PR #2731, opened 2018-08-16
Changes to LASCO color tables

### LINKED ISSUE #2724: LASCO C3 Maps from JP2 sources are badly stretched
See attached: probably need to fix the stretching.

figure_1

wafels:
The existing LASCO color tables don't work well with helioviewer JP2 images.  The color tables are highly compressed with over half the colors simply being white, so the colors are crammed in to an unnecessarily small range of values.  This PR replaces those color tables with matplotlib colormaps that make nice looking helioviewer JP2 images and reasonable looking images of Level 0.5 image data (bearing in mind that Level 0.5 data is not background subtracted).  This PR fixes #2724 .


Cadair:
ooo a bold move!!

What's the justification for removing them rather than just making them not the default?

wafels:
I suppose we could keep them for historical reasons but they are really bad color tables.  They are highly compressed, and in my opinion work _against_ the desire to see features in the data. The new C2 color table looks superficially like the old C2 color table, so I expect few complaints about that.  The C3 color table is quite a bit different, but you do see more detail.  I think these are improvements to SunPy's ability to display LASCO images nicely.  The only reason to keep them is for historical reasons.

Cadair:
I agree. imo it's probably worth keeping them in `sunpy.cm` just in case people want them for something specific and weird.

sunpy/cm/color_tables.py
```diff
@@ -349,6 +356,23 @@ def lasco_color_table(number):
...
+    if number == 2:
+        rgb = cm.get_cmap('gist_heat')(np.linspace(0, 1, 256)) * 255
+    elif number == 3:
+        rgb = cm.get_cmap('bone')(np.linspace(0, 1, 256)) * 255
+    else:
+        raise ValueError("Invalid LASCO number. Valid values are 2, 3.")
+
+    # Now create the color dictionary in the correct format
+    cdict = create_cdict(rgb[:, 0], rgb[:, 1], rgb[:, 2])
+    return colors.LinearSegmentedColormap('SOHO LASCO C{:s}'.format(str(number)), cdict)
```
Cadair: These cms and the old ones have the same name.

Do we need this? Could we not just change the map defaults to the mpl names?

sunpy/map/sources/soho.py
```diff
@@ -130,7 +130,7 @@ def __init__(self, data, header, **kwargs):
             self.meta['date_obs'] = self.meta['date-obs']
         self._nickname = self.instrument + "-" + self.detector
         self.plot_settings['cmap'] = plt.get_cmap('soholasco{det!s}'.format(det=self.detector[1]))
-        self.plot_settings['norm'] = ImageNormalize(stretch=source_stretch(self.meta, PowerStretch(0.5)))
+        self.plot_settings['norm'] = ImageNormalize(stretch=source_stretch(self.meta, LinearStretch()))
```
Cadair: Is this also the best stretch for the actual calibrated lasco data?

sunpy/cm/color_tables.py
wafels: I did this for consistency with the names we give to the other solar physics derived color tables.

sunpy/map/sources/soho.py
wafels: Level 0.5 and Level 1 data don't look great with any amount of stretch since you really need to subtract a background model in order to see anything like a CME.  Slightly better choices might be, say PowerStretch(0.5), but the advantage is not great.

sunpy/cm/color_tables.py
wafels: I'm not super happy with the name of the historical color table functions.  Maybe we should write "instrument_team_lasco_color_table", or "sswidl_lasco_color_table" ?

ehsteve: I like ''sswidl_lasco_color_table''. I agree we should keep the old version just in case.

sunpy/map/sources/soho.py
```diff
@@ -11,7 +11,7 @@
 from matplotlib import colors
 
 from astropy.units import Quantity
-from astropy.visualization import PowerStretch
```
ehsteve: Why are you importing LinearStretch if you then don't use it?

sunpy/cm/color_tables.py
wafels: Ok.

sunpy/map/sources/soho.py
wafels: removed.

Cadair:
I think you will have to make a change in `cm/cm.py` to register the legacy colormaps with matplotlib?

sunpy/cm/cm.py
```diff
@@ -137,14 +143,14 @@ def get_cmap(name):
         raise ValueError("Colormap {name!s} is not recognized".format(name=name))
 
 
-def show_colormaps(filter=None):
+def show_colormaps(search=None):
```
Cadair: can you add a second changelog file under `breaking` for this please?

sunpy/cm/color_tables.py
```diff
@@ -341,14 +348,30 @@ def lasco_color_table(number):
...
     return colors.LinearSegmentedColormap('SOHO LASCO C{:s}'.format(str(number)), cdict)
 
+
+def lasco_color_table(number):
+    """Returns one of the fundamental color tables for SOHO LASCO images.
+    These are not the same as those used in SSWIDL.  This is because the
+    SSWIDL color scaling for LASCO level 0.5 and 1.0 is highly compressed
+    and does not display the data well."""
+    if number == 2:
+        rgb = cm.get_cmap('gist_heat')(np.linspace(0, 1, 256)) * 255
```
Cadair: This seems... redundant?

The operation we want to do here is to give the same colourmap a different name. We can do that with a simple call to https://matplotlib.org/api/cm_api.html#matplotlib.cm.register_cmap

I think therefore my advice is to move the contents of this function directly into `cm/cm.py` and just call

```python
mpl.cm.register_cmap("SOHO LASCO C2", mpl.cm.get_cmap("gist_heat"))
```

Cadair: Also as an aside, we should probably not use `import matplotlib.cm as cm` in a sunpy submodule named `cm` where there is also a file called `cm` it get's very confusing very quick!!

sunpy/cm/cm.py
wafels: So I add a file called 2731.breaking.rst ?

Cadair: yes

sunpy/cm/color_tables.py
```diff
@@ -329,8 +331,12 @@ def eit_color_table(wavelength):
         244.,  248.,  250.]), 255 * np.ones(181)))
 
 
-def lasco_color_table(number):
-    """Returns one of the fundamental color tables for SOHO LASCO images."""
+def sswidl_lasco_color_table(number):
+    """Returns one of the previously used color tables for SOHO LASCO images.
```
ehsteve: Remove use of word "previous" as this will likely not mean much in a few years. Replace with SSW-defined tables.

PR #2735, opened 2018-08-31
Revert disallowing a no-op HPC<>HPC transformation without obstime

Cadair:
On second thougts it dosen't actually make any sense to allow a noop transfrom with obstime if we are not allowing it without obstime. Appart from anything else, this breaks doing a `frame.transform_to` blindly to make sure your frame is what you expect it to be.

### DIFF
--- a/sunpy/coordinates/transformations.py
+++ b/sunpy/coordinates/transformations.py
@@ -57,7 +57,10 @@ def _carrington_offset(obstime):
     return get_sun_L0(obstime)
 
 
-def _observers_are_equal(obs_1, obs_2):
+def _observers_are_equal(obs_1, obs_2, string_ok=False):
+    if string_ok:
+        if obs_1 == obs_2:
+            return True
     if not (isinstance(obs_1, BaseCoordinateFrame) and isinstance(obs_2, BaseCoordinateFrame)):
         raise ValueError("To compare two observers, both must be instances of BaseCoordinateFrame. "
                          "Cannot compare two observers {} and {}.".format(obs_1, obs_2))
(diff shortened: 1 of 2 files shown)


sunpy/coordinates/transformations.py
```diff
@@ -57,7 +57,10 @@ def _carrington_offset(obstime):
     return get_sun_L0(obstime)
 
 
-def _observers_are_equal(obs_1, obs_2):
+def _observers_are_equal(obs_1, obs_2, string_ok=False):
+    if string_ok:
+        if obs_1 == obs_2:
```
wtbarnes: Shouldn't this just be `return obs_1 == obs_2`

Cadair: I didn't do that because if its false we need to run the rest of the function.

sunpy/coordinates/transformations.py
```diff
@@ -57,7 +57,10 @@ def _carrington_offset(obstime):
     return get_sun_L0(obstime)
 
 
-def _observers_are_equal(obs_1, obs_2):
+def _observers_are_equal(obs_1, obs_2, string_ok=False):
+    if string_ok:
```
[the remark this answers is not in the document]
Cadair: yeah I *think* it only applies to the `hpc_to_hpc` transform, so I added it as a flag rather than anything else.

sunpy/coordinates/transformations.py
wtbarnes: Oh I see now. Yeah I think I did the same thing when I wrote the first version of this function...and then forgot!

Issue #2741, opened 2018-09-07
Add a HMI overplotting on AIA example


Cadair:
Should use Composite map to over plot the magnetic field strength contours on an AIA image.

Chicrala:
The idea is to create an example like the ones that are accessible via tutorial from a sample data or make a general routine for this?

I'm currently making plots of Poynting flux over HMI Br component so I may be able to help with this if I understand it correctly.

Cadair:
We already have this functionality in `sunpy.map.compositemap`, the issue is to write an example like [this](http://docs.sunpy.org/en/stable/generated/gallery/maps/map_rotation.html#sphx-glr-generated-gallery-maps-map-rotation-py) which live in the [`examples/`](https://github.com/sunpy/sunpy/tree/master/examples) directory.

Cadair:
(note: composite map will only over plot if the data is already aligned, so both the AIA and HMI image should be thrown through `aiaprep` first)

(note, note: we don't actually have any HMI data in our sample data at the moment, so we should really add that as well)

Chicrala:
I think I understood.
May I try to solve this then?

Cadair:
Just open the Pull Request, it makes it a lot easier to give feedback on :smiley:

Chicrala:
I haven't thought on this haha

PR #2749, opened 2018-09-11
Aiaprep fix

### LINKED ISSUE #2747: aiaprep can not be used to align AIA and HMI images
`aiaprep` explicitly checks the type of the map to make sure it's `AIAMap` when really it should allow `HMIMap` as well because then you can align AIA images with HMI, and the metadata structure is sufficiently similar that there is no reason why this would not work.

The fix I think would be best would be to allow both `AIAMap` and `HMIMap` in [this](https://github.com/sunpy/sunpy/blob/master/sunpy/instr/aia.py#L45) `isinstance` check. On the other hand you could just drop it altogether and just hope it works :laughing:

SolarDrew:
Allow `aiaprep()` to take a HMI map as input, plus associated test. Wasn't sure what the 'correct' HMI file for this purpose would be so I used the only one I had locally in the test data folder with 'hmi' in the name. Happy to change it to a better file if there is one.

When I find some time (hopefully soon) I'll change the name of the function to reflect this change but I wanted to get this PR'd so people can start using it.

Closes #2747

### DIFF
--- a/changelog/2749.bugfix.rst
+++ b/changelog/2749.bugfix.rst
@@ -1,0 +1,1 @@
+Add support for `HMIMap` objects as input to `sunpy.instr.aia.aiaprep()`.
--- a/sunpy/instr/aia.py
+++ b/sunpy/instr/aia.py
@@ -5,7 +5,7 @@ Provides processing routines for data captured with the AIA instrument on SDO.
 import numpy as np
 import astropy.units as u
 
-from sunpy.map.sources.sdo import AIAMap
+from sunpy.map.sources.sdo import AIAMap, HMIMap
 
 __all__ = ['aiaprep']
 
@@ -42,7 +42,7 @@ def aiaprep(aiamap):
     therefore differ from the original file.
     """
 
-    if not isinstance(aiamap, AIAMap):
+    if not isinstance(aiamap, (AIAMap, HMIMap)):
         raise ValueError("Input must be an AIAMap")
 
     # Target scale is 0.6 arcsec/pixel, but this needs to be adjusted if the map
--- a/sunpy/instr/tests/test_aia.py
+++ b/sunpy/instr/tests/test_aia.py
@@ -13,9 +13,11 @@ from sunpy.instr.aia import aiaprep
 # functions
 
 
-@pytest.fixture
-def original():
-    return sunpy.map.Map(test.get_test_filepath("aia_171_level1.fits"))
+@pytest.fixture(scope="module",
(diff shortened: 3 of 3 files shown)


wtbarnes:
Doesn't this still assume a plate scale of 0.6 arcseconds per pixel? Which is not true for HMI?

larrymanley:
For purposes of over-plotting HMI should be re-scaled to 0.6 arcsec per
pixel.  Maybe add a plate_scale keyword with a default of 0.6 and document
that native HMI is 0.5?

SolarDrew:
Should the WCS not account for the difference in plate scale if you try to overplot HMI on AIA? Either way, I was thinking hard-code the native HMI scale in for HMI input. I think adding this option to `Map.rotate()` and telling people to use that instead for cases where they want to overplot makes more sense, and `aiaprep()` should be kept for the one very narrow purpose of just making scaling and rotation consistent between images.

wtbarnes:
It might be best to change the name of this function as well once it is applicable to HMI data as `aiaprep` would be a bit misleading.

Cadair:
> Should the WCS not account for the difference in plate scale if you try to overplot HMI on AIA?

Yes, if you use the `WCSAxes` contouring routine. However, if you want image overlays then no, you need to regrid the pixels using either this or reproject.

> Maybe add a plate_scale keyword with a default of 0.6 and document that native HMI is 0.5?

I would rather see this feature implemented directly in `GenericMap.rotate` where as well as the scale factor you can currently specify you could specify a *target* pixel scale.

Im my eyes the purpose of this change is to allow people to align HMI data to AIA data if they know what they are doing. I think keeping the name as `aiaprep` and the default pixel scale as 0.6 makes sense. If we were to call this `hmiprep` then I would expect it to scale to 0.5. I think we should be trying to take functionality out of this function not put more in :wink:

Cadair:
Unless anyone else whishes to disagree with my opinion that this patch is acceptable as-is, I will merge it as I would like to do a 0.9.3 release.

I do think we should look at further improvement to `aiaprep` and `GenericMap.rotate` in tandem.

wtbarnes:
Ok that sounds good to me. I didn't know if the 0.6 plate scale would be an issue or not. Seeing as aligning AIA and HMI images is an extremely useful capability, I'm all for this fix. 

We should also include an example in the gallery of how to do AIA/HMI overplotting once this is merged.

Cadair:
For the latter see #2741

Issue #2712, opened 2018-08-06
Incorrect transformation between HCC and HPC with different observers


wtbarnes:
There seems to be a bug when transforming between a Heliocentric Cartesian (HCC) frame and a Helioprojective (HPC) frame, where the two frames are defined by different observers. 

Best illustrated by example (see below).

```python
>>> from astropy.coordinates import SkyCoord
>>> import astropy.units as u
>>> import astropy.constants as const
>>> from sunpy.coordinates import Heliocentric,Helioprojective
# First observer
>>> obs_1 = SkyCoord(lon=0*u.deg, lat=0*u.deg, radius=const.au, frame='heliographic_stonyhurst')
# HCC frame defined by first observer
>>> hcc_frame = Heliocentric(observer=obs_1)
# Point on the surface of the Sun in the new HCC frame
>>> p = SkyCoord(x=0*u.m,y=0*u.m,z=const.R_sun,frame=hcc_frame)
# Second observer rotated 90 deg in longitude
>>> obs_2 = SkyCoord(lon=90*u.deg, lat=0*u.deg, radius=const.au, frame='heliographic_stonyhurst')
# HPC frame defined by second observer
>>> hpc_frame = Helioprojective(observer=obs_2)
```

Because this second observer is rotated by 90 degrees, we would expect `p`, expressed in the HPC frame, to be on the limb and to the left, i.e. theta_x <0. 

```python
>>> p.transform_to(hpc_frame)
<SkyCoord (Helioprojective: obstime=None, rsun=695508.0 km, observer=<HeliographicStonyhurst Coordinate (obstime=None): (lon, lat, radius) in (deg, deg, m)
    (90., 0., 1.49597871e+11)>): (Tx, Ty, distance) in (arcsec, arcsec, km)
    (0., 0., 1.48902171e+08)>
```

This is not the case. It stays at the center of the disk for some reason. However, if we do an intermediate transform to an HGS frame, we get the expected result,

```python
>>> p.transform_to('heliographic_stonyhurst').transform_to(hpc_frame)
<SkyCoord (Helioprojective: obstime=None, rsun=695508.0 km, observer=<HeliographicStonyhurst Coordinate (obstime=None): (lon, lat, radius) in (deg, deg, m)
    (90., 0., 1.49597871e+11)>): (Tx, Ty, distance) in (arcsec, arcsec, km)
    (-959.22081361, 0., 1.49599488e+08)>
```

[...]

PR #2754, opened 2018-09-16
Functionality to convert hpc to hcc with different observer and vice versa issue: #2712

souravghosh97:
Previously an user cannot convert hpc to hcc for different observer (#2712). This changes will allow user to convert between hpc to hcc. This is done by converting hcc ->hgs->hcc 

Closes #2754


sunpy/coordinates/tests/test_transformations.py
```diff
@@ -225,21 +224,40 @@ def test_hcc_to_hpc_different_observer():
...
+    # This test checks transformation HCC->HCC in the case of same observer
+
+    D0 = 1 * u.km
+    observer_1 = HeliographicStonyhurst(lat=0 * u.deg, lon=0 * u.deg, radius=D0)
+    observer_2 = observer_1
+    hcc_frame = Heliocentric(observer=observer_1)
+    hpc_frame = Helioprojective(observer=observer_2)
+    hpccoord = SkyCoord(Tx=0 * u.arcsec, Ty=0 * u.arcsec, frame=hpc_frame)
+    hcccoord_out = hpccoord.transform_to(hcc_frame)
+    hcccoord_expected = hpccoord.transform_to(HeliographicStonyhurst).transform_to(hcc_frame)
```
Cadair: Rather than doing this transformation here, why not just compare to the original values to make sure they have not changed?

sunpy/coordinates/transformations.py
```diff
@@ -117,12 +117,9 @@ def hcc_to_hpc(helioccoord, heliopframe):
...
     """
     if not _observers_are_equal(helioccoord.observer, heliopframe.observer):
-        raise ConvertError("Cannot directly transform heliocentric coordinates to "
-                           "helioprojective coordinates for different "
-                           "observers {} and {}. See discussion in this GH issue: "
-                           "https://github.com/sunpy/sunpy/issues/2712. Try converting to "
-                           "an intermediate heliographic Stonyhurst frame.".format(
-                               helioccoord.observer, heliopframe.observer))
+        heliocframe = Heliocentric(observer=heliopframe.observer)
+        new_helioccoord = hcc_to_hcc(helioccoord,heliocframe)
```
Cadair: Rather than explicitly calling the transformation function here, this should use the transformation machinery:
```python
new_helioccoord = helioccoord.transform_to(heliocframe)
```

sunpy/coordinates/transformations.py
```diff
@@ -148,12 +145,10 @@ def hpc_to_hcc(heliopcoord, heliocframe):
...
     if not _observers_are_equal(heliopcoord.observer, heliocframe.observer):
-        raise ConvertError("Cannot directly transform helioprojective coordinates to "
-                           "heliocentric coordinates for different "
-                           "observers {} and {}. See discussion in this GH issue: "
-                           "https://github.com/sunpy/sunpy/issues/2712. Try converting to "
-                           "an intermediate heliographic Stonyhurst frame.".format(
-                               heliopcoord.observer, heliocframe.observer))
+        heliocframe_heliopobs = Heliocentric(observer=heliopcoord.observer)
+        helioccoord_heliopobs = heliopcoord.transform_to(heliocframe_heliopobs)
+        helioccoord = hcc_to_hcc(helioccoord_heliopobs,heliocframe)
```
Cadair: Same here, should call `transform_to`.

sunpy/coordinates/tests/test_transformations.py
souravghosh97: I think i have done a mistake here. The function name should be ```test_hpc_to_hcc_same_observer()``` instead of ```test_hcc_to_hcc_same_observer```.

Cadair: oh! makes sense!

souravghosh97: there is need for another test_hcc_to_hcc_same_observer

sunpy/coordinates/tests/test_transformations.py
```diff
@@ -225,21 +224,40 @@ def test_hcc_to_hpc_different_observer():
...
+    assert_quantity_allclose(hcccoord_out.x, hcccoord_expected.x)
+    assert_quantity_allclose(hcccoord_out.y, hcccoord_expected.y)
+    assert_quantity_allclose(hcccoord_out.z, hcccoord_expected.z)
+
+def test_hcc_to_hcc_same_observer():
+    # This test checks transformation HCC->HCC in the case of same observer
+
+    D0 = 1 * u.km
+    observer_1 = HeliographicStonyhurst(lat=0 * u.deg, lon=0 * u.deg, radius=D0)
+    observer_2 = observer_1
```
wtbarnes: Why are there two observers? I think you only need one as you are checking HPC->HCC for the same observer. Also as you noted below, test name should be changed.

sunpy/coordinates/transformations.py
```diff
@@ -361,4 +356,18 @@ def hgs_to_hgs(from_coo, to_frame):
         return from_coo.transform_to(HCRS).transform_to(to_frame)
 
 
+@frame_transform_graph.transform(FunctionTransform, Heliocentric, Heliocentric)
+def hcc_to_hcc(from_coo, to_frame):
```
wtbarnes: To be consistent with the other transform functions, I think the input args should be `hcccoord` and `hccframe`. I realize that the `hgs_to_hgs` transform uses this same naming, though I would claim that that should be changed as well.

sunpy/coordinates/transformations.py
```diff
@@ -361,4 +356,18 @@ def hgs_to_hgs(from_coo, to_frame):
...
+    """
+    Convert from  Heliocentric to Heliocentric
+    """
+    if _observers_are_equal(from_coo.observer, to_frame.observer, string_ok=True):
+        return to_frame.realize_frame(from_coo._data)
+
+    hgs = from_coo.transform_to(HeliographicStonyhurst)
```
wtbarnes: Similar comment to above. Naming scheme should be a bit more explicit. `hgs`->`hgscoord`, `hcc`->`hcccoord`.

changelog/2754.bugfix.rst
```diff
@@ -0,0 +1 @@
+User can convert between HPC and HCC Co-ordinates with different observers
```
wtbarnes: Typo. "Co-ordinates" should be "coordinates"

Cadair:
Thanks a lot @souravghosh97 !!

Issue #2760, opened 2018-09-24
AttributeError: 'str' object has no attribute 'decode'


MjdMahasneh:
Hello! so am attempting to query Helioviewer.org with SunPy and i get this AttributeError randomly (sometime i get it and sometimes i dont when using the same exact parameters) this is a very strange behaviour and i tried to understand whats causing it but i couldnt figure it out. any help in this regard is very much appreciated.

DeltiKron:
Hi, I've been seeing  the same problem since some time last week and would also be interested in what is causing this. I have the impression that the error occurs more frequently if files are downloaded in quick succession. Usually waiting for a minute or so and rerunning the same command works for me. (ofc this is still not ideal for batch operations... ) Maybe the cause is some caching or temporary file error?

MjdMahasneh:
this can be something to look at and can still be fixed as usually data will be accessed as batches and this issue makes it very slow and inconvenient.

nabobalis:
Can you provide a small example that I can run to debug this please?

DeltiKron:
```
from sunpy.net.helioviewer import HelioviewerClient
import datetime

# get hmi reference
hv = HelioviewerClient()
date = datetime.datetime.now()
file = hv.download_jp2(
    date,
    observatory='SDO',
    instrument='HMI',
    detector='HMI',
    measurement='continuum'
    )

```
for me this is an example sufficient to produce the error. I ran it twice just now and one time the error popped up and one time it worked as expected.

MjdMahasneh:
i can indeed but the code is customised for a certain mission and depends on a data set that i have so it wont work on your machine. however here is the main part of the code that generates the queries:


`
hv = HelioviewerClient()
#loop through the list of required images  
for i, f in enumerate(imlist): 
  corrosponding_EIT_name = f #where f is a string of format date-time i.e. '2012/07/05 00:30:00'
 
 #download image
 file = hv.download_png(corrosponding_EIT_name, 2.4, '[1,1,100]', x0=0, y0=0, width=1024, height=1024)

`

nabobalis:
Thanks, I'll look into this now.

MjdMahasneh:
@nabobalis Also it would be nice if one could download the .fits or at least being able to access the header part of the fits file that corresponds to the png image.

Many thank in advance :)

nabobalis:
I think I have a fix locally, I will see about pushing up and getting a review and hopefully a bug fix release this week. 

[...]

PR #2761, opened 2018-09-24
[0.9] Helioviewer default file type change

nabobalis:
To address #2760.

It seems that because `default=u'file'`, when name is not returned it tries to decode a `str` which is correct. 

Honestly, I am not sure if this is the best fix but it seems to not fail anymore for me. 
Because of this, I was not planning to submit this to master but just the 0.9 branch.

### DIFF
--- a/changelog/2761.bugfix.rst
+++ b/changelog/2761.bugfix.rst
@@ -1,0 +1,1 @@
+Changed default file type for Helioviewer to prevent decode errors.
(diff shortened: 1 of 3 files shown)


nabobalis:
DELETE IT INSTEAD

nabobalis:
WHY WONT YOU LEAVE ME ALONE

PR #2764, opened 2018-09-24
Alternative Fix for Helioviewer

nabobalis:
Just delete it? 

Seems to not error locally.

### DIFF
--- a/sunpy/net/helioviewer.py
+++ b/sunpy/net/helioviewer.py
@@ -133,6 +133,7 @@ class HelioviewerClient(object):
 
         >>> data_sources = hv.get_data_sources()  # doctest: +REMOTE_DATA
         >>> file = hv.download_jp2('2012/07/03 14:30:00', sourceId=data_sources['SOHO']['LASCO']['C2']['white-light']['sourceId'])   # doctest: +REMOTE_DATA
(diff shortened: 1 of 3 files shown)


sunpy/net/helioviewer.py
```diff
@@ -133,6 +133,7 @@ def download_jp2(self, date, directory=None, overwrite=False, **kwargs):
 
         >>> data_sources = hv.get_data_sources()  # doctest: +REMOTE_DATA
         >>> file = hv.download_jp2('2012/07/03 14:30:00', sourceId=data_sources['SOHO']['LASCO']['C2']['white-light']['sourceId'])   # doctest: +REMOTE_DATA
+        >>> file = hv.download_jp2(datetime.datetime.now(), observatory='SDO', instrument='HMI', detector='HMI', measurement='continuum')   # doctest: +REMOTE_DATA
```
wafels: Shouldn't have _file =_  - file is a built-in object.

wafels:
I don't think this PR is necessary.  The original bug report was probably due to the Helioviewer API sending out malformed data due to the server being overloaded by requests.  I was able to reproduce the original bug report during a time of heavy usage, and at time of writing (when we don't have a lot of demand) the download_jp2 request works just fine.

nabobalis:
In this case the API would return a blank string for the filename. That then triggers the `if not` statement which then goes to decode a `str`. I think there is a bug here.

PR #2766, opened 2018-09-28
Make sure package does not import itself during setup

drdavella:
This bug came up during a discussion with @Cadair while we were trying to test `egg_info` in a completely clean environment. More on that later.

This is only a problem when running `setup.py` for the very first time in a completely clean environment, which is why most users/developers never notice it. It also has the curious property of not affecting release builds (where `.dev` is not in the version string), which is why it doesn't affect users who install the package using `pip` or `conda`.

### DIFF
--- a/sunpy/__init__.py
+++ b/sunpy/__init__.py
@@ -21,14 +21,21 @@ try:
 except ImportError:
     __githash__ = ''
 
-import os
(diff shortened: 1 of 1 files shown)


Cadair:
Thanks @drdavella !

PR #2769, opened 2018-10-01
Clean up some invalid escape sequences

dstansby:


nabobalis:
Seems to cause some doc build failures. Out of the 6, I could only find two (SunPy build log is too messy). 

```/root/project/docs/code_ref/net.rst:34:<autosummary>:1: WARNING: Inline emphasis start-string without end-string.```

```/root/project/docs/code_ref/net.rst:34:<autosummary>:1: WARNING: Inline strong start-string without end-string.```

PR #2770, opened 2018-10-02
Fix/copyedit outdated documentation

### LINKED ISSUE #2763: Solar Cycle Gallery example out of date
The example includes the following text 

> For this example we will use the SunPy sample data, if you want the current data, delete the argument to the create function. i.e. noaa = lc.NOAAIndicesLightCurve.create()

This text is inline and therefore not checked during build so was not caught. This should be fixed and this behavior should be discouraged.

astroengisci:
Fixes #2763 

Removed the outdated instruction as well as copyedited the documentation for readability and professionalism.

### DIFF
--- a/changelog/2770.doc.rst
+++ b/changelog/2770.doc.rst
@@ -1,0 +1,1 @@
+Copyedited and corrected the solar cycles example.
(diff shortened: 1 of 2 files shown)


nabobalis:
This looks good, could you please add a changelog entry?

PR #2771, opened 2018-10-04
 [0.9] Helioviewer default file type change

### LINKED ISSUE #2760: AttributeError: 'str' object has no attribute 'decode'

nabobalis:
Change only for 0.9. No intention to patch master. 

Fixes #2760.

It seems that because `default=u'file'`, when name is not returned it tries to decode a `str` which is correct.


nabobalis:
Rhessi test fail due to trying to creating a lightcurve with today's date when there has been no data for the last few days online.

sunpy/util/net.py
```diff
@@ -78,18 +78,18 @@ def get_filename(sock, url):
...
-def get_system_filename(sock, url, default=u"file"):
+def get_system_filename(sock, url, default="file"):
     """ Get filename from given urllib2.urlopen object and URL.
     First, attempts to extract Content-Disposition, second, extract
     from URL, eventually fall back to default. Returns bytestring
     in file system encoding. """
     name = get_filename(sock, url)
     if not name:
-        name = default.decode("ascii", "ignore")
+        name = six.text_type(default).decode("ascii", "ignore")
```
Cadair: drop the decode here.

Cadair:
I still think this needs to be on master

Issue #2774, opened 2018-10-08
Creating LASCO C3 Map object fails


dstansby:
```python
from sunpy.net import Fido, attrs as a
import sunpy.map as smap

timerange = a.Time('1996/05/24 11:00', '1996/05/24 12:00')
instrument = a.Instrument('C3')

result = Fido.search(timerange, instrument)
downloaded_files = Fido.fetch(result)

m = smap.Map(downloaded_files[0])
```

Fails with
```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-25-a94b4971e236> in <module>()
----> 1 m = smap.Map(downloaded_files[0])

~/miniconda3/lib/python3.7/site-packages/sunpy/map/map_factory.py in __call__(self, *args, **kwargs)
    285 
    286             try:
--> 287                 new_map = self._check_registered_widgets(data, meta, **kwargs)
    288             except (NoMatchError, MultipleMatchError, ValidationFunctionError):
    289                 if not silence_errors:

~/miniconda3/lib/python3.7/site-packages/sunpy/map/map_factory.py in _check_registered_widgets(self, data, meta, **kwargs)
    341         WidgetType = candidate_widget_types[0]
    342 
--> 343         return WidgetType(data, meta, **kwargs)
    344 
    345 

~/miniconda3/lib/python3.7/site-packages/sunpy/map/sources/soho.py in __init__(self, data, header, **kwargs)
    111         GenericMap.__init__(self, data, header, **kwargs)
    112 
--> 113         self.meta['CUNIT1'] = self.meta['CUNIT1'].lower()
    114         self.meta['CUNIT2'] = self.meta['CUNIT2'].lower()
    115 

~/miniconda3/lib/python3.7/site-packages/sunpy/util/metadata.py in __getitem__(self, key)
     41     def __getitem__(self, key):
     42         """Override [] indexing"""
---> 43         return OrderedDict.__getitem__(self, key.lower())
     44 
     45     def __setitem__(self, key, value):

KeyError: 'cunit1'
```

Cadair:
the `self.meta['CUNIT2'].lower()` should probably be re-written as `self.meta.get('CUNIT2', 'arcsec').lower()`

PR #2775, opened 2018-10-09
Fix LASCOMap

dstansby:
Suggested by @Cadair here: https://github.com/sunpy/sunpy/issues/2774#issuecomment-427869475 - fixes the problem locally for me. Maybe needs a test adding too?

### DIFF
--- a/changelog/2775.bugfix.rst
+++ b/changelog/2775.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed loading in LASCO C3 data.
--- a/sunpy/map/sources/soho.py
+++ b/sunpy/map/sources/soho.py
@@ -110,8 +110,8 @@ class LASCOMap(GenericMap):
 
         GenericMap.__init__(self, data, header, **kwargs)
 
-        self.meta['CUNIT1'] = self.meta['CUNIT1'].lower()
-        self.meta['CUNIT2'] = self.meta['CUNIT2'].lower()
+        self.meta['CUNIT1'] = self.meta.get('CUNIT1', 'arcsec').lower()
+        self.meta['CUNIT2'] = self.meta.get('CUNIT2', 'arcsec').lower()
 
         # Fill in some missing or broken info
         # Test if change has already been applied


Cadair:
I would be interested in knowing what @wafels or someone else who has a little more idea of the LASCO data thinks about this. In general we should probably shouldn't assume the existence of critical headers anyway.

Cadair:
FTR we assume arcsec later on anyway: https://github.com/sunpy/sunpy/blob/master/sunpy/map/mapbase.py#L693

nabobalis:
Do we not have a test for each map source? 

@dstansby Just add a changelog entry and we can merge this.

dstansby:
I think this might be wrong... Trying to draw the limb defaults to drawing it at ~950arsec (assuming an Earth based observer, which SOHO basically is right?), which is not where the disk is in the created LASCO maps after this fix.

Cadair:
your lasco data is missing all the required metadata to calculate where the limb should be, see discussion in #2788

dstansby:
Yes, but assuming the units are arsec is incorrect (as this PR does) - is there an option to just leave the data in "pixel number" and not physical units? Or to error instead if there isn't the right metadata? Either of those options would be much preferred to silently doing the wrong thing.

Cadair:
As I mentioned further up, we are already assuming arcsec in `GenericMap` anyway. So if we are going to change that then it's bigger than this PR.

It's also worth noting that Map is designed to support images *with WCS* metadata. While I agree with your comment on matrix that we should probably throw a better error, I am really not sure we want to support a completely metadata-less Map.

dstansby:
@sunpy/sunpy-developers - can someone revert this? It's definitely the wrong thing to do (https://github.com/sunpy/sunpy/pull/2775#issuecomment-428933507) ...

nabobalis:
It doesn't work? Should it be set to no units?

dstansby:
The LASCO maps that I am downloading and reading do not have units of 'arcsec' (I don't know what units they do have...), so at the very least it shouldn't silently assume 'arcsec'.

PR #2787, opened 2018-10-11
Bare bones downloading and plotting example

### LINKED ISSUE #2772: Need a clear tutorial on downloading and viewing data
Although there seem to be some examples/tutorials in the documentation that use the sample data, there are no obvious examples on how to download and then view/analyse non-sample data. The ones I can find are

- http://docs.sunpy.org/en/stable/generated/gallery/acquiring_data/2011_06_07_sampledata_overview.html#sphx-glr-generated-gallery-acquiring-data-2011-06-07-sampledata-overview-py - only has sample data
- http://docs.sunpy.org/en/stable/generated/gallery/acquiring_data/skip_searching_vso.html#sphx-glr-generated-gallery-acquiring-data-skip-searching-vso-py - only shows how to search for data
- http://docs.sunpy.org/en/stable/guide/tour.html#obtaining-data - only shows how to search for data
- http://docs.sunpy.org/en/stable/guide/tour.html#maps only has sample data.

And that's it really... A linked up tutorial on searching for, downloading, and plotting some data, which assumes no previous knowledge of SunPy would be much appreciated.

dstansby:
Fixes #2772. This is a very bare bones example for downloading and plotting LASCO C3 data - if anyone wants to make changes or improvements feel free to push to my branch.

### DIFF
--- a/examples/acquiring_data/downloading_and_plotting.py
+++ b/examples/acquiring_data/downloading_and_plotting.py
@@ -1,37 +1,56 @@
 """
-=============================
-Downloading and plotting data
-=============================
+======================================
+Downloading and plotting LASCO C3 data
+======================================
+
+In this example, we show how to acquire and plot SOHO/LASCO C3 data using SunPy.
 """
-###############################################################################
 # Import the required modules.
 from sunpy.net import Fido, attrs
 import sunpy.map
-import matplotlib.pyplot as plt
+from sunpy.io.file_tools import read_file
 
 ###############################################################################
-# Define the timerange and instrument.
+# In order to download the required data, we will use
+# `Fido <sunpy.net.fido_factory.UnifiedDownloaderFactory>`, a downloader client.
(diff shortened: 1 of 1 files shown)


examples/acquiring_data/downloading_and_plotting.py
```diff
@@ -0,0 +1,37 @@
...
+###############################################################################
+# Define the timerange and instrument.
+timerange = attrs.Time('1996/05/24 11:00', '1996/05/24 12:00')
+instrument = attrs.Instrument('C3')
+
+###############################################################################
+# Search for data that is from the requested instrument and in the requested
+# timerange.
+result = Fido.search(timerange, instrument)
+result
```
dpshelio: If we use this as a script `results` won't be printed. If you want them to be printed, then we should have `print(results)`.

nabobalis:
@dstansby I thought I would rebase the pull request and add some comments here to pad out the example. 

I think long term we actually want an example for each map source.

PR #2796, opened 2018-10-16
Fix Slicing a JSOC UnifiedResponse

### LINKED ISSUE #2781: Slicing a JSOC UnifiedResponse object causes Fido.fetch to fail
```python
tstart = '2011/06/07 06:32:45'
tend = '2011/06/07 06:33:15'                                                                                                                                                                 

res = Fido.search(a.jsoc.Time(tstart, tend), a.jsoc.Series('hmi.M_45s'))

Fido.fetch(res[0,0])
```
```
<ipython-input-6-488027c4a327> in <module>
----> 1 Fido.fetch(res[0,0])

~/Git/sunpy/sunpy/net/fido_factory.py in fetch(self, *query_results, **kwargs)
    360         for query_result in query_results:
    361             for block in query_result.responses:
--> 362                 reslist.append(block.client.fetch(block, **kwargs))
    363 
    364         results = DownloadResponse(reslist)

~/Git/sunpy/sunpy/net/jsoc/jsoc.py in fetch(self, jsoc_response, path, overwrite, progress, max_conn, downloader, sleep)
    532 
    533         # Make staging request to JSOC
--> 534         responses = self.request_data(jsoc_response)
    535         # Make response iterable
    536         if not isiterable(responses):

~/Git/sunpy/sunpy/net/jsoc/jsoc.py in request_data(self, jsoc_response, **kwargs)
    428         requests = []
    429         self.query_args = jsoc_response.query_args
--> 430         for block in jsoc_response.query_args:
    431 
    432             ds = self._make_recordset(**block)

TypeError: 'NoneType' object is not iterable
```

eamanu:
### Description

The problem reported on issue #2781  is because, when slice a JSOC UnifiedResponse:

`_handle_record_slice(...)`

the query_args is not set.

Fixes #2781

### DIFF
--- a/changelog/2796.bugfix.rst
+++ b/changelog/2796.bugfix.rst
@@ -1,0 +1,1 @@
+Fix the slicing a JSOC UnifiedResponse object. This bug was that when slice a Unifiedresponse this not saved the query_args attribute.
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -40,10 +40,24 @@ class JSOCResponse(Sequence):
         """
         table : `astropy.table.Table`
         """
-
-        self.table = table
-        self.query_args = None
-        self.requests = None
+        if isinstance(table, list):
+            # fido_factory._handle_record_slice
+            # will return a list containing the
+            # response that has been indexed,
+            # so we have to remove that.
+            new_table = table[0]
+            if isinstance(new_table, type(self)):
+                self.table = new_table.table
+                self.query_args = new_table.query_args
+                self.requests = new_table.requests
+            else:
+                self.table = table
+                self.query_args = None
+                self.requests = None
+        else:
+            self.table = table
+            self.query_args = None
+            self.requests = None
 
     def __str__(self):
         return str(self.table)
@@ -61,7 +75,10 @@ class JSOCResponse(Sequence):
             return len(self.table)
 
     def __getitem__(self, item):
-        return type(self)(self.table[item])
+        resp = type(self)(self.table[item])
+        resp.query_args = self.query_args
+        resp.requests = self.requests
+        return resp
 
     def __iter__(self):
         return (t for t in [self])
--- a/sunpy/net/tests/test_fido.py
+++ b/sunpy/net/tests/test_fido.py
@@ -314,3 +314,15 @@ def test_repr(query):
         else:
             assert "Provider" not in rep_meth()
             assert "Providers" in rep_meth()
+
+
+@pytest.mark.remote_data
+def test_slicing_JSOC():
+    tstart = '2011/06/07 06:32:45'
+    tend = '2011/06/07 06:33:15'
+    res = Fido.search(a.Time(tstart, tend), a.jsoc.Series('hmi.M_45s'),
+                      a.jsoc.Notify('jsoc@cadair.com'))
+    result = Fido.fetch(res[0, 0])
+    assert isinstance(result, list)
+    assert result[0].endswith('hmi.m_45s.20110607_063300_TAI.2.magnetogram.fits')
+    assert result[1].endswith('hmi.m_45s.20110607_063345_TAI.2.magnetogram.fits')


sunpy/net/fido_factory.py
```diff
@@ -99,6 +99,7 @@ def _handle_record_slice(self, client_resp, record_slice):
         ret = resp_type(resp)
         # Make sure we pass the client back out again.
         ret.client = client_resp.client
+        ret.query_args = client_resp.query_args
```
Cadair: As the `query_args` attribute is specific to `JSOCResponse` this is the wrong place for this. It should be added in this https://github.com/sunpy/sunpy/blob/master/sunpy/net/jsoc/jsoc.py#L66 `__getitem__`

eamanu: Hi thanks for the review. 

I can add on __getitem__ to get the table and the query_args too. 

But now I have the problem when 

```
# Reconstruct a response object with the sub-indexed records.
ret = resp_type(resp)
```

This do, for some reason, the ret variable have `query_args = None` after `resp_type(resp)`. I don't know how to make the "cast" without erase query_args. 

Any idea?

nabobalis:
I think this needs a unit test to make sure we never break it again.

eamanu:
Ok, I can add it.

@Cadair  the change making it's ok for you?

Now, I will work on unittest.

sunpy/net/jsoc/jsoc.py
```diff
@@ -43,10 +43,20 @@ def __init__(self, table=None):
         """
         table : `astropy.table.Table`
         """
-
-        self.table = table
-        self.query_args = None
-        self.requests = None
+        if isinstance(table, list):
```
Cadair: I am not sure why you are checking for `list` here, I think what you want to be doing is checking for `JSOCResponse`. Like this:

```suggestion
        if isinstance(table, JSOCResponse):
            self.table = new_table.table
            self.query_args = new_table.query_args
            self.requests = new_table.requests
        else:
            self.table = table
            self.query_args = None
            self.requests = None
```

(there should only be one `else` block.

sunpy/net/jsoc/jsoc.py
```diff
@@ -64,7 +74,11 @@ def __len__(self):
             return len(self.table)
 
     def __getitem__(self, item):
-        return type(self)(self.table[item])
+        type_self = type(self)
```
Cadair: I think I would rather write this as:

```suggestion
        resp = type(self)(self.table[item])
```

sunpy/net/jsoc/jsoc.py
eamanu: Hi ! sorry for the delay. I try to check for a JSOCResponse but in fido_factory.py in the lines:
```
if isinstance(record_slice, int):
    resp = [client_resp[record_slice]]
else:
    resp = client_resp[record_slice]
# Reconstruct a response object with the sub-indexed records.
ret = resp_type(resp)
```
create a list, and then in if I use `isinstance(table, JSOCResponse)` this will be ignore.

sunpy/net/jsoc/jsoc.py
eamanu: Great! thanks!

eamanu:
@Cadair @nabobalis The tests fail because they are waiting for a valid email. What can we do with that?

nabobalis:
We have one for the other tests, we just need to borrow that.

nabobalis:
I was wrong, we don't actually have a email address. But I found a bug/issue in the mean time.

eamanu:
How could we continue with this PR? the test fail because they are waiting for an email.

nabobalis:
Adding `attrs.Notify('jsoc@cadair.com'))` should fix that part.

Going off my memory, I noticed that the download seemed to download the entire response instead of the sliced one. 

Also there seems to a few properties that aren't updated when we slice the response, I think it was `num_files` or something.  But I would have to check again since it has been a while.

nabobalis:
So I added the notify attr to fix the test and rebased it upon of current master to fix the tests that are failing.

sunpy/net/jsoc/jsoc.py
```diff
@@ -40,10 +40,24 @@ def __init__(self, table=None):
...
+                self.query_args = new_table.query_args
+                self.requests = new_table.requests
+            else:
+                self.table = table
+                self.query_args = None
+                self.requests = None
+        else:
+            self.table = table
+            self.query_args = None
+            self.requests = None
```
dpshelio: is there a different way to repeat all these lines multiple times?
What about define them at first, as such, and then only change them when needed?

changelog/2796.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix the slicing a JSOC UnifiedResponse object. This bug was that when slice a Unifiedresponse this not saved the query_args attribute.
```
Cadair: This could do with rewording to be a little clearer imo

PR #2188, opened 2017-07-11
Rewritten JSOC Client backed by drms

nitinkgp23:

PR #2807, opened 2018-10-25
A dependency cleanup

Cadair:
@dstansby @nabobalis and I have been having some conversations about the number of dependencies SunPy has and how as we get more packages relying on SunPy not all users might want all dependencies all the time.

I noticed that since #2188 was merged the only point we were using `requests` was for one very simple check in `vso.py` so I replaced that with a standard library version, which means we can drop requests completely as a sunpy dep.

Finally, I have taken the executive decision that we are going to start putting imports in functions! We shouldn't do this when we don't have to, but things that are optional dependancies should probably be imported inside functions if they are in files which have a lot of functionality which do not use the optional dep and only one that does. I have also done the same thing with `sunpy.map` in a few places, because `sunpy.map` takes a *long* time to import, so it makes sense to delay pulling in all that until it's actually needed.

I am very open to discussion and debate on any of this, and in fact am fully expecting it :stuck_out_tongue: 

**Update: I decided to move the Rhessi stuff into it's own PR. #2808**

### DIFF
--- a/sunpy/time/time.py
+++ b/sunpy/time/time.py
@@ -138,22 +138,27 @@ def convert_time(time_string, **kwargs):
     raise ValueError("'{tstr!s}' is not a valid time string!".format(tstr=time_string))
 
 
-@convert_time.register(pandas.Timestamp)
-def convert_time_pandasTimestamp(time_string, **kwargs):
-    return time_string.to_pydatetime()
+# Only register pandas if we can import pandas
+try:
+    import pandas
 
+    @convert_time.register(pandas.Timestamp)
+    def convert_time_pandasTimestamp(time_string, **kwargs):
+        return time_string.to_pydatetime()
 
-@convert_time.register(pandas.Series)
-def convert_time_pandasSeries(time_string, **kwargs):
-    if 'datetime64' in str(time_string.dtype):
-        return np.array([dt.to_pydatetime() for dt in time_string])
-    else:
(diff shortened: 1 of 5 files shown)


sunpy/map/mapbase.py
```diff
@@ -18,7 +18,6 @@
 
 import sunpy.io as io
 import sunpy.coordinates
-import sunpy.cm
```
Cadair: This shouldn't be removed, it's registering them with mpl.

sunpy/time/time.py
```diff
@@ -140,22 +139,27 @@ def convert_time(time_string, **kwargs):
     raise ValueError("'{tstr!s}' is not a valid time string!".format(tstr=time_string))
 
 
-@convert_time.register(pandas.Timestamp)
-def convert_time_pandasTimestamp(time_string, **kwargs):
-    return time_string.to_pydatetime()
+# Only register pandas if we can import pandas
```
Cadair: This probably needs doing on the time branch rather than here.

ayshih:
> Finally, I have taken the executive decision that we are going to start putting imports in functions!

Yay!  Down with PEP8! =P

We already have to allow such imports in our codebase to prevent circular imports (e.g., in `coordinates`), and allowing them to avoid unnecessary dependencies and performance hits makes sense to me.

PR #2714, opened 2018-08-06
[0.9] RHESSI get_obssum_filename will now download datafile for final day in input timerange

aringlis:
See #2672  - previously, this function did not return the datafile from the final day of the request. This showed up in FIDO queries for RHESSI data as well, which rely on this function.

PR #2808, opened 2018-10-25
The Great RHESSI Cleanup of 1.0

### LINKED ISSUE #215: Write unit tests for sunpy.instr.rhessi
Currently there is 0% coverage for instr/rhessi.py

### LINKED ISSUE #2231: Improve RHESSI instrument module
Just a tidy up really:

- some repeated code e.g. `parse_obssumm_file` and `parse_obssumm_hdulist`
- refactor methods so more tests can be run offline
- refactor tests to use reprsentative data so can be run offline  

Some of the issue raised above have been addressed in PR #2048

### LINKED ISSUE #2382: Rhessi Fido queries can not span multiple months
The code in `instr.rhessi` only returns the database files for the month of the start time. It needs to return a list for every month in the time range.

### LINKED ISSUE #2672: RHESSI FIDO Queries Give Warnings and Exclude The Last Day from Results
When querying FIDO for RHESSI data spanning multiple days the following warning is showed:

> "UserWarning: Currently only support providing data from one whole day."
 
But the function then returns the correct days. minus the last day of the sequence.

For example, the following gives a user warning but works, with the exception of giving the last days result:
```
import sunpy.timeseries as ts
results = Fido.search(a.Time('2017/09/06 11:00', '2017/09/09 12:00'), a.Instrument('rhessi'))
print('Results:\n',results)
lis_downloaded_files = Fido.fetch(results)
print('downloaded:\n',lis_downloaded_files)
ts_rhessi = ts.TimeSeries(lis_downloaded_files, source='rhessi', concatenate=True)
print('Result from',ts_rhessi.data.index[0],'to',ts_rhessi.data.index[-1])
```

This is independent of [Issue #2382](https://github.com/sunpy/sunpy/issues/2382)

Cadair:
I noticed that importing `sunpy.timeseries` was importing `sunpy.map` (which is stupid) and I traced that to `sunpy.instr.rhessi`, which was also being imported by `sunpy.net` which meant that `sunpy.net` was also importing `sunpy.map`!! This lead me to completely refactor `sunpy.instr.rhessi` so that all the data download code was in `sunpy.net.dataretriever.sources.rhessi` (i.e. only exposed through Fido) and then I removed all the code from `sunpy.instr.rhessi`.

fixes #2382, fixes #215, fixes #2672 and fixes #2231 

[This also removes a few functions in `sunpy.instr.rhessi` or moves them into `net`. I think we should probably deprecate everything as a removal in our next 0.9 release]

Think the best option to remove the python-dateutil module at this point is to put `dateutil.rrule` into `extern/`. We should decide if that is something we want to do, or if we are happy with the dep.


sunpy/instr/rhessi.py
```diff
@@ -5,106 +5,32 @@
...
 from sunpy.coordinates import get_sunearth_distance
-import sunpy.map
-import sunpy.io
 
-from sunpy.extern.six.moves import urllib
-from sunpy.extern.six.moves.urllib.request import urlopen, urlretrieve
-from sunpy.extern.six.moves.urllib.error import URLError
-
-
-__all__ = ['get_obssumm_dbase_file', 'parse_obssumm_dbase_file',
```
Cadair: `get_obsumm_dbase_file`, `get_obssum_filename` and `get_obssumm_file` should all be marked as deprecated in 0.9 and say "use Fido".

nabobalis: I can make a note and do a PR for that unless you really want to.

Cadair: @ehsteve Does this seem reasonable to you, or are there reasons to keep these as independent functions?

samaloney:
Would agree with moving `parse_observing_summary_hdulist` to `RHESSITimeseries`. Not sure about moving the summary base parser to net, while it's not used in another way, does it really belong in net?

Cadair:
This now also includes #2714

sunpy/instr/rhessi.py
ehsteve: I think we should keep that functionality here for two reasons with Fido calling it.
1] It is easier for someone to find the code for a particular instrument rather than it being buried in Fido.
2] I think there is more information in the dbase files that could be used for other reasons besides Fido.

ehsteve:
As I noted above my opinion is that we keep all instrument specific code in `instr` and other parts of SunPy make use of it. I think it would be easier from a developer perspective to ask someone from an instrument to fix a particular instrument specific problem and not have them have to learn how Fido works (or other part of SunPy).

Cadair:
A general point on this:

> As I noted above my opinion is that we keep all instrument specific code in `instr` and other parts of SunPy make use of it.

This is not the case at the moment, and probably will never completely be the case. I.e all the map sources, timeseries sources and net sources are all "instrument specific" (especially timeseries which has a bunch of file parsing routines in it).

I feel like the `instr` subpackage is really poorly defined, it has a bunch of disparate code in it that feels like it doesn't really belong. The two bits of instr I can see are the fermi rhessi and lyra code for parsing event lists, and the goes emission measure stuff. Then a few extras like aia prep and a weird iris helper.

I half feel that we should have a better idea of what we want to allow in instr and what should go in other places and what should go in affiliated packages.

---

A little more specific to this PR:

I am happy to have the file parsing code live in `instr` because I can see that a) sometimes we might need to parse a file not into a map or a timeseries and b) sometimes you might want to parse a file for other purposes other than creating a map or timeseries.

However, I strongly feel that file download should be handled exclusively by `net` (having just looked through instr, I now realise this is a bigger task than just rhessi). Partly this is a dependency issue, in my mind it is the net submodule that should be the one that has dependencies related to network access and file download, not `instr` or anywhere else. Secondly, I think it just makes conceptual sense!!

Therefore, on this specific PR, I am happy to move `parse_obssumm_dbase` back into `instr`. However, functions starting with `get_` I really do think should live in the submodule designed for the downloading of data. The way I have implemented `get_observing_summary_dbase_file` is as a static method, so it is trivial to still use it in isolation if that is what people want to do. (it's just `RHESSIClient.get_observing_summary_dbase_file("2011/01/01")` for instance).

ehsteve:
@Cadair all good points and I'm fine with proceeding with this PR as it stands but the other consideration is how we might support future affiliated packages. Say there is a rhessi package. it would make more sense for the RHESSI team to proivde all of the baseline download codes (and maintain them) and we just use them in SunPy. Having things in `instr` is a way to model that potential future.

Cadair:
With regards to net, affiliated packages can happily register clients with Fido. This is something I am planning on doing with the DKIST user tools early next year, so I imagine it will motivate me to make sure it is well supported. Also I believe it is something heliopy are considering doing as well.

Cadair:
@ehsteve if you are ok with this as it is can you review and approve? :angel:

Cadair:
Ok, I think this is ready for a final review...

sunpy/net/dataretriever/client.py
```diff
@@ -300,6 +300,15 @@ def _get_full_filenames(self, qres, filenames, path):
 
         return paths
 
+    def _get_time_for_url(self, urls):
+        """
+        This method allows clients to customise the timerange displayed for
+        each URL.
+
+        It should return a sunpy.time.TimeRange object per URL.
```
dpshelio: a TimeRange? what if you don't know from the URL what's the obs-end?

sunpy/net/dataretriever/sources/rhessi.py
```diff
@@ -1,16 +1,133 @@
...
+            return server
+        except (URLError, socket.timeout):
+            pass
+
+    raise IOError('Unable to find an online HESSI server from {0}'.format(data_servers))
+
 
 class RHESSIClient(GenericClient):
 
+    def get_observing_summary_filename(self, time_range):
```
dpshelio: I thought we had this done using the scraper..

sunpy/net/dataretriever/tests/test_rhessi.py
```diff
@@ -1,68 +1,185 @@
...
 from sunpy.net import attrs as a
-from sunpy.time import parse_time
-from sunpy.time.timerange import TimeRange
-from sunpy.net.vso.attrs import Time, Instrument
-from sunpy.net.dataretriever.client import QueryResponse
+from sunpy.time import TimeRange, parse_time
 from sunpy.net.fido_factory import UnifiedResponse
+from sunpy.net.dataretriever.client import QueryResponse
 
 LCClient = rhessi.RHESSIClient()
```
dpshelio: Since we don't hae LC anymore... shouldn't we call this differently?

sunpy/net/dataretriever/client.py
Cadair: :rofl: you wrote this function, I just added the docstring!

If you don't know the end time, I guess you make it up! It makes no sense to return anything other than a range here, all files will have some range, I would really hope you know what it is.

sunpy/net/dataretriever/sources/rhessi.py
Cadair: We had? I don't remember that. You got a link?

sunpy/net/dataretriever/client.py
dpshelio: Really?? Then of course we know it!!
I'll have to look into it with more detail... but not know. Let's merge this and one day I'll get my head around the scrapper again.

sunpy/net/dataretriever/sources/rhessi.py
dpshelio: I think it's on one of the automated closed PRs... I got a list to work on these... so one day (hopefully before 1.0).

dpshelio:
More conflicts!! I won't be able to press any merge button today?? :crying_cat_face:

nabobalis:
Merging in those six and future changes probably has messed up every open PR.

nabobalis:
Since its failing, I probably fucked up the merging.

nabobalis:
ITS ALL ALIVE (BAR THE COVERAGE)

PR #2838, opened 2018-11-08
Astropy dev fixes

Cadair:
This should hopefully fix the astropy dev build.


Cadair:
@nabobalis we really need to release this as a 0.9 patch before astropy 3.1 comes out. (Though it is only tests...)

nabobalis:
merge it can we can cook up a release.

PR #2839, opened 2018-11-09
Tox our CI

nabobalis:
We now use Tox for run all our CI.

It also allow us to pre-define and generate the same conda environment as the CI locally so we can debug issues this way now. 

```bash
$ tox -al
py37-offline
py36-offline
py37-online
py37-astropydev
py37-numpydev
py37-build_docs
py36-online
py36-astropydev
py36-numpydev
py36-build_docs
figure
```

The old ways are still supported. **FOR NOW**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,25 +8,18 @@ apt-run: &apt-install
     apt update
     apt install -y graphviz build-essential
 
-docs-install: &docs-install
-  name: Install Python dependencies
+tox-install: &tox-install
+  name: Install Tox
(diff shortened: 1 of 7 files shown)


nabobalis:
IT ALL PASSED

Cadair:
@nabobalis This looks ok to me, but giles and codecov haven't reported on it?

Cadair:
Ah I think that's because it has built from your fork. I am going to merge and we can see what the damage is.

drdavella:
🎆

PR #2842, opened 2018-11-13
Added docs_require and fixed RTD build.

nabobalis:
It seems I somehow removed these changes from one of the PRs I already merged.

Fun fact that adding `git+https://github.com/hawkowl/towncrier.git` to setup.cfg and having it parsed by whatever is used by setup.py breaks.


readthedocs.yml
```diff
@@ -0,0 +1,8 @@
+build:
+  image: latest
+
+python:
+  version: 3.6
+  pip_install: true
+  extra_requirements:
+        - all
```
nabobalis: I swear it worked without this file. But it seems that I need this at least.

Cadair:
yeah setup.py can't resolve git links. Can we add the dev version of towncrier to the rtd yaml file?

nabobalis:
Looks like thats a no.

Cadair:
Balls. That's kinda a problem, we have needed git versions of the gallery and such before. Maybe we need to go back to requirements for the docs?

nabobalis:
I will see what I can do.

nabobalis:
This seems to work locally, it updates the pip version of town crier and the changelog runs on locally and shows the fancy fragments but for some reason it does not update the changelog on my fork on RTD.

PR #2843, opened 2018-11-13
Added docs_require and fixed RTD build

nabobalis:
It seems I somehow removed these changes from one of the PRs I already merged.

So I forgot that you can wipe the env on RTD and doing that made it work. 
Since we are also broken right now with towncrier the package, I have removed it from the setup.cfg


rtd.txt
```diff
@@ -0,0 +1,3 @@
+# This file is for special versions of packages we need
```
Cadair: why don't we put this file in `docs/`?

tox.ini
```diff
@@ -17,8 +17,7 @@ conda_deps =
     astropydev,numpydev: cython
     offline,online,build_docs: numpy
     offline,online,build_docs: astropy
-    # bug with doc folder path
```
Cadair: Oh this is fixed under tox right? Not under setup.py?

Issue #1832, opened 2016-06-14
List of accepted metadata for custom SunPy Maps


aringlis:
When creating custom SunPy maps (i.e. from a 2D array), there is no obvious way to see what the accepted metadata keywords are. I have been rooting around in mapbase.py to try and find the expected input. 

For user-friendliness, it would be nice to have a full list of the metadata keywords that can be input.

Cadair:
I agree, it will change dependant upon the instrument, but for the moment we should at least document the defaults in `GenericMap`

dpshelio:
we could have it interactively in the way `iraf` used to work :)

Cadair:
I have no idea what that is.

Cadair:
what is that?

dpshelio:
@Cadair you don't need to look back to `iraf`, but in short, it opened a text-like editor with the fields to write them in, and then generated the object based on these things. It was kind of ncurses but based on vi.

Alex-Ian-Hamilton:
From my investigations, with metadata in a dictionary called meta, I needed to add the following values.

For WCS, HMI has typical values:
    meta['cdelt1'] = 0.504219
    meta['cdelt2'] = 0.504219
    meta['cunit1'] = 'arcsec'
    meta['cunit2'] = 'arcsec'
    meta['dsun_obs'] = 147141258270.0 # Else Warning: Missing metadata for Sun-spacecraft separation
    meta['crpix1'] = 2045.131592
    meta['crpix2'] = 2042.132812
    meta['crval1'] = 0.0
    meta['crval2'] = 0.0

    For non-earth observations you need:
    meta['HGLN_OBS'] = 70.438797
    meta['HGLT_OBS'] = -4.521303

    Which HMI actually stores as:
    meta['crln_obs'] = 70.438797
    meta['crlt_obs'] = -4.521303

For a observation time:
dic_meta['DATE_OBS'] = '2011-02-15T00:58:57.30'

To be recognised as an SDO HMI map you need:    
    dic_meta['telescop'] = 'SDO/HMI'
    dic_meta['instrume'] = 'HMI' # generally 'HMI_FRONT2', but only checks using .startswith.
    dic_meta['content'] = 'MAGNETOGRAM'

If all the above are give, plots get generated without warnings and you can make a map sequence plot work (that's temperamental otherwise).
I can dig around the source to look for alternatives if people thing there are some.

wtbarnes:
Doesn't #3083 solve this?

[...]

PR #2847, opened 2018-11-19
Don't assume arcseconds when CUNIT metadata isn't present

dstansby:
Blindly assuming that data is in units of arcseconds seems like a Bad Idea. A concrete example is LASCO C3 data that doesn't have any metadata (see https://github.com/sunpy/sunpy/pull/2775#issuecomment-428933507 and comments below). Currently LASCO C3 data with missing metadata is assumed to be in arcseconds, which is completely wrong and results in e.g. the draw-solar-limb function putting the solar limb in completely the wrong place without even a warning.

At the very least I think assuming arcseconds when metadata is missing should be done on an instrument-by-instrument level, and not in `mapbase`. This PR removes the arcsecod assumption in mapbase.

### DIFF
--- a/changelog/2847.breaking.rst
+++ b/changelog/2847.breaking.rst
@@ -1,0 +1,2 @@
+Maps no longer assume that the pixel units are arcseconds if the units aren't
+explicitly set.
--- a/changelog/2847.breaking.rst
+++ b/changelog/2847.breaking.rst
@@ -1,2 +1,3 @@
 Maps no longer assume that the pixel units are arcseconds if the units aren't
-explicitly set.
+explicitly set. In addition to this if critical metadata is missing from when
+creating a map, the map will fail to initialize and will raise an error.
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -40,6 +40,10 @@ SpatialPair = namedtuple('SpatialPair', 'axis1 axis2')
 __all__ = ['GenericMap']
 
 
+class MapMetaValidationError(AttributeError):
+    pass
+
+
 class GenericMap(NDData):
     """
     A Generic spatially-aware 2D data array
@@ -681,8 +685,8 @@ class GenericMap(NDData):
         """
         Image coordinate units along the x and y axes (i.e. cunit1, cunit2).
         """
-        return SpatialPair(u.Unit(self.meta.get('cunit1', 'arcsec')),
-                           u.Unit(self.meta.get('cunit2', 'arcsec')))
+        return SpatialPair(u.Unit(self.meta.get('cunit1')),
+                           u.Unit(self.meta.get('cunit2')))
 
     @property
     def rotation_matrix(self):
@@ -775,8 +779,19 @@ class GenericMap(NDData):
             CUNIT1, CUNIT2, WAVEUNIT
 
         """
+        msg = ('Image coordinate units for axis {} not present in metadata.')
+        err_message = []
+        for i in [1, 2]:
+            if self.meta.get(f'cunit{i}') is None:
+                err_message.append(msg.format(i, i))
 
-        for meta_property in ('cunit1', 'cunit2', 'waveunit'):
+        if err_message:
+            err_message.append(
+                'See http://docs.sunpy.org/en/stable/code_ref/map.html#fixing-map-metadata` for '
+                'instructions on how to add missing metadata.')
+            raise MapMetaValidationError('\n'.join(err_message))
+
+        for meta_property in ('waveunit', ):
             if (self.meta.get(meta_property) and
                 u.Unit(self.meta.get(meta_property),
                        parse_strict='silent').physical_type == 'unknown'):
--- a/sunpy/map/sources/iris.py
+++ b/sunpy/map/sources/iris.py
@@ -37,6 +37,9 @@ class SJIMap(GenericMap):
     """
 
     def __init__(self, data, header, **kwargs):
+        # Assume pixel units are arcesc if not given
+        header['cunit1'] = header.get('cunit1', 'arcsec')
+        header['cunit2'] = header.get('cunit2', 'arcsec')
         GenericMap.__init__(self, data, header, **kwargs)
 
         self.meta['detector'] = "SJI"
--- a/sunpy/map/sources/soho.py
+++ b/sunpy/map/sources/soho.py
@@ -60,11 +60,15 @@ class EITMap(GenericMap):
     """
 
     def __init__(self, data, header, **kwargs):
-        GenericMap.__init__(self, data, header, **kwargs)
+        # Assume pixel units are arcesc if not given
+        header['cunit1'] = header.get('cunit1', 'arcsec')
+        header['cunit2'] = header.get('cunit2', 'arcsec')
 
+        GenericMap.__init__(self, data, header, **kwargs)
         # Fill in some missing info
         self.meta['detector'] = "EIT"
         self.meta['waveunit'] = "Angstrom"
+
         self._fix_dsun()
         self._nickname = self.detector
         self.plot_settings['cmap'] = plt.get_cmap(self._get_cmap_name())
@@ -172,7 +176,9 @@ class MDIMap(GenericMap):
     """
 
     def __init__(self, data, header, **kwargs):
-
+        # Assume pixel units are arcesc if not given
+        header['cunit1'] = header.get('cunit1', 'arcsec')
+        header['cunit2'] = header.get('cunit2', 'arcsec')
         GenericMap.__init__(self, data, header, **kwargs)
 
         # Fill in some missing or broken info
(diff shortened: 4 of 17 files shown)


sunpy/map/mapbase.py
```diff
@@ -672,8 +672,12 @@ def spatial_units(self):
         """
         Image coordinate units along the x and y axes (i.e. cunit1, cunit2).
         """
-        return SpatialPair(u.Unit(self.meta.get('cunit1', 'arcsec')),
-                           u.Unit(self.meta.get('cunit2', 'arcsec')))
+        cunit1 = self.meta.get('cunit1')
+        cunit2 = self.meta.get('cunit2')
+        if (cunit1 is None) or (cunit2 is None):
+            raise AttributeError('Image coordinate units not present in metadata')
```
dpshelio: Should we have an argument to force the use of a particular unit? Or  a helpful message on how to create a map from this type of data?

dstansby: I could add a little bit more to this message that says how to set the unit metadata of the map.

dstansby: How about now?

dstansby:
This is kind of all working, but I'm not really sure what to do about the EIT sample map that's missing the metadata?

sunpy/map/mapbase.py
```diff
@@ -678,8 +678,17 @@ def spatial_units(self):
...
-        return SpatialPair(u.Unit(self.meta.get('cunit1', 'arcsec')),
-                           u.Unit(self.meta.get('cunit2', 'arcsec')))
+        cunit1 = self.meta.get('cunit1')
+        cunit2 = self.meta.get('cunit2')
+        msg = ('Image coordinate units for axis {} not present in metadata. '
+               'It can be set using `map.meta["cunit{}"] = unit`.')
+        if cunit1 is None:
+            raise AttributeError(msg.format('1', '1'))
+        if cunit2 is None:
+            raise AttributeError(msg.format('2', '2'))
```
dpshelio:
```python
cunits = {f'cunit{i}: self.meta.get(f'cunit{i}') for i in [1, 2]}
msg = ('Image coordinate units for axis {} not present in metadata. '
       'It can be set using `map.meta["{}"] = unit`.')
for lab in cunits:
    if cunits[cunit_i] is None:
         raise AttributeError(msg.format(cunit_i[-1], cunit_i))
```

Also, the problem I see here is that you need to run it once, fix `cunit1`, then encounter other error, run it again...  What about:

```python
err_message = []
for lab in cunits:
    if cunits[cunit_i] is None:
         err_message.append(msg.format(cunit_i[-1], cunit_i))

if err_message:
    raise AttributeError('\n'.join(err_message))

```

sunpy/map/sources/soho.py
```diff
@@ -109,8 +109,13 @@ def __init__(self, data, header, **kwargs):
...
 
-        self.meta['CUNIT1'] = self.meta['CUNIT1'].lower()
-        self.meta['CUNIT2'] = self.meta['CUNIT2'].lower()
+        cunit1 = self.meta['CUNIT1']
+        if cunit1 is not None:
+            self.meta['CUNIT1'] = cunit1.lower()
+
+        cunit2 = self.meta['CUNIT2']
+        if cunit2 is not None:
+            self.meta['CUNIT2'] = cunit2.lower()
```
dpshelio:
```python
for cunit in ['CUNIT1', 'CUNIT2']:
    if self.meta[cunit] is not None:
        self.meta[cunit] = self.meta[cunit].lower()
```

dstansby:
wooops, force pushed an old version, so will have to dig the correct new version out from my laptop.

dstansby:
🎉 it works!

sunpy/map/mapbase.py
```diff
@@ -679,8 +679,21 @@ def spatial_units(self):
         """
         Image coordinate units along the x and y axes (i.e. cunit1, cunit2).
         """
-        return SpatialPair(u.Unit(self.meta.get('cunit1', 'arcsec')),
-                           u.Unit(self.meta.get('cunit2', 'arcsec')))
+        cunit1 = self.meta.get('cunit1')
+        cunit2 = self.meta.get('cunit2')
+
+        msg = ('Image coordinate units for axis {} not present in metadata. '
```
Cadair: I just remembered that we have `GenericMap._validate_meta` as a place for doing these checks. Which causes the errors to be raised on instantiation, which is probably preferable than at a random point later.

dstansby: I think doing this will completely prevent creation of the `Map` object, and therefore not give the user a way to manually specify the units before doing other operations on the map.

dpshelio: What do we want? allow the creation of a `Map` raising this error when it's used by:
- `plot`, 
- `superpixel`
- `draw_rectangle`
- `shift`
- `_reference_longitdue`, `_reference_latitude`
- `get_lon_lat`
   - `world_to_pixel`, 
   - `resample`
   - `rotate` 
   - `submap`
- `wcs`
  - `coordinate_frame`
  - `as_mpl_axes` 

Or tell the user at creation: Hey!! your map doesn't has enough information to allow you to play with it correctly.
I just imagine that could be really annoying to get this type of message after I've done some analysis (don't know which) to then been told that my map is not "good". What's worse... I could have been saving these maps which in the future I want to plot, to discover that I have to reprocess all of them (imagine 1000s).

From my point of view, I think we should [fail fast](https://en.wikipedia.org/wiki/Fail-fast).

dstansby: The use case I was thinking of was using `sunpy.map.Map` to load a `.fits` file, and extract the data.  For this the coordinate system doesn't have to be defined, but `sunpy.map.Map` needs to be able to make a map that doesn't have enough metadata to do any of the plotting/coordinate stuff. Unless there's another way to load data from a `.fits` file without going through `Map`?

dstansby: Looks like `sunpy.io` stuff can be used, so I'd be 👍 on failing fast.

Cadair: I think we should couple this validation with the discussion starting in #1832 about formalising the minimum metadata needed to build a map. Then we should document all of this as well.

@dstansby will you be able to shuffle this around into the validation method and then we can merge this and continue from there?

sunpy/map/mapbase.py
```diff
@@ -773,8 +773,17 @@ def _validate_meta(self):
...
+               'It can be set using `map.meta["cunit{}"] = unit`.')
+        err_message = []
+        for i in [1, 2]:
+            if self.meta.get(f'cunit{i}') is None:
+                err_message.append(msg.format(i, i))
 
-        for meta_property in ('cunit1', 'cunit2', 'waveunit'):
+        if err_message:
+            raise AttributeError('\n'.join(err_message))
```
Cadair: As [discussed](https://matrix.to/#/!MeRdFpEonLoCwhoHeT:matrix.org/$15519781816270xRRRX:cadair.com?via=cadair.com&via=matrix.org&via=openastronomy.org) if you make this raise a specific error to map header validation then it will be simple to patch `MapFactory` to skip invalid HDUs.

Cadair: This is then the patch to `map_factory.py` that I think will do the job: https://github.com/Cadair/sunpy/commit/020944b0c56b40b0f63477e7fb1553c12f91d268

sunpy/map/sources/iris.py
```diff
@@ -37,6 +37,9 @@ class SJIMap(GenericMap):
     """
 
     def __init__(self, data, header, **kwargs):
+        # Assume pixel units are arcesc if not given
```
Cadair: is this still needed with the change to `MapFactory`?

Cadair:
@dstansby one of the online test fails is real.

sunpy/map/mapbase.py
```diff
@@ -773,8 +777,17 @@ def _validate_meta(self):
             CUNIT1, CUNIT2, WAVEUNIT
 
         """
+        msg = ('Image coordinate units for axis {} not present in metadata. '
```
dstansby: Note to self: this message needs changing to tell users how to load data if the metadata is missing.

dstansby:
So it seems like we need to provide users for instructions on how to fix metadata before creating a `Map` from a `.fits` file with missing `cunit` metadata. Does anyone know a simple way to do this?

changelog/2847.breaking.rst
```diff
@@ -0,0 +1,2 @@
+Maps no longer assume that the pixel units are arcseconds if the units aren't
+explicitly set.
```
Cadair: Can you add a comment here that they will fail on instantiation now as well?

sunpy/map/sources/iris.py
dstansby: Not sure; I'll double check all these statements I've added.

PR #2853, opened 2018-11-22
Always use _default_wrap_angle rather than hard coding wrap angle

Cadair:


nabobalis:
Test issues are online hec ones.

PR #2856, opened 2018-11-26
Make sure we are only slicing with ints and not numpy.int64s

Cadair:
I discovered this while working on some dkist stuff, it's a pretty subtle distinction but it's just better practice to cast back to `int`.


nabobalis:
All of `hec` is still failing. Only seems to be that.

PR #2859, opened 2018-11-28
Fix doc building when ruamel package not installed

dstansby:
`html_context` needs to be defined so it can be accessed later.


nabobalis:
Woops, I clearly missed that when I added it to the `conf.py` file. Cheers!

PR #2863, opened 2018-12-06
Update CI configs to install tox-conda from pypi

drdavella:
Now that I've released [tox-conda on pypi](https://pypi.org/project/tox-conda/), I thought this might be a good opportunity to take it for a spin.


Cadair:
What is with that windows build on azure, and also Helio...

PR #2865, opened 2018-12-10
Added new example for getting colormaps

ehsteve:
### Description
Provide a simple example to get at the SunPy colormaps and using them with matplotlib


nabobalis:
So I rebased and decided that we also do not need the reference colormap example. I added it to the new example.

Since the CI is on fire, the only build we care about is the docs one.

Cadair:
Can we set the second image (the one with all the colourmaps) as the thumbnail? i.e. https://sphinx-gallery.readthedocs.io/en/latest/auto_examples/plot_choose_thumbnail.html

nabobalis:
And like magic.

examples/plotting/sunpy_matplotlib_colormap.py
```diff
@@ -6,6 +6,7 @@
 This examples shows how you can use the SunPy colormaps with Matplotlib.
 Also the full range of colormaps we provide.
 """
+sphinx_gallery_thumbnail_number = 2
```
Cadair: this shouldn't be here.

examples/plotting/sunpy_matplotlib_colormap.py
```diff
@@ -43,6 +44,7 @@
 print(cm.cmlist.keys())
 
 ###############################################################################
+# Ignore this: sphinx_gallery_thumbnail_number = 2
```
Cadair: or this.

examples/plotting/sunpy_matplotlib_colormap.py
```diff
@@ -43,6 +44,7 @@
...
 # We also provide a function that will display all our colormaps:
```
Cadair: Add 
```
# sphinx_gallery_thumbnail_number = 2
```
here.

examples/plotting/sunpy_matplotlib_colormap.py
nabobalis: GOD DAMN IT

examples/plotting/sunpy_matplotlib_colormap.py
Cadair: as done here: https://sphinx-gallery.readthedocs.io/en/latest/_downloads/plot_choose_thumbnail.py

examples/plotting/sunpy_matplotlib_colormap.py
nabobalis: I WILL FIX THIS

nabobalis: I HAVE THE POWER

PR #2880, opened 2018-12-20
Handle saving out FITS from Astropy. 

### LINKED ISSUE #2879: Crash while saving sunpy Map if it has comments
### Description
If a level1 fits file has comments in it, then to make it level 1.5, aiaprep is invoked, then saving the resultant Map will lead to crash throwing Value Error

### Expected behavior
The save method should run normally and save the map in FITS file.

### Actual behavior
The Map save method threw Value Error

### Steps to Reproduce

1. Download hmi.ic_nolimbdark_720s.20151217_000000_TAI.1.continuum.fits HMI 720 s Limb Darkening corrected fits file, for the given date (the fits file has comments).
2. ```fits_hdu = fits.open('hmi.ic_nolimbdark_720s.20151217_000000_TAI.1.continuum.fits')[1]```
3. ```fits_hdu.verify('fix')```
4. ```fits_hdu.header['HGLN_OBS'] = 0.0```
4. ```aia_map = sunpy.map.Map(fits_hdu.data,fits_hdu.header)```
5. ```aia_map = aiaprep(aia_map)```
6. ```aia_map.save('aiaprep/hmi.ic_nolimbdark_720s.20151217_000000_TAI.1.continuum.fits')```

### System Details

  - SunPy Version: 0.9.3
  - Python Version: 3.7.0
  - OS information: macOS Mojave

harshmathur1990:
### Description
If a Sunpy map has comments in it (which we can see by checking mapobject's meta attribute), and if we go ahead and save it, It will crash because the code is mistyped for checking 'comments' instead of 'comment' and it is calling wrong method of astropy i.e. 'add_comments' instead of 'add_comment'.

Fixes #2879

### DIFF
--- a/changelog/2880.bugfix.rst
+++ b/changelog/2880.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed `sunpy.io.fits.write` to handle the keyword ``COMMENT`` correctly.
--- a/sunpy/io/fits.py
+++ b/sunpy/io/fits.py
@@ -120,7 +120,7 @@ def get_header(afile):
         close = True
 
     try:
-        headers= []
+        headers = []
         for hdu in hdulist:
             try:
                 comment = "".join(hdu.header['COMMENT']).strip()
@@ -139,7 +139,7 @@ def get_header(afile):
             keydict = {}
             for card in hdu.header.cards:
                 if card.comment != '':
-                    keydict.update({card.keyword:card.comment})
+                    keydict.update({card.keyword: card.comment})
             header['KEYCOMMENTS'] = keydict
             header['WAVEUNIT'] = extract_waveunit(header)
 
@@ -204,11 +204,11 @@ def header_to_fits(header):
 
     for k, v in header.items():
         if isinstance(v, fits.header._HeaderCommentaryCards):
-            if k == 'comments':
+            if k.upper() == 'COMMENT':
                 comments = str(v).split('\n')
                 for com in comments:
-                    fits_header.add_comments(com)
-            elif k == 'history':
+                    fits_header.add_comment(com)
+            elif k.upper() == 'HISTORY':
                 hists = str(v).split('\n')
                 for hist in hists:
                     fits_header.add_history(hist)
@@ -306,5 +306,5 @@ def extract_waveunit(header):
                 waveunit = m.group(1)
                 break
     if waveunit == '':
-        return None # To fix problems associated with HMI FITS.
+        return None  # To fix problems associated with HMI FITS.
     return waveunit
--- a/sunpy/io/tests/test_fits.py
+++ b/sunpy/io/tests/test_fits.py
@@ -1,11 +1,12 @@
-from astropy.io import fits
+import os
+from collections import OrderedDict
+
+import astropy.io.fits as fits
 
 import sunpy.io.fits
-from sunpy.io.fits import get_header, extract_waveunit
-
 import sunpy.data.test
-import os
-
+from sunpy.util import MetaDict
+from sunpy.io.fits import get_header, extract_waveunit
 from sunpy.data.test.waveunit import MEDN_IMAGE, MQ_IMAGE, NA_IMAGE, SVSM_IMAGE
 
 testpath = sunpy.data.test.rootdir
@@ -107,3 +108,11 @@ def test_simple_write_compressed(tmpdir):
     with fits.open(str(outfile)) as hdul:
         assert len(hdul) == 2
         assert isinstance(hdul[1], fits.CompImageHDU)
+
+def test_write_with_metadict_header_astropy(tmpdir):
+    fits_file = fits.open(AIA_171_IMAGE)
+    data, header = fits_file[0].data, fits_file[0].header
+    meta_header = MetaDict(OrderedDict(header))
+    temp_file = tmpdir / "temp.fits"
+    sunpy.io.fits.write(str(temp_file), data, meta_header)
+    assert temp_file.exists()
--- a/sunpy/instr/aia.py
+++ b/sunpy/instr/aia.py
@@ -43,7 +43,7 @@ def aiaprep(aiamap):
     """
 
     if not isinstance(aiamap, (AIAMap, HMIMap)):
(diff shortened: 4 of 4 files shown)


harshmathur1990:
> Thanks for the PR @harshmathur1990 :)
> 
> Could you add a test to this? Probably best to manually construct a FITS header, and then run it through the read and save routines. I just want to make sure this bug dosen't come back :)

I'll do it soon and update here.

harshmathur1990:
> Thanks for the PR @harshmathur1990 :)
> 
> Could you add a test to this? Probably best to manually construct a FITS header, and then run it through the read and save routines. I just want to make sure this bug dosen't come back :)

Hi,
I am sorry but I am unable to understand the testing framework. As to where should I write the method to test this? Any guidance would be appreciated.

nabobalis:
Probably best to add the test to `https://github.com/sunpy/sunpy/blob/master/sunpy/io/tests/test_fits.py`.

If you look at `test_extra_comment_write`, that is probably similar in style to what test we need for this. We use pytest as our framework. 

Hopefully this helps!

harshmathur1990:
> Probably best to add the test to `https://github.com/sunpy/sunpy/blob/master/sunpy/io/tests/test_fits.py`.
> 
> If you look at `test_extra_comment_write`, that is probably similar in style to what test we need for this. We use pytest as our framework.
> 
> Hopefully this helps!

Thanks for prompt reply. I thought it is a good idea to run the existing tests on my system before I write  my own test. Hence, when I run:

```py.test```

I got the error:

```
usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --doctest-rst
  inifile: /Users/harshmathur/Documents/CourseworkRepo/sunpy/setup.cfg
  rootdir: /Users/harshmathur/Documents/CourseworkRepo/sunpy
```

To rectify this, I opened setup.cfg file and changed the line 40

From
```addopts = -p no:warnings --doctest-rst``` 

To
```addopts = "-p no:warnings --doctest-rst"```

Then I am facing another issue

```E   KeyError: 'FLOAT_CMP'```

```https://github.com/sunpy/sunpy/issues/2597```

This I found relevant, but it is suggested there to disable the doctests, If you can suggest it how to do that.


I am sorry for asking so many questions, I have not used pytest before.

**Comment by nabobalis** [comment]
That is fine, it's what we are here for. 

First, you do not need to change that line and I would suggest making sure it goes back to the way it was. 

What I think is the issue is we have extra dependencies for our tests. If you install `pytest-astropy` I beleive that those errors should go. Also you will need `hypothesis, mock, pytest-mock`.

**Comment by harshmathur1990** [comment]
> That is fine, it's what we are here for.
> 
> First, you do not need to change that line and I would suggest making sure it goes back to the way it was.
> 
> What I think is the issue is we have extra dependencies for our tests. If you install `pytest-astropy` I beleive that those errors should go. Also you will need `hypothesis, mock, pytest-mock`.

Thank you. It worked, I have written tests, please review.

I had to use astropy.io.fits.open method here to open the fits file because I needed the comment to be a list instead of just a key and a value, as in this case is only when the comment is multiline. The suny fits read method, automatically converts multiline comment into a single line and it will never reaches the code which actually fails. Hence ,  two different methods to open fits file. I have tested it, you guys can also review it and give comments.

Thanks
HM

File: changelog/2880.bugfix.rst [hunk 1]
```diff
@@ -0,0 +1 @@
+Fixed Crash while saving Sunpy Maps if they have comments in them
```
nabobalis: I would suggest rewording this to, `Fixed io.fits write if the header has the keyword COMMENT`.

sunpy/io/tests/test_fits.py
```diff
@@ -95,3 +98,26 @@ def test_extra_comment_write(tmpdir):
...
+
+    fits_file = astropy.io.fits.open(str(temp_file))
+    data, header = fits_file[0].data, fits_file[0].header
+
+    aiamap = sunpy.map.Map(data, header)
+
+    out_file = tmpdir / "test.fits"
+    aiamap.save(str(out_file))
+
+    assert out_file.exists()
```
nabobalis: I think we can simplify this test. 

I would make the comment shorter, I don't think we need 6 lines of strings. 
We can remove the map import. The bug is with the `io.fits` and not `Map`.

You want this test to do two things in my mind. 

1. Able to save out `COMMENT` in the header
2. Check that when you read the file again, the header `COMMENT` is the same.

harshmathur1990: The making comment shorter is fine.

But I am not able to recreate the issue by just calling sunpy.io.fits.write method. Somehow the issue does not come up if one directly saves it avoiding maps.

I am still working on it, may be a day or two :)

nabobalis: There is no rush. 

If the issue is with Map.save, then the test should be `https://github.com/sunpy/sunpy/blob/master/sunpy/map/tests/test_mapbase.py`. Sorry about that. 

But `Map.save` calls `io.write_file(filepath, self.data, self.meta, filetype=filetype, **kwargs)`, so the issue should crop up using that directly.

sunpy/io/fits.py
```diff
@@ -177,10 +177,10 @@ def write(fname, data, header, **kwargs):
 
     for k, v in header.items():
         if isinstance(v, fits.header._HeaderCommentaryCards):
-            if k == 'comments':
+            if k == 'comment':
```
Cadair: should be `if k.lower() == 'comment'`

nabobalis:
@harshmathur1990 We were talking about this issue in our weekly online meeting and it turns out, the issue is slightly more complex than we thought. So I went ahead and finished this up (hopefully).

harshmathur1990:
Thanks. I was out of town since then and hence wasn't able to look into it since then. Go ahead with it as you see fit. :)

nabobalis:
Test fail is GOES time related on windows only.

nabobalis:
Thank you @harshmathur1990!

harshmathur1990:
> Thank you @harshmathur1990!

Thanks a lot :)

PR #2881, opened 2018-12-21
Fix some inconsistent online tests

Cadair:
This also adds a much better error message when we can't connect to any HELIO servers.

TODO:

- [x] A test for the helio error

### DIFF
--- a/sunpy/net/dataretriever/sources/rhessi.py
+++ b/sunpy/net/dataretriever/sources/rhessi.py
@@ -62,8 +62,8 @@ class RHESSIClient(GenericClient):
         --------
         >>> from sunpy.net.dataretriever.sources.rhessi import RHESSIClient
         >>> RHESSIClient().get_observing_summary_filename(('2011/04/04', '2011/04/05'))   # doctest: +REMOTE_DATA
-        ['https://hesperia.gsfc.nasa.gov/hessidata/metadata/catalog/hsi_obssumm_20110404_042.fits',
-        'https://hesperia.gsfc.nasa.gov/hessidata/metadata/catalog/hsi_obssumm_20110405_031.fits']
+        ['https://.../hessidata/metadata/catalog/hsi_obssumm_20110404_042.fits',
+        'https://.../hessidata/metadata/catalog/hsi_obssumm_20110405_031.fits']
         """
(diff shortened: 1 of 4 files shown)


sunpy/net/dataretriever/sources/rhessi.py
```diff
@@ -62,8 +62,8 @@ def get_observing_summary_filename(self, time_range):
         --------
         >>> from sunpy.net.dataretriever.sources.rhessi import RHESSIClient
         >>> RHESSIClient().get_observing_summary_filename(('2011/04/04', '2011/04/05'))   # doctest: +REMOTE_DATA
-        ['https://hesperia.gsfc.nasa.gov/hessidata/metadata/catalog/hsi_obssumm_20110404_042.fits',
-        'https://hesperia.gsfc.nasa.gov/hessidata/metadata/catalog/hsi_obssumm_20110405_031.fits']
+        ['https://.../hessidata/metadata/catalog/hsi_obssumm_20110404_042.fits',
+        'https://.../hessidata/metadata/catalog/hsi_obssumm_20110405_031.fits']
```
nabobalis: I thought of that but then I was like, it makes the example opaque.

Cadair: Yeah, it's not great but it's better than skipping it.

sunpy/net/tests/test_helio.py
```diff
@@ -172,7 +171,8 @@ def test_wsdl_retriever_no_content(mock_endpoint_parser):
     """
     No links found? Return None
     """
-    assert wsdl_retriever() is None
+    with pytest.raises(ValueError):
+        assert wsdl_retriever() is None
```
dpshelio: Shouldn't this be a simple call to the function without the `assert`

sunpy/net/tests/test_helio.py
```diff
@@ -172,7 +171,8 @@ def test_wsdl_retriever_no_content(mock_endpoint_parser):
     """
     No links found? Return None
```
dpshelio: This comment should change too?

Cadair:
@dpshelio updated.

Cadair:
alright @dpshelio I think I have actually done it now :)

PR #2882, opened 2018-12-23
Added constants list to gallery

ehsteve:
### Description

Adds an example to the gallery that simply lists all of the constants available similar to the colormap example.

### DIFF
--- a/examples/plotting/constants_reference.py
+++ b/examples/plotting/constants_reference.py
@@ -1,0 +1,20 @@
+# coding: utf-8
+"""
+===============
+SunPy Constants
+===============
+
+The example shows all of the constants provided by sunpy.
+"""
+
+from sunpy.sun import constants as con
+
+##############################################################################
+# All constants are stored in a dictionary. A list of all available keys
+# which describe the constant can be had with the following command.
+con.constants.keys()
+
+##############################################################################
+# The following command will display all constants as well as their values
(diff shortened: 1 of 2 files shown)


nabobalis:
You need to update your master.

examples/gallery/constants_reference.py
```diff
@@ -0,0 +1,20 @@
...
+
+##############################################################################
+# All constants are stored in a dictionary. A list of all available keys
+# which describe the constant can be had with the following command.
+con.constants.keys()
+
+##############################################################################
+# The following command will display all constants as well as their values
+# in an astropy `Table <http://docs.astropy.org/en/stable/table/index.html>`_
+print(con.print_all())
```
dpshelio: This `print` seems redundant.

examples/gallery/constants_reference.py
```diff
@@ -0,0 +1,20 @@
...
+
+The example shows all of the constants provided by sunpy.
+"""
+
+from sunpy.sun import constants as con
+
+##############################################################################
+# All constants are stored in a dictionary. A list of all available keys
+# which describe the constant can be had with the following command.
+con.constants.keys()
```
dpshelio: Whereas I would put a `print` in here.

nabobalis:
Since the CI is on fire, the only build we care about is the docs one.

Issue #2822, opened 2018-10-30
Support as input, and use pathlib everwhere


Cadair:
We should use pathlib now we are Python 3 only. We should make sure all the functions that take path-like inputs accept `Path` objects as well as using pathlib in places like `map.Map` to standardise out slashes over platforms and other quirks of paths.

First attempt at changing test files: #2891

If you plan to undertake this, the best suggestion is to focus on a submodule, one at a time and send that as a pull request.

jains8844:
For the mentioned issue, I have made a certain changes to the following files : 

1. sunpy/map/map_factory.py
https://github.com/jains8844/Contri-to-Issues/blob/master/map_factory.py
2. sunpy/io/file_tools.py
https://github.com/jains8844/Contri-to-Issues/blob/master/file_tools.py
3. sunpy/io/ana.py
https://github.com/jains8844/Contri-to-Issues/blob/master/ana.py
4. sunpy/io/jp2.py
https://github.com/jains8844/Contri-to-Issues/blob/master/jp2.py
5. sunpy/io/fits.py
https://github.com/jains8844/Contri-to-Issues/blob/master/fits.py

These changes would make the reading of headers and data from maps in certain directories to be completely based on pathlib rather than os module and will be working on pathlib objects in place of strings. Whenever string path is needed to be passed, the object can be converted into string else it remains as an object only.
I have not yet worked in writing data to files.

So kindly see if the changes done would work or not .

nabobalis:
It would be easier to review any changes as a pull request. 

But one suggestion off the top is that we do not allow `from pathlib import *`. You will have to use `import pathlib` instead.

jains8844:
Sure :+1:

nabobalis:
Also you will need to fork the SunPy repo and then make those changes your fork, so you can then submit the pull request. I'm not sure we can take the code changes via another method.

darthgera123:
can i work on this issue?

nabobalis:
Sure

darthgera123:
umm the required suggestions are already implemented in the mentioned files above

nabobalis:
Those files are not in our repo.

darthgera123:
so what exactly am i required to do? because the repo i forked & cloned had the suggestions implemented

nabobalis:
What repo did you fork? I'm not aware that our repo (sunpy/sunpy) has pathlib support. 

While there is a pull request open that tries to add support. It has been untouched for a while now. You could start there.

[...]

PR #2884, opened 2019-01-02
Pathlib Functionalities added

darthgera123:
### Description

Added pathlib functionalities in files with testing. Need a review before i progress further. Also i have specifically not touched astropy_helper files since i have no ways of testing the changes there
Fixes issue #2822


nabobalis:
@pep8speaks quiet 

I am not sure what has happened yet, but it seems your changes cause some problems with our tests.

darthgera123:
yeah one change. i ll fix it. could you just check if the rest of the files are inline

nabobalis:
So I have a few comments/suggestions.

First, you change to pathlib over os, you can remove the `import os` if it isn't used in a file.
Second, I thought the API for pathlib is `pathlib.Path(path).exists()` and not `pathlib.Path.exists(path)`.
Thirdly, I don't think we want to use the full import of `pathlib.Path`. `Path` should be enough, which would require changing the import a bit.  

The next stage after those comments would to be tackle the file input and output which might be a large job.

darthgera123:
yeah that was the error there. I forgot to change the exists thing in one file. Actually i wanted to ask how to test my changes in files which dont have unit tests written for them like the entire astropy helpers sub module and other files

nabobalis:
Ignore the astropy helpers sub module. We do not make any changes there. 

I do not just mean `exists`, I mean for any and all extra calls on pathlib. I think you got the API mixed up. 

What do you mean by test your changes specially? As I can see, so far the changes are just in our test files which means if they still run and pass, they are ok.

darthgera123:
yes so for the files which dont have unit tests written how do i test my changes. Also in some files i had syntax errors or something(for testing) but the test didnt figure them. Lastly i would be grateful if you could point out where i have mixed up the calls Ill fix them all of them. Thanks

nabobalis:
I also suggest undoing the changes to `ah_bootstrap.py`. I think but I would have to check but we need that to be python 2 compatible.  Just focus on files inside the sunpy folder for now. 

You mean unit tests for your current changes? As far as I can tell with a quick glance, the changes right now focus on our tests and as such if you just run them that should be enough.

If there are functions that lack unit tests that deal with input and output, we will have to write those. 

It happens every place you have done `Path.something`, those methods (`exists,is_dir` etc) should be called on the Path instance and not class.

darthgera123:
sorry didnt notice the name. Thanks for the help. Ill make the remaining changes within a day or 2

nabobalis:
That is fine, there is no rush!

darthgera123:
@nabobalis  i am getting permission denied errors while running the tests. Could you please help me out

darthgera123:
the errors are on os.remove

nabobalis:
Permission denied? Could you expand on that?

darthgera123:
sorry to bother but i fixed it. However could you please help me out on why my test is failing. Im not able to figure it out.

nabobalis:
Well to start, you need to redo the rebase for this pull request. It seems to have picked up commits that do not belong here.

`git remote add upstream https://github.com/sunpy/sunpy.git`
`git remote update`
`git rebase -i upstream/master`

should clear it. 

Secondly, undo all the changes in `ah_bootstrap.py`, this file comes from outside and should not be changed unless it is necessary. The same goes for `setup.py`, this file should not be changed. 

Doing this should fix your test failures.

darthgera123:
Thanks a lot

nabobalis:
What are the commands you are running for the rebase/merging in master?

darthgera123:
im using git rebase master on the first branch

nabobalis:
Closed in favour of #2891

PR #2887, opened 2019-01-07
CircleCI now builds PRs on top of master

nabobalis:
Suggestion in the matrix chat to make sure that CircleCI aligns with Travis in how it checkouts PRs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,14 @@ skip-check: &skip-check
   name: Check for [ci skip]
   command: bash .circleci/early_exit.sh
 
+merge-check: &merge-check
+  name: Check if we need to merge upstream master
+  command: |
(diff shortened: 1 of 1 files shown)


nabobalis:
Seems that figure test failure is to due pytest 4.2 and pytest-cov. This should be fixed upstream but waiting for package update.

nabobalis:
32bit build fails due to:

```[gw1] linux -- Python 3.6.7 /opt/python/cp36-cp36m/bin/python
Fixture "aia171_test_map" called directly. Fixtures are not meant to be called directly,
but are created automatically when test functions request them as parameters.
See https://docs.pytest.org/en/latest/fixture.html for more information about fixtures, and
https://docs.pytest.org/en/latest/deprecations.html#calling-fixtures-directly about how to update your code.```

PR #2894, opened 2019-01-13
ImageAnimatorWCS unit_x_axis and unit_y_axis follows image_axes

### LINKED ISSUE #2855: ImageAnimatorWCS unit_x_axis and unit_y_axis do not follow image_axes
The units are always applied to the last two dimensions of the WCS, which is just wrong.

Naman9639:
### Description

Fixes #2855

### DIFF
--- a/sunpy/visualization/animator/image.py
+++ b/sunpy/visualization/animator/image.py
@@ -224,12 +224,14 @@ class ImageAnimatorWCS(ImageAnimator):
         return axes
 
     def _set_unit_in_axis(self, axes):
+        x_index = self.slices_wcsaxes.index("x")
+        y_index = self.slices_wcsaxes.index("y")
         if self.unit_x_axis is not None:
-            axes.coords[2].set_format_unit(self.unit_x_axis)
-            axes.coords[2].set_ticks(exclude_overlapping=True)
+            axes.coords[x_index].set_format_unit(self.unit_x_axis)
+            axes.coords[x_index].set_ticks(exclude_overlapping=True)
         if self.unit_y_axis is not None:
-            axes.coords[1].set_format_unit(self.unit_y_axis)
-            axes.coords[1].set_ticks(exclude_overlapping=True)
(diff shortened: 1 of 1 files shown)


nabobalis:
That looks like it might work but we need to check. We need some examples

Naman9639:
> That looks like it might work but we need to check. We need some examples

Examples that show the unit?

Naman9639:
I'll think of an example and add it in the documentation

nabobalis:
Not an example for the documentation. We need a code example to make sure the bug is fixed.

Naman9639:
Ahh ok

Naman9639:
I checked that the indexes of x-axis and y-axis as per the self.slices_wcsaxes list and the value of x_index and y_index matches so what kind of example do we need to add?

nabobalis:
It is not an example we need to add. It is a code example that shows that the code works and fixes the bug. Since this most likely has no tests for it, we can not just change the code and hope it works.

Naman9639:
I understand that. To check that I added print statements to see if the index matches the position of X and Y in the self.slices_wcsaxes.

nabobalis:
That does not mean the bug is fixed. We need an example that shows the bug and that the fix works.

Naman9639:
I am having trouble with the bugged example. I mean it is not clear as to how I show that the bug is fixed.

nabobalis:
@Cadair Any ideas?

Naman9639:
We still need that bugged example right?

nabobalis:
If we can't create one, the changes should not break anything. So I think it is fine to be merged.

Naman9639:
Oh ok

Naman9639:
Um can I work on another issue till this is getting reviewed?

nabobalis:
Sure.

Naman9639:
Did it work?

nabobalis:
Yes it did.

Naman9639:
That's great

PR #2899, opened 2019-01-16
[0.9] Fix CI and backport patches.

### LINKED ISSUE #2710: [0.9] fermi.met_to_utc does not account for leap seconds
@Cadair, @wafels, @ehsteve. 

Fermi Mission Elapsed Time (MET) is the number of seconds elapsed since a pre-defined fiducial, or base time, in this case 2001 January 01 00:00 UT. The current version of the function fermi.met_to_utc is incorrect. It does not account for leap seconds that have occurred since the fiducial.

The in-development version of this function that uses Astropy.time appears to be correct, it accounts for the leap seconds, and gives the same answer provided by XTIME (https://heasarc.gsfc.nasa.gov/cgi-bin/Tools/xTime/xTime.pl). It's possible that XTIME also uses AstroPy, but importantly the same time is returned by tools in SolarSoftWare written by Kim Tolbert and Richard Schwartz.

For example:

Fermi MET = 390009602.0

The current fermi_met_to_utc function will return this as 2013-05-12 00:00:02.000 UTC, but the correct answer is actually 2013-05-11 23:59:59.000 UTC. This is because datetime does not know that three of the elapsed seconds were leap seconds.

This whole situation is complicated by the fact that the headers in Fermi GBM data files show the incorrect time in string format. Their string headers also do not seem to account for leap seconds, which I think is why I didn't catch this error in the first place. When testing this, I saw that the times returned matched the file headers and assumed it to be correct.

This could be fixed by incorporating a lookup table of leap seconds, but if the switch to Astropy.time fixes this anyway it may not be worth it.

---

Note: This is fixed in the 1.0 branch due to the transition to `astropy.time`. 
This is now labelled as `Wont Fix` because we are not fixing it in the 0.9 series.

nabobalis:
It grew too large. 

Fixes #2710  by adding a large warning to the doc string.


nabobalis:
Travis?!

nabobalis:
App Failure due to GOES test on 3.6

nabobalis:
CI fail is due to 
```
E           hypothesis.errors.Flaky: Hypothesis test_transform(lon=<Quantity -165.21875365 deg>, lat=<Quantity -56.50144255 deg>) produces unreliable results: Falsified on the first call but did not on a subsequent one
```

nabobalis:
The warning in fermi: https://2233-85053060-gh.circle-artifacts.com/0/root/project/docs/_build/html/api/sunpy.instr.fermi.met_to_utc.html#sunpy.instr.fermi.met_to_utc

nabobalis:
Also Windows Py2.7 Issues is still there, it just xpasses the build now.

sunpy/instr/fermi.py
```diff
@@ -451,6 +451,11 @@ def met_to_utc(timeinsec):
     """
     Converts Fermi Mission Elapsed Time (MET) in seconds to a datetime object.
 
+    .. warning::
+        This function does not account for leap seconds and can not be fixed in
+        SunPy's 0.9.x series of releases. However, it is fixed in the next major version
+        coming by May 2019.
```
Cadair: I would say this will be fixed in 1.0, putting release dates in the docs makes me nervous :p

nabobalis: Sorted mate

PR #2900, opened 2019-01-18
Added mocking in fido_search for test_noaa

yashrsharma44:
1. Note that this is test commit, demonstrating the mocking of `Fido.search` and `Fido.fetch`
2. Further tests using mock would be written after this pattern of mocking is useful

### Description

Partial work for #2874

PR #2901, opened 2019-01-19
remove mock dependency

dpshelio:
From @yashrsharma44's pull request #2900 I've noticed we were using `mock` in other places, but mock is since 3.3 part of python's unittest. Since we are requiring python 3.6 this dependency can be removed.


nabobalis:
Looks like a real test fail.

nabobalis:
Issue with windows build seems to be an upstream DLL issue with pillow.

PR #2904, opened 2019-01-21
Add support to download header information form helioviewer API.

### LINKED ISSUE #2762: Add support to download the header information from helioviewer
Currently our HelioViewer Client (https://github.com/sunpy/sunpy/blob/master/sunpy/net/helioviewer.py) allows us to download jpeg files from the service. They (Helioviewer) have an a better API now (v2 which is linked in the original post) that allows us to also download the metadata of the jpeg files with a seperate function. We do not have that seperate function, which is what this issue is tracking. 

The API returns XML data so we want to convert it to a nicer dictionary format.

Naman9639:
### Description

Fixes #2762


Naman9639:
We require some new modules to be added. That is why the tests are failing. How to solve this?

nabobalis:
You will need to add the new package to setup.cfg under the net list. You will need to also add it to the tox.ini file under deps/condadeps (depending on if it's on conda or not)

Naman9639:
I googled on how to convert xml doc to dictionary so this was the most supported solution

Naman9639:
I still face problem in rebasing and keeping my branch up to date with master

nabobalis:
What was the problem?

Naman9639:
> Is there no way to do it without the new package?
> 
> You will need to add the new package to setup.cfg under the net list. You will need to also add it to the tox.ini file under deps/condadeps (depending on if it's on conda or not).
> 
> Also make sure your up to date with master. We had to merge in some fixes for the tests.

`net_requires = drms, zeep, beautifulsoup4, python-dateutil` 
shall I add the new packages to this list?

Naman9639:
> What was the problem?

My history gets messed up and then I am generally not able to rebase not even merge it with master. I had to clone the updated repository again and that's how it always works for me

nabobalis:
Yes that is the list.

Make sure you have done:
```
git remote add upstream git@github.com:sunpy/sunpy.git
git remote update
(on this branch then do)
git rebase -i upstream/master
git push --force
```
That should work. I would check the log after the rebase.

Naman9639:
Ok I will try that

nabobalis:
I would try to avoid `#` in the branch name. It can cause issues.

Naman9639:
> I would try to avoid `#` in the branch name. It can cause issues.

Oh ok I will keep that in mind

Naman9639:
While doing this 
`git rebase -i upstream/master`
Do I need to pick every commit? I always mess up at this step

nabobalis:
There should be only one commit which is this one.

Naman9639:
Looks like I already messed up because I am getting a whole lot of commits

nabobalis:
What are those commits?

Naman9639:
commits
 They are some of the many

nabobalis:
Well in that case, it would seem somehow your locally git history is incorrect.  

I would exit and abort the rebase.

Could you try:
`git remote prune upstream`
`git remote update`

That should refresh your local version of upstream hopefully.

Naman9639:
That worked

nabobalis:
Do you need to use `requests`? We used to have it but we removed it.

nabobalis:
But we don't seem to need it to access the API for the other functions.

Naman9639:
I can try it using the self. _request function once

changelog/2904.feature.rst
```diff
@@ -0,0 +1 @@
+Header information can be downloaded using get_jp2_header function. The API returns the information in XML format but using xmltodict library we convert the xml file into a dictionary. Two new libraries are added to support the same.
```
nabobalis: This needs to be reworded. 

```python
Provided access to the Helioviewer header information using `~sunpy.net.helioviewer.HelioviewerClient.get_jp2_header` function. This requires the installation of the `xmltodict` library, if you want to use this function.
```

sunpy/net/helioviewer.py
```diff
@@ -14,6 +14,9 @@
 from sunpy.time import parse_time
 from sunpy.util.net import download_fileobj
 
+import requests
+import xmltodict
```
nabobalis: These imports need to be merged into the ones above in the correct places.

sunpy/net/helioviewer.py
```diff
@@ -189,6 +192,63 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
 
         return self._get_file(params, directory=directory, overwrite=overwrite)
 
+    def get_jp2_header(self, date, observatory=None, instrument=None, detector=None,
+                       measurement=None, jp2_id=None, directory=None, overwrite=False):
```
nabobalis: Why are `directory=None, overwrite=False` here?

sunpy/net/helioviewer.py
```diff
@@ -189,6 +192,63 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+            Detector name
+        directory : `str`
+            Directory to download JPEG 2000 image to.
+        overwrite : bool
+            Defaults to False.
+            If set to True, will overwrite any files with the same name.
+
+        Returns
+        -------
+        out : `Dict`
```
nabobalis: lower case D here

sunpy/net/helioviewer.py
```diff
@@ -189,6 +192,63 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+        --------
+        >>> from sunpy.net import helioviewer
+        >>> hv = helioviewer.HelioviewerClient()  # doctest: +REMOTE_DATA
+        >>> header = hv.get_jp2_header('2012/07/03', observatory='SDO',
+        ...                            instrument='HMI', detector=None, measurement='continuum')  # doctest: +REMOTE_DATA
+        """
+        if jp2_id is None:
+            try:
+                key = (observatory, instrument, detector, measurement)
+                print(key)
```
nabobalis: We can probably remove the print now.

sunpy/net/helioviewer.py
Naman9639: I wanted to ask this. Do we wish to provide the freedom to download the headers in any of the directory or shall we remove this?

I mean do we really need to keep these two parameters or can we omit them?

sunpy/net/helioviewer.py
```diff
@@ -189,6 +192,63 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+                source_id = self.data_sources[key]
+                metadata = self.get_closest_image(date, source_id = source_id)
+                jp2_id = metadata['id']
```
nabobalis: We could probably condense these 4 lines into one line.

sunpy/net/helioviewer.py
```diff
@@ -189,6 +192,63 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+            except KeyError:
+                raise KeyError("The values used for observatory, instrument, detector, measurement "
+                               "do not correspond to a source_id. Please check the list using "
+                               "HelioviewerClient.data_sources.")
+
+        URL = ('https://api.helioviewer.org/v2/getJP2Header/?id=' + (jp2_id))
```
nabobalis: Do we need to provide the URL here?

sunpy/net/helioviewer.py
```diff
@@ -189,6 +192,63 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+
+        params = {
+            "action": "getJP2Header",
+            "id" : jp2_id,
+        }
```
nabobalis: What does this do?

nabobalis:
You will also need to add the new dependency to the .circleci/config file for the 32 bit section. That just has a list of all the dependencies.

docs/guide/acquiring_data/helioviewer.rst
```diff
@@ -20,6 +20,7 @@ We provide the follwing functions:
 2. As above but return the full JSON response instead of an image.
 3. Return all available datasources.
 4. Generates custom screenshots that allow labels and layers of images.
+5. Download the XML header embedded in a JPEG2000 image. This includes the FITS header as well as a section of Helioviewer-specific metadata.
```
nabobalis: This should be reworded, I am not a fan of a copy and paste.

sunpy/net/helioviewer.py
nabobalis: We are returning a dictionary. I don't think we are meant to download anything.

Naman9639: But to convert the XML data into dictionary I first download the XML file and then convert it into a dictionary. That is also the reason for using `requests` module

sunpy/net/helioviewer.py
Naman9639: To get the header from the API we need to pass a unique JP2 image identifier which is the jp2_id

docs/guide/acquiring_data/helioviewer.rst
Naman9639: Sure I will change that

sunpy/net/helioviewer.py
Naman9639: We can provide it in the `self._get_header_information` function as well.

sunpy/net/helioviewer.py
nabobalis: Do you have to create a downloaded file to read it and parse it?

sunpy/net/helioviewer.py
```diff
@@ -390,3 +450,11 @@ def _request(self, params):
...
         """Formats a date for Helioviewer API requests"""
         return parse_time(date).isot + "Z"
+
+    def _get_header_information(self, URL):
+        response = requests.get(URL)
+        with open('header.xml', 'wb') as file:
+            file.write(response.content)
+        with open('header.xml') as fd:
+            doc = xmltodict.parse(fd.read())
+        return doc
```
nabobalis: Do you need to do `with open` twice?

sunpy/net/helioviewer.py
Naman9639: I guess so. That is an easy way of converting the XML file into a dictionary

nabobalis: Can't you convert the response directly?

sunpy/net/helioviewer.py
nabobalis: params is not used anywhere.

sunpy/net/helioviewer.py
Naman9639: Oh yeah sorry. I did that while testing the code on my machine

sunpy/net/helioviewer.py
nabobalis: I mean we don't need to use the URL directly in the other functions. So why here?

nabobalis:
What does the output look like from the function?

nabobalis:
We will also need a unit test or two for the new function.

sunpy/net/helioviewer.py
Naman9639: I don't think so because the API just redirects you to a page containing the XML response.

Naman9639:
> What does the output look like from the function?

It is messy
output

nabobalis:
The keys for the returned dictionary should be mentioned somewhere.

Naman9639:
> You will also need to add the new dependency to the .circleci/config file for the 32 bit section. That just has a list of all the dependencies.

Um where is this file? I am not able to find it

nabobalis:
Here: https://github.com/sunpy/sunpy/blob/master/.circleci/config.yml#L96

Naman9639:
> Here: https://github.com/sunpy/sunpy/blob/master/.circleci/config.yml#L96

I don't have this file in my local repo and I don't have a lot of other files as well

nabobalis:
The file should be there. 

What does a `git status` report?

nabobalis:
If you are on linux, those are hidden by default with a file manager or `ls`. They exist you just have to open it with tab completion or typing the full name.

Naman9639:
> If you are on linux, those are hidden by default with a file manager or `ls`. They exist you just have to open it with tab completion or typing the full name.

Oh ok

nabobalis:
`ls -a` should display them.

Naman9639:
> The keys for the returned dictionary should be mentioned somewhere.

`header.keys()` only gives one key i.e ['meta'].

nabobalis:
Well in that case, we should at least index that layer out of the return.

Naman9639:
> Well in that case, we should at least index that layer out of the return.

I am able to go 2 level deep with the following keys
`header['meta']['fits']`

nabobalis:
Yes but we want to return the other keyword too.

Naman9639:
I tried something
We can get all the keys of fits header as well as keys for helioviewer metadata

.circleci/config.yml
```diff
@@ -97,6 +97,9 @@ jobs:
       - run:
           name: Run tests
           command: PYTHONHASHSEED=42 /opt/python/cp36-cp36m/bin/python setup.py test --parallel=4
+      - run:
+          name: xml to dictionary conversion dependencies
+          command: conda install -c conda-forge xmltodict
```
nabobalis: This entire block needs to be removed. The package needs to be added to the list I linked earlier.

docs/guide/acquiring_data/helioviewer.rst
```diff
@@ -20,6 +20,9 @@ We provide the follwing functions:
 2. As above but return the full JSON response instead of an image.
 3. Return all available datasources.
 4. Generates custom screenshots that allow labels and layers of images.
+5. Download the header information present in a JPEG2000 image. This includes:
+    5.1. FITS header
```
nabobalis: These probably should just be bullet points

sunpy/net/helioviewer.py
```diff
@@ -189,6 +192,57 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+                jp2_id = self.get_closest_image(date, observatory, instrument, detector, measurement)['id']
+            except KeyError:
+                raise KeyError("The values used for observatory, instrument, detector, measurement "
+                               "do not correspond to a source_id. Please check the list using "
+                               "HelioviewerClient.data_sources.")
+
+        params = {
+            "action": "getJP2Header",
+            "id" : jp2_id,
+        }
```
nabobalis: I think it would be nice if we split this line up.

sunpy/net/helioviewer.py
```diff
@@ -189,6 +192,57 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+
+        responses = self._request(params).read().decode('utf-8')
```
nabobalis: Just have this return the line above instead

nabobalis:
What do you mean?

I think we can just remove the outer meta dictionary on the return. That would probably be enough

sunpy/net/helioviewer.py
nabobalis: These should be referring to the line below.

sunpy/net/helioviewer.py
```diff
@@ -6,6 +6,8 @@
 import errno
 import codecs
 import urllib
+import requests
```
nabobalis: Now you don't need requests, you can remove it.

setup.cfg
```diff
@@ -14,7 +14,7 @@ extra_requires = database, image, jpeg2000, net, tests, docs, asdf
 database_requires = sqlalchemy
 image_requires = scikit-image
 jpeg2000_requires = glymur
-net_requires = drms, zeep, beautifulsoup4, python-dateutil
+net_requires = drms, zeep, beautifulsoup4, python-dateutil, xmltodict, requests
```
nabobalis: We can now remove requests here too

sunpy/net/helioviewer.py
```diff
@@ -189,6 +192,57 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+                               "HelioviewerClient.data_sources.")
+
+        params = {
+            "action": "getJP2Header",
+            "id" : jp2_id,
+        }
+
+        responses = self._request(params).read().decode('utf-8')
+        header_info = xmltodict.parse(responses)
+        return header_info
```
nabobalis: Have this return the line above

sunpy/net/helioviewer.py
nabobalis: I feel like this to too many operations in one line. Also a comment here would be good

tox.ini
```diff
@@ -42,6 +42,8 @@ conda_deps =
     scipy
     sqlalchemy
     zeep
+    xmltodict
+    requests
```
nabobalis: We can remove this as well now.

Naman9639:
> What do you mean?
> 
> I think we can just remove the outer meta dictionary on the return. That would probably be enough

`
fits_header = header['meta']['fits']  #The keys 'meta' and 'fits' can be used to get the fits header information
helioviewer_meta_data = header['meta']['helioviewer'] #The keys 'meta' and 'helioviewer' can be used to extract the helioviewer specific metadata.
`
I meant this. We can remove the outer meta dictionary on the return but in this way we will be able to provide more details to the user

Naman9639:
No no it will just help in showing the two data separately i.e the fits header and helioviewer metadata.
The XML file contains two main tags under meta tag 
The fits tag and the helioviewer tag. So I thought it will be good if we let the user access both thr information separately

Naman9639:
Letting the user access the two data i.e fits header and helioviewer data separately. I just added two commands to access that information.

nabobalis:
But they can do that already. Which is why I am confused.

Naman9639:
Yeah they can. I just thought of adding it to example in case someone needs it. I'll remove it

nabobalis:
Oh, you want to add it as an example. That's fine.

nabobalis:
I think we should also add the keywords in the return dict to the doc string for the function under the `Returns` section. 

Also in the test that the return has a len of 2 (i think it should) and that it has the two keywords we expect.

nabobalis:
This shows how little I am aware of the SunPy codebase, but we seem to have some sort of xml to dict function in SunPy. See `sunpy/util/xml.py`

Could you try to use that instead of the new module?

Naman9639:
> This shows how little I am aware of the SunPy codebase, but we seem to have some sort of xml to dict function in SunPy. See `sunpy/util/xml.py`
> 
> Could you try to use that instead of the new module?

Sure I will do that

Naman9639:
> I think we should also add the keywords in the return dict to the doc string for the function under the `Returns` section.
> 
> Also in the test that the return has a len of 2 (i think it should) and that it has the two keywords we expect.

I'm sorry but I didn't get that

nabobalis:
In the documentation for the function, under the return section, we should say that the returned dictionary has two keys and name them. 

In the test you wrote we should also check that the dictionary has the required length and that the two keys are in the dictionary.

sunpy/net/helioviewer.py
```diff
@@ -189,6 +192,57 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+
+        Examples
+        --------
+        >>> from sunpy.net import helioviewer
+        >>> hv = helioviewer.HelioviewerClient()  # doctest: +REMOTE_DATA
+        >>> header = hv.get_jp2_header('2012/07/03', observatory='SDO',
+        ...                            instrument='HMI', detector=None, measurement='continuum')  # doctest: +REMOTE_DATA
+        >>> fits_header = header['meta']['fits']  #The keys 'meta' and 'fits' can be used to get the fits header information
+        >>> helioviewer_meta_data = header['meta']['helioviewer'] #The keys 'meta' and 'helioviewer' can be used to extract the
+                                                                  #helioviewer specific metadata.
```
nabobalis: These comments need to be moved. They should be above the line of code, otherwise the line is too long. 

Also these will need `# doctest: +REMOTE_DATA`

sunpy/net/helioviewer.py
```diff
@@ -189,6 +192,57 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
 
         return self._get_file(params, directory=directory, overwrite=overwrite)
 
+    def get_jp2_header(self, date, observatory=None, instrument=None, detector=None,
+                       measurement=None, jp2_id=None):
```
nabobalis: Can this be made one line?

Naman9639: We cannot remove anything as these are essential parameters. We can keep it in one line but it will be a too much in a single line

nabobalis: Is it too long if its one line?

Naman9639: I guess so.

.circleci/config.yml
```diff
@@ -93,10 +93,10 @@ jobs:
           command: /opt/python/cp36-cp36m/bin/pip install --progress-bar off numpy==1.15.3 scipy astropy!=3.1.1 matplotlib==1.5.3
       - run:
           name: Install dependencies two
-          command: /opt/python/cp36-cp36m/bin/pip install --progress-bar off sqlalchemy scikit-image==0.13.1 glymur drms zeep beautifulsoup4 requests python-dateutil pytest pytest-cov pytest-mock pytest-xdist hypothesis pytest-astropy pytest-rerunfailures
+          command: /opt/python/cp36-cp36m/bin/pip install --progress-bar off sqlalchemy scikit-image==0.13.1 glymur drms zeep beautifulsoup4 requests python-dateutil pytest pytest-cov pytest-mock pytest-xdist hypothesis pytest-astropy pytest-rerunfailure
       - run:
           name: Run tests
-          command: PYTHONHASHSEED=42 /opt/python/cp36-cp36m/bin/python setup.py test --parallel=4
+          command: PYTHONHASHSEED=42 /opt/python/cp36-cp36m/bin/python setup.py test --parallel=4
```
nabobalis: Can you return this file back to normal?

sunpy/net/helioviewer.py
```diff
@@ -189,6 +191,58 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+        measurement : `str`
+            Measurement name
+        detector : `str`
+            Detector name
+
+        Returns
+        -------
+        out : `dict`
+            Returns a dictionary containing the header information of JPEG 2000 image. The returned dictionary
+            has two keys namely *fits* and *helioviewer*
```
nabobalis: Can you move the sentence to a new line? 

Also change it to: `The returned dictionary has two keys: *fits* and *helioviewer*`

sunpy/net/helioviewer.py
```diff
@@ -189,6 +191,58 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+
+        Examples
+        --------
+        >>> from sunpy.net import helioviewer
+        >>> hv = helioviewer.HelioviewerClient()  # doctest: +REMOTE_DATA
+        >>> header = hv.get_jp2_header('2012/07/03', observatory='SDO',
+        ...                            instrument='HMI', detector=None, measurement='continuum')  # doctest: +REMOTE_DATA
+        #The key 'fits' can be used to get the fits header information
+        >>> fits_header = header['fits']  # doctest: +REMOTE_DATA
+        #The keys 'meta' and 'helioviewer' can be used to extract the helioviewer specific metadata.
```
nabobalis: remove `meta` here.

sunpy/net/helioviewer.py
```diff
@@ -189,6 +191,58 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+                               "do not correspond to a source_id. Please check the list using "
+                               "HelioviewerClient.data_sources.")
+
+        params = {
+            "action": "getJP2Header",
+            "id" : jp2_id,
+        }
+
+        responses = self._request(params)
+        responses = responses.read().decode('utf-8')  #Reads the output from HTTPResponse object and decodes it.
```
nabobalis: can the comment be moved to a new line above this line of code?

sunpy/net/tests/test_helioviewer.py
```diff
@@ -74,6 +74,18 @@ def test_download_jp2(self, client):
...
+        """
+        Tests getJP2Header API method
+        """
+        header1 = client.get_jp2_header('1994/01/01', observatory='SOHO',
+                                              instrument='EIT', measurement='304')
+        header2 = client.get_jp2_header('1994/01/01', jp2_id = 1795504)
+        keys = header1.keys()
+        assert cmp(header1 == header2)
+        assert len(header1) == 2
+        assert 'fits' and 'helioviewer' in keys
```
nabobalis: Can we do this check for both headers?

Naman9639: Sure I will add that

Naman9639:
I am sorry but can you tell me how to test for online data as well on my local machine?

nabobalis:
`pytest --remote-data=any <path to test file>`

.circleci/config.yml
```diff
@@ -93,7 +93,7 @@ jobs:
           command: /opt/python/cp36-cp36m/bin/pip install --progress-bar off numpy==1.15.3 scipy astropy!=3.1.1 matplotlib==1.5.3
       - run:
           name: Install dependencies two
-          command: /opt/python/cp36-cp36m/bin/pip install --progress-bar off sqlalchemy scikit-image==0.13.1 glymur drms zeep beautifulsoup4 requests python-dateutil pytest pytest-cov pytest-mock pytest-xdist hypothesis pytest-astropy pytest-rerunfailures
+          command: /opt/python/cp36-cp36m/bin/pip install --progress-bar off sqlalchemy scikit-image==0.13.1 glymur drms zeep beautifulsoup4 requests python-dateutil pytest pytest-cov pytest-mock pytest-xdist hypothesis pytest-astropy pytest-rerunfailure
```
nabobalis: This needs to be fixed too.

Naman9639: Ahh now I get it why it was failing every time

sunpy/net/helioviewer.py
```diff
@@ -189,6 +191,59 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+        ----------
+        observatory : `str`
+            Observatory name
+        instrument : `str`
+            Instrument name
+        measurement : `str`
+            Measurement name
+        detector : `str`
+            Detector name
+
```
nabobalis: Need a entry for `jp2_id`

nabobalis: Also for `date`

sunpy/net/tests/test_helioviewer.py
```diff
@@ -74,6 +74,21 @@ def test_download_jp2(self, client):
...
+        header1 = client.get_jp2_header('1994/01/01', observatory='SOHO',
+                                              instrument='EIT', measurement='304')
+        header2 = client.get_jp2_header('1994/01/01', jp2_id = 1795504)
+        keys_header1 = header1.keys()
+        keys_header2 = header2.keys()
+        assert cmp(header1 == header2)
+        assert len(header1) == 2
+        assert len(header2) == 2
+        assert 'fits' and 'helioviewer' in keys_header1
+        assert 'fits' and 'helioviewer' in keys_header2
```
nabobalis: We can probably shorten the lines. 

```
        assert cmp(header1 == header2)
        assert len(header1) == len(header2) == 2
        assert 'fits' and 'helioviewer' in header1.keys()
        assert 'fits' and 'helioviewer' in header2.keys()  

 ```
I also wonder if the last two lines could be made into one.

File: sunpy/net/helioviewer.py (hunk 27 above)
Comment (Naman9639) [reply sunpy/net/helioviewer.py]: Oh ok I will add that

**Comment by Naman9639** [comment]
Ok, I guess there is a twist. Not every dictionary has both the keys. The API said that XML will contain fits header and helioviewer metadata but I guess that is not the case with every example. I don't know how these asserts statements were working till now but for this `header1 = client.get_jp2_header('1994/01/01', observatory='SOHO', instrument='EIT', measurement='304') `
the length of dictionary is 1 and it only contain fits data but they still passed the test I don't  know how

**Comment by nabobalis** [comment]
Well we need to make a comment somewhere about that fact. 

Also I don't know if the tests have passed, the one build that runs that test timed out.

**Comment by Naman9639** [comment]
But we don't know how many keys the dictionary will have. It can either be 1 or 2. So we can't add the test for that and same goes for the test that checks if both 'fits' and 'helioviewer' are present as keys

**Comment by nabobalis** [comment]
But we do know.

**Comment by Naman9639** [comment]
Sorry I didn't get that.

**Comment by nabobalis** [comment]
Well the test code is fixed. We know its outcome.

**Comment by Naman9639** [comment]
> Well the test code is fixed. We know its outcome.

Yeah for the test we can do that. I will add something in the function description regarding this thing that we came across. But I am still confused. It means the definition provided by the API is wrong? Because the API says 
`Get the XML header embedded in a JPEG2000 image. Includes the FITS header as well as a section of Helioviewer-specific metadata.`

This means both FITS and Helioviewer specific metadata should be present

**Comment by nabobalis** [comment]
I don't know, I would have to ask someone that worked on the API.

I would assume that maybe they did not add the metadata for older data but started doing it on newer data, or maybe that data doesn't need any specific helioviewer information.

**Comment by Naman9639** [comment]
Why is there an error in this?
I am not able to understand this one

**Comment by nabobalis** [comment]
I think you need to add `>>>` to the comment lines.

**Comment by Naman9639** [comment]
Can't we add comments without that?

**Comment by nabobalis** [comment]
I do not think so in the examples.

File: sunpy/net/helioviewer.py [hunk 29]
```diff
@@ -189,6 +191,65 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+        Examples
+        --------
+        >>> from sunpy.net import helioviewer
+        >>> hv = helioviewer.HelioviewerClient()  # doctest: +REMOTE_DATA
+        >>> header = hv.get_jp2_header('2012/07/03', observatory='SDO',
+        ...                            instrument='HMI', detector=None, measurement='continuum')  # doctest: +REMOTE_DATA
+        >>> #The key 'fits' can be used to get the fits header information
+        >>> fits_header = header['fits']  # doctest: +REMOTE_DATA
+        >>> #The keys 'helioviewer' can be used to extract the helioviewer specific metadata.
```
nabobalis: Can I get a space between the `#` and the comment text? For both lines?

sunpy/net/helioviewer.py
```diff
@@ -189,6 +191,65 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+                               "do not correspond to a source_id. Please check the list using "
+                               "HelioviewerClient.data_sources.")
+
+        params = {
+            "action": "getJP2Header",
+            "id" : jp2_id,
+        }
+
+        responses = self._request(params)
+        #Reads the output from HTTPResponse object and decodes it.
```
nabobalis: Same as above.

sunpy/net/tests/test_helioviewer.py
```diff
@@ -74,6 +74,19 @@ def test_download_jp2(self, client):
...
+    def test_get_jp2_header(self, client):
+        """
+        Tests getJP2Header API method
+        """
+        header1 = client.get_jp2_header('1994/01/01', observatory='SOHO',
+                                              instrument='EIT', measurement='304')
+        header2 = client.get_jp2_header('1994/01/01', jp2_id = 1795504)
+        keys_header1 = header1.keys()
+        keys_header2 = header2.keys()
+        assert (header1 == header2)
```
nabobalis: Can you remove the extra brackets here.

nabobalis:
I think just some really minor text changes and it is good to go.

Naman9639:
Text changes? Like some changes in documentation and all that?

nabobalis:
No I just meant the comments I just left. It is just some text changes.

Naman9639:
Ok I will do that

Naman9639:
@nabobalis Can you restart those test?

Naman9639:
These errors are not related to helioviewer right?

Naman9639:
Why do it times out again and again?

nabobalis:
Some server is just not responding and the test seems to hang.  With that in mind, I will merge this now.

nabobalis:
Thanks again @Naman9639

Naman9639:
Oh, ok. Thanks

PR #2907, opened 2019-01-24
CI test Fix and hack around for goes time

nabobalis:


sunpy/net/tests/strategies.py
```diff
@@ -99,8 +100,13 @@ def goes_time(draw, time=Times(
 
     tr = TimeRange(t1, t2)
     # There is no GOES data for this date.
-    assume(Time('1983-5-1') not in tr)
-    assume(Time('1983-5-1') + draw(delta) not in tr)
+    # We seem to pick up an extra micro second when we create t1
+    # TODO: Track this issue down.
+    failing_time = Time('1983-05-01')
```
Cadair: Oh gods, does this actually work?

nabobalis: It does seem that way using pdb that I can now trigger the `assume`

nabobalis: @vn-ki you got any advice?

nabobalis:
note to self: squash merge.

PR #2922, opened 2019-02-01
moved to using appdirs for sunpyrc config

### LINKED ISSUE #2869: Use appdirs for our config file locations
https://pypi.org/project/appdirs/

nabobalis:
### Description

Fixes #2869

TODO: 

- [x] Write tests

### DIFF
--- a/sunpy/util/config.py
+++ b/sunpy/util/config.py
@@ -1,11 +1,19 @@
 """SunPy configuration file functionality"""
 import os
-import tempfile
 import configparser
+from pathlib import Path
 
 import sunpy
+from sunpy.extern.appdirs import AppDirs
 
-__all__ = ['load_config', 'print_config']
+
+__all__ = ['load_config', 'print_config', 'CONFIG_DIR']
+
+# This is to avoid creating a new config dir for each new dev version.
+# We use AppDirs to locate and create the config directory.
+dirs = AppDirs("sunpy", "sunpy")
+# Default one set by AppDirs
+CONFIG_DIR = dirs.user_config_dir
 
 
 def load_config():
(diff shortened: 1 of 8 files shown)


sunpy/util/config.py
```diff
@@ -24,40 +36,59 @@ def load_config():
...
 
 
+def _find_config_files():
+    """Finds locations of SunPy configuration files"""
+    config_files = []
+    config_filename = 'sunpyrc'
+
+    # find default configuration file
+    module_dir = Path(sunpy.__file__).parent
+    config_files.append(str(module_dir / 'data' / 'sunpyrc'))
```
dpshelio: I believe this works:
```suggestion
    config_files.append(f"{module_dir / 'data' / 'sunpyrc'}")
```

sunpy/util/config.py
```diff
@@ -24,40 +36,59 @@ def load_config():
...
-    filepaths = [
-        ('downloads', 'download_dir'),
-        ('downloads', 'sample_dir')
-    ]
-    _fix_filepaths(config, filepaths)
+    working_dir = Path(config.get('general', 'working_dir'))
+    sample_dir = Path(config.get('downloads', 'sample_dir'))
+    download_dir = Path(config.get('downloads', 'download_dir'))
+    config.set('downloads', 'sample_dir', str((working_dir / sample_dir).expanduser().resolve()))
+    config.set('downloads', 'download_dir', str((working_dir / download_dir).expanduser().resolve()))
```
dpshelio: Can't config use `Path` directly?

sunpy/util/config.py
dpshelio: Similarly in other places.

sunpy/util/config.py
```diff
@@ -78,54 +109,13 @@ def print_config():
...
-
-    return config_files
+    return Path(p).is_dir() and os.access(p, os.W_OK)
 
 
 def _get_user_configdir():
     """
     Return the string representing the configuration dir.
-    The default is "HOME/.sunpy".  You can override this with the
+    The default is "HOME/.sunpy". You can override this with the
```
dpshelio: I think it makes more sense if it is `$HOME`
```suggestion
    The default is "$HOME/.sunpy". You can override this with the
```

sunpy/util/config.py
```diff
@@ -1,11 +1,23 @@
...
+
+# This is to avoid creating a new config dir for each new dev version.
+# We use AppDirs to locate and create the config directory.
+dirs = AppDirs("sunpy", "sunpy")
+# Default one set by AppDirs
+CONFIG_DIR = dirs.user_config_dir
+
+# This is to avoid creating a new config dir for each new dev version.
+# We use AppDirs to locate and create the config directory.
+dirs = AppDirs("sunpy", "sunpy")
```
dpshelio: Why is this repeated also on l14?

nabobalis: Because I can't do a rebase merge correctly.

sunpy/util/config.py
nabobalis: That is a good question and one I did not plan to test out until we plan to move to `Path` in more files.

sunpy/util/config.py
nabobalis: I mean it works fine, it feels a little overkill to str something?

nitish705702: i think, it work good,very little change f over str.

Cadair: There is also `Path.as_posix()` for a third option.

I think `str()` is probably lower overhead on the interpreter and a little more familiar to people, but I don't really think it makes any real difference.

nabobalis:
Seems that `hec.HECClient()` found no servers.

PR #2926, opened 2019-02-05
Simplified user access to the sourceID variable in the HelioviewerClient. 

### LINKED ISSUE #2878: Accessing the Helioviewer sourceid with fewer keywords.
Currently for the Helioviewer Client, two functions (`get_closest_image` and `download_jp2`) need you to specify 4 keywords to access the `source_id` dictionary. These are `observatory=None, instrument=None, detector=None, measurement=None`.

We want to be able to support a user submitting say two keywords like for example, `observatory=SDO, measurement=171` or `instrument=AIA, measurement=171`.
This is enough information in theory to get the right source_id for **most** of them.
If it can not, it should through a nice error.   

Any change would be inside our functions. This is not related to the API of Helioviewer.
Ideally where we directly index the dictionary for the `source_id`, we would replace that part with something much smarter. What? I don't know.

Naman9639:
### Description

Fixes #2878


sunpy/net/helioviewer.py
```diff
@@ -109,10 +117,15 @@ def get_closest_image(self, date, observatory=None, instrument=None,
         >>> print(metadata['date'])  # doctest: +REMOTE_DATA
         2012-01-01T00:00:07.000
         """
+        data_dictionary = self.data_sources
```
nabobalis: Why is this defined here?

sunpy/net/helioviewer.py
```diff
@@ -109,10 +117,15 @@ def get_closest_image(self, date, observatory=None, instrument=None,
...
         if source_id is None:
             try:
                 key = (observatory, instrument, detector, measurement)
-                source_id = self.data_sources[key]
+                source_id_list = list(self._partial_match(key, dictionary=data_dictionary))
+                if len(source_id_list) == 1:
+                    source_id = source_id_list[0]
+                else:
+                    raise KeyError
```
nabobalis: Why does this raise a different `KeyError` and not just the one below?

sunpy/net/helioviewer.py
```diff
@@ -171,13 +184,19 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
```
nabobalis: The same as above.

sunpy/net/helioviewer.py
```diff
@@ -171,13 +184,19 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
         >>> hv = helioviewer.HelioviewerClient()  # doctest: +REMOTE_DATA
         >>> filepath = hv.download_jp2('2012/07/03 14:30:00', observatory='SDO',
         ...                            instrument='HMI', detector=None, measurement='continuum')  # doctest: +REMOTE_DATA
+        >>> filepath1 = hv.download_jp2('2012/07/03 14:30:00', observatory='SDO', measurement='continuum')  # doctest: +REMOTE_DATA
         >>> aia = sunpy.map.Map(filepath)  # doctest: +REMOTE_DATA
         >>> aia.peek()  # doctest: +SKIP
         """
+        data_dictionary = self.data_sources
```
nabobalis: The same as above.

sunpy/net/helioviewer.py
```diff
@@ -171,13 +184,19 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
         >>> hv = helioviewer.HelioviewerClient()  # doctest: +REMOTE_DATA
         >>> filepath = hv.download_jp2('2012/07/03 14:30:00', observatory='SDO',
         ...                            instrument='HMI', detector=None, measurement='continuum')  # doctest: +REMOTE_DATA
+        >>> filepath1 = hv.download_jp2('2012/07/03 14:30:00', observatory='SDO', measurement='continuum')  # doctest: +REMOTE_DATA
```
nabobalis: `filepath1` -> `filepath`

sunpy/net/tests/test_helioviewer.py
```diff
@@ -73,6 +73,18 @@ def test_download_jp2(self, client):
                                        instrument='MDI', measurement='continuum')
         assert "2011_01_11__22_39_00_000__SOHO_MDI_MDI_continuum.jp2" in filepath
         os.remove(filepath)
+        
+    def test_partial_match(self, client):
+        """
+        Tests the function that helps to access the helioviewer source id 
+        with fewer keywords.
+        """
```
nabobalis: Not sure we need this doc string.

sunpy/net/tests/test_helioviewer.py
```diff
@@ -73,6 +73,18 @@ def test_download_jp2(self, client):
...
+        data_dict = client.data_sources
+        key_1 = ('SOHO', 'MDI', None, 'continuum')
+        source_id_1 = data_dict[key_1]
+        key_2 = ('SOHO', None, None, 'continuum')
+        source_id_2 = list(client._partial_match(key=key_2, dictionary=data_dict))
```
nabobalis: Why do you use `list` here? Also why are these passed in as keywords?

sunpy/net/helioviewer.py
```diff
@@ -57,6 +57,14 @@ def data_sources(self):
         # Sort the output for printing purposes
         return OrderedDict(sorted(data_sources_dict.items(), key=lambda x: x[1]))
 
+    def _partial_match(self, key, dictionary):
+        """
+        A function that returns source id based on partial key.
+        """
```
nabobalis: The function is more general than that.

Also if you found this online, I would link to it so it adds more context for someone reading the code.

changelog/2926.trivial.rst
```diff
@@ -0,0 +1 @@
+User can access the Helioviewer API using fewer keywords. Either by `observatory` and `measurement` or `instrument` and `measurement`.
```
nabobalis: I would expand upon this change log, maybe with a short example.

sunpy/net/helioviewer.py
Naman9639: We can define this in the init function. We need the dictionary because it is used to get the value from the partial key. We can define it for once and then use.

nabobalis: But it is used one?

sunpy/net/helioviewer.py
Naman9639: I will add some description

sunpy/net/tests/test_helioviewer.py
nabobalis: That is fine then. But why are these passed in as keywords?

sunpy/net/helioviewer.py
nabobalis: Well you use it once, so why not just call the original and not create a new variable?

sunpy/net/tests/test_helioviewer.py
Naman9639: You mean why is the key passed as  keyword?

nabobalis: Yes.

sunpy/net/helioviewer.py
```diff
@@ -57,6 +57,14 @@ def data_sources(self):
...
+        for k, v in dictionary.items():
+            if all(k1 == k2 or k2 is None  for k1, k2 in zip(k, key)):
+                yield v
```
nabobalis: Actually, this probably should be in a `util` file somewhere. It is too general to be a client function.

sunpy/net/tests/test_helioviewer.py
Naman9639: Actually I forgot to remove it. I was debugging the code so I added it. I will remove it

sunpy/net/helioviewer.py
Naman9639: So shall I remove this function from `helioviewer.py` and add it to `util`?

nabobalis: Yeah, not sure what file to add it to but I think it belongs there.

Naman9639: We can add it in `util.py` in the `util` folder

nabobalis: Yeah that would work. Also the another test would need to be added to the util folder that works on a generic dictionary.

Naman9639: Ok sure I will do that.

sunpy/net/helioviewer.py
Naman9639: Can I initialise the dictionary inside the init function? Then we can use it anywhere we need.

nabobalis: We already have tho, its called `self.data_sources`

Naman9639: Oh, looks like I got confused with something but now it's clear. Thanks

changelog/2926.trivial.rst
Naman9639: Example showing that we can access the source ID by just using two keywords?

nabobalis: Yeah, something like that. The changelog to explain this change better since it's unclear to me right now.

Naman9639:
Are these changes ok?

sunpy/net/helioviewer.py
```diff
@@ -171,13 +176,18 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
                 key = (observatory, instrument, detector, measurement)
-                source_id = self.data_sources[key]
+                source_id_list = list(partial_match(key, self.data_sources))
+                if len(source_id_list) == 1:
+                    source_id = source_id_list[0]
+                else:
+                    source_id = self.data_sources[key]
             except KeyError:
                 raise KeyError("The values used for observatory, instrument, detector, measurement "
                                "do not correspond to a source_id. Please check the list using "
```
nabobalis: I think since we repeat this entire block of code in two functions, we should move this out into a private function. This way we only need to modify it once.

sunpy/util/util.py
```diff
@@ -212,3 +212,28 @@ def expand_list_generator(inp):
                 yield nested_item
         else:
             yield item
+
+def partial_match(key, dictionary):
+    """
+    Return value/values based on the partial key .
```
nabobalis: This needs a bit more explanation. Also the space between `key` and `.` should be removed.

sunpy/util/util.py
```diff
@@ -212,3 +212,28 @@ def expand_list_generator(inp):
...
+
+    Parameters
+    ----------
+    key : `tuple`
+          A tuple containig the partial key.
```
nabobalis: `containig` -> `containing`

sunpy/util/util.py
```diff
@@ -212,3 +212,28 @@ def expand_list_generator(inp):
...
+    dictionary: `dict`
+          A dictionary having all the (key, value) pair which you want to search
```
nabobalis: This needs rephrasing.

sunpy/util/util.py
```diff
@@ -212,3 +212,28 @@ def expand_list_generator(inp):
...
+
+    Returns
+    -------
+    `list`
+        A list consisting of all the values matching the partial key.
+
+    References
+    ----------
+    Taken from :https://stackoverflow.com/questions/18893624/partial-match-dictionary-keyof-tuples-in-python
```
nabobalis: `from :h` -> `from: h`

changelog/2926.trivial.rst
```diff
@@ -0,0 +1,8 @@
+User can access the source ID of HelioviewerClient using fewer keywords. Either by `observatory` and `measurement` or `instrument` and `measurement` as this much information is enough to get the source ID for most of the cases.
+
+For example if we want to use `sunpy.net.helioviewer.HelioviewerClient.download_jp2` function we can use it in the following ways:
+1. filepath = hv.download_jp2('2012/07/03 14:30:00', observatory='SDO', instrument='HMI', detector=None, measurement='continuum')
+2. filepath = hv.download_jp2('2012/07/03 14:30:00', observatory='SDO', measurement='continuum')
+3. filepath = hv.download_jp2('2012/07/03 14:30:00', instrument='HMI', measurement='continuum')
+
+All the above three ways will give the same result. An error message is thrown if we get more than one value for a partial key.
```
nabobalis: The rendered version looks like your missing whitespace: 

screenshot from 2019-02-08 12-22-40

Naman9639: But I gave proper spaces while writing that. Do I need to build the docs to see how it looks after rendering?

nabobalis: Yes or have a look at the giles link in the checks section.

Naman9639: Ok

sunpy/util/util.py
```diff
@@ -212,3 +212,35 @@ def expand_list_generator(inp):
...
+def partial_match(key, dictionary):
+    """
+    Return value/values based on the partial key. Each element of the partial key is matched against the keys of the
+    dictionary and if a match is found the value of the key is returned. Even a partial match works here i.e 
+    even if the key matches partially a value is returned. Note that this function uses a generator so we need to 
+    type cast the output to a list to get the values.
+    For eg: If we have a dictionary like
+    `d = {('abc','def','ghi') : 1, ('abc', 'def', 'xyz') : 2, ('pqr', 'lmn', 'tuv') : 3}` then a command
+    `key = list(partial_match(('abc', 'def', None), d))` will return a list containing the values `[1,2]` as the given
+    key partially matches with two keys i.e ('abc','def','ghi') and ('abc','def','xyz') in the dictionary.
```
nabobalis: This should be in an example section and not here.

Naman9639: Can we add description in the example section? I mean can we describe how things work in the example section?

nabobalis: I mean you can describe it here and have this code actually run in an example section.

sunpy/net/helioviewer.py
```diff
@@ -171,25 +166,79 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+        >>> from sunpy.net import helioviewer
+        >>> hv = helioviewer.HelioviewerClient()  # doctest: +REMOTE_DATA
+        >>> header = hv.get_jp2_header('2012/07/03', observatory='SDO',
+        ...                            instrument='HMI', detector=None, measurement='continuum')  # doctest: +REMOTE_DATA
+        >>> # The key 'fits' can be used to get the fits header information
+        >>> fits_header = header['fits']  # doctest: +REMOTE_DATA
+        >>> # The keys 'helioviewer' can be used to extract the helioviewer specific metadata.
+        >>> helioviewer_meta_data = header['helioviewer']  # doctest: +REMOTE_DATA
+        """
+        if jp2_id is None:
```
nabobalis: What happened here?

sunpy/net/helioviewer.py
```diff
@@ -450,3 +499,17 @@ def _request(self, params):
...
         return parse_time(date).isot + "Z"
+
+    def _get_source_id(self, source_id, key):
+        if source_id is None:
+            try:
+                source_id_list = list(partial_match(key,self.data_sources))
+                if len(source_id_list) == 1:
+                    source_id = source_id_list[0]
+                else:
+                    source_id = self.data_sources[key]
```
nabobalis: Do we need to do this anymore?

sunpy/net/helioviewer.py
```diff
@@ -450,3 +499,17 @@ def _request(self, params):
     def _format_date(self, date):
         """Formats a date for Helioviewer API requests"""
         return parse_time(date).isot + "Z"
+
+    def _get_source_id(self, source_id, key):
+        if source_id is None:
+            try:
+                source_id_list = list(partial_match(key,self.data_sources))
+                if len(source_id_list) == 1:
```
nabobalis: I think if a partial match is found that is >1, we should add that to the `raise` so the user knows what was found.

sunpy/net/helioviewer.py
Naman9639: Sorry I didn't get that

sunpy/net/helioviewer.py
Naman9639: We can replace it with the list of values returned by the function

sunpy/net/helioviewer.py
nabobalis: Why is this in the github difference?

Naman9639: I have no idea. I have not changed anything in that function

sunpy/net/helioviewer.py
Naman9639: Shall we raise a KeyError or a ValueError saying that `Got more values than expected` and then just showing the list of all the values returned by the function?

nabobalis: I would say its still a KeyError but we can change it if we need to later.

sunpy/net/helioviewer.py
nabobalis: Maybe a rebase would help.

Naman9639:
What is this test `codecov/patch` for? And why is it failing?

nabobalis:
It checks for line coverage of our code through our test suite. It failed because it seems that this pull request lowers the amount of line coverage. 

Sometime this happens due to fails in the CI or the git history becomes tangled instead of an actual issue.

changelog/2926.trivial.rst
```diff
@@ -0,0 +1,9 @@
+User can access the source ID of HelioviewerClient using fewer keywords. Either by `observatory` and `measurement` or `instrument` and `measurement` as this much information is enough to get the source ID for most of the cases.
```
nabobalis: I think this line need a tidy up.

```
Users can now use fewer keywords in our `~sunpy.net.HelioviewerClient` to access the available sources.
```

changelog/2926.trivial.rst
```diff
@@ -0,0 +1,9 @@
...
+
+For example if we want to use `sunpy.net.helioviewer.HelioviewerClient.download_jp2` function we can use it in the following ways:
```
nabobalis: Same here:

```
For example, if we want to use `~sunpy.net.helioviewer.HelioviewerClient.download_jp2` you can now do the  following:
```

changelog/2926.trivial.rst
```diff
@@ -0,0 +1,9 @@
...
+
+1. filepath = hv.download_jp2('2012/07/03 14:30:00', observatory='SDO', instrument='HMI', detector=None, measurement='continuum')
+2. filepath = hv.download_jp2('2012/07/03 14:30:00', observatory='SDO', measurement='continuum')
+3. filepath = hv.download_jp2('2012/07/03 14:30:00', instrument='HMI', measurement='continuum')
+
+All the above three ways will give the same result. An error message is thrown if we get more than one value for a partial key.
```
nabobalis:
```
These examples will now give the same result. An error message will be raised if an incomplete match is found.
```

sunpy/net/helioviewer.py
```diff
@@ -109,14 +110,8 @@ def get_closest_image(self, date, observatory=None, instrument=None,
...
-        if source_id is None:
-            try:
-                key = (observatory, instrument, detector, measurement)
-                source_id = self.data_sources[key]
-            except KeyError:
-                raise KeyError("The values used for observatory, instrument, detector, measurement "
-                               "do not correspond to a source_id. Please check the list using "
-                               "HelioviewerClient.data_sources.")
+        key = (observatory, instrument, detector, measurement)
+        source_id = self._get_source_id(source_id, key)
```
nabobalis: Can we not pass the above line into this function?

sunpy/net/helioviewer.py
```diff
@@ -171,25 +166,79 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
         >>> hv = helioviewer.HelioviewerClient()  # doctest: +REMOTE_DATA
         >>> filepath = hv.download_jp2('2012/07/03 14:30:00', observatory='SDO',
         ...                            instrument='HMI', detector=None, measurement='continuum')  # doctest: +REMOTE_DATA
+        >>> filepath = hv.download_jp2('2012/07/03 14:30:00', observatory='SDO', measurement='continuum')  # doctest: +REMOTE_DATA
         >>> aia = sunpy.map.Map(filepath)  # doctest: +REMOTE_DATA
         >>> aia.peek()  # doctest: +SKIP
         """
-        if source_id is None:
```
nabobalis: Do we not still need this check?

sunpy/net/helioviewer.py
```diff
@@ -171,25 +166,79 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+        --------
+        >>> from sunpy.net import helioviewer
+        >>> hv = helioviewer.HelioviewerClient()  # doctest: +REMOTE_DATA
+        >>> header = hv.get_jp2_header('2012/07/03', observatory='SDO',
+        ...                            instrument='HMI', detector=None, measurement='continuum')  # doctest: +REMOTE_DATA
+        >>> # The key 'fits' can be used to get the fits header information
+        >>> fits_header = header['fits']  # doctest: +REMOTE_DATA
+        >>> # The keys 'helioviewer' can be used to extract the helioviewer specific metadata.
+        >>> helioviewer_meta_data = header['helioviewer']  # doctest: +REMOTE_DATA
+        """
```
nabobalis: Since this is now in the difference, I think a rebase should be tried to remove this.

sunpy/net/helioviewer.py
```diff
@@ -171,25 +166,79 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
...
+        if jp2_id is None:
             try:
-                key = (observatory, instrument, detector, measurement)
-                source_id = self.data_sources[key]
+                jp2_id = self.get_closest_image(date, observatory, instrument, detector, measurement)['id']
             except KeyError:
                 raise KeyError("The values used for observatory, instrument, detector, measurement "
                                "do not correspond to a source_id. Please check the list using "
```
nabobalis: Do we need the `KeyError` here now?

sunpy/net/helioviewer.py
```diff
@@ -450,3 +499,18 @@ def _request(self, params):
...
+                source_id_list = list(partial_match(key,self.data_sources))
+                if len(source_id_list) == 1:
+                    source_id = source_id_list[0]
+                else:
+                    raise KeyError("The values used for observatory, instrument, detector, measurement "
+                                   "do not correspond to a source_id. Please check the list using "
+                                   "HelioviewerClient.data_sources. Expected 1 got " + str(len(source_id_list)) + 
+                                   " values for source_id ", source_id_list)
+            except KeyError:
+                raise
```
nabobalis: Why is this here?

sunpy/net/helioviewer.py
```diff
@@ -450,3 +499,18 @@ def _request(self, params):
...
+        if source_id is None:
+            try:
+                source_id_list = list(partial_match(key,self.data_sources))
+                if len(source_id_list) == 1:
+                    source_id = source_id_list[0]
+                else:
+                    raise KeyError("The values used for observatory, instrument, detector, measurement "
+                                   "do not correspond to a source_id. Please check the list using "
+                                   "HelioviewerClient.data_sources. Expected 1 got " + str(len(source_id_list)) + 
+                                   " values for source_id ", source_id_list)
```
nabobalis: Have you seen how the error message is printed out in a terminal?

sunpy/util/util.py
```diff
@@ -212,3 +212,37 @@ def expand_list_generator(inp):
...
+
+    References
+    ----------
+    Taken from: https://stackoverflow.com/questions/18893624/partial-match-dictionary-keyof-tuples-in-python
+
+    Examples
+    ----------
+    >>> d = {('abc','def','ghi') : 1, ('abc', 'def', 'xyz') : 2, ('pqr', 'lmn', 'tuv') : 3}
+    >>> key = list(partial_match(('abc', 'def', None), d))
+
```
nabobalis: You should unassign `key`, so it prints the results in the example.

sunpy/util/util.py
```diff
@@ -212,3 +212,37 @@ def expand_list_generator(inp):
...
         else:
             yield item
+
+def partial_match(key, dictionary):
+    """
+    Return value/values based on the partial key. Each element of the partial key is matched against the keys of the
+    dictionary and if a match is found the value of the key is returned. Even a partial match works here i.e 
+    even if the key matches partially a value is returned. Note that this function uses a generator so we need to 
+    type cast the output to a list to get the values. The example below will return a list containing the values `[1,2]` 
+    as the given key partially matches with two keys i.e ('abc','def','ghi') and ('abc','def','xyz') in the dictionary.
```
nabobalis: You should line break between the first sentence and the rest of the text here. 

I think the last line can be removed.

sunpy/net/helioviewer.py
Naman9639: We can do this but then I guess we should pass key as a keyword argument

sunpy/net/helioviewer.py
Naman9639: i did a rebase earlier as well

sunpy/net/helioviewer.py
Naman9639: Yes we need it because this function does not use source_id as a parameter but as a way to get jp2_id. But we need to check if the user has entered the correct source id

sunpy/net/helioviewer.py
Naman9639: I will upload a screenshot

Naman9639: error

sunpy/net/helioviewer.py
nabobalis: Why would we need to pass it in as a keyword?

sunpy/net/helioviewer.py
nabobalis: I can try myself if you want, but I do not see why this should be here.

sunpy/net/helioviewer.py
nabobalis: But the error now gets raised from inside `get_closest_image`?

sunpy/net/helioviewer.py
nabobalis: The error message doesn't make it clear to the user what keys were returned.

sunpy/net/helioviewer.py
Naman9639: I mean it will be better I guess to pass it as a keyword argument just for clear understanding

sunpy/net/helioviewer.py
Naman9639: Even I don't understand why they are here. I didn't touched that function.

sunpy/net/helioviewer.py
Naman9639: Oh, Ok I got it. Yes I will remove it

sunpy/net/helioviewer.py
Naman9639: Ok I will rephrase it

sunpy/net/helioviewer.py
nabobalis: I don't think so but I could be persuaded otherwise.

sunpy/net/helioviewer.py
nabobalis: I think you need to make sure your rebasing with upstream updated and might be worth squashing your commits into one so it clears up the history.

Naman9639: Ok, I will try it again.

sunpy/net/helioviewer.py
Naman9639: Ok, no problem. We won't keep it as a keyword argument

sunpy/net/helioviewer.py
Naman9639: I think we need this because if we don't check source_id for none then even if we pass in some value for source_id our function won't take that value and just try to calculate the source_id from the passed parameters like observatory, instrument etc which is wrong.

Naman9639:
Do I need to skip the result because most of the tests are failing as the result is getting printed

nabobalis:
Skip what?

sunpy/net/helioviewer.py
nabobalis: So you plan to add the check back?

sunpy/net/helioviewer.py
```diff
@@ -450,3 +433,18 @@ def _request(self, params):
     def _format_date(self, date):
         """Formats a date for Helioviewer API requests"""
         return parse_time(date).isot + "Z"
+
+    def _get_source_id(self, source_id, key):
+        """Returns source_id based on the key"""
```
nabobalis: Change this to:

```
"""
Returns source_id based on the key.
"""
```

sunpy/net/helioviewer.py
```diff
@@ -450,3 +433,18 @@ def _request(self, params):
...
+        if source_id is None:
```
nabobalis: Why does it check if its None?

sunpy/net/helioviewer.py
```diff
@@ -450,3 +433,18 @@ def _request(self, params):
     def _format_date(self, date):
         """Formats a date for Helioviewer API requests"""
         return parse_time(date).isot + "Z"
+
+    def _get_source_id(self, source_id, key):
```
nabobalis: Why is source_id an input?

sunpy/net/helioviewer.py
```diff
@@ -450,3 +433,18 @@ def _request(self, params):
...
+        if source_id is None:
+            source_id_list = list(partial_match(key,self.data_sources))
+            if len(source_id_list) == 1:
+                source_id = source_id_list[0]
+                return source_id
+            else:
+                raise KeyError("The values used for observatory, instrument, detector, measurement "
+                               "do not correspond to a source_id. Please check the list using "
+                               "HelioviewerClient.data_sources. Expected 1 got " + str(len(source_id_list)) + 
+                               " values for source_id ", source_id_list)
```
nabobalis: How does this error message look now?

sunpy/util/util.py
```diff
@@ -212,3 +212,36 @@ def expand_list_generator(inp):
...
+
+    References
+    ----------
+    Taken from: https://stackoverflow.com/questions/18893624/partial-match-dictionary-keyof-tuples-in-python
+
+    Examples
+    ----------
+    >>> d = {('abc','def','ghi') : 1, ('abc', 'def', 'xyz') : 2, ('pqr', 'lmn', 'tuv') : 3}
+    >>> list(partial_match(('abc', 'def', None), d))
+
```
nabobalis: You now need to add the output from this line here.

Naman9639:
> Skip what?

I mean all the tests are failing because of this
`Examples
>>> d = {('abc','def','ghi') : 1, ('abc', 'def', 'xyz') : 2, ('pqr', 'lmn', 'tuv') : 3}
>>> list(partial_match(('abc', 'def', None), d))
Expected nothing
Got:
    [1, 2] 
`

nabobalis:
It means you have to add it to the line below to make the example pass.

sunpy/net/helioviewer.py
Naman9639: We need to check this because if the user passes the source_id then we don't need to calculate it using our private function

sunpy/net/helioviewer.py
Naman9639: Because of the same reason that if a user enters the source_id which is allowed as of now then we don't want to calculate the source_id using _get_source_id function and just return the source_id entered by the user

sunpy/net/helioviewer.py
Naman9639: Ahh ok

sunpy/util/util.py
Naman9639: Ok I will add that

Naman9639:
Um, is the rebase ok now?

sunpy/net/helioviewer.py
Naman9639: It is same as before. It looks like that only. I tried it

sunpy/net/helioviewer.py
nabobalis: But this function should calculate a souirce_id, not check if we need one.

sunpy/net/helioviewer.py
nabobalis: But this is meant to get the source_id, I don't see why we should use this to check it.

sunpy/net/helioviewer.py
Naman9639: It do calculates the source_id but what if the user has passed it explicitly? Then it needs to return the same back to the function

sunpy/net/helioviewer.py
nabobalis: If I recall, it needed rephrasing.

nabobalis:
Seems to be. The github difference does not have the extra function now.

sunpy/net/helioviewer.py
nabobalis: But if the user has passed in one, we do not need to call this function.

sunpy/net/helioviewer.py
Naman9639: I'm confused. What should it tell the user. I mean shall I make it more clear that the values passed by the user resulted in the following source_id list and then just show the list?
Or shall it just be a simple error saying that the key is wrong

sunpy/net/helioviewer.py
Naman9639: Then we need a check in the function from which this private function is being called. In that case we need to add checks at two places so I thought to add it here

nabobalis: I don't see why we should call out to a function only to have it return the same value.

sunpy/net/helioviewer.py
nabobalis: The user needs to be able to understand what went wrong and how to fix it. 
Printing out some source_id numbers does not help a user.

nabobalis: So I think I have two issues with it as it stands. It is at one point really long and two does not give the user enough information to fix their issue.

sunpy/net/helioviewer.py
```diff
@@ -450,3 +433,18 @@ def _request(self, params):
...
         """Formats a date for Helioviewer API requests"""
         return parse_time(date).isot + "Z"
+
+    def _get_source_id(self, source_id, key):
+        """Returns source_id based on the key"""
+        if source_id is None:
+            source_id_list = list(partial_match(key,self.data_sources))
+            if len(source_id_list) == 1:
+                source_id = source_id_list[0]
+                return source_id
```
nabobalis: `return source_id_list[0]`

sunpy/net/helioviewer.py
nabobalis: So maybe we want to change it something like:

```
                raise KeyError(f"The values used: {observatory}, {instrument}, {detector}, {measurement} "
                                "do not correspond to one source_id but str(len(source_id_list)) source_id(s)."
                                " Please check the list using HelioviewerClient.data_sources.")
```

Do you know if it's possible to get all the keys that partially match?

nabobalis: Tho I don't think my version works since I have no accessed the right keys.

Naman9639: Yeah it return all the values that matches partially

sunpy/net/helioviewer.py
Naman9639: But we do need to check it somewhere. We can check it in the calling function

sunpy/net/helioviewer.py
Naman9639: But we do need to add the check somewhere. We can add it in the calling function

nabobalis: We can check it before we call the function.

sunpy/net/helioviewer.py
nabobalis: But that isn't the keys but the values?

Naman9639: Yes. It tries to get all the values that matches the partial key

sunpy/net/helioviewer.py
Naman9639: Ok no problem I'll add it there

sunpy/util/util.py
```diff
@@ -212,3 +212,37 @@ def expand_list_generator(inp):
...
         else:
             yield item
+
+def partial_match(key, dictionary):
+    """
+    Return value/values based on the partial key. Each element of the partial key is matched against the keys of the dictionary and if a match is found 
+    the value of the key is returned.
+    Even a partial match works here i.e even if the key matches partially
+    a value is returned. Note that this function uses a generator so we need to type cast the output to a list to get the values.
+
```
nabobalis: The doc string here. You want to have a new line at a full stop and re-flow the text after if its too long.

Naman9639: Um sorry I didn't get that. You mean I should have a new line after every full stop?

nabobalis: Yes.

sunpy/net/helioviewer.py
```diff
@@ -450,3 +435,15 @@ def _request(self, params):
...
+        """
+        Returns source_id based on the key.
+        """
+        source_id_list = list(partial_match(key,self.data_sources))
+        if len(source_id_list) == 1:
+            return source_id_list[0]
+        else:
+            raise KeyError("The values used: {observatory}, {instrument}, {detector}, {measurement} "
+                            "do not correspond to one source_id but str(len(source_id_list)) source_id(s)."
+                            " Please check the list using HelioviewerClient.data_sources.")
```
nabobalis: How does this error message look like now?

sunpy/util/util.py
```diff
@@ -212,3 +212,38 @@ def expand_list_generator(inp):
...
+          The target dictionary from which we want to retrieve the value based on the partial key.
+
+    Returns
+    -------
+    `list`
+        A list consisting of all the values matching the partial key.
+
+    References
+    ----------
+    Taken from: https://stackoverflow.com/questions/18893624/partial-match-dictionary-keyof-tuples-in-python
```
nabobalis: `Taken from:` -> `*`

sunpy/net/helioviewer.py
Naman9639: I will upload a screen shot

sunpy/util/util.py
Naman9639: Um what?

sunpy/net/helioviewer.py
Naman9639: error

sunpy/util/util.py
nabobalis: It will make it a bullet point.

sunpy/net/helioviewer.py
nabobalis: You need a `f`  in front of the string. It has not rendered correctly.

Naman9639: I did it with `f` but there was no change

nabobalis: You did `(f"`?

nabobalis: If that doesn't work, we will need to use `.format()` instead.

Naman9639: Oh ok I will go with format because the `f` didn't work for me

Naman9639: What exactly do you want it to render? Do we need the values of `observatory`, `instrument` etc to be shown in the error message?

Naman9639: error
Is this ok?

nabobalis: Yeah that looks like it works now.

Naman9639: it was not recognising observatory etc but when I replaced it with {key} it worked

Naman9639:
Are the servers down because so many tests failed and they are not related to helioviewer I guess?

nabobalis:
Have you checked to see why they errored?

Naman9639:
They are not related to helioviewer. I checked the log

nabobalis:
I have restarted them.

nabobalis:
The `pip` one is due to a change in master that would require this PR to be rebased. So we can ignore that.

The `windows` one seems to be an issue with `numpy` update in anaconda. We can ignore that too.

Naman9639:
Ah ok

> The `pip` one is due to a change in master that would require this PR to be rebased. So we can ignore that.
> 
> The `windows` one seems to be an issue with `numpy` update in anaconda. We can ignore that too.

Naman9639:
Um should I change anything?

nabobalis:
Nothing that I am aware of. Unfortunately, you will just have to wait until we get another review.

Naman9639:
I needed some help writing my proposal

nabobalis:
I would suggest raising that in the chat room instead of this PR. 
Just to remind that, that we are not aware that we have even been accepted by GSoC yet (February 26, 2019).

Naman9639:
Ok I will wait till 26th

sunpy/net/helioviewer.py
```diff
@@ -450,3 +435,15 @@ def _request(self, params):
...
+        """
+        Returns source_id based on the key.
+        """
+        source_id_list = list(partial_match(key,self.data_sources))
+        if len(source_id_list) == 1:
+            return source_id_list[0]
+        else:
+            raise KeyError(f"The values used: {key} do not correspond to one source_id "
+                            "but "+ str(len(source_id_list)) + " source_id(s)."
+                            " Please check the list using HelioviewerClient.data_sources.")
```
dpshelio: I would rewrite this in a different way...
```python
if len(source_id_list) > 1: # or maybe != 1
    raise KeyError...
return source_id_list[0]
```
I prefer this way because it's obvious that the error message is for the case when there's more than one possibility (or not 1). As it is now it could make one think that you can change this by a `try/except` block.

sunpy/net/helioviewer.py
```diff
@@ -450,3 +435,15 @@ def _request(self, params):
...
+    def _get_source_id(self, key):
+        """
+        Returns source_id based on the key.
+        """
+        source_id_list = list(partial_match(key,self.data_sources))
+        if len(source_id_list) == 1:
+            return source_id_list[0]
+        else:
+            raise KeyError(f"The values used: {key} do not correspond to one source_id "
+                            "but "+ str(len(source_id_list)) + " source_id(s)."
```
dpshelio:
```suggestion
                            f"but {len(source_id_list)} source_id(s)."
```
I believe this works....

sunpy/util/util.py
```diff
@@ -212,3 +212,38 @@ def expand_list_generator(inp):
...
+    the value of the key is returned.
+    Even a partial match works here i.e even if the key matches partially a value is returned.
+    Note that this function uses a generator so we need to type cast the output to a list to get the values.
+
+    Parameters
+    ----------
+    key : `tuple`
+          A tuple containing the partial key.
+    dictionary: `dict`
+          The target dictionary from which we want to retrieve the value based on the partial key.
```
dpshelio: To make it more general, could this second argument be just a list of tuples? I'm thinking that way could be used in the case of dictionaries like:
```python
partial_match(('a','b', None), d.keys())
```
Or maybe... if we want this to be dictionaries, then `partial_key_match` makes more sense for the function name. So with the `key` in the name I can think about it as the `key` of a dictionary.
But... since it return the values... It's not a matching function, but a `getting` one.

I don't know... too many questions...

sunpy/util/util.py
```diff
@@ -212,3 +212,38 @@ def expand_list_generator(inp):
...
         else:
             yield item
+
+def partial_match(key, dictionary):
+    """
+    Return value/values based on the partial key.
+    Each element of the partial key is matched against the keys of the dictionary and if a match is found 
+    the value of the key is returned.
+    Even a partial match works here i.e even if the key matches partially a value is returned.
+    Note that this function uses a generator so we need to type cast the output to a list to get the values.
```
dpshelio: You get values via a generator too! That's only if you want to get them all at once.

sunpy/util/util.py
```diff
@@ -212,3 +212,38 @@ def expand_list_generator(inp):
...
+
+    Examples
+    ----------
+    >>> d = {('abc','def','ghi') : 1, ('abc', 'def', 'xyz') : 2, ('pqr', 'lmn', 'tuv') : 3}
+    >>> list(partial_match(('abc', 'def', None), d))
+        [1, 2]
+
+    """
+    for k, v in dictionary.items():
+        if all(k1 == k2 or k2 is None  for k1, k2 in zip(k, key)):
```
dpshelio: I suppose we don't need the case when there's a `None` in the input dictionary... Though that should be easy to add with a `k1 is None`...

sunpy/util/util.py
Naman9639: Here we need all the values so that is why I added that line. I will change it and make it say if `if you need to get all the values`

sunpy/util/util.py
Naman9639: I have to check if the 2nd argument can be a list of tuples. 
I will change the name if I am not able to match it in tuples

sunpy/util/util.py
Naman9639: I guess we don't need that because here we are only concerned with returning the keys that match the partial key given by the user.

sunpy/util/util.py
dpshelio: Change the function is easy, as you only need to change the `dictionary.keys()` for that input. But, if the function is returning the values of these dictionaries, then yeah, update the function name.

sunpy/util/util.py
dpshelio: I don't think these notes should be here. I understand they can be useful for beginners, but there should be other place to put that. You don't get that kind of information when you look at the docstring of `range` or `map`. They say "Make an iterator...", then the person can look up how an `iterator` works. Here you could say, it returns a `generator`.

sunpy/util/util.py
Naman9639: Yes it is returning the values of the dictionary so I guess I will change the function name. It will be a general function to fetch values from a dictionary given a partial key

sunpy/net/helioviewer.py
```diff
@@ -450,3 +435,15 @@ def _request(self, params):
...
         return parse_time(date).isot + "Z"
+
+    def _get_source_id(self, key):
+        """
+        Returns source_id based on the key.
+        """
+        source_id_list = list(partial_key_match(key,self.data_sources))
+        if len(source_id_list) > 1:  # or maybe != 1
+            raise KeyError(f"The values used: {key} do not correspond to one source_id "
+                            "but" f" {len(source_id_list)} source_id(s)."
```
nabobalis: Does this work?

Naman9639: Yeah though it is a bit weird. I will send a screenshot
error

nabobalis:
```suggestion
                            f"but {len(source_id_list)} source_id(s)." 
```

changelog/2926.trivial.rst
```diff
@@ -0,0 +1,9 @@
+Users can now use fewer keywords in our `~sunpy.net.HelioviewerClient` to access the available sources. Either by `observatory` and `measurement` or `instrument` and `measurement` as this much information is enough to get the source ID for most of the cases.
+
+For example, if we want to use `~sunpy.net.helioviewer.HelioviewerClient.download_jp2` you can now do the  following:
+
+1. filepath = hv.download_jp2('2012/07/03 14:30:00', observatory='SDO', instrument='HMI', detector=None, measurement='continuum')
+2. filepath = hv.download_jp2('2012/07/03 14:30:00', observatory='SDO', measurement='continuum')
+3. filepath = hv.download_jp2('2012/07/03 14:30:00', instrument='HMI', measurement='continuum')
```
nabobalis: We had a discussion on this and in hindsight, this should be in the documentation string for the functions to tell people they can do this instead of the changelog.

Naman9639: oh ok I will do that

nabobalis: We need to add some comments to each doc string that uses sourceID, telling users that you can do this.

Naman9639: Like telling the user that in place of source ID we can use the partial key to retrieve the source ID?

nabobalis:
Yes.

changelog/2926.trivial.rst
nabobalis: Yes

Naman9639:
Sorry for the delay. I was having my exams. I will be active again

sunpy/net/helioviewer.py
```diff
@@ -169,19 +165,16 @@ def download_jp2(self, date, observatory=None, instrument=None, detector=None,
         >>> import sunpy.map
         >>> from sunpy.net import helioviewer
         >>> hv = helioviewer.HelioviewerClient()  # doctest: +REMOTE_DATA
+        >>> # We can just pass the observatory and measurement or instrument and measurement to get the value for source ID.
```
nabobalis: I would not put this in the example section. But above in the description.

sunpy/net/helioviewer.py
```diff
@@ -77,6 +78,8 @@ def get_closest_image(self, date, observatory=None, instrument=None,
                           detector=None, measurement=None, source_id=None):
         """
         Finds the closest image available for the specified source and date.
+        We can use `observatory` and `measurement` or `instrument` and `measurement` to get the value for source ID which
```
nabobalis: Actually in fact, I wonder if wrapping this in a rst .. note:: would be better.

Naman9639: Yeah I guess that would be better. I will change it

Cadair:
I got a little hasty with the merge button there, feel free to open another PR to tidy up the docs :smile:

sunpy/net/helioviewer.py
Naman9639: I need to build the docs to see if it is getting rendered properly right?

nabobalis: Yes.

Naman9639:
> I got a little hasty with the merge button there, feel free to open another PR to tidy up the docs

No issues. I will open another one

Issue #2886, opened 2019-01-05
Remove all warnings from our test suite (module by module)


nabobalis:
Currently, we have have silenced all the warnings of our test suite using within `setup.cfg`:

```
addopts = -p no:warnings
``` 

Removing this reveals a multitude of issues that need fixing. 
**Wile it might tedious and "easy", this will require a large amount of time.**

This issue involves several steps we suggest several items to be aware of.

* `addopts = -p no:warnings` is located in `setup.cfg`. We do not want this removed in any pull request. So you have to make sure that when you change the file locally, **you never commit it**. So make sure to do `git add <test file changed>` to avoid adding this change to any commit. 

* Since this involves the entire test suite, this is too much to do in one go. So do this file by file, a pull request is allowed to be small and file by file if need be. 

* Once you have removed the setting, you want to run `pytest --remote-data=any <file to run>`. This will only run the tests in that file. The tests are named `test_<filename>.py` and should be in folder called `tests` for each module. 

All this means that the work is dividable into lots of smaller parts that can be completed independently of each other.

lzeee:
There are two warnings in sunpy.physics
Both caused by the package pywt.  In its' _utils.py,
"using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated"

Rita-Okonkwo:
IS THIS STILL OPEN?

nabobalis:
@Rita-Okonkwo While many of the warnings are squashed, I do believe that some still remain.

Rita-Okonkwo:
can I work on this then?

nabobalis:
You are free to work on any issue, including this one. 

If you find that there are no warnings, we can close this issue as well.

dstansby:
🥳

PR #2927, opened 2019-02-05
Remove warnings when testing sunpy.cm

dstansby:
xref #2886. Yes, the actual change is in `sunpy.database` and not `sunpy.cm`.


nabobalis:
Seems to be a real import error:

`ImportError: cannot import name 'OrderedDict'`

PR #2928, opened 2019-02-05
Remove test warnings from sunpy.coordinates

dstansby:
xref #2886; mainly using `representation_type` instead of `representation`.


nabobalis:
Test fail is unrelated.

PR #2936, opened 2019-02-20
Changed Pandas DataFrames in sunpy/sun/models.py to Astropy Tables.

### LINKED ISSUE #2933: Convert sunpy/sun/models.py to using astropy tables
Currently importing `sunpy.sun.models` raises the following pandas warnings:
```python
import sunpy.sun.models
/Users/dstansby/github/sunpy/sunpy/sun/models.py:63: UserWarning: Pandas doesn't allow columns to be created via a new attribute name - see https://pandas.pydata.org/pandas-docs/stable/indexing.html#attribute-access
  'temperature': Quantity(1e6, 'K'), 'density': Quantity(1, 'g cm**-3')}
/Users/dstansby/github/sunpy/sunpy/sun/models.py:91: UserWarning: Pandas doesn't allow columns to be created via a new attribute name - see https://pandas.pydata.org/pandas-docs/stable/indexing.html#attribute-access
  'central temperature': Quantity(1e6, 'K'), 'time': Quantity(1e9, 'year')}
```
due to e.g. the following line of code:
```python
evolution.units = {'radius': con.radius, 'luminosity': con.luminosity,
                   'central temperature': Quantity(1e6, 'K'), 'time': Quantity(1e9, 'year')}
```
where `evolution` is a DataFrame. I'm guessing the way to fix this is changing `models.py` to using astropy tables instead of pandas DataFrames.

See documentation of Astropy tables: http://docs.astropy.org/en/stable/table/index.html.
We probably want to use QTable instead of a normal Table.

AbhigyanBose:
### Description

Changed Pandas DataFrames in sunpy/sun/models.py to Astropy Tables.

Fixes #2933

### DIFF
--- a/sunpy/sun/models.py
+++ b/sunpy/sun/models.py
@@ -17,7 +17,7 @@ Object
 
 """
 
-from astropy.units import Quantity
+from astropy import units as u
 from astropy.table import QTable
 import sunpy.sun.constants as con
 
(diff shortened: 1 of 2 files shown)


sunpy/sun/models.py
```diff
@@ -20,6 +20,7 @@
 
 import pandas
 from astropy.units import Quantity
+from astropy.table import Table
```
nabobalis: I was wondering if using `QTable` would be better.

nabobalis:
Could you print the new table and show us what it looks like?

sunpy/sun/models.py
AbhigyanBose: I'm new to this so I don't know. If you think QTable would be better then I could try to use it and we can see which one works better.

nabobalis: Yeah its just to try it. QTable is meant for tables with units as per my understanding.

AbhigyanBose:
I made another commit changing Tables to QTable and adding indices to the tables.

sunpy/sun/models.py
```diff
@@ -56,12 +57,14 @@
...
-interior.index.name = 'radius'
-interior.units = {'radius': con.radius, 'mass': con.mass, 'luminosity': con.luminosity,
+_d = {'radius': _radius, 'mass': _mass, 'luminosity': _luminosity, 'temperature': _temperature, 'density': _density}
+_units = {'radius': con.radius, 'mass': con.mass, 'luminosity': con.luminosity,
                   'temperature': Quantity(1e6, 'K'), 'density': Quantity(1, 'g cm**-3')}
+interior = QTable(_d)
 interior.source = 'Turck-Chieze et al. (1988)'
+for labels in _units:
+    interior[labels].unit = _units[labels]
+interior.add_index('radius')
```
nabobalis: Can you tell me what this does?

AbhigyanBose: Lines 65 and 66 add units to each of the columns. 
Line 67 makes the 'radius' column the index (since the radius was the index in the DataFrames as well).

The QTable does not print out any differently.

nabobalis: Ah ok, I was not aware that radius was the index. Thanks!

AbhigyanBose: I just saw that there is a way to create QTables directly from DataFrames. Should I use that instead of creating the QTables from scratch ??

nabobalis: No. What I would suggest to do is check that your changes are the same as if you had done that.

AbhigyanBose: My bad, doing that doesn't solve the initial issue as we will get the warning when we create the pandas DataFrame.

AbhigyanBose: Okay, will check that.

AbhigyanBose: I looked into it and conversion by using the method actually has some issues like it ignores the index column so in interior the final table ends up not having 'radius' column. 
I ran the tests and they are running successfully, please tell me if there are any changes that I need to make.

nabobalis: I think as it stands, you do not need to make any more changes from my point of view.

We will need to wait for others to have a look over this pull request and make comments.

sunpy/sun/models.py
```diff
@@ -9,9 +9,9 @@
 
 Object
 ------
-    interior : pandas.DataFrame
+    interior : astropy.table.QTable
```
nabobalis: Add a backtick on each side please.

sunpy/sun/models.py
```diff
@@ -9,9 +9,9 @@
...
         The standard model of the solar interior
-    evolution : pandas.DataFrame
+    evolution : astropy.table.QTable
```
nabobalis: Same as above.

sunpy/sun/models.py
```diff
@@ -20,6 +20,7 @@
 
 import pandas
```
nabobalis: Do we still need to import pandas?

AbhigyanBose: No we don't. My mistake, I overlooked it.

sunpy/sun/models.py
AbhigyanBose: Okay, will do. Also there are a few additional changes that I need to make in the documentation part, which I'll do and make a commit.

sunpy/sun/models.py
```diff
@@ -56,12 +57,14 @@
...
-interior = pandas.DataFrame(_d, index = _radius)
-interior.index.name = 'radius'
-interior.units = {'radius': con.radius, 'mass': con.mass, 'luminosity': con.luminosity,
+_d = {'radius': _radius, 'mass': _mass, 'luminosity': _luminosity, 'temperature': _temperature, 'density': _density}
+_units = {'radius': con.radius, 'mass': con.mass, 'luminosity': con.luminosity,
                   'temperature': Quantity(1e6, 'K'), 'density': Quantity(1, 'g cm**-3')}
+interior = QTable(_d)
 interior.source = 'Turck-Chieze et al. (1988)'
+for labels in _units:
+    interior[labels].unit = _units[labels]
```
dpshelio: I would suggest that instead of adding the units as a separate array, it would be better for the future if we add them in the data (so we can remove the comments that precedes them), so like:
```python
_density = [147.74, ...., 1.99e-7] * u.g/u.cm**3
```
So then, we don't need to loop around the units and it's easier to manipulate them.
This is how it's shown in [astropy's docs](http://docs.astropy.org/en/stable/table/mixin_columns.html#quantity-and-qtable)

AbhigyanBose: I understand. Will make that change.

AbhigyanBose: I don't think that will be practical. As trying to represent a unit like 10^6 K results in all values of the column being multiplied by 10^6 and the unit being K.
Example:

sunpy/sun/models.py
```diff
@@ -20,6 +20,7 @@
 
 import pandas
 from astropy.units import Quantity
+from astropy.table import QTable
 import sunpy.sun.constants as con
```
Cadair: can you add an `__all__` list to this file with just the models in it please?

AbhigyanBose: Made the changes, could you tell me if that is what you meant ?

sunpy/sun/models.py
```diff
@@ -31,65 +31,57 @@
 _radius = [0, 0.01, 0.022, 0.061, 0.090, 0.120,
           0.166, 0.202, 0.246, 0.281, 0.317,
           0.370, 0.453, 0.611, 0.7304, 0.862,
-          0.965, 1.0000]
+          0.965, 1.0000] * u.Rsun
```
AbhigyanBose: @dpshelio I made the changes that you suggested. Could you check them out once?

sunpy/sun/models.py
AbhigyanBose: Done

sunpy/sun/models.py
```diff
@@ -2,26 +2,27 @@
 Solar Physical Models
 ---------------------
 This module contains standard models of the sun from various sources. All data is saved
-in pandas DataFrames with two added attributes
+in Astropy QTables with an added attribute
```
nabobalis:
```suggestion
in a `~astropy.table.QTable` with an added attribute.
```

sunpy/sun/models.py
```diff
@@ -2,26 +2,27 @@
...
 
 * source : names the source of the data
-* units : a dictionary with the units of each of the columns
 
 Object
 ------
-    interior : pandas.DataFrame
+    **interior** : ``astropy.table.QTable``
```
nabobalis:
```suggestion
    interior : `astropy.table.QTable`
```

sunpy/sun/models.py
```diff
@@ -2,26 +2,27 @@
...
         The standard model of the solar interior
-    evolution : pandas.DataFrame
+    **evolution** : ``astropy.table.QTable``
```
nabobalis:
```suggestion
    evolution : `astropy.table.QTable`
```

nabobalis:
Can you rebase this on top of our current master branch version? 
Also can you  fix those pep8 issues reported from the bot?

sunpy/sun/models.py
AbhigyanBose: I didn't understand this properly before, my bad. Will make the changes.

nabobalis:
Thanks @AbhigyanBose

AbhigyanBose:
@nabobalis Welcome. I'm so happy, this was my first contribution to an Open Source project.

Issue #2932, opened 2019-02-12
astropy_helpers is not automatically being initialized by setup.py


dstansby:
This could well be related to #2920 

---

With the latest `master` branch checked out, running `python setup.py make_docs` gives me

```
$ python setup.py build_docs
The requested path 'astropy_helpers' for importing astropy_helpers does not exist, or does not contain a copy of the astropy_helpers package.
Downloading 'astropy-helpers'; run setup.py with the --offline option to force offline installation.
Traceback (most recent call last):
  File "setup.py", line 12, in <module>
    from astropy_helpers.setup_helpers import setup, register_commands # noqa
ImportError: cannot import name 'setup' from 'astropy_helpers.setup_helpers' (/Users/dstansby/miniconda3/lib/python3.7/site-packages/astropy_helpers/setup_helpers.py)
```

I've checked that `astropy_helpers` is installed properly using `pip install astropy_helpers --upgrade`.

nabobalis:
You might need to check that the submodule is pointing at master.

dstansby:
Possibly caused by https://github.com/sunpy/sunpy/pull/2920, judging by the `setup.py` blame.

dstansby:
Do I really have to have a submodule checked out to master in order to build (and therefore contribute to) the docs?

nabobalis:
Right now (probably) yes. That PR made some changes to setup.py that require the upstream master of helpers. 

Cadair is working on planning to remove the submodule in the future. But more upstream work needs to occur.

nabobalis:
In fact, I am not sure that anything works without the submodule. But I have never actually tried.

dstansby:
Are there instructions on how to do that in the SunPy docs somewhere? (I can't find them). It seems slightly crazy to depend on the master branch of another project, and expect other people to be able to keep up with that.

nabobalis:
Normally the submodule is checked out at a release version but it just happens to be set differently right now. When Astropy 3.2 comes out, we will update that.

The command should be in the developer documentation.

But the command should be `git submodule update --init`.

dstansby:
👍 that worked a treat, thanks! Maybe worth adding a note to that effect to https://docs.sunpy.org/en/latest/dev_guide/documentation.html#usage ?

nabobalis:
We should add it to more places, currently the message is only here: https://docs.sunpy.org/en/latest/dev_guide/newcomers.html#astropy-helpers

But I am in the process of redoing the entire dev docs.

[...]

PR #2938, opened 2019-02-23
Modified /sunpy/instr/goes.py so that 'hek' is imported only in the f…

### LINKED ISSUE #2821: Unexpected error on import of goes module
### Description

When importing the goes instr module to get two simple functions
```
from sunpy.instr.goes import flareclass_to_flux, flux_to_flareclass
```
I ran into an error because it was trying to import the `hek` module which requires a package I had not yet installed.

### Expected behavior

Given that the functions I wanted are simple and should not require the `hek`, I was not expecting to need to have a package installed to support the `hek` module.

### Actual behavior

Should have no error!

I would recommend the import of `hek` is moved into the only function that uses it which is `get_goes_event_list`.

Akram9:
…unction that uses it.

### Description
Solves the unexpected error as described in #2821 by moving `import hek` into `get_goes_event_list`.

Fixes #2821

### DIFF
--- a/sunpy/instr/goes.py
+++ b/sunpy/instr/goes.py
@@ -105,8 +105,7 @@ def get_goes_event_list(timerange, goes_class_filter=None):
         e.g. 'M1', 'X2'.
 
     """
-
-    # import hek here so that other functions don't import hek as they don't need it
+    # Importing hek here to avoid calling code that relies on optional dependencies.
     from sunpy.net import hek
 
     # use HEK module to search for GOES events
--- a/sunpy/instr/goes.py
+++ b/sunpy/instr/goes.py
@@ -59,7 +59,6 @@ from scipy import interpolate
 from scipy.integrate import trapz, cumtrapz
 
 from sunpy import sun
-from sunpy.net import hek
 from sunpy import timeseries
 from sunpy.time import parse_time
 from sunpy.util.net import check_download_file
@@ -106,6 +105,10 @@ def get_goes_event_list(timerange, goes_class_filter=None):
         e.g. 'M1', 'X2'.
 
     """
+
+    # import hek here so that other functions don't import hek as they don't need it
+    from sunpy.net import hek
+
     # use HEK module to search for GOES events
     client = hek.HEKClient()
     event_type = 'FL'


nabobalis:
You have also pulled in changes to `astropy_helpers`. You need to remove them. 

Since you have committed the change, the best way would be to run:

```
git reset HEAD~
git submodule update --init
git add .
git commit
```
That should commit the change to the file only (in theory).

sunpy/instr/goes.py
```diff
@@ -106,6 +105,10 @@ def get_goes_event_list(timerange, goes_class_filter=None):
         e.g. 'M1', 'X2'.
 
     """
+
+    #import hek here as this is the only function using it
```
nabobalis: This should be clearer, the reason this is here is not since its the only function using it per say, but to avoid importing and calling code that requires an optional dependency.

Akram9: Should I change and push again?

nabobalis: Yes please.

Akram9:
I am having some problem with `astropy_helpers`. I had missed your earlier comment on reverting the changes and now I am not able to push. What should I do?
Will deleting this PR and doing it again help?

PS: I don't get how the astropy changes occured.

nabobalis:
No, deleting it won't really help. How it happend, I am unsure since in theory it should have been checked out at the correct version. I wonder if this is the bug mentioned in #2932.

You won't be able to normally push if you change the history, you'd have to force push. 
You can do the same commands as above but with:

```
git reset HEAD~2
git submodule update --init
git status (check here that astropy_helpers is unchanged. 
git add .
git commit -m <message>
git push --force
```

Akram9:
I did as per your above code and it seems better now. Please check.

nabobalis:
Yes that looks much beter now.

sunpy/instr/goes.py
```diff
@@ -106,6 +105,10 @@ def get_goes_event_list(timerange, goes_class_filter=None):
         e.g. 'M1', 'X2'.
 
     """
+
+    # import hek here so that other functions don't import hek as they don't need it
+    from sunpy.net import hek
+
```
nabobalis: Remove the new line.

sunpy/instr/goes.py
```diff
@@ -106,6 +105,10 @@ def get_goes_event_list(timerange, goes_class_filter=None):
         e.g. 'M1', 'X2'.
 
     """
+
```
nabobalis: Remove the new line.

sunpy/instr/goes.py
Akram9: 111 or 108?

sunpy/instr/goes.py
```diff
@@ -106,6 +105,10 @@ def get_goes_event_list(timerange, goes_class_filter=None):
...
+    # import hek here so that other functions don't import hek as they don't need it
```
nabobalis: Change this to:
```
 # Importing hek here to avoid calling code that relies on optional dependencies. 
```

Akram9: Okay, doing it.....

sunpy/instr/goes.py
nabobalis: Both.

sunpy/instr/goes.py
Akram9: Do I push again, normally?

dpshelio: You should be able to...

Akram9:
So what do I do next?

nabobalis:
Nothing, this will be merged once the checks clear.

Akram9:
Thanks a lot!

nabobalis:
Thank you @Akram9.

Akram9:
Oh, thanks to you people. This was my first PR, and an enjoyable one at that.

PR #2930, opened 2019-02-07
increased tolerance on one test and fixed another flaky test

nabobalis:
### Description

Fixes my nightmares.

PR #2939, opened 2019-02-23
updated python version checking in test_rhessi

nabobalis:
See comment here: https://github.com/sunpy/sunpy/pull/2930#discussion_r259587254

### DIFF
--- a/sunpy/instr/tests/test_rhessi.py
+++ b/sunpy/instr/tests/test_rhessi.py
@@ -4,6 +4,7 @@ Unit tests for `sunpy.instr.rhessi`
 """
 import sys
 import textwrap
+from distutils.version import LooseVersion
 
 from unittest import mock
 import numpy as np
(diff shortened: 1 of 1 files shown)


nabobalis:
YOU CAND DRAFT PRS NOW. WHAT MAGIC

sunpy/instr/tests/test_rhessi.py
```diff
@@ -136,7 +137,7 @@ def test_parse_observing_summary_dbase_file_mock():
     """
 
     # We need to mock this test differently for < 3.7.2.
-    if sys.version_info[2] < 2 or sys.version_info[1] < 7:
+    if LooseVersion(sys.version) < LooseVersion("3.7.2"):
```
dpshelio: `sys.version` is the whole date and stuff from when it was built... but I think this works fine, anyway.

nabobalis: I did try passing in your one and it broke Loose and Strict.

PR #2946, opened 2019-02-28
Mapbase converts crln_obs to hgln_obs if it is found in the header

### LINKED ISSUE #2942: Mapbase should convert crln_obs to hgln_obs if it is found in the header
This specifically applies to HMI data which uses crln/t_obs rather than hgln/t

dodoextinct:
### Description

Fixes #2942

### DIFF
--- a/changelog/2946.bugfix.rst
+++ b/changelog/2946.bugfix.rst
@@ -1,1 +1,1 @@
-Mapbase now converts Carrington longitude(crln_obs) to Stonyhurst longitude(hgln_obs), if Carrington longitude(crln_obs) is found in the header.
+If Carrington longitude ("crln_obs") is found in the FITS header, `~sunpy.map.Map` converts this to the correct Heliographic longitude. 
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -612,18 +612,15 @@ class GenericMap(NDData):
 
         if heliographic_longitude is None:
             if self.meta.get('crln_obs', None) is not None:
-                heliographic_longitude = self.meta['crln_obs']*(u.deg) - get_sun_L0(self.date)
-            
+                heliographic_longitude = self.meta['crln_obs'] * u.deg - get_sun_L0(self.date)
             else:
-                #this parts gets executed when crln_obs is also not present in the header.
                 if self._default_heliographic_longitude is None:
-                    warnings.warn_explicit(
-                                "Missing metadata for heliographic longitude: "
-                                "assuming longitude of 0 degrees",
-                                Warning, __file__,
-                                inspect.currentframe().f_back.f_lineno)
+                    warnings.warn_explicit("Missing metadata for heliographic longitude: "
+                                           "assuming longitude of 0 degrees",
+                                           Warning, __file__,
+                                           inspect.currentframe().f_back.f_lineno)
                     self._default_heliographic_longitude = 0
-                    heliographic_longitude = self._default_heliographic_longitude
+                heliographic_longitude = self._default_heliographic_longitude
 
         if isinstance(heliographic_longitude, str):
             heliographic_longitude = float(heliographic_longitude)
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -611,14 +611,19 @@ class GenericMap(NDData):
         heliographic_longitude = self.meta.get('hgln_obs', None)
 
         if heliographic_longitude is None:
-            if self._default_heliographic_longitude is None:
-                warnings.warn_explicit(
-                    "Missing metadata for heliographic longitude: "
-                    "assuming longitude of 0 degrees",
-                    Warning, __file__,
-                    inspect.currentframe().f_back.f_lineno)
-                self._default_heliographic_longitude = 0
-            heliographic_longitude = self._default_heliographic_longitude
+            if self.meta.get('crln_obs', None) is not None:
+                heliographic_longitude = self.meta['crln_obs']*(u.deg) - get_sun_L0(self.date)
+            
+            else:
+                #this parts gets executed when crln_obs is also not present in the header.
+                if self._default_heliographic_longitude is None:
+                    warnings.warn_explicit(
+                                "Missing metadata for heliographic longitude: "
+                                "assuming longitude of 0 degrees",
+                                Warning, __file__,
+                                inspect.currentframe().f_back.f_lineno)
+                    self._default_heliographic_longitude = 0
+                    heliographic_longitude = self._default_heliographic_longitude
 
         if isinstance(heliographic_longitude, str):
             heliographic_longitude = float(heliographic_longitude)
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -3,7 +3,6 @@
 Test Generic Map
 """
 import os
-import glob
 import tempfile
 import warnings
 
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -196,7 +196,7 @@ def test_coordinate_frame(aia171_test_map):
     assert frame.obstime == aia171_test_map.date
     
 def test_HMI_file(genric_map):
-    assert generic_map.resampled_hmi.fits == 0
+    assert generic_map.resampled_hmi.fits == genric_map.carrington_longitude - sunpy.coordinates.get_sun_L0(generic_map.date)
 
 
 # ==============================================================================
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -196,7 +196,7 @@ def test_coordinate_frame(aia171_test_map):
     assert frame.obstime == aia171_test_map.date
     
 def test_HMI_file(genric_map):
-    assert generic_map.resampled_hmi.fits == genric_map.carrington_longitude - sunpy.coordinates.get_sun_L0(generic_map.date)
+    assert generic_map.sunpy.data.test.resampled_hmi.fits == genric_map.carrington_longitude - sunpy.coordinates.get_sun_L0(generic_map.date)
 
 
 # ==============================================================================
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -30,7 +30,7 @@ testpath = sunpy.data.test.rootdir
 
 @pytest.fixture
 def hmi_test_map():
-    return sunpy.map.Map(glob.glob(os.path.join(testpath, "resampled_hmi.fits")))
+    return sunpy.map.Map(os.path.join(testpath, "resampled_hmi.fits"))
 
 @pytest.fixture
 def aia171_test_map():
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -3,6 +3,7 @@
 Test Generic Map
 """
 import os
+import glob
 import tempfile
 import warnings
 
@@ -27,6 +28,9 @@ from sunpy.time import parse_time
 
 testpath = sunpy.data.test.rootdir
 
+@pytest.fixture
+def HMI_file():
+    return sunpy.map.Map(glob.glob(os.path.join(testpath, "resampled_hmi.fits")))
 
 @pytest.fixture
 def aia171_test_map():
@@ -195,9 +199,9 @@ def test_coordinate_frame(aia171_test_map):
     assert frame.observer.radius == aia171_test_map.observer_coordinate.frame.radius
     assert frame.obstime == aia171_test_map.date
     
-def test_HMI_file(genric_map):
-    assert generic_map.sunpy.data.test.resampled_hmi.fits == genric_map.carrington_longitude - sunpy.coordinates.get_sun_L0(generic_map.date)
-
+def test_heliographic_longitude_crln(HMI_file):
+    if HMI_file.heliographic_longitude == 0:
+        assert HMI_file.heliographic_longitude == HMI_file.carrington_longitude - sunpy.coordinates.get_sun_L0(HMI_file.date)
 
 # ==============================================================================
 # Test Rotation WCS conversion
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -29,7 +29,7 @@ from sunpy.time import parse_time
 testpath = sunpy.data.test.rootdir
 
 @pytest.fixture
-def HMI_file():
+def hmi_test_map():
     return sunpy.map.Map(glob.glob(os.path.join(testpath, "resampled_hmi.fits")))
 
 @pytest.fixture
@@ -199,9 +199,8 @@ def test_coordinate_frame(aia171_test_map):
     assert frame.observer.radius == aia171_test_map.observer_coordinate.frame.radius
     assert frame.obstime == aia171_test_map.date
     
-def test_heliographic_longitude_crln(HMI_file):
-    if HMI_file.heliographic_longitude == 0:
-        assert HMI_file.heliographic_longitude == HMI_file.carrington_longitude - sunpy.coordinates.get_sun_L0(HMI_file.date)
+def test_heliographic_longitude_crln(hmi_test_map):
+    assert hmi_test_map.heliographic_longitude == hmi_test_map.carrington_longitude - sunpy.coordinates.get_sun_L0(hmi_test_map.date)
 
 # ==============================================================================
 # Test Rotation WCS conversion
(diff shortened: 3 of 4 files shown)


sunpy/map/mapbase.py
```diff
@@ -611,14 +611,19 @@ def heliographic_longitude(self):
...
-            if self._default_heliographic_longitude is None:
-                warnings.warn_explicit(
-                    "Missing metadata for heliographic longitude: "
-                    "assuming longitude of 0 degrees",
-                    Warning, __file__,
-                    inspect.currentframe().f_back.f_lineno)
-                self._default_heliographic_longitude = 0
-            heliographic_longitude = self._default_heliographic_longitude
+            if self.meta.get('crln_obs', None) is not None:
+                heliographic_longitude = self.meta.get('crln_obs', None) - get_sun_L0(self.date)
```
Cadair: If we make it to this line we know the value exists, so we should access it like this:
```suggestion
                heliographic_longitude = self.meta['crln_obs'] - get_sun_L0(self.date)
```

dodoextinct:
> Thanks for the PR!
> 
> I am not sure why all the tests have failed, can you investigate?
> 
> Also, could you write a test in the [`test_genericmap`](https://github.com/sunpy/sunpy/blob/master/sunpy/map/tests/test_mapbase.py) file, that uses the [test HMI file](https://github.com/sunpy/sunpy/blob/master/sunpy/data/test/resampled_hmi.fits) we have (which uses `crln/t_obs`).

The error is when the code is performing subtraction,

> Can only apply 'subtract' function to dimensionless quantities when other argument is not a quantity (unless the latter is all zero/infinity/nan)

dodoextinct:
> Thanks for the PR!

> 
> I am not sure why all the tests have failed, can you investigate?
> 
> Also, could you write a test in the [`test_genericmap`](https://github.com/sunpy/sunpy/blob/master/sunpy/map/tests/test_mapbase.py) file, that uses the [test HMI file](https://github.com/sunpy/sunpy/blob/master/sunpy/data/test/resampled_hmi.fits) we have (which uses `crln/t_obs`).

I think the error is when performing subtraction it is saying,

> "The value must be a valid Python or Numpy numeric type."

Can you help?

nabobalis:
The error seems to imply that one variable is a `astropy.unit` and the other is not. Therefore, it can not be subtracted. I would see what type those two variables are and go from there.

dodoextinct:
> The error seems to imply that one variable is a `astropy.unit` and the other is not. Therefore, it can not be subtracted. I would see what type those two variables are and go from there.

What can I do to help?

nabobalis:
I would see what type those two variables within the subtraction are and go from there.

sunpy/map/mapbase.py
```diff
@@ -611,14 +611,19 @@ def heliographic_longitude(self):
...
-            if self._default_heliographic_longitude is None:
-                warnings.warn_explicit(
-                    "Missing metadata for heliographic longitude: "
-                    "assuming longitude of 0 degrees",
-                    Warning, __file__,
-                    inspect.currentframe().f_back.f_lineno)
-                self._default_heliographic_longitude = 0
-            heliographic_longitude = self._default_heliographic_longitude
+            if self.meta.get('crln_obs', None) is not None:
+                heliographic_longitude = float(self.meta['crln_obs']) - float(get_sun_L0(self.date))
```
nabobalis: What was your thought process here? 

I would suggest also running code locally, so you can debug it.

dodoextinct: It was showing a type  error, so I thought since both values are returning values in degree type, hence it's not able to subtract, so I converted it to float.

nabobalis: Unfortunately it doesn't seem to have worked. The tests are still failing. 

I would suggest picking a failing test and trying to run it with a `pdb` or `breakpoint` to find its cause.

Cadair: I would *guess* that you need to multiply `self.meta['crln_obs']` by `u.deg` to put it in degrees.

dodoextinct: > What was your thought process here?
> 
> I would suggest also running code locally, so you can debug it.

How can I do it locally?

nabobalis: You could write a short script that imports and runs `Map` on the HMI  sample file or you could use `pytest` to run the tests.

dodoextinct: > What was your thought process here?
> 
> I would suggest also running code locally, so you can debug it.

How can I do it locally?

dodoextinct: > You could write a short script that imports and runs `Map` on the HMI sample file or you could use `pytest` to run the tests.

Okay, I will do that.

dodoextinct:
> The error seems to imply that one variable is a `astropy.unit` and the other is not. Therefore, it can not be subtracted. I would see what type those two variables are and go from there.

dodoextinct:
> I would see what type those two variables within the subtraction are and go from there.

Did you see the type?

nabobalis:
Sorry, I do not follow.

dodoextinct:
> Sorry, I do not follow.

You were going to inspecting the types of the variables like you said?

sunpy/map/mapbase.py
dodoextinct: > You could write a short script that imports and runs `Map` on the HMI sample file or you could use `pytest` to run the tests.

I was writing the script for testing but this error came up while importing `sunpy.map`

> ModuleNotFoundError: No module named 'numpy.core._multiarray_umath'

nabobalis:
Ah sorry, I meant that you should have a look to see what's wrong. 

The numpy issue probably means it needs to be updated (if possible) or downgraded. Hopefully they could fix it.

sunpy/map/mapbase.py
dodoextinct: @nabobalis can you help me with this?

nabobalis: The numpy issue? I think you should check that all your installed packages are up to date.

Cadair:
@dodoextinct I think a quick poke around in this with the debugger should be sufficient to get this working. If you run `pytest sunpy/coordinates --pdb` you will be presented with a [python debugger](https://pymotw.com/2/pdb/#module-pdb) where the error occurs and you can inspect the variables and see what's going on.

Akram9:
Doesn't `get_sun_L0(self.date)` give `astropy.coordinates.angles.Longitude` type?
That doesn't seem to convert to float on applying `float()`.

Also, numpy issue can be solved by upgrading to v1.16 using pip.

dodoextinct:
> @dodoextinct I think a quick poke around in this with the debugger should be sufficient to get this working. If you run `pytest sunpy/coordinates --pdb` you will be presented with a [python debugger](https://pymotw.com/2/pdb/#module-pdb) where the error occurs and you can inspect the variables and see what's going on.

Okay, I can try that.

dodoextinct:
> Doesn't `get_sun_L0(self.date)` give `astropy.coordinates.angles.Longitude` type?
> That doesn't seem to convert to float on applying `float()`.
> 
> Also, numpy issue can be solved by upgrading to v1.16 using pip.

Exactly it doesn't get converted into float.

nabobalis:
> > Doesn't `get_sun_L0(self.date)` give `astropy.coordinates.angles.Longitude` type?
> > That doesn't seem to convert to float on applying `float()`.
> > Also, numpy issue can be solved by upgrading to v1.16 using pip.
> 
> Exactly it doesn't get converted into float.

Yes, in this case you want to keep them as astropy.units and fix the subtract another way. 
You can see the documentation here: http://docs.astropy.org/en/stable/units/

dodoextinct:
> > > Doesn't `get_sun_L0(self.date)` give `astropy.coordinates.angles.Longitude` type?
> > > That doesn't seem to convert to float on applying `float()`.
> > > Also, numpy issue can be solved by upgrading to v1.16 using pip.
> > 
> > 
> > Exactly it doesn't get converted into float.
> 
> Yes, in this case you want to keep them as astropy.units and fix the subtract another way.
> You can see the documentation here: http://docs.astropy.org/en/stable/units/

I have tried changing units and debugging it and I am afraid, I couldn't solve the problem. Can anybody help?

nabobalis:
Hey @dodoextinct, I had a look at the code and discovered the issue was something else. It seemed the indentation was the issue that broke the code. 

Regarding what is left to do on this PR, we need a changelog entry and a unittest for the new behaviour.

dodoextinct:
> Hey @dodoextinct, I had a look at the code and discovered the issue was something else. It seemed the indentation was the issue that broke the code.
> 
> Regarding what is left to do on this PR, we need a changelog entry and a unittest for the new behaviour.
> 
> I also rebased the PR since we have had quite some changes to master.
> So locally you will need to do:
> 
> ```
> git remote update
> git checkout master
> git branch -D patch-1
> git checkout -b origin/patch-1
> ```
> That should restore your local version to what is here.

Okay, I will do the following.

dodoextinct:
> Hey @dodoextinct, I had a look at the code and discovered the issue was something else. It seemed the indentation was the issue that broke the code.
> 
> Regarding what is left to do on this PR, we need a changelog entry and a unittest for the new behaviour.
> 
> I also rebased the PR since we have had quite some changes to master.
> So locally you will need to do:
> 
> ```
> git remote update
> git checkout master
> git branch -D patch-1
> git checkout -b origin/patch-1
> ```
> That should restore your local version to what is here.

@nabobalis Is this a feature or a bugfix, what should I write in changelog?

nabobalis:
It is a bugfix

dodoextinct:
> It is a bugfix

[changelog/2946.bugfix.rst](url)
Is it correct?

changelog/2946.bugfix.rst
```diff
@@ -0,0 +1 @@
+Mapbase now converts Carrington longitude(crln_obs) to Stonyhurst longitude(hgln_obs), if Carrington longitude(crln_obs) is found in the header.
```
nabobalis:
```suggestion
If Carrington longitude ("crln_obs") is found in the FITS header, `~sunpy.map.Map` converts this to the correct Heliographic longitude. 
```

nabobalis:
Just a unit test left now

dodoextinct:
> Looks good to go but just a unit test left now.

Do I have to make changes in this code for the test?

`def test_heliographic_longitude(generic_map):
    assert generic_map.heliographic_longitude == 0.
`
[https://github.com/sunpy/sunpy/blob/76c87eb450d3b2c1f8ec43d394d33ab507d42cf5/sunpy/map/tests/test_mapbase.py#L182](url)

dodoextinct:
> Looks good to go but just a unit test left now.

Do I have to make changes in the following code?

`def test_heliographic_longitude(generic_map):
assert generic_map.heliographic_longitude == 0.
`
I am a bit confused. Can you help?

nabobalis:
You will want to edit that file, but I don't think you want to edit a test, but create a new one.

nabobalis:
You can see cadair's comment at the top for what file to test as well.

dodoextinct:
> You can see cadair's comment at the top for what file to test as well.

I have written the test, but I am afraid, it must be wrong. I am terribly confused about it. Can you guide me?

nabobalis:
> > You can see cadair's comment at the top for what file to test as well.
> 
> I have written the test, but I am afraid, it must be wrong. I am terribly confused about it. Can you guide me?

I would suggest having a look at the other tests in the file to see how the other tests are written. 

But you essentially need to create a test that loads in the `https://github.com/sunpy/sunpy/blob/master/sunpy/data/test/resampled_hmi.fits` 
and passes it into the generic map function and checks that the correct change was made to the attribute, that your code changed. 

You should read up on [pytest](https://docs.pytest.org/en/latest/contents.html) as well to try and understand the framework. Also I found this [introduction](http://pythontesting.net/framework/pytest/pytest-introduction/#example).

sunpy/map/tests/test_mapbase.py
```diff
@@ -194,6 +194,9 @@ def test_coordinate_frame(aia171_test_map):
     assert frame.observer.lon == aia171_test_map.observer_coordinate.frame.lon
     assert frame.observer.radius == aia171_test_map.observer_coordinate.frame.radius
     assert frame.obstime == aia171_test_map.date
+    
+def test_HMI_file(genric_map):
+    assert generic_map.sunpy.data.test.resampled_hmi.fits == genric_map.carrington_longitude - sunpy.coordinates.get_sun_L0(generic_map.date)
```
nabobalis: I strongly recommend running any test code that you write. To do so, you need to `pytest sunpy/map/tests/test_mapbase.py`, and this will run the entire file. 

What you need to do for this test is:

1. Create a `pytest.fixture`for the hmi test file much like `aia171_test_map` at the top of the file. Here is the [hmi test file to see how to access the test file](https://github.com/sunpy/sunpy/blob/cd8618e961b4140f2d65cc955f48760098e068f9/sunpy/map/sources/tests/test_hmi_source.py).

2. You will will want to assert that the `hmi_map.heliographic_longitude` is the same as if you opened the hmi file, loaded the header/meta information and done the operation on the crln_obs yourself.

dodoextinct: > I strongly recommend running any test code that you write. To do so, you need to `pytest sunpy/map/tests/test_mapbase.py`, and this will run the entire file.
> 
> What you need to do for this test is:
> 
> 1. Create a `pytest.fixture`for the hmi test file much like `aia171_test_map` at the top of the file. Here is the [hmi test file to see how to access the test file](https://github.com/sunpy/sunpy/blob/cd8618e961b4140f2d65cc955f48760098e068f9/sunpy/map/sources/tests/test_hmi_source.py).
> 2. You will will want to assert that the `hmi_map.heliographic_longitude` is the same as if you opened the hmi file, loaded the header/meta information and done the operation on the crln_obs yourself.

I have run the tests locally and it passed, hope it's correct.

sunpy/map/tests/test_mapbase.py
```diff
@@ -194,7 +198,10 @@ def test_coordinate_frame(aia171_test_map):
     assert frame.observer.lon == aia171_test_map.observer_coordinate.frame.lon
     assert frame.observer.radius == aia171_test_map.observer_coordinate.frame.radius
     assert frame.obstime == aia171_test_map.date
-
+    
+def test_heliographic_longitude_crln(HMI_file):
+    if HMI_file.heliographic_longitude == 0:
```
nabobalis: Why the if statement?

sunpy/map/tests/test_mapbase.py
```diff
@@ -27,6 +28,9 @@
 
 testpath = sunpy.data.test.rootdir
 
+@pytest.fixture
+def HMI_file():
```
nabobalis: Can you rename this?

sunpy/map/tests/test_mapbase.py
dodoextinct: To check if heliographic_longitude is none in the HMI file. Shouldn't it check?

sunpy/map/tests/test_mapbase.py
dodoextinct: Rename HMI_file to hmi_file?

sunpy/map/tests/test_mapbase.py
nabobalis: Would it be zero?

sunpy/map/tests/test_mapbase.py
nabobalis: No, to something similar as the other pytest fixtures in the file.

sunpy/map/tests/test_mapbase.py
dodoextinct: `None` will be a better option?

sunpy/map/tests/test_mapbase.py
dodoextinct: Okay, `hmi_test_map`  will do?

sunpy/map/tests/test_mapbase.py
nabobalis: I mean I would remove it. I don't see why there should be an if statement.

dodoextinct: Okay, I will do that.

sunpy/map/tests/test_mapbase.py
```diff
@@ -27,6 +28,9 @@
 
 testpath = sunpy.data.test.rootdir
 
+@pytest.fixture
+def hmi_test_map():
+    return sunpy.map.Map(glob.glob(os.path.join(testpath, "resampled_hmi.fits")))
```
Cadair:
```suggestion
    return sunpy.map.Map(os.path.join(testpath, "resampled_hmi.fits"))
```

sunpy/map/tests/test_mapbase.py
```diff
@@ -3,6 +3,7 @@
 Test Generic Map
 """
 import os
+import glob
```
Cadair:
```suggestion
```

changelog/2946.bugfix.rst
nabobalis: Could this change be made? I think we can merge this after.

Cadair:
Thanks @dodoextinct

PR #2950, opened 2019-03-03
Fix most warnings in sunpy.image tests

dstansby:
xref #2886

The only warnings I ~~could~~ couldn't get rid of were
```
sunpy/image/tests/test_transform.py::test_nan_skimage_low
  /Users/dstansby/miniconda3/lib/python3.7/site-packages/numpy/core/_methods.py:32: RuntimeWarning: invalid value encountered in reduce
    return umr_minimum(a, axis, None, out, keepdims, initial)
  /Users/dstansby/miniconda3/lib/python3.7/site-packages/numpy/core/_methods.py:28: RuntimeWarning: invalid value encountered in reduce
    return umr_maximum(a, axis, None, out, keepdims, initial)

-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

### DIFF
--- a/sunpy/image/tests/test_transform.py
+++ b/sunpy/image/tests/test_transform.py
@@ -105,40 +105,41 @@ def test_scale(scale_factor):
     rmatrix = np.array([[1.0, 0.0], [0.0, 1.0]])
 
     # Check a scaled image against the expected outcome
-    newim = tf.rescale(original/original.max(), scale_factor, order=4,
-                       mode='constant') * original.max()
+    newim = tf.rescale(original / original.max(), scale_factor, order=4,
+                       mode='constant', multichannel=False, anti_aliasing=False) * original.max()
     # Old width and new center of image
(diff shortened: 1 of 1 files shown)


nabobalis:
Does `pytest.warns` not work for that error?

dstansby:
Probably, but I'm not sure whether this warning is meant to be emitted or not. (blindly just catching all the warnings without knowing whether they should be present seems like a bad idea!)

nabobalis:
Fair point. 

I would have to check but I assume the warning is raised in that test since we purposely feeding a nan in to make sure it is still there after we call the function.

dstansby:
Might be worth raising a warning on the SunPy level like the rest of the warnings then instead of propagating it down to numpy where the warning is currently thrown?

nabobalis:
Yeah, I'd agree with that.

nabobalis:
Thanks @dstansby

PR #2952, opened 2019-03-04
Update models.py

nitish705702:
tested so we can discover the changes.
: interior
: evolution
They should return the appropriate tables for each.

### DIFF
--- a/sunpy/sun/models.py
+++ b/sunpy/sun/models.py
@@ -57,7 +57,7 @@ _density = [147.74, 146.66, 142.73, 116.10, 93.35,
            0.042, 0.00361, 1.99e-7]
 
 _d = {'mass': _mass, 'luminosity': _luminosity, 'temperature': _temperature, 'density': _density}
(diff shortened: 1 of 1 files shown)


nabobalis:
Sorry again @nitish705702, but we already have an open pull request addressing this issue (https://github.com/sunpy/sunpy/pull/2936). This pull request removes pandas and replaces with a different object. 

I would suggest having a look at another issue to tackle! You can find them here: https://github.com/sunpy/sunpy/issues?q=is%3Aissue+is%3Aopen+label%3A%22Package+Novice%22

nabobalis:
It seems that there is still an issue with the sythanx:

```
  File "/root/project/.tox/py37-build_docs/lib/python3.7/site-packages/sunpy/sun/models.py", line 60
    interior = pandas.DataFrame((_d, index = _radius),(columns=['mass', 'luminosity', 'temperature','density' ,'radius']))
                                           ^
SyntaxError: invalid syntax
```

PR #2954, opened 2019-03-04
move to setuptools_scm

nabobalis:
Moved from astropy_helpers ot setuptools_scm for our version number.


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "wheel", "numpy"]
+requires = ["setuptools", "setuptools_scm", "wheel", "numpy"]
```
Cadair: We should pin numpy to the minimum version here: `numpy==1.13.0`

setup.cfg
```diff
@@ -9,7 +9,7 @@ license = BSD 2-Clause
 url = https://sunpy.org
 edit_on_github = True
 github_project = sunpy/sunpy
-version = 1.0.dev
+version = 1.0
```
Cadair: Delete this completely

setup.py
```diff
@@ -27,6 +27,7 @@
 ################################################################################
 # Programmatically generate some extras combos.
 ################################################################################
+setups = read_configuration("setup.cfg")['options']['setup_requires']
```
Cadair: We shouldn't need this, it should be auto-read from the config. (We only read in extras because we modify it)

setup.cfg
nabobalis: It has to be there in some form otherwise it seems to error.

setup.py
nabobalis: Got it.

pyproject.toml
nabobalis: Done.

setup.cfg
Cadair: It really shouldn't:

> Important: Ensure neither the [metadata] version option nor the [egg_info] section are defined, as these will interfere with setuptools_scm.

https://github.com/pypa/setuptools_scm/#setupcfg

setup.cfg
```diff
@@ -20,11 +20,12 @@ classifiers =
 python_requires = >=3.6
 packages = find:
 include_package_data = True
+setup_requires = setuptools_scm
 install_requires =
-  numpy>=1.11
+  numpy>=1.13
   scipy
-  matplotlib>=1.3
```
Cadair: Even though they are really old does it make sense to keep these here, just in case someone installs sunpy in to an ancient env?

nabobalis: Well for pandas, the numpy version probably limits some of them

I can add back the matplotib one

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "wheel", "numpy"]
+requires = ["setuptools", "setuptools_scm", "wheel", "numpy>=1.13.0"]
```
Cadair: This needs to be `==` we always need to build the C extension with the oldest version we support.

sunpy/__init__.py
```diff
@@ -31,15 +31,6 @@ class UnsupportedPythonError(Exception):
     import builtins
     builtins._SUNPY_SETUP_ = False
 
-try:
```
Cadair: We should put this back.

nabobalis: Woops.

nabobalis:
CircleCI fails should resolve when this is merged.

Cadair:
Needs a changelog for the numpy version bump.

sunpy/version.py
```diff
@@ -0,0 +1,5 @@
+from pkg_resources import get_distribution, DistributionNotFound
```
Cadair: This should probably have a comment stating
`# This file is for compatibility with astropy_helpers`

PR #2957, opened 2019-03-06
Fixed Most Warnings in sunpy.io 

Pac23:
With Reference to #2886 
Fixed clobber Deprecation Warning,
All other tests in io work fine,except for a blank Header warning which is to be solved by upstream.


nabobalis:
Doing a search of the term clobber: https://github.com/sunpy/sunpy/search?q=clobber&type=Code

It seems its in three places total.

Pac23:
> Doing a search of the term clobber: https://github.com/sunpy/sunpy/search?q=clobber&type=Code
> 
> It seems its in three places total.

I hadn't tested the other two originally,fixed them now.

nabobalis:
Thanks @Pac23

PR #2963, opened 2019-03-07
Bug in sunpy.time convert_time_str function

hayesla:
### Description
In sunpy.time.time.py the astropy.time.Time.strptime is called but this does not exist, it comes from an older version of sunpy when the time types were datetimes
 

Fixes #<Issue Number>

### DIFF
--- a/sunpy/time/time.py
+++ b/sunpy/time/time.py
@@ -224,6 +224,7 @@ def convert_time_str(time_string, **kwargs):
                 break
             if ts is None:
                 continue
+            #problem!!
             return Time.strptime(ts, time_format, **kwargs) + time_delta
         except ValueError:
             pass


nabobalis:
Hello @hayesla, do you have an example I can use to debug?

Cadair:
This should exist as long as you have astropy 3.1 installed?

hayesla:
Oh maybe that's the problem - updated my astropy and now seems to work

Was also just learning from @DanRyanIrish how to make pull request to SunPy

DanRyanIrish:
Even if this isn't merged it's a good learning experience.  Also shows the value of opening a PR sooner rather than later.

nabobalis:
Yeah, you want to see the older PRs that wafels or stevie have opened. 

Several times they commited sunpy into sunpy.

Cadair:
Jack holds the dubious honor of being the only person to cause us to reset the master branch 😜😜

PR #2966, opened 2019-03-08
Update database.rst

nitish705702:
### Description

Fixes #<Issue Number>

### DIFF
--- a/docs/guide/acquiring_data/database.rst
+++ b/docs/guide/acquiring_data/database.rst
@@ -1,9 +1,8 @@
 .. The tests in the module are very fragile to change in ordering of VSO
    results. In their current form they are very hard to maintain.
-****************************************
-****************************************
-..  _database_guide:
-****************************************
+
+.. _database_guide:
+
 --------------------------
 Using the database package
 --------------------------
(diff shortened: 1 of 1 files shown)


nabobalis:
Hi @nitish705702, can you explain your thought process behind these changes? 

Looking at them at a glance, I can see some changes that I do not think are restructuredtext syntax. Also I do not understand the choice to add the HelioViewer Changes into the database.rst file.  

For reference here is a guide on RST I still load up: http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html

docs/guide/acquiring_data/database.rst
```diff
@@ -55,11 +53,13 @@ can also be committed explicitly using the :meth:`Database.commit` method.
...
-
-2. Adding new entries
+________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________
+2. <Adding new entries>
 ---------------------
-Each entry in a database is an instance of the class
-:class:`tables.DatabaseEntry` with the following attributes:
+             Each entry in a database is an instance of the class
+             :class:tables.DatabaseEntry (
+                  https://docs.sunpy.org/en/latest/api/sunpy.database.tables.DatabaseEntry.html#sunpy.database.tables.DatabaseEntry
```
nitish705702: Adding url of the data base entry file

docs/guide/acquiring_data/database.rst
```diff
@@ -800,3 +791,196 @@ to 10 and therefore removes the 5 entries that been used least recently.
      18    2011-06-07 20:37:52 ...               19.5               19.5
      58    2011-06-07 06:33:29 ... 17.400000000000002 17.400000000000002
      59    2011-06-06 00:00:00 ...                N/A                N/A
+****************************************************************************************************************************************
```
nitish705702: Let's add a file to a new file, and the name of file is 'QueryingHelioviewer.org_with_sunPy'  .
which is helped for 'SunPy can be used to make several basic requests using the The Helioviewer.org API including generating a PNG screenshot and downloading a JPEG 2000 image.'  likely related problem.

docs/guide/acquiring_data/database.rst
nabobalis: So the code at line 60 here, or the old line 62, already (at least it should) embeds a link to that when the documentation is built. So we don't have to do it manually.

docs/guide/acquiring_data/database.rst
nabobalis: What related problem?

We already have all of this text in the helioviewer.rst file, so I am still confused as to its inclusion here.

nitish705702: here my mistake ,sorry.

nabobalis: That is no problem. If you do need any help or advice, you can ask!

PR #2967, opened 2019-03-08
Update CITATION.rst

nitish705702:
### Description

Fixes #<Issue Number>


nabobalis:
Sorry @nitish705702, I already have an open PR updating this file. 

If you do want to tackle an issue, I would suggest having a look at these: https://github.com/sunpy/sunpy/issues?q=is%3Aissue+is%3Aopen+label%3A%22Package+Novice%22 

Some are assigned (the top right), if so you should keep that issue.

PR #2975, opened 2019-03-10
Fixed #2777 and added aspect control in LineAnimator

### LINKED ISSUE #2777: LineAnimator display of small y-axis values
The aspect ratio of `LineAnimator` appears to be determined through absolute scaling of x and y values. For example if y ~ 10^-9 and x ~ 1, the y-axis is not visible. We need something like an `aspect` kwarg as in `plt.imshow` so that the length of each axis is determined by the plotting area and not the data range. This issue probably applies to the other animated classes.

Akram9:
### Description

Fixes #2777 
Changed y-axis pad size in `base.py` to fixed size and added `aspect` option in `LineAnimator`.

### DIFF
--- a/sunpy/visualization/animator/base.py
+++ b/sunpy/visualization/animator/base.py
@@ -243,22 +243,22 @@ class BaseFuncAnimator:
         self.divider = make_axes_locatable(self.axes)
         pad = 0.01  # Padding between axes
         pad_size = Size.Fraction(pad, Size.AxesX(self.axes))
-        large_pad_size = Size.Fraction(0.1, Size.Fixed(3.1))
+        large_pad_size = Size.Fraction(0.1, Size.AxesY(self.axes))
 
         # Define size of useful axes cells, 50% each in x 20% for buttons in y.
         small_x = Size.Fraction((1.-2.*pad)/10, Size.AxesX(self.axes))
-        ysize = Size.Fraction((1.-2.*pad)/15., Size.Fixed(3.1))
+        ysize = Size.Fraction((1.-2.*pad)/15., Size.AxesY(self.axes))
 
         # Set up grid, 3x3 with cells for padding.
         if self.num_buttons > 0:
             xsize = Size.Fraction((1.-2.*pad)/self.num_buttons, Size.AxesX(self.axes))
             horiz = [xsize] + [pad_size, xsize]*(self.num_buttons-1) + \
-                    [Size.Fraction(0.1, Size.Fixed(3.1)), small_x]
+                    [Size.Fraction(0.1, Size.AxesY(self.axes)), small_x]
             vert = [ysize, pad_size] * self.num_sliders + \
-                   [large_pad_size, large_pad_size, Size.Fixed(3.1)]
+                   [largr_pad_size, large_pad_size, Size.AxesY(self.axes)]
         else:
-            vert = [ysize, pad_size] * self.num_sliders + \
-                   [large_pad_size, Size.Fixed(3.1)]
+            vert = [ysize, large_pad_size] * self.num_sliders + \
+                   [Size.AxesY(self.axes)]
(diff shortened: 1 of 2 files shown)


nabobalis:
What do the new plots look like now?

Akram9:
new0
new2
new1

They don't fluctuate in y-axis height now.

nabobalis:
What was it like before the fix?

Akram9:
aspect_0 25-3
aspect_0 25-2
aspect_8-2
aspect_0 25-5

Akram9:
The y-axis padding height was based on the input data. I changed that to fixed size padding.

sunpy/visualization/animator/line.py
```diff
@@ -77,7 +78,7 @@ class LineAnimator(ArrayAnimator):
           In this scenario the same x-axis values are used in every frame of the animation.
        b) Second, the x-axis array can have the same shape as the data array.
           In this scenario the x-axis is refreshed for each frame. For example, if
-          ``data.shape == axis_ranges[plot_axis_index] == (4, 3)``,
+          ``data.shape == axis_ranges[plot_axis_index] == (4, 3)``,     #shouldn't it be axis_ranges[plot_axis_index].shape ?
```
nabobalis: Can you check this?

nabobalis:
Looks much better than before.

sunpy/visualization/animator/line.py
Akram9: Yeah, I was going to say, I forgot the comment there.

Akram9: If you don't mind, I'll check this about 14 hours later.? I'm a bit occupied now.

nabobalis: Sure, there is no rush.

nabobalis:
I wonder if a new figure test is what we want here.

Akram9:
I will see what I can do here.

sunpy/visualization/animator/line.py
Akram9: Yes, it should be as I expect it.

Akram9:
I am new to tests. Should it be a unit test comparing data shapes?

Akram9:
Okay, I got a problem here. Changing the window size destroys the look as the pad size is fixed. I am trying a fix.

nabobalis:
I am not sure how best to test this behavior, if it can be done at all. Maybe the axes size of a plot? Maybe a figure test?

Akram9:
Yes, axes size of the plot to figure size is what I had in mind.

Cadair:
If I had to guess looking at your diff I would think that you don't need the changes in `base.py` for this to work?

Akram9:
I had thought the same and then I made changes in `base` which sort of solved the problem so I had gone ahead with that. 
Anyway, I will keep it unchanged and try only from `line`.

Akram9:
Applied a fix which fixes the aforementioned issues but I have only tried the example as in the docs. I will say it is yet not complete.

sunpy/visualization/animator/base.py
```diff
@@ -285,7 +285,7 @@ def _add_widgets(self):
             x = i*2
             # The i+1/10. is a bug that if you make two axes directly on top of
             # one another then the divider doesn't work.
-            self.buttons.append(self.fig.add_axes((0., 0., 0.+i/10., 1.)))
+            self.buttons.append(self.fig.add_axs((0., 0., 0.+i/10., 1.)))
```
DanRyanIrish: Is this a typo of a correct change?

DanRyanIrish:
I think this PR now needs a ```git pull upstream master```.

sunpy/visualization/animator/base.py
Akram9: Oops, a typo.

PR #2976, opened 2019-03-10
Fixed docstring and improved readability of hec.py

cquark7:
### Description

While working on an open issue I encountered some problems in [hec.py](https://github.com/sunpy/sunpy/blob/master/sunpy/net/helio/hec.py). To fix those problem I am opening this PR.
- Fixed incorrect return type in the docstring:
In `make_table_list` method return type was incorrectly mentioned as `str`, it should be `bytes`

- Handled edge cases:
Existing code will throwing `IndexError` if user input is '0' and other invalid strings.

- Improved code readability

### DIFF
--- a/sunpy/net/helio/hec.py
+++ b/sunpy/net/helio/hec.py
@@ -85,7 +85,7 @@ class HECClient(object):
         end_time : str, `~sunpy.time.parse_time` parsable objects
             The time where the query window closes
 
-        table : str
+        table : bytes
             The table to query from. If the table is unknown, the user will be
             prompted to pick from a list of tables.
 
@@ -107,7 +107,7 @@ class HECClient(object):
 
         """
         while table is None:
-            table = self.make_table_list()
+            table = self.select_table()
         start_time = parse_time(start_time)
         end_time = parse_time(end_time)
         results = self.hec_client.service.TimeQuery(STARTTIME=start_time.isot,
@@ -145,7 +145,7 @@ class HECClient(object):
         tables = votable_handler(etree.tostring(results))
         return tables.array
 
-    def make_table_list(self):
+    def select_table(self):
         """
         Creates a list of table names and prompts the user for a choice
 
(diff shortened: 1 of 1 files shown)


sunpy/net/helio/hec.py
```diff
@@ -156,39 +156,29 @@ def make_table_list(self):
 
         Returns
         -------
-        temp: `str`
+        `bytes`
```
nabobalis: Should this be bytes and not a string?

sunpy/net/helio/hec.py
```diff
@@ -156,39 +156,29 @@ def make_table_list(self):
...
+            user_input = input(f"\nPlease enter a table number between 1 and {len(table_list)} "
+                               "('e' to exit): ")
+            if user_input.lower() == "e" or user_input.lower() == "exit":
+                return None
+            if user_input.isdigit() and 1 <= int(user_input) <= len(table_list):
+                table_no = int(user_input)
+                return table_list[table_no - 1]
             else:
-                print("Choice outside of bounds")
-        return temp
```
nabobalis: We don't want to return anything anymore?

sunpy/net/helio/hec.py
cquark7: Yes.  Method `make_table_list` returns a `bytes` object, not `str`.

cquark7: @nabobalis I want to mention that this service will work if `table` is either string or bytes object. But it would be best to use bytes because that how client and TimeQuery service parses them. 
```python3
TimeQuery(STARTTIME=start_time.isot, ENDTIME=end_time.isot, FROM=table, MAXRECORDS=max_records)
```

nabobalis: Yeah that is fine.

sunpy/net/helio/hec.py
```diff
@@ -156,39 +156,29 @@ def make_table_list(self):
...
-                             "{elem:d} "
-                             "('e' to exit): ".format(elem=len(table_list)))
-            if stdinput.lower() == "e" or stdinput.lower() == "exit":
-                temp = None
-                break
-            temp = [int(s) for s in stdinput.split() if s.isdigit()]
-            temp = temp[0] - 1
-            if temp in range(0, len(table_list)):
-                temp = table_list[temp]
-                break
```
nabobalis: Do we not still need the break?

cquark7: We don't need them because return statements not only break out of the while loop, they also come out of function definition. They transfer control back to the statement which invoked the function.

nabobalis: Fair enough.

nabobalis:
Thanks @cquark7

PR #2982, opened 2019-03-12
Fixes issue-2981 by raising a ConnectionError

### LINKED ISSUE #2981: VSOClient should fail to initilize if it can't read the WSDL
Currently what happens is this:

```
/home/stuart/Git/sunpy/sunpy/net/vso/vso.py:93: UserWarning: Connection failed with error <urlopen error [Errno -3] Temporary failure in name resolution>. 
 Retrying with different url and port.
  "Connection failed with error {}. \n Retrying with different url and port.".format(e))
  File "test_download.py", line 14, in <module>
    a.Wavelength(171*u.angstrom) | a.Wavelength(94*u.angstrom))
  File "/home/stuart/Git/sunpy/sunpy/net/fido_factory.py", line 297, in search
    return UnifiedResponse(query_walker.create(query, self))
  File "/home/stuart/Git/sunpy/sunpy/net/attr.py", line 411, in create
    return self.createmm(self, *args, **kwargs)
  File "/home/stuart/Git/sunpy/sunpy/util/multimethod.py", line 120, in __call__
    return fun(*args, **kwargs)
  File "/home/stuart/Git/sunpy/sunpy/net/fido_factory.py", line 238, in _create_or
    qblocks.extend(walker.create(attr.and_(attrblock), factory))
  File "/home/stuart/Git/sunpy/sunpy/net/attr.py", line 411, in create
    return self.createmm(self, *args, **kwargs)
  File "/home/stuart/Git/sunpy/sunpy/util/multimethod.py", line 120, in __call__
    return fun(*args, **kwargs)
  File "/home/stuart/Git/sunpy/sunpy/net/fido_factory.py", line 231, in _create_and
    return [factory._make_query_to_client(*query.attrs)]
  File "/home/stuart/Git/sunpy/sunpy/net/fido_factory.py", line 431, in _make_query_to_client
    return tmpclient.search(*query), tmpclient
  File "/home/stuart/Git/sunpy/sunpy/net/vso/vso.py", line 303, in search
    QueryRequest = self.api.get_type('VSO:QueryRequest')
AttributeError: 'NoneType' object has no attribute 'get_type'


```

cquark7:
### Description

Fixes #2981 

Fixes the issue by checking if `get_online_vso_url` function returns a valid API.

### DIFF
--- a/docs/guide/acquiring_data/database.rst
+++ b/docs/guide/acquiring_data/database.rst
@@ -258,7 +258,7 @@ sections.
     >>> client = vso.VSOClient()  # doctest: +REMOTE_DATA
 
 .. testsetup:: *
-   >>> from unittest.mock import Mock
+   >>> from unittest.mock import Mock, patch
    >>> from sunpy.tests.mocks import MockObject
    >>> class MockQR:
    ...    def __init__(self, start_time, instrument, wavelength):
@@ -266,8 +266,9 @@ sections.
    ...        self.time = MockObject(start=start_time, end=end_time)
    ...        self.instrument = instrument
    ...        self.wave = MockObject(wavemin=wavelength, wavemax=wavelength, waveunit='nm')
-   >>> vso.VSOClient().search = Mock(return_value=[MockQR(f"2011050800000{t}", 'AIA', w) for
-   ...                                   t,w in zip([0,0,2,3], [17.1, 21.1, 9.4, 33.5])]) # doctest: +REMOTE_DATA
+   >>> with patch.object(vso.VSOClient, '__init__', lambda x: None):
+   ...    vso.VSOClient().search = Mock(return_value=[MockQR(f"2011050800000{t}", 'AIA', w) for
+   ...                                   t, w in zip([0,0,2,3], [17.1, 21.1, 9.4, 33.5])])
 
 After initialising the VSO client:
 
--- a/docs/guide/acquiring_data/database.rst
+++ b/docs/guide/acquiring_data/database.rst
@@ -267,7 +267,7 @@ sections.
    ...        self.instrument = instrument
    ...        self.wave = MockObject(wavemin=wavelength, wavemax=wavelength, waveunit='nm')
    >>> vso.VSOClient().search = Mock(return_value=[MockQR(f"2011050800000{t}", 'AIA', w) for
-   ...                                   t,w in zip([0,0,2,3], [17.1, 21.1, 9.4, 33.5])])
+   ...                                   t,w in zip([0,0,2,3], [17.1, 21.1, 9.4, 33.5])]) # doctest: +REMOTE_DATA
 
 After initialising the VSO client:
 
--- a/sunpy/net/vso/vso.py
+++ b/sunpy/net/vso/vso.py
@@ -91,15 +91,23 @@ def check_connection(url):
     except (socket.error, socket.timeout, HTTPError, URLError) as e:
         warnings.warn(
             "Connection failed with error {}. \n Retrying with different url and port.".format(e))
+        return None
 
 
 def get_online_vso_url(api, url, port):
-    if api is None and (url is None or port is None):
-        for mirror in DEFAULT_URL_PORT:
-            if check_connection(mirror['url']):
-                api = zeep.Client(mirror['url'], port_name=mirror['port'])
-                api.set_ns_prefix('VSO', 'http://virtualsolar.org/VSO/VSOi')
-                return api
+    if isinstance(api, zeep.client.Client):
+        if check_connection(api.wsdl.location):
+            return api
+
+    if url and check_connection(url):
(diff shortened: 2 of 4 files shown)


cquark7:
Inside `get_online_vso_url` there's a function named `check_connection`. It returns `True` only when HTTP response is 200, otherwise it returns `None`.

Cadair:
Looks like this has actually broken one of the tests, can you please fix it. Also if you could add an explicit test for this by mocking the return value of `get_online_vso_url` that would be great.

nabobalis:
I wonder if `check_connection` needs to be checked as well.

cquark7:
@Cadair Sure. I'll update this PR with updated tests.

cquark7:
@Cadair @nabobalis 
I fixed the test due to which tests were failing, but while writing a test for `get_online_vso_url` I noticed another major bug. If a user provides their own `url`, `port` or `api` to initialize VSOClient then `get_online_vso_url` will return `None`. I don't think this kind of behaviour is expected from VSOClient.

To fix this I need to modify  `get_online_vso_url`, then I can proceed and write a test for it.

Reference code:

```python3
class VSOClient(BaseClient)
    def __init__(self, url=None, port=None, api=None):
        api = get_online_vso_url(api, url, port)
        self.api = api
```

```python3
def get_online_vso_url(api, url, port):
    if api is None and (url is None or port is None):
        for mirror in DEFAULT_URL_PORT:
            if check_connection(mirror['url']):
                api = zeep.Client(mirror['url'], port_name=mirror['port'])
                api.set_ns_prefix('VSO', 'http://virtualsolar.org/VSO/VSOi')
                return api
```

nabobalis:
Sure?

docs/guide/acquiring_data/database.rst
```diff
@@ -267,7 +267,7 @@ sections.
    ...        self.instrument = instrument
    ...        self.wave = MockObject(wavemin=wavelength, wavemax=wavelength, waveunit='nm')
    >>> vso.VSOClient().search = Mock(return_value=[MockQR(f"2011050800000{t}", 'AIA', w) for
-   ...                                   t,w in zip([0,0,2,3], [17.1, 21.1, 9.4, 33.5])])
+   ...                                   t,w in zip([0,0,2,3], [17.1, 21.1, 9.4, 33.5])]) # doctest: +REMOTE_DATA
```
Cadair: why? this should be setting this up for offline doc tests.

sunpy/net/vso/vso.py
```diff
@@ -92,15 +92,23 @@ def check_connection(url):
...
 
 def get_online_vso_url(api, url, port):
-    if api is None and (url is None or port is None):
-        for mirror in DEFAULT_URL_PORT:
-            if check_connection(mirror['url']):
-                api = zeep.Client(mirror['url'], port_name=mirror['port'])
-                api.set_ns_prefix('VSO', 'http://virtualsolar.org/VSO/VSOi')
-                return api
+    if isinstance(api, zeep.client.Client):
+        if check_connection(api.wsdl.location):
```
Cadair: I wonder if this is needed, I would assume that you would be unable to create a `zeep.Client` object if the WSDL was offline. I think we can just do
```
if isinstance(api, zeep.client.Client):
    return api
```

here.

sunpy/net/vso/vso.py
```diff
@@ -257,6 +265,8 @@ class VSOClient(BaseClient):
 
     def __init__(self, url=None, port=None, api=None):
         api = get_online_vso_url(api, url, port)
+        if api is None:
+            raise ConnectionError("Cannot find an online VSO mirror")
```
Cadair:
```suggestion
            raise ConnectionError("Cannot find an online VSO mirror.")
```

docs/guide/acquiring_data/database.rst
cquark7: `VSOClient` cannot be instantiated without internet access, that's why I had to add `+REMOTE_DATA`. What would be the best way to handle this? The easiest way I can think of is to make `search` a `staticmethod` in `VSOClient`.

sunpy/net/vso/vso.py
cquark7: You're right. It's not possible to create a `zeep.Client` object when you're offline. I thought it would be safe to check WSDL status before returning API (just in case zeep.Client object was created a while ago)

Cadair: I think we are probably safe to remove the extra check, I would imagine that anyone setting the api by themselves would be doing so for weird edge cases anyway.

docs/guide/acquiring_data/database.rst
cquark7: I can mock it or I can temporarily override the constructor of `VSOClient` using a helper function. I'll see which approach is more elegant.

nabobalis:
Thanks @cquark7

PR #2986, opened 2019-03-13
Improve scaling of animator and adding aspect keyword to lineanimator

### LINKED ISSUE #2777: LineAnimator display of small y-axis values

Akram9:
### Description

Improve scaling of animator and adding aspect keyword to lineanimator, reference to  #2777


sunpy/visualization/animator/base.py
```diff
@@ -257,8 +257,8 @@ def _make_axes_grid(self):
             vert = [ysize, pad_size] * self.num_sliders + \
                    [large_pad_size, large_pad_size, Size.AxesY(self.axes)]
         else:
-            vert = [ysize, pad_size] * self.num_sliders + \
-                   [large_pad_size, Size.AxesY(self.axes)]
+            vert = [ysize, large_pad_size] * self.num_sliders + \
+                   [Size.AxesY(self.axes)]
```
nabobalis: Deleting one of the pad_sizes is ok?

Akram9: Frankly, I need to test it a bit more.

Akram9: Well this does solve most of the problem, bu tvery satisfactorily. Also, it's specific to the example in the docs.

nabobalis:
Is this finished?

sunpy/visualization/animator/base.py
nabobalis: How well does this fix the issue?

Akram9: Changing y-axis values no longer affects the plot size as earlier but changing the window size does affect the whole plot. Everything goes out of ratio in y-axis again and depending on how it's done, the slider comes on top of the plot.

Akram9:
Not yet.

Akram9:
I would actually like someone to check out the changes.

Akram9:
To run -
```
import numpy as np

from sunpy.visualization.animator import LineAnimator

data_shape0 = (10, 20)
data0 = np.random.rand(*data_shape0)

plot_axis0 = 1
slider_axis0 = 0

xdata = np.tile(np.linspace(0, 100, data_shape0[plot_axis0]+1), (data_shape0[slider_axis0], 1))
ani = LineAnimator(data0, plot_axis_index=plot_axis0, axis_ranges=[None, xdata])

plt.show()
```

nabobalis:
I had a quick try with that code and it does scale a bit better than normal.

I am not sure if it solves the original issue without checking.

Akram9:
Thanks. Also try to give `aspect` a try by - `ani = LineAnimator(data0, plot_axis_index=plot_axis0, axis_ranges=[None, xdata], aspect=float)`. It behaves pretty wierd at times.

nabobalis:
Yeah, I'm unsure on the aspect keyword.

Akram9:
I'll leave the aspect for now and see if I can better the scaling.

Akram9:
Please check the last change. I think this solves all problems here. 
The aspect feature still remains glitchy though.

nabobalis:
I think the fix helps the scaling issues with our current plots and adding aspect to lineanimator helps. But I am not sure if it solves the original issue. But I am happy to merge this in regardless.

nabobalis:
Thanks @Akram9!

nabobalis:
Having checked the code given in the original issue, this does in fact close #2777.

PR #2987, opened 2019-03-13
Changed test_online_fido to be much faster

nabobalis:
Until we mock more clients, we have parameterized test_online_fido unlike the offline version.

This should speed this test up and the online build.

### DIFF
--- a/sunpy/net/dataretriever/sources/rhessi.py
+++ b/sunpy/net/dataretriever/sources/rhessi.py
@@ -66,7 +66,7 @@ class RHESSIClient(GenericClient):
         """
         dt = TimeRange(time_range)
         # remove time from dates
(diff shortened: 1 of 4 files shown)


sunpy/net/dataretriever/sources/rhessi.py
```diff
@@ -66,7 +66,7 @@ def get_observing_summary_filename(self, time_range):
         """
         dt = TimeRange(time_range)
         # remove time from dates
-        dt = TimeRange(dt.start.datetime.date(), dt.end.datetime.date())
+        dt = TimeRange(dt.start.strftime('%Y-%m-%d'), dt.end.strftime('%Y-%m-%d'))
```
nabobalis: This is a bug I found if you pass in a time with a leap second.

PR #2993, opened 2019-03-16
Improvements to sample and download data directories

### LINKED ISSUE #2959: Download Sample Data into a hidden directory
Currently `import sunpy.data.sample` will download our [sample data](http://data.sunpy.org) to `~/sunpy` as specified by [working dir](https://github.com/sunpy/sunpy/blob/master/sunpy/data/sunpyrc#L24) in the config file.

I propose that we move to using [appdirs](https://github.com/ActiveState/appdirs) to determine the "user data directory" which is platform dependant and used for applications to put data they need (not to be directly accessed by the user). This move would mean that we wouldn't create `~/sunpy` until the user used `Fido` with the default `path=` argument.

My reasoning for this is that we only expose the sample data to users through the `import sunpy.data.sample` interface, so it is irrelevant to the users where those files actually reside on disk, and that if this is the case then we shouldn't be creating a directory in their home directory to do it.

Cadair:
This is a PR in three commits.

The first commit transitions our sample data download to parfive, which required a major refactoring of how the sample data download worked, I also removed support for zip files in the process as we were not using it.

The second commit fixes #2959 by changing the default sample data download location to the platform specific user data directory.

The third commit makes a change to the download dir for all test runs so that all tests which use the default download path download to a tmp dir rather than the actual download dir.

### DIFF
--- a/changelog/2993.feature.rst
+++ b/changelog/2993.feature.rst
@@ -1,0 +1,2 @@
+The default location of the sunpy sample data has changed to be in the platform
+specific data directory as provided by `appdirs <https://github.com/ActiveState/appdirs>`__.
--- a/sunpy/data/_sample.py
+++ b/sunpy/data/_sample.py
@@ -1,24 +1,18 @@
 # -*- coding: utf-8 -*-
 """SunPy sample data files"""
-import socket
-import os.path
 import warnings
-from shutil import move
-from zipfile import ZipFile
+from pathlib import Path
+from collections import namedtuple
 from urllib.parse import urljoin
 
-from astropy.utils.data import download_file
+import parfive
 
-from sunpy.util.net import url_exists
 from sunpy.util.config import get_and_create_sample_dir
 from sunpy.util.exceptions import SunpyUserWarning
 
-__author__ = "Steven Christe"
-__email__ = "steven.christe@nasa.gov"
-
 _base_urls = (
     'http://data.sunpy.org/sunpy/v1/',
-    'https://github.com/sunpy/sample-data/raw/master/sunpy/v1/'
+    'https://github.com/sunpy/sample-data/raw/master/sunpy/v1/',
 )
 
 # Shortcut requirements:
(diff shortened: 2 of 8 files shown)


sunpy/data/sunpyrc
```diff
@@ -35,12 +35,11 @@ time_format = %Y-%m-%d %H:%M:%S
...
 ; SunPy working directory.
 ; Default value: data/
-;download_dir = /tmp
 download_dir = data
 
-; Location where the sample data will be downloaded. Path should be specified
-; relative to the SunPy working directory.
-sample_dir = data/sample_data
+; Location where the sample data will be downloaded. If not specified, will be
+; downloaded to platform specific user data directory.
```
ehsteve: could you add some examples here for different systems?

sunpy/data/_sample.py
```diff
@@ -63,96 +58,66 @@
...
+    # Creating the directory for sample files to be downloaded
+    sampledata_dir = Path(get_and_create_sample_dir())
 
+    dl = parfive.Downloader(overwrite=overwrite)
 
-def get_sample_file(filename, url_list, show_progress=True, overwrite=False,
-                    timeout=None):
-    """
-    Downloads a sample file. Will download  a sample data file and move it to
-    the sample data directory. Also, uncompresses zip files if necessary.
```
ehsteve: and deletes the zip file

sunpy/data/_sample.py
```diff
@@ -63,96 +58,66 @@
...
-    Returns the local file if exists.
```
ehsteve: if it exists.

changelog/2993.feature.rst
```diff
@@ -0,0 +1,3 @@
+The default location of the sunpy sample data has changed to be in the platform
```
ehsteve: Suggest to just say "The default location of the sample data is now placed in platform specific data directory as provided by X".

sunpy/data/sunpyrc
nabobalis: Should we put that in the sunpyrc  file and not maybe a doc string?

Cadair: You mean like where they will be put by default? I was assuming that this file isn't a way to document the defaults, and that if the user is here changing this they are not going to care to set it to something close to the default?

ehsteve: From a user's perspective I would be annoyed because if I want to know this I now have to search. Maybe just link to the package that does this?

nabobalis: I am ok linking it here. I think I did at the top of the file when I switched to appdirs.

ehsteve: I'm fine with this PR once and also likely an update to the docs page on sample data as well.

Cadair: The docs currently state:

> This will download the data to your sample-data directory which can be customized by editing the sunpyrc file (see Customizing SunPy).

Which still holds true, so I think they are ok.

PR #2994, opened 2019-03-16
Some more Python 2 cleanup

### LINKED ISSUE #2287: Drop Python 2 Support
This is the meta issue for dropping Python 2 in SunPy. This is scheduled to be done *after* the 0.9 release.

# Cleaning and Tidying

- ~`sunpy.extern.six` can be removed and all six imports in the code can be replaced by the Python 3 only ones.~
- ~All `__future__` imports.~
- ~Search for and remove any if `six.PY2` blocks and also search for `sys.version_info` or any usage of `minversion` or `Looseversion`.~
- Remove the Python 2 / 3 compatibility checks in the ANA C-API code..
- Search the code for comments mentioning Python to see if changes can be made under 3 only.

# Code Improvements / Style Changes

- ~Replace all instances of `u.quantity_input(keyword=unit)` with the function annotiation equivalent. (This means it will form part of the function definition and be documented by sphinx).~

See astropy/astropy#6234 for their version of this issue.

Cadair:
I think this is the last of it, closes #2287

### DIFF
--- a/sunpy/database/database.py
+++ b/sunpy/database/database.py
@@ -380,13 +380,9 @@ class Database(object):
         """
         self.session.commit()
 
-    def _download_and_collect_entries(self, query_result, **kwargs):
-
-        client = kwargs.pop('client', None)
-        path = kwargs.pop('path', None)
-        progress = kwargs.pop('progress', False)
-        methods = kwargs.pop('methods', ('URL-FILE_Rice', 'URL-FILE'))
-        overwrite = kwargs.pop('overwrite', False)
+    def _download_and_collect_entries(self, query_result, client=None,
+                                      path=None, progress=False, methods=None,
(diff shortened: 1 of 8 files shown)


sunpy/database/database.py
```diff
@@ -442,12 +438,8 @@ def _download_and_collect_entries(self, query_result, **kwargs):
                 entry.download_time = datetime.utcnow()
                 yield entry
 
-
     def fetch(self, *query, **kwargs):
-
         """
-        fetch(*query[, path, overwrite, client, progress, methods])
```
nabobalis: Isn't this for the doc builds like we have in fido?

Cadair: You shouldn't specify the call sig in the docstring it should be self explanatory.

PR #3000, opened 2019-03-18
Tests for animator class

Akram9:
Progressing on the test file `test_basefuncanimator` of animator classes.


Cadair:
Congrats on #3000 😉

Akram9:
Please check if I'm going right.

The tests for `key_press` and `_sanitize_axis_ranges` contain only one case while the code actually has lots of conditions. Do I need to test every possible combination of conditions?

nabobalis:
> Do I need to test every possible combination of conditions?

In an ideal world yes but it's not always possible and it might involve too much work to do feasibly.

Akram9:
> > Do I need to test every possible combination of conditions?
> 
> In an ideal world yes but it's not always possible and it might involve too much work to do feasibly.

Right. I'll see if I can get some major conditions done.

Akram9:
So I did all conditions of sanitizer test and found the 6th and 7th to fail. Seems to be actual code failure - 
1. the 6th test failure is because of creating a weirdly shaped array that numpy doesn't like - line 541 of `base.py`.
2. the 7th test failure has to do with the length of shape and dimensions of the input `axis_range` - line 586 of `base.py`.

nabobalis:
If the test fails, you should see if you can patch the code.

sunpy/visualization/animator/base.py
```diff
@@ -526,7 +526,7 @@ def _sanitize_axis_ranges(self, axis_ranges, data_shape):
                     # and APIs using both [min, max] pair and manual definition of each pixel
                     # values can be unambiguously and simultanously supported.
                     extent += [axis_ranges[i][0], axis_ranges[i][-1]]
-                    axis_ranges[i] = edges_to_centers_nd(axis_ranges[i])
+                    axis_ranges[i] = edges_to_centers_nd(axis_ranges[i], i)     # probably this change is required(?)
```
nabobalis: We don't need this comment.

Akram9: Removed.

sunpy/visualization/animator/base.py
```diff
@@ -526,7 +526,7 @@ def _sanitize_axis_ranges(self, axis_ranges, data_shape):
                     # and APIs using both [min, max] pair and manual definition of each pixel
                     # values can be unambiguously and simultanously supported.
                     extent += [axis_ranges[i][0], axis_ranges[i][-1]]
-                    axis_ranges[i] = edges_to_centers_nd(axis_ranges[i])
+                    axis_ranges[i] = edges_to_centers_nd(axis_ranges[i], i)
```
nabobalis: Can you also see if this happens in other places?

Akram9:
Right, I'll see what I can do.

sunpy/visualization/animator/base.py
Akram9: Sure, though it'll take some time.

Akram9:
The sixth failure was because of wrong data type assumption in the test, which has been resolved.
The patch for the seventh failure is at #3013.

nabobalis:
I think it would be better of you discover a bug while testing, to fix it in this PR.

Akram9:
> I think it would be better of you discover a bug while testing, to fix it in this PR.

I'll take care of this.

sunpy/visualization/animator/base.py
```diff
@@ -558,7 +558,7 @@ def plot_start_image(self):  # pragma: no cover
         raise NotImplementedError("Must be defined and used by subclasses of {0}.".format(self.__class__))
 
     @abc.abstractmethod
-    def update_plot(self, val, slider):  # pragma: no cover
+    def update_plot(self, val, im=None, slider):  # pragma: no cover
```
nabobalis: Can you explain this for me?

Akram9: Running the ImageAnimatorWCS example from docs, you get :
``` File "/mnt/h/Documents/Projects/sunpy-dev/sunpy/visualization/animator/image.py", line 119, in update_plot
    super().update_plot(val, im, slider)
TypeError: update_plot() takes 3 positional arguments but 4 were given
Aborted (core dumped)
```
error when playing the animation. Apparently, `update_plot` in base takes in two inputs, while three were given in ImageAnimatorWCS class.
(Perhaps there's a better fix?)

What does `# pragma: no cover` mean?

nabobalis: `# pragma: no cover` lets any code checker to ignore this line. You should remove it, I am not sure why it is there. 

But `im` is not used in that function as I can see? I'm not sure why it would need to be passed in.

nabobalis: I would at least remove the `=None`.

Akram9: Traceback:
```Traceback (most recent call last):
  File "/home/akram/miniconda3/envs/sunpy-dev/lib/python3.7/site-packages/matplotlib/backend_bases.py", line 1238, in _on_timer
    ret = func(*args, **kwargs)
  File "/mnt/h/Documents/Projects/sunpy-dev/sunpy/visualization/animator/base.py", line 363, in _step
    s.set_val(s.val+1)
  File "/home/akram/miniconda3/envs/sunpy-dev/lib/python3.7/site-packages/matplotlib/widgets.py", line 438, in set_val
    func(val)
  File "/mnt/h/Documents/Projects/sunpy-dev/sunpy/visualization/animator/base.py", line 332, in _slider_changed
    self.slider_functions[slider.slider_ind](val, self.im, slider)
  File "/mnt/h/Documents/Projects/sunpy-dev/sunpy/visualization/animator/image.py", line 220, in update_plot
    super().update_plot(val, im, slider)
  File "/mnt/h/Documents/Projects/sunpy-dev/sunpy/visualization/animator/image.py", line 119, in update_plot
    super().update_plot(val, im, slider)
TypeError: update_plot() takes 3 positional arguments but 4 were given
Aborted (core dumped)
```

So yeah, lines 119 and 220 of image.py call `update_plot` from base ArrayAnimator.

Akram9: As for `im`'s usage, you are right, it is not needed. So well I'll just delete it from 119 & 220 of image.(?)

Akram9: I tried it that way and it doesn't seem to work.

Akram9: > I would at least remove the `=None`. If the other two values are not set, I think that `im` should be the same.

Well, as for that, it needs to be checked if there are any other calls where setting `im` could cause a problem.

nabobalis: You could remove `im`. You have to check the functions correctly. Since I think it goes base -> image -> imageWCS.

Akram9: There is a call as such(super) in `line`, which just calls the two values.

nabobalis: What about those lines?

Akram9: They call `update_plot` with all the three arguments.

nabobalis: Can you also change line 41 to 
```
widget instance passed to them, e.g., ``update_slider(val, im, slider)``
```

Akram9: Sure.

nabobalis: Well then we should add `im` to that function and leave it as is.

Akram9: I feel I got a solution. I'll push...

sunpy/visualization/animator/line.py
```diff
@@ -156,4 +156,5 @@ def update_plot(self, val, line, slider):
                 line.set_xdata(self.xdata[tuple(item)])
             slider.cval = val
         # Update slider label to reflect real world values in axis_ranges.
-        super().update_plot(val, slider)
+        im = None
```
nabobalis:
```suggestion
```

sunpy/visualization/animator/line.py
```diff
@@ -156,4 +156,5 @@ def update_plot(self, val, line, slider):
...
+        super().update_plot(val, im, slider)
```
nabobalis:
```suggestion
        super().update_plot(val, None, slider)
```

nabobalis: Can you check if this happens in other places?

Akram9: Okay, this looks clean. Let me try.

sunpy/visualization/animator/tests/test_basefuncanimator.py
```diff
@@ -63,3 +86,49 @@ def test_edges_to_centers_nd():
     expected[:, edges_axis] = np.arange(10.5, 19)
     output = base.edges_to_centers_nd(axis_range, edges_axis)
     assert np.array_equal(output, expected)
+
+
+class ArrayAnimatorTest(ArrayAnimator):
```
nabobalis: If there is an `arrayanimator` test file, these should go in there.

Akram9:
Are these tests good enough?

sunpy/visualization/animator/base.py
```diff
@@ -558,7 +558,7 @@ def plot_start_image(self):  # pragma: no cover
         raise NotImplementedError("Must be defined and used by subclasses of {0}.".format(self.__class__))
 
     @abc.abstractmethod
-    def update_plot(self, val, slider):  # pragma: no cover
+    def update_plot(self, val, im, slider):
```
Cadair: it feel weird calling this `im` here, as it is either `im` or `line` or some other plot object. Maybe `artist` is a good name?

sunpy/visualization/animator/line.py
```diff
@@ -156,4 +156,4 @@ def update_plot(self, val, line, slider):
                 line.set_xdata(self.xdata[tuple(item)])
             slider.cval = val
         # Update slider label to reflect real world values in axis_ranges.
-        super().update_plot(val, slider)
+        super().update_plot(val, None, slider)
```
Cadair: I think it makes more sense to pass `line` here and not `None`

sunpy/visualization/animator/tests/test_basefuncanimator.py
```diff
@@ -29,15 +33,34 @@ def button_func1(*args, **kwargs):
...
                            button_labels=buttons[1])
 
     tfa.label_slider(0, "hello")
+    assert tfa.sliders[0]._slider.label.get_text() == "hello"
+
+    tfa._set_active_slider(1)
+    assert tfa.active_slider == 1
+
+    fig = plt.figure()
+    event = mback.KeyEvent(name='key_press_event', canvas=fig.canvas, key='down')
```
Cadair: *nice*

Cadair:
@Akram9 You can see the lines which have been executed on codecov: https://codecov.io/gh/sunpy/sunpy/pull/3000/src/sunpy/visualization/animator/base.py?before=sunpy/visualization/animator/base.py or by running `tox -e py37-offline -- --cov sunpy --cov-report=html` locally.

I would encourage you to try and get all of `base.py` tested in this PR if you are willing to continue working on this.

sunpy/visualization/animator/line.py
Akram9: If I remember right, this causes some issue. Let me try.

Akram9: Okay, right it works (obviously). My bad.

sunpy/visualization/animator/tests/test_basefuncanimator.py
Akram9: Thanks!

Akram9:
I'll give this a try right away. 
And yeah, I would continue pushing to this, at least till I reach figure tests.

sunpy/visualization/animator/base.py
Akram9: Sure, it's better. I'll update this.

Akram9:
If I could have some pointers to what all I am missing in unit tests in base, except `make_axes_grid`, that will be great.

Most of the functions left out seem to be figure related.

sunpy/visualization/animator/tests/test_basefuncanimator.py
```diff
@@ -29,15 +34,46 @@ def button_func1(*args, **kwargs):
...
+    tfa._key_press(event)
+    assert tfa.active_slider == 0
+
+    slide = tfa.active_slider
+    slider = widgets.Slider(fig.gca(), str(slide), ranges[slide][0], ranges[slide][1], valinit=0)
```
Akram9: I am not sure of the valinit thing. That way I was able to pass the tests anyway. Apparently, initial value of slider is default (0.5), and therefore gives assertion error in line 73.

Akram9:
How do I test `make_axes_grid` and `add_widgets`? Figure tests?

sunpy/visualization/animator/tests/test_basefuncanimator.py
```diff
@@ -29,15 +34,56 @@ def button_func1(*args, **kwargs):
...
+    event = mback.KeyEvent(name='key_press_event', canvas=fig.canvas, key='down')
+    tfa._key_press(event)
+    assert tfa.active_slider == 0
+
+    event = mback.KeyEvent(name='key_press_event', canvas=fig.canvas, key='up')
+    tfa._key_press(event)
+    assert tfa.active_slider == 1
+
+    slide = tfa.active_slider
+    slider = widgets.Slider(fig.gca(), str(slide), ranges[slide][0], ranges[slide][1], valinit=0)
```
Cadair: Out of interest why are you creating this slider here rather than using one from the `tfa` object?

sunpy/visualization/animator/tests/test_basefuncanimator.py
```diff
@@ -29,15 +34,56 @@ def button_func1(*args, **kwargs):
...
+
+    tfa._set_active_slider(1)
+    assert tfa.active_slider == 1
+
+    fig = plt.figure()
+    event = mback.KeyEvent(name='key_press_event', canvas=fig.canvas, key='down')
+    tfa._key_press(event)
+    assert tfa.active_slider == 0
+
+    event = mback.KeyEvent(name='key_press_event', canvas=fig.canvas, key='up')
```
Cadair: Can you add similar tests for left, right and `p`  here?

sunpy/visualization/animator/tests/test_basefuncanimator.py
```diff
@@ -63,3 +114,54 @@ def test_edges_to_centers_nd():
...
+    def __init__(self, data):
+        self.naxis = data.ndim
+        self.image_axes = [1]
+        self.slider_axes = [0]
+
+    def plot_start_image(self, ax):
+        pass
+
+    def update_plot(self, val, artist, slider):
+        pass
```
Cadair: Adding the super call in here will enable you to test the code in the base class.

```suggestion
        super().update_plot(val, artist, slider)
```

sunpy/visualization/animator/tests/test_basefuncanimator.py
Cadair: I commented `valint` out and it didn't seem to fail the test.

Cadair:
> How do I test `make_axes_grid` and `add_widgets`? Figure tests?

The code is already being executed through constructing the class, but I think a figure test would ensure that the layout is constant. So yes I think it would be worth writing some figure tests for construction of various animators.

sunpy/visualization/animator/tests/test_basefuncanimator.py
Akram9: Right, it doesn't. I was doing something wrong earlier perhaps?

sunpy/visualization/animator/tests/test_basefuncanimator.py
Akram9: I don't see how I can do that? Shouldn't it be the same for the buttons widgets too?

Akram9: Well, actually I can. I am an idiot.

sunpy/visualization/animator/tests/test_basefuncanimator.py
Akram9: I was messing up with the sliders.

sunpy/visualization/animator/tests/test_basefuncanimator.py
Akram9: Sliders are supposed to be `widget` objects, while the sliders in `tfa` object are `axes` objects.
Is this the same slider you are pointing at? Apart from that, I don't see any other `widget` sliders in `tfa`.

sunpy/visualization/animator/tests/test_basefuncanimator.py
Akram9: Will update the tests with them.

sunpy/visualization/animator/tests/test_basefuncanimator.py
Akram9: Am I thinking right?

Cadair: if you do `tfa.sliders[0]._slider` you get the `Slider` object.

Akram9: Oh yes, I had used it elsewhere too. Very stupid of me.

Akram9: Yeah, the same had to be done for the button widget too.

Akram9:
Hashes already present for the figure tests or do we need to generate them?

nabobalis:
You need to generate any new hashes for figures and add them to the json file.

Akram9:
I tried with the figure test for some time but I don't get where I am really failing.

nabobalis:
I just pushed what I think is a fix for the figure test.

Akram9:
Thanks for the correction. The tests pass now. Apparently, I was changing the hash in the `tmp` folder.
So is the hash and image generated by tox? That's what happens when I remove the line from the JSON file and run the test again.

nabobalis:
Yes those in that folder are generated by tox, the fixed file is in the tests subfolder

Cadair:
Could you add an `ArrayAnimator` figure test as well and then I think we can call this one done!

Akram9:
Yup. Will do so by this weekend.

Akram9:
An ArrayAnimator figure test would require defined `plot_start_image` and `update_plot`, which, if I go the LineAnimator way, will effectively become LineAnimator figure test. Same for ImageAnimator. So should I anyway minimally define those two or rather do an ImageAnimator figure test?

Cadair:
Ah, you probably want to be testing `ImageAnimator` rather than `ArrayAnimator` my bad. The difference should be that it should be plotting an image rather than lines (obviously lol)

Akram9:
Alrighty, doubts cleared. Will do so in a day or two. A bit busy with college now.

Cadair:
No worries.

Akram9:
For the last test, I pretty much copy-pasted from the example. Am I thinking right that it can be made a bit shorter?

nabobalis:
Probably the  least amount of code for the image, but could you use the test data instead please? I am not sure we support using sample-data in our tests.

Akram9:
Okay, I'll try to improve it soon.

Akram9:
Okay, I'll try to improve it soon.

sunpy/visualization/animator/tests/test_basefuncanimator.py
```diff
@@ -180,3 +181,23 @@ def test_lineanimator_figure():
     fig = plt.figure()
     ani = LineAnimator(data0, plot_axis_index=plot_axis0, axis_ranges=[None, xdata], fig=fig)
     ani.plot_start_image(ax=fig.gca())
+
+
+@figure_test
+def test_imageanimator_figure():
+    from itertools import product
```
Cadair: can you move the imports to the top of the file?

Akram9: Yup, I will.

Akram9:
Seems too many tests are failing. Did I break something?

Cadair:
Ah, no that's on me looks like I broke it with parfive 1.0 I will try and fix that later today. (Don't import the sample data into tests and a few more of the builds should pass though)

Akram9:
So I corrected the mouse button test and simplified the code at the cost of pep8.

PR #3005, opened 2019-03-19
made tests use default config

### LINKED ISSUE #3004: Tests should not make use of user config file
### Description
Tests make use of the user provided config file but have hard coded dependencies on expectations from the default configuration. This is most easily seen in tests for time. Many tests are expecting a particular time format namely `%Y-%m-%d %H:%M:%S` but the user could have set that to another format. Those tests would then unexpectedly fail.  

### Expected behavior
I expected the tests to pass.

### Actual behavior
The tests failed.

### Steps to Reproduce
Change your default config time_format from `%Y-%m-%d %H:%M:%S` to `%Y-%m-%d %H:%M` and watch the tests fail.

nabobalis:
Fixes #3004 (probably).

### DIFF
--- a/sunpy/conftest.py
+++ b/sunpy/conftest.py
@@ -1,3 +1,4 @@
+import os
 import json
 import pathlib
 import warnings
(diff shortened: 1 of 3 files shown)


Cadair:
Are there tests to check that the env var is correctly obeyed?

nabobalis:
In `test_get_user_configdir` I manually set it and check it worked.

PR #3009, opened 2019-03-20
Added function to generate user config file

### LINKED ISSUE #3006: Provide an easy way to generate a user config
There is no simple way to put a config file in your config directory. maybe ` sunpy.util.config.write_config` function that saves the current state of the config to the user config dir?

cquark7:
### Description

Fixes #3006

### DIFF
--- a/sunpy/util/config.py
+++ b/sunpy/util/config.py
@@ -3,11 +3,14 @@ import os
 import configparser
 from pathlib import Path
 import shutil
+import warnings
 
 import sunpy
 from sunpy.extern.appdirs import AppDirs
+from sunpy.util.exceptions import SunpyUserWarning
 
-__all__ = ['load_config', 'print_config', 'CONFIG_DIR']
+
+__all__ = ['load_config', 'write_config', 'print_config', 'CONFIG_DIR']
 
 # This is to avoid creating a new config dir for each new dev version.
 # We use AppDirs to locate and create the config directory.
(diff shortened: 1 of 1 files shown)


sunpy/util/config.py
```diff
@@ -2,11 +2,11 @@
 import os
 import configparser
 from pathlib import Path
+import shutil
 
 import sunpy
 from sunpy.extern.appdirs import AppDirs
 
-
 __all__ = ['load_config', 'print_config', 'CONFIG_DIR']
```
nabobalis: New function should be added here.

sunpy/util/config.py
```diff
@@ -123,3 +123,26 @@ def _get_user_configdir():
...
+        raise RuntimeError(f'Could not write to SunPy config directory="{user_config_dir}"')
+
+    # No config file exists in user directory
+    if not user_config_file.exists():
+        shutil.copyfile(config_file, user_config_file)
+
+    # File exists and needs to be overwritten
+    elif overwrite:
+        os.remove(user_config_file)
+        shutil.copyfile(config_file, user_config_file)
```
nabobalis: I feel like this this if/elif block could be rewritten in a different manner. 

Also I think if the file already exists, then you we should tell the user that it does and if they want to write to it, they have to use overwrite.

cquark7: @nabobalis I wanted to add an `else` after the `elif` block to display a warning message, but I was unsure if it's necessary. That's why I structured the if/elif block in this way.  Now that you mention it, I think it can be rewritten in a better way.

How should I tell the user about existing config file -- using `warnings.warn(message)` or a simple `print(message)`? I think a warning message would be better, but that will also require importing another library simply for this.

cquark7: P.S. Any idea why some tests are failing?

nabobalis: The test failures seem to be unrelated.

I think a warning.warn would be a better idea.

cquark7:
@Cadair Thanks for the review. I will add the test and changelog entry tomorrow.

Raahul-Singh:
@nabobalis so just a test is required here, right?

nabobalis:
@Raahul-Singh I think that's all that was needed.

PR #3010, opened 2019-03-20
Fixed missing positional argument while calling function

Akram9:
Fixes missing positional argument in `animator/base.py`.


nabobalis:
Thanks @Akram9

PR #3011, opened 2019-03-21
Fix for example hmi.py

nabobalis:
Should stop the example from failing but I think it suggests a deeper problem.

### DIFF
--- a/examples/acquiring_data/hmi.py
+++ b/examples/acquiring_data/hmi.py
@@ -27,26 +27,31 @@ result = Fido.search(a.Time('2015/11/04 12:00:00', '2015/11/04 12:10:00'),
 
 ###############################################################################
 # Now we can see what results we obtained from our search.
+# Notice we have two files. One is the full disk image we plan to display
+# and the other is a synoptic version of said image.
+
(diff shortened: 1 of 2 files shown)


examples/acquiring_data/hmi.py
```diff
@@ -27,26 +27,28 @@
...
 
 ###############################################################################
 # Once we are happy with the results obtained from the search.
 # We can download the data with `Fido.fetch` .
 # In this case we only want one file so we can index the result.
 
-# Notice we have two files. One is the full disk image we plan to display
-# and the other is a synoptic version of said image.
-downloaded_files = Fido.fetch(result)
+downloaded_file = Fido.fetch(result[0, 0])
```
Cadair: Given the slightly peculiar semantics of this slicing I suggest we add a comment like
```
# Slice the first record returned by the first client
```
or similar to explain both the indices.

nabobalis: Good point

nabobalis: Added a sentence abov e

PR #3012, opened 2019-03-23
Fixed the colour of mask

### LINKED ISSUE #3007: "Masking the solar disk" gallery example does not have a black mask
### Description
```python
palette.set_bad('black')
```

### Steps to Reproduce
See example gallery in docs.

ankursikarwar:
### Description

Fixes #3007

### DIFF
--- a/changelog/3012.bugfix.rst
+++ b/changelog/3012.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed the bug that made the mask color white in map_disk.py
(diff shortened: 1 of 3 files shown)


nabobalis:
Thanks @ankursikarwar, this only fixes the example and not what seems to be an issue with setting the `palette`?

ankursikarwar:
@nabobalis  But the original image itself has the same brightness around the edges. 

sphx_glr_aia_example_001

nabobalis:
Sure. But the aim of the example as I see it is to mask out the middle and brighten the edges. Which is what the normalisation aims to do. So I don't think we can set it to None.

nabobalis:
Sorry @ankursikarwar, I don't think we can go ahead and use this fix for the issue. 

If you want to make more changes or try a different issue, you can either reopen this or make a new PR.

PR #3013, opened 2019-03-23
Modify `edges_to_centers_nd` calls in `base.py`

Akram9:
`edges_to_centers_nd` is dependent on `edge_axis` which apparently affects some calls. Probably there is more to fix.


Akram9:
The problem arises when `axis_ranges[i].ndim == 1` and `i == 1` hold true as while applying `edges_to_centers_nd` in this case will cause index error while slicing. Probably a better solution could be to alter `edges_to_centers_nd` to incorporate such issues. But I would like someone's input here.

PR #3016, opened 2019-03-25
Remove some warnings from timeseries tests

dstansby:
More warning removal, this time for timeseries tests.


nabobalis:
Thanks @dstansby, just looks like you need to import `SunpyUserWarning` in the test file.

nabobalis:
It seems to error now that no warning is raised. Removing the flag raises the warning but with it, it seems to fail the test? It seems odd to me.

dstansby:
I can reproduce this behaviour with the standard `setup.cfg`. Might be something to do with custom warnings not playing nicely with the way pytest filters warnings/how the `no-warnings` command line option works?

nabobalis:
Yeah that is a good point, I am not sure how pytest handles custom warnings. We probably want to look into or ignore those warnings for now.

PR #3019, opened 2019-03-27
Document that Helioprojective observer can be a string

dstansby:
e.g. it it is fine to pass `observer='earth'`.

### DIFF
--- a/sunpy/coordinates/frames.py
+++ b/sunpy/coordinates/frames.py
@@ -319,8 +319,10 @@ class Helioprojective(SunPyBaseCoordinateFrame):
     obstime: SunPy Time
         The date and time of the observation, used to convert to heliographic
         carrington coordinates.
-    observer: `~sunpy.coordinates.frames.HeliographicStonyhurst`
-        The coordinate of the observer in the solar system.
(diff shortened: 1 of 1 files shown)


nabobalis:
Thanks again @dstansby

sunpy/coordinates/frames.py
```diff
@@ -319,8 +319,10 @@ class Helioprojective(SunPyBaseCoordinateFrame):
     obstime: SunPy Time
         The date and time of the observation, used to convert to heliographic
         carrington coordinates.
-    observer: `~sunpy.coordinates.frames.HeliographicStonyhurst`
-        The coordinate of the observer in the solar system.
+    observer: `~sunpy.coordinates.frames.HeliographicStonyhurst`, str
+        The coordinate of the observer in the solar system. If a string must
+        be a solar system body that can be parsed by
+        ~`sunpy.coordinates.ephemeris.get_body_heliographic_stonyhurst`.
```
Cadair: This tick and `~` are in the wrong order?

nabobalis: This is what I get for being blind. I will book mark this and fix it.

PR #3022, opened 2019-03-29
Trick GitHub into seeing we have a code of conduct for the repo

Cadair:
It appears that GitHub hasn't automatically detected our rst code of conduct file, so this is a hack.


nabobalis:
Do we need to do this for the licence too?

Cadair:
I don't think so.

PR #3025, opened 2019-03-31
Catch some more warnings

dstansby:
A bit of a random selection, but gets rid of some more warnings in the tests. Slowly getting there...


nabobalis:
Doc fail due to timeout or connection issue on download.

Issue #2849, opened 2018-11-20
Narrative documentation, one sentence per line


nabobalis:
I personally am not a fan that we break up sentences in our rst files. 
Not only do we break sentences in nearly random places, any change made makes the diff much larger than it needs to be.

Thus I am suggesting we move to one sentence per line in all our RST files and **only** the RST files. 

However, I probably should ask first before I butcher all the RST files.

Cadair:
I could get on board with this.

dstansby:
I'm 👍 for this if someone wants to put in a PR to update everything... that we should do this is already documented in our developer docs.

dpshelio:
This [pre-pep thread about OSPL](https://discuss.python.org/t/one-sentence-per-line-for-peps-and-more/13920) and the [one above Semantic Line Breaks](https://discuss.python.org/t/semantic-line-breaks/13874) are quite interesting. I'm happy to keep it OSPL, but I'd wish there's a way to lint the RST files automatically into that, so it's not such a manual work. I did it using [an emacs function](https://stackoverflow.com/a/41638368/1087595), but I had to be careful as various reStructuredText tags were converted incorrectly.

wtbarnes:
Having a quick look through our narrative docs, it looks like we now adhere to this.

This is now codified in the developer docs: https://docs.sunpy.org/en/latest/dev_guide/contents/documentation.html#sunpy-specific-rules so I'm going to close this. If there are specific places where this is still not the case, I would suggest we open up specific issues to address that.

PR #3031, opened 2019-04-03
Doc guidance on signing up for an issue by comment

### LINKED ISSUE #3029: Add guidance about how to sign up for solving an issue to the documentation
### Description

To reduce the likelihood of people working on the same issue at the same time there needs to be guidance provided to let people know how to `sign up`. The contribution docs should provide guidance on this.

### Additional context

None.

ehsteve:
### Description

Solves issue #3029 to provide guidance for developers to let people know that they are working on a particular issue.

Fixes #3029

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -75,12 +75,14 @@ Generally the more people that can look over a pull request the better it will t
 
 .. _pull requests: https://github.com/sunpy/sunpy/pulls
 
-Code
-----
+Code (to solve Issues)
+----------------------
 
 If you would prefer to code instead, we have several repositories you can investigate.
 The main one is the SunPy repository with where all the known `issues`_ with SunPy are detailed.
 Each issue should have a series of labels that provide information about nature of the issue.
+If you found an issue you'd like to work on, please make sure to add a comment on the issue
+to let people know that you are working on it!
 
 If you are unsure where to start we suggest the `Package Novice label`_.
 These are issues that have been deemed a good way to be eased into SunPy and are achievable with little understanding of the SunPy codebase.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -75,20 +75,22 @@ Generally the more people that can look over a pull request the better it will t
 
 .. _pull requests: https://github.com/sunpy/sunpy/pulls
 
-Code
-----
+Code (to solve Issues)
```
nabobalis: What if it's not an issue but the code is useful?

CONTRIBUTING.rst
```diff
@@ -75,20 +75,22 @@ Generally the more people that can look over a pull request the better it will t
...
+----------------------
 
 If you would prefer to code instead, we have several repositories you can investigate.
 The main one is the SunPy repository with where all the known `issues`_ with SunPy are detailed.
 Each issue should have a series of labels that provide information about nature of the issue.
+If you found an issue you'd like to work on, please make sure to add a comment on the issue
+to let people know that you are working on it!
```
nabobalis: Any chance of removing the line break?

CONTRIBUTING.rst
Cadair: Yeah I am not sure we need this caveat here either.

ehsteve: All the text below talks about working on issues which I thought was on purpose. If we want to expand the scope of this section, that's fine.

CONTRIBUTING.rst
ehsteve: That's a bridge too far!

CONTRIBUTING.rst
nabobalis: Yeah, I think adding a paragraph on what if you want to just contribute code upstream too would be good.

ehsteve: Done.

CONTRIBUTING.rst
ehsteve: Done.

nabobalis:
You will hate me but please can you only break each sentence on a new line and not midsentance?

ehsteve:
@nabobalis I will only do that if you can point to me where that is requirement is described in the contribution guide for docs and if it does not exist then show me your PR to fix that :P

nabobalis:
Yes that is a good point, that line should be in the dev guide's doc file and not the contributing guide. 

We do have an open issue https://github.com/sunpy/sunpy/issues/2849 that no one pushed back on.

nabobalis:
it be easier if I just fix it in a separate PR and not worry about it here.

nabobalis:
Thanks @ehsteve

Issue #2984, opened 2019-03-12
TimeSeries support for EVE/ESP level 1 data


hayesla:
### Description

Currently the timeseries source in timeseries.sources.eve.py only supports level 0CS data and cannot handle the eve_l1_esp fits files (which would be easy to implement) or the level 2 emission lines time series (see [#Issue 2407 ](https://github.com/sunpy/sunpy/issues/2407))

Should this be included in the eve.py source or a new source created for the different levels?
Would be nice to see timeseries being able to support this! :)

aringlis:
Additionally, it may be that there is no value in the Level 0 data anymore once we are providing the Level 1 and Level 2 data instead...

hayesla:
This has now been fixed with #3032

PR #3032, opened 2019-04-04
EVE ESP timeseries source added

hayesla:
### Description

Currently the EVE timeseries supports only level 0 data and cannot handle level1 EVE/ESP data. ESP is a detector which is part of EVE on SDO, providing 4 EUV and 1 soft X-ray irradiance measurements. This follows from #2984 

I added a source class names ```ESPTimeSeries``` within ```sunpy/timeseries/sources/eve.py```.

Example
```

res = Fido.search(a.Time('2011-02-15 00:00', '2011-02-15 22:00'), a.Instrument('ESP'))
files = Fido.fetch(res)
esp_ts = ts.TimeSeries(files)
esp_ts.peek()
```

I also added some tests. We should probably include a sample fits for this too.


Cadair:
There is one critical thing wrong with this PR as it stands. It ruins our lovely line count statistics :stuck_out_tongue_closed_eyes: The test data file you added needs to be marked as binary in git, the easiest way to do this is to put a `.fits` file extension on the name of the file.

Cadair:
Also it's 39Mb which is way too big to check into the repo. Can you truncate it or modify it so it's more about ~~100Kb. Thanks!

sunpy/timeseries/sources/eve.py
```diff
@@ -2,19 +2,121 @@
...
+
+        * 'QD' - sum of 4 quad diodes, this is the soft X-ray measurments 0.1-7nm
+        * 'CH_18' - EUV irradiance 18nm
+        * 'CH_26' - EUV irradiance 26nm
+        * 'CH_30' - EUV irradiance 30nm
+        * 'CH_36' - EUV irradiance 36nm        
+
+    References
+    ----------
+    * `SDO Mission Homepage <https://sdo.gsfc.nasa.gov/>`_
```
Cadair: All of these URLs should have `__` after rather than `_`.

sunpy/timeseries/sources/eve.py
```diff
@@ -2,19 +2,121 @@
...
+            ax.legend(loc='upper right')
+        axes[-1].set_xlabel('Time (UT) ' + str(self.data.index[0])[0:11])
+        axes[-1].xaxis.set_major_formatter(dates.DateFormatter('%H:%M'))
+        plt.tight_layout()
+        plt.subplots_adjust(hspace=0.05)
+
+        figure.show()
+
+    @classmethod
+    def _parse_file(cls, filepath):
```
Cadair: I would need to double check this, but I don't think you need this method for fits files.

nabobalis:
@ehsteve Second one*

sunpy/timeseries/sources/eve.py
hayesla: thanks for edits - have updated now and added a truncated version of fits file :P

hayesla:
hrmm can't figure out what its failing on - something to do with .peek - anyone any suggestions?

nabobalis:
You created a new figure test with `@figure_test` and there is no hash to do the comparison to.
`E           Failed: Hash not present: sunpy.timeseries.tests.test_timeseriesbase.test_esp_peek` 

So you will need to add the hash to the file of hashes located `sunpy/tests/figure_hashes_py36.json`. To work out the hash tho involves running the figure test under the figure env using `tox -e figure`, it should dump out the results into a `tmp` folder in the root of the sunpy repo.

sunpy/timeseries/sources/eve.py
```diff
@@ -50,11 +152,11 @@ class EVESpWxTimeSeries(GenericTimeSeries):
 
     References
     ----------
-    * `SDO Mission Homepage <https://sdo.gsfc.nasa.gov/>`_
-    * `EVE Homepage <http://lasp.colorado.edu/home/eve/>`_
-    * `Level 0CS Definition <http://lasp.colorado.edu/home/eve/data/>`_
-    * `EVE Data Acess <http://lasp.colorado.edu/home/eve/data/data-access/>`_
-    * `Instrument Paper <https://doi.org/10.1007/s11207-009-9487-6>`_
+    * `SDO Mission Homepage <https://sdo.gsfc.nasa.gov/>`__
```
nabobalis: I think they all can be just a single underscore.

hayesla: oh I changed it as @Cadair had mentioned it should be __ - which should it be?

hayesla:
> You created a new figure test with `@figure_test` and there is no hash to do the comparison to.
> `E Failed: Hash not present: sunpy.timeseries.tests.test_timeseriesbase.test_esp_peek`
> 
> So you will need to add the hash to the file of hashes located `sunpy/tests/figure_hashes_py36.json`. To work out the hash tho involves running the figure test under the figure env using `tox -e figure`, it should dump out the results into a `tmp` folder in the root of the sunpy repo.

So i ran this but cant seem to figure out where the hash is 🤔

sunpy/timeseries/sources/eve.py
nabobalis: In that case, ignore me.

nabobalis:
I think it saves out the hashes with the figures in the `tmp` folder.

hayesla:
so in my tmp folder there seems to be a sunpydata.sqlite file and a hidden .hypothesis directory with an examples folder full of empty directories - am I missing something or doing something stupid?

nabobalis:
Good question, normally when I run the command in my sunpy repository folder, it saves a folder called `tmp` there (`sunpy/tmp/figure_something/`. I am not sure if is different on a macbook.

hayesla:
oh wait - ran it for a third time and it now is populating my tmp directory 🙉

sunpy/timeseries/sources/eve.py
Cadair: double underscores are generally a better idea for URLs.

hayesla:
hmm there doesn't seem to be a baseline image for comparison? 

maybe why the hash doesn't match?

nabobalis:
Yeah, I'm not sure. I think we can ignore it and if it fails on merge, we can update it separately.

Cadair:
There wont be a comparison image because it hasn't been added to the image repo: https://github.com/sunpy/sunpy-figure-tests which we should do after merging (and also automate!!)

nabobalis:
We could just squash merge it?

Cadair:
Yeah, but there is no way to set that on a PR ahead of time, so I just want to make sure we think about it before pressing the button.

tox.ini
```diff
@@ -29,6 +29,7 @@ commands = sphinx-build docs docs/_build/html -W -b html
 
 # This env requires tox-conda.
 [testenv:figure]
+changedir = {toxinidir}
```
Cadair: @nabobalis I put this here so that when you run the tests locally you get access to the generated files. We don't need the C extensions for this anyway right?

nabobalis: That should be ok but you will need to update the docs, and the circleci path to match.

nabobalis: When I ran it locally, it just made a tmp dir in the sunpy repo folder, it didn't actually use my real tmp volumne.

Cadair:
Wow this PR has pointed out how spectacularly out of date our reference images are :man_facepalming:

nabobalis:
We should update them.

Cadair:
I opened #3033 as a placeholder for automating it, I will manually update them after this PR lands.

Cadair:
@hayesla I pushed a commit here which fixed the figure tests, I assume there was something different on your mac setup that tox wasn't controlling correctly. We should probably look into that at somepoint but it's working now :smile:

hayesla:
Great thanks @Cadair !

PR #3046, opened 2019-04-15
Fix #3045 return an empty HEK table

### LINKED ISSUE #3045: Cryptic error message when querying the HEK with a search that returns no results
### Description
Cryptic error message when querying the HEK with a search that returns no results.

### Expected behavior

Rise a more informative error message or return an empty HEKTable

### Actual behavior

Raises an unexpected error
```ValueError: Inconsistent data column lengths: set()```

### Steps to Reproduce

```
from sunpy.time import TimeRange
from sunpy.net import hek


client = hek.HEKClient()

tstart = '1985-05-03 00:00:00'
tend = '1985-05-04 00:00:00'

result = client.search(hek.attrs.Time(tstart, tend),
                       hek.attrs.EventType('FL'),
                       hek.attrs.OBS.Observatory == 'GOES')
```
### System Details

  - SunPy Version: sunpy-1.0.0.dev1274+gdc0c8cf4f.dist-info (I pip installed from master)
  - Python Version: 3.6.3
  - OS information: macOS 10.14.3

### PR DESCRIPTION
**Pull request by samaloney**
### Description

Check for an empty query result and if found return empty HEKTable instance and test to cover this case.

Fixes #3045

### DIFF
--- a/sunpy/net/hek/hek.py
+++ b/sunpy/net/hek/hek.py
@@ -16,6 +16,7 @@ from itertools import chain
 from astropy.table import Table, Row, Column
 from astropy.time import Time
 
+from sunpy import log
 from sunpy.net import attr
 from sunpy.util import unique
 from sunpy.net.hek import attrs
@@ -73,7 +74,12 @@ class HEKClient(object):
             results.extend(result['result'])
 
             if not result['overmax']:
-                return HEKTable(results)
+                if len(results) > 0:
+                    return HEKTable(results)
+                else:
+                    log.info('Search returned no results')
+                    return HEKTable()
+
             page += 1
 
     def search(self, *query):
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: sunpy/net/hek/hek.py [hunk 1]
```diff
@@ -73,7 +74,12 @@ def _download(self, data):
             results.extend(result['result'])
 
             if not result['overmax']:
-                return HEKTable(results)
+                if len(results) > 0:
+                    return HEKTable(results)
+                else:
+                    log.info('Search returned no results')
```
Cadair: I am not sure if there is value in logging something here, I mean it's obvious there is no results because the table is empty??? :grin:

Cadair:
Thanks @samaloney this looks great.

sunpy/net/hek/hek.py
samaloney: That's true I was trying to be explicit but it probably overkill I'll remove it 👍

nabobalis:
I will merge this and backport it this evening. Thanks @samaloney

nabobalis:
The HEKTable was never backported.

PR #3050, opened 2019-04-16
Squash test warnings in sunpy.database

dstansby:
All just files that are smaller than expected (I guess because they have been manually truncated to be small test files)


Cadair:
Thanks again!!

PR #3052, opened 2019-04-17
Squash timeseries test warnings

dstansby:
Got a bit bored so there's some left, but this should get rid of most of them.


nabobalis:
Thanks again @dstansby

PR #3053, opened 2019-04-17
Add a home for map utility functions

### LINKED ISSUE #3020: A home for functions which operate on Map
We seem to have a couple of things waiting in the wings ([#2972](https://github.com/sunpy/sunpy/pull/2972/files#diff-e5b590f063aafc6198499904a424fc22R547) and #2278) which are a set of functions which provide utilities based on taking a `GenericMap` object (or subclass) as input (or output) and performing some check, or to help people construct a map etc.

 @wafels added a bunch of functions which operate on a map in #2972 the ones relevant for this discussion are:

* `map_edges`
* `contains_full_disk`
* `is_all_off_disk`
* `is_all_on_disk`
* `contains_limb`
* `coordinate_is_on_disk`
* `on_disk_bounding_coordinates`

The other one is a header generation utility as described in #2278, which is a little different in that it doesn't operate on a map but I feel still generally in the same bin as these helpers.

This was discussed at the community meeting (2019-03-27) and it was agreed that we need somewhere in core to put these kind of functions but that we also need to come up with **a strong definition of what the scope of the namespace is so we limit the type of helpers that are in it**.

samaloney:
### Description

Adds a file to contain utility map functions and an example.

Fixes #3020

### DIFF
--- a/sunpy/map/maputils.py
+++ b/sunpy/map/maputils.py
@@ -3,8 +3,5 @@ Holds utility map function which do not belong in map objects
 """
 
 
-__all__ = ['test_function']
+__all__ = []
 
-
-def test_function(map):
-    return map.name
--- a/sunpy/map/maputils.py
+++ b/sunpy/map/maputils.py
@@ -1,5 +1,5 @@
 """
-Holds utility map function which do not belong in map objects
+This submodule provides utility functions to act on `sunpy.map.Map` instances.
 """
 
 
--- a/sunpy/map/maputils.py
+++ b/sunpy/map/maputils.py
@@ -1,0 +1,9 @@
+"""
+Hold utility map function which do not belong in map objects
+"""
+
+
+__all__ = ['test_function']
+
+def test_function(map):
+	return map.name
--- a/sunpy/map/maputils.py
+++ b/sunpy/map/maputils.py
@@ -1,9 +1,10 @@
 """
-Hold utility map function which do not belong in map objects
+Holds utility map function which do not belong in map objects
 """
 
 
 __all__ = ['test_function']
 
+
 def test_function(map):
-	return map.name
+    return map.name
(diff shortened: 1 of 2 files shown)


sunpy/map/maputils.py
```diff
@@ -0,0 +1,10 @@
+"""
+Holds utility map function which do not belong in map objects
+"""
+
+
+__all__ = ['test_function']
+
+
+def test_function(map):
+    return map.name
```
nabobalis: Could you remove this function?

nabobalis:
I forgot if we had decided on what should be moved here. 

I think we can merge this and move functions after the fact.

samaloney:
Yea I think that was the idea but could also wait until the other PRs are merged and then I could move them here?

sunpy/map/maputils.py
```diff
@@ -0,0 +1,7 @@
+"""
+Holds utility map function which do not belong in map objects
+"""
+
+
+__all__ = ['test_function']
```
nabobalis: Will have to remove this until we add more functions.

nabobalis:
Yeah that would make the most sense.

sunpy/map/maputils.py
Cadair: I think it can just be an empty list.

sunpy/map/maputils.py
```diff
@@ -0,0 +1,7 @@
+"""
+Holds utility map function which do not belong in map objects
```
nabobalis:
```suggestion
This submodule provides utility functions to act on `sunpy.map.Map` instances.
```

Cadair:
@samaloney if you get a moment, could you accept Nabil's suggestion and then we will merge this. Thanks

Cadair:
Thanks @samaloney

Issue #3039, opened 2019-04-12
HPC coordinates need to factor in light travel time


ayshih:
Our coordinates code currently ignores any corrections due to light travel time when transforming into helioprojective coordinates (HPC).  It's often "okay" to ignore light travel time for typical observations of the Sun at a distance of 1 AU because in 8 minutes the Sun moves <~0.01 arcseconds in HPC.  However, there are times with light travel time is quite important, such as when determining the apparent position of a planetary body in HPC.

For comparison, Astropy's function `get_body()` – as opposed to Astropy's `get_body_barycentric()` and SunPy's `get_body_heliographic_stonyhurst()` – includes a correction for light travel time, but only for an Earth-based observer.  SunPy ought to have a generalized correction for an arbitrary observer location.

Cadair:
Yeah I was thinking about this with regards to doing a direct HPC <> (H/I)CRS conversion the other day, I don't know if astropy has any helpers for this in ERFA.

ayshih:
(I've edited the first post a little.)

I haven't looked for any helpers in Astropy's code, but given that their `get_body()` has an explicitly coded loop to converge on the corrected position for light travel time for an Earth-based observer, I doubt that they have something more general available elsewhere in their code base.

PR #3055, opened 2019-04-17
Add capability for light-travel-time correction

ayshih:
Addresses #3039 by adding functionality to `get_body_heliographic_stonyhurst()`.  When one specifies an observer through the keyword `observer`, an iterative loop determines the emission time in the past that results in photons arriving at the observation time.  In the following example, Venus's apparent HGS coordinates have been corrected by the 5 Mm that the planet moves in the 144 seconds of light travel time:
```python
>>> from sunpy.coordinates import get_body_heliographic_stonyhurst, get_earth
>>> t = '2012-06-05 22:34:48.350'

>>> without_correction = get_body_heliographic_stonyhurst('venus', t)
>>> print(without_correction)
<HeliographicStonyhurst Coordinate (obstime=2012-06-05T22:34:48.350): (lon, lat, radius) in (deg, deg, AU)
    (359.92620234, 0.02752007, 0.72602872)>

>>> with_correction = get_body_heliographic_stonyhurst('venus', t, observer=get_earth(t))
Apparent location factors in 144.063097622999 seconds of light travel time
>>> print(with_correction)
<HeliographicStonyhurst Coordinate (obstime=2012-06-05T22:34:48.350): (lon, lat, radius) in (deg, deg, AU)
    (359.92355609, 0.02734159, 0.72602853)>

>>> distance_traveled = (with_correction.cartesian - without_correction.cartesian).norm().to('km')
>>> print(distance_traveled)
<Quantity 5027.81720838 km>
```

Only `get_body_heliographic_stonyhurst()` has been modified because corrections for light travel time are usually significant only for planetary bodies, where the transverse angular motion of the body is appreciable relative to the light travel time.  In principle, our coordinate transformations need to take into account the motion of the Sun in barycenter coordinates, but the Sun moves only ~5 km in the ~8 minutes that it takes light to traverse 1 AU, which is a negligible correction (i.e., <0.01 arcseconds in helioprojective coordinates).

~Needs a test.~


sunpy/coordinates/ephemeris.py
```diff
@@ -43,8 +47,25 @@ def get_body_heliographic_stonyhurst(body, time='now'):
...
 
-    body_icrs = ICRS(get_body_barycentric(body, obstime))
-    body_hgs = body_icrs.transform_to(HGS(obstime=obstime))
+    if observer is None:
+        body_icrs = get_body_barycentric(body, obstime)
+    else:
+        observer_icrs = SkyCoord(observer).icrs.cartesian
+
+        # This implementation is modeled after Astropy's `_get_apparent_body_position`
+        light_travel_time = 0.*u.s
```
Cadair:
```suggestion
        light_travel_time = 0. * u.s
```

sunpy/coordinates/ephemeris.py
```diff
@@ -43,8 +47,25 @@ def get_body_heliographic_stonyhurst(body, time='now'):
...
+        emitted_time = obstime
+        delta_light_travel_time = 1.*u.s # placeholder value
```
Cadair:
```suggestion
        delta_light_travel_time = 1. * u.s  # placeholder value
```

sunpy/coordinates/ephemeris.py
```diff
@@ -43,8 +47,25 @@ def get_body_heliographic_stonyhurst(body, time='now'):
...
+        while np.any(np.fabs(delta_light_travel_time) > 1.0e-8*u.s):
+            body_icrs = get_body_barycentric(body, emitted_time)
+            distance = (body_icrs - observer_icrs).norm()
+            delta_light_travel_time = light_travel_time - distance / speed_of_light
+            light_travel_time = distance / speed_of_light
+            emitted_time = obstime - light_travel_time
+
+        print('Apparent location factors in {} seconds of light travel time'.format(light_travel_time.to('s').value))
```
Cadair: This should be logger or @ehsteve will flip out :stuck_out_tongue:

sunpy/coordinates/ephemeris.py
```diff
@@ -43,8 +47,25 @@ def get_body_heliographic_stonyhurst(body, time='now'):
...
+    if observer is None:
+        body_icrs = get_body_barycentric(body, obstime)
+    else:
+        observer_icrs = SkyCoord(observer).icrs.cartesian
+
+        # This implementation is modeled after Astropy's `_get_apparent_body_position`
+        light_travel_time = 0.*u.s
+        emitted_time = obstime
+        delta_light_travel_time = 1.*u.s # placeholder value
+        while np.any(np.fabs(delta_light_travel_time) > 1.0e-8*u.s):
```
Cadair:
```suggestion
        while np.any(np.fabs(delta_light_travel_time) > 1.0e-8 * u.s):
```

sunpy/coordinates/ephemeris.py
ehsteve: Too late!!!!

use
```python
from sunpy import log

log.info(f"Apparent location factors in {light_travel_time.to('s')} of light travel time.')
```

ehsteve:
Tested this and it works!

sunpy/coordinates/ephemeris.py
Cadair: info? I think we should print this by default?

sunpy/coordinates/ephemeris.py
ayshih: Is it our style guidelines to have spaces around the "*" even for units?  I see it done both ways in our codebase.  I prefer not having the spaces for such constants, but I'll do whatever is our guideline.

Cadair: Yeah, I used to not but then my linter nagged me into it. I wonder if @nabobalis has legislated for this.

nabobalis: I'm not sure it's an internal rule but I think it's a PEP8 issue? I'm sure flake8 tells me to put spaces.

Cadair: It might be worth making an explicit note for units, because as @ayshih points out sometimes it feels weird to put the spaces around the operator before units.

nabobalis: > Thus the PEP, in its current form, does not dictate whether or not you should use spaces around the + operator (or other arithmetic operators like * and / and **). You are free to "use your own judgement".

So we should pick one?

Cadair: Ok, I think we should move this to a new issue.

ayshih:
I added a test.

PR #3056, opened 2019-04-17
Make AIAMap use the HAE coordinates in the header

### LINKED ISSUE #3038: AIA FITS headers have inaccurate HGS coordinates
AIA FITS headers apparently have inaccurate Heliographic Stonyhurst (HGS) coordinates (`HGLN_OBS` and `HGLT_OBS`).  For example, the distance from Earth center does not match the orbit radius.  We currently use these keywords to generate the `observer_coordinate` for an AIA map, so we provide an inaccurate observer location.

The headers also have Heliographic Aries Ecliptic (HAE) coordinates (`HAEX_OBS`, `HAEY_OBS`, and `HAEZ_OBS`), and the HAE coordinates are inconsistent with the HGS coordinates in the same header.  We have previously verified the accuracy of SunPy's transformation from HAE to HGS (e.g., https://github.com/sunpy/sunpy/issues/2445#issuecomment-364531159).   The HAE coordinates appear to be credible, and likely should be trusted over the HGS coordinates.

My recommendation is for `AIAMap` to override the generation of `observer_coordinate` to use the HAE coordinates.  Discuss.

ayshih:
Fixes #3038 by overriding the properties in `AIAMap` to use the HAE coordinates in the header rather than the HGS coordinates.

No action is currently taken with respect to the HGS coordinates (and HGC coordinates) in the header (`HGLN_OBS` et al.).  Possible actions include:
- Do nothing (i.e., retain the inaccurate HGS coordinates in the metadata despite the inconsistency)
- Remove the HGS coordinates from the metadata to have a "clean" header
- Modify the HGS coordinates in the metadata to be more accurate


sunpy/map/sources/sdo.py
```diff
@@ -47,6 +49,43 @@ def __init__(self, data, header, **kwargs):
         self.plot_settings['cmap'] = plt.get_cmap(self._get_cmap_name())
         self.plot_settings['norm'] = ImageNormalize(stretch=source_stretch(self.meta, AsinhStretch(0.01)))
 
+    @property
+    def observer_coordinate(self):
+        """
+        The Heliographic Stonyhurst Coordinate of the observer.
+        The HAE coordinates in the header are used instead of the HGS coordinates in the header.
```
Cadair: Probably worth saying why

Cadair: Add a line break 
```suggestion

        The HAE coordinates in the header are used instead of the HGS coordinates in the header.
```

sunpy/map/sources/sdo.py
```diff
@@ -47,6 +49,43 @@ def __init__(self, data, header, **kwargs):
...
+        """
+        vector = CartesianRepresentation(self.meta['haex_obs'],
+                                         self.meta['haey_obs'],
+                                         self.meta['haez_obs'])
+        coord = SkyCoord(vector * u.m, frame=HeliocentricTrueEcliptic, obstime=self.date)
+        return coord.heliographic_stonyhurst
+
+    @property
+    def heliographic_latitude(self):
+        """Heliographic latitude (calculated from HAE coordinates)."""
```
Cadair: I think it could be better if we put this comment in the `observer_coordinate` docstring and not in all of these?

Cadair:
My default position on the HGS keys is to do nothing. I am open to counter arguments though.

ayshih:
Need to fix this to use `HeliocentricMeanEcliptic` instead of `HeliocentricTrueEcliptic` for Astropy 3.2 (see #3066 and #3075)

PR #3059, opened 2019-04-19
Deprecation warning squash in resample.py

dstansby:


nabobalis:
Thanks again @dstansby

PR #3062, opened 2019-04-19
fixed mask colour

### LINKED ISSUE #3007: "Masking the solar disk" gallery example does not have a black mask

ravindra1307:
### Description

Fixes #3007 <Issue Number>


nabobalis:
Thanks @ravindra1307 but it seems there is an error in the code:

```
norm = mpl.colors.Normalize(vmin=-3000, vmax=4000)
NameError: name 'mpl' is not defined
```

nabobalis:
Hi @ravindra1307, the conversation on the original issue seems to suggest a fix that is upstream in astropy. As a result, I am not sure this fix would work and we would need to tackle this in another way. So I will close this for now. Please feel free to tackle another issue.

PR #3063, opened 2019-04-19
Fix two bugs in scraper code.

### LINKED ISSUE #1954: Transient test failure in scraper code
This seems to happen in Travis builds from time to time.

```python
=================================== FAILURES ===================================
_________________________ testURL_patternMilliseconds __________________________
    def testURL_patternMilliseconds():
        s = Scraper('fd_%Y%m%d_%H%M%S_%e.fts')
>       assert s._URL_followsPattern('fd_20130410_231211_119.fts')
E       assert False
E        +  where False = <bound method Scraper._URL_followsPattern of <sunpy.util.scraper.Scraper object at 0x7fdfda7c0400>>('fd_20130410_231211_119.fts')
E        +    where <bound method Scraper._URL_followsPattern of <sunpy.util.scraper.Scraper object at 0x7fdfda7c0400>> = <sunpy.util.scraper.Scraper object at 0x7fdfda7c0400>._URL_followsPattern
sunpy/util/tests/test_scraper.py:110: AssertionError
============================= 121 tests deselected =============================
============ 1 failed, 935 passed, 121 deselected in 62.03 seconds =============
```

### LINKED ISSUE #2684: Scraper errors when directory in pattern doesn't exist
i.e. when you have a pattern like `http://mywebsite/%Y/%Y%m%d` if you ask for a year that doesn't exist it will fail because it will try and get the directory.

Brhstone:
Fixed parser failing due to missing directory (#2684) and when milliseconds were less than 100 (#1954).

### Description

Addressed two issues:
* #2684 Fixed by catching the HTTP error and ignoring it if the error code was 404 (Not Found). Added a test asserting the fix.
* #1954 Fixed the randomly failing test by addressing the underlying bug -- milliseconds were not being zero-padded. The `_URL_followsPattern` logic relies on URL length, which changes if zero-padding isn't used. The randomness was due to the random chance of the current milliseconds time being less than 100 (_derived from microseconds_). Added a test asserting the fix by using a mocked datetime of the edge case. I was not able to get the two other tests mentioned in the ticket to fail.

Fixes #2684 and fixes #1954

### DIFF
--- a/sunpy/util/tests/test_scraper.py
+++ b/sunpy/util/tests/test_scraper.py
@@ -166,8 +166,6 @@ def testFilesRange_sameDirectory_local():
     # tries to open a directory as a binary file.
     s = Scraper('/'.join(['file:/', rootdir,
                           'EIT', 'efz%Y%m%d.%H%M%S_s.fits']))
-    print(s.pattern)
-    print(s.now)
     startdate = parse_time((2004, 3, 1, 4, 0))
     enddate = parse_time((2004, 3, 1, 6, 30))
     assert len(s.filelist(TimeRange(startdate, enddate))) == 3
--- a/sunpy/util/tests/test_scraper.py
+++ b/sunpy/util/tests/test_scraper.py
@@ -1,3 +1,6 @@
+import datetime
+from unittest.mock import patch, Mock
+
 import pytest
 
 import astropy.units as u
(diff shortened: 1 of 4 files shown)


sunpy/util/tests/test_scraper.py
```diff
@@ -140,18 +143,27 @@ def testURL_pattern():
...
+    with patch('datetime.datetime', now=now_mock):
+        s = Scraper('fd_%Y%m%d_%H%M%S_%e.fts')
+    now_mock.assert_called_once()
+    assert s.now == 'fd_20190419_000000_004.fts'
+
+
 @pytest.mark.xfail
 def testFilesRange_sameDirectory_local():
+    # Fails due to an IsADirectoryError, wrapped in a URLError, after `requests`
+    # tries to open a directory as a binary file.
```
nabobalis: So this is something we can fix or should this test actually always fail?

Brhstone: Excellent question. I don't know if scanning directories was a functioning feature at some point or not. It's definitely fixable/implementable, but it'd take some effort.

nabobalis: Yeah, that does seem like an undertaking. I wonder if we want to support that in the future or just drop it. What do you think @Cadair?

Cadair: I honestly have no idea. If there isn't an issue for this already we should probably open one. Also I am guessing @dpshelio wrote this test.

Cadair:
Thanks for the patch @Brhstone :smile: 

I think this is good to go, if you can add a changelog entry (one per bug fix) as per the instructions in the [changelog readme](https://github.com/sunpy/sunpy/blob/master/changelog/README.rst) that would be awesome.

sunpy/util/tests/test_scraper.py
samaloney: There is already a check for `ftp` type urls (uris? I never know which) maybe add a check for `file` type urls and raise not implemented exception and alter the test to match? I mean it can't work as as it is?

sunpy/util/tests/test_scraper.py
```diff
@@ -140,18 +143,27 @@ def testURL_pattern():
...
     s = Scraper('/'.join(['file:/', rootdir,
                           'EIT', 'efz%Y%m%d.%H%M%S_s.fits']))
     print(s.pattern)
```
samaloney: I know these print statements were not added in this PR but they don't belong, remove?

nabobalis: Probably just remove it when the test gets updated/fixed.

PR #3064, opened 2019-04-20
squashed AstropyDeprecationWarning in vso.attrs.Wavelength

### LINKED ISSUE #3058: vso.attrs.Wavelength raises AstropyDeprecationWarning
Can't work out how to fix this at the moment, so leaving a bug as a placeholder:
```python
from sunpy.net import vso
import astropy.units as u
vso.attrs.Wavelength(12 * u.AA, 16 * u.AA)
```
results in 
```
WARNING: AstropyDeprecationWarning: The truth value of a Quantity is ambiguous. In the future this will raise a ValueError. [astropy.units.quantity]
```

nabobalis:
Fixes #3058

### DIFF
--- a/sunpy/net/vso/attrs.py
+++ b/sunpy/net/vso/attrs.py
@@ -18,7 +18,6 @@ Instrument('aia') & Instrument('eit').
 import collections
 
 import astropy.units as u
-from astropy.time import Time
 
 from sunpy.time import TimeRange as _TimeRange
 from sunpy.net.attr import (Attr, AttrWalker, AttrAnd,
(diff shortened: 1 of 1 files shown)


sunpy/net/vso/attrs.py
```diff
@@ -18,7 +18,6 @@
 import collections
 
 import astropy.units as u
-from astropy.time import Time
```
nabobalis: So this change I'm unsure about. Within the `Time` attr, the method `pad` returns a `Time`but I assume this must be a new `Time` attr and not astropy.time.Time instance. If that is the case, then we don't need this import.

nabobalis:
Online build failure is database doc example complaining about the database being locked.

sunpy/net/vso/attrs.py
Cadair: This import would get clobbered by the `Time` attr anyway right? I think you should be good here.

nabobalis: It's what the linter said should happen.

PR #3065, opened 2019-04-20
fixed mask_color

### LINKED ISSUE #3007: "Masking the solar disk" gallery example does not have a black mask

ravindra1307:
### Description

Fixes #3007<Issue Number>

### DIFF
--- a/examples/computer_vision_techniques/mask_disk.py
+++ b/examples/computer_vision_techniques/mask_disk.py
@@ -42,6 +42,9 @@ r = np.sqrt(hpc_coords.Tx ** 2 + hpc_coords.Ty ** 2) / aia.rsun_obs
 mask = ma.masked_less_equal(r, 1)
 palette = aia.plot_settings['cmap']
 palette.set_bad('black')
+
+###############################################################################
+# Now we normalize our image to brighten edges 
 norm = mpl.colors.Normalize(vmin=-1000, vmax=3000)
 ###############################################################################
 # Now we create a new custom aia with our new mask and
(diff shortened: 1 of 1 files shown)


examples/computer_vision_techniques/mask_disk.py
```diff
@@ -41,7 +42,7 @@
 mask = ma.masked_less_equal(r, 1)
 palette = aia.plot_settings['cmap']
 palette.set_bad('black')
-
+norm = mpl.colors.Normalize(vmin=-1000, vmax=3000)
```
nabobalis: Can you add a comment here as to why we had to add this line?

ravindra1307: i have added comment, i have tried for this image only for others i think we have to experiment with different values

examples/computer_vision_techniques/mask_disk.py
```diff
@@ -42,6 +43,9 @@
 palette = aia.plot_settings['cmap']
 palette.set_bad('black')
 
+###############################################################################
```
nabobalis: I don't feel this comment explains enough, why is it negative, why this range and etc.

examples/computer_vision_techniques/mask_disk.py
nabobalis: That's a problem. We can't expect our users to have to manually change the norm in some (what feels like random) way to get the mask to work. It's ok for this example but I don't think we can count it as a fix for the original issue.

nabobalis:
Sorry again @ravindra1307 but having looked at this again, I don't think that making this change is a good idea. 

We need to change the underlying setting somehow instead of asking our users to change the normalization like that.

Issue #3066, opened 2019-04-22
Astropy dev test failure for HGS->HCRS. 


nabobalis:
Currently with the latest version of astropy master. 

```python
def test_hgs_hcrs():
        # This test checks the HGS->HCRS transformation by transforming from HGS to
        # HeliocentricTrueEcliptic (HTE).  It will fail if there are errors in Astropy's
        # HCRS->ICRS or ICRS->HTE transformations.
    
        # Use published HGS coordinates in the Astronomical Almanac (2013), pages C6-C7
        obstime = Time('2013-01-28')
        earth_hgs = SkyCoord(0*u.deg, -5.73*u.deg, 0.9848139*u.AU, frame=HeliographicStonyhurst,
                             obstime=obstime)
    
        # Transform to HTE at observation-time equinox
        earth_hte = earth_hgs.transform_to(HeliocentricTrueEcliptic(equinox=obstime))
    
        # Validate against published values from the Astronomical Almanac (2013), page C6 per page E2
        # The dominant source of inaccuracy is the limited precision of the published B0 used above
>       assert quantity_allclose(earth_hte.lon, Angle('308d13m30.51s') - 180*u.deg, atol=5*u.arcsec)
E       assert False
E        +  where False = quantity_allclose(<Longitude 128.23007468 deg>, (<Angle 308.22514167 deg> - (180 * Unit("deg"))), atol=(5 * Unit("arcsec")))
E        +    where <Longitude 128.23007468 deg> = <SkyCoord (HeliocentricTrueEcliptic: equinox=2013-01-28 00:00:00.000, obstime=2013-01-28 00:00:00.000): (lon, lat, distance) in (deg, deg, AU)\n    (128.23007468, 0.00023427, 0.9848139)>.lon
E        +    and   <Angle 308.22514167 deg> = Angle('308d13m30.51s')
E        +    and   Unit("deg") = u.deg
E        +    and   Unit("arcsec") = u.arcsec
```

Need to figure out what's changed and if it's ok.

Cadair:
ping @ayshih I will have a go at bisecting this later.

Cadair:
Is this also failing on 0.9?

ayshih:
<whimper>

nabobalis:
Not a clue if it happens on 0.9.

Cadair:
[...]

Issue #3057, opened 2019-04-18
HMI FITS headers may not have accurate observer coordinates


ayshih:
HMI FITS headers appear to have only heliographic Carrington (HGC) coordinates for the observer (SDO).  These appear to correspond very closely with the HGC coordinates in AIA FITS headers at the same time.  However, SunPy does not agree with the HGC longitude in the AIA FITS header at the level of arcminutes (see #3038), and part of the discrepancy is likely due to a slightly different definition of HGC longitude (see #2224).

Because the observer location for AIA files is determined using other coordinates than the HGC coordinates, it currently will not match the observer location for HMI files.

Cadair:
I assume HMI dosen't have the same headers as AIA for the other observer coordinates?

ayshih:
Here's a super-relevant comment from the HMI folks: https://github.com/sunpy/sunpy/issues/3735#issuecomment-582487046

wtbarnes:
We can close this right? If I recall, this was addressed by the HMI team.

wtbarnes:
I'm going to close this issue for the reasons stated above. Feel free to reopen if there's something we still haven't resolved.

Issue #2224, opened 2017-08-03
Resolve different ways to calculate Carrington longitude


ayshih:
Now that PR #2186 is merged, our ephemeris calculations have accurate correspondence with published sources *except* for arcmin-level discrepancies with Carrington longitudes provided by the _Astronomical Almanac_.  Here are my comments from PR #2186 (starting [here](https://github.com/sunpy/sunpy/pull/2186#discussion_r129573004)):

> I've mainly used two references for verification. I see a discrepancy of up to ~2 arcmin with the Carrington longitudes published in the Astronomical Almanac (2013 edition, at least, I haven't checked other years yet), but curiously I have an exact match to a Carrington longitude calculated in Astronomical Algorithms by Meeus, even though this code doesn't include any implementation of Meeus's equations. I'm not currently sure what to make of this discrepancy, but I wouldn't be surprised if there are multiple, inconsistent definitions of Carrington longitude.

> Actually, now that I think more about it, the issue is probably simply the exact start time of the first Carrington rotation. I use the same time that Meeus uses (JD 2398167.4), hence the exact match isn't surprising, while the Astronomical Almanac doesn't (explicitly) state what it uses.

Currently, `sunpy.coordinates.ephemeris.get_sun_L0()` effectively propagates from the first Carrington rotation in 1853 (a la Meeus).  In contrast, Thompson (2006) provides a definition referenced to J2000.0:

> Seidelmann et al. (2002) gives the angle of the prime meridian from the ascending node as 84.10 deg at J2000.0.

Switching to the latter would probably improve the agreement with non-Meeus sources, but at the expense of breaking agreement with Meeus.  With @dpshelio's [comment](https://github.com/sunpy/sunpy/pull/2186#discussion_r130982132) that we also have similar-magnitude discrepancies with HELIO (which presumably agrees better with the _Astronomical Almanac_), the agreement with Meeus may not be worth keeping.

Resolve this.

ayshih:
I decided to break the correspondence with Meeus, in favor of agreement with all modern sources.

PR #3067, opened 2019-04-24
Attempt to correctly read observer locations for MDI and EIT

### LINKED ISSUE #2430: Map not recognizing EIT coordinates.
Something is messed up with the EIT map coordinates. The meta data and WCS looks OK at first glance, but the coordinate grid isn't being translated properly. 

`from sunpy.map import Map`
`eit=Map(sunpy.data.sample.EIT_195_IMAGE)`
`eit.peek()`

eit_example

### LINKED ISSUE #3040: MDI observer longitude not properly extracted from header
The MDI FITS header has the observer Carrington longitude under the keyword `OBS_L0` (and `L0`) rather than `CRLN_OBS`, so our code currently misses this information and assumes an Earth-based observer.  We shouldn't use this value blindly, however, because I have reasons to be skeptical about Carrington-longitude calculations.  The good news is that the FITS header *also* has the Earth Carrington longitude under the keyword `EARTH_L0`, and the *difference* of the two longitudes should be an accurate observer Stonyhurst longitude.

Cadair:
### Description

This is inspired by #3056 and conversations around it. This changes the SoHo sources so they do not actually modify the header, but implement the properties needed to override things.

Fixes #2430 and fixes #3040 

This now builds on #3115

### DIFF
--- a/changelog/3067.bugfix.rst
+++ b/changelog/3067.bugfix.rst
@@ -1,0 +1,1 @@
+Correctly extract observer location from MDI and EIT data
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -441,7 +441,10 @@ class GenericMap(NDData):
     @property
     def waveunit(self):
         """The `~astropy.units.Unit` of the wavelength of this observation."""
-        return u.Unit(self.meta.get("waveunit", ""))
+        unit = self.meta.get("waveunit")
+        if unit is None:
+            return u.one
+        return u.Unit(unit)
 
     @property
     def wavelength(self):
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -436,13 +436,18 @@ class GenericMap(NDData):
     def measurement(self):
         """Measurement name, defaults to the wavelength of image."""
         return u.Quantity(self.meta.get('wavelnth', 0),
-                          self.meta.get('waveunit', ""))
+                          self.waveunit)
+
+    @property
+    def waveunit(self):
+        """The `~astropy.units.Unit` of the wavelength of this observation."""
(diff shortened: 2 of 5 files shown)


ehsteve:
See issue #3070.

ayshih:
Need to fix this to use HeliocentricMeanEcliptic instead of HeliocentricTrueEcliptic for Astropy 3.2 (see #3066 and #3075)

ayshih:
Your MDI implementation doesn't fix #3040.  It looks like you've put their value for Carrington longitude where the Stonyhurst longitude should go.

I believe our options are to trust their Carrington longitude (which I don't like because I can't match Stonyhurst-Carrington transformations exactly), we try to calculate a Stonyhurst longitude by subtracting two header values (which at first glance doesn't give accurate results), or we kludge it by saying that the Stonyhurst longitude is zero and document that kludge.

Cadair:
@ayshih I went with the convert to stonyhurst approach. It seems the least kludgy? We already actually do that in mapbase by default: https://github.com/sunpy/sunpy/blob/master/sunpy/map/mapbase.py#L606 which is maybe something we might want to consider?! (HMI uses it as they don't have HGS in the header)

ayshih:
I think you made the right choice, but it's an uncomfortable choice since I still don't trust HGC coordinates (see #2224).  For cross-reference purposes, #3057 contains my anxiety over HMI.

Cadair:
I think this test fail might actually be real.

Cadair:
@ayshih It seems that the distance for EIT isn't coming out right by doing this, which is causing the fails in the differential rotation code.

ayshih:
@Cadair I've pushed the fix

Cadair:
oh dear :woman_facepalming: oh dear oh dear oh dear.

Issue #2278, opened 2017-08-23
Provide easy method/workflow to generate a map header


ehsteve:
Currently there is no good way for a user to know what should be or must be provided in a map header dictionary. The current process is cumbersome. This issue is to, at minimum, provide a convenience function to do this. At best we should provide a meta class object which solves all of our problems.

Cadair:
I think the gen-0 version of this should just be a utility function (probably in `sunpy.utils`?) which outputs a `MetaDict` with the following keys (values provided for examples):
```
meta = MetaDict({
    'ctype1': 'HPLN-TAN',
    'ctype2': 'HPLT-TAN',
    'cunit1': 'arcsec', 'cunit2': 'arcsec',
    'crpix1': (data.shape[0] + 1)/2., 'crpix2': (data.shape[1] + 1)/2.,
    'cdelt1': 1.0, 'cdelt2': 1.0, 'crval1': 0.0, 'crval2': 0.0,
    'hgln_obs': 0.0,
    'hglt_obs': 0.0,
    'dsun_obs': const.au.to(u.m).value,
    'dsun_ref': const.au.to(u.m).value,
    'rsun_ref': const.R_sun.to(u.m).value,
    'rsun_obs': ((const.R_sun / const.au).decompose()
                 * u.radian).to(u.arcsec).value,
    't_obs': time_now.iso,
    'date-obs': time_now.iso,
})
```
it should be sufficient to pass this utility function, a SkyCoord for the observer from which hgln/lt_obs, dsun_obs, rsun_obs, t_obs, date-obs could be extracted. Then it should also take a Quantity for Cdelt and crval (defaulting to 0,0) and then the shape of the array and crpix should be optional defaulting to the centre of the array.

Cadair:
@Alex-Ian-Hamilton has done a bit of research into required metadata as well here https://github.com/sunpy/sunpy/issues/1832#issuecomment-467987877

Cadair:
I am going to deem this closed by #3083

PR #3083, opened 2019-04-30
Helper map header function 

hayesla:
### Description
A first pass at creating helper functionality for creating a Map header from Issue #2278

At the moment I've put it in ```maputils.py``` as ```meta_helper``` class object. 
The object can be called through ```map.meta_helper``` . 

It currently print the accepted meta keywords that is used in ```GenericMap```  to create a map, and describes what each keyword means.

For example
```
>>> from sunpy import map
>>> map.meta_helper.acceptable_meta_keywords()
cunit1  :  Units of the coordinate increments along naxis1 e.g. arcsec **required
cunit2  :  Units of the coordinate increments along naxis2 e.g. arcsec **required
crval1  :  Coordinate value at reference point on naxis1 **required
crval2  :  Coordinate value at reference point on naxis2 **required
cdelt1  :  Spatial scale of pixels for naxis1, i.e. coordinate increment at reference point
cdelt2  :  Spatial scale of pixels for naxis2, i.e. coordinate increment at reference point
.....
```
It can also check if a meta keyword is accepted 
```
>>> map.meta_helper.acceptable_meta_keywords('crpix1')
crpix1  :  Pixel coordinate at reference point naxis1
>>> map.meta_helper.acceptable_meta_keywords('lalala')
lalala  is not in the accepted meta data required for sunpy.map.Map 
```
The class also has a utility function that can create a map header, stored in a `MetaDict`. This function can either take no keywords, and defaults to an Earth-based observer at current time (assuming cunit = arcsec), or can take a `SkyCoord` or `sunpy.coordinates.frame` as a reference point from which ctype, cunit, hglt/ln_obs, dsun_ref, rsun_ref and obstime can be pulled. It then can take optional crval, cdelt and crpix which are set to default 0, 1, center of data array if none are given.
  
For example

```
>>> from sunpy import map
>>> from sunpy.coordinates import frames
>>> from astropy import units as u
>>> import numpy as np

>>> data = np.random.rand(1024, 1024)
>>> header = map.meta_helper.make_header(data)
>>> header
MetaDict([('ctype1', 'HPLN-TAN'),
          ('ctype2', 'HPLT-TAN'),
          ('cunit1', 'arcsec'),
          ('cunit2', 'arcsec'),
          ('crpix1', 512.5),
          ('crpix2', 512.5),
          ('cdelt1', 1),
          ('cdelt2', 1),
          ('crval1', 0),
          ('crval2', 0),
          ('hgln_obs', 0.0),
          ('hglt_obs', 0.0),
          ('date_obs', '2019-04-30 19:29:47.129'),
          ('t_obs', '2019-04-30 19:29:47.129'),
          ('dsun_obs', 149597870700.0),
          ('rsun_obs', 959.2277285408243),
          ('rsun_ref', 695700000.0)])
```

or a `SkyCoord` or `frame` can be passed such that
```
>>> d = SkyCoord(0*u.arcsec, 0*u.arcsec, obstime="2017-08-01", observer = 'earth', frame=frames.Helioprojective)
>>> header = map.meta_helper.make_header(data, d)
>>> header
MetaDict([('ctype1', 'HPLN-TAN'),
          ('ctype2', 'HPLT-TAN'),
          ('cunit1', Unit("arcsec")),
          ('cunit2', Unit("arcsec")),
          ('crpix1', 512.5),
          ('crpix2', 512.5),
          ('cdelt1', 1),
          ('cdelt2', 1),
          ('crval1', 0),
          ('crval2', 0),
          ('hgln_obs', <Longitude 0. deg>),
          ('hglt_obs', <Latitude 5.78339799 deg>),
          ('date_obs',
           <Time object: scale='utc' format='isot' value=2017-08-01T00:00:00.000>),
          ('t_obs',
           <Time object: scale='utc' format='isot' value=2017-08-01T00:00:00.000>),
          ('dsun_obs', <Distance 1.51837236e+11 m>),
          ('rsun_obs', 945.0805956220959),
          ('rsun_ref', <Quantity 6.957e+08 m>)])
```
This is based on astropy.wcs.utils.celestial_frame_to_wcs. additonaly kwargs can be passed through `make_map` to define the metadict keywords

```
>>> header = map.meta_helper.make_header(data, d, crval = [100,100], cdelt = [10,10], exptime = 2, telescop = 'my_telescope', crota2 = -0.15)
```

I dont know if this is the type of functionality we want but it's a start I guess!


sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+import random
+import matplotlib.pyplot as plt
+import astropy
+from sunpy.coordinates import frames
 
+__all__ = ['meta_helper']
 
-__all__ = []
 
+class map_helper:
```
wafels: CamelCase for classes, i.e, MapHelper

changelog/3083.feature.rst
```diff
@@ -0,0 +1 @@
+Add a helper utility for creating a `sunpy.map.Map` header
```
nabobalis: How about:
```suggestion
Added a helper function that allows users to create a meta header for custom created `sunpy.map.Map`.
```

I think we want to expand on what its for and just to hopefully make it stand out in the changelog!

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+import random
+import matplotlib.pyplot as plt
+import astropy
+from sunpy.coordinates import frames
 
+__all__ = ['MapHelper']
 
-__all__ = []
 
+class MapHelper:
```
nabobalis: Does this need to be a class and not just a function?

ehsteve:
@hayesla I think this is a great start! I would also recommend that this PR come with a guide (in addition to the normal documentation).

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+    """
+    A helper class for knowing what is required in a `sunpy.map.Map` header.
+    A `sunpy.map.Map` header consists of a dictionary that contains meta information
```
nabobalis: A line break above here.

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+    about the map data. This meta is generally found in the FITS header of an image
```
nabobalis:
```suggestion
    about the map data. This meta information is generally found in the FITS header of an image
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+    file. This helper is defined to assist the user in finding out what meta data
+    keywords are accepted in a map header, and can produce a header MetaDict from a
```
nabobalis:
```suggestion
    keywords are accepted in a map header, and can produce a header `sunpy.util.meta.MetaDict from a 
```

I forgot where MetaDict is actually stored. That might not be correct in the slightest.

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
-__all__ = []
 
+class MapHelper:
+    """
+    A helper class for knowing what is required in a `sunpy.map.Map` header.
+    A `sunpy.map.Map` header consists of a dictionary that contains meta information
+    about the map data. This meta is generally found in the FITS header of an image
+    file. This helper is defined to assist the user in finding out what meta data
+    keywords are accepted in a map header, and can produce a header MetaDict from a 
+    coordinate object, or through default meta that can be then passed into `Map`
```
nabobalis:
```suggestion
    coordinate object, or through default meta that can be then passed into `sunpy.map.Map` 
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
 
+class MapHelper:
+    """
+    A helper class for knowing what is required in a `sunpy.map.Map` header.
+    A `sunpy.map.Map` header consists of a dictionary that contains meta information
+    about the map data. This meta is generally found in the FITS header of an image
+    file. This helper is defined to assist the user in finding out what meta data
+    keywords are accepted in a map header, and can produce a header MetaDict from a 
+    coordinate object, or through default meta that can be then passed into `Map` 
+    to create a Map object.
```
nabobalis:
```suggestion
    to create a map.
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+
+    `map_helper.acceptable_meta_keywords()` prints the accepted meta that is used in
```
nabobalis:
```suggestion
    `sunpy.map.MapHelper.acceptable_meta_keywords()` prints the accepted meta that is used in 
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+    A helper class for knowing what is required in a `sunpy.map.Map` header.
+    A `sunpy.map.Map` header consists of a dictionary that contains meta information
+    about the map data. This meta is generally found in the FITS header of an image
+    file. This helper is defined to assist the user in finding out what meta data
+    keywords are accepted in a map header, and can produce a header MetaDict from a 
+    coordinate object, or through default meta that can be then passed into `Map` 
+    to create a Map object.
+
+    `map_helper.acceptable_meta_keywords()` prints the accepted meta that is used in 
+    creating a sunpy.map.Map
```
nabobalis:
```suggestion
    creating a `sunpy.map.Map`.
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+
+    `map_helper.make_header(data)` will create a map header for the numpy.ndarray data,
```
nabobalis:
```suggestion
    `sunpy.map.MapHelper.make_header(data)` will create a map header for the `numpy.ndarray` data, 
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+    file. This helper is defined to assist the user in finding out what meta data
+    keywords are accepted in a map header, and can produce a header MetaDict from a 
+    coordinate object, or through default meta that can be then passed into `Map` 
+    to create a Map object.
+
+    `map_helper.acceptable_meta_keywords()` prints the accepted meta that is used in 
+    creating a sunpy.map.Map
+
+    `map_helper.make_header(data)` will create a map header for the numpy.ndarray data, 
+    either with default meta keywords, or a SkyCoord or frame can be passed to extract
```
nabobalis:
```suggestion
    either with default meta keywords, or a `astropy.coordinates.SkyCoord` or frame can be passed to extract
```

I am not sure if that is right either. I forgot where SkyCoord lives.

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+    reference meta
```
nabobalis:
```suggestion
    reference meta.
```

Does meta refer to "meta information" or to something else?

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+
+    Examples
+    --------
+    Printing meta keywords:
+
+        * print the accepeted meta keywords that are used in a sunpy.map.Map header
```
nabobalis:
```suggestion
        * Print the accepted meta keywords that are used in a `sunpy.map.Map` header:
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+          >>> from sunpy import map
+          >>> map.MapHelper.acceptable_meta_keywords() 
+          cunit1  :  Units of the coordinate increments along naxis1 e.g. arcsec **required
+          ...
+          CD2_2  :  Matrix element CDi_j describing the rotation required to align solar North with the top of the image.
+
+        * check is a meta keyword is acceptable and print information about it
```
nabobalis:
```suggestion
        * Check is a meta keyword is acceptable and print information about it:
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+          >>> map.MapHelper.acceptable_meta_keywords('exptime')
+          exptime is an accepted keyword  :  exposure time of observation, in seconds e.g 2
+
+          >>> map.MapHelper.acceptable_meta_keywords('keyword_test')
+          keyword_test  is not in the accepted meta data required for sunpy.map.Map 
+    
+    Creating headers from map_helper:
+
+        * Create a default header with meta data assumed to be Earth-based observer at current time and arcsec units
```
nabobalis:
```suggestion
        * Create a default header with meta data assumed to be Earth-based observer at current time and arcsec units:
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+          >>> from sunpy.coordinates import frames
+          >>> from astropy.coordinates import SkyCoord
+          >>> from astropy import units as u
+          >>> import numpy as np
+
+          >>> data = np.random.rand(1024, 1024)
+          >>> header = map.MapHelper.make_header(data)
+          >>> my_map = map.Map(data, header)
+        
+        * Create a header from a reference SkyCoord object
```
nabobalis:
```suggestion
        * Create a header from a reference SkyCoord object:
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+          >>> my_coord = SkyCoord(100*u.arcsec, 100*u.arcsec, obstime = '2013-10-28 01:00', frame = frames.Helioprojective)
+          >>> header = map.MapHelper.make_header(data, my_coord)
+          >>> my_map = map.Map(data, header)
+
+
+
```
nabobalis:
```suggestion
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+          >>> data = np.random.rand(1024, 1024)
+          >>> header = map.MapHelper.make_header(data)
+          >>> my_map = map.Map(data, header)
+        
+        * Create a header from a reference SkyCoord object
+          >>> my_coord = SkyCoord(100*u.arcsec, 100*u.arcsec, obstime = '2013-10-28 01:00', frame = frames.Helioprojective)
+          >>> header = map.MapHelper.make_header(data, my_coord)
+          >>> my_map = map.Map(data, header)
+
+
```
nabobalis:
```suggestion
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+          >>> data = np.random.rand(1024, 1024)
+          >>> header = map.MapHelper.make_header(data)
+          >>> my_map = map.Map(data, header)
+        
+        * Create a header from a reference SkyCoord object
+          >>> my_coord = SkyCoord(100*u.arcsec, 100*u.arcsec, obstime = '2013-10-28 01:00', frame = frames.Helioprojective)
+          >>> header = map.MapHelper.make_header(data, my_coord)
+          >>> my_map = map.Map(data, header)
+
```
nabobalis:
```suggestion
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+
+
+    """
+    def acceptable_meta_keywords(self, *key):
+        """
+        prints the accepted metadata keywords that are used when
```
nabobalis:
```suggestion
        Prints the accepted metadata keywords that are used when
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+                else:
+                    print(k, ' is not in the accepted meta data required for sunpy.map.Map \n The accepted keywords are : \n', list(acceptable_meta.keys()))
+            return
+
+        for k in acceptable_meta:
+            print(k, ' : ', acceptable_meta[k])
+
+    def make_header(self, data, *coordinate, crval=None, crpix=None, cdelt=None, **kwargs):
+        """
+        Function to create a MetaDict from a coordinate object (SkyCoord or coordinate frame)
```
nabobalis:
```suggestion
        Function to create a `sunpy.util.meta.MetaDict` from a coordinate object (`astropy.coordinates.SkyCoord` or coordinate frame)
```

Have to check those paths first.

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+        that is required to create a `sunpy.map.Map`
+
+        Parameters
+        ----------
+        data : `~numpy.ndarray`
+            array data of Map for which a header is requried
+
```
nabobalis:
```suggestion
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+        coordinates : `~astropy.coordinates.SkyCoord` or `~astropy.coordinates.BaseFrame`, optional if not set default parameters are set
+            coordinate object to get meta information for map header
+
```
nabobalis:
```suggestion
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+        crval : array or list of size of 2, default (0, 0)
+            coordinate of reference pixel along each axis
+            first index is crval1, second index is crval2
+
```
nabobalis:
```suggestion
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+        crpix : array or list of size of 2, default center of data array ((data.shape[0] + 1)/2., (data.shape[1] + 1)/2.)
+            reference pixel along each axis
+            first index is crpix1, second index is crpix2
+
```
nabobalis:
```suggestion
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+        cdelt : array or list of size of 2, default (0, 0)
+            pixel scaling along each axis (i.e. the spatial scale of the pixels)
+            first index is cdelt1, second index is cdelt2
+
```
nabobalis:
```suggestion
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+    def make_header(self, data, *coordinate, crval=None, crpix=None, cdelt=None, **kwargs):
+        """
+        Function to create a MetaDict from a coordinate object (SkyCoord or coordinate frame)
+        that is required to create a `sunpy.map.Map`
+
+        Parameters
+        ----------
+        data : `~numpy.ndarray`
+            array data of Map for which a header is requried
```
nabobalis:
```suggestion
            Array data of Map for which a header is required.
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+
+        coordinates : `~astropy.coordinates.SkyCoord` or `~astropy.coordinates.BaseFrame`, optional if not set default parameters are set
```
nabobalis:
```suggestion
        coordinates : `~astropy.coordinates.SkyCoord` or `~astropy.coordinates.BaseFrame`, optional
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+            coordinate object to get meta information for map header
```
nabobalis:
```suggestion
            Coordinate object to get meta information for map header.
             If not set, default parameters are used.
```

What are the default parameters?

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+
+        crval : array or list of size of 2, default (0, 0)
```
nabobalis:
```suggestion
        crval : `numpy.array` or `list` of size of 2, optional
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+            coordinate of reference pixel along each axis
```
nabobalis:
```suggestion
            Coordinate of reference pixel along each axis
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+            first index is crval1, second index is crval2
```
nabobalis:
```suggestion
            first index is ``crval1``, second index is ``crval2``
            Defaults to ``(0, 0)``.
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+
+        crpix : array or list of size of 2, default center of data array ((data.shape[0] + 1)/2., (data.shape[1] + 1)/2.)
+            reference pixel along each axis
```
nabobalis:
```suggestion
            Reference pixel along each axis
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+            array data of Map for which a header is requried
+
+        coordinates : `~astropy.coordinates.SkyCoord` or `~astropy.coordinates.BaseFrame`, optional if not set default parameters are set
+            coordinate object to get meta information for map header
+
+        crval : array or list of size of 2, default (0, 0)
+            coordinate of reference pixel along each axis
+            first index is crval1, second index is crval2
+
+        crpix : array or list of size of 2, default center of data array ((data.shape[0] + 1)/2., (data.shape[1] + 1)/2.)
```
nabobalis:
```suggestion
        crpix : `numpy.array` or `list` of size of 2, optional
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+            reference pixel along each axis
+            first index is crpix1, second index is crpix2
```
nabobalis:
```suggestion
            first index is ``crpix1``, second index is ``crpix2``.
            Defaults to the center of data array, ``((data.shape[0] + 1)/2., (data.shape[1] + 1)/2.)``.
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+
+        cdelt : array or list of size of 2, default (0, 0)
+            pixel scaling along each axis (i.e. the spatial scale of the pixels)
```
nabobalis:
```suggestion
            Pixel scaling along each axis (i.e., the spatial scale of the pixels)
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+
+        crval : array or list of size of 2, default (0, 0)
+            coordinate of reference pixel along each axis
+            first index is crval1, second index is crval2
+
+        crpix : array or list of size of 2, default center of data array ((data.shape[0] + 1)/2., (data.shape[1] + 1)/2.)
+            reference pixel along each axis
+            first index is crpix1, second index is crpix2
+
+        cdelt : array or list of size of 2, default (0, 0)
```
nabobalis:
```suggestion
        cdelt : `numpy.array` or `list` of size of 2, optional
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+            pixel scaling along each axis (i.e. the spatial scale of the pixels)
+            first index is cdelt1, second index is cdelt2
```
nabobalis:
```suggestion
            First index is ``cdelt1``, second index is ``cdelt2``.
            Defaults to ``(0, 0)``.
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+
+        **kwargs
```
nabobalis:
```suggestion
        **kwargs: 
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+            additional arguments that will be put into metadict header. These must be in the
+            list of acceptable meta keywords
```
nabobalis:
```suggestion
            list of acceptable meta keywords.
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+        crpix : array or list of size of 2, default center of data array ((data.shape[0] + 1)/2., (data.shape[1] + 1)/2.)
+            reference pixel along each axis
+            first index is crpix1, second index is crpix2
+
+        cdelt : array or list of size of 2, default (0, 0)
+            pixel scaling along each axis (i.e. the spatial scale of the pixels)
+            first index is cdelt1, second index is cdelt2
+
+        **kwargs
+            additional arguments that will be put into metadict header. These must be in the
```
nabobalis:
```suggestion
            Additional arguments that will be put into metadict header. These must be in the
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+            list of acceptable meta keywords
+
```
nabobalis: I don't see where the `**kwargs`` are passed in. Maybe I'm blind.

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+        Returns
+        -------
+        meta_dict : `~sunpy.util.MetaDict`
```
nabobalis:
```suggestion
        `~sunpy.util.MetaDict`
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+            a MetaDict that contains the header information required for making a Map
```
nabobalis:
```suggestion
            A `~sunpy.util.MetaDict` that contains the header information required for making a `sunpy.map.Map`.
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+        >>> from sunpy.coordinates import frames
+        >>> from astropy.coordinates import SkyCoord
+        >>> from astropy import units as u
+        >>> import numpy as np
+
+        >>> data = np.random.rand(1024, 1024)
+        >>> my_coord = SkyCoord(0*u.arcsec, 0*u.arcsec, obstime="2017-08-01", observer = 'earth', frame=frames.Helioprojective)
+        >>> my_header = map.MapHelper.make_header(data, my_coord, crval = [0,0], cdelt = [2,2])
+        >>> my_map = map.Map(data, my_header)
+
```
nabobalis:
```suggestion
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+        """
+
+        if coordinate:
+            if not isinstance(coordinate[0], (SkyCoord, frames.BaseCoordinateFrame)):
+                raise ValueError('coordinate needs to be an Astropy coordinate frame or SkyCoord instance.')
```
nabobalis:
```suggestion
                raise ValueError('Input coordinate needs to be an Astropy coordinate frame or SkyCoord instance.')
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+
+            meta = _get_meta_from_coordinate(coordinate[0])
+
+        else:
+            print('Using default meta assuming Earth based observer at current time with `arcsec` cunit type')
```
nabobalis: Should be the logger here?

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+        >>> my_map = map.Map(data, my_header)
+
+        """
+
+        if coordinate:
+            if not isinstance(coordinate[0], (SkyCoord, frames.BaseCoordinateFrame)):
+                raise ValueError('coordinate needs to be an Astropy coordinate frame or SkyCoord instance.')
+
+            meta = _get_meta_from_coordinate(coordinate[0])
+
```
nabobalis:
```suggestion
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+        >>> my_coord = SkyCoord(0*u.arcsec, 0*u.arcsec, obstime="2017-08-01", observer = 'earth', frame=frames.Helioprojective)
+        >>> my_header = map.MapHelper.make_header(data, my_coord, crval = [0,0], cdelt = [2,2])
+        >>> my_map = map.Map(data, my_header)
+
+        """
+
+        if coordinate:
+            if not isinstance(coordinate[0], (SkyCoord, frames.BaseCoordinateFrame)):
+                raise ValueError('coordinate needs to be an Astropy coordinate frame or SkyCoord instance.')
+
```
nabobalis:
```suggestion
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+        for key in kwargs:
+            if key in acceptable_meta:
+                meta_dict[key] = kwargs[key]
+
+        return meta_dict
+
+
+def _default_meta():
+    """
+    Creates default meta data if none is specified
```
nabobalis:
```suggestion
    Creates a `dict` with the basic metadata.
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+    coord_meta['t_obs'] = parse_time('now').iso
+    coord_meta['date_obs'] = parse_time('now').iso
+
+    return coord_meta
+
+
+def _get_meta_from_coordinate(coordinate):
+    """
+    Function to get WCS meta data from a SkyCoord or frame using
+    `astropy.wcs.utils.celestial_frame_to_wcs`
```
nabobalis:
```suggestion
    `astropy.wcs.utils.celestial_frame_to_wcs`.
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+    coord_meta['rsun_obs'] = ((con.radius / con.au).decompose() * u.radian).to(u.arcsec).value
+    coord_meta['t_obs'] = parse_time('now').iso
+    coord_meta['date_obs'] = parse_time('now').iso
+
+    return coord_meta
+
+
+def _get_meta_from_coordinate(coordinate):
+    """
+    Function to get WCS meta data from a SkyCoord or frame using
```
nabobalis:
```suggestion
    Function to get WCS meta data from a `~astropy.coordinates.SkyCoord` or frame using
```

Not sure if that is right.

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+    `astropy.wcs.utils.celestial_frame_to_wcs`
+
+    Parameters
+    ----------
+    coordinate : `~astropy.coordinates.SkyCoord` or `~astropy.coordinates.BaseFrame`
+        coordinate object to pull meta data from
```
nabobalis:
```suggestion
        Coordinate object to pull metadata from.
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+
+    Returns
+    -------
+    coord_meta : dict
```
nabobalis:
```suggestion
    `dict`
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+        containing the meta information required for `sunpy.map.Map` header
```
nabobalis:
```suggestion
        Containing the meta information required for `sunpy.map.Map` header:
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+            * ctype1, ctype2
+            * cunit1, cunit2
+            * hgln_obs, hglt_obs
+            * dsun_obs
+            * rsun_obs
+            * t_obs, date_obs
+
```
nabobalis:
```suggestion
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+    """
+    coord_meta = {}
+
+    if not isinstance(coordinate, (SkyCoord, frames.BaseCoordinateFrame)):
+        raise ValueError('takes an Astropy coordinate frame or SkyCoord instance.')
```
nabobalis:
```suggestion
        raise ValueError('Input must be an Astropy coordinate frame or SkyCoord instance.')
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+
+    if isinstance(coordinate, SkyCoord):
+        if coordinate.obstime is None or coordinate.frame is None:
+            raise ValueError('SkyCoord needs an observation time and a frame')
+        skycoord_wcs = astropy.wcs.utils.celestial_frame_to_wcs(coordinate.frame)
+    elif isinstance(coordinate, frames.SunPyBaseCoordinateFrame):
+        if coordinate.obstime is None:
+            raise ValueError('frame needs an observation time')
```
nabobalis:
```suggestion
            raise ValueError('Frame needs an observation time.')
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+
+    """
+    coord_meta = {}
+
+    if not isinstance(coordinate, (SkyCoord, frames.BaseCoordinateFrame)):
+        raise ValueError('takes an Astropy coordinate frame or SkyCoord instance.')
+
+    if isinstance(coordinate, SkyCoord):
+        if coordinate.obstime is None or coordinate.frame is None:
+            raise ValueError('SkyCoord needs an observation time and a frame')
```
nabobalis:
```suggestion
            raise ValueError('SkyCoord needs an observation time and a frame.')
```

sunpy/map/maputils.py
```diff
@@ -1,7 +1,309 @@
...
+    coord_meta['cunit1'], coord_meta['cunit2'] = skycoord_wcs.wcs.cunit
+    coord_meta['hgln_obs'], coord_meta['hglt_obs'] = skycoord_wcs.heliographic_observer.lon, skycoord_wcs.heliographic_observer.lat
+    coord_meta['dsun_obs'] = skycoord_wcs.heliographic_observer.radius.to(u.m)
+    coord_meta['rsun_ref'] = skycoord_wcs.rsun.to(u.m)
+    coord_meta['rsun_obs'] = ((skycoord_wcs.rsun.to(u.AU)/skycoord_wcs.heliographic_observer.radius.to(u.AU)).decompose() * u.radian).to(u.arcsec).value
+    coord_meta['t_obs'], coord_meta['date_obs'] = skycoord_wcs.heliographic_observer.obstime, skycoord_wcs.heliographic_observer.obstime
+
+    return coord_meta
+
+MapHelper = MapHelper()
```
nabobalis: What if we made the function call lower case?

Cadair:
I think that it would be best to move this into it's own file (should `maputils` be a directory?) and drop the class (using the file to group all the parts).

I will go through the exact header logic tomorrow morning probably.

sunpy/map/maputils.py
```diff
@@ -1,7 +1,308 @@
...
+                if k in acceptable_meta.keys():
+                    print(k, " : ", acceptable_meta[k])
+                else:
+                    print(k, " is not in the accepted meta data required for sunpy.map.Map \n The accepted keywords are : \n", list(acceptable_meta.keys()))
+            return
+
+        for k in acceptable_meta:
+            print(k, " : ", acceptable_meta[k])
+
+    def make_header(self, data, *coordinate, crval=None, crpix=None, cdelt=None, **kwargs):
```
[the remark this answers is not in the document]
hayesla: yeah I agree, I think we should probably get rid of the default map, and just have detailed docs with examples

sunpy/map/maputils.py
hayesla: in the last line in the function 
```for key in kwargs: ```

File: sunpy/map/maputils.py [hunk 63]
```diff
@@ -1,7 +1,308 @@
...
+        else:
+            meta['crpix1'], meta['crpix2'] = crpix[0], crpix[1]
+
+        # if no cdelt options set, default to 1, 1
+        if cdelt is None:
+            meta['cdelt1'], meta['cdelt2'] = 1, 1
+        else:
+            meta['cdelt1'], meta['cdelt2'] = cdelt[0], cdelt[1]
+
+        meta_dict = MetaDict({
```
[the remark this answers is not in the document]
hayesla: yes.. yes it does

sunpy/map/maputils.py
```diff
@@ -1,7 +1,308 @@
...
+    coord_meta = {}
+
+    if not isinstance(coordinate, (SkyCoord, frames.BaseCoordinateFrame)):
+        raise ValueError("Input must be an Astropy coordinate frame or SkyCoord instance.")
+
+    if isinstance(coordinate, SkyCoord):
+        if coordinate.obstime is None or coordinate.frame is None:
+            raise ValueError("SkyCoord needs an observation time and a frame.")
+        skycoord_wcs = astropy.wcs.utils.celestial_frame_to_wcs(coordinate.frame)
+    elif isinstance(coordinate, frames.SunPyBaseCoordinateFrame):
```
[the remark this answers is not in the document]
hayesla: but should it not need a frame also?

hayesla:
I've updated this PR but I think we need to think about what else we want the helper function to do. At the moment I've updated it so that its a function that now lives in `sunpy.map.maphelper.py`

The two functions that live here are 

`meta_keywords` which prints the list of accepted fits keywords that are used to make a `sunpy.map.GenericMap`. 

The other one is `make_header` which takes an `astropy.coordinates.Skycoord`, from which the `cravl` meta is taken from the units given in the SkyCoord object, then the `date_obs` is taken from the `obstime` given in the SkyCoord, and the `cunit`, `ctype`,`hgln_obs`, `hglt_obs`, `dsun_obs` can all been taken from the SkyCoord.frame and defined observed in the SkyCoord.
The function also takes two optional inputs of astropy quantities of `reference_pixel` and `scale` from which `crpix` and `cdelt` are taken from. 

I guess the question is if we also want a function that doesn't include the passing of a SkyCoord - something similar to the SSW `make_map` https://hesperia.gsfc.nasa.gov/ssw/gen/idl/maps/make_map.pro

such that it would look something like:

```
def make_header(data, date_obs = parse_time('now'), xcen_coord = 0.0, ycen_coord = 0.0, dx = 1.0, dy = 1.0, xunit = 'arcsec', yunit = 'arcsec', coord_system = 'Helioprojective'):

	meta = {}
	meta['crval1'] = xcen_coord
	meta['crval2'] = ycen_coord
	meta['cdelt1'] = dx
	meta['cdelt2'] = dy
	meta['crpix1'] = (data.shape[0] + 1)/2.
	meta['crpix2'] = (data.shape[1] + 1)/2.
	meta['cunit1'] = xunit
	meta['cunit2'] = yunit
	meta['date_obs'] = date_obs 

	if coord_system.lower() == ('helioprojective' or 'heliographicstoneyhurst'):
		ctype1, ctype2 = 'HPLN-TAN', 'HPLT-TAN'
	elif coord_system.lower() == ('heliographiccarrington'):
		ctype1, ctype2 = 'CRLN-TAN', 'CRLT-TAN'
	elif coord_system.lower() == ('heliocentric'):
		ctype1, ctype2 = 'SOLX', 'SOLY'

	meta['ctype1'] = ctype1
	meta['ctype2'] = ctype2

	return MetaDict(meta)
```

hayesla:
will be working on this all day tomorrow with the sprints - so any suggestions on what this should look like let me know!

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,200 @@
...
+            * date_obs
+    """
+
+    coord_meta = {}
+
+    if not isinstance(coordinate, (SkyCoord, frames.BaseCoordinateFrame)):
+        raise ValueError("Input must be an Astropy coordinate frame or SkyCoord instance.")
+
+    if isinstance(coordinate, SkyCoord):
+        if coordinate.obstime is None or coordinate.frame is None:
```
Cadair: I don't think it's possible to have a `SkyCoord` without a frame. I suggest you convert a `SkyCoord` to a frame here:
```python
if isinstance(coordinate, SkyCoord):
    coordinate = coordinate.frame
```

and then just implement all the logic after that for frames and not SkyCoord.

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,200 @@
...
+        skycoord_wcs = astropy.wcs.utils.celestial_frame_to_wcs(coordinate.frame)
+    elif isinstance(coordinate, frames.SunPyBaseCoordinateFrame):
+        if coordinate.obstime is None:
+            raise ValueError("Frame needs an observation time.")
+        skycoord_wcs = astropy.wcs.utils.celestial_frame_to_wcs(coordinate)
+
+    cunit1, cunit2 = skycoord_wcs.wcs.cunit
+    coord_meta = dict(skycoord_wcs.to_header())
+    coord_meta['cunit1'], coord_meta['cunit2'] = cunit1, cunit2  # this is done as to_header transforms cunit to deg
+    coord_meta['hgln_obs'], coord_meta['hglt_obs'] = skycoord_wcs.heliographic_observer.lon, skycoord_wcs.heliographic_observer.lat
```
Cadair: All four of these assignments here should not be quantities, so they should all be using `to_value`.

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,200 @@
...
+    coord_meta['dsun_obs'] = skycoord_wcs.heliographic_observer.radius.to(u.m)
+    coord_meta['rsun_ref'] = skycoord_wcs.rsun.to(u.m)
+    coord_meta['rsun_obs'] = ((skycoord_wcs.rsun.to(u.AU)/skycoord_wcs.heliographic_observer.radius.to(u.AU)).decompose() * u.radian).to(u.arcsec).value
```
Cadair: Is this equation correct? Shouldn't this be an angular quantity?

Also you don't need to to `.to(u.AU)` on both sides of this, the `Quantity` magic will handle all the required unit conversion for you.

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,200 @@
...
+from sunpy.coordinates import frames
+import warnings
+from sunpy.util.exceptions import SunpyUserWarning
+
+__all__ = ['meta_keywords', 'make_header']
+
+
+def meta_keywords(*key):
+    """
+    Prints the accepted metadata keywords that are used when
```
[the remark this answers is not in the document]
hayesla: yes so accepted meta keywords here means essentially what keywords are actually taken from the meta and used in the generation of the map, of course more can be given they just wont necessarily be called in `GenericMap` (for example #1832). I think this function is helpful so that the user can see what the actual keywords are being used - but it may be better suited on the docs? Or if there was a way we could scrape from `mapbase.py` so that it gets updated

ehsteve:
Just to sow discord. I'm not sure I am on board with the functional approach...

Cadair:
:grin: That @ehsteve  is because you love yourself a class.

To my mind there is no state being stored anywhere here and therefore no need for a class at all. The main API here is one function to generate a map header.

sunpy/map/maphelper.py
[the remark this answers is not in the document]
Cadair: > I think this function is helpful so that the user can see what the actual keywords are being used

I can see the utility here, and I think we should keep it, I do think we need a more appropriate word than "accepted" though.

> Or if there was a way we could scrape from mapbase.py so that it gets updated.

Not easily enough that it's worth doing in this PR, I have an idea or two about how we could go about doing this if we wanted to. I don't think this set of data is likely to change very often so I am happy to keep it as it is.

If this is just for reference then maybe we should just be auto-generating some documentation rather than providing it as a function?

nabobalis:
That CI fail looks like something I need to fix and is unrelated to the PR

sunpy/map/maphelper.py
hayesla: is there a reason they shouldn't be quantities?

Cadair: My concern is consistency with values that come out of headers of FITS files, which will be floats and not quantities. I worry that it's implicitly assumed that they are not in GenericMap (especially `.save`)

ehsteve:
@Cadair,  @hayesla and I discussed the design and scope for this PR and we should all discuss to all get on the same page. We could do that on the phone call or earlier if you think there will already be lots to do on the call.

sunpy/map/maphelper.py
hayesla: Good question - I'm actually not too sure?

hayesla:
So with these commits several things have changes:

* tried to fix all the comments  - still unsure is the dsun_obs calculation is correct, but it seems to be right when playing around with calculating it 

* The function `accepted_meta` is now called `map_meta_keywords`

* The function `make_meta` is now called `make_fitswcs_header` following discussion yesterday

* if a coordinate system other than helioprojective is given, the coordinate is transformed to helioprojective 

*the `_get_meta_from_coordinate` function is now broken into two separate functions `_get_wcs_meta` and `_get_observer_meta`

* added a function `get_instrument_meta` that helps the user pass more logical keywords i.e. ``instrument`` vs ``instrume``. I don't know if we want to keep this, but following some discussion yesterday I threw it in

An example of using this helper is in this [gist](https://gist.github.com/hayesla/4cfc7846ff3d58b6660df95be0b88c76)

hayesla:
@Cadair  @ehsteve  let me know if you have any comments

wafels:
The calculation or rsun_obs is a small angle approximation. It should be replaced with the full (arctan I think) calculation to account for all situations.

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,246 @@
...
+    Parameters
+    ----------
+    data : `~numpy.ndarray`
+        Array data of Map for which a header is required.
+    coordinates : `~astropy.coordinates.SkyCoord`
+        Coordinate object to get meta information for map header.
+    reference_pixel :`~astropy.units.Quantity` of size 2, optional
+        Reference pixel along each axis. These are expected to be Cartestian ordered, i.e
+        the first index is the x axis, second index is the y axis.
+        Defaults to the center of data array, ``(data.shape[0] + 1)/2., (data.shape[1] + 1)/2.)``.
```
Cadair: This is back-to-front

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,246 @@
...
+    scale : `~astropy.units.Quantity` of size 2, optional
+        Pixel scaling along x and y axis (i.e. the spatial scale of the pixels (dx, dy)). These are
+        expected to be Cartestian ordered, i.e [dx, dy].
+        Defaults to ``([1., 1.] arcsec)``.
```
Cadair:
```suggestion
        Defaults to ``([1., 1.] arcsec/pixel)``.
```

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,246 @@
...
+        ``obsrvtry``, ``wavelnth``, ``exptime`` respectively.
+
+    Returns
+    -------
+    `~sunpy.util.MetaDict`
+        A `sunpy.util.MetaDict` that contains the header information required for making a `sunpy.map.Map`
+
+    Examples
+    --------
+    >>> from sunpy import map
```
Cadair: `map` is a Python builtin:

```suggestion
    >>> import sunpy.map
```

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,246 @@
...
+    >>> from sunpy.coordinates import frames
+    >>> from astropy.coordinates import SkyCoord
+    >>> from astropy import units as u
+    >>> import numpy as np
+
+    >>> data = np.random.rand(1024, 1024)
+    >>> my_coord = SkyCoord(0*u.arcsec, 0*u.arcsec, obstime="2017-08-01", observer = 'earth', frame=frames.Helioprojective)
+    >>> my_header = map.make_fits_wcs(data, my_coord)
+    >>> my_map = map.Map(data, my_header)
```
Cadair:
```suggestion
    >>> my_map = sunpy.map.Map(data, my_header)
```

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,246 @@
...
+    --------
+    >>> from sunpy import map
+    >>> from sunpy.coordinates import frames
+    >>> from astropy.coordinates import SkyCoord
+    >>> from astropy import units as u
+    >>> import numpy as np
+
+    >>> data = np.random.rand(1024, 1024)
+    >>> my_coord = SkyCoord(0*u.arcsec, 0*u.arcsec, obstime="2017-08-01", observer = 'earth', frame=frames.Helioprojective)
+    >>> my_header = map.make_fits_wcs(data, my_coord)
```
Cadair:
```suggestion
    >>> my_header = sunpy.map.make_fits_wcs(data, my_coord)
```

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,246 @@
...
+    >>> my_map = map.Map(data, my_header)
+    """
+
+    if not isinstance(coordinate, SkyCoord):
+        raise ValueError("coordinate needs to be a SkyCoord instance.")
+
+    if coordinate.obstime is None:
+        raise ValueError("The coordinate needs an observation time, `obstime`.")
+
+    coordinate = coordinate.transform_to(frames.Helioprojective)
```
Cadair:
```suggestion
```

We shouldn't do this, it's perfectly valid to construct a FITS-WCS header in other frames, we just need to only add the observer info where an observer is specified.

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,246 @@
...
+    -------
+    `dict`
+        Containing the WCS meta information
+            * hgln_obs, hglt_obs
+            * dsun_obs
+            * rsun_obs
+    """
+    coord_meta = {}
+
+    skycoord_wcs = astropy.wcs.utils.celestial_frame_to_wcs(coordinate.frame)
```
Cadair: Do we want this function to also do this? My understanding after our chat on Wed was this one would *only* do the observer stuff?

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,246 @@
...
+    >>> from astropy import units as u
+    >>> import numpy as np
+
+    >>> data = np.random.rand(1024, 1024)
+    >>> my_coord = SkyCoord(0*u.arcsec, 0*u.arcsec, obstime="2017-08-01", observer = 'earth', frame=frames.Helioprojective)
+    >>> my_header = map.make_fits_wcs(data, my_coord)
+    >>> my_map = map.Map(data, my_header)
+    """
+
+    if not isinstance(coordinate, SkyCoord):
```
Cadair: Should we also support frame objects in this function?

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,246 @@
...
+        raise ValueError("coordinate needs to be a SkyCoord instance.")
+
+    if coordinate.obstime is None:
+        raise ValueError("The coordinate needs an observation time, `obstime`.")
+
+    coordinate = coordinate.transform_to(frames.Helioprojective)
+
+    meta_wcs = _get_wcs_meta(coordinate)
+
+    meta_observer = _get_observer_meta(coordinate)
```
Cadair:
```suggestion
    if hasattr(coordinate, "observer"):
        meta_observer = _get_observer_meta(coordinate)
```

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,246 @@
...
+
+    meta_wcs.update(meta_observer)
+    meta_wcs.update(meta_instrument)
+
+    if reference_pixel is None:
+        reference_pixel = u.Quantity([(data.shape[1] + 1)/2.*u.pixel, (data.shape[0] + 1)/2.*u.pixel])
+    if scale is None:
+        scale = u.Quantity([1.0*u.arcsec, 1.0*u.arcsec])
+
+    meta_wcs['crval1'], meta_wcs['crval2'] = coordinate.Tx.to_value(meta_wcs['cunit1']), coordinate.Ty.to_value(meta_wcs['cunit2'])
```
Cadair: We need to do this in a frame agnostic way, which is hard. I suggest for the moment, like we do in map we don't allow HCC coordinates (and only spherical ones) and do what we do in `mapbase` to get lon / lat:
https://github.com/sunpy/sunpy/blob/master/sunpy/map/mapbase.py#L254

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,246 @@
...
+
+    coord_meta['hgln_obs'], coord_meta['hglt_obs'] = skycoord_wcs.heliographic_observer.lon.to_value(), skycoord_wcs.heliographic_observer.lat.to_value()
+    coord_meta['dsun_obs'] = skycoord_wcs.heliographic_observer.radius.to(u.m).to_value()
+    coord_meta['rsun_ref'] = skycoord_wcs.rsun.to(u.m).to_value()
+    coord_meta['rsun_obs'] = ((skycoord_wcs.rsun/skycoord_wcs.heliographic_observer.radius).decompose() * u.radian).to(u.arcsec).to_value()
+
+    return coord_meta
+
+
+def _get_instrument_meta(**kwargs):
```
Cadair: As in the call I am not sure I am 100% on board with this, but I am happy to have it. It does however need a couple of tweaks:

1) It needs to pop the values from `kwargs` or else we will end up raising an error later on in the header gen function. (see in line suggestion)
2) We need to document all of these kwargs in the docstring for `make_fitswcs_header`.

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,246 @@
...
+    """
+    Function to correctly name keywords from kwargs
+    """
+    coord = {}
+
+    conversion = {'instrument': 'instrume', 'telescope': 'telescop', 'observatory': 'obsrvtry', 'wavelength': 'wavelnth', 'exposure': 'exptime'}
+
+    for key in kwargs:
+        if key in conversion:
+            coord[conversion[key]] = kwargs[key]
```
Cadair:
```suggestion
            coord[conversion[key]] = kwargs.pop(key)
```

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,246 @@
...
+            * dsun_obs
+            * rsun_obs
+    """
+    coord_meta = {}
+
+    skycoord_wcs = astropy.wcs.utils.celestial_frame_to_wcs(coordinate.frame)
+
+    coord_meta['hgln_obs'], coord_meta['hglt_obs'] = skycoord_wcs.heliographic_observer.lon.to_value(), skycoord_wcs.heliographic_observer.lat.to_value()
+    coord_meta['dsun_obs'] = skycoord_wcs.heliographic_observer.radius.to(u.m).to_value()
+    coord_meta['rsun_ref'] = skycoord_wcs.rsun.to(u.m).to_value()
```
Cadair:
```suggestion
    coord_meta['rsun_ref'] = skycoord_wcs.to_value(u.m)
```

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,246 @@
...
+            * hgln_obs, hglt_obs
+            * dsun_obs
+            * rsun_obs
+    """
+    coord_meta = {}
+
+    skycoord_wcs = astropy.wcs.utils.celestial_frame_to_wcs(coordinate.frame)
+
+    coord_meta['hgln_obs'], coord_meta['hglt_obs'] = skycoord_wcs.heliographic_observer.lon.to_value(), skycoord_wcs.heliographic_observer.lat.to_value()
+    coord_meta['dsun_obs'] = skycoord_wcs.heliographic_observer.radius.to(u.m).to_value()
```
Cadair:
```suggestion
    coord_meta['dsun_obs'] = skycoord_wcs.heliographic_observer.radius.to_value(u.m)
```

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,246 @@
...
+        Containing the WCS meta information
+            * hgln_obs, hglt_obs
+            * dsun_obs
+            * rsun_obs
+    """
+    coord_meta = {}
+
+    skycoord_wcs = astropy.wcs.utils.celestial_frame_to_wcs(coordinate.frame)
+
+    coord_meta['hgln_obs'], coord_meta['hglt_obs'] = skycoord_wcs.heliographic_observer.lon.to_value(), skycoord_wcs.heliographic_observer.lat.to_value()
```
Cadair:
```suggestion
    coord_meta['hgln_obs'], coord_meta['hglt_obs'] = skycoord_wcs.heliographic_observer.lon.to_value(u.deg), skycoord_wcs.heliographic_observer.lat.to_value(u.deg)
```

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,246 @@
...
+    coord_meta['dsun_obs'] = skycoord_wcs.heliographic_observer.radius.to(u.m).to_value()
+    coord_meta['rsun_ref'] = skycoord_wcs.rsun.to(u.m).to_value()
+    coord_meta['rsun_obs'] = ((skycoord_wcs.rsun/skycoord_wcs.heliographic_observer.radius).decompose() * u.radian).to(u.arcsec).to_value()
```
Cadair: As @wafels pointed out this needs to use the full equation not the small angle approximation.

Cadair:
```suggestion
    coord_meta['rsun_obs'] = ((skycoord_wcs.rsun/skycoord_wcs.heliographic_observer.radius).decompose() * u.radian).to_value(u.arcsec)
```

sunpy/map/maputils.py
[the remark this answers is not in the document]
Cadair: I am pretty sure `SkyCoord` always had a frame.

Cadair:
@hayesla I have left a bunch more in line comments. I have also gone through and resolved all the comments I think are sorted or out of date, so the open comments on the PR are all things that need looking at (I think)

sunpy/map/maphelper.py
hayesla: yeah so at the moment it just does the observer stuff right? It just pulls it from the heliographic_observer info in the frame - should we pull it rather from the skycoord observer keyword rather than the frame?

sunpy/map/maphelper.py
hayesla: yeah so I originally had that - but if we want observer info some frames do not have that? what do you think?

sunpy/map/maphelper.py
hayesla: oh following on the previous comment - should we just have that if the observer information isn't given we just let map deal with that rather than requiring it here?

sunpy/map/maphelper.py
Cadair: Yeah I would suggest you use the coordinate directly.

sunpy/map/maphelper.py
Cadair: yes :)

sunpy/map/maphelper.py
hayesla: ping @Cadair

hayesla:
Ok so I think I've fixed the open comments - let me know what you think @Cadair 

and any other comments/docstrings etc, still need to do some tests!

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,253 @@
...
+    if not isinstance(coordinate, (SkyCoord, frames.BaseCoordinateFrame)):
+        raise ValueError("coordinate needs to be a coordinate frame or an SkyCoord instance.")
+
+    if coordinate.obstime is None:
+        raise ValueError("The coordinate needs an observation time, `obstime`.")
+
+    if isinstance(coordinate, SkyCoord):
+        coordinate = coordinate.frame
+
+    if coordinate.name == 'heliocentric':
```
Cadair: I would you replace this with an `isinstance` check: (I think that's the correct frame name)
```suggestion
    if isinstance(coordinate, frames.Heliocentric):
```

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,253 @@
...
+        raise ValueError("SunPy Map does not support three dimensional data "
```
Cadair: I would reword this error to be less map specific. Just something along the lines of "this function currently does not support heliocentric coordinates."

sunpy/map/maphelper.py
```diff
@@ -0,0 +1,253 @@
...
+                         "and therefore cannot represent heliocentric coordinates.")
+
+    meta_wcs = _get_wcs_meta(coordinate)
+
+    if hasattr(coordinate, "observer"):
+        meta_observer = _get_observer_meta(coordinate)
+        meta_wcs.update(meta_observer)
+
+    meta_instrument = _get_instrument_meta(**kwargs)
+
```
Cadair:
```suggestion
```

PR #3089, opened 2019-05-02
backport rhessi test fix

nabobalis:
Did not backport this but somehow did not fail on any of the CI. 

ALso I forgot I have black enabled on my work laptop.


nabobalis:
Skipped it since it fails on a range of different versions.

Issue #2590, opened 2018-04-16
Sample AIA images in the gallery do not show SunPy plotting very well


wafels:
The colors selected AIA images in the gallery are badly scaled.  This is probably because the data contain one or two very bright pixels which is throwing the existing scaling off.

Users going to the gallery are seeing badly scaled images, and that's not a good advertisement for SunPy.  Either the images should be changed, the default AIA scaling should be changed, or some clipping of the intensity scale should be included in the example gallery code as appropriate.

sapnamish:
I want to work on this issue

nabobalis:
@sapnamish You don't need to ask for permission to work on this issue! You can go right ahead!

ravindra1307:
is this issue still open ? i want to work on this issue. can you please guide me to the page where i can see badly scaled image ?

nabobalis:
Hi @ravindra1307, yes this issue is still open. I assume that @wafels is referring to the `Sample Data Set Overview` gallery example. But also to a more general point that we use the sample data in many of our examples and the scaling of the AIA images might need to be looked at either within the examples or inside the AIA Map code.

Cadair:
now that #3100 is merged we should modify the examples and the docs where appropriate to use that functionality to show the images with better scaling.

PR #3100, opened 2019-05-09
Clipping Map plots to a percentile interval

ayshih:
This PR adds the capability to easily clip Map plots to a percentile interval.  Facilitates fixing of #2590.  Discuss.

## No clipping (the default)
```python
>>> import sunpy.data.sample
>>> from sunpy.map import Map
>>> m = Map(sunpy.data.sample.AIA_171_IMAGE)
>>> m.peek()
```
a

## With clipping
Here, the clipping percentile interval is (1%, 99.95%).  That is, the lowest 1% and the highest 0.05% of pixels are being clipped out.
```python
>>> import astropy.units as u
>>> import sunpy.data.sample
>>> from sunpy.map import Map
>>> m = Map(sunpy.data.sample.AIA_171_IMAGE)
>>> m.peek(clip_interval=(1., 99.95)*u.percent)
```
b

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1655,29 +1655,30 @@ class GenericMap(NDData):
         figure.show()
 
     @toggle_pylab
-    def plot(self, annotate=True, axes=None, title=True, clip_percentile=None, **imshow_kwargs):
+    @u.quantity_input(clip_interval=u.percent)
+    def plot(self, annotate=True, axes=None, title=True, clip_interval=None, **imshow_kwargs):
         """
         Plots the map object using matplotlib, in a method equivalent
         to plt.imshow() using nearest neighbour interpolation.
 
         Parameters
         ----------
-        annotate : bool
-            If True, the data is plotted at it's natural scale; with
+        annotate : `bool`, optional
+            If `True`, the data is plotted at its natural scale; with
             title and axis labels.
 
         axes: `~matplotlib.axes` or None
             If provided the image will be plotted on the given axes. Else the
             current matplotlib axes will be used.
 
-        title : bool
-            If True, include the title
+        title : `bool`, optional
+            If `True`, include the title.
 
-        clip_percentile : tuple of 2 numbers
+        clip_interval : two-element `~astropy.units.Quantity`, optional
             If provided, the data will be clipped to the percentile interval bounded by the two
-            numbers
+            numbers.
 
-        **imshow_kwargs  : dict
+        **imshow_kwargs  : `dict`
             Any additional imshow arguments that should be used
             when plotting.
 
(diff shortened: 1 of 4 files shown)


nabobalis:
A changelog would be nice. Test wise, we could create a figure test?

sunpy/map/mapbase.py
```diff
@@ -1669,6 +1670,13 @@ def plot(self, annotate=True, axes=None, title=True, **imshow_kwargs):
             If provided the image will be plotted on the given axes. Else the
             current matplotlib axes will be used.
 
+        title : bool
```
```suggestion
        title : `bool`
```

sunpy/map/mapbase.py
```diff
@@ -1669,6 +1670,13 @@ def plot(self, annotate=True, axes=None, title=True, **imshow_kwargs):
...
+            If True, include the title
```
```suggestion
            If `True`, include the title.
```

sunpy/map/mapbase.py
```diff
@@ -1669,6 +1670,13 @@ def plot(self, annotate=True, axes=None, title=True, **imshow_kwargs):
...
+
+        clip_percentile : tuple of 2 numbers
```
```suggestion
        clip_percentile : `tuple` of 2 numbers, optional
```

sunpy/map/mapbase.py
```diff
@@ -1669,6 +1670,13 @@ def plot(self, annotate=True, axes=None, title=True, **imshow_kwargs):
...
+            If provided, the data will be clipped to the percentile interval bounded by the two
+            numbers
```
```suggestion
            numbers.
```

sunpy/map/mapbase.py
```diff
@@ -1732,6 +1740,15 @@ def plot(self, annotate=True, axes=None, title=True, **imshow_kwargs):
             imshow_args.update({'extent': x_range + y_range})
         imshow_args.update(imshow_kwargs)
 
+        if clip_percentile is not None:
+            if len(clip_percentile) == 2:
+                vmin, vmax = AsymmetricPercentileInterval(*clip_percentile).get_limits(self.data)
+            else:
+                raise ValueError("Clip percentile interval must be a tuple of 2 numbers")
```
```suggestion
                raise ValueError("Clip percentile interval must be have a  length of  2.")
```

sunpy/map/mapbase.py
Cadair: Should this be a `u.Quantity` object?

Cadair:
I think this is a great little feature. I think we should probably edit the examples and docs to use it where appropriate with our sample data?

sunpy/map/mapbase.py

sunpy/map/mapbase.py
```diff
@@ -1654,22 +1655,30 @@ def peek(self, draw_limb=False, draw_grid=False,
...
 
         axes: `~matplotlib.axes` or None
             If provided the image will be plotted on the given axes. Else the
             current matplotlib axes will be used.
 
-        **imshow_kwargs  : dict
+        title : `bool`, optional
+            If `True`, include the title.
+
+        clip_interval : two-element `~astropy.units.Quantity`, optional
```
samaloney: Does this *really* need to be as astropy quantity? If so and the quantity is known should probably check for it raise and error if doesn't match?

ayshih: The `u.quantity_input` decorator takes care of (some of) the error checking.

It may be slightly cumbersome for this to be a `Quantity`, but it's our guidelines to use units as much as possible.  As a case in point, for this PR I played around with some of the underlying Matplotlib machinery, and I was initially confused because I didn't know whether I was supposed to be supplying percentages or fractions as inputs, so I had to empirically figure it out.

Cadair: We are generally pretty stringent with everything being quantities where that makes sense. the `@u.quantity_input` will validate that the unit is `u.percent`

ehsteve: Thumbs up on using units here though the `percent` quantity is not so useful beyond just identifying itself as such.

ayshih: Note that one could choose to specify `clip_interval=(0.01, 0.9995)*u.one` in the example in the first post and get the same clipping interval (1% to 99.95%)

samaloney: Ok, I just really, really dislike the extra typing in cases like this.

Cadair: heh:
```python
>>> (0.99*u.one).to(u.percent)                                                                                                                                                                                                 
<Quantity 99. %>
```
Didn't know that worked...

PR #3105, opened 2019-05-15
Cleanup sample data example

dstansby:
Once #3103 has gone in, this is a little cleanup of the sample data example that cuts out some of the un-needed plotting lines.


ehsteve:
would you be willing to expand the scope of this PR and remove all uses of `peek()` in the example gallery :)

dstansby:
I'm confused, I've changed everything here to use `.peek()`, not get rid of it!

ehsteve:
Sorry! We just had a discussion during the dev meeting and it was general consensus that we should not be using `peek()` in the gallery at all since it is not really how scientists should learn to interact with sunpy and plotting. Would you be willing to remove peek() from all examples instead of adding it?

dstansby:
Here's some, I got bored so if anyone else wants to do the rest feel free to force push (otherwise I'll finish it off in the future)

dstansby:
I think that's everything 🎉

nabobalis:
Thanks @dstansby

PR #3106, opened 2019-05-15
Tidy up warning when plotting on non-wcsaxes

### LINKED ISSUE #3096: It is not clear how not using WCSAxes could lead to a problem when plotting a map
### Description

It is possible to pass a standard axes object to a `map.plot(axes=ax)`. This will lead to unexpected behavior (likely positions will be off by 0.5 pixels).

### Expected behavior

It should likely not be allowed but at the very least, a clear warning should be issued that makes it clear

### Actual behavior

Currently a vague warning is issued about the map not being oriented properly which suggests that there is something wrong with the map, not something wrong with the plotting.

### Steps to Reproduce

```python
ax = plt.subplot()
map1.plot(axes=ax)
```

I strongly recommend that the error not only describe the problem but give the user the information to correct it.

dstansby:
Fixes #3096. Previously only the first warning would be raised, now raise the more important warning (no `WCSAxes`) first, and then do the additional check for the wrong rotation.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1701,7 +1701,7 @@ class GenericMap(NDData):
                               SunpyUserWarning)
                 # Check if the image is properly oriented
                 if not np.array_equal(self.rotation_matrix, np.identity(2)):
-                    warnings.warn("This map is not properly oriented. Plot axes may be incorrect.",
+                    warnings.warn("The axes of this map are not aligned to the pixel grid. Plot axes may be incorrect.",
                                   SunpyUserWarning)
 
         # Normal plot
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1694,16 +1694,15 @@ class GenericMap(NDData):
             axes = wcsaxes_compat.gca_wcs(self.wcs)
 
         if not _basic_plot:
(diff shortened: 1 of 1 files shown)


sunpy/map/mapbase.py
```diff
@@ -1694,16 +1694,15 @@ def plot(self, annotate=True, axes=None, title=True, **imshow_kwargs):
...
-            elif not wcsaxes_compat.is_wcsaxes(axes):
+            if not wcsaxes_compat.is_wcsaxes(axes):
                 warnings.warn("WCSAxes not being used as the axes object for this plot."
-                              " Plots may have unexpected behaviour.",
+                              " Plots may have unexpected behaviour. To fix this pass "
+                              "'projection=map' when creating the axes",
                               SunpyUserWarning)
+                # Check if the image is properly oriented
+                if not np.array_equal(self.rotation_matrix, np.identity(2)):
+                    warnings.warn("This map is not properly oriented. Plot axes may be incorrect.",
```
Cadair: I think this message could be better, and we should take this opportunity to fix it. How's this for a suggestion:
```suggestion
                    warnings.warn("The axes of this map are not aligned to the pixel grid. Plot axes may be incorrect.",
```

Cadair:
Thanks @dstansby

PR #3111, opened 2019-05-16
Remove header from licence

Cadair:
I think this was causing GitHub to not correctly identify the licence type. Also it's not that informative (and kinda wrong)


nabobalis:
IM DOING GOOD WORK

PR #3112, opened 2019-05-17
Sample data limits

### LINKED ISSUE #2590: Sample AIA images in the gallery do not show SunPy plotting very well

Cadair:
fixes #2590

### DIFF
--- a/docs/guide/tour.rst
+++ b/docs/guide/tour.rst
@@ -123,7 +123,6 @@ plots can be built by combining SunPy with matplotlib.
     aia.plot()
     aia.draw_limb()
     aia.draw_grid()
-    aia.draw_limb()
     plt.colorbar()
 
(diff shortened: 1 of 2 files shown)


nabobalis:
Let's hope they look nicer.

Cadair:
We use them in a few other places I can find, but in more tutorial'y style examples and I don't want to add the extra overhead of clipping in what should be simple examples.

Issue #3110, opened 2019-05-15
Add the ability to correct observer information


ehsteve:
### Description

It is now abundantly clear that many of the data sources that we support do not provide accurate observer information if it is provided at all (see #3057, #3038, #3037).

Thankfully JPL Horizons can provide satellite location information for many (if not all) data sources supported by SunPy. Also, [astroquery](https://astroquery.readthedocs.io/en/latest/jplhorizons/jplhorizons.html) provides a Python interface to that information. There is the limitation that it may not provide information for recent positions (~1 month). I would suggest that the map factory try to correct or add observer information on when the map is instantiated.

### Additional context

Correct observer information is necessary for the coordinate functionality to work accurately.

ayshih:
Query for SOHO's location via `astroquery` would look something like this:
```python
>>> from astropy.time import Time
>>> from astroquery.jplhorizons import Horizons
>>> from astropy.coordinates import SkyCoord, CartesianRepresentation

>>> obstime = Time('2012-06-01 11:22:33')
>>> soho = Horizons(id='SOHO', id_type='majorbody', location='500@0', epochs=obstime.jd)
>>> soho_vector = CartesianRepresentation(soho.vectors()[0]['x', 'y', 'z'])*u.AU
>>> soho_coord = SkyCoord(CartesianRepresentation(soho_vector)*u.AU, frame='icrs', obstime=obstime)
```

Cadair:
> I would suggest that the map factory try to correct or add observer information when the map is instantiated if it is known that the data is missing or likely incorrect.

No.

We could provide a utility for doing this, although I am not in favour of adding an astroquery dep to core. So maybe just an example?

dstansby:
Here's a working example:
```python
import astropy.io

import sunpy.map
from sunpy.coordinates import get_horizons_coord
from sunpy.map.header_helper import get_observer_meta
from sunpy.net import Fido
from sunpy.net import attrs as a

# Download a single LASCO file
result = Fido.search(a.Instrument('lasco'), a.Time('2010-01-01 00:00:00', '2010-01-01 00:10:00'))

# Load, and print the assumed observer coordinate
filepath = Fido.fetch(result[0, 0])[0]
lasco_map = sunpy.map.Map(filepath)
print(lasco_map.observer_coordinate)

# Get actual coordinate using JPL Horizons
soho_coord = get_horizons_coord('SOHO', time=lasco_map.date)
print(soho_coord)

[...]

PR #3113, opened 2019-05-17
Added the capability to query JPL HORIZONS

ayshih:
Here's a function to easily query JPL HORIZONS (via HTTPS) for the location of a solar-system body.  This function requires the astroquery package to be installed.  Be aware that the output for spacecraft may not be accurate.

## Examples of querying by name
```python
>>> from sunpy.coordinates import get_horizons_coord
>>> get_horizons_coord('SOHO', '2011-06-01 11:22:33')
<SkyCoord (HeliographicStonyhurst: obstime=2011-06-01T11:22:33.000): (lon, lat, radius) in (deg, deg, AU)
    (359.74707192, -0.70191324, 1.00451497)>
>>> get_horizons_coord('SDO', '2011-06-01 11:22:33')
<SkyCoord (HeliographicStonyhurst: obstime=2011-06-01T11:22:33.000): (lon, lat, radius) in (deg, deg, AU)
    (0.01222499, -0.65400412, 1.01407332)>
>>> get_horizons_coord('Deimos', '2011-06-01 11:22:33')
<SkyCoord (HeliographicStonyhurst: obstime=2011-06-01T11:22:33.000): (lon, lat, radius) in (deg, deg, AU)
    (137.86655038, 4.67297496, 1.42850571)>
```

## Example of an ambiguous query and a query by ID
```python
>>> get_horizons_coord('STEREO', '2011-06-01 11:22:33')
ValueError: Ambiguous target name; provide unique id:
  ID#      Name                               Designation  IAU/aliases/other
  -------  ---------------------------------- -----------  -------------------
     -234  STEREO-A (spacecraft)                           AHEAD

     -235  STEREO-B (spacecraft)                           BEHIND

  -234900  STEREO Third Stage (spacecraft)
>>> get_horizons_coord(-234, '2011-06-01 11:22:33', id_type='id')
<SkyCoord (HeliographicStonyhurst: obstime=2011-06-01T11:22:33.000): (lon, lat, radius) in (deg, deg, AU)
    (94.27943046, 7.34584532, 0.95739986)>
```

Could help address #3110

### DIFF
--- a/changelog/3113.feature.rst
+++ b/changelog/3113.feature.rst
@@ -1,1 +1,1 @@
-The new function `~sunpy.coordinates.get_horizons_coord()` enables querying JPL HORIZONS for the locations of a wide range of solar-system bodies, including spacecraft.
+The new function `~sunpy.coordinates.get_horizons_coord` enables querying JPL HORIZONS for the locations of a wide range of solar-system bodies, including spacecraft.
--- a/changelog/3113.feature.rst
+++ b/changelog/3113.feature.rst
@@ -1,0 +1,1 @@
+The new function `~sunpy.coordinates.get_horizons_coord()` enables querying JPL HORIZONS for the locations of a wide range of solar-system bodies, including spacecraft.
--- a/sunpy/coordinates/ephemeris.py
+++ b/sunpy/coordinates/ephemeris.py
@@ -299,10 +299,41 @@ def get_horizons_coord(body, time='now', id_type='majorbody'):
     out : `~astropy.coordinates.SkyCoord`
         Location of the solar-system body
 
+    Notes
+    -----
+    Be aware that there can be discrepancies between the coordinates returned by JPL HORIZONS,
+    the coordinates reported in mission data files, and the coordinates returned by
+    `~sunpy.coordinates.get_body_heliographic_stonyhurst()`.
+
     References
     ----------
     * `JPL HORIZONS <https://ssd.jpl.nasa.gov/?horizons>`_
     * `Astroquery <https://astroquery.readthedocs.io/en/latest/>`_
+
+    Examples
+    --------
+    >>> from sunpy.coordinates import get_horizons_coord
+
+    Query the location of Venus
+
+    >>> get_horizons_coord('Venus barycenter', '2001-02-03 04:05:06')  # doctest: +REMOTE_DATA
+    INFO: Obtained JPL HORIZONS location for Venus Barycenter (2) [sunpy.coordinates.ephemeris]
+    <SkyCoord (HeliographicStonyhurst: obstime=2001-02-03T04:05:06.000): (lon, lat, radius) in (deg, deg, AU)
+        (326.06844114, -1.64998481, 0.71915147)>
+
+    Query the location of the SDO spacecraft
+
+    >>> get_horizons_coord('SDO', '2011-11-11 11:11:11')  # doctest: +REMOTE_DATA
+    INFO: Obtained JPL HORIZONS location for Solar Dynamics Observatory (spac [sunpy.coordinates.ephemeris]
+    <SkyCoord (HeliographicStonyhurst: obstime=2011-11-11T11:11:11.000): (lon, lat, radius) in (deg, deg, AU)
+        (0.01018888, 3.29640407, 0.99011042)>
+
+    Query the location of the SOHO spacecraft via its ID number (-21)
+
+    >>> get_horizons_coord(-21, '2004-05-06 11:22:33', 'id')  # doctest: +REMOTE_DATA
+    INFO: Obtained JPL HORIZONS location for SOHO (spacecraft) (-21) [sunpy.coordinates.ephemeris]
+    <SkyCoord (HeliographicStonyhurst: obstime=2004-05-06T11:22:33.000): (lon, lat, radius) in (deg, deg, AU)
+        (0.2523461, -3.55863351, 0.99923086)>
     """
     obstime = parse_time(time)
 
--- a/sunpy/coordinates/ephemeris.py
+++ b/sunpy/coordinates/ephemeris.py
@@ -14,6 +14,17 @@ from astropy.coordinates import (SkyCoord, Angle, Longitude,
 from astropy.coordinates.representation import CartesianRepresentation, SphericalRepresentation
 from astropy._erfa.core import ErfaWarning
 from astropy.constants import c as speed_of_light
+# Versions of Astropy that do not have HeliocentricMeanEcliptic have the same frame
+# with the incorrect name HeliocentricTrueEcliptic
+try:
+    from astropy.coordinates import HeliocentricMeanEcliptic
+except ImportError:
+    from astropy.coordinates import HeliocentricTrueEcliptic as HeliocentricMeanEcliptic
+try:
+    from astroquery.jplhorizons import Horizons
+    ASTROQUERY_PRESENT = True
+except ImportError:
+    ASTROQUERY_PRESENT = False
 
 from sunpy.time import parse_time
 from sunpy import log
@@ -23,7 +34,7 @@ from .transformations import _SUN_DETILT_MATRIX
 
 __all__ = ['get_body_heliographic_stonyhurst', 'get_earth',
            'get_sun_B0', 'get_sun_L0', 'get_sun_P', 'get_sunearth_distance',
-           'get_sun_orientation']
+           'get_sun_orientation', 'get_horizons_coord']
 
 
 def get_body_heliographic_stonyhurst(body, time='now', observer=None):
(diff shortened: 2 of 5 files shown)


sunpy/coordinates/ephemeris.py
```diff
@@ -14,6 +14,17 @@
...
 from astropy._erfa.core import ErfaWarning
 from astropy.constants import c as speed_of_light
+# Versions of Astropy that do not have HeliocentricMeanEcliptic have the same frame
+# with the incorrect name HeliocentricTrueEcliptic
+try:
+    from astropy.coordinates import HeliocentricMeanEcliptic
+except ImportError:
+    from astropy.coordinates import HeliocentricTrueEcliptic as HeliocentricMeanEcliptic
+try:
+    from astroquery.jplhorizons import Horizons
```
nabobalis: We will want to add this as an extra dependency. I am not sure what group.

nabobalis:
I think just a test and a changelog and this is good to go.

sunpy/coordinates/ephemeris.py
Cadair: I am really not sure we do, astroquery is continuously deployed because it's always adapting to upstream changes. It might be best to just have users install this if they need this function?

sunpy/coordinates/ephemeris.py
```diff
@@ -14,6 +14,17 @@
 from astropy.coordinates.representation import CartesianRepresentation, SphericalRepresentation
 from astropy._erfa.core import ErfaWarning
 from astropy.constants import c as speed_of_light
+# Versions of Astropy that do not have HeliocentricMeanEcliptic have the same frame
+# with the incorrect name HeliocentricTrueEcliptic
+try:
+    from astropy.coordinates import HeliocentricMeanEcliptic
+except ImportError:
+    from astropy.coordinates import HeliocentricTrueEcliptic as HeliocentricMeanEcliptic
+try:
```
Cadair: We could just import this in the function? :scream_cat: It seems like it's a bit overkill to do this all the time.

sunpy/coordinates/ephemeris.py
nabobalis: Well if that's the case, I think we should move this import into the function and make a note that it is required top be installed separately.

Cadair: I agree with that.

sunpy/coordinates/tests/test_ephemeris.py
```diff
@@ -111,3 +112,17 @@ def test_get_sun_orientation():
     # Check the Southern Hemisphere
     angle = get_sun_orientation(EarthLocation(lat=-40*u.deg, lon=-75*u.deg), '2017-02-18 13:00')
     assert_quantity_allclose(angle, -110.8*u.deg, atol=0.1*u.deg)
+
+
+@pytest.mark.remote_data
+def test_get_horizons_coord():
```
Cadair: We need a way of marking this test as needs astroquery. Then we should add astroquery to our online test deps.

nabobalis: pytest importer skip?

ayshih: Okay, and where do I add astroquery as a dependency for online tests?

Cadair: In tox.ini?

nabobalis: Yeah that would work. 

Adding it `online: astropyquery` to the last part of the `deps=` list should work.

ayshih:
@Cadair @nabobalis Can one of you explain to me how to fix the test failure? (`OSError: Could not find unix home directory to search for astropy config dir`)

nabobalis:
It's a warning I think. I've seen it before but I didn't it breaks anything.

nabobalis:
Oh it's used for astroquery. I'll have to look at it tomorrow.

nabobalis:
Maybe if we set 

            if 'HOME' in os.environ:
                homedir = os.environ['HOME'] 
In the tox ini  it would work.

nabobalis:
If we add `setenv = HOME = {toxinidir}` it might do something? I will need to pull the PR and test it myself.

nabobalis:
Adding:

```
HOME = {toxinidir}/..
```

in the setenv seems to work locally for me.

nabobalis:
I pushed a fix but it breaks under windows, so maybe @Cadair will tell us the real fix.

ayshih:
@Cadair @nabobalis Fixed the environment variable so tests work.  Approve and merge?

nabobalis:
Theres a homedir for tox? Can't believe I didn't find it. The other changes I made in my commit are actually things we wanted. But I can redo them in a separate PR.

changelog/3113.feature.rst
```diff
@@ -0,0 +1 @@
+The new function `~sunpy.coordinates.get_horizons_coord()` enables querying JPL HORIZONS for the locations of a wide range of solar-system bodies, including spacecraft.
```
Cadair:
```suggestion
The new function `~sunpy.coordinates.get_horizons_coord` enables querying JPL HORIZONS for the locations of a wide range of solar-system bodies, including spacecraft.
```

sunpy/coordinates/ephemeris.py
```diff
@@ -271,3 +277,77 @@ def _sun_north_angle_to_z(frame):
...
+    body : `str`
+        The solar-system body for which to calculate positions
+    id_type : `str`
+        If 'majorbody', search by name for planets or satellites.  If 'id', search by ID number.
+    time : various
+        Time to use as `~astropy.time.Time` or in a parse_time-compatible format
+
+    Returns
+    -------
+    out : `~astropy.coordinates.SkyCoord`
```
Cadair:
```suggestion
    `~astropy.coordinates.SkyCoord`
```

sunpy/coordinates/ephemeris.py
```diff
@@ -271,3 +277,77 @@ def _sun_north_angle_to_z(frame):
...
+        Location of the solar-system body
+
+    Notes
+    -----
+    Be aware that there can be discrepancies between the coordinates returned by JPL HORIZONS,
+    the coordinates reported in mission data files, and the coordinates returned by
+    `~sunpy.coordinates.get_body_heliographic_stonyhurst()`.
```
Cadair:
```suggestion
    `~sunpy.coordinates.get_body_heliographic_stonyhurst`.
```

tox.ini
```diff
@@ -7,6 +7,7 @@ changedir = tmp
 setenv =
     MPLBACKEND = agg
     COLUMNS = 180
+    HOME = {homedir}
```
Cadair: I need to work out a better solution than this.

ayshih: With Astropy's `_find_home()` relying on the `HOME` environment variable (at least currently), I don't know what better solution there could be.  Why do you want to avoid creating this environment variable?

Cadair: I think the best thing to do is to use the [`set_temp_config`](https://github.com/astropy/astropy/blob/master/astropy/config/paths.py#L195) helper in astropy config so that we don't actually write to the users home directory who is running the tests.

The tests should be stateless themselves without having to rely on the clean VMs of the CI service.

ayshih: Okay, I see what you're saying, but doesn't SunPy currently assume that its config is under the user's home directory, so we're already broken?  It seems that you should create a new issue for this.

Cadair: We create a temp sunpy config during test runs.

Cadair: https://github.com/sunpy/sunpy/blob/master/sunpy/conftest.py#L40

ayshih: We do?  [This line](https://github.com/sunpy/sunpy/blob/master/sunpy/util/tests/test_config.py#L20) was causing a test failure when @nabobalis accidentally set `HOME` to an incorrect directory.

ayshih: I guess I don't understand the temp SunPy config if the directories aren't changed

Cadair: :confused: Yeah I don't know either.

sunpy/coordinates/ephemeris.py
```diff
@@ -271,3 +277,80 @@ def _sun_north_angle_to_z(frame):
...
+    obstime = parse_time(time)
+
+    # Import here so that astroquery is not a module-level dependency
+    from astroquery.jplhorizons import Horizons
+    query = Horizons(id=body, id_type=id_type,
+                     location='500@10',      # Heliocentric (mean ecliptic)
+                     epochs=obstime.tdb.jd)  # Time must be provided in JD TDB
+    try:
+        result = query.vectors()
+    except:  # Catch and re-raise all exceptions, and also provide query URL if generated
```
Cadair: It dosen't really matter here, but this should probably be:
```suggestion
    except Exception:  # Catch and re-raise all exceptions, and also provide query URL if generated
```

tox.ini
nabobalis: I think also that failure is due to just being a bad test. Normally on the CI where the clone repo is, is the same place as the user. But on windows for azure, the home user and the config location is are different.

Notice that everything is occurring under `D:\a\1\s\.tox\py36-offline`  where the USER directory is.
But for some reason the CONFIG_DIR is `C:\\Users\\VssAdministrator\\AppData\\Local\\sunpy\\sunpy` 

I suspect this could be a bug in how we set the dir for windows.

ayshih: In that test failure, `USER` is set to `D:\a\1\s\.tox\py36-offline` because you had set `HOME` to `{toxinidir}` instead of `{homedir}`, so it was a one-time error in the configuration

nabobalis: OH it was the test from that. I thought it was the test results  when it was set correctly. Ignore me then.

Cadair:
Great, the CI has died apparently because GH webhooks are having issues.

ayshih:
Yay!

PR #3115, opened 2019-05-20
Change the way observer location is calculated in Map

### LINKED ISSUE #3068: Change how the observer information is generated for Map classes
Observer information for Map classes is carried in FITS headers via a variety of possible keywords, under different coordinate systems, and not necessarily internally consistent.

Currently, the base Map class determines the heliographic Stonyhurst (HGS) longitude and latitude by cascading through a few common keywords, as well as a fallback to heliographic Carrington (HGC) if the HGS keywords are not there.  The `observer_coordinate` property is then constructed from these properties.  A recent PR (#3038) has `AIAMap` reverses this approach: the `observer_coordinate` property is constructed from the Heliocentric Aries Ecliptic keywords, and then the HGS/HGC longitudes and latitude properties are constructed from the `observer_coordinate` property.

Suggestions:
- Change the base Map to follow the approach of constructing the `observer_coordinate` property first, and then deriving the HGS/HGC properties from it.
- Re-factor the base Map class so that derived Map classes can simply specify the keywords to use for the `observer_coordinate` property rather than explicitly overriding the property.

Cadair:
This is essentially a re-do and generalisation of #3056 because of the error @wafels has encountered in #2972 

In #2972 @wafels changes the observer location in the header (to HGS) which dosen't work with an `AIAMap` because we now extract a different set of kwargs from the header and therefore to overwrite the location successfully we would have to write HEA coordinates to the header.

This PR changes AIAMap to default to extracting HEA but to fall back to HGS or HGC if those keys are present but not the previous ones.

This however, causes a change in behaviour of Map because it now checks for all three of the observer location keys *in one go* rather than checking them individually and then defaulting them individually to Earth if they can't be found. This means that if some of the header information is there and some of it isn't before this PR we could end up with an inconsistent observer location, which now wouldn't happen.

fixes #3068

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -28,7 +28,7 @@ from sunpy.time import parse_time, is_time
 from sunpy.image.transform import affine_transform
 from sunpy.image.resample import reshape_image_to_4d_superpixel
 from sunpy.image.resample import resample as sunpy_image_resample
-from sunpy.coordinates import get_sun_B0, get_sun_L0, get_sunearth_distance, get_earth
+from sunpy.coordinates import get_earth
 from sunpy.util.exceptions import SunpyUserWarning
 
 from astropy.nddata import NDData
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -29,6 +29,7 @@ from sunpy.image.transform import affine_transform
 from sunpy.image.resample import reshape_image_to_4d_superpixel
 from sunpy.image.resample import resample as sunpy_image_resample
 from sunpy.coordinates import get_earth
+from sunpy.util import expand_list
 from sunpy.util.exceptions import SunpyUserWarning
 
 from astropy.nddata import NDData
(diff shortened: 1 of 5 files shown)


Cadair:
@wafels I think what you would need to do in #2972 once this is merged is pop the HEA keys from the map meta before changing the HGS ones.

Cadair:
I thought about it a little more and added a helper function for you @wafels

ayshih:
See also #3068!

Cadair:
Ah yes. This fixes that 😂🤣

sunpy/map/mapbase.py
```diff
@@ -562,71 +549,75 @@ def coordinate_system(self):
...
 
-        if heliographic_latitude is None:
-            if self._default_heliographic_latitude is None:
-                warnings.warn("Missing metadata for heliographic latitude: assuming Earth-based observer.",
-                              SunpyUserWarning)
-                self._default_heliographic_latitude = get_sun_B0(self.date)
-            heliographic_latitude = self._default_heliographic_latitude
+        all_keys = [str(e[0]) for e in self._supported_observer_coordinates]
+        all_keys = '\n'.join(all_keys)
+        warning_message = ("Missing metadata for observer: assuming Earth-based observer."
```
dstansby: I think it would be useful to tell the user exactly which metadata is missing here (instead of just telling them that something is missing).

Cadair: It's not trivial to work out what ones are missing, i.e. you could have a weird header with a mix of crln and hglt and it would be hard to work out which one the user was aiming for.

I will give it some thought.

hayesla: I think improved docs on map meta will really help here

Cadair: I opened an issue to keep track of this because I didn't want to block this on it.

PR #3118, opened 2019-05-21
Allow coordinate frames to have no observer set

Cadair:
~(This is post-1.0: While I think this is a bit of an edge case I was thinking about this the other day and wanted to get something together so I didn't forget.)~

It is currently impossible to not have an observer set on a HPC / HCC coordinate frame (which is very possible as @ehsteve @hayesla and I were discussing in context of #3083 ). If you don't explicitly set one (or set it to `None`) it will default to Earth. This PR adds the ability to set the observer to `'none'` which will result in no observer being set on the frame.

While I think this is useful  I am a little concerned by the usability of this feature. We can not have `None` be the "no observer" option without changing the default to be "no observer" however `'none'` feels weird as well.

Finally, I need to work out how the transformations should handle this case. I think what we want is if the observer attribute (different from kwarg) on either side of the transform is `None` then it should raise an error about not being able to do the transform without an observer set.

---

This does form part of the larger discussion that's going on about how and what we should default the observer to. Currently both coordinates and map default to Earth, but is this actually want we want to do? Do we instead want to default to no observer and not let any transformations happen?

### DIFF
--- a/changelog/3118.feature.rst
+++ b/changelog/3118.feature.rst
@@ -1,0 +1,2 @@
+Enable explicitly setting the observer of a coordinate frame to `'none'` if it
+is incorrect to default to Earth.
--- a/sunpy/coordinates/frameattributes.py
+++ b/sunpy/coordinates/frameattributes.py
@@ -136,7 +136,9 @@ class ObserverCoordinateAttribute(CoordinateAttribute):
             # If the observer is a string and we have obstime then calculate
             # the position of the observer.
             if isinstance(observer, str):
-                if obstime is not None:
+                if observer == 'none':
+                    return None
+                elif obstime is not None:
                     new_observer = self._convert_string_to_coord(observer.lower(), obstime)
                     new_observer.object_name = observer
                     setattr(instance, '_' + self.name, new_observer)
--- a/sunpy/coordinates/tests/test_frameattributes.py
+++ b/sunpy/coordinates/tests/test_frameattributes.py
@@ -92,6 +92,16 @@ def test_string_coord():
     assert coord.obstime == parse_time(obstime)
 
 
+def test_none_observer_coord():
+    obstime = "2011-01-01"
+
+    obs = Helioprojective(observer="none", obstime=obstime).observer
+    assert obs is None
+
+    obs = Helioprojective(observer="none", obstime=None).observer
+    assert obs is None
+
+
 def test_coord_get():
 
     # Test default (instance=None)


Cubostar:
If this is added, should I update [my PR](https://github.com/sunpy/sunpy/pull/3262) to mention this feature?

ayshih:
@Cubostar Your PR should be merged well before this one is merged, so don't worry about it.

ayshih:
Note to everyone: as discussed at last dev meeting, please add comments here about the use cases for this functionality, as well as possible confusions for the user.

ehsteve:
I don't like this idea. It would create a special and isolated set of coordinates which can be transformed only to each other. I would much prefer we make it easier to set a default to the observer (like Earth).

ehsteve:
Without an observer the coordinate frame would not be fully specified.

Cadair:
> I would much prefer we make it easier to set a default to the observer (like Earth).

I mean `"earth"` is currently the default. To change the default to other bodies, you just have to set it as a string of the body name, I don't know how much easier you want to make this. :stuck_out_tongue: 

> Without an observer the coordinate frame would not be fully specified.

This is the point. The fact is you can have a HPC frame without knowing where your observer is, it is still valid for use in `Map`. You can use all the wonderful `pixel_to_world` and `rotate` and plotting with HPC grids etc, and never have gone to the effort of locating your observer at all.

I appreciate this is probably a niche that most people wont find themselves in, and I opened this more because I thought it *should* be possible rather than thinking about any specific usecases. Ones I can think of off the top of my head, are things like simulated datasets, if you have a synthetic observation you might not care for an observation time at all. Which means that you *already* can't transform out of HPC (default observer being 'earth' and no obstime set) so why have a default observer at all? Maybe later if you are comparing your synthetic data to observations (maybe many observations with different observers) you will just copy the observer from a different frame.

Anyway, I can speculate about usecases as much as I want, the point is there is theoretical validity in having observer-less HPC frames, so the question is are we going to make people hack this (by not setting obstime) or are we going to explicitly support it?

ehsteve:
Still not completely convinced...not sure how this would work in the machinery but how about we solve this with an explicit SimulationMap class which does not require time or observers and such?

Cadair:
A map subclass could remove the defaults from map, but would have no way of telling the `map.coordinate_frame` attribute to not have an observer set without this PR.

Cadair:
(We should try and keep the scope of this discussion to just the coordinate frames, defaults in map are a separate and even thornier discussion for a later date)

wtbarnes:
> Still not completely convinced...not sure how this would work in the machinery but how about we solve this with an explicit `SimulationMap` class which does not require time or observers and such?

We seem to now be mixing two different issues as @Cadair has pointed out. I don't think the simulated datasets is a very compelling use case and I really don't like this idea of having separate classes for simulations. If you're simulating a particular instrument, you should be using its Map class. If not, why are you using Map? This deserves its own issue though.

I think the most compelling case for having a `None` observer, as @ayshih pointed out in the telecon today, is that it provides a safeguard against users. If you've not specified an observer, you cannot transform out of that frame and it forces the user to think about how their frame is defined before doing any sort of transformation. If we assume `'earth'`, we're silently allowing the user to continue on with a potentially incorrect definition of their coordinate frame.

I'd also argue that having a `None` observer is more in keeping with the "Explicit is better than implicit" part of [PEP 20](https://www.python.org/dev/peps/pep-0020/#the-zen-of-python).

ayshih:
After mulling this over, and having just tangled with default observers in #3291, my position has changed.  Here's where I think we should go:

- Remove any notion of a default observer for our coordinate frames that are observer-based:
  - `observer` is `None` will mean that no observer is defined.
  - Defining `obstime` will not imply anything about the observer unless `observer` is a string instead of `None`.
- The transformation machinery will do the Right Thing<sup>TM</sup> when `observer` is `None`:
  - If both source and destination frames are observer-based and  `observer` is `None` for both frames, the transformation will behave as if the observer is the same between the two (even if the `obstime` is different).  I could see issuing a warning to be safe.
  - If both source and destination frames are observer-based and  `observer` is `None` for one frame but not the other, an exception will be raised.  We don't know what the user intends with an observer mismatch like this.
  - If one frame is observer-based with `observer` is `None` and the other frame is not observer-based, an exception will be raised.  We have no idea how to proceed with the transformation.
- `GenericMap` can continue to have Earth as a default observer if so desired, but it will achieve this by explicitly specifying `observer='earth'` for the frame.  That is, the default-observer behavior will be handled in `sunpy.map` rather than in `sunpy.coordinates`.

ehsteve:
Sorry to muddy the waters a bit. What I was trying to say is that my opinion is that this should only go forward if there is an associated change to define a special map class that defines a kind of boundary for users. Otherwise I don't see the benefit of going forward with this.

Cadair:
@ayshih I am in agreement with that plan, I think it makes a lot of sense, and is an initial decision about coordinates which I regret. My only concern is working out a sensible deprecation path.

@ehsteve This is not only useful for Map, the coordinates infrastructure is used in many places both directly by users and by other code like WCSAxes, reproject etc. The decisions we take here have an effect on Map, but Map is not the only, or even primary, consumer of this code.

ayshih:
I have turned this PR into two issues (#3386 and #3387).

PR #3128, opened 2019-05-23
Fixed uses of a deprecated keyword

ayshih:
#3115 and #3067 used the now-deprecated `representation` keyword for `SkyCoord` instead of `representation_type`


Cadair:
Whoops!

PR #3131, opened 2019-05-23
[1.0] Extra backports for next RC/Release

nabobalis:
Few extra backports.


nabobalis:
```  File "/root/project/examples/time_series/goes_hek_m25.py", line 39, in <module>
    goes.plot()
AttributeError: 'list' object has no attribute 'plot'
```

Look into it and check(?)

nabobalis:
the pandas commit neeeds to be forward ported.

sunpy/timeseries/timeseriesbase.py
```diff
@@ -573,6 +573,8 @@ def to_array(self, **kwargs):
         `~numpy.ndarray`
             If the data is heterogeneous and contains booleans or objects, the result will be of ``dtype=object``.
         """
+        if pd.__version__ < "0.24.0":
```
Cadair: I don't think this version check will work, you probably need to use LooseVersion or whatever the "correct" way of doing version checking is.

nabobalis: I feel like you can do basic string checking like that. It did seem to work locally when I installed pandas 0.23

PR #3134, opened 2019-05-24
Fixes for sunpy.sun.sun

ayshih:
- `sun.print_params()` was broken and nobody noticed.  Fixed and added a simple test
- Docstring tweaks, particularly to functions that don't use our coordinates framework

### DIFF
--- a/sunpy/sun/sun.py
+++ b/sunpy/sun/sun.py
@@ -389,7 +389,7 @@ def print_params(t='now'):
     from sunpy.coordinates.ephemeris import (get_sun_L0, get_sun_B0,
                                              get_sun_P, get_sunearth_distance)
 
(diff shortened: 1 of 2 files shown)


nabobalis:
Thanks @ayshih!

nabobalis:
Backported to 1.0

Cadair:
@ayshih with regards to "this function dosen't use the coordinates framework" messages should we be thinking about deprecating / removing these functions?

ayshih:
> @ayshih with regards to "this function dosen't use the coordinates framework" messages should we be thinking about deprecating / removing these functions?

I plan to re-implement them using the coordinates framework, and I've started doing so, but it's not a 1.0 thing.  The notion would be to not change the outward API, so that would presumably mean not deprecating or removing these functions.

PR #3137, opened 2019-05-26
Re-implementation of `sunpy.sun.sun` functions

### LINKED ISSUE #472: Agree on solar ephemerids calculations
@wafels said: We should unify sun_pos.py and sunpy.sun but agree on which calculation to use on PR #452

### LINKED ISSUE #3021: Missing calculations for sunpy.sun functions
Three functions are missing the mathematics to calculate them.
These seem to be:

* longitude_Sun_perigee
* true_latitude 
* apparent_latitude

We need to track down what they need to calculate from sources and implement that.

ayshih:
I have re-implemented the `sunpy.sun.sun` functions using the coordinates framework and Astropy's underlying machinery (ERFA).  Of note:
- Accuracy is now much improved, and can be compared against published values (e.g., _Astronomical Almanac_) to the <0.1-arcsecond level
- Functions to calculate intermediate products (e.g., mean anomaly) have been deleted since they are non-trivial to re-implement, and folks aren't typically interested in the intermediate products
- Re-defines the output of some functions (`position()` and `true_obliquity_of_ecliptic()`), so definitely a breaking PR, even if the above functions were preserved
- The conversion of ecliptic to equatorial coordinates is explicitly coded.  In principle, the coordinates framework could be used, but the precise frames needed do not (yet) exist in Astropy.
- Curiously, the previous implementations of `true_declination()` and `apparent_declination()` were each bugged, not just inaccurate.
- The RA/dec functions default to providing equinox-of-date values, but have the option to provide J2000.0 values instead.

Closes #472 and closes #3021


sunpy/sun/sun.py
```diff
@@ -51,124 +59,28 @@ def solar_semidiameter_angular_size(t='now'):
...
-
-    Notes
-    -----
-    Based on formulae in "Astronomical Formulae for Calculators" by Meeus
     """
-    ra = true_rightascension(t)
-    dec = true_declination(t)
+    ra = apparent_rightascension(t)
+    dec = apparent_declination(t)
     return ra, dec
```
Cadair: Should this return a `SkyCoord`?

ayshih:
Now even more awesome.  I realized that I can easily use the coordinates framework to get J2000.0 RA/declination, so I added that in (via the boolean keyword `equinox_of_date`).

sunpy/sun/sun.py
ayshih: In principle, yes, but the appropriate frame doesn't currently exist in Astropy.  The frame needs to be `PrecessedGeocentric` plus nutation corrections (i.e., true equinox of date instead of mean equinox of date).  So, for now, a tuple is what we can do.

Cadair:
Thanks @ayshih this is awesome :smile:

nabobalis:
Backported.

Issue #3126, opened 2019-05-23
Map Header Helper should support rotation matrix


Cadair:
I don't know how we managed to miss this the first time around but #3083 should have included the ability to specify the rotation of the pixel grid with respect to the world coordinates. (i.e with an angle or rotation matrix)

hayesla:
what would you invision this to look like? If you give the CROTA or PCi keywords you can still make it into the header

Cadair:
I think a keyword arg (or two) much like `Map.rotate` would be the best thing here where you can specify either a rotation angle and the rotation matrix is calculated from the angle and injected into the header as `PCi_j` or the matrix is just put in as `PCi_j`.

hayesla:
so should we have a keywords in the `make_fitswcs_header` function such that you can call `rotate = xx` in the function call? similar to scale and reference_pixel?

Cadair:
yeah I would think so

hayesla:
groovy, I'll have some time today to work on this

PR #3139, opened 2019-05-27
Adding rotation arguments to the header helper function

hayesla:
### Description
This is a first pass to include the ability to add the rotation information into the map header helper function #3083 following #3126. 

At the moment it's just a patch job which now includes extra keyword arguments for `rotation_angle` (in u.deg) and then either a `rotation_matrix_pc` and `rotation_matrix_cd` for PCi_j and CDi_j rotation matrix. There's probably a nicer way to do this - any ideas?

### DIFF
--- a/changelog/3139.bugfix.rst
+++ b/changelog/3139.bugfix.rst
@@ -1,1 +1,1 @@
-Add support for rotation parameters to `sunpy.map.make_fitswcs_header`
+Add support for rotation parameters to `sunpy.map.make_fitswcs_header`.
--- a/changelog/3139.bugfix.rst
+++ b/changelog/3139.bugfix.rst
@@ -1,0 +1,1 @@
+Add support for rotation parameters to `sunpy.map.make_fitswcs_header`
--- a/sunpy/map/header_helper.py
+++ b/sunpy/map/header_helper.py
@@ -34,8 +34,7 @@ def meta_keywords():
 def make_fitswcs_header(data, coordinate, reference_pixel: u.pix = None,
                         scale: u.arcsec/u.pix = None,
                         rotation_angle: u.deg = None,
-                        rotation_matrix_pc=None,
-                        rotation_matrix_cd=None,
+                        rotation_matrix=None,
                         **kwargs):
     """
     Function to create a FITS-WCS header from a coordinate object
@@ -57,13 +56,12 @@ def make_fitswcs_header(data, coordinate, reference_pixel: u.pix = None,
         expected to be Cartestian ordered, i.e [dx, dy].
         Defaults to ``([1., 1.] arcsec/pixel)``.
     rotation_angle : `~astropy.unit.Quantity`, optional
-        Coordinate system reference angle i.e. the rotation of the pixel array wrt the WCS.
-    rotation_matrix_pc : `~np.ndarray` of dimensions 2x2, optional
+        Coordinate system rotation angle, will be converted to a rotation
+        matrix and stored in the `PCi_j` matrix. Can not be specified with
+        ``rotation_matrix``.
+    rotation_matrix : `~numpy.ndarray` of dimensions 2x2, optional
         Matrix describing the rotation required to align solar North with
-        the top of the image in PCi_j convention.
-    rotation_matrix_cd : `~np.ndarray` of dimensions 2x2, optional
-        Matrix describing the rotation required to align solar North with
-        the top of the image in CDi_j convention.
+        the top of the image in FITS PCi_j convention. Can not be specified with ``rotation_angle``.
     **kwargs:
         Additional arguments that will be put into the metadict header if they
         are in the list returned by `~sunpy.map.meta_keywords`. Additional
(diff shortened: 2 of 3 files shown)


sunpy/map/header_helper.py
```diff
@@ -52,6 +56,14 @@ def make_fitswcs_header(data, coordinate, reference_pixel: u.pix = None,
         Pixel scaling along x and y axis (i.e. the spatial scale of the pixels (dx, dy)). These are
         expected to be Cartestian ordered, i.e [dx, dy].
         Defaults to ``([1., 1.] arcsec/pixel)``.
+    rotation_angle : `~astropy.unit.Quantity`, optional
+        Coordinate system reference angle i.e. the rotation of the pixel array wrt the WCS.
+    rotation_matrix_pc : `~np.ndarray` of dimensions 2x2, optional
+        Matrix describing the rotation required to align solar North with
+        the top of the image in PCi_j convention.
+    rotation_matrix_cd : `~np.ndarray` of dimensions 2x2, optional
```
Cadair: I suggest we don't support this. Map can't support having `CDi_j` without `scale` which means you have the scale infomation duplicated in the matrix and the `CDELT` parameters. So I would just support adding `PCi_j`.

sunpy/map/header_helper.py
```diff
@@ -119,6 +131,15 @@ def make_fitswcs_header(data, coordinate, reference_pixel: u.pix = None,
     meta_wcs['cdelt1'], meta_wcs['cdelt2'] = (scale[0]*meta_wcs['cunit1']/u.pixel,
                                               scale[1]*meta_wcs['cunit2']/u.pixel)
 
+    if rotation_angle is not None:
+        meta_wcs['crota2'] = rotation_angle.to_value()
```
Cadair: We should convert this to a `PCi_j` matrix and store it in the header that way. `CROTA` is deprecated in the FITS standard.

sunpy/map/header_helper.py
```diff
@@ -119,6 +131,15 @@ def make_fitswcs_header(data, coordinate, reference_pixel: u.pix = None,
...
+
+    if rotation_matrix_pc is not None:
```
Cadair: We should check that only `rotation_angle` or `rotation_matrix` is set. We should also validate that the rotation matrix is the right shape.

sunpy/map/header_helper.py
```diff
@@ -52,6 +56,14 @@ def make_fitswcs_header(data, coordinate, reference_pixel: u.pix = None,
         Pixel scaling along x and y axis (i.e. the spatial scale of the pixels (dx, dy)). These are
         expected to be Cartestian ordered, i.e [dx, dy].
         Defaults to ``([1., 1.] arcsec/pixel)``.
+    rotation_angle : `~astropy.unit.Quantity`, optional
+        Coordinate system reference angle i.e. the rotation of the pixel array wrt the WCS.
+    rotation_matrix_pc : `~np.ndarray` of dimensions 2x2, optional
```
Cadair:
```suggestion
    rotation_matrix_pc : `~numpy.ndarray` of dimensions 2x2, optional
```

sunpy/map/header_helper.py
hayesla: ok will fix

sunpy/map/header_helper.py
hayesla: but I guess it's done in mapbase - so would it not just be duplicating the code?

sunpy/map/header_helper.py
hayesla: silly question - what do you meet by 'is set'?

Also yes will add in that check

sunpy/map/header_helper.py
Cadair: yeah it would be, but I think this function should generate the most compliant FITS header it can.

sunpy/map/header_helper.py
Cadair: I mean "is not None" in this case, i.e. just that the user hasn't specified both.

sunpy/map/header_helper.py
```diff
@@ -52,6 +55,13 @@ def make_fitswcs_header(data, coordinate, reference_pixel: u.pix = None,
         Pixel scaling along x and y axis (i.e. the spatial scale of the pixels (dx, dy)). These are
         expected to be Cartestian ordered, i.e [dx, dy].
         Defaults to ``([1., 1.] arcsec/pixel)``.
+    rotation_angle : `~astropy.unit.Quantity`, optional
+        Coordinate system rotation angle, will be converted to a rotation
+        matrix and stored in the `PCi_j` matrix. Can not be specified with
```
nabobalis:
```suggestion
        matrix and stored in the ``PCi_j`` matrix. Can not be specified with
```

nabobalis:
```suggestion
        matrix and stored in the "PCi_j" matrix. Can not be specified with
```
Maybe?

changelog/3139.bugfix.rst
```diff
@@ -0,0 +1 @@
+Add support for rotation parameters to `sunpy.map.make_fitswcs_header`
```
nabobalis:
```suggestion
Add support for rotation parameters to `sunpy.map.make_fitswcs_header`.
```

sunpy/map/header_helper.py
Cadair: I think it will read better in verbatim, but needs to be double back tick?

nabobalis: Either works for me.

nabobalis:
Online test failure is `ValueError: No online HELIO servers can be found.`

nabobalis:
Thanks @hayesla!

nabobalis:
Backported.

PR #3143, opened 2019-05-27
Update the submodule stability matrix

Cadair:


docs/dev_guide/sunpy_stability.yaml
```diff
@@ -1,45 +1,45 @@
...
-  status: stable
-  comments: Stable in 0.9, mature in 1.0.
+  status: mature
+  comments: Any future changes should largely maintain backwards compatibility.
 net:
   status: stable
-  comments: We intend to maintain backwards compatibility post 0.9.
+  comments: Any future changes should largely maintain backwards compatibility.
 physics:
   status: dev
```
Cadair: @wafels what do you think we should put here after the diff rot changes?

nabobalis:
Backported.

PR #3147, opened 2019-05-29
tidy up of the install/bug docs

nabobalis:
I looked over the install and reporting bug part of the guide.


nabobalis:
To be backported.

PR #3149, opened 2019-05-29
Fix differential rotation of map

wafels:
### Description

The current differential rotation warping of maps does not take in to account the location of the observer.  This PR is an attempt to fix this.

Fixes #<Issue Number>


Cadair:
The notebook comparing this routine to the equivalent implementation in reproject is here: https://gist.github.com/Cadair/3606b5aa25059b688fca24e950cfb25a

Cadair:
@wafels let me know if you want to quickly call up and chat about this, or feel free to leave a review on here if you prefer.

wtbarnes:
This need not be resolved before 1.0, but it seems that, now with the fix to account for the different observer and the example provided in @Cadair's notebook, labeling this function as just `differential_rotate` is potentially a bit confusing for users. 

Perhaps I'm confused, but this can now be used for arbitrary reprojection of a map to a new observer right? and differential rotation is just a special case of this in which the new observer's location is determined by the differential rotation of the Sun.

This is also related to the larger question of potential overlap with and dependence on the reproject package, but could this function here be renamed as something a bit more general (e.g. `reproject_map`) and then `differential_rotate` could be a wrapper around `reproject_map` that takes a time (or a time delta), constructs an observer based on that time, and passes it to `reproject_map`?

changelog/3149.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Correctly implement the image warping when transforming Maps for differential
```
wafels: Change to "Implement the image warping when transforming Maps for differential".  This is needed if we are saying the code is experimental.

sunpy/physics/differential_rotation.py
```diff
@@ -360,22 +360,29 @@ def _get_bounding_coordinates(coords):
...
 def _warp_sun_coordinates(xy, smap, new_observer, **diff_rot_kwargs):
     """
-    Helper function that returns a new list of coordinates for each input coord.
-    This is an inverse function needed by `skimage.transform.warp`
+    This function takes pixel coordinates in the warped image (`xy`) and
+    calculates the pixel locations of those pixels in the map.
 
-    function.
+    To do this it converts the input pixel coordinates to helioprojective
+    coordinates as seen by new_observer, then transforms them to Heliographic
```
wafels: "heliographic"

sunpy/physics/differential_rotation.py
```diff
@@ -360,22 +360,29 @@ def _get_bounding_coordinates(coords):
...
 
     Returns
     -------
     xy2 : `numpy.ndarray`
-        Array with the inverse transformation.
+        Pixel coordinates in the map corresponding to the input pixels in the
+        warped image.
     """
     # We start by converting the pixel to world
     with warnings.catch_warnings():
```
wafels: How about adding in an explanation as to why this catch is here?

wafels:
Yes, `differential_rotate` can now be used for arbitrary re-projection (with the caveat that it is still experimental).  I suppose you can look at the current `differential_rotate` as handling the general case when both the observer location and time have changed.  Map re-projection could be a substantial project in a post 1.0 world?

wafels:
Also, +1 on the PR.

PR #3150, opened 2019-05-29
Removed `sunpy.sun.sun.solar_cycle_number`

ayshih:
This function has been broken since its [inception in 2011](https://github.com/sunpy/sunpy/commit/64d9f78aa4d568ded6dbae36d8e6c8863f62057d#diff-5074c05f93f63925e350880e4ca213f9R39).  This was [noted as broken in 2013](https://github.com/sunpy/sunpy/issues/472#issuecomment-20257906) and [re-noted as broken in 2014](https://github.com/sunpy/sunpy/pull/1030#discussion_r13520694), but no one ever did anything about it.  Now I am.  Bye bye!

To be explicit, it's fundamentally broken because:
- The solar-cycle numbers that it returns are nonsensical; they increase by 1 every year and wrap every 28 years.
- Even if the function were fixed to give something close to reasonable (by changing the modulo to a division by a nominal solar-cycle period), we can't possibly return something that's uncontroversial, so why even try?

### DIFF
--- a/changelog/3150.removal.rst
+++ b/changelog/3150.removal.rst
@@ -1,0 +1,1 @@
+Removed `~sunpy.sun.sun.solar_cycle_number` because it was fundamentally flawed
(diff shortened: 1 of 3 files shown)


Cadair:
:fire: :fire: :fire:

PR #3153, opened 2019-05-30
Removal of image utils file due to un_norm and to_norm not being used after #3149

wafels:
### Description

After #3149, this PR should be merged.  This PR removes the image utils file and its test file since the functions they provide (and test) are no longer required anywhere in SunPy after #3149 is accepted.  Until then this PR will likely fail CI.

Fixes #<Issue Number>


wafels:
Needs a changelog

PR #3157, opened 2019-05-30
More improvements to sunpy.sun.sun tests

ayshih:
- Obliquity tests are now against published values instead of just regression-only tests
- Explicitly show the arithmetic expressions from published values rather than just the final value
- A silly-ish test for `position()` to get to 100% test coverage


nabobalis:
STOP SUN IS ALREADY DEAD

nabobalis:
Thanks again @ayshih !

PR #3163, opened 2019-05-31
The new `sunpy.coordinates.sun`

### LINKED ISSUE #3158: Decide the fate  of sunpy.sun.sun or sunpy.coordinates.ephemeris
We have two places for very similar kinds of functions, we should define what goes where to make sure we don't have duplication, and potentially move one into the other.

ayshih:
As proposed in #3158, this PR gathers all functions from `sunpy.sun.sun` and some functions from `sunpy.coordinates.ephemeris` into a new `sunpy.coordinates.sun`.  Of note:

- Deprecation warnings abound.  ~However, do our deprecation warnings not show by default?  I have to turn them on (via `warnings.simplefilter`)~  Fixed by @Cadair 
- To avoid circular imports, I had to leave the `sunpy.coordinates.ephemeris` functions' code where it was.  I renamed the functions to be private functions, and call them in place with deprecated hooks and call them from their future locations with their future names.  It's weird, I know, but it can all be cleaned up once we're past the deprecation period.
- ~I don't understand why the docs are failing (but maybe it's because of my contortions above?)~  Ah, the automatic docs ignore functions that have `__module__` different from the current module because they look to be non-local.

Closes #3158

### DIFF
--- a/sunpy/coordinates/ephemeris.py
+++ b/sunpy/coordinates/ephemeris.py
@@ -23,6 +23,7 @@ except ImportError:
 
 from sunpy.time import parse_time
 from sunpy import log
+from sunpy.util.decorators import deprecated
 
 from .frames import HeliographicStonyhurst as HGS
 from .transformations import _SUN_DETILT_MATRIX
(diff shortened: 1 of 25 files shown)


sunpy/coordinates/sun.py
```diff
@@ -0,0 +1,421 @@
...
+                                                                        B0(t).to_string()))
+    print('Position angle of north pole = {}'.format(P(t)))
+    print('Carrington Rotation Number = {}'.format(carrington_rotation_number(t)))
+
+
+# The code for these functions should be moved in from sunpy.coordinates.ephemeris after the
+# deprecation period
+
+B0 = _B0
+B0.__module__ = __name__  # to be picked up by docs
```
Cadair: Oh the horror. If it works, it works I guess.

Cadair:
@ayshih well spotted on the deprecation warning front, I have fixed that in https://github.com/sunpy/sunpy/pull/3163/commits/421a2b4ad78d31ed5f3a847c59983c9f59ab7269

Cadair:
Grumble, GOES is offline again and it's breaking our doc builds.

ayshih:
To explain, the worst circular imports were due to `sunpy.coordinates.sun` depending on `sunpy.coordinates.ephemeris` on import (to create some constants), but the deprecation hooks in `sunpy.coordinates.ephemeris` needing to point to the new functions in `sunpy.coordinates.sun` also on import.  By using the admittedly weird approach with private functions, where the hooks in `sunpy.coordinates.ephemeris` point within the same module, the dependencies stay in one direction.

Cadair:
The problem with the doc build here is that it's failing right on an example at the top, and then stopping, which means we have no idea if there are errors in the rest of the build :man_facepalming:

sunpy/coordinates/ephemeris.py
```diff
@@ -109,15 +113,110 @@ def get_earth(time='now'):
...
+    solar-system body at a specified time.  This location is the instantaneous or "true" location,
+    and is not corrected for light travel time or observer motion.
+
+    .. note::
+        This function requires the Astroquery package to be installed and
+        requires an Internet connection.
+
+    Parameters
+    ----------
+    body : `str`
```
wafels: Is there a list of strings somewhere that identify the bodies?  Might be good to point to that.

sunpy/coordinates/ephemeris.py
```diff
@@ -109,15 +113,110 @@ def get_earth(time='now'):
...
+        The solar-system body for which to calculate positions
+    id_type : `str`
+        If 'majorbody', search by name for planets or satellites.  If 'id', search by ID number.
```
wafels: Same here - what does ID number mean and where can I find out what it means and what are acceptable values?

sunpy/coordinates/ephemeris.py
ayshih: Unfortunately, nothing that can be easily referenced (that I have found).  JPL HORIZONS's web form to search through their database can't be linked to directly.  Honestly, just asking for what you want is typically the best way to find out if it's there or not.

ayshih: Okay, I take that back.  It looks like I can link reasonably to their web form, so that can be added to the docstring.

sunpy/coordinates/sun.py
```diff
@@ -0,0 +1,394 @@
...
+        If True, output is referred to the true equinox of date.  Otherwise, output is referred to
+        the J2000.0 epoch.
+    """
+    ra = apparent_rightascension(t, equinox_of_date=equinox_of_date)
+    dec = apparent_declination(t, equinox_of_date=equinox_of_date)
+    return ra, dec
+
+
+@add_common_docstring(**_variables_for_parse_time_docstring())
+def carrington_rotation_number(t='now'):
```
wafels: Minor - can this function be considered a coordinate?  Would it be better off somewhere else?

wafels: I suppose it is a coordinate in the sense that time is a coordinate....

wafels: Maybe we can re-write this in Astropy time?

ayshih: I'd argue that this is a coordinate, since it's intimately related to the heliographic longitude of disk center (L0).

wafels: ok

wafels:
Other than my bigger complaint about 'now', +1!

ayshih:
> Other than my bigger complaint about 'now', +1!

Heh, good enough!

I'll add the JPL HORIZONS web form to docstring.

PR #3178, opened 2019-06-02
Fixed Carrington longitude and Carrington rotation number

### LINKED ISSUE #1646: Carrington rotation number
The SunPy and SSWIDL Carrington rotation numbers differ, and the difference is significant.

In [2]: import sunpy.sun
In [3]: sunpy.sun.carrington_rotation_number('2015-12-25 11:57:00')
Out[3]: 2172.011791498784

In [4]: sunpy.sun.carrington_rotation_number('2015-12-25 04:13:54')
Out[4]: 2172.0000007128997

IDL> print, tim2carr('2015-12-25 11:57:00',/dc)
       2172.0001

For reference, 
http://alpo-astronomy.org/solar/rotn_nos.html
http://umtof.umd.edu/pm/crn/

The SunPy Carrington rotation numbers appear to be the minority opinion.

### LINKED ISSUE #2224: Resolve different ways to calculate Carrington longitude

ayshih:
Welp, I didn't quite manage to get this fix into 1.0, but now it's done!  The calculations of Carrington longitude and Carrington rotation number have been fixed to use modern definitions, and the outputs match the *Astronomical Almanac*.  If there are discrepancies with other sources (e.g., SSW), it's likely those other sources are doing it wrong.

Closes #2224 and closes #1646

### DIFF
--- a/changelog/3178.bugfix.rst
+++ b/changelog/3178.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed accuracy issues with the calculations of Carrington longitude (`~sunpy.coordinates.sun.L0`) and Carrington rotation number (`~sunpy.coordinates.sun.carrington_rotation_number`).
--- a/docs/code_ref/coordinates.rst
+++ b/docs/code_ref/coordinates.rst
@@ -131,7 +131,7 @@ coordinates is::
        (0., 0.)>
    >>> c.transform_to(frames.HeliographicCarrington)
    <SkyCoord (HeliographicCarrington: obstime=2017-07-26T00:00:00.000): (lon, lat, radius) in (deg, deg, km)
-      (283.99298362, 5.31701821, 695700.00000125)>
+      (283.95274241, 5.31701821, 695700.00000125)>
 
 It is also possible to transform to any coordinate system implemented in Astropy. This can be used to find the position of the solar limb in AltAz equatorial coordinates::
 
--- a/sunpy/coordinates/ephemeris.py
+++ b/sunpy/coordinates/ephemeris.py
@@ -269,7 +269,7 @@ def _L0(time='now'):
 
     Returns
     -------
-    ~astropy.coordinates.Longitude`
+    `~astropy.coordinates.Longitude`
         The Carrington longitude
 
     Notes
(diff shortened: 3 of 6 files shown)


ayshih:
I don't understand why the CI is failing, and I can't reproduce it locally.  Help?

nabobalis:
Restarted the CI, it seems to pass now.

nabobalis:
Online failure due to missing database package that should be part of the standard library. I think we can ignore.

PR #3184, opened 2019-06-03
Fix two *small* errors in header helper

### LINKED ISSUE #3179: Back (or double?) great circles drawn when used `draw_grid` from any position.
### Description

We get the lines behind the sun when using `draw_grid`.

### Expected behavior

That only the front lines ares shown and the backs are not drawn! (or at least @Cadair said: *weeeird, bug report time*)

### Actual behavior

From SDO:
Earth_sun

From other places:
NoEarth_sun

### Steps to Reproduce

1. Get the observer location
```python
sdo_pos = get_horizons_coord('SDO', '2019-03-14 00:10:00')
```

2. Generate a map from such observer
```python
def sun_seen_from(position, name):
    
    # Find out the size of the sun
    solar_r = (constants.radius / position.radius).to(u.arcsec, equivalencies=u.dimensionless_angles())
    th = np.linspace(0 * u.deg, 360 * u.deg)

    # Create a FD map from there
    half_image = 100
    datax, datay = np.mgrid[-half_image:half_image, -half_image:half_image]
    data = 1 - np.sqrt(datax**2 + datay**2)
    coord = SkyCoord(0*u.arcsec, 0*u.arcsec, obstime=position.obstime, observer=position, frame=frames.Helioprojective)
    size = solar_r / half_image / u.pixel
    header = sunpy.map.header_helper.make_fitswcs_header(data, coord,
                                                         reference_pixel=u.Quantity([half_image, half_image]*u.pixel),
                                                         scale=u.Quantity([size, size]),
                                                         instrument='sunlike', detector=f'Sun seen from {name}')

    return sunpy.map.Map(data, header)


sdo_map = sun_seen_from(sdo_pos, name='SDO')
```

3. Plot that map with the `draw_grid`
```python
fig = plt.figure(figsize=(7,8))
ax = fig.add_subplot(1, 1, 1, projection=sdo_test_map)
sdo_test_map.plot(axes=ax)
sdo_test_map.draw_grid(axes=ax)
fig.savefig('Earth_sun.png')
plt.show()
```
4. Repeat for other positions around the heliosphere.

### System Details

  - SunPy Version: v1.0 :tada:
  - Python Version: 3.6.7
  - OS information: Linux using miniconda (conda v.4.5.11)

Cadair:
Fixes #3179

### DIFF
--- a/sunpy/map/header_helper.py
+++ b/sunpy/map/header_helper.py
@@ -50,7 +50,7 @@ def make_fitswcs_header(data, coordinate, reference_pixel: u.pix = None,
     reference_pixel :`~astropy.units.Quantity` of size 2, optional
         Reference pixel along each axis. These are expected to be Cartestian ordered, i.e
         the first index is the x axis, second index is the y axis. Defaults to
-        the center of data array, ``(data.shape[1] + 1)/2., (data.shape[0] + 1)/2.)``,
+        the center of data array, ``(data.shape[1] - 1)/2., (data.shape[0] - 1)/2.)``,
         this argument is zero indexed (Python convention) not 1 indexed (FITS
         convention).
     scale : `~astropy.units.Quantity` of size 2, optional
--- a/sunpy/map/header_helper.py
+++ b/sunpy/map/header_helper.py
@@ -124,8 +124,8 @@ def make_fitswcs_header(data, coordinate, reference_pixel: u.pix = None,
     meta_wcs['crval1'], meta_wcs['crval2'] = (coordinate.spherical.lat.to_value(meta_wcs['cunit1']),
                                               coordinate.spherical.lon.to_value(meta_wcs['cunit2']))
 
-    meta_wcs['crpix1'], meta_wcs['crpix2'] = (reference_pixel[0].to_value(u.pixel),
-                                              reference_pixel[1].to_value(u.pixel))
+    meta_wcs['crpix1'], meta_wcs['crpix2'] = (reference_pixel[0].to_value(u.pixel) + 1,
+                                              reference_pixel[1].to_value(u.pixel) + 1)
 
     meta_wcs['cdelt1'], meta_wcs['cdelt2'] = (scale[0].to_value(meta_wcs['cunit1']/u.pixel),
                                               scale[1].to_value(meta_wcs['cunit2']/u.pixel))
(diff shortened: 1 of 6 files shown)


sunpy/map/header_helper.py
```diff
@@ -124,8 +124,8 @@ def make_fitswcs_header(data, coordinate, reference_pixel: u.pix = None,
     meta_wcs['crval1'], meta_wcs['crval2'] = (coordinate.spherical.lat.to_value(meta_wcs['cunit1']),
                                               coordinate.spherical.lon.to_value(meta_wcs['cunit2']))
 
-    meta_wcs['crpix1'], meta_wcs['crpix2'] = (reference_pixel[0].to_value(u.pixel),
-                                              reference_pixel[1].to_value(u.pixel))
+    meta_wcs['crpix1'], meta_wcs['crpix2'] = (reference_pixel[0].to_value(u.pixel) + 1,
+                                              reference_pixel[1].to_value(u.pixel) + 1)
```
hayesla: what if someone assumes this already when they add in what the reference pixel should be?

Cadair: We need to document this clearer if we expect this to be 1 indexed or 0 indexed. We definitely need to fix it if the user doesn't specify it. I am not sure however if we should expect the user to give us these values 0 or 1 indexed.

This is the only kwarg we need to make this decision on though.

sunpy/map/header_helper.py
```diff
@@ -49,8 +49,10 @@ def make_fitswcs_header(data, coordinate, reference_pixel: u.pix = None,
         Coordinate object to get meta information for map header.
     reference_pixel :`~astropy.units.Quantity` of size 2, optional
         Reference pixel along each axis. These are expected to be Cartestian ordered, i.e
-        the first index is the x axis, second index is the y axis.
-        Defaults to the center of data array, ``(data.shape[1] + 1)/2., (data.shape[0] + 1)/2.)``.
+        the first index is the x axis, second index is the y axis. Defaults to
+        the center of data array, ``(data.shape[1] + 1)/2., (data.shape[0] + 1)/2.)``,
```
ayshih:
```suggestion
        the center of data array, ``(data.shape[1] - 1)/2., (data.shape[0] - 1)/2.)``,
```
If we're asking for the input to be zero indexed, then the default should be represented as zero indexed too.

nabobalis:
Rebased, fixed the doc test issues and snuck in a doc dev update.

nabobalis:
Thanks @Cadair

PR #3186, opened 2019-06-04
Added missing HGC<->HGC loopback transformation

ayshih:
Yes, my objective here is to fix the one missing arrow in our transformation graph.  I don't think people are rushing out to use this transformation.


nabobalis:
Will we want a test for this?

ayshih:
Good point, added tests

PR #3190, opened 2019-06-05
Bugfix in differential rotation when passing time and sign error

### LINKED ISSUE #3189: Differential rotation throws an error when passed a time
### Description
When passing a time or time interval to `sunpy.physics.differential_rotation.differential_rotate`, an error is thrown due to the new observer not being in the correct coordinate frame.

A sample code snippet,
```python
import astropy.units as u
from sunpy.map import Map
from sunpy.physics.differential_rotation import differential_rotate
m = Map('/path/to/my_map.fits')
m_rot = differential_rotate(m, 1*u.day)
```

This error is almost certainly due to the fact that the new observer constructed from this new time is in the GCRS reference frame and thus has no lon/lat attributes.

### Expected behavior
Should return a `Map` `m_rot` which has been warped to account for the differential rotation on the Sun over 1 day.

### Actual behavior
```python-traceback
AttributeError                            Traceback (most recent call last)
<ipython-input-17-d7d9f076cb62> in <module>()
----> 1 m_rot = differential_rotate(m,time=1*u.day)

~/anaconda/envs/sunpy-paper/lib/python3.6/site-packages/sunpy/physics/differential_rotation.py in differential_rotate(smap, observer, time, **diff_rot_kwargs)
    580
    581     # Add a new HGS observer
--> 582     out_meta['hglt_obs'] = new_observer.lat.value
    583     out_meta['hgln_obs'] = new_observer.lon.value
    584     out_meta['dsun_obs'] = new_observer.radius.to(u.m).value

~/anaconda/envs/sunpy-paper/lib/python3.6/site-packages/astropy/coordinates/sky_coordinate.py in __getattr__(self, attr)
    598         # Fail
    599         raise AttributeError("'{0}' object has no attribute '{1}'"
--> 600                              .format(self.__class__.__name__, attr))
    601
    602     def __setattr__(self, attr, val):

AttributeError: 'SkyCoord' object has no attribute 'lat'
```

### System Details

  - SunPy Version: 1.0.0
  - Python Version: 3.6.5
  - OS information: macOS 10.14.5

wtbarnes:
Fixes #3189 

This PR addresses two issues with `sunpy.physics.differential_rotation.differential_rotate`:

* an error was thrown when passing a time because the new observer was not in HGS (see #3189 )
* images were being warped the opposite direction due to a sign error

Regarding the second point, my only evidence for this fix being the right thing to do is that the resulting images now look "correct." It could be that there is a larger issue here.


wtbarnes:
Given the test failure, I think something else is going on Re: point 2.

Cadair:
Those test failures are just testing the previous output for regression, I don't think they are indicative of errors.

wtbarnes:
@Cadair Ok that's kind of what I suspected. How should that test be modified?

Cadair:
Yeah something isn't right here...

wtbarnes:
I've really muddied this PR so I'm closing it opening a new one on a different branch.

PR #3191, opened 2019-06-06
Culled-down version of the transformation graph for our docs

### LINKED ISSUE #2846: Remove some coordinate frames from transformation graph in docs?
Thanks to yours truly, it's now possible to create a custom coordinate transformation graph for documentation as of astropy 3.1: https://github.com/astropy/astropy/pull/7135

It might be nice to remove some of the non-solar physics coordinate systems in the current graph to make it more readable? Current graph is here: http://docs.sunpy.org/en/stable/code_ref/coordinates.html#module-sunpy.coordinates.transformations

ayshih:
- Removes most of the astronomical frames from the diagram in the docs
- Adds a box for "Other frames in Astropy"
- Edits the description text before the diagram because the default talks only about Astropy
- Moves the description and graph from `sunpy.coordinates.transform` to `sunpy.coordinates`

Here's what the graph looks like:

graph

Closes #2846

### DIFF
--- a/sunpy/coordinates/__init__.py
+++ b/sunpy/coordinates/__init__.py
@@ -20,10 +20,9 @@ below.  See the documentation for `astropy.coordinates` for more information.
 from .frames import *
 from .offset_frame import *
 from . import transformations
-from .transformations import _make_sunpy_graph
 from .ephemeris import *
 from . import sun
 
 from .wcs_utils import *
 
-__doc__ += _make_sunpy_graph()
+__doc__ += transformations._make_sunpy_graph()
--- a/sunpy/coordinates/__init__.py
+++ b/sunpy/coordinates/__init__.py
@@ -1,7 +1,29 @@
+"""
+This subpackage contains:
+
+* A robust framework for working with coordinate systems
+* Functions to obtain the locations of solar-system bodies
+* Functions to calculate Sun-specific coordinate information
+  (`sunpy.coordinates.sun`)
+
+The diagram below shows all of Sun-based and Earth-based coordinate systems
+available through `sunpy.coordinates`, as well as the transformations between
+them.  Each frame is labeled with the name of its class and its alias (useful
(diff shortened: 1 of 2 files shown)


ayshih:
Built off of #3186, so merge that first

sunpy/coordinates/transformations.py
```diff
@@ -371,4 +388,91 @@ def hcc_to_hcc(hcccoord, hccframe):
...
+        pass
+
+    @graph.transform(FunctionTransform, ICRS, Astropy)
+    def fake_transform2():
+        pass
+
+
+def _tweak_graph(docstr):
+    # Edit the diagram description
+    output = docstr.replace('all of the coordinate systems built into the\n`~astropy.coordinates` '
```
nabobalis: Do we need `\n`?

sunpy/coordinates/__init__.py
```diff
@@ -1,7 +1,29 @@
...
+The frames colored in cyan are implemented in `astropy.coordinates`, and there
+are other astronomical frames that can be transformed to that are not shown
+below.  See the documentation for `astropy.coordinates` for more information.
+
+"""
+
 from .frames import *
 from .offset_frame import *
 from . import transformations
+from .transformations import _make_sunpy_graph
```
Cadair: Why import this like this when we are already importing `transformations` above?

sunpy/coordinates/__init__.py
```diff
@@ -1,7 +1,29 @@
...
 from .ephemeris import *
 from . import sun
 
 from .wcs_utils import *
+
+__doc__ += _make_sunpy_graph()
```
Cadair:
```suggestion
__doc__ += transformations._make_sunpy_graph()
```

sunpy/coordinates/__init__.py
ayshih: Because I can't read. =P  Fixed.

PR #3197, opened 2019-06-07
Test tweaks for `sunpy.coordinates.sun`

ayshih:
It turns out that our accuracy is even better than I thought.  I was looking in the wrong place for the high-precision obliquity numbers, and I got mixed up about degrees versus hour angles.


nabobalis:
Thanks again!

PR #3203, opened 2019-06-10
Added tests that check against SunSPICE transformations

ayshih:
I've added tests to explicitly check that we give the same answers as SunSPICE for the transformations between our frames.  We should still check against published values (e.g., the *Astronomical Alamanc*) when possible, but SunSPICE enables additional testing coverage.


Cadair:
This looks great, it's nice to have a concrete reference for something for the HCC -> HGS conversion. Do you know if you could calculate these numbers with HelioPy?

ayshih:
> This looks great, it's nice to have a concrete reference for something for the HCC -> HGS conversion. Do you know if you could calculate these numbers with HelioPy?

I could be wrong, but I don't think it's possible yet to do so in HelioPy.  I imagine that it's possible to do so using spicepy, but SunSPICE is much easier to use.

nabobalis:
Thanks @ayshih

PR #3210, opened 2019-06-12
pin astropy<3.2 for RTD build

nabobalis:
Let us test 3.2 as well.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,9 +7,9 @@ variables:
 jobs:
 - template: azure-templates.yml
   parameters:
-    name: Linux_37_offline
+    name: Linux_36_offline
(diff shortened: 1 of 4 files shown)


azure-pipelines.yml
```diff
@@ -7,9 +7,9 @@ variables:
 jobs:
 - template: azure-templates.yml
   parameters:
-    name: Linux_37_offline
+    name: Linux_36_offline
```
Cadair: ooi, why did you switch the python versions over?

nabobalis: It was to avoid the DB issue.

nabobalis: So I switched it back to normal

PR #3212, opened 2019-06-12
New coordinate frames! (HEE, GSE, HCI, and GEI)

ayshih:
I've added four coordinate frames:

- Heliocentric Earth Ecliptic (HEE)
- Geocentric Solar Ecliptic (GSE)
- Heliocentric Inertial (HCI)
- Geocentric Earth Equatorial (GEI)

All transformations are implemented and are tested against SunSPICE (see #3203).  Here's the transformation graph (with the frame names of Astropy 3.2.x):

graph

Miscellaneous notes:

- `GeocentricSolarEcliptic` and `GeocentricEarthEquatorial` are transformed through heliocentric frames because of how Astropy handles aberration due to Earth motion.
- ~SunSPICE doesn't appear to be accounting for Earth's ecliptic latitude (very small but non-zero) when transforming from HEE to GSE.  I need to chat with Bill Thompson.~  I hadn't implemented HEE and GSE quite correctly, but that is now fixed.
- ~The implementation for longitude wrap angles has been tweaked for the better.~  Moved to separate PR (#3223)
- `GeocentricEarthEquatorial` defaults to J2000.0 (which is what SunSPICE assumes), but one can also use it with the mean equinox of date (i.e., GEI_D) by specifying `equinox` with the same time as `obstime`.

### DIFF
--- a/sunpy/coordinates/frames.py
+++ b/sunpy/coordinates/frames.py
@@ -12,6 +12,7 @@ from astropy.coordinates.baseframe import BaseCoordinateFrame, RepresentationMap
 from astropy.coordinates.representation import (CartesianRepresentation, SphericalRepresentation,
                                                 CylindricalRepresentation,
                                                 UnitSphericalRepresentation)
+from astropy.time import Time
 
 from sunpy.sun.constants import radius as _RSUN
 from sunpy.util.decorators import add_common_docstring
@@ -21,6 +22,8 @@ from .frameattributes import TimeFrameAttributeSunPy, ObserverCoordinateAttribut
 
 from sunpy.util.decorators import add_common_docstring
 from sunpy.time.time import _variables_for_parse_time_docstring
+
+_J2000 = Time('J2000.0', scale='tt')
 
 __all__ = ['HeliographicStonyhurst', 'HeliographicCarrington',
            'Heliocentric', 'Helioprojective',
@@ -556,7 +559,7 @@ class GeocentricEarthEquatorial(SunPyBaseCoordinateFrame):
 
     - The origin is the center of the Earth
     - The z-axis is aligned with the Earth's rotation axis
-    - The x-axis is aligned with the vernal equinox (mean J2000.0)
+    - The x-axis is aligned with the mean (not true) vernal equinox
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


Cadair:
You really are on a mission @ayshih !!

ayshih:
> * SunSPICE doesn't appear to be accounting for Earth's ecliptic latitude (very small but non-zero) when transforming from HEE to GSE.  I need to chat with Bill Thompson.

Ah, I think I now understand how HEE and GSE are defined slightly differently from how I implemented them.  Stay tuned.

ayshih:
@Cadair I've changed how longitude wrap angles are handled in our frames after seeing this suggestion: https://github.com/astropy/astropy/issues/7462#issuecomment-482694086. The existing code works in basic cases, but it falls apart once transformations get involved (including when I tried rewriting the HGS<->HGC transformation as a rotation).

Cadair:
I will put this on my review queue to go through properly, but it will be a while before I find the time to wrap my head around these frames and their transforms.

ayshih:
@Cadair I'd recommend that you don't go too deep down the rabbit hole of the math of the transformations – the tests against SunSPICE are a rather good safety net there – and stick as much as possible to reviewing just the implementation.  But, if you do want to really chew on the math, focus on the transformation between `HeliocentricMeanEcliptic` and `HeliocentricInertial` because the discrepancies versus SunSPICE are notably larger there (although still sub-arcsecond).

ayshih:
I've rebased this PR

ayshih:
@Cadair Review away!  I've added GEI and did a bunch of cleanup.

sunpy/coordinates/frames.py
```diff
@@ -432,3 +439,122 @@ def calculate_distance(self):
...
+    """
+    A coordinate or frame in the Heliocentric Earth Ecliptic system.
+
+    - The origin is the center of the Sun
+    - The x-axis is aligned with the Sun-Earth vector
+    - The z-axis is aligned with the component of the mean ecliptic pole at the observation time
+
+    Parameters
+    ----------
+    data: `~astropy.coordinates.BaseRepresentation` subclass instance
```
Cadair: All these parameters are shared between the frames, why don't you feed them into `add_common_docstring` to prevent from repeating yourself.

ayshih: Sure, I created a new issue (#3314) for this.

Cadair:
This (as a new feature) needs a second review.

sunpy/coordinates/frames.py
ayshih: To be clear, I want to implement this in a new PR because I want the approach to cover the other frames as well.

Cadair: Yeah I figured, seems wise.

ayshih:
Hooray!

Cadair:
Thanks for this awesome addition @ayshih :smile:

PR #3217, opened 2019-06-14
SDO Map sources header cleanup

Cadair:


sunpy/map/sources/sdo.py
```diff
@@ -107,24 +103,22 @@ def __init__(self, data, header, **kwargs):
 
         GenericMap.__init__(self, data, header, **kwargs)
 
-        self.meta['detector'] = "HMI"
-#        self.meta['instrme'] = "HMI"
-#        self.meta['obsrvtry'] = "SDO"
+        self.meta['detector'] = self.meta.get('detector') or "HMI"
```
nabobalis: Would `self.meta.get('detector', "HMI")` not work?

Cadair: Yes that would be much better.

nabobalis:
Thanks @Cadair

PR #3218, opened 2019-06-14
Support custom projection in make_fitswcs_header

Cadair:


sunpy/map/header_helper.py
```diff
@@ -182,13 +189,19 @@ def _get_wcs_meta(coordinate):
...
+def get_observer_meta(observer, rsun: u.Mm):
     """
     Function to get observer meta from coordinate frame.
 
     Parameters
     ----------
     coordinate : ~`astropy.coordinates.BaseFrame`
+        The coordinate of the observer, must be transformable to Heliographic
+        Stonyhurst.
+
```
nabobalis:
```suggestion
```

sunpy/map/header_helper.py
```diff
@@ -110,10 +117,10 @@ def make_fitswcs_header(data, coordinate, reference_pixel: u.pix = None,
     if isinstance(coordinate, frames.Heliocentric):
         raise ValueError("This function does not currently support heliocentric coordinates.")
 
-    meta_wcs = _get_wcs_meta(coordinate)
+    meta_wcs = _get_wcs_meta(coordinate, projection_code)
 
     if hasattr(coordinate, "observer") and isinstance(coordinate.observer, frames.BaseCoordinateFrame):
-        meta_observer = _get_observer_meta(coordinate)
+        meta_observer = get_observer_meta(coordinate.observer, coordinate.rsun)
```
hayesla: is there a reason that coordinate.rsun is pulled out here? just curious why the coordinate isn't just passed

sunpy/map/header_helper.py
```diff
@@ -197,14 +209,16 @@ def _get_observer_meta(coordinate):
             * hgln_obs, hglt_obs
             * dsun_obs
             * rsun_obs
+            * rsun_ref
     """
+    observer = observer.transform_to(frames.HeliographicStonyhurst(obstime=observer.obstime))
```
hayesla: is this just to make life easier when pulling out meta?

sunpy/map/header_helper.py
Cadair: The main reason I wanted to do this was to make the `get_observer_meta` useful in isolation. It might be occasionally useful to convert a SkyCoord for observer to header keys, for instance in the differential rotation code etc.

sunpy/map/header_helper.py
Cadair: This is to ensure that we are always in HGS before we build the header (which when coordinate was passed in was kind of enforced by the frame).

nabobalis:
Thanks @Cadair and @hayesla for your review!

PR #3219, opened 2019-06-14
Fix the detection of distance in Helioprojective.calculate_distance

Cadair:
This is a fix to the behaviour of the HelioProjective frame if you are for some weird reason, trying to instantiate a frame with a non-spherical representation.


Cadair:
The online and 32bit fails are being talked about / fixed in #3221

PR #3223, opened 2019-06-17
Improved handling of longitude wrap angles in frames

ayshih:
(This was originally part of #3212, but I feel that it qualifies as a bug fix to get into 1.0.2 rather than wait until 1.1.)

We intend to have all of our frames wrap longitudes at 180 degrees (i.e., the range is -180 deg to +180 deg, rather than 0 deg to +360 deg).  Prior to this PR, the longitude wrap angle would get set to the desired value if a frame was created with a `(Unit)SphericalRepresentation`.  However, that meant that if the frame was created with a `CartesianRepresentation`, and then *converted* to a `SphericalRepresentation`, the longitude wrap angle would not get set.  This happens frequently after frame transformations, such as for `AIAMap.observer_coordinate`, which gets converted from HEC coordinates and often shows HGS longitudes of ~359.99 deg rather than a small negative number.

This PR takes a different approach.  Rather than modifying the representation as it goes into the frame, the approach is to modify the representation as it comes out.  That is, the internal data can be whatever, but whenever the user asks for it as a `(Unit)SphericalRepresentation`, it is provided with the desired wrap angle.

### DIFF
--- a/sunpy/coordinates/frames.py
+++ b/sunpy/coordinates/frames.py
@@ -23,12 +23,30 @@ __all__ = ['HeliographicStonyhurst', 'HeliographicCarrington',
 
 class SunPyBaseCoordinateFrame(BaseCoordinateFrame):
     """
-    Inject a nice way of representing the object which the coordinate represents.
+    * Defines a default longitude wrap angle of 180 degrees, which can be overridden by the class
+      variable `_wrap_angle`.
+    * Inject a nice way of representing the object which the coordinate represents.
     """
+    _wrap_angle = 180*u.deg
 
     def __init__(self, *args, **kwargs):
         self.object_name = None
+
+        # If wrap_longitude=False is passed in, do not impose a specific wrap angle for the frame
+        if not kwargs.pop('wrap_longitude', True):
+            self._wrap_angle = None
+
         return super().__init__(*args, **kwargs)
+
+    def represent_as(self, base, s='base', in_frame_units=False):
+        """
+        If a frame wrap angle is set, use that wrap angle for any spherical representations.
+        """
+        data = super().represent_as(base, s, in_frame_units=in_frame_units)
+        if self._wrap_angle is not None and \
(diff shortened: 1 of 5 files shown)


sunpy/coordinates/frames.py
```diff
@@ -23,13 +23,31 @@
...
+            self._wrap_angle = None
+
         return super().__init__(*args, **kwargs)
 
+    def represent_as(self, base, s='base', in_frame_units=False):
+        """
+        If a frame wrap angle is set, use that wrap angle for any spherical representations.
+        """
+        data = super().represent_as(base, s, in_frame_units=in_frame_units)
+        if self._wrap_angle is not None and \
```
Cadair: line break using `()` rather than `\`?

nabobalis:
This fixes the 32bit build error!

ayshih:
> This looks like a clever way of doing it. Presuming that `represent_as` is always called when we need it to be?

It bears repeating in this PR that the cleverness is not mine; I copied it from here: https://github.com/astropy/astropy/issues/7462#issuecomment-482694086.

As far as I've seen, yes, `represent_as` is called when needed.  It *isn't* called when using `BaseCoordinateFrame.data` (i.e., asking for the raw coordinate data), but arguably a user shouldn't rely on the raw coordinate data having a particular representation, much less wrap angle, and should always ask for the desired representation.

sunpy/coordinates/frames.py
ayshih: I originally had it that way, but when using the `(`, the way that I want to indent the following line becomes exactly 4 spaces, so it's easy to confuse for an actual indention (and the bot complained as well).  So, I changed it to the `\` so that I could indent the folllowing line only 3 spaces and break that confusion.  It's admittedly silly.  How would you indent it?

Cadair: Fair enough. I just have a pathological hatred for `\` lol

PR #3228, opened 2019-06-18
Clarifying `astroquery` dependency in gallery example

MSKirk:
### Description
Added clarifying hyperlinks to the gallery example `Setting the correct position for SOHO in a LASCO C3 Map` to link to `astroquery` docs page. Explained the dependency in the first paragraph.


nabobalis:
Thanks @MSKirk !

PR #3229, opened 2019-06-18
Remove pytest warnings capture

dstansby:
xref #2886 - I think the time may have come to face up to the test suite warnings, and not just silently ignore them. There's still a handful to fix, but maybe others will feel more motivated to fix them if they see them in their test logs 😉


nabobalis:
Thanks again @dstansby

PR #3231, opened 2019-06-18
Remove coordinate test warnings

dstansby:
These are just warnings because the time is very in the future.


nabobalis:
Thanks again @dstansby

dstansby:
slowly I will beat the warnings

nabobalis:
You know you can tackle more interesting problems!

dstansby:
but these are easy wins 😛

Cadair:
These warnings are the bane of my life in the DKIST package (where I use dates generated like 40 years into the future).

PR #3237, opened 2019-06-21
More warning removal

dstansby:


sunpy/net/fido_factory.py
```diff
@@ -8,7 +8,7 @@
 `Fido.fetch <sunpy.net.fido_factory.UnifiedDownloaderFactory.fetch>`.
 
 """
-from collections import Sequence
```
Cadair: @nabobalis Should should backport these ones for 3.8 compatibility?

nabobalis: Yes.

setup.cfg
```diff
@@ -83,6 +83,8 @@ addopts = --doctest-rst -m "not figure"
 markers =
     online: marks this test function as needing online connectivity.
     figure: marks this test function as using hash-based Matplotlib figure verification. This mark is not meant to be directly applied, but is instead automatically applied when a test function uses the @sunpy.tests.helpers.figure_test decorator.
+    flaky
+    array_compare
```
nabobalis: What do these do?

dstansby: pytest raises a warning now if any markers used aren't in this list (presumably so users have a definitive list of markers?)

nabobalis: Could you add some descriptions like the other two?

Cadair: Both of these are really fails on the plugin to not register their marks.

nabobalis:
Thanks @dstansby

PR #3240, opened 2019-06-22
Fix: Util function to ensure list of dicts all have same keys for hek

nabobalis:
Fixes #3238 via a large HACK. 

Not really sure this is a sane way to do it however.

PR #3245, opened 2019-06-25
Fix the handling of diffrot warping

### LINKED ISSUE #3226: Differential rotation seems to rotate the wrong way
### Description
This was previously raised in #3190, but I'm breaking this into a separate issue.

When warping a map using the new differential rotation code, the image is warped in the opposite direction, e.g. when rotating a map forward two days in the future, the new rotated map appears rotated "backwards in time." 

See example below, where Map A (top panel) is being rotated to the observer location of Map B (middle panel). Because Map B is in the future (by two days),  it is expected that Map A will be warped in the direction of the solar rotation. However, the opposite happens.

### Expected behavior
image

### Actual behavior
fig_diff_rot_1600

### Steps to Reproduce

The code to reproduce this example can be found [here](https://github.com/sunpy/sunpy-1.0-paper/blob/master/figures/diff_rot_example_plot.py).

### System Details
- SunPy Version: 1.0.1
- Python Version: 3.6
- OS information: macOS

Cadair:
TODO:

- [x] Fix existing tests
- [x] Add an array comparison test for diff rot warping.
- [ ] ~Add a test comparing pixel coordinates for points (to ensure the to/from pixel coords is consistent)~

Fixes #3226

Figure_1


wafels:
The warp transformation requires us to go from pixels as seen from the transformed data back to the original data.  If I understand this Pr correctly then the answer to #3226 is to (1) subtract the differential rotation (from the transformed pixels back to the original, L417) and (2) implement the transformation from the HGS frame at the new time to the HGS frame at the original time, L410.  Step (2) is required because the HGS frames at different times are different because the Earth moves.  So there has to be an explicit transformation to take care of this.

wafels:
There are some CI errors I don't recognize.  I also expect that there will be some explicit test errors.  Also, the `test_differential_rotation` code should be tightened up; for example, the tests on the array sizes should really be explicit.

nabobalis:
Yeah going to have to be careful how to read the CI failures. Most are fixed in #3240 but it isn't ready to be merged.

ayshih:
The code changes look good, but of course the real test failures should fixed.

nabobalis:
Rebased (due to the CI being fixed hopefully now) and fixed the one failing test.

Cadair:
@wafels I have added an array comparison test which will make sure that our output image is constant. Do you have any suggestions for other tests we can run?

PR #3247, opened 2019-06-26
Fixing transformations to handle a change in observation time and to handle velocities

ayshih:
- The following transformations can now handle a change in observation time (which means all of ours now can)
  - HGS<->HGC
  - HGS<->HCC
  - HCC<->HPC
- The above transformations are now `FunctionTransformWithFiniteDifference` rather than `FunctionTransform` to support velocities.  (Incidentally, the underlying implementations are now the combination of matrix transforms and affine transforms, so someone in the future could "easily" re-implement everything as `AffineTransform`, but that is not done here.)
- Fixed the handling of velocities so that they are handled properly by the HCRS<->HGS transformations, so that anyone already using velocities with Astropy frames won't encounter any issues when walking into/through a SunPy frame.  (We won't tell users broadly that velocities are supported until SunPy 1.1.)

### DIFF
--- a/docs/code_ref/coordinates.rst
+++ b/docs/code_ref/coordinates.rst
@@ -130,8 +130,8 @@ coordinates is::
    <SkyCoord (Helioprojective: obstime=2017-07-26T00:00:00.000, rsun=695700.0 km, observer=<HeliographicStonyhurst Coordinate for 'earth'>): (Tx, Ty) in arcsec
        (0., 0.)>
    >>> c.transform_to(frames.HeliographicCarrington)
-   <SkyCoord (HeliographicCarrington: obstime=2017-07-26T00:00:00.000): (lon, lat, radius) in (deg, deg, km)
-      (283.95274241, 5.31701821, 695700.00000125)>
+   <SkyCoord (HeliographicCarrington: obstime=2017-07-26T00:00:00.000): (lon, lat, radius) in (deg, deg, AU)
+      (283.95274241, 5.31701821, 0.00465047)>
 
 It is also possible to transform to any coordinate system implemented in Astropy. This can be used to find the position of the solar limb in AltAz equatorial coordinates::
 
--- a/sunpy/coordinates/transformations.py
+++ b/sunpy/coordinates/transformations.py
@@ -18,10 +18,11 @@ from copy import deepcopy
 import numpy as np
 
 import astropy.units as u
-from astropy.coordinates import ICRS, HCRS, ConvertError, BaseCoordinateFrame, get_body_barycentric
+from astropy.coordinates import (ICRS, HCRS, ConvertError, BaseCoordinateFrame,
+                                 get_body_barycentric, get_body_barycentric_posvel)
 from astropy.coordinates.baseframe import frame_transform_graph
 from astropy.coordinates.representation import (CartesianRepresentation, SphericalRepresentation,
-                                                UnitSphericalRepresentation)
+                                                UnitSphericalRepresentation, CartesianDifferential)
 from astropy.coordinates.transformations import (FunctionTransform, AffineTransform,
                                                  FunctionTransformWithFiniteDifference)
 from astropy.coordinates.matrix_utilities import rotation_matrix, matrix_transpose
@@ -321,10 +322,19 @@ def hcrs_to_hgs(hcrscoord, hgsframe):
         raise ValueError("To perform this transformation the coordinate"
                          " Frame needs an obstime Attribute")
 
+    # Check whether differentials are involved on either end
+    has_differentials = ((hcrscoord._data is not None and hcrscoord.data.differentials) or
+                         (hgsframe._data is not None and hgsframe.data.differentials))
+
     # Determine the Sun-Earth vector in ICRS
     # Since HCRS is ICRS with an origin shift, this is also the Sun-Earth vector in HCRS
-    sun_pos_icrs = get_body_barycentric('sun', hgsframe.obstime)
-    earth_pos_icrs = get_body_barycentric('earth', hgsframe.obstime)
+    # If differentials exist, also obtain Sun and Earth velocities
+    if has_differentials:
+        sun_pos_icrs, sun_vel = get_body_barycentric_posvel('sun', hgsframe.obstime)
+        earth_pos_icrs, earth_vel = get_body_barycentric_posvel('earth', hgsframe.obstime)
+    else:
+        sun_pos_icrs = get_body_barycentric('sun', hgsframe.obstime)
+        earth_pos_icrs = get_body_barycentric('earth', hgsframe.obstime)
     sun_earth = earth_pos_icrs - sun_pos_icrs
 
     # De-tilt the Sun-Earth vector to the frame with the Sun's rotation axis parallel to the Z axis
@@ -349,10 +359,15 @@ def hcrs_to_hgs(hcrscoord, hgsframe):
     if np.any(hcrscoord.obstime != hgsframe.obstime):
         sun_pos_old_icrs = get_body_barycentric('sun', hcrscoord.obstime)
         offset_icrf = sun_pos_icrs - sun_pos_old_icrs
-        offset = offset_icrf.transform(total_matrix)
     else:
-        offset = CartesianRepresentation(0, 0, 0)*u.m
+        offset_icrf = sun_pos_icrs * 0  # preserves obstime shape
 
+    # Add velocity if needed (at the HGS observation time)
+    if has_differentials:
+        vel_icrf = (sun_vel - earth_vel).represent_as(CartesianDifferential)
+        offset_icrf = offset_icrf.with_differentials(vel_icrf)
+
+    offset = offset_icrf.transform(total_matrix)
     return total_matrix, offset
 
 
@@ -366,7 +381,14 @@ def hgs_to_hcrs(hgscoord, hcrsframe):
 
     # Invert the transformation to get the HGS->HCRS transformation
     reverse_matrix = matrix_transpose(total_matrix)
-    reverse_offset = (-offset).transform(reverse_matrix)
+    # If differentials exist, properly negate the velocity
+    if offset.differentials:
+        pos = -offset.without_differentials()
+        vel = -offset.differentials['s']
+        offset = pos.with_differentials(vel)
+    else:
+        offset = -offset
+    reverse_offset = offset.transform(reverse_matrix)
 
     return reverse_matrix, reverse_offset
 
--- a/sunpy/coordinates/transformations.py
+++ b/sunpy/coordinates/transformations.py
@@ -18,8 +18,8 @@ from copy import deepcopy
 import numpy as np
 
 import astropy.units as u
-from astropy.coordinates import ICRS, HCRS, ConvertError, BaseCoordinateFrame, get_body_barycentric
-from astropy.coordinates.baseframe import frame_transform_graph
+from astropy.coordinates import ICRS, HCRS, ConvertError, SkyCoord, get_body_barycentric
+from astropy.coordinates.baseframe import BaseCoordinateFrame, frame_transform_graph
 from astropy.coordinates.representation import (CartesianRepresentation, SphericalRepresentation,
                                                 UnitSphericalRepresentation)
 from astropy.coordinates.transformations import (FunctionTransform, AffineTransform,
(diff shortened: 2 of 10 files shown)


sunpy/coordinates/transformations.py
```diff
@@ -76,108 +63,155 @@ def _observers_are_equal(obs_1, obs_2, string_ok=False):
...
+        raise ValueError("To perform this transformation the coordinate"
+                         " Frame needs an obstime Attribute")
+
+    # If obstime is None or the obstime matches, nothing needs to be done
+    if obstime is None or np.all(frame.obstime == obstime):
+        return frame
+
+    if isinstance(frame, BaseCoordinateFrame):
+        frame_class = frame.__class__
+    else:  # Assume that it is like SkyCoord and contains the frame
```
Cadair: Is this ever a `SkyCoord` here? I thought the transform graph always operated on frames.

sunpy/coordinates/transformations.py
```diff
@@ -76,108 +63,155 @@ def _observers_are_equal(obs_1, obs_2, string_ok=False):
...
+    if frame.obstime is None:
+        raise ValueError("To perform this transformation the coordinate"
+                         " Frame needs an obstime Attribute")
+
+    # If obstime is None or the obstime matches, nothing needs to be done
+    if obstime is None or np.all(frame.obstime == obstime):
+        return frame
+
+    if isinstance(frame, BaseCoordinateFrame):
+        frame_class = frame.__class__
```
Cadair: I generally prefer `frame_class = type(frame)` but makes little difference.

sunpy/coordinates/transformations.py
```diff
@@ -76,108 +63,155 @@ def _observers_are_equal(obs_1, obs_2, string_ok=False):
...
 def hgs_to_hgc(hgscoord, hgcframe):
     """
-    Transform from Heliographic Stonyhurst to Heliograpic Carrington.
+    Convert from Heliographic Stonyhurst to Heliographic Carrington.
     """
-    if hgcframe.obstime is None or np.any(hgcframe.obstime != hgscoord.obstime):
-        raise ValueError("Can not transform from Heliographic Stonyhurst to "
-                         "Heliographic Carrington, unless both frames have matching obstime.")
+    # First transform the HGS coord to the HGC obstime
+    int_coord = _transform_obstime(hgscoord, hgcframe.obstime)
```
Cadair: I am not seeing much meaning in `int_coord` as a variable name?

sunpy/coordinates/transformations.py
```diff
@@ -76,108 +63,155 @@ def _observers_are_equal(obs_1, obs_2, string_ok=False):
...
-    z = helioccoord.z.to(u.m)
+    # Loopback transform HCC coord to obstime and observer of HPC frame
+    int_frame = Heliocentric(obstime=heliopframe.obstime, observer=heliopframe.observer)
+    int_coord = helioccoord.transform_to(int_frame)
 
-    # d is calculated as the distance between the points
-    # (x,y,z) and (0,0,D0).
-    distance = np.sqrt(x**2 + y**2 + (helioccoord.observer.radius - z)**2)
+    # Shift the origin from the Sun to the observer
+    distance = SkyCoord(int_coord.observer).heliographic_stonyhurst.radius
```
Cadair: I don't think we want to incur the performance hit of the `SkyCoord` initializer here, would be better to do:
```suggestion
    distance = int_coord.observer.transform_to("heliographic_stonyhurst").radius
```

Although, because `int_coord` is always an instance of `Heliocentric` we actually know that `.observer` will always be a HGS frame, because the Descriptor for the frame always returns it in that frame irrespective of what was passed in.

sunpy/coordinates/transformations.py
```diff
@@ -76,108 +63,155 @@ def _observers_are_equal(obs_1, obs_2, string_ok=False):
...
 
-    cosx = np.cos(x)
-    sinx = np.sin(x)
-    cosy = np.cos(y)
-    siny = np.sin(y)
+    # Permute/swap axes from HPC equivalent Cartesian to HCC
+    newrepr = heliopcoord.cartesian.transform(matrix_transpose(_matrix_hcc_to_hpc()))
+
+    # Shift the origin from the observer to the Sun
+    distance = SkyCoord(heliocframe.observer).heliographic_stonyhurst.radius
```
Cadair: Same comment here about `SkyCoord`.

sunpy/coordinates/transformations.py
```diff
@@ -76,108 +63,155 @@ def _observers_are_equal(obs_1, obs_2, string_ok=False):
...
+                            [0, 1, 0]])
 
-    representation = CartesianRepresentation(
-        rx.to(u.km), ry.to(u.km), rz.to(u.km))
-    return heliocframe.realize_frame(representation)
+    # Rotate in latitude and longitude (sign difference because of direction difference)
+    lat_matrix = rotation_matrix(latitude, 'y')
+    lon_matrix = rotation_matrix(-longitude, 'z')
 
+    return matrix_product(matrix_product(lon_matrix, lat_matrix), axes_matrix)
```
Cadair: The future is now:
```suggestion
    return lon_matrix @ lat_matrix @ axes_matrix
```
(Also the docstring for `matrix_product` tells me it exists so you don't need to nest calls to it.)

sunpy/coordinates/transformations.py
```diff
@@ -187,71 +221,59 @@ def hcc_to_hgs(helioccoord, heliogframe):
...
                            "without `obstime` being specified.".format(helioccoord.observer))
 
-    x = helioccoord.x.to(u.m)
-    y = helioccoord.y.to(u.m)
-    z = helioccoord.z.to(u.m)
-
-    l0_rad = helioccoord.observer.lon
-    b0_deg = helioccoord.observer.lat
+    # Make sure the observer is in HGS
+    obs = helioccoord.observer.\
```
Cadair: I really hate line continuation characters, can't you line break after the `transform_to(`?

sunpy/coordinates/transformations.py
```diff
@@ -187,71 +221,59 @@ def hcc_to_hgs(helioccoord, heliogframe):
                            "heliographic coordinates for observer '{}' "
                            "without `obstime` being specified.".format(helioccoord.observer))
 
-    x = helioccoord.x.to(u.m)
-    y = helioccoord.y.to(u.m)
-    z = helioccoord.z.to(u.m)
-
-    l0_rad = helioccoord.observer.lon
-    b0_deg = helioccoord.observer.lat
+    # Make sure the observer is in HGS
```
Cadair:
```suggestion
    # Make sure the observer is in HGS at the correct obstime.
```

sunpy/coordinates/transformations.py
```diff
@@ -187,71 +221,59 @@ def hcc_to_hgs(helioccoord, heliogframe):
...
-    hglt = np.arcsin((y * cosb + z * sinb) / hecr)
+    # Transform from HCC to HGS at the HCC obstime
+    newrepr = helioccoord.cartesian.transform(total_matrix)
+    int_coord = HeliographicStonyhurst(newrepr, obstime=helioccoord.obstime)
 
-    representation = SphericalRepresentation(
-        np.rad2deg(hgln), np.rad2deg(hglt), hecr.to(u.km))
-    return heliogframe.realize_frame(representation)
+    # Loopback transform HGS if there is a change in obstime
+    if np.any(helioccoord.obstime != heliogframe.obstime):
```
Cadair: Isn't this if statement done in `_transform_obstime`?

sunpy/coordinates/transformations.py
```diff
@@ -187,71 +221,59 @@ def hcc_to_hgs(helioccoord, heliogframe):
...
                            "heliocentric coordinates for observer '{}' "
                            "without `obstime` being specified.".format(heliocframe.observer))
 
-    l0_rad = heliocframe.observer.lon.to(u.rad)
-    b0_deg = heliocframe.observer.lat
-
-    lon = np.deg2rad(hglon)
-    lat = np.deg2rad(hglat)
+    # Make sure the observer is in HGS
+    obs = heliocframe.observer.\
```
Cadair: *twitch*

sunpy/coordinates/transformations.py
```diff
@@ -187,71 +221,59 @@ def hcc_to_hgs(helioccoord, heliogframe):
...
-    cosb = np.cos(b0_deg.to(u.rad))
-    sinb = np.sin(b0_deg.to(u.rad))
-
-    lon = lon - l0_rad
-
-    cosx = np.cos(lon)
-    sinx = np.sin(lon)
-    cosy = np.cos(lat)
-    siny = np.sin(lat)
+    # If HGS coord is only directional, place it on the surface of the Sun
```
Cadair: I don't think this is the place we should be doing this, we should be using `calculate_distance` (renamed) to do this.

sunpy/coordinates/transformations.py
```diff
@@ -187,71 +221,59 @@ def hcc_to_hgs(helioccoord, heliogframe):
...
+    r = heliogcoord.spherical.distance
+    if r.unit is u.one and u.allclose(r, 1*u.one):
+        int_repr = heliogcoord.spherical * RSUN_METERS
```
Cadair: We shoudn't be using `RSUN_METERS` here we should be using the `rsun` property of the `hcc` coord.

Cadair:
Given how large a shift this is in the way we implement these frames, I wonder if it would be better milestoned to 1.1? @ayshih Do you think it's worth splitting this into just the obstime fixes and the other things?

docs/code_ref/coordinates.rst
```diff
@@ -130,8 +130,8 @@ coordinates is::
    <SkyCoord (Helioprojective: obstime=2017-07-26T00:00:00.000, rsun=695700.0 km, observer=<HeliographicStonyhurst Coordinate for 'earth'>): (Tx, Ty) in arcsec
        (0., 0.)>
    >>> c.transform_to(frames.HeliographicCarrington)
-   <SkyCoord (HeliographicCarrington: obstime=2017-07-26T00:00:00.000): (lon, lat, radius) in (deg, deg, km)
-      (283.95274241, 5.31701821, 695700.00000125)>
+   <SkyCoord (HeliographicCarrington: obstime=2017-07-26T00:00:00.000): (lon, lat, radius) in (deg, deg, AU)
```
Cadair: This feels like a change we probably shouldn't do in a bugfix? (changing the default unit?)

ayshih: I'm a bit mystified here.  I made this change (and others like it) so that the doc tests would pass, but when I run the same code locally, I still get km, not AU.  I don't know where AU is coming from.

sunpy/coordinates/transformations.py
ayshih: I can go with that

sunpy/coordinates/transformations.py
ayshih: Heh, this is a holdover because I was testing this function on both frames and SkyCoords.  I believe you're correct that the latter will never be needed by the machinery.  Removing

sunpy/coordinates/transformations.py
ayshih: It's short for "intermediate coordinate".  I use that and `int_frame` quite a bit in my new-frames PR (although that use may go down now that I came up with the `_transform_obstime()` approach).

sunpy/coordinates/transformations.py
ayshih: Using `transform_to("heliographic_stonyhurst")` would actually be a bit problematic in some cases, but if it's guaranteed to be HGS, then I'll just take it out.

sunpy/coordinates/transformations.py
ayshih: Really?  Sweet!

sunpy/coordinates/transformations.py
ayshih: Moot

sunpy/coordinates/transformations.py
ayshih: Moot (since HGS is guaranteed)

sunpy/coordinates/transformations.py
ayshih: Moot

sunpy/coordinates/transformations.py
ayshih: Yup

sunpy/coordinates/transformations.py
ayshih: I'm going to remove this code because the HGS docstring says that it will always be 3D.  (In actuality, it can be only directional in some circumstances, but I'll consider that a bug to be fixed.)

ayshih:
> What is the motivation for changing to `FunctionTransformWithFiniteDifference` if we aren't doing anything with them?

A fair question.  Really, the best reason is to not have any `FunctionTransform` just like Astropy doesn't.  This enables undocumented and totally unverified support for velocities, which is both good and bad.  I could be convinced to switch everything back to `FunctionTransform` until I actually start working on supporting velocities.

> Also, what would be the advantage of moving to `AffineTransform` at some point in the future?

It should slightly improve performance, and it should make transforming velocities a bit cleaner.

> Given how large a shift this is in the way we implement these frames, I wonder if it would be better milestoned to 1.1? @ayshih Do you think it's worth splitting this into just the obstime fixes and the other things?

I think it'd be a bit awkward and inelegant to insert the changed-obstime behavior in all of the right places without overhauling the implementation at the same time.  I understand that it feels bigger than a bugfix, but the user-facing side is essentially unchanged, so it also feels too small to be a feature to be put off until 1.1.  I vote for 1.0.3.

sunpy/coordinates/transformations.py
Cadair: Isn't that what `.heliographic_stonyhurst` is doing underneath anyway?

Cadair:
I think I am happy to put my trust in our test suite and keep this milestoned 1.0.3 as it does facilitate a fix to a really annoying bug.

I would like to understand the change in units in the doctests though.

No idea what's going on with the tests :/

docs/code_ref/coordinates.rst
ayshih: Okay, I don't know why I was getting weird answers locally.  (Maybe I was running in an old shell.)  My local results currently match the CI results.

Are you of the mind that I should force everything to "km"?  It wasn't actually a default defined by the frame, but rather the unit was hard-coded as the output of transformations.  It seems a bit unseemly to force the unit if the sole reason is to exactly match the previous output.

Cadair: Ok, I think that makes sense. The idea of units after all is that it doesn't matter what unit it is in, and AU is much nicer than km anyway.

sunpy/physics/solar_rotation.py
```diff
@@ -64,6 +64,10 @@ def calculate_solar_rotate_shift(mc, layer_index=0, **kwargs):
 
     # Calculate the rotations and the shifts
     for i, m in enumerate(mc):
+        # Skip the reference layer
```
Cadair: how did this end up here :grinning:

ayshih: I rebased this PR to include #3255 because the tweaking of transformations in this PR caused even more not-exactly-zero numerical-precision bugs that resulted in test failures.  At the time, all tests were green.  (At the time of this writing, tests are not green for an unrelated reason.)

ayshih: As an aside, I intended to leave #3255 for someone else to do, but I was "forced" to do it myself so that the tests would pass here.  Apparently if you want to squash some pesky bug in SunPy, just find some way for it to impact coordinates...

changelog/3247.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed all transformations to properly handle a change in observation time.
```
Cadair: can you add a second changelog entry for the velocities stuff.

Cadair:
@ayshih I merged #3255 so can you rebase this as well :smile:

changelog/3247.bugfix.rst
ayshih: Done

ayshih:
Done

PR #3267, opened 2019-07-10
Make rotate respect Dask arrays

### LINKED ISSUE #3266: rotate does not support Map data stored as a Dask array
Maps can be created using Dask arrays (as of #2689) such that the metadata is stored in memory while the underlying image data is loaded "lazily" and only loaded when necessary. However, when `.rotate()` is called, the resulting Map data is a normal Numpy array.

This is because some operations in `.rotate()` force the array data into memory instead of leaving it as a lazily loaded array. There are three culprits here:

* `np.pad` (https://github.com/sunpy/sunpy/blob/master/sunpy/map/mapbase.py#L1082)
* `np.nan_to_num` (https://github.com/sunpy/sunpy/blob/master/sunpy/image/transform.py#L109)
* `skimage.transform.warp` (https://github.com/sunpy/sunpy/blob/master/sunpy/image/transform.py#L129)

The first two are easily solved as Dask already provides lazy versions of these functions (see https://docs.dask.org/en/latest/array-api.html). The last one is a bit more tricky as their is no drop-in replacement for a lazy affine transform. However, its fairly easy to get around this by wrapping it in a `dask.delayed` call (https://docs.dask.org/en/latest/array-creation.html#using-dask-delayed).

Whether this is a bug report or a feature request is probably a matter of opinion. My personal feeling is that this is a bug. Because Map supports storing the underlying image data as a Dask array, the methods on Map should return a Dask array if the original underlying data is a Dask array.

I'm happy to make a PR with the proposed changes, but wanted to get other's opinions first. As this would involve making changes to check if the array data is a specific data structure from another library that is not even a hard dependency of sunpy, we should probably proceed cautiously.

wtbarnes:
Fixes #3266 

This PR makes a few minor changes in `Map.rotate()` to allow a Dask array to pass through without being brought into memory. This means that `.rotate()` returns a lazily evaluated Dask array rather than a NumPy array if the underlying data in Map is represented as a Dask array.

### DIFF
--- a/sunpy/image/transform.py
+++ b/sunpy/image/transform.py
@@ -10,11 +10,20 @@ from sunpy.util.exceptions import SunpyUserWarning
 
 try:
     import skimage.transform
-    scikit_image_not_found = False
 except ImportError:
     warnings.warn("scikit-image could not be imported. Image rotation will use scipy",
                   ImportWarning)
-    scikit_image_not_found = True
+    SCIKIT_IMAGE_NOT_FOUND = True
+else:
+    SCIKIT_IMAGE_NOT_FOUND = False
+
+
+try:
+    import dask
+except ImportError:
+    DASK_FOUND = False
+else:
+    DASK_FOUND = True
 
 
 __all__ = ['affine_transform']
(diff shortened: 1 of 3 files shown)


wafels:
Tests?

wtbarnes:
There's another bug here in that the "laziness" of the Dask array is not actually preserved, i.e. it is brought into memory at some point,likely by a NumPy operation.

wtbarnes:
It might also be worth replacing the `dask.delayed` messiness in `affine_transform` with something from [dask-image](https://github.com/dask/dask-image) though the scipy affine_transform is not yet implemented.

wtbarnes:
I'm going close this as I do not think this is the right approach here. More generally, we need an affine transform that is "Dask"-y, but that is tricky and larger than the scope of sunpy.

I'll probably create another issue that catalogues more generally the inconsistencies in sunpy with respect to Dask.

PR #3271, opened 2019-07-17
Improved the layout of the frame transform graph

ayshih:
For improved readability, here's a new layout for the graph showing the coordinate frames and their transformations:
graphLR


nabobalis:
Thanks @ayshih

PR #3287, opened 2019-07-31
Change treatment of non-finite entries before image coalignment

### LINKED ISSUE #3277: Do not try to remove non-finite entries before image coalignment
Currently, before coaligning two images, [`repair_image_nonfinite`](https://github.com/sunpy/sunpy/blob/master/sunpy/image/coalignment.py#L290) is run on the image to remove non-finite entries and replace them with the local mean.

This step should be removed and the user should be expected to this themselves before coaligning images. If NaNs/non-finite entries are present, a warning should be thrown.

xuanxu:
### Description

This PR removes the step of reparing images (replacing non-finite entries with local mean) before coaligning them and throws a warning if any of the images has NaNs/non-finite entries.

Fixes #3277

### DIFF
--- a/sunpy/image/coalignment.py
+++ b/sunpy/image/coalignment.py
@@ -423,7 +423,8 @@ def apply_shifts(mc, yshift: u.pix, xshift: u.pix, clip=True, **kwargs):
 
 
 def calculate_match_template_shift(mc, template=None, layer_index=0,
-                                   func=_default_fmap_function):
+                                   func=_default_fmap_function,
+                                   repair_nonfinite=True):
     """
     Calculate the arcsecond shifts necessary to co-register the layers in a
     `~sunpy.map.MapSequence` according to a template taken from that
@@ -493,7 +494,8 @@ def calculate_match_template_shift(mc, template=None, layer_index=0,
         this_layer = func(m.data)
 
         # Calculate the y and x shifts in pixels
-        yshift, xshift = calculate_shift(this_layer, tplate)
+        yshift, xshift = calculate_shift(this_layer, tplate,
+            repair_nonfinite=repair_nonfinite)
 
         # Keep shifts in pixels
         yshift_keep[i] = yshift
@@ -515,7 +517,7 @@ def calculate_match_template_shift(mc, template=None, layer_index=0,
 # Coalignment by matching a template
 def mapsequence_coalign_by_match_template(mc, template=None, layer_index=0,
                                           func=_default_fmap_function, clip=True,
-                                          shift=None, **kwargs):
+                                          shift=None, repair_nonfinite=True, **kwargs):
     """
     Co-register the layers in a `~sunpy.map.MapSequence` according to a
     template taken from that `~sunpy.map.MapSequence`. This method REQUIRES
@@ -594,7 +596,8 @@ def mapsequence_coalign_by_match_template(mc, template=None, layer_index=0,
     if shift is None:
         shifts = calculate_match_template_shift(mc, template=template,
                                                 layer_index=layer_index,
-                                                func=func)
+                                                func=func,
+                                                repair_nonfinite=repair_nonfinite)
         xshift_arcseconds = shifts['x']
         yshift_arcseconds = shifts['y']
     else:
--- a/sunpy/image/coalignment.py
+++ b/sunpy/image/coalignment.py
@@ -299,14 +299,10 @@ def check_for_nonfinite_entries(layer_image, template_image):
     template_image : `numpy.ndarray`
         A two-dimensional `numpy.ndarray`.
     """
-
-    bad_index_layer = np.where(np.logical_not(np.isfinite(layer_image)))
-    bad_index_template = np.where(np.logical_not(np.isfinite(template_image)))
-
-    if bad_index_layer[0].size != 0:
+    if not np.all(np.isfinite(layer_image)):
         warnings.warn("The layer image has noninfinite entries.", SunpyUserWarning)
 
-    if bad_index_template[0].size != 0:
+    if not np.all(np.isfinite(template_image)):
         warnings.warn("The template image has noninfinite entries.", SunpyUserWarning)
 
 
--- a/sunpy/image/coalignment.py
+++ b/sunpy/image/coalignment.py
@@ -27,13 +27,13 @@ import astropy.units as u
 
 import sunpy.map
 from sunpy.map.mapbase import GenericMap
-from sunpy.util import SunpyUserWarning
+from sunpy.util import (SunpyUserWarning, SunpyDeprecationWarning)
 
 __all__ = ['calculate_shift', 'clip_edges', 'calculate_clipping',
            'match_template_to_layer', 'find_best_match_location',
            'get_correlation_shifts', 'parabolic_turning_point',
-           'check_for_nonfinite_entries', 'apply_shifts',
-           'mapsequence_coalign_by_match_template',
+           'repair_image_nonfinite', 'check_for_nonfinite_entries',
+           'apply_shifts', 'mapsequence_coalign_by_match_template',
            'calculate_match_template_shift']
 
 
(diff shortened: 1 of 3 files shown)


sunpy/image/coalignment.py
```diff
@@ -287,53 +288,26 @@ def parabolic_turning_point(y):
...
-        # ignoring nans
-        subarray = repaired_image[y - 1: y + 2, x - 1: x + 2]
-        repaired_image[by, bx] = np.mean(subarray[np.isfinite(subarray)])
-        bad_index = np.where(np.logical_not(np.isfinite(repaired_image)))
-    return repaired_image
+
+    bad_index_layer = np.where(np.logical_not(np.isfinite(layer_image)))
+    bad_index_template = np.where(np.logical_not(np.isfinite(template_image)))
+
+    if bad_index_layer[0].size != 0:
```
ayshih:
```suggestion
    if not np.all(np.isfinite(layer_image):
```
Since you don't actually use the output from `np.where()`, this is a simpler/cleaner pattern to use, and you can remove the `bad_index_layer = ...` line above.

sunpy/image/coalignment.py
```diff
@@ -287,53 +288,26 @@ def parabolic_turning_point(y):
...
+        warnings.warn("The layer image has noninfinite entries.", SunpyUserWarning)
+
+    if bad_index_template[0].size != 0:
```
ayshih: See above comment

ayshih:
You've changed the name and functionality of a function in the public API, but we actually need to go through a deprecation period.  It'd be better to create a new function and deprecate the old one, and then the old one can be deleted later down the line.

sunpy/image/coalignment.py
xuanxu: nice!

sunpy/image/coalignment.py
xuanxu: done!

sunpy/image/coalignment.py
```diff
@@ -306,6 +325,7 @@ def repair_image_nonfinite(image):
         the next non-finite value is replaced by the mean of its finite
         valued nearest neighbors.
     """
+    warnings.warn('This function has been deprecated.', SunpyDeprecationWarning)
```
ayshih: You should instead use the `deprecated` decorator in `sunpy.util.decorators` to mark this function as deprecated.

xuanxu: done!

Cadair:
Am I right in thinking this is still a significant behaviour change by default? If so it should not be a bug fix.

I think the best thing to do would be to add a flag somewhere to turn on the new behaviour, which we can then flip the default on later and then remove.

ayshih:
> Am I right in thinking this is still a significant behaviour change by default? If so it should not be a bug fix.

I agree that this not merely a bug fix, so it should be milestoned instead for 1.1 rather than 1.0.3.

xuanxu:
> Am I right in thinking this is still a significant behaviour change by default? If so it should not be a bug fix.

The changelog entry type is `removal`

wtbarnes:
> Am I right in thinking this is still a significant behaviour change by default? If so it should not be a bug fix.
> 
> I think the best thing to do would be to add a flag somewhere to turn on the new behaviour, which we can then flip the default on later and then remove.

@cadair Do you mean something along the lines of this?
```python
def calculate_shift(this_layer, template, repair_nonfinite=True):
    ...
    # Current behaviour
    if repair_nonfinite:
        # Repair any NANs, Infs, etc in the layer and the template
        this_layer = repair_image_nonfinite(this_layer)
        template = repair_image_nonfinite(template)
    # New behaviour
    else:
        # Warn user if any NANs, Infs, etc are present in the layer or the template
        check_for_nonfinite_entries(this_layer, template)
    ...
```

So would this be deprecated in 1.1, default behaviour changed in 1.2 (i.e. `repair_nonfinite` defaults to False), and then `repair_image_nonfinite` dropped in 2.0?

Cadair:
Oh isn't this a fun test of our new deprecation policy...

@wtbarnes yes, that code example is what I meant.

I *think* what we could do is add the kwarg and add a deprecation warning (for the behaviour change and kwarg removal) now, leave that in place until 2.0, then change the behaviour for 2.1 and remove it in 3.0.

This is because our deprecation policy is to provide a deprecation warning for at least one LTS release (for those people who stick on LTSs).

wtbarnes:
@xuanxu are you still interested in working on this? If so, could you make the change described in my previous comment?

sunpy/image/coalignment.py
```diff
@@ -64,9 +66,8 @@ def calculate_shift(this_layer, template):
         Pixel shifts ``(yshift, xshift)`` relative to the offset of the template
         to the input array.
     """
-    # Repair any NANs, Infs, etc in the layer and the template
```
wtbarnes: Per the deprecation policy, we need to preserve this as the default behavior for now. See my comment on the main thread for one possible solution.

sunpy/image/coalignment.py
```diff
@@ -287,6 +288,25 @@ def parabolic_turning_point(y):
...
+        A two-dimensional `numpy.ndarray`.
+    """
+    if not np.all(np.isfinite(layer_image)):
+        warnings.warn('The layer image has noninfinite entries.', SunpyUserWarning)
+
+    if not np.all(np.isfinite(template_image)):
+        warnings.warn('The template image has noninfinite entries.', SunpyUserWarning)
+
+
+@deprecated("1.1")
```
wtbarnes: I think we will have to wait til 2.1 to deprecate.

Cadair: We can deprecate now, just wont remove till 2.1

xuanxu:
> @xuanxu are you still interested in working on this? If so, could you make the change described in my previous comment?

@wtbarnes Sure, I have some time now, I'm going for it.

xuanxu:
@wtbarnes I've push the changes. To keep the current behaviour while allowing to use the new one, I had to propagate the kwarg to a couple of functions calling `calculate_shift` internally. Let me know what you think.

sunpy/image/coalignment.py
wtbarnes: Ah ok. Clearly I'm still a bit confused about what the policy is 😄

wtbarnes:
Looks like the test failures are all online tests and not related to this PR.

sunpy/image/coalignment.py
```diff
@@ -64,9 +66,16 @@ def calculate_shift(this_layer, template):
         Pixel shifts ``(yshift, xshift)`` relative to the offset of the template
         to the input array.
     """
-    # Repair any NANs, Infs, etc in the layer and the template
-    this_layer = repair_image_nonfinite(this_layer)
-    template = repair_image_nonfinite(template)
+    if repair_nonfinite:
+        # Repair any NANs, Infs, etc in the layer and the template
+        # This behaviour is deprecated
+        warnings.warn('The value True for the repair_nonfinite kwarg is deprecated',
```
Cadair: Is the plan to leave this kwarg in and just default it to `False` post 2.0, or remove it altogether.

From the wording of this warning it sounds like the former, but then you are deprecating the repair function, so I am a little confused.

sunpy/image/coalignment.py
```diff
@@ -287,6 +296,25 @@ def parabolic_turning_point(y):
...
+
+    Parameters
+    ----------
+    layer_image : `numpy.ndarray`
+        A two-dimensional `numpy.ndarray`.
+    template_image : `numpy.ndarray`
+        A two-dimensional `numpy.ndarray`.
+    """
+    if not np.all(np.isfinite(layer_image)):
+        warnings.warn('The layer image has noninfinite entries.', SunpyUserWarning)
```
Cadair: I would like these warnings to be a little clearer, specifically I think it should answer the following two questions: What happens if I ignore this warning? What should I do to make it not occur?

sunpy/image/coalignment.py
wtbarnes: So I'm still fuzzy on the policy here, but I think your comment from above makes the most sense:

> I think what we could do is add the kwarg and add a deprecation warning (for the behaviour change and kwarg removal) now, leave that in place until 2.0, then change the behaviour for 2.1 and remove it in 3.0.

So `repair_nonfinite` will default to `True` for now, will default to `False` starting at 2.1, and then be removed, along with the `repair_image_nonfinite` function, in 3.0

Cadair: OK, makes sense. I feel this message should be changed to make it clear the behavior of repairing is deprecated not just the True value.

sunpy/image/coalignment.py
xuanxu: :+1: Done!

sunpy/image/coalignment.py
xuanxu: Done!

PR #3290, opened 2019-08-02
Reverted the return arguments for ImageAnimatorWCS example

yashrsharma44:
This PR fixes the return arguments of `ImageAnimatorWCS` example in `examples/plotting/imageanimatorwcs.py`.


Cadair:
(This will make the astropydev build fail but all others should pass)

PR #3291, opened 2019-08-05
Fixed observer-related bugs

ayshih:
This PR fixes three observer-related bugs in `coordinates`:

- Our frames default to an Earth observer.  Previously, if the user specifies `obstime` but not `observer`, this default observer is resolved to a full coordinate, but was still considered a "default" value, which meant that `observer` can get overwritten with something different by the SkyCoord machinery.  That is, `Heliocentric(obstime='2019-06-01')` behaves slightly differently from `Heliocentric(obstime='2019-06-01', observer='earth')`, even though they appear to be identical to the typical user.  This PR considers `observer` to no longer be a "default" value if `obstime` is specified.
- The HPC->HCC transformation had a bug if the observer changed between the two frames.
- ~The transformation from HPC to HGS (or beyond) goes through HCC, but note that HGS (or beyond) does not have the `observer` attribute.  The SkyCoord machinery creates an intermediate HCC frame that combines the `obstime` of the destination frame with the `observer` of the source (HPC) frame, and they may not be at the same time.  This PR forces the intermediate frame to have the same `obstime` as `observer`.  Accounting for a change in `obstime` in the overall transformation is thus pushed off to a subsequent transformation step.~  The HCC<->HGS transformations (as well as the HCC loopback) failed to handle properly when the frame's `obstime` wasn't the same as the `obstime` of the observer.  This unusual circumstance could happen "under the hood" when transforming between HPC and HGS (or beyond) while also changing the observation time.


sunpy/coordinates/transformations.py
```diff
@@ -180,12 +180,17 @@ def hpc_to_hcc(heliopcoord, heliocframe):
...
+    distance = heliopcoord.observer.radius
     newrepr += CartesianRepresentation(0*u.m, 0*u.m, distance)
 
     # Complete the conversion of HPC to HCC at the obstime and observer of the HPC coord
     int_coord = Heliocentric(newrepr, obstime=heliopcoord.obstime, observer=heliopcoord.observer)
 
+    # When a SkyCoord is transformed through HCC, the obstime and observer's obstime may not match
+    # because they may come from different frames.  The fix is to force the HCC obstime to match
+    # the observer's obstime, and let a later transformation step update the obstime.
+    heliocframe = heliocframe.replicate_without_data(obstime=heliocframe.observer.obstime)
```
Cadair: Should this be `heliocframe.obstime`? Why is it pulling from `observer`?

nabobalis:
IT FINALLY BROKE HERE!

PR #3299, opened 2019-08-09
Removed Travis config and fixed some minor CI issues.

nabobalis:
Astropy Dev failures are real but hopefully will be fixed upstream.

A failing test (`test_velocity_hcrs_hgs`) that happens under astropy 3.1.2 has been skipped.

Conda has to be pinned to astropy 3.1 as the asdf version is older than the current release and it tries to import pytest_plugins from astropy which is gone now from astropy 3.2.


sunpy/coordinates/tests/test_transformations.py
```diff
@@ -388,6 +389,7 @@ def test_hgs_hcc_sunspice():
     assert_quantity_allclose(new.z, 688539.32*u.km, atol=1e-2*u.km)
 
 
+@pytest.mark.skipif(astropy.__version__ < '3.2.0', reason="Not supported by Astropy <3.2")
```
ayshih: In principle, you should use `pkg_resources.parse_version()` (or similar) for proper version comparison, although of course it's incredibly unlikely that this conditional will still be present in our code when there's Astropy 10.0 or if there's ever Astropy 3.10.

nabobalis: Yeah I should.

nabobalis:
Oh some of the online tests are down, looks like we are getting less results than we have been. I assume a server is down.

examples/plotting/imageanimatorwcs.py
```diff
@@ -77,7 +77,12 @@
 # of ``wcs_anim`` allows us to set various properties.
 # We assign them to a easy to remember variable name.
 # Note the order of assignment out from ``wcs_anim.axes.coords``.
-time, solar_y, solar_x = wcs_anim.axes.coords
+
```
Cadair: We shouldn't be doing this in the example gallery, we should let astropydev fail.

Cadair:
The title says you are removing travis but you actually aren't?

examples/plotting/imageanimatorwcs.py
nabobalis: Fine.

nabobalis:
Well I did at the start. Then I thought I was being too hasty.

Cadair:
image

nabobalis:
Bar the servers all broken upstream, are we good here?

nabobalis:
THE SERVERS ALL BACK.

PR #3301, opened 2019-08-13
SUVI fido client

### LINKED ISSUE #3300: Add support to Fido to download SUVI data
### Description

The title says it all.

ehsteve:
### Description

Adds the ability to search for and download SUVI level 1b and 2 data.

Fixes #3300


sunpy/net/dataretriever/sources/goes.py
```diff
@@ -17,7 +19,32 @@
...
 TIME_FORMAT = config.get("general", "time_format")
 
-__all__ = ['XRSClient']
+__all__ = ["XRSClient", "SUVIClient"]
+
+import urllib.request
+import requests
+from bs4 import BeautifulSoup
+
+def url_is_alive(url):
```
Cadair: I am pretty sure we already have a function to do this in `util.net`

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +202,111 @@ def _can_handle_query(cls, *query):
...
+            wavelength = kwargs["wavelength"]
+        wavelength = wavelength.to(u.Angstrom, equivalencies=u.spectral())
+
+        if "level" in kwargs:
+            level = kwargs["level"]
+        else:
+            level = "l2"
+
+        if level == "l2":
+            base_url += f'l2/data/suvi-{level}-ci{int(wavelength.to("Angstrom").value):03}/'
```
Cadair: If this data source has a predictable URL pattern could you not use the [Scraper](http://docs.sunpy.org/en/stable/api/sunpy.util.scraper.Scraper.html#sunpy.util.scraper.Scraper)?

sunpy/net/dataretriever/sources/goes.py
nabobalis: On a large side note: We should use that to create a test that reports what servers are down or up that we access for our tests and examples.

Cadair: :thinking: nice idea.

nabobalis: Opened an issue: https://github.com/sunpy/sunpy/issues/3303

sunpy/net/dataretriever/sources/goes.py
ehsteve: I wasn't sure. The filenames depend on when the image was taken and so cannot be predicted. I request all of the filenames and then compare to the time range.

sunpy/net/dataretriever/sources/goes.py
ehsteve: I'm pretty sure we do. I was just trying to work fast here. I'll get rid of this function which is currently not even used anyway.

ehsteve:
@Cadair I could use some help on what class functions I need to implement to make this work with `Fido`...<cough> I mean `dataretriever`.

sunpy/net/dataretriever/sources/goes.py
Cadair: That's exactly what Scraper does iirc. You give it a filename pattern and it works out which file names exist within your time.

Cadair:
I think you have the main two you need, you also want a [`_makeimap`](https://github.com/sunpy/sunpy/blob/master/sunpy/net/dataretriever/sources/norh.py#L111).

Also you might want to look at [this](https://github.com/sunpy/sunpy/blob/master/sunpy/net/dataretriever/sources/norh.py#L121) as it's a better pattern than the one you have copied in this PR.

sunpy/net/dataretriever/sources/goes.py
hayesla: @Cadair  I was having similar trouble with `Scraper` - for the SUVI files the filenames aren't predictable and have two dates within each filename e.g.`dr_suvi-l2-ci171_g16_s20190302T000000Z_e20190302T000400Z_v1-0-0.fits` - how can you go about this using Scraper?

sunpy/util/scraper.py
```diff
@@ -116,7 +116,7 @@ def _URL_followsPattern(self, url):
             pattern = pattern.replace(k, v)
         matches = re.match(pattern, url)
         if matches:
-            return matches.end() == matches.endpos == len(self.now)
+            return matches.end() == matches.endpos
```
Cadair: This is what it was?! :laughing:

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +178,95 @@ def _can_handle_query(cls, *query):
...
+
+        if "level" in kwargs.keys():
+            level = str(kwargs["level"])
+        else:
+            level = "2"
+
+        if str(level) == "2":
+            search_pattern = base_url + 'l{level}/data/suvi-l{level}-ci{wave:03}/%Y/%m/%d/dr_suvi-l{level}-ci{wave:03}_g16_s%Y%m%dT%H%M%SZ_.*\.fits'
+        elif str(level) == "1b":
+            if wavelength in u.Quantity([131, 171, 195, 284], 'Angstrom'):
```
Cadair: I feel like this set of if statements is repeating itself a lot, as long as I am parsing this correctly and the only difference is the prefix before the wave, I would recommend you make a dict of prefixes so you only need the template once.

```python
search_pattern = base_url + 'l{level}/suvi-l{level}-fe{wave:03}/%Y/%m/%d/OR_SUVI-L{level}-{prefix}{wave:03}_G16_s%Y%j%H%M%S.*\.fits.gz'

...
scraper = Scraper(search_pattern, level=level, wave=wave,
                              goes_number=satellitenumber, prefix=prefixes[wave])
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +178,95 @@ def _can_handle_query(cls, *query):
...
+        query : list of query objects
+
+        Returns
+        -------
+        boolean
+            answer as to whether client can service the query
+        """
+        chk_var = 0
+        for x in query:
+            if x.__class__.__name__ == 'Instrument' and x.value.lower() == 'suvi':
```
Cadair: Can you take a look at [this](https://github.com/sunpy/sunpy/blob/master/sunpy/net/dataretriever/sources/norh.py#L121) and maybe re-write this, I am not sure we should be perpetuating this pattern, it's pretty hard to read.

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +178,95 @@ def _can_handle_query(cls, *query):
...
+        level : str
+            The level of the data.
+
+        Note
+        ----
+        SUVI data does not exist for GOES satellite < 16.
+        """
+        base_url = "https://data.ngdc.noaa.gov/platforms/solar-space-observing-satellites/goes/goes{goes_number}/"
+
+        if "wavelength" not in kwargs.keys() or not kwargs["wavelength"]:
```
Cadair:
```suggestion
        if not kwargs.get("wavelength", None):
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +178,95 @@ def _can_handle_query(cls, *query):
...
+            raise ValueError("Queries must specify a wavelength.")
+        else:
+            wavelength = kwargs["wavelength"]
+        wavelength = wavelength.to(u.Angstrom, equivalencies=u.spectral())
+
+        if "satellitenumber" in kwargs.keys():
```
Cadair: This whole if block should just be:
```suggestion
satellitenumber = int(kwargs.get("satellitenumber", 16))
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +178,95 @@ def _can_handle_query(cls, *query):
...
+            satellitenumber = int(kwargs["satellitenumber"])
+        else:
+            satellitenumber = 16
+
+        if "level" in kwargs.keys():
```
Cadair: and this one should be:
```suggestion
level = str(kwargs.get("level", "2"))
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +178,95 @@ def _can_handle_query(cls, *query):
...
+            level = str(kwargs["level"])
+        else:
+            level = "2"
+
+        if str(level) == "2":
```
Cadair:
```suggestion
        if level == "2":
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +178,95 @@ def _can_handle_query(cls, *query):
...
+            search_pattern = base_url + 'l{level}/data/suvi-l{level}-ci{wave:03}/%Y/%m/%d/dr_suvi-l{level}-ci{wave:03}_g16_s%Y%m%dT%H%M%SZ_.*\.fits'
+        elif str(level) == "1b":
```
Cadair:
```suggestion
        elif level == "1b":
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +178,95 @@ def _can_handle_query(cls, *query):
...
+            if wavelength in u.Quantity([131, 171, 195, 284], 'Angstrom'):
+                search_pattern = base_url + 'l{level}/suvi-l{level}-fe{wave:03}/%Y/%m/%d/OR_SUVI-L{level}-Fe{wave:03}_G16_s%Y%j%H%M%S.*\.fits.gz'
+            elif wavelength == 304 * u.Angstrom:
+                search_pattern = base_url + 'l{level}/suvi-l{level}-he{wave:03}/%Y/%m/%d/OR_SUVI-L{level}-He{wave_minus1:03}_G16_s%Y%j%H%M%S.*\.fits.gz'
+            elif wavelength == 94 * u.Angstrom:
+                search_pattern = base_url + 'l{level}/suvi-l{level}-fe{wave:03}/%Y/%m/%d/OR_SUVI-L{level}-Fe{wave_minus1:03}_G16_s%Y%j%H%M%S.*\.fits.gz'
+        else:
+            return []
+        wave = int(wavelength.to('angstrom').value)
```
Cadair:
```suggestion
        wave = int(wavelength.to_value('angstrom'))
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +178,95 @@ def _can_handle_query(cls, *query):
...
+
+        if search_pattern.count('wave_minus1'):
```
Cadair: Aside: I feel like `Scraper` should just ignore extra parameters which are not in the pattern to prevent things like this being needed.

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +178,95 @@ def _can_handle_query(cls, *query):
...
+                    """
+        Answers whether client can service the query.
+
+        Parameters
+        ----------
+        query : list of query objects
+
+        Returns
+        -------
+        boolean
```
Cadair:
```suggestion
        `bool`
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +178,95 @@ def _can_handle_query(cls, *query):
...
+        return scraper.filelist(timerange)
+
+        @classmethod
+        def _can_handle_query(cls, *query):
+                    """
+        Answers whether client can service the query.
+
+        Parameters
+        ----------
+        query : list of query objects
```
Cadair:
```suggestion
        query : `tuple`
            All specified query objects
```

Cadair:
> Under what conditions should a GenericClient return an error?

The idea is that the `_can_handle_query` will filter out most "bad attr" type errors, otherwise it should throw errors as normal.

> Also how should defaults be handled? For examples, right now I default to level 2 data if it is not specified because that's what another client does but I don't default to any particular wavelength if it is not specified.

I would not default to any given wavelength, that seems like a bad idea. Defaulting to the highest data level makes sense to me on a general note, but all these decisions have been left to the clients. (In some cases VSO has the high level data and we have clients which have low level data).

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +178,95 @@ def _can_handle_query(cls, *query):
...
+            search_pattern = base_url + 'l{level}/data/suvi-l{level}-ci{wave:03}/%Y/%m/%d/dr_suvi-l{level}-ci{wave:03}_g16_s%Y%m%dT%H%M%SZ_.*\.fits'
+        elif str(level) == "1b":
+            if wavelength in u.Quantity([131, 171, 195, 284], 'Angstrom'):
+                search_pattern = base_url + 'l{level}/suvi-l{level}-fe{wave:03}/%Y/%m/%d/OR_SUVI-L{level}-Fe{wave:03}_G16_s%Y%j%H%M%S.*\.fits.gz'
+            elif wavelength == 304 * u.Angstrom:
+                search_pattern = base_url + 'l{level}/suvi-l{level}-he{wave:03}/%Y/%m/%d/OR_SUVI-L{level}-He{wave_minus1:03}_G16_s%Y%j%H%M%S.*\.fits.gz'
+            elif wavelength == 94 * u.Angstrom:
+                search_pattern = base_url + 'l{level}/suvi-l{level}-fe{wave:03}/%Y/%m/%d/OR_SUVI-L{level}-Fe{wave_minus1:03}_G16_s%Y%j%H%M%S.*\.fits.gz'
+        else:
+            return []
```
Cadair: I think you want to modify `_can_handle_query` to not be a valid query for this client if it's level 1b and not one of the allowed wavelengths?

sunpy/net/dataretriever/sources/goes.py
Cadair: I wasn't suggesting wildcards, I was suggesting that it not error if you pass it kwargs it has no place for in the template.

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -4,20 +4,25 @@
 
 import os
 from urllib.parse import urlsplit
+from datetime import timedelta
```
Cadair: please run isort on this file.

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +180,121 @@ def _can_handle_query(cls, *query):
...
+        satellitenumber = int(kwargs.get("satellitenumber", 16))
+        if "satellitenumber" not in kwargs.keys():
+            log.info(f'No satellite number provided so assuming GOES-{satellitenumber}.')
+
+        level = str(kwargs.get("level", "2"))
+        if "level" not in kwargs.keys():
+            log.info(f"No data level provided so assuming level {level}.")
+
+        if level not in supported_levels:
+            warnings.warn(f"Level not supported. Supported are f{supported_levels}.", SunpyUserWarning)
```
Cadair: This shouldn't even get here if level is not in supported levels.

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +180,121 @@ def _can_handle_query(cls, *query):
...
+        if wave not in supported_waves:
+            warnings.warn(f"Wavelength {wavelength} not supported. Supported are {supported_waves * u.Angstrom}.", SunpyUserWarning)
```
Cadair: Same here, this is what `_can_handle_query` is for.

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +180,121 @@ def _can_handle_query(cls, *query):
...
+        if satellitenumber < minimum_supported_satellitenumber:
+            warnings.warn(f"Satellite number not supported. Supported are >{minimum_supported_satellitenumber}.", SunpyUserWarning)
```
Cadair: And once more.

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -243,17 +242,22 @@ def _get_url_for_timerange(self, timerange, **kwargs):
         base_url = "https://data.ngdc.noaa.gov/platforms/solar-space-observing-satellites/goes/goes{goes_number}/"
         supported_waves = [94, 131, 171, 195, 284, 304]
         supported_levels = ("2", "1b")
+        min_satellite_number = 16  # when SUVI was first included
 
         # these are optional requirements so if not provided assume defaults
         # if wavelength is not provided assuming all of them
         if "wavelength" in kwargs.keys():
+            if int(kwargs.get("wavelength").to_value('Angstrom')) not in supported_waves:
```
Cadair: can you test what happens if wavelength is a range. i.e `a.Wavelength(90*u.AA, 150*u.AA)` should get both 94 and 131. I don't think this will work like that though.

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +180,120 @@ def _can_handle_query(cls, *query):
...
+        if ops and ops != optional:
+            return False
+
+        # if we get this far we have either Instrument and Time
+        # or Instrument, Time and Wavelength
+        check_var_count = 0
+        for x in query:
+            if isinstance(x, a.Instrument) and x.value.lower() == 'suvi':
+                check_var_count += 1
+            if isinstance(a, a.Wavelength) and int(x.value.to_value('Angstrom')) is in supported_waves:
```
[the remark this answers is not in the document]
ehsteve: so this is no longer checked in this function as we discussed and agreed. How does this range get passed to `_get_url_for_timerange`?

sunpy/net/dataretriever/sources/goes.py
ehsteve: How does this range get passed to `_get_url_for_timerange`?

sunpy/net/dataretriever/sources/goes.py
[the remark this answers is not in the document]
Cadair: I honestly can't remember, try it and find out!

*I would guess it's a tuple*

sunpy/net/dataretriever/sources/goes.py
Cadair: see above.

sunpy/net/dataretriever/sources/goes.py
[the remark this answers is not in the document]
ehsteve: It's a minmax object but only if it is called `a.Wavelength(q1, q2)` otherwise the bare quantity object get passed. This is weird. and probably very few clients are currently handling this. Would make more sense to always pass a `minmax` object but with the same values in both.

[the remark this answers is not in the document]
ehsteve: Why does it pass the Wavelength attr object directly?!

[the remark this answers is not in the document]
ehsteve: `minmax` is a named tuple. 
```
namedtuple("minmax", "{0}min {0}max".format(prefix))
```

sunpy/net/dataretriever/tests/test_goes_ud.py
```diff
@@ -129,3 +130,17 @@ def test_time_for_url(LCClient, time):
...
 
     assert all([tr == t2 for t2 in times])
+
+
+
+@pytest.mark.remote_data
+@pytest.mark.parametrize("time,instrument", [
+    (a.Time('2019/05/13 00:00', '2019/05/13 00:10'), a.Instrument('suvi')),
+])
+def test_query(time, instrument):
```
Cadair: This test should be in `test_goes_suvi.py` and you should use the `suvi_client` fixture in that file.

ehsteve: done.

sunpy/net/dataretriever/tests/test_goes_ud.py
```diff
@@ -129,3 +130,17 @@ def test_time_for_url(LCClient, time):
...
+    qr1 = sclient.search(time, instrument)
+    assert isinstance(qr1, QueryResponse)
+    assert len(qr1) == 3
+    assert qr1.time_range().start == time.start
+    assert qr1.time_range().end == time.end
```
nabobalis: DELETE THIS

sunpy/net/dataretriever/tests/test_goes_suvi.py
```diff
@@ -0,0 +1,205 @@
...
+def test_fido_waverange_level1b(start, end, wave1, wave2, expected_num_files):
+    """check that we get all wavelengths if no wavelength is given"""
+    result = Fido.search(a.Time(start, end), a.Instrument('suvi'),
+                         a.Wavelength(wave1 * u.Angstrom, wave2 * u.Angstrom),
+                         a.Level('1b'))
+    assert result.file_num == expected_num_files
+
+
+@pytest.mark.remote_data
+@pytest.mark.parametrize("start, end, expected_num_files",
```
nabobalis: rename these?

ehsteve: Why?

ehsteve:
@Cadair please update your review. I think the test fails are not my fault (I hope)? They pass on my machine.

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +177,163 @@ def _can_handle_query(cls, *query):
...
+            wavelength_input = kwargs.get("wavelength")
+            if isinstance(wavelength_input, u.Quantity):  # not a range
+                if int(wavelength_input.to_value('Angstrom')) not in supported_waves:
+                    raise ValueError(f"Wavelength {kwargs.get('wavelength')} not supported.")
+                else:
+                    wavelength = [kwargs.get("wavelength")]
+            else:  #  _Range was provided
+                compress_index = [wavelength_input.wavemin <= this_wave <= wavelength_input.wavemax for this_wave in (supported_waves * u.Angstrom)]
+                if not any(compress_index):
+                    raise ValueError(
```
Cadair: I am not sure throwing a `ValueError` here is the correct thing to do. Doing this will kill outright any `Fido` query which misses the windows. Take for example the following query (where AIA has both and SUVI only has one passband):

```python
Fido.search(a.Instrument('SUVI') | a.Instrument('AIA'), a.Wavelength(30.4*u.nm) | a.Wavelength(211*u.AA))
```

This error here would cause it to fail, where as if it returned no results this query would work.

I think this is yet another case of the need to give users feedback on query validity, I just don't think this is the way to do it.

changelog/3301.feature.rst
```diff
@@ -0,0 +1 @@
+Adds support for searching and downloading SUVI data
```
nabobalis:
```suggestion
Adds support for searching and downloading SUVI data.
```

Maybe a url for suvi incase anyone wants to look it up?

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +177,163 @@ def _can_handle_query(cls, *query):
...
+                else:
+                    wavelength = list(compress(supported_waves, compress_index)) * u.Angstrom
+        else:  # no wavelength provided return all of them
+            wavelength = supported_waves * u.Angstrom
+        # check that the input wavelength can be converted to angstrom
+        waves = [int(this_wave.to_value('angstrom', equivalencies=u.spectral())) for this_wave in wavelength]
+        # use the given satellite number or choose the best one
+        satellitenumber = int(kwargs.get("satellitenumber", self._get_goes_sat_num(timerange.start)))
+        if satellitenumber < 16:
+            raise ValueError(f"Satellite number {satellitenumber} not supported.")
```
Cadair: Same potentially applies here, at least for consistency, although it would be a lot harder to come up with a breaking query.

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +177,163 @@ def _can_handle_query(cls, *query):
...
+        # default to the highest level of data
+        level = str(kwargs.get("level", "2"))  # make string in case the input is a number
+
+        if level not in supported_levels:
+            raise ValueError(f"Level {level} is not supported.")
```
Cadair: Same applies here as well. As other clients support Level

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -3,21 +3,23 @@
...
 
-from datetime import timedelta
-from sunpy.time import parse_time, TimeRange
 from ..client import GenericClient
-from sunpy import config
 
 TIME_FORMAT = config.get("general", "time_format")
 
-__all__ = ['XRSClient']
+__all__ = ["XRSClient", "SUVIClient"]
```
nabobalis: Did you run a linter?

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +177,163 @@ def _can_handle_query(cls, *query):
...
+
+class SUVIClient(GenericClient):
+
+    def _get_goes_sat_num(self, date):
+        """
+        Determines the best satellite number for a given date.
+
+        Parameters
+        ----------
+
```
nabobalis:
```suggestion
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +177,163 @@ def _can_handle_query(cls, *query):
...
+        date : `astropy.time.Time`
```
nabobalis: We don't want to use the fancy time doc string magic?

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +177,163 @@ def _can_handle_query(cls, *query):
...
+            # if no satellites were found then raise an exception
+            raise ValueError(f"No operational SUVI instrument on {date.strftime(TIME_FORMAT)}")
+
+    def _get_url_for_timerange(self, timerange, **kwargs):
+        """
+        Returns a URL to the SUVI data for the specified time range.
+
+        Parameters
+        ----------
+        timerange: sunpy.time.TimeRange
```
nabobalis:
```suggestion
        timerange: `sunpy.time.TimeRange`
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +177,163 @@ def _can_handle_query(cls, *query):
...
+            time range for which data is to be downloaded.
```
nabobalis:
```suggestion
            Time range for which data is to be downloaded.
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +177,163 @@ def _can_handle_query(cls, *query):
...
+        level : str
```
nabobalis:
```suggestion
        level : `str`
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +177,163 @@ def _can_handle_query(cls, *query):
...
+            The level of the data.
+        wavelength :
```
nabobalis: This lacks a type?

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +177,163 @@ def _can_handle_query(cls, *query):
...
+            Wavelength band.
+        satellitenumber : int
```
nabobalis:
```suggestion
        satellitenumber : `int`
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +177,163 @@ def _can_handle_query(cls, *query):
...
+
+    @classmethod
+    def _can_handle_query(cls, *query):
+        """
+        Answers whether client can service the query.
+
+        Parameters
+        ----------
+        query : `tuple`
+            All specified query objects
```
nabobalis:
```suggestion
            All specified query objects.
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +177,163 @@ def _can_handle_query(cls, *query):
...
+
+        Returns
+        -------
+        `bool`
+            answer as to whether client can service the query
```
nabobalis:
```suggestion
            answer as to whether client can service the query.
```

sunpy/net/dataretriever/tests/test_goes_suvi.py
nabobalis: You fixed it.

sunpy/net/dataretriever/tests/test_goes_ud.py
```diff
@@ -13,6 +13,7 @@
 from sunpy.net import attrs as a
 from sunpy.net.tests.strategies import goes_time
 from sunpy.time import parse_time, is_time_equal
+from sunpy.net.tests.strategies import time_attr
```
nabobalis:
```suggestion
```

sunpy/net/dataretriever/sources/goes.py
Cadair: We should.

sunpy/net/dataretriever/sources/goes.py
ehsteve: yes, I likely ran black at some point.

sunpy/net/dataretriever/sources/goes.py
ehsteve: This is consistent with how the `XRSClient` does it.

sunpy/net/dataretriever/sources/goes.py
ehsteve: I was just following what the `XRSClient` does. I'm not saying it's the right thing to do but since this PR is not meant to be controversial, I was just following the example of code that was already being used.

sunpy/net/dataretriever/sources/goes.py
Cadair: Like this: https://github.com/sunpy/sunpy/blob/master/sunpy/time/time.py#L258

sunpy/net/dataretriever/sources/goes.py
ehsteve: This is a namedtuple or a quantity. I'm not sure how how to document that since it expects the named tuple to be of a certain kind.

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -168,9 +177,162 @@ def _can_handle_query(cls, *query):
...
+
+class SUVIClient(GenericClient):
+
+    def _get_goes_sat_num(self, date):
+        """
+        Determines the best satellite number for a given date.
+
+        Parameters
+        ----------
+        date : {parse_time_types}
```
ayshih: This function needs the decorator `@add_common_docstring(**_variables_for_parse_time_docstring())` for this field to expand as intended.

hayesla:
Looks great @ehsteve!  - I wonder how to include the wavelength info in the `UnifiedResponse` when no wavelength is given? and the start and end times of the files rather than the search timerange for each search result


For example:
```
>>> res = Fido.search(a.Time('2018/06/11 00:00', '2018/06/11 01:00'), a.Instrument('SUVI'), a.Level('1b')) 
>>> res
<sunpy.net.fido_factory.UnifiedResponse object at 0x125035ef0>
Results from 1 Provider:

358 Results from the SUVIClient:
       str19               str19         str4     str4       str3   
2018-06-11 00:00:00 2018-06-11 01:00:00   GOES       SUVI        nan
2018-06-11 00:00:00 2018-06-11 01:00:00   GOES       SUVI        nan
2018-06-11 00:00:00 2018-06-11 01:00:00   GOES       SUVI        nan
2018-06-11 00:00:00 2018-06-11 01:00:00   GOES       SUVI        nan
2018-06-11 00:00:00 2018-06-11 01:00:00   GOES       SUVI        nan
2018-06-11 00:00:00 2018-06-11 01:00:00   GOES       SUVI        nan
...
```

I can't figure out where this info should be given - I guess in the `QueryResponseBlock` in `client.py` - @Cadair ?

ehsteve:
@hayesla my understanding is that that is not possible currently with Fido and needs to be fixed. The clients only return a url list and nothing more. @Cadair and I talked about this and this is something that needs to be fixed going forward.

Cadair:
The CI fails look like they are something in astroquery being sad. I would like to see a passing doc build before merge though.

sunpy/net/dataretriever/tests/test_goes_suvi.py
```diff
@@ -0,0 +1,203 @@
...
+
+    mock_qr = mock_qr[0]
+    qr = qr1[0]
+
+    assert mock_qr.source == qr.source
+    assert mock_qr.provider == qr.provider
+    assert mock_qr.physobs == qr.physobs
+    assert mock_qr.instrument == qr.instrument
+    assert mock_qr.url == qr.url
+    assert mock_qr.time == qr.time
```
hayesla: this time check is now failing - I think its now because the time timedates from `get_time_for_url`?

ayshih:
I'll point out that the commit history of this PR is atrocious, and really ought to have commits squashed and messages improved before it's merged.  I'm not going to raise (too much of) a fuss here, because it's hardly the first instance of such a commit history being accepted into our repo, but see what we say in our own developer's guide: https://docs.sunpy.org/en/latest/dev_guide/maintainer_workflow.html?highlight=commit#when-to-remove-or-combine-squash-commits.  Either we mean it or we don't.

Cadair:
:+1: to squash merging this.

ehsteve:
I was going to suggest squashing the commit history myself so I'm definitely fine with it. Perhaps we should even have a rule that PRs should really consist of just one or a few commits.

PR #3302, opened 2019-08-13
New doc entry on requirements for data source docstring

ehsteve:
### Description

The purpose of this documentation is to provide requirements for data source docstrings (e.g. Map, TimeSeries) so that we provide consistent and complete information about our data sources.


ehsteve:
We have no sub package maintainers for developer docs...which makes me sad.

docs/dev_guide/new_objects.rst
```diff
@@ -4,6 +4,37 @@
 Creating new SunPy Subclasses (Maps, TimeSeries)
 ************************************************
 
+Documenting new data sources
+----------------------------
+Each subclass of `~sunpy.map.GenericMap` or `~sunpy.timeseries.TimeSeries`
```
Cadair:
```suggestion
Each subclass of `~sunpy.map.GenericMap` or `~sunpy.timeseries.GenericTimeSeries`
```

ehsteve:
@wafels any comments?

ehsteve:
calling all TImeSeries sub package maintainers (@hayesla @dstansby), comments on this PR?

docs/dev_guide/new_objects.rst
```diff
@@ -4,6 +4,37 @@
...
+* the name of the mission and instrument and the institution that built it
+* a short description of the instrument (e.g. Cassegrain reflector, Wolter-1 grazing incidence x-ray, coronagraph) including the type of detector
+* a description of the platform (e.g. satellite in 28 deg inclined orbit, a telescope on the summit of Mauna Kea in Hawaii)
+* a description of the primary purpose or science goals of the instrument.
+* a list of all wavelength(s) or passbands in appropriate units
+* a description of the emission processes which dominate in those passbands
+* the field of view and resolution (e.g. angular resolution)
+* the measurement cadence
+* a description of the operational concept (e.g. operates 24/7, observes from 7 am to 5 pm UT)
+* the start and end of the data set
```
hayesla: It might be also good to have a caveats section, such as 'be warned, this instrument looks off at the crab nebula every second tuesday' or the fact that it only supports a certain data product level or things of note along those lines for instrument specific cases.

docs/dev_guide/new_objects.rst
```diff
@@ -4,6 +4,37 @@
...
+Each subclass of `~sunpy.map.GenericMap` or `~sunpy.timeseries.GenericTimeSeries`
+must provide a detailed docstring providing an overview of the data source that
+the object represents. In order to maintain consistency and completeness the
+following information must be provided by a data source docstring, if available,
+and preferably in the following order:
+
+* the name of the mission and instrument and the institution that built it
+* a short description of the instrument (e.g. Cassegrain reflector, Wolter-1 grazing incidence x-ray, coronagraph) including the type of detector
+* a description of the platform (e.g. satellite in 28 deg inclined orbit, a telescope on the summit of Mauna Kea in Hawaii)
+* a description of the primary purpose or science goals of the instrument.
```
dstansby: Science goals seems like a very broad/vague requirement - much better to link to an instrument paper that provides this information.

docs/dev_guide/new_objects.rst
ehsteve: I was thinking that that would be part of the operational concept section. I'll add text to make this more clear.

docs/dev_guide/new_objects.rst
```diff
@@ -4,6 +4,37 @@
...
+* a list of all wavelength(s) or passbands in appropriate units
```
dstansby: This only appliess for maps (not necessarily timeseries data). Same for a lot of the points below, these should be specified as `Map` specific.

dstansby:
Looks good overall, but there are several points that are specific to `Map` objects, and not general `Timeseries` data.

docs/dev_guide/new_objects.rst
ehsteve: I was thinking that providing a single sentence summarizing what the instrument was designed to do would be useful in most cases.

docs/dev_guide/new_objects.rst
ehsteve: Sorry, how is this specific to `Map`? Like letting users know which passbands XRS data comes from or the passbands the RHESSI summary light curve data. I'm confused by your comment here.

dstansby: Sorry, I was thinking about this with my `HelioPy` hat on (which uses `TimeSeries`), but I guess if this bit of doc is `SunPy` specific (not recommendations for other packages that use `TimeSeries`) then it holds.

Cadair: I think this could be made a little more generic, i.e "A list of physical parameters represented by the data type, i.e. wavelength bands, line of sight magnetic field, particle types".

Not all Map data corresponds to a wavelength, i.e. HMI LOS Magnetic field is sufficiently detached from a pass band imo.

Cadair:
> you can revert back to this monstrous format of text writing

Not on my watch :stuck_out_tongue:

docs/dev_guide/new_objects.rst
```diff
@@ -4,43 +4,88 @@
 Creating new SunPy Subclasses (Maps, TimeSeries)
 ************************************************
 
+Documenting new data sources
```
Cadair: I don't think this should go above the instructions for how to make the class, it should go below.

Cadair: I think it makes sense to have it in here. I would suppose it's more likely people will see it if it's here.

ehsteve:
@nabobalis please update your review! thanks.

ehsteve:
thanks @nabobalis! @Cadair ping!

PR #3304, opened 2019-08-14
Bye Bye Travis

Cadair:


dpshelio:
Thanks for all the :fish:.... 

Should we have some place the history of our CI and why the changes? It would help future generations to know what was missing or good from the different CIs

Cadair:
My philosophy on CI services is never to get too attached to one or another. Azure is great for now, it's configurable, supports templates and a load of other magic and gives us the most build time.

dpshelio:
I know what's your philosophy, and I totally agree with it. But I think it would help to have a better description of why we stop to use one or start to use another. So we save from future contributions that say: "¡hey! did you know about Travis, here's a file that will make it run there".

nabobalis:
I want to try to keep any CI changes the same on 1.0.X and master.

nabobalis:
> I know what's your philosophy, and I totally agree with it. But I think it would help to have a better description of why we stop to use one or start to use another. So we save from future contributions that say: "¡hey! did you know about Travis, here's a file that will make it run there".

Then Stuart can reply saying why we don't use it or point to 0.9 and go we did.

Cadair:
Documenting our CI is a good idea, means that hopefully it's not just stored in @nabobalis's brain :laughing:

Cadair:
I turned off the cron on master.

nabobalis:
We do have some words on it in the dev guide. It probably needs updating.

nabobalis:
https://docs.sunpy.org/en/latest/dev_guide/pr_review_procedure.html#continuous-integration

nabobalis:
> Documenting our CI is a good idea, means that hopefully it's not just stored in @nabobalis's brain

It is also stored in your brain so I don't see an issue.

PR #3313, opened 2019-08-21
Delay the map import in the asdf tag

Cadair:
This code is imported when the asdf entrypoint is loaded, and the map import has many dependencies and takes some time, so this is just a little performance improvement.


Cadair:
This looks like it might have introduced a bug, better not merge it till I have investigated a little more.

Cadair:
yep this doesn't work.

PR #3317, opened 2019-08-24
Overhaul of coordinate-frame docstrings

### LINKED ISSUE #3314: Leverage `add_common_docstring` for the coordinate-frame docstrings
Currently, the coordinate-frame docstrings repeat *a lot* of the same parameter text between them.  We should leverage `add_common_docstring` to get rid of the repetition, which will also help future maintenance.

ayshih:
- Coordinate-frame docstrings now leverage `add_common_docstring()` for repetitive parts
- Documented `representation_type` and `copy` as optional keywords
- Minor cleanup

Closes #3314

### DIFF
--- a/sunpy/coordinates/frames.py
+++ b/sunpy/coordinates/frames.py
@@ -24,6 +24,55 @@ __all__ = ['HeliographicStonyhurst', 'HeliographicCarrington',
            'Heliocentric', 'Helioprojective']
 
 
+def _frame_parameters():
+    """
+    Returns formatting dictionary to use with add_common_docstring to populate frame docstrings
+    """
+    ret = {}
+
+    # Each text block is missing the first indent because it already exists in the frame docstring
+    ret['data'] = ("data : `~astropy.coordinates.BaseRepresentation` or ``None``\n"
+                   "        A representation object or ``None`` to have no data\n"
+                   "        (or use the coordinate component arguments, see below).")
+    ret['common'] = (f"obstime : {_variables_for_parse_time_docstring()['parse_time_types']}\n"
+                     "        The time of the observation.  This is used to determine the\n"
+                     "        position of solar-system bodies (e.g., the Sun and the Earth) as\n"
+                     "        needed to define the origin and orientation of the frame.\n"
+                     "    representation_type : `~astropy.coordinates.BaseRepresentation`, str, optional\n"
+                     "        A representation class or string name of a representation class.\n"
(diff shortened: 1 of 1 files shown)


Cadair:
The fact the doc build is failing really needs addressing before we merge this. (I don't think it's related)

nabobalis:
```
/home/circleci/project/examples/units_and_coordinates/AIA_limb_STEREO.py failed to execute correctly: Traceback (most recent call last):
  File "/home/circleci/project/.tox/build_docs/lib/python3.7/site-packages/sphinx_gallery/gen_rst.py", line 480, in _memory_usage
    out = func()
  File "/home/circleci/project/.tox/build_docs/lib/python3.7/site-packages/sphinx_gallery/gen_rst.py", line 465, in __call__
    exec(self.code, self.globals)
  File "/home/circleci/project/examples/units_and_coordinates/AIA_limb_STEREO.py", line 49, in <module>
    for m in sunpy.map.Map(downloaded_files)}
  File "/home/circleci/project/examples/units_and_coordinates/AIA_limb_STEREO.py", line 47, in <dictcomp>
    maps = {m.detector: m.submap(SkyCoord([-1100, 1100], [-1100, 1100],
  File "/home/circleci/project/.tox/build_docs/lib/python3.7/site-packages/sunpy/map/mapbase.py", line 216, in __getitem__
    "The ability to index Map by physical"
NotImplementedError: The ability to index Map by physical coordinate is not yet implemented.
```
Some server is done.

ayshih:
The docs build fine locally, so I can believe that the CI failure was because of not being able to reach a server.

*However*, I'm befuddled about the error and traceback.  It looks like a broken Map object was created from whatever Fido gave it, but shouldn't `Map()` not even return an object?  It's obviously unrelated to this PR, but I wanted to point it out in case it needs to be tracked as a new issue.

Cadair:
@ayshih What happens is that instead of two files Fido gives `Map` one file and therefore it returns a `Map` rather than a list, it's not an issue with map so much as it is a weird side effect of the way the example is setup.

ayshih:
Ah, I understand now.  The issue is that `m in map_object` calls `__getitem__()`, which raises the `NotImplementedError`, but the error message is totally misleading, since no indexing by physical coordinate is being attempted.

nabobalis:
the doc build is taking 15 minutes to upload. We should either fix that by uploading less of the build folder or seeing if its circle being slow one time.

PR #3320, opened 2019-08-28
update docs and fix init format

### LINKED ISSUE #3319: SunPy import python check uses Python 3.6 syntax
Line 25 in `__init__.py` checks for the python version but assumes at least Python 3.6 since it uses an f-string.

```python

raise UnsupportedPythonError(f"sunpy does not support Python < {__minimum_python_version__}")
```
This means that users with versions below 3.6 will not actually see the intended output but will just get a less helpful error. This line should be able to run in Python 2.

nabobalis:
Fixes #3319 again and updates docs/conf to support only sphinx 2 or higher.


docs/conf.py
```diff
@@ -31,18 +31,14 @@
...
 from distutils.version import LooseVersion
 
-from sphinx import __version__
-SPHINX_LT_17 = LooseVersion(__version__) < LooseVersion('1.7')
-
 # -- Convert Sphinx Warnings to output to stdout not stderr---------------------
-if not SPHINX_LT_17:
-    import logging
-    from sphinx.util.logging import NAMESPACE, WarningStreamHandler, SafeEncodingWriter
+import logging
```
Cadair: Does this even work? Can we drop this hack of mine?

nabobalis: Not a clue. I can remove it.

nabobalis: Well the warnings are all there before so I'm gonna say no it doesn't.

PR #3269, opened 2019-07-11
Add support for SUVI images

wafels:
### Description

Adds support for SUVI images as a map source

Fixes #2013

PR #3322, opened 2019-08-28
Test data for the SUVI map client

wafels:
A rebinned (and therefore smaller) SUVI Level 2 file for use for testing with issue #3269.  Please let me know if this is small enough, and sufficient.  I used a relatively recent SUVI file.  Please let me know if you think I should also include a rebinned SUVI Level 1 file also.  Let me know also if you think the file should be included in #3269.

### Description

Fixes #3269


Cadair:
yeah I would say so.

ayshih:
> Let me know also if you think the file should be included in #3269.

I certainly think that this file should be included in #3269.  How will that PR have good unit tests without a test file?  And this PR basically has no value without #3269 being merged first.

> Fixes #3269

Note that by writing this line, merging this PR would close #3269, which I'm sure isn't what you intend.

Cadair:
> this PR would close #3269

what the! Why does that work for PRs as well as issues lol

ayshih:
It could make sense: if two PRs solve the same issue through different implementations, you could have the merge of one automatically close the other.  Of course, "fixes" isn't the right verb in that situation.

wafels:
I didn't want to make a PR with a file that is too big.  That is why this file is in a separate PR.  It is true that #3269 is worth less without the test data.  I'll close this PR and add the file to #3269.  Thanks for your comments.

PR #3330, opened 2019-08-30
Add Debug logging to VSOClient 

Cadair:
This logs at the DEBUG level all the VSO envelopes that are sent between our client and the VSO. You can enable it by setting the sunpy logger to debug:

```python
>>> from sunpy.net import Fido, attrs as a
>>> import astropy.units as u
>>> from sunpy import log

>>> log.setLevel("DEBUG")                                                                                                                                                                                        

>>> Fido.search(a.Time('2012-03-04','2012-03-06'), a.Instrument("EUVI"), a.Wavelength(304*u.AA), a.Sample(100*u.min))                                
```
```                                                                                                                                                      
DEBUG: VSO Request:
 <soap-env:Envelope xmlns:VSO="http://virtualsolar.org/VSO/VSOi" xmlns:soap-env="http://schemas.xmlsoap.org/soap/envelope/">
  <soap-env:Body>
    <VSO:Query>
      <body>
        <block>
          <time>
            <start>20120304000000</start>
            <end>20120306000000</end>
          </time>
          <instrument>EUVI</instrument>
          <wave>
            <wavemin>304.0</wavemin>
            <wavemax>304.0</wavemax>
            <waveunit>Angstrom</waveunit>
          </wave>
          <sample>6000.0</sample>
        </block>
      </body>
    </VSO:Query>
  </soap-env:Body>
</soap-env:Envelope>
 [sunpy.net.vso.zeep_plugins]
DEBUG: VSO Response:
 <soap:Envelope xmlns:VSO="http://virtualsolar.org/VSO/VSOi" xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <soap:Body>
    <VSO:QueryResponse>
      <body>
        <provideritem>
          <provider>SSC</provider>
          <error>VSO-D404 Bad Request -- Invalid Time Sample : contains non digit</error>
          <version>1</version>
        </provideritem>
      </body>
    </VSO:QueryResponse>
  </soap:Body>
</soap:Envelope>
 [sunpy.net.vso.zeep_plugins]
Out[2]: 
<sunpy.net.fido_factory.UnifiedResponse object at 0x7efbf7608790>
Results from 1 Provider:

0 Results from the VSOClient:
Start Time End Time  Source Instrument   Type 
 float64   float64  float64  float64   float64
---------- -------- ------- ---------- -------

### DIFF
--- a/changelog/3330.breaking.rst
+++ b/changelog/3330.breaking.rst
@@ -1,0 +1,2 @@
+The `sunpy.net.vso.vso.get_online_vso_url` function has been broken into two components, the new `sunpy.net.vso.vso.get_online_vso_url` function takes no arguments (it used to take three) and now only returns an online VSO mirror or None.
+The construction of a `zeep.Client` object is now handled by `sunpy.net.vso.vso.build_client` which has a more flexible API for customising the `zeep.Client` interface.
--- a/changelog/3330.feature.rst
+++ b/changelog/3330.feature.rst
@@ -1,0 +1,2 @@
+Log all VSO XML requests and responses to the SunPy logger at the ``DEBUG``
+level.
--- a/sunpy/net/tests/test_vso.py
+++ b/sunpy/net/tests/test_vso.py
@@ -8,7 +8,7 @@ from parfive import Results
 from sunpy.time import TimeRange, parse_time
 from sunpy.net import vso
 from sunpy.net.vso import attrs as va
-from sunpy.net.vso.vso import VSOClient, get_online_vso_url
+from sunpy.net.vso.vso import VSOClient, get_online_vso_url, build_client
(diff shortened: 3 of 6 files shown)


Cadair:
This as it is is walking a line with an API change `sunpy.net.vso.vso.get_online_vso_url` now does what it says on the tin and gives you the port, url pair for the first online VSO mirror. Before it used to give you the `zeep.Client` object and it used to take 3 args and now takes 0.

I don't really consider something in the `sunpy.net.vso.vso` namespace "public" API, but I am open to being persuaded...

PR #3332, opened 2019-08-30
Update to _Range object

ehsteve:
### Description

This change allows for the following
```python
171 * u.mm in a.Wavelength(100 * u.mm, 200 * u.mm)
```

Fixes NA

### DIFF
--- a/sunpy/net/vso/attrs.py
+++ b/sunpy/net/vso/attrs.py
@@ -61,8 +61,10 @@ class _Range:
         return new
 
     def __contains__(self, other):
-        return self.min <= other.min and self.max >= other.max
-
+        if isinstance(other, _Range):
+            return self.min <= other.min and self.max >= other.max
+        else:
+            return self.min <= other <= self.max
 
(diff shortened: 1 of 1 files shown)


Cadair:
Looks good, changelog entry and tests please!

ehsteve:
@Cadair I hope you understand that this part of my master plan to have the attrs sent to the clients. ;)

Cadair:
You mean instead of the kwargs for `_get_url_for_timerange`?

ehsteve:
Just pass it as a list to an attr keyword. Then a client could just check what was in the list and not need to figure out how attrs are related to keywords and what gets passed in. Presumably they would already have some familiarity with attrs as a user.

Cadair:
Yeah, I think in general it makes sense to use the attrs as dev api as well. This probably isn't the right place for this discussion though :laughing:

ehsteve:
Just my quick idea anyway

wafels:
could do the same for a.Time ?

PR #3334, opened 2019-09-01
Great Arc now calculable when two different observers are defined

wafels:
### Description

Previously the great arc code could only generate arc coordinates if both the start and end point of the arcs had the same observer location.  This PR now allows users to pass in start and end coordinates that have different observers.  The coordinates are returned from the point of view of the start observer.

### DIFF
--- a/changelog/3334.bugfix.rst
+++ b/changelog/3334.bugfix.rst
@@ -1,1 +1,1 @@
-GreatArc now accounts for the start and end points of the arc having different observers.
+`~sunpy.coordinates.utils.GreatArc` now accounts for the start and end points of the arc having different observers.
--- a/changelog/3334.bugfix.rst
+++ b/changelog/3334.bugfix.rst
@@ -1,0 +1,1 @@
+GreatArc now accounts for the start and end points of the arc having different observers.
--- a/sunpy/coordinates/tests/test_utils.py
+++ b/sunpy/coordinates/tests/test_utils.py
@@ -32,8 +32,20 @@ def test_great_arc_calculable(start, end):
                  observer=frames.HeliographicStonyhurst(0*u.deg, 0*u.deg, 1*u.AU))
     gc = GreatArc(c, d)
 
-    assert gc.start == c
-    assert gc.end == d
+    assert gc.start.x == c.transform_to(frames.Heliocentric).x
+    assert gc.start.y == c.transform_to(frames.Heliocentric).y
+    assert gc.start.z == c.transform_to(frames.Heliocentric).z
+    assert gc.start.observer.lat == 0*u.deg
+    assert gc.start.observer.lon == 0*u.deg
+    assert gc.start.observer.radius == 1 * u.AU
+
+    assert gc.end.x == d.transform_to(frames.Heliocentric(observer=c.observer)).x
+    assert gc.end.y == d.transform_to(frames.Heliocentric(observer=c.observer)).y
+    assert gc.end.z == d.transform_to(frames.Heliocentric(observer=c.observer)).z
+    assert gc.end.observer.lat == 0*u.deg
+    assert gc.end.observer.lon == 0*u.deg
+    assert gc.end.observer.radius == 1 * u.AU
+
     np.testing.assert_almost_equal(gc.inner_angle.to('deg').value, 45.0)
     np.testing.assert_almost_equal(gc.radius.to('km').value, sun.constants.radius.to('km').value)
     np.testing.assert_almost_equal(gc.distance.to('km').value, sun.constants.radius.to('km').value * 2 * np.pi/8, decimal=1)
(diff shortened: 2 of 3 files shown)


sunpy/coordinates/tests/test_utils.py
```diff
@@ -32,8 +32,20 @@ def test_great_arc_calculable(start, end):
                  observer=frames.HeliographicStonyhurst(0*u.deg, 0*u.deg, 1*u.AU))
     gc = GreatArc(c, d)
 
-    assert gc.start == c
-    assert gc.end == d
+    assert gc.start.x == c.transform_to(frames.Heliocentric).x
```
ayshih: Here and later, perform the transformation once for each set of components rather than once for each component.

sunpy/coordinates/tests/test_utils.py
```diff
@@ -60,7 +72,7 @@ def test_great_arc_calculable(start, end):
 def test_great_arc_coordinates(points_requested, points_expected, first_point,
                                last_point, last_inner_angle, last_distance):
 
-    m = sunpy.map.Map(sunpy.map.Map(sunpy.data.test.get_test_filepath('aia_171_level1.fits')))
+    m = sunpy.map.Map(sunpy.data.test.get_test_filepath('aia_171_level1.fits'))
```
ayshih: Make this a fixture

ayshih:
@wafels I see that you have milestone this for 0.9.x.  Because of the bevy of updates to the transformation machinery in 1.0.x, I'm skeptical whether this would actually work as desired in 0.9.x for all possible inputs.

wafels:
> You stripped the `obstime` stuff. What is your intent when `start` and `end` have different values for `obstime`?

Hmmm.  Good point.  I missed that.  I will put the obstime back in, but it will be the start obstime, since the returned frame is also the start frame.  I will include a note in the doc string to make this clear.

Cadair:
@ayshih Presumably this fix is still valid for 0.9 though even if it's only to the limitation of the coordinate stack.

nabobalis:
Is this even possible to auto backport to 0.9?

Cadair:
I don't see why not. The worst that will happen is that it will conflict horribly.

sunpy/coordinates/utils.py
```diff
@@ -80,11 +84,21 @@ class GreatArc:
...
+        # Observation time
+        self.obstime = start.obstime
+
         # Start point of the great arc
-        self.start = start
+        self.start = start.transform_to(Heliocentric)
 
         # End point of the great arc
-        self.end = end
+        self.end = end.transform_to(Heliocentric(observer=self.observer))
```
ayshih: You don't specify the `obstime` for the destination frame.  But instead of fixing that, you should probably write `end.transform_to(self.start.frame)`, which makes it more clear what you are doing.

ayshih:
> @ayshih Presumably this fix is still valid for 0.9 though even if it's only to the limitation of the coordinate stack.

The fix here may be valid, but I suspect that the underlying transformation machinery might actually give the wrong answers.  (I have not confirmed this, so my fear may not be warranted, but I fixed a *lot* of bugs in 1.0.x.)  Backporting this fix to 0.9.x makes sense only if the output is actually correct.

Cadair:
We are nearing the end of support for 0.9 now anyway. Given @ayshih's concerns I think we will just release this in the 1.0 series, unless you have particular motivation for it to go in 0.9 @wafels ?

nabobalis:
Was this finished @wafels?

nabobalis:
I merged in master, addressed Alberts comments and found a "bug" I re-jigged the tests for but left a comment.

changelog/3334.bugfix.rst
```diff
@@ -0,0 +1 @@
+GreatArc now accounts for the start and end points of the arc having different observers.
```
ayshih:
```suggestion
`~sunpy.coordinates.utils.GreatArc` now accounts for the start and end points of the arc having different observers.
```

nabobalis:
Do you want to open an issue so we don't forget to do that renaming later on?

nabobalis:
I can edit the file via the web ui but i cant add commit suggestions?! What?

ayshih:
> Do you want to open an issue so we don't forget to do that renaming later on?

Just created #3460

nabobalis:
Looks like online test issue: 
```
def test_fetch_working(tmpdir):
        """
        Tests if the online server for noaa is working.
        Uses the url : ftp://ftp.swpc.noaa.gov/pub/weekly/RecentIndices.txt
        """
        qr1 = LCClient.search(Time('2012/10/4', '2012/10/6'),
                              Instrument('noaa-indices'))
 ```
and
```
>>> hek2vso.translate_results_to_query(q[0])  # doctest: +REMOTE_DATA
Expected:
    [[<Time(<Time object: scale='utc' format='isot' value=2011-08-08T01:30:04.000>, <Time object: scale='utc' format='isot' value=2011-08-10T00:00:04.000>, None)>, <Source('SDO')>, <Instrument('HEK')>, <Wavelength(0.0, 0.0, 'Angstrom')>]]
Got:
    [[<Time(<Time object: scale='utc' format='isot' value=2011-08-08T01:30:04.000>, <Time object: scale='utc' format='isot' value=2011-08-10T00:00:04.000>, None)>, <Source('SDO')>, <Instrument('HMI')>, <Wavelength(6172.999999999998, 6172.999999999998, 'Angstrom')>]]
```
unrelated to this PR.

ayshih:
Okay, rebased again, and cleaned up the commit history a bit too

PR #3346, opened 2019-09-11
Small Improvements to ImageAnimatorWCS

Cadair:


sunpy/visualization/animator/image.py
```diff
@@ -140,6 +138,8 @@ class ImageAnimatorWCS(ImageAnimator):
     ----------
```
dstansby: Does this not get inherited from the base class?

sunpy/visualization/animator/image.py
```diff
@@ -160,10 +160,10 @@ class ImageAnimatorWCS(ImageAnimator):
     """
     def __init__(self, data, wcs=None, image_axes=[-1, -2], unit_x_axis=None, unit_y_axis=None,
                  axis_ranges=None, **kwargs):
-        if not isinstance(wcs, (astropy.wcs.WCS, BaseLowLevelWCS)):
-            raise ValueError("wcs data should be provided.")
+        if not isinstance(wcs, BaseLowLevelWCS):
```
dstansby: Why this change?

Cadair: `astropy.wcs.WCS` inherits `BaseLowLevelWCS` so it was redundant.

sunpy/visualization/animator/image.py
Cadair: The docstring? I would assume so, but we are adding a `wcs` parameter (that was previously not included in the docstring).

Cadair:
I need to double check this works as intended with astropy 3.1.

Cadair:
@dstansby I checked the rendered docs and noticed more docstring issues.

Cadair:
This does indeed work on Astropy 3.1.

sunpy/visualization/animator/image.py
```diff
@@ -140,13 +138,15 @@ class ImageAnimatorWCS(ImageAnimator):
...
     image_axes: `list`, optional
         A list of the axes order that make up the image.
-    unit_x_axis: `astropy.units.Unit`
+    unit_x_axis: `astropy.units.Unit`, optional
         The unit of X axis.
-    unit_y_axis: `astropy.units.Unit`
+    unit_y_axis: `astropy.units.Unit`, optional
         The unit of Y axis.
-    axis_ranges: `list` of physical coordinates for the `numpy.ndarray`, optional
+    axis_ranges: `list`, optional
```
DanRyanIrish: Is there not a advantage of also giving the required type of the elements of the list?

sunpy/visualization/animator/image.py
```diff
@@ -140,13 +138,15 @@ class ImageAnimatorWCS(ImageAnimator):
...
         Defaults to `None` and array indices will be used for all axes.
         The `list` should contain one element for each axis of the `numpy.ndarray`.
```
DanRyanIrish: Replace "`numpy.ndaray`" with "input data array".

sunpy/visualization/animator/image.py
Cadair: That is in the comment for the parameter. The numpy doc format is pretty clear on what format this line should take, and a lot of sphinx themes are built around it being short.

sunpy/visualization/animator/image.py
Cadair:
```suggestion
        The `list` should contain one element for each axis of the input data array.
```

Cadair:
The coverage upload is broken, am investigating.

PR #3351, opened 2019-09-12
Fix construction of HG frames to be 3D

### LINKED ISSUE #3350: Creating a HeliographicStonyhurst frame with SkyCoord does not always return a 3D frame
### Description
```python
import astropy.units as u

SkyCoord([1,2,3,4], [1,3,4,5], unit=u.deg, frame=frames.HeliographicStonyhurst(obstime="2011-06-07T06:32:11.000"))
```
### Expected behaviour
The returned frame should be 3D with the rsun value as the distance coordinate.

### Actual behaviour

The frame is has a `UnitSpherical` representation:
```
<SkyCoord (HeliographicStonyhurst: obstime=2011-06-07T06:32:11.000): (lon, lat) in deg
    [(1., 1.), (2., 3.), (3., 4.), (4., 5.)]>
```

Cadair:
Fixes #3350

### DIFF
--- a/changelog/3351.bugfix.rst
+++ b/changelog/3351.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed bug where Heliographic frames were not constructed as 3D as intended.
+Fixed situations where 2D coordinates provided to `~sunpy.coordinates.frames.HeliographicStonyhurst` and `~sunpy.coordinates.frames.HeliographicCarrington` were not converted to 3D as intended.  Furthermore, the stored data will always be the post-conversion, 3D version.
--- a/changelog/3351.bugfix.rst
+++ b/changelog/3351.bugfix.rst
@@ -1,0 +1,2 @@
+Ensure that Heliographic frames are always constructed as 3D even when
+constructed through `~astropy.coordinates.SkyCoord`.
--- a/sunpy/coordinates/frames.py
+++ b/sunpy/coordinates/frames.py
@@ -221,8 +221,10 @@ class HeliographicStonyhurst(SunPyBaseCoordinateFrame):
         super().__init__(*args, **kwargs)
 
         # Make 3D if specified as 2D
-        # If representation was explicitly passed, do not change the rep.
-        if not _rep_kwarg:
+        # If representation was explicitly passed, do not change the rep, unless
+        # it's specified as SphericalRepresentation, in which case we should
+        # upgrade it to that rep if we only have UnitSphericalRepresentation
+        if not _rep_kwarg or _rep_kwarg is SphericalRepresentation:
             if isinstance(self._data, UnitSphericalRepresentation):
                 self._data = self.spherical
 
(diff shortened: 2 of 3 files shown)


ayshih:
While this PR does fix the issue, I suspect that there might be a better/cleaner way to fix it.  Or I could just be imagining things.  I'd like to do some tinkering before letting this through, unless a bugfix release is just around the corner.

Cadair:
@ayshih It's not the end of the world if this floats for a little while, but it is breaking me messing around with #3353. 

I thought a decent amount about this, and I am pretty sure this is the best way. When you construct the frame with an explict `SphericalRepresentation` it seems to me the only logical thing to do is to throw it through our code to add R.

ayshih:
I just PR-ed to your branch.  After poking around a bunch, I've decided that the whole approach to keyword checking and `represent_as` was unnecessarily complicated.  (The `represent_as` stuff is [my fault](https://github.com/sunpy/sunpy/commit/ec89603db6065af12784d1e31e9e41967e5f256b).)  Instead it will now always upgrade to 3D, as already promised in the docstring.  If a user chooses to specify a `representation_type` of `UnitSphericalRepresentation`, that's what they'll be shown, with component attributes to match, but the underlying data will be 3D.

Cadair:
Do the changes to the `represent_as` function have any effect here? Is this still just a bug fix?

ayshih:
In terms of user-facing API, the change here is what `.data` returns.  Previously, if a user specifically defined the `representation_type`, the class would try to preserve the underlying data and its representation, even if it was 2D rather than 3D.  The `.represent_as()` stuff was there so that the class was still able to work with the underlying data as if it were 3D.  Thus, previously, `.data` could still return a 2D representation, although this behavior was entirely undocumented.  (The bug identified in the corresponding issue was because the SkyCoord machinery can specify a `representation_type` that is different from the type of the representation.)  I think this still qualifies as a bug fix because these shenanigans were in violation of the docstring, which says that the data will always be converted to 3D.

Cadair:
Could you perhaps suggest some appropriate wording for the changelog just to make that obvious to everyone? (or I can and then you can comment on it?)

ayshih:
Here's a wordier version:
```
Fixed situations where 2D coordinates provided to `~sunpy.coordinates.frames.HeliographicStonyhurst` and `~sunpy.coordinates.frames.HeliographicCarrington` were not converted to 3D as intended.  Furthermore, the stored data will always be the post-conversion, 3D version.
```

changelog/3351.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed bug where Heliographic frames were not constructed as 3D as intended.
```
Cadair:
```suggestion
Fixed situations where 2D coordinates provided to `~sunpy.coordinates.frames.HeliographicStonyhurst` and `~sunpy.coordinates.frames.HeliographicCarrington` were not converted to 3D as intended.  Furthermore, the stored data will always be the post-conversion, 3D version.
```

Cadair:
Point of order, it seems that this backport failed because of the merge commit.

PR #3354, opened 2019-09-15
Ran a bunch of tools on the codebase

nabobalis:
i can remove any strong formatting (most of it) from that PR
and just autopep8 and then add a CI job to test it
then formatting style can be moved to another time
ill leave in the rest and change the precommits


ayshih:
Yikes!  Some comments from a quick glance at `coordinates` files:

- Well, this destroys `git blame`...
- I'm not a fan of adding spaces around the "*" when defining a constant `Quantity`, but I don't know if there's a way around that with auto-formatting
- Some `isort` output oddities:
  - Mistakenly treats `ICRS` (and similar frames) as a constant because it is all capitals, so it puts it at the start of import lists
  - `try`...`except` code in the import area for third-party stuff gets moved to after the `sunpy` imports
- `black` sure loves newlines, and the (lack of) indention for continued lines is a bit confusing

Cadair:
We had a little chat about this in the matrix channel, and we agreed that black should probably be delayed and discussed in a SEP.

I am keen on using isort, so would be interested in looking at fixing the ICRS constant import thing. The try except thing is probably deliberate, but I can see that it's a change in how we currently operate.

nabobalis:
Since this was generated a lot of negative comments, I will just leave it for someone else in the future.

PR #3356, opened 2019-09-16
Fix off by one error in header helper

### LINKED ISSUE #3355: make_fitswcs_header has an off by one error in the default crpix calculation
This needs double checking but it looks like we  are +1 off in the default array shape halving

Cadair:
Fixes #3355

### DIFF
--- a/changelog/3356.bugfix.rst
+++ b/changelog/3356.bugfix.rst
@@ -1,3 +1,3 @@
-Fix off by one error in `sunpy.map.make_fiteswcs_header` where when using the
+Fix off by one error in `sunpy.map.make_fitswcs_header` where when using the
 default ``reference_pixel=None`` keyword argument the pixel coordinate of the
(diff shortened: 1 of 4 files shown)


astrofrog:
Seems reasonable to me!

Cadair:
Online test fails are unrelated.

changelog/3356.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Fix off by one error in `sunpy.map.make_fiteswcs_header` where when using the
```
wtbarnes: `sunpy.map.make_fiteswcs_header` should be `sunpy.map.make_fitswcs_header`

Cadair:
```suggestion
Fix off by one error in `sunpy.map.make_fitswcs_header` where when using the
```

PR #3361, opened 2019-09-17
Add remove_column() method to TimeSeries

dstansby:


sunpy/timeseries/tests/test_timeseriesbase.py
```diff
@@ -566,6 +566,26 @@ def test_add_column_from_quantity(eve_test_ts, add_column_from_quantity_ts,
...
+    assert len(removed.columns) == len(removed.units)
+
+    # Check data updated correctly
+    assert len(removed.columns) == removed.data.shape[1]
+
+    # Check that removing a non-existant column errors
+    with pytest.raises(ValueError):
+        eve_test_ts.remove_column('random column name')
+
+
```
Cadair: bonus blank lines?

dstansby:
> Looks useful and unoffensive to me. Is this API pandas like or astropy Table like or both or neither?

It's [astropy like API](https://docs.astropy.org/en/stable/api/astropy.table.Table.html#astropy.table.Table.remove_column), and definitely not pandas-like,

dstansby:
The only difference is `astropy` does everything in-place, wheras here I've copied `add_column` in SunPy and made `remove_column` return a new `TimeSeries`.

Cadair:
I am generally :+1: on more following the table api than pandas (mostly personal preference) so that's great by me.

Issue #3337, opened 2019-09-04
Data for GOES assumes it knows what satellites are available based on date range


Cadair:
Currently in `XRSClient` we hard code the operational dates of the goes satellite and use the highest number we can for they query. For some days in the operational range of the satellite the files might not exist on the server. We should be smart about how we choose satellite number.

aringlis:
So do you suggest that the client should be aware of all the operational satellites that have data available for the chosen date(s), and fall back to another satellite if data cannot be found? 

Related to this, it is possible to search using the ```a.goes.SatelliteNumber``` attribute to manually specify which satellite you want, but it does not seem to work as intended, e.g.:

```In [68]: result = Fido.search(a.Time('2010/06/01','2010/06/02'),a.Instrument('XRS'),a.goes.SatelliteNumber(15))                      

In [69]: result                                                                                                                      
Out[69]: 
<sunpy.net.fido_factory.UnifiedResponse object at 0x7f8c38647eb8>
Results from 1 Provider:

2 Results from the XRSClient:
     Start Time           End Time      Source Instrument Wavelength
       str19               str19         str4     str4       str3   
------------------- ------------------- ------ ---------- ----------
2010-06-01 00:00:00 2010-06-01 23:59:59   nasa       goes        nan
2010-06-02 00:00:00 2010-06-02 23:59:59   nasa       goes        nan ```

The above should not work, as it is before GOES-15 data are available. But, the search result gives the impression that results are found. However:

```In [70]: Fido.fetch(result)                                                                                                          
Files Downloaded:   0%|                                                                                      | 0/2 [00:00<?, ?file/s]
Out[70]: 
<parfive.results.Results object at 0x7f8c3863aeb8>
[]
Errors:
(error(filepath_partial=<function Downloader.enqueue_file.<locals>.filepath at 0x7f8c4c7467b8>, url='https://umbra.nascom.nasa.gov/goes/fits/2010/go1520100602.fits', exception=FailedDownload()```

I truncated the error message, but you can see that when you try to actually download the GOES data, it does not exist. The same search and retrieval works correctly when GOES-14 is specified, as it should.

aringlis:
The hardcoded dates for the GOES satellite operations also do not account for all the times that data are available. For example, 2016 and 2017 both have data available from GOES-13 and GOES-14 as well as GOES-15.

IIRC the dates are sourced from NOAA and represent in some way which spacecraft is 'primary'.

[...]

PR #3367, opened 2019-09-24
Scraper fido updates

### LINKED ISSUE #3349: Dataretriever data sources should all use Scraper
### Description 

Some of the Fido source clients do not use `Scraper` and just build a url string based upon the pattern that is usually used, hence a `Fido.search` using these clients (that don't use `Scraper`) will always return a result even though they might necessarily exist, so when you do a Fido.fetch on the result of the search, an error is returned. 

This is true for GOES XRS, LYRA and the RHESSI clients, and follows on from #3337 error shown by @aringlis 

For example with LYRAClient (which is same for when GOES has no data):
```
>>> res = Fido.search(a.Time('2009/06/11 00:00', '2009/06/11 01:00'), a.Instrument('lyra')) 
>>> res 
<sunpy.net.fido_factory.UnifiedResponse object at 0x126e50c88>
Results from 1 Provider:

1 Results from the LYRAClient:
       str19               str19         str6     str4       str3   
2009-06-11 00:00:00 2009-06-11 01:00:00 Proba2       lyra        nan
>>> Fido.fetch(res)
Files Downloaded: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00,  5.17file/s]
<parfive.results.Results object at 0x126e50eb8>
[]
Errors:
(error(filepath_partial=<function Downloader.enqueue_file.<locals>.filepath at 0x12613b2f0>, url='http://proba2.oma.be/lyra/data/bsd/2009/06/11/lyra_20090611-000000_lev2_std.fits', response=<ClientResponse(http://proba2.oma.be/lyra/data/bsd/2009/06/11/lyra_20090611-000000_lev2_std.fits) [404 Not Found]>
<CIMultiDictProxy('Server': 'nginx/0.7.65', 'Date': 'Wed, 11 Sep 2019 15:18:38 GMT', 'Content-Type': 'text/html', 'Transfer-Encoding': 'chunked', 'Connection': 'keep-alive', 'Content-Encoding': 'gzip')>
))
```
Hence it looks like the file exists but theres then an error opening it. With `Scraper` nothing would be returned in the `Fido.search`. 

Also it's weird that the Files Downloaded 100% is output from parfive @Cadair ?

hayesla:
### Description
Updating both the dataretriever source files of for GOES XRS and lyra to use Scraper 

Fixes #3349 and also fixes some of what was talked about in #3337


Cadair:
I think a lot of the tests for those clients were offline, and they are now online, so I would guess that that is one of the things that's broken the tests here (without actually looking).

changelog/3367.bugfix.rst
```diff
@@ -0,0 +1 @@
+Updating both GOES XRS and LYRA dataretriver cleints to use Scraper
```
ehsteve: misspelled clients

hayesla:
arg another bug I think in Scraper - 

For example:
```
>>> from sunpy.time import TimeRange
>>> from sunpy.util.scraper import Scraper

>>> timerange = TimeRange(['2011-06-07 04:00', '2011-06-07 12:00'])
>>> goes_pattern = ("https://umbra.nascom.nasa.gov/goes/fits/%Y/go{satellitenumber}%Y%m%d.fits")
>>> goes_files = Scraper(goes_pattern, satellitenumber=15)
>>> urls = goes_files.filelist(timerange)
>>> urls
[]

```
The issue here is that the files have the time pattern of the date that doesn't include hours, minutes etc. So in scraper in the `_extractDateUrl` function it adds `00:00:00` to the date from the url and then when it checks if the date from the url > timerange.start and date from url < timerange.end , this condition is never met (i.e from the example above the time from `_extractDateURL` returns '2011-06-07 00:00:00' so its a problem! 

Should I raise this as an issue? The url it should return is available

hayesla:
ping @Cadair @ehsteve

hayesla:
or I could just change it in the goes file such that the start time is changed to %y%m%d ? probably best for this PR - but something we should be aware of for scraper!

hayesla:
this now works locally for me - but its having issues here with generating the time-series example?

Also how do you resolve these conflicts? Git is confusing me ..... 🤕

wtbarnes:
It looks like even the offline tests are failing because there are some tests in `net/tests/test_fido.py` that are not marked as `remote_data`.

changelog/3367.bugfix.rst
```diff
@@ -0,0 +1 @@
+Updating both GOES XRS and LYRA dataretriver clients to use Scraper
```
Cadair: Is this actually fixing a bug? Does it have any external facing changes at all?

nabobalis: I mean it might not be user facing but the change could in theory fix something we've overlooked.

hayesla: it doesn't have external facing changes - other than before the Fido search would have returned a query result that may not actually exist so it fails on download e.g. #3349 and #3337

Cadair: If it's not actually fixing a bug I would probably classify it as "internal". I have a feeling it might be fixing bugs where some files were missing on the server though. If it is a bug fix I think the changelog entry should describe the bug it's fixing.

Cadair: It returning results that aren't on the server counts as a bug in my mind, can you edit the changelog to describe that @hayesla ? :smile:

changelog/3367.bugfix.rst
```diff
@@ -0,0 +1 @@
+Updating both GOES XRS and LYRA dataretriver clients to use `sunpy.util.scraper.Scraper` as to make sure that files are actually on servers being queried
```
nabobalis:
```suggestion
Updating both GOES XRS and LYRA dataretriever clients to use `~sunpy.util.scraper.Scraper`, to make sure that files are actually on the servers being queried.
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -118,43 +118,26 @@ def _get_url_for_timerange(self, timerange, **kwargs):
...
-        satellitenumber : int
-            GOES satellite number (default = 15)
-        data_type : str
-            Data type to return for the particular GOES satellite. Supported
-            types depend on the satellite number specified. (default = xrs_2s)
+        timerange : `~sunpy.time.TimeRange`
+
+        Returns
+        -------
+        list :
```
nabobalis:
```suggestion
        `list`
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -118,43 +118,26 @@ def _get_url_for_timerange(self, timerange, **kwargs):
...
         ----------
-        timerange: sunpy.time.TimeRange
-            time range for which data is to be downloaded.
-        satellitenumber : int
-            GOES satellite number (default = 15)
-        data_type : str
-            Data type to return for the particular GOES satellite. Supported
-            types depend on the satellite number specified. (default = xrs_2s)
+        timerange : `~sunpy.time.TimeRange`
+
```
nabobalis:
```suggestion
        The time range you want files for.
         
```

sunpy/net/dataretriever/sources/lyra.py
```diff
@@ -36,44 +36,21 @@ class LYRAClient(GenericClient):
     """
     def _get_url_for_timerange(self, timerange, **kwargs):
         """
-        Returns list of URLS corresponding to value of input timerange.
-
+        Return URL(s) for corresponding timerange.
         Parameters
```
nabobalis:
```suggestion

        Parameters
```

sunpy/net/dataretriever/sources/lyra.py
```diff
@@ -36,44 +36,21 @@ class LYRAClient(GenericClient):
...
         ----------
-        timerange: sunpy.time.TimeRange
-            time range for which data is to be downloaded.
-
+        timerange : `~sunpy.time.TimeRange`
```
nabobalis:
```suggestion
        timerange : `~sunpy.time.TimeRange`
            The time range you want files for.
```

sunpy/net/dataretriever/sources/lyra.py
```diff
@@ -36,44 +36,21 @@ class LYRAClient(GenericClient):
...
-            list of URLs corresponding to the requested time range
-        """
-        days = timerange.get_dates()
-        urls = []
-        for day in days:
-            urls.append(self._get_url_for_date(day, **kwargs))
-        return urls
-
-    def _get_url_for_date(self, date, **kwargs):
+        list :
```
nabobalis:
```suggestion
        `list`
```

sunpy/net/dataretriever/tests/test_lyra_ud.py
```diff
@@ -36,12 +37,7 @@ def test_get_url_for_time_range(timerange, url_start, url_end):
     assert urls[-1] == url_end
 
 
-def test_get_url_for_date():
-    url = LCClient._get_url_for_date(parse_time((2013, 2, 13)))
-    assert url == 'http://proba2.oma.be/lyra/data/bsd/2013/02/13/lyra_20130213-000000_lev2_std.fits'
```
nabobalis: Stupid question: Why was this test removed?

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -118,43 +118,26 @@ def _get_url_for_timerange(self, timerange, **kwargs):
...
-            url = base_url + regex.format(sat=satellitenumber)
-            result.append(url)
-        return result
+        timerange = TimeRange(timerange.start.strftime('%Y-%m-%d'), timerange.end)
+        if timerange.start < parse_time("1999/01/15"):
+            goes_pattern = ("https://umbra.nascom.nasa.gov/goes/fits/"
+                            "%Y/go{satellitenumber:02d}%y%m%d.fits")
+        else:
+            goes_pattern = ("https://umbra.nascom.nasa.gov/goes/fits/"
+                            "%Y/go{satellitenumber}%Y%m%d.fits")
```
dpshelio: Yay! The scraper being used :hugs: 

Here I'd suggest the following:

```python
if timerange.start < parse_time("1999/01/15"):
    goes_file = "%Y/go{satellitenumber:02d}%y%m%d.fits"
else:
    goes_file = "%Y/go{satellitenumber}%Y%m%d.fits"

goes_pattern = f"https://umbra.nascom.nasa.gov/goes/fits/{goes_file}"
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -118,43 +118,26 @@ def _get_url_for_timerange(self, timerange, **kwargs):
...
+
+        satellitenumber = kwargs.get("satellitenumber", self._get_goes_sat_num(timerange.start))
+        goes_files = Scraper(goes_pattern, satellitenumber=satellitenumber)
+
+        urls = goes_files.filelist(timerange)
+        return urls
```
dpshelio:
```python
return goes_files.filelist(timerange)
```
No need to create the intermediate variable

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -118,43 +118,26 @@ def _get_url_for_timerange(self, timerange, **kwargs):
...
-            else:
-                regex += date.strftime("%Y%m%d") + ".fits"
-            satellitenumber = kwargs.get(
-                "satellitenumber", self._get_goes_sat_num(date)
-            )
-            url = base_url + regex.format(sat=satellitenumber)
-            result.append(url)
-        return result
+        timerange = TimeRange(timerange.start.strftime('%Y-%m-%d'), timerange.end)
+        if timerange.start < parse_time("1999/01/15"):
```
ayshih: You check only `timerange.start` here, so it's possible that `timerange.end` will be after this date, so you may need to split the time range in two.

sunpy/net/dataretriever/tests/test_lyra_ud.py
hayesla: The function `_get_url_for_date` no longer exists - it used to be a separate function that would return a url for a particular date and was called within `_get_url_for_timerange`  but now scraper takes care of it and its all done now inside `_get_url_for_timerange`

sunpy/net/dataretriever/sources/goes.py
hayesla: I've added functionality now to take this into account 👍

sunpy/net/dataretriever/sources/lyra.py
```diff
@@ -36,44 +36,24 @@ class LYRAClient(GenericClient):
     """
     def _get_url_for_timerange(self, timerange, **kwargs):
         """
-        Returns list of URLS corresponding to value of input timerange.
-
+        Return URL(s) for corresponding timerange.
+
```
Cadair:
```suggestion

```

Cadair:
@hayesla There are a couple of suggestions which need applying, then I think it can be merged :smile:

nabobalis:
I just opened the files with my editor and it whitespace things and updated the changelog.

nabobalis:
All other comments had been addressed so I plan to merge this as soon as the CI passes.

nabobalis:
I can not replicate that doc failure locally. Both the example and the helioviewer tests pass for me.

PR #3371, opened 2019-09-25
Fix lat lon makefitswcs

### LINKED ISSUE #3370: Lat and Lon the wrong way around when put into fitshe`sunpy.map.make_fitswcs_header`
# Description
I think its a bug in the `make_fitswcs_header` here https://github.com/sunpy/sunpy/blob/master/sunpy/map/header_helper.py#L134

The x and y values of the input SkyCoord (i.e. lon and lat) are put into the fits header as 
crval1 = lat and crval2 = lon - shouldn't it be the other way around? 

For example:
```
import sunpy.map
from astropy import units


coord1 = SkyCoord(1000*u.arcsec, 0*u.arcsec, observer = 'earth', 
                  obstime = parse_time('now'), frame = frames.Helioprojective)
data = data1 = np.zeros((4096, 4096))
map_header = sunpy.map.make_fitswcs_header(data, coord1)
my_map = sunpy.map.Map(data, map_header)
my_map.peek(draw_limb=True)

```

image

Given that the reference coordinate of the center pixel should be (1000, 0) rather than (0, 1000) as shown. 

This is an easy fix - does this make sense or am I missing something?

hayesla:
Fixes #3370 👍


Cadair:
@hayesla can you rebase out the other commits that change net? Also can you add a test to make sure we get the lon / lat consistent with ctype please?

hayesla:
woops sorry didn't mean that to also have those net changes

hayesla:
> woops sorry didn't mean that to also have those net changes

git hurts my brain sometimes🤦‍♂

Cadair:
That's a healthy thing :stuck_out_tongue: I will squash this to tidy up the history when I merge it, don't worry about it.

Just need tests :smile:

hayesla:
> That's a healthy thing 😛 I will squash this to tidy up the history when I merge it, don't worry about it.
> 
> Just need tests 😄

hrmm I'm kinda confused how these tests should look? - the way the frame_to_wcs functionality is working is that it will return ctype1, ctype2 as HGLN, HGLT  or (CRLN, CRLT) respectively, so now the way the `make_fitswcs_header` is working it should always be the case now that crval1, crval2 will be lon and lat right?

Cadair:
yeah just write a regression test which makes sure that the lon value is in crval1 and that crtype one contains `LN` and the same for lat.

Cadair:
(this is definitely on me as I merged this PR, but this should have been rebased on merge because the lively git history really confused the backport bot)

PR #3376, opened 2019-09-30
Put Matplotlib imports inside TimeSeries methods

### LINKED ISSUE #3254: Import matplotlib only in plot methods?
Currently `from sunpy import timeseries` requires a working Matplotlib installation; it would be nice if it was possible to use the `TimeSeries` object without needing a full on Matplotlib installation.

Relevant line:
```python
  File "/Users/dstansby/miniconda3/envs/helios-alphas/lib/python3.7/site-packages/sunpy/timeseries/timeseriesbase.py", line 15, in <module>
```

dstansby:
Fixes #3254

### DIFF
--- a/changelog/3376.breaking.rst
+++ b/changelog/3376.breaking.rst
@@ -1,0 +1,2 @@
+Importing `sunpy.timeseries.timeseriesbase` no longer automatically imports
+Matplotlib.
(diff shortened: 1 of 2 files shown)


hayesla:
I think just a changelog is needed 👍

PR #3379, opened 2019-10-01
Add missing import in the GenericTimeSeries docstring

xuanxu:
Otherwise running the example code fails with a `NameError: name 'u' is not defined`

### DIFF
--- a/sunpy/timeseries/timeseriesbase.py
+++ b/sunpy/timeseries/timeseriesbase.py
@@ -59,6 +59,7 @@ class GenericTimeSeries:
     >>> from sunpy.timeseries import TimeSeries
     >>> from sunpy.time import parse_time
     >>> from astropy.time import TimeDelta
+    >>> import astropy.units as u
     >>> import numpy as np
     >>> import pandas as pd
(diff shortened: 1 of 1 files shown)


ayshih:
@Cadair @nabobalis Why wasn't this caught by doctest?

xuanxu:
> Why wasn't this caught by doctest?

@ayshih There's a SKIP flag: https://github.com/sunpy/sunpy/blob/master/sunpy/timeseries/timeseriesbase.py#L68 
(Maybe because the output is too verbose, more noise than signal for a docstring)

ayshih:
The `peek()` call is skipped (because it's graphical), but the import error should occur four lines earlier.  This suggests to me that this entire docstring is being skipped somehow.

nabobalis:
I'll have to run this locally to check but as far as I was aware, none of the docstrings should be skipped.

nabobalis:
I would speculate that the in-code examples has access to the imports in the code file.

ayshih:
Ah, yes, that's exactly why.  Each doctest has a copy of the module's globals ([ref](https://docs.python.org/3/library/doctest.html#what-s-the-execution-context)).  I wasn't aware of that.

PR #3388, opened 2019-10-02
Allow observer-based coordinate frames to have no defined observer

### LINKED ISSUE #3386: Allow observer-based coordinate frames to have no defined observer
This issue is inspired by PR #3118, specifically my [suggested alternative](https://github.com/sunpy/sunpy/pull/3118#issuecomment-519636169).  In a nutshell, the coordinate machinery would no longer have a default observer (e.g., Earth) for observer-based frames (i.e., Heliocentric and Helioprojective), but instead allow for the observer to be `None`.  Users would be able to use such frames up until the point where an observer needs to be defined (e.g., transformations).

ayshih:
Observer-based coordinate frames can have no defined observer, and will no longer have a default observer of Earth.  (Maps still have a default observer of Earth, because that is separate code.)  All non-trivial transformations will raise an error.  This is implemented as a change without any deprecation period.

Closes #3386


sunpy/coordinates/transformations.py
```diff
@@ -127,6 +127,14 @@ def _observers_are_equal(obs_1, obs_2, string_ok=False):
             obs_1.obstime == obs_2.obstime)
 
 
+def _check_observer_defined(frame):
+    if frame.observer is None:
+        raise ConvertError("This transformation cannot be performed with observer=None.")
```
Cadair: Given you call this function on input and output frames in the transforms would it be wise to specify which frame is missing the observer here?

sunpy/coordinates/transformations.py
```diff
@@ -127,6 +127,14 @@ def _observers_are_equal(obs_1, obs_2, string_ok=False):
...
+    elif isinstance(frame.observer, str):
+        raise ConvertError("This transformation cannot be performed without a specified obstime "
```
Cadair: Same here about telling the user what frame we are talking about.

sunpy/coordinates/transformations.py
ayshih: Done

sunpy/coordinates/transformations.py
ayshih: Done

sunpy/coordinates/frameattributes.py
```diff
@@ -127,11 +127,8 @@ def _convert_string_to_coord(self, out, obstime):
     def __get__(self, instance, frame_cls=None):
         # If instance is None then we can't get obstime so it doesn't matter.
         if instance is not None:
-            # Get observer if the instance has one, or the default.
-            observer = getattr(instance, '_' + self.name, self.default)
-
-            # We have an instance of a frame, so get obstime
-            obstime = getattr(instance, 'obstime', None)
+            observer = getattr(instance, '_' + self.name)
+            obstime = getattr(instance, 'obstime', None)  # TODO: Why is this `None` needed?
```
ayshih: I'm not sure why this line is getting called on objects without `obstime`, but there can be errors during transformations of the `None` fallback is not present.  Curiously, it's not needed for the `observer` line just above it.   I've marked it with a comment for future investigation (beyond the scope of this PR).

ayshih:
I tried to implement deprecation, but I feel that it just doesn't work when removing a default.  You can't have both the deprecated approach where `observer=None` means a default observer and the new approach where `observer=None` means no observer.  That is, a user wouldn't be able to use the new functionality.

One possibility is that I could emit a warning the first time that a frame has `observer=None` so that a user can be aware that the default behavior has changed.  What do you think?

Cadair:
Does throwing a warning on construct buy us much? It's going to error as soon as they do anything with the frame anyway? (i.e. it's not like they aren't going to notice!)

ayshih:
Heh, no worries, I was already planning on adding those tests

Cadair:
@ayshih I spoke to Nadia about the asdf schema version bump and she recommends we don't bump the schema version because it's a backwards compatible change.

ayshih:
Okay, test added

Cadair:
@sunpy/sunpy-maintainers This needs a second review please.

sunpy/coordinates/tests/test_transformations.py
```diff
@@ -667,3 +669,32 @@ def test_gei_gei():
     new = old.transform_to(GeocentricEarthEquatorial(equinox=t, obstime=t)).cartesian
 
     assert_quantity_allclose(new.xyz, gei_d.xyz)
+
+
+def test_no_observer():
```
dstansby: I think it is worth testing here two different coordinate objects initialised with exactly the same parameters (but that are distinct objects) cannot be transformed, ie.:

```python
frame1 = Heliocentric(0*u.km, 0*u.km, 0*u.km, observer=None)
frame2 = Heliocentric(0*u.km, 0*u.km, 0*u.km, observer=None)
frame1.transform_to(frame2)   # This should fail
```

sunpy/coordinates/transformations.py
```diff
@@ -118,15 +118,28 @@ def _observers_are_equal(obs_1, obs_2, string_ok=False):
     if string_ok:
         if obs_1 == obs_2:
             return True
-    if not (isinstance(obs_1, BaseCoordinateFrame) and isinstance(obs_2, BaseCoordinateFrame)):
-        raise ValueError("To compare two observers, both must be instances of BaseCoordinateFrame. "
-                         "Cannot compare two observers {} and {}.".format(obs_1, obs_2))
+
+    if not isinstance(obs_1, BaseCoordinateFrame):
+        raise ConvertError(f"Source observer {obs_1} is not fully specified.")
```
dstansby: I think the original message was clearer here, ie. mentioning explicitly that `obs_1` must be an instance of `BaseCoordinateFrame`.

sunpy/coordinates/tests/test_transformations.py
ayshih: As discussed on the telecon, this will actually succeed in the transformation machinery.

sunpy/coordinates/transformations.py
ayshih: I've made the error messages more helpful

ayshih:
Ready for review again

Cadair:
Thanks @ayshih another awesome coordinates PR :grin:

PR #3389, opened 2019-10-02
Rename `calculate_distance()` to `make_3d()`

### LINKED ISSUE #3222: Change the name of `Helioprojective.calculate_distance()`
`Helioprojective.calculate_distance()` is used to convert a 2D HPC coordinate into a 3D HPC coordinate by filling in the `distance` attribute with the distance from the observer to the surface of the Sun.  I have issues with the name of the method:

- The name suggests that it is used to calculate the distance to something user-supplied (e.g., a la `SkyCoord.separation_3d()`).
- A user might expect it to return some distance value, not a new `Helioprojective` frame.

I think a better name for this method would be along the lines of `make_3d()` or `convert_to_3d()`.  Discuss.

ayshih:
See #3222.  For `Helioprojective`, I've renamed `calculate_distance()` to `make_3d()`, with deprecation.

Closes #3222


Cadair:
@ayshih Figure test fail is real, looks like it's calling the deprecated function.

ayshih:
Heh, it figures that I'd neglect to change to the renamed function in the one place where it's actually used (outside of tests)!

I've updated the deprecation approach so that it shares the method and docstring (rather than having the old method call the new one).

Cadair:
Test fails seem unrelated. @sunpy/sunpy-maintainers can this get a second review please.

PR #3396, opened 2019-10-07
Resolve  deprecation warnings for test_sysinfo.py

guntbert:
xref #2886
using the module `distro` for determining the linux_distribution instead of `platform`.

### Description
Goal: resolve all warnings when running tests for  `sunpy/util/tests/test_sysinfo.py `
Currently this resolves one deprecation warning.

I have a few questions:
- this needs to import the module `distro` - where should this be documented?
- further deprecation warnings concern 
  - `lib/python3.7/site-packages/html5lib/_trie/_base.py`
  - `lib/python3.7/site-packages/zeep/loader.py`

should I change those files here as well?

### DIFF
--- a/sunpy/util/sysinfo.py
+++ b/sunpy/util/sysinfo.py
@@ -4,6 +4,8 @@ This module provides functions to retrieve system information.
 import datetime
 import platform
 
+from distro import linux_distribution
+
 __all__ = ['get_sys_dict', 'system_info']
 
 
@@ -118,7 +120,7 @@ def system_info():
         print('{} : {}'.format(sys_info, sys_prop[sys_info]))
 
     if sys_prop['System'] == "Linux":
-        distro = " ".join(platform.linux_distribution())
+        distro = " ".join(linux_distribution())
         print("OS: {} (Linux {} {})".format(distro, platform.release(), sys_prop['Processor']))
     elif sys_prop['System'] == "Darwin":
         print("OS: Mac OS X {} ({})".format(platform.mac_ver()[0], sys_prop['Processor']))
--- a/sunpy/util/sysinfo.py
+++ b/sunpy/util/sysinfo.py
@@ -4,7 +4,7 @@ This module provides functions to retrieve system information.
 import datetime
 import platform
 
-from distro import linux_distribution
+from sunpy.extern.distro import linux_distribution
 
 __all__ = ['get_sys_dict', 'system_info']
 
--- a/changelog/3396.trivial.rst
+++ b/changelog/3396.trivial.rst
@@ -1,0 +1,1 @@
+Copy the library `distro` into `sunpy/extern`: replaces the deprecated `platform/linux_distribution`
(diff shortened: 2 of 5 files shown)


nabobalis:
Is `distro` a separate PyPi package?

guntbert:
I installed it via conda
   distro-1.4.0               |             py_0          19 KB  conda-forge
according to my research it cannot be avoided (project `salt` uses it as well.

nabobalis:
Hmm, I would suggest that we add distro the same we we have `appdirs` in SunPy, under the `extern` folder and we import it from there instead of adding as an extra dependency.

Could you do that?

guntbert:
I will try, looks easy enough :-)
Immediately I have three questions:
- How do you keep this library up to date?
- Anything special for importing (like giving the path,...)?
- Do I understand correctly that I should uninstall `distro` from my environment?

nabobalis:
> I will try, looks easy enough :-)
> Immediately I have three questions:
> 
> * How do you keep this library up to date?
> * Anything special for importing (like giving the path,...)?
> * Do I understand correctly that I should uninstall `distro` from my environment?

* Manually update it when we feel like it.
* import should be along the lines of `sunpy.extern.distro` 
* No, they should not conflict.

guntbert:
Should be complete - I added a changelog entry (I guess the label doesn't apply since I worked outside of the tests too)

nabobalis:
Thanks @guntbert, it all looks fine to me. We will need two reviews before we can merge this but I see no problems.

guntbert:
Thx again for your guidance - you make it a pleasure to contribute :bowing_man:

sunpy/util/sysinfo.py
```diff
@@ -118,7 +120,7 @@ def system_info():
         print('{} : {}'.format(sys_info, sys_prop[sys_info]))
 
     if sys_prop['System'] == "Linux":
-        distro = " ".join(platform.linux_distribution())
+        distro = " ".join(linux_distribution())
```
dstansby: Is it really worth ~1500 extra lines of code just to make it possible to distinguish between linux distros? How often has this information actually come in handy?

nabobalis: I don't think it's bad idea in the long run. We should be looking to expand how smart and what detailed info we can get from our sysinfo and  think this is part 1 of that.

nabobalis: Or if you prefer we can just add it as a dependency  like we are doing for `importlib_metadata`

Cadair:
I think that the call on if this should go in extern or not is a close one. This code is pretty minimal in it's use, but I think @nabobalis makes a good point that it's in our own best interests to provide rich info in this function. If a user is calling this function then something has probably already gone wrong, and making them install a package (if we didn't extern it and didn't make it a dep) is another hurdle to them submitting a bug report.

If we added it as a linux only dep, it's yet another package that needs to be installed and I am more hesitant to add dependencies these days.

I think that adding it to extern is the thing that makes our users lives easier, so I am going to merge this on that basis.

PR #3397, opened 2019-10-08
Add an annotate=False kwarg to map.draw_grid

### LINKED ISSUE #3208: Add an annotate=False kwarg to map.draw_grid
This would disable the axes labels and the ticks on the top and right axes.

For example: http://docs.sunpy.org/en/latest/generated/gallery/map/submaps_and_cropping.html#sphx-glr-generated-gallery-map-submaps-and-cropping-py

In the submap, you have the top and right axes which correspond to the grid and we would like to turn that off.

gowthamraj198:
- added a dict with a new key and value
- then updated the dict with kwargs

### Description

Fixes #3208

### DIFF
--- a/changelog/3397.feature.rst
+++ b/changelog/3397.feature.rst
@@ -1,1 +1,1 @@
-Added a new key - annotation to kwergs of map.draw_grid
+Now able to turn off the tick and axes labels (``annotate=False``) to `sunpy.map.GenericMap.draw_grid`.
--- a/changelog/3397.feature.rst
+++ b/changelog/3397.feature.rst
@@ -1,0 +1,1 @@
+Added a new key - annotation to kwergs of map.draw_grid
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1416,7 +1416,7 @@ class GenericMap(NDData):
 # #### Visualization #### #
 
     @u.quantity_input
-    def draw_grid(self, axes=None, grid_spacing: u.deg = 15*u.deg, **kwargs):
+    def draw_grid(self, axes=None, annotation=False, grid_spacing: u.deg = 15*u.deg, **kwargs):
         """
         Draws a coordinate overlay on the plot in the Heliographic Stonyhurst
         coordinate system.
@@ -1441,18 +1441,16 @@ class GenericMap(NDData):
         Notes
         -----
         Keyword arguments are passed onto the `sunpy.visualization.wcsaxes_compat.wcsaxes_heliographic_overlay` function.
+        :param annotation:
         """
 
         if not axes:
             axes = wcsaxes_compat.gca_wcs(self.wcs)
         if not wcsaxes_compat.is_wcsaxes(axes):
             raise TypeError("Overlay grids can only be plotted on WCSAxes plots.")
-        c_kw = {'annotate': False
-                }
-        c_kw.update(kwargs)
-        return wcsaxes_compat.wcsaxes_heliographic_overlay(axes,
+        return wcsaxes_compat.wcsaxes_heliographic_overlay(axes, annotation,
                                                            grid_spacing=grid_spacing,
-                                                           **c_kw)
+                                                           **kwargs)
 
     def draw_limb(self, axes=None, **kwargs):
         """
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1416,7 +1416,7 @@ class GenericMap(NDData):
 # #### Visualization #### #
 
     @u.quantity_input
-    def draw_grid(self, axes=None, annotation=True, grid_spacing: u.deg = 15*u.deg, **kwargs):
+    def draw_grid(self, axes=None, annotation=False, grid_spacing: u.deg = 15*u.deg, **kwargs):
         """
         Draws a coordinate overlay on the plot in the Heliographic Stonyhurst
         coordinate system.
@@ -1433,6 +1433,8 @@ class GenericMap(NDData):
             Spacing for longitude and latitude grid, if length two it specifies
             (lon, lat) spacing.
 
+        annotation: to turn off the tick and axes labels
+
         Returns
         -------
         overlay: `~astropy.visualization.wcsaxes.coordinates_map.CoordinatesMap`
@@ -1441,7 +1443,6 @@ class GenericMap(NDData):
         Notes
         -----
         Keyword arguments are passed onto the `sunpy.visualization.wcsaxes_compat.wcsaxes_heliographic_overlay` function.
-        :param annotation:
         """
 
         if not axes:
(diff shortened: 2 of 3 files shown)


nabobalis:
Thanks @gowthamraj198 for the PR.

I think you misunderstand the original issue slightly. My understanding is that we want to add a keyword argument to  `draw_grid` similar to how `grid_spacing` is a keyword. Then, we would pass this keyword to  `wcsaxes_heliographic_overlay` which will turn off the axis labels in the top and right axes. The logic to turn off these labels is not implemented either and would need to be written in `wcsaxes_heliographic_overlay`  for this to work.

gowthamraj198:
> Thanks @gowthamraj198 for the PR.
> 
> I think you misunderstand the original issue slightly. My understanding is that we want to add a keyword argument to `draw_grid` similar to how `grid_spacing` is a keyword. Then, we would pass this keyword to `wcsaxes_heliographic_overlay` which will turn off the axis labels in the top and right axes. The logic to turn off these labels is not implemented either and would need to be written in `wcsaxes_heliographic_overlay` for this to work.

Thanks for the clarification @nabobalis I will do the changes and push again!

gowthamraj198:
@nabobalis if I am not wrong, only labels should be disabled.

nabobalis:
> @nabobalis if I am not wrong, only axis labels should be disabled.
> 
> below 2 lines inside wcsaxes_heliographic_overlay() should be on or off based on the logic.
> lon.set_axislabel('Solar Longitude', minpad=0.8)
> lat.set_axislabel('Solar Latitude', minpad=0.9)
> Please confirm

The issue says "axes labels and the ticks" so I would do that for now and we can check later on if that's what we want.

gowthamraj198:
@nabobalis i missed the ticks. 
So I should include the below lines also inside the logic.

    lon.set_ticks_position('tr')
    lat.set_ticks_position('tr')

nabobalis:
I think that just sets the tick locations, it doesn't remove them. You will have to run the code to find out, I unfortunately can not remember.

gowthamraj198:
@nabobalis 
    lon.set_ticks(spacing=lon_space, color=tick_color)
    lat.set_ticks(spacing=lat_space, color=tick_color)

this sets the ticks i guess!

nabobalis:
It looks like it might.

sunpy/map/mapbase.py
```diff
@@ -1416,7 +1416,7 @@ def superpixel(self, dimensions: u.pixel, offset: u.pixel=(0, 0)*u.pixel, func=n
 # #### Visualization #### #
 
     @u.quantity_input
-    def draw_grid(self, axes=None, grid_spacing: u.deg = 15*u.deg, **kwargs):
+    def draw_grid(self, axes=None, annotation=False, grid_spacing: u.deg = 15*u.deg, **kwargs):
```
```suggestion
    def draw_grid(self, axes=None, annotation=True, grid_spacing: u.deg = 15*u.deg, **kwargs):
```
I think we want this to default to be on.

sunpy/visualization/wcsaxes_compat.py
```diff
@@ -113,7 +113,7 @@ def default_wcs_grid(axes):
 
 
 @u.quantity_input
-def wcsaxes_heliographic_overlay(axes, grid_spacing: u.deg = 10*u.deg, **kwargs):
+def wcsaxes_heliographic_overlay(axes, annotate=False, grid_spacing: u.deg = 10*u.deg, **kwargs):
```
nabobalis:
```suggestion
def wcsaxes_heliographic_overlay(axes, annotate=True, grid_spacing: u.deg = 10*u.deg, **kwargs):
```
Same here.

sunpy/visualization/wcsaxes_compat.py
```diff
@@ -135,6 +135,7 @@ def wcsaxes_heliographic_overlay(axes, grid_spacing: u.deg = 10*u.deg, **kwargs)
     Notes
     -----
     Keywords are passed to `~astropy.visualization.wcsaxes.coordinates_map.CoordinatesMap.grid`.
+    :param annotate:
```
nabobalis: What is this?

sunpy/map/mapbase.py
```diff
@@ -1441,13 +1441,14 @@ def draw_grid(self, axes=None, grid_spacing: u.deg = 15*u.deg, **kwargs):
         Notes
         -----
         Keyword arguments are passed onto the `sunpy.visualization.wcsaxes_compat.wcsaxes_heliographic_overlay` function.
+        :param annotation:
```
nabobalis: What is this?

changelog/3397.feature.rst
```diff
@@ -0,0 +1 @@
+Added a new key - annotation to kwergs of map.draw_grid
```
nabobalis:
```suggestion
Now able to turn off the tick and axes labels (``annotate=False``) to `sunpy.map.GenericMap.draw_grid`.
```

nabobalis: Just to tidy up the changelog entry.

nabobalis:
I am not sure it worked. If you see here: https://19359-2165383-gh.circle-artifacts.com/0/root/project/figure_test_images/fig_comparison.html

This is our figure test suite and you can see the comparison to the new and older images in the middle column.  It seems the changes only made the grid spacing different and did not remove the ticks or the axis labels. 

The tricky part next is writing a unit test for the new keyword, but we can worry about that after we have the code working.

nabobalis:
I would strongly suggest that you create a  local developer environment to test your changes, we have a guide here: http://docs.sunpy.org/en/latest/dev_guide/newcomers.html

It will be quicker than making a change here and then waiting for CircleCI to run the tests.

gowthamraj198:
> I would strongly suggest that you create a local developer environment to test your changes, we have a guide here: http://docs.sunpy.org/en/latest/dev_guide/newcomers.html
> 
> It will be quicker than making a change here and then waiting for CircleCI to run the tests.

Thanks  for this. I am already in the process of setting up the dev env. 1 quick qn.

Here - 
https://19380-2165383-gh.circle-artifacts.com/0/root/project/figure_test_images/fig_comparison.html

I am referring to sunpy.map.tests.test_plotting.test_draw_grid_aia171 images. Am I referring the correct one? Also the expected is we should get rid off the axes values (1000, 500, 0 ,-1000, -500) also the small black '-' before the values. Is my understanding correct? @nabobalis

nabobalis:
See this example for the bug: http://docs.sunpy.org/en/latest/generated/gallery/map/submaps_and_cropping.html#sphx-glr-generated-gallery-map-submaps-and-cropping-py

nabobalis:
Closed in favour of https://github.com/sunpy/sunpy/pull/3673 

If you want to try another issue or help on that PR please go ahead!

PR #3400, opened 2019-10-10
Less sunpy

### LINKED ISSUE #3286: Too many SunPy's in the User guide menu
### Description

The side bar in the docs contains a whole bunch of unnecessary "in/of/with SunPy"s.  It's obvious that this describes SunPy.

### Expected behavior

Less "in/of/with SunPy"s.

### Actual behavior

The extra "in/of/with SunPy"s make it much harder to read the sidebar.

### Steps to Reproduce

Look at the User's guide.

Screen Shot 2019-07-29 at 11 49 04 AM

### System Details

n/a

eigenbrot:
### Description

Remove the "Sunpy"'s in the online table of contents. Fixes  #3286.

### DIFF
--- a/docs/guide/roi.rst
+++ b/docs/guide/roi.rst
@@ -1,6 +1,6 @@
-==================
-Region of Interest
-==================
+===================
+Regions of Interest
+===================
 
 .. warning:: This module is under development.
 
--- a/docs/guide/index.rst
+++ b/docs/guide/index.rst
@@ -12,12 +12,12 @@ For more details checkout the :ref:`reference`.
    :maxdepth: 3
 
    installation/index
-   tour
-   acquiring_data/index
-   data_types/index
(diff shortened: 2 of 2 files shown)


docs/guide/roi.rst
```diff
@@ -1,6 +1,6 @@
-==================
-Region of Interest
-==================
+===================
+Regions of Interest
+===================
```
nabobalis: I think this should be reverted.

eigenbrot: Is that because the ``sunpy.roi.roi.roi`` objects are for a single region of interest? Even in this case I feel the pluralization is correct because the module can be used to create an arbitrary number of ROIs.

nabobalis: Sure.

PR #3403, opened 2019-10-12
Delay all matplotlib imports to inside plot_settings and plot methods

Cadair:
This turned out to be a rather larger diff than I expected.

I took the opportunity of touch all the sources files to tidy them up a little.

### DIFF
--- a/sunpy/map/sources/hinode.py
+++ b/sunpy/map/sources/hinode.py
@@ -1,10 +1,4 @@
 """Hinode XRT and SOT Map subclass definitions"""
-#pylint: disable=W0221,W0222,E1101,E1121
-
-__author__ = ["Jack Ireland, Jose Ivan Campos-Rozo, David Perez-Suarez"]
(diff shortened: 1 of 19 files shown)


Cadair:
I have flagged as draft as I need to work out how we can test this properly, and also I need to double check the rest of the code base.

sunpy/map/sources/hinode.py
```diff
@@ -130,7 +133,10 @@ def __init__(self, data, header, **kwargs):
                  'SOT/SP': 'intensity',  # For the 1st 2 dimensions
                  }
 
-        self.plot_settings['cmap'] = plt.get_cmap('hinodesot' + color[self.instrument])
+        plot_settings = super()._default_plot_settings()
+        plot_settings['cmap'] = plt.get_cmap('hinodesot' + color[self.instrument])
```
dstansby: Is it possible to just store the string here instead of the full on `Colormap`?

Issue #3154, opened 2019-05-30
Changes after Astropy 3.1


ayshih:
This issue is a list of changes that can be made once we drop support for Astropy 3.1 (i.e. depend on Astropy 3.2+)

## Coordinates

- [x] [#3919] Change `get_horizons_coord()` to use `HeliocentricEclipticIAU76` to be more accurate
- ~Change `apparent_longitude()` and `apparent_latitude()` to use `GeocentricTrueEcliptic` (no change in accuracy)~  Using `GeocentricTrueEcliptic` gives incorrect answers, so I need to confirm whether this is a bug in Astropy's definition.  Update: confirmed to be a bug and fixed in astropy/astropy#10129
- [x] [#4005] Remove `try`...`except` blocks for importing `HeliocentricMeanEcliptic` and `GeocentricMeanEcliptic`
- [x] [#4005] Remove the `try`...`except` block for importing `make_transform_graph_docs` in coordinates.
- [x] [#4005] Simplify the [saving/loading example](https://docs.sunpy.org/en/latest/generated/gallery/saving_and_loading_data/coordinates_in_asdf.html) because SkyCoord can be saved

## Visualization

- [ ] `ArrayAnimatorWCS` has a compatibility wrapper for `Quantity.to_string`. [#3407]

ayshih:
> * Change `get_horizons_coord()` to use `HeliocentricEclipticIAU76` to be more accurate

I finally checked, and making this change decreases the discrepancy for the position of Earth from tens of kilometers to a few kilometers.  If I also make Astropy use a JPL ephemeris rather than its built-in ephemeris, the discrepancy drops to below a *meter* (not kilometer).

Cubostar:
Does this still need to be done? I can work on this.

ayshih:
@Cubostar This will need to be done, but not now.  We need to wait until SunPy depends on Astropy 3.2, and there isn't a compelling reason yet to bump up the dependency.

Cadair:
@ayshih I have hijacked this for all Astropy 3.1 compatibility changes.

Cadair:
I think we have decided that we are 3.2+ for 2.0 right? If so we should make all these changes?

PR #3407, opened 2019-10-14
Add support for WCS when animating line plots

Cadair:
I recently implemented astropy/astropy#9266 which enables the use of WCSAxes for 1D plots. This makes it much easier to provide a consistent plotting interface between 2D and 1D WCS aware plots. As part of an ongoing effort in sunpy/ndcube#207 I am trying to unify all the NDCube plotting to using WCS.

This PR implements a new `ArrayAnimatorWCS` class which allows you to plot 1D or 2D plots using a WCS, it provides a much reduced API interface, always relying on the WCS object for everything it can (no `axis_ranges` etc).

To make this work, I needed to change the base `ArrayAnimator` class to support (and always use underneath) functions to get the display coordinates for the sliders, this should be a backwards compatible change to the other animators.

TODO:

- [x] Error nicely if Astropy 4.0 isn't installed and 1D plots wont work.
- [x] Implement some method of setting WCSAxes properties on the coordinate object (format unit, label etc).
- [x] Work out how to deal with user overrides of `ylim` (@DanRyanIrish is this something we really need?)
- [x] Tests?! :scream: 
- [x] ~it would be nice to be able to set the display unit for the slider labels, but only because `astropy.wcs.WCS` forgets it and defaults to SI base.~

An animation of this working for an IRIS SJI cube (bottom slider is time, top slider is latitude):

line_animator_2


dstansby:
Can you ping me for again when you've finished making changes 😛

Cadair:
Sure! I had hoped the tests would pass, but you know! :roll_eyes:

sunpy/visualization/animator/image.py
```diff
@@ -173,7 +173,9 @@ def __init__(self, data, wcs, image_axes=[-1, -2], unit_x_axis=None, unit_y_axis
         self.slices_wcsaxes = list_slices_wcsaxes[::-1]
         self.unit_x_axis = unit_x_axis
         self.unit_y_axis = unit_y_axis
-        super().__init__(data, image_axes=image_axes, axis_ranges=axis_ranges, **kwargs)
+
+        # For some reason using `super()` here causes an error.
+        ImageAnimator.__init__(self, data, image_axes=image_axes, axis_ranges=axis_ranges, **kwargs)
```
ayshih: I don't expect you to know the answer since you say "for some reason" and "wtaf", but what's the deal?

Cadair: I have no flaming idea, the error being raised was:
```
  File "/home/circleci/project/.tox/build_docs/lib/python3.7/site-packages/sunpy/visualization/animator/image.py", line 176, in __init__
    super().__init__(data, image_axes=image_axes, axis_ranges=axis_ranges, **kwargs)
TypeError: super(type, obj): obj must be an instance or subtype of type
```

sunpy/visualization/animator/base.py
```diff
@@ -532,30 +536,36 @@ def _sanitize_axis_ranges(self, axis_ranges, data_shape):
                     axis_ranges[i] = edges_to_centers_nd(axis_ranges[i], i)
                 else:
                     raise ValueError(incompatible_axis_ranges_error_message(i))
-        # For each axis validate and translate the axis_ranges.
+
+        # For each slider axis validate and translate the axis_ranges.
+        def get_pixel_to_world_callable(array):
+            def pixel_to_world(pixel):
+                return array[pixel]
+            return pixel_to_world
```
DanRyanIrish: What is this code block doing...?

sunpy/visualization/animator/wcs.py
```diff
@@ -0,0 +1,241 @@
...
+    ylabel: `string`, optional
+       The yaxis label to use when drawing a line plot. Setting the label on
+       the y-axis on an image plot should be done via ``coord_params``.
+
+    """
+
+    def __init__(self, data, wcs, slices, coord_params=None, ylim=None, ylabel=None, **kwargs):
+        if not isinstance(wcs, BaseLowLevelWCS):
+            raise ValueError("A WCS object should be provided that implements the astropy WCS API.")
+        if wcs.pixel_n_dim is not data.ndim:
```
DanRyanIrish: Should this not be ```wcs.pixel_n_dim =! data.ndim``` to keep it general to any number of dimensions?  (Although I realize if the dimensions were greater than the Python-pre-assigned ints the data cube will likely by too big for most systems...)

sunpy/visualization/animator/wcs.py
```diff
@@ -0,0 +1,241 @@
...
+            raise ValueError("slices should be the same length as the number of pixel dimensions.")
+        if "x" not in slices:
+            raise ValueError("slices should contain at least 'x' to indicate the axis to plot on the x axis.")
+
+        self.plot_dimensionality = 1
+
+        image_axes = [slices[::-1].index("x")]
+        if "y" in slices:
+            image_axes.append(slices[::-1].index("y"))
+            self.plot_dimensionality = 2
```
DanRyanIrish: Does the way this is done mean that you can't customize which dimensions are on the x and y axis for image animations?

sunpy/visualization/animator/wcs.py
```diff
@@ -0,0 +1,241 @@
...
+
+        super().__init__(data, image_axes=image_axes, axis_ranges=None, **kwargs)
+
+    def _sanitize_axis_ranges(self, *args):
+        """
+        This overrides the behaviour of ArrayAnimator to generate axis_ranges
+        based on the WCS.
+        """
+
+        def partial_pixel_to_world(pixel_dimension, pixel_coord):
```
DanRyanIrish: What's the advantage of defining a function within a function here?

sunpy/visualization/animator/wcs.py
```diff
@@ -0,0 +1,241 @@
...
+
+            axislabel = params.get("axislabel", None)
+            if axislabel:
+                coord.set_axislabel(axislabel)
+
+            grid = params.get("grid", None)
+            if grid is not None:
+                if not isinstance(grid, dict):
+                    grid = {}
+                coord.grid(**grid)
```
DanRyanIrish: So the changes/actions of ```coord``` doesn't have to be done to the version attached to ```self```?

sunpy/visualization/animator/wcs.py
```diff
@@ -0,0 +1,241 @@
...
+            self._add_colorbar(im)
+        return im
+
+    def update_plot_2d(self, val, im, slider):
+        """
+        Update the image plot.
+        """
+        if val != slider.cval:
+            self.axes.reset_wcs(wcs=self.wcs, slices=self.slices_wcsaxes)
+            im.set_array(self.data_transposed)
```
DanRyanIrish: So this means the data is re-transposed every time the plot is updated?  Is this not a performance concern?

sunpy/visualization/animator/base.py
Cadair: This function now converts all the slider axes to functions, which are then called by the base animator when the slider is updated. So this is just moving the array slice into a function.

sunpy/visualization/animator/wcs.py
Cadair: yes, yes it should.

sunpy/visualization/animator/wcs.py
Cadair: It does not. You just specify `('y', 'x')` rather than `('x', 'y')` to specify the order.

sunpy/visualization/animator/wcs.py
Cadair: There isn't much either way really, it's just where the code belonged to me!

There is a potential performance optimisation in using closures here (for the lines in this function which are not dependant upon the pixel coord), which would require the function to be defined here, so I think it's probably a sane place for it.

sunpy/visualization/animator/wcs.py
Cadair: No, coord is attached to `self.axes` and coord is just the element in the dict, so this will happily work by reference.

sunpy/visualization/animator/wcs.py
Cadair: It does, I don't think transpose is a particularly performance sensitive operation in numpy, as it doesn't modify the data, it just creates a transposed view into the array.

sunpy/visualization/animator/base.py
DanRyanIrish: Is there a reason it's a function defined in a function defined in a function?

Cadair: yes, because it uses `array` from the defining scope (which is called a closure)

sunpy/visualization/animator/wcs.py
DanRyanIrish: OK. So long as there's a reason to justify its ugliness ;)

sunpy/visualization/animator/wcs.py
DanRyanIrish: So ```coord``` is a shallow copy of something attached to ```self.axes``` not a new deepcopy?

Cadair: it's not even a copy really, it's just a variable name pointing at the object in the dict.

sunpy/visualization/animator/wcs.py
Cadair: On further reflection I have moved it to its own method.

Cadair:
If you run the full test suite with figures you get 100% test coverage on astropy 4.0 :rofl: :sob:

Cadair:
@dstansby I am done with this, please review if you get a chance.

dstansby:
Looks like figure tests are failing.

Cadair:
The hashes just needed updating. Passes now.

dstansby:
Okay, I think that all makes sense - will give the PR a proper review now. Yes I think having an astropy-dev figure test run would be good (there's already one for none-figure tests right?)

sunpy/visualization/animator/image.py
dstansby: Is this because of a funny interaction with the `@deprecated` dectorator? If you remove the dectorator does `super()` work?

sunpy/visualization/animator/tests/test_basefuncanimator.py
```diff
@@ -165,8 +169,8 @@ def test_sanitize_axis_ranges(axis_ranges, exp_extent, exp_axis_ranges):
     out_axis_ranges, out_extent = aanim._sanitize_axis_ranges(axis_ranges=axis_ranges,
                                                               data_shape=data_shape)
     assert exp_extent == out_extent
-    assert np.array_equal(exp_axis_ranges[0], out_axis_ranges[0])
     assert np.array_equal(exp_axis_ranges[1], out_axis_ranges[1])
+    assert callable(out_axis_ranges[0])
```
dstansby: Even though this is a callable now, can we still do a comparision with expected ranges (as in the old line 168 which has now gone)?

sunpy/visualization/animator/tests/test_basefuncanimator.py
```diff
@@ -206,5 +211,8 @@ def test_imageanimator_figure():
     time_diff = (t1 - t0).to(u.s)
     wcs_input_dict.update({'CTYPE1': 'Time', 'CUNIT1': time_diff.unit.name, 'CDELT1': time_diff.value})
     wcs = astropy.wcs.WCS(wcs_input_dict)
-    wcs_anim = ImageAnimatorWCS(sequence_array, wcs=wcs, vmax=1000, image_axes=[0, 1])
-    plt.show()
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        wcs_anim = ImageAnimatorWCS(sequence_array, wcs=wcs, vmax=1000, image_axes=[0, 1])
```
dstansby: What warnings are being thrown up, and is it possible to do a more specific warning filter instead of just ignoring everything?

sunpy/visualization/animator/tests/test_wcs.py
```diff
@@ -0,0 +1,155 @@
...
+from astropy.wcs import WCS
+
+from sunpy.tests.helpers import figure_test
+from sunpy.visualization.animator.wcs import ArrayAnimatorWCS
+
+
+@pytest.fixture
+def wcs_4d():
+    header = dedent("""\
+    WCSAXES =                    4 / Number of coordinate axes
```
dstansby: 👍 I approve of extensive commenting here

sunpy/visualization/animator/tests/test_wcs.py
```diff
@@ -0,0 +1,155 @@
...
+                assert isinstance(a, u.Quantity)
+                assert a.value == wcs_4d.pixel_to_world_values(*[0]*wcs_4d.world_n_dim)[i]
+                assert a.unit == wcs_4d.world_axis_units[i]
+        else:
+            assert arange is None
+
+
+def test_constructor_errors(wcs_4d):
+    # WCS is not BaseLowLevelWCS
+    with pytest.raises(ValueError):
```
dstansby: Could you add the `match='string that is in error message'` bit to `pytest.raises` here, so it's easier to read exactly what is being caught.

sunpy/visualization/animator/line.py
```diff
@@ -148,8 +148,7 @@ def update_plot(self, val, line, slider):
         """
         val = int(val)
         ax_ind = self.slider_axes[slider.slider_ind]
-        ind = int(np.argmin(np.abs(self.axis_ranges[ax_ind] - val)))
```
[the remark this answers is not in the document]
Cadair: yes. I am not sure how this snuck in originally, but it was actually wrong and broke things.

sunpy/visualization/animator/tests/test_basefuncanimator.py
Cadair: it was to filter the `SunPyDeprecationWarning` I will change it :+1:

DanRyanIrish:
Shifting the play button to the left has worked a treat!  One other request though.  In cases where the slider is in pixel units, can it say "pix" in the unit position to be explicit?

sunpy/visualization/animator/base.py
```diff
@@ -532,30 +536,36 @@ def _sanitize_axis_ranges(self, axis_ranges, data_shape):
...
+
         for i in self.slider_axes:
             if axis_ranges[i] is None:
                 # If axis range not supplied, set pixel center values as integers starting at 0.
-                axis_ranges[i] = np.arange(data_shape[i])
-            else:
+                axis_ranges[i] = get_pixel_to_world_callable(np.arange(data_shape[i]))
```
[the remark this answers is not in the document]
Cadair: It's a change that was needed for the WCS to work, we need to be able to call `pixel_to_world` to get the world coordinate for the slider. I don't really want to go through the whole code and change it, also it's only the slider axes which are callables (the others are not).

(Also this is not a user facing change).

sunpy/visualization/animator/base.py
```diff
@@ -604,3 +617,16 @@ def edges_to_centers_nd(axis_range, edges_axis):
     lower_edges = axis_range[tuple(lower_edge_indices)]
 
     return (upper_edges - lower_edges) / 2 + lower_edges
+
+
+def format_quantity_as_string(quantity):
+    """
+    An Astropy 3.1 compatibility wrapper for Quantity.to_string.
```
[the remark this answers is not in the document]
Cadair: I have already added it to the list in #3154, will add a comment.

sunpy/visualization/animator/image.py
Cadair: You are correct, good spot! I have updated the comment.

Cadair:
@dstansby I think I have addressed all your comments. I added a couple more commits following an in-person review/testing exercise with @DanRyanIrish just in an attempt to make it have better formatting and defaults.

Cadair:
The figure test comparison (https://22838-2165383-gh.circle-artifacts.com/0/root/project/figure_test_images/fig_comparison.html) seems to indicate it's all fine apart from the colorbar one which is borked.

DanRyanIrish:
Is there anything more you need from me on this PR, @Cadair?

Cadair:
Another review once I have finished with all the CI issues would be good, but not right now.

DanRyanIrish:
Cool.

sunpy/tests/figure_hashes_py36.json
```diff
@@ -28,5 +28,9 @@
     "sunpy.timeseries.tests.test_timeseriesbase.test_norh_peek": "22e7a1d5cbb7e5dea4c886752e6be961b938515030ac936c821ba21177e4a840",
     "sunpy.timeseries.tests.test_timeseriesbase.test_rhessi_peek": "fcd235c9627ad38afd1ca90605bbfed05e5c6fb299c3afcd5dced78de06a061b",
     "sunpy.visualization.animator.tests.test_basefuncanimator.test_imageanimator_figure": "43febed509078c145da4e2b78f8b1a8d5c90ed89a8ce880e8eaa73e2785b6e29",
-    "sunpy.visualization.animator.tests.test_basefuncanimator.test_lineanimator_figure": "22addc09bcee836322349c6f086b1305b8c1dabf4176fb89afbd70da83dc79c8"
+    "sunpy.visualization.animator.tests.test_basefuncanimator.test_lineanimator_figure": "360673f30fa79c3218b4962bafb0aff85ff29e05949a1c5f25a0d78d0f200aba",
```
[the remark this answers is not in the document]
Cadair: yeah, it looks fine in the figure comparison page.

Cadair:
Ok, @DanRyanIrish @dstansby I think this is done.

Cadair:
I can write one, but the best way to use this is with `NDCube`. We are in a bit of a weird situation where this code is mostly used by ndcube but is currently in core. There are plans afoot to move it into ndcube at somepoint, probably around the time ndcube 2 is ready.

Cadair:
I found the old `ImageAnimatorWCS` example and updated it for this new class.

examples/plotting/imageanimatorwcs.py
```diff
@@ -1,99 +0,0 @@
...
-
-# Now we can change the spacing and we have to use `astropy.units` here.
-# We are setting the spacing to be quite small here and this will cause overlap.
-solar_x.set_ticks(spacing=1*u.arcmin, color='black')
-solar_y.set_ticks(spacing=1*u.arcmin, color='black')
-
-# We can make sure the ticks do not overlap.
-solar_x.set_ticklabel(exclude_overlapping=True)
-# Set this one to false, so we can see the difference.
-solar_y.set_ticklabel(exclude_overlapping=False)
```
nabobalis: Why we deleting this?

Cadair: Because it doesn't work, and I haven't implemented support for tick labels in `ArrayAnimatorWCS` as the new defaults in `WCSAxes` should generally be good enough.

Cadair: (specifically all these settings do not survive updating the animation)

nabobalis: Just remove that bit?
The old example was deleted for a reason.

Cadair: What? I renamed this file to refelect the new name of the class.

nabobalis: Then why did you say you just updated the old example?

Cadair: Well I updated it and then renamed it, so it's very similar, just in a different file name.

nabobalis: Ok that's fine. I thought you reverted to the old one.

Cadair:
Updated animation of using this thing in line mode:

line_animator_3

sunpy/visualization/animator/base.py
```diff
@@ -292,15 +298,15 @@ def _add_widgets(self):
...
-                nx1 = 1
+                nx1 = 3
             else:
-                nx1 = -3
-            locator = self.divider.new_locator(nx=0, ny=x, nx1=nx1)
+                nx1 = -2
+            locator = self.divider.new_locator(nx=2, ny=y, nx1=nx1)
             self.sliders[-1].set_axes_locator(locator)
-            sframe = widgets.Slider(self.sliders[-1], f"{i:d}",
+            sframe = widgets.Slider(self.sliders[-1], "",
```
DanRyanIrish: What's going on in these lines and the ~10 above?

sunpy/visualization/animator/base.py
```diff
@@ -292,15 +298,15 @@ def _add_widgets(self):
         self.sliders = []
         self.slider_buttons = []
         for i in range(self.num_sliders):
-            x = i * 2
+            y = i * 2
```
DanRyanIrish: Is changing this to y to fix one of the issues we identified when you were here?  Or is this another issue?

sunpy/visualization/animator/base.py
DanRyanIrish: Still don't understand this but as I see you've thought it through a second time I trust it's fine ;)

sunpy/visualization/animator/base.py
Cadair: It's just because it's a row index not a column index, so it was confusing to me re-reading the code that it was `x` and not `y`.

sunpy/visualization/animator/base.py
Cadair: It is computing the layout for the sliders. I am not going to attempt to explain it in detail in a GH comment, but basically the whole thing works by generating a  x by y grid and then placing axes on the grid which span a given number of rows and columns. So calls to `new_locator` specify the start index for the axes and how many rows and columns it spans. (`nx1`, `ny1`).

Cadair:
I have added one more commit to the end of this, which changes the default behaviour for line plots, so that the global min/max of the array don't need to be computed by default. As this is likely to cause memory and/or compute issues when using larger than memory memmaps or dask arrays.

nabobalis:
Test failures all seem to be servers not letting us download data from them.

PR #3408, opened 2019-10-14
Explicitly sort maps for glob and directory patterns

### LINKED ISSUE #3318: Passing glob pattern to Map returns unsorted list of Maps
When passing a glob pattern to `sunpy.map.Map`, the list of maps that it returns is not sorted, even if the filenames in the glob can be sorted, e.g. by time.

For example,

```python
from sunpy.map import Map
maps = Map('/path/to/my/files/*.fits')  # these maps are not sorted!
```

where all of the filenames in `'/path/to/my/files/'` include a timestamp such that they can be sorted sequentially. The resulting list of files, `maps`, is not sorted.

The reason for this is that [we are using `glob.glob()` to get the list of files in the case that the user passes in a pattern](https://github.com/sunpy/sunpy/blob/master/sunpy/map/map_factory.py#L214) and `glob()` does not sort the resulting list of files by default.

This can easily be solved by just wrapping a `sorted` argound `glob.glob(...)`. The question is do we want to make the choice to sort the files for the user? I find it a bit unexpected that the resulting Maps are not sorted if the filenames allow for it.

wtbarnes:
# Description 
Fixes #3318 by explicitly sorting maps that are created when passing in a directory or glob.

Also fixes minor pep8 issues in `test_map_factory.py` and adds factory tests for SWAP and XRT maps.

### DIFF
--- a/changelog/3408.feature.rst
+++ b/changelog/3408.feature.rst
@@ -1,4 +1,0 @@
-The returned list of `sunpy.map.Map` objects is now sorted by filename when passing a
-directory or glob pattern to `sunpy.map.MapFactory`. Single character wildcards and
-character ranges can now also be used as glob patterns. `sunpy.map.Map` also now accepts
-`pathlib.Path` objects, in addition to `str`, when passing a file system path.
--- a/changelog/3408.feature.rst
+++ b/changelog/3408.feature.rst
@@ -1,0 +1,4 @@
+The returned list of `sunpy.map.Map` objects is now sorted by filename when passing a
+directory or glob pattern to `sunpy.map.MapFactory`. Single character wildcards and
+character ranges can now also be used as glob patterns. `sunpy.map.Map` also now accepts
+`pathlib.Path` objects, in addition to `str`, when passing a file system path.
--- a/sunpy/map/map_factory.py
+++ b/sunpy/map/map_factory.py
@@ -207,13 +207,13 @@ class MapFactory(BasicRegistrationFactory):
             elif (isinstance(arg, str) and
                   os.path.isdir(os.path.expanduser(arg))):
                 path = os.path.expanduser(arg)
-                files = [os.path.join(path, elem) for elem in os.listdir(path)]
+                files = sorted([os.path.join(path, elem) for elem in os.listdir(path)])
                 for afile in files:
                     data_header_pairs += self._read_file(afile, **kwargs)
 
             # Glob
             elif (isinstance(arg, str) and '*' in arg):
-                files = glob.glob(os.path.expanduser(arg))
+                files = sorted(glob.glob(os.path.expanduser(arg)))
                 for afile in files:
                     data_header_pairs += self._read_file(afile, **kwargs)
 
--- a/sunpy/map/map_factory.py
+++ b/sunpy/map/map_factory.py
@@ -197,32 +197,29 @@ class MapFactory(BasicRegistrationFactory):
                     data_header_pairs.append(pair)
                     i += 1    # an extra increment to account for the data-header pairing
 
-            # File name
-            elif (isinstance(arg, str) and pathlib.Path(arg).expanduser().is_file()):
+            # File system path (file or directory)
+            elif _is_path(arg):
                 path = pathlib.Path(arg).expanduser()
-                pairs = self._read_file(path, **kwargs)
-                data_header_pairs += pairs
-
-            # Directory
-            elif (isinstance(arg, str) and pathlib.Path(arg).expanduser().is_dir()):
-                path = pathlib.Path(arg).expanduser()
-                files = sorted(list(path.glob('*')))
-                for afile in files:
-                    data_header_pairs += self._read_file(afile, **kwargs)
+                if path.is_file():
+                    pairs = self._read_file(path, **kwargs)
+                    data_header_pairs += pairs
+                elif path.is_dir():
+                    for afile in sorted(path.glob('*')):
+                        data_header_pairs += self._read_file(afile, **kwargs)
+                else:
+                    raise ValueError(f'{path} is neither a file nor a directory')
 
             # Glob
-            elif (isinstance(arg, str) and '*' in arg):
-                files = sorted(glob.glob(arg))
-                for afile in files:
+            elif isinstance(arg, str) and len(glob.glob(arg)) >= 1:
+                for afile in sorted(glob.glob(arg)):
                     data_header_pairs += self._read_file(afile, **kwargs)
 
             # Already a Map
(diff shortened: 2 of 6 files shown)


sunpy/map/map_factory.py
```diff
@@ -207,13 +207,13 @@ def _parse_args(self, *args, **kwargs):
             elif (isinstance(arg, str) and
                   os.path.isdir(os.path.expanduser(arg))):
                 path = os.path.expanduser(arg)
-                files = [os.path.join(path, elem) for elem in os.listdir(path)]
+                files = sorted([os.path.join(path, elem) for elem in os.listdir(path)])
```
Cadair: We should probably refactor this to use pathlib? i.e.

```suggestion
                files = sorted(list(Path(path).glob("*")))
```

sunpy/map/map_factory.py
```diff
@@ -207,13 +207,13 @@ def _parse_args(self, *args, **kwargs):
...
                 for afile in files:
                     data_header_pairs += self._read_file(afile, **kwargs)
 
             # Glob
             elif (isinstance(arg, str) and '*' in arg):
-                files = glob.glob(os.path.expanduser(arg))
+                files = sorted(glob.glob(os.path.expanduser(arg)))
```
Cadair: Also pathlib?

sunpy/map/map_factory.py
wtbarnes: Sure, but currently `_read_file` does not support `pathlib.Path` objects...

wtbarnes: Can always cast these to a string of course, before passing them, but I wonder if we should support pathlib at a lower I/O level?

Cadair: Yeah we should totally make `_read_file` support Path objects.

wtbarnes: Do we also want`Map` to support being passed `pathlib.Path`s? This could make some things a bit cleaner, though it may make the input parsing code a bit messy as we would have to check for either `str` or `pathlib.Path` in the case of directories, filenames, and globs.

Cadair: I am surprised it doesn't already so yes?!

wtbarnes: This is a bit trickier than I thought...So much of the I/O stuff relies on explicitly passing strings (e.g. in `sunpy.io.file_tools.read_file`, the filepath is pulled apart to check the extension).

wtbarnes:
So the discussion has devolved a bit into how we should support `pathlib.Path` arguments into `Map`. Ideally, the map factory would accept either `str` or `pathlib.Pathlib` objects as valid ways to specify a file/files and then specifics of that would be handled by the underlying reader functions.

I think this may deserve its own PR/issue(s) as there are going to need to be changes to parts of the `sunpy.io` as well.

sunpy/map/map_factory.py
```diff
@@ -197,23 +197,22 @@ def _parse_args(self, *args, **kwargs):
...
+                path = pathlib.Path(arg).expanduser()
+                files = sorted(list(path.glob('*')))
                 for afile in files:
                     data_header_pairs += self._read_file(afile, **kwargs)
 
             # Glob
             elif (isinstance(arg, str) and '*' in arg):
-                files = glob.glob(os.path.expanduser(arg))
+                path = pathlib.Path(arg).expanduser().parent
+                files = sorted(list(path.glob('*')))
```
ayshih: This doesn't look correct...  Will this return just the files specified by the wildcard pattern, or will this return every file in the directory?

Cadair: yeah the latter.

wtbarnes: Oof you're right. This is not what should be happening here...

wtbarnes: Should be fixed now.

wtbarnes:
Based on the Windows test failures, it looks like `pathlib.WindowsPath` does not like being passed a glob pattern(?) As I do not have a Windows machine, I cannot investigate this much further than that. 

Maybe it is best to just abandon `pathlib` altogether for now and leave it for another day...

wtbarnes:
Refactored the logic in `MapFactory._parse_args` to:
* allow `Map` to accept `pathlib.Path` objects
* simplify file and directory checks
* allow for more complex glob patterns than `*`
* (hopefully) Fix the Windows test failure

It still might be worth opening a separate PR to allow `pathlib.Path`s trickle all the way down to the I/O level. In this PR, I've just made sure to cast to a string in `MapFactory._read_file` before passing to `read_file`.

sunpy/map/map_factory.py
```diff
@@ -196,34 +197,29 @@ def _parse_args(self, *args, **kwargs):
...
+                    for afile in sorted(path.glob('*')):
+                        data_header_pairs += self._read_file(afile, **kwargs)
+                else:
+                    raise ValueError(f'{path} is neither a file nor a directory')
 
             # Glob
-            elif (isinstance(arg, str) and '*' in arg):
-                files = glob.glob(os.path.expanduser(arg))
-                for afile in files:
+            elif isinstance(arg, str) and len(glob.glob(arg)) >= 1:
```
Cadair: We can just do this right:
```suggestion
            elif isinstance(arg, str) and len(glob.glob(arg)):
```
Will `glob.glob` happily take any valid string without erroring?

wtbarnes:
I think all of the test failures are unrelated? They seem to be due to the unavailability of JPL Horizons

changelog/3408.feature.rst
```diff
@@ -0,0 +1,4 @@
+The returned list of `sunpy.map.Map` objects is now sorted by filename when passing a
+directory or glob pattern to `sunpy.map.MapFactory`. Single character wildcards and
+character ranges can now also be used as glob patterns. `sunpy.map.Map` also now accepts
```
dstansby: I think the acceptance of `Path` objects is worthy of it's own changelog entry.

wtbarnes: The changelog now has three separate entries: 
1. sorted output from glob and directories
2. support for glob patterns other than `*`
3. support for `pathlib.Path`

wtbarnes:
> Looks great! Just one requested change: document the new valid types of inputs in the docstring for `MapFactory`.

I've added a few pathlib examples to the docstring

> For extra credit, I suggest outputting a logger debug line (i.e., hidden by default) for each file that gets read. I foresee that being useful when passing in a directory or a wildcard pattern.

Ah good idea. Done! @ehsteve will be happy to hear that the logger is being used

ayshih:
Great!  Looks good to me.

PR #3410, opened 2019-10-15
Move sunpy.cm to sunpy.visualization.colormaps

### LINKED ISSUE #3392: Move cm subpackage into visualization subpackage
The `cm` (i.e. colormap) subpackage should be moved under the `visualization` subpackage. There's not really any reason it needs to stand on its own. Also, `cm` is a not a very descriptive name.

This was discussed at the dev telecon yesterday, but it seems there was not an issue created for it.

Cadair:
Fixes #3392


sunpy/cm/__init__.py
```diff
@@ -1,7 +1,9 @@
...
-The following colormaps are provided by this module:
-"""
-from sunpy.cm.cm import *  # noqa
+from sunpy.util.decorators import deprecated as _d
 
-for cmname in cmlist.keys():  # noqa
-    __doc__ += f"\n* '`~sunpy.cm.cm.{cmname}`'\n"
+_d("1.1",
+   "The sunpy.cm module has been moved.",
+   alternative="sunpy.visualization.colormaps")
```
ayshih: How is this deprecation warning triggered?  I don't see it when I import `sunpy.cm`.

sunpy/cm/__init__.py
```diff
@@ -1,7 +1,9 @@
-"""
-The following colormaps are provided by this module:
-"""
```
ayshih: It's a bit odd that this package no longer has a docstring.  How about importing `sunpy.visualization.colormaps` and copying over the `__doc__`?

Cadair:
So another fun thing, we define an `__all__` list in `sunpy.visualization.colormaps.cm` which excludes 99% of the objects in there, so only `show_colormaps` and `cmlist` (which isn't a list) get imported.

This is pretty much fine with the deprecated version, but what now doesn't work is this:

```
In [7]: from sunpy.cm.cm import sdoaia191                                                                                                                                                                                          
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-7-c83856b488f1> in <module>
----> 1 from sunpy.cm.cm import sdoaia191

ModuleNotFoundError: No module named 'sunpy.cm.cm'
```

ayshih:
Ha ha, of course, `cmlist` isn't a list.  Awesome.

So, `sunpy.cm.cm.sdoaia171` works, but `from sunpy.cm.cm import sdoaia171` doesn't.  Okay.

Cadair:
> Ha ha, of course, cmlist isn't a list. Awesome.

This might actually be a good time to change it. 

> So, sunpy.cm.cm.sdoaia171 works, but from sunpy.cm.cm import sdoaia171 doesn't. Okay.

Are we happy with this, because it did work before this change?

sunpy/cm/__init__.py
```diff
@@ -1,7 +1,12 @@
...
-from sunpy.cm.cm import *  # noqa
+import warnings as _w
+from sunpy.util.exceptions import SunpyDeprecationWarning
 
-for cmname in cmlist.keys():  # noqa
-    __doc__ += f"\n* '`~sunpy.cm.cm.{cmname}`'\n"
+_w.warn("The sunpy.cm module has been moved in 1.1 to sunpy.visualization.colormaps, it will be removed in 2.1", category=SunpyDeprecationWarning)
```
wafels: How about "The functionality of the sunpy.cm module is now in sunpy.visualization.colormaps as of SunPy 1.1.  The ability to import sunpy.cm as a module may be removed in a future version of SunPy."

Cadair:
Ok, with a little bit of high grade hackery, I fixed the `import from` issue. So if the CI agrees I think this is good to go.

ayshih:
The figure-test build is still broken, though...

nabobalis:
```
______________________________________________________________________ ERROR collecting sunpy/cm/__init__.py _______________________________________________________________________
sunpy/cm/__init__.py:9: in <module>
    category=_SDW)
E   sunpy.util.exceptions.SunpyDeprecationWarning: The functionality of the sunpy.cm module is now in sunpy.visualization.colormaps as of SunPy 1.1. The ability to import sunpy.cm as a module may be removed in a future version of SunPy.
```
Breaking on raising the DepWarn.

ayshih:
Yeah, the open question during the telecon was why the figure-test build encounters that problem when collecting tests, but not, for example, the 32-bit build.

sunpy/cm/__init__.py
```diff
@@ -1,7 +1,28 @@
...
+from sunpy.visualization.colormaps import cm
+
+_sys.modules['sunpy.cm.cm'] = sunpy.visualization.colormaps.cm
+_sys.modules['sunpy.cm.color_tables'] = sunpy.visualization.colormaps.color_tables
+
+# Import the two original functions in this namespace and use __all__ to keep
+# import * behaviour the same.
+from sunpy.visualization.colormaps.cm import *
+
+__all__ = ['show_colormaps', 'cmlist']
```
ayshih: Technically, you're changing the behavior of `from sunpy.cm import *`, but that's because it didn't previously have an `__all__`.  (For example, `cmname` would get imported.)  However, the previous behavior is essentially a bug, and no one would have relied on it, so I support this change.

Cadair: I think it is only cname.

This `__all__` is copied from `sunpy.cm.cm.__all__` which we were importing [with import *](https://github.com/sunpy/sunpy/blob/master/sunpy/cm/__init__.py#L4)

Cadair:
Can someone else please review this and then merge it

docs/dev_guide/sunpy_stability.yaml
```diff
@@ -1,6 +1,3 @@
-cm:
-  status: mature
-  comments: Only forwards compatible changes expected.
```
dstansby: 😆

PR #3411, opened 2019-10-16
Give Map cmap and norm properties

dstansby:
Description to follow...

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -197,18 +197,14 @@ class GenericMap(NDData):
         self._shift = SpatialPair(0 * u.arcsec, 0 * u.arcsec)
 
         if self.dtype == np.uint8:
-            norm = None
+            self.norm = None
         else:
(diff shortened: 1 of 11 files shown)


sunpy/map/mapbase.py
```diff
@@ -1414,6 +1410,44 @@ def superpixel(self, dimensions: u.pixel, offset: u.pixel=(0, 0)*u.pixel, func=n
...
+                'norm': self.norm,
+                'interpolation': 'nearest',
+                'origin': 'lower'}
+
+    @plot_settings.setter
+    def plot_settings_set(self, plot_settings):
+        if 'cmap' in plot_settings:
+            self.cmap = plot_settings['cmap']
+        if 'norm' in plot_settings:
+            self.norm = plot_settings['norm']
```
Cadair: You still need to set `plot_settings` here, as it's a way of setting arbitrary keyword argument to imshow.

PR #3412, opened 2019-10-16
Store colormap names instead of Colormap objects

dstansby:
Matplotlib doesn't care where a colormap is actually a `Colormap` or just a `str` that is in their internal colormap registry. This saves having to import Matplotlib in each different map source.

### DIFF
--- a/changelog/3412.feature
+++ b/changelog/3412.feature
@@ -1,2 +1,2 @@
-`Map` objects now have a `.cmap` attribute, which returns the full colormap
+`~sunpy.map.GenericMap` objects now have a ``.cmap`` attribute, which returns the full `~matplotlib.colormap.Colormap`.
 object.
--- a/changelog/3412.feature
+++ b/changelog/3412.feature
@@ -1,0 +1,2 @@
+`Map` objects now have a `.cmap` attribute, which returns the full colormap
+object.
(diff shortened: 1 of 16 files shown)


Cadair:
yes this is a very good idea.

Cadair:
Could probably use a test for `.cmap`

changelog/3412.feature
```diff
@@ -0,0 +1,2 @@
+`Map` objects now have a `.cmap` attribute, which returns the full colormap
```
Cadair:
```suggestion
`~sunpy.map.GenericMap` objects now have a ``.cmap`` attribute, which returns the full `~matplotlib.colormap.Colormap`.
```

docs/guide/data_types/maps.rst
```diff
@@ -478,9 +478,10 @@ very easily when initializing the normalization variable. For example::
 This clips out many of the brightest pixels. If you'd like to see what areas of
 your images got clipped set the following values::
 
-    >>> cmap = cmap.plot_settings['cmap']  # doctest: +SKIP
+    >>> cmap = map.cmap  # doctest: +SKIP
     >>> cmap.set_over('red', 1.0)  # doctest: +SKIP
     >>> cmap.set_under('green', 1.0)  # doctest: +SKIP
+    >>> map.cmap = cmap   # doctest: +SKIP
```
ayshih:
```suggestion
```
This line doesn't work – you can't set this property – and even if it did work, it'd be unnecessary because you've already modified the colormap directly.

dstansby: Ahh, good catch.

PR #3416, opened 2019-10-17
Support pathlib.Path input to map_factory

dstansby:
This was a surprisingly simple change... A longer term goal should be cleaning up the internal `map_factory` code to use `pathlib`, but I leave that as an exercise for the contributor.

xref #2822


ayshih:
See #3408

dstansby:
Ah thanks, I didn't spot that.

PR #3421, opened 2019-10-21
[1.0] Revert "skip a test on 1.0 only as it's causing the release to stall."

Cadair:
In the process of releasing 1.0.4 I skipped this test that was returning no data from the VSO. This is to un-skip it once the online build passes again.


nabobalis:
We don't need this?

PR #3422, opened 2019-10-22
Reduce the number of files we upload after doc build

Cadair:
This should hopefully speed up the build on circle ci, and does three things:

1) Changes the location of the `.doctrees` pickles, so we don't end up uploading them.
2) Deletes the `_sources` folder which is a *lot* of files and is isn't involved in the main rendering of the docs.
3) Only uploads on PRs (credit @bsipocz astropy/astropy#9426)

@bsipocz If this works I will PR a version to Astropy.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,6 +126,17 @@ jobs:
       - run: *sudo-apt-install
       - run: pip install -U tox
       - run: tox -e build_docs
+      - run:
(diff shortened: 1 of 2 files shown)


Cadair:
On the previous CircleCI build the upload took 4:14 on this one it took 2:25 so that's a pretty significant saving.

PR #3430, opened 2019-10-24
Clarify definition of map.dsun

dstansby:
It occurred to me that "the Sun" is technically not a point, so this should read "center of the Sun".

### DIFF
--- a/changelog/3430.doc.rst
+++ b/changelog/3430.doc.rst
@@ -1,0 +1,1 @@
+Clarified the meaning of :attr:`GenericMap.dsun`.
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -624,7 +624,7 @@ class GenericMap(NDData):
 
     @property
     def dsun(self):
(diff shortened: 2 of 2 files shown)


Cadair:
>  "the Sun" is technically not a point

your approximations are weak :stuck_out_tongue:

dstansby:
> > "the Sun" is technically not a point
> 
> your approximations are weak 😛

I know, but I have had to go a step further and consider the Sun as a perfectly round sphere.

dstansby:
Woops, did not mean to close...

nabobalis:
In the future a way for a bot to create a changelog entry from a commit message or title would be nice.

dstansby:
Yes definitely... something that can automatically be done through a comment instead of having to faff around with the source code and git again would be great.

Cadair:
Yeah I think we all had the same idea at the same time :laughing: 

I am about 95% of the way to having rewritten sunpy bot (based on big changes to Baldrick) would be easy enough to add a command to make a changelog entry I think.

PR #3431, opened 2019-10-24
Put parse_time example in an example folder

dstansby:
It is currently by itself at the top: https://docs.sunpy.org/en/latest/generated/gallery/index.html , but probably should be in a folder.


nabobalis:
Maybe Units and Coordinates?

Cadair:
We could change timeseries to "time and time series"?!

nabobalis:
Seems a bit forced but maybe?

Cadair:
yeah or `Units, Coordinates and Times` might work better?

nabobalis:
I prefer that myself. 
What about `Coordinates, Times and Units`?

Cadair:
sure that does read nicer.

nabobalis:
I forgot how gallery works out the section titles.

dstansby:
I like the efficient decision making! Will update the location/folder title.

nabobalis:
Screenshot from 2019-10-24 17-19-24
Looking good expect the cutoff on my vertical monitor.

nabobalis:
On a larger side note that is even more unrelated to this PR,someone  should fix those online tests.

nabobalis:
Thanks @dstansby

PR #3438, opened 2019-10-24
Fix a database remote test (maybe?)

dstansby:
This is one of the tests failing in CI. I can get it to pass locally with these changes, but it looks different to the error method on CI...

As an aside, presumably CI fails every time these dates are updated on the remote server, which is a bit annoying/dull.


nabobalis:
Why they seem to keep changing is the question.

dstansby:
The files have `'MISSIONLONG'` in them, so I'm guessing they're updated to cover the whole mission as it progresses?

nabobalis:
Online tests failed due to permission error and Helio servers are down. This is good to merge.

PR #3439, opened 2019-10-24
Put path in error message when write permission denied

dstansby:
Won't fix CI, but hopefully a nicer error message...


nabobalis:
I wonder if those CI path issues is that they aren't using a tmpdir for that test. But I'm sure we set the entire test suite to be in a tmpdir so it shouldn't error.

Cadair:
I think the root of that error is that we were creating the target directory later in the process and we are now checking if the whole path exists first. I think we need to check the write ability of the prefix which actually exists rather than the target path.

nabobalis:
> I think the root of that error is that we were creating the target directory later in the process and we are now checking if the whole path exists first. I think we need to check the write ability of the prefix which actually exists rather than the target path.

```
PermissionError: You do not have permission to write to the directory /tmp/pytest-of-vsts/pytest-0/popen-gw0/test_save_path2/{instrument}/{level}
```
Since that is  what it checks for.

Cadair:
I think we might as well fix it in this, I might be able to take a look later.

nabobalis:
I mean I "fixed" it but it's so bad.

nabobalis:
Now the only PR with a fully passing CI.

sunpy/net/fido_factory.py
```diff
@@ -347,9 +347,9 @@ def fetch(self, *query_results, path=None, max_conn=5, progress=True,
         >>> filepaths = Fido.fetch(filepaths)  # doctest: +SKIP
 
         """
-        if path is not None and not os.access(path, os.W_OK):
+        if path is not None and not os.access(str(path).split("{")[0], os.W_OK):
```
nabobalis: There has to be a better way than this. This was all I could come up with.

sunpy/net/fido_factory.py
```diff
@@ -347,9 +347,9 @@ def fetch(self, *query_results, path=None, max_conn=5, progress=True,
...
             raise PermissionError('You do not have permission to write'
-                                  ' file in this directory')
+                                  f' to the directory {path}')
```
nabobalis: I should probably change the error message too, so people aren't confused as to why the `{}` parts of the path are here but not checked for.

sunpy/net/fido_factory.py
dstansby: I think you want to take [`path.parts`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parts), and just use all the parts that come before the first part that contains a `{`.

Cadair: more than that, you want to check the highest folder that actually exists, because the downloader will create folders in the tree if they don't

Cadair: the best way I could figure out was this:
```python
>>> path = Path("/home/stuart/does/not/exist")                                                                                                                                                                                     
>>> exists = list(filter(lambda p: p.exists(), path.parents))                                                                                                                                                                      
>>> os.access(exists[0], os.W_OK)                                                                                                                                                                                                  
True
```

nabobalis: Copied and pasted into the code.

nabobalis:
Using the github editor is always a mistake.

PR #3427, opened 2019-10-23
Explicitly set the clipping to False for the default normalizers for Map sources

ayshih:
When looking at [our documentation for clipping Map data when plotting](https://docs.sunpy.org/en/latest/guide/data_types/maps.html#masking-and-clipping-data), I thought that it could be updated to incorporate the `clip_interval` functionality.  However, I was initially befuddled why the over/under functionality of colormaps didn't appear to work.  It turns out that the documentation explicitly uses Matplotlib's `Normalize`, which has a clipping keyword that defaults to `False`, while our default normalizers use Astropy's derived `ImageNormalize`, which changes the default value of the clipping keyword to `True` (as of astropy/astropy#7800 a year ago).  As noted in [Matplotlib's documentation for `Normalize`](https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.colors.Normalize.html), setting the clipping keyword to `True` "silently defeats the purpose of setting the over, under, and masked colors in the colormap, so it is likely to lead to surprises".  I completely agree, and I have created an issue on Astropy (astropy/astropy#9449).

This PR explicitly sets the clipping to `False` for the default normalizers for Map sources.

Example code for what this PR enables:
```python
from sunpy.map import Map
from sunpy.data.sample import AIA_171_IMAGE
import astropy.units as u

m = Map(AIA_171_IMAGE)
m.plot_settings['cmap'].set_over('blue')
m.plot_settings['cmap'].set_under('red')
m.peek(clip_interval=[0.01, 0.99]*u.one)
```
Figure_1

PR #3444, opened 2019-10-25
[Backport 1.0] Explicitly set the clipping to False for the default normalizers for Map sources

ayshih:
Manual backport of #3427


nabobalis:
Thanks @ayshih

Issue #3445, opened 2019-10-25
Reduce import time for `sunpy`


ayshih:
(I was getting annoyed by our import times, and was inspired by astropy/astropy#4598 to look deeper.)

Importing `sunpy` takes just about a second on my machine (see below).  Over half of that time is spent importing `pkg_resources` for `sunpy.version`, which seems horribly excessive.  Can we get rid of using `pkg_resources`?  (See astropy/astropy#7647 for when they re-factored out `pkg_resources` to speed up their import time.)

```
python -X importtime -c "import sunpy" 2> sunpy.log
tuna sunpy.log
```
![sunpy](https://user-images.githubusercontent.com/991759/67575349-63f14280-f70a-11e9-8bd4-7065a9809178.png)

nabobalis:
It might be possible to replace it with something like:

```
from setuptools_scm import get_version
version = get_version(root='..', relative_to=__file__)
```
Not sure if that means adding `setuptools_scm` as an install dep tho.

nabobalis:
Or maybe:

```
>>> from importlib.metadata import version  
>>> version('wheel')  
'0.32.3'
```
from
https://docs.python.org/3/library/importlib.metadata.html

Cadair:
I don't think we can remove `pkg_resources` for the version and continue to use setuptools_scm, we don't want to add scm as a runtime dep, and doing what @nabobalis suggests above probably uses pkg_resources anyway!

ayshih:
Neato, created PR #3449

dstansby:
The bottleneck seems to be astropy and numpy now, so I guess there's not much more to be done here?
<img width="1657" alt="Screenshot 2019-11-12 at 12 00 06" src="https://user-images.githubusercontent.com/6197628/68670123-0cddd100-0544-11ea-83ec-7e1809d139a5.png">

PR #3449, opened 2019-10-25
Use a faster library to get the package version

ayshih:
This PR uses `importlib_metadata` instead of `pkg_resources` to get the package version, and reduces the time to import `sunpy` by ~0.4 s (see #3445).  This library is standard in Python 3.8, but is a separate package in earlier versions of Python.  I don't know how conceivable it is that it wouldn't be installed, but I kept a `pkg_resources` fallback to be safe.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,7 @@ install_requires =
   pandas
   astropy>=3.1
   parfive[ftp]
-  importlib_resources;python_version<"3.7"
+  importlib_resources;python_version<"3.8"
 
 [options.extras_require]
 database = sqlalchemy
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,7 +29,9 @@ classifiers =
 python_requires = >=3.6
 packages = find:
 include_package_data = True
-setup_requires = setuptools_scm
+setup_requires =
+  setuptools_scm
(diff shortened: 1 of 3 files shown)


Cadair:
We can add the standalone package as a dependency only on Python 3.7, see [this](https://github.com/DKISTDC/dkist/blob/master/setup.cfg#L34)

Cadair:
interestingly it seems pytest is now using that package on <3.8

setup.cfg
```diff
@@ -29,14 +29,17 @@ classifiers =
 python_requires = >=3.6
 packages = find:
 include_package_data = True
-setup_requires = setuptools_scm
+setup_requires =
+  setuptools_scm
+  importlib_resources;python_version<"3.7"
```
Cadair: do we need this here?

nabobalis: Probably not, I was wondering if I added it there could I remove the check for `pkg_resources` and the answer was a no. Which I find odd since surely it should be installed at setup time like `scm` is.

Cadair: Where is this version calc even called during setup time? I would have expected scm to just handle it. Or do we import this into setup.py in some way?

nabobalis: Who knows.

PR #3469, opened 2019-10-31
Report incorrect filetype in write_file() error message

dstansby:
As well as including a more descriptive error message,

- Remove code duplication
- Use pathlib to get the file extension
- Allow pathlib objects to go through `write_file()`

### DIFF
--- a/sunpy/io/file_tools.py
+++ b/sunpy/io/file_tools.py
@@ -2,6 +2,7 @@
 This module provides a generic file reader.
 """
 import os
+import pathlib
 import re
 
 try:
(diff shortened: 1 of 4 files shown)


sunpy/io/file_tools.py
```diff
@@ -144,18 +145,15 @@ def write_file(fname, data, header, filetype='auto', **kwargs):
     * This routine currently only supports saving a single HDU.
     """
     if filetype == 'auto':
-        if not isinstance(fname, str):
-            raise ValueError("Can not automatically detect filetype for non-string fname argument")
```
dstansby: I've removed this error, because anything that can't be coerced into a Path will throw an error at the `pathlib.Path` stage.

wtbarnes:
Why has the code coverage decreased?

dstansby:
Overall the coverage change seems fine; the diff coverage is not testing error, I can add a test for that if required?

wtbarnes:
I don't really have an opinion either way. And it looks like the online test failures are due to the HELIO servers being down so I'm fine to merge this as is.

PR #3470, opened 2019-10-31
Remove u.Quantity calls from examples

dstansby:
Most of the calls can be simplified using normal multiplication with units, which I think makes the code a bit nicer/readable.

### DIFF
--- a/examples/saving_and_loading_data/coordinates_in_asdf.py
+++ b/examples/saving_and_loading_data/coordinates_in_asdf.py
@@ -68,7 +68,7 @@ def semi_circular_loop(length: u.m, theta0: u.deg=0*u.deg):
 
     return SkyCoord(
         x=r.to(u.cm) * np.sin(phi.to(u.radian)),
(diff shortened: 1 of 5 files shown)


examples/saving_and_loading_data/coordinates_in_asdf.py
```diff
@@ -68,7 +68,7 @@ def r_2_func(x):
 
     return SkyCoord(
         x=r.to(u.cm) * np.sin(phi.to(u.radian)),
-        y=u.Quantity(r.shape[0] * [0 * u.cm]),
+        y=r.shape[0] * u.cm * 0,
```
Cadair: What is this doing!?

ayshih: Even simpler:
```python
np.zeros_like(r)
```

ayshih: (The code as written works because the `0 * u.cm` gets broadcasted anyway, so you could just replace it all with a scalar Quantity.)

PR #3478, opened 2019-11-03
Add __citation__

TomAugspurger:
### Description

Based off https://www.astropy.org/acknowledging.html, this adds
a `__citation__` variable to the top-level of the package, to make it easier
to find the citation info.

(Discussed with @hayesla at the NumFOCUS summit).

One note, this duplicates the citation. In theory, the source of truth could be in `sunpy/__init__.py`, and `CITATION.rst` could execute that? But that seems maybe more complicated than necessary.


hayesla:
Thanks @TomAugspurger ! I wonder if we should also add our new V1.0 paper @ehsteve @Cadair @mbobra ?

Cadair:
Thanks a lot @TomAugspurger. I would really like to unify this with the rst file, as we (reasonably) recently unified the main sunpy.org and documentation site with this one `CITATION.rst` file because they had got out of sync with each other!

We could pull the code block out of the rst with minimal effort?

TomAugspurger:
I believe so... My two thoughts are

1. Using `html_context` similar to pandas in
https://github.com/pandas-dev/pandas/blob/165d5ee40862ba155b5988045164e21aaaefd556/doc/source/conf.py#L312-L335.
Though that may be HTML output only, which probably isn't acceptable.
2. A small sphinx extension. Shouldn't be too difficult.

I can take a look over the next couple days.

Cadair:
Are you considering having the `__init__` be the source of truth or the `CITATION.rst`? I worry about dependency issues with a sphinx extension, either importing `sunpy.__citation__` would need sphinx or building the sunpy website would need dev sunpy (not impossible but would slow our site build a little).

TomAugspurger:
> Are you considering having the __init__ be the source of truth or the CITATION.rst? 

That was my initial thought.

It seems like astropy actually reads in their citation from the file system in `astropy/__init__.py`: https://github.com/astropy/astropy/blob/47575b3d92732de6e8694b004d850f6f45a241f1/astropy/__init__.py#L344-L355

Does that seem reasonable?

nabobalis:
Yeah if that works, lets do the same here.

nabobalis:
I've done the change to use astropy's method.

Thanks again @TomAugspurger

TomAugspurger:
Thanks for that. Got busy last week :)

nabobalis:
> Thanks for that. Got busy last week :)
> […](#)
> On Nov 16, 2019, at 06:52, Nabil Freij ***@***.***> wrote: ﻿ I've done the change to use astropy's method. Thanks again @TomAugspurger — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.

That is perfectly fine!

PR #3483, opened 2019-11-06
Allow `map.make_fitswcs_header` to take a shape or an array

Cadair:
This makes a lot of sense if you don't have an array. For instance a lot of the reprojection examples.

### DIFF
--- a/sunpy/map/header_helper.py
+++ b/sunpy/map/header_helper.py
@@ -49,8 +49,9 @@ def make_fitswcs_header(data, coordinate,
 
     Parameters
     ----------
-    data : `~numpy.ndarray`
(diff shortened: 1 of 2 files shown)


sunpy/map/header_helper.py
```diff
@@ -49,8 +49,9 @@ def make_fitswcs_header(data, coordinate,
 
     Parameters
     ----------
-    data : `~numpy.ndarray`
-        Array data of Map for which a header is required.
+    data : `~numpy.ndarray` or `tuple`
+        Array data of Map for which a header is required, or the shape of an
```
hayesla: sorry for being pedantic - but I think it would be best to say 'or that shape of _the_ data array`

PR #3477, opened 2019-11-01
Apply expanduser before passing to glob in map factory

wtbarnes:
### Description

Apply `os.path.expanduser` before passing a pattern to glob. This allows for passing patterns like `~/sunpy/data/aia*.fits` to the MapFactory.

Fixes #3476

PR #3485, opened 2019-11-06
[1.0] Backport 3477 to 1.0

nabobalis:
Backport of https://github.com/sunpy/sunpy/pull/3477


nabobalis:
Added #3408 for good measure

changelog/3408.feature.1.rst
```diff
@@ -0,0 +1,2 @@
+The returned list of `~sunpy.map.Map` objects is now sorted by filename when
```
Cadair: My opinion is that this should be a backport, but it's actually a subtle behaviour change.

@wtbarnes @ayshih thoughts?

nabobalis: I backported it since the other PR was tagged and made no sense without it.

I also think that that change isn't a feature but a bugfix. YOu would or should expect map to sort the input.

ayshih: I'm not a fan of backporting #3408 as-is for a couple of (nitpicky) reasons:

- The changelog descriptions are labeled and written as if they are new features, as opposed to fixing bugs in intended behavior.
- #3408 actually includes three "features", and I think it's difficult to argue that the `pathlib` stuff is merely a bugfix.

In an ideal world, I'd want to see the `pathlib` stuff separated out and not backported, and the changelog entries re-written for the other two "features" – sorting and additional glob patterns – to make them bugfixes.

nabobalis: I can probably just remove the pathlib stuff manually from this PR and rename the changelogs.

nabobalis: I've removed the pathlib feature and its just the other two bug fixes now.

PR #3488, opened 2019-11-08
Add an example showing fading between two maps

dstansby:
Some code I had lying around, which might be a nice example for the gallery.

### DIFF
--- a/examples/plotting/fading_between_maps.py
+++ b/examples/plotting/fading_between_maps.py
@@ -19,28 +19,36 @@ import sunpy.map
 from sunpy.data.sample import AIA_171_IMAGE, AIA_1600_IMAGE
 
 ###############################################################################
-# Start by loading two AIA maps from the sample data
+# Start by loading two AIA maps from the sample data.
+#
+# Note that these maps must be co-aligned. If they are taken at different times
+# or from different observation locations they will need to be aligned using
+# rotate() or reproject().
 map_171 = sunpy.map.Map(AIA_171_IMAGE)
 map_1600 = sunpy.map.Map(AIA_1600_IMAGE)
 
 ###############################################################################
 # First we create a figure, and add the axes that will show the maps. Then
-# plot both of the images on the same axes.
+# plot both of the images on the same axes. Finally, we add another axes that
+# contains the slider.
 fig = plt.figure()
+
+# Add the main axes. Note this is resized to leave room for the slider axes
 ax = fig.add_axes([0.1, 0.2, 0.9, 0.8], projection=map_171)
 
 im_1600 = map_1600.plot(axes=ax)
 im_171 = map_171.plot(axes=ax, alpha=0.5)
 ax.set_title('AIA 171 + 1600')
 
-###############################################################################
-# Now add another axes to the figure, which contains the slider
+# Add the slider axes
 ax_slider = fig.add_axes([0.25, 0.1, 0.65, 0.03])
 slider = Slider(ax_slider, 'Alpha', 0, 1, valinit=0.5)
 
+
 ###############################################################################
 # Finally, define what happens when the slider is changed and link this to the
-# slider
+# slider we set up above. In this case we just set the alpha (ie. transparencey)
+# of the 171 image.
 def update(val):
     alpha = slider.val
     im_171.set_alpha(alpha)
--- a/examples/plotting/fading_between_maps.py
+++ b/examples/plotting/fading_between_maps.py
@@ -1,4 +1,5 @@
 """
+=======================
 Fading between two maps
 =======================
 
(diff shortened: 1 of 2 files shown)


wtbarnes:
I have no idea what's up with the docs build. I've just restarted it to see if that helps?

wtbarnes:
I think you need to place the `add_axes` call and the `plot` calls all in the same block (i.e. not under different `#####` headings). Not entirely sure why this is, but this is what we're doing in all of our other examples.

dstansby:
Thanks for the feedback - I've added some motivation text to the top of the example, but any suggestions on that are very welcome.

wtbarnes:
Really a shame that the interactive output does not show up in the example gallery.

Cadair:
> Really a shame that the interactive output does not show up in the example gallery.

That's because it's saving the plot before the slider and stuff is added.

dstansby:
Thanks for the feedback.

> Really a shame that the interactive output does not show up in the example gallery.

Maybe this is a good excuse to build a bokeh plotting backend? 😛

nabobalis:
> Thanks for the feedback.
> 
> > Really a shame that the interactive output does not show up in the example gallery.
> 
> Maybe this is a good excuse to build a bokeh plotting backend?

Don't you mean plotly?

Cadair:
I think you need to refactor this due to https://github.com/sphinx-gallery/sphinx-gallery/issues/363

Specifically:

> [...] we are currently closing all the figures at the end of a code block. So unless I am wrong, the basic assumption is that a figure is completely built within a code block.

You can (and should) still add inline Python comments to explain the steps. This will mean that the figure actually displays as intended with the slider in the output.

Cadair:
Also, good luck writing something equivalent to WCSAxes in either bokeh or plotly... :stuck_out_tongue_closed_eyes:

dstansby:
I did combine them, so now the example looks like this: https://23569-2165383-gh.circle-artifacts.com/0/home/circleci/project/docs/_build/html/generated/gallery/plotting/fading_between_maps.html - I think this is fine, just needs some axes placement tweaking.

Cadair:
I wish the API to do this with `sunpy.visualization.animator` was cleaner, it would do all the layout for you, but you have to subclass the thing.

examples/plotting/fading_between_maps.py
```diff
@@ -0,0 +1,59 @@
...
+
+import sunpy.map
+from sunpy.data.sample import AIA_171_IMAGE, AIA_1600_IMAGE
+
+###############################################################################
+# Start by loading two AIA maps from the sample data.
+#
+# Note that these maps must be co-aligned. If they are taken at different times
+# or from different observation locations they will need to be aligned using
+# rotate() or reproject().
```
Cadair:
```suggestion
# `~sunpy.map.GenericMap.rotate` or `reproject`.
```

examples/plotting/fading_between_maps.py
```diff
@@ -0,0 +1,59 @@
+"""
+Fading between two maps
```
nabobalis:
```suggestion
=======================
Fading between two maps
```

examples/plotting/fading_between_maps.py
```diff
@@ -0,0 +1,59 @@
...
+
+import sunpy.map
+from sunpy.data.sample import AIA_171_IMAGE, AIA_1600_IMAGE
+
+###############################################################################
+# Start by loading two AIA maps from the sample data.
+#
+# Note that these maps must be co-aligned. If they are taken at different times
+# or from different observation locations they will need to be aligned using
+# `~sunpy.map.GenericMap.rotate` or `reproject`.
```
nabobalis: Does reproject link now?

examples/plotting/fading_between_maps.py
```diff
@@ -0,0 +1,59 @@
...
+map_171 = sunpy.map.Map(AIA_171_IMAGE)
```
nabobalis:
```suggestion

map_171 = sunpy.map.Map(AIA_171_IMAGE)
```

examples/plotting/fading_between_maps.py
```diff
@@ -0,0 +1,59 @@
...
+map_1600 = sunpy.map.Map(AIA_1600_IMAGE)
+
+###############################################################################
+# First we create a figure, and add the axes that will show the maps. Then
+# plot both of the images on the same axes. Finally, we add another axes that
+# contains the slider.
+fig = plt.figure()
```
nabobalis:
```suggestion

fig = plt.figure()
```

examples/plotting/fading_between_maps.py
```diff
@@ -0,0 +1,59 @@
...
+# Add the slider axes
+ax_slider = fig.add_axes([0.25, 0.05, 0.65, 0.03])
+slider = Slider(ax_slider, 'Alpha', 0, 1, valinit=0.5)
+
+
+###############################################################################
+# Finally, define what happens when the slider is changed and link this to the
+# slider we set up above. In this case we just set the alpha (ie. transparencey)
+# of the 171 image.
+def update(val):
```
nabobalis:
```suggestion

def update(val):
```

examples/plotting/fading_between_maps.py
Cadair: It will post merge of the reproject example PR

PR #3489, opened 2019-11-10
Document some kwargs in mapsaving methods

dstansby:
Should probably check that the intersphinx links work properly here.


dstansby:
🎉 the links work.

PR #3491, opened 2019-11-11
Add some missing doc requirements to conf.py

dstansby:
Building the docs in a new conda environment, these are the packages that I had to install to build wihtout errors. There must be a better way to do this than all of these try except blocks...

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,6 +1,3 @@
-# -*- coding: utf-8 -*-
-# Licensed under a 3-clause BSD style license - see LICENSE.rst
-#
 # SunPy documentation build configuration file.
 #
 # This file is execfile()d with the current directory set to its containing dir.
@@ -25,6 +22,12 @@
 # Thus, any C-extensions that are needed to build the documentation will *not*
 # be accessible, and the documentation will not build correctly.
 
+# ITS THE WILD WEST IN HERE
+# flake8: noqa
+
+import sunpy.data.sample
+import numpy as np
+from pkg_resources import get_distribution
 import os
 import sys
 import pathlib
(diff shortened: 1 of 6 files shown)


Cadair:
Do we need to be doing this at all? The way we should be recommending people to build the docs is `tox -e build_docs` and then we don't have to worry about this.

nabobalis:
> Building the docs in a new conda environment, these are the packages that I had to install to build wihtout errors. There must be a better way to do this than all of these try except blocks...

The answer is use tox if you don't want to install them in your local env. But I think we should add them since we check for other deps too.

dstansby:
Oh, maybe not - out of long time habit I've been doing `cd docs; make html` - if the way to build should be tox, then I think running the above command should error out.

nabobalis:
> Oh, maybe not - out of long time habit I've been doing `cd docs; make html` - if the way to build should be tox, then I think running the above command should error out.

I'ts why I deleted the makefile ages ago before finally adding it back.

Cadair:
It seems weird to break your workflow (with `make html`) just because you haven't installed the deps (which will just error out at some point right?). I assume we haven't documented doing this anywhere?

Cadair:
At the very least we should turn these whole bunch of try except statements into a loop over a list of deps or something.

docs/conf.py
```diff
@@ -82,6 +82,27 @@
           'the "glymur" package to be installed')
     sys.exit(1)
 
+try:
+    import sqlalchemy
+except ImportError:
+    print('ERROR: sqlalchemy could not be imported. Building the documentation requires '
```
wtbarnes: Rather than using `print` and `sys.exit`, I think it'd be better to catch and reraise the `ImportError` with a custom message, e.g.

```python
try:
    import sqlalchemy
except ImportError as e:
    raise Exception(e, 'ERROR: sqlalchemy could not be imported. ...')
```

nabobalis: Sold.

nabobalis:
> At the very least we should turn these whole bunch of try except statements into a loop over a list of deps or something.

These deps in the setup.cfg we can do that?  Have to check if the new docs ones are there too tho.

nabobalis:
> It seems weird to break your workflow (with `make html`) just because you haven't installed the deps (which will just error out at some point right?). I assume we haven't documented doing this anywhere?

Doing what? running make html?

wtbarnes:
So what is the recommended way to build the docs? I was under the impression that it was with,
`python setup.py build_docs` since this is what we recommend in [the dev guide](https://docs.sunpy.org/en/latest/dev_guide/documentation.html#usage). Is this not the case anymore?

nabobalis:
> So what is the recommended way to build the docs? I was under the impression that it was with,
> `python setup.py build_docs` since this is what we recommend in [the dev guide](https://docs.sunpy.org/en/latest/dev_guide/documentation.html#usage). Is this not the case anymore?

We never formally moved to the use tox part. We planned it as an experiment that eventually we say is the only officially supported way. 

As sooner or later, `python setup.py build_docs` will not work. (Let alone when we delete setup.py.)

Cadair:
We should remove all mentions of `setup.py xxxx` from the dev docs before 1.1 I would like to ditch astropy_helpers for 2.0.

nabobalis:
I changed the docs to account for that.

docs/dev_guide/documentation.rst
```diff
@@ -106,17 +106,19 @@ To build the documentation locally you must have all the dependencies (``pip ins
...
-If you'd like to start from scratch then just delete the build directory or run::
+If you'd like to start from scratch then just delete the tox build directory::
 
-    $ python setup.py build_docs -l
+    $ rm .tox/build_docs/
 
-to clean previous builds before building new ones.
+or::
+
+    $ tox --recreate -e build_docs
```
Cadair: `tox --recreate -e build_docs -- -aE`

Issue #3448, opened 2019-10-25
Reduce import time for `sunpy.map`


ayshih:
(I was getting annoyed by our import times, and was inspired by astropy/astropy#4598 to look deeper.)

Oof.  Importing `sunpy.map` takes a whopping 6+ seconds on my machine (see below).  There are a lot of contributing factors (see #3445, #3446, #3447), but what sticks out like a sore thumb is that 1.6 seconds (25% of the total time) is spent importing `sunpy.database` (which then imports `sunpy.net`) so that `map_factory` can handle `DatabaseEntry`.  Can that be re-factored so `map_factory` learns about `DatabaseEntry` only after a user has intentionally imported `sunpy.database`?

```
python -X importtime -c "import sunpy.map" 2> map.log
tuna map.log
```
![map](https://user-images.githubusercontent.com/991759/67576555-c6e3d900-f70c-11e9-9841-8ba77164ffd0.png)

Cadair:
Also for `database` we only need the type to know how to read it, perhaps we can refactor `database` so we import less to know about the type.

Cadair:
Closed by #3495??

nabobalis:
Yes.

PR #3495, opened 2019-11-12
Reduce sunpy.map import time

dstansby:
xref #3448. This reduces import time from 3.8s to 3.0s for me locally.

### DIFF
--- a/changelog/3495.feature.rst
+++ b/changelog/3495.feature.rst
@@ -1,0 +1,2 @@
+Made a couple of module imports lazy to reduce the import time of sunpy.map by
+~20%.
--- a/changelog/3495.feature.rst
+++ b/changelog/3495.feature.rst
@@ -1,2 +1,2 @@
 Made a couple of module imports lazy to reduce the import time of sunpy.map by
-~20%.
+~40%.
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -8,7 +8,7 @@ import textwrap
 
 import numpy as np
 import matplotlib.pyplot as plt
-from matplotlib import patches, cm, colors
+from matplotlib import cm
 
 import astropy.wcs
 import astropy.units as u
@@ -25,7 +25,6 @@ from sunpy.visualization import wcsaxes_compat, axis_labels_from_ctype, peek_sho
 from sunpy.sun import constants
 from sunpy.coordinates import sun
 from sunpy.time import parse_time, is_time
-from sunpy.image.transform import affine_transform
 from sunpy.image.resample import reshape_image_to_4d_superpixel
 from sunpy.image.resample import resample as sunpy_image_resample
 from sunpy.coordinates import get_earth
@@ -199,6 +198,8 @@ class GenericMap(NDData):
         if self.dtype == np.uint8:
             norm = None
         else:
+            # Put import here to reduce sunpy.map import time
+            from matplotlib import colors
             norm = colors.Normalize()
 
         # Visualization attributes
@@ -1026,6 +1027,9 @@ class GenericMap(NDData):
         transformations, situations when the underlying data is modified prior
         to rotation, and differences from IDL's rot().
         """
+        # Put the import here to reduce sunpy.map import time
+        from sunpy.image.transform import affine_transform
+
         if angle is not None and rmatrix is not None:
             raise ValueError("You cannot specify both an angle and a matrix")
         elif angle is None and rmatrix is None:
@@ -1485,6 +1489,8 @@ class GenericMap(NDData):
         -----
         Keyword arguments are passed onto `matplotlib.patches.Circle`.
         """
+        # Put import here to reduce sunpy.map import time
+        from matplotlib import patches
 
         if not axes:
             axes = wcsaxes_compat.gca_wcs(self.wcs)
--- a/sunpy/util/types.py
+++ b/sunpy/util/types.py
@@ -3,7 +3,8 @@ A collection of ABCs to use to shortcut having to import things.
 """
 import abc
 
+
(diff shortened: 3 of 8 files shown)


nabobalis:
Any more changes you want to make?

dstansby:
These are the only simple ones I can see. Will add a changelog in a bit.

nabobalis:
> These are the only simple ones I can see. Will add a changelog in a bit.

How much effort is to to remake the import time figure?

Cadair:
I wouldn't backport this personally, mainly as we are so close to 1.1. We can have "Import Performance Improvements" in the 1.1 changelog :grin: 

I am still unsure if I like imports in functions, but it's worth it for a 20% import time improvement.

sunpy/map/mapbase.py
```diff
@@ -25,7 +25,6 @@
 from sunpy.sun import constants
 from sunpy.coordinates import sun
 from sunpy.time import parse_time, is_time
-from sunpy.image.transform import affine_transform
 from sunpy.image.resample import reshape_image_to_4d_superpixel
```
Cadair: Do these imports not pull in `affine_transform` anyway?

dstansby: Nope; what moving this import really does is avoid having to import skimage.

nabobalis:
> I wouldn't backport this personally, mainly as we are so close to 1.1. We can have "Import Performance Improvements" in the 1.1 changelog 
> 
> I am still unsure if I like imports in functions, but it's worth it for a 20% import time improvement.

Fine. I removed the backport.

I also can't say I am a fan of imports in functions but since the import time is slow, it's a sacrifice worth making.

sunpy/map/mapbase.py
Cadair: I figured, I was just surprised.

dstansby:
> > These are the only simple ones I can see. Will add a changelog in a bit.
> 
> How much effort is to to remake the import time figure?

Very little, here's the new one:

Screenshot 2019-11-12 at 16 12 46

nabobalis:
Hmm. Could be other places to hit but nothing simple.

dstansby:
Yes, quite a lot of the potential places for improvements (e.g. pandas, matplotlib) need to be imported to register various external classes with an internal SunPy registry (e.g. for time conversion), and I'm not sure if there's an way to get around that.

Cadair:
I got this down to sub 1s on Python 3.8

Screen Shot 2019-11-12 at 16 33 16

It does have one proper weird work around though, which I will open as a PR to this branch, because it's weird.

nabobalis:
What on earth did you do?

Cadair:
https://github.com/dstansby/sunpy/pull/2

I don't think the times are comparable, I think my PC might just be a bit faster than Davids. Also it's a very clean Python 3.8 env. I only ran (`pip install -e .[all]`).

Cadair:
For comparison without the patch I get this:
Screen Shot 2019-11-12 at 16 40 34

nabobalis:
> [dstansby#2](https://github.com/dstansby/sunpy/pull/2)
> 
> I don't think the times are comparable, I think my PC might just be a bit faster than Davids. Also it's a very clean Python 3.8 env. I only ran (`pip install -e .[all]`).

Isn't all everything but testing and docs deps? I would have thought it would import all of the optional deps?

Cadair:
The only thing is `dask` which we import if it's installed. Otherwise I think it will be ok.

nabobalis:
Uh oh.

sunpy/util/types.py
```diff
@@ -0,0 +1,9 @@
+"""
+A collection of ABCs to use to shortcut having to import things.
+"""
+import abc
+
```
Cadair:
```suggestion


```

sunpy/util/types.py
```diff
@@ -0,0 +1,9 @@
...
+class DatabaseEntryType(metaclass=abc.ABCMeta):
+    """
+    A Base class for sunpy.database.DatabaseEntry.
```
Cadair:
```suggestion
    A base class for `sunpy.database.tables.DatabaseEntry`.
```

sunpy/util/types.py
nabobalis: You don't need to import abc for metaclasses to work?

Cadair: The suggestion is adding an extra line for the pep8 checker.

nabobalis:
The ABC patch broke the entire suite. 

```
class DatabaseEntry(DatabaseEntryType, Base):

TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

Not a clue how to fix it.

changelog/3495.feature.rst
```diff
@@ -0,0 +1,2 @@
+Made a couple of module imports lazy to reduce the import time of sunpy.map by
+~20%.
```
ayshih: On my machine, the total time improvement is now ~40% (from ~6 seconds to ~3.5 seconds).

sunpy/util/types.py
```diff
@@ -0,0 +1,10 @@
+"""
+A collection of ABCs to use to shortcut having to import things.
```
ayshih: I think this docstring should be a little more descriptive, including elaborating on "things".

changelog/3495.feature.rst
nabobalis: Go with the worst case? Or we just remove the % and say its just faster.

ayshih: Oh, I believe the "~20%" was estimated prior to the `database` improvement, so I suspect that ~40% is more accurate for everyone.

I like having a value in the changelog entry.

PR #3497, opened 2019-11-12
Do not assume the type of header entries

### LINKED ISSUE #3496: PSP WISPR data doesn't work out of the box with SunPy
This is probably because the `.fits` files are weird, and have an integer for the detector (?!?!?!), but putting it here in case anyone else comes across the same issue.

```python
import sunpy.map

instr = a.Instrument('WISPR')
time = a.Time('2018-10-31', '2018-11-01')
result = Fido.search(instr, time)

files = Fido.fetch(result)

wispr = sunpy.map.Map(files[0])
```

```python
----> 1 wispr = sunpy.map.Map(files[0])

~/github/sunpy/sunpy/map/map_factory.py in __call__(self, composite, sequence, silence_errors, *args, **kwargs)
    292 
    293             try:
--> 294                 new_map = self._check_registered_widgets(data, meta, **kwargs)
    295                 new_maps.append(new_map)
    296             except (NoMatchError, MultipleMatchError, ValidationFunctionError):

~/github/sunpy/sunpy/map/map_factory.py in _check_registered_widgets(self, data, meta, **kwargs)
    327 
    328             # Call the registered validation function for each registered class
--> 329             if self.registry[key](data, meta, **kwargs):
    330                 candidate_widget_types.append(key)
    331 

~/github/sunpy/sunpy/map/sources/stereo.py in is_datasource_for(cls, data, header, **kwargs)
    121     def is_datasource_for(cls, data, header, **kwargs):
    122         """Determines if header corresponds to an COR image"""
--> 123         return header.get('detector', '').startswith('COR')
    124 
    125 

AttributeError: 'int' object has no attribute 'startswith'
```

Cadair:
Don't ever assume that a FITS key is a string, **especially** not in `is_datasource_for`.

Fixes #3496

### DIFF
--- a/changelog/3497.bugfix.rst
+++ b/changelog/3497.bugfix.rst
@@ -1,0 +1,2 @@
+Fix multiple instances of `sunpy.map.sources` assuming the type of FITS Header
+values.
--- a/docs/code_ref/map.rst
+++ b/docs/code_ref/map.rst
@@ -117,7 +117,7 @@ demonstrated by the following example.
        @classmethod
        def is_datasource_for(cls, data, header, **kwargs):
             """Determines if header corresponds to an AIA image"""
-            return header.get('instrume', '').startswith('FUTURESCOPE')
+            return str(header.get('instrume', '')).startswith('FUTURESCOPE')
 
 
 This class will now be available through the `Map <sunpy.map.map_factory.MapFactory>` factory as long as this
--- a/sunpy/map/sources/iris.py
+++ b/sunpy/map/sources/iris.py
@@ -49,7 +49,7 @@ class SJIMap(GenericMap):
     @classmethod
     def is_datasource_for(cls, data, header, **kwargs):
         """Determines if header corresponds to an IRIS SJI image"""
-        tele = header.get('TELESCOP', '').startswith('IRIS')
-        obs = header.get('INSTRUME', '').startswith('SJI')
+        tele = str(header.get('TELESCOP', '')).startswith('IRIS')
+        obs = str(header.get('INSTRUME', '')).startswith('SJI')
         level = header.get('lvl_num') == 1
         return tele and obs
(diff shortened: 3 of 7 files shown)


Cadair:
I am unconvinced that we should backport the fits header change to 1.0. Opinions on how much people are going to want to load WISPR data without upgrading to 1.1?

sunpy/map/mapbase.py
```diff
@@ -265,7 +265,15 @@ def wcs(self):
...
         The `~astropy.wcs.WCS` property of the map.
         """
-        w2 = astropy.wcs.WCS(naxis=2)
+        # Construct the WCS based on the FITS header, but don't "do_set" which
+        # analyses the FITS header for correctness.
+        w2 = astropy.wcs.WCS(header=self.fits_header, _do_set=False)
+
+        # If the FITS header is > 2D pick the first 2 and move on.
+        # This will require the FITS header to be valid.
+        if w2.naxis != 2:
```
wtbarnes: To silently do this seems a bit...questionable. At the very least, shouldn't we tell the user that the higher dimensions are being skipped?

Cadair: I thought we already did that somewhere on map load?

sunpy/map/mapbase.py
```diff
@@ -265,7 +265,15 @@ def wcs(self):
         """
         The `~astropy.wcs.WCS` property of the map.
         """
-        w2 = astropy.wcs.WCS(naxis=2)
+        # Construct the WCS based on the FITS header, but don't "do_set" which
+        # analyses the FITS header for correctness.
+        w2 = astropy.wcs.WCS(header=self.fits_header, _do_set=False)
```
dstansby: This seems to be changing the way the `wcs` is computed quite substantially, ie. now it does it from the header instead of from just the map properties (below)

nabobalis:
I thought the str header was gonna be backported?

Cadair:
I have moved all the other changes out into #3501

nabobalis:
I want to know how badly the backport to 0.9 goes.

PR #3501, opened 2019-11-13
Make Map work with WISPR Data

### LINKED ISSUE #3496: PSP WISPR data doesn't work out of the box with SunPy

Cadair:
Fixes #3496 

This makes the following changes:

* Change `GenericMap.wcs` to use the raw FITS header, this is so we can get the extra parameters from the FITS headers for the unusual projections in WISPR data. This is a *large* change in approach, although I am reasonably confident it shouldn't break anything.
* Fix IRIS so that the above change dosen't completely destroy it.
* Fix some parfive incompatibility issues in `vso.py`

requires #3497  to actually make WISPR data load.

### DIFF
--- a/changelog/3501.feature.rst
+++ b/changelog/3501.feature.rst
@@ -1,1 +1,5 @@
-Sunpy Maps now can read WISPR Data.
+`sunpy.map.GenericMap.wcs` now uses the full FITS header to construct the WCS.
+This adds support for instruments with more complex projections, such as WISPR,
+however does mean that Map will be more sensative to incorrect or invalid FITS
+headers. If you are using custom headers with SunPy Map you might encounter
+issues relating to this change.
--- a/changelog/3501.feature.rst
+++ b/changelog/3501.feature.rst
@@ -1,0 +1,1 @@
+Sunpy Maps now can read WISPR Data.
--- a/changelog/3501.feature.rst
+++ b/changelog/3501.feature.rst
@@ -1,5 +1,5 @@
 `sunpy.map.GenericMap.wcs` now uses the full FITS header to construct the WCS.
 This adds support for instruments with more complex projections, such as WISPR,
-however does mean that Map will be more sensative to incorrect or invalid FITS
+however does mean that Map will be more sensitive to incorrect or invalid FITS
 headers. If you are using custom headers with SunPy Map you might encounter
 issues relating to this change.
--- a/sunpy/io/fits.py
+++ b/sunpy/io/fits.py
@@ -200,24 +200,25 @@ def header_to_fits(header):
     key_comments = header.pop('KEYCOMMENTS', False)
 
     for k, v in header.items():
-        if isinstance(v, fits.header._HeaderCommentaryCards):
-            if k.upper() == 'COMMENT':
-                comments = str(v).split('\n')
-                for com in comments:
-                    fits_header.add_comment(com)
-            elif k.upper() == 'HISTORY':
-                hists = str(v).split('\n')
-                for hist in hists:
-                    fits_header.add_history(hist)
-            elif k != '':
+        if len(k) > 8:
+            warnings.warn(f"The meta key {k} is too long, dropping from the FITS header.", SunpyUserWarning)
+            continue
+        if k.upper() in ('COMMENT', 'HV_COMMENT'):
+            comments = str(v).split('\n')
+            for com in comments:
+                fits_header.add_comment(com)
+        elif k.upper() == 'HISTORY':
+            hists = str(v).split('\n')
+            for hist in hists:
+                fits_header.add_history(hist)
+        elif isinstance(v, fits.header._HeaderCommentaryCards):
+            if k != '':
                 fits_header.append(fits.Card(k, str(v).split('\n')))
-
-        # For some horrific reason, we save a list to the wavelnth key in
-        # sources/rhessi.py. This is the least invasive fix for that stupidity.
-        elif isinstance(v, list):
-            v = str(v)
-
         else:
+            # For some horrific reason, we save a list to the wavelnth key in
+            # sources/rhessi.py. This is the least invasive fix for that stupidity.
+            if isinstance(v, list):
+                v = str(v)
             fits_header.append(fits.Card(k, v))
 
     if isinstance(key_comments, dict):
--- a/sunpy/io/fits.py
+++ b/sunpy/io/fits.py
@@ -212,6 +212,11 @@ def header_to_fits(header):
             elif k != '':
                 fits_header.append(fits.Card(k, str(v).split('\n')))
 
+        # For some horrific reason, we save a list to the wavelnth key in
+        # sources/rhessi.py. This is the least invasive fix for that stupidity.
+        elif isinstance(v, list):
+            v = str(v)
+
         else:
             fits_header.append(fits.Card(k, v))
 
--- a/sunpy/io/fits.py
+++ b/sunpy/io/fits.py
@@ -200,6 +200,7 @@ def header_to_fits(header):
     key_comments = header.pop('KEYCOMMENTS', False)
 
     for k, v in header.items():
+        # Drop any keys which are too long to save into FITS
         if len(k) > 8:
             warnings.warn(f"The meta key {k} is too long, dropping from the FITS header.", SunpyUserWarning)
             continue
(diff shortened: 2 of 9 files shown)


sunpy/map/sources/iris.py
```diff
@@ -40,12 +40,17 @@ def __init__(self, data, header, **kwargs):
...
-        GenericMap.__init__(self, data, header, **kwargs)
+        super().__init__(data, header, **kwargs)
 
         self.meta['detector'] = "SJI"
         self.meta['waveunit'] = "Angstrom"
         self.meta['wavelnth'] = header['twave1']
 
+        # We have to change this or else the WCS doesn't parse properly, even
+        # though we don't care about the third dimension.
+        if self.meta['cdelt3'] == 0:
```
Cadair:
```suggestion
        if self.meta.get('cdelt3', None) == 0:
```

wafels:
This

https://github.com/sunpy/sunpy/blob/master/sunpy/map/tests/test_mapbase.py#L101

looks off.  Should this be

`assert all(wcs.wcs.cdelt == [aia171_test_map.scale.axis1.value, aia171_test_map.scale.axis2.value])`

?

nabobalis:
Ideally want a test with a new test file.

nabobalis:
> This
> 
> https://github.com/sunpy/sunpy/blob/master/sunpy/map/tests/test_mapbase.py#L101
> 
> looks off. Should this be
> 
> `assert all(wcs.wcs.cdelt == [aia171_test_map.scale.axis1.value, aia171_test_map.scale.axis2.value])`
> 
> ?

Changed it.

sunpy/map/mapbase.py
```diff
@@ -266,7 +266,15 @@ def wcs(self):
         """
         The `~astropy.wcs.WCS` property of the map.
         """
-        w2 = astropy.wcs.WCS(naxis=2)
+        # Construct the WCS based on the FITS header, but don't "do_set" which
+        # analyses the FITS header for correctness.
+        w2 = astropy.wcs.WCS(header=self.fits_header, _do_set=False)
```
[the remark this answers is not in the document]
nabobalis: The keyword is set and documented in their source code.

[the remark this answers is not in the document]
Cadair: I also don't think we have a choice. We *do not* want to validate the header at this point, as we are later on fixing the header based on our metadata mapping. While I agree this whole setup isn't great, the alternative is about 6 months of work on completely refactoring how we represent metadata... (which I do want to do)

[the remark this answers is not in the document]
dstansby: Okay, I guess we should convince astropy to remove the underscore if it really is intended as a user facing kwarg. I'll open an issue over there.

sunpy/map/sources/rhessi.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, data, header, **kwargs):
             self.meta['ctype2'] = 'HPLT-TAN'
 
         self.meta['waveunit'] = 'keV'
-        self.meta['wavelnth'] = [self.meta['energy_l'], self.meta['energy_h']]
+        self.meta['wavelnth'] = f"{self.meta['energy_l']}-{self.meta['energy_h']}"
```
Cadair: I have no idea if this is a good idea. It's definitely a breaking change, it's been in SunPy for 5 years, on the other hand it means you can now actually save a RHESSI map to FITS.

@ehsteve this was your fault, if you want to weigh in here? :stuck_out_tongue:

Cadair: Ok, I changed this fix, so that we only fix it on save to FITS. Everything is still terrible but at least we don't break this.

sunpy/io/fits.py
```diff
@@ -200,19 +200,25 @@ def header_to_fits(header):
...
-            if k.upper() == 'COMMENT':
-                comments = str(v).split('\n')
-                for com in comments:
-                    fits_header.add_comment(com)
-            elif k.upper() == 'HISTORY':
-                hists = str(v).split('\n')
-                for hist in hists:
-                    fits_header.add_history(hist)
-            elif k != '':
+        if len(k) > 8:
```
nabobalis: Lol?

sunpy/io/fits.py
```diff
@@ -200,19 +200,25 @@ def header_to_fits(header):
...
+            hists = str(v).split('\n')
+            for hist in hists:
+                fits_header.add_history(hist)
+        elif isinstance(v, fits.header._HeaderCommentaryCards):
+            if k != '':
                 fits_header.append(fits.Card(k, str(v).split('\n')))
-
         else:
+            # For some horrific reason, we save a list to the wavelnth key in
+            # sources/rhessi.py. This is the least invasive fix for that stupidity.
```
nabobalis: LOL

sunpy/io/fits.py
dstansby: Add a comment explaining the (currently) seemingly arbitrary 8?

sunpy/io/tests/test_fits.py
```diff
@@ -124,3 +125,14 @@ def test_write_with_metadict_header_astropy(tmpdir):
     temp_file = tmpdir / "temp.fits"
     sunpy.io.fits.write(str(temp_file), data, meta_header)
     assert temp_file.exists()
+
+
+def test_fitsheader():
+    extensions = ('fts', 'fits')
```
dstansby: Add a short comment here explaining what this test is doing? (Testing all the sample data parses???)

sunpy/net/vso/vso.py
```diff
@@ -793,7 +793,7 @@ def download_all(self, response, methods, downloader, path, qr, info=None):
                             qr[dataitem.fileiditem.fileid[0]]
                         )
                     except NoData:
-                        results.add_error('', DownloadFailed(dresponse))
+                        results.add_error('', '', DownloadFailed(dresponse))
```
dstansby: What are all these here for?

sunpy/io/fits.py
Cadair: We are saving to FITS headers, you can't have keys in a FITS header longer than 8.

sunpy/net/vso/vso.py
Cadair: There was an API change in parfive that broke this code.

sunpy/io/fits.py
Cadair:
```suggestion
        # Drop any keys which are too long to save into FITS
        if len(k) > 8:
```

sunpy/io/tests/test_fits.py
Cadair:
```suggestion
    """Test that all test data can be converted back to a FITS header."""
    extensions = ('fts', 'fits')
```

changelog/3501.feature.rst
```diff
@@ -0,0 +1,5 @@
+`sunpy.map.GenericMap.wcs` now uses the full FITS header to construct the WCS.
+This adds support for instruments with more complex projections, such as WISPR,
+however does mean that Map will be more sensative to incorrect or invalid FITS
```
Cadair:
```suggestion
however does mean that Map will be more sensitive to incorrect or invalid FITS
```

PR #3515, opened 2019-11-18
edited documentation to reflect that the threshold is being made at 10% of max intensity

### LINKED ISSUE #3503: Error in documentation for "Finding bright regions with ndimage" example.
### Description

There seems to be an error in the documentation for the "Finding bright regions with ndimage" example.

In the part where a mask is made, the surrounding text states: " We choose the criterion that the data should be at least 5% of the maximum value. " However, if you look at the code immediately below, the threshold is based off 10% the max value:
`mask = aiamap.data < aiamap.max() * 0.10`

### Expected behavior

Documentation needs to be modified to reflect that the threshold is based off a 10% threshold.

neerajkulk:
trivial change to make the documentation consistent with the code

Fixes #3503

### DIFF
--- a/examples/map/image_bright_regions_gallery_example.py
+++ b/examples/map/image_bright_regions_gallery_example.py
@@ -22,7 +22,7 @@ aiamap = sunpy.map.Map(AIA_193_IMAGE)
 
 ##############################################################################
 # First we make a mask, which tells us which regions are bright. We
-# choose the criterion that the data should be at least 5% of the maximum
+# choose the criterion that the data should be at least 10% of the maximum
 # value. Pixels with intensity values greater than this are included in the
 # mask, while all other pixels are excluded.
 mask = aiamap.data < aiamap.max() * 0.10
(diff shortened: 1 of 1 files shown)


nabobalis:
Thanks @neerajkulk.

nabobalis:
Doc build passed and the example has updated. Merged.

nabobalis:
The bot did not run. So a manual backport is in order.

neerajkulk:
I don't see the change reflected in the [example](https://docs.sunpy.org/en/stable/generated/gallery/map/image_bright_regions_gallery_example.html#sphx-glr-generated-gallery-map-image-bright-regions-gallery-example-py). Is there somewhere else I need to make a change?

wtbarnes:
@neerajkulk I think you want to be looking at the "latest" version of the docs on Read the Docs: https://docs.sunpy.org/en/latest/generated/gallery/map/image_bright_regions_gallery_example.html. The "stable" version does not contain your changes because we've not done a release since your fix was merged. Once that's done, this change will be propagated up to the "stable" version.

neerajkulk:
got it! Thanks

PR #3517, opened 2019-11-20
Fixes to `NorthOffsetFrame`

ayshih:
- `NorthOffsetFrame` now accepts a north pole specified in Cartesian coordinates (or actually, any representation)
- Cleaned up docstring and fixed an error message


nabobalis:
Online failures we need to track down but unrelated to this PR.

PR #3525, opened 2019-11-20
Add reproject to rtd env

Cadair:


dstansby:
Lets merge and see if it fixes things: https://readthedocs.org/projects/sunpy/builds/

Cadair:
It did but:
```
generating gallery for generated/gallery/sunpy_other_packages... [100%] reprojection_aia_euvi_mosaic.py
Killed


Command killed due to excessive memory consumption
```
whoops!

wtbarnes:
Same https://readthedocs.org/projects/aiapy/builds/9994608/. 

Will RTD increase the memory limit on the container?

Cadair:
Probably if you ask nicely.

dstansby:
It looks like the limit is 1GB RAM - the other thing to check would be that memory isn't building up in the background from previous examples for some reason.

wtbarnes:
We're using full resolution, full disk images in the aiapy examples so I guess this is not too surprising. I've sent a nice email so we'll see what they say.

PR #3528, opened 2019-11-21
Reduce memory consumption of aiai_euvi_mosaic

Cadair:


examples/sunpy_other_packages/reprojection_aia_euvi_mosaic.py
```diff
@@ -79,14 +85,16 @@
 ax.set_rlim(0, 1.3)
 
 plt.legend()
+plt.show()
```
nabobalis: I might have removed these from more of the reproduction examples. I thought one show at the end was enough.

examples/sunpy_other_packages/reprojection_aia_euvi_mosaic.py
```diff
@@ -48,6 +48,12 @@
 
 maps = sunpy.map.Map(sorted(files))
 
+######################################################################
+# To reduce memory consumption we also downsample these maps before continuing,
+# you can disable this.
+
+maps = [m.resample((1024, 1024)*u.pix) for m in maps]
```
nabobalis: Is it this just this examples we need to do this for?

Cadair: The main mem issues here seem to be inside reproject's coadd function so I hope it's only this one.

nabobalis:
Did we not backport the reproject stuff?

Cadair:
We did not, they use a few small fixes on master.

PR #3533, opened 2019-11-21
Use exact rotation to transform from Carrington > Carrington frames

### LINKED ISSUE #3530: HGC > HGC transformation with different frame times has non-zero change in lattitude
### Description
When transforming from one HeliographicCarrington (HGC) frame, to another but with a different obstime, a non-zero change in latitude seems to be observerd. I would expect exactly zero change in latitude, since the difference between two HCG frames at different times is just a change in longitude.

### Steps to Reproduce
```python
from astropy.time import Time
import astropy.units as u

t1 = Time('2019-01-01')
t2 = Time('2019-01-06')

frame1 = frames.HeliographicCarrington(obstime=t1)
coord1 = SkyCoord(0*u.deg, 0*u.deg, frame=frame1)
print(coord1)
# <SkyCoord (HeliographicCarrington: obstime=2019-01-01 00:00:00.000): (lon, lat, radius) in (deg, deg, km)
    (0., 0., 695700.)>

frame2 = frames.HeliographicCarrington(obstime=t2)
coord2 = coord1.transform_to(frame2)
print(coord2)
# <SkyCoord (HeliographicCarrington: obstime=2019-01-06 00:00:00.000): (lon, lat, radius) in (deg, deg, km)
    (288.58265103, -0.05073824, 696211.1121367)>
```

Note how the latitude at the end != 0.

### System Details

  - SunPy Version: 1.0.5
  - Astropy Version: 3.2.3
  - Python Version: 3.7
  - OS information: MacOS 10.15

dstansby:
Fixes #3530.

This introduces an exact rotation matrix transformation for going between two HeliographicCarrington frames, which removes the inaccuracies of going through various other transformations.

This allows the accuracy of the latitude and radius of the transformation in the tests to be beaten down to zero, as it should be.

### DIFF
--- a/sunpy/coordinates/tests/test_transformations.py
+++ b/sunpy/coordinates/tests/test_transformations.py
@@ -331,9 +331,10 @@ def test_hgc_hgc():
     old = SkyCoord(90*u.deg, 10*u.deg, 1*u.AU, frame=HeliographicCarrington(obstime=obstime))
     new = old.transform_to(HeliographicCarrington(obstime=obstime + 1*u.day))
 
-    assert_quantity_allclose(new.lon, old.lon - 14.1844*u.deg, atol=1e-4*u.deg)  # solar rotation
-    assert_quantity_allclose(new.lat, old.lat, atol=1e-4*u.deg)
-    assert_quantity_allclose(new.radius, old.radius, atol=1e-5*u.AU)
+    assert_quantity_allclose(new.lon, 76.82974*u.deg, atol=1e-4*u.deg)  # solar rotation
+    # Lattitude and radius shouldn't change
+    assert_quantity_allclose(new.lat, old.lat, atol=1e-10*u.deg)
+    assert_quantity_allclose(new.radius, old.radius, atol=1e-10*u.R_sun)
 
 
 def test_hcc_hcc():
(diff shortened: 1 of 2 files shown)


sunpy/coordinates/tests/test_transformations.py
```diff
@@ -331,9 +331,10 @@ def test_hgc_hgc():
     old = SkyCoord(90*u.deg, 10*u.deg, 1*u.AU, frame=HeliographicCarrington(obstime=obstime))
     new = old.transform_to(HeliographicCarrington(obstime=obstime + 1*u.day))
 
-    assert_quantity_allclose(new.lon, old.lon - 14.1844*u.deg, atol=1e-4*u.deg)  # solar rotation
-    assert_quantity_allclose(new.lat, old.lat, atol=1e-4*u.deg)
-    assert_quantity_allclose(new.radius, old.radius, atol=1e-5*u.AU)
+    assert_quantity_allclose(new.lon, 76.82974*u.deg, atol=1e-4*u.deg)  # solar rotation
```
dstansby: I think this changes, because the change in longitude is not (and should not) be exactly `14.1844*u.deg`, but this should be checked.

ayshih:
I'm closing this PR because implementing this transformation as an exact rotation violates the SkyCoord being a point in inertial space.  Continue discussion in #3530 if desired.

PR #3538, opened 2019-11-24
Fixed wrong header being used in "AIA as Seen from Mars" example

ayshih:
NT


nabobalis:
The bot seems to be dead.

Issue #3532, opened 2019-11-21
HGS > HGS transformation with different frame times has non-zero change in distance


dstansby:
### Description
When transforming from one HeliographicStonyhurst (HGS) frame, to another but with a different obstime, a non-zero change in radius happens. I would expect exactly zero change in radius, since the difference between two HGS frames is just a rotation about the center of the Sun.

### Steps to reproduce
```python
from sunpy.coordinates import frames
from astropy.coordinates import SkyCoord
from astropy.time import Time
import astropy.units as u

t1 = Time('2019-01-01')
t2 = Time('2019-01-06')

frame1 = frames.HeliographicStonyhurst(obstime=t1)
coord1 = SkyCoord(0*u.deg, 0*u.deg, frame=frame1)
print(coord1)
# <SkyCoord (HeliographicStonyhurst: obstime=2019-01-01 00:00:00.000): (lon, lat, radius) in (deg, deg, km)
#    (0., 0., 695700.)>

frame2 = frames.HeliographicStonyhurst(obstime=t2)
coord2 = coord1.transform_to(frame2)
print(coord2)
# <SkyCoord (HeliographicStonyhurst: obstime=2019-01-06 00:00:00.000): (lon, lat, radius) in (deg, deg, km)
#    (-4.62021654, -0.05059666, 698159.27645488)>
```
Note how the radius values are not the same.

dstansby:
Eek the difference in distance becomes larger and larger for larger time differences: with a 6 month difference a point on the surface of the Sun ends up at `r = 777179 m` instead of `695700 m`.

ayshih:
The current behavior is correct, although it can be non-intuitive.  The SkyCoord is a point in *inertial* space, and the Sun moves over time (relative to the solar-system barycenter).  Thus, a point that was on the surface of the Sun at one time will not be exactly on the surface of the Sun at other times.

ayshih:
Since this is very similar to #3530 and has the same underlying reason, I'm going to close this issue as a duplicate.

PR #3540, opened 2019-11-25
Context manager to ignore the inertial motion of the Sun for transformations

ayshih:
Normally, our coordinates machinery assumes that a coordinate points to a location in inertial space (at rest relative to the solar-system barycenter).  When one transforms a coordinate between frames with different `obstime` values, the representation changes because the origin and orientation of the coordinate frames are different, but the location in inertial space itself does not change.  However, this behavior is awkward for coordinates that are intended to point to locations on the Sun itself, because the Sun moves around in inertial space.  Thus, a coordinate that is on the Sun's surface at one time is no longer exactly on the surface at a different time.  In the following example, note that the point is no longer at the Sun's radius after the HGS->HGC transformation with a one-week difference.

```python
>>> import sunpy.coordinates.frames as f
>>> import astropy.units as u

>>> loc = f.HeliographicStonyhurst(0*u.deg, 0*u.deg, obstime='2001-01-01')
>>> print(loc)
<HeliographicStonyhurst Coordinate (obstime=2001-01-01T00:00:00.000): (lon, lat, radius) in (deg, deg, km)
    (0., 0., 695700.)>

>>> print(loc.transform_to(f.HeliographicCarrington(obstime='2001-01-08')))
<HeliographicCarrington Coordinate (obstime=2001-01-08T00:00:00.000): (lon, lat, radius) in (deg, deg, km)
    (117.35959212, 0.08151069, 690181.02067937)>
```

This PR provides the capability – as a context manager – to ignore this inertial motion of the Sun for transformations.  In essence, coordinates behave as if they move along with Sun over time.  While this would be appropriate for coordinates for Sun features, it would not be appropriate for other solar-system objects (e.g., planets or spacecraft).  When the same transformation in the previous example is performed using the context manager, note that the point remains at the Sun's radius.

```python
>>> from sunpy.coordinates import transform_with_sun_center

>>> with transform_with_sun_center():
>>>     print(loc.transform_to(f.HeliographicCarrington(obstime='2001-01-08')))
<HeliographicCarrington Coordinate (obstime=2001-01-08T00:00:00.000): (lon, lat, radius) in (deg, deg, km)
    (117.98984699, 0., 695700.)>
```

Due to the implementation approach, this context manager works for only five of our coordinate frames and transformations between them:

- `HeliographicStonyhurst`
- `HeliographicCarrington`
- `HeliocentricInertial`
- `Heliocentric`
- `Helioprojective`

Fortunately, it would be rare to identify a Sun feature using a frame other than one of these five.

Still needs:
- [x] Settle on the name
- [x] Decide whether to have this available in the `sunpy.coordinates` namespace
- [x] Documentation example
- [x] Tests

Addresses #3530 and #3532

### DIFF
--- a/changelog/3540.feature.rst
+++ b/changelog/3540.feature.rst
@@ -1,0 +1,1 @@
+Added a context manager (`~sunpy.coordinates.transform_with_sun_center`) to ignore the motion of the center of the Sun for coordinate transformations.
--- a/sunpy/coordinates/transformations.py
+++ b/sunpy/coordinates/transformations.py
@@ -14,6 +14,7 @@ This module contains the functions for converting one
 
 """
 import logging
+from contextlib import contextmanager
 from copy import deepcopy
 from functools import wraps
 
@@ -55,7 +56,8 @@ except ImportError:
 
 RSUN_METERS = constants.get('radius').si.to(u.m)
 
-__all__ = ['hgs_to_hgc', 'hgc_to_hgs', 'hcc_to_hpc',
+__all__ = ['transform_with_sun_center',
+           'hgs_to_hgc', 'hgc_to_hgs', 'hcc_to_hpc',
            'hpc_to_hcc', 'hcc_to_hgs', 'hgs_to_hcc',
            'hpc_to_hpc',
            'hcrs_to_hgs', 'hgs_to_hcrs',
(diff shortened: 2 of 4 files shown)


ayshih:
Updated OP with description

sunpy/coordinates/transformations.py
```diff
@@ -66,6 +68,28 @@
...
 
 
+# Boolean flag for whether to ignore the inertial motion of the Sun
+_ignore_sun_motion = False
+
+
+@contextmanager
+def ignore_sun_motion():
+    """
+    Context manager for ignoring the inertial motion of the Sun for transformations.
```
dstansby: The motion of the Sun presumable isn't quite inertial (because it interacts via. gravity with other bodies)? I would suggest:

```suggestion
    Context manager for ignoring the motion of the center of the Sun for transformations.
```

sunpy/coordinates/transformations.py
```diff
@@ -66,6 +68,28 @@
...
+
+    Essentially, coordinates will be assumed to move along with the Sun over time (e.g., for
```
dstansby:
```suggestion
    Essentially, coordinates will be fixed with respect to the center of the Sun and **not** an inertial frame of reference (e.g., for
```

sunpy/coordinates/transformations.py
```diff
@@ -66,6 +68,28 @@
...
+    features on the Sun itself).
```
dstansby: It should be clear here that this doesn't take into account any rotation or differential rotation of the surface of the Sun.

ayshih:
The name of this context manager is up for discussion.  `follow_sun_center` was just suggested in the dev meeting, and I'll just brainstorm some other options right now:

- `as_solar_feature`
- `maintain_sun_location`
- `keep_sun_still`
- `freeze_sun`
- `no_sun_motion`

I actually kinda like `as_solar_feature` over any of the other options because it captures the "why" rather than the "what".  Comments?

ayshih:
I've decided to call the context manager `transform_with_sun_center` and to have it available in the `sunpy.coordinates` namespace.

changelog/3540.feature.rst
```diff
@@ -0,0 +1 @@
+Added a context manager (`~sunpy.coordinates.transformations.ignore_sun_motion`) to ignore the motion of the center of the Sun for transformations.
```
Cadair: This needs updating.

ayshih: Done

ayshih:
Added a test to make sure that the context manager doesn't accidentally make a permanent change to transformations

wafels:
The name `as_solar_feature` may imply that solar differential rotation has been applied, which is not the case.

sunpy/coordinates/transformations.py
```diff
@@ -66,6 +68,69 @@
...
+def transform_with_sun_center():
+    """
+    Context manager for coordinate transformations to ignore the motion of the center of the Sun.
+
+    Normally, coordinates refer to a point in inertial space (relative to the barycenter of the
+    solar system).  Transforming to a different observation time does not move the point at all,
+    but rather only updates the coordinate representation as needed for the origin and axis
+    orientations at the new observation time.  However, the center of the Sun moves over time.
+    Thus, a coordinate that points to a solar feature (e.g., the center of the Sun) will not point
+    to the same solar feature at other observation times.
```
dstansby: I think this needs tightening up a bit, because even with coordinates of a solar feature (e.g. active region) will not point to the solar feature at a later time due to solar rotation. Maybe

"Thus, a coordinate on the surface of the Sun will not remain on the surface of the Sun at other observation times"

sunpy/coordinates/transformations.py
```diff
@@ -66,6 +68,69 @@
...
+        (-156.66825767, 5.96399877, 0.00027959)>
+    >>> with transform_with_sun_center():
+    ...     sun_center.transform_to(end_frame)  # now following Sun center
+    <SkyCoord (HeliographicStonyhurst: obstime=2001-02-01T00:00:00.000): (lon, lat, radius) in (deg, deg, AU)
+        (0., 0., 0.)>
+
+    Notes
+    -----
+    This context manager accounts only for the motion of the center of the Sun, i.e.,
+    translational motion.  The motion of solar features due to the (differential) rotation of the
```
dstansby:
```suggestion
    translational motion.  The motion of solar features due to any rotation of the
```

Cadair:
@ayshih If you want to integrate @dstansby 's comments I think this is good to go.

sunpy/coordinates/transformations.py
ayshih: Edited

sunpy/coordinates/transformations.py
ayshih: Accepted

ayshih:
Done

PR #3543, opened 2019-11-27
Change code block to remove warning

nabobalis:
Some warning now appears.

### DIFF
--- a/docs/dev_guide/documentation.rst
+++ b/docs/dev_guide/documentation.rst
@@ -50,9 +50,11 @@ SunPy Specific Rules
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


docs/dev_guide/documentation.rst
```diff
@@ -50,9 +50,11 @@ SunPy Specific Rules
 
 .. code-block:: python
```
Cadair: make this not a Python code block? This is rst right? (I assume that would also fix the warning?)

nabobalis: Well it was meant to be a python docstring block.

PR #3548, opened 2019-11-27
Block `graphviz` version 2.42 on RTD due to the bug with scaling

### LINKED ISSUE #3545: Class-inheritance diagrams for `coordinates` are being cropped rather than zoomed out
See:

- https://docs.sunpy.org/en/latest/code_ref/coordinates.html#class-inheritance-diagram
- https://docs.sunpy.org/en/latest/code_ref/coordinates.html#id3

But not this one:

- https://docs.sunpy.org/en/latest/code_ref/coordinates.html#id8
(Although this one is just one box, do note that that the box is notably smaller than the boxes that are visible in the earlier inheritance diagrams.)

I don't know when this started, but the same diagram looks fine in 1.0.x (see https://docs.sunpy.org/en/v1.0.6/code_ref/coordinates.html#class-inheritance-diagram).

ayshih:
Fixes #3545

The solution is to block `graphviz` version 2.42.x on RTD.  I confirmed that this solution works  through a test branch on the main repo.

### DIFF
--- a/.rtd-environment.yml
+++ b/.rtd-environment.yml
@@ -30,6 +30,7 @@ dependencies:
   - sphinx-astropy
   - sphinx-gallery
   - towncrier
+  - graphviz!=2.42.*
   - pip:
     - sunpy-sphinx-theme
     - jplephem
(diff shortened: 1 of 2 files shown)


nabobalis:
I assume we want to block this on both 1.0 and 1.1?

ayshih:
We should, yes

nabobalis:
Merge it once the "bot" come back.

PR #3550, opened 2019-11-27
Fix tests for data_manager

Cadair:
So these tests were never being run on the CI.

### DIFF
--- a/changelog/3550.trivial.rst
+++ b/changelog/3550.trivial.rst
@@ -1,0 +1,1 @@
+Fix tests for `sunpy.data.data_manager` and ensure they are correctly executed with pytest.
--- a/setup.cfg
+++ b/setup.cfg
@@ -91,8 +91,7 @@ all_files = 1
 [tool:pytest]
 minversion = 3.0
 testpaths = "sunpy" "docs"
-# Skip sunpy/data to prevent importing the sample data (there are no tests in that dir anyway)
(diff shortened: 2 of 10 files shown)


nabobalis:
Changelog or no?

Cadair:
Yeah it needs a changelog.

I am a little concerned around me removing @vn-ki 's patching of the global cache during the tests. The fact that nothing fails is interesting, but I worry that running the tests locally will now interfere with the "real" cache people are using on their machines.

Also todo, is that if we are to remove that we should remove the code in conftest related to it.

Cadair:
It would be better if it didn't use the real cache for the test suite, so I will go back and look at that again.

> Please note that sunpy_cache function serves a different purpose and shouldn't be removed.

I haven't removed this though right?

vn-ki:
> I haven't removed this though right?

No.

Cadair:
So what tests interact with the main cache here? Is it just the ones for the instr code which uses the data manager? or is it other ones?

vn-ki:
> It would be better if it didn't use the real cache for the test suite, so I will go back and look at that again.

I had an idea but it was not too elegant for my taste.

The idea was to be data manager to be patched using `mocker.patch`. So, in the tests which use data manager, in the beginning of the tests the real data manager in the module will be patched to the mock one using mocker patch.

This could work but the problem is that the responsibility of doing this is on the person writing the module tests and not on the one who knows data manager. But if you think this is a good trade off, I'm sure this is a reasonable way of implementing it.

Cadair:
I want to merge this and then open an issue for ensuring that tests don't interact with the main cache / manager. This is mainly because I don't want to hold 1.1 on this any longer.

Cadair:
I am currently spinning up our own copy of the old backport bot, will merge this after that is complete.

PR #3563, opened 2019-11-29
Corrected docstring for _B0

fluxtransport:
Original description was:
"Return the B0 angle for the Sun at a specified time, which is the heliographic latitude of the Sun-disk center as seen from Earth. "
which has the opposite sense as the revised docstring:
"Return the B0 angle for the Sun at a specified time, which is the tilt of the solar North rotational axis toward the observer (heliographic latitude of the observer)."

### Description

Fixes #<Issue Number>

### DIFF
--- a/sunpy/coordinates/ephemeris.py
+++ b/sunpy/coordinates/ephemeris.py
@@ -223,7 +223,7 @@ def _B0(time='now'):
     Return the B0 angle for the Sun at a specified time, which is the heliographic latitude of the
     of the center of the disk of the Sun as seen from Earth. The range of B0 is +/-7.23 degrees.
 
-    Section 7 of `Thompson, A&A 449, 791-803 (2006) <https://doi.org/10.1051/0004-6361:20054262>__`
+    Table 1 of `Thompson, A&A 449, 791-803 (2006) <https://doi.org/10.1051/0004-6361:20054262>__`
     defines it as: "Tilt of the solar North rotational axis toward the
     observer (heliographic latitude of the observer)"
 
--- a/sunpy/coordinates/ephemeris.py
+++ b/sunpy/coordinates/ephemeris.py
@@ -223,9 +223,9 @@ def _B0(time='now'):
     Return the B0 angle for the Sun at a specified time, which is the heliographic latitude of the
     of the center of the disk of the Sun as seen from Earth. The range of B0 is +/-7.23 degrees.
 
-    Table 1 of `Thompson, A&A 449, 791-803 (2006) <https://doi.org/10.1051/0004-6361:20054262>__`
-    defines it as: "Tilt of the solar North rotational axis toward the
-    observer (heliographic latitude of the observer)"
+    Equivalent definitions include:
+        * The heliographic latitude of Earth
+        * The tilt of the solar North rotational axis toward Earth
 
 
     Parameters
--- a/sunpy/coordinates/ephemeris.py
+++ b/sunpy/coordinates/ephemeris.py
@@ -88,7 +88,7 @@ def get_body_heliographic_stonyhurst(body, time='now', observer=None):
             emitted_time = obstime - light_travel_time
 
         log.info(f"Apparent body location accounts for {light_travel_time.to('s').value:.2f}"
-                  " seconds of light travel time")
+                 " seconds of light travel time")
 
     body_hgs = ICRS(body_icrs).transform_to(HGS(obstime=obstime))
 
@@ -220,11 +220,14 @@ def get_horizons_coord(body, time='now', id_type='majorbody'):
 @add_common_docstring(**_variables_for_parse_time_docstring())
 def _B0(time='now'):
     """
-    Return the B0 angle for the Sun at a specified time, which is the tilt of the solar North rotational
-    axis toward the observer (heliographic latitude of the observer).  The range of B0 is +/-7.23 degrees.
-    Ref: see section 7 of Thompson, A&A 449, 791-803 (2006), 
-         https://www.aanda.org/articles/aa/abs/2006/14/aa4262-05/aa4262-05.html
-    
+    Return the B0 angle for the Sun at a specified time, which is the heliographic latitude of the
+    of the center of the disk of the Sun as seen from Earth. The range of B0 is +/-7.23 degrees.
+
+    Section 7 of `Thompson, A&A 449, 791-803 (2006) <https://doi.org/10.1051/0004-6361:20054262>__`
+    defines it as: "Tilt of the solar North rotational axis toward the
+    observer (heliographic latitude of the observer)"
+
+
     Parameters
     ----------
     time : {parse_time_types}
--- a/sunpy/coordinates/ephemeris.py
+++ b/sunpy/coordinates/ephemeris.py
@@ -220,9 +220,11 @@ def get_horizons_coord(body, time='now', id_type='majorbody'):
 @add_common_docstring(**_variables_for_parse_time_docstring())
 def _B0(time='now'):
     """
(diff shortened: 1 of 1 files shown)


sunpy/coordinates/ephemeris.py
```diff
@@ -217,9 +217,11 @@ def get_horizons_coord(body, time='now', id_type='majorbody'):
 @add_common_docstring(**_variables_for_parse_time_docstring())
 def _B0(time='now'):
     """
-    Return the B0 angle for the Sun at a specified time, which is the heliographic latitude of the
-    Sun-disk center as seen from Earth.  The range of B0 is +/-7.23 degrees.
-
+    Return the B0 angle for the Sun at a specified time, which is the tilt of the solar North rotational
+    axis toward the observer (heliographic latitude of the observer).  The range of B0 is +/-7.23 degrees.
+    Ref: see section 7 of Thompson, A&A 449, 791-803 (2006), 
+         https://www.aanda.org/articles/aa/abs/2006/14/aa4262-05/aa4262-05.html
```
Cadair: The indentation of this line is causing the docs to fail to build. (It's very very fussy!).

You could render it as a hyperlink by doing
```
`Thompson, A&A 449, 791-803 (2006) <https://www.aanda.org/articles/aa/abs/2006/14/aa4262-05/aa4262-05.html
>`__
```

Also, I think we generally try to use DOI links to articles to make sure we are immune to any changes of web addresses etc.

ayshih:
@fluxtransport I'm confused why you believe that the original wording has the "opposite sense" to your suggested wording.  The "heliographic latitude of the Sun-disk center as seen from Earth" is equal to the "heliographic latitude of the observer [Earth]" (which is also equal to the tilt of the Sun's north pole towards Earth).  I'm not saying that the original wording is crystal clear, but I don't understand how you're reading it to be "opposite".  Can you explain?

I feel that it's quite common to see B0 defined as the heliographic latitude of disk center, so I'm reluctant to remove that wording completely.  (That's how the Astronomical Almanac and Meeus define it, and I imagine that would be the case for most literature intended for Earth-based observers.)   That said, I have no objection to *adding* equivalent definitions to the docstring to be as helpful as possible.

sunpy/coordinates/ephemeris.py
fluxtransport: Ok will do. Thanks.

fluxtransport:
Hi @ayshih,

I see your point. Let me explain how I interpreted the docstring. 
Say we are looking at the Sun in September. The north pole is tilted toward us. 
B0 ~  +7 deg as returned by the function. I think the way I interpreted the docstring was not as intended and that may be entirely my fault.  I guess I misinterpreted "Sun-disk center" as the intersection between the Solar equator (in the rotating frame) and central meridian great circles. When I use that interpretation, +7 deg would mean the South pole were tilted toward me. That's what I meant by opposite sense. 

I guess others would interpret it in the intended sense.  I think the explanation in Thompson (2006) is less ambiguous. Let me know if I should just supplement the original with the Thompson (2006) explanation. 

Cheers,
Mark

ayshih:
> Let me know if I should just supplement the original with the Thompson (2006) explanation.

@fluxtransport Sorry for taking so long to respond to you.  Yes, please do that.  At the same time, can you also change my wording of "of the Sun-disk center as seen from Earth" to "of the center of the disk of the Sun as seen from Earth"?  I had been trying to avoid multiple uses of "of the", but I think verbosity will also help to avoid potential confusion here.  Thanks!

nabobalis:
Hi @fluxtransport, I hope it was ok but I (hopefully) addressed @ayshih and @Cadair comments on this PR. 

Since I made these changs, I would like (if you have time) to review what I did and see if they are ok with you.

fluxtransport:
Thanks so much !

sunpy/coordinates/ephemeris.py
```diff
@@ -221,7 +221,12 @@ def get_horizons_coord(body, time='now', id_type='majorbody'):
 def _B0(time='now'):
     """
     Return the B0 angle for the Sun at a specified time, which is the heliographic latitude of the
-    Sun-disk center as seen from Earth.  The range of B0 is +/-7.23 degrees.
+    of the center of the disk of the Sun as seen from Earth. The range of B0 is +/-7.23 degrees.
+
+    Section 7 of `Thompson, A&A 449, 791-803 (2006) <https://doi.org/10.1051/0004-6361:20054262>__`
```
dstansby:
```suggestion
    Table 1 of `Thompson, A&A 449, 791-803 (2006) <https://doi.org/10.1051/0004-6361:20054262>__`
```

nabobalis: The button worked?

sunpy/coordinates/ephemeris.py
```diff
@@ -221,7 +221,12 @@ def get_horizons_coord(body, time='now', id_type='majorbody'):
 def _B0(time='now'):
     """
     Return the B0 angle for the Sun at a specified time, which is the heliographic latitude of the
-    Sun-disk center as seen from Earth.  The range of B0 is +/-7.23 degrees.
+    of the center of the disk of the Sun as seen from Earth. The range of B0 is +/-7.23 degrees.
+
+    Table 1 of `Thompson, A&A 449, 791-803 (2006) <https://doi.org/10.1051/0004-6361:20054262>__`
+    defines it as: "Tilt of the solar North rotational axis toward the
+    observer (heliographic latitude of the observer)"
```
ayshih: I suggest instead saying:
```
Equivalent definitions include:
* The heliographic latitude of Earth
* The tilt of the solar North rotational axis toward Earth
```
The Thompson (2006) reference is a bit of red herring to call out specifically since it's just one of many places where various definitions are stated.

sunpy/coordinates/ephemeris.py
```diff
@@ -221,7 +221,12 @@ def get_horizons_coord(body, time='now', id_type='majorbody'):
 def _B0(time='now'):
     """
     Return the B0 angle for the Sun at a specified time, which is the heliographic latitude of the
-    Sun-disk center as seen from Earth.  The range of B0 is +/-7.23 degrees.
+    of the center of the disk of the Sun as seen from Earth. The range of B0 is +/-7.23 degrees.
+
+    Table 1 of `Thompson, A&A 449, 791-803 (2006) <https://doi.org/10.1051/0004-6361:20054262>__`
+    defines it as: "Tilt of the solar North rotational axis toward the
```
nabobalis:
```suggestion
```

sunpy/coordinates/ephemeris.py
nabobalis:
```suggestion
```

sunpy/coordinates/ephemeris.py
```diff
@@ -221,7 +221,12 @@ def get_horizons_coord(body, time='now', id_type='majorbody'):
 def _B0(time='now'):
     """
     Return the B0 angle for the Sun at a specified time, which is the heliographic latitude of the
-    Sun-disk center as seen from Earth.  The range of B0 is +/-7.23 degrees.
+    of the center of the disk of the Sun as seen from Earth. The range of B0 is +/-7.23 degrees.
+
+    Table 1 of `Thompson, A&A 449, 791-803 (2006) <https://doi.org/10.1051/0004-6361:20054262>__`
```
nabobalis:
```suggestion
    Equivalent definitions include:
        * The heliographic latitude of Earth
        * The tilt of the solar North rotational axis toward Earth
```

sunpy/coordinates/ephemeris.py
nabobalis: Better now?

nabobalis:
Squash merged.

Thank you everyone!

PR #3568, opened 2019-11-30
More helpful error message for Map(fname) when fname doesn't exist

### LINKED ISSUE #3552: File not found error should include the file path when trying to load Map
Currently trying to load a non-existent file using `Map` raises a very generic error message. It would be good if the file that is trying to be loaded could be listed in the error message.

Current message:

```python
  File "main.py", line 84, in <module>
    aia_map = aia_helpers.create_synoptic_map(datetime.now())
  File "/Users/dstansby/github/psp-connection/aia_helpers.py", line 99, in create_synoptic_map
    aia_synop_map = synop_reproject(aia_map, shape)
  File "/Users/dstansby/github/psp-connection/aia_helpers.py", line 82, in synop_reproject
    new_map = Map(synop_map_path)
  File "/Users/dstansby/miniconda3/lib/python3.7/site-packages/sunpy/map/map_factory.py", line 276, in __call__
    data_header_pairs, already_maps = self._parse_args(*args, **kwargs)
  File "/Users/dstansby/miniconda3/lib/python3.7/site-packages/sunpy/map/map_factory.py", line 238, in _parse_args
    raise ValueError("File not found or invalid input")
ValueError: File not found or invalid input
```

dstansby:
Fixes #3552

- Check if `arg` can be co-erced into a `Path` object.
- If so, step through checking if it is a file, dir, or glob, and use appropriate logic in each case
- Add a test that `arg` is included in the error message when it's not available.

### DIFF
--- a/sunpy/map/map_factory.py
+++ b/sunpy/map/map_factory.py
@@ -219,7 +219,7 @@ class MapFactory(BasicRegistrationFactory):
                 data_header_pairs += pairs
 
             # File system path (file or directory or glob)
-            elif _is_path(arg):
+            elif _possibly_a_path(arg):
                 path = pathlib.Path(arg).expanduser()
                 if _is_file(path):
                     pairs = self._read_file(path, **kwargs)
@@ -349,7 +349,7 @@ def _is_url(arg):
     return True
 
 
-def _is_path(arg):
+def _possibly_a_path(arg):
     """
     Check if arg can be coerced into a Path object.
     Does *not* check if the path exists.
--- a/sunpy/map/map_factory.py
+++ b/sunpy/map/map_factory.py
@@ -203,22 +203,9 @@ class MapFactory(BasicRegistrationFactory):
                     data_header_pairs.append(pair)
                     i += 1    # an extra increment to account for the data-header pairing
 
-            # File system path (file or directory)
-            elif _is_path(arg):
-                path = pathlib.Path(arg).expanduser()
-                if path.is_file():
-                    pairs = self._read_file(path, **kwargs)
-                    data_header_pairs += pairs
-                elif path.is_dir():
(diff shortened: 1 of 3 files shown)


sunpy/map/map_factory.py
```diff
@@ -231,12 +218,25 @@ def _parse_args(self, *args, **kwargs):
...
+                path = pathlib.Path(arg).expanduser()
+                if _is_file(path):
+                    pairs = self._read_file(path, **kwargs)
+                    data_header_pairs += pairs
+                elif _is_dir(path):
+                    for afile in sorted(path.glob('*')):
+                        data_header_pairs += self._read_file(afile, **kwargs)
+                elif glob.glob(os.path.expanduser(arg)):
+                    for afile in sorted(glob.glob(os.path.expanduser(arg))):
+                        print(afile)
```
Cadair: remove the print?

Cadair:
Please add a changelog and this looks good to go.

sunpy/map/map_factory.py
dstansby: doh good catch

sunpy/map/map_factory.py
```diff
@@ -350,12 +350,33 @@ def _is_url(arg):
 
 
 def _is_path(arg):
```
wtbarnes: I think the naming of this function is now misleading. I initially thought the `_parse_args` code above was broken because a glob is not a valid path (though it can be coerced into a `pathlib.Path` object).

sunpy/map/map_factory.py
Cadair: this still needs to go :grinning:

Cadair:
Would you be up for adding a couple of extra tests to cover some of the edge cases here and up the coverage?

PR #3569, opened 2019-11-30
Fix __repr__ for unknown coordinate systems

### LINKED ISSUE #2452: The repr of GenericMap raises an error if the WCS is an unknown type.
Make a WCS with some nonsense in CTYPE1 CTYPE2 and then call `repr(smap)`

dstansby:
Fixes #2452

### DIFF
--- a/changelog/3569.bugfix.rst
+++ b/changelog/3569.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed ``map.__repr__`` when the coordinate system information contained in the
+``CUNIT1/2`` metadata is not set to a known value.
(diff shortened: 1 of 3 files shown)


wtbarnes:
Out of curiosity, in what context do you have an unknown coordinate system?

dstansby:
@Cadair's original example was when `'CUNIT1'` was `'STUART1'`, so maybe he invented his own coordinate system? I'm not sure to be honest, but I feel like if one can make an object without an error then `__repr__` shouldn't raise an error.

PR #3570, opened 2019-12-01
Remove un-used csvfile and fitsfile kwargs from _remove_lytaf_events

dstansby:
These are un-tested and un-used keyword arguments for a private function, so remove them and their corresponding code paths. If someone has a good reason to keep them, they should be made public and tested by adding them to the public `remove_lytaf_events`.

### DIFF
--- a/changelog/3570.trivial.rst
+++ b/changelog/3570.trivial.rst
@@ -1,0 +1,2 @@
+Removed un-used and un-tested code paths in the private ``_remove_lytaf_events`` function
+in `sunpy.instr.lyra`.
--- a/sunpy/instr/lyra.py
+++ b/sunpy/instr/lyra.py
@@ -131,8 +131,7 @@ def remove_lytaf_events_from_timeseries(ts, artifacts=None,
 
 
 def _remove_lytaf_events(time, channels=None, artifacts=None,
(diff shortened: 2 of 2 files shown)


wtbarnes:
This looks good to me. However, I find it odd that this function is documented as if it is part of the public-facing API, i.e. it has a fairly extensive example in the docstring that makes it look as if it is supposed to be used directly.

dstansby:
> This looks good to me. However, I find it odd that this function is documented as if it is part of the public-facing API, i.e. it has a fairly extensive example in the docstring that makes it look as if it is supposed to be used directly.

Yes, I did think this - maybe a rummage in the git blame could explain this?

wtbarnes:
Looks like @DanRyanIrish is the one who originally wrote this. Perhaps he can give some insight as to whether this is meant to be a user-facing function or not?

nabobalis:
Isn't this technically an API change? Won't be need to add a warning  in 1.0?

dstansby:
No, because it's changing the API for a non-public (ie. underscored) function.

nabobalis:
> No, because it's changing the API for a non-public (ie. underscored) function.

Good point, I glossed over the `_`

PR #2999, opened 2019-03-17
sunpy.time python file review

nabobalis:
Time for `sunpy.time`. 

This probably has some controversial changes. Specifically the removal of *maybe* useful user facing functions? 

TODO: 
- [ ] Check radiospectra

PR #3571, opened 2019-12-01
Add test for julian_centuries

dstansby:
Should bump test coverage a bit.

It looks like this was meant to be deleted in https://github.com/sunpy/sunpy/pull/2999 but wasn't (see the changelog for that PR), so this should probably be deprecated and removed.

### DIFF
--- a/sunpy/time/time.py
+++ b/sunpy/time/time.py
@@ -14,7 +14,7 @@ from astropy.time import Time
 
 # This is not called but imported to register it
 from sunpy.time.utime import TimeUTime  # noqa
-from sunpy.util.decorators import add_common_docstring
+from sunpy.util.decorators import add_common_docstring, deprecated
 
 __all__ = [
     'find_time', 'parse_time', 'is_time',
@@ -354,6 +354,7 @@ def is_time_in_given_format(time_string, time_format):
         return False
 
 
+@deprecated("2.0", alternative='(time.jd - 2415020) / 36525')
 def julian_centuries(t='now'):
     """
     Returns the number of Julian centuries since J1900.0 (noon on 1900 January
--- a/changelog/3571.removal.rst
+++ b/changelog/3571.removal.rst
@@ -1,0 +1,1 @@
+`sunpy.time.julian_centuries` is deprecated.
--- a/sunpy/time/tests/test_time.py
+++ b/sunpy/time/tests/test_time.py
@@ -9,6 +9,7 @@ from astropy.time import Time
 
 import sunpy.time as time
 from sunpy.time import is_time_equal, parse_time, julian_centuries
+from sunpy.util.exceptions import SunpyDeprecationWarning
 
 LANDING = Time('1966-02-03', format='isot')
 
@@ -339,5 +340,6 @@ def test_is_time_in_given_format():
 
 
 def test_julian_centuries():
-    assert julian_centuries('1899-12-31 12:00:00') == 0
-    assert julian_centuries('1999-12-31 12:00:00') == 0.9999726214921287
+    with pytest.warns(SunpyDeprecationWarning):
+        assert julian_centuries('1899-12-31 12:00:00') == 0
+        assert julian_centuries('1999-12-31 12:00:00') == 0.9999726214921287
--- a/sunpy/time/tests/test_time.py
+++ b/sunpy/time/tests/test_time.py
@@ -8,7 +8,7 @@ import astropy.time
 from astropy.time import Time
 
 import sunpy.time as time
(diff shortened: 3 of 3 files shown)


wtbarnes:
The [diff for #2999](https://github.com/sunpy/sunpy/pull/2999/files#diff-5db3f3c0d2a3561c7f8f115b9ad85947) clearly shows this function being deleted. How is it still here???

dstansby:
It reappeared in the same diff here: https://github.com/sunpy/sunpy/pull/2999/files#diff-e4d7773dc71c1ca30c1741901255ba37R357

dstansby:
I've followed the policy strictly and added a deprecation for `2.0`; if there's a strong feeling for doing this earlier (maybe sneak into `1.1`) then I can easily change this.

sunpy/time/time.py
```diff
@@ -354,6 +354,7 @@ def is_time_in_given_format(time_string, time_format):
         return False
 
 
+@deprecated("2.0", alternative='(time.jd - 2415020) / 36525')
```
Cadair: The alternative here is knowing these magic numbers? Isn't there a way to do it with `Time`?

dstansby: Dunno/I don't even know where the magic numbers are documented or officially defined elsewhere.

wtbarnes:
I would argue for sneaking this into 1.1 and not even bothering with the deprecation. We'v already advertised that we've removed it so we might as well actually do it.

sunpy/time/time.py
ayshih: I believe it traces back to [Newcomb's _Tables of the Sun_](https://en.wikipedia.org/wiki/Newcomb%27s_Tables_of_the_Sun).  A rendition that makes better use of `Time` would be `(time - Time('J1900.0', scale='tt')) / (36525*u.day)` (and I guess a `.value` for the whole thing for just the number).

ayshih:
<sigh>  There's technically a bug in this function because it doesn't explicitly set the scale (e.g., TT versus UTC) for the time input before converting to JD.  Of course, the difference would be very small on the scale of Julian centuries.

sunpy/time/time.py
ayshih: There doesn't appear to be a pre-defined unit in Astropy for Julian century.

Cadair: I realise I am re-litigating this somewhat, but this seems like it's non-trivial to do without this function, so why did it get removed?

nabobalis: I removed it probably.

ayshih: I'm fine with keeping it, but the scale bug needs to be fixed and the docstring ought to be improved.

nabobalis:
I did remove (and clearly in a rebase didn't delete all the functions) a few functions during that PR. My thought process was that it was not used in the core library and seems so specific to have no use in normal code. 

We can or should restore functions if i was too zealous if there any reason to.

vn-ki:
> My thought process was that it was not used in the core library and seems so specific to have no use in normal code.

I remember having the discussion for my GSoC too. The decision was to deprecate/remove it because it was not being used anywhere specifically.

dstansby:
If we're going to keep it we should probably make a conscious decision to put back all the similar stuff that went in #2999. I'm +1 for just getting rid of it. How about vote on this comment, 👍 to keep and improve, 👎 to remove.

vn-ki:
> If we're going to keep it we should probably make a conscious decision to put back all the similar stuff that went in #2999

Apart from the julian stuff, the other functions was being used by the code base prior to switch to astropy time. Due to the switch, what the functions did became trivial or not required. That is why the functions were removed IIRC.

wtbarnes:
@dstansby I'd give it another day and then remove the function/tests (provided there are no other objections)

I've posted in the chat to nudge people to comment here if they have objections

ayshih:
I (weakly) object to removing `julian_centuries`.  I think there needs to be a high bar to removing something, even if we don't use it elsewhere in the library.  In this case, equivalent functionality is non-trivial.

Regardless, even if the consensus is to remove it, there definitely needs to be a deprecation period.  It doesn't matter that a changelog entry claims that it's been removed; it's part of the public API, and someone could be using it

dstansby:
Does anyone have any specific objections to the current changes then? I'm pro improving docstring etc. eventually, but don't really have the enthusiasm at the moment.

wtbarnes:
I know this is out of scope of the original PR, but since we are not removing this now, could you incorporate @ayshih's suggested change to the actual `julian_centuries` function? Or just open an issue to fix it. I'm fine with either. The issue could include fixing the docstring as well.

PR #3572, opened 2019-12-03
Update to outmap meta to improve title in final plot

ehsteve:
### Description

Small patch to update meta data so final plot has a better title.

Fixes nothing but sometimes you gotta yolo.


examples/sunpy_other_packages/reprojection_aia_euvi_mosaic.py
```diff
@@ -185,6 +185,10 @@
 outmap = sunpy.map.Map((array, header))
 outmap.plot_settings = maps[0].plot_settings
 
+outmap.nickname = 'AIA + EUVI/A + EUVI/B'
+outmap.meta.update({'wavelnth': maps[0].meta['wavelnth']})
```
Cadair: these should be moved into the header generation above.

nabobalis: Done.

Issue #3233, opened 2019-06-19
Add an example that shows how to rotate an HMI map


ehsteve:
### Description

HMI fits file need to be rotated by default. This is a common user function and an example would be very useful.

ayshih:
> HMI fits file need to be rotated by default.

To be clear, you're just saying that all HMI maps need to be rotated to have solar north pointing up, and that's something that many users will want to do immediately.  You're not actually making the assertion/request that `HMIMap` should rotate internally as a default action irrespective of what a user might want, correct?

Cadair:
Yeah we should provide a gallery example that makes it clear how to do this, but they don't need to "do it by default" as not all operations need you to have a coordinate system aligned with the grid.

wtbarnes:
Could we just address this in the [existing HMI example](https://docs.sunpy.org/en/stable/generated/gallery/acquiring_data/downloading_hmi.html) or do you think this warrants its own entry in the gallery?

PaulJWright:
I think addressing it at the end of the existing HMI example would be an ideal location for it. Also adding a note in the [map rotation](https://docs.sunpy.org/en/stable/generated/gallery/map/map_rotation.html) mentioning that rotating HMI data is addressed elsewhere?

mbobra:
The [existing HMI example](https://docs.sunpy.org/en/stable/generated/gallery/acquiring_data/downloading_hmi.html) has a few issues:
* It queries for line-of-sight magnetic field data at a cadence of 720s, but downloads the 45s data.
* The wavelength is incorrect (says 6174 Angstroms, not the correct value of 6173).

It also looks like rotating the HMI image seems to work, but somehow Map shows a zoomed-in image. Does anyone know why this is? These are the lines I appended to the end of the example: 
```
hmi_rotated = hmi_map.rotate(angle=180.0 * u.deg)
hmi_rotated.plot()
plt.show()
```

wtbarnes:
>     * It queries for line-of-sight magnetic field data at a cadence of 720s, but downloads the 45s data.

This is a bit confusing, but I think what is happening is that it is sampling the `45_s` series at a 720 s cadence. `attrs.Sample` does not change the series, but rather just the cadence at which the series is queried. I suspect the VSO does not even bother with the `720_s` series as it is already included in `45_s` (I could be wrong about this so someone more familiar with the VSO should probably weigh in too).

>     * The wavelength is incorrect (says 6174 Angstroms, not the correct value of 6173).

[...]

PR #3573, opened 2019-12-03
Fix rotation issue in HMI example (#3233)

mbobra:
This PR fixes issue #3233, which suggests rotating the HMI image in the example gallery entry titled [Downloading and plotting an HMI magnetogram](https://docs.sunpy.org/en/stable/generated/gallery/acquiring_data/downloading_hmi.html#sphx-glr-generated-gallery-acquiring-data-downloading-hmi-py) such that solar North is pointed up.


examples/acquiring_data/downloading_hmi.py
```diff
@@ -32,15 +36,28 @@
...
 
 ###############################################################################
-# Now load it into a map and plot it
+# Now load it into a map and plot it.
+# We see that solar North is pointed down instead of up in this image, which is
+# indicated by the coordinates (that range from positive to negative, rather
+# than negative to positive).
+# We can verify the rotation of the image by querying the keyword `CROTA2`,
+# which indicates the number degrees the image must be rotated for solar North
+# to point up.
```
ayshih: It's peculiar to make this statement but then not demonstrate it.

examples/acquiring_data/downloading_hmi.py
```diff
@@ -32,15 +36,28 @@
...
 hmi_map = sunpy.map.Map(downloaded_file[0])
 fig = plt.figure()
 hmi_map.plot()
 plt.show()
+
+###############################################################################
+# Rotate the image such that solar North is pointed up. The map automatically
+# incorporates the value of `CROTA2` into the rotation.
+hmi_rotated = hmi_map.rotate(order=3)
```
ayshih: You need to explain why you're choosing to use `order=3` (or a reader might interpret that `order=3` is a required keyword argument to utilize the `CROTA2` value).

ayshih:
Also, it looks like `drms` doesn't like the email address provided?
```python
Warning, treated as error:
/home/circleci/project/examples/acquiring_data/downloading_hmi.py failed to execute correctly: Traceback (most recent call last):
  File "/home/circleci/project/.tox/build_docs/lib/python3.7/site-packages/sphinx_gallery/gen_rst.py", line 440, in _memory_usage
    out = func()
  File "/home/circleci/project/.tox/build_docs/lib/python3.7/site-packages/sphinx_gallery/gen_rst.py", line 425, in __call__
    exec(self.code, self.globals)
  File "/home/circleci/project/examples/acquiring_data/downloading_hmi.py", line 41, in <module>
    downloaded_file = Fido.fetch(result[0])
  File "/home/circleci/project/.tox/build_docs/lib/python3.7/site-packages/sunpy/net/fido_factory.py", line 383, in fetch
    wait=False, **kwargs))
  File "/home/circleci/project/.tox/build_docs/lib/python3.7/site-packages/sunpy/net/jsoc/jsoc.py", line 482, in fetch
    responses = self.request_data(jsoc_response)
  File "/home/circleci/project/.tox/build_docs/lib/python3.7/site-packages/sunpy/net/jsoc/jsoc.py", line 419, in request_data
    cd = drms.Client(email=block.get('notify', ''))
  File "/home/circleci/project/.tox/build_docs/lib/python3.7/site-packages/drms/client.py", line 653, in __init__
    self.email = email      # use property for email validation
  File "/home/circleci/project/.tox/build_docs/lib/python3.7/site-packages/drms/client.py", line 866, in email
    raise ValueError('Email address is invalid or not registered')
ValueError: Email address is invalid or not registered
```

mbobra:
Yep, I'm not going to put my own e-mail address in an example. The [docs that describe the SunPy JSOC interface](https://docs.sunpy.org/en/stable/guide/acquiring_data/jsoc.html) also use sunpy@sunpy.org.

examples/acquiring_data/downloading_hmi.py
mbobra: 👍 I will take it out.

examples/acquiring_data/downloading_hmi.py
mbobra: 👍 I will add this.

wtbarnes:
I think it is worth explaining *why* we are rotating the data as well (i.e. removing the roll angle of the instrument) and noting that this info is stored in the header of the observation

examples/acquiring_data/downloading_hmi.py
nabobalis: What is the reason? It's just faster?

nabobalis:
I changed the email address to the ones we use in the test. That seems to work locally for me. Why the sunpy email address isn't working while it seems to in our other docs is something I don't understand.

examples/acquiring_data/downloading_hmi.py
```diff
@@ -3,24 +3,27 @@
...
-# the observation type,
-# and a cadence of images spaced every 720 seconds.
-result = Fido.search(a.Time('2011/11/09 17:40:00', '2011/11/09 17:55:00'),
-                     a.Instrument('hmi'),
-                     a.Sample(720*u.s),
-                     a.vso.Physobs('LOS_magnetic_field'))
+# `Fido <sunpy.net.fido_factory.UnifiedDownloaderFactory>`, a downloader client,
+# to query the Joint Science Operations Center, or JSOC, where HMI data are stored.
+# First define the search variables, a timerange,
+# a [data series](http://jsoc.stanford.edu/JsocSeries_DataProducts_map.html),
```
wtbarnes: You're using Markdown instead of RST 😁

examples/acquiring_data/downloading_hmi.py
```diff
@@ -3,24 +3,27 @@
...
+# keywords, and your e-mail address (to notify you when the download is complete).
+# See the JSOC e-mail address registration page
+# [here](http://jsoc.stanford.edu/ajax/register_email.html).
```
wtbarnes: Same as above

examples/acquiring_data/downloading_hmi.py
```diff
@@ -32,15 +35,26 @@
 # don't provide a path it will download the file into the sunpy data directory.
 # The output provides the path of the downloaded files.
 # The result can be from several data clients, so we have to index the client first
-# client and then index the file.
+# and then index the file.
 
 # Slice the first record returned by the first client.
-downloaded_file = Fido.fetch(result[0, 0])
```
wtbarnes: I think the slicing syntax that you've removed here is actually correct. Indexing it as `[0]` just returns the same response so this will still download more than 1 file if your query returns more than one file. However, it seems like `[0,0]` should not work either given #2781?

examples/acquiring_data/downloading_hmi.py
```diff
@@ -32,15 +35,26 @@
...
+# than negative to positive).
 hmi_map = sunpy.map.Map(downloaded_file[0])
 fig = plt.figure()
 hmi_map.plot()
 plt.show()
+
+###############################################################################
+# Rotate the image such that solar North is pointed up.
+# The order specifies the type of interpolation; in this case, 3 refers to
+# bi-cubic.
```
wtbarnes: See my above comment. I still think this could use 1 sentence (or phrase) explaining why just calling `rotate` with no specification for the angle removes the roll angle. People may find this mysterious.

examples/acquiring_data/downloading_hmi.py
wtbarnes: It's because `order=3` doesn't convert all of your NaNs to zeros. If you use the default (`order=4`), you can't see the edge of the disk.

examples/acquiring_data/downloading_hmi.py
mbobra: Yes, `[0,0]` returns an error, so I kept it the following line:  
`downloaded_file = Fido.fetch(result[0])`. 
 
 Because the search only spans a few minutes, executing this line only downloads one file.

examples/acquiring_data/downloading_hmi.py
mbobra: Eek sorry, fixed this 🙈

examples/acquiring_data/downloading_hmi.py
```diff
@@ -32,15 +35,34 @@
...
 plt.show()
+
+###############################################################################
+# Now rotate the image such that solar North is pointed up. 
+# We have to do this because the HMI instrument is mounted upside-down 
+# relative to the AIA instrument on the SDO satellite, which means most 
+# of the images are taken with solar North pointed up. 
+# The roll angle of the instrument is reported in the FITS header
+# keyword `CROTA2` (see Figure 17 of 
+# `Couvidat et al. (2016)<https://dx.doi.org/10.1007/s11207-016-0957-3>`_,
```
wtbarnes:
```suggestion
# `Couvidat et al. (2016) <https://dx.doi.org/10.1007/s11207-016-0957-3>`_, 
```

I think this is what is causing the docs build to fail?

mbobra: Ugh, sorry Will!

wtbarnes: Wow that actually fixed it 😆

examples/acquiring_data/downloading_hmi.py
```diff
@@ -32,15 +35,34 @@
...
+
+###############################################################################
+# Now rotate the image such that solar North is pointed up. 
+# We have to do this because the HMI instrument is mounted upside-down 
+# relative to the AIA instrument on the SDO satellite, which means most 
+# of the images are taken with solar North pointed up. 
+# The roll angle of the instrument is reported in the FITS header
+# keyword `CROTA2` (see Figure 17 of 
+# `Couvidat et al. (2016) <https://dx.doi.org/10.1007/s11207-016-0957-3>`_, 
+# which states that "the nominal CROTA2 for HMI is ≈179.93".
```
wtbarnes:
```suggestion
# which states that "the nominal CROTA2 for HMI is ≈179.93").
```

Sorry for yet another suggestion! I just noticed that the parentheses were not closed 😭

mbobra: Blah sorry I guess I was doing this too fast -- sorry for the lack of attention to detail, and thanks for catching all these!

wtbarnes:
Despite very much needing to be fixed, test failures are not related to this PR.

Cadair:
**Needs squash merging**.

PR #3576, opened 2019-12-04
Fixed bugs with some transformations when `obstime` is `None` on the destination frame

ayshih:
There's a bug where some coordinate-frame transformations do not succeed when `obstime` is `None` on the destination frame but can be assumed to be the same as the `obstime` of the source frame (which happens automatically when using `SkyCoord`s).  This assumption is "reasonable" when the destination frame is an observer-independent frame.  For implementation reasons, I have difficulty dealing with this situation when crossing the HGS<->HCRS transformation leg, but I can at least fix it for the subgraph of HGS, HGC, HCI, HCC, and HPC.  I think any user encountering this bug would be working within this subgraph.

I encountered this bug when providing a non-HGS coordinate for the `observer` frame attribute, because it gets auto-converted to HGS with `obstime=None` as the destination frame.  That's why there's a test in `frameattributes.py`.

### DIFF
--- a/sunpy/coordinates/transformations.py
+++ b/sunpy/coordinates/transformations.py
@@ -203,7 +203,7 @@ def hgs_to_hgc(hgscoord, hgcframe):
     total_matrix = _rotation_matrix_hgs_to_hgc(int_coord.obstime)
     newrepr = int_coord.cartesian.transform(total_matrix)
 
-    return hgcframe.realize_frame(newrepr)
+    return hgcframe._replicate(newrepr, obstime=int_coord.obstime)
 
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference,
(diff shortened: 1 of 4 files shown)


ayshih:
Ugh.  I wrote a proper test, and there are a whole bunch more situations to fix (and maybe a reason to re-think the approach entirely).  Back to WIP.

ayshih:
Okay, second try.  Go unit tests!

sunpy/coordinates/transformations.py
```diff
@@ -526,7 +526,9 @@ def hgs_to_hgs(from_coo, to_frame):
     """
     Convert between two Heliographic Stonyhurst frames.
     """
-    if np.all(from_coo.obstime == to_frame.obstime):
+    if to_frame.obstime is None:
+        return from_coo.replicate()
```
Cadair: what's the difference between `.replicate()` here and `_replicate()` used in a bunch of other places?

ayshih: `._replicate()` allows one to replace the data and attributes at the same time.  `.replicate()` replaces attributes but keeps the data, while `.realize_frame()` replaces data but keeps the attributes.  Both `.replicate()` and `.realize_frame()` call `._replicate()`.

Cadair: nice and simple then :rofl:

PR #3582, opened 2019-12-05
[joss_paper] Add a footnote about author list sorting

Cadair:


joss_paper/paper.md
```diff
@@ -424,6 +424,9 @@ aas-doi: Add DOI HERE
 aas-journal: The Astrophysical Journal
 ...
 
+[^*]: The author list in this paper is sorted by number of commits to the core
```
nabobalis:
```suggestion
[^*]: The author list in this paper is sorted by number of non-merge commits to the core
```
maybe?

Cadair: except that it includes merge commits as currently authored. I don't want to change the ordering on a whim as it will be a bit of work.

ayshih:
Neato

joss_paper/paper.md
nabobalis: Oh never mind me then.

ehsteve: I would not use the word "core" which is unclear. Why not just say "sunpy package"?

Cadair: "core sunpy repository" makes it pretty clear doesn't it? "package" isn't really the right term here because we are talking about the git repo.

nabobalis: Should we say for master only as well?

PR #3595, opened 2019-12-09
created BaseHeliographic class

### LINKED ISSUE #3584: Consider not deriving HeliographicCarrington from HeliographicStonyhurst
Currently, `HeliographicCarrington` (HGC) is defined as a subclass of `HeliographicStonyhurst` (HGS), which is a straightforward way to avoid some code repetition.  However, this has the potentially surprising behavior that `isinstance()` cannot be used to check whether a frame is of a desired type: `isinstance(frame, HeliographicStonyhurst)` will return `True` even when `frame` is `HeliographicCarrington`.  `isinstance()` is used in this fashion within Astropy's `CoordinateAttribute` class, which we use for the `observer` attribute in our frames.

The solution is to define a new frame (e.g., `BaseHeliographic`) that has the common code but no defined transformations, and then have both HGS and HGC be subclasses of that new frame.  Since HGC isn't actually a version of HGS, having HGC be a subclass of HGS isn't really appropriate anyway.

abhijeetmanhas:
### Description

I have created new Base class `BaseHelipgraphic` which is inherited by HGS and HGC frames. I think `BaseHeliographic` will be more of an Abstract class, its object won't be created, so I haven't added any examples stating how to use it.

Fixes #3584

### DIFF
--- a/changelog/3595.bugfix.rst
+++ b/changelog/3595.bugfix.rst
@@ -1,0 +1,4 @@
+`isinstance(frame, HeliographicStonyhurst)` will no longer
+return true when frame is `HeliographicCarrington` .
+Created new class `Base_Heliographic` inherited by both
+`HGC` and `HGS` .
--- a/changelog/3595.bugfix.rst
+++ b/changelog/3595.bugfix.rst
@@ -1,4 +1,0 @@
-`isinstance(frame, HeliographicStonyhurst)` will no longer
-return true when frame is `HeliographicCarrington` .
-Created new class `Base_Heliographic` inherited by both
-`HGC` and `HGS` .
--- a/sunpy/coordinates/frames.py
+++ b/sunpy/coordinates/frames.py
@@ -136,12 +136,12 @@ class SunPyBaseCoordinateFrame(BaseCoordinateFrame):
             return super().__str__()
 
 
-@add_common_docstring(**_frame_parameters())
 class BaseHeliographic(SunPyBaseCoordinateFrame):
     """
     Base class for HeliographicCarrington (HGC) and HeliographicStonyhurst (HGS) frames.
+
+    This class is not intended to be used directly and has no transformations defined.
     """
-    name = "base_heliographic"
     default_representation = SphericalRepresentation
 
     frame_specific_representation_info = {
--- a/sunpy/coordinates/frames.py
+++ b/sunpy/coordinates/frames.py
@@ -137,7 +137,45 @@ class SunPyBaseCoordinateFrame(BaseCoordinateFrame):
 
 
 @add_common_docstring(**_frame_parameters())
-class HeliographicStonyhurst(SunPyBaseCoordinateFrame):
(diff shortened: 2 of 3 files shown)


abhijeetmanhas:
I will add changelog, if there are any more changes required regarding code; let me know.

changelog/3595.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+`isinstance(frame, HeliographicStonyhurst)` will no longer
```
ayshih:
```suggestion
``isinstance(frame, HeliographicStonyhurst)`` will no longer
```

changelog/3595.bugfix.rst
```diff
@@ -0,0 +1,4 @@
...
+return true when frame is `HeliographicCarrington` .
```
ayshih:
```suggestion
return ``True`` when ``frame`` is `~sunpy.coordinates.frames.HeliographicCarrington` .
```

sunpy/coordinates/frames.py
```diff
@@ -137,7 +137,45 @@ def __str__(self):
 
 
 @add_common_docstring(**_frame_parameters())
-class HeliographicStonyhurst(SunPyBaseCoordinateFrame):
+class BaseHeliographic(SunPyBaseCoordinateFrame):
+    """
+    Base class for HeliographicCarrington (HGC) and HeliographicStonyhurst (HGS) frames.
+    """
+    name = "base_heliographic"
```
ayshih:
```suggestion
```
Don't give this class an explicit name.  It will have the name `baseheliographic` by default.

sunpy/coordinates/frames.py
```diff
@@ -137,7 +137,45 @@ def __str__(self):
 
 
 @add_common_docstring(**_frame_parameters())
-class HeliographicStonyhurst(SunPyBaseCoordinateFrame):
+class BaseHeliographic(SunPyBaseCoordinateFrame):
+    """
+    Base class for HeliographicCarrington (HGC) and HeliographicStonyhurst (HGS) frames.
```
ayshih:
```suggestion
    Base class for HeliographicCarrington (HGC) and HeliographicStonyhurst (HGS) frames.

    This class is not intended to be used directly and has no transformations defined.
```

sunpy/coordinates/frames.py
```diff
@@ -137,7 +137,45 @@ def __str__(self):
 
 
 @add_common_docstring(**_frame_parameters())
```
ayshih:
```suggestion
```
The docstring for this base class isn't making use of the `add_common_docstring()` functionality.

abhijeetmanhas:
@ayshih done suggested changes in files you asked for

ayshih:
@abhijeetmanhas Thanks for making those changes!  I've revised the wording of the changelog entry, but otherwise, it's good to go.

nabobalis:
Thank you @abhijeetmanhas

PR #3598, opened 2019-12-11
Remove astropy_helpers and update packaging to use recommendations in APE 17

Cadair:
This PR contains a few packaging changes to follow the work being done in [APE 17](https://github.com/astropy/astropy-APEs/pull/52)

The only noticeable change here is that `python setup.py test` and `python setup.py build_docs` are no longer supported. If you still run those commands the alternatives are printed, with links to the documentation. 

I will hold off on merging this, so that a) maintainers see it and know about the changes to the setup commands and b) wait until extension helpers has a chance to mature a little.


pyproject.toml
```diff
@@ -5,6 +5,7 @@ requires = [
          "wheel",
          "numpy==1.14.5; python_version<='3.7'",
          "numpy==1.17.3; python_version>='3.8'",
```
astrofrog: Could try using ``oldest-compatible-numpy``?

Cadair:
This is an interesting meta question about the changelog. Is the changelog just for users or also for developers? Personally, I think this is an infrastructure change which shouldn't go in the changelog.

I will be sending info about this around to `sunpy-dev` before we merge it though.

dstansby:
Isn't there a `dev` or similar category in the changelog types?

Cadair:
There is an internal one, I suppose it could go there. I still think that there is a policy decision to be made here (but this probably isn't the place).

ayshih:
I'd like to see `make html` mentioned as an option for building the docs (speaking as someone who keeps forgetting whether `tox` actually works on his setup).

I support adding a changelog entry for this.  There may be some users using the `python setup.py ...` commands, e.g., to make sure that their setup isn't borked or to have docs available offline.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{36,37,38}-{offline,online,astropydev,numpydev,astropy31},build_docs,figure,figure_astropydev,conda
```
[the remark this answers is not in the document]
nabobalis: Oh I see. you can remove them from the list but still call the directly.

nabobalis:
I added a changelog and added astropyfrog's suggestion.

nabobalis:
Do we want to backport this to 1.1?

tox.ini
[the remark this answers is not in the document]
Cadair: Yeah it means that you can run `tox` and get a minimal set locally.

Cadair:
The main thing I want to wait for here is a non-pre release of extension-helpers.

nabobalis:
Seems an API change from extension_helpers came.

tox.ini
```diff
@@ -27,7 +26,7 @@ deps =
     astropydev,build_docs: astroquery
     astropydev,build_docs: jplephem
 commands =
-    offline,astropydev,numpydev,astropy31: {env:PYTEST_COMMAND} {posargs}
```
Cadair: @nabobalis did you drop the numpydev build?

nabobalis: I did

Cadair:
Given the release of extension helpers, I think this is good for merging once approved.

Issue #3507, opened 2019-11-15
MapSequence has a non-functioning keyword `derotate`


ayshih:
The `MapSequence` constructor accepts the keyword `derotate`, with the following documentation:
```
derotate : `bool`
    Apply a derotation to the data. Default to False.
```

The code that is called when `derotate == True` is:
```python
def _derotate(self):
    """Derotates the layers in the MapSequence"""
    pass
```

That is, the keyword doesn't actually do anything.  I'm mystified.

Raahul-Singh:
Hi, I was going through the docs and realized that a lot can be done here.
What would you recommend as the optimal funtionality that needs to be added here?

nabobalis:
Probably needs to be removed. Adding that feature might be non trivial.

nabobalis:
Having said that, I wonder if all you want to do is call rotate on each map.

Raahul-Singh:
Just for the challenge, I'd try implementing the non trivial solution. I assume ``mapsequence_solar_derotate`` which uses  ``sunpy.physics.solar_rotation`` would work, or should I go for differential rotation?
Also, could you suggest some data with more than two maps so that I can test whether I am on the right track or not.

nabobalis:
It would be what function or code that map.rotate uses. Probably to rotate the following maps to the first one in the sequence.  

In terms of data, you'll need to acquire some (using one of the gallery examples as a base) or construct one from the test data.

Raahul-Singh:
Cool. Thanks! :smiley:

wtbarnes:
`map.rotate` is used for rotating about an axis perpendicular to the plane of the image. My understanding is that the intention of the `derotate` kwarg in `MapSequence` was to align all maps in the sequence to some reference image. This can be done in various ways, but ultimately the rotation is going to be about the rotation axis of the Sun rather than about the axis perpendicular to the plane of the image so I don't think we want to be using `.rotate` here.

One approach would be to do this using the existing differential rotation functionality, though it would probably be better to wait until #3537 and #3540 are merged.

nabobalis:
Ah I was mistaken. I just think `derotate` is a strange name for that since it's not the opposite of `rotate`.

Raahul-Singh:
@wtbarnes Thanks! I'll pick this up after those PRs are merged.

ayshih:
While implementing actual functionality would be great, in the meantime I suggest bugfixing existing releases such that using this keyword raises a `NotImplementedError`.

[...]

PR #3613, opened 2019-12-18
Bugfixes ``sunpy.map.mapsequence.MapSequence._derotate()`` and adds tests

Raahul-Singh:
### Description

Updates #3507


Raahul-Singh:
This is my first time working with ``pytest``. Please let me know if any improvements can be made.

nabobalis:
Thanks for the pull request but it seems you also have the older commits from your previous pull request. If you don't want to remove them, we can sort that out when we merge this (via the web UI on github).

sunpy/map/mapsequence.py
```diff
@@ -345,3 +345,9 @@ def all_meta(self):
         Return all the meta objects as a list.
         """
         return [m.meta for m in self.maps]
+
+    def derotate(self):
+        """
+         Return derotated layers in the MapSequence
+        """
+        return self._derotate()
```
nabobalis: Why was this function created?

Raahul-Singh: I was reading somewhere that internal functions should not be directly called outside the class, so created this.

Raahul-Singh:
@nabobalis 
> Thanks for the pull request but it seems you also have the older commits from your previous pull request. If you don't want to remove them, we can sort that out when we merge this (via the web UI on github).

I have been a little careless. Please tell me which option would be more convenient for you. I can open a new PR or resolve these during merging. Sorry for this. :sweat_smile:

sunpy/map/mapsequence.py
nabobalis: But this function isn't called that I could see?

nabobalis:
In that case, we can resolve it when we merge it via GitHub.

sunpy/map/mapsequence.py
Raahul-Singh: It is called in the ``test_mapsequence.py``
For testing the raising of the error, I did not want call the ``_derotate()``outside the class hence created it. 
Would you rather have me call ``_derotate()`` directly for the test?

Raahul-Singh:
> In that case, we can resolve it when we merge it via GitHub.

Thank You!

sunpy/map/mapsequence.py
nabobalis: Well the init of the class calls the underscore method.I don't see the need for this new function. Unless there is that I'm not aware of?

Raahul-Singh: Hmm. I could directly make the object with the ``derorate`` parameter set as ``True``. Would that be fine?

nabobalis: Sorry, I don't follow.

changelog/3613.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Updated ``sunpy.map.mapsequence.MapSequence`` so that calling ``_derotate()`` raises ``NotImplementedError``.
```
nabobalis:
```suggestion
Updated `sunpy.map.mapsequence.MapSequence` so that calling ``_derotate()`` raises ``NotImplementedError``.
```

sunpy/map/mapsequence.py
Raahul-Singh: It was dumb on my part. Rather than make a new function, I have changed the test. In the test, I am making the object with the ``derotate`` set as true, so the ``NotImplementedError`` getting successfully raised is tested.

Raahul-Singh:
@nabobalis Thank you so much for helping me out on this and the other PRs!

nabobalis:
I have manually cherry-picked the commit and pushed it to both 1.0 and 1.1 branches.

PR #3614, opened 2019-12-18
Fixed bug resulting from 8-bit floats in AIA test data

ayshih:
SciPy 1.4 has new FFT code, which now requires floating-point arrays to have depth 16 bits or greater.  This FFT code is used by scikit-image's `match_template`, which is used by our image co-alignment code.

For some reason, our AIA test file has 8-bit floats.  Thus, the tests for the image co-alignment code need to increase the bit depth of the array.

### DIFF
--- a/sunpy/image/tests/test_coalignment.py
+++ b/sunpy/image/tests/test_coalignment.py
@@ -37,7 +37,7 @@ def aia171_test_shift():
 
 @pytest.fixture
 def aia171_test_map_layer(aia171_test_map):
-    return aia171_test_map.data
+    return aia171_test_map.data.astype('float32')  # SciPy 1.4 requires at least 16-bit floats
 
 
 @pytest.fixture


ayshih:
For reference, [this line in C](https://github.com/scipy/scipy/blob/c0dc7fccc53d8a8569cde5d55673fca284bca191/scipy/fft/_pocketfft/pypocketfft.cxx#L199) is where SciPy accepts only `f32` (`float`), `f64` (`double`), and `flong` (`ldbl_t`).  And, [this function](https://github.com/scipy/scipy/blob/c0dc7fccc53d8a8569cde5d55673fca284bca191/scipy/fft/_pocketfft/helper.py#L80) upgrades any `np.float16` array to `np.float32` before hitting the C code.  There is no automatic upgrading for 8-bit float arrays.

ayshih:
It seems that I incorrectly diagnosed this problem, although the fix is still fine.  Our AIA test file doesn't use 8-bit floats – which in retrospect isn't really a thing – but rather has big-endian 8-byte floats.  I imagine the big-endianness was what was breaking the code.

PR #3623, opened 2019-12-19
Squelch possible return value from `plot_coord` in a doctest

ayshih:
See #3616.  This is a second attempt, because #3616 didn't solve the problem.  Apparently doctests don't behave the same as interactive shells with respect to multiple statements separated by semicolons in that the output for each statement is still shown.

Here, I instead explicitly trash the return value if one is returned.  It's slightly ugly, but I think still preferable to doing something more complicated.

### DIFF
--- a/docs/guide/units-coordinates.rst
+++ b/docs/guide/units-coordinates.rst
@@ -229,7 +229,7 @@ Using Coordinates with SunPy Map
    >>> ax = plt.subplot(projection=m)  # doctest: +REMOTE_DATA
    >>> m.plot()  # doctest: +REMOTE_DATA
    <matplotlib.image.AxesImage object at ...>
-   >>> ax.plot_coord(c, 'o');  # doctest: +REMOTE_DATA
+   >>> _ = ax.plot_coord(c, 'o')  # doctest: +REMOTE_DATA
 
 For more information on coordinates see :ref:`sunpy-coordinates` section of
 the :ref:`reference`.


ayshih:
Let's wait for tests to finish this time. =)

ayshih:
The one test fail is unrelated

docs/guide/units-coordinates.rst
```diff
@@ -229,7 +229,7 @@ Using Coordinates with SunPy Map
    >>> ax = plt.subplot(projection=m)  # doctest: +REMOTE_DATA
    >>> m.plot()  # doctest: +REMOTE_DATA
    <matplotlib.image.AxesImage object at ...>
-   >>> ax.plot_coord(c, 'o');  # doctest: +REMOTE_DATA
+   >>> _ = ax.plot_coord(c, 'o')  # doctest: +REMOTE_DATA
```
dstansby: I think this is potentially confusing for a user; would it be possible to add a descriptive variable name that captures what ax.plot_coord returns?

ayshih: The problem is that `plot_coord()` returns a value in Astropy 4.0, and doesn't return a value for earlier versions.  Add a descriptive variable name for a function that doesn't return anything would arguably be more confusing.

ayshih: It's not a widely used code pattern, admittedly, but people do use `_ = ...` to indicate when the return value is not actually going to be used, which is the case here.

nabobalis:
An unrelated failure:
```
163     >>> eve = sunpy.timeseries.TimeSeries("http://lasp.colorado.edu/eve/data_access/quicklook/quicklook_data/L0CS/LATEST_EVE_L0CS_DIODES_1m.txt", source='EVE')  # doctest: +REMOTE_DATA
UNEXPECTED EXCEPTION: NoMatchError('File not found or invalid input')
```

PR #3624, opened 2019-12-19
add .cmap and .norm property to plot_settings

### LINKED ISSUE #3415: Replace plot_settings with .cmap and .norm properties
(as discussed on the call today)

Currently `Map` has a keyword argument `plot_settings`. This is just a dictionary of settings that are passed to `plt.imshow()` when showing the map. However, only a subset of these (`cmap` and `norm`) are "map source specific". It would make for a cleaner API if everything else (e.g. `interpolation`) was passed at "plot time" to either `peek()` or `show()`. To achieve this we need to:

- [ ] Create `.cmap` and `.norm` attributes
- [ ] Add `cmap` and `norm` keyword arguments to `Map.__init__`
- [ ] Deprecate handing `plot_settings` to `Map.__init__`

### LINKED ISSUE #3713: Update docs for new cmap and norm keywords
Once #3624 is merged, the `plot_settings` kwarg will be deprecated in favor of passing `cmap` and `norm` explicitly to the `Map` constructor and the remaining keywords to `plot()` (or `peek()`). The docs need to be updated in several places to account for this, mainly in `guide/data_types/map.rst`, though there may be other places as well. The examples in the example gallery have largely been fixed.

sashank27:
### Description

Replaces up `plot_settings` in ``sunpy.map.mapbase``, and adds up `.cmap` and `.norm` so as to provide for a cleaner and usable API.

Fixes #3415 
Fixes #3713


sashank27:
@dstansby @Cadair @nabobalis please review this

nabobalis:
I'll review it later this week but it seems you have broken some of the tests with the init changes:

```
>           return smap._new_instance(out_data, out_meta)
E           TypeError: _new_instance() missing 2 required positional arguments: 'cmap' and 'norm'
```

The other test fails seem to be fixed in master, so you might want to merge in master.

sunpy/map/mapbase.py
```diff
@@ -1673,7 +1696,7 @@ def peek(self, draw_limb=False, draw_grid=False,
 
     @u.quantity_input
     def plot(self, annotate=True, axes=None, title=True,
-             clip_interval: u.percent = None, **imshow_kwargs):
+             clip_interval: u.percent = None, plot_settings=None, **imshow_kwargs):
```
wtbarnes: My interpretation of @dstansby's comment in #3415 was that we should support `interpolation` and `origin` as explicit kwargs to `plot` (and consequentially `peek`). This just means setting the `imshow_args` in `plot` a bit more explicitly.

sunpy/map/mapbase.py
```diff
@@ -1625,7 +1646,7 @@ def draw_contours(self, levels: u.percent, axes=None, **contour_args):
 
     @peek_show
     def peek(self, draw_limb=False, draw_grid=False,
-             colorbar=True, **matplot_args):
+             colorbar=True, plot_settings=None, **matplot_args):
```
wtbarnes: See comment on `plot`

sunpy/map/mapbase.py
```diff
@@ -195,21 +196,26 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
...
-        self.plot_settings = {'cmap': 'gray',
-                              'norm': norm,
-                              'interpolation': 'nearest',
-                              'origin': 'lower'
-                              }
         if plot_settings:
+            warnings.warn("Handling of plot_settings is deprecated. "
+                          "Pass the plot specific settings to either peek() or show()",
+                          DeprecationWarning)
             self.plot_settings.update(plot_settings)
```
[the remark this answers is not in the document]
wtbarnes: In order to continue to support `plot_settings`, but move away from it internally, we could (temporarily) add it as a property with a setter such that the `norm` and `cmap` are adjusted any time `plot_settings` is modified, e.g.

```python
@property
def plot_settings(self):
    return self._plot_settings

@plot_settings.setter
def plot_settings(self, plot_settings):
    warnings.warn("Handling of plot_settings is deprecated. "
                          "Pass the plot specific settings to either peek() or show()",
                          DeprecationWarning)
    self._plot_settings = plot_settings
    if 'norm' in plot_settings:
        self.norm = plot_settings['norm']
    if 'cmap' in plot_settings:
        self.cmap = plot_settings['cmap']
```

Then in `__init__`, we would just need

```python
self.plot_settings = plot_settings
```

and throughout the rest of the code, we can just deal with `cmap` and `norm` internally. Once the deprecation period has passed, we can just remove this property.

sunpy/map/mapbase.py
```diff
@@ -560,7 +566,7 @@ def shift(self, axis1: u.deg, axis2: u.deg):
                                self.spatial_units[1] + axis2).to(self.spatial_units[1])).value
 
         # Create new map with the modification
-        new_map = self._new_instance(self.data, new_meta, self.plot_settings)
+        new_map = self._new_instance(self.data, new_meta)
```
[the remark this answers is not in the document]
sashank27: Well, in that case, we will need to modify `_new_instance` calls in `compositemap` and `mapsequence` as well

sunpy/map/mapbase.py
sashank27: I am thinking of defining `imshow_args` as below:
```
imshow_args = copy.deepcopy(imshow_kwargs)
if self.plot_settings is not None:
	imshow_args.update(self.plot_settings)
```

Also in some scenarios like https://github.com/sunpy/sunpy/blob/6a5fab2591c0e12f29c0c622f5f52f06f4eb3efd/sunpy/map/compositemap.py#L425 the `cmap` property is still accessed as `"cmap": m.plot_settings['cmap']`, and it expects to provide for a default value to it.
Do all the instances like these need to be edited?

sunpy/map/mapbase.py
```diff
@@ -195,21 +196,19 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
...
-                              'norm': norm,
-                              'interpolation': 'nearest',
-                              'origin': 'lower'
-                              }
-        if plot_settings:
-            self.plot_settings.update(plot_settings)
+        self.plot_settings = plot_settings
+        self.cmap = cmap
+
+        if norm is None:
```
wtbarnes: I think this whole if block could also go in the `norm` setter. Sorry for not pointing this out in the initial review. Then, this all could be replaced by `self.norm = norm`

sunpy/map/mapbase.py
sashank27: @wtbarnes?

wtbarnes: You don't want to initialize `imshow_args` with `imshow_kwargs` as we want to be able to override any defaults with the keyword arguments passed to `plot`. Additionally, we want to move away from any internal use of `plot_settings`.

The fix is even simpler I think. Instead of initializing `imshow_args` with `plot_settings`, just create a dictionary explicitly from `norm`, `cmap`, etc.

```python
imshow_args = {
    'cmap': self.cmap,
    'norm': self.norm,
    'interpolation': interpolation,
    'origin': origin,
}
```

where `interpolation` and `origin` are now explicit kwargs for `plot` with default values of 'nearest' and 'lower', respectively.

wtbarnes: Actually, for the sake of simplicity, just initialize `imshow_args` as,

```python
imshow_args = {
    'cmap': self.cmap,
    'norm': self.norm,
    'interpolation': 'nearest',
    'origin': 'lower',
}
```

and ignore my comment about passing additional keyword arguments to plot. These can just be absorbed into `imshow_kwargs`.

sashank27: Okay! So I'm defining `imshow_args` as below:
```python
if 'interpolation' not in imshow_kwargs:
	interpolation = 'nearest'
else:
	interpolation = imshow_kwargs['interpolation']
        
if 'origin' not in imshow_kwargs:
	origin = 'lower'
else:
	origin = imshow_kwargs['origin']
        
imshow_args = {
	'cmap': self.cmap,
	'norm': self.norm,
	'interpolation': interpolation,
	'origin': origin,
}
        
```

sashank27: Also, what about this?https://github.com/sunpy/sunpy/blob/6a5fab2591c0e12f29c0c622f5f52f06f4eb3efd/sunpy/map/compositemap.py#L425
Do I need to remove all these instances by `cmap = self.cmap`? Or `plot_settings` need to be handled for this?

wtbarnes: All the checking of `imshow_kwargs` is  unnecessary. Anything passed in with `imshow_kwargs` will override the values in `imshow_args` in [this line](https://github.com/sunpy/sunpy/blob/6a5fab2591c0e12f29c0c622f5f52f06f4eb3efd/sunpy/map/mapbase.py#L1753).

sashank27: Yeah I saw that later :sweat_smile:

wtbarnes: Yes, all references to `plot_settings['cmap']` should be replaced with `self.cmap` (same with `norm`). I think you got most of them?

sashank27: Yeah! But then also build fails :disappointed:
I'm unsure of what is causing the issue now

wtbarnes: The problem seems to be with the gallery examples. If you can, try to build the docs (and gallery) locally (https://docs.sunpy.org/en/latest/dev_guide/documentation.html) to see which examples are causing the problems. The CI seems to be stopping every time it hits a gallery example that fails.

sunpy/map/mapbase.py
```diff
@@ -195,21 +196,10 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
...
         # Visualization attributes
-        self.plot_settings = {'cmap': 'gray',
-                              'norm': norm,
-                              'interpolation': 'nearest',
-                              'origin': 'lower'
-                              }
-        if plot_settings:
-            self.plot_settings.update(plot_settings)
+        self.plot_settings = plot_settings
+        self.cmap = cmap
```
vn-ki: Is the original way of providing cmap and norm using `plot_settings` still supported?

sashank27: @vn-ki If `plot_settings` is passed in `__init__`, then the setter updated the `cmap` and `norm`, if these are passed in `plot_settings`

sunpy/map/mapbase.py
```diff
@@ -54,9 +54,10 @@ class GenericMap(NDData):
         A 2d list or ndarray containing the map data.
     header : dict
         A dictionary of the original image header tags.
-    plot_settings : dict, optional
```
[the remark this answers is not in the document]
nabobalis: I am not sure there is one. Adding a note saying it is deprecated makes the most sense to me.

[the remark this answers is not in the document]
sashank27: @nabobalis would this work?
```
plot_settings : dict, optional
        (DEPRECATED) Plot settings.
```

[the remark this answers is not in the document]
nabobalis: Works for me. But ideally will want to say what version it will be removed in as well (which i guess would be 2.1?)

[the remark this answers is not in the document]
sashank27: So this?
```
plot_settings : dict, optional
        Plot settings (Deprecated, will be removed in 2.1)
```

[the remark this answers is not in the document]
nabobalis: Yeah.

sunpy/map/mapbase.py
```diff
@@ -195,21 +196,26 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
         self._validate_meta()
         self._shift = SpatialPair(0 * u.arcsec, 0 * u.arcsec)
 
-        if self.dtype == np.uint8:
-            norm = None
+        # Visualization attributes
```
wtbarnes: As discussed before, I think it is best to leave all the `cmap` and `norm` logic in the setters. However, I'll backtrack a bit and say that we should abandon the idea of using a property-setter approach for plot settings. Instead, I think the constructor should just look like

```python
def __init__(self, data, header, cmap='gray', norm=None, plot_settings=None, **kwargs):
    # Other stuff
    self.cmap = cmap
    self.norm = norm
    if plot_settings:
        self.plot_settings = plot_settings
        warnings.warn("Handling of plot_settings is deprecated. "
                                 "Pass the plot specific settings to either peek() or show()",
                                 DeprecationWarning)
        if 'norm' in self.plot_settings:
            self.norm = self.plot_settings['norm']
        if 'cmap' in self.plot_settings:
            self.cmap = self.plot_settings['cmap']
    else:
        self._plot_settings = {}
    # Other stuff
```

sunpy/map/mapbase.py
```diff
@@ -1437,13 +1445,53 @@ def cmap(self):
...
-        return cmap
+        return self._cmap
+
+    @cmap.setter
+    def cmap(self, n):
+        if isinstance(n, str):
+            self._cmap = plt.get_cmap(n)
+        else:
+            self._cmap = n
+        self.plot_settings['cmap'] = self._cmap
```
wtbarnes: Remove this line. After thinking about it a bit, there's no reason we need to continually update `plot_settings` since it will not be used internally to set anything.

sunpy/map/mapbase.py
```diff
@@ -1437,13 +1445,53 @@ def cmap(self):
...
+        if n is None:
+            if self.dtype == np.uint8:
+                self._norm = None
+            else:
+                # Put import here to reduce sunpy.map import time
+                from matplotlib import colors
+                self._norm = colors.Normalize()
+        else:
+            self._norm = n
+        self.plot_settings['norm'] = self._norm
```
wtbarnes: Remove this line. See note in `cmap` setter above. Sorry again for going back on what I said before 😄

sunpy/map/mapbase.py
```diff
@@ -1437,13 +1445,53 @@ def cmap(self):
...
+
+    @property
```
wtbarnes: Remove this. See my note in `__init__`

sunpy/map/mapbase.py
```diff
@@ -1437,13 +1445,53 @@ def cmap(self):
...
+    def plot_settings(self):
+        return self._plot_settings
+
+    @plot_settings.setter
```
wtbarnes: Remove this. See my note in `__init__`. While this does mean that subsequent setting of `plot_settings` will have no effect, I think it is worth the degree of complexity it introduces (particularly since we are removing it in 2.1). We could add a note in the initial warning stating that subsequently setting `plot_settings` will have no effect.

sunpy/map/mapbase.py
```diff
@@ -1726,7 +1774,13 @@ def plot(self, annotate=True, axes=None, title=True,
...
 
         # Normal plot
-        imshow_args = copy.deepcopy(self.plot_settings)
+        imshow_args = {
+                        'cmap': self.cmap,
+                        'norm': self.norm,
+                        'interpolation': 'nearest',
+                        'origin': 'lower',
+                      }
+
```
wtbarnes: I think we should add a `imshow_args.update(self.plot_settings)` here. Otherwise, the `interpolation` and `origin` keywords in `plot_settings` (if they are present) will have no effect. As with the `plot_settings` logic in `__init__`, we can just remove this line in 2.1

sunpy/map/mapbase.py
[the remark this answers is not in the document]
sashank27: I actually didn't add the documentation, as I was waiting for the review :sweat_smile:

sashank27:
@wtbarnes I was thinking to update all the examples on a separate PR, as this already has a lot of changes

sunpy/map/mapbase.py
```diff
@@ -54,9 +54,12 @@ class GenericMap(NDData):
...
     header : dict
         A dictionary of the original image header tags.
+    cmap : `matplotlib.colors.Colormap`, str
+        Colormap of the map image. Defaults to 'grey'
+    norm : `matplotlib.colors.Normalize`
+        Normalization function used. Defaults to None
     plot_settings : dict, optional
-        Plot settings.
-
+        Plot settings (Deprecated, will be removed in 2.1)
```
wtbarnes:
```suggestion
        Keyword arguments to be passed to `~matplotlib.pyplot.imshow` (Deprecated, will be removed in 2.1)
```

sunpy/map/mapbase.py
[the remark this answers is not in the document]
wtbarnes: Oops. I think I was confusing your comments with commits. Could you add backticks around `dict` as well? (for some reason GitHub won't let me suggest this change)

wtbarnes:
> @wtbarnes I was thinking to update all the examples on a separate PR, as this already has a lot of changes

That's completely fine, but it looks like (at least) one of the examples is causing the docs build to fail. Could you try to build the examples locally and fix up the ones that fail? Others can be left to separate PR.

sashank27:
> > @wtbarnes I was thinking to update all the examples on a separate PR, as this already has a lot of changes
> 
> That's completely fine, but it looks like (at least) one of the examples is causing the docs build to fail. Could you try to build the examples locally and fix up the ones that fail? Others can be left to separate PR.

Well, previously this was the issue. I had to update all the examples to make the build pass.
I've decided now to update the examples here only.

wtbarnes:
Ok so the docs build is failing because one of the examples is constructing the color map name incorrectly. The reason is because the observatory name (SDO in this case) is being excluded when creating the new header. 

If you pass the additional keyword argument `observatory=map_aia.observatory` when creating the new header here:

https://github.com/sunpy/sunpy/blob/eaab3362e91b65d73de2c41f0fa4b46eb09cfa7d/examples/sunpy_other_packages/reprojection_different_observers.py#L130

I think this should fix the problem. Previously, this was not raising an exception because we were not validating the colormap when creating the Map and it was later being overridden by the old `plot_settings` dict which had the correct colormap name.

sashank27:
> Ok so the docs build is failing because one of the examples is constructing the color map name incorrectly. The reason is because the observatory name (SDO in this case) is being excluded when creating the new header.
> 
> If you pass the additional keyword argument `observatory=map_aia.observatory` when creating the new header here:
> 
> https://github.com/sunpy/sunpy/blob/eaab3362e91b65d73de2c41f0fa4b46eb09cfa7d/examples/sunpy_other_packages/reprojection_different_observers.py#L130
> 
> I think this should fix the problem. Previously, this was not raising an exception because we were not validating the colormap when creating the Map and it was later being overridden by the old `plot_settings` dict which had the correct colormap name.

Well this worked, but with a small change. Instead of `observatory=map_aia.observatory` we had to do `telescope=map_aia.observatory`, as AIAMap loads the observatory name from the `telescop` of `meta`.

wtbarnes:
If you want to rebase on top of master, those test failures should be fixed (by #3661).

sunpy/map/mapbase.py
```diff
@@ -195,21 +199,21 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
...
+            warnings.warn("Handling of ``plot_settings`` is deprecated."
+                          "Subsequently setting ``plot_settings`` will have no effect on the plot."
+                          "Pass the plot specific settings to either ``.peek()`` or ``.show()``",
+                          DeprecationWarning)
+            if 'norm' in self.plot_settings:
+                self.norm = self.plot_settings['norm']
+            if 'cmap' in self.plot_settings:
+                self.cmap = self.plot_settings['cmap']
+        else:
+            self.plot_settings = {}
```
nabobalis: What is the reason for an empty plot settings?

sunpy/map/tests/test_mapbase.py
```diff
@@ -206,6 +206,11 @@ def test_cmap(generic_map):
     assert generic_map.cmap == plt.get_cmap('gray')
 
 
+def test_norm(generic_map):
+    from matplotlib import colors
+    assert isinstance(generic_map.norm, colors.Normalize)
+
```
nabobalis: Do we want a test to ensure you can create a map with norm and cmap?

sunpy/map/mapbase.py
```diff
@@ -195,21 +199,21 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
...
-                              }
+        self.cmap = cmap
+        self.norm = norm
         if plot_settings:
-            self.plot_settings.update(plot_settings)
+            self.plot_settings = plot_settings
+            warnings.warn("Handling of ``plot_settings`` is deprecated."
+                          "Subsequently setting ``plot_settings`` will have no effect on the plot."
+                          "Pass the plot specific settings to either ``.peek()`` or ``.show()``",
+                          DeprecationWarning)
```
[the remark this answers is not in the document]
sashank27: I didn't knew of that before. Will edit it to `SunpyDeprecationWarning`

sunpy/map/mapbase.py
```diff
@@ -159,7 +163,7 @@ def __init_subclass__(cls, **kwargs):
         if hasattr(cls, 'is_datasource_for'):
             cls._registry[cls] = cls.is_datasource_for
 
-    def __init__(self, data, header, plot_settings=None, **kwargs):
+    def __init__(self, data, header, cmap='gray', norm=None, plot_settings=None, **kwargs):
```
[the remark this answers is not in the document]
sashank27: Except `cmap` and `norm`, everything would work as expected.

sunpy/map/tests/test_mapbase.py
sashank27: Sorry I didn't understood

sunpy/map/mapbase.py
sashank27: > As discussed before, I think it is best to leave all the `cmap` and `norm` logic in the setters. However, I'll backtrack a bit and say that we should abandon the idea of using a property-setter approach for plot settings. Instead, I think the constructor should just look like
> 
> ```python
> def __init__(self, data, header, cmap='gray', norm=None, plot_settings=None, **kwargs):
>     # Other stuff
>     self.cmap = cmap
>     self.norm = norm
>     if plot_settings:
>         self.plot_settings = plot_settings
>         warnings.warn("Handling of plot_settings is deprecated. "
>                                  "Pass the plot specific settings to either peek() or show()",
>                                  DeprecationWarning)
>         if 'norm' in self.plot_settings:
>             self.norm = self.plot_settings['norm']
>         if 'cmap' in self.plot_settings:
>             self.cmap = self.plot_settings['cmap']
>     else:
>         self._plot_settings = {}
>     # Other stuff
> ```

@nabobalis As previously discussed in this thread and in the chatroom, ensuring 100% cross-compatibility is not possible. BY setting `plot_settings={}` we can ensure the plot settings other than `cmap` and `norm` are correctly handled by the plotting functions. Otherwise, `plot_settings` would default to `None` and one can't pass plot specific settings according to previous method.

nabobalis: That's fair.

sunpy/map/mapbase.py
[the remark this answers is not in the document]
nabobalis: Ok!

sunpy/map/tests/test_mapbase.py
nabobalis: We added two new keywords to the init but unless I missed it (did the review late at night), I don't see a test to ensure that they work?

sunpy/map/mapbase.py
wtbarnes: Now that I see this, there's really no reason not to just leave it as `None`. Since we're not using the setter anymore, setting `plot_settings` (or any keys in it) will have no effect on how the Map is plotted.

sunpy/map/tests/test_mapbase.py
sashank27: Oh sorry, I missed adding the test! Will add it soon!

sunpy/map/mapbase.py
sashank27: @wtbarnes so `plot_settings = {}` should be done or not?

sunpy/map/mapbase.py
```diff
@@ -195,21 +199,21 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
...
+            self.plot_settings = plot_settings
+            warnings.warn("Handling of ``plot_settings`` is deprecated."
+                          "Subsequently setting ``plot_settings`` will have no effect on the plot."
+                          "Pass the plot specific settings to either ``.peek()`` or ``.show()``",
+                          DeprecationWarning)
+            if 'norm' in self.plot_settings:
+                self.norm = self.plot_settings['norm']
+            if 'cmap' in self.plot_settings:
+                self.cmap = self.plot_settings['cmap']
+        else:
```
wtbarnes: Remove this `else` block.

sunpy/map/mapbase.py
```diff
@@ -1726,7 +1749,15 @@ def plot(self, annotate=True, axes=None, title=True,
...
         # Normal plot
-        imshow_args = copy.deepcopy(self.plot_settings)
+        imshow_args = {
+                        'cmap': self.cmap,
+                        'norm': self.norm,
+                        'interpolation': 'nearest',
+                        'origin': 'lower',
+                      }
+
+        imshow_args.update(self.plot_settings)
```
wtbarnes: Change to

```python
if self.plot_settings:
    imshow_args.update(self.plot_settings)
```

sunpy/map/mapbase.py
wtbarnes: Yeah I don't really like the idea of an empty `dict` hanging around if we haven't set it. See my comments below. I don't think removing it affects anything else. Of course, this will break someone's code where they set `plot_settings` after the fact, but it won't have any affect anyway so 🤷‍♀

sunpy/map/mapbase.py
```diff
@@ -195,21 +199,19 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
...
-
         # Visualization attributes
-        self.plot_settings = {'cmap': 'gray',
-                              'norm': norm,
-                              'interpolation': 'nearest',
-                              'origin': 'lower'
-                              }
+        self.cmap = cmap
+        self.norm = norm
         if plot_settings:
```
wtbarnes:
```suggestion
        if self.plot_settings:
```

sunpy/map/mapbase.py
```diff
@@ -195,21 +199,19 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
...
-            self.plot_settings.update(plot_settings)
+            self.plot_settings = plot_settings
```
wtbarnes: Move this just outside the `if` block. That way, the `plot_settings` attribute always exists.

sunpy/map/mapbase.py
```diff
@@ -1726,7 +1747,16 @@ def plot(self, annotate=True, axes=None, title=True,
...
         # Normal plot
-        imshow_args = copy.deepcopy(self.plot_settings)
+        imshow_args = {
+                        'cmap': self.cmap,
+                        'norm': self.norm,
+                        'interpolation': 'nearest',
+                        'origin': 'lower',
+                      }
+
+        if hasattr(self, 'plot_settings') and self.plot_settings is not None:
```
wtbarnes:
```suggestion
        if self.plot_settings:
```

Given my above suggestion, we don't need to check for the existence of `self.plot_settings`.

wtbarnes: Is there a reason to check explicitly if it is not `None`?

sunpy/map/tests/test_mapbase.py
```diff
@@ -262,6 +267,33 @@ def test_rotation_matrix_cd_cdelt():
     np.testing.assert_allclose(cd_map.rotation_matrix, np.array([[0., -1.], [1., 0]]))
 
 
+def test_rotation_matrix_cd_cdelt():
```
wtbarnes: This new test has the same name as an already existing one and I think it could be greatly simplified. 

```python
def test_norm_cmap_kwargs(generic_map):
    cmap = 'cool'
    from matplotlib import colors, pyplot as plt
    norm = colors.Normalize(vmin=1., vmax=1.)
    smap = sunpy.map.Map(generic_map.data, generic_map.meta, cmap=cmap, norm=norm)
    assert smap.cmap == plt.get_cmap(cmap)
    assert smap.norm == norm 
```

We probably need one additional test for the new deprecation warning as well 😁

sunpy/map/tests/test_mapbase.py
wtbarnes: See my comment below w.r.t. tests

wtbarnes:
The test failures are unrelated, but I've kicked the py38-online one anyway.

Looks good to me. Thanks @sashank27!

changelog/3624.feature.rst
```diff
@@ -0,0 +1 @@
+Creates sources-specific ``.cmap`` and ``.norm`` attributes in `~sunpy.map.GenericMap.__init__()`, and deprecate handling of ``.plot_settings``, while shifting it to ``peek()`` and ``show()``, so as to provide for a cleaner API.
```
Cadair:
```suggestion
Creates sources-specific ``.cmap`` and ``.norm`` attributes in ``sunpy.map.GenericMap.__init__``, and deprecate handling of ``.plot_settings``, while shifting it to ``peek()`` and ``show()``, so as to provide for a cleaner API.
```

docs/guide/data_types/maps.rst
```diff
@@ -351,8 +351,9 @@ dictionary. In the following example we change the title of the plot by changing
     import sunpy.data.sample
     import matplotlib.pyplot as plt
     smap = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
-    smap.plot_settings['title'] = "My Second SunPy Plot"
-    smap.plot_settings['cmap'] = plt.cm.Blues_r
+    plot_settings = {}
```
Cadair: This example doesn't seem to be right? What are we recommending people do now? Pass these in as kwargs to `plot`?

sunpy/map/tests/test_mapbase.py
```diff
@@ -206,6 +206,11 @@ def test_cmap(generic_map):
     assert generic_map.cmap == plt.get_cmap('gray')
 
 
+def test_norm(generic_map):
+    from matplotlib import colors
```
Cadair: I think this import could be at the top?

sunpy/map/tests/test_mapbase.py
```diff
@@ -284,6 +289,27 @@ def test_rotation_matrix_cd_cdelt_square():
     np.testing.assert_allclose(cd_map.rotation_matrix, np.array([[0., -1], [1., 0]]))
 
 
+def test_norm_cmap_kwargs(generic_map):
+    cmap = 'cool'
+    from matplotlib import colors, pyplot as plt
```
Cadair: likewise here, I don't think we need imports inside tests.

sunpy/map/tests/test_mapbase.py
```diff
@@ -270,9 +275,9 @@ def test_rotation_matrix_cd_cdelt_square():
         'CRPIX1': 5,
         'CRPIX2': 5,
         'CDELT1': 10,
-        'CDELT2': 10,
+        'CDELT2': 9,
```
Cadair: What's motivated these changes?

sashank27: Idk how these values got changed :/ will edit them back

docs/guide/data_types/maps.rst
sashank27: since there is no `plot_settings` property associated with `sunpy.map.Map`, we have to pass the plot-specific attributes like `title` as kwargs to `plot`.
I just realized this can have a better implementation as below:
```python
plot_settings = {'title':'My Second SunPy Plot'}
smap.cmap = plt.cm.Blues_r
fig = plt.figure()
smap.plot()
```

Cadair: My point was you aren't actually passing the `plot_settings` anywhere after you create it. What I think we should be documenting is:
```python
fig = plt.figure()
smap.plot(cmap="Blues_r", title="My Second Plot")
```
and not constructing `plot_settings` or setting `map.norm` / `map.cmap` etc

sashank27: oops sorry, didn't notice I wasn't passing `plot_settings` anywhere.
Can't we pass `cmap` as property of `map`?
```python
smap.cmap = plt.cm.Blues_r
fig = plt.figure()
smap.plot(title="My Second Plot")
````
In this way user can adhere to the recent API as well

**Comment by sashank27** [comment]
> This generally looks good, thanks for the work @sashank27 😄
> 
> My only real concern here is how we are handling the docs and the example changes made in this PR. Are we going to uniformly recommend people use `plot(norm=, cmap=)` over `map.norm` / `map.cmap`? If so we should make sure we are doing this everywhere that is appropriate in the docs and examples. (Even if we plan to update them separately later, any changes made in this PR should subscribe to the new way not some interim situation).

`cmap` and `norm` are not plot-specific, so better would be to pass them as attributes. Other properties like `title` should be passed to `plot`

**Comment by Cadair** [comment]
I think as a pattern to document, having `map.cmap` be the default for that map, and encouraging users to pass overrides into `plot()` is better than the recommended pattern be to override the source defaults.

I don't know if others have an opinion on this @dstansby it was your issue originally? :grinning:

**Comment by wtbarnes** [comment]
That sounds sensible to me, but if that's how we want to proceed, why bother with the `cmap` and `norm` in the constructor or bothering with properties at all?

File: examples/sunpy_other_packages/reprojection_different_observers.py [hunk 30]
```diff
@@ -133,13 +134,15 @@
...
+    telescope=map_aia.observatory,
     wavelength=map_aia.wavelength
 )
 
 ######################################################################
 # Once again we need to generate a `~astropy.wcs.WCS` and then manually
 # set the observer location.
 
+print('header:', mars_header)
 mars_wcs = WCS(mars_header)
```
wtbarnes: why was this added?

examples/sunpy_other_packages/reprojection_different_observers.py
```diff
@@ -133,13 +134,15 @@
...
     instrument="AIA",
+    telescope=map_aia.observatory,
     wavelength=map_aia.wavelength
 )
 
 ######################################################################
 # Once again we need to generate a `~astropy.wcs.WCS` and then manually
 # set the observer location.
 
+print('header:', mars_header)
```
wtbarnes: Why was this added?

dstansby:
> I think as a pattern to document, having `map.cmap` be the default for that map, and encouraging users to pass overrides into `plot()` is better than the recommended pattern be to override the source defaults.
> 
> I don't know if others have an opinion on this @dstansby it was your issue originally? 😀

If a user wants to set `m.norm` or `m.cmap` they can, but also on a one time basis they can hand keywrods to `plot()`. Will take a look at the changelog at some point soon...

examples/sunpy_other_packages/reprojection_different_observers.py
sashank27: I am extremely sorry, I added it while debugging, and forgot to remove

wtbarnes: No problem 😅 surprised we didn’t catch it before

Cadair:
I think as @dstansby said, it makes sense to allow the setting of `m.cmap/norm` but generally the pattern should be to override on calls to plot. I would approve of not adding this to the constructor of `GenericMap` and handling setting of the properties in `_new_instance`.

sunpy/map/tests/test_mapbase.py
```diff
@@ -206,6 +207,11 @@ def test_cmap(generic_map):
     assert generic_map.cmap == plt.get_cmap('gray')
 
 
+def test_norm(generic_map):
+    from matplotlib import colors
```
nabobalis:
```suggestion
```

nabobalis:
So the plan is now to not set `norm` and `cmap` in examples and pass them into plot?

sashank27:
> So the plan is now to not set `norm` and `cmap` in examples and pass them into plot?

According to the previous discussions by cadair and dstansby I believe, Yes.
> it makes sense to allow the setting of `m.cmap/norm` but generally the pattern should be to override on calls to plot.

docs/guide/data_types/maps.rst
```diff
@@ -347,23 +347,9 @@ plot changes the default AIA color table to use an inverse Grey color table.
...
-    import sunpy.data.sample
-    import matplotlib.pyplot as plt
-    smap = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
-    smap.plot_settings['title'] = "My Second SunPy Plot"
-    smap.plot_settings['cmap'] = plt.cm.Blues_r
-    fig = plt.figure()
-    smap.plot()
-    plt.colorbar()
-    plt.show()
+Earlier, you had and option to view or make changes to the default settings through
```
wtbarnes: Remove this. We don't need to mention deprecated options in the docs

docs/guide/data_types/maps.rst
```diff
@@ -472,6 +454,13 @@ Note how the color in the colorbar does not change since these two maps share
 the same colormap while the data values associated with each color do because
 the normalization is different.
 
+Since, both colormap and the normalization are map-source specific, there is an option to change
```
wtbarnes:
```suggestion
You may also explicitly set both the colormap and normalization::
```

sunpy/map/mapbase.py
```diff
@@ -54,9 +54,13 @@ class GenericMap(NDData):
         A 2d list or ndarray containing the map data.
     header : dict
         A dictionary of the original image header tags.
-    plot_settings : dict, optional
-        Plot settings.
-
+    cmap : `matplotlib.colors.Colormap`, `str`
```
wtbarnes: Remove this

sunpy/map/mapbase.py
```diff
@@ -54,9 +54,13 @@ class GenericMap(NDData):
...
+        Colormap of the map image. Defaults to 'grey'.
+    norm : `matplotlib.colors.Normalize`
```
wtbarnes: Remove this

sunpy/map/mapbase.py
wtbarnes: Given conversations with @Cadair and @dstansby, we don't want to accept `cmap` and `norm` into the constructor so remove these two kwargs

sunpy/map/mapbase.py
```diff
@@ -195,21 +199,20 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
...
-
         # Visualization attributes
-        self.plot_settings = {'cmap': 'gray',
-                              'norm': norm,
-                              'interpolation': 'nearest',
-                              'origin': 'lower'
-                              }
-        if plot_settings:
-            self.plot_settings.update(plot_settings)
+        self.cmap = cmap
```
wtbarnes: Remove. See above comment

sunpy/map/mapbase.py
```diff
@@ -195,21 +199,20 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
...
+        self.norm = norm
```
wtbarnes: Remove. See above comment.

sunpy/map/mapbase.py
```diff
@@ -246,12 +249,12 @@ def __repr__(self):
...
     @classmethod
-    def _new_instance(cls, data, meta, plot_settings=None, **kwargs):
+    def _new_instance(cls, data, meta, **kwargs):
         """
         Instantiate a new instance of this class using given data.
-        This is a shortcut for ``type(self)(data, meta, plot_settings)``.
+        This is a shortcut for ``type(self)(data, meta)``.
         """
-        return cls(data, meta, plot_settings=plot_settings, **kwargs)
+        return cls(data, meta, **kwargs)
```
wtbarnes: So given our removal of `cmap` and `norm` from the constructor, in order to preserve the cmap and norm across new instance creation, what I think we want here is

```suggestion
        new_instance =  cls(data, meta, **kwargs)
        new_instance.cmap = self.cmap
        new_instance.norm = self.norm
        return new_instance
```

sunpy/map/mapbase.py
```diff
@@ -564,7 +567,7 @@ def shift(self, axis1: u.deg, axis2: u.deg):
                                self.spatial_units[1] + axis2).to(self.spatial_units[1])).value
 
         # Create new map with the modification
-        new_map = self._new_instance(self.data, new_meta, self.plot_settings)
+        new_map = self._new_instance(self.data, new_meta, norm=self.norm, cmap=self.cmap)
```
wtbarnes:
```suggestion
        new_map = self._new_instance(self.data, new_meta)
```

sunpy/map/mapbase.py
```diff
@@ -997,7 +1000,7 @@ def resample(self, dimensions: u.pixel, method='linear'):
         new_meta['crval2'] = lat.value
 
         # Create new map instance
-        new_map = self._new_instance(new_data, new_meta, self.plot_settings)
+        new_map = self._new_instance(new_data, new_meta, norm=self.norm, cmap=self.cmap)
```
wtbarnes:
```suggestion
        new_map = self._new_instance(new_data, new_meta)
```

sunpy/map/mapbase.py
```diff
@@ -1112,7 +1115,7 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=4, scale=1.0,
         pixel_array_center = (np.flipud(new_data.shape) - 1) / 2.0
 
         # Create a temporary map so we can use it for the data to pixel calculation.
-        temp_map = self._new_instance(new_data, new_meta, self.plot_settings)
+        temp_map = self._new_instance(new_data, new_meta, norm=self.norm, cmap=self.cmap)
```
wtbarnes:
```suggestion
        temp_map = self._new_instance(new_data, new_meta)
```

sunpy/map/mapbase.py
```diff
@@ -1181,7 +1184,7 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=4, scale=1.0,
         new_meta.pop('CD2_2', None)
 
         # Create new map with the modification
-        new_map = self._new_instance(new_data, new_meta, self.plot_settings)
+        new_map = self._new_instance(new_data, new_meta, norm=self.norm, cmap=self.cmap)
```
wtbarnes:
```suggestion
        new_map = self._new_instance(new_data, new_meta)
```

sunpy/map/mapbase.py
```diff
@@ -1340,10 +1343,11 @@ def submap(self, bottom_left, top_right=None):
         if self.mask is not None:
             new_mask = self.mask[yslice, xslice].copy()
             # Create new map with the modification
-            new_map = self._new_instance(new_data, new_meta, self.plot_settings, mask=new_mask)
+            new_map = self._new_instance(new_data, new_meta, mask=new_mask,
```
wtbarnes:
```suggestion
            new_map = self._new_instance(new_data, new_meta, mask=new_mask)
```

sunpy/map/mapbase.py
```diff
@@ -1340,10 +1343,11 @@ def submap(self, bottom_left, top_right=None):
...
+                                         norm=self.norm, cmap=self.cmap)
```
wtbarnes:
```suggestion
```

sunpy/map/mapbase.py
```diff
@@ -1340,10 +1343,11 @@ def submap(self, bottom_left, top_right=None):
...
             return new_map
         # Create new map with the modification
-        new_map = self._new_instance(new_data, new_meta, self.plot_settings)
+        new_map = self._new_instance(new_data, new_meta, norm=self.norm, cmap=self.cmap)
```
wtbarnes:
```suggestion
        new_map = self._new_instance(new_data, new_meta)
```

sunpy/map/mapbase.py
```diff
@@ -1431,7 +1435,8 @@ def superpixel(self, dimensions: u.pixel, offset: u.pixel=(0, 0)*u.pixel, func=n
             new_mask = None
 
         # Create new map with the modified data
-        new_map = self._new_instance(new_data, new_meta, self.plot_settings, mask=new_mask)
+        new_map = self._new_instance(new_data, new_meta, mask=new_mask,
```
wtbarnes:
```suggestion
        new_map = self._new_instance(new_data, new_meta, mask=new_mask)
```

sunpy/map/mapbase.py
```diff
@@ -1431,7 +1435,8 @@ def superpixel(self, dimensions: u.pixel, offset: u.pixel=(0, 0)*u.pixel, func=n
...
+                                     norm=self.norm, cmap=self.cmap)
```
wtbarnes:
```suggestion
```

sunpy/map/tests/test_mapbase.py
```diff
@@ -284,6 +289,25 @@ def test_rotation_matrix_cd_cdelt_square():
     np.testing.assert_allclose(cd_map.rotation_matrix, np.array([[0., -1], [1., 0]]))
 
 
+def test_norm_cmap_kwargs(generic_map):
+    cmap = 'cool'
+    norm = colors.Normalize(vmin=1., vmax=1.)
+    smap = sunpy.map.Map(generic_map.data, generic_map.meta, cmap=cmap, norm=norm)
```
wtbarnes:
```suggestion
    smap = sunpy.map.Map(generic_map.data, generic_map.meta)
    smap.cmap = cmap
    smap.norm = norm
```

sunpy/map/tests/test_mapbase.py
```diff
@@ -284,6 +289,25 @@ def test_rotation_matrix_cd_cdelt_square():
     np.testing.assert_allclose(cd_map.rotation_matrix, np.array([[0., -1], [1., 0]]))
 
 
+def test_norm_cmap_kwargs(generic_map):
```
wtbarnes: Could possibly have a better name, but `kwargs` didn't seem appropriate given the above changes.

```suggestion
def test_norm_cmap_setters(generic_map):
```

docs/guide/data_types/maps.rst
```diff
@@ -472,6 +454,13 @@ Note how the color in the colorbar does not change since these two maps share
...
+then from the map itself as follows::
```
wtbarnes:
```suggestion
```

examples/plotting/hmi_synoptic_maps.py
```diff
@@ -47,8 +47,8 @@
 
 ###############################################################################
 # Let's also fix the plot settings
-syn_map.plot_settings['cmap'] = 'hmimag'
-syn_map.plot_settings['norm'] = plt.Normalize(-1500, 1500)
+syn_map.cmap = 'hmimag'
```
wtbarnes: Instead of setting `cmap` and `norm` explicitly, pass these as keyword arguments to `syn_map.plot()`

examples/sunpy_other_packages/reprojection_aia_euvi_mosaic.py
```diff
@@ -183,7 +184,8 @@
 # little.
 
 outmap = sunpy.map.Map((array, header))
-outmap.plot_settings = maps[0].plot_settings
+outmap.cmap = maps[0].cmap
```
wtbarnes: Pass these as kwargs to `plot` instead of setting them explicitly. To incorporate the altered `vmin`, you may need to do something like,

```python
norm = maps[0].norm
norm.vmin = 400
```

and then

```python
im = outmap.plot(cmap=maps[0].cmap, norm=norm)
```

examples/sunpy_other_packages/reprojection_aia_euvi_mosaic.py
```diff
@@ -114,10 +114,11 @@
 
 ######################################################################
 # To display the output we construct a new map using the new array and our
-# generated header. We also borrow the plot settings from the AIA map.
+# generated header. We also borrow the colormap and normalization attributes from the AIA map.
```
wtbarnes:
```suggestion
# generated header.
```

examples/sunpy_other_packages/reprojection_aia_euvi_mosaic.py
```diff
@@ -114,10 +114,11 @@
...
 
 outmap = sunpy.map.Map((array, header))
-outmap.plot_settings = maps[0].plot_settings
+outmap.cmap = maps[0].cmap
```
wtbarnes: Pass these as kwargs to `plot()`

examples/sunpy_other_packages/reprojection_align_aia_hmi.py
```diff
@@ -45,9 +45,8 @@
 # reduce the computation time.
 
 map_aia, map_hmi = [m.resample((1024, 1024)*u.pix) for m in sunpy.map.Map(sorted(files))]
-# Why do we have to do this?
-map_hmi.plot_settings['cmap'] = "hmimag"
-map_hmi.plot_settings['norm'] = plt.Normalize(-2000, 2000)
+map_hmi.cmap = "hmimag"
```
wtbarnes: Pass these as kwargs to `map_hmi.plot()` instead of setting explicitly

examples/sunpy_other_packages/reprojection_align_aia_hmi.py
```diff
@@ -74,8 +73,8 @@
 # Construct an output map and set some nice plotting defaults.
 
 out_hmi = sunpy.map.Map(output, map_aia.wcs)
-out_hmi.plot_settings['cmap'] = "hmimag"
-out_hmi.plot_settings['norm'] = plt.Normalize(-1500, 1500)
+out_hmi.cmap = "hmimag"
```
wtbarnes: Same as above

examples/sunpy_other_packages/reprojection_different_observers.py
```diff
@@ -98,7 +98,8 @@
 # to the AIA image.
 
 outmap = sunpy.map.Map(output, out_header)
-outmap.plot_settings = map_stereo.plot_settings
+outmap.cmap = map_stereo.cmap
```
wtbarnes: Same as above

examples/sunpy_other_packages/reprojection_different_observers.py
```diff
@@ -154,7 +156,8 @@
 # We generate the output map and plot it next to the original image.
 
 outmap = sunpy.map.Map((output, mars_header))
-outmap.plot_settings = map_aia.plot_settings
+outmap.cmap = map_aia.cmap
```
wtbarnes: Same as above

examples/sunpy_other_packages/reprojection_heliographic_stonyhurst.py
```diff
@@ -57,7 +57,8 @@
 
 array, footprint = reproject_interp(aia_map, out_wcs, shape_out=shape_out)
 outmap = sunpy.map.Map((array, header))
-outmap.plot_settings = aia_map.plot_settings
+outmap.cmap = aia_map.cmap
```
wtbarnes: Same as above

wtbarnes:
Everything looks good to me, but something weird is happening with one of the examples and is causing the docs build to fail. For some reason, it is trying to smash together "stereo a" and "aia 171" as the color map name and then failing because that is not a valid color map name (obviously :laughing: )

wtbarnes:
Thanks @sashank27. One more approval and we'll be good to go.

I'm a bit concerned by why that change in the ordering of the maps was needed...Have the filenames changed that caused the ordering to be swapped? Or perhaps there's something strange going on in the factory? Either way, it is unrelated to this PR.

changelog/3624.feature.rst
```diff
@@ -0,0 +1 @@
+Creates sources-specific ``.cmap`` and ``.norm`` attributes in ``sunpy.map.GenericMap.__init__``, and deprecate handling of ``.plot_settings``, while shifting it to ``peek()`` and ``show()``, so as to provide for a cleaner API.
```
Cadair: This changelog type should be `removal`.
```suggestion
`sunpy.map.GenericMap.plot_settings` is now deprecated in favour of passing keyword arguments to the `sunpy.map.GenericMap.plot` method. New properties `~sunpy.map.GenericMap.cmap` and `~sunpy.map.GenericMap.norm` properties contain the source specific colormap and norm settings. These properties are settable so can be overridden.
```

examples/plotting/map_editcolormap.py
```diff
@@ -17,12 +17,12 @@
...
 
 ###############################################################################
-# All plot settings for a map are stored in the `plot_settings` attribute.
+# The colormap and normalization for a map are accessed as a property of the map.
 # How a Map is displayed is determined by its colormap, which sets the colors
 # , and the normalization, which sets how data values are translated to colors.
 # Lets replace the colormap and normalization.
-aiamap.plot_settings['cmap'] = plt.get_cmap('Greys_r')
-aiamap.plot_settings['norm'] = colors.LogNorm(100, aiamap.max())
+aiamap.cmap = plt.get_cmap('Greys_r')
```
Cadair: I am a little concerned that this example isn't really describing what we think is the best way of doing things. It should probably be completely re-written?

sunpy/map/mapbase.py
```diff
@@ -195,21 +195,21 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
...
-            self.plot_settings.update(plot_settings)
+        self.cmap = 'gray'
+        self.norm = None
+
+        self.plot_settings = plot_settings
+
+        if self.plot_settings:
+            warnings.warn("Handling of ``plot_settings`` is deprecated."
+                          "In the future setting ``plot_settings`` will have no effect on the plot."
+                          "Pass the plot specific settings to either ``.peek()`` or ``.show()``",
```
ayshih:
```suggestion
                          "Pass the plot specific settings to either ``.peek()`` or ``.plot()``",
```

sashank27: 🤦‍♂

ayshih:
I worry that there are additional use cases for `plot_settings`, beyond colormap and normalization, that will no longer be supported.  Being able to store the plot settings with each map means that a user could more seamlessly work with a collection(/list/sequence) of maps, including slicing, without having to keep and synchronize a separate collection of plot settings.

However, during the dev meeting, we didn't see any plot settings would conceivably be used in this way by a user, aside from maybe interpolation.

Cadair:
hey @sashank27 

Thank you so much for working on this. We chatted about it today at the community meeting, and generally we all struggled to remember why we wanted to make this change in the first place, and we wanted to stop and think about it in terms of the future changed to Map (i.e. integration with ndcube).

Therefore we are going to put this on the agenda for the SunPy Coordination meeting at the end of March, and talk about it with all the other things relating to Map.

Sorry that we don't feel like we can just accept this as it is, the code is great and thank you very much for all your effort on it.

nabobalis:
With the decision in https://github.com/sunpy/sunpy/issues/3415#issuecomment-611522080 I will be closing this PR as I believe the new change probably needs to be done from scratch (especially with all the file conflicts).

I want to thank you @sashank27 for the effort you put into this PR and I am sorry it was never merged due to us being undecided on the new formal API for this change.

Issue #3331, opened 2019-08-30
a.Sample does not work for VSO for STEREO Data


Cadair:
When doing the following VSO query:

```python
Fido.search(a.Time('2012-03-04','2012-03-06'), a.Instrument("EUVI"), 
            a.Wavelength(304*u.AA), a.Sample(100*u.min))
```

The VSO responds with:
```
DEBUG: VSO Response:
 <soap:Envelope xmlns:VSO="http://virtualsolar.org/VSO/VSOi" xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <soap:Body>
    <VSO:QueryResponse>
      <body>
        <provideritem>
          <error>VSO-D404 Bad Request -- Invalid Time Sample : contains non digit</error>
          <provider>SSC</provider>
          <version>1</version>
        </provideritem>
      </body>
    </VSO:QueryResponse>
  </soap:Body>
</soap:Envelope>
```

because we send Sample as a float and not an integer. This works for JSOC backed data (AIA, HMI, etc) but does not work for SSC (the data provider for STEREO).

It is possible to patch this in SunPy by doing the following:
```
diff --git a/sunpy/net/vso/attrs.py b/sunpy/net/vso/attrs.py
index f778883a8..eee1db660 100644
--- a/sunpy/net/vso/attrs.py
+++ b/sunpy/net/vso/attrs.py
@@ -520,6 +520,10 @@ walker.add_converter(SimpleAttr)(
     lambda x: ValueAttr({(x.__class__.__name__.lower(), ): x.value})
 )
 
+walker.add_converter(Sample)(
+    lambda x: ValueAttr({(x.__class__.__name__.lower(), ): int(x.value)})
+)
+
 walker.add_converter(Wavelength)(
     lambda x: ValueAttr({
             ('wave', 'wavemin'): x.min.value,
```

however, doing that applies to all clients, not just the `VSOClient` and restricts the minimum Sample resolution to 1s, so I am reluctant to do it, and have asked the VSO team to look into fixing it server side. I am opening this issue to keep track.

Cadair:
> doing that applies to all clients

My recent delve into attrs makes me think this isn't true. I would need to look into how we use attrs in dataretriever

ejm4567:
[...]

Issue #3248, opened 2019-06-26
Strange key added to map.meta '\x00\x00\x00\x00\x00\x00\x00\x00'


amunozj:
### Description

Trying to convert meta to fits_header using Map.fits_header fails because a very odd key, but this key is not in the original header.  It is added somewhere in the construction of the map.   

### Expected behavior
For the meta not to have this key

### Actual behavior
Original header:

```
SIMPLE  =                    T  / FITS STANDARD                                 
BITPIX  =                  -32 / FITS BITS/PIXEL                                
NAXIS   =                    2  / NUMBER OF AXES                                
NAXIS1  =                  512  /                                               
NAXIS2  =                  512  /                                               
BUNIT   = 'GAUSS'                                                               
ORIGIN  = 'MT. WILSON'                                                          
TELESCOP= '150 ft'                                                              
OBSERVER= 'SBEW RET'                                                            
OBJECT  = 'SUN'                                                                 
DATA-TYP= 'MAGNETOGRAM'                                                         
DATE-OBS= '1981.02.01_18:08:00_TAI'                                             
TAVG-UT =              19.2890  / Avg UT during obs                             
CARROT  =            1704.7341  / Carr rotation                                 
CENLON  =              95.7238  / Carr long of cent med                         
B0      =            -0.106516  / Sun tilt(rad)                                 
X0      =             256.5000  / image x center                                
Y0      =             256.5000  / image y center                                
R0      =             256.0000  / image radius                                  
LAMBDA  =              5250.20  / Wavelength(Angstroms)                         
GV      =             3.000000  / Lande g-factor                                
DXB_IMG =               3.8035  / image x-arcsec/pixel                          
DYB_IMG =               3.8035  / image y-arcsec/pixel                          
DXB_OBS =               9.7679  / obs avg x-arcsec/pixel                        
DYB_OBS =              12.6425  / obs avg y-arcsec/pixel                        
REMARK1 = 'TSAFARG   .M'                                                        
REMARK2 =                                                                       
PSEPMA  =             0.000000  / port separation in milliangstroms             
ICHK    =                    0  / int. ratio flag, 0=no, 1=yes, 2=1./ratio      
MRKMOD  =                  256  / hrdwre 0=datex,256=ael,1280-1283=fm,1284=24ch 

CUNIT1  = 'arcsec  '                                                            
CUNIT2  = 'arcsec  '                                                            
CDELT1  =               3.8035                                                  
CDELT2  =               3.8035                                                  
[...]

PR #3627, opened 2019-12-20
Difference image example for the gallery

jmason86:
### Description
This is my first gallery example so may have done something wrong, especially with sphinx stuff. 

Partially address Issue #3.


examples/map/difference_images.py
```diff
@@ -0,0 +1,64 @@
...
+diff = maps[1].data - maps[0].data
+
+###########################################################################
+# But we have to decide what to do with the metadata. For example, what
+# time does this difference image correspond to? The time of the first or
+# second image? The mean time? You'll have to decide what makes most sense
+# for your application. Here we'll just use the metadata from the second
+# image. Then we can store the difference and header back in a Map.
+
+header = maps[1].fits_header
```
Cadair: You would be better off using `maps[1].meta` here rather than `fits_header`.

examples/map/difference_images.py
```diff
@@ -0,0 +1,64 @@
+"""
+================
```
Cadair: These lines need to be the same length as the title.

examples/map/difference_images.py
```diff
@@ -0,0 +1,64 @@
...
+import sunpy.map
+
+###########################################################################
+# First we'll download a couple images and store them in a
+# sunpy.map.MapSequence(). This could be from whatever imager.
+# Here, we use SDO/AIA 304 Å.
+
+instrument = a.Instrument('AIA')
+wave = a.Wavelength(30 * u.nm, 31 * u.nm)
+result1 = Fido.search(a.Time('2015-06-18T00:00:00', '2015-06-18T00:00:10'), instrument, wave)
```
Cadair: I think you can do this in one search like this:

```
result = Fido.search(a.Time('2015-06-18T00:00:00', '2015-06-18T00:00:10') | a.Time('2015-06-18T01:03:30', '2015-06-18T01:03:35'), instrument, wave)

File: examples/map/difference_images.py [hunk 4]
```diff
@@ -0,0 +1,64 @@
...
+header = maps[1].fits_header
+diff_map = sunpy.map.Map(diff, header)
+
+###########################################################################
+# Finally, we'll plot it. We'll apply a colormap and renormalize the
+# intensity so that it shows up well.
+
+plt.figure()
+ax_diff = plt.subplot(projection=diff_map)
+diff_map.plot_settings['cmap'] = plt.get_cmap('Greys_r')
```
Cadair: I think it might be nicer to just pass these as keyword arguments to `plot()`?

examples/map/difference_images.py
jmason86: Done (I think)

examples/map/difference_images.py
jmason86: No reason. Fixed.

examples/map/difference_images.py
jmason86: Done. And updated variable name accordingly.

examples/map/difference_images.py
jmason86: Somehow didn't realize that was an option. Done.

jmason86:
Note that there are several impediments to doing this with coronagraph data right now due to other issues. Unfortunately, that's the class of data that probably does the most difference imaging. 

For example, Issues:
* #3331
* #3248 
* and (probably a related issue to #3248), trying to plot LASCO at all can cause: 

> WARNING: SunpyUserWarning: Could not determine coordinate frame from map metadata [sunpy.map.mapbase].

and

> ValueError: FITS header values must contain standard printable ASCII characters; 'offset_bias.pro\t1.37, 12/18/14, 631.118V15 18 Jun 1997 MAKE_FITS_HDR2' contains characters not representable in ASCII or non-printable characters.

examples/map/difference_images.py
```diff
@@ -0,0 +1,62 @@
...
+Plotting a difference image
+===========================
+How to compute and plot a difference image.
+The example uses `sunpy.map.MapSequence` to compute a difference image and then plot it.
+This basic method works for base difference or running difference. Just change whether
+you're subtracting the previous image or the first image in a sequence.
+"""
+###########################################################################
+
+#sphinx_gallery_thumbnail_number = 2
```
wtbarnes: For some reason this is showing up in the built documentation. Presumably this should be hidden?

jmason86: I noticed that too. I dunno how to make it go away, though it does appear to be functional as well since the thumbnail in the gallery is the second image.

wtbarnes: [This example from sphinx-gallery](https://sphinx-gallery.github.io/stable/auto_examples/plot_4_choose_thumbnail.html#sphx-glr-auto-examples-plot-4-choose-thumbnail-py) shows how to use the `sphinx_gallery_thumbnail_number` option, but they seem to output the whole code snippet at the end? Maybe the best option is just to stick it somewhere where it doesn't interrupt the narrative text of your example.

examples/map/difference_images.py
```diff
@@ -0,0 +1,62 @@
...
+
+import matplotlib.pyplot as plt
+import matplotlib.colors as colors
+import astropy.units as u
+from sunpy.net import Fido, attrs as a
+import sunpy.map
+
+###########################################################################
+# First we'll download a couple images and store them in a
+# sunpy.map.MapSequence(). This could be from whatever imager.
```
wtbarnes:
```suggestion
# `sunpy.map.MapSequence`. This could be from any channel of any imager.
```

examples/map/difference_images.py
```diff
@@ -0,0 +1,62 @@
...
+# Here, we use SDO/AIA 304 Å.
+
+instrument = a.Instrument('AIA')
+wave = a.Wavelength(30 * u.nm, 31 * u.nm)
+result = Fido.search(a.Time('2015-06-18T00:00:00', '2015-06-18T00:00:10') |
```
wtbarnes: I think this query would be a bit more readable written as,

```python
 result = Fido.search(
    a.Time('2015-06-18T00:00:00', '2015-06-18T00:00:10') | a.Time('2015-06-18T01:03:30', '2015-06-18T01:03:35'), 
    instrument,
    wave
)
```

(Sorry for the lack of a suggestion. Apparently GitHub does not like multi-line suggested changes).

examples/map/difference_images.py
```diff
@@ -0,0 +1,62 @@
...
+meta = maps[1].meta
+diff_map = sunpy.map.Map(diff, meta)
+
+###########################################################################
+# Finally, we'll plot it. We'll apply a colormap and renormalize the
+# intensity so that it shows up well.
+
+plt.figure()
+ax_diff = plt.subplot(projection=diff_map)
+diff_map.plot(cmap=plt.get_cmap('Greys_r'),
```
wtbarnes:
```suggestion
diff_map.plot(cmap='Greys_r',
```

examples/map/difference_images.py
jmason86: I've moved it to the very end in an upcoming commit. We'll see if that breaks anything. If it works, then the gallery template (examples/example_template/example_template.py) should be updated -- I was just following that template.

examples/map/difference_images.py
jmason86: I did the line break for `instrument` and `wave` (and changed to `wavelength`) but didn't put the two times on a single line since that's > 80 characters and breaks PEP8 (unless we ignore that rule?)

examples/map/difference_images.py
```diff
@@ -0,0 +1,62 @@
+"""
+===========================
+Plotting a difference image
+===========================
+How to compute and plot a difference image.
```
nabobalis:
```suggestion

How to compute and plot a difference image.
```

examples/map/difference_images.py
```diff
@@ -0,0 +1,62 @@
...
+# Finally, we'll plot it. We'll apply a colormap and renormalize the
+# intensity so that it shows up well.
+
+plt.figure()
+ax_diff = plt.subplot(projection=diff_map)
+diff_map.plot(cmap='Greys_r',
+              norm=colors.Normalize(vmin=-50, vmax=50))
+plt.show()
+
+#sphinx_gallery_thumbnail_number = 2
```
nabobalis: We should do this everywhere.

examples/map/difference_images.py
```diff
@@ -0,0 +1,63 @@
...
+
+###########################################################################
+# First we'll download a couple images and store them in a
+# `sunpy.map.MapSequence`. This could be from any channel of any imager.
+# Here, we use SDO/AIA 304 Å.
+
+instrument = a.Instrument('AIA')
+wave = a.Wavelength(30 * u.nm, 31 * u.nm)
+result = Fido.search(a.Time('2015-06-18T00:00:00', '2015-06-18T00:00:10') |
+                     a.Time('2015-06-18T01:03:30', '2015-06-18T01:03:35'), instrument, wave)
```
Cadair:
```suggestion
                     a.Time('2015-06-18T01:03:30', '2015-06-18T01:03:35'),
                     instrument,
                     wave)
```

examples/map/difference_images.py
```diff
@@ -0,0 +1,63 @@
...
+###########################################################################
+# And now we can do take the actual difference.
+
+diff = maps[1].data - maps[0].data
+
+###########################################################################
+# But we have to decide what to do with the metadata. For example, what
+# time does this difference image correspond to? The time of the first or
+# second image? The mean time? You'll have to decide what makes most sense
+# for your application. Here we'll just use the metadata from the second
```
Cadair: Could we set some property in the meta to be custom here, just so that there is something for someone who want's to edit it to copy?

examples/map/difference_images.py
Cadair: Why is this showing up at all?!

examples/map/difference_images.py
jmason86: .... weird deja vu. I could've sworn this was one of Will's suggestions too and I committed the change. Hmpf.

examples/map/difference_images.py
jmason86: I'm not opposed. But you have a much better idea than me what property would be best.

Cadair: I think the only one which would show up on the plot would be `DATE-OBS`.

jmason86: Tried this but it crashes unless `date-obs` is a properly formatted timestamp. I tried this useful thing: 

```
meta = maps[1].meta
timediff = Time(maps[1].meta['date-obs']) - Time(maps[0].meta['date-obs'])
meta['date-obs'] = '{} Minute Difference Image'.format(int(round(timediff.sec / 60.)))
diff_map = sunpy.map.Map(diff, meta)
```

but that causes:
```
ValueError: Input values did not match any of the formats where the format keyword is optional ['datetime', 'iso', 'isot', 'yday', 'datetime64', 'fits', 'byear_str', 'jyear_str']
```

Ditto if I just set `meta['date-obs'] = 'Turtle'`. The error occurs at the line:
```
ax_diff = plt.subplot(projection=diff_map)
```

wtbarnes: Yeah, 'date-obs` is going to have to follow a specific format. I think setting the `date-obs` as the second time (i.e. in diff time_2 - time_1).

examples/map/difference_images.py
wtbarnes: I searched through the sphinx-gallery and could not find an example of where this is used. Is the expectation that it just shows up in the example code?

examples/map/difference_images.py
jmason86: Setting it to the second time is what it originally/presently does

examples/map/difference_images.py
nabobalis: Aren't all comments rendered by sphinx? Why would this one be any different.

nabobalis:
What happened to: https://docs.sunpy.org/en/v0.9.9/generated/gallery/maps/base_and_running_difference_mapcubes.html#sphx-glr-generated-gallery-maps-base-and-running-difference-mapcubes-py ?

examples/map/difference_images.py
jmason86: Looks like my sphinx n00bness commented on in the initial PR is turning out to be the sticking point. I have no idea how that stuff works, I just saw that setting the thumbnail ≠ 1 was an option in the template (examples/example_template/example_template.py). Guess we've uncovered something unexpected.

examples/map/difference_images.py
Cadair: ok, lets just leave it. :smile:

Cadair:
This should probably be squash merged.

PR #3628, opened 2019-12-20
Make WCS creation and header creation more lenient

Cadair:
@jmason86 found that the new WCS header parsing was failing on LASCO data. This is a double pronged fix, which a) makes it more likely we can generate a valid FITS header from a map meta and b) falls back to generating a simple 2D WCS like we used to if we can't.

### DIFF
--- a/sunpy/io/fits.py
+++ b/sunpy/io/fits.py
@@ -200,9 +200,9 @@ def header_to_fits(header):
     key_comments = header.pop('KEYCOMMENTS', False)
 
     for k, v in header.items():
-        # Drop any keys which are too long to save into FITS
-        if len(k) > 8:
-            warnings.warn(f"The meta key {k} is too long, dropping from the FITS header.", SunpyUserWarning)
+        # Drop any keys which are too long to save into FITS, or have non-ascii characters
+        if len(k) > 8 or not fits.Card._ascii_text_re.match(str(v)):
+            warnings.warn(f"The meta key {k} is too long or value not ascii, dropping from the FITS header.", SunpyUserWarning)
             continue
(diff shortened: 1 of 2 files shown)


Cadair:
(marked as no changelog because it's an additional fix to the original PR

sunpy/io/fits.py
```diff
@@ -200,9 +200,9 @@ def header_to_fits(header):
     key_comments = header.pop('KEYCOMMENTS', False)
 
     for k, v in header.items():
-        # Drop any keys which are too long to save into FITS
-        if len(k) > 8:
-            warnings.warn(f"The meta key {k} is too long, dropping from the FITS header.", SunpyUserWarning)
+        # Drop any keys which are too long to save into FITS, or have non-ascii characters
+        if len(k) > 8 or not fits.Card._ascii_text_re.match(str(v)):
```
dstansby: Would be nice to split these checks and raise a different more descriptive error in each case (instead of "either this or this is wrong").

dstansby:
@Cadair happy to pick this up and add my suggested changes if you'd like?

Cadair:
@dstansby Always, I probably wont get to it till next week if you don't.

dstansby:
I've opened #3645 in place of this - feel free to push to my branch over there if you want to make any more changes @Cadair

Cadair:
You are more than welcome to push to my PRs ftr. :smile:

dstansby:
👍  - it didn't look like the option was turned on, but maybe I missed something

Cadair:
oh yeah, what's that all about?!

PR #3633, opened 2019-12-23
Update pandas plotting registration

dstansby:
This is the (probably relatively new??) proper way to register the pandas converters.

### DIFF
--- a/changelog/3633.bugfix.rst
+++ b/changelog/3633.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed pandas plotting registration in `sunpy.timeseries`.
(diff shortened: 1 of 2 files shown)


nabobalis:
I added the backport for good measure.

Cadair:
I think it's important to know what the minimum version of pandas this works with is.

nabobalis:
Using git blame on pandas, it seems to be around since the plotting module was created which was 2 ish years ago.

Cadair:
If we are now relying on a min version of pandas, we should update our deps to account for that.

nabobalis:
We should.

Issue #3622, opened 2019-12-19
Map cannot open FITS files directly retrieved from DRMS


bmorris3:
### Description

I'm trying to initialize a map with an output from `drms` and having some trouble.

### Expected behavior

```python
url = 'http://jsoc.stanford.edu/SUM9/D757819487/S00009/continuum.fits'
from sunpy.map import Map
m = Map(url)
```
should produce a map.

### Actual behavior

```python
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-17-1ed630106a5c> in <module>
----> 5 m = Map(url)

//anaconda3/lib/python3.7/site-packages/sunpy/map/map_factory.py in __call__(self, composite, sequence, silence_errors, *args, **kwargs)
    298 
    299         if not len(new_maps):
--> 300             raise RuntimeError('No maps loaded')
    301 
    302         # If the list is meant to be a sequence, instantiate a map sequence

RuntimeError: No maps loaded
```

### Steps to Reproduce

```
import drms
from astropy.io import fits

c = drms.Client()
keys, segments = c.query('hmi.Ic_45s[{0}]'.format('2015.11.02'), 
                         key=drms.const.all, seg='continuum')

# Download the file(s)
url = 'http://jsoc.stanford.edu' + segments.continuum[0]

from sunpy.map import Map

m = Map(url)
```

### System Details

  - SunPy Version: 1.0.6
  - Astropy Version: 4.1 
  - Python Version: 3.7
  - OS information: Mac OS X 10.14.6

Cadair:
So it looks like the reason is there is no unit metadata in the file.
```
In [2]: sunpy.map.Map("continuum.fits")                                                                                                                                                                                                                        
/home/stuart/Git/sunpy/sunpy/map/map_factory.py:294: UserWarning: One of the data, header pairs failed to validate with: Image coordinate units for axis 1 not present in metadata.
Image coordinate units for axis 2 not present in metadata.
See https://docs.sunpy.org/en/stable/code_ref/map.html#fixing-map-metadata` for instructions on how to add missing metadata.
  warnings.warn(f"One of the data, header pairs failed to validate with: {e}")
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-2-f266a6e276fb> in <module>
----> 1 sunpy.map.Map("continuum.fits")

[...]

PR #3646, opened 2020-01-05
Print all possible map loading errors in MapFactory

dstansby:
Sort of inspired by #3622, I *think* we should be printing warnings whenever a map fails to validate for any reason. None of the other errors seem to be triggered at all in the test suite, so I'm not really sure how to trigger them to check the warnings work as intended.

I've labelled this as a technically breaking change, as the code is now more strict when metadata validation failed (e.g. on the IRIS file I edited in the tests).

### DIFF
--- a/sunpy/map/map_factory.py
+++ b/sunpy/map/map_factory.py
@@ -37,9 +37,6 @@ try:
 except ImportError:
     pass
 
-__authors__ = ["Russell Hewett, Stuart Mumford"]
-__email__ = "stuart@mumford.me.uk"
-
 __all__ = ['Map', 'MapFactory']
 
 
@@ -286,13 +283,11 @@ class MapFactory(BasicRegistrationFactory):
             try:
                 new_map = self._check_registered_widgets(data, meta, **kwargs)
                 new_maps.append(new_map)
-            except (NoMatchError, MultipleMatchError, ValidationFunctionError):
+            except (NoMatchError, MultipleMatchError,
+                    ValidationFunctionError, MapMetaValidationError) as e:
                 if not silence_errors:
                     raise
-            except MapMetaValidationError as e:
                 warnings.warn(f"One of the data, header pairs failed to validate with: {e}")
-            except Exception:
-                raise
 
         new_maps += already_maps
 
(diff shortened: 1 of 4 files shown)


wtbarnes:
The test failures look unrelated. Seems like some RHESSI files have changed on a remote server...

sunpy/map/map_factory.py
```diff
@@ -37,9 +37,6 @@
 except ImportError:
     pass
 
-__authors__ = ["Russell Hewett, Stuart Mumford"]
```
wtbarnes: I'm in support of removing the `__authors__` and `__email__` lists everywhere in the codebase, but we should probably ensure that those whose names appear there (i.e. mostly the early developers) are OK with it. This could be offloaded to a separate issue/PR (e.g. "Clean out old author info from the whole codebase")

dstansby: I *think* I remember @Cadair doing it as he went along, but may be mistaken. Would be +1 for doing it together in one PR.

wtbarnes: Doing a quick search on the code base shows that this is only 1/2 left so looks like this is a non-issue 😆

ayshih: Hmm, I wasn't aware of a desire to strip out all of the authors/emails from the codebase.  I'll point out that there are a bunch more if you search for `__author__` instead of `__authors__`, including my information a couple of times.

wtbarnes: Ah ok. That's my mistake. There clearly needs to be a larger discussion on this then. If people, particularly the early contributors such as yourself, would rather leave them in, I'm not opposed to that. My thinking was that we might as well remove them as we're now recording "authorship" through git.

ayshih: I created an issue (#3650) for discussion

nabobalis: I've removed it all in my isort PR since its pointless.

ayshih: I disagree with the blanket "pointless", but that's a discussion to have on #3650

Cadair: +1 for removing this in this particular file btw, it's very wrong!

PR #3648, opened 2020-01-06
Force zero warnings in test suite

dstansby:
It would be great to consciously make sure the right warnings are being thrown in the right places, so lets try turning them all into errors! The test suite will probably 💥 , but I might leave this PR to keep chipping away at the warnings in the hope of one day squashing them all.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -85,6 +85,7 @@ addopts = --doctest-rst -m "not figure" --doctest-ignore-import-errors
 asdf_schema_tests_enabled = true
 asdf_schema_root = sunpy/io/special/asdf/schemas/
 markers =
+    remote_data: marks this test function as needing remote data.
     online: marks this test function as needing online connectivity.
     figure: marks this test function as using hash-based Matplotlib figure verification. This mark is not meant to be directly applied, but is instead automatically applied when a test function uses the @sunpy.tests.helpers.figure_test decorator.
     flaky
--- a/setup.cfg
+++ b/setup.cfg
@@ -128,6 +128,8 @@ filterwarnings =
     ignore:The private astropy._erfa module has been made into its own package
     # See https://github.com/astropy/astropy/pull/10570
     ignore:leap-second auto-update failed
+    # See https://github.com/astropy/extension-helpers/issues/23
+    ignore:Distutils was imported before Setuptools
 
 
 [pycodestyle]
--- a/setup.cfg
+++ b/setup.cfg
@@ -93,7 +93,7 @@ remote_data_strict = True
 # Pin junit behaviour; we might want to update this to xunit2 at some point
 junit_family=xunit1
 filterwarnings =
-    ignore
+    error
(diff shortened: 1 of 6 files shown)


nabobalis:
Is the plan to still split these up into smaller PRs or now this is just the final stretch?

dstansby:
The plan is to do this in smaller PRs as much as possible; I'll mark this as work in progress.

setup.cfg
```diff
@@ -114,6 +114,7 @@ filterwarnings =
     ignore:leap-second auto-update failed
     ignore:defusedxml.lxml is no longer supported:DeprecationWarning
     ignore:Using or importing the ABCs from:DeprecationWarning
+    ignore:direct construction of DocTestModulePlus has been deprecated
```
Cadair: This is actually an issue with pytest 5.4. See astropy/astropy#10039

It might be prudent to pin pytest to `<5.4` rather than ignore this warning. (That should be done on master).

setup.cfg
```diff
@@ -118,6 +118,8 @@ filterwarnings =
     # Warnings about open jp2 library not being found
     ignore:The openjp2 library:UserWarning
     ignore:Neither the openjp2 nor the openjpeg library could be loaded:UserWarning
+# Set the junit version to supress a depreaction warning
+junit_family=xunit1
```
Cadair: I was looking into this the other day and wondering if we actually want to bump it up rather than pin it down. Probably more a question for the azure templates though.

dstansby:
🎉 🎉 🎉 🎉 🎉  getting close!

sunpy/visualization/animator/tests/test_wcs.py
```diff
@@ -11,6 +11,10 @@
 from sunpy.visualization.animator.wcs import ArrayAnimatorWCS
 
 
+pytestmark = pytest.mark.fitlerwarnings('ignore:target cannot be converted to ICRS, so will not be '
```
Cadair, line 14: god I hate this warning.

Cadair:
@dstansby are you winning this battle slowly?! :grinning:

dstansby:
Yep, almost done! The only one now is in the figure tests, which is a rabbit (black)hole I attempted to go down this morning and didn't get very far with. For the sake of getting this done I might just filter the warning and open an issue to fix it at some point.

sunpy/net/dataretriever/sources/norh.py
```diff
@@ -61,7 +61,7 @@ def _get_url_for_timerange(self, timerange, **kwargs):
 
         # We allow queries with no Wavelength but error here so that the query
         # does not get passed to VSO and spit out garbage.
-        if 'wavelength' not in kwargs.keys() or not kwargs['wavelength']:
+        if 'wavelength' not in kwargs.keys():
```
Cadair, line 64:
```suggestion
        if 'wavelength' not in kwargs:
```

Cadair:
Thanks for all your work on this quest @dstansby :rocket:

PR #3652, opened 2020-01-08
Fix warnings in sunpy.coordinates

dstansby:
I'm going to split #3648 into some smaller PRs to make reviewing easier.

No major changes here, just catching some warnings in the coordinate tests.

### DIFF
--- a/sunpy/coordinates/tests/test_frames.py
+++ b/sunpy/coordinates/tests/test_frames.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+import warnings
+
 import numpy as np
 
 import pytest
@@ -411,7 +413,14 @@ def test_skycoord_hpc(args, kwargs):
     sc = SkyCoord(*args, **kwargs, frame="helioprojective",
                   observer='earth', obstime="2011-01-01T00:00:00")
     # Test the transform to HGS because it will force a `make_3d` call.
-    hgs = sc.transform_to("heliographic_stonyhurst")
+    #
+    # astropy emits some warnings here because of invalid NaN comparisons,
+    # which will be removed in a future astropy release
+    # (see https://github.com/astropy/astropy/pull/9843), so filter the warnings out.
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message='invalid value encountered',
+                                category=RuntimeWarning)
+        hgs = sc.transform_to("heliographic_stonyhurst")
 
     assert isinstance(hgs.frame, HeliographicStonyhurst)
 
--- a/sunpy/coordinates/tests/test_sun.py
+++ b/sunpy/coordinates/tests/test_sun.py
@@ -1,3 +1,5 @@
+import warnings
+
 import pytest
 
 import numpy as np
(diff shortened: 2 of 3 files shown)


sunpy/coordinates/tests/test_wcs_utils.py
```diff
@@ -100,7 +103,8 @@ def test_wcs_extras():
               'rsun_ref': 690000000}
     generic_map = sunpy.map.Map((data, header))
 
-    wcs = generic_map.wcs
+    with pytest.warns(FITSFixedWarning, match='Set MJD-OBS to 40587.000000 from DATE-OBS'):
```
dstansby: I think this is only for astropy 4.0, so needs some sort of conditional check.

sunpy/coordinates/tests/test_sun.py
```diff
@@ -263,7 +265,9 @@ def test_carrington_rotation_number(date, day_fraction, rotation_number):
                           ])
 def test_carrington_rotation_starttime(crot, julian_days):
     # Stated precision in the docstring is 0.11 seconds
-    assert_quantity_allclose(sun.carrington_rotation_time(crot).tt.jd * u.day, julian_days * u.day, atol=0.11*u.s)
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=ErfaWarning)
```
Cadair: `ErfaWarning` is a very broad category unfortunately. What exactly are you catching here?

ayshih: > `ErfaWarning` is a very broad category unfortunately.

As far as I know, `ErfaWarning` is virtually always the result of providing "dubious" times to ERFA, where "dubious" is defined by being too far in the past or too far in the future in ERFA's judgment.  Here, it's because the test dates are all far in the past.

sunpy/coordinates/tests/test_sun.py
```diff
@@ -7,7 +9,7 @@
 from astropy.time import Time
 import astropy.units as u
 from astropy.tests.helper import assert_quantity_allclose
-from astropy.utils.exceptions import ErfaWarning
+from astropy.utils.exceptions import AstropyWarning, ErfaWarning
```
ayshih: Why add `AstropyWarning`?  Is this a vestigial edit?

sunpy/coordinates/tests/test_frames.py
```diff
@@ -411,7 +413,11 @@ def test_skycoord_hpc(args, kwargs):
     sc = SkyCoord(*args, **kwargs, frame="helioprojective",
                   observer='earth', obstime="2011-01-01T00:00:00")
     # Test the transform to HGS because it will force a `make_3d` call.
-    hgs = sc.transform_to("heliographic_stonyhurst")
+    # astropy emits some incorrect warnings here, that will be fixed in a
+    # future release, so filter them out
```
ayshih: I think this comment needs elaboration on the precise "incorrect" warnings, because something may change and we may undesirably filter out "correct" warnings.

sunpy/coordinates/tests/test_sun.py
dstansby: It's a hangover from a change I didn't need to make in the end, will remove and squash the commits.

sunpy/coordinates/tests/test_sun.py
Cadair: That certainly seems to be the most common one, but in theory it can be a lot more (any warning from the erfa layer at all iirc.).

Cadair:
If you merge in master the CI should fix.

PR #3655, opened 2020-01-08
Install extension helpers on rtd before install

Cadair:


Cadair:
wait, why did we want to backport this? We only use extension helpers on master right?!

nabobalis:
Yeah dont

PR #3657, opened 2020-01-09
Reduce memory consuption of different observers example

Cadair:
This is to try and fix the RTD issues.


Cadair:
I have tested all the rest locally and none of them clear 1Gb.

PR #3669, opened 2020-01-10
hopefully fix astropy dev build

Cadair:


Cadair:
Yeah, at somepoint we will need to change it and skip on `<=4.0.1`

PR #3675, opened 2020-01-10
add property to access QueryResponse as Table

### LINKED ISSUE #3279: make it easy to access the UnifiedResponse table
if you want to access keys you should be able to get at the table

sashank27:
### Description

Provides for a property to access the `QueryResponse` as an `astropy`table`Table`, which then can be used for Indexing of results.

Example:
```python
result = Fido.search(a.Time('2012/03/04 00:00', '2012/03/04 00:02'),
                     a.Instrument('aia'),
                     a.Wavelength(171*u.angstrom) | a.Wavelength(94*u.angstrom))
for table in result.tables:
    print(table['Start Time','End Time'])
	print()
```
```
Output:
 Start Time [1]       End Time [1]   
------------------- -------------------
2012-03-04 00:00:00 2012-03-04 00:00:01
2012-03-04 00:00:12 2012-03-04 00:00:13
2012-03-04 00:00:24 2012-03-04 00:00:25
2012-03-04 00:00:36 2012-03-04 00:00:37
2012-03-04 00:00:48 2012-03-04 00:00:49
2012-03-04 00:01:00 2012-03-04 00:01:01
2012-03-04 00:01:12 2012-03-04 00:01:13
2012-03-04 00:01:24 2012-03-04 00:01:25
2012-03-04 00:01:36 2012-03-04 00:01:37
2012-03-04 00:01:48 2012-03-04 00:01:49
2012-03-04 00:02:00 2012-03-04 00:02:01

   Start Time [1]       End Time [1]   
------------------- -------------------
2012-03-04 00:00:02 2012-03-04 00:00:03
2012-03-04 00:00:14 2012-03-04 00:00:15
2012-03-04 00:00:26 2012-03-04 00:00:27
2012-03-04 00:00:38 2012-03-04 00:00:39
2012-03-04 00:00:50 2012-03-04 00:00:51
2012-03-04 00:01:02 2012-03-04 00:01:03
2012-03-04 00:01:14 2012-03-04 00:01:15
2012-03-04 00:01:26 2012-03-04 00:01:27
2012-03-04 00:01:38 2012-03-04 00:01:39
2012-03-04 00:01:50 2012-03-04 00:01:51
```
Fixes #3279


sashank27:
@Cadair your review on this?

nabobalis:
Please try not to @ people unnecessarily. Cadair watches this repo and is already notified of all events nor is he the only person who can review pull requests.

changelog/3675.feature.rst
```diff
@@ -0,0 +1 @@
+Creates a `table` property in `~sunpy.net.fido_factory.UnifiedResponse`, which allows to access the `QueryResponse` as an `~astropy.table.Table`, which then can be used for indexing of results.
```
nabobalis:
```suggestion
Creates a ``table`` property for `~sunpy.net.fido_factory.UnifiedResponse`, which allows to access the `QueryResponse` as an `~astropy.table.Table`, which then can be used for indexing of results.
```

sunpy/net/dataretriever/client.py
```diff
@@ -85,15 +85,15 @@ def response_block_properties(self):
...
     def __str__(self):
-        return str(self._build_table())
+        return str(self.build_table())
 
     def _repr_html_(self):
-        return self._build_table()._repr_html_()
+        return self.build_table()._repr_html_()
 
-    def _build_table(self):
+    def build_table(self):
```
nabobalis: I think it's better to keep this as a private function.

sunpy/net/fido_factory.py
```diff
@@ -158,6 +158,24 @@ def response_block_properties(self):
             s.intersection(self.get_response(i).response_block_properties())
         return s
 
+    @property
+    def table(self):
+        """
+        Returns `astopy.table.Table` for all responses present in a specific
```
nabobalis:
```suggestion
        Returns an `astopy.table.Table` for all responses present in a specific
```

sunpy/net/fido_factory.py
```diff
@@ -158,6 +158,24 @@ def response_block_properties(self):
...
+        `~sunpy.net.fido_factory.UnifiedResponse` object. They can then be used
+        to perform key-based indexing of QueryResponses
```
nabobalis:
```suggestion
        to perform key-based indexing of QueryResponses.
```
Can you link to this as you did for UnifiedResponse?

sunpy/net/fido_factory.py
```diff
@@ -158,6 +158,24 @@ def response_block_properties(self):
...
+
+        Returns
+        -------
+        tables : `astropy.table.Table`, list
```
nabobalis:
```suggestion
        `astropy.table.Table`, list
```

sunpy/net/fido_factory.py
```diff
@@ -158,6 +158,24 @@ def response_block_properties(self):
...
+            a list or a single table consisting of data from the QueryResponse
+        """
+
+        tables = []
+        for block in self.responses:
+            tables.append(block.build_table())
+        return tables[0] if len(tables) == 1 else tables
```
nabobalis: So we either get a list of tables or just one table?

sunpy/net/tests/test_fido.py
```diff
@@ -222,6 +223,17 @@ def test_unifiedresponse_slicing_reverse():
...
+@pytest.mark.remote_data
+def test_table():
+    results = Fido.search(
+        a.Time("2012/1/1", "2012/1/5"), a.Instrument("lyra"))
+    table = results.table
+    assert isinstance(table, Table)
+    columns = ['Start Time', 'End Time', 'Source', 'Instrument', 'Wavelength']
+    assert columns == table.colnames
+    assert len(table) == 5
+
```
nabobalis: Can you add a test for more than one table?

sashank27:
> Please try not to @ people unnecessarily. Cadair watches this repo and is already notified of all events nor is he the only person who can review pull requests.

Actually I thought the sub-package maintainer reviews the PR specific to submodules.
Sorry for the trouble and will keep this in mind!

> So we either get a list of tables or just one table?

For most of the scenarious, we generally have a single `UnifiedResponse` returned per search, so it would be better to return it as a single element, rather than returning a one-sized list.
Therefore, if multiple responses are present then we get list of tables, for each specific response. Otherwise it returns only a single Table.

One possible way to maintain consistency is to concatenate the Tables and then return it.
Which one would be better?

nabobalis:
> Actually I thought the sub-package maintainer reviews the PR specific to submodules.
> Sorry for the trouble and will keep this in mind!

They also get a notification. 
 
> For most of the scenarious, we generally have a single `UnifiedResponse` returned per search, so it would be better to return it as a single element, rather than returning a one-sized list.
> Therefore, if multiple responses are present then we get list of tables, for each specific response. Otherwise it returns only a single Table.
> One possible way to maintain consistency is to concatenate the Tables and then return it.
> Which one would be better?

That is a good question. I don't have an answer but I would lean to if there are multiple responses you should probably not concatenate  them just in case. But I will wait for others to comment on it before I would do anything.

sunpy/net/dataretriever/client.py
sashank27:
```python
AttributeError: 'QueryResponse' object has no attribute '_build_table'
```
This occurs if we keep it as a private function.

nabobalis: I wonder why that is. In that case leave it.

sunpy/net/fido_factory.py
vn-ki: I'm not a fan of any API that returns different things at different times. 

@sashank27 Is the returning of list only affected by the user input? In other words, can the user know that they will get a list or table? (Even then this is not elegant, but probably justifiable)

vn-ki: From your tests below, it is not very obvious to me when the `.table` will return a `Table` or a list.

I vote for renaming to `.tables` and always returning a list.

sashank27: If there is only one `QueryResponse` then only a `Table` is returned, otherwise, list.
Yeah, it seems that this is not evident from the Users point of view.
Will shift it to always return a list.

sunpy/net/fido_factory.py
```diff
@@ -158,6 +158,25 @@ def response_block_properties(self):
...
+    def tables(self):
+        """
+        Returns a list of `astropy.table.Table` for all responses present in a specific
+        `~sunpy.net.fido_factory.UnifiedResponse` object. They can then be used
+        to perform key-based indexing of `sunpy.net.dataretriever.client.QueryResponse`
+        objects.
+
+        Returns
+        -------
+        list
```
nabobalis:
```suggestion
        `list`
```

sunpy/net/fido_factory.py
```diff
@@ -158,6 +158,25 @@ def response_block_properties(self):
...
+            a list of `astropy.table.Table`, consisting of data from the QueryResponse
```
nabobalis:
```suggestion
            A list of `astropy.table.Table`, consisting of data from the `~sunpy.net.dataretriever.client.QueryResponse`.
```

sashank27:
> Thanks for the PR @sashank27 it looks good.
> 
> The main issue here is by adding the `tables` property to the unified response, we need to make sure that all response classes (VSO, JSOC and dataretriever) all have a `build_table()` function.

Dataretriver: https://github.com/sunpy/sunpy/blob/3801ddba22c49e3527c215862d7d453820909f09/sunpy/net/dataretriever/client.py#L96-L110

VSO: https://github.com/sunpy/sunpy/blob/3801ddba22c49e3527c215862d7d453820909f09/sunpy/net/vso/vso.py#L182-L237

nabobalis:
We might need a test to ensure that all three response classes have a working `.tables` attribute.

Cadair:
@sashank27 

you need to add a `build_table` function to the jsoc client:

```
diff --git a/sunpy/net/jsoc/jsoc.py b/sunpy/net/jsoc/jsoc.py
index c24a851f6..3f10e0177 100644
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -68,6 +68,9 @@ class JSOCResponse(Sequence):
         else:
             self.table = astropy.table.vstack([self.table, table])
 
+    def build_table(self):
+        return self.table
+
     def response_block_properties(self):
         """
         Returns a set of class attributes on all the response blocks.
```

it would also be good if you could add a test which tests jsoc.

sunpy/net/fido_factory.py
```diff
@@ -158,6 +158,26 @@ def response_block_properties(self):
...
+        to perform key-based indexing of objects of either type
+        `sunpy.net.dataretriever.client.QueryResponse`, `sunpy.net.vso.QueryResponse` or
+        `sunpy.net.jsoc.JSOCClient`
+
+        Returns
+        -------
+        list
+            a list of `astropy.table.Table`, consisting of data from the QueryResponse
+        """
+
```
nabobalis:
```suggestion
```

PR #3683, opened 2020-01-13
Fix Python version for requiring importlib_resources

olebole:
### Description

The `importlib_resources` is replaces by `importlib.resources` already in Python 3.7; so the requirement is adjusted accordingly. Otherwise, the tests fail on Python 3.7 on Debian, since we don't have an `importlib_resources` package.

### DIFF
--- a/changelog/3683.trivial.rst
+++ b/changelog/3683.trivial.rst
@@ -1,0 +1,1 @@
+Fix Python version for requiring importlib_resources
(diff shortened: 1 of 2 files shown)


nabobalis:
Thanks @olebole

nabobalis:
Doesn't need to be backported to 1.0. Mistake on my part.

PR #3685, opened 2020-01-13
Use Giles over SunPy bot for milestones and changelogs

Cadair:
The main advantage to doing this is no more comment from SunPyBot which means no more *multiple* comments from SunPyBot.

All the info that it put in the comment is now shown on the checks page, and we can make the checks page even more detailed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -180,4 +180,4 @@ workflows:
 
 notify:
   webhooks:
-    - url: https://giles.cadair.com/circleci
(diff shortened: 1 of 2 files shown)


nabobalis:
Should something show up?

Cadair:
No we have to merge it first, it only reads the configs off of master.

Cadair:
We should check the documentation link though, that should still show I think.

nabobalis:
There is a giles link?

Cadair:
:shipit:

PR #3686, opened 2020-01-14
Improves missing metadata warning in `sunpy.map`

### LINKED ISSUE #3117: Improve missing metadata warning in Map
If some of the keys for observer metadata are missing it would be good to inform the user which ones they should add. See this comment https://github.com/sunpy/sunpy/pull/3115#discussion_r285666447

Raahul-Singh:
### Description

Updates the `~sunpy.map.mapbase.observer_coordinate` property to warn the user of specific missing metadata for each frame.
Omits warning about frames where all metadata is missing or all meta is present.
Omits warning about a frame when only metadata which is frame independent is missing.

I tried to make this as general as possible.
This will support addition of any new frames that may come up in the future.
Any refactoring suggestions are welcome!

Fixes #3117

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -180,4 +180,4 @@ workflows:
 
 notify:
   webhooks:
-    - url: https://giles.cadair.com/circleci
+    - url: https://giles.cadair.dev/circleci
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,50 +8,32 @@ requires = [
          ]
 build-backend = 'setuptools.build_meta'
 
-[ tool.sunpy-bot ]
-  # disable astropy checks
-  changelog_check = false
-  autoclose_stale_pull_request = false
+[ tool.gilesbot ]
+  [ tool.gilesbot.circleci_artifacts ]
+    enabled = true
 
-  # SunPy Checks
-  check_towncrier_changelog = true
-  check_milestone = true
-  post_pr_comment = true
+  [ tool.gilesbot.circleci_artifacts.giles ]
+    url = "html/index.html"
+    message = "Click details to preview the HTML documentation."
 
-  all_passed_message = """
+  [ tool.gilesbot.pull_requests ]
+    enabled = true
 
-  Thanks for the pull request @{pr_handler.user}! Everything looks great!
-"""
-
-  [ tool.sunpy-bot.towncrier_changelog ]
+  [ tool.gilesbot.towncrier_changelog ]
+    enabled = true
     verify_pr_number = true
     changelog_skip_label = "No Changelog Entry Needed"
-    help_url = "https://github.com/Cadair/sunpy/blob/towncrier/changelog/README.rst"
+    help_url = "https://github.com/sunpy/sunpy/blob/master/changelog/README.rst"
 
-    missing_file_message = """
+    changelog_missing_long = "There isn't a changelog file in this pull request. Please add a changelog file to the `changelog/` directory following the instructions in the changelog [README](https://github.com/sunpy/sunpy/blob/master/changelog/README.rst)."
 
+    type_incorrect_long = "The changelog file you added is not one of the allowed types. Please use one of the types described in the changelog [README](https://github.com/sunpy/sunpy/blob/master/changelog/README.rst)"
 
-* I didn't detect a changelog file in this pull request. Please add a changelog file to the `changelog/` directory following the instructions in the changelog [README](https://github.com/sunpy/sunpy/blob/master/changelog/README.rst).
-"""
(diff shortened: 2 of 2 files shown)


sunpy/map/mapbase.py
```diff
@@ -635,10 +635,62 @@ def observer_coordinate(self):
...
+            if count[frame] != 0 and count[frame] != 3:
+                missing_dict[str(frame)] = []
+                for i in frame_key_dict[frame]:
+                    if i not in meta_present[frame] and i not in shared_meta.keys():
+                        missing_dict[str(frame)].append(i)
+
+        missing_str = ""
+
+        if not all(shared_meta[i] for i in shared_meta):
+            missing_str += "The following frame indipendent metadata is missing : "
```
vn-ki: This whole block can be reduced to (I think)

```python
missing_str += "The following frame indipendent metadata is missing : " + \
				" ".join(str(i) for i in shared_meta if not i) + \
				"\n"
```

sunpy/map/mapbase.py
```diff
@@ -635,10 +635,62 @@ def observer_coordinate(self):
...
+        for frame in count:
+            if count[frame] != 0 and count[frame] != 3:
+                missing_dict[str(frame)] = []
+                for i in frame_key_dict[frame]:
+                    if i not in meta_present[frame] and i not in shared_meta.keys():
+                        missing_dict[str(frame)].append(i)
+
+        missing_str = ""
+
+        if not all(shared_meta[i] for i in shared_meta):
```
vn-ki: Is 
```python
not all(shared_meta)
```

not enough?

sunpy/map/mapbase.py
```diff
@@ -635,10 +635,62 @@ def observer_coordinate(self):
...
+            missing_str += "The following frame indipendent metadata is missing : "
+            for i in shared_meta:
+                if shared_meta[i] is not True:
+                    missing_str += i + " "
+            missing_str += "\n"
+
+        for i in missing_dict:
+            if len(missing_dict[i]) == 0:
+                continue
+            missing_str += "For frame '{:s}' the following metadata is missing : ".format(str(i))
```
vn-ki: Same as above.

sunpy/map/mapbase.py
```diff
@@ -635,10 +635,62 @@ def observer_coordinate(self):
...
+                    meta_present[key_frame_dict[key]].append(key)
+                except KeyError:
+                    meta_present[key_frame_dict[key]] = [key]
+                count[key_frame_dict[key]] += 1
+                if key in shared_meta.keys():
+                    shared_meta[key] = True
+
+        for frame in count:
+            if count[frame] != 0 and count[frame] != 3:
+                missing_dict[str(frame)] = []
```
vn-ki:
```python
missing_dict[str(frame)] = [i for i in frame_key_dict[frame] 
									if i not in meta_present[frame] and i not in shared_meta.keys()]
```

sunpy/map/mapbase.py
Raahul-Singh: `shared_meta` is to keep track of all the meta that will be shared between frames.
`not all(shared_meta)` would evaluate to true as long as there is no `None` or `False` in the keys.
Here I want to keep track of which meta are present and so, I must evaluate over the key values.

sunpy/map/mapbase.py
Raahul-Singh: Further, in case all the frame independent meta are present, the line 
`The following frame independent metadata is missing : `
will be appended for no reason.

sunpy/map/mapbase.py
Raahul-Singh: This is better! Thanks!

sunpy/map/mapbase.py
Raahul-Singh: Applied. Thanks!

sunpy/map/mapbase.py
Raahul-Singh: Thanks!

Raahul-Singh:
Okay, I need some help understanding why these checks failed. :sweat_smile:

nabobalis:
Looks some package has broken

wtbarnes:
It appears the problem crops up when installing the `gdal` package from PyPI though I've no idea why this is even a dependency at all? Looking at [their release history](https://pypi.org/project/GDAL/#history), it seems v3.0.3 was released Monday

sunpy/map/mapbase.py
vn-ki: Oh. Then 

```python
not all(shared_meta.values())
```

should do

PR #3687, opened 2020-01-14
Fix top_right and bottom_left coordinates

### LINKED ISSUE #2908: Tighter definition of the bottom left and top right coordinate map properties required?
### Description

Map provides the bottom left and top right coordinates as properties. The issue is that the meaning of the positions described by these coordinates is not obvious.

### Expected behavior

The expected behavior is that the meaning of these coordinates is consistent.

### Actual behavior

It appears that map.bottom_left_coord is the position of the lower left hand vertex of that pixel, and map.top_right_coord is the position of the upper right hand vertex of that pixel.  Therefore the positions that these coordinates refer to are different relative to the pixel.  This is not explicit to the user.  There may be an implicit assumption that these positions define the extent of the map in the world.

### Steps to Reproduce

```
import numpy as np
import sunpy.data.test 
import sunpy.map
import os
import astropy.units as u

testpath = sunpy.data.test.rootdir

m = sunpy.map.Map((os.path.join(testpath, 'aia_171_level1.fits')))
m_bl = m.bottom_left_coord
m_tr = m.top_right_coord

x, y = np.meshgrid(*[np.arange(v.value) for v in m.dimensions]) * u.pix
ac = m.pixel_to_world(x, y)

ac_bl = np.nanmin(ac.Tx.value), np.nanmin(ac.Ty.value)
ac_tr = np.nanmax(ac.Tx.value), np.nanmax(ac.Ty.value)

print(m_bl, ac_bl)
print(m_tr, ac_tr)
```

### System Details

==========================================================
 SunPy Installation Information

==========================================================

###########
 General
###########
System : Darwin
Processor : i386
Arch : 64bit
SunPy : 1.0.0.dev10506
SunPy_git : 10a1729ebc2c9acce33c849fe531e547a359b5fe
OS: Mac OS X 10.13.6 (i386)

###########
 Required Libraries 
###########
Python: 3.6.8
NumPy: 1.15.4
SciPy: 1.1.0
matplotlib: 3.0.2
Astropy: 3.1.1
Pandas: 0.23.4

###########
 Recommended Libraries 
###########
beautifulsoup: 4.6.3
PyQt: NOT INSTALLED
Zeep: NOT INSTALLED
Sqlalchemy: 1.2.15

dstansby:
Fixes #2908

This correctly selects the bottom left and top left coordinates of the map, such that the square defined by these coordinates is the bounding box of the image. The issue was needing an extra 0.5 pixel offset, to account for the pixel coordinates labeling the centre of the pixels.

This modifies a few figure tests, but I think in a good way, because some axis labels that weren't previously shown are now shown. There's quite a big change in one of the resampled AIA image figures, which needs checking to make sure the change is correct.

Also clarified the doc wording, to hopefully make clear what these properties actually return.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -509,16 +509,22 @@ class GenericMap(NDData):
     @property
     def bottom_left_coord(self):
         """
-        The physical coordinate for the bottom left [0,0] pixel.
+        The physical coordinate for the bottom left corner of the [0,0] pixel.
         """
-        return self.pixel_to_world(0*u.pix, 0*u.pix)
+        # (0, 0) is the centre of the pixel, so offset by half a pixel to get
+        # corner.
+        return self.pixel_to_world(-0.5*u.pix, -0.5*u.pix)
 
     @property
     def top_right_coord(self):
         """
-        The physical coordinate for the top left pixel.
+        The physical coordinate for the top right corner of the top right ([-1, -1]) pixel.
         """
-        return self.pixel_to_world(*self.dimensions)
+        # self.dimensions is the centre of the pixel, so offset by half a pixel to get
+        # corner.
+        x, y = self.dimensions
+        x, y = x + 0.5 * u.pix, y + 0.5 * u.pix
+        return self.pixel_to_world(x, y)
 
     @property
     def center(self):


nabobalis:
Not sure that issue link is correct?

dstansby:
Whoops, should be fixed.

nabobalis:
We want to backport this?

dstansby:
Not sure, it could either be considered a bug fix or a breaking change (and I think the previous documentation wording was ambiguous enough that I'm not sure which way to go, so would be good to hear what others think)

Cadair:
I need to think about this PR a little more, but I would consider this a breaking change and not a bugfix.

dstansby:
Maybe a good option is to create a new property (`image_bounds`?) that does this to use with the plotting limtis, and just fix the docs for the existing properties.

Cadair:
From reading this PR and the original linked issue it feels like they are not describing / solving the same issue?

#2908 says:

> It appears that map.bottom_left_coord is the position of the lower left hand vertex of that pixel, and map.top_right_coord is the position of the upper right hand vertex of that pixel.

This PR appears to be changing the code *to* have that behaviour described as a bug in #2908. Am I just not understanding what is going on?

dstansby:
As the code stands (and from my investigations) they return the coordinates of the centers of the pixels (can be seen because they give integer values of the pixels, which correspond to pixel centers in FITS), not the corners, so I think #2908 is wrong on that point.

Cadair:
Interesting, ok.

So, is the reason you want to change them to the edges just for non-wcsaxes plotting? or does it have other uses?

dstansby:
Well for `wcsaxes` plotting you want to set the limits to show all of the edge pixels anyway presumably, and if you look at how the test images have changed this (for some reason) makes axis labels appear correctly in some cases.

Cadair:
Ok, if the primary motivation for this is plotting, I think we should do as you suggested and create a new attribute for it. I would propose it be private or at least well named because it's not actually the bounding box of the image it's the coordinates of the corners. (we used to have `extent` properties, but I removed them because they were very misleading (essentially assuming aligned data)).

PR #3699, opened 2020-01-16
Remove glymur pins for infra, and exclude 0.9.0 in setup.cfg

Cadair:
Round two.

### DIFF
--- a/sunpy/tests/helpers.py
+++ b/sunpy/tests/helpers.py
@@ -22,7 +22,7 @@ except ImportError:
     SKIP_GLYMUR = True
 else:
     # See if we have a C backend
-    if any((glymur.lib.openjp2.OPENJP2, glymur.lib.openjpeg.OPENJPEG)):
+    if glymur.lib.openjp2.OPENJP2:
         SKIP_GLYMUR = False
     else:
         SKIP_GLYMUR = True
--- a/.rtd-environment.yml
+++ b/.rtd-environment.yml
@@ -13,7 +13,7 @@ dependencies:
   - parfive
   - sqlalchemy
   - scikit-image
-  - glymur<0.9
+  - glymur
   - beautifulsoup4
   - drms
   - python-dateutil
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,7 +12,7 @@ resources:
     type: github
     endpoint: sunpy
     name: OpenAstronomy/azure-pipelines-templates
-    ref: latest-images
+    ref: master
 
 trigger:
   branches:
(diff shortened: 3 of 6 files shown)


nabobalis:
U BREAK

Cadair:
Oh fun. Looks like some other unrelated things have changed in Glymur 0.9 which need patching.

nabobalis:
```
    if any((glymur.lib.openjp2.OPENJP2, glymur.lib.openjpeg.OPENJPEG)):
E   AttributeError: module 'glymur.lib' has no attribute 'openjpeg'
```

dstansby:
Maybe add the exclusions everywhere, and then we can add support for glymur 0.9 in `2.0`? (and maybe backport to minor releases too)

dstansby:
I might(?) have fixed it.

sunpy/tests/helpers.py
```diff
@@ -22,7 +22,7 @@
     SKIP_GLYMUR = True
 else:
     # See if we have a C backend
-    if any((glymur.lib.openjp2.OPENJP2, glymur.lib.openjpeg.OPENJPEG)):
+    if glymur.lib.openjp2.OPENJP2:
```
dstansby: Unless we want to support testing with glymur<0.9 when openjp2 doesn't work but openjpeg does, this change should fix things.

dstansby:
Fun times: `RuntimeError: You must have a version of OpenJPEG at least as high as 2.3.0 before you can read JPEG2000 images with glymur.  Your version is 0.0.0`

sunpy/tests/helpers.py
nabobalis: Does this chance actually mean we have a min version of 0.9.1 for sunpy now?

dstansby:
@wtbarnes why the approval, there's still a broken test 😆

sunpy/tests/helpers.py
dstansby: Or we have a min version of openjpeg of 0.2.3, which was released 4th Oct 2017? (http://www.openjpeg.org/)

nabobalis: I'm not sure how we enforce that, which we can do with python packages?

dstansby: Sorry I got a bit confused, the requirement I suggested above is only for running the tests. Users should be able to go on using openjpeg<2 with glymur<0.9 if they want.

Alternatively we could add some fancier conditional logic here based on `LooseVersion(glymur)` to handle <0.9 and >0.9.

nabobalis: I also got confused, I didn't notice this was a test file.

Cadair: This is to work out if we should skip the openjpeg tests if glymur is installed but openjpeg isn't

Cadair:
**DO NOT MERGE**

nabobalis:
> **DO NOT MERGE**

You don't control me.

wtbarnes:
@dstansby oops. For some reason I thought those failures were unrelated

PR #3709, opened 2020-01-17
List possible TimeSeriesMetaData signatures

dstansby:
This code is a bit of a mess, but I've gone through and listed all the possible signatures to `TimeSeriesMetaData` to help. Also cleaned up some of the logic in the code, but no functionality changes.

### DIFF
--- a/sunpy/timeseries/metadata.py
+++ b/sunpy/timeseries/metadata.py
@@ -20,6 +20,14 @@ class TimeSeriesMetaData:
     multiple `sunpy.timeseries.TimeSeries` metadata to be concatenated in an
     organized fashion.
 
+    Possible signatures::
+
+        TimeSeriesMetaData(meta=dict, timerange=TimeRange, colnames=list)
+        TimeSeriesMetaData(meta=tuple)
+        TimeSeriesMetaData(meta=list(tuples))
+        TimeSeriesMetaData(timerange=TimeRange)
+        TimeSeriesMetaData(timerange=TimeRange, colnames=list)
+
     Parameters
     ----------
     meta : `dict`, `MetaDict`, `tuple`, `list`
@@ -78,8 +86,10 @@ class TimeSeriesMetaData:
     def __init__(self, meta=None, timerange=None, colnames=None):
         self.metadata = []
         # Parse in arguments
-        if not isinstance(meta, type(None)):
-            if isinstance(meta, (dict, MetaDict)) and isinstance(timerange, TimeRange) and isinstance(colnames, list):
+        if meta is not None:
+            if (isinstance(meta, (dict, MetaDict)) and
+                isinstance(timerange, TimeRange) and
+                    isinstance(colnames, list)):
                 # Given a single metadata entry as a dictionary with additional timerange and colnames.
                 self.metadata.append((timerange, colnames, meta))
             elif isinstance(meta, tuple):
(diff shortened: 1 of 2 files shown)


sunpy/timeseries/metadata.py
```diff
@@ -78,8 +86,10 @@ class TimeSeriesMetaData:
     def __init__(self, meta=None, timerange=None, colnames=None):
         self.metadata = []
         # Parse in arguments
-        if not isinstance(meta, type(None)):
-            if isinstance(meta, (dict, MetaDict)) and isinstance(timerange, TimeRange) and isinstance(colnames, list):
+        if meta is not None:
+            if (isinstance(meta, (dict, MetaDict)) and
```
wtbarnes: Is there anyway to format this conditional a bit more nicely?

dstansby: The whole thing, or just this line? Maybe if `isinstance(MetaDict, dict) == True` then it could be `isinstance(meta, dict)`.

wtbarnes: Sorry, I meant just the multi-line `if` statement starting on line 90

dstansby: Agree it's a bit unweildy, but not sure how it can be made simpler.

sunpy/timeseries/metadata.py
```diff
@@ -78,8 +86,10 @@ class TimeSeriesMetaData:
...
+                isinstance(timerange, TimeRange) and
```
Cadair: It's weird to me that the indentation on this line and the one above is different.

sunpy/timeseries/metadata.py
```diff
@@ -90,13 +100,14 @@ def __init__(self, meta=None, timerange=None, colnames=None):
...
-                              "This will mean the metadata can't be linked "
-                              "to columns in data.", SunpyUserWarning)
+            if isinstance(timerange, TimeRange):
+                if isinstance(colnames, list):
+                    self.metadata.append((timerange, colnames, MetaDict()))
+                else:
+                    self.metadata.append((timerange, [], MetaDict()))
+                    warnings.warn("No time range given for metadata. "
+                                  "This will mean the metadata can't be linked "
+                                  "to columns in data.", SunpyUserWarning)
```
Cadair: columns not rows? I am a little confused by this.

dstansby: 🤷‍♂ this is just what was there to begin with, I haven't changed it.

sunpy/timeseries/metadata.py
dstansby: Otherwise PEP8 complains that the line below and the line below that are the same indentation level - not really sure what the fix is.

sunpy/timeseries/metadata.py
```diff
@@ -20,6 +20,14 @@ class TimeSeriesMetaData:
     multiple `sunpy.timeseries.TimeSeries` metadata to be concatenated in an
     organized fashion.
 
+    Possible signatures::
+
+        TimeSeriesMetaData(meta=dict, timerange=TimeRange, colnames=list)
+        TimeSeriesMetaData(meta=tuple)
```
hayesla: Am I missing something here, but it seems that` TimeSeriesMeta` needs a timerange no? If only a meta is given and no timerange the `TimeSeriesMeta` that is returned is empty - i.e. the meta is not saved within the object.

dstansby: You can give the `meta` kwarg a tuple of `(TimeRange, list, dict)` here: https://github.com/sunpy/sunpy/pull/3709/files#diff-5c1988a7acaca3d3c5035b0f0808a8c3R95 I think.

PR #3712, opened 2020-01-17
Package updates

nabobalis:
1. Updated tox by copying astropy
2. Added basic pre-commit for people who want to use one.
3. Removed version.py by again coping astropy.

~Probs more to do.~


.pre-commit-config.yaml
```diff
@@ -0,0 +1,21 @@
+repos:
+-   repo: https://github.com/pre-commit/mirrors-autopep8
```
Cadair: Does this end up applying to the whole file when you commit it? Does that make it likely to generate large style diffs?

nabobalis: Yes. I would suggest we don't allow anyone to use it until we lint the entire package. But I thought to add it now.

setup.cfg
```diff
@@ -102,23 +112,35 @@ markers =
...
 [flake8]
-max-line-length = 100
+max-line-length = 110
+exclude =
+    .git,
+    __pycache__,
+    docs/conf.py,
+    build,
+    sunpy/data/sample.py,
+    sunpy/extern/distro.py,
```
Cadair: can we ignore a whole directory?

nabobalis: That makes more sense, I will check and add it.

setup.py
```diff
@@ -40,12 +45,17 @@
...
-from setuptools.config import read_configuration
-
-from extension_helpers import get_extensions
+VERSION_TEMPLATE = """
+# Note that we need to fall back to the hard-coded version if either
+# setuptools_scm can't be imported or setuptools_scm can't determine the
+# version, so we catch the generic 'Exception'.
+try:
+    from setuptools_scm import get_version
+    version = get_version(root='..', relative_to=__file__)
```
Cadair: An interesting suggestion from the astropy discussion here is to change this to:
```suggestion
    __version__ = get_version(root='..', relative_to=__file__)
```

and then update `__init__.py` so it does `from version import __version__`.

.pre-commit-config.yaml
Cadair: hmmm, I wonder if this is useful before then anyway.

nabobalis: If you dont have pre commit installed it wont do anything or tell you its here.

Cadair: I was more meaning people could install it and use it before we apply it to all the files in the repo, for isort and the more basic checks.

sunpy/__init__.py
```diff
@@ -31,8 +32,6 @@ class UnsupportedPythonError(Exception):
     import builtins
     builtins._SUNPY_SETUP_ = False
```
Cadair: I wonder if we can remove this?

tox.ini
```diff
@@ -1,76 +1,147 @@
...
+# E722 - do not use bare except
+# E901 - SyntaxError or IndentationError
+# E902 - IOError
+# E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree
+# F822: undefined name in __all__
+# F823: local variable name referenced before assignment
+skip_install = true
+description = check code style, e.g. with flake8
+deps = flake8
+commands = flake8 sunpy --count --select=E101,W191,W291,W292,W293,W391,E111,E112,E113,E30,E502,E722,E901,E902,E999,F822,F823
```
Cadair: Should we limit the pre-commit flake 8 to this list of checks?

setup.cfg
```diff
@@ -67,6 +67,16 @@ docs =
   sphinx-gallery
   sunpy-sphinx-theme
   towncrier
+  reproject
+  astroquery
+  jplephem
+format =
```
Cadair: Are we using this for anything?

.pre-commit-config.yaml
nabobalis: Sure but if we never tell people it exists they wont hopefully?

sunpy/__init__.py
nabobalis: Let me check.

tox.ini
nabobalis: I mean I can? I thought the idea was that we check a subset on the CI but make IDE users sort all problems?

setup.cfg
nabobalis: No I just thought that pydoc style would use this in the future to check th entire code base using all these packages.

sunpy/__init__.py
nabobalis: I've removed it, it seems to work locally.

setup.cfg
Cadair: Is it possible to make pre-commit do all this?

nabobalis: In the future I don't see why not.

nabobalis:
**SQUASH MERGE THIS**

tox.ini
Cadair: I like the idea of having the pre-commit be a way to make the style CI pass.

nabobalis: Already done.

setup.cfg
Cadair: I meant we can make pre-commit install these things, so do we really need a new extra for them?

tox.ini
```diff
@@ -1,76 +1,124 @@
 [tox]
-envlist = py{36,37,38}-{offline},online,build_docs,figure
-isolated_build = True
+envlist =
+    py{36,37,38}-test{,-oldestdeps,-devdeps,-online}
```
Cadair: are we doing anything with `-test`? Is is clearer to have `py38-test` than just `py38`?

tox.ini
```diff
@@ -1,76 +1,124 @@
...
     pytest-xdist
-    astropydev: cython
-    astropydev,figure_astropydev: git+https://github.com/astropy/astropy
-    astropy31: astropy<3.2
+
+    # The devdeps factor is intended to be used to install the latest developer version.
+    # of key dependencies.
+    devdeps: cython
+    devdeps,figure_astropydev: git+https://github.com/astropy/astropy
+    devdeps: astropy<3.2
```
Cadair: wait how are we installing git astropy and astropy <3.2?!

tox.ini
```diff
@@ -1,76 +1,124 @@
...
+
+    # All our tests run in parallel which this plugin allows us to.
     pytest-xdist
-    astropydev: cython
-    astropydev,figure_astropydev: git+https://github.com/astropy/astropy
-    astropy31: astropy<3.2
+
+    # The devdeps factor is intended to be used to install the latest developer version.
+    # of key dependencies.
+    devdeps: cython
```
Cadair: Once astropy/astropy#9726 is merged we probably don't need this any more.

tox.ini
Cadair: Also, we should add some more oldest deps to this, like mpl, numpy whatever else we pin in setup.cfg.

setup.cfg
nabobalis: Oh I thought I removed them.

tox.ini
nabobalis: Nope, i can remove it.

tox.ini
nabobalis: Woops.

nabobalis: Woops.

tox.ini
nabobalis: Removed.

setup.cfg
nabobalis: Removed.

sunpy/__init__.py
```diff
@@ -9,6 +9,12 @@
 """
 # Enforce Python version check during package import.
 # This is the same check as the one at the top of setup.py
+from sunpy.util.logger import _init_log
```
Cadair: isort?! :stuck_out_tongue:

tox.ini
```diff
@@ -1,76 +1,128 @@
 [tox]
-envlist = py{36,37,38}-{offline},online,build_docs,figure
-isolated_build = True
+envlist =
+    py{36,37,38}{,-oldestdeps,-devdeps,-online}
+    build_docs
+    figure
+    figure_astropydev
+    conda
```
Cadair:
```suggestion
    py36-conda
```
At least this is what the env is further down?

sunpy/__init__.py
nabobalis: Fixed,

tox.ini
nabobalis: Fixed

nabobalis:
Should check that before we merge this.

Cadair:
It looks like it could well happen, but only with an old version of asdf. Not sure what's happening there.

nabobalis:
Can we delete asdf and just use a real file format like GeoTIFF?

PR #3714, opened 2020-01-20
Refactoring sunpy.net.attrs

Cadair:
The current state of `sunpy.net.attrs` is a bit of a mess as it has grown organically with the rest of net but never had a cleanup itself. The current state is that all the attrs are defined in `net/vso/attrs.py` and we import the ones which are generally applicable up into the `sunpy.net.attrs` namespace.

This PR makes the following changes:

### vso.attrs

All the classes defined in `vso.attrs` which are used outside of the VSO client have been moved into `net/_attrs.py`.

### `AttrWalker`

The `AttrWalker` class is designed to parse a query (collection of `Attr`s) and return a type specific to the client. This PR adds some limited inline documentation (I am planning a followup PR with narrative docs), as well as a basic set of API tests.

The `AttrWalker` was also the last place in sunpy still using `sunpy/util/multimethod` which I have refactored to use `functools.singledispatch` and a modified version thereof which dispatches on the second argument to the function and not the first.

I have now deprecated the use of `MultiMethod`.

### Generic cleanup

* `_Range` is no longer a mixin class but a subclass of `Attr` and should be used accordingly.
* I have endeavoured to cleanup the various `attrs.` namespaces to be free from all other variables so that your tab completion experience is as nice as possible.

## To Do:

Currently there is no deprecation of the classes that used to be in the `vso.attrs` namespace. Before I deprecate these we need to decide if the goal is to have the client specific attrs be only classes used exclusively by those clients (as it is now in this PR) or all the ones applicable to those clients (i.e. also including the generic ones) which would mean this PR isn't an API change for vso at all.

### DIFF
--- a/sunpy/net/_attrs.py
+++ b/sunpy/net/_attrs.py
@@ -12,22 +12,25 @@ from sunpy.time import parse_time, TimeRange
 
 from .attr import Range, SimpleAttr
 
+from sunpy.util.decorators import add_common_docstring
+from sunpy.time.time import _variables_for_parse_time_docstring
 
 __all__ = ['Resolution', 'Detector', 'Sample', 'Level', 'Instrument', 'Wavelength', 'Time']
 
 
+@add_common_docstring(**_variables_for_parse_time_docstring())
 class Time(Range):
     """
     Specify the time range of the query.
 
     Parameters
     ----------
-    start : SunPy time string or `~sunpy.time.TimeRange`.
+    start : {parse_time_types}
         The start time in a format parseable by `~sunpy.time.parse_time` or
         a `sunpy.time.TimeRange` object.
-    end : SunPy Time String
+    end : {parse_time_types}
         The end time of the range.
-    near : SunPy Time String
+    near : {parse_time_types}
         Return a singular record closest in time to this value as possible,
         inside the start and end window. Note: not all providers support this
         functionality.
--- a/sunpy/net/_attrs.py
+++ b/sunpy/net/_attrs.py
@@ -25,10 +25,8 @@ class Time(Range):
     start : SunPy time string or `~sunpy.time.TimeRange`.
         The start time in a format parseable by `~sunpy.time.parse_time` or
         a `sunpy.time.TimeRange` object.
-
     end : SunPy Time String
         The end time of the range.
-
     near : SunPy Time String
         Return a singular record closest in time to this value as possible,
         inside the start and end window. Note: not all providers support this
@@ -86,7 +84,6 @@ class Wavelength(Range):
         ----------
         wavemin : `~astropy.units.Quantity`
             The lower bounds of the range.
-
         wavemax : `~astropy.units.Quantity`
             The upper bound of the range, if not specified it will default to
             the lower bound.
@@ -140,7 +137,7 @@ class Instrument(SimpleAttr):
 
     Parameters
     ----------
-    value : str
+    value : `str`
 
     Notes
     -----
@@ -162,7 +159,7 @@ class Level(SimpleAttr):
 
     Parameters
     ----------
-    value : float or str
+    value : `float` or `str`
 
         The value can be entered in of three ways:
 
@@ -196,7 +193,7 @@ class Detector(SimpleAttr):
 
     Parameters
     ----------
-    value : str
+    value : `str`
     """
     pass
 
@@ -207,7 +204,7 @@ class Resolution(SimpleAttr):
 
     Parameters
     ----------
-    value : float or str
+    value : `float` or `str`
 
         The value can be entered in of three ways:
 
--- a/sunpy/net/_attrs.py
+++ b/sunpy/net/_attrs.py
@@ -1,0 +1,229 @@
+"""
+Implementation of global attrs.
+
+These are defined in here to keep the `sunpy.net.attrs` namespace clean, and to
+prevent circular imports.
+"""
+import collections
+
+import astropy.units as u
+
+from sunpy.time import parse_time, TimeRange
+
+from .attr import Range, SimpleAttr
+
+
+__all__ = ['Resolution', 'Detector', 'Sample', 'Level', 'Instrument', 'Wavelength', 'Time']
+
+
+class Time(Range):
+    """
+    Specify the time range of the query.
+
+    Parameters
+    ----------
+    start : SunPy time string or `~sunpy.time.TimeRange`.
+        The start time in a format parseable by `~sunpy.time.parse_time` or
+        a `sunpy.time.TimeRange` object.
+
+    end : SunPy Time String
+        The end time of the range.
+
+    near : SunPy Time String
+        Return a singular record closest in time to this value as possible,
+        inside the start and end window. Note: not all providers support this
+        functionality.
+
+    """
+    def __init__(self, start, end=None, near=None):
+        if end is None and not isinstance(start, TimeRange):
+            raise ValueError("Specify start and end or start has to be a TimeRange")
+        if isinstance(start, TimeRange):
+            self.start = start.start
+            self.end = start.end
+        else:
+            self.start = parse_time(start)
+            self.end = parse_time(end)
+
+        if self.start > self.end:
+            raise ValueError("End time must be after start time.")
+        self.near = None if near is None else parse_time(near)
+
+        super().__init__(self.start, self.end)
+
+    def __hash__(self):
+        if not (isinstance(self.start, collections.Hashable) and
+                isinstance(self.end, collections.Hashable)):
+            # The hash is the hash of the start and end time
+            return hash((self.start.jd1, self.start.jd2, self.start.scale,
+                         self.end.jd1, self.end.jd2, self.end.scale))
+        else:
+            return super().__hash__()
+
+    def collides(self, other):
+        return isinstance(other, self.__class__)
+
+    def __xor__(self, other):
+        if not isinstance(other, self.__class__):
+            raise TypeError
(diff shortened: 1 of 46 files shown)


sunpy/net/attr.py
```diff
@@ -429,9 +429,12 @@ class AttrWalker:
       type and return a different one.
 
     """
+    @staticmethod
+    def _unknown_type(*args, **kwargs):
+        raise TypeError(f"{args[1]} or any of its parents have not been registered with the AttrWalker")
     def __init__(self):
```
Cadair:
```suggestion

    def __init__(self):
```

sunpy/net/vso/attrs.py
```diff
@@ -536,34 +282,37 @@ def _create(wlk, root, api):
 # AttrAnd and AttrOr obviously are - in the HEK module). If we defined the
 # filter method as a member of the attribute classes, we could only filter
 # one type of data (that is, VSO data).
-filter_results = MultiMethod(lambda *a, **kw: (a[0], ))
+
+@_singledispatch
+def _filter_results(*args, **kwargs):
```
Cadair: Well we don't have any tests for this, so hopefully it's not broken :see_no_evil:

sunpy/net/hek2vso/hek2vso.py
```diff
@@ -92,12 +93,12 @@ def vso_attribute_parse(phrase):
     [<Time(<Time object: scale='utc' format='isot' value=2011-08-09T07:22:38.000>, <Time object: scale='utc' format='isot' value=2011-08-09T08:32:02.000>, None)>, <Source('SDO')>, <Instrument('AIA')>, <Wavelength(210.99999999999997, 210.99999999999997, 'Angstrom')>]
     """
     try:
-        query = [vso.attrs.Time(phrase['event_starttime'],
+        query = [a.Time(phrase['event_starttime'],
                                 phrase['event_endtime']),
-                 vso.attrs.Source(phrase['obs_observatory']),
-                 vso.attrs.Instrument(phrase['obs_instrument'])]
+                 a.vso.Source(phrase['obs_observatory']),
```
nabobalis: Source is still vso specific?

sunpy/util/functools.py
```diff
@@ -0,0 +1,32 @@
...
+import functools
+
+
+__all__ = ['seconddispatch']
+
+
+def seconddispatch(func):
+    """
+    A variant of `functools.singledispatch` which dispatches on type of the second argument.
+
```
nabobalis:
```suggestion
```

Cadair:
Discussed at the community meeting, should deprecate names in `vso.attrs`.

sunpy/net/vso/attrs.py
nabobalis: You add one?

sunpy/net/attr.py
```diff
@@ -261,10 +261,49 @@ def __repr__(self):
...
+    An attribute that represents a range of a value.
+
+    This type of attribute would be applicable for types like Wavelength or Time.
+    The range is inclusive of both the min and ma
+
+    Parameters
+    ----------
+    min_ : `object`
+        The lower bound of the range.
+
```
nabobalis:
```suggestion
```

sunpy/net/_attrs.py
```diff
@@ -0,0 +1,229 @@
...
+    pass
+
+
+class Resolution(SimpleAttr):
+    """
+    Resolution level of the data.
+
+    Parameters
+    ----------
+    value : float or str
```
nabobalis:
```suggestion
    value : `float` or `str`
```

sunpy/net/_attrs.py
```diff
@@ -0,0 +1,229 @@
...
+        self.value = value.to_value(u.s)
+
+
+class Detector(SimpleAttr):
+    """
+    The detector from which the data comes from.
+
+    Parameters
+    ----------
+    value : str
```
nabobalis:
```suggestion
    value : `str`
```

sunpy/net/_attrs.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+class Level(SimpleAttr):
+    """
+    Specifies the data processing level to search for.  The data processing
+    level is specified by the instrument PI.  May not work with all archives.
+
+    Parameters
+    ----------
+    value : float or str
```
nabobalis:
```suggestion
    value : `float` or `str`
```

sunpy/net/_attrs.py
```diff
@@ -0,0 +1,229 @@
...
+                                                            self.unit)
+
+
+class Instrument(SimpleAttr):
+    """
+    Specifies the Instrument name for the search.
+
+    Parameters
+    ----------
+    value : str
```
nabobalis:
```suggestion
    value : `str`
```

sunpy/net/_attrs.py
```diff
@@ -0,0 +1,229 @@
...
+
+    Notes
+    -----
+    More information about each instrument supported by the VSO may be found
+    within the VSO Registry. For a list of instruments see
+    https://sdac.virtualsolar.org/cgi/show_details?keyword=INSTRUMENT.
```
nabobalis: Turn into a `.. note::` section?

sunpy/net/_attrs.py
```diff
@@ -0,0 +1,229 @@
...
+
+        wavemax : `~astropy.units.Quantity`
+            The upper bound of the range, if not specified it will default to
+            the lower bound.
+
+        Notes
+        -----
+        The VSO understands the 'wavelength' in one of three units, Angstroms,
+        kHz or keV. Therefore any unit which is directly convertible to these
+        units is valid input.
```
nabobalis: Turn into a `.. note::`?

sunpy/net/_attrs.py
```diff
@@ -0,0 +1,229 @@
...
+class Wavelength(Range):
+    def __init__(self, wavemin, wavemax=None):
+        """
+        Specifies the wavelength or spectral energy range of the detector.
+
+        Parameters
+        ----------
+        wavemin : `~astropy.units.Quantity`
+            The lower bounds of the range.
+
```
nabobalis:
```suggestion
```

sunpy/net/_attrs.py
```diff
@@ -0,0 +1,229 @@
...
+    end : SunPy Time String
+        The end time of the range.
+
+    near : SunPy Time String
+        Return a singular record closest in time to this value as possible,
+        inside the start and end window. Note: not all providers support this
+        functionality.
+
+    """
```
nabobalis: Update this to use the new time doc add?

sunpy/net/_attrs.py
```diff
@@ -0,0 +1,229 @@
...
+
+    Parameters
+    ----------
+    start : SunPy time string or `~sunpy.time.TimeRange`.
+        The start time in a format parseable by `~sunpy.time.parse_time` or
+        a `sunpy.time.TimeRange` object.
+
+    end : SunPy Time String
+        The end time of the range.
+
```
nabobalis:
```suggestion
```

sunpy/net/_attrs.py
```diff
@@ -0,0 +1,229 @@
...
+class Time(Range):
+    """
+    Specify the time range of the query.
+
+    Parameters
+    ----------
+    start : SunPy time string or `~sunpy.time.TimeRange`.
+        The start time in a format parseable by `~sunpy.time.parse_time` or
+        a `sunpy.time.TimeRange` object.
+
```
nabobalis:
```suggestion
```

sunpy/net/hek2vso/hek2vso.py
Cadair: for now.

Cadair:
> I have a silly question - why is it necessary to put from sunpy.net import attrs as a within the _can_handle_query() for the dataretriever clients (goes and norh)? How will this prevent circular imports?

The issue is that the clients are imported by the `net/__init__.py` file, which also imports the attrs. So when the `net/dataretriever/goes/` files are imported the `sunpy.net.attrs` namespace doesn't exist yet, because we are still importing `sunpy.net`.

Or to put it another way `sunpy.net` imports `sunpy.net.dataretriever.goes` which imports `sunpy.net.attrs` which breaks because `sunpy.net` doesn't exist.

 By putting the imports in the `_can_handle_query` method importing `sunpy.net.dataretriever` doesn't import `sunpy.net.attrs` that import happens at runtime.

sunpy/net/vso/attrs.py
Cadair: I would rather fold it into https://github.com/sunpy/sunpy/issues/3733 really.

Cadair:
Tests are fine other than the HEK having got upset.

sunpy/net/vso/attrs.py
nabobalis: Fine by me.

sunpy/net/vso/attrs.py
```diff
@@ -619,3 +370,24 @@ def _(attr, results):
...
+
+for _name in _deprecated_names:
+    # Dynamically construct a class which inherits the class with the
+    # deprecation warning in the __init__ first and the class it's deprecating
+    # second.
+    cls = type(_name, (_DeprecatedAttr, getattr(_attrs, _name)), {})
+    # Add the new class to the modules namespace
+    setattr(sys.modules[__name__], _name, cls)
+
+__all__ += _deprecated_names
```
nabobalis: 10/10

PR #3719, opened 2020-01-23
Warning for VSO errors

### LINKED ISSUE #3338: Return VSO errors to the user
Currently we hide any VSO error codes from the user at Search time. If there is a reliable way we can extract this info we should throw warnings or similar.

From #3331 we can see that under certain conditions we get an error key in the returned XML from the VSO:
```
DEBUG: VSO Response:
 <soap:Envelope xmlns:VSO="http://virtualsolar.org/VSO/VSOi" xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <soap:Body>
    <VSO:QueryResponse>
      <body>
        <provideritem>
          <error>VSO-D404 Bad Request -- Invalid Time Sample : contains non digit</error>
          <provider>SSC</provider>
          <version>1</version>
        </provideritem>
      </body>
    </VSO:QueryResponse>
  </soap:Body>
</soap:Envelope>
```

The question here is, is this layout consistent over all providers and error types the VSO might throw back at us, so should we be parsing this and raising it as a warning during `VSOClient.search`.

jains8844:
Fixes #3338


jains8844:
For JSOC one, there was some issue. Kindly refer this [https://github.com/sunpy/sunpy/issues/3338#issuecomment-576259172](https://github.com/sunpy/sunpy/issues/3338#issuecomment-576259172)

jains8844:
I formatted the code with black which resulted in so many changes. The main change is just a single line :`) [change](https://github.com/sunpy/sunpy/pull/3719/files#diff-0c4a0061c4c5eefc3ba432fa7a2b1280L367)

nabobalis:
I will have review this PR later on but we can't accept black formatted. I suggest you undo that formatting.

jains8844:
Okay no problem.

jains8844:
Done.

nabobalis:
Thanks.

nabobalis:
Can you construct a test for this? It might be a pain to make but we want to make sure it catches/reports errors how we think it does.

jains8844:
How do I test if some code raises a warning ??

nabobalis:
Using pytest: https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

jains8844:
Okay thanks :)

jains8844:
Implemented the test.

jains8844:
Kindly review this

sunpy/net/tests/test_fido.py
```diff
@@ -229,6 +230,17 @@ def test_vso_unifiedresponse():
     assert isinstance(uresp, UnifiedResponse)
 
 
+@pytest.mark.remote_data
+def test_vso_error():
+    with pytest.warns(SunpyUserWarning,
+        match="VSO-D404 Bad Request -- Invalid Time \
+Sample : contains non digit"):
```
nabobalis: I don't think this line continuation, can you reformat this please?

jains8844: Done

sunpy/net/tests/test_fido.py
```diff
@@ -229,6 +230,16 @@ def test_vso_unifiedresponse():
     assert isinstance(uresp, UnifiedResponse)
 
 
+@pytest.mark.remote_data
+def test_vso_error():
+    with pytest.warns(SunpyUserWarning,
+        match="VSO-D404 Bad Request -- Invalid Time Sample : contains non digit"):
```
dstansby: This seems like an odd warning to be raised - the query below looks fine to me. Do you know why it raises this warning?

Cadair: I think it's https://github.com/sunpy/sunpy/issues/3331

jains8844: Yes it is #3331

nabobalis: Do we want to have a test that might get invalidated down the road?

jains8844: Can you explain a little bit ? I didn't get it

jains8844:
This should be a trivial change right ?

sunpy/net/tests/test_fido.py
nabobalis: The test you wrote is using a bug to raise the warning which could get fixed in the near future.

jains8844: Oh yeah got it. I couldn't find another query so I used this .

jains8844: I got one that throws 404 Not Found error. Would that work ?

nabobalis:
Yeah trivial would work.

sunpy/net/tests/test_fido.py
nabobalis: Can you add it in addition to this test? Just add another with block.

changelog/3719.trivial.rst
```diff
@@ -0,0 +1 @@
+Added support to return errors from VSO as warnings in `sunpy.net.vso`
```
nabobalis:
```suggestion
Errors from a VSO search will now be raised to the user .
```

jains8844:
Commited :)

nabobalis:
Thanks @jains8844

Issue #3688, opened 2020-01-14
Provide a keyword for map sequence to skip over bad files and continue loading.


Cadair:
When loading a map sequence, if one file fails then it kills the whole load without telling you which file it failed on.

jmason86:
e.g., 

```
start_time = '2010-03-01T20:00:00'
end_time = '2010-03-02T04:00:00'

off_sun_earth_line_imager = (a.vso.Source('STEREO_B') &
                             a.Instrument('EUVI') &
                             a.Time(start_time, end_time))

wavelength = a.Wavelength(17 * u.nm, 18 * u.nm)
result_right = Fido.search(wavelength, off_sun_earth_line_imager)
downloaded_files_right = Fido.fetch(result_right)
maps_right = sunpy.map.Map(downloaded_files_right, sequence=True)
```
results in:
```
Traceback (most recent call last):
  File "/Users/jmason86/anaconda3/envs/cme_speed_astropy_dev/lib/python3.7/site-packages/sunpy/io/fits.py", line 81, in read
    pairs.append(HDPair(hdu.data, header))
  File "/Users/jmason86/anaconda3/envs/cme_speed_astropy_dev/lib/python3.7/site-packages/astropy/utils/decorators.py", line 744, in __get__
    val = self.fget(obj)
  File "/Users/jmason86/anaconda3/envs/cme_speed_astropy_dev/lib/python3.7/site-packages/astropy/io/fits/hdu/image.py", line 230, in data
    data = self._get_scaled_image_data(self._data_offset, self.shape)
  File "/Users/jmason86/anaconda3/envs/cme_speed_astropy_dev/lib/python3.7/site-packages/astropy/io/fits/hdu/image.py", line 697, in _get_scaled_image_data
    raw_data = self._get_raw_data(shape, code, offset)
  File "/Users/jmason86/anaconda3/envs/cme_speed_astropy_dev/lib/python3.7/site-packages/astropy/io/fits/hdu/base.py", line 507, in _get_raw_data
    return self._file.readarray(offset=offset, dtype=code, shape=shape)
  File "/Users/jmason86/anaconda3/envs/cme_speed_astropy_dev/lib/python3.7/site-packages/astropy/io/fits/file.py", line 336, in readarray
    buffer=self._mmap)
TypeError: buffer is too small for requested array
```

jmason86:
My suggestion is that it skip the broken file, complete the load, but issue a warning that includes the filename(s) that failed

[...]

PR #3720, opened 2020-01-23
Fix issue #3688

twentyse7en:
parses arguments inside try block.
print the name of first argument which
caused the error and raise the error.

### Description

Fixes #<3688>


twentyse7en:
Merge Request was by mistake .

PR #3721, opened 2020-01-23
Fix issue #3688

### LINKED ISSUE #3688: Provide a keyword for map sequence to skip over bad files and continue loading.

twentyse7en:
parses arguments inside try block.
print the name of first argument which
caused the error and raise the error.

### Description

Fixes #3688


nabobalis:
You will need to pull in the current sunpy master as it will fix the CI.

PR #3722, opened 2020-01-23
Added function to generate user config file and tests

Raahul-Singh:
In continuation of #3009
I've added two tests. Please let me know if any more are required.

### DIFF
--- a/changelog/3722.feature.rst
+++ b/changelog/3722.feature.rst
@@ -1,1 +1,1 @@
-Added `sunpy.util.config.write_config` that copies the default config file to the user's config directory.
+Added `sunpy.util.config.copy_default_config` that copies the default config file to the user's config directory.
--- a/changelog/3722.feature.rst
+++ b/changelog/3722.feature.rst
@@ -1,0 +1,1 @@
+Added `sunpy.util.config.write_config` that copies the default config file to the user's config directory.
--- a/sunpy/util/config.py
+++ b/sunpy/util/config.py
@@ -12,7 +12,7 @@ from sunpy.extern.appdirs import AppDirs
 from sunpy.util.exceptions import SunpyUserWarning
 
 
-__all__ = ['load_config', 'write_config', 'print_config', 'CONFIG_DIR']
+__all__ = ['load_config', 'copy_default_config', 'print_config', 'CONFIG_DIR']
 
 # This is to avoid creating a new config dir for each new dev version.
 # We use AppDirs to locate and create the config directory.
@@ -152,7 +152,7 @@ def _get_user_configdir():
     return configdir
 
 
-def write_config(overwrite=False):
+def copy_default_config(overwrite=False):
     """
     Copies the default sunpy config file to the user's config directory.
 
@@ -178,7 +178,7 @@ def write_config(overwrite=False):
             shutil.copyfile(config_file, user_config_file)
         else:
             message = "User config file already exists. " \
-                      "To overwrite it use `write_config(overwrite=True)`"
+                      "To overwrite it use `copy_default_config(overwrite=True)`"
             warnings.warn(message, SunpyUserWarning)
     else:
         shutil.copyfile(config_file, user_config_file)
--- a/sunpy/util/config.py
+++ b/sunpy/util/config.py
@@ -154,16 +154,20 @@ def _get_user_configdir():
 
 def write_config(overwrite=False):
     """
-    Copies default config file to user config directory.
+    Copies the default sunpy config file to the user's config directory.
+
+    Parameters
+    ----------
+    overwrite : `bool`
+        If True, existing config file will be overwritten.
     """
     config_filename = 'sunpyrc'
-    module_dir = Path(sunpy.__file__).parent
-    config_file = module_dir / 'data' / config_filename
+    config_file = Path(sunpy.__file__).parent / 'data' / config_filename
     user_config_dir = Path(_get_user_configdir())
     user_config_file = user_config_dir / config_filename
 
     if not _is_writable_dir(user_config_dir):
-        raise RuntimeError(f'Could not write to SunPy config directory="{user_config_dir}"')
+        raise RuntimeError(f'Could not write to SunPy config directory {user_config_dir}')
 
     if user_config_file.exists():
         if overwrite:
--- a/sunpy/util/config.py
+++ b/sunpy/util/config.py
@@ -4,11 +4,15 @@ This module provides SunPy's configuration file functionality.
 import os
 import configparser
 from pathlib import Path
+import shutil
+import warnings
 
 import sunpy
 from sunpy.extern.appdirs import AppDirs
+from sunpy.util.exceptions import SunpyUserWarning
 
-__all__ = ['load_config', 'print_config', 'CONFIG_DIR']
+
+__all__ = ['load_config', 'write_config', 'print_config', 'CONFIG_DIR']
 
 # This is to avoid creating a new config dir for each new dev version.
 # We use AppDirs to locate and create the config directory.
@@ -146,3 +150,28 @@ def _get_user_configdir():
     if not _is_writable_dir(configdir):
         raise RuntimeError(f'Could not write to SUNPY_CONFIGDIR="{configdir}"')
     return configdir
+
+
+def write_config(overwrite=False):
+    """
+    Copies default config file to user config directory.
+    """
+    config_filename = 'sunpyrc'
+    module_dir = Path(sunpy.__file__).parent
+    config_file = module_dir / 'data' / config_filename
+    user_config_dir = Path(_get_user_configdir())
+    user_config_file = user_config_dir / config_filename
+
+    if not _is_writable_dir(user_config_dir):
+        raise RuntimeError(f'Could not write to SunPy config directory="{user_config_dir}"')
+
+    if user_config_file.exists():
+        if overwrite:
+            os.remove(user_config_file)
+            shutil.copyfile(config_file, user_config_file)
+        else:
+            message = "User config file already exists. " \
+                      "To overwrite it use `write_config(overwrite=True)`"
+            warnings.warn(message, SunpyUserWarning)
+    else:
+        shutil.copyfile(config_file, user_config_file)
--- a/sunpy/util/config.py
+++ b/sunpy/util/config.py
@@ -171,7 +171,10 @@ def write_config(overwrite=False):
 
     if user_config_file.exists():
         if overwrite:
-            os.remove(user_config_file)
+            message = "User config file already exists. " \
+                      "This will be overwritten with a backup written in the same location."
+            warnings.warn(message, SunpyUserWarning)
+            os.rename(str(user_config_file), str(user_config_file) + ".bak")
             shutil.copyfile(config_file, user_config_file)
         else:
             message = "User config file already exists. " \
--- a/sunpy/util/tests/test_config.py
+++ b/sunpy/util/tests/test_config.py
@@ -10,7 +10,7 @@ from sunpy import config
 from sunpy.util import SunpyUserWarning
 from sunpy.util.config import (get_and_create_sample_dir, get_and_create_download_dir,
                                CONFIG_DIR, print_config, _find_config_files, dirs,
-                               _get_user_configdir, _is_writable_dir, write_config)
+                               _get_user_configdir, _is_writable_dir, copy_default_config)
 
 USER = os.path.expanduser('~')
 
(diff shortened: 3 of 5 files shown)


changelog/3722.feature.rst
```diff
@@ -0,0 +1 @@
+Added function to generate user config file that copies default config file to user config directory.
```
nabobalis:
```suggestion
Added `sunpy.util.config.write_config` that copies the default config file to the user's config directory.
```

sunpy/util/config.py
```diff
@@ -144,3 +148,28 @@ def _get_user_configdir():
     if not _is_writable_dir(configdir):
         raise RuntimeError(f'Could not write to SUNPY_CONFIGDIR="{configdir}"')
     return configdir
+
+
+def write_config(overwrite=False):
+    """
+    Copies default config file to user config directory.
+    """
```
nabobalis: Docstring missing parameters section.

sunpy/util/config.py
```diff
@@ -144,3 +148,28 @@ def _get_user_configdir():
     if not _is_writable_dir(configdir):
         raise RuntimeError(f'Could not write to SUNPY_CONFIGDIR="{configdir}"')
     return configdir
+
+
+def write_config(overwrite=False):
+    """
+    Copies default config file to user config directory.
```
nabobalis:
```suggestion
    Copies  the default  sunpy config file to the user's config directory.
```

sunpy/util/tests/test_config.py
```diff
@@ -80,3 +86,17 @@ def test_get_and_create_sample_dir():
...
+    module_dir = Path(sunpy.__file__).parent
+    config_file = module_dir / 'data' / config_filename
+    user_config_dir = Path(_get_user_configdir())
+    user_config_file = user_config_dir / config_filename
+
+    write_config()
+    assert open(user_config_file, 'r').read() == open(config_file, 'r').read()
+
+    with pytest.warns(SunpyUserWarning):
+        write_config(overwrite=False)
```
nabobalis: Does this test use a real user's directory?

Raahul-Singh: Yes. This is a test to check if the default sunpy config file is copied to the user's config directory.

nabobalis: We can't have a test that actually interferes with a running user's system. 

Can you check this does not affect an existing config file.

Raahul-Singh: It would. That was the whole point of the write function in the original PR. To copy the default config to the user config file. Create one if there isn't any. Overwrite it if the user wishes so.

nabobalis: Yes but we can't have a test actually overwrite a real users config.

Raahul-Singh: Should I make a temporary folder and use it to test everything?

nabobalis: Yes, override the config location, create a fake file and see if it overwrites and all that jazz. There are other config tests around that have done the first bit before.

Raahul-Singh: There's this problem that the `write_config` function by default writes in the user's config location.
I can override it by passing a parameter to the function to allow user-specified locations. But would such an option to write config files wherever the user wants to, make sense?

nabobalis: It would be better to change the location inside the test only.

nabobalis:
Looks like the codestyle build failed:

```
sunpy/util/tests/test_config.py:114:1: W293 blank line contains whitespace
sunpy/util/tests/test_config.py:118:1: W391 blank line at end of file
```

sunpy/util/config.py
```diff
@@ -144,3 +148,33 @@ def _get_user_configdir():
...
+    Copies the default sunpy config file to the user's config directory.
+
+    Parameters
+    ----------
+    overwrite : `bool`
+        If True, existing config file will be overwritten.
+    """
+    config_filename = 'sunpyrc'
+    module_dir = Path(sunpy.__file__).parent
+    config_file = module_dir / 'data' / config_filename
```
wtbarnes: If `module _dir` isn't used anywhere else (it looks like it is not?), I'd recommend doing all of this in one line
```suggestion
    config_file = Path(sunpy.__file__).parent / 'data' / config_filename
```

sunpy/util/tests/test_config.py
```diff
@@ -80,3 +86,32 @@ def test_get_and_create_sample_dir():
...
+
+    tmp_config_dir = tmpdir.mkdir("config_here")
+    os.environ["SUNPY_CONFIGDIR"] = tmp_config_dir.strpath
+
+    assert tmp_config_dir == _get_user_configdir()
+
+    module_dir = Path(sunpy.__file__).parent
+    config_file = module_dir / 'data' / config_filename
+    user_config_dir = tmp_config_dir
+    user_config_file = user_config_dir / config_filename
```
wtbarnes: Same as above.

```suggestion
    config_file = Path(sunpy.__file__).parent / 'data' / config_filename
    user_config_file = tmp_config_dir / config_filename
```

sunpy/util/tests/test_config.py
```diff
@@ -80,3 +86,32 @@ def test_get_and_create_sample_dir():
...
+    # Have to write to the file otherwise its seen as a directory(?!)
+    tmp_file.write("content")
+
+    write_config()
+    assert open(user_config_file, 'r').read() == open(config_file, 'r').read()
+
+    with pytest.warns(SunpyUserWarning):
+        write_config(overwrite=False)
+
+    if not (os.name == "nt"):
```
wtbarnes: This if statement is a bit mysterious. Could you add a clarifying comment?

sunpy/util/config.py
```diff
@@ -144,3 +148,33 @@ def _get_user_configdir():
...
+    user_config_dir = Path(_get_user_configdir())
+    user_config_file = user_config_dir / config_filename
+
+    if not _is_writable_dir(user_config_dir):
+        raise RuntimeError(f'Could not write to SunPy config directory="{user_config_dir}"')
```
wtbarnes:
```suggestion
        raise RuntimeError(f'Could not write to SunPy config directory {user_config_dir}')
```

sunpy/util/tests/test_config.py
Raahul-Singh: I've added the comments that were there in a similar test above.
Let me know if this needs more clarification. I'll change the comments in both places.

sunpy/util/tests/test_config.py
```diff
@@ -80,3 +86,31 @@ def test_get_and_create_sample_dir():
...
+    user_config_file = tmp_config_dir / config_filename
+
+    tmp_file = tmpdir.join("hello.txt")
+    # Have to write to the file otherwise its seen as a directory(?!)
+    tmp_file.write("content")
+
+    write_config()
+    assert open(user_config_file, 'r').read() == open(config_file, 'r').read()
+
+    with pytest.warns(SunpyUserWarning):
```
dpshelio: I normally prefer to have small tests that test one thing only. In this case, I would have one that creates, one that overwrites, and another that doesn't.

sunpy/util/tests/test_config.py
```diff
@@ -80,3 +86,31 @@ def test_get_and_create_sample_dir():
...
+        write_config(overwrite=False)
```
dpshelio: Isn't `overwrite=False` the default behaviour? I would not add it here if that's what we expect by default. Otherwise if someone changes the default we won't notice the test failing.

sunpy/util/config.py
```diff
@@ -144,3 +148,32 @@ def _get_user_configdir():
...
+    config_file = Path(sunpy.__file__).parent / 'data' / config_filename
+    user_config_dir = Path(_get_user_configdir())
+    user_config_file = user_config_dir / config_filename
+
+    if not _is_writable_dir(user_config_dir):
+        raise RuntimeError(f'Could not write to SunPy config directory {user_config_dir}')
+
+    if user_config_file.exists():
+        if overwrite:
+            os.remove(user_config_file)
```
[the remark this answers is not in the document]
Raahul-Singh: When I adopted this PR, it was mentioned that only tests are required for the code so I did not change the `write_config` function.
As for checking the difference, before we remove it, we can write the incoming file in a temporary file in the same folder, compare line by line, print the differences and then overwrite the original. Would this work?
There's a library that might be useful `difflib`

sunpy/util/tests/test_config.py
```diff
@@ -80,3 +86,31 @@ def test_get_and_create_sample_dir():
     # Set the config back
     os.rmdir(new_path)
     config.set('downloads', 'sample_dir', os.path.join(USER, 'sunpy', 'data', 'sample_data'))
+
+
+def test_write_config(tmpdir, undo_config_dir_patch):
+    config_filename = 'sunpyrc'
+
+    # Try to set a manual one (already created)
+    tmp_config_dir = tmpdir.mkdir("config_here")
```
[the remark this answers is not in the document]
Raahul-Singh: I'll change this.
I was following the code and the names that the tests already present in the same file contain.
May I change them also?

sunpy/util/tests/test_config.py
```diff
@@ -80,3 +86,31 @@ def test_get_and_create_sample_dir():
...
+    os.environ["SUNPY_CONFIGDIR"] = tmp_config_dir.strpath
+
+    assert tmp_config_dir == _get_user_configdir()
+
+    config_file = Path(sunpy.__file__).parent / 'data' / config_filename
+    user_config_file = tmp_config_dir / config_filename
+
+    tmp_file = tmpdir.join("hello.txt")
+    # Have to write to the file otherwise its seen as a directory(?!)
+    tmp_file.write("content")
```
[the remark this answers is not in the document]
Raahul-Singh: Ah, that's an oversight on my part. I have removed it. I was using it locally while playing with the tests when I did not know how to mock the folders properly. It wasn't needed now.

sunpy/util/tests/test_config.py
nabobalis: This is already the case?

sunpy/util/config.py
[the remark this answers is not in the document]
nabobalis: I added a warning to be raised and copied it to a backup file.

sunpy/util/tests/test_config.py
```diff
@@ -80,3 +84,68 @@ def test_get_and_create_sample_dir():
...
     os.rmdir(new_path)
     config.set('downloads', 'sample_dir', os.path.join(USER, 'sunpy', 'data', 'sample_data'))
+
+
+def test_write_config_create_new_config(tmpdir, undo_config_dir_patch):
+    config_filename = 'sunpyrc'
+
+    # Try to set a manual one (already created)
+    tmp_config_dir = tmpdir.mkdir("sunpy_test_configdir")
+    os.environ["SUNPY_CONFIGDIR"] = tmp_config_dir.strpath
```
dstansby, line 94: Is it possible to do this environment setting in a context manager?

nabobalis: Or we can monkeypatch it. Not sure I want to fix it now tho but it would be much better to do so.

sunpy/util/config.py
```diff
@@ -146,3 +150,35 @@ def _get_user_configdir():
     if not _is_writable_dir(configdir):
         raise RuntimeError(f'Could not write to SUNPY_CONFIGDIR="{configdir}"')
     return configdir
+
+
+def write_config(overwrite=False):
```
wtbarnes, line 155: Can the name of this be changed to `copy_default_config`? This is a bit more descriptive and accurate.

nabobalis: Done.

sunpy/util/tests/test_config.py
nabobalis: Done.

sunpy/util/config.py
```diff
@@ -146,3 +150,35 @@ def _get_user_configdir():
...
+    user_config_file = user_config_dir / config_filename
+
+    if not _is_writable_dir(user_config_dir):
+        raise RuntimeError(f'Could not write to SunPy config directory {user_config_dir}')
+
+    if user_config_file.exists():
+        if overwrite:
+            message = "User config file already exists. " \
+                      "This will be overwritten with a backup written in the same location."
+            warnings.warn(message, SunpyUserWarning)
```
wtbarnes, line 176: Should we be using the logger here instead of `warnings`?

sunpy/util/config.py
```diff
@@ -146,3 +150,35 @@ def _get_user_configdir():
...
+            os.rename(str(user_config_file), str(user_config_file) + ".bak")
+            shutil.copyfile(config_file, user_config_file)
+        else:
+            message = "User config file already exists. " \
+                      "To overwrite it use `copy_default_config(overwrite=True)`"
+            warnings.warn(message, SunpyUserWarning)
```
wtbarnes, line 182: Same as above

changelog/3722.feature.rst
```diff
@@ -0,0 +1 @@
+Added `sunpy.util.config.write_config` that copies the default config file to the user's config directory.
```
wtbarnes, line 1:
```suggestion
Added `sunpy.util.config.copy_default_config` that copies the default config file to the user's config directory.
```

sunpy/util/config.py
Cadair: no, warnings raised like this are captured by the logger, this is the correct pattern.

PR #3723, opened 2020-01-24
Adds a Unit Range Object Class

### LINKED ISSUE #1515: Create a Unit Range object
Should be able to handle any kind of quantities including times such as wavelength, dimensions (e.g. arcsec).

TimeRange would subclass this object.

Raahul-Singh:
In `TimeRange`, they use `astropy.time.TimeDelta`. This gives range object specific to handle time data.
There is no general purpose `QuantityDelta` that can handle ranges independently. May I implement one, modeling it after `TimeDelta` and import it?

Fixes #1515

### DIFF
--- a/changelog/3723.feature.rst
+++ b/changelog/3723.feature.rst
@@ -1,0 +1,1 @@
+Adds a Unit Range Object class that can handle any kind of quantities such as wavelength, dimensions (e.g. arcsec).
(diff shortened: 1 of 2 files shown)


nabobalis:
When steve opened the issue, im sure he had a use case for it. But yeah maybe in fact a `QuantityDelta ` should exist upstream in astropy?

dstansby:
Yeah, I think this should go upstream to https://github.com/astropy/astropy/ - sorry @Raahul-Singh!

PR #3727, opened 2020-01-25
Map now raises the file that failed to load if it errors on reading.

### LINKED ISSUE #3688: Provide a keyword for map sequence to skip over bad files and continue loading.

twentyse7en:
If one of the file fails to load, an IOError is now raised informing the user which file failed to load and why.

Partially addresses #3688

### DIFF
--- a/changelog/3727.bugfix.rst
+++ b/changelog/3727.bugfix.rst
@@ -1,1 +1,1 @@
-Now when `~sunpy.map.GenericMap` fails to load a file, it will raise the filename.
+Now when `~sunpy.map.GenericMap` fails to load a file, the filename that failed to load will now be part of the error message.
--- a/changelog/3727.bugfix.rst
+++ b/changelog/3727.bugfix.rst
@@ -1,0 +1,1 @@
+Now when `~sunpy.map.GenericMap` fails to load a file, it will raise the filename.
--- a/sunpy/map/map_factory.py
+++ b/sunpy/map/map_factory.py
@@ -19,6 +19,7 @@ from sunpy.io.file_tools import read_file
 from sunpy.io.header import FileHeader
 
 from sunpy.util import expand_list
+from sunpy.util import SunpyUserWarning
 from sunpy.util.metadata import MetaDict
 from sunpy.util.types import DatabaseEntryType
 
@@ -184,55 +185,58 @@ class MapFactory(BasicRegistrationFactory):
         # For each of the arguments, handle each of the cases
         i = 0
         while i < len(args):
+            try:
+                arg = args[i]
 
-            arg = args[i]
+                # Data-header or data-WCS pair
+                if isinstance(arg, SUPPORTED_ARRAY_TYPES):
+                    arg_header = args[i+1]
+                    if isinstance(arg_header, WCS):
+                        arg_header = args[i+1].to_header()
 
-            # Data-header or data-WCS pair
-            if isinstance(arg, SUPPORTED_ARRAY_TYPES):
-                arg_header = args[i+1]
-                if isinstance(arg_header, WCS):
-                    arg_header = args[i+1].to_header()
+                    if self._validate_meta(arg_header):
+                        pair = (args[i], OrderedDict(arg_header))
+                        data_header_pairs.append(pair)
+                        i += 1    # an extra increment to account for the data-header pairing
 
-                if self._validate_meta(arg_header):
-                    pair = (args[i], OrderedDict(arg_header))
-                    data_header_pairs.append(pair)
-                    i += 1    # an extra increment to account for the data-header pairing
+                # A database Entry
+                elif isinstance(arg, DatabaseEntryType):
+                    data_header_pairs += self._read_file(arg.path, **kwargs)
 
-            # A database Entry
-            elif isinstance(arg, DatabaseEntryType):
-                data_header_pairs += self._read_file(arg.path, **kwargs)
+                # Already a Map
+                elif isinstance(arg, GenericMap):
+                    already_maps.append(arg)
 
-            # Already a Map
-            elif isinstance(arg, GenericMap):
-                already_maps.append(arg)
-
-            # URL
-            elif isinstance(arg, str) and _is_url(arg):
-                url = arg
-                path = str(cache.download(url).absolute())
-                pairs = self._read_file(path, **kwargs)
-                data_header_pairs += pairs
-
-            # File system path (file or directory or glob)
-            elif _possibly_a_path(arg):
-                path = pathlib.Path(arg).expanduser()
-                if _is_file(path):
+                # URL
+                elif isinstance(arg, str) and _is_url(arg):
+                    url = arg
+                    path = str(cache.download(url).absolute())
                     pairs = self._read_file(path, **kwargs)
                     data_header_pairs += pairs
-                elif _is_dir(path):
-                    for afile in sorted(path.glob('*')):
-                        data_header_pairs += self._read_file(afile, **kwargs)
-                elif glob.glob(os.path.expanduser(arg)):
-                    for afile in sorted(glob.glob(os.path.expanduser(arg))):
-                        data_header_pairs += self._read_file(afile, **kwargs)
+
+                # File system path (file or directory or glob)
+                elif _possibly_a_path(arg):
+                    path = pathlib.Path(arg).expanduser()
+                    if _is_file(path):
+                        pairs = self._read_file(path, **kwargs)
+                        data_header_pairs += pairs
+                    elif _is_dir(path):
+                        for afile in sorted(path.glob('*')):
+                            data_header_pairs += self._read_file(afile, **kwargs)
+                    elif glob.glob(os.path.expanduser(arg)):
+                        for afile in sorted(glob.glob(os.path.expanduser(arg))):
+                            data_header_pairs += self._read_file(afile, **kwargs)
+
+                    else:
+                        raise ValueError(f'Did not find any files at {arg}')
 
                 else:
-                    raise ValueError(f'Did not find any files at {arg}')
+                    raise ValueError(f"Invalid input: {arg}")
 
-            else:
-                raise ValueError(f"Invalid input: {arg}")
-
-            i += 1
+                i += 1
+            except Exception as e:
+                warnings.warn(f'Error reading file {arg}', SunpyUserWarning)
+                raise
 
         # TODO:
         # In the end, if there are already maps it should be put in the same
--- a/sunpy/map/map_factory.py
+++ b/sunpy/map/map_factory.py
@@ -42,79 +42,50 @@ __all__ = ['Map', 'MapFactory']
 class MapFactory(BasicRegistrationFactory):
     """
     Map(\\*args, \\*\\*kwargs)
-
     Map factory class.  Used to create a variety of Map objects.  Valid map types
     are specified by registering them with the factory.
-
-
     Examples
     --------
     >>> import sunpy.map
     >>> from astropy.io import fits
     >>> import sunpy.data.sample  # doctest: +REMOTE_DATA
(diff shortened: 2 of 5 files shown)


twentyse7en:
@nabobalis can you check why sunpy.sunpy test is failing

nabobalis:
That build failure doesn't seem important.

twentyse7en:
what else I need to do for merging ?

nabobalis:
Nothing for now, reviews will come in and you'll need to respond to those.

sunpy/map/tests/test_map_factory.py
```diff
@@ -152,6 +153,15 @@ def test_errors(self):
...
             sunpy.map.Map(78)
 
+        # If one file failed to load, make sure it's listed in warning msg
+        nonexist_file = os.path.join(filepath, 'nonexist.fits')
+        files = [AIA_171_IMAGE, nonexist_file]
+        with pytest.warns(SunpyUserWarning, match='Error reading file {0}'.format(nonexist_file)):
+            try:
+                sunpy.map.Map(files)
+            except Exception:
+                pass
```
nabobalis: Why was this put into a try block?

sunpy/map/map_factory.py
```diff
@@ -186,55 +187,58 @@ def _parse_args(self, *args, **kwargs):
...
-                url = arg
-                path = str(cache.download(url).absolute())
-                pairs = self._read_file(path, **kwargs)
-                data_header_pairs += pairs
-
-            # File system path (file or directory or glob)
-            elif _possibly_a_path(arg):
-                path = pathlib.Path(arg).expanduser()
-                if _is_file(path):
+            try:
```
nabobalis: What was the need to put the entire block of code in the try? Could this not be focused on the part that does the reading of a file?

sunpy/map/tests/test_map_factory.py
twentyse7en: Second file doesn't exist. So it will throw a value error. Feature added was to provide warning before error. To validate what happened before error , I put into a try block.

twentyse7en:
I initially thought the same. Then implement approach require 4 lines of code, for the second one this lines has to be repeated for all reading.

Is other approach better one ?

nabobalis:
> I initially thought the same. Then implemented approach require 4 lines of code, for the second one this lines has to be repeated for all reading.
> 
> Is other approach better one ?

What needs repeating over and over?

sunpy/map/tests/test_map_factory.py
nabobalis: The value error should be caught by pytest and not by the try block. Please see https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

twentyse7en: Okay

twentyse7en:
try : and exception :

twentyse7en:
I will implement try block only on code that read the file and resubmit.

nabobalis:
> try : and exception :

Why would that need repeating?

nabobalis:
> I will implement try block only on code that read the file and resubmit.

You can just update this PR. No need to resubmit anything.

twentyse7en:
For database entry , there is a read file function, for url there is read file , for file / dir / glob there is also read file , so all read function should be in try : block, right ?

twentyse7en:
Or are you  just mentioning (file / dir / glob) part

nabobalis:
If the issue is with `_read_file` then maybe that should pathed instead of just adding a large try block. 
My only concern might be it become a larger refactor doing that but on the surface that seems to be the best place for it.

nabobalis:
Those line removals should be undone.

twentyse7en:
@nabobalis in `_read_file( )` there is another `read_file' which does the reading. That line is put inside a try block. hope this solves the issue.

nabobalis:
I think it needs tidying up but it works. It does bring up the question of how we should deal with this behaviour by default.

twentyse7en:
@nabobalis  I didn't get your question

nabobalis:
It's fine. It's more aimed at the other maintainers

twentyse7en:
okay , does this request get merged ?

nabobalis:
It needs to be reviewed first.

sunpy/map/map_factory.py
```diff
@@ -43,79 +44,50 @@
...
     * DatabaseEntry
-
     >>> mymap = sunpy.map.Map(db_result)   # doctest: +SKIP
-
     * Lists of any of the above
-
     >>> mymap = sunpy.map.Map(['file1.fits', 'file2.fits', 'file3.fits', 'directory1/'])  # doctest: +SKIP
-
     * Any mixture of the above not in a list
-
```
nabobalis: You need to undo all these line changes.

sunpy/map/map_factory.py
```diff
@@ -128,7 +100,11 @@ def _read_file(self, fname, **kwargs):
         # NOTE: use os.fspath so that fname can be either a str or pathlib.Path
         # This can be removed once read_file supports pathlib.Path
         log.debug(f'Reading {fname}')
-        pairs = read_file(os.fspath(fname), **kwargs)
+        try:
+            pairs = read_file(os.fspath(fname), **kwargs)
+        except Exception as e:
+            warnings.warn(f'failed to read {fname}',SunpyUserWarning)
+            raise
```
nabobalis: I'm not sure a warn and a raise makes any sense. 

The error from the failed attempt at reading a file should already be sent to the user.

nabobalis: I think what we want to do to is more selective filtering based on the exception raised.

If the file does not exist, raise and stop working but if the file fails to be read then raise a warning but continue.

twentyse7en: > I'm not sure a warn and a raise makes any sense.
> 
> The error from the failed attempt at reading a file should already be sent to the user.

issue was "if one file fails then it kills the whole load without telling you which file it failed on."

Initially @jmason86 had a TypeError and error doesn't tell which file caused the error.My attempt was to print name of the file that failed to load.

twentyse7en: > I think what we want to do to is more selective filtering based on the exception raised.
> 
> If the file does not exist, raise and stop working but if the file fails to be read then raise a warning but continue.

By default , If file doesn't exist it raise and stop working.

If the reading of one file fails, Now it raise warning and stop working. So I think instead of raising error for `read_file( )` it's better to skip there.

nabobalis: But if all we want is to report what file failed, we can just add text to the raised expectation and that would be enough.

twentyse7en: It's better to skip files that failed to read as you suggested.
then all it need is to remove the `raise Exception` in current solution.

`read_file` is the part of `sunpy.io.file_tools` that reads the file.

So is this the better approach.

nabobalis:
Hi @twentyse7en, I undid some of the line space changes and tidied up the exception raising.

sunpy/map/tests/test_map_factory.py
```diff
@@ -141,17 +143,25 @@ def test_patterns(self):
...
         directory = pathlib.Path(filepath, nonexist_dir)
         with pytest.raises(ValueError, match=nonexist_dir):
-            maps = sunpy.map.Map(os.fspath(directory))
+            sunpy.map.Map(os.fspath(directory))
 
         with pytest.raises(ValueError, match='Invalid input: 78'):
             # Check a random unsupported type (int) fails
             sunpy.map.Map(78)
 
+        # If one file failed to load, make sure it's listed in warning msg
```
wtbarnes: This tests if an error is raised, not a warning.

sunpy/map/map_factory.py
```diff
@@ -287,7 +287,7 @@ def __call__(self, *args, composite=False, sequence=False, silence_errors=False,
                     ValidationFunctionError, MapMetaValidationError) as e:
                 if not silence_errors:
                     raise
-                warnings.warn(f"One of the data, header pairs failed to validate with: {e}")
+                log.warning(f"One of the data, header pairs failed to validate with: {e}")
```
Cadair: we should still be `warnings.warn` here shouldn't we?

nabobalis: Does that pass it to the logger?

Cadair: yes, the logger captures warnings.

sunpy/map/map_factory.py
```diff
@@ -120,15 +119,19 @@ class MapFactory(BasicRegistrationFactory):
...
         # call a fits file or a jpeg2k file, etc
         # NOTE: use os.fspath so that fname can be either a str or pathlib.Path
         # This can be removed once read_file supports pathlib.Path
         log.debug(f'Reading {fname}')
-        pairs = read_file(os.fspath(fname), **kwargs)
+        try:
+            pairs = read_file(os.fspath(fname), **kwargs)
+        except Exception as e:
+            msg = f'Failed to read {fname} because: {str(e)}'
+            raise IOError(msg)
```
Cadair: We should chain these exceptions so the user can see why the map loaded.
```suggestion
            raise IOError(msg) from e
```

sunpy/map/map_factory.py
```diff
@@ -120,15 +119,19 @@ class MapFactory(BasicRegistrationFactory):
...
+        # File gets read here. This needs to be generic enough to seamlessly
         # call a fits file or a jpeg2k file, etc
         # NOTE: use os.fspath so that fname can be either a str or pathlib.Path
         # This can be removed once read_file supports pathlib.Path
         log.debug(f'Reading {fname}')
-        pairs = read_file(os.fspath(fname), **kwargs)
+        try:
+            pairs = read_file(os.fspath(fname), **kwargs)
+        except Exception as e:
+            msg = f'Failed to read {fname} because: {str(e)}'
```
Cadair: If we do the exception chaining below, we don't need this.
```suggestion
            msg = f"Failed to read {fname}."
```

wtbarnes:
Why does this not need a changelog?

sunpy/map/map_factory.py
nabobalis: Fixed.

sunpy/map/map_factory.py
nabobalis: Fixed.

sunpy/map/map_factory.py
nabobalis: Fixed.

sunpy/map/tests/test_map_factory.py
nabobalis: Fixed.

nabobalis:
> Why does this not need a changelog?

Added a bad changelog.

nabobalis:
**SQUASH MERGE THS**

Cadair:
The windows test fail is real.

wtbarnes:
It looks like the `\` in the path are not being escaped properly at some point

nabobalis:
Trying to fix this windows path issue is painful. I might have to setup a windows venv.

README.rst
```diff
@@ -48,23 +48,6 @@ Developing
 ==========
 
 If you want to develop SunPy you will need to install from GitHub.
-The best way to do this is to create a new conda environment and install the git version of SunPy in it:
```
nabobalis, line 51: Why on earth I ever wrote these god awful instructions I will never know.

nabobalis:
The windows failure was pytest trying to re.match a string so I cheated it.

changelog/3727.bugfix.rst
```diff
@@ -0,0 +1 @@
+Now when `~sunpy.map.GenericMap` fails to load a file, it will raise the filename.
```
Cadair, line 1:
```suggestion
Now when `~sunpy.map.GenericMap` fails to load a file, the filename that failed to load will now be part of the error message.
```

nabobalis:
Thanks @twentyse7en

dstansby:
🎉

PR #3677, opened 2020-01-11
Updates map.draw_rectangle API to be consistent with map.submap

Raahul-Singh:
### Description

With the proposed changes, running the following code will generate the attached figures.

```python
import sunpy.map
import sunpy.data.sample
import matplotlib.pyplot as plt
from astropy.coordinates import SkyCoord
import astropy.units as u

swap_map = sunpy.map.Map(sunpy.data.sample.SWAP_LEVEL1_IMAGE)

top_right = SkyCoord(0 * u.arcsec, -200 * u.arcsec, frame=swap_map.coordinate_frame)
bottom_left = SkyCoord(-900 * u.arcsec, -900 * u.arcsec, frame=swap_map.coordinate_frame)
swap_submap = swap_map.submap(bottom_left, top_right)

fig = plt.figure()
ax = fig.add_subplot(111, projection=swap_submap)
swap_submap.plot()
plt.show()

fig = plt.figure()
ax = fig.add_subplot(111, projection=swap_map)
swap_map.plot()
swap_map.draw_rectangle(bottom_left,top_right=top_right)
plt.show()
```

Figure_1
Figure_2

Fixes #3325

PR #3737, opened 2020-01-28
Adds Rectangle coordinate parser and associated tests.

Raahul-Singh:
@Cadair Is this good?

### DIFF
--- a/changelog/3737.feature.rst
+++ b/changelog/3737.feature.rst
@@ -1,0 +1,1 @@
+Added a helper function (`~sunpy.coordinates.util.get_rectangle_coordinates`) for defining a rectangle in longitude and latitude coordinates.
--- a/sunpy/coordinates/tests/test_utils.py
+++ b/sunpy/coordinates/tests/test_utils.py
@@ -9,7 +9,7 @@ from sunpy.coordinates import sun
 import sunpy.map as smap
 import sunpy.data.test as test
 from sunpy.coordinates import frames
-from sunpy.coordinates.utils import GreatArc
+from sunpy.coordinates.utils import GreatArc, get_rectangle_coordinates
 
 
 @pytest.fixture
@@ -237,3 +237,90 @@ def test_great_arc_different_observer(aia171_test_map):
     np.testing.assert_almost_equal(end.x.value, b2h.x.value)
     np.testing.assert_almost_equal(end.y.value, b2h.y.value)
     np.testing.assert_almost_equal(end.z.value, b2h.z.value)
+
+
+@pytest.fixture
+def rectangle_args():
+    bottom_left = SkyCoord(0 * u.arcsec, 0 * u.arcsec, frame='heliographic_stonyhurst')
+    top_right = SkyCoord(10 * u.arcsec, 10 * u.arcsec, frame='heliographic_stonyhurst')
+    width = 10 * u.arcsec
+    height = 10 * u.arcsec
+
+    return bottom_left, top_right, width, height
+
+
+def test_rectangle_incomplete_input(rectangle_args):
+    bottom_left, top_right, width, height = rectangle_args
+
+    with pytest.raises(ValueError):
+        get_rectangle_coordinates(bottom_left, height=height)
+
+
+def test_rectangle_invalid_input(rectangle_args):
+    bottom_left, top_right, width, height = rectangle_args
+
+    with pytest.raises(TypeError):
+        get_rectangle_coordinates(width, height=height)
+
+
+def test_rectangle_all_parameters_passed(rectangle_args):
+    bottom_left, top_right, width, height = rectangle_args
+
+    with pytest.raises(ValueError):
+        get_rectangle_coordinates(bottom_left, width=width, top_right=top_right, height=height)
+
+
+def test_rectangle_width_height(rectangle_args):
+    bottom_left, top_right, width, height = rectangle_args
+
+    bottom_left_1, top_right_1 = get_rectangle_coordinates(bottom_left, width=width, height=height)
+
+    assert bottom_left.spherical.lon + width == top_right_1.spherical.lon
+    assert bottom_left.spherical.lat + height == top_right_1.spherical.lat
+
+
+def test_rectangle_mismatching_frames_missing_parameters(rectangle_args):
+    bottom_left, top_right, width, height = rectangle_args
+    top_right = SkyCoord(10 * u.arcsec, 10 * u.arcsec, frame='heliographic_carrington')
+
+    with pytest.raises(ValueError):
+        bottom_left, top_right = get_rectangle_coordinates(bottom_left, top_right=top_right)
+
+
+def test_rectangle_top_right(rectangle_args):
+    bottom_left, top_right, width, height = rectangle_args
+
+    bottom_left_1, top_right_1 = get_rectangle_coordinates(bottom_left, top_right=top_right)
+
+    assert bottom_left.spherical.lon == bottom_left_1.spherical.lon
+    assert bottom_left.spherical.lat == bottom_left_1.spherical.lat
+    assert top_right.spherical.lon == top_right_1.spherical.lon
+    assert top_right.spherical.lat == top_right_1.spherical.lat
+
+
+def test_rectangle_bottom_left_different_types(rectangle_args):
+    bottom_left, top_right, width, height = rectangle_args
+
+    bottom_left_1, top_right_1 = get_rectangle_coordinates(bottom_left.frame, width=width, height=height)
+
+    assert bottom_left.spherical.lon + width == top_right_1.spherical.lon
+    assert bottom_left.spherical.lat + height == top_right_1.spherical.lat
+    assert type(bottom_left_1) == type(top_right_1) == type(bottom_left.frame)
+
+    bottom_left_2, top_right_2 = get_rectangle_coordinates(bottom_left, width=width, height=height)
+
+    assert bottom_left.spherical.lon + width == top_right_2.spherical.lon
+    assert bottom_left.spherical.lat + height == top_right_2.spherical.lat
+    assert type(bottom_left_2) == type(top_right_2) == type(bottom_left)
+
+
+def test_rectangle_bottom_left_vector():
+    bottom_left_vector = SkyCoord([0 * u.arcsec, 10 * u.arcsec], [0 * u.arcsec, 10 * u.arcsec],
+                                  frame='heliographic_stonyhurst')
+
+    bottom_left, top_right = get_rectangle_coordinates(bottom_left_vector)
+
+    assert bottom_left.spherical.lon == bottom_left_vector[0].spherical.lon
+    assert bottom_left.spherical.lat == bottom_left_vector[0].spherical.lat
+    assert top_right.spherical.lon == bottom_left_vector[1].spherical.lon
+    assert top_right.spherical.lat == bottom_left_vector[1].spherical.lat
--- a/sunpy/coordinates/utils.py
+++ b/sunpy/coordinates/utils.py
@@ -293,6 +293,30 @@ def get_rectangle_coordinates(bottom_left, *, top_right = None, width: u.deg = N
         The bottom left coordinate of the rectangular region of interest.
     `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
         The top right coordinate of the rectangular region of interest.
+
+    Examples
+    --------
+    >>> import astropy.units as u
+    >>> from astropy.coordinates import SkyCoord
+    >>> from sunpy.coordinates.frames import HeliographicStonyhurst
+    >>> from sunpy.coordinates.utils import get_rectangle_coordinates
+
+    # With bottom left as a SkyCoord, width and height
+    >>> bottom_left = SkyCoord(0 * u.arcsec, 0 * u.arcsec, frame='heliographic_stonyhurst')
+    >>> width = 10 * u.arcsec
+    >>> height = 10 * u.arcsec
+    >>> bottom_left, top_right = get_rectangle_coordinates(bottom_left, width=width, height=height)
+
+    #With bottom left of shape (2,)
+    >>> bottom_left_vector = SkyCoord([0 * u.arcsec, 10 * u.arcsec], [0 * u.arcsec, 10 * u.arcsec], frame='heliographic_stonyhurst')
+    >>> bottom_left, top_right = get_rectangle_coordinates(bottom_left_vector)
+
+    # With bottom left as a BaseCoordinateFrame instance, width and height
+    >>> bottom_left =  HeliographicStonyhurst(0 * u.arcsec, 0 * u.arcsec)
+    >>> width = 10 * u.arcsec
+    >>> height = 10 * u.arcsec
+    >>> bottom_left, top_right = get_rectangle_coordinates(bottom_left, width=width, height=height)
+
     """
     if not (hasattr(bottom_left, 'transform_to') and
             hasattr(bottom_left, 'shape') and
@@ -324,6 +348,11 @@ def get_rectangle_coordinates(bottom_left, *, top_right = None, width: u.deg = N
         top_right = top_right.transform_to(bottom_left)
 
     else:
+        # If bottom left is a ``SkyCoord``, top right is constructed
+        # as a SkyCoord using width and height. If bottom left is a
+        # ``frame``, the top right is typecasted to its respective
+        # frame. This is done to ensure that the output coordinates
+        # are of the same type.
         top_right = SkyCoord(bottom_left.spherical.lon + width,
                             bottom_left.spherical.lat + height,
                             frame=bottom_left)
--- a/sunpy/coordinates/utils.py
+++ b/sunpy/coordinates/utils.py
@@ -5,7 +5,7 @@
 import numpy as np
 
 import astropy.units as u
-from astropy.coordinates import SkyCoord
+from astropy.coordinates import SkyCoord, BaseCoordinateFrame
 
 from sunpy.coordinates import Heliocentric
 
@@ -265,3 +265,70 @@ class GreatArc:
                         obstime=self.obstime,
                         observer=self.observer,
                         frame=Heliocentric).transform_to(self.start_frame)
+
+
+def get_rectangle_coordinates(bottom_left, *, top_right = None, width: u.deg = None, height: u.deg = None):
+    """
+    Specify a rectangular region of interest in longitude and latitude in a given coordinate frame.
+
+    Parameters
+    ----------
+    bottom_left : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The bottom-left coordinate of the rectangle.
+        Supports passing both the bottom left and top right coordinates by passing with a shape of ``(2,)``.
+    top_right : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The top-right coordinate of the rectangle.
+        If in a different frame than ``bottom_left`` and all required metadata for frame conversion is present,
+        ``top_right`` will be transformed to ``bottom_left`` frame.
+    width : `~astropy.units.Quantity`
+        The width of the rectangle.
+        Must be omitted if the coordinates of both corners have been specified.
+    height : `~astropy.units.Quantity`
+        The height of the rectangle.
+        Must be omitted if the coordinates of both corners have been specified.
+
+    Returns
+    -------
+    `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The bottom left coordinate of the rectangular region of interest.
+    `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The top right coordinate of the rectangular region of interest.
+    """
+    if not (hasattr(bottom_left, 'transform_to') and
+            hasattr(bottom_left, 'shape') and
+            hasattr(bottom_left, 'spherical')):
+        raise TypeError("Invalid input, bottom_left must be of type SkyCoord or BaseCoordinateFrame.")
+
+    if (top_right is not None and not ((hasattr(top_right, 'transform_to') and
+                                        hasattr(top_right, 'shape') and
+                                        hasattr(top_right, 'spherical')))):
+        raise TypeError("Invalid input, top_right must be of type SkyCoord or BaseCoordinateFrame.")
+
+    if bottom_left.shape == (2,) and any((x is not None for x in (width, height, top_right))):
+        raise ValueError("Invalid input, if bottom_left.shape == (2,) "
+                         "other parameters should not be passed.")
+
+    if all(x is not None for x in (width, height, top_right)):
+        raise ValueError("Invalid input, width, height and top_right "
+                         "parameters should not be passed simultaneously.")
+
+    if top_right is None and bottom_left.shape != (2,) and not all((width, height)):
+        raise ValueError("Invalid input, either bottom_left and top_right "
+                         "or bottom_left and height and width should be provided.")
+
+    if bottom_left.shape == (2,):
+        top_right = bottom_left[1]
+        bottom_left = bottom_left[0]
+
+    elif top_right is not None:
+        top_right = top_right.transform_to(bottom_left)
+
+    else:
+        top_right = SkyCoord(bottom_left.spherical.lon + width,
+                            bottom_left.spherical.lat + height,
+                            frame=bottom_left)
+
+        if isinstance(bottom_left, BaseCoordinateFrame):
+            top_right = top_right.frame
+
+    return bottom_left, top_right


sunpy/util/util.py
```diff
@@ -195,3 +197,35 @@ def hash_file(path):
...
+                raise IndexError("If top_right is not specified bottom_left must have length two or width and height must be provided.")
+            else:
+                top_right = bottom_left[1]
+                bottom_left = bottom_left[0]
+
+        if type(bottom_left) is not type(top_right):
+            raise ValueError("top_right must be of same type as bottom_left")
+
+        if isinstance(bottom_left, (SkyCoord, BaseCoordinateFrame)):
+            top_right = top_right.transform_to(bottom_left.frame.name)
```
Cadair:
```suggestion
            top_right = top_right.transform_to(bottom_left.frame)
```

sunpy/util/util.py
```diff
@@ -195,3 +197,35 @@ def hash_file(path):
...
+            width = (top_right.spherical.lon - bottom_left.spherical.lon).to(u.deg)   # Getting the difference in Longitudes.
+            height = (top_right.spherical.lat - bottom_left.spherical.lat).to(u.deg)  # Getting the difference in Latitudes.
+
+            if width <= 0*u.deg or height <= 0*u.deg:
+                raise ValueError("top_right must be to the right and above bottom_left.")
+        else:
+            raise ValueError("Invalid value passed for bottom_left")
+
+    elif top_right is None:
+        top_right = SkyCoord(bottom_left.data.lon + width, bottom_left.data.lat + height, frame=bottom_left.frame.name)
```
Cadair:
```suggestion
        top_right = SkyCoord(bottom_left.unit_spherical.lon + width,
                             bottom_left.unit_spherical.lat + height,
                             frame=bottom_left.frame)
```

sunpy/util/util.py
```diff
@@ -195,3 +197,35 @@ def hash_file(path):
...
+                top_right = bottom_left[1]
+                bottom_left = bottom_left[0]
+
+        if type(bottom_left) is not type(top_right):
+            raise ValueError("top_right must be of same type as bottom_left")
+
+        if isinstance(bottom_left, (SkyCoord, BaseCoordinateFrame)):
+            top_right = top_right.transform_to(bottom_left.frame.name)
+
+            width = (top_right.spherical.lon - bottom_left.spherical.lon).to(u.deg)   # Getting the difference in Longitudes.
```
Cadair:
```suggestion
            width = top_right.unit_spherical.lon - bottom_left.unit_spherical.lon   # Getting the difference in Longitudes.
```

sunpy/util/util.py
```diff
@@ -195,3 +197,35 @@ def hash_file(path):
...
+            height = (top_right.spherical.lat - bottom_left.spherical.lat).to(u.deg)  # Getting the difference in Latitudes.
```
Cadair:
```suggestion
            height = top_right.unit_spherical.lat - bottom_left.unit_spherical.lat  # Getting the difference in Latitudes.
```

sunpy/util/util.py
Raahul-Singh: I think it'll be `bottom_left.spherical.lon` rather than `bottom_left.unit_spherical.lon`

Raahul-Singh:
> Thanks @Raahul-Singh let me know if you think any of the following comment makes this less useful for your other PR, I am mostly focusing on this as a standalone function here.
> 
> The scope for this function is defining a 2D rectangle in world coordinates, as it's restricted to 2D we can only work in unit spherical coordinate representations, which means we are in degrees for width and height everywhere. I have suggested inline some changes to the coordinates functions to make this explicit.
> 
> On the general logic of this function, I think given we are scoping this to a 2D rectangle in spatial coordinates we can assert some things at the start of the function.
> 
>     * `bottom_left` and (if specified) `top_right` must be `SkyCoord` or `BaseCoordinateFrame` objects (they don't have to both be one or the other you can mix those two types. So we can make the first type check if statement at the top of the function something like:
> 
> 
> ```python
> if (not isinstance(bottom_left, (SkyCoord, BaseCoordinateFrame)) and 
>   not isinstance(top_right, (None, SkyCoord, BaseCoordinateFrame))):
>     raise TypeError("...")
> ```
> 
> As we are accepting either of these classes as input for both these coordinates, after this at the top of the function I would convert any `SkyCoord` objects to `BaseCoordinateFrame` objects (by `SkyCoord.frame) to make things simpler in the rest of the function.
> 
>     * I would also rearrange the flow logic to be as flat (not nested) as you can. It's a good Python paradigm to do tests for what things aren't at the top of the function and raise errors or return. i.e.
> 
> 
> ```python
> if not isinstance(thing, int):
>     raise TypeError
> ```
> 
> is better than
> 
> ```python
> if isinstance(thing, int):
>     # do a few lines of thing with int
> raise TypeError
> ```
> 
> So I would suggest you try and work out a way to sanitise your input and then process it and hopefully you will have a couple less nested ifs.

I'm not sure if I understand the conversion from `SkyCoord` to `BaseCoordinateFrame` properly.
`SkyCoord.frame` converts a `astropy.coordinates.sky_coordinate.SkyCoord` to whatever frame that SkyCoord is in, rather than converting it to a `BaseCoordinateFrame`

Cadair:
Sorry, I meant "a subclass of `BaseCoordinateFrame`.

sunpy/util/util.py
```diff
@@ -195,3 +197,32 @@ def hash_file(path):
...
     return sha256.hexdigest()
+
+
+def get_rectangle_coordinates(bottom_left, *, top_right = None, width: u.deg = None, height: u.deg = None):
+
+    if (not isinstance(bottom_left, (SkyCoord, BaseCoordinateFrame)) and 
+        not isinstance(top_right, (None, SkyCoord, BaseCoordinateFrame))):
+        raise TypeError("bottom_left and top_right must be of type SkyCoord or BaseCoordinateFrame")
+
+    elif len(bottom_left.shape) != 0 and bottom_left.shape[0] == 2:
```
Cadair: Would this work:
```suggestion
    elif bottom_left.shape) == (2,):
```

Cadair:
I suggest you convert all your `.spherical` to `.unit_spherical` so that it's really clear we are only operating in 2D

Raahul-Singh:
> .unit_spherical

I can't find `unit_spherical` function, the [astropy docs](https://docs.astropy.org/en/stable/coordinates/skycoord.html) do not list it in either `SkyCoord` or `SkyCoord.frame`

Raahul-Singh:
> Sorry, I meant "a subclass of `BaseCoordinateFrame`".

Since you asked me to let you know if any of these suggestions would affect the other PR, changing a `SkyCoord` to a `BaseCoordinateFrame` would require some refactoring on the the other PR. Nothing significant though. Only `_get_lon_lat` method would need to be replaced.

Raahul-Singh:
> I suggest you convert all your `.spherical` to `.unit_spherical` so that it's really clear we are only operating in 2D

May I add this in a docstring?

Cadair:
is it `.unitspherical`? there is definitely a way of getting the unit spherical representation of a frame.

Raahul-Singh:
> is it `.unitspherical`? there is definitely a way of getting the unit spherical representation of a frame.

I can't find any direct function to get the unit spherical representation of a `SkyCoord`. There is a way to use `.represent_as(UnitSphericalRepresentation)`.

Cadair:
If there isn't a shortcut we can leave it, it's good enough :smile:

sunpy/util/util.py
```diff
@@ -195,3 +197,32 @@ def hash_file(path):
...
+        if width <= 0*u.deg or height <= 0*u.deg:
+            raise ValueError("top_right must be to the right and above bottom_left.")
+
+    elif width is not None and height is not None:
+        top_right = SkyCoord(bottom_left.spherical.lon + width,
+                             bottom_left.spherical.lat + height,
+                             frame=bottom_left.frame)
+
+    else:
+        raise ValueError("Invalid input, either bottom_left and top_right \
```
Cadair: We aren't checking for the situation where the user specifies *all* the keyword arguments (or where `bottom_left` is shape 2 and width and height are passed) we should do this at the top.

sunpy/util/util.py
```diff
@@ -195,3 +197,32 @@ def hash_file(path):
             sha256.update(data)
 
     return sha256.hexdigest()
+
+
+def get_rectangle_coordinates(bottom_left, *, top_right = None, width: u.deg = None, height: u.deg = None):
+
```
Cadair: needs a docstring :smile:

sunpy/util/util.py
```diff
@@ -195,3 +197,32 @@ def hash_file(path):
...
+    if (not isinstance(bottom_left, (SkyCoord, BaseCoordinateFrame)) and 
+        not isinstance(top_right, (None, SkyCoord, BaseCoordinateFrame))):
+        raise TypeError("bottom_left and top_right must be of type SkyCoord or BaseCoordinateFrame")
+
+    elif bottom_left.shape == (2,):
+        top_right = bottom_left[1]
+        bottom_left = bottom_left[0]
+
+    elif top_right is not None:
+        width = top_right.spherical.lon - bottom_left.spherical.lon   # Getting the difference in Longitudes.
```
Cadair: The comments here are probably not needed, it's reasonably obvious that's what happening from the code.

sunpy/util/util.py
```diff
@@ -195,3 +197,32 @@ def hash_file(path):
...
+    elif width is not None and height is not None:
+        top_right = SkyCoord(bottom_left.spherical.lon + width,
+                             bottom_left.spherical.lat + height,
+                             frame=bottom_left.frame)
+
+    else:
+        raise ValueError("Invalid input, either bottom_left and top_right \
+                         or bottom_left and height and width must be provided")
+
+    return (bottom_left, top_right)
```
Cadair: multiple return is implicitly a tuple
```suggestion
    return bottom_left, top_right
```

Raahul-Singh:
> If there isn't a shortcut we can leave it, it's good enough smile

Is anything else required here?

Raahul-Singh:
And finally, I learnt how to squash and rebase successfully without screwing things up! :smile:

sunpy/util/util.py
```diff
@@ -195,3 +197,54 @@ def hash_file(path):
...
+            The height of the rectangle.
+            Must be omitted if ``bottom_left`` has shape ``(2,)`` or ``top_right`` is provided.
+
+    """
+    if ((all((width, height)) and (top_right is not None or bottom_left.shape == (2,))) or
+        ((top_right is None and bottom_left.shape != (2,)) and not all((width, height)))):
+        raise ValueError("Invalid input, either bottom_left and top_right \
+                          or bottom_left and height and width must be provided")
+
+    elif (not isinstance(bottom_left, (SkyCoord, BaseCoordinateFrame)) and
```
Cadair: This can just be a separate statement
```suggestion
    if (not isinstance(bottom_left, (SkyCoord, BaseCoordinateFrame)) and 
```

sunpy/util/util.py
```diff
@@ -195,3 +197,54 @@ def hash_file(path):
...
+        width : `astropy.units.Quantity`
+            The width of the rectangle.
+            Must be omitted if ``bottom_left`` has shape ``(2,)`` or ``top_right`` is provided.
+
+        height : `astropy.units.Quantity`
+            The height of the rectangle.
+            Must be omitted if ``bottom_left`` has shape ``(2,)`` or ``top_right`` is provided.
+
+    """
+    if ((all((width, height)) and (top_right is not None or bottom_left.shape == (2,))) or
```
Cadair: For the sake of both readability of this if statement :grinning: and the specificity of the error I suggest you break this up into one if statement which raises an error "if bottom_left and top_right specified width and height must be None" and one which says the opposite.

sunpy/util/util.py
```diff
@@ -195,3 +197,54 @@ def hash_file(path):
...
     return sha256.hexdigest()
+
+
+def get_rectangle_coordinates(bottom_left, *, top_right = None, width: u.deg = None, height: u.deg = None):
+    """
+    Returns bottom_left and top_right coordinates of a 2D rectangular region of interest.
+
+    Parameters
+    ----------
+        bottom_left : `astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
```
Cadair: iirc there shouldn't be this indent after the heading, so all these lines should be dedented 4 spaces.

sunpy/util/util.py
```diff
@@ -195,3 +197,54 @@ def hash_file(path):
...
+            ``bottom_left`` has shape ``(2,)``.
+
+        width : `astropy.units.Quantity`
+            The width of the rectangle.
+            Must be omitted if ``bottom_left`` has shape ``(2,)`` or ``top_right`` is provided.
+
+        height : `astropy.units.Quantity`
+            The height of the rectangle.
+            Must be omitted if ``bottom_left`` has shape ``(2,)`` or ``top_right`` is provided.
+
```
Cadair: You should add a 
```
Returns
-------
```
section describing the return values.

sunpy/util/util.py
```diff
@@ -195,3 +197,54 @@ def hash_file(path):
             sha256.update(data)
 
     return sha256.hexdigest()
+
+
+def get_rectangle_coordinates(bottom_left, *, top_right = None, width: u.deg = None, height: u.deg = None):
+    """
+    Returns bottom_left and top_right coordinates of a 2D rectangular region of interest.
```
Cadair:
```suggestion
    Specify a rectangular region of interest in spherical world coordinates.
```

sunpy/util/util.py
Cadair:
```suggestion
        bottom_left : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
```

sunpy/util/util.py
```diff
@@ -195,3 +197,54 @@ def hash_file(path):
...
+
+    Parameters
+    ----------
+        bottom_left : `astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+            The bottom_left coordinate of the rectangle. If a `SkyCoord` it can
+            have shape ``(2,)`` and also define ``top_right``.
+
+        top_right : `astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
```
Cadair:
```suggestion
        top_right : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
```

sunpy/util/tests/test_util.py
```diff
@@ -62,3 +64,42 @@ def test_partial_key_match():
...
+
+def test_get_rectangle_coordinates():
+
+    bottom_left = SkyCoord(0 * u.arcsec, 0 * u.arcsec, frame='galactic')
+    top_right = SkyCoord(10 * u.arcsec, 10 * u.arcsec, frame='galactic')
+    width = 10 * u.arcsec
+    height = 10 * u.arcsec
+    
+    with pytest.raises(ValueError):
+        util.get_rectangle_coordinates(bottom_left)
```
Cadair: Good testing practice is for each test function to test one thing, so each call to `get_rectangle_coordinates` should be in it's own test function. To prevent repeating yourself with the parameters for the calls, I suggest you make a fixture for then all:

```python
@pytest.fixture
def rectangle_args():
    ...
    return bottom_left, top_right, width, height

def test1(rectangle_args):
    bottom_left, top_right, width, height = rectangle_args
    ...
```

Cadair:
@ayshih Do you have opinions on if this should live in `sunpy.coordinates.utils`?

ayshih:
@Cadair Conceptually, I think having this in `sunpy.coordinates.utils` probably makes more sense than anywhere else.  That said, the return has me a little bothered: I'd expect a sequence of coordinates that outline the rectangle (e.g., four corners as opposed to only two).

Cadair:
:+1: on `coordinates.utils` @Raahul-Singh can you move it there please?

On the return value, four coordinates is redundant information, you only need two to describe the rectangle given the constraints in this function (as far as I see it). Also this came out of `submap` and `draw_rectangle` both of which fit into this two coordinate model quite well.

ayshih:
Sure, it's redundant, but that redundancy is useful if I wanted to, say, plot the rectangle.  It seems that we can do the additional manipulations for the user, and if the user doesn't need the redundant corners, they can just ignore them.  It depends on the expectation of how a user would use this function.  Certainly we could have a separate function to specifically plot a rectangle based on two coordinates.

Cadair:
My intended use case for this is mostly as a utility for developers not users, I want to use it to validate and process rectangles in a way we can reuse everywhere we want users to specify rectangles.

> doesn't this code completely fail to treat the case where the provided coordinates are not in the same frame?

it used to do this, it might have gone walkies.

Raahul-Singh:
> Unless I'm missing something, doesn't this code completely fail to treat the case where the provided coordinates are not in the same frame?

I think I accidentally deleted that conversion while refactoring the code. I have added it now.

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,69 @@ def coordinates(self, points=None):
...
+    if all((width, height, top_right)):
+        raise ValueError("Invalid input, width, height and top_right \
+                          parameters should not be passed simultaneously.")
+
+    if top_right is None and bottom_left.shape != (2,) and not all((width, height)):
+        raise ValueError("Invalid input, either bottom_left and top_right \
+                         or bottom_left and height and width should be provided.")
+
+    if (not isinstance(bottom_left, (SkyCoord, BaseCoordinateFrame)) and 
+        not isinstance(top_right, (None, SkyCoord, BaseCoordinateFrame))):
```
ayshih: Not a big deal, but a duck-typing approach would be nicer than explicitly checking the type.  That is, checking for the existence of a `.transform_to()` method and a `.spherical` property, which is all that you use.

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,69 @@ def coordinates(self, points=None):
...
+        raise TypeError("bottom_left and top_right must be of type SkyCoord or BaseCoordinateFrame")
+
+    elif bottom_left.shape == (2,):
+        top_right = bottom_left[1]
+        bottom_left = bottom_left[0]
+
+    elif top_right is not None:
+        top_right = top_right.transform_to(bottom_left.frame)
```
ayshih:
```suggestion
        top_right = top_right.transform_to(bottom_left)
```
The `.frame` errors if the input is a frame instance rather than a SkyCoord, whereas the transform will still work if supplied a SkyCoord instead of a frame.

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,69 @@ def coordinates(self, points=None):
...
+        width = top_right.spherical.lon - bottom_left.spherical.lon
+        height = top_right.spherical.lat - bottom_left.spherical.lat
+
+        if width <= 0 * u.deg or height <= 0 * u.deg:
+            raise ValueError("top_right must be to the right and above bottom_left.")
+
+    else:
+        top_right = SkyCoord(bottom_left.spherical.lon + width,
+                             bottom_left.spherical.lat + height,
+                             frame=bottom_left.frame)
```
ayshih:
```suggestion
                             frame=bottom_left)
```
The `.frame` errors if the input is a frame instance rather than a SkyCoord, whereas the constructor will still work if supplied a SkyCoord instead of a frame.

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,69 @@ def coordinates(self, points=None):
...
+    elif top_right is not None:
+        top_right = top_right.transform_to(bottom_left.frame)        
+        width = top_right.spherical.lon - bottom_left.spherical.lon
+        height = top_right.spherical.lat - bottom_left.spherical.lat
+
+        if width <= 0 * u.deg or height <= 0 * u.deg:
+            raise ValueError("top_right must be to the right and above bottom_left.")
+
+    else:
+        top_right = SkyCoord(bottom_left.spherical.lon + width,
```
ayshih: So, `top_right` will always be a SkyCoord, even if `bottom_left` is supplied as a frame instance?  That's a bit weird.

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,69 @@ def coordinates(self, points=None):
...
+        not isinstance(top_right, (None, SkyCoord, BaseCoordinateFrame))):
+        raise TypeError("bottom_left and top_right must be of type SkyCoord or BaseCoordinateFrame")
+
+    elif bottom_left.shape == (2,):
+        top_right = bottom_left[1]
+        bottom_left = bottom_left[0]
+
+    elif top_right is not None:
+        top_right = top_right.transform_to(bottom_left.frame)        
+        width = top_right.spherical.lon - bottom_left.spherical.lon
```
ayshih: This pure subtraction doesn't work as you'd like across a longitudinal wrap.  For example, consider if a user wants to specify a 2-degree-wide-in-longitude rectangle that goes from 359 degrees to 1 degrees when wrapping at 360 degrees, or +179 degrees to -179 degrees when wrapping at 180 degrees.

Since the intent appears to be to error on rectangles than span more than 180 degrees, one way to solve these situations is to cast the difference as a `Longitude` with a wrap angle of 180 degrees.

sunpy/coordinates/tests/test_utils.py
```diff
@@ -2,14 +2,14 @@
 
 import numpy as np
 
-from astropy.coordinates import SkyCoord
+from astropy.coordinates import SkyCoord, BaseCoordinateFrame, Galactic
```
ayshih:
```suggestion
from astropy.coordinates import SkyCoord
```
You don't use `BaseCoordinateFrame` at all in this file.

sunpy/coordinates/tests/test_utils.py
```diff
@@ -237,3 +237,58 @@ def test_great_arc_different_observer(aia171_test_map):
     np.testing.assert_almost_equal(end.x.value, b2h.x.value)
     np.testing.assert_almost_equal(end.y.value, b2h.y.value)
     np.testing.assert_almost_equal(end.z.value, b2h.z.value)
+
+
+@pytest.fixture
+def rectangle_args():
+    bottom_left = SkyCoord(0 * u.arcsec, 0 * u.arcsec, frame='galactic')
+    top_right = SkyCoord(10 * u.arcsec, 10 * u.arcsec, frame='galactic')
```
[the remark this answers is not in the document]
Raahul-Singh: I didn't know about this.

sunpy/coordinates/utils.py
Raahul-Singh: > This pure subtraction doesn't work as you'd like across a longitudinal wrap. For example, consider if a user wants to specify a 2-degree-wide-in-longitude rectangle that goes from 359 degrees to 1 degrees when wrapping at 360 degrees, or +179 degrees to -179 degrees when wrapping at 180 degrees.
> 
> Since the intent appears to be to error on rectangles than span more than 180 degrees, one way to solve these situations is to cast the difference as a `Longitude` with a wrap angle of 180 degrees.

In accordance with your comment above about `top_right` being "right" and "above" `bottom_left` I've removed this section all together.

Raahul-Singh:
I have a doubt :
`>>> top_right = SkyCoord(10 * u.arcsec, 10 * u.arcsec, frame='heliographic_stonyhurst',obstime="2017-07-26")`

`>>> top_right`

`>>> <SkyCoord (HeliographicStonyhurst: obstime=2017-07-26T00:00:00.000): (lon, lat, radius) in (deg, deg, km)
    (0.00277778, 0.00277778, 695700.)>`  

Then, 

`>>> top_right.is_transformable_to("heliographic_carrington")`

gives :

`>>> False`

but the following works :

`>>>top_right.transform_to("heliographic_carrington")`

`>>><SkyCoord (HeliographicCarrington: obstime=2017-07-26T00:00:00.000): (lon, lat, radius) in (deg, deg, km)
  (283.95552018, 0.00277778, 695700.)>`

The docs for `.transform_to` says :
Determines if this coordinate frame can be transformed to another given frame.
It being `False` implies that such a conversion should not be possible, right?

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,79 @@ def coordinates(self, points=None):
...
+def get_rectangle_coordinates(bottom_left, *, top_right = None, width: u.deg = None, height: u.deg = None):
+    """
+    Specify a rectangular region of interest in spherical world coordinates.
+
+    Parameters
+    ----------
+    bottom_left : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The bottom_left coordinate of the rectangle. If a `SkyCoord` it can
+        have shape ``(2,)`` and also define ``top_right``.
+
```
nabobalis:
```suggestion
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,79 @@ def coordinates(self, points=None):
...
+    top_right : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The top_right coordinate of the rectangle. Can only be omitted if
+        ``bottom_left`` has shape ``(2,)``.
+        If in a different frame than ``bottom_left`` and all required metadata for frame conversion is present,
+        ``top_right`` will be transformed to ``bottom_left`` frame.
+        Otherwise a ``ValueError`` will be incurred.
+
```
nabobalis:
```suggestion
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,79 @@ def coordinates(self, points=None):
...
+    width : `~astropy.units.Quantity`
+        The width of the rectangle.
+        Must be omitted if ``bottom_left`` has shape ``(2,)`` or ``top_right`` is provided.
+
```
nabobalis:
```suggestion
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,79 @@ def coordinates(self, points=None):
...
+    height : `~astropy.units.Quantity`
+        The height of the rectangle.
+        Must be omitted if ``bottom_left`` has shape ``(2,)`` or ``top_right`` is provided.
+
+    Returns
+    -------
+    bottom_left : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
```
nabobalis:
```suggestion
    `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,79 @@ def coordinates(self, points=None):
...
+        The bottom_left coordinate of the rectangular region of interest in spherical world coordinates.
```
nabobalis:
```suggestion
        The bottom left coordinate of the rectangular region of interest in spherical world coordinates.
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,79 @@ def coordinates(self, points=None):
...
+
```
nabobalis:
```suggestion
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,79 @@ def coordinates(self, points=None):
...
+    top_right : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
```
nabobalis:
```suggestion
    `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`    
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,79 @@ def coordinates(self, points=None):
...
+        The height of the rectangle.
+        Must be omitted if ``bottom_left`` has shape ``(2,)`` or ``top_right`` is provided.
+
+    Returns
+    -------
+    bottom_left : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The bottom_left coordinate of the rectangular region of interest in spherical world coordinates.
+
+    top_right : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`    
+        The top_right coordinate of the rectangular region of interest in spherical world coordinates.
```
nabobalis:
```suggestion
        The top right coordinate of the rectangular region of interest in spherical world coordinates.
```

ayshih:
> `>>> top_right.is_transformable_to("heliographic_carrington")`
> 
> gives :
> 
> `>>> False`

This is because `.is_transformable_to()` takes as input a frame class, not a string, to work as intended (e.g., `HeliographicCarrington` instead of `"heliographic_carrington"`).  Incidentally, while `SkyCoord.transform_to()` does allow for string input, be aware that the `.transform_to()` method of frame classes does not.

Raahul-Singh:
@ayshih is anything else required here?

Cadair:
@Raahul-Singh Can you fix the code style issues: https://dev.azure.com/sunpy/sunpy/_build/results?buildId=5448&view=logs&j=4cdc857c-0d48-50f8-2772-3b048c8bb2f3&t=fe62049c-f9e9-5b31-8bb6-a78aae0c6ebe&l=33

You can auto-fix some with:
```
pip install pre-commit
pre-commit run --all-files
```

Raahul-Singh:
Looks like I still don't know all the cases of squashing after all. :sweat_smile:

Raahul-Singh:
The tests are failing because of a `JSONDecodeError` which is unrelated to this PR.
I am facing trouble with `JSONDecodeError` elsewhere too. Is it because of the servers?

nabobalis:
If you merge in master, the CI has updated to pass again.

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,75 @@ def coordinates(self, points=None):
...
+        Must be omitted if ``bottom_left`` has shape ``(2,)`` or ``top_right`` is provided.
+
+    Returns
+    -------
+    `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The bottom left coordinate of the rectangular region of interest in spherical world coordinates.
+    `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The top right coordinate of the rectangular region of interest in spherical world coordinates.
+    """
+    try:
```
Cadair: This try block seems wasteful. It's two whole coordinate transforms for a start. @ayshih is this really worth it over assuming type?

ayshih: Er, not exactly what I had in mind...  These should all be `hasattr()` checks, not actual transformations.

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,74 @@ def coordinates(self, points=None):
...
+            hasattr(bottom_left, 'shape') and
+            hasattr(bottom_left, 'spherical')):
+        raise TypeError("Invalid input, bottom_left must be of type SkyCoord or BaseCoordinateFrame")
+
+    if (top_right is not None and (hasattr(top_right, 'transform_to') and
+                                   hasattr(top_right, 'shape') and
+                                   hasattr(top_right, 'spherical'))):
+        raise TypeError("Invalid input, top_right if passed, must be of type SkyCoord or BaseCoordinateFrame")
+
+    if bottom_left.shape == (2,) and any((width, height, top_right)):
```
Cadair: I think width and height can be array-like, (a Quantity), which means doing a truthy operation on it would potentially raise an error.

I think it would be safer to do:
```suggestion
    if bottom_left.shape == (2,) and any((x is not None for x in (width, height, top_right))):
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,74 @@ def coordinates(self, points=None):
...
+        raise ValueError("Invalid input, if bottom_left.shape == (2,) "
+                         "other parameters should not be passed.")
+
+    if all((width, height, top_right)):
```
Cadair: Same here.
```suggestion
    if all(x is not None for x in (width, height, top_right):
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,74 @@ def coordinates(self, points=None):
...
+    if (top_right is not None and (hasattr(top_right, 'transform_to') and
+                                   hasattr(top_right, 'shape') and
+                                   hasattr(top_right, 'spherical'))):
+        raise TypeError("Invalid input, top_right if passed, must be of type SkyCoord or BaseCoordinateFrame")
+
+    if bottom_left.shape == (2,) and any((x is not None for x in (width, height, top_right))):
+        raise ValueError("Invalid input, if bottom_left.shape == (2,) "
+                         "other parameters should not be passed.")
+
+    if all(x is not None for x in (width, height, top_right):
```
ayshih:
```suggestion
    if all(x is not None for x in (width, height, top_right)):
```

changelog/3737.feature.rst
```diff
@@ -0,0 +1 @@
+Adds Rectangle coordinate parser in `~sunpy.util` and associated tests.
```
ayshih:
```suggestion
Added a helper function (`~sunpy.coordinates.util.get_rectangle_coordinates`) for defining a rectangle in longitude and latitude coordinates.
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,74 @@ def coordinates(self, points=None):
...
+
+def get_rectangle_coordinates(bottom_left, *, top_right = None, width: u.deg = None, height: u.deg = None):
+    """
+    Specify a rectangular region of interest in spherical world coordinates.
+
+    Parameters
+    ----------
+    bottom_left : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The bottom_left coordinate of the rectangle. If a `SkyCoord` it can
+        have shape ``(2,)`` and also define ``top_right``.
```
ayshih: Why "If a SkyCoord..."?  Does the function not work if `bottom_left` is a frame (not a SkyCoord) and has shape `(2,)`?

ayshih: Also, this sentence doesn't make sense.  Having `shape (2,)` and defining `top_right` are two exclusionary circumstances, so shouldn't be joined by "and also".

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,74 @@ def coordinates(self, points=None):
...
+
+
+def get_rectangle_coordinates(bottom_left, *, top_right = None, width: u.deg = None, height: u.deg = None):
+    """
+    Specify a rectangular region of interest in spherical world coordinates.
+
+    Parameters
+    ----------
+    bottom_left : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The bottom_left coordinate of the rectangle. If a `SkyCoord` it can
```
ayshih:
```suggestion
        The bottom-left coordinate of the rectangle. If a `SkyCoord` it can
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,74 @@ def coordinates(self, points=None):
...
+        have shape ``(2,)`` and also define ``top_right``.
+    top_right : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The top_right coordinate of the rectangle. Can only be omitted if
```
ayshih:
```suggestion
        The top-right coordinate of the rectangle. Can only be omitted if
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,74 @@ def coordinates(self, points=None):
...
+        ``bottom_left`` has shape ``(2,)``.
+        If in a different frame than ``bottom_left`` and all required metadata for frame conversion is present,
+        ``top_right`` will be transformed to ``bottom_left`` frame.
+        Otherwise a ``ValueError`` will be incurred.
```
ayshih:
```suggestion
        Otherwise a ``ValueError`` will be raised.
```
I don't think "incurred" is typical terminology.  This sentence could also be removed altogether.

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,74 @@ def coordinates(self, points=None):
                         obstime=self.obstime,
                         observer=self.observer,
                         frame=Heliocentric).transform_to(self.start_frame)
+
+
+def get_rectangle_coordinates(bottom_left, *, top_right = None, width: u.deg = None, height: u.deg = None):
```
ayshih: What's the thinking behind the bare asterisk here?  Why not allow `top_right` to be a positional argument?

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,74 @@ def coordinates(self, points=None):
...
+    """
+    Specify a rectangular region of interest in spherical world coordinates.
```
ayshih:
```suggestion
    Specify a rectangular region of interest in longitude and latitude in a given coordinate frame.
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,74 @@ def coordinates(self, points=None):
...
+        The width of the rectangle.
+        Must be omitted if ``bottom_left`` has shape ``(2,)`` or ``top_right`` is provided.
+    height : `~astropy.units.Quantity`
+        The height of the rectangle.
+        Must be omitted if ``bottom_left`` has shape ``(2,)`` or ``top_right`` is provided.
+
+    Returns
+    -------
+    `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The bottom left coordinate of the rectangular region of interest in spherical world coordinates.
```
ayshih:
```suggestion
        The bottom left coordinate of the rectangular region of interest.
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,74 @@ def coordinates(self, points=None):
...
+    `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The top right coordinate of the rectangular region of interest in spherical world coordinates.
```
ayshih:
```suggestion
        The top right coordinate of the rectangular region of interest.
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,74 @@ def coordinates(self, points=None):
...
+    """
+    if not (hasattr(bottom_left, 'transform_to') and
+            hasattr(bottom_left, 'shape') and
+            hasattr(bottom_left, 'spherical')):
+        raise TypeError("Invalid input, bottom_left must be of type SkyCoord or BaseCoordinateFrame")
+
+    if (top_right is not None and (hasattr(top_right, 'transform_to') and
+                                   hasattr(top_right, 'shape') and
+                                   hasattr(top_right, 'spherical'))):
+        raise TypeError("Invalid input, top_right if passed, must be of type SkyCoord or BaseCoordinateFrame")
```
ayshih:
```suggestion
        raise TypeError("Invalid input, top_right must be of type SkyCoord or BaseCoordinateFrame")
```
The user has explicitly passed in `top_right` to see this error – and if the bare asterisk in the function argument stands, the user had to even explicitly name the argument – so it's unnecessary to say "if passed".

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,74 @@ def coordinates(self, points=None):
...
+        bottom_left = bottom_left[0]
+
+    elif top_right is not None:
+        top_right = top_right.transform_to(bottom_left)
+
+    else:
+        if isinstance(bottom_left, BaseCoordinateFrame):
+            base_frame = type(bottom_left)
+            top_right = base_frame(bottom_left.spherical.lon + width,
+                                   bottom_left.spherical.lat + height)
```
ayshih: This needs to be a `bottom_left.realize_frame(...)` call because otherwise you're losing all frame attributes.

sunpy/coordinates/utils.py
Cadair: My thinking was that allowing a second positional argument is ambiguous between the two inputs. See my long comment here: https://github.com/sunpy/sunpy/pull/3677#issuecomment-575908202 for more thoughts on why this function signature.

sunpy/coordinates/utils.py
Cadair: whoops my bad (I put this line in a suggestion).

sunpy/coordinates/utils.py
Cadair: yeah we don't need the "if a SkyCoord" specifier here, this argument can only ever be a coordinate. I think the also is to indicate that this argument can also specify the top right coordinate of the rectangle (not the argument to the function), so clearly needs rewording.

sunpy/coordinates/utils.py
ayshih: It's not that ambiguous here because there are different numbers of arguments, but sure.

The docstring should show example call strings, which I think would help greatly in preparing users..

Cadair:
@Raahul-Singh the tests are failing :smile:

Raahul-Singh:
> @Raahul-Singh the tests are failing smile

One of the tests was failing because using `realize_frame` converted a SkyCoord to its frame. I've removed the test because I remember you telling me that for this particular function, the two can be used interchangably.

Raahul-Singh:
I'll add the examples in the docstring once the implementation is approved.

sunpy/coordinates/utils.py
Cadair: Does it matter here that the `bottom_left` coord could be a SkyCoord with some stored attrrs and we end up creating a `top_right` which is (currently a frame) but even if we cast to `SkyCoord` it wouldn't have the same history as the `bottom_left`.

ayshih: What was previously there was fine structurally: two code paths depending on whether `bottom_left` was a frame or a SkyCoord.  The frame path needed to call `realize_frame()`, while the SkyCoord path could just use the SkyCoord constructor to indirectly call `realize_frame()` for you.

If the preference is to have a largely single path, you should use the SkyCoord constructor to create a `top_right` as a SkyCoord (which would look the same regardless of the type of `bottom_left`), and then "downgrade" it to a frame if `bottom_left` was actually just a frame.  So, no explicit call to `realize_frame()`.

sunpy/coordinates/utils.py
Cadair: yeah, this should be conditional on the input type.
I think you can do
```
top_right = bottom_left.realize_frame(...)
if isinstance(bottom_left, SkyCoord):
    top_right = SkyCoord(top_right.data, frame=bottom_left)
```
not sure if that's the best way to do it though, @ayshih ?

Raahul-Singh: What I'm doing right now is basically this but in opposite order.
top_right becomes a SkyCoord first  and if bottom_left is a frame, I downgrade it to the corresponding frame.
Would the opposite order be more efficient?

ayshih: It's fine either way.  The SkyCoord constructor is a little additional overhead, but I'd argue that it's more readable.  Adding a comment to the downgrading would be prudent.

(If you change it to @Cadair's way, use `SkyCoord(top_right)`.)

sunpy/coordinates/utils.py
Cadair: I agree that a comment about the type change would be useful. I am :+1: on the implementation as-is now.

sunpy/coordinates/tests/test_utils.py
```diff
@@ -237,3 +237,89 @@ def test_great_arc_different_observer(aia171_test_map):
...
+
+
+def test_rectangle_bottom_left_different_types(rectangle_args):
+    bottom_left, top_right, width, height = rectangle_args
+
+    bottom_left_1, top_right_1 = get_rectangle_coordinates(bottom_left.frame, width=width, height=height)
+
+    assert bottom_left.spherical.lon + width == top_right_1.spherical.lon
+    assert bottom_left.spherical.lat + height == top_right_1.spherical.lat
+    assert type(bottom_left_1) == type(top_right_1)
```
Cadair: You should also be asserting that it's the same type as the input:
```suggestion
    assert type(bottom_left_1) == type(top_right_1) == type(bottom_left.frame)
```

sunpy/coordinates/tests/test_utils.py
```diff
@@ -237,3 +237,89 @@ def test_great_arc_different_observer(aia171_test_map):
...
+
+    bottom_left_2, top_right_2 = get_rectangle_coordinates(bottom_left, width=width, height=height)
+
+    assert bottom_left.spherical.lon + width == top_right_2.spherical.lon
+    assert bottom_left.spherical.lat + height == top_right_2.spherical.lat
+    assert type(bottom_left_2) == type(top_right_2)
```
Cadair: same here
```suggestion
    assert type(bottom_left_2) == type(top_right_2) == type(bottom_left)
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,72 @@ def coordinates(self, points=None):
...
+
+def get_rectangle_coordinates(bottom_left, *, top_right = None, width: u.deg = None, height: u.deg = None):
+    """
+    Specify a rectangular region of interest in longitude and latitude in a given coordinate frame.
+
+    Parameters
+    ----------
+    bottom_left : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The bottom-left coordinate of the rectangle.
+        Supports passing both ``bottom_left`` and ``top_right`` as an array of shape ``(2,)``.
```
Cadair:
```suggestion
        Supports passing both the bottom left and top right coordinates by passing with a shape of ``(2,)``.
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,72 @@ def coordinates(self, points=None):
...
+    top_right : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The top-right coordinate of the rectangle. Can only be omitted if
+        ``bottom_left`` has shape ``(2,)``.
```
Cadair: You can also omit it if you pass `width` and `height`.
```suggestion
        The top-right coordinate of the rectangle.
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,72 @@ def coordinates(self, points=None):
...
+        If in a different frame than ``bottom_left`` and all required metadata for frame conversion is present,
+        ``top_right`` will be transformed to ``bottom_left`` frame.
+        Otherwise a ``ValueError`` will be raised.
```
Cadair:
```suggestion
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,72 @@ def coordinates(self, points=None):
...
+    width : `~astropy.units.Quantity`
+        The width of the rectangle.
+        Must be omitted if ``bottom_left`` has shape ``(2,)`` or ``top_right`` is provided.
```
Cadair:
```suggestion
        Must be omitted if the coordinates of both corners have been specified.
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,72 @@ def coordinates(self, points=None):
...
+    height : `~astropy.units.Quantity`
+        The height of the rectangle.
+        Must be omitted if ``bottom_left`` has shape ``(2,)`` or ``top_right`` is provided.
```
Cadair:
```suggestion
        Must be omitted if the coordinates of both corners have been specified.
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,72 @@ def coordinates(self, points=None):
...
+    -------
+    `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The bottom left coordinate of the rectangular region of interest.
+    `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
+        The top right coordinate of the rectangular region of interest.
+    """
+    if not (hasattr(bottom_left, 'transform_to') and
+            hasattr(bottom_left, 'shape') and
+            hasattr(bottom_left, 'spherical')):
+        raise TypeError("Invalid input, bottom_left must be of type SkyCoord or BaseCoordinateFrame")
```
Cadair:
```suggestion
        raise TypeError("Invalid input, bottom_left must be of type SkyCoord or BaseCoordinateFrame.")
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,72 @@ def coordinates(self, points=None):
...
+
+    if (top_right is not None and not ((hasattr(top_right, 'transform_to') and
+                                        hasattr(top_right, 'shape') and
+                                        hasattr(top_right, 'spherical')))):
+        raise TypeError("Invalid input, top_right must be of type SkyCoord or BaseCoordinateFrame")
```
Cadair:
```suggestion
        raise TypeError("Invalid input, top_right must be of type SkyCoord or BaseCoordinateFrame.")
```

sunpy/coordinates/utils.py
```diff
@@ -265,3 +265,72 @@ def coordinates(self, points=None):
...
+
+    if bottom_left.shape == (2,):
+        top_right = bottom_left[1]
+        bottom_left = bottom_left[0]
+
+    elif top_right is not None:
+        top_right = top_right.transform_to(bottom_left)
+
+    else:
+        top_right = SkyCoord(bottom_left.spherical.lon + width,
```
Cadair: Add a comment here about type casting, as discussed in a couple of threads.

(new comment to track it)

sunpy/coordinates/tests/test_utils.py
```diff
@@ -237,3 +237,89 @@ def test_great_arc_different_observer(aia171_test_map):
...
+
+
+def test_rectangle_bottom_left_vector():
+    bottom_left_vector = SkyCoord([0 * u.arcsec, 10 * u.arcsec], [0 * u.arcsec, 10 * u.arcsec], frame='heliographic_stonyhurst')
```
Cadair:
```suggestion
    bottom_left_vector = SkyCoord([0 * u.arcsec, 10 * u.arcsec], [0 * u.arcsec, 10 * u.arcsec],
                                  frame='heliographic_stonyhurst')
```

Cadair:
Online test fails are HELIO being offline.

Raahul-Singh:
Thanks! @Cadair and @ayshih! :)

PR #3738, opened 2020-01-29
feat(netcdf): Support SUVI netCDF files

zflamig:
### Description

Supports SUVI netcdf files from GOES-16 & GOES-17

Example of using this, first install libraries:
`conda install netcdf4 hdf5`

then:

```
import astropy.units as u
import sunpy.map

smap = sunpy.map.Map("http://noaa-goes16.s3.amazonaws.com/SUVI-L1b-Fe171/2020/001/03/OR_SUVI-L1b-Fe171_G16_s20200010302242_e20200010302252_c20200010302513.nc")

ax = plt.subplot(projection=smap)
smap.plot(clip_interval=(0.05, 99.95)*u.percent)
plt.colorbar(extend='both')
plt.savefig('test_sunpy.png', dpi=200)
```

### DIFF
--- a/sunpy/io/file_tools.py
+++ b/sunpy/io/file_tools.py
@@ -20,13 +20,19 @@ try:
 except ImportError:
     ana = None
 
+try:
+    from . import netcdf
+except ImportError:
+    netcdf = None
+
 __all__ = ['read_file', 'read_file_header', 'write_file']
 
 # File formats supported by SunPy
 _known_extensions = {
     ('fts', 'fits'): 'fits',
     ('jp2', 'j2k', 'jpc', 'jpt'): 'jp2',
-    ('fz', 'f0'): 'ana'
+    ('fz', 'f0'): 'ana',
+    ('nc'): 'netcdf'
 }
 
 
@@ -48,7 +54,8 @@ class Readers(dict):
 _readers = Readers({
             'fits': fits,
             'jp2': jp2,
-            'ana': ana
+            'ana': ana,
+            'netcdf': netcdf
 })
 
 
--- a/sunpy/io/netcdf.py
+++ b/sunpy/io/netcdf.py
@@ -1,0 +1,78 @@
+"""
+This module provides a netCDF file reader.
+"""
+import collections
+import numpy.ma as ma
+
+from sunpy.io.header import FileHeader
+
+__all__ = ['read', 'get_header', 'write']
+
+HDPair = collections.namedtuple('HDPair', ['data', 'header'])
+
+
+def read(filepath, **kwargs):
+    """
+    Reads a netCDF file.
+
+    Parameters
+    ----------
+    filepath : `str`
+        The file to be read.
+
+    Returns
+    -------
+    pairs : `list`
+        A list of (data, header) tuples.
+    """
+    # Put import here to speed up sunpy.io import time
+    from netCDF4 import Dataset
+    header = get_header(filepath)
+
+    nc = Dataset(filepath, 'r')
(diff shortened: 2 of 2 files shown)


zflamig:
test_sunpy
Example output

Cadair:
Thanks a lot for this PR @zflamig 

The big comment I have here is that the `sunpy.io` submodule is supposed to be independent of the data it is reading. In your implementation here, the only netCDF files it could read are the GOES SUVI ones.

I don't know enough about the way that netCDF files are laid out to know if it's practical / possible to read all the datasets in a file into a list of array & metadata pairs? The metadata doesn't have to conform to FITS as far as `sunpy.io` is concerned, but will need to follow at least the coordinate metadata standards for map to work (although we could add support to map for metadata translation if we needed to).

I hope that makes some sense!

zflamig:
I understand. This is actually really tricky with netCDF because it is a general purpose container format. So the data & metadata are intermixed.

It would be possible to add an if check based on other values in the file for setting `instrume`, however I am not sure how to figure out if the data field is not `RAD`.

Do we know of any other datasets available in netCDF to compare against?

nabobalis:
Yeah this might make more sense as an instr goes method but that brings up he question as the future of the instr subpackage. 

But I am in favour of adding support for these netcdf files in sunpy.

wtbarnes:
Given the discussion on the telecon yesterday, I think what makes the most sense is to move `io/netcdf.py` into `instr/suvi.py` as this functionality is instrument specific. Then, the `read` function could be renamed something like `from_netcdf` and would return a `SUVIMap` instead of a data-header pair. We may want to handle the optional dependency on `netcdf4` a bit more gracefully, e.g. by adding as an optional dependency in `setup.cfg`

wtbarnes:
A follow up question that also came up on the telecon yesterday: what is the canonical source of the SUVI data? Is it these netCDF files on AWS? Is it the FITS files provided [here](https://data.ngdc.noaa.gov/platforms/solar-space-observing-satellites/goes/)? (this is where the `SUVIClient` in sunpy retrieves them). Or somewhere else? 

ping @wafels, @ehsteve, and @hayesla as they've previously worked on supporting SUVI data in sunpy.

zflamig:
Little confused by the feedback here. This will likely work for more than just SUVI, however without seeing other data in netcdf format it is hard to know if we have the abstractions right. 

NOAA is providing the SUVI data in both NetCDF and FITS. I believe the NetCDF is converted to FITS for distribution by them so both are canonical.

nabobalis:
> Little confused by the feedback here. This will likely work for more than just SUVI, however without seeing other data in netcdf format it is hard to know if we have the abstractions right.

This  is the main issue that made us decide on the path to move it out of `io` and into `instr`. If this is not "general" enough to open other NETCDF solar data, it becomes instrument specific. Which means there is a bit more work to be done instead of this approach unfortunately.

ehsteve:
If these data are exactly the same as those provided in the fits files then I don’t see why we would also support the netcdf unless there is something inherently superior to this file format. Thought?

ehsteve:
I’m sorry. I take that back. We should support the files but not necessarily support two ways of getting suvi data if they are the same.

nabobalis:
Do we know or can we work out a list of all known sources of solar data that use NETCDF?

dstansby:
Thanks for opening this PR! We had a chat about this on the weekly sunpy call, and decided that the approach is not generic enough to work. In particular, `sunpy.map.Map` currently does not support custom loaders based on data source. Because `sunpy.io` tools should be generic, and not instrument specific, a lot of modification needs to happen to the `sunpy.map` to add support for data-source dependent file loading. As such I'm going to close this, as I think it needs building from the ground up.

PR #3743, opened 2020-01-31
Fixed broken link in documentation for Timeseries

### LINKED ISSUE #3682: List of available sources for TimeSeries?
The [docs](https://docs.sunpy.org/en/stable/code_ref/timeseries.html#instrument-timeseries-classes) state:

> A number of instruments are supported through subclasses of the base GenericTimeSeries class. To see Instrument TimeSeries Classes for a list of all of them.

But there is nothing there, except for the XRS example. Is there a list of supported sources? If so, could it be linked in this page?

pkj-m:
This PR tries to fix the broken link for timeseries documentation (for v1.0+). 
Fixes #3682 

@nabobalis


nabobalis:
I am not a fan of adding it back to the init. I think it should be handled via the rst file instead.

pkj-m:
Sure. I'll look into it immediately.

pkj-m:
@nabobalis I looked around and it seems in order to print the all the GenericTimeseries classes in the docs, they either need to imported explicitly (like done in this PR) or at least exported using the `__all__` in the timeseries' `__init__.py` file. 

Alternatively, we can add another automod line in the rst, such as `.. automodapi:: sunpy.timeseries.sources` below `.. automodapi:: sunpy.timeseries` but that creates 2 disjoint tables and inheritance diagrams. Any way to combine both of them?

nabobalis:
How does it look?

pkj-m:
Screenshot from 2020-02-05 02-26-39

We can also hide the inheritance diagram with just the metadata and generic timeseries labels if that makes it prettier.

nabobalis:
That looks ok to me.

nabobalis:
But yeah we can hide the first graph.

pkj-m:
@nabobalis I think I messed up something with git while adding a remote to my forked repo and updating it with the master. Is it okay if I close this PR and create a new one?

nabobalis:
Ok ahead.

pkj-m:
Closing this PR and opening another one at #3757

PR #3746, opened 2020-02-03
create .data property so as to encourage cleaner API

### LINKED ISSUE #2917: Timeseries should encourage users to not use .data to get a dataframe
Given that we generally want the Timeseries API to not be tied to any particular timeseries data structure, i.e pandas or Astropy Timeseries. Users should be using `to_dataframe` rather than `.data`.

sashank27:
### Description

Fixes #2917

### DIFF
--- a/sunpy/timeseries/timeseriesbase.py
+++ b/sunpy/timeseries/timeseriesbase.py
@@ -121,7 +121,7 @@ class GenericTimeSeries:
 
     @property
     def data(self):
-        warnings.warn("Using .data to access the dataframe is discouraged;"
+        warnings.warn("Using .data to access the dataframe is discouraged; "
                       "use .to_dataframe() instead.", SunpyUserWarning)
         return self._data
 
--- a/sunpy/timeseries/timeseriesbase.py
+++ b/sunpy/timeseries/timeseriesbase.py
@@ -91,18 +91,18 @@ class GenericTimeSeries:
 
     # kwargs are not used here but are passed in for sources.
     def __init__(self, data, meta=None, units=None, **kwargs):
-        self.data = data
+        self._data = data
         tr = self.time_range
         # Check metadata input
         if meta is None:
             # No meta given, so default
-            self.meta = TimeSeriesMetaData(MetaDict(), tr, list(self.data.columns.values))
+            self.meta = TimeSeriesMetaData(MetaDict(), tr, list(self._data.columns.values))
         elif isinstance(meta, (dict, OrderedDict, MetaDict)):
             # Given the values for metadata (dict) and infer timerange and colnames from the data
-            self.meta = TimeSeriesMetaData(meta, tr, list(self.data.columns.values))
+            self.meta = TimeSeriesMetaData(meta, tr, list(self._data.columns.values))
         elif isinstance(meta, tuple):
             # Given the values all in a tuple
-            self.meta = TimeSeriesMetaData(meta, tr, list(self.data.columns.values))
+            self.meta = TimeSeriesMetaData(meta, tr, list(self._data.columns.values))
         else:
             # Should have a list of 3-tuples giving a complex metadata list.
             self.meta = meta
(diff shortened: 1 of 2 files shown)


sunpy/timeseries/timeseriesbase.py
```diff
@@ -119,6 +119,15 @@ def __init__(self, data, meta=None, units=None, **kwargs):
 
 # #### Attribute definitions #### #
 
+    @property
+    def data(self):
+        warnings.warn("use `.to_dataframe` instead of `.data` to get the data of the timeseries", SunpyUserWarning)
```
Cadair:
```suggestion
        warnings.warn("use `.to_dataframe` instead of `.data` to get the timeseries as a pandas dataframe.", SunpyUserWarning)
```

Cadair:
Looks pretty good to me, I will let one of the timeseries maintainers review it :smile:

sunpy/timeseries/timeseriesbase.py
```diff
@@ -119,6 +119,15 @@ def __init__(self, data, meta=None, units=None, **kwargs):
...
+        return self._data
+
+    @data.setter
+    def data(self,d):
```
nabobalis:
```suggestion
    def data(self, d):
```

nabobalis: Stupid question, why do we need this setter?

sunpy/timeseries/timeseriesbase.py
```diff
@@ -119,6 +119,15 @@ def __init__(self, data, meta=None, units=None, **kwargs):
 
 # #### Attribute definitions #### #
 
+    @property
+    def data(self):
+        warnings.warn("use `.to_dataframe` instead of `.data` to get the timeseries as a pandas dataframe.", SunpyUserWarning)
```
nabobalis:
```suggestion
        warnings.warn("use `.to_dataframe` instead of `.data` to get the timeseries as a pandas dataframe.",
                                SunpyUserWarning)
```

dstansby:
Is the plan to deprecate using `.data` in a future release? I think it should be, which means the warning should be tweaked slightly to say it's deprecated.

Cadair:
My plan was, much like `map`, to keep it around as a reference to the underlying data source. However, like in `map` there is no restriction on the type returned by `.data` i.e. it's perfectly possible (and encouraged) to have `.data` be a dask array rather than a numpy array.

dstansby:
Then if `.data.` is a valid operation, it shouldn't emit a deprecation warning, but the docstring should be changed instead, and users notified by the release notes?

Cadair:
It's not currently emitting a deprecation warning, it's just emitting a user warning. (I do agree that the message should be changed to make it clear that it's only the type that's undefined.)

My concern with just adding this to the docs is that it's much more likely that people wont notice it or ignore it like that. If we do ever end up transitioning timeseries to use `astropy.table` underneath rather than pandas those still using `.data` and expecting a dataframe are in for a shock.

I am happy to go with whatever you think the best option here is though. (Although making `.data` a property does feel like a good idea).

dstansby:
I would vouch for deprecating it, and the if we swap to using `astropy.table` (or similar) under the hood we can add a `to_table()` (or similar) method (or mayble a `.table` property).

Cadair:
>add a to_table() (or similar) method 

think we already have this.

My main reservation about deprecating it is that we have it on all the other objects (NDData, map, ndcube, other things inheriting from ndcube) etc.

sunpy/timeseries/timeseriesbase.py
```diff
@@ -119,6 +119,16 @@ def __init__(self, data, meta=None, units=None, **kwargs):
 
 # #### Attribute definitions #### #
 
+    @property
+    def data(self):
+        warnings.warn("use `.to_dataframe` instead of `.data` to get the timeseries as a pandas dataframe.",
```
dstansby: I think something more like 'Using `.data` to access the dataframe is discouraged; use `.to_dataframe()` instead'

sunpy/timeseries/timeseriesbase.py
```diff
@@ -119,6 +119,16 @@ def __init__(self, data, meta=None, units=None, **kwargs):
 
 # #### Attribute definitions #### #
 
+    @property
+    def data(self):
+        warnings.warn("Using .data to access the dataframe is discouraged;"
```
dstansby:
```suggestion
        warnings.warn("Using .data to access the dataframe is discouraged; "
```
Sorry to bother, but apart from this space all looks 👍 👍 👍

nabobalis:
Thanks  @sashank27

PR #3760, opened 2020-02-04
skip another example

nabobalis:
Should restore the doc build and has already committed to the other branches. 

Why this errors I can not tell you.


azure-pipelines.yml
```diff
@@ -42,6 +42,7 @@ jobs:
     envs:
       - macos: py36
         name: py36_test
+        posargs: -n=1
```
dstansby: Why this change?

examples/example_template/example_template.py
```diff
@@ -87,3 +80,11 @@ def dummy():
 # code directly will see the plots; this is not necessary for creating the docs
 
 plt.show()
+
```
dstansby: And why has this block moved?

azure-pipelines.yml
nabobalis: Race condition on the macos builds only.

examples/example_template/example_template.py
nabobalis: To make sure it does not interrupt the flow of an example.

nabobalis:
> Should probably also open a corresponding issue so we don't forget to fix the example.

I will open one.

PR #3762, opened 2020-02-05
Fix mapbase test warnings

dstansby:
Taken out of #3648

### DIFF
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -95,9 +95,7 @@ def generic_map():
 
 def test_fits_data_comparison(aia171_test_map):
     """Make sure the data is the same in pyfits and SunPy"""
-    with warnings.catch_warnings():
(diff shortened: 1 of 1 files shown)


sunpy/map/tests/test_mapbase.py
```diff
@@ -82,7 +95,10 @@ def generic_map():
 
 def test_fits_data_comparison(aia171_test_map):
     """Make sure the data is the same in pyfits and SunPy"""
-    data = fits.open(os.path.join(testpath, 'aia_171_level1.fits'))[0].data
+    with warnings.catch_warnings():
```
Cadair: out of interest, why the `warnings.catch_warnings` here and not `pytest.warns`?

nabobalis: We don't care about the warning?

dstansby: Um I'm not sure, let me get back to you on that one...

dstansby: Fixed

Issue #1615, opened 2015-11-20
Alignment of Halpha & HMI composite maps


slyardley:
I am trying to create a composite map consisting of submaps of H-alpha and HMI data. The H-alpha data seems to be offset in both the x and y direction by ~100", compared with HMI.

Halpha fits: 
ftp://ftp.kso.ac.at/Ha4M/2011/20110605/processed/kanz_halph_fi_20110605_065522.fts
(password for Halpha available here: http://cesar.kso.ac.at/main/ftp.php)

HMI fits: 
http://jsoc.stanford.edu/SUM72/D761559934/S00000/
(hmi.M_720s.20110605_070000_TAI.1.magnetogram.fits)

``` python
hmap = sunpy.map.Map('kanz_halph_fi_20110605_065522.fts').rotate()
hmap.meta['waveunit'] = 'AA' #changing ANG to proper units
hmap.meta['CTYPE1'] = 'HPLN-TAN'
hmap.meta['CTYPE2'] = 'HPLT-TAN'
smap = hmap.submap(u.Quantity([-100, 500], u.arcsec),
                   u.Quantity([-600, -250], u.arcsec))
hmimap = sunpy.map.Map('hmi.M_720s.20110605_070000_TAI.1.magnetogram.fits').rotate()
shmimap = hmimap.submap(u.Quantity([-100, 500], u.arcsec),
                   u.Quantity([-600, -250], u.arcsec))

comp_map = sunpy.map.Map(smap,shmimap, composite=True)
comp_map.set_alpha(1, 0.75)
comp_map.set_colors(1,'winter')
comp_map.set_levels(1, [-300,-500,300,500])
fig = plt.figure(figsize=(12,8))
comp_map.plot()
plt.title('2011-06-05 06:55:22')

```

![comp_plot](https://cloud.githubusercontent.com/assets/5773943/11302049/e31dd534-8f92-11e5-819f-30ac8e586830.png)

Cadair:
The alignment between the two images is only as good as the respective pointing information in the HMI and H alpha headers. I would assume the H aplha header is not accurate?

slyardley:
I couldn't see anything in the headers (unless I missed it). Also I have made the same plot before in IDL with no alignment issues?

Cadair:
@slyardley interesting. I will try and take a look over the weekend, but I have limited ideas and all the ones I do have are scary. Have you used these two exact files in IDL?

slyardley:
[...]

PR #3763, opened 2020-02-05
Implemented Groundclients in continuation with #2619

abhijeetmanhas:
### Description

It is continuation of PR https://github.com/sunpy/sunpy/pull/2619, it implements kanzelhohe, gong, farside, vsm and bbso clients.


ayshih:
A question: should we render the observatory as "[Kanzelhoehe](https://en.wikipedia.org/wiki/Kanzelhoehe_Solar_Observatory)", or even more correctly as "[Kanzelhöhe](https://www.kso.ac.at/index_en.php)", rather than "Kanzelhohe"?

One reason that I ask is that Kanzelhoehe is spelled differently in another issue than it is here, and that frustrated my initial attempt at searching.

ayshih:
I'll point out that if this PR gets merged, we really ought to fix how Kanzelhohe/Kanzelhoehe/Kanzelhöhe files are read in by `sunpy.map` (see #1615).

abhijeetmanhas:
On it .
I'll have to make changes to `norh` as well, for making it pass tests.

nabobalis:
There are a lot of doc string changes to be made having glanced over it.

dstansby:
Would it be possible/easy to split this into one PR per source? That would make it a *lot* easier to review

abhijeetmanhas:
Yes, I can do it,  since all are independent, should this pr be closed now?

dstansby:
Feel free to leave this PR open if you want, and open other smaller ones that split this one up.

abhijeetmanhas:
Ok

abhijeetmanhas:
@nabobalis @dstansby  it fails on `sunpy/net/dataretriever/tests/test_goes_suvi.py`, the file I haven't even touched. So I then even checked with upstream
```
!git clone https://github.com/sunpy/sunpy.git
%cd sunpy
!pip install -e .[all,dev]
!pytest --remote-data=any sunpy/net/dataretriever/tests/test_goes_suvi.py
```
But I'm still getting the same errors, why is it so?

nabobalis:
The servers are broken.

abhijeetmanhas:
So is this PR reviewable now? Since it have passed all tests of its individual clients. OR I need to create separate PR for every client?

nabobalis:
Yeah, I say split it up into one PR per client and then one PR for what looks like a bug fix for the scrapper? 

Doing so means the commit history will be erased but I am ok with that.

abhijeetmanhas:
That was my only doubt, with commit history being erased . Thanks for
clearing it. Creating separate PRs now.

abhijeetmanhas:
@nabobalis  @dstansby All 4 splits  can be reviewed now, it have passed all the tests except suvi and changelog is also there.

Cadair:
A comment about all the clients, we should make sure that the VSO hasn't started indexing these data in the meantime. I am pretty sure you can at least get some GONG data from VSO now.

ejm4567:
The VSO currently has all the GONG data, available under the provider NSO. We also have KSO data. The BBSO data is currently offline under provider = HANET. We'll let you know when it's restored. I'll have to check on VSM and FARSIDE.

ejm4567:
The VSO has VSM data as well under provider = NSO and source = SOLIS.
See https://sdac.virtualsolar.org/cgi/show_details?instrument=vsm for details.

The Farside GONG data is currently not available thru the VSO, but is available directly from NSO's farside website.

abhijeetmanhas:
So except kanzelhohe, should PRs about bbso, gong, farside and vsm should be closed?

Cadair:
We should double check that the clients and the VSO are returning exactly the same data products, but if they are we should close them. In my opinion there is no good reason to implement code we have to maintain if the VSO are already doing it for us?

nabobalis:
> In my opinion there is no good reason to implement code we have to maintain if the VSO are already doing it for us?

This is mine too. I rather not keep duplicate code if the VSO already handles sources for us.

abhijeetmanhas:
I don't found data from ``tucson`` observatory on VSO, but data can be accessed through GONGClient implemented here. 

Also, can't find `vector_magnetic_filed` `equivalence_width` physobs data on VSO, but it is accessible in this VSMClient.

There is bbso on VSO, but don't know how to get data from it, as I applied a filter time interval or two years and it displayed something like "no records found" .

I think we have to first check whether the search queries on VSO can search the correct data (same as these clients, if implementation is correct, or verify from official observatories' website) and then is data on VSO is subset of actual data? It is in the case of `FARSIDE`.
I couldn't find matching data on VSO with same time queries, through these clients and VSO website. It was less for all clients, but this fact should be checked, **do VSO have access to all ground client data?**

dpshelio:
> We should double check that the clients and the VSO are returning exactly the same data products, but if they are we should close them. In my opinion there is no good reason to implement code we have to maintain if the VSO are already doing it for us?

If I remember right... I had worked on all these clients mostly because they were providing data in real time whereas VSO was not (needed for SolarMonitor). This may have changed.

abhijeetmanhas:
> The VSO has VSM data as well under provider = NSO and source = SOLIS.
> See https://sdac.virtualsolar.org/cgi/show_details?instrument=vsm for details.
> 
> The Farside GONG data is currently not available thru the VSO, but is available directly from NSO's farside website.

@ejm4567 I feel we can't have access to `physobs` equivalence width or vector magnetic field through VSO. But it is accessible through client. Can you recheck?

abhijeetmanhas:
> The VSO currently has all the GONG data, available under the provider NSO. We also have KSO data. The BBSO data is currently offline under provider = HANET. We'll let you know when it's restored. I'll have to check on VSM and FARSIDE.

@ejm4567 does VSO covers data from `tucson` observatory too? I couldn't find in their search form.

PR #3765, opened 2020-02-05
Fix error cause by trying to download flies which don't exist in jsoc DB

### LINKED ISSUE #3735: Some JSOC data fails to download using Fido
### Description
I'm trying to download SDO HMI magnetic data for selected time ranges. Sometimes there are MISSING entries in the search results. If you try to download those "corrupted" UnifiedResponse objects, Fido fails

### Expected behavior
Working download without any corrupted objects or ability to filter them out.

### Actual behavior
Fido.fetch() crashes with error (see in the gist below)

### Steps to Reproduce
This gist: https://gist.github.com/vit1-irk/19d9ebbc69281fd142316f52a2f019ce

Errors also happen when downloading from website

See: <https://jsoc.stanford.edu/ajax/lookdata.html>
Looks like the problem is on JSOC website, but maybe Sunpy should handle these MISSING entries itself

### System Details
Included in the gist. Btw, sorry, but this time additional debugging info (like query dumps) didn't work out properly

samaloney:
### Description

Fixes #3735 

* Add new Quality attr to jsoc
* Add default Quality('? QUALITY >=0 ?) to all jsoc queries

### DIFF
--- a/sunpy/net/jsoc/attrs.py
+++ b/sunpy/net/jsoc/attrs.py
@@ -83,13 +83,40 @@ class Notify(SimpleAttr):
 
 class Quality(SimpleAttr):
     """
-    A comparison test to run again the quality value (e.g. '>= 0').
+    A comparison test and number to run against the quality value (e.g. `>= 0`).
 
-    Quality information is encoded into a  bit mask represented as a number (i.e [1,0,0,0,1] = 2**4
-    + 2**0 = 17). The exact meaning of each bit in the bit mask differs from series to series see
+    Quality information is encoded into 32 bit flags e.g. `0b01010101010101010101010101010101`
+    which can be represented in hexadecimal as `0x55555555`,  or as in integer `1431655765`.
+    The exact meaning of each bit defers from series to series see
     `MDI <http://jsoc.stanford.edu/MDI/MDI_Global.html>`_,
-    `HMI <http://jsoc.stanford.edu/doc/data/hmi/Quality_Bits/quallev1.html>`_, and
+    `HMI <http://jsoc.stanford.edu/doc/data/hmi/Quality_Bits/quallev1.txt>`_, and
     `AIA <http://jsoc.stanford.edu/~jsoc/keywords/AIA/AIA02840_K_AIA-SDO_FITS_Keyword_Document.pdf>`_.
+
+    .. warning::
+        The quality field is returned as an **unsigned 32 bit** integer representation of the bit
+        flags but the attribute expects a **signed 32 bit**  integer representation of the bit
+        flags.
+
+    For example the bit flags `0b10000000000000000000000000000000` or `0x80000000` in hex have an
+    unsigned 32 bit integer value of `2147483648` or a signed 32 bit integer value of `-2147483648`.
+
+    For HMI the missing data flag is `0x80000000` or `0b10000000000000000000000000000000` this can
+    be converted to a signed int using::
+
+        import numpy as np
+        np.array(0x80000000).astype(np.int32)
+        array(-2147483648, dtype=int32)
+
+    This value could then be used with the quality attribute `Quality('>-2147483648)` however this
+    may not remove all missing data as other flag or bits may be set. For singed 32 bit these
+    numbers would be smaller (`0x80000000 < 0x88000000`, `-2147483648 < -2013265920`) and so the
+    comparision test would be false. In this case it is important to note all signed 32 bit numbers
+    with the top bit set (`0x80000000`) are negative so it is sufficient to query for quality values
+    the are greater than or equal to zero `Quality('>=0')`
+
+    Another example with HMI the QUAL_ECLIPSE flag (`0x00000200`) as a signed 32 bit number is `512`
+    setting any additional flags can only increase this value so to filter out any results with the
+    QUAL_ECLIPSE flag `Quality( < 512)` would be used.
     """
 
 
--- a/sunpy/net/jsoc/attrs.py
+++ b/sunpy/net/jsoc/attrs.py
@@ -83,7 +83,7 @@ class Notify(SimpleAttr):
 
 class Quality(SimpleAttr):
     """
-    A quality flag
+    A comparison to run again the quality value e.g. '>= 0'
 
     """
     pass
--- a/sunpy/net/jsoc/attrs.py
+++ b/sunpy/net/jsoc/attrs.py
@@ -81,6 +81,14 @@ class Notify(SimpleAttr):
         self.value = value
 
 
+class Quality(SimpleAttr):
+    """
+    A quality flag
+
+    """
+    pass
+
+
 walker = AttrWalker()
 
 
--- a/sunpy/net/jsoc/attrs.py
+++ b/sunpy/net/jsoc/attrs.py
@@ -5,7 +5,7 @@ from sunpy.net._attrs import Wavelength, Time
 from sunpy.net.attr import AttrWalker, AttrAnd, AttrOr, DataAttr, SimpleAttr
 
 
-__all__ = ['Series', 'Protocol', 'Notify', 'Segment', 'Keys', 'PrimeKey']
+__all__ = ['Series', 'Protocol', 'Notify', 'Segment', 'Keys', 'PrimeKey', 'Quality']
 
 
 class Series(SimpleAttr):
@@ -83,10 +83,14 @@ class Notify(SimpleAttr):
 
 class Quality(SimpleAttr):
     """
-    A comparison to run again the quality value e.g. '>= 0'
+    A comparison test to run again the quality value (e.g. '>= 0').
 
+    Quality information is encoded into a  bit mask represented as a number (i.e [1,0,0,0,1] = 2**4
+    + 2**0 = 17). The exact meaning of each bit in the bit mask differs from series to series see
+    `MDI <http://jsoc.stanford.edu/MDI/MDI_Global.html>`_,
+    `HMI <http://jsoc.stanford.edu/doc/data/hmi/Quality_Bits/quallev1.html>`_, and
+    `AIA <http://jsoc.stanford.edu/~jsoc/keywords/AIA/AIA02840_K_AIA-SDO_FITS_Keyword_Document.pdf>`_.
     """
-    pass
 
 
 walker = AttrWalker()
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -107,7 +107,8 @@ class JSOCClient(BaseClient):
 
         Not all data returned by a `~sunpy.net.jsoc.JSOCClient.search` is available for download.
         The status of data is encoded into the `Quality` key and its interpretation varies
-        by series and can be filtered using `~sunpy.net.jsoc.attrs.Quality` see Example 3 below.
+        by series and can be filtered using the `sunpy.net.jsoc.attrs.Quality` attribute
+        see Example 3 below.
 
     Notes
     -----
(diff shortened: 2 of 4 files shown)


sunpy/net/jsoc/jsoc.py
```diff
@@ -670,6 +671,11 @@ def _make_recordset(self, series, start_time='', end_time='', wavelength='',
         if sample:
             sample = f'@{sample}s'
 
+        # Extract and format quality
+        quality = kwargs.get('quality', '')
```
Cadair: I would add the default here, rather than in line 327 above.

sunpy/net/jsoc/jsoc.py
```diff
@@ -323,7 +324,7 @@ def search(self, *query, **kwargs):
         """
 
         return_results = JSOCResponse()
-        query = and_(*query)
+        query = and_(*(*query, attrs.Quality('?QUALITY >= 0?')))
```
Cadair: I think adding this here will cause errors if the user tries to specify their own Quality attr. See my comment further down about where I think the default would be best put.

Cadair:
Thanks @samaloney I will see if I can rustle up some documentation on the Quality flag we can add to the docstring for the attr to explain what's going on.

sunpy/net/jsoc/jsoc.py
Cadair: I don't really know enough about this, but should we only be asking the user to specify the bit after `QUALITY` i.e. you would do something like `a.jsoc.Quality(">= 0")`?

samaloney: Yea me neither AFAIK they are defined as bitflags and then you'd need convert to number and pass that in. I'm assuming there is a convention that if the data is missing the corresponding number is negative?

And the exact meaning seem to be dependant on the series and level i.e. http://jsoc.stanford.edu/jsocwiki/Lev1Doc

Cadair: oh goodness, I can totally read that C code. :eyes: 

The `?QUALITY` bit of the query is always going to be the same right? If we leave the specifier section freeform we shouldn't be restricting what people can ask for?

samaloney:
Ok so based on the updated issue I think what will work is don't do anything on on search side in fetch if any of the results passed have "MISSING" flag add the quality flag and then indicate which files were not download in the `download.Results`

sunpy/net/jsoc/attrs.py
```diff
@@ -82,6 +82,14 @@ def __init__(self, value):
         self.value = value
 
 
+class Quality(SimpleAttr):
+    """
+    A quality flag
```
Cadair: I think the last thing to do is to take a run at this docstring :worried:

samaloney:
So there's a bug here which was only caught in the doc build 😟 The issue is the only primary key we always return is T_REC but T_REC is always a valid value even if the file is missing. Literally any other keyword would work the current code relies on the INSTURME key but this is not guaranteed  to be present. One possible solution is to add another key which is always retuned e.g. T_OBS but I don't know if this is a good idea?

sunpy/net/jsoc/jsoc.py
```diff
@@ -768,9 +780,10 @@ def _make_recordset(self, series, start_time='', end_time='', wavelength='',
             error_message = "Atleast one PrimeKey must be passed."
             raise ValueError(error_message)
 
-        dataset = '{series}{primekey}{segment}'.format(series=series,
-                                                       primekey=pkstr,
-                                                       segment=segment)
+        dataset = '{series}{primekey}{quality}{segment}'.format(series=series,
+                                                                primekey=pkstr,
+                                                                quality=quality,
+                                                                segment=segment)
```
Cadair: This should be an f-string :grinning:

samaloney:
Ok JSOC query showing some results with one result missing data.

image

Work out the integer numbers from the quality hex fields:
`0x00010000 = 65536`
`0xc0000000 = 3221225472`
add quaily query to select everying equal to or greater than zero `[? QUALITY >= 0 ?]`

image

Lose only the last result with quality flag of 3221225472?!? 😖💥 

So it looks like Quality is a signed 32bit integer.

`np.frombuffer(0xc0000000.to_bytes(32, 'little'), count=1, dtype=np.int32)`
`array([-1073741824], dtype=int32)`

`np.frombuffer(0x00010000.to_bytes(32, 'little'), count=1, dtype=np.int32)
`array([65536], dtype=int32)`

So the DRMS module is not returning the right value.

sunpy/net/jsoc/jsoc.py
```diff
@@ -226,6 +232,48 @@ class JSOCClient(BaseClient):
...
+    Handling missing dataQuery JSOC for some HMI observables and specif keys
+
+        >>> from sunpy.net import jsoc
+        >>> from sunpy.net import attrs as a
+        >>> client = jsoc.JSOCClient()
+        >>> segments = a.jsoc.Segment('field') & jsoc.Segment('inclination') & a.jsoc.Segment('azimuth') & a.jsoc.Segment('disambig')
+        >>> keys = ['T_REC', 'TELESCOP', 'INSTRUME', 'WAVELNTH', 'CAR_ROT', 'QUALITY']
+        >>> response = client.search(a.Time('2017/09/06 05:40', '2017/09/06 06:30'),
+        ...                          a.jsoc.Series('hmi.B_720s'), segments, a.jsoc.Keys(keys),
+        ...                          a.jsoc.Notify("sunpy@sunpy.org")) # doctest: +REMOTE_DATA
```
Cadair: If you aren't doing a `fetch()` then you don't need `Notify`.

sunpy/net/jsoc/jsoc.py
```diff
@@ -226,6 +232,48 @@ class JSOCClient(BaseClient):
...
+        2017.09.06_06:00:00_TAI  SDO/HMI HMI_COMBINED   6173.0  2194.0      65536
+        2017.09.06_06:12:00_TAI  SDO/HMI      MISSING   6173.0      -- 3221225472
+
+    Attempting to download this data with `client.fetch(response)` will result in an error as the
+    underling data file for the last entry is missing. We can use the `sunpy.net.jsoc.attrs.Quality`
+    to filter out data with a non-zero quality flag.
+
+        >>> response = client.search(a.Time('2017/09/06 05:40', '2017/09/06 06:30'),
+        ...                          a.jsoc.Series('hmi.B_720s'), segments, a.jsoc.Keys(keys),
+        ...                          a.jsoc.Quality('>= 0'), a.jsoc.Notify("sunpy@sunpy.org")) # doctest: +REMOTE_DATA
```
Cadair: Also Notify here

sunpy/net/jsoc/attrs.py
```diff
@@ -81,6 +81,45 @@ def __init__(self, value):
...
+    `MDI <http://jsoc.stanford.edu/MDI/MDI_Global.html>`_,
+    `HMI <http://jsoc.stanford.edu/doc/data/hmi/Quality_Bits/quallev1.txt>`_, and
+    `AIA <http://jsoc.stanford.edu/~jsoc/keywords/AIA/AIA02840_K_AIA-SDO_FITS_Keyword_Document.pdf>`_.
+
+    .. warning::
+        The quality field is returned as an **unsigned 32 bit** integer representation of the bit
+        flags but the attribute expects a **signed 32 bit**  integer representation of the bit
+        flags.
+
+    For example the bit flags `0b10000000000000000000000000000000` or `0x80000000` in hex have an
```
Cadair: oh my mind :boom:

sunpy/net/jsoc/jsoc.py
```diff
@@ -226,6 +233,50 @@ class JSOCClient(BaseClient):
...
 
+    *Example 3*
+
+    Handling missing data, first query JSOC for some HMI observables and specif keys with known
+    missing data::
+
+        >>> from sunpy.net import jsoc
+        >>> from sunpy.net import attrs as a
+        >>> client = jsoc.JSOCClient()
+        >>> segments = a.jsoc.Segment('field') & jsoc.Segment('inclination') & a.jsoc.Segment('azimuth') & a.jsoc.Segment('disambig')
```
Cadair: Could you wrap this line by putting it in `()` please.

sunpy/net/jsoc/jsoc.py
```diff
@@ -103,6 +103,13 @@ class JSOCClient(BaseClient):
         JSOC requires you to register your email address before requesting
         data. `See this on how to register <http://jsoc.stanford.edu/ajax/register_email.html>`__.
 
+    .. note::
+
+        Not all data returned by a `~sunpy.net.jsoc.JSOCClient.search` is available for download.
+        The status of data is encoded into the `Quality` key and its interpretation varies
+        by series and can be filtered using the `sunpy.net.jsoc.attrs.Quality` attribute
+        see Example 3 below.
```
dstansby:
```suggestion
        See Example 3 below.
```

sunpy/net/jsoc/jsoc.py
```diff
@@ -103,6 +103,13 @@ class JSOCClient(BaseClient):
         JSOC requires you to register your email address before requesting
         data. `See this on how to register <http://jsoc.stanford.edu/ajax/register_email.html>`__.
 
+    .. note::
+
+        Not all data returned by a `~sunpy.net.jsoc.JSOCClient.search` is available for download.
+        The status of data is encoded into the `Quality` key and its interpretation varies
+        by series and can be filtered using the `sunpy.net.jsoc.attrs.Quality` attribute
```
dstansby:
```suggestion
        by series and can be filtered using the `sunpy.net.jsoc.attrs.Quality` attribute.
```

sunpy/net/jsoc/jsoc.py
```diff
@@ -226,6 +233,51 @@ class JSOCClient(BaseClient):
 
         >>> res.wait(progress=True)   # doctest: +SKIP
 
+    *Example 3*
+
+    Handling missing data, first query JSOC for some HMI observables and specif keys with known
```
dstansby:
```suggestion
    Handling missing data, first query JSOC for some HMI observables and specify keys with known
```

davisite:
To help SunPy and other DRMS export clients handle DRMS records with no images, I will change the default behavior of jsoc_info and jsoc_fetch. Currently, regardless of the value of the QUALITY keyword, jsoc_info and jsoc_fetch will attempt to export all DRMS records specified by the provided record-set specification. This includes exporting DRMS records for which no image exists. This is currently a problem because an attempt to export a record that has no image will cause an export request to fail. A request for many records will therefore fail if any one of those records lacks an image.

According to the new, yet-to-be-implemented behavior, the DRMS server would skip any record that has a QUALITY value that indicates that the record contains no image. SunPy would experience this future behavior with current SunPy code. However, I would add a flag to jsoc_info/jsoc_fetch to allow the SunPy package to retain the current behavior. I'm aware that in some cases SunPy users would like to download all records, even if some lack images. Currently this causes a failure, but I would remedy this by returning a FITS file that has metadata keywords (as is the case now) but with no image if none is available (NAXIS would be 0). SunPy could then create a flag in its API methods that would allow SunPy users to override the new behavior and fetch all DRMS records. In essence, this would get the SunPy package out of the business of determining which records get downloaded - the SunPy package would merely be a conduit between the SunPy user and DRMS, and it would be the SunPy user that would control completely which records get downloaded. The SunPy package should not add any QUALITY filter to the user's record-set specification - i.e., it should not add "[? QUALITY >= 0 ?]" - when calling jsoc_info/jsoc_fetch. By setting the new SunPy API flag, the SunPy user could filter records before a download occurs.

There is a second issue to address which involves DRMS int-data-type keywords whose values are expressed as hex strings. All DRMS ints are signed 32-bit numbers, and when jsoc_info returns the value of such a keyword, it returns a hex string representing a two's complement 32-bit pattern representing the signed 32-bit value. SunPy is converting these hex string values to unsigned integers - i.e., it is not interpreting the hex string as a representation of a 32-bit signed number, it is interpreting it as an unsigned number. QUALITY is one such keyword, but any int keyword may be expressed in this manner. QUALITY is actually a bit-field, not an integer. But since there is no bit-field DRMS data type, DRMS uses a signed 32-bit int type to store the 32-bit bit-field. 

To allow the SunPy user to properly use any DRMS int keyword value that is represented by a hex string, the conversion to a Python integer that the SunPy package uses would need to be modified. Here's one way to solve this problem:

1. SunPy would interpret all integer values expressed as hex strings as two's complement 32-bit integers (int() interprets its first argument as the pattern for an unsigned integer):
>>> key_hex_str = '0x80C07500'
>>> key_value = twos(int(key_hex_str, 16), 4)
>>> key_value
-2134870784

This would provide the correct, signed value to the SunPy user. Since the SunPy package has no knowledge whether this value is a bit-field, this is the proper thing to do.

2. A SunPy user who wishes to filter on a bit-field keyword would need to perform bitwise operations; to do so, they would need to recover the original bit-field (bit manipulation on negative values in Python can lead to unexpected artifacts since Python has no limit on the number of bytes used to represent integers; negative numbers are represented with two's complement bit patterns, but because there is no limit on the number of bits in an integer, the 31st bit might be unexpectedly set after bitwise operations):
>>> key_value_bytes = int(key_value).to_bytes(4, byteorder=sys.byteorder, signed=True)
>>> key_value_bytes
b'\x00u\xc0\x80'

This would produce a byte array that is identical to the original unsigned integer's byte array:
>>> int(key_hex_str, 16).to_bytes(4, byteorder=sys.byteorder, signed=False)
b'\x00u\xc0\x80'

This byte array could then be used to generate an int that could be used for bitwise operations:
>>> key_value_unsigned = int.from_bytes(key_value_bytes, byteorder=sys.byteorder, signed=False)
>>> key_value_unsigned
2160096512

3. The SunPy user could then filter on bits:
>>> key_value_unsigned & (1 << 31) != 0
True
>>> key_value_unsigned & (1 << 30) != 0
False
>>> key_value_unsigned & (1 << 8) != 0
True
>>>

This would allow the user to filter DRMS records post-download.

4. It might be nice for SunPy to offer a higher-level mechanism for a SunPy user to filter on bits. There are some Python external packages (like bitarray) that provide a bit-field type with bitwise operations. numpy also provides some bitwise operations on arrays of bits (https://numpy.org/doc/stable/reference/routines.bitwise.html).


`twos` is a function that interprets the bit pattern for the unsigned int `value` parameter as a `bytes`-bytes bit pattern for a two's complement integer. Here's one possibility:

import sys
def twos(value, bytes):
&nbsp;&nbsp;&nbsp;&nbsp;value_bytes = value.to_bytes(bytes, byteorder=sys.byteorder, signed=False)                                                          
&nbsp;&nbsp;&nbsp;&nbsp;return int.from_bytes(value_bytes, byteorder=sys.byteorder, signed=True)

PR #3769, opened 2020-02-06
Cleanup to Attrs

Cadair:
Missed a couple of little things in #3714


abhijeetmanhas:
There are no `Filter`, `Physobs`, `Source` classes in `_atts.py` however it is in `vso/attrs.py`. Should they also be added in `net/_attrs` ? I mean should _attrs in net should have all classes defined in vso/attrs ?

Cadair:
You are moving the Filter, Physobs, and Source attrs out of vso.attrs in your PR, in master they were still specific to VSO, which is why these two PRs haven't moved them.

PR #3772, opened 2020-02-06
Fix the calculation of Carrington longitude as seen from Earth (again!)

ayshih:
In my continuing battle with Carrington longitude (see #2224 and #3178), here's the latest fix for the calculation of the Carrington longitude of disk center as seen from Earth (i.e., `sunpy.coordinates.sun.L0()`.  I have updated the approach from the pre-2007 approach to the post-2007 approach (see [this 2007 IAU report](https://link.springer.com/article/10.1007/s10569-007-9072-y) for context).  The biggest difference is that the approach now accounts for the varying amount of light travel time because the Sun-Earth distance isn't always 1 AU, but there is also subtlety with respect to having to apply a correction for the aberration due to Earth motion.

The old method resulted in errors no greater than 7 arcseconds in Carrington longitude (see plot).
In fact, the errors were essentially below the precision level of the published reference we use for tests (_The Astronomical Almanac_).
difference2
_The Astronomical Almanac_ updated their approach starting with their 2009 edition, and in [their write-up of the impact](http://asa.usno.navy.mil/static/files/sun_rotation_change.pdf), you can see that their Figure 5 is the same as the above plot (just for a different year).

Larger changes will be coming down the pipe for Carrington longitude to account for non-Earth observers, but those won't be mere bug fixes.


nabobalis:
Changelog is missing.

ayshih:
Ha ha, I know about the missing changelog! =)  I guess I ought to have labeled this "WIP" while I was waiting to see what tests broke.

Cadair:
You can open PRs as draft (dropdown next to the open PR button), and then it also doesn't request reviews, for future reference.

ayshih:
Oh, do draft PRs still run the CI?  I'll definitely do that next time then.

Cadair:
Yeah the CI still runs.

nabobalis:
Looks like three test failures.

ayshih:
I still have one broken test to fix, and the backports will presumably need to be done manually because I needed to fix a doctest in a 2.0-only file.

changelog/3772.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed a bug with the calculation of Carrington longitude as seen from Earth where the true Sun-Earth distance was not properly considered.
+This bug resulted in errors no greater than 7 arcseconds in Carrington longitude when using `~sunpy.coordinates.sun.L0` and `~sunpy.coordinates.frames.HeliographicCarrington`.
```
Cadair: It makes me happy that this is the level of accuracy we are working to :grinning:

PR #3537, opened 2019-11-22
Coordinate frames that include solar differential rotation

ayshih:
This PR integrates solar (differential) rotation into the coordinates framework.  The approach here is to enable defining a frame that is a modified version of a base coordinate frame by a duration of solar (differential) rotation.  The class factory `RotatedSunFrame` is supplied the "base" coordinate frame, and creates a base-specific rotated-Sun frame class.

Closes #3357

Still needs:
- [x] Add tutorial ([see this recent build](https://30287-2165383-gh.circle-artifacts.com/0/home/circleci/project/docs/_build/html/code_ref/coordinates/rotatedsunframe.html))
- [x] Finish writing tests
- [x] Support array `obstime`/`duration`/`rotated_time`

# Walkthrough
```python
>>> import astropy.units as u
>>> from sunpy.coordinates import RotatedSunFrame
>>> import sunpy.coordinates.frames as f
```

## Propagating a coordinate
Let's define a coordinate array of points in a rotated-Sun frame that in the base HGS frame were on the meridian as seen from Earth.  Here, the rotated-Sun frame applies +1 day of solar rotation.
```python
>>> hgs = f.HeliographicStonyhurst(obstime='2001-01-01')
>>> rotated_arc = RotatedSunFrame([0]*11*u.deg, range(-75, 90, 15)*u.deg, base=hgs, duration=1*u.day)
>>> rotated_arc
<RotatedSunHeliographicStonyhurst Coordinate (base=<HeliographicStonyhurst Frame (obstime=2001-01-01T00:00:00.000)>, duration=1.0 d, rotation_model=howard): (lon, lat, radius) in (deg, deg, km)
    [(0., -75., 695700.), (0., -60., 695700.), (0., -45., 695700.),
     (0., -30., 695700.), (0., -15., 695700.), (0.,   0., 695700.),
     (0.,  15., 695700.), (0.,  30., 695700.), (0.,  45., 695700.),
     (0.,  60., 695700.), (0.,  75., 695700.)]>
```
The coordinate components are relative to the distorted axes of the rotated-Sun frame, and thus are *already* differentially rotated in real space.  Convert the coordinate array back to the HGS frame to see what I mean.
```python
>>> rotated_arc.transform_to(hgs)
<HeliographicStonyhurst Coordinate (obstime=2001-01-01T00:00:00.000): (lon, lat, radius) in (deg, deg, km)
    [(10.7550473 , -75., 695700.), (11.70697161, -60., 695700.),
[...]

PR #3775, opened 2020-02-07
Fixes to the transformation test for NorthOffsetFrame

ayshih:
- Fixed the specifications of the absolute tolerances
- Extended the `hypothesis` deadline to 1 second (from the default of 200 ms)

Context: The first call to `NorthOffsetFrame` takes longer than subsequent calls (using the same base frame) because it needs to create the class.  This was causing unreliable test timings, with a misleading error message that the first call was "falsified".

### DIFF
--- a/changelog/3775.trivial.rst
+++ b/changelog/3775.trivial.rst
@@ -1,0 +1,1 @@
+Fixed the transformation test for `~sunpy.coordinates.metaframes.NorthOffsetFrame`, which would intermittently fail.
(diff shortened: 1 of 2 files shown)


Cadair:
@nabobalis you noticed this fail anywhere other than master?

nabobalis:
Nope. The last time I saw the branches they were all ok.

nabobalis:
But would it cost much to backport?

nabobalis:
I wonder why that failed.

ayshih:
The backports have to be done manually because these lines were modified earlier in #3537, which can't be backported.  I'll whip up the backports later this morning.

nabobalis:
No rush!

PR #3776, opened 2020-02-08
Added example of world_to_pixel to Units & Coordinates guide

### LINKED ISSUE #3278: Units and Coordinates page of the Guide should show a world_to_pixel example.
http://docs.sunpy.org/en/stable/guide/units-coordinates.html#physical-coordinates-in-sunpy

pkj-m:
### Description

Added an example of `world_to_pixel` to the Units & Coordinates guide. 
Example is taken from https://docs.astropy.org/en/stable/wcs/wcsapi.html

Fixes #3278


nabobalis:
Personally, I'm not sure that using the astropy example makes any sense for two reasons.

1. It isn't solar specific.
2. A link to their example in the docs would make more sense if that example were to change. 

I think adding an extension to the subsection in the original issue might be the way to go.

pkj-m:
Got it. Also, can you please explain what you mean by adding an extension to the subsection in the original issue? Are we talking about the astropy docs and adding a solar example there or something else?

nabobalis:
The original issue links to a subsection of that documentation page with an example. Adding the `world_to_pixel` to that part to me seems the best place for it.

pkj-m:
@nabobalis for creating the WCS object should I use the `AIA_171_IMAGE` fits file or should I use the one given in the astropy docs?

nabobalis:
You could use the one that exists on the current documentation page?

pkj-m:
In that case I'll need to create a new SkyCoord object since the dimension of `AIA_171_IMAGE` file does not match with the current SkyCoord object's dimension. Is that alright?

nabobalis:
Sure.

Cadair:
Thanks for the PR @pkj-m ! This should be using the map APIs for these transforms, i.e. `sunpy.map.GenericMap.world_to_pixel`.

pkj-m:
@Cadair could you please review the PR? Also, can you tell me what is failing the CI test? It says 'Fix End-of-File` test is failing but I cannot figure out why it would say so?

docs/guide/units-coordinates.rst
```diff
@@ -231,5 +231,9 @@ Using Coordinates with SunPy Map
    <matplotlib.image.AxesImage object at ...>
    >>> _ = ax.plot_coord(c, 'o')  # doctest: +REMOTE_DATA
 
-For more information on coordinates see :ref:`sunpy-coordinates` section of
-the :ref:`reference`.
```
nabobalis: Why was this reference deleted?

changelog/3776.doc.rst
```diff
@@ -0,0 +1 @@
+Added an example for `~sunpy.map.GenericMap.world_to_pixel` function in the Units & Coordinates guide
```
nabobalis:
```suggestion
Added an example for `~sunpy.map.GenericMap.world_to_pixel` function in the Units & Coordinates guide.
```

docs/guide/units-coordinates.rst
pkj-m: Oops. My bad, I'll revert the change.

nabobalis:
**NOTE TO WHO MERGES THIS: SQUASH MERGE**

pkj-m:
@nabobalis could you please rerun the tests for this PR, I think the failures are due to some server issues?

nabobalis:
The documentation build passed, that is enough.

nabobalis:
Thanks @pkj-m

PR #3781, opened 2020-02-10
[Backport 1.0] Fixes to Map plotting in user guide

ayshih:
Backport #3454.


ayshih:
Better late than never! =P

PR #3782, opened 2020-02-10
Carrington coordinates for non-Earth observers

ayshih:
Currently, our Carrington coordinate frame (`HeliographicCarrington`) essentially assumes an observer at Earth when transforming between that frame and other frames.  The location of the observer matters because the Sun rotates during the time it takes light to travel between the Sun and the observer.  If the observer is at a distance from the Sun that is not the same as the Sun-Earth distance, that observer will see the Sun at a different phase of its rotation than observers at Earth.  Note that only the Sun-observer distance matters for this consideration, not the heliographic longitude/latitude of the observer.  While this effect is very small for observers at ~1 AU from the Sun (e.g., STEREO), it will matter for imagery from Solar Orbiter (now launched!).

With this PR, `HeliographicCarrington` is changed to be an observer-based frame:

- The default is `observer=None`, which is *not* the same as our current `HeliographicCarrington` frame.  ~Instead, it gives the instantaneous (or "true") Carrington coordinates (i.e., assuming zero light travel time).~  This approach turned out to have issues.  Instead, `observer=None` will be treated the same way as with other observer-based frames: valid to be created, but can't be transformed with.  (If one wants "true" Carrington coordinates, one can specify an observer at Sun center!  Edit: This will actually be slightly different from "true" because there will still be a light-travel-time correction to the Sun surface.)
- To get our current behavior*, you need to use something like `HeliographicCarrington(observer='earth', obstime=...)`.
- To get the frame for, say, an AIA observation, you would use `HeliographicCarrington(observer=aiamap.observer_coordinate, obstime=aiamap.date)`.
- If you're working with a `SkyCoord` in an observer-based frame (e..g, `Helioprojective`), the transformation `.heliographic_carrington` should propagate the `observer`/`obstime` information automatically.

\* Technically, it will not be our "current behavior", because I have introduced a ~20 arcsecond shift.  See below.

Some considerations:

- The phase of the Sun's rotation here is in the sense of the Carrington coordinate frame, which rotates non-differentially.
- Interactions with `RotatedSunFrame` may be confusing, since that will mix non-differential rotation with differential rotation.

## Major update ##

The calculation approach for Carrington coordinates has been changed to exclude the application of any stellar-aberration correction due to motion of the observer (Earth or otherwise).  For Earth motion, that means a increase of ~20 arcseconds in Carrington longitude.  This change is critical to enable the precise co-alignment of solar images from different observatories (see the long write-up in this PR).  Unfortunately, this change breaks our agreement the *Astronomical Almanac* (when using default settings for `sun.L0()`), as well as earlier versions of SunPy.


dstansby:
Just to make sure I'm getting this, this means `HeliographicCarrington('earth', time)` is the Carrington frame at a time `time - light_travel_time_from_earth_to_sun_center`?

ayshih:
> Just to make sure I'm getting this, this means `HeliographicCarrington('earth', time)` is the Carrington frame at a time `time - light_travel_time_from_earth_to_sun_center`?

Yup.

ayshih:
Per dev meeting:

- [ ] Check whether grid overlays work
- [ ] Add more comments to explain the fake observer "hack" for `.observer_coordinate`
- [ ] Make sure that `.observer_coordinate` doesn't return the fake observer in the `SkyCoord`

Cadair:
Is this ready for review @ayshih ?

ayshih:
> Is this ready for review @ayshih ?

Yes!

I just checked against SPICE this morning to confirm its options for handling observer motion, and I added corresponding footnotes to the write-up.

ayshih:
@Cadair I just realized that I need to update the ASDF schema to also have the `observer`.  I presume the schema version should be bumped?

Cadair:
> I presume the schema version should be bumped?

I think we need to consider what happens if someone loads a file written with the current schemas with this PR. i.e. what is the default is observer is not specified? If we don't need to (want to) special case that situation then we probably don't need to bump the schema. If we want to special case the situation where you are loading an old schema to default the observer to Earth or something then we will need to bump the version and keep both implementations around for a while.

ayshih:
I have bumped the version of the ASDF schema, and loading an old schema will be assigned the observer as Earth.

ayshih:
Okay, the dust has finally settled.  Review away!

docs/code_ref/coordinates/carrington.rst
```diff
@@ -0,0 +1,126 @@
...
+There are two primary effects for why the apparent sub-observer point is not the same as the "true" sub-observer point.
+
+Light travel time
+^^^^^^^^^^^^^^^^^
+
+It takes time for light to travel from the Sun to any observer (e.g., ~500 seconds for an observer at Earth).
+Thus, the farther the observer is from the Sun, the farther back in time the Sun will appear to be.
+This effect is sometimes called "planetary aberration".
+
+As an additional detail, the observer is closer to the sub-observer point than to the center of the Sun (i.e., the radius of the Sun), which corresponds to ~2.3 lightseconds.
```
dpshelio: In the discussion above you said that  `HeliographicCarrington('earth', time)` is the Carrington frame at a time `time - light_travel_time_from_earth_to_sun_center`,  wouldn't be at sun surface?

docs/code_ref/coordinates/carrington.rst
```diff
@@ -0,0 +1,126 @@
...
+A more subtle question is whether observer motion also shifts the apparent sub-observer Carrington longitude.
+
+Stellar aberration causes the apparent ecliptic longitude of an observer to be different from its true ecliptic longitude.
+This shift in ecliptic longitude also means a shift in the heliographic longitude of the observer.
+For example, for an Earth observer, the apparent heliographic longitude of Earth is ~0.006 degrees (~20 arcseconds) less than its true heliographic longitude.
+For some purposes, this shift in the heliographic longitude of the observer should also shift the apparent sub-observer Carrington longitude.
+
+However, shifting the sub-observer Carrington longitude due to observer motion is not always appropriate.
+For certain types of observations (e.g., imagery of features on the surface of the Sun), stellar aberration does not shift the positions of solar features **relative** to the center of the disk of the Sun; everything shifts in tandem.
+Thus, for the purposes of co-aligning images from different observatories, it is critical to **exclude** any correction for observer motion.
```
dpshelio: Very clear section!!

docs/code_ref/coordinates/carrington.rst
```diff
@@ -0,0 +1,126 @@
...
+The exclusion of a stellar-aberration correction is appropriate for purposes such as image co-alignment, where Carrington longitudes must be consistently associated with features on the Sun's surface.
+
+Comparisons to other sources
+----------------------------
+
+Compared to |AA| and older methods of calculation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+|AA| publishes the apparent sub-Earth Carrington longitude (|L0|), and these values include the stellar-aberration correction.
+Consequently, SunPy values will be consistently ~0.006 degrees (~20 arcseconds) greater than |AA| values, although these discrepancies are not always apparent at the printed precision (0.01 degrees).
```
dpshelio: Knowing this is consistent is very good! Nice point about the printed precision!

docs/code_ref/coordinates/carrington.rst
```diff
@@ -0,0 +1,126 @@
...
+In `JPL Horizons <https://ssd.jpl.nasa.gov/?horizons>`_, one can request the "Obs sub-long & sub-lat".
+JPL Horizons appears to start from the IAU parameters and to include the correction for light travel time but not the correction for observer motion (i.e., equivalent to specifying "LT" to `subpnt`_ in `SPICE`_).
+The discrepancy between SunPy values and JPL Horizons values is no greater than 0.01 arcseconds.
+
+Compared to SunSPICE
+^^^^^^^^^^^^^^^^^^^^
+
+In `SunSPICE <https://stereo-ssc.nascom.nasa.gov/sunspice.shtml>`_, one can convert to and from the "Carrington" coordinate system using the function `convert_sunspice_coord <https://hesperia.gsfc.nasa.gov/ssw/packages/sunspice/idl/convert_sunspice_coord.pro>`_.
+However, these Carrington longitudes are "true" rather than "apparent" because the observer is not specified, so there are no corrections for light travel time or for observer motion.
+Thus, SunPy values will be consistently greater than SunSPICE's coordinate transformation by ~0.82 degrees.
```
dpshelio: 0.8 degrees? That's a big chunk. Is that much the difference?

dpshelio: Also, for all the comparisons. Are these discrepancies only in longitude, in latitude or combined?

sunpy/coordinates/ephemeris.py
```diff
@@ -263,15 +264,26 @@ def _detilt_lon(coord):
...
 @add_common_docstring(**_variables_for_parse_time_docstring())
-def _L0(time='now'):
+def _L0(time='now',
+        light_travel_time_correction=True,
+        nearest_point=True,
+        aberration_correction=False):
     """
-    Return the L0 angle for the Sun at a specified time, which is the Carrington longitude of the
-    Sun-disk center as seen from Earth.
+    Return the L0 angle for the Sun at a specified time, which is the apparent Carrington longitude
```
dpshelio: it's only the `apparent` if run with the default values, right?

sunpy/coordinates/frames.py
```diff
@@ -252,19 +252,23 @@ class HeliographicCarrington(BaseHeliographic):
     - The Z-axis (+90 degrees latitude) is aligned with the Sun's north pole.
     - The X-axis and Y-axis rotate with a period of 25.38 days.
 
-    This system differs from Stonyhurst Heliographic (HGS) in its definition of longitude.
+    This system differs from Stonyhurst Heliographic (HGS) in its definition of longitude.  This
+    longitude is an "apparent" longitude because it takes into account the time it takes for light
```
dpshelio: I thought we were converting it to the "true" longitude, and not the apparent.

sunpy/coordinates/tests/test_sun.py
```diff
@@ -180,19 +181,87 @@ def test_B0():
...
+                                      -7.219984,
+                                      -6.543092,
+                                      -4.168016,
+                                      -0.663804,
+                                       2.873030,
+                                       5.784693,
+                                       7.194559,
+                                       6.719875,
+                                       4.378979,
+                                       0.881068]*u.deg, atol=0.01*u.arcsec, rtol=0)
```
dpshelio: To make this more readable I would create a dict or list with times and results with somethink like:
```python
jpl_values = {'2013-01-01': -3.034687,
              '2013-02-01': -6.031910,
             ...}
# and then calculate the times and the tests as:
sun_B0 = sun.B0(Time(jpl_values.keys(), scale='tt'))
assert_quantity_allclose(sun_B0, jpl_values.values() * u.deg, atol=0.01*u.arcsec, rtol=0)
```
It's easier to match date and values.

sunpy/coordinates/tests/test_sun.py
```diff
@@ -180,19 +181,87 @@ def test_B0():
...
+                              -7.2199937,
+                              -6.5430498,
+                              -4.1679382,
+                              -0.66371004,
+                               2.8731155,
+                               5.7847503,
+                               7.1945709,
+                               6.7198381,
+                               4.3789008,
+                               0.88096965]*u.deg, atol=0.005*u.arcsec, rtol=0)
```
dpshelio: Same as above.

sunpy/coordinates/tests/test_sun.py
```diff
@@ -180,19 +181,87 @@ def test_B0():
...
+
+
+def test_L0_defaults():
+    assert_quantity_allclose(sun.L0('1992-Oct-13'),
+                             sun.L0('1992-Oct-13',
+                                    light_travel_time_correction=True,
+                                    aberration_correction=False,
+                                    nearest_point=True))
```
dpshelio: Would it better to hard-code the precission we expect?

docs/code_ref/coordinates/carrington.rst
```diff
@@ -0,0 +1,126 @@
+.. _sunpy-coordinates-carrington:
+
+Calculating Carrington longitude
+================================
+
+Carrington coordinates are a heliographic coordinate system, where the longitude is determined assuming that the Sun has rotated at a sidereal period of ~25.38 days starting at zero longitude at a specific time on 1853 November 9.
+For a given observer at a given observation time, the center of the disk of the Sun as seen by that observer (also known as the apparent sub-observer point) has a Carrington latitude and longitude.
+When the observer is at Earth, the apparent sub-Earth Carrington latitude and longitude are also known as |B0| and |L0|, respectively.
```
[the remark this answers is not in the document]
ayshih: > * Why we said here "apparent"?

The apparent sub-observer point is different from the true sub-observer point because the observer is seeing light from the past.  For example, the apparent center of the disk for an observer at Earth is ~500 seconds in the past; whatever the observer sees at disk center is no longer actually there in "true" coordinates.

> * What does "sub" in sub-Earth means? Should it be "sun-Earth"?

"sub-Earth" is a special case of "sub-observer".  "sub-observer" point means the point on the surface directly "below" the observer.

docs/code_ref/coordinates/carrington.rst
ayshih: Ah, good point.  That earlier comment pre-dated some of the finer corrections implemented later.  I'll amend the earlier comment.

docs/code_ref/coordinates/carrington.rst
ayshih: Yup.  It's the difference between "true" and "apparent" coordinates, so it shouldn't match.

ayshih: The short answer is that we agree in latitude with JPL Horizons and SPICE.  *The Astronomical Almanac* does the same thing with applying a stellar-aberration correction, so that causes us to disagree with them up to a few arcseconds, but that would almost never show up at their printed precision.

docs/code_ref/coordinates/index.rst
```diff
@@ -206,8 +206,8 @@ coordinates is::
    <SkyCoord (Helioprojective: obstime=2017-07-26T00:00:00.000, rsun=695700.0 km, observer=<HeliographicStonyhurst Coordinate for 'earth'>): (Tx, Ty) in arcsec
```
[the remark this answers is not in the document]
ayshih: > Shouldn't [line 54](https://github.com/sunpy/sunpy/pull/3782/files#diff-0ed4b40bbcab180c906609a092ecf502R54) and 58 say heliographic instead of heliocentric?

Ha!  Good catch.

> Should we also mention on this page (under section of [line 166](https://github.com/sunpy/sunpy/pull/3782/files#diff-0ed4b40bbcab180c906609a092ecf502R54)) that the observer is needed on Carrington's or it will be set as `None`?

I'd argue that there's no reason to over-stuff this section (which is just about accessing coordinate components).  I'll add a sentence in the observer section.

sunpy/coordinates/ephemeris.py
ayshih: Yes, if you turn off both the light-travel-time correction and the stellar-aberration correction, you will get the true Carrington longitude.

sunpy/coordinates/frames.py
ayshih: Nope, the point of `HeliographicCarrington` is to do apparent longitudes for the specified observer.

sunpy/coordinates/tests/test_sun.py
ayshih: Good suggestion, will do

sunpy/coordinates/tests/test_sun.py
ayshih: This unit test is just about checking defaults, so the return values should match exactly.  I could use `==` checks instead of `allclose` to make that clear.

docs/code_ref/coordinates/carrington.rst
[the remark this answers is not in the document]
dpshelio: Re. "aparent": Gotya! I understood that later on in the text. I wonder if that could have this comment as a footnote. 
Re. "sub-Earth": Aaahhh, I didn't know that. Is that for the Earth a constant value or there are variations from observatories at higher mountains and at different places on Earth? Maybe a foot-note with that definition there could be also helpful. 

Maybe I've forgotten all that I knew about observatories and observations... so maybe it's just me and it's common knowledge across the community. If that's the case, then we don't need to define these terms. :grin:

sunpy/coordinates/ephemeris.py
dpshelio: OK, then that first sentence could be misleading as to what the function does (or can do).

docs/code_ref/coordinates/carrington.rst
dpshelio: Ok, I thought I got few things wrong. First I thought to understand that it was ~20 arcsecs, not 0.8 degrees. But now, reading the rest of the comments I am getting to a different conclussion. The 20 arcsecs are between the AA and us using apparent (because they are including stellar aberration). And the apparent is what makes it easier to align images.. right? (I would have thought the opposite, as the true values will keep it closer to truth).

sunpy/coordinates/ephemeris.py
ayshih: I added a sentence

docs/code_ref/coordinates/carrington.rst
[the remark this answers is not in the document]
ayshih: I added a footnote

docs/code_ref/coordinates/carrington.rst
ayshih: First, there are three categories here:

1. true (which is the raw coordinate transformation in SunSPICE)
2. apparent, excluding stellar aberration (which is what we do and what JPL Horizons does)
3. apparent, including stellar aberration (which is what *The Astronomical Almanac* does)

The difference between (1) and (2) is ~0.82 degrees in longitude, and the difference between (2) and (3) is ~0.006 degrees (~20 arcseconds) in longitude.

Second, for co-aligning images, we need to use (2).  We can't use (1) because the images that we record are affected by light travel time (for example, Earth observers record where solar features were ~500 seconds in the past), so we need coordinates that are also affected by light travel time.  We can't use (3) because observer motion doesn't actually shift the positions of solar features relative to disk center.

docs/code_ref/coordinates/carrington.rst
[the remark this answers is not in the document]
ayshih: I forgot to add a footnote for "sub-Earth".

Sub-Earth, without any qualification, means the center of the Earth.  Of course, ground-based observatories aren't at the center of the Earth, so there will be some very small differences in their sub-observer longitudes.

docs/code_ref/coordinates/carrington.rst
dpshelio: And know I understand it all!! :tada:  Thanks (and also thanks for your patience ;) )

ayshih: Thank *you* for reviewing the PR!

sunpy/coordinates/transformations.py
```diff
@@ -606,11 +619,17 @@ def hgc_to_hgc(from_coo, to_frame):
     """
     Convert between two Heliographic Carrington frames.
     """
-    if np.all(from_coo.obstime == to_frame.obstime):
+    if _observers_are_equal(from_coo.observer, to_frame.observer) and \
```
Cadair: *twitch*

docs/code_ref/coordinates/carrington.rst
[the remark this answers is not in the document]
ayshih: Footnote added

sunpy/coordinates/ephemeris.py
```diff
@@ -296,14 +309,18 @@ def _L0(time='now'):
...
     # This expression is an approximation to reduce computations (e.g., it does not account for the
     # inclination of the Sun's rotation axis relative to the ecliptic), but the estimated error is
     # <0.2 arcseconds
-    dlon_earth -= 20.496*u.arcsec * 1*u.AU / earth.radius
+    if aberration_correction:
+        dlon_earth -= 20.496*u.arcsec * 1*u.AU / earth.radius
 
     # Antedate the observation time to account for light travel time for the Sun-Earth distance
-    antetime = obstime - earth.radius / speed_of_light
+    antetime = obstime - distance / speed_of_light if light_travel_time_correction else obstime
```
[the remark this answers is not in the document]
ayshih: Done

sunpy/map/mapbase.py
```diff
@@ -313,7 +313,7 @@ def wcs(self):
         # need to do this only based on .meta.
         ctypes = {c[:4] for c in w2.wcs.ctype}
         # Check that the ctypes contains one of these two pairs of axes.
```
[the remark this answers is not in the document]

sunpy/io/special/asdf/schemas/sunpy.org/sunpy/coordinates/frames/heliographic_carrington-1.1.0.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - $ref: "tag:astropy.org:astropy/coordinates/frames/baseframe-1.0.0"
+  - properties:
+      frame_attributes:
+        type: object
+        properties:
+          observer:
+            anyOf:
+              - $ref: "heliographic_stonyhurst-1.0.0"
+              - type:
+                  string
```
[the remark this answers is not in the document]
ayshih: Done

docs/code_ref/coordinates/carrington.rst
```diff
@@ -0,0 +1,127 @@
...
+For a given observer at a given observation time, the center of the disk of the Sun as seen by that observer (also known as the apparent [#apparent]_ sub-observer point) has a Carrington latitude and longitude.
+When the observer is at Earth, the apparent sub-Earth Carrington latitude and longitude are also known as |B0| and |L0|, respectively.
+
+Quick summary
+-------------
+SunPy calculates the sub-observer Carrington longitude in a manner that enables co-alignment of images of the Sun's surface from different observatories at different locations/velocities in the solar system, but that results in a ~20-arcsecond discrepancy with |AA|.
+
+Observer effects
+----------------
+There are two primary effects for why the apparent sub-observer point is not the same as the "true" sub-observer point.
```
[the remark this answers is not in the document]
ayshih: Sure, done

docs/code_ref/coordinates/carrington.rst
```diff
@@ -0,0 +1,127 @@
...
+In `JPL Horizons <https://ssd.jpl.nasa.gov/?horizons>`_, one can request the "Obs sub-long & sub-lat".
+JPL Horizons appears to start from the IAU parameters and to include the correction for light travel time but not the correction for observer motion (i.e., equivalent to specifying "LT" to `subpnt`_ in `SPICE`_).
+The discrepancy between SunPy values and JPL Horizons values is no greater than 0.01 arcseconds.
+
+Compared to SunSPICE
+^^^^^^^^^^^^^^^^^^^^
+
+In `SunSPICE <https://stereo-ssc.nascom.nasa.gov/sunspice.shtml>`_, one can convert to and from the "Carrington" coordinate system using the function `convert_sunspice_coord <https://hesperia.gsfc.nasa.gov/ssw/packages/sunspice/idl/convert_sunspice_coord.pro>`_.
+However, these Carrington longitudes are "true" rather than "apparent" because the observer is not specified, so there are no corrections for light travel time or for observer motion.
+Thus, SunPy values will be consistently greater than SunSPICE's coordinate transformation by ~0.82 degrees.
```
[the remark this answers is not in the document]
ayshih: Clarified

docs/code_ref/coordinates/carrington.rst
```diff
@@ -0,0 +1,127 @@
...
+
+Footnotes
+---------
+
+.. [#apparent] The term "apparent" (as opposed to "true") is used to indicate that the observer perceives the locations and orientations of objects different from how they truly are due to effects that include light travel time.
+.. [#Carrington] Carrington (1863), *Observations of the Spots on the Sun*, p. 244
```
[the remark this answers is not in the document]
ayshih: Links added to archive.org

docs/code_ref/coordinates/carrington.rst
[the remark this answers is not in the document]
ayshih: Using the same footnote twice looks weird (at least in our docs style), so I added a sentence instead.

sunpy/coordinates/transformations.py
ayshih: Heh, no comment

Cadair:
:shipit:

PR #3803, opened 2020-02-14
Update citation information with ApJ and JOSS papers

Cadair:
I am not sure what we want to do with regards to recommending people site the ApJ and JOSS papers?


nabobalis:
To cite both?

sunpy/CITATION.rst
```diff
@@ -13,33 +13,35 @@ Please add the following line within your methods, conclusion or acknowledgement
    *This research has made use of SunPy vX.Y (version citation), an open-source and free
    community-developed solar data analysis Python package (citation).*
 
-The citation should be to the `SunPy paper`_ and the version number should cite the specific `Zenodo DOI`_ for the version used in your work.
+The citation should be to the `SunPy paper`_, the `SunPy JOSS paper`_ and the version number should cite the specific `Zenodo DOI`_ for the version used in your work.
```
hayesla: yes maybe we should be clearer here that both should be cited?

dstansby:
I think I'm -1 on asking people to cite both papers - it's extra effort for the citer, and I'm not sure what the added benefit is for SunPy. If we want to track citations (and therefore 'impact' of sunpy), then citing one paper should be good enough for that.

nabobalis:
I think if we ask to cite one, it would be  just the ApJ paper in my mind.

wtbarnes:
I don't think it is too onerous to ask people to cite both. It is just a matter of adding another entry to your BibTeX file...

If we really want to go with one or the other though, I would say we should ask people to cite the JOSS paper as that constitutes a review of the code itself while the ApJ paper was just a review of the paper.

dstansby:
> I don't think it is too onerous to ask people to cite both. It is just a matter of adding another entry to your BibTeX file...

It's not a lot of extra effort, but it is still extra effort, for (as far as I can see) no gain on the SunPy side.

wtbarnes:
Sure, but if we are not going to ask people to cite the JOSS paper, why did we write and submit it in the first place?

I think this probably needs more discussion, perhaps during one of the community meetings.

mbobra:
This is the recommendation I gave in the SunPy survey paper:
 
> To cite scientific software, follow these two steps:
> 
> 1. Cite the refereed journal article describing the research software. To find this article, visit the website for a software package and look for citation instructions. For example, the SunPy website includes dedicated citation instructions and an associated BibTex entry.
> 2. Cite the software archive. Publicly available digital repositories, such as Zenodo, issue a Digital Object Identifier (DOI) for archived software. (Some institutions also provide digital repositories as part of their library system). Generally, open source software projects in the Python scientific stack will archive their software every time they release a new version. For example, the SunPy Github page includes the Zenodo DOI for the most recent release (as of this writing, v1.1.0); clicking on it leads to the Zenodo deposit, which provides an associated BibTex entry.
> 
> Many projects release multiple versions of software per year, but only write a refereed journal article once in a while (for example, the SunPy Project published a paper about the v1.0 release, but will not publish a paper on the v1.1.0 release). Therefore, creating reproducible results requires citing both the journal article and the software archive. Here is an example: This analysis used version v1.1.0 (Mumford et al., 2020a) of the SunPy open source software package (The SunPy Community et al., 2020).
> 

I'm happy to change this if people don't agree (the survey paper is in review, so none of this is published). But as it stands, I recommended two citations: citing the ApJ paper and the Zenodo archive for whatever version used in the research. 

I agree with @wtbarnes that citing both ApJ and JOSS is not a problem per se, but I worry people will not want three different citations to SunPy in their papers. I also think if we can amass a huge number of citations for the ApJ paper, instead of splitting it with JOSS, we can use this number to our advantage when pursuing funding opportunities in the future.

Cadair:
The suggestion was brought up at the community meeting today that we do mention the JOSS paper in the citation instructions as an optional thing to do?

dpshelio:
I would change the "order" towards giving more importance to the zenodo entry for that release they used and suggest to include at least, one of the two papers. I don't have a strong opinion on one or the other or both. My motivation doesn't go to get high numbers but replicable science, so if people are not sharing the code that they used, then the citation to the zenodo may not be that important, but refering to the papers do.

The problem with zenodo is that people may not be fully aware of what they are citing to, all the versions, one in particular, the latest even though they've not run it...  Maybe we should have a function that provides the right one for the sunpy you are using. Which reminds me... this PR would also need to change __init__.py which is only returning a single "journal" citation and not all.

Cadair:
> this PR would also need to change `__init__.py` which is only returning a single "journal" citation and not all

The `__init__.py` reads this file to extract the bibtex.

I like the idea of extracting the bibtex / citation info for the running sunpy in the `__citation__`, but that should be a separate issue.

I am going to rewrite this, heavily prioritise the ApJ paper (and Zenodo) but still mention the JOSS publication.

Cadair:
Ok, I added a separate section entitled "other sunpy publications" in which I include a link to the JOSS paper (and it's bibtex) as well as links to the old v0.5 paper. Thoughts?

sunpy/__init__.py
```diff
@@ -32,15 +32,18 @@ class UnsupportedPythonError(Exception):
...
-        if len(refs) == 0:
-            return ''
-        bibtexreference = r"@ARTICLE{}".format(refs[0])
-    return bibtexreference
+        # Extract the first bibtex block:
+        ref = citation.read().partition(".. code:: bibtex\n\n")[2]
+        lines = ref.split("\n")
+        # Only read the lines which are indented
+        lines = lines[:[l.startswith("    ") for l in lines].index(False)]
+        ref = textwrap.dedent(''.join(lines))
```
dpshelio: The formating of the bibtex is completely messed up using this - all in one line. And it's only getting the APJ paper, not JOSS. Good if that's what we want. Maybe this sorts it 
```suggestion
        ref = textwrap.dedent('\n'.join(lines))
```

Cadair: I am deliberately only getting the ApJ paper.

Cadair:
> I suggest changing the title from "Acknowledging or Citing SunPy" to "Citing SunPy"

I don't really mind this, but there is also a bit on Talks and Posters at the bottom which is where the Acknowledging bit comes in I think. I have pushed a couple of other wording changes, but left the title. Let me know if you want the title changed or have other suggestions!

Cadair:
Thanks for your help everyone :smile:

PR #3805, opened 2020-02-16
Add TimeRange.contains(t) method

### LINKED ISSUE #3789: Document t in TimeRange
`TimeRange` had a `.__contains__` method, but this isn't well documented. We should add some more obvious documentation that one can do `t in TimeRange`.

dstansby:
Potentially fixes #3789

### DIFF
--- a/sunpy/time/timerange.py
+++ b/sunpy/time/timerange.py
@@ -405,7 +405,8 @@ class TimeRange:
         """
         Checks whether the given time lies within this range. Both limits are
         inclusive (i.e., ``__contains__(t1)`` and ``__contains__(t2)`` always
-        return `True).that.
+        return `True`, where ``t1, t2`` are the start and end times of the
+        range.
 
         Parameters
         ----------
@@ -414,7 +415,7 @@ class TimeRange:
 
         Returns
         -------
-        `bool`
+        bool
             `True` if time lies between start and end, `False` otherwise.
 
         Examples
@@ -431,4 +432,5 @@ class TimeRange:
         this_time = parse_time(time)
         return this_time >= self.start and this_time <= self.end
 
+    # Create TimeRange.contains, so it shows up in the docs.
     contains = __contains__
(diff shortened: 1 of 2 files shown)


nabobalis:
I wonder if we need to import `autodoc_default_options` first.

nabobalis:
To see it working nicely: https://39302-2165383-gh.circle-artifacts.com/0/docs/_build/html/api/sunpy.time.TimeRange.html?highlight=timerange

wtbarnes:
Does this need to be documented as an explicit method on `TimeSeries`? It seems like this would encourage people to use the `contains`  method directly. There's not anything wrong with that *per se*, but it seems to defeat the purpose of the `__contains__`...

Instead of adding this to the public API for the sake of documentation, could we not just add an example of how the `__contains__` method is used (i.e. the example with `this_time in that_time`) to the main `TimeSeries` docstring?

Cadair:
I am also kinda +0 on adding a new method which does the same thing. The big advantage of doing this to me is tab complete, but at the same time it feels redundant. I agree with @wtbarnes that isn't there a better way we can document this?

nabobalis:
> I am also kinda +0 on adding a new method which does the same thing. The big advantage of doing this to me is tab complete, but at the same time it feels redundant. I agree with @wtbarnes that isn't there a better way we can document this?

If you have one?

sunpy/time/timerange.py
```diff
@@ -405,7 +405,8 @@ def __contains__(self, time):
         """
         Checks whether the given time lies within this range. Both limits are
```
Cadair: We should modify this docstring to be more logical for `.contains()` and also to demonstrate the `t in tr` syntax as well.

dstansby:
For the documentation was the initial motivation, but I also think doing `TimeRange.contains(t)` is slghtly more obvious syntax than `t in TimeRange`.

Consensus seems to be to add documentation to the docstring of `TimeRange` though instead of adding `.contains()`, so I will do that and open an alternative PR.

PR #3806, opened 2020-02-16
Avoid use of .data in TimeSeries examples

dstansby:
Since using `.data` now raises a warning, fix the examples so they use best practice methods etc. Marked as no changelog needed because I think it falls under the PR that made the original change.

### DIFF
--- a/examples/time_series/power_spectra_example.py
+++ b/examples/time_series/power_spectra_example.py
@@ -26,7 +26,7 @@ ts = sunpy.timeseries.TimeSeries(RHESSI_TIMESERIES)
 
 x_ray = ts.columns[0]
 # The suitable value for fs would be 0.25 Hz as the time step is 4 s.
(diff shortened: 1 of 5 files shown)


examples/time_series/power_spectra_example.py
```diff
@@ -26,7 +26,7 @@
 
 x_ray = ts.columns[0]
 # The suitable value for fs would be 0.25 Hz as the time step is 4 s.
-freq, spectra = signal.periodogram(ts.data[x_ray], fs=0.25)
+freq, spectra = signal.periodogram(ts.quantity(x_ray), fs=0.25)
```
nabobalis: Is this just how timeseries works?

dstansby: Yes, to get a 'column' it's recommended to use `.quantity` which returns an astropy `Quantity`

nabobalis: Fancy.

PR #3808, opened 2020-02-16
Fix Matplotlib deprecation warning in animator

dstansby:
I think this is the trivial fix, but I've put it in it's own PR just to get some scrutiny. Previously `fig.add_subplot(111)` would just return the axes if it already existed in the figure, but this will no longer be the case in a later version of Matplotlib. I think the fix is to check if there is a axes present, and if not create it.

I'm unsure what "Allow replacement of main axes by subclassing." means though...

### DIFF
--- a/sunpy/visualization/animator/base.py
+++ b/sunpy/visualization/animator/base.py
@@ -251,7 +251,9 @@ class BaseFuncAnimator:
         """
         Allow replacement of main axes by subclassing.
         """
-        return self.fig.add_subplot(111)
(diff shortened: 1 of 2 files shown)


Cadair:
I think this looks ok, do any of the other animator classes need this method updating? Maybe the `WCS` animators which probably override this method.

dstansby:
I don't think it was being triggered by the tests, but I've checked and fixed the possible other instance.

nabobalis:
Thanks again @dstansby

PR #3810, opened 2020-02-17
Implemented Kanzelhohe Client 

abhijeetmanhas:
Split of PR #3763 .

### DIFF
--- a/sunpy/net/dataretriever/sources/kanzelhohe.py
+++ b/sunpy/net/dataretriever/sources/kanzelhohe.py
@@ -17,18 +17,20 @@ class KanzelhoheClient(GenericClient):
     The earliest data for H-alpha 2k - 20-Jul-2000
                           Ca-II k - 31-Jul-2010
                           Continuum - 7-Jan-2011
+
     Parameters
     ----------
-    timerange: sunpy.time.TimeRange
+    timerange: `~sunpy.time.TimeRange`
         time range for which data is to be downloaded.
         Example value - TimeRange('2015-12-30 00:00:00','2015-12-31 00:01:00')
     Instrument: Fixed argument = 'kanzelhohe'
     Wavelength: Fixed argument = astropy.units.quantity.Quantity
                 The physical value of wavelength will belong to any of [5460, 6563, 32768]
                 and units will be Angstroms.
+
     Returns
     -------
-    urls: list
+    urls: `list`
     list of urls corresponding to requested time range.
     Examples
     --------
@@ -52,55 +54,47 @@ class KanzelhoheClient(GenericClient):
     def _get_url_for_timerange(self, timerange, **kwargs):
         """
         Return list of URLS corresponding to value of input timerange.
+
         Parameters
         ----------
-        timerange: `sunpy.time.TimeRange`
+        timerange: `~sunpy.time.TimeRange`
             time range for which data is to be downloaded.
+
         Returns
         -------
-        urls : list
+        urls : `list`
             list of URLs corresponding to the requested time range
         """
         wave = kwargs['wavelength']
         wave_float = (wave.min if isinstance(wave, a.Wavelength) else wave.min()).value
-        table = {6563: ['halpha2k/recent', 'halph_fr'],
-                 32768: ['caiia', 'caiik_fi'],
-                 5460: ['phokada', 'bband_fi']}
+        table = {6563: {'datatype': ['halpha2k/recent', 'halph_fr'],
+                        'start_time': datetime.datetime(2000, 7, 20, 7, 45, 46)},
+                 32768: {'datatype': ['caiia', 'caiik_fi'],
+                         'start_time': datetime.datetime(2000, 7, 20, 7, 45, 46)},
+                 5460: {'datatype': ['phokada', 'bband_fi'],
+                        'start_time': datetime.datetime(2000, 7, 20, 7, 45, 46)}}
         # Checking if value is close enough to a wavelength value.
         # Converting from one unit to other introduces precision errors.
-        try:
-            # The isclose function in numpy can only check two arrays if they
-            # are close to each other within some precision. Standalone values such
-            # as int, doubles etc can't be checked that way. In order to do that, we put the
-            # two indiviual values in two seperate arrays da and db and then apply
-            # isclose on both those arrays.
-            da = list()
-            da.append(wave_float)
-            for wave_nums in table.keys():
-                db = list()
-                db.append(wave_nums)
-                if np.isclose(da, db, 1e-10, 1e-10):
-                    wave = wave_nums
-
-            date_table = {6563: datetime.datetime(2000, 7, 20, 7, 45, 46),
-                          5460: datetime.datetime(2011, 1, 7, 10, 7, 33),
-                          32768: datetime.datetime(2010, 7, 31, 8, 10, 59)}
-            START_DATE = date_table[wave]
-            if timerange.start < START_DATE:
-                msg = 'Earliest date for which Kanzelhohe data is available is {:%Y-%m-%d}'
-                raise ValueError(msg.format(START_DATE))
-            prefix = "http://cesar.kso.ac.at/{datatype}/%Y/"
-            suffix = ""
-            if wave != 6563:
-                suffix = "%Y%m%d/processed/"
-            url_pattern = prefix + suffix + "kanz_{datatype1}_%Y%m%d_%H%M%S.fts.gz"
-            crawler = Scraper(url_pattern, datatype=table[wave][0], datatype1=table[wave][1])
-            if not timerange:
-                return []
-            result = crawler.filelist(timerange)
-            return result
-        except BaseException:
-            raise ValueError("Enter wavelength with proper values and units")
+        wavelengths = table.keys()
+        wave_list = list(filter(lambda x: np.isclose(x, wave_float), wavelengths))
+        if(len(wave_list) == 0):
+            raise ValueError("no value is enough closer to allowed wavelengths")
+        wave = wave_list[0]
+        start_date = table[wave]['start_time']
+        if timerange.start < start_date:
+            msg = 'Earliest date for which Kanzelhohe data is available is {:%Y-%m-%d}'
+            raise ValueError(msg.format(start_date))
+        prefix = "http://cesar.kso.ac.at/{datatype}/%Y/"
+        suffix = ""
+        if wave != 6563:
+            suffix = "%Y%m%d/processed/"
+        url_pattern = prefix + suffix + "kanz_{datatype1}_%Y%m%d_%H%M%S.fts.gz"
+        dtypes = table[wave]['datatype']
+        scraper = Scraper(url_pattern, datatype=dtypes[0], datatype1=dtypes[1])
+        if not timerange:
+            return []
+        result = scraper.filelist(timerange)
+        return result
 
     def _makeimap(self):
         """
@@ -115,9 +109,11 @@ class KanzelhoheClient(GenericClient):
     def _can_handle_query(cls, *query):
         """
         Answers whether client can service the query.
+
         Parameters
         ----------
         query : list of query objects
+
         Returns
         -------
         boolean: answer as to whether client can service the query
--- a/sunpy/net/dataretriever/sources/kanzelhohe.py
+++ b/sunpy/net/dataretriever/sources/kanzelhohe.py
@@ -42,10 +42,9 @@ class KanzelhoheClient(GenericClient):
     Results from 1 Provider:
     <BLANKLINE>
     1 Results from the KanzelhoheClient:
-         Start Time           End Time      ...   Instrument      Wavelength
-           str19               str19        ...     str14           str15
-    ------------------- ------------------- ... -------------- ---------------
-    2015-12-28 00:00:00 2015-12-28 09:03:00 ... Kanzelhohe HA2 6563.0 Angstrom
+         Start Time           End Time      Source   Instrument      Wavelength
+    ------------------- ------------------- ------ -------------- ---------------
+    2015-12-28 00:00:00 2015-12-28 09:03:00    KSO Kanzelhohe HA2 6563.0 Angstrom
     <BLANKLINE>
     <BLANKLINE>
     >>> response = Fido.fetch(results)  #doctest: +SKIP
@@ -78,7 +77,8 @@ class KanzelhoheClient(GenericClient):
         wavelengths = table.keys()
         wave_list = list(filter(lambda x: np.isclose(x, wave_float), wavelengths))
         if(len(wave_list) == 0):
-            raise ValueError("no value is enough closer to allowed wavelengths")
+            raise ValueError("no value is enough closer to allowed wavelengths: "
+                             + ",".join(map(str, table.keys())) + " angstroms")
         wave = wave_list[0]
         start_date = table[wave]['start_time']
         if timerange.start < start_date:
@@ -100,10 +100,10 @@ class KanzelhoheClient(GenericClient):
         """
         Helper Function:used to hold information about source.
         """
-        self.map_['source'] = 'Global Halpha Network'  # TODO: check with kanzelhohe
+        self.map_['source'] = 'KSO'  # TODO: check with kanzelhohe
         self.map_['instrument'] = 'Kanzelhohe HA2'
         self.map_['phyobs'] = 'irradiance'
-        self.map_['provider'] = 'Kanzelhohe'
+        self.map_['provider'] = 'KSO'
 
     @classmethod
     def _can_handle_query(cls, *query):
--- a/sunpy/net/dataretriever/sources/kanzelhohe.py
+++ b/sunpy/net/dataretriever/sources/kanzelhohe.py
@@ -1,0 +1,132 @@
+import datetime
+import numpy as np
+
+import astropy.units as u
+
+from sunpy.net.dataretriever.client import GenericClient
+from sunpy.util.scraper import Scraper
+from sunpy.net import attrs as a
+
+__all__ = ['KanzelhoheClient']
+
+
+class KanzelhoheClient(GenericClient):
+    """
+    Returns a list of URLS to Kanzelhohe H-alpha files corresponding to value of input timerange.
+    URL source: `http://cesar.kso.ac.at/`.
+    The earliest data for H-alpha 2k - 20-Jul-2000
+                          Ca-II k - 31-Jul-2010
+                          Continuum - 7-Jan-2011
+    Parameters
+    ----------
+    timerange: sunpy.time.TimeRange
+        time range for which data is to be downloaded.
+        Example value - TimeRange('2015-12-30 00:00:00','2015-12-31 00:01:00')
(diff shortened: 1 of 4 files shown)


ayshih:
Copying over my two relevant comments from #3763:

> A question: should we render the observatory as "[Kanzelhoehe](https://en.wikipedia.org/wiki/Kanzelhoehe_Solar_Observatory)", or even more correctly as "[Kanzelhöhe](https://www.kso.ac.at/index_en.php)", rather than "Kanzelhohe"?
>
> One reason that I ask is that Kanzelhoehe is spelled differently in another issue than it is here, and that frustrated my initial attempt at searching.

_Originally posted by @ayshih in https://github.com/sunpy/sunpy/pull/3763#issuecomment-582457925_

> I'll point out that if this PR gets merged, we really ought to fix how Kanzelhohe/Kanzelhoehe/Kanzelhöhe files are read in by `sunpy.map` (see #1615 and https://github.com/sunpy/sunpy/issues/1615#issuecomment-542704687).

_Originally posted by @ayshih in https://github.com/sunpy/sunpy/pull/3763#issuecomment-582461245_

abhijeetmanhas:
I will change the name of the client, as per reviewers' instructions; after they all decide.

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,132 @@
...
+            # as int, doubles etc can't be checked that way. In order to do that, we put the
+            # two indiviual values in two seperate arrays da and db and then apply
+            # isclose on both those arrays.
+            da = list()
+            da.append(wave_float)
+            for wave_nums in table.keys():
+                db = list()
+                db.append(wave_nums)
+                if np.isclose(da, db, 1e-10, 1e-10):
+                    wave = wave_nums
```
dpshelio: We could get the close wavelents with:
```python
wave = list(filter(lambda x: np.isclose(x, wave_float), table.keys()))
```
or running `np.isclose` once:

```python
wave = [wave for wave, close in zip(table.keys(), np.isclose(list(table.keys()), wave_float)) if close]
```
or all in numpy:

```python
wave = np.array(list(table.keys()))[np.isclose(list(table.keys()), wave_float)]
```

Probably in all of these it would be cleaner if the keys from `table` are stored in a variable first.
Then after this an error could be raised if no value found.

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,132 @@
...
+
+            date_table = {6563: datetime.datetime(2000, 7, 20, 7, 45, 46),
+                          5460: datetime.datetime(2011, 1, 7, 10, 7, 33),
+                          32768: datetime.datetime(2010, 7, 31, 8, 10, 59)}
```
dpshelio: I would prefer to have this table merged with `table`, so there's one place where we need to update if they decide to add a new filter to their observations.

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,132 @@
...
+            START_DATE = date_table[wave]
```
dpshelio: I wouldn't use all caps variables unless they are global variables.

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,132 @@
...
+            if timerange.start < START_DATE:
+                msg = 'Earliest date for which Kanzelhohe data is available is {:%Y-%m-%d}'
+                raise ValueError(msg.format(START_DATE))
```
dpshelio: I believe this `ValueError` would be misleading by the final error message obtained from the `try` block

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,132 @@
...
+            prefix = "http://cesar.kso.ac.at/{datatype}/%Y/"
+            suffix = ""
+            if wave != 6563:
+                suffix = "%Y%m%d/processed/"
+            url_pattern = prefix + suffix + "kanz_{datatype1}_%Y%m%d_%H%M%S.fts.gz"
+            crawler = Scraper(url_pattern, datatype=table[wave][0], datatype1=table[wave][1])
```
dpshelio: I would use `scrapper` instead of `crawler`. So to keep consistency.

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,132 @@
...
+                 32768: ['caiia', 'caiik_fi'],
+                 5460: ['phokada', 'bband_fi']}
+        # Checking if value is close enough to a wavelength value.
+        # Converting from one unit to other introduces precision errors.
+        try:
+            # The isclose function in numpy can only check two arrays if they
+            # are close to each other within some precision. Standalone values such
+            # as int, doubles etc can't be checked that way. In order to do that, we put the
+            # two indiviual values in two seperate arrays da and db and then apply
+            # isclose on both those arrays.
```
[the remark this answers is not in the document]
abhijeetmanhas: maybe at the time when the PR was initially made, numpy might didn't allowed it. So now I think I should remove this comments.

sunpy/net/dataretriever/sources/kanzelhohe.py
abhijeetmanhas: replacing with `start_date`

sunpy/net/dataretriever/sources/kanzelhohe.py
abhijeetmanhas:
```python3
wavelengths = table.keys()
wave = list(filter(lambda x: np.isclose(x, wave_float), wavelengths))
if(len(wave) == 0):    #if wave is empty
    raise ValueError("no value is enough closer to allowed wavelengths")

```


Can I have something like this here?

sunpy/net/dataretriever/sources/kanzelhohe.py
abhijeetmanhas: Do we even need a try except block here? What I think is this exceptions and errors should be handled by `~sunpy.util.scraper` which obtains gets a list of url and checks whether they are existent or not.
Should I remove the try except block?

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,132 @@
...
+            da = list()
+            da.append(wave_float)
+            for wave_nums in table.keys():
+                db = list()
+                db.append(wave_nums)
+                if np.isclose(da, db, 1e-10, 1e-10):
```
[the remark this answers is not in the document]
abhijeetmanhas: Should it be fine to just use `np.isclose` without providing parameters like `1e-10` tolerance as you used in the following comment?

sunpy/net/dataretriever/sources/kanzelhohe.py
abhijeetmanhas: I didn't understood this one. Is it we should predefine it in `sunpy.database` or we should use `~sunpy.database.table` instead of python dictionary?

sunpy/net/dataretriever/sources/kanzelhohe.py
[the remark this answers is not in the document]
dpshelio: I believe the default tolerances are OK, but you could set them to something else. I was just confused why using the same number without any explanation.

sunpy/net/dataretriever/sources/kanzelhohe.py
dpshelio: That's ok. I will remove the comment `#if wave...` as it's redundant, and maybe add the valid wavelengths to the error message.

sunpy/net/dataretriever/sources/kanzelhohe.py
dpshelio: I mean, that this information about the dates could be in the same dictionary where the names are kept above. So if in the future we need to add a new instrument, we can only change one dictionary and not two.

sunpy/net/dataretriever/sources/kanzelhohe.py
dpshelio: The external one? probably you should.

abhijeetmanhas:
@dpshelio did all the suggested changes. The failing tests are from HEK(not from the changes I made), maybe go away by rerunning the tests.
Any more changes?

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,123 @@
...
+        if not timerange:
+            return []
+        result = scraper.filelist(timerange)
+        return result
+
+    def _makeimap(self):
+        """
+        Helper Function:used to hold information about source.
+        """
+        self.map_['source'] = 'Global Halpha Network'  # TODO: check with kanzelhohe
```
dpshelio: Probably, to keep it inline with VSO, we should use `KSO` as source and provider.

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,123 @@
...
+        self.map_['instrument'] = 'Kanzelhohe HA2'
```
dpshelio: This is not fully correct as there are three instruments: https://www.kso.ac.at/instrumente/instrumente_en.php#strahlung
But I vote for leave it as it is and then ask someone from Kanzelhohe to fix it.

abhijeetmanhas: @dpshelio done the suggested changes. Left this instrument issue btw.

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,128 @@
...
+                 32768: {'datatype': ['caiia', 'caiik_fi'],
+                         'start_time': datetime.datetime(2000, 7, 20, 7, 45, 46)},
+                 5460: {'datatype': ['phokada', 'bband_fi'],
+                        'start_time': datetime.datetime(2000, 7, 20, 7, 45, 46)}}
+        # Checking if value is close enough to a wavelength value.
+        # Converting from one unit to other introduces precision errors.
+        wavelengths = table.keys()
+        wave_list = list(filter(lambda x: np.isclose(x, wave_float), wavelengths))
+        if(len(wave_list) == 0):
+            raise ValueError("no value is enough closer to allowed wavelengths")
```
dpshelio:
```suggestion
            raise ValueError("no value is enough closer to allowed wavelengths: " + ",".join(table.keys()) + " angstroms")
```
Or something like that.

dpshelio: Also, could you write a test that checks this ValueError with `pytest.raises`?

abhijeetmanhas: Sure, on it, I'll write a test and make error message more meaningful.

abhijeetmanhas: @dpshelio added pytest raises for valuerror.

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,128 @@
...
+        query : list of query objects
+
+        Returns
+        -------
+        boolean: answer as to whether client can service the query
+        """
+        chk_var = 0
+        for x in query:
+            if x.__class__.__name__ == 'Instrument' and isinstance(
+                    x.value, str) and x.value.lower() == 'kanzelhohe':
```
dpshelio: To keep everyone happy and since [wikipedia](https://en.wikipedia.org/wiki/Kanzelhoehe_Solar_Observatory), their official website (or so [it says google](https://www.google.com/search?q=kanzelhoehe+site%3Awww.kso.ac.at&oq=kanzelhoehe+site%3Awww.kso.ac.at), I've not found where exactly) and [their (?) Facebook group](https://www.facebook.com/pages/Kanzelhoehe-Solar-Observatory/127081834002472) puts the ö as oe I would suggest we accept the three of them as:

```suggestion
                    x.value, str) and x.value.lower() in ['kanzelhohe', 'kanzelhoehe', 'kanzelhöhe']:
```

I don't know German, but it seems it's a common practice to substitute the ö by oe when they don't have a keyboard layout that allows to put the umlaut.

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,128 @@
...
+from sunpy.net.dataretriever.client import GenericClient
+from sunpy.util.scraper import Scraper
+from sunpy.net import attrs as a
+
+__all__ = ['KanzelhoheClient']
+
+
+class KanzelhoheClient(GenericClient):
+    """
+    Returns a list of URLS to Kanzelhohe H-alpha files corresponding to value of input timerange.
```
dpshelio:
```suggestion
    Returns a list of URLS to Kanzelhöhe H-alpha files corresponding to value of input timerange.
```
We should be able to put ö on docstrings.

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,128 @@
...
+import astropy.units as u
+
+from sunpy.net.dataretriever.client import GenericClient
+from sunpy.util.scraper import Scraper
+from sunpy.net import attrs as a
+
+__all__ = ['KanzelhoheClient']
+
+
+class KanzelhoheClient(GenericClient):
```
dpshelio: This is more tricky... if we don't put `ö` on the class name, should we put `oe`? 

@segfaulthunter what would you say?

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,128 @@
...
+                        'start_time': datetime.datetime(2000, 7, 20, 7, 45, 46)},
+                 32768: {'datatype': ['caiia', 'caiik_fi'],
+                         'start_time': datetime.datetime(2000, 7, 20, 7, 45, 46)},
+                 5460: {'datatype': ['phokada', 'bband_fi'],
+                        'start_time': datetime.datetime(2000, 7, 20, 7, 45, 46)}}
+        # Checking if value is close enough to a wavelength value.
+        # Converting from one unit to other introduces precision errors.
+        wavelengths = table.keys()
+        wave_list = list(filter(lambda x: np.isclose(x, wave_float), wavelengths))
+        if(len(wave_list) == 0):
```
dpshelio:
```suggestion
        if len(wave_list) == 0:
```
No need to put the parentheses.

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,128 @@
...
+            raise ValueError("no value is enough closer to allowed wavelengths: "
```
dpshelio: I'm not a native English speaker, but this sounds better to me:
```suggestion
            raise ValueError("no value is closer enough to available wavelengths: "
```
Feel free to ignore it if you think I'm wrong. 
However, I'm changing `allowed` to `available` as this feels more precise to what we want to tell the users.

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,128 @@
...
+                             + ",".join(map(str, table.keys())) + " angstroms")
+        wave = wave_list[0]
+        start_date = table[wave]['start_time']
+        if timerange.start < start_date:
+            msg = 'Earliest date for which Kanzelhohe data is available is {:%Y-%m-%d}'
```
dpshelio:
```suggestion
            msg = 'Earliest date for which Kanzelhöhe data is available is {:%Y-%m-%d}'
```

sunpy/net/dataretriever/sources/kanzelhohe.py
```diff
@@ -0,0 +1,128 @@
...
+            return []
+        result = scraper.filelist(timerange)
+        return result
+
+    def _makeimap(self):
+        """
+        Helper Function:used to hold information about source.
+        """
+        self.map_['source'] = 'KSO'  # TODO: check with kanzelhohe
+        self.map_['instrument'] = 'Kanzelhohe HA2'
```
dpshelio:
```suggestion
        self.map_['instrument'] = 'Kanzelhöhe HA2'
```

abhijeetmanhas: Should I remove todo in Line 103?

dpshelio: I wouldn't. In any case, once we merge it we may create an issue about it too.

abhijeetmanhas:
@dpshelio @nabobalis for naming, can we just have `KANZ` since VSO uses it? And I checked here http://vso3.stanford.edu/cgi-bin/show_details?instrument=KANZ vso covers only one wavelength and this client covers two other wavelengths too.

nabobalis:
You mean call it the KANZClinet?

abhijeetmanhas:
> You mean call it the KANZClinet?

yep `KANZClient`. If it sounds good. Just making initial name same as VSO. We can also have `Kanzelhoehe' but it will confuse some users with ö.

nabobalis:
Sure.

sunpy/net/dataretriever/tests/test_kanzelhohe.py
```diff
@@ -0,0 +1,75 @@
...
+
+from astropy import units as u
+from sunpy.time.timerange import TimeRange
+from sunpy.net import attrs as a
+from sunpy.net.dataretriever.client import QueryResponse
+from sunpy.net.fido_factory import UnifiedResponse
+from sunpy.net import Fido
+import sunpy.net.dataretriever.sources.kanzelhohe as kanzelhohe
+
+KClient = kanzelhohe.KANZClient()
```
nabobalis, line 12: This should be turned into a pytest fixture.

sunpy/net/dataretriever/tests/test_kanzelhohe.py
```diff
@@ -0,0 +1,75 @@
...
+    assert isinstance(qr, QueryResponse)
+    assert len(qr) == 2
+    assert qr.time_range().start == datetime.datetime(2015, 1, 10)
+    assert qr.time_range().end == datetime.datetime(2015, 1, 10, 12, 0, 0)
+    with pytest.raises(ValueError, match="no value is closer enough to the available wavelengths: "
+                                         "6563,32768,5460 angstroms"):
+        KClient.search(TRANGE, a.Wavelength(6000 * u.AA))
+
+
+# This test downloads 3 files
```
nabobalis, line 53: Is it possible to download 1 file?

sunpy/net/dataretriever/tests/test_kanzelhohe.py
abhijeetmanhas: Done

sunpy/net/dataretriever/tests/test_kanzelhohe.py
abhijeetmanhas: Fixed.

PR #3812, opened 2020-02-17
Implemented Bbso Client

abhijeetmanhas:
Split of PR #3763

### DIFF
--- a/sunpy/util/scraper.py
+++ b/sunpy/util/scraper.py
@@ -29,20 +29,17 @@ TIME_CONVERSIONS = {'%Y': r'\d{4}', '%y': r'\d{2}',
 class Scraper:
     """
     A Scraper to scrap web data archives based on dates.
-
     Parameters
     ----------
     pattern : `str`
         A string containing the url with the date encoded as datetime formats,
         and any other parameter as ``kwargs`` as a string format.
-
     Attributes
     ----------
     pattern : `str`
         A converted string with the kwargs.
     now : `datetime.datetime`
         The pattern with the actual date.
-
     Examples
     --------
     >>> # Downloading data from SolarMonitor.org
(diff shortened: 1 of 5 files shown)


sunpy/util/scraper.py
```diff
@@ -231,8 +232,11 @@ def filelist(self, timerange):
...
                     for link in soup.find_all("a"):
                         href = link.get("href")
-                        if href.endswith(self.pattern.split('.')[-1]):
-                            fullpath = directory + href
+                        if href is not None and href.endswith(self.pattern.split('.')[-1]):
+                            if href[0] == '/':
+                                fullpath = self.domain + href[1:]
+                            else:
+                                fullpath = directory + href
                             if self._URL_followsPattern(fullpath):
```
nabobalis, line 240: Are these changes are bug fix or just specific to this client?

abhijeetmanhas: It is a bugfix. Luckily in all our previous clients, we don't check whether `href` can be `None`; because those html pages opened by `BeautifulSoup` never had one.

Another fix is about relative paths of file returned by href, the directory opened by scraper was like `https://some_domain/pub/keep/2016/05/18` and href returned something like `/pub/keep/2016/05/18/file_nameHHMMSS.fts` so full url becomes `https://some_domain/pub/keep/2016/05/18/pub/keep/2016/05/18/file_nameHHMMSS.fts` which is wrong, and it should be just `domain + href[1:]`. Checking whether file href starts from `\` to know href are relative to just domain, not the directory.

nabobalis: Could this be spun off into its own PR?

abhijeetmanhas: Ok

PR #3822, opened 2020-02-19
Compatibility with parfive 1.1

### LINKED ISSUE #3844: In the jsoc client, for the fetch function, `max_conn` is listed as a parameter in the documentation but its not a valid parameter.
### Description
The title is pretty self-explanatory. This is at line *397* in file `~sunpy/net/jsoc/jsoc.py`
The function snippet below contains the bug. Either `max_conn` be removed from the docstring or should be implemented. 
```
    def fetch(self, jsoc_response, path=None, progress=True, overwrite=False,
              downloader=None, wait=True, sleep=10):
        """
        Make the request for the data in a JSOC response and wait for it to be
        staged and then download the data.

        Parameters
        ----------
        jsoc_response : `~sunpy.net.jsoc.jsoc.JSOCResponse` object
            A response object

        path : `str`
            Path to save data to, defaults to SunPy download dir

        progress : `bool`, optional
            If `True` show a progress bar showing how many of the total files
            have been downloaded. If `False`, no progress bar will be shown.

        overwrite : `bool` or `str`, optional
            Determine how to handle downloading if a file already exists with the
            same name. If `False` the file download will be skipped and the path
            returned to the existing file, if `True` the file will be downloaded
            and the existing file will be overwritten, if `'unique'` the filename
            will be modified to be unique.

        max_conns : `int`
            Maximum number of download connections.

        downloader : `parfive.Downloader`, optional
            The download manager to use.

        wait : `bool`, optional
           If `False` ``downloader.download()`` will not be called. Only has
           any effect if `downloader` is not `None`.

        sleep : `int`
            The number of seconds to wait between calls to JSOC to check the status
            of the request.

        Returns
        -------
        results : a `~sunpy.net.download.Results` instance
            A Results object

        """
        # Make staging request to JSOC
        responses = self.request_data(jsoc_response)

        # Make response iterable
        if not isiterable(responses):
            responses = [responses]

        # Add them to the response for good measure
        jsoc_response.requests = [r for r in responses]
        time.sleep(sleep/2.)

        for response in responses:
            response.wait(verbose=progress)

        return self.get_request(responses, path=path, overwrite=overwrite,
                                progress=progress, downloader=downloader,
                                wait=wait)

```

Raahul-Singh:
Created to verify the new parfive release against sunpy's CI.

(Also fixes #3844 as a side effect)

### DIFF
--- a/changelog/3822.feature.rst
+++ b/changelog/3822.feature.rst
@@ -1,0 +1,1 @@
+Add support for parfive 1.1. This sets a limit on the number of open connections to JSOC when downloading files to 10.
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,8 +38,8 @@ install_requires =
   scipy>=1.0.0
   matplotlib>=2.1.2
   pandas>=0.23.0
-  astropy>=3.2
-  parfive[ftp]>=1.0
+  astropy>=3.2,!=4.0.1,!=4.0.1.post1
+  parfive[ftp]>=1.0.2
   importlib_resources;python_version<"3.7"
 
 [options.extras_require]
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,7 @@ install_requires =
   scipy>=1.0.0
   matplotlib>=2.1.2
   pandas>=0.23.0
-  astropy>=3.2,!=4.0.1,!=4.0.1.post1
+  astropy>=3.2
   parfive[ftp]>=1.0.2
   importlib_resources;python_version<"3.7"
 
--- a/sunpy/net/dataretriever/sources/norh.py
+++ b/sunpy/net/dataretriever/sources/norh.py
@@ -89,7 +89,7 @@ class NoRHClient(GenericClient):
             timerange = TimeRange(timerange.start.strftime('%Y-%m-%d'),
                                   timerange.end)
 
-        norh = Scraper(BASEURL, freq=freq)
+        norh = Scraper(BASEURL, freq=freq, append_login=False)
         # TODO: warn user that some files may have not been listed, like for example:
         #       tca160504_224657 on ftp://solar-pub.nao.ac.jp/pub/nsro/norh/data/tcx/2016/05/
         #       as it doesn't follow pattern.
@@ -98,7 +98,7 @@ class NoRHClient(GenericClient):
 
     def _get_time_for_url(self, urls):
         freq = urls[0].split('/')[-1][0:3]  # extract the frequency label
-        crawler = Scraper(BASEURL, freq=freq)
+        crawler = Scraper(BASEURL, freq=freq, append_login=False)
         times = list()
         for url in urls:
             t0 = crawler._extractDateURL(url)
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -453,7 +453,7 @@ class JSOCClient(BaseClient):
         return requests
 
     def fetch(self, jsoc_response, path=None, progress=True, overwrite=False,
-              downloader=None, wait=True, sleep=10):
+              downloader=None, wait=True, sleep=10, max_conn=4, **kwargs):
         """
         Make the request for the data in a JSOC response and wait for it to be
         staged and then download the data.
@@ -477,7 +477,7 @@ class JSOCClient(BaseClient):
             and the existing file will be overwritten, if `'unique'` the filename
             will be modified to be unique.
 
-        max_conns : `int`
+        max_conn : `int`
             Maximum number of download connections.
 
         downloader : `parfive.Downloader`, optional
@@ -500,6 +500,9 @@ class JSOCClient(BaseClient):
         # Make staging request to JSOC
         responses = self.request_data(jsoc_response)
 
+        defaults = {'max_splits': 2}
+        defaults.update(kwargs)
+
         # Make response iterable
         if not isiterable(responses):
             responses = [responses]
@@ -513,10 +516,10 @@ class JSOCClient(BaseClient):
 
         return self.get_request(responses, path=path, overwrite=overwrite,
                                 progress=progress, downloader=downloader,
-                                wait=wait)
+                                wait=wait, max_conn=max_conn, **defaults)
 
     def get_request(self, requests, path=None, overwrite=False, progress=True,
-                    downloader=None, wait=True):
+                    downloader=None, wait=True, max_conn=4, **kwargs):
         """
         Query JSOC to see if the request(s) is ready for download.
 
@@ -557,6 +560,8 @@ class JSOCClient(BaseClient):
         """
         c = drms.Client()
 
+        kwargs['max_splits'] = kwargs.get('max_splits', 2)
+
         # Convert Responses to a list if not already
         if isinstance(requests, str) or not isiterable(requests):
             requests = [requests]
@@ -595,10 +600,18 @@ class JSOCClient(BaseClient):
                 fname = os.path.expanduser(fname)
                 paths.append(fname)
 
+        if max_conn * kwargs['max_splits'] > 10:
+            warnings.warn(("JSOC does not support more than 10 parallel connections. " +
+                           "Changing the number of parallel connections to 8."), SunpyUserWarning)
+            kwargs['max_splits'] = 2
+            max_conn = 4
+
         dl_set = True
         if not downloader:
             dl_set = False
-            downloader = Downloader(progress=progress, overwrite=overwrite)
+            downloader = Downloader(progress=progress, overwrite=overwrite, max_conn=max_conn)
+        else:
+            downloader.max_conn = max_conn
 
         urls = []
         for request in requests:
@@ -615,7 +628,7 @@ class JSOCClient(BaseClient):
                 print_message = "{0} URLs found for download. Full request totalling {1}MB"
                 print(print_message.format(len(urls), request._d['size']))
             for aurl, fname in zip(urls, paths):
-                downloader.enqueue_file(aurl, filename=fname)
+                downloader.enqueue_file(aurl, filename=fname, **kwargs)
 
         if dl_set and not wait:
             return Results()
--- a/sunpy/net/jsoc/tests/test_jsoc.py
+++ b/sunpy/net/jsoc/tests/test_jsoc.py
@@ -7,9 +7,11 @@ import astropy.time
 import astropy.units as u
 import pytest
 from parfive import Results
+from unittest import mock
 
 from sunpy.net.jsoc import JSOCClient, JSOCResponse
 import sunpy.net.attrs as a
+from sunpy.util.exceptions import SunpyUserWarning
 
 client = JSOCClient()
 
@@ -346,3 +348,25 @@ def test_can_handle_query_no_series():
     assert not JSOCClient._can_handle_query(a.Time("2020/01/02", "2020/01/03"))
     assert not JSOCClient._can_handle_query(a.Wavelength(17.1*u.nm))
     assert JSOCClient._can_handle_query(a.jsoc.Series("hmi.M_45s"))
+
+
+@pytest.mark.remote_data
+def test_max_parallel_connections():
+    responses = client.search(
+        a.Time('2014/1/1T1:00:36', '2014/1/1T01:01:38'),
+        a.jsoc.Series('hmi.M_45s'), a.jsoc.Notify('jsoc@cadair.com'),
+        a.jsoc.Protocol("as-is"))
+
+    path = tempfile.mkdtemp()
+
+    with mock.patch(
+                    "parfive.Downloader.download",
+                    new_callable=mock.MagicMock
+                  ) as download:
+
+        download.side_effect = ["Mocked Downloader"]
+
+        with pytest.warns(SunpyUserWarning):
+            client.fetch(responses, path=path, max_conn=5, max_splits=5)
+
+    assert download.called
--- a/sunpy/util/scraper.py
+++ b/sunpy/util/scraper.py
@@ -75,6 +75,10 @@ class Scraper:
                 end=self.pattern[milliseconds.end():]
             ))
 
+        self.append_login = True
+        if 'append_login' in kwargs:
+            self.append_login = kwargs['append_login']
+
     def matches(self, filepath, date):
         return date.strftime(self.pattern) == filepath
 
@@ -263,8 +267,13 @@ class Scraper:
                         if (datehref >= timerange.start and
                                 datehref <= timerange.end):
                             filesurls.append(fullpath)
-        filesurls = ['ftp://anonymous:data@sunpy.org@' + "{0.netloc}{0.path}".format(urlsplit(url))
+
+        login = ''
+        if self.append_login:
+            login = "anonymous:data@sunpy.org@"
+        filesurls = [f'ftp://{login}' + "{0.netloc}{0.path}".format(urlsplit(url))
                      for url in filesurls]
+
         return filesurls
 
     def _localfilelist(self, timerange):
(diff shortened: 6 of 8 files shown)


Raahul-Singh:
I find it curious that python3.8 tests should fail. I read somewhere that they are changing the way `mock` is used in 3.8. Could that be the reason?

Cadair:
It's not python 3.8, the only online test build happens to be the python 3.8 one.

Cadair:
There looks to be a real issue with the example gallery and one of the online test issues look real (the SUVI tests are a known issue).

Raahul-Singh:
You were right @Cadair setting the `max_splits = 1` solved the `JSOC` errors. `NORH` error remain now.
In parfive, `max_splits` is set to be 12 by default.

Cadair:
I am chatting to the people who maintain JSOC, we might want to reduce the default from 12 in parfive (12*5 in parfive leads to a default of 60 connections, which seems like a lot!). However, should we also consider if we can override the default reliably in the JSOC client?

Cadair:
Conclusion with the JSOC folk is we probably shouldn't be hammering them with a lot of open connections, I propose that the `JSOCClient` defaults `max_splits` to two, and keeps `max_conn` at 5.

Raahul-Singh:
> I am chatting to the people who maintain JSOC, we might want to reduce the default from 12 in parfive (12*5 in parfive leads to a default of 60 connections, which seems like a lot!). However, should we also consider if we can override the default reliably in the JSOC client?

That is what I did here. Set the default to 1 and the tests worked. For this, we can simply implement `max_conns` in `JSOCClient.fetch` and pass it to `JSOCClient.get_request` as a kwarg.

Raahul-Singh:
> Conclusion with the JSOC folk is we probably shouldn't be hammering them with a lot of open connections, I propose that the `JSOCClient` defaults `max_splits` to two, and keeps `max_conn` at 5.

Doesn't `max_splits` override `max_conn` ?

Cadair:
> Doesn't `max_splits` override `max_conn` ?

In a series of poor naming choices, `max_conn` is the number of parallel files that will be downloaded and `max_splits` is the number of parts used for each file.

We should override the max_splits default in JSOCClient but allow users to override it I would think.

Raahul-Singh:
One of the two `NORH` test now works. The other should have passed as well. Oh, and `NOAA` test felt lonely and decided it will fail all of a sudden for no reason whatsoever.

Cadair:
Latest intel on JSOC is that the max connection limit is 10 per IP. So I suggest we go for `max_splits=2 max_conn=4` so we run at 8 not the max of 10.

Raahul-Singh:
I've applied your suggestions for JSOC. NORH fails despite the changes

Cadair:
I wasn't really suggesting it would be a fix for NORH. 

I think we should allow the user, through the  `JSOCClient.fetch()` method, to override these keyword argument and all the other arguments to `enqueue_file` via `**kwargs`.

A pattern like:
```python
def func(**kwargs):
    defaults = {'max_conn': 4}
    defaults.update(kwargs)
    enqueue_file(**defaults)
```
(all be it with better variable names) works well for this situation.

sunpy/net/jsoc/jsoc.py
```diff
@@ -579,8 +582,8 @@ def get_request(self, requests, path=None, overwrite=False, progress=True,
         dl_set = True
         if not downloader:
             dl_set = False
-            downloader = Downloader(progress=progress, overwrite=overwrite)
-
+            downloader = Downloader(progress=progress, overwrite=overwrite, max_conn=4)
```
Cadair: max_conn should be treated the same as `max_splits`.

sunpy/net/jsoc/jsoc.py
```diff
@@ -500,7 +500,7 @@ def fetch(self, jsoc_response, path=None, progress=True, overwrite=False,
                                 wait=wait)
```
Cadair: you need to pass `**kwargs` through here (and in the call signature to `fetch`).

sunpy/net/jsoc/jsoc.py
Raahul-Singh: max_conn goes to Downloader constructor. I don't think it can be passed as a kwarg. Can it be?

Cadair: it would have to be explicit (and not in `**kwargs`) I guess.

Raahul-Singh:
> I wasn't really suggesting it would be a fix for NORH.

Yeah I know. Was just wondering if I'll get lucky and it'll fix this too. :grin: 

> I think we should allow the user, through the `JSOCClient.fetch()` method, to override these keyword arguments and all the other arguments to `enqueue_file` via `**kwargs`.
> 
> A pattern like:
> 
> ```python
> def func(**kwargs):
>     defaults = {'max_conn': 4}
>     defaults.update(kwargs)
>     enqueue_file(**defaults)
> ```
> 
> (all be it with better variable names) works well for this situation.

I have applied the changes. However, the change of the variable names would need to be done first in parfive.

Cadair:
>  the change of the variable names

oh I was mostly referring to `defaults` lol

Raahul-Singh:
The `NORH` test fails with the following message :
`Waiting for ('230', '33x') but got 530 [" Can't change from guest user."]`

Raahul-Singh:
The `NORH` test fails with the following message :
`Waiting for ('230', '33x') but got 530 [" Can't change from guest user."]`
But even with parfive==1.0.0, it doesn't download the files and returns an empty list.

sunpy/net/jsoc/jsoc.py
```diff
@@ -582,7 +582,7 @@ def get_request(self, requests, path=None, overwrite=False, progress=True,
         dl_set = True
         if not downloader:
```
Cadair: So the issue here is that this if statement will not be hit from `Fido.fetch` as we pass in a `Downloader` instance.

Raahul-Singh: `max_conn` is an instance variable.
My OOPS Instructor would shake his head in disapproval if I modify an instance variable without getters and setters after an object has been created. Should we implement that as a property in parfive first?

Cadair: I am not sure it's worth it. Interface control is great and all, but what's a property in Python doing if not just changing an instance variable underneath :grinning: 

The other thing we could do here is just disable multipart altogether for JSOC?

Raahul-Singh: Which one would you prefer? 
I for one think that we should change the instance variable here for now and implement checks to see if the values passed by the users exceed 8 connections. Raise a warning and reduce the connections to our default values.

Cadair: Sounds good but let's warn on >10 and default to 8. (Given 10 is the max allowed).

nabobalis: If the limit is 10, do we want to be that close to it by default?

Cadair: Well you would have to be running two separate python processes or other downloads from JSOC to go over it. We want to get people their data as quickly as we can imo. We should obviously document this so if it goes wrong people have an idea why.

sunpy/net/jsoc/tests/test_jsoc.py
```diff
@@ -318,3 +319,14 @@ def test_results_filenames():
     assert len(files) == len(responses)
     for hmiurl in files:
         assert os.path.isfile(hmiurl)
+
+
+@pytest.mark.remote_data
```
Cadair: Rather than making this an online test, you should mock `parfive.download` as we don't actually need to download the files.

Raahul-Singh: How should I mock the JSOCResponse object that goes into this,
` client.fetch(responses, path=path, max_conn=5, max_splits=5)`
Hard code one ?

Cadair: yeah, if it's not too much effort you might as well. Even if you end up keeping the remote_data for the search, it's worth not actually doing the download.

sunpy/net/jsoc/tests/test_jsoc.py
```diff
@@ -320,6 +322,27 @@ def test_results_filenames():
...
+
+    with mock.patch(
+                    "parfive.Downloader.download",
+                    new_callable=mock.MagicMock
+                  ) as download:
+
+        download.side_effect = ["Mocked Downloader"]
+
+        with pytest.warns(SunpyUserWarning):
+            client.fetch(responses, path=path, max_conn=5, max_splits=5)
```
Cadair: We need to think about this a little more, as the images are still being staged for export here which is something we should try and avoid.

Cadair: I guess we could use `as-is` as the export method and then it wouldn't stage them, that might be the easiest way.

Raahul-Singh: Sorry, it took three days. I was kinda out of action.

Cadair: No worries!

Raahul-Singh:
Any idea why the database tests would be failing?

Raahul-Singh:
The  `test_get_request` test fails because I am adding  `max_splits` through `fetch` and not through `get_request`.
Should I change the test or add a default `max_splits` to `get_request` as well?
Perhaps check the `kwargs` before using `max_splits` ?

sunpy/net/jsoc/jsoc.py
```diff
@@ -547,6 +547,9 @@ def get_request(self, requests, path=None, overwrite=False, progress=True,
         """
         c = drms.Client()
 
+        if 'max_splits' not in kwargs:
+            kwargs['max_splits'] = 2
```
dpshelio: You could do:
```suggestion
        kwargs['max_splits'] = kwargs.get('max_splits', 2)
```

Raahul-Singh:
NORH test fail once again

Cadair:
I don't know why those tests are failing, I suggest you get in there with the debugger (run pytest / tox with `--pdb`) and see what's going on :grin:

sunpy/net/dataretriever/sources/norh.py
```diff
@@ -90,7 +90,7 @@ def _get_url_for_timerange(self, timerange, **kwargs):
             timerange = TimeRange(timerange.start.strftime('%Y-%m-%d'),
                                   timerange.end)
 
-        norh = Scraper(BASEURL, freq=freq)
+        norh = Scraper(BASEURL, freq=freq, appendLogin=False)
```
dpshelio: variables are preferred to use snake_case rather than camelCase.

sunpy/util/scraper.py
```diff
@@ -262,7 +266,12 @@ def _ftpfileslist(self, timerange):
                         if (datehref >= timerange.start and
                                 datehref <= timerange.end):
                             filesurls.append(fullpath)
-        filesurls = ['ftp://anonymous:data@sunpy.org@' + url[domain + 2:] for url in filesurls]
+
+        if self.appendLogin:
+            filesurls = ['ftp://anonymous:data@sunpy.org@' + url[domain + 2:] for url in filesurls]
+        else:
+            filesurls = ['ftp://' + url[domain + 2:] for url in filesurls]
```
dpshelio: Let's make this a bit DRY-er
```suggestion
        login = ""
        if self.appendLogin:
            login = "anonymous:data@sunpy.org@"
        filesurls = [f'ftp://{login}' + url[domain + 2:] for url in filesurls]
```

sunpy/util/scraper.py
```diff
@@ -74,9 +74,9 @@ def __init__(self, pattern, **kwargs):
                 end=self.pattern[milliseconds.end():]
             ))
 
-        self.appendLogin = True
-        if 'appendLogin' in kwargs:
-            self.appendLogin = kwargs['appendLogin']
+        self.append_login = True
+        if 'append_login' in kwargs:
```
Cadair, line 78: why not add this as an explict kwarg to the signature?

Raahul-Singh: Right now, there was this one isolated case where we needed the scraper to not append the login details to the FTP URL. I thought it'll be better if we added this as a kwarg rather than make changes to the function prototype.

Cadair: tbh I am a little confused by why we are doing this at all. Why can't we append the login in the things that call the scraper that need it?

nabobalis:
We will need to backport parts of this to 1.1 and 1.0 is my gut feeling or at least pinning older versions to use <1.1 of parfive.

Cadair:
yeah that's true, we definitely need to do one of those two things.

sunpy/util/scraper.py
Raahul-Singh: Right now, every FTP URL that called the scraper was being appended with the login. If  you think that it'll be better to let the URLs decide if they need the login appended, then I believe that passing append_login as an explicit kwarg would be better. May I make it so?

Cadair: I think if the URLs need the login appended it should be appended by the caller before it gets to scraper, I don't understand why this code is in scraper at all lol. @dpshelio ?

changelog/3822.feature.rst
```diff
@@ -0,0 +1 @@
+Integrating `parfive==1.1rc2` with sunpy.
```
Cadair:
```suggestion
Add support for parfive 1.1. This sets a limit on the number of open connections to JSOC when downloading files to 10.
```

setup.cfg
```diff
@@ -38,8 +38,8 @@ install_requires =
   scipy>=1.0.0
   matplotlib>=2.1.2
   pandas>=0.23.0
+  parfive[ftp]>=1.1rc2
```
Cadair, line 41: we should remove this before we merge once the CI has passed.

changelog/3822.feature.rst
Raahul-Singh: I'll do this along with the change suggested below. I was resolving a merge conflict and now can't commit this suggestion. :man_facepalming:

Cadair:
@Raahul-Singh can you rebase this again to fix the doc build please?

Raahul-Singh:
@Cadair, Nabobalis helped me and it looks like it tried to download the sample data and failed
is the sample data ok?

setup.cfg
Raahul-Singh: I have set it back to the >= 1.0 as 1.1 has not been released as of yet.

changelog/3822.feature.rst
```diff
@@ -0,0 +1 @@
+Add support for parfive 1.1. This sets a limit on the number of open connections to JSOC when downloading files to 10.
```
nabobalis, line 1: I thought it was 8?

nabobalis:
The doc build has downloaded the example data so far, have to see if the example fails again.

nabobalis:
The doc build passed. It must have been trainsist.

changelog/3822.feature.rst
Cadair: We always set it to 8 but we allow 10 if explicitly passed by the user.

nabobalis: Oh i see.

Raahul-Singh:
@Cadair These failed jsoc tests are because I switched the parfive version from 1.1rc2 to >=1.0 ?

Cadair:
Well we definitely shouldn't have broken support for 1.0 :worried:

Raahul-Singh:
> Well we definitely shouldn't have broken support for 1.0

Most likely because of the new kwargs that we've added to JSOC and are passing on parfive, I believe.

setup.cfg
```diff
@@ -38,8 +38,8 @@ install_requires =
   scipy>=1.0.0
   matplotlib>=2.1.2
   pandas>=0.23.0
-  astropy>=3.2
-  parfive[ftp]>=1.0
+  astropy>=3.2,!=4.0.1,!=4.0.1.post1
```
Cadair, line 41:
```suggestion
  astropy>=3.2
```

nabobalis:
Devdeps failed on downloading a file in an example.

nabobalis:
Thanks again @Raahul-Singh

PR #3829, opened 2020-02-21
Document types of jsoc.attrs.PrimeKey

dstansby:
I was trying to pass an `int` to a `PrimeKey`, which I don't think you're meant to do; according to https://docs.sunpy.org/en/stable/guide/acquiring_data/jsoc.html everything should be a string, so document this.

### DIFF
--- a/sunpy/net/jsoc/attrs.py
+++ b/sunpy/net/jsoc/attrs.py
@@ -25,6 +25,11 @@ class Keys(SimpleAttr):
 class PrimeKey(DataAttr):
     """
     Prime Keys
+
+    Parameters
+    ----------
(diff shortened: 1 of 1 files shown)


nabobalis:
This brings up a question in general that springs from my ignorance. 

For classes, should parameters be documented on the main class docstring or the classes init method?

dstansby:
The numpydoc spec says to put them in the main class docstring: https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring

nabobalis:
Thanks again @dstansby.

`net` needs a full doc reviews. I might have to get back to that.

PR #3838, opened 2020-02-23
fix issue with 'as-is' protocol

### LINKED ISSUE #2631: JSOC Client errors with `Protocol("as-is")`
```python
query = a.jsoc.Time("2014-12-18T18:00:00", "2014-12-18T18:00:00") & a.jsoc.Series("hmi.m_720s") & a.jsoc.Notify(email)
res = Fido.search(query)
files = Fido.fetch(res)
```
```
  File "plot_jsoc_map_asis_sunpy", line 20, in <module>
    files = Fido.fetch(res)
  File "/home/stuart/Git/sunpy/sunpy/net/fido_factory.py", line 362, in fetch
    reslist.append(block.client.fetch(block, **kwargs))
  File "/home/stuart/Git/sunpy/sunpy/net/jsoc/jsoc.py", line 547, in fetch
    progress=progress, results=r)
  File "/home/stuart/Git/sunpy/sunpy/net/jsoc/jsoc.py", line 651, in get_request
    url_dir = request.request_url + '/'
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

sashank27:
### Description

Fixes #2631


sunpy/net/jsoc/tests/test_jsoc.py
```diff
@@ -119,6 +119,18 @@ def test_wait_get():
     assert len(res) == 1
 
 
+@pytest.mark.remote_data
+def test_wait_get_as_is():
+    responses = client.search(
+        a.Time('2012/1/1T1:00:36', '2012/1/1T01:00:38'),
+        a.jsoc.Series('hmi.M_45s'), a.jsoc.Notify('jsoc@cadair.com'),
```
dstansby: Is @Cadair okay with having this email address here?

nabobalis: I think we have a test email address for sunpy somewhere in this file.

sashank27: Everywhere in the file, the same email is present.
example: https://github.com/sunpy/sunpy/blob/11a0d65fc7ecfe9d85210603521e67a02cf5a7fe/sunpy/net/jsoc/tests/test_jsoc.py#L56
https://github.com/sunpy/sunpy/blob/11a0d65fc7ecfe9d85210603521e67a02cf5a7fe/sunpy/net/jsoc/tests/test_jsoc.py#L67

nabobalis: Ah that is the test email. I thought there was a sunpy@sunpy one.

Cadair: Now I have the ability to create `@sunpy.org` emails, I should probably put changing this on my todo. I did, however, create this email alias for this purpose (and then deleted it as soon as I had completed the registration).

sunpy/net/jsoc/tests/test_jsoc.py
```diff
@@ -318,3 +330,17 @@ def test_results_filenames():
...
         assert os.path.isfile(hmiurl)
+
+
+@pytest.mark.remote_data
+def test_results_filenames_as_is():
+    responses = client.search(
+        a.Time('2014/1/1T1:00:36', '2014/1/1T01:01:38'),
+        a.jsoc.Series('hmi.M_45s'), a.jsoc.Notify('jsoc@cadair.com'),
+        a.jsoc.Protocol('as-is'))
+    path = tempfile.mkdtemp()
```
Cadair: rather than using `tempfile` like this it would be better to use the [`tmp_path`](https://docs.pytest.org/en/latest/tmpdir.html#the-tmp-path-fixture) pytest fixture.

sunpy/net/jsoc/tests/test_jsoc.py
```diff
@@ -119,6 +119,18 @@ def test_wait_get():
     assert len(res) == 1
 
 
+@pytest.mark.remote_data
+def test_wait_get_as_is():
+    responses = client.search(
+        a.Time('2012/1/1T1:00:36', '2012/1/1T01:00:38'),
+        a.jsoc.Series('hmi.M_45s'), a.jsoc.Notify('jsoc@cadair.com'),
+        a.jsoc.Protocol('as-is'))
+    path = tempfile.mkdtemp()
```
Cadair: same here about tempfile.

sunpy/net/jsoc/tests/test_jsoc.py
```diff
@@ -318,3 +330,17 @@ def test_results_filenames():
     assert len(files) == len(responses)
     for hmiurl in files:
         assert os.path.isfile(hmiurl)
+
+
+@pytest.mark.remote_data
+def test_results_filenames_as_is():
+    responses = client.search(
+        a.Time('2014/1/1T1:00:36', '2014/1/1T01:01:38'),
```
Cadair: How many files is this downloading? Also I can't see a significant difference between this and the other new test above, what are they testing?

In general tests which download data are slow and prone to failure, if we can adequately test this with a single download then we should.

sashank27: The default protocol is `fits`. In the newly added tests, the protocol is explicitly specified as `as-is` so as to check if the edits work correctly or not

Cadair: What I was trying to ask was you have two tests which both use `as-is` and to me they seem very similar, what's the difference between them?

sashank27: The first one has a single result, while the second one has 2.
Yeah, I believe we can remove the former test

changelog/3838.bugfix.1.rst
```diff
@@ -0,0 +1 @@
+Add support for passing paths of type `pathlib.Path` in `~sunpy.net.jsoc.JSOCClient`
```
Cadair:
```suggestion
Add support for passing paths of type `pathlib.Path` in `sunpy.net.jsoc.JSOCClient.fetch`.
```

changelog/3838.bugfix.rst
```diff
@@ -0,0 +1 @@
+Add explicit support for dealing with download urls for files, under 'as-is' protocol in `sunpy.net.jsoc.JSOCClient.get_request`
```
Cadair:
```suggestion
Add explicit support for dealing with download urls for files, under 'as-is' protocol in `sunpy.net.jsoc.JSOCClient.get_request`.
```

sunpy/net/jsoc/tests/test_jsoc.py
```diff
@@ -318,3 +318,16 @@ def test_results_filenames():
...
         assert os.path.isfile(hmiurl)
+
+
+@pytest.mark.remote_data
+def test_results_filenames_as_is(tmp_path):
+    responses = client.search(
+        a.Time('2014/1/1T1:00:36', '2014/1/1T01:01:38'),
+        a.jsoc.Series('hmi.M_45s'), a.jsoc.Notify('jsoc@cadair.com'),
+        a.jsoc.Protocol('as-is'))
+    files = client.fetch(responses, path=tmp_path)
```
Cadair: Can you add an:
```python
assert len(responses) == X
```
here to check we are getting the correct number of results before we download them?

PR #3840, opened 2020-02-24
Used parse_time function to compare Time

### LINKED ISSUE #3839: Error parsing time in VSO search filtering
### Description

VSO fails to filter results having Time as Attribute. The bug lies in the imoplementation for filtering time specifically in the line (in `vso/attrs.py`) - 
`attr.min <= _attrs.Time.strptime(it.time.end, _TIMEFORMAT)`
The comparison is occuring between a string and a Time object. Moreover, there is no attribute named `strptime` in Time attribute. 

### Expected behavior

The Time attribute would have been successfully filtered with the new query.

### Actual behavior

See gist link below.

### Steps to Reproduce

Gist link - https://gist.github.com/aribalam/4aef458ebaeeb7cb23d45720f61ced2f

aribalam:
### Description

Removed the old code for filtering Time attribute in `vso/attrs.py`. Converted the time string (`it.time.start` and `it.time.end`) to Time objects using `parse_time` helper function.

Fixes #3839

### DIFF
--- a/changelog/3840.bugfix.rst
+++ b/changelog/3840.bugfix.rst
@@ -1,0 +1,1 @@
+Used the `parse_time` function to compare time in VSO post-search filtering code
--- a/changelog/3840.bugfix.rst
+++ b/changelog/3840.bugfix.rst
@@ -1,1 +1,1 @@
-Used the `parse_time` function to compare time in VSO post-search filtering code
+Updated the method used to filter time in the VSO post-search filtering function.
--- a/sunpy/net/vso/attrs.py
+++ b/sunpy/net/vso/attrs.py
@@ -25,6 +25,8 @@ from functools import singledispatch as _singledispatch
 import astropy.units as u
 
 from sunpy.util.exceptions import SunpyDeprecationWarning
+from sunpy.time import parse_time
+
 from .. import _attrs
 from .. import attr as _attr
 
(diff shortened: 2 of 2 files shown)


sunpy/net/vso/attrs.py
```diff
@@ -353,11 +355,11 @@ def _(attr, results):
         if
         it.time.end is not None
         and
-        attr.min <= _attrs.Time.strptime(it.time.end, _TIMEFORMAT)
+        attr.min <= parse_time(it.time.end)
```
vn-ki: What was the use of `TIMEFORMAT`?

vn-ki: If it's not used anywhere else, remove it.

aribalam: `_TIMEFORMAT` is being used in the converter in the same file.

vn-ki: Given we already know the format, I think using `Time.strptime` might be actually better here.

ayshih: I just pointed out in https://github.com/sunpy/sunpy/issues/3839#issuecomment-594110352 that the code was originally `astropy.time.Time`, but got accidentally changed to `_attrs.Time` in a bad edit, which was why it broke.

Cadair: yeah I agree, we don't need to use `parse_time` we just need to undo the change as @ayshih pointed out in the issue.

aribalam: @vn-ki Check my comment in https://github.com/sunpy/sunpy/issues/3839.

changelog/3840.bugfix.rst
```diff
@@ -0,0 +1 @@
+Used the `parse_time` function to compare time in VSO post-search filtering code
```
Cadair: This changelog is now inaccurate.

nabobalis: Fixed.

nabobalis:
Thanks @aribalam

PR #3842, opened 2020-02-24
Adds checks to ensure that the downloaded FITS files are valid FITS files.

Raahul-Singh:
### Description

Fixes: #3843

It has been observed that many times even if the files are downloaded successfully, they are empty.
This PR adds checks to the code to ensure that the user is informed if something like this happens.

### DIFF
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -4,6 +4,7 @@ import time
 import urllib
 import warnings
 from collections.abc import Sequence
+from sunpy.map import Map
 
 import drms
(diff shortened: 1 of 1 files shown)


Raahul-Singh:
@dstansby how do you propose we check the performance impact? A simple test to see the time it takes to run the code for different download sizes would do?

Cadair:
I don't like this because it makes net import map, let alone the performance overhead of actually reading the file.

I am tempted to just flag this issue as a JSOC bug and say we wont fix it. Other options are to try and verify the file with just `astropy.io.fits` but :man_shrugging:

PR #3770, opened 2020-02-06
Further Fido Refactoring

Cadair:
This PR fixes a number of things in net, I have gone to great lengths to keep the commits independent. If you want to review them one by one you should be able to. There are probably more tests that could do with writing, if you see some missing then let me know.

1) ~Allows clients to register new attrs namespaces, i.e. `a.dkist`~
1) Implements an ABC for `QueryResponse` and makes all the built in clients use it.
1) Allows Fido searches without Time, but makes sure all the clients that require Time don't respond to queries without it.
1) Allows multiple Fido clients to return results from a single AND section of a query. So both `RHESSIClient` and `VSO` now service queries for `a.Instrument("RHESSI")`.

fixes #3830 fixes #3368 fixes #2992 fixes #3814

PR #3852, opened 2020-02-26
fix indexing of QueryResponse for fetch

### LINKED ISSUE #2781: Slicing a JSOC UnifiedResponse object causes Fido.fetch to fail

sashank27:
### Description

Fixes #2781
Continuation of #2796

### DIFF
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -68,6 +68,7 @@ class JSOCResponse(BaseQueryResponse):
         ret = type(self)(self.table[item])
         ret.query_args = self.query_args
         ret.requests = self.requests
+        ret.client = self._client
 
         warnings.warn("Downloading of sliced JSOC results is not supported. "
                       "All the files present in the original response will be downloaded.",
@@ -349,7 +350,6 @@ class JSOCClient(BaseClient):
             # Update blocks with deep copy of iargs because in _make_recordset we use .pop() on element from iargs
             blocks.append(copy.deepcopy(iargs))
             return_results.append(self._lookup_records(iargs))
-
         return_results.query_args = blocks
         return return_results
 
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -40,8 +40,8 @@ class JSOCResponse(BaseQueryResponse):
         """
         super().__init__()
         self.table = table or astropy.table.QTable()
-        self.query_args = None
-        self.requests = None
+        self.query_args = getattr(table, 'query_args', None)
+        self.requests = getattr(table, 'requests', None)
         self._client = client
 
     @property
@@ -63,7 +63,16 @@ class JSOCResponse(BaseQueryResponse):
             return len(self.table)
 
     def __getitem__(self, item):
-        return type(self)(self.table[item])
+        if isinstance(item, int):
+            item = slice(item, item + 1)
+        ret = type(self)(self.table[item])
+        ret.query_args = self.query_args
+        ret.requests = self.requests
+
+        warnings.warn("Downloading of sliced JSOC results is not supported. "
+                      "All the files present in the original response will be downloaded.",
+                      SunpyUserWarning)
+        return ret
 
     def __iter__(self):
         return (t for t in [self])
--- a/sunpy/net/tests/test_fido.py
+++ b/sunpy/net/tests/test_fido.py
@@ -471,15 +471,12 @@ def test_fido_no_time(mocker):
 @pytest.mark.remote_data
 def test_slice_jsoc():
     tstart = '2011/06/07 06:32:45'
-    tend = '2011/06/07 06:35:15'
+    tend = '2011/06/07 06:33:15'
     res = Fido.search(a.Time(tstart, tend), a.jsoc.Series('hmi.M_45s'),
                       a.jsoc.Notify('jsoc@cadair.com'))
 
+    with pytest.warns(SunpyUserWarning):
+        Fido.fetch(res[0, 0])
+
     with pytest.warns(SunpyUserWarning) as record:
-        result = Fido.fetch(res[0, 0])
-        assert isinstance(result, Results)
-
-        result = Fido.fetch(res[0, 0:2])
-        assert isinstance(result, Results)
-
-    assert len(record) == 2
+        Fido.fetch(res[0, 0:1])
--- a/sunpy/net/tests/test_fido.py
+++ b/sunpy/net/tests/test_fido.py
@@ -467,3 +467,19 @@ def test_fido_no_time(mocker):
     Fido.search(a.jsoc.Series("test"))
 
     jsoc_mock.assert_called_once()
+
+@pytest.mark.remote_data
+def test_slice_jsoc():
+    tstart = '2011/06/07 06:32:45'
+    tend = '2011/06/07 06:35:15'
+    res = Fido.search(a.Time(tstart, tend), a.jsoc.Series('hmi.M_45s'),
+                      a.jsoc.Notify('jsoc@cadair.com'))
+
+    with pytest.warns(SunpyUserWarning) as record:
+        result = Fido.fetch(res[0, 0])
+        assert isinstance(result, Results)
+
+        result = Fido.fetch(res[0, 0:2])
+        assert isinstance(result, Results)
+
+    assert len(record) == 2
--- a/changelog/3852.bugfix.rst
+++ b/changelog/3852.bugfix.rst
@@ -1,0 +1,1 @@
+Fix failing of fetching of the indexed JSOCResponses using `Fido.fetch`.


Cadair:
There is some overlap here with #3770 although I think that it's a good idea to have this PR to bugfix this issue on the 1.1 and 1.0 branches.

I also think I might have not fixed this bug in #3770 despite thinking I did.

sunpy/net/fido_factory.py
```diff
@@ -91,10 +91,7 @@ def _handle_record_slice(self, client_resp, record_slice):
 
         # Make sure we always have an iterable, as most of the response objects
         # expect one.
-        if isinstance(record_slice, int):
-            resp = [client_resp[record_slice]]
-        else:
-            resp = client_resp[record_slice]
+        resp = list(client_resp[record_slice])
```
Cadair: I don't think this logic is quite the same as what it is removing. For example `list('hello')` isn't the same operation as `['hello']` given I think I have changed this block in #3770 I propose you revert this change.

sunpy/net/jsoc/jsoc.py
```diff
@@ -58,7 +63,10 @@ def __len__(self):
             return len(self.table)
 
     def __getitem__(self, item):
-        return type(self)(self.table[item])
+        resp = type(self)(self.table[item])
```
Cadair: The pattern I went with in #3770 (although not in the `JSOCClient` apparently) was this https://github.com/sunpy/sunpy/pull/3770/files#diff-0c4a0061c4c5eefc3ba432fa7a2b1280R157-R160

Would this apply here?

sunpy/net/fido_factory.py
sashank27: For single-element slicing, the previous implementation works but fails for multi-element slicing.
Eg: If we do slicing as `res[0, 0]`, then `resp` would be a list of `JSOCResponse`. But if we do slicing as `res[0, 0:2]`, then it gives a `Table`. In that case, we have to handle `JSOCResponse` differently. 
So I thought to maintain the uniformity, we should cast all of the slicings as a list.

Cadair: I think the issue is

> But if we do slicing as res[0, 0:2], then it gives a Table.

not how we are handling it here.

I am not 100% sure though, it's hard to think about this without really jumping into it.

sunpy/net/jsoc/jsoc.py
sashank27: but JSOCResponse takes only `table` as an argument. Should I add `**kwargs` as well?

Cadair: I was referring to indexing like `[0:1]` rather than `[0]` so you get an iterable object back.

sashank27: oh okay

sunpy/net/fido_factory.py
```diff
@@ -95,12 +99,15 @@ def _handle_record_slice(self, client_resp, record_slice):
             resp = [client_resp[record_slice]]
         else:
             resp = client_resp[record_slice]
-
         # Reconstruct a response object with the sub-indexed records.
         ret = resp_type(resp)
         # Make sure we pass the client back out again.
         ret.client = client_resp.client
 
+        warnings.warn("Downloading of sliced results is not supported. "
```
Cadair: This warning will be triggered for all clients, even ones which do support this. This should be in `JSOCClient.__getitem__`.

sunpy/net/jsoc/jsoc.py
```diff
@@ -38,9 +38,14 @@ def __init__(self, table=None):
...
-        self.query_args = None
-        self.requests = None
+        if isinstance(table, list) and isinstance(table[0], type(self)):
+            self.table = table[0].table
+            self.query_args = table[0].query_args
+            self.requests = table[0].requests
+        else:
+            self.table = table
+            self.query_args = table.query_args if hasattr(table, 'query_args') else None
+            self.requests = table.requests if hasattr(table, 'requests') else None
```
Cadair:
```suggestion
            self.query_args = getattr(table, 'query_args', None)
            self.requests = getattr(table, 'requests', None)
```

Cadair: You can refactor this logic a little to make it so that it's fewer lines.

```suggestion
        if isinstance(table, list) and isinstance(table[0], type(self)):
            table = table[0].table
        self.table = table
        self.query_args = table.query_args if hasattr(table, 'query_args') else None
        self.requests = table.requests if hasattr(table, 'requests') else None
```

sunpy/net/jsoc/jsoc.py
```diff
@@ -58,7 +63,13 @@ def __len__(self):
             return len(self.table)
 
     def __getitem__(self, item):
-        return type(self)(self.table[item])
+        if isinstance(item, int):
+            item = slice(item, item+1)
```
Cadair:
```suggestion
            item = slice(item, item + 1)
```

sunpy/net/jsoc/jsoc.py
```diff
@@ -38,9 +38,14 @@ def __init__(self, table=None):
         table : `astropy.table.Table`
         """
 
-        self.table = table
-        self.query_args = None
-        self.requests = None
+        if isinstance(table, list) and isinstance(table[0], type(self)):
```
Cadair: What happens if it's a list and longer than length 1? Also what happens if it's passed an instance of `type(self)` not in a list?

sunpy/net/fido_factory.py
nabobalis: What clients support this?

Cadair: All of them but JSOC.

sunpy/net/jsoc/jsoc.py
sashank27: I believe these cases will never arise. The list of length 1 is only created when we handle slicing in Fido.
 https://github.com/sunpy/sunpy/blob/c6b475659c1c366c574e8b755e7a828c31b6182f/sunpy/net/fido_factory.py#L94-97

I'm not sure still of these corner cases appearing in any scenario. Let me know if there is one.

Cadair: OK, I think this is fine as a bugfix, the "correct" fix is probably in #3770 but is much deeper and affects all clients. :+1:

sunpy/net/fido_factory.py
```diff
@@ -14,6 +14,8 @@
 
 from parfive import Downloader, Results
 
+import warnings
```
Cadair:
```suggestion
```

sunpy/net/jsoc/jsoc.py
```diff
@@ -58,7 +60,16 @@ def __len__(self):
...
 
     def __getitem__(self, item):
-        return type(self)(self.table[item])
+        if isinstance(item, int):
+            item = slice(item, item + 1)
+        ret = type(self)(self.table[item])
+        ret.query_args = self.query_args
+        ret.requests = self.requests
+
+        warnings.warn("Downloading of sliced results is not supported. "
```
Cadair:
```suggestion
        warnings.warn("Downloading of sliced JSOC results is not supported. "
```

sunpy/net/jsoc/jsoc.py
```diff
@@ -58,7 +60,16 @@ def __len__(self):
             return len(self.table)
 
     def __getitem__(self, item):
-        return type(self)(self.table[item])
+        if isinstance(item, int):
+            item = slice(item, item + 1)
+        ret = type(self)(self.table[item])
+        ret.query_args = self.query_args
+        ret.requests = self.requests
```
nabobalis: If these are `None` what happens?

sashank27: `__getitem__` is just used to get the slice of the response. All these lines preserve the properties of the original response, and so if they are originally `None` they can be safely passed on as `None`

sunpy/net/tests/test_fido.py
```diff
@@ -478,3 +478,18 @@ def test_client_fetch_wrong_type(mock_fetch):
...
+def test_slice_jsoc():
+    tstart = '2011/06/07 06:32:45'
+    tend = '2011/06/07 06:35:15'
+    res = Fido.search(a.Time(tstart, tend), a.jsoc.Series('hmi.M_45s'),
+                      a.jsoc.Notify('jsoc@cadair.com'))
+
+    with pytest.warns(SunpyUserWarning):
+        result = Fido.fetch(res[0, 0])
+        assert isinstance(result, Results)
+
```
nabobalis: Do we not need to with blocks to make sure the second one raises?

sashank27: Sorry I didn't understand 😅

nabobalis: I meant to say that you slice the results twice and make sure that pytest checks that there is a warning. But without a second with block, can you confirm that the second one has also raised its warning?

sunpy/net/jsoc/jsoc.py
nabobalis: Fair enough.

sunpy/net/tests/test_fido.py
sashank27: Ohh sorry, I thought it would be duplication of code. So I believe I would have to add `with` block for every slicing.

sashank27: or we can check for the count of warnings raised

sunpy/net/jsoc/jsoc.py
```diff
@@ -40,8 +40,8 @@ def __init__(self, table=None, client=None):
         """
         super().__init__()
         self.table = table or astropy.table.QTable()
```
nabobalis: The `if isinstance(table, list) and isinstance(table[0], type(self)):` is not here now but it didn't seem to break anything so ?

nabobalis:
Online failure is a doctest issue in `docs/guide/acquiring_data/fido.rst` and is unrelated.

sunpy/net/jsoc/jsoc.py
nabobalis: Fun fact, this was needed on the backports but not on master with the Fido changes.

PR #3869, opened 2020-03-06
Implement client specific attrs registration

dstansby:
Another part of #3770

### DIFF
--- a/sunpy/net/dataretriever/sources/goes.py
+++ b/sunpy/net/dataretriever/sources/goes.py
@@ -191,6 +191,10 @@ class XRSClient(GenericClient):
                 return all(chklist)
         return False
 
+    @classmethod
+    def _attrs_module(cls):
+        return 'goes', 'sunpy.net.dataretriever.attrs.goes'
+
 
 class SUVIClient(GenericClient):
     """
@@ -389,3 +393,7 @@ class SUVIClient(GenericClient):
             return True
         else:
             return False
+
+    @classmethod
+    def _attrs_module(cls):
+        return 'goes', 'sunpy.net.dataretriever.attrs.goes'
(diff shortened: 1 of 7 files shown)


changelog/3770.feature.rst
```diff
@@ -0,0 +1 @@
+Fido clients (subclasses of `sunpy.net.base_client.BaseClient`) can now register their own attrs modules with `sunpy.net.attrs`. This allows clients which require attr classes specific to that client to register modules that can be used by the user i.e. ``a.vso``. It also allows clients implemented externally to sunpy to register attrs.
```
dstansby:
```suggestion
Fido clients (subclasses of `sunpy.net.base_client.BaseClient`) can now register their own attrs modules with `sunpy.net.attrs`. This allows clients which require attr classes specific to that client to register attributes that can be used by the user i.e. ``a.vso``. It also allows clients implemented externally to sunpy to register attrs.
```
... unless I'm reading something wrong?

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -191,6 +191,10 @@ def _can_handle_query(cls, *query):
                 return all(chklist)
         return False
 
+    @classmethod
+    def _attrs_module(cls):
```
dstansby: There should probably be some docs on how to write this function (ie. should be called `_attrs_module()`, should return these values...) for people wanting to do this themselves?

nabobalis: I believe there is a seperate PR to update the docs.

changelog/3770.feature.rst
Cadair: You register a module (i.e. `a.vso` maps to `sunpy.net.vso.attrs`) which contains attrs (but is just a module). [here](https://github.com/sunpy/sunpy/pull/3869/files#diff-85c0984f67d7abb954d0f4a8d56fcb4fR46)

Cadair:
@dstansby you happy with this?

PR #3870, opened 2020-03-07
Fix peek warning

dstansby:
This prevents an AstropyWarning when calling `if draw_grid:`, where `draw_grid` is a Quantity. Instead, force `draw_grid` to be a bool, and if it's started as a quantity just save the value in another variable.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1665,6 +1665,14 @@ class GenericMap(NDData):
 
         im = self.plot(axes=axes, **matplot_args)
 
+        grid_spacing = None
+        # Handle case where draw_grid is actually the grid sapcing
+        if isinstance(draw_grid, u.Quantity):
+            grid_spacing = draw_grid
+            draw_grid = True
+        elif not isinstance(draw_grid, bool):
+            raise TypeError("draw_grid should be a bool or an astropy Quantity.")
+
         if colorbar:
             if draw_grid:
                 pad = 0.12  # Pad to compensate for ticks and axes labels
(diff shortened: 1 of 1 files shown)


Cadair:
Looks good but I think you made it angry!

ayshih:
I think these changes make the code harder to read.  If the only problematic line is [this one](https://github.com/sunpy/sunpy/pull/3870/files#diff-9eadd681e99c7419c431865d29b5ca1dL1669), why not simply change it to `if draw_grid is not None:`?  The other `if draw_grid:` is protected by a type check.

ayshih:
Oops, I see that `draw_grid` defaults to `False`.  Okay, so that one line would be a little more complex – `if isinstance(draw_grid, bool) and draw_grid:` – but it's still cleaner to change that one line than to juggle another variable.

dpshelio:
I would prefer that `draw_grid` is just boolean and don't give the option to provide a size. That's making life harder when it could have a single variable for a single thing. (No that this is something you introduced, but I think it is a good place to mention it).

Cadair:
I kind of agree, but I think :ship: has sailed on that now.

wtbarnes:
Looks like the figure tests are failing because `grid_size` is getting propagated further down in the `Patch` object than it should?

No idea what is going on with the Azure tests...

dstansby:
> Oops, I see that `draw_grid` defaults to `False`. Okay, so that one line would be a little more complex – `if (isinstance(draw_grid, bool) and draw_grid) or isinstance(draw_grid, u.Quantity):` – but it's still cleaner to change that one line than to juggle another variable.

I think adding the new variable makes it much clearer what is being stored (ie. the grid spacing) when `draw_grid` isn't a bool. I've added a comment to that effect.

dstansby:
I also moved the type checking up so that all the type checking is in one place now.

sunpy/map/mapbase.py
```diff
@@ -1675,13 +1683,11 @@ def peek(self, draw_limb=False, draw_grid=False,
         if draw_limb:
             self.draw_limb(axes=axes)
 
-        if isinstance(draw_grid, bool):
-            if draw_grid:
+        if draw_grid:
+            if grid_spacing is None:
```
wtbarnes: Maybe this is ignorant question, but what's the reasoning behind adding the `if-else` block here? Could it not just be 
```python
self.draw_grid(axes=axes, grid_spacing=grid_spacing)
```
independent of whether `grid_spacing` is `None`? Is there some reason to not pass `None` to `draw_grid`?

dstansby: Passing `grid_spacing=None` to `draw_grid()` isn't supported according to the docstring, and I think I tried this a while ago and it indeed was not supported.

PR #3874, opened 2020-03-09
Improve subclassing attrs.Time and JSOC _can_handle_query

Cadair:
This is a couple of small fixes from #3770 

The first change is motivated by the DKIST client in dkistdc/dkist#73 where I am subclassing `a.Time` for a different purpose. Changing our checks to be for type rather than instance means that you can combine two different subclasses of Time in one query without issue.

The other is to make sure that the JSOC client can only handle queries if `a.jsoc.Series` is provided, otherwise queries with just `a.Wavelength` or `a.Time` match the JSOC but then .search errors.

### DIFF
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -860,10 +860,19 @@ class JSOCClient(BaseClient):
 
     @classmethod
     def _can_handle_query(cls, *query):
-        chkattr = ['Series', 'Protocol', 'Notify', 'Wavelength', 'Time',
-                   'Segment', 'Keys', 'PrimeKey', 'Sample']
+        # Import here to prevent circular imports
+        from sunpy.net import attrs as a
 
-        return all([x.__class__.__name__ in chkattr for x in query])
+        required = {a.jsoc.Series}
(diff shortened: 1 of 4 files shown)


dstansby:
Looks like this is good to go pending a rebase

Cadair:
rebased.

sunpy/net/jsoc/jsoc.py
```diff
@@ -860,10 +860,19 @@ def _lookup_records(self, iargs):
...
+        required = {a.jsoc.Series}
+        optional = {a.jsoc.Protocol, a.jsoc.Notify, a.Wavelength, a.Time,
+                    a.jsoc.Segment, a.jsoc.Keys, a.jsoc.PrimeKey, a.Sample}
+        query_attrs = {type(x) for x in query}
+        all_attrs = required.union(optional)
+
+        if not required.issubset(query_attrs) or not query_attrs.issubset(all_attrs):
+            return False
+
+        return True
```
dpshelio:
```suggestion
        return required.issubset(query_attrs) and query_attrs.issubset(all_attrs):
 ```

PR #3887, opened 2020-03-11
Fixed missing `sunpy.coordinates.utils` in docs

ayshih:
See title


ayshih:
Backports will need to be done manually

PR #3894, opened 2020-03-12
Fixed bug where some transformation errors raised ValueError instead ConvertError

ayshih:
Some of our transformations raise `ValueError` instead of `ConvertError` (from Astropy).  This PR fixes them to all raise `ConvertError`.


ayshih:
Strictly speaking, this is an API change, so perhaps should not be backported

ayshih:
The test failure was actually just during test collection – for whatever reason – and I'm sure that the test suite would succeed if re-ran.

Regardless, I'm closing this PR because it's moot.  This commit was already merged in as part of #3782.  I rebased that PR off of this commit to fix a test.  Sorry!

Issue #3824, opened 2020-02-20
Use the Zenodo API to get the citation for the current SunPy version


Cadair:
Our `sunpy.__citation__` functionality should give you the info for the zenodo record for the version you run it with.

Cadair:
See #3803 

@dpshelio you want to take this one?

mehulfollytobevice:
Hey , I want to work on this .

Cadair:
I started looking into this and couldn't find any documentation on how one would do this with the zenodo API. So the biggest task here will be working out how to ask Zenodo for the info.

mehulfollytobevice:
I will try to get the information about the cititation .

mehulfollytobevice:
> Our `sunpy.__citation__` functionality should give you the info for the zenodo record for the version you run it with.

hey was this functionality used in the earlier versions , or do we have to create it

Cadair:
We have it returning the ApJ paper citation, the idea of this issue is that it would also return the citation info for the running version of SunPy.

dstansby:
Looks like the API we need is still under testing and not public: https://developers.zenodo.org/#rest-api

wtbarnes:
> Looks like the API we need is still under testing and not public: https://developers.zenodo.org/#rest-api

Looking at that link, it seems like this API is now ready to use.

PR #3896, opened 2020-03-13
Changed quoted sentence (that we suggest authors add to their research papers) in CITATION.rst

mbobra:
I changed the sentence we suggest authors add to their research papers. Previously, it was: 

> This research has made use of SunPy vX.Y (software citation), an open-source and free community-developed solar data analysis Python package (project citation).

I suggest the following:

> This research used version vX.Y.Z (software citation) of the SunPy open source software package (project citation).

Here are my reasons:

- It is easier to read (and eliminates the passive voice)
- I added vX.Y.Z instead of vX.Y because we archive each vX.Y.Z release in Zenodo, therefore a reference to vX.Y is ambiguous when referring to the software citation
- I think we chose not to put a dash between open and source in the SunPy v1.0 paper
- It is more likely someone will add a short sentence into the methods or conclusion section of a paper (whereas a longer sentence seems more likely to end up in the acknowledgements section of a paper, which gets less visibility).

### DIFF
--- a/sunpy/CITATION.rst
+++ b/sunpy/CITATION.rst
@@ -9,8 +9,8 @@ Citing SunPy in Publications
 
 Please add the following line within your methods, conclusion or acknowledgements sections:
 
-   *This research has made use of SunPy vX.Y (software citation), an open-source and free
-   community-developed solar data analysis Python package (project citation).*
+   *This research used version vX.Y.Z (software citation) of the SunPy open source
+   software package (project citation).*
 
 The project citation should be to the `SunPy paper`_, and the software citation should be the specific `Zenodo DOI`_ for the version used in your work.
 
--- a/sunpy/CITATION.rst
+++ b/sunpy/CITATION.rst
@@ -9,7 +9,7 @@ Citing SunPy in Publications
 
 Please add the following line within your methods, conclusion or acknowledgements sections:
 
-   *This research used version vX.Y.Z (software citation) of the SunPy open source
+   *This research used version X.Y.Z (software citation) of the SunPy open source
    software package (project citation).*
 
 The project citation should be to the `SunPy paper`_, and the software citation should be the specific `Zenodo DOI`_ for the version used in your work.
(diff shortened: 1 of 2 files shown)


nabobalis:
The astropy version is:
```
This research made use of Astropy,\footnote{http://www.astropy.org} a community-developed core Python package for Astronomy \citep{astropy:2013, astropy:2018}.
```
I wonder if we co-opt that style more.

mbobra:
They only suggest citing the article describing the software, not citing the DOI for the software archive. The [AAS Policy Statement on Software](https://journals.aas.org/policy-statement-on-software/) says: "Ideally, both forms of citation should be included."

nabobalis:
What about:

```
This research made use of SunPy vX.Y.Z (software citation), a Python library for Solar Physics (project citation).
```

dpshelio:
I wonder whether we could be more direct and show what to copy paste (like the latex format that astropy uses).

Cadair:
I suggest we combine that with #3824 as then we can give them the full citation (as part of `__citation__`?)

dpshelio:
#3824 may take a bit longer... This is a small change that I'm in favour. Though as @nabobalis said, maybe removing the "open source" makes it more readable.

sunpy/CITATION.rst
```diff
@@ -9,8 +9,8 @@ Citing SunPy in Publications
 
 Please add the following line within your methods, conclusion or acknowledgements sections:
 
-   *This research has made use of SunPy vX.Y (software citation), an open-source and free
-   community-developed solar data analysis Python package (project citation).*
+   *This research used version vX.Y.Z (software citation) of the SunPy open source
```
nabobalis:
```suggestion
   *This research used version X.Y.Z (software citation) of the SunPy open source
```

sunpy/CITATION.rst
```diff
@@ -9,8 +9,8 @@ Citing SunPy in Publications
 
 Please add the following line within your methods, conclusion or acknowledgements sections:
```
nabobalis: Do want to change up this sentence too?

mbobra: Maybe we don't have to be that specific

sunpy/CITATION.rst
mbobra: Good idea, I committed this one

PR #3899, opened 2020-03-15
updated HEKClient to return json again

nabobalis:
Fixes #3898

PR #3902, opened 2020-03-17
Fixed incorrect module name in test on 1.1

ayshih:
This bug was introduced in a backport (#3858)

(an analogous bug was introduced to 1.0 in #3857)


Cadair:
Test fails are a result of #3899 not having been backported yet.

PR #3909, opened 2020-03-20
Add contour map method

dstansby:
I'm doing some contouring at the moment, and thought a helper function might be helpful. My current use case is contouring AIA 193 maps to isolate coronal holes, and then plotting those contours on other maps.

Essentially this just uses `skimage.measure.contours`, but returns the result as `SkyCoord` objects. Here is what the example I have added to the gallery looks like:

sphx_glr_map_contouring_001

Needs

- [x] A documentation example
- [x] What's new
- [ ] Tests

Test ideas:
Sample maps with:
- No contours
- One contour

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -2091,6 +2091,33 @@ class GenericMap(NDData):
 
         return ret
 
+    def contour(self, level, **kwargs):
+        """
+        Returns coordinates of the contours for a given level value.
+
+        For details of the contouring algorithm see `skimage.measure.find_contours`.
+
+        Parameters
+        ----------
+        level : float
+            Value along which to find contours in the array.
+        kwargs :
+            Additional keyword arguments are passed to `skimage.measure.find_contours`.
+
+        Returns
+        -------
+        contours: list of (n,2) `~astropy.coordinates.SkyCoord`
+            Coordinates of each contour.
+
+        See also
+        --------
+        `skimage.measure.find_contours`
+        """
+        from skimage import measure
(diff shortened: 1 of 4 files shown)


dpshelio:
I think it's very useful!! Whether in the future we want these contours as RoIs and be able to extract properties based on them on other datasets is something we could discuss beyond this PR. For now, having this available simple as it's now I think is good. :+1:  from me!

sunpy/map/mapbase.py
```diff
@@ -1821,6 +1821,33 @@ def plot(self, annotate=True, axes=None, title=True,
...
+        contours: list of (n,2) `~astropy.coordinates.SkyCoord`
+            Coordinates of each contour.
+
+        See also
+        --------
+        `skimage.measure.find_contours`
+        """
+        from skimage import measure
+        contours = measure.find_contours(self.data, level=level, **kwargs)
+        contours = [self.wcs.array_index_to_world(c[:, 0], c[:, 1]) for c in contours]
```
Cadair, line 1848:
```suggestion
        contours = [self.wcs.array_index_to_world(*c.T) for c in contours]
```

hayesla:
I think just a small test is now needed

ehsteve:
I feel like this provides 90% of the functionality of compositemap. I am not complaining! Just another reason to potentially remove compositemap. Ping @cadair.

dstansby:
Thanks! Since this won't make it in to 2.0 anyway, I'm waiting until I have a bit more time to add tests.

dstansby:
I think methods are more pythonic than functions, but happy to go either way if someone wants to make a decision.

Issue #1155, opened 2014-08-26
Make Map (and other datatypes) have setters for metadata


Cadair:
I know when @rhewett and I were building the new `Map` class we chose not to do this because we didn't want to encourage the user to break things, however, if we are to maintain the ability to save out files using similar meta data to the file read in, implementing setters so that the metadata -> property mapping works both ways would be very useful.

Also I see a situation where the user will want to manually update the metadata, because they have written a function to do something custom, and having to manually know what the keys in the dicts are rather than being able to use instrument independent setters seems a little silly.

abit2:
@Cadair do all the metadata need setters?

samaloney:
I guess with great power comes great responsibility - how would implicitly related keywords be handled I'm specifically thinking of `DATE_OBS`, `DATE_BEG`, DATE_AVG`, `DATE_END` that all SolarOrbiter files will have and some current fit files have `DATE_OBS` and `DATE_END`?

dstansby:
As a slight aside, regardless of setting properties, should we be adding properties for all of these extra FITS keywords?

dstansby:
Closing this as per https://github.com/sunpy/sunpy/pull/3911#issuecomment-614728637

PR #3911, opened 2020-03-22
Add setter for map.date

dstansby:
I find myself wanting to change the date of a map fairly regularly at the minute, so it would be good to have a way to do this without having to go into the metadata. See also #1155.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -462,6 +462,13 @@ class GenericMap(NDData):
 
     @date.setter
     def date(self, val):
+        """
+        Set the image observation time.
+
+        Notes
+        -----
+        This sets the DATE-OBS fits keyword in the map metadata.
+        """
         val = parse_time(val)
         self.meta['date-obs'] = val.isot
 
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -460,6 +460,11 @@ class GenericMap(NDData):
             time = self._default_time
         return parse_time(time)
 
+    @date.setter
+    def date(self, val):
+        val = parse_time(val)
+        self.meta['date-obs'] = val.isot
+
     @property
     def detector(self):
         """Detector name."""
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -175,6 +175,12 @@ def test_date_aia(aia171_test_map):
     assert aia171_test_map.date == parse_time('2011-02-15T00:00:00.34')
 
 
+def test_date_setter(generic_map):
+    dtime_old = generic_map.date
+    generic_map.date = '2011-02-15T00:00:00.34'
+    assert generic_map.date == parse_time('2011-02-15T00:00:00.34') != dtime_old
+
+
 def test_detector(generic_map):
     assert generic_map.detector == 'bar'
 
--- a/changelog/3911.feature.rst
+++ b/changelog/3911.feature.rst
@@ -1,0 +1,1 @@
+Added the ability to set the ``.date`` property of a `~sunpy.map.GenericMap`.


wtbarnes:
This looks good to me (other than needing a changelog 😅 ). The online build failures look unrelated to me.

A change like this seems quite related to the question of map mutability though, i.e. to what degree do we want to provide users the ability to change an existing map versus creating a new map with the desired properties. I'd like to hear some other opinions on this (maybe another topic for the coordination meeting?). It looks like this conversation was sort of started in the very old issue you linked above.

dstansby:
> This looks good to me (other than needing a changelog 😅 )

I'm confused by this, there is a changelog file but the bot is complaining

> 
> A change like this seems quite related to the question of map mutability though, i.e. to what degree do we want to provide users the ability to change an existing map versus creating a new map with the desired properties. I'd like to hear some other opinions on this (maybe another topic for the coordination meeting?). It looks like this conversation was sort of started in the very old issue you linked above.

I think the big thing here is being able to provide a level of abstraction above the FITS standard - if I want to create a new map with a different date, I have to get the FITS header, know where and how to change the date key, and then feed that back into the `Map` constructor, which is a lot of work and knowledge needed for a simple task.

sunpy/map/tests/test_mapbase.py
```diff
@@ -175,6 +175,12 @@ def test_date_aia(aia171_test_map):
     assert aia171_test_map.date == parse_time('2011-02-15T00:00:00.34')
 
 
+def test_date_setter(generic_map):
+    dtime_old = generic_map.date
+    generic_map.date = '2011-02-15T00:00:00.34'
+    assert generic_map.date == parse_time('2011-02-15T00:00:00.34') != dtime_old
```
dpshelio:
```suggestion
    new_date = '2011-02-15T00:00:00.34'
    generic_map.date = new_date
    assert generic_map.date == parse_time(new_date) != dtime_old
```
sorry for my pedancity of becoming DRY-er. This way we can change the date or parametrise it to use multiple range of dates if we need.

sunpy/map/tests/test_mapbase.py
```diff
@@ -175,6 +175,12 @@ def test_date_aia(aia171_test_map):
...
+
+
```
dpshelio: Should we also add a negative test? What are we not going to accept? Are the tests on `parse_time` enough?

sunpy/map/mapbase.py
```diff
@@ -460,6 +460,11 @@ def date(self):
             time = self._default_time
         return parse_time(time)
 
+    @date.setter
+    def date(self, val):
+        val = parse_time(val)
+        self.meta['date-obs'] = val.isot
```
dpshelio: Is there any extra check we need to do? Do we keep a record somewhere that the date has been changed? What's the motivation for changing the date?

wtbarnes:
Now it is saying it found the changelog and it didn't find the changelog. perfect 😭

dstansby:
> @dstansby why are you doing this so frequently? Is it because the "raw" data is faulty? is it to enable some map operations that need the same time?...

My use case is that I have a synoptic map with dtime `t1`, and a bunch of coordinates in Carrington lon/lat, but with dtime `d2`. If I want to plot one on top of the other, the datetimes have to match, so I'm changing the map dtime to match the coordinate dtime.

sunpy/map/tests/test_mapbase.py
dstansby: I think parse_time is generic enough, since the reason parse_time exists is to massage a range of sensible inputs into an `astropy.time.Time`.

sunpy/map/mapbase.py
dstansby: I think as long as we're updating the meta we *should* be fine, as that is where all properties draw the map metadata from.

samaloney:
What happens to other date related headers i.e `DATE_END` shouldn't they also be updated?

dstansby:
It doesn't look like any other date related FITS keywords are used to get properties of the map. I'll make it clear in the docstring that setting `.date` sets the `DATE-OBS` FITS keyword in the header.

samaloney:
I was thinking more about headers that implicitly relate to DATE_OBS rather than explicit properties but if no one else see’s an issue maybe it fine. I also made a comment in the related issue too.

Cadair:
I have pretty strong reservations about this because it's a can of worms. There are a whole bunch of use cases for metadata setters in general, but without some consideration of the wider picture of various bits of metadata having to move together I would really rather leave setting metadata as an explicit operation the user has to do rather than potentially making it look like it's sanctioned in some way.

dstansby:
> I have pretty strong reservations about this because it's a can of worms. There are a whole bunch of use cases for metadata setters in general, but without some consideration of the wider picture of various bits of metadata having to move together I would really rather leave setting metadata as an explicit operation the user has to do rather than potentially making it look like it's sanctioned in some way.

I closed this, but I still think having setters is a good idea. Doing `m.date = my_date` is still explicitly an operation the user has to do.

As I've noted for the observer coordinate in https://github.com/sunpy/sunpy/issues/4189, setters are a natural way to abstract above FITS, avoiding a user having to learn what FITS metadata is, what key needs changing, what exact format the key is etc. With a setter all they need is a `parse_time` compatible time.

PR #3221, opened 2019-06-14
Restore online test suite. 

nabobalis:
Should restore the online tests. 
Need to decide what to do about the 32bit issue.

Issue #3918, opened 2020-03-25
Make a policy for the minimum version of Astropy that we will support


ayshih:
We should define a policy for the minimum version of Astropy that we will support for each future SunPy release (e.g., latest LTS, N versions back, ...).  We need to consider our release schedule versus Astropy's release schedule.

(See #3154, where we don't actually know when Astropy v3.2+ will become the minimum version.)

Cadair:
My vote would be either the newest stable version 12 months ago or previous 2 releases. Pinning to a LTS is troublesome because of our out of phase cycles.

dstansby:
For an example of a minimum version policy, see https://matplotlib.org/devel/min_dep_policy.html. Since astropy release a LTS every two years, supporting astropy latest LTS versions upwards seems like one sensible option.

Cadair:
There is also NEP 29 https://numpy.org/neps/nep-0029-deprecation_policy.html we could adopt which in that we could take all our minimum versions to be 24 months like it suggests for Numpy. Although it should be noted that Astropy effectively provides our minimum numpy requirement (as they are often newer than NEP 29 especially for their LTS releases (so the min numpy version isn't 4 years old by the end of their LTS cycle).

I am for the record a hard -1 on supporting the last Astropy LTS release, we are too tightly coupled to Astropy to be using something which is potentially (due to phase issues) 30 months old.

Cadair:
(I am also kind of +0 on a SEP for this, but I am open to what others think on that)

dstansby:
Also +0 for a SEP. In the meantime, for sunpy 2.1 I propose we specify astropy >= 4, will have been out for 1y when sunpy 2.1 is released hopefully and is a LTS.

Cadair:
We appear to have adopted a "bump it by a version" each release and master has already been bumped to 4.0. This does kind of get a little sticky with the super delayed 4.1 release, but otherwise I am pretty happy with the emergent status quo. Perhaps we should write it down.

ayshih:
> master has already been bumped to 4.0

Which reminds me that we should make some progress on #4000, although none of the stuff identified has any real benefit to users.

ayshih:
For the record:

- 2.0 depends on Astropy 3.2
- 2.1 depends on Astropy 4.0

If we continue this way,
- 3.0 can depend on Astropy 4.1

dstansby:
Fixed by https://github.com/sunpy/sunpy/pull/5365

PR #3919, opened 2020-03-25
Improve the accuracy of `get_horizons_coord()`

ayshih:
By bumping our Astropy dependency to v3.2, I can better match the coordinate frame that JPL Horizons uses for calculations, and hence more accurately create a SkyCoord from its output.  This PR significantly decreases small discrepancies between Astropy (using a JPL ephemeris rather than Astropy's built-in ephemeris) and JPL Horizons.

Miscellaneous notes:
- This PR sets Astropy to use a JPL ephemeris (DE432s), which means an additional ~10 MB download for the online test build.
- The Earth discrepancy improves from tens of kilometers to tens of meters.  I can't reproduce my earlier claim (https://github.com/sunpy/sunpy/issues/3154#issuecomment-502938102) that the discrepancy would be sub-meter, so my current assumption is that I made a calculation error last year.
- This PR adds some (light) `hypothesis` testing to hopefully avoid issues like https://github.com/sunpy/sunpy/pull/3221#issuecomment-502646325)

### DIFF
--- a/changelog/3919.bugfix.rst
+++ b/changelog/3919.bugfix.rst
@@ -1,0 +1,1 @@
+The accuracy of the output of :func:`sunpy.coordinates.ephemeris.get_horizons_coord` is significantly improved.
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,7 @@ install_requires =
   scipy>=1.0.0
   matplotlib>=2.1.2
   pandas>=0.23.0
-  astropy>=3.1
+  astropy>=3.2
   parfive[ftp]
   importlib_resources;python_version<"3.7"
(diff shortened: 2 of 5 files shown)


ayshih:
Whether this can be backported depends on whether we want to bump the Astropy dependency for v1.x

dstansby:
Do we have a policy on minimum versions of dependencies (e.g. see https://matplotlib.org/devel/min_dep_policy.html for an example). If not maybe we should create one next week...

Cadair:
See #3918 for a discussion about astropy, which I consider to be a special case. We have much less strict requirements on many others.

Issue #3922, opened 2020-03-27
VSO Results from RHESSI seem to not have consistent ordering


Cadair:
If you run the following query multiple times:
```python
Fido.search(a.Time('2012/3/4', '2012/3/4 02:00'), a.Instrument('rhessi'))
```

```
In [1]: Results from 2 Providers:

3 Results from the VSOClient:
   Start Time [1]       End Time [1]    Source Instrument     Type    Wavelength [2]
                                                                           keV      
------------------- ------------------- ------ ---------- ----------- --------------
2012-03-03 22:57:40 2012-03-04 00:33:20 RHESSI     RHESSI PARTIAL_SUN 3.0 .. 17000.0
2012-03-04 01:45:40 2012-03-04 02:09:00 RHESSI     RHESSI PARTIAL_SUN 3.0 .. 17000.0
2012-03-04 00:33:20 2012-03-04 01:45:40 RHESSI     RHESSI PARTIAL_SUN 3.0 .. 17000.0

1 Results from the RHESSIClient:
     Start Time           End Time      Source Instrument Wavelength
------------------- ------------------- ------ ---------- ----------
2012-03-04 00:00:00 2012-03-04 23:59:59 rhessi     rhessi        nan
```

```
In [1]: Results from 2 Providers:

3 Results from the VSOClient:
   Start Time [1]       End Time [1]    Source Instrument     Type    Wavelength [2]
                                                                           keV      
------------------- ------------------- ------ ---------- ----------- --------------
2012-03-04 01:45:40 2012-03-04 02:09:00 RHESSI     RHESSI PARTIAL_SUN 3.0 .. 17000.0
2012-03-04 00:33:20 2012-03-04 01:45:40 RHESSI     RHESSI PARTIAL_SUN 3.0 .. 17000.0
2012-03-03 22:57:40 2012-03-04 00:33:20 RHESSI     RHESSI PARTIAL_SUN 3.0 .. 17000.0

1 Results from the RHESSIClient:
     Start Time           End Time      Source Instrument Wavelength
------------------- ------------------- ------ ---------- ----------
2012-03-04 00:00:00 2012-03-04 23:59:59 rhessi     rhessi        nan

```

```
In [1]: Results from 2 Providers:

3 Results from the VSOClient:
   Start Time [1]       End Time [1]    Source Instrument     Type    Wavelength [2]
                                                                           keV      
------------------- ------------------- ------ ---------- ----------- --------------
[...]

PR #3923, opened 2020-03-27
Skip the problematic VSO doctest

Cadair:
This is a workaround for #3922 so we can get the bug fix releases out.


Cadair:
(┛◉Д◉)┛彡┻━┻

nabobalis:
I wonder why.

PR #3925, opened 2020-03-27
Bump astropy version in figure tests

dstansby:
I'm running into some warnings thrown by astropy 3.1 in the figure tests (e.g. https://app.circleci.com/pipelines/github/sunpy/sunpy/4264/workflows/492b5606-4a71-4a74-a521-5656b44502f7/jobs/43381).

Since astropy 3.1 isn't supported any more, bump the version we run the figure tests with to 3.2 (which I don't think is supported either, but if it doesn't break the tests there's no harm in bumping and it may fix the warning).


dstansby:
Hmm, but if we still support astropy `3.1` it would be good to run the figure tests against that version...

Cadair:
I think @ayshih was planning on bumping to 3.2 for 2.0 anyway in #3919

PR #3932, opened 2020-03-30
Move oldtestdep astropy pin from 3.1 to 3.2

dstansby:
#3925 was merged, bumping the astropy version we test against for figure test. Do the same here for the normal tests.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -45,7 +45,7 @@ deps =
     devdeps: cython
     devdeps,figure_astropydev: git+https://github.com/astropy/astropy
     # Oldest deps we pin against.
-    oldestdeps: astropy<3.2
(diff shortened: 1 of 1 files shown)


Cadair:
I feel like we should be doing `==` style pinnings here, but sure.

dstansby:
> I feel like we should be doing `==` style pinnings here, but sure.

I assumed the point as if a new point release came out we would automatically pick it up this way.

Cadair:
That's fair but doesn't pinning `==3.2` pick up `3.2.2`?

Cadair:
No, it doesn't. Ignore me.

ayshih:
As with #3925, this bump is also in #3919.  Let's merge #3919 already! =)

PR #3936, opened 2020-03-30
Document that Astropy 3.2 is now the minimum required version

### LINKED ISSUE #3935: Document our astropy minimum version bump
#3919 bumped the minimum astropy version for sunpy 2.0. This should be documented.

ayshih:
#3919 bumped our Astropy dependency to 3.2 or later, but it wasn't clearly documented

Fixes #3935

### DIFF
--- a/changelog/3936.breaking.rst
+++ b/changelog/3936.breaking.rst
@@ -1,0 +1,1 @@
+Astropy 3.2 is now the minimum required version of that dependency.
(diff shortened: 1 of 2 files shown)


ayshih:
Even though this PR itself isn't actually "breaking", I classified the changelog entry as "breaking" so that the notion of the version bump is categorized in the right place.

PR #3938, opened 2020-03-31
Improve installation docs

dstansby:
- Add a list of requirements
- Instead of a large wall of text at the top, present readers with two options (main install, or advanced install)
- SunPy > sunpy (SunPy is the project, sunpy is the software package)
- Remove miscellaneous redundant words

### DIFF
--- a/docs/guide/installation/index.rst
+++ b/docs/guide/installation/index.rst
@@ -2,59 +2,63 @@
 Installation
 ============
 
-SunPy relies on and enables the use of the wider ecosystem of scientific Python packages for solar physics.
-Therefore a working SunPy installation is more about installing the scientific Python ecosystem than SunPy itself.
-SunPy v1.X or greater is compatible with Python 3.6+.
-If you still want a Python 2 version, you have to install SunPy v0.9.X.
-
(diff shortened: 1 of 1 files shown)


docs/guide/installation/index.rst
```diff
@@ -2,59 +2,63 @@
...
 If you already have a working Python / Scientific Python environment then you can skip to the :ref:`advanced-install` section.
 
+Requirements
+------------
+
+- python >= 3.6
+- astropy >= 3.2
+
+sunpy v1.X or greater is compatible with Python 3.6+.
+If you need a Python 2 compatible version, you have to install SunPy v0.9.X.
```
nabobalis: Should we advertise this? We don't support it anymore.

dstansby: Good point, especially since 2.0 will be a second python 3+ LTS

docs/guide/installation/index.rst
```diff
@@ -2,59 +2,63 @@
...
+
 .. _main-install:
 
-Installing Scientific Python and SunPy
+Installing Scientific python and sunpy
```
Cadair: Python should be capitalized no?
```suggestion
Installing Scientific Python and sunpy
```

PR #3940, opened 2020-03-31
Add a other issue template

### LINKED ISSUE #3933: Create an Other issue template
Currently the only choices provided are Feature Request and Report a bug but there are things that fall between these two and a default template would be helpful.

ehsteve:
### Description

Creates another issue template for those issues that do not fit in either a feature request or a bug. That way other issues are not created using a totally blank template which does not refer to our code of conduct of checking other existing issues.

Fixes #3933


.github/ISSUE_TEMPLATE/other_issue.md
```diff
@@ -0,0 +1,23 @@
+---
+name: Issue 
+about: Log an issue to improve SunPy!
+---
+
+<!-- This comments are hidden when you submit the issue so you do not need to remove them!
```
dstansby:
```suggestion
<!-- These comments are hidden when you submit the issue so you do not need to remove them!
```

.github/ISSUE_TEMPLATE/other_issue.md
```diff
@@ -0,0 +1,23 @@
+---
+name: Issue 
+about: Log an issue to improve SunPy!
```
dstansby: How is this different to a feature request (which is also a suggestion to improve sunpy)?

ehsteve: For example, a suggestion to remove something from a sub-package is neither a feature request nor a bug fix. I'm sure we can think of others. I just wanted a catch-all so that no one ever uses the blank form.

PR #3951, opened 2020-03-31
Added an HTML summary for Map and MapSequence

ayshih:
This PR adds an HTML summary for both Map and MapSequence, which comes up readily in Jupyter notebooks and elsewhere needs a method call.  If the user clicks on the image, it toggles between coordinate space and pixel space.  The method `quicklook()` could eventually replace `peek()`!

## In a Jupyter notebook
aia_171_roll
aia_1600

## From a terminal (opens in the default web browser)
```python
>>> sunpy.map.Map(sunpy.data.sample.HMI_LOS_IMAGE).quicklook()
```
hmi
```python
>>> sunpy.map.Map(sunpy.data.sample.EIT_195_IMAGE).quicklook()
```
eit

## MapSequence too
For MapSequence, there are Javascript-powered buttons to step through the individual maps.

mapsequence


sunpy/map/mapbase.py
```diff
@@ -243,7 +246,45 @@ def __repr__(self):
                                refpix=u.Quantity(self.reference_pixel),
                                refcoord=u.Quantity((self.reference_coordinate.data.lon,
                                                     self.reference_coordinate.data.lat)),
-                               tmf=TIME_FORMAT) + self.data.__repr__()
+                               tmf=TIME_FORMAT)
+
+    def __repr__(self):
+        return self._text_summary() + "\n" + self.data.__repr__()
```
Cadair: I have been chatting to Dan about this in NDCube recently, and this is a good time to do it here as well. I strongly feel that all our repr's should have the `object.__repr__` at the start, so it's clear what the type is and what the id is.

 ```suggestion
        return object.__repr__(self) + "\n" + self._text_summary() + "\n" + self.data.__repr__()
```

**Comment by nabobalis** [comment]
I like it but I am unsure about the reason behind the pixel histogram?

**Comment by Cadair** [comment]
I love this idea.

I would like to see the text section of this look more html'y and less codey. Also could we see if we could use more rich coordinate information, like observer and frame (both potentially in and out of the html version).

**Comment by dstansby** [comment]
Nice! I like the display of the WCS info side by side with the image.

Agree with @nabobalis, I'm not sure whether it's worth having the histogram. Also not sure what a "histogram normalized image" is? I think we should be just doing very basic showing the data as is. Perhaps worth adding a colorbar too.

File: sunpy/map/mapbase.py (hunk 1 above)

**Comment by ayshih** [comment]
I think the pixel histogram is exceptionally useful to appreciate the distribution of pixel values, particularly to understand whether there are extreme values that are at too few counts to be easily seen in an image (e.g., noise spikes, dead/hot pixels).

> Also not sure what a "histogram normalized image" is?

Histogram equalization is an accepted method of processing an image (see [this example on Wikipedia](https://en.wikipedia.org/wiki/Histogram_equalization#Full-sized_image)).  For an arbitrary image, it provides a balanced representation of the full range of pixel values.  You wouldn't want to do it to produce a science plot, but it's quite appropriate for the purpose here.

**Comment by ayshih** [comment]
> I would like to see the text section of this look more html'y and less codey.

See the updated version

> Also could we see if we could use more rich coordinate information, like observer and frame (both potentially in and out of the html version).

Sure, but that requires more thought than just hacking at what's currently in the text repr.

**Comment by dstansby** [comment]
> Histogram equalization is an accepted method of processing an image (see [this example on Wikipedia](https://en.wikipedia.org/wiki/Histogram_equalization#Full-sized_image)). For an arbitrary image, it provides a balanced representation of the full range of pixel values. You wouldn't want to do it to produce a science plot, but it's quite appropriate for the purpose here.

What do you mean by 'accepted'? I've certainly never heard of it, do you think most solar physicists will have? We don't do this when we `peek` or `plot` a map, so I am -1 on doing it here.

**Comment by dstansby** [comment]
(and in addition it means the same value in two different maps might not show up as the same colour once they are independently histogram normalised)

**Comment by Cadair** [comment]
I think the choice of normalisation here comes right back to what the purpose of this is. The idea here I think isn't to be a `plot()` or even to be a "plot" at all, it's supposed to be much closer to `repr(map.data)` in a way, it lets you see the array. I think hist normalisation is a good way of allowing you to see the data in a predictable and simple way without having to special case things to the dynamic ranges of each instrument etc.

**Comment by dstansby** [comment]
> I think the choice of normalisation here comes right back to what the purpose of this is. The idea here I think isn't to be a `plot()` or even to be a "plot" at all, it's supposed to be much closer to `repr(map.data)` in a way, it lets you see the array. I think hist normalisation is a good way of allowing you to see the data in a predictable and simple way without having to special case things to the dynamic ranges of each instrument etc.

If it's letting you 'see' the array, then shouldn't it just show the array without any transformation on the values? I worry that applying a non-trival transform to the data (that is a function of the data itself, and therefore to me not predictable or simple at all) may end up leading to a lot of confusion and mis-interpretation (whatever the intentions are of the `repr`).

**Comment by dstansby** [comment]
Okay, on second thoughts I think I am coming around to the normalisation (especially since only a grey colourmap is applied)

File: sunpy/map/mapbase.py [hunk 2]
```diff
@@ -243,7 +246,51 @@ def __repr__(self):
...
                                                     self.reference_coordinate.data.lat)),
-                               tmf=TIME_FORMAT) + self.data.__repr__()
+                               tmf=TIME_FORMAT)
+
+    def __repr__(self):
+        return object.__repr__(self) + "\n" + self._text_summary() + "\n" + self.data.__repr__()
+
+    def _repr_html_(self):
+        buf = BytesIO()
+        fig = Figure(figsize=(4.8, 4.8), constrained_layout=True)
```
dstansby, line 256: This and the below line can be condensed into
```
fig, ax = plt.subplots(figsize=(4.8, 4.8), constrained_layout=True)
```
(and similarly below)

sunpy/map/mapbase.py
```diff
@@ -243,7 +246,51 @@ def __repr__(self):
...
+        ax.set_title('Histogram-equalized image')
+        fig.savefig(buf, format='png')
+        image_src = b64encode(buf.getvalue()).decode('utf-8')
+
+        buf = BytesIO()
+        fig = Figure(figsize=(4.8, 2.4), constrained_layout=True)
+        ax = fig.subplots()
+        ax.hist(self.data.ravel(), bins=100, histtype='stepfilled')
+        ax.semilogy()
+        ax.xaxis.get_major_formatter().set_powerlimits((-3, 4))
```
dstansby, line 271: What's the reasoning behind manually setting this?

sunpy/map/mapbase.py
```diff
@@ -243,7 +246,51 @@ def __repr__(self):
...
+        ax.set_xlabel('X pixel')
+        ax.set_ylabel('Y pixel')
+        ax.set_title('Histogram-equalized image')
+        fig.savefig(buf, format='png')
+        image_src = b64encode(buf.getvalue()).decode('utf-8')
+
+        buf = BytesIO()
+        fig = Figure(figsize=(4.8, 2.4), constrained_layout=True)
+        ax = fig.subplots()
+        ax.hist(self.data.ravel(), bins=100, histtype='stepfilled')
```
dstansby, line 269: I think setting bins=100 will fail if any of the map data contains a NaN, so this should probably be `np.linspace(np.nanmin(self.data), np.nanmax(self.data), 100)`.

sunpy/map/mapbase.py
```diff
@@ -243,7 +246,51 @@ def __repr__(self):
...
+        ax.set_xlabel('Pixel value')
+        ax.set_ylabel('# of pixels')
+        fig.savefig(buf, format='png')
+        hist_src = b64encode(buf.getvalue()).decode('utf-8')
+
+        text_to_table = "<table style='text-align:left'><tr><th>" +\
+                        self._text_summary()[20:].replace('\n', '</td></tr><tr><th>').replace(':\t', '</th><td>') +\
+                        "</td></tr></table>"
+
+        html = [
```
dstansby, line 281: Might be a bit nicer to have this as one big block, ie.

```
r"""
repr
goes
here
"""
```

sunpy/map/mapbase.py
ayshih: Perhaps there's a workaround, but it seems that if I create the figure through `pyplot`, the Jupyter notebook will automatically display the created figure, which is not what I want to happen here.

sunpy/map/mapbase.py
ayshih: It's possibly a side effect of using `constrained_layout=True`, but the tick labels were overlapping each other.  According to [the docstring for `ScalarFormatter`](https://matplotlib.org/api/ticker_api.html#matplotlib.ticker.ScalarFormatter.set_powerlimits), this setting of the power limits is an old default.

I welcome suggestions for less hacky alternatives.

sunpy/map/mapbase.py
dstansby: Ah yes, that makes sense.

ayshih:
Bwa ha ha!  I've added a `quicklook()` method to bring up the same summary when not using Jupyter notebook.  It'll create a temporary file and open it in the default web browser.

sunpy/map/mapbase.py
ayshih: NaNs don't appear to cause any issue (at least on my version of Matplotlib).

ayshih: I added an example in the original post for a Map with NaNs.

dstansby: Ah maybe it's infs that I've had problems with in the past

sunpy/map/mapbase.py
```diff
@@ -243,7 +249,66 @@ def __repr__(self):
...
+
+        buf = BytesIO()
+        fig = Figure(figsize=(4.8, 2.4), constrained_layout=True)
+        ax = fig.subplots()
+        ax.hist(self.data.ravel(), bins=100, histtype='stepfilled')
+        ax.semilogy()
+        ax.xaxis.get_major_formatter().set_powerlimits((-3, 4))
+        ax.set_xlabel('Pixel value')
+        ax.set_ylabel('# of pixels')
+        fig.savefig(buf, format='png')
```
Cadair, line 282: I think this is where you can give this arg:
```suggestion
        fig.savefig(buf, format='png', transparent=True)
```

sunpy/map/mapbase.py
ayshih: It turns out that I missed warnings due to the NaNs.  The code now better handles infinites and NaNs.

sunpy/map/mapbase.py
ayshih: `transparent=True` seems to work as desired for the images, but not for this histogram...  I don't know what's going on.

ayshih: It turns out that `transparent=True` was not only working, but working too well and overriding other colors.  Anyway, fixed now.

ayshih:
Lots of updates: WCS plotting, toggling between two images, etc.

@Cadair: I now detect for pyplot use in `plot()`, so no need for an additional keyword.

ayshih:
Running this on an HMI synoptic map (see https://docs.sunpy.org/en/stable/generated/gallery/plotting/hmi_synoptic_maps.html):

syn_map

I guess there aren't enough NaN pixels for them to be rendered in the image.

ayshih:
Can't stop the awesome train: Now MapSequence has joined the fun!

ayshih:
The MapSequence buttons don't appear to work in Internet Explorer.  Investigating...

nabobalis:
I would not support IE11, its EOL and the new edge (aka chrome with some changes) is going to be the new default browser for windows soon enough.

ayshih:
If any heroics (e.g., browser detection) were needed to support IE11, I wouldn't bother.  But, in this case, I just need to convert one object to another type to be able to iterate over it, so I think it's worth doing.  It's also not just IE11, but any web browser version older than a few years (e.g., if a user is on a platform where they can't update the web browser).

ayshih:
I have fixed the interpolation bug that was showing up in the online Sphinx build

ayshih:
It looks like the code as written requires Matplotlib 3.1.0 (released May 2019).  I'll look into whether it's easy to tweak the code to retain support for earlier versions of Matplotlib.

ayshih:
It was an easy fix to recover support for Matplotlib versions back to 2.2.2 (released March 2018).  Our `setup.cfg` claims 2.1.2 as the minimum version, so I bumped it to 2.2.2.  (Incidentally, our `tox.ini` only tests as old as 2.2.5 anyway.)  2.1.2 doesn't work because it doesn't accept the keyword `constrained_layout`.

dstansby:
Calling `fig.tight_layout()` might work as an alternative for older mpl versions?

Cadair:
I am fine with relying on a version of matplotlib from 2+ years ago when 2.0 comes out :+1:

sunpy/map/__init__.py
```diff
@@ -8,3 +11,38 @@
...
 from .mapsequence import MapSequence
+
+
+# If the documentation is being built using Sphinx, append output HTML to the docstsrings for the
+#   GenericMap and MapSequence `quicklook()` methods.
+# This code lives here instead of the respective source files because the source files need to be
+#   fully imported before the following code will successfully run.
+# This code detects a Sphinx build via the simplistic approach of checking if Sphinx has been
+#   imported, and thus there is the potential for false positives.
+if 'sphinx' in sys.modules:
```
Cadair, line 22: This is polluting this namespace a fair bit, might be nice if it were just a private function call.

wafels:
Quick question - for the spherical screen, isn't the distance 'd' between the observer and a point on the screen always the same as the distance between the observer and sun center?

ayshih:
@wafels Wrong issue =)

wafels:
> @wafels Wrong issue =)

Long time since I githubbed.... apologies

dstansby:
Does this work well with a Map that doesn't have all of the metadata displayed in the examples above? (e.g. the map [in this example](https://docs.sunpy.org/en/stable/generated/gallery/map/map_from_numpy_array.html#sphx-glr-generated-gallery-map-map-from-numpy-array-py))

ayshih:
Even in that example, essentially all of the metadata is there, and the HTML version looks fine:

fake

sunpy/map/__init__.py
ayshih: Thinking about this some more, technically this approach doesn't actually pollute the namespace at all.  A user who doesn't import Sphinx won't add any of these new variables and nothing will show up through tab completion.  A Sphinx build does create these variables, but they don't get picked up by our docs, so that pollution is entirely hidden.

Defining a function, even a private one, would pollute the namespace more than what's currently done.  Thoughts?

ayshih: After considering the benefits for future maintenance, I have moved the code to private class methods.

ayshih:
Latest versions:
- https://46163-2165383-gh.circle-artifacts.com/0/docs/_build/html/api/sunpy.map.GenericMap.html#sunpy.map.GenericMap.quicklook
- https://46163-2165383-gh.circle-artifacts.com/0/docs/_build/html/api/sunpy.map.MapSequence.html#sunpy.map.MapSequence.quicklook

sunpy/map/mapsequence.py
```diff
@@ -85,7 +89,154 @@ def __len__(self):
...
+
+        Examples
+        --------
+        >>> from sunpy.map import Map
+        >>> import sunpy.data.sample  # doctest: +REMOTE_DATA
+        >>> seq = Map(sunpy.data.sample.HMI_LOS_IMAGE,
+        ...           sunpy.data.sample.AIA_1600_IMAGE,
+        ...           sunpy.data.sample.EIT_195_IMAGE,
+        ...           sequence=True)  # doctest: +REMOTE_DATA
+        >>> seq.quicklook()  # doctest: +REMOTE_DATA
```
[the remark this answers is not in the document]
ayshih: Heh, whoops!  I'm inclined just to do a doctest skip here since the point of a doctest is to match stated output, and there's no stated output to match.  Testing `quicklook()` (with mocking of the web browser) is already covered by a unit test.

sunpy/map/mapbase.py
```diff
@@ -243,7 +251,160 @@ def __repr__(self):
...
+
+        Clicking on the image to switch between pixel space and coordinate space requires
+        Javascript to be supported by the web browser.
+
+        Examples
+        --------
+        >>> from sunpy.map import Map
+        >>> import sunpy.data.sample  # doctest: +REMOTE_DATA
+        >>> smap = Map(sunpy.data.sample.AIA_171_IMAGE)  # doctest: +REMOTE_DATA
+        >>> smap.quicklook()  # doctest: +REMOTE_DATA
```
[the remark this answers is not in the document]
ayshih: Same as above, I'm going to change this to a skip.

sunpy/map/mapbase.py
```diff
@@ -243,7 +251,160 @@ def __repr__(self):
...
+        # Use a grayscale colormap with histogram equalization (and red for bad values)
+        cmap = cm.get_cmap('gray')
+        cmap.set_bad(color='red')
+        norm = ImageNormalize(stretch=HistEqStretch(finite_data))
+
+        # Plot the image in pixel space
+        fig = Figure(figsize=(5.2, 4.8))
+        # Figure instances in matplotlib<3.1 do not create a canvas by default
+        if fig.canvas is None:
+            FigureCanvasBase(fig)
```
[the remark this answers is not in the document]
ayshih: For context and posterity, here's the PR (matplotlib/matplotlib#12450) where the bejavior was changed.

sunpy/map/mapsequence.py
```diff
@@ -85,7 +89,154 @@ def __len__(self):
...
+        # Assemble the individual HTML repr from each Map, all hidden initally
+        repr_list = [f"<div style='display: none' index={i}>{m._repr_html_()}</div>"
+                     for i, m in enumerate(self.maps)]
+
+        # Unhide the first Map
+        repr_list_html = "\n".join(repr_list).replace('display: none', 'display: ', 1)
+
+        # Return HTML with Javascript-powered buttons
+        # To avoid potential conflicts, the Javascript code does not use any user-defined functions
+        return textwrap.dedent(f"""\
```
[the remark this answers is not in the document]
ayshih: I think that's busy work. =P

I don't know which way is better or worse for maintenance.  At least I'm not adding this giant string to `mapbase.py`!

sunpy/map/mapsequence.py
```diff
@@ -85,7 +89,154 @@ def __len__(self):
...
+        Display a quicklook summary of the MapSequence instance using the default web browser.
+
+        Click on the ``<`` and ``>`` buttons to step through the individual maps.
+
+        Notes
+        -----
+        The image colormap uses
+        `histogram equalization <https://en.wikipedia.org/wiki/Histogram_equalization>`__.
+
+        Interactive elements require Javascript to be supported by the web browser.
```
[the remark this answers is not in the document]
ayshih: I tweaked the sentence so it says "...require Javascript support to be enabled in the web browser."  I actually did encounter an issue when testing in IE because IE was blocking Javascript execution by default.

sunpy/map/mapsequence.py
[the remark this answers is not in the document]
Cadair: yeah fair, I was suggesting it mainly because it moves the js stuff to it's own file, but I am not that bothered by it either way.

PR #3955, opened 2020-04-01
pin pillow

nabobalis:


sunpy_version_pins.txt
```diff
@@ -0,0 +1 @@
+pillow!=7.1.0
```
Cadair, line 1:
```suggestion
pillow!=7.1.0
pytest<5.4
```

nabobalis: Are we broken with pytest?

Cadair: doctestplus is, it causes a warning.

nabobalis: A warning or an error?

Cadair: warning.

nabobalis: Is it worth pinning for a warning?

PR #3956, opened 2020-04-01
Using urlsplit in scraper

abhijeetmanhas:
`urlsplit` used in scraper for `_ftpfileslist` and `filelist` function to make code more interpretable.


abhijeetmanhas:
`ftp.cwd` seems giving error. Making changes accordingly.

nabobalis:
I feel like a change like this needs some sort of extra test for the scrapper.

abhijeetmanhas:
@nabobalis what particular functionality we need to check? I mean this is an internal change using an external library, there is already `test_ftp` function in `test_scraper.py`, which checks only length of the response, adding one more assert to check correctness of string of urls returned will be enough?

nabobalis:
You said that `ftp.cwd` was giving errors, was that picked up by a test? 
If we change how we parse urls, we should make sure we check that the resulting url is still correct.

abhijeetmanhas:
`ftp.cwd` was giving error before its correction commit because I initially passed wrong directory path to it. Now I have corrected it and it no longer gives error.  I checked the argument passed to it is exaclty same as that in master branch. 

So I guess just matching the final urls and returned by ftp and asserting they are as expected shall be enough.

abhijeetmanhas:
> You said that `ftp.cwd` was giving errors, was that picked up by a test?
> If we change how we parse urls, we should make sure we check that the resulting url is still correct.

@nabobalis I have added test to check resulting url is correct. Any more changes?

nabobalis:
Looks fine to me.

nabobalis:
Thanks again @abhijeetmanhas

PR #3960, opened 2020-04-01
Add deprecation warning to aiaprep

### LINKED ISSUE #3282: Remove sunpy.instr.aia.aiaprep
The `aiaprep` function should be removed from the `sunpy.instr` subpackage. Any AIA specific functionality should transition to the new [`aiapy` package](https://gitlab.com/LMSAL_HUB/aia_hub/aiapy).

This is part of the broader goal of moving instrument-specific functionality out of the core sunpy package and into affiliated packages.

wtbarnes:
Fixes #3282. Now that aiapy has had its first release, we can recommend that users use the `register` function in that package 🎉 .

I've put `2.0` in deprecated since and the deprecation decorator appends "scheduled for removal in 3.1" which I think is right?

The message covers multiple lines (its more than 100 characters), but I'd like it to appear as only 1 line in the output to the user. I'm not sure if there is a better way to format that than with the `\`.

### DIFF
--- a/changelog/3960.removal.rst
+++ b/changelog/3960.removal.rst
@@ -1,0 +1,3 @@
+Deprecate `sunpy.instr.aia.aiaprep` in favor of the `register` function in the
+[aiapy](https://gitlab.com/LMSAL_HUB/aia_hub/aiapy) package.
+`sunpy.instr.aia.aiaprep` will be removed in version 2.1
--- a/sunpy/instr/aia.py
+++ b/sunpy/instr/aia.py
@@ -7,10 +7,13 @@ import numpy as np
 import astropy.units as u
 
 from sunpy.map.sources.sdo import AIAMap, HMIMap
+from sunpy.util.decorators import deprecated
 
 __all__ = ['aiaprep']
 
 
+@deprecated("2.0", alternative="`register` in aiapy (https://aiapy.readthedocs.io) for converting \
+AIA images to level 1.5")
 def aiaprep(aiamap):
     """
     Processes a level 1 `~sunpy.map.sources.sdo.AIAMap` into a level 1.5
@@ -18,7 +21,8 @@ def aiaprep(aiamap):
 
     Rotates, scales and translates the image so that solar North is aligned
     with the y axis, each pixel is 0.6 arcsec across, and the center of the
-    Sun is at the center of the image. The actual transformation is done by Map's `~sunpy.map.mapbase.GenericMap.rotate` method.
+    Sun is at the center of the image. The actual transformation is done by Map's
+    `~sunpy.map.mapbase.GenericMap.rotate` method.
 
     This function is similar in functionality to ``aia_prep`` in SSWIDL, but
     it does not use the same transformation to rotate the image and it handles


sunpy/instr/aia.py
```diff
@@ -7,10 +7,14 @@
...
 
 from sunpy.map.sources.sdo import AIAMap, HMIMap
+from sunpy.util.decorators import deprecated
 
 __all__ = ['aiaprep']
 
 
+@deprecated("2.0", message="For converting AIA images to level 1.5, please \
+use the `update_pointing` and `register` functions in aiapy: \
+https://gitlab.com/LMSAL_HUB/aia_hub/aiapy")
```
hayesla, line 17: Do we need to say that this will be removed in future versions of sunpy?

wtbarnes: That gets automatically appended to the end of this message, with the removal version automatically calculated from the `since` version, 2.0 in this case. The output looks like,

```python
>>> m_prep = aiaprep(m)
WARNING: SunpyDeprecationWarning: For converting AIA images to level 1.5, please use the `update_pointing` and `register` functions in aiapy: https://gitlab.com/LMSAL_HUB/aia_hub/aiapy This is scheduled for removal in 3.1. [sunpy.util.decorators]
WARNING: SunpyUserWarning: Input data has been cast to float64. [sunpy.image.transform]
```

Cadair:
> I've put 2.0 in deprecated since and the deprecation decorator appends "scheduled for removal in 3.1" which I think is right?

I don't actually think this is right (sounds like a bug in the decorator logic) if we deprecate it in 2.0 that's our "one LTS release" so we could actually remove it in 2.1?

changelog/3960.removal.rst
```diff
@@ -0,0 +1 @@
+Add `deprecated` decorator to `sunpy.instr.aia.aiaprep` in preparation for removal in 3.1
```
Cadair, line 1: This should be user facing, so it should explain what's been deprecated, what the alternative is and when it's going away.

sunpy/instr/aia.py
Cadair, line 17: Would it be better to link to rendered install / getting started docs here?

Cadair, line 17:
```suggestion
https://gitlab.com/LMSAL_HUB/aia_hub/aiapy .")
```

ayshih:
> I don't actually think this is right (sounds like a bug in the decorator logic) if we deprecate it in 2.0 that's our "one LTS release" so we could actually remove it in 2.1?

Ugh, there's actually more than one bug in the decorator logic.  I'll create a separate issue.

wtbarnes:
#3982 should be merged first

changelog/3960.removal.rst
```diff
@@ -0,0 +1,4 @@
+Add `sunpy.util.decorators.deprecated` decorator to `sunpy.instr.aia.aiaprep`
+This function is being deprecated in favor of the `register` function in the
```
Cadair, line 2:
```suggestion
Deprecate `sunpy.instr.aia.aiaprep` in favor of the `register` function in the 
```

Issue #2690, opened 2018-07-15
Should sunpy.map.Map be able to take file handlers and a filetype like Timeseries


Cadair:
<EXAMPLE TO COME>

Cadair:
No I think I mean like this:
```python
with open("filename.fits") as fd:
    sunpy.map.Map(fd)
```

devanshshukla99:
Hey @Cadair, I've added support for file-handlers in both `map_factory` and `timeseries_factory`.
However, since both ana and jp2 don't intrinsically support file-like objects, they will still extract path from the object.
Any suggestion on how to improve this further?

nabobalis:
Hi @devanshshukla99, it is hard to say without looking at the code. Can you open a pull request?

devanshshukla99:
Yeah, will do.

amishgog:
Hey @Cadair I was working on this issue and when I directly entered the file name into the map function it works correctly. I think the problem lies in opening a FITS file with "open()" . I tried this with TimeSeries as well but it shows a different error saying "Input read by sunpy.io can not find a matching class for reading multiple HDUs". Is the TimeSeries function having trouble reading multiple HUDs?. Can you give some tips on how we can address that.

PredictiveManish:
Is this issue has been resolved??

nabobalis:
I see no pull request that closes this issue linked.

PredictiveManish:
Outdated?

nabobalis:
I think we still want to support the original issue.

wtbarnes:
Running the code in the original issue, I get a fun new exception,

```python traceback
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
Cell In[41], line 4
      2 import sunpy.map
      3 with open(sp.AIA_171_IMAGE) as fd:
----> 4     sunpy.map.Map(fd)

File [~/Documents/codes/sunpy/sunpy/map/map_factory.py:293](http://localhost:8888/lab/workspaces/auto-I/tree/Documents/projects/active-region-heating-properties/~/Documents/codes/sunpy/sunpy/map/map_factory.py#line=292), in MapFactory.__call__(self, composite, sequence, allow_errors, *args, **kwargs)
    267 def __call__(self, *args, composite=False, sequence=False, allow_errors=False, **kwargs):
    268     """Method for running the factory. Takes arbitrary arguments and
    269     keyword arguments and passes them to a sequence of pre-registered types
[...]

PR #3963, opened 2020-04-02
Allow Map to use i/o objects

dstansby:
xref #2690

### DIFF
--- a/sunpy/map/map_factory.py
+++ b/sunpy/map/map_factory.py
@@ -3,6 +3,7 @@ from collections import OrderedDict
 import warnings
 import pathlib
 import glob
+import io
 
 import numpy as np
 import astropy.io.fits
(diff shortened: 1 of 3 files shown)


sunpy/map/map_factory.py
```diff
@@ -189,6 +190,11 @@ def _parse_args(self, *args, **kwargs):
 
             arg = args[i]
 
+            # File object
+            if isinstance(arg, io.IOBase):
+                # Extract the name and then rely on the filename logic below
+                arg = arg.name
```
Cadair, line 196: If we are doing this, is there really any point?

dstansby: What do you mean by "is there any point"?

Cadair: Well we aren't using the file object, or respecting the context manager. We will be keeping our own file handle open to the fits file once the context manager is closed or once `f.close()` was called, which to me is us outright lying to the user lol

dstansby: Ha, good point... it's going to need a bit more code wrangling then to do it properly.

Cadair:
I do like the idea of doing this though, I just think we need to pass the file handle through the whole stack, which I appreciate is a bigger effort.

dstansby:
Yep 👍

PR #3967, opened 2020-04-02
Fix astropy pinning

dstansby:
astropy released a new version with a different version number but no actual other changes, so we have to pin against the new version too. Should fix the doc build... Does this need backporting?


dstansby:
Wooo time to unbreak

dstansby:
Does this need backporting?

nabobalis:
Yeah

nabobalis:
THIS SHOULD WORK YOU STUPID BOT

PR #3971, opened 2020-04-02
Update figure test artefact upload directory

dstansby:
I think since we started using tox this has been broken. This should now point to the correct directory to upload the test figures.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,10 +88,11 @@ jobs:
       - run: tox -e figure
       - run: codecov
       - store_artifacts:
-          path: figure_test_images
+          path:  .tmp/figure/figure_test_images
+
       - run:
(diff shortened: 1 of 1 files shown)


nabobalis:
I can't believe I updated the docs one and forgot to do the figure tests.

nabobalis:
The url doesn't work for me still. I think we need to cut more.

nabobalis:
`https://44568-2165383-gh.circle-artifacts.com/0/.tmp/figure/figure_test_images/fig_comparison.html`
is what it should be, I forgot I had to remove `${CIRCLE_WORKING_DIRECTORY/#\~/$HOME}/` for the docs

dstansby:
Ah I didn't check that, but at least the artefact upload now works via. the artefact tab

PR #3974, opened 2020-04-02
Sorted response tables in vso

### LINKED ISSUE #3922: VSO Results from RHESSI seem to not have consistent ordering

abhijeetmanhas:
### Description

Fixes #3922 . I have added a new function `iter_sort_responses` in vso so that always response tables are sorted by start time. Let me know about new tests/ addons or changes.


abhijeetmanhas:
Don't know why tests are failing. Can someone please re-run them? It shows internal error on azure

abhijeetmanhas:
PIL is responsible for errors, as I checked.

nabobalis:
I would like a quick test for that function just to make sure it works and maybe a documentation string.

abhijeetmanhas:
Sure @nabobalis adding a function to check and adding a docstring too

sunpy/net/vso/vso.py
```diff
@@ -188,7 +195,8 @@ def search(self, *query):
 
     @classmethod
     def create(cls, queryresult):
-        return cls(list(iter_records(queryresult)), queryresult)
```
Cadair, line 191: Do we use this `iter_records` anywhere else? If not can we remove it, if so should we be sorting that as well?

sunpy/net/vso/vso.py
```diff
@@ -77,6 +77,13 @@ def iter_records(response):
         yield from prov_item.record.recorditem
 
 
+def iter_sort_response(response):
+    for prov_item in response.provideritem:
+        if not hasattr(prov_item, 'record') or not prov_item.record:
+            continue
+        yield from sorted(prov_item.record.recorditem, key=lambda x: x.time.start)
```
Cadair, line 84: If I am reading this right this is assuming that if a `provideritem` block has a `record` subblock, then that has a `recorditem` and that has a `time` with a `start`.

I might just run this past the VSO team to make sure this is a safe assumption.

Cadair:
Looks pretty good @abhijeetmanhas if you rebase this onto master the CI should pass.

sunpy/net/vso/vso.py
nabobalis: If we sort in the original method? Keyword arg to specific the type of sorting?

sunpy/net/vso/vso.py
abhijeetmanhas: I just checked it for 3-4 types of VSO queries, and it didn't failed so assumed. It would be great if it can be confirmed with VSO if it is the case always.

sunpy/net/vso/vso.py
abhijeetmanhas: No `iter_records` is not used anywhere, except it was used earlier in `create` which I changed to `iter_sort_responses()` . @nabobalis I didn't understand type of sorting. Is it like ascending or descending? Or it is just sort or not sort(keep the default)?

sunpy/net/vso/vso.py
abhijeetmanhas: @Cadair I can't find the reference in their docs.

abhijeetmanhas: I just checked that we do assume in `merge` that `record` will have a `recorditem`, and in their docs it is written that record item will have `time` with `start` in it. Just not sure that will `provideritem` will have `record` in it. Should I assume it will have? Or just sort by any algorithm and use continue with `hasattr(record)` ?

abhijeetmanhas:
I have removed `iter_records` and added a mocktest for `iter_sort_response` function, and docstring for the same. Have yet to make changes in docs of `fido.rst` having sorted records.

sunpy/net/vso/vso.py
```diff
@@ -70,11 +70,24 @@ def _parse_date(string):
...
+    Sorts the VSO queryresults by their start time.
+
+    Parameters
+    ----------
+    response : `zeep.objects.QueryResponse`
+        A SOAP Object of a VSO queryresult
+    Returns
+    -------
+
+    `generator`
```
Cadair, line 84:
```suggestion
        A SOAP Object of a VSO queryresult
        
    Returns
    -------
    `generator`
```

sunpy/net/vso/vso.py
jhourcle: `provideritem` might not have a `record` subblock, or it might have an empty block.

The main case when this would happen is if there's an error message from the provider, and use that mechanism to return a warning message from the JSOC when it switches into 'summary row' mode because you asked for something that matched too many records.

You should be able to trigger it by asking for 3 hours worth of AIA data.  Here's an example from IDL:

    IDL> a = vso_search( '2011-01-01','2011-01-01 03:00', inst='aia' );
    Records Returned : JSOC : 0/
        JSOC : Large Query (over 5000 records); Summary rows returned; Each record describes multiple observations
    Records Returned : JSOC : 7202/7202
    IDL> help, a
    A               STRUCT    = -> VSORECORD Array[174]

There were talks in years past about including model results in the VSO, and so it's possible that in the future, not all data will have a time associated with it.  (but I lean towards having it like 'quicklook' data where you have to specifically request it if we don't put it as a completely different method to keep it isolated from observational data)

jhourcle: I should mention -- that return structure should've had two `provideritem`s, one with no records, the other with 174.

Cadair: Thanks for the input @jhourcle 

> provideritem might not have a record subblock, or it might have an empty block.

Cool, I think we handle this.

Maybe it is worth just guarding this whole block in an `except AttributeError` so that if there isn't a time we just return the results unsorted?

abhijeetmanhas: @Cadair that would mean that if any single record doesn't have time, or record is empty, then entire result is unsorted, (default as it was earlier). If it is Ok then I will do it using try except on entire yield.

jhourcle: > Thanks for the input @jhourcle
> 
> > provideritem might not have a record subblock, or it might have an empty block.
> 
> Cool, I think we handle this.
> 
> Maybe it is worth just guarding this whole block in an `except AttributeError` so that if there isn't a time we just return the results unsorted?

If it were me, I'd assume some default time ... either "now" or the epoch, so they end up either sorting to the end of the list, or the beginning of the list, rather than having the sort fail entirely.  Currently, all items under `record` should have a start time and end time.  (but I'll leave it for the scientists to argue if sorting by start time or the middle of the observation is best)

Cadair: defaulting to "now" (only for sorting) seems reasonable, and start time is good enough rather than calculating the middle :laughing:

abhijeetmanhas: Seems reasonable. Ok making these changes now.

abhijeetmanhas: Have made the required checks and changes. Can check.

sunpy/net/vso/vso.py
```diff
@@ -71,11 +71,39 @@ def _parse_date(string):
...
+
+    Parameters
+    ----------
+    response : `zeep.objects.QueryResponse`
+        A SOAP Object of a VSO queryresult
+
+    Returns
+    -------
+    `list`
+    Sorted recorditems w.r.t. their start time.
```
nabobalis, line 86:
```suggestion
        Sorted record items w.r.t. their start time.
```

sunpy/net/vso/vso.py
```diff
@@ -71,11 +71,39 @@ def _parse_date(string):
...
+            continue
+        rec_item = prov_item.record.recorditem
+        for rec in rec_item:
+            if hasattr(rec, 'time'):
+                if hasattr(rec.time, 'start') and rec.time.start is not None:
+                    has_time_recs.append(rec)
+                else:
+                    has_notime_recs.append(rec)
+            else:
+                has_notime_recs.append(rec)
```
Cadair, line 103: I think you can collapse this into one block:
```suggestion
            if hasattr(rec, 'time') and hasattr(rec.time, 'start') and rec.time.start is not None:
                has_time_recs.append(rec)
            else:
                has_notime_recs.append(rec)
```

docs/guide/acquiring_data/fido.rst
```diff
@@ -134,8 +135,8 @@ operator would::
                                                    ...                  keV
```
Cadair, line 135: this example has a `+SKIP` on it, which I hope can be removed now?

changelog/3974.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed a bug of inconsistent ordering of results in `~sunpy.net.vso.vso.QueryResponse` by sorting the results by start time.
```
Cadair, line 1:
```suggestion
Fixed a bug of where results of VSO searches would have inconsistent ordering in `~sunpy.net.vso.vso.QueryResponse` by always sorting the results by start time.
```

docs/guide/acquiring_data/fido.rst
abhijeetmanhas: yep

abhijeetmanhas:
Hope the tests shall pass, after rebasing and cleaning the git history.

nabobalis:
Thanks again @abhijeetmanhas

Issue #2384, opened 2017-12-18
Post Python 2: Refactor Map and Timeseries factory to use singledispatch in `__call__`


Cadair:
This would remove the large amount of nested if statements, and also allow user code and other packages to add extra argument parsing to the factories.

[singledispatch](https://docs.python.org/3/library/functools.html#functools.singledispatch)

*This can not be done until after 0.9*

talsperre:
Hi @Cadair. 
I was looking at this issue and wanted to work on it. Are you referring to the nested if statements in ```_parse_args``` that is called in ```__call__``` function? Some of the conditionals here require the use of ```self``` i.e. class methods such as:
```
# A URL
elif (isinstance(arg, six.string_types) and
      _is_url(arg)):
    url = arg
    path = download_file(url, get_and_create_download_dir())
    pairs = self._read_file(path, **kwargs)
    data_header_pairs += pairs
```
 How should ```functools.singledispatch``` be used with instance methods?  Would something like [this](https://stackoverflow.com/a/24602374) do?

PR #3977, opened 2020-04-03
Sanitize args list in map factory

dstansby:
I tried making a start on using singledispatch for the map factory (#2384), and it ended up as a bit of a mess. This PR is some refactoring that is a pre-requisite to using singledispatch.

Before looping through all the args, first find the data, header pairs, and put them in a tuple. This then allows a `for arg in args` loop, instead of the previous `while` loop and index bookkeeping. Now (almost) everything directly under `for arg in args` is a `if isinstance(..., arg)` statement.


nabobalis:
Rebase this and if the CI passes, it's good to go?

PR #3979, opened 2020-04-03
Helper method for checking against required and optional query attrs

dstansby:
The logic for checking a query in `jsoc` seems quite common, and I need to use similar logic in the SPDF client I am writing. It seems sensible to put it in a helper method. Questions

I'm not sure what it should be called? (`simple_can_handle_query` is a bit of a rubbish name...)

### DIFF
--- a/changelog/3979.trivial.rst
+++ b/changelog/3979.trivial.rst
@@ -1,0 +1,3 @@
+Added `sunpy.net.base_client.BaseClient.check_attr_types_in_query` as a helper method
+to check if a query contains a set of required attributes, and is a subset of optional
+attributes.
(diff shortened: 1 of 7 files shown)


nabobalis:
What about `query_handler`?

Cadair:
I think this would be useful as a helper as a part of a more complex `_can_handle_query`, i.e. you could call this first and then check specific values of attrs later having verified the types are there with this, so I think the name should reflect that as well. (Also I think this pattern might be useable in some of the dataretriever clients as well, maybe you could check?)

So maybe something like `check_attr_types_in_query`?

PR #3981, opened 2020-04-03
Add AttrWalker __repr__

dstansby:
Prints something like
```
<sunpy.net.attr.AttrWalker object at 0x1232f99d0>
Registered creators: [<class 'object'>, <class 'sunpy.net.attr.AttrOr'>, <class 'sunpy.net.attr.AttrAnd'>]
Registered appliers: [<class 'object'>, <class 'sunpy.net.attrs.Time'>, <class 'heliopy.data.cdaweb.attrs.Dataset'>]
```

### DIFF
--- a/sunpy/net/attr.py
+++ b/sunpy/net/attr.py
@@ -445,8 +445,8 @@ class AttrWalker:
 
     """
     def __repr__(self):
-        creators = [key for key in self.createmm.registry.keys()]
-        appliers = [key for key in self.applymm.registry.keys()]
+        creators = list(self.createmm.registry.keys())
+        appliers = list(self.applymm.registry.keys())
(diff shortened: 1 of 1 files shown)


sunpy/net/attr.py
```diff
@@ -444,6 +444,13 @@ class AttrWalker:
       type and return a different one.
 
     """
+    def __repr__(self):
+        creators = [key for key in self.createmm.registry.keys()]
+        appliers = [key for key in self.applymm.registry.keys()]
```
Cadair, line 449:
```suggestion
        creators = list(self.createmm.registry.keys())
        appliers = list(self.applymm.registry.keys())
```

dstansby:
Should probably have added a changelog to this.

PR #3990, opened 2020-04-06
Fix the automatic creation of slider labels from WCS

Cadair:
### Description

In the situation where a WCS is passed to `ArrayAnimatorWCS` with more world dimensions than pixel dimensions our automatic calculation of the labels for the sliders failed. This patch robustly calculates them based on the axis_correlation_matrix.


Cadair:
Looks like my labels have changed ordering, I wonder if they were wrong before or if they are wrong now :thinking:

Cadair:
done

PR #3993, opened 2020-04-06
Updated VSOClient's response_block_properties method

### LINKED ISSUE #2574: JSOC response_block_properties is empty
Once PR #2188 is merged. 

JSOC requests returns an empty response_block_properties as can be seen [here.](https://github.com/sunpy/sunpy/pull/2188/files#diff-61f5137ce8bef90efb1dd91756542acbR83)

So the fundamental problem with response_block_properties appears to be in the way that JSOC separates it's queries and it's downloads. We need to work out a  way of mapping from the JSOCResponse(query) object to the records that are actually downloaded by the downloader.

DShivansh:
implemented response_block_properties method of jsoc, optimised response_block_properties of vso and also fixed its bug.

### DIFF
--- a/changelog/3993.bugfix.rst
+++ b/changelog/3993.bugfix.rst
@@ -1,0 +1,1 @@
+Updated VSOClient.response_block_properties to check if "None" is in the return.
--- a/sunpy/net/vso/vso.py
+++ b/sunpy/net/vso/vso.py
@@ -303,9 +303,12 @@ class QueryResponse(BaseQueryResponse):
         """
         s = {a if not a.startswith('_') else None for a in dir(self[0])}
         for resp in self[1:]:
+            if len(s) == 0:
+                break
             s = s.intersection({a if not a.startswith('_') else None for a in dir(resp)})
 
-        s.remove(None)
+        if None in s:
+            s.remove(None)
         return s
 
 
--- a/sunpy/net/dataretriever/sources/norh.py
+++ b/sunpy/net/dataretriever/sources/norh.py
@@ -26,6 +26,7 @@ class NoRHClient(GenericClient):
 
     Examples
     --------
+    >>> import astropy.units as u
     >>> from sunpy.net import Fido, attrs as a
     >>> results = Fido.search(a.Time("2016/1/1", "2016/1/2"),
     ...                       a.Instrument.norh, a.Wavelength(17*u.GHz))  #doctest: +REMOTE_DATA
--- a/sunpy/net/vso/tests/test_vso.py
+++ b/sunpy/net/vso/tests/test_vso.py
@@ -538,3 +538,11 @@ def test_vso_repr(client):
     """
     output = str(client)
     assert output[:50] == 'sunpy.net.vso.vso.VSOClient\n\nAllows queries and do'
+
+
+@pytest.mark.remote_data
+def test_response_block_properties(client):
+    res = client.search(a.Time('2012/3/4', '2012/3/6'), a.Instrument('aia'), a.Wavelength(171 * u.angstrom),
+                        a.vso.Sample(10 * u.minute))
+    properties = res.response_block_properties()
+    assert len(properties) == 0


sunpy/net/jsoc/jsoc.py
```diff
@@ -90,9 +90,21 @@ def append(self, table):
...
-        return set()
+        s = {a if not a.startswith('_') else None for a in dir(self[0])}
+        for resp in self[1:]:
+            if len(s) == 0:
+                break
+            s = s.intersection({a if not a.startswith('_') else None for a in dir(resp)})
+
+        if None in s:
+            s.remove(None)
+        return s
```
nabobalis, line 107: This will need a test.

DShivansh: VSO one also does not have any test for this.

nabobalis: Then one should be added for both.

Cadair:
This approach doesn't work for JSOC as the form of the `QueryResponse` object is very different for VSO. 

More fundamentally, I think this method can never be implemented for the JSOC due to limitations in the way search and fetch work together.

To explain, the purpose of this method is to show people the things they can specify as format strings in the filenames generated by the `fetch()` method. I think the fundamental issue with the JSOC results is the way fetch works. We can only fetch the whole thing the user searches for not individual records. So if I remember correctly we can't associate files that we download (and therefore generate filenames for) with specific records in the search results. This means that the `response_block_properties` method for the JSOC client could never work as we could never hook the string interpolation into the `fetch` method.

sunpy/net/vso/vso.py
```diff
@@ -271,9 +271,12 @@ def response_block_properties(self):
         """
         s = {a if not a.startswith('_') else None for a in dir(self[0])}
         for resp in self[1:]:
+            if len(s) == 0:
```
Cadair, line 274: These fixes look good :+1:

nabobalis:
So the original issue should be closed?

Cadair:
Well the subtlety in the original issue is:

> We need to work out a way of mapping from the JSOCResponse(query) object to the records that are actually downloaded by the downloader.

If we could do this then that would be great, but that isn't straight forward and probably needs someone to work with the JSOC team to see if it's actually possible.

DShivansh:
ok

sunpy/net/vso/vso.py
DShivansh: should I push another commit that contains changes only in vso file and remove the changes in JSOC.
```python
s.remove(None) # removing none without checking
```
the above line tries to remove None without checking whether set contains None or not which throws KeyError so it is a bug.

Cadair: yeah that would be great :)

Cadair, line 274: Could this check actually be moved up outside of the for loop?

```python
s = ...
if not s:
    return
```

DShivansh: no, because at first, the set will contain some elements it's after the first intersection that set may become empty.

DShivansh: no, because set may not be empty at the start but may become empty after the first or second iteration.

changelog/3993.bugfix.rst
```diff
@@ -0,0 +1 @@
+optimised response_block_properties of vso and also fixed its bug of removing None from set without checking if None is present in it or not.
```
nabobalis, line 1: Updated `VSOClient.response_block_properties`  to check if "None" is in  the return.

sunpy/net/vso/vso.py
```diff
@@ -271,9 +271,12 @@ def response_block_properties(self):
...
+                break
             s = s.intersection({a if not a.startswith('_') else None for a in dir(resp)})
 
-        s.remove(None)
+        if None in s:
+            s.remove(None)
         return s
```
nabobalis, line 280: I think this needs a test.

sunpy/net/tests/test_fido.py
```diff
@@ -489,3 +489,10 @@ def test_slice_jsoc():
 
     with pytest.warns(SunpyUserWarning) as record:
         Fido.fetch(res[0, 0:1])
+
+
+@pytest.mark.remote_data
+def test_response_block_properties():
+    res = Fido.search(a.Time('2012/3/4', '2012/3/6'), a.Instrument('aia'), a.Wavelength(171 * u.angstrom),
+                      a.vso.Sample(10 * u.minute))
+    assert len(res) == 0
```
nabobalis, line 498: How does this test the `response_block_properties` function?

DShivansh: sorry I forgot to use the function.

sunpy/net/tests/test_fido.py
```diff
@@ -489,3 +489,11 @@ def test_slice_jsoc():
...
     with pytest.warns(SunpyUserWarning) as record:
         Fido.fetch(res[0, 0:1])
+
+
+@pytest.mark.remote_data
+def test_response_block_properties():
+    res = Fido.search(a.Time('2012/3/4', '2012/3/6'), a.Instrument('aia'), a.Wavelength(171 * u.angstrom),
+                      a.vso.Sample(10 * u.minute))
+    properties = res.response_block_properties()
+    assert len(properties) == 0
```
nabobalis, line 499: Why is this zero? What does the return from this function look like if there is actual data? 

I think also, since this is a test for a VSO Client method, the test should live in that file.

DShivansh: output of this function is a set of properties common to all the rows,
I included this test inside Fido test because if I included it inside test_vso then I had to import Fido inside test_vso just for the one test so I thought of including the test inside the test_Fido file.

nabobalis: Can't you use the vso client directly instead of fido?

DShivansh: No, because this function works on unifiedResponse which is returned by FIDO.

nabobalis: Is the method not implemented on the VSOClient?

DShivansh: yes, I checked and it worked with vso query response. I'll change the location of the test case.

sunpy/net/tests/test_vso.py
```diff
@@ -505,3 +505,11 @@ def test_incorrect_content_disposition(client):
     ])
 def test_can_handle_query(query, handle):
     assert VSOClient._can_handle_query(*query) is handle
+
+
+@pytest.mark.remote_data
+def test_response_block_properties():
```
Cadair, line 511:
```suggestion
def test_response_block_properties(client):
```

dstansby:
Could you rebase this on to the current master branch? Otherwise this looks good to me.

nabobalis:
Online fails are due to HELIO.

nabobalis:
Thanks again @DShivansh

PR #3994, opened 2020-04-06
Removed xfail label in scraper test & enabled to scrap local files

abhijeetmanhas:
`test_scraper.py` earlier had a test labelled `@pytest.mark.xfail`, so removed it. And added a new method `_localfilelist` to scrap local data files.
The method is called for `file://` protocols.

### DIFF
--- a/changelog/3994.feature.rst
+++ b/changelog/3994.feature.rst
@@ -1,0 +1,1 @@
+Added `_localfilelist` method in `~sunpy.util.scraper.Scraper` to scrap local data archives.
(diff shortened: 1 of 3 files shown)


abhijeetmanhas:
It is failing because file system in Windows is different from linux or Unix. Need to check whether sunpy rootdir need to be changed or some os specific changes in my function.

abhijeetmanhas:
MacOS azure is stuck at installing tox.

changelog/3994.feature.rst
```diff
@@ -0,0 +1 @@
+Added `_localfilelist` method in `~sunpy.util.scraper.Scraper` to scrap local data archives.
```
dstansby, line 1: This isn't a user facing change, so not sure it needs to be a changelog. Is this fixing a bug? If so I would mention the bug that has been fixed here instead.

nabobalis: It feels like a mix of the two.

nabobalis:
Thanks again @abhijeetmanhas

PR #3995, opened 2020-04-07
Cleanup Dependencies and Re-enable asdf schema tests

Cadair:
### Description

Cleanup pins in the setup.cfg file we don't need after reproject 0.7.

Also clean up things which are deps of pytest astropy and also re-enable asdf schema tests.


Cadair:
@nabobalis did we backport the astropy pins? If so we should backport this right?

nabobalis:
Sure. We might need to do a manual pass over the 1.0 and 1.1 branch anyway.

PR #4002, opened 2020-04-08
Automatically upload reference figures to the repo

Cadair:
Also a CI cleanup

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -60,7 +60,7 @@ jobs:
       - linux: py38-online
         name: py38_test_online
 
-      - linux32: py38
+      - linux32: py38-32bit
         name: py38_test_32bit
 
       - linux: codestyle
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -92,5 +93,6 @@ jobs:
         - py36_test
         - py37_test
         - py38_test_online
+        - py38_test_32bit
         - py36_test_oldestdeps
```
nabobalis, line 97: Do we need two 36 tests?

Cadair: the other one is macos

Cadair: You think one should be 3.7?

nabobalis: Or remove the first 36 test?

Cadair: that's the macos build.

nabobalis: Run oldest deps on macos then?

PR #4004, opened 2020-04-08
Take 2 on auto figure uploads

Cadair:
This time from a branch on sunpy/sunpy so I can get the ssh key and test it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,3 +1,5 @@
+version: 2.1
+
 no-backports: &no-backports
   name: Skip any branches called cherry-pick
   command: |
@@ -24,7 +26,6 @@ apt-run: &apt-install
     apt update
     apt install -y graphviz build-essential libopenjp2-7
 
-version: 2
 jobs:
   twine-check:
     docker:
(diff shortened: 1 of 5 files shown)


azure-pipelines.yml
```diff
@@ -72,7 +76,6 @@ jobs:
         name: py36_test_oldestdeps
 
       - linux: py38-devdeps
-        name: py38_test_devdeps
```
nabobalis, line 75: why no name?

.circleci/config.yml
```diff
@@ -145,25 +89,76 @@ jobs:
...
+      - run: git clone git@github.com:sunpy/sunpy-figure-tests.git -b sunpy-test ~/sunpy-figure-tests/
+
+      - run: cp -r .tmp/figure_py36/figure_test_images/*.png ~/sunpy-figure-tests/figures/figure_py36/
+
+      - run: cp -r .tmp/figure_py37_astropydev/figure_test_images/*.png ~/sunpy-figure-tests/figures/figure_py37_astropydev/
+
+      - run: git config --global user.email "sunpy@circleci" && git config --global user.name "SunPy Circle CI"
+
+      - run: cd ~/sunpy-figure-tests/; git add . && git commit -m "Update reference figures from ${CIRCLE_BRANCH}" && git push
+
```
nabobalis, line 139: Why all these line breaks?

azure-pipelines.yml
Cadair: it doesn't need a name because we don't reference it in the upload task later.

nabobalis: But it won't have a nice name?

Cadair: the name is not used anywhere else iirc

.circleci/config.yml
```diff
@@ -145,25 +89,83 @@ jobs:
...
+      - deploy-reference-images:
+          requires:
+            - "figure_py36"
+            - "figure_py37_astropydev"
+          filters:
+            branches:
+              only:
+                - master
+                # - 1.1
+                # - 1.0
```
nabobalis, line 167: Will this be enabled?

azure-pipelines.yml
nabobalis: Is it not in the GUI?

Cadair: No, I don't think so the template generates that from the URL.and tox env iirc

.circleci/config.yml
Cadair: Yeah, we need to create the branch on the repo first and then back port this.

Cadair:
Need to fix the figure comparison page I think.

dstansby:
I've unsubscribed from this pr, but please do ping me when it's ready for review

Cadair:
@dstansby I think this is ready for review. I am not 100% sure it's working, but I can't really tell until it's on master now.

nabobalis:
You want to remove that blank test file and the figure comparison can't get the old images (I assume its a non master branch issue?)

azure-pipelines.yml
```diff
@@ -72,7 +79,6 @@ jobs:
...
```
dstansby, line 75: Why has this gone?

Cadair: See https://github.com/sunpy/sunpy/pull/4004#discussion_r405604079 TL;DR it isn't used for anything.

Cadair:
The wheels/sdists are broken because of the PIL issue, and we don't have the same pinning things. They are only showing up here because this PR is a branch on sunpy/sunpy

PR #4005, opened 2020-04-08
Miscellanous cleanup in coordinates now that we depend on Astropy 3.2+

ayshih:
Now that the minimum required version of Astropy is 3.2, we can do some cleanup in `sunpy.coordinates`.

~Curiously, some of the functions are now returning different values (this is technically the first time I'm making use of Astropy's `GeocentricTrueEcliptic`).  I'm investigating, but it's possibly a bug in Astropy.~  Removed any use of `GeocentricTrueEcliptic` from this PR pending further thought.

Addresses part of #3154


nabobalis:
No I don't think so, since the bump was made in another PR.

Issue #4006, opened 2020-04-08
sunpy.map.Map() fails to load list of urls


fluxtransport:
### Description
sunpy.map.Map() fails to load list of urls. 
e.g
urls = ['http://jsoc.stanford.edu/SUM80/D136597189/S00000/image_lev1.fits'
 'http://jsoc.stanford.edu/SUM79/D136597240/S00000/image_lev1.fits'
 'http://jsoc.stanford.edu/SUM76/D136597241/S00000/image_lev1.fits'
 'http://jsoc.stanford.edu/SUM82/D136597237/S00000/image_lev1.fits'
 'http://jsoc.stanford.edu/SUM83/D136597187/S00000/image_lev1.fits'
 'http://jsoc.stanford.edu/SUM81/D136597190/S00000/image_lev1.fits']

maps = Map(urls)

### Expected behavior
Expected to return a list of maps objects. 

### Actual behavior
Files Downloaded: 100%|████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:01<00:00,  1.36s/file]
Files Downloaded: 100%|████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:01<00:00,  1.47s/file]
---------------------------------------------------------------------------
```python                                                                    
AttributeError                            Traceback (most recent call last)
<ipython-input-23-1a0872c98bf5> in <module>
----> 1 maps = Map(urls)

~/anaconda3/lib/python3.7/site-packages/sunpy/map/map_factory.py in __call__(self, composite, sequence, silence_errors, *args, **kwargs)
    274         """
    275 
--> 276         data_header_pairs, already_maps = self._parse_args(*args, **kwargs)
    277 
    278         new_maps = list()

~/anaconda3/lib/python3.7/site-packages/sunpy/map/map_factory.py in _parse_args(self, *args, **kwargs)
    215             elif isinstance(arg, str) and _is_url(arg):
    216                 url = arg
--> 217                 path = str(cache.download(url).absolute())
    218                 pairs = self._read_file(path, **kwargs)
    219                 data_header_pairs += pairs

AttributeError: 'str' object has no attribute 'absolute'
```
### Steps to Reproduce
```python
from sunpy.map import Map

urls = ['http://jsoc.stanford.edu/SUM80/D136597189/S00000/image_lev1.fits',
 'http://jsoc.stanford.edu/SUM79/D136597240/S00000/image_lev1.fits',
[...]

PR #4007, opened 2020-04-09
fix for str return from the cache

nabobalis:
It works but its a fudge.

Fixes https://github.com/sunpy/sunpy/issues/4006

### DIFF
--- a/sunpy/map/tests/test_map_factory.py
+++ b/sunpy/map/tests/test_map_factory.py
@@ -250,3 +250,14 @@ class TestMap:
         assert isinstance(xrt, sunpy.map.sources.XRTMap)
 
     # TODO: Test SXTMap
+
+
+@pytest.mark.remote_data
+def test_map_list_urls_cache():
+    """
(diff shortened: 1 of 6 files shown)


sunpy/map/tests/test_map_factory.py
```diff
@@ -250,3 +250,14 @@ def test_xrt(self):
...
+
+@pytest.mark.remote_data
+def test_map_list_urls_cache():
+    """
+    Test for https://github.com/sunpy/sunpy/issues/4006
+    """
+    urls = ['http://jsoc.stanford.edu/SUM80/D136597189/S00000/image_lev1.fits',
+            'http://jsoc.stanford.edu/SUM79/D136597240/S00000/image_lev1.fits']
+
+    sunpy.map.Map(urls)
```
nabobalis, line 263: Do we have to mock or pass in a fixture to use the cache?

nabobalis:
The doc changes need to be backport manually for 1.0

Issue #4009, opened 2020-04-10
Automate mocking for online tests 🧙‍♂️


samaloney:
### Description

As good internet citizens we should try to avoid making unnecessary requests to web services and servers, at best it needlessly increases traffic and at worst it can overload upstream services and servers. This is especially true when the requests are for testing and not actual user requests.

Much of sunpy's CI testing infrastructure relies upon such requests. Due to this fact and the unreliability of networks and services this can result in many spurious failures of sunpy CI testing. Efforts are already underway to manually mock these request and responses (see #2874) this is pretty arduous and will have to be manually updated in the future.

This issue looks at the possibility of automating this process by serialising the requests and response to files which can then be replayed at test time significantly reducing the need for actual web requests an hopefully increasing CI stability.

Below is a brief overview of the python libraries I could find possibly providing this type of functionality.

| Name | pytest integration | maintained | used by/stars | notes |
|-------|------------------|------------|-------------|-------|
| [VCRpy](https://github.com/kevin1024/vcrpy) | Yes [1](https://github.com/ktosiek/pytest-vcr), [2](https://github.com/kiwicom/pytest-recording) | unclear | 1.8k / 1.7k | supports aiohttp, urllib, boto, boto2 ... |
| [HTTPretty](https://httpretty.readthedocs.io/en/latest/index.html) | Yes [1](https://pypi.org/project/pytest-httpretty/) | active | 3k / 1.8k | Not clear how to replay from saved file |
| [cassette](https://github.com/uber/cassette) | No | dead | | | 
| [responses](https://github.com/getsentry/responses) | Yes builtin | active | 5.1k/1.7k | only supports requests, manual plumbing required |

VCRpy looks like its worth some more investigation as it supports both aiohttp ,urllib, etc and integrates with pytest.

wtbarnes:
Is this something we are still interested in pursuing given the age of this issue?

PR #4010, opened 2020-04-10
VCRpy investigation examples for goes, eve and jsoc

samaloney:
### Description

Initial investigation of using VCRpy to automatically mock some online tests. It **works** but there are some issues:
1. The naming convention for the 'cassettes' there could be collisions with similarly named tests and it doesn't handle parametrised test very well.
1. Obviously it can't work with hypothesis tests.
1. Not clear how to integrate with --remote-data removed the decorators and comment out `remote_data_strict = True`
1. The yaml files are big and some are nested very deeply - possibly keep in a subrepo?
1. ...

For the curious turn off you network and watch the magic:
```
pytest --vcr-record=none sunpy/net/dataretriever/tests/test_eve.py
pytest --vcr-record=none sunpy/net/dataretriever/tests/ test_goes_ud.py
pytest --vcr-record=none sunpy/net/jsoc/tests/test_jsoc.py
```

Not in this PR as the files were way too big (you end up with a yaml serialisation of binary files) but it does work on fetch for eve and goes but hangs on replay for jsoc fetches some threading issue. But it would make no sense to actually do this for data requests, you could do it for the headers and maybe spoof the content.

Addresses issue #4009


nabobalis:
I think VCRpy offers some interesting choices but as it stands, I think we might need to hold off on it for now.

PR #4012, opened 2020-04-13
Add the summary report to tox pytest runs

ayshih:
I like having it: https://docs.pytest.org/en/stable/usage.html#detailed-summary-report

I think it's particularly useful for making sure that relevant test failures aren't overlooked by being swamped by unrelated test failures.  It's a little difficult to illustrate this use since our CI tests are largely coming in green currently, but I'm sure that won't always be the case. =)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -33,7 +33,7 @@ description =
 setenv =
     MPLBACKEND = agg
     COLUMNS = 180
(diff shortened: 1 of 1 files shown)


ayshih:
Hmm, it shows up on some builds but not all of them...

nabobalis:
I assume it only shows up if there is something to report?

ayshih:
> I assume it only shows up if there is something to report?

Ah, I think you're right.  It invariably shows up on our offline builds because it lists the skips, but it's not showing up on successful builds without any skips.

PR #4017, opened 2020-04-14
Tidy/add some Map factory tests

dstansby:
In order of commits:

- Remove the outer `class TestMap`, which wasn't doing anything
- Use more `pathlib` throughout
- parameterize the source tests
- Add a test that pairs of valid input to `sunpy.map.Map` parse (xref #3977; I thought I would check this worked on current master)

### DIFF
--- a/sunpy/map/tests/test_map_factory.py
+++ b/sunpy/map/tests/test_map_factory.py
@@ -204,62 +204,6 @@ class TestMap:
         backin = sunpy.map.Map(afilename)
         assert isinstance(backin, sunpy.map.sources.EITMap)
 
-# ==============================================================================
-# Sources Tests
-# ==============================================================================
-    def test_sdo(self):
-        # Test an AIAMap
-        aia = sunpy.map.Map(AIA_171_IMAGE)
-        assert isinstance(aia, sunpy.map.sources.AIAMap)
-        # TODO: Test a HMIMap
-
-    def test_soho(self):
-        # Test EITMap, LASCOMap & MDIMap
-        eit = sunpy.map.Map(os.path.join(filepath, "EIT", "efz20040301.000010_s.fits"))
-        assert isinstance(eit, sunpy.map.sources.EITMap)
-
-        lasco = sunpy.map.Map(os.path.join(filepath, "lasco_c2_25299383_s.fts"))
-        assert isinstance(lasco, sunpy.map.sources.LASCOMap)
-
-        mdi_c = sunpy.map.Map(os.path.join(filepath, "mdi_fd_Ic_6h_01d.5871.0000_s.fits"))
-        assert isinstance(mdi_c, sunpy.map.sources.MDIMap)
-
-        mdi_m = sunpy.map.Map(os.path.join(filepath, "mdi_fd_M_96m_01d.5874.0005_s.fits"))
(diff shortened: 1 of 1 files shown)


sunpy/map/tests/test_map_factory.py
```diff
@@ -18,235 +18,226 @@
...
-        assert isinstance(eitmap, sunpy.map.GenericMap)
-
-        # Directory
-        directory = pathlib.Path(filepath, "EIT")
+AIA_171_IMAGE = filepath / 'aia_171_level1.fits'
+RHESSI_IMAGE = filepath / 'hsi_image_20101016_191218.fits'
+
+amap = sunpy.map.Map(AIA_171_IMAGE)
+
+valid_map_inputs = [(a_fname, ),
```
Cadair, line 30: If `amap` is a fixture this would need to be a fixture, but then you can't use it in `parametrize` but that's probably not the end of the world looking at how you are using it.

sunpy/map/tests/test_map_factory.py
```diff
@@ -18,235 +18,226 @@
 import sunpy.data.test
 
 
-filepath = sunpy.data.test.rootdir
+filepath = pathlib.Path(sunpy.data.test.rootdir)
 a_list_of_many = [os.fspath(f) for f in pathlib.Path(filepath, "EIT").glob("*")]
```
Cadair, line 22: these also should probably be fixtures :see_no_evil:

sunpy/map/tests/test_map_factory.py
```diff
@@ -18,235 +18,226 @@
...
+    assert isinstance(sequence, sunpy.map.MapSequence)
+
+
+def test_composite():
+    # Test making a CompositeMap
+    comp = sunpy.map.Map(AIA_171_IMAGE, RHESSI_IMAGE, composite=True)
+    assert isinstance(comp, sunpy.map.CompositeMap)
+
+
+def test_patterns():
```
Cadair, line 64: it might be nice if these were individual tests, but that's *really* nitpicky :grinning:

dstansby: Yeah, I thought about splitting them up and making them nicer (maybe parametrizing them), but held off since it seemed to be non-trivial

sunpy/map/tests/test_map_factory.py
dstansby: Am 👍 for making `amap` a fixture, but I had a trawl through the pytest docs and I can't work out how to make it play nicely with `parametrize`.

dstansby: I have tried a few things and gone down a few rabbit holes, and am struggling to work out a way to do this nicely, other than definitely a bunch of functions to return the map types, then calling the functions within the test (sort of like a manual fixture I guess).

nabobalis: Personally I wouldn't waste too much time on it. We can always come back to this. I would rather we get this merged either in its current state or with more test cases than with more fixtures.

dstansby: In that case I think this is ready for review/merging

dstansby:
Removing the test class turned into a rebase nightmare, so I've just left this at the new double-input tests, and parametrizing the source tests.

PR #4020, opened 2020-04-15
Removes appending login from norh and scraper.

Raahul-Singh:
With parfive 1.1 in place, we can safely remove the login appending for ftp urls from scraper.


Cadair:
This has introduced a real online test fail. We need to fix it.

Issue #2284, opened 2017-09-07
Fetching HMI data fails whenn no attribute vso.Physobs is provided


Hackempluf:
When trying to download HMI data without specifying a vso.Physobs attribute, the query fails with the following error: 
>WebFault: Server raised fault: 'Element '' can't be allowed in valid XML message. Died. at /opt/vso/lib/perl5/SOAP/Lite.pm line 1480.
'

Searching for the data is succesful, and when at least one vso.Physobs attribute is specified, the download is successful.

Example code:
```python
import sunpy
from sunpy.net import Fido, attrs as a

#this queryresult cannot be downloaded
qr = Fido.search(a.Time('2017-09-02 23:52:00', '2017-09-02 23:52:30') 
                 & a.Instrument('HMI'))
print(qr)
Fido.fetch(qr)

#this queryresult can be downloaded
qr = Fido.search(a.Time('2017-09-02 23:52:00', '2017-09-02 23:52:30') 
                 & a.Instrument('HMI')
                 & (a.vso.Physobs('intensity') | a.vso.Physobs('LOS_magnetic_field')))
print(qr)
```

Traceback:
[traceback.txt](https://github.com/sunpy/sunpy/files/1284485/traceback.txt)

Cadair:
Thanks for the bug report @Hackempluf. This looks like a VSO side issue to me, I will try and get some VSO peoples eyes on it.

jhourcle:
This is believed to be the message that SunPy was sending to the VSO:

    <?xml version="1.0" encoding="UTF-8"?>
    <SOAP-ENV:Envelope xmlns:SOAP-ENV="http://schemas.xmlsoap.org/soap/envelope/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:SOAP-ENC="http://schemas.xmlsoap.org/soap/encoding/" xmlns:ns0="http://schemas.xmlsoap.org/soap/encoding/" xmlns:ns1="http://virtualsolar.org/VSO/VSOi" xmlns:ns2="http://schemas.xmlsoap.org/soap/envelope/" SOAP-ENV:encodingStyle="http://schemas.xmlsoap.org/soap/encoding/">
       <SOAP-ENV:Header/>
       <ns2:Body>
          <ns1:GetData>
             <body>
                <request>
                   <method>
                      <methodtype>URL-FILE_Rice</methodtype>
                      <methodtype>URL-FILE</methodtype>
                   </method>
[...]

PR #4026, opened 2020-04-16
VSO fetch fix

### LINKED ISSUE #3734: Downloading from VSO is broken when using specific providers
### Description
VSO returns entries with Start Time: None when querying data for SOHO MDI

### Expected behavior
No corrupted entries or a way to filter them out and not download

### Actual behavior
Fido fails when trying to download the result UnifiedResponse object

### Steps to Reproduce
<https://gist.github.com/vit1-irk/a3891d0948c3aff347fd5a82c9ece79f>

Also, see this: #2141 #3372 

### System Details

```
#######
General
#######
Time : Monday, 27. January 2020 04:54PM UT
System : Linux
Processor : 
Arch : 64bit
SunPy : 1.1.0
OS: Manjaro Linux 18.1.5 Juhraya (Linux 5.4.14-2-MANJARO )


##################
Required Libraries
##################
Python: 3.8.1
NumPy: 1.18.1
SciPy: 1.4.1
matplotlib: 3.1.2
Astropy: 4.0
Pandas: 0.25.3
parfive: 1.0.0


#####################
Recommended Libraries
#####################
beautifulsoup: 4.8.2
PyQt4: NOT INSTALLED
PyQt5: 5.14.1
Zeep: 3.4.0
Sqlalchemy: NOT INSTALLED
drms: 0.5.7
```

abhijeetmanhas:
### Description

Fixes #3734 

In the issue Viktor created a **[notebook](https://gist.github.com/vit1-irk/2e518c02c792b5d1109c656431472254)** which shows the fetch fails for *SHA* provider. I inspected and found that whenever we used MDI and the `DataRequestitem` had multiple fileids , the fetch always failed. So I fixed it by having one fileid per datarequestitem. After this fix, I ran the same commands as in the notebook, and all 118 files were downloaded without error, here https://gist.github.com/abhijeetmanhas/089f4a4935cf3fd5e4a431c91f781dc0 .

In fact, this fix is not new , https://github.com/sunpy/sunpy/blob/master/sunpy/net/vso/vso.py#L748-#L767 here we pop the JSOC provider, and have one fileid per request, which solved this issue #2284 .

I have added a minimum test with only 3 files; which will fail on upstream.

However the issue is marked as remote server issue, but it can be fixed at our end. Open for discussions and suggestions.

### DIFF
--- a/sunpy/net/tests/test_vso.py
+++ b/sunpy/net/tests/test_vso.py
@@ -447,6 +447,16 @@ def test_vso_hmi(client, tmpdir):
         assert all([s == series[0] for s in series])
 
 
+@pytest.mark.remote_data
+def test_vso_mdi(client):
+    timerange = a.Time('2005/01/01 00:12', '2005/01/01 00:21')
+    result = client.search(timerange, a.Instrument('mdi'),
+                           a.Wavelength(6768*u.angstrom) & a.Physobs('LOS_magnetic_field'))
+    assert len(result) == 3
+    files = client.fetch(result)
+    assert len(files) == 3
+
+
 @mock.patch('sunpy.net.vso.vso.check_connection', return_value=None)
 def test_get_online_vso_url(mock_urlopen):
     """
--- a/sunpy/net/tests/test_vso.py
+++ b/sunpy/net/tests/test_vso.py
@@ -438,8 +438,7 @@ def test_vso_hmi(client, tmpdir):
     # Extract the DRIs from the request
     dris = dr.request.datacontainer.datarequestitem
 
-    # 3 HMI series and one AIA
-    assert len(dris) == 4
+    assert len(dris) == 90
 
     # For each DataRequestItem assert that there is only one series in it.
     for dri in dris:
--- a/sunpy/net/tests/test_vso.py
+++ b/sunpy/net/tests/test_vso.py
@@ -438,6 +438,7 @@ def test_vso_hmi(client, tmpdir):
     # Extract the DRIs from the request
     dris = dr.request.datacontainer.datarequestitem
 
+    # 9 HMI files and 81 AIA
     assert len(dris) == 90
 
     # For each DataRequestItem assert that there is only one series in it.
@@ -502,7 +503,7 @@ def test_incorrect_content_disposition(client):
     files = client.fetch(results[0:1])
 
     assert len(files) == 1
-    assert  files[0].endswith("mdi_vw_v_9466622_9466622.tar")
+    assert files[0].endswith("mdi_vw_v_9466622_9466622.tar")
     assert "Content" not in files[0]
 
 
--- a/sunpy/net/vso/vso.py
+++ b/sunpy/net/vso/vso.py
@@ -742,29 +742,21 @@ class VSOClient(BaseClient):
         if 'email' not in info:
             info['email'] = 'sunpy'
 
-        # For the JSOC provider we need to make a DataRequestItem for each
-        # series, not just one for the whole provider.
+        """ Create datarequest from maps mapping data provider to
+        fileids and methods, """
+        if info is None:
+            info = {}
 
-        # Remove JSOC provider items from the map
-        jsoc = maps.pop('JSOC', [])
-        # Make DRIs for everything that's not JSOC one per provider
+        if 'email' not in info:
+            info['email'] = 'sunpy'
+
+        reqitems = []
+        for k in maps.keys():
+            for ids in maps[k]:
+                reqitems.append([k, ids])
+
         dris = [self.make('DataRequestItem', provider=k, fileiditem={'fileid': v})
-                for k, v in maps.items()]
-
-        def series_func(x):
-            """ Extract the series from the fileid. """
-            return x.split(':')[0]
-
-        # Sort the JSOC fileids by series
-        # This is a precursor to groupby as recommended by the groupby docs
-        series_sorted = sorted(jsoc, key=series_func)
-        # Iterate over the series and make a DRI for each.
-        # groupby creates an iterator based on a key function, in this case
-        # based on the series (the part before the first ':')
-        for series, fileids in itertools.groupby(series_sorted, key=series_func):
-            dris.append(self.make('DataRequestItem',
-                                  provider='JSOC',
-                                  fileiditem={'fileid': list(fileids)}))
+                for k, v in reqitems]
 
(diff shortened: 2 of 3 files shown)


abhijeetmanhas:
I guess the test fails in windows and html-docs will vanish by a simple re-run. The failed tests have nothing to do with these changes.

sunpy/net/tests/test_vso.py
```diff
@@ -438,8 +438,7 @@ def test_vso_hmi(client, tmpdir):
     # Extract the DRIs from the request
     dris = dr.request.datacontainer.datarequestitem
 
-    # 3 HMI series and one AIA
-    assert len(dris) == 4
+    assert len(dris) == 90
```
Raahul-Singh, line 441: I'm just trying to understand this.
What causes this increase in the number of results?
Further, you should update the comment which still mentions 4 results.

abhijeetmanhas: Thanks, I need to remove that line.
The number of results are still same, but what happened is earlier there were multiple files in same requestitem, so number of results are less.

Now all fileids have different `DataRequestitem`, so no. of results increased. But the files shall always be same.

Raahul-Singh: Cool!

sunpy/net/tests/test_vso.py
```diff
@@ -438,8 +438,8 @@ def test_vso_hmi(client, tmpdir):
     # Extract the DRIs from the request
     dris = dr.request.datacontainer.datarequestitem
 
-    # 3 HMI series and one AIA
-    assert len(dris) == 4
+    # 9 HMI files and 81 AIA
```
Cadair, line 441: Why have these numbers changed?

sunpy/net/tests/test_vso.py
```diff
@@ -448,6 +448,16 @@ def test_vso_hmi(client, tmpdir):
         assert all([s == series[0] for s in series])
 
 
+@pytest.mark.remote_data
+def test_vso_mdi(client):
+    timerange = a.Time('2005/01/01 00:12', '2005/01/01 00:21')
+    result = client.search(timerange, a.Instrument('mdi'),
+                           a.Wavelength(6768*u.angstrom) & a.Physobs('LOS_magnetic_field'))
+    assert len(result) == 3
```
Cadair, line 456: What is this test actually checking? The number of DRIs in the fetch request? Can we actually the thing we want to make sure is happening rather than relying on the return values from VSO?

sunpy/net/vso/vso.py
```diff
@@ -742,29 +742,13 @@ def create_getdatarequest(self, maps, methods, info=None):
...
-        series_sorted = sorted(jsoc, key=series_func)
-        # Iterate over the series and make a DRI for each.
-        # groupby creates an iterator based on a key function, in this case
-        # based on the series (the part before the first ':')
-        for series, fileids in itertools.groupby(series_sorted, key=series_func):
-            dris.append(self.make('DataRequestItem',
-                                  provider='JSOC',
-                                  fileiditem={'fileid': list(fileids)}))
+        reqitems = []
+        for k in maps.keys():
```
Cadair, line 746: Might be a good opportunity to use some more descriptive names here.
```suggestion
        for provider in maps.keys():
```

sunpy/net/vso/vso.py
```diff
@@ -742,29 +742,13 @@ def create_getdatarequest(self, maps, methods, info=None):
...
-        # This is a precursor to groupby as recommended by the groupby docs
-        series_sorted = sorted(jsoc, key=series_func)
-        # Iterate over the series and make a DRI for each.
-        # groupby creates an iterator based on a key function, in this case
-        # based on the series (the part before the first ':')
-        for series, fileids in itertools.groupby(series_sorted, key=series_func):
-            dris.append(self.make('DataRequestItem',
-                                  provider='JSOC',
-                                  fileiditem={'fileid': list(fileids)}))
+        reqitems = []
```
Cadair, line 745: Can you add a comment here describing why this code exists? If I am interpreting this correctly we are creating one `DataRequestItem` per fileid?

abhijeetmanhas: Yes sure. We are creating one `DataRequestItem` for on fileid.

sunpy/net/tests/test_vso.py
abhijeetmanhas: Earlier in `dris` there used to be 3 HMI datarequestitems, and one AIA. But it was like they had array of fileids containing more than one fileid.
For example, the AIA one had 81 files for one item, and hmi had 3 each, so the total files accessible where still 90 but dris list had four items only.

sunpy/net/tests/test_vso.py
abhijeetmanhas: I added this check which will fail on upstream but not after the fix. Though it is not checking anything; just that fix works for a query which won't work on master. So I think I should check in the test that records in getdatarequest have one fileid per record in this test.

abhijeetmanhas:
I recommend going through this [gist](https://gist.github.com/abhijeetmanhas/3ccff55b148dc9d7708cd00f8f473ceb) for debug info.

jhourcle:
> I don't understand why we need this fix, which definitely concerns me, I will be asking for some input from the VSO team.

Sorry for the slow response -- and I apologize for not following up on #2284 -- that was right about the time that I was fired from the SDAC, so was a bit distracted on things.  I also don't know if I access to trace things as I used to.

I'd rather try to figure out *why* this is going wrong, and if you're having problems with it, there's a possibility that other clients might have problems, too, if it's something on the VSO side of things.  (SHA is a new-ish data provider ... as it had a complete rewrite in ~2018.  (because they moved the data into NetDRMS (the system that JSOC uses), so the old system stopped working ... and the person who wrote it didn't know that I had forked the JSOC code, but had been waiting years for info to finish getting info to map VSO concepts to MDI datasets.)

As for would this cause a problem with VSO?  It actually might.  The VSO "data providers" speak a simplified version of the VSO API, and only accept a single DataRequestItem ... so the VSO "Core" will make a separate request for each one ... and it does it in parallel if the server running the VSO Core has threading enabled.  Normally, we're not making so many connections to cause a DoS on the service, but if you did this with hundreds or thousands of fileIDs, it might.

nabobalis:
I will close this for now. If any updates need to occur, we can open a new PR or a new issue to track them.

PR #4034, opened 2020-04-18
Improve sample data doc

dstansby:
The list of files isn't quite rendering properly: https://docs.sunpy.org/en/latest/code_ref/data.html#module-sunpy.data.sample

Maybe this will fix it


nabobalis:
it worked: https://46888-2165383-gh.circle-artifacts.com/0/docs/_build/html/code_ref/data.html?highlight=data%20sample#module-sunpy.data.sample

PR #4037, opened 2020-04-20
Use singledispatch in Map factory 

### LINKED ISSUE #2384: Post Python 2: Refactor Map and Timeseries factory to use singledispatch in `__call__`

dstansby:
Solves the map part of #2384, and builds on #3977. Some notes

- This converts url strings into `urllib.request.Request` objects, solely for the purpose of having a unique class to dispatch on. I think this is the right choice?
- I had to vendor a `singledispatch` version that works on methods, because `singledispatchmethod` was only added in python 3.8.
- Should I add a changelog?

Otherwise everything should work as before, but much neater code, and easier to add other dispatching options in the future.

### DIFF
--- a/sunpy/map/map_factory.py
+++ b/sunpy/map/map_factory.py
@@ -3,6 +3,7 @@ from collections import OrderedDict
 import warnings
 import pathlib
 import glob
+from urllib.request import urlopen, Request
 
 import numpy as np
 import astropy.io.fits
@@ -27,7 +28,6 @@ from sunpy.util.datatype_factory_base import BasicRegistrationFactory
 from sunpy.util.datatype_factory_base import NoMatchError
 from sunpy.util.datatype_factory_base import MultipleMatchError
 from sunpy.util.datatype_factory_base import ValidationFunctionError
-from urllib.request import urlopen
 
 SUPPORTED_ARRAY_TYPES = (np.ndarray,)
 try:
@@ -197,6 +197,9 @@ class MapFactory(BasicRegistrationFactory):
                 header = args.pop(i)
                 args.insert(i, (data, header))
                 nargs -= 1
+            elif isinstance(arg, str) and _is_url(arg):
+                # Repalce URL string with a Request object to dispatch on later
+                args[i] = Request(arg)
             i += 1
 
         # For each of the arguments, handle each of the cases
@@ -220,8 +223,8 @@ class MapFactory(BasicRegistrationFactory):
                 already_maps.append(arg)
 
             # URL
-            elif isinstance(arg, str) and _is_url(arg):
-                url = arg
+            elif isinstance(arg, Request):
+                url = arg.full_url
                 path = str(cache.download(url).absolute())
                 pairs = self._read_file(path, **kwargs)
                 data_header_pairs += pairs
--- a/sunpy/map/map_factory.py
+++ b/sunpy/map/map_factory.py
@@ -200,11 +200,12 @@ class MapFactory(BasicRegistrationFactory):
             elif isinstance(arg, str) and _is_url(arg):
                 # Repalce URL string with a Request object to dispatch on later
                 args[i] = Request(arg)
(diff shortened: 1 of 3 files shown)


nabobalis:
I think this should get a changelog.

sunpy/map/map_factory.py
```diff
@@ -179,14 +180,11 @@ def _parse_args(self, *args, **kwargs):
...
         """
-        data_header_pairs = list()
-        already_maps = list()
-
         # Account for nested lists of items
         args = expand_list(args)
 
-        # Take array, header pairs and put them in a tuple
-        # This makes each item in the args list correspond to exactly one data, header pair
+        # Sanitise the input so that each 'type' of input corresponsds to a different
```
nabobalis, line 186:
```suggestion
        # Sanitise the input so that each 'type' of input corresponds to a different
```

sunpy/map/map_factory.py
```diff
@@ -3,6 +3,7 @@
 import warnings
 import pathlib
 import glob
+from urllib.request import urlopen, Request
```
nabobalis, line 6:
```suggestion
from urllib.request import Request
```

dstansby: I think this is still needed in `_is_url`?

nabobalis: My find and search must have been playing up.

nabobalis: Seriously, i can't find it.

dstansby: https://github.com/sunpy/sunpy/pull/4037/files#diff-9306ac2bb55014e8e339ffdedaab50c5L347

sunpy/util/functools.py
```diff
@@ -27,5 +27,19 @@ def wrapper(*args, **kwargs):
...
 
+def _singledispatchmethod(func):
+    """
+    A variant of `functools.singledispatch` that can be used on methods.
+    As of python 3.8, this can be replaced by the stdlib `functools.singledispatchmethod` decorator.
+    """
+    dispatcher = functools.singledispatch(func)
+
+    def wrapper(*args, **kw):
+        return dispatcher.dispatch(args[1].__class__)(*args, **kw)
```
Cadair, line 38: isn't this actually the same as "seconddispatch", because it's just dispatching on the second arg? The `singledispatchmethod` in py 38 is smarter than this right?

sunpy/util/functools.py
```diff
@@ -27,5 +27,19 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def _singledispatchmethod(func):
```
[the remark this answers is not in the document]
dstansby: To make it a private method, so we can break API/remove this without having to deprecate etc.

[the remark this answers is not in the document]
Cadair: Fair. Although, I am not sure I extend our usual deprecation policy to everything in utils, that way madness may lie. Still fine with me :+1:

sunpy/util/functools.py
dstansby: Huh, oh yeah... I'll get rid of all this and just use seconddispatch

Cadair: You could conditionally import `singledispatchmethod` if it exists?

dstansby: Can do if you want, although it would make the code a bit messier

dstansby: Oh, but the reason you made the dispatchonsecondarg was for methods right? In that case then yes, I should conditionally import.

dstansby: Just trying the replacement, and it's a bit messy, can I do it in another PR?

PR #4039, opened 2020-04-20
Fixes to the STEREO/SECCHI COR2 starfield gallery example

ayshih:
The [STEREO/SECCHI COR2 starfield gallery example](https://docs.sunpy.org/en/stable/generated/gallery/units_and_coordinates/STEREO_SECCHI_starfield.html) has a few minor issues:

- It repeats the notion of having to install `astroquery`.
- It uses a list of scalar coordinates instead of a single array coordinate.
- It sets the `obstime` for an ICRS `SkyCoord`, which is allowed, but could be potentially confusing.  The ICRS coordinate frame doesn't change over time.  The `obstime` for the transformation chain will be read from the destination frame (`cor2.coordinate_frame`).
- It sets the "very far away" distance as `1e12 * u.km`, which is actually only ~0.1 light-years, which looks silly as a distance to stars.
- There's a use of `frames.Helioprojective(observer=cor2.observer_coordinate)` instead of simply using `cor2.coordinate_frame`, which will have the `obstime` set too.

### DIFF
--- a/changelog/4039.doc.rst
+++ b/changelog/4039.doc.rst
@@ -1,0 +1,1 @@
+Made improvements to the gallery example :ref:`sphx_glr_generated_gallery_units_and_coordinates_stereo_secchi_starfield.py`.
--- a/examples/units_and_coordinates/STEREO_SECCHI_starfield.py
+++ b/examples/units_and_coordinates/STEREO_SECCHI_starfield.py
@@ -35,22 +35,22 @@ f = hv.download_jp2('2014/05/15 07:54', observatory='STEREO_A',
 cor2 = sunpy.map.Map(f)
 
 ###############################################################################
-# To efficiently search the star field we need to know what stars are near the
+# To efficiently search the star field, we need to know what stars are near the
 # Sun as observed by STEREO. We need the vector that points from STEREO to the Sun.
-# By converting to HCRS we get the vector from the Sun to STEREO
+# The location of STEREO in HCRS provides the Sun-to-STEREO vector.
 
 sun_to_stereo = cor2.observer_coordinate.transform_to('hcrs')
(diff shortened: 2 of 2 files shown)


ayshih:
I think that the `.transform_to(cor2.coordinate_frame)` calls can actually be removed (on appropriately fixed versions of Astropy), but I haven't confirmed that.

nabobalis:
Can you add a changelog please?

nabobalis:
It seems that the doc build doesn't like the reference:

```
whatsnew/latest_changelog.txt:124: WARNING: undefined label: sphx-glr-generated-gallery-units-and-coordinates-stereo-secchi-starfield-py (if the link has no caption the label must precede a section header)
```

nabobalis:
Those gallery references are a nightmare.

ayshih:
> Those gallery references are a nightmare. Something I should add to the dev docs about.

Is there a way to reference a gallery example that works?  That is, a way that is smarter than just linking the URL, of course.

nabobalis:
> > Those gallery references are a nightmare. Something I should add to the dev docs about.
> 
> Is there a way to reference a gallery example that works? That is, a way that is smarter than just linking the URL, of course.

When I figure that out, I plan to add it to the docs.

ayshih:
> When I figure that out, I plan to add it to the docs.

This page (https://sphinx-gallery.github.io/stable/advanced.html#know-your-gallery-files) appears to explain how to do the reference.  Let's see if it works.

ayshih:
Yay, the reference worked! (see https://47188-2165383-gh.circle-artifacts.com/0/docs/_build/html/whatsnew/changelog.html#improved-documentation)

PR #4046, opened 2020-04-22
Added a custom directive for Sphinx for generated HTML output

ayshih:
This PR defines a custom directive for Sphinx (called `generate`) for generated HTML output, which cleans up a clunky solution in #3951.  For example, now the docstring for `MapSequence.quicklook()` simply has at the end:
```
    .. generate:: html
        :html_border:
    
        from sunpy.map import Map
        import sunpy.data.sample
        seq = Map(sunpy.data.sample.HMI_LOS_IMAGE,
                  sunpy.data.sample.AIA_1600_IMAGE,
                  sunpy.data.sample.EIT_195_IMAGE,
                  sequence=True)
        print(seq._repr_html_())
```
The Python code is run, and the printed-out HTML is included in the page generated by Sphinx.  No on-the-fly modification of docstrings, no private methods, and no having to detect Sphinx.

This will also make it super-easy to show off the HTML summary anywhere else in the documentation.


docs/conf.py
```diff
@@ -262,6 +265,54 @@ def rstjinja(app, docname, source):
...
+
+        attributes = {'format': ' '.join(self.arguments[0].lower().split())}
+
+        source, lineno = self.state_machine.get_source_and_line(self.lineno)
+
+        old_stdout, sys.stdout = sys.stdout, StringIO()
+        try:
+            # Exceute the Python code and capture its output
+            exec('\n'.join(self.content))
+            text = sys.stdout.getvalue()
```
Cadair, line 298: Isn't this lovely :rofl:

docs/conf.py
```diff
@@ -262,6 +265,54 @@ def rstjinja(app, docname, source):
...
+
+            # Wrap HTML output in a black border if requested
+            if attributes['format'] == 'html' and 'html_border' in self.options:
+                text = f"<div style='border:1px solid black'>{text}</div>"
```
Cadair, line 302: Maybe adding a slightly larger gap between the border and the text might be nice?
```suggestion
                text = f"<div style='border:1px solid black; padding:3px'>{text}</div>"
```

nabobalis:
I feel like writing this change down somewhere might be useful in case we want to do more HTML and forget this exists.

docs/conf.py
ayshih: Good idea, done

ayshih:
I've added a subsection and example to the Sphinx part of our developer's guide.

ayshih:
For the record, I figured out how to write this custom directive by studying these two places:

- [The `raw` directive in `docutils`](https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/docutils/parsers/rst/directives/misc.py)
- [This Stack Overflow thread](https://stackoverflow.com/questions/7250659/how-to-use-python-to-programmatically-generate-part-of-sphinx-documentation)

PR #4047, opened 2020-04-22
Warn when assuming CUNIT in map sources

dstansby:
I was taking a look at map sources and to see examples of fixing metadata, and it struck me that there should probably be a warning if anything is being assumed that isn't being cross checked against other present header keys.

### DIFF
--- a/changelog/4047.trivial.rst
+++ b/changelog/4047.trivial.rst
@@ -1,0 +1,1 @@
+Several map sources in `sunpy.map.sources` assume 'arcsec' units. If this assumption is raised a warning is now raised, instead of the assumption happening silently.
--- a/docs/code_ref/map.rst
+++ b/docs/code_ref/map.rst
@@ -105,12 +105,11 @@ demonstrated by the following example.
 .. code-block:: python
 
     import sunpy.map
+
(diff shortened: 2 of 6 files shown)


dstansby:
I updated this to cover all places where this is happening.

Cadair:
I am not sure on this, I understand your logic, we are making assumptions about the metadata. However, isn't making assumptions about the metadata the entire point of the Map sources? In the AIA map we choose to read the position info from other keys, in the HMI map we fix something I think, to me this is the point, we know that if `INSTRUME = rhessi` then the units are missing but the file is in those units because that's what the documentation (lololol) for that data says it is?

dstansby:
I would agree if there was some sort of cross check against other (non-standard) FITS keys, or an link to some documentation, but in all these cases the assumption is completely blind.

Cadair:
We can't really check things which are non-standard reliably, and I am game for better documentation. I am not really in favour of throwing these warnings at the users when there is nothing they can do (or would want to do) to remove them. I think if we are not careful with warnings they will confuse new users and make users ignore the important ones we throw at them.

Cadair:
I am not sure where the best place to document this is, maybe just in the docstrings for the sources?

PR #4050, opened 2020-04-22
Remove requirement for JSOC client to have a primekey

### LINKED ISSUE #4049: Searching JSOC without a PrimeKey fails
It should be possible to search JSOC without a PrimeKey, e.g.
```python
from sunpy.net.jsoc import attrs as jsoca

series = jsoca.Series('hmi.mrsynop_small_720s')

result = Fido.search(series)
```
 but this fails with 
```python
  File "../examples/finding_data/plot_hmi.py", line 11, in <module>
    result = Fido.search(series)
  File "/Users/dstansby/github/sunpy/sunpy/net/fido_factory.py", line 270, in search
    results = query_walker.create(query, self)
  File "/Users/dstansby/github/sunpy/sunpy/net/attr.py", line 466, in create
    return self.createmm(self, *args, **kwargs)
  File "/Users/dstansby/github/sunpy/sunpy/util/functools.py", line 19, in wrapper
    return dispatcher.dispatch(args[1].__class__)(*args, **kwargs)
  File "/Users/dstansby/github/sunpy/sunpy/net/fido_factory.py", line 199, in _create_data
    return factory._make_query_to_client(query)
  File "/Users/dstansby/github/sunpy/sunpy/net/fido_factory.py", line 419, in _make_query_to_client
    results.append(tmpclient.search(*query))
  File "/Users/dstansby/github/sunpy/sunpy/net/jsoc/jsoc.py", line 352, in search
    return_results.append(self._lookup_records(iargs))
  File "/Users/dstansby/github/sunpy/sunpy/net/jsoc/jsoc.py", line 867, in _lookup_records
    ds = self._make_recordset(**iargs)
  File "/Users/dstansby/github/sunpy/sunpy/net/jsoc/jsoc.py", line 791, in _make_recordset
ValueError: Atleast one PrimeKey must be passed.
```

Doing the following search works fine, even though the time attribute to the search is completely ignored and all the synoptic maps that exist are returned by the search
```python
from sunpy.net.jsoc import attrs as jsoca

time = a.Time('2010/01/01', '2010/01/01')
series = jsoca.Series('hmi.mrsynop_small_720s')

result = Fido.search(time, series)
```

dstansby:
Fixes #4049, but might have other side effects? See #4049 for a valid search that has no primekey.

### DIFF
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -785,11 +785,6 @@ class JSOCClient(BaseClient):
                 # break because we can skip adding {} at the end of pkstr, if the primekey
                 # dict is empty.
 
-        if not pkstr:
-            # pkstr cannot be totally empty
-            error_message = "Atleast one PrimeKey must be passed."
-            raise ValueError(error_message)
-
         dataset = '{series}{primekey}{segment}'.format(series=series,
                                                        primekey=pkstr,
                                                        segment=segment)
--- a/sunpy/net/jsoc/tests/test_jsoc.py
+++ b/sunpy/net/jsoc/tests/test_jsoc.py
@@ -182,16 +182,18 @@ def test_lookup_records_errors():
 
 
 @pytest.mark.remote_data
-def test_make_recordset_errors():
-    d1 = {'series': 'aia.lev1_euv_12s'}
-    with pytest.raises(ValueError):
-        client._make_recordset(**d1)
+def test_make_recordset_series():
+    # Check that a series by itself works
+    client._make_recordset(series='hmi.mrsynop_small_720s')
 
-    d1.update({
-        'end_time': astropy.time.Time('2014-01-01 01:00:35', scale='tai'),
-        'start_time': astropy.time.Time('2014-01-01 00:00:35', scale='tai'),
-        'primekey': {'T_REC': '2014.01.01_00:00:35_TAI-2014.01.01_01:00:35_TAI'}
-    })
+
+@pytest.mark.remote_data
+def test_make_recordset_errors():
+    d1 = {'series': 'aia.lev1_euv_12s',
+          'end_time': astropy.time.Time('2014-01-01 01:00:35', scale='tai'),
+          'start_time': astropy.time.Time('2014-01-01 00:00:35', scale='tai'),
+          'primekey': {'T_REC': '2014.01.01_00:00:35_TAI-2014.01.01_01:00:35_TAI'}
+          }
 
     with pytest.raises(ValueError):
         client._make_recordset(**d1)


dstansby:
I've put this as no changelog needed as it's building on the Fido overhaul in 2.0

Cadair:
I would like to double check this with people who know a little more about DRMS. I am wondering if we should just be considering series a prime key.

nabobalis:
There is one online test fail but I am not sure if its due to this PR or just transient

```
    @pytest.mark.remote_data
    def test_wait_get():
        responses = client.search(
            a.Time('2012/1/1T1:00:36', '2012/1/1T01:00:38'),
            a.jsoc.Series('hmi.M_45s'), a.jsoc.Notify('jsoc@cadair.com'))
        path = tempfile.mkdtemp()
>       res = client.fetch(responses, path=path)

../../.tox/py38-online/lib/python3.8/site-packages/sunpy/net/jsoc/tests/test_jsoc.py:125: 
```

dstansby:
> I would like to double check this with people who know a little more about DRMS. I am wondering if we should just be considering series a prime key.

👍 , feel free to milestone to `2.1`.

Cadair:
If we got the definition of primekey wrong I would consider this a bugfix anyway :smile:

mbobra:
I am not totally sure I understand this -- maybe I am missing something? Every series must have at least one prime key. A prime key, or combination of prime keys, uniquely defines a record in PostgreSQL. The JSOC database uses PostgreSQL tables to store metadata keywords and values.

The DRMS series linked in #4049, `hmi.mrsynop_small_720s`, does have a prime key. It is `CAR_ROT`, the Carrington rotation number. You can verify the prime keys of any drms series using the [`Client.pkeys()` method in the drms package](https://docs.sunpy.org/projects/drms/en/latest/generated/drms.Client.pkeys.html#drms.Client.pkeys) like this:
```python
import drms                                                                                                                                        
c = drms.Client()
c.pkeys('hmi.mrsynop_small_720s')                                                                                                                  
```

dstansby:
> I am not totally sure I understand this -- maybe I am missing something? Every series must have at least one prime key. A prime key, or combination of prime keys, uniquely defines a record in PostgreSQL. The JSOC database uses PostgreSQL tables to store metadata keywords and values.
> 
> The DRMS series linked in #4049, `hmi.mrsynop_small_720s`, does have a prime key. It is `CAR_ROT`, the Carrington rotation number. You can verify the prime keys of any drms series using the [`Client.pkeys()` method in the drms package](https://docs.sunpy.org/projects/drms/en/latest/generated/drms.Client.pkeys.html#drms.Client.pkeys) like this:
> 
> ```python
> import drms                                                                                                                                        
> c = drms.Client()
> c.pkeys('hmi.mrsynop_small_720s')                                                                                                                  
> ```

But it seems like even though every record might have at least one primekey associated with it, it's possible to perform a search with just the series (no primekeys) to return all the records in that series?

mbobra:
> But it seems like even though every record might have at least one primekey associated with it, it's possible to perform a search with just the series (no primekeys) to return all the records in that series?

Ah, yes, technically possible and fine for small series like in your example. But we don't want to encourage this, because if someone does it on, say, `hmi.M_720s`, then JSOC will melt 😭

dstansby:
Ah that makes sense! I will change this PR into a comment in the code saying that.

PR #4053, opened 2020-04-23
add HMISynoptic Map

sashank27:
### Description

Add support for Synoptic Maps from HMI

### DIFF
--- a/sunpy/map/sources/sdo.py
+++ b/sunpy/map/sources/sdo.py
@@ -1,11 +1,11 @@
 """SDO Map subclass definitions"""
 
+import numpy as np
+
 import astropy.units as u
 from astropy.coordinates import CartesianRepresentation, HeliocentricMeanEcliptic
 from astropy.visualization import AsinhStretch
 from astropy.visualization.mpl_normalize import ImageNormalize
-
-import numpy as np
 
 from sunpy.map import GenericMap
 from sunpy.map.sources.source_type import source_stretch
@@ -138,7 +138,7 @@ class HMISynopticMap(HMIMap):
             # the spacing should be modified to 180/pi times the original value
             # Reference: Section 5.5, Thompson 2006
             self.meta['cdelt2'] = 180 / np.pi * self.meta['cdelt2']
-            self.meta['cdelt1'] *= -1
+            self.meta['cdelt1'] = np.abs(self.meta['cdelt1'])
 
     @classmethod
     def is_datasource_for(cls, data, header, **kwargs):
--- a/sunpy/map/sources/sdo.py
+++ b/sunpy/map/sources/sdo.py
@@ -5,10 +5,12 @@ from astropy.coordinates import CartesianRepresentation, HeliocentricMeanEclipti
 from astropy.visualization import AsinhStretch
 from astropy.visualization.mpl_normalize import ImageNormalize
 
+import numpy as np
+
 from sunpy.map import GenericMap
 from sunpy.map.sources.source_type import source_stretch
 
-__all__ = ['AIAMap', 'HMIMap']
+__all__ = ['AIAMap', 'HMIMap', 'HMISynopticMap']
 
 
 class AIAMap(GenericMap):
@@ -42,7 +44,7 @@ class AIAMap(GenericMap):
     """
 
     def __init__(self, data, header, **kwargs):
-        GenericMap.__init__(self, data, header, **kwargs)
+        super().__init__(data, header, **kwargs)
 
         # Fill in some missing info
         self.meta['detector'] = self.meta.get('detector', "AIA")
@@ -95,7 +97,7 @@ class HMIMap(GenericMap):
 
     def __init__(self, data, header, **kwargs):
 
-        GenericMap.__init__(self, data, header, **kwargs)
+        super().__init__(data, header, **kwargs)
 
         self.meta['detector'] = self.meta.get('detector', "HMI")
         self._nickname = self.detector
@@ -117,4 +119,32 @@ class HMIMap(GenericMap):
     @classmethod
     def is_datasource_for(cls, data, header, **kwargs):
         """Determines if header corresponds to an HMI image"""
-        return str(header.get('instrume', '')).startswith('HMI')
+        return (str(header.get('TELESCOP', '')).endswith('HMI') and
+                not HMISynopticMap.is_datasource_for(data, header))
+
+
+class HMISynopticMap(HMIMap):
+
+    def __init__(self, data, header, **kwargs):
+        super().__init__(data, header, **kwargs)
+
+        if self.meta['cunit1'] == 'Degree':
+            self.meta['cunit1'] = 'degree'
+
+        if self.meta['cunit2'] == 'Sine Latitude':
+            self.meta['cunit2'] = 'degree'
+
+            # Since, this map uses the cylindrical equal-area (CEA) projection,
+            # the spacing should be modified to 180/pi times the original value
+            # Reference: Section 5.5, Thompson 2006
+            self.meta['cdelt2'] = 180 / np.pi * self.meta['cdelt2']
+            self.meta['cdelt1'] *= -1
+
+    @classmethod
+    def is_datasource_for(cls, data, header, **kwargs):
+        """
+        Determines if header corresponds to an HMI synoptic map.
+        """
+        return (str(header.get('TELESCOP', '')).endswith('HMI') and
+                str(header.get('CONTENT', '')) ==
+                'Carrington Synoptic Chart Of Br Field')
--- a/sunpy/map/sources/tests/test_hmi_synoptic_source.py
+++ b/sunpy/map/sources/tests/test_hmi_synoptic_source.py
@@ -5,9 +5,8 @@ import pytest
 
 from astropy.io import fits
 
-from sunpy.map.sources.sdo import HMISynopticMap, HMIMap
 from sunpy.map import Map
-from sunpy.util import MetaDict
+from sunpy.map.sources.sdo import HMISynopticMap
 
 
 @pytest.fixture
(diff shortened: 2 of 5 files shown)


sunpy/map/sources/sdo.py
```diff
@@ -121,4 +123,28 @@ def observatory(self):
     @classmethod
     def is_datasource_for(cls, data, header, **kwargs):
         """Determines if header corresponds to an HMI image"""
-        return str(header.get('instrume', '')).startswith('HMI')
+        return (str(header.get('TELESCOP', '')).endswith('HMI') and
+                str(header.get('CONTENT', '')) !=
+                'Carrington Synoptic Chart Of Br Field')
```
Cadair, line 128:
```suggestion
                not HMISynopticMap.is_datasource_for(data, header))
```

sunpy/map/sources/sdo.py
```diff
@@ -121,4 +123,28 @@ def observatory(self):
...
+
+class HMISynopticMap(HMIMap):
+    
+    def __init__(self, data, header, **kwargs):
+        GenericMap.__init__(self, data, header, **kwargs)
```
Cadair, line 133: This should be:
```suggestion
        super().__init__(data, header, **kwargs)
```
(also this needs changing in HMIMap)

sashank27:
@dstansby Should I fetch the FITS using Fido and then do the test?
Also, what about the existing example in the docs regarding handling of HMI synoptic maps?

sunpy/map/sources/tests/test_hmi_synoptic_source.py
```diff
@@ -0,0 +1,76 @@
...
+from sunpy.map.sources.sdo import HMISynopticMap, HMIMap
+from sunpy.map import Map
+from sunpy.util import MetaDict
+
+
+@pytest.fixture
+def hmi_synoptic():
+    # sample HMI synoptic map
+    # Please do not edit this header, it is a 1:1 copy of a HMI map header
+    header = MetaDict([
```
Cadair, line 20: It would be better to extract the full fits header as ascii I think, you can do that with the `fitsheader` command provided by astropy. See https://github.com/sunpy/sunpy/blob/master/sunpy/visualization/animator/tests/test_wcs.py#L15 for an example

sunpy/map/sources/sdo.py
```diff
@@ -121,4 +123,31 @@ def observatory(self):
...
+        super().__init__(data, header, **kwargs)
+
+        if self.meta['cunit1'] == 'Degree':
+            self.meta['cunit1'] = 'degree'
+
+        if self.meta['cunit2'] == 'Sine Latitude':
+            self.meta['cunit2'] = 'degree'
+
+            # Instead of the spacing in sin(lat), this should be 180/pi times
+            # that value (see Thompson 2005)
```
ayshih, line 142: I recommend either (or both) specifically pointing to Section 5.5 in that paper or stating explicitly that this map uses the cylindrical equal-area (CEA) projection.

Cadair:
This applies to both #4054 and this PR.

There is some interesting subtlety here between the names of these maps `Synoptic`, what the actual purpose of them is (to fix some metadata in files from a specific source) and what we are using to detect them.

I am mostly ok with this one (although I think that the `is_datasource_for` could be more specific) but the `CTYPE` check alone in #4054 is not sufficient. I think it's to keep in mind while working on these synoptic maps is to remember we only want them to be used for the files and not really a custom header which happens to look like a synoptic map but might not be one?

dstansby:
> and not really a custom header which happens to look like a synoptic map but might not be one?

What does this mean? That there can be non-synoptic maps that have the same projection? Come to think of it, what's the correct definition of 'synoptic map'?

Cadair:
> That there can be non-synoptic maps that have the same projection?

yes.

> Come to think of it, what's the correct definition of 'synoptic map'?

good question (presumably one that's made up because the sun rotates?). This is why I am bringing up the name "SynopticMap" here because it's less about that and more about the data provided by JSOC which has these metadata issues.

dstansby:
I think you'll need to rebase this on to the current master branch to fix the failing tests.

sunpy/map/sources/sdo.py
```diff
@@ -115,4 +117,32 @@ def observatory(self):
...
+            self.meta['cunit1'] = 'degree'
+
+        if self.meta['cunit2'] == 'Sine Latitude':
+            self.meta['cunit2'] = 'degree'
+
+            # Since, this map uses the cylindrical equal-area (CEA) projection,
+            # the spacing should be modified to 180/pi times the original value
+            # Reference: Section 5.5, Thompson 2006
+            self.meta['cdelt2'] = 180 / np.pi * self.meta['cdelt2']
+            self.meta['cdelt1'] *= -1
```
dstansby, line 139: Does anyone know why this change needs to be made? A negative CDELT seems valid to me...

ayshih: Ugh.  I think it's quite likely that the CDELT1 is intentionally negative, so it'd be incorrect to invert it like this.  The Carrington longitude of disk center decreases over time (i.e., from 360 degrees to 0 degrees), so it'd be reasonable to construct a synoptic map with a decreasing horizontal axis.

The gallery example boldly claims that `CDELT` has the "wrong sign" without any justification.

ayshih: On the other hand, it appears that the horizontal axis for the image provided by HMI (http://jsoc.stanford.edu/data/hmi/synoptic/hmi.Synoptic_Mr.2191.png):
hmi Synoptic_Mr 2191
matches our gallery example (which negates CDELT1):
sphx_glr_hmi_synoptic_maps_001

So maybe HMI *does* have the incorrect sign for CDELT1 in their header?  Do we have written confirmation of this?

ayshih: <sigh> I don't know why their CDELT1 needs to be inverted, but it definitely needs to be inverted to get the right answers (see the magnetograms below).  Let's document this assumption explicitly in our docstring.

Here's the magnetogram when the Carrington longitude of disk center is ~90 degrees:
shmi_maglc_fd_20170615_224645

Here's the magnetogram when the Carrington longitude of disk center is ~300 degrees:
shmi_maglc_fd_20170531_143444

ayshih: That is, their provided images show the correct axis values, which is not consistent with their CDELT1 sign.

Cadair: > That is, their provided synoptic images show the correct axis values, which is not consistent with their CDELT1 sign.

iirc this is why it's `*-1`. Because otherwise it didn't match their published synoptic images.

dstansby: Thanks for checking this out @ayshih. I am quite reluctant to do `self.meta['cdelt1'] *= -1` without cross checking this against something else in the header that indicates we should do this. It could be possible that JSOC fix the CDELT value in the future and then we end up erroneously multiplying by -1. To be on the safe side I would say we shouldn't make this fix for now (it's easy enough to add in the future), but wonder what others think?

ayshih: Er, not fixing the CDELT and providing maps with coordinates that don't match publicly released plots sounds like a bad idea.

We can protect against them fixing the sign in their headers if it were instead `self.meta['cdelt1'] = np.abs(self.meta['cdelt'])`.

dstansby: 👍 that sounds like a good solution

mbobra: If you are interested, [here is the code](http://jsoc.stanford.edu/data/hmi/synoptic/displaysynoptic.pro) that generates the HMI synoptic maps (apologies for the code quality, I wrote it back in my IDL days).

mbobra:
> Come to think of it, what's the correct definition of 'synoptic map'?

My understanding is that a synoptic map simply shows the Sun throughout one Carrington rotation. They're usually constructed by taking a column of data from the central meridian of every image observed within a Carrington rotation and stacking them together horizontally. So they're diachronic by nature.

The HMI magnetic field team also came up with another concept, called a synoptic frame. This is similar in concept, but replaces the first 120° latitude swath of data in the synoptic map with synchronic data from a single magnetogram. This method does not blur out quickly-evolving photospheric features.

sunpy/map/sources/sdo.py
dstansby, line 139:
```suggestion
            self.meta['cdelt1'] = np.abs(self.meta['cdelt1'])
```

sunpy/map/sources/sdo.py
```diff
@@ -117,4 +119,32 @@ def observatory(self):
...
+            self.meta['cunit1'] = 'degree'
+
+        if self.meta['cunit2'] == 'Sine Latitude':
+            self.meta['cunit2'] = 'degree'
+
+            # Since, this map uses the cylindrical equal-area (CEA) projection,
+            # the spacing should be modified to 180/pi times the original value
+            # Reference: Section 5.5, Thompson 2006
+            self.meta['cdelt2'] = 180 / np.pi * self.meta['cdelt2']
+            self.meta['cdelt1'] *= np.abs(self.meta['cdelt1'])
```
dstansby, line 141: An extra * has snuck in here
```suggestion
            self.meta['cdelt1'] = np.abs(self.meta['cdelt1'])
```

dstansby:
I think this is good to merge given two approvals. Should get the CI to pass though. @sashank27 could you rebase this on to the current master branch?

sashank27:
@dstansby I have rebased the branch, but I see there are failed tests, related to `TimeSeries` I believe

dstansby:
Since test failures look unrelated and are fixed on master, and this has two approvals, I'm going to merge this, thanks a lot @sashank27!

Issue #4048, opened 2020-04-22
How to search for VSO EXTENT_TYPE?


dstansby:
I'm trying to use `Fido` to search the vso by the [EVENT_TYPE keyword](https://sdac.virtualsolar.org/cgi/show_details?keyword=EXTENT_TYPE), but I'm not entirely sure what attr I should be using for this. Sepcifically I want to search for EXTENT_TYPE = 'synotpic'.

The most obvious one seems to be [`sunpy.net.vso.attrs.Extent`](https://docs.sunpy.org/en/latest/code_ref/net.html#module-sunpy.net.vso.attrs), but this requires x, y, width, length values which I don't want to specify. Does anyone have any idea?

abhijeetmanhas:
```python
class Extent(_attr.DataAttr):
    """
    Specify the spatial field-of-view of the query. Due to a bug in the VSO,
    the Extent attribute is not used.

    """
```
In the docstring for vso attrs file, it is written their is bug in VSO. May I know what's the bug and whether it is fixed now?

dstansby:
Looks like @wafels added that comment a few years back, perhaps he remembers?

ejm4567:
I think the issue with Extent in the VSO is that not all Data Providers support searches against it, and if they do, only the Extent[type] attribute. The VSO never implemented searches against the FOV dimensions, or the co-ordinates of the reference pixel. We are working on implementing a full search of all the Extent variables for Data Providers that have that data available.

Cadair:
Looks like we should maybe make a new attr `ExtentType` to support these queries?

dstansby:
For refernce sunpy has an [`ExtentType` attr](https://docs.sunpy.org/en/stable/generated/api/sunpy.net.attrs.ExtentType.html#sunpy.net.attrs.ExtentType) that is used to search in the `GONGClient`, the work to fix this is presumably make the `VSOClient` understand that attr and search with it.

wtbarnes:
Well I've once again run into this issue and was very confused by the fact that we have an `ExtentType` attr, the VSO returns an `ExtentType` in the response, yet I'm not able to search the VSO using the `ExtentType` attr!

As an example, the following fails,

```Python
from sunpy.net import Fido, attrs as a
q = Fido.search(
    a.Time('2020-11-09T18:19:00', '2020-11-09T18:28:00'),
    a.Instrument('XRT'),
    a.ExtentType('FULLDISK')
)
```
with
```python-traceback
[...]

PR #4055, opened 2020-04-23
Created Fido Client for Gong synoptic maps

abhijeetmanhas:
### Description

Fixes https://github.com/dstansby/pfsspy/issues/165

This will enable users to search and download the meged data product for GONG using Fido, giving the timerange. The url pattern is like this,

`https://gong2.nso.edu/oQR/zqs/%Y%m/mrzqs%y%m%d/mrzqs%y%m%dt%H%Mc\d4_\d3.fits.gz`
Here `\d4` and `\d3` are Carringtion rotation number and Initital carr-long for synoptic map respectively.

Used scraper partially, I can't write client only using scraper since it scraper needs everything to be defined in the pattern, which won't be the case for these two attributes. It doesn't supports custom regex.

### DIFF
--- a/changelog/4055.feature.rst
+++ b/changelog/4055.feature.rst
@@ -1,2 +1,1 @@
-Created `~sunpy.net.dataretriever.sources.GongSynopticClient` for accessing merged data product
-for GONG Synoptic Maps.
+Created `~sunpy.net.dataretriever.sources.GongSynopticClient` for accessing Magnetogram Synoptic Map Archives of NSO-GONG.
--- a/changelog/4055.feature.rst
+++ b/changelog/4055.feature.rst
@@ -1,1 +1,1 @@
-Created `~sunpy.net.dataretriever.sources.GongSynopticClient` for accessing Magnetogram Synoptic Map Archives of NSO-GONG.
+Created `~sunpy.net.dataretriever.sources.GONGClient` for accessing Magnetogram Synoptic Map Archives of NSO-GONG.
--- a/changelog/4055.feature.rst
+++ b/changelog/4055.feature.rst
@@ -1,0 +1,2 @@
+Created `~sunpy.net.dataretriever.sources.GongSynopticClient` for accessing merged data product
+for GONG Synoptic Maps.
--- a/sunpy/net/attrs.py
+++ b/sunpy/net/attrs.py
@@ -20,6 +20,7 @@ attrs specific to them, under:
 * `a.vso <sunpy.net.vso.attrs>`
 * `a.jsoc <sunpy.net.jsoc.attrs>`
 * `a.goes <sunpy.net.dataretriever.attrs.goes>`
+* `a.gong_synoptic <sunpy.net.dataretriever.attrs.gong_synoptic>`
 
 """
 from ._attrs import (
--- a/sunpy/net/attrs.py
+++ b/sunpy/net/attrs.py
@@ -20,11 +20,11 @@ attrs specific to them, under:
 * `a.vso <sunpy.net.vso.attrs>`
 * `a.jsoc <sunpy.net.jsoc.attrs>`
 * `a.goes <sunpy.net.dataretriever.attrs.goes>`
-* `a.gong_synoptic <sunpy.net.dataretriever.attrs.gong_synoptic>`
 
 """
 from ._attrs import (
     Detector,
+    ExtentType,
     Instrument,
     Level,
     Physobs,
@@ -37,8 +37,8 @@ from ._attrs import (
 )
 
 # Trick the docs into thinking these attrs are defined in here.
-for _a in (Time, Instrument, Wavelength, Level, Sample, Detector, Resolution, Physobs, Source, Provider):
+for _a in (Time, Instrument, Wavelength, Level, Sample, Detector, Resolution, Physobs, Source, Provider, ExtentType):
     _a.__module__ = __name__
 
-__all__ = ['Time', 'Instrument', 'Wavelength', 'Level',
+__all__ = ['Time', 'Instrument', 'Wavelength', 'Level', 'ExtentType',
            'Sample', 'Detector', 'Resolution', 'Physobs', 'Source', 'Provider']
--- a/sunpy/net/dataretriever/sources/gong_synoptic.py
+++ b/sunpy/net/dataretriever/sources/gong_synoptic.py
@@ -1,5 +1,4 @@
 from sunpy.net.dataretriever import GenericClient
-from sunpy.util.scraper import Scraper
 
 __all__ = ['GongSynopticClient']
 
@@ -20,73 +19,18 @@ class GongSynopticClient(GenericClient):
     Results from 1 Provider:
     <BLANKLINE>
     3 Results from the GongSynopticClient:
-         Start Time           End Time      Source Instrument Wavelength
-    ------------------- ------------------- ------ ---------- ----------
-    2019-12-31 22:00:00 2020-01-01 02:00:00    NSO       GONG        nan
-    2019-12-31 22:00:00 2020-01-01 02:00:00    NSO       GONG        nan
-    2019-12-31 22:00:00 2020-01-01 02:00:00    NSO       GONG        nan
+         Start Time           End Time      Instrument ... Source ExtentType
+    ------------------- ------------------- ---------- ... ------ ----------
+    2019-12-31 22:14:00 2019-12-31 22:14:59       GONG ...    NSO   SYNOPTIC
+    2019-12-31 23:04:00 2019-12-31 23:04:59       GONG ...    NSO   SYNOPTIC
+    2019-12-31 23:54:00 2019-12-31 23:54:59       GONG ...    NSO   SYNOPTIC
     <BLANKLINE>
     <BLANKLINE>
 
     """
-
-    def _get_url_for_timerange(self, timerange, **kwargs):
-        """
-        Return URL(s) for corresponding timerange.
-
-        Parameters
-        ----------
-        timerange : `~sunpy.time.TimeRange`
-            The time range required.
-
-        Returns
-        -------
-        `list`
-            The URL(s) for the corresponding timerange.
-        """
-        prefix = r'https://gong2.nso.edu/oQR/zqs/%Y%m/mrzqs%y%m%d/'
-        pattern = prefix + r'mrzqs%y%m%dt%H%Mc(\d){4}_(\d){3}\.fits.gz'
-        gongsynoptic_files = Scraper(pattern, regex=True)
-        urls = gongsynoptic_files.filelist(timerange)
-        return urls
-
-    def _makeimap(self):
-        """
-        Helper Function:used to hold information about source.
-        """
-        self.map_['instrument'] = 'GONG'
-        self.map_['source'] = 'NSO'
-
-    @classmethod
-    def _can_handle_query(cls, *query):
-        """
-        Answers whether client can service the query.
-
-        Parameters
-        ----------
-        query : list of query objects
-
-        Returns
-        -------
-        bool : Answer as to whether client can service the query.
-        """
-        from sunpy.net import attrs as a
-
-        required = {a.Time, a.Instrument}
-        optional = {a.Physobs, a.gong_synoptic.ExtentType}
-        if not cls.check_attr_types_in_query(query, required, optional):
-            return False
-
-        matches = True
-        for x in query:
-            if isinstance(x, a.Instrument) and x.value.lower() != 'gong':
-                matches = False
-            if isinstance(x, a.Physobs) and x.value.lower() != 'los_magnetic_field':
-                matches = False
-            if isinstance(x, a.gong_synoptic.ExtentType) and x.value.lower() != 'synoptic':
-                matches = False
-
-        return matches
+    baseurl = (r'https://gong2.nso.edu/oQR/zqs/%Y%m/mrzqs%y%m%d/mrzqs%y%m%dt%H%Mc'
+               r'(\d){4}_(\d){3}\.fits.gz')
+    pattern = '{}/zqs/{year:4d}{month:2d}/mrzqs{:4d}{day:2d}/mrzqs{:6d}t{hour:2d}{minute:2d}c{}'
 
     @classmethod
     def _attrs_module(cls):
@@ -95,9 +39,8 @@ class GongSynopticClient(GenericClient):
     @classmethod
     def register_values(cls):
         from sunpy.net import attrs
-        adict = {attrs.Instrument: [
-            ("GONG", "Global Oscillation Network Group.")],
-            attrs.gong_synoptic.ExtentType: [
-            ("SYNOPTIC", "Coverage of a complete solar rotation synthesized over time")],
-            attrs.Physobs: [("LOS_MAGNETIC_FIELD", "Line of sight magnetic field")]}
+        adict = {attrs.Instrument: [("GONG", "Global Oscillation Network Group.")],
+                 attrs.Physobs: [("LOS_MAGNETIC_FIELD", "Line of sight magnetic field")],
+                 attrs.Source: [('NSO', 'National Solar Observatory.')],
+                 attrs.gong_synoptic.ExtentType: [("SYNOPTIC", "Coverage of a complete solar rotation synthesized over time")]}
         return adict
--- a/sunpy/net/dataretriever/sources/gong_synoptic.py
+++ b/sunpy/net/dataretriever/sources/gong_synoptic.py
@@ -57,76 +57,11 @@ class GongSynopticClient(GenericClient):
         `list`
             The URL(s) for the corresponding timerange.
         """
-        prefix = 'https://gong2.nso.edu/oQR/zqs/%Y%m/mrzqs%y%m%d/'
-        file_pattern = 'mrzqs\d{6}t\d{4}c\d+_\d+.fits.gz'
-        pref_scrap = Scraper(prefix)
-        directories = pref_scrap.range(timerange)
-        filesurls = []
-        for directory in directories:
-            try:
-                opn = urlopen(directory)
-                try:
-                    soup = BeautifulSoup(opn, "html.parser")
-                    for link in soup.find_all("a"):
-                        href = link.get("href")
-                        if re.match(file_pattern, href):
-                            # extracting timestamp from the filename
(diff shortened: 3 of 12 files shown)


abhijeetmanhas:
Main things to bee reviewed:

- The name of Client
- The `_makemap` routine, what more can be added ?
- Should add its own fetch function, which will extract the files too?
- Add a function to convert the fits data to pfsspy input compatible ?
- Should allow users to specify the `CAR_ROT` and `LONG0` ?
- `_can_handle_query` , what more `phsyobs`, `level`, etc should it support ?

Sunpy scraper should be able to handle custom regex, something by which we can solve issue https://github.com/sunpy/sunpy/issues/3337 by making scraper selecting only those goes_satellites for which data is present. Can be handled in different PR.

dstansby:
> Main things to bee reviewed:
> 
>     * The name of Client
Seems fine to me
> 
>     * The `_makemap` routine, what more can be added ?
> 
I'm not sure what this is, so pass...
>     * Should add its own fetch function, which will extract the files too?
> 
Do you mean un-zip them from `.fits.gz` to `.fits`?
>     * Add a function to convert the fits data to pfsspy input compatible ?
> 
That's out of scope for sunpy, so no
>     * Should allow users to specify the `CAR_ROT` and `LONG0` ?
> 
I think searching by time is fine
>     * `_can_handle_query` , what more `phsyobs`, `level`, etc should it support ?
> 
Again, just searching by time is fine. We can always expand this later.
> 
> Sunpy scraper should be able to handle custom regex, something by which we can solve issue #3337 by making scraper selecting only those goes_satellites for which data is present. but it can be handled in different PR.

abhijeetmanhas:
Yes, I was thinking of unzipping them into fits. Should we do it here? Or just leave gz files downloaded?

dstansby:
Ideally we want to unzip them, if possible.

Cadair:
I am pretty sure the SRS client handles unpacking if you want some prior art.

abhijeetmanhas:
@Cadair thanks for the suggestion.
I discovered your comment after pushing. SRS uses Tarfile library, I am using gzip. 
Which one would be better here?

abhijeetmanhas:
@dstansby the test failure is remote server error, rest all tests have passed. It is good for review now.

sunpy/net/dataretriever/sources/gong_synoptic.py
```diff
@@ -0,0 +1,157 @@
...
+        -------
+        boolean
+            answer as to whether client can service the query
+        """
+        chkattr = ['Time', 'Instrument']
+        chklist = [x.__class__.__name__ in chkattr for x in query]
+        for x in query:
+            if x.__class__.__name__ == 'Instrument' and x.value.lower() == 'gong':
+                return all(chklist)
+        return False
```
dstansby, line 157: It is possible to download different types of data from GONG, so I think we need to put some more thought into making `_can_handle_query` more specific. I think what we might need is an attribute that specifies we are looking for a synoptic map (as opposed to full disk observation), and an attribute that specifies that we are looking for magnetic field (more specifically, radial magnetic field) data. I think I will open this in a new issue, so people can see/comment on ideas for these attributes (and whether we need to make any new ones).

dstansby:
Thanks for all the work on this! On the technical side it looks 👍. As I've said in a comment, more thought needs to go in on the sunpy/Fido API side to come up with a more specific way to search for this data. I will open a new issue to discuss that.

nabobalis:
Would a specific attr help?

sunpy/net/dataretriever/sources/gong_synoptic.py
abhijeetmanhas: Thanks for the reivew, I have added physobs and extentType now.

abhijeetmanhas:
@nabobalis I have added optional `ExtentType` attrs like satellite no. in goes to differentiate b/w `FULL_DISK` and `SYNOPTIC` data, and tests for the same too. Also I have added optional `physobs` which should by `LOS_MAGNETIC_FIELD` as discussed with @dstansby .

sunpy/net/dataretriever/sources/gong_synoptic.py
abhijeetmanhas: @dstansby this is green now, can be reviewed, I have done all suggested changes till now.

sunpy/net/dataretriever/sources/gong_synoptic.py
```diff
@@ -0,0 +1,172 @@
...
+                    soup = BeautifulSoup(opn, "html.parser")
+                    for link in soup.find_all("a"):
+                        href = link.get("href")
+                        if re.match(file_pattern, href):
+                            # extracting timestamp from the filename
+                            timestring = href[5:16]
+                            tstamp = parse_time('20'+timestring.upper()+'00')
+                            if tstamp >= timerange.start and tstamp <= timerange.end:
+                                fullpath = directory + href
+                                filesurls.append(fullpath)
```
dpshelio, line 78: Shouldn't all that be "done" by the scrapper? I believe you can put a wildcard if you don't know how it goes (like the last digits)... Ok, you can't! but with my new pr #4107

sunpy/net/dataretriever/sources/gong_synoptic.py
```diff
@@ -0,0 +1,172 @@
...
+
+        for url, filename in zip(urls, paths):
+            downloader.enqueue_file(url, filename=filename)
+
+        resobj = downloader.download()
+
+        for loc_fpath, fpath in zip(local_paths, paths):
+            f = gzip.open(fpath, 'r')
+            g = gzip.open(loc_fpath, 'w')
+            g.write(f.read())
```
dpshelio, line 126: Why is this needed?

dpshelio: Aren't you removing the `.gz` on line 112? why are you then zipping it back without the `gz` suffix?

abhijeetmanhas: Yes, I removed `.gz` in line 112 to just get what the filename of extracted file should be. I haven't done any extraction there.
`local_fpath` contains path to future extracted file, which is currently empty when I opened it in write mode, and write `f` object to it, which contains the extracted file object in `fits`.

sunpy/net/dataretriever/sources/gong_synoptic.py
abhijeetmanhas: This PR will be very helpful. I'll change the code to depend on #4107 after it is merged.

sunpy/net/dataretriever/sources/gong_synoptic.py
dpshelio: sorry I've not commented earlier. I don't understand why you want to unzip the file. I believe fits, and map, reads zipped files fine. I'd prefer to keep the files zipped in my computer.

abhijeetmanhas: Cool, keeping it zipped then. Thanks for the review!

changelog/4055.feature.rst
```diff
@@ -0,0 +1,2 @@
+Created `~sunpy.net.dataretriever.sources.GongSynopticClient` for accessing merged data product
```
nabobalis, line 1: What is a merged data product?

abhijeetmanhas: Changed to 
```Created `~sunpy.net.dataretriever.sources.GongSynopticClient` for accessing Magnetogram Synoptic Map Archives of NSO-GONG.```.

**Comment by abhijeetmanhas** [comment]
`helio` Test failure is not related to the change.

File: sunpy/net/dataretriever/sources/tests/test_gong_synoptic.py [hunk 5]
```diff
@@ -0,0 +1,112 @@
...
+    "time, instrument",
+    [(a.Time('2019/10/4', '2019/10/4 2:00'), a.Instrument('gong')),
+     (a.Time('2019/12/31 21:00', '2020/1/1'), a.Instrument('gong'))])
+def test_fido(time, instrument):
+    qr = Fido.search(time, instrument)
+    assert isinstance(qr, UnifiedResponse)
+    response = Fido.fetch(qr)
+    assert len(response) == qr._numfile
+    for res in response:
+        assert res.split('.')[-1] == 'gz'
```
dpshelio, line 99: I don't really like `assert` on loops, if one fails, you only get a "report" of that one, and not whether others did too.
```suggestion
    assert all(map(lambda x: x.endswith('.gz'), response))
```

sunpy/net/dataretriever/sources/gong_synoptic.py
```diff
@@ -0,0 +1,105 @@
...
+from sunpy.util.scraper import Scraper
+
+__all__ = ['GongSynopticClient']
+
+
+class GongSynopticClient(GenericClient):
+    """
+    Provides access to the Magnetogram products of NSO-GONG synoptic Maps.
+
+    Searches data hosted by the `National Solar Observatory <gong2.nso.edu/oQR/zqs/>`__
```
nabobalis, line 11:
```suggestion
    Searches data hosted by the `National Solar Observatory. <gong2.nso.edu/oQR/zqs/>`__
```

sunpy/net/dataretriever/sources/gong_synoptic.py
```diff
@@ -0,0 +1,105 @@
...
+
+    Examples
+    --------
+
```
nabobalis, line 15:
```suggestion
```

sunpy/net/dataretriever/sources/gong_synoptic.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+
+    def _get_url_for_timerange(self, timerange, **kwargs):
+        """
+        Return URL(s) for corresponding timerange.
+
+        Parameters
+        ----------
+        timerange : `~sunpy.time.TimeRange`
+            The time range you want the files for.
```
nabobalis, line 41:
```suggestion
            The time range required.
```

sunpy/net/dataretriever/sources/gong_synoptic.py
```diff
@@ -0,0 +1,105 @@
...
+        """
+        Answers whether client can service the query.
+
+        Parameters
+        ----------
+        query : list of query objects
+
+        Returns
+        -------
+        boolean
```
nabobalis, line 72:
```suggestion
        bool : 
```

sunpy/net/dataretriever/sources/gong_synoptic.py
```diff
@@ -0,0 +1,105 @@
...
+            answer as to whether client can service the query
```
nabobalis, line 73:
```suggestion
            Answer as to whether client can service the query,
```

sunpy/net/dataretriever/sources/gong_synoptic.py
```diff
@@ -0,0 +1,105 @@
...
+        """
+        from sunpy.net import attrs as a
```
nabobalis, line 75: Does this need to be in the function?

abhijeetmanhas: I thought it prevents circular imports.

nabobalis: Can you check?

abhijeetmanhas: yes, it prevents. https://github.com/sunpy/sunpy/blob/master/sunpy/net/dataretriever/sources/goes.py#L380 similar to what we did for `goes` and `norh`.

sunpy/net/dataretriever/sources/gong_synoptic.py
```diff
@@ -0,0 +1,103 @@
...
+        if not cls.check_attr_types_in_query(query, required, optional):
+            return False
+
+        matches = True
+        for x in query:
+            if isinstance(x, a.Instrument) and x.value.lower() != 'gong':
+                matches = False
+            if isinstance(x, a.Physobs) and x.value.lower() != 'los_magnetic_field':
+                matches = False
+            if isinstance(x, a.gong_synoptic.ExtentType) and x.value.lower() != 'synoptic':
```
Cadair, line 86: Why do we have an attr for a thing we only support one type of?

dstansby: I was thinking that we could re-use the VSO `EXTENT_TYPE` keyword here, but it's currently broken: https://github.com/sunpy/sunpy/issues/4048

Cadair: If we don't have any use for this attr I suggest we don't implement it at this time. The extent type for this new client should be set in the `_map` metadata in the client class.

abhijeetmanhas: I added extentType after this https://github.com/sunpy/sunpy/pull/4055#discussion_r414496406 . We can either remove it too.

Cadair: Is gong data with different extent types available through the VSO? but we can't filter on it because of #4048 ?

dstansby: No, I don't think GONG synoptic maps are on the VSO, but I think it would be good if we can re-purpose the EXTENT_TYPE attr in VSO for search (ie. move it from a vso specific attr to a general one like `Time` or `Instrument`) to avoid creating new attrs.

dstansby: Okay, I did some digging/research and the EXTENT_TYPE attr for VSO is properly broken, so I think including a new attr is fine here.

If it was working then these are the valid values: https://sdac.virtualsolar.org/cgi/show_details?keyword=EXTENT_TYPE - @Cadair do you think it's worth putting the new attr in `sunpy.net.attrs` instead of in `sunpy.net.gong_synoptic.attrs`?

sunpy/net/dataretriever/sources/gong_synoptic.py
```diff
@@ -0,0 +1,103 @@
...
+    >>> results  #doctest: +REMOTE_DATA
+    <sunpy.net.fido_factory.UnifiedResponse object at ...>
+    Results from 1 Provider:
+    <BLANKLINE>
+    3 Results from the GongSynopticClient:
+         Start Time           End Time      Source Instrument Wavelength
+    ------------------- ------------------- ------ ---------- ----------
+    2019-12-31 22:00:00 2020-01-01 02:00:00    NSO       GONG        nan
+    2019-12-31 22:00:00 2020-01-01 02:00:00    NSO       GONG        nan
+    2019-12-31 22:00:00 2020-01-01 02:00:00    NSO       GONG        nan
```
dstansby, line 27: This result looks odd, do you know why this is picking up three files, but all with the same timestamp?

sunpy/net/dataretriever/sources/gong_synoptic.py
dstansby: Going to resolve this conversation, and move discussion on where this new attr should go to an issue.

dstansby:
@nabobalis which PR is this waiting on?

nabobalis:
> @nabobalis which PR is this waiting on?

https://github.com/sunpy/sunpy/pull/4321

Cadair:
#4321 has been merged so this also needs a rewrite.

sunpy/net/dataretriever/sources/gong_synoptic.py
```diff
@@ -0,0 +1,46 @@
...
+    def _attrs_module(cls):
+        return 'gong_synoptic', 'sunpy.net.dataretriever.attrs.gong_synoptic'
+
+    @classmethod
+    def register_values(cls):
+        from sunpy.net import attrs
+        adict = {attrs.Instrument: [("GONG", "Global Oscillation Network Group.")],
+                 attrs.Physobs: [("LOS_MAGNETIC_FIELD", "Line of sight magnetic field")],
+                 attrs.Source: [('NSO', 'National Solar Observatory.')],
+                 attrs.gong_synoptic.ExtentType: [("SYNOPTIC", "Coverage of a complete solar rotation synthesized over time")]}
```
nabobalis, line 45: I don't have a better suggestion but `attrs.gong_synoptic.ExtentType` seems a bit long.

sunpy/net/dataretriever/sources/gong_synoptic.py
abhijeetmanhas: Fixed now.

sunpy/net/dataretriever/sources/gong_synoptic.py
abhijeetmanhas: Can we just remove it?
It will only happen that there won't be an extra column of name `extent type` in results whoes value is "synoptic". 
Nothing else would change.

dstansby: I think it would be reasonable to move `ExtentType` out of `attrs.gong_synoptic` and into `attrs` as a more generic attribute, don't know what others think about that?

nabobalis: IF we need to use it for this client, I would be in favour of that instead of it being for this client only.

dstansby: Given one can also get full disc magnetograms from GONG, I actually think we should rename the client from `gong_synoptic` > `gong`, and keep this keyword as a way to differentiate between a full disk map and a synoptic map. Then we avoid having separate `gong` and a `gong_synoptic` clients.

abhijeetmanhas: Replaced gongsynoptic with gong and moved extentType to attrs.

sunpy/net/attrs.py
```diff
@@ -20,10 +20,12 @@
 * `a.vso <sunpy.net.vso.attrs>`
 * `a.jsoc <sunpy.net.jsoc.attrs>`
 * `a.goes <sunpy.net.dataretriever.attrs.goes>`
+* `a.gong_synoptic <sunpy.net.dataretriever.attrs.gong_synoptic>`
```
nabobalis, line 23:
```suggestion
```

nabobalis (approved):

kakirastern (approved):

dstansby (approved):

PR #4056, opened 2020-04-27
Update JSOC warning

dstansby:
Replaces #4050

- "Atleast" > "At least"
- pkstr contains the time and wavelength too, so passing one of these is enough to fix the error. Mention this in the error message.
- Add a comment explaining why we don't allow searching only by series (see https://github.com/sunpy/sunpy/pull/4050#issuecomment-619166920)

### DIFF
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -787,8 +787,11 @@ class JSOCClient(BaseClient):
 
         if not pkstr:
             # pkstr cannot be totally empty
-            error_message = "Atleast one PrimeKey must be passed."
-            raise ValueError(error_message)
+            #
+            # Note that whilst it is technically posisble to just search by series,
+            # this is not allowed here, because some of these would be very large
+            # searches that would make JSOC sad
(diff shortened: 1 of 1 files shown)


sunpy/net/jsoc/jsoc.py
```diff
@@ -787,8 +787,11 @@ def _make_recordset(self, series, start_time='', end_time='', wavelength='',
 
         if not pkstr:
             # pkstr cannot be totally empty
-            error_message = "Atleast one PrimeKey must be passed."
-            raise ValueError(error_message)
+            #
+            # Note that whilst it is technically posisble to just search by series,
+            # this is not allowed here, because some of these would be very large
+            # searches that would make JSOC sad
+            raise ValueError("At least one PrimeKey, time, or wavelength must be passed.")
```
Cadair, line 794: This might confuse people who actually understand JSOC as Time and Wavelength are pkeys, we just expose them differently in the search. Not 100% on this reword but hows:
```suggestion
            raise ValueError("Time, Wavelength or an explicit PrimeKey must be specified.")
```

Issue #4057, opened 2020-04-27
Fido has some issues to download into current working directory


hayesla:
### Description

When downloading results from a Fido search it's not intuitive on how to save to your current working directory. For example:

```
>>> from sunpy.net import Fido, attrs as a
>>> result = Fido.search(a.Time("2013-10-28", "2013-10-29"), a.Instrument("XRS"))
>>> Fido.fetch(result, path="./")

~/sunpy_dev/sunpy/sunpy/net/fido_factory.py in fetch(self, path, max_conn, progress, overwrite, downloader, *query_results, **kwargs)
    334         if path is not None:
    335             exists = list(filter(lambda p: p.exists(), Path(path).parents))
--> 336             if not os.access(exists[0], os.W_OK):
    337                 raise PermissionError('You do not have permission to write'
    338                                       f' to the directory {exists[0]}.')

IndexError: list index out of range
```
This can be fixed by including the file name from the client such that:
```
>>> Fido.fetch(result, path="./{file}")
```

Is there a reason that you have to define the file name from the client rather than just using `./`?

If so, we should raise an error that says this. Otherwise can we fix it to allow `./` to be allowed?

Cadair:
I see no reason why that shouldn't work.

hayesla:
The issue is in the checking the write access - the path `./` returns an empty list in here `exists = list(filter(lambda p: p.exists(), Path(path).parents))` and so breaks

Otherwise it works - (i.e. if you comment out this condition)

Cadair:
I think that could be:
```
exists = list(filter(lambda p: p.exists(), Path(path).resolve().parents))
```

hayesla:
Yep - that fixes it! 🥳 

Shall I make a PR to fix this?

nabobalis:
Yeah and add a test if you can.

hayesla:
will do!

PR #4058, opened 2020-04-27
Fixing bug in Fido to allow download to path="./" 

hayesla:
### Description
Small fix to bug within `fido_factory.py` to allow download to `path = "./"` in `Fido.fetch()`.

I think I might need some help/advice on the test written for this using the pytest `tmpdir` or `tmp_path`

Fixes #<Issue Number>
#4057


sunpy/net/tests/test_fido.py
```diff
@@ -124,6 +124,15 @@ def test_save_path_pathlib(tmpdir):
         assert f"eve{os.path.sep}0" in f
 
 
+@pytest.mark.remote_data
+def test_save_path_cwd(tmpdir):
+    qr = Fido.search(a.Instrument('EVE'), a.Time("2016/10/01", "2016/10/02"), a.Level(0))
+
+    # Test when path is ./ for current working directory
+    files = Fido.fetch(qr, path="./")
```
nabobalis, line 132:
```suggestion
    files = Fido.fetch(qr, path=tmpdir.strpath+"./")
```
I think?

hayesla: hmm will this work? this seems to be creating a path with just a ./ at the end. I'm a bit confused about `tmpdir` - is this just a temporary directory created for the test?

Also then how would you test it's in your current directory?

hayesla: Also silly question - is it currently passing or is are the tests just skipping online tests?

nabobalis: Yeah `tmpdir` create a temporary directory created for the test. 

I don't actually think we need to use tmpdir here, it should be ok as it.

The windows CI build is failing, not sure why its missing from the checks at the bottom of the PR but it is there on the commit.

sunpy/net/tests/test_fido.py
```diff
@@ -124,6 +124,15 @@ def test_save_path_pathlib(tmpdir):
         assert f"eve{os.path.sep}0" in f
 
 
+@pytest.mark.remote_data
+def test_save_path_cwd(tmpdir):
```
nabobalis, line 128:
```suggestion
def test_save_path_cwd():
```

sunpy/net/tests/test_fido.py
Cadair: We should change directory into the tmp dir before running fetch. Otherwise this will pollute whatever directory the tests are run in.

hayesla: what about this ? I think this might work ...

hayesla:
does this need a changelog?

nabobalis:
I think it does since its a bugfix.

sunpy/net/tests/test_fido.py
nabobalis: Looks good to me.

changelog/4058.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixes a bug in fido_factory to allow  path="./" in fido.fetch().
```
Cadair, line 1: The codestyle check thinks there is a stray end of line here: https://dev.azure.com/sunpy/sunpy/_build/results?buildId=7271&view=logs&j=4cdc857c-0d48-50f8-2772-3b048c8bb2f3&t=fe62049c-f9e9-5b31-8bb6-a78aae0c6ebe&l=40

dstansby:
I just fixed the PEP8

sunpy/net/tests/test_fido.py
```diff
@@ -124,6 +124,16 @@ def test_save_path_pathlib(tmpdir):
...
+    qr = Fido.search(a.Instrument('EVE'), a.Time("2016/10/01", "2016/10/02"), a.Level(0))
+
+    # Test when path is ./ for current working directory
+    os.chdir(tmpdir)  # move into temp directory
```
dstansby, line 132: Do we want to exit the directory after the test is done?

nabobalis: I was gonna hope/pray that it doesn't matter since the test is done and it all resets ok.

wtbarnes: Is there not a way to do directory changes with pytest to ensure that nothing blows up?

Cadair: :shrug:

hayesla: its all a mystery ...

dstansby: The tests don't seem to have blown up so shall we just merge this?

nabobalis: Yeah

Issue #4051, opened 2020-04-22
Unskip gallery examples before 2.0


Cadair:
A couple of examples are skipped, including the reproject one, we should unskip them before release

dstansby:
Are these skipped just to avoid running them over and over again between releases, or because they're broken?

nabobalis:
We normally skip them if they are broken.

PR #4060, opened 2020-04-29
Unskip non-reprojection examples

dstansby:
xref #4051 - these work fine locally for me.


nabobalis:
Let's hope the CI agrees.

nabobalis:
```
generating gallery for generated/gallery/plotting... [ 87%] solar_cycle_example.py 
Exception occurred:
 
  File "/root/project/.tox/build_docs/lib/python3.7/site-packages/matplotlib/axis.py", line 1576, in convert_units
 
    f'units: {x!r}') from e
 
matplotlib.units.ConversionError: Failed to convert value(s) to axis units: '2011-01-01'
 
```

Should unskip any others. If we hit memory limits on rtd, we can worry about that later.

mbobra:
This works for me locally too. However, [this comment](https://github.com/sunpy/sunpy/pull/3765#issuecomment-582991873) in #3765 pointed out it fails the doc builds (even though I see it passed here). What was the problem?

dstansby:
I un-un-skipped the solar cycle example, will try and fix that elsewhere

PR #4061, opened 2020-04-29
Simplify solar cycle example

dstansby:
This should make it pass hopefully. I got rid of some of the complicated bits that were probably causing the test to fail, but this still shows how to plot the solar cycle from the sample data. To be honest we should probably replace this with how to download/plot current data, but for now this should be fine.


nabobalis:
Do we have any left skipped examples now?

nabobalis:
We also get 
```
/root/project/.tox/build_docs/lib/python3.7/site-packages/sunpy/timeseries/timeseriesbase.py:126: SunpyUserWarning: Using .data to access the dataframe is discouraged; use .to_dataframe() instead.
  "use .to_dataframe() instead.", SunpyUserWarning)
/
``` 
do we want to fix that?

dstansby:
The reprojection ones are still skipped.

nabobalis:
Should we just skip those?

dstansby:
This is a quick fix to make the example run, so I'd rather not add anything new here.

nabobalis:
I opened an issue here about that https://github.com/sunpy/sunpy/issues/4082

PR #4080, opened 2020-05-01
Un-skip reprojection examples

### LINKED ISSUE #4051: Unskip gallery examples before 2.0

dstansby:
Fixes #4051. This is the last of the un-skips. I also took the oppurtunity to re-use the same search/data for these two examples, to save having to download two different sets of maps. I verified locally that the changed example still works fine.


examples/map_transformations/reprojection_different_observers.py
```diff
@@ -29,16 +29,17 @@
 plt.rcParams['figure.figsize'] = (16, 8)
 
 ######################################################################
-# Let’s download an EUV image from both AIA and STEREO A, when their
-# separation was around 90 degrees.
+# Let’s download an EUV image from both AIA and STEREO A
```
nabobalis, line 32: Do we not want `# separation was around 90 degrees.`?

dstansby: It isn't any more, because I changed the images so it's more like 120 degrees (off the top of my head). Can edit it to 120 deg instead?

nabobalis: If you know what it is sure, if not leave it.

ayshih:
Don't merge yet

ayshih:
The 4x resampling/rescaling bug wasn't ever actually fixed in this gallery example, so it's now fixed.  Now the Sun isn't freakishly zoomed out.

nabobalis:
Was that the only example with that issue?

ayshih:
I don't know.  The issue was because a resampling step was added to reduce memory consumption, but the rest of the example still assumed the original pixel scale.

ayshih:
Nothing else jumps out at me.

nabobalis:
When the docs build, I will check them before I merge.

Cadair:
In the observers example we are saying "AIA" but using an "EUVI" image?

ayshih:
There's both AIA and EUVI, but things are borked in the example.  I'll fix it.

examples/map_transformations/reprojection_different_observers.py
```diff
@@ -29,52 +32,55 @@
...
 ######################################################################
-# Create a map for each image.
+# Create a map for each image, after making sure to sort by the
+# appropriate name attribute (i.e., "AIA" and "EUVI") so that the
+# order is reliable.
 
-map_aia, map_stereo = sunpy.map.Map(sorted(files))
+map_list = sunpy.map.Map(files)
+map_list.sort(key=lambda m: m.detector)
+map_aia, map_euvi = map_list
```
ayshih, line 58: This approach is a little fancy, of course, but it seems like the most robust approach (i.e., it doesn't depend on filenames).

ayshih:
> In the observers example we are saying "AIA" but using an "EUVI" image?

@Cadair All fixed now (see [recent build](https://48341-2165383-gh.circle-artifacts.com/0/docs/_build/html/generated/gallery/map_transformations/reprojection_different_observers.html#sphx-glr-generated-gallery-map-transformations-reprojection-different-observers-py))

dstansby:
Thanks for the fixes @ayshih !

PR #4084, opened 2020-05-03
Use to_dataframe() in goes.py

dstansby:
Catching a few more instances where using `.data` is now discouraged/throws a warning, and converting them to using `to_dataframe()`.

### DIFF
--- a/sunpy/instr/goes.py
+++ b/sunpy/instr/goes.py
@@ -191,8 +191,8 @@ def calculate_temperature_em(goests, abundances="coronal",
         Contains same metadata and data as input timeseries with the
         following two additional data columns:
 
-        | ts_new.data.temperature - Array of temperatures [MK]
-        | ts_new.data.em - Array of volume emission measures [cm**-3]
+        | ts_new.to_dataframe().temperature - Array of temperatures [MK]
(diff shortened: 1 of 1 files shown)


wtbarnes:
I have absolutely no problem with this change, but I wonder how we should handle development on the `instr` subpackage moving forward. Most of this functionality has been moved to [sunpy/sunkit-instruments](https://github.com/sunpy/sunkit-instruments).

Should development happen in both places? Can we automatically mirror PRs on this subpackage to that repo or *vice-versa*? Should we freeze development here and move all development over to `sunkit-instruments`?

dstansby:
Oh, if that's the case shouldn't we deprecate all this stuff in sunpy core (like with aiapy?)

wtbarnes:
I'm not sure how the interaction between core and `sunkit-instruments` should look like. We briefly discussed this at the coordination meeting I think, but I don't remember if we came up with a concrete solution. This should be a topic at the next telecon.

wtbarnes:
Oops meant to approve that first 🤷 

Didn't think that discussion should block this being merged. I've added a point to the agenda for the next meeting that essentially encapsulates my above point.

PR #4085, opened 2020-05-03
Make _observers_are_equal compatible with 4.1

dstansby:
https://github.com/astropy/astropy/pull/10154 has made coordinate equality checking more stringent, which has broken `_observers_are_equal` (e.g. see https://dev.azure.com/sunpy/sunpy/_build/results?buildId=7437&view=logs&jobId=d3468eda-fa0c-5333-fa0e-f689a94bf857&j=7f33e5bd-7764-5d8a-ba2e-506e078b9c3f).

This PR fixes it to make it compatible with astropy 4.1. According to the astropy changelog `==` used to be equivalent to `is`, so that's the change I've made.

### DIFF
--- a/sunpy/coordinates/transformations.py
+++ b/sunpy/coordinates/transformations.py
@@ -175,7 +175,7 @@ def _transformation_debug(description):
 
 def _observers_are_equal(obs_1, obs_2):
     # Note that this also lets pass the situation where both observers are None
-    if obs_1 == obs_2:
+    if obs_1 is obs_2:
         return True
 
     # obs_1 != obs_2


sunpy/coordinates/transformations.py
```diff
@@ -174,17 +174,23 @@ def wrapped_func(*args, **kwargs):
...
         raise ConvertError("The source observer is set to None, but the destination observer is "
                            f"{obs_2}.")
     if obs_2 is None:
         raise ConvertError("The destination observer is set to None, but the source observer is "
                            f"{obs_2}.")
+    # Make sure obs_1 and obs_2 are in the same frame
+    if not np.all(obs_1.is_equivalent_frame(obs_2)):
+        return False
+
+    if np.all(obs_1 == obs_2):
```
dstansby, line 190: This now returns a boolean array if there are multiple coordinates, hence the `np.all` call.

sunpy/coordinates/transformations.py
```diff
@@ -174,17 +174,23 @@ def wrapped_func(*args, **kwargs):
...
-
-    # obs_1 != obs_2
     if obs_1 is None:
         raise ConvertError("The source observer is set to None, but the destination observer is "
                            f"{obs_2}.")
     if obs_2 is None:
         raise ConvertError("The destination observer is set to None, but the source observer is "
                            f"{obs_2}.")
+    # Make sure obs_1 and obs_2 are in the same frame
+    if not np.all(obs_1.is_equivalent_frame(obs_2)):
```
dstansby, line 187: Previously astropy would run this check, and return False, but now astropy raises an error instead, so we have to check for this ourselves.

sunpy/coordinates/transformations.py
```diff
@@ -174,17 +174,23 @@ def wrapped_func(*args, **kwargs):
 
 
 def _observers_are_equal(obs_1, obs_2):
-    # Note that this also lets pass the situation where both observers are None
-    if obs_1 == obs_2:
+    # Handle None cases
+    if obs_1 is None and obs_2 is None:
```
dstansby, line 178: In order to do the `is_equivalent_frame` check below, all the `None` cases have to be dealt with first.

dstansby:
I've tagged for backport - I'm not sure what the sunpy policy is on astropy version support, but it should be pretty easy to backport.

Cadair:
Is this PR supposed to fix the `test_hpc_hgs_implicit_hcc ` test?

dstansby:
Nope, that was done by @ayshih upstream

PR #4089, opened 2020-05-04
Fix a doctest to work on multiple versions of Astropy

ayshih:
Because of astropy/astropy#10232, the traceback and error message has slightly changed when supplying something that not a `Quantity` when units are required.  This PR enables the doctest to pass both before and after this change.


ayshih:
Ugh...

ayshih:
Now good to go.  Rebased on #4085 to confirm that it works, so merge that PR first.

Issue #3865, opened 2020-03-06
sunpy.data.manager API docs missing


dstansby:
https://docs.sunpy.org/en/latest/code_ref/data.html contains no docs for `sunpy.data.manager` - there should be some there, which should be linked to from https://docs.sunpy.org/en/stable/generated/gallery/remote_data_manager.html.

Cadair:
Argh, I thought I fixed this ages ago.

Cadair:
Oh. The docs that I expect to be there are [here](https://docs.sunpy.org/en/latest/api/sunpy.data.data_manager.DataManager.html#sunpy.data.data_manager.DataManager) `sunpy.data.manager` is an instance of that class (iirc) so much like `Map` and `Fido` is a right PITA to make sphinx document properly. What needs to happen is the links to `sunpy.data.manager` needs to be updated to the same format as the links for `Map` and `Fido` as described here: https://docs.sunpy.org/en/latest/dev_guide/documentation.html#sunpy-specific-rules

amogh-jrules:
I'll do it 👍

PR #4098, opened 2020-05-05
Document variables (e.g. Map, Fido, TimeSeries...)

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/3865. Because `sunpy.data.{manager,cache}` are variables, they need an extra option to make `automodapi` document them.

### DIFF
--- a/changelog/4098.docfix.rst
+++ b/changelog/4098.docfix.rst
@@ -1,0 +1,2 @@
+Added `sunpy.data.manager`, `sunpy.data.cache`, `sunpy.net.Fido`, `sunpy.map.Map`,
+and `sunpy.timeseries.TimeSeries` to the docs.
--- a/docs/code_ref/data.rst
+++ b/docs/code_ref/data.rst
@@ -5,6 +5,7 @@ The SunPy data module contains ways to access sample data and small test files
 for running the SunPy test suite.
 
 .. automodapi:: sunpy.data
+   :include-all-objects:
 
 .. automodapi:: sunpy.data.sample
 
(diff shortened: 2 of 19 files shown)


dstansby:
See https://48805-2165383-gh.circle-artifacts.com/0/docs/_build/html/code_ref/data.html for the variables showing up in the docs.

Cadair:
ooo, nice! Can we do this for `Fido`, `Map` and `Timeseries` as well?

Cadair:
(whoops wrong button)

dstansby:
I presume so? Shall I do it here?

Cadair:
This is awesome, so glad you found the answer to this :smile: 

Can you update this section of the dev docs as well: https://github.com/sunpy/sunpy/blob/master/docs/dev_guide/documentation.rst#sunpy-specific-rules as it presumably doesn't matter any more.

ayshih:
It's not the fault of this PR, but we should revise the docstrings with actual cross-linking because they all are dead-ends:

- https://48905-2165383-gh.circle-artifacts.com/0/docs/_build/html/api/sunpy.map.Map.html
- https://48905-2165383-gh.circle-artifacts.com/0/docs/_build/html/api/sunpy.net.Fido.html (has basically no content)
- https://48905-2165383-gh.circle-artifacts.com/0/docs/_build/html/api/sunpy.timeseries.TimeSeries.html

Cadair:
The Map and Timeseries ones look about as I would expect them to. What are you suggesting @ayshih ?

ayshih:
I mean "dead ends" in the sense that it's not clear where to go to find out more about what comes out of these entities.  For example, Map should provide a description of what it actually returns and an intersphinx link to `GenericMap` to be able to jump to methods and attributes.  I think TimeSeries tries to do that, but it just links to itself.  It wasn't so a problem before since these used to be embedded in surrounding text, but now these can be pulled up in isolation.

Cadair:
I will re-write the docstrings and push.

PR #4100, opened 2020-05-05
ran unify_section_headings.py

nabobalis:
Maybe?


Cadair:
I approve of this in theory but can we customise the ordering it uses to minimize the changes? Also can it be a pre-commit hook?

nabobalis:
We can create a copy of the script and change it?

Cadair:
where is the script?

nabobalis:
https://github.com/astropy/astropy-procedures/blob/master/discontinued_usage/unify_section_headings.py

Cadair:
I feel like we haven't historically had `*` at the top level of our titles?

nabobalis:
We used whatever? I'm not 100% what is the official ordering of headers (even if there is one)

Cadair:
> We used whatever?

yeah but I think we had `=` as the top level more often than not? I do wonder how this interacts with the default heading levels in `automodapi`?

nabobalis:
Badly since the doc build failed.

I can change the order to a version we can agree on?

Cadair:
Well this ordering from astropy is likely to fit with the defaults where `automodapi` is used like how astropy uses it. If I had to guess I would say that we use `=` as our top level in most places so that might make it a smaller change?

PR #4102, opened 2020-05-06
Improve support for velocities in coordinates

ayshih:
- Our frames now define the default representation of the velocity, so the velocity (if present) will now show up in the text repr of a coordinate.
- The ephemeris functions `get_body_heliographic_stonyhurst()`, `get_earth()`, and `get_horizons_coord()` all have a toggle `include_velocity` to include the velocity of the body in the output coordinate.

```python
>>> get_body_heliographic_stonyhurst('mars', '2001-02-03', include_velocity=True)
<HeliographicStonyhurst Coordinate (obstime=2001-02-03T00:00:00.000): (lon, lat, radius) in (deg, deg, AU)
    (63.03105777, -5.20656151, 1.6251161)
 (d_lon, d_lat, d_radius) in (arcsec / s, arcsec / s, km / s)
    (0.007552, 0.00037353, -28.43105538)>

>>> get_horizons_coord('Juno', '1995-07-18 07:17', 'smallbody', include_velocity=True)
INFO: Obtained JPL HORIZONS location for 3 Juno [sunpy.coordinates.ephemeris]
<SkyCoord (HeliographicStonyhurst: obstime=1995-07-18T07:17:00.000): (lon, lat, radius) in (deg, deg, AU)
    (-25.16107572, 14.59098456, 3.17667662)
 (d_lon, d_lat, d_radius) in (arcsec / s, arcsec / s, km / s)
    (-0.00514936, -0.00205857, 8.89781348)>
```

### DIFF
--- a/sunpy/coordinates/ephemeris.py
+++ b/sunpy/coordinates/ephemeris.py
@@ -5,8 +5,18 @@ import numpy as np
 
 import astropy.units as u
 from astropy.constants import c as speed_of_light
-from astropy.coordinates import ICRS, HeliocentricEclipticIAU76, SkyCoord, get_body_barycentric
-from astropy.coordinates.representation import CartesianRepresentation
+from astropy.coordinates import (
+    ICRS,
+    HeliocentricEclipticIAU76,
+    SkyCoord,
+    get_body_barycentric,
+    get_body_barycentric_posvel,
+)
+from astropy.coordinates.representation import (
+    CartesianDifferential,
+    CartesianRepresentation,
+    SphericalRepresentation,
+)
 
 from sunpy import log
 from sunpy.time import parse_time
@@ -22,7 +32,7 @@ __all__ = ['get_body_heliographic_stonyhurst', 'get_earth',
 
 
 @add_common_docstring(**_variables_for_parse_time_docstring())
-def get_body_heliographic_stonyhurst(body, time='now', observer=None):
+def get_body_heliographic_stonyhurst(body, time='now', observer=None, *, include_velocity=False):
     """
     Return a `~sunpy.coordinates.frames.HeliographicStonyhurst` frame for the location of a
     solar-system body at a specified time.  The location can be corrected for light travel time
@@ -39,6 +49,11 @@ def get_body_heliographic_stonyhurst(body, time='now', observer=None):
         If not None, the returned coordinate is the astrometric location (i.e., accounts for light
         travel time to the specified observer)
 
+    Keyword Arguments
+    -----------------
+    include_velocity : `bool`
+        If True, include the body's velocity in the output coordinate.  Defaults to False.
+
     Returns
     -------
     out : `~sunpy.coordinates.frames.HeliographicStonyhurst`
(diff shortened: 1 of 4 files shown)


sunpy/coordinates/ephemeris.py
```diff
@@ -74,37 +126,72 @@ def get_body_heliographic_stonyhurst(body, time='now', observer=None):
...
+        If True, include the Earth's velocity in the output coordinate.  Defaults to False.
 
     Returns
     -------
     out : `~astropy.coordinates.SkyCoord`
         Location of the Earth in the `~sunpy.coordinates.frames.HeliographicStonyhurst` frame
+
+    Notes
+    -----
+    The Earth's velocity in the output coordinate will invariably be negligible because the
```
dstansby, line 161:
```suggestion
    The Earth's velocity in the output coordinate will invariably be zero because the
```
?

sunpy/coordinates/ephemeris.py
```diff
@@ -38,6 +48,8 @@ def get_body_heliographic_stonyhurst(body, time='now', observer=None):
         If None, the returned coordinate is the instantaneous or "true" location.
         If not None, the returned coordinate is the astrometric location (i.e., accounts for light
         travel time to the specified observer)
+    include_velocity : `bool`
+        If True, include the body's velocity in the output coordinate.  Defaults to False.
```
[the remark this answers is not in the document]
ayshih: Yes and yes:

- Using a coordinate with a velocity can lead to warnings/errors for certain operations that do not result in warning/errors when using a coordinate without a velocity.
- Calculating the velocity takes additional computation time on some ephemeris selections.

sunpy/coordinates/ephemeris.py
ayshih: The velocity should be zero, but it generally won't be *exactly* zero because:

- The rotation of the frame is calculated via finite-difference methods on the positions of Earth, which will always have some inaccuracy compared to the actual velocity of Earth.
- For the built-in Astropy ephemeris, the returned velocity is inaccurate compared to the derivative of the returned positions.  I think this is documented somewhere.  The JPL ephemeris is fine.
- Unavoidable numerical noise

That's why I'm reluctant from sounding like it's a promise that it will be zero.  I also don't want to force it to be zero because that may lead to inconsistencies.

ayshih:
> Looks like this still needs some tests and codestyle fixes.

Technically it's thoroughly tested (against regressions) via doctests, but I'll grant you that I didn't bother to write separate unit tests.

sunpy/coordinates/ephemeris.py
```diff
@@ -22,7 +32,7 @@
 
 
 @add_common_docstring(**_variables_for_parse_time_docstring())
-def get_body_heliographic_stonyhurst(body, time='now', observer=None):
+def get_body_heliographic_stonyhurst(body, time='now', observer=None, include_velocity=False):
```
Cadair, line 35: The `include_velocity` kwarg here feels like a good candidate for a keyword only argument to me. Specifying a boolean as a positional argument is really meaningless when reading the code?

ayshih: Done.  I've moved the docstring documentation for the keyword out of "Parameters" to a new section "Keyword Arguments" to highlight that it is a keyword-only argument.

Cadair (approved):

Issue #4101, opened 2020-05-06
data manager issue with sunpy.instr.goes


hayesla:
### Description
I know we are planning to move the instr out of core, but I'm playing around with `sunpy.instr. calculate_temperature_em()`, and there seems to be a bug with the data manager to get the required chianti files. 

### Expected behavior
```python
>>> goes_ts = TimeSeries('go1520120514.fits').truncate('2012-05-14 13:32', '2012-05-14 13:45')
>>> ts_new = goes.calculate_temperature_em(goes_ts)   
```
and this should return a lovely timeseries with two extra columns for temperature and emission measure

### Actual behavior
```python
/Users/laurahayes/anaconda3/lib/python3.7/functools.py:824: UserWarning: Discarding nonzero nanoseconds in conversion
  return dispatch(args[0].__class__)(*args, **kw)
Files Downloaded: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00,  8.08file/s]
WARNING: SunpyUserWarning: [Errno 2] No such file or directory: '/Users/laurahayes/sunpy/data_manager/goes_chianti_temp_cor.csv' [sunpy.data.data_manager.cache]
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-9-e8e77b0651c9> in <module>
----> 1 ts_new = goes.calculate_temperature_em(goes_ts)

~/sunpy_dev/sunpy-git/sunpy/instr/goes.py in calculate_temperature_em(goests, abundances, download, download_dir)
    268         satellite=goests.meta.metas[0]["TELESCOP"].split()[1],
    269         date=goests.to_dataframe().index[0],
--> 270         abundances=abundances, download=download, download_dir=download_dir)
    271 
    272     ts_new = timeseries.XRSTimeSeries(meta=copy.deepcopy(goests.meta),

~/anaconda3/lib/python3.7/site-packages/astropy/units/decorators.py in wrapper(*func_args, **func_kwargs)
    232             # Call the original function with any equivalencies in force.
    233             with add_enabled_equivalencies(self.equivalencies):
--> 234                 return_ = wrapped_function(*func_args, **func_kwargs)
    235             if wrapped_signature.return_annotation not in (inspect.Signature.empty, None):
    236                 return return_.to(wrapped_signature.return_annotation)

[...]

PR #4103, opened 2020-05-06
Improve cache downloader error

dstansby:
xref #4101 - I think this is an improvement?

### DIFF
--- a/sunpy/data/data_manager/cache.py
+++ b/sunpy/data/data_manager/cache.py
@@ -167,6 +167,6 @@ class Cache:
                 return download(url)
             except Exception as e:
                 warn(e, SunpyUserWarning)
+                error = e
         else:
-            raise RuntimeError("One or more downloads failed, "
-                               "see above warnings for the errors raised.")
+            raise RuntimeError from error
--- a/sunpy/data/data_manager/cache.py
+++ b/sunpy/data/data_manager/cache.py
@@ -168,4 +168,5 @@ class Cache:
             except Exception as e:
                 warn(e, SunpyUserWarning)
         else:
-            raise RuntimeError("Download failed")
+            raise RuntimeError("One or more downloads failed, "
+                               "see above warnings for the errors raised.")


wtbarnes:
> I think we should ideally do
> 
> ```
> raise RuntimeError from e
> ```

@vn-ki could you add this as a suggestion and we can just approve that?

dstansby:
That wouldn't work though if there are multiple errors, I think the current approach makes sure that all the errors get printed as warnings, and then at the end an error is raised to stop the program.

vn-ki:
I meant something along these lines

```python
In [5]: err = None 
   ...: for i in range(10): 
   ...:     try: 
   ...:         raise Exception(f'{i}') 
   ...:     except Exception as e: 
   ...:         err = e 
   ...:         print(i) 
   ...: else: 
   ...:     raise RuntimeError('asd') from err 
   ...:                                 
0
1
2
3
4
5
6
7
8
9
Exception                                 Traceback (most recent call last)
<ipython-input-5-a8775b499c41> in <module>
      3     try:
----> 4         raise Exception(f'{i}')
      5     except Exception as e:

Exception: 9

The above exception was the direct cause of the following exception:

RuntimeError                              Traceback (most recent call last)
<ipython-input-5-a8775b499c41> in <module>
      7         print(i)
      8 else:
----> 9     raise RuntimeError('asd') from err
     10 

RuntimeError: asd
```

Not the best approach, but if theres only one file (usually the case), we'll get a meaningful stack trace.

wtbarnes:
@vn-ki can you take a second look at that and see if that's what you had in mind?

sunpy/data/data_manager/cache.py
```diff
@@ -167,5 +167,6 @@ def download(url):
                 return download(url)
             except Exception as e:
                 warn(e, SunpyUserWarning)
+                error = e
```
Cadair, line 170: why do we need this line, is `e` really not still accessible after the except block?

wtbarnes: It is not unfortunately. I thought the same thing at first, but my editor complained about it being undefined in the scope of that last `else` block.

nabobalis: Do we not want to append each error to a list and then raise the entire bunch of errors instead of the last error?

wtbarnes: I agree that it is ugly

wtbarnes: 🤷 we could? I don't have a strong opinion on it. They each get raised as warnings along the way anyway.

wtbarnes: The disadvantage would be that you can't raise a `RuntimeError` from a list of exceptions.

sunpy/data/data_manager/cache.py
```diff
@@ -167,5 +167,6 @@ def download(url):
                 return download(url)
             except Exception as e:
                 warn(e, SunpyUserWarning)
```
wtbarnes, line 169: Is there a reason we are using `warnings.warn` and not the logger here?

vn-ki: I don't remember a particular reason. I think I might have copied the pattern from other (old?) parts of the code.

PR #4109, opened 2020-05-07
Remove VSO deprecations

### LINKED ISSUE #4108: Check for deprecated functionality which needs removing before 2.0
Before we finalise 2.0 we should make sure we have removed all the deprecated code

dstansby:
Fixes #4108 (unless we should remove stuff deprecated in 1.1 too?)


Cadair:
> unless we should remove stuff deprecated in 1.1 too?

Nope only stuff deprecated in 1.0

Cadair:
Should have a `removal` type changelog :smile:

PR #4110, opened 2020-05-07
Add a custom User Agent for sunpy parfive downloads

Cadair:
This is something I had completely forgot about, but here we go.

The idea of this is to put the sunpy version in the user agent of all our download requests. This should provide us some version statistics via our friends at the VSO or similar services.

I would like to see this in 2.0 so that we can start collecting useful information as soon as possible.


Cadair:
https://dev.azure.com/sunpy/sunpy/_build/results?buildId=7560&view=logs&j=58e52e63-bf2d-579e-8f74-7ab3fe4e7110&t=09892a10-6d69-5c3e-c7ad-308f1fc1d02b&l=4302

That implies that as I suspected this would bump our min version of parfive to 1.1 as currently implemented. I could do some function inspection magic to not have to do the bump, but maybe given the support for proxy and faster downloading it's a good idea to dump it anyway?

nabobalis:
👍 on bump

sunpy/util/parfive_helpers.py
```diff
@@ -0,0 +1,27 @@
...
+import parfive
+from parfive import Results
+
+import sunpy
+
+__all__ = ['Downloader', 'Results']
+
+
+# Overload the parfive downloader class to set the User-Agent string
+class Downloader(parfive.Downloader):
```
Raahul-Singh, line 14: Why make a full class for this?
Wouldn't simply passing values be good enough?

Cadair: Because it makes it transparent to all the other code in the package that uses it, you just change the import and now it sets the header etc. I did consider defining `User-Agent` and then passing it `Downloader(headers=SUNPY_USER_AGENT)` everywhere but that felt like more effort.

PR #4113, opened 2020-05-07
Use astropy to create `sunpy.sun.constants` docstring RST table

### LINKED ISSUE #4041: Constants __doc__ creation
We currently add a list of all constants into an rst table manually:

https://github.com/sunpy/sunpy/blob/master/sunpy/sun/constants.py#L92
 
This means if a name changes and it becomes wider than the column, it will break the documentation build. We need a way to construct the table that can account for this.

jairideout:
### Description

The RST table in `sunpy.sun.constants` docstring is now built with an astropy `Table` object and the `ascii.rst` writer. The previous implementation manually built an RST table and could break if names changed to become longer than the fixed column widths.

Note: the new implementation uses right-aligned columns, while the previous implementation used centered columns.

Fixes #4041.

Previous docstring:

Screen Shot 2020-05-07 at 9 06 17 AM

New docstring:

Screen Shot 2020-05-07 at 10 54 24 AM

### Questions

1. Is a changelog entry necessary for this PR?
2. Would you like me to add unit test(s) for this functionality? I checked that the docs build and render successfully.

### DIFF
--- a/sunpy/sun/constants.py
+++ b/sunpy/sun/constants.py
@@ -1,6 +1,8 @@
 """
 This module provides fundamental solar physical constants.
 """
+import io
+
 from astropy.table import Table
 from astropy.time import Time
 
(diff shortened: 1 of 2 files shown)


nabobalis:
> ### Questions
> 1. Is a changelog entry necessary for this PR?

No, I don't feel so.

> 2. Would you like me to add unit test(s) for this functionality? I checked that the docs build and render successfully.

One unit test can't hurt.

jairideout:
Thanks for reviewing, @nabobalis! I added a unit test. Please let me know if you'd like a different approach to test the generated docstring.

jairideout:
I think the test failures are unrelated (the errors are coming from `sunpy/visualization/animator/tests/test_wcs.py`).

nabobalis:
Thanks @jairideout

PR #4116, opened 2020-05-08
Updated STEREO SECCHI starfield example to use all of the information in the star catalog

ayshih:
In the STEREO SECCHI starfield example, we weren't using all of the information in the star catalog.  We were using a false distance even though it can be determined from the measured parallax, and we weren't updating the position based on the time between the reference epoch for the catalog and the actual observation.  These improvements do not produce any visually discernable change in the final plot, but we still should do everything the proper way.

That said, ERFA now raises a warning when it updates the star positions because it doesn't like a proper motion in the catalog.  Oh well.

### DIFF
--- a/changelog/4116.doc.rst
+++ b/changelog/4116.doc.rst
@@ -1,0 +1,1 @@
+Updated the gallery example :ref:`sphx_glr_generated_gallery_units_and_coordinates_stereo_secchi_starfield.py` to use all of the information in the star catalog.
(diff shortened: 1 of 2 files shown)


dstansby:
I think this could do with a changelog since it's slightly changing the method used in the example - otherwise looks 👍  to me

ayshih:
Changelog added, and I now also read the reference epoch from the star catalog instead of hard-coding it.

ayshih:
I imagine the backport failed because #4039 wasn't backported...

PR #4118, opened 2020-05-08
Improve fits header helper doc

dstansby:
It wasn't obvious to me that `coordinates` was the coord of the reference pixel. Hopefully this is clearer.


sunpy/map/header_helper.py
```diff
@@ -53,7 +53,7 @@ def make_fitswcs_header(data, coordinate,
         Array data of Map for which a header is required, or the shape of the
         data array (in numpy order, i.e. ``(y_size, x_size)``).
     coordinates : `~astropy.coordinates.SkyCoord` or `~astropy.coordinates.BaseFrame`
```
Cadair, line 55:
```suggestion
    coordinate : `~astropy.coordinates.SkyCoord` or `~astropy.coordinates.BaseFrame`
```

Cadair:
now was this in 1.0 or 1.1? :grin:

PR #4121, opened 2020-05-08
Fix map.{top_right_coord, center}

### LINKED ISSUE #3793: top_right_coord may be off by one
The `top_right_coord` attribute on `sunpy.map.GenericMap` is calculated as,

https://github.com/sunpy/sunpy/blob/2a3d9a5cb34ba496fcf55d861e9307e3cec783db/sunpy/map/mapbase.py#L521

While WCS indexing does begin at 1, in astropy.wcs.WCS, our origin is defined such that `(0,0)*u.pixel` corresponds to the center of the bottom left corner. This would mean that, for an `N`-by-`N` image, the top right corner should correspond to `(N-1,N-1)*u.pixel`.

Further testing is needed to determine whether this is actually the case.

dstansby:
Fixes #3793.

It turns out `map.center` was off-by-one too, so I have fixed that and clarified its behaviour in the docstring. For reference the changes to the test images can be found here: https://49358-2165383-gh.circle-artifacts.com/0/.tmp/figure_py36/figure_test_images/fig_comparison.html


sunpy/map/mapbase.py
```diff
@@ -672,7 +672,8 @@ def top_right_coord(self):
         """
         The physical coordinate at the center of the the top left ([-1, -1]) pixel.
         """
-        return self.pixel_to_world(*self.dimensions)
+        return self.pixel_to_world(self.dimensions[0] - 1 * u.pix,
+                                   self.dimensions[1] - 1 * u.pix)
```
Cadair, line 676: I *think* an easier fix might be
```suggestion
        return self.pixel_to_world(*self.dimensions, origin=1)
```

ayshih:
Question regarding this off by one stuff:  If we do a `submap()` with `GenericMap.bottom_left_coord` and `GenericMap.top_right_coord`, do we get back the whole map?

dstansby:
> Question regarding this off-by-one stuff: If we do a `submap()` with `GenericMap.bottom_left_coord` and `GenericMap.top_right_coord`, do we get back the whole map?

A good question - the exact behaviour of `submap()` is very undocumented, so I think I will open other issues/PRs to check this and improve the documentation.

dstansby:
Eek, this changes some of the superpixel plot axes quite a bit!

sunpy/map/mapbase.py
```diff
@@ -672,14 +672,18 @@ def top_right_coord(self):
         """
         The physical coordinate at the center of the the top left ([-1, -1]) pixel.
```
ayshih, line 673:
```suggestion
        The physical coordinate at the center of the the top right ([-1, -1]) pixel.
```

Cadair:
I am not surprised by the changes to the nonwcsaxes plots, they use centre iirc. The composite map ones are a little surprising, but I guess they might not use wcsaxes either.

changelog/4121.breaking.rst
```diff
@@ -0,0 +1,3 @@
+`sunpy.map.GenericMap.top_right_coord` and `~sunpy.map.map.GenericMap.center`
```
Cadair, line 1: I think "bugfix" might be more appropriate for this changelog type, according to the changelog readme, a breaking change is:

> A change which requires users to change code and is not backwards compatible.

Which I don't think this really is, it's just a particularly bad bug fix lol

sunpy/map/tests/test_mapbase.py
```diff
@@ -66,6 +66,19 @@ def aia171_test_map_with_mask(aia171_test_map):
     return sunpy.map.Map(np.ma.array(aia171_test_map.data, mask=mask), aia171_test_map.meta)
 
 
+@pytest.fixture
+def simple_map():
+    # A 2x2 map, with it's center at (0, 0), and scaled differently in
+    # each direction
+    data = np.arange(4).reshape((2, 2))
+    ref_coord = SkyCoord(0, 0, frame='helioprojective', obstime='now', unit='deg')
+    ref_pix = [1.5, 1.5] * u.pix
```
wtbarnes, line 75: Isn't this inconsistent with #4154? If the reference pixel is now zero-based, shouldn't this be (0.5,0.5)? Or is the input to `make_fitswcs_header` expected to be one-based?

Cadair: yeah this now needs updating post-#4154

wtbarnes: After drawing pictures for 10 minutes, really had a crisis and thought I understood nothing until I realized we switched conventions 😆

PR #4124, opened 2020-05-08
Added a custom directive for Sphinx to include mini-galleries

ayshih:
This PR adds a custom directive for Sphinx that makes it really easy to add a mini-gallery where desired to the documentation.  For example, for the `RotatedSunFrame` mini-gallery, this is all you need to write:
```rst
.. minigallery:: coordinates.RotatedSunFrame
```
If you want the mini-gallery to cover multiple objects, just separate them by spaces, e.g.:
```rst
.. minigallery:: coordinates.sun.earth_distance coordinates.transform_with_sun_center
```

Edit: For reference, here's the [`sphinx-gallery` documentation on mini-galleries](https://sphinx-gallery.github.io/stable/configuration.html#add-mini-galleries-for-api-documentation).


docs/conf.py
```diff
@@ -318,6 +317,44 @@ def run(self):
             sys.stdout = old_stdout
 
 
+class MiniGallery(Directive):
+    """
+    Custom directive to insert a mini-gallery
+
+    The required argument is the name of the object, excluding the initial `sunpy.`.  There can be
```
Cadair, line 324: I get why you have dropped the initial `sunpy.` but I think it would be better if we kept it. For a start affiliated packages could import this from core and use it in their docs which would obviously not work so good :D

Cadair:
This is awesome :grinning:

nabobalis:
Just a changelog.

mbobra:
Wow this is really really awesome. Have you considered a PR to sphinx?

docs/conf.py
ayshih: Okay, changed

ayshih:
> Wow this is really really awesome. Have you considered a PR to sphinx?

(To be clear, any PR would be to `sphinx-gallery`.)  Considered?  Sure.  Would it be anywhere on the top few pages of my to-do list?  Probably not.

dstansby:
I'm not going to block this but I strongly think this should go upstream instead of in SunPy, so that the wider sphinx/python ecosystem can benefit from it, and so it gets reviewers who are more experienced and knowledgeable about sphinx/sphinx-gallery than sunpy maintainers are. If we really want it quickly we can always vendor what gets merged until a new sphinx-gallery release is made.

ayshih:
@dstansby I just submitted an issue to `sphinx-gallery` to see if they'd be interested in this directive.  I'll see what they say, but judging from some of their discussion threads, they appear to have an guideline to not provide any directives at all, with one reason being that they can be prone to breakage when Sphinx is updated.

PR #4129, opened 2020-05-08
Maintain coherence between keys and keycomments in MetaDict

### LINKED ISSUE #2750: Maintain coherence between keycomments and the metadict
See #2748 

This is probably best implemented by adding the functionality to our `MetaDict` object or something, so that we don't have to do it manually everywhere.

jairideout:
### Description

`sunpy.util.metadata.MetaDict` now maintains coherence between its keys and their corresponding keycomments.

If the key ``'keycomments'`` exists, its value must be a dictionary mapping keys in the `MetaDict` to their comments. The casing of keys in the keycomments dictionary is not significant. If a key is removed from the `MetaDict`, it will also be removed from the keycomments dictionary. Additionally, any extraneous keycomments will be removed when the `MetaDict` is instantiated.

Calling `del` on a `MetaDict` object key is now case-insensitive.

Fixes #2750

### DIFF
--- a/sunpy/util/metadata.py
+++ b/sunpy/util/metadata.py
@@ -1,6 +1,6 @@
 """
 This module provides a generalized dictionary class that deals with header
-parsing and normalization.
+parsing, normalization, and maintaining coherence between keys and keycomments.
 """
 from collections import OrderedDict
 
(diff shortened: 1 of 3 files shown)


sunpy/util/metadata.py
```diff
@@ -14,27 +14,66 @@ class MetaDict(OrderedDict):
...
         """
-        # Store all keys as upper-case to allow for case-insensitive indexing
+        # Store all keys as lower-case to allow for case-insensitive indexing
         # OrderedDict can be instantiated from a list of lists or a tuple of tuples
         tags = dict()
         if args:
             args = list(args)
             adict = args[0]
             if isinstance(adict, list) or isinstance(adict, tuple):
-                tags = OrderedDict((k.upper(), v) for k, v in adict)
```
jairideout, line 29: These `upper` calls aren't necessary because the keys are converted to lowercase in the `super()` call below.

jairideout:
I think the test failures are unrelated.

dstansby:
Thanks for the PR! We're all a bit busy trying to get version 2.0 finished, but I'll try and get round to looking at this soon.

dstansby:
I'm just going to close and re-open to cycle the CI, since the failing devdeps build has disappeared.

nabobalis:
I had to rebase this on master due to the shear number of commits that have changed.

nabobalis:
Failures are known and I will merge this.

Thanks again @jairideout.

PR #4132, opened 2020-05-10
href fix in scraper

abhijeetmanhas:
### Description

More info here https://github.com/sunpy/sunpy/pull/3812#discussion_r422460309
In short, we didn't checked href can be None and may give error here https://github.com/sunpy/sunpy/blob/553d60538ecb6fb955fc28d428372c17f0501557/sunpy/util/scraper.py#L234 
By working on #3812 , I found the source website do have some href which are **None**.
Also, sometimes path to files maybe relative to just domain, not always to directory they are in, so checking that by verifying if href for files begin with `\` .


nabobalis:
Can I get a changelog please?

abhijeetmanhas:
Sure

abhijeetmanhas:
https://github.com/dpshelio/sunpy/blob/5eba84ce1ab16b64852bc6441a2027823e6fd5ab/sunpy/util/scraper.py#L234
This fix was also done in old groundclients PR.

PR #4134, opened 2020-05-10
 Fix/overhaul/test submap 

### LINKED ISSUE #4128: Fix difference in pixel and coordinate input to submap()
Currently it seems that submap has slightly different of handling any pixels that overlap with the boundary given. Reading the code the current behaviour is:

- If you give pixel input, these are rounded via. a call to `int()` (so to the nearest integer)
- If you give coordinate input, these are converted to pixels and both are rounded up.

The difference in behaviour should be fixed, but in which direction? I am inclined to say we should just be rounding to the nearest integer, which is currently done for pixel input.

dstansby:
Previously the behaviour of `submap()` wasn't very well documented, so I decided to work out what it actually does. Along the way I discovered a lack of tests, some *interesting* code, and a lack of consistency between what happened with pixel and coordinate inputs.

This PR somewhat overhauls `submap` to
- Have well defined behaviour for pixels on the boundary of the rectangle, that is consistent for all boundaries (previously pixels on the lower and left boundary were included, but on the upper and right boundary excluded). This means the shape of a submap is now often 1 pixel larger in each dimension.
- Extend the tests
- Refactor the internals to remove code duplication

Fixes #4128, and is built on top of #4127

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1488,12 +1488,14 @@ class GenericMap(NDData):
         x_pixels.sort()
         y_pixels.sort()
 
-        # Round down the lower left pixel to the nearest integer
-        x_pixels[0] = np.round(x_pixels[0])
-        y_pixels[0] = np.round(y_pixels[0])
+        # Round the lower left pixel to the nearest integer
+        # We want 0.5 to be rounded up to 1, so use floor(x + 0.5)
+        x_pixels[0] = np.floor(x_pixels[0] + 0.5)
+        y_pixels[0] = np.floor(y_pixels[0] + 0.5)
         # Round the top right pixel to the nearest integer, then add 1 for array indexing
-        x_pixels[1] = np.round(x_pixels[1]) + 1
-        y_pixels[1] = np.round(y_pixels[1]) + 1
+        # We want e.g. 2.5 to be rounded down to 2, so use ceil(x - 0.5)
+        x_pixels[1] = np.ceil(x_pixels[1] - 0.5) + 1
+        y_pixels[1] = np.ceil(y_pixels[1] - 0.5) + 1
 
         x_pixels = np.array(x_pixels)
         y_pixels = np.array(y_pixels)
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1492,8 +1492,8 @@ class GenericMap(NDData):
         x_pixels[0] = np.round(x_pixels[0])
         y_pixels[0] = np.round(y_pixels[0])
         # Round the top right pixel to the nearest integer, then add 1 for array indexing
-        x_pixels[1] = round(x_pixels[1]) + 1
-        y_pixels[1] = round(y_pixels[1]) + 1
+        x_pixels[1] = np.round(x_pixels[1]) + 1
+        y_pixels[1] = np.round(y_pixels[1]) + 1
 
         x_pixels = np.array(x_pixels)
         y_pixels = np.array(y_pixels)
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1379,6 +1379,9 @@ class GenericMap(NDData):
         Returns a submap of the map defined by the rectangle given by the
         ``[bottom_left, top_right]`` coordinates.
 
+        Any pixels which have at least part of their area inside the rectangle
+        are returned.
+
         Parameters
         ----------
         bottom_left : `astropy.units.Quantity` or `~astropy.coordinates.SkyCoord`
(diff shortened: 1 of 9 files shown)


sunpy/map/mapbase.py
```diff
@@ -1369,6 +1369,9 @@ def submap(self, bottom_left, top_right=None):
         Returns a submap of the map defined by the rectangle given by the
         ``[bottom_left, top_right]`` coordinates.
 
+        Any pixels which have at least part of their area inside the rectangle
+        are returned.
```
Cadair, line 1373: Was this the behaviour before when using world coordinates?

sunpy/map/mapbase.py
```diff
@@ -1452,58 +1455,43 @@ def submap(self, bottom_left, top_right=None):
...
         # Sort the pixel values so we always slice in the correct direction
         x_pixels.sort()
         y_pixels.sort()
 
+        # Round down the lower left pixel to the nearest integer
+        x_pixels[0] = np.round(x_pixels[0])
+        y_pixels[0] = np.round(y_pixels[0])
+        # Round the top right pixel to the nearest integer, then add 1 for array indexing
+        x_pixels[1] = np.round(x_pixels[1]) + 1
+        y_pixels[1] = np.round(y_pixels[1]) + 1
```
Cadair, line 1486: Following on from mycomment in the docstring about including a pixel with any part, should this not always round down? Or am I mis-interpreting where we are indexing pixels from, i.e. are they referenced from the centre (FITS WCS) or bottom left (array indexing)?

If your "top_right" pixel coord is `(5.8, 5.8)`, it rounds up to 6 and then gets one added and ends up at 7, which is one pixel too large?

Cadair:
Thanks a lot for doing this, I don't think the core logic of submap has been looked at since we migrated it to Quantity many many moons ago.

sunpy/map/mapbase.py
dstansby: This rounds up because when you're slicing an array, the "stop" part of the slice has to be one larger than the pixel number you want (e.g. to get pixel number 3 you would slice [3:4].

sunpy/map/mapbase.py
dstansby: I think so

sunpy/map/mapbase.py
Cadair: I understand that as to why the +1 is there, but what I am trying to understand is the combo of the round up and the +1. 

So am I right in saying that if you have a 3x3 array and indexing from `(0, 0)` left you submap from `(0, 0)` to `(1.8, 1.8)` what's the output we expect? I think it should be a 2x2 array of the first 4 pixels right? i.e. the same as `[0:2]`

Well if you round up 1.8 and then add 1 you are doing `[0:3]` which gives you a 3x3 array out:
```python
>>> a                                                                                                                                                                                                                                                          
array([[4, 6, 2],
       [0, 4, 7],
       [1, 8, 8]])
>>> a[0:int(np.round(1.8) + 1), 0:int(np.round(1.8) + 1)]                                                                                                                                                                                                          
array([[4, 6, 2],
       [0, 4, 7],
       [1, 8, 8]])
```

If we are defining the centre of the pixels to be the index point (so our 3x3 array goes from `(-0.5, -0.5) -> (2.5, 2.5)`) then what we have is closer to correct, but I think we have issues with round down then (I would need to brush up on my WCS indexing).

---

(This is why we should use get item syntax for submap operations in pixel coordinates...)

dstansby: The 1.8 is rounded up because it has gone over the pixel border between pixels 1 and 2, that lies at a value of 1.5. Perhaps a picture helps:

Screenshot 2020-05-11 at 10 21 26

Cadair: I wonder if using an implementation like this: https://github.com/spacetelescope/gwcs/blob/master/gwcs/utils.py#L52 which has been tested in a WCS framework would be a good idea?

I will think about this a little more, now I know we are in ints at centre of pixel mode.

dstansby: `floor(x + 0.5)` is the same as `round(x)`, so I the gwcs stuff is functionally doing the same. Edge cases (ie. 0.5) might be different though.

ayshih: > Edge cases (ie. 0.5) might be different though.

Yeah, definitely don't use `np.round()` for this code because it follows the convention of rounding to the closest even number.

dstansby: Woops, thanks for spotting that. As a note to myself, we want to make sure that for an edge (ie. 0.5) we include the pixel inside the rectangle but exclude the pixel outside.

sunpy/map/tests/test_mapbase.py
```diff
@@ -416,9 +429,39 @@ def test_shift_history(generic_map):
     assert resultant_shift[1] == y_shift1 + y_shift2
 
 
+@pytest.mark.parametrize('rect, submap_out',
```
wtbarnes, line 432: I think we might want to expand this test to a 3x3 array. I think this is what the discussion above was getting at, but `np.round(0.5)=0` while `np.round(1.5)=2`, meaning `[(0,0),(1.5,1.5)]` would actually return the whole 3-by-3 array when it should return only `[0:2,0:2]`. So as it stands, this test would fail if extended beyond a 2x2. We should make sure we're testing a general enough case.

sunpy/map/mapbase.py
dstansby: I've updated the rounding so that it correctly takes into account boundary (ie. n + 0.5) values.

sunpy/map/tests/test_mapbase.py
dstansby: I've updated the rounding to get rid of `np.round` completely - I can still add a 3x3 test map?

wtbarnes: Yeah, I think it would still be a good idea to have a pixel center with a coordinate >1. Presumably this should just be changing the `simple_map` fixture to `np.arange(9).reshape(3,3)` and then adding one line to the test parametrization here?

wtbarnes:
Still needs a changelog and for the `test_submap` test to be expanded a bit. I'll have this done by the end of the day.

dstansby:
Ah great thanks, I didn't have any time to update this today.

wtbarnes:
So I've found a fun little problem while extending the tests to a 3-by-3 by map.

If you specify a sub-map in pixel coordinates with the corners (0,0) and (0.5,0.5), this should select only the `map.data[0,0]` pixel as the edge of the first pixel in both directions is 0.5. Thus, this selection covers no other pixels except `map.data[0,0]`. When the cutout selection is specified in pixel coordinates, this works as expected. However, when these pixel coordinates are first converted to world coordinates and then passed into `submap` (i.e. the pixel coordinates are round-tripped through `pixel_to_world` -> `world_to_pixel`), in the 3-by-3 case, there is enough round-off error introduced that the resulting y-pixel coordinate is now greater than 0.5 (by an amount O(1e-12)). 

This means that when converting the pixel coordinates to array indices, the selection becomes [0:2,0:1] and both the first and second pixels in the y-direction are selected. For some reason, this doesn't pop up in the 2-by-2 case, but seems to persist for dimensions > 3-by-3.

wtbarnes:
I've separated out the tests into separate ones for the world and pixel submaps. In the world case, I've not included the test for the case of being right on the boundary (i.e. 0.5,0.5) as we cannot reliably test this.

sunpy/map/tests/test_mapbase.py
```diff
@@ -505,17 +517,17 @@ def test_submap_data_header(generic_map, unit):
...
-    assert simple_map.reference_pixel.x == 0.5 * u.pix
-    assert simple_map.reference_pixel.y == 0.5 * u.pix
+    assert simple_map.reference_pixel.x == 1 * u.pix
+    assert simple_map.reference_pixel.y == 1 * u.pix
 
 
 def test_resample(simple_map):
     # Test resampling a 2x2 map
     resampled = simple_map.resample([1, 1] * u.pix)
     # Should be the mean of [0, 1, 2, 3]
```
dstansby, line 527:
```suggestion
    # Should be the mean of [0,1,2,3,4,5,6,7,8,9]
```

Issue #4091, opened 2020-05-04
Log something to stdout when sample data download fails


Cadair:
If we fall over to our mirrors for the sample data download we should log something. Mainly so that we can easily see it happening on the CI etc

rutujasurve94:
Hi, I've raised PR: https://github.com/sunpy/sunpy/pull/4137 for this issue.

nabobalis:
Fixed by https://github.com/sunpy/sunpy/pull/4137

PR #4137, opened 2020-05-10
Issue #4091: Logging for download sample data failure

rutujasurve94:
Adding info and debug log for failure on downloading of sample data - first download and subsequent retries.

### DIFF
--- a/sunpy/data/_sample.py
+++ b/sunpy/data/_sample.py
@@ -3,6 +3,7 @@ import warnings
 from pathlib import Path
 from collections import namedtuple
 from urllib.parse import urljoin
+from sunpy import log
 
 from sunpy.util.config import get_and_create_sample_dir, _is_writable_dir
 from sunpy.util.exceptions import SunpyUserWarning
@@ -102,6 +103,8 @@ def download_sample_data(overwrite=False):
 
     if not results.errors:
         return results
+    else:
+        log.info('Failed to download sample data, retrying with a mirror.')
 
     for retry_url in _base_urls[1:]:
         for i, err in enumerate(results.errors):
@@ -119,6 +122,8 @@ def download_sample_data(overwrite=False):
 
     for err in results.errors:
         file_name = Path(err.url).name
+        log.info('Failed to download file' + str(err.filepath_partial) +
+                 'from url: ' + str(err.url) + ' with error: ' + str(err.exception))
         warnings.warn(f"File {file_name} not found.", SunpyUserWarning)
 
     return results
--- a/changelog/4137.trivial.rst
+++ b/changelog/4137.trivial.rst
@@ -1,0 +1,1 @@
+Added log messages when the sample data fails to download.


changelog/4137.featuredownloadsamplelog.rst
```diff
@@ -0,0 +1 @@
+Added logs for failure while downloading sample data.
```
nabobalis, line 1:
```suggestion
Added log messages when the sample data fails to download.
```
This file should also be named `4137.trival.rst`

sunpy/data/_sample.py
```diff
@@ -116,6 +119,9 @@ def download_sample_data(overwrite=False):
 
         if not results.errors:
             return results
+        else:
+            log.debug('Failed to download file' + str(err.filepath_partial) +
```
nabobalis, line 123: Should this be info to match the one above or should log above be debug?

rutujasurve94: Since we wish to log to std output, we could keep it 'info' for consistency. Though this log will be useful for debugging purposes as well.

changelog/4137.featuredownloadsamplelog.rst
rutujasurve94: Done.

sunpy/data/_sample.py
```diff
@@ -116,6 +119,9 @@ def download_sample_data(overwrite=False):
 
         if not results.errors:
             return results
+        else:
+            log.info('Failed to download file' + str(err.filepath_partial) +
+                      'from url: ' + str(new_url) + ' with error: ' + str(err.exception))
```
nabobalis, line 124: I wonder if this log info should be in the for loop below?

rutujasurve94: Yes, that makes sense, added it below.

sunpy/data/_sample.py
```diff
@@ -119,6 +122,8 @@ def download_sample_data(overwrite=False):
 
     for err in results.errors:
         file_name = Path(err.url).name
+        log.info('Failed to download file' + str(err.filepath_partial) +
+                 'from url: ' + str(new_url) + ' with error: ' + str(err.exception))
```
nabobalis, line 126:
```suggestion
                 'from url: ' + str(err.url) + ' with error: ' + str(err.exception))
```

sunpy/data/_sample.py
```diff
@@ -102,6 +103,8 @@ def download_sample_data(overwrite=False):
 
     if not results.errors:
         return results
+    else:
+        log.info('Failed to download sample data, retrying')
```
nabobalis, line 107:
```suggestion
        log.info('Failed to download sample data, retrying with a mirror.')
```

sunpy/data/_sample.py
rutujasurve94: Done

sunpy/data/_sample.py
rutujasurve94: Done

sunpy/data/_sample.py
```diff
@@ -102,14 +103,16 @@ def download_sample_data(overwrite=False):
...
+        log.info('Failed to download sample data, retrying with a mirror.')
 
     for retry_url in _base_urls[1:]:
         for i, err in enumerate(results.errors):
             file_name = Path(err.url).name
             # Overwrite the parfive error to change the url to a mirror
             new_url = urljoin(retry_url, file_name)
             results._errors[i] = _error(err.filepath_partial,
-                                        new_url,
+                                        err.url,
```
nabobalis, line 115: This should be undone.

sunpy/data/_sample.py
nabobalis: You only need to change this line.

sunpy/data/_sample.py
rutujasurve94: done.

nabobalis:
Thanks again @rutujasurve94

ayshih:
Well, this already got merged, but why are these not `log.warn()` and `log.error()`?  `log.info()` doesn't seem like the appropriate level for these messages.

nabobalis:
Failing to download the sample data is an error but isn't that important if it fail so I was ok with info.

It could be raised to a higher level.

rutujasurve94:
I've raised a PR for adding warnings and error in case info needs to be removed here:
https://github.com/sunpy/sunpy/pull/4149

PR #4152, opened 2020-05-11
Fix reference pixel in header helper

### LINKED ISSUE #4120: make_fitswcs_header indexes from 1, not 0 as claimed in docs
Using the following toy example, I would expect the reference coordinate (0, 0) to correspond to the center of the [1, 1] pixel in the array, ie. the top right pixel. Looking at the figure below it seems like it refers to the [0, 0] pixel, which would mean `make_fitswcs_header` is indexing from 1 instead of 0.

The docstring [claims to be indexing from 0.](https://docs.sunpy.org/en/stable/api/sunpy.map.make_fitswcs_header.html#sunpy.map.make_fitswcs_header)

```python
import numpy as np
import sunpy.map
import sunpy.coordinates
import astropy.units as u

data = np.arange(4).reshape((2, 2))
ref_coord = SkyCoord(0, 0, frame='helioprojective', obstime='now', unit='deg')
ref_pix = [1, 1] * u.pix
scale = [2, 2] * u.arcsec / u.pix
print(header)

m = sunpy.map.Map(data, header)
m.peek()
```

Figure_1

dstansby:
Fixes #4120. Indexing from 0 as input is consistent with the current docstring, and other pixel inputs in the codebase.

### DIFF
--- a/sunpy/map/header_helper.py
+++ b/sunpy/map/header_helper.py
@@ -133,15 +133,16 @@ def make_fitswcs_header(data, coordinate,
     meta_wcs.update(meta_instrument)
 
     if reference_pixel is None:
-        reference_pixel = u.Quantity([(shape[1] + 1)/2.*u.pixel, (shape[0] + 1)/2.*u.pixel])
+        reference_pixel = u.Quantity([(shape[1] - 1)/2.*u.pixel, (shape[0] - 1)/2.*u.pixel])
     if scale is None:
         scale = [1., 1.] * (u.arcsec/u.pixel)
 
     meta_wcs['crval1'], meta_wcs['crval2'] = (coordinate.spherical.lon.to_value(meta_wcs['cunit1']),
                                               coordinate.spherical.lat.to_value(meta_wcs['cunit2']))
 
-    meta_wcs['crpix1'], meta_wcs['crpix2'] = (reference_pixel[0].to_value(u.pixel),
-                                              reference_pixel[1].to_value(u.pixel))
+    # Add 1 to go from input 0-based indexing to FITS 1-based indexing
+    meta_wcs['crpix1'], meta_wcs['crpix2'] = (reference_pixel[0].to_value(u.pixel) + 1,
+                                              reference_pixel[1].to_value(u.pixel) + 1)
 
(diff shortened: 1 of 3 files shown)


sunpy/map/header_helper.py
```diff
@@ -133,15 +133,16 @@ def make_fitswcs_header(data, coordinate,
...
     if scale is None:
         scale = [1., 1.] * (u.arcsec/u.pixel)
 
     meta_wcs['crval1'], meta_wcs['crval2'] = (coordinate.spherical.lon.to_value(meta_wcs['cunit1']),
                                               coordinate.spherical.lat.to_value(meta_wcs['cunit2']))
 
-    meta_wcs['crpix1'], meta_wcs['crpix2'] = (reference_pixel[0].to_value(u.pixel),
-                                              reference_pixel[1].to_value(u.pixel))
+    # Add 1 to go from input 0-based indexing to FITS 1-based indexing
+    meta_wcs['crpix1'], meta_wcs['crpix2'] = (reference_pixel[0].to_value(u.pixel) + 1,
```
hayesla, line 144: but what if the reference pixel given is already in FITS based indexing - i.e. you pull it from the header of another map 🤔 

but maybe this is a silly question ...

hayesla:
good catch @dstansby  - this is really an issue with radio data when the scale is several arcsecs :S

sunpy/map/header_helper.py
hayesla: We will probably have to be more explicit in docs

Cadair: The docs already explicitly say that the reference pixel is zero indexed. Which was a lie :see_no_evil: 

> this argument is zero indexed (Python convention) not 1 indexed (FITS convention)

wtbarnes:
Wasn’t there an off by one error fixed in this function previously? Is this reverting that fix?

Cadair:
> Is this reverting that fix?

I *think* that was fixing the default value (i.e. centre of given array) and not the values passed with the kwarg. (but I might be wrong)

sunpy/map/header_helper.py
dstansby: Ideally one could do `m.reference_pixel` to get a zero index reference pixel, but that just returns the one-based index pixels at the moment... if we're going to make everything zero indexed then I guess we should change the `reference_pixel` property too?

PR #4164, opened 2020-05-14
Fixes for linear upsampling

dstansby:
- The code for calculating new/old interpolating coordinates baffled me, so I re-wrote it to make it clearer what is going on (especially with respect to applying offsets)
- The interpolation was using a fill value of `min(old_coords[-1])` to fill in any data around the edges. This is crazy because `old_coords` are coordinates and not data, and is just arbitrarily filling with a value. I've changed this to extrapolate beyond the data edges.

These changes now mean that `np.mean(resampled.data) == np.mean(simple_map.data)` for upsampling in a linear fashion, which seems like a sensible result to me.

### DIFF
--- a/changelog/4164.bugfix.rst
+++ b/changelog/4164.bugfix.rst
@@ -1,0 +1,4 @@
+Previously `sunpy.map.GenericMap.resample` with ``method='linear'`` was
+using an incorrect and constant value to fill edges when upsampling a map. Values
+near the edges are now correct extrapolated using the ``fill_value=extrapolate``
+option to `scipy.interpolate.interp1d`.
--- a/changelog/4164.bugfix.rst
+++ b/changelog/4164.bugfix.rst
@@ -1,4 +1,4 @@
 Previously `sunpy.map.GenericMap.resample` with ``method='linear'`` was
 using an incorrect and constant value to fill edges when upsampling a map. Values
-near the edges are now correct extrapolated using the ``fill_value=extrapolate``
+near the edges are now correctly extrapolated using the ``fill_value=extrapolate``
 option to `scipy.interpolate.interp1d`.
--- a/sunpy/image/resample.py
+++ b/sunpy/image/resample.py
@@ -85,22 +85,21 @@ def _resample_nearest_linear(orig, dimensions, method, offset, m1):
 
     Parameters
     ----------
-    orig :
+    orig : array-like
         Original data.
-    dimensions :
-        Dimensions of resamled data.
-    offset :
+    dimensions : `tuple`
+        Dimensions of resampled data.
+    method : `str`
+        Interpolation method passed to `~scipy.interpolate.interp1d`
+    offset : `float`
         Either 0 or 0.5, depending on whether interpolation is at the edge or
         centers of bins.
-    m1 :
+    m1 : 0 or 1
         For ``orig.shape = (i,j)`` & new dimensions ``= (x,y)``, if set to `False`
         (default) ``orig`` is resampled by factors of ``(i/x) * (j/y)``,
         otherwise ``orig`` is resampled by ``(i-1)/(x-1) * (j-1)/(y-1)``.
         This prevents extrapolation one element beyond bounds of input array.
     """
-    dimlist = []
-
-    # Calculate old coords. Add 0.5 to get centers of pixels
     old_coords = [np.arange(i, dtype=np.float) + offset for i in orig.shape]
     scale = (orig.shape - m1) / (dimensions - m1)
     new_coords = [(np.arange(dimensions[i], dtype=np.float) + offset) * scale[i] for i in
--- a/sunpy/image/resample.py
+++ b/sunpy/image/resample.py
@@ -82,31 +82,43 @@ def resample(orig, dimensions, method='linear', center=False, minusone=False):
 def _resample_nearest_linear(orig, dimensions, method, offset, m1):
     """
     Resample Map using either linear or nearest interpolation.
+
+    Parameters
+    ----------
+    orig :
+        Original data.
+    dimensions :
+        Dimensions of resamled data.
+    offset :
+        Either 0 or 0.5, depending on whether interpolation is at the edge or
+        centers of bins.
+    m1 :
+        For ``orig.shape = (i,j)`` & new dimensions ``= (x,y)``, if set to `False`
+        (default) ``orig`` is resampled by factors of ``(i/x) * (j/y)``,
+        otherwise ``orig`` is resampled by ``(i-1)/(x-1) * (j-1)/(y-1)``.
+        This prevents extrapolation one element beyond bounds of input array.
     """
     dimlist = []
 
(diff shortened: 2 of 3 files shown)


sunpy/image/resample.py
```diff
@@ -82,31 +82,44 @@ def resample(orig, dimensions, method='linear', center=False, minusone=False):
...
 
-    # calculate new dims
-    for i in range(orig.ndim):
-        base = np.arange(dimensions[i])
-        dimlist.append((orig.shape[i] - m1) / (dimensions[i] - m1) *
-                       (base + offset) - offset)
-
-    # specify old coordinates
-    old_coords = [np.arange(i, dtype=np.float) for i in orig.shape]
+    # Calculate old coords. Add 0.5 to get centers of pixels
```
Cadair, line 103: I am a little confused by this comment. You are adding `offset` right, which could be 0.5 or not?

Cadair:
(This should get a second review from someone who knows a little more about resampling)

changelog/4164.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+Previously `sunpy.map.GenericMap.resample` with ``method='linear'`` was
+using an incorrect and constant value to fill edges when upsampling a map. Values
+near the edges are now correct extrapolated using the ``fill_value=extrapolate``
```
wtbarnes, line 3:
```suggestion
near the edges are now correctly extrapolated using the ``fill_value=extrapolate``
```

sunpy/image/resample.py
```diff
@@ -82,31 +82,44 @@ def resample(orig, dimensions, method='linear', center=False, minusone=False):
 def _resample_nearest_linear(orig, dimensions, method, offset, m1):
     """
     Resample Map using either linear or nearest interpolation.
+
+    Parameters
+    ----------
+    orig :
+        Original data.
+    dimensions :
+        Dimensions of resamled data.
```
wtbarnes, line 91:
```suggestion
        Dimensions of resampled data.
```

sunpy/image/resample.py
```diff
@@ -82,31 +82,44 @@ def resample(orig, dimensions, method='linear', center=False, minusone=False):
...
+    offset :
```
wtbarnes, line 92:
```suggestion
    method : `str`
        Keyword specifying interpolation method passed to `~scipy.interpolate.interp1d`
    offset : `float`
```

sunpy/image/resample.py
```diff
@@ -82,31 +82,44 @@ def resample(orig, dimensions, method='linear', center=False, minusone=False):
...
+        Either 0 or 0.5, depending on whether interpolation is at the edge or
+        centers of bins.
+    m1 :
```
wtbarnes, line 95:
```suggestion
    m1 : `bool`
```

sunpy/image/resample.py
```diff
@@ -82,31 +82,44 @@ def resample(orig, dimensions, method='linear', center=False, minusone=False):
 def _resample_nearest_linear(orig, dimensions, method, offset, m1):
     """
     Resample Map using either linear or nearest interpolation.
+
+    Parameters
+    ----------
+    orig :
+        Original data.
+    dimensions :
```
wtbarnes, line 90:
```suggestion
    orig : array-like
        Original data.
    dimensions : `tuple`
```

sunpy/image/resample.py
```diff
@@ -82,31 +82,44 @@ def resample(orig, dimensions, method='linear', center=False, minusone=False):
...
 
     trorder = [orig.ndim - 1] + list(range(orig.ndim - 1))
     for i in range(orig.ndim - 2, -1, -1):
         new_data = new_data.transpose(trorder)
-
         mint = scipy.interpolate.interp1d(old_coords[i], new_data,
-                                          bounds_error=False, fill_value=min(old_coords[i]), kind=method)
-        new_data = mint(dimlist[i])
+                                          bounds_error=False,
+                                          fill_value='extrapolate',
```
wtbarnes, line 120: Why don't we allow users to pass in a value for `fill_value` and then default to `'extrapolate'`?

wtbarnes:
(This should get a ~second~ third review from someone who knows ~a little more~ about resampling)

sunpy/image/resample.py
ayshih: Actually, 0 or 1

sunpy/image/resample.py
```diff
@@ -82,31 +82,44 @@ def resample(orig, dimensions, method='linear', center=False, minusone=False):
...
+    dimensions :
+        Dimensions of resamled data.
+    offset :
+        Either 0 or 0.5, depending on whether interpolation is at the edge or
+        centers of bins.
+    m1 :
+        For ``orig.shape = (i,j)`` & new dimensions ``= (x,y)``, if set to `False`
+        (default) ``orig`` is resampled by factors of ``(i/x) * (j/y)``,
+        otherwise ``orig`` is resampled by ``(i-1)/(x-1) * (j-1)/(y-1)``.
+        This prevents extrapolation one element beyond bounds of input array.
```
ayshih, line 99: I know this sentence is just copied from the docstring of the user-facing function, but I'm confused regardless.  What does this sentence mean?  What does it have to do with a parameter that just modifies the scaling factor?

dstansby: It's gibberish to me too, and seems to have been lifted directy from the IDL CONGRID docs: https://www.harrisgeospatial.com/docs/CONGRID.html

sunpy/image/resample.py
dstansby: Yep, old comment

sunpy/image/resample.py
dstansby: We could, but I'm not sure what the use case is for having a constant fill_value?

wtbarnes: Fair enough. Outside the scope of this PR either way.

sunpy/image/resample.py
ayshih: Ah, I get it now.  It actually makes more sense with the flipped order of sentences in the IDL documentation.

Consider `i=3` and `x=6`.

With `m1=0` (`minusone=False` in the user-facing function):
```
orig: |         |         |         |
new : |    |    |    |    |    |    |
```
3 pixels are being mapped to 6 pixels, so the factor is `i/x=3/6=0.5`.

The issue is that the outermost pixels of the output array require extrapolating off of the edges of the input array.  If you instead resample by the "minus one" dimensions, the outermost pixels in the input and output arrays are lined up, so no extrapolation is needed.

With `m1=1` (`minusone=True` in the user-facing function):
```
orig: |         |         |         |
new :    |   |   |   |   |   |   |
```
That is, here `2=0.5+1+0.5` pixels are being resampled to `5=0.5+4+0.5` pixels, which is why the factor is , so the factor is `(i-1)/(x-1)=2/5=0.4`.

PR #4179, opened 2020-05-18
Check for unintended behavior on pixels instead of coords

dstansby:
See  https://github.com/sunpy/sunpy/pull/3677#pullrequestreview-412087551 for discussion on this. It is better to perform the check on pixels, because `lon1 < lon2` does  not necessarily imply that `world2pix(lon1)` < `world2pix(lon2)` for all maps (in particular for heliographic maps).

### DIFF
--- a/sunpy/coordinates/utils.py
+++ b/sunpy/coordinates/utils.py
@@ -319,6 +319,11 @@ def get_rectangle_coordinates(bottom_left, *, top_right=None, width: u.deg = Non
     >>> height = 10 * u.arcsec
     >>> bottom_left, top_right = get_rectangle_coordinates(bottom_left, width=width, height=height)
 
+    Notes
+    -----
+    This function does not check whether ``top_right`` is above and to the right of ``bottom_left``
+    in coordinate space. Whether this condition holds when plotting a rectangle depends on the
+    orientation of the axes, and therefore needs a WCS for context.
     """
     if not (hasattr(bottom_left, 'transform_to') and
             hasattr(bottom_left, 'shape') and
--- a/sunpy/coordinates/utils.py
+++ b/sunpy/coordinates/utils.py
@@ -268,7 +268,8 @@ class GreatArc:
                         frame=Heliocentric).transform_to(self.start_frame)
 
 
-def get_rectangle_coordinates(bottom_left, *, top_right=None, width: u.deg = None, height: u.deg = None):
+def get_rectangle_coordinates(bottom_left, *, top_right=None,
+                              width: u.deg = None, height: u.deg = None):
     """
     Specify a rectangular region of interest in longitude and latitude in a given coordinate frame.
 
--- a/sunpy/coordinates/utils.py
+++ b/sunpy/coordinates/utils.py
@@ -1,7 +1,6 @@
 """
 Miscellaneous utilities related to coordinates
 """
-import warnings
 
 import numpy as np
 
@@ -9,7 +8,6 @@ import astropy.units as u
 from astropy.coordinates import BaseCoordinateFrame, SkyCoord
 
 from sunpy.coordinates import Heliocentric
-from sunpy.util.exceptions import SunpyUserWarning
 
 __all__ = ['GreatArc', 'get_rectangle_coordinates']
 
--- a/sunpy/coordinates/utils.py
+++ b/sunpy/coordinates/utils.py
@@ -275,12 +275,13 @@ def get_rectangle_coordinates(bottom_left, *, top_right=None, width: u.deg = Non
     Parameters
     ----------
     bottom_left : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
-        The bottom-left coordinate of the rectangle.
-        Supports passing both the bottom left and top right coordinates by passing with a shape of ``(2,)``.
+        The bottom-left coordinate of the rectangle. Supports passing both the
+        bottom left and top right coordinates by passing with a shape of ``(2,)``.
     top_right : `~astropy.coordinates.BaseCoordinateFrame` or `~astropy.coordinates.SkyCoord`
         The top-right coordinate of the rectangle.
-        If in a different frame than ``bottom_left`` and all required metadata for frame conversion is present,
-        ``top_right`` will be transformed to ``bottom_left`` frame.
+        If in a different frame than ``bottom_left`` and all required metadata
+        for frame conversion is present, ``top_right`` will be transformed to
+        ``bottom_left`` frame.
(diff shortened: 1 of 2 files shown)


sunpy/map/mapbase.py
```diff
@@ -1826,6 +1835,13 @@ def draw_rectangle(self, bottom_left, *, width: u.deg=None, height: u.deg=None,
 
         width = Longitude(top_right.spherical.lon - bottom_left.spherical.lon)
         height = Latitude(top_right.spherical.lat - bottom_left.spherical.lat)
+        if width < 0:
```
Cadair, line 1838: As an alternative to moving this from the utility into here, why not add a flag to the utility which disables this warning. I suggest this mainly because the idea of the utility was so that it could be used in other places (I plan to use it in the `dkist` package where the warning will be useful).

sunpy/map/mapbase.py
```diff
@@ -1537,16 +1537,25 @@ def submap(self, bottom_left, *, top_right=None, width: (u.deg, u.pix)=None, hei
...
+            raise ValueError("Both bottom_left and top_right when specified as Quantity objects "
+                             "must have shape (2,)")
 
         elif height is not None and width is not None:
             top_right = u.Quantity([bottom_left[0] + width, bottom_left[1] + height])
 
         x_pixels = u.Quantity([bottom_left[0], top_right[0]]).to_value(u.pix)
         y_pixels = u.Quantity([top_right[1], bottom_left[1]]).to_value(u.pix)
+        if x_pixels[0] > x_pixels[1]:
+            warnings.warn("The rectangle is inverted in the left/right (longitude) direction,  "
```
Cadair, line 1553: I don't think adding latitude & longitude to these error messages is a good idea, there is no reason other than convention that the left/right direction is normally longitude. You can easily construct WCSes where this would appear to be right but would actually be wrong and all manner of other weirdness.

sunpy/map/mapbase.py
dstansby: I'm not sure that the warning is appropriate for the utility, whose job it is to
- Convert `top_right` into the frame of `bottom_left`
or
- Add `width` and `height` to `bottom_left`.

Whether a warning makes sense then depends what you want to do with the coordinates. In addition, as you say above you can easily construct WCSes where x-y don't correspond to lon-lat, so you really need a WCS for context to see if it makes sense to raise a warning.

Cadair: hmm, yeah ok, it probably is better done in the context of the calling function. Maybe a little write up of this to the `Notes` section of the utility docstring so we don't wonder why it's not there in 3 months time :grinning: ?

ayshih: > you can easily construct WCSes where x-y don't correspond to lon-lat

I'll point out that this utility as written converts `bottom_left` to spherical coordinates and adds width as longitude and height as latitude (even if "longitude" and "latitude" aren't the correct terms in that coordinate frame).  Whether the WCS is something other than spherical (e.g., Cartesian) is ignored.

sunpy/map/mapbase.py
```diff
@@ -1829,6 +1839,13 @@ def draw_rectangle(self, bottom_left, *, width: u.deg = None, height: u.deg = No
 
         width = Longitude(top_right.spherical.lon - bottom_left.spherical.lon)
         height = Latitude(top_right.spherical.lat - bottom_left.spherical.lat)
+        if width < 0:
+            warnings.warn("The rectangle is inverted in the left/right (longitude) direction,  "
```
Cadair, line 1843:
```suggestion
            warnings.warn("The rectangle is inverted in the left/right direction,  "
```

Cadair:
We discussed this at the community meeting, and it's worth remembering that only submap of the known (3) usecases needs to / should do this check in pixel space. 

However, this is still an improvement on the current situation, so we should merge it pending review.

sunpy/coordinates/utils.py
```diff
@@ -318,6 +318,11 @@ def get_rectangle_coordinates(bottom_left, *, top_right=None, width: u.deg = Non
     >>> height = 10 * u.arcsec
     >>> bottom_left, top_right = get_rectangle_coordinates(bottom_left, width=width, height=height)
 
+    Notes
+    -----
+    This function does not check whether ``top_right`` is above and to the right of ``bottom_left``
+    in coordinate space. Whether this condition holds when plotting a rectangle depends on the
+    orientation of the axes, and therefore needs a WCS for context.
```
ayshih, line 325:
```suggestion
    ``width`` is always treated as an increase in longitude, but ``bottom_left`` may have a higher
    value of longitude than ``top_right`` due to the wrapping of the longitude angle.  Appropriate
    care should be taken when using this function's output to define a range of longitudes.

    ``height`` is always treated as an increase in latitude, but this function does not enforce
    that ``bottom_left`` has a lower value of latitude than ``top_right``, in case that orientation
    is valid for the intended use.
```

sunpy/map/mapbase.py
dstansby: In `draw_rectangle` width is hard-coded to apply to the longitude direction (see the above two lines), so having longitude here is correct. I can still remove if it would make for a better warning?

dstansby:
> Can you add wordage to the docstrings of both `.submap()` and `draw_rectangle()` to make it explicit that the rectangles are defined in spherical coordinates?

Sorry to be a pain, but can I do this in another PR given this one is still an improvement on what exists? Just tried to do it here, and I think more changes need making to test/catch/error on different width/height inputs.

ayshih:
> Sorry to be a pain, but can I do this in another PR given this one is still an improvement on what exists?

Okay

sunpy/map/mapbase.py
```diff
@@ -1787,6 +1797,13 @@ def draw_rectangle(self, bottom_left, *, width: u.deg = None, height: u.deg = No
         """
         Draw a rectangle defined in world coordinates on the plot.
 
+        This draws a rectangle that has corners at ``(bottom_left, top_right)``,
+        and has sides parallel to the x and y axes of the plot.
```
Cadair, line 1801: This wording is potentially confusing. They have axes aligned with the coordinate grid not the axes of the plot. I think they might be best described as lines of constant latitude / longitude.

dstansby: > This wording is potentially confusing.

I think you meant "wrong" 😆 . Will fix.

dstansby: I think latitude / longitude is potentially confusing (for a helioprojective map it might be interpreted as lines of Carrington lat/long), so I just went with coordinate axes, which is still a bit vague but I think true.

PR #4182, opened 2020-05-18
Add colormaps.color_tables to docs

dstansby:
I noticed this was missing from the docs. Once the docs build might need some fixes to the docstrings since they've not been built by sphinx before.


wtbarnes:
Online failures are unrelated

PR #4183, opened 2020-05-18
Fix bug in selecting hdu indices in sunpy.io.fits.read

### LINKED ISSUE #4181: Example of reading data/header with io.fits.read is broken
In the docs, we have [the following example of how to read the data and header directly from a file using `sunpy.io.fits.read`](https://docs.sunpy.org/en/stable/code_ref/map.html#fixing-map-metadata),

```python
[data,header] = sunpy.io.fits.read(filename)
```

There are a few problems here.

1. If you have 2 HDUs, this returns the first to `data` and the second to `header`, obviously not the intended behavior.
2. You can use the `hdus` kwarg to select only a specific HDU but it turns out this is broken because the rest of the function expects an `HDUList` and after we slice it, we don't cast it back to an `HDUList` so everything after this explodes in fun and unexpected ways.
3. Even accounting for the first two problems, this function returns a list of data,header pairs, not a list of data and headers so you get a `ValueError` when trying to return this to a tuple as above. Fixing items 1 and 2, this example could be amended to,

```python
data, header = sunpy.io.fits.read(filename)[0]
```

wtbarnes:
Fixes #4181 and cleans up a few tests for `sunpy.io.fits`

### DIFF
--- a/changelog/4183.docfix.rst
+++ b/changelog/4183.docfix.rst
@@ -1,2 +1,2 @@
-Fixed example usage of `~sunpy.io.fits.read` to account for the fact that it returns a list
+Fixed example usage of :func:`~sunpy.io.fits.read` to account for the fact that it returns a list
 of data-header pairs rather than the data-header pairs directly.
--- a/changelog/4183.docfix.rst
+++ b/changelog/4183.docfix.rst
@@ -1,0 +1,2 @@
+Fixed example usage of `~sunpy.io.fits.read` to account for the fact that it returns a list
+of data-header pairs rather than the data-header pairs directly.
--- a/sunpy/io/fits.py
+++ b/sunpy/io/fits.py
@@ -48,7 +48,7 @@ def read(filepath, hdus=None, memmap=None, **kwargs):
     ----------
     filepath : `str`
         The fits file to be read.
-    hdu: `int` or iterable
+    hdus: `int` or iterable
         The HDU indexes to read from the file.
 
     Returns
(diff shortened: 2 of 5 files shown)


sunpy/io/fits.py
```diff
@@ -71,7 +71,8 @@ def read(filepath, hdus=None, memmap=None, **kwargs):
             elif isinstance(hdus, collections.Iterable):
                 hdulist = [hdulist[i] for i in hdus]
 
-        hdulist.verify('silentfix+warn')
+        hdulist = fits.hdu.HDUList(hdulist)
+        [h.verify('silentfix+warn') for h in hdulist]
```
ayshih, line 75: Is this acceptable coding style compared to an explicit loop or to a `map()` call?  It's a list comprehension to create a list that isn't actually used...  The answer can be "yes"; I'm simply asking.

wtbarnes: From the [standard library docs](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions):

> List comprehensions provide a concise way to create lists.

So no, this should probably just be a `for` loop.

Cadair:
Can haz changelog?

sunpy/io/fits.py
wtbarnes: Fixed!

wtbarnes:
I added separate changelogs for the docfix and the bugfix. Do they need to be numbered as well? i.e. `4183.bugfix.1.rst` and `4183.docfix.2.rst`?

ayshih:
Nope

changelog/4183.docfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed example usage of `~sunpy.io.fits.read` to account for the fact that it returns a list
```
ayshih, line 1:
```suggestion
Fixed example usage of :func:`~sunpy.io.fits.read` to account for the fact that it returns a list
```

PR #3637, opened 2019-12-24
Attrs registration for clients

nabobalis:
Todo:
- [x] Have `print(Client)` print its attrs it supports
- [x] Fixes #3575

I have no idea why I wrote these or what they mean:
- [x] Automatic doc generation for attrs.
- [x] Make `print(Attr)` work
- [x] Add this
```
    @classmethod
    def register_values(cls):
        adict = {attrs.Instrument: []}
        return adict
```
So it would basically be `attrs.Instrument: '*'`
- [x]  make the keys (client, key) pairs
- [x]  Cache VSO attrs to prevent the long network call
- [x] Fido prints a list of clients.
- [x] printed the long name and description like
```
In [2]: a.Instrument.aia                                                                                                                                                                                                                                       
Out[2]: <Instrument('AIA'): Atmospheric Imaging Assembly>
```
This probably needs to be nice when the description is really long, like some of the physobs ones.

Fido printing:

![Screenshot from 2020-04-19 16-41-47](https://user-images.githubusercontent.com/1392107/79692380-b72cf380-825c-11ea-8678-a67493ceed89.png)

PR #4185, opened 2020-05-19
Fix specification of STEREO B in an example

ayshih:
This example was broken in #3637 because it was mistakenly switched to downloading from both STEREO spacecraft, when we actually want only STEREO B.  The example would non-deterministically error depending on which STEREO Map was actually used for the coordinate frame.


wtbarnes:
I was wondering why those docs were only failing intermittently...

nabobalis:
I forgot to undo that during the attrs change.

PR #4195, opened 2020-05-21
Clarify draw_rectangle behaviour

dstansby:
The code in `draw_rectangle` was doing all kinds of acrobatics with long/lat and units, which is fine for a plot where longitude-latitude align with the x-y axes, but I have no idea what wild things would happen if they weren't. The code is much simpler with a call to `world_to_pixel`.

In addition, this means that the behavior for an arbitrary WCS is well defined, as now indicated in the docstring:
```
This draws a rectangle that has corners at ``(bottom_left, top_right)``,
and that has sides parallel to the x and y axes of the plot.
```

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1836,8 +1836,8 @@ class GenericMap(NDData):
                                                            top_right=top_right,
                                                            width=width,
                                                            height=height)
-        bottom_left = self.world_to_pixel(bottom_left)
-        top_right = self.world_to_pixel(top_right)
+        bottom_left = self.world_to_pixel(bottom_left).to_value(u.pix)
+        top_right = self.world_to_pixel(top_right).to_value(u.pix)
         width = top_right[0] - bottom_left[0]
         height = top_right[1] - bottom_left[1]
 
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1785,7 +1785,10 @@ class GenericMap(NDData):
     def draw_rectangle(self, bottom_left, *, width: u.deg = None, height: u.deg = None,
                        axes=None, top_right=None, **kwargs):
         """
-        Draw a rectangle defined in world coordinates on the plot.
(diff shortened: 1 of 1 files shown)


sunpy/map/mapbase.py
```diff
@@ -1785,7 +1785,17 @@ def draw_limb(self, axes=None, **kwargs):
...
+
+        This draws a rectangle that has corners at ``(bottom_left, top_right)``,
+        and that has sides parallel to the x and y axes of the plot.
+
+        If ``width`` and ``height`` are specified, they are respectively added to the
+        longitude and latitude of the ``bottom_left`` coordinate to calculate a
+        ``top_right`` coordinate. This means that ``width`` and ``height`` will only
+        correspond to the width and height of the draw rectangle if the longitude
+        and latitude coordiante axes are parallel with the x and y axes of the pixels
+        that comprise the map.
```
dstansby, line 1798: @ayshih here is some text about what `width` and `height` mean. It feels a bit wordy to me, so suggestions/improvements very welcome.

Cadair:
This isn't making assumptions about which axes is x and which is y, the "x" coordinate is just the first one in the WCS. WCSAxes handles transforming them to pixel coordinates as defined by the WCS.

dstansby:
> This isn't making assumptions about which axes is x and which is y, the "x" coordinate is just the first one in the WCS.

The issue is the computation of `width` and `height`. Width is computed as
```
width = Longitude(top_right.spherical.lon - bottom_left.spherical.lon)
```
which is the difference in longitude between the coordinates, but when the rectangle is drawn this is used as the width along the x-axis. So if Longitude is on my y-axis, the `width` argument would currently specify the height of the rectangle as drawn.

> I don't really see any good reason to not use the transformation machinery in WCSAxes/matplotlib to do this.

I agree with this, but as far as I can tell the transformation machinery doesn't affect the `width` and `height` arguments to `plt.rectangle`, so these need to be computed manually anyway I think.

I'll try and come up with some simple examples to check the current behaviour of `draw_rectangle.`

PR #4197, opened 2020-05-21
Improve submap argument parsing

dstansby:
This builds on #4179 to improve the argument parsing of `submap()`. In particular:

- The allowed args/kwargs are different depending on whether `bottom_left` is a Quantity or SkyCoord, so dispatch on type to make the code clearer
- Add more specific error messages (e.g. instead of "any values of top_right, width or height specified must be Quantity", there are now separate checks and messages for top_right and width or height)
- Check for the right argument combination (e.g. top_right OR width and height) first
- If bottom_left is a SkyCoord, check that top_right is a SkyCoord
- If bottom_left is a SkyCoord, check that width and height are in degrees

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1588,7 +1588,6 @@ class GenericMap(NDData):
         Should take any valid input to submap() and return bottom_left and
         top_right in pixel coordinates.
         """
-        pass
 
     @_parse_submap_input.register(u.Quantity)
     def _parse_submap_quantity_input(self, bottom_left, top_right, width, height):
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1521,16 +1521,11 @@ class GenericMap(NDData):
             [268.24377, 254.83157, 268.24377, 321.89252],
             [249.99167, 265.14267, 274.61206, 240.5223 ]], dtype=float32)
         """
-        # Handle the case where bottom_left and top_right are really in bottom_left
-        if isinstance(bottom_left, SkyCoord) and bottom_left.shape == (2,):
-            top_right = bottom_left[1]
-            bottom_left = bottom_left[0]
-
         # Check that we have been given a valid combination of inputs
+        # [False, False, False] is valid if bottom_left contains the two corner coords
         if ([arg is not None for arg in (top_right, width, height)]
-                not in [[True, False, False], [False, True, True]]):
-            raise ValueError("If bottom_left is not a SkyCoord either top_right alone or "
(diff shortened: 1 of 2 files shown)


sunpy/map/mapbase.py
```diff
@@ -1520,34 +1521,28 @@ def submap(self, bottom_left, *, top_right=None, width: (u.deg, u.pix) = None, h
...
         """
-        if not isinstance(bottom_left, u.Quantity):
-            bottom_left, top_right = get_rectangle_coordinates(bottom_left,
-                                                               top_right=top_right,
-                                                               width=width,
-                                                               height=height)
+        # Handle the case where bottom_left and top_right are really in bottom_left
+        if isinstance(bottom_left, SkyCoord) and bottom_left.shape == (2,):
+            top_right = bottom_left[1]
+            bottom_left = bottom_left[0]
```
Cadair, line 1527: This is done in the `get_rectangle_helper` right? Why are we repeating it here? Can't we pass `top_right` through as none to the generic function?

sunpy/map/mapbase.py
```diff
@@ -1592,6 +1587,56 @@ def submap(self, bottom_left, *, top_right=None, width: (u.deg, u.pix) = None, h
...
+    def _parse_submap_coord_input(self, bottom_left, top_right, width, height):
+        if top_right is not None:
+            if not isinstance(top_right, SkyCoord):
+                raise TypeError("When bottom_left is a SkyCoord, top_right "
+                                "must also be a SkyCoord.")
+        else:
+            if not (width.unit.is_equivalent(u.deg) and
+                    height.unit.is_equivalent(u.deg)):
+                raise TypeError("When bottom_left is a SkyCoord, width and height "
+                                "must both be a Quantity in units convertable to degrees.")
```
Cadair, line 1629: Isn't this check also done in `get_rectangle_coordinates`? I would have thought that all the validation we need for this branch is in the helper, so can't we just call the helper and convert to pixels in this function?

sunpy/map/mapbase.py
```diff
@@ -1602,6 +1582,50 @@ def submap(self, bottom_left, *, top_right=None, width: (u.deg, u.pix) = None, h
         new_map = self._new_instance(new_data, new_meta, self.plot_settings)
         return new_map
 
+    @seconddispatch
+    def _parse_submap_input(self, bottom_left, top_right, width, height):
+        """
+        Should take any valid input to submap() and return bottom_left and
+        top_right in pixel coordinates.
+        """
+        pass
```
Cadair, line 1591:
```suggestion
```

PR #4199, opened 2020-05-25
Fix printing an Attr when registry is empty

Cadair:
If an attr has not entries in the values registry it should still print / repr


Cadair:
Fails are unreleated

PR #4203, opened 2020-05-27
Fixed non-relative paths in sphinx-gallery configuration

ayshih:
The paths in the sphinx-gallery configuration are supposed to be relative paths, not absolute.  I believe this will fix the failing doc build on CI.


nabobalis:
html-docs passed.

PR #4204, opened 2020-05-28
Add a couple of map error tests

dstansby:
I got distracted from something else and ended up with this. Should bump coverage a tiny little bit...


sunpy/map/tests/test_mapbase.py
```diff
@@ -717,7 +728,7 @@ def test_rotate_new_matrix(generic_map):
 
 
 def test_rotate_rmatrix_angle(generic_map):
-    with pytest.raises(TypeError):
+    with pytest.raises(ValueError, match="You cannot specify both an angle and a rotation matrix."):
```
dstansby, line 731: I think this is what was always intended for this test, previously it would `TypeError` because `angle` wasn't a quantity.

wtbarnes:
Looks good to me. I restarted the online 3.8 tests just for kicks, but I think it is just a JSOC thing

PR #4205, opened 2020-05-28
Improve Map histogram example

dstansby:
- Use `top_right=top_right` to avoid a new sunpy 2.0 deprecation warning
- Plot the submap before diving into the histogramming
- Change variable name from `bins` > `bin_edges` for clarity
- Use matplotlib built in `hist`
- Set the yscale to log is it is easier to see the histogram at low values

If anyone has any other suggestions for this example, happy to add them here.


nabobalis:
If I were to be really nitpicky, the legend could be moved to the top? 

https://52214-2165383-gh.circle-artifacts.com/0/docs/_build/html/generated/gallery/map/map_data_histogram.html#sphx-glr-generated-gallery-map-map-data-histogram-py

examples/map/map_data_histogram.py
```diff
@@ -50,8 +51,8 @@
             one_sigma[0], one_sigma[1]))
 plt.axvline(one_sigma[0], color='green')
 plt.axvline(one_sigma[1], color='red')
+plt.yscale('log')
 plt.legend()
```
nabobalis, line 55:
```suggestion
plt.legend(loc=9)
```

PR #4226, opened 2020-06-01
Reduced the tolerance of affine-transform tests to avoid numerical-precision issues

ayshih:
Follow-up to #4076

The affine-transform tests still have intermittent failures on OS X, and I no longer see indications of pixel shifts.  I suspect, without proof, that the lingering issue may be due to numerical precision at the limit of 64-bit floats (IEEE 754 standard).  The relative tolerance had been set to 1e-15, which is looser than the minimum precision possible in the standard (~1e-16), but not by terribly much.  I've learned that library math functions are not universally mandated to maintain accuracy to the full IEEE 754 precision – although OS X is allegedly better about this than other OSes – so such a stringent relative tolerance may be hitting some weirdness.

Accordingly, I've reduced the relative-tolerance value, and hopefully that will put the intermittent failures to bed.

### DIFF
--- a/sunpy/image/tests/test_transform.py
+++ b/sunpy/image/tests/test_transform.py
@@ -7,7 +7,7 @@ from sunpy.image.transform import affine_transform
 from sunpy.util import SunpyUserWarning
 
 # Tolerance for tests
-RTOL = 1.0e-15
+RTOL = 1.0e-10
 
 
 @pytest.fixture


ayshih:
Interesting read: https://stackoverflow.com/a/26502092

ayshih:
Upon thinking about this more, it'd be a miracle if this fixes the OS X builds because I believe the default non-zero absolute tolerance (for `np.allclose()`) will always end up trumping even an extremely stringent relative tolerance.

I've created #4235 to dump out a lot more diagnostics for any future failure.

PR #4231, opened 2020-06-02
Add rotated map plotting test

dstansby:
I thought it would be good to test that plotting of a rotated map works as expected, and that drawing a rectangle on to the rotated map also works as expected.

### DIFF
--- a/sunpy/map/tests/test_plotting.py
+++ b/sunpy/map/tests/test_plotting.py
@@ -47,6 +47,21 @@ def test_plot_aia171(aia171_test_map):
     aia171_test_map.plot()
 
 
+def test_plot_rotated_aia171(aia171_test_map):
+    # Check that plotting a rotated map and a rectangle works as expected
+
+    # Set rotation metadata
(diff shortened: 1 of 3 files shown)


sunpy/map/tests/test_plotting.py
```diff
@@ -47,6 +47,22 @@ def test_plot_aia171(aia171_test_map):
     aia171_test_map.plot()
 
 
+@figure_test
+def test_plot_rotated_aia171(aia171_test_map):
+    # Check that plotting a rotated map and a rectangle works as expected
+
+    # Set rotation metadata
+    aia171_test_map.meta['CROTA2'] = 45
```
dstansby, line 55: Maybe this line should call `.rotate()` instead of modifying the header?

Cadair: I think this is fine, because it's going to be a lot quicker than actually rotating the array, and it tests the code just fine.

dstansby:
Yep, just done. It plotted as expected 🎉 (and I now understand how plot_rectangle works)

PR #4233, opened 2020-06-02
The Search Events Object.

Raahul-Singh:
This is just the skeleton for the final SEO.
I still need to add tests, which I'll do once we finalize the API.

Will also add the changelog if we decide to keep it in `sunpy` and not move it to `astropy`.

### DIFF
--- a/sunpy/util/seo.py
+++ b/sunpy/util/seo.py
@@ -1,0 +1,86 @@
+"""
+This module implements functions that can be used to match rows across various Data Catalogues.
+All these functions will be coalesced into the Search Events Object.
+"""
+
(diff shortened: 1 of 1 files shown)


nabobalis:
What ended up happening with this code and the GSoC project?

Raahul-Singh:
> What ended up happening with this code and the GSoC project?

We successfully made the SEO, it's in my GSoC repository.
For now we are keeping it there, depending on further use, we may move it somewhere it is more useful.

nabobalis:
In that case I will close this and if we plan to make it an affiliated package, it can start off there before we consider moving it.

PR #4236, opened 2020-06-02
Adding an example using creating a Map from observations in RA-DEC

hayesla:
### Description
This is an example of how you to create a sunpy map from observations in celestial coordinates (RA/DEC) with LOFAR  observations to a helioprojective map. 

Here's an example of plotting the results after it creates the map:

Figure_1

and then after its rotated and cropped
Figure_2

I hope the comments are clear but feedback would be great!


hayesla:
This is also dependent on a PR to the sample-data repo https://github.com/sunpy/sample-data/pull/13

examples/units_and_coordinates/radec_to_hpc_map.py
```diff
@@ -0,0 +1,134 @@
...
+###############################################################################
+# To create a new `~sunpy.map.Map` header we need convert the reference coordinate
+# in RA-DEC to Helioprojective. To do this we will first create an `astropy.coordinates.SkyCoord`
+# of the reference coordinate from the header information. LOFAR observation are in
+# the J2000 epoch (this information is in the header).
+reference_coord = SkyCoord(header['crval1']*u.deg, header['crval2']*u.deg,
+                           frame='gcrs',
+                           obstime=obstime,
+                           distance=sun.earth_distance(obstime),
+                           equinox='J2000')
```
ayshih, line 60: Questions/comments:

- Are you sure that the coordinates provided are in GCRS, as opposed to the ICRF-based frame of ICRS with a shifted origin?  The critical difference is whether to include stellar aberration due to observer motion (which shifts coordinates by ~20 arcseconds).
- If using GCRS is in fact correct, you should use the `GCRS` keyword arguments `obsgeoloc` and `obsgeovel` to precisely specify the observer location.
- The argument `equinox='J2000'` doesn't actually do anything here.  It's not used by `GCRS`, nor by any frame in the transformation path to `Helioprojective`.

examples/units_and_coordinates/radec_to_hpc_map.py
```diff
@@ -0,0 +1,134 @@
...
+# Now we need to get the other parameters from the header that will be used
+# to create the new header - here we can get the cdelt1 and cdelt2 which are
+# the spatial scales of the data axes.
+cdelt1 = (np.abs(header['cdelt1'])*u.deg).to(u.arcsec)
+cdelt2 = (np.abs(header['cdelt2'])*u.deg).to(u.arcsec)
+
+##########################################################################
+# We will also required the P angle at the observation time - the image will
+# need to be rotated by this angle.
+P1 = sun.P(obstime)
```
ayshih, line 84: Strictly speaking, this is not the correct angle.  `P()` returns the angle between Earth north and solar north as seen from Earth, but Earth north is not exactly aligned with GCRS north.

ayshih: This is not particularly anything to be concerned about.  The difference will likely be very small; that 's why I said "strictly speaking".

examples/units_and_coordinates/radec_to_hpc_map.py
ayshih: Incidentally, since you specifically note that LOFAR say that their coordinates are at the reference epoch of J2000.0, they may not be using an ICRF-based frame like ICRS or GCRS, which don't really have a reference epoch.  They could instead be using an equatorial coordinate system with a reference epoch (e..g, Astropy's `PrecessedGeocentric` for the mean equatorial coordinate system, which even at J2000.0 isn't exactly the same as GCRS).

Aren't coordinates fun?

examples/units_and_coordinates/radec_to_hpc_map.py
```diff
@@ -0,0 +1,134 @@
...
+                           obstime=obstime,
+                           distance=sun.earth_distance(obstime),
+                           equinox='J2000')
+
+###########################################################################
+# To convert the reference coordinate to Helioprojective we need the location of the
+# observer (i.e. where the observation was taken). We first establish the location on
+# Earth from which the observation takes place. We can convert this to a SkyCoord at the
+# observation time.
+lofar_loc = EarthLocation(lat=52.905329712*u.deg, lon=6.867996528*u.deg)
```
dstansby, line 67: Is it possible to get this info from the header?

examples/units_and_coordinates/radec_to_hpc_map.py
```diff
@@ -0,0 +1,134 @@
...
+
+new_header = sunpy.map.make_fitswcs_header(data, reference_coord_arcsec,
+                                           reference_pixel=u.Quantity([header['crpix1']-1, header['crpix2']-1]*u.pixel),
+                                           scale=u.Quantity([cdelt1, cdelt2]*u.arcsec/u.pix),
+                                           rotation_angle=-P1,
+                                           wavelength=frequency.to(u.MHz),
+                                           observatory='LOFAR')
+
+##########################################################################
+# Let's now inspect the new header
```
dstansby, line 104:
```suggestion
# Lets inspect the new header
```

examples/units_and_coordinates/radec_to_hpc_map.py
```diff
@@ -0,0 +1,134 @@
...
+##########################################################################
+# We will also required the P angle at the observation time - the image will
+# need to be rotated by this angle.
+P1 = sun.P(obstime)
+
+##########################################################################
+# Now we can use this information to create a new header using the helper
+# function `~sunpy.map.make_fitswcs_header()`. This will create a MetaDict
+# which we contain all the necessay WCS information to create a `~sunpy.map.Map`.
+# What we provide to this includes a reference coordinate (in HPC), the spatial
```
dstansby, line 90:
```suggestion
# We provide a reference coordinate (in HPC), the spatial
```

examples/units_and_coordinates/radec_to_hpc_map.py
```diff
@@ -0,0 +1,134 @@
...
+##########################################################################
+# Lets now create a `~sunpy.map.Map`
+lofar_map = sunpy.map.Map(data, new_header)
+
+##########################################################################
+# We can now plot this map and inspect it
+fig = plt.figure()
+ax = fig.add_subplot(projection=lofar_map)
+lofar_map.plot(cmap='viridis')
+lofar_map.draw_limb()
```
dstansby, line 116: Slightly off topic question, if you just load the original data/header, is sunpy clever enough draw the limb correctly on the RA-DEC grid?

examples/units_and_coordinates/radec_to_hpc_map.py
```diff
@@ -0,0 +1,134 @@
...
+plt.show()
+
+##########################################################################
+# We can now rotate the image and create a submap in the field of view of interest.
```
dstansby, line 120:
```suggestion
# We can now rotate the image so that solar north is pointing up and create a submap in the field of view of interest.
```

examples/units_and_coordinates/radec_to_hpc_map.py
```diff
@@ -0,0 +1,134 @@
...
+lofar_map_rotate = lofar_map.rotate()
+bl = SkyCoord(-1500*u.arcsec, -1500*u.arcsec, frame=lofar_map_rotate.coordinate_frame)
+tr = SkyCoord(1500*u.arcsec, 1500*u.arcsec, frame=lofar_map_rotate.coordinate_frame)
+lofar_submap = lofar_map_rotate.submap(bl, tr)
+
+##########################################################################
+# Now lets plot this map, and overplot the contours
```
dstansby, line 127:
```suggestion
# Now lets plot this map, and overplot some contours
```

hayesla:
thanks @dstansby  I'll get on this later today - I've somewhat neglected this PR :P


> * Would be good to have replies to @ayshih's comments before this is merged
indeed!
> * Has the new sample data been uploaded? It looks like the documentation build is struggling to find it.
Yep - this'll be fixed in next PR

examples/units_and_coordinates/radec_to_hpc_map.py
```diff
@@ -0,0 +1,134 @@
...
+
+import sunpy.data.sample
+import sunpy.map
+from sunpy.coordinates import frames, sun
+
+##############################################################################
+# We will first begin be reading in the header and data from the FITS file.
+hdu = fits.open(sunpy.data.sample.LOFAR_IMAGE)
+header = hdu[0].header
+data = np.squeeze(hdu[0].data)
```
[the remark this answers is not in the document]
hayesla: the data has a shape of (1, 1, 800, 800) - to allow for data cubes of time and polarization. The data in this header is just one intensity image so needs to be squeezed to (800, 800)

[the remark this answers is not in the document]
hayesla: The data has a shape (1, 1, 800, 800) to allow for the data to be in a cube to include different frequencies and stokes to be included. This file however is just one frequency so need to squeeze to be (800, 800)

examples/units_and_coordinates/radec_to_hpc_map.py
hayesla: it really should be but its not in the header output information :(

examples/units_and_coordinates/radec_to_hpc_map.py
hayesla: coordinates 🤯

examples/units_and_coordinates/radec_to_hpc_map.py
hayesla: it _should_ work be if there was some sort of observer location in the header - but there is not!

examples/units_and_coordinates/radec_to_hpc_map.py
hayesla: good question - so as far as I can tell it should be GCRS (this is mainly based on testing with an eclipse observation as we know exactly where the moon should be).

I also reached out to some people at ASTRON and they think its also in gcrs with comparisons with NRH observations, but they are going to look into it some more

hayesla: also the more I seem to learn about coordinates the more i'm getting confused

....ignorance was bliss

hayesla: So yes, according to them it should be in gcrs apparently (this example gives same results compared to their own codes)

examples/units_and_coordinates/radec_to_hpc_map.py
```diff
@@ -0,0 +1,133 @@
...
```
dstansby, line 38: Add a comment explaining this line? I'm not sure what it does.

examples/units_and_coordinates/radec_to_hpc_map.py
```diff
@@ -0,0 +1,133 @@
...
+##########################################################################
+# Now we can convert the `reference_coord` to the HPC coordinate frame
+reference_coord_arcsec = reference_coord.transform_to(frames.Helioprojective(observer=lofar_coord))
+
+##########################################################################
+# Now we need to get the other parameters from the header that will be used
+# to create the new header - here we can get the cdelt1 and cdelt2 which are
+# the spatial scales of the data axes.
+cdelt1 = (np.abs(header['cdelt1'])*u.deg).to(u.arcsec)
+cdelt2 = (np.abs(header['cdelt2'])*u.deg).to(u.arcsec)
```
dstansby, line 77: Can you get the units from `cunit{1,2}` instead of manually supplying them as degrees?

examples/units_and_coordinates/radec_to_hpc_map.py
```diff
@@ -0,0 +1,133 @@
...
+
+##########################################################################
+# We will also required the P angle at the observation time - the image will
```
dstansby, line 80: Might be good to spell out what the P angle is?

dstansby (approved):
This is great! I have left a few optional comments, but not blocking.

dstansby:
Not sure why there's a coordinate devdeps doctest failing here, maybe @ayshih has an idea? Otherwise this looks 👍 to me

ayshih:
We're going to have coordinates devdeps doctest failures because astropy/astropy#10970 was just merged.  It's fine to ignore for this PR.

dstansby (approved):

ayshih (changes requested):
(pending)

examples/units_and_coordinates/radec_to_hpc_map.py
ayshih: I don't think a change to the code is needed, but I'd like to see it noted in the comment.

ayshih (changes requested):

examples/units_and_coordinates/radec_to_hpc_map.py
```diff
@@ -0,0 +1,139 @@
...
+frequency = header['crval3']*u.Hz
+
+###############################################################################
+# To create a new `~sunpy.map.Map` header we need convert the reference coordinate
+# in RA-DEC to Helioprojective. To do this we will first create an `astropy.coordinates.SkyCoord`
+# of the reference coordinate from the header information.
+reference_coord = SkyCoord(header['crval1']*u.Unit(header['cunit1']), header['crval2']*u.Unit(header['cunit2']),
+                           frame='gcrs',
+                           obstime=obstime,
+                           distance=sun.earth_distance(obstime))
```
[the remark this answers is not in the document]
ayshih: In other words:
```python
lofar_pos, lofar_vel = lofar_loc.get_gcrs_posvel(obstime)
reference_coord = SkyCoord(...
                           frame='gcrs',
                           obstime=obstime,
                           obsgeoloc=lofar_pos, obsgeovel=lofar_vel,
                           distance=sun.earth_distance(obstime))
)
```

Strictly speaking, the Sun-observer distance isn't the same as the Sun-Earth distance, so you should actually do something like:
```python
lofar_gcrs = SkyCoord(lofar_loc.get_gcrs(obstime))
reference_coord = SkyCoord(...
                           frame='gcrs',
                           obstime=obstime,
                           obsgeoloc=lofar_gcrs.cartesian, obsgeovel=lofar_gcrs.velocity,
                           distance=lofar_gcrs.hcrs.distance)  # the `distance` component in HCRS is the distance from the Sun
)

File: examples/units_and_coordinates/radec_to_hpc_map.py (hunk 2 above)
Comment (hayesla) [reply examples/units_and_coordinates/radec_to_hpc_map.py:97]: I don't know exactly what should be said here without it being too clunky in regards to this @ayshih - suggestions?

File: sunpy/data/_sample.py [hunk 14]
```diff
@@ -46,6 +46,7 @@
     "GBM_TIMESERIES": "glg_cspec_n5_110607_v00.pha",
     "RHESSI_TIMESERIES": "hsi_obssumm_20110607_025.fits",
     "NORH_TIMESERIES": "tca110607.fits",
+    "LOFAR_IMAGE": "LOFAR_70MHZ_20190409_131136.fits",
```
nabobalis, line 49: Was this added (or already exist in the sample data repo?

hayesla: yep it exists - it was added here https://github.com/sunpy/sample-data/pull/14

examples/units_and_coordinates/radec_to_hpc_map.py
```diff
@@ -0,0 +1,148 @@
...
+# Now we need to get the other parameters from the header that will be used
+# to create the new header - here we can get the cdelt1 and cdelt2 which are
+# the spatial scales of the data axes.
+cdelt1 = (np.abs(header['cdelt1'])*u.deg).to(u.arcsec)
+cdelt2 = (np.abs(header['cdelt2'])*u.deg).to(u.arcsec)
+
+##################################################################################
+# We will also need knowledge of the P-angle at the observation time, which is the
+# position angle between geocentric north and solar north as observed from Earth.
+# The image will need to be rotated by this angle so that solar north is pointed up.
```
ayshih, line 96:
```suggestion
##################################################################################
# Finally, we need to specify the orientation of the HPC coordinate grid because
# GCRS north is not in the same direction as HPC north.  For convenience, we use
# :func:`~sunpy.coordinates.sun.P` to calculate this relative rotation angle,
# although due to subtleties in definitions, the returned value is inaccurate by
# 2 arcmin, equivalent to a worst-case shift of 0.6 arcsec for HPC coordinates
# on the disk.
```

examples/units_and_coordinates/radec_to_hpc_map.py
ayshih: I've suggested an expansion of the comment.  While the inaccuracy by using `P()` is very small, it's larger than I would have guessed, so I think it's worthwhile to include an estimate of the inaccuracy.

ayshih (approved):
Looks good, should be squash merged

nabobalis:
Squash merged.

PR #4248, opened 2020-06-04
Fix figure tests astropydev installation in tox.ini

dstansby:


tox.ini
```diff
@@ -51,7 +51,7 @@ deps =
     # The devdeps factor is intended to be used to install the latest developer version.
     # of key dependencies.
     devdeps: cython
-    devdeps,figure_astropydev: git+https://github.com/astropy/astropy
+    devdeps,figure_py37_astropydev: git+https://github.com/astropy/astropy
```
nabobalis, line 54: We should update the envlist at the top of the file for both figure environments.

PR #4256, opened 2020-06-05
Ensure that SunpyDeprecationWarnings are raised as errors in the test suite

Cadair:
### Description

It seems that at some point in the past we stopped raising SunpyDeprecationWarnings as errors in the test suite. This is *bad* :sob:

### DIFF
--- a/examples/units_and_coordinates/SDO_to_STEREO_Coordinate_Conversion.py
+++ b/examples/units_and_coordinates/SDO_to_STEREO_Coordinate_Conversion.py
@@ -54,13 +54,11 @@ for i, m in enumerate(maps.values()):
 
 ###############################################################################
 # We are now going to pick out a region around the south west corner:
-aia_width = 200 * u.arcsec
-aia_height = 250 * u.arcsec
 aia_bottom_left = SkyCoord(-800 * u.arcsec,
                            -300 * u.arcsec,
                            frame=maps['AIA'].coordinate_frame)
-aia_top_right = SkyCoord(aia_bottom_left.Tx + aia_width,
-                         aia_bottom_left.Ty + aia_height,
+aia_top_right = SkyCoord(-600 * u.arcsec,
+                         -50 * u.arcsec,
                          frame=maps['AIA'].coordinate_frame)
 
 ###############################################################################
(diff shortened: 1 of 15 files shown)


Cadair:
There was one real warning in the online build, the html build fails looked like server issues.

examples/units_and_coordinates/SDO_to_STEREO_Coordinate_Conversion.py
```diff
@@ -54,13 +54,11 @@
...
 ###############################################################################
 # We are now going to pick out a region around the south west corner:
-aia_width = 200 * u.arcsec
-aia_height = 250 * u.arcsec
 aia_bottom_left = SkyCoord(-800 * u.arcsec,
                            -300 * u.arcsec,
                            frame=maps['AIA'].coordinate_frame)
-aia_top_right = SkyCoord(aia_bottom_left.Tx + aia_width,
-                         aia_bottom_left.Ty + aia_height,
+aia_top_right = SkyCoord(-600 * u.arcsec,
```
wtbarnes, line 60: Maybe just get rid of `aia_top_right` and just use `aia_width` and `aia_height`?

sunpy/instr/tests/test_aia.py
```diff
@@ -21,7 +22,8 @@ def original(request):
 
 @pytest.fixture
 def prep_map(original):
-    return aiaprep(original)
+    with pytest.warns(SunpyDeprecationWarning):
```
wtbarnes, line 25: What does this do in the fixture? Just catches the warning so pytest doesn't raise an error for this deprecation warning?

Cadair: yep

examples/units_and_coordinates/SDO_to_STEREO_Coordinate_Conversion.py
Cadair: It was marginally easier lower down to do it this way around. :shrug:

sunpy/instr/tests/test_aia.py
dstansby: And also fails the test if the code *doesn't* raise a warning.

wtbarnes: Ah ok. So works the same in a fixture as in a normal test 👍

PR #4257, opened 2020-06-05
Transforming a SunPy frame to a class now preserves shared frame attributes

### LINKED ISSUE #4237: HeliocentricEarthEcliptic frame does not copy obstime when transforming to HGS
```python
>>> hee = HeliocentricEarthEcliptic(45*u.deg, 35*u.deg, 17*u.km, obstime="2020-01-01") 
>>> hee.transform_to(HeliographicStonyhurst) 
```
```
ConvertError                              Traceback (most recent call last)
<ipython-input-39-8ab95e21b2a5> in <module>
      1 hee = HeliocentricEarthEcliptic(45*u.deg, 35*u.deg, 17*u.km, obstime="2020-01-01")
----> 2 hee.transform_to(HeliographicStonyhurst)

~/.virtualenvs/ndcube-dev/lib/python3.8/site-packages/astropy/coordinates/baseframe.py in transform_to(self, new_frame)
   1196             msg = 'Cannot transform from {0} to {1}'
   1197             raise ConvertError(msg.format(self.__class__, new_frame.__class__))
-> 1198         return trans(self, new_frame)
   1199 
   1200     def is_transformable_to(self, new_frame):

~/.virtualenvs/ndcube-dev/lib/python3.8/site-packages/astropy/coordinates/transformations.py in __call__(self, fromcoord, toframe)
   1390 
   1391             curr_toframe = t.tosys(**frattrs)
-> 1392             curr_coord = t(curr_coord, curr_toframe)
   1393 
   1394         # this is safe even in the case where self.transforms is empty, because

~/.virtualenvs/ndcube-dev/lib/python3.8/site-packages/astropy/coordinates/transformations.py in __call__(self, fromcoord, toframe)
   1208     def __call__(self, fromcoord, toframe):
   1209 
-> 1210         M, vec = self.transform_func(fromcoord, toframe)
   1211         newrep = self._apply_transform(fromcoord, M, vec)
   1212 

~/.virtualenvs/ndcube-dev/lib/python3.8/site-packages/sunpy/coordinates/transformations.py in wrapped_func(*args, **kwargs)
    174                 _layer_level += 1
    175 
--> 176             result = func(*args, **kwargs)
    177 
    178             if debug_output:

~/.virtualenvs/ndcube-dev/lib/python3.8/site-packages/sunpy/coordinates/transformations.py in hcrs_to_hgs(hcrscoord, hgsframe)
    541     """
    542     if hgsframe.obstime is None:
--> 543         raise ConvertError("To perform this transformation, the coordinate"
    544                            " frame needs a specified `obstime`.")
    545 

ConvertError: To perform this transformation, the coordinate frame needs a specified `obstime`.
```

ayshih:
Fixes #4237

Transforming a SunPy coordinate frame to a class now instantiates the destination frame with the (non-default) values in the original frame for any shared frame attributes (e.g., `obstime`) instead of being all default values.  This is particularly important for us because `obstime` in our frames defaults to `None`.

Thus, this now works:
```python
>>> from sunpy.coordinates import HeliocentricEarthEcliptic, HeliographicStonyhurst
>>> hee = HeliocentricEarthEcliptic(45*u.deg, 35*u.deg, 17*u.km, obstime="2020-01-01") 
>>> hee.transform_to(HeliographicStonyhurst) 
<HeliographicStonyhurst Coordinate (obstime=2020-01-01T00:00:00.000): (lon, lat, radius) in (deg, deg, km)
    (46.71542196, 28.19372091, 17.)>
```

This also means that there will no longer be transformation issues when supplying a non-HGS coordinate as the `observer` frame attribute (as long as that non-HGS coordinate is in a SunPy frame).

I've already created an upstream PR for this (astropy/astropy#10453), so assuming that PR is accepted, this fix can eventually be removed.

### DIFF
--- a/changelog/4257.bugfix.rst
+++ b/changelog/4257.bugfix.rst
@@ -1,0 +1,1 @@
+Transforming any `sunpy.coordinates` frame to a frame class (as opposed to an instantiated frame) will now preserve the values for any shared frame attributes (e.g., ``obstime``).
--- a/sunpy/coordinates/frames.py
+++ b/sunpy/coordinates/frames.py
@@ -4,6 +4,8 @@ Common solar physics coordinate systems.
 This submodule implements various solar physics coordinate frames for use with
 the `astropy.coordinates` module.
 """
+import inspect
+
 import numpy as np
 
 import astropy.units as u
@@ -131,6 +133,25 @@ class SunPyBaseCoordinateFrame(BaseCoordinateFrame):
         property in `~astropy.coordinates.BaseCoordinateFrame`.
         """
         return self.data.size if self.has_data else 0
+
+    def transform_to(self, new_frame):
+        """
+
+        Notes
+        -----
+        If a class is specified, the new frame will be instantiated using the values
+        from this frame for any shared frame attributes (e.g., ``obstime``) that do
+        not have default values.  All other frame attributes will have default values.
+        """
+        # The docstring above will be prepended by the parent method's docstring
+
+        if inspect.isclass(new_frame):
+            nondefault = set(self.frame_attributes.keys()) - set(self._attr_names_with_defaults)
+            shared_attr = nondefault.intersection(new_frame.frame_attributes.keys())
+            new_frame = new_frame(**dict([(attr, getattr(self, attr)) for attr in shared_attr]))
+        return super().transform_to(new_frame)
+
+    transform_to.__doc__ = BaseCoordinateFrame.transform_to.__doc__ + transform_to.__doc__
 
     def __str__(self):
         """
--- a/sunpy/coordinates/tests/test_transformations.py
+++ b/sunpy/coordinates/tests/test_transformations.py
@@ -753,13 +753,15 @@ def test_array_obstime():
     assert isinstance(t2.frame, Helioprojective)
 
 
-_frameset1 = [HeliographicStonyhurst, HeliocentricInertial]
-_frameset2 = [HeliographicCarrington, Heliocentric, Helioprojective]
+_frames_wo_observer = [HeliographicStonyhurst, HeliocentricInertial,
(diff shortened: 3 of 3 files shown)


wtbarnes:
🎉  This is great!

Cadair:
@ayshih Do you think we should merge this before the concept gets a review upstream? I am hesitant to deviate away from the core behaviour like this.

ayshih:
I think it's safe to merge, but I'm hardly objective.  I haven't come up with how this could cause problems (given that `SkyCoord` already runs rampant with preserving frame attributes), but it's possible that I'm missing something.

ayshih:
I'll note that I haven't yet looked into the test failures on the Astropy PR, so definitely should do that before merging this PR.

dstansby:
For the example above, I would have guessed that `hee.transform_to(HeliographicStonyhurst) ` would raise an error, since the `HeliographicStonyhurst` frame is poorly defined without an `obstime`. Instead of silently propagating the obstime in `hee`, perhaps it would be better to require `hee.transform_to(HeliographicStonyhurst(obstime=hee.obstime))` instead?

ayshih:
I'm converting this PR back to a "draft", and it will likely be closed pending discussion over on Astropy.

> For the example above, I would have guessed that `hee.transform_to(HeliographicStonyhurst) ` would raise an error, since the `HeliographicStonyhurst` frame is poorly defined without an `obstime`.

I'll point out that if `hee` were a `SkyCoord`, that call already works.  It's essentially equivalent to `.transform_to("heliographic_stonyhurst")` or simply `.heliographic_stonyhurst`.  The silent propagation of attributes is the default behavior of `SkyCoord`.

The question is what should happen for that same call for `hee` when it is *not* a `SkyCoord`.

ayshih:
No longer the approach

PR #4264, opened 2020-06-06
Remove warnings from test_mapbase.py

dstansby:
- Filter out inverted submap/rectangle warnings
- Add full observer info to the `generic_map` and `simple_map` fixtures, to avoid having to catch a missing observer warning whenever they're used

### DIFF
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -227,7 +227,8 @@ def test_heliographic_latitude(generic_map):
 
 
 def test_heliographic_longitude(generic_map):
-    assert u.allclose(generic_map.heliographic_longitude, 0 * u.deg)
+    # Needs a small tolerance to account for 32bit rounding errors
+    assert u.allclose(generic_map.heliographic_longitude, 0 * u.deg, atol=1e-15*u.deg)
 
 
 def test_units(generic_map):
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -11,7 +11,7 @@ import pytest
 
 import astropy.units as u
 import astropy.wcs
-from astropy.coordinates import SkyCoord
+from astropy.coordinates import Latitude, SkyCoord
 from astropy.io import fits
 from astropy.io.fits.verify import VerifyWarning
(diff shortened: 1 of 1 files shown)


sunpy/map/tests/test_mapbase.py
```diff
@@ -101,7 +108,9 @@ def simple_map():
     ref_pix = [1, 1] * u.pix
     scale = [2, 1] * u.arcsec / u.pix
     header = sunpy.map.make_fitswcs_header(data, ref_coord, reference_pixel=ref_pix, scale=scale)
-
+    header['crlt_obs'] = 0
```
wtbarnes, line 111: Instead of setting these explicitly, couldn't you just make this the observer of `ref_coord`?

sunpy/map/tests/test_mapbase.py
```diff
@@ -82,7 +86,10 @@ def generic_map():
         'PC2_2': 0,
         'NAXIS1': 6,
         'NAXIS2': 6,
-        'date-obs': '1970-01-01T00:00:00',
+        'date-obs': dobs.isot,
```
wtbarnes, line 89: Is there a reason we explicitly construct header as a dictionary here rather than using `make_fitswcs_header`? I understand this was already in place before this PR.

sunpy/map/tests/test_mapbase.py
```diff
@@ -194,8 +203,7 @@ def test_detector(generic_map):
 
 
 def test_dsun(generic_map):
-    with pytest.warns(SunpyUserWarning, match='Missing metadata for observer: assuming Earth-based observer.*'):
```
wtbarnes, line 197: Removing the need to catch this everywhere is certainly a good idea, but do we still have a test somewhere that makes sure this warning is raised when the observer metadata is missing?

sunpy/map/tests/test_mapbase.py
Cadair: this probably existed long before `make_fitswcs_header`

Cadair:
I don't really see much point in backporting these kind of changes?

dstansby:
Yeah, seems unnecessary faff to backport.

dstansby:
I think this is good to go?

Cadair:
Could do with a second review.

PR #4267, opened 2020-06-08
Fixed missing transformation errors when `obstime` is `None`

ayshih:
The transformations to/from SunPy 1.0 frames are fairly permissive when they encounter `obstime=None` for a frame and try to transfer it from the other frame in the transformation, or even the `observer` attribute if it's an observer-based frame.  The behavior is no longer permissive if attempting to transform beyond that subgraph.  (The permissive behavior was primarily motivated by our `observer` frame attribute, but the specific bug there has been fixed by #4266, and the concept of under-defined `SkyCoord`-less transformations may be going away in the future.  This PR leaves this permissive behavior untouched.)

The transformations to/from SunPy 1.1 frames (HEE, GSE, GEI, HCI) don't fully handle all situations with `obstime=None`, and weren't raising transformation errors.  Some of the transformation output in edge cases could be outright wrong.

This PR adds a bunch of transformation errors.

### DIFF
--- a/changelog/4267.bugfix.rst
+++ b/changelog/4267.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed a bug where transformation errors were not getting raised in some situations when a coordinate frame had ``obstime`` set to the default value of ``None`` and `~astropy.coordinates.SkyCoord` was not being used.
+Users are recommended to use `~astropy.coordinates.SkyCoord` to manage coordinate transformations unless they have a specific reason not to.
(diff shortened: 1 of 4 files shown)


ayshih:
The idea is to always have the check when transforming between an Astropy frame and a SunPy frame, for when our frame has `obstime=None`, because copying may be weird.  The check already exists for the HCRS<->HGS transformations, so we need analogous checks for the HME<->HEE and HME<->GEI transformations.

The HGS<->HCI checks are a bit different: it'll only error when `obstime=None` on both ends, so the transformations are not defined.

The HEE<->GSE checks should actually be the same way as the HGS<->HCI checks, and so I'll fix that.

ayshih:
Okay, HEE<->GSE transformations are now fixed.  Plus, I realized that there were unnecessary computations in those transformations, so I also simplified them.

PR #4270, opened 2020-06-10
Allow hiding of ticks and labels via coord_params in ArrayAnimatorWCS

Cadair:
I had use for this in NDCube

### DIFF
--- a/changelog/4270.feature.rst
+++ b/changelog/4270.feature.rst
@@ -1,0 +1,2 @@
+Allow `sunpy.visualization.animators.ArrayAnimatorWCS` to disable ticks for
+a coordinate, by setting ``ticks: False`` in the ``coord_params`` dictionary.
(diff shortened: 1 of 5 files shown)


Cadair:
Yeah, that will work if you don't have `plt.ion()` get I guess, but there is still nothing you can do if you do. I will have a play.

Cadair:
> Given that we know the WCS isn't changing, only the slices argument[...]

Thinking about this more, we actually have an even stronger guarantee than this as we know we are not moving `'x'` and `'y'` around in the slices list, so the axes/is being displayed on the plot are not changing. Under this assumption there are lots of potential optimisations that could be done at the `WCSAxes` layer to prevent recalculation and redrawing of a bunch of things.

All this being said, I would like this PR to be a bug fix, a small change which allows something to work in the same API that already exists. So I would rather we keep it in the `coord_params` dict for now. I am happy to move it to it's own key `tick_visibility` or something if you think that would make it less confusing?

dstansby (changes requested):
Okay, have looked over this again and it seems fine. The docstring of `ArrayAnimatorWCS` just needs updating to reflect this new option, and a rebase and passing CI would be good too.

Cadair:
Good timing I have taken a trip back into NDCube this week :laughing:

dstansby:
Removed the backport tag as this is a new feature.

nabobalis (approved):

changelog/4270.feature.rst
```diff
@@ -0,0 +1,2 @@
+Allow `sunpy.visualization.animators.ArrayAnimatorWCS` to disable ticks for
```
dstansby, line 1: Can you extend this a bit to say how this is done? Something like "This can be done by setting `ticks=False` in the `coord_params` dictionary."

Cadair: done

dstansby (approved):
Anyone can merge if the CI passes, or the failures are un-related (which I think they will be)

nabobalis:
Looks like some servers are having issues.

PR #4275, opened 2020-06-11
Fix some time comparisions in tests

dstansby:
I'm not sure what changed in astropy to make these fail, but these are just tests, and in both cases the previous code was odd anyway, so I haven't worked out why these are failing.

### DIFF
--- a/sunpy/instr/goes.py
+++ b/sunpy/instr/goes.py
@@ -1357,7 +1357,6 @@ def flux_to_flareclass(goesflux: u.watt/u.m**2):
 
 
 def _assert_chrono_order(obstime):
-    chrono_check = np.array(obstime) - np.roll(obstime, 1)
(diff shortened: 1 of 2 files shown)


sunpy/instr/goes.py
```diff
@@ -1357,7 +1357,6 @@ def flux_to_flareclass(goesflux: u.watt/u.m**2):
 
 
 def _assert_chrono_order(obstime):
-    chrono_check = np.array(obstime) - np.roll(obstime, 1)
```
dstansby, line 1360: `np.roll(obstime)` was returning an array of `Time` (instead of just a `Time`), which was causing this to fail.

Cadair:
I agree these tests are a bit weird and wrong. On the other hand should we report this upstream to see if they actually meant to change these things?

PR #4284, opened 2020-06-13
Drop support for Python 3.6

Cadair:
This is the master branch version of #4283 seeing how it was "broken" anyway, we might as well drop it rather than fix it.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -74,7 +74,7 @@ jobs:
       - linux: py38-hypothesis
 
       - linux: py38-conda
-        name: py368_conda
+        name: py38_conda
         libraries: {}
 
       - linux: py37-oldestdeps
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -55,8 +55,8 @@ jobs:
         - freetype-devel
         - libpng-devel
     envs:
-      - macos: py36
-        name: py36_test
+      - macos: py38
+        name: py38_test
         posargs: -n=1
 
       - windows: py37
@@ -73,12 +73,12 @@ jobs:
 
       - linux: py38-hypothesis
 
-      - linux: py36-conda
-        name: py36_conda
+      - linux: py38-conda
+        name: py368_conda
         libraries: {}
 
-      - linux: py36-oldestdeps
-        name: py36_test_oldestdeps
+      - linux: py37-oldestdeps
+        name: py37_test_oldestdeps
 
       - linux: py38-devdeps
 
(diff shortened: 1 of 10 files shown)


azure-pipelines.yml
```diff
@@ -73,12 +73,12 @@ jobs:
 
       - linux: py38-hypothesis
 
-      - linux: py36-conda
-        name: py36_conda
+      - linux: py38-conda
+        name: py368_conda
```
nabobalis, line 77:
```suggestion
        name: py38_conda
```

nabobalis:
Needs a changelog.

azure-pipelines.yml
```diff
@@ -97,9 +97,9 @@ jobs:
...
       dependsOn:
-        - py36_test
         - py37_test
+        - py38_test
         - py38_test_online
         - py38_test_32bit
-        - py36_test_oldestdeps
-        - py36_conda
+        - py38_test_oldestdeps
+        - py38_conda
```
dstansby, line 105: Is the whole test suite run on `py38_conda`? If so it seems a bit of a waste given there's already a set of tests run on `py38_test`.

sunpy/__init__.py
```diff
@@ -16,26 +16,13 @@
...
 except ImportError:
     print("version.py not found, please reinstall sunpy.")
     __version__ = "unknown"
 
 
-class UnsupportedPythonError(Exception):
-    pass
-
-
-if sys.version_info < tuple(int(val) for val in __minimum_python_version__.split('.')):
```
dstansby, line 33: Is there still a check for python version somewhere that gives a nice error?

azure-pipelines.yml
Cadair: yeah, it needs to run the whole test suite. We could, but I am not sure I want to swap the conda build to macos to remove the duplication but I would rather keep them separate.

sunpy/__init__.py
Cadair: I suppose not at runtime, I will put this back.

Cadair:
Now also bumping our other core versions. Matplotlib, scipy and pandas are now at the major version released two years before our scheduled 2.1 release (so ~Nov/Dec). Astropy gets its regular bump, so we are now at 4.0. Numpy is bumped in line with NEP 29, but also the minimum version in astropy 4.0.

azure-pipelines.yml
nabobalis, line 77:
```suggestion
        name: py38_conda
```

PR #4289, opened 2020-06-15
Don't default to py_36 for toxenv in figure comparisons

dstansby:


sunpy/tests/helpers.py
```diff
@@ -181,7 +182,9 @@ def _patch_coverage(testdir, sourcedir):  # pragma: no cover
 def _generate_fig_html(fname):
     generated_image = figure_base_dir / (fname + '.png')
 
-    envname = os.environ.get("TOXENV", "figure_py36")
+    envname = os.environ.get("TOXENV", None)
+    if envname is None:
+        raise RuntimeError("Could not find a TOXENV environment variable")
```
Cadair, line 187: This is likely to happen locally, we should probably just print something rather than error. Of course we should really have this as a arg to pytest or something.

dstansby: The thinking was checking this was a way to verify the right packages were installed. I feel like this should definitely error on CI, but maybe just warn elsewhere - does that sound sensible?

Cadair: Oh I didn't realise tox injects this env var, that's super awesome! fine as is.

Cadair:
I squash merged to aid in the presumably hellish backport to 2.0 when we work all this out.

nabobalis:
Not sure that matters for the bot.

Issue #4290, opened 2020-06-15
Intermittent image-rotation test failures on OS X when using conda


ayshih:
We have intermittent failures of our image rotation tests and these failures appear to be isolated to OS X when using conda(-forge).  #4235 added raw output for these failures occurred, and some of the output is truly bizarre.  I will add investigative stuff in separate posts.  My current conjecture is there is nothing wrong with SunPy code, but rather a C extension in the `numpy`/`scipy`/`scikit-image` ecosystem is not being compiled for conda(-forge) with the correct compile options for OS X such that there's the intermittent potential for bad memory access of arrays.

---

Edit: go down to https://github.com/sunpy/sunpy/issues/4290#issuecomment-676573472 for a summary of the current understanding

ayshih:
One example failure:
```
___________________________ test_rotation[-270.0-1] ____________________________

original = array([[156., 157., 160., ..., 152., 152., 152.],
       [156., 157., 159., ..., 152., 152., 152.],
       [158., 157....121., 113., 111.],
       [121., 123., 126., ..., 121., 113., 111.],
       [121., 123., 126., ..., 121., 113., 111.]])
angle = -4.71238898038469, k = 1

    @pytest.mark.parametrize("angle, k", [(90.0, 1), (-90.0, -1), (-270.0, 1),
                                          (-90.0, 3), (360.0, 0), (-360.0, 0)])
    def test_rotation(original, angle, k):
        # Test rotation against expected outcome
        angle = np.radians(angle)
        c = np.round(np.cos(angle))
        s = np.round(np.sin(angle))
        rmatrix = np.array([[c, -s], [s, c]])
        expected = np.rot90(original, k=k)
    
        # Run the tests at order 4 as it produces more accurate 90 deg rotations
        rot = affine_transform(original, order=4, rmatrix=rmatrix)
>       assert compare_results(expected, rot)
E       assert False
[...]

PR #4293, opened 2020-06-16
Ongoing investigation into test failures on OS X when using conda-forge

Cadair:
See #4290


Cadair:
@ayshih I merged #4284 and #4295 so I rebased this.

PR #4295, opened 2020-06-17
Do not install any dependencies via pip under tox conda

### LINKED ISSUE #4285: Our tox-conda build does a poor job of picking up conda build issues as it uses a mix of conda and pip to install the dependencies.
The only solution I can think of is to enable `--no-deps` in the pip install command for tox on that build and make sure that *all* possible dependencies are in the `conda_deps` list in the tox.ini file

ayshih:
For the tox conda build:
- The pip command will no longer install dependencies
- List out all of the installed packages

Closes #4285

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -108,6 +108,7 @@ deps =
 conda_deps =
     astropy
     beautifulsoup4
+    conda
     dask
     drms
     glymur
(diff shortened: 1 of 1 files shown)


ayshih:
I didn't change the Python version of the tox conda build from 3.6, but the CI build can't succeed because of the incompatibility of Python 3.6 and parfive 1.1.0.  This nicely shows that the `--no-deps` is working as intended!

The Python version of the tox conda build will be bumped to 3.8 in #4284.

PR #4298, opened 2020-06-17
Fixed missing asdf dependency for tox conda build

ayshih:
Follow-on to #4295 

I'm blind.  It didn't register that the build was complaining about `asdf` being missing.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -133,7 +133,7 @@ conda_deps =
     sqlalchemy
     zeep
     pillow < 7.1.0
-conda_channels = sunpy
+conda_channels = conda-forge
 install_command = pip install --no-deps {opts} {packages}
 commands =
     conda list
--- a/tox.ini
+++ b/tox.ini
@@ -108,6 +108,7 @@ basepython = python3.8
 extras =
 deps =
 conda_deps =
+    asdf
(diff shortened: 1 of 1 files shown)


ayshih:
Yeesh!  Looks like there's much more wrong...

ayshih:
Okay, it looks like adding the `asdf` dependency is responsible for completely breaking the conda solution.  A conda solution can be found when using the `conda-forge` channel instead of the `sunpy` channel, so the `sunpy` channel must be missing one or more critical packages.  Either the tox conda build needs to simply omit `asdf` (the `asdf` tests are then skipped), the `sunpy` channel needs to be fixed, or the channel needs to be changed to `conda-forge`.

Cadair:
>  the channel needs to be changed to conda-forge

I vote for this one.

nabobalis:
> > the channel needs to be changed to conda-forge
> 
> I vote for this one.

I do too. Since this is also the channel used for our conda recipe.

ayshih:
@Cadair Heh, oops, I didn't see that you added a commit here.  I just created a fresh PR (#4299) which also covers some more missing dependencies, so I'm going to close this PR.

PR #4299, opened 2020-06-18
Switch to using conda-forge for tox conda build

ayshih:
Follow-on to #4295

- For the tox conda build, switch to using the `conda-forge` channel instead of the (incomplete) `sunpy` channel
- Add missing conda dependencies


nabobalis:
Conda build passes.

Issue #4301, opened 2020-06-19
Mapsequence.plot docstring lacks information on plot_function signature


Cadair:
it should either link to the animator docs or copy the function signature information

PR #4306, opened 2020-06-21
Fix MapSequence doc

dstansby:
See https://docs.sunpy.org/en/stable/api/sunpy.map.MapSequence.html#mapsequence for the current slightly-broken version.


Cadair:
While you are here I don't suppose you want to fix #4301 ?

dstansby:
Not really...

PR #4308, opened 2020-06-21
Improve missing observer transformation errors

dstansby:
- Add a description of what is wrong in the error message (`needs to be set` instead of just saying the coordinate is `None`)
- Fix the observer printed in the second error message (it was `obs_2`, but I think it is meant to be `obs_1`)

### DIFF
--- a/sunpy/coordinates/transformations.py
+++ b/sunpy/coordinates/transformations.py
@@ -197,11 +197,13 @@ def _observers_are_equal(obs_1, obs_2):
 
     # obs_1 != obs_2
     if obs_1 is None:
-        raise ConvertError("The source observer is set to None, but the destination observer is "
-                           f"{obs_2}.")
+        raise ConvertError("The source observer is set to None, but the transformation requires "
+                           "the source observer to be specified, as the destination observer "
+                           f"is set to {obs_2}.")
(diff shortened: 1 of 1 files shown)


sunpy/coordinates/transformations.py
```diff
@@ -197,11 +197,11 @@ def _observers_are_equal(obs_1, obs_2):
 
     # obs_1 != obs_2
     if obs_1 is None:
-        raise ConvertError("The source observer is set to None, but the destination observer is "
-                           f"{obs_2}.")
+        raise ConvertError("The source observer is None, but needs to be set as "
+                           f"the destination observer is {obs_2}.")
```
Cadair, line 201: My concern with this wording is it could be read as "you need to set the source observer to {obs_2}". Also I think "is set to None" is better, but it's very possible the user hasn't actually *set* the observer at all, i.e. if they are transforming to a frame class etc.

dstansby: How about:

"The source observer is set to None, but the transformation requires the source observer to be specified, as the destination observer is set to {obs_2}."?

Cadair: sounds good.

PR #4213, opened 2020-05-28
Dataretriever QueryResponse can show more metadata columns

abhijeetmanhas:
### Description
* `show()` method in `base_client()` to specify which columns to be shown in response.
* Dataretriever clients `QueryResponse` have more metadata info like goes `SatelliteNumber` , `Detector`, `Level`, etc. got from url.
* Correct satellite number returned in goes overlapping urls.

### New metadata cols added for
 * [x] Eve
 * [x] Fermi
 * [x] Lyra
 * [x] NorH
 * [x] XRS
 * [x] SUVI
 * [ ] Rhessi
 * [ ] Noaa
### Updated Tests
 * [x] Docs
 * [ ] Database
 * [x] net
 * [x] util
### Fix `parse` dependency
 * [x] Pypi
 * [x] Conda
### Miscellaneous
 * [ ] What to do with `EndTime` for clients?
 * [ ] Add `docs` for new functionality
 * [ ] Add tests for new metadata cols
 * [ ] Discuss better approaches for returning metadata from scrapper

### Some snippets
```python
from sunpy.net import Fido, attrs as a
res=Fido.search(a.Time('2019/05/25 00:50', '2019/05/25 00:52'),a.Instrument('suvi'))
res

<sunpy.net.fido_factory.UnifiedResponse object at 0x7f6c418c2cd0>
Results from 1 Provider:

6 Results from the SUVIClient:
Level SatelliteNumber        StartTime                EndTime           Wavelength   Source Provider Physobs Instrument
----- --------------- ----------------------- ----------------------- -------------- ------ -------- ------- ----------
    2              16 2019-05-25T00:52:00.000 2019-05-25T00:56:00.000  94.0 Angstrom   GOES     NOAA    flux       SUVI
    2              16 2019-05-25T00:52:00.000 2019-05-25T00:56:00.000 131.0 Angstrom   GOES     NOAA    flux       SUVI
    2              16 2019-05-25T00:52:00.000 2019-05-25T00:56:00.000 171.0 Angstrom   GOES     NOAA    flux       SUVI
    2              16 2019-05-25T00:52:00.000 2019-05-25T00:56:00.000 195.0 Angstrom   GOES     NOAA    flux       SUVI
    2              16 2019-05-25T00:52:00.000 2019-05-25T00:56:00.000 284.0 Angstrom   GOES     NOAA    flux       SUVI
[...]

PR #4313, opened 2020-06-23
Added method to extract metadata from url using parse

abhijeetmanhas:
### Description

It is a split of #4213 in which a new method is added for scraper changes which would possibly be backward compatible.

### DIFF
--- a/sunpy/util/scraper.py
+++ b/sunpy/util/scraper.py
@@ -333,7 +333,7 @@ class Scraper:
         except Exception:
             raise
 
-    def _extractMetaURLs(self, timerange, extractor=None, translator=None):
+    def _extract_files_meta(self, timerange, extractor=None):
         """
         Returns metadata information contained in URLs.
 
@@ -343,8 +343,6 @@ class Scraper:
             Time interval where to find the directories for a given pattern.
         extractor: `str`
             Pattern to extract metadata by parsing the URL.
-        translator: `dict`
-            Convert metadata retrieved from URL to table displayable format.
 
         Returns
         -------
@@ -362,12 +360,5 @@ class Scraper:
             else:
                 metadict = {}
             metadict['url'] = url
-            if translator is None:
-                metalist.append(metadict)
-            else:
-                for k in translator.keys():
-                    inurl = metadict[k]
-                    intable = translator[k][inurl]
-                    metadict[k] = intable
-                    metalist.append(metadict)
+            metalist.append(metadict)
         return metalist
--- a/sunpy/util/scraper.py
+++ b/sunpy/util/scraper.py
@@ -15,6 +15,7 @@ from bs4 import BeautifulSoup
 import astropy.units as u
 from astropy.time import Time, TimeDelta
 
+from sunpy.extern.parse import parse
 from sunpy.util.exceptions import SunpyUserWarning
 
 __all__ = ['Scraper']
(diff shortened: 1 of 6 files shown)


abhijeetmanhas:
Two tests are failing which are in extern parse code.
https://dev.azure.com/sunpy/sunpy/_build/results?buildId=8518&view=logs&j=90fb3984-273f-5c22-0ea1-da9b62a2f74c&t=0369260d-660a-5340-720a-a028839aef79

nabobalis:
Need to exclude the entire extern location from pytest in setup.cfg. I would have thought it would have been.

sunpy/util/scraper.py
```diff
@@ -331,3 +332,42 @@ def _smallerPattern(self, directoryPattern):
                 return None
         except Exception:
             raise
+
+    def _extractMetaURLs(self, timerange, extractor=None, translator=None):
```
Cadair, line 336: This function name isn't great, python method names are supposed to use snake_case.

I am not sure what a good name for this function is, it's doing a few things (calling filelist as well as running things through parse). Do we want to see this become the main entry point for scraper in a land where you use parse for everything?

sunpy/util/scraper.py
```diff
@@ -331,3 +332,42 @@ def _smallerPattern(self, directoryPattern):
...
+        """
+        Returns metadata information contained in URLs.
+
+        Parameters
+        ----------
+        timerange : `~sunpy.time.TimeRange`
+            Time interval where to find the directories for a given pattern.
+        extractor: `str`
+            Pattern to extract metadata by parsing the URL.
+        translator: `dict`
```
Cadair, line 346: What are you using the translator for out of interest?

sunpy/util/scraper.py
abhijeetmanhas: Yes, `filelist` will become internal function for scraper and the client would only be using the `extractMetaURLs` method.

sunpy/util/scraper.py
abhijeetmanhas: I'm using it for converting the things in url to what we want to show them in the table.
As for example.
In `NorHClient`, if url contains `tca` then it translates to `17*u.GHz` in the response table. It is optional param, not always required for most of the clients.

Cadair: On one hand this feels like something which should be done after the call to this method and not as a part of it. What do you think, is this sufficiently useful to warrant it adding extra complexity here?

abhijeetmanhas: Agreed. We it should be a `GenericClient` method, not a part of scraper method. Since as a function, it should only return the metadata from url string and nothing else. The final dict can be formed later by using this extracted info and other fixed client attr values.

Cadair:
if you pull in master the figure test build should pass.

nabobalis:
Can you rebase this onto master?

nabobalis:
Windows failures unrelated.

```
FAILED ..\..\.tox\py37\lib\site-packages\sunpy\coordinates\tests\test_metaframes.py::test_rotatedsun_transforms[HeliocentricInertial] - hypothesis.errors.FailedHealthCheck: Data...
FAILED ..\..\.tox\py37\lib\site-packages\sunpy\coordinates\tests\test_metaframes.py::test_rotatedsun_transforms[HeliographicCarrington] - hypothesis.errors.DeadlineExceeded: Tes...
```

nabobalis:
Doc failure is due to 
```Files Downloaded: 100%|██████████| 1/1 [00:02<00:00,  2.91s/file]INFO: Failed to download one or more sample data files, retrying with a mirror. [sunpy.data._sample]
DEBUG: Failed to download SWAP_LEVEL1_IMAGE from http://data.sunpy.org/sunpy/v1/swap_lv1_20110607_063329.fits: Timeout on reading data from socket [sunpy.data._sample]
loading translations [en]... done
``` 
which is unrelated to the PR.

PR #4321, opened 2020-06-26
Redesigning genericClient

### LINKED ISSUE #2314: Wavelength attribute and namedtuple should have the same way to refer to their things
In dataretriever a [wavelength namedtuple](https://github.com/sunpy/sunpy/blob/master/sunpy/net/dataretriever/client.py#L202) is created with `wavemin` and `wavemax` names. However, a [`Wavelength` attribute](https://github.com/sunpy/sunpy/blob/master/sunpy/net/vso/attrs.py#L132) has `min` and `max` to access the min and max value of that range.

This is at times confusing, because fido will change for being a Wavelength to being a named tuple and then we have to use `min` or `wavemin` for different inputs.

### LINKED ISSUE #3306: GenericClient could provide a much nicer interface for _can_handle_query
Currently `_can_handle_query` takes the whole query as an argument and returns `True` or `False` as a way of determining if the client should be used for the given query.

With some work we could have a helper method on `dataretriever.GenericClient` which returns a list of attrs with which it should match. i.e. `a.Time(start, end), a.Instrument("AIA"), a.Wavelength(start, end)`.

While this would not be as flexible (it can only match the intersection of all provided) it would cover many of the usecases we have in dataretriever at the moment.

### LINKED ISSUE #3321: Dataretriever Clients should be able to specify what columns are in their results table
For instance the GOES based satellites probably want satellite number in the table.

This is actually more complex because there isn't a method currently for subclasses of `GenericClient` to return a list of results with meta data for each file.

### LINKED ISSUE #3337: Data for GOES assumes it knows what satellites are available based on date range

### LINKED ISSUE #3715: `GenericClient.search` can't know URLs' timeranges without additional function
### Description
While implementing a Client in `sunpy/net/dataretriever/sources`, which is subclass of `GenericClient`; the `GenericClient.search` can't use `_get_url_for_timerange` function, to know timeranges for urls returned by it.
 An additional ` _get_time_for_url` function is required to be implemented so that search can know the time range for each individual URL.
Only one function should be able to handle urls and timeranges associated with it, rather than it requiring to implement two functions for all source Clients.

### Additional context
If this is expected output

```
>>> from sunpy.net import Fido
>>> from sunpy.net import attrs as a
>>> results = Fido.search(a.Time('2016/5/18 00:00:00', '2016/5/20 00:03:00'), a.Instrument('swepam'))
>>> print(results)
         Start Time           End Time      Source Instrument
            str19               str19         str3     str6
    ------------------- ------------------- ------ ----------
    2016-05-18 00:00:00 2016-05-19 00:00:00    ACE     swepam
    2016-05-19 00:00:00 2016-05-20 00:00:00    ACE     swepam
    2016-05-20 00:00:00 2016-05-21 00:00:00    ACE     swepam
```

But without implementing `get_time_for_url` , we get same start and end times.
```

         Start Time           End Time      Source Instrument 
           str19               str19         str3     str6       
    2016-05-18 00:00:00 2016-05-20 00:03:00    ACE     swepam        
    2016-05-18 00:00:00 2016-05-20 00:03:00    ACE     swepam        
    2016-05-18 00:00:00 2016-05-20 00:03:00    ACE     swepam        
```

abhijeetmanhas:
### Description

Key highlights:

- Easy to define a new "simple" Fido client. Only define `pattern`, `baseurl`, and `register_values()`. This further uses scraper and `~sunpy.extern.parse` for extracting metadata.
- More metadata is shown in the response tables. Like `SatelliteNumber`, `Level`, `Detector`, `Resolution`,  `Wavelength` will be shown which earlier were either not present or `np.nan`.
- Nothing is default Attr for any search made to client. So if client supports 3 levels, and we don't specify level in the query, data fo all levels shall be shown.
- Refactoring in `GenericClient`, removed/generalized/updated a lot of methods.

Fixes #3715, Fixes #3337, Fixes  #3321, Fixes #3306, Fixes #2314.

### DIFF
--- a/changelog/4321.breaking.1.rst
+++ b/changelog/4321.breaking.1.rst
@@ -1,0 +1,1 @@
+Removed :meth:`~sunpy.net.dataretriever.QueryResponse.create`.
--- a/changelog/4321.breaking.rst
+++ b/changelog/4321.breaking.rst
@@ -1,0 +1,1 @@
+Removed `~sunpy.net.dataretriever.QueryResponseBlock` as `~sunpy.net.dataretriever.QueryResponse` now uses a `list` of `~collections.OrderedDict` for building the response table.
--- a/changelog/4321.feature.1.rst
+++ b/changelog/4321.feature.1.rst
@@ -1,0 +1,3 @@
+Additions in `~sunpy.util.scraper` to support the refactoring of `~sunpy.net.dataretriever.GenericClient`:
+- :meth:`~sunpy.util.scraper.Scraper.findDatewith_extractor` that parses the url using extractor to return its start time.
+- A ``matcher`` in :meth:`~sunpy.util.scraper.Scraper._extract_files_meta` which validates the extracted metadata by using the dictionary returned from :meth:`~sunpy.net.dataretriever.GenericClient._get_match_dict`.
--- a/changelog/4321.feature.rst
+++ b/changelog/4321.feature.rst
@@ -1,0 +1,4 @@
+Refactoring of `~sunpy.net.dataretriever` which adds these capabilities to `~sunpy.net.dataretriever.QueryResponse`:
+
+- Any ``attr`` shall not be defaulted to a hard-coded value in all subclasses of `~sunpy.net.dataretriever.GenericClient`; thus records for all possible ``attrs`` shall be returned if it is not specified in the query.
+- `~sunpy.net.dataretriever.QueryResponse` can now show more columns; thus all metadata extractable from matching file URLs shall be shown and for a client, non-spported ``attrs`` shall not be shown in the response tables.
--- a/docs/guide/acquiring_data/fido.rst
+++ b/docs/guide/acquiring_data/fido.rst
@@ -62,19 +62,6 @@ Using ``Instrument`` as the first example, if you print the object::
 You get a full list of known values, a description and what "Clients" support those values (if you want to use a specific data source).
 This is supported for most attributes including the client specific ones.
 
-For VSO::
-
-    >>> print(a.vso.Provider)
-    sunpy.net.vso.attrs.Provider
-    <BLANKLINE>
-    Specifies the VSO data provider to search for data for.
-    <BLANKLINE>
-    Attribute Name Client Full Name                                   Description
-    -------------- ------ --------- --------------------------------------------------------------------------------
-    hao            VSO    HAO       High Altitude Observatory, NCAR
-    jsoc           VSO    JSOC      SDO Joint Science Operations Center
-    kis            VSO    KIS       Kiepenheuer-Institut für Sonnenphysik
-    ...
 
 For JSOC::
 
@@ -98,7 +85,7 @@ Searching for Data Using Fido
 
 For example::
 
-    >>> result = Fido.search(a.Time('2012/3/4', '2012/3/6'), a.Instrument.lyra) # doctest: +REMOTE_DATA
+    >>> result = Fido.search(a.Time('2012/3/4', '2012/3/6'), a.Instrument.lyra, a.Level.two) # doctest: +REMOTE_DATA
 
 this returns an `~sunpy.net.fido_factory.UnifiedResponse` object containing
 information on the available online files which fit the criteria specified by
@@ -113,11 +100,11 @@ variable set to the Fido search, in this case, result::
     Results from 1 Provider:
     <BLANKLINE>
     3 Results from the LYRAClient:
-         Start Time           End Time      Source Instrument Wavelength
-    ------------------- ------------------- ------ ---------- ----------
-    2012-03-04 00:00:00 2012-03-06 00:00:00 Proba2       lyra        nan
-    2012-03-04 00:00:00 2012-03-06 00:00:00 Proba2       lyra        nan
-    2012-03-04 00:00:00 2012-03-06 00:00:00 Proba2       lyra        nan
+         Start Time           End Time      Instrument ... Source Provider Level
+    ------------------- ------------------- ---------- ... ------ -------- -----
+    2012-03-04 00:00:00 2012-03-04 23:59:59       LYRA ... PROBA2      ESA     2
+    2012-03-05 00:00:00 2012-03-05 23:59:59       LYRA ... PROBA2      ESA     2
+    2012-03-06 00:00:00 2012-03-06 23:59:59       LYRA ... PROBA2      ESA     2
     <BLANKLINE>
     <BLANKLINE>
 
@@ -133,11 +120,11 @@ passbands can be searched for by supplying an `~astropy.units.Quantity` to the
     Results from 1 Provider:
     <BLANKLINE>
     3 Results from the NoRHClient:
-         Start Time           End Time      Source Instrument   Wavelength
-    ------------------- ------------------- ------ ---------- --------------
-    2012-03-04 00:00:00 2012-03-05 00:00:00   NAOJ       NORH 17000000.0 kHz
-    2012-03-05 00:00:00 2012-03-06 00:00:00   NAOJ       NORH 17000000.0 kHz
-    2012-03-06 00:00:00 2012-03-07 00:00:00   NAOJ       NORH 17000000.0 kHz
+         Start Time           End Time      Instrument Source Provider Wavelength
+    ------------------- ------------------- ---------- ------ -------- ----------
+    2012-03-04 00:00:00 2012-03-04 23:59:59       NORH   NAOJ      NRO   17.0 GHz
+    2012-03-05 00:00:00 2012-03-05 23:59:59       NORH   NAOJ      NRO   17.0 GHz
+    2012-03-06 00:00:00 2012-03-06 23:59:59       NORH   NAOJ      NRO   17.0 GHz
     <BLANKLINE>
     <BLANKLINE>
 
@@ -193,10 +180,11 @@ operator would::
     <sunpy.net.fido_factory.UnifiedResponse object at ...>
     Results from 3 Providers:
     <BLANKLINE>
-    1 Results from the LYRAClient:
-         Start Time           End Time      Source Instrument Wavelength
-    ------------------- ------------------- ------ ---------- ----------
-    2012-03-04 00:00:00 2012-03-04 02:00:00 Proba2       lyra        nan
+    2 Results from the LYRAClient:
+         Start Time           End Time      Instrument ... Source Provider Level
+    ------------------- ------------------- ---------- ... ------ -------- -----
+    2012-03-04 00:00:00 2012-03-04 23:59:59       LYRA ... PROBA2      ESA     2
+    2012-03-04 00:00:00 2012-03-04 23:59:59       LYRA ... PROBA2      ESA     3
     <BLANKLINE>
     3 Results from the VSOClient:
        Start Time [1]       End Time [1]    Source ...     Type    Wavelength [2]
@@ -207,9 +195,9 @@ operator would::
     2012-03-04 01:45:40 2012-03-04 02:09:00 RHESSI ... PARTIAL_SUN 3.0 .. 17000.0
     <BLANKLINE>
     1 Results from the RHESSIClient:
-         Start Time           End Time      Source Instrument Wavelength
-    ------------------- ------------------- ------ ---------- ----------
-    2012-03-04 00:00:00 2012-03-04 23:59:59 rhessi     rhessi        nan
+         Start Time           End Time      Instrument ... Source Provider
+    ------------------- ------------------- ---------- ... ------ --------
+    2012-03-04 00:00:00 2012-03-04 23:59:59     RHESSI ... RHESSI     NASA
     <BLANKLINE>
     <BLANKLINE>
 
@@ -227,7 +215,7 @@ the `~sunpy.net.dataretriever.LYRAClient`, and EVE data from the
 `~sunpy.net.vso.VSOClient`::
 
     >>> from sunpy.net import Fido, attrs as a
-    >>> results = Fido.search(a.Time("2012/1/1", "2012/1/2"),
+    >>> results = Fido.search(a.Time("2012/1/1", "2012/1/2"), a.Level.two,
     ...                       a.Instrument.lyra | a.Instrument.eve)  # doctest: +REMOTE_DATA
 
 If you then wanted to inspect just the LYRA data for the whole time range
@@ -239,10 +227,10 @@ results returned by the `~sunpy.net.dataretriever.LYRAClient`::
     Results from 1 Provider:
     <BLANKLINE>
     2 Results from the LYRAClient:
-         Start Time           End Time      Source Instrument Wavelength
-    ------------------- ------------------- ------ ---------- ----------
-    2012-01-01 00:00:00 2012-01-02 00:00:00 Proba2       lyra        nan
-    2012-01-01 00:00:00 2012-01-02 00:00:00 Proba2       lyra        nan
+         Start Time           End Time      Instrument ... Source Provider Level
+    ------------------- ------------------- ---------- ... ------ -------- -----
+    2012-01-01 00:00:00 2012-01-01 23:59:59       LYRA ... PROBA2      ESA     2
+    2012-01-02 00:00:00 2012-02-02 23:59:59       LYRA ... PROBA2      ESA     2
     <BLANKLINE>
     <BLANKLINE>
 
@@ -253,10 +241,10 @@ Or, equivalently::
     Results from 1 Provider:
     <BLANKLINE>
     2 Results from the LYRAClient:
-         Start Time           End Time      Source Instrument Wavelength
-    ------------------- ------------------- ------ ---------- ----------
-    2012-01-01 00:00:00 2012-01-02 00:00:00 Proba2       lyra        nan
-    2012-01-01 00:00:00 2012-01-02 00:00:00 Proba2       lyra        nan
+         Start Time           End Time      Instrument ... Source Provider Level
+    ------------------- ------------------- ---------- ... ------ -------- -----
+    2012-01-01 00:00:00 2012-01-01 23:59:59       LYRA ... PROBA2      ESA     2
+    2012-01-02 00:00:00 2012-02-02 23:59:59       LYRA ... PROBA2      ESA     2
     <BLANKLINE>
     <BLANKLINE>
 
--- a/examples/time_series/goes_hek_m25.py
+++ b/examples/time_series/goes_hek_m25.py
@@ -15,7 +15,7 @@ from sunpy.timeseries import TimeSeries
 
 ###############################################################################
 # Let's first grab GOES XRS data for a particular time of interest
-tr = TimeRange(['2011-06-07 04:00', '2011-06-07 12:00'])
+tr = TimeRange(['2011-06-07 00:00', '2011-06-07 12:00'])
 results = Fido.search(a.Time(tr), a.Instrument.xrs)
 
 ###############################################################################
--- a/examples/time_series/goes_hek_m25.py
+++ b/examples/time_series/goes_hek_m25.py
@@ -15,7 +15,7 @@ from sunpy.timeseries import TimeSeries
 
 ###############################################################################
 # Let's first grab GOES XRS data for a particular time of interest
-tr = TimeRange(['2011-06-07 00:00', '2011-06-07 12:00'])
+tr = TimeRange(['2011-06-07 04:00', '2011-06-07 12:00'])
 results = Fido.search(a.Time(tr), a.Instrument.xrs)
 
 ###############################################################################
--- a/sunpy/database/tests/test_database.py
+++ b/sunpy/database/tests/test_database.py
@@ -74,12 +74,11 @@ def fido_search_result():
     # No JSOC query
     return Fido.search(
         net_attrs.Time("2012/1/1", "2012/1/2"),
-        net_attrs.Instrument('lyra') | net_attrs.Instrument('eve') |
+        net_attrs.Instrument('lyra') & net_attrs.Level.two | net_attrs.Instrument('eve') |
         net_attrs.Instrument('XRS') | net_attrs.Instrument('noaa-indices') |
         net_attrs.Instrument('noaa-predict') |
         (net_attrs.Instrument('norh') & net_attrs.Wavelength(17*units.GHz)) |
-        (net_attrs.Instrument('rhessi') & net_attrs.Physobs("summary_lightcurve")) |
-        (net_attrs.Instrument('EVE') & net_attrs.Level(0))
+        (net_attrs.Instrument('rhessi') & net_attrs.Physobs("summary_lightcurve"))
     )
 
 
@@ -1107,8 +1106,8 @@ def test_split_database(split_function_database, database):
         split_function_database, database, net_attrs.Instrument('EIA'))
 
     observed_source_entries = split_function_database.search(
-        vso.attrs.Provider('xyz'), sortby='id')
-    observed_destination_entries = database.search(vso.attrs.Provider('xyz'))
+        net_attrs.Provider('xyz'), sortby='id')
+    observed_destination_entries = database.search(net_attrs.Provider('xyz'))
 
     assert observed_source_entries == [
         DatabaseEntry(id=1, instrument='RHESSI', provider='xyz'),
--- a/sunpy/database/tests/test_tables.py
+++ b/sunpy/database/tests/test_tables.py
@@ -55,12 +55,11 @@ def fido_search_result():
     # No JSOC query
     return Fido.search(
         net_attrs.Time("2012/1/1", "2012/1/2"),
-        net_attrs.Instrument('lyra') | net_attrs.Instrument('eve') |
+        net_attrs.Instrument('lyra') & net_attrs.Level.two | net_attrs.Instrument('eve') |
         net_attrs.Instrument('XRS') | net_attrs.Instrument('noaa-indices') |
         net_attrs.Instrument('noaa-predict') |
         (net_attrs.Instrument('norh') & net_attrs.Wavelength(17 * u.GHz)) |
-        (net_attrs.Instrument('rhessi') & net_attrs.Physobs("summary_lightcurve")) |
-        (net_attrs.Instrument('EVE') & net_attrs.Level(0))
+        (net_attrs.Instrument('rhessi') & net_attrs.Physobs("summary_lightcurve"))
     )
 
 
@@ -74,15 +73,15 @@ def query_result():
 def qr_with_none_waves():
     return vso.VSOClient().search(
         net_attrs.Time('20121224T120049.8', '20121224T120049.8'),
-        vso.attrs.Provider('SDAC'), net_attrs.Instrument('VIRGO'))
+        net_attrs.Provider('SDAC'), net_attrs.Instrument('VIRGO'))
 
 
 @pytest.fixture
 def qr_block_with_missing_physobs():
     return vso.VSOClient().search(
         net_attrs.Time('20130805T120000', '20130805T121000'),
-        net_attrs.Instrument('SWAVES'), vso.attrs.Source('STEREO_A'),
-        vso.attrs.Provider('SSC'), net_attrs.Wavelength(10 * u.kHz, 160 * u.kHz))[0]
+        net_attrs.Instrument('SWAVES'), net_attrs.Source('STEREO_A'),
+        net_attrs.Provider('SSC'), net_attrs.Wavelength(10 * u.kHz, 160 * u.kHz))[0]
 
 
 @pytest.fixture
@@ -127,14 +126,23 @@ def test_entries_from_fido_search_result(fido_search_result):
     assert len(entries) == 66
     # First 2 entries are from lyra
     assert entries[0] == DatabaseEntry(
-        source='Proba2', provider='esa', physobs='irradiance',
+        source='PROBA2', provider='ESA', physobs='irradiance',
         fileid='http://proba2.oma.be/lyra/data/bsd/2012/01/01/lyra_20120101-000000_lev2_std.fits',
         observation_time_start=datetime(2012, 1, 1, 0, 0),
-        observation_time_end=datetime(2012, 1, 2, 0, 0),
+        observation_time_end=datetime(2012, 1, 1, 23, 59, 59, 999000),
         wavemin=np.nan, wavemax=np.nan,
-        instrument='lyra')
-    # 54 entries from EVE
+        instrument='LYRA')
+    # 2 entries from eve, level 0
     assert entries[2] == DatabaseEntry(
+        source='SDO', provider='LASP', physobs='irradiance',
+        fileid=("http://lasp.colorado.edu/eve/data_access/evewebdata/quicklook"
+                "/L0CS/SpWx/2012/20120101_EVE_L0CS_DIODES_1m.txt"),
+        observation_time_start=datetime(2012, 1, 1, 0, 0),
+        observation_time_end=datetime(2012, 1, 1, 23, 59, 59, 999000),
+        wavemin=np.nan, wavemax=np.nan,
+        instrument='EVE')
+    # 54 entries from EVE
+    assert entries[4] == DatabaseEntry(
         source='SDO', provider='LASP', physobs='irradiance',
         fileid='EVE_L1_esp_2012001_00',
         observation_time_start=datetime(2012, 1, 1, 0, 0),
@@ -143,56 +151,47 @@ def test_entries_from_fido_search_result(fido_search_result):
         instrument='EVE',
         wavemin=0.1, wavemax=30.4)
     # 2 entries from goes
-    assert entries[56] == DatabaseEntry(
-        source='nasa', provider='sdac', physobs='irradiance',
+    assert entries[58] == DatabaseEntry(
+        source='NASA', provider='SDAC', physobs='irradiance',
         fileid='https://umbra.nascom.nasa.gov/goes/fits/2012/go1520120101.fits',
         observation_time_start=datetime(2012, 1, 1, 0, 0),
         observation_time_end=datetime(2012, 1, 1, 23, 59, 59, 999000),
         wavemin=np.nan, wavemax=np.nan,
-        instrument='goes')
+        instrument='XRS')
     # 1 entry from noaa-indices
-    assert entries[58] == DatabaseEntry(
-        source='sdic', provider='swpc', physobs='sunspot number',
+    assert entries[60] == DatabaseEntry(
+        source='SIDC', provider='SWPC', physobs='sunspot number',
         fileid='https://services.swpc.noaa.gov/json/solar-cycle/observed-solar-cycle-indices.json',
         observation_time_start=datetime(2012, 1, 1, 0, 0),
         observation_time_end=datetime(2012, 1, 2, 0, 0),
         wavemin=np.nan, wavemax=np.nan,
-        instrument='noaa-indices')
+        instrument='NOAA-Indices')
     # 1 entry from noaa-predict
-    assert entries[59] == DatabaseEntry(
-        source='ises', provider='swpc', physobs='sunspot number',
+    assert entries[61] == DatabaseEntry(
+        source='ISES', provider='SWPC', physobs='sunspot number',
         fileid='https://services.swpc.noaa.gov/json/solar-cycle/predicted-solar-cycle.json',
         observation_time_start=datetime(2012, 1, 1, 0, 0),
         observation_time_end=datetime(2012, 1, 2, 0, 0),
         wavemin=np.nan, wavemax=np.nan,
-        instrument='noaa-predict')
+        instrument='NOAA-Predict')
     # 2 entries from norh
-    assert entries[60] == DatabaseEntry(
-        source='NAOJ', provider='NRO', physobs="",
+    assert entries[62] == DatabaseEntry(
+        source='NAOJ', provider='NRO', physobs=None,
         fileid=("ftp://solar-pub.nao.ac.jp/"
                 "pub/nsro/norh/data/tcx/2012/01/tca120101"),
         observation_time_start=datetime(2012, 1, 1, 0, 0),
-        observation_time_end=datetime(2012, 1, 2, 0, 0),
+        observation_time_end=datetime(2012, 1, 1, 23, 59, 59, 999000),
         wavemin=17634850.470588233, wavemax=17634850.470588233,
         instrument='NORH')
     # 1 entry from rhessi
-    assert entries[62] == DatabaseEntry(
-        source="rhessi", provider='nasa', physobs='irradiance',
+    assert entries[64] == DatabaseEntry(
+        source="RHESSI", provider='NASA', physobs='summary_lightcurve',
         fileid=("https://hesperia.gsfc.nasa.gov/"
                 "hessidata/metadata/catalog/hsi_obssumm_20120101_032.fits"),
         observation_time_start=datetime(2012, 1, 1, 0, 0),
         observation_time_end=datetime(2012, 1, 1, 23, 59, 59, 999000),
         wavemin=np.nan, wavemax=np.nan,
-        instrument='rhessi')
-    # 2 entries from eve, level 0
-    assert entries[64] == DatabaseEntry(
-        source='SDO', provider='LASP', physobs='irradiance',
-        fileid=("http://lasp.colorado.edu/eve/data_access/evewebdata/quicklook"
-                "/L0CS/SpWx/2012/20120101_EVE_L0CS_DIODES_1m.txt"),
-        observation_time_start=datetime(2012, 1, 1, 0, 0),
-        observation_time_end=datetime(2012, 1, 2, 0, 0),
-        wavemin=np.nan, wavemax=np.nan,
-        instrument='eve')
+        instrument='RHESSI')
 
 
 @pytest.mark.remote_data
@@ -214,12 +213,12 @@ def test_from_fido_search_result_block(fido_search_result):
     entry = DatabaseEntry._from_fido_search_result_block(
         fido_search_result[0, 0][0].get_response(0).blocks[0])
     expected_entry = DatabaseEntry(
-        source='Proba2', provider='esa', physobs='irradiance',
+        source='PROBA2', provider='ESA', physobs='irradiance',
         fileid='http://proba2.oma.be/lyra/data/bsd/2012/01/01/lyra_20120101-000000_lev2_std.fits',
         observation_time_start=datetime(2012, 1, 1, 0, 0),
-        observation_time_end=datetime(2012, 1, 2, 0, 0),
+        observation_time_end=datetime(2012, 1, 1, 23, 59, 59, 999000),
         wavemin=np.nan, wavemax=np.nan,
-        instrument='lyra')
+        instrument='LYRA')
     assert entry == expected_entry
 
 
--- a/sunpy/net/_attrs.py
+++ b/sunpy/net/_attrs.py
@@ -245,11 +245,35 @@ class Physobs(SimpleAttr):
     """
 
 
-class Source(SimpleAttr):
-    """
-    """
-
-
 class Provider(SimpleAttr):
     """
+    Specifies the data provider to search for data using Fido clients.
+
+    Parameters
+    ----------
+    value : str
+
+    Notes
+    -----
+    For VSO, More information about each provider may be found within in the VSO Registry.
+    For a list of VSO providers see
+    https://sdac.virtualsolar.org/cgi/show_details?keyword=PROVIDER.
+    """
+
+
+class Source(SimpleAttr):
+    """
+    Data sources that Fido clients can search on.
+
+    Parameters
+    ----------
+    value : str
+
+    Notes
+    -----
+    For VSO, More information about each source may be found within in the VSO Registry.
+    User Interface programmers should note that some names may be encoded as
+    UTF-8. Please note that 'Source' is used internally by VSO to represent
+    what the VSO Data Model refers to as 'Observatory'.  For a list of VSO sources
+    see https://sdac.virtualsolar.org/cgi/show_details?keyword=SOURCE.
     """
--- a/sunpy/net/_attrs.py
+++ b/sunpy/net/_attrs.py
@@ -14,7 +14,7 @@ from sunpy.util.decorators import add_common_docstring
 from .attr import Range, SimpleAttr
 
 __all__ = ['Physobs', 'Resolution', 'Detector', 'Sample',
-           'Level', 'Instrument', 'Wavelength', 'Time']
+           'Level', 'Instrument', 'Wavelength', 'Time', 'Source', 'Provider']
 
 
 @add_common_docstring(**_variables_for_parse_time_docstring())
@@ -242,4 +242,14 @@ class Physobs(SimpleAttr):
     More information about the values of physobs used by the VSO
     registry can be found at
     https://sdac.virtualsolar.org/cgi/show_details?keyword=PHYSOBS.
+    """
+
+
+class Source(SimpleAttr):
+    """
+    """
+
+
+class Provider(SimpleAttr):
+    """
     """
--- a/sunpy/net/_attrs.py
+++ b/sunpy/net/_attrs.py
@@ -247,33 +247,33 @@ class Physobs(SimpleAttr):
 
 class Provider(SimpleAttr):
     """
-    Specifies the data provider to search for data using Fido clients.
+    Specifies the data provider to search for data using Fido.
 
     Parameters
     ----------
     value : str
+        A keyword describing the Provider for the data.
 
     Notes
     -----
-    For VSO, More information about each provider may be found within in the VSO Registry.
-    For a list of VSO providers see
-    https://sdac.virtualsolar.org/cgi/show_details?keyword=PROVIDER.
+    For VSO, more information about each provider may be found within in the VSO Registry.
+    See `VSO providers <https://sdac.virtualsolar.org/cgi/show_details?keyword=PROVIDER>`__.
     """
 
 
 class Source(SimpleAttr):
     """
-    Data sources that Fido clients can search on.
+    Data sources that Fido can search with.
 
     Parameters
     ----------
     value : str
+        A keyword describing the Data Source.
 
     Notes
     -----
-    For VSO, More information about each source may be found within in the VSO Registry.
-    User Interface programmers should note that some names may be encoded as
-    UTF-8. Please note that 'Source' is used internally by VSO to represent
-    what the VSO Data Model refers to as 'Observatory'.  For a list of VSO sources
-    see https://sdac.virtualsolar.org/cgi/show_details?keyword=SOURCE.
+    For VSO, more information about each source may be found within in the VSO Registry.
+    See `VSO sources <https://sdac.virtualsolar.org/cgi/show_details?keyword=SOURCE>`__.
+    Please note that 'Source' is used internally by VSO to represent
+    what the VSO Data Model refers to as 'Observatory'.
     """
--- a/sunpy/net/dataretriever/client.py
+++ b/sunpy/net/dataretriever/client.py
@@ -98,6 +98,16 @@ class GenericClient(BaseClient):
             d[attrname] = []
             for val, desc in a[i]:
                 d[attrname].append(val)
+        for elem in args:
+            if isinstance(elem, Time):
+                timerange = TimeRange(elem.start, elem.end)
+                d['timerange'] = timerange
+            elif hasattr(elem, 'value'):
+                d[elem.__class__.__name__] = [str(elem.value)]
+            else:
+                raise ValueError("GenericClient can not add {} to the map_ dictionary to pass to the Client.".format(elem.__class__.__name__))
+        for k in kwargs:
+            d[k] = [kwargs[k]]
         return cls.baseurl, cls.pattern, d
 
     @classmethod
@@ -111,7 +121,11 @@ class GenericClient(BaseClient):
 
         required = {a.Time, a.Instrument}
         adict = cls.register_values()
-        optional = {i for i in adict.keys()}
+        optional = {i for i in adict.keys()} - required
+        if hasattr(cls, 'required'):
+            required = cls.required
+        if hasattr(cls, 'optional'):
+            optional = cls.optional
         if not cls.check_attr_types_in_query(query, required, optional):
             return False
         for x in query:
@@ -128,7 +142,8 @@ class GenericClient(BaseClient):
         map_['Start Time'] = start.strftime(TIME_FORMAT)
         map_['End Time'] = end.strftime(TIME_FORMAT)
         map_['Instrument'] = matchdict['Instrument'][0]
-        map_['Phsyobs'] = matchdict['Physobs'][0]
+        if 'Physobs' in matchdict:
+            map_['Phsyobs'] = matchdict['Physobs'][0]
         map_['Source'] = matchdict['Source'][0]
         map_['Provider'] = matchdict['Provider'][0]
         for k in exdict:
@@ -179,16 +194,7 @@ class GenericClient(BaseClient):
         """
         baseurl, pattern, matchdict = self.pre_hook(*args, **kwargs)
         scraper = Scraper(baseurl, regex=True)
-        for elem in args:
-            if isinstance(elem, Time):
-                timerange = TimeRange(elem.start, elem.end)
-            elif hasattr(elem, 'value'):
-                matchdict[elem.__class__.__name__] = [str(elem.value)]
-            else:
-                raise ValueError("GenericClient can not add {} to the map_ dictionary to pass to the Client.".format(elem.__class__.__name__))
-        for k in kwargs:
-            matchdict[k] = [str(kwargs[k])]
-        filesmeta = scraper._extract_files_meta(timerange, extractor=pattern, matcher=matchdict)
+        filesmeta = scraper._extract_files_meta(matchdict['timerange'], extractor=pattern, matcher=matchdict)
         metalist = []
         for i in filesmeta:
             map_ = self.post_hook(i, matchdict)
--- a/sunpy/net/dataretriever/client.py
+++ b/sunpy/net/dataretriever/client.py
@@ -7,7 +7,7 @@ from astropy.time import TimeDelta
 
 import sunpy
 from sunpy import config
-from sunpy.net._attrs import Time
+from sunpy.net._attrs import Time, Wavelength
 from sunpy.net.base_client import BaseClient, BaseQueryResponse
 from sunpy.time import TimeRange, parse_time
 from sunpy.util.parfive_helpers import Downloader
@@ -107,6 +107,8 @@ class GenericClient(BaseClient):
                 d['Time'] = timerange
             elif hasattr(elem, 'value'):
                 d[elem.__class__.__name__] = [str(elem.value)]
+            elif isinstance(elem, Wavelength):
+                d['Wavelength'] = elem
             else:
                 raise ValueError("GenericClient can not add {} to the map_ dictionary to pass to the Client.".format(elem.__class__.__name__))
         for k in kwargs:
--- a/sunpy/net/dataretriever/client.py
+++ b/sunpy/net/dataretriever/client.py
@@ -66,6 +66,7 @@ class QueryResponse(BaseQueryResponse):
             return astropy.table.Table()
         meta0 = self._data[0]
         meta0.pop('url', None)
+        meta0.pop('Time', None)
         columns = OrderedDict(((col, [])) for col in meta0.keys())
         for qrblock in self:
             for colname in columns.keys():
@@ -86,9 +87,11 @@ class GenericClient(BaseClient):
     set of results for files available through the service the client is
     querying and the latter downloads that data.
     """
+    baseurl = None
+    pattern = None
 
     @classmethod
-    def pre_hook(cls, *args, **kwargs):
+    def _get_match_dict(cls, *args, **kwargs):
         """
         """
         a = cls.register_values()
@@ -101,14 +104,21 @@ class GenericClient(BaseClient):
         for elem in args:
             if isinstance(elem, Time):
                 timerange = TimeRange(elem.start, elem.end)
-                d['timerange'] = timerange
+                d['Time'] = timerange
             elif hasattr(elem, 'value'):
                 d[elem.__class__.__name__] = [str(elem.value)]
             else:
                 raise ValueError("GenericClient can not add {} to the map_ dictionary to pass to the Client.".format(elem.__class__.__name__))
         for k in kwargs:
             d[k] = [kwargs[k]]
-        return cls.baseurl, cls.pattern, d
+        return d
+
+    @classmethod
+    def pre_search_hook(cls, *args, **kwargs):
+        """
+        """
+        matchdict = cls._get_match_dict(*args, **kwargs)
+        return cls.baseurl, cls.pattern, matchdict
 
     @classmethod
     def _can_handle_query(cls, *query):
@@ -134,7 +144,7 @@ class GenericClient(BaseClient):
                 return True
         return False
 
-    def post_hook(self, exdict, matchdict):
+    def post_search_hook(self, exdict, matchdict):
         """
         """
         map_ = OrderedDict()
@@ -142,9 +152,10 @@ class GenericClient(BaseClient):
         if 'month' in exdict and 'day' in exdict:
             start = parse_time("{}/{}/{}".format(exdict['year'], exdict['month'], exdict['day']))
             end = start + almost_day
-        else:
+        elif 'year' in exdict:
             start = parse_time("{}/{}/{}".format(exdict['year'], 1, 1))
             end = parse_time("{}/{}/{}".format(exdict['year'], 12, 31)) + almost_day
+        map_['Time'] = TimeRange(start, end)
         map_['Start Time'] = start.strftime(TIME_FORMAT)
         map_['End Time'] = end.strftime(TIME_FORMAT)
         map_['Instrument'] = matchdict['Instrument'][0]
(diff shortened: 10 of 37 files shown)


abhijeetmanhas:
### Tasks
 * [ ] Generalize `fermi`, `eve`, `lyra`
 * [ ] Generalize `goes`, `noaa`, `norh`
 * [ ] Move `rhessi` out of Generic
 * [ ] Update all tests and docs

sunpy/net/_attrs.py
```diff
@@ -243,3 +243,13 @@ class Physobs(SimpleAttr):
     registry can be found at
     https://sdac.virtualsolar.org/cgi/show_details?keyword=PHYSOBS.
     """
+
+
+class Source(SimpleAttr):
```
Cadair, line 248: aren't these in VSO at the moment?

sunpy/net/dataretriever/client.py
```diff
@@ -89,10 +88,17 @@ class GenericClient(BaseClient):
     """
 
     @classmethod
-    def _get_patterns_for_attrs(cls, **kwargs):
+    def pre_hook(cls, *args, **kwargs):
```
Cadair, line 91: What are the args here? Does this need to be a full `*args, **kwargs`?

I think the query needs to be passed in here? Does anything else?

sunpy/net/dataretriever/client.py
```diff
@@ -225,7 +107,34 @@ def _can_handle_query(cls, *query):
         `sunpy.net.fido_factory.UnifiedDownloaderFactory`
         class uses to dispatch queries to this Client.
         """
-        raise NotImplementedError
+        from sunpy.net import attrs as a
+
+        required = {a.Time, a.Instrument}
```
Cadair, line 112: This needs to be a class attr?

sunpy/net/_attrs.py
nabobalis, line 248: Are these not defined under the VSO namespace?

sunpy/net/dataretriever/client.py
```diff
@@ -225,7 +107,34 @@ def _can_handle_query(cls, *query):
...
+        adict = cls.register_values()
+        optional = {i for i in adict.keys()}
```
Cadair, line 114: This needs to be everything in register_values that isn't in required?
```suggestion
        optional = {i for i in adict.keys()}difference(required)
```

nabobalis:
Should be rebased on top of master.

sunpy/net/dataretriever/client.py
```diff
@@ -225,7 +107,34 @@ def _can_handle_query(cls, *query):
...
+                return True
+        return False
+
+    def post_hook(self, exdict, matchdict):
+        """
+        """
+        map_ = OrderedDict()
+        start = parse_time("{}/{}/{}".format(exdict['year'], exdict['month'], exdict['day']))
+        end = start + TimeDelta(1 * u.day - 1 * u.millisecond)
+        map_['Start Time'] = start.strftime(TIME_FORMAT)
```
Cadair, line 128: Why are we converting these to strings to put them in the response, can't we just pass astropy time objects to Table?

sunpy/net/dataretriever/client.py
```diff
@@ -137,86 +85,20 @@ class GenericClient(BaseClient):
...
-                    # Anything that hits this will require special code to
-                    # convert it into the map_ dict.
-                    raise ValueError(
-                        "GenericClient can not add {} to the map_ dictionary to pass "
-                        "to the Client.".format(elem.__class__.__name__))  # pragma: no cover
-        self._makeimap()
-
     @classmethod
-    def _get_url_for_timerange(cls, timerange, **kwargs):
+    def pre_hook(cls, *args, **kwargs):
```
Cadair, line 91:
```suggestion
    def pre_search_hook(cls, *args, **kwargs):
```

sunpy/net/dataretriever/client.py
```diff
@@ -264,35 +173,27 @@ def _get_full_filenames(self, qres, filenames, path):
...
+            elif hasattr(elem, 'value'):
+                matchdict[elem.__class__.__name__] = [str(elem.value)]
+            else:
+                raise ValueError("GenericClient can not add {} to the map_ dictionary to pass to the Client.".format(elem.__class__.__name__))
+        for k in kwargs:
+            matchdict[k] = [str(kwargs[k])]
+        filesmeta = scraper._extract_files_meta(timerange, extractor=pattern, matcher=matchdict)
+        metalist = []
+        for i in filesmeta:
+            map_ = self.post_hook(i, matchdict)
```
Cadair, line 194: What do you think the pros/cons of putting this loop inside `post_hook` are? Are there cases where the source would want to handle all the returns as a group? i.e. to filter them or something mad?

sunpy/net/dataretriever/client.py
abhijeetmanhas: yeah, True will change it accordingly

sunpy/net/dataretriever/client.py
abhijeetmanhas: Yes, `*args` and `**kwargs` of query shall be used for translation here. This pre-hook will be overwritten for some clients like `norh`. Where wavelength shall translate to either `tcz` or `tca` in the url.

sunpy/net/dataretriever/client.py
abhijeetmanhas: For most clients, `Instrument` and `Time` are required attrs so the method is same for all clients.
If any client deviates, it will be implemented in the  subclass.

sunpy/net/dataretriever/client.py
abhijeetmanhas: Passing large `filesmeta`list of dictionaries which may contain a lot of metada to a method will increase memory usage (I'm not very certain about it).
There are no filtering left, all that is handled by the `matchdict` passed in scraper, as we decided in the last meeting.
The `post_hook` will be overwritten for clients like `norh` to convert `tcz` or `tca` in the url to `17*u.GHz` or `34 *u.GHz`.

sunpy/net/dataretriever/client.py
Cadair: yeah but they shouldn't need to override this method to set this, they should just be able to set a class attr?

sunpy/net/dataretriever/sources/noaa.py
```diff
@@ -200,6 +149,22 @@ def _get_url_for_timerange(self, timerange, **kwargs):
...
         return result
 
+    def search(self, *args, **kwargs):
+        extractor1 = '{}/warehouse/{:4d}/SRS/{year:4d}{month:2d}{day:2d}SRS.txt'
+        extractor2 = '{}/warehouse/{year:4d}/{}'
+        matchdict = {'Instrument': ['SOON'], 'Source': ['SWPC'], 'Provider': ['NOAA'], 'Physobs': ['SRS']}
+        metalist = []
+        for elem in args:
+            if isinstance(elem, a.Time):
+                timerange = TimeRange(elem.start, elem.end)
```
Cadair, line 159: Can you not replace this with a call to `pre_hook`?

sunpy/net/dataretriever/sources/noaa.py
```diff
@@ -200,6 +149,22 @@ def _get_url_for_timerange(self, timerange, **kwargs):
             result.append(url)
         return result
 
+    def search(self, *args, **kwargs):
+        extractor1 = '{}/warehouse/{:4d}/SRS/{year:4d}{month:2d}{day:2d}SRS.txt'
+        extractor2 = '{}/warehouse/{year:4d}/{}'
+        matchdict = {'Instrument': ['SOON'], 'Source': ['SWPC'], 'Provider': ['NOAA'], 'Physobs': ['SRS']}
```
Cadair, line 155: Also this would come out of `pre_hook` right?

sunpy/net/dataretriever/sources/norh.py
```diff
@@ -35,119 +34,55 @@ class NoRHClient(GenericClient):
...
-        """
-        Helper Function used to hold information about source.
-        """
-        self.map_['source'] = 'NAOJ'
-        self.map_['provider'] = 'NRO'
-        self.map_['instrument'] = 'NORH'
-        self.map_['physobs'] = ''
+    baseurl = r'ftp://solar-pub.nao.ac.jp/pub/nsro/norh/data/tcx/%Y/%m/(\w){3}%y%m%d'
+    pattern = '{}/tcx/{year:4d}/{month:2d}/{Wavelength:3l}{:4d}{day:2d}'
+    optional = {a.Wavelength, a.Source, a.Provider}
```
Cadair, line 47: I thought the plan was to have `required` as a class attribute and derive optional from `register_attrs`?

sunpy/net/dataretriever/client.py
```diff
@@ -137,86 +85,30 @@ class GenericClient(BaseClient):
     `~sunpy.net.dataretriever.client.GenericClient.fetch` the former generates a
     set of results for files available through the service the client is
     querying and the latter downloads that data.
-
-    The `~sunpy.net.dataretriever.client.GenericClient.search` method takes a
-    set of `sunpy.net.attrs` objects and then converts these into a call to
-    `~sunpy.net.dataretriever.client.GenericClient._get_url_for_timerange`. It
-    does this through the `map\\_` dictionary which represents the
-    `~sunpy.net.attrs` objects as a dictionary.
     """
```
Cadair, line 89:
```suggestion
    baseurl = None
    pattern = None
```

sunpy/net/_attrs.py
abhijeetmanhas: Moved from vso to net

abhijeetmanhas: Moved from vso to net

sunpy/net/dataretriever/client.py
abhijeetmanhas: Done

sunpy/net/dataretriever/sources/noaa.py
abhijeetmanhas: Done

sunpy/net/dataretriever/client.py
abhijeetmanhas: Done

sunpy/net/dataretriever/sources/noaa.py
abhijeetmanhas: Done

sunpy/net/dataretriever/sources/norh.py
abhijeetmanhas: This can't be done, since we can't derive `wavelength` from register_attrs as we planned not to register it.

sunpy/net/dataretriever/client.py
abhijeetmanhas: Done

sunpy/net/_attrs.py
```diff
@@ -243,3 +243,37 @@ class Physobs(SimpleAttr):
     registry can be found at
     https://sdac.virtualsolar.org/cgi/show_details?keyword=PHYSOBS.
     """
+
+
+class Provider(SimpleAttr):
+    """
+    Specifies the data provider to search for data using Fido clients.
```
nabobalis, line 250:
```suggestion
    Specifies the data provider to search for data using Fido.
```

sunpy/net/_attrs.py
```diff
@@ -243,3 +243,37 @@ class Physobs(SimpleAttr):
...
+
+    Parameters
+    ----------
+    value : str
+
```
nabobalis, line 255: This should be completed.

sunpy/net/_attrs.py
```diff
@@ -243,3 +243,37 @@ class Physobs(SimpleAttr):
...
+    Notes
+    -----
+    For VSO, More information about each provider may be found within in the VSO Registry.
```
nabobalis, line 258:
```suggestion
    For VSO, more information about each provider may be found within in the VSO Registry.
```

sunpy/net/_attrs.py
```diff
@@ -243,3 +243,37 @@ class Physobs(SimpleAttr):
...
+    For a list of VSO providers see
+    https://sdac.virtualsolar.org/cgi/show_details?keyword=PROVIDER.
```
nabobalis, line 260:
```suggestion
```

sunpy/net/_attrs.py
```diff
@@ -243,3 +243,37 @@ class Physobs(SimpleAttr):
...
+    Specifies the data provider to search for data using Fido clients.
+
+    Parameters
+    ----------
+    value : str
+
+    Notes
+    -----
+    For VSO, More information about each provider may be found within in the VSO Registry.
+    For a list of VSO providers see
```
nabobalis, line 259:
```suggestion
    See `VSO providers <https://sdac.virtualsolar.org/cgi/show_details?keyword=PROVIDER>`__.
```

sunpy/net/_attrs.py
```diff
@@ -243,3 +243,37 @@ class Physobs(SimpleAttr):
...
+    https://sdac.virtualsolar.org/cgi/show_details?keyword=PROVIDER.
+    """
+
+
+class Source(SimpleAttr):
+    """
+    Data sources that Fido clients can search on.
```
nabobalis, line 266:
```suggestion
    Data sources that Fido can search with.
```

sunpy/net/_attrs.py
```diff
@@ -243,3 +243,37 @@ class Physobs(SimpleAttr):
...
+
+    Parameters
+    ----------
+    value : str
```
nabobalis, line 270: This should be completed.

sunpy/net/_attrs.py
```diff
@@ -243,3 +243,37 @@ class Physobs(SimpleAttr):
...
+
+    Notes
+    -----
+    For VSO, More information about each source may be found within in the VSO Registry.
```
nabobalis, line 274:
```suggestion
    For VSO, more information about each source may be found within in the VSO Registry.
```

sunpy/net/_attrs.py
```diff
@@ -243,3 +243,37 @@ class Physobs(SimpleAttr):
...
+    User Interface programmers should note that some names may be encoded as
```
nabobalis, line 275:
```suggestion
    See `VSO sources <https://sdac.virtualsolar.org/cgi/show_details?keyword=SOURCE>`__.
```

sunpy/net/_attrs.py
```diff
@@ -243,3 +243,37 @@ class Physobs(SimpleAttr):
...
+    UTF-8. Please note that 'Source' is used internally by VSO to represent
```
nabobalis, line 276:
```suggestion
    Please note that 'Source' is used internally by VSO to represent
```

sunpy/net/_attrs.py
```diff
@@ -243,3 +243,37 @@ class Physobs(SimpleAttr):
...
+    what the VSO Data Model refers to as 'Observatory'.  For a list of VSO sources
```
nabobalis, line 277:
```suggestion
    what the VSO Data Model refers to as 'Observatory'.
```

sunpy/net/_attrs.py
```diff
@@ -243,3 +243,37 @@ class Physobs(SimpleAttr):
...
+    see https://sdac.virtualsolar.org/cgi/show_details?keyword=SOURCE.
```
nabobalis, line 278:
```suggestion
```

sunpy/net/vso/attrs.py
```diff
@@ -64,40 +63,6 @@ def collides(self, other):
         return isinstance(other, self.__class__)
 
 
-class Provider(_attr.SimpleAttr):
```
nabobalis, line 67: You need to add these to the list of removed attrs at the bottom of this file.

sunpy/net/dataretriever/client.py
```diff
@@ -137,86 +93,59 @@ class GenericClient(BaseClient):
...
-        This `map\\_` dictionary is passed through to the
-        `_get_url_for_timerange` method to get the URL results.
+        Returns a dictionary to validate the metadata of searched files using
+        query and registered values for Attrs for the respective client.
 
         Parameters
         ----------
         \\*args: `tuple`
-            The query attributes.
+            `sunpy.net.attrs` objects representing the query.
```
nabobalis, line 110:
```suggestion
```

sunpy/net/dataretriever/client.py
nabobalis: This is still in the method and not a class attr?

sunpy/net/dataretriever/client.py
```diff
@@ -281,18 +257,22 @@ def search(self, *args, **kwargs):
         ----------
         \\*args: `tuple`
             `sunpy.net.attrs` objects representing the query.
+
```
nabobalis, line 260:
```suggestion
```

sunpy/net/dataretriever/sources/fermi_gbm.py
```diff
@@ -37,118 +36,28 @@ class GBMClient(GenericClient):
...
-                                 ('CTIME', 'counts accumulated every 0.256 seconds in 8 energy channels')],
+                 attrs.Physobs: [('flux', 'a measure of the amount of radiation received by an object from a given source.')],
+                 attrs.Source: [('FERMI', 'The Fermi Gamma-ray Space Telescope.')],
+                 attrs.Provider: [('NASA', 'The National Aeronautics and Space Administration.')],
                  attrs.Detector: [
             (f"n{x}", f"GBM Detector short name for the detector NAI_{x:02}") for x in range(12)],
             attrs.Resolution: [
-            ("CSPEC", "CSPEC 128 channel spectra every 4.096 seconds."),
-            ("CTIME", "CTIME provides 8 channel spectra every 0.256 seconds")]
+            ("cspec", "CSPEC 128 channel spectra every 4.096 seconds."),
```
nabobalis, line 60: I feel like the attrs should be case insensitive, if they are not, we might want to change that.

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -223,183 +85,86 @@ class SUVIClient(GenericClient):
...
-        if check_var_count == 1:
-            return True
+    baseurl1b = (r'https://data.ngdc.noaa.gov/platforms/solar-space-observing-satellites/goes/goes'
+                 r'{SatelliteNumber}/l1b/suvi-l1b-{elem:2}{wave:03}/%Y/%m/%d/OR_SUVI-L1b.*\.fits.gz')
+    pattern1b = ('{}/goes/goes{SatelliteNumber:2d}/l{Level:2w}/suvi-l1b-{}{Wavelength:03d}/'
+                 '{year:4d}/{month:2d}/{day:2d}/{}_s{:7d}{hour:2d}{minute:2d}{second:2d}'
+                 '{:1d}_e{:7d}{ehour:2d}{eminute:2d}{esecond:2d}{:1d}_{}')
+    baseurl2 = (r'https://data.ngdc.noaa.gov/platforms/solar-space-observing-satellites/goes/goes{SatelliteNumber}/'
+                r'l2/data/suvi-l2-ci{wave:03}/%Y/%m/%d/dr_suvi-l2-ci{wave:03}_g{SatelliteNumber}_s%Y%m%dT%H%M%SZ_.*\.fits')
+    pattern2 = ('{}/goes/goes{SatelliteNumber:2d}/{}/dr_suvi-l{Level}-ci{Wavelength:03d}_g{SatelliteNumber:2d}_s'
```
nabobalis, line 95: In theory this could be reduced if the scrapper was smarter?

sunpy/net/_attrs.py
abhijeetmanhas: Done

sunpy/net/_attrs.py
abhijeetmanhas: Done

sunpy/net/vso/attrs.py
abhijeetmanhas: Done

sunpy/net/dataretriever/client.py
abhijeetmanhas: yes, since it is by default `instrument` and `Time`. If any client want to overwrite it, then it can be defined as class attributes for that clients, and the generic `_can_handle_query` checks if they are defined as class attributes or not. If not, it will follow the default rule.

sunpy/net/dataretriever/sources/fermi_gbm.py
abhijeetmanhas: That would have been ideal, but the problem is, the arhive website may contain these attrs in any case, and our pattern matching will fail if we don't follow the website.

If we want to ignore case, that can also be done, but we have to then overwrite `pre_search_hook()` method (which norh does for converting 34 ghz to `tca`). But defining pre-hook for converting `CTIME` to `ctime` or `Ctime` looked like overkill.
So I defined the case the archive follows in register values itself.

sunpy/net/dataretriever/sources/goes.py
abhijeetmanhas: Yeah, only convertors need to befined. (to convert level 2 to `/data/l2/` in baseurl). We can do everything with one baseurl itself.

abhijeetmanhas:
`fido_factory` tests updated. `test_fermi.py` can be reviewed as an example how all other sources clients test would look like.
Have to yet update/add tests for other clients, and look for failures outside net.

nabobalis:
Having had a quick glance over the last two commits, I see no reason not to carry on.

abhijeetmanhas:
Database tests only remaining to be updated.

sunpy/net/dataretriever/sources/tests/test_lyra_ud.py
```diff
@@ -19,21 +22,22 @@ def LCClient():
 
 @pytest.mark.remote_data
 @pytest.mark.parametrize("timerange,url_start,url_end", [
-    (TimeRange('2012/1/7', '2012/1/7'),
+    (Time('2012/1/7', '2012/1/7'),
```
nabobalis, line 25: Does this return a timerange?

abhijeetmanhas: It was `from sunpy.net.attrs import Time` So I just used it.

nabobalis: Fair enough.

docs/guide/acquiring_data/fido.rst
```diff
@@ -62,19 +62,6 @@ Using ``Instrument`` as the first example, if you print the object::
 You get a full list of known values, a description and what "Clients" support those values (if you want to use a specific data source).
 This is supported for most attributes including the client specific ones.
 
-For VSO::
```
Cadair, line 65: Are we really out of VSO specific attrs we register values for?

sunpy/net/dataretriever/client.py
Cadair: But why not just set required on this class, so it's always defined and then use it consistently in this method rather than doing `hasattr`. I would also think it would be a good idea to define `optional = set()` as a class attr as well, so these are always defined and we don't have to check if they are there.

sunpy/net/dataretriever/client.py
```diff
@@ -225,7 +176,63 @@ def _can_handle_query(cls, *query):
...
+        from sunpy.net import attrs as a
+
+        required = {a.Time, a.Instrument}
+        adict = cls.register_values()
+        optional = {i for i in adict.keys()} - required
+        if hasattr(cls, 'required'):
+            required = cls.required
+        if hasattr(cls, 'optional'):
+            optional = cls.optional
+        if not cls.check_attr_types_in_query(query, required, optional):
```
Cadair, line 188: As per the thread above:
```suggestion
        if not cls.check_attr_types_in_query(query, cls.required, cls.optional):
```

sunpy/net/dataretriever/client.py
```diff
@@ -225,7 +176,63 @@ def _can_handle_query(cls, *query):
...
+        for key in adict:
+            all_vals = [i[0].lower() for i in adict[key]]
+            for x in query:
+                if isinstance(x, key) and str(x.value).lower() not in all_vals:
+                    return False
+        return True
+
+    def post_search_hook(self, exdict, matchdict):
+        """
+        Helper function used after `search()` which makes the
```
Cadair, line 199:
```suggestion
        Helper function used after `:func:~sunpy.net.dataretriever.GenericClient.search` which makes the
```

sunpy/net/dataretriever/client.py
```diff
@@ -225,7 +176,63 @@ def _can_handle_query(cls, *query):
...
+        almost_day = TimeDelta(1 * u.day - 1 * u.millisecond)
+        if 'month' in exdict and 'day' in exdict:
+            start = parse_time("{}/{}/{}".format(exdict['year'], exdict['month'], exdict['day']))
+            end = start + almost_day
+        elif 'year' in exdict:
+            start = parse_time("{}/{}/{}".format(exdict['year'], 1, 1))
+            end = parse_time("{}/{}/{}".format(exdict['year'], 12, 31)) + almost_day
+        map_['Time'] = TimeRange(start, end)
+        map_['Start Time'] = start.strftime(TIME_FORMAT)
+        map_['End Time'] = end.strftime(TIME_FORMAT)
```
Cadair, line 226: Why are we injecting the timerange and the start and end times, why not just one or the other?

sunpy/net/dataretriever/client.py
```diff
@@ -1,63 +1,45 @@
...
 from sunpy.util.parfive_helpers import Downloader
+from sunpy.util.scraper import Scraper
 
 TIME_FORMAT = config.get("general", "time_format")
 
 __all__ = ['QueryResponse', 'GenericClient']
 
 
-class QueryResponseBlock:
+class QueryResponseBlock(OrderedDict):
```
Cadair, line 21: This PR feels like a good opportunity to get rid of this class? Or shall we do that later?

nabobalis: I thought this was removed?

docs/guide/acquiring_data/fido.rst
abhijeetmanhas: Yes, though there are `Pscale`, `Extent`, `Pixel` etc, but nothing is registered for them.

sunpy/net/dataretriever/client.py
abhijeetmanhas: This was done so that we have `Time` as a Quantity in the table which can be used for inspection. And start time/endtime as strings which can be used for showing in the response table, so that results look similar as it was before for time.

If we inject only `Time`, we had to do that thing in `build_table`, and as I also had to preserve order of colnames in the table, it was relatively difficult to implement.

sunpy/net/dataretriever/client.py
abhijeetmanhas: I had to bring back `QueryResponseBlock` due to two reasons:
- All vso, jsoc, dataretriever individual records (rows in table) used to be classes, and to access their attributes, we have to do something like `row.url` instead of `row['ur']` in my implementation. So this ensured uniformity.
- A huge changes might have to be done in `database` tests, first by checking what type of response it is and whether to do `['url']` or `.url`. So it saved me from it.

examples/time_series/goes_hek_m25.py
```diff
@@ -15,7 +15,7 @@
 
 ###############################################################################
 # Let's first grab GOES XRS data for a particular time of interest
-tr = TimeRange(['2011-06-07 04:00', '2011-06-07 12:00'])
+tr = TimeRange(['2011-06-07 00:00', '2011-06-07 12:00'])
```
nabobalis, line 18: What caused this change?

sunpy/net/dataretriever/client.py
nabobalis: The question is, is it worth removing this class and just doing those annoying little changes everywhere?

examples/time_series/goes_hek_m25.py
abhijeetmanhas: The goes `xrs` had one day files, so my implementation defaults the start time of a record to `00:00` on that day.
When the start time is given `4:00` it isn't less than `0:00` so scraper doesn't return that file.

sunpy/net/dataretriever/client.py
abhijeetmanhas: Done

sunpy/net/dataretriever/client.py
abhijeetmanhas: Done

sunpy/net/dataretriever/client.py
abhijeetmanhas: Removed in the latest commit, there weren't lot of changes.
Now the question is, should we keep the way to access fields from row data of QueryResponse (earlier QueryResponseBlock ) different for VSO and DR?
I think it would be rarely used. And after `:func:show` added, it won't be required.

nabobalis: Different how?

abhijeetmanhas: Like in VSO, the zeep client makes each block an object, so time is accessed like `block.time` or `getattr(block, 'time')`.

But I make them a dictionary, so it is accessed like `block['time']` or `block.get('time')`

nabobalis: On the surface I would think we go for the latter, is it not how astropy.tables are indexed?

abhijeetmanhas: Exactly, that's how we index cols in astropy table, like I'm doing in dict

sunpy/net/dataretriever/client.py
abhijeetmanhas: Done

sunpy/net/dataretriever/client.py
```diff
@@ -137,86 +114,60 @@ class GenericClient(BaseClient):
...
-    def _makeargs(self, *args):
+    @classmethod
+    def _get_match_dict(cls, *args, **kwargs):
         """
-        Construct the `map\\_` internal representation of the query.
-
-        This `map\\_` dictionary is passed through to the
-        `_get_url_for_timerange` method to get the URL results.
+        Returns a dictionary to validate the metadata of searched files using
+        query and registered values for Attrs for the respective client.
```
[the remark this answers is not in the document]
abhijeetmanhas: Reprhased

changelog/4321.feature.rst
```diff
@@ -0,0 +1,14 @@
+Refactoring of `~sunpy.net.dataretriever` which lead to the following changes:
```
nabobalis, line 1: This should be split into multiple changelogs

abhijeetmanhas: Done

changelog/4321.breaking.1.rst
```diff
@@ -0,0 +1 @@
+Removed `:func:~sunpy.net.dataretriever.QueryResponse.create`.
```
nabobalis, line 1: Can you check if ":func:" should be outside or inside of the backticks?

changelog/4321.breaking.2.rst
```diff
@@ -0,0 +1 @@
+Removed `~sunpy.net.dataretriever.QueryResponseBlock` as `~sunpy.net.dataretriever.QueryResponse` now uses a ``list`` of `~collections.OrderedDict` for building the response table.
```
nabobalis, line 1:
```suggestion
Removed `~sunpy.net.dataretriever.QueryResponseBlock` as `~sunpy.net.dataretriever.QueryResponse` now uses a `list` of `~collections.OrderedDict` for building the response table.
```

changelog/4321.feature.1.rst
```diff
@@ -0,0 +1,3 @@
+Additions in `~sunpy.util.scraper` to support the refactoring of ``GenericClient``:
```
nabobalis, line 1:
```suggestion
Additions in `~sunpy.util.scraper` to support the refactoring of ~sunpy.net.dataretriever.GenericClient:
```

changelog/4321.breaking.1.rst
abhijeetmanhas: According to docs
```
python function :py:func:`package.name.function`
:py:meth:`package.name.ClassName.method`.
```
So ideally , we should use `:meth:` here. But we have used both meth and func across sunpy, in changelog and the rst docs. However, in this PR I'll use meth. And func should be outside those ticks, which also need to be changed.

nabobalis: Can you use the correct one?

abhijeetmanhas: Yes, I have used the correct one, in all changelogs now.

examples/time_series/goes_hek_m25.py
```diff
@@ -15,7 +15,7 @@
 
 ###############################################################################
 # Let's first grab GOES XRS data for a particular time of interest
-tr = TimeRange(['2011-06-07 04:00', '2011-06-07 12:00'])
```
hayesla, line 18: hmmmm this is an issue now I guess - will `tr = TimeRange(['2011-06-07 04:00', '2011-06-07 12:00'])` this now not return the file for this day? it really should - many people will search only for a flare time so this is important!

sunpy/database/tests/test_database.py
```diff
@@ -74,12 +74,11 @@ def fido_search_result():
     # No JSOC query
     return Fido.search(
         net_attrs.Time("2012/1/1", "2012/1/2"),
-        net_attrs.Instrument('lyra') | net_attrs.Instrument('eve') |
+        net_attrs.Instrument('lyra') & net_attrs.Level.two | net_attrs.Instrument('eve') |
```
hayesla, line 77:
```suggestion
        net_attrs.Instrument('lyra') & net_attrs.Level(2) | net_attrs.Instrument('eve') |
```

sunpy/database/tests/test_tables.py
```diff
@@ -55,12 +55,11 @@ def fido_search_result():
     # No JSOC query
     return Fido.search(
         net_attrs.Time("2012/1/1", "2012/1/2"),
-        net_attrs.Instrument('lyra') | net_attrs.Instrument('eve') |
+        net_attrs.Instrument('lyra') & net_attrs.Level.two | net_attrs.Instrument('eve') |
```
hayesla, line 58:
```suggestion
        net_attrs.Instrument('lyra') & net_attrs.Level(2) | net_attrs.Instrument('eve') |
```

sunpy/net/dataretriever/sources/tests/test_goes_ud.py
```diff
@@ -23,27 +23,29 @@ def LCClient():
 @pytest.mark.remote_data
 @pytest.mark.parametrize(
     "timerange,url_start,url_end",
-    [(TimeRange('1995/06/03', '1995/06/05'),
+    [(Time('1995/06/03', '1995/06/05'),
```
hayesla, line 26: just a general comment - we should probably have some tests that also include %H%m to check this works too

hayesla:
This is great! 

I have one issue that needs to be addressed and fixed before this can be approved.
If for example you search over a timerange during a certain day doesn't return anything - 
for example:
```python
>>> result = Fido.search(a.Time('2013-10-28 01:00', '2013-10-28 03:00'), a.Instrument('XRS')) 
>>> print(result)
Results from 1 Provider:

0 Results from the XRSClient:
<No columns>

```

this is a bug that needs to be fixed!

hayesla:
this is similarly a problem is someone spans two days but not at 00:00

for example:
```python
>>> result = Fido.search(a.Time('2013-10-28 01:00', '2013-10-29 03:00'), a.Instrument('XRS')) 
>>> print(result)
Results from 1 Provider:

1 Results from the XRSClient:
     Start Time           End Time      Instrument  Physobs   Source Provider SatelliteNumber
------------------- ------------------- ---------- ---------- ------ -------- ---------------
2013-10-29 00:00:00 2013-10-29 23:59:59        XRS irradiance   NASA     SDAC              15
```
it only returns the second day.

docs/guide/acquiring_data/fido.rst
```diff
@@ -98,7 +85,7 @@ Searching for Data Using Fido
 
 For example::
 
-    >>> result = Fido.search(a.Time('2012/3/4', '2012/3/6'), a.Instrument.lyra) # doctest: +REMOTE_DATA
+    >>> result = Fido.search(a.Time('2012/3/4', '2012/3/6'), a.Instrument.lyra, a.Level.two) # doctest: +REMOTE_DATA
```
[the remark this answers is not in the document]
Cadair: I think we need to decide which one we are doing and then do it everywhere. I think @nabobalis changed it everywhere applicable.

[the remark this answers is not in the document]
nabobalis: Do whatever you want.

hayesla:
and just note this is an issue with all the clients it seems not just XRS

docs/guide/acquiring_data/fido.rst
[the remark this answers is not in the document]
abhijeetmanhas: Both will work by the way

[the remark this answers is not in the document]
hayesla: either is fine - just a personal preference nothing else!

examples/time_series/goes_hek_m25.py
abhijeetmanhas: This issue is due to all clients assuming that if a filename contains the timestamp of a file, 
For example goes contain `2011/6/7` in the url, so it start time is considered as `2011/6/7 00:00` since there's no info about `HHMMSS` and there's one file per day. 
Since scraper compares `start_time_in_attr<=start_time_extracted`, so if we do `2011/4/7 4:00`, so file for `2011/6/7` is not returned. What best thing we should be doing? 

One thing in my mind is compare inferred endtime of a file with the end time provided in the attr. If one of them is true, then we will return the file.

sunpy/net/dataretriever/client.py
```diff
@@ -225,7 +170,57 @@ def _can_handle_query(cls, *query):
...
+            start = parse_time("{}/{}/{}".format(exdict['year'], exdict['month'], exdict['day']))
+            end = start + almost_day
+        elif 'year' in exdict:
+            start = parse_time("{}/{}/{}".format(exdict['year'], 1, 1))
+            end = parse_time("{}/{}/{}".format(exdict['year'], 12, 31)) + almost_day
+        map_['Time'] = TimeRange(start, end)
+        map_['Start Time'] = start.strftime(TIME_FORMAT)
+        map_['End Time'] = end.strftime(TIME_FORMAT)
+        map_['Instrument'] = matchdict['Instrument'][0].upper()
+        if 'Physobs' in matchdict:
```
Cadair, line 216: why only the if statement for physobs, what if my new client doesn't define a source, provider or even instrument? This whole method feels like it's pretty specialised for what exists at the moment rather than the design of the `GenericClient`.

Am I right in thinking that this is a converter from `cls.pattern` to a response table? Can we not generalise this by restricting (via docs) the possible names of parameters in `cls.pattern`?

sunpy/net/dataretriever/client.py
```diff
@@ -225,7 +170,57 @@ def _can_handle_query(cls, *query):
...
+
+        Returns
+        -------
+        map_: `~collections.OrderedDict`
+            An Ordered Dictionary which is used by `QueryResponse`
+            to show results.
+        """
+        map_ = OrderedDict()
+        almost_day = TimeDelta(1 * u.day - 1 * u.millisecond)
+        if 'month' in exdict and 'day' in exdict:
```
Cadair, line 206: Now I think about it, the fact there is `"year", "month" and "day"` in this dict at all is weird.

sunpy/net/dataretriever/client.py
```diff
@@ -225,7 +170,57 @@ def _can_handle_query(cls, *query):
...
+            map_['Physobs'] = matchdict['Physobs'][0]
+        map_['Source'] = matchdict['Source'][0].upper()
+        map_['Provider'] = matchdict['Provider'][0].upper()
+        for k in exdict:
+            if k not in ['year', 'month', 'day']:
+                map_[k] = exdict[k]
+        return map_
 
     def _get_full_filenames(self, qres, filenames, path):
         """
```
Cadair, line 226: This docstring is telling lies, it downloads nothing.

sunpy/net/dataretriever/client.py
Cadair: In fact, couldn't we use the fields present in the pattern to know which things we need to process here?

examples/time_series/goes_hek_m25.py
kakirastern: I agree with @hayesla in that it is customary to return the file for the day being searched, otherwise it would be against the usual user expectations...

sunpy/net/dataretriever/client.py
```diff
@@ -281,18 +267,22 @@ def search(self, *args, **kwargs):
...
-            if times and times is not NotImplemented:
-                return QueryResponse.create(self.map_, urls, times, client=self)
-        return QueryResponse.create(self.map_, urls, client=self)
+        baseurl, pattern, matchdict = self.pre_search_hook(*args, **kwargs)
+        scraper = Scraper(baseurl, regex=True)
+        filesmeta = scraper._extract_files_meta(matchdict['Time'], extractor=pattern,
+                                                matcher=matchdict)
+        metalist = []
+        for i in filesmeta:
+            map_ = self.post_search_hook(i, matchdict)
```
Cadair, line 283: :x: 

*hisses at `map_`*

examples/time_series/goes_hek_m25.py
hayesla: hmm is there a way to implement within goes.py? such that if the hh:mm is given it will automatically trim them to 00:00?

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -222,184 +84,108 @@ class SUVIClient(GenericClient):
...
+        map_['SatelliteNumber'] = i['SatelliteNumber']
+        map_['Level'] = i['Level']
+        map_['Wavelength'] = i['Wavelength']*u.Angstrom
+        map_['url'] = i['url']
+        return map_
+
+    def search(self, *args, **kwargs):
+        supported_waves = [94, 131, 171, 195, 284, 304]*u.Angstrom
+        all_waves = []
+        baseurl1b, pattern1b, baseurl2, pattern2, matchdict = self.pre_search_hook(*args, **kwargs)
```
Cadair, line 148: why even bother with a call to `pre_search_hook` in this client?

sunpy/net/dataretriever/sources/noaa.py
```diff
@@ -32,55 +34,33 @@ class NOAAIndicesClient(GenericClient):
...
-                return all(chklist)
-        return False
+    required = {a.Instrument}
+    optional = {a.Time, a.Source, a.Physobs, a.Provider}
+
+    def search(self, *args, **kwargs):
+        map_ = self._get_match_dict(*args, **kwargs)
+        for key in map_:
+            if isinstance(map_[key], list):
+                map_[key] = map_[key][0]
```
Cadair, line 51: I think I know what's happening here, but a comment would help.

sunpy/net/dataretriever/sources/noaa.py
```diff
@@ -102,59 +82,33 @@ class NOAAPredictClient(GenericClient):
...
```
Cadair, line 99: same here.

sunpy/net/dataretriever/sources/noaa.py
```diff
@@ -200,6 +158,21 @@ def _get_url_for_timerange(self, timerange, **kwargs):
...
+        extractor1 = '{}/warehouse/{:4d}/SRS/{year:4d}{month:2d}{day:2d}SRS.txt'
+        extractor2 = '{}/warehouse/{year:4d}/{}'
+        matchdict = self._get_match_dict(*args, **kwargs)
+        timerange = matchdict['Time']
+        metalist = []
+        for url in self._get_url_for_timerange(timerange):
+            exdict1 = parse(extractor1, url)
+            exdict2 = parse(extractor2, url)
+            exdict = (exdict2 if exdict1 is None else exdict1).named
+            map_ = super().post_search_hook(exdict, matchdict)
```
Cadair, line 171: why `super()` we haven't overriden it right?

sunpy/net/dataretriever/client.py
```diff
@@ -137,86 +101,67 @@ class GenericClient(BaseClient):
...
-    The `~sunpy.net.dataretriever.client.GenericClient.search` method takes a
-    set of `sunpy.net.attrs` objects and then converts these into a call to
-    `~sunpy.net.dataretriever.client.GenericClient._get_url_for_timerange`. It
-    does this through the `map\\_` dictionary which represents the
-    `~sunpy.net.attrs` objects as a dictionary.
     """
+    baseurl = None
+    pattern = None
+    required = {a.Time, a.Instrument}
+    optional = set()
```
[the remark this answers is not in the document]
Cadair: I found [this](https://github.com/sunpy/sunpy/pull/4321/files#r449536787) old thread where it was pointed out that the main issue is that `Wavelength` (and I assume `Time`) isn't registered. This makes me think we should be registering them, even if we can't register values for them.

@nabobalis what do you think about that? Could we just pass `{a.Wavelength: '*'}` and everything just work?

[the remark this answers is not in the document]
nabobalis: Try it and see?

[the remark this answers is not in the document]
nabobalis: I am not sure I ever tested non-simple attrs with `*`

changelog/4321.feature.1.rst
```diff
@@ -0,0 +1,3 @@
+Additions in `~sunpy.util.scraper` to support the refactoring of `~sunpy.net.dataretriever.GenericClient`:
+- :meth:`~sunpy.util.scraper.Scraper.findDatewith_extractor` which uses the parses the url using extractor to return its start time.
```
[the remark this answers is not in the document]
abhijeetmanhas: Yes,it was wrong.
Changed to
```- :meth:`~sunpy.util.scraper.Scraper.findDatewith_extractor` which parses the url using extractor to return its start time.
```

sunpy/net/dataretriever/client.py
```diff
@@ -137,86 +101,67 @@ class GenericClient(BaseClient):
...
     set of results for files available through the service the client is
     querying and the latter downloads that data.
-
-    The `~sunpy.net.dataretriever.client.GenericClient.search` method takes a
-    set of `sunpy.net.attrs` objects and then converts these into a call to
-    `~sunpy.net.dataretriever.client.GenericClient._get_url_for_timerange`. It
-    does this through the `map\\_` dictionary which represents the
-    `~sunpy.net.attrs` objects as a dictionary.
     """
+    baseurl = None
```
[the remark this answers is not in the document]
abhijeetmanhas: Done.

sunpy/net/dataretriever/client.py
```diff
@@ -225,7 +170,57 @@ def _can_handle_query(cls, *query):
         `sunpy.net.fido_factory.UnifiedDownloaderFactory`
         class uses to dispatch queries to this Client.
         """
-        raise NotImplementedError
+        adict = cls.register_values()
```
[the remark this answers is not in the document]
abhijeetmanhas: Changed to `regattrs_dict`

sunpy/net/dataretriever/client.py
```diff
@@ -137,86 +101,67 @@ class GenericClient(BaseClient):
     `~sunpy.net.dataretriever.client.GenericClient.fetch` the former generates a
```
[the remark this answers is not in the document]
abhijeetmanhas: Added some details of post and pre hooks. Also explained `baseurl` `pattern` etc.

sunpy/net/dataretriever/client.py
```diff
@@ -137,86 +101,67 @@ class GenericClient(BaseClient):
...
         \\*args: `tuple`
-            The query attributes.
+            `sunpy.net.attrs` objects representing the query.
+        \\*\\*kwargs: `dict`
+             Any extra keywords to refine the search.
 
+        Returns
+        -------
+        matchdict: `dict`
+            A dictionary having a `list` of all possible Attr values
```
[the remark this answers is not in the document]
abhijeetmanhas: It knows them all by `register_values()`. Only those which are registered in that method will be known to the client.

sunpy/net/dataretriever/client.py
```diff
@@ -225,7 +170,57 @@ def _can_handle_query(cls, *query):
...
+            Contains attr values accessed from `register_values()`
+            and the search query itself.
+
+        Returns
+        -------
+        map_: `~collections.OrderedDict`
+            An Ordered Dictionary which is used by `QueryResponse`
+            to show results.
+        """
+        map_ = OrderedDict()
```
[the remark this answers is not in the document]
abhijeetmanhas: changed to `rowdict` everywhere.

sunpy/net/dataretriever/sources/goes.py
abhijeetmanhas: You're right, not calling now.

sunpy/net/dataretriever/sources/noaa.py
abhijeetmanhas: Added.

sunpy/net/dataretriever/sources/noaa.py
abhijeetmanhas: Actually the `SRSClient` is weird. It changes its pattern on the basis of year. So I first know the correct pattern and then use the `super().search` cause it saves me from adding scraper code.

sunpy/net/dataretriever/client.py
abhijeetmanhas: Yes, updated.

abhijeetmanhas:
Things which are yet to be resolved are
- Handling of time in `client.py`.
- Handling of wildcard attrs, optionaldict.

sunpy/net/dataretriever/client.py
```diff
@@ -225,7 +170,57 @@ def _can_handle_query(cls, *query):
...
-        raise NotImplementedError
+        adict = cls.register_values()
+        if cls.optional == set():
+            cls.optional = {k for k in adict.keys()} - cls.required
+        if not cls.check_attr_types_in_query(query, cls.required, cls.optional):
+            return False
+        for key in adict:
+            all_vals = [i[0].lower() for i in adict[key]]
+            for x in query:
+                if isinstance(x, key) and str(x.value).lower() not in all_vals:
```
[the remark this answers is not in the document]
abhijeetmanhas: yes it will fail, the code assumes that the one who will write client won't add wildcard. I will workout how it can be supported here.

sunpy/net/dataretriever/client.py
abhijeetmanhas: Because scraper needs a `TimeRange` for generating urls.

sunpy/net/dataretriever/client.py
abhijeetmanhas: https://github.com/sunpy/sunpy/pull/4321#discussion_r463059042
https://github.com/sunpy/sunpy/pull/4321#discussion_r463059949
this all is for getting end times for a url. If file contains only `year` then end time is defaulted to end of that year.
If file contains only `y/m/d` then endtime is defaulted to end of the day,

It works for all files because none of the file have time in format `%j` or other format. In `SUVICLient` however the endtime was in the filename itself, so `post_search_hook` was overwritten.
All DR clients have daily files, except `SRS` has yearly file. But I think it is a bad assumption. Some scraper changes can be helpful to address it.

sunpy/net/dataretriever/client.py
abhijeetmanhas: Some fields are not present in the `pattern`, like `Source`, `Instrument`, `Provider` etc. So we can't do that way. I'm using matchdict for that work. It uses `register_values` for getting those attrs. The problem is some things may not be even registered as attrs. (Like Time and Wavelength). So we need an individual compare after running loop over the things available from `cls.pattern`.

Yes you are right I shoudn't check only for `Physobs`. Infact I should check for all other attrs too.

sunpy/net/dataretriever/client.py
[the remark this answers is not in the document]
Cadair: This seems flawed. The big change to GOES in particular was us being able to search all the available satellites, but because we are only search for ones which we have listed in `register_values` if a new satellite becomes operational we would need the user to upgrade sunpy before they could get results for that satellite.

This pulls us back to redesigning scraper doesn't it?

sunpy/net/dataretriever/client.py
Cadair: This brings up an interesting thought about how we validate `register_values` and the pattern. If things aren't in the pattern they should have one value in `register_values` and then all the results take that value? If they have more than one value in `register_values` then they should be in the pattern?

Can you not use this logic to determine how to build the results? It would then be much more flexible. I wouldn't expect you to need to call out anything explicitly in this method other than maybe time?

sunpy/net/dataretriever/sources/noaa.py
Cadair: My point  was `self.post_search_hook` should be fine.

changelog/4321.breaking.rst
```diff
@@ -0,0 +1 @@
+Removed :meth:`~sunpy.net.dataretriever.GenericClient._makeargs`, :meth:`~sunpy.net.dataretriever.GenericClient._get_url_for_timerange`, :meth:`~sunpy.net.dataretriever.GenericClient._makeimap`, :meth:`~sunpy.net.dataretriever.GenericClient_get_time_for_url.`, :meth:`~sunpy.net.dataretriever.GenericClient._link`.
```
Cadair, line 1: We don't need a changelog entry for private (`_`) methods.

abhijeetmanhas: Ok, removed.

sunpy/net/dataretriever/sources/noaa.py
abhijeetmanhas: Oh ok, corrected.

sunpy/net/dataretriever/client.py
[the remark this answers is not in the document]
hayesla: I think it's probably ok given that its somewhat of a rare occurrence that there'll be a new GOES satellite with science quality XRS data so I think this is ok to have it 'hard-coded' into the register_values. 

Like the more I think about this - people really will only want just GOES data from one instrument given that they are _supposed_ to be the same data (calibrated to be the same anyway). Those that are interested in using other GOES satellites (e.g. wanting to use GOES 13 when there is GOES 15) will could just use the `SatelliteNumber` attrs. The majority of the community will really just want the lightcurves from the most current instrument at the time of their observation. We could just have a nice example of how to do this etc. 

But as @abhijeetmanhas mentioned in the matrix chat - how would scraper work when the pattern is `../satellite_number/y/m/d...` which is the case for GOES 16 and 17 (rather than just listing all files under `../goes_data/fits/y/m/d..` like for previous data) you would have to register it somewhere times when this data is available and iterate to list the available data in the search right?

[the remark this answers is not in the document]
abhijeetmanhas: yes, because else we won't be able to know the directory names where our scraper would search data for.

sunpy/net/dataretriever/client.py
abhijeetmanhas: Yes, not assuming existence of specific attrs now. Applied that logic in `post_search_hook`

sunpy/net/dataretriever/client.py
```diff
@@ -225,7 +170,57 @@ def _can_handle_query(cls, *query):
...
+            to show results.
+        """
+        map_ = OrderedDict()
+        almost_day = TimeDelta(1 * u.day - 1 * u.millisecond)
+        if 'month' in exdict and 'day' in exdict:
+            start = parse_time("{}/{}/{}".format(exdict['year'], exdict['month'], exdict['day']))
+            end = start + almost_day
+        elif 'year' in exdict:
+            start = parse_time("{}/{}/{}".format(exdict['year'], 1, 1))
+            end = parse_time("{}/{}/{}".format(exdict['year'], 12, 31)) + almost_day
```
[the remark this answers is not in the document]
abhijeetmanhas: https://github.com/sunpy/sunpy/pull/4321#discussion_r464013973

sunpy/net/dataretriever/client.py
abhijeetmanhas: Also we would lose those milliseconds in endtime, if we use stringified time.

sunpy/net/dataretriever/client.py
```diff
@@ -225,7 +170,57 @@ def _can_handle_query(cls, *query):
...
+            and the search query itself.
+
+        Returns
+        -------
+        map_: `~collections.OrderedDict`
+            An Ordered Dictionary which is used by `QueryResponse`
+            to show results.
+        """
+        map_ = OrderedDict()
+        almost_day = TimeDelta(1 * u.day - 1 * u.millisecond)
```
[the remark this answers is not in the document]
abhijeetmanhas: Is it Ok if I add method to get timerange for a file in each client itself?

sunpy/net/dataretriever/client.py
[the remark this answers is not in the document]
abhijeetmanhas: Removed optional by adding Time and Wavelength in `register_values()`.

sunpy/net/dataretriever/client.py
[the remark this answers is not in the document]
abhijeetmanhas: https://github.com/sunpy/sunpy/pull/4321#discussion_r466726547

sunpy/net/dataretriever/client.py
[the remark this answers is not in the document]
Cadair: Conceptually I don't see why the scraper shouldn't be able to handle a pattern like that, list all directories, use the values. Bonus points for checking that the format is something like `{:d2}` or whatever.

sunpy/util/scraper.py
```diff
@@ -137,10 +137,24 @@ def _URL_followsPattern(self, url):
...
+        exdict = parse(self.extractor, url).named
+        datetypes = ['year', 'month', 'day']
+        timetypes = ['hour', 'minute', 'second']
+        dtlist = []
+        for d in datetypes:
+            dtlist.append(exdict.get(d, 1))
+        for t in timetypes:
+            dtlist.append(exdict.get(t, 0))
+        dt = datetime.datetime(*dtlist)
+        return Time(dt)
```
abhijeetmanhas, line 150: This is not a method which follows generalization.

abhijeetmanhas: Much general method added now.

examples/time_series/goes_hek_m25.py
abhijeetmanhas: @hayesla this is fixed now with some tweaks in scraper; by adding :func:`get_timerange_with_extractor` in scraper.

sunpy/net/dataretriever/client.py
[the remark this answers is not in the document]
abhijeetmanhas: This is fixed, we don't have defaulted timerange to one day in `client.py` now.

sunpy/net/dataretriever/sources/rhessi.py
```diff
@@ -151,65 +151,23 @@ def get_observing_summary_dbase_file(time):
...
-        return matches
+    def search(self, *args, **kwargs):
+        baseurl, pattern, matchdict = self.pre_search_hook(*args, **kwargs)
+        timerange = matchdict['Time']
+        metalist = []
+        for url in self.get_observing_summary_filename(timerange):
+            exdict = parse(pattern, url).named
+            exdict['url'] = url
+            map_ = super().post_search_hook(exdict, matchdict)
+            metalist.append(map_)
```
nabobalis, line 162: Is `map_` a good name for this variable?

sunpy/net/dataretriever/sources/noaa.py
```diff
@@ -223,7 +200,7 @@ def fetch(self, qres, path=None, error_callback=None, **kwargs):
             name = url.split('/')[-1]
 
             # temporary fix !!! coz All QRBs have same start_time values
-            day = Time(qre.time.start.strftime('%Y-%m-%d')) + TimeDelta(i*u.day)
+            day = Time(qre['Time'].start.strftime('%Y-%m-%d')) + TimeDelta(i*u.day)
```
nabobalis, line 203: What on earth is this comment?!

sunpy/net/dataretriever/sources/rhessi.py
abhijeetmanhas: Changed to `rowdict` now everywhere.

sunpy/net/dataretriever/sources/noaa.py
abhijeetmanhas: Don't know what are `QRB` (may be queryresponse) here, but it probably mentions about the assumption that all file have `00:00` as start time values, which is used in `fetch` defined differently here to name the downloaded files.

sunpy/net/dataretriever/sources/tests/test_goes_ud.py
abhijeetmanhas: Added many tests with `%H%M` for goes.

.pre-commit-config.yaml
```diff
@@ -66,13 +66,13 @@ repos:
         args: ['--count', '--select', 'E101,E11,E111,E112,E113,E121,E122,E123,E124,E125,E126,E127,E128,E129,E131,E133,E20,E211,E231,E241,E242,E251,E252,E26,E265,E266,E27,E301,E302,E303,E304,E305,E306,E401,E402,E502,E701,E711,E712,E713,E714,E722,E731,E901,E902,F822,F823,W191,W291,W292,W293,W391,W601,W602,W603,W604,W605,W690']
```
nabobalis, line 66: Why are these changes in the diff?

abhijeetmanhas: I messed up in git, let me check.

nabobalis: If you are having issues, it might be better to checkout the old version of this PR. Squash each commit and then rebase it.

abhijeetmanhas: Resolved now.

sunpy/net/dataretriever/sources/tests/test_eve.py
```diff
@@ -134,29 +134,29 @@ def mock_query_object(LCClient):
     """
     # Creating a Query Response Object
     start = '2016/1/1'
-    end = '2016/1/2'
+    end = '2016/1/1 23:59:59'
```
hayesla, line 137: Just out of curiousity - why are all the end times for the tests (incl. ones for other instruments) all changed to 2016/01/01 23:59:59 rather than just 2016/01/02? - is this left over from the problem with scraper before?

hayesla (approved):

hayesla:
This looks great - and the problems with overlapping times and querying times on the same day now is fixed! 🎉

sunpy/net/dataretriever/sources/tests/test_eve.py
abhijeetmanhas: This is query response object to check the end time from the responses returned from table.
Earlier we defaulted it to end date for some clients and enddate - 1 microseconds for some. Now it is consistent to enddate -1 microseconds.

examples/time_series/goes_hek_m25.py
nabobalis: Can we revert this now?

sunpy/net/dataretriever/sources/noaa.py
nabobalis: Can we remove this "temp" fix?

nabobalis:
> This is great!
> 
> I have one issue that needs to be addressed and fixed before this can be approved.
> If for example you search over a timerange during a certain day doesn't return anything -
> for example:
> 
> ```python
> >>> result = Fido.search(a.Time('2013-10-28 01:00', '2013-10-28 03:00'), a.Instrument('XRS')) 
> >>> print(result)
> Results from 1 Provider:
> 
> 0 Results from the XRSClient:
> <No columns>
> ```
> 
> this is a bug that needs to be fixed!

Can we add this as a test case?

nabobalis:
> Things which are yet to be resolved are
> 
> * Handling of time in `client.py`.
> * Handling of wildcard attrs, optionaldict.

Are both of these resolved?

examples/time_series/goes_hek_m25.py
abhijeetmanhas: yes, will revert.

abhijeetmanhas:
> > Things which are yet to be resolved are
> > 
> > * Handling of time in `client.py`.
> > * Handling of wildcard attrs, optionaldict.
> 
> Are both of these resolved?

yes

sunpy/net/dataretriever/sources/noaa.py
abhijeetmanhas: I feel we should keep it, `SRSClient` is exceptional in terms of having its own `fetch()` method, so if there's generalization later done with `fetch`, this comment can be helpful.

abhijeetmanhas:
> > This is great!
> > I have one issue that needs to be addressed and fixed before this can be approved.
> > If for example you search over a timerange during a certain day doesn't return anything -
> > for example:
> > ```python
> > >>> result = Fido.search(a.Time('2013-10-28 01:00', '2013-10-28 03:00'), a.Instrument('XRS')) 
> > >>> print(result)
> > Results from 1 Provider:
> > 
> > 0 Results from the XRSClient:
> > <No columns>
> > ```
> > 
> > 
> > this is a bug that needs to be fixed!
> 
> Can we add this as a test case?

added

nabobalis (approved):

Cadair (approved):
This looks complete, and it's holding up other stuff so let's do it!

PR #4323, opened 2020-06-27
Return Latitude from B0

### LINKED ISSUE #4322: Should coordinates.sun.B0 return a Latitude object?
https://docs.sunpy.org/en/stable/api/sunpy.coordinates.sun.B0.html, returns "the heliographic latitude of the of the center of the disk of the Sun as seen from Earth.". Currently the return type is an `Angle`, and in a few tests both here and in `heliopy` I've had to convert this to a `Latitude` object to compare with other coordinates etc.

So should the return type of `B0` be `Latitude`? I think it should be, but thought I'd open an issue to solicit input before making a PR.

dstansby:
Fixes #4322

Needs:
- [x] Changelog


changelog/4323.trivial.rst
```diff
@@ -0,0 +1,2 @@
+Changed `sunpy.coordinates.sun.B0` return type from `~astropy.coordinates.Angle`
```
ayshih, line 1:
```suggestion
Changed :func:`sunpy.coordinates.sun.B0` return type from `~astropy.coordinates.Angle`
```

dstansby: I don't think this is necessary, the link works fine without: https://sunpy--4323.org.readthedocs.build/en/4323/whatsnew/changelog.html

ayshih: It adds the parentheses

Cadair:
Is this really a breaking change given `Latitude` is a subclass of `Angle`? I suppose as a choice between "trivial" and "breaking" it's maybe better to land on breaking...

ayshih:
It'd be pathological, but someone could have code out there that depends on the output to *not* be of a `Latitude` class.

PR #4326, opened 2020-06-29
Use default mpl colour cycle for RHESSI peek

dstansby:
- Reduces dependence of `timeseries` on `instr`
- Previously hard coding the colors prevented the user handing over the `colors` kwarg
- It didn't look very nice:
sunpy timeseries tests test_timeseriesbase test_rhessi_peek

Would be good to get a review from someone who uses RHESSI data, in case these are commonly used colours associated with this data and they should stay - I suspect they aren't though.

Also remove hardcoded linewidth, so the user can specify it.

### DIFF
--- a/changelog/4326.feature.rst
+++ b/changelog/4326.feature.rst
@@ -1,0 +1,7 @@
+`sunpy.timeseries.RHESSISummaryTimeSeries.peek()` has had the following minor
+changes:
+
+- The default matplotlib color cycle is now used
+- The default matplotlib linewidth is now used
+- It is now possible to pass in a user specified linewidth
+- Seconds have been added to the x-axis labels (previously it was just hours and minutes)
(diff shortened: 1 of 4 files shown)


hayesla:
I think these are just the colors from the RHESSI GUI (see below)

Screen Shot 2020-06-29 at 9 58 32 AM


But I don't think they are necessarily used in paper publications etc. I vote to change it to a default matplotlib table 👍 

also there's something funky going on with the time axis on the plot ....

dstansby:
Oh yeah, I'll try and fix the time axis too.

Cadair:
:+1: much nicer.

Cadair:
huh, is codecov broken on the figure tests?

ayshih:
Erm.  I'm not a fan *at all* of changing the color cycle.  The prior color cycle is the standard RHESSI color cycle (via the procedure `hsi_linecolors`).  Here's one of our regular data products:
hsi_20120531_234540_corrected_rate

@hayesla is not incorrect; people don't always use `hsi_linecolors` when making their own plots for publication.  However, I feel that it's enough of a standard that it's a bad idea to change to a color cycle that is unfamiliar to most RHESSI folks.

PR #4328, opened 2020-06-30
Set v{min,max} on the norm when plotting

dstansby:
This is in preparation for a deprecation warning in Matplotlib 3.3, which means that a norm and v{min, max} cannot be simulataneously handed to `imshow`. See https://app.circleci.com/pipelines/github/sunpy/sunpy/6725/workflows/45cad83c-7349-4cc1-93c9-796bdfabc1a6/jobs/55990/steps for the actual warning.

In addition, this raises an error if the user hands v{min,max}, but it is already set on the norm. This is breaking behavior without a depreaction warning, but I think is much better than undefined/multiple v{min,max} values. Happy to turn into a deprecation warning though if that's better.


nabobalis:
Worth adding a unit test to check for the ValueError being rasied?

PR #4330, opened 2020-06-30
Change rhessi colors

dstansby:
xref https://github.com/sunpy/sunpy/pull/4326

Before #4326:
sunpy timeseries tests test_timeseriesbase test_rhessi_peek-1
With this PR:
sunpy timeseries tests test_timeseriesbase test_rhessi_peek-2

### DIFF
--- a/sunpy/timeseries/sources/rhessi.py
+++ b/sunpy/timeseries/sources/rhessi.py
@@ -82,12 +82,18 @@ class RHESSISummaryTimeSeries(GenericTimeSeries):
         # Check we have a timeseries valid for plotting
         self._validate_data_for_plotting()
 
+        # These are a matplotlib version of the default RHESSI color cycle
+        default_colors = ('black', 'tab:pink', 'tab:green', 'tab:cyan',
+                          'tab:olive', 'tab:red', 'tab:blue', 'tab:orange',
+                          'tab:brown')
+        colors = kwargs.pop('colors', default_colors)
+
         figure = plt.figure()
         axes = plt.gca()
 
-        for item, frame in self.to_dataframe().items():
-            axes.plot_date(self.to_dataframe().index, frame.values, '-',
-                           label=item, **kwargs)
+        for color, (item, frame) in zip(colors, self.to_dataframe().items()):
+            axes.plot(self.to_dataframe().index, frame.values,
+                      color=color, label=item, **kwargs)
 
         axes.set_yscale("log")
         axes.set_xlabel(datetime.datetime.isoformat(self.to_dataframe().index[0])[0:10])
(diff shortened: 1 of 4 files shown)


sunpy/instr/rhessi.py
```diff
@@ -183,8 +183,9 @@ def hsi_linecolors():
     ----------
     `hsi_linecolors.pro <https://hesperia.gsfc.nasa.gov/ssw/hessi/idl/gen/hsi_linecolors.pro>`__
     """
-    return ('black', 'magenta', 'lime', 'cyan', 'y', 'red', 'blue', 'orange',
-            'olive')
+    return ('black', 'tab:pink', 'tab:green', 'tab:cyan',
```
Cadair, line 186: I think we should probably add this as a new cycle so people can opt back into the old one?

dstansby: Should I add it to `timeseries` to avoid adding anything new to `instr`? I think ideally any plot related things should be under `timeseries`, in a similar way to how the e.g. aia colormap is in `map` and not `instr`.

Cadair: yeah I am in favour of that.

sunpy/timeseries/sources/rhessi.py
```diff
@@ -76,18 +76,26 @@ def peek(self, title="RHESSI Observing Summary Count Rate", **kwargs):
...
+                          'tab:brown')
+        colors = kwargs.pop('colors', default_colors)
+
         figure = plt.figure()
         axes = plt.gca()
 
-        for item, frame in self.to_dataframe().items():
-            axes.plot_date(self.to_dataframe().index, frame.values, '-',
-                           label=item, **kwargs)
+        for color, (item, frame) in zip(colors, self.to_dataframe().items()):
```
Cadair, line 96: What's the faliure mode look like here if colors is the wrong length?

Cadair:
Generally looks good to me.

sunpy/timeseries/sources/rhessi.py
dstansby: If colors is too long it will just use the first *n* colors, where *n* is the number of lines in the normal plot.
If colors is too short it will just plot the first *m* lines, where *m* is the number of colors. Should probably change that, but I'm not sure what to? Maybe cycling through the colors provided?

Cadair: yeah cycling sounds good.

nabobalis: Once this change is done, we can merge this?

dstansby: I think so - can someone leave me a "needs changes" review to block the PR and remind me to do this?

nabobalis:
Not sure what happened that made the CI break.

PR #4338, opened 2020-07-07
added deprecation to map.size property

### LINKED ISSUE #4337: Deprecate map.size
Given the discussion in #4202, we should deprecate the `size` attribute on `Map` for consistency as we do not want to add a `shape` attribute for the reasons listed in that issue.

dipanshu231099:
### Description

Added the required deprecation for the map.size property to use map.data.shape instead.

Fixes #4337

### DIFF
--- a/changelog/4338.removal.rst
+++ b/changelog/4338.removal.rst
@@ -1,2 +1,2 @@
-Depreacted the use of `sunpy.map.Map.size` property
+Deprecated the use of `sunpy.map.Map.size` property
 with suggested use of `sunpy.map.Map.data.size` instead.
--- a/changelog/4338.removal.rst
+++ b/changelog/4338.removal.rst
@@ -1,0 +1,2 @@
+depreacted the use of `sunpy.map.Map.size` property
+with suggested use of `sunpy.map.Map.data.size` instead.
--- a/changelog/4338.removal.rst
+++ b/changelog/4338.removal.rst
@@ -1,2 +1,2 @@
-depreacted the use of `sunpy.map.Map.size` property
+Depreacted the use of `sunpy.map.Map.size` property
 with suggested use of `sunpy.map.Map.data.size` instead.
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -531,7 +531,7 @@ class GenericMap(NDData):
         return self.data.dtype
 
     @property
-    @deprecated(since="2.1", message="Use map.data.shape instead", alternative="map.data.shape")
+    @deprecated(since="2.1", message="Use map.data.size instead", alternative="map.data.size")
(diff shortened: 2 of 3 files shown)


nabobalis:
You will need to update:

```
FAILED ..\..\.tox\py37\lib\site-packages\sunpy\map\tests\test_mapbase.py::test_size - sunpy.util.exceptions.SunpyDeprecationWarning: Use map.data.shape instead
```
to catch and test for that exception.

dstansby:
I don't think we should backport deprecations, so I've removed the label. @dipanshu231099 let us know if you need any help updating the tests!

nabobalis:
> I don't think we should backport deprecations, so I've removed the label. @dipanshu231099 let us know if you need any help updating the tests!

I think we should so any people on the LTS will be aware of it and since it does have an alternative. If it did not have one, I would be against it.

Cadair:
We wont be able to remove it until 3.1 but I don't think we should backport deprecations. Things which don't make the cut have to wait.

nabobalis:
Ok, does this get a breaking changelog?

Cadair:
No deprecations and removal of deprecated features go in `removal` according to the changelog readme.

sunpy/map/mapbase.py
```diff
@@ -531,6 +531,7 @@ def dtype(self):
         return self.data.dtype
 
     @property
+    @deprecated(since="2.1", message="Use map.data.shape instead", alternative="map.data.shape")
```
Cadair, line 534:
```suggestion
    @deprecated(since="2.1", message="Use map.data.size instead", alternative="map.data.size")
```

dipanshu231099: I guess the issue was to use map.data.shape instead of map.size property? or is it map.data.size?

dipanshu231099:
@nabobalis the test to catch exception has to be done for the map.size property right ?

nabobalis:
> @nabobalis the test to catch exception has to be done for the map.size property right ?

Yes.

changelog/4338.removal.rst
```diff
@@ -0,0 +1,2 @@
+depreacted the use of `sunpy.map.Map.size` property
```
nabobalis, line 1:
```suggestion
Depreacted the use of `sunpy.map.Map.size` property
```

changelog/4338.removal.rst
```diff
@@ -0,0 +1,2 @@
+Depreacted the use of `sunpy.map.Map.size` property
```
nabobalis, line 1:
```suggestion
Deprecated the use of `sunpy.map.Map.size` property
```

nabobalis:
Thanks @dipanshu231099

PR #4343, opened 2020-07-10
Always use the same ERFA package used by Astropy

dstansby:
Edit by @ayshih: This PR underwent a few significant variations, so the following discussion may be confusing when compared to the actual changes.

This fixes deprecation warnings like

```python
pfsspy/__init__.py:12: in <module>

../../../miniconda/envs/testenv/lib/python3.8/site-packages/sunpy/coordinates/__init__.py:21: in <module>

    from . import sun

../../../miniconda/envs/testenv/lib/python3.8/site-packages/sunpy/coordinates/sun.py:7: in <module>

    from astropy import _erfa as erfa

../../../miniconda/envs/testenv/lib/python3.8/site-packages/astropy/_erfa/__init__.py:13: in <module>

    warnings.warn('The private astropy._erfa module has been made into its '

E   astropy.utils.exceptions.AstropyDeprecationWarning: The private astropy._erfa module has been made into its own package, pyerfa, 
```

Not sure if it wants to be backported?

### DIFF
--- a/sunpy/coordinates/transformations.py
+++ b/sunpy/coordinates/transformations.py
@@ -20,7 +20,6 @@ from contextlib import contextmanager
 import numpy as np
 
 import astropy.units as u
-from astropy._erfa import obl06
 from astropy.constants import c as speed_of_light
 from astropy.coordinates import (
     HCRS,
@@ -41,6 +40,8 @@ from astropy.coordinates.representation import (
     SphericalRepresentation,
     UnitSphericalRepresentation,
 )
+# Import erfa via astropy to make sure we are using the same ERFA library as Astropy
+from astropy.coordinates.sky_coordinate import erfa
 from astropy.coordinates.transformations import (
     AffineTransform,
     FunctionTransform,
@@ -889,7 +890,7 @@ def _rotation_matrix_obliquity(time):
     """
     Return the rotation matrix from Earth equatorial to ecliptic coordinates
     """
-    return rotation_matrix(obl06(*get_jd12(time, 'tt'))*u.radian, 'x')
+    return rotation_matrix(erfa.obl06(*get_jd12(time, 'tt'))*u.radian, 'x')
 
 
 @frame_transform_graph.transform(FunctionTransformWithFiniteDifference,
--- a/setup.cfg
+++ b/setup.cfg
@@ -124,8 +124,6 @@ filterwarnings =
     ignore:direct construction of AsdfSchemaExampleItem has been deprecated
     # See https://github.com/scikit-image/scikit-image/issues/4848
     ignore:Converting `np.inexact` or `np.floating` to a dtype is deprecated
-    # See https://github.com/sunpy/sunpy/pull/4343
-    ignore:The private astropy._erfa module has been made into its own package
     # See https://github.com/astropy/astropy/pull/10570
     ignore:leap-second auto-update failed
     # See https://github.com/astropy/extension-helpers/issues/23
--- a/sunpy/coordinates/sun.py
+++ b/sunpy/coordinates/sun.py
@@ -4,7 +4,6 @@ Sun-specific coordinate calculations
 import numpy as np
 
 import astropy.units as u
-from astropy import _erfa as erfa
 from astropy.constants import c as speed_of_light
 from astropy.coordinates import (
     ITRS,
@@ -20,6 +19,8 @@ from astropy.coordinates import (
 )
 from astropy.coordinates.builtin_frames.utils import get_jd12
 from astropy.coordinates.representation import CartesianRepresentation, SphericalRepresentation
+# Import erfa via astropy to make sure we are using the same ERFA library as Astropy
+from astropy.coordinates.sky_coordinate import erfa
 from astropy.time import Time
 
 from sunpy import log


ayshih:
I've changed this PR to "draft" status because the issue needs more deliberation.  `pyerfa` doesn't become a dependency of Astropy until Astropy 4.2, so the tests fail because nothing is triggering `pyerfa` to be installed.  Although our policy hasn't been formally established (see #3918), I believe our intention is to stick with Astropy 4.0 as our dependency through the next release.

An alternative solution is for SunPy to have an explicit dependency on `pyerfa`, instead of an indirect one through Astropy.  I believe @Cadair thinks that that is the way to go.  However, I worry about the small danger that a user could find themselves using a version of Astropy that has the embedded ERFA while SunPy is using `pyerfa`, and there could be slight differences in the libraries.

dstansby:
I did think about doing something like
```python
try:
    import erfa
except ImportError:
    import astropy._erfa as erfa
```
Are there versions of `erfa` it is desirable to pin to? As far as I can tell at the moment it's just whatever version astropy ships with.

nabobalis:
If we did bump to astropy 4.2 for 2.1 would that make this change easier?

ayshih:
> Are there versions of `erfa` it is desirable to pin to? As far as I can tell at the moment it's just whatever version astropy ships with.

My position is that I want `sunpy.coordinates` to be using the same ERFA as the users's `astropy.coordinates` is using.  If the user is running Astropy 4.2, then it's straightforward: use the system's `pyerfa`.  If the user is running Astropy 4.1 or earlier, it should be the ERFA that ships within Astropy, which means *disregarding* `pyerfa` even if it is installed.

> If we did bump to astropy 4.2 for 2.1 would that make this change easier?

Even though our Astropy-dependency policy isn't established, I don't think we can sanely go from SunPy 2.0 depending on Astropy 3.2 to SunPy 2.1 depending on Astropy 4.2 (which isn't even released yet).

nabobalis:
> Even though our Astropy-dependency policy isn't established, I don't think we can sanely go from SunPy 2.0 depending on Astropy 3.2 to SunPy 2.1 depending on Astropy 4.2 (which isn't even released yet).

I thought that 2.0 depended on 4.0 or was that 2.1?

dstansby:
I've changed it to a conditional import on astropy version
- Maybe there's a cleverer way than checking the astropy version?
- Not sure if this is the best way to check for the astropy version
- Not sure if `sunpy.util.sysinfo` is the best place to put the check

nabobalis:
What if we just do:

```python
try:
    import erfa
except ImportError:
    from astropy import _erfa as erfa
```
at the top of both files instead of a check in "sunpy.util.sysinfo"?

Cadair:
Yeah I think that makes more sense than checking astropy version.

dstansby:
That's different to what @ayshih said earlier:

> My position is that I want sunpy.coordinates to be using the same ERFA as the users's astropy.coordinates is using. If the user is running Astropy 4.2, then it's straightforward: use the system's pyerfa. If the user is running Astropy 4.1 or earlier, it should be the ERFA that ships within Astropy, which means disregarding pyerfa even if it is installed.

In order to not use `pyerfa` in astropy<4.2 I think there has to be a version check somewhere.

ayshih:
The "elegant" solution – which I believe unsettles @Cadair – is to import *through* Astropy.  That is, simply replace
```python
from astropy import _erfa as erfa
```
with
```python
from astropy.coordinates.sky_coordinate import erfa
```
Thus, SunPy will be guaranteed to use the same ERFA library that Astropy is using.  No version checking, no logic branches, etc.  However, it means that the implied dependency on `pyerfa` is completely hidden when looking at the SunPy codebase.

nabobalis:
I am ok with that.

Cadair:
:nauseated_face: I guess I am ok with it as a horrific hack until we drop support for astropy 4.1

Cadair:
Actually, hang on. If we are doing that why not just import the hidden `_erfa` package still and squash the warning?

ayshih:
> Actually, hang on. If we are doing that why not just import the hidden `_erfa` package still and squash the warning?

It's plausible that we will still be supporting Astropy 4.1 when the latest development version of Astropy removes the hidden `_erfa` package.  So, if we just squash a potential deprecation warning, we'd also need to `try`...`except` the import to fallback on importing `pyerfa` for such a development version.

Or we can just import through Astropy.  I disagree that it's a "horrific hack".  It accurately represents the fact that our coordinates framework should use the *exact* ERFA library that Astropy coordinates is using, regardless of the package environment.

If you want a horrific hack, the ERFA imports could be removed entirely by back-deriving the necessary values (e.g., obliquity, nutation) from transformation matrices that Astropy calculates...

sunpy/coordinates/transformations.py
```diff
@@ -41,6 +40,7 @@
     SphericalRepresentation,
     UnitSphericalRepresentation,
 )
+from astropy.coordinates.sky_coordinate.erfa import obl06
```
ayshih, line 43: I believe what you actually need to do is:
```suggestion
from astropy.coordinates.sky_coordinate import erfa
```
and then change the call on line 892 from `obl06()` to `erfa.obl06()`.

nabobalis (approved):

PR #4353, opened 2020-07-13
Fix some GOES warnings

dstansby:
Split out of #3648 for easier review. This removes a bunch of GOES related warnings in the tests (and one in actual `goes.py`).


nabobalis:
Doc failure is a sample data issue and conda failed due to `ERROR ../../docs/whatsnew/1.0.rst::1.0.rst - ValueError: invalid version number '6.0.0rc1` I would say its safe to ignore and fixed in another PR.

PR #4354, opened 2020-07-13
Fix various depreaction warnings

dstansby:
Split out of #3648 for easier review. These are all minor code changes to avoid raising warnings during the running of actual code. If you are reviewing this and don't understand some of the changes, let me know and I can post more info.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -286,7 +286,7 @@ class GenericMap(NDData):
             bad_pixel_text += ", ".join(text_list)
 
         # Use a grayscale colormap with histogram equalization (and red for bad values)
-        cmap = cm.get_cmap('gray')
+        cmap = copy.copy(cm.get_cmap('gray'))
         cmap.set_bad(color='red')
         norm = ImageNormalize(stretch=HistEqStretch(finite_data))
 
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -286,6 +286,8 @@ class GenericMap(NDData):
             bad_pixel_text += ", ".join(text_list)
 
         # Use a grayscale colormap with histogram equalization (and red for bad values)
+        # Make a copy of the colormap to avoid modifying the matplotlib instance when
+        # doing set_bad()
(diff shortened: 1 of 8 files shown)


sunpy/map/mapbase.py
```diff
@@ -286,7 +286,7 @@ def _repr_html_(self):
             bad_pixel_text += ", ".join(text_list)
 
         # Use a grayscale colormap with histogram equalization (and red for bad values)
-        cmap = cm.get_cmap('gray')
+        cmap = copy.copy(cm.get_cmap('gray'))
```
nabobalis, line 289: We copy to prevent modification?

sunpy/net/jsoc/jsoc.py
```diff
@@ -742,7 +742,7 @@ def _make_recordset(self, series, start_time='', end_time='', wavelength='',
             else:
                 timestr = ''
 
-        if wavelength:
+        if wavelength != '':
```
nabobalis, line 745: What was the issue here?

sunpy/tests/tests/test_self_test.py
```diff
@@ -1,4 +1,4 @@
-import imp
```
nabobalis, line 1: This was an import?!

sunpy/map/mapbase.py
dstansby: This is because we do `.set_bad` on the line below, which without copying would modify the default copy of the colormap in matplotlib, which raises a warning now. I'll add a comment to the code saying why we copy.

sunpy/net/jsoc/jsoc.py
dstansby: If wavelength is a Quantity, `if wavelength:` raises a warning, and will eventually fail, as in the future `Quantity` can't be treated like a bool.

nabobalis:
Looks like an isort issue from `python_codestyle`

nabobalis:
Online failures are due to JSOC 500 errors, I have restarted the build.

nabobalis:
Less failed this time. So thats good?

PR #4355, opened 2020-07-13
Filter out some warnings from tests

dstansby:
More stuff pulled out of #3648, this time ignoring warnings. This is stuff that I understand well and have commented the reasons for ignoring the various warnings. The one thing that isn't explicitly commented is setting up new clients in the tests, which has to be done as `hypothesis` does not automatically reset fixtures.

Again, if anything does not make sense please let me know!

### DIFF
--- a/sunpy/coordinates/tests/test_ephemeris.py
+++ b/sunpy/coordinates/tests/test_ephemeris.py
@@ -126,7 +126,7 @@ def test_get_horizons_coord_array_time():
 
 # Ignore the warning hypothesis throws, as in this case we don't care that
 # the fixutre isn't reset bewteen tests
-@pytest.mark.filterwarnings('ignore:.*function-scoped fixtures should not be used with.*')
+@pytest.mark.filterwarnings('ignore:.*which is reset between function calls but not between test cases.*')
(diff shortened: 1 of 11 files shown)


sunpy/coordinates/tests/test_ephemeris.py
```diff
@@ -124,6 +124,9 @@ def test_get_horizons_coord_array_time():
     assert_quantity_allclose(e[3].radius, 0.9908173*u.AU, atol=5e-7*u.AU)
 
 
+# Ignore the warning hypothesis throws, as in this case we don't care that
+# the fixutre isn't reset bewteen tests
```
nabobalis, line 128:
```suggestion
# the fixture isn't reset between tests
```

nabobalis:
I am not 100% sure about any cooridainte or frame warnings we skip over. @ayshih can you double check please?

ayshih:
> @ayshih can you double check please?

Coordinate stuff looks good to me

nabobalis:
> > @ayshih can you double check please?
> 
> Coordinate stuff looks good to me

Thanks.

PR #4357, opened 2020-07-14
added functionality to check if ther server is active or not

### LINKED ISSUE #4243: Skip remote tests for HELIO if server is unreachable
### Description

There should be a way to skip all remote tests in helio after checking if server is reachable or not.

### Additional context

This is useful https://docs.pytest.org/en/latest/skipping.html#id1 . We can prior check the server state and use `@pytest.skipif(condition)`

dipanshu231099:
### Description

Added the functionality to check whether client so requested is not down at the moment the test is run.

Fixes #4243

### DIFF
--- a/sunpy/net/helio/tests/test_helio.py
+++ b/sunpy/net/helio/tests/test_helio.py
@@ -266,9 +266,12 @@ def test_link_test_on_urlerror(mock_link_test):
 @pytest.mark.remote_data
 @pytest.fixture(scope="session")
 def client():
-    working_links = list(filter(link_test, webservice_parser()))
-    taverna_link = taverna_parser(working_links[0])[0]
-    return HECClient(taverna_link)
+    try:
+        client = HECClient()
+        return client
+    # If no links are found, the client should raise a ValueError
+    except ValueError:
+        pytest.xfail("No HELIO working links found.")
 
 
 @pytest.mark.remote_data
@@ -298,3 +301,10 @@ def test_time_query(client):
     table_name = b'rhessi_hxr_flare'
     res = client.time_query(start, end, table=table_name, max_records=10)
     assert len(res.array) == 10
+
+
+def test_client_mock_fail(client):
+    """
+    This test will xfail on the offline tests.
+    It just passes if the client is working.
+    """
--- a/sunpy/net/helio/parser.py
+++ b/sunpy/net/helio/parser.py
@@ -215,7 +215,8 @@ def wsdl_retriever(service='HEC'):
         this function to take a while to return. Timeout duration can be
         controlled through the LINK_TIMEOUT value
     """
-    def fail(): raise ValueError("No online HELIO servers can be found.")
+    def fail():
+        raise ValueError("No online HELIO servers can be found.")
     service_links = webservice_parser(service=service)
     wsdl = None
     wsdl_links = None


sunpy/net/helio/tests/test_helio.py
```diff
@@ -259,16 +260,39 @@ def test_link_test_on_urlerror(mock_link_test):
...
     working_links = list(filter(link_test, webservice_parser()))
     taverna_link = taverna_parser(working_links[0])[0]
-    return HECClient(taverna_link)
+    # first check if the server is active
+    ip, port = taverna_link.split('/')[2].split(':')
+    server_status = isOpen(ip, port, 3)
+    if(server_status):
+        return HECClient(taverna_link)
+    else:
+        return None
```
nabobalis, line 288: If this return was just `pytest.xfail` would that work?

dipanshu231099: I guess pytest.xfail has to be called for test functions so that those funcitons depending on its data do not fail. because client is not failing in any way its just not catching any data which could be normal... while its a error for those relying on this data

sunpy/net/helio/tests/test_helio.py
```diff
@@ -259,16 +260,39 @@ def test_link_test_on_urlerror(mock_link_test):
...
+    except socket.gaierror:
+        return False
+    finally:
+        s.close()
+
+
 @pytest.mark.remote_data
 @pytest.fixture(scope="session")
 def client():
     working_links = list(filter(link_test, webservice_parser()))
```
nabobalis, line 280: In theory this link_test should already filter out any bad URLS. I wonder if we need to check that works and then if this returns nothing we pytest.xfail?

sunpy/net/helio/tests/test_helio.py
nabobalis: If you did `pytest.xfail("server broke")` would that not work instead of returning None?

sunpy/net/helio/tests/test_helio.py
dipanshu231099: Maybe it passes the urls whose servers are down at the moment, if I am not wrong i guess that was the issue that server when down thus resulted in failure. isOpen ensures that not be the issue.

nabobalis: Then we could fix `link_test` instead of creating a new function?

dipanshu231099: so shall i implement this function in link_test if you want. or if we wish to keep this   way also, i mean link_test will test the link while isOpen will check if the server is up. I will update it accordingly if you recommend.

nabobalis: I don't understand why we need another function to do the test? Why can't link_test do it all?

dipanshu231099: okay i got your point... test_link itself can do check this functionality thanks for pointing me out i will remove this redundant function then

nabobalis:
Can you also come up with a test to make sure it works?

dipanshu231099:
> Can you also come up with a test to make sure it works?

a test for link_test when the url passed is not valid would that be fine?

nabobalis:
Sure

sunpy/net/helio/tests/test_helio.py
```diff
@@ -263,12 +263,18 @@ def test_link_test_on_urlerror(mock_link_test):
...
+        taverna_link = taverna_parser(working_links[0])[0]
+    except IndexError:
+        return None
     return HECClient(taverna_link)
 
 
 @pytest.mark.remote_data
 def test_get_table_names(client):
+    # xfail the test if the server is down
+    if client is None:
```
nabobalis, line 276: I would rather that it gets xfailed by the figure instead of checking this in the test.

dipanshu231099: sure sir, i will do the required changes

nabobalis: Please do not call me sir.

dipanshu231099:
@nabobalis as for the test i guess `test_link_test_on_urlerror` and `test_link_test_on_valueerror` also ensure that bad url or when server is down None is returned by `link_test`.

nabobalis:
> @nabobalis as for the test i guess `test_link_test_on_urlerror` and `test_link_test_on_valueerror` also ensure that bad url or when server is down None is returned by `link_test`.

You mean those tests already check for the server being down?

dipanshu231099:
> > @nabobalis as for the test i guess `test_link_test_on_urlerror` and `test_link_test_on_valueerror` also ensure that bad url or when server is down None is returned by `link_test`.
> 
> You mean those tests already check for the server being down?

Since a down server may not be permanantly detected ( I am guessing so). Hence I tried passing a bad url in that case the tests already present are already doing the job of checking a bad url

nabobalis:
Sounds good.

sunpy/net/helio/tests/test_helio.py
```diff
@@ -260,11 +260,14 @@ def test_link_test_on_urlerror(mock_link_test):
 
 
 @pytest.mark.remote_data
-@pytest.fixture(scope="session")
+@pytest.fixture
 def client():
-    working_links = list(filter(link_test, webservice_parser()))
-    taverna_link = taverna_parser(working_links[0])[0]
-    return HECClient(taverna_link)
```
nabobalis, line 267: Looking at it, the client itself should handle this already. If it does not, its a bug in the client.

sunpy/net/helio/tests/test_helio.py
```diff
@@ -291,3 +294,10 @@ def test_time_query(client):
     table_name = b'rhessi_hxr_flare'
     res = client.time_query(start, end, table=table_name, max_records=10)
     assert len(res.array) == 10
+
+
+@mock.patch('sunpy.net.helio.parser.webservice_parser', return_value=None)
+def test_client_mock_fail(mock_parser, client):
+    """
+    This test will always xfail.
+    """
```
nabobalis, line 303: Dummy test that always XFAILS.

nabobalis:
Thanks @dipanshu231099

Issue #4346, opened 2020-07-11
Add support for extra NOAA JSON timeseries


nabobalis:
From https://github.com/sunpy/sunpy/pull/4340#issuecomment-656253760:

> I wonder if it would be an interesting feature/extension for the nooa TimeSeries class, to can handle also other json files (with t.s.) from https://services.swpc.noaa.gov/json/. For example https://services.swpc.noaa.gov/json/goes/ (/primary/xrays-1-day.json) contain json timeseries for soft x-rays provided in real time (contrary to the goes.py class where the fits files are available after 1-2 days). I think this may be something interesting for the sunpy user to have. I attach a simple example of how I process and plot this xrays-1-day.json file, similarly it works for for the other xrays-?-day.json files. Something similar could be tailored to the noaa class, right?
> [GOES_JSON_FLARE.zip](https://github.com/sunpy/sunpy/files/4898400/GOES_JSON_FLARE.zip)

AthKouloumvakos:
@nabobalis thank you for opening this issue. I was trying to see some details for the products provided and information detailed by the https://services.swpc.noaa.gov/json/. So for the **GOES json files** that are provided under the /primary/ or /secondary/ sub-directories, the TimeSeries class could probably handle the following products:

1. **GOES json files (main-url: https://services.swpc.noaa.gov/json/goes/):**

    -  **GOES X-ray Flux** (more info at: https://www.swpc.noaa.gov/products/goes-x-ray-flux)
        **X-ray Flux**, 1-minute averaged: _xrays-1-day.json, xrays-3-day.json, xrays-6-hour.json, xrays-7-day.json_ (note: the files are for different time spans, i.e. now-?days/hours)

    - **GOES Proton Flux** (more info at: https://www.swpc.noaa.gov/products/goes-proton-flux)
[...]

PR #4364, opened 2020-07-15
Improve goes peek()

dstansby:
This

- Fixes goes `peek()` on Matplotlib dev
- Puts the labels in the center of the flare classes instead of the boundaries, as in https://www.swpc.noaa.gov/products/goes-x-ray-flux

Here's the new plot:

sunpy timeseries tests test_timeseriesbase test_goes_peek

### DIFF
--- a/changelog/4364.bugfix.rst
+++ b/changelog/4364.bugfix.rst
@@ -1,0 +1,2 @@
+The flare class labels in GOES ``peek()`` plots are now drawn at the center of
+the flare classes. Previously they were (ambiguously) drawn on the boundaries.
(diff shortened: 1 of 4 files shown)


dstansby:
Would like to get a second review on this to double check before it's merged.

AthKouloumvakos:
Looks good! I will also revise this part for #4346.
I wonder if a: `axes.get_shared_x_axes().join(axes, ax2)` is needed after the ax2 creation.
So that after the figure is created and if someone needs to adjust the ylims of the left axis 
the update to pass automatically to right axis also and the labels to be synchronized.
Is that right?

nabobalis:
I am not sure we let users adjust peek that much. If you want a figure to display with custom limits, we normally suggest you use plot.

dstansby:
I checked, and as it stands the axes are linked so they move together.

PR #4366, opened 2020-07-16
More warnings

dstansby:
More stuff from #3648. Again, all should be self-explanatory or have comments. Again, let me know if anything doesn't make sense.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -100,6 +100,10 @@ filterwarnings =
     # add a comment or ideally a link to an issue that explains why the warning
     # is being ignored
     #
+    # Several timeseries data products have nanoseconds, that python/pandas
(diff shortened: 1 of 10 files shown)


setup.cfg
```diff
@@ -90,6 +90,8 @@ markers =
     flaky
     array_compare
 remote_data_strict = True
+# Pin junit behaviour; we might want to update this to xunit2 at some point
```
nabobalis, line 93: What's the difference?

nabobalis:
That seems like more failures than normal.

dstansby:
Woops, that was a missing warning. Figure test failures are due to upstream matplotlib changes, will open another PR for them....

PR #4368, opened 2020-07-16
Linux OpenBLAS test

ayshih:
Previously the `libopenblas` feedstock for conda-forge provided a build for Linux using OpenMP threads.  Yesterday, they added a build for Linux that uses pthreads threads instead, and that is the new default for conda solutions.  We're getting severe test failures when using the pthreads build, so this PR tests whether we get test failures if we explicitly specify the OpenMP build.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -122,6 +122,7 @@ conda_deps =
     lxml
     matplotlib
     numpy
+    libopenblas=*=*openmp*
     openjpeg
     pandas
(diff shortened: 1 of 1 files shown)


ayshih:
Conclusions:
* All varieties of 0.3.10 Linux builds fail
* 0.3.9 Linux build passes

PR #4369 is a clean PR for the pin

PR #4372, opened 2020-07-16
added contains example to timerange class doc string

### LINKED ISSUE #3789: Document t in TimeRange

nabobalis:
Fixes #3789

### DIFF
--- a/sunpy/time/timerange.py
+++ b/sunpy/time/timerange.py
@@ -26,6 +26,10 @@ class TimeRange:
        Regardless of how a `sunpy.time.TimeRange` is constructed it will always
        provide a positive time range where the start time is before the end time.
 
+       `~sunpy.time.TimeRange.__contains__` has been implemented which means you can
+       check if a time is within the time range you have created.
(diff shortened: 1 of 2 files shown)


sunpy/time/timerange.py
```diff
@@ -26,6 +26,10 @@ class TimeRange:
        Regardless of how a `sunpy.time.TimeRange` is constructed it will always
        provide a positive time range where the start time is before the end time.
 
+       `~sunpy.time.TimeRange.__contains__` has been implented which means you can
```
dstansby, line 29:
```suggestion
       `~sunpy.time.TimeRange.__contains__` has been implemented which means you can
```

nabobalis: Damn it.

nabobalis: Fixed. This is what I get for writing english.

PR #4374, opened 2020-07-16
Custom SunpyMetadataWarning

### LINKED ISSUE #4008: Missing metadata warnings aren't very helpful
I'm currently working on some code the raises a lot of Missing metadata warnings, that show up like this
```
WARNING: SunpyUserWarning: Missing metadata for observer: assuming Earth-based observer.
 [sunpy.map.mapbase]
```
This isn't very helpful though, as I don't know what line in my code, or object, is the cause of these warnings.

It would be good to add some extra information so the user knows what is causing the warnings, and therefore making them easier to diagnose and fix.

nabobalis:
Fixes #4008

By not using `AstropyWarning` we restore the information on where the warning is raised which is useful for debugging. The logger from astropy hides this and there isn't an easy way to change this.

If Astropy want to show a warning location, they do not inherit from `AstropyWarning` and create a custom warning. So we could create a custom metadata warning that handles this?

### DIFF
--- a/sunpy/util/exceptions.py
+++ b/sunpy/util/exceptions.py
@@ -8,7 +8,7 @@ but rather in the particular package.
 from astropy.utils.exceptions import AstropyWarning
 
 __all__ = ["SunpyWarning", "SunpyUserWarning", "SunpyDeprecationWarning",
-           "SunpyPendingDeprecationWarning"]
+           "SunpyPendingDeprecationWarning", "SunpyMetadataWarning"]
 
 
 class SunpyWarning(AstropyWarning):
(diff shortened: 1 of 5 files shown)


nabobalis:
> LGTM but why is `SunpyMetadataWarning` in `mapbase.py`? Shouldn't it go in the util module somewhere (i.e. where we've already defined `SunpyUserWarning`)?

I thought I would leave in mapbase since its a warning we might not want to repeat outside of that file. I can move it, I'm not too fussed where it lives.

Cadair:
I think this should be in `utils/exceptions.py` but otherwise lgtm.

nabobalis:
Moved.

sunpy/util/exceptions.py
```diff
@@ -29,6 +29,14 @@ class SunpyUserWarning(UserWarning, SunpyWarning):
     """
 
 
+class SunpyMetadataWarning(UserWarning):
+    """
+    Warning class for cases metadata is missing.
```
Cadair, line 34:
```suggestion
    Warning class for cases metadata is missing.
    
```

nabobalis: Fixed.

PR #4375, opened 2020-07-16
Don't assume earth observer when calculating rsun_obs

dstansby:
Currently `sunpy` ignores any observer coordinate information when choosing a default value of `rsun_obs`, and assumes the observer is Earth. This PR changes that to calculate `rsun_obs` from `map.dsun`. `map.dsun` in turn is drawn from `self.observer_coordinate.radius.to('m')`.

This keeps the previous behaviour when `observer_coordinate` is not defined, because that defaults to Earth anyway.

This is pretty important I think because `draw_limb` relies on `map.rsun_obs`, so Solar Orbiter images would currently not have the limb drawn properly.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -748,7 +748,8 @@ class GenericMap(NDData):
         Notes
         -----
         This value is taken the ``'rsun_obs'``, ``'solar_r'``, or ``radius``
-        FITS keywords.
+        FITS keywords. If none of these keys are present the photospheric limb
+        as seen from the observer coordinate is returned.
         """
         rsun_arcseconds = self.meta.get('rsun_obs',
                                         self.meta.get('solar_r',
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -756,11 +756,14 @@ class GenericMap(NDData):
                                                                     None)))
 
         if rsun_arcseconds is None:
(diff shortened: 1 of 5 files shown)


sunpy/map/sources/tests/test_euvi_source.py
```diff
@@ -51,8 +52,8 @@ def test_rsun_missing():
     """Tests output if 'rsun' is missing"""
     euvi_no_rsun = Map(fitspath)
     euvi_no_rsun.meta['rsun'] = None
-    with pytest.warns(SunpyUserWarning, match='Missing metadata for solar radius'):
-        assert euvi_no_rsun.rsun_obs.value == sun.angular_radius(euvi.date).to('arcsec').value
+    r = euvi_no_rsun.observer_coordinate.radius
+    assert euvi_no_rsun.rsun_obs == sun._angular_radius(constants.radius, r)
```
dstansby, line 56: Turns out that EUVI was at 0.95 AU, so this value was always a bit wrong.

sunpy/map/mapbase.py
```diff
@@ -742,18 +742,28 @@ def rsun_meters(self):
...
     @property
     def rsun_obs(self):
-        """Radius of the Sun."""
+        """
+        Angular radius of the Sun.
+
+        Notes
+        -----
+        This value is taken the ``'rsun_obs'``, ``'solar_r'``, or ``radius``
+        FITS keywords.
```
ayshih, line 751: Might as well expand this description to also note the fallback of using the observer coordinate.

PR #4369, opened 2020-07-16
Pin tox-conda to use OpenBLAS below 0.3.10 due to apparent bugs on Linux

ayshih:
The conda-forge Linux builds of OpenBLAS 0.3.10 appear to result in catastrophic errors, from massively inaccurate calculations to segmentation faults.  These errors appear to be associated with OpenBLAS multithreading, because there are no errors if the environmental variable `OMP_NUM_THREADS=1` is set to force OpenBLAS to use only one thread.  (Due to a quirk in the code, this environmental variable works regardless of whether OpenBLAS is using OpenMP threads or pthreads threads.  On the other hand, the similarly named `OPENBLAS_NUM_THREADS`, which you'd think would be more general, does not work for both types of threading.)

For the time being, we'll pin our tox-conda dependencies to below 0.3.10 so that our CI will function.

Fixes #4356

PR #4384, opened 2020-07-17
Test for bugs on Linux with OpenBLAS 0.3.10

ayshih:
We apparently have massive issues with OpenBLAS 0.3.10 on Linux.  This PR undoes the pinning in #4369 for investigation into the bugs (see also #4368).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -75,7 +75,14 @@ jobs:
       - linux: py38-hypothesis
 
       - linux: py38-conda
-        name: py38_conda
+        name: py38_conda_linux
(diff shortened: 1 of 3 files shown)


ayshih:
It looks like the underlying issue was fixed in xianyi/OpenBLAS#2729 (staged for version 0.3.11) and backported to conda-forge's 0.3.10 builds in conda-forge/openblas-feedstock#103.

Re-running our tests

ayshih:
The test run `py38_conda_linux` now passes, so I've created a PR (#4425) to remove the version pin.

Issue #4324, opened 2020-06-29
Fix CRDER nan values in HMI maps


dstansby:
Loading a HMI map (e.g. https://docs.sunpy.org/en/stable/generated/gallery/acquiring_data/downloading_hmi.html#sphx-glr-generated-gallery-acquiring-data-downloading-hmi-py) results in a bunch of these warnings:
```
WARNING: FITSFixedWarning: CRDER1 = 'nan ' 
a floating-point value was expected. [astropy.wcs.wcs]
WARNING: FITSFixedWarning: CRDER2 = 'nan ' 
a floating-point value was expected. [astropy.wcs.wcs]
```

It would be good to fix the header in the `HMI` map source to prevent these warnings.

ayshih:
Related discussions: astropy/astropy#10153 and astropy/astropy#10165

In [this comment](https://github.com/astropy/astropy/pull/10165#issuecomment-616011425) from astropy/astropy#10165, @ndl303 said:
> On another note: I was speaking to the SDO HMI people and they have already started the process of making their FITS files compliant, which is probably a good thing.

Given that, we may not want to take any steps towards squelching these warnings.

Cadair:
I think that in general exposing warnings to the user about invalid FITS is a good thing, if we always hide these people will never learn :rofl:

dstansby:
How about warning once when loading the map and then fixing it? Currently the warning is thrown a lot, as it's every time `.wcs` is accessed on the map (I think), which can lead to a lot of warnings cluttering a terminal/notebook.

dstansby:
More generally, should we be warning every time we change some metadata in the map source `__init__`s? Should probably have a consistent policy about how to handle stuff there 😆

Cadair:
> Currently the warning is thrown a lot, as it's every time .wcs is accessed on the map (I think), which can lead to a lot of warnings cluttering a terminal/notebook.

Ah this would be an unintended side effect of the PSP changes to the generation of the WCS object. This is probably worth considering. (tbh I would expect these to be raised the first time you load the header out of the FITS file).

> More generally, should we be warning every time we change some metadata in the map source __init__s? Should probably have a consistent policy about how to handle stuff there

I think we did discuss this or something close to it reasonably recently. I can't remember in what context tho... :thinking:

nabobalis:
So what do we want to fix here now?

dstansby:
[...]

PR #4391, opened 2020-07-27
Warn whenever metadata assumptions are made

dstansby:
In the map sources, there are quite a few times where sunpy assumes default values. This PR makes sure that a warning is emitted every time a value is assumed with no reference to other information in the header to validate the assumption.

### DIFF
--- a/changelog/4391.feature.rst
+++ b/changelog/4391.feature.rst
@@ -1,0 +1,7 @@
+sunpy currently makes a number of assumptions about missing or incorrect values
+stored in files from various sources. Previously it was not always clear what
+or when these assumptions were being made. Now any assumptions raise a warning
+stating which key is being modified, and the new value it is being set to.
+
+In order to prevent these warnings, missing metadata must be explicitly set
+before being loaded by `sunpy.map.Map`.
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -421,6 +421,29 @@ class GenericMap(NDData):
         r = frame.represent_as(UnitSphericalRepresentation)
         return r.lon.to(self.spatial_units[0]), r.lat.to(self.spatial_units[1])
 
+    @staticmethod
+    def _fix_and_warn_header(header, key, value, replace_old=False):
+        """
+        Update *key* in *header* with a new *value*, and raise a warning. Any present values
+        are only overritten if *replace_old* is True.
+
+        This modifies *header* in place.
+        """
+        if key not in header:
+            warnings.warn(f'Did not find {key} in header, setting it to "{value}".',
+                          SunpyUserWarning)
+        elif replace_old:
+            warnings.warn(f'Fixing "{key}" header entry by replacing the old value '
+                          f'"{header[key]}" with "{value}".',
+                          SunpyUserWarning)
+        else:
+            return
+
+        header[key] = value
+
+    def _fix_and_warn(self, key, value, replace_old=False):
+        self._fix_and_warn_header(self.meta, key, value, replace_old)
+
     @property
     def _meta_hash(self):
         return self.meta.item_hash()
--- a/sunpy/map/sources/mlso.py
+++ b/sunpy/map/sources/mlso.py
@@ -35,10 +35,9 @@ class KCorMap(GenericMap):
         # Fill in some missing info
         self.meta['observatory'] = 'MLSO'
         self.meta['detector'] = 'KCor'
-        self.meta['waveunit'] = 'nanometer'
-        # Since KCor is on Earth, no need to raise the warning in mapbase
-        self.meta['dsun_obs'] = (sun.earth_distance(self.date)).to(u.m).value
-        self.meta['hgln_obs'] = 0.0
+        self._fix_and_warn('waveunit', 'nanometer')
+        self._fix_and_warn('dsun_obs', (sun.earth_distance(self.date)).to(u.m).value)
+        self._fix_and_warn('hgln_obs', 0)
         self._nickname = self.detector
 
         self.plot_settings['cmap'] = self._get_cmap_name()
--- a/sunpy/map/sources/rhessi.py
+++ b/sunpy/map/sources/rhessi.py
@@ -34,24 +34,19 @@ class RHESSIMap(GenericMap):
     """
 
     def __init__(self, data, header, **kwargs):
-        # Assume pixel units are arcesc if not given
-        header['cunit1'] = header.get('cunit1', 'arcsec')
-        header['cunit2'] = header.get('cunit2', 'arcsec')
-
+        # Fix some broken/misapplied keywords
+        if header['ctype1'] == 'arcsec':
+            header['cunit1'] = 'arcsec'
+            self._fix_and_warn_header(header, 'ctype1', 'HPLN-TAN', replace_old=True)
+        if header['ctype2'] == 'arcsec':
+            header['cunit2'] = 'arcsec'
+            self._fix_and_warn_header(header, 'ctype2', 'HPLT-TAN', replace_old=True)
         super().__init__(data, header, **kwargs)
 
         self._nickname = self.detector
         # TODO Currently (8/29/2011), cannot read fits files containing more
         # than one image (schriste)
-        # Fix some broken/misapplied keywords
-        if self.meta['ctype1'] == 'arcsec':
-            self.meta['cunit1'] = 'arcsec'
-            self.meta['ctype1'] = 'HPLN-TAN'
-        if self.meta['ctype2'] == 'arcsec':
-            self.meta['cunit2'] = 'arcsec'
-            self.meta['ctype2'] = 'HPLT-TAN'
-
-        self.meta['waveunit'] = 'keV'
+        self._fix_and_warn('waveunit', 'keV')
         self.meta['wavelnth'] = [self.meta['energy_l'], self.meta['energy_h']]
         self.plot_settings['cmap'] = 'rhessi'
 
--- a/docs/code_ref/map.rst
+++ b/docs/code_ref/map.rst
@@ -111,6 +111,11 @@ demonstrated by the following example.
             super(FutureMap, self).__init__(data, header, **kwargs)
 
             # Any Future Instrument specific keyword manipulation
+            #
+            # If any assumptions are made about metadata that is not present
+            # in the header, a warning MUST be raised to inform the user that
+            # sunpy is modifying the metadata. This can be done using the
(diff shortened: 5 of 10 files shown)


sunpy/map/sources/rhessi.py
```diff
@@ -46,12 +44,12 @@ def __init__(self, data, header, **kwargs):
         # Fix some broken/misapplied keywords
         if self.meta['ctype1'] == 'arcsec':
             self.meta['cunit1'] = 'arcsec'
-            self.meta['ctype1'] = 'HPLN-TAN'
+            self._fix_and_warn('ctype1', 'HPLN-TAN')
```
Cadair, line 47: There is a difference between your error message "we assumed a missing value for you" and "we are fixing a header to make it standard compliant" which is what this is. I am not sure what the best path forward for that is.

In this case, we know what the header means and we are fixing it so that it's WCS spec compliant. I am not convinced this warrants a warning.

sunpy/map/sources/yohkoh.py
```diff
@@ -56,15 +56,12 @@ def __init__(self, data, header, **kwargs):
         # use simple trigonometry to calculate the distance of the center of
         # the Sun from the spacecraft.  Note that the small angle approximation
         # is used, and the solar radius stored in SXT FITS files is in arcseconds.
-        self.meta['dsun_apparent'] = constants.au
         if 'solar_r' in self.meta:
             self.meta['dsun_apparent'] = constants.radius/(np.deg2rad(self.meta['solar_r']/3600.0))
+        else:
+            self._fix_and_warn('dsun_apparent', constants.au)
```
Cadair, line 62: I don't think we need this to throw either, this looks like we are merrily injecting our own header value.

Cadair:
While in general I support this idea, I think we should be warning when we inject missing values, I am very concerned about warning fatigue. Hence my comments about fixing vs injecting.

sunpy/map/sources/yohkoh.py
ayshih: Ugh, I don't like this Yohkoh handling.  We shouldn't add a new item to the dictionary for this.

This comment appears to suggest that the Yohkoh header has the wrong information in the FITS keyword `DSUN_OBS`, which is why a custom header keyword is created:
https://github.com/sunpy/sunpy/blob/db314114f0148d16610dda10f18ea0876942cd61/sunpy/map/sources/yohkoh.py#L52-L54

But, the `dsun` property is written as if `DSUN_OBS` is accurate if present (i.e, `meta['dsun_apparent']` may not actually be used):
https://github.com/sunpy/sunpy/blob/db314114f0148d16610dda10f18ea0876942cd61/sunpy/map/sources/yohkoh.py#L67

So, is `DSUN_OBS` inaccurate or is it not defined?

It seems to me that:

- If `DSUN_OBS` is wrong, it should be fixed and the warning raised upon fixing.
- If `DSUN_OBS` is just sometimes missing, the warning should be raised on access in the `dsun` property.
- Either way, there's no reason to add a new item to the dictionary.
- The fallback should never have a path to a fixed `constants.au`, but rather we should be falling back on an Earth observer.

sunpy/map/sources/rhessi.py
dstansby: I sort of threw this together after https://github.com/sunpy/sunpy/issues/4324#issuecomment-651158078 where you said that "in general exposing warnings to the user about invalid FITS is a good thing, if we always hide these people will never learn".

dstansby:
I've tried to limit this PR to cases where we are blindly injecting without validating against any other metadata - if there are some specific cases where that's not the case I can revert those specific changes.

sunpy/map/sources/rhessi.py
dstansby: Ah, I see what you mean. I don't think in this case we know that ctype1 has to be be `HPLN-TAN`, so I still think a warning should be raised, but I see that the current warning is not the right warning.

sunpy/map/sources/yohkoh.py
dstansby: Might be best to open this in a new issue? I'll revert the changes here.

changelog/4391.feature.rst
```diff
@@ -0,0 +1,7 @@
+sunpy currently makes a number of assumptions about missing or incorrect values
+stored in files from various sources. Previously it was not always clear what
+or when these assumptions were being made. Now any assumptions raise a warning
+stating which key is being modified, and the new value it is being set to.
+
+In order to prevent these warnings, missing metadata must be explicitly set
+before being loaded by `sunpy.map.Map`.
```
ayshih, line 7:
```suggestion
`sunpy.map.Map` classes now explicitly raise warnings when assumptions are made to fix missing or known incorrect values in file headers from various sources.
In order to prevent such a warning, the problematic metadata identified by the warning must be fixed before being loaded by `~sunpy.map.Map`.
```
I think this reads better.

sunpy/map/sources/yohkoh.py
ayshih: Created: #4395

changelog/4391.feature.rst
dstansby: Agreed, thanks a lot!

sunpy/map/mapbase.py
```diff
@@ -416,6 +416,38 @@ def _get_lon_lat(self, frame):
...
 
+    def _fix_and_warn(self, key, value):
+        """
+        Replace ``self.meta[key]`` with *value* and warn about the change.
+        """
+        warnings.warn(f'Fixing the header entry {key} by replacing the old value '
+                      f'({self.meta[key]}) with "{value}".',
+                      SunpyUserWarning)
+        header[key] = value
+
```
Cadair, line 427: This method is redefined below?
```suggestion
```

sunpy/map/sources/rhessi.py
```diff
@@ -46,12 +44,13 @@ def __init__(self, data, header, **kwargs):
         # Fix some broken/misapplied keywords
         if self.meta['ctype1'] == 'arcsec':
             self.meta['cunit1'] = 'arcsec'
-            self.meta['ctype1'] = 'HPLN-TAN'
+            self._fix_and_warn('cunit1', 'arcsec', replace_old=True)
+            self._fix_and_warn('ctype1', 'HPLN-TAN', replace_old=True)
         if self.meta['ctype2'] == 'arcsec':
-            self.meta['cunit2'] = 'arcsec'
-            self.meta['ctype2'] = 'HPLT-TAN'
+            self._fix_and_warn('cunit2', 'arcsec', replace_old=True)
```
Cadair, line 50: haven't we already injected this above?

sunpy/map/sources/rhessi.py
```diff
@@ -46,12 +44,13 @@ def __init__(self, data, header, **kwargs):
         # Fix some broken/misapplied keywords
         if self.meta['ctype1'] == 'arcsec':
             self.meta['cunit1'] = 'arcsec'
-            self.meta['ctype1'] = 'HPLN-TAN'
+            self._fix_and_warn('cunit1', 'arcsec', replace_old=True)
```
Cadair, line 47: same here

sunpy/map/mapbase.py
dstansby: 🤦

sunpy/map/sources/rhessi.py
dstansby: I've moved this editing to before `super().__init__` is called, and only warned for replacing `ctype` and not `cunit`.

wtbarnes:
Could the HMI header fixes in #4324 also be implemented here using `_fix_and_warn`? That way, the nan-related warnings are only emitted once.

dstansby:
Good idea, I'll put those fixes in this PR.

dstansby:
Actually we don't want to re-write the metadata for the HMI issue, we just want to get rid of it altogether, so I don't think `_fix_and_warn` is suited for that and I'll leave it for another PR.

sunpy/map/sources/mlso.py
```diff
@@ -35,10 +35,9 @@ def __init__(self, data, header, **kwargs):
         # Fill in some missing info
         self.meta['observatory'] = 'MLSO'
         self.meta['detector'] = 'KCor'
-        self.meta['waveunit'] = 'nanometer'
-        # Since KCor is on Earth, no need to raise the warning in mapbase
-        self.meta['dsun_obs'] = (sun.earth_distance(self.date)).to(u.m).value
-        self.meta['hgln_obs'] = 0.0
+        self._fix_and_warn('waveunit', 'nanometer')
+        self._fix_and_warn('dsun_obs', (sun.earth_distance(self.date)).to(u.m).value)
+        self._fix_and_warn('hgln_obs', 0)
```
Cadair, line 40: this makes me wonder, do we need to warn here? We know that this telescope is on Earth, and we know that the headers are missing this information (there is no standard enforcing they need to be there it's just part of sunpy's datamodel). Why are we going to throw warnings for this everytime?

dstansby:
Thanks for the comments! I definitely agree that we should have a good think about this, come and come up with a well thought out and documented policy for what to do when sunpy is given a .fits file that is not standards compliant. I also emphasize with the warning soup that sunpy sometimes emits...

I've opened https://github.com/sunpy/sunpy/issues/4410 for discussion - eventually we should probably SEP a policy on this?

dstansby:
Going to close this since it needs a SEP. Can always re-open if this is the desired solution.

nabobalis:
So is the plan to create a SEP or just update the behaviour first?

dstansby:
I am firmly of the camp that we shouldn't be blindly making assumptions about metadata without at least informing the user, and would like to push strongly for this to go into 2.1. I understand @Cadair's reservations in https://github.com/sunpy/sunpy/pull/4391#pullrequestreview-462407347, but I think it is much better to err on the side of caution to avoid situations where people are at best confused, and at worst completely misinterpreting their data because of an assumption sunpy is silently making.

dstansby:
Actually, lets leave this for 3.0.

Cadair:
I agree that silently assuming things about the metadata is a bad idea.

I am still a hard :-1: on raising warnings for things we aren't asking the users to change (we probably do this already but I would argue we shouldn't). If we spew warnings for stuff when we aren't asking for user action, people will start ignoring all the warnings as a matter of course and then they will ignore our important deprecation warnings and then get angry when "this version of sunpy broke my code and you didn't tell me" because they missed the deprecation warning.

So I agree with your premise, I just disagree with the implementation.

PR #4392, opened 2020-07-27
Update dev figure test hashes

dstansby:
Should fix dev figure tests. These changes are due to pandas fixes in matplotlib converters that make our images look good again!


dstansby:
💥 everything is broken! But at least this fixes the figure tests...

PR #4393, opened 2020-07-27
Fix differential rotation test

dstansby:
Looks like this was missing a line


sunpy/physics/tests/test_differential_rotation.py
```diff
@@ -264,6 +264,7 @@ def test_differential_rotate_time_straddles_limb(straddles_limb_map):
 def test_differential_rotate_time_off_disk(all_off_disk_map):
     # Test a map that is entirely off the disk of the Sun
     # Should report an error
+    new_time = all_off_disk_map.date + 48*u.hr
```
nabobalis, line 267: Woops.

dstansby:
😍 so many green ticks!

PR #4394, opened 2020-07-27
Functionality to client to download new GOES 16/17 data and reprocessed 13/14/15

### LINKED ISSUE #4173: New GOES XRS (16- 17) data will be in netcdf format
### Description
GOES-15 X-ray sensor is currently no longer operational (i.e no more GOES data from GOES-15). 
GOES-16 and 17 are now taking data are are being used by SWPC for flare classification. GOES-16 has actually been observing since 2017 but the XRS data has not been publicity released - but hurrah as far as I'm aware NOAA are planning to release this data next week (both the new data and the data which overlaps GOES-15) 🎉 

This data will be hosted here https://www.ngdc.noaa.gov/stp/satellite/goes-r.html and will be in netcdf format.

Similarly now NOAA has just released re-processed science quality data for GOES13-15, also in netcdf. 

There were rumblings of a netcdf reader in sunpy but is there one/a plan for one? If not we're going to need one 🌞

### LINKED ISSUE #4376: GOES 13-15 XRS data out of date
### Description
NOAA recalibrated the historical XRS data for GOES13-15 in May 2020, removing the scaling factors so that the data is now reported in true irradiance values (W/m^2).  SunPy still returns the older values with the scaling factors included.  

### Additional context
This is related to https://github.com/sunpy/sunpy/issues/4173 -- the newly calibrated data was only made available in netcdf format [here for example](https://www.ngdc.noaa.gov/stp/satellite/goes-r.html).

hayesla:
### Description
This is adding the functionality to the XRSClient to get access to the GOES XRS 16 and 17 data. 

It also now updates the client so that the default GOES13/14/15 data queried will return the new reprocessed data from the NOAA site rather than the old data. The old data can still be accessed through a new attrs `a.goes.VersionData`. I'm not too sure if this is how we want to go about this - but I think we should still provide an option to be able to download the old data (i.e. the data that this client used to return). 

This is for the high time resolution data (2s for 13/14/15 and 1s for 16/17)

Fixes #4173 
Fixes #4376 

These new data files are in netCDF format so at the moment do not work with `sunpy.timeseries` (which will be in another PR)


sunpy/net/dataretriever/attrs/goes.py
```diff
@@ -1,9 +1,18 @@
...
+__all__ = ["SatelliteNumber", "VersionData"]
 
 
 class SatelliteNumber(SimpleAttr):
     """
     The GOES Satellite Number
     """
+
+
+class VersionData(SimpleAttr):
```
abhijeetmanhas, line 12: Can you please add it to the class method `register_values`? So that it is known what can be the possible values for `a.VersionData`.

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -91,19 +91,19 @@ def _get_goes_sat_num(self, date):
...
-            uripath = urlsplit(uri).path
-
-            # Extract the yymmdd or yyyymmdd timestamp
-            datestamp = os.path.splitext(os.path.split(uripath)[1])[0][4:]
-
-            # 1999-01-15 as an integer.
-            if int(datestamp) <= 990115:
-                start = Time.strptime(datestamp, "%y%m%d")
+            if uri.endswith('fits'):
+                datestamp = os.path.basename(uri)[4:12]
```
abhijeetmanhas, line 95: If it ends with `fits`, then seems like it doesn't check whether the pattern is `%Y%m%d` or `%y%m%d`, i.e., that check around the date `1999/1/15`.

hayesla: good catch - thanks

sunpy/net/dataretriever/attrs/goes.py
hayesla: I've just added this to the register_values - can you see if what I've written is ok?

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -199,7 +213,9 @@ def register_values(cls):
         adict = {attrs.Instrument: [
             ("GOES", "The Geostationary Operational Environmental Satellite Program."),
             ("XRS", "GOES X-ray Flux")],
-            attrs.goes.SatelliteNumber: [(str(x), f"GOES Satellite Number {x}") for x in goes_number]}
+            attrs.goes.SatelliteNumber: [(str(x), f"GOES Satellite Number {x}") for x in goes_number], 
+            attrs.goes.VersionData: [('VersionData', ("If `old` then will use the 13, 14, and 15 " 
+                                     "old calibrated data before NOAA re-processed it"))]}
```
abhijeetmanhas, line 218:
```python
attrs.goes.VersionData: [('old', "old calibrated XRS data before NOAA re-processed it")]}
```
It should be like this, the adict contains list of tuples of possible values that an attr can take; and its description.
So `old` is one possible value, since we can do `a.goes.VersionData.old` and the second element will explain what old means here.

sunpy/net/dataretriever/attrs/goes.py
abhijeetmanhas: I have suggested some changes to it.

sunpy/database/tests/test_tables.py
```diff
@@ -56,7 +56,7 @@ def fido_search_result():
     return Fido.search(
         net_attrs.Time("2012/1/1", "2012/1/2"),
         net_attrs.Instrument('lyra') | net_attrs.Instrument('eve') |
-        net_attrs.Instrument('XRS') | net_attrs.Instrument('noaa-indices') |
+        net_attrs.Instrument('XRS')  | net_attrs.Instrument('noaa-indices') |
```
abhijeetmanhas, line 59: I guess this extra space should be removed.

sunpy/net/dataretriever/attrs/goes.py
abhijeetmanhas: https://github.com/sunpy/sunpy/pull/4394#discussion_r461384748

sunpy/net/dataretriever/sources/goes.py
hayesla: thanks!

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -159,7 +173,7 @@ def _makeimap(self):
         """
         Helper function used to hold information about source.
         """
-        self.map_["source"] = "nasa"
+        self.map_["source"] = "noaa"
```
abhijeetmanhas, line 176: Should we keep `NOAA/NASA` or just `NOAA`?

hayesla: good question - GOES XRS is a NOAA mission - however for data before GOES 13 the data we pull is from a NASA server. I think it's probably ok to say just NOAA as its not actually NASA data

sunpy/database/tests/test_tables.py
```diff
@@ -144,8 +144,9 @@ def test_entries_from_fido_search_result(fido_search_result):
         wavemin=0.1, wavemax=30.4)
     # 2 entries from goes
     assert entries[56] == DatabaseEntry(
-        source='nasa', provider='sdac', physobs='irradiance',
-        fileid='https://umbra.nascom.nasa.gov/goes/fits/2012/go1520120101.fits',
+        source='noaa', provider='sdac', physobs='irradiance',
```
wtbarnes, line 147: The SDAC is no longer the provider if we are now pulling from NOAA

wtbarnes:
This is great! 🎉

sunpy/database/tests/test_tables.py
hayesla: good point!

hayesla: actually following up on this - I guess how can this be handled in the case where data before GOES 13 is pulled from a NASA server, but data of > GOES 13 is now from the NOAA site?

wtbarnes: Oh right. Would it be possible to add the GOES < 13 (and the old calibration for 13-15) as a separate client? And then toggle between the old and new client based on the version number and satellite number attrs? I'm not sure if what I'm asking really makes any sense

hayesla: yeah that could make sense - however with this it would mean that the user would have to put in the satellite number for looking for the data which they may not know. At the moment if a user does 
` res = Fido.search(a.Time(tstart, tend), a.Instrument('XRS')`
and no satellite number is given it will find the available satellite that was taking data over that time and return that. 

Actually now that I think about it - how could you have two clients that both have a.Instrument('XRS') - would that work?

hayesla: oh actually I wonder if we could just have the [`_makeimap`](https://github.com/sunpy/sunpy/blob/master/sunpy/net/dataretriever/sources/goes.py#L158) call the [_get_url_for_timerange](https://github.com/sunpy/sunpy/blob/master/sunpy/net/dataretriever/sources/goes.py#L110) and check to see the url and then populate the server/source accordingly 

that will probably make more sense to do...

dstansby (dismissed):
Marking as needs changes as per https://github.com/sunpy/sunpy/pull/4394#pullrequestreview-457163618

Cadair:
@hayesla #4321 has been merged so this needs a re-work :smile:

hayesla:
> @hayesla #4321 has been merged so this needs a re-work 😄

on it 🌞

hayesla:
This is still a work in progress - the main issue being how we deal with the old GOES 13, 14 and 15 data that's on the NASA server. Really people should be using the new re-processed science quality data from the NOAA site (in netcdf), but maybe we should still allow access to the old fits files? The only difference is the scaling factor which is removed in the new netcdf data (which will change the peak flux value/GOES class). 

At the moment I've added a `a.goes.VersionData` attrs - but I'm not married to that idea. Actually its pretty messy. 

What are peoples thoughts on this? @wtbarnes @Cadair

wtbarnes:
Seems reasonable to me. Would this be something that the user always has to specify? Or does it take on a default value to always return the latest (netCDF) data?

hayesla:
> Seems reasonable to me. Would this be something that the user always has to specify? Or does it take on a default value to always return the latest (netCDF) data?

It defaults to the netCDF data, and the user would have to specify VersionData if they actually wanted the old data - it's probably the best way given that most people should be using the new data and are probably unaware that things have changed 🤦‍♀️

hayesla:
Final to do's

- [ ] Make it work for overlapping times before GOES 13 (i.e. searches that span before and after 2009-Sept)
- [ ] Write test for overlapping times
- [ ] Write proper documentation on the different data sources (probably in another PR)

hayesla:
Ok i think this should now work (in theory)! Its grown into something a little messy, but it gets the job done. 

I'll get onto writing some docs for the new GOES data etc.

hayesla:
unless I'm misreading the fails here, I don't think they are related to this PR?

dstansby:
A rebase might get rid of some of them if you haven't rebased recently?

hayesla:
failing doc build is due to https://github.com/sunpy/sunpy/pull/4592 for the example gallery (not in netcdf not fits)

I think everything else in this PR is working

dstansby:
Looks like the doc build is failing on some GOES examples.

changelog/4394.feature.rst
```diff
@@ -0,0 +1 @@
+Add support to the GOES XRS fido client to download new GOES 16 & 17 data reprocessed GOES 13, 14 and 15 data.
```
nabobalis, line 1:
```suggestion
Add support to the GOES XRS Fido client to download new GOES 16 & 17 data and reprocessed GOES 13, 14 and 15 data.
```

sunpy/database/tests/test_tables.py
```diff
@@ -151,9 +151,12 @@ def test_entries_from_fido_search_result(fido_search_result):
...
         wavemin=0.1, wavemax=30.4)
     # 2 entries from goes
+
     assert entries[58] == DatabaseEntry(
-        source='NASA', provider='SDAC', physobs='irradiance',
-        fileid='https://umbra.nascom.nasa.gov/goes/fits/2012/go1520120101.fits',
+        source='GOES', provider='NOAA', physobs='irradiance',
+        fileid='https://satdat.ngdc.noaa.gov/sem/goes/data/science/xrs/goes15/'
+               'gxrs-l2-irrad_science/2012/01/sci_gxrs-l2-irrad_g15_d20120101_v0-0-0.nc',
+
```
nabobalis, line 159:
```suggestion
```

sunpy/database/tests/test_tables.py
```diff
@@ -151,9 +151,12 @@ def test_entries_from_fido_search_result(fido_search_result):
         instrument='EVE',
         wavemin=0.1, wavemax=30.4)
     # 2 entries from goes
+
```
nabobalis, line 154:
```suggestion
```

nabobalis:
I merged in master and made some style changes.

kakirastern (approved):

hayesla:
there was actually a broken test there too. 

Yeah I don't know why those `HTTP Error 429` errors occur - must be from the NOAA website. But seems only issue when using scraper it seems 🤷‍♀️

nabobalis:
> Yeah I don't know why those `HTTP Error 429` errors occur - must be from the NOAA website. But seems only issue when using scraper it seems 🤷‍♀️

Yeah its failing sometimes on our CI. I am not sure how we want to handle this.

kakirastern:
The `build_docs` check failed as a result of 2 warnings treated as errors. I wonder if we can instead treat warnings as warnings and errors as errors.

kakirastern:
One salient feature I noticed is the `HTTPError` logs in the error output. Wondering if there is some quick fix by updating one's request credentials as discussed a while back, instead of recycling another individual's credentials for access?

kakirastern:
Regarding the `build_docs` check, we have the warnings:
```

    goes_ts = sunpy.timeseries.TimeSeries(goes_files, source='XRS', concatenate=True)
  File "/home/vsts/work/1/s/.tox/build_docs/lib/python3.8/site-packages/sunpy/timeseries/timeseries_factory.py", line 406, in __call__
    new_ts = self._check_registered_widgets(filepath=filepath, **kwargs)
  File "/home/vsts/work/1/s/.tox/build_docs/lib/python3.8/site-packages/sunpy/timeseries/timeseries_factory.py", line 529, in _check_registered_widgets
    data, meta, units = WidgetType._parse_file(filepath)
  File "/home/vsts/work/1/s/.tox/build_docs/lib/python3.8/site-packages/sunpy/timeseries/sources/goes.py", line 165, in _parse_file
    hdus = sunpy.io.read_file(filepath)
  File "/home/vsts/work/1/s/.tox/build_docs/lib/python3.8/site-packages/sunpy/io/file_tools.py", line 89, in read_file
    readername = _detect_filetype(filepath)
  File "/home/vsts/work/1/s/.tox/build_docs/lib/python3.8/site-packages/sunpy/io/file_tools.py", line 214, in _detect_filetype
    raise UnrecognizedFileTypeError("The requested filetype is not currently "
sunpy.io.file_tools.UnrecognizedFileTypeError: The requested filetype is not currently supported by SunPy.


/home/vsts/work/1/s/examples/time_series/goes_hek_m25.py failed leaving traceback:
  File "/home/vsts/work/1/s/examples/time_series/goes_hek_m25.py", line 24, in <module>
    goes = TimeSeries(files)
  File "/home/vsts/work/1/s/.tox/build_docs/lib/python3.8/site-packages/sunpy/timeseries/timeseries_factory.py", line 406, in __call__
    new_ts = self._check_registered_widgets(filepath=filepath, **kwargs)
  File "/home/vsts/work/1/s/.tox/build_docs/lib/python3.8/site-packages/sunpy/timeseries/timeseries_factory.py", line 529, in _check_registered_widgets
    data, meta, units = WidgetType._parse_file(filepath)
TypeError: cannot unpack non-iterable NotImplementedType object
```

hayesla:
The build_doc is failing as its waiting on #4592

kakirastern:
> The build_doc is failing as its waiting on #4592

I see. Thanks for the heads up!

Cadair:
Oh fun, the goes server is giving us `urllib.error.HTTPError: HTTP Error 429: Too Many Requests`

Cadair:
Also happened on the online tests, I wonder if this is the scraper making too many calls to the server or similar. The first thing we should check would be to make sure we aren't being inefficient with our use of the scraper or at any other point in the high level code. The next thing we could do is catch a 409 response and implement a delay of a few hundred ms and retry.

sunpy/extern/parse.py
```diff
@@ -471,7 +471,7 @@
 
 __all__ = 'parse search findall with_pattern'.split()
 
-log = logging.getLogger(__name__)
+log = logging.getLogger("parse")
```
Cadair, line 474: This is annoying, as we now have to remember to change this on every update.

Cadair: For context, not doing this was causing an error to be raised when the log level was set to debug. I assume because the name of this logger started with `sunpy` it was being sent through some of the astropy logging infrastructure in a way that it wasn't supposed to be and throwing a traceback.

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -22,7 +22,22 @@ class XRSClient(GenericClient):
     """
     Provides access to the GOES XRS fits files archive.
 
-    Searches data hosted by the `Solar Data Analysis Center <https://umbra.nascom.nasa.gov/goes/fits/>`__.
+    Searches data for the GOES XRS data both on NASA servers prior
```
Cadair, line 25: both? I thought we defaulted to just new servers?

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -22,7 +22,22 @@ class XRSClient(GenericClient):
...
+    to re-processed GOES 13, 14 and 15. For satellite numbers > 13
+    the XRSClient searches the NOAA archive, and returns the re-processed
+    science-quality data for GOES 13, 14 and 15, and also the new GOES-R series 16 and 17.
+
+    Note - the new science quality factors have scaling factors removed for 13, 14 and 15
```
Cadair, line 30:
```suggestion
    Note - the new science quality data have scaling factors removed for 13, 14 and 15
```
??

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -44,8 +59,105 @@ class XRSClient(GenericClient):
...
+                metalist.append(rowdict)
+
+        # if for some reason a user wants the old data.
+        elif (matchdict["Time"].end >= "2009-09-01" and matchdict["Provider"] == ["sdac"]):
+            scraper = Scraper(self.baseurl_old, regex=True)
+            filemeta = scraper._extract_files_meta(
+                matchdict['Time'], extractor=self.pattern_old, matcher=matchdict)
+            for i in filemeta:
+                rowdict = self.post_search_hook(i, matchdict)
+                metalist.append(rowdict)
```
Cadair, line 135: These two blocks seem identical. Can't they be combined into one?

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -44,8 +59,105 @@ class XRSClient(GenericClient):
...
+
+            for sat in [13, 14, 15]:
+                formdict = {"SatelliteNumber": sat}
+                urlpattern = self.baseurl_new.format(**formdict)
+                scraper = Scraper(urlpattern)
+                filemeta = scraper._extract_files_meta(
+                    matchdict["Time"], extractor=self.pattern_new, matcher=matchdict)
+                for i in filemeta:
+                    rowdict = self.post_search_hook(i, matchdict)
+                    metalist.append(rowdict)
```
Cadair, line 158: rather than repeating the whole block here, could you not set a variable like `new_pattern = self.pattern_r` and the same for base url in the if block and then only write the actual scraping code once using those variables?

Cadair (changes requested):
Looks good, I think there is some deduplication that can happen on the calls to scraper though.

sunpy/net/dataretriever/sources/goes.py
hayesla: yep - but for the data prior to GOES 13 the NASA server data is correct (NOAA haven't provided re-processed data)

sunpy/net/dataretriever/sources/goes.py
hayesla: hmm I'm not too sure if this would work due to the fact that there are days when they are available GOES 13, 14, 15, 16 and 17 data so needs to do this in an append method I think

Cadair: Ah yeah I think I see what you are saying. It still feels a little to repetitive here, I wonder if moving this pattern into a function which could be used in these 3 different conditions would be worth it?

hayesla: yeah thats probably the best idea, I'll do that

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -44,8 +60,91 @@ class XRSClient(GenericClient):
...
+        metalist = []
+        # the data before the re-processed GOES 13, 14, 15 data.
+        if (matchdict["Time"].end < "2009-09-01") or (matchdict["Time"].end >= "2009-09-01" and matchdict["Provider"] == ["sdac"]):
+            metalist += self._get_metalist_fn(matchdict, self.baseurl_old, self.pattern_old)
+
+        # new data from NOAA.
+        else:
+            if matchdict["Time"].end >= "2017-02-07":
+                for sat in [16, 17]:
+                    metalist += self._get_metalist_fn(matchdict, self.baseurl_r.format(**{"SatelliteNumber": sat}), self.pattern_r)
```
Cadair, line 141:
```suggestion
                    metalist += self._get_metalist_fn(matchdict, self.baseurl_r.format(SatelliteNumber=sat), self.pattern_r)
```

sunpy/net/dataretriever/sources/goes.py
```diff
@@ -44,8 +60,91 @@ class XRSClient(GenericClient):
...
+
+            if matchdict["Time"].end <="2020-03-04":
+                for sat in [13, 14, 15]:
+                    metalist += self._get_metalist_fn(matchdict, self.baseurl_new.format(**{"SatelliteNumber": sat}), self.pattern_new)
```
Cadair, line 145:
```suggestion
                    metalist += self._get_metalist_fn(matchdict, self.baseurl_new.format(SatelliteNumber=sat), self.pattern_new)
```

Cadair (approved):

Cadair:
Thanks @hayesla this and #4592 was an awesome effort :tada:

kakirastern:
🎉

PR #4397, opened 2020-07-29
Only put changed figures on the figure test report page

dstansby:
The figure test page is a bit large with all the figures, so just put the ones that changed. Perhaps there is a good reason to put them all on, that I can't think of at the moment?


dstansby:
Seems to work, none of the figures changed and there are no figures on the figure test page.

PR #4399, opened 2020-07-29
update RTD and fix docs

nabobalis:
Fix a deprecation warning in the docs alongside some minor RTD changes.


nabobalis:
The parallel write/read settings for RTD. They build with `-j auto`

PR #4400, opened 2020-07-30
Remove more warnings from tests

dstansby:
Some more stuff pulled out of #3648.

### DIFF
--- a/sunpy/map/tests/test_compositemap.py
+++ b/sunpy/map/tests/test_compositemap.py
@@ -7,7 +7,7 @@ import pytest
 
 import sunpy.data.test
 import sunpy.map
-from sunpy.tests.helpers import figure_test
+from sunpy.tests.helpers import figure_test, fix_map_wcs
 
 testpath = sunpy.data.test.rootdir
 
--- a/sunpy/map/tests/test_compositemap.py
+++ b/sunpy/map/tests/test_compositemap.py
@@ -11,6 +11,9 @@ from sunpy.tests.helpers import figure_test, fix_map_wcs
 
 testpath = sunpy.data.test.rootdir
 
+# Ignore missing metadata warnings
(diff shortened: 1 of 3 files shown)


sunpy/map/tests/test_compositemap.py
```diff
@@ -7,10 +7,13 @@
 
 import sunpy.data.test
 import sunpy.map
-from sunpy.tests.helpers import figure_test
+from sunpy.tests.helpers import figure_test, fix_map_wcs
```
nabobalis, line 10: This is unused?

sunpy/net/helio/tests/test_helio.py
```diff
@@ -12,6 +12,8 @@
     wsdl_retriever,
 )
 
+pytestmark = pytest.mark.filterwarnings('ignore:Unverified HTTPS request is being made')
```
Cadair, line 15: why are we making unverified https requests?!

dstansby: 🤷 is this something we should fix instead of ignoring?

Cadair: I think we should understand it at least. Could be cret issues with the helio servers I guess?

nabobalis: We had to disable the SSL check for helio.

dstansby: I'll open an issue with the traceback

Cadair: > We had to disable the SSL check for helio.

We did? I thought it didn't actually fix it?

nabobalis: That was for some instr data.

PR #4405, opened 2020-08-03
Allow jsocClient to download .tar archives from JSOC

### LINKED ISSUE #4291: Support url-tar method downloads in the JSOCClient
### Description

Currently we hard code downloading files from the JSOC using the `url` method. This in general is the nicest way for our users as they get the number of files they expect to get from the search results returned to them, and they can be opened with no further manual steps.

However, sometimes if you are requesting *a lot* of files, this isn't a good situation for the JSOC who have limited resources to service the download request. We should at the very least enable people to manually select url-tar when they need to. Also we should probably document doing that if you are planning on downloading >N files where N is large.

### Additional context

This request came from the JSOC team as they have had people using "sunpy" trying to download a sufficient number of files they have had to block their IPs.

lzivadinovic:
### Description

This PR allows JSOCClient to download all staged files as single .tar archive while still maintaining file list in base object.

Fixes #4291

### DIFF
--- a/changelog/4405.feature.rst
+++ b/changelog/4405.feature.rst
@@ -1,0 +1,1 @@
+`sunpy.net.jsoc.JSOCClient.request_data` now support additional parameter `download_tar` which allows user to download staged data as single .tar file.
--- a/changelog/4405.feature.rst
+++ b/changelog/4405.feature.rst
@@ -1,1 +1,1 @@
-`sunpy.net.jsoc.JSOCClient.request_data` now support additional parameter `method` which allows user to download staged data as single .tar file.
+`sunpy.net.jsoc.JSOCClient.request_data` now support additional parameter "method" which allows user to download staged data as single .tar file.
--- a/changelog/4405.feature.rst
+++ b/changelog/4405.feature.rst
@@ -1,1 +1,1 @@
-`sunpy.net.jsoc.JSOCClient.request_data` now support additional parameter `download_tar` which allows user to download staged data as single .tar file.
+`sunpy.net.jsoc.JSOCClient.request_data` now support additional parameter `method` which allows user to download staged data as single .tar file.
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -447,11 +447,11 @@ class JSOCClient(BaseClient):
             protocol = block.get('protocol', 'fits')
 
             if protocol not in supported_protocols:
-                error_message = "Protocols other than fits and as-is "\
+                error_message = f"Protocols other than {','.join(supported_protocols)} "\
                                 "are not supported."
                 raise TypeError(error_message)
             if method not in supported_methods:
-                error_message = "Methods other than url-tar, url and url-quick "\
+                error_message = f"Methods other than {','.join(supported_methods)} "\
                                 "are not supported."
                 raise TypeError(error_message)
 
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -410,7 +410,7 @@ class JSOCClient(BaseClient):
 
         return res
 
-    def request_data(self, jsoc_response, download_tar=False, **kwargs):
+    def request_data(self, jsoc_response, method='url', **kwargs):
         """
         Request that JSOC stages the data for download. This method will not
         wait for the request to be staged.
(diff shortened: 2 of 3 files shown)


sunpy/net/jsoc/jsoc.py
```diff
@@ -410,7 +410,7 @@ def search_metadata(self, *query, **kwargs):
 
         return res
 
-    def request_data(self, jsoc_response, **kwargs):
+    def request_data(self, jsoc_response, download_tar=False, **kwargs):
```
Cadair, line 413: Would it not be more flexible to allow the user to specify `method=` and `protocol=` explicitly here?

Cadair:
Thanks for opening this PR, sorry I was a bit slow getting to it. I have one major question, but otherwise it looks great.

sunpy/net/jsoc/jsoc.py
lzivadinovic: True, but do we want to support all methods/protocols combinations? If so, i can create changes for that. I guess, supporting ftp is ok, but not so sure regarding jpeg/mpeg/mp4 because of processing and additional options you have to select in JSOC export (colormap, scaling, img size).

Cadair: I think you are correct that we don't really want to support all the possible options, but I think building the API in a way which is extensible into the future so we can add them if we want is a better choice.

I would suggest you allow things which work with minimal modification and disallow everything else, and then there is simple (non-API changing) path forward for adding them in the future.

lzivadinovic: Ok, i agree. I will re-work it.

sunpy/net/jsoc/jsoc.py
```diff
@@ -432,18 +438,25 @@ def request_data(self, jsoc_response, **kwargs):
...
 
             ds = self._make_recordset(**block)
             cd = drms.Client(email=block.get('notify', ''))
             protocol = block.get('protocol', 'fits')
 
-            if protocol != 'fits' and protocol != 'as-is':
-                error_message = "Protocols other than fits and as-is are "\
+            if protocol not in supported_protocols:
+                error_message = "Protocols other than fits and as-is "\
+                                "are not supported."
```
Cadair, line 451:
```suggestion
                error_message = f"Protocols other than {','.join(supported_protocols)} "\
                                "are not supported."
```

changelog/4405.feature.rst
```diff
@@ -0,0 +1 @@
+`sunpy.net.jsoc.JSOCClient.request_data` now support additional parameter `download_tar` which allows user to download staged data as single .tar file.
```
Cadair, line 1: This needs updating based on the changes.

Cadair (commented):
Looks like it's basically there.

sunpy/net/jsoc/jsoc.py
```diff
@@ -432,18 +439,25 @@ def request_data(self, jsoc_response, **kwargs):
 
         requests = []
         self.query_args = jsoc_response.query_args
+        supported_protocols = {'fits', 'as-is'}
+        supported_methods = {'url-tar', 'url', 'url-quick'}
         for block in jsoc_response.query_args:
 
             ds = self._make_recordset(**block)
             cd = drms.Client(email=block.get('notify', ''))
             protocol = block.get('protocol', 'fits')
```
Cadair, line 448: We don't have to fix this here, but it feels weird that email, protocol and notify are not all the same?

Cadair (approved):

sunpy/net/jsoc/jsoc.py
```diff
@@ -420,6 +420,13 @@ def request_data(self, jsoc_response, **kwargs):
         jsoc_response : `~sunpy.net.jsoc.jsoc.JSOCResponse` object
             The results of a query
 
+        method : `str`
```
nabobalis, line 423:
```suggestion
        method : {'url-tar', 'url', 'url-quick'}
```

sunpy/net/jsoc/jsoc.py
```diff
@@ -420,6 +420,13 @@ def request_data(self, jsoc_response, **kwargs):
...
+            Method for requesting JSOC data, can be 'url-tar', 'url' and 'url-quick'
```
nabobalis, line 424:
```suggestion
            Method for requesting JSOC data, can be 'url-tar', 'url' (the default) and 'url-quick'
```

sunpy/net/jsoc/jsoc.py
lzivadinovic: I dont understand this change, i assumed that we want to say what data type given parameter is.

nabobalis: We try to follow the https://numpydoc.readthedocs.io/en/latest/format.html guide for our docs.
For parameters that have a fixed set of values, this is the format.

changelog/4405.feature.rst
```diff
@@ -0,0 +1 @@
+`sunpy.net.jsoc.JSOCClient.request_data` now support additional parameter `method` which allows user to download staged data as single .tar file.
```
nabobalis, line 1:
```suggestion
`sunpy.net.jsoc.JSOCClient.request_data` now support additional parameter "method" which allows user to download staged data as single .tar file.
```

nabobalis:
Stupid question: To use fido with this, you would pass in `method=url-tar` to fetch?

nabobalis (approved):

Cadair:
> Stupid question: To use fido with this, you would pass in `method=url-tar` to fetch?

This doesn't add Fido support at the moment

lzivadinovic:
Should i rebase this PR?

Issue #4406, opened 2020-08-04
32 bit tests failing


dstansby:
The 32 bit tests are currently failing, due to a failed attempt to install the `cryptography` library from source. See e.g. https://dev.azure.com/sunpy/sunpy/_build/results?buildId=9154&view=logs&jobId=a7b3aa55-7d57-562f-3433-7f6b2d4252da&j=2920fa54-56c6-5dfd-d3d8-7b1b694d90cb&t=bba4d609-c264-5f9c-e072-020a8a3b573f

nabobalis:
My interpretation of the issue is as follows:

"cryptography" is looking for `SSL_OP_NO_DTLSv1_2`, `SSL_OP_NO_DTLSv1`, `X509_CHECK_FLAG_ALWAYS_CHECK_SUBJECT` and `X509_CHECK_FLAG_MULTI_LABEL_WILDCARDS`.

The first two are not defined in openssl 1.0.1 (https://github.com/openssl/openssl/blob/OpenSSL_1_0_1e/ssl/ssl.h) but are defined in openssl 1.0.2 (https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/ssl/ssl.h).

The issue is that the linux image used for the 32bit image is manylinux 2010 which installs version "1.0.1e" of openssl. The current version of cryptography is only tested against ,OpenSSL 1.0.2-latest, OpenSSL 1.1.0-latest and OpenSSL 1.1.1-latest and will soon drop support for 1.0.2.

We either figure out what version of cryptography to pin for in the 32bit build so it works again or we update the 32bit image used.

nabobalis:
cryptography 2.9.2 being the version we pin to.

Cadair:
How about jumping back to `manylinux1` on the 32bit build?

nabobalis:
What version of openssl does it use?

nabobalis:
It seems they don't provide any 32bit wheels for linux anymore:
https://pypi.org/project/cryptography/#files

nabobalis:
Since we have to compile it from source, we should just see about pinning its version for the 32 bit build.

nabobalis:
I pinned it in tox here https://github.com/sunpy/sunpy/pull/4388

PR #4407, opened 2020-08-04
Install ssl dev headers on 32 bit tests

dstansby:
Might help with https://github.com/sunpy/sunpy/issues/4406

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -68,6 +68,10 @@ jobs:
 
       - linux32: py38-32bit
         name: py38_test_32bit
+        libraries:
+          yum:
+            - libopenjp2-7
+            - openssl-devel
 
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -68,6 +68,10 @@ jobs:
 
       - linux32: py38-32bit
         name: py38_test_32bit
+        libraries:
+          apt:
+            - libopenjp2-7
+            - openssl-devel
```
lzivadinovic, line 74: -devel are sufix for fedora like distros, on ubuntu its -dev

try `libssl-dev` instead and maybe `libopenjp2-7-dev` if we use lib for some compile 

-dev packages contains header files required for compilation, without dev, they are just compiled libraries.

PR #4409, opened 2020-08-05
Retain nanosecond precision in parse_time for pandas

dstansby:
Currently `parse_time` called with a `pandas.Timestamp` uses `time_string.to_pydatetime()` to convert the timestamp into something that astropy `Time` understands. This loses nanosecond precision however.

Instead, go via a numpy datetime64, which does not lose nanosecond precision.


dstansby:
🎉 these warnings have been plaguing me for years, feels good to finally fix them!

PR #4178, opened 2020-05-16
Enable sphinx nitpicky mode

dstansby:
This makes the doc build error if links are broken. 
Fixes #3539.

PR #4418, opened 2020-08-07
Add some missing bits to the API docs

dstansby:
Pulled out of https://github.com/sunpy/sunpy/pull/4178. Adds some previously missing API docs that are linked to in other bits of the documentation.

### DIFF
--- a/sunpy/sun/constants.py
+++ b/sunpy/sun/constants.py
@@ -30,12 +30,13 @@ def get(key):
 
     Returns
     -------
-    constant : `~astropy.units.Constant`
+    constant : `~astropy.constants.Constant`
 
     See Also
     --------
     `sunpy.sun.constants`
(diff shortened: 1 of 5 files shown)


sunpy/sun/constants.py
```diff
@@ -134,5 +139,8 @@ def _build_docstring():
 # Observable parameters
 average_angular_size = get('average angular size')
 sidereal_rotation_rate = get('sidereal rotation rate')
+
 first_carrington_rotation = Time(get('first Carrington rotation (JD TT)'), format='jd', scale='tt')
+first_carrington_rotation.__doc__ = 'Time of the start of the first Carrington rotation'
```
Cadair, line 144: Does not doing this work?
```suggestion
"""Time of the start of the first Carrington rotation"""
```

dstansby: I used the same syntax as above (`#:` above the instance).

nabobalis (approved):

PR #4420, opened 2020-08-18
Un-pin pytest

dstansby:
Remove pytest pinning since pytest-doctestplus should now be fixed, and pytest-xdist requires pytest>=6 which is causing CI builds to fail at the moment.


dstansby:
Going to self-merge to fix CI giving one approval

PR #4424, opened 2020-08-19
Remove all transformations from a frame instance to a frame class

ayshih:
This PR replaces all internal cases where a frame instance is transformed to a frame class with an explicitly instantiated destination frame.  All results are unchanged.  This change anticipates the pending astropy/astropy#10475, but does not need to wait for that PR to be merged.


Cadair (approved):
Nice, much clearer on attributes.

PR #4426, opened 2020-08-19
Raise an error if a file contains no HDUs with NAXIS > 1

### LINKED ISSUE #4413: Exporting SDO FITS Files without Image Data Arrays
### Description

I'd like to propose a change to the JSOC export system to help users handle image files that are requested but missing and cannot be downloaded. In addition to affecting the JSOC export web-page users, this change would affect users of software that uses the JSOC export-system CGI. This includes various existing scripts that use URL-opening libraries or linux commands to access these CGI scripts. As such, this will affect users of the drms/securedrms Python package and, consequently, SunPy users that download SDO data with fido or drms directly. 

The Problem
-----------
Internally, the JSOC stores image metadata (i.e., FITS keyword data) separately from the actual image data, and there are cases where metadata exist, like the time of an observation, but no image exists. An example is the JSOC data storage that results when an eclipse prevents SDO from capturing an image of the sun. The observation time and other metadata are stored at the JSOC, but no image exists so no image is stored. Storing metadata separate from image data can lead to an export-system user having access to the metadata, but not having access to an image. Software that queries the JSOC database to obtain metadata often assumes that because metadata exist, so do the corresponding images. The software does know ahead of time if images are missing, so this is a reasonable supposition. The software will then attempt to download the non-existent images and fail. Things get worse if the software has requested many images, but only one image is non-existent - the entire export request will fail.

The JSOC does have a not-well-known means for checking on the presence of an image file, but download-software generally has not been implemented to use it. And even if software were to use this method, it is a bit inefficient and cumbersome to have to make two queries every time you want to download a range of image data that is mostly present. Why not ask us for a range of image data and we give you what we have within that range, telling you when an expected image is missing?

Blah, Blah, Blah (skippable details)
------------------------------------
DRMS provides an INCONSISTENT means for notifying download-software of ONE cause of this problem. For certain series a metadata keyword named QUALITY exists. The value of this keyword is a 32-bit bitmap, with one bit indicating whether an image exists or not. I say QUALITY is inconsistent because some DRMS data series have the QUALITY keyword, and some do not. And the meaning of the various bits is series-specific. I will set up a different thread to deal with the inconsistency of QUALITY. It is an important problem that must also be dealt with. But for the present discussion, let's assume your software has attempted to download a 'missing' image.

There are other reasons why an image might be missing. The most likely one is an image file is on tape, but not on disk - offline data. The JSOC has a finite amount of disk space so not all data can be online at the same time. Exporting offline data is currently not supported due to the potential for the need for manual intervention to place data online. In addition, there are other less likely reasons to be in this situation: a disk crashes, a network switch dies, a file system goes belly up. The point is, no matter how careful one is, there is no way to prevent with 100% certainty an export user from finding themselves in a missing-image scenario.

Proposal
--------
Here is what I'd like to do: if an export user has access to image metadata, but not the actual image, I'd like to send the user a file that has the image metadata, but no image. In this way, as long as the normal HTTP mechanisms work, the requestor will not be in the situation where they attempted to download an image file, which failed, potentially causing the request for many images to fail also. At the very least, the user will end up with a file that contains the metadata. At the moment, we are focusing on data series with a FITS-file container, especially since this scenario, where metadata exist but no image exists, is going to almost always concern FITS files. So in the FITS case, the user will end up with a FITS file that has FITS keywords (but NAXIS will be 0 since no image exists), but no image array. However, the principle should apply to all types of data-series file containers. For a user of drms.py/securedrms.py, the client.export() call will ultimately result in the download of FITS keywords, just as the client.query() call does. 

If I were to make this change, would this help the SunPy community? What problems might it cause?

Cadair:
In #4413 it was proposed that in the future some of the files returned by the JSOC might have no array data in them, but these files would be potentially mixed in with a bunch of files which did in the results of a search. This lead me to wonder what would happen if you gave sunpy a list of FITS, where one has no array data in it, but all the others do, can you still load the ones that do, do you get errors, warnings or anything else?

So I created such a file with:
```python
fits.writeto(header=fits.Header(head), data=np.array(()), filename="test.fits", overwrite=True)
```

and then loaded it into Map, both on it's own and with other files. Map completely ignored it, no error (unless it was the only file, where a `NoMap` error was raised) no warning, error or anything.

On further investigation, it seems that we are dropping all HDUs with `NAXIS = 1` here:

https://github.com/sunpy/sunpy/blob/7a8e7f9a74b558908a58ef35891ec6c8ff18e2ab/sunpy/map/map_factory.py#L171-L174

This PR changes this behaviour to raise an error if *all* HDUs in the file have `NAXIS = 1` and are therefore dropped. This error is then later caught during the `__call__` method and converted to a warning if `silence_errors=True` in the call to `Map`.

Fixes #4413

### DIFF
--- a/sunpy/map/map_factory.py
+++ b/sunpy/map/map_factory.py
@@ -24,6 +24,7 @@ from sunpy.util.datatype_factory_base import (
     NoMatchError,
     ValidationFunctionError,
 )
+from sunpy.util.exceptions import NoMapsInFileError, SunpyUserWarning
 from sunpy.util.functools import seconddispatch
 from sunpy.util.metadata import MetaDict
 from sunpy.util.types import DatabaseEntryType
@@ -172,6 +173,10 @@ class MapFactory(BasicRegistrationFactory):
                 data = filedata
                 meta = MetaDict(filemeta)
                 new_pairs.append((data, meta))
+
+        if not new_pairs:
+            raise NoMapsInFileError(f"Found no HDUs with >= 2D data in '{fname}'.")
+
         return new_pairs
 
     def _validate_meta(self, meta):
@@ -185,7 +190,7 @@ class MapFactory(BasicRegistrationFactory):
         else:
             return False
 
-    def _parse_args(self, *args, **kwargs):
+    def _parse_args(self, *args, silence_errors=False, **kwargs):
         """
         Parses an args list into data-header pairs.
 
@@ -236,7 +241,12 @@ class MapFactory(BasicRegistrationFactory):
         # Note that this list can also contain GenericMaps if they are directly given to the factory
         data_header_pairs = []
         for arg in args:
-            data_header_pairs += self._parse_arg(arg, **kwargs)
+            try:
+                data_header_pairs += self._parse_arg(arg, **kwargs)
+            except NoMapsInFileError as e:
+                if not silence_errors:
+                    raise
+                warnings.warn(f"One of the arguments failed to parse with error: {e}", SunpyUserWarning)
 
         return data_header_pairs
 
@@ -320,7 +330,7 @@ class MapFactory(BasicRegistrationFactory):
         Extra keyword arguments are passed through to `sunpy.io.read_file` such
         as `memmap` for FITS files.
         """
-        data_header_pairs = self._parse_args(*args, **kwargs)
+        data_header_pairs = self._parse_args(*args, silence_errors=silence_errors, **kwargs)
         new_maps = list()
 
         # Loop over each registered type and check to see if WidgetType
@@ -339,7 +349,7 @@ class MapFactory(BasicRegistrationFactory):
                     ValidationFunctionError, MapMetaValidationError) as e:
                 if not silence_errors:
                     raise
-                warnings.warn(f"One of the data, header pairs failed to validate with: {e}")
+                warnings.warn(f"One of the data, header pairs failed to validate with: {e}", SunpyUserWarning)
 
         if not len(new_maps):
             raise RuntimeError('No maps loaded')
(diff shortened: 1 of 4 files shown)


sunpy/map/map_factory.py
```diff
@@ -172,6 +179,10 @@ def _read_file(self, fname, **kwargs):
                 data = filedata
                 meta = MetaDict(filemeta)
                 new_pairs.append((data, meta))
+
+        if not new_pairs:
+            raise NoMapsInFileError(f"Reading the file '{fname}' found no HDUs with NAXIS > 1, so no maps could be loaded.")
```
dstansby, line 184: This isn't actually a check on the `NAXIS` keyword, but on the shape of the data, so I think the check should be changed or the warning should be changed to reflect the check (that the data has >= two dimensions).

sunpy/map/map_factory.py
```diff
@@ -38,6 +39,12 @@
 __all__ = ['Map', 'MapFactory']
 
 
+class NoMapsInFileError(Exception):
```
dstansby, line 42: Huh, is there not a custom `SunpyException` to inherit from?

sunpy/map/map_factory.py
dstansby, line 184:
```suggestion
            raise NoMapsInFileError(f"Found no HDUs in '{fname}' with NAXIS > 1, so no maps could be loaded.")
```
Although the message might change, I think it can be made shorter.

dstansby (changes requested):

dstansby:
I can try and pick this up and finish it?

nabobalis (approved):

Cadair (commented):
Looks good :+1:

dstansby (approved):

PR #4427, opened 2020-08-19
Follow-up: Removed one more transformation from a frame instance to a frame class

ayshih:
Follow-up on #4424 

Not sure how I missed this one call.  Now fixed.


ayshih:
Failures are unrelated, someone approve =)

nabobalis (approved):

Issue #4390, opened 2020-07-24
Fido download of NOAA SRS broken for dates before 1996


samaloney:
### Description

Trying to download SRS data before 1996 results in an error.

### Expected behavior

SRS files downloaded

### Actual behavior

Download appears to work but inspecting the results reveals an error.

The format of the NOAA archive changes pre 1996 (ftp://ftp.swpc.noaa.gov/pub/warehouse) and the current code doesn't support the other format e.g. 

https://github.com/sunpy/sunpy/blob/49279e22213b8a50753ce90c3c0cf0130b0a02f7/sunpy/net/dataretriever/sources/noaa.py#L186-L201

To fix would need to add code generate correct urls for dates before 1996 and also add code to extract the SRS files from the different structured tar.gz (I think its just a tar.gz of all the contents for the year). 

### Steps to Reproduce

```python
from sunpy.net import Fido, attrs as a
query = Fido.search(a.Time('1990-05-01', '1990-05-02'), a.Instrument.srs_table)
#query = Fido.search(a.Time('1990-05-01', '1990-05-02'), a.Instrument('srs_table'))
files = Fido.fetch(query[0,0])
files 

<parfive.results.Results object at 0x113188090>
['/Users/shane/sunpy/data/19900501SRS.txt']
Errors:
(error(filepath_partial=<function Downloader.enqueue_file.<locals>.filepath at 0x104e07170>, url='ftp://ftp.swpc.noaa.gov/pub/warehouse/1990/1990_SRS.tar.gz', exception=StatusCodeError("Waiting for ('1xx',) but got 550 [' /pub/warehouse/1990/1990_SRS.tar.gz: No such file or directory']")))
```

### System Details

- SunPy Version: 2.0.1
- Astropy Version:
- Python Version:
- OS information:

samaloney:
Actually I think the data pre 1996 can't be parsed as it's free text so I think we just need to manual add a minimum date date of 1996/01/01

PR #4432, opened 2020-08-23
Ignore out-of-range years for an SRS query

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/4390.


Cadair:
Easiest way to check for no results is running the `SRSClient.search` method directly, that should return an object with `len() == 0`.

Cadair (changes requested):
This should probably wait for #4321

dstansby:
Just given this an update, and it's all required.

nabobalis (approved):

dstansby:
Maybe we want to backport this?

Issue #3885, opened 2020-03-11
Reformatting Colour Tables


nabobalis:
The code (https://github.com/sunpy/sunpy/tree/master/sunpy/visualization/colormaps) that deals with our color tables needs some reformatting to be cleaner.

Ideally we need to do:

* Change the color tables into a series of enums to allow someone to access the array's directly without the use of a function. 
* This means deprecation of most of the functions in those files.
* Move the arrays into their own file whether it be python or a plain text file.

abhijeetmanhas:
I want to give it a go. There are many arrays in that file. Should I keep all of them in `sunpy/data` or create a folder named `sunpy/visualization/colormaps/colortables` containing all the arrays stored as `json` or python numpy arrays in it?

nabobalis:
I would put them in the same folder as the code to start off. Moving then if we want can be done after.

Cadair:
I would think something like csv would be best, ascii and easy to read in a text editor.

ehsteve:
csv files also render nicely on GitHub which might be convenient.

jeffreypaul15:
Should the conversion of the tables into enums still be done? I can give it a go if required

nabobalis:
In theory I don't see why not.

kakirastern:
Well said

jeffreypaul15:
I'm a little confused as to what exactly has to be converted to enums here? Correct me if I'm wrong but the functions in color_tables.py have to be removed and replaced with them accessing the emuns which are the variables such as aia_wave_dict.

nabobalis:
A design/API was never finalised, I would suggest seeing what you think is easier/pretty and opening that PR so we can iterate on it.

jeffreypaul15:
```Python
The following constants in color_tables.py are the  same but part of 2 different functions, the constants c0, c1, c2 & c3 all follow the same calculations but differ in type, one is float and the other int. This might cause an issue during creation of enums, what should be done here?


c0 = np.arange(256, dtype='f')
c1 = (np.sqrt(c0) * np.sqrt(255.0)).astype('f')
c2 = (np.arange(256)**2 / 255.0).astype('f')
c3 = ((c1 + c2 / 2.0) * 255.0 / (c1.max() + c2.max() / 2.0)).astype('f')


# these are part of iris_sji_color_table() function and follow the same calculations but different datatype
c0 = np.arange(0, 256)
[...]

PR #4433, opened 2020-08-23
Move colormap data to .csv files

dstansby:
xref https://github.com/sunpy/sunpy/issues/3885.  This moves some colormap data into their own .csv files which makes them easier to use/human-read.


Cadair (approved):
Looks good to me

nabobalis (approved):

dstansby:
CI failures look un-related; I won't merge as is my PR, but if anyone else wants to this has two approvals.

dpshelio:
:clap: thanks @dstansby this had been bugging me for a long time! I was going to suggest to add some metadata as headers on the csv files so there's no need to depend on the filename to know which instrument it's for or where/how the colour table had been created but it was merged already. I suppose we can create an issue for that.

dstansby:
Yes, please do open an issue, that would be a very nice thing to do!

PR #4435, opened 2020-08-25
precommit update/CI updates

nabobalis:


Cadair (approved):

nabobalis:
I might look at the failing tests and update them here too.

dstansby:
Might be worth xfailing the `devdeps` test that fails, as I don't think a mpl fix is incoming in the near future (but the PR to fix it is marked as release critical for the next mpl release).

sunpy/map/mapbase.py
```diff
@@ -2049,7 +2049,7 @@ def plot(self, annotate=True, axes=None, title=True,
                               SunpyUserWarning)
 
         # Normal plot
-        plot_settings = copy.deepcopy(self.plot_settings)
+        plot_settings = copy.copy(self.plot_settings)
```
nabobalis, line 2052: This breaks things.

dstansby: Yeah, it has to be `deepcopy`.

nabobalis:
Dev deps failed trying to open the sample data.

sunpy/coordinates/ephemeris.py
```diff
@@ -285,10 +285,12 @@ def get_horizons_coord(body, time='now', id_type='majorbody', *, include_velocit
...
     try:
         result = query.vectors()
-    except Exception:  # Catch and re-raise all exceptions, and also provide query URL if generated
+    except Exception as e:  # Catch and re-raise all exceptions, and also provide query URL if generated
         if query.uri is not None:
             log.error(f"See the raw output from the JPL HORIZONS query at {query.uri}")
-        raise
+        raise e
+    finally:
+        query._session.close()
```
dstansby, line 293: 👎 👎 👎 for using a private attribute from another package, is there another way to do this?

nabobalis: I didn't have a close look at the astroquery docs but I couldn't see one. I could see if it could be done in a context manager?

Cadair: I can't see one from a quick poke around the astroquery source code.

dstansby: If the session isn't closing (or there's no public API to close the session) then this should also be reported upstream.

Cadair: it feels like the object should clean the session when it's deleted, but yeah opening an issue upstream is never a bad idea.

nabobalis:
So dev deps is still broken due to upstream MPL and there isn't anything to be done about it outside of not running the exmaples.

sunpy/coordinates/ephemeris.py
nabobalis: I will open an issue. But I see what happens if I try it as a context.

sunpy/map/mapbase.py
```diff
@@ -2049,7 +2049,11 @@ def plot(self, annotate=True, axes=None, title=True,
                               SunpyUserWarning)
 
         # Normal plot
-        plot_settings = copy.deepcopy(self.plot_settings)
+        try:
```
nabobalis, line 2052: I've been unable to get a dev MPL to work in my WSL2 setup so time to use the CI to test this.

sunpy/coordinates/ephemeris.py
nabobalis: I opened an issue, but I have no issue with using this private function as it stands.

sunpy/map/mapbase.py
nabobalis: Turns out I can't read. I installed MPL master and it doesn't seem to error.

sunpy/map/mapbase.py
```diff
@@ -2049,7 +2049,11 @@ def plot(self, annotate=True, axes=None, title=True,
...
+            plot_settings = copy.deepcopy(self.plot_settings)
+        except NotImplementedError:
```
Cadair, line 2054: I mean this feels like a horrific hack, but as long as we revert it it's better than a load of fails I suppose. Not so sure about backporting this commit though.

nabobalis: Fair about the backport but I feel like the rest does need to be backported. I can do it manually?

Cadair: yeah seems good.

sunpy/coordinates/ephemeris.py
dstansby: 🤷 if it's private for a reason so we probably shouldn't use it, and if there's no reason for it not to be private we can ask upstream to make it public. As long as someone opened an issue.

PR #4437, opened 2020-08-26
downloading_hmi now uses the VSO

nabobalis:
Let us not bother JSOC.

### DIFF
--- a/examples/acquiring_data/downloading_hmi.py
+++ b/examples/acquiring_data/downloading_hmi.py
@@ -12,34 +12,20 @@ from sunpy.net import Fido
 from sunpy.net import attrs as a
 
 ###############################################################################
-# To download the required data, we use
-# `sunpy.net.Fido`, a downloader client,
-# to query the Joint Science Operations Center, or JSOC, where HMI data are stored.
-# First define the search variables, a timerange,
-# a `data series <http://jsoc.stanford.edu/JsocSeries_DataProducts_map.html>`_,
-# keywords, and your e-mail address (to notify you when the download is complete).
-# See the JSOC e-mail address registration page
-# `here <http://jsoc.stanford.edu/ajax/register_email.html>`_.
-
-result = Fido.search(a.Time('2020/01/20 00:00:00', '2020/01/20 00:04:00'),
-                     a.jsoc.Series("hmi.M_720s"),
-                     a.jsoc.Keys(["T_REC, CROTA2"]),
-                     a.jsoc.Notify("jsoc@cadair.com"))
+# To download the required data, we use `sunpy.net.Fido`, a downloader client,
+# to query the Virtual Solar Observatory to acquire HMI data.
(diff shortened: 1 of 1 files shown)


examples/acquiring_data/downloading_hmi.py
```diff
@@ -12,34 +12,21 @@
...
-# `here <http://jsoc.stanford.edu/ajax/register_email.html>`_.
-
-result = Fido.search(a.Time('2020/01/20 00:00:00', '2020/01/20 00:04:00'),
-                     a.jsoc.Series("hmi.M_720s"),
-                     a.jsoc.Keys(["T_REC, CROTA2"]),
-                     a.jsoc.Notify("jsoc@cadair.com"))
+# To download the required data, we use `sunpy.net.Fido`, a downloader client,
+# to query the Virtual Solar Observatory to acquire HMI data.
+result = Fido.search(a.Time('2020/01/20 00:00:00', '2020/01/20 00:01:00'),
+                     a.Instrument.hmi)
```
Cadair, line 18: I think you need to add `a.Physobs.los_magnetic_field` to get the same two results?

nabobalis: Good point.

examples/acquiring_data/downloading_hmi.py
```diff
@@ -12,34 +12,20 @@
...
 ###############################################################################
 # The following shows how to download the results. If we
 # don't provide a path it will download the file into the sunpy data directory.
 # The output provides the path of the downloaded files.
-# The result can be from several data clients, so we have to index the client first
-# and then index the file.
-
-# Slice the first record returned by the first client.
-downloaded_file = Fido.fetch(result[0, 0])
+downloaded_file = Fido.fetch(result)
```
Cadair, line 28: I would leave this, we are now printing two file names where as before we showed how to only print one?

nabobalis: There is only one file in the result.

Cadair: huh ok, why does JSOC return an extra file?

nabobalis: Not a clue.

nabobalis: Current download:
```python
Results from 1 Provider:

1 Results from the VSOClient:
   Start Time [1]       End Time [1]    Source Instrument   Type    Wavelength [2]
                                                                       Angstrom
------------------- ------------------- ------ ---------- -------- ----------------
2020-01-20 00:00:22 2020-01-20 00:00:23    SDO        HMI FULLDISK 6173.0 .. 6174.0


Files Downloaded: 100%|█████████████████████████████████████████████████████████████████| 1/1 [00:01<00:00,  1.91s/file]
['/home/nabil/sunpy/data/hmi_m_45s_2020_01_20_00_01_30_tai_magnetogram.fits']
```

Cadair (approved):

PR #4446, opened 2020-08-28
Faster contains_full_disk

wafels:
### Description

This PR speeds up checking that a map contains the full disk.  I get about a ~20x speed up (2016 MacBook Pro, 13'')

This PR is a direct result of the AIApy acceleration project, part of HelioHackweek 2020 - see https://github.com/heliohackweek/AIA_acceleration/issues/2

### DIFF
--- a/sunpy/map/maputils.py
+++ b/sunpy/map/maputils.py
@@ -179,10 +179,10 @@ def contains_full_disk(smap):
     top_, bottom, left_hand_side, right_hand_side = map_edges(smap)
 
     # Coordinates of the edges
-    horizontal1 = smap.pixel_to_world(_convert_pixels(top_))
-    horizontal2 = smap.pixel_to_world(_convert_pixels(bottom))
-    vertical1 = smap.pixel_to_world(_convert_pixels(left_hand_side))
-    vertical2 = smap.pixel_to_world(_convert_pixels(right_hand_side))
+    horizontal1 = smap.pixel_to_world(*_convert_pixels(top_))
+    horizontal2 = smap.pixel_to_world(*_convert_pixels(bottom))
+    vertical1 = smap.pixel_to_world(*_convert_pixels(left_hand_side))
+    vertical2 = smap.pixel_to_world(*_convert_pixels(right_hand_side))
 
     # The radius of the Sun
     radius = smap.rsun_obs
--- a/sunpy/map/maputils.py
+++ b/sunpy/map/maputils.py
@@ -14,7 +14,7 @@ __all__ = ['all_pixel_indices_from_map', 'all_coordinates_from_map',
            'map_edges', 'solar_angular_radius', 'sample_at_coords',
            'contains_full_disk', 'is_all_off_disk', 'is_all_on_disk',
            'contains_limb', 'coordinate_is_on_solar_disk',
-           'on_disk_bounding_coordinates']
+           'on_disk_bounding_coordinates', '_convert_pixels']
 
 
 def all_pixel_indices_from_map(smap):
@@ -124,6 +124,28 @@ def sample_at_coords(smap, coordinates):
     return smap.data[smap.wcs.world_to_array_index(coordinates)]
 
 
+def _convert_pixels(edge):
+    """
+    Helper function to convert a list of edge pixels of the form [(x0, y0), (x1, y1),..., (xn,yn)]
+    into the form ([x0, x1,...,xn], [y0, y1,...,yn])
+
+    Parameter
+    ---------
+    edge : ~list
+        A list of pairs of tuples.
+
+    Returns
+    -------
+    pixel_edge : ~tuple
+        A tuple containing two lists.  The first entry is a list containing the first coordinate of
+        each pixel, and similarly for the second entry.
+
+    """
+    x = [p[0] for p in edge.value] * u.pix
+    y = [p[1] for p in edge.value] * u.pix
+    return x, y
+
+
 def contains_full_disk(smap):
     """
     Checks if a map contains the full disk of the Sun.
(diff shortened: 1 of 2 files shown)


sunpy/map/maputils.py
```diff
@@ -124,15 +124,38 @@ def sample_at_coords(smap, coordinates):
...
 
 
+def _convert_pixels(edge):
+    """
+    Helper function to convert a list of edge pixels of the form [(x0, y0), (x1, y1),..., (xn,yn)]
+    into the form ([x0, x1,...,xn], [y0, y1,...,yn])
+
+    Parameter
+    ---------
+    edge : ~list
```
nabobalis, line 134:
```suggestion
    edge : list
```

sunpy/map/maputils.py
```diff
@@ -124,15 +124,38 @@ def sample_at_coords(smap, coordinates):
...
+        A list of pairs of tuples.
+
+    Returns
+    -------
+    pixel_edge : ~tuple
```
nabobalis, line 139:
```suggestion
    tuple
```

sunpy/map/maputils.py
```diff
@@ -124,15 +124,38 @@ def sample_at_coords(smap, coordinates):
...
+        A tuple containing two lists.  The first entry is a list containing the first coordinate of
```
nabobalis, line 140:
```suggestion
        A tuple containing two lists. The first entry is a list containing the first coordinate of
```

sunpy/map/maputils.py
```diff
@@ -124,15 +124,38 @@ def sample_at_coords(smap, coordinates):
...
+        each pixel, and similarly for the second entry.
+
```
nabobalis, line 142:
```suggestion
```

sunpy/map/maputils.py
```diff
@@ -153,21 +176,66 @@ def contains_full_disk(smap):
...
-
-    # Test if all the edge pixels are more than one solar radius distant
-    # and that the whole map is not all off disk.
-    return np.all(coordinate_angles > solar_angular_radius(edge_of_world)) and ~is_all_off_disk(smap)
+    # Get the edge pixels
+    top_, bottom, left_hand_side, right_hand_side = map_edges(smap)
+
+    # Coordinates of the edges
+    x, y = _convert_pixels(top_)
+    horizontal1 = smap.pixel_to_world(x, y)
```
nabobalis, line 184:
```suggestion
    horizontal1 = smap.pixel_to_world(_convert_pixels(top_))
```
Is there a better name than `horizontal1`? (The same applies for the other lines.

sunpy/map/maputils.py
wafels: Fixed

sunpy/map/maputils.py
wafels: Fixed

sunpy/map/maputils.py
wafels: Fixed

sunpy/map/maputils.py
wafels: Fixed

sunpy/map/maputils.py
```diff
@@ -153,21 +175,59 @@ def contains_full_disk(smap):
...
-    coordinate_angles = np.sqrt(edge_of_world.Tx ** 2 + edge_of_world.Ty ** 2)
-
-    # Test if all the edge pixels are more than one solar radius distant
-    # and that the whole map is not all off disk.
-    return np.all(coordinate_angles > solar_angular_radius(edge_of_world)) and ~is_all_off_disk(smap)
+    # Get the edge pixels
+    top_, bottom, left_hand_side, right_hand_side = map_edges(smap)
+
+    # Coordinates of the edges
+    horizontal1 = smap.pixel_to_world(*_convert_pixels(top_))
```
nabobalis, line 182: I forgot about the need to unpack it.

sunpy/map/maputils.py
```diff
@@ -153,21 +175,59 @@ def contains_full_disk(smap):
...
+        rhs = v1
+    elif np.all(v1 < -radius):
+        lhs = v1
+
+    v2 = vertical2.Tx
+    if np.all(v2 > radius):
+        rhs = v2
+    elif np.all(v2 < -radius):
+        lhs = v2
+
```
nabobalis, line 224: These basically repeat but changed out for a different variable. There should be a neater way to write this but I don't have an answer at the moment.

sunpy/map/maputils.py
wafels: I struggled coming up with these names.  For a N x N image, the `top_ `pixels are (0, N-1),...(N-1, N-1), that is, the second axis value is constant at its maximum value.  A more descriptive name would be `second_axis_value_is_max`.   Following this scheme, we also have `second_axis_value_is_min`, `first_axis_value_is_max` and `first_axis_value_is_min`.  Is this better?

sunpy/map/maputils.py
wafels: Fixed.

sunpy/map/maputils.py
wafels: Yeah, I agree.  I wrote the function quickly and in a long-winded way to support the hackweek.

sunpy/map/maputils.py
```diff
@@ -124,15 +124,37 @@ def sample_at_coords(smap, coordinates):
...
+    Returns
+    -------
+    tuple
+        A tuple containing two lists. The first entry is a list containing the first coordinate of
+        each pixel, and similarly for the second entry.
+    """
+    x = [p[0] for p in edge.value] * u.pix
+    y = [p[1] for p in edge.value] * u.pix
+    return x, y
```
ayshih, line 145: This looks slow (performance-wise).  At the very least, the list comprehension should not be done twice.

ayshih:
If you want performance, I think you'd do better if you reversed the check (at least for high-resolution images.  Instead of making sure all of the edges of the map are at least a solar radius away, how about creating an array coordinate for the limb of the Sun and seeing if it all lands in the array when converted into pixel space?

wafels:
I'll try that.

sunpy/map/maputils.py
wafels: I could just replace `map_edges` with a function that returns the pixels in the form that they are needed.

Cadair: isn't this `zip(egde.value)`? or the reverse of that maybe?

nabobalis:
If we can benchmark this using https://github.com/joerick/pyinstrument would be nice.

nabobalis:
I think that #4463 has become more advanced and supersedes this.

Issue #1753, opened 2016-04-25
Add support for Map's units parameter


ehsteve:
Since Map uses [NDData](http://docs.astropy.org/en/latest/api/astropy.nddata.NDData.html) they can carry a units parameter which is currently not filled in when a file is opened. 

This information could be stored in the file header and could be passed into Map when we load a file.

nadgeri14:
I would like to work on this problem. Could I get some details on how to proceed forward ?

agneet42:
@ehsteve I'd like to work on this. Could you help me by letting me know which file to start looking at?

swapsha96:
@Cadair @dpshelio I would like to fix this issue. Can you suggest me which unit should I set as default? An example in [NDData documentation](http://docs.astropy.org/en/stable/api/astropy.nddata.NDData.html) uses 'u.cm'. Should I create a PR setting default as u.cm?

swapsha96:
@ehsteve @Cadair I would like to work on this.

dstansby:
Is there a standard FITS keyword for data units?

Cadair:
`BUNIT`?

dstansby:
I'll take a look at this for 2.1. There's a couple of places (e.g. contouring) where it would be good to add support for unitful-data, so I think some sort of global review of this will be helpful.

PR #4451, opened 2020-09-02
Use bunit in sunpy.map

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/1753

Several design decisions I have made, that are up for discussion:
- Override `NDData`s `unit()` method, and use `self.meta['bunit']` as the immutable ground truth
- Warn if 'bunit' is present but cannot be parsed
- Return `None` if `bunit` not present or can't be parsed

Still needs tests and a changelog.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -45,6 +45,8 @@ TIME_FORMAT = config.get("general", "time_format")
 PixelPair = namedtuple('PixelPair', 'x y')
 SpatialPair = namedtuple('SpatialPair', 'axis1 axis2')
 
+_META_FIX_URL = 'https://docs.sunpy.org/en/stable/code_ref/map.html#fixing-map-metadata'
+
 __all__ = ['GenericMap']
 
 
@@ -585,10 +587,13 @@ class GenericMap(NDData):
         if unit_str is None:
             return
         else:
-            unit = u.Unit(unit_str, parse_strict='warn')
+            unit = u.Unit(unit_str, format='fits', parse_strict='silent')
             if isinstance(unit, u.UnrecognizedUnit):
-                warnings.warn(f'Could not parse unit string "{unit_str}". '
-                              'See the above warning for more information',
+                warnings.warn(f'Could not parse unit string "{unit_str}" as a valid FITS unit.\n'
+                              f'See {_META_FIX_URL} for how to fix metadata before loading it '
+                              'with sunpy.map.Map.\n'
+                              'See https://fits.gsfc.nasa.gov/fits_standard.html for'
+                              'the FITS unit standards.',
                               SunpyMetadataWarning)
                 unit = None
             return unit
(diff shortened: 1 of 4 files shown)


sunpy/map/mapbase.py
```diff
@@ -572,6 +572,27 @@ def max(self, *args, **kwargs):
...
+
+        This is taken from the 'BUNIT' FITS keyword. If no 'BUNIT' entry is
+        present in the metadata then this returns `None`. If the 'BUNIT' value
+        cannot be parsed into a unit a warning is raised, and `None` returned.
+        """
+        unit_str = self.meta.get('bunit', None)
+        if unit_str is None:
+            return
+        else:
+            unit = u.Unit(unit_str, parse_strict='warn')
```
Cadair, line 588: We should really be parsing these units with just the FITS unit convention?
```suggestion
            unit = u.Unit(unit_str, format="fits", parse_strict='warn')
```

sunpy/map/mapbase.py
```diff
@@ -572,6 +572,27 @@ def max(self, *args, **kwargs):
...
+            if isinstance(unit, u.UnrecognizedUnit):
+                warnings.warn(f'Could not parse unit string "{unit_str}". '
+                              'See the above warning for more information',
```
Cadair, line 591: Would it not be cleaner to capture and parse the astropy warning, it's pretty long and mostly irrelevant for the map user here (i.e. it suggests things they might mean, which isn't really how you interact with a FITS header?)

dstansby: 👍 maybe a link to the FITS unit standard would be good in this warning so a user can go and look up and alternative to fix the metadata

dstansby: How about now?

dpshelio (approved):

sunpy/map/mapbase.py
```diff
@@ -572,6 +574,30 @@ def max(self, *args, **kwargs):
...
+        Unit of the map data.
+
+        This is taken from the 'BUNIT' FITS keyword. If no 'BUNIT' entry is
+        present in the metadata then this returns `None`. If the 'BUNIT' value
+        cannot be parsed into a unit a warning is raised, and `None` returned.
+        """
+        unit_str = self.meta.get('bunit', None)
+        if unit_str is None:
+            return
+        else:
```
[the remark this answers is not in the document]
Cadair: I would do the fail early thing, but drop the else:
```suggestion
        if unit_str is None:
            return
```

As you don't need it.

Cadair (approved):

Cadair:
Looks like the RTD fail is real:

> WARNING: :88: (ERROR/3) Unknown interpreted text role "method".

nabobalis:
I can't see where that is in the diff.

Cadair:
It's possible it snuck into master with all the examples being on fire?

Cadair:
yeah, the RTD fail is not specific to this PR.

nabobalis:
> It's possible it snuck into master with all the examples being on fire?

This makes sense. I will hunt it down.

PR #2041, opened 2017-03-15
Remove unnecessary code from image.transform.affine_transform

larrymanley:

PR #4452, opened 2020-09-03
Don't convert float32 to float64 in affine_transform

dstansby:
Large (e.g. AIA) maps can take up lots of memory (100MB+). As long as accuracy can be sacrificed, these can be converted from 64 to 32 bit floats to save memory. However, `affine_transform` casts anything that isn't float64 up to float64. Looking at the git history this seems to be to avoid passing integer dytpes to `warp`, but I don't think this needs to be done for float32.

This PR changes the casting check to make it so that only integer dtypes are cast to float64, avoiding float32 being cast up and reducing memory usage.

### DIFF
--- a/sunpy/image/transform.py
+++ b/sunpy/image/transform.py
@@ -1,6 +1,7 @@
 """
 Functions for geometrical image transformation and warping.
 """
+import numbers
 import warnings
 
 import numpy as np
@@ -113,9 +114,10 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
         skmatrix[:2, 2] = shift
         tform = skimage.transform.AffineTransform(skmatrix)
 
-        # Transform the image using the skimage function
-        if not np.issubdtype(image.dtype, np.float64):
-            warnings.warn("Input data has been cast to float64.", SunpyUserWarning)
+        if issubclass(image.dtype.type, numbers.Integral):
+            warnings.warn("Integer input data has been cast to float64, "
+                          "which is required for the skikit-image transform.",
+                          SunpyUserWarning)
             adjusted_image = image.astype(np.float64)
         else:
             adjusted_image = image.copy()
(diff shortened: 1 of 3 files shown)


ayshih:
I believe the casting from integer to float is a vestige from when we were normalizing the data array before passing it to `warp()`, which apparently is no longer needed.  `warp()` should work fine on integer arrays because it will internally convert them to float arrays, although I believe we'll want to use the `preserve_range` keyword.

In other words, I'd strip out the casting code entirely, add `preserve_range=True` to `warp()`, and see if everything still works.

ayshih:
A bit tangential to this PR, but...

#2041 removed the normalization code and cast everything to float64.  However, I believe the removal of the normalization code should not have been done.  Scikit-image's documentation is quite clear that it always assumes that floats are in the range of [0, 1].  While `warp()` does appear to work fine when that assumption is violated, we are currently relying on undefined behavior.

I suggest that the normalization code be re-inserted.

sunpy/image/transform.py
```diff
@@ -106,25 +107,35 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
...
-        if not np.issubdtype(image.dtype, np.float64):
-            warnings.warn("Input data has been cast to float64.", SunpyUserWarning)
-            adjusted_image = image.astype(np.float64)
-        else:
-            adjusted_image = image.copy()
-        if np.any(np.isnan(adjusted_image)) and order >= 4:
+        tform = _affine_transform_matrix(rmatrix, shift)
+
+        rotated_image = image.copy()
+        # skikit-image expects floating point images to be in the range -1, 1
```
wtbarnes, line 113:
```suggestion
        # scikit-image expects floating point images to be in the range 0, 1
```

dstansby: Hmm, perhaps we have to be cleverer to deal with negative values too.

dstansby:
I had a play around with suggestions, and decided they were more complicated than they seem to implement. In the interests of fixing the original bug, and maintaining other behaviour so this can be backported to 2.0.x, I have kept this simple and opened https://github.com/sunpy/sunpy/issues/4453.

nabobalis (approved):

Cadair (approved):

nabobalis:
Failures are server related.

PR #4463, opened 2020-09-10
Maputil speedups

dstansby:
This is an attempt at something simpler than https://github.com/sunpy/sunpy/pull/4446 to speed up `contains_full_disk` (and other maputils functions). Times are for a 4k AIA image.

*`contains_full_disk`*
The bottleneck is running `~is_all_off_disk(smap)`, which is unnescessary. If the solar limb doesn't intersect the edges of the map (the first test), then the full disc is in the map iff (0, 0) is in the map. This is pretty easy to test for by just checking if both of the corner coordinates change sign, which is what this PR does.
11s > 0.3s

*`is_all_off_disk`*
Again, do not need to check if all the pixels are off disc, it's sufficient to check that the edges are off disk (ie. more than a solar radius away) and the solar center isn't in the image.
8s > 0.3s

*`is_all_on_disk`*
This is only true if all the edges are on the disc.
7s > 0.1s

*`contains_limb`*
Shortcut the case where the full disc is in the map. If not, it is sufficient to check if some edge pixels are both on and off the disc.
7s > 0.3s

A typical profile now looks like
```python
0.129 <module>  test.py:1
└─ 0.129 is_all_on_disk  sunpy/map/maputils.py:260
   ├─ 0.123 _edge_coordinates  sunpy/map/maputils.py:127
   │  ├─ 0.103 wrapper  astropy/units/decorators.py:165
   │  │  └─ 0.103 pixel_to_world  sunpy/map/mapbase.py:1099
   │  │     ├─ 0.049 coordinate_frame  sunpy/map/mapbase.py:482
   │  │     │  ├─ 0.042 wcs  sunpy/map/mapbase.py:421
   │  │     │  │  ├─ 0.015 __init__  astropy/wcs/wcs.py:359
   │  │     │  │  │     [31 frames hidden]  astropy
   │  │     │  │  ├─ 0.015 fits_header  sunpy/map/mapbase.py:983
   │  │     │  │  │  └─ 0.015 header_to_fits  sunpy/io/fits.py:194
   │  │     │  │  │     ├─ 0.007 __init__  astropy/io/fits/card.py:156
   │  │     │  │  │     │     [24 frames hidden]  astropy
   │  │     │  │  │     ├─ 0.004 append  astropy/io/fits/header.py:1157
   │  │     │  │  │     │     [5 frames hidden]  astropy
   │  │     │  │  │     └─ 0.003 [self]  
   │  │     │  │  └─ 0.010 observer_coordinate  sunpy/map/mapbase.py:830
   │  │     │  │     └─ 0.009 __getattr__  astropy/coordinates/sky_coordinate.py:577
   │  │     │  │           [32 frames hidden]  astropy
   │  │     │  │              0.008 __call__  astropy/coordinates/transformations.py:1208
   │  │     │  │              ├─ 0.004 wrapped_func  sunpy/coordinates/transformations.py:154
   │  │     │  │              │  └─ 0.004 hcrs_to_hgs  sunpy/coordinates/transformations.py:531
   │  │     │  └─ 0.006 wcs_to_celestial_frame  astropy/wcs/utils.py:182
   │  │     │        [3 frames hidden]  astropy
   │  │     │           0.005 solar_wcs_frame_mapping  sunpy/coordinates/wcs_utils.py:16
   │  │     │           └─ 0.005 __init__  sunpy/coordinates/frames.py:118
   │  │     │              └─ 0.005 __init__  astropy/coordinates/baseframe.py:415
   │  │     │                 └─ 0.005 __get__  sunpy/coordinates/frameattributes.py:129
   │  │     │                    └─ 0.004 __get__  astropy/coordinates/attributes.py:95
   │  │     │                       └─ 0.004 convert_input  sunpy/coordinates/frameattributes.py:100
   │  │     │                          └─ 0.004 convert_input  astropy/coordinates/attributes.py:415
   │  │     │                                [7 frames hidden]  astropy
   │  │     │                                   0.002 __call__  astropy/coordinates/transformations.py:913
   │  │     │                                   └─ 0.002 wrapped_func  sunpy/coordinates/transformations.py:154
   │  │     │                                      └─ 0.002 hgs_to_hgs  sunpy/coordinates/transformations.py:612
   │  │     ├─ 0.046 wcs  sunpy/map/mapbase.py:421
   │  │     │  ├─ 0.016 fits_header  sunpy/map/mapbase.py:983
   │  │     │  │  └─ 0.016 header_to_fits  sunpy/io/fits.py:194
   │  │     │  │     ├─ 0.008 __init__  astropy/io/fits/card.py:156
   │  │     │  │     │     [25 frames hidden]  astropy, <built-in>
   │  │     │  │     └─ 0.006 append  astropy/io/fits/header.py:1157
   │  │     │  │           [3 frames hidden]  astropy
   │  │     │  ├─ 0.014 __init__  astropy/wcs/wcs.py:359
   │  │     │  │     [37 frames hidden]  astropy, <built-in>
   │  │     │  ├─ 0.012 observer_coordinate  sunpy/map/mapbase.py:830
   │  │     │  │  └─ 0.010 __getattr__  astropy/coordinates/sky_coordinate.py:577
   │  │     │  │        [35 frames hidden]  astropy, <__array_function__ internal...
   │  │     │  │           0.009 __call__  astropy/coordinates/transformations.py:1208
   │  │     │  │           ├─ 0.004 wrapped_func  sunpy/coordinates/transformations.py:154
   │  │     │  │           │  └─ 0.004 hcrs_to_hgs  sunpy/coordinates/transformations.py:531
   │  │     │  │           │     └─ 0.002 _rotation_matrix_reprs_to_xz_about_z  sunpy/coordinates/transformations.py:491
   │  │     │  │           │        └─ 0.002 _rotation_matrix_reprs_to_reprs  sunpy/coordinates/transformations.py:470
   │  │     │  └─ 0.002 date  sunpy/map/mapbase.py:633
   │  │     │     └─ 0.002 parse_time  sunpy/time/time.py:253
   │  │     │        └─ 0.002 wrapper  functools.py:870
   │  │     │           └─ 0.002 convert_time_str  sunpy/time/time.py:198
   │  │     ├─ 0.004 __init__  astropy/coordinates/sky_coordinate.py:210
   │  │     │     [5 frames hidden]  astropy, <built-in>
   │  │     │        0.002 __init__  sunpy/coordinates/frames.py:118
   │  │     │        └─ 0.002 __init__  astropy/coordinates/baseframe.py:415
   │  │     │           └─ 0.002 __get__  sunpy/coordinates/frameattributes.py:129
   │  │     │              └─ 0.002 __get__  astropy/coordinates/attributes.py:95
   │  │     │                    [3 frames hidden]  astropy
   │  │     │        0.001 __get__  astropy/coordinates/attributes.py:95
   │  │     │        └─ 0.001 convert_input  sunpy/coordinates/frameattributes.py:32
   │  │     │           └─ 0.001 __eq__  astropy/time/core.py:2044
   │  │     │                 [7 frames hidden]  astropy
   │  │     └─ 0.004 wcs_pix2world  astropy/wcs/wcs.py:1386
   │  │           [6 frames hidden]  astropy, <built-in>
   │  └─ 0.020 map_edges  sunpy/map/maputils.py:57
   │     └─ 0.020 __rmul__  astropy/units/core.py:701
   │           [4 frames hidden]  astropy, <built-in>
   └─ 0.006 wrapper  astropy/units/decorators.py:165
      └─ 0.006 coordinate_is_on_solar_disk  sunpy/map/maputils.py:198
         ├─ 0.003 __getattr__  astropy/coordinates/sky_coordinate.py:577
         │     [2 frames hidden]  astropy
         │        0.003 __getattr__  astropy/coordinates/baseframe.py:1527
         │        └─ 0.003 represent_as  sunpy/coordinates/frames.py:133
         │           └─ 0.002 wrap_angle  astropy/coordinates/angles.py:652
         │                 [4 frames hidden]  astropy
         └─ 0.002 wrapper  astropy/units/decorators.py:165
               [7 frames hidden]  astropy, <built-in>
```


nabobalis:
What is the rest of the time spent on?

sunpy/map/maputils.py
```diff
@@ -153,21 +164,25 @@ def contains_full_disk(smap):
...
-    return np.all(coordinate_angles > solar_angular_radius(edge_of_world)) and ~is_all_off_disk(smap)
+    return np.all(coordinate_angles > solar_angular_radius(edge_of_world)) and contains_solar_center(smap)
+
+
+def contains_solar_center(smap):
+    """
+    Returns `True` if smap contains the solar center.
+    """
+    bottom_left = smap.pixel_to_world(-0.5 * u.pix, -0.5 * u.pix)
+    top_right = smap.pixel_to_world(*(u.Quantity(smap.dimensions) - 0.5 * u.pix))
```
Cadair, line 181: doesn't map have a `top_right` and `bottom_left` property?

Cadair (commented):
This is awesome!

Two things I can think of:

* `contains_solar_center` might be able to be fooled if you throw it a really large field of view, like the whole sky?
* Are we just assuming all maps are in HPC in maputils already or is that an assumption you have made in this diff?

sunpy/map/maputils.py
```diff
@@ -124,6 +124,17 @@ def sample_at_coords(smap, coordinates):
     return smap.data[smap.wcs.world_to_array_index(coordinates)]
 
 
+def _edge_coordinates(smap):
+    # Calculate all the edge pixels
+    edges = map_edges(smap)
+    edge_pixels = list(chain.from_iterable([edges[0], edges[1], edges[2], edges[3]]))
+    x = [p[0] for p in edge_pixels] * u.pix
+    y = [p[1] for p in edge_pixels] * u.pix
```
Cadair, line 132: as @ayshih pointed out on #4446 doing this comprehension twice is a waste, also this is just a `zip` (I think)

```suggestion
    x, y = list(zip(edge_pixels)) * u.pix
```

sunpy/map/maputils.py
```diff
@@ -124,6 +124,17 @@ def sample_at_coords(smap, coordinates):
...
+
+    # Calculate the edge of the world
+    return smap.pixel_to_world(x, y)
```
Cadair, line 135: Having said that wouldn't casting to a numpy array be a better idea?

```suggestion
    edge_pixels = list(chain.from_iterable([edges[0], edges[1], edges[2], edges[3]])) * u.pix
    return smap.pixel_to_world(*edge_pixels.T)
```

sunpy/map/maputils.py
```diff
@@ -153,21 +164,25 @@ def contains_full_disk(smap):
...
 
     # Test if all the edge pixels are more than one solar radius distant
     # and that the whole map is not all off disk.
-    return np.all(coordinate_angles > solar_angular_radius(edge_of_world)) and ~is_all_off_disk(smap)
+    return np.all(coordinate_angles > solar_angular_radius(edge_of_world)) and contains_solar_center(smap)
+
+
+def contains_solar_center(smap):
+    """
+    Returns `True` if smap contains the solar center.
```
Cadair, line 178: *mutters something about a fully fleshed out docstring :wink: *

sunpy/map/maputils.py
dstansby: Yes, but they're the coordinates of the centers of the pixels, not the edges.

Cadair: Ah right 👍

sunpy/map/maputils.py
dstansby: Done something slightly different, but shaved around 50% of the new times 🕐

dstansby:
> contains_solar_center might be able to be fooled if you throw it a really large field of view, like the whole sky?

Shouldn't be fooled, as by definition of the coordinate origin is the center of the Sun.

> Are we just assuming all maps are in HPC in maputils already or is that an assumption you have made in this diff?

That assumption is all over maputils already. Wouldn't be a bad idea to open an issue/PR to validate the map and throw a nice error if it's not HPC tho.

Cadair (approved):
Looks good to me, I wonder if @wafels or @ayshih want to take a look?

sunpy/map/maputils.py
```diff
@@ -124,6 +124,17 @@ def sample_at_coords(smap, coordinates):
     return smap.data[smap.wcs.world_to_array_index(coordinates)]
 
 
+def _edge_coordinates(smap):
+    edges = map_edges(smap)
+    # Calculate all the edge pixels
+    edge_pixels = np.concatenate(edges)
+    # Need this line for compatability with old versions of astropy, where
+    # edge_pixels is not a full on Quantity. This can be removed in the future
+    edge_pixels = edge_pixels.value * edges[0].unit
```
ayshih, line 133: (I can't check myself right now...)  What is `edge_pixels` if it's not a `Quantity`?  What happens if you instead used:
```python
edge_pixels = np.concatenate(edges) << edges[0].unit
```

dstansby: I'm not sure on old versions of astropy, but this was what I had to do to get the oldtestdeps build to pass. Your suggestion seems to work fine (on new versions of astropy), but I've never seen `<<` before (the internet says it is a bitwise shift operator?) and I think the current version is a bit clearer

Cadair: astropy overloaded `<<` to mean "convert to quantity without copying the array" iirc.

Cadair: https://docs.astropy.org/en/stable/units/quantity.html#astropy-units-quantity-no-copy

ayshih: It doesn't have to be the insertion/left-shift operator (`<<`), which is admittedly unusual to those who aren't familiar with C++ streams or with bit operations.  It could instead be the `Quantity()` initializer:
```python
edge_pixels = Quantity(np.concatenate(edges), edges[0].unit, copy=False)
```
I don't like your current code pattern because it utilizes specific aspects of the API rather than abstractly requesting what you want, it creates a copy when one isn't needed, and it performs an operation even when `edge_pixels` is already a `Quantity`.

ayshih: I'm still not clear what `edge_pixels` is if it's not a `Quantity`; why does `edge_pixels.value` work in that case?

dstansby: Some investigation; the problem occurs with our oldest supported version of numpy (`1.16.6`), but is not an issue for `numpy>=1.17`. The `<<` suggestion doesn't work (where the current PR does), failing with
```
  File "test.py", line 10, in <module>
    coords = sunpy.map.is_all_on_disk(m)
  File "/Users/dstansby/github/sunpy/sunpy/map/maputils.py", line 285, in is_all_on_disk
    edge_of_world = _edge_coordinates(smap)
  File "/Users/dstansby/github/sunpy/sunpy/map/maputils.py", line 133, in _edge_coordinates
    edge_pixels = np.concatenate(edges) << edges[0].unit
  File "/Users/dstansby/miniconda3/envs/dev/lib/python3.8/site-packages/astropy/units/quantity.py", line 887, in __lshift__
    return self.__class__(self, other, copy=False, subok=True)
  File "/Users/dstansby/miniconda3/envs/dev/lib/python3.8/site-packages/astropy/units/quantity.py", line 300, in __new__
    value = value.to(unit)
  File "/Users/dstansby/miniconda3/envs/dev/lib/python3.8/site-packages/astropy/units/quantity.py", line 689, in to
    return self._new_view(self._to_value(unit, equivalencies), unit)
  File "/Users/dstansby/miniconda3/envs/dev/lib/python3.8/site-packages/astropy/units/quantity.py", line 660, in _to_value
    return self.unit.to(unit, self.view(np.ndarray),
AttributeError: 'NoneType' object has no attribute 'to'
```
`edge_pixels` is of type `<class 'astropy.units.quantity.Quantity'>`, but for some reason does not have a unit attribute properly propagated to it during the `np.concatenate` call.

wtbarnes: I've noticed that this can happen with some numpy and scipy functions. They will return a `Quantity` but for some reason the units will be stripped off and it has never been completely clear to me why. Maybe at this stage you should explicitly strip units off of `edges` before passing to `np.concatenate` so that it is completely clear that the `edge_pixels` returned in L30 is a numpy array and then just remove the `.value` in L33.

sunpy/map/maputils.py
```diff
@@ -124,6 +124,17 @@ def sample_at_coords(smap, coordinates):
     return smap.data[smap.wcs.world_to_array_index(coordinates)]
 
 
+def _edge_coordinates(smap):
+    # We need to strip the units from edges before handing it to np.concatenate,
+    # as the .unit attribute is not propagated in np.concatenate for numpy<1.17
+    # When sunpy depends on numpy>=1.17 this unit replacing code can be removed
+    edges = [edge.to_value(u.pix) for edge in map_edges(smap)]
+    # Calculate all the edge pixels
+    edge_pixels = np.concatenate(edges) * u.pix
```
Cadair, line 133: It's a marginal improvement, but seeing how this PR is about perf, we might as well not copy the array one more time.
```suggestion
    edge_pixels = u.Quantity(np.concatenate(edges), unit=u.pix, copy=False)
```

Cadair (approved):
Looks good to me.

changelog/4463.trivial.rst
```diff
@@ -0,0 +1,9 @@
+The following functions in `sunpy.map` have had their performance greatly increased,
+with runtimes typically improving by a factor of 20x. This has been achieved by
+improving many of the checks so that they only require checking the edge pixels of a
+map as opposed to all of the pixels.
+
+- `~sunpy.map.contains_full_disk`
+- `~sunpy.map.is_all_off_disk`
+- `~sunpy.map.is_all_on_disk`
+- `~sunpy.map.contains_limb`
```
ayshih, line 9:
```suggestion
- :func:`~sunpy.map.contains_full_disk`
- :func:`~sunpy.map.is_all_off_disk`
- :func:`~sunpy.map.is_all_on_disk`
- :func:`~sunpy.map.contains_limb`
```

sunpy/map/maputils.py
```diff
@@ -124,6 +124,17 @@ def sample_at_coords(smap, coordinates):
     return smap.data[smap.wcs.world_to_array_index(coordinates)]
 
 
+def _edge_coordinates(smap):
+    # We need to strip the units from edges before handing it to np.concatenate,
+    # as the .unit attribute is not propagated in np.concatenate for numpy<1.17
+    # When sunpy depends on numpy>=1.17 this unit replacing code can be removed
+    edges = [edge.to_value(u.pix) for edge in map_edges(smap)]
+    # Calculate all the edge pixels
+    edge_pixels = u.Quantity(np.concatenate(edges), unit=u.pix, copy=False)
```
ayshih, line 133: If performance is the driving concern here, why strip the unit only to put it back, even when such an operation is not needed (i.e., newer versions of NumPy)?  Especially a list comprehension!  Maybe the performance hit is negligible in actuality, but it looks silly.

How about using a private function:
```suggestion
    edge_pixels = np.concatenate(map_edges(smap))
    # The .unit attribute is not propagated in np.concatenate for numpy<1.17
    # When sunpy depends on numpy>=1.17 this unit replacing code can be removed
    if edge_pixels.unit != u.pix:
        edge_pixels._set_unit(u.pix)
```

ayshih (changes requested):

sunpy/map/maputils.py
dstansby: I think the performance hit is tiny, there's only 4 items in `map_edges(smap)`, and as profiled the bottleneck is now (re-)computing the WCS. In addition:

- I (personally) refuse to use private API from another library
- I don't think the code snippet would work because the `.unit` attribute doesn't exist for older versions of astropy

Cadair: Do you actually need to strip the units? Can't you just run whatever the output of `np.concatenate` is through `u.Quantity` and let that fix it?

dstansby: I originally did that but changed it in response to https://github.com/sunpy/sunpy/pull/4463#discussion_r491718451

Cadair, line 133: You can drop the list comprehension:

```suggestion
    # Calculate all the edge pixels
    edge_pixels = u.Quantity(np.concatenate(edges).value, unit=u.pix, copy=False)
```
works with both old and new numpy's and avoids all copies.

ayshih: I believe that `.value` creates a copy

ayshih: > * I don't think the code snippet would work because the `.unit` attribute on `edge_pixels` doesn't exist for older versions of numpy after going through `np.concatenate`

Well, that's broken then.  `.unit` is a property, so that means that `._unit` is `None`.  You could check for `None` instead, and I think the use of private API is justifiable in this case.

dstansby:
Thanks for the reviews! I got a bit confused by them all, so I've resolved all the conversations, tidied up @Cadair's latest suggestion, and force pushed away extra commits. If anyone has more reviews, please could you open new inline threads in the code?

wtbarnes (approved):

nabobalis:
Build warning is due to another PR (woops on my part)

sunpy/map/maputils.py
Cadair: Why would it do that and not just give you the underlying buffer :sob:

Cadair:
Thoughts on backporting this?

This isn't an API change, so could be backported without affecting users. Are we willing to establish a policy of performance fixes being considered bug fixes and being backported where there is no behaviour changes associated with them?

Cadair (approved):

sunpy/map/maputils.py
nabobalis: Don't know if any of this is helpful info: https://docs.astropy.org/en/stable/units/quantity.html#creating-and-converting-quantities-without-copies

nabobalis:
> Thoughts on backporting this?
> 
> This isn't an API change, so could be backported without affecting users. Are we willing to establish a policy of performance fixes being considered bug fixes and being backported where there is no behaviour changes associated with them?

It was why I did initially put on the backport label.

Cadair:
(Test failures seem to be unrelated)

I think I am +1 on backporting perf fixes, it seems to me to be pointless to hold them back from people. I am open to counter arguments thought.

ayshih:
> I think I am +1 on backporting perf fixes, it seems to me to be pointless to hold them back from people. I am open to counter arguments thought.

What if users are depending on the slower performance?!? =P  (I jest, but sometimes a significant performance improvement can break user code, e.g., if there are parallel operations that implicitly depended on timing.)

+1 to backporting

sunpy/map/maputils.py
ayshih: > Why would it do that and not just give you the underlying buffer 😭

So people don't muck up the underlying buffer?

Cadair:
@ayshih you still have a request changes on this PR, you good with it now?

sunpy/map/maputils.py
```diff
@@ -153,21 +164,39 @@ def contains_full_disk(smap):
...
+    return np.all(coordinate_angles > solar_angular_radius(edge_of_world)) and contains_solar_center(smap)
+
+
+def contains_solar_center(smap):
+    """
+    Returns `True` if smap contains the solar center.
+
+    This is the case if and only if the solar center is inside the edges of the map. This
+    is checked by seeing if the sign of both the x and y coordintaes of the corners are opposite
+    (ie. the (0, 0) point is contained within the map).
```
ayshih, line 182: Maybe I'm missing something, but why wouldn't you do a `.world_to_pixel()` for Helioprojective (0, 0) and see if it lands in your pixel array?  Does that approach perform worse?

ayshih (approved):
Just one question about the choice of implementation for `contains_solar_center()`, but that doesn't need to block this PR.

sunpy/map/maputils.py
dstansby: Hah, no reason that I can think of, maybe I was being a bit *too* clever here 😆. I'll try `.world_to_pixel()` on (0, 0) and compare the performance.

PR #4464, opened 2020-09-11
str function for mapbase

### LINKED ISSUE #4201: Improve Map __str__
`Map` has an nice `__repr__`, but currently `print(map)` just prints the data array. It would be nice if `print(map)` could print some extra info like in `__repr__` (maybe even the same info?).

sidhu1012:
### Description

Fixes #4201

### DIFF
--- a/changelog/4464.trivial.rst
+++ b/changelog/4464.trivial.rst
@@ -1,0 +1,2 @@
+Changes __repr__ function in sunpy.map.mapbase .
+Added __str__ function in sunpy.map.mapbase .
--- a/changelog/4464.trivial.rst
+++ b/changelog/4464.trivial.rst
@@ -1,2 +1,1 @@
-Changes __repr__ function in sunpy.map.mapbase .
-Added __str__ function in sunpy.map.mapbase .
+Improved the output when you print a sunpy Map.
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -254,8 +254,11 @@ class GenericMap(NDData):
                                                     self._reference_latitude)),
                                tmf=TIME_FORMAT)
 
+    def __str__(self):
+        return object.__str__(self) + "\n" + self._text_summary() + "\n" + self.data.__str__()
+
     def __repr__(self):
-        return object.__repr__(self) + "\n" + self._text_summary() + "\n" + self.data.__repr__()
+        return f"{object.__repr__(self)}\n{self}"
 
     def _repr_html_(self):
         """
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -255,7 +255,7 @@ class GenericMap(NDData):
                                tmf=TIME_FORMAT)
 
     def __str__(self):
-        return object.__str__(self) + "\n" + self._text_summary() + "\n" + self.data.__str__()
+        return f"{self._text_summary()}\n{self.data.__repr__()}"
 
     def __repr__(self):
         return f"{object.__repr__(self)}\n{self}"
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -1104,5 +1104,7 @@ def test_contour(simple_map):
 
 
 def print_map(generic_map):
-    out = generic_map.__repr__()
-    assert isinstance(out, str)
+    out_repr = generic_map.__repr__()
+    assert isinstance(out_repr, str)
+    out_str = generic_map.__str__()
+    assert isinstance(out_str, str)
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -1103,7 +1103,7 @@ def test_contour(simple_map):
     assert u.allclose(contour.Ty, [0.5, 0, -0.5, 0, 0.5] * u.arcsec, atol=1e-10 * u.arcsec)
 
 
-def print_map(generic_map):
+def test_print_map(generic_map):
     out_repr = generic_map.__repr__()
     assert isinstance(out_repr, str)
     out_str = generic_map.__str__()
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -890,7 +890,7 @@ def test_bad_coordframe_repr(generic_map):
     generic_map.meta['CTYPE2'] = "STUART2"
     with pytest.warns(UserWarning,
                       match="Could not determine coordinate frame from map metadata"):
-        assert 'Unknown' in generic_map.__repr__()
+        assert 'Unknown' in generic_map.__str__()
 
 
 def test_bad_header_final_fallback():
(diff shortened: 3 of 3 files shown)


sidhu1012:
Why these checks failing on such simple change

nabobalis:
Some of those checks use servers which are prone to failing. 

What does the output now look like?

sidhu1012:
> A unit test would be ideal as well.

 Okay


> Can you also change the repr as suggested in the original issue?

Sure

nabobalis:
How do the new print and repr look on output?

sunpy/map/tests/test_mapbase.py
```diff
@@ -1101,3 +1101,8 @@ def test_contour(simple_map):
     assert contour.obstime == simple_map.date
     assert u.allclose(contour.Tx, [0, -1, 0, 1, 0] * u.arcsec, atol=1e-10 * u.arcsec)
     assert u.allclose(contour.Ty, [0.5, 0, -0.5, 0, 0.5] * u.arcsec, atol=1e-10 * u.arcsec)
+
+
+def print_map(generic_map):
```
wtbarnes, line 1106:
```suggestion
def test_print_map(generic_map):
```

sunpy/map/tests/test_mapbase.py
```diff
@@ -890,7 +890,7 @@ def test_bad_coordframe_repr(generic_map):
     generic_map.meta['CTYPE2'] = "STUART2"
     with pytest.warns(UserWarning,
                       match="Could not determine coordinate frame from map metadata"):
-        assert 'Unknown' in generic_map.__repr__()
+        assert 'Unknown' in generic_map.__str__()
```
wtbarnes, line 893: I don't think we should remove the test for `__repr__`. There should just be an additional test for `__str__`

sunpy/map/mapbase.py
```diff
@@ -254,8 +254,11 @@ def _text_summary(self):
                                                     self._reference_latitude)),
                                tmf=TIME_FORMAT)
 
+    def __str__(self):
+        return object.__str__(self) + "\n" + self._text_summary() + "\n" + self.data.__str__()
```
wtbarnes, line 258: Why is `__str__` now the old `__repr__`? Why can't it just be the same as `__repr__`? I'm not necessarily against this, I'm just not sure I fully understand the logic here.

ayshih: `__str__()`, unlike `__repr__()`, shouldn't return the object-level representation.

```suggestion
        return f"{self._text_summary()}\n{self.data.__repr__()}"
```
The first part (`object.__str__(self)`) shouldn't be there.  And it should stay `self.data.__repr__()` instead of `self.data.__str__()` because (I feel) we want to preserve the current output when it gets wrapped for `__repr__()`.

sidhu1012: repr should not change then?

ayshih: I don't understand your question.  To be clear, here's what I'm suggesting:
```python
    def __str__(self):
        return f"{self._text_summary()}\n{self.data.__repr__()}"

    def __repr__(self):
        return f"{object.__repr__(self)}\n{self}"
```

sidhu1012:
Does it seems fine now?

nabobalis:
How do the new print and repr look on output?

sidhu1012:
> How do the new print and repr look on output?

New print : 

image


New repr : 

image

sunpy/map/tests/test_mapbase.py
```diff
@@ -1101,3 +1101,8 @@ def test_contour(simple_map):
     assert contour.obstime == simple_map.date
     assert u.allclose(contour.Tx, [0, -1, 0, 1, 0] * u.arcsec, atol=1e-10 * u.arcsec)
     assert u.allclose(contour.Ty, [0.5, 0, -0.5, 0, 0.5] * u.arcsec, atol=1e-10 * u.arcsec)
+
+
+def print_map(generic_map):
+    out = generic_map.__repr__()
+    assert isinstance(out, str)
```
nabobalis, line 1108: This test needs expanding to test the print.

nabobalis:
A changelog is required for this change.

sunpy/map/tests/test_mapbase.py
sidhu1012: ok

sidhu1012:
> A changelog is required for this change.

okay

sunpy/map/tests/test_mapbase.py
sidhu1012: ok

sidhu1012:
> A changelog is required for this change. 

How to add changelog?

nabobalis:
> > A changelog is required for this change.
> 
> How to add changelog?

Instructions are here: https://github.com/sunpy/sunpy/blob/master/changelog/README.rst

sunpy/map/tests/test_mapbase.py
```diff
@@ -1101,3 +1101,10 @@ def test_contour(simple_map):
     assert contour.obstime == simple_map.date
     assert u.allclose(contour.Tx, [0, -1, 0, 1, 0] * u.arcsec, atol=1e-10 * u.arcsec)
     assert u.allclose(contour.Ty, [0.5, 0, -0.5, 0, 0.5] * u.arcsec, atol=1e-10 * u.arcsec)
+
+
+def test_print_map(generic_map):
+    out_repr = generic_map.__repr__()
+    assert isinstance(out_repr, str)
```
Cadair, line 1108: A simple addition here to check the repr has the base repr in it.
```suggestion
    assert isinstance(out_repr, str)
    assert object.__repr__(generic_map) in out_repr
```

Cadair (approved):
Thanks for working on this @sidhu1012 it looks good :smile:

sidhu1012:
> Thanks for working on this @sidhu1012 it looks good :smile: 

Thanks 😀

sunpy/map/tests/test_mapbase.py
```diff
@@ -1101,3 +1101,11 @@ def test_contour(simple_map):
...
+    assert object.__repr__(generic_map) in out_repr
+    out_str = generic_map.__str__()
+    assert isinstance(out_str, str)
```
ayshih, line 1111:
```suggestion
    assert isinstance(out_str, str)
    assert out_str in out_repr
```
Might as well test a bit more

ayshih (changes requested):
Still needs a changelog entry (maybe categorized under `trivial`?)

sidhu1012:
> Still needs a changelog entry (maybe categorized under `trivial`?)

Will add it

changelog/4464.trivial.rst
```diff
@@ -0,0 +1,2 @@
+Changes __repr__ function in sunpy.map.mapbase .
+Added __str__ function in sunpy.map.mapbase .
```
nabobalis, line 2:
```suggestion
Improved the output when you print a sunpy Map.
```

nabobalis (approved):

nabobalis:
RTD failure due to issue fixed in master.

Cadair (approved):

Cadair (changes requested):
`WARNING: :87: (ERROR/3) Unknown interpreted text role "method".`

not sure what this is about

Cadair (approved):
fixed on master apparently

Cadair:
Thanks a lot @sidhu1012 :rocket:

dstansby:
This changes the functionality of `__str__`, which has caused some breakages downstream in `pfsspy`, meaning I'm having to pin `sunpy!=2.0.2`. In general I think things like this that change behaviour and are not bug-fixes shouldn't be backported, and in general we should be a bit more careful about what we're backporting.

PR #4467, opened 2020-09-12
Cache map.wcs

dstansby:
Although creating the `.wcs` property is relatively in-expensive once, creating it many times can turn into a bottleneck for some code. As an example, `contains_full_disk` (after https://github.com/sunpy/sunpy/pull/4463) spends ~70% of its time constructing the `.wcs` property from the metadata 4 times. If we cache `.wcs`, the runtime improves from 0.3s to 0.1s.

Another advantage of caching the `.wcs` property is that warnings raised in its creation are only emitted once - currently they can be emitted a lot of times in a row when doing some tasks such as plotting.

This PR manually caches the `.wcs` property. In order to protect against changes in the metadata, `.wcs` is regenerated if any changes are made to `.meta`. Feedback very welcome on whether this is a sensible/silly idea!

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -36,7 +36,7 @@ from sunpy.image.resample import reshape_image_to_4d_superpixel
 from sunpy.sun import constants
 from sunpy.time import is_time, parse_time
 from sunpy.util import MetaDict, expand_list
-from sunpy.util.decorators import deprecate_positional_args_since, deprecated
+from sunpy.util.decorators import cached_property_based_on, deprecate_positional_args_since, deprecated
 from sunpy.util.exceptions import SunpyMetadataWarning, SunpyUserWarning
 from sunpy.util.functools import seconddispatch
 from sunpy.visualization import axis_labels_from_ctype, peek_show, wcsaxes_compat
@@ -421,7 +421,11 @@ class GenericMap(NDData):
         r = frame.represent_as(UnitSphericalRepresentation)
         return r.lon.to(self.spatial_units[0]), r.lat.to(self.spatial_units[1])
 
+    def _meta_hash(self):
+        return self.meta.item_hash()
+
     @property
+    @cached_property_based_on(_meta_hash)
     def wcs(self):
         """
         The `~astropy.wcs.WCS` property of the map.
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -425,7 +425,7 @@ class GenericMap(NDData):
         return self.meta.item_hash()
 
     @property
-    @cached_property_based_on(_meta_hash)
+    @cached_property_based_on('_meta_hash')
     def wcs(self):
         """
         The `~astropy.wcs.WCS` property of the map.
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -421,6 +421,7 @@ class GenericMap(NDData):
         r = frame.represent_as(UnitSphericalRepresentation)
         return r.lon.to(self.spatial_units[0]), r.lat.to(self.spatial_units[1])
 
+    @property
     def _meta_hash(self):
         return self.meta.item_hash()
 
--- a/sunpy/util/decorators.py
+++ b/sunpy/util/decorators.py
@@ -356,6 +356,7 @@ def cached_property_based_on(meth):
         """
         prop: the property method being decorated
         """
+        @wraps(outer)
         def inner(instance):
             """
             instance: the class instance of the property being decorated
--- a/sunpy/util/decorators.py
+++ b/sunpy/util/decorators.py
@@ -325,3 +325,52 @@ def deprecate_positional_args_since(since, keyword_only=False):
             return f(**kwargs)
         return inner_f
     return deprecate_positional_args
+
+
+_NOT_FOUND = object()
+
+
+def cached_property_based_on(meth):
+    """
+    A decorator to cache the value of a property based on the output of a
+    different callable.
+
+    This decorator caches the value of ``meth(instance)`` and ``prop(instance)``.
+    When the decorated property is accessed, ``meth(instance)`` is recomputed.
+    If this returns the same as its cached value, the cached value of ``prop``
+    is returned. Otherwise both ``meth`` and ``prop`` are recomputed, cached,
+    and the new value of ``prop`` is returned.
+
+    Parameters
+    ----------
+    meth : callable
+        A function that takes exactly one argument (a class instance) and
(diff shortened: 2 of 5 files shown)


nabobalis:
Makes sense to me that we could cache it instead of creating it each time.

sunpy/map/mapbase.py
```diff
@@ -200,6 +200,7 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
         self._default_carrington_longitude = None
         self._default_heliographic_latitude = None
         self._default_heliographic_longitude = None
+        self._wcs = None
```
Cadair, line 203: It's a good practice to define all class attributes in the init and not in methods.
```suggestion
        self._meta_hash = None
```

sunpy/util/metadata.py
```diff
@@ -102,6 +101,17 @@ def __delitem__(self, key):
         OrderedDict.__delitem__(self, key.lower())
         self._prune_keycomments()
 
+    def item_hash(self):
+        """
+        Create a hash based on the stored items.
+
+        This relies on all the items themselves being hashable. For this reason
```
Cadair, line 108: Probably worth failing back to the old behaviour somehow if someone has put something unhashable in the meta dict manually.

dstansby:
Hmm, I am running into `AttributeError: 'dict' object has no attribute 'item_hash'` on one of the tests. The issue is that `map.meta` isn't guaranteed to be a `MetaDict`, but can be a plain `dict`. Not sure whether the solution is to handle this case (I'm struggling to work that out...) or force `m.meta` to be a `MetaDict` during `m.__init__`?

nabobalis:
I would force it be a metadict.

wtbarnes:
I'm not totally against this and I do think we should increasingly be taking performance into account and this obviously is a pretty easy way to speed up things like the `contains_full_disk` function.

However, I'm a bit worried about this becoming a slippery slope. We could likely do this for quite a few properties that we calculate on the fly and probably see a performance bump in a decent number of places. But we do add caching at the cost of code complexity and readability and I'm just a bit worried about doing this every time we find a place where our on the fly approach is costing us in terms of compute time.

Again, I'm not really against this PR and my concerns should probably be documented in a separate PR/issue, but I think adding complexity for the sake of performance is something we should always consider very carefully and perhaps be reluctant to do most of the time.

dstansby:
Yes, that's a very good point. In this case, I think the much nicer solution would be to create a decorator along the lines of
```python
@cached_based_on(self.meta.meta_hash())
def wcs(self):
    ...
```
which acts in a similar way to the standard library [@cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property), but caches based on changes in the argument passed to the decorator.

sunpy/util/decorators.py
```diff
@@ -325,3 +325,52 @@ def inner_f(*args, **kwargs):
...
+    meth : callable
+        A function that takes exactly one argument (a class instance) and
+        returns an arbitrary output. Changes in the output are used to decide
+        whether or not to recompute the property.
+
+    Notes
+    -----
+    The cached value of ``meth(instance)`` is stored under the key ``meth.__name__``.
+    """
+    def outer(prop):
```
dstansby, line 355: Just remembered I forgot to copy the docstring across, leaving as a not for myself.

sunpy/util/metadata.py
```diff
@@ -34,11 +33,14 @@ def __init__(self, *args):
             args = list(args)
             adict = args[0]
             if isinstance(adict, list) or isinstance(adict, tuple):
-                tags = OrderedDict((k.lower(), v) for k, v in adict)
+                items = adict
```
nabobalis, line 36: This isn't related to the this PR but if its a tuple or a list, how does it have a "k,v" for the code below?

nabobalis:
I like the new approach, I do think it should be documented in a dev doc somewhere (not got any bright ideas on where tho).

wtbarnes:
I like this new strategy too! I think this pretty much alleviates all of my concerns in my previous comment.

sunpy/util/decorators.py
Cadair: doesn't `wraps` do that?

dstansby: Yep, this was a comment from before I added wraps

sunpy/util/metadata.py
dstansby: I think because it's meant to be a list of lists, where each sub-list is a 2-length key-item pair, e.g. `[['key1', 'item1'], ['key2', 'item2']]`

sunpy/util/decorators.py
```diff
@@ -325,3 +325,53 @@ def inner_f(*args, **kwargs):
...
     return deprecate_positional_args
+
+
+_NOT_FOUND = object()
+
+
+def cached_property_based_on(meth):
+    """
+    A decorator to cache the value of a property based on the output of a
+    different callable.
```
Cadair, line 336: Wouldn't making the input here a attribute / property and not a callable be a little clearer? Is the reason it's a method to deal with scoping issues when defining the class?

An alternative could be to do it by name rather than passing in the actual method
```
@cached_property_based_on("_meta_hash")
```
which then does a `getattr` on the object?

dstansby: The reason I did this is because handing over an object (unless it's handing it by reference and not value?) doesn't work, as there needs to be some way to detect changes. Didn't think of doing `getattr` though, and agree that is a bit of a nicer solution, so will update it.

Cadair: (I would expect a property to work as it's really a callable underneath, so it's effectively by reference. A generic attribute may or may not work depending I guess :+1: )

dstansby: I think I've updated this to what you meant now?

dstansby:
As a performance note, locally running the tests in `sunpy/map/tests/test_mapbase.py` takes 17s before this PR and 13s with this PR for me, will be interesting to compare CI run times before/after.

sunpy/map/mapbase.py
```diff
@@ -418,7 +418,11 @@ def _get_lon_lat(self, frame):
         r = frame.represent_as(UnitSphericalRepresentation)
         return r.lon.to(self.spatial_units[0]), r.lat.to(self.spatial_units[1])
 
+    def _meta_hash(self):
```
Cadair, line 421: I was anticipating this:
```suggestion
    @property
    def _meta_hash(self):
```

Cadair:
I think the implementation here is good, all the usual tests etc :grinning:

Cadair (approved):

nabobalis (approved):

PR #4475, opened 2020-09-14
Raise nice error if MetaDict gets non-str keys

dstansby:
Pulled out of https://github.com/sunpy/sunpy/pull/4467, this provides a nicer error message if the user tries to provide MetaDict with non-string keys. This is good anyway, and a non-breaking pre-requisite to forcing `GenericMap.meta` to always be a metadict.


nabobalis:
Just a unit test if possible.

dstansby:
Might just leave this lumped into https://github.com/sunpy/sunpy/pull/4476 on second thoughts

PR #4476, opened 2020-09-14
Force map.meta to be a MetaDict

dstansby:
This is a requirement for https://github.com/sunpy/sunpy/pull/4467.

Previously it was possible that `map.meta` could just be a simple dict if a map was constructed with `map.GenericMap`. This brings `GenericMap` in line with the map factory (ie. `Map`) and forces `meta` to always be a `MetaDict`.


sunpy/map/tests/test_mapbase.py
```diff
@@ -901,13 +901,22 @@ def test_bad_coordframe_repr(generic_map):
...
+
+
 def test_bad_header_final_fallback():
     # Checks that if a WCS cannot be constructed from the
     # header, a warning is raised and a simple WCS is created
     # instead
     header = {'cunit1': 'arcsec',
               'cunit2': 'arcsec',
-              None: None,  # Cannot parse this into WCS
+              'None': None,  # Cannot parse this into WCS
```
dstansby, line 919: I'm not sure how to construct a valid `MetaDict` that isn't a valid wcs to trigger the warning below now 😞

nabobalis: Maybe it isn't possible.

dstansby: 🤣 of course it was me that wrote this test originally...

sunpy/map/tests/test_mapbase.py
```diff
@@ -901,22 +901,13 @@ def test_bad_coordframe_repr(generic_map):
         assert 'Unknown' in generic_map.__repr__()
 
 
-def test_bad_header_final_fallback():
-    # Checks that if a WCS cannot be constructed from the
-    # header, a warning is raised and a simple WCS is created
-    # instead
```
dstansby, line 907: As far as I can tell, as long as everything in the header is a string (which is now mandated), astropy will happily parse it and we will never get to the point where a simple WCS needs to be created. Warnings like
```
VerifyWarning("Keyword name '#none' is greater than 8 characters or contains characters not allowed by the FITS standard; a HIERARCH card will be created.")
``` 
are raised when I try and put some gibberish in a header. So I've removed this test. I suppose there's no harm in have the fallback, but I can't think of anything that will actually exercise it.

nabobalis (approved):

sunpy/map/tests/test_mapbase.py
Cadair: Interesting, and it's good that this improves the validation and makes the error better.

I think there is a class of errors for which the fallback in GenericMap is still useful and might not be being tested, which is a valid FITS header but an invalid WCS. i.e. something which when passed to `astropy.wcs.WCS` errors because of some issue (like invalid values etc).

dstansby: I'm still struggling to find something that trips the warning. Changing ctype, crval, crpix, or cdelt to garbage values just makes the checks below error instead. I've also tried with the header (I think a WISPR one?) that caused this warning to be put in the code, and that doesn't trigger the warning either.

nabobalis:
I feel like this is an API change, backporting this might be off the cards.

Cadair (approved):

PR #4477, opened 2020-09-15
Normalize images before transformation when using skimage

### LINKED ISSUE #4453: Normalise images before handing them to skimage.transform.warp
From @ayshih at https://github.com/sunpy/sunpy/pull/4452#issuecomment-686504683:

> #2041 removed the normalization code and cast everything to float64. However, I believe the removal of the normalization code should not have been done. Scikit-image's documentation is quite clear that it always assumes that floats are in the range of [0, 1]. While warp() does appear to work fine when that assumption is violated, we are currently relying on undefined behavior.

> I suggest that the normalization code be re-inserted.

I believe the work here is to normalise floating point input to `skimage.transform.warp` to the range [0, 1] in `sunpy.image.transform.affine_transform`, and then un-normalise it after the transform.

chouhanaryan:
### Description
- Normalize the image to [0, 1] when using `skimage.transform.warp`, and later rescale it back to the original range
- In the case of an image consisting entirely of NaNs, pass it down as it is (this was done to conform with the tests - specifically, `sunpy.image.tests.test_transform.test_nan_skimage_low`)

Fixes #4453


chouhanaryan:
There was some unexpected behaviour that I encountered when running `test_nan_skimage_low`. The problem was that `np.nanmin` and `np.nanmax` (which are called during normalization) throw a Runtime warning when the array passed only has NaNs (which this particular test case checks for). I chose to handle it as follows - if the array consists entirely of NaNs, I skip the normalization entirely and pass it down as it is to `skimage.transform.warp`. Later, the image is not rescaled back either. Even though this works, let me know if this should be handled any differently.

Note - I also chose to keep the `preserve_range` keyword when calling `skimage.transform.warp` because it prevents one extra call to `img_as_float` further down the line. We already have a float image so there is no need to process it again.

sunpy/image/transform.py
```diff
@@ -126,8 +130,27 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
...
 
+        # Scale image to range [0, 1] if it is valid (not made up entirely of NaNs)
+        is_nan_image = np.all(np.isnan(adjusted_image))
+        if is_nan_image:
+            adjusted_missing = missing
+        else:
+            im_min = np.nanmin(adjusted_image)
+            adjusted_image -= im_min
+            im_max = np.nanmax(adjusted_image)
+            if im_max > 0:
```
dstansby, line 141: Why this if clause? If the minimum value in the image is now zero (by definition because you have subtracted it), surely the maximum value will always be greater than zero?

dstansby (commented):
Looks good overall - I have one question. There's also a couple of formatting issues - see https://docs.sunpy.org/en/latest/dev_guide/code_standards.html#formatting for how to run our checks and fix them.

sunpy/image/transform.py
chouhanaryan: From what I can gather, that code was introduced [here](https://github.com/sunpy/sunpy/commit/982bb0dadd736a6c7d91defb596f4dc4cdcd1679). If we have a flat array with the same values, after its minimum is subtracted, we're left with an array of zeros. It's maximum is evaluted to zero and a 0/0 division occurs in the next line.
This code handles that 😃

And yes, I'll fix the formatting issues, thanks! 👍

dstansby (approved):

changelog/4477.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed :meth:`~sunpy.image.transform.affine_transform` to scale images to [0, 1] before
+passing them to :meth:`skimage.transform.warp` and later rescale them back.
```
nabobalis, line 2:
```suggestion
passing them to :func:`skimage.transform.warp` and later rescale them back.
```
I think wrap is a function.

chouhanaryan: You're right, thanks! 😅

sunpy/image/transform.py
```diff
@@ -116,7 +120,7 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
 
         if issubclass(image.dtype.type, numbers.Integral):
             warnings.warn("Integer input data has been cast to float64, "
-                          "which is required for the skikit-image transform.",
+                          "which is required for the scikit-image transform.",
```
ayshih, line 123: Can you remove the "which is required..." from this warning?  It's not actually a requirement.

sunpy/image/transform.py
```diff
@@ -126,8 +130,27 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
...
+            if im_max > 0:
+                adjusted_image /= im_max
+                adjusted_missing = (missing - im_min) / im_max
+            else:
+                adjusted_missing = missing - im_min
+
         rotated_image = skimage.transform.warp(adjusted_image, tform, order=order,
-                                               mode='constant', cval=missing,
-                                               preserve_range=True)
+                                               mode='constant', cval=adjusted_missing, preserve_range=True)
```
ayshih, line 148:
```suggestion
                                               mode='constant', cval=adjusted_missing)
```
Given that we're doing the keyword `preserve_range` applies only to passing in an integer array, and our code never does that, it's probably best to remove this non-functional keyword argument.

sunpy/image/transform.py
```diff
@@ -126,8 +130,27 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
...
+            adjusted_missing = missing
+        else:
+            im_min = np.nanmin(adjusted_image)
+            adjusted_image -= im_min
+            im_max = np.nanmax(adjusted_image)
+            if im_max > 0:
+                adjusted_image /= im_max
+                adjusted_missing = (missing - im_min) / im_max
+            else:
+                adjusted_missing = missing - im_min
```
ayshih, line 145:
```suggestion
                # The input array is all one value (aside from NaNs), so no scaling is needed
                adjusted_missing = missing - im_min
```

sunpy/image/transform.py
```diff
@@ -59,14 +59,18 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
...
 
-    When using for `skimage.transform.warp` with order >= 4 or using
-    `scipy.ndimage.affine_transform` at all, "NaN" values will
-    replaced with zero prior to rotation. No attempt is made to retain the NaN
-    values.
+    When using `skimage.transform.warp` with order >= 4 or using
+    `scipy.ndimage.affine_transform` at all, "NaN" values will be replaced with
+    zero prior to rotation. No attempt is made to retain the "NaN" values except when
+    the entire image consists of "NaN" values only and `skimage.transform.warp` is used,
+    where it is passed down as it is.
```
ayshih, line 66: This last part ("passed down as it is") isn't actually true.  As currently implemented, you still go through with the `warp()` call on the all-NaN array, which (I think) has the potential of returning an array that is not all NaNs for parts outside the image boundaries.

ayshih (changes requested):

sunpy/image/transform.py
chouhanaryan: Yes, you're right, but that explanation doesn't refer to the values returned - only the values passed to `warp`, which are NaNs only (for an all-NaN array). It doesn't talk about the return values at all 😅

I've made a commit with all of the other changes, if you still want me to change this let me know and I'll fix it.

ayshih: Ah, I now understand how to read your sentence, but it's still inaccurate/misleading given the current implementation.  Your code's logic branch for an image of all NaNs comes *after* the code to set NaN values to zero if it is needed for a selected transform.  So:

- If the selected transform requires us to replace NaN values, an image full of NaNs will also be entirely replaced.  The NaN values will not be retained even for an image full of NaN values.
- If the selected transform does not require us to replace NaN values, and image full of NaNs is not treated any differently than an image that has only a few NaN values.  The NaN values are not replaced in either image.  Note that the overall paragraph isn't talking about this case anyway.

chouhanaryan: Ahh, got it. I'll remove the last part of the sentence, thanks.

changelog/4477.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed :meth:`~sunpy.image.transform.affine_transform` to scale images to [0, 1] before
```
ayshih, line 1:
```suggestion
Fixed :func:`~sunpy.image.transform.affine_transform` to scale images to [0, 1] before
```

ayshih (approved):
One small fix, and then looks good to me

nabobalis:
Failures are SUVI related.

nabobalis:
Thanks @chouhanaryan

PR #4466, opened 2020-09-11
Fix registration of wildcard attrs

Cadair:
And fix the docs

Issue #4194, opened 2020-05-21
Modifying our approach to Map metadata


ayshih:
The current storage of our Map metadata (`.meta`) is confusing because it is the combination of:

- keywords in the original header that we trust and use actively for calculations
- keywords in the original header that we know have wrong values (e.g., #3906, #3057, #3038)
- keywords in the original header that we have no opinion on
- keywords that we've *overwritten* from the original header
- keywords that we've *added* compared to the original header (e.g., `PCi_j`)
- keywords that we've *removed* compared to the original header (e.g., `CROTA2`)

While we our API does have properties as an interface to `.meta`, it's unavoidable for users to directly access `.meta`, and then it's opaque to the user whether they can trust what's in `.meta`.

I suggest a modified approach to Map metadata.  Rather than having a single dictionary `.meta`, we would have two dictionaries:

- `.original_header` would preserve the original values from the header, and we don't modify this header at all after it is created
- `.meta` would contain *only* those keywords that we "bless", whether they are copied from `.original_header` or are generated through our own code

For example, if there's a `PCi_j` keyword in `.original_header`, we'd simply copy that to `.meta`.  If there's only `CROTA2`, we'd calculate the appropriate `PCi_j` keywords and put them in `.meta`, but we wouldn't modify `.original_header`.  The `CROTA2` keyword would never show up in `.meta`.

For observer coordinates, we'd copy over to `.meta` only the set of coordinates that we trust from `.original_header`.  In the unfortunate cases where there are no correct values (e.g., #3057, #3037), we could choose to copy over *nothing* to `.meta` so that it's explicit that there are no "blessed" values.  Even if we automatically determine the observer location through external sources (e.g., JPL Horizons), we'd put those "blessed" values in `.meta` but still not touch `.original_header`.

It's an open question what we save when saving to a file.  I haven't thought much about the considerations there.

Thoughts?

dstansby:
I'm 👍 for saving the original 100% untouched header somewhere. Whilst writing map classes for synoptic maps it is often helpful to validity check metadata fixes against parts of the original header that may be subsequently modified by the map class, so keeping a copy of the original header would be very useful in at least that case.

ayshih:
I'll preserve here some Riot discussion between myself and @Cadair:

> > > @Cadair:
> > > but my thinking was we would have something quite a lot like that in some kinda metadata object
> >
> > @ayshih:
[...]

PR #4479, opened 2020-09-17
More BUNIT fixes for AIAMap

wtbarnes:
This is an add-on to #4451 to better handle units in AIA images.

Update `DN` to `ct` in the case of normalized and un-normalized counts. 

Also, add `BUNIT` in the case where an image contains the `PIXLUNIT` keyword which is not a FITS standard keyword. However, it seems that this is the keyword that AIA has chosen to denote units in their data products. See http://jsoc.stanford.edu/~jsoc/keywords/AIA/AIA02840_K_AIA-SDO_FITS_Keyword_Document.pdf

### DIFF
--- a/sunpy/map/sources/sdo.py
+++ b/sunpy/map/sources/sdo.py
@@ -48,11 +48,15 @@ class AIAMap(GenericMap):
 
         # Fill in some missing info
         self.meta['detector'] = self.meta.get('detector', "AIA")
+        if 'pixlunit' in self.meta:
+            self.meta['bunit'] = self.meta['pixlunit']  # PIXLUNIT is not a FITS standard keyword
         self._nickname = self.detector
         self.plot_settings['cmap'] = self._get_cmap_name()
         self.plot_settings['norm'] = ImageNormalize(
             stretch=source_stretch(self.meta, AsinhStretch(0.01)), clip=False)
-        # DN/s is not a FITS standard unit, so convert to counts/second
+        # DN is not a FITS standard unit, so convert to counts
+        if self.meta.get('bunit', None) == 'DN':
+            self.meta['bunit'] = 'ct'
         if self.meta.get('bunit', None) == 'DN/s':
             self.meta['bunit'] = 'ct/s'
 
--- a/sunpy/map/sources/sdo.py
+++ b/sunpy/map/sources/sdo.py
@@ -49,7 +49,8 @@ class AIAMap(GenericMap):
         # Fill in some missing info
         self.meta['detector'] = self.meta.get('detector', "AIA")
         if 'bunit' not in self.meta and 'pixlunit' in self.meta:
-            self.meta['bunit'] = self.meta['pixlunit']  # PIXLUNIT is not a FITS standard keyword
+            # PIXLUNIT is not a FITS standard keyword
+            self.meta['bunit'] = self.meta['pixlunit']
         self._nickname = self.detector
         self.plot_settings['cmap'] = self._get_cmap_name()
         self.plot_settings['norm'] = ImageNormalize(
--- a/sunpy/map/sources/sdo.py
+++ b/sunpy/map/sources/sdo.py
@@ -48,7 +48,7 @@ class AIAMap(GenericMap):
 
         # Fill in some missing info
         self.meta['detector'] = self.meta.get('detector', "AIA")
-        if 'pixlunit' in self.meta:
+        if 'bunit' not in self.meta and 'pixlunit' in self.meta:
             self.meta['bunit'] = self.meta['pixlunit']  # PIXLUNIT is not a FITS standard keyword
         self._nickname = self.detector
         self.plot_settings['cmap'] = self._get_cmap_name()
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -370,7 +370,10 @@ def test_save(aia171_test_map, generic_map):
     aiamap.save(afilename, filetype='fits', overwrite=True)
     loaded_save = sunpy.map.Map(afilename)
     assert isinstance(loaded_save, sunpy.map.sources.AIAMap)
-    assert loaded_save.meta == aiamap.meta
+    # Compare metadata without considering ordering of keys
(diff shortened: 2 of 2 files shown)


sunpy/map/sources/sdo.py
```diff
@@ -48,11 +48,15 @@ def __init__(self, data, header, **kwargs):
 
         # Fill in some missing info
         self.meta['detector'] = self.meta.get('detector', "AIA")
+        if 'pixlunit' in self.meta:
```
dstansby, line 51:
```suggestion
        if 'bunit' not in self.meta and 'pixlunit' in self.meta:
```
This guards against a user set 'bunit' being overwritten.

dstansby (changes requested):

wtbarnes:
Should we delete `PIXLUNIT` after updating `BUNIT` as it is a non-standard keyword and potentially conflicts with `BUNIT`?

wtbarnes:
These test failures are real and are the result of the metadata between the saved and in-memory map not being the same because of the addition of the `BUNIT` key.

wtbarnes:
Hmm I take back my previous comment. I can't seem to reproduce this locally.

wtbarnes:
This is failing simply because of the order in which these keys appear. Why are we testing the ordering of the keys?
```
E         -           ('bunit', 'ct'),
E                     ('comment',
E                      'FITS (Flexible Image Transport System) format is defined in '
E                      "'Astronomyand Astrophysics', volume 376, page 359; bibcode: "
E                      '2001A&A...376..359H'),
E                     ('history', ''),
E                     ('keycomments',
E                      {'BITPIX': 'array data type',
E                       'NAXIS': 'number of array dimensions',
E         -             'SIMPLE': 'conforms to FITS standard'})],
E         ?                                                    -
E         +             'SIMPLE': 'conforms to FITS standard'}),
E         +           ('bunit', 'ct')],
E           )

sunpy/map/tests/test_mapbase.py:373: AssertionError
```

nabobalis:
We probably shouldn't be.

wtbarnes:
Yeah I've adjusted the check to be independent of order. I'm not sure if we're doing this elsewhere.

dstansby:
> Should we delete `PIXLUNIT` after updating `BUNIT` as it is a non-standard keyword and potentially conflicts with `BUNIT`?
> 
> I suppose precedent would probably say no as we do not do this in any other `Map` source.

See https://github.com/sunpy/sunpy/issues/4194 for discussion around this. I think what we really should do is keep the original header, and then modify a copy (in which case deleting pixlunit would be fine on the copy). I think this PR is fine for now though.

dstansby (approved):
But I think it would be worth waiting for the doc build to be fixed elsewhere and passed here before merging.

nabobalis:
Why would the doc build be important here?

dstansby:
🤷‍♂️ not sure, but it has caught bugs that the test suite hasn't caught in the past.

wtbarnes:
I'm quite confused by those doc build failures but I haven't yet tried to reproduce them locally. I don't think they're related to my PR at all, but I've not seen that before and am concerned

nabobalis:
It is fixed by https://github.com/sunpy/sunpy/pull/4466

wtbarnes:
Ok I suspected that was the case. I can wait until that is merged and then rebase to do the docs build check. There's no urgency here.

nabobalis:
It has now been merged in, so the doc build should pass with a rebase.

dstansby (approved):
If/when tests pass

sunpy/map/tests/test_mapbase.py
```diff
@@ -370,7 +370,10 @@ def test_save(aia171_test_map, generic_map):
     aiamap.save(afilename, filetype='fits', overwrite=True)
     loaded_save = sunpy.map.Map(afilename)
     assert isinstance(loaded_save, sunpy.map.sources.AIAMap)
-    assert loaded_save.meta == aiamap.meta
+    # Compare metadata without considering ordering of keys
+    assert loaded_save.meta.keys() == aiamap.meta.keys()
+    for k in aiamap.meta:
+        assert loaded_save.meta[k] == aiamap.meta[k]
```
nabobalis, line 376: Can we not compare `.items()` directly instead of a loop?

nabobalis: Google seems to suggest not.

sunpy/map/sources/sdo.py
```diff
@@ -48,11 +48,15 @@ def __init__(self, data, header, **kwargs):
 
         # Fill in some missing info
         self.meta['detector'] = self.meta.get('detector', "AIA")
+        if 'bunit' not in self.meta and 'pixlunit' in self.meta:
+            self.meta['bunit'] = self.meta['pixlunit']  # PIXLUNIT is not a FITS standard keyword
```
nabobalis, line 52:
```suggestion
			# PIXLUNIT is not a FITS standard keyword
            self.meta['bunit'] = self.meta['pixlunit']
```

sunpy/map/sources/sdo.py
```diff
@@ -48,11 +48,15 @@ def __init__(self, data, header, **kwargs):
...
         self._nickname = self.detector
         self.plot_settings['cmap'] = self._get_cmap_name()
         self.plot_settings['norm'] = ImageNormalize(
             stretch=source_stretch(self.meta, AsinhStretch(0.01)), clip=False)
-        # DN/s is not a FITS standard unit, so convert to counts/second
+        # DN is not a FITS standard unit, so convert to counts
```
nabobalis, line 57:
```suggestion
        # DN and DN/s are not FITS standard units, so convert to counts
```

nabobalis (approved):

sunpy/map/sources/sdo.py
wtbarnes: I'd argue that it's really just `DN` that's non-standard and that any composite unit containing DN is then obviously non-standard

sunpy/map/tests/test_mapbase.py
wtbarnes: I'm not sure. I know this is not the nicest approach. I'm happy to do something different here if there is a better option.

PR #4486, opened 2020-09-20
Add aiapy docs to the intersphinx mapping

wtbarnes:
As the title says. This may be useful in the future if in the docs, we want to defer to aiapy for AIA-specific functionality.


nabobalis (approved):

dstansby:
Not sure what the point is in backporting?

Issue #4490, opened 2020-09-22
The 'Creating a Composite map' example in the gallery shows warnings


chouhanaryan:
### Description
The code outputs in the [Creating a Composite map](https://docs.sunpy.org/en/stable/generated/gallery/map/composite_map_AIA_HMI.html) example in the gallery might be bugged.

### Expected vs Actual behavior
The code in the example should show appropriate outputs, but it currently doesn't.

Edit: Well, the code actually runs correctly (just ran it locally), but a number of `SunpyUserWarnings` and `FITSFixedWarnings` pop up. Is this expected behaviour?

ayshih:
In short, the warnings are ignorable.

The `FITSFixedWarning` warnings are expected because the HMI header makes use of a non-standard `'nan'` value (see #4324).  If you want to stop the warnings, you can add the lines:
```python
hmi_map.meta['crder1'] = 0
hmi_map.meta['crder2'] = 0
```

The `SunpyUserWarning` warnings are arguably a bug.  We emit a warning when it looks like the user may be supplying a rectangle incorrectly, in this case for the `.submap()` call.  However, the check is performed in pixel space (see #4179, ping @dstansby), and doesn't currently take into account that map may be rotated.  HMI maps are ~180 degrees rotated (i.e., basically upside down), so the check incorrectly fails.  You can de-rotate the HMI map with the line:
```python
hmi_map = hmi_map.rotate()
```
I should point out that while this line will prevent the `SunpyUserWarning` associated with the `.submap()` call, the `.rotate()` call itself will emit a different `SunpyUserWarning` because HMI maps have NaN values that aren't allowed with the default rotation settings.

PR #4491, opened 2020-09-22
Remove submap warnings on pixel ordering

dstansby:
On second thoughts, prompted by https://github.com/sunpy/sunpy/issues/4490, I think that there shouldn't be a check on the pixel ordering. Even though our API names the two corners of the rectangle as `bottom_left` and `top_right`, it does not really matter as a rectangle is uniquely defined by two pairs of un-ordered coordinates.

This removes a bit of the sunpy warning soup, but happy to hear counter arguments as to why this warning should be kept.

Fixes https://github.com/sunpy/sunpy/issues/4490

### DIFF
--- a/changelog/4491.trivial.rst
+++ b/changelog/4491.trivial.rst
@@ -1,0 +1,4 @@
+If the ``top_right`` corner given to :meth:`sunpy.map.Map.submap` is
+below or to the right of the ``bottom_left`` corner, a warning is no longer
+raised (as the rectangle is still well defined), but a message is still logged
+at the debug level to the sunpy logger.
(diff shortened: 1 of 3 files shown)


ayshih:
Rather than removing the warnings entirely, how about emitting debug-level logger messages (i.e., use `log.debug()`)?  If any user reports that `submap()` is doing something weird, we can ask them to turn on debug-level messages for additional diagnostic output.  (Of course, that would also turn on all of the `coordinates` diagnostic output...)

Cadair:
Should we consider this a bugfix?

nabobalis (approved):

nabobalis:
CI failure due to upstream dep warning.

PR #4499, opened 2020-09-24
Extend TimeSeries concatenate to allow iterables

### LINKED ISSUE #4495: Typo / missing functionality in TimeSeries concatenate method
### Description
The current method signature is `def concatenate(self, otherts, same_source=False, **kwargs):` so there is a small typo 'otherts' -> 'others' but currently the method can only take a single timeseries so should probably be 'other'. Simplest fix would be to just change 'otherts' to 'other' but I think it might be more beneficial to change to 'others' and add the missing code to handle a list/iterable of timeseries and is reference in the code comments.

Works fine to concatenate a time series with a single other time series but doesn't work combing a list of time series.

Currently doing this type of thing quite, often where the TimeSeries are all of the same type

```python
my_timeseries = TimeSeries(listofiles)
combined_ts = my_timeseries[0]
for ts in my_timeseries[1:]:
     combined_ts = combined_ts.concatenate(bk)
```

### Steps to Reproduce
```python

```

### System Details
- SunPy Version: 2.0.1
- Astropy Version: 3.2.3
- Python Version: 3.7.0
- OS information: macOS 10.15.6

chouhanaryan:
### Description

The `concatenate` method of `TimeSeries` currently only allows a single time series object to be passed. This PR will extend that method to allow for an iterable containing multiple objects to be passed as well.

This will involve two things - first, extending TimeSeriesMetaData's concatenate method to allow for iterables. Second, using that to extend TimeSeries's concatenate method.

- [x] TimeSeriesMetaData concatenate
- [x] TimeSeriesMetaData tests
- [x] TimeSeries concatenate
- [x] TimeSeries tests
- [x] 100% codecov
- [x] Documentation

Fixes #4495 

Note for myself: I first tried overloading the TSMD concatenate method using `@seconddispatch` but while registering the decorator with the `TimeSeriesMetaData` type, it simply wouldn't work. This could be because I'm basically trying to register the data type as the class of which the method actually belongs to. So I defaulted to simple if-else checks. Will have to dive deeper to understand if it's still possible.


chouhanaryan:
Not sure if appropriate or not - but pinging @nabobalis 😅 PR hasn't been reviewed yet

sunpy/timeseries/metadata.py
```diff
@@ -338,22 +339,36 @@ def get(self, keys, time=None, colname=None):
...
         ----------
-        tsmetadata2 : `~sunpy.timeseries.TimeSeriesMetaData`
-            The second TimeSeriesMetaData object.
+        tsmetadata2 : `~sunpy.timeseries.TimeSeriesMetaData` or `collections.abc.Iterable`
+            The second `~sunpy.timeseries.metadata.TimeSeriesMetaData` object or an iterable
+            containing multiple `~sunpy.timeseries.metadata.TimeSeriesMetaData` objects.
         """
+        # Check if the type of the object passed is valid
+        if not isinstance(tsmetadata2, (self.__class__, Iterable)):
+            raise TypeError(f"Invalid type provided: {type(tsmetadata2)}")
```
nabobalis, line 355: It would be useful to tell a user what is valid.

sunpy/timeseries/metadata.py
```diff
@@ -338,22 +339,36 @@ def get(self, keys, time=None, colname=None):
...
+
+        # If an individual TimeSeriesMetaData object is to be concatenated, wrap it in a list
+        # Else, it must already be an iterable, so check if all items within it are valid
+        if isinstance(tsmetadata2, self.__class__):
+            tsmetadata2 = [tsmetadata2]
+        elif not all(isinstance(series, self.__class__) for series in tsmetadata2):
```
nabobalis, line 361: The same here.

sunpy/timeseries/metadata.py
```diff
@@ -338,22 +339,36 @@ def get(self, keys, time=None, colname=None):
...
+            raise TypeError("Invalid type within iterable")
```
nabobalis, line 362: I feel like these if/elses could be condensed.

sunpy/timeseries/tests/test_timeseriesmetadata.py
```diff
@@ -158,12 +158,30 @@ def test_append_invalid_timerange(basic_1_md):
...
 
+def test_concatenate_list(basic_ascending_append_md, basic_3_md, basic_4_md, complex_append_md):
+    concatenated = copy.deepcopy(basic_ascending_append_md)
+    concatenated = concatenated.concatenate([basic_3_md, basic_4_md])
+    assert concatenated == complex_append_md
+
+
+def test_concatenate_invalid_type(basic_ascending_append_md):
+    concatenated = copy.deepcopy(basic_ascending_append_md)
+    with pytest.raises(TypeError):
```
nabobalis, line 175: It is worth to match the message as well just to make sure you are raising the correct error.

sunpy/timeseries/tests/test_timeseriesmetadata.py
```diff
@@ -158,12 +158,30 @@ def test_append_invalid_timerange(basic_1_md):
...
+        concatenated = concatenated.concatenate(100)
+
+
+def test_concatenate_invalid_type_within_list(basic_ascending_append_md):
+    concatenated = copy.deepcopy(basic_ascending_append_md)
+    with pytest.raises(TypeError):
```
nabobalis, line 181: It is worth to match the message as well just to make sure you are raising the correct error.

sunpy/timeseries/tests/test_timeseriesbase.py
```diff
@@ -481,6 +522,20 @@ def test_concatenation_of_slices(eve_test_ts, concatenated_slices_test_ts):
...
+    assert_frame_equal(concatenated_slices_test_list.to_dataframe(), eve_test_ts.to_dataframe())
+    # Otherwise: concatenated_ts.data.equals(eve_test_ts)
+    # Compare timeranges from before and after match for both metadata and TS
+    assert eve_test_ts.meta.time_range == concatenated_slices_test_list.meta.time_range
+    assert eve_test_ts.time_range == concatenated_slices_test_list.time_range
+    # Test metadata MetaDict matches
+    eve_test_ts.meta.metadata[0][
+        2] == concatenated_slices_test_list.meta.metadata[0][
+            2] == concatenated_slices_test_list.meta.metadata[1][2]
+    # ToDo: Will TSMD.concatenate() want to re-merge the metadata entries back into one?
```
nabobalis, line 536: What do you mean by this?

nabobalis:
The style checking job failed, seems something is wrong with the changelog entry.

sunpy/timeseries/tests/test_timeseriesbase.py
chouhanaryan: So I copied most of the code for that test from what was already present and that comment was part of what I copied over.

I dug around a bit and I think what that's asking is if we want `concatenate()` to combine the metadata of the concatenated TS objects.

Right now, say I have -

```python
>>> ts_1.meta.metadata = [(
        <sunpy.time.timerange.TimeRange object at 0x208ac7304c8>
        Start: 2016-09-30 05:01:00
        End:   2016-10-01 05:00:00
        Center:2016-09-30 17:00:30
        Duration:0.9993055555555556 days or
               23.983333333333334 hours or
               1439.0 minutes or
               86340.0 seconds,
        ["intensity"],
        MetaDict([("key", "value")])
    )]

>>> ts_2.meta.metadata = [(
        <sunpy.time.timerange.TimeRange object at 0x208a7683e48>
        Start: 2016-10-01 05:01:00
        End:   2016-10-02 05:00:00
        Center:2016-10-01 17:00:30
        Duration:0.9993055555555556 days or
               23.983333333333334 hours or
               1439.0 minutes or
               86340.0 seconds,
        ["intensity"],
        MetaDict([("key", "value")])
    )]
```

then `ts_1.concatenate(ts_2).meta.metadata` will simply give a list with both ts_1 and ts_2's metadata (`[*ts_1.meta.metadata, *ts_2.meta.metadata]`). This behavior carries over to when ts_2 is an iterable.

So instead of this, do we want to combine all of these metadata entries back into one or not.

sunpy/timeseries/timeseriesbase.py
```diff
@@ -356,15 +357,16 @@ def extract(self, column_name):
 
     def concatenate(self, otherts, same_source=False, **kwargs):
```
hayesla, line 358: I'm a bit confused about the word `otherts`? Shouldn't this be `others?`

sunpy/timeseries/timeseriesbase.py
```diff
@@ -356,15 +357,16 @@ def extract(self, column_name):
...
+        Concatenate with another `~sunpy.timeseries.TimeSeries` or an iterable containing multiple
+        `~sunpy.timeseries.TimeSeries`. This function will check and remove any duplicate times.
+        It will keep the column values from the original timeseries to which the new time
+        series is being added.
 
         Parameters
         ----------
-        otherts : `~sunpy.timeseries.TimeSeries`
-            Another `~sunpy.timeseries.TimeSeries`.
+        otherts : `~sunpy.timeseries.TimeSeries` or `collections.abc.Iterable`
```
hayesla, line 367:
```suggestion
        others : `~sunpy.timeseries.TimeSeries` or `collections.abc.Iterable`
```

sunpy/timeseries/timeseriesbase.py
```diff
@@ -356,15 +357,16 @@ def extract(self, column_name):
...
+            Another `~sunpy.timeseries.TimeSeries` or an iterable containing multiple
+            `~sunpy.timeseries.TimeSeries`.
         same_source : `bool`, optional
             Set to `True` to check if the sources of the time series match. Defaults to `False`.
```
hayesla, line 372: It would also be nice to put an example here

sunpy/timeseries/tests/test_timeseriesbase.py
nabobalis: You mean to combine them into one metadata object?

sunpy/timeseries/timeseriesbase.py
nabobalis: I thought that "otherts" stands for "other timeseries"

sunpy/timeseries/tests/test_timeseriesbase.py
chouhanaryan: Yeah. It might make sense for when the resulting series is continuous. So we set Start and End to the minimum and maximum of all Start and all End times and sum up the Duration too. If the source is the same, the columns and relevant MetaDict are probably going to be the same too, so they can be combined as well. This is for a specific sub-case, but yes - we could add a kwarg like `combine` which could do this.

I'm curious if instantiating TimeSeries with `concatenate=True` does this already or not. I'll look into it.

sunpy/timeseries/timeseriesbase.py
chouhanaryan: Changed to `others` 👍

chouhanaryan:
I did not mean to close the PR, sorry about that 🤦‍♂️

nabobalis:
It's fine. It doesn't effect anything.

sunpy/timeseries/timeseriesbase.py
```diff
@@ -376,28 +378,55 @@ def concatenate(self, otherts, same_source=False, **kwargs):
...
-        kwargs['sort'] = kwargs.pop('sort', False)
-        meta = self.meta.concatenate(otherts.meta)
-        data = pd.concat([self._data.copy(), otherts.to_dataframe()], **kwargs)
+        if same_source:
+            if isinstance(others, Iterable):
+                if not all(isinstance(series, self.__class__) for series in others):
+                    raise TypeError("TimeSeries classes must match if 'same_source' is specified.")
+            else:
+                if not isinstance(others, self.__class__):
+                    raise TypeError("TimeSeries classes must match if 'same_source' is specified.")
```
chouhanaryan, line 408: I feel like this might be refactorable to a certain extent since both error messages are the same.

The reason we check if it's an iterable first is that if it's not, the call to `all()` throws a TypeError which stops execution. Another way of doing it might be to use a try-except, but even that isn't significantly different from what this already does.
Something along the lines of -

```python
if same_source:
    try:
        if not all(isinstance(series, self.__class__) for series in others):
            raise TypeError("...")
    except TypeError:
        if not instance(others, self.__class__):
            raise TypeError("...")
```

nabobalis: Would an `or` not work?

chouhanaryan: Yeah, I tried that! So say we write the following -

```python
if same_source:
    if (
        isinstance(others, Iterable)
        and not all(isinstance(series, self.__class__) for series in others)
    ) or not isinstance(others, self.__class__):
        raise TypeError("TimeSeries classes must match if 'same_source' is specified.")
```

and have `goes_1` and `goes_data = [goes_2, goes_3, goes_4]` and you try the following - `goes_1.concatenate(goes_data, same_source=True)`, it will throw an error (even though it shouldn't).

What's happening is -

`isinstance(others, Iterable)` = True (which we have in place to short-circuit the first check if an iterable isn't passed) and `not all(isinstance(series, self.__class__) for series in others)` = False, so the first check becomes False. It goes on to the second check where `isinstance(others, self.__class__)` = True because `others` is a list. 

So to solve it we need to add something to short-circuit the second check if an iterable is passed to it which is basically `not isinstance(others, Iterable)`.

If that's what we're writing then we might as well make it more readable with cleaner if-statements.

Something tells me that I might be missing something blatantly obvious though, but I just can't figure out what 😓

sunpy/timeseries/timeseriesbase.py
chouhanaryan: I've renamed `otherts` to `others` in `timeseriesbase.py` and `tsmetadata2` to `others` in `metadata.py` to make it consistent.

sunpy/timeseries/timeseriesbase.py
nabobalis: Maybe a chain of if/elif/else statements makes more sense.

sunpy/timeseries/tests/test_timeseriesbase.py
chouhanaryan: No, it doesn't do that. It just calls the `concatenate` method and it doesn't do that so invariably it doesn't either.

sunpy/timeseries/timeseriesbase.py
chouhanaryan: Something like this?

```python
if (
    same_source
    and isinstance(others, Iterable)
    and not all(isinstance(series, self.__class__) for series in others)
):
    raise TypeError("TimeSeries classes must match if 'same_source' is specified.")
elif (
    same_source
    and not isinstance(others, Iterable)
    and not isinstance(others, self.__class__)
):
    raise TypeError("TimeSeries classes must match if 'same_source' is specified.")
```

nabobalis: Sure

nabobalis:
Sorry @chouhanaryan, I forgot to review this.

nabobalis (approved):

kakirastern (approved):

Cadair (approved):

nabobalis:
Thanks for the PR @chouhanaryan

PR #4502, opened 2020-09-24
Changes to `map.rotate`: Compatibility with custom affine_transforms via `method=` argument

### LINKED ISSUE #4478: OpenCV as alternative library for map.image.affine_transform
### Description

As part of the [Helio Hackweek](https://github.com/heliohackweek/AIA_acceleration) a few weeks ago, Raphael Attie (@WaaallEEE) and I experimented with using the OpenCV python library, specifically `cv2.warpAffine`, as an alternative to `scipy.ndimage.transform` and `skimage.warp`. I have now figured out the nuances regarding different coordinate systems and order of operations, and have been able to essentially reproduce the current behavior of `map.rotate` and `image.affine_transform` using `cv2.warpAffine` as the underlying image manipulator.

I've added my [changes to a local fork](https://github.com/cbard/sunpy/blob/opencv2_warp/sunpy/image/transform.py#L123) and run [some benchmarking tests](https://github.com/cbard/sunpy/blob/opencv2_warp/open_cv_tests/CV_rotate_tests2.ipynb). Basically, `openCV.warpAffine` is 3x faster than `skimage.warp` on my laptop.

I have not created a pull request, since discussions with @wtbarnes and @Cadair suggest that it may be better to add optional argument(s) in `image.affine_transform` and `map.rotate` that allow for the passing in of custom `affine_transform` functions and their related `**kwargs`. My current fork has the openCV implementation hardcoded, so I wanted to get some information on the best way to proceed.

A difficulty with passing through a general function (+arguments) is that the required arguments for the underlying image manip routines are calculated within `image.affine_transform` or `map.rotate`; i.e. they are hidden from the general user. Even passing in an angle or rotation matrix to `map.rotate` may require some additional processing to calculate a rotation matrix and put it in the right format for the underlying routine. It seems weird to me to have the user do the required calculations and then pass the resulting arguments through `rotate()`. That kinda defeats the purpose of having `rotate()` in the first place.

cbard:
### Description

Fixes #4478.  Changes `map.rotate` and `image.transform.affine_transform` to allow for passing-in of custom affine_transform functions; otherwise keeps the default functionality of `skimage.transform.warp` and `scipy.ndimage.interpolation.affine_transform`. Changes `use_scipy=` argument to `method=` argument. Added example of passing in a custom function to `map.rotate`.

NOTE: I'm not sure where the tests are; I imagine changing the function arguments of `rotate` and `affine_transform` may cause some to fail.


Cadair:
Thanks for opening this PR @cbard I will try and get to it in the next few days :smile:

Cadair:
sorry for not getting to this yet, it's a PR with a lot of ramifications and I want to make sure I give it the attention it deserves.

sunpy/image/transform.py
```diff
@@ -55,32 +60,206 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
...
     IDL's ``rot`` calls the `POLY_2D <https://www.harrisgeospatial.com/docs/poly_2d.html>`__
     method to calculate the inverse mapping of original to target pixel
     coordinates. This is a polynomial geometrical transformation.
     Then optionally it uses a bicubic convolution interpolation
     algorithm to map the original to target pixel values.
+
+    If a custom transform function is passed to `method`, it must have the function signature:
+    `foo(image,rmatrix,order,scale,missing,image_center,recenter)`
+    (identical to the parent affine_transform, without the `method` argument),
+    and it must return the new rotated, scaled, translated image.
```
Cadair, line 86: My largest concern about this approach is encapsulated in these two sentences. Arguably however, it's a problem which already exists.

It's very easy for us to specify what the arguments of the function must be, takes these arguments etc. It's a lot harder however for us to describe the desired (default) behaviour of the function, such that it would be comparable to the default implementation. This probably isn't something we need to worry about in this PR however, although at somepoint we really should test out the properties of our transform methods.

examples/map/rotate_custom_affine_transform_opencv.py
```diff
@@ -0,0 +1,158 @@
...
+#     The point in the image to rotate around (axis of rotation).
+#     Defaults to the center of the array.
+# recenter : `bool` or array-like, optional
+#     Move the axis of rotation to the center of the array or recenter coords.
+#     Defaults to `True` i.e., recenter to the center of the array.
+
+
+# NOTE: the required libraries have already been imported globally
+# (see above: cv2, np, numbers)
+def cv_rotate(image, rmatrix, order, scale, missing, image_center, recenter):
```
Cadair, line 47: I think we should add this as an option to the built in affine_transform function and not just as an example :)

sunpy/image/transform.py
```diff
@@ -55,32 +60,206 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
...
+    missing : `float`
+        The value to replace any missing data after the transformation.
+    image_center : tuple, optional
+        The point in the image to rotate around (axis of rotation).
+        Defaults to the center of the array.
+    recenter : `bool` or array-like, optional
+        Move the axis of rotation to the center of the array or recenter coords.
+        Defaults to `True` i.e., recenter to the center of the array.
+
+    """
```
Cadair, line 214: we should move the scipy import into this function as well.

Cadair (commented):
Thanks so much for taking this on @cbard I think this has the possibility to be a real improvement in this very crusty corner of the library! That being said there should definitely be a here be dragons warning on walking into this :rofl: 

I have added a set of initial comments based on the biggest things I think should be adjusted. There are some other considerations, like deprecation and improving the example which I think we should look at when we have settled on the design.

I think that if we are going to add this flexibility to the functions we should add the opencv implementation as a built in option.

Finally, I would love to see this in the 2.1 release if you think you have enough time to wrap it up in the next couple of weeks. I promise I will be faster on subsequent reviews, sorry it took me so long to do this one!

sunpy/image/transform.py
cbard: Yeah, I'm not sure the best way to describe the desired result of `rotate`. The chief problem I had developing the openCV equivalent was that the library handles rotation, translation, and scaling operations in a different order and a different coordinate system than skimage/scipy. Even skimage and scipy have different systems between one another! 

Ultimately, we can't control how external libraries implement their manip routines. I think all we can do is to say that the image is rotated by `angle` degrees counterclockwise, translated to the image center (or input argument), and scaled by the `scale` factor; the user would have to do the legwork to verify that their external library operation gives a similar result as the current implementation.

We could make that legwork easier by defining a comparison metric for testing. I have attempted to do this via the symmetric mean absolute percentage error (as seen in the CV example). Otherwise, I'm perfectly happy to boot this to future developers 😄

Cadair: I think having a more detailed description of the algorithm in the docs and then maybe a gallery example describing how to do some comparisons would be great. Otherwise, yeah I think we should punt this to a new issue or something.

sunpy/image/transform.py
```diff
@@ -55,32 +60,206 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
...
+        raise ValueError("argument `method` must be a string or function")
+
+    # do affine transform with selected method
+    try:
+        rotated_image = method(image=image, rmatrix=rmatrix,
+                               order=order, scale=scale,
+                               missing=missing, image_center=image_center,
+                               recenter=recenter)
+    except (ImportError, NameError):
+        warnings.warn('library in affine transform `method` was not found. Using `scipy` method',
```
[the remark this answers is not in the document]
cbard: Done. Code is a bit messy in this portion, but reproduces old behavor (if `method = 'none'`).

examples/map/rotate_custom_affine_transform_opencv.py
cbard: Done.

sunpy/image/transform.py
```diff
@@ -11,14 +12,15 @@
 
 __all__ = ['affine_transform']
 
-
 def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
-                     recenter=False, missing=0.0, use_scipy=False):
+                     recenter=False, missing=0.0, method='skimage'):
```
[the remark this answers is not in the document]
cbard: What's the best way to do this? I have raised a `UserDeprecationWarning`, but is there a decorator I should be using?

sunpy/image/transform.py
cbard: Ok, I've taken a stab at the description of the algorithm in the docstring, and added a gallery example with map comparisons (in addition to the current one demonstrating a drop-in rotation function). Let me know how these could be improved!

examples/map/compare_rotation_results.py
```diff
@@ -0,0 +1,47 @@
+# # Comparing Results of `map.rotate` Between Library Implementations
```
dstansby, line 1: The formatting of this example is a bit off, in particular
- Needs a short title
- Needs a description after the title
- The top comment should be formatted with a multiline comment, ie. `"""` 
- The comments below should have lines of `#` to create sections

Looking at other examples should help, but do ask if any of that doesn't make sense!

examples/map/compare_rotation_results.py
```diff
@@ -0,0 +1,47 @@
...
+import sunpy.map
+import numbers
+import numpy as np
+import cv2
+
+# Rotating a map in Sunpy has a choice between three libraries: `skimage', `scipy`, and `cv2`.
+# One can also create a custom rotation function and drop it into `map.rotate`.
+# Since all these options use different algorithms, we need a basis for comparison:
+# what is an acceptable margin of difference in the final, rotated data product?
+# One option is to use the Symmetric Mean Absolute Percentage Error
```
dstansby, line 14: I'm not sure calling this the "error" is right for this application. The rotation algorithms are just different (one is not "correct" compared to the others), and we're measuring differences between them, not errors. Could we rename "error" > "difference" in this section?

examples/map/compare_rotation_results.py
```diff
@@ -0,0 +1,47 @@
...
+# (https://en.wikipedia.org/wiki/Symmetric_mean_absolute_percentage_error;
+# the third equation in the link). This is a bit rudimentary, since it looks
+# at all of the pixels in the image (and not just the solar disk), but it works
+# for a quick check.
+
+# SymmetricMeanAbsolutePercentageError
+# returns between 0%-100% of average error between arr1 and arr2
```
dstansby, line 21: These two lines should be moved to a docstring of the function using a multi-line `"""` comment.

examples/map/compare_rotation_results.py
```diff
@@ -0,0 +1,47 @@
...
+print("sk vs. cv", smape(map_r_sk.data, map_r_cv.data))
+
+# This essentially means that the average difference between every pixel in the `skimage` and
+# `scipy` images is about 1.9\%, and the `cv2` implementation results in about a 1.3-1.4\%
+# average difference in the final rotated image. So, the `openCV` implementation is at least within
+# the accepted difference between `scipy` and `skimage`.
+
+# A visual comparison:
+
+map_r_sk.peek()
```
dstansby, line 45: Would it also be possible to create and plot difference maps between each of the rotation methods?

examples/map/rotate_custom_affine_transform_opencv.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python
+# coding: utf-8
+
```
dstansby, line 3:
```suggestion
```
Don't need these lines

examples/map/rotate_custom_affine_transform_opencv.py
```diff
@@ -0,0 +1,118 @@
...
+# However, the `method=` argument in `sunpy.map.rotate` can accept a custom function designed
+# to use an external library. Here, we illustrate this process by defining a function
+# with the OpenCV library; this will be identical to the built-in `cv2` option.
+
+# First, our custom method must have a similar function call to
+# `sunpy.image.transform.affine_transform` and return the same output,
+# namely the rotated, scaled, and transformed image.
+# The required arguments are:
+
+# Parameters
```
dstansby, line 29: I don't think this should be documented in the example as it makes the comment quite long. Can you link to the equivalent bit in the docs for `rotate` instead?

examples/map/rotate_custom_affine_transform_opencv.py
dstansby: Does that mean this example should be removed?

sunpy/image/transform.py
```diff
@@ -3,22 +3,25 @@
...
 import numpy as np
-import scipy.ndimage.interpolation
 
-from sunpy.util.exceptions import SunpyUserWarning
+from sunpy.util.exceptions import SunpyUserWarning, SunpyDeprecationWarning
 
 __all__ = ['affine_transform']
 
-
+# TODO: add deprecation for `use_scipy`
```
dstansby, line 14: Is this still TODO?

sunpy/image/transform.py
```diff
@@ -41,10 +46,19 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
...
+        If `scipy`, uses :func:`scipy.ndimage.interpolation.affine_transform`.
+        If `cv2`, uses :func:`cv2.warpAffine`.
+        2. Elif function, uses user-defined function to perform affine transform.
+        See `notes` for function requirements.
+        Default: `none`: Will attempt to use :func:`skimage.transform.warp`;
+        on ImportError, will use :func:`scipy.ndimage.interpolation.affine_transform`.
+        (This behavior is identical to the now-deprecated `use_scipy=False`)
+        (Option `none` will be deprecated in future releases. Please explicitly set method.)
+    use_scipy : `bool`, to be deprecated
+        Equivalent to setting `method=scipy`; please do that instead.
```
dstansby, line 61:
```suggestion
```
Don't need to document a depreacted parameter

sunpy/image/transform.py
```diff
@@ -41,10 +46,19 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
...
+        1. If `string`: `skimage`, `scipy`, or `cv2'
+        If `skimage`, uses :func:`skimage.transform.warp`.
+        If `scipy`, uses :func:`scipy.ndimage.interpolation.affine_transform`.
+        If `cv2`, uses :func:`cv2.warpAffine`.
+        2. Elif function, uses user-defined function to perform affine transform.
+        See `notes` for function requirements.
+        Default: `none`: Will attempt to use :func:`skimage.transform.warp`;
+        on ImportError, will use :func:`scipy.ndimage.interpolation.affine_transform`.
+        (This behavior is identical to the now-deprecated `use_scipy=False`)
+        (Option `none` will be deprecated in future releases. Please explicitly set method.)
```
dstansby, line 59: Can the default be `'scipy'` instead of `'none'`, to avoid us having to deprecate this in the future?

sunpy/image/transform.py
```diff
@@ -55,32 +69,326 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
...
     IDL's ``rot`` calls the `POLY_2D <https://www.harrisgeospatial.com/docs/poly_2d.html>`__
     method to calculate the inverse mapping of original to target pixel
     coordinates. This is a polynomial geometrical transformation.
     Then optionally it uses a bicubic convolution interpolation
     algorithm to map the original to target pixel values.
+
+    Method
+    ------
+    If a custom transform function is passed to `method`, it must have the function signature:
+    `foo(image,rmatrix,order,scale,missing,image_center,recenter)`
```
dstansby, line 98:
```suggestion
    `custom_transform(image, rmatrix, order, scale, missing, image_center, recenter)`
```

dstansby (commented):
I haven't looked at the implementations yet, but here's some comments on the API and docs. I hope they're helpful!

cbard:
Thanks @dstansby! I will incorporate these suggestions and upload new commits shortly.

examples/map/compare_rotation_results.py
cbard: It would. Not sure the best colormap to use, so I have used the default sunpy cmap.

examples/map/compare_rotation_results.py
cbard: Ok, I have modified the formatting according to the other examples.

sunpy/image/transform.py
cbard: Another issue (which I've just realized) is that the current default behavior of `rotate` is to attempt `skimage`, but on an `ImportError`, defaults to `scipy`. I have kept this functionality, but this will have to be deprecated in the future in favor of explicitly setting method each time.

nabobalis:
Hello @cbard, sorry for the long delay in replies. 

Albert recently rewrote the API for rotate and the affine transform, you can check out https://github.com/sunpy/sunpy/pull/5867 It incorpataes a ton of the changes you did to the API here.

As a result, it will be easier to start again from scratch adding a new opencv2 function than rebasing the current code here.

What I will be doing is taking your examples and adding them in a new PR as a comparison.

nabobalis:
As a follow up, https://github.com/sunpy/sunpy/pull/6089 is now planning to add opencv2 support! If you would like to have a look @cbard, I would greatly appreciate it!

PR #4503, opened 2020-09-25
Show only headings in gallery sidebar

### LINKED ISSUE #4501: The gallery has no easy to read table of contents
https://docs.sunpy.org/en/stable/generated/gallery/index.html has no easy to read table of contents, listing each of the gallery sections. Ideally I think the table of contents in the LH sidebar shouldn't have links to the specific examples, but adding a table of contents at the top would also work.

Cadair:
fixes #4501

@dstansby is this what you had in mind?

### DIFF
--- a/docs/_templates/gallery_sidebar.html
+++ b/docs/_templates/gallery_sidebar.html
@@ -1,0 +1,25 @@
+{% set toctree = toctree(maxdepth=2, collapse=True, includehidden=True) %}
+{% if toctree %}
+{% if logo %}
+<div class="sidelogo">
+    <img class="sidelogo" src="{{ pathto('_static/' + logo, 1) }}">
+</div>
+{% else %}
+<div class="sidenologo"></div>
+{% endif %}
+<p class="doc-title">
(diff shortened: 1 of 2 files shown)


Cadair:
This is a little more hacky than I would have liked. With some modifications to the theme repo it might be able to be made better.

nabobalis:
What would need changing the theme?

wtbarnes:
I think there is overlap here with sunpy/sunpy-sphinx-theme#117

dstansby:
Yes, this is what I had in mind. I think https://github.com/sunpy/sunpy-sphinx-theme/pull/117 might be a nicer solution though.

Cadair:
I am not sure there is a one size fits all solution here, imo sunpy/sunpy-sphinx-theme#117 is too shallow for the dev guide. so i think we end up doing this either for the gallery and making sunpy/sunpy-sphinx-theme#117 depth 3 or we keep sunpy/sunpy-sphinx-theme#117 and then do this for code ref. either way it's a bit messy.

what I think might work is iterating here using this approach with the previews and the sections and then work out what the right pr to the gallery is.

nabobalis:
I have copied this to the theme repo.

PR #4506, opened 2020-09-26
[2.0] remove warning

nabobalis:


dstansby:
Which PR is this a backport of?

nabobalis:
It isn't. It removing an added dep warning.

dstansby:
As in it's undoing a backport that shouldn't have ended up in the 2.0.x branch?

nabobalis:
> As in it's undoing a backport that shouldn't have ended up in the 2.0.x branch?

No just undoing the warnings from a backport.

Cadair:
Yeah when the functionality for the new json was added we added a deprecation warning. but we shouldn't backport new dep warnings, even though we need to backport the changes.

PR #4510, opened 2020-09-28
Remove global ignore of nanoseconds warning in tests

dstansby:
I think https://github.com/sunpy/sunpy/pull/4409 should have got rid of these warnings, so see what happens if globally ignoring these warnings is removed.


nabobalis:
Test failures aren't due to this.

nabobalis (approved):

PR #4518, opened 2020-09-29
Fixed a miniscule inaccuracy in the calculation of the angular radius of the Sun

ayshih:
The function `coordinates.sun.angular_radius()` incorrectly used the arcsine instead of the arctangent for the calculation, which resulted in an inaccuracy of ~0.01 arcseconds.

### DIFF
--- a/sunpy/coordinates/sun.py
+++ b/sunpy/coordinates/sun.py
@@ -60,7 +60,7 @@ def angular_radius(t='now'):
 
 
 def _angular_radius(sol_radius, distance):
-    solar_semidiameter_rad = np.arcsin(sol_radius / distance)
+    solar_semidiameter_rad = np.arctan(sol_radius / distance)
     return Angle(solar_semidiameter_rad.to(u.arcsec))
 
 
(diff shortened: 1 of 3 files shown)


nabobalis (approved):

sunpy/coordinates/sun.py
```diff
@@ -60,7 +60,7 @@ def angular_radius(t='now'):
 
 
 def _angular_radius(sol_radius, distance):
-    solar_semidiameter_rad = np.arcsin(sol_radius / distance)
+    solar_semidiameter_rad = np.arctan(sol_radius / distance)
```
nabobalis, line 63: I don't get why codecov is saying this line was missed by the tests.

Neel2904:
I think that **arcsin()** should better work because we have approximated the distance between center of earth and sun is equal to tangent-length from earth to sun. Now, If we consider solar diameter radian then it should be like arcsin(sol_radius/tangent-length).

ayshih:
@Neel2904 Argh, you're correct that it should be arcsine, so `sun.py` was previously correct (and I had done this correctly myself in #4239).  Let this be a lesson to the dangers of multitasking during a meeting.

PR #4377, opened 2020-07-16
removed minigallery

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/4167

PR #4520, opened 2020-09-29
Fixed heading now that there is no SunPy-specific minigallery directive

ayshih:
Our version of the `minigallery` Sphinx directive was removed in #4377, so a heading in the docs should be changed.


nabobalis (approved):

nabobalis:
Example failed which failed the doc build.

Issue #4516, opened 2020-09-29
SunPy has two different functions that calculate the Sun's angular radius


wafels:
### Description

SunPy has two different functions that calculate the Sun's angular radius:

https://github.com/sunpy/sunpy/blob/master/sunpy/map/maputils.py#L84
https://github.com/sunpy/sunpy/blob/master/sunpy/coordinates/sun.py#L62

### Expected vs Actual behavior

I expect that there should be only one function in SunPy to calculate the Sun's angular radius.

### System Details

This is in the currently released version of SunPy.

ayshih:
For completeness, I'll note that this line doesn't call either function:
https://github.com/sunpy/sunpy/blob/2b8ee1b25ee16199b92e9d396675e51e457b1368/sunpy/map/header_helper.py#L227

ayshih:
As an aside, the arcsin/arctan disagreement is fixed by #4518

BassCoder2808:
So basically  we want to have only one function which will calculate the expected as well as the actual Sun's angular radius. 
But these functions are in 2 different folders. So which function do you want to retail ?
And do we have to create a new function in all or just modify it.
I would like to contribute on this issue.

nabobalis:
Hi @BassCoder2808, yes you are correct but a PR has been just opened addressing this. 

We would strongly suggest having a look at some other issues.

ayshih:
#4524 fixes all of the underlying code issues, but there still is an API issue of whether we want two functions in two different places that both return the Sun's angular radius, even though the inputs are quite different.  I'll point out that the function in `map/maputils.py` is a bit odd since it doesn't actually involve a Map in anyway, so maybe it should be moved to `coordinates/utils.py` regardless?

The residual issue is no longer a "good first issue", so I am removing the label.

jeffreypaul15:
Has this issue been resolved? Or as @ayshih stated above should the code from `map/maputils.py` be moved into `coordinates/utils.py`?

nabobalis:
I don't believe the move was made in a pull request. We should probably have go.

PR #4524, opened 2020-09-30
Fixed all calculations of the solar angular radius to use the same correct code (arcsine)

ayshih:
Context:
- #4239 fixed `coordinates.sun.angular_radius()` to use arcsine
- #4375 split off the calculation code to a private function so that it could be used in `map/mapbase.py`
- #4516 identified that other places in the codebase did not use the same code, and differed by using arctangent instead of arcsine.
- In #4518, I incorrectly changed `coordinates.sun.angular_radius()` to use arctangent to match the other places in the code, but @Neel2904 pointed out that was wrong.

This PR:
- Effectively reverts #4518 
- `map/maputils.py` and `map/header_helper.py` now use the correct arcsine instead of the incorrect arctangent because they use the private function from `coordinates.sun`.  The differences are on the order of ~0.01 arcsecond.
- Docstrings now explain why the arcsine is correct
- Fixed a bug with in `map/tests/test_mapbase.py` where the observer location was specified as 1 meter from the center of the Sun

### DIFF
--- a/changelog/4524.bugfix.rst
+++ b/changelog/4524.bugfix.rst
@@ -1,0 +1,3 @@
+All calculations of the angular radius of the Sun now use the same underlying code with the correct calculation.
+The inaccuracies were at the level of ~0.01 arcseconds.
+A previous fix to :func:`sunpy.coordinates.sun.angular_radius` was incorrect and has been reverted.
--- a/changelog/4524.bugfix.rst
+++ b/changelog/4524.bugfix.rst
@@ -1,3 +1,2 @@
-All calculations of the angular radius of the Sun now use the same underlying code with the correct calculation.
-The inaccuracies were at the level of ~0.01 arcseconds.
-A previous fix to :func:`sunpy.coordinates.sun.angular_radius` was incorrect and has been reverted.
+All calculations of the angular radius of the Sun now use the same underlying code with the accurate calculation.
+The previous inaccuracy was a relative error of ~0.001% (0.01 arcseconds) for an observer at 1 AU, but could be as large as ~0.5% for Parker Solar Probe perihelia.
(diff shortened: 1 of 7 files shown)


wafels:
After a couple of minutes with a pen and paper and most importantly, straight lines drawn on the right diagram, I agree that using arcsine is correct.

ayshih:
So no one makes this mistake again, here's a diagram showing why arcsine is correct and arctangent is not:
angular radius of the Sun

wafels:
Thanks @ayshih !  That is very clear.

changelog/4524.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+All calculations of the angular radius of the Sun now use the same underlying code with the correct calculation.
+The inaccuracies were at the level of ~0.01 arcseconds.
+A previous fix to :func:`sunpy.coordinates.sun.angular_radius` was incorrect and has been reverted.
```
Cadair, line 3: You can delete the other changelog entry. No need to confuse it.

ayshih: Sure, but should I remove this sentence too then?  It may be confusing to refer to a "previous fix" when there is no corresponding changelog entry.

Cadair: yeah, I feel like we should be writing the changelog to be read at release time, we don't really need to document release dev fluctuations in it.

ayshih: Okay, done

nabobalis (approved):

ayshih:
I clarified that the 0.01 arcsecond inaccuracy was specifically for an observer at 1 AU, and I added the inaccuracy for Parker Solar Probe perihelia.

Issue #4522, opened 2020-09-30
Remove things which are in the sunpy/.github repo


Cadair:
https://github.com/sunpy/.github/  including issue templates and CONTRIBUTING.rst

PR #4525, opened 2020-09-30
Tidy up github files

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/4522


docs/dev_guide/newcomers.rst
```diff
@@ -1 +1,269 @@
-.. include:: ../../CONTRIBUTING.rst
+.. _newcomers:
```
nabobalis, line 1: Copied this here for now.

Cadair: Isn't this somewhat defeating the purpose? It's fine for now though.

nabobalis: Sure it is but I don't have a better idea.

Cadair: yeah I guess the only real alternative is to use MyST and download it on build.

nabobalis: I will do that.

Cadair: Which I am not sure I am on board with. That being said maybe we should re-write `CONTRIBUTING.md` to be more generic and higher level and link to this?

Cadair: yeah maybe, I wonder if anyone else has opinions.

nabobalis: Will make the project wide one the same as the website.

Cadair (approved):

PR #4528, opened 2020-09-30
Added an example for drawing rectangles on maps

### LINKED ISSUE #4500: Add "drawing a rectangle on a map" example
As far as I can see there is no simple example in the gallery showing how to plot a rectangle on the map; this should have examples showing how to define the corners as pixels, and defining the corners as coordinates.

chouhanaryan:
### Description

Added an example for drawing rectangles on maps.

Fixes #4500


examples/plotting/plot_rectangle.py
```diff
@@ -0,0 +1,100 @@
...
+In this example we are going to look at how we can plot rectangles on maps
+while describing the dimensions using different methods.
+"""
+
+################################################################################
+# Fundamentally, we will be using the `~sunpy.map.GenericMap.draw_rectangle` method to
+# draw the rectangles. Its function signature is like so:
+#
+# `draw_rectangle(bottom_left, *, width: Unit(‘deg’) = None,
+# height: Unit(‘deg’) = None, axes=None, top_right=None, **kwargs)`
```
wtbarnes, line 15: Remove this. We already link to `draw_rectangle` in the API docs above. There's no need to reproduce that here.

examples/plotting/plot_rectangle.py
```diff
@@ -0,0 +1,100 @@
...
+#
+# All of the extra kwargs are passed on to the `~matplotlib.patches.Rectangle` instance,
+# which allows us to modify the color, linestyle, etc. of the plotted rectangle.
+# A full overview of available options can be found in the Rectangle documentation.
```
wtbarnes, line 19: I think this is redundant. You've already said this above and the intersphinx mapping links to the matplotlib API docs for `Rectangle`.

examples/plotting/plot_rectangle.py
```diff
@@ -0,0 +1,100 @@
...
+# Therefore, the different forms in which an input can be provided to the function are:
+#
+# * `~astropy.coordinates.SkyCoord` ((x1, x2), (y1, y2))
+# * `~astropy.coordinates.SkyCoord` bottom_left, `~astropy.coordinates.SkyCoord` top_right
+# * `~astropy.coordinates.SkyCoord` bottom_left, `~astropy.units.Quantity` width, `~astropy.units.Quantity` height
+#
+# Note that from SunPy v2.1, every argument to `~sunpy.map.GenericMap.draw_rectangle` apart from
+# 'bottom_left' (i.e. 'top_right', 'width' and 'height') should be passed as a keyword argument.
+#
+# We also have a fourth option - to specify the dimensions in terms of pixel coordinates.
```
wtbarnes, line 30: I don't think this needs to be explicitly stated here. Or at the very least, this should be broken up and interspersed with the code below.

examples/plotting/plot_rectangle.py
```diff
@@ -0,0 +1,100 @@
...
+# `~sunpy.map.GenericMap.bottom_left_coord` and `~sunpy.map.GenericMap.top_right_coord` properties
+# as parameters for `~sunpy.map.GenericMap.draw_rectangle`.
+
+import matplotlib.pyplot as plt
+
+import astropy.units as u
+from astropy.coordinates import SkyCoord
+
+import sunpy.data.sample
+import sunpy.map
```
wtbarnes, line 43: Move these to the top of the example

examples/plotting/plot_rectangle.py
```diff
@@ -0,0 +1,100 @@
...
+
+aia_1 = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
+aia_2 = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
+aia_3 = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
+aia_4 = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
```
wtbarnes, line 48: You only need one map.

examples/plotting/plot_rectangle.py
```diff
@@ -0,0 +1,100 @@
...
+# A full overview of available options can be found in the Rectangle documentation.
+#
+# Therefore, the different forms in which an input can be provided to the function are:
+#
+# * `~astropy.coordinates.SkyCoord` ((x1, x2), (y1, y2))
+# * `~astropy.coordinates.SkyCoord` bottom_left, `~astropy.coordinates.SkyCoord` top_right
+# * `~astropy.coordinates.SkyCoord` bottom_left, `~astropy.units.Quantity` width, `~astropy.units.Quantity` height
+#
+# Note that from SunPy v2.1, every argument to `~sunpy.map.GenericMap.draw_rectangle` apart from
+# 'bottom_left' (i.e. 'top_right', 'width' and 'height') should be passed as a keyword argument.
```
wtbarnes, line 28: I'm not sure about the specific mention of a version here, especially given that 2.1 hasn't been released here. I think we should just rely on the example code itself and the API docs to communicate this.

wtbarnes (changes requested):
Thanks for adding this @chouhanaryan. I think all of the components are here, but this example needs a bit of reorganization. I started adding a few line-by-line suggestions, but I'm not sure that that is terribly helpful. I would recommend having a look at the other examples in the gallery or looking at the example template in `examples/example_template/example_template.py`. In particular, the code and the text should be more interspersed as opposed to having a chunk of text at the top and then a chunk of code at the bottom.

examples/plotting/plot_rectangle.py
nabobalis: Depends on if we decide to backport this example or not.

wtbarnes: Yep that too. On top of that though, it just feels like extraneous info for an example. If I'm a user and just coming to see how to draw a rectangle on a map and I see that the `top_right` kwarg is used, that's all I care about. I'm not concerned about which versions this is/is not supported in.

chouhanaryan:
Yep, with respect to how the example is structured - I first tried to make it just like you described, but for some reason when I locally built the website, it wouldn't work. The plot just wouldn't show up.
I'll factor in all of the changes you've suggested and push it again irrespective of how it looks locally.

examples/plotting/plot_rectangle.py
```diff
@@ -0,0 +1,85 @@
...
+# We also have a fourth option - to specify the dimensions in terms of pixel coordinates.
+#
+# To do this, we need to first create a `~sunpy.map.GenericMap.submap` by passing
+# `~astropy.units.Quantity` objects with the units as pixel. Then we can use the
+# `~sunpy.map.GenericMap.bottom_left_coord` and `~sunpy.map.GenericMap.top_right_coord` properties
+# as parameters for `~sunpy.map.GenericMap.draw_rectangle`.
+fig.add_subplot(111, projection=aia_map)
+bottom_left_pixel = [400, 400] * u.pixel
+top_right_pixel = [700, 600] * u.pixel
+aia_submap = aia_map.submap(bottom_left_pixel, top_right=top_right_pixel)
```
hayesla, line 77: Instead of using submap here I'd suggest its better to use the `sunpy.map.GenericMap.pixel_to_world` to get the coordinates of top right and bottom left 

e.g.
```python
bottom_left_coord = aia_map.pixel_to_world(400*u.pixel, 400*u.pixel)
top_right_coord = aia_map.pixel_to_world(700*u.pixel, 600*u.pixel)
aia_map.draw_rectangle(bottom_left_coord, top_right=top_right_coord)
```

hayesla (changes requested):

examples/plotting/plot_rectangle.py
chouhanaryan: Done, thanks!

hayesla (approved):

kakirastern:
It is really weird, the RTD docs build failed with error output "`Build exited due to time out`". Let's see what we can do on our side.

chouhanaryan:
@kakirastern Welp, that's unexpected. Do let me know if I need to make any changes 👍🏻
Hopefully it's just a random occurrence.

examples/plotting/plot_rectangle.py
```diff
@@ -0,0 +1,82 @@
...
+import sunpy.map
+
+################################################################################
+# We use the `~sunpy.map.GenericMap.draw_rectangle` method to draw the rectangles.
+aia_map = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
+fig = plt.figure(figsize=(5, 5))
+
+################################################################################
+# The first option is to provide an input `~astropy.coordinates.SkyCoord` as a
+# (2,)-shaped tuple in the form of ((x1, x2), (y1, y2)).
```
wtbarnes, line 24:
```suggestion
# The first option is to provide the top left and bottom right corners as a 
# single `~astropy.coordinates.SkyCoord`.
```

examples/plotting/plot_rectangle.py
```diff
@@ -0,0 +1,82 @@
...
+fig.add_subplot(111, projection=aia_map)
+coords = SkyCoord((-400, 200) * u.arcsec, (-300, 100) * u.arcsec, frame=aia_map.coordinate_frame)
```
wtbarnes, line 26:
```suggestion
coords = SkyCoord(Tx=(-400, 200) * u.arcsec, Ty=(-300, 100) * u.arcsec, frame=aia_map.coordinate_frame)
```

examples/plotting/plot_rectangle.py
```diff
@@ -0,0 +1,82 @@
...
+    coords,
+    color="blue",
+    linestyle="-",
+)
+aia_map.plot()
+plt.show()
+
+################################################################################
+# The second option is to pass `~astropy.coordinates.SkyCoord` bottom_left and
+# `~astropy.coordinates.SkyCoord` top_right.
```
wtbarnes, line 37:
```suggestion
# The second option is to pass the bottom left and top right corners as 
# separate `~astropy.coordinates.SkyCoord` objects
```

examples/plotting/plot_rectangle.py
```diff
@@ -0,0 +1,82 @@
...
+aia_map.plot()
+plt.show()
+
+################################################################################
+# The third option is to provide `~astropy.coordinates.SkyCoord` bottom_left,
+# `~astropy.units.Quantity` width and `~astropy.units.Quantity` height.
+fig.add_subplot(111, projection=aia_map)
+bottom_left = SkyCoord(-500 * u.arcsec, -200 * u.arcsec, frame=aia_map.coordinate_frame)
+width = 700 * u.arcsec
+height = 500 * u.arcsec
```
wtbarnes, line 56: I think this should be the same bottom left corner as the first two. We should show how to draw the same rectangle in several different ways. 

```suggestion
width = 700 * u.arcsec
height = 500 * u.arcsec
```

examples/plotting/plot_rectangle.py
```diff
@@ -0,0 +1,82 @@
...
+    top_right=top_right,
+    color="green",
+    linestyle="--",
+)
+aia_map.plot()
+plt.show()
+
+################################################################################
+# The third option is to provide `~astropy.coordinates.SkyCoord` bottom_left,
+# `~astropy.units.Quantity` width and `~astropy.units.Quantity` height.
```
wtbarnes, line 52:
```suggestion
# The third option is to provide the coordinate of the bottom left corner along with the
# width and the height of the rectangle.
```

wtbarnes (changes requested):
Sorry this took me a bit to get back to. This is looking great @chouhanaryan! Just some minor rewording of a few of the explanations. I've also suggested not including the pixel case, but if people are in favor of keeping it, I'm ok with that.

examples/plotting/plot_rectangle.py
```diff
@@ -0,0 +1,82 @@
...
+    color="yellow",
+    linestyle="-.",
+)
+aia_map.plot()
+plt.show()
+
+################################################################################
+# We also have a fourth option - to specify the dimensions in terms of pixel coordinates.
+#
+# To do this, we use the `~sunpy.map.GenericMap.pixel_to_world` method to convert the
```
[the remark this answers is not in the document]
wtbarnes: Thinking more about it, I think this is fine to keep. the rectangle drawn here the same as the others above? If not, I think the pixel values should be adjusted so that it is (approximately) the same as above.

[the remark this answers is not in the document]
chouhanaryan: Yep - I changed it in the last commit to make it approx equal.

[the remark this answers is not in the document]
wtbarnes: 👍

wtbarnes (approved):

wtbarnes:
That 32-bit test failure is unrelated. I've no idea about the RTD fail though,

```
ERROR: Requested iniconfig from https://files.pythonhosted.org/packages/44/39/e96292c7f7068e58877f476908c5974dc76c37c623f1fa332fe4ed6dfbec/iniconfig-1.1.0.tar.gz#sha256=150a59361017218f4621a68ea9984772675a7f6e074ff7d02e115152f1804dc6 (from pytest>=4.6->pytest-astropy>=0.8->sunpy==2.1.dev474+g6aef806e4) has different version in metadata: '0.0.0'
```

ayshih (changes requested):
* You really should have at least one `plt.show()`, otherwise running this script in an interpreter does not show any plots.
* You repeatedly add the same subplot to the same `Figure` instance, and it's making matplotlib and/or WCSAxes unhappy.
* Why not just create one plot with all four rectangles on it (e.g., in the four quadrants of the solar disk)?  What's the value of five separate plots?
* You're not the only gallery example with this "sin", but you really should plot the AIA image using `clip_interval` or manually setting `vmin`/`vmax` because the AIA image does not look good.

ayshih:
> * Why not just create one plot with all four rectangles on it (e.g., in the four quadrants of the solar disk)?  What's the value of five separate plots?

If there is a desire for multiple plots, particularly one without any rectangles, you should explicitly specify which plot to use for the thumbnail, because right now the thumbnail doesn't have any rectangle.

chouhanaryan:
> You really should have at least one `plt.show()`, otherwise running this script in an interpreter does not show any plots.

Ah, I see. I'll add that back in, thanks!

> You repeatedly add the same subplot to the same `Figure` instance, and it's making matplotlib and/or WCSAxes unhappy.
> Why not just create one plot with all four rectangles on it (e.g., in the four quadrants of the solar disk)?  What's the value of five separate plots?

Yeah, that's what I originally planned to do, but the explanation for that plot wouldn't be consistent with how examples are usually structured (interspersed with details for each option explicitly) - I'd have to put all of the code together.
Is there any other way to avoid the matplotlib error?

> You're not the only gallery example with this "sin", but you really should plot the AIA image using `clip_interval` or manually setting `vmin`/`vmax` because the AIA image does not look good.

I see. Are you talking about the blank, original image only or all of them with rectangles added? I haven't come across `clip_interval` but I'll definitely have a look, thanks.

>
> If there is a desire for multiple plots, particularly one without any rectangles, you should explicitly specify which plot to use for the thumbnail, because right now the thumbnail doesn't have any rectangle.

How can I do this? I wanted to but had no idea this was possible.

ayshih:
> > You repeatedly add the same subplot to the same `Figure` instance, and it's making matplotlib and/or WCSAxes unhappy.
>
> Is there any other way to avoid the matplotlib error?

You should create a fresh `Figure` instance for each plot.

> > Why not just create one plot with all four rectangles on it (e.g., in the four quadrants of the solar disk)?  What's the value of five separate plots?
> 
> Yeah, that's what I originally planned to do, but the explanation for that plot wouldn't be consistent with how examples are usually structured (interspersed with details for each option explicitly) - I'd have to put all of the code together.

You can still have separate code blocks interspersed with explanation.  If you don't want the reader to wait until the end to see all four rectangles on a single plot, you could show your first plot with a rectangle, and then don't bother plotting the other three because the plots are essentially the same.

> > You're not the only gallery example with this "sin", but you really should plot the AIA image using `clip_interval` or manually setting `vmin`/`vmax` because the AIA image does not look good.
> 
> I see. Are you talking about the blank, original image only or all of them with rectangles added? I haven't come across `clip_interval` but I'll definitely have a look, thanks.

In all of the plots, the AIA image looks dark because the color scale is trying to accommodate a few extremely bright pixels.  `clip_interval` is described in [our docs](https://docs.sunpy.org/en/latest/guide/data_types/maps.html?highlight=clip_interval#masking-and-clipping-data), and a bunch of our examples use it.

> > If there is a desire for multiple plots, particularly one without any rectangles, you should explicitly specify which plot to use for the thumbnail, because right now the thumbnail doesn't have any rectangle.
> 
> How can I do this? I wanted to but had no idea this was possible.

You'd add a comment line saying `# sphinx_gallery_thumbnail_number = ...` ([documentation](https://sphinx-gallery.github.io/stable/configuration.html#choosing-thumbnail)).  But, I strongly feel that this script only needs one plot, so there's no need to select the plot to use for the thumbnail.  The plot with no rectangles is unnecessary.

kakirastern:
@chouhanaryan Only some final changes requested by @ayshih before we can merge your PR. Keep trying.

kakirastern (approved):

chouhanaryan:
@kakirastern Hi! I am so sorry for the delay. I have been on a hiatus for the past month to attend to some personal work. Since this PR doesn't affect release or anything substantial - I'll apply the final touches in a few weeks time. However, if it's urgent then do let me know, I'll get to it right now. Again, really sorry 😅

kakirastern:
@chouhanaryan No worries, it's good to hear from you again. Take your time to do a really good job though. 💪🏼

examples/plotting/plot_rectangle.py
```diff
@@ -0,0 +1,92 @@
+"""
+============================
+Drawing a rectangle on a map
+============================
+
+How to draw a rectangle on a map
```
nabobalis, line 6:
```suggestion
This example will demonstrate how to draw a rectangle on a sunpy map.
```

examples/plotting/plot_rectangle.py
```diff
@@ -0,0 +1,92 @@
...
+import astropy.units as u
+from astropy.coordinates import SkyCoord
+
+import sunpy.data.sample
+import sunpy.map
+
+################################################################################
+# The purpose of this example is to demonstrate how to draw a rectangle on a map
+# using :meth:`~sunpy.map.GenericMap.draw_rectangle`.  Let's start with a sample
+# AIA image.
```
nabobalis, line 19:
```suggestion
# Let's start with a sample AIA image.
```

nabobalis:
I rebased this and added my own corrections.

**SQUASH MERGE THIS**

changelog/4528.doc.rst
```diff
@@ -0,0 +1 @@
+Added a gallery example for drawing rectangles on maps.
```
ayshih, line 1:
```suggestion
Added a gallery example (:ref:`sphx_glr_generated_gallery_plotting_plot_rectangle.py`) for drawing rectangles on maps.
```

ayshih (approved):
Just needs the link to the example in the changelog entry, otherwise good to go

nabobalis:
Thanks again for the PR @chouhanaryan!

PR #4530, opened 2020-10-01
Update README.rst

muditshamz:
### Description

Fixes #<Issue Number>


kakirastern:
Hi @muditshamz Thanks for the PR, but I have a feeling the choice of spelling of `Newcomers'` instead of `Newcomer's` might have been intentional. I will let the other reviewers to chime in and see whether this change is desired or not...

nabobalis:
Yes, unless it can be proved to be incorrect that is intentional spelling.

I would suggest having a look at our open issues!

PR #4540, opened 2020-10-02
Fixed label in changelog readme

wtbarnes:
Relabels `doc` as `docfix` as we seem to only use `docfix` in all of our changelogs

- [x] Include both `doc` and `docfix`
- [x] Fix `pyproject.toml`

### DIFF
--- a/changelog/README.rst
+++ b/changelog/README.rst
@@ -21,7 +21,8 @@ Each file should be named like ``<PULL REQUEST>.<TYPE>[.<COUNTER>].rst``, where 
 * ``breaking``: A change which requires users to change code and is not backwards compatible. (Not to be used for removal of deprecated features.)
 * ``feature``: New user facing features and any new behavior.
 * ``bugfix``: Fixes a reported bug.
-* ``docfix``: Documentation addition or improvement, like rewording an entire session or adding missing docs.
+* ``doc``: Documentation addition or improvement, like rewording an entire session or adding missing docs.
+* ``docfix``: Correction to existing documentation, such as fixing a typo or adding a missing input parameter.
 * ``removal``: Feature deprecation and/or feature removal.
 * ``trivial``: A change which has no user facing effect or is tiny change.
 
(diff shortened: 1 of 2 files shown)


nabobalis:
IF doc applies to everything, then it is docfix that should be removed from the changelog entries.

wtbarnes:
Sure, but then do we relabel all of our changelog entries? There aren't any that are labeled just `doc`.

nabobalis:
> Sure, but then do we relabel all of our changelog entries? There aren't any that are labeled just `doc`.

It isn't a big thing to rename changelog files. 

I still propose that `doc` and `docfix` should exist.

wtbarnes:
👍 Sounds good to me. I don't have a strong opinion. We should just be consistent. I was confused because I saw an entry labeled with `docfix` and thought it was wrong at first because it isn't listed in the README.

nabobalis:
If we want to do this, we need to update the readme, add the config support and rename the changelogs to fit in the correct category.

dstansby (dismissed):
Marking this as needs changes, as either the config file needs changing, or we should just use `doc`.

FWIW I think it's worth having a distinction between documentation fixes and new documentation (maybe `docfix` and `docfeature`?)

nabobalis:
> Marking this as needs changes, as either the config file needs changing, or we should just use `doc`.
> 
> FWIW I think it's worth having a distinction between documentation fixes and new documentation (maybe `docfix` and `docfeature`?)

How about, doc and docfix. docfeature is a tad verbose for me.

wtbarnes:
FWIW, we discussed this at the community meeting today and the general consensus, prompted by Nabil, was to use `doc` and `docfix`

nabobalis (approved):

Cadair (approved):

nabobalis:
How do we want to handle backports?

Cadair:
yeah we should backport this

Issue #4546, opened 2020-10-05
Add coordinate-system checks to maputils functions


dstansby:
Lots of `sunpy/map/maputils.py` assumes that a helioprojective map is being passed. Therefore, if (e.g.) a synoptic map is passed into lots of those functions, I suspect there will be lots of unhelpful errors. We should

- Create a helper function `is_helioprojective_map` that returns `True` if a map is helioprojective, `False` otherwise
- Use this check in `maputil` functions
- Raise a user-friendly error if the check fails

ayshih:
I changed the title of this issue to reflect its contents.  A separate issue can be created for enhancing `maputils` functions as possible to support synoptic maps.

I'll note that there are other projected coordinate systems beyond helioprojective (e.g., any equatorial-like system using RA/dec) that all `maputils` functions in principle can support.  One sticking point is that `Helioprojective` carries with it the radius of the Sun to use, and other coordinate frames do not.

chouhanaryan:
To check if a map is helioprojective or not, is the following adequate?
```python
>>> from sunpy.coordinates import Helioprojective
>>> isinstance(map.coordinate_frame, Helioprojective)
```

ayshih:
Yup.  One `maputils` function already has such a check:
https://github.com/sunpy/sunpy/blob/332dc4ebd91e1ae62f10253c3e7ff81c06aa04fe/sunpy/map/maputils.py#L225-L226

chouhanaryan:
Would it be better to have the function check if the frame passed is helioprojective or not, as opposed to the map passed? As in, have the function `is_helioprojective_frame`. I'm asking because some functions require a check if the coordinates (not maps) they use are helioprojective or not. This function would then be able to handle both - frames of coordinates and maps.

nabobalis:
What about a more generic assert function? You pass in a map/frame and the coordinate system you are after and it checks it for you and raises an error.

chouhanaryan:
Wouldn't that force the user to pass an extra parameter/object that they're not necessarily working with directly (eg. specifying `frame='Helioprojective'/Helioprojective`)? It'd be as good as them manually checking the type using `isinstance`. I'm not too sure if that's a good thing.
In the long run, if we're supporting different types of maps then why not simply overload the function and define default behavior for when something weird is passed?
I'm an amateur when it comes to design choice so do let me know if I'm missing something.

nabobalis:
[...]

PR #4550, opened 2020-10-08
Add coordinate-system checks to maputils functions, issue #4546

abhishekpandey1880:
Added a coordinate-system checks to functions ```solar_angular_radius(coordinates)``` and ```sample_at_coords(smap, coordinates)``` in sunpy/map/maputils.py file.


dstansby (changes requested):
Thanks for opening the pull request! This is the right approach, but it would be a bit nicer to define a new function `_check_helioprojective_coords(coords)` that runs the check and raises the error. This would reduce code duplication and improve readability.

PR #4554, opened 2020-10-09
Fix ylim setting by LineAnimator

DanRyanIrish:
This PR fixes a bug which occurs in setting the ylims by ```LineAnimator``` when there are non-finite values in the data array to be animated.


dstansby:
Could you use `nanmin` and `nanmax` instead? Should make the code a little simpler.

nabobalis:
> Could you use `nanmin` and `nanmax` instead? Should make the code a little simpler.

I have updated the PR with some tests. See if it is satisfactory.

kakirastern (approved):

hayesla (approved):

Issue #4555, opened 2020-10-09
JSOC download unexpectedly returns


dstansby:
Running a medium sized download from JSOC via. Fido, the actual download is unexpectedly stopping before it is complete. The `Fido.fetch` command just seems to be returning early (before all files downloaded), as the print statement underneath is run fine.

Luckily sunpy is clever enough to cache files, and JSOC is clever enough to recognise repeat requests that I can just re-run the code a few times to get all the downloads to complete, so this isn't a massive issue.

```python
from sunpy.net import Fido, attrs as a
import sunpy.map

series = a.jsoc.Series('hmi.mrsynop_small_720s')
time = a.Time('2010/01/01', '2010/01/01')

result = Fido.search(time, series, a.jsoc.Notify("my@email.address"))

Fido.fetch(result)
print(result)
```
The search result looks like
```python
Results from 1 Provider:

138 Results from the JSOCClient:
     T_REC      TELESCOP  INSTRUME WAVELNTH CAR_ROT
--------------- -------- --------- -------- -------
Invalid KeyLink  SDO/HMI HMI_SIDE1   6173.0    2097
Invalid KeyLink  SDO/HMI HMI_SIDE1   6173.0    2098
Invalid KeyLink  SDO/HMI HMI_SIDE1   6173.0    2099
Invalid KeyLink  SDO/HMI HMI_SIDE1   6173.0    2100
Invalid KeyLink  SDO/HMI HMI_SIDE1   6173.0    2101
Invalid KeyLink  SDO/HMI HMI_SIDE1   6173.0    2102
Invalid KeyLink  SDO/HMI HMI_SIDE1   6173.0    2103
Invalid KeyLink  SDO/HMI HMI_SIDE1   6173.0    2104
Invalid KeyLink  SDO/HMI HMI_SIDE1   6173.0    2105
Invalid KeyLink  SDO/HMI HMI_SIDE1   6173.0    2106
Invalid KeyLink  SDO/HMI HMI_SIDE1   6173.0    2107
Invalid KeyLink  SDO/HMI HMI_SIDE1   6173.0    2108
Invalid KeyLink  SDO/HMI HMI_SIDE1   6173.0    2109
Invalid KeyLink  SDO/HMI HMI_SIDE1   6173.0    2110
Invalid KeyLink  SDO/HMI HMI_SIDE1   6173.0    2111
            ...      ...       ...      ...     ...
[...]

PR #4556, opened 2020-10-09
Log URLs downloaded by jsoc client

dstansby:
This might make it easier to diagnose things like https://github.com/sunpy/sunpy/issues/4555 by turning on the debug log.

### DIFF
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -16,6 +16,7 @@ import astropy.units as u
 from astropy.utils.misc import isiterable
 
 from sunpy import config
+from sunpy import log
 from sunpy.net.attr import and_
 from sunpy.net.base_client import BaseClient, BaseQueryResponse
 from sunpy.net.jsoc.attrs import walker
@@ -646,6 +647,7 @@ class JSOCClient(BaseClient):
                 print_message = "{0} URLs found for download. Full request totalling {1}MB"
                 print(print_message.format(len(urls), request._d['size']))
             for aurl, fname in zip(urls, paths):
+                log.debug(f'Adding {aurl} to download queue.')
                 downloader.enqueue_file(aurl, filename=fname, **kwargs)
 
         if dl_set and not wait:


sunpy/net/jsoc/jsoc.py
```diff
@@ -646,6 +647,7 @@ def get_request(self, requests, path=None, overwrite=False, progress=True,
                 print_message = "{0} URLs found for download. Full request totalling {1}MB"
                 print(print_message.format(len(urls), request._d['size']))
```
nabobalis, line 648: Why do we have a print statement here?

nabobalis (approved):
Happy to merge this, but I wonder if we want to do this for all downloaders?

dstansby:
Oddly this doesn't seem to be doing anything for me, so might be worth holding off until I can work out why it's not working.

Yes, I think rolling it out elsewhere would be good. Seems like a nice "good first issue"?

nabobalis (changes requested):
Holding off.

sunpy/net/jsoc/jsoc.py
```diff
@@ -646,6 +647,7 @@ def get_request(self, requests, path=None, overwrite=False, progress=True,
...
             for aurl, fname in zip(urls, paths):
+                log.debug(f'Adding {aurl} to download queue.')
```
dstansby, line 650: Well and truly baffled, this doesn't work (ie. a log message doesn't get printed), but if I put `print(log)` before this line it does...

nabobalis: Yeah this looks like it should be fine.

ayshih: > Well and truly baffled, this doesn't work (ie. a log message doesn't get printed), but if I put `print(log)` before this line it does...

Er, are you turning on debug output properly?
```python
log.setLevel("DEBUG")
```
I just tried this on my machine, and the debug log message shows up fine.

ayshih:
```
DEBUG: Adding http://jsoc.stanford.edu/SUM14/D1333490357/S00000/hmi.m_45s.20200101_010130_TAI.2.magnetogram.fits to down
load queue. [sunpy.net.jsoc.jsoc]
```

dstansby: Yeah, I'm setting the level properly. Maybe it's just something funny with my setup.

Cadair:
This might be better done in parfive?

dstansby:
> This might be better done in parfive?

That's a good idea, I shall close this in favour of some parfive logging

Issue #4553, opened 2020-10-09
Helioviewer SOHO/LASCO coronagraph images: CROTA2 header correct?


johan12345:
### Description

When looking at the FITS headers of the JPEG2000 files supplied by the Helioviewer API for SOHO/LASCO C3 and C2, I noticed that the CROTA2 header is set to non-zero values. If I understood correctly, this header contains the rotation of the image in degrees relative to solar north.

SunPy [uses this information to assign a rotation matrix to the Map](https://github.com/sunpy/sunpy/blob/master/sunpy/map/mapbase.py#L977-L989), so it is important that this is correct when e.g. plotting other features into these images.

But when looking at the Helioviewer images themselves, it looks like they are already rotated to the correct orientation, as e.g. the occulter arm in C3 is not always on the same side of the image (in contrast to unprocessed images from VSO, which have the same value in their CROTA2 header).

So, could it be that SunPy actually needs to ignore the CROTA2 header for Helioviewer images and set the rotation to zero? Or is it an issue on the Helioviewer side that this header is set in the first place?

### Steps to Reproduce

```python
from sunpy.io import read_file
from sunpy.map import Map
from sunpy.net import helioviewer
from sunpy.net import Fido
from sunpy.net import attrs as a
import datetime as dt
import matplotlib.pyplot as plt

date = dt.datetime(2013, 5, 13, 16, 54)

fig = plt.figure()

# Helioviewer image
hv = helioviewer.HelioviewerClient()
file = hv.download_jp2(date, observatory='SOHO', instrument='LASCO', detector='C3')
data, header = read_file(file)[0]

print(header['CROTA2'])  # -> -173.555

map = Map(data, header)

ax1 = fig.add_subplot(1, 2, 1, projection=map)
map.plot(axes=ax1)

# VSO image
result = Fido.search(a.Time(date, date + dt.timedelta(minutes=10)), a.Instrument.lasco, a.Detector.c3)
downloaded_files = Fido.fetch(result)
 
data, header = read_file(downloaded_files[0])[0]

print(header['CROTA2'])  # -> -173.554

map2 = sunpy.map.Map(data, header)
ax2 = fig.add_subplot(1, 2, 2, projection=map2)
map2.plot(axes=ax2)

plt.show()
```

![Figure_3](https://user-images.githubusercontent.com/5310424/95572737-af3f5680-0a2a-11eb-8362-d8df2187250c.png)

[...]

PR #4561, opened 2020-10-12
LASCOMap: clear rotation metadata for Helioviewer JPEG2000 images

johan12345:
As discussed in #4553, SOHO/LASCO coronagraph images in JPEG2000 format by Helioviewer.org are already rotated so that solar north is pointing up, but they still have the rotation metadata (CROTA2 keyword) from the original FITS file, which no longer applies here.

As suggested by @ayshih, I have added some code to clear the rotation metadata for files coming from Helioviewer.


sunpy/map/sources/tests/test_lasco_source.py
```diff
@@ -14,6 +15,10 @@
 fitspath = glob.glob(os.path.join(path, "lasco_c2_25299383_s.fts"))
 lasco = Map(fitspath)
 
+jp2path = glob.glob(os.path.join(path, "2013_05_13__16_54_06_137__SOHO_LASCO_C3_white-light.jp2"))
+lasco_helioviewer = Map(jp2path)
```
nabobalis, line 19: I know you are following the style of this file but ideally this should be a pytest fixture like so:
```python
@pytest.fixture()
def lasco_helioviewer():
    return Map(jp2path)
```

Then you can pass "lasco_helioviewer" into the test you wrote.

johan12345: sure, that makes sense. Should I do this only for `lasco_helioviewer` or also adjust the `lasco` file to use a fixture?

nabobalis: If you can that would be great.

johan12345: done in 8a458bc.

nabobalis: Thanks.

nabobalis:
Looks good to me but I will let someone with more knowledge of maps to approve this.

dstansby (changes requested):
Thanks for the PR! I think this is broadly the right approach, but we need to protect against users subsequently rotating the map, saving it, and loading it in to map again. In this case, the rotation metadata should be preserved.

In other words we need to find a robust way of knowing whether a file is a completely untouched file that has been loaded from a helioviewer file, and I don't think just checking for 'helioviewer' in the metadta is sufficient.

johan12345:
I just tested what happens in this case:

```python
# download a Helioviewer image
date = dt.datetime(2013, 5, 13, 13, 00)
hv = helioviewer.HelioviewerClient()
file = hv.download_jp2(date, observatory='SOHO', instrument='EIT', measurement='171')
data, header = read_file(file)[0]
map = Map(data, header)
print(map.rotation_matrix)
# > [[ 1. -0.] [ 0.  1.]]

map_rot = map.rotate(90 * degree)
print(map_rot.rotation_matrix)
# > [[ 6.123234e-17  1.000000e+00] [-1.000000e+00  6.123234e-17]]

map_rot.save('test.fts')
# WARNING: SunpyUserWarning: The meta key history is not valid ascii, dropping from the FITS header [sunpy.io.fits]
# WARNING: SunpyUserWarning: The meta key helioviewer is too long, dropping from the FITS header (maximum allowed key length is 8 characters). [sunpy.io.fits]

map_rot_reload = Map('test.fts')
print(map_rot_reload.rotation_matrix)
# > [[ 6.123234e-17  1.000000e+00] [-1.000000e+00  6.123234e-17]]
```

So, at the moment, the `helioviewer` key is dropped from the metadata when saving to FITS as it is too long. So, it works correctly - even though that's obviously not a perfect solution.

But I think even if the `helioviewer` key would not be dropped it would still work, because the `rotate()` function removes the `CROTAn` headers and adds `pc1_1, pc1_2, pc2_1, pc_2_2` headers instead to save the rotation matrix.

ayshih:
> So, at the moment, the `helioviewer` key is dropped from the metadata when saving to FITS as it is too long.

I think there's an argument for explicitly dropping the `helioviewer` keyword when we start mucking around with the metadata.

> In other words we need to find a robust way of knowing whether a file is a completely untouched file that has been loaded from a helioviewer file, and I don't think just checking for 'helioviewer' in the metadta is sufficient.

> But I think even if the `helioviewer` key would not be dropped it would still work, because the `rotate()` function removes the `CROTAn` headers and adds `pc1_1, pc1_2, pc2_1, pc_2_2` headers instead to save the rotation matrix.

The detection in this PR can be made more robust by checking for the existence of all three `CROTAn` keywords and performing the cleanup only if they are *all* there.  If after this cleanup code is run, some user wants to avoid using `.rotate()` and instead chooses to directly modify `CROTA2`, that would not trigger the cleanup code again as long as the user doesn't *also* inject both `CROTA` and `CROTA1` (which aren't supported keywords anyway).

johan12345:
@ayshih 
> The detection in this PR can be made more robust by checking for the existence of all three CROTAn keywords and performing the cleanup only if they are all there. If after this cleanup code is run, some user wants to avoid using .rotate() and instead chooses to directly modify CROTA2, that would not trigger the cleanup code again as long as the user doesn't also inject both CROTA and CROTA1 (which aren't supported keywords anyway).

Okay, that makes sense. I have added a check for all three CROTAn keywords exist before removing them.

Cadair:
Have we thrown rocks at helioviewer about this? it's naughty they are shipping data with lies in the header :p

ayshih:
@wafels gave a detailed comment in the associated issue (https://github.com/sunpy/sunpy/issues/4553#issuecomment-706196247).  In short, they preserved the original FITS keyword values instead of updating the metadata be consistent with modifications such as de-rotation because the files were not designed to be reused as science data.

wtbarnes:
If these files are not meant for science and are really just for a quick look , should we raise a warning indicating as much?

johan12345:
@wtbarnes 
> If these files are not meant for science and are really just for a quick look , should we raise a warning indicating as much?

Well, that might make sense, however if I understand correctly they are the only LASCO images available within SunPy that are already processed (background subtraction etc.) - the raw images that can be downloaded from VSO with `Fido` are not of much use for science either...

ayshih:
I think raising a warning would be silly.  The Helioviewer files aren't fundamentally flawed and are still worthwhile for science, even with the caveats raised by @wafels.  The files simply intentionally preserve metadata that isn't consistent with the data because their design hadn't anticipated that someone (e.g., SunPy) would blindly rely on that metadata as opposed to referring to it as a historical record.

dstansby (commented):
Looks good to me now; could you rebase this on to the current master branch, and hopefully that will trigger a run of the tests?

johan12345:
Sure, I have rebased, now waiting for the CI to run the tests :)

johan12345:
Some tests still seem to be failing, also the one that I added:
```
test_helioviewer_rotation

failed on setup with "OSError: Failed to read d:\a\1\s\.tox\py37\lib\site-packages\sunpy\data\test\2013_05_13__16_54_06_137__SOHO_LASCO_C3_white-light.jp2."
```
I'm not sure why reading this file fails, it works for me locally. It also doesn't seem to be failing in all of the test environments.

ayshih:
The reading is failing because the Windows environment doesn't have OpenJPEG for Glymur.
```
E           RuntimeError: You must have a version of OpenJPEG at least as high as 2.3.0 before you can read JPEG2000 images with glymur.  Your version is 0.0.0
```
I don't remember whether there's a workaround.

(The other test failures can be ignored.)

ayshih:
Oh, right, you can decorate the test with `@skip_glymur` so that the test will be skipped if glymur isn't there or doesn't see OpenJPEG

johan12345:
okay, let's see whether it works now

johan12345:
Now all tests are passing :)

dstansby (approved):

sunpy/map/sources/soho.py
```diff
@@ -123,6 +124,17 @@ def __init__(self, data, header, **kwargs):
         self.plot_settings['norm'] = ImageNormalize(
             stretch=source_stretch(self.meta, PowerStretch(0.5)), clip=False)
 
+        # For Helioviewer images, clear rotation metadata, as these have already been rotated.
+        # Also check that all CROTAn keywords exist to make sure that it's an untouched
+        # Helioviewer file.
+        if 'helioviewer' in self.meta \
+                and 'crota' in self.meta and 'crota1' in self.meta and 'crota2' in self.meta:
```
Cadair, line 131: I have a hatred of line continuation characters :stuck_out_tongue: 
```suggestion
        if ('helioviewer' in self.meta 
                and 'crota' in self.meta and 'crota1' in self.meta and 'crota2' in self.meta):
```

Cadair (approved):

sunpy/map/sources/soho.py
johan12345: Haha, no problem 😅 Replaced in f6496e2.

Cadair:
@johan12345 just so you know I have rebased this to try and fix the CI.

Cadair:
Thanks a lot @johan12345 :rocket:

johan12345:
Great, thank you for your help 🌞

PR #4562, opened 2020-10-12
fix names of test functions that were incorrectly named test_fitstoEIT

johan12345:
When working on #4561 I noticed that some tests in the `sunpy.map.sources.tests` module were incorrectly named `test_fitstoEIT` even though they are for other instruments than EIT (LASCO, COR, and EUVI). Probably someone forgot to rename these accordingly after copy/paste. I have now adjusted the names of these test functions.

### DIFF
--- a/sunpy/map/sources/tests/test_cor_source.py
+++ b/sunpy/map/sources/tests/test_cor_source.py
@@ -17,7 +17,7 @@ cor = Map(fitspath)
 # COR Tests
 
 
-def test_fitstoEIT():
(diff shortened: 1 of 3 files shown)


nabobalis (approved):

johan12345:
Hm, it looks like there are some test failures in CI, but if I understand it correctly these are not related to the change I made...

nabobalis:
Yeah it is a known failure.

nabobalis:
Thanks @johan12345

PR #4565, opened 2020-10-12
Propagate max_conn to fetch methods

dstansby:
Previously this wasn't handed, so for clients that created their own `Downloader` (e.g. JSOC) `max_conn` wasn't respected when passed to `Fido.fetch()`. Not sure what the best way is to add a test for this?


nabobalis (approved):

nabobalis:
Looks like it broke.

Cadair:
Doesn't Fido always pass a downloader object to the clients though? I don't think this is needed (or I am misunderstanding what's going on)

dstansby:
Not sure what's really going on underneath, but these changes update the `fetch()` methods to conform to the signature of `BaseClient.fetch()`, which is probably a good thing anyway?

dstansby:
Actual fix in https://github.com/sunpy/sunpy/pull/4567

PR #4569, opened 2020-10-13
Add Python 3.9 builds

chouhanaryan:
Add Python 3.9 builds to the CI.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -31,6 +31,9 @@ trigger:
     - '*pre*'
     - '*post*'
 
+pr:
+  autoCancel: true
+
 jobs:
 - template: run-tox-env.yml@OpenAstronomy
   parameters:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -53,6 +53,9 @@ jobs:
     envs:
       - linux: py39
         name: py39_test
+
+      - macos: py38
+        name: py38_test
         posargs: -n=1
 
       - windows: py37
@@ -63,10 +66,6 @@ jobs:
 
       - linux32: py38-32bit
         name: py38_test_32bit
-
-      - linux: codestyle
-        name: python_codestyle
-        pytest: false
 
       - linux: py38-hypothesis
 
@@ -81,6 +80,10 @@ jobs:
 
       - linux: build_docs
         posargs: " "
+        pytest: false
+
+      - linux: codestyle
+        name: python_codestyle
         pytest: false
 
 # On branches which aren't master, and not Pull Requests, build the wheels but only upload them on tags
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -51,8 +51,8 @@ jobs:
         - libpng-devel
 
     envs:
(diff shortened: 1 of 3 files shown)


dstansby:
Thanks for this! Looks like we might have to wait for https://github.com/actions/virtual-environments/issues/1740 to be resolved so that python 3.9 is available on azure

chouhanaryan:
Yep, I was just about to add a comment linking to it 😅

chouhanaryan:
I'll keep track of that issue and update the PR once it's resolved on their end.

Cadair:
We could for the moment use a docker based build if we are feeling keen: https://openastronomy-azure-pipelines.readthedocs.io/en/latest/run_tox_env.html#other-docker-images

nabobalis:
> We could for the moment use a docker based build if we are feeling keen: [openastronomy-azure-pipelines.readthedocs.io/en/latest/run_tox_env.html#other-docker-images](https://openastronomy-azure-pipelines.readthedocs.io/en/latest/run_tox_env.html#other-docker-images)

Is there really a point to this?

Cadair:
probs not

wtbarnes:
Are we sure that all of our dependencies are 3.9 compatible? I guess there's only one way to find out 😅

Cadair:
3.9 builds are live, but I don't think skimage will build without some hackery at the moment.

chouhanaryan:
Opening and closing to trigger a build. Let's see what happens.

Cadair:
ha lol @ that fail. Can you try the 3.9 build not on macos? also ping @ayshih for macos weirdness.

ayshih:
Not weirdness per se, but the NumPy check is successfully detecting that Accelerate is being linked rather than a non-buggy BLAS/LAPACK library.  From reading the log, it looks it's not the NumPy wheel that's complaining, but rather it's a bundled version of NumPy within Matplotlib 3.3.2 that's complaining?  (Is Matplotlib being built from source here?)  You could try pinning to an earlier version of Matplotlib.

Cadair:
it's probably building numpy from source, as I don't think there are wheels for 3.9 yet.

azure-pipelines.yml
```diff
@@ -50,8 +50,8 @@ jobs:
         - libpng-devel
 
     envs:
-      - macos: py38
-        name: py38_test
+      - macos: py39
```
Cadair, line 53:
```suggestion
      - linux: py39
```

ayshih:
Ah, I get the log now.  It built NumPy from source, but it's not until Matplotlib is being built from source that NumPy is imported and hence runs that check for the first time.  For NumPy <1.20, you'd need to explicitly block the linking of the Accelerate libraries on MacOS by setting an environment variable when building NumPy (see https://numpy.org/doc/stable/user/building.html#disabling-atlas-and-other-accelerated-libraries).

nabobalis (approved):

kakirastern (approved):

nabobalis:
Thanks again @chouhanaryan

Issue #3284, opened 2019-07-26
Fix the sunpy test suite to not error if packages are missing


shessweb:
### Description

I just had to reinstall my conda and python installations (anaconda3, python3.7), along with a lot of packages I need, including sunpy. I just installed it and ran sunpy.self_test(), and got several errors.

### Expected behavior

sunpy.self_test() should run with no errors.

### Actual behavior

The install was missing a couple dependencies (hypothesis and asdf). After installing the missing modules, I am getting more errors regarding the `markers` configuration options.

(conda install flaky gets rid of a few more errors, but several still occur.)

Here's the initial self_test() output I got:

```
========================================================================================= test session starts ==========================================================================================
platform linux -- Python 3.7.3, pytest-5.0.1, py-1.8.0, pluggy-0.12.0
rootdir: /my/dir/anaconda3, inifile: setup.cfg
plugins: openfiles-0.3.2, remotedata-0.3.1, arraydiff-0.3, doctestplus-0.3.0
collected 1304 items / 15 errors / 1 skipped / 1288 selected                                                                                                                                           

================================================================================================ ERRORS ================================================================================================
______________________________________________________ ERROR collecting lib/python3.7/site-packages/sunpy/coordinates/tests/test_offset_frame.py _______________________________________________________
ImportError while importing test module '/my/dir/anaconda3/lib/python3.7/site-packages/sunpy/coordinates/tests/test_offset_frame.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../anaconda3/lib/python3.7/site-packages/sunpy/coordinates/tests/test_offset_frame.py:2: in <module>
    import hypothesis.strategies as st
E   ModuleNotFoundError: No module named 'hypothesis'
___________________________________________________________ ERROR collecting lib/python3.7/site-packages/sunpy/database/tests/test_tables.py ___________________________________________________________
'flaky' not found in `markers` configuration option
_____________________________________________________________ ERROR collecting lib/python3.7/site-packages/sunpy/instr/tests/test_fermi.py _____________________________________________________________
'flaky' not found in `markers` configuration option
_____________________________________________________________ ERROR collecting lib/python3.7/site-packages/sunpy/instr/tests/test_goes.py ______________________________________________________________
'array_compare' not found in `markers` configuration option
_______________________________________________ ERROR collecting lib/python3.7/site-packages/sunpy/io/special/asdf/tags/tests/test_coordinate_frames.py ________________________________________________
[...]

PR #4586, opened 2020-10-18
Remove self_test() references from docs

dstansby:
From a user perspective, using `self_test()` does not seem terribly useful, especially since it requires many un-documented requirements to pass. This has recently caused confusion (https://github.com/sunpy/sunpy/issues/3284, https://github.com/sunpy/sunpy/issues/4480), so I think we should remove mention of it from the user facing docs.

If it is to return, we should run a CI build to check that `self_test()` works fine. I think I am +1 for deprecating and removing `self_test()` altogether though, as I'm not sure what purpose it serves.

### DIFF
--- a/docs/guide/installation/advanced.rst
+++ b/docs/guide/installation/advanced.rst
@@ -67,7 +67,7 @@ However if you want to develop sunpy we would strongly recommend reading :ref:`n
     You can read more about how to do this in the `pip documentation <https://pip.pypa.io/en/stable/user_guide/#user-installs>`__.
 
     Do **not** install sunpy or other third-party packages using ``sudo``.
-    
+
 sunpy's Requirements
 ********************
 
--- a/docs/guide/installation/advanced.rst
+++ b/docs/guide/installation/advanced.rst
@@ -67,25 +67,7 @@ However if you want to develop sunpy we would strongly recommend reading :ref:`n
     You can read more about how to do this in the `pip documentation <https://pip.pypa.io/en/stable/user_guide/#user-installs>`__.
 
     Do **not** install sunpy or other third-party packages using ``sudo``.
-
-.. _testing-sunpy:
-
-Testing sunpy
-*************
-
-.. warning::
-    The tests will fail if you do not install all the optional dependencies.
-    If you have installed sunpy with conda-forge, you will be missing the ``pytest-mock`` package which will cause the test suite to fail.
-
-The easiest way to test your installed version of sunpy is running correctly is to use the :func:`sunpy.self_test`::
-
-    import sunpy
-    sunpy.self_test()
-
-which will run many of the sunpy tests.
-
-The tests should run and print out any failures, which you can report at the `sunpy issue tracker <https://github.com/sunpy/sunpy/issues>`__.
-
+    
 sunpy's Requirements
 ********************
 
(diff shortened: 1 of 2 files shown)


nabobalis (approved):

nabobalis:
I am also in favour of removing it wholesale from the docs and from the code itself.

ayshih:
I think `self_test()` should be improved rather than removed.

> From a user perspective, using self_test() does not seem terribly useful

It verifies that something isn't borked in their Python environment for SunPy.  Depending on how the user installed SunPy, this may be unlikely right after an install, but environments change over time.

> especially since it requires many un-documented requirements to pass.

I think it's actually just one: `pytest-mock`.  I think there is protection for anything else missing.

> This has recently caused confusion (#3284, #4480), so I think we should remove mention of it from the user facing docs.

This seems like a failing that we can easily fix.  Instead of `self_test()` being a direct call to the test suite, how about making it a thin wrapper that checks for any critical dependency that may be missing and explains to the user how to fix the issue?

> I think I am +1 for deprecating and removing `self_test()` altogether though, as I'm not sure what purpose it serves.

It can also help us help users:

- Users need more guidance to use `pytest`/`tox` than to run `self_test()`, which can be done right in the Python terminal.
- Because the installation instructions tell the user to run it, it gives us the opportunity to find out about bugs that might result from odd combinations of package versions .

dstansby:
> It verifies that something isn't borked in their Python environment for SunPy. Depending on how the user installed SunPy, this may be unlikely right after an install, but environments change over time.

It doesn't seem like it was written for that purpose - it currently runs the tests, instead of checking for a non-broken sunpy installation. Being able to run the tests might be a valid proxy for having a working installation, but the error messages when the tests fail are not tailored to users who just want to know if they have a valid install.

> I think it's actually just one: `pytest-mock`. I think there is protection for anything else missing.

When I tried setting up a `tox` environment to run `self_test()` in, at least `asdf` was also required (and I didn't check past that)

> This seems like a failing that we can easily fix. Instead of `self_test()` being a direct call to the test suite, how about making it a thin wrapper that checks for any critical dependency that may be missing and explains to the user how to fix the issue?

Sounds good, but I would still advocate for removing it from the docs whilst it is confusing and not working (for a typical user install).

nabobalis:
What we want is a function that checks the current env and lists missing packages, not to run the test suite.

Cadair:
> What we want is a function that checks the current env and lists missing packages, not to run the test suite.

Why not both? I already wrote a thing in conf.py which can verify dependencies. Let's do that first, and if all the deps (including test deps) are installed let's run the test suite?

Cadair:
I do also think the current implementation of self-test is very complex, so we could also clean that up in the process.

nabobalis:
> > What we want is a function that checks the current env and lists missing packages, not to run the test suite.
> 
> Why not both? I already wrote a thing in conf.py which can verify dependencies. Let's do that first, and if all the deps (including test deps) are installed let's run the test suite?

Sure.

nabobalis:
I still think this should be removed until we fix self_test and dependency handling. Then we can add it back.

Cadair:
If I haven't opened a patch in a few days let's merge this.

nabobalis:
> If I haven't opened a patch in a few days let's merge this.

Well if you don't get to it, I might have a go this weekend.

dstansby:
Since presumably the `self_test()` won't be modified in the `2.0.x` branch, I would advocate for definitely merging this into the `2.0.x` branch whatever.

ayshih:
> Since presumably the `self_test()` won't be modified in the `2.0.x` branch

Heh, depends on your view on backporting! =P

Cadair (approved):

PR #4589, opened 2020-10-19
Allow sunpy.timeseries.TimeSeries to read in from pathlib.Path

hayesla:
This PR is to allow `pathlib.Path` objects to be passed to to `sunpy.timeseries.TimeSeries()`

sunpy.map.Map already allows this

### DIFF
--- a/changelog/4589.feature.rst
+++ b/changelog/4589.feature.rst
@@ -1,0 +1,1 @@
+Add support for pathlib.Path objects to be passed to `sunpy.timeseries.TimeSeries`.
--- a/changelog/4589.feature.rst
+++ b/changelog/4589.feature.rst
@@ -1,1 +1,1 @@
-Add support for pathlib.Path objects to be passed to `sunpy.timeseries.TimeSeries`.
+Add support for `pathlib.Path` objects to be passed to `sunpy.timeseries.TimeSeries`.
--- a/sunpy/timeseries/timeseries_factory.py
+++ b/sunpy/timeseries/timeseries_factory.py
@@ -4,7 +4,6 @@ This module provies the `~sunpy.timeseries.TimeSeriesFactory` class.
 import os
 import copy
 import glob
-import pathlib
 from collections import OrderedDict
 from urllib.request import urlopen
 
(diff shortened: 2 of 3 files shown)


nabobalis (approved):

dstansby (commented):
Does this work with windows paths? `PosixPath` is a subset of all possible `Path` types (see the diagram at the top of https://docs.python.org/3/library/pathlib.html). Off the top of my head I think the check should be `isinstance(arg, pathlib.Path)` instead.

nabobalis:
Assuming the offline tests pass on windows, we should find out.

hayesla:
good point - it seems to be the windows fail

sunpy/timeseries/timeseries_factory.py
```diff
@@ -247,6 +248,7 @@ def _parse_args(self, *args, **kwargs):
         * tuples of (data, header, unit) (1)
         * data, header not in a tuple (1)
         * filename, which will be read
+        * pathlib.PosixPath, which is the filepath to a file to be read.
```
dstansby, line 251:
```suggestion
        * `pathlib.Path`, which is the filepath to a file to be read.
```

dstansby (changes requested):
Needs a changelog, but otherwise looks 👍

nabobalis:
Oh the changelog bot is broken.

hayesla:
what type of changelog would this be? A feature/trivial/bugfix?

nabobalis:
Lets go with feature.

sunpy/timeseries/timeseries_factory.py
```diff
@@ -324,6 +326,11 @@ def _parse_args(self, *args, **kwargs):
                 result = self._read_file(path, **kwargs)
                 data_header_pairs, filepaths = _apply_result(data_header_pairs, filepaths, result)
 
+            # pathlib.Path
+            elif (isinstance(arg, pathlib.Path)):
```
Cadair, line 330: It might be better to use `os.PathLike` both here and in Map? (I only learned this existed today).

hayesla: sure, probably better.

Map doesn't use `isinstance` like here, it uses the [_possibly_pathlike] (https://github.com/sunpy/sunpy/blob/master/sunpy/map/map_factory.py#L397)

dstansby (approved):

changelog/4589.feature.rst
```diff
@@ -0,0 +1 @@
+Add support for pathlib.Path objects to be passed to `sunpy.timeseries.TimeSeries`.
```
Cadair, line 1:
```suggestion
Add support for `pathlib.Path` objects to be passed to `sunpy.timeseries.TimeSeries`.
```

Cadair (approved):

dstansby:
Test failures look unrelated

Issue #4174, opened 2020-05-16
Functionality for parse_time to take list as input


hayesla:
### Description
It would be super handy if `sunpy.time.parse_time` could take in a list of time inputs and return a list of the `astropy.time.Time` objects.

I find myself again and again doing things like
```python

time_dt = [parse_time(x) for x in list_of_times]
```

where it would be great just to do
```python
time_dt = parse_time(list_of_times)
```

What are peoples thoughts on this?

ayshih:
Is there a chance that something is out of date in your setup?  Currently, if you call `parse_time()` on a list of time inputs, you get back an array `astropy.time.Time` object.

Or, are you specifically requesting that it return a list of scalar `astropy.time.Time` objects *instead* of a single, array `astropy.time.Time` object?  I'd be resistant to that notion.

ayshih:
Ah, my mistake!  It's been a while since I've worked with inputs that couldn't be directly parsed by `astropy.time.Time` (e.g., `'2001-Jan-01'` instead of `'2001-01-01'`).  This is only a problem for certain types of inputs in a list.  For example, the current behavior is:
```python
>>> parse_time(['2001-01-01', '2001-01-02'])
 <Time object: scale='utc' format='iso' value=['2001-01-01 00:00:00.000' '2001-01-02 00:00:00.000']>

>>> parse_time(['2001-Jan-01', '2001-Jan-02'])
(errors out)
```

Absolutely, `parse_time()` should be fixed so that the two examples above return the same output.

hayesla:
oh strange! - yes I was working with `%Y-%b-%d` formats. Yes an array of `astropy.time.Time` object return is exactly what I'm looking for (rather than a list of `astropy.time.Time` objects)

I'l look into this some more for a fix

dstansby:
I think if you pass the right time format to `parse_time` this should work fine?

hayesla:
I think this is actually a 🐛 

For example this works:

```python
>>> time_list1 = ['2001-01-01', '2001-01-02', '2001-01-03']
>>> parse_time(time_list1[0])
        <Time object: scale='utc' format='isot' value=2001-01-01T00:00:00.000>
>>> parse_time(time_list1)
       <Time object: scale='utc' format='iso' value=['2001-01-01 00:00:00.000' '2001-01-02 00:00:00.000'
 '2001-01-03 00:00:00.000']>
```

[...]

PR #4590, opened 2020-10-19
Correctly parse lists of times

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/4174

### DIFF
--- a/sunpy/time/time.py
+++ b/sunpy/time/time.py
@@ -198,11 +198,14 @@ def convert_time_astropy(time_string, **kwargs):
 @convert_time.register(list)
 def convert_time_list(time_list, format=None, **kwargs):
     item = time_list[0]
+    # If we have a list of strings, need to get the correct format from our
+    # list of custom formats
     if isinstance(item, str):
         if not format:
             format = _get_time_fmt(item)
         return Time.strptime(time_list, format, **kwargs)
 
+    # Otherwise return the default method
     return convert_time.dispatch(object)(time_list, format, **kwargs)
 
 
--- a/sunpy/time/time.py
+++ b/sunpy/time/time.py
@@ -199,10 +199,9 @@ def convert_time_astropy(time_string, **kwargs):
 def convert_time_list(time_list, format=None, **kwargs):
     item = time_list[0]
     # If we have a list of strings, need to get the correct format from our
-    # list of custom formats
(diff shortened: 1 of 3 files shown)


hayesla (approved):

nabobalis (approved):

nabobalis:
I think we should add a bugfix changelog.

sunpy/time/time.py
```diff
@@ -195,6 +195,20 @@ def convert_time_astropy(time_string, **kwargs):
...
 
+@convert_time.register(list)
+def convert_time_list(time_list, format=None, **kwargs):
+    item = time_list[0]
+    # If we have a list of strings, need to get the correct format from our
+    # list of custom formats
+    if isinstance(item, str):
+        if not format:
+            format = _get_time_fmt(item)
+        return Time.strptime(time_list, format, **kwargs)
```
ayshih, line 206: Test failure is real.  The `format` that can get passed in is a `Time` format (e.g., `'isot'`), but you're passing that on as if it were a string format (e.g., `'%Y-%m-%d'`).

ayshih (changes requested):

sunpy/time/time.py
dstansby: Thanks - I think this is fixed now. I assumed if a user passed a custom format they just wanted to use `Time`, so avoided doing anything else in that case.

sunpy/time/time.py
```diff
@@ -195,6 +195,19 @@ def convert_time_astropy(time_string, **kwargs):
...
 
 
+@convert_time.register(list)
+def convert_time_list(time_list, format=None, **kwargs):
+    item = time_list[0]
+    # If we have a list of strings, need to get the correct format from our
+    # list of custom formats.
+    if isinstance(item, str) and format is None:
+        format = _get_time_fmt(item)
+        return Time.strptime(time_list, format, **kwargs)
```
ayshih, line 205:
```suggestion
        string_format = _get_time_fmt(item)
        return Time.strptime(time_list, string_format, **kwargs)
```
I suggest using a different name for the variable to avoid confusing future maintainers

ayshih (approved):
Just one suggestion

ayshih:
Squash merging via GitHub

Issue #3997, opened 2020-04-07
Re-projection of LASCO images


wafels:
### Description

It would be nice if SunPy could re-project LASCO images from different points of view.  A fundamental issue here is you have to make an assumption about where the off-solar surface material is in space.  One assumption is to assume that the LASCO image represents a plane in space perpendicular to the observer going through the center of Sun.  Another possible assumption (brought up by ayshih) is that the off-solar surface material lies on a sphere centered at the observer that passes through the center of the Sun.

### Additional context

Use cases for this functionality include being able to visualize LASCO and STEREO COR data from each other's POVs.  Further extensions of this functionality would be to visualize and re-project off disk data in general, for example STEREO HI1, HI2 and PSP WISPR data.

ehsteve:
This sounds like a terrible idea to me given that you'd have to inject some major assumptions about the three-dimensional nature of the structures in LASCO images. You might argue that we already do this for the low-corona in EUV images but you can see through so much more once you are above the solar limb.

ayshih:
As you say, the default assumption is already not correct (especially for, say, a halo CME), so this would just be a different assumption.  If you haven't already looked at it, #4003 implements this as a context manager, so it'd be an active choice by a user that they are using this different assumption.  (I'll point out that a user can't "do this themselves" by just manually specifying fully 3D coordinates because you need to affect how the `reproject` machinery works.)  Here are example images from that PR (I particularly like the lower-left panel of the AIA set):

![screen](https://user-images.githubusercontent.com/991759/78792054-a20fb500-797e-11ea-8b28-a2a91eb6392e.png)

![lasco](https://user-images.githubusercontent.com/991759/78804316-fd956f00-798d-11ea-9d18-58bbdaac9e72.png)

ehsteve:
The AIA one is cool. I just don’t think that anyone in their right mind should try to infer 3D structure from only an image without it being driven by an eruption model which goes beyond what is being described for this machinery. This is not my area of expertise so perhaps there are “default” ways of doing this accepted by the commmunity and with limitations that are understood that SunPy could support. I’ve never seen it myself though. I think we need an expert to weigh in here.

ayshih:
Closed by #4003

PR #4591, opened 2020-10-19
3D map plotting

dstansby:
This is a very very rough draft of code to enable 3D plots to be made with `Map`. This works by:

1. Calculating the coordinates of all the corners of the pixels in a map
2. Converting these to `HeliocentricInertial` coordinates (an arbitrary, but fixed choice of coordinate frame).
3. Using the x, y, z coordinate values in these cooridnates to define the corners of mesh quadrilaterals
4. Creating a `pyvista` (a python interface to VTK) mesh from these mesh coordinates
5. Plotting in `pyvista`

This is super useful when combining visualisation of multiple objects (e.g. adding magnetic field lines, or adding a LASCO map and an AIA map as in https://github.com/sunpy/sunpy/issues/3997#issuecomment-627959946)

If you checkout this branch and run `test.py` hopefully you should get something like:
aia_3d

All feedback welcome!


Cadair:
Maybe this would be better as a submodule in `sunpy.visualization`?

hayesla:
This is super cool 🌞

dstansby:
😢 readthedocs doesn't support a headless display, so I don't think pyvista examples will run on RTD (https://github.com/readthedocs/readthedocs.org/issues/1951#issuecomment-486453150)

nabobalis:
> 😢 readthedocs doesn't support a headless display, so I don't think pyvista examples will run on RTD ([readthedocs/readthedocs.org#1951 (comment)](https://github.com/readthedocs/readthedocs.org/issues/1951#issuecomment-486453150))

Given this, what about skipping the example just so people know it's possible? 
We can leave a note saying why its static.

dstansby:
Yep, that's what I was planning. Could also possibly embed a .gif or a .png with the output run locally.

Cadair:
It is technically possible to compile VTK in headless mode so it doesn't need an X server...

Cadair:
Another (easier idea) is maybe to spin this up in the conf.py: https://pypi.org/project/xvfbwrapper/

Cadair:
> It is technically possible to compile VTK in headless mode so it doesn't need an X server...

There is supposed to be a conda-forge vtk variant which is compiled with osmesa (off screen mesa GL), but it seems to have been disabled for the newest builds (despite me seemingly being able to use it with a rc of vtk 9.0) so it's possible there is a decent solution to this not that far away.

dstansby:
> Another (easier idea) is maybe to spin this up in the conf.py: https://pypi.org/project/xvfbwrapper/

Hmm, I gave this a go, but VTK doesn't seem to be detecting/using the xvfb wrapper and is just spinning up it's own window for me locally.

dstansby:
I'm wondering about spinning this off into a separate package, so I can more rapidly develop and test it without putting it all in what would be a mega PR. I think there's enough things different (needing pyvista/vtk installed, and really need a way to do image comparison tests with pyvista, which is possible, but I think worth trialing elsewhere first) to warrant this.

Issue #4389, opened 2020-07-23
Discussion on how we support GOES-XRS Data


wtbarnes:
This issue is for discussing the larger problem of how best to support both querying and loading GOES XRS data. We have a lot of PRs and issues related to this larger issue and so this (meta) issue just serves to try coordinate all of these efforts as best we can and come up with a coherent plan of how best to tackle this. 

Currently, `XRSClient` queries data from the SDAC (https://umbra.nascom.nasa.gov/goes/fits). The data are provided in FITS files and go up to GOES-15.

However, as pointed out in #4173, GOES-15 is no longer operational and is superseded by GOES-16 and GOESS-17. The new 16 and 17 data, however, are in netcdf format and we currently have no way of reading netcdf files in sunpy.

To further complicate things, as discussed in #4376 the data from GOES-13 through GOES-15 have been re-calibrated so as to match the calibration in GOES-16 and 17. This newly calibrated data, also only available in netcdf, significantly changes things like flare classification which is based on the peak GOES flux. We probably want to return the netcdf data with the new calibration by default. We should also still consider providing access to the older (GOES-15 and below) FITS data with the old calibration so that users can produce past results. At the very least, this will require a deprecation period.

In addition, PR #4359 adds support for GOES XRS JSON files that are provided by NOAA which seem to be quick-look data products from GOES-15 and 16 for 6 hour, 1 day, 3 day, and 7 day averages. I'm not quite sure how these data differ from those provided by the FITS or netcdf files or whether we should even support them at all if they can be derived easily from the data provided by the netcdf files.

## Primary Questions

* What is the canonical data product for GOES XRS? (FITS? JSON? netcdf?)
* What is the best way to support netcdf files in sunpy?
* By default, what do we return when a user makes a query for XRS data?
* Should we support querying and/or loading the longer time-averaged JSON files?

## All the Relevant Issues and PRs

* #3337
* #4173
* #4346 
* #4359 
* #4376

ayshih:
> In addition, PR #4359 adds support for GOES XRS JSON files that are provided by NOAA which seem to be quick-look data products from GOES-15 and 16 for 6 hour, 1 day, 3 day, and 7 day averages. I'm not quite sure how these data differ from those provided by the FITS or netcdf files or whether we should even support them at all if they can be derived easily from the data provided by the netcdf files.

[...]

PR #4592, opened 2020-10-21
TimeSeries support for new GOES netcdf files 

hayesla:
This is a first pass at having support for the new GOES XRS netcdf files to be read into a `sunpy.timeseries.GenericTimeSeries`. 

It works for the new GOES 13, 14, 15 re-processed files, and also for the new GOES-R series 16, 17 instruments.

Addresses : #4173 #4389 #4376 

At the moment the netcdf reader is within `sunpy.timeseries.sources.goes.py` and should be pulled out as a generic netcdf reader to `sunpy.io`. 

Because of this, the source has to be given as a keyword argument when reading in the new goes files
e.g.
```python
>>> new_goes = ts.TimeSeries("sci_xrsf-l2-flx1s_g17_d20201016_v2-0-1.nc", source='xrs')
>>> new_goes.peek()
```
Figure_1


hayesla:
some points:

* If we are to use h5netcdf it seems to have some issues with other netcdf files I played around with (e.g. TIMES/SEE data). Do we want to use another netcdf reader?
* The other `sunpy.io.read_file`  return the contents in a fits type format - this doesn't really make sense for netcdf files .... should we return something like an astropy.table that contains the attributes and variables of the netcdf?

hayesla:
I wonder if netCDF4 https://pypi.org/project/netCDF4/ would be better?

DO we have a preference on netcdf reader?

nabobalis:
Try it and if it works for all of them, we just need to figure out our depencies.

Cadair:
> If we are to use h5netcdf it seems to have some issues with other netcdf files I played around with (e.g. TIMES/SEE data). Do we want to use another netcdf reader?

Can you elaborate? I would much rather keep the binary dependency to h5py as it's much better supported and more widely used. On the other hand, if we are going to need to switch in the future, then maybe we should just accept our fate.

dstansby:
It looks (at a glance) that `netcdf4` is more mature (1.5.4 vs. 0.8.1), and maintained by an organisation as opposed to an individual. Both projects look active, but `netcdf4` is more widely used (3500 users vs. 100). I was going say go for `netcdf4`, but then I read this bit in the requirements section:

- The netCDF-4 C library from the github releases page. Version 4.1.1 or higher is required (4.2 or higher recommended). Be sure to build with --enable-netcdf-4 --enable-shared, and set CPPFLAGS="-I $HDF5_DIR/include" and LDFLAGS="-L $HDF5_DIR/lib", where $HDF5_DIR is the directory where HDF5 was installed. If you want OPeNDAP support, add --enable-dap. If you want HDF4 SD support, add --enable-hdf4 and add the location of the HDF4 headers and library to $CPPFLAGS and $LDFLAGS.

which seems quite scary 👻.

nabobalis:
I think it has wheels. I don't normally have to install the c library when I use it on Linux

dstansby:
Hmm, installing it in a clean environment only installed `cftime-1.2.1 netcdf4-1.5.4 numpy-1.19.2` for me, so I suspect that it doesn't work out of the box from `pip install netcdf4`...

Cadair:
Yeah the issue is the clib dependency, the hdf5 C lib is used by a lot more people than the netcdf4 one, so I would be much happier depending on that than the netcdf lib, it gives us the ability to support other hdf5 based files and astropy already have an optional dep on it etc.

Unless someone has some concrete arguments on the functionality side, or can correct me on the packaging issues, I am still on the side of hdf5 based code.

sunpy/timeseries/sources/goes.py
```diff
@@ -215,6 +183,40 @@ def _parse_hdus(cls, hdulist):
...
+        Parses the netCDF GOES files to return the data, header and associated units.
+
+        Parameters
+        ----------
+        filepath : `~str`
+            The path of the file to parse
+        """
+        import h5netcdf
+
+        d = h5netcdf.File(filepath, mode='r')
```
wtbarnes, line 198: Should this all be wrapped in a context manager? Or at least have a close at the end? I've never used `h5netcdf` but I think if it works the same as `h5py`, you'll need one or the other in order to not potentially leave the file open.

hayesla:
not too sure where else the dependency of h5netcdf should go? Seems to be the issue causing the failures

nabobalis:
I see this on the CI:

```python
        if filepath:
>           data, meta, units = WidgetType._parse_file(filepath)
E           TypeError: cannot unpack non-iterable ValueError object
```

hayesla:
> I see this on the CI:
> 
> ```python
>         if filepath:
> >           data, meta, units = WidgetType._parse_file(filepath)
> E           TypeError: cannot unpack non-iterable ValueError object
> ```

thanks - this should be fixed now (issue with `.fits.gz`)

Cadair:
> not too sure where else the dependency of h5netcdf should go? Seems to be the issue causing the failures

I pushed a fix (hopefully).

sunpy/timeseries/sources/goes.py
```diff
@@ -162,8 +124,13 @@ def _parse_file(cls, filepath):
...
         """
-        hdus = sunpy.io.read_file(filepath)
-        return cls._parse_hdus(hdus)
+        if ".fits" in Path(filepath).suffixes:
+            hdus = sunpy.io.read_file(filepath)
+            return cls._parse_hdus(hdus)
+        elif Path(filepath).suffix == ".nc":
+            return cls._parse_netcdf(filepath)
+        else:
+            return ValueError(f"{Path(filepath).name} is not supported. Only fits and netCDF (nc) can be read.")
```
Cadair, line 133: This feels like it's not robust enough to me. Inside `sunpy.io.read_file` we do a lot more clever stuff (https://github.com/sunpy/sunpy/blob/master/sunpy/io/file_tools.py#L79-L90) to detect if a file is FITS. We should write this in a way that uses that, and really, we should implement something similar for hdf5 / netcdf4 files.

I think the simplest option would just be to do the following, but we should really make our netcdf detection smarter than just the file extension at some point.
```suggestion
        if Path(filepath).suffix == ".nc":
            return cls._parse_netcdf(filepath)
        try:
            hdus = sunpy.io.read_file(filepath)
            return cls._parse_hdus(hdus)
        except UnrecognizedFileTypeError:
            return ValueError(f"{Path(filepath).name} is not supported. Only fits and netCDF (nc) can be read.")
```

The error class would need importing from here: https://github.com/sunpy/sunpy/blob/8e5a00849d8dbd355eaeb2cfc47109cdc63dd9d8/sunpy/io/file_tools.py#L218

sunpy/timeseries/sources/goes.py
```diff
@@ -214,14 +182,53 @@ def _parse_hdus(cls, hdulist):
...
+            xrsa = np.array(d["a_flux"])
+            xrsb = np.array(d["b_flux"])
+            times = parse_time("1970-01-01") + TimeDelta(d["time"], format="sec")
+        elif "xrsa_flux" in d.variables:
+            xrsa = np.array(d["xrsa_flux"])
+            xrsb = np.array(d["xrsb_flux"])
+            times = parse_time("2000-01-01 12:00:00") + TimeDelta(d["time"], format="sec")
+
+        else:
+            raise ValueError("Don't know how to parse this file")
```
Cadair, line 210:
```suggestion
            raise ValueError("The file {filepath} doesn't seem to be a GOES netcdf file.")
```

sunpy/timeseries/sources/goes.py
```diff
@@ -214,14 +182,53 @@ def _parse_hdus(cls, hdulist):
...
-        if 'meta' in kwargs.keys():
-            return kwargs['meta'].get('TELESCOP', '').startswith('GOES')
+        if "source" in kwargs.keys():
+            if kwargs.get("source", ""):
+                return kwargs.get("source", "").lower().startswith(cls._source)
+        if "meta" in kwargs.keys():
+            return kwargs["meta"].get("TELESCOP", "").startswith("GOES")
+
+        if "filepath" in kwargs.keys():
+            return ("xrs" and ".nc" in kwargs.get("filepath", ""))
```
Cadair, line 234: A lot of my comments above about file type detection also apply here. It would probably be wise for this check to do at least the following in order:

* Check the file extension
* If file extension is unknown, open the file and see if it looks like a netcdf4 file.
* If the file looks like a netcdf4 file (based on file extension or opening it) see if it looks like a GOES file

return False if the last one isn't true, or you never make it to the last one. A lot of this logic could also be reused for the other places in this PR where we need to check type.

Cadair (changes requested):
This generally looks great. My only concern here is that we need to be more robust in how we detect both the filetype and then decide if the file looks like a GOES netcdf.

What I would propose is that we do a little refactor of the `io.file_tools` file, so that we firstly, extend [`_detect_filetype`](https://github.com/sunpy/sunpy/blob/master/sunpy/io/file_tools.py#L159) to detect a hdf5 file based on header bytes (which I assume must exist). Then maybe extend that to detect if the hdf5 file is actually a netcdf file, but I would need to look into that more.

Also we should probably refactor file_tools to move the file extension handling in that file into a utility we can use outside of the auto loaders, so that we can use it here as well. Maybe adding a new ultimate "detect file type" function which first checks the extension and then checks the bytes.

hayesla:
thanks for this @Cadair  - yeah the filetype checking needs to be more robust, and I think having this within file_tools is the best way forward.

hayesla:
This updated PR now includes a addition of a `detect_filetype` in `file_tools.py` which includes the netcdf4 checks and the others that were originally in _`detect_filetype`. 

`_detect_filetype` now calls `detect_filetype` to check if the readers exist - so dosn't change anything on that side. 

The check for netcdf checks the header bytes, which from far as I can tell are also the same for h5 files.

In regards to the `_is_datasource_for` - it may be a bit clunky but think it probably much for robust than before

Is this along the lines of what you were thinking @Cadair ?

hayesla:
Failing issue is with exception returning False - probably a better way to do this . Anyone have thoughts?

nabobalis:
I can't see a test or a test file, do we want to add one?

hayesla:
> I can't see a test or a test file, do we want to add one?

Yes there should be a test, I'll get on this

kakirastern:
For the `codestyle` check, try running `pre-commit run --all-files` locally then do `git add .` and `git commit -m "..."` before pushing.

kakirastern:
(Codestyle `flake8` failed.)

hayesla:
yep - I've purposely done a `--no-verify` as I want to see what @Cadair thinks about how the `_is_datasource_for` is currently handled. The fail is on the fact that the except returns `None`

kakirastern:
> yep - I've purposely done a `--no-verify` as I want to see what @Cadair thinks about how the `_is_datasource_for` is currently handled. The fail is on the fact that the except returns `None`

Yeah, sounds cool 😇

setup.cfg
```diff
@@ -40,6 +40,7 @@ install_requires =
   astropy>=4.0
   parfive[ftp]>=1.1.0
   importlib_metadata;python_version<"3.8"
+  h5netcdf>=0.8.1
```
dstansby, line 43: Do we really want to add this as a base dependency, instead of an optional dependency, given it's only needed for new GOES files? Yes is an acceptable answer, but just want to keep dependency creep in check a bit.

sunpy/timeseries/sources/goes.py
```diff
@@ -214,14 +184,58 @@ def _parse_hdus(cls, hdulist):
...
+                return kwargs.get("source", "").lower().startswith(cls._source)
+        if "meta" in kwargs.keys():
+            return kwargs["meta"].get("TELESCOP", "").startswith("GOES")
+
+        if "filepath" in kwargs.keys():
+            try:
+                if sunpy.io.detect_filetype(kwargs.get("filepath", "")) == "nc":
+                    with h5netcdf.File(kwargs.get("filepath", ""), mode="r") as f:
+                        return "XRS" in f.attrs["summary"].astype("str")
+            except:
```
dstansby, line 240:
```suggestion
            except Exception as e:
```
This should placate `pre-commit`

dstansby (approved):
Looks great, thanks for working on this! CI (apart from pre-commit) failures look unrelated. I have 1 query about dependencies (see above)...

setup.cfg
wtbarnes: Fwiw my vote here would be no and that this is added to a `timeseries` or the like `extra_requires` category as HDF5 is a heavy dependency.

sunpy/timeseries/sources/goes.py
hayesla: great thanks @dstansby !

setup.cfg
hayesla: I would also agree with this. I think @Cadair had a reason for this though? however I could be wrong.

nabobalis: Going to speak again for stuart as I like to do.

The ability for time series or map to open and parse a file type it supports is core functionality.

Cadair: Nabil is basically correct with his channelling. More specifically *not* making this a core dependency is a significant change in policy. At the moment core is pretty much "the requirements needed to make map and timeseries work". I am happy to reconsider this, but we should probably do it in a dedicated issue and not on this PR.

Cadair (approved):
I think this is looking great, I have one quite pedantic comment that we should say `detect_filetype` is in fact detecting hdf5 and not netcdf4, so we should change that but otherwise I think this is good to go.

setup.cfg
Cadair: I opened https://github.com/sunpy/sunpy/issues/4655

sunpy/timeseries/sources/goes.py
```diff
@@ -214,14 +184,58 @@ def _parse_hdus(cls, hdulist):
...
+        if "source" in kwargs.keys():
+            if kwargs.get("source", ""):
+                return kwargs.get("source", "").lower().startswith(cls._source)
+        if "meta" in kwargs.keys():
+            return kwargs["meta"].get("TELESCOP", "").startswith("GOES")
+
+        if "filepath" in kwargs.keys():
+            try:
+                if sunpy.io.detect_filetype(kwargs.get("filepath", "")) == "nc":
+                    with h5netcdf.File(kwargs.get("filepath", ""), mode="r") as f:
```
Cadair, line 238: You have already checked for the existence of the key.
```suggestion
                    with h5netcdf.File(kwargs["filepath"], mode="r") as f:
```

sunpy/timeseries/sources/goes.py
```diff
@@ -214,14 +184,58 @@ def _parse_hdus(cls, hdulist):
...
-                return kwargs.get('source', '').lower().startswith(cls._source)
-        if 'meta' in kwargs.keys():
-            return kwargs['meta'].get('TELESCOP', '').startswith('GOES')
+        if "source" in kwargs.keys():
+            if kwargs.get("source", ""):
+                return kwargs.get("source", "").lower().startswith(cls._source)
+        if "meta" in kwargs.keys():
+            return kwargs["meta"].get("TELESCOP", "").startswith("GOES")
+
+        if "filepath" in kwargs.keys():
```
Cadair, line 235: I think the logic here is sound :+1:

setup.cfg
wtbarnes: Fair enough! I feel like we've had this discussion in the past and I've agreed with this line of reasoning and then forgotten 😅

sunpy/timeseries/sources/goes.py
```diff
@@ -162,8 +126,13 @@ def _parse_file(cls, filepath):
...
         """
-        hdus = sunpy.io.read_file(filepath)
-        return cls._parse_hdus(hdus)
+        if sunpy.io.detect_filetype(filepath) == "nc":
+            return cls._parse_netcdf(filepath)
+        try:
+            hdus = sunpy.io.read_file(filepath)
+            return cls._parse_hdus(hdus)
+        except UnrecognizedFileTypeError:
+            return ValueError(f"{Path(filepath).name} is not supported. Only fits and netCDF (nc) can be read.")
```
wtbarnes, line 135:
```suggestion
        try:
            hdus = sunpy.io.read_file(filepath)
        except UnrecognizedFileTypeError:
            raise ValueError(f"{Path(filepath).name} is not supported. Only fits and netCDF (nc) can be read.")
        else:
            return cls._parse_hdus(hdus)
```

sunpy/timeseries/sources/goes.py
```diff
@@ -214,14 +184,58 @@ def _parse_hdus(cls, hdulist):
...
+        elif "xrsa_flux" in d.variables:
+            xrsa = np.array(d["xrsa_flux"])
+            xrsb = np.array(d["xrsb_flux"])
+            start_time_str = d["time"].attrs["units"].astype(str).lstrip("seconds since")
+            times = parse_time(start_time_str) + TimeDelta(d["time"], format="sec")
+
+        else:
+            raise ValueError(f"The file {filepath} doesn't seem to be a GOES netcdf file.")
+
+        d.close()
```
wtbarnes, line 214: If neither `a_flux` or `xrsa_flux` are in `d.variables` and the `ValueError` is raised, doesn't this mean that the file handle will be left open? I would still argue for a context manager here, but maybe I'm just being overly cautious. 

```suggestion
        with h5netcdf.File(filepath, mode="r") as d:
            header =  MetaDict(OrderedDict(d.attrs))
            if "a_flux" in d.variables:
                xrsa = np.array(d["a_flux"])
                xrsb = np.array(d["b_flux"])
                start_time_str = d["time"].attrs["units"].astype(str).lstrip("seconds since").rstrip("UTC")
                times = parse_time(start_time_str) + TimeDelta(d["time"], format="sec")
            elif "xrsa_flux" in d.variables:
                xrsa = np.array(d["xrsa_flux"])
                xrsb = np.array(d["xrsb_flux"])
                start_time_str = d["time"].attrs["units"].astype(str).lstrip("seconds since")
                times = parse_time(start_time_str) + TimeDelta(d["time"], format="sec")
            else:
                raise ValueError(f"The file {filepath} doesn't seem to be a GOES netcdf file.")
```

sunpy/timeseries/sources/goes.py
```diff
@@ -214,14 +184,58 @@ def _parse_hdus(cls, hdulist):
...
         Determines if header corresponds to a GOES lightcurve
         `~sunpy.timeseries.TimeSeries`.
         """
-        if 'source' in kwargs.keys():
-            if kwargs.get('source', ''):
-                return kwargs.get('source', '').lower().startswith(cls._source)
-        if 'meta' in kwargs.keys():
-            return kwargs['meta'].get('TELESCOP', '').startswith('GOES')
+        if "source" in kwargs.keys():
+            if kwargs.get("source", ""):
```
wtbarnes, line 230: I'm confused as to why this additional `if` is needed? Doesn't the preceding conditional already tell you that `'source'` is in `kwargs`? Maybe I'm just confused here.

wtbarnes (approved):
Nothing that I think should block a merge. Just some nitpicky things

nabobalis (changes requested):
I am going to block until the comments are addressed.

hayesla:
> I am going to block until the comments are addressed.

working on it now!

hayesla:
done - and tests are passing 🎉

nabobalis (approved):

kakirastern (approved):

PR #4598, opened 2020-10-23
Rewrite overplotting SRS active regions example

kathivashkiv:
Modified `skip_magnetogram_active_regions_example.py` to use
HMI file from sample data instead of downloading it with Fido.

Meanwhile `Fido.search()` for SRS files needs to be fixed. It downloads file for the wrong date (or several of files). Correct year and wrong month and day. So this example works, but uses wrong SRS file.


kathivashkiv:
@Cadair @nabobalis please have a look

nabobalis:
You will need to rename example to not have "skip_" in the name.

kathivashkiv:
> You will need to rename the example to not have "skip_" in the name.

Despite the final result is incorrect cause of wrong SRS file?

nabobalis:
> > You will need to rename the example to not have "skip_" in the name.
> 
> Despite the final result is incorrect cause of wrong SRS file?

The example is still broken then?

kathivashkiv:
Yes, still broken. I can modify it to use 'path-to-file'. But this will require from users to find and download SRS file separately, as I did.

nabobalis:
I can't say I am pro changing an example that requires a manual step to make it work. I don't have any good ideas tho.

hayesla:
actually looking at this some more - its really a bug in the srs client that needs to be investigated https://github.com/sunpy/sunpy/blob/master/sunpy/net/dataretriever/sources/noaa.py#L119

kathivashkiv:
From the start the problem was in downloading HMI file with Fido. This was bypassed by changing downloading it with the HMI file from sample data. This was the first task. And then downloading of wrong SRS with Fido poped out. I don't know. Can these problems be related?

hayesla:
> From the start the problem was in downloading HMI file with Fido. This was bypassed by changing downloading it with the HMI file from sample data. This was the first task. And then downloading of wrong SRS with Fido poped out. I don't know. Can these problems be related?

I'm not too sure - it looks like the SRS client is definitely acting weird which I don't think is related to the HMI issue - I think you just discovered another bug 😆  haha!

hayesla:
it looks the SRS data for that day is available from NOAA - but the sunpy SRS client is getting the wrong day for some reason

nabobalis:
> it looks the SRS data for that day is available from NOAA - but the sunpy SRS client is getting the wrong day for some reason

Could we have broken this in master with the refactor?

hayesla:
> > it looks the SRS data for that day is available from NOAA - but the sunpy SRS client is getting the wrong day for some reason
> 
> Could we have broken this in master with the refactor?

potentially, I'm having a look at it now

hayesla:
@nabobalis  - arg its a bug here https://github.com/sunpy/sunpy/blob/master/sunpy/net/dataretriever/sources/noaa.py#L211 for files in years previous that are in tar.gz it only provides the first two days of the year (and this is what the tests are running for 🤦 )

@kathivashkiv thanks for finding this bug!

I'll open an issue and a fixing PR

nabobalis:
> @nabobalis - arg its a bug here [https://github.com/sunpy/sunpy/blob/master/sunpy/net/dataretriever/sources/noaa.py#L211](https://github.com/sunpy/sunpy/blob/master/sunpy/net/dataretriever/sources/noaa.py?rgh-link-date=2020-10-23T11%3A34%3A30Z#L211) for files in years previous that are in tar.gz it only provides the first two days of the year (and this is what the tests are running for 🤦 )

Good spot.

kathivashkiv:
> @nabobalis - arg its a bug here https://github.com/sunpy/sunpy/blob/master/sunpy/net/dataretriever/sources/noaa.py#L211 for files in years previous that are in tar.gz it only provides the first two days of the year (and this is what the tests are running for facepalm )

Oh, now it makes sense. I'm glad this was useful in the end)

nabobalis:
> > @nabobalis - arg its a bug here [https://github.com/sunpy/sunpy/blob/master/sunpy/net/dataretriever/sources/noaa.py#L211](https://github.com/sunpy/sunpy/blob/master/sunpy/net/dataretriever/sources/noaa.py?rgh-link-date=2020-10-23T11%3A55%3A24Z#L211) for files in years previous that are in tar.gz it only provides the first two days of the year (and this is what the tests are running for facepalm )
> 
> Oh, now it makes sense. I'm glad this was useful in the end)

So I would wait for that bug to be fixed and then make sure the example works as intended and we should be good to go.

kathivashkiv:
> > > You will need to rename the example to not have "skip_" in the name.
> > 
> > 
> > Despite the final result is incorrect cause of wrong SRS file?
> 
> The example is still broken then?

Actually if SRS search works correctly all example will work. This must be checked after bugfix

kathivashkiv:
One more thing


> > From the start the problem was in downloading HMI file with Fido. This was bypassed by changing downloading it with the HMI file from sample data. This was the first task. And then downloading of wrong SRS with Fido poped out. I don't know. Can these problems be related?
> 
> I'm not too sure - it looks like the SRS client is definitely acting weird which I don't think is related to the HMI issue - I think you just discovered another bug laughing haha!

Now it works. HMI file downloading with Fido. And I can't figure out why)..

nabobalis:
> One more detail
> 
> > > From the start the problem was in downloading HMI file with Fido. This was bypassed by changing downloading it with the HMI file from sample data. This was the first task. And then downloading of wrong SRS with Fido poped out. I don't know. Can these problems be related?
> > 
> > 
> > I'm not too sure - it looks like the SRS client is definitely acting weird which I don't think is related to the HMI issue - I think you just discovered another bug laughing haha!
> 
> Now it works. HMI file downloading with Fido. And I can't figure out why)..

Leave it with the sample data, we don't need it to download any data outside of the SRS data.

kathivashkiv:
Good)

kathivashkiv:
This is output for 4 regions: 
image
This is for 9:
image
Should I leave it as it is or change it? I mean overlaying region numbers

nabobalis:
I would leave it as is, others might have differning thoughts.

kakirastern:
Some additional info for the debugging: 

client = <sunpy.net.helio.hec.HECClient object at 0x7fa3d3db52e0>

    @pytest.mark.remote_data
    def test_get_table_names(client):
        tables = client.get_table_names()
        assert len(tables) == 126
        table = tables[0][0]
>       assert isinstance(table, bytes)
E       AssertionError: assert False
E        +  where False = isinstance('timed_see_flare', bytes)

../../.tox/py38-online/lib/python3.8/site-packages/sunpy/net/helio/tests/test_helio.py:282: AssertionError

kakirastern:
AttributeError: 'str' object has no attribute 'decode'

kakirastern:
@kathivashkiv You have some failures on Azure that will need to be processed before this PR can be merged, including but not limited to the codestyle test.

kakirastern:
But overall really impressive input seen in this PR

nabobalis:
We have merged in the fix for the SRS client, so you should be able to update the example to download the correct data as the sample file.

kathivashkiv:
Example works, "build_docs" succeeded.

kakirastern:
That's awesome!

examples/plotting/magnetogram_active_regions.py
```diff
@@ -14,27 +14,28 @@
 
 import sunpy.coordinates
 import sunpy.map
+import sunpy.data.sample
```
nabobalis, line 17:
```suggestion
import sunpy.data.sample
import sunpy.map
```

nabobalis: Should fix the style check.

examples/plotting/magnetogram_active_regions.py
```diff
@@ -14,27 +14,28 @@
...
 ##############################################################################
-# Let's select only the first file, download it and create a map.
-result = results[0, 0]
-file_name = Fido.fetch(result)
-smap = sunpy.map.Map(file_name)
+# Then extract the observation time of the sample image. We use it to
+# calculate the time interval for the SRS file search.
+
+start_time = parse_time(smap.date)
+end_time = start_time + TimeDelta(23*u.hour + 59*u.minute + 59*u.second)
```
nabobalis, line 34:
```suggestion
end_time = start_time + TimeDelta(23*u.day)
```
Can we not do this instead?

nabobalis:
Rendered example: https://sunpy--4598.org.readthedocs.build/en/4598/generated/gallery/plotting/magnetogram_active_regions.html#sphx-glr-generated-gallery-plotting-magnetogram-active-regions-py

nabobalis:
I think this should have a docfix changelog entry as well.

examples/plotting/magnetogram_active_regions.py
kathivashkiv: No, SRS files are published at the same time each day. I checked this, TimeDelta(1*u.day) gives two files

nabobalis: So it does work we just need to index out the first file? 

Can you add a comment as to why we had to be very specific with this timedelta?

kakirastern: > No, SRS files are published at the same time each day. I checked this, TimeDelta(1*u.day) gives two files

Good catch!

kakirastern:
For the codestyle fix, do this as prescribed by the CI/CD tests:

```python
 from astropy.time import TimeDelta
 
 import sunpy.coordinates
-import sunpy.map
 import sunpy.data.sample
+import sunpy.map
 from sunpy.io.special import srs
 from sunpy.net import Fido
ERROR: InvocationError for command /home/vsts/work/1/s/.tox/codestyle/bin/pre-commit run --verbose --all-files --show-diff-on-failure (exited with code 1)
```

In other words, move the `import sunpy.map` the right way around...

changelog/4598.docfix.rst
```diff
@@ -0,0 +1 @@
+Modified `skip_magnetogram_active_regions_example.py` to use HMI file from sample data instead of downloading it with Fido.
```
nabobalis, line 1:
```suggestion
Modified :ref:`sphx_glr_generated_gallery_magnetogram_active_regions_example.py` to use HMI file from sample data instead of downloading it with Fido.
```
I think this is correct.

kakirastern:
Try collecting all commit suggestions from @nabobalis in a batch before committing in the `/files` page. Then everything should be in order.

kakirastern:
> Try collecting all commit suggestions from @nabobalis in a batch before committing in the `/files` page. Then everything should be in order.

Or just hit the "Files changed" button up top...

examples/plotting/magnetogram_active_regions.py
kathivashkiv: > So it does work we just need to index out the first file?
> 
Yes, it works. We chose the first file later:
image

> Can you add a comment as to why we had to be very specific with this timedelta?
Perhaps then it is appropriate to replace the `srs_downloaded_files` list with `srs_downloaded_file` variable and then add such a comment about this timedelta?

nabobalis: Sounds good to me.

kathivashkiv: No, `Fido.fetch()` returns a list object, so I leave it with index access. Only adding a comment

examples/plotting/magnetogram_active_regions.py
```diff
@@ -13,34 +13,37 @@
...
 
 ##############################################################################
-# Let's select only the first file, download it and create a map.
-result = results[0, 0]
-file_name = Fido.fetch(result)
-smap = sunpy.map.Map(file_name)
+# Then extract the observation time of the sample image. We use it to
+# calculate the time interval for the SRS file search. SRS files are published
+# at the same time each day. Therefore to define `TimeDelta` we use '23*u.hour
+# + 59*u.minute + 59*u.second' interval (not '1*u.day').
```
nabobalis, line 33:
```suggestion
# Then using the observation time of the sample image we will download the corresponding
# NOAA SWPC solar region summary. These are published once a day, so we will define an
# ``end_time`` that ends just before the next day, otherwise you will download two files.
```

examples/plotting/magnetogram_active_regions.py
```diff
@@ -13,34 +13,37 @@
...
+
+start_time = parse_time(smap.date)
+end_time = start_time + TimeDelta(23*u.hour + 59*u.minute + 59*u.second)
 
 ##############################################################################
-# Download the SRS file.
+# Search and download the SRS file.
```
nabobalis, line 39:
```suggestion
#  Here we use `sunpy.net.Fido` to acquire the data.
```

kakirastern:
For the docs_build, found this which may be of some use:
```
<unknown>:196: WARNING: undefined label: sphx_glr_generated_gallery_magnetogram_active_regions_example.py (if the link has no caption the label must precede a section header)

```

changelog/4598.docfix.rst
```diff
@@ -0,0 +1 @@
+Modified :ref:`sphx_glr_generated_gallery_magnetogram_active_regions_example.py` to use HMI file from sample data instead of downloading it with Fido.
```
nabobalis, line 1: I messed up this reference.

changelog/4598.docfix.rst
```diff
@@ -0,0 +1 @@
+Modified :ref:`sphx_glr_generated_gallery_plotting_magnetogram_active_regions_example.py` to use HMI file from sample data instead of downloading it with Fido.
```
nabobalis, line 1:
```suggestion
Modified :ref:`sphx_glr_generated_gallery_plotting_magnetogram_active_regions.py` to use HMI file from sample data instead of downloading it with Fido.
```

nabobalis (approved):

kakirastern:
Thanks @kathivashkiv for your contribution! All the best for Outreachy. We appreciate your working with us.

kathivashkiv:
Thank you all for your help too!

kathivashkiv:
> For the codestyle fix, do this as prescribed by the CI/CD tests:
> 

> 
> In other words, move the `import sunpy.map` the right way around...

Is this because of alphabetical order? For the future

kakirastern:
I normally don't care and let the machine handle it for me by doing `pre-commit run --all-files` at the Terminal (I am using a MacBook Pro)

kakirastern:
Otherwise too many packages including `isort` to consider with the human mind

kakirastern:
But good question

nabobalis:
> > For the codestyle fix, do this as prescribed by the CI/CD tests:
> 
> > In other words, move the `import sunpy.map` the right way around...
> 
> Is this because of alphabetical order? For the future

Yeah. We sort all imports by alphabetical order expect the standard library which is by length.

kakirastern:
Typo: *expect -> except

PR #4605, opened 2020-10-27
Hash based figure tests with pytest-mpl

Cadair:
This is a work in progress port to using pytest-mpl with new hash based functionality implemented in https://github.com/matplotlib/pytest-mpl/pull/101

TODO:
- [x] Await merge and release upstream
- [x] Fix the baseline image updates after merge.


tox.ini
```diff
@@ -87,7 +88,7 @@ commands =
     figure: /bin/bash -c "mkdir -p ./figure_test_images; python -c 'import matplotlib as mpl; print(mpl.ft2font.__file__, mpl.ft2font.__freetype_version__, mpl.ft2font.__freetype_build_type__)' > ./figure_test_images/figure_version_info.txt"
     figure: /bin/bash -c "pip freeze >> ./figure_test_images/figure_version_info.txt"
     figure: /bin/bash -c "cat ./figure_test_images/figure_version_info.txt"
-    figure: {env:PYTEST_COMMAND} -m "figure" {posargs}
+    figure: {env:PYTEST_COMMAND} -m "mpl_image_compare" --mpl --remote-data=any --mpl-generate-summary --mpl-baseline-path=https://raw.githubusercontent.com/sunpy/sunpy-figure-tests/sunpy-master/figures/{envname}/ {posargs}
```
Cadair, line 91: If we ever needed a different set of figures per-branch we would have to manually change this URL.

Cadair:
We can't merge this until the pytest-mpl PRs land and a release happens, but it would be good to get a review from this side so I can adjust in pytest-mpl as needed.

sunpy/map/tests/test_plotting.py
```diff
@@ -82,12 +82,12 @@ def test_peek_grid_aia171(aia171_test_map):
...
 @figure_test
 def test_peek_grid_spacing_aia171(aia171_test_map):
-    aia171_test_map.peek(draw_grid=(5, 5) * u.deg)
+    aia171_test_map.peek(draw_grid=(10, 5) * u.deg)
 
 
 @figure_test
 def test_peek_limb_aia171(aia171_test_map):
-    aia171_test_map.peek(draw_limb=True)
+    aia171_test_map.peek(draw_limb=False)
```
nabobalis, line 90: This changes just to break the CI?

sunpy/tests/helpers.py
```diff
@@ -79,45 +79,25 @@ def figure_test(test_function):
...
         plt.plot([0,1])
     """
-    @pytest.mark.figure
+    ft2_version = f"{mpl.ft2font.__freetype_version__.replace('.', '')}"
+    mpl_version = "dev" if "+" in mpl.__version__ else mpl.__version__.replace('.', '')
+    astropy_version = "dev" if "dev" in astropy.__version__ else astropy.__version__.replace('.', '')
+    hash_library_name = f"figure_hashes_mpl_{mpl_version}_ft_{ft2_version}_astropy_{astropy_version}.json"
+    hash_library_file = Path(__file__).parent / hash_library_name
+
+    @pytest.mark.figure  # Add for backwards compatibility
```
nabobalis, line 88: Do we need to keep this?

tox.ini
nabobalis: This is a good question, so far we have not needed this but I wonder if that will always be the case.

setup.cfg
```diff
@@ -124,8 +126,7 @@ filterwarnings =
     ignore:leap-second auto-update failed
     # See https://github.com/astropy/extension-helpers/issues/23
     ignore:Distutils was imported before Setuptools
-    # toolz internal deprecation warning https://github.com/pytoolz/toolz/issues/500
-    ignore:The toolz.compatibility module is no longer needed:DeprecationWarning
+    ignore:Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
```
dstansby, line 130: Why is this warning being thrown?

sunpy/map/tests/test_plotting.py
dstansby, line 90: Why these changes?

sunpy/tests/helpers.py
dstansby, line 88: Is the plan to leave this here forever?

sunpy/tests/helpers.py
```diff
@@ -159,79 +139,6 @@ def _patch_coverage(testdir, sourcedir):  # pragma: no cover
     cov.save()
 
 
-html_intro = '''
```
dstansby, line 162: Noice, did you put the HTML comparison stuff in pytest-mpl too?

dstansby (commented):
Looks good overall; I've left a few comments.

sunpy/map/tests/test_plotting.py
Cadair: yep

sunpy/tests/helpers.py
Cadair: yep

setup.cfg
nabobalis: I thought this warning was thrown when you set MPL to use agg. I have seen this locally at times with the figure tests.

sunpy/tests/helpers.py
Cadair: Maybe we don't need it at all. The reason to keep it was to keep `-m "figure"` working

sunpy/map/tests/test_plotting.py
Cadair: Just to break the tests, so I could see if the HTML worked.

setup.cfg
Cadair: pytest-mpl dynamically switches the backend, but I don't understand how we avoided this in our own code before this PR. As I thought we set the backend to Agg anyway?

sunpy/tests/helpers.py
nabobalis: What would it be without this? `-m mpl-figure`?

setup.cfg
nabobalis: IN tox we do set it to AGG and I have seen this warning but it is a good question on how have we avoided this before.

sunpy/tests/helpers.py
Cadair: `-m mpl_image_compare` as done in tox.ini

setup.cfg
Cadair: we set Agg in the conftest.py as well: https://github.com/sunpy/sunpy/blob/master/sunpy/conftest.py#L17-L22

sunpy/tests/helpers.py
nabobalis: That isn't so bad. I feel like we could remove "figure"

setup.cfg
nabobalis: Do we need both?

Cadair: :man_shrugging:

dstansby: What's the stack trace when this warning is thrown? Maybe pytest-mpl needs to be setting it to agg somewhere?

sunpy/map/tests/test_plotting.py
dstansby: 👍 will leave this un-resolved so we don't forget to remove

sunpy/tests/helpers.py
dstansby: Yeah, I would say that there are probably so few people running the figure tests locally that changing this is fine.

setup.cfg
Cadair: I think this is being raised because pytest-mpl is setting it to agg and we are calling .show inside of our peek methods.

Cadair: I can't replicate this inside tox, but the tests should run outside it just as well. The full traceback is:

```
___________________________________________________________ test_heliographic_peek ____________________________________________________________

cls = <class 'matplotlib.backends.backend_agg._BackendAgg'>

    @classmethod
    def show(cls, *, block=None):
        """
        Show all figures.
    
        `show` blocks by calling `mainloop` if *block* is ``True``, or if it
        is ``None`` and we are neither in IPython's ``%pylab`` mode, nor in
        `interactive` mode.
        """
        managers = Gcf.get_all_fig_managers()
        if not managers:
            return
        for manager in managers:
            try:
>               manager.show()  # Emits a warning for non-interactive backend.

../../.virtualenvs/pytest-mpl/lib/python3.8/site-packages/matplotlib/backend_bases.py:3524: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <matplotlib.backend_bases.FigureManagerBase object at 0x7ff5ab60a1c0>

    def show(self):
        """
        For GUI backends, show the figure window and redraw.
        For non-GUI backends, raise an exception, unless running headless (i.e.
        on Linux with an unset DISPLAY); this exception is converted to a
        warning in `.Figure.show`.
        """
        # This should be overridden in GUI backends.
        if cbook._get_running_interactive_framework() != "headless":
>           raise NonGuiException(
                f"Matplotlib is currently using {get_backend()}, which is "
                f"a non-GUI backend, so cannot show the figure.")
E           matplotlib.backend_bases.NonGuiException: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.

../../.virtualenvs/pytest-mpl/lib/python3.8/site-packages/matplotlib/backend_bases.py:2728: NonGuiException

During handling of the above exception, another exception occurred:

heliographic_test_map = <sunpy.map.mapbase.GenericMap object at 0x7ff571b1b0a0>
SunPy Map
---------
Observatory:             
Instrument:              
Detector:               ...    -0.02115 ],
       [-0.030485, -0.030435, -0.030385, ..., -0.03059 , -0.030535,
        -0.030485]], dtype=float32)

    @figure_test
    def test_heliographic_peek(heliographic_test_map):
>       heliographic_test_map.peek()

sunpy/map/tests/test_plotting.py:179: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
sunpy/visualization/visualization.py:22: in show_figure
    plt.show()
../../.virtualenvs/pytest-mpl/lib/python3.8/site-packages/matplotlib/pyplot.py:353: in show
    return _backend_mod.show(*args, **kwargs)
../../.virtualenvs/pytest-mpl/lib/python3.8/site-packages/matplotlib/backend_bases.py:3526: in show
    cbook._warn_external(str(exc))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

message = 'Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.', category = None

    def _warn_external(message, category=None):
        """
        `warnings.warn` wrapper that sets *stacklevel* to "outside Matplotlib".
    
        The original emitter of the warning can be obtained by patching this
        function back to `warnings.warn`, i.e. ``cbook._warn_external =
        warnings.warn`` (or ``functools.partial(warnings.warn, stacklevel=2)``,
        etc.).
        """
        frame = sys._getframe()
        for stacklevel in itertools.count(1):  # lgtm[py/unused-loop-variable]
            if frame is None:
                # when called in embedded context may hit frame is None
                break
            if not re.match(r"\A(matplotlib|mpl_toolkits)(\Z|\.(?!tests\.))",
                            # Work around sphinx-gallery not setting __name__.
                            frame.f_globals.get("__name__", "")):
                break
            frame = frame.f_back
>       warnings.warn(message, category, stacklevel)
E       UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.

../../.virtualenvs/pytest-mpl/lib/python3.8/site-packages/matplotlib/cbook/__init__.py:2108: UserWarning
```

dstansby: 🤦 of course this is exactly what is expected, silly me for not realising. Could you drop a comment in above the warning ignore along the lines of ".peek() calls plt.show(), that raises a warning in the test suite because we use the agg backend"

Cadair:
Ok I think this is now ready to go. The pytest-mpl release it out, and I am pretty sure I have sorted the upload to the figure repo in the circle ci config now as well.

nabobalis (approved):

.circleci/config.yml
```diff
@@ -50,29 +50,40 @@ jobs:
...
       # Add private key for deploying to the figure tests repo
       - add_ssh_keys:
           fingerprints: "7b:8f:c8:1c:13:a3:b4:46:ec:9a:a5:0d:3f:62:69:78"
       - run: ssh-keyscan github.com >> ~/.ssh/known_hosts
-      - run: git clone git@github.com:sunpy/sunpy-figure-tests.git -b sunpy-${CIRCLE_BRANCH} ~/sunpy-figure-tests/
-      - run: cd .tmp/ && find . -maxdepth 1 -name "*-figure*" -type d -exec bash -c 'mkdir ~/sunpy-figure-tests/figures/$1; cp -v "$1"/figure_test_images/*png "/home/circleci/sunpy-figure-tests/figures/$1"' Cp {} \;
       - run: git config --global user.email "sunpy@circleci" && git config --global user.name "SunPy Circle CI"
-      - deploy:
-          command: cd ~/sunpy-figure-tests/; git add . && git commit -m "Update reference figures from ${CIRCLE_BRANCH}" || echo "No changes to reference images to deploy" && git push
+      - run: git clone git@github.com:sunpy/sunpy-figure-tests.git -b sunpy-${CIRCLE_BRANCH} ~/sunpy-figure-tests/
```
dstansby, line 74: Can the depth of this clone be reduced (perhaps to 1)?

.circleci/config.yml
```diff
@@ -50,29 +50,40 @@ jobs:
...
+      # Generate Reference images
+      - run: pip install --user -U tox codecov tox-pypi-filter
+      - run: tox -- --mpl-generate-path=/home/circleci/sunpy-figure-tests/figures/$TOXENV | tee toxlog
+      - run: |
+          hashlib=$(grep "^figure_hashes.*\.json$" toxlog)
```
dstansby, line 79: This seems hacky, is there any other way to extract the hash library name (perhaps from the environment name)?

dstansby (changes requested):
Seems good overall; could you add a removals changelog that states that `sunpy/tests/hash.py` has gone. Not documented and not public API so fine to delete the code, but you never know if someone out there is using it...

.circleci/config.yml
Cadair: Not that I can think of, which is why we ended up here. I don't want to put a manual map of tox name to lib versions in the circle ci config (by hard coding the hash name for each env var). Only by extracting the hash name from the code that generates it in the pytest decorator can we be sure to get it right. I think my regex is pretty restrictive, so it should only match the explicit print line in the tox.ini.

nabobalis (approved):
Assuming codecov is ok.

dstansby:
Since this PR only affects tests, I'm going to remove the 2.1 milestone

Cadair:
@dstansby are you ok with this as it is?

setup.cfg
```diff
@@ -83,18 +84,23 @@ doctest_optionflags = NORMALIZE_WHITESPACE FLOAT_CMP ELLIPSIS
...
     array_compare
+    mpl_image_compare
 remote_data_strict = True
 # Pin junit behaviour; we might want to update this to xunit2 at some point
 junit_family=xunit1
 filterwarnings =
     error
     error::sunpy.util.exceptions.SunpyDeprecationWarning
+    # Do not fail on pytest config issues (i.e. missing plugins) but do show them
+    always::pytest.PytestConfigWarning
```
dstansby, line 103: Does this fix https://github.com/sunpy/sunpy/issues/4612?

dstansby (approved):

setup.cfg
Cadair: I think it fixes some of the issues. I don't *think* it would fix the asdf issue tho.

PR #4610, opened 2020-10-28
Add documentation on coordinates with velocities

### LINKED ISSUE #3188: Properly support velocities in our coordinate frames
Stuff to do:

- [x] Fix the Astropy bug with the HCRS<->ICRS transformations (astropy/astropy#8794)
- [x] Re-implement all of our transformations away from `FunctionTransform`.  ~I'm thinking that HGS<->HGC and HGS<->HCC transformations will be `DynamicMatrixTransform`, and HCC<->HPC and all loopback transformations will be `FunctionTransformWithFiniteDifference`.~  The transformations that I was thinking to implement as `DynamicMatrixTransform` actually have an origin shift if transforming between different observation times.  While they could technically be `AffineTransform`s, it's probably more tractable to make them all `FunctionTransformWithFiniteDifference`. (#3247)
- [x] Fix the missing induced-velocity component for the HCRS<->HGS transformations (#3247)
- [x] Define `default_differential` (and maybe `frame_specific_representation_info`) for each of our frames (#4102)
- [x] Add documentation and guide (#4610)
- [x] Convert HCRS<->HGS transformation `FunctionTransformWithFiniteDifference` to fix the fact that `AffineTransform` can't account for an induced angular velocity (#4613)

ayshih:
Coordinates with velocities are already supported.  This closes #3188 by adding user documentation.

### DIFF
--- a/changelog/4610.doc.rst
+++ b/changelog/4610.doc.rst
@@ -1,0 +1,1 @@
+Added :ref:`sunpy-coordinates-velocities` to explain how to use velocity information in the coordinates framework.
--- a/docs/code_ref/coordinates/index.rst
+++ b/docs/code_ref/coordinates/index.rst
@@ -318,6 +318,7 @@ More Detailed Information
 
    carrington
    rotatedsunframe
+   velocities
    wcs
    other_api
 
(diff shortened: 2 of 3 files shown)


Cadair (approved):
Reads well to me :+1:

nabobalis (approved):

PR #4614, opened 2020-10-29
Remove multimethod

dstansby:
As promised by the deprecation warning, this should be removed for 2.1.


nabobalis (approved):

sunpy/net/attr.py
```diff
@@ -7,10 +7,6 @@
 (a AND b AND c) OR (d AND e).
 
 Walkers are used to traverse the tree that results from combining attributes.
-They are implemented using sunpy.util.multimethod. Multimethods are functions
-that are not assigned to classes but still dispatch by type of one or more
-of their arguments. For more information about multimethods, refer to
-sunpy.util.multimethod.
```
Cadair, line 13: We should probably re-write this to explain they use singledispatch instead.

Cadair (approved):
Praise the holy fire

nabobalis:
I wonder if there is anything else we have not removed.

sunpy/net/attr.py
dstansby: Can someone else who know more about this provide some words (either as a comment here or push to my branch)?

sunpy/net/attr.py
```diff
@@ -7,10 +7,6 @@
 (a AND b AND c) OR (d AND e).
 
 Walkers are used to traverse the tree that results from combining attributes.
```
Cadair, line 9:
```suggestion
Walkers are used to traverse the tree that results from combining attributes.
They are implemented using `functools.singledispatch` modified to dispatch on the second argument to the function.
```

nabobalis:
```FAILED ../../.tox/py38-online/lib/python3.8/site-packages/sunpy/net/helio/tests/test_helio.py::test_get_table_names - AssertionError: assert False
FAILED ../../.tox/py38-online/lib/python3.8/site-packages/sunpy/net/helio/tests/test_helio.py::test_select_table - AttributeError: 'str' object has no attribute 'decode'
``` only.

PR #4616, opened 2020-10-29
Remove deprecated positional arguments to map methods

dstansby:
These were deprecated in 2.0 (an LTS), so good to go in 2.1.

### DIFF
--- a/changelog/4616.breaking.rst
+++ b/changelog/4616.breaking.rst
@@ -1,0 +1,6 @@
+The following `~sunpy.map.Map` methods have had support for specific positional
+arguments removed. They must now be passed as keyword arguments
+(i.e. ``m.method(keyword_arg=value)``).
+
+- :meth:`~sunpy.map.Map.submap`: ``width``, ``height``.
+- :meth:`~sunpy.map.Map.draw_rectangle`: ``width``, ``height``, ``axes``, ``top_right``.
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1445,8 +1445,6 @@ class GenericMap(NDData):
         top_right : `astropy.units.Quantity` or `~astropy.coordinates.SkyCoord`, optional
             The top-right coordinate of the rectangle. If ``top_right`` is
             specified ``width`` and ``height`` must be omitted.
-            Passing this as a positional argument is deprecated, you must pass
-            it as ``top_right=...``.
         width : `astropy.units.Quantity`, optional
             The width of the rectangle. Required if ``top_right`` is omitted.
         height : `astropy.units.Quantity`
@@ -1889,13 +1887,10 @@ class GenericMap(NDData):
             have shape ``(2,)`` to simultaneously define ``top_right``.
         top_right : `~astropy.coordinates.SkyCoord`
             The top-right coordinate of the rectangle.
-            Passing this as a positional argument is deprecated.
         width : `astropy.units.Quantity`, optional
             The width of the rectangle. Required if ``top_right`` is omitted.
-            Passing this as a positional argument is deprecated.
         height : `astropy.units.Quantity`
             The height of the rectangle. Required if ``top_right`` is omitted.
-            Passing this as a positional argument is deprecated.
         axes : `matplotlib.axes.Axes`
             The axes on which to plot the rectangle, defaults to the current
             axes.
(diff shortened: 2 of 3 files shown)


nabobalis (approved):

ayshih:
It's not an "obsolete" decorator.  The decorator rewrites the function signature so that the arguments are not positional only.  See #3677 

This was deprecated in 2.0, so it's okay to remove (I think), but the motivation here is not correct.

ayshih (changes requested):
I'll block this until there's a discussion about whether we want to remove this deprecated behavior.

nabobalis:
> It's not an "obsolete" decorator. The decorator rewrites the function signature so that the arguments are not keyword only. See #3677
> 
> This was deprecated in 2.0, so it's okay to remove (I think), but the motivation here is not correct.

Is it not obsolete in the sense that we don't need it anymore? I still forgot if we remove this in 2.1 or 3.0+ but as far as I can see it seems to be ok for me. @Cadair can be the final judge.

Cadair:
By the deprecation policy this is up for removal, so pulling it seems fine to me unless people think we should move slower in methods of Map?

ayshih:
Sure, but @dstansby appears to mean "obsolete" in the sense that he thought that the decorator wasn't do anything anymore, but it most definitely is.  That is, the decorator makes it so the function can accept positional arguments (and raises a deprecation warning), even though the function signature looks like it can't.  The PR should state a valid reason for a change, even if the change is acceptable by some other policy.

dstansby:
Ah I got confused and thought that the signature as written on the decorated function was the signature, and the decorator just raised a warning if arguments were passed as positional arguments (this is how our decorators work in Matplotlib). Seems like these are fine to remove with appropriate updates to the tests and a changelog though so I'll update this PR as such.

ayshih:
> By the deprecation policy this is up for removal, so pulling it seems fine to me unless people think we should move slower in methods of Map?

Maybe it's just how COVID has affected my sense of time, but it certainly doesn't feel like it's been half a year since this functionality was deprecated.  <shrug>

ayshih (commented):
.

nabobalis:
> Sure, but @dstansby appears to mean "obsolete" in the sense that he thought that the decorator wasn't do anything anymore, but it most definitely is. That is, the decorator makes it so the function can accept positional arguments (and raises a deprecation warning), even though the function signature looks like it can't. The PR should state a valid reason for a change, even if the change is acceptable by some other policy.

Good point we should remove the * in each function.

dstansby:
I think it's been 6 months (since 2.0), but since the deprecation is in an LTS release it's good to remove now (https://github.com/sunpy/sunpy-SEP/blob/master/SEP-0009.md)

dstansby:
> > Sure, but @dstansby appears to mean "obsolete" in the sense that he thought that the decorator wasn't do anything anymore, but it most definitely is. That is, the decorator makes it so the function can accept positional arguments (and raises a deprecation warning), even though the function signature looks like it can't. The PR should state a valid reason for a change, even if the change is acceptable by some other policy.
> 
> Good point we should remove the * in each function.

No we shouldn't, that's what makes the arguments keyword only!

nabobalis:
> > > Sure, but @dstansby appears to mean "obsolete" in the sense that he thought that the decorator wasn't do anything anymore, but it most definitely is. That is, the decorator makes it so the function can accept positional arguments (and raises a deprecation warning), even though the function signature looks like it can't. The PR should state a valid reason for a change, even if the change is acceptable by some other policy.
> > 
> > 
> > Good point we should remove the * in each function.
> 
> No we shouldn't, that's what makes the arguments keyword only!

Remove the other one then.

ayshih (approved):
I don't know what's going on with our test suite, but the code changes look good

changelog/4616.breaking.rst
```diff
@@ -0,0 +1,6 @@
+The following `~sunpy.map.Map` methods have had support for specific positional
+arguments removed. They must now be passed as keyword arguments
+(i.e. ``m.method(keyword_arg=value)``).
+
+- :meth:`~sunpy.map.Map.submap`: ``width``, ``height``.
+- :meth:`~sunpy.map.Map.draw_rectangle`: ``width``, ``height``, ``axes``, ``top_right``.
```
Cadair, line 6: does our changelog rendering actually respect whitespace like this?

Cadair (approved):
Looks good tho the style check is angry

changelog/4616.breaking.rst
nabobalis: Only one way to find out.

Cadair: the line break isn't rendered but the list is.

Issue #3643, opened 2020-01-03
Update to support new features in wcslib 7.1


Cadair:
wcslib 7.1 includes:

>  Support Solar keywords RSUN_REF, DSUN_OBS, CRLN_OBS, CRLT_OBS, HGLN_OBS, and HGLT_OBS by accomodating them within an auxiliary struct, auxprm, within the wcsprm struct.  Now filled by the header parsers, wcspih() and wcsbth() and handled routinely by wcsinit(), wcssub(), wcscompare(), wcsfree(), wcsprt(), wcsset(), and wcshdo().
Requested by Stuart Mumford (SunPy) with input from Bill Thompson.

While we can't rely on wcslib 7.1 yet, as it's not even been merged into astropy (astropy/astropy#9829), we can work out how to support it in parallel with our old hacks.

What this allows us to do is to use the underlying wcslib strcuts for storing the heliographic observer location and rsun reference value. The most important place to support the new functionality is in the code which converts WCSes to and from coordinate frames [here](https://github.com/sunpy/sunpy/blob/master/sunpy/coordinates/wcs_utils.py#L22-L30) and [here](https://github.com/sunpy/sunpy/blob/master/sunpy/coordinates/wcs_utils.py#L56-L64) as this will support people loading compatible headers directly with `astropy.wcs.WCS`.

We also then need to update [`GenericMap.wcs`](https://github.com/sunpy/sunpy/blob/master/sunpy/map/mapbase.py#L265) to "do the right thing" with ensuring that the information is in the struct, and more importantly is consistent with the `GenericMap.observer_coordinate` property, which means removing any inconsistent HG or CR headers and converting the `observer_coordinate` back to HG and putting back in the header at the right place.

Once all of this is done, we need to make sure we test all these options, as well as adding tests for direct parsing of `WCS` objects and headers to `sunpy.Map`, and through the WCS <> coordinate converters. Then finally we need to document this madness somewhere.  

Addressing this would close #2910

Cadair:
It should be noted that 7.1 is now in astropy master and will be included in 4.0.1, so this is now easy to hack on.

dstansby:
To add an example, the goal is to make this code add the observer coordinate properly:

```python
from astropy.wcs.utils import wcs_to_celestial_frame
from astropy.wcs import WCS
from sunpy.util import MetaDict
import sunpy.coordinates.wcs_utils

[...]

Issue #2910, opened 2019-01-25
Support Observer Location and other Properties on Astropy WCS objects


Cadair:
In Astropy 3.1 new APIs were added to `astropy.wcs.WCS` [1] these APIs (specifically for the context of the rest of this issue, `WCS.pixel_to_world`) make use of the existing [`wcs_to_celestial_frame`](https://github.com/astropy/astropy/blob/master/astropy/wcs/utils.py#L178) functionality in Astropy, which SunPy [extends](https://github.com/sunpy/sunpy/blob/master/sunpy/coordinates/wcs_utils.py) to convert from a `WCS` object to a coordinate frame.

These functions work by extracting information from the underlying `WCS.wcs` structure (which is implemented in `wcslib` at the C level). However, this structure does not extract the keys that are specified in [Thompson (2006) [Section 9.1]](https://ui.adsabs.harvard.edu/#abs/2006A&A...449..791T) to determine the observer location (normally `DSUN_OBS`, `HGLN_OBS`, and `HGLT_OBS`). In addition to this it is often required to extract the expected radius of the sun in meters to correctly specify the transform from the helioprojective to the heliocentric coordinate system [2]. Without these extra pieces of metadata any transformation out of the coordinate system described by the `WCS` object will be inaccurate, or flatly wrong.

SunPy currently works around this issue in `sunpy.map` by attaching [extra attributes](https://github.com/sunpy/sunpy/blob/master/sunpy/map/mapbase.py#L274) to the `WCS` returned by the `GenericMap` object, specifically `observer_coordinate` and `rsun`, which [`wcs_to_celestial_frame`](https://github.com/astropy/astropy/blob/master/astropy/wcs/utils.py#L178) uses when constructing the relevant coordinate frame object for the wcs. This is clearly sub-optimal for `WCS` objects created outside of `GenericMap` and is a hack at best.

The best solution to this issue would be to have `wcslib` support the `*_OBS` keywords in the wcs struct and therefore expose them to our `wcs_to_celestial_frame` function. This would require contacting Mark Calabretta (the maintainer of wcslib) and proposing this to him, something I have yet to do.

I do not have a clear picture of what workarounds we could implement in the short term to mitigate this issue without having modifications made in `wcslib`. I am open to suggestions, although I would *really* like to refrain from having to subclass `astropy.wcs.WCS`.

---

[1] http://docs.astropy.org/en/stable/wcs/wcsapi.html#wcsapi
[...]

PR #4620, opened 2020-11-02
Support observer information in WCS

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/3643, fixes https://github.com/sunpy/sunpy/issues/2910.

This is the minimal update needed to get observer coordinates stored in the WCS working properly. There are more general ways to do this, but I went down that rabbit hole and it is not pretty (AIA, I'm looking at you and your many observer coordinate keywords...).

Hopefully this is a good base to start from and make improvements though. I have milestoned as 2.1 as it would be good to get this in 1 release before a LTS to make sure it works.

### DIFF
--- a/changelog/4620.feature.rst
+++ b/changelog/4620.feature.rst
@@ -1,0 +1,3 @@
+sunpy now sets auxillary parameters on `sunpy.map.Map.wcs` using the
+`astropy.wcs.wcs.aux` attribute. This stores observer information, along with
+the reference solar radius if present.
--- a/sunpy/coordinates/wcs_utils.py
+++ b/sunpy/coordinates/wcs_utils.py
@@ -1,6 +1,11 @@
+import warnings
+
+import astropy.units as u
 import astropy.wcs.utils
+from astropy.coordinates import BaseCoordinateFrame
 from astropy.wcs import WCS
 
+from sunpy.util.exceptions import SunpyUserWarning
 from .frames import (
     BaseCoordinateFrame,
     Heliocentric,
@@ -32,17 +37,47 @@ def solar_wcs_frame_mapping(wcs):
 
     dateobs = wcs.wcs.dateobs or None
 
-    # SunPy Map adds 'heliographic_observer' and 'rsun' attributes to the WCS
-    # object. We check for them here, and default to None.
+    # Get observer coordinate from the WCS auxillary information
+    required_attrs = {HeliographicStonyhurst: ['hgln_obs', 'hglt_obs', 'dsun_obs'],
+                      HeliographicCarrington: ['crln_obs', 'hglt_obs', 'dsun_obs']}
+
+    observer = None
+    for frame, attr_names in required_attrs.items():
+        attrs = [getattr(wcs.wcs.aux, attr_name) for attr_name in attr_names]
+        if all([attr is not None for attr in attrs]):
+            observer = frame(attrs[0] * u.deg,
+                             attrs[1] * u.deg,
+                             attrs[2] * u.m,
+                             obstime=dateobs)
+
+    # Get rsun from the WCS auxillary information
+    rsun = wcs.wcs.aux.rsun_ref
+    if rsun is not None:
+        rsun *= u.m
+
+    # These custom attributes were always used in sunpy < 2.1; these warnings
+    # can be converted into errors in sunpy 3.1
     if hasattr(wcs, 'heliographic_observer'):
-        observer = wcs.heliographic_observer
-    else:
-        observer = None
+        warnings.warn('Support for the .heliographic_observer attribute on a WCS is deprecated. '
+                      'Set observer keywords in the FITS header, or directly set the wcs.wcs.aux '
+                      'values instead.',
+                      SunpyUserWarning)
+        if observer is None:
+            observer = wcs.heliographic_observer
+        else:
+            warnings.warn('Observer information present in WCS auxillary information, ignoring '
+                          '.heliographic_observer')
 
     if hasattr(wcs, 'rsun'):
-        rsun = wcs.rsun
-    else:
-        rsun = None
+        warnings.warn('Support for the .rsun attribute on a WCS is deprecated. '
+                      'Set observer keywords in the FITS header, or directly set the wcs.wcs.aux '
+                      'values instead.',
+                      SunpyUserWarning)
+        if rsun is None:
+            rsun = wcs.rsun
+        else:
+            warnings.warn('rsun information present in WCS auxillary information, ignoring '
+                          '.rsun')
 
     # Truncate the ctype to the first four letters
     ctypes = {c[:4] for c in wcs.wcs.ctype}
@@ -58,6 +93,31 @@ def solar_wcs_frame_mapping(wcs):
 
     if {'SOLX', 'SOLY'} <= ctypes:
         return Heliocentric(obstime=dateobs, observer=observer)
+
+
+def _set_wcs_aux_obs_coord(wcs, obs_frame):
+    """
+    Set (in-place) observer coordinate information on a WCS.
+
+    Parameters
+    ----------
+    wcs : astropy.wcs.WCS
+    obs_frame : astropy.coordinates.SkyCoord, astropy.coordinates.CoordinateFrame
+    """
+    # Sometimes obs_coord can be a frame (with data); convert to a SkyCoord to
+    # ensure the .frame attribute is present
+    if not isinstance(obs_frame, BaseCoordinateFrame):
+        obs_frame = obs_frame.frame
+
+    if isinstance(obs_frame, HeliographicStonyhurst):
+        wcs.wcs.aux.hgln_obs = obs_frame.lon.to_value(u.deg)
+    elif isinstance(obs_frame, HeliographicCarrington):
+        wcs.wcs.aux.crln_obs = obs_frame.lon.to_value(u.deg)
+    else:
+        raise ValueError('obs_coord must be in a Stonyhurst or Carrington frame')
+    # These two keywords are the same for Carrington and Stonyhurst
+    wcs.wcs.aux.hglt_obs = obs_frame.lat.to_value(u.deg)
+    wcs.wcs.aux.dsun_obs = obs_frame.radius.to_value(u.m)
 
 
 def solar_frame_to_wcs_mapping(frame, projection='TAN'):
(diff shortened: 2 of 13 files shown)


sunpy/coordinates/wcs_utils.py
```diff
@@ -52,6 +59,29 @@ def solar_wcs_frame_mapping(wcs):
...
+    Set (in-place) observer coordinate information on a WCS.
+
+    Parameters
+    ----------
+    wcs : astropy.wcs.WCS
+    obs_coord : astropy.coordinates.SkyCoord
+    """
+    # Sometimes obs_coord can be a frame (with data); convert to a SkyCoord to
+    # ensure the .frame attribute is present
+    obs_coord = SkyCoord(obs_coord)
```
Cadair, line 73: I suggest you do this the other way around:
```
if not isinstance(obs_coord, CoordinateFrame):
    obs_coord = obs_coord.frame
```
the SkyCoord constructor is very heavy we don't want to call it if we don't have to.

sunpy/coordinates/wcs_utils.py
```diff
@@ -24,17 +26,22 @@ def solar_wcs_frame_mapping(wcs):
 
     dateobs = wcs.wcs.dateobs or None
```
dstansby, line 28: Note to self: raise a warning here if `wcs` has `heliographic_observer` or `rsun` attributes.

sunpy/map/mapbase.py
```diff
@@ -476,12 +476,23 @@ def wcs(self):
...
-        if ({'HPLN', 'HPLT'} <= ctypes or {'SOLX', 'SOLY'} <= ctypes or {'CRLN', 'CRLT'} <= ctypes):
-            w2.heliographic_observer = self.observer_coordinate
+        # Set observer coordinate information
+        #
+        # Clear all the aux information that was set earlier. This is to avoid
+        # issues with maps that store multiple observer coordinate keywords.
+        # Note that we have to create a new WCS as it's not possible to modify
+        # wcs.wcs.aux in place.
+        header = w2.to_header()
+        for kw in ['crln_obs', 'dsun_obs', 'hgln_obs', 'hglt_obs']:
```
Cadair, line 486: you are missing `rsun_ref`: https://github.com/astropy/astropy/blob/master/astropy/wcs/src/wcslib_auxprm_wrap.c#L178-L182

dstansby: That's deliberate, since it's fine to set `rsun_ref` using the first call to `WCS` above; only the observer coordinate keywords need to be cleared here to avoid duplicate or misleading information that doesn't come from `m.observer_coordinate`.

sunpy/coordinates/wcs_utils.py
```diff
@@ -32,17 +37,31 @@ def solar_wcs_frame_mapping(wcs):
...
-    # SunPy Map adds 'heliographic_observer' and 'rsun' attributes to the WCS
-    # object. We check for them here, and default to None.
+    # These custom attributes were used in sunpy < 2.1; these warnings can
+    # be removed in sunpy 3.1
     if hasattr(wcs, 'heliographic_observer'):
-        observer = wcs.heliographic_observer
-    else:
-        observer = None
-
+        warnings.warn('sunpy no longer uses the .heliographic_observer attribute.',
```
Cadair, line 43: If this isn't going to work, should this be an error? Also, it should point people at aux, so they know how to fix it.

sunpy/coordinates/wcs_utils.py
```diff
@@ -60,6 +79,31 @@ def solar_wcs_frame_mapping(wcs):
...
+    # ensure the .frame attribute is present
+    if not isinstance(obs_frame, BaseCoordinateFrame):
+        obs_frame = obs_frame.frame
+
+    if isinstance(obs_frame, HeliographicStonyhurst):
+        wcs.wcs.aux.hgln_obs = obs_frame.lon.to_value(u.deg)
+    elif isinstance(obs_frame, HeliographicCarrington):
+        wcs.wcs.aux.crln_obs = obs_frame.lon.to_value(u.deg)
+    else:
+        raise ValueError('obs_coord must be in a Stonyhurst or Carrington frame')
```
Cadair, line 101: What's the reasoning for not converting here but doing so in mapbase?

sunpy/map/mapbase.py
```diff
@@ -476,12 +476,23 @@ def wcs(self):
...
-        ctypes = {c[:4] for c in w2.wcs.ctype}
-        # Check that the ctypes contains one of these three pairs of axes.
-        if ({'HPLN', 'HPLT'} <= ctypes or {'SOLX', 'SOLY'} <= ctypes or {'CRLN', 'CRLT'} <= ctypes):
-            w2.heliographic_observer = self.observer_coordinate
+        # Set observer coordinate information
+        #
+        # Clear all the aux information that was set earlier. This is to avoid
+        # issues with maps that store multiple observer coordinate keywords.
+        # Note that we have to create a new WCS as it's not possible to modify
+        # wcs.wcs.aux in place.
```
Cadair, line 484: We should open an astropy issue and link to it here, I don't see any fundamental reason why this shouldn't be supported.

sunpy/map/mapbase.py
```diff
@@ -476,12 +476,23 @@ def wcs(self):
...
+        header = w2.to_header()
+        for kw in ['crln_obs', 'dsun_obs', 'hgln_obs', 'hglt_obs']:
+            header.pop(kw, None)
```
Cadair, line 487: is this case sensitive? I don't think so right?

sunpy/tests/figure_hashes_mpl_332_ft_261_astropy_403.json
```diff
@@ -22,7 +22,7 @@
     "sunpy.map.tests.test_plotting.test_plot_masked_aia171_nowcsaxes": "69e28235fa42379a94f7da6cc39b375f361d1e26dc0f94e6b8c5f05894cc64c7",
     "sunpy.map.tests.test_plotting.test_plot_masked_aia171_superpixel": "154586d5635f895b5f9170353f70f427e23ab0be4ccabf743e668864653f54e3",
     "sunpy.map.tests.test_plotting.test_plot_masked_aia171_superpixel_nowcsaxes": "ec4cfe2687ea0a1f62f1a80f07e69ef5a89ef5aa2dbebc58599fb16b98ebafcb",
-    "sunpy.map.tests.test_plotting.test_plot_rotated_aia171": "878fedb8b890018434ca66aa09b8aacc5c2bd6a39b4cbefed36121998aca2bcc",
+    "sunpy.map.tests.test_plotting.test_plot_rotated_aia171": "06ed44d872a6e1e5ec2dd6a522ee81f68a2c8420dea1bc7491bc64d9e3d3e147",
```
Cadair, line 25: ooi, what changed?

changelog/4620.feature.rst
```diff
@@ -0,0 +1,5 @@
+sunpy now sets auxillary parameters using the `astropy.wcs.wcs.aux` attribute.
+This stores observer information, along with the reference solar radius if present.
+
+This means the sunpy specific attributes ``.heliographic_observer`` and ``.rsun``
+are no longer set on the `~astropy.wcs.WCS` returned by `sunpy.map.Map.wcs`.
```
Cadair, line 5: duplication with the other entry? Do we need this here?

Cadair (commented):
Thanks for working on this @dstansby it looks great. A few minor comments.

sunpy/tests/figure_hashes_mpl_332_ft_261_astropy_403.json
dstansby: A few pixels, nothing major. I think due to the precision at which the auxillary information is stored in the FITS header.

changelog/4620.feature.rst
dstansby: This is a feature entry, and the other one is a behaviour change entry. I think we should have both (so the changes show up in both sections), but happy to wordsmith either message.

sunpy/coordinates/wcs_utils.py
dstansby: Dunno, having these attributes isn't going to do any harm, they just won't be used. I was thinking this was sort of a deprecation period, after which we can just remove the warnings.

sunpy/coordinates/wcs_utils.py
dstansby: This provides a more stringent check on obs_frame, which doesn't let anything through that isn't stonyhurst/carrington, which is then the behaviour of `solar_frame_to_wcs_mapping`, which I think is desired (to error instead of transforming).

Then, if elsewhere (e.g. `Map`) we want to be more lenient, we can always transform to `stonyhurst` before setting this metadata to let any coordinate system through.

That's my thinking at least, open to other options.

changelog/4620.feature.rst
Cadair: I get having both entries, I am just not sure they need the same info in both. We could have implemented this feature while still keeping these attributes, so not saying this in the feature entry seems ok to me?

sunpy/coordinates/wcs_utils.py
Cadair: I suppose that makes some sense for this helper, if we consider this a straight up wrapper of the WCS functionality. On the other hand, the point of our coordinate infra is that the transforms shouldn't ever really need to be special cased, they will either work or they wont. So I am not sure what the reasoning would be to not want the transform done if you are calling this helper function :thinking:

dstansby: Maybe @ayshih can have a casting vote - I'm unsure what the best option is.

sunpy/coordinates/wcs_utils.py
Cadair: The normal flow for a deprecation is that the thing still works but shouts at you, and then it is removed. Here we have it raise a warning and then not work. I think we should either keep it working or make it raise an error, as you need to edit your code to get the expected behaviour.

sunpy/map/mapbase.py
dstansby: Looking in the code for `astropy.io.fits.Header` it seems like it is, though it's not clear from the docs. Opened https://github.com/astropy/astropy/issues/11031 but seems safe to take this approach for now.

sunpy/coordinates/wcs_utils.py
dstansby: I've put back support for using the old method, with deprecation warnings. Still need to add tests for this and update the changelog.

sunpy/map/mapbase.py
dstansby: Confirmed with tests in https://github.com/astropy/astropy/pull/11039 that this is fine.

sunpy/coordinates/wcs_utils.py
dstansby: Updated changelog and should be tested too.

sunpy/coordinates/wcs_utils.py
dstansby: Since this is a private function, and for the sake of getting this shipped, I would suggest leaving this as is and we can always revisit?

sunpy/map/mapbase.py
dstansby: Opened https://github.com/astropy/astropy/issues/11041

sunpy/coordinates/wcs_utils.py
Cadair: Sounds good.

Cadair (approved):
:shipit:

nabobalis (approved):

Issue #4000, opened 2020-04-08
Changes we can make now we depend on Astropy 4.0+


Cadair:
This is a list of things which we can change once we depend on Astropy 4.0+. This is a follow on from #3154

## Visualization

- [x] (fixed in https://github.com/sunpy/sunpy/pull/4621) Line plotting in `ArrayAnimatorWCS` is conditional on Astropy 4.0, if it's our minimum requirement we can drop all the conditional stuff and expect it to be there.

## Coordinates

- [x] (Fixed in #4609) Remove explicit transformations of coordinates prior to calling `.plot_coord()` (e.g., [here](https://docs.sunpy.org/en/latest/generated/gallery/units_and_coordinates/STEREO_SECCHI_starfield.html) and [here](https://docs.sunpy.org/en/latest/generated/gallery/saving_and_loading_data/coordinates_in_asdf.html)) because the transformation will be performed automatically.
- [x] (Fixed in #4577) Remove custom `size` property of `SunPyBaseCoordinateFrame` because that workaround is no longer needed (see astropy/astropy#10295).
- [x] (Fixed in #4584) Change `apparent_longitude()` and `apparent_latitude()` to use `GeocentricTrueEcliptic` (no change in accuracy) (requires Astropy 4.0.2, see astropy/astropy#10129)

nikita-0209:
Since Astropy 4.0 has been released, can I begin working on this issue? I am a novice.

nikita-0209:
Oh alright. I'll look for another issue

mridullpandey:
hi, @Cadair  may I please take this up.

nabobalis:
@mridullpandey we still do not depend on astropy 4.0 only, I would suggest another issue.

mridullpandey:
thanks

Cadair:
I still think it's too early to do this, but following the pattern of incrementing our astropy dep by one each release, we can make these changes on master in a month or so.

kathivashkiv:
Are changes listed above relevant for now? (After #3918 discussion)

nabobalis:
Yes.

kathivashkiv:
About one of them, "delete size property": 
1) there is one mention of this property along with SunPyBaseCoordinateFrame, in metaframe.py:
https://github.com/sunpy/sunpy/blob/master/sunpy/coordinates/metaframes.py#L184
Should this check be deleted after size property removal, or should it stay there for another reason (different than described)?
[...]

PR #4621, opened 2020-11-03
Update ArrayAnimatorWCS class

kathivashkiv:
Checks related to Astropy ver 4.0 have been removed
(#4000 issue)

### DIFF
--- a/sunpy/visualization/animator/wcs.py
+++ b/sunpy/visualization/animator/wcs.py
@@ -102,10 +102,8 @@ class ArrayAnimatorWCS(ArrayAnimator):
         """
         Read first the axes names property of the wcs and fall back to physical types.
         """
-        # world_axis_names was only added to the APE 14 API in 4.0, so do this for backwards compatibility.
-        world_axis_names = self.wcs.world_axis_names
         # Return the name if it is set, or the physical type if it is not.
-        return [l or t for l, t in zip(world_axis_names, self.wcs.world_axis_physical_types)]
+        return [l or t for l, t in zip(self.wcs.world_axis_names, self.wcs.world_axis_physical_types)]
 
     def _compute_slider_labels_from_wcs(self, slices):
         """
--- a/sunpy/visualization/animator/wcs.py
+++ b/sunpy/visualization/animator/wcs.py
@@ -80,13 +80,6 @@ class ArrayAnimatorWCS(ArrayAnimator):
             image_axes.append(slices[::-1].index("y"))
             self.plot_dimensionality = 2
 
-        if self.plot_dimensionality == 1:
-            try:
-                from astropy.visualization.wcsaxes.frame import RectangularFrame1D  # NOQA
-            except ImportError as e:
-                raise ImportError(
-                    "Astropy 4.0 must be installed to do line plotting with WCSAxes.") from e
-
         self.naxis = data.ndim
         self.num_sliders = self.naxis - self.plot_dimensionality
         self.slices_wcsaxes = list(slices)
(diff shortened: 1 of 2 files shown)


kathivashkiv:
@Cadair @nabobalis please have a look

sunpy/visualization/animator/wcs.py
```diff
@@ -110,7 +103,10 @@ def _get_wcs_labels(self):
         Read first the axes names property of the wcs and fall back to physical types.
         """
         # world_axis_names was only added to the APE 14 API in 4.0, so do this for backwards compatibility.
-        world_axis_names = getattr(self.wcs, "world_axis_names", [''] * self.wcs.world_n_dim)
+        try:
+            world_axis_names = self.wcs.world_axis_names
+        except AttributeError:
+            world_axis_names = [''] * self.wcs.world_n_dim
```
nabobalis, line 109: I am not sure we need this.

sunpy/visualization/animator/wcs.py
```diff
@@ -80,13 +80,6 @@ def __init__(self, data, wcs, slices, coord_params=None, ylim='dynamic', ylabel=
             image_axes.append(slices[::-1].index("y"))
             self.plot_dimensionality = 2
 
-        if self.plot_dimensionality == 1:
-            try:
-                from astropy.visualization.wcsaxes.frame import RectangularFrame1D  # NOQA
```
nabobalis, line 85: Do we still not need this import for something?

kathivashkiv: In `wcs.py` it is the only mention of it

nabobalis: Must have just been a check then.

sunpy/visualization/animator/wcs.py
kathivashkiv: @Cadair said that this method can be rewritten not to use `getattr`, but I could have misunderstood the way to do it

nabobalis, line 109:
```suggestion
        world_axis_names = self.wcs.world_axis_names
```
I think this is all we need.

Cadair: I thought we still needed to generate the list of strings but [we don't](https://github.com/astropy/astropy/blob/master/astropy/wcs/wcsapi/low_level_api.py#L301) as that's the default in the base class. So @nabobalis's suggestion is correct.

sunpy/visualization/animator/wcs.py
Cadair: yeah this was how I was checking if you were running an astropy version which supported 1D or not.

sunpy/visualization/animator/wcs.py
```diff
@@ -110,7 +103,7 @@ def _get_wcs_labels(self):
         Read first the axes names property of the wcs and fall back to physical types.
         """
         # world_axis_names was only added to the APE 14 API in 4.0, so do this for backwards compatibility.
-        world_axis_names = getattr(self.wcs, "world_axis_names", [''] * self.wcs.world_n_dim)
+        world_axis_names = self.wcs.world_axis_names
         # Return the name if it is set, or the physical type if it is not.
         return [l or t for l, t in zip(world_axis_names, self.wcs.world_axis_physical_types)]
```
Cadair, line 108: Might as well drop the intermediate variable as well I think:
```suggestion
        world_axis_names = self.wcs.world_axis_names
        # Return the name if it is set, or the physical type if it is not.
        return [l or t for l, t in zip(self.wcs.world_axis_names, self.wcs.world_axis_physical_types)]
```

Cadair (approved):

sunpy/visualization/animator/wcs.py
kathivashkiv: Then delete the first comment too?

kathivashkiv: `world_axis_names was only added to the APE 14 API in 4.0, so do this for backwards compatibility.`

dstansby: Yep, deleting the comment seems like the right thing to do

dstansby (approved):
Thanks for the PR 👍

PR #4601, opened 2020-10-24
fixed wsdl parser bug

dstansby:
Pulled out of https://github.com/sunpy/sunpy/pull/4358/ to
- Make that PR more manageable
- Allow an easier backport

Fixes https://github.com/sunpy/sunpy/pull/4358

PR #4625, opened 2020-11-11
Revert "fixed wsdl parser bug"

dstansby:
Reverts sunpy/sunpy#4601


dstansby:
🤷  this is failing the same way, I think this bug is independent of the change and can be fixed in a new PR.

Issue #4564, opened 2020-10-12
improve setuptools_scm dev version handling


Cadair:
copy this: https://github.com/astropy/astropy/pull/10774

PR #4629, opened 2020-11-11
dev scm update

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/4564

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -71,13 +71,16 @@ docs =
   jplephem
 
 
+[options.packages.find]
+exclude = sunpy._dev
+
 [options.entry_points]
 asdf_extensions =
   sunpy = sunpy.io.special.asdf.extension:SunpyExtension
 
 [tool:pytest]
 testpaths = "sunpy" "docs"
-norecursedirs = ".tox" "build" "docs[\/]_build" "docs[\/]generated" "*.egg-info" "examples" "sunpy[/\]cm" ".jupyter" ".history" "tools" "sunpy[\/]extern"
+norecursedirs = ".tox" "build" "docs[\/]_build" "docs[\/]generated" "*.egg-info" "examples" "sunpy[/\]_dev" ".jupyter" ".history" "tools" "sunpy[\/]extern"
 doctest_plus = enabled
 doctest_optionflags = NORMALIZE_WHITESPACE FLOAT_CMP ELLIPSIS
 addopts = --doctest-rst -m "not figure" --doctest-ignore-import-errors
--- a/setup.cfg
+++ b/setup.cfg
@@ -70,7 +70,6 @@ docs =
   astroquery
   jplephem
 
-
 [options.packages.find]
 exclude = sunpy._dev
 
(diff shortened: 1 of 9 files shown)


sunpy/__init__.py
```diff
@@ -13,18 +13,13 @@
 
 from sunpy.tests.runner import SunPyTestRunner
 from sunpy.util import system_info
-from sunpy.util.config import load_config, print_config
```
Cadair, line 16: why? This is a thing we provide in the base namespace so people can see what config they are running.

sunpy/__init__.py
```diff
@@ -54,17 +49,10 @@ def _get_bibtex():
 
 
 __citation__ = __bibtex__ = _get_bibtex()
-
```
Cadair, line 57: I feel like removing *all* this whitespace makes it less readable, single space between different bits would be clearer to me, but whatever.

setup.cfg
```diff
@@ -71,13 +71,16 @@ docs =
   jplephem
```
Cadair, line 73:
```suggestion
```

Cadair (approved):

sunpy/__init__.py
nabobalis: It isn't in the all at least.

sunpy/__init__.py
nabobalis: Wrong.

setup.cfg
nabobalis: Whitespace is important for readability.

sunpy/__init__.py
Cadair: hmm, fair, but it will still be there if you do:
```python
import sunpy
sunpy.print_config
```
so removing it is still a change. I think we should keep it, and maybe add it to the `__all__`.

sunpy/__init__.py
```diff
@@ -13,18 +13,13 @@
 
 from sunpy.tests.runner import SunPyTestRunner
 from sunpy.util import system_info
-from sunpy.util.config import load_config, print_config
+from sunpy.util.config import load_config
```
Cadair, line 16:
```suggestion
from sunpy.util.config import load_config, print_config
```

sunpy/__init__.py
nabobalis: Sure, we can't hide it but I feel like it should either be removed or in the all.

sunpy/__init__.py
```diff
@@ -54,17 +49,10 @@ def _get_bibtex():
...
 # Load user configuration
 config = load_config()
-
-
 # Use the root logger as a dummy log before initializing Astropy's logger
 log = logging.getLogger()
-
 log = _init_log(config=config)
-
 __all__ = ['config', 'self_test', 'system_info']
```
Cadair, line 58:
```suggestion

__all__ = ['config', 'self_test', 'system_info', 'print_config']
```

nabobalis: Fixed.

sunpy/__init__.py
nabobalis: Fixed.

sunpy/__init__.py
nabobalis: Fixed.

sunpy/__init__.py
nabobalis: Fixed.

setup.cfg
nabobalis: Fixed.

dstansby:
Remote test failures look unrelated

PR #4643, opened 2020-11-14
updated return types for the HECClient

### LINKED ISSUE #4631: Update Helio Code to expect str not bytes
Astropy 4.1 changed the output of [`parse_single_table`](https://keflavich-astropy.readthedocs.io/en/latest/api/astropy.io.votable.parse_single_table.html#astropy.io.votable.parse_single_table) to return `str` not `bytes`. https://github.com/sunpy/sunpy/pull/4628 modified our code to convert it back to bytes, but this was a short term solution for the 2.0 branch, we should be forwards-compatible and change the code to always expect `str` and decode the bytes to str after the function call (and then be able to remove that code when we depend on astropy > 4.1.

nabobalis:
Fixes #4631

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -134,6 +134,8 @@ filterwarnings =
     ignore:Distutils was imported before Setuptools
     # Figure tests use agg backend so show doesn't work but we use it in peek
     ignore:Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
+    # We raise this for our HECClient to avoid HTTPS issues with non https urls.
+    ignore:Unverified HTTPS request
 
 [pycodestyle]
 max_line_length = 100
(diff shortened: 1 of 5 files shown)


setup.cfg
```diff
@@ -134,6 +134,8 @@ filterwarnings =
     ignore:Distutils was imported before Setuptools
     # Figure tests use agg backend so show doesn't work but we use it in peek
     ignore:Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
+    # We raise this for our HECClient to avoid HTTPS issues with non https urls.
+    ignore:Unverified HTTPS request
```
nabobalis, line 138: This only came up when I ran the doctests locally. I have not seen this on the CI.

Cadair (approved):

Cadair (approved):

PR #4646, opened 2020-11-15
updated extern files

nabobalis:


setup.cfg
```diff
@@ -39,7 +39,6 @@ install_requires =
   pandas>=0.23.0
   astropy>=4.0
   parfive[ftp]>=1.1.0
-  importlib_metadata;python_version<"3.8"
```
nabobalis, line 42: Not sure we are using this for anything anymore?

kakirastern: The package `importlib_metadata` is intended to be ported to Python 3.8... But then something is off (or fishy) if that is the case.

Cadair: I can't remember what we were using this for in the first place?

nabobalis: We were using to to check the python version according to the changelog.

kakirastern: I see. Thanks @nabobalis!

kakirastern (approved):

Cadair (approved):

Issue #4651, opened 2020-11-16
map.contour should make use of map.units


dstansby:
We should check that the level given to the new `contour` method has the same units as the map data. Currently working on this, but tagging as 2.1 release critical since `contour` is new for 2.1.

PR #4652, opened 2020-11-16
Force use of units when contouring

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/4651. This forces the use of a quantity to contour when a map has it's unit attribute set.

### DIFF
--- a/examples/map/map_contouring.py
+++ b/examples/map/map_contouring.py
@@ -18,11 +18,17 @@ from sunpy.data.sample import AIA_193_IMAGE
 aiamap = sunpy.map.Map(AIA_193_IMAGE)
 
 ###############################################################################
-# Now find the contours
+# In find a set of contours, we have to provide the level to contour in the
+# same units as the map data. To find out the units we can inspect
+# `sunpy.map.Map.unit`.
+print(aiamap.unit)
+###############################################################################
+# We can see that the units of this map are ``ct``, or counts. We can now
+# chose a contour level, and use the contour method to extract the contours.
 contours = aiamap.contour(50000 * u.ct)
 
 ##############################################################################
-# Finally, plot the map and add the contours
+# Finally, we can plot the map, and add each of the contours in turn.
 plt.figure()
 ax = plt.subplot(projection=aiamap)
 aiamap.plot()
--- a/examples/map/map_contouring.py
+++ b/examples/map/map_contouring.py
@@ -6,8 +6,9 @@ Finding contours of a map
 
 This example shows how to find and plot contours on a map.
 """
-
 import matplotlib.pyplot as plt
+
+import astropy.units as u
 
 import sunpy.map
 from sunpy.data.sample import AIA_193_IMAGE
@@ -18,7 +19,7 @@ aiamap = sunpy.map.Map(AIA_193_IMAGE)
 
 ###############################################################################
 # Now find the contours
-contours = aiamap.contour(50000)
+contours = aiamap.contour(50000 * u.ct)
 
 ##############################################################################
 # Finally, plot the map and add the contours
--- a/examples/map/map_contouring.py
+++ b/examples/map/map_contouring.py
@@ -24,7 +24,8 @@ aiamap = sunpy.map.Map(AIA_193_IMAGE)
 print(aiamap.unit)
 ###############################################################################
 # We can see that the units of this map are ``ct``, or counts. We can now
-# chose a contour level, and use the contour method to extract the contours.
+# chose a contour level, and use the :meth:`~sunpy.map.GenericMap.contour`
+# method to extract the contours.
 contours = aiamap.contour(50000 * u.ct)
 
 ##############################################################################
--- a/examples/map/map_contouring.py
+++ b/examples/map/map_contouring.py
@@ -18,7 +18,7 @@ from sunpy.data.sample import AIA_193_IMAGE
 aiamap = sunpy.map.Map(AIA_193_IMAGE)
 
 ###############################################################################
-# In find a set of contours, we have to provide the level to contour in the
+# In finding a set of contours, we have to provide the level to contour in the
 # same units as the map data. To find out the units we can inspect
 # `sunpy.map.GenericMap.unit`.
 print(aiamap.unit)
--- a/examples/map/map_contouring.py
+++ b/examples/map/map_contouring.py
@@ -20,8 +20,9 @@ aiamap = sunpy.map.Map(AIA_193_IMAGE)
 ###############################################################################
 # In find a set of contours, we have to provide the level to contour in the
 # same units as the map data. To find out the units we can inspect
-# `sunpy.map.Map.unit`.
+# `sunpy.map.GenericMap.unit`.
 print(aiamap.unit)
+
 ###############################################################################
 # We can see that the units of this map are ``ct``, or counts. We can now
 # chose a contour level, and use the :meth:`~sunpy.map.GenericMap.contour`
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -2203,7 +2203,7 @@ class GenericMap(NDData):
         if self.unit is not None:
             try:
                 level = level.to_value(self.unit)
-            # Catch cases where level is a float or can't be converted
+            # Catch cases where level can't be converted or isn't a Quantity
             except (AttributeError, u.UnitConversionError) as e:
                 raise ValueError(f'level must be an astropy quantity convertible to {self.unit}') from e
 
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -2181,6 +2181,20 @@ class GenericMap(NDData):
         contours: list of (n,2) `~astropy.coordinates.SkyCoord`
             Coordinates of each contour.
 
+        Examples
+        --------
+        >>> import astropy.units as u
+        >>> from astropy.coordinates import SkyCoord
+        >>> import sunpy.map
+        >>> import sunpy.data.sample  # doctest: +REMOTE_DATA
+        >>> aia = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)  # doctest: +REMOTE_DATA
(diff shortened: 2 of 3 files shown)


examples/map/map_contouring.py
```diff
@@ -18,7 +19,7 @@
 
 ###############################################################################
 # Now find the contours
-contours = aiamap.contour(50000)
+contours = aiamap.contour(50000 * u.ct)
```
hayesla, line 22: this example should also explicitly show users how to find the units of the data - i.e. 


'we can inspect the units of the data by doing:'
``` print(aiamap.units)```

File: sunpy/map/mapbase.py [hunk 2]
```diff
@@ -2177,8 +2177,10 @@ def contour(self, level, **kwargs):
 
         Parameters
         ----------
-        level : float
-            Value along which to find contours in the array.
+        level : float, astropy.units.Quantity
```
hayesla, line 2180: just giving this functionality - it would be nice if the users could also pass a percentage level of data, this would then be consistent with `draw_contours`

sunpy/map/mapbase.py
```diff
@@ -2177,8 +2177,10 @@ def contour(self, level, **kwargs):
...
+            Value along which to find contours in the array. If the map unit attribute
+            is not `None`, this must be a `~astropy.units.Quantity` with units
+            equivalent to the map data units.
         kwargs :
             Additional keyword arguments are passed to `skimage.measure.find_contours`.
```
hayesla, line 2186: can you add simple example here in docs?

hayesla (changes requested):

hayesla:
also a  `@u.quantity_input` decorator I think

sunpy/map/mapbase.py
dstansby: Percentage relative to what? I am reluctant to do this, given the different ways one can take a percentage of the data, and also given that it's pretty easy for a user to explicitly calculate a level as a percentage.

hayesla: percentages of the maximum values of the maps - yeah I agree its pretty easy for user. I guess it was as when I originally played with this, I assumed the levels would be in percentages as `draw_contours` is

ayshih: Heh, @dstansby...  Are we going to argue about this again? =P  (xref #4206)

dstansby:
> also a `@u.quantity_input` decorator I think

We still have to be able to pass floats through for when `unit == None`, so I don't think this decorator can be added.

sunpy/map/mapbase.py
dstansby: I would say this is a feature request to go on top of the original PR (https://github.com/sunpy/sunpy/pull/3909), instead of in this PR.

ayshih: I concur

examples/map/map_contouring.py
```diff
@@ -17,11 +18,17 @@
 aiamap = sunpy.map.Map(AIA_193_IMAGE)
 
 ###############################################################################
-# Now find the contours
-contours = aiamap.contour(50000)
+# In find a set of contours, we have to provide the level to contour in the
+# same units as the map data. To find out the units we can inspect
+# `sunpy.map.Map.unit`.
+print(aiamap.unit)
+###############################################################################
```
nabobalis, line 25:
```suggestion

###############################################################################
```

examples/map/map_contouring.py
```diff
@@ -17,11 +18,17 @@
 aiamap = sunpy.map.Map(AIA_193_IMAGE)
 
 ###############################################################################
-# Now find the contours
-contours = aiamap.contour(50000)
+# In find a set of contours, we have to provide the level to contour in the
+# same units as the map data. To find out the units we can inspect
+# `sunpy.map.Map.unit`.
```
ayshih, line 23: This won't actually link to anything (as opposed to `sunpy.map.GenericMap.unit`)

examples/map/map_contouring.py
```diff
@@ -17,11 +18,17 @@
...
+print(aiamap.unit)
+###############################################################################
+# We can see that the units of this map are ``ct``, or counts. We can now
+# chose a contour level, and use the contour method to extract the contours.
```
ayshih, line 27:
```suggestion
# chose a contour level, and use the :meth:`~sunpy.map.GenericMap.contour`
# method to extract the contours.
```

sunpy/map/mapbase.py
```diff
@@ -2187,11 +2189,33 @@ def contour(self, level, **kwargs):
         contours: list of (n,2) `~astropy.coordinates.SkyCoord`
             Coordinates of each contour.
 
+        Examples
+        --------
+        >>> import astropy.units as u
+        >>> from astropy.coordinates import SkyCoord
```
ayshih, line 2195:
```suggestion
```

kakirastern:
Please add a changelog entry to complete the PR.

sunpy/map/mapbase.py
```diff
@@ -2187,11 +2189,33 @@ def contour(self, level, **kwargs):
...
         See also
         --------
         `skimage.measure.find_contours`
         """
         from skimage import measure
+
+        if self.unit is not None:
+            try:
+                level = level.to_value(self.unit)
+            # Catch cases where level is a float or can't be converted
```
ayshih, line 2215: What's the issue if `level` is a float?  Should that part of the comment be removed?

dstansby: If `level` is a float but `self.unit != None` then this errors, since we want the user to give `level` as a `Quantity`. I'll improve the comment here to make it clearer.

ayshih: It does?  I just tried it (`aia.contour(50000.1*u.ct)`) and no error...  Am I misunderstanding something?

ayshih: I guess it depends on the type of the data array?

dstansby: `50000.1 * u.ct` should be a `Quantity`, not a `float` - I'm interpreting `float` in the strict sense of `isinstance(num, float) == True`.

ayshih: Oh!  So you really mean a non-`Quantity`, whether it's a `float`, `int`, or whatever.

dstansby: Yep, I have updated the comment to "not a Quantity" to be a bit clearer

dstansby:
I don't think this needs a changelog (since it's just making `countour` behave like it should have done from the start, and `contour` is new), but happy to add one if desired.

kakirastern:
Just to please the CI/CD checks 😉

wtbarnes:
> Just to please the CI/CD checks 😉

This is what the "No change log needed" label is for

kakirastern:
> > Just to please the CI/CD checks 😉
> 
> This is what the "No change log needed" label is for

Awesome, was wondering about this myself. Thanks for the tip!

examples/map/map_contouring.py
```diff
@@ -17,11 +18,19 @@
 aiamap = sunpy.map.Map(AIA_193_IMAGE)
 
 ###############################################################################
-# Now find the contours
-contours = aiamap.contour(50000)
+# In find a set of contours, we have to provide the level to contour in the
```
wtbarnes, line 21:
```suggestion
# In finding a set of contours, we have to provide the level to contour in the
```

I'm assuming this is what you meant?

sunpy/map/mapbase.py
```diff
@@ -2191,11 +2193,32 @@ def contour(self, level, **kwargs):
...
         `skimage.measure.find_contours`
         """
         from skimage import measure
+
+        if self.unit is not None:
+            try:
+                level = level.to_value(self.unit)
+            # Catch cases where level can't be converted or isn't a Quantity
+            except (AttributeError, u.UnitConversionError) as e:
+                raise ValueError(f'level must be an astropy quantity convertible to {self.unit}') from e
```
Cadair, line 2220:
```suggestion
                raise u.UnitsError(f'level must be an astropy quantity convertible to {self.unit}') from e
```
maybe?

examples/map/map_contouring.py
```diff
@@ -17,11 +18,19 @@
...
+# same units as the map data. To find out the units we can inspect
+# `sunpy.map.GenericMap.unit`.
+print(aiamap.unit)
+
+###############################################################################
+# We can see that the units of this map are ``ct``, or counts. We can now
+# chose a contour level, and use the :meth:`~sunpy.map.GenericMap.contour`
+# method to extract the contours.
+contours = aiamap.contour(50000 * u.ct)
```
wtbarnes, line 30:
```suggestion
contours = aiamap.contour(50000 * aiamap.unit)
```

I understand if you'd rather be explicit that this is in units of `ct`.

Cadair (approved):
Looks fine, maybe raise a more unit specific error?

wtbarnes (approved):

examples/map/map_contouring.py
dstansby: Yeah, I think it's worth keeping it explicit here.

examples/map/map_contouring.py
```diff
@@ -17,11 +18,19 @@
 aiamap = sunpy.map.Map(AIA_193_IMAGE)
 
 ###############################################################################
-# Now find the contours
-contours = aiamap.contour(50000)
+# In finding a set of contours, we have to provide the level to contour in the
```
hayesla, line 21: do you think it's worth explicitly mentioning here that the level here is different to level given for `draw_contours`? I feel like users may be confused slightly, and just to explain the difference is using the input `level`?

hayesla (approved):

examples/map/map_contouring.py
kakirastern: Agreed...

nabobalis:
I did a quick rebase and test fix.

examples/map/map_contouring.py
nabobalis: Do we need both methods?

PR #4662, opened 2020-11-19
created more extra deps

### LINKED ISSUE #4655: Move Map and Timeseries and their dependencies to extra requirements
Given the discussion in https://github.com/sunpy/sunpy/pull/4592#discussion_r525395690 about adding support for netcdf4 files to Timeseries, I think now is a good (ish) time to reconsider the standing policy that doing `pip install sunpy` installs all the dependencies for map and timeseries to be fully operational.

The current core deps are:
  numpy>=1.16.0
  scipy>=1.2.0
  matplotlib>=2.2.2
  pandas>=0.23.0
  astropy>=4.0
  parfive[ftp]>=1.1.0

There are a few variants of this plan, but I propose that the best option (other than the status quo) is to add a `timeseries` and a `map` extra. This would leave core with numpy, astropy and parfive. 

matplotlib is an interesting one, I think we could leave it optional on both map and timeseries, but then `cm` also wouldn't work, so do we add an extra for that as well? (Maybe we need to leave mpl in core for 2.1 and work these things out??)

Of the rest scipy is Map (and image), pandas is timeseries.

We would need someone to do a good audit of the code make sure we aren't rushing this and miss something though.

nabobalis:
Fixes #4655


docs/guide/installation/advanced.rst
```diff
@@ -36,14 +36,18 @@ It is however *strongly* recommended to have all the dependencies installed (wit
 
 There are multiple options depending on how many optional dependencies you want to install:
 
-To install sunpy with ``pip`` including optional dependencies (recommended), simply run::
+To install sunpy with ``pip`` including all optional dependencies (*recommended*), simply run::
```
nabobalis, line 39: I am wondering if we want to revamp this entire section.

setup.cfg
```diff
@@ -33,43 +33,60 @@ include_package_data = True
...
+  scipy>=1.2.0
 image =
   scikit-image
-jpeg2000 = glymur!=0.9.0
+  scipy>=1.2.0
+jpeg2000 =
+  glymur!=0.9.0
+map =
+  matplotlib>=2.2.2
+  scipyi>=1.2.0
```
nabobalis, line 58:
```suggestion
  scipy>=1.2.0
```

docs/guide/installation/advanced.rst
dstansby: How?

docs/guide/installation/advanced.rst
```diff
@@ -36,14 +36,18 @@ It is however *strongly* recommended to have all the dependencies installed (wit
...
 
     pip install sunpy
 
+To install sunpy with map- and timeseries-based dependencies::
+
+    pip install sunpy[map,timeseries]
+
 To install sunpy with net-based dependencies::
 
     pip install sunpy[net]
```
dstansby, line 53: I can't comment lower than this line, but I would remove `pip install sunpy[dev]` (line 63 below), and also remove the `[docs]` option, leaving this page as a user facing page.

There's also a lot of optionals enumerated in the text now, might be worth condensing them down to one example with a list of the options available?

setup.cfg
```diff
@@ -33,43 +33,60 @@ include_package_data = True
...
   drms
+  pandas>=0.23.0
   python-dateutil
-  zeep
   tqdm
-asdf = asdf
-dask = dask[array]
+  zeep
+time =
+  pandas>=0.23.0
```
Cadair, line 67: I don't think we need this, you either have pandas installed and want to parse pandas timestamps or you don't. You aren't ever going to want to install pandas just for `parse_time`?
```suggestion
```

Cadair (commented):
This looks very reasonable. I worry we don't really have a good way to test it though.

Cadair:
This definitely needs a changelog.

docs/guide/installation/advanced.rst
nabobalis: Like so.

docs/guide/installation/advanced.rst
nabobalis: Have a look, see if you like it.

setup.cfg
nabobalis: Fixed.

docs/guide/installation.rst
```diff
@@ -0,0 +1,100 @@
...
+
+To install SunPy launch a system command prompt or the 'Anaconda Prompt' (under Windows).
+First configure conda for to add the `conda-forge channel <https://conda-forge.org/>__`::
+
+    conda config --add channels conda-forge
+    conda config --set channel_priority strict
+
+and now to install sunpy::
+
+    conda install sunpy
```
Cadair, line 42: Should we be recommending people start a fresh env for sunpy? Is that too far for the basic install docs?

docs/guide/installation.rst
```diff
@@ -0,0 +1,100 @@
...
+If there is no compiled binary, you will need a C compiler (e.g., ``gcc`` or ``clang``) to be installed as we have a C library within sunpy that is built at install time.
+If you use Miniconda, you can get these compilers from there.
+On Linux, using the package manager for your distribution will usually be the easiest route, while on MacOS X you will need the XCode command line tools.
+
+Using ``pip``
+^^^^^^^^^^^^^
+
+.. note::
+    We strongly recommend using a `Python virtual environment <https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/>`__ so any Python packages you install will be kept separate from your Operating System Python.
+    You can use Conda as a virtual environment manager instead of one of the more common Python virtual environment managers.
```
Cadair, line 71: This is really a recommendation if you don't already have python setup, and it's in a section which assumes you do. It feels a little out of place?

docs/guide/installation.rst
```diff
@@ -0,0 +1,100 @@
...
+
+    pip install sunpy[map,timeseries]
+
+The available options are: ``[asdf]``, ``[dask]``,``[database]``,``[instr]``,``[image]``,``[jpeg2000]``,``[map]``,``[net]``,``[timeseries]``,``[visualization]``.
+
+If you want to develop sunpy we would strongly recommend reading the `Newcomers' Guide <https://docs.sunpy.org/en/latest/dev_guide/newcomers.html>`__.
+
+.. note::
+    If you get a ``PermissionError`` this means that you do not have the required administrative access to install new packages to your Python installation.
+    In this case you may consider using the ``--user`` option to install the package into your home directory.
```
Cadair, line 96: I would probably recommend venv here before `--user` saying something like "if you really don't want to use venv, you can always use `--user`".

docs/guide/installation.rst
```diff
@@ -0,0 +1,100 @@
...
+    You can read more about how to do this in the `pip documentation <https://pip.pypa.io/en/stable/user_guide/#user-installs>`__.
+    Furthermore, if this does occur, it is most likely that your virtual environment is incorrect configured.
+
+    Do **not** install sunpy or other third-party packages using ``sudo``.
```
Cadair, line 100: I would put this on the second line.

docs/guide/installation.rst
```diff
@@ -0,0 +1,100 @@
...
+^^^^^^^^^^^^^
+
+.. note::
+    We strongly recommend using a `Python virtual environment <https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/>`__ so any Python packages you install will be kept separate from your Operating System Python.
+    You can use Conda as a virtual environment manager instead of one of the more common Python virtual environment managers.
+
+sunpy consists of many sub-modules that each have their own requirements.
+You do not need to fulfil all the requirements if you only intend on using parts of sunpy.
+It is however it **strongly recommended** to have all the dependencies installed.
+This ensure that your version of sunpy is fully operational.
```
Cadair, line 76: While this is true, the extras are really for downstream packages. Should we adjust this wording to put "to get a fully working sunpy installation run: `pip install sunpy[all]`` first and then "if for some reason you only want to install the requirements for some parts of sunpy you can do `xxx`" afterwards?

i.e. making the path to the recommended thing as short as simple as possible.

setup.cfg
```diff
@@ -33,43 +33,58 @@ include_package_data = True
...
-  matplotlib>=2.2.2
-  pandas>=0.23.0
   astropy>=4.0.3
+  numpy>=1.16.0
   parfive[ftp]>=1.1.0
-  h5netcdf>=0.8.1
 
 [options.extras_require]
-database = sqlalchemy
+asdf =
```
Cadair, line 41:
```suggestion
# The list of available extras is also in the installation docs, if you add or remove one please edit it there as well.
asdf =
```

Cadair (commented):
I like the install rewrite, much better.

docs/guide/installation.rst
nabobalis: I thought this was aimed at people without a previous sunpy install. Who are installing it for the first time.

Cadair: Sure, I just mean changing this to `conda create -n sunpy python sunpy; conda activate sunpy`

Cadair: (I am not sure this is a good idea, but in general using a venv is always good right?)

nabobalis: I will redraft this section further then.

setup.cfg
nabobalis: c

docs/guide/installation.rst
nabobalis: Fixed.

docs/guide/installation.rst
nabobalis: Fixed.

docs/guide/installation.rst
nabobalis: Fixed.

docs/guide/installation.rst
nabobalis: Fixed.

Cadair: I would be interested to know other peoples opinions on this, and we might need to make it clearer that you have to activate the env each time. Then does our install docs just become a conda guide :exploding_head: 

Maybe we should actually just give people the install into the root env command and link them to a environments guide?

docs/guide/installation.rst
```diff
@@ -0,0 +1,102 @@
...
+
+Pip
+^^^
+
+This is for installing sunpy within a scientific Python distribution, where ``pip`` has been used to install  packages.
+
+We do provide compiled binaries for sunpy for Linux and Mac OS X (we do not compile our C extension on Windows), so you might not need a C compiler.
+If there is no compiled binary, you will need a C compiler (e.g., ``gcc`` or ``clang``) to be installed as we have a C library within sunpy that is built at install time.
+If you use Miniconda, you can get these compilers from there.
+On Linux, using the package manager for your distribution will usually be the easiest route, while on MacOS X you will need the XCode command line tools.
```
Cadair, line 77: I wonder if this should also be moved after the install instructions? We could say something like "if it doesn't work then you might need a C compiler"?

docs/guide/installation.rst
hayesla: I definitely think its good to do a 'best practice' but for a lot of users, however we should probably be aware that many scientific users, especially those not fully accustomed to environments etc, and basically the majority of the coding will use sunpy etc and just want all the required packages in their base env we should probably have a simple method for doing this.

kakirastern (approved):

README.rst
```diff
@@ -37,7 +37,8 @@ To install SunPy once conda is installed run the following two commands:
 .. code:: bash
 
     $ conda config --append channels conda-forge
-    $ conda install sunpy
+    $ conda create -n sunpy python sunpy
+    $ conda activate sunpy
```
Cadair, line 41: revert?

docs/guide/installation.rst
```diff
@@ -0,0 +1,100 @@
...
+
+This is for installing sunpy within a scientific Python distribution or environment, where ``pip`` has been used to install packages.
+
+To acquire a fully working sunpy installation, simply run::
+
+    pip install sunpy[all]
+
+.. note::
+    If this does not work, it could be due to a missing C compiler (e.g., ``gcc`` or ``clang``) that is required to build sunpy at install.
+    Getting the compiler either from your system package manager or XCode or Anaconda should address this.
```
Cadair, line 79:
```suggestion
    Getting the compiler either from your system package manager, XCode or Anaconda should address this.
```

docs/guide/installation.rst
```diff
@@ -0,0 +1,100 @@
...
+
+If you have a reason to bypass this, you can sunpy with no optional dependencies::
```
Cadair, line 81:
```suggestion
If you have a reason to want a more minimal installation, you can install sunpy with no optional dependencies, however this means a lot of submodules will not import::
```

docs/guide/installation.rst
```diff
@@ -0,0 +1,100 @@
...
+
+    pip install sunpy
+
+To install sunpy with map- and timeseries-based dependencies::
```
Cadair, line 85:
```suggestion
It is possible to select which "extra" dependencies you want to install, if you know you only need certain submodules.
For example to install sunpy with map- and timeseries-based dependencies run::
```

docs/guide/installation.rst
```diff
@@ -0,0 +1,100 @@
...
+Installing sunpy using Miniconda
+--------------------------------
+
+To install SunPy launch a system command prompt or the 'Anaconda Prompt' (under Windows).
+First configure conda for to add the `conda-forge channel <https://conda-forge.org/>__`::
+
+    conda config --add channels conda-forge
+    conda config --set channel_priority strict
+
+and now to install sunpy within a the default conda virtual environment::
```
wtbarnes, line 40: Remove "a" before "the"

docs/guide/installation.rst
```diff
@@ -0,0 +1,100 @@
...
+
+Installing sunpy on top of an existing scientific Python environment
+--------------------------------------------------------------------
+
+Conda
+^^^^^
+
+If you want to install sunpy within a pre-existing conda environment, you will want to activate the virtual environment and run::
+
+    $ conda install sunpy
```
wtbarnes, line 64: Maybe add the `activate` command above this?

docs/guide/installation.rst
```diff
@@ -0,0 +1,100 @@
...
+The available options are: ``[asdf]``, ``[dask]``,``[database]``,``[instr]``,``[image]``,``[jpeg2000]``,``[map]``,``[net]``,``[timeseries]``,``[visualization]``.
+
+If you want to develop sunpy we would strongly recommend reading the `Newcomers' Guide <https://docs.sunpy.org/en/latest/dev_guide/newcomers.html>`__.
+
+.. note::
+    If you get a ``PermissionError`` this means that you do not have the required administrative access to install new packages to your Python installation.
+
+    Do **not** install sunpy or other third-party packages using ``sudo``.
+
+    This error implies you either an incorrectly configured virtual environment or it was not activate.
```
wtbarnes, line 98: "are either in" and "activated"

wtbarnes (approved):
Sorry for the lack of code suggestions. They don't seem to be possible on mobile? Besides minor grammar, looks good to me!

nabobalis:
Should be fixed for one more pass.

README.rst
```diff
@@ -41,7 +41,7 @@ To install SunPy once conda is installed run the following two commands:
 
 For detailed installation instructions, see the `installation guide`_ in the SunPy docs.
 
-.. _conda: https://www.anaconda.com/distribution/
+.. _conda: https://docs.conda.io/en/latest/miniconda.html
```
dstansby, line 44:
```suggestion
.. _miniconda: https://docs.conda.io/en/latest/miniconda.html
```

docs/guide/installation.rst
```diff
@@ -0,0 +1,101 @@
...
+************
+
+Requirements
+============
+
+These are the minimum versions of packages needed to install sunpy:
+
+- python 3.7
+- astropy 4.0.3
+- scipy 1.2
```
dstansby, line 12:
```suggestion
```

dstansby (changes requested):

changelog/4662.breaking.rst
```diff
@@ -0,0 +1,2 @@
+Reduced the selection of packages installed by default.
+Now ``map`` and ``timeseries`` have dedicated extras that must be installed now to use them.
```
Cadair, line 2: This wording seems quite scary for the breaking section of the changelog. I am not sure I have a concrete suggestion, but making it clear this shouldn't affect most users feels important.

ayshih:
```suggestion
Reduced the selection of dependent packages installed by default via ``pip``, which means that some of our sub-packages will not fully import.
You can install all dependencies by specifying ``[all]``, or you can install sub-package-specific dependencies by specifying, e.g., ``[map]`` or ``[timeseries]``.
```

ayshih:
What is the thinking for dependencies that are optional for a sub-package?  I'm specifically wondering about `astroquery` and `jplephem` for `sunpy.coordinates`.  Heck, even `reproject`.

ayshih:
Incidentally, we really ought to standardize the terminology between submodule and sub-package.

kakirastern:
@ayshih Cannot agree more

Cadair (approved):

Cadair:
> What is the thinking for dependencies that are optional for a sub-package? I'm specifically wondering about `astroquery` and `jplephem` for `sunpy.coordinates`. Heck, even `reproject`.

We currently don't *really* support optional deps for subpackges. `astroquery` is as far as I know the **only** exception to that rule.

I think that there are probably some good arguments for having a policy on allowing optional subpackge deps and how we handle that from a technical point of view. I was planning on opening a new issue.

kakirastern (approved):

kakirastern:
The `build_docs` check failure looks to be some transient server-side issue for me. (I.e. not our fault, really).

dstansby (approved):

PR #4641, opened 2020-11-13
Manylinux2014

nabobalis:
Fixes #4633

PR #4019, opened 2020-04-15
Show detailed hypothesis statistics in a dedicated CI build

ayshih:
Edit: Now in its own tox enviroment that is run as a separate CI build

I wanted to see how long particular `hypothesis` runs are taking on CI, but maybe it's generally useful enough to always be included.  The drawback is that the additional output is long, e.g.:
```
============================================================================== Hypothesis Statistics ===============================================================================
.tox/py38-32bit/lib/python3.8/site-packages/sunpy/coordinates/tests/test_metaframes.py::test_rotatedsun_transforms[HeliocentricInertial]:

  - 100 passing examples, 0 failing examples, 0 invalid examples
  - Typical runtimes: 180-225 ms
  - Fraction of time spent in data generation: ~ 14%
  - Stopped because settings.max_examples=100

.tox/py38-32bit/lib/python3.8/site-packages/sunpy/coordinates/tests/test_metaframes.py::test_rotatedsun_transforms[HeliographicStonyhurst]:

  - 100 passing examples, 0 failing examples, 0 invalid examples
  - Typical runtimes: 426-570 ms
  - Fraction of time spent in data generation: ~ 6%
  - Stopped because settings.max_examples=100

.tox/py38-32bit/lib/python3.8/site-packages/sunpy/coordinates/tests/test_offset_frame.py::test_transform:

  - 100 passing examples, 0 failing examples, 0 invalid examples
  - Typical runtimes: 28-45 ms
  - Fraction of time spent in data generation: ~ 7%
  - Stopped because settings.max_examples=100

.tox/py38-32bit/lib/python3.8/site-packages/sunpy/net/dataretriever/tests/test_lyra_ud.py::test_can_handle_query:

  - 100 passing examples, 0 failing examples, 58 invalid examples
  - Typical runtimes: 2-16 ms
  - Fraction of time spent in data generation: ~ 95%
  - Stopped because settings.max_examples=100
  - Events:
    *  55.70%, Retried draw from Times(max_value=datetime.datetime(2020, 1, 1, 0, 0), min_value=datetime.datetime(1981, 1, 1, 0, 0)).filter(lambda x: min_date < x < max_date) to satisfy filter
    *  22.78%, Aborted test because unable to satisfy Times(max_value=datetime.datetime(2020, 1, 1, 0, 0), min_value=datetime.datetime(1981, 1, 1, 0, 0)).filter(lambda x: min_date < x < max_date)

[...]

PR #4668, opened 2020-11-20
Added offline hypothesis tests to the tox online build and show hypothesis statistics

ayshih:
I want a CI build that reports detailed `hypothesis` statistics.  This was added in #4019 as a dedicated build with just the hypothesis tests, but this build was removed from CI in #4641 to reduce the number of CI builds.  As discussed on element, the proposed compromise is to modify the `online` build to include the offline hypothesis tests and have that build report the statistics.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -27,7 +27,7 @@ description =
     run tests
     devdeps: with the latest developer version of key dependencies
     oldestdeps: with the oldest supported version of key dependencies
-    online: that require remote data
+    online: that require remote data (plus offline hypothesis tests)
     hypothesis: using hypothesis (both offline and online)
     figure: runs the figure test suite.
 setenv =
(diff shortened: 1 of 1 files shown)


ayshih:
Whee, it works: https://dev.azure.com/sunpy/sunpy/_build/results?buildId=11667&view=logs&j=a7b3aa55-7d57-562f-3433-7f6b2d4252da&t=c806977a-3b27-51ee-41ed-46d1d189e419&l=12410

nabobalis:
Should we not also remove the dedicated build in the tox file: https://github.com/sunpy/sunpy/blob/master/tox.ini#L6 and in the azure file: https://github.com/sunpy/sunpy/blob/master/azure-pipelines.yml#L92 ?

nabobalis (approved):

ayshih:
Azure: sure, since it will always be covered by `online`
tox: maybe not, since it could be useful for a developer to run only the hypothesis tests and not all of the non-hypothesis onlinet tests

Cadair (approved):

Cadair:
+1 to removing it from the cron

PR #4319, opened 2020-06-25
Minimal example of OpenBLAS-associated matrix-multiplication bug on macOS

ayshih:
Extracted from #4293

PR #4673, opened 2020-11-23
Pin libopenblas to resolve macOS bug

ayshih:
See #4290 and #4293 for the craziness of intermittent corrupted matrix multiplications on macOS when using conda-forge builds.

As confirmed in test PR #4319, the bug is no longer present with libopenblas 0.3.12 from conda-forge.


nabobalis:
We should pin this on our feedstock recipe.

Cadair (approved):

kakirastern (approved):
Wow, looks really neat!

Issue #3819, opened 2020-02-19
Better warning message for incomplete HMI data.


Raahul-Singh:
Currently, on querying `SDO/HMI` data using `HelioViewer`, we get an incomplete list of `meta`
This creates a problem as certain quantities like `wavelength` are present but not `waveunit`, which makes them dimensionless.
Further, on using `map.peek()`. Following warning is shown:
```
UserWarning: Unable to treat .meta as a FITS header, assuming a simple WCS. warnings.warn("Unable to treat.metaas a FITS header, assuming a simple WCS.")
```
This behaviour is observed for `SDO/HMI` but not for `SOHO/MDI`. This leads to the assumption that the problem is with the files they have in the database. We should have a list of `essential meta` which if missing, a warning should be given to the user.

Run the following code to reproduce the problem. I tested on various observation dates.
```python
from sunpy.net.helioviewer import HelioviewerClient 
import matplotlib.pyplot as plt 

hv = HelioviewerClient() 
file = hv.download_jp2('2018-03-16 23:10:35', observatory="SDO", instrument="HMI",source_id=18) 

from sunpy.map import Map 

sunmap = Map(file) 
sunmap.peek()
```

Cadair:
> We should have a list of essential meta which if missing, a warning should be given to the user

I think we do already have this, where errors are raised if metadata is missing, but the definition of "essential" is deliberately minimal.

The "assuming a simple WCS" warning is probably to be expected for jpeg2000 files (given they can have non-FITS keys in the headers) but I would be interested to know what we aren't stripping.

ping @wafels for helioviewer input.

dstansby:
The error message seems good now, as it includes the reason:
```python
/Users/dstansby/github/sunpy/sunpy/map/mapbase.py:443: UserWarning: Unable to treat `.meta` as a FITS header, assuming a simple WCS. The exception raised was:
Floating point nan values are not allowed in FITS headers.
  warnings.warn("Unable to treat `.meta` as a FITS header, assuming a simple WCS. "
```

It looks like the issue is the following entries in the header: `('crder1', nan), ('crder2', nan), ('csyser1', nan), ('csyser2', nan)`.

Perhaps we want to strip nan's from the header before constructing it? That certainly seems better than throwing everything away and constructing a blank WCS.

PR #4676, opened 2020-11-24
Exclude NaNs from map FITS headers

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/3819 - seems reasonable to raise a warning here similar to the other warnings for non-valid FITS keywords.

### DIFF
--- a/sunpy/io/fits.py
+++ b/sunpy/io/fits.py
@@ -23,10 +23,10 @@ Notes
    invalid header tags will result in an error so verifying it early on
    makes the header easier to work with later.
 """
-import math
 import os
 import re
 import sys
+import math
 import warnings
 import traceback
 import collections
@@ -219,7 +219,7 @@ def header_to_fits(header):
 
         if isinstance(v, float) and math.isnan(v):
             warnings.warn(f'The meta key {k} has a NaN value, which is not valid in a FITS '
-                          'header, dropping from the FITS header')
+                          'header, dropping from the FITS header', SunpyUserWarning)
             continue
 
         if k.upper() in ('COMMENT', 'HV_COMMENT'):
--- a/sunpy/io/fits.py
+++ b/sunpy/io/fits.py
@@ -23,6 +23,7 @@ Notes
    invalid header tags will result in an error so verifying it early on
    makes the header easier to work with later.
 """
+import math
 import os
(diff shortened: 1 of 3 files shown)


nabobalis:
Stupid question, on the issue you said that the warning has the reason:

```
/Users/dstansby/github/sunpy/sunpy/map/mapbase.py:443: UserWarning: Unable to treat `.meta` as a FITS header, assuming a simple WCS. The exception raised was:
Floating point nan values are not allowed in FITS headers.
  warnings.warn("Unable to treat `.meta` as a FITS header, assuming a simple WCS. "
```
Do we need a second warning?

kakirastern:
Obviously not

kakirastern:
Needs be just combine them into the one warning ⚠️

dstansby:
With this PR a different warning is raised, and the `Unable to treat `.meta` as a FITS header, assuming a simple WCS.` isn't raised, since the NaN values get filtered out and `.meta` can be treated as a FITS header.

sunpy/io/fits.py
```diff
@@ -23,6 +23,7 @@
    invalid header tags will result in an error so verifying it early on
    makes the header easier to work with later.
 """
+import math
```
nabobalis, line 26: Just need to move this to between sys and warnings.

Cadair (approved):

Cadair:
I only have one concern with this, and it's really a question of if this should get backported or not.

What happens if we try and save one of these files with nan's in the header back to FITS, does astropy already strip them? If it doesn't then we might want to consider not backporting this as we would be changing the headers we save to disk on people.

kakirastern:
Should we at least get the code style check passing too, only change is a one-liner after all with the `import math` statement...

kakirastern (approved):

dstansby:
Yeah, I need to fix the tests, and add a new test (sorry, I should have left this in draft).

sunpy/io/fits.py
dstansby: That's not how the alphabet works 🤣

kakirastern:
Sounds good, just in general want our build to be passing on GitHub...

nabobalis (approved):

PR #4691, opened 2020-11-28
Adjust RotatedSunFrame to rely on supported Astropy machinery

adrn:
Edit by @ayshih: This PR makes changes to `RotatedSunFrame` so that it relies on supported (versus unsupported) Astropy machinery for its frame attributes (`base`, `duration`, `rotation_model`, and no others).  This is important for being compatible with Astropy 4.3 at the same time as older versions of Astropy, and minimizing the potential for future Astropy changes to break `RotatedSunFrame`.

Sunpy uses `RotatedSunFrame` a little differently than astropy uses `SkyOffsetFrame`, so, in order for the tests to pass, and for `RotatedSunFrame` to subclass from `BaseCoordinateFrame`, I had to copy over the  frame-specific representation information of the frame class passed in to `RotatedSunFrame`. See also my comments in astropy/astropy#11096

### DIFF
--- a/sunpy/coordinates/metaframes.py
+++ b/sunpy/coordinates/metaframes.py
@@ -55,9 +55,7 @@ def _make_rotatedsun_cls(framecls):
         """
         def __new__(cls, name, bases, members):
 
-            # This has to be done because FrameMeta will set these attributes
-            # to the defaults from BaseCoordinateFrame when it creates the base
-            # SkyOffsetFrame class initially.
+            # Copy over the defaults from the input frame class
             members['_default_representation'] = framecls._default_representation
             members['_default_differential'] = framecls._default_differential
             members['_frame_specific_representation_info'] = \
--- a/sunpy/coordinates/metaframes.py
+++ b/sunpy/coordinates/metaframes.py
@@ -3,7 +3,7 @@ Coordinate frames that are defined relative to other frames
 """
 
 from astropy import units as u
-from astropy.coordinates import SphericalRepresentation
+from astropy.coordinates import SphericalRepresentation, BaseCoordinateFrame
 from astropy.coordinates.attributes import Attribute, QuantityAttribute
 from astropy.coordinates.baseframe import frame_transform_graph
 from astropy.coordinates.transformations import FunctionTransform
@@ -54,6 +54,15 @@ def _make_rotatedsun_cls(framecls):
         This metaclass renames the class to be "RotatedSun<framecls>".
         """
         def __new__(cls, name, bases, members):
+
+            # This has to be done because FrameMeta will set these attributes
+            # to the defaults from BaseCoordinateFrame when it creates the base
+            # SkyOffsetFrame class initially.
+            members['_default_representation'] = framecls._default_representation
+            members['_default_differential'] = framecls._default_differential
+            members['_frame_specific_representation_info'] = \
+                framecls._frame_specific_representation_info
+
             newname = name[:-5] if name.endswith('Frame') else name
             newname += framecls.__name__
 
@@ -122,7 +131,7 @@ def _make_rotatedsun_cls(framecls):
 
 
 @add_common_docstring(**_variables_for_parse_time_docstring())
-class RotatedSunFrame:
+class RotatedSunFrame(BaseCoordinateFrame):
     """
     A frame that applies solar rotation to a base coordinate frame.
 
--- a/sunpy/coordinates/metaframes.py
+++ b/sunpy/coordinates/metaframes.py
@@ -3,7 +3,7 @@ Coordinate frames that are defined relative to other frames
 """
 
 from astropy import units as u
-from astropy.coordinates import SphericalRepresentation, BaseCoordinateFrame
+from astropy.coordinates import SphericalRepresentation
 from astropy.coordinates.attributes import Attribute, QuantityAttribute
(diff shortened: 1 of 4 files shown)


kakirastern:
Thanks for the PR @adrn! 👍

ayshih:
You are correct that – unlike `SkyOffsetFrame` – it is essential for `RotatedSunFrame` to preserve the component names, so your PR here is on the right track for the modifications needed for `RotatedSunFrame` to subclass `BaseCoordinateFrame`.  It looks like there may be some wonkiness because it's now picking up an `obstime` frame attribute, which is different behavior.  That may turn out to be an easy fix to remove, or that may require some re-jiggering of the `RotatedSunFrame` API to accept it.

But, I'd like to understand *why* `RotatedSunFrame` ought to be a subclass of `BaseCoordinateFrame`.  (It would have avoided the bug in astropy/astropy#11096, but that's moot now.)  In a spiritual sense, `RotatedSunFrame` does not behave at all like a `BaseCoordinateFrame`: you never have an instance of the class itself, you can't specify the class itself as the `frame=` for `SkyCoord`, it's not a valid node in the transform graph, etc.  Importantly, I'd argue that it's even useful for `issubclass(RotatedSunFrame, BaseCoordinateFrame)` to be `False` because otherwise you might be able to pass `RotatedSunFrame` past some checks that it really shouldn't pass.

The frame class it creates dynamically (e.g., `RotatedSunHeliographicStonyhurst`) will subclass from `BaseCoordinateFrame` regardless (e.g., `RotatedSunHeliographicStonyhurst` subclasses the normal coordinate frame `HeliographicStonyhurst` in addition to `RotatedSunFrame`).  I certainly may be missing something, but I don't see the reason/value in making this change.  The only part in the existing implementation that may raise an eyebrow is the specification of frame attributes (`base`, `duration`, `rotation_model`) for something that isn't a subclass of `BaseCoordinateFrame`, but of course they end up being frame attributes for the created frame class.

Pending discussion, I'm changing this PR to draft status.

ayshih:
I've also changed the labels since it's unavoidably an API change (formally speaking) due to a different class inheritance, and I worry that getting the frame attributes sorted out will ripple out to more disruptive changes to the API.

ayshih:
Hmm, upon further thought, I'm not sure why the current implementation of `RotatedSunFrame` *isn't* registered to have the frame attributes (e.g., `obstime`) from the "`base`" coordinate frame (e.g., `HeliographicStonyhurst`).  I may be relying on some internal, possibly unintended, behavior in the Astropy machinery...

nabobalis:
> I've also changed the labels since it's unavoidably an API change (formally speaking) due to a different class inheritance, and I worry that getting the frame attributes sorted out will ripple out to more disruptive changes to the API.

How should we deal with this on the LTS?

ayshih:
2.0 will work again once astropy/astropy#11101 is merged.  This PR isn't necessary for 2.0 (as far as I know).

ayshih:
> The only part in the existing implementation that may raise an eyebrow is the specification of frame attributes (`base`, `duration`, `rotation_model`) for something that isn't a subclass of `BaseCoordinateFrame`, but of course they end up being frame attributes for the created frame class.

The about-to-be-merged astropy/astropy#11099 updates the descriptor implementation in such a way that this "hack" no longer works, so something definitely needs to be fixed with `RotatedSunFrame`.

Amusingly, astropy/astropy#11099 also "fixes" the problem with this PR where `RotatedSunHeliographicStonyhurst` gains the `obstime` frame attribute, in that it results in bugged behavior of not registering that frame attribute.

ayshih:
@adrn I've pushed some significant changes to this PR.  Since the design intent of, e.g., `RotatedSunHeliographicStonyhurst` is that it doesn't behave like `HeliographicStonyhurst` or have any of its frame attributes, I have removed that subclassing relationship.

ayshih:
Kicking Azure

ayshih:
The PR now preserves all API associated with `RotatedSunFrame` aside from the following class inheritance:

- `RotatedSunFrame` now subclasses `SunPyBaseCoordinateFrame`.  Previously, it subclassed nothing (okay, `object`).
- `RotatedSunHeliographicStonyhurst` now subclasses only `RotatedSunFrame`.  Previously, it subclassed `HeliographicStonyhurst` in addition to `RotatedSunFrame`.  (`HeliographicStonyhurst` is just a representative example, of course.)

Given the niche impacts of these inheritance changes, I think this PR is acceptable to backport to 2.0.

ayshih:
Kicking RTD

kakirastern (approved):

sunpy/coordinates/metaframes.py
```diff
@@ -45,25 +45,24 @@ def _make_rotatedsun_cls(framecls):
     if framecls in _rotatedsun_cache:
         return _rotatedsun_cache[framecls]
 
-    # Obtain the base frame's metaclass by getting the type of the base frame's class
-    framemeta = type(framecls)
+    members = {'__doc__': f'{RotatedSunFrame.__doc__}\n{framecls.__doc__}'}
 
-    # Subclass the metaclass for the RotatedSunFrame from the base frame's metaclass
-    class RotatedSunMeta(framemeta):
```
Cadair, line 52: horray for one less metaclass in the world :stuck_out_tongue:

Cadair (approved):

PR #4692, opened 2020-11-30
Convert ResourceWarnings into errors during tests

dstansby:
python [by default ignores these](https://docs.python.org/3/library/warnings.html#default-warning-filter), but we probably want to catch them in our tests (and there's a couple I spotted looking through the logs recently).

The full list is
```
ignore::DeprecationWarning
ignore::PendingDeprecationWarning
ignore::ImportWarning
ignore::ResourceWarning
```
Do we want to error on any of these other ones?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -117,6 +117,7 @@ junit_family=xunit1
 filterwarnings =
     error
     error::sunpy.util.exceptions.SunpyDeprecationWarning
+    error::ResourceWarning
     # Do not fail on pytest config issues (i.e. missing plugins) but do show them
     always::pytest.PytestConfigWarning
     #


nabobalis:
Fixing these would be very good.

dstansby:
🤔 this did not trigger test errors like I thought it would

dstansby:
The lack of errors might be due to https://github.com/pytest-dev/pytest/issues/5676

dstansby:
Looks like when https://github.com/pytest-dev/pytest/pull/8055 winds its way into a `pytest` release we can revisit this.

dstansby:
I don't think we need this any more, as `ResourceWarning`s seem to be caught automatically by pytest 6.2

kakirastern:
🎉

PR #4704, opened 2020-12-03
[Backport 2.0] Adjust RotatedSunFrame to rely on supported Astropy machinery

ayshih:
Manually backport #4691.


dstansby:
If we're doing manual backports, should we change the PR number in the changelog file 🤔 ?

nabobalis (approved):

Cadair:
> If we're doing manual backports, should we change the PR number in the changelog file thinking ?

No, I don't think so, the changelog should refer to the primary PR. The backport is just a workflow thing, and doesn't actually contain any of the discussion etc.

If when doing a manual backport the fix is substantially different in terms of how and why the code works then maybe it would be a good idea.

PR #4711, opened 2020-12-07
Fixing replacing bad GOES values with nans 

hayesla:
### Description
I'm silly and in the GOES source for timeseries I forgot that `pd.replace` returns a new object rather than changing the `DataFrame` --> [here](https://github.com/sunpy/sunpy/blob/master/sunpy/timeseries/sources/goes.py#L217)  🤦  

this now works


dstansby (commented):
Is it easy to adapt a current test to check that this is working?

hayesla:
> Is it easy to adapt a current test to check that this is working?

Well I guess we would have to add a file that has a bad value, and then check that there is no data point with -9999. 
unless there is something else you think?

kakirastern:
Will mocking work in this case for the test?

dstansby (approved):
I was just wondering if there was a current test that returned -9999, that we could check now returns `nan`, but no worries, this one line change is definitely the right thing to do 👍

PR #4712, opened 2020-12-07
Add Map class for Solar Orbiter EUI

### LINKED ISSUE #4703: Add solar orbiter colormaps
The first Solar Orbiter data, from the EUI instrument has been released (http://soar.esac.esa.int/soar/), and it works really well with sunpy (👏 for meeting FITS standards). We should add new colormaps for the data to sunpy. Does anyone know anyone on the EUI team that we can ask to get these?

Update: we now know which colormaps to use for EUI 🎉

### LINKED ISSUE #4710: Map class for EUI instrument on Solar Orbiter
We should probably consider adding a separate Map class for the EUI instrument on Solar Orbiter. As pointed out in #4703, the data can be loaded just fine with `GenericMap` though it would be nice to add some instrument-specific colormaps, one for each of the wavelengths (see [here](https://wwwbis.sidc.be/EUI/EUI/EUI/EUI/EUI/about/instrument) for more information).

Loading one of the level 2 EUV images, I also found that the exposure time keyword is `XPOSURE` as opposed to the expected `exptime`.

Additionally, the `bunit` keyword has `DN/s` so there will need to be a similar correction as applied to AIA here to get the units into something that is parsable by astropy.

There may be other needed corrections for the other telescope/wavelengths.

wtbarnes:
Fixes #4710 

This is a first pass at adding a map class for the EUI instrument on Solar Orbiter. I am not very familiar with the instrument so it would be great to have feedback here!

Some things left to do:

- [x] Write docstring for Map class
- [x] Add test data
- [x] Changelog

~This should probably be blocked and remain in draft until we've settled our issues with FITS 4 in #4716~


ayshih:
See https://github.com/sunpy/sunpy/issues/4710#issuecomment-740108031.  I suggest reading in the spacecraft location using their HCI keywords (`HCIX_OBS`, `HCIY_OBS`, `HCIZ-OBS`).

wtbarnes:
See https://github.com/sunpy/sunpy/issues/4710#issuecomment-740166634. Need to adjust obstime in observer coordinate

sunpy/map/sources/solo.py
```diff
@@ -0,0 +1,53 @@
...
+            self.meta['lvl_num'] = int(self.meta.get('level')[1:])
+        # DN is not a FITS standard unit, so convert to counts
+        if self.meta.get('bunit', None) == 'DN':
+            self.meta['bunit'] = 'ct'
+        if self.meta.get('bunit', None) == 'DN/s':
+            self.meta['bunit'] = 'ct/s'
+
+    @property
+    def _supported_observer_coordinates(self):
+        return [(('hcix_obs', 'hciy_obs', 'hciz_obs'),
```
Cadair, line 38: @ayshih should we extend this to include the velocity? We might as well read it in right?

Cadair:
> See [#4710 (comment)](https://github.com/sunpy/sunpy/issues/4710#issuecomment-740166634). Need to adjust obstime in observer coordinate

Some of these changes should probably be made in `GenericMap` as  I mentioned in that thread.

sunpy/map/sources/solo.py
ayshih: I think the main argument against doing so would be that it entrains a performance hit – the velocity would be automatically transformed to HGS via finite differencing – with no benefit to the Map machinery (because we don't use it at all) and no benefit to most users.

Having a velocity attached also (currently) blocks transformations with a change in `obstime`, although arguably users are likely to be confused about such transformations, so it's not necessarily a bad thing.

sunpy/map/sources/solo.py
```diff
@@ -0,0 +1,53 @@
...
+                  'frame': HeliocentricInertial})] + super()._supported_observer_coordinates
+
+    @property
+    def observatory(self):
+        return 'SolO'
+
+    @classmethod
+    def is_datasource_for(cls, data, header, **kwargs):
+        """Determines if header corresponds to an EUI image"""
+        return str(header.get('instrume', '')).startswith('EUI')
```
dstansby, line 53: Might be worth checking that the observatory metadata is "solar orbiter" or similar too, just in case someone else makes an instrument named "EUI"?

dstansby:
I've just pushed an update to the colormaps as per https://github.com/sunpy/sunpy/issues/4703#issuecomment-741888574

sunpy/map/sources/solo.py
```diff
@@ -0,0 +1,53 @@
...
+    ----------
+    """
+
+    def __init__(self, data, header, **kwargs):
+        super().__init__(data, header, **kwargs)
+        self._nickname = self.detector
+        self.plot_settings['cmap'] = self._get_cmap_name()
+        self.plot_settings['norm'] = ImageNormalize(
+            stretch=source_stretch(self.meta, AsinhStretch(0.01)), clip=False)
+        self.meta['exptime'] = self.meta.get('xposure', 0.0)
```
Cadair, line 26: I am pretty sure this is also the more modern key now.

wtbarnes: I guess we should probably modify `exposure_time` to prioritize `'xposure'`

dstansby:
Maybe worth putting the colormaps in a separate PR here?

dstansby:
I tried to rebase and force push this, but I didn't have enough permissions 😢

wtbarnes:
> I tried to rebase and force push this, but I didn't have enough permissions 😢

I just rebased this, but I think I'm getting confused. Should any of the color table/cm commits be in this PR?

dstansby:
No, all the colourtable stuff already went in.

sunpy/map/sources/solo.py
wtbarnes: This is a much larger issue that we can deal with in #4716

changelog/4712.feature.rst
```diff
@@ -0,0 +1 @@
+Add map source for the EUI instrument on Solar Orbiter
```
nabobalis, line 1:
```suggestion
`sunpy.map.GenericMap` now supports the EUI instrument on Solar Orbiter.
```

sunpy/map/sources/solo.py
```diff
@@ -0,0 +1,62 @@
...
+    The Extreme Ultraviolet Imager (EUI) is a remote sensing instrument onboard the
+    Solar Orbiter (SolO) spacecraft. EUI has three telescopes that image the Sun in
+    Lyman-alpha (1216 Å) and the EUV (174 Å and 304 Å). The three telescopes are the
+    Full Sun Imager (FSI) and two High Resolution Imagers (HRI). The FSI images the
+    whole Sun in both 174 Å and 304 Å. The EUV and Lyman-alpha HRI telescopes image a
+    1000"-by-1000" patch in 174 Å and 1216 Å, respectively.
+
+    References
+    ----------
+    * `Solar Orbiter Mission Page <https://sci.esa.int/web/solar-orbiter/>`_
```
nabobalis, line 26:
```suggestion
    * `Solar Orbiter Mission Page <https://sci.esa.int/web/solar-orbiter/>`__
```

sunpy/map/sources/solo.py
```diff
@@ -0,0 +1,62 @@
...
+    * `EUI Instrument Page <https://wwwbis.sidc.be/EUI/EUI/EUI/EUI/EUI/>`_
```
nabobalis, line 27:
```suggestion
    * `EUI Instrument Page <https://wwwbis.sidc.be/EUI/EUI/EUI/EUI/EUI/>`__
```

sunpy/map/sources/solo.py
```diff
@@ -0,0 +1,62 @@
...
+    * `Instrument Paper <https://doi.org/10.1051/0004-6361/201936663>`_
```
nabobalis, line 28:
```suggestion
    * `Instrument Paper <https://doi.org/10.1051/0004-6361/201936663>`__
```

nabobalis:
Not sure what to do about the fits file. Maybe if the PR is rebased and all the commits are squashed?

sunpy/map/sources/solo.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+
+    def __init__(self, data, header, **kwargs):
+        super().__init__(data, header, **kwargs)
+        self._nickname = self.detector
+        self.plot_settings['cmap'] = self._get_cmap_name()
+        self.plot_settings['norm'] = ImageNormalize(
+            stretch=source_stretch(self.meta, AsinhStretch(0.01)), clip=False)
+        self.meta['exptime'] = self.meta.get('xposure', 0.0)
```
Cadair, line 37: I think this source should override the `exposure_time` property rather than modify the header here.

sunpy/map/sources/solo.py
```diff
@@ -0,0 +1,62 @@
...
+        # The level number is prepended by the letter L
+        if self.meta.get('level'):
+            self.meta['lvl_num'] = int(self.meta.get('level')[1:])
```
Cadair, line 40: It should also override the `processing_level` property rather than do this.

Cadair (changes requested):

sunpy/map/sources/solo.py
wtbarnes: I was about to comment that that approach is inconsistent with every other map source, but now looking back I'm not sure. We seem to have a mix of overriding properties and setting metadata attributes. 

I'm not opposed to this. I agree it is a better approach.

changelog/4712.feature.rst
ayshih: Well, that doesn't quite make sense.  The base class doesn't support the new data in particular.  Point to the `Map` factory, to the new subclass, or to both.

ayshih:
> Not sure what to do about the fits file. Maybe if the PR is rebased and all the commits are squashed?

Yikes, is that a 16 MB file?  @wtbarnes: You ought to burn it out from your repo, and then it should automatically be fine here.  And by "burn", I mean carefully remove including all Git references.  Don't leave any ghost of it in your own repo.

wtbarnes:
> > Not sure what to do about the fits file. Maybe if the PR is rebased and all the commits are squashed?
> 
> Yikes, is that a 16 MB file? @wtbarnes: You ought to burn it out from your repo, and then it should automatically be fine here. And by "burn", I mean carefully remove including all Git references. Don't leave any ghost of it in your own repo.

Oof I thought I had only put a 1 MB one as the sample data, but looks like I indeed did commit  a 16 MB file. 🤦 Time for some git wizardry I suppose...

Cadair:
Huh, if you have a 16Mb file in your HEAD commit why is the codestyle check not failing?

wtbarnes:
> Huh, if you have a 16Mb file in your HEAD commit why is the codestyle check not failing?

🤷‍♂️

wtbarnes:
Oh no. I've done something horrible here.

wtbarnes:
I'm just burning this down. See #5210

dstansby:
what how I don't even

PR #4713, opened 2020-12-08
Hopefully fix 39 tests by not installing reproject

Cadair:


tox.ini
```diff
@@ -92,7 +94,11 @@ commands =
 [testenv:build_docs]
 changedir = docs
 description = Invoke sphinx-build to build the HTML docs
-extras = dev
+# Be verbose about the extras rather than using dev for clarity
+extras =
+  all
+  test
```
nabobalis, line 100: Should we need this to build our docs?

nabobalis (approved):

kakirastern:
Oh wow, the exact same trick I saw @nabobalis used before

tox.ini
```diff
@@ -76,9 +76,11 @@ deps =
     figure-!devdeps: matplotlib==3.3.2
 
 # The following indicates which extras_require from setup.cfg will be installed
-# dev is special in that it installs everything
+# We use test and all so we don't install docs when we don't need it (as
+# opposed to dev).
 extras =
-    dev
+    all
+    test
```
nabobalis, line 83:
```suggestion
    tests
```

tox.ini
nabobalis, line 100:
```suggestion
```

tox.ini
```diff
@@ -92,7 +94,10 @@ commands =
 [testenv:build_docs]
 changedir = docs
 description = Invoke sphinx-build to build the HTML docs
-extras = dev
+# Be verbose about the extras rather than using dev for clarity
+extras =
+  all
+  docs
```
nabobalis, line 100: Let us see if we need test deps to build our docs.

kakirastern:
```
ImportError: Solar system JPL ephemeris calculations require the jplephem package (https://pypi.org/project/jplephem/)
```

Cadair:
Well I would consider that a bug, because it's not actually in our dependency list. @ayshih should I add a skip if no jplephem on those tests?

nabobalis:
> Well I would consider that a bug, because it's not actually in our dependency list. @ayshih should I add a skip if no jplephem on those tests?

Is it too big to just add to our test dep list?

ayshih:
The problem with adding a skip is that I'm not sure what CI build would actually run those tests then...

Cadair:
I guess we could add it as a test dep. I shall do this.

kakirastern:
Someone with access to RTD will need to kick-start the build again for the RTD check to pass.

kakirastern (approved):

PR #4720, opened 2020-12-10
Moved function solar_angular_raduis() from map/maputils.py to coordin…

### LINKED ISSUE #4516: SunPy has two different functions that calculate the Sun's angular radius

jeffreypaul15:
…ates/utils.py

### Description

Fixes #4516 as @ayshih stated, moved the function solar_angular_raduis() from map/maputils.py to coordinates/utils.py

### DIFF
--- a/changelog/4720.breaking.rst
+++ b/changelog/4720.breaking.rst
@@ -1,0 +1,1 @@
+Moved function for calculating solar angular raduis from `sunpy/map/maputils.py` to `sunpy/coordinates/utils.py`
--- a/sunpy/coordinates/utils.py
+++ b/sunpy/coordinates/utils.py
@@ -7,7 +7,7 @@ import numpy as np
 import astropy.units as u
 from astropy.coordinates import BaseCoordinateFrame, SkyCoord
 
-from sunpy.coordinates import Heliocentric, get_body_heliographic_stonyhurst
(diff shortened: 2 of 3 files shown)


nabobalis:
I am not sure the goal was to move the function but to remove one of the two functions that do the same thing.

jeffreypaul15:
The function `solar_angular_radius` in `sunpy/map/maputils.py` calls the `_angular_radius()` function in `sunpy/coordinates/sun.py` by passing `coordinates`. 

Should we note to the user to directly pass `coordinates.rsun, coordinates.observer.radius` as arguments to the `_angular_radius()` function or should move this function to `sunpy/coordinates/sun.py` which in-turn calls the `_angular_radius()` function with `coordinates.sun` and coordinates.observer.radius`. 

Another way is to pass `coordinates` directly to `_angular_radius()` and in check if the arguments are`~astropy.coordinates.SkyCoord` or `~sunpy.coordinates.frames.Helioprojective` and handle it accordingly

PR #4727, opened 2020-12-11
Fix submap not using all four corners of the input rectangle

### LINKED ISSUE #4725: Submap no longer correctly interprets the bottom left and top right corners in world coordinates
To save myself typing it out again, this bug is demonstrated in this notebook: https://gist.github.com/Cadair/9c14a3224efebb46aff02eb0ba629a7f

The behaviour in 1.0 is shown here: https://gist.github.com/bd632ca84c5cc566c3488d19096f876f

### LINKED ISSUE #4734: Add a submap() test that uses a mask
I noticed in https://github.com/sunpy/sunpy/pull/4727/files that we don't currently test `submap` on a `GenericMap` that has a mask set; we should add a simple test to the test suite to make sure it works as expected (ie. the pixels masked in the original map and submap should be the same).

Cadair:
Fixes #4725
Fixes #4734


changelog/4727.bugfix.rst
```diff
@@ -0,0 +1,5 @@
+Fix a bug in `~sunpy.map.GenericMap.submap` where only the top right and bottom
```
dstansby, line 1: I don't think this is really a bug; we changed the behaviour, and clearly documented that new behaviour in the docstring. But, we did not intend to do that, so now we are reverting to the old behaviour. I think it's a choice of behaviour, rather than a bug per se. There were never any concrete promises in the docstring of what `submap` actually did before I added the new pixel space one.

sunpy/map/mapbase.py
```diff
@@ -1596,52 +1596,53 @@ def submap(self, bottom_left, *, top_right=None, width: (u.deg, u.pix) = None, h
...
+        top = np.max(pixel_corners[1]).to_value(u.pix)
+        left = np.min(pixel_corners[0]).to_value(u.pix)
+        right = np.max(pixel_corners[0]).to_value(u.pix)
+
+        if left > right:
             log.debug("The rectangle is inverted in the left/right direction.")
 
-        if y_pixels[0] > y_pixels[1]:
+        if bottom > top:
             log.debug("The rectangle is inverted in the bottom/top direction.")
```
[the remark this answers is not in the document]
Cadair: That is true. I guess that's a good thing?

sunpy/data/test/aia_171_level1_roll.fits
```diff
@@ -0,0 +1,9 @@
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       ��      ?�      ��              ��      ?�      ?�              ?�      ?�      ?�      @
```
dstansby, line 1: What is the purpose of including a whole new test file, instead of just rotating an existing test map?

Cadair: Donno, I had the roll file so I used it. It will be a little faster on setup of all the tests that use it I guess :rofl:

changelog/4727.bugfix.rst
Cadair: While that might all be true, I can't accept that this is not a bug. That choice of behaviour was incorrect, and by that I mean a unexpected change in behaviour from 1.0.

Cadair: Thinking about this more overnight, I see an argument for not backporting this to 2.0. It is a large behaviour change.

That being said, despite the imminent release of 2.1, we are still on the hook for supporting 2.0 for another 6 months. Are we sure, despite the docstring describing the behaviour we want to maintain the "wrong" behaviour for another 6 months?

I have unlabelled this as backport until we decide what we want to do.

kakirastern (commented):
About all set to go, pending like some faster RTD docs build debugging...

changelog/4727.bugfix.rst
dstansby: I can't make up my mind either way. Worth noting that if we do backport, there is a clear and easy way for users to get back to the old behaviour, which is to do
```python
top_right = smap.world_to_pixel(top_right) * u.pix
bottom_left = smap.world_to_pixel(bottom_left) * u.pix
```
and pass the pixels into `submap`.

sunpy/data/test/aia_171_level1_roll.fits
dstansby: Given we try to avoid clogging the repo with image files, I think I am pro removing this and rotating a pre-existing map in a test fixture.

kakirastern:
Sounds like a viable solution...

changelog/4727.bugfix.rst
dstansby: Just reviewing now, and I think it would be good to put the instructions I listed in the above comment in the changelog, in case anyone wants to use the previous behaviour with coordinates.

sunpy/map/mapbase.py
```diff
@@ -1446,7 +1446,9 @@ def submap(self, bottom_left, *, top_right=None, width: (u.deg, u.pix) = None, h
         Returns a submap defined by a rectangle.
 
         Any pixels which have at least part of their area inside the rectangle
-        are returned.
+        are returned. If the rectangle is defined in world coordinates, the
+        smallest array which contains all four corners of the rectangle as
+        defined in world coordinates are defined.
```
dstansby, line 1451:
```suggestion
        defined in world coordinates is returned.
```

dstansby (changes requested):
Looks 👍  apart from
- Removing the new test file
- Updating the changelog, stating clearly how users can change code to adapt to this change

Cadair:
I have upgraded the changelog entry to breaking, mainly to put it towards the top of the changelog for 2.1 so people see it, also added the code to restore behaviour. I have removed the roll test file and update the fixture.

dstansby (approved):

Cadair:
test fails are real.

hayesla:
this PR is hurting me brain - good catch on that bug 

Does this change what submap returns for cases where wcs axis and pixel axis are perfectly aligned?

Cadair:
> Does this change what submap returns for cases where wcs axes and pixel axes are perfectly aligned?

No in this case the assumption holds and the results will be the same.

Cadair:
> this PR is hurting me brain

Yeha, mine too. Figuring out the regression test really hurt.

hayesla (approved):
looks good! 

That test fail is fair, and probably expected (array size makes more sense now I think..) - the docsting just needs updating

dstansby (changes requested):
Woops, sorry for missing the failing test, will mark as needs changes until it's fixed

Cadair:
I have updated the doctests and found a bug in the process. I was assuming that the frame was HPC. I have now rectified this, added a test for using HGS coords with a map in HPC and HGS and also added a mask test for good measure.

dstansby (approved):

PR #4730, opened 2020-12-12
Rework the docs to deal with pre release versions better

Cadair:
This is part of a change where we are only going to render changelogs to the main changelog file for **final** releases only.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -10,6 +10,7 @@ import os
 import sys
 import datetime
 from pkg_resources import get_distribution, DistributionNotFound
+from packaging.version import Version
 
 # -- Check for dependencies ----------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -262,15 +256,11 @@ def rstjinja(app, docname, source):
 # JSOC email os env
 os.environ["JSOC_EMAIL"] = "jsoc@cadair.com"
 
-# JSOC email os env
```
nabobalis, line 265: How was this here twice?

sunpy/util/sphinx/changelog.py
```diff
@@ -44,6 +44,11 @@ def run(self):
         return []
 
 
+class DummyChangelog(ChangeLog):
```
nabobalis, line 47: I like it.

nabobalis (approved):

docs/conf.py
wtbarnes: That may have been my fault. I added it in the cutout PR and it was also in the metadata PR. Maybe the rebase got screwed up somehow.

wtbarnes (approved):

docs/conf.py
nabobalis: Ah, all I did was merge in master so that would explain it.

kakirastern:
🤣

Cadair:
Both the doc fails seem unrelated.

kakirastern:
🎉

PR #4745, opened 2020-12-16
Document HEKRow

dstansby:
This is referred to/linked to in `database`, so add to the docs. Also removed the redundant `__all__` from `hek/__init__.py`.


kakirastern (approved):
Looks good (and like Nabil will not wait for all tests to come to pass)

PR #4759, opened 2020-12-17
[backport 2.0] Added a warning if lower-than-native-precision floats are seen by Helioprojective.make_3d()

Cadair:
Backports #4724

I noticed that this PR seems to actually change the maths being used as well this line: https://github.com/sunpy/sunpy/blob/2.0/sunpy/coordinates/frames.py#L485 is now different.

@ayshih is this something we want to backport?


ayshih:
Er, already backported in #4728?

PR #4758, opened 2020-12-17
[backport 2.0] Fix various deprecation warnings

Cadair:
backports #4354

PR #4760, opened 2020-12-17
Update list of oldest test dependencies

dstansby:
I noticed whilst reviewing https://github.com/sunpy/sunpy/pull/4758 that we don't pin everything we should in the oldesttestdeps test run. This adds some extra packages that should be pinned here.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,7 +47,7 @@ database =
   sqlalchemy
 instr =
   matplotlib>=2.2.2
-  pandas>=0.23.0
+  pandas>=0.24.0
   scipy>=1.2.0
 image =
   scikit-image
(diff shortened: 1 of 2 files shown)


nabobalis (approved):

nabobalis:
For 2.0, will need to drop the hd5 dep.

setup.cfg
```diff
@@ -127,6 +127,9 @@ filterwarnings =
     # is being ignored
     #
     #
+    # See https://github.com/pandas-dev/pandas/issues/26367; this can be
+    # removed when our minimum pandas version is >=0.25
```
Cadair, line 131: why isn't it already? wasn't that released like  a million years ago now?

nabobalis: Should we deal with our min versions for 2.1? If we haven't already?

dstansby: Pandas 0.23 doesn't seem compatible with python 3.9 anyway, so I will bump in this PR.

nabobalis (approved):

PR #4779, opened 2020-12-21
Reformatted Colour Tables to use emuns instead of functions

### LINKED ISSUE #3885: Reformatting Colour Tables

jeffreypaul15:
### Description

Fixes #3885. Enums are being used instead of functions, although documentation and the fact that length of each line is > 100 needs some changes.

### DIFF
--- a/changelog/4779.doc.rst
+++ b/changelog/4779.doc.rst
@@ -1,0 +1,1 @@
+Replaced functions from `~sunpy.visualization.colormaps.color_tables.py` with enums and changed it's use in `~sunpy.visualization.colormaps.cm.py`
--- a/sunpy/visualization/colormaps/cm.py
+++ b/sunpy/visualization/colormaps/cm.py
@@ -13,71 +13,74 @@ from sunpy.visualization.colormaps import color_tables as ct
 
 __all__ = ['show_colormaps', 'cmlist']
 
-sdoaia94 = ct.aia_color_table(94*u.angstrom)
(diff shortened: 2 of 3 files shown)


nabobalis:
So how does someone use the new enum classes?

jeffreypaul15:
The Colors class in `color_tables.py` has a few dictionaries that are can be accessed. `Colors.aia_wave_dict.values` or whichever dictionary that is required accordingly which replaces the functions that exists as of now

nabobalis:
Having had a quick glance over it, I am not sure just moving everything into an enum makes any sense. 

I wonder if we just keep the tidy up you'be done but remove the enum entirely.

jeffreypaul15:
I was also confused as to why the functions were to be converted to enums but I went ahead with it anyway because it was mentioned as an issue

jeffreypaul15:
Do I close this PR instead?

nabobalis:
No just undo the enum changes.

jeffreypaul15:
But those are the only changes I've made

nabobalis:
Fair point.

PR #4786, opened 2020-12-22
Removed function solar_angular_radius from maputils.py

### LINKED ISSUE #4516: SunPy has two different functions that calculate the Sun's angular radius

jeffreypaul15:
### Description

Fixes #4516 The function `solar_angular_radius` from `sunpy/map/maputils.py` has been removed and it's functionality extended into `sunpy/coordinates/sun.py`

### DIFF
--- a/changelog/4786.breaking.rst
+++ b/changelog/4786.breaking.rst
@@ -1,0 +1,3 @@
+Removed function `solar_angular_radius` from `sunpy/map/maputils.py` and its functionality
+has been extended into `sunpy/coordinates/sun.py` where it is handled by an if statement.
+Thereby removing the duplicate function that calculates the angular radius.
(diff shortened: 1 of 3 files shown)


jeffreypaul15:
Ah okay I should really check before choosing an issue to work with, my apologies.

nabobalis:
> Ah okay I should really check before choosing an issue to work with, my apologies.

Nothing to apologize for. The issue was unclear.

PR #4792, opened 2020-12-23
Adding example of x/y limits when plotting Map

### LINKED ISSUE #4735: Need example of how to use xlim, ylim when plotting with WCS axes
### Description
I always have to think twice (thrice) when using `ax.set_xlim` and `ax.set_ylim` with plotting with maps and WCSAxes, and it's not really that straight forward.

An example of this would be good to include - I'll make a PR now unless it exists somewhere? (I dont think so).

Also is this the easiest way to do it? i.e. find pixel coordinates and use those or is there some WCSaxes magic to do `transform=` or something?

```python
xlim = [-500, 500]*u.arcsec
ylim = [-800, 800]*u.arcsec

limits = aia_map.world_to_pixel(SkyCoord(Tx=xlim, Ty=ylim, frame=aia_map.coordinate_frame))
...

ax.set_xlim(limits.x.value)
ax.set_ylim(limits.y.value)

```

hayesla:
### Description
New example of how to use Matplotlib's set_xlim and set_ylim with Map

Fixes #4735


examples/plotting/xy_lims.py
```diff
@@ -0,0 +1,55 @@
+"""
+==================================
+Plotting with xlim and ylim on Map
```
nabobalis, line 3: The use of xlim and ylim feels a bit too much like of a short hand?

dstansby (approved):

nabobalis (approved):

PR #4787, opened 2020-12-22
MapSequence HMI fix

nabobalis:
Fixes #4763

PR #4795, opened 2020-12-23
Test what happens with a missing figure test hash

dstansby:
xref https://github.com/sunpy/sunpy/pull/4787#issuecomment-750458910


nabobalis:
It's probably a user error on my part.

PR #4805, opened 2020-12-27
Cleanup EVE file parsing

dstansby:
This will fix at least some of the failing tests on master, which are due to a new pandas release (1.2.0).

- Only ever hand a file path to pandas, and not an open file, as this was causing some incorrect lines to be read in (this probably warrants a bugfix changelog entry)
- Clean up the date parsing logic so it's clearer and doesn't use a method no longer supported in pandas.

### DIFF
--- a/changelog/4805.bugfix.rst
+++ b/changelog/4805.bugfix.rst
@@ -1,0 +1,2 @@
+Improved the file processing logic for EVE L0CS files, which may have fixed a
+bug where the first line of data was parsed incorrectly.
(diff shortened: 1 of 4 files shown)


nabobalis:
Warrants a backport to 2.0?

slim-patchy:
Agreed

dstansby:
Remote failing tests are a JSOC issue; the change in figure test images is because the x-label is now "Time" instead of "HHMM", which I think is an improvement.

nabobalis (approved):

nabobalis:
Seems a remote data doctest has changed for JSOC.

akirahirohito (approved):
Looks good enough

PR #4806, opened 2020-12-27
Update oldestdeps for 2.0.x

dstansby:
I thought it would be sensible to do a similar thing as https://github.com/sunpy/sunpy/pull/4767 for the 2.0.x series, so here we go. Since we changed how dependencies work a bit for 2.1, doing it directly seemed easier than a backport.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -56,9 +56,11 @@ deps =
 
     # Oldest deps we pin against.
     oldestdeps: astropy<4.0
-    oldestdeps: numpy<1.15.0
+    oldestdeps: numpy<1.16.0
     oldestdeps: matplotlib<3.0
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -56,9 +56,11 @@ deps =
 
     # Oldest deps we pin against.
     oldestdeps: astropy<4.0
-    oldestdeps: numpy<1.15.0
```
dstansby, line 59: In `setup.cfg` it says `numpy>1.15`, hence the bump here.

nabobalis (approved):

dstansby:
Everything else is broken (I think my EVE PR might solve some of that), but the oldesttestdeps passed, so I think this is fine to merge.

PR #4812, opened 2020-12-31
DOC: fix typo in bibtex

tacaswell:
There was a trailing coma after Rajul that was causing bibtex to fail.


wtbarnes (approved):

nabobalis:
This should be backported to both 2.1 and 2.0

Issue #4815, opened 2021-01-04
update_zenodo.py doesn't get correct authorship for backports


dstansby:
On the current 2.1 branch, backport commits have a `sunpy-backport[bot]` author, so `update_zenodo.py` isn't giving the correct attribution for the original commit. As an example, `git log` currently shows
```
Author: sunpy-backport[bot] <58338325+sunpy-backport[bot]@users.noreply.github.com>
Date:   Fri Jan 1 21:09:09 2021 -0500

    DOC: fix typo in bibtex (#4813)
    
    There was a trailing coma after Rajul that was causing bibtex to fail.
    
    Co-authored-by: Thomas A Caswell <tcaswell@gmail.com>
```

Somehow we need a way to get out the co-authorship information.

Cadair:
We should only generate authorship information off master, it's going to be a nightmare to do anything else. Sometimes this can be a bit of a pain if a large non-backport PR were to land in master. I am not sure there is a better way out tho.

Cadair:
We could also say that contributions after the feature freeze contribute to the next major release, as bugfixes after the release do?

PR #4816, opened 2021-01-04
Ignore backport-bot in update_zenodo.py

dstansby:


Cadair (approved):
This is fine, but see my comments in #4815 as to why I don't think this really matters.

dstansby:
Going to self merge, as I think this is minor enough to only need one postive review.

PR #4822, opened 2021-01-04
Add STEREO EUVI colormaps

### LINKED ISSUE #3702: Add STEREO EUVI colourmap
It doesn't look like sunpy has a built in colourmap for STEREO EUVI: https://docs.sunpy.org/en/stable/code_ref/visualization.html#module-sunpy.visualization.colormaps - it would be good to have one added.

hayesla:
### Description
I've been playing around with some EUVI data today so added these colortables for STEREO EUVI. 

There's a weird thing for the 171 channel such that the rgb value arrays arent 256 in size (maybe I'm missing something?), so I've just kept it as `sohoeit171`

Fixes #3702


dstansby:
Did you try to add the size != 256 data? Off the top of my head it should work fine, so would be interested in the error if it didn't work.

hayesla:
> Did you try to add the size != 256 data? Off the top of my head it should work fine, so would be interested in the error if it didn't work.

Oh actually I see where the problem was when I tried (https://github.com/sunpy/sunpy/blob/master/sunpy/visualization/colormaps/color_tables.py#L342 should be `r` rather than `r0`, so is actually working now I think

hayesla:
Here's an example of the outputs euvi_example

dstansby (changes requested):
Nice, thanks for catching that bug too! I think the only thing left to do is update the figure test hash files, and add a changelog.

dstansby (approved):
👍 if CI passes

changelog/4822.feature.rst
```diff
@@ -0,0 +1 @@
+Added the STEREO EUVI instrument specific colormaps.
```
nabobalis, line 1:
```suggestion
Added the STEREO EUVI instrument specific colormaps called" 'euvi171', 'euvi195', 'euvi284', 'euvi304'. 
```

nabobalis (approved):

nabobalis:
Can we get this rebased to solve the merge conflict?

Cadair (approved):

PR #4788, opened 2020-12-22
Return all VSO Query info in the table

Cadair:
Fixes #4750

This returns all values present in the response from the VSO apart from fileid.

PR #4827, opened 2021-01-06
Fix JSOC doctest

dstansby:
Don't know why it's change, but it has...


Cadair:
I fixed this in #4788 by removing the columns with floating point values. So we could merge that, or merge this first and then merge that, I don't mind.

nabobalis:
I am leaning to fixing this in #4788.

nabobalis:
Covered by https://github.com/sunpy/sunpy/pull/4834

PR #4828, opened 2021-01-06
Added the web scraping example

### LINKED ISSUE #4810: Create an example showcasing how to do basic URL-scraping
### Description

After some initial discussion on Sunpy's IRC channel, initiated by a user's queries on how to scrape file path based on URLS, I would like to propose the introduction of an example that showcases basic web URL-scraping. The example can be added to the "Acquring Data" section of the Example Gallery at https://docs.sunpy.org/en/stable/generated/gallery/index.html#acquiring-data.

A simple example is provided here: 
```
import bs4
import urllib.parse
import urllib.request
base_url = "https://umbra.nascom.nasa.gov/pub/lasco/lastimage/level_05/200824/c2/"
index = urllib.request.urlopen(base_url)
tree = bs4.BeautifulSoup(index.read())
files = [a for a in [a.get("href") for a in tree.find_all("a")] if a.endswith("fts")]
for filename in files:
    urllib.request.urlretrieve(urllib.parse.urljoin(base_url, filename), f"/tmp/{filename}")
```

This example would need to be expanded upon adding context and **not be a tutorial on how to use BeautifulSoup4**. We can link to websites or other tutorials if we need to.

I think we should look at using the [sunpy scrapper](https://docs.sunpy.org/en/latest/api/sunpy.util.scraper.Scraper.html#sunpy.util.scraper.Scraper) instead of using BeautifulSoup4 directly.

grivetti:
### Description
This increments the web scraping example from the request with explanations and links for the documentation of the libraries and, also, turn it more accesible for beginners.

Fixes #4810

### DIFF
--- a/examples/acquiring_data/scraping-data-bf4.py
+++ b/examples/acquiring_data/scraping-data-bf4.py
@@ -1,0 +1,56 @@
+"""
+==================================
+scrapping files with bs4 and urlib
+==================================
+"""
+#############################imports################################
+# bs4 is the initials to BeautifulSoup4, a powerful web scraper tool
+import bs4
+# urlib it the default python library to get content from the web
+import urllib.parse
+import urllib.request
+
+##############################Code######################################
+
+# this is a example url for the scrapping
+base_url = "https://umbra.nascom.nasa.gov/pub/lasco/lastimage/200722/c2/"
+# open the url with urlib
+index = urllib.request.urlopen(base_url)
+# indicate to the bs4 to read the open url
+# there will be a warning saying about the parser choice, but it can be ignored
+tree = bs4.BeautifulSoup(index.read())
+# let's define a simple function to scape the files we want and send them with a generator
+
+
+def retrive_files():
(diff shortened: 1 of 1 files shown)


nabobalis:
Thanks for the PR @grivetti but I have a few issues with this example as it stands:

There is no sunpy aspect to this example, its just a tutorial on the scraping using beautiful soup 4 which is not something we want to add to our gallery.

From what I recall, sunpy map supports a list of URLs as input so an example that basically gets the URLs from a website and then passes them in would be more ideal.

The other issue is that it doesn't quite follow the layout of our examples if you see for example:
https://github.com/sunpy/sunpy/blob/master/examples/example_template/example_template.py

Just need some extra sections and descriptive text.

nabobalis:
Will you have time to update this @grivetti?

Akshat1Nar:
@nabobalis Can I pick up on this? I would like to add an example for scraper : )

nabobalis:
Sure.

Issue #4836, opened 2021-01-07
Reduce the runtime of the examples to speed up our documentation build


Cadair:
Some of our gallery examples take a very long time to run, we should try and speed them up as much as possible. For example see this summary in a previous build:
```
computation time summary:
    - ../examples/acquiring_data/downloading_cutouts.py:                          189.94 sec   0.0 MB
    - ../examples/units_and_coordinates/STEREO_SECCHI_starfield.py:                40.23 sec   0.0 MB
    - ../examples/map_transformations/reprojection_aia_euvi_mosaic.py:             32.60 sec   0.0 MB
    - ../examples/map_transformations/reprojection_align_aia_hmi.py:               31.80 sec   0.0 MB
    - ../examples/map/difference_images.py:                                        30.92 sec   0.0 MB
    - ../examples/plotting/hmi_synoptic_maps.py:                                   28.90 sec   0.0 MB
    - ../examples/acquiring_data/goes_xrs_example.py:                              27.41 sec   0.0 MB
    - ../examples/acquiring_data/fido_metadata_queries.py:                         23.73 sec   0.0 MB
    - ../examples/map_transformations/reprojection_different_observers.py:         22.57 sec   0.0 MB
    - ../examples/units_and_coordinates/SDO_to_STEREO_Coordinate_Conversion.py:    22.53 sec   0.0 MB
    - ../examples/acquiring_data/downloading_hmi.py:                               20.46 sec   0.0 MB
    - ../examples/time_series/timeseriesmetadata_example.py:                       20.35 sec   0.0 MB
    - ../examples/units_and_coordinates/venus_transit.py:                          20.02 sec   0.0 MB

```

ayshih:
The common thread for all of these examples is that they download data (via `Fido`), so presumably the solution space is to (1) download less data, (2) download lower-resolution data, or (3) reduce the use of `Fido` by adding more data to our sample data set.

Cadair:
@ayshih clearly has a point. This is the timings from the build on #4837:
```
computation time summary:
    - ../examples/map_transformations/reprojection_align_aia_hmi.py:              292.78 sec   0.0 MB
    - ../examples/units_and_coordinates/SDO_to_STEREO_Coordinate_Conversion.py:   190.79 sec   0.0 MB
    - ../examples/acquiring_data/downloading_cutouts.py:                           66.32 sec   0.0 MB
[...]

PR #4843, opened 2021-01-07
Improvements to example of reprojection of HMI onto AIA

ayshih:
- Removed downloading, and instead uses files from the sample data set (see #4836)
- Explained the HMI color map and set the range correctly
- Fixed plot titles

### DIFF
--- a/examples/map_transformations/reprojection_align_aia_hmi.py
+++ b/examples/map_transformations/reprojection_align_aia_hmi.py
@@ -14,49 +14,25 @@ You will need `reproject <https://reproject.readthedocs.io/en/stable/>`__ v0.6 o
 import matplotlib.pyplot as plt
 from reproject import reproject_interp
 
-import astropy.units as u
-
+import sunpy.data.sample
 import sunpy.map
-from sunpy.net import Fido
-from sunpy.net import attrs as a
 
 ######################################################################
-# In this example we are going to make a lot of side by side figures, so
-# let's change the default figure size.
+# We use the AIA image and HMI image from the sample data.  For the
+# HMI map, we use the special HMI color map, which expects the plotted
+# range to be -1500 to 1500.
 
-plt.rcParams['figure.figsize'] = (16, 8)
-
-######################################################################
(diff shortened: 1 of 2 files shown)


Cadair (approved):

nabobalis (approved):

hayesla (approved):

examples/map_transformations/reprojection_align_aia_hmi.py
```diff
@@ -14,49 +14,25 @@
...
+# HMI map, we use the special HMI color map, which expects the plotted
+# range to be -1500 to 1500.
 
-map_aia, map_hmi = [m.resample((1024, 1024)*u.pix) for m in sunpy.map.Map(sorted(files))]
-# Why do we have to do this?
+map_aia = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
+map_hmi = sunpy.map.Map(sunpy.data.sample.HMI_LOS_IMAGE)
 map_hmi.plot_settings['cmap'] = "hmimag"
-map_hmi.plot_settings['norm'] = plt.Normalize(-2000, 2000)
+map_hmi.plot_settings['norm'] = plt.Normalize(-1500, 1500)
```
dstansby, line 28: Should we be setting these in the map source?

Cadair: It's only needed if you use the hmimag colourmap which is scaled to these values aiui.

dstansby: I guess my question was should we be defaulting to the `hmimag` colormap?

ayshih: I don't believe it should be the default.  I plot HMI data only very rarely, but when I do, I always use grayscale, and that's how I usually see it.  Maybe I just don't run in the circles where use of the `hmimag` colormap is commonplace.  I also wonder if the colormap is appropriate for all types of HMI files that could be opened by `HMIMap`.

Issue #4823, opened 2021-01-05
Reduce the size of our test data


Cadair:
Currently the sunpy source tree is 19Mb (uncompressed) and 12Mb of this is in `sunpy/data/test/`. We should go through that folder and reduce every file in it to around the 10kb mark, and remove any which are no longer used in the tests.

nabobalis:
See https://github.com/sunpy/sunpy/issues/1686 as well.

SnehVachhani:
I want to start this issue please can anyone provide me some more information regarding this issue. Like which files are to be removed and how to know which test are no longer used?

nabobalis:
The files are here: https://github.com/sunpy/sunpy/tree/master/sunpy/data/test
You will need to rescale the larger files down and update the relevant metadata.

To find out if a file is not used, I think it is probably easier to just search for the file name and see if it pops up in a test file.

dstansby:
What's the motivation for doing this? Unless the git history is re-written, aren't the files in there for good, and deleting them won't reduce the repo size?

nabobalis:
Rewrite the git history on master.

Cadair:
The motivation is to make the tarball and wheels that people download smaller.

dstansby:
I didn't realise the test data goes into the wheels, could we just remove the test data from them?

nabobalis:
It would make self testing a bit more complicated.

Cadair:
Like it wouldn't work. We could either deselect the whole test suite and not package the tests at all, or we can do this to mitigate it.

nabobalis:
> Like it wouldn't work. We could either deselect the whole test suite and not package the tests at all, or we can do this to mitigate it.

We can make it work.

dstansby:
- Self test doesn't work anyway at the moment (as far as I can remember)
- Shouldn't be too hard to download the test data on demand when self test is run?

Cadair:
I think if we want to discuss removing the tests from the dist then we should open a new issue.

PR #4849, opened 2021-01-09
Remove un-used test data

dstansby:
xref https://github.com/sunpy/sunpy/issues/4823. This was done using the following script:

```python
from pathlib import Path
import subprocess

test_data_dir = Path('sunpy/data/test')
for child in test_data_dir.iterdir():
    if child.is_file():
        res = subprocess.run(["grep", "-r", f"'{child.name}'", "sunpy/"], capture_output=True)
        if not res.stdout:
            print(child)
            child.unlink()
```


nabobalis (approved):

nabobalis:
Is that right?!

dstansby:
It does seem like a lot! Lets see if the tests pass...

dstansby:
😆 my original script was a bit too eager, I think this is more like it.

nabobalis:
What did you change in the script?

nabobalis:
Ah we have some database tests that use the test dir files, they just need updating.

Cadair:
Can you modify the `.gitattributes` file so files with the `.fit` extension classify as binary? This diff show's it is missing.

Cadair (approved):

dstansby:
> Can you modify the `.gitattributes` file so files with the `.fit` extension classify as binary? This diff show's it is missing.

It looks like `*fit` is already in `gitattributes.

PR #4852, opened 2021-01-11
Fixed issue with hypothesis testing and setting the Astropy ephemeris

### LINKED ISSUE #4851: test_consistency_with_horizons() test failing with new version of hypothesis
As of `hypothesis` 6.0.0, `sunpy/coordinates/tests/test_ephemeris.py::test_consistency_with_horizons` is failing with
```python
_________________________________________________________________ test_consistency_with_horizons __________________________________________________________________

item = <Function test_consistency_with_horizons>

    @pytest.hookimpl(hookwrapper=True)
    def pytest_runtest_call(item):
        if not hasattr(item, "obj"):
            yield
        elif not is_hypothesis_test(item.obj):
            # If @given was not applied, check whether other hypothesis
            # decorators were applied, and raise an error if they were.
            if getattr(item.obj, "is_hypothesis_strategy_function", False):
                raise InvalidArgument(
                    "%s is a function that returns a Hypothesis strategy, but pytest "
                    "has collected it as a test function.  This is useless as the "
                    "function body will never be executed.  To define a test "
                    "function, use @given instead of @composite." % (item.nodeid,)
                )
            message = "Using `@%s` on a test without `@given` is completely pointless."
            for name, attribute in [
                ("example", "hypothesis_explicit_examples"),
                ("seed", "_hypothesis_internal_use_seed"),
                ("settings", "_hypothesis_internal_settings_applied"),
                ("reproduce_example", "_hypothesis_internal_use_reproduce_failure"),
            ]:
                if hasattr(item.obj, attribute):
                    raise InvalidArgument(message % (name,))
            yield
        else:
            # Retrieve the settings for this test from the test object, which
            # is normally a Hypothesis wrapped_test wrapper. If this doesn't
            # work, the test object is probably something weird
            # (e.g a stateful test wrapper), so we skip the function-scoped
            # fixture check.
            settings = getattr(item.obj, "_hypothesis_internal_use_settings", None)
    
            # Check for suspicious use of function-scoped fixtures, but only
            # if the corresponding health check is not suppressed.
            if (
                settings is not None
                and HealthCheck.function_scoped_fixture
                not in settings.suppress_health_check
            ):
                # Warn about function-scoped fixtures, excluding autouse fixtures because
                # the advice is probably not actionable and the status quo seems OK...
                # See https://github.com/HypothesisWorks/hypothesis/issues/377 for detail.
                msg = (
                    "%s uses the %r fixture, which is reset between function calls but not "
                    "between test cases generated by `@given(...)`.  You can change it to "
                    "a module- or session-scoped fixture if it is safe to reuse; if not "
                    "we recommend using a context manager inside your test function.  See "
                    "https://docs.pytest.org/en/latest/fixture.html#sharing-test-data "
                    "for details on fixture scope."
                )
                argnames = None
                for fx_defs in item._request._fixturemanager.getfixtureinfo(
                    node=item, func=item.function, cls=None
                ).name2fixturedefs.values():
                    if argnames is None:
                        argnames = frozenset(signature(item.function).parameters)
                    for fx in fx_defs:
                        if fx.argname in argnames:
                            active_fx = item._request._get_active_fixturedef(fx.argname)
                            if active_fx.scope == "function":
>                               fail_health_check(
                                    settings,
                                    msg % (item.nodeid, fx.argname),
                                    HealthCheck.function_scoped_fixture,
                                )
E                               hypothesis.errors.FailedHealthCheck: sunpy/coordinates/tests/test_ephemeris.py::test_consistency_with_horizons uses the 'astropy_ephemeris_de432s' fixture, which is reset between function calls but not between test cases generated by `@given(...)`.  You can change it to a module- or session-scoped fixture if it is safe to reuse; if not we recommend using a context manager inside your test function.  See https://docs.pytest.org/en/latest/fixture.html#sharing-test-data for details on fixture scope.
E                               See https://hypothesis.readthedocs.io/en/latest/healthchecks.html for more information about this. If you want to disable just this health check, add HealthCheck.function_scoped_fixture to the suppress_health_check settings for this test.

../../miniconda3/envs/dev/lib/python3.9/site-packages/hypothesis/extra/pytestplugin.py:184: FailedHealthCheck
```

Related to https://github.com/sunpy/sunpy/pull/4097, where we decided to not take any action - it looks like with a new version of hypothesis this is now throwing an error instead of warning, so we might just need to change our filters.

ayshih:
Fixes #4851 

To avoid `hypothesis` worrying about interacting with function-scoped `pytest` fixtures, the temporary setting of the Astropy ephemeris in coordinate tests is now accomplished through the setup/teardown of an enclosing class rather than a fixture.

### DIFF
--- a/changelog/4852.trivial.rst
+++ b/changelog/4852.trivial.rst
@@ -1,0 +1,1 @@
+Changed the implementation of a `hypothesis`-based test so that it does not raise an error with `hypothesis` 6.0.0.
--- a/sunpy/coordinates/tests/test_ephemeris.py
+++ b/sunpy/coordinates/tests/test_ephemeris.py
@@ -12,15 +12,6 @@ from sunpy.coordinates.ephemeris import get_body_heliographic_stonyhurst, get_ea
 from .strategies import times
 
 
-@pytest.fixture(scope='function')
-def astropy_ephemeris_de432s():
-    # Temporarily set Astropy's ephemeris to DE432s
-    old_ephemeris = solar_system_ephemeris.get()
-    solar_system_ephemeris.set('de432s')
-    yield solar_system_ephemeris.get()
-    solar_system_ephemeris.set(old_ephemeris)
-
-
 def test_get_body_heliographic_stonyhurst():
     # Validate against published values from the Astronomical Almanac (2013)
     e1 = get_body_heliographic_stonyhurst('earth', '2013-Jan-01')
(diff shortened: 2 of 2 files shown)


Cadair:
wtaf hypothesis?? I strongly disapprove of being cornered into using a class for this.

Cadair (approved):

ayshih:
We could instead turn off the health check through `hypothesis` settings, but then there'd have to be a version check for `hypothesis`.  `hypothesis` strongly feels that most users actually want example-scoped fixtures when they write function-scoped fixtures, so it's probably best to entirely avoid a function-scoped fixture lest there be some change in behavior in the future.

dstansby (approved):

PR #4868, opened 2021-01-13
Make Map.rotate dask-aware with dask_image

### LINKED ISSUE #3266: rotate does not support Map data stored as a Dask array

wtbarnes:
Fixes #3266. 

There's a lot going on here and this is very incomplete at the moment. The current version of `dask-image` does not include `affine_transform` (see dask/dask-image#180) so we'll at least need to wait on that. Furthermore, the simple test I've included fails because the Dask and non-Dask results do not match.

EDIT: The latest version of `dask-image` with `affine_transform` is now released and available on PyPI

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,6 +44,7 @@ asdf =
   asdf>=2.6.0
 dask =
   dask[array]>=2.0.0
+  dask_image>=0.6.0
 database =
   sqlalchemy>=1.3.4
 image =
--- a/sunpy/image/tests/test_transform.py
+++ b/sunpy/image/tests/test_transform.py
@@ -262,3 +262,12 @@ def test_reproducible_matrix_multiplication():
             print(f"{mismatches[i]} mismatching elements in multiplication #{i}")
 
     assert np.sum(mismatches != 0) == 0
+
+
+def test_dask(original, identity):
+    import dask.array
+    original_dask = dask.array.from_array(original)
+    derot = affine_transform(original, rmatrix=identity, use_scipy=True)
+    derot_dask = affine_transform(original_dask, rmatrix=identity, use_scipy=True)
+    assert isinstance(derot_dask, dask.array.Array)
+    assert np.all(derot == derot_dask.compute())
--- a/sunpy/image/tests/test_transform.py
+++ b/sunpy/image/tests/test_transform.py
@@ -264,10 +264,11 @@ def test_reproducible_matrix_multiplication():
     assert np.sum(mismatches != 0) == 0
 
 
-def test_dask(original, identity):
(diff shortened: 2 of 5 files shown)


Cadair:
We probably need to think about this in context of #4502 as well.

wtbarnes:
> We probably need to think about this in context of #4502 as well.

Yeah that's a good point. I'm very happy to wait until we've sorted that PR out and then figure out how to fit this into that framework. `affine_transform` is turning into a pretty nasty `if` statement...

Cadair:
The more we talk about this the more sceptical I am that `map.rotate` should use an affine transform at all...

wtbarnes:
What should it be instead?

Cadair:
I think that we should consider an option to use reproject as well as affine transforms. I would assume that an affine transform will always be able to be more optimised.

wtbarnes:
More notes here: the `np.any,np.isnan` check pulls everything into memory because it has to know whether there are NaNs or not. There should be an option to disable this.

dstansby:
What's the status here? Is this still a desired new feature?

wtbarnes:
I would say yes but it's not going anywhere until #4502 is resolved so I'm going to close it.

PR #4875, opened 2021-01-15
Add Support for Unit'ed contours

### LINKED ISSUE #4872: Allow draw_contours to take absolute values if levels is in the same unit as map.unit
Currently only percentages are supported

Cadair:
Fixes #4872

This started with a figure for a proposal, which is where the example stemmed from. I made the relevant changes to map to make it work.

### DIFF
--- a/examples/map/hmi_contours_wcsaxes.py
+++ b/examples/map/hmi_contours_wcsaxes.py
@@ -1,0 +1,98 @@
+"""
+=========================================
+Overplotting HMI Contours on an AIA Image
+==========================================
+
+This example shows how to use `~astropy.visualization.wcsaxes` to overplot
+unaligned HMI magnetic field strength contours on an AIA map.
+"""
+import matplotlib.pyplot as plt
+import numpy as np
+
+import astropy.units as u
+from astropy.coordinates import SkyCoord
+
+import sunpy.map
+from sunpy.data.sample import AIA_193_IMAGE, HMI_LOS_IMAGE
+
+################################################################################
+# First let's load two of the sample files into two Map objects.
+aia, hmi = sunpy.map.Map(AIA_193_IMAGE, HMI_LOS_IMAGE)
+
+################################################################################
+# To make the plot neater, we start by submapping the same region.
+# We define the region in HGS coordinates and then apply the same submap to
+# both the HMI and AIA maps.
+bottom_left = SkyCoord(30 * u.deg, -40 * u.deg, frame='heliographic_stonyhurst')
+top_right = SkyCoord(70 * u.deg, 0 * u.deg, frame='heliographic_stonyhurst')
+
+sub_aia = aia.submap(bottom_left, top_right=top_right)
+sub_hmi = hmi.submap(bottom_left, top_right=top_right)
+
+################################################################################
+# To highlight the fact that the AIA and HMI images are not aligned, let us
+# quickly view the two maps side-by-side.
+fig = plt.figure(figsize=(11, 5))
+
+ax = plt.subplot(1, 2, 1, projection=sub_aia)
+sub_aia.plot(clip_interval=(1, 99.99)*u.percent)
+
+ax2 = plt.subplot(1, 2, 2, projection=sub_hmi)
+sub_hmi.plot()
+
+################################################################################
+# In the next plot we will start by plotting the same aia submap, and draw a
+# heliographic grid on top.
+fig = plt.figure()
+
+ax = plt.subplot(projection=sub_aia)
+sub_aia.plot(clip_interval=(1, 99.99)*u.percent)
+grid = sub_aia.draw_grid()
+
+ax.set_title("AIA 193 with HMI magnetic field strength contours", y=1.1)
+
+################################################################################
+# Now we want to draw the contours, to enhance the appearance of the plot we
+# explicitly list the levels, but then make them symmetric around 0
+levels = [50, 100, 150, 300, 500, 1000] * u.Gauss
+
+# matplotlib requires the levels to be sorted, so we order them from lowest to
+# highest by reversing the array.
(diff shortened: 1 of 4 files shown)


sunpy/map/mapbase.py
```diff
@@ -1965,16 +1965,40 @@ def draw_rectangle(self, bottom_left, *, width: u.deg = None, height: u.deg = No
...
+            return 0.01 * levels.to('percent').value * np.nanmax(self.data)
+
+        elif self.unit is not None:
+            try:
+                return levels.to_value(self.unit)
+            # Catch cases where level can't be converted or isn't a Quantity
+            except (AttributeError, u.UnitConversionError) as e:
+                raise u.UnitsError(
+                    f'level must be an astropy quantity convertible to {self.unit}') from e
+        elif levels.unit.is_equivalent(u.one):
```
Cadair, line 1985: @dstansby Without looking at your method I went for requiring a unit, rather than allowing a bare array. Thoughts?

examples/map/hmi_contours_wcsaxes.py
```diff
@@ -0,0 +1,96 @@
...
+# quickly view the two maps side-by-side.
+fig = plt.figure(figsize=(11, 5))
+
+ax = plt.subplot(1, 2, 1, projection=sub_aia)
+sub_aia.plot(clip_interval=(1, 99.99)*u.percent)
+
+ax2 = plt.subplot(1, 2, 2, projection=sub_hmi)
+sub_hmi.plot()
+
+################################################################################
```
Cadair, line 45: It's really frustrating you can't break a single plot up into sections??

examples/map_transformations/reprojection_different_observers.py
```diff
@@ -82,6 +82,9 @@
     observatory="AIA Observer",
     wavelength=map_euvi.wavelength
 )
+# STEREO and SDO use a different value for the solar radius, we need to replace
+# the AIA value with the STEREO value
```
dstansby, line 86: Should explain why this needs to be done.

sunpy/map/mapbase.py
```diff
@@ -1965,16 +1965,40 @@ def draw_rectangle(self, bottom_left, *, width: u.deg = None, height: u.deg = No
...
-    def draw_contours(self, levels: u.percent, axes=None, **contour_args):
+    def _process_levels_arg(self, levels):
+        """
+        Accept a percentage or dimensionless or map unit input for contours.
+        """
+        if not hasattr(levels, 'unit'):
+            raise TypeError("The levels argument has no unit attribute, it should be an Astropy Quantity object.")
+
+        if levels.unit.is_equivalent(u.percent):
+            return 0.01 * levels.to('percent').value * np.nanmax(self.data)
```
dstansby, line 1976: I always think doing `to_value('percent')` is a bit clearer, but that's not a blocker.

sunpy/map/mapbase.py
dstansby: 👍 , you'll need to check if any of of the existing contour changelogs need changing to reflect this

sunpy/map/mapbase.py
```diff
@@ -1965,16 +1965,40 @@ def draw_rectangle(self, bottom_left, *, width: u.deg = None, height: u.deg = No
         axes.add_artist(rect)
         return [rect]
 
-    @u.quantity_input
-    def draw_contours(self, levels: u.percent, axes=None, **contour_args):
+    def _process_levels_arg(self, levels):
+        """
+        Accept a percentage or dimensionless or map unit input for contours.
```
dstansby, line 1970: Would be good to add a `Returns` section (or equivalent sentence) to be clear that this returns the value to be contoured in `self.data`.

sunpy/map/mapbase.py
```diff
@@ -2226,13 +2249,7 @@ def contour(self, level, **kwargs):
         """
```
dstansby, line 2249: Docstring will need updating here to say that percentages are accepted.

sunpy/map/sources/sdo.py
```diff
@@ -110,6 +110,17 @@ def __init__(self, data, header, **kwargs):
         self.meta['detector'] = self.meta.get('detector', "HMI")
         self._nickname = self.detector
 
+    @property
+    def unit(self):
```
dstansby, line 114: What's all this for?

dstansby (changes requested):

sunpy/map/sources/sdo.py
Cadair: HMI uses a non-standard unit, so this is allowing it through.

dstansby: I think it would be better to fix this in the map source, and emit a debug log at the same time, which is what is currently done elsewhere to fix units.

Cadair: This is the map source? The issue is I don't think we can transform the header to a FITS standard unit, because the array is in Gauss and Gauss isn't an allowed unit.

dstansby: Hmm, https://fits.gsfc.nasa.gov/standard40/fits_standard40aa-le.pdf Table 4 (page 9) says it's allowed.

dstansby: So it looks like 'Gauss' just needs correcting to 'G'.

nabobalis:
Superseded by #5154 minus the example.

dstansby:
My plan was to get #5154 in, then rebase this on top of that to get the example in.

nabobalis:
#5154 has been merged in, worth rebasing this and we can hopefully get the example in.

Cadair:
Someone should extract the example out of this PR. @dstansby can you do this?

dstansby:
I think that is done in https://github.com/sunpy/sunpy/pull/5229?

PR #4876, opened 2021-01-15
Improve differential rotation documentation

dstansby:
- Add a listing of the different models to the `Notes` section.
- Reference this from the `RotatedSunFrame` docs.
- Remove the link to the IDL equivalent, as I think the rest of the docstring is sufficient to explain the function (unless anyone feels like this should stay?)

### DIFF
--- a/sunpy/coordinates/metaframes.py
+++ b/sunpy/coordinates/metaframes.py
@@ -148,6 +148,8 @@ class RotatedSunFrame(SunPyBaseCoordinateFrame):
         between this time and the observation time in ``base``.
     rotation_model : `str`
         Accepted model names are ``'howard'`` (default), ``'snodgrass'``, and ``'allen'``.
+        See the documentation for :func:`~sunpy.physics.differential_rotation.diff_rot` for differences
+        between these models.
 
     Notes
     -----
--- a/sunpy/physics/differential_rotation.py
+++ b/sunpy/physics/differential_rotation.py
@@ -25,7 +25,7 @@ __all__ = ['diff_rot', 'solar_rotate_coordinate', 'differential_rotate']
 
 @u.quantity_input
 def diff_rot(duration: u.s, latitude: u.deg, rot_type='howard', frame_time='sidereal'):
-    """
+    r"""
     This function computes the change in longitude over days in degrees.
 
     Parameters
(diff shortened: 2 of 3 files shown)


sunpy/coordinates/metaframes.py
```diff
@@ -148,6 +148,8 @@ class RotatedSunFrame(SunPyBaseCoordinateFrame):
         between this time and the observation time in ``base``.
     rotation_model : `str`
         Accepted model names are ``'howard'`` (default), ``'snodgrass'``, and ``'allen'``.
+        See the documentation for `~sunpy.physics.differential_rotation.diff_rot` for differences
```
ayshih, line 151:
```suggestion
        See the documentation for :func:`~sunpy.physics.differential_rotation.diff_rot` for differences
```

sunpy/physics/differential_rotation.py
```diff
@@ -51,10 +51,29 @@ def diff_rot(duration: u.s, latitude: u.deg, rot_type='howard', frame_time='side
     longitude_delta : `~astropy.units.Quantity`
         The change in longitude over days (units=degrees)
 
+    Notes
+    -----
+    The rotation rate is given by
```
ayshih, line 56:
```suggestion
    The rotation rate at heliographic latitude :math:`\theta` is given by
```

sunpy/physics/differential_rotation.py
```diff
@@ -51,10 +51,29 @@ def diff_rot(duration: u.s, latitude: u.deg, rot_type='howard', frame_time='side
...
+
+    ========= ===== ====== ====== ==========
+    Model     A     B      C      Unit
+    ========= ===== ====== ====== ==========
+    howard    2.894 -0.428 -0.370 microrad/s
+    snodgrass 2.851 -0.343 -0.474 microrad/s
+    allen     14.44 -3.0   0      deg/day
+    ========= ===== ====== ====== ==========
+
+    1 mirorad/s is approximately 4.95 deg/day.
```
ayshih, line 72:
```suggestion
    1 microrad/s is approximately 4.95 deg/day.
```

ayshih (commented):
Can you also tweak the code examples in the `diff_rot()` docstring so that each actually shows output instead of just assigning it to a variable?

ayshih (approved):
Minor fix for line length, and might as well add a `docfix` changelog entry

wafels:
I don't have a problem removing the reference to the IDL code as this is a hangover from the early days of SunPy.  The better references are the papers.  Explicitly writing out the function used in these references is also an improvement.

nabobalis (changes requested):
Changelog just missing.

nabobalis (approved):

PR #4879, opened 2021-01-18
Added longitude argument and broadcasting capabilites to carrington_rotation_time()

### LINKED ISSUE #4821: Add a way of specifying rotation number and longitude to carrington_rotation_time
Currently `carrington_rotation_time` accepts a floating point number as input, which is the fractional Carrington rotation number. However, calculating this number from a rotation number and Carrington longitude is a bit confusing, because Carrington longitudes run backwards in time, going from 360 > 0. e.g.

```
2210, 90 deg > 2210.75
2210, 180 deg > 2210.5
2210, 270 deg > 2210.25
```
(note that in the order written, these numbers/coordinates are going backwards in time)

It would be nice if the API somehow allowed one to input rotation number and Longitude, so it was easy to input these values directly read from a synoptic map (e.g. https://sdo.gsfc.nasa.gov/data/synoptic/).

### LINKED ISSUE #4863: Document that carrington_rotation_time accepts arrays as input
https://docs.sunpy.org/en/stable/api/sunpy.coordinates.sun.carrington_rotation_time.html currently just says it accepts a `float` as input, so I thought I would have to loop through my array and call it one value at a time. I tried it with an array though, and it seemed to work fine, so we should document that arrays are also a valid input.

In addition, if there isn't already a test that uses this function with an array input, we should add one.

kc611:
### Description

Fixes #4821
Fixes #4863 

Added the longitude feature as requested in #4821. Also in #4863 it was mentioned that this function also works with Numpy arrays so I added a kind of "broadcasting" capability so that we can calculate multiple rotation times at once.

Would appreciate feedback before adding tests for this.

### DIFF
--- a/changelog/4879.feature.rst
+++ b/changelog/4879.feature.rst
@@ -1,0 +1,1 @@
+Added a ``longitude`` keyword argument to :func:`~sunpy.sun.carrington_rotation_time` as an alternate way to specify a fractional Carrington rotation.
--- a/changelog/4879.feature.rst
+++ b/changelog/4879.feature.rst
@@ -1,1 +1,1 @@
-Added a ``longitude`` keyword argument to :func:`~sunpy.sun.carrington_rotation_time` as an alternate way to specify a fractional Carrington rotation.
+Added a ``longitude`` keyword argument to :func:`~sunpy.coordinates.sun.carrington_rotation_time` as an alternate way to specify a fractional Carrington rotation.
--- a/sunpy/coordinates/sun.py
+++ b/sunpy/coordinates/sun.py
@@ -118,6 +118,17 @@ def carrington_rotation_time(crot, longitude: u.deg = None):
     Returns
     -------
     `astropy.time.Time`
+
+    Examples
+    --------
+    >>> from sunpy.coordinates.sun import carrington_rotation_time
+    >>> import astropy.units as u
+    >>> carrington_rotation_time(2242)
+    <Time object: scale='utc' format='iso' value=2021-03-17 22:31:37.030>
+    >>> carrington_rotation_time(2000.25)
+    <Time object: scale='utc' format='iso' value=2003-02-27 02:52:57.315>
+    >>> carrington_rotation_time(2000, 270*u.deg)
+    <Time object: scale='utc' format='iso' value=2003-02-27 02:52:57.315>
     """
     crot = crot << u.one
     if longitude is not None:
--- a/sunpy/coordinates/sun.py
+++ b/sunpy/coordinates/sun.py
@@ -22,6 +22,7 @@ from astropy.coordinates.builtin_frames.utils import get_jd12
 from astropy.coordinates.representation import CartesianRepresentation, SphericalRepresentation
 # Import erfa via astropy to make sure we are using the same ERFA library as Astropy
 from astropy.coordinates.sky_coordinate import erfa
+from astropy.tests.helper import quantity_allclose
 from astropy.time import Time
 
 from sunpy import log
@@ -91,23 +92,40 @@ def sky_position(t='now', equinox_of_date=True):
     return ra, dec
 
 
-def carrington_rotation_time(crot):
+@u.quantity_input
+def carrington_rotation_time(crot: u.one, longitude: u.deg = None):
     """
     Return the time of a given Carrington rotation.
 
+    Fractional Carrington rotation numbers can be provided in two ways:
+    * Fractional numbers to ``crot``
+    * Integer numbers to ``crot`` and a Carrington longitude to ``longitude``
+
+    Inputs can be arrays.  If both ``crot`` and ``longitude`` are provided, the
+    output shape will be the broadcasted combination.
     The round-trip from this method to `carrington_rotation_number` has
     absolute errors of < 0.11 seconds.
 
     Parameters
     ----------
-    crot : float
-        Carrington rotation number. Can be a fractional cycle number.
+    crot : `~astropy.units.Quantity`
+        Carrington rotation number(s). Can be a fractional rotation number.
+
+    longitude : `~astropy.units.Quantity`
+        Carrington longitude(s), which must be > 0 degrees and <= 360 degrees.
+        If provided, ``crot`` must be strictly integral.
 
     Returns
     -------
-    astropy.time.Time
+    `astropy.time.Time`
     """
-    estimate = (constants.mean_synodic_period.to_value('day') *
+    if longitude is not None:
+        if not quantity_allclose(crot%1, 0):
+            raise ValueError("Carrington rotation number(s) must be integral if `longitude` is provided.")
+        if (longitude <= 0*u.deg).any() or (longitude > 360*u.deg).any():
+            raise ValueError("Carrington longitude(s) must be > 0 degrees and <= 360 degrees.")
+        crot = crot + (1 - longitude/(360*u.deg))
+    estimate = (constants.mean_synodic_period *
                 (crot - 1)) + constants.first_carrington_rotation
 
     # The above estimate is inaccurate (see comments below in carrington_rotation_number),
--- a/sunpy/coordinates/sun.py
+++ b/sunpy/coordinates/sun.py
@@ -93,7 +93,7 @@ def sky_position(t='now', equinox_of_date=True):
 
 
 @u.quantity_input
-def carrington_rotation_time(crot: u.one, longitude: u.deg = None):
+def carrington_rotation_time(crot, longitude: u.deg = None):
     """
     Return the time of a given Carrington rotation.
 
@@ -108,7 +108,7 @@ def carrington_rotation_time(crot: u.one, longitude: u.deg = None):
 
     Parameters
     ----------
-    crot : `~astropy.units.Quantity`
+    crot : `int`, `float`, `~astropy.units.Quantity`
         Carrington rotation number(s). Can be a fractional rotation number.
 
     longitude : `~astropy.units.Quantity`
@@ -119,6 +119,7 @@ def carrington_rotation_time(crot: u.one, longitude: u.deg = None):
     -------
     `astropy.time.Time`
     """
+    crot = crot << u.one
     if longitude is not None:
         if not quantity_allclose(crot%1, 0):
             raise ValueError("Carrington rotation number(s) must be integral if `longitude` is provided.")
--- a/sunpy/coordinates/tests/test_sun.py
+++ b/sunpy/coordinates/tests/test_sun.py
@@ -1,5 +1,6 @@
 import warnings
 
+import numpy as np
 import pytest
 from numpy.testing import assert_allclose
 
@@ -470,6 +471,39 @@ def test_carrington_rotation_starttime(crot, julian_days):
         warnings.filterwarnings("ignore", category=ErfaWarning)
         assert_quantity_allclose(sun.carrington_rotation_time(crot).tt.jd * u.day,
                                  julian_days * u.day, atol=0.11*u.s)
+
+
+@pytest.mark.parametrize("crot, longitude, crot_fractional",
+                         [(2000, 360, 2000.0),
+                          (2000.0, 270, 2000.25)
+                          ])
+def test_carrington_rotation_time_longitude(crot, longitude, crot_fractional):
+    assert sun.carrington_rotation_time(crot*u.one, longitude*u.deg) == \
+        sun.carrington_rotation_time(crot_fractional*u.one)
+
+
+@pytest.mark.parametrize("crot, longitude, crot_fractional",
+                         [
+                             (np.array([2000, 2000]), np.array(
+                                 [180, 90]), np.array([2000.5, 2000.75])),
+                             (2000, np.array([180, 90]), np.array([2000.5, 2000.75])),
+                             (np.array([2000, 2000]), 180, np.array([2000.5, 2000.5]))
+                         ])
+def test_carrington_rotation_time_longitude_numpy(crot, longitude, crot_fractional):
+    assert all(sun.carrington_rotation_time(crot*u.one, longitude*u.deg) ==
+               sun.carrington_rotation_time(crot_fractional*u.one))
+
+
+@pytest.mark.parametrize("crot, longitude",
+                         [
+                             (2000, 0),
+                             (2000, -10),
+                             (2000, 400),
+                             (2000.5, 180),
+                         ])
+def test_carrington_rotation_time_longitude_err(crot, longitude):
+    with pytest.raises(ValueError):
+        sun.carrington_rotation_time(crot*u.one, longitude*u.deg)
 
 
 def test_carrington_rotation_roundtrip():


dstansby (changes requested):
Thanks for the PR, and welcome to sunpy! The approach overall looks good, but the following things need changing:
- `longitude` should be given as a an `astropy.units.Quantity`, see https://docs.astropy.org/en/stable/units/quantity.html?highlight=quantity_input#functions-that-accept-quantities for more information on how to do this
- For a new paragraph in the documentation strings, use a blank line instead of an indentation
- If `longitude` is given, we should check that `crot` is an integer (or an integer array), and error if it isn't
- A test needs to be added to check that the function is returning the right values when longitude is given. The tests for this code live in `sunpy/coordinates/tests/test_sun.py`. If you have any questions about this, feel free to ask!

kc611:
> A test needs to be added to check that the function is returning the right values when longitude is given.

Can these values be simply hard-coded ? Also should tests be also added for checking whether it throws errors correctly ?

( I'll add these once the main function has been finalized. )

dstansby:
Code is looking good! One more thing I noticed is that there should be an error if `longitude` is not in the range 0 to 360 degrees.

For the tests, I think it would be good to add:
- A check that passing a floating point rotation number and `longitude` raises an error
- A check that passing `longitude=360 * u.deg` returns the same result as not passing longitude
- A check that passing `crot=1, longitude=360*u.deg` gives the same result as `crot=2`
- A check that passing `longitude=400*u.deg` or `longitude=-10*u.deg` raises an error

ayshih:
> - A check that passing `longitude=360 * u.deg` returns the same result as not passing longitude
> - A check that passing `crot=1, longitude=360*u.deg` gives the same result as `crot=2`

Presumably you mean `0*u.deg` for the second check.  The fact that `0*u.deg` and `360*u.deg` are expected to give different answers is a bit disconcerting; for example, one wouldn't be able to use the `Longitude` class.

Other comments:
- The docstring needs to explain this mode of providing the `longtiude` argument before actually getting to the parameter list.
- The type check for the rotation number is a bit odd.  We're not going to accept a float with zero fractional part?
- The broadcasting code looks clunky and not sufficiently general, and I'm not sure why it's even needed.  Why isn't the broadcasting already taken care of automatically in the arithmetic expression with `crot` and `longitude`?

sunpy/coordinates/sun.py
```diff
@@ -100,13 +100,35 @@ def carrington_rotation_time(crot):
...
+        assert np.issubdtype(crot.dtype, int), "Carrington Number(s) should be of type int"
+        longitude = np.array(longitude.to('deg').value)
+        if crot.shape != longitude.shape:
+            if not longitude.shape or longitude.shape[0] == 1:
+                longitude = np.full_like(crot, longitude)
+            elif not crot.shape or crot.shape[0] == 1:
+                crot = np.full_like(longitude, crot)
+            else:
+                raise Exception("Array of rotation number and longitudes are of unequal length")
+        crot = crot + (1 - longitude/360)
```
ayshih, line 131:
```suggestion
        crot = crot + (1 - longitude/(360*u.deg))
```

dstansby:
> Presumably you mean `0*u.deg` for the second check. The fact that `0*u.deg` and `360*u.deg` are expected to give different answers is a bit disconcerting; for example, one wouldn't be able to use the `Longitude` class.

Yes, thanks for catching, I've edited it now. They're different in the sense that 00:00 and 24:00 are different times (the beginning and end of a day), but they are both "midnight". To reduce confusion, I'd suggest that we limit the longitude values to (0, 360], ie. make the interval open on the left and disallow 0 degrees input.

sunpy/coordinates/sun.py
kc611: This change raises an error since `Quantity` (even dimensionless ones) cannot be subtracted from `int` objects, `1` in this case. ( Even if the above line which converts it into `longitude` into a numpy array is removed ). I think we should rather avoid this since `crot` would be passed ahead as `Quantity` rather than a numpy array even if we fix this.

kc611:
@dstansby @ayshih Sorry for late response. I think the latest commit addresses all the points highlighted by you.

changelog/4879.feature.rst
```diff
@@ -0,0 +1 @@
+Added longitude argument and broadcasting capabilites to carrington_rotation_time.
```
ayshih, line 1:
```suggestion
Added a ``longitude`` keyword argument to :func:`~sunpy.sun.carrington_rotation_time` as an alternate way to specify a fractional Carrington rotation.
```

sunpy/coordinates/sun.py
```diff
@@ -91,7 +91,7 @@ def sky_position(t='now', equinox_of_date=True):
     return ra, dec
 
 
-def carrington_rotation_time(crot):
+def carrington_rotation_time(crot, longitude=None):
```
ayshih, line 94:
```suggestion
@u.quantity_input
def carrington_rotation_time(crot: u.one, longitude: u.deg = None):
```

sunpy/coordinates/sun.py
```diff
@@ -100,13 +100,30 @@ def carrington_rotation_time(crot):
 
     Parameters
     ----------
-    crot : float
-        Carrington rotation number. Can be a fractional cycle number.
+    crot : int, float, numpy.array
+        Carrington rotation number. Can be a integral or fractional cycle number
+        or a Numpy array of such Carrington rotation numbers. If a single
+        Carrington rotation number is passed with multiple logitudes then
+        rotation times at each of those longitudes is calculated.
```
ayshih, line 107:
```suggestion
    crot : `~astropy.units.Quantity`
        Carrington rotation number(s). Can be a fractional rotation number.
```

sunpy/coordinates/sun.py
```diff
@@ -100,13 +100,30 @@ def carrington_rotation_time(crot):
...
+
+    longitude : astropy.units.Quantity
+        Carrington longitude (in degrees).If a single longitude is passed with
+        multiple Carrington rotation numbers then the same rotation is applied
+        to each Carrington rotation number while calculating rotation times.
+        When this argument is passed Carrington rotation number(s) should be
+        strictly integral.
```
ayshih, line 114:
```suggestion
    longitude : `~astropy.units.Quantity`
        Carrington longitude(s), which must be > 0 degrees and <= 360 degrees.
        If provided, ``crot`` must be strictly integral.
```

sunpy/coordinates/sun.py
```diff
@@ -100,13 +100,30 @@ def carrington_rotation_time(crot):
...
 
     Returns
     -------
     astropy.time.Time
```
ayshih, line 118:
```suggestion
    `astropy.time.Time`
```

sunpy/coordinates/sun.py
```diff
@@ -100,13 +100,30 @@ def carrington_rotation_time(crot):
...
     """
+    if longitude is not None:
+        crot = np.array(crot)
+        longitude = np.array(longitude.to('deg').value)
```
ayshih, line 122:
```suggestion
```

sunpy/coordinates/sun.py
```diff
@@ -100,13 +100,30 @@ def carrington_rotation_time(crot):
...
+        assert not (crot % 1).all(), "Carrington Number(s) should be integral"
+        assert (longitude > 0).all() and (longitude <= 360).all(), \
+            "Carrington longitude(s) should be positive and/or less than 360"
+        crot = crot + (1 - longitude/360)
     estimate = (constants.mean_synodic_period.to_value('day') *
```
ayshih, line 127:
```suggestion
    estimate = (constants.mean_synodic_period *
```

sunpy/coordinates/sun.py
```diff
@@ -91,7 +91,7 @@ def sky_position(t='now', equinox_of_date=True):
     return ra, dec
 
 
-def carrington_rotation_time(crot):
+def carrington_rotation_time(crot, longitude=None):
     """
     Return the time of a given Carrington rotation.
```
ayshih, line 97:
```suggestion

    Fractional Carrington rotation numbers can be provided in two ways:
    * Fractional numbers to ``crot``
    * Integer numbers to ``crot`` and a Carrington longitude to ``longitude``
      
    Inputs can be arrays.  If both ``crot`` and ``longitude`` are provided, the
    output shape will be the broadcasted combination.
```

sunpy/coordinates/sun.py
```diff
@@ -100,13 +100,30 @@ def carrington_rotation_time(crot):
...
 
     Returns
     -------
     astropy.time.Time
     """
+    if longitude is not None:
+        crot = np.array(crot)
+        longitude = np.array(longitude.to('deg').value)
+        assert not (crot % 1).all(), "Carrington Number(s) should be integral"
+        assert (longitude > 0).all() and (longitude <= 360).all(), \
```
ayshih, line 124: These should not be `assert` checks, but rather `quantity_allclose()` checks that raise a `ValueError` when needed

sunpy/coordinates/sun.py
ayshih: You should switch to using `Quantity` everywhere

sunpy/coordinates/tests/test_sun.py
```diff
@@ -472,6 +473,39 @@ def test_carrington_rotation_starttime(crot, julian_days):
...
+
+@pytest.mark.parametrize("crot, longitude",
+                         [
+                             (2000, 0),
+                             (2000, -10),
+                             (2000, 400),
+                             (2000.5, 180),
+                         ])
+def test_carrington_rotation_time_longitude_err(crot, longitude):
+    with pytest.raises(AssertionError):
```
ayshih, line 505: The function should raise `ValueError`

ayshih (changes requested):
My suggestions make this function properly unit-aware

sunpy/coordinates/sun.py
kc611: It's kinda hard to formulate the second condition into an `quantity_allclose()` so I let it be. Did you have some kind of formulation in mind for that?

sunpy/coordinates/sun.py
```diff
@@ -91,23 +92,44 @@ def sky_position(t='now', equinox_of_date=True):
...
-    astropy.time.Time
-    """
-    estimate = (constants.mean_synodic_period.to_value('day') *
+    `astropy.time.Time`
+    """
+    if longitude is not None:
+        try:
+            quantity_allclose(crot%1, np.zeros_like(crot))
+        except AssertionError:
+            raise ValueError("Carrington Number(s) should be integral")
```
ayshih, line 126:
```suggestion
        if not quantity_allclose(crot%1, 0):
            raise ValueError("Carrington rotation number(s) must be integral if `longitude` is provided.")
```

sunpy/coordinates/sun.py
```diff
@@ -91,23 +92,44 @@ def sky_position(t='now', equinox_of_date=True):
...
+        try:
+            assert (longitude > 0).all() and (longitude <= 360).all()
+        except AssertionError:
+            raise ValueError("Carrington longitude(s) should be positive and/or less than 360")
```
ayshih, line 130:
```suggestion
        if (longitude <= 0*u.deg).any() or (longitude > 360*u.deg).any():
            raise ValueError("Carrington longitude(s) must be > 0 degrees and <= 360 degrees.")
```

ayshih (changes requested):

sunpy/coordinates/sun.py
ayshih: No, you're right: that doesn't need to use `quantity_allclose()`

sunpy/coordinates/tests/test_sun.py
```diff
@@ -472,6 +473,39 @@ def test_carrington_rotation_starttime(crot, julian_days):
                                  julian_days * u.day, atol=0.11*u.s)
 
 
+@pytest.mark.parametrize("crot, longitude, crot_fractional",
+                         [(2000, 360, 2000.0),
+                          (2000.0, 270, 2000.25)
+                          ])
+def test_carrington_rotation_time_longitude(crot, longitude, crot_fractional):
+    assert sun.carrington_rotation_time(crot*u, longitude*u.deg) == \
```
ayshih, line 481:
```suggestion
    assert sun.carrington_rotation_time(crot, longitude*u.deg) == \
```

ayshih (changes requested):

dstansby (approved):
I just have a small non-blocking question, but otherwise this looks great, thanks a lot!

sunpy/coordinates/sun.py
```diff
@@ -91,23 +92,40 @@ def sky_position(t='now', equinox_of_date=True):
     return ra, dec
 
 
-def carrington_rotation_time(crot):
+@u.quantity_input
+def carrington_rotation_time(crot: u.one, longitude: u.deg = None):
```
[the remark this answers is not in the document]
ayshih: Fair point, let's tweak it
```suggestion
def carrington_rotation_time(crot, longitude: u.deg = None):
```

sunpy/coordinates/sun.py
```diff
@@ -91,23 +92,40 @@ def sky_position(t='now', equinox_of_date=True):
...
+    Inputs can be arrays.  If both ``crot`` and ``longitude`` are provided, the
+    output shape will be the broadcasted combination.
     The round-trip from this method to `carrington_rotation_number` has
     absolute errors of < 0.11 seconds.
 
     Parameters
     ----------
-    crot : float
-        Carrington rotation number. Can be a fractional cycle number.
+    crot : `~astropy.units.Quantity`
```
ayshih, line 111:
```suggestion
    crot : `int`, `float`, `~astropy.units.Quantity`
```

sunpy/coordinates/sun.py
```diff
@@ -91,23 +92,40 @@ def sky_position(t='now', equinox_of_date=True):
...
+        If provided, ``crot`` must be strictly integral.
 
     Returns
     -------
-    astropy.time.Time
-    """
-    estimate = (constants.mean_synodic_period.to_value('day') *
+    `astropy.time.Time`
+    """
+    if longitude is not None:
```
ayshih, line 122:
```suggestion
    crot = crot << u.one
    if longitude is not None:
```

ayshih (changes requested):

sunpy/coordinates/sun.py
```diff
@@ -91,23 +92,41 @@ def sky_position(t='now', equinox_of_date=True):
...
     Return the time of a given Carrington rotation.
 
+    Fractional Carrington rotation numbers can be provided in two ways:
+    * Fractional numbers to ``crot``
+    * Integer numbers to ``crot`` and a Carrington longitude to ``longitude``
+
+    Inputs can be arrays.  If both ``crot`` and ``longitude`` are provided, the
+    output shape will be the broadcasted combination.
     The round-trip from this method to `carrington_rotation_number` has
     absolute errors of < 0.11 seconds.
```
nabobalis, line 108: Is it worth adding an example to the docstring?

sunpy/coordinates/sun.py
```diff
@@ -91,23 +92,41 @@ def sky_position(t='now', equinox_of_date=True):
...
+        Carrington longitude(s), which must be > 0 degrees and <= 360 degrees.
+        If provided, ``crot`` must be strictly integral.
 
     Returns
     -------
-    astropy.time.Time
-    """
-    estimate = (constants.mean_synodic_period.to_value('day') *
+    `astropy.time.Time`
+    """
```
ayshih, line 121:
```suggestion
    `astropy.time.Time`

    Examples
    --------
    >>> from sunpy.coordinates.sun import carrington_rotation_time
    >>> import astropy.units as u
    >>> carrington_rotation_time(2242)
    <Time object: scale='utc' format='iso' value=2021-03-17 22:31:37.030>
    >>> carrington_rotation_time(2000.25)
    <Time object: scale='utc' format='iso' value=2003-02-27 02:52:57.315>
    >>> carrington_rotation_time(2000, 270*u.deg)
    <Time object: scale='utc' format='iso' value=2003-02-27 02:52:57.315>
    """
```

changelog/4879.feature.rst
```diff
@@ -0,0 +1 @@
+Added a ``longitude`` keyword argument to :func:`~sunpy.sun.carrington_rotation_time` as an alternate way to specify a fractional Carrington rotation.
```
ayshih, line 1:
```suggestion
Added a ``longitude`` keyword argument to :func:`~sunpy.coordinates.sun.carrington_rotation_time` as an alternate way to specify a fractional Carrington rotation.
```

ayshih (changes requested):

ayshih (approved):
This looks good to me, but it'd be prudent to rebase to see if the CI sees any error from the past ~200 commits

nabobalis:
Thanks for the PR @kc611!

PR #4881, opened 2021-01-18
Handle time scale FITS metadata in map.date

dstansby:
Section 9.2.1 https://fits.gsfc.nasa.gov/standard40/fits_standard40aa-le.pdf says there is a TIMESYS keyword, that can be set to denote the time scale of times in a FITS header. This PR:

- Sets the time scale from TIMESYS if present
- Explicitly sets the timescale to 'utc', which is the mandated default, if TIMESYS isn't present.

It looks like the timescale was implicitly set to UTC anyway by default (either by us or astropy?), so this shouldn't be a big issue, but we should still support alternative time scales from the FITS header.

### DIFF
--- a/changelog/4881.bugfix.rst
+++ b/changelog/4881.bugfix.rst
@@ -1,0 +1,4 @@
+`sunpy.map.GenericMap.date` now has its time scale set from the 'TIMESYS' FITS keyword,
+if it is present. If it isn't present the time scale defaults to 'utc', which is unchanged
+default behaviour, so this change will only affect maps with a 'TIMESYS' keyword
+that is not set to 'UTC'.
--- a/changelog/4881.bugfix.rst
+++ b/changelog/4881.bugfix.rst
@@ -1,4 +1,4 @@
 `sunpy.map.GenericMap.date` now has its time scale set from the 'TIMESYS' FITS keyword,
-if it is present. If it isn't present the time scale defaults to 'utc', which is unchanged
+if it is present. If it isn't present the time scale defaults to 'UTC', which is unchanged
 default behaviour, so this change will only affect maps with a 'TIMESYS' keyword
 that is not set to 'UTC'.
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -666,7 +666,7 @@ class GenericMap(NDData):
             timesys_meta = self.meta.get('timesys', '').upper()
             if timesys_meta != 'TAI':
                 warnings.warn('Found "TAI" in time string, ignoring TIMESYS keyword '
-                              f'which is set to "{timesys_meta}".')
+                              f'which is set to "{timesys_meta}".', SunpyUserWarning)
         else:
             # UTC is the FITS standard default
             timesys = self.meta.get('timesys', 'UTC')
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -657,6 +657,20 @@ class GenericMap(NDData):
         This is taken from the 'DATE-OBS' FITS keyword.
         """
         time = self.meta.get('date-obs', None)
+        # Get the time scale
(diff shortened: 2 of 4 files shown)


sunpy/map/mapbase.py
```diff
@@ -657,6 +657,16 @@ def date(self):
         This is taken from the 'DATE-OBS' FITS keyword.
         """
         time = self.meta.get('date-obs', None)
+        # Get the time scale if present
+        # SDO specifies the 'TAI' scale in their time string, so detect that
+        # and special case it (even though it's probably not FITS compliant)
+        if time is not None and 'TAI' in time:
+            timesys = 'TAI'
+            if 'timesys' in self.meta:
+                warnings.warn('Found "TAI" in time string, ignoring TIMESYS keyword')
```
Cadair, line 666: shouldn't we special case this in the sdo source?

dstansby:
Note to self (discussed on call): we should fix the funny TAI date in SDO map source, and deprecate TAI parsing in `parse_time`.

Cadair:
I am not sure we should deprecate it in `parse_time`, but not relying on it in Map would be good given the other changes in here.

sunpy/map/mapbase.py
Cadair, line 666:
```suggestion
                warnings.warn('Found "TAI" in time string, ignoring TIMESYS keyword', SunpyUserWarning)
```

wtbarnes (approved):

Cadair (approved):

sunpy/map/mapbase.py
```diff
@@ -657,6 +657,20 @@ def date(self):
...
+        # Get the time scale
+        if time is not None and 'TAI' in time:
+            # SDO specifies the 'TAI' scale in their time string, which is parsed
+            # by parse_time(). If a different timescale is also present, warn the
+            # user that it will be ignored.
+            timesys = 'TAI'
+            timesys_meta = self.meta.get('timesys', '').upper()
+            if timesys_meta != 'TAI':
+                warnings.warn('Found "TAI" in time string, ignoring TIMESYS keyword '
+                              f'which is set to "{timesys_meta}".')
```
Cadair, line 669:
```suggestion
                              f'which is set to "{timesys_meta}".', SunpyUserWarning)
```

changelog/4881.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+`sunpy.map.GenericMap.date` now has its time scale set from the 'TIMESYS' FITS keyword,
+if it is present. If it isn't present the time scale defaults to 'utc', which is unchanged
```
Cadair, line 2:
```suggestion
if it is present. If it isn't present the time scale defaults to 'UTC', which is unchanged
```

Cadair (approved):

sunpy/map/mapbase.py
```diff
@@ -657,6 +657,20 @@ def date(self):
         This is taken from the 'DATE-OBS' FITS keyword.
         """
         time = self.meta.get('date-obs', None)
+        # Get the time scale
+        if time is not None and 'TAI' in time:
+            # SDO specifies the 'TAI' scale in their time string, which is parsed
```
nabobalis, line 662: Stupid question, why was this not moved to the SDO map source? I can't find a comment or any reference.

nabobalis (approved):

sunpy/map/mapbase.py
dstansby: Why was what not moved to the map source?

nabobalis: This code deals with SDO specific time handing no?

dstansby: Just in case there is other data out there that has this TAI timestamp, I think it's worth keeping it in here.

PR #4882, opened 2021-01-18
Add support for TIMEUNIT FITS keyword

dstansby:
https://fits.gsfc.nasa.gov/standard40/fits_standard40aa-le.pdf section 9.7 says that the unit for exposure time can be specified using the TIMEUNIT keyword. Add support for this, falling back on the mandated default of seconds (which is what sunpy previously always assumed).

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -677,13 +677,23 @@ class GenericMap(NDData):
         return self.meta.get('detector', "")
 
     @property
+    def timeunit(self):
+        """
+        The `~astropy.units.Unit` of the exposure time of this observation.
+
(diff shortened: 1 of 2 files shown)


wtbarnes:
I'm all for this being the default unit for exposure time but if that's it's only purpose why do we need a separate property for it? Isn't providing it as the unit on exposure time sufficient?

dstansby:
In the FITS standard it's used in a few more places to, e.g. for the units of TIMEDEL, TSTART, TSTOP, TELAPSE, so I thought it might be worth having a property, and I was kind of following the philosophy of `.waveunit` existing. Happy to remove though.

hayesla (approved):

Cadair (approved):

PR #4885, opened 2021-01-20
Reduce the length of the BaseClient repr

### LINKED ISSUE #4884: Limit the number of rows in the BaseClient repr
The VSOClient repr in particular is very long, it makes sense to print it all in the `__str__` method but not in `__repr__`

nakul-shahdadpuri:
### Description

Fixes #4884
Removes the part in __repr__ method of the BaseClient class so that the method only returns a VSOClient object.

### DIFF
--- a/sunpy/net/base_client.py
+++ b/sunpy/net/base_client.py
@@ -358,8 +358,8 @@ def _print_client(client, html=False, visible_entries=None):
     lines = [class_name, dedent(client.__doc__.partition("\n\n")[0])]
     if html:
         lines = [f"<p>{line}</p>" for line in lines]
-        lines.extend(t.pformat_all(max_lines=visible_entries, show_dtype=False,
-                                   max_width=width, align="<", html=html))
+    lines.extend(t.pformat_all(max_lines=visible_entries, show_dtype=False,
+                               max_width=width, align="<", html=html))
     return '\n'.join(lines)
 
 
@@ -423,7 +423,7 @@ class BaseClient(ABC):
         """
         Returns the normal repr plus the pretty client __str__.
         """
-        return _print_client(visible_entries=15, client=self)
+        return object.__repr__(self) + "\n" + _print_client(visible_entries=15, client=self)
 
     def __str__(self):
         """
--- a/sunpy/net/base_client.py
+++ b/sunpy/net/base_client.py
@@ -326,7 +326,7 @@ def convert_row_to_table(func):
     return wrapper
 
 
-def _print_client(client, html=False):
+def _print_client(client, html=False, visible_entries=None):
     """
     Given a BaseClient instance will print out each registered attribute.
 
@@ -358,7 +358,8 @@ def _print_client(client, html=False):
     lines = [class_name, dedent(client.__doc__.partition("\n\n")[0])]
     if html:
         lines = [f"<p>{line}</p>" for line in lines]
-    lines.extend(t.pformat_all(show_dtype=False, max_width=width, align="<", html=html))
+        lines.extend(t.pformat_all(max_lines=visible_entries, show_dtype=False,
+                                   max_width=width, align="<", html=html))
     return '\n'.join(lines)
 
 
@@ -422,19 +423,19 @@ class BaseClient(ABC):
         """
         Returns the normal repr plus the pretty client __str__.
         """
-        return object.__repr__(self) + "\n" + str(self)
+        return _print_client(visible_entries=15, client=self)
 
     def __str__(self):
         """
         This enables the "pretty" printing of BaseClient.
         """
-        return _print_client(self)
+        return _print_client(client=self)
 
     def _repr_html_(self):
         """
         This enables the "pretty" printing of the BaseClient with html.
         """
-        return _print_client(self, html=True)
+        return _print_client(visible_entries=15, client=self, html=True)
 
     @abstractmethod
     def search(self, *args, **kwargs):
--- a/sunpy/net/fido_factory.py
+++ b/sunpy/net/fido_factory.py
@@ -517,7 +517,7 @@ class UnifiedDownloaderFactory(BasicRegistrationFactory):
         return results
 
     def __repr__(self):
-        return object.__repr__(self) + "\n" + str(self)
+        return object.__repr__(self) + "\n" + self._print_clients(visible_entries=15)
 
     def __str__(self):
         """
@@ -529,9 +529,9 @@ class UnifiedDownloaderFactory(BasicRegistrationFactory):
         """
         This enables the "pretty" printing of the Fido Clients with html.
         """
-        return self._print_clients(html=True)
+        return self._print_clients(visible_entries=15, html=True)
 
-    def _print_clients(self, html=False) -> str:
+    def _print_clients(self, html=False, visible_entries=None) -> str:
         width = -1 if html else get_width()
 
         t = Table(names=["Client", "Description"], dtype=["U80", "U120"])
@@ -541,7 +541,8 @@ class UnifiedDownloaderFactory(BasicRegistrationFactory):
         for key in BaseClient._registry.keys():
             t.add_row((key.__name__, dedent(
                 key.__doc__.partition("\n\n")[0].replace("\n    ", " "))))
-        lines.extend(t.pformat_all(show_dtype=False, max_width=width, align="<", html=html))
+        lines.extend(t.pformat_all(max_lines=visible_entries,
+                                   show_dtype=False, max_width=width, align="<", html=html))
         return '\n'.join(lines)
 
 


nabobalis:
What do the repr and print statements look like now?

sunpy/net/base_client.py
```diff
@@ -206,6 +206,43 @@ def _print_client(client, html=False):
     return '\n'.join(lines)
 
 
+def _repr_client(client, html=False):
```
nabobalis, line 209: I am unsure why we need a new function, can you explain?

nakul-shahdadpuri:
@nabobalis Sorry for the late reply. I am attaching screenshots of the output.

**Print Statement**
Screenshot from 2021-01-21 02-29-43

**str**
Screenshot from 2021-01-21 02-29-55

**__repr__**
Screenshot from 2021-01-21 02-30-02

sunpy/net/base_client.py
nakul-shahdadpuri: I just thought that if in case the user wanted to display the entire table they could use _print_client and also I was not sure if _print_client was called someplace else. Should I add the visible_entries parameter in _print_client itself?

nabobalis: Why not just add a new arg to the print client function to limit the line length?

nakul-shahdadpuri: Yup, that sounds better. I'll do that.

sunpy/net/base_client.py
```diff
@@ -266,19 +266,19 @@ def __repr__(self):
...
         """
-        return object.__repr__(self) + "\n" + str(self)
+        return _print_client(visible_entries=15, client=self)
 
     def __str__(self):
         """
         This enables the "pretty" printing of BaseClient.
         """
-        return _print_client(self)
+        return _print_client(visible_entries=-1, client=self)
```
nabobalis, line 275: Should this not be None?

nakul-shahdadpuri: I used list splicing for limiting rows. -1 index just gives the last element of the list. So that it prints the entire list.

nabobalis: I think this is only correct if you index the list directly aka alist[-1]. If you slice it, it excludes the last element.

nakul-shahdadpuri: My bad. I'll just add an if statement for that block.

sunpy/net/base_client.py
```diff
@@ -266,19 +268,19 @@ def __repr__(self):
...
```
nabobalis, line 277:
```suggestion
        return _print_client(visible_entries=None, client=self)
```

sunpy/net/base_client.py
```diff
@@ -203,7 +203,9 @@ def _print_client(client, html=False):
     if html:
         lines = [f"<p>{line}</p>" for line in lines]
     lines.extend(t.pformat_all(show_dtype=False, max_width=width, align="<", html=html))
-    return '\n'.join(lines)
+    if visible_entries == -1:
+        return '\n'.join(lines)
```
nabobalis, line 207:
```suggestion
```

sunpy/net/base_client.py
```diff
@@ -170,7 +170,7 @@ def show(self, *cols):
         return self.table[valid_cols]
 
 
-def _print_client(client, html=False):
+def _print_client(visible_entries, client, html=False):
```
nabobalis, line 173:
```suggestion
def _print_client(client, visible_entries=None, html=False):
```

sunpy/net/base_client.py
```diff
@@ -266,19 +268,19 @@ def __repr__(self):
...
         """
-        return _print_client(self)
+        return _print_client(visible_entries=-1, client=self)
 
     def _repr_html_(self):
         """
         This enables the "pretty" printing of the BaseClient with html.
         """
-        return _print_client(self, html=True)
+        return _print_client(visible_entries=15, client=self, html=True)
```
nabobalis, line 283:
```suggestion
        return _print_client(visible_entries=10, client=self, html=True)
```

sunpy/net/base_client.py
```diff
@@ -266,19 +268,19 @@ def __repr__(self):
         """
         Returns the normal repr plus the pretty client __str__.
         """
-        return object.__repr__(self) + "\n" + str(self)
+        return _print_client(visible_entries=15, client=self)
```
nabobalis, line 271:
```suggestion
        return _print_client(visible_entries=10, client=self)
```

sunpy/net/base_client.py
```diff
@@ -203,7 +203,9 @@ def _print_client(client, html=False):
     if html:
         lines = [f"<p>{line}</p>" for line in lines]
     lines.extend(t.pformat_all(show_dtype=False, max_width=width, align="<", html=html))
-    return '\n'.join(lines)
+    if visible_entries is None:
+        return '\n'.join(lines)
```
nabobalis, line 207:
```suggestion
```

sunpy/net/base_client.py
```diff
@@ -203,7 +203,9 @@ def _print_client(client, html=False):
...
+    return '\n'.join(lines[0:visible_entries])
```
nabobalis, line 208:
```suggestion
    return '\n'.join(lines[slice(0:visible_entries)])
```

nakul-shahdadpuri: Hi, @nabobalis ```lines[slice(0:visible_entries)``` was giving an error so I used ```lines[slice(visible_entries)]``` instead.

nabobalis: Yeah, I made a big mistake in my suggestion.

sunpy/net/base_client.py
```diff
@@ -203,7 +203,9 @@ def _print_client(client, html=False):
     if html:
         lines = [f"<p>{line}</p>" for line in lines]
     lines.extend(t.pformat_all(show_dtype=False, max_width=width, align="<", html=html))
-    return '\n'.join(lines)
+    if visible_entries is None:
+        return '\n'.join(lines)
+    return '\n'.join(lines[slice(visible_entries)])
```
nabobalis, line 208:
```suggestion
    return '\n'.join(lines[slice(0,visible_entries)])
```

nabobalis: This should be the correct suggestion now.

nabobalis: Sorry about that.

nakul-shahdadpuri: No worries, I have made the changes. I am sorry if I bothered you too much with this pull request, I will try to be more careful from now onwards.

nabobalis: No, that's what we are here for.

sunpy/net/base_client.py
nabobalis: You don't need this anymore.

sunpy/net/base_client.py
```diff
@@ -203,7 +203,9 @@ def _print_client(client, html=False):
     if html:
         lines = [f"<p>{line}</p>" for line in lines]
     lines.extend(t.pformat_all(show_dtype=False, max_width=width, align="<", html=html))
-    return '\n'.join(lines)
+    if visible_entries is None:
+        return '\n'.join(lines)
+    return '\n'.join(lines[slice(0,visible_entries)])
```
nabobalis, line 208:
```suggestion
    return '\n'.join(lines[slice(0, visible_entries)])
```

nabobalis (approved):

sunpy/net/base_client.py
```diff
@@ -203,7 +203,7 @@ def _print_client(client, html=False):
     if html:
         lines = [f"<p>{line}</p>" for line in lines]
     lines.extend(t.pformat_all(show_dtype=False, max_width=width, align="<", html=html))
-    return '\n'.join(lines)
+    return '\n'.join(lines[slice(0, visible_entries)])
```
Cadair, line 206: I think if we truncate this output (i.e. `visible_entries is not None`) then we should append `...` to the output or similar so that it's obvious it's been truncated.

sunpy/net/base_client.py
```diff
@@ -203,7 +203,7 @@ def _print_client(client, html=False):
     if html:
         lines = [f"<p>{line}</p>" for line in lines]
     lines.extend(t.pformat_all(show_dtype=False, max_width=width, align="<", html=html))
```
Cadair, line 205: There is an argument to `pformat_all` `max_lines=` which we should use instead of truncating the results: https://docs.astropy.org/en/latest/api/astropy.table.Table.html#astropy.table.Table.pformat_all

Cadair (changes requested):
Thanks for the PR @nakul-shahdadpuri Just one small tweak I think: We should use the flag to `pformat_all` rather than truncating the output.

sunpy/net/base_client.py
nakul-shahdadpuri: Hi, @Cadair I have made the change. The latest commit used pformat_all for limiting rows.

sunpy/net/base_client.py
```diff
@@ -202,10 +202,14 @@ def _print_client(client, html=False):
     lines = [class_name, dedent(client.__doc__.partition("\n\n")[0])]
     if html:
         lines = [f"<p>{line}</p>" for line in lines]
-    lines.extend(t.pformat_all(show_dtype=False, max_width=width, align="<", html=html))
+    if visible_entries is not None:
+        lines.extend(t.pformat_all(max_lines=visible_entries, show_dtype=False, max_width=width, align="<", html=html))
+    else:
+        lines.extend(t.pformat_all(show_dtype=False, max_width=width, align="<", html=html))
```
Cadair, line 208: it's perfectly allowed for max_lines to be None, so this can just be:
```suggestion
   lines.extend(t.pformat_all(max_lines=visible_entries, show_dtype=False, max_width=width, align="<", html=html))
```

Cadair:
@nakul-shahdadpuri You can fix the code style check locally by running `tox -e style_check` it should make a lot of the changes for you by running that, and then you can commit them :smile:

nakul-shahdadpuri:
@Cadair, I was using pylint for it. I ll try it with ```tox -e style_check```. Thanks.

nabobalis:
Thanks again for the PR @nakul-shahdadpuri, I noticed we had some other places unrelated to the original issue we could change this. So I went ahead and did that.

Cadair (approved):

nabobalis:
Failures are due to Helioviewer.

PR #4891, opened 2021-01-20
Updated Great Arc

### LINKED ISSUE #3460: `GreatArc` needs to have its API clarified and documented
The class `sunpy.coordinates.utils.GreatArc` has a whole bunch of public attributes that are created as part of `__init__()` and used in other methods, but are not documented:

- `.start` and `.end` could be confusing that they are not the user-provided inputs of the same name, but rather transformed versions of those coordinates.
- `.start_frame` is particularly confusing because it is the frame of the `start` is provided by the user, not the frame of the attribute `.start`.
- `.inner_angle` is also confusing since it is so similarly named to the method `.inner_angles()`.
- `.start_cartesian`, `.center_cartesian`, `end_cartesian`, `.v1`, `.v2`, `.v3`, etc. are all intermediate calculated values.

Which of these, if any, are intended to be public API?  Anything that is public needs to be documented, and everything else should be made private.

wafels:
### Description

A first attempt to fix #3460.

This PR also allows the user to create a great circle passing through any two points on the disk, and to change the angle subtended by the arc as it passes through those points.  A helper object for coordinate visibility is also defined as it allows the user to easily calculate arc visibility properties that are useful for plotting, etc.

### DIFF
--- a/examples/plotting/great_arc_example.py
+++ b/examples/plotting/great_arc_example.py
@@ -7,14 +7,14 @@ Drawing and using a Great Arc and a Great Circle
 How to define and draw a great arc on an image of the
 Sun, and to extract intensity values along that arc.
 """
-import numpy as np
 import matplotlib.pyplot as plt
+import numpy as np
 
 import astropy.units as u
 from astropy.coordinates import SkyCoord
 
 import sunpy.map
-from sunpy.coordinates.utils import GreatArc, CoordinateVisibility
+from sunpy.coordinates.utils import CoordinateVisibility, GreatArc
 from sunpy.data.sample import AIA_171_IMAGE
 
 from sunpy.net import Fido
--- a/examples/plotting/great_arc_example.py
+++ b/examples/plotting/great_arc_example.py
@@ -1,21 +1,25 @@
 # coding: utf-8
 """
-=============================
-Drawing and using a Great Arc
-=============================
+================================================
+Drawing and using a Great Arc and a Great Circle
+================================================
 
 How to define and draw a great arc on an image of the
 Sun, and to extract intensity values along that arc.
 """
-import matplotlib.pyplot as plt
 import numpy as np
+import matplotlib.pyplot as plt
 
 import astropy.units as u
 from astropy.coordinates import SkyCoord
 
 import sunpy.map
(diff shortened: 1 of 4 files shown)


ayshih:
There's useful functionality here with `CoordinateVisibility`, but the API seems very clunky for how you make use of it.  I'm moving this PR to draft status since I think it needs a bunch of discussion.

wafels:
I think `CoordinateVisibility` as implemented is incorrect because the tangent from the observer to the edge of the solar disk does not have a heliocentric z coordinate of zero (except when the observer is at infinity).  This is essentially the same error as solved by #4524.  A solution would be to compare the solar angular radius as seen from the observer's position to the angle subtended by the coordinate(s) with the heliocentric z axis.  Coordinates having angles smaller than the solar angular radius would be observable.

wtbarnes:
I second Albert's comment regarding `CoordinateVisibility`. It would be extremely useful to have a *function* that takes in a coordinate and an observer (this could be optional and default to the observer of the coordinate, if it has one) and returns a mask of whether each point in the coordinate is visible or not.

I think this should be separated out into its own PR as its functionality is more general than just in the context of `GreatArc`. For example, when plotting fieldlines from a field extrapolation expressed in HGS, you may want to see what fieldlines are visible for a particular observer location. This  of course includes coordinates that could be off-limb.

wafels:
I've been working on a coordinate visibility function that relies on the observer property of the coordinate, returning a mask.  "Visibility" means that it is possible to draw a straight line from the observer to the coordinate without passing through the sphere of the Sun.

wtbarnes:
> I've been working on a coordinate visibility function that relies on the observer property of the coordinate, returning a mask. "Visibility" means that it is possible to draw a straight line from the observer to the coordinate without passing through the sphere of the Sun.

I've implemented something very similar, but it is rough and definitely needs to be generalized: https://github.com/wtbarnes/synthesizAR/blob/e26882a5d6fbbd57ebd94dcf1dc614f792df2501/synthesizAR/util/util.py#L22

As I said above, I think just having this as a function that returns a mask on the coordinate would be extremely useful in a lot of contexts.

PR #4899, opened 2021-01-22
Remove old _FORCE_NO_WCSAXES variable

dstansby:
I noticed that `_FORCE_NO_WCSAXES` doesn't seem to be used anywhere in the tests, so remove it as I presume it's old and uneeded? ping @Cadair who put this in 6 yrs ago in case I'm missing something.

### DIFF
--- a/sunpy/visualization/wcsaxes_compat.py
+++ b/sunpy/visualization/wcsaxes_compat.py
@@ -6,7 +6,6 @@ import matplotlib.pyplot as plt
 import astropy.units as u
 from astropy.visualization import wcsaxes
 
-
 __all__ = ["is_wcsaxes", "gca_wcs", "get_world_transform",
            "default_wcs_grid", "wcsaxes_heliographic_overlay"]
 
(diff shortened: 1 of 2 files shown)


Cadair:
@dstansby can you update this?

nabobalis (approved):

sunpy/visualization/wcsaxes_compat.py
```diff
@@ -29,10 +26,7 @@ def is_wcsaxes(axes):
     `bool`
         Result of the test.
     """
-    if not _FORCE_NO_WCSAXES:
-        return isinstance(axes, wcsaxes.WCSAxes)
-    else:
-        return False
+    return isinstance(axes, wcsaxes.WCSAxes)
```
nabobalis, line 29: Not sure we need this method anymore but we can punt down the road.

sunpy/visualization/wcsaxes_compat.py
```diff
@@ -6,9 +6,6 @@
 import astropy.units as u
 from astropy.visualization import wcsaxes
 
-# Force is put here to enable disabling all checks in this module.
-# It should only be used by tests and other such hacks.
-_FORCE_NO_WCSAXES = False
```
nabobalis, line 9:
```suggestion
```

sunpy/visualization/wcsaxes_compat.py
dstansby: Yeah, might as well wait till 3.0 since a deprecation needs to be in one LTS anyway.

PR #4912, opened 2021-01-26
Propagate units to map statistical methods

dstansby:
- This is a breaking change, but I think one worth making. Does anyone else have opinions on this?
- Unfortunately e.g. `np.mean(NDData_instance)` doesn't propagate the unit, so I had to create a quantity. I've done this in place though, so there shouldn't be (much) memory overhead
- Still needs a changelog, but in the meantime feedback welcome.

### DIFF
--- a/changelog/4912.breaking.rst
+++ b/changelog/4912.breaking.rst
@@ -1,0 +1,4 @@
+``Map`` statistical methods (`~sunpy.map.GenericMap.std`,
+`~sunpy.map.GenericMap.mean`, `~sunpy.map.GenericMap.min`,
+`~sunpy.map.GenericMap.max`) now return a `~astropy.units.Quantity`.
+To get the raw number use the ``.value`` attribute of the returned Quantity.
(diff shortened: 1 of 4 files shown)


nabobalis:
Any downsides to this? It looks good to me.

dstansby:
Don't think so, as long as it's documented that users can do `.value` to get the bare numbers back. Since this is a new feature we should probably leave it to 3.0 anyway.

wtbarnes:
The docs build failures look to be real.

dstansby:
Looks like doing `imshow` with unit-ful data has some bugs that need to be ironed out in astropy and/or matplotlib, I'll convert this to a draft and link to upstream issues.

dstansby:
See https://github.com/astropy/astropy/issues/11306 for issues using imshow+Quantity+LogNorm+colorbar.

dstansby:
I don't think this is going to be fixed upstream any time soon, so I'm going to close for now. Can always revisit in the future.

PR #4916, opened 2021-01-28
Fixed the introspection of arguments for subclasses of DataAttr

### LINKED ISSUE #4913: Add explicit signatures to Fido attribute documentation
Currently Fido attributes have a generic (`*args, **kwargs`) function signature in the docs: https://docs.sunpy.org/en/stable/code_ref/net.html#classes

It would be good to convert these to explicit function signatures, to allow users to more easily understand what arguments each attribute takes. It looks like the function signatures themselves are fine, but somehow they are being modified when sphinx is documenting them during the doc build.

ayshih:
Fixes #4913 

> Currently Fido attributes have a generic (*args, **kwargs) function signature in the docs: https://docs.sunpy.org/en/stable/code_ref/net.html#classes

> Detective work on element has uncovered that this is because `DataAttr` has a `__new__(*args, **kwargs)`, which takes precedence over the `__init__` signatures defined in the sub-classes.

This PR fixes the introspected arguments for the `Fido` attributes.  It uses `__init_subclasses__()` to create a `__new__()` in any subclass of `DataAttr`, which simply calls the inherited `__new__()`, but with its signature matching the `__init__()` signature.

### DIFF
--- a/changelog/4916.docfix.rst
+++ b/changelog/4916.docfix.rst
@@ -1,0 +1,1 @@
+Fixed the display of arguments in the documentation for `~sunpy.net.Fido` attributes (`sunpy.net.attrs`).
--- a/sunpy/net/attr.py
+++ b/sunpy/net/attr.py
@@ -14,6 +14,7 @@ Please note that & is evaluated first, so A & B | C is equivalent to
 """
 import re
 import string
+import inspect
 import keyword
 import textwrap
 from textwrap import dedent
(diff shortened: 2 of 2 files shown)


ayshih:
https://sunpy--4916.org.readthedocs.build/en/4916/code_ref/net.html#classes

nabobalis (approved):

dstansby (approved):

PR #4926, opened 2021-01-29
Don't overwrite metadata in map sources

dstansby:
If maps have metadata set, we shouldn't be overwriting it. I've been through all the map sources, and I think this fixes them all.

### DIFF
--- a/sunpy/map/sources/iris.py
+++ b/sunpy/map/sources/iris.py
@@ -42,8 +42,8 @@ class SJIMap(GenericMap):
         super().__init__(data, header, **kwargs)
 
         self.meta['detector'] = "SJI"
-        self.meta['waveunit'] = "Angstrom"
-        self.meta['wavelnth'] = header['twave1']
+        self.meta['waveunit'] = header.get('waveunit', "Angstrom")
+        self.meta['wavelnth'] = header.get('wavelnth', 'twave1')
 
     @classmethod
     def is_datasource_for(cls, data, header, **kwargs):
--- a/sunpy/map/sources/iris.py
+++ b/sunpy/map/sources/iris.py
@@ -43,7 +43,7 @@ class SJIMap(GenericMap):
 
         self.meta['detector'] = "SJI"
         self.meta['waveunit'] = header.get('waveunit', "Angstrom")
-        self.meta['wavelnth'] = header.get('wavelnth', 'twave1')
+        self.meta['wavelnth'] = header.get('wavelnth', header.get('twave1'))
 
     @classmethod
     def is_datasource_for(cls, data, header, **kwargs):
--- a/sunpy/map/sources/mlso.py
+++ b/sunpy/map/sources/mlso.py
@@ -35,10 +35,11 @@ class KCorMap(GenericMap):
         # Fill in some missing info
         self.meta['observatory'] = 'MLSO'
         self.meta['detector'] = 'KCor'
-        self.meta['waveunit'] = 'nanometer'
+        self.meta['waveunit'] = self.meta.get('waveunit', 'nanometer')
         # Since KCor is on Earth, no need to raise the warning in mapbase
-        self.meta['dsun_obs'] = (sun.earth_distance(self.date)).to(u.m).value
-        self.meta['hgln_obs'] = 0.0
+        self.meta['dsun_obs'] = self.meta.get('dsun_obs',
+                                              sun.earth_distance(self.date).to(u.m).value)
+        self.meta['hgln_obs'] = self.meta.get('hgln_obs', 0.0)
         self._nickname = self.detector
 
         self.plot_settings['cmap'] = self._get_cmap_name()
--- a/sunpy/map/sources/mlso.py
+++ b/sunpy/map/sources/mlso.py
@@ -35,7 +35,7 @@ class KCorMap(GenericMap):
         # Fill in some missing info
         self.meta['observatory'] = 'MLSO'
         self.meta['detector'] = 'KCor'
-        self.meta['waveunit'] = self.meta.get('waveunit', 'nanometer')
+        self.meta['waveunit'] = self.meta.get('waveunit', 'nm')
         # Since KCor is on Earth, no need to raise the warning in mapbase
         self.meta['dsun_obs'] = self.meta.get('dsun_obs',
                                               sun.earth_distance(self.date).to(u.m).value)
--- a/sunpy/map/sources/rhessi.py
+++ b/sunpy/map/sources/rhessi.py
@@ -46,14 +46,15 @@ class RHESSIMap(GenericMap):
         # than one image (schriste)
         # Fix some broken/misapplied keywords
         if self.meta['ctype1'] == 'arcsec':
-            self.meta['cunit1'] = 'arcsec'
-            self.meta['ctype1'] = 'HPLN-TAN'
+            self.meta['cunit1'] = self.meta.get('cunit1', 'arcsec')
+            self.meta['ctype1'] = self.meta.get('ctype1', 'HPLN-TAN')
         if self.meta['ctype2'] == 'arcsec':
-            self.meta['cunit2'] = 'arcsec'
-            self.meta['ctype2'] = 'HPLT-TAN'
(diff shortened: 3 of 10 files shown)


sunpy/map/sources/iris.py
```diff
@@ -42,8 +42,8 @@ def __init__(self, data, header, **kwargs):
         super().__init__(data, header, **kwargs)
 
         self.meta['detector'] = "SJI"
-        self.meta['waveunit'] = "Angstrom"
-        self.meta['wavelnth'] = header['twave1']
+        self.meta['waveunit'] = header.get('waveunit', "Angstrom")
+        self.meta['wavelnth'] = header.get('wavelnth', 'twave1')
```
Cadair, line 46: This is closer to correct, but could still end up with `None` in there. Although it would error currently.
```suggestion
        self.meta['wavelnth'] = header.get('wavelnth', header.get('twave1'))
```

sunpy/map/sources/mlso.py
```diff
@@ -35,10 +35,11 @@ def __init__(self, data, header, **kwargs):
         # Fill in some missing info
         self.meta['observatory'] = 'MLSO'
         self.meta['detector'] = 'KCor'
-        self.meta['waveunit'] = 'nanometer'
+        self.meta['waveunit'] = self.meta.get('waveunit', 'nanometer')
```
Cadair, line 38: Should this not be "nm"?
```suggestion
        self.meta['waveunit'] = self.meta.get('waveunit', 'nm')
```

Cadair (changes requested):
A couple of minor tweaks.

Cadair:
Test fails seem to be real.

ayshih:
The pattern `self.meta['waveunit'] = header.get('waveunit', ...)` doesn't work as intended because we apparently always set `header['WAVEUNIT']` to something as we're reading in the file, including `None`:
https://github.com/sunpy/sunpy/blob/0fcc39241395e5770ba01069fed6960f6ed6b9ae/sunpy/io/fits.py#L142
Perhaps that line can be revised to set `header['WAVEUNIT']` only if  `extract_waveunit()` returns something other than `None`?

sunpy/map/sources/rhessi.py
```diff
@@ -45,14 +45,15 @@ def __init__(self, data, header, **kwargs):
...
         if self.meta['ctype1'] == 'arcsec':
-            self.meta['cunit1'] = 'arcsec'
-            self.meta['ctype1'] = 'HPLN-TAN'
+            self.meta['cunit1'] = self.meta.get('cunit1', 'arcsec')
+            self.meta['ctype1'] = self.meta.get('ctype1', 'HPLN-TAN')
         if self.meta['ctype2'] == 'arcsec':
-            self.meta['cunit2'] = 'arcsec'
-            self.meta['ctype2'] = 'HPLT-TAN'
+            self.meta['cunit2'] = self.meta.get('cunit2', 'arcsec')
+            self.meta['ctype2'] = self.meta.get('ctype2', 'HPLT-TAN')
```
ayshih, line 52: This metadata overwriting appears to be necessary with the RHESSI headers.  `CTYPEi` mistakenly contains `CUNITi` information.

hayesla (approved):

sunpy/map/sources/rhessi.py
ayshih: If you want to be safe, you can first check whether `CTYPEi` contains "arcsec", and if so, replace with "HPLT-TAN".

ayshih (changes requested):
See earlier comments

sunpy/map/sources/rhessi.py
hayesla: ah yes. Is that true for all RHESSI image headers?

ayshih: I haven't checked

dstansby:
Thanks for the review and catching various things. Thoughts on leaving this to 3.0, where hopefully we'll have improvements to metadata handling going in anyway?

nabobalis:
This feels separate enough to warrant getting backported as bug fixes?

dstansby:
I think I'd rather this wasn't backported anywhere, as it changes how some map sources are handled away from what sunpy has done from a while, and as a user I wouldn't want to be surprised by that in a bugfix release. Obviously I've tried to be careful, but we might miss something subtle (as indeed I already did before this was reviewed).

nabobalis:
If we aren't potentially screwing up metadata that could affect someone's work in a manner, then im ok with dropping the backport.

Cadair:
> Perhaps that line can be revised to set `header['WAVEUNIT']` only if `extract_waveunit()` returns something other than `None`?

The first thing `extract_waveunit` checks is the `WAVEUNIT` key, so if waveunit is already in the header it will return the string unit representation of `WAVEUNIT`. i.e. it will convert exponents to unit strings.

ayshih:
The issue is that `extract_waveunit()` returns `None` if `WAVEUNIT` is not in the header, and the code line quoted above then *creates* `WAVEUNIT` in the header with the value `None`.  So, any map source can't tell the difference between `WAVEUNIT` being `None` because the header actually had it that way (e.g., dimensionless values) or because we inserted it.  And `header.get('waveunit', ...)` will never use the default argument.

sunpy/map/sources/yohkoh.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, data, header, **kwargs):
         # use simple trigonometry to calculate the distance of the center of
         # the Sun from the spacecraft.  Note that the small angle approximation
         # is used, and the solar radius stored in SXT FITS files is in arcseconds.
-        self.meta['dsun_apparent'] = constants.au
+        self.meta['dsun_apparent'] = self.meta.get('dsun_apparent', constants.au)
         if 'solar_r' in self.meta:
             self.meta['dsun_apparent'] = constants.radius/(np.deg2rad(self.meta['solar_r']/3600.0))
```
ayshih, line 61:
```suggestion
        if 'dsun_apparent' not in self.meta:
            if 'solar_r' in self.meta:
                self.meta['dsun_apparent'] = constants.radius/(np.deg2rad(self.meta['solar_r']/3600.0))
            else:
                self.meta['dsun_apparent'] = (sun.earth_distance(self.date)).to(u.m).value
```
And import `sun` from `sunpy.coordinates` up top.

ayshih: Some comments:

- `dsun_apparent` is a SunPy-generated keyword, so we could choose to not trust it in a file even if it's there.
- I hate the 1 AU assumption.
- I don't know whether SXT files ever don't have `solar_r`, so maybe that code branch isn't actually used.

dstansby: I think changing/fixing the dsun_apparent keyword from au to the actual sun-earth distance is a bugfix that's orthogonal to this PR.

Cadair: I have moved this to it's own issue as I want to get this PR in before 2.1.

Cadair (approved):

ayshih (changes requested):
I think [the test failure](https://dev.azure.com/sunpy/sunpy/_build/results?buildId=14170&view=logs&j=0609049e-3a2a-5217-995e-2cc283dcaa52&t=93461a68-a5a2-5d84-2961-2a4e1719748a&l=4578) is real and due to the fact that we're no longer always adding a WAVEUNIT keyword.  Fix the test ("17"-->"16"), and there should probably be a second changelog entry (bugfix?) to note this behavior change.

nabobalis:
> I think [the test failure](https://dev.azure.com/sunpy/sunpy/_build/results?buildId=14170&view=logs&j=0609049e-3a2a-5217-995e-2cc283dcaa52&t=93461a68-a5a2-5d84-2961-2a4e1719748a&l=4578) is real and due to the fact that we're no longer always adding a WAVEUNIT keyword. Fix the test ("17"-->"16"), and there should probably be a second changelog entry (bugfix?) to note this behavior change.

Fixed the test and added a basic changelog.

ayshih (approved):
Looks good

PR #4929, opened 2021-01-29
Fixed miscellaneous bugs with GenericMap.draw_rectangle()

ayshih:
This PR fixes miscellaneous bugs with `GenericMap.draw_rectangle()`:
- Removes code associated with handling positional arguments, which could potentially cause wackiness now.  Should not be backported to 2.0.
- Transforms the corners of the rectangle to the map's coordinate frame *before* determining the dimensions of the rectangle.
- Does not use `Latitude` for the "height" of the rectangle because the height is a latitude difference, which can be greater than 90 degrees.
- Specifies the color of the rectangle using `edgecolor` instead of `color` so that it can be properly overridden by the user.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1941,11 +1941,6 @@ class GenericMap(NDData):
         Extra keyword arguments to this function are passed through to the
         `~matplotlib.patches.Rectangle` instance.
         """
-        if isinstance(top_right, u.Quantity) and isinstance(width, u.Quantity):
-            # The decorator assigns the first positional arg to top_right and so on.
-            height = width
-            width = top_right
-            top_right = None
 
         bottom_left, top_right = get_rectangle_coordinates(bottom_left,
                                                            top_right=top_right,
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1941,11 +1941,13 @@ class GenericMap(NDData):
         Extra keyword arguments to this function are passed through to the
         `~matplotlib.patches.Rectangle` instance.
         """
-
         bottom_left, top_right = get_rectangle_coordinates(bottom_left,
                                                            top_right=top_right,
                                                            width=width,
                                                            height=height)
+
+        bottom_left = bottom_left.transform_to(self.coordinate_frame)
+        top_right = top_right.transform_to(self.coordinate_frame)
 
         width = Longitude(top_right.spherical.lon - bottom_left.spherical.lon)
         height = top_right.spherical.lat - bottom_left.spherical.lat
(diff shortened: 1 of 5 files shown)


ayshih:
The bullets aside from the first bullet probably ought to be (manually) backported to 2.0.

jeffreypaul15:
> This PR fixes miscellaneous bugs with `GenericMap.draw_rectangle()`:
> 
>     * Removes code associated with handling positional arguments, which could potentially cause wackiness now.  Should not be backported to 2.0.
> 
>     * Transforms the corners of the rectangle to the map's coordinate frame _before_ determining the dimensions of the rectangle.
> 
>     * Does not use `Latitude` for the "height" of the rectangle because the height is a latitude difference, which can be greater than 90 degrees.
> 
>     * Specifies the color of the rectangle using `edgecolor` instead of `color` so that it can be properly overridden by the user.

The working of this should be similar to ```draw_quadrangle``` right? Mind if I have a go at this?

sunpy/map/mapbase.py
```diff
@@ -1941,19 +1941,16 @@ def draw_rectangle(self, bottom_left, *, width: u.deg = None, height: u.deg = No
...
-            width = top_right
-            top_right = None
-
         bottom_left, top_right = get_rectangle_coordinates(bottom_left,
                                                            top_right=top_right,
                                                            width=width,
                                                            height=height)
 
+        bottom_left = bottom_left.transform_to(self.coordinate_frame)
+        top_right = top_right.transform_to(self.coordinate_frame)
```
dstansby, line 1950: This is a major behaviour change if your coordinate frame of the corners and map is different right? Even if it's a bug fix the specific change in behavour should be clearly documented in the changelog.

dstansby (changes requested):
It  would also be good to add a test with input coordinates that are in a different frame as the map they're being plotted on - I don't think this exists already?

ayshih:
> It would also be good to add a test with input coordinates that are in a different frame as the map they're being plotted on - I don't think this exists already?

It's already in this PR: a rectangle in HGS coordinates plotted on an HPC map.

sunpy/map/mapbase.py
ayshih: Well, it's a major behavior change in the sense that it's a fix of a major bug.  Previously, `bottom_left` was transformed to the map's coordinate frame only *after* the width and height of the rectangle had already been calculated, but those width and height values are used as if they are in the map's coordinate frame (e.g., a rectangle specified with a width of 10 degrees in HGS would be plotted as an rectangle with a width of 10 degrees in HPC).  This PR fixes the logic.

ayshih: I've updated the changelog

Cadair (approved):

sunpy/map/mapbase.py
dstansby, line 1950: Not for this PR, but I wonder if it would be worth adding a `coordinate_frame` argument to `get_rectangle_coordinates` so the conversion is automatically handled in that function.

dstansby (approved):

sunpy/map/mapbase.py
ayshih: I think that over-complicates `get_rectangle_coordinates()`.  Ideally, `get_rectangle_coordinates()` would return a 2-element `SkyCoord` rather than a tuple.  Then, a subsequent `.transform_to()` call would be a single line.

PR #4931, opened 2021-01-30
Added an alternate histogram and colormap shading to Map quicklook

### LINKED ISSUE #4864: Unclear which way round colourmap is in Map quicklook
I have a map with data scaled from 0 > 1, and quicklooking it in a notebook gives me the following output:

Screenshot 2021-01-12 at 13 36 00

It's not clear from looking at this (maybe I'm missing something?) whether black/white areas of the map are high/low values.

ayshih:
Closes #4864

Map `quicklook()`'s histogram of pixel values has been enhanced with shading to correspond to the colormap.  For many maps, this is not exceptionally useful since the colormap is not distributed well across a linear scale of pixel values.  I've also added a secondary histogram – switch to it by clicking on the histogram – that uses a symmetric-log scale for the pixel values.

For our stock EIT 195 image:
hist1

And if you click on the histogram:
hist2

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -345,7 +345,7 @@ class GenericMap(NDData):
 
         # Plot the CDF of the pixel values using a symmetric-log horizontal scale
         fig = Figure(figsize=(4.8, 2.4), constrained_layout=True)
-        # Figure instances in matplotlib<3.1 do not create a canvas by default
+        # TODO: Figure instances in matplotlib<3.1 do not create a canvas by default
         if fig.canvas is None:
             FigureCanvasBase(fig)
         ax = fig.subplots()
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -328,15 +328,41 @@ class GenericMap(NDData):
         if fig.canvas is None:
             FigureCanvasBase(fig)
         ax = fig.subplots()
-        ax.hist(finite_data.ravel(), bins=100, histtype='stepfilled')
+        values, bins, patches = ax.hist(finite_data.ravel(), bins=100)
+        norm_centers = norm(0.5 * (bins[:-1] + bins[1:])).data
+        for c, p in zip(norm_centers, patches):
+            plt.setp(p, "facecolor", cmap(c))
+        ax.plot(np.array([bins[:-1], bins[1:]]).T.ravel(),
+                np.array([values, values]).T.ravel())
         ax.set_facecolor('white')
         ax.semilogy()
         # Explicitly set the power limits for the X axis formatter to avoid text overlaps
         ax.xaxis.get_major_formatter().set_powerlimits((-3, 4))
-        ax.set_xlabel('Pixel value')
+        ax.set_xlabel('Pixel value (linear binning)')
         ax.set_ylabel('# of pixels')
-        ax.set_title('Distribution of pixels with finite values')
+        ax.set_title('Distribution of pixel values [click for cumulative]')
         hist_src = _figure_to_base64(fig)
+
+        # Plot the CDF of the pixel values using a symmetric-log horizontal scale
+        fig = Figure(figsize=(4.8, 2.4), constrained_layout=True)
+        # Figure instances in matplotlib<3.1 do not create a canvas by default
+        if fig.canvas is None:
+            FigureCanvasBase(fig)
+        ax = fig.subplots()
+        n_bins = 256
+        bins = norm.inverse(np.arange(n_bins + 1) / n_bins)
+        values, _, patches = ax.hist(finite_data.ravel(), bins=bins, cumulative=True)
+        for i, p in enumerate(patches):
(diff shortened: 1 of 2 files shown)


ayshih:
Here's the interactive example from the doc build: https://sunpy--4931.org.readthedocs.build/en/4931/api/sunpy.map.MapSequence.html#sunpy.map.MapSequence.quicklook

dstansby:
Whilst this looks really neat, I'm a bit worried about feature creep, especially for `quicklook()` which should be a simple non-quantitative way to quickly look at the data. My specific questions would be:

- Is there a good reason why we're not using the default colormap and norm to show the images?
- I would advocate for keeping it simple and choosing only one way to show the histogram. Could we can't use the default norm for the map as the x-axis norm in the histogram? Otherwise I'd say defualting to linear scales is reasonable as the simplest choice, unless we know a-priori that almost all data without a default norm will be better with another scale.

nabobalis:
Yeah I have to agree, I am worried about quicklook growing out of control (even tho it is very fancy).

ayshih:
> - Is there a good reason why we're not using the default colormap and norm to show the images?

Yes.  The principle of `quicklook()` (to me) is to quickly evaluate the data file, without trusting that the header values are all correct or SunPy's assumptions are appropriate (cf. MDI files).  That's also the reason for histogram equalization, which reveals structure at all pixel values, unlike a default norm, which can hide issues in the data, particularly if the data doesn't exactly conform with the assumptions behind that norm.

If you want to make a plot with the default colormap and norm, just use `plot()` (or `peek()`).  If that plot looks "off", then call `quicklook()`.

This PR is not scope creep.

dstansby:
👍 on norm/colormap

What's the purpose of the histogram in quicklook()?

I really like the colormap shading of it, agree that it's not ideal on a linear scale, but think the CDF is a bit weird on a log-scale because visually it looks like there are as many pixels below 10e-2 as between 10e0 and 10e1 in the example above.

hayesla (approved):
This looks great!

ayshih:
> What's the purpose of the histogram in quicklook()?

In my mind, the distribution of pixel values is useful for determining normalization limits and/or whether your default choices are appropriate.  Imagine that you normally plot HMI magnetograms clipped to -1500 to 1500; the histogram will show you whether that's inappropriate for the file you actually have.

It can also be useful for checking whether the file conforms to your expectations of already-performed processing.  For example, the EIT image shows a clump of negative-values pixels (presumably due to subtracting dark current from map pixels that are not on the sensor?) and a handful of high-valued pixels (presumably hot pixels on the sensor or cosmic-ray hits).  The histogram thus helps you see that you may want to submap to avoid the corners and/or run a de-spiker.

dstansby:
Thanks for the explanation. I'm still not sure how to usefully interpret the CDF, since the pixel values on the x-axis are normalised. Is there any way to change the x-axis so that the tick values are the actual data values at those tick points instead of the normalised values?

ayshih:
The X values in the CDF aren't normalized; they're the actual pixel values.  Maybe I'm misunderstanding your question?

dstansby:
I mean the x-axis in this plot:
Screenshot 2021-02-07 at 14 47 46

sunpy/map/mapbase.py
```diff
@@ -328,16 +328,42 @@ def _repr_html_(self):
...
-        ax.set_xlabel('Pixel value')
+        ax.set_xlabel('Pixel value (linear binning)')
         ax.set_ylabel('# of pixels')
-        ax.set_title('Distribution of pixels with finite values')
+        ax.set_title('Distribution of pixel values [click for cumulative]')
         hist_src = _figure_to_base64(fig)
 
+        # Plot the CDF of the pixel values using a symmetric-log horizontal scale
+        fig = Figure(figsize=(4.8, 2.4), constrained_layout=True)
+        # Figure instances in matplotlib<3.1 do not create a canvas by default
```
nabobalis, line 348:
```suggestion
        # TODO: Figure instances in matplotlib<3.1 do not create a canvas by default
```

Just so we can change this in future.

nabobalis:
> I meant the x-axis in this plot:
> Screenshot 2021-02-07 at 14 47 46
> Are these the actual values? If so great, and I just got confused

The histogram code, `ax.hist(finite_data.ravel(), bins=bins, cumulative=True)` So I would assume those are actual values.

dstansby (changes requested):
Could the "(linear binning)" and "(equalized binning)" be removed from the x-axis labels on the new histogram? They confused me, and I think they're not needed if the values are just the actual pixel values.

ayshih:
I don't want to entirely remove the description of the type of binning because it's already important information for any histogram, and it's especially important since it's different between the two histograms.  How about I change the labels to "Pixel value in linear bins" and "Pixel value in equalized bins"?

dstansby:
👍 that sounds good

ayshih:
Done

dstansby (approved):

PR #4939, opened 2021-01-30
Fix animator axes test

dstansby:
This fixes the devdeps build. It seems like the order of axes in `fig.axes` cannot be relied upon to stay the same, so instead take a more careful approach that sets the `.axes` attribute, instead of trying to return something from `fig.axes`.


nabobalis (approved):

nabobalis:
Style check issue with the WCS import.

PR #4948, opened 2021-02-03
Support reading GOES XRS daily background files

wtbarnes:
### Description

This adds support to `XRSTimeseries` for the XRS daily background files. These are described in section 7 of the [user guide](https://data.ngdc.noaa.gov/platforms/solar-space-observing-satellites/goes/goes16/l2/docs/GOES-R_XRS_L2_Data_Users_Guide.pdf) and more details are provided in appendix B of that document.

I'm not completely sure if this is a good idea or if perhaps this needs its own client or if we should even support these at all?

### DIFF
--- a/sunpy/timeseries/sources/goes.py
+++ b/sunpy/timeseries/sources/goes.py
@@ -210,7 +210,11 @@ class XRSTimeSeries(GenericTimeSeries):
                 xrsb = np.array(d["xrsb_flux"])
                 start_time_str = d["time"].attrs["units"].astype(str).lstrip("seconds since")
                 times = parse_time(start_time_str) + TimeDelta(d["time"], format="sec")
-
+            elif "bkd1d_xrsa_flux" in d.variables:
+                xrsa = np.array(d["bkd1d_xrsa_flux"])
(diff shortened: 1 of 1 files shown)


wtbarnes:
An example file for this kind of data can be found here: https://data.ngdc.noaa.gov/platforms/solar-space-observing-satellites/goes/goes17/l2/data/xrsf-l2-bkd1d_science/

And calling `peek` gives,

image

wtbarnes:
Thoughts/notes from dev meeting discussion

* At least needs a test
* GOES 13/14/15 also have background files. Are the variables named differently?
* If we support these background files, do we support the other types of files that XRS produces?
* Does it need its own source? Only very minor differences, but we should indicate somehow that these are background files in the TS object we return.
* There are other variables in these files (the normal flux ones). Should we be parsing these?

hayesla:
just after having a quick look - it seems that the GOES 13/14/15 files have the same variable names which is good

wtbarnes:
I don't think there's a clear way forward on this without having a deeper think about how to support other variables in the XRS files so I'm going to close this for now.

PR #4954, opened 2021-02-03
pin drms to 0.6.1

wtbarnes:
Pin `drms` to v0.6.1 as it fixes a bug that was introduced in 0.6.0. 

I almost forgot about this, but that little bug is annoying and potentially means downloads of files from JSOC are skipped so I think this is needed. Ideally no one should be on 0.6.0.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -60,7 +60,7 @@ map =
 net =
   beautifulsoup4
   pandas>=0.24.0
-  drms>=0.6
+  drms>=0.6.1
   python-dateutil
   tqdm
   zeep


Cadair:
we are pinning this in  2.0?

nabobalis (approved):

wtbarnes:
> we are pinning this in 2.0?

I don't see why not. If we do another bug fix release for 2.0, I think it should get in there. It manifests as a bug when using the JSOCClient.

Cadair (approved):

PR #4961, opened 2021-02-04
Make sure figures are closed after animator tests

dstansby:
I think this should fix the remote tests failure. Not sure why it's only happening on the remote tests, but it's probably good practice to close all figures after individual tests to stop them leaking.

### DIFF
--- a/sunpy/visualization/animator/tests/test_basefuncanimator.py
+++ b/sunpy/visualization/animator/tests/test_basefuncanimator.py
@@ -47,7 +47,7 @@ def test_base_func_init(fig, colorbar, buttons):
     tfa._set_active_slider(1)
     assert tfa.active_slider == 1
 
-    fig = plt.figure()
+    fig = tfa.fig
     event = mback.KeyEvent(name='key_press_event', canvas=fig.canvas, key='down')
     tfa._key_press(event)
     assert tfa.active_slider == 0


Cadair:
> to close all figures after individual tests to stop them leaking

this should be happening in pytest-mpl and if it isn't somehow then it should fixed there as well.

ayshih:
I think this test failure (or at least the failure that gets reported) is because a created figure (from `plt.figure()`) is being [passed in as a parameterized argument](https://github.com/sunpy/sunpy/blob/master/sunpy/visualization/animator/tests/test_basefuncanimator.py#L31-L32), and the test is [getting rerun](https://dev.azure.com/sunpy/sunpy/_build/results?buildId=13785&view=logs&j=edaf935d-435e-52ad-defd-fb8d072e3acf&t=2eea3e3c-5f3e-542f-03b5-1c1d34d7ce7d&l=7299).

So, of course Matplotlib raises a warning on the rerun because the same axes gets added twice to the same figure.  What I can't tell is why the test fails the first time, whether Matplotlib or something else.  I don't see this failure locally.

ayshih:
That is, the failure that I can see is not from a figure leaking, but rather a figure explicitly being re-used when the test is rerun.

dstansby:
Ah, the documentation for `plt.figure()` says:

> Create a new figure, or activate an existing figure.

So the call to `plt.figure()` in the parametrization is probably not creating a new figure, but reusing the previous one. If that's right closing the figure should work, which is the current patch.

ayshih:
I disagree.  It should only activate an existing figure if provided a number identifier:

> If there is no figure with the identifier or num is not given, a new figure is created, made active and returned.

I suspect that there is some other bug at play here that is causing the first failure.

ayshih:
Importantly, this online CI build (https://dev.azure.com/sunpy/sunpy/_build/results?buildId=13754&view=logs&s=2cb4d22b-bdb6-5735-535c-40e7906c1eb8&j=edaf935d-435e-52ad-defd-fb8d072e3acf) from yesterday morning passed, which suggests that the first failure may be intermittent in some non-obvious way.  I'd prefer us trying to isolate the failure before using the blanket solution of this PR.

ayshih:
So the first failure is [the same warning](https://dev.azure.com/sunpy/sunpy/_build/results?buildId=13815&view=logs&j=edaf935d-435e-52ad-defd-fb8d072e3acf&t=2eea3e3c-5f3e-542f-03b5-1c1d34d7ce7d&l=9106), but I have a new theory:

`plt.figure()` as the parameterized argument probably gets executed at test collection, but the test itself doesn't get run until later.  There may be an intervening `plt.gcf()` call buried in some preceding test that grabs this empty figure sitting on the stack and adds an axes to it.  By the time the test actually gets run, it will try to add an axes to a figure that already has axes.

ayshih:
Confirmed.  If you add to the top of `test_basefuncanimator.py` this silly test function:
```python
def test_call_gcf():
    fig = plt.gcf()
    fig.add_subplot(111)
```
it will run after test collection but before `test_base_func_init` is run, so it will add axes to the figure generated by `plt.figure()` for the parameterized argument before that argument is fed into `test_base_func_init`.

PR incoming

PR #4963, opened 2021-02-05
Fixed the timing of figure creation for a test

ayshih:
The lesson here is don't create a (pyplot) figure within the `pytest.mark.parametrize` decorator because it will get created during test collection, and may get mucked with by other tests (via pyplot) before it is actually passed to the test function.

See #4961 for related discussion

### DIFF
--- a/sunpy/visualization/animator/tests/test_basefuncanimator.py
+++ b/sunpy/visualization/animator/tests/test_basefuncanimator.py
@@ -29,8 +29,12 @@ def button_func1(*args, **kwargs):
 
 
 @pytest.mark.parametrize('fig, colorbar, buttons', ((None, False, [[], []]),
-                                                    (plt.figure(), True, [[button_func1], ["hi"]])))
+                                                    (plt.figure, True, [[button_func1], ["hi"]])))
 def test_base_func_init(fig, colorbar, buttons):
+    # We need to create figures within the test function rather than in parametrize()
+    if callable(fig):
+        fig = fig()
+
     data = np.random.random((3, 10, 10))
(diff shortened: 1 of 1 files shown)


ayshih:
I'll point out that `BaseFuncAnimator()` throws a fit in various ways if passed a non-pyplot figure (i.e., created via `matplotlib.figure.Figure()`).  Requiring it to accept only pyplot figures or fixing it to work with non-pyplot figures is beyond the scope of this PR.

nabobalis (approved):

sunpy/visualization/animator/tests/test_basefuncanimator.py
```diff
@@ -95,6 +99,10 @@ def test_base_func_init(fig, colorbar, buttons):
     tfa._mouse_click(event)
     assert tfa.active_slider == 0
 
+    # Close the figure if it is a pyplot figure
+    if fig in [plt.figure(i) for i in plt.get_fignums()]:
+        plt.close(fig)
```
dstansby, line 104: Why not just do a `plt.close('all')` here?

ayshih: I think it's poor form to blindly close all figures because that could hide bugs or even create new problems (if there's some legitimate reason for a figure to be present on the stack).  If you're going to change the global state, it should be to intentionally restore it to the pre-test state: for example, you should use `plt.close('all')` only if the test explicitly starts with an assertion that the pyplot stack is initially empty (e.g., `assert not plt.get_fignums()`).

PR #4968, opened 2021-02-05
Allow non-pyplot figures to be used by animators

dstansby:
This PR:
- Allows one to use a directly created figure with `BaseFuncAnimator`
- Updates the test to use a specific figure instance, thus prevent issues like the one fixed in https://github.com/sunpy/sunpy/pull/4963 Unless one specifically wants to test pyplot behaviour, it's good practice to use directly created figures.

### DIFF
--- a/changelog/4968.behaviour.rst
+++ b/changelog/4968.behaviour.rst
@@ -1,7 +1,0 @@
-The default axes used by :meth:`sunpy.visualisation.animator.BaseFuncAnimator.get_animation`
-is now ``BaseFuncAnimator.axes``, instead of the currently active axes (accessed via.
-:func:`matplotlib.pyplot.gca`). The allows animations to be created on figures
-not created via the pyplot interface.
-
-To revert to the previous behaviour of using the current axes,
-give ``axes=plt.gca()`` to ``get_animation()``.
--- a/changelog/4968.behaviour.rst
+++ b/changelog/4968.behaviour.rst
@@ -1,0 +1,7 @@
+The default axes used by :meth:`sunpy.visualisation.animator.BaseFuncAnimator.get_animation`
+is now ``BaseFuncAnimator.axes``, instead of the currently active axes (accessed via.
+:func:`matplotlib.pyplot.gca`). The allows animations to be created on figures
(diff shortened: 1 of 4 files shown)


Cadair (approved):
Looks good, but do we still have a test using plt.figure?

dstansby:
Good point - just to check I'll add a test to make sure that when a `plt.figure()` figure is passed, the current axes is correctly set.

ayshih:
Other methods in animator also assume the figure is pyplot, which is why I gave up on fixing this

ayshih:
Specifically:
https://github.com/sunpy/sunpy/blob/b08428321482dc005b7c5d5252e3e29ce58bbb49/sunpy/visualization/animator/base.py#L165
https://github.com/sunpy/sunpy/blob/b08428321482dc005b7c5d5252e3e29ce58bbb49/sunpy/visualization/animator/base.py#L208
They look straightforward to fix in isolation, but you'd need to verify that the resulting behavior overall for a non-pyplot figure is as intended.

dstansby:
It looks like the minimum version of Matplotlib we support will need bumping for this approach to work.  We're currently on 2.2.5, 3.0.0 was released in September 2018, so questions:

- Are people happy with the approach and changes made?
- If so, are people happy bumping the minimum version of Matplotlib we support in sunpy 3.0?

I think this can probably not be backported as there is a minor behaviour change (see changelog), and it is otherwise minor changes for better testing.

nabobalis:
I am happy with bumping the version for sunpy 3.0

Cadair (approved):
Looks fine, just need the CI to be happy.

Cadair:
+1 on bumping to mpl 3.0 as well, the NEP 29 rule is 24 months from scheduled release (for numpy), so 3.0 seems fine for that.

dstansby:
Hmm, even a `Matplotlib` version bump doesn't seem to be good enough for this approach, I think I will edit this at some point to use a `try: plt.gca(); except` block instead.

Cadair (approved):
Seems reasonable to me

nabobalis (approved):

PR #4969, opened 2021-02-08
Safer handling of pyplot figures by tests

ayshih:
This PR is a broader version of what @dstansby first toyed with in #4961.

- Raise a `SunpyUserWarning` (nominally converted to an error) if the pyplot figure stack is not empty prior to running a test.
- After a running a test, clear any pyplot figures that are still open, with an associated info message.  To see these messages, you can provide `-o log_cli=true` as a command-line argument for pytest.

### DIFF
--- a/changelog/4969.trivial.rst
+++ b/changelog/4969.trivial.rst
@@ -1,0 +1,1 @@
+The testing suite now raises a warning if the `~matplotlib.pyplot` figure stack is not empty prior to running a test, and it closes all open figures after finishing each test.
--- a/sunpy/conftest.py
+++ b/sunpy/conftest.py
@@ -1,4 +1,5 @@
 import os
+import logging
 import tempfile
 import importlib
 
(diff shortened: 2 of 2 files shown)


dstansby:
My instincts are that it would be best to be strict and error if any figures are left open, ie. something like
```
open_figs = plt.get_fignums()
if HAVE_MATPLOTLIB:
    open_figs = plt.get_fignums()
    if open_figs:
        plt.close('all')
        raise RuntimeError(f"There were {open_figs} pyplot figures left open by {item.name}")
```

It's probably also worth checking what the performance hit is of running these checks with every test.

ayshih:
We could be strict, but there are currently many tests that leave pyplot figures on the stack, so that'd be a slog to fix.  And besides, there's nothing particularly *wrong* with leaving figures on the stack and leaving it for a teardown to clean up.  It's just being scoped like local variables that we don't need to manually delete at the end of a function.

Qualitatively, the entire test suite does not take appreciably longer to run.

dstansby:
Sounds good for now 👍 - I'm always happy to go through and make it strict later.

nabobalis (approved):

dstansby:
Should this be visibly logging anything when the tests are run? I can't find anything showing up in the circleCI logs.

ayshih:
Not by default.  The logging info messages require a command-line option to be made visible, and the warning messages should be emitted only if something has gone wrong.

ayshih:
We could set the configuration option `log_cli=true` so that it doesn't need to be passed in by command line

ayshih:
But that could get quite noisy to have on by default since we have open figures everywhere

dstansby:
No worries, just wanted to double check 👍

dstansby (approved):

PR #4971, opened 2021-02-09
Only update wcs animator when slider index changes

dstansby:
This should increase performance when dragging the slider. Previously the update functions would have been called at many intermediate values. Now the update functions are only called when the integer index changes.

### DIFF
--- a/sunpy/visualization/animator/wcs.py
+++ b/sunpy/visualization/animator/wcs.py
@@ -213,6 +213,8 @@ class ArrayAnimatorWCS(ArrayAnimator):
         the matplotlib artist.
         """
         ind = int(val)
+        if ind == int(slider.cval):
+            return
         ax_ind = self.slider_axes[slider.slider_ind]
         self.frame_slice[ax_ind] = ind
         self.slices_wcsaxes[self.wcs.pixel_n_dim - ax_ind - 1] = ind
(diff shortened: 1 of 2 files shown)


sunpy/visualization/animator/wcs.py
```diff
@@ -223,7 +225,6 @@ def update_plot(self, val, artist, slider):
             self.update_plot_2d(val, artist, slider)
 
         self._apply_coord_params(self.axes)
-        return super().update_plot(val, artist, slider)
```
Cadair, line 226: Why the delete?

Cadair (approved):
Oh, neat!

sunpy/visualization/animator/wcs.py
dstansby: I think it always returns `None`, and is never used.

Cadair: The base class has logic in it's update_plot method: https://github.com/sunpy/sunpy/blob/master/sunpy/visualization/animator/base.py#L614 so unless I am missing something, I think this should stay.

sunpy/visualization/animator/wcs.py
```diff
@@ -203,6 +203,7 @@ def plot_start_image(self, ax):
         elif self.plot_dimensionality == 2:
             artist = self.plot_start_image_2d(ax)
 
+        self.axes.reset_wcs(wcs=self.wcs, slices=self.slices_wcsaxes)
```
Cadair, line 206: Do we need to do this? `plot_start_image` should be called on initialisation just after a fresh axes has been created so as long as `self.wcs` and `self.slices_wcsaxes` haven't been modified between construction and this method getting called we don't need this.

The `reset_wcs` method is the one which does all the setup for `WCSAxes` so it's really expensive.

dstansby: I think this does some stuff that `plot_start_image` doesn't currently do. I thought this would fix some of the image failures, but it doesn't, so I might put this change in another draft PR to see what happens and perhaps fix `plot_start_image`.

dstansby: re. performance, `reset_wcs` is called **every time** the slider is changed, so unless you're plotting and saving the figure without changing the slider I'm not sure performance is too much of an issue.

Cadair: Running through the `reset_wcs` twice on startup if it's redundant seems like a waste, but if it makes writing a `plot_start_image` method easier then it is probably worth it.

nabobalis:
We want to add a changelog for this?

dstansby:
Yes, and there's still an outstanding issue with units not being set on the sliders at the moment.

nabobalis (changes requested):
Unit issue to be fixed.

dstansby:
I've taken off the backport labels, as I think this will require the new minimum version of matplotlib.

dstansby:
I think this is good for review now

nabobalis (approved):

nabobalis:
Online fails are related to the changes.

nabobalis:
> not being set on the sliders at the moment.

What was the change that made this work again?

PR #4973, opened 2021-02-10
Update wcs when creating animator

dstansby:
In #4971 adding this line seemed to improve some of the initial images (I think mainly by correctly initialising some of the axes). It might be that the correct solution is to modify `plot_start_image` instead of this heavy handed approach, but I thought it would be good to start here to see if anything does need fixing.

### DIFF
--- a/sunpy/visualization/animator/wcs.py
+++ b/sunpy/visualization/animator/wcs.py
@@ -203,6 +203,7 @@ class ArrayAnimatorWCS(ArrayAnimator):
         elif self.plot_dimensionality == 2:
             artist = self.plot_start_image_2d(ax)
 
+        self.axes.reset_wcs(wcs=self.wcs, slices=self.slices_wcsaxes)
         return artist
 
     def update_plot(self, val, artist, slider):


Cadair:
(going to discuss this here)

The call stack is deep but [`__init__`](https://github.com/sunpy/sunpy/blob/b08428321482dc005b7c5d5252e3e29ce58bbb49/sunpy/visualization/animator/base.py#L113) calls [`make_axes_grid`](https://github.com/sunpy/sunpy/blob/b08428321482dc005b7c5d5252e3e29ce58bbb49/sunpy/visualization/animator/base.py#L261) which calls [`setup_main_axes`](https://github.com/sunpy/sunpy/blob/b08428321482dc005b7c5d5252e3e29ce58bbb49/sunpy/visualization/animator/wcs.py#L194) which both initialises the WCSAxes instance and applies the coord params. This change causes the coord param changes to be reverted to default, which is why the tests are failing.

dstansby:
I would expect a call to `reset_wcs()` to be a no-op after `plot_start_image()` though, so perhaps there's something that needs fixing? The changes to the figures look sensible: https://63225-2165383-gh.circle-artifacts.com/0/.tmp/py38-figure/figure_test_images/fig_comparison.html In particular, it looks like the y-axis is updated by calling `reset_wcs()`

dstansby:
Ignore me, I should have read the actual tests...

PR #4986, opened 2021-02-13
Experiment: Move some CI onto GitHub actions

Cadair:
This is a proposal to attempt to reduce the amount of times the CI gets backlogged on Azure by moving as many builds as is practical onto GitHub Actions.

It moves some of the phase 2 test builds onto actions, the `-oldestdeps` and `build_docs` (these two are chosen as they do not require any of the more complex features of the Azure template that I don't want to rewrite into the workflows). Also the codestyle check is moved to actions.

The issue with moving these off onto Actions is we want to prevent the release builds running if these jobs fail, so to still have this safety check, these jobs are still in the Azure config, but in a section which only gets triggered just before the wheels are built.

Finally, as a bonus new feature, I have added a second GitHub Actions workflow which is only run for PRs which have the "Extra CI" label. The jobs in this workflow are the jobs which run on the nightly cron on Azure. This is to facilitate easy debugging of these jobs in PRs.

---

This is a reasonably large jump in the complexity of our CI, we need to decide if it is worth it for the speed ups (which benefit all sponsored packages as well as core).


nabobalis:
You'll need to update the CI page if we do this.

.github/workflows/run_tox.yml
```diff
@@ -0,0 +1,62 @@
...
+            python: 3.7
+
+          - toxenv: build_docs
+            os: ubuntu-latest
+            python: 3.8
+            cov: false
+
+    name: ${{ matrix.name || format('{0} [{1}]', matrix.toxenv, matrix.os) }}
+    runs-on: ${{ matrix.os }}
+
```
Cadair, line 29: Skip these builds on not-master as we run those on Azure anyway.

.github/workflows/run_tox.yml
```diff
@@ -0,0 +1,62 @@
...
+      matrix:
+        include:
+          - toxenv: codestyle
+            os: ubuntu-latest
+            python: 3.8
+            cov: false
+
+          - toxenv: py37-oldestdeps
+            os: ubuntu-latest
+            python: 3.7
```
Cadair, line 20: Do we put this here, or leave it on Azure?

nabobalis:
I am adding Azure stages to the 2.0 CI.

nabobalis:
I think we should swing back to this if we need more CI time.

PR #4979, opened 2021-02-12
Fixes to differential rotation (both implementations)

ayshih:
So, it turns out that both `solar_rotate_coordinate()` and `RotatedSunFrame` have bugged transformations, both due to improper handling of a change in `obstime`.  The bug in the former resulted in a larger inaccuracy, but I'm more annoyed that there was a bug in the latter.  The bug in `solar_rotate_coordinate()` is why the bounds of the differentially-rotated map could be incorrect (see #4977).

There's also another bug in `sunpy.physics.differential_rotation`: it wasn't written with awareness of the translational motion of the Sun.  (To be fair, it might have been written before that was rigorously handled in `sunpy.coordinates`.)  For the use cases of `sunpy.physics.differential_rotation`, the user would want to ignore the translational motion of the Sun, so I've wrapped the relevant transformations with the `transform_with_sun_center()` context manager.

Fixes #4977
Closes #3931

Issue #4670, opened 2020-11-21
Skipped tests that generate warnings


nabobalis:
I found out in #4358 that when we ran only online tests, this meant that no doctests were collected anywhere.
This change was reverted and several doctests were fixed or skipped in order to prevent the CI from breaking.

While this will not be a full list, some of the changes made that need to be updated/fixed:

```
____________________________________________________________________ [doctest] sunpy.map.map_factory.MapFactory ____________________________________________________________________
095     `sunpy.io.header.FileHeader` or `sunpy.util.metadata.MetaDict` classes.
096 
097     * data, header pairs, not in tuples
098 
099     >>> mymap = sunpy.map.Map(data, header)   # doctest: +SKIP
100 
101     * data, wcs object, in tuple
102 
103     >>> from astropy.wcs import WCS
104     >>> wcs = WCS(sunpy.data.sample.AIA_171_ROLL_IMAGE)     # doctest: +REMOTE_DATA +IGNORE_WARNINGS
UNEXPECTED EXCEPTION: VerifyWarning('Verification reported errors:')
Traceback (most recent call last):
  File "/opt/miniconda/envs/sunpy/lib/python3.8/doctest.py", line 1336, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest sunpy.map.map_factory.MapFactory[7]>", line 1, in <module>
  File "/opt/miniconda/envs/sunpy/lib/python3.8/site-packages/astropy/wcs/wcs.py", line 425, in __init__
    header_string = header.tostring().rstrip()
  File "/opt/miniconda/envs/sunpy/lib/python3.8/site-packages/astropy/io/fits/header.py", line 684, in tostring
    s = str(card)
  File "/opt/miniconda/envs/sunpy/lib/python3.8/site-packages/astropy/io/fits/card.py", line 205, in __str__
    return self.image
  File "/opt/miniconda/envs/sunpy/lib/python3.8/site-packages/astropy/io/fits/card.py", line 521, in image
    self.verify('fix+warn')
  File "/opt/miniconda/envs/sunpy/lib/python3.8/site-packages/astropy/io/fits/verify.py", line 117, in verify
    warnings.warn(line, VerifyWarning)
astropy.io.fits.verify.VerifyWarning: Verification reported errors:
```

```
_____________________________________________________________ [doctest] sunpy.timeseries.sources.eve.EVESpWxTimeSeries _____________________________________________________________
[...]

PR #4358, opened 2020-07-14
Made Metadata clients Fido compatible

abhijeetmanhas:
### Description

Fixes #2744 , Fixes #2032 , Fixes #3293, Fixes #3941 , Fixes #2002
Queries like this can be made, and Fido fetch will ignore helio without giving any errors.
```python
from sunpy.net import attrs as a, Fido
from sunpy.net.helio import attrs as ha
res = Fido.search(a.Time('2005/1/3', '2005/1/4'), (ha.MaxRecords(10) & ha.Catalogue('hec') & ha.TableName(b'rhessi_hxr_flare')) | a.Instrument.xrs)
Fido.fetch(res)
```
Infile doctest can be seen for more details.

Similarly queries can be made with HEK too.

PR #5002, opened 2021-02-16
Fixed solar_rotate_coordinate() docstring examples

ayshih:
For `solar_rotate_coordinate()`, especially after the fixes in #4979:

- I fixed the output for the example skipped in #4358 (see also #4670).  The example can be unskipped when astropy/pytest-doctestplus#152 is in a release.
- I modified the other docstring example to use a different time and to not use Astropy's `get_body()`.
- I also changed internal code to not use Astropy's `get_body()`.  Because it's being called for Earth, there's no actual difference in the output, but for any other object, it would give the apparent location as seen by an Earth observer (with stellar aberration, but not with planetary aberration), as opposed to the true location.

### DIFF
--- a/sunpy/physics/differential_rotation.py
+++ b/sunpy/physics/differential_rotation.py
@@ -4,10 +4,10 @@ from copy import deepcopy
 import numpy as np
 
 import astropy.units as u
-from astropy.coordinates import BaseCoordinateFrame, Longitude, SkyCoord, get_body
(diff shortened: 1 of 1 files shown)


ayshih:
This will need to be included in #4999 to backport to 2.0

ayshih:
It turns out that I can't unskip the first docstring example due to a bug that prevents the use of `+IGNORE_WARNINGS` in our online test build.  See astropy/pytest-doctestplus#152.

nabobalis (approved):

PR #5019, opened 2021-02-19
Map drawing methods should use add_patch instead of add_artist

ayshih:
For whatever reason, the Map drawing methods (`draw_*`) use `add_artist()` instead of `add_patch()`, which breaks the usage of labels for legends.  This PR fixes that.

This PR will need to be manually backported because of `draw_quadrangle()`.


nabobalis (approved):

dstansby:
I would like to approve this, but as far as I can see there is no easy way of checking how the test images have changed? (I don't really have space on my laptop to create a whole new environment just to generate them locally)

nabobalis:
> I would like to approve this, but as far as I can see there is no easy way of checking how the test images have changed? (I don't really have space on my laptop to create a whole new environment just to generate them locally)

I would assume the difference would be nothing. The figure hash change being from adding the legend.

dstansby:
I assume so too, but it would be good to check.

Issue #4883, opened 2021-01-19
we should set the COLUMNS env var in conftest.py


Cadair:
Our doctests will not pass unless COLUMNS=180 which we currently do in tox.ini, but this means that if you run pytest outside of tox the doctests fail. We should set it in the pytest configure hook and unset it in the deconfigure hook, reverting it back to the previous value

PR #5021, opened 2021-02-19
Set COLUMNS environment variable in conftest

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/4883 - I think I did it right?


nabobalis:
To stop yourself having to add it manually?

nabobalis (approved):

PR #5022, opened 2021-02-20
Fix scraper issue with Windows and URL path separation 

samaloney:
### Description

Scraper subtly broken on windows when using regex patterns. In the `scraper.range` call `os.path.dirname` is used to obtain the base directory from the pattern e.g. `http://a.server.dom/some/path/afile(\d){3}.ext`. On windows `os.path.dirname` correctly interprets backslashes in the url pattern as path separators.  So a **URL** like above which should become `http://a.server.dom/some/path` instead becomes `http://a.server.dom/some/path/afile`.

I replaced `os.path.dirname` which assumes it is working with a path with code used the fact the pattern is a url and not a path. I looked at using `urllib` `parserurl` but that didn't seem to work with the regex patterns either.

### DIFF
--- a/sunpy/util/scraper.py
+++ b/sunpy/util/scraper.py
@@ -116,8 +116,8 @@ class Scraper:
             Notice that these directories may not exist in the archive.
         """
         # find directory structure - without file names
-        directorypattern = os.path.dirname(self.pattern) + '/'
-        # TODO what if there's not slashes?
+        if '/' in self.pattern:
+            directorypattern = '/'.join(self.pattern.split('/')[:-1]) + '/'
         timestep = self._smallerPattern(directorypattern)
         if timestep is None:
             return [directorypattern]
--- a/changelog/5022.bugfix.rst
+++ b/changelog/5022.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed bug in `sunpy.until.scraper.Scraper` which caused URL patterns containing backslashes to be incorrectly parsed on Windows.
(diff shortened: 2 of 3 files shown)


nabobalis:
If we can fix it in this PR, we can backport it as a bugifix.

sunpy/util/scraper.py
```diff
@@ -116,8 +116,8 @@ def range(self, timerange):
             Notice that these directories may not exist in the archive.
         """
         # find directory structure - without file names
-        directorypattern = os.path.dirname(self.pattern) + '/'
-        # TODO what if there's not slashes?
+        if '/' in self.pattern:
+            directorypattern = '/'.join(self.pattern.split('/')[:-1]) + '/'
```
nabobalis, line 120: I don't understand how this fixed it on windows.

samaloney: So `os.path.dirname` calls different code paths on windows and nix/posix systems. On windows `\` is obviously a path separator [1](https://github.com/python/cpython/blob/216cb1469f566ba5493bf53a73da9ccdac05ccfc/Lib/ntpath.py#L34)  while on posix it's `/`  [2](https://github.com/python/cpython/blob/216cb1469f566ba5493bf53a73da9ccdac05ccfc/Lib/posixpath.py#L41) same as urls. So the `\` in the regex is treated as a path separator on windows.

samaloney:
I guess I should open an issue for this too?

nabobalis:
> I guess I should open an issue for this too?

If you want to keep an issue in the history for it.

nabobalis (approved):

nabobalis:
Just a bugfix changelog and this is good to go.

nabobalis (approved):

PR #5025, opened 2021-02-21
added dynamic z axis option

### LINKED ISSUE #4718: Add a dynamic z axis option to the animators
So if your frames have a wide dynamic range the z scale changes

jeffreypaul15:
### Description

Fixes #4718 
As @dstansby requested, on setting ```vmax='auto'``` while creating an animation, sets ```vmin```/```vmax``` to a ```clip_interval=(1, 90)*u.percent``` of the current image that is being shown.

* Added a separate method ```update_bounds()``` which calculates ```vmin/vmax``` for the current 2D image plot by using astropy's ```AsymmetricPercentileInterval```.
* Calls this method on initial display of the 2D plot as well as when the plot is being updated.

The results for  [example](https://docs.sunpy.org/en/latest/generated/gallery/plotting/arrayanimatorwcs.html#sphx-glr-generated-gallery-plotting-arrayanimatorwcs-py) when not specifying ```vmax``` are
image
and
image

After specifying ```vmax='auto'``` the results are
image
and
image

What I did find odd was specifying ```vmax='auto'``` because ```vmin``` is also being automatically set, wouldn't it be better to specify either ```vmax='auto'``` or ```vmin='auto'``` or having a separate variable to set the bounds as automatic?


nabobalis:
I wonder if we want it to default to "auto".

jeffreypaul15:
> I wonder if we want it to default to "auto".
All of the figure hashes will have to changed in that case :sweat_smile:

changelog/5025.feature.rst
```diff
@@ -0,0 +1 @@
+Added dynamic z axis option to the animators.
```
nabobalis, line 1: This changelog needs expanding.

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +303,30 @@ def plot_start_image_2d(self, ax):
 
         return im
 
+    def update_bounds(self):
```
nabobalis, line 306: The function name is a bit ambiguous.

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +303,30 @@ def plot_start_image_2d(self, ax):
...
+        """
+        Update vmin/vmax when set to 'auto'.
+        """
+        print(self.imshow_kwargs)
```
nabobalis, line 310:
```suggestion
```

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +303,30 @@ def plot_start_image_2d(self, ax):
...
+        if 'vmax' in self.imshow_kwargs and self.imshow_kwargs['vmax'] == 'auto':
+            # set clip interval to 90% of maximum value
+            clip_interval = (1, 90)*u.percent
+            clip_percentages = clip_interval.to('%').value
+            vmin, vmax = AsymmetricPercentileInterval(*clip_percentages).get_limits(self.data_transposed)
+            return vmin, vmax
+        else:
+            return None
```
nabobalis, line 318:
```suggestion
```

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +303,30 @@ def plot_start_image_2d(self, ax):
 
         return im
 
+    def update_bounds(self):
+        """
+        Update vmin/vmax when set to 'auto'.
+        """
+        print(self.imshow_kwargs)
+        if 'vmax' in self.imshow_kwargs and self.imshow_kwargs['vmax'] == 'auto':
+            # set clip interval to 90% of maximum value
```
nabobalis, line 312:
```suggestion
```

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +303,30 @@ def plot_start_image_2d(self, ax):
 
         return im
 
+    def update_bounds(self):
+        """
+        Update vmin/vmax when set to 'auto'.
+        """
+        print(self.imshow_kwargs)
+        if 'vmax' in self.imshow_kwargs and self.imshow_kwargs['vmax'] == 'auto':
```
nabobalis, line 311:
```suggestion
        if self.imshow_kwargs.get('vmax') == 'auto':
```
I think?

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +303,30 @@ def plot_start_image_2d(self, ax):
...
+            # set clip interval to 90% of maximum value
+            clip_interval = (1, 90)*u.percent
+            clip_percentages = clip_interval.to('%').value
```
nabobalis, line 314: Can these two lines not be combined?

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +303,30 @@ def plot_start_image_2d(self, ax):
...
+            vmin, vmax = AsymmetricPercentileInterval(*clip_percentages).get_limits(self.data_transposed)
+            return vmin, vmax
```
nabobalis, line 316: Same here?

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +303,30 @@ def plot_start_image_2d(self, ax):
...
         """
         Update the image plot.
         """
         self.axes.reset_wcs(wcs=self.wcs, slices=self.slices_wcsaxes)
         im.set_array(self.data_transposed)
+
+        bounds = self.update_bounds()
+        if bounds is not None:
+            vmin, vmax = bounds
+            im.set_clim(vmin, vmax)
```
nabobalis, line 330: Could the function itself not update the im instead of calling it and then checking if its none?

jeffreypaul15: This would be an issue because the initial plots are updated by the vmin/vmax passed to it. So hence I did it this way

sunpy/visualization/animator/wcs.py
jeffreypaul15: Do you want them to be in one single line? Or do you want it split further? I'm a bit confused

nabobalis: Can they not be one single line?

sunpy/visualization/animator/wcs.py
jeffreypaul15: Also, the data to be displayed (2D plot is being updated here, so I thought updating the data and the vmin/vmax together would work better.

sunpy/visualization/animator/wcs.py
jeffreypaul15: This is necessary because I'm calling this method without passing vmax='auto' would case an unpacking error

nabobalis: I thought functions return none if you don't trigger a manual return?

jeffreypaul15: Yeah, sorry got a tad bit confused there. Making the changes now.

sunpy/visualization/animator/wcs.py
```diff
@@ -281,6 +282,9 @@ def plot_start_image_2d(self, ax):
                        'origin': 'lower'}
         imshow_args.update(self.imshow_kwargs)
 
+        if 'vmax' in imshow_args and imshow_args['vmax'] == 'auto':
```
nabobalis, line 285:
```suggestion
        if self.imshow_kwargs.get('vmax') == 'auto':
```

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +303,27 @@ def plot_start_image_2d(self, ax):
...
         return im
 
+    def get_2d_plot_clip_bounds(self):
+        """
+        Update vmin/vmax when set to 'auto'.
+        """
+        if self.imshow_kwargs.get('vmax') == 'auto':
+            clip_interval = (1, 90)*u.percent
+            clip_percentages = clip_interval.to('%')
+            clip_limits = AsymmetricPercentileInterval(*clip_percentages.value)
```
nabobalis, line 313: Can these lines not be combined into one?

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +303,27 @@ def plot_start_image_2d(self, ax):
...
+            vmin, vmax = clip_limits.get_limits(self.data_transposed)
+            return vmin, vmax
```
nabobalis, line 315: Can these lines not be combined into one?

sunpy/visualization/animator/wcs.py
jeffreypaul15: Sorry :sweat_smile:, I'm a bit confused as to what you mean

sunpy/visualization/animator/wcs.py
jeffreypaul15: Same here :sweat_smile:

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +303,27 @@ def plot_start_image_2d(self, ax):
...
     def update_plot_2d(self, val, im, slider):
         """
         Update the image plot.
         """
         self.axes.reset_wcs(wcs=self.wcs, slices=self.slices_wcsaxes)
         im.set_array(self.data_transposed)
+
+        bounds = self.get_2d_plot_clip_bounds()
+        if bounds is not None:
+            vmin, vmax = bounds
```
nabobalis, line 326: I don't like the fact we have to check if its none and then unpack the return after the fact. 

The issue is that I don't have a solution to this.

changelog/5025.feature.rst
```diff
@@ -0,0 +1,3 @@
+Added dynamic z axis option to `sunpy.visualization.animator`.
```
nabobalis, line 1: Is this a z axis?

sunpy/visualization/animator/wcs.py
nabobalis: Why is this not one line of python code?

sunpy/visualization/animator/wcs.py
nabobalis: Why is this not one line of python code?

changelog/5025.feature.rst
jeffreypaul15: Was confused about that actually, when I asked about it in the #4718 this is what @dstansby stated so I was under the assumption that it's correct.

sunpy/visualization/animator/wcs.py
jeffreypaul15: Oh, I thought you wanted them to not be in a single line :facepalm:. Sorry I'll change that

changelog/5025.feature.rst
nabobalis: This sets the image min and max limits on a 2D image. I don't think that is a Z axis.

jeffreypaul15: Yeah, so I guess the PR as a whole was pointing to a separate issue being mentioned under #4718

nabobalis: No this does address that issue.

sunpy/visualization/animator/wcs.py
jeffreypaul15: What if ```get_2d_plot_clip_bounds``` returns a ```[None, None] ```, that way  
```python
vmin, vmax = self.get_2d_plot_clip_bounds()
``` 
would update the plot as ```None```

changelog/5025.feature.rst
jeffreypaul15: So, should I change the name of this PR to something else?

nabobalis: The name of the PR doesn't really matter.

sunpy/visualization/animator/wcs.py
nabobalis: What happens if you pass in None,None to the limit function?

jeffreypaul15: Is this better instead of having to check for None and unpack?

jeffreypaul15: > What happens if you pass in None,None to the limit function?

I think this is a better way to go about it

```python
def get_2d_plot_clip_bounds(self):
    """
    Update vmin/vmax when set to 'auto'.
    """
    vmin, vmax = None, None
    if self.imshow_kwargs.get('vmax') == 'auto':
        clip_interval = (1, 90)*u.percent
        clip_percentages = clip_interval.to('%')
        clip_limits = AsymmetricPercentileInterval(*clip_percentages.value)
        vmin, vmax = clip_limits.get_limits(self.data_transposed)
        return vmin, vmax

def update_plot_2d(self, val, im, slider):
    """
    Update the image plot.
    """
    self.axes.reset_wcs(wcs=self.wcs, slices=self.slices_wcsaxes)
    im.set_array(self.data_transposed)

    vmin, vmax = self.get_2d_plot_clip_bounds()
    im.set_clim(vmin, vmax)

    slider.cval = val
```

nabobalis: Hmmm, I like it more than the current way. I think the name of the function still needs some work.

jeffreypaul15: I'm can't really thing of anything better. :sweat_smile:

nabobalis: For the function name? Well it doesn't get the bounds of an image, and it only works if vmax is set to auto.  

It currently gives the impression it will update the bounds of an image every step.

jeffreypaul15: changed it to ```get_2d_plot_limts``` hoping that it is better :sweat_smile:

changelog/5025.feature.rst
```diff
@@ -0,0 +1,3 @@
...
+Specifying ``vmax='auto'`` calculates and updates the colormap
+according to the current slice shown.
```
nabobalis, line 3:
```suggestion
Added the ability to dynamically scale `sunpy.visualization.animator` instances.
By specifying ``vmax="auto"`` will now clip the minimum and maximum at each slider step to 1% and 90% respectively.
```

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +303,24 @@ def plot_start_image_2d(self, ax):
 
         return im
 
+    def get_2d_plot_limts(self):
+        """
+        Update vmin/vmax when set to 'auto'.
```
nabobalis, line 308:
```suggestion
        Get vmin, vmax of a data slice when "vmax" is set to "auto".
```

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +303,24 @@ def plot_start_image_2d(self, ax):
...
+        """
+        vmin, vmax = None, None
+        if self.imshow_kwargs.get('vmax') == 'auto':
+            vmin, vmax = AsymmetricPercentileInterval(*((1, 90)*u.percent).to('%').value)\
+                         .get_limits(self.data_transposed)
```
nabobalis, line 313:
```suggestion
            percent_limits = (1, 90)*u.percent).to('%').value
            return AsymmetricPercentileInterval(percent_limits).get_limits(self.data_transposed)
```
Would this not work?

jeffreypaul15: This would but when ```vmax``` isn't explicitly specified, we would have to specify a separate else and return [None, None]. It looked neater having to return ```vmin, vmax```

nabobalis:
I forget, do we backport features?

changelog/5025.feature.rst
```diff
@@ -0,0 +1,2 @@
+Added the ability to dynamically scale `sunpy.visualization.animator` instances.
+By specifying ``vmax="auto"`` will now clip the minimum and maximum at each slider step to 1% and 90% respectively.
```
nabobalis, line 2:
```suggestion
By specifying ``vmax="auto"``, it will now clip the minimum and maximum at each slider step to 1% and 90% respectively.
```

sunpy/visualization/animator/wcs.py
nabobalis: I updated it to account for that.

jeffreypaul15:
> I forget, do we backport features?

```draw quadrangle``` was a feature, but that wasn't backported

**Comment by nabobalis** [comment]
> > I forget, do we backport features?
> 
> `draw quadrangle` was a feature, but that wasn't backported

This question was not aimed at you. Don't worry.

File: sunpy/visualization/animator/wcs.py [hunk 19]
```diff
@@ -299,10 +303,24 @@ def plot_start_image_2d(self, ax):
 
         return im
 
+    def get_2d_plot_limts(self):
+        """
+        Get vmin, vmax of a data slice when "vmax" is set to "auto".
+        """
+        vmin, vmax = None, None
+        if self.imshow_kwargs.get('vmax') == 'auto':
+            percent_limits = (1, 90)*u.percent).to('%').value
```
jeffreypaul15, line 312: Unmatched parenthesis here, I'll make that fix

sunpy/visualization/animator/wcs.py
```diff
@@ -281,6 +282,9 @@ def plot_start_image_2d(self, ax):
                        'origin': 'lower'}
         imshow_args.update(self.imshow_kwargs)
 
+        if self.imshow_kwargs.get('vmax') == 'auto':
```
nabobalis, line 285: Do we need to do this check twice?

jeffreypaul15: I guess this would work regardless of having to check it as we're passing ```None``` if not specified.

nabobalis:
I think we need a test for this new behavior?

nabobalis:
Question, what if someone wanted to fix the vmin or vmax axis but let the other one be auto? Would someone even want to do this?

sunpy/visualization/animator/wcs.py
```diff
@@ -281,6 +282,8 @@ def plot_start_image_2d(self, ax):
                        'origin': 'lower'}
         imshow_args.update(self.imshow_kwargs)
 
+        imshow_args['vmin'], imshow_args['vmax'] = self.get_2d_plot_limts()
```
nabobalis, line 285: That reminds me, what if someone added "vmax" or "vmin", would they not be replaced with None by this? 
Maybe we do need the check?

jeffreypaul15:
> I think we need a test for this new behavior?



> Question, what if someone wanted to fix the vmin or vmax axis but let the other one be auto? Would someone even want to do this?

In that case wouldn't it be better to separately specify ```vmin='auto'``` as well? As of now ```vmax='auto'``` sets both of them to be auto.

sunpy/visualization/animator/wcs.py
jeffreypaul15: Yes, I was just about to being with the tests and I thought of this as well :sweat_smile:, the check first check for 'vmax' in ```self.imshow_kwargs``` prevents this from happening

dstansby:
Just tuning in here, lets just worry about making both limits auto instead of one or the other. If desired (I don't think it is though) we can always modify that later in another PR.

changelog/5025.feature.rst
```diff
@@ -0,0 +1,2 @@
+Added the ability to dynamically scale `sunpy.visualization.animator` instances.
+By specifying ``vmax="auto"``, it will now clip the minimum and maximum at each slider step to 1% and 90% respectively.
```
dstansby, line 2: We should make these limits symmetrical, e.g. use 1% and 99% or 10% and 90%. I think there's somewhere else in the library that we use 1% and 99%, so I would say lets go with that.

dstansby (changes requested):
This looks great overall 👍 Some comments:
- The new option needs documenting in the docstring
- We should change the limits to 1% and 99%.
- I think we should check if `vmin == 'auto'` as well as `vmax == 'auto'`
- This needs a new test. I think the best way to do this is to copy an existing image test and, apply 'auto' limits

changelog/5025.feature.rst
jeffreypaul15: @dstansby 99% causes the image to not be as visible but if that's the standard then I shall make that change.
I'll make the change for 2 specifying either of them separately.
While writing tests should I write 3 separate tests for ```vmax='auto'``` , ```vmin='auto'```, and both, or should I just have one test where both ```vmax='auto'``` and ```vmin='auto'```?

jeffreypaul15:
@nabobalis figure tests fail because of the absence of the png file at sunpy-figure-tests, I'm a little confused as to how this works, and whether a png file should be added there

nabobalis:
> @nabobalis figure tests fail because of the absence of the png file at sunpy-figure-tests, I'm a little confused as to how this works, and whether a png file should be added there

The figure tests fail since the hash is wrong and since the hash is wrong, it goes to fetch the original image to do a comparison and then fails there.

jeffreypaul15:
> > @nabobalis figure tests fail because of the absence of the png file at sunpy-figure-tests, I'm a little confused as to how this works, and whether a png file should be added there
> 
> The figure tests fail since the hash is wrong and since the hash is wrong, it goes to fetch the original image to do a comparison and then fails there, since the figure has never been uploaded.
> 
> It's an issue with the figure test script and it is something we need to fix.

Oh okay, do I let it be for now?

nabobalis:
We need to fix the hashes.

jeffreypaul15:
> We need to fix the hashes.

Figure test doesn't generate the hashes so I'm not really sure how to get the hashes.

nabobalis:
> > We need to fix the hashes.
> 
> Figure test doesn't generate the hashes so I'm not really sure how to get the hashes.

You generate them locally using the commands from https://docs.sunpy.org/en/latest/dev_guide/contents/tests.html#figure-unit-tests

jeffreypaul15:
> > > We need to fix the hashes.
> > 
> > 
> > Figure test doesn't generate the hashes so I'm not really sure how to get the hashes.
> 
> You generate them locally using the commands from https://docs.sunpy.org/en/latest/dev_guide/contents/tests.html#figure-unit-tests

Ah yeah I remember this being an issue locally, I'll give it a shot

jeffreypaul15:
The tox test doesn't seem to work locally for me at all, is there an alternative?

jeffreypaul15:
The tox test doesn't seem to work locally for me at all, is there an alternative?

nabobalis:
What is the error?

If you can't get it to work, you will need someone else to run them for you.

jeffreypaul15:
> What is the error?
> 
> If you can't get it to work, you will need someone else to run them for you.

Coverage.py warning: Module sunpy was previously imported, but not measured (module-not-measured).
I think this was the same issue I was facing last time as well..

nabobalis:
That's not an error should stop the test run

jeffreypaul15:
> That's not an error that should stop the test run

ERROR ../.. - DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working

This showed up, although it didn't make sense to me because it was just a Depreciation Warning

nabobalis:
The test suite is set to error on those warnings. You can change setup.cfg to ignore it but that error shouldn't happen in the first place.

nabobalis:
Is there any more information about where that line comes from?

jeffreypaul15:
> The test suite is set to error on those warnings. You can change setup.cfg to ignore it but that error shouldn't happen in the first place.

I made a change to the setup.cfg (commented a few lines) and the tests passed locally

nabobalis:
> @nabobalis The figure-tests passed, a little more light shed on adding documentation in the docstring would be helpful.

We have https://docs.sunpy.org/en/latest/dev_guide/contents/documentation.html

> Also should I add another test where only `vmax` or `vmin` is specified? In cases where the user wants to specify one as `auto` and the other to be left as it is.

Yes please.

jeffreypaul15:
> > @nabobalis The figure-tests passed, a little more light shed on adding documentation in the docstring would be helpful.
> 
> We have https://docs.sunpy.org/en/latest/dev_guide/contents/documentation.html
> 
> > Also should I add another test where only `vmax` or `vmin` is specified? In cases where the user wants to specify one as `auto` and the other to be left as it is.
> 
> Yes please.

Should I add both or just either vmax or vmin?

nabobalis:
> > > @nabobalis The figure-tests passed, a little more light shed on adding documentation in the docstring would be helpful.
> > 
> > 
> > We have [docs.sunpy.org/en/latest/dev_guide/contents/documentation.html](https://docs.sunpy.org/en/latest/dev_guide/contents/documentation.html)
> > > Also should I add another test where only `vmax` or `vmin` is specified? In cases where the user wants to specify one as `auto` and the other to be left as it is.
> > 
> > 
> > Yes please.
> 
> Should I add both or just either vmax or vmin?

Both if you can.

jeffreypaul15:
> > > > @nabobalis The figure-tests passed, a little more light shed on adding documentation in the docstring would be helpful.
> > > 
> > > 
> > > We have [docs.sunpy.org/en/latest/dev_guide/contents/documentation.html](https://docs.sunpy.org/en/latest/dev_guide/contents/documentation.html)
> > > > Also should I add another test where only `vmax` or `vmin` is specified? In cases where the user wants to specify one as `auto` and the other to be left as it is.
> > > 
> > > 
> > > Yes please.
> > 
> > 
> > Should I add both or just either vmax or vmin?
> 
> Both if you can.

I've added 2 more tests, the docstring documentation however may not be how you may want it to be so please let me know if that change has to be made.

nabobalis:
> > > > > @nabobalis The figure-tests passed, a little more light shed on adding documentation in the docstring would be helpful.
> > > > 
> > > > 
> > > > We have [docs.sunpy.org/en/latest/dev_guide/contents/documentation.html](https://docs.sunpy.org/en/latest/dev_guide/contents/documentation.html)
> > > > > Also should I add another test where only `vmax` or `vmin` is specified? In cases where the user wants to specify one as `auto` and the other to be left as it is.
> > > > 
> > > > 
> > > > Yes please.
> > > 
> > > 
> > > Should I add both or just either vmax or vmin?
> > 
> > 
> > Both if you can.
> 
> I've added 2 more tests, the docstring documentation however may not be how you may want it to be so please let me know if that change has to be made.

I don't them to be figure tests, I personally want to make sure that we aren't changing vim or vmax values by mistake with the logic.

jeffreypaul15:
> > > > > > @nabobalis The figure-tests passed, a little more light shed on adding documentation in the docstring would be helpful.
> > > > > 
> > > > > 
> > > > > We have [docs.sunpy.org/en/latest/dev_guide/contents/documentation.html](https://docs.sunpy.org/en/latest/dev_guide/contents/documentation.html)
> > > > > > Also should I add another test where only `vmax` or `vmin` is specified? In cases where the user wants to specify one as `auto` and the other to be left as it is.
> > > > > 
> > > > > 
> > > > > Yes please.
> > > > 
> > > > 
> > > > Should I add both or just either vmax or vmin?
> > > 
> > > 
> > > Both if you can.
> > 
> > 
> > I've added 2 more tests, the docstring documentation however may not be how you may want it to be so please let me know if that change has to be made.
> 
> I don't them to be figure tests, I personally want to make sure that we aren't changing vim or vmax values by mistake with the logic.

Shall I remove the figure tests then? How do you want me to write the tests in this case?

sunpy/visualization/animator/wcs.py
```diff
@@ -59,7 +60,10 @@ class ArrayAnimatorWCS(ArrayAnimator):
     ylabel: `string`, optional
        The yaxis label to use when drawing a line plot. Setting the label on
        the y-axis on an image plot should be done via ``coord_params``.
-
+    vmin: `int` or `str`, optional
+        if specified as "auto", clips the minimum and maximum at each slider step to 1% and 99%
```
nabobalis, line 64:
```suggestion
        Sets the minimum threshold for all slices of the animation. 
        If specified as "auto", clips the minimum and maximum at each slider step to 1% and 99%
```

sunpy/visualization/animator/wcs.py
```diff
@@ -59,7 +60,10 @@ class ArrayAnimatorWCS(ArrayAnimator):
...
+    vmax: `int` or `str`, optional
+        if specified as "auto", clips the minimum and maximum at each slider step to 1% and 99%
```
nabobalis, line 66:
```suggestion
        Sets the maximum threshold for all slices of the animation. 
        If specified as "auto", clips the minimum and maximum at each slider step to 1% and 99%
```

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +308,28 @@ def plot_start_image_2d(self, ax):
...
         return im
 
+    def get_2d_plot_limts(self):
+        """
+        Get vmin, vmax of a data slice when "vmax" or "vmin" is set to "auto".
+        """
+        vmin, vmax = None, None
+        if self.imshow_kwargs.get('vmin') == 'auto':
+            percent_limits = ((1, 99)*u.percent).to('%').value
+            vmin, _ = AsymmetricPercentileInterval(*percent_limits).get_limits(self.data_transposed)
```
nabobalis, line 318: Why not work this out once and then set the return value based on the keyword argument?

nabobalis:
> > > > > > > @nabobalis The figure-tests passed, a little more light shed on adding documentation in the docstring would be helpful.
> > > > > > 
> > > > > > 
> > > > > > We have [docs.sunpy.org/en/latest/dev_guide/contents/documentation.html](https://docs.sunpy.org/en/latest/dev_guide/contents/documentation.html)
> > > > > > > Also should I add another test where only `vmax` or `vmin` is specified? In cases where the user wants to specify one as `auto` and the other to be left as it is.
> > > > > > 
> > > > > > 
> > > > > > Yes please.
> > > > > 
> > > > > 
> > > > > Should I add both or just either vmax or vmin?
> > > > 
> > > > 
> > > > Both if you can.
> > > 
> > > 
> > > I've added 2 more tests, the docstring documentation however may not be how you may want it to be so please let me know if that change has to be made.
> > 
> > 
> > I don't them to be figure tests, I personally want to make sure that we aren't changing vim or vmax values by mistake with the logic.
> 
> Shall I remove the figure tests then? How do you want me to write the tests in this case? Also if a specific vmax or vmin value is specified, the vmin and vmax does get updated in the `update_plot_2d()` method but this doesn't change in the aminator, as I mentioned earlier, if vmax or vmin is explicitly specified, the preference does go to the explicitly specified values even though we are setting then to `None`. This is what I was testing and experimenting with locally but I want to know how to go about this testing..

I don't know, maybe it is possible to get the vmin or vmax values out of the returned arrayanimator.

sunpy/visualization/animator/wcs.py
```diff
@@ -59,7 +60,12 @@ class ArrayAnimatorWCS(ArrayAnimator):
     ylabel: `string`, optional
        The yaxis label to use when drawing a line plot. Setting the label on
        the y-axis on an image plot should be done via ``coord_params``.
-
+    vmin: `int` or `str`, optional
+        Sets the minimum threshold for all slices of the animation. 
+        If specified as "auto", clips the minimum and maximum at each slider step to 1% and 99%
```
nabobalis, line 65: Is this true anymore? Does it not just set the vmin at 1%?

jeffreypaul15: Oh yeah, vmin and vmax will have to be separated in the docstring

sunpy/visualization/animator/wcs.py
jeffreypaul15: Sorry @nabobalis, not quite sure what you meant by this.

nabobalis: This blocks of code are repeated with only the value you return changed. Why not combine them?

sunpy/visualization/animator/wcs.py
```diff
@@ -59,7 +60,12 @@ class ArrayAnimatorWCS(ArrayAnimator):
     ylabel: `string`, optional
        The yaxis label to use when drawing a line plot. Setting the label on
        the y-axis on an image plot should be done via ``coord_params``.
-
+    vmin: `int` or `str`, optional
+        Sets the minimum threshold for all slices of the animation. 
+        If specified as "auto", clips the minimum at each slider step to 1%
```
nabobalis, line 65:
```suggestion
        If specified as "auto", clips the minimum at each slider step to 1%.
```

sunpy/visualization/animator/wcs.py
```diff
@@ -59,7 +60,12 @@ class ArrayAnimatorWCS(ArrayAnimator):
...
+    vmax: `int` or `str`, optional
+        Sets the maximum threshold for all slices of the animation. 
+        If specified as "auto", clips the maximum at each slider step to 99%
```
nabobalis, line 68:
```suggestion
        If specified as "auto", clips the maximum at each slider step to 99%.
```

ayshih:
Sorry for joining this party late, but I'm a strong :-1: on this API.  We already have analogous auto-scaling functionality for `GenericMap`  – namely the `clip_interval` keyword argument – and the API here is both different and less flexible.  There's no reason to hard-code the 1% and 99% percentile limits, and seemingly overloading `imshow()` keyword arguments also seems unnecessarily confusing.  Let's not re-invent the wheel; use a `clip_interval` keyowrd argument here too.

jeffreypaul15:
> Sorry for joining this party late, but I'm a strong 👎 on this API. We already have analogous auto-scaling functionality for `GenericMap` – namely the `clip_interval` keyword argument – and the API here is both different and less flexible. There's no reason to hard-code the 1% and 99% percentile limits, and seemingly overloading `imshow()` keyword arguments also seems unnecessarily confusing. Let's not re-invent the wheel; use a `clip_interval` keyowrd argument here too.

Oh, my apologies. Didn't know that clip interval could be used here as well. I'll work get started on another issue then.

ayshih:
No, you misunderstand me.  I'm suggesting that you *create* the `clip_interval` keyword argument here to mimic the existing API present in `GenericMap.plot()`.  It doesn't already exist.

nabobalis:
> Sorry for joining this party late, but I'm a strong 👎 on this API. We already have analogous auto-scaling functionality for `GenericMap` – namely the `clip_interval` keyword argument – and the API here is both different and less flexible. There's no reason to hard-code the 1% and 99% percentile limits, and seemingly overloading `imshow()` keyword arguments also seems unnecessarily confusing. Let's not re-invent the wheel; use a `clip_interval` keyowrd argument here too.

I forgot completely we had that API on map. Does map not take at all vmin/vmax?

ayshih:
`GenericMap.plot()` also handles `vmin`/`vmax`, so you can specify a fixed range.  It looks like using `clip_interval` there will override any provided `vmin`/`vmax`, so it's not possible to have the functionality that's in this PR of, for example, setting `vmin` to a fixed value and auto-scaling `vmax`.  I don't know how valuable that use case is.

nabobalis:
> `GenericMap.plot()` also handles `vmin`/`vmax`, so you can specify a fixed range. It looks like using `clip_interval` there will override any provided `vmin`/`vmax`, so it's not possible to have the functionality that's in this PR of, for example, setting `vmin` to a fixed value and auto-scaling `vmax`. I don't know how valuable that use case is.

Hmm, I think we might want to decide on an API for this then as a way to keep it similar to map. Thanks for bringing it up.

sunpy/visualization/animator/wcs.py
ayshih: To be clear, the issue is that if both `vmin` and `vmax` are set to "auto", you compute the limits twice unnecessarily with the repeated call.  It'd be cleaner to do something like:
```python
        if self.imshow_kwargs.get('vmin') == 'auto' or self.imshow_kwargs.get('vmax') == 'auto':
            percent_limits = ((1, 99)*u.percent).to('%').value
            vmin_temp, vmax_temp = AsymmetricPercentileInterval(*percent_limits).get_limits(self.data_transposed)
        vmin = vmin_temp if self.imshow_kwargs.get('vmin') == 'auto' else None
        vmax = vmax_temp if self.imshow_kwargs.get('vmax') == 'auto' else None
        return vmin, vmax
```

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +310,28 @@ def plot_start_image_2d(self, ax):
 
         return im
 
+    def get_2d_plot_limts(self):
```
ayshih, line 313: Is this misspelling of "limits" intentional?

sunpy/visualization/animator/wcs.py
```diff
@@ -281,6 +287,11 @@ def plot_start_image_2d(self, ax):
                        'origin': 'lower'}
         imshow_args.update(self.imshow_kwargs)
 
+        if self.imshow_kwargs.get('vmin') == 'auto':
+            imshow_args['vmin'], _ = self.get_2d_plot_limts()
+        if self.imshow_kwargs.get('vmax') == 'auto':
+            _, imshow_args['vmax'] = self.get_2d_plot_limts()
```
ayshih, line 293: As with https://github.com/sunpy/sunpy/pull/5025/files#r582872523, you shouldn't call your method twice.  In fact, the doubled call here and the doubled call within the method means that the underlying calculation is getting repeated *four* times.

sunpy/visualization/animator/wcs.py
ayshih: Also, this should be a private method

nabobalis:
> Hmm, I think we might want to decide on an API for this then as a way to keep it similar to map. Thanks for bringing it up.

So I assume the animators already handle vmin and vmax args correctly. 
So if we change this PR to add a `clip_interval` keyword that will do this scaling via a user provided limit?

ayshih:
Yes, that's precisely what I meant with my suggestion

nabobalis:
> Yes, that's precisely what I meant with my suggestion

We should do your suggestion.

sunpy/visualization/animator/wcs.py
jeffreypaul15: Misspelling was not intentional, didn't check that, sorry.

jeffreypaul15:
> > Yes, that's precisely what I meant with my suggestion
> 
> We should do your suggestion.

Shall I go on with that then?

nabobalis:
> > > Yes, that's precisely what I meant with my suggestion
> > 
> > 
> > We should do your suggestion.
> 
> Shall I go on with that then? Should I make that change in a separate PR or just edit this one?

Yes, just do it in this PR unless you want to start fresh.

jeffreypaul15 (commented):
> > > > Yes, that's precisely what I meant with my suggestion
> > > 
> > > 
> > > We should do your suggestion.
> > 
> > 
> > Shall I go on with that then? Should I make that change in a separate PR or just edit this one?
> 
> Yes, just do it in this PR unless you want to start fresh.

Also I was thinking, for example if a user decides to use the animator for 2 maps, and wants to specify ```vmax``` for both of them separately, isn't that not possible as of now?

nabobalis:
> Also I was thinking, for example if a user decides to use the animator for 2 maps, and wants to specify `vmax` for both of them separately, isn't that not possible as of now?

I am not sure that is possible and I am not sure we want to support that.

jeffreypaul15:
I've added ```clip_interval``` in the form of ```kwargs``` for now, although ```GenericMap.plot()``` has a separate ```clip_interval``` argument, so I was wondering if you wanted it implemented it that way..

nabobalis:
I don't see why it can't be a proper keyword arg.

jeffreypaul15:
> I don't see why it can't be a proper keyword arg.

Shall I make that change then?

nabobalis:
> > I don't see why it can't be a proper keyword arg.
> 
> Shall I make that change then?

Yes please.

changelog/5025.feature.rst
```diff
@@ -0,0 +1,2 @@
+Added the ability to dynamically scale `sunpy.visualization.animator` instances.
+By specifying a ``clip_interval``, it will now clip the minimum and maximum at each slider step to the specified interval.
```
nabobalis, line 2:
```suggestion
By specifying the ``clip_interval`` keyword, it will now clip the minimum and maximum at each slider step to the specified interval.
```

sunpy/visualization/animator/wcs.py
```diff
@@ -60,9 +61,13 @@ class ArrayAnimatorWCS(ArrayAnimator):
        The yaxis label to use when drawing a line plot. Setting the label on
        the y-axis on an image plot should be done via ``coord_params``.
```
nabobalis, line 63:
```suggestion
```

sunpy/visualization/animator/wcs.py
```diff
@@ -60,9 +61,13 @@ class ArrayAnimatorWCS(ArrayAnimator):
...
 
+    clip_interval : two-element `~astropy.units.Quantity`, optional
```
nabobalis, line 64: Does this take any quantity?

sunpy/visualization/animator/wcs.py
```diff
@@ -60,9 +61,13 @@ class ArrayAnimatorWCS(ArrayAnimator):
...
+        If provided, the data for each map will be clipped to the percentile interval bounded by the two
+        numbers.
```
nabobalis, line 66:
```suggestion
        If provided, the data for each step will be clipped to the percentile interval bounded by the two numbers.
```

sunpy/visualization/animator/wcs.py
jeffreypaul15: Percentage right? I assumed that's how it's supposed to be written.

nabobalis: Sure it is the type but since it has to be a percent quantity, I wonder if its worth changing to that.

jeffreypaul15: All other docstrings in sunpy have ```astropy.units.Quantity``` regardless of them only working with one particular type though

sunpy/visualization/animator/wcs.py
```diff
@@ -281,6 +285,9 @@ def plot_start_image_2d(self, ax):
                        'origin': 'lower'}
         imshow_args.update(self.imshow_kwargs)
 
+        if self.clip_interval is not None:
```
nabobalis, line 288:
```suggestion
        if self.clip_interval:
```

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +306,27 @@ def plot_start_image_2d(self, ax):
...
+
+
     def update_plot_2d(self, val, im, slider):
         """
         Update the image plot.
         """
         self.axes.reset_wcs(wcs=self.wcs, slices=self.slices_wcsaxes)
         im.set_array(self.data_transposed)
+
+        if self.clip_interval is not None:
```
nabobalis, line 328:
```suggestion
        if self.clip_interval:
```

sunpy/visualization/animator/wcs.py
```diff
@@ -299,10 +306,27 @@ def plot_start_image_2d(self, ax):
 
         return im
 
+    def _get_2d_plot_limits(self):
+        """
+        Get vmin, vmax of a data slice when clip_interval is specified.
+        """
+        if len(self.clip_interval) == 2:
```
nabobalis, line 313: I wonder if this check is better to happen in the init instead. I don't think we should allow a bad clip_interval to be passed in and return the class.

sunpy/visualization/animator/wcs.py
nabobalis: Leave it then.

sunpy/visualization/animator/wcs.py
jeffreypaul15: Okay, I'll make this check within init itself.

jeffreypaul15:
Also, @nabobalis ```WARNING: AstropyDeprecationWarning: The truth value of a Quantity is ambiguous. In the future this will raise a ValueError. [astropy.units.quantity]``` shows up if we check for ```if clip_interval:```. That's why I used ```if clip_interval is not None```

nabobalis:
> Also, @nabobalis `WARNING: AstropyDeprecationWarning: The truth value of a Quantity is ambiguous. In the future this will raise a ValueError. [astropy.units.quantity]` shows up if we check for `if clip_interval:`. That's why I used `if clip_interval is not None`

Fair, add it back then.

sunpy/visualization/animator/wcs.py
```diff
@@ -88,6 +91,11 @@ def __init__(self, data, wcs, slices, coord_params=None, ylim='dynamic', ylabel=
         self.ylim = ylim
         self.ylabel = ylabel
 
+        if clip_interval is not None and len(clip_interval) == 2:
```
nabobalis, line 94: Is this check correct?

jeffreypaul15:
> > Also, @nabobalis `WARNING: AstropyDeprecationWarning: The truth value of a Quantity is ambiguous. In the future this will raise a ValueError. [astropy.units.quantity]` shows up if we check for `if clip_interval:`. That's why I used `if clip_interval is not None`
> 
> Fair, add it back then.

Why are the figure tests failing though

sunpy/visualization/animator/wcs.py
jeffreypaul15: Think there's an issue, gimme a second to verify

jeffreypaul15: Oh, sorry. I didn't see this

nabobalis:
Can you please stop requesting a review everytime you make some changes?

nabobalis (approved):

nabobalis:
You will need to fix the failing, Core Tests style_check [linux], but otherwise this looks good to me.

jeffreypaul15:
> Can you please stop requesting a review everytime you make some changes?

My apologies, I assumed that was the norm. I shall keep that in mind.

jeffreypaul15:
@nabobalis Why are the 2 build tests failing? I don't suppose it's something to do with the added code right?

nabobalis:
No, its ok.

Cadair (approved):

dstansby (approved):
Thanks for the PR! There's a couple of bits of tests we should add, but I'm going to merge this and someone can do that in another PR.

jeffreypaul15:
> Thanks for the PR! There's a couple of bits of tests we should add, but I'm going to merge this and someone can do that in another PR.

What exactly has to be tested? I could give it a go.

PR #5033, opened 2021-02-24
MDI synoptic map does not need crder

wtbarnes:
This allows MDI synoptic maps to not have `CRDER` keywords. I don't know how common this is or if we should be applying a different fix here, but I've been working with some MDI synoptic maps that don't have this and the MDI synoptic source barfs because they're missing `CRDER{1,2}`.


dstansby (approved):

nabobalis (approved):

nabobalis:
Did we backport synoptic maps to 2.0? I feel like we might not have?

PR #5038, opened 2021-02-27
added support to more than one registry url for hec

nabobalis:
Current reg is down, found another one but decided that we should just check each one.

### DIFF
--- a/sunpy/net/helio/parser.py
+++ b/sunpy/net/helio/parser.py
@@ -2,26 +2,29 @@
 This module is meant to parse the HELIO registry and return WSDL endpoints to
 facilitate the interfacing between further modules and HELIO.
 """
-import urllib
 import xml.etree.ElementTree as EL
 from contextlib import closing
+from urllib.request import urlopen
 
 from bs4 import BeautifulSoup
 
-from sunpy.net.helio import registry_links as RL
+from sunpy import log
 
 __all__ = ['webservice_parser', 'endpoint_parser', 'wsdl_retriever']
 
-# Lifespan in seconds before a link times-out
-LINK_TIMEOUT = 3
+
+LINK_TIMEOUT = 10
+REG_LINKS = [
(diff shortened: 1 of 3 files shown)


sunpy/net/helio/parser.py
```diff
@@ -37,17 +39,21 @@ def webservice_parser(service='HEC'):
...
-    link = RL.LINK + '/' + service.lower()
-    xml = link_test(link)
+    xml = None
+    for REG_LINK in REG_LINKS:
+        link = REG_LINK + service.lower()
+        xml = link_test(link)
+        if xml:
+            break
     if xml is None:
         return None
```
nabobalis, line 58: I feel like this logic could be made simpler and nicer.

nabobalis:
Online failure is the cutout service.

sunpy/net/helio/parser.py
```diff
@@ -175,9 +181,10 @@ def link_test(link):
         None
     """
     try:
-        with closing(urllib.request.urlopen(link, timeout=LINK_TIMEOUT)) as fd:
```
Cadair, line 178: what's closing do?

sunpy/net/helio/parser.py
Cadair: it'll suffice

sunpy/net/helio/parser.py
nabobalis: I assume it was before a time the urlopen was unable to be opened as a context manager.

Cadair (approved):

sunpy/net/helio/parser.py
Cadair: huuuh. mybbe.

nabobalis: https://docs.python.org/3/library/contextlib.html#contextlib.closing

sunpy/net/helio/parser.py
nabobalis: I rather do better than, "it'll suffice".

nabobalis:
The doctests failed and returned None, I don't want to merge a PR where I've made the tests more flaky.

PR #5040, opened 2021-02-27
[2.0] Collection of backports

nabobalis:
Backport of #5038 among others.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -77,8 +77,8 @@ deps =
     # Figure tests need a tightly controlled environment
     figure-!devdeps: astropy==4.0.1.post1
     figure-!devdeps: matplotlib==3.3.2
-
-    build_docs: git+https://github.com/Cadair/sphinx-gallery@skip_on_fail#egg=sphinx-gallery
+    # Temp measure until release
+    build_docs: git+https://github.com/sphinx-gallery/sphinx-gallery.git
 
 # The following indicates which extras_require from setup.cfg will be installed
 # dev is special in that it installs everything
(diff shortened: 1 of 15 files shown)


nabobalis:
I need to fix some tests

dstansby (changes requested):
I don't think we should be merging or releasing workflows (readthedocs) in this case that rely on git branches instead of released packages.

tox.ini
```diff
@@ -77,6 +78,8 @@ deps =
     figure-!devdeps: astropy==4.0.1.post1
     figure-!devdeps: matplotlib==3.3.2
 
+    build_docs: git+https://github.com/Cadair/sphinx-gallery@skip_on_fail#egg=sphinx-gallery
+
```
Cadair, line 82: We don't technically need this here as the tox build runs with `-W` so the behaviour is unchanged with or without the branch.

Cadair:
> I don't think we should be merging or releasing workflows (readthedocs in this case) that rely on git branches instead of released packages.

We are already using this on the 2.1 and master branches to prevent issues on RTD for PR previews and the actual release build following [a conversation in #sunpy-development:openastronomy.org](https://matrix.to/#/!JYqfIVJjWANcHnfktY:cadair.com/$16133960461201mMqai:cadair.com?via=cadair.com&via=matrix.org). 

I am mostly indifferent to if we backport it or not, but I just thought I would explain the backstory.

nabobalis:
Since the PR has been merged, I have updated it to point to upstream sphinx gallery.

Cadair (approved):

PR #5041, opened 2021-02-27
added tests for map sequence animator

### LINKED ISSUE #2296: Mapcube animation is not tested
#2295 points out the fact we don't have a test suite for the plotting capability of MapSequence.
That's not good.
So we need a MapSequence plot test.

https://github.com/sunpy/sunpy/blob/master/sunpy/map/mapsequence.py
https://github.com/sunpy/sunpy/blob/master/sunpy/map/tests/test_mapsequence.py

MapSequence  animation relies on https://github.com/sunpy/sunpy/blob/master/sunpy/visualization/animator/mapsequenceanimator.py 

This code has no tests - see https://github.com/sunpy/sunpy/tree/master/sunpy/visualization/animator/tests 

Tests for MapSequenceAnimator go in this subdirectory.
All the methods for the MapSequenceAnimator object will need tests.

A previous attempt was made here (when it was called MapCube): https://github.com/sunpy/sunpy/pull/2564

jeffreypaul15:
### Description

Fixes #2296
Added tests for ```MapSequenceAnimator```. 
Figure hashes have yet to be added.


sunpy/visualization/animator/tests/test_mapsequence.py
```diff
@@ -0,0 +1,72 @@
...
+
+def test_construct_map_sequence_animator(test_map_sequence):
+    map_animator = MapSequenceAnimator(test_map_sequence)
+    assert isinstance(map_animator, MapSequenceAnimator)
+
+
+@figure_test
+def test_map_sequence_animator_wcs_simple_plot(test_map_sequence):
+    map_animator = MapSequenceAnimator(test_map_sequence)
+    return map_animator.fig
```
dpshelio, line 31: What are this test (and all others that are "returning") testing?

jeffreypaul15: I've tested this class similar to the ```arrayanimator``` class where the basic plots are just being tested, was that wrong?

sunpy/visualization/animator/tests/test_mapsequence.py
```diff
@@ -0,0 +1,72 @@
...
+        button_labels=bl)
+    map_animator.updatefig(
+        1,
+        map_animator.im,
+        map_animator.sliders[0]._slider)
+    return map_animator.fig
+
+
+@figure_test
+def test_map_sequence_animator_wcs_colorbar_buttons_default_labels(
```
nabobalis, line 63: What are these tests testing?

jeffreypaul15: This just tests whether the display of the buttons are happening, drawing parrallels from arrayanimator like I previously mentioned.

nabobalis: I see, do we need 2 of them?

jeffreypaul15: Also, not too sure if getting the buttons from the matplotlib figure are possible (for testing via code isntead of figure tests).

nabobalis: I guess my point of view is that, we should map out what the API is of MapsequenceAnimator and work towards testing as much of it a we can. Ideally with code if we can and using figure test where either a pure code test is difficult/impossible or we want to verify something purely visual.

jeffreypaul15: Hmm okay, the arrayanimator class is also tested this way with respect to the buttons and figures, and I couldn't find any other method of testing the buttons through code. Do you have any suggestions?

nabobalis: No, I've not worked on these kind of tests before nor am I too familiar with MapsequenceAnimator.

jeffreypaul15: Okay then, I shall let this be. I'll add 2 more methods which test a few more functions of MapsequenceAnimator in a bit and then you can make a decision on what to do.

nabobalis:
Looking at our codecov (you will need to login via github unfortunately): https://codecov.io/gh/sunpy/sunpy/src/7d2b42b73a4130846b32ba3c9ecb2b22390b8e89/sunpy/map/mapsequence.py
https://codecov.io/gh/sunpy/sunpy/src/7d2b42b73a4130846b32ba3c9ecb2b22390b8e89/sunpy/visualization/animator/mapsequenceanimator.py

If I am reading it right, in fact MapSequenceAnimator has at least some test coverage while the plot method for MapSequence does not.


So I think we should also add a figure test for MapSequence.plot while we are here.

sunpy/visualization/animator/tests/test_mapsequence.py
nabobalis: I feel like these two tests that deal with the "wcs_colorbar" probably don't really add anything.

sunpy/visualization/animator/tests/test_mapsequence.py
```diff
@@ -0,0 +1,75 @@
...
+    assert isinstance(map_animator.axes, WCSAxes)
+    start_img = map_animator.plot_start_image(map_animator.axes)
+    assert isinstance(start_img.axes, WCSAxes)
+
+
+@figure_test
+def test_map_sequence_animator_wcs_update_plot(test_map_sequence):
+    map_animator = MapSequenceAnimator(test_map_sequence)
+    map_animator.updatefig(1, map_animator.im, map_animator.sliders[0]._slider)
+    return map_animator.fig
```
nabobalis, line 46: I feel like this is possible to test without a figure.

jeffreypaul15: I suppose, I'll make this change soon

jeffreypaul15:
> Looking at our codecov (you will need to login via github unfortunately): https://codecov.io/gh/sunpy/sunpy/src/7d2b42b73a4130846b32ba3c9ecb2b22390b8e89/sunpy/map/mapsequence.py
> https://codecov.io/gh/sunpy/sunpy/src/7d2b42b73a4130846b32ba3c9ecb2b22390b8e89/sunpy/visualization/animator/mapsequenceanimator.py
> 
> If I am reading it right, in fact MapSequenceAnimator has at least some test coverage while the plot method for MapSequence does not.
> 
> So I think we should also add a figure test for MapSequence.plot while we are here.

I shall test that at  `test_mapsequence.py` then?

nabobalis:
Yeah.

sunpy/visualization/animator/tests/test_mapsequence.py
jeffreypaul15: How exactly would the tests change? `updatefig` changes the current figures data and axes.

sunpy/visualization/animator/tests/test_mapsequence.py
jeffreypaul15: Should I get rid of them then?

nabobalis: Yeah, I don't see the point in them. Especially having two of them.

jeffreypaul15: The same base animator is tested in array animator so I guess I shall remove them.

jeffreypaul15:
@nabobalis I've updated the tests accordingly and added a test for plot. Let me know if this is fine

nabobalis:
You need to rebase the PR and I think a trivial changelog entry makes sense.

jeffreypaul15:
> You need to rebase the PR and I think a trivial changelog entry makes

jeffreypaul15:
> You need to rebase the PR and I think a trivial changelog entry makes sense.

I accidentally rebased this after pulling from the updated code, would this be an issue?

jeffreypaul15:
> You need to rebase the PR and I think a trivial changelog entry makes sense.



> You need to rebase the PR and I think a trivial changelog entry makes sense.

Having a bit of trouble with this, I'm not too sure as to what's the need for rebase.

nabobalis:
> > You need to rebase the PR and I think a trivial changelog entry makes sense.
> 
> > You need to rebase the PR and I think a trivial changelog entry makes sense.
> 
> Having a bit of trouble with this, I'm not too sure as to what's the need for rebase.

You can merge in master if it's easier. There is conflict with master and this needs to be addressed.

jeffreypaul15:
> > > You need to rebase the PR and I think a trivial changelog entry makes sense.
> > 
> > 
> > > You need to rebase the PR and I think a trivial changelog entry makes sense.
> > 
> > 
> > Having a bit of trouble with this, I'm not too sure as to what's the need for rebase.
> 
> You can merge in master if it's easier. There is conflict with master and this needs to be addressed.

So what exactly would I have to do?

jeffreypaul15:
Actually, I shall reopen this PR. Its a lot easier

Issue #5043, opened 2021-03-01
Creating a SUVIMap from SUVI L1b/L2 files is broken in SunPy 2.1.1


cbethge:
### Description

The sunpy.map.sources.SUVIMap class expects data/header input from astropy.io.fits. Probably due to changes in WCS-related modules from SunPy 2.0.1 -> 2.1.1, creating a SUVIMap from SUVI L1b/L2 files is no longer working. Metadata in the headers seems to be missing to make it work, but I can't tell which additional metadata is required now.

### Expected vs actual behavior

**Expected behaviour:** when creating a SUVIMap from data/header inputs in SunPy 2.0.1, it would throw an AstropyDeprecationWarning ("WARNING: AstropyDeprecationWarning: The private astropy._erfa module has been made into its own package, pyerfa, which is a dependency of astropy and can be imported directly using "import erfa" [astropy._erfa]"), but it would still create it.

**Actual behaviour in SunPy 2.1.1:** Quits with "TypeError: Can not create a MetaDict from this type input".

### Steps to Reproduce

```
In [1]: from astropy.io import fits
   ...: from sunpy.map.sources import SUVIMap
   ...: hdu = fits.open('dr_suvi-l2-ci195_g16_s20200220T180000Z_e20200220T180400Z_v1-0-0.fits')
   ...: suvimap = SUVIMap(hdu[1].data, hdu[1].header)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-777075e1d3d8> in <module>
      2 from sunpy.map.sources import SUVIMap
      3 hdu = fits.open('dr_suvi-l2-ci195_g16_s20200220T180000Z_e20200220T180400Z_v1-0-0.fits')
----> 4 suvimap = SUVIMap(hdu[1].data, hdu[1].header)

~/opt/anaconda3/lib/python3.7/site-packages/sunpy/map/sources/suvi.py in __init__(self, data, header, **kwargs)
     72     def __init__(self, data, header, **kwargs):
     73 
---> 74         super().__init__(data, header, **kwargs)
     75 
     76         # Fill in some missing info

~/opt/anaconda3/lib/python3.7/site-packages/sunpy/map/mapbase.py in __init__(self, data, header, plot_settings, **kwargs)
    187                           "Data will be truncated to the first two dimensions.", SunpyUserWarning)
    188 
--> 189         super().__init__(data, meta=MetaDict(header), **kwargs)
    190 
[...]

PR #5047, opened 2021-03-01
Be more lenient when constructing a MetaDict

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/5043. Instead of checking explicitly for a `dict`, try `dict(object)` instead, and error if that isn't possible.

### DIFF
--- a/sunpy/util/metadata.py
+++ b/sunpy/util/metadata.py
@@ -3,6 +3,7 @@ This module provides a generalized dictionary class that deals with header
 parsing, normalization, and maintaining coherence between keys and keycomments.
 """
 from collections import OrderedDict
+from collections.abc import Mapping
 
 __all__ = ['MetaDict']
 
@@ -34,8 +35,11 @@ class MetaDict(OrderedDict):
             adict = args[0]
             if isinstance(adict, list) or isinstance(adict, tuple):
                 items = adict
-            elif isinstance(adict, dict):
-                items = adict.items()
+            elif isinstance(adict, Mapping):
+                # Cast to a dict here, since a Mapping doesn't have to define
+                # a .items() method (but has enough methods to be converted to
+                # a dict)
+                items = dict(adict).items()
             else:
                 raise TypeError(f"Can not create a MetaDict from this input of type {type(adict)}")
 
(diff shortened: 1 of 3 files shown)


sunpy/util/metadata.py
```diff
@@ -34,10 +34,12 @@ def __init__(self, *args):
             adict = args[0]
             if isinstance(adict, list) or isinstance(adict, tuple):
                 items = adict
-            elif isinstance(adict, dict):
-                items = adict.items()
             else:
-                raise TypeError(f"Can not create a MetaDict from this input of type {type(adict)}")
+                try:
+                    items = dict(adict).items()
```
nabobalis, line 39: Why does the type check for the dict not wokr?

nabobalis:
I am unsure if we need to backport this beyond 2.1, but I added the tag just in case.

nabobalis:
If you want to add a changelog, I leave that up to you.

sunpy/util/metadata.py
hayesla: I've also just tried this (i.e. `dict(adict).items()` from the example before from #5043  and it doesn't work, but `dict(adict)` does

dstansby: I think classes can implement special methods that tell python how to convert them to `dict`, without themselves being a child class of `dict`.

dstansby:
Added a changelog. Just backporting to 2.1 should be fine, I removed the 2.0 label.

nabobalis:
A regression test just to make sure for future?

sunpy/util/metadata.py
```diff
@@ -34,10 +34,12 @@ def __init__(self, *args):
             adict = args[0]
             if isinstance(adict, list) or isinstance(adict, tuple):
                 items = adict
-            elif isinstance(adict, dict):
```
Cadair, line 37: A better check might be to check for an instance of `collections.abc.MutableMapping`? It's generally better to avoid try except blocks if you can and I am like 90% sure that `fits.Header` is a Mapping.

dstansby: Good idea - I don't think being mutable matters here, so I just went for `Mapping`.

dstansby:
Still need to add a test, which it turns out is non-trivial due to the level of non-FITS compliance in the sample data.

dstansby:
Test added

Cadair (approved):

PR #5051, opened 2021-03-02
Raise warning in GenericMap.rotate() if `missing` is incompatible with the data array

### LINKED ISSUE #5036: Map.rotate pads with zeros in addition to the value of "missing"
### Description

`Map.rotate()` pads the input data when rotating. The `missing` keyword provides a value for filling the missing values. But this value is not used everywhere where I expect it: there is also a stripe of zero values.

The value of `missing` should be used in all pixels of the output map where there is no value from the input map (after rotation).

This is not the case: some of these pixels have a value of zero.
 
### Steps to Reproduce

```python
from sunpy.data.sample import AIA_171_ROLL_IMAGE
import sunpy.map
import numpy as np

aiamap = sunpy.map.Map(AIA_171_ROLL_IMAGE)
aia_aligned = aiamap.rotate(order=3, missing=np.nan)
aia_aligned
```
In a notebook, this gives
image

The red pixels are the expected NaNs, the black stripe is unexpected zeros (it should also be NaNs), as can be displayed with `aia_aligned.peek()`.

The result is the same with `order=1`.

#######
General
#######
Time : Friday, 26. February 2021 09:29PM UT
System : Linux
Processor : x86_64
Arch : 64bit
SunPy : 2.1.0
OS: Ubuntu 20.10 groovy (Linux 5.8.0-43-generic x86_64)

##################
Required Libraries
##################
Python: 3.8.6
NumPy: 1.19.4
SciPy: 1.5.4
matplotlib: 3.3.3
Astropy: 4.1
Pandas: 1.1.0
parfive: 1.1.1

#####################
Recommended Libraries
#####################
beautifulsoup: 4.9.1
PyQt4: NOT INSTALLED
PyQt5: 5.15.0
Zeep: 3.4.0
Sqlalchemy: 1.3.19
drms: 0.6.1

jeffreypaul15:
### Description

Fixes #5036
Raises a warning if `missing` datatype is not compatible with GenericMap data

### DIFF
--- a/changelog/5051.bugfix.rst
+++ b/changelog/5051.bugfix.rst
@@ -1,2 +1,2 @@
-Added conversion of `int` to `float` when `sunpy.map.GenericMap.rotate` is called,
-fixing faulty plots when `skimage` is used.
+Added warning to `sunpy.map.GenericMap.rotate` when specified `missing` value is not compatible
+with `GenericMap`.
--- a/changelog/5051.bugfix.rst
+++ b/changelog/5051.bugfix.rst
@@ -1,2 +1,2 @@
-Added warning to `sunpy.map.GenericMap.rotate` when specified ``missing`` value is not compatible
-with the data type of the `~sunpy.map.GenericMap`.
+Added warning to :meth:`sunpy.map.GenericMap.rotate` when specified ``missing`` value is not compatible
+with the number type of the data array.
--- a/changelog/5051.bugfix.rst
+++ b/changelog/5051.bugfix.rst
@@ -1,2 +1,2 @@
-Added warning to `sunpy.map.GenericMap.rotate` when specified `missing` value is not compatible
-with `GenericMap`.
+Added warning to `sunpy.map.GenericMap.rotate` when specified ``missing`` value is not compatible
+with the data type of the `~sunpy.map.GenericMap`.
--- a/changelog/5051.bugfix.rst
+++ b/changelog/5051.bugfix.rst
@@ -1,0 +1,2 @@
+Added conversion of `int` to `float` when `sunpy.map.GenericMap.rotate` is called,
+fixing faulty plots when `skimage` is used.
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1484,10 +1484,9 @@ class GenericMap(NDData):
         pad_x = int(np.max((diff[1], 0)))
         pad_y = int(np.max((diff[0], 0)))
 
-        if version.parse(np.__version__) >= version.parse("1.20.0"):
-            raise ImportError('Missing NaN value is not supported for Integer map')
-
         if issubclass(self.data.dtype.type, numbers.Integral) and (missing == np.nan):
+            if version.parse(np.__version__) >= version.parse("1.20.0"):
+                raise ImportError('Missing NaN value is not supported for Integer map')
             warnings.warn("Integer map data is incompatible with specified missing value",
                           SunpyUserWarning)
 
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1482,7 +1482,10 @@ class GenericMap(NDData):
         pad_x = int(np.max((diff[1], 0)))
         pad_y = int(np.max((diff[0], 0)))
 
-        new_data = np.pad(self.data,
+        if issubclass(aiamap.data.dtype.type, numbers.Integral):
+            new_data = self.data.astype(np.float64)
+
+        new_data = np.pad(new_data,
                           ((pad_y, pad_y), (pad_x, pad_x)),
                           mode='constant',
                           constant_values=(missing, missing))
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1443,6 +1443,7 @@ class GenericMap(NDData):
         -----
         This function will remove old CROTA keywords from the header.
         This function will also convert a CDi_j matrix to a PCi_j matrix.
+        This function always type casts integer data to float
 
         See :func:`sunpy.image.transform.affine_transform` for details on the
         transformations, situations when the underlying data is modified prior
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -3,6 +3,7 @@ Map is a generic Map class from which all other Map classes inherit from.
 """
 import copy
 import html
+import numbers
 import textwrap
 import warnings
 import webbrowser
@@ -1482,7 +1483,9 @@ class GenericMap(NDData):
         pad_x = int(np.max((diff[1], 0)))
         pad_y = int(np.max((diff[0], 0)))
 
-        if issubclass(aiamap.data.dtype.type, numbers.Integral):
+        if issubclass(self.data.dtype.type, numbers.Integral):
+            warnings.warn("Integer input data has been cast to float64.",
+                          SunpyUserWarning)
             new_data = self.data.astype(np.float64)
 
         new_data = np.pad(new_data,
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1484,7 +1484,7 @@ class GenericMap(NDData):
         pad_y = int(np.max((diff[0], 0)))
 
         if issubclass(self.data.dtype.type, numbers.Integral) and isinstance((missing), float):
-            warnings.warn("Integer map data is incompatibe with specified missing value",
+            warnings.warn("Integer map data is incompatible with specified missing value",
                           SunpyUserWarning)
 
         new_data = np.pad(self.data,
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -17,7 +17,6 @@ import numpy as np
 from matplotlib import cm
 from matplotlib.backend_bases import FigureCanvasBase
 from matplotlib.figure import Figure
-from packaging import version
 
 import astropy.units as u
 import astropy.wcs
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1483,10 +1483,11 @@ class GenericMap(NDData):
         pad_x = int(np.max((diff[1], 0)))
         pad_y = int(np.max((diff[0], 0)))
 
-        if issubclass(self.data.dtype.type, numbers.Integral) and (missing % 1 != 0)):
-            warnings.warn("The specified `missing` value is not an integer, but the data "
-                          "array is of integer type, so the output may be strange.",
-                          SunpyUserWarning)
+        if issubclass(self.data.dtype.type, numbers.Integral) and (missing % 1 != 0):
+            warnings.warn(
+                "The specified `missing` value is not an integer, but the data "
+                "array is of integer type, so the output may be strange.",
+                SunpyUserWarning)
 
         new_data = np.pad(self.data,
                           ((pad_y, pad_y), (pad_x, pad_x)),
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1443,7 +1443,6 @@ class GenericMap(NDData):
         -----
         This function will remove old CROTA keywords from the header.
         This function will also convert a CDi_j matrix to a PCi_j matrix.
-        This function always type casts integer data to float
 
         See :func:`sunpy.image.transform.affine_transform` for details on the
         transformations, situations when the underlying data is modified prior
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1488,7 +1488,7 @@ class GenericMap(NDData):
                           SunpyUserWarning)
             new_data = self.data.astype(np.float64)
         else:
-             new_data = self.data
+            new_data = self.data
 
         new_data = np.pad(new_data,
                           ((pad_y, pad_y), (pad_x, pad_x)),
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1484,10 +1484,9 @@ class GenericMap(NDData):
         pad_x = int(np.max((diff[1], 0)))
         pad_y = int(np.max((diff[0], 0)))
 
-        if issubclass(self.data.dtype.type, numbers.Integral) and (np.isnan(missing)):
-            if version.parse(np.__version__) >= version.parse("1.20.0"):
-                raise ImportError('Missing NaN value is not supported for Integer map')
-            warnings.warn("Integer map data is incompatible with specified missing value",
+        if issubclass(self.data.dtype.type, numbers.Integral) and (missing % 1 != 0)):
+            warnings.warn("The specified `missing` value is not an integer, but the data "
+                          "array is of integer type, so the output may be strange.",
                           SunpyUserWarning)
 
         new_data = np.pad(self.data,
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1484,7 +1484,7 @@ class GenericMap(NDData):
         pad_x = int(np.max((diff[1], 0)))
         pad_y = int(np.max((diff[0], 0)))
 
-        if issubclass(self.data.dtype.type, numbers.Integral) and (missing == np.nan):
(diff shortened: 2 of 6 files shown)


sunpy/map/mapbase.py
```diff
@@ -1482,7 +1483,14 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=4, scale=1.0,
         pad_x = int(np.max((diff[1], 0)))
         pad_y = int(np.max((diff[0], 0)))
 
-        new_data = np.pad(self.data,
+        if issubclass(self.data.dtype.type, numbers.Integral):
+            warnings.warn("Integer input data has been cast to float64.",
```
dstansby, line 1487: Someone might correct me, but I don't think we need to add a new warning here, as the data is always cast to a `float` when `rotate` is used.

dstansby (dismissed):
Looks like a good approach, but I think it's worth
- Adding a note to the docstring (under the notes section) that integer data is always cast to float64.
- Adding a test to make sure the original issue is fixed

sunpy/map/mapbase.py
jeffreypaul15: An explicit warning is given by ```sunpy.image.transform``` when converted to float, considering this step is skipped, I thought of adding the warning here.

jeffreypaul15:
>     * Adding a test to make sure the original issue is fixed

How would the test have to be written? Should it be a figure test?

sunpy/map/mapbase.py
dstansby: Ah, that makes sense.

jeffreypaul15:
> Looks like a good approach, but I think it's worth
> 
>     * Adding a note to the docstring (under the notes section) that integer data is always cast to float64.
> 
>     * Adding a test to make sure the original issue is fixed

I've written a test for where the ```skimage``` rotation is compared with ```scipy``` rotation of the map. ```affine_transform```  from ```scipy``` currently still converts the ```np.nan``` to ```0```, so the image is crosschecked with the number of ```nan``` in both the rotated maps. Not too sure if this is the right way to go about this though.

sunpy/map/mapbase.py
```diff
@@ -1482,7 +1484,14 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=4, scale=1.0,
         pad_x = int(np.max((diff[1], 0)))
         pad_y = int(np.max((diff[0], 0)))
 
-        new_data = np.pad(self.data,
+        if issubclass(self.data.dtype.type, numbers.Integral):
```
Cadair, line 1487: I do not think that always casting to float64 is a good idea. People often want to use rotate-like functionality to rotate many maps, and if you take a unit8 image as input here and cast it up to float64 you are multiplying the memory usage by a factor of 8! I know that if scikit-image is installed by default we cast up anyway, but with efforts like #4502 I really don't think doing this in an implementation-agnostic way is a good idea.

Why don't we raise a warning if missing is set to nan and the dtype is not float?

jeffreypaul15: I guess so, do we raise a warning and then proceed to cast it when missing=np.nan or just raise a warning and let the erroneous plot show up?

Cadair: The other option is to raise an exception if the missing value doesn't have the same dtype (or a compatible one) with the data array? I am not sure if that's a good idea though.

jeffreypaul15: The plots do say the "bad values (NaN) are given in red". If nan is considered bad then I guess raising an warning and converting to 0 is better?

Cadair: Another option is we throw a warning and then cast the data array to float64 if the missing value is an incompatible dtype with the array.

It basically comes down to which dtype do we want to prioritise. If the dtype of the array is the most important then we should probably error if missing is incompatible with the array dtype (which is potentially problematic if the default causes an exception). If we want to prioritise the dtype of missing then we should cast the array to a compatible dtype if it is incompatible with the missing value.

I think I need some thoughts from other people on this.

nabobalis: Do we have no control over the missing value?

jeffreypaul15: I suppose, I shall leave this PR open for that then, thanks!

jeffreypaul15: @nabobalis depending on how it's being cast, if passing np.nan (float) to an int map causes this issue.

nabobalis:
I think we decided that we would add a warning where the missing_value is not compatible with the datatype of the data and do no changes otherwise and leave that up to you the user.

sunpy/map/mapbase.py
```diff
@@ -1442,6 +1443,7 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=4, scale=1.0,
         -----
         This function will remove old CROTA keywords from the header.
         This function will also convert a CDi_j matrix to a PCi_j matrix.
+        This function always type casts integer data to float
```
nabobalis, line 1446: Is this true?

sunpy/map/mapbase.py
```diff
@@ -1482,6 +1484,10 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=4, scale=1.0,
         pad_x = int(np.max((diff[1], 0)))
         pad_y = int(np.max((diff[0], 0)))
 
+        if issubclass(self.data.dtype.type, numbers.Integral):
+            warnings.warn("Integer map data is incompatilbe with specified missing value",
```
nabobalis, line 1488: I think a test to make sure this is raised would be nice.

sunpy/map/mapbase.py
jeffreypaul15: Forgot about that, need remove it.

sunpy/map/mapbase.py
jeffreypaul15: I'm not sure as to what datatype the testing map is, should I have to load a separate map in the test to make sure this works?

nabobalis: You can find out by loading the map? Or you can make a fake one with a fixed data type. Both would work.

nabobalis, line 1488:
```suggestion
            warnings.warn("Integer map data is incompatible with specified missing value",
```

sunpy/map/mapbase.py
```diff
@@ -1482,6 +1484,10 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=4, scale=1.0,
         pad_x = int(np.max((diff[1], 0)))
         pad_y = int(np.max((diff[0], 0)))
 
+        if issubclass(self.data.dtype.type, numbers.Integral):
```
nabobalis, line 1487: Do we not need to check the missing value as well?

sunpy/map/mapbase.py
jeffreypaul15: Sorry, I should really spell check.

jeffreypaul15: Yeah. I'll get this done.

sunpy/map/mapbase.py
jeffreypaul15: yeah, it slipped my mind. I shall make a commit soon fixing all this.

jeffreypaul15:
Okay, so the newer versions of Numpy do not allow padding of `int` and `nan` datatypes hence the failing  Core Tests py39 [linux] test. That essentially solves this entire issue right?

nabobalis:
Not everyone has newer versions of numpy installed.

changelog/5051.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Added warning to `sunpy.map.GenericMap.rotate` when specified `missing` value is not compatible
+with `GenericMap`.
```
nabobalis, line 2:
```suggestion
Added warning to `sunpy.map.GenericMap.rotate` when specified ``missing`` value is not compatible
with the data type of the `~sunpy.map.GenericMap`.
```

sunpy/map/mapbase.py
```diff
@@ -1482,6 +1483,10 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=4, scale=1.0,
         pad_x = int(np.max((diff[1], 0)))
         pad_y = int(np.max((diff[0], 0)))
 
+        if issubclass(self.data.dtype.type, numbers.Integral) and isinstance((missing), float):
+            warnings.warn("Integer map data is incompatibe with specified missing value",
```
nabobalis, line 1487:
```suggestion
            warnings.warn("Integer map data is incompatible with specified missing value",
```
If the value was a float of 1.0, that is fine as an int, even if it was 1.5, it gets rounded by you can cast the value.

I wonder if this is only for nans.

jeffreypaul15: any `float` value except for `.0` will result in the same issue, just checked. So I guess unless the user specifies a `missing=1.0`. Either way the warning should be raised right?

jeffreypaul15:
> Not everyone has newer versions of numpy installed.

Should I run pytest.mark.xflail then?

sunpy/map/mapbase.py
jeffreypaul15: I wonder if raising a different message such as "irregular behavior might occur" is more appropriate?

nabobalis: I am not sure I want to raise a warning when a float is also a int without a loss of information.
Otherwise yeah I think for 1.1, it should raise the warning.

nabobalis:
> > Not everyone has newer versions of numpy installed.
> 
> Should I run `pytest.mark.xfail` then?

No, we should handle that error in our code.

jeffreypaul15:
> > > Not everyone has newer versions of numpy installed.
> > 
> > 
> > Should I run `pytest.mark.xfail` then?
> 
> No, we should handle that error in our code.

But shouldn't I run the test only if numpy is below a certain version?

nabobalis:
> > > > Not everyone has newer versions of numpy installed.
> > > 
> > > 
> > > Should I run `pytest.mark.xfail` then?
> > 
> > 
> > No, we should handle that error in our code.
> 
> But shouldn't I run the test only if numpy is below a certain version?

We run numpy on a range of versions in our test suite depending on the python version. We have a fix min version and it has to work on that as well as the latest version.

jeffreypaul15:
> > > > > Not everyone has newer versions of numpy installed.
> > > > 
> > > > 
> > > > Should I run `pytest.mark.xfail` then?
> > > 
> > > 
> > > No, we should handle that error in our code.
> > 
> > 
> > But shouldn't I run the test only if numpy is below a certain version?
> 
> We run numpy on a range of versions in our test suite depending on the python version. We have a fixed min version and it has to work on that as well as the latest version.

How would I write this test then?

nabobalis:
Not sure, we probably need to catch the error for newer versions of numpy and then raise forward it with a nicer message.

jeffreypaul15:
> message

This change takes place in version `1.20.0` of numpy but the issue arises only when `nan` is passed, if any other value is passed `np.pad` converts it to 0.

nabobalis:
Hmm, that is tricky.

jeffreypaul15:
> Hmm, that is tricky.

So ideally the warning should not only contain `nan` is incompatible but also state that when Integer map data is passed, passing any float value as missing would trigger a faulty plot.

jeffreypaul15:
> Hmm, that is tricky.

Turns out I was wrong, any float value is ceiled and only `nan` results in a true faulty plot, although,  on running 
```python
from sunpy.data.sample import AIA_171_ROLL_IMAGE
import sunpy.map
import numpy as np

aiamap = sunpy.map.Map(AIA_171_ROLL_IMAGE)
aia_aligned = aiamap.rotate(order=3, missing=1.9)
aia_aligned
```
This faulty plot shows up
image
but the data is consistent with replacing the missing values as
```python
plt.imshow(aia_aligned.data, vmax=100)
``` 
gives :
image

So `aia_aligned` on a notebook will result in a faulty plot but the `aia_aligned` is consistent with the correct behavior of how this is supposed to work

ayshih:
It's misleading to call the plot "faulty"; it's doing exactly what it should be doing.

If `missing` is not an integer (e.g., 1.9), but the array is of integer `dtype`, the padding in `.rotate()` before the `affine_transform()` call will pad by the integer amount (1), while the filling within `affine_transform()` will be by the float amount (1.9) because the array has already been cast to a float.  Thus, for `missing=1.9`, there are sections of the resulting data array that are 1 and other sections that are 1.9.

The notebook graphical representation (which is the same as the output of `.quiclook()`) uses histogram equalization for its normalization, which makes the difference between 1 and 1.9 glaringly apparent.  The difference would not be visible with typical linear or power-law normalizations.

jeffreypaul15:
> It's misleading to call the plot "faulty"; it's doing exactly what it should be doing. The problem is in the data array.
> 
> If `missing` is not an integer (e.g., 1.9), but the array is of integer `dtype`, the padding in `.rotate()` before the `affine_transform()` call will pad by the integer amount (1), while the filling within `affine_transform()` will be by the float amount (1.9) because the array has already been cast to a float. Thus, for `missing=1.9`, there are sections of the resulting data array that are 1 and other sections that are 1.9.
> 
> The notebook graphical representation (which is the same as the output of `.quiclook()`) uses histogram equalization for its normalization, which makes the difference between 1 and 1.9 glaringly apparent because they each are associated with large numbers of pixels. The difference would not be visible with typical linear or power-law normalizations.

Ah, thanks @ayshih. I was quite confused about this.

jeffreypaul15:
@nabobalis I've added a test to check for numpy version and I've also raised an appropriate error as well as warning if the values aren't the same dtype as Integer.

sunpy/map/mapbase.py
```diff
@@ -1482,6 +1484,12 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=4, scale=1.0,
         pad_x = int(np.max((diff[1], 0)))
         pad_y = int(np.max((diff[0], 0)))
 
+        if issubclass(self.data.dtype.type, numbers.Integral) and (np.isnan(missing)):
+            if version.parse(np.__version__) >= version.parse("1.20.0"):
+                raise ImportError('Missing NaN value is not supported for Integer map')
+            warnings.warn("Integer map data is incompatible with specified missing value",
+                          SunpyUserWarning)
```
ayshih, line 1491:
```suggestion
        if issubclass(self.data.dtype.type, numbers.Integral) and (missing % 1 != 0)):
            warnings.warn("The specified `missing` value is not an integer, but the data "
                          "array is of integer type, so the output may be strange.",
                          SunpyUserWarning)
```
- As noted in the discussion, we ought to raise a warning for all non-integer `missing`, not just NaN.
- Ditch the version check for NumPy.  It's going to error anyway from NumPy, and our warning message makes it clear why.  (Plus, the error shouldn't be `ImportError`.)

sunpy/map/mapbase.py
```diff
@@ -16,6 +17,7 @@
 from matplotlib import cm
 from matplotlib.backend_bases import FigureCanvasBase
 from matplotlib.figure import Figure
+from packaging import version
```
ayshih, line 20:
```suggestion
```

changelog/5051.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Added warning to `sunpy.map.GenericMap.rotate` when specified ``missing`` value is not compatible
+with the data type of the `~sunpy.map.GenericMap`.
```
ayshih, line 2:
```suggestion
Added warning to :meth:`sunpy.map.GenericMap.rotate` when specified ``missing`` value is not compatible
with the number type of the data array.
```

ayshih (changes requested):
And update the test to match

jeffreypaul15:
@ayshih won't the tests remain the same as above numpy version `1.20.0` it would cause an error? Or should I use xfail instead?

ayshih:
> Or should I use xfail instead?

Yes, that is what I'd suggest.

Cadair (approved):

ayshih (approved):
Boom

nabobalis:
I've backported this to 2.0 myself.

PR #5053, opened 2021-03-03
Remove global vars in cm/color_tables

nabobalis:
Removes the globals.

### DIFF
--- a/sunpy/visualization/colormaps/color_tables.py
+++ b/sunpy/visualization/colormaps/color_tables.py
@@ -14,34 +14,85 @@ __all__ = [
     'aia_color_table', 'sswidl_lasco_color_table', 'eit_color_table',
     'sxt_color_table', 'xrt_color_table', 'trace_color_table',
     'sot_color_table', 'hmi_mag_color_table', 'suvi_color_table',
-    'rhessi_color_table', 'std_gamma_2', 'euvi_color_table',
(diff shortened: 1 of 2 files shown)


dstansby (changes requested):
Seems a bit overkill (especially for backporting) when the fix could just be a a copy statement when creating `solohri_lya1216`?

nabobalis:
I think its the right amount of kill. 

We have too many globals floating around and it's caught us here and it should never happen again.

nabobalis:
Figure tests need updating since the colourtable figure had the bug in it.

Cadair:
Do we not want to do this on master? I do not like having the globals around even if they are not writeable.

sunpy/visualization/colormaps/color_tables.py
```diff
@@ -312,51 +369,10 @@ def std_gamma_2():
...
     try:
         return cmap_from_rgb_file(f'EUVI {str(wavelength)}', f'euvi_{int(wavelength.value)}.csv')
     except OSError:
         raise ValueError(
             "Invalid EUVI wavelength. Valid values are "
             "171, 195, 284, 304."
         )
-
-
-def cmap_from_rgb_file(name, fname):
```
nabobalis, line 325: I moved this first just because they get called by other functions and I can't handle them being at the end of the file.

nabobalis:
I've removed the backport and rebased this.

dstansby (approved):

Issue #5052, opened 2021-03-03
Issues with AIA color tables?


cbethge:
### Description

First of all, no one else seems to have this issue, so maybe it is a problem with my installation. I was wondering if anyone can reproduce it. The issue: I am getting odd colortables for AIA 171 and 304 when I call them directly. I recently updated to SunPy 2.1.1.

### Expected vs Actual behavior

Expected behaviour for plotting an AIA 304 image with the 304 colortable is the "normal" one that looks like a heatmap. 
Actual behaviour is getting a pink-ish colortable. This does not happen when I call peek() or plot() in sunpy.map.

### Steps to Reproduce

```
from astropy.io import fits
from astropy import units as u
from sunpy.visualization.colormaps import color_tables as spct
import matplotlib.pyplot as plt
import matplotlib.colors as colors
hdu = fits.open('aia.lev1.304A_2020-10-29T13_40_05.12Z.image_lev1.fits')
fig = plt.figure(figsize=(10,10))
fig.subplots_adjust(bottom=0, top=1, left=0, right=1)
plt.imshow(hdu[1].data, norm=colors.PowerNorm(gamma=0.5, vmin=0, vmax=75), cmap=spct.aia_color_table(304*u.angstrom), origin='lower')
plt.axis('off')
plt.show()
```

![Figure_1](https://user-images.githubusercontent.com/68035714/109846442-70917f00-7c0b-11eb-8bc4-aff3f45a1286.jpg)

```
import sunpy.map
aia_map = sunpy.map.Map('aia.lev1.304A_2020-10-29T13_40_05.12Z.image_lev1.fits')
aia_map.peek()
```

![Figure_2](https://user-images.githubusercontent.com/68035714/109846825-ca924480-7c0b-11eb-94c4-75cb60dcad55.jpg)

### System Details

- SunPy Version: 2.1.1
- Astropy Version: 4.2
- Python Version: 3.7.10
- OS information: Mac OS X 10.16 (i386)

nabobalis:
Can you tell me what `spct.aia_color_table(304*u.angstrom)` returns?

cbethge:
Sure!

```
In [2]: spct.aia_color_table(304*u.angstrom)
Out[2]: <matplotlib.colors.LinearSegmentedColormap at 0x7fb826abf950>
```

nabobalis:
That did not tell me what I had hoped for. 

If I do:

```python
from astropy.io import fits
from astropy import units as u
[...]

PR #5054, opened 2021-03-03
Stop color table arrays from being modified in place

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/5052; a simpler alternative to https://github.com/sunpy/sunpy/pull/5053 (which I still think is worth putting in master, but is too much refactoring for a bugfix).


nabobalis (approved):

nabobalis:
Just need to update the figure hashes.

Issue #5064, opened 2021-03-10
Venus transit gallery example FOV is much too wide


wtbarnes:
<img width="1141" alt="image" src="https://user-images.githubusercontent.com/8676141/110674084-88cf4400-819f-11eb-8863-8c358c212148.png">

There is something wonky going on with the `submap` call

nabobalis:
Once this has been tracked down, a figure test replicating this bug should be added to prevent this happening agian.

PR #5069, opened 2021-03-10
revert to add_artist for draw_limb to fix bug

nabobalis:
Check the venus example to make sure it worked.

Fixes https://github.com/sunpy/sunpy/issues/5064

### DIFF
--- a/changelog/5069.bugfix.rst
+++ b/changelog/5069.bugfix.rst
@@ -1,0 +1,1 @@
+Reverted change for `.draw_limb` that made it use "add_artist" as it was changing the FOV of the plotted image.
--- a/changelog/5069.bugfix.rst
+++ b/changelog/5069.bugfix.rst
@@ -1,1 +1,1 @@
-Reverted change for `.draw_limb` that made it use "add_artist" as it was changing the FOV of the plotted image.
+Reverted change for `sunpy.map.GenericMap.draw_limb` that made it use "add_artist" as it was changing the FOV of the plotted image.
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1977,7 +1977,7 @@ class GenericMap(NDData):
         c_kw.update(kwargs)
 
         circ = patches.Circle([0, 0], **c_kw)
-        axes.add_patch(circ)
+        axes.add_artist(circ)
 
         return [circ]
 


nabobalis:
The example is back: https://sunpy--5069.org.readthedocs.build/en/5069/generated/gallery/units_and_coordinates/venus_transit.html#sphx-glr-generated-gallery-units-and-coordinates-venus-transit-py

Cadair (approved):
Fine with me, would like to know what @dstansby has to say as he understands mpl better than I.

ayshih:
Copying over my understanding from Element:
> #5019 (backported to 2.1.1 and 2.0.8) changed the limb (which is a Matplotlib patch) so that it is added through add_patch() instead of add_artist(). The plot_coord() call triggers Matplotlib to change the plot limits to include the entire limb because it now realizes that it is there.

> That change to draw_limb() can be reverted on all branches. We ought to use add_patch() for patches that are intended to be in a legend, but it might be okay if we assume that the limb doesn't need to be in a legend...

Presumably `plot_coord()` triggers Matplotlib to change the FOV to include all fully registered patches (ones added through `add_patch()`), but doesn't take into account patches that aren't fully registered (ones added through `add_artist()` instead).

dstansby:
I've opened an upstream issue here: https://github.com/matplotlib/matplotlib/issues/19685. This seems 👍 for now, but it doesn't feel like a long term solution to me. Will merge this, and see what MPL folk say on that issue.

dstansby (changes requested):
👍 , but needs a bugfix changelog (unless the `add_patch()` change hasn't found a way into a release yet)

nabobalis:
Changelog added.

dstansby (approved):

dstansby:
Actually it looks like the doc build fail is real, with
```python
<unknown>:34: WARNING: more than one target found for cross-reference 'draw_limb': sunpy.map.CompositeMap.draw_limb, sunpy.map.GenericMap.draw_limb
```

changelog/5069.bugfix.rst
```diff
@@ -0,0 +1 @@
+Reverted change for `.draw_limb` that made it use "add_artist" as it was changing the FOV of the plotted image.
```
dstansby, line 1:
```suggestion
Reverted change for `sunpy.map.GenericMap.draw_limb` that made it use "add_artist" as it was changing the FOV of the plotted image.
```

nabobalis:
I forgot the tilda! Thanks

nabobalis:
Another issue came up: 
```
/home/vsts/work/1/s/docs/guide/data_types/timeseries.rst:263: WARNING: Exception occurred in plotting timeseries-2
```

nabobalis:
> Another issue came up:
> 
> ```
> /home/vsts/work/1/s/docs/guide/data_types/timeseries.rst:263: WARNING: Exception occurred in plotting timeseries-2
> ```

I've fixed this in another PR.

PR #5070, opened 2021-03-10
ignore warning created in figure-dev-deps

nabobalis:


dstansby (changes requested):
We should work out why this is warning, and try and fix it. As a minimum, ignoring this warning needs to have a link to a sunpy issue so we remember ignoring this doesn't get into a release.

nabobalis:
It comes from this PR: https://github.com/astropy/astropy/pull/11260/files and astropy themselves ignore it.

dstansby:
Thanks for the PR link - I think we should look into this further, as we don't want users being spammed with these warnings if they're spurious. I'll try and take a look in the next week or so.

nabobalis:
True, this warning will most likely be emitted when we use the next version of astropy. 

I don't think that holding back on fixing the CI until we investigate this is worth it. We could open an issue to track this?

dstansby:
Yeah, sorry if I wasn't clear earlier, I'm fine with opening an issue and merging this 👍

Cadair:
This warning is most likely emitted from WCSLIB so I don't expect there to be a quick fix. Someone should track this down to a MWE, but in a new issue.

Cadair (approved):

setup.cfg
```diff
@@ -171,6 +171,8 @@ filterwarnings =
     # an astropy table inside a test, which is generally only done when
     # debugging.
     ignore:the imp module is deprecated in favour of importlib:DeprecationWarning:ipykernel.iostream
+    # Recent change to astropy has made our figure tests emit this warning.
```
Cadair, line 174:
```suggestion
    # Recent change to astropy has made our figure tests emit this warning.
    # https://github.com/sunpy/sunpy/issues/5073
```

PR #5084, opened 2021-03-13
Refactored RotatedSunFrame transformations

### LINKED ISSUE #5082: Observer-based RotatedSunFrame can't transform to a non-observer-based frame
```python
>>> from astropy import units as u

>>> from sunpy.coordinates import HeliographicStonyhurst, Helioprojective, RotatedSunFrame

>>> r = RotatedSunFrame(0*u.arcsec, 0*u.arcsec, base=Helioprojective(observer='earth', obstime='2021-01-01'))
>>> r
<RotatedSunHelioprojective Coordinate (base=<Helioprojective Frame (obstime=2021-01-01T00:00:00.000, rsun=695700.0 km, observer=<HeliographicStonyhurst Coordinate for 'earth'>)>, duration=0.0 d, rotation_model=howard): (Tx, Ty) in arcsec
    (0., 0.)>

>>> r.transform_to(HeliographicStonyhurst(obstime=r.base.obstime))
...
ConvertError: This transformation cannot be performed because the Heliocentric frame has observer=None.
```
This is because neither `RotatedSunHelioprojective` nor `HeliographicStonyhurst` have the `observer` frame attribute, but the HCC->HGS transformation step needs to have a defined observer.  (The `observer` frame attribute that you see in the repr for `RotatedSunHelioprojective` is a frame attribute of the `base` frame attribute.)  There are several approaches to fix this, but I need to investigate which approach makes the most sense.

ayshih:
For philosophical reasons, the transformations for `RotatedSunFrame` were written to externally go to/from the corresponding "base" frame, and those transformations would internally go through `HeliocentricInertial` (HCI), which intermediately goes through `HeliographicStonyhurst` (HGS).  Calculating the differential rotation in HGS instead of HCI does not change the result, so the transformation loop HGS->HCI->HGS is unnecessary.  Furthermore, if the overall transformation for `RotatedSunFrame` goes to anything other than the "base" frame, there is an unnecessary transformation detour of going all the way to the "base" frame and then transforming back out of it.

This PR refactors those transformations to externally go to/from HGS.  This strictly reduces the number of transformation legs (i.e., a performance boost) with no change in the calculation result.  It also happens to fix #5082.

### DIFF
--- a/changelog/5084.bugfix.rst
+++ b/changelog/5084.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug where some `~sunpy.coordinates.metaframes.RotatedSunFrame` transformations could fail with an "observer=None" error.
--- a/changelog/5084.trivial.rst
+++ b/changelog/5084.trivial.rst
@@ -1,0 +1,1 @@
+Refactored `~sunpy.coordinates.metaframes.RotatedSunFrame` transformations for improved performance.
--- a/sunpy/coordinates/metaframes.py
+++ b/sunpy/coordinates/metaframes.py
@@ -12,7 +12,7 @@ from sunpy import log
 from sunpy.time import parse_time
 from sunpy.time.time import _variables_for_parse_time_docstring
 from sunpy.util.decorators import add_common_docstring
-from .frames import BaseHeliographic, HeliocentricInertial, SunPyBaseCoordinateFrame
+from .frames import BaseHeliographic, HeliographicStonyhurst, SunPyBaseCoordinateFrame
 from .offset_frame import NorthOffsetFrame
 from .transformations import _transformation_debug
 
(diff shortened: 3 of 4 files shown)


ayshih:
Using hypothesis tests for benchmarking, the transformations are roughly twice as fast.

Before this PR:
```
sunpy/coordinates/tests/test_metaframes.py::test_rotatedsun_transforms[HeliographicStonyhurst]:
    - Typical runtimes: 65-72 ms, ~ 15% in data generation
sunpy/coordinates/tests/test_metaframes.py::test_rotatedsun_transforms[HeliographicCarrington]:
    - Typical runtimes: 156-170 ms, ~ 6% in data generation
sunpy/coordinates/tests/test_metaframes.py::test_rotatedsun_transforms[HeliocentricInertial]:
    - Typical runtimes: 26-30 ms, ~ 36% in data generation
```
After this PR:
```
sunpy/coordinates/tests/test_metaframes.py::test_rotatedsun_transforms[HeliographicStonyhurst]:
    - Typical runtimes: 27-30 ms, ~ 35% in data generation
sunpy/coordinates/tests/test_metaframes.py::test_rotatedsun_transforms[HeliographicCarrington]:
    - Typical runtimes: 96-104 ms, ~ 10% in data generation
sunpy/coordinates/tests/test_metaframes.py::test_rotatedsun_transforms[HeliocentricInertial]:
    - Typical runtimes: 57-63 ms, ~ 17% in data generation
```

nabobalis (approved):

Cadair (approved):

PR #5089, opened 2021-03-14
Fix bug where DataManager breaks on deleting db file

### LINKED ISSUE #3853: Remote data manager does not recover well from its database being deleted while a session is running
### Description
You can permanently corrupt the data manager state. So nothing works.

### Expected behavior
It recovers from my abuse and fixes itself.

### Steps to Reproduce

1. Delete all your data_manager stuff so you have a clean slate.
2. Run the following code inside an interactive prompt:
```python
from sunpy.data import manager
@manager.require('test_file',
                  ['http://data.sunpy.org/sample-data/predicted-sunspot-radio-flux.txt'],
                  '4c85b04a5528aa97eb84a087450eda0421c71833820576330bba148564089b11')
def test_function():
    pass

test_function()
```
3. Observe how everything works.
4. do a rm ~/sunpy/data_manager/*
5. **Without** starting a new python session run: `test_function()`, observe the following error:
```
OperationalError                          Traceback (most recent call last)
<ipython-input-2-7e96d146b416> in <module>
----> 1 test_function()

~/.virtualenvs/sunpy-release/lib/python3.8/site-packages/sunpy/data/data_manager/manager.py in wrapper(*args, **kwargs)
     64                     file_path = self._cache.download(urls, redownload=True)
     65                 else:
---> 66                     details = self._cache.get_by_hash(sha_hash)
     67                     if not details:
     68                         # In case we are matching by hash and file does not exist

~/.virtualenvs/sunpy-release/lib/python3.8/site-packages/sunpy/data/data_manager/cache.py in get_by_hash(self, sha_hash)
    127             SHA-1 hash of the file.
    128         """
--> 129         details = self._storage.find_by_key('file_hash', sha_hash)
    130         return details
    131 

~/.virtualenvs/sunpy-release/lib/python3.8/site-packages/sunpy/data/data_manager/storage.py in find_by_key(self, key, value)
    161         with self.connection() as conn:
    162             cursor = conn.cursor()
--> 163             cursor.execute(f'''SELECT * FROM {self._table_name}
    164                                       WHERE {key}="{value}"''')
    165             row = cursor.fetchone()

OperationalError: no such table: cache_storage

```
6. Restart your Python session and re-run:
```python
from sunpy.data import manager
@manager.require('test_file',
                  ['http://data.sunpy.org/sample-data/predicted-sunspot-radio-flux.txt'],
                  '4c85b04a5528aa97eb84a087450eda0421c71833820576330bba148564089b11')
def test_function():
    pass

test_function()
```
7. Observe how it's still broken.

Note, that if you delete the contents of the `data_manager` dir and then *restart* your Python session before calling the data manager again it recovers fine.

### System Details

  - SunPy Version: 1.1.1
  - Python Version: 3.8

adwaitbhope:
### Description

Fixes #3853 where `DataManager` does not recover after we delete its `.db` file while a session is running.

If we follow the steps to reproduce described in the issue, it works as expected.


nabobalis:
Thanks for the PR, I think to confirm it was fixed, we need a test that replicates the issue.

adwaitbhope:
Sure. Should I add it as another commit in this PR itself?
Or would we need to run the test on the current master branch first, to see if the test indeed fails?

nabobalis:
Yes please can you add it to this PR.

sunpy/data/data_manager/tests/test_manager.py
```diff
@@ -144,3 +145,26 @@ def test_file_changed(data_function, storage):
     # Now it should error
     with pytest.warns(SunpyUserWarning):
         data_function()
+
+
+def test_delete_db(manager, sqlstorage):
```
nabobalis, line 150: Have you ran this test without the fix you added to make sure it fails?

adwaitbhope: I had not, and turns out it passed. I've updated the test now - it fails without the bug fix, and passes with it.

nabobalis:
Needs a bugfix changelog and the os import moved to the top and that should solve the remaining two checks.

sunpy/data/data_manager/tests/conftest.py
```diff
@@ -55,6 +55,16 @@ def manager(tmp_path, downloader, storage, mocker):
     yield manager
 
 
+@pytest.fixture
+def sqlmanager(tmp_path, downloader, sqlstorage, mocker):
```
nabobalis, line 59: What is the difference between this and the original fixture?

nabobalis: Ah, I see, it uses the sqlstorage instead of the inmem one.

sunpy/data/data_manager/storage.py
```diff
@@ -122,16 +122,12 @@ def __init__(self, path):
             self._setup()
 
     def _setup(self):
-        schema = ' text, '.join(self.COLUMN_NAMES) + ' text'
         with self.connection(commit=True) as conn:
-            # Do this in a try...except to prevent race conditions in the tests
```
nabobalis, line 127: I wonder if we need to worry about this anymore in our test suite.
I assume that "IF NOT EXISTS" might not avoid this problem?

sunpy/data/data_manager/tests/conftest.py
adwaitbhope: This one uses an `sqlstorage` instance for the `cache`, instead of `storage` which was an instance of `InMemStorage`. Since this bug pertained to deleting the SQLite db file, I needed a `manager` that uses `sqlstorage`.

sunpy/data/data_manager/storage.py
adwaitbhope: I'm sorry, did you mean "might avoid this problem"?

IF NOT EXISTS should take care of the exception that's being handled, since it pertains to creating a table that already existed.

nabobalis: Yeah, sorry, I did mean "might avoid this problem".

changelog/5089.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed bug where `DataManager` used to break upon deleting `data_manager.db`.
```
nabobalis, line 1:
```suggestion
Fixed bug where `sunpy.data.data_manager.storage.DataManager` would fail to recover upon deleting the sqlite database file.
```

nabobalis (approved):

Cadair (approved):
Looks good, thanks a lot.

nabobalis:
Thanks for the PR @adwaitbhope

PR #5091, opened 2021-03-14
Re-ordered keyword-only arguments of draw_rectangle

### LINKED ISSUE #4025: Re-order and remove the deprecation on the submap and draw_rectangle parameters after the 2.0 release.
### Description

As [mentioned here](https://github.com/sunpy/sunpy/pull/3677#discussion_r409606416), the new API for specifying rectangular coordinates requires keyword-only arguments.
The current implementation has a decorator for deprecating the old API which will be supported until the official 2.0 release.
This is the issue to keep track of the eventual removal of the deprecation and the reordering of the parameters to submap and draw_rectangle.

samriddhi-dev:
@Cadair as mentioned in the issue summary, I have re-ordered the positional arguments of draw_rectangle. Though, it is to be noted the modification of the decorator and re-ordering of the sub_map was already done. As mentioned  https://github.com/sunpy/sunpy/pull/3677#discussion_r409606416 

### Description

Fixes #4025

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -2055,8 +2055,8 @@ class GenericMap(NDData):
         return quad
 
     @u.quantity_input
-    def draw_rectangle(self, bottom_left, *, width: u.deg = None, height: u.deg = None,
-                       axes=None, top_right=None, **kwargs):
+    def draw_rectangle(self, bottom_left, *,top_right=None, width: u.deg = None, height: u.deg = None,
+                       axes=None,  **kwargs):
         """
         Draw a rectangle defined in world coordinates on the plot.
 
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -2055,8 +2055,8 @@ class GenericMap(NDData):
         return quad
 
     @u.quantity_input
-    def draw_rectangle(self, bottom_left, *,top_right=None, width: u.deg = None, height: u.deg = None,
-                       axes=None,  **kwargs):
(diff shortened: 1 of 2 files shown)


nabobalis:
I think this will need a breaking changelog entry.

sunpy/map/mapbase.py
```diff
@@ -2048,8 +2048,8 @@ def draw_quadrangle(self, bottom_left, *, width: u.deg = None, height: u.deg = N
         return quad
 
     @u.quantity_input
-    def draw_rectangle(self, bottom_left, *, width: u.deg = None, height: u.deg = None,
-                       axes=None, top_right=None, **kwargs):
+    def draw_rectangle(self, bottom_left, *,top_right=None, width: u.deg = None, height: u.deg = None,
```
nabobalis, line 2051:
```suggestion
    def draw_rectangle(self, bottom_left, *, top_right=None, width: u.deg = None, height: u.deg = None,
```

ayshih:
I'll point out that you've re-ordered the keyword-only arguments, not positional arguments.  The changelog entry should be "trivial".

nabobalis:
> I'll point out that you've re-ordered the keyword-only arguments, not positional arguments. The changelog entry should be "trivial".

Right, I forgot about the * operator. Otherwise it would be breaking. My bad.

sunpy/map/mapbase.py
nabobalis: If we can get this change in and a changelog, this PR is ready to go.

nabobalis (approved):

nabobalis:
Thanks for the PR @samriddhi-dev

wtbarnes (approved):

PR #5100, opened 2021-03-16
update online db test

nabobalis:
Fixes the online test failure for me locally.

### DIFF
--- a/sunpy/database/tests/test_database.py
+++ b/sunpy/database/tests/test_database.py
@@ -1046,7 +1046,7 @@ def test_fetch_separate_filenames():
 def test_disable_undo(database, download_query, tmpdir):
     entry = DatabaseEntry()
     with disable_undo(database) as db:
-        db.set_cache_size(5)
+        db.set_cache_size(10)
         db.add(entry)
         db.commit()
         db.remove(entry)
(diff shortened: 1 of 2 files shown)


sunpy/database/tests/test_database.py
```diff
@@ -1046,7 +1046,7 @@ def test_fetch_separate_filenames():
 def test_disable_undo(database, download_query, tmpdir):
     entry = DatabaseEntry()
     with disable_undo(database) as db:
-        db.set_cache_size(5)
+        db.set_cache_size(10)
```
Cadair, line 1049: For why?

Cadair (approved):

sunpy/database/tests/test_database.py
nabobalis: So the test was originally erroring on the keyerror not being raised instead of a sqlalchemy one. 
Fixing that led me to trigger another issue with the cache removal in the remove function.

It seems the cache we had at 5 now, got keys that went from 5-9 instead. So when it went to pop up the 4th entry, it errored. Increasing the cache to 10 seems to have fixed it. But this seems like a bug in the original code.

Issue #5107, opened 2021-03-17
Improve performance of VSOClient


dstansby:
I have just tried searching for a day's worth of WISPR data using the following code:
```python

from sunpy.net import Fido
from sunpy.net import attrs as a

result = Fido.search(a.Time('2020/01/28 00:00:00', '2020/01/29 00:00:00'),
                     a.Instrument('WISPR'))
print(result)
```
This takes ~1 minute to execute, and it turns out the majority of the time (~42s) is spent converting time strings in astropy! There must be a way to speed this up...

EDIT: original bug report was from sunpy 2.0.x, I've updated the trace below for 2.1.x. It seems a bit better, but there's still a lot of time taken up converting time strings.

```python
39.451 <module>  downloading_hmi.py:1
├─ 35.274 search  sunpy/net/fido_factory.py:301
│     [696 frames hidden]  sunpy, zeep, requests, urllib3, http,...
│        9.724 readinto  socket.py:655
│        └─ 9.724 socket.recv_into  ../../<built-in>:0
│        5.958 readinto  socket.py:655
│        └─ 5.946 socket.recv_into  ../../<built-in>:0
│        12.494 convert_time_str  sunpy/time/time.py:212
│        ├─ 4.469 __add__  astropy/time/core.py:2079
│        │  ├─ 3.008 _set_scale  astropy/time/core.py:536
│        │  │  ├─ 0.775 jd2_filled  astropy/time/formats.py:238
│        │  │  │  └─ 0.754 masked  astropy/time/formats.py:232
│        │  │  ├─ 0.517 day_frac  astropy/time/utils.py:18
│        │  │  ├─ 0.425 [self]  
│        │  │  ├─ 0.404 __init__  astropy/time/formats.py:136
│        │  │  └─ 0.401 utctai  erfa/core.py:17050
│        │  │        [4 frames hidden]  erfa
│        │  └─ 0.882 replicate  astropy/time/core.py:987
│        │     └─ 0.864 _apply  astropy/time/core.py:1018
│        │        └─ 0.442 __init__  astropy/time/formats.py:136
│        ├─ 4.454 strptime  astropy/time/core.py:1621
│        │  ├─ 3.507 __init__  astropy/time/core.py:1515
│        │  │  └─ 3.384 _init_from_vals  astropy/time/core.py:343
│        │  │     └─ 3.291 _get_time_fmt  astropy/time/core.py:400
│        │  │        └─ 3.255 __init__  astropy/time/formats.py:136
[...]

PR #5108, opened 2021-03-17
Shortcut adding zero timedelta in parse_time()

dstansby:
xref https://github.com/sunpy/sunpy/issues/5107 - there is a non-trival performance overhead in adding a zero timedelta, when instead one can just check for zero and not add the timedelta. This should possibly go upstream to astropy too?

### DIFF
--- a/changelog/5108.trivial.rst
+++ b/changelog/5108.trivial.rst
@@ -1,0 +1,3 @@
+Significantly sped up calls to :func:`~sunpy.time.parse_time` for string
+arguments. This will have knock on effects, including improved performance of
+querying the VSO.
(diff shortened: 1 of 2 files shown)


ayshih:
I recommend a complete change in the fix approach.  `time_delta` can only have two values: `TimeDelta(0*u.day)` and `TimeDelta(1*u.day)`.  I suggest getting rid of `time_delta` entirely by changing the private function `_regex_parse_time()` to return a Boolean instead for its second argument, and that Boolean would drive if a day needs to be added in `convert_time_str()`.

dstansby:
Good idea, thanks for digging a bit deeper; modified as such.

ayshih (approved):

Cadair (approved):

Cadair:
I think `parse_time` might be a great place to start writing some benchmarks?

PR #5110, opened 2021-03-17
pin oldest deps

nabobalis:
Lets see what I broke.

So I can say that on py3.9, I can not get the oldest deps to work, scipy needs numpy pre-installed and h5py also refuses to compile. There are wheels for older versions of python, so it is ok for those.


setup.cfg
```diff
@@ -40,55 +41,55 @@ install_requires =
...
-  matplotlib>=3.1.1
+  matplotlib>=3.1.0
 tests =
-  pytest-astropy >= 0.8  # 0.8 is the first release to include filter-subpackage
-  pytest-doctestplus >= 0.5 # We require the newest version of doctest plus to use +IGNORE_WARNINGS
+  jplephem  # For some coordinates tests
+  pytest-astropy
+  pytest-doctestplus
   pytest-mock
-  pytest-mpl >= 0.12
```
Cadair, line 77: why?

nabobalis: I can add them back? I don't think it really matters for the test deps.

Cadair: Well if we rely on a feature in a particular release we should tell pip about that when people are installing so it auto upgrades things if they are already installed.

Cadair (approved):

PR #5111, opened 2021-03-18
Namespace downloaded files with module name

### LINKED ISSUE #3908: Should remote data manager namespace file names?
While working on sunpy/sunxspex#34 I thought, "what would happen if two affiliated packages both used the same name for a data file?" So I was wondering if we could do some mad introspection stuff to automatically namespace variables so it would be `"sunxspex.xray_abundancies"`?

adwaitbhope:
### Description

Added namespaces for files downloaded using `DataManager`, by prepending the file name with module name. 

Closes #3908.

### DIFF
--- a/sunpy/data/data_manager/cache.py
+++ b/sunpy/data/data_manager/cache.py
@@ -38,7 +38,7 @@ class Cache:
         self._cache_dir = Path(cache_dir)
         self._expiry = expiry if expiry is None else TimeDelta(expiry)
 
-    def download(self, urls, namespace='sunpy.', redownload=False):
+    def download(self, urls, namespace='', redownload=False):
         """
         Downloads the files from the urls.
 
@@ -140,7 +140,7 @@ class Cache:
         details = self._storage.find_by_key('url', url)
         return details
 
-    def _download_and_hash(self, urls, namespace='sunpy.'):
+    def _download_and_hash(self, urls, namespace=''):
         """
         Downloads the file and returns the path, hash and url it used to download.
 
--- a/sunpy/data/data_manager/cache.py
+++ b/sunpy/data/data_manager/cache.py
@@ -38,7 +38,7 @@ class Cache:
         self._cache_dir = Path(cache_dir)
         self._expiry = expiry if expiry is None else TimeDelta(expiry)
 
-    def download(self, urls, namespace, redownload=False):
+    def download(self, urls, namespace='sunpy.', redownload=False):
         """
         Downloads the files from the urls.
 
@@ -140,7 +140,7 @@ class Cache:
         details = self._storage.find_by_key('url', url)
         return details
 
-    def _download_and_hash(self, urls, namespace):
+    def _download_and_hash(self, urls, namespace='sunpy.'):
         """
         Downloads the file and returns the path, hash and url it used to download.
 
--- a/sunpy/data/data_manager/cache.py
+++ b/sunpy/data/data_manager/cache.py
@@ -38,7 +38,7 @@ class Cache:
         self._cache_dir = Path(cache_dir)
         self._expiry = expiry if expiry is None else TimeDelta(expiry)
 
-    def download(self, urls, redownload=False):
+    def download(self, urls, namespace, redownload=False):
         """
         Downloads the files from the urls.
 
@@ -83,7 +83,7 @@ class Cache:
             else:
                 return Path(details['file_path'])
 
-        file_path, file_hash, url = self._download_and_hash(urls)
+        file_path, file_hash, url = self._download_and_hash(urls, namespace)
 
         self._storage.store({
             'file_hash': file_hash,
@@ -140,7 +140,7 @@ class Cache:
         details = self._storage.find_by_key('url', url)
         return details
 
-    def _download_and_hash(self, urls):
+    def _download_and_hash(self, urls, namespace):
         """
         Downloads the file and returns the path, hash and url it used to download.
 
@@ -155,7 +155,7 @@ class Cache:
             Path, hash and URL of the file.
         """
         def download(url):
-            path = self._cache_dir / get_filename(urlopen(url), url)
+            path = self._cache_dir / (namespace + get_filename(urlopen(url), url))
             # replacement_filename returns a string and we want a Path object
             path = Path(replacement_filename(str(path)))
             self._downloader.download(url, path)
--- a/sunpy/data/data_manager/manager.py
+++ b/sunpy/data/data_manager/manager.py
@@ -64,7 +64,8 @@ class DataManager:
                         file_hash = hash_file(file_path)
                     else:
                         file_path, file_hash, _ = self._cache._download_and_hash(
-                                                    [replace['uri']], self._namespace)
+                            [replace['uri']], self._namespace
+                        )
                     if replace['hash'] and file_hash != replace['hash']:
                         # if hash provided to replace function doesn't match the hash of the file
                         # raise error
--- a/sunpy/data/data_manager/manager.py
+++ b/sunpy/data/data_manager/manager.py
@@ -1,3 +1,4 @@
+import inspect
 import pathlib
 import warnings
 import functools
@@ -26,6 +27,7 @@ class DataManager:
 
         self._file_cache = {}
 
+        self._namespace = None
         self._skip_hash_check = False
         self._skip_file: Dict[str, str] = {}
 
@@ -49,6 +51,12 @@ class DataManager:
         def decorator(func):
             @functools.wraps(func)
             def wrapper(*args, **kwargs):
+                self._namespace = func.__module__.lstrip('.').split('.')[0]
+                if self._namespace == '__main__':
+                    self._namespace = ''
+                else:
+                    self._namespace += '.'
+
                 replace = self._skip_file.get(name, None)
                 if replace:
                     uri_parse = urlparse(replace['uri'])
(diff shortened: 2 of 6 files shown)


adwaitbhope:
I've updated the existing tests, but how should I go about adding new tests for `manager` that simulate a different module name?

nabobalis:
Create a fake external module to be used in a test?

adwaitbhope:
So something like a `fake_module` directory in the root of the repo or maybe under the `tools` directory?

nabobalis:
> So something like a `fake_module` directory in the root of the repo or maybe under the `tools` directory?

In the test folder?

adwaitbhope:
Well, if it's inside the main `sunpy` folder, the module name will be `sunpy`. I wanted to have a test that simulates a different module name.

sunpy/data/data_manager/cache.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, downloader, storage, cache_dir, expiry=10*u.day):
         self._cache_dir = Path(cache_dir)
         self._expiry = expiry if expiry is None else TimeDelta(expiry)
 
-    def download(self, urls, redownload=False):
+    def download(self, urls, namespace='sunpy.', redownload=False):
```
nabobalis, line 41: I think if we can avoid adding the "." to the namespace, that would be nice.

wtbarnes:
I've just installed this and run the aiapy tests against it and everything seems to work as expected:

image

sunpy/data/data_manager/cache.py
wtbarnes: Why does this have to be hardcoded as `sunpy`? Isn't this already sorted out by `_get_module`?

adwaitbhope: I added this hardcoded default because a few places in sunpy's codebase were using `cache.download()` directly, instead of going through `manager.require()`, which is what provides `_get_module`.

Cadair: I am not convinced we need to do this namespacing for direct invocations of `cache.download()`. The cache is designed to be able to be used by user code or by libraries, whereas the manager is really a library interface.

adwaitbhope: I see. How about we pass an empty string as the default argument?

Cadair: `None` would be marginally cleaner than `""` although either works.

adwaitbhope: I'll go with `''` because I'll have to introduce a check with `None`, whereas concatenation will happen with `''` in a concise way.

adwaitbhope:
Since we were talking about how this PR would affect `manager.override_file()` on the community call last week, I wrote a [test](https://github.com/sunpy/sunpy/pull/5111/commits/837e201bbc56a6c9a31bb663d444b4f9eaafeacc#diff-f17a536f701c02b6143960385de228a2b105e65dabe205f0df9b9697069053edR185) for it. Can someone confirm if the test addresses our concern in the first place and if that is indeed the expected behavior?

sunpy/data/data_manager/manager.py
```diff
@@ -172,10 +181,17 @@ def get(self, name):
...
+        return pathlib.Path(self._file_cache[name][self._namespace])
 
     def _cache_has_file(self, urls):
         for url in urls:
             if self._cache._get_by_url(url):
                 return True
         return False
+
+    def _get_module(self, func):
+        # Made into a separate function for the ability to patch in tests
```
nabobalis, line 193: I think this should get a docstring

sunpy/data/data_manager/tests/test_manager.py
```diff
@@ -160,3 +160,65 @@ def test_function():
...
+    assert len(storage._store) == 2
+    assert downloader.times_called == 2
+
+    # Check if the files are namespaced correctly
+    assert Path(storage._store[0]['file_path']).name == 'sunpy.test_file'
+    assert Path(storage._store[1]['file_path']).name == 'fake_module.test_file'
+
+
+def test_namespacing_with_manager_override_file(
+        module_patched_manager, downloader, storage, data_function_from_fake_module):
```
nabobalis, line 186: This is line broken strangely.

sunpy/data/data_manager/tests/test_manager.py
```diff
@@ -160,3 +160,65 @@ def test_function():
 
     # SQLite should not throw an error
     test_function()
+
+
+def test_same_file_id_different_module(
+        downloader, storage, data_function, data_function_from_fake_module):
```
nabobalis, line 166: This is line broken strangely.

sunpy/data/data_manager/tests/test_manager.py
adwaitbhope: Would this be better?
```python3
def test_namespacing_with_manager_override_file(module_patched_manager,
        downloader, storage, data_function_from_fake_module):
```

sunpy/data/data_manager/manager.py
adwaitbhope: [Added](https://github.com/adwaitbhope/sunpy/blob/cefba371a5e761afe1a3b87ce32975590c4c9c94/sunpy/data/data_manager/manager.py#L192)

sunpy/data/data_manager/tests/test_manager.py
nabobalis: I guess.

sunpy/data/data_manager/manager.py
```diff
@@ -172,10 +181,30 @@ def get(self, name):
...
+        func : `function`
+            A function who's module is to be found.
+
+        Returns
+        -------
+        `module`:
+            A `str` that represents the module name appended with a dot.
+        """
+        module = func.__module__.lstrip('.').split('.')[0] + '.'
+        if module == '__main__.':
```
nabobalis, line 208: Is there a test for this case?

sunpy/data/data_manager/manager.py
```diff
@@ -172,10 +181,30 @@ def get(self, name):
...
+
+    def _get_module(self, func):
+        """
+        Returns the name of the module (appended with a dot) that a function belongs to.
+        This is made into a separate function for the ability to patch in tests.
+
+        Parameters
+        ----------
+        func : `function`
+            A function who's module is to be found.
```
nabobalis, line 200:
```suggestion
        func : function
            A function whose module is to be found.
```

sunpy/data/data_manager/manager.py
```diff
@@ -172,10 +181,30 @@ def get(self, name):
...
+
+        Returns
+        -------
+        `module`:
+            A `str` that represents the module name appended with a dot.
```
nabobalis, line 205:
```suggestion
    A `str` that represents the module name appended with a dot.
```

sunpy/data/data_manager/manager.py
```diff
@@ -172,10 +181,30 @@ def get(self, name):
...
     def _cache_has_file(self, urls):
         for url in urls:
             if self._cache._get_by_url(url):
                 return True
         return False
+
+    def _get_module(self, func):
+        """
+        Returns the name of the module (appended with a dot) that a function belongs to.
+        This is made into a separate function for the ability to patch in tests.
```
nabobalis, line 195:
```suggestion
        Returns the name of the module (appended with a dot) that a function belongs to.
```

sunpy/data/data_manager/manager.py
```diff
@@ -96,16 +100,21 @@ def wrapper(*args, **kwargs):
...
                                                    "The remote file on the server might have changed.")
                         else:
                             file_path = details['file_path']
 
-                self._file_cache[name] = file_path
-                return func(*args, **kwargs)
+                if name not in self._file_cache:
+                    self._file_cache[name] = {}
+                self._file_cache[name][self._namespace] = file_path
+
```
nabobalis, line 114:
```suggestion
```

sunpy/data/data_manager/manager.py
adwaitbhope: No, writing such a test is similar to the problem I faced before, where I was not able to fake the module name of a function for the tests. I dabbled in `sys.module` but with no luck. I think this has got something to do with `pytest`, but I'm not sure. I ended up patching the `_get_module` function instead.

Do you have any suggestions on how we could do this?

sunpy/data/data_manager/manager.py
adwaitbhope: Oops

sunpy/data/data_manager/manager.py
nabobalis: Nope.

sunpy/data/data_manager/manager.py
adwaitbhope: For some reason the "Commit suggestion" button is grayed out. Should I add make the changes and commit manually?

nabobalis: Sure.

sunpy/data/data_manager/tests/test_manager.py
adwaitbhope: I modified it a bit since it was failing the `flake8` check.

Cadair (approved):
As long as @wtbarnes with his aiapy hat on is ok with this, let's get it in 3.0.

nabobalis (approved):

nabobalis:
Thanks for the PR @adwaitbhope

PR #5121, opened 2021-03-20
new gallery example for selecting pixels in a map based on a coordinate transform

### LINKED ISSUE #5095: Add an example to the gallery of selecting pixels in a map based on a coordinate transform
Extracting the pixel values of a region of interest defined by a box in some world coordinate space is quite a common operation. A simple example would be extracting a region of interest in HGS coordinates:

```python
import numpy as np
import astropy.units as u

import sunpy.map
from sunpy.data.sample import AIA_171_ROLL_IMAGE

smap = sunpy.map.Map(AIA_171_ROLL_IMAGE)

all_hpc = sunpy.map.all_coordinates_from_map(smap)
all_hgs = all_hpc.transform_to("heliographic_stonyhurst")

segment = np.logical_and(all_hgs.lon<10*u.deg, all_hgs.lon>0*u.deg).nonzero()
smap.data[segment] = 0
smap.peek()
```

This can be extended to more exotic transformations by using `NorthOffsetFrame` (for example the second half of [this](https://gist.github.com/Cadair/494b1d7026918cd97c57c21def52fc31) old notebook) or the newer `RotatedSunFrame`.

A good example would show off a few of these options and also present a couple of things to do with the extracted mask, for example, plotting the mask, averaging the pixel values in the mask etc.

jeffreypaul15:
### Description

Fixes #5095
Added new gallery example for selecting pixels based on coordinate transform.
The pixels are segmented by converting to 3 different frames and all of them are displayed although I'm not too sure if the newly created maps of the segmented data have the correct meta information.


examples/map/map_segment.py
```diff
@@ -0,0 +1,118 @@
...
+
+###############################################################################
+# A utility function gives us access to the helioprojective coordinate of each
+# pixel. From this we then transform the coordinates to the required frame.
+all_hpc = sunpy.map.all_coordinates_from_map(smap)
+all_hgs = all_hpc.transform_to("heliographic_stonyhurst")
+
+###############################################################################
+# Let's then segment the data based on coordinates and convert them to pixel
+# index. We convert the segmented pixel values to 0.
```
Cadair, line 33: I think this comment needs to explain a little more about what this line is doing. It's creating a boolean mask where True is the masked out segment (I think).

examples/map/map_segment.py
```diff
@@ -0,0 +1,118 @@
...
+segment = np.logical_and(all_hgs.lon < 10 * u.deg, all_hgs.lon > 0 * u.deg)
+
+###############################################################################
+# Let's plot the segment separately
+new_data = np.asarray(segment, dtype='float')
+new_data = np.where(new_data == False, np.nan, new_data)
+
+new_frame_map = sunpy.map.Map(new_data * smap.data, smap.meta)
```
Cadair, line 41: If you want to mask out the other bits of the map, I think setting `map.mask` would be a better pattern.

I think this would work:
```suggestion
smap.mask = segment
smap.plot()
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,118 @@
...
+new_data = np.asarray(segment, dtype='float')
+new_data[new_data == False] = np.nan
+
+ofsetted_map = sunpy.map.Map(new_data * smap.data, smap.meta)
+fig = plt.figure()
+fig.add_subplot(projection=smap)
+ofsetted_map.plot()
+plt.show()
+
+###############################################################################
```
Cadair, line 73: Now I see this, I wonder if it's worth having this in here. I wonder if a section on doing more mathematical operations with the mask, like sum / mean / etc instead.

Cadair (commented):
This is looking great, thanks for working on it. I think that refocusing this on generating and working with masks would make it a little more useful.

examples/map/map_segment.py
jeffreypaul15: This change would basically remove the segment from `smap`, I was thinking of plotting the segment separately.

Cadair: You mean it would mask out the segment rather than masking out the rest of the image?

jeffreypaul15: yeah, wouldn't that be more appropriate for the example?

Cadair: yeah that makes sense, in that case run the mask through `np.logical_not` first to reverse it.

examples/map/map_segment.py
jeffreypaul15: Yeah, I shall word it properly

examples/map/map_segment.py
jeffreypaul15: I have added this

examples/map/map_segment.py
jeffreypaul15: Should I do that instead? Like add another part to this example performing mathematical operations.
Do we keep the `NorthOffsetFrame` example though?

examples/map/map_segment.py
```diff
@@ -0,0 +1,119 @@
...
+import numpy as np
+from reproject import reproject_interp
+
+import astropy.units as u
+from astropy.coordinates import SkyCoord
+from astropy.wcs import WCS
+
+import sunpy.map
+from sunpy.coordinates import Helioprojective, NorthOffsetFrame, RotatedSunFrame, transform_with_sun_center
+from sunpy.data.sample import AIA_171_ROLL_IMAGE
```
Cadair, line 19: You shouldn't use roll image (just use `AIA_171_IMAGE`) because your example is killing RTD because of memory usage. The roll image is a full size 4k x 4k image and the others are downsampled.

examples/map/map_segment.py
Cadair: I think maybe just drop the `RotatedSunFrame` stuff and add some examples of using the mask in operations to the end of both the other sections.

examples/map/map_segment.py
jeffreypaul15: Ah, I see. I shall fix this..

examples/map/map_segment.py
```diff
@@ -0,0 +1,84 @@
...
+fig = plt.figure()
+fig.add_subplot(projection=smap)
+ofsetted_map.plot()
+plt.show()
+
+###############################################################################
+# We can also find the maximum, minimum or average pixel values of the segment
+mean_pix = np.average(segment)
+max_pix = np.max(segment*smap.data)
+min_pix = np.min(segment*smap.data)
```
Cadair, line 84: All three of these need to be using the original map data? Also would using a numpy masked array be better here?
By which I mean something along the lines of:
```
np.average(np.ma.array(ofsetted_map,data, mask=ofsetted_map.mask))
```

Finally, printing these values out and comparing them to the original `smap` would be nice, just to show it doing something.

examples/map/map_segment.py
```diff
@@ -0,0 +1,84 @@
...
+    ofsetted_coords.lon > -20 * u.deg)
+
+###############################################################################
+# Let's plot the ofsetted segment separately
+ofsetted_map = sunpy.map.Map(
+    smap.data,
+    smap.meta,
+    mask=np.logical_not(segment))
+fig = plt.figure()
+fig.add_subplot(projection=smap)
```
Cadair, line 76: I would replace this with:
```suggestion
fig = plt.figure()
ax = plt.subplot(projection=smap)
```

examples/map/map_segment.py
jeffreypaul15: Oh yeah, I forgot about printing it :facepalm:

jeffreypaul15: The mask should be in a `logical_not` though right? Otherwise `array.mean()` would produce the mean of the entire map without the segment.

examples/map/map_segment.py
```diff
@@ -0,0 +1,88 @@
...
+plt.show()
+
+###############################################################################
+# We can perform various mathematical operations on the extracted segment such
+# as averaging the pixel values or finding the sum of the segment.
+new_frame_mask = np.ma.array(smap.data, mask=np.logical_not(segment))
+
+print(f'Original Map : mean = {smap.data.mean()}, sum = {smap.data.sum()}')
+print(
+    f'Segmented Map : mean = {new_frame_mask.mean()}, sum = {new_frame_mask.sum()}')
```
Cadair, line 53:
```suggestion
print(f"Segmented Map : mean = {new_frame_mask.mean()}, sum = {new_frame_mask.sum()}")
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,88 @@
...
+plt.show()
+
+###############################################################################
+# We can also find the maximum, minimum or average pixel values of the segment
+# and compare it with the original map.
+offset_mask = np.ma.array(smap.data, mask=np.logical_not(segment))
+print(
+    f'Original Map : mean = {smap.data.mean()}, maximum value = {smap.data.max()}, minimum value = {smap.data.min()}')
+print(
+    f'Segmented Map : mean = {offset_mask.mean()}, maximum value = {offset_mask.max()}, minimum value = {offset_mask.min()}')
```
Cadair, line 88:
```suggestion
print(f"Original Map : mean = {smap.data.mean()}, "
      f"maximum value = {smap.data.max()}, "
      f"minimum value = {smap.data.min()}")
print(f"Segmented Map : mean = {offset_mask.mean()}, "
      f"maximum value = {offset_mask.max()}, "
      f"minimum value = {offset_mask.min()}")
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,88 @@
...
+# Let's plot the segment separately, we create a new map with the segment as the
+# mask.
+new_frame_map = sunpy.map.Map(
+    smap.data,
+    smap.meta,
+    mask=np.logical_not(segment))
+fig = plt.figure()
+ax = plt.subplot(projection=smap)
+new_frame_map.plot()
+plt.show()
```
Cadair, line 44:
```suggestion
new_frame_map.draw_grid()
plt.show()
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,88 @@
...
+
+###############################################################################
+# We can perform various mathematical operations on the extracted segment such
+# as averaging the pixel values or finding the sum of the segment.
+new_frame_mask = np.ma.array(smap.data, mask=np.logical_not(segment))
+
+print(f'Original Map : mean = {smap.data.mean()}, sum = {smap.data.sum()}')
```
Cadair, line 51:
```suggestion
print(f"Original Map : mean = {smap.data.mean()}, sum = {smap.data.sum()}")
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,88 @@
...
+
+###############################################################################
+# Let's plot the ofsetted segment separately
+offsetted_map = sunpy.map.Map(
+    smap.data,
+    smap.meta,
+    mask=np.logical_not(segment))
+fig = plt.figure()
+ax = plt.subplot(projection=smap)
+offsetted_map.plot()
```
Cadair, line 78: I think this plot needs a grid or two so that people can see why the segment we are showing here has been selected. Perhaps try showing both the "normal" HGS grid and the offset grid?

examples/map/map_segment.py
```diff
@@ -0,0 +1,88 @@
+"""
+=======================================================
+Segmenting a Map based on transformation of coordinates
+=======================================================
+
+A demostration creating a segment of a particular map from
+transformed coordinates.
```
Cadair, line 7:
```suggestion
A demonstration extracting a region of a particular map based on
world coordinates in different systems.
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,88 @@
...
+from sunpy.coordinates import NorthOffsetFrame
+from sunpy.data.sample import AIA_171_IMAGE
+
+###############################################################################
+# We start with the sample data.
+smap = sunpy.map.Map(AIA_171_IMAGE)
+
+###############################################################################
+# A utility function gives us access to the helioprojective coordinate of each
+# pixel. From this we then transform the coordinates to the required frame.
```
Cadair, line 25:
```suggestion
# A utility function gives us access to the helioprojective coordinate of each
# pixel. From this we then transform the coordinates to the required frame.
# For this example we are going to extract a region based on the 
# heliographic Stonyhurst coordinates, so we transform to that frame.
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,88 @@
...
+all_hpc = sunpy.map.all_coordinates_from_map(smap)
+all_hgs = all_hpc.transform_to("heliographic_stonyhurst")
+
+###############################################################################
+# Let's then segment the data based on coordinates and create a boolean mask
+# where True indicates the segmented mask.
+segment = np.logical_and(all_hgs.lon < 10 * u.deg, all_hgs.lon > 0 * u.deg)
```
Cadair, line 32: I think we should start off doing the segmentation using the numpy convention of `True` being invalid data, rather than applying `np.logical_not` later on. So I would reverse the logic here.
```suggestion
###############################################################################
# Let's then segment the data based on coordinates and create a boolean mask
# where True indicates invalid or deselected data.
# This is the convention used by the `numpy.ma` module.
segment = np.logical_and(all_hgs.lon > 10 * u.deg, all_hgs.lon < 0 * u.deg)
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,88 @@
...
+
+###############################################################################
+# Let's plot the segment separately, we create a new map with the segment as the
+# mask.
+new_frame_map = sunpy.map.Map(
+    smap.data,
+    smap.meta,
+    mask=np.logical_not(segment))
```
Cadair, line 40:
```suggestion
    mask=segment)
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,88 @@
...
+f = NorthOffsetFrame(north=north)
+
+###############################################################################
+# We then transform coordinates to the Offsetted frame and segment the data
+# based on conditions
+all_hpc = sunpy.map.all_coordinates_from_map(smap)
+ofsetted_coords = all_hpc.transform_to(f)
+segment = np.logical_and(
+    ofsetted_coords.lon < 30 * u.deg,
+    ofsetted_coords.lon > -20 * u.deg)
```
Cadair, line 68: same here, flip the logic.

examples/map/map_segment.py
```diff
@@ -0,0 +1,88 @@
...
+
+###############################################################################
+# Let's plot the ofsetted segment separately
+offsetted_map = sunpy.map.Map(
+    smap.data,
+    smap.meta,
+    mask=np.logical_not(segment))
```
Cadair, line 75:
```suggestion
    mask=segment)
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,88 @@
...
+fig = plt.figure()
+ax = plt.subplot(projection=smap)
+offsetted_map.plot()
+plt.show()
+
+###############################################################################
+# We can also find the maximum, minimum or average pixel values of the segment
+# and compare it with the original map.
+offset_mask = np.ma.array(smap.data, mask=np.logical_not(segment))
```
Cadair, line 84:
```suggestion
offset_mask = np.ma.array(smap.data, mask=segment)
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,88 @@
...
+    mask=np.logical_not(segment))
+fig = plt.figure()
+ax = plt.subplot(projection=smap)
+new_frame_map.plot()
+plt.show()
+
+###############################################################################
+# We can perform various mathematical operations on the extracted segment such
+# as averaging the pixel values or finding the sum of the segment.
+new_frame_mask = np.ma.array(smap.data, mask=np.logical_not(segment))
```
Cadair, line 49:
```suggestion
new_frame_mask = np.ma.array(smap.data, mask=segment)
```

Cadair (commented):
Thanks a lot for working on this, I think it needs a little more work to make it a little easier to understand as an introduction to masks. It is pretty likely that people reading this might not have any experience with numpy's masked arrays, or the conventions surrounding it.

A good addition to the text, as an extra to my suggestions, would be to say a little more about numpy masked arrays and link to https://numpy.org/doc/stable/reference/maskedarray.generic.html#using-numpy-ma

examples/map/map_segment.py
jeffreypaul15: https://numpy.org/doc/stable/reference/generated/numpy.ma.mean.html
As per numpy's masked arrays, the values marked `False` are considered right? So here we want only the segment and we'd have to consider everything else as True hence the `np.logical_not`.
Correct me if I'm wrong, I'm a bit confused here.

examples/map/map_segment.py
jeffreypaul15: https://numpy.org/doc/stable/reference/generated/numpy.ma.mean.html
As per numpy's masked arrays, the values marked `False` are considered right? So here we want only the segment and we'd have to consider everything else as True hence the `np.logical_not`.
Correct me if I'm wrong, I'm a bit confused here.

examples/map/map_segment.py
jeffreypaul15: Same here, have we decided to show the entire map without the segment?
image
This would be the result

examples/map/map_segment.py
jeffreypaul15: This doesn't really perform the segmentation. It produces a segment which is `False` completely.

examples/map/map_segment.py
jeffreypaul15: Not working for the above logic.

examples/map/map_segment.py
Cadair: Does it need to be logical_or when you reverse the signs? I don't have time to experiment right now, but we should really try and build the logic the right way around. I think that would make it easier to understand.

ayshih: Yes, `logical_or()`

jeffreypaul15: But based on `False` being the masked arrays accepted value, I'm not sure if we can skip the `np.logical_not`. 
`np.logical_or` doesn't provide the right plot either :sweat_smile:

jeffreypaul15: image
This works but I'm not too sure why the rest of the image isn't being as `all_hpc` only contains coordinates till 90deg? I'm not too sure about this

ayshih: You also need to mask out where `all_hgs.lon` is NaN

examples/map/map_segment.py
```diff
@@ -0,0 +1,113 @@
...
+all_hpc = sunpy.map.all_coordinates_from_map(smap)
+all_hgs = all_hpc.transform_to("heliographic_stonyhurst")
+
+######################################################################
+# Let's then segment the data based on coordinates and create a boolean mask
+# where True indicates invalid or deselected data.
+# This is the convention used by the `numpy.ma` module.
+segment = np.logical_or(
+    all_hgs.lon >= 50 * u.deg,
+    all_hgs.lon <= 0 * u.deg)
```
Cadair, line 37: We can combine all this logic into the one statement.
```suggestion
segment = np.logical_or(
    all_hgs.lon >= 50 * u.deg,
    all_hgs.lon <= 0 * u.deg,
    ~np.isnan(all_hgs.lon))
```

Alternatively `segment = np.logical_or(segment, np.logical_not(np.isnan(all_hgs.lon)))` is more explicit than the multiplication below.

jeffreypaul15: This doesn't produce the right plot though

jeffreypaul15: This still produces the issue where the `NaN` values aren't masked
image

Cadair: wait well then use `np.isnan`

ayshih: Heh, I suppose `np.inf` can't be a possibility, so sure.

jeffreypaul15: Oh yeah, forgot that existed.

examples/map/map_segment.py
```diff
@@ -0,0 +1,121 @@
...
+# Numpy's masked arrays allow for a combination of standart numpy array and
+# a mask array. When an element of the mask is True, the corresponding element
+# of the associated array is said to be masked (invalid).
+# More information about Numpy's masked arrays :
+# https://numpy.org/doc/stable/reference/maskedarray.generic.html#using-numpy-ma
+# From the segment, we now mask out all values where `all_hgs.lon` is
+# NaN.
+segment = np.logical_or(
+    segment, np.isnan(
+        all_hgs.lon))
```
Cadair, line 51:
```suggestion
segment = np.logical_or(segment, np.isnan(all_hgs.lon))
```

Cadair: This whole file is rather linebreak happy. Are you running something like black on it?

nabobalis: I am not sure that black would do that normally.

jeffreypaul15: I've been using autopep8, that was because of the `np.logical_not(np.isnan())`.
I shall make this change.

Cadair:
You seem to have a real warning in the doc build: https://dev.azure.com/sunpy/sunpy/_build/results?buildId=15270&view=logs&j=090c8010-07b3-5084-af5f-2960d012d632&t=d581a89c-705b-5160-02b7-5f27e289c243&l=287

jeffreypaul15:
> You seem to have a real warning in the doc build: https://dev.azure.com/sunpy/sunpy/_build/results?buildId=15270&view=logs&j=090c8010-07b3-5084-af5f-2960d012d632&t=d581a89c-705b-5160-02b7-5f27e289c243&l=287

Oops.

examples/map/map_segment.py
```diff
@@ -0,0 +1,103 @@
+"""
+=======================================================
+Segmenting a Map based on transformation of coordinates
+=======================================================
+
+A demonstration extracting a region of a particular map based on
+world coordinates in different systems.
```
nabobalis, line 7:
```suggestion
This example demonstrates extracting a region of a particular map based on
world coordinates in different systems.
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,103 @@
...
+# A utility function gives us access to the helioprojective coordinate of each
+# pixel. From this we then transform the coordinates to the required frame.
+# For this example we are going to extract a region based on the
+# heliographic Stonyhurst coordinates, so we transform to that frame.
+all_hpc = sunpy.map.all_coordinates_from_map(smap)
+all_hgs = all_hpc.transform_to("heliographic_stonyhurst")
+
+######################################################################
+# Let's then segment the data based on coordinates and create a boolean mask
+# where True indicates invalid or deselected data.
```
nabobalis, line 33:
```suggestion
# where `True` indicates invalid or deselected data.
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,103 @@
...
+# This is the convention used by the `numpy.ma` module.
+segment = np.logical_or(all_hgs.lon >= 50 * u.deg, all_hgs.lon <= 0 * u.deg)
+
+######################################################################
+# Let's plot the segment separately, we create a new map with the segment as the
+# mask.
+# Numpy's masked arrays allow for a combination of standart numpy array and
```
nabobalis, line 40:
```suggestion
# mask. Numpy's masked arrays allow for a combination of standard numpy array and
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,103 @@
...
+# a mask array. When an element of the mask is True, the corresponding element
```
nabobalis, line 41:
```suggestion
# a mask array. When an element of the mask is `True`, the corresponding element
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,103 @@
...
+# of the associated array is said to be masked (invalid).
+# More information about Numpy's masked arrays :
+# https://numpy.org/doc/stable/reference/maskedarray.generic.html#using-numpy-ma
```
nabobalis, line 44:
```suggestion
# `More information about numpy's masked arrays <https://numpy.org/doc/stable/reference/maskedarray.generic.html#using-numpy-ma
>`__
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,103 @@
...
+######################################################################
+# We can perform various mathematical operations on the extracted segment such
+# as averaging the pixel values or finding the sum of the segment.
+new_frame_mask = np.ma.array(smap.data, mask=segment)
+
+print(f"Original Map : mean = {smap.data.mean()}, sum = {smap.data.sum()}")
+print(f"Segmented Map : mean = {new_frame_mask.mean()}, sum = {new_frame_mask.sum()}")
+
+######################################################################
+# This example can also be extended while using NorthOffsetFrame
```
nabobalis, line 65:
```suggestion
# This example can also be extended while using `sunpy.coordinates.NorthOffsetFrame`
```
I think this is correct.

examples/map/map_segment.py
```diff
@@ -0,0 +1,103 @@
...
+# Let us offset the north pole and create the frame
```
nabobalis, line 65:
```suggestion
# Let us offset the north pole and create the frame.
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,103 @@
...
+north = SkyCoord(20 * u.deg, 20 * u.deg, frame="heliographic_stonyhurst")
+f = NorthOffsetFrame(north=north)
+
+######################################################################
+# We then transform coordinates to the Offsetted frame and segment the data
+# based on conditions
```
nabobalis, line 71:
```suggestion
# based on conditions.
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,103 @@
...
+print(f"Segmented Map : mean = {new_frame_mask.mean()}, sum = {new_frame_mask.sum()}")
+
+######################################################################
+# This example can also be extended while using NorthOffsetFrame
+# Let us offset the north pole and create the frame
+north = SkyCoord(20 * u.deg, 20 * u.deg, frame="heliographic_stonyhurst")
+f = NorthOffsetFrame(north=north)
+
+######################################################################
+# We then transform coordinates to the Offsetted frame and segment the data
```
nabobalis, line 70:
```suggestion
# We then transform coordinates to the offsetted frame and segment the data
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,103 @@
...
+# based on conditions
+all_hpc = sunpy.map.all_coordinates_from_map(smap)
+ofsetted_coords = all_hpc.transform_to(f)
+segment = np.logical_or(ofsetted_coords.lon >= 30 * u.deg,
+                        ofsetted_coords.lon <= -20 * u.deg)
+
+######################################################################
+# Masking out the NaN values of `ofsetted_coords.lon`, we get
```
nabobalis, line 78:
```suggestion
# Masking out the NaN values of ``ofsetted_coords.lon``, we get:
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,103 @@
...
+segment = np.logical_or(segment, np.isnan(ofsetted_coords.lon))
+
+######################################################################
+# Let's plot the ofsetted segment separately
```
nabobalis, line 82:
```suggestion
# Let's plot the ofsetted segment separately.
```

nabobalis:
I would suggest adding the commits into a batch instead of doing it one by one.

jeffreypaul15:
> I would suggest adding the commits into a batch instead of doing it one by one.

Wasn't thinking of that, sorry.

examples/map/map_segment.py
```diff
@@ -0,0 +1,100 @@
...
+segment = np.logical_or(all_hgs.lon >= 50 * u.deg, all_hgs.lon <= 0 * u.deg)
+
+######################################################################
+# Let's plot the segment separately, we create a new map with the segment as the
+# mask. Numpy's masked arrays allow for a combination of standard numpy array and
+# a mask array. When an element of the mask is `True`, the corresponding element
+# of the associated array is said to be masked (invalid).
+# `More information about numpy's masked arrays <https://numpy.org/doc/stable/reference/maskedarray.generic.html#using-numpy-ma>`
+# From the segment, we now mask out all values where `all_hgs.lon` is
+# NaN.
```
nabobalis, line 43:
```suggestion
# From the segment, we now mask out all values where ``all_hgs.lon`` is NaN.
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,99 @@
...
+segment = np.logical_or(all_hgs.lon >= 50 * u.deg, all_hgs.lon <= 0 * u.deg)
+
+######################################################################
+# Let's plot the segment separately, we create a new map with the segment as the
+# mask. Numpy's masked arrays allow for a combination of standard numpy array and
+# a mask array. When an element of the mask is `True`, the corresponding element
+# of the associated array is said to be masked (invalid).
+# `More information about numpy's masked arrays <https://numpy.org/doc/stable/reference/maskedarray.generic.html#using-numpy-ma>`
+# From the segment, we now mask out all values where ``all_hgs.lon`` is NaN.
+segment = np.logical_or(segment, np.isnan(all_hgs.lon))
```
ayshih, line 43:
```suggestion
segment_mask |= np.isnan(all_hgs.lon)
```
This is cleaner, but some may not like it

ayshih, line 43: This stuff should be merged with the above block.  There's no reason to separate it out, nor to explain masked arrays twice.

examples/map/map_segment.py
```diff
@@ -0,0 +1,99 @@
...
+
+new_frame_map = sunpy.map.Map(smap.data, smap.meta, mask=segment)
```
ayshih, line 45:
```suggestion
new_frame_map = sunpy.map.Map(smap.data, smap.meta, mask=segment_mask)
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,99 @@
...
+fig = plt.figure()
+ax = plt.subplot(projection=smap)
+new_frame_map.plot()
+new_frame_map.draw_grid()
```
ayshih, line 49:
```suggestion
new_frame_map.draw_grid(color='red')
```
This way, we can see the grid lines outside of the segment too

examples/map/map_segment.py
```diff
@@ -0,0 +1,99 @@
...
+plt.show()
+
+######################################################################
+# We can perform various mathematical operations on the extracted segment such
+# as averaging the pixel values or finding the sum of the segment.
+new_frame_mask = np.ma.array(smap.data, mask=segment)
```
ayshih, line 55:
```suggestion
masked_data = np.ma.array(smap.data, mask=segment_mask)
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,99 @@
...
+
+print(f"Original Map : mean = {smap.data.mean()}, sum = {smap.data.sum()}")
+print(f"Segmented Map : mean = {new_frame_mask.mean()}, sum = {new_frame_mask.sum()}")
```
ayshih, line 58:
```suggestion
print(f"Segment : mean = {masked_data.mean()}, sum = {masked_data.sum()}")
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,99 @@
...
+# This example can also be extended while using `sunpy.coordinates.NorthOffsetFrame`
+# Let us offset the north pole and create the frame.
+north = SkyCoord(20 * u.deg, 20 * u.deg, frame="heliographic_stonyhurst")
+f = NorthOffsetFrame(north=north)
+
+######################################################################
+# We then transform coordinates to the offsetted frame and segment the data
+# based on conditions.
+all_hpc = sunpy.map.all_coordinates_from_map(smap)
+ofsetted_coords = all_hpc.transform_to(f)
```
ayshih, line 70:
```suggestion
offsetted_coords = all_hpc.transform_to(offset_frame)
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,99 @@
...
+new_frame_mask = np.ma.array(smap.data, mask=segment)
+
+print(f"Original Map : mean = {smap.data.mean()}, sum = {smap.data.sum()}")
+print(f"Segmented Map : mean = {new_frame_mask.mean()}, sum = {new_frame_mask.sum()}")
+
+######################################################################
+# This example can also be extended while using `sunpy.coordinates.NorthOffsetFrame`
+# Let us offset the north pole and create the frame.
+north = SkyCoord(20 * u.deg, 20 * u.deg, frame="heliographic_stonyhurst")
+f = NorthOffsetFrame(north=north)
```
ayshih, line 64:
```suggestion
offset_frame = NorthOffsetFrame(north=north)
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,99 @@
...
+
+######################################################################
+# We then transform coordinates to the offsetted frame and segment the data
+# based on conditions.
+all_hpc = sunpy.map.all_coordinates_from_map(smap)
+ofsetted_coords = all_hpc.transform_to(f)
+segment = np.logical_or(ofsetted_coords.lon >= 30 * u.deg,
+                        ofsetted_coords.lon <= -20 * u.deg)
```
ayshih, line 72:
```suggestion
segment_mask = np.logical_or(offsetted_coords.lon >= 30 * u.deg,
                             offsetted_coords.lon <= -20 * u.deg)
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,99 @@
...
+
+######################################################################
+# Masking out the NaN values of ``ofsetted_coords.lon``, we get:
+segment = np.logical_or(segment, np.isnan(ofsetted_coords.lon))
```
ayshih, line 76:
```suggestion
segment_mask |= np.isnan(offsetted_coords.lon)
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,99 @@
...
+######################################################################
+# We then transform coordinates to the offsetted frame and segment the data
+# based on conditions.
+all_hpc = sunpy.map.all_coordinates_from_map(smap)
+ofsetted_coords = all_hpc.transform_to(f)
+segment = np.logical_or(ofsetted_coords.lon >= 30 * u.deg,
+                        ofsetted_coords.lon <= -20 * u.deg)
+
+######################################################################
+# Masking out the NaN values of ``ofsetted_coords.lon``, we get:
```
ayshih, line 75:
```suggestion
# Masking out the NaN values of ``offsetted_coords.lon``, we get:
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,99 @@
...
+segment = np.logical_or(segment, np.isnan(ofsetted_coords.lon))
+
+######################################################################
+# Let's plot the ofsetted segment separately.
+offsetted_map = sunpy.map.Map(smap.data, smap.meta, mask=segment)
```
ayshih, line 80:
```suggestion
offsetted_map = sunpy.map.Map(smap.data, smap.meta, mask=segment_mask)
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,99 @@
...
+fig = plt.figure()
+ax = plt.subplot(projection=smap)
+offsetted_map.plot()
+overlay = ax.get_coords_overlay(f)
+overlay[0].set_ticks(spacing=30. * u.deg)
+overlay.grid(ls='--', color='blue')
+offsetted_map.draw_grid()
```
ayshih, line 87:
```suggestion
offsetted_map.draw_grid(color='red')
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,99 @@
...
+plt.show()
+
+######################################################################
+# We can also find the maximum, minimum or average pixel values of the segment
+# and compare it with the original map.
+offset_mask = np.ma.array(smap.data, mask=segment)
```
ayshih, line 93:
```suggestion
offset_masked_data = np.ma.array(smap.data, mask=segment_mask)
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,99 @@
...
+print(f"Original Map : mean = {smap.data.mean()}, "
+      f"maximum value = {smap.data.max()}, "
+      f"minimum value = {smap.data.min()}")
+print(f"Segmented Map : mean = {offset_mask.mean()}, "
+      f"maximum value = {offset_mask.max()}, "
+      f"minimum value = {offset_mask.min()}")
```
ayshih, line 99:
```suggestion
print(f"Offsetted segment : mean = {offset_masked_data.mean()}, "
      f"maximum value = {offset_masked_mask.max()}, "
      f"minimum value = {offset_masked_mask.min()}")
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,99 @@
...
+ofsetted_coords = all_hpc.transform_to(f)
+segment = np.logical_or(ofsetted_coords.lon >= 30 * u.deg,
+                        ofsetted_coords.lon <= -20 * u.deg)
+
+######################################################################
+# Masking out the NaN values of ``ofsetted_coords.lon``, we get:
+segment = np.logical_or(segment, np.isnan(ofsetted_coords.lon))
+
+######################################################################
+# Let's plot the ofsetted segment separately.
```
ayshih, line 79:
```suggestion
# Let's plot the offsetted segment separately.
```

ayshih (changes requested):

examples/map/map_segment.py
jeffreypaul15: I see..

examples/map/map_segment.py
jeffreypaul15: apparently I can't spell :facepalm:

examples/map/map_segment.py
```diff
@@ -0,0 +1,98 @@
...
+from sunpy.coordinates import NorthOffsetFrame
+from sunpy.data.sample import AIA_171_IMAGE
+
+######################################################################
+# We start with the sample data.
+smap = sunpy.map.Map(AIA_171_IMAGE)
+
+######################################################################
+# A utility function gives us access to the helioprojective coordinate of each
+# pixel. From this we then transform the coordinates to the required frame.
```
Cadair, line 24: While this function does give us the helioprojective coordinates for this map, as that is it's native coordinate system, it's not universally true. So just to be pedantic:

```suggestion
# A utility function gives us access to the helioprojective coordinate of each
# pixel in this map. From this we then transform the coordinates to the required frame.
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,98 @@
...
+# For this example we are going to extract a region based on the
+# heliographic Stonyhurst coordinates, so we transform to that frame.
+all_hpc = sunpy.map.all_coordinates_from_map(smap)
+all_hgs = all_hpc.transform_to("heliographic_stonyhurst")
+
+######################################################################
+# Let's then segment the data based on coordinates and create a boolean mask
+# where `True` indicates invalid or deselected data.
+# mask. Numpy's masked arrays allow for a combination of standard numpy array and
```
Cadair, line 33:
```suggestion
# Numpy's masked arrays allow for a combination of standard numpy array and
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,98 @@
...
+# a mask array. When an element of the mask is `True`, the corresponding element
+# of the associated array is said to be masked (invalid).
+# `More information about numpy's masked arrays <https://numpy.org/doc/stable/reference/maskedarray.generic.html#using-numpy-ma>`
```
Cadair, line 36:
```suggestion
# For more information about numpy's masked arrays see :mod:`numpy.ma`.
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,98 @@
...
+# From the segment, we now mask out all values where ``all_hgs.lon`` is NaN.
```
Cadair, line 37:
```suggestion
# We now mask out all values not in our coordinate range and where the coordinates are NaN.
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,98 @@
...
+# This is the convention used by the `numpy.ma` module.
```
Cadair, line 38:
```suggestion
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,98 @@
...
+segment_mask = np.logical_or(all_hgs.lon >= 35 * u.deg, all_hgs.lon <= -35 * u.deg)
+segment_mask |= np.isnan(all_hgs.lon)
+
+######################################################################
+# Let's plot the segment separately, we create a new map with the segment as the mask
```
Cadair, line 43:
```suggestion
# To plot the segment separately, we create a new map with the segment as the mask
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,98 @@
...
+######################################################################
+# We can perform various mathematical operations on the extracted segment such
+# as averaging the pixel values or finding the sum of the segment.
+masked_data = np.ma.array(smap.data, mask=segment_mask)
+
+print(f"Original Map : mean = {smap.data.mean()}, sum = {smap.data.sum()}")
+print(f"Segment : mean = {masked_data.mean()}, sum = {masked_data.sum()}")
+
+######################################################################
+# This example can also be extended while using `sunpy.coordinates.NorthOffsetFrame`
```
Cadair, line 60: I think this will work and create a subheading.
```suggestion
# Using `sunpy.coordinates.NorthOffsetFrame`
# ------------------------------------------
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,98 @@
...
+fig = plt.figure()
+ax = plt.subplot(projection=smap)
+new_frame_map.plot()
+new_frame_map.draw_grid(color='red')
+plt.show()
+
+######################################################################
+# We can perform various mathematical operations on the extracted segment such
+# as averaging the pixel values or finding the sum of the segment.
+masked_data = np.ma.array(smap.data, mask=segment_mask)
```
Cadair, line 54:
```suggestion
masked_data = np.ma.array(new_frame_map.data, mask=new_frame_map.mask)
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,98 @@
...
+overlay = ax.get_coords_overlay(offset_frame)
+overlay[0].set_ticks(spacing=30. * u.deg)
+overlay.grid(ls='--', color='blue')
+offsetted_map.draw_grid(color='red')
+plt.show()
+
+######################################################################
+# We can also find the maximum, minimum or average pixel values of the segment
+# and compare it with the original map.
+offset_masked_data = np.ma.array(smap.data, mask=segment_mask)
```
Cadair, line 92:
```suggestion
offset_masked_data = np.ma.array(offsetted_map.data, mask=offsetted_map.mask)
```

examples/map/map_segment.py
```diff
@@ -0,0 +1,98 @@
...
+print(f"Original Map : mean = {smap.data.mean()}, "
+      f"maximum value = {smap.data.max()}, "
+      f"minimum value = {smap.data.min()}")
+print(f"Offsetted segment : mean = {offset_masked_data.mean()}, "
```
Cadair, line 96:
```suggestion
print(f"Offset segment : mean = {offset_masked_data.mean()}, "
```

Cadair (changes requested):
With the small changes in these comments I think this is good to go. Thanks so much for working on this, it's looking really good.

examples/map/map_segment.py
nabobalis: I forgot that exists.

examples/map/map_segment.py
```diff
@@ -0,0 +1,98 @@
...
+all_hpc = sunpy.map.all_coordinates_from_map(smap)
+all_hgs = all_hpc.transform_to("heliographic_stonyhurst")
+
+######################################################################
+# Let's then segment the data based on coordinates and create a boolean mask
+# where `True` indicates invalid or deselected data.
+# Numpy's masked arrays allow for a combination of standard numpy array and
+# a mask array. When an element of the mask is `True`, the corresponding element
+# of the associated array is said to be masked (invalid).
+# For more information about numpy's masked arrays see :mod:`numpy.ma`.
```
jeffreypaul15, line 36: @nabobalis Do you have any suggestions for this part?

nabobalis: No.

Cadair (approved):

examples/map/map_segment.py
```diff
@@ -0,0 +1,98 @@
...
+# We now mask out all values not in our coordinate range and where the coordinates are NaN.
```
ayshih, line 37:
```suggestion
# We now mask out all values not in our coordinate range or where the
# coordinates are NaN (because they could not be transformed to the
# surface of the Sun).
```

ayshih (approved):
Just one fix, otherwise looks good!

nabobalis:
Thanks again @jeffreypaul15

PR #5122, opened 2021-03-20
GenericMap now passes docstrings to its subclasses

### LINKED ISSUE #3716: Automatically include Map docstring when displaying help for subclasses?
Originally raised in #396. Ideally, the `GenericMap` docstring should also be included when using one of the source-specific maps. For example, when calling `help(sunpy.map.sources.AIAMap)`, I should see the docstring for `GenericMap` and `AIAMap`.

asinghgaba:
### Description

Helps GenericMap to pass its docstring to its subclasses.
Fixes #3716

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -189,6 +189,7 @@ class GenericMap(NDData):
         `is_datasource_for` attribute.
         This is then passed into the Map Factory so we can register them.
         """
+        cls.__doc__ = cls.__doc__ + '\n' + GenericMap.__doc__.split("data array")[1].split("Examples")[0]
         super().__init_subclass__(**kwargs)
         if hasattr(cls, 'is_datasource_for'):
             cls._registry[cls] = cls.is_datasource_for
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -68,6 +68,52 @@ and `~sunpy.util.MetaDict.modified_items` properties of MetaDict
 to query how the metadata has been modified.
 """
 
+# The notes live here so we can reuse it in the source maps
+_notes_doc = """
+
+Notes
+-----
+
+A number of the properties of this class are returned as two-value named
+tuples that can either be indexed by position ([0] or [1]) or be accessed
+by the names (.x and .y) or (.axis1 and .axis2). Things that refer to pixel
+axes use the ``.x``, ``.y`` convention, where x and y refer to the FITS
+axes (x for columns y for rows). Spatial axes use ``.axis1`` and ``.axis2``
+which correspond to the first and second axes in the header. ``axis1``
+corresponds to the coordinate axis for ``x`` and ``axis2`` corresponds to
+``y``.
+
+This class makes some assumptions about the WCS information contained in
+the meta data. The first and most extensive assumption is that it is
+FITS-like WCS information as defined in the FITS WCS papers.
+
+Within this scope it also makes some other assumptions.
+
+* In the case of APIS convention headers where the CROTAi/j arguments are
+    provided it assumes that these can be converted to the standard PCi_j
+    notation using equations 32 in Thompson (2006).
+
+* If a CDi_j matrix is provided it is assumed that it can be converted to a
+    PCi_j matrix and CDELT keywords as described in
+    `Greisen & Calabretta (2002) <https://doi.org/10.1051/0004-6361:20021327>`_
+
+* The 'standard' FITS keywords that are used by this class are the PCi_j
+    matrix and CDELT, along with the other keywords specified in the WCS
+    papers. All subclasses of this class must convert their header
+    information to this formalism. The CROTA to PCi_j conversion is done in
+    this class.
+
+.. warning::
+    This class currently assumes that a header with the CDi_j matrix
+    information also includes the CDELT keywords, without these keywords
+    this class will not process the WCS.
+    Also the rotation_matrix does not work if the CDELT1 and CDELT2
+    keywords are exactly equal.
+    Also, if a file with more than two dimensions is feed into the class,
+    only the first two dimensions (NAXIS1, NAXIS2) will be loaded and the
+    rest will be discarded.
+"""
+
 __all__ = ['GenericMap']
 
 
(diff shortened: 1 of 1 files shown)


sunpy/map/mapbase.py
```diff
@@ -173,6 +173,10 @@ def __init_subclass__(cls, **kwargs):
         `is_datasource_for` attribute.
         This is then passed into the Map Factory so we can register them.
         """
+        if cls.__doc__ is None:
+            cls.__doc__ = GenericMap.__doc__
```
nabobalis, line 177: What is this for?

asinghgaba: When the subclass has no doctring.

nabobalis: Do we have a subclass without a docstring?

asinghgaba: Yes, there is one, `HMISynopticMap` in sunpy/map/sources/sdo.py

nabobalis: I think a docstring should be added to it, and these lines removed.

asinghgaba: Okay, but can you please suggest what the doctoring should be for this class? I am getting familiar with sunpy, never used before.

ayshih: These lines should be removed because:
- All of our subclasses already have their own docstrings, so there is no situation in our code where `__doc__` is `None`.
- If any subclass had no docstring, Python *already* gives it the docstring for its parent class.  This code is unnecessary because it is already done by Python.

sunpy/map/mapbase.py
```diff
@@ -173,6 +173,10 @@ def __init_subclass__(cls, **kwargs):
...
+        else:
+            cls.__doc__ = GenericMap.__doc__ + '\n' + cls.__doc__
```
ayshih, line 179:
```suggestion
            cls.__doc__ = cls.__doc__ + '\n' + GenericMap.__doc__
```

sunpy/map/mapbase.py
asinghgaba: @ayshih When I remove that condition, I get this error:
`TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'`
And this is pointing to `HMISynopoticMap` class which has no doctstring.

dstansby:
I think this is a change in behaviour, not a trivial fix, so I've milestoned for 3.0 instead of backporting.

sunpy/map/mapbase.py
dstansby: Yep, it looks like maybe sphinx automatically inherits docstring, but python doesn't.

dstansby: I think these lines are fine, as if we don't have a docstring it's better to have the GenericMap one than nothing.

dstansby (commented):
Looks good to me - I'm happy to approve when @ayshih suggested change to the order has been made.

dstansby (approved):

ayshih (dismissed):
I think the output from the PR is confusing ([example for CORMap](https://sunpy--5122.org.readthedocs.build/en/5122/api/sunpy.map.sources.CORMap.html#sunpy.map.sources.CORMap)):
- The initial docstring of `GenericMap` ("A Generic spatially-aware 2D data array") is not stripped out or otherwise prefaced.
- The example in the docstring `GenericMap` is not stripped out, which means that every non-`AIAMap` class shows an example of using `AIAMap`.  (As an aside, why does the docstring for `GenericMap` not show `GenericMap`?)

This demonstrates that we can't do a simple string addition.  There needs to be a deeper level of refactoring so that docstring elements can be re-used in the docstrings of both `GenericMap` and the subclasses.

sunpy/map/mapbase.py
ayshih: > Yep, it looks like maybe sphinx automatically inherits docstring, but python doesn't.

It's actually that Python will read the parent class docstring for `help()`, but the `__doc__` attribute itself is not inherited.



> I think these lines are fine, as if we don't have a docstring it's better to have the GenericMap one than nothing.

I disagree, because we should never have no docstring.  We should mandate that every subclass have a docstring, even if it's one sentence.  Otherwise, you get silly stuff like this (https://sunpy--5122.org.readthedocs.build/en/5122/code_ref/map.html#id3), where `HMISynopticMap` claims to be "A Generic spatially-aware 2D data array"

asinghgaba:
@ayshih I think this should be done by adding docstrings individually to each subclass. We could then strip as much we want according to the subclass. What do you think about doing it this way?

ayshih:
I (mostly) disagree.  We want the creation of the merged docstring to be under `GenericMap` to facilitate maintenance and to ensure a standard presentation.  A person creating a new `GenericMap` subclass should be able to write a simple docstring, and have the merged docstring be automatically created.  If we want to support customization for unusual situations, `GenericMap.__init_subclass__` can call an overridable class method (e.g., `_make_docstring()`.

dstansby:
What do people think of adding some text along the lines of

> For more information on map attributes and methods, see documentation for `sunpy.map.GenericMap`

instead? That way there's still a clear distinction between the docstrings of the two classes, but anyone using `help(MyMapSource)` has a pointer to `GenericMap`.

nabobalis:
I wonder if we want to either do that or break up the important part of the Generic Map docstring and append that. Similar to how we do that with Fido Clients.

sunpy/map/mapbase.py
```diff
@@ -174,6 +174,7 @@ def __init_subclass__(cls, **kwargs):
         `is_datasource_for` attribute.
         This is then passed into the Map Factory so we can register them.
         """
+        cls.__doc__ = cls.__doc__ + '\n' + GenericMap.__doc__.split("data array")[1].split("Examples")[0]
```
nabobalis, line 177: Not very pretty but it cuts out only the parameters section. Leaves the notes to the side.

sunpy/map/mapbase.py
```diff
@@ -162,9 +162,15 @@ class GenericMap(NDData):
...
         rest will be discarded.
     """
-
     _registry = dict()
 
+    @classmethod
+    def _construct_subclass_doc(cls):
+        split_subclass_doc = cls.__doc__.split("References")
+        genmap_params = GenericMap.__doc__.split("data array\n\n")[1].split("Examples")[0]
+        references = "References" + split_subclass_doc[1] if len(split_subclass_doc) > 1 else ""
```
nabobalis, line 171: I feel like there has to be a nicer way to do this.

Cadair:
Rather than taking this approach why not do something similar to what we do in coordinates, where the parameters are templated into the docstrings? Splitting up docstrings on magic sequences feels prone to hard to detect issues to me.

Cadair (approved):

nabobalis:
Final result: https://sunpy--5122.org.readthedocs.build/en/5122/api/sunpy.map.sources.AIAMap.html#aiamap

Cadair:
I don't think the double notes section is an issue, but I am kinda surprised that sphinx doesn't get angry about that, I feel like old numpydoc would have done lol.

nabobalis:
Thanks for getting us started on this @asinghgaba!

Issue #5109, opened 2021-03-17
Default map title for WISPR plots is not quite right


dstansby:
With the file downloaded from here: https://vso1.nascom.nasa.gov/data/psp/wispr/L1/20200125/psp_L1_wispr_20200125T000229_V1_2302.fits, running:

```python
wispr_map = sunpy.map.Map('/Users/dstansby/Downloads/psp_L1_wispr_20200125T000229_V1_2302.fits')
fig = plt.figure()
wispr_map.plot()
plt.show()
```
gives the plot below. The title is a bit odd, in particular the first "2 0" bit. We should work out how to fix this.

![Screenshot 2021-03-17 at 21 19 05](https://user-images.githubusercontent.com/6197628/111539952-68d2ee00-8766-11eb-980a-4bc93395f220.png)

jeffreypaul15:
The downloaded map has no meta info for `wavelength`or `waveunit` so it defaults to 0 and no unit. The `nickname` property is set by `meta['detector']` which is basically `2` for this particular map. Hence `2 0  2020-01-25 00:02:29`.

ayshih:
I don't think this is a "Good First Issue", and maybe not some of the other labels as well.  This WISPR file is being opened as a `GenericMap` because we have no instrument-specific subclass for this.  As such, there's no fix for this that can be WISPR-specific.  We can change what goes into `nickname`, but that changes it for any file that gets loaded as a `GenericMap`.

![Untitled](https://user-images.githubusercontent.com/991759/111890461-d168d780-89bf-11eb-83ee-c85b07e0bebe.png)

nabobalis:
Would adding a WISPR Map source be the easiest thing to do?

dstansby:
Yeah, I think adding a WISPR map class is the best way to go and we can them customize the nickname which is used as the plot title.

PR #5124, opened 2021-03-21
Only include measurement in map names if it is given

dstansby:
Improves the situation in https://github.com/sunpy/sunpy/issues/5109. I'm going to put this as no-backport as it changes what plots look like.

Note that this is superseeded by #5126, if/when that PR is merged.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -679,13 +679,21 @@ class GenericMap(NDData):
     @property
     def name(self):
         """Human-readable description of the Map."""
-        return self._base_name().format(measurement=self.measurement)
+        if self.measurement == 0 * u.Angstrom:
+            # Don't include measurement if it's the default fallback value
+            return self._base_name().format(measurement='')
+        else:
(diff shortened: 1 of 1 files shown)


sunpy/map/mapbase.py
```diff
@@ -679,13 +679,21 @@ def _base_name(self):
     @property
     def name(self):
         """Human-readable description of the Map."""
-        return self._base_name().format(measurement=self.measurement)
+        if self.measurement == 0 * u.Angstrom:
```
Cadair, line 682: My real question here is not that this isn't an improvement (it definitely is) but why are we defaulting the measurement to `0 * u.AA` rather than `None`?

wtbarnes:
Should this wait on #5126?

dstansby:
Should have cross posted, but this is superseeded by https://github.com/sunpy/sunpy/pull/5126

nabobalis:
> Should have cross posted, but this is superseeded by #5126

So we should close this in favour of #5126?

dstansby:
Sounds good

PR #5126, opened 2021-03-22
Return None for unknown pieces of metadata

dstansby:
xref https://github.com/sunpy/sunpy/pull/5124#discussion_r598600405. I have always wondered why these return default values when they are unknown. Instead, return `None`. This is a breaking change, but I think fine as long as we document it well? Opinions very welcome.


sunpy/map/mapbase.py
```diff
@@ -783,10 +792,8 @@ def waveunit(self):
 
         This is taken from the 'WAVEUNIT' FITS keyword.
         """
-        unit = self.meta.get("waveunit")
-        if unit is None:
-            return u.one
-        return u.Unit(unit)
+        if 'waveunit' in self.meta:
```
Cadair, line 795: This will break wavelength if `'wavelnth'` is set and waveunit isn't?

Cadair: I think this default actually makes sense, `u.one` is a better "null" value than `None` for a unit?

dstansby: I did think about that, but a dimensionless unit is still a unit and not really a null value.

Cadair: We could have this return None but then all the other things that rely on it need changing to not explode if it's None.

sunpy/map/mapbase.py
```diff
@@ -672,10 +680,12 @@ def unit(self):
...
-        return "{nickname} {{measurement}} {date}".format(
-            nickname=self.nickname,
-            date=parse_time(self.date).strftime(TIME_FORMAT)
-        )
+        if self.measurement is None:
+            format_str = "{nickname} {date}"
+        else:
+            format_str = "{nickname} {{measurement}} {date}"
+        return format_str.format(nickname=self.nickname,
+                                 date=parse_time(self.date).strftime(TIME_FORMAT))
```
Cadair, line 688: This is shorter and slightly less repetitive, but is it actually better??

```suggestion
        measurement_str = "{{measurement}} " if measurement else ""
        format_str = "{nickname} {measurement_str}{date}"
        return format_str.format(nickname=self.nickname,
                                 measurement=measurement_str
                                 date=parse_time(self.date).strftime(TIME_FORMAT))
```

nabobalis: Not want a space between `{measurement_str}{date}`?

Cadair: No, the space is conditionally in `measurement_str`.

dstansby:
Anyone got any ideas why my attempt at ignoring the warning didn't work?

sunpy/map/mapbase.py
dstansby: I think I prefer my version, but not too fussed.

nabobalis (approved):

sunpy/map/mapbase.py
Cadair, line 688: How about this?
```suggestion
        segments = [self.nickname, self.date.strftime(TIME_FORMAT)]
        if self.measurement is not None:
            segments.insert(1, "{measurement}")
        return " ".join(segments)
```

This also drops the `parse_time` as that should have been called in `self.date` *riiight*?

Cadair (approved):

dstansby:
@Cadair can you remember what we decided to do if `WAVEUNIT` Isn't present but `WAVELNTH` is? Should `map.wavelength` return:

1. None
2. A bare number
3. A `Quantity` with `u.dimensionless` units?

Cadair:
I would go for quantity with dimensionless units. I think thats what happens at the moment in this code anyway.

PR #5127, opened 2021-03-22
Transformations now ignore time differences below numerical precision

ayshih:
#5108 had the consequence that we stopped adding zero seconds to most parsed times.  It turns out that adding zero seconds causes a UTC time to be converted to TAI and then back to UTC, which can result in a delta of tens of picoseconds (see https://github.com/astropy/astropy/issues/8742#issuecomment-494562784).  So, once we stopped doing the unnecessary conversion in `parse_time()`, later manipulation of the `Time` instance could cause it to change by tens of picoseconds.

This is particularly problematic when working with 2D Helioprojective coordinates that are off the limb and transforming the coordinate to its own frame.  When the obstimes (and observer obstimes) are equal, we can bail out and return the same 2D coordinate.  When the obstimes are different, even by only tens of picoseconds, we would run through the whole HPC->HCC->HGS->HCC->HPC chain, so we would convert the coordinate to 3D, and so off-the-limb coordinates turn into NaN.  This is the cause of our current test failures.

~This PR modifies all time-equality checks in `sunpy.coordinates.transformations` to consider a time difference of only tens of picoseconds to be equal.  Once we depend on Astropy 4.2, we can use `Time.isclose()` (see astropy/astropy#10646), but that wouldn't really make the code here much cleaner.~

This PR modifies all time-equality checks in `sunpy.coordinates.transformations` to be performed in TAI.  This is stricter than the above approach of a tolerance.

### DIFF
--- a/sunpy/coordinates/transformations.py
+++ b/sunpy/coordinates/transformations.py
@@ -41,6 +41,7 @@ from astropy.coordinates.representation import (
 # Import erfa via astropy to make sure we are using the same ERFA library as Astropy
 from astropy.coordinates.sky_coordinate import erfa
 from astropy.coordinates.transformations import FunctionTransform, FunctionTransformWithFiniteDifference
+from astropy.time import Time
 
 from sunpy import log
 from sunpy.sun import constants
@@ -213,7 +214,7 @@ def _observers_are_equal(obs_1, obs_2):
     return np.atleast_1d((u.allclose(obs_1.lat, obs_2.lat) and
                           u.allclose(obs_1.lon, obs_2.lon) and
                           u.allclose(obs_1.radius, obs_2.radius) and
-                          obs_1.obstime == obs_2.obstime)).all()
+                          _times_are_equal(obs_1.obstime, obs_2.obstime))).all()
 
 
 def _check_observer_defined(frame):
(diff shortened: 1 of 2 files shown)


sunpy/coordinates/transformations.py
```diff
@@ -230,6 +231,15 @@ def _check_observer_defined(frame):
                                "but this is valid for only HeliographicCarrington frames.")
 
 
+def _times_are_equal(time_1, time_2):
```
dstansby, line 234: Might be a bit nicer to put this in `sunpy/time/time.py`?

ayshih: I prefer to keep this in `transformations.py`.  It's a helper function with some specialization, with the handling of possible non-`Time` input (namely `None`) and the included use of `np.all()`.  It could call a more general function in `sunpy.time`, but that'll be unnecessary once we have `Time.isclose()` with Astropy 4.2.

ayshih:
Okay, I've modified the fix approach.  By checking for equality in the TAI scale, we don't need to specify a tolerance.

nabobalis (approved):

sunpy/coordinates/transformations.py
```diff
@@ -230,6 +231,17 @@ def _check_observer_defined(frame):
...
+    # Checks whether times are equal
+    if isinstance(time_1, Time) and isinstance(time_2, Time):
+        # We explicitly perform the check in TAI to avoid possible numerical precision differences
+        # between a time in UTC and the same time after a UTC->TAI->UTC conversion
+        return np.all(time_1.tai == time_2.tai)
```
Cadair, line 239: Out of curiosity: Why do this in TAI rather than with jd1 and jd2?

Cadair (approved):

sunpy/coordinates/transformations.py
ayshih: `jd1`/`jd2` is just the format, not the scale.  Comparing `jd1`/`jd2` explicitly – without taking the extra step of forcing the scale conversion to TAI – isn't any better than the previous equality check and would exhibit the same undesirable behavior.

Cadair: Ah right, makes sense.

Issue #5006, opened 2021-02-17
Convert results from the HEKClient to Astropy Time and Quantity objects


dstansby:
As far as I can tell, all the data returned by a HEK search is either a bare number or a string.

My first instinct would be to automate it inside the client, but more generally, it would be good if there were a schema for the return type and we could automate these conversions in the client.

For times at least, it seems doing `sunpy.time.parse_time(result['hek']['my_time_column'])` does the job.
For quantities I haven't worked out if there is any information on the units in the query result yet.

1. Consider using the XML API (but this probably involves a rewrite) as it seems to have unit info whereas the json endpoint does not.
2. We work could work out post search functions to apply to known searches.

See https://github.com/sunpy/sunpy/issues/1792 for an older discussion on this topic.

Cadair:
More generally, it would be good if there were a schema for the return type and we could automate these conversions in the client.

dstansby:
Yes, definitely bonus points for that!

For times at least, it seems doing `sunpy.time.parse_time(result['hek']['my_time_column'])` does the job. For quantities I haven't worked out if there is any information on the units in the query result yet.

nabobalis:
> More generally, it would be good if there were a schema for the return type and we could automate these conversions in the client.

This feels like a separate issue.

Cadair:
Well my first instinct would be to automate it, not just write a big doc with "this column is this datatype" etc.

Kate028:
Hey @dstansby , I am having some doubts

> As far as I can tell, all the data returned by a HEK search is either a bare number or a string.

Are we talking about float or string datatype that we getting as output in this file https://www.lmsal.com/hek/VOEvent_Spec.html . 
where's the input file or maybe we downloading it from this url https://www.lmsal.com/hek/her ?

> It would be good to add a section to https://docs.sunpy.org/en/latest/guide/acquiring_data/hek.html explaining how to convert relevant columns to astropy.time.Time or astropy.units.Quantity objects.

Let me know if these are the right place to look up
`sunpy/net/hek/hek.py`
`docs/guide/acquiring_data/hek.rst`

Also, in this code block we are using method  `get_voevent()` to return string instead of astropy.time.Time or astropy.units.Quantity objects. 
Am i thinking clearly? Could you please elaborate a bit further on what we are supposed to do!
```
def get_voevent(self, as_dict=True,
                    base_url="http://www.lmsal.com/hek/her?"):
        """Retrieves the VOEvent object associated with a given event and
[...]

PR #5131, opened 2021-03-23
Converted time results from HEK Client to Astropy Time

jeffreypaul15:
### Description

Ref #5006
The `_convert_to_Time()` method replaces the `str` with the appropriate `astropy.time.core.Time` object.
The conversion to `astropy.units.quantity.Quantity` is still yet to be done based on input.

### DIFF
--- a/changelog/5131.feature.rst
+++ b/changelog/5131.feature.rst
@@ -1,0 +1,2 @@
+Added a :meth:`~sunpy.net.hek._convert_to_Time` method to `sunpy.net.hek`
+that converts times from `str` to `astropy.time.core.Time` object.
--- a/changelog/5131.feature.rst
+++ b/changelog/5131.feature.rst
@@ -1,2 +1,2 @@
-Added a :meth:`~sunpy.net.hek._convert_to_Time` method to `sunpy.net.hek`
+Added a :meth:`sunpy.net.hek.convert_time` method to `sunpy.net.hek`
 that converts times from `str` to `astropy.time.core.Time` object.
--- a/sunpy/net/hek/__init__.py
+++ b/sunpy/net/hek/__init__.py
@@ -1,1 +1,2 @@
 from sunpy.net.hek.hek import *
+# from sunpy.net.hek.hek import convert_time
--- a/sunpy/net/hek/__init__.py
+++ b/sunpy/net/hek/__init__.py
@@ -1,2 +1,1 @@
 from sunpy.net.hek.hek import *
-from sunpy.net.hek.hek import convert_time
--- a/sunpy/net/hek/__init__.py
+++ b/sunpy/net/hek/__init__.py
@@ -1,2 +1,2 @@
 from sunpy.net.hek.hek import *
-# from sunpy.net.hek.hek import convert_time
+from sunpy.net.hek.hek import convert_time
--- a/sunpy/net/hek/hek.py
+++ b/sunpy/net/hek/hek.py
@@ -55,6 +55,13 @@ class HEKClient(BaseClient):
     def __init__(self, url=DEFAULT_URL):
         self.url = url
 
+    def _convert_to_Time(self, results):
+        """ Convert time from str to astropy.time.core.Time"""
+        for result in results:
+            result['event_starttime'] = Time(result['event_starttime'])
+            result['event_endtime'] = Time(result['event_endtime'])
+            result['event_peaktime'] = Time(result['event_peaktime'])
+
     def _download(self, data):
         """ Download all data, even if paginated. """
         page = 1
@@ -73,6 +80,8 @@ class HEKClient(BaseClient):
             finally:
                 fd.close()
             results.extend(result['result'])
+
+            self._convert_to_Time(results)
 
             if not result['overmax']:
                 if len(results) > 0:
--- a/sunpy/net/hek/hek.py
+++ b/sunpy/net/hek/hek.py
@@ -9,13 +9,13 @@ from itertools import chain
 
 import astropy.table
 from astropy.table import Row
-from astropy.time import Time
 
 import sunpy.net._attrs as core_attrs
 from sunpy import log
 from sunpy.net import attr
 from sunpy.net.base_client import BaseClient, QueryResponseTable
 from sunpy.net.hek import attrs
+from sunpy.time import parse_time
 from sunpy.util import dict_keys_same, unique
 from sunpy.util.xml import xml_to_dict
 
@@ -35,11 +35,13 @@ def _freeze(obj):
 
 def convert_time(response_obj):
     """ Helper method that convert time from str to astropy.time.core.Time"""
-    _table = response_obj['hek']
-    _table.replace_column
-    _table.replace_column(name='event_starttime', col=Time(_table['event_starttime']))
-    _table.replace_column(name='event_endtime', col=Time(_table['event_endtime']))
-    _table.replace_column(name='event_peaktime', col=Time(_table['event_peaktime']))
+    if isinstance(response_obj, HEKTable):
+        _table = response_obj
+    else:
+        _table = response_obj['hek']
+    _table.replace_column(name='event_starttime', col=parse_time(_table['event_starttime']))
+    _table.replace_column(name='event_endtime', col=parse_time(_table['event_endtime']))
+    _table.replace_column(name='event_peaktime', col=parse_time(_table['event_peaktime']))
 
 
 class HEKClient(BaseClient):
--- a/sunpy/net/hek/hek.py
+++ b/sunpy/net/hek/hek.py
@@ -34,7 +34,9 @@ def _freeze(obj):
 
 
 def convert_time(response_obj):
-    """ Helper method that convert time from str to astropy.time.core.Time"""
+    """
+    Converts time from str to astropy.time.core.Time for a HEKTable.
+    """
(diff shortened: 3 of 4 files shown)


Cadair:
Thanks for this PR, but following on from the conversation at the community meeting today I think that we shouldn't take this approach of converting known columns, we should work with the HEK's VO table or other mechanisms to solve this in a general manner.

The only other thing I would suggest is turning this into a helper function which can optionally be applied to the results to convert the known columns.

If you don't want to turn this PR into the latter option, then please close this.

jeffreypaul15:
> Thanks for this PR, but following on from the conversation at the community meeting today I think that we shouldn't take this approach of converting known columns, we should work with the HEK's VO table or other mechanisms to solve this in a general manner.
> 
> The only other thing I would suggest is turning this into a helper function which can optionally be applied to the results to convert the known columns.
> 
> If you don't want to turn this PR into the latter option, then please close this.

Okay, that makes sense. What would the helper function contain though? How would we want to write it?

nabobalis:
The helper function would probably look like the function you wrote, we would probably not call it automatically but tell users in a guide or docstring it's there for use.

jeffreypaul15:
> The helper function would probably look like the function you wrote, we would probably not call it automatically but tell users in a guide or docstring it's there for use.

So, just convert this into a method that would be called only if the user desires right? How would we want to design that? Have a separate argument for this within the HEK client?

nabobalis:
> > The helper function would probably look like the function you wrote, we would probably not call it automatically but tell users in a guide or docstring it's there for use.
> 
> So, just convert this into a method that would be called only if the user desires right?

Yeah.

> How would we want to design that? Have a separate argument for this within the HEK client?

I don't think this would be an argument, maybe a function a user can import separately and then apply on the results object.

I am unsure of the path required in code to make that an argument.

jeffreypaul15:
> maybe a function a user can import separately and then apply on the results object.

Yeah this makes more sense.

jeffreypaul15:
> > > The helper function would probably look like the function you wrote, we would probably not call it automatically but tell users in a guide or docstring it's there for use.
> > 
> > 
> > So, just convert this into a method that would be called only if the user desires right?
> 
> Yeah.
> 
> > How would we want to design that? Have a separate argument for this within the HEK client?
> 
> I don't think this would be an argument, maybe a function a user can import separately and then apply on the results object.
> 
> I am unsure of the path required in code to make that an argument.

Would this be exclusive to HEK Client responses? Or should I place this method elsewhere?

nabobalis:
> Would this be exclusive to HEK Client responses? Or should I place this method elsewhere?

I don't see why it wouldn't be.

jeffreypaul15:
> > Would this be exclusive to HEK Client responses? Or should I place this method elsewhere?
> 
> I don't see why it wouldn't be.

I'm a bit confused on where to place this method exactly, if it's not exclusive then do you have any suggestions?

nabobalis:
> > > Would this be exclusive to HEK Client responses? Or should I place this method elsewhere?
> > 
> > 
> > I don't see why it wouldn't be.
> 
> I'm a bit confused on where to place this method exactly, if it's not exclusive then do you have any suggestions?

Same place as the client?

jeffreypaul15:
> > > > Would this be exclusive to HEK Client responses? Or should I place this method elsewhere?
> > > 
> > > 
> > > I don't see why it wouldn't be.
> > 
> > 
> > I'm a bit confused on where to place this method exactly, if it's not exclusive then do you have any suggestions?
> 
> Same place as the client?

But we want this to work for all clients right?

nabobalis:
> > > > > Would this be exclusive to HEK Client responses? Or should I place this method elsewhere?
> > > > 
> > > > 
> > > > I don't see why it wouldn't be.
> > > 
> > > 
> > > I'm a bit confused on where to place this method exactly, if it's not exclusive then do you have any suggestions?
> > 
> > 
> > Same place as the client?
> 
> But we want this to work for all clients right?

No?

jeffreypaul15:
> > > > > > Would this be exclusive to HEK Client responses? Or should I place this method elsewhere?
> > > > > 
> > > > > 
> > > > > I don't see why it wouldn't be.
> > > > 
> > > > 
> > > > I'm a bit confused on where to place this method exactly, if it's not exclusive then do you have any suggestions?
> > > 
> > > 
> > > Same place as the client?
> > 
> > 
> > But we want this to work for all clients right?
> 
> No?

Oh okay, I misunderstood. Sorry.

jeffreypaul15:
```python
from sunpy.net import Fido
from sunpy.net.hek.hek import convert_time
import sunpy
from astropy.time import Time

event_type = "FL"
tstart = "2013/10/28"
tend = "2013/10/29"
result = Fido.search(a.Time(tstart, tend),
                     a.hek.EventType(event_type),
                     a.hek.OBS.Observatory == "GOES")

convert_time(result)

```
This should produce the required result.
I do have a few questions though :
- Would it be better to ` __init__` this method so the user imports it from from `sunpy.net.hek` instead of from `sunpy.net.hek.hek`?
- I'm importing `from astropy.time import Time`, should I just use `sunpy.time.parse_time` instead of having that import?
- Does this need a test?

nabobalis:
> * Would it be better to ` __init__` this method so the user imports it from from `sunpy.net.hek` instead of from `sunpy.net.hek.hek`?

Sure.

> * I'm importing `from astropy.time import Time`, should I just use `sunpy.time.parse_time` instead of having that import?

Yes.

> * Does this need a test?

Yes.

Ideally the function would do more, but it's ok for as a starting platform.

jeffreypaul15:
@nabobalis is this good to go?

changelog/5131.feature.rst
```diff
@@ -0,0 +1,2 @@
+Added a :meth:`~sunpy.net.hek._convert_to_Time` method to `sunpy.net.hek`
```
nabobalis, line 1:
```suggestion
Added a :meth:`~sunpy.net.hek.convert_time` method to `sunpy.net.hek`
```

sunpy/net/hek/__init__.py
```diff
@@ -1 +1,2 @@
 from sunpy.net.hek.hek import *
+from sunpy.net.hek.hek import convert_time
```
nabobalis, line 2:
```suggestion
```

sunpy/net/hek/hek.py
```diff
@@ -9,13 +9,13 @@
...
-from astropy.time import Time
 
 import sunpy.net._attrs as core_attrs
 from sunpy import log
 from sunpy.net import attr
 from sunpy.net.base_client import BaseClient, QueryResponseTable
 from sunpy.net.hek import attrs
+from sunpy.time import parse_time
 from sunpy.util import dict_keys_same, unique
 from sunpy.util.xml import xml_to_dict
```
nabobalis, line 21: Add the function to the `__all__` below this.

sunpy/net/hek/hek.py
```diff
@@ -33,6 +33,17 @@ def _freeze(obj):
     return obj
 
 
+def convert_time(response_obj):
+    """ Helper method that convert time from str to astropy.time.core.Time"""
```
nabobalis, line 37:
```suggestion
    """
    Converts time from str to astropy.time.core.Time for a hek response.
    """
```

sunpy/net/hek/tests/test_hek.py
```diff
@@ -197,3 +199,12 @@ def test_mixed_results_get_angstrom():
     result = client.search(attrs.Time(tstart, tend), attrs.hek.EventType(event_type))
     assert len(result) == 13
     assert result[0]["SOL_standard"] == 'SOL2014-10-24T20:53:46L247C106'
+
+
+@pytest.mark.remote_data
+def test_convert_time(hek_client_creator):
```
nabobalis, line 205: Does it work with a return from fido?

jeffreypaul15: Yes, it does. Would we need a test for that as well?

nabobalis: Sure.

jeffreypaul15: > Sure.

Where would this test go though? Should I import fido and write it here itself?

nabobalis: Yeah.

sunpy/net/hek/hek.py
```diff
@@ -33,6 +33,19 @@ def _freeze(obj):
     return obj
 
 
+def convert_time(response_obj):
+    """
+    Converts time from str to astropy.time.core.Time for a HEKTable.
+    """
```
nabobalis, line 39: I think this docstring needs expanding to have a parameters section, also a note to say this does the conversion in place. 

Maybe we want it to return a new repesonse?

dstansby:
I'm -1 to merging something like this that isn't a generic solution, as it just increases the amount of code we have to maintain without *really* solving the issue in the long run (as it stands we're just guessing at column names that need converting).

dstansby:
I think it's worth discussing this specific PR at our next weekly meeting, and we can come to a collective decision on whether to go forward with it or not.

nabobalis:
Following up in this weekly meeting. We don't want to add code to the client that deals with conversion of the table columns in an ad hoc manner. We need to work out a schema for the conversion to start with.

PR #5142, opened 2021-03-24
Show the inherited attributes/methods in the docs for Map subclasses

ayshih:
I don't know why this wasn't done before.  For example, https://docs.sunpy.org/en/stable/api/sunpy.map.sources.AIAMap.html doesn't show nearly any of the attributes/methods that a user would actually want to use.


nabobalis:
The rendered version: https://sunpy--5142.org.readthedocs.build/en/5142/api/sunpy.map.sources.AIAMap.html#sunpy.map.sources.AIAMap

nabobalis (approved):

nabobalis (approved):

nabobalis (approved):

PR #2301, opened 2017-10-13
Creating MapCube.save(prefix_path)

jerinphilip:
Attempting to build https://github.com/sunpy/sunpy/issues/2210.

Now how do I test this? I tried looking for a test for the save function in map, didn't find any.
I've already found the tests located at `.../map/tests/*`, and once went through the [testing documentation](http://docs.sunpy.org/en/latest/dev_guide/tests.html).

Module level tests are failing in my machine.
```bash
$ py.test sunpy/map/tests/
sunpy/map/tests/test_header.py .
sunpy/map/tests/test_map_factory.py ...........
sunpy/map/tests/test_mapbase.py ............................................................
sunpy/map/tests/test_mapcube.py ....
sunpy/map/tests/test_plotting.py FFFFFFFFFFFFFFFFFFF
```

@wafels @Punyaslok ^_^

PR #5145, opened 2021-03-25
save method for MapSequence

jeffreypaul15:
### Description

Fixes #2301
Adds a save method to `MapSequence` where it saves each map from the sequence with `_{map_index_number}` as the suffix.
Does this have to be tested?

### DIFF
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -537,7 +537,7 @@ class MapSequence:
         >>> smap = Map(sunpy.data.sample.HMI_LOS_IMAGE,
         ...            sunpy.data.sample.AIA_1600_IMAGE,
         ...            sequence=True)  # doctest: +REMOTE_DATA
-        >>> smap.save('map_{index:03}.fits')  # doctest: +REMOTE_DATA
+        >>> smap.save('map_{index:03}.fits')  # doctest: +SKIP
 
         """
         if filepath.format(index=0) == filepath:
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -513,7 +513,8 @@ class MapSequence:
         return [m.meta for m in self.maps]
 
     def save(self, filepath, filetype='auto', **kwargs):
-        """Saves the SunPy Map Sequence object to a file.
+        """
+        Saves the SunPy Map Sequence object to a file.
 
         Currently SunPy can only save files in the FITS format.
 
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -532,15 +532,11 @@ class MapSequence:
         Notes
         -----
         The individual maps are saved as with the specified
-        `filepath` followed by an `_` with a sequence number.
-        If an extension isn't specified it is defaulted to `fits`.
+        `filepath` based on {index} specified by the user.
         """
-        filepath_ = Path(filepath)
         for index, map_seq in enumerate(self.maps):
-            extension = filepath_.suffix
-            path = (filepath_.parent / filepath_.stem).as_posix()
-            if extension == '':
-                path = f"{path}_{index:03}.fits"
+            filepath_ = filepath.format(index=index)
+            if filepath_ == filepath:
+                raise ValueError("{index} must be specified")
             else:
-                path = f"{path}_{index:03}{extension}"
-            map_seq.save(path, filetype, **kwargs)
+                map_seq.save(filepath_, filetype, **kwargs)
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -5,8 +5,8 @@ import textwrap
 import warnings
 import webbrowser
 from copy import deepcopy
-from tempfile import NamedTemporaryFile
 from pathlib import Path
+from tempfile import NamedTemporaryFile
 
 import matplotlib.animation
 import numpy as np
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -5,6 +5,7 @@ import textwrap
 import warnings
 import webbrowser
 from copy import deepcopy
+from tempfile import NamedTemporaryFile
 from pathlib import Path
 
 import matplotlib.animation
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -515,8 +515,7 @@ class MapSequence:
     def save(self, filepath, filetype='auto', **kwargs):
         """Saves the SunPy Map Sequence object to a file.
 
-        Currently SunPy can only save files in the FITS format. In the future
-        support will be added for saving to other formats.
+        Currently SunPy can only save files in the FITS format.
 
         Parameters
         ----------
@@ -532,7 +531,7 @@ class MapSequence:
         Notes
         -----
         The individual maps are saved as with the specified
-        filepath followed by an `_` with the sequence number.
+        `filepath` followed by an `_` with a sequence number.
         If an extension isn't specified it is defaulted to `fits`.
         """
         filepath_ = Path(filepath)
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -5,7 +5,7 @@ import textwrap
 import warnings
 import webbrowser
 from copy import deepcopy
-from tempfile import NamedTemporaryFile
+from pathlib import Path
 
 import matplotlib.animation
 import numpy as np
@@ -513,7 +513,7 @@ class MapSequence:
         return [m.meta for m in self.maps]
 
     def save(self, filepath, filetype='auto', **kwargs):
-        """Saves the SunPy Map object to a file.
+        """Saves the SunPy Map Sequence object to a file.
 
         Currently SunPy can only save files in the FITS format. In the future
         support will be added for saving to other formats.
@@ -528,7 +528,19 @@ class MapSequence:
         kwargs :
             Any additional keyword arguments are passed to
             `~sunpy.io.write_file`.
+
+        Notes
+        -----
+        The individual maps are saved as with the specified
+        filepath followed by an `_` with the sequence number.
+        If an extension isn't specified it is defaulted to `fits`.
         """
+        filepath_ = Path(filepath)
         for index, map_seq in enumerate(self.maps):
-            _filepath = filepath + "_{}".format(index)
-            map_seq.save(_filepath, filetype, **kwargs)
+            extension = filepath_.suffix
+            path = (filepath_.parent / filepath_.stem).as_posix()
+            if extension == '':
+                path = f"{path}_{index:03}.fits"
+            else:
+                path = f"{path}_{index:03}{extension}"
+            map_seq.save(path, filetype, **kwargs)
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -535,7 +535,7 @@ class MapSequence:
         --------
         >>> from sunpy.map import Map
         >>> import sunpy.data.sample # doctest: +REMOTE_DATA
-        >>> smap = Map(sunpy.data.sample.AIA_171_IMAGE, sunpy.data.sample.AIA_193_IMAGE, sequence=True) # doctest: +REMOTE_DATA
+        >>> smap = Map(sunpy.data.sample.HMI_LOS_IMAGE, sunpy.data.sample.AIA_1600_IMAGE, sequence=True) # doctest: +REMOTE_DATA
         >>> smap.save('map_{index:03}.fits') # doctest: +REMOTE_DATA
 
         """
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -5,7 +5,6 @@ import textwrap
 import warnings
 import webbrowser
 from copy import deepcopy
-from pathlib import Path
 from tempfile import NamedTemporaryFile
 
 import matplotlib.animation
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -530,5 +530,5 @@ class MapSequence:
             `~sunpy.io.write_file`.
         """
         for index, map_seq in enumerate(self.maps):
-            _filepath = filepath+"_{}".format(index)
+            _filepath = filepath + "_{}".format(index)
             map_seq.save(_filepath, filetype, **kwargs)
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -511,3 +511,24 @@ class MapSequence:
         Return all the meta objects as a list.
         """
         return [m.meta for m in self.maps]
+
+    def save(self, filepath, filetype='auto', **kwargs):
+        """Saves the SunPy Map object to a file.
+
+        Currently SunPy can only save files in the FITS format. In the future
+        support will be added for saving to other formats.
+
+        Parameters
+        ----------
+        filepath : str
+            Location to save file to. 
+            Each map from the sequence will be saved separately.
+        filetype : str
+            'auto' or any supported file extension.
+        kwargs :
+            Any additional keyword arguments are passed to
+            `~sunpy.io.write_file`.
+        """
+        for index, map_seq in enumerate(self.maps):
+            _filepath = filepath+"_{}".format(index)
+            map_seq.save(_filepath, filetype, **kwargs)
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -522,16 +522,21 @@ class MapSequence:
         filepath : str
             Location to save file to.
             Each map from the sequence will be saved separately.
+            The individual maps are saved based on {index} specified
+            by the user.
         filetype : str
             'auto' or any supported file extension.
         kwargs :
             Any additional keyword arguments are passed to
             `~sunpy.io.write_file`.
 
-        Notes
-        -----
-        The individual maps are saved as with the specified
-        `filepath` based on {index} specified by the user.
+        Examples
+        --------
+        >>> from sunpy.map import Map
+        >>> import sunpy.data.sample
+        >>> smap = Map(sunpy.data.sample.AIA_171_IMAGE, sunpy.data.sample.AIA_193_IMAGE, sequence=True)
+        >>> smap.save('map_{index:03}.fits')
+
         """
         for index, map_seq in enumerate(self.maps):
             filepath_ = filepath.format(index=index)
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -521,7 +521,7 @@ class MapSequence:
         Parameters
         ----------
         filepath : str
-            Location to save file to. 
+            Location to save file to.
             Each map from the sequence will be saved separately.
         filetype : str
             'auto' or any supported file extension.
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -534,9 +534,9 @@ class MapSequence:
         Examples
         --------
         >>> from sunpy.map import Map
-        >>> import sunpy.data.sample
-        >>> smap = Map(sunpy.data.sample.AIA_171_IMAGE, sunpy.data.sample.AIA_193_IMAGE, sequence=True)
-        >>> smap.save('map_{index:03}.fits')
+        >>> import sunpy.data.sample # doctest: +REMOTE_DATA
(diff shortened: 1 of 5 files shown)


ayshih:
> Does this have to be tested?

The answer is always "yes" =P

sunpy/map/mapsequence.py
```diff
@@ -511,3 +511,24 @@ def all_meta(self):
         Return all the meta objects as a list.
         """
         return [m.meta for m in self.maps]
+
+    def save(self, filepath, filetype='auto', **kwargs):
+        """Saves the SunPy Map object to a file.
```
nabobalis, line 516:
```suggestion
        """
        Saves the SunPy Map object to a file.
```

sunpy/map/mapsequence.py
```diff
@@ -511,3 +511,24 @@ def all_meta(self):
...
+
+        Currently SunPy can only save files in the FITS format. In the future
+        support will be added for saving to other formats.
```
nabobalis, line 519:
```suggestion
        Currently this only supports saving files in the FITS format.
```

sunpy/map/mapsequence.py
```diff
@@ -511,3 +511,24 @@ def all_meta(self):
...
+
+        Parameters
+        ----------
+        filepath : str
+            Location to save file to.
+            Each map from the sequence will be saved separately.
```
nabobalis, line 525:
```suggestion

```

sunpy/map/mapsequence.py
```diff
@@ -511,3 +511,24 @@ def all_meta(self):
...
+        filetype : str
+            'auto' or any supported file extension.
```
nabobalis, line 527: If we only support Fits, should this not default to fits for now?

jeffreypaul15: Yeah, I was thinking about that even for `GenericMap`

nabobalis: Hmm, I assume auto just defaults to fits in the io.write function?

jeffreypaul15: No, it doesn't, we get a `ValueError: The filetype provided () is not supported`.

sunpy/map/mapsequence.py
```diff
@@ -511,3 +511,24 @@ def all_meta(self):
...
+        filepath : str
+            Location to save file to.
+        filetype : str
+            'auto' or any supported file extension.
+        kwargs :
+            Any additional keyword arguments are passed to
+            `~sunpy.io.write_file`.
+        """
+        for index, map_seq in enumerate(self.maps):
+            _filepath = filepath+"_{}".format(index)
```
nabobalis, line 533:
```suggestion
            _filepath = filepath + "_{}".format(index)
```
Does the style check not complain about this?

sunpy/map/tests/test_mapsequence.py
```diff
@@ -196,3 +198,23 @@ def test_norm_animator(hmi_test_map):
...
     seq = sunpy.map.Map([aia171_test_map, hmi_test_map], sequence=True)
     seq.plot()
+
+
+def test_save(aia171_test_map, hmi_test_map):
+    """Tests the map save function"""
+    seq = sunpy.map.Map([aia171_test_map, hmi_test_map], sequence=True)
+    afilename = tempfile.NamedTemporaryFile(suffix='fits').name
+    seq.save(afilename, filetype='fits', overwrite=True)
+    test_seq = sunpy.map.Map(afilename+"_0", afilename+"_1", sequence=True)
```
nabobalis, line 208: So the saved files don't have the fits extension?

sunpy/map/mapsequence.py
nabobalis: I am not sure I am pro this as a name but at the same time, I don't have a better idea.

sunpy/map/tests/test_mapsequence.py
jeffreypaul15: Nope, here it just adds fits to the end of the filename, not really an extension. Took this from `GenericMap` and updated it, should I change that as well?

sunpy/map/mapsequence.py
nabobalis: Sorry, I mean that if you provide "auto" for save, it will always default to a fits file?

sunpy/map/tests/test_mapsequence.py
nabobalis: You mean the Map load adds the "fits" extension on load?

jeffreypaul15: Oh wait, I got confused, it is of `FITS` type but has no extension such as `filename.fits`. 
This is provided the `filetype` attribute is set = `fits`.

sunpy/map/mapsequence.py
jeffreypaul15: The default value in this method is `auto` right? Even if 'auto' is provided the same error occurs.

sunpy/map/tests/test_mapsequence.py
nabobalis: So the saved file has no "fits" extension in the file name?

sunpy/map/mapsequence.py
nabobalis: What errors? I am slightly confused.

sunpy/map/tests/test_mapsequence.py
jeffreypaul15: No, they're of type `FITS` but are saved with just the filename

sunpy/map/mapsequence.py
jeffreypaul15: When `filename=auto` is provided to `save`, we get `ValueError: The filetype provided () is not supported.`

sunpy/map/tests/test_mapsequence.py
nabobalis: Ah I see. I think changing that is outside of scope for this PR.

sunpy/map/mapsequence.py
nabobalis: I wonder if thats a bug. I am not sure what is going on.

jeffreypaul15: Ah, so this occurs if we don't specify `filepath` with an extension. 
```python
map.save(filepath = "/home/map.fits", filetype='auto') # works fine.
```
```python
map.save(filepath = "/home/map", filetype='auto') # bugs out
```
The fault is in `sunpy.io.file_tools`
`filetype = pathlib.Path(fname).suffix[1:]`
Here, `filetype` is an empty string if no extension ("/home/map") is provided and this  raises the ValueError

nabobalis: I see, leave it then.

jeffreypaul15: So, that works as intended right?

nabobalis: Yeah

jeffreypaul15:
I wonder why the Core Test py39 is failing with the astropy warning. It should be using the same maps that are tested in `test_mapbase.py` and there are no warnings there though.

nabobalis:
Are both maps saved out in the mapbase tests?

jeffreypaul15:
> Are both maps saved out in the mapbase tests?

Ah, that could be the reason.

jeffreypaul15:
@nabobalis is this good to go?

nabobalis:
Will need two reviews.

sunpy/map/mapsequence.py
```diff
@@ -511,3 +511,24 @@ def all_meta(self):
...
+        """
+        Saves the maps in the map sequence to disk.
+        Each map from the sequence will be saved separately.
+
+        Currently this only supports saving files in the FITS format.
+
+        Parameters
+        ----------
+        filepath : str
+            Location to save file to.
```
ayshih, line 525: This is unclear whether this is supposed to be a directory name or a filename stub.

sunpy/map/mapsequence.py
```diff
@@ -511,3 +511,24 @@ def all_meta(self):
...
+        filetype : str
+            'auto' or any supported file extension.
+        kwargs :
+            Any additional keyword arguments are passed to
+            `~sunpy.io.write_file`.
+        """
+        for index, map_seq in enumerate(self.maps):
+            _filepath = filepath + "_{}".format(index)
+            map_seq.save(_filepath, filetype, **kwargs)
```
ayshih, line 534:
```suggestion
            map_seq.save(f"{filepath}_{index}", filetype, **kwargs)
```
But see my overarching comment

sunpy/map/tests/test_mapsequence.py
```diff
@@ -196,3 +198,23 @@ def test_norm_animator(hmi_test_map):
 def test_map_sequence_plot(aia171_test_map, hmi_test_map):
     seq = sunpy.map.Map([aia171_test_map, hmi_test_map], sequence=True)
     seq.plot()
+
+
+def test_save(aia171_test_map, hmi_test_map):
+    """Tests the map save function"""
+    seq = sunpy.map.Map([aia171_test_map, hmi_test_map], sequence=True)
+    afilename = tempfile.NamedTemporaryFile(suffix='fits').name
```
ayshih, line 206: This usage of `tempfile` is poor form.  You've created a temporary filename, but then you create two actual files with modified versions of that filename.  Not only have you opened yourself to the (admittedly unlikely) possibility of a filename collision, you aren't doing any cleanup of the files (and `tempfile` won't be able to clean them up automatically).

You should be using `tempfile.TemporaryDirectory()` and working within that space.

sunpy/map/mapsequence.py
```diff
@@ -511,3 +511,24 @@ def all_meta(self):
         Return all the meta objects as a list.
         """
         return [m.meta for m in self.maps]
+
+    def save(self, filepath, filetype='auto', **kwargs):
+        """
+        Saves the maps in the map sequence to disk.
+        Each map from the sequence will be saved separately.
```
ayshih, line 518: The naming approach and how `filepath` is used needs to be documented

ayshih (changes requested):
I'm :-1: on the filenames for the saved files in the current implementation.  Whatever string that is passed as `filepath` is simply appended to with an underscore+number (e.g., if `filepath='test/blah'`, the first map is saved as `test/blah_0`).  I think it's highly unsatisfying to not have a way to achieve normal file extensions ('.fits') on the saved files.

sunpy/map/tests/test_mapsequence.py
jeffreypaul15: I did use the same method used in `GenericMap` I shall make this change.

nabobalis: I think we can update the code in GenericMap as well then.

jeffreypaul15: Should I use Pytest's `tmpdir` instead of `tempfile` here?

nabobalis: Sure.

ayshih: > I think we can update the code in GenericMap as well then.

The code in the test for `GenericMap.save()` is fine because only one file is being created, and it's properly using the temporary filename provided by `tempfile`.  Thus, it was also be cleaned up automatically.

ayshih:
My preference is to see an API like where `filepath` can be specified as `'mydirectory/mymapsequence_{index}.fits'`, and the files will be named appropriately.  Bonus points if `'mydirectory/mymapsequence_{index:03}.fits'` can be provided for formatting the index (here, with padded zeros).

ayshih:
Padding with zeros means that the filenames stay the same length, and is more friendly to (non-smart) sorting of filenames.  That is, the difference between:
- `mymapsequence_1.fits` to `mymapsequence_100.fits`, where `mymapsequence_2.fits` might get sorted after `mymapsequence_100.fits`
- `mymapsequence_001.fits` to `mymapsequence_100.fits`

jeffreypaul15:
`map.save('example.fits', filetype='fits', overwrite=True)` should save the sequence as :
`example_001.fits`
`example_002.fits`

`map.save('example.fits', filetype='auto', overwrite=True)` would also do the same (similar to how `GenericMap` save was intended).

`map.save('example', filetype='fits', overwrite=True)` saves the sequence with a `.fits` extension by default but this can be removed if not required.

ayshih:
I think you didn't understand me with respect to API.  It should be both more explicit and more general than you have implemented.  For example:

- If the user wants `example_001.fits`, `example_002.fits`, etc., they should explicitly provide `filepath='example_{index:03}.fits'`.
- If the user wants `example_1.fits`, `example_2.fits`, etc., they should explicitly provide `filepath='example_{index}.fits'`.
- If the user wants something quite different, e.g., `frame0001ofsequence.fts`, `frame0002ofsequence.fts`, etc., they would simply provide `filepath='frame{index:04}ofsequence.fts'`.

We can defer the flexible number formatting to a future PR, but I see the explicit designation of `{index}` in the string to be very important.  The input `filepath='example.fits'` should not be accepted.

jeffreypaul15:
> I think you didn't understand me with respect to API. It should be both more explicit and more general than you have implemented. For example:
> 
>     * If the user wants `example_001.fits`, `example_002.fits`, etc., they should explicitly provide `filepath='example_{index:03}.fits'`.
> 
>     * If the user wants `example_1.fits`, `example_2.fits`, etc., they should explicitly provide `filepath='example_{index}.fits'`.
> 
>     * If the user wants something quite different, e.g., `frame0001ofsequence.fts`, `frame0002ofsequence.fts`, etc., they would simply provide `filepath='frame{index:04}ofsequence.fts'`.
> 
> 
> We can defer the flexible number formatting to a future PR, but I see the explicit designation of `{index}` in the string to be very important. The input `filepath='example.fits'` should not be accepted.

Oh, my apologies, I misunderstood.
I shall make this fix.

jeffreypaul15:
> I think you didn't understand me with respect to API. It should be both more explicit and more general than you have implemented. For example:
> 
>     * If the user wants `example_001.fits`, `example_002.fits`, etc., they should explicitly provide `filepath='example_{index:03}.fits'`.
> 
>     * If the user wants `example_1.fits`, `example_2.fits`, etc., they should explicitly provide `filepath='example_{index}.fits'`.
> 
>     * If the user wants something quite different, e.g., `frame0001ofsequence.fts`, `frame0002ofsequence.fts`, etc., they would simply provide `filepath='frame{index:04}ofsequence.fts'`.
> 
> 
> We can defer the flexible number formatting to a future PR, but I see the explicit designation of `{index}` in the string to be very important. The input `filepath='example.fits'` should not be accepted.

This API should properly work now, the documentation needs some work though.

sunpy/map/mapsequence.py
```diff
@@ -511,3 +511,31 @@ def all_meta(self):
         Return all the meta objects as a list.
         """
         return [m.meta for m in self.maps]
+
+    def save(self, filepath, filetype='auto', **kwargs):
+        """Saves the SunPy Map Sequence object to a file.
```
nabobalis, line 516:
```suggestion
        """
        Saves the SunPy Map Sequence object to a file.
```

sunpy/map/mapsequence.py
```diff
@@ -511,3 +511,31 @@ def all_meta(self):
...
+        ----------
+        filepath : str
+            Location to save file to.
+            Each map from the sequence will be saved separately.
+        filetype : str
+            'auto' or any supported file extension.
+        kwargs :
+            Any additional keyword arguments are passed to
+            `~sunpy.io.write_file`.
+
```
nabobalis, line 530: I think a doc example might be useful.

sunpy/map/mapsequence.py
```diff
@@ -511,3 +511,31 @@ def all_meta(self):
...
+        Notes
+        -----
+        The individual maps are saved as with the specified
+        `filepath` based on {index} specified by the user.
```
nabobalis, line 534: I feel like this should be in the "filepath" part. This seems to important to be at the bottom.

sunpy/map/tests/test_mapsequence.py
```diff
@@ -196,3 +197,28 @@ def test_norm_animator(hmi_test_map):
 def test_map_sequence_plot(aia171_test_map, hmi_test_map):
     seq = sunpy.map.Map([aia171_test_map, hmi_test_map], sequence=True)
     seq.plot()
+
+
+def test_save(aia171_test_map, hmi_test_map, tmp_path):
+    """Tests the MapSequence save function"""
```
nabobalis, line 203:
```suggestion
    """
    Tests the MapSequence save function
    """
```

sunpy/map/mapsequence.py
jeffreypaul15: I'm confused as to where exactly you want this part

sunpy/map/mapsequence.py
jeffreypaul15: Alright, I shall add that

sunpy/map/mapsequence.py
nabobalis: In the filepath doc part under parameters.

sunpy/map/mapsequence.py
```diff
@@ -511,3 +511,37 @@ def all_meta(self):
...
+        >>> smap = Map(sunpy.data.sample.AIA_171_IMAGE, sunpy.data.sample.AIA_193_IMAGE, sequence=True) # doctest: +REMOTE_DATA
+        >>> smap.save('map_{index:03}.fits') # doctest: +REMOTE_DATA
+
+        """
+        for index, map_seq in enumerate(self.maps):
+            filepath_ = filepath.format(index=index)
+            if filepath_ == filepath:
+                raise ValueError("{index} must be specified")
+            else:
+                map_seq.save(filepath_, filetype, **kwargs)
```
ayshih, line 547:
```suggestion
        if filepath.format(index=0) == filepath:
            raise ValueError("'{index}' must be appear in the string")

        for index, map_seq in enumerate(self.maps):
            map_seq.save(filepath.format(index=index), filetype, **kwargs)
```

sunpy/map/tests/test_mapsequence.py
```diff
@@ -196,3 +197,30 @@ def test_norm_animator(hmi_test_map):
...
+    Tests the MapSequence save function
+    """
+    seq = sunpy.map.Map([aia171_test_map, hmi_test_map], sequence=True)
+
+    with pytest.raises(ValueError, match="{index} must be specified"):
+        seq.save(f"{tmp_path.as_posix()}_index:03", filetype='fits', overwrite=True)
+    seq.save(f"{tmp_path.as_posix()}_{{index:03}}.fits", filetype='auto', overwrite=True)
+
+    test_seq = sunpy.map.Map(f"{tmp_path.as_posix()}_000.fits",
+                             f"{tmp_path.as_posix()}_001.fits", sequence=True)
```
ayshih, line 213:
```suggestion
    with pytest.raises(ValueError, match="'{index}' must be appear in the string"):
        seq.save("index", filetype='fits', overwrite=True)

    base_str = (tmp_path / "map").as_posix()
    seq.save(f"{base_str}_{{index:03}}.fits", filetype='auto', overwrite=True)

    test_seq = sunpy.map.Map(f"{base_str}_000.fits",
                             f"{base_str}_001.fits", sequence=True)
```
You should be creating files within the `tmp_path` directory.

ayshih (changes requested):

sunpy/map/tests/test_mapsequence.py
jeffreypaul15: Oh yeah, my bad.

jeffreypaul15:
@nabobalis Do we need the py38-online [linux] test to pass? There's a warning which is causing it to fail.

nabobalis:
If a test is now causing a warning due to this PR, that has to be addressed.

jeffreypaul15:
> If a test is now causing a warning due to this PR, that has to be addressed.

The header

> If a test is now causing a warning due to this PR, that has to be addressed.

The header contains a `BLANK` keyword (warns when trying to save) which is removed when creating the pytest fixtures but the doctest seems to be having issues with this keyword while saving. Is something like pytest.warns for doctests?

nabobalis:
Not sure.

jeffreypaul15:
> Not sure.

I'm assuming I can use +SKIP because we're already testing with pytest for the save method.

jeffreypaul15:
I think the PR is clear of test failures now. Test might be failing due to some other reason?

nabobalis:
Yeah those look like unreleated issues.

ayshih (approved):
Looks good.  Must be squash merged

nabobalis (approved):

PR #5148, opened 2021-03-26
Added docstring in class  MDISynopticMap.

Vedant1parekh:
### Description

Fixes #<Issue Number>


nabobalis:
Hi @Vedant1parekh thanks for the PR. 

Technically this does not add a docstring to MDISynopticMap, it already had one. You are just updating it to be more useful. 

I also am not a fan of copy and pasting text from another website without the proper attribution.
I do wonder if we need all that text as well.

nabobalis:
I am going to close this since this is a replica of https://github.com/sunpy/sunpy/pull/5149

PR #5156, opened 2021-03-29
pin importlib_metadata in test deps

nabobalis:
Attempt to bypass the warning issue with py37.

Setting the ignore text seems to not work at all.

https://github.com/python/importlib_metadata/blob/main/CHANGES.rst#v390


dstansby (approved):
If CI passes. The failing azure doc build seems to be due to EVE data not being found so is fine.

nabobalis:
Manually added to the 2.1 branch.

PR #5161, opened 2021-03-30
update figure dev docs and added tox-pypi-filter

nabobalis:
Probably look at the online helio issue while I am here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,7 @@ apt-run: &apt-install
   name: Install apt packages
   command: |
     sudo apt update
-    sudo apt install -y graphviz build-essential libopenjp2-7 python3-dev
+    sudo apt install -y libopenjp2-7 python3-dev
 
 jobs:
   figure:
--- a/docs/dev_guide/contents/tests.rst
+++ b/docs/dev_guide/contents/tests.rst
@@ -259,12 +259,14 @@ Inside this folder will be all the images created, as well as a json file with t
 The current hashes are located within "sunpy/tests/figure_hashes_mpl_<ver>_ft_<ver>_astropy_<ver>.json" and this will be where you will need to update old hashes or create new figure entries if anything changes.
 The filenames are the versions of matplotlib, freetype and astropy used.
 If these versions differ to your local setup, the figure tests will not run.
+In theory Python version does not change the results since we have pinned the packages that can cause the hash to vary.
 
 If you are adding a new figure test you will need to do a few more steps::
 
-    $ tox -e py38-figure -- --mpl-generate-hash-library=sunpy/tests/figure_hashes_mpl_332_ft_261_astropy_403.json
+    $ tox -e py38-figure -- --mpl-generate-hash-library=sunpy/tests/figure_hashes_mpl_332_ft_261_astropy_42.json
 
-and for the development figure tests::
+The filename might have changed since the version of astropy or matplotlib or freetype gets updated.
+For the development figure tests::
 
     $ tox -e py38-figure-devdeps -- --mpl-generate-hash-library=sunpy/tests/figure_hashes_mpl_dev_ft_261_astropy_dev.json
 
--- a/docs/dev_guide/contents/tests.rst
+++ b/docs/dev_guide/contents/tests.rst
@@ -259,13 +259,14 @@ Inside this folder will be all the images created, as well as a json file with t
 The current hashes are located within "sunpy/tests/figure_hashes_mpl_<ver>_ft_<ver>_astropy_<ver>.json" and this will be where you will need to update old hashes or create new figure entries if anything changes.
 The filenames are the versions of matplotlib, freetype and astropy used.
 If these versions differ to your local setup, the figure tests will not run.
-In theory Python version does not change the results since we have pinned the packages that can cause the hash to vary.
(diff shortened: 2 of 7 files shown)


nabobalis:
The tox change and the SSL check disable would need to be backported manually.

sunpy/net/helio/hec.py
```diff
@@ -84,9 +84,9 @@ def __init__(self, link=None):
         if link is None:
             # The default wsdl file
             link = parser.wsdl_retriever()
-        # Disable SSL check.
         session = Session()
-        session.verify = True
+        # Disable SSL check.
+        session.verify = False
```
Cadair, line 89: OK, this has to stop. We can't keep doing this when their servers fall over.

We should make this configurable somehow so we can set it for the test run or something.

nabobalis: Hmm, could make it an env var we set in tox?

Cadair: Probably easiest.

sunpy/net/helio/hec.py
```diff
@@ -85,8 +86,8 @@ def __init__(self, link=None):
             # The default wsdl file
             link = parser.wsdl_retriever()
         session = Session()
-        # Disable SSL check.
-        session.verify = False
+        # This is for use in our test suite.
+        session.verify = int(os.environ.get("HELIO_SSL", 1))
```
nabobalis, line 90: I'm being very thick here but it works.

docs/dev_guide/contents/tests.rst
```diff
@@ -259,12 +259,14 @@ Inside this folder will be all the images created, as well as a json file with t
 The current hashes are located within "sunpy/tests/figure_hashes_mpl_<ver>_ft_<ver>_astropy_<ver>.json" and this will be where you will need to update old hashes or create new figure entries if anything changes.
 The filenames are the versions of matplotlib, freetype and astropy used.
 If these versions differ to your local setup, the figure tests will not run.
+In theory Python version does not change the results since we have pinned the packages that can cause the hash to vary.
```
hayesla, line 262:
```suggestion
In theory, Python version does not change the results since we have pinned the packages that can cause the hash to vary.
```

docs/dev_guide/contents/tests.rst
```diff
@@ -259,12 +259,14 @@ Inside this folder will be all the images created, as well as a json file with t
...
 
 If you are adding a new figure test you will need to do a few more steps::
 
-    $ tox -e py38-figure -- --mpl-generate-hash-library=sunpy/tests/figure_hashes_mpl_332_ft_261_astropy_403.json
+    $ tox -e py38-figure -- --mpl-generate-hash-library=sunpy/tests/figure_hashes_mpl_332_ft_261_astropy_42.json
 
-and for the development figure tests::
+The filename might have changed since the version of astropy or matplotlib or freetype gets updated.
```
hayesla, line 268:
```suggestion
The filename might have changed if the version of astropy or matplotlib or freetype gets updated.
```
does `if` make more sense here - unless I'm missing something.

hayesla (approved):

docs/dev_guide/contents/tests.rst
nabobalis: Yeah if makes more sense.

sunpy/net/helio/hec.py
nabobalis: If I passed a bool in the tox env, it came back as "True" or "False" which is always true. 
So int a 0 or 1 gets me the boolean I need. But it seems a bit silly.

.circleci/config.yml
```diff
@@ -24,7 +24,7 @@ apt-run: &apt-install
   name: Install apt packages
   command: |
     sudo apt update
-    sudo apt install -y graphviz build-essential libopenjp2-7 python3-dev
+    sudo apt install -y libopenjp2-7 python3-dev
```
nabobalis, line 27: We don't need those two installs. "python3-dev" gives us the python 3.7 we need for the figure tests.

sunpy/net/tests/test_helioviewer.py
```diff
@@ -29,119 +29,130 @@ def client():
 
 
 @pytest.mark.remote_data
-class TestHelioviewerClient:
```
nabobalis, line 32: The class had to go.

sunpy/net/tests/test_helioviewer.py
```diff
@@ -29,119 +29,130 @@ def client():
...
+    header2 = client.get_jp2_header('1994/01/01', jp2_id=1795504)
+    assert header1 == header2
+    assert len(header1) == len(header2) == 1
+    assert ('fits' in header1.keys()) and ('fits' in header2.keys())
+
+
+@skip_glymur
+@figure_test
+@pytest.mark.remote_data
+@pytest.mark.filterwarnings("ignore::sunpy.util.SunpyMetadataWarning")
```
nabobalis, line 100: To fix the figure test, I decided to filter the warning over asserting it via pytest.warns.

sunpy/net/helio/hec.py
Cadair: All we really need to be testing for is if this var is set at all, the value is somewhat irrelevant.
```suggestion
        session.verify = bool(os.environ.get("VERIFY_HELIO_SSL", 1))
```

Issue #5165, opened 2021-03-31
Map mutates input WCS


dstansby:
When creating a `Map` using `Map(array, wcs)`, the input `wcs` does not evaluate equal to the resulting `map.wcs` object. See below for an example - this is contrived to be simple, but is an issue when doing reprojection, when one wants to force two maps to have the same `wcs` so you can plot them one on top of the other.

```python
import sunpy.data.sample
import sunpy.map

map_aia = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
map_hmi = sunpy.map.Map(sunpy.data.sample.HMI_LOS_IMAGE)

map_aia = sunpy.map.Map(map_aia.data, map_hmi.wcs)
# This prints False
print(map_aia.wcs == map_hmi.wcs)
```

Cadair:
Correct, the input WCS is converted to a header and then read through the normal map property -> WCS pipe. There almost certainly isn't a fix for this which isn't rewrite most of maps metadata handling ;)

Cadair:
The most general way to assert if two WCS objects are equal is to run a sample of pixel locations through a pixel-world-pixel round trip and see if the output's match.

dstansby:
I think this is probably just a floating point accuracy error since `dsun_obs` is going back and forth through at least one transformation, with a bit more digging I think this should be fixable.

dstansby:
Here's a smaller example, showing the issue:
```python
import sunpy.data.sample
import sunpy.map

map_hmi = sunpy.map.Map(sunpy.data.sample.HMI_LOS_IMAGE)
print("{:.20f}".format(map_hmi.meta['dsun_obs']))
print("{:.20f}".format(map_hmi.wcs.wcs.aux.dsun_obs))
```
This prints:
```
151846023892.98999023437500000000
151846023892.99002075195312500000
```
ie. the two values are slightly different. I think this is because of small changes in a `HGS` > `HGC` transformation that takes place in `GenericMap.observer_coordinate`, which is used to populate the wcs keywords.

ayshih:
> This prints:
> 
> ```
> 151846023892.98999023437500000000
> 151846023892.99002075195312500000
> ```

This slight difference (~30 microns) is the machine epsilon for IEEE-754 64-bit floats (2^-52, or 2.22e-16):

```python
>>> import numpy as np
>>> np.finfo(float).eps
2.220446049250313e-16
[...]

PR #5166, opened 2021-03-31
Don't include CROTA when PC present in WCS

dstansby:
The FITS standard states (https://fits.gsfc.nasa.gov/standard40/fits_standard40aa-le.pdf, section 8.1):

> the CROTA2 must not occur with the PCij keywords

This is part of fixing https://github.com/sunpy/sunpy/issues/5165.


Cadair:
Whooops my bad, this wasn't a problem until we were generating the WCS from the whole header.

Cadair (approved):

nabobalis (approved):

PR #5167, opened 2021-03-31
Only caclulate observer coord for WCS if needed

dstansby:
This is a minor performance improvement, and also avoids small floating point accuracy changes to the metadata due to round trips in the coordinate transformation machinery.

Part of fixing https://github.com/sunpy/sunpy/issues/5165

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -527,23 +527,27 @@ class GenericMap(NDData):
         if w2.wcs.ctype[1].lower() in ("solar-y", "solar_y"):
             w2.wcs.ctype[1] = 'HPLT-TAN'
 
-        # Set observer coordinate information
-        #
-        # Clear all the aux information that was set earlier. This is to avoid
-        # issues with maps that store multiple observer coordinate keywords.
-        # Note that we have to create a new WCS as it's not possible to modify
-        # wcs.wcs.aux in place.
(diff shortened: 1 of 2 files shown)


dstansby:
Hmm, the test failures are interesting, and need more investigation.

sunpy/map/mapbase.py
```diff
@@ -527,23 +527,27 @@ def wcs(self):
...
-            header.pop(kw, None)
-        w2 = astropy.wcs.WCS(header)
-
-        # Get observer coord, and transform if needed
-        obs_coord = self.observer_coordinate
-        if not isinstance(obs_coord.frame, (HeliographicStonyhurst, HeliographicCarrington)):
-            obs_coord = obs_coord.transform_to(HeliographicStonyhurst(obstime=self.date))
-
-        sunpy.coordinates.wcs_utils._set_wcs_aux_obs_coord(w2, obs_coord)
+        if not sunpy.coordinates.wcs_utils._has_complete_observer_info(w2):
```
Cadair, line 530: I don't think this approach is valid. Take AIA for instance, we don't use the heliographic keys for AIA, we read the observer location out of the more accurate keys in a different frame. This would prevent that from happening.

dstansby: Ah, that would explain the test issues.

PR #5168, opened 2021-04-01
Added 2 small missing tests for GenericMap

jeffreypaul15:
### Description

Based on codecov, these `GenericMap's` `_fix_bitpix()` and `_parse_submap_quantity_input()` methods weren't fully tested.


ayshih:
You should avoid directly testing private methods.  They are, after all, not part of the supported API, so significant changes to their behavior can be perfectly acceptable.  Instead, tests should stick to the public API.  For example, you would provide the same variety of inputs to `.submap()`, which would then internally call `._parse_submap_quantity_input()` and get you the same coverage.

jeffreypaul15:
> You should avoid directly testing private methods. They are, after all, not part of the supported API, so significant changes to their behavior can be perfectly acceptable. Instead, tests should stick to the public API. For example, you would provide the same variety of inputs to `.submap()`, which would then internally call `._parse_submap_quantity_input()` and get you the same coverage.

Yeah this makes sense, will do that. Thanks!

nabobalis (approved):

nabobalis:
Thanks again @jeffreypaul15

Issue #5171, opened 2021-04-01
`make_fitswcs_header` doesn't create a `rsun_obs` keyword


hayesla:
### Description
when using `make_fitswcs_header`, the `rsun_obs` keyword is no longer populated - removed in #4620.
This also makes the [`get_observer_meta()` ](https://github.com/sunpy/sunpy/blob/master/sunpy/map/header_helper.py#L199) redundant.

This is really not too much of an issue, but when you create a map from this, and then use `map.draw_limb()` it raises a warning as it calculates the `rsun_obs` assuming photospheric limb. 

However we should probably have it such that `rsun_obs` is populated with `make_fitswcs_header`, maybe to a similar way to  it was done in [`get_observer_meta()` ](https://github.com/sunpy/sunpy/blob/master/sunpy/map/header_helper.py#L199)?

ayshih:
> This is really not too much of an issue, but when you create a map from this, and then use `map.draw_limb()` it raises a warning as it calculates the `rsun_obs` assuming photospheric limb.

#5172 should prevent this warning.

Even so, it might be useful for the helper to create the `rsun_obs` keyword explicitly.

jeffreypaul15:
Would this require writing a helper to populate the `rsun_obs` keyword?

Cadair:
Can this be closed?

PR #5172, opened 2021-04-01
`GenericMap.rsun_obs` now uses `rsun_ref` if it is in the header

ayshih:
Related to #5171 

When the property `GenericMap.rsun_obs` (used by `.draw_limb()`) doesn't find any information for the solar angular radius, it falls back to using the solar physical radius at the observer distance.  However, it doesn't make use of the `rsun_ref` keyword if it is present.  In that case, it doesn't need to raise a warning, because no assumption is being made.


ayshih:
This could be backported to 2.0 as well, although I don't know how many real-world FITS headers do not have `rsun_obs` (or equivalent) but do have `rsun_ref`.  I set the backport to 2.1 because we started creating such headers with 2.1.

nabobalis (approved):

PR #5177, opened 2021-04-02
Adding calc of rsun_obs to make_fitswcs_header

hayesla:
### Description

This comes from #5171  - it adds back in the trig calc for the `rsun_obs` keyword, and removes a redundant function that's no longer used following #4620 

Thoughts on this - I'm in favour, but also could be persuaded. Also is there ever a case that someone would want to pass their own `rsun_obs`?

### DIFF
--- a/changelog/5177.bugfix.rst
+++ b/changelog/5177.bugfix.rst
@@ -1,0 +1,1 @@
+Added "rsun_obs" to the output of `sunpy.map.make_fitswcs_header` if "rsun" is in the coordinate argument.
--- a/changelog/5177.bugfix.rst
+++ b/changelog/5177.bugfix.rst
@@ -1,1 +1,1 @@
-Added "rsun_obs" to the output of `sunpy.map.make_fitswcs_header` if "rsun" is in the coordinate argument.
+Added a "rsun_obs" keyword to the output of :func:`sunpy.map.make_fitswcs_header` if the coordinate argument has a "rsun" frame attribute.
--- a/sunpy/map/header_helper.py
+++ b/sunpy/map/header_helper.py
@@ -161,6 +161,11 @@ def make_fitswcs_header(data, coordinate,
          meta_wcs['PC2_1'], meta_wcs['PC2_2']) = (rotation_matrix[0, 0], rotation_matrix[0, 1],
                                                   rotation_matrix[1, 0], rotation_matrix[1, 1])
 
+    if hasattr(coordinate, 'rsun') and isinstance(coordinate.observer, frames.BaseCoordinateFrame):
+        observer = coordinate.observer.transform_to(frames.HeliographicStonyhurst(obstime=coordinate.observer.obstime))
+        meta_wcs['rsun_obs'] = sun._angular_radius(getattr(coordinate, 'rsun'),
+                                                   observer.radius).to_value(u.arcsec)
+
     meta_dict = MetaDict(meta_wcs)
 
     return meta_dict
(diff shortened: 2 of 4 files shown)


nabobalis:
Do want to backport this?

ayshih:
The restoration of `rsun_obs` can be backported to 2.1, but the removal of `get_observer_meta()` may technically require a deprecation period and can't be backported.

ayshih:
> Also is there ever a case that someone would want to pass their own `rsun_obs`?

Maybe, but it seems too pathological to support that in the helper.  The user can modify the header themselves.

dstansby (changes requested):
Seems 👍 , but `get_observer_meta` should have a deprecation period if it's going to be removed. I think it might be worth keeping a public facing function for this though.

nabobalis:
If we restore `get_observer_meta`, we can backport the change in the other function back to 2.0 as a bugfix?

Then a new Pr to add a deprecation for `get_observer_meta`?

ayshih:
> If we restore `get_observer_meta`, we can backport the change in the other function back to 2.0 as a bugfix?

2.1.  2.0 is fine..

sunpy/map/tests/test_header_helper.py
```diff
@@ -90,16 +88,19 @@ def test_make_fits_header(map_data, hpc_test_header, hgc_test_header,
     assert header['crpix1'] == 5.5
     assert header['ctype1'] == 'HPLN-TAN'
     assert u.allclose(header['dsun_obs'], hpc_test_header.frame.observer.radius.to_value(u.m))
+    assert header.get('rsun_obs') is not None
```
nabobalis, line 91: I don't know how to work out this value without replicating the code in the header function.

nabobalis:
I added a changelog and added a basic check to the test to check if the keyword is in the return.

sunpy/map/header_helper.py
```diff
@@ -161,6 +160,13 @@ def make_fitswcs_header(data, coordinate,
          meta_wcs['PC2_1'], meta_wcs['PC2_2']) = (rotation_matrix[0, 0], rotation_matrix[0, 1],
                                                   rotation_matrix[1, 0], rotation_matrix[1, 1])
 
+    if hasattr(coordinate, 'rsun') and isinstance(coordinate.observer, frames.BaseCoordinateFrame):
+        observer = coordinate.observer.transform_to(
+            frames.HeliographicStonyhurst(obstime=coordinate.observer.obstime)
+        )
```
ayshih, line 166:
```suggestion
```
These lines, as written, should never do anything.  The `coordinate.observer` attribute is already in `HeliographicStonyhurst` at the `coordinate.observer.obstime` obstime.

sunpy/map/tests/test_header_helper.py
ayshih: Well, yeah, you're basically going to have to replicate the call to `sun._angular_radius()`.  I'd do:
```suggestion
    assert u.allclose(header['rsun_ref'] * u.m, hpc_test_header.frame.rsun)
    assert u.allclose(header['rsun_obs'] * u.arcsec,
                      sun._angular_radius(header['rsun_ref'] * u.m, header['dsun_obs'] * u.m))
```

sunpy/map/tests/test_header_helper.py
```diff
@@ -90,16 +88,19 @@ def test_make_fits_header(map_data, hpc_test_header, hgc_test_header,
...
     assert isinstance(WCS(header), WCS)
 
     # Check no observer info for HGS
     header = sunpy.map.make_fitswcs_header(map_data, hgs_test_header)
     assert header.get('dsun_obs') is None
+    assert header.get('rsun_obs') is None
```
ayshih, line 97:
```suggestion
    assert 'dsun_obs' not in header
    assert 'rsun_obs' not in header
```

sunpy/map/tests/test_header_helper.py
```diff
@@ -90,16 +88,19 @@ def test_make_fits_header(map_data, hpc_test_header, hgc_test_header,
...
     assert isinstance(WCS(header), WCS)
 
     # Check for observer info for HGC
     header = sunpy.map.make_fitswcs_header(map_data, hgc_test_header)
     assert u.allclose(header['dsun_obs'], hgc_test_header.frame.observer.radius.to_value(u.m))
+    assert header.get('rsun_obs') is None
```
ayshih, line 103:
```suggestion
    assert 'rsun_obs' not in header
```

sunpy/map/header_helper.py
```diff
@@ -161,6 +160,13 @@ def make_fitswcs_header(data, coordinate,
...
+        meta_wcs['rsun_obs'] = sun._angular_radius(getattr(coordinate, 'rsun'),
```
ayshih, line 167:
```suggestion
        meta_wcs['rsun_obs'] = sun._angular_radius(coordinate.rsun),
```

changelog/5177.bugfix.rst
```diff
@@ -0,0 +1 @@
+Added "rsun_obs" to the output of `sunpy.map.make_fitswcs_header` if "rsun" is in the coordinate argument.
```
ayshih, line 1:
```suggestion
Added a "rsun_obs" keyword to the output of :func:`sunpy.map.make_fitswcs_header` if the coordinate argument has a `rsun` frame attribute.
```

hayesla (commented):
thanks @nabobalis for updating this - looks good to me!

dstansby (approved):

nabobalis (approved):

PR #5183, opened 2021-04-05
fix online test

nabobalis:


sunpy/net/dataretriever/sources/tests/test_noaa.py
```diff
@@ -173,19 +173,10 @@ def test_srs_tar_unpack_midyear():
...
 
 
-@no_vso
-@pytest.mark.remote_data
-@mock.patch('ftplib.FTP.nlst', side_effect=[[''], ['20200101SRS.txt', '20200102SRS.txt']])
-def test_srs_missing_tarball(mock_ftp_nlst):
-    qr = Fido.search(a.Time('2020-01-01', '2020-01-02'), a.Instrument.srs_table)
-    urls = [qrblock['url'] for qrblock in qr[0]]
-    assert urls[0].endswith('20200101SRS.txt')
-    assert urls[1].endswith('20200102SRS.txt')
```
nabobalis, line 183: This test was failing and we don't need it anyway.

wtbarnes: Why don't we need this anymore?

nabobalis: This does the same thing as the "test_srs_current_year" but broken now since they removed the text files for 2010.

dstansby (approved):

PR #5184, opened 2021-04-05
Updated np.linspace for both generalized axis

satyamroy001:
Updated np.linspace for both generalized axis (X,Y)

### Description

Fixes #<Issue Number>

### DIFF
--- a/examples/example_template/example_template.py
+++ b/examples/example_template/example_template.py
@@ -30,7 +30,8 @@ import numpy as np
 # inlined figure.
 
 x = np.linspace(-np.pi, np.pi, 300)
-xx, yy = np.meshgrid(x, x)
+y= np.linspace(-np.pi, np.pi, 300)
+xx, yy = np.meshgrid(x, y)
 z = np.cos(xx) + np.cos(yy)
 
 plt.figure()


examples/example_template/example_template.py
```diff
@@ -30,7 +30,8 @@
 # inlined figure.
 
 x = np.linspace(-np.pi, np.pi, 300)
-xx, yy = np.meshgrid(x, x)
+y= np.linspace(-np.pi, np.pi, 300)
+xx, yy = np.meshgrid(x, y)
```
nabobalis, line 34: Can I ask why we need this change?

satyamroy001: Yes, as linspace can be coined as Lin i.e linear + space. We are here plotting 2D plot as an example, so it's useful and more generalized way to put array on X,Y.

nabobalis: I mean that is true. However, since the values are the same for x and y, I don't see the need for this change.

nabobalis:
Many thanks for the PR @satyamroy001 unfortunately I can not see the need to merge this. This example is not rendered in our gallery as it is a placeholder template for people to use when they create their own examples.

I would strongly suggest looking at https://github.com/sunpy/sunpy/labels/Good%20First%20Issue if you want to work on some issues for sunpy!

PR #5193, opened 2021-04-07
Adds support for file-handlers and file-type in Map and TimeSeries

### LINKED ISSUE #2690: Should sunpy.map.Map be able to take file handlers and a filetype like Timeseries

devanshshukla99:
### Description

Fixes #2690

Adds support for file-object and file-type in `sunpy.map.Map` & `sunpy.timeseries.TimeSeries`.

Currently, the support for `sunpy.timeseries.TimeSeries` is broken

### DIFF
--- a/changelog/5193.feature.rst
+++ b/changelog/5193.feature.rst
@@ -1,0 +1,1 @@
+Added support for file-object and file-type in `~sunpy.map.Map` and `~sunpy.timeseries.TimeSeries`.
--- a/sunpy/io/ana.py
+++ b/sunpy/io/ana.py
@@ -10,10 +10,8 @@ This is a modified version of `pyana <https://github.com/tvwerkhoven/pyana>`__.
 import io
 import os
 import collections
-from warnings import warn
 
 from sunpy.io.header import FileHeader
-from sunpy.util.exceptions import SunpyUserWarning
 
 try:
     from sunpy.io import _pyana
@@ -48,8 +46,7 @@ def read(filename, debug=False, **kwargs):
     >>> data = sunpy.io.ana.read(filename)  # doctest: +SKIP
     """
     if isinstance(filename, io.IOBase):
-        warn("Reader does not support file-object, falling back to using filepath", SunpyUserWarning)
-        filename = filename.name
+        raise TypeError("Reader does not support file-handler")
 
     if not os.path.isfile(filename):
         raise OSError("File does not exist!")
--- a/sunpy/io/ana.py
+++ b/sunpy/io/ana.py
@@ -7,6 +7,7 @@ This is a modified version of `pyana <https://github.com/tvwerkhoven/pyana>`__.
 
     The reading and writing of ana files is not supported under Windows.
 """
+import io
 import os
 import collections
 
@@ -49,6 +50,10 @@ def read(filename, debug=False, **kwargs):
 
     if _pyana is None:
         raise ImportError("C extension for ANA is missing, please rebuild.")
+
+    # NOTE: This can be removed after adding support for file-obj in `sunpy.io._pyana`.
+    if isinstance(filename, io.IOBase):
+        filename = filename.name  # Extracting path from the file-obj
 
     data = _pyana.fzread(filename, debug)
     return [HDPair(data['data'], FileHeader(data['header']))]
--- a/sunpy/io/ana.py
+++ b/sunpy/io/ana.py
@@ -10,8 +10,10 @@ This is a modified version of `pyana <https://github.com/tvwerkhoven/pyana>`__.
 import io
 import os
 import collections
+from warnings import warn
 
 from sunpy.io.header import FileHeader
+from sunpy.util.exceptions import SunpyUserWarning
 
 try:
     from sunpy.io import _pyana
@@ -45,15 +47,15 @@ def read(filename, debug=False, **kwargs):
     --------
     >>> data = sunpy.io.ana.read(filename)  # doctest: +SKIP
     """
+    if isinstance(filename, io.IOBase):
+        warn("Reader does not support file-object, falling back to using filepath", SunpyUserWarning)
+        filename = filename.name
+
     if not os.path.isfile(filename):
         raise OSError("File does not exist!")
 
     if _pyana is None:
         raise ImportError("C extension for ANA is missing, please rebuild.")
-
-    # NOTE: This can be removed after adding support for file-obj in `sunpy.io._pyana`.
-    if isinstance(filename, io.IOBase):
-        filename = filename.name  # Extracting path from the file-obj
 
     data = _pyana.fzread(filename, debug)
     return [HDPair(data['data'], FileHeader(data['header']))]
--- a/sunpy/io/file_tools.py
+++ b/sunpy/io/file_tools.py
@@ -1,6 +1,7 @@
 """
 This module provides a generic file reader.
 """
+import io
 import os
 import re
 import pathlib
@@ -77,19 +78,24 @@ def read_file(filepath, filetype=None, **kwargs):
     -----
     Other keyword arguments are passed to the reader used.
     """
-    filepath = str(pathlib.Path(filepath))
+    if isinstance(filepath, io.IOBase):
+        _filepath = pathlib.Path(filepath.name)
+
+    if not isinstance(filepath, pathlib.Path):
+        _filepath = pathlib.Path(_filepath)
+
     # Use the explicitly passed filetype
     if filetype is not None:
-        return _readers[filetype].read(filepath, **kwargs)
+        return _readers[filetype].read(_filepath, **kwargs)
 
     # Go through the known extensions
     for extension, readername in _known_extensions.items():
-        if filepath.endswith(extension) or filetype in extension:
-            return _readers[readername].read(filepath, **kwargs)
+        if _filepath.suffix.endswith(extension) or filetype in extension:
+            return _readers[readername].read(_filepath, **kwargs)
 
     # If filetype is not apparent from the extension, attempt to detect it
-    readername = _detect_filetype(filepath)
-    return _readers[readername].read(filepath, **kwargs)
+    readername = _detect_filetype(_filepath)
+    return _readers[readername].read(_filepath, **kwargs)
 
 
 def read_file_header(filepath, filetype=None, **kwargs):
--- a/sunpy/io/file_tools.py
+++ b/sunpy/io/file_tools.py
@@ -5,6 +5,7 @@ import io
 import os
 import re
 import pathlib
+from itertools import chain
 
 try:
     from . import fits
@@ -60,7 +61,7 @@ def read_file(filepath, filetype=None, **kwargs):
 
     Parameters
     ----------
-    filepath : path-like
+    filepath : `str` or file-like
         The file to be read.
     filetype : `str`, optional
         Supported reader or extension to manually specify the filetype.
@@ -78,20 +79,24 @@ def read_file(filepath, filetype=None, **kwargs):
     -----
     Other keyword arguments are passed to the reader used.
     """
+    if not isinstance(filepath, (str, io.IOBase, pathlib.Path)):
+        raise TypeError("expected str, IO or pathlib.Path object")
+
     if isinstance(filepath, io.IOBase):
         _filepath = pathlib.Path(filepath.name)
-
-    if not isinstance(filepath, pathlib.Path):
-        _filepath = pathlib.Path(_filepath)
+    else:
+        _filepath = pathlib.Path(filepath)
 
     # Use the explicitly passed filetype
     if filetype is not None:
-        return _readers[filetype].read(_filepath, **kwargs)
+        if filetype in chain(*_known_extensions):
+            return _readers[filetype].read(filepath, **kwargs)
+        raise UnrecognizedFileTypeError("The requested filetype is not currently supported by SunPy.")
 
     # Go through the known extensions
     for extension, readername in _known_extensions.items():
-        if _filepath.suffix.endswith(extension) or filetype in extension:
-            return _readers[readername].read(_filepath, **kwargs)
+        if _filepath.suffix.endswith(extension):
+            return _readers[readername].read(filepath, **kwargs)
 
     # If filetype is not apparent from the extension, attempt to detect it
     readername = _detect_filetype(_filepath)
--- a/sunpy/io/file_tools.py
+++ b/sunpy/io/file_tools.py
@@ -79,19 +79,15 @@ def read_file(filepath, filetype=None, **kwargs):
     -----
     Other keyword arguments are passed to the reader used.
     """
-    if not isinstance(filepath, (str, io.IOBase, pathlib.Path)):
-        raise TypeError("expected str, IO or pathlib.Path object")
-
-    if isinstance(filepath, io.IOBase):
-        _filepath = pathlib.Path(filepath.name)
-    else:
-        _filepath = pathlib.Path(filepath)
+    _filepath = filepath.name if isinstance(filepath, io.IOBase) else filepath
+    _filepath = pathlib.Path(_filepath)
 
     # Use the explicitly passed filetype
-    if filetype is not None:
-        if filetype in chain(*_known_extensions):
-            return _readers[filetype].read(filepath, **kwargs)
-        raise UnrecognizedFileTypeError("The requested filetype is not currently supported by SunPy.")
+    if filetype:
+        if filetype not in chain(*_known_extensions):
+            raise UnrecognizedFileTypeError(
+                "The requested filetype is not currently supported by SunPy.")
+        return _readers[filetype].read(filepath, **kwargs)
 
     # Go through the known extensions
     for extension, readername in _known_extensions.items():
--- a/sunpy/io/jp2.py
+++ b/sunpy/io/jp2.py
@@ -3,11 +3,9 @@ This module provides a JPEG 2000 file reader.
 """
 import io
 import collections
-from warnings import warn
 from xml.etree import cElementTree as ET
 
 from sunpy.io.header import FileHeader
-from sunpy.util.exceptions import SunpyUserWarning
 from sunpy.util.xml import xml_to_dict
 
 __all__ = ['read', 'get_header', 'write']
@@ -30,8 +28,7 @@ def read(filepath, **kwargs):
         A list of (data, header) tuples.
     """
     if isinstance(filepath, io.IOBase):
-        warn("Reader does not support file-object, falling back to using filepath", SunpyUserWarning)
-        filepath = filepath.name
+        raise TypeError("Reader does not support file-handler")
 
     # Put import here to speed up sunpy.io import time
     from glymur import Jp2k
--- a/sunpy/io/jp2.py
+++ b/sunpy/io/jp2.py
@@ -1,6 +1,7 @@
 """
 This module provides a JPEG 2000 file reader.
 """
+import io
 import collections
 from xml.etree import cElementTree as ET
 
@@ -26,6 +27,10 @@ def read(filepath, **kwargs):
     pairs : `list`
         A list of (data, header) tuples.
     """
+    # NOTE: This can be removed after support for file-obj in `glymur.Jp2k`.
+    if isinstance(filepath, io.IOBase):
+        filepath = filepath.name  # Extracting path from the file-obj
+
     # Put import here to speed up sunpy.io import time
     from glymur import Jp2k
     header = get_header(filepath)
--- a/sunpy/io/jp2.py
+++ b/sunpy/io/jp2.py
@@ -3,9 +3,11 @@ This module provides a JPEG 2000 file reader.
 """
 import io
 import collections
+from warnings import warn
 from xml.etree import cElementTree as ET
 
 from sunpy.io.header import FileHeader
+from sunpy.util.exceptions import SunpyUserWarning
 from sunpy.util.xml import xml_to_dict
 
 __all__ = ['read', 'get_header', 'write']
@@ -27,9 +29,9 @@ def read(filepath, **kwargs):
     pairs : `list`
         A list of (data, header) tuples.
     """
-    # NOTE: This can be removed after support for file-obj in `glymur.Jp2k`.
     if isinstance(filepath, io.IOBase):
-        filepath = filepath.name  # Extracting path from the file-obj
+        warn("Reader does not support file-object, falling back to using filepath", SunpyUserWarning)
+        filepath = filepath.name
 
     # Put import here to speed up sunpy.io import time
     from glymur import Jp2k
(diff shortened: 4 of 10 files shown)


changelog/5193.doc.rst
```diff
@@ -0,0 +1 @@
+Added documentation strings related to file-type & file-obj in :class:`sunpy.map.Map` and :class:`sunpy.timeseries.TimeSeries`.
```
nabobalis, line 1: Remove this file.

changelog/5193.trivial.rst
```diff
@@ -0,0 +1 @@
+Added file-object & file-type related tests for :class:`sunpy.map.Map` and :class:`sunpy.timeseries.TimeSeries`.
```
nabobalis, line 1: Remove this file.

changelog/5193.feature.rst
```diff
@@ -0,0 +1 @@
+Added support for file-object and file-type in :class:`~sunpy.map.Map` and :class:`~sunpy.timeseries.TimeSeries`.
```
nabobalis, line 1:
```suggestion
Added support for file-object and file-type in `~sunpy.map.Map` and `~sunpy.timeseries.TimeSeries`.
```

sunpy/io/ana.py
```diff
@@ -50,6 +51,10 @@ def read(filename, debug=False, **kwargs):
     if _pyana is None:
         raise ImportError("C extension for ANA is missing, please rebuild.")
 
+    # NOTE: This can be removed after adding support for file-obj in `sunpy.io._pyana`.
```
nabobalis, line 54:
```suggestion
```
I don't think this will ever happen.

sunpy/io/ana.py
```diff
@@ -50,6 +51,10 @@ def read(filename, debug=False, **kwargs):
...
+    if isinstance(filename, io.IOBase):
+        filename = filename.name  # Extracting path from the file-obj
```
nabobalis, line 56:
```suggestion
        filename = filename.name
```

sunpy/io/file_tools.py
```diff
@@ -77,18 +78,28 @@ def read_file(filepath, filetype=None, **kwargs):
     -----
     Other keyword arguments are passed to the reader used.
     """
+    _filepath = filename
+    # Type checking and extracting filepath.
```
nabobalis, line 82:
```suggestion
```

sunpy/io/file_tools.py
```diff
@@ -53,13 +54,13 @@ def __getitem__(self, key):
...
 
-def read_file(filepath, filetype=None, **kwargs):
+def read_file(filename, filetype=None, **kwargs):
     """
     Automatically determine the filetype and read the file.
 
     Parameters
     ----------
-    filepath : `str`
+    filename : `str` or file-like
```
nabobalis, line 63: Is this still not a filepath?

sunpy/io/file_tools.py
```diff
@@ -77,18 +78,28 @@ def read_file(filepath, filetype=None, **kwargs):
...
     """
+    _filepath = filename
+    # Type checking and extracting filepath.
+    if isinstance(filename, (str, io.IOBase, pathlib.Path)):
+        if isinstance(filename, io.IOBase):
+            _filepath = pathlib.Path(filename.name)
+        elif isinstance(filename, str):
+            _filepath = pathlib.Path(filename)
+    else:
+        raise TypeError("filename should be str, IO or Path Object")
```
nabobalis, line 89: This feels like it could be reworked to be a bit neater.

sunpy/io/jp2.py
```diff
@@ -26,6 +27,10 @@ def read(filepath, **kwargs):
     pairs : `list`
         A list of (data, header) tuples.
     """
+    # NOTE: This can be removed after support for file-obj in `glymur.Jp2k`.
```
nabobalis, line 30:
```suggestion
```
This probably won't happen.

sunpy/io/jp2.py
```diff
@@ -26,6 +27,10 @@ def read(filepath, **kwargs):
...
+    if isinstance(filepath, io.IOBase):
+        filepath = filepath.name  # Extracting path from the file-obj
```
nabobalis, line 32:
```suggestion
        filepath = filepath.name
```

nabobalis: Is this tested in some form?

sunpy/io/ana.py
nabobalis: Is this tested in some form?

sunpy/map/map_factory.py
```diff
@@ -54,6 +55,10 @@ class MapFactory(BasicRegistrationFactory):
         Inputs to parse for map objects. See the examples section for a
         detailed list of accepted inputs.
 
+    filetype : `str`, optional
```
nabobalis, line 58: I feel like this should be the last one in the list of keywords.

sunpy/map/map_factory.py
```diff
@@ -146,20 +156,23 @@ class MapFactory(BasicRegistrationFactory):
 
     * Any mixture of the above not in a list
 
-    >>> mymap = sunpy.map.Map(((data, header), data2, header2, 'file1.fits', url_str, 'eit_*.fits'))  # doctest: +SKIP
+    >>> mymap = sunpy.map.Map(((data, header), data2, header2, 'file1.fits', url_str, 'eit_*.fits'))\
+        # doctest: +SKIP
```
nabobalis, line 160:
```suggestion
    >>> mymap = sunpy.map.Map(((data, header), data2, header2, 'file1.fits', url_str, 'eit_*.fits')) # doctest: +SKIP
```

sunpy/map/map_factory.py
```diff
@@ -146,20 +156,23 @@ class MapFactory(BasicRegistrationFactory):
...
+
+    * Manually Specify the file-type
```
nabobalis, line 162:
```suggestion
    * Manually specify the file-type
```

sunpy/map/map_factory.py
```diff
@@ -190,7 +203,7 @@ def _validate_meta(self, meta):
         else:
             return False
 
-    def _parse_args(self, *args, silence_errors=False, **kwargs):
+    def _parse_args(self, *args, filetype=None, silence_errors=False, **kwargs):
```
nabobalis, line 206: Does this need to be passed in? Would it not be covered by "**kwargs"?

sunpy/timeseries/tests/test_timeseries_factory.py
```diff
@@ -148,6 +148,76 @@ def test_implicit_esp(self):
         ts_esp = sunpy.timeseries.TimeSeries(esp_filepath)
         assert isinstance(ts_esp, sunpy.timeseries.sources.eve.ESPTimeSeries)
 
+    # Test with File-Type
```
nabobalis, line 151: I am not sure we need to repeat this test for each source type.

sunpy/timeseries/timeseries_factory.py
```diff
@@ -118,10 +128,14 @@ class TimeSeriesFactory(BasicRegistrationFactory):
 
     >>> my_timeseries = sunpy.timeseries.TimeSeries((data, header), data2, header2,
     ...                                             'file1.fits', url, 'eit_*.fits')  # doctest: +SKIP
+
+    * Manually Specify the file-type
```
nabobalis, line 132:
```suggestion
    * Manually specify the file-type
```

sunpy/map/map_factory.py
devanshshukla99: Yes, I thought explicit keyword arg will be better for readability and further developments, will do the change👍

nabobalis: That is true, but if I am honest, I dislike the entire concept of the "filetype" keyword in our IO module.

sunpy/io/ana.py
devanshshukla99: Not currently, should it be?

nabobalis: It might be tricky with ana but I am not pro-adding this without making sure it works in some form.

sunpy/map/map_factory.py
devanshshukla99: I can understand, I'll do the needful changes, 
and should I close this and open a new PR, to avoid a commit mess?

sunpy/io/file_tools.py
devanshshukla99: It was just a rename, to distinguish between the arg and the path extracted from IO object

devanshshukla99:
@nabobalis I can't find why test `py37-figure-devdeps` is failing even though tox is reporting no bug, could you please help me regarding this?

nabobalis:
The hash has changed. You can ignore it.

sunpy/io/ana.py
```diff
@@ -44,6 +45,9 @@ def read(filename, debug=False, **kwargs):
     --------
     >>> data = sunpy.io.ana.read(filename)  # doctest: +SKIP
     """
+    if isinstance(filename, io.IOBase):
+        filename = filename.name
```
nabobalis, line 49: If a reader can't handle this, then we should not support it.

sunpy/map/map_factory.py
```diff
@@ -60,6 +61,10 @@ class MapFactory(BasicRegistrationFactory):
     composite : `bool`, optional
         Return a `sunpy.map.CompositeMap` object comprised of all the parsed maps.
 
+    filetype : `str`, optional
+        Manually specify the filetype.
+        Supported extensions are ('fts', 'fits', 'jp2', 'j2k', 'jpc', 'jpt', 'fz', 'f0')
```
nabobalis, line 66: I would rather avoid adding this keyword if we can avoid it.

sunpy/io/ana.py
devanshshukla99: How about raising a warning like,
`warning: reader doesn't support file-obj, falling back to file path.`(or something similar), rather than a exception?

nabobalis: I am not sure, maybe a warning is enough.

sunpy/map/map_factory.py
devanshshukla99: That's doable,

But, suppose a user has a file with an unknown type,
- they can either read the header and supply the filetype directly, bypassing the whole detection stack
- or they run the detection once and supply the filetype in the future,

assuming the user doesn't wanna rename the file;
So wouldn't the filetype keyword be valuable in this context?

nabobalis: I feel like a user should know what type their file is.

devanshshukla99: That's fair

sunpy/map/mapbase.py
```diff
@@ -178,7 +178,7 @@ def __init_subclass__(cls, **kwargs):
         if hasattr(cls, 'is_datasource_for'):
             cls._registry[cls] = cls.is_datasource_for
 
-    def __init__(self, data, header, plot_settings=None, **kwargs):
+    def __init__(self, data, header, plot_settings=None, mask=None, **kwargs):
```
nabobalis, line 181: Was mask missing and not handled by kwargs?

sunpy/io/file_tools.py
```diff
@@ -77,18 +79,28 @@ def read_file(filepath, filetype=None, **kwargs):
     -----
     Other keyword arguments are passed to the reader used.
     """
+    if not isinstance(filepath, (str, io.IOBase, pathlib.Path)):
+        raise TypeError("expected str, IO or pathlib.Path object")
+
+    if isinstance(filepath, io.IOBase):
+        _filepath = pathlib.Path(filepath.name)
```
nabobalis, line 86: Is there a reason for the underscore?

sunpy/map/mapbase.py
devanshshukla99: The parent class `NDData` doesn't support kwargs, so I made `mask` an  explicit arg and removed kwargs from super().
https://github.com/sunpy/sunpy/blob/0f7946d90dae9eeb6ae8e81913a32e0745ee7b44/sunpy/map/mapbase.py#L196

I think there's a test, but since `mask` is an explicit arg in `nddata` it didn't raise any exceptions.

nabobalis: So the test we have didn't work?

devanshshukla99: the test passes.
But an exception is raised if we pass any kwarg which is not in `nddata`.
for example,
```python
sunpy.map.Map(fname, filetype='fits')
```
would raise an exception.

nabobalis: Oh I see.

sunpy/io/file_tools.py
devanshshukla99: feel free to correct me, but doesn't the detection logic requires str filepath?
https://github.com/sunpy/sunpy/blob/0f7946d90dae9eeb6ae8e81913a32e0745ee7b44/sunpy/io/file_tools.py#L102-L103

devanshshukla99: on looking closely, I feel like making the distinction there is a better way; we'll have to extract filetype anyway for,
https://github.com/sunpy/sunpy/blob/0f7946d90dae9eeb6ae8e81913a32e0745ee7b44/sunpy/io/file_tools.py#L97-L99
plus it appears unnecessary to add compatibiliy/support for file-handlers in the detection logic.


ps - looks like there are no tests for the detection function, shall I add one? https://github.com/sunpy/sunpy/blob/0f7946d90dae9eeb6ae8e81913a32e0745ee7b44/sunpy/io/file_tools.py#L196

nabobalis: Sure add a test for `detect_filetype`. 

I am not sure I follow on making the distinction, what do you want to change?

sunpy/io/ana.py
```diff
@@ -44,6 +47,10 @@ def read(filename, debug=False, **kwargs):
     --------
     >>> data = sunpy.io.ana.read(filename)  # doctest: +SKIP
     """
+    if isinstance(filename, io.IOBase):
+        warn("Reader does not support file-object, falling back to using filepath", SunpyUserWarning)
+        filename = filename.name
```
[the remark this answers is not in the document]
nabobalis: I think this should error, can you make those changes?

sunpy/io/file_tools.py
devanshshukla99: I think the detection works by reading some lines from the file, so its probably better to have a separate file obj in the detection function(current implementation) to avoid unnecessary bugs and ResourceWarnings. (probably bugs like unintentionally closing original file-obj, etc)

nabobalis: Ok.

nabobalis:
> > Sure add a test for `detect_filetype`.
> 
> @nabobalis is there any `hdf5` test file in sunpy?

https://github.com/sunpy/sunpy/blob/main/sunpy/data/test/goes_truncated_test_goes15.nc
and 
https://github.com/sunpy/sunpy/blob/main/sunpy/data/test/goes_truncated_test_goes16.nc
should work

sunpy/timeseries/timeseries_factory.py
```diff
@@ -537,6 +557,9 @@ def _apply_result(data_header_pairs, filepaths, result):
     if read:
         data_header_pairs.append(result)
     else:
+        if isinstance(result, io.IOBase):
+            warn("Falling back to using filepath", SunpyUserWarning)
+            result = result.name
```
nabobalis, line 562: This should error.

nabobalis:
So just to double check, we want any places where the code cannot use an open file handler to error instead of use the filename and then open the file.

ayshih:
As discussed during the dev meeting, it'd be good to support (and test!) file-like objects that do not actually correspond to a real on-disk file.  For example, `socket.socket.makefile()` returns a file-like object for a socket.

devanshshukla99:
> As discussed during the dev meeting, it'd be good to support (and test!) file-like objects that do not actually correspond to a real on-disk file. For example, `socket.socket.makefile()` returns a file-like object for a socket.

So, if I understand correctly, we want to support socket file-obj in `sunpy.io.read_file` for reading remote files?

If so, don't we'll still have to download them locally for processing considering socket file-obj aren't seekable?

sunpy/timeseries/timeseries_factory.py
devanshshukla99: 👍
is this error prompt appropiate?
```python
        if isinstance(result, io.IOBase):
            raise TypeError("file-handler not supported for unreadable files")
```

nabobalis: Maybe a bit more info on what reader?

devanshshukla99: uhh, I don't think its specifically attached to any reader; its triggered when the filetype is not supported(in the detection logic for timeseries)

sunpy/map/tests/test_map_factory.py
```diff
@@ -98,6 +102,11 @@ def test_patterns(self):
         assert isinstance(maps, list)
         assert ([isinstance(amap, sunpy.map.GenericMap) for amap in maps])
 
+        # File-like object
+        with open(a_fname, "rb") as fd:
```
nabobalis, line 106: What happens if you pass random binary data to map?

sunpy/io/file_tools.py
```diff
@@ -59,7 +61,7 @@ def read_file(filepath, filetype=None, **kwargs):
 
     Parameters
     ----------
-    filepath : path-like
+    filepath : `str` or file-like
```
nabobalis, line 64: Does a pathlib.Path not work?

sunpy/map/tests/test_map_factory.py
devanshshukla99: it returns a `ValueError`

sunpy/io/file_tools.py
devanshshukla99: no it works,
will update👍

sunpy/map/tests/test_map_factory.py
nabobalis: Is there a test for this in this file?

devanshshukla99: Not currently, should I add one?

nabobalis: If you could.

nabobalis:
> As discussed during the dev meeting, it'd be good to support (and test!) file-like objects that do not actually correspond to a real on-disk file. For example, `socket.socket.makefile()` returns a file-like object for a socket.

Any work on this? Or we could open an issue to support this in future?

devanshshukla99:
> > As discussed during the dev meeting, it'd be good to support (and test!) file-like objects that do not actually correspond to a real on-disk file. For example, `socket.socket.makefile()` returns a file-like object for a socket.
> 
> Any work on this? Or we could open an issue to support this in future?

Nothing till now.

So, if I understand correctly, we'll have to read the socket fileobj then pass the data to appropriate readers?
(considering the file-object returned by socket is not seekable)

nabobalis:
> So, if I understand correctly, we'll have to read the socket fileobj then pass the data to appropriate readers?
> (considering the file-object returned by socket is not seekable)

Sure, his use case was makefile which returns a file-like handler.

devanshshukla99:
> > So, if I understand correctly, we'll have to read the socket fileobj then pass the data to appropriate readers?
> > (considering the file-object returned by socket is not seekable)
> 
> Sure, his use case was makefile which returns a file-like handler.

Looks doable!
Will do👍

nabobalis:
I think we can bump the socket request from Albert an issue and leave this PR as it for review.

@devanshshukla99, could you rebase this PR for us?

devanshshukla99:
> Hi @devanshshukla99, sorry for the lack of reviews on this PR.
> 
> I am trying to get someone to have a look but could you also rebase this if you have time? If not, I can do so.

Apologies from my side too
Will do 👍🏻

sunpy/io/ana.py
```diff
@@ -44,6 +45,9 @@ def read(filename, debug=False, **kwargs):
     --------
     >>> data = sunpy.io.ana.read(filename)  # doctest: +SKIP
     """
+    if isinstance(filename, io.IOBase):
+        raise TypeError("Reader does not support file-handler")
```
Cadair, line 49:
```suggestion
        raise TypeError("The ANA reader does not support file handles.")
```

sunpy/io/jp2.py
```diff
@@ -26,6 +27,9 @@ def read(filepath, **kwargs):
     pairs : `list`
         A list of (data, header) tuples.
     """
+    if isinstance(filepath, io.IOBase):
+        raise TypeError("Reader does not support file-handler")
```
Cadair, line 31:
```suggestion
        raise TypeError("The ANA reader does not support file handles.")
```

sunpy/io/file_tools.py
```diff
@@ -77,19 +79,24 @@ def read_file(filepath, filetype=None, **kwargs):
     -----
     Other keyword arguments are passed to the reader used.
     """
-    filepath = str(pathlib.Path(filepath))
+    _filepath = filepath.name if isinstance(filepath, io.IOBase) else filepath
```
Cadair, line 82: Not all objects which implement `io.IOBase` have a `.name` attribute:

```python
>>> import io
>>> isinstance(io.BytesIO(), io.IOBase)
True
>>> io.BytesIO().name
AttributeError: '_io.BytesIO' object has no attribute 'name'
```

I don't think we should do this at all when dealing with io objects, I think it would be better to work out the type from the magic bytes at the start of the file and not any reference to the filename. (Which already happens if this filename doesn't match anything I think).

sunpy/io/tests/test_filetools.py
```diff
@@ -44,6 +44,15 @@ def test_read_file_fits(self):
         assert all([isinstance(p[1],
                                sunpy.io.header.FileHeader) for p in pairs])
 
+        # Test file-object
```
Cadair, line 47: We should add a test here to check it works as expected with other file objects, something like this:

```python
with open(AIA_171_IMAGE, 'rb') as fd:
    fits_bytes = fd.read()

sunpy.io.read_file(io.BytesIO(fits_bytes))
...
```

sunpy/map/tests/test_map_factory.py
```diff
@@ -80,6 +80,10 @@ def test_patterns(self):
         eitmap = sunpy.map.Map(a_fname)
         assert isinstance(eitmap, sunpy.map.GenericMap)
 
+        # File name with file-type
+        eitmap = sunpy.map.Map(a_fname, filetype=pathlib.Path(a_fname).suffix[1:])
+        assert isinstance(eitmap, sunpy.map.GenericMap)
```
Cadair, line 85: You aren't really testing this properly here as you are giving it a filename which it can detect the type for. You would be better off giving it a random filename and the type.

(Although it will also use magic bytes to detect file type if the filename doesn't pan out so I am not sure exactly how to make this test fail.)

Cadair (changes requested):
Thanks for the work on this @devanshshukla99, sorry it took so long for me to review it. I think it's close, I just think we need to remove the last bit of name checking stuff for file handlers.

PR #5195, opened 2021-04-08
Adding Python 2 support to sunpy

nabobalis:
No more typing in sunpy. Why would we want good things in sunpy? NEVER AGAIN - I SAY IT'S TIME WE GO BACK TO PYTHON 2


Cadair (approved):

Cadair:
(for the benefit of the spectators) We currently don't have any type hints in sunpy, I think that adding them should be a decision we take consciously having thought about it.

PR #5197, opened 2021-04-09
update pre-commit to check for all large files

nabobalis:
I had to increase the "maxkb=1256" since we have several files that are 500 kb to 1000 kb in size.

Files that fail:
```
sunpy/data/test/cor1_20090615_000500_s4c1A.fts (535 KB) exceeds 500 KB.
sunpy/data/test/HinodeXRT.fits (527 KB) exceeds 500 KB.
sunpy/data/test/test_ana.fz (846 KB) exceeds 500 KB.
sunpy/data/test/go1520110607.fits (678 KB) exceeds 500 KB.
sunpy/data/test/2013_06_24__17_31_30_84__SDO_AIA_AIA_193.jp2 (1053 KB) exceeds 500 KB.
sunpy/data/test/annotation_science.db (540 KB) exceeds 500 KB.
sunpy/data/test/waveunit/svsm_e3100_S2_20110625_1856.fts (653 KB) exceeds 500 KB.
```

TODO: Prune the docs from the tarball.
TODO: LOOK AT HEADER ONLY TEST DATA AND WRITE FILES.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -87,6 +87,7 @@ repos:
       - id: check-yaml
       - id: debug-statements
       - id: check-added-large-files
+        args: ['--enforce-all','--maxkb=100']
       - id: end-of-file-fixer
         exclude: ".*(.fits|.fts|.fit|.txt|tca.*)$"
       - id: mixed-line-ending


ayshih:
This is silly, but can we use something other than 1256?  Of course, the choice is arbitrary, but 1256 just seems _strangely_ arbitrary.  Given that one file is above 1024 KB, so we can't set it at 1024, the following are alternatives that I'd each prefer over 1256:

- 1234
- 1100
- 1152 (=1024+128)
- 1280 (=1024+256)
- 1250
- 1500
- 1080 (=360*3)

nabobalis:
I picked the number that would trigger the warning :P

ayshih:
> I picked the number that would not trigger the warning :P

Given that the largest failing file in your printout was 1053 KB, I don't know how you landed on 1256 KB as the limit

nabobalis:
> > I picked the number that would not trigger the warning :P
> 
> Given that the largest failing file in your printout was 1053 KB, I don't know how you landed on 1256 KB as the limit

It's a nicer number. 

I plan to set it to 100kb and redo the test files.

nabobalis:
Now set to 100kb but will fail until the PR that fixes the test data is opened and merged.

nabobalis:
Will need #5261 before we can update this PR.

PR #5199, opened 2021-04-09
Don't modify the header in the constructor of `GenericMap` subclasses

Cadair:
I am not entirely sure how I let this one get away from me this much, but here goes.

The objective of the map sources system is to provide a standard API to access the metadata from a disparate set of headers. This is done via the use of properties which abstract where a bit of information comes from so other places in map can use it (i.e. WCS).

We seem to have forgotten this however, and have widely (for a large number of years) taken to editing the header directly in the constructor of the map source. This PR moves all header modification (with a couple of exceptions) into properties and leaves the actual headers the user gives us alone.

This is relying quite heavily on the sources tests, which I don't have a huge amount of faith in. Also I am taking this as an opportunity to fix issues with defaults if people want to pick on them

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -655,6 +655,19 @@ class GenericMap(NDData):
         """
         return self.data.max(*args, **kwargs)
 
+    @staticmethod
+    def _parse_fits_unit(unit_str):
+        unit = u.Unit(unit_str, format='fits', parse_strict='silent')
+        if isinstance(unit, u.UnrecognizedUnit):
+            warnings.warn(f'Could not parse unit string "{unit_str}" as a valid FITS unit.\n'
+                          f'See {_META_FIX_URL} for how to fix metadata before loading it '
+                          'with sunpy.map.Map.\n'
+                          'See https://fits.gsfc.nasa.gov/fits_standard.html for'
+                          'the FITS unit standards.',
+                          SunpyMetadataWarning)
+            unit = None
+        return unit
+
     @property
     def unit(self):
         """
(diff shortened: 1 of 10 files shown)


ayshih:
> The objective of the map sources system is to provide a standard API to access the metadata from a disparate set of headers.

> We seem to have forgotten this however, and have widely (for a large number of years) taken to editing the header directly in the constructor of the map source.

I'll say up front that I like the approach in this PR.

However, what you assert in the OP is highly contrary to my understanding of your position in the past.  I have repeatedly questioned you why we have the source subclasses if they are just metadata standardizers that don't make use of class inheritance beyond overriding `__init__()`, and you repeatedly defended the thin-wrapper approach and said that the primary reason for subclasses was for representation purposes (e.g., so that `type()` returns something nice).

Admittedly, the last time I brought this up was in 2016 (when hiking in Boulder), so it's been many years.  But, I'd still like to see a debate between @Cadair of 2016 and @Cadair of 2021.

Cadair:
Ok, well apparently I have done a poor job in explaining my position in the past, or my opinion has shifted and I can't remember.

I have always maintained the sole purpose of the map sources is to standardise metadata, and resisted any attempt to for a source to deviate from the API of `GenericMap` (i.e. a magnetogram subclass). As for

> you repeatedly defended the thin-wrapper approach and said that the primary reason for subclasses was for representation purposes (e.g., so that `type()` returns something nice).

I could have sworn my position on this is almost the opposite, I really don't care what type returns. Having subclasses provides a way to override the default places to go look for metadata. Given my desire for the API of all the sources to be the same, having them all be one type would be fine.

---

*Disclaimer: While are are talking metadata let's discuss future plans for metadata. However, I would say that in my opinion this PR is how it *should always have been* currently, and not really a major change.*

My stance on the future of the map sources is really that all the metadata handling should be spun out into it's own class structure, which would remove the need for instrument specific sources in map like there is at the moment. Also any future changes to meta should draw a line between "metadata we process and modify" and "metadata we don't understand and leave alone".

dstansby:
@Cadair do you still plan to work on this in the near future? If not I can try and pick it up for 3.1.

Cadair:
I am really busy at the moment, so if you have the time to run with this please be my guest.

PR #5200, opened 2021-04-11
Moved timeseries peek to plot()

### LINKED ISSUE #5066: Move custom timeseries plotting from from peek to plot
Related to #5058 move custom timeseries plotting from from peek to plot and then have peek call plot. Check and update docs as needed.

Okay, I've made the changes that I wanted.  I didn't update the `peek()` docstrings for the derived TimeSeries classes because their implementations needs to be cleaned up (i.e., source-specific plotting code should be accessible through a `plot()` method, but is only found in the `peek()` method).  I think such changes should be put off to a different PR.

_Originally posted by @ayshih in https://github.com/sunpy/sunpy/issues/3103#issuecomment-494839081_

jeffreypaul15:
### Description

Fixes #5066 
Drawing parallels from GenericMap, I suppose this is along the lines of what we want?
This does override the base `plot()` method defined for TimeSeries though.


jeffreypaul15:
@ayshih Is this something along the lines of what we want?

nabobalis:
This looks like something we want, tho It will need to be done for each time series source since they all have their own peek method.

jeffreypaul15:
> This looks like something we want, tho It will need to be done for each time series source since they all have their own peek method.

Yeah, I just made the change for this one source to confirm if this is what we want. Shall I go about with the rest?

nabobalis:
Sure

jeffreypaul15:
@nabobalis I've added all the TimeSeries sources, but like I mentioned, this does overwrite the `plot()` method defined in the base TimeSeries class, how would we want to handle that?

nabobalis:
> @nabobalis I've added all the TimeSeries sources, but like I mentioned, this does overwrite the `plot()` method defined in the base TimeSeries class, how would we want to handle that?

You mean it overloads it?

sunpy/timeseries/sources/eve.py
```diff
@@ -18,7 +18,7 @@
 from sunpy.util.metadata import MetaDict
 from sunpy.visualization import peek_show
 
-__all__ = ['EVESpWxTimeSeries', 'ESPTimeSeries']
+__all__ = ["EVESpWxTimeSeries", "ESPTimeSeries"]
```
nabobalis, line 21: Did you run a linter?

jeffreypaul15: No I ran codestyle on it, isn't this what we should be using instead?

jeffreypaul15:
> You mean it overloads it?

No, I am overriding the plot method right

sunpy/timeseries/sources/eve.py
nabobalis: What do you mean? The code style shouldn't make any of these changes.

nabobalis:
> > You mean it overloads it?
> 
> No, I am overriding the plot method right

I don't see the issue. What would be the problem.

sunpy/timeseries/sources/eve.py
jeffreypaul15: Oh yeah, my apologies. These were changed by `autopep8`.

jeffreypaul15:
> > > You mean it overloads it?
> > 
> > 
> > No, I am overriding the plot method right
> 
> I don't see the issue. What would be the problem.

We lose that functionality, basically the ability to call the base plot method. Is that fine?

nabobalis:
> > > > You mean it overloads it?
> > > 
> > > 
> > > No, I am overriding the plot method right
> > 
> > 
> > I don't see the issue. What would be the problem.
> 
> We lose that functionality, basically the ability to call the base plot method. Is that fine?

Don't see why not.

jeffreypaul15:
What changelog would this PR require?

nabobalis:
> What changelog would this PR require?

Not sure. It feels kind of breaking to make plot do all this formatting but it doesn't actually break anything.

jeffreypaul15:
> Not sure. It feels kind of breaking to make plot do all this formatting but it doesn't actually break anything.

Shall I go with feature then?

nabobalis:
Sure.

jeffreypaul15:
> Sure.

Need to word the changelog better, please make the required changes.

jeffreypaul15:
@nabobalis I don't suppose extra tests are needed right? The test for `peek()` covers `plot()` so would we need separate tests for `plot()`?

nabobalis:
> @nabobalis I don't suppose extra tests are needed right? The test for `peek()` covers `plot()` so would we need separate tests for `plot()`?

No it should be ok.

jeffreypaul15:
@nabobalis Anything else to be done in this PR?

nabobalis:
It will need a review.

changelog/5200.feature.rst
```diff
@@ -0,0 +1 @@
+Added `plot()` method to all `sunpy.timeseries.sources` and `peek()` now calls plot instead.
```
nabobalis, line 1:
```suggestion
Added ``plot`` method to all `~sunpy.timeseries.Timeseries` sources and the ``peek`` method calls this instead.
This ensures that ``plot`` and ``peek`` will output figures that look the same.
```

sunpy/timeseries/sources/eve.py
```diff
@@ -69,23 +85,12 @@ def peek(self, title='EVE/ESP Level1', **kwargs):
         """
 
         self._validate_data_for_plotting()
```
nabobalis, line 87: Does this not belong in plot?

sunpy/timeseries/sources/eve.py
```diff
@@ -69,23 +85,12 @@ def peek(self, title='EVE/ESP Level1', **kwargs):
...
-
-        names = ('Flux \n 0.1-7nm', 'Flux \n 18nm', 'Flux \n 26nm', 'Flux \n 30nm', 'Flux \n 36nm')
-
         axes = self.to_dataframe().plot(subplots=True, sharex=True, **kwargs)
```
nabobalis, line 88: Same here?

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -59,6 +59,24 @@ class GBMSummaryTimeSeries(GenericTimeSeries):
     # Class attribute used to specify the source class of the TimeSeries.
     _source = 'gbmsummary'
 
+    def plot(self, axes=None, **kwargs):
```
nabobalis, line 62: Docstring missing.

sunpy/timeseries/sources/eve.py
```diff
@@ -57,8 +57,24 @@ class ESPTimeSeries(GenericTimeSeries):
 
     _source = 'esp'
 
+    def plot(self, axes=None, title="EVE/ESP Level1", **kwargs):
```
nabobalis, line 60: Is it not possible to call the generic time series plot function to get a figure and then update the plot?

sunpy/timeseries/sources/eve.py
jeffreypaul15: Should this be moved there? It'd make sense I suppose.

nabobalis: Well why would we only want to validate the data for peek but not plot?

sunpy/timeseries/sources/eve.py
jeffreypaul15: Similar to how GenericMap is written, I've set `axes` in peek and then passed it to plot. Do we not want this?

nabobalis: Does plot not need an axes?

sunpy/timeseries/sources/eve.py
jeffreypaul15: Why though? I'm a bit confused

sunpy/timeseries/sources/eve.py
jeffreypaul15: Yeah that makes sense, I'll make that change.

sunpy/timeseries/sources/eve.py
jeffreypaul15: If plot is called directly it still gets the axes from `plt.gca()` right

sunpy/timeseries/sources/eve.py
nabobalis: It might some code that could be useful?

sunpy/timeseries/sources/eve.py
nabobalis: Then do we need this line?

sunpy/timeseries/sources/eve.py
jeffreypaul15: So what would the generic time series plot function produce? As of now it just plots the timeseries dataframe and isn't specific to any source right

sunpy/timeseries/sources/eve.py
jeffreypaul15: Sorry, I misunderstood this part. There is redundant code here, the reason I did this was in all sources `axes` was set in `peek` and then passed to `plot`. We don't use `plt.gca()` here but for the sake of uniformity I left it as it is, if plot is called directly, the same code is repeated there as well. I'll make the change if this is redundant.

sunpy/timeseries/sources/eve.py
nabobalis: Well if the generic one creates the plot for us, we can call it and then update the figure?

sunpy/timeseries/sources/eve.py
nabobalis: I am not sure, it feels redundant. You might have to have a play around to decide.

jeffreypaul15: This is the only source that behaves this way, the other sources axes aren't defined in this manner so I guess I'll make an exception for this?

sunpy/timeseries/sources/eve.py
jeffreypaul15: Hmm I suppose we could do that as well. Would that be better? I'll re-structure this PR then.

sunpy/timeseries/sources/eve.py
nabobalis: I would like to know why we make this exception for this source.

jeffreypaul15: This source fetches the `axes` from a pandas DataFrame object which returns a list of 2 axes.

sunpy/timeseries/sources/eve.py
jeffreypaul15: I don't suppose, `eve esp` and `lyra` would be exceptions because as I mentioned above, they contain multiple axes so we can't call plot which deals with just one axes. Correct me if I'm wrong, I haven't gone into too much depth.

nabobalis: Not sure.

jeffreypaul15: Nope, I was wrong, setting `subplot=True` would fix the multiple axes problem. Shall I restructure this then?

nabobalis: Sure.

jeffreypaul15:
@nabobalis On continuing with using the source having to call the main class to plot the figure, we get 
`UserWarning: To output multiple subplots, the figure containing the passed axes is being cleared`
The reason is because generic time series just fetches the axes for us and then returns the axes to the source where I'm plotting it. 
Do we just ignore this warning?

nabobalis:
I would like to avoid rasing it if possible

jeffreypaul15:
> I would like to avoid rasing it if possible

As of now it seems to be unavoidable, another problem that I've stumbled upon is that the `noaa`source creates it's axes from it's dataframe like any other source but only certain parts of the entire dataframe is plotted (columns based on user input). Calling plot() here would be redundant as we'd have extra `2D lines` which aren't what is required by the user.
Based on this and the previous warning issue with multiple plots I suggest that it'd be better to not call generic time series to plot, and instead leave that functionality to a source-less timeseries object.

nabobalis:
Sure.

jeffreypaul15:
> Sure.

Should the docstrings have examples?

nabobalis:
No its ok.

sunpy/timeseries/sources/eve.py
```diff
@@ -57,11 +57,14 @@ class ESPTimeSeries(GenericTimeSeries):
 
     _source = 'esp'
 
-    @peek_show
-    def peek(self, title='EVE/ESP Level1', **kwargs):
+    def plot(self, axes=None, title="EVE/ESP Level1", **kwargs):
         """
+        Plots the EVE ESP Leve1 timeseries data.
```
nabobalis, line 62: Does plot not return the figure/axes?

sunpy/timeseries/sources/eve.py
nabobalis: Fair enough.

sunpy/timeseries/sources/eve.py
jeffreypaul15: No, do we want them to return the axes?

ayshih: Yes

sunpy/timeseries/sources/eve.py
```diff
@@ -57,11 +57,14 @@ class ESPTimeSeries(GenericTimeSeries):
...
+
         Parameters
         ----------
+        axes : `matplotlib.axes.Axes`
+            The axes on which to plot the TimeSeries. Defaults to current axes.
```
nabobalis, line 67: Now we need a return section in the doc string.

jeffreypaul15: Ah, will do that.

sunpy/timeseries/sources/eve.py
```diff
@@ -57,35 +57,63 @@ class ESPTimeSeries(GenericTimeSeries):
...
         title : `str`, optional
             Plot title.
         **kwargs : `dict`
             Additional plot keyword arguments that are handed to `pandas.DataFrame.plot`.
+
+        Returns
+        -------
+        axes : `~matplotlib.axes.Axes`
```
nabobalis, line 75:
```suggestion
        `~matplotlib.axes.Axes`
```

jeffreypaul15: Is this the format that's followed everywhere?
Won't we need to change it here as well? https://github.com/sunpy/sunpy/blob/3940d01382c1863e3250b5cc7f96531265b29c43/sunpy/timeseries/timeseriesbase.py#L462

nabobalis: It is the format that should be used.

jeffreypaul15: Alright then, I'll make that change.

jeffreypaul15:
@nabobalis Tests seem to be unrelated issues?

nabobalis:
Yeah.

sunpy/timeseries/sources/eve.py
```diff
@@ -57,35 +57,63 @@ class ESPTimeSeries(GenericTimeSeries):
 
     _source = 'esp'
 
-    @peek_show
-    def peek(self, title='EVE/ESP Level1', **kwargs):
+    def plot(self, axes=None, title="EVE/ESP Level1", **kwargs):
         """
+        Plots the EVE ESP Leve1 timeseries data.
```
nabobalis, line 62:
```suggestion
        Plots the EVE ESP Level 1 timeseries data.
```

sunpy/timeseries/sources/eve.py
```diff
@@ -57,35 +57,63 @@ class ESPTimeSeries(GenericTimeSeries):
 
     _source = 'esp'
 
-    @peek_show
-    def peek(self, title='EVE/ESP Level1', **kwargs):
+    def plot(self, axes=None, title="EVE/ESP Level1", **kwargs):
```
nabobalis, line 60:
```suggestion
    def plot(self, axes=None, title="EVE/ESP Level 1", **kwargs):
```

sunpy/timeseries/sources/eve.py
```diff
@@ -57,35 +57,63 @@ class ESPTimeSeries(GenericTimeSeries):
...
         """
+        Plots the EVE ESP Leve1 timeseries data.
+
         Parameters
         ----------
+        axes : `matplotlib.axes.Axes`
```
nabobalis, line 66:
```suggestion
        axes : `matplotlib.axes.Axes`, optional
```

sunpy/timeseries/sources/eve.py
```diff
@@ -57,35 +57,63 @@ class ESPTimeSeries(GenericTimeSeries):
...
+            The axes on which to plot the TimeSeries. Defaults to current axes.
```
nabobalis, line 67: Is "Defaults to current axes." true?

jeffreypaul15: Sorry, this should be changed.

sunpy/timeseries/sources/eve.py
```diff
@@ -57,35 +57,63 @@ class ESPTimeSeries(GenericTimeSeries):
...
+        Returns
+        -------
+        `~matplotlib.axes.Axes`
+            The plot axes.
         """
 
         self._validate_data_for_plotting()
 
-        names = ('Flux \n 0.1-7nm', 'Flux \n 18nm', 'Flux \n 26nm', 'Flux \n 30nm', 'Flux \n 36nm')
+        if axes is None:
```
nabobalis, line 81: So this function will do nothing if you provide an axes?

nabobalis: We might need some plot tests after all.

jeffreypaul15: We'd need to provide it with a list of axes, should I mention that in the docstring?

nabobalis: Maybe.

jeffreypaul15: As a note? Or should under parameters would do?

nabobalis: Would a list of axes change anything?

jeffreypaul15: This source in particular requires a list of axes right so not passing a list would return an index out of range error, I suppose I'd have to handle that error in the code?

nabobalis: Sure. I think we need plot tests for each source to ensure it works as we want. They won't need to be figure tests. We can leave that to peek.

jeffreypaul15: How would we test these plots without a figure test?

nabobalis: We check the inputs work as we expect them to?

jeffreypaul15: I'm a bit confused as to how exactly we'd be testing the axes input. 
How would we vary the inputs to test?

nabobalis: Its fine, lets not test this.

sunpy/timeseries/sources/eve.py
```diff
@@ -57,35 +57,63 @@ class ESPTimeSeries(GenericTimeSeries):
...
+
+        Parameters
+        ----------
+        title : `str`, optional
+            Plot title.
+        **kwargs : `dict`
+            Additional plot keyword arguments that are handed to `pandas.DataFrame.plot`.
+        """
+        self._validate_data_for_plotting()
+        axes = self.to_dataframe().plot(subplots=True, sharex=True, **kwargs)
```
nabobalis, line 111: This are in the plot function, so I don't see why we can't remove them here?

jeffreypaul15: Sorry, must've forgotten

jeffreypaul15:
@nabobalis The eve, lyra, noaa sources all plot from a pandas dataframe exclusively and each source has their own number of columns. For these particular sources, we can't pass axes as an argument so I have removed that. We don't need tests for these sources as they don't have axes as arguments. 
The remaining sources (rhessi, norh, goes and fermi_gbm) work similar to how GenericMap works and `plt.gca()` can be used, thereby allowing for axes to be passed via :
```python
fig, ax = plt.subplots()
axs = ts.plot(axes=ax)
```
I have written tests to confirm if the returned axes are `matplotlib.axes._subplots.AxesSubplot` but I don't know if this is required. If not I can get rid of them.

nabobalis:
> I have written tests to confirm if the returned axes are `matplotlib.axes._subplots.AxesSubplot` but I don't know if this is required. If not I can get rid of them.

It would be useful to check that the input axes gets updated by plot.

jeffreypaul15:
> It would be useful to check that the input axes gets updated by plot.

Is there a simple way to check for that? This is where I'm a bit confused.

nabobalis:
Not sure.

jeffreypaul15:
> Not sure.

How about checking for the title change in `ax.title`, if the input axes is updated the title is also updated?

nabobalis:
If it works.

jeffreypaul15:
> If it works.

Should be fine now. I've added `title` arguments to the sources that didn't have them as well.

sunpy/timeseries/sources/eve.py
```diff
@@ -57,35 +57,59 @@ class ESPTimeSeries(GenericTimeSeries):
...
-    def peek(self, title='EVE/ESP Level1', **kwargs):
+    def plot(self, title="EVE/ESP Level 1", **kwargs):
         """
+        Plots the EVE ESP Level 1 timeseries data.
+
         Parameters
         ----------
         title : `str`, optional
-            Plot title.
+            The title of the plot. Defaults to "EVE/ESP Level1".
```
nabobalis, line 67:
```suggestion
            The title of the plot. Defaults to "EVE/ESP Level 1".
```

sunpy/timeseries/sources/eve.py
```diff
@@ -57,35 +57,59 @@ class ESPTimeSeries(GenericTimeSeries):
...
+    def plot(self, title="EVE/ESP Level 1", **kwargs):
         """
+        Plots the EVE ESP Level 1 timeseries data.
+
         Parameters
         ----------
         title : `str`, optional
-            Plot title.
+            The title of the plot. Defaults to "EVE/ESP Level 1".
         **kwargs : `dict`
```
nabobalis, line 68: I always forget if this need an optional.

sunpy/timeseries/sources/eve.py
```diff
@@ -57,35 +57,59 @@ class ESPTimeSeries(GenericTimeSeries):
...
+    @peek_show
+    def peek(self, title="EVE/ESP Level 1", **kwargs):
+        """
+        Plots the EVE ESP Level 1 timeseries data by calling
+        `~sunpy.timeseries.sources.eve.ESPTimeSeries.plot`.
 
+        Parameters
+        ----------
+        title : `str`, optional
+            The title of the plot. Defaults to "EVE/ESP Level1".
```
nabobalis, line 103:
```suggestion
            The title of the plot. Defaults to "EVE/ESP Level 1".
```

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -59,10 +59,55 @@ class GBMSummaryTimeSeries(GenericTimeSeries):
...
+
+        data_lab = self.to_dataframe().columns.values
+
+        for d in data_lab:
+            axes.plot(self.to_dataframe().index, self.to_dataframe()[d], label=d, **kwargs)
+
+        axes.set_yscale("log")
+        if title is None:
+            axes.set_title('Fermi GBM Summary data ' + str(self.meta.get(
+                'DETNAM').values()))
```
nabobalis, line 95: This is a strange line break?

jeffreypaul15: Not sure why this happened. I'll fix this

sunpy/timeseries/sources/eve.py
jeffreypaul15: Other parts of the code follow the same behavior so I guess we do it here too?

jeffreypaul15:
Test failures look like unrelated issues.

sunpy/timeseries/sources/eve.py
```diff
@@ -57,35 +57,59 @@ class ESPTimeSeries(GenericTimeSeries):
...
-            Plot title.
+            The title of the plot. Defaults to "EVE/ESP Level1".
         **kwargs : `dict`
             Additional plot keyword arguments that are handed to `pandas.DataFrame.plot`.
+
+        Returns
+        -------
+        `~matplotlib.axes.Axes`
+            The plot axes.
         """
```
nabobalis, line 76: Can you remove all these blank lines?

sunpy/timeseries/sources/eve.py
```diff
@@ -57,35 +57,59 @@ class ESPTimeSeries(GenericTimeSeries):
...
 
         self._validate_data_for_plotting()
 
-        names = ('Flux \n 0.1-7nm', 'Flux \n 18nm', 'Flux \n 26nm', 'Flux \n 30nm', 'Flux \n 36nm')
```
nabobalis, line 73: Did this have to move?

sunpy/timeseries/sources/eve.py
```diff
@@ -57,35 +57,59 @@ class ESPTimeSeries(GenericTimeSeries):
...
+        axes[-1].set_xlabel("Time (UT) " + str(self.to_dataframe().index[0])[0:11])
+        axes[-1].xaxis.set_major_formatter(dates.DateFormatter("%H:%M"))
+
+        return axes
+
+    @peek_show
+    def peek(self, title="EVE/ESP Level 1", **kwargs):
+        """
+        Plots the EVE ESP Level 1 timeseries data by calling
+        `~sunpy.timeseries.sources.eve.ESPTimeSeries.plot`.
```
nabobalis, line 98:
```suggestion
        Displays the EVE ESP Level 1 timeseries data by calling
        `~sunpy.timeseries.sources.eve.ESPTimeSeries.plot`.
```

sunpy/timeseries/sources/eve.py
jeffreypaul15: I moved it a line below, I don't suppose that changes anything right?

nabobalis: No but why?

jeffreypaul15: Looked neater that way, no worries, I'll move it back.

sunpy/timeseries/sources/eve.py
```diff
@@ -57,35 +57,59 @@ class ESPTimeSeries(GenericTimeSeries):
...
+            The title of the plot. Defaults to "EVE/ESP Level1".
+        **kwargs : `dict`
+            Additional plot keyword arguments that are handed to `pandas.DataFrame.plot`.
+        """
+
+        axes = self.plot(title=title, **kwargs)
         fig = axes[0].get_figure()
         fig.tight_layout()
         fig.subplots_adjust(hspace=0.05)
+
```
nabobalis, line 112:
```suggestion
```

sunpy/timeseries/sources/eve.py
```diff
@@ -57,35 +57,59 @@ class ESPTimeSeries(GenericTimeSeries):
...
+        `~sunpy.timeseries.sources.eve.ESPTimeSeries.plot`.
 
+        Parameters
+        ----------
+        title : `str`, optional
+            The title of the plot. Defaults to "EVE/ESP Level1".
+        **kwargs : `dict`
+            Additional plot keyword arguments that are handed to `pandas.DataFrame.plot`.
+        """
+
```
nabobalis, line 107:
```suggestion
```

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -59,10 +59,53 @@ class GBMSummaryTimeSeries(GenericTimeSeries):
     # Class attribute used to specify the source class of the TimeSeries.
     _source = 'gbmsummary'
 
+    def plot(self, axes=None, title=None, **kwargs):
+        """
+        Plots the GBM lightcurve TimeSeries with data from its pandas dataframe.
```
nabobalis, line 64:
```suggestion
        Plots the GBM lightcurve timeseries with data from its pandas dataframe.
```

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -59,10 +59,53 @@ class GBMSummaryTimeSeries(GenericTimeSeries):
...
+
+        return axes
+
     @peek_show
     def peek(self, **kwargs):
         """
-        Plots the GBM lightcurve TimeSeries. An example can be seen below:
+        Plots the GBM lightcurve TimeSeries by calling
+        `~sunpy.timeseries.sources.fermi_gbm.GBMSummaryTimeSeries.plot`.
+        An example can be seen below:
```
nabobalis, line 108:
```suggestion
```

sunpy/timeseries/sources/eve.py
jeffreypaul15: I've removed all extra lines locally, I shall commit separately.

sunpy/timeseries/sources/fermi_gbm.py
jeffreypaul15: Every `peek()` method has this, do we want to get rid of this line?

nabobalis: Sure.

sunpy/timeseries/sources/eve.py
```diff
@@ -57,35 +57,53 @@ class ESPTimeSeries(GenericTimeSeries):
...
+        title : `str`, optional
+            The title of the plot. Defaults to "EVE/ESP Level1".
+        **kwargs : `dict`
+            Additional plot keyword arguments that are handed to `pandas.DataFrame.plot`.
+        """
+        axes = self.plot(title=title, **kwargs)
         fig = axes[0].get_figure()
         fig.tight_layout()
         fig.subplots_adjust(hspace=0.05)
+
```
nabobalis, line 106:
```suggestion
```

sunpy/timeseries/sources/eve.py
```diff
@@ -175,7 +193,7 @@ class EVESpWxTimeSeries(GenericTimeSeries):
     @peek_show
     def peek(self, column=None, **kwargs):
         """
-        Plots the time series in a new figure. An example is shown below:
+        Plots the time series in a new figure.
 
         ..
```
nabobalis, line 198: Can you remove this and see if it works now?

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -59,10 +59,50 @@ class GBMSummaryTimeSeries(GenericTimeSeries):
...
+        **kwargs : `dict`
+            Additional plot keyword arguments that are handed to `~matplotlib.axes.Axes.plot`
+            functions.
+
+        Returns
+        -------
+        `~matplotlib.axes.Axes`
+            The plot axes.
+        """
+
```
nabobalis, line 81:
```suggestion
```

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -59,10 +59,50 @@ class GBMSummaryTimeSeries(GenericTimeSeries):
...
+        self._validate_data_for_plotting()
+
```
nabobalis, line 83:
```suggestion
```

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -59,10 +59,50 @@ class GBMSummaryTimeSeries(GenericTimeSeries):
...
+            axes.plot(self.to_dataframe().index, self.to_dataframe()[d], label=d, **kwargs)
+        axes.set_yscale("log")
+        if title is None:
+            axes.set_title('Fermi GBM Summary data ' + str(self.meta.get('DETNAM').values()))
+        else:
+            axes.set_title(title)
+        axes.set_xlabel('Start time: ' + self.to_dataframe().index[0].strftime('%Y-%m-%d %H:%M:%S UT'))
+        axes.set_ylabel('Counts/s/keV')
+        axes.legend()
+
```
nabobalis, line 97:
```suggestion
```

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -59,10 +59,50 @@ class GBMSummaryTimeSeries(GenericTimeSeries):
...
+        return axes
+
     @peek_show
     def peek(self, **kwargs):
         """
-        Plots the GBM lightcurve TimeSeries. An example can be seen below:
+        Displays the GBM lightcurve timeseries by calling
+        `~sunpy.timeseries.sources.fermi_gbm.GBMSummaryTimeSeries.plot`.
+        An example can be seen below:
```
nabobalis, line 105:
```suggestion
```

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -77,22 +117,8 @@ def peek(self, **kwargs):
...
-
-        ax.set_yscale("log")
-        ax.set_title('Fermi GBM Summary data ' + str(self.meta.get(
-            'DETNAM').values()))
-        ax.set_xlabel('Start time: ' + self.to_dataframe().index[0].strftime(
-            '%Y-%m-%d %H:%M:%S UT'))
-        ax.set_ylabel('Counts/s/keV')
-        ax.legend()
+        self.plot(axes=ax, **kwargs)
         fig.autofmt_xdate()
```
nabobalis, line 123:
```suggestion
```

sunpy/timeseries/sources/goes.py
```diff
@@ -64,63 +64,87 @@ class XRSTimeSeries(GenericTimeSeries):
...
+        axes.yaxis.grid(True, "major")
+        axes.xaxis.grid(False, "major")
+        axes.legend()
 
         # TODO: display better tick labels for date range (e.g. 06/01 - 06/05)
-        formatter = matplotlib.dates.DateFormatter('%H:%M')
-        ax.xaxis.set_major_formatter(formatter)
+        formatter = matplotlib.dates.DateFormatter("%H:%M")
+        axes.xaxis.set_major_formatter(formatter)
+
```
nabobalis, line 121:
```suggestion

```

sunpy/timeseries/sources/goes.py
```diff
@@ -64,63 +64,87 @@ class XRSTimeSeries(GenericTimeSeries):
...
         ax2.set_yticklabels([])
 
-        ax.yaxis.grid(True, 'major')
-        ax.xaxis.grid(False, 'major')
-        ax.legend()
+        axes.yaxis.grid(True, "major")
+        axes.xaxis.grid(False, "major")
+        axes.legend()
 
         # TODO: display better tick labels for date range (e.g. 06/01 - 06/05)
```
nabobalis, line 118: Can we do this?

sunpy/timeseries/sources/goes.py
```diff
@@ -64,63 +64,87 @@ class XRSTimeSeries(GenericTimeSeries):
...
+
+        ax2 = axes.twinx()
         ax2.set_yscale("log")
         ax2.set_ylim(1e-9, 1e-2)
-        labels = ['A', 'B', 'C', 'M', 'X']
+        labels = ["A", "B", "C", "M", "X"]
         centers = np.logspace(-7.5, -3.5, len(labels))
         ax2.yaxis.set_minor_locator(mticker.FixedLocator(centers))
         ax2.set_yticklabels(labels, minor=True)
         ax2.set_yticklabels([])
```
nabobalis, line 113:
```suggestion
```

sunpy/timeseries/sources/goes.py
```diff
@@ -64,63 +64,87 @@ class XRSTimeSeries(GenericTimeSeries):
...
+        ----------
+        title : `str`, optional
+            The title of the plot. Defaults to "GOES Xray Flux".
+        **kwargs : `dict`
+            Additional plot keyword arguments that are handed to `~matplotlib.axes.Axes.plot`
+            functions.
+        """
+        fig, ax = plt.subplots()
+        self.plot(axes=ax, title=title, **kwargs)
         fig.autofmt_xdate()
```
nabobalis, line 149:
```suggestion

```

sunpy/timeseries/sources/lyra.py
```diff
@@ -54,46 +54,68 @@ class LYRATimeSeries(GenericTimeSeries):
...
+    def plot(self, title=None, names=3, **kwargs):
         """
-        Plots the LYRA data. An example is shown below:
-
-        .. plot::
-
-            import sunpy.timeseries
-            import sunpy.data.sample
-            lyra = sunpy.timeseries.TimeSeries(sunpy.data.sample.LYRA_LEVEL3_TIMESERIES, source='LYRA')
-            lyra.peek()
```
nabobalis, line 67: Restore this?

sunpy/timeseries/sources/goes.py
jeffreypaul15: The resultant plots are the same so I suppose we can right

sunpy/timeseries/sources/lyra.py
jeffreypaul15: It's part of peek in the method below, this is plot()

sunpy/timeseries/sources/goes.py
jeffreypaul15: Do we not need an empty line between the method and it's return statement?

sunpy/timeseries/sources/eve.py
jeffreypaul15: The plot? It works right

sunpy/timeseries/sources/goes.py
nabobalis: No?

jeffreypaul15: Oh wow okay, I had been doing that for every PR that I had made until now. I'll make this change.

nabobalis: It's just a style choice.

jeffreypaul15:
@nabobalis Do I need to rebase this as well to check for figure tests?

nabobalis:
Sure.

nabobalis:
Is this ready now you think?

jeffreypaul15:
> Is this ready now you think?

I seem to have messed up the git commands here while rebasing. My aplogies.

jeffreypaul15:
@nabobalis The new figure tests revealed that the previous peek method had title = 
image
and the updated one has : 
image

That's the reason for change in the figure hashes.

nabobalis:
Sure.

jeffreypaul15:
> Sure.

The failing test seems to be unrelated, everything else seems good to go I guess.

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -59,10 +59,46 @@ class GBMSummaryTimeSeries(GenericTimeSeries):
     # Class attribute used to specify the source class of the TimeSeries.
     _source = 'gbmsummary'
 
+    def plot(self, axes=None, title=None, **kwargs):
+        """
+        Plots the GBM lightcurve timeseries with data from its pandas dataframe.
```
samaloney, line 64:
```suggestion
        Plots the GBM timeseries with data from its pandas dataframe.
```

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -59,10 +59,46 @@ class GBMSummaryTimeSeries(GenericTimeSeries):
...
+        axes.set_xlabel('Start time: ' + self.to_dataframe().index[0].strftime('%Y-%m-%d %H:%M:%S UT'))
+        axes.set_ylabel('Counts/s/keV')
+        axes.legend()
+        return axes
+
     @peek_show
     def peek(self, **kwargs):
         """
-        Plots the GBM lightcurve TimeSeries. An example can be seen below:
+        Displays the GBM lightcurve timeseries by calling
```
samaloney, line 100:
```suggestion
        Displays the GBM timeseries by calling
```

sunpy/timeseries/sources/eve.py
```diff
@@ -175,15 +191,14 @@ class EVESpWxTimeSeries(GenericTimeSeries):
...
+        .. plot::
 
-                import sunpy.timeseries
-                from sunpy.data.sample import EVE_TIMESERIES
-                eve = sunpy.timeseries.TimeSeries(EVE_TIMESERIES, source='eve')
-                eve.peek(subplots=True)
+            import sunpy.timeseries
+            from sunpy.data.sample import EVE_TIMESERIES
+            eve = sunpy.timeseries.TimeSeries(EVE_TIMESERIES, source='eve')
+            eve.peek(subplots=True)
```
samaloney, line 201: For some reason this doesn't seem to be showing up in the built version of the [docs]( https://sunpy--5200.org.readthedocs.build/en/5200/api/sunpy.timeseries.sources.ESPTimeSeries.html#sunpy.timeseries.sources.ESPTimeSeries) like the rest

samaloney (approved):

sunpy/timeseries/sources/eve.py
jeffreypaul15: What do you mean?

sunpy/timeseries/sources/eve.py
```diff
@@ -57,32 +57,48 @@ class ESPTimeSeries(GenericTimeSeries):
...
-            ax.legend(loc='upper right')
-        axes[-1].set_xlabel('Time (UT) ' + str(self.to_dataframe().index[0])[0:11])
-        axes[-1].xaxis.set_major_formatter(dates.DateFormatter('%H:%M'))
+            ax.legend(loc="upper right")
+        axes[-1].set_xlabel("Time (UT) " + str(self.to_dataframe().index[0])[0:11])
+        axes[-1].xaxis.set_major_formatter(dates.DateFormatter("%H:%M"))
+        return axes
 
+    @peek_show
+    def peek(self, title="EVE/ESP Level 1", **kwargs):
```
nabobalis, line 89: Does peek need to take title?

jeffreypaul15: I don't see an issue here right? Do we not want that?

nabobalis: Why would we want it on peek?

nabobalis: I do wonder if we want it on plot as well.

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -59,10 +59,46 @@ class GBMSummaryTimeSeries(GenericTimeSeries):
     # Class attribute used to specify the source class of the TimeSeries.
     _source = 'gbmsummary'
 
+    def plot(self, axes=None, title=None, **kwargs):
```
nabobalis, line 62: I feel like the API on plot should be the same for all?

sunpy/timeseries/sources/noaa.py
```diff
@@ -59,60 +60,89 @@ class NOAAIndicesTimeSeries(GenericTimeSeries):
     # Class attribute used to specify the source class of the TimeSeries.
     _source = 'noaaindices'
 
-    @peek_show
-    def peek(self, type='sunspot SWO', **kwargs):
+    def plot(self, title='Solar Cycle Progression', plot_type='sunspot SWO', dataframe=None, **kwargs):
```
nabobalis, line 63: I know I wrote this but I do wonder if this should just output all of the plot types in big subplot.

sunpy/timeseries/sources/fermi_gbm.py
jeffreypaul15: This changes depending on the number of axes right? This source has only one and isn't deriving it from it's pandas dataframe while EVE/ESP and a few others are

sunpy/timeseries/sources/rhessi.py
```diff
@@ -136,59 +136,77 @@ class RHESSISummaryTimeSeries(GenericTimeSeries):
...
+        Returns
+        -------
+        `~matplotlib.axes.Axes`
+            The plot axes.
         """
-        # Check we have a timeseries valid for plotting
         self._validate_data_for_plotting()
-
+        if axes is None:
+            axes = plt.gca()
```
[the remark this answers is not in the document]
jeffreypaul15: Sure, I'll add that in

sunpy/timeseries/sources/noaa.py
jeffreypaul15: Should I make that change?

sunpy/timeseries/sources/eve.py
nabobalis: I think at the very least, peek shouldn't take any arguments.

sunpy/timeseries/sources/fermi_gbm.py
nabobalis: Sure but I guess on some level I am uncomfortable for a function that is named the same on a subclass of time series to a different API.

I don't really have a good answer to this.

jeffreypaul15: They still return the axes similar to the others right so I thought it wouldn't matter but if that's the case then I'm open to removing the plot functions where the axes are not singular

nabobalis: Removing?

sunpy/timeseries/sources/noaa.py
nabobalis: It might be, I will let others have a review and decide.

sunpy/timeseries/sources/fermi_gbm.py
jeffreypaul15: I meant keeping removing it as it was before this PR

nabobalis: Sorry, I've lost track. Removing the axes keywords in plot?

jeffreypaul15: Removing this method altogether, if that might be better (for those with multiple axes, but 3 of them will have multiple axes if we decide NOAA to be one as well)

nabobalis: You mean removing plot? I think we want to keep plot, in future I plan to push to remove peek.

jeffreypaul15: Ah okay, then I shall wait for a better solution to this

sunpy/timeseries/sources/rhessi.py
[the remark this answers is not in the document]
jeffreypaul15: @nabobalis `return fig` from every timeseries `peek()` method is tested via their figure tests right?
I haven't altered those figure test, yet codecov says that the coverage for `return fig` from every source is removed?
That's odd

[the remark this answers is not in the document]
nabobalis: I think that is correct since the return is captured by the decorator?

sunpy/timeseries/timeseriesbase.py
```diff
@@ -461,7 +461,7 @@ def plot(self, axes=None, **plot_args):
```
nabobalis, line 461: I can't highlight the line, but it says above the `**plot_args` are passed to :meth:`pandas.DataFrame.plot`. But is that true?

sunpy/timeseries/sources/rhessi.py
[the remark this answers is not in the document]
jeffreypaul15: I'm sorry I don't follow, so how has the coverage been altered when I'm still returning `fig`?

sunpy/timeseries/sources/eve.py
```diff
@@ -57,32 +57,48 @@ class ESPTimeSeries(GenericTimeSeries):
 
     _source = 'esp'
 
-    @peek_show
-    def peek(self, title='EVE/ESP Level1', **kwargs):
+    def plot(self, title="EVE/ESP Level 1", **kwargs):
```
nabobalis, line 60: On the axes in plot, the original plot method on time series base takes an axes and I think each plot method should as a result at least keep that as a named argument. The idea is to keep the API uniform, we can add title after the fact but I am also leaning toward removing title as an arg fully.

Thoughts?

sunpy/timeseries/sources/rhessi.py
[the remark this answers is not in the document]
nabobalis: Don't know.

sunpy/timeseries/timeseriesbase.py
jeffreypaul15: `self._data` returns a `pandas.core.frame.DataFrame`, so arguments passed to `self._data.plot()` are sent to pandas

sunpy/timeseries/sources/eve.py
jeffreypaul15: I guess that makes sense but the user should pass the axes from a pandas dataframe, else this won't really behave the same way it does when `ax` from `fig, ax = plt.subplots()` is passed to it directly, and `plt.gca` won't work either

dstansby: The issue is that the plots created have multiple axes, so it's not possible to have a single `axes` argument here. I would remove the title argument though, and just always set the title to be what is currently the default. It's easy enough for the user to override this themselves after calling `.plot()`.

sunpy/timeseries/sources/eve.py
```diff
@@ -57,32 +57,48 @@ class ESPTimeSeries(GenericTimeSeries):
...
         title : `str`, optional
-            Plot title.
+            The title of the plot. Defaults to "EVE/ESP Level 1".
         **kwargs : `dict`
             Additional plot keyword arguments that are handed to `pandas.DataFrame.plot`.
-        """
 
+        Returns
+        -------
+        `~matplotlib.axes.Axes`
```
dstansby, line 73: This actually returns an array of `Axes`.
```suggestion
        array of `~matplotlib.axes.Axes`
```

sunpy/timeseries/sources/eve.py
dstansby: Lets leave this for now (we shouldn't change the signature without deprecating anyway), we can always revisit the signature of `peek()` in another issue or PR.

sunpy/timeseries/sources/eve.py
dstansby: I think we should just get rid of these plot example lines, we don't have them for other `peek()` methods.

sunpy/timeseries/sources/fermi_gbm.py
dstansby: I think we should remove the title argument, but leave axes.

dstansby:
I think it's worth holding off on this until we've discussed what signature new `plot()` functions should have at the weekly meeting this week. Once we've discussed I'll post another comment about what we decided.

nabobalis:
Discussion in community meeting:

Make sure we don't add or break/remove the original plot/peek api for timeseries. 
Move all custom plotting to plot as required. 

We would need to deal with plots that have multiple axes but I think we can handle that.

jeffreypaul15:
> We would need to deal with plots that have multiple axes but I think we can handle that.

nabobalis:
> > Discussion in community meeting:
> > Make sure we don't add or break/remove the original plot/peek api for timeseries.
> > Move all custom plotting to plot as required.
> > We would need to deal with plots that have multiple axes but I think we can handle that.
> 
> So, just removing the title arguments?

If thats all the difference is, then yes.

jeffreypaul15:
> If thats all the difference is, then yes.

Do we still have axes as an argument? Because of the multiple axes issue

nabobalis:
> > If thats all the difference is, then yes.
> 
> Do we still have axes as an argument? Because of the multiple axes issue

If its part of the original API, yes.

jeffreypaul15:
> > > If thats all the difference is, then yes.
> > 
> > 
> > Do we still have axes as an argument? Because of the multiple axes issue
> 
> If its part of the original API, yes.

Then I'll have to add the axes back to the sources where there are multiple axes which are being derived from a pandas dataframe, otherwise I don't think there are any changes other than the original plot method now producing the required plot instead of the one in timeseries base

jeffreypaul15:
When we have multiple axes, how do we handle the situation then? For now any axes passed will be overwritten by 
`axes = self.to_dataframe().plot(subplots=True, sharex=True, **kwargs)`
Which is wrong, but I'm not really sure how to proceed with this.

jeffreypaul15:
Do we keep the title argument in peek? Since peek is calling plot do we get rid of the title here as well?

nabobalis:
> Do we keep the title argument in peek? Since peek is calling plot do we get rid of the title here as well?

if title was a keyword taken in peek, it stays.

jeffreypaul15:
@nabobalis The multiple axes issue shouldn't be a problem now, although if the user does:
```py
eve = ts.TimeSeries(esp_filepath)
fig, ax = plt.subplots(5,1)
eve.plot(axes=ax)
```
image

the colours aren't in accordance with when the user doesn't pass any axes, we could fix this by having a colours array but that is specific to this dataframe and I don't suppose that's right to do

```py
eve.plot()
```
produces :
image

nabobalis:
> the colours aren't in accordance with when the user doesn't pass any axes as we're plotting each column of the dataframe separately on each of the axes passed, we could fix this by having a colours array but that is specific to this dataframe and I don't suppose that's right to do

Why?

sunpy/timeseries/sources/eve.py
```diff
@@ -57,32 +57,55 @@ class ESPTimeSeries(GenericTimeSeries):
...
+            for i, ax in enumerate(axes):
+                ax.plot(data_[data_.columns[i]], label=data_.columns[i])
+                ax.set_ylabel(names[i])
+                ax.legend(loc="upper right")
+        elif axes is None:
+            axes = self.to_dataframe().plot(subplots=True, sharex=True, **kwargs)
+            axes[-1].set_xlim(self.to_dataframe().index[0], self.to_dataframe().index[-1])
+            for i, ax in enumerate(axes):
+                ax.set_ylabel(names[i])
+                ax.legend(loc="upper right")
```
nabobalis, line 89: Can this block not be made simpler?

sunpy/timeseries/sources/eve.py
jeffreypaul15: All of them have examples, esp is the only one without

jeffreypaul15:
> 
> 
> > the colours aren't in accordance with when the user doesn't pass any axes as we're plotting each column of the dataframe separately on each of the axes passed, we could fix this by having a colours array but that is specific to this dataframe and I don't suppose that's right to do
> 
> Why?

I was under the assumption that this would be used for a variety of dataframes where the columns might exceed the default number of colours that we specify right?

nabobalis:
> > > the colours aren't in accordance with when the user doesn't pass any axes as we're plotting each column of the dataframe separately on each of the axes passed, we could fix this by having a colours array but that is specific to this dataframe and I don't suppose that's right to do
> > 
> > 
> > Why?
> 
> I was under the assumption that this would be used for a variety of dataframes where the columns might exceed the default number of colours that we specify right?

Maybe?

sunpy/timeseries/sources/eve.py
nabobalis: Can you check that each example shows up and looks nice?

jeffreypaul15:
> 
> 
> > > > the colours aren't in accordance with when the user doesn't pass any axes as we're plotting each column of the dataframe separately on each of the axes passed, we could fix this by having a colours array but that is specific to this dataframe and I don't suppose that's right to do
> > > 
> > > 
> > > Why?
> > 
> > 
> > I was under the assumption that this would be used for a variety of dataframes where the columns might exceed the default number of colours that we specify right?
> 
> Maybe?

`dataframe.plot()` handles this by itself, including the rotation of the `xticklabels` so by plotting it separately we would need to manually set all of this

sunpy/timeseries/sources/eve.py
jeffreypaul15: https://sunpy--5200.org.readthedocs.build/en/5200/api/sunpy.timeseries.sources.EVESpWxTimeSeries.html#sunpy.timeseries.sources.EVESpWxTimeSeries
Looks bad, the others are fine I suppose

nabobalis: I think we should address this.

nabobalis:
> > > > > the colours aren't in accordance with when the user doesn't pass any axes as we're plotting each column of the dataframe separately on each of the axes passed, we could fix this by having a colours array but that is specific to this dataframe and I don't suppose that's right to do
> > > > 
> > > > 
> > > > Why?
> > > 
> > > 
> > > I was under the assumption that this would be used for a variety of dataframes where the columns might exceed the default number of colours that we specify right?
> > 
> > 
> > Maybe?
> 
> `dataframe.plot()` handles this by itself, including the rotation of the `xticklabels` so by plotting it separately we would need to manually set all of this

I guess we add it then?

sunpy/timeseries/sources/eve.py
jeffreypaul15: The figure has too much info to fit in standard figsize, either we increase the figsize which would be inconsistent in the docs, or plot the figures with a few columns (column argument already exists), I'd say the latter approach is better

nabobalis: Why not make it bigger? 

Can we also not change the subplots to use the same x axis?

jeffreypaul15: I thought making it bigger would make it inconsistent with the other plots? If not then we could do that.
> Can we also not change the subplots to use the same x axis?

Yeah, `sharex=True` argument isn't passed this would case them to use the same x axis.

nabobalis: They don't have to be the same size, they display different amounts of data.

jeffreypaul15:
> I guess we add it then?

I've added the exact colours that are cycled and also updated the docs figure to be of a `figsize=(20,10)`, the plot should now be rendered properly.

jeffreypaul15:
@nabobalis the failing tests seem unrelated right?

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -59,10 +59,40 @@ class GBMSummaryTimeSeries(GenericTimeSeries):
...
+        for d in data_lab:
+            axes.plot(self.to_dataframe().index, self.to_dataframe()[d], label=d, **kwargs)
+        axes.set_yscale("log")
+        axes.set_xlabel('Start time: ' + self.to_dataframe().index[0].strftime('%Y-%m-%d %H:%M:%S UT'))
+        axes.set_ylabel('Counts/s/keV')
+        axes.legend()
+        return axes
+
     @peek_show
     def peek(self, **kwargs):
```
nabobalis, line 92: This is lacking a title argument, if all the other peeks have one.

sunpy/timeseries/sources/noaa.py
```diff
@@ -205,11 +227,42 @@ class NOAAPredictIndicesTimeSeries(GenericTimeSeries):
...
+        dataframe['sunspot low'].plot(linestyle='--', color='b', **plot_args)
+        axes.set_ylim(0)
+        axes.set_ylabel('Sunspot Number')
+        axes.yaxis.grid(True, 'major')
+        axes.xaxis.grid(True, 'major')
+        axes.legend()
+        return axes
+
     @peek_show
     def peek(self, **plot_args):
```
nabobalis, line 262: Same here with the title arg.

sunpy/timeseries/sources/norh.py
```diff
@@ -54,10 +54,42 @@ class NoRHTimeSeries(GenericTimeSeries):
...
+        axes.plot(self.to_dataframe().index, self.to_dataframe(), label=data_lab, **kwargs)
+        axes.set_yscale("log")
+        axes.set_ylim(1e-4, 1)
+        axes.set_xlabel('Start time: ' + self.to_dataframe().index[0].strftime(TIME_FORMAT))
+        axes.set_ylabel('Correlation')
+        axes.legend()
+        return axes
+
     @peek_show
     def peek(self, **kwargs):
```
nabobalis, line 89: This one as well. Did we not set this on enough peek methods?

sunpy/timeseries/sources/fermi_gbm.py
jeffreypaul15: Didn't we decide to not add one?
Do I add it in xD

nabobalis: I don't see why some peek methods should have it and some don't.

sunpy/timeseries/sources/norh.py
jeffreypaul15: I think we decided that we wouldn't be adding it in if it had it before I think

sunpy/timeseries/sources/fermi_gbm.py
jeffreypaul15: Yeah, the some of the peek methods had it initially as well, I thought we wouldn't be changing how it was before, so I left it in

nabobalis: I think we wanted to not break the old API but we should also try to unify them where possible.

jeffreypaul15: > > > Discussion in community meeting:
> > > Make sure we don't add or break/remove the original plot/peek api for timeseries.
> > > Move all custom plotting to plot as required.
> > > We would need to deal with plots that have multiple axes but I think we can handle that.
> > 
> > 
> > So, just removing the title arguments?
> 
> If thats all the difference is, then yes.

Okay, so add title arguments for every peek method but not plot right?

nabobalis: Pretty much.

jeffreypaul15: Alright, I'll add that in.

nabobalis: Thanks.

jeffreypaul15:
@nabobalis A few sources (lyra and fermi_gbm) derive their title from the dataframe :
`axes[0].set_title("LYRA ({0:{1}})".format(self.to_dataframe().index[0], TIME_FORMAT))`
Do we still accept a default `title` argument here?

nabobalis:
> @nabobalis A few sources (lyra and fermi_gbm) derive their title from the dataframe :
> `axes[0].set_title("LYRA ({0:{1}})".format(self.to_dataframe().index[0], TIME_FORMAT))`
> Do we still accept a default `title` argument here?

Sure.

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -73,28 +103,18 @@ def peek(self, **kwargs):
...
+        title : `str`, optional
+            The title of the plot.
         **kwargs : `dict`
             Additional plot keyword arguments that are handed to `~matplotlib.axes.Axes.plot`
             functions.
         """
-        # Check we have a timeseries valid for plotting
-        self._validate_data_for_plotting()
-
+        if title:
```
nabobalis, line 112:
```suggestion
        if title is None:
```

sunpy/timeseries/sources/lyra.py
```diff
@@ -54,49 +56,75 @@ class LYRATimeSeries(GenericTimeSeries):
...
+        title : `str`, optional
+            The title of the plot.
+        names : `int`, optional
+            The number of columns to plot. Defaults to 3.
+        **kwargs : `dict`
+            Additional plot keyword arguments that are handed to
+            :meth:`pandas.DataFrame.plot`.
+        """
+        axes = self.plot(names=names, **kwargs)
+        if title:
```
nabobalis, line 123:
```suggestion
        if title is None:
```

sunpy/timeseries/sources/noaa.py
```diff
@@ -56,60 +57,83 @@ class NOAAIndicesTimeSeries(GenericTimeSeries):
     # Class attribute used to specify the source class of the TimeSeries.
     _source = 'noaaindices'
 
-    @peek_show
-    def peek(self, type='sunspot SWO', **kwargs):
+    def plot(self, axes=None, plot_type='sunspot SWO', **kwargs):
```
nabobalis, line 60: The change to "plot_type" will need a breaking changelog entry.

nabobalis:
Can this be squash merged and rebased please?

sunpy/timeseries/sources/noaa.py
jeffreypaul15: I changed this because type is a keyword, should I revert to type?

nabobalis: No, it shouldn't have been `type` in the first place.

changelog/5200.breaking.rst
```diff
@@ -0,0 +1 @@
+:meth:`sunpy.timeseries.sources.NOAAIndicesTimeSeries.peek` accepts ``plot_type`` as an argument instead of ``type``.
```
nabobalis, line 1:
```suggestion
:meth:`sunpy.timeseries.sources.NOAAIndicesTimeSeries.plot` accepts ``plot_type`` as an argument instead of ``type``.
```

jeffreypaul15: peek as well right?

nabobalis: Did `peek` take the argument before?

jeffreypaul15: yeah it did

nabobalis: My bad.

nabobalis (approved):

nabobalis:
I wanted to check what plot vs peek looked light right now.
Each plot is plot then peek.

# noaa_json_pre
noaa_json_pre_plot
noaa_json_pre_peek

# noaa_pre_json
noaa_pre_json_peek
noaa_pre_json_plot

# norh
norh plot broke for me but peek worked.
norh_peek

# rhessi
rhessi_plot
rhessi_peek

# esp
esp_plot
esp_peek

# eve
eve_plot
eve_peek

# fermi_gbm
fermi_gbm_plot
fermi_gbm_peek

# goes
goes_plot
goes_peek

# lyra
lyra_plot
lyra_peek

There are still some big differences? Is there anyway to resolve this?

nabobalis (changes requested):
Undoing my approval.

jeffreypaul15:
@nabobalis These differences are resolved in this PR right?
The peek and plot all look alike for me

nabobalis:
I will double check I had checked out the correct code but this was from testing this PR.

jeffreypaul15:
> I will double check I had checked out the correct code but this was from my attempt at testing this PR.

Sure, the current release version has the same differences that you've mentioned

nabobalis:
I had sunpy installed twice in my env and this created the conflict.

nabobalis (approved):

nabobalis:
Thanks again @jeffreypaul15

Issue #4492, opened 2020-09-23
Passing draw_grid=True to CompositeMap.peek is broken


wtbarnes:
### Description

As it says in the title, passing `draw_grid=True` to `CompositeMap` is broken. This was brought up by @schonfsj.

### Expected vs Actual behavior

A grid to be drawn on my composite map. Instead, it throws an error.

### Steps to Reproduce

MWE courtesy of @schonfsj

```python
import matplotlib.pyplot as plt
import sunpy.map
from sunpy.data.sample import AIA_171_IMAGE, HMI_LOS_IMAGE
aia_map = sunpy.map.Map(AIA_171_IMAGE)
hmi_map = sunpy.map.Map(HMI_LOS_IMAGE)
comp_map = sunpy.map.Map(aia_map, hmi_map, composite=True)
comp_map.peek(draw_grid=True)
```
which throws the exception
```python-traceback
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-d5ad2e978908> in <module>
      5 hmi_map = sunpy.map.Map(HMI_LOS_IMAGE)
      6 comp_map = sunpy.map.Map(aia_map, hmi_map, composite=True)
----> 7 comp_map.peek(draw_grid=True)

~/anaconda/envs/sunpy-user/lib/python3.7/site-packages/sunpy/visualization/visualization.py in show_figure(*args, **kwargs)
     19     @wraps(func)
     20     def show_figure(*args, **kwargs):
---> 21         _ = func(*args, **kwargs)
     22         plt.show()
     23 

~/anaconda/envs/sunpy-user/lib/python3.7/site-packages/sunpy/map/compositemap.py in peek(self, colorbar, basic_plot, draw_limb, draw_grid, **matplot_args)
    482         if isinstance(draw_grid, bool):
    483             if draw_grid:
--> 484                 self.draw_grid(axes=axes)
    485 
    486         elif isinstance(draw_grid, (int, float)):

~/anaconda/envs/sunpy-user/lib/python3.7/site-packages/astropy/units/decorators.py in wrapper(*func_args, **func_kwargs)
    232             # Call the original function with any equivalencies in force.
    233             with add_enabled_equivalencies(self.equivalencies):
--> 234                 return_ = wrapped_function(*func_args, **func_kwargs)
    235             if wrapped_signature.return_annotation not in (inspect.Signature.empty, None):
    236                 return return_.to(wrapped_signature.return_annotation)

[...]

PR #5208, opened 2021-04-14
CompositeMap now uses WCSAxes

ayshih:
`CompositeMap` plotting is currently not WCSAxes-aware.  Trying to provide it a WCSAxes axes creates weird results (e.g., https://github.com/sunpy/sunpy/issues/4492#issuecomment-697082363)  This PR (partially) fixes it by properly figuring out how to place each image on a WCSAxes axes.

The following code:
```python

import astropy.units as u

import sunpy.map
from sunpy.data.sample import AIA_171_IMAGE, HMI_LOS_IMAGE

# Use the bottom hemisphere of the AIA image
aia_submap = aia_map.submap(SkyCoord([-1200, 1200]*u.arcsec, [-1200, 0]*u.arcsec, frame=aia_map.coordinate_frame))

# Use the western hemisphere of the HMI image (but recall that it will be upside down)
hmi_map = sunpy.map.Map(HMI_LOS_IMAGE)
hmi_submap = hmi_map.submap(SkyCoord([0, 1000]*u.arcsec, [-1000, 1000]*u.arcsec, frame=hmi_map.coordinate_frame))

comp_map = sunpy.map.Map(aia_submap, hmi_submap, composite=True)

fig = plt.figure()
ax = fig.add_subplot(111, projection=aia_map)
comp_map.plot(axes=ax)
comp_map.draw_limb(axes=ax, index=0, color='red')
comp_map.draw_grid(axes=ax, index=1, color='magenta')
plt.show()
```
produces
good

instead of
bad

Issues:
- This usage of `extent` and `imshow` works only if the maps are oriented the same or upside down (180 degrees rotated).  Edit: There is now a warning for each map that is not simply oriented.
- The internal use of `Helioprojective.assume_spherical_screen` should be discussed.
- ~I haven't bothered to preserve the old, non-WCSAxes behavior.~  Now restored (with fixing some bugs with the previous implementation)
- The grid overlay is below every map but the first one due to the current lack of `zorder` functionality in WCSAxes.

### DIFF
--- a/sunpy/map/compositemap.py
+++ b/sunpy/map/compositemap.py
@@ -2,6 +2,8 @@
 
 Author: `Keith Hughitt <keith.hughitt@nasa.gov>`
 """
+import warnings
+
 import matplotlib.pyplot as plt
 import numpy as np
 
@@ -10,6 +12,7 @@ import astropy.units as u
 from sunpy.coordinates.frames import Helioprojective
 from sunpy.map import GenericMap
 from sunpy.util import expand_list
+from sunpy.util.exceptions import SunpyUserWarning
 from sunpy.visualization import axis_labels_from_ctype, peek_show, wcsaxes_compat
 
 __all__ = ['CompositeMap']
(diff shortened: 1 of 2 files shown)


ayshih:
> * The grid overlay is below every map but the first one due to the current lack of `zorder` functionality in WCSAxes.

This PR technically fixes #4492, but the results are unsatisfactory in some cases due to the above issue.

ayshih:
Because I have `plot()` now create a WCSAxes, the plotting code doesn't actually need the explicit figure/axes creation or the specification of the projection:
```python
comp_map = sunpy.map.Map(aia_submap, hmi_submap, composite=True)

comp_map.plot()
comp_map.draw_limb(index=0, color='red')
comp_map.draw_grid(index=1, color='magenta')
plt.show()
```

ayshih:
Ha!  The existing figure tests are failing miserably because the AIA and HMI test files are for wildly different observation times (February 2011 and March 2014).  Thus, now that I've made `CompositeMap` more coordinate-aware, the composite is no longer co-registered:
result

ayshih:
I'm closing this PR in favor of #5255, which can handle all alignment issues

PR #5209, opened 2021-04-14
Move the imports after the Python version check in `__init__.py`

Cadair:
I noticed that we import sunpy before doing the python version check. This is a bad idea as the sunpy imports probably use Python features not supported in older releases, which means we will error before we check the version.


setup.cfg
```diff
@@ -188,12 +188,13 @@ exclude =
     build,
     sunpy/data/sample.py,
     sunpy/extern/**,
+    sunpy/__init__.py,
```
Cadair, line 191: This might be overkill, but flake8 was testing that imports were at the top of the file.

nabobalis (approved):

nabobalis:
Backport? Worthy of a changelog?

.pep8speaks.yml
```diff
@@ -6,6 +6,7 @@ pycodestyle:
   exclude:
     - setup.py
     - docs/conf.py
+    - sunpy/__init__.py,
```
nabobalis, line 9:
```suggestion
    - sunpy/__init__.py
```

Issue #4716, opened 2020-12-09
Update metadata model to FITS 4


wtbarnes:
Our metadata model needs updating regarding FITS 4. This was partially motivated by #4712. The purpose of this issue is to discuss exactly how to do this. This will probably need to be split into multiple issues and maybe become a project, but we can leave it as a single issue for now.

Some properties on `GenericMap` that need updating:

* `exposure_time` -- we currently expect `exp_time`, but `xposure` is also (more?) valid
* `date` -- we currently use `date-obs`, but we should think about adding additional properties for `date-avg`, `date-start`, and `date-end`. We can always fall back to `date-obs` for FITS 3

In particular, `date-avg` is the "more correct" obstime to use for defining our coordinate frame, but currently we use `Map.date` (which is defined using `date-obs`) everywhere, including for defining our coordinate frame and observer coordinate.

dstansby:
See https://fits.gsfc.nasa.gov/fits_standard.html for the FITS standard document(s).

dstansby:
I think this is done now - see above linked PRs.

wtbarnes:
We are actually still using `exp_time` by default instead of the FITS-4 `xposure` key for exposure time.

wtbarnes:
I'm closing this in favor of the more specific #6513

PR #5210, opened 2021-04-14
Add Map class for Solar Orbiter EUI

### LINKED ISSUE #4710: Map class for EUI instrument on Solar Orbiter

wtbarnes:
Fixes #4710

This is a first pass at adding a map class for the EUI instrument on Solar Orbiter. I am not very familiar with the instrument so it would be great to have feedback here!

TODO: Figure out test data situation. Possibly just add FITS header as plaintext plus fake data array? The tests are going to fail until this is dealt with.

This is resurrecting #4712 which will never be spoken of again


nabobalis:
I say add a header and test it that way.

sunpy/map/sources/solo.py
```diff
@@ -0,0 +1,68 @@
+"""Solar Orbiter Map subclass definitions"""
```
Cadair, line 1:
```suggestion
"""
Solar Orbiter Map subclass definitions.
"""
```

Cadair (commented):
Looks good, just needs a test.

sunpy/map/sources/tests/test_eui_source.py
```diff
@@ -0,0 +1,265 @@
...
+        WAVEMIN =                  250 / [Angstrom] min wavelength resp. > 0.05 of max
+        WAVEMAX =                  350 / [Angstrom] max wavelength resp. > 0.05 of max
+        SOOPNAME= 'not defined'        / name of the SOOP Campaign that the data belong
+        SOOPTYPE= '000     '           / campaign ID(s) that the data belong to
+        OBS_MODE= 'GENERIC_CALIB'      / observation mode
+        OBS_TYPE= '2ZpG    '           / encoded version of OBS_MODE
+        OBS_ID  = 'SEUI_021A_000_000_2ZpG_11K' / unique ID of the individual observation
+        TARGET  = 'not defined'        / type of target from planning
+        BTYPE   = 'Intensity'          / type of data
+        BUNIT   = 'ct      '           / units of physical value, after BSCALE, BZERO
```
Cadair, line 61: Did you fix this header? or if not why is there the unit fix in the constructor?

wtbarnes: I cheated and roundtripped the file through sunpy first to fix the `WAVEUNIT` issue.

wtbarnes:
Needs squash merging

nabobalis (approved):

sunpy/map/sources/solo.py
```diff
@@ -0,0 +1,70 @@
...
+    * `EUI Instrument Page <https://wwwbis.sidc.be/EUI/EUI/EUI/EUI/EUI/>`__
+    * `Instrument Paper <https://doi.org/10.1051/0004-6361/201936663>`__
+    """
+
+    def __init__(self, data, header, **kwargs):
+        super().__init__(data, header, **kwargs)
+        self._nickname = self.detector
+        self.plot_settings['cmap'] = self._get_cmap_name()
+        self.plot_settings['norm'] = ImageNormalize(
+            stretch=source_stretch(self.meta, AsinhStretch(0.01)), clip=False)
```
dstansby, line 38: What's the motivation for using a `AsinhStretch`, instead of just leaving this unset and defaulting to linear?

sunpy/map/sources/solo.py
```diff
@@ -0,0 +1,70 @@
...
+        # DN is not a FITS standard unit, so convert to counts
+        if self.meta.get('bunit', None) == 'DN':
+            self.meta['bunit'] = 'ct'
+        if self.meta.get('bunit', None) == 'DN/s':
+            self.meta['bunit'] = 'ct/s'
```
dstansby, line 43: This is now done automatically in `sunpy.map.GenericMap._fix_unit()`, so I think these lines can safely go.

sunpy/map/sources/solo.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @property
+    def processing_level(self):
+        if self.meta.get('level'):
+            # The level number is prepended by the letter L
+            return int(self.meta.get('level')[1:])
+
+    @property
+    def exposure_time(self):
+        return self.meta.get('xposure', 0.0) * self.timeunit
```
dstansby, line 53: Not for this PR, but am I right in thinking we should fix this in `GenericMap`?

wtbarnes: Yes. This will have to be part of the larger FITS-4 refactor though. See #4716

sunpy/map/sources/solo.py
wtbarnes: 🤷‍♂️  This is what is done in `AIAMap` so I just repeated it here. I don't think a linear scaling is often a good default for this data, but I could be wrong

wtbarnes: I take that back. Every test image I've looked at looks better with a linear scaling.

wtbarnes:
Online test failure is unrelated

dstansby (approved):

PR #5211, opened 2021-04-15
Adds a `rsun` frame attribute to `Heliographic*` frames

ayshih:
This PR modifies `HeliographicStonyhurst` and `HeliographicCarrington` to be more like `Helioprojective`: rather than automatically converting 2D input to a 3D coordinate upon instantiation (with a hard-coded assumption of the default solar radius), a 2D coordinate is converted to 3D only upon transformations, and uses a `rsun` frame attribute to map to the surface of the Sun.

Before this PR:
```python
>>> import astropy.units as u
>>> from sunpy.coordinates import HeliographicStonyhurst
>>> HeliographicStonyhurst(10*u.deg, 20*u.deg)
<HeliographicStonyhurst Coordinate (obstime=None): (lon, lat, radius) in (deg, deg, km)
    (10., 20., 695700.)>
```

After this PR:
```python
>>> HeliographicStonyhurst(10*u.deg, 20*u.deg)
<HeliographicStonyhurst Coordinate (obstime=None, rsun=695700.0 km): (lon, lat) in deg
    (10., 20.)>
>>> HeliographicStonyhurst(10*u.deg, 20*u.deg).make_3d()
<HeliographicStonyhurst Coordinate (obstime=None, rsun=695700.0 km): (lon, lat, radius) in (deg, deg, km)
    (10., 20., 695700.)>
```

This PR also modifies `GenericMap.draw_grid()` and `sunpy.visualization.wcsaxes_compat.wcsaxes_heliographic_overlay()` so that `rsun` can be passed through to plotting the heliographic grid.  This is important for plotting the grid correctly on Helioprojective maps with a custom `rsun` value (e.g., AIA images).


ayshih:
This PR fixes inaccuracies and visibility issues with grid overlays.  For example, before this PR:
baseline
And after this PR:
result (1)

sunpy/coordinates/wcs_utils.py
```diff
@@ -41,11 +41,31 @@ def solar_wcs_frame_mapping(wcs):
...
+    if hasattr(wcs, 'rsun'):
+        warnings.warn('Support for the .rsun attribute on a WCS is deprecated. '
+                      'Set observer keywords in the FITS header, or directly set the wcs.wcs.aux '
+                      'values instead.',
+                      SunpyUserWarning)
+        if rsun is None:
+            rsun = wcs.rsun
+        else:
+            warnings.warn('rsun information present in WCS auxillary information, ignoring '
+                          '.rsun')
```
Cadair, line 60: This should be a sunpyuserwarning.

Cadair (approved):
It's always a pleasure to review your PRs @ayshih.

This makes a lot of sense and honestly I am not sure why I didn't implement it like this right at the start.

sunpy/coordinates/wcs_utils.py
ayshih: Since this would need to be backported, I created #5219

changelog/5211.feature.rst
```diff
@@ -0,0 +1,2 @@
+`~sunpy.coordinates.frames.HeliographicStonyhurst` and `~sunpy.coordinates.frames.HeliographicCarrington` now have an ``rsun`` frame attribute to specify the radius of the Sun, which defaults to the value in `sunpy.sun.constants`.
+This frame attribute is used when converting a 2D coordinate to a 3D coordinate.
```
dstansby, line 2: Might be worth being clear that `rsun` is not a radial coordinate, but is taken as the default when converting, e.g. something like:

> When the radial coordinate is not explicitly specified, this frame attribute is used as the default radial coordinate when converting a 2D coordinate to a 3D coordinate.

sunpy/coordinates/frames.py
```diff
@@ -87,6 +87,10 @@ def _frame_parameters():
                        "        it must be a solar system body that can be parsed by\n"
                        "        `~sunpy.coordinates.ephemeris.get_body_heliographic_stonyhurst`\n"
                        "        at the time ``obstime``. Defaults to Earth center.")
+    ret['rsun'] = ("rsun : `~astropy.units.Quantity`\n"
+                   "        The physical (length) radius of the Sun. Used to convert a 2D\n"
```
dstansby, line 91:
```suggestion
                   "        The radius of the Sun. Used to convert a 2D\n"
```
? Unless having "physical (length)" here has a technical meaning I've not come across?

sunpy/coordinates/frames.py
```diff
@@ -87,6 +87,10 @@ def _frame_parameters():
...
+                   "        to a 3D coordinate as needed for transformations by assuming that the\n"
+                   "        coordinate is on the surface of the Sun. Defaults to the solar radius.")
```
dstansby, line 93: Can we be more specific here instead of just "the solar radius", perhaps "photospheric radius" or however the default value is defined?

sunpy/coordinates/frames.py
```diff
@@ -169,17 +173,26 @@ class BaseHeliographic(SunPyBaseCoordinateFrame):
...
     }
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    rsun = QuantityAttribute(default=_RSUN, unit=u.km)
 
-        self._make_3d()
+    def make_3d(self):
+        """
+        This method calculates the third coordinate of the frame.
```
dstansby, line 180:
```suggestion
        This method calculates the third (radial) coordinate of the frame.
```
If it's always the radial coordinate that's added with this method?

sunpy/coordinates/frames.py
```diff
@@ -169,17 +173,26 @@ class BaseHeliographic(SunPyBaseCoordinateFrame):
...
+        It assumes that the coordinate point is on the surface of the Sun.
 
-    def _make_3d(self):
-        # Make 3D if specified as 2D
+        Returns
+        -------
+        new_frame : `~sunpy.coordinates.frames.BaseHeliographic`
+            A new frame instance with all the attributes of the original but
```
dstansby, line 186: It should be clarified that if the coordinate is already 3D, it is simply returned and is not a new instance.

sunpy/io/special/asdf/schemas/sunpy.org/sunpy/coordinates/frames/helioprojective-1.0.0.yaml
```diff
@@ -31,8 +31,10 @@ examples:
                     wrap_angle: !<tag:astropy.org:astropy/coordinates/angle-1.0.0> {unit: !unit/unit-1.0.0 deg,
                       value: 180.0}
                 type: SphericalRepresentation
-              frame_attributes: {obstime: !time/time-1.1.0 '2011-01-01T01:01:01.000'}
-            obstime: !time/time-1.1.0 2011-01-01T01:01:01.000
+              frame_attributes:
+                obstime: !time/time-1.1.0 '2011-01-01T01:01:01.000'
+                rsun: !unit/quantity-1.1.0 {unit: !unit/unit-1.0.0 km, value: 695700.0}
+            obstime: !time/time-1.1.0 '2011-01-01T01:01:01.000'
```
dstansby, line 37: I don't know much about these, `.yaml` files, but just wanted to check that it was intended to create new ones for carrington and stonyhurst, but modify an existing one for helioprojective?

sunpy/visualization/wcsaxes_compat.py
```diff
@@ -104,7 +106,8 @@ def default_wcs_grid(axes):
 
 
 @u.quantity_input
-def wcsaxes_heliographic_overlay(axes, grid_spacing: u.deg = 10*u.deg, annotate=True, **kwargs):
+def wcsaxes_heliographic_overlay(axes, grid_spacing: u.deg = 10*u.deg, annotate=True,
```
dstansby, line 109: These new arguments should probably get a feature changelog entry?

sunpy/io/special/asdf/schemas/sunpy.org/sunpy/coordinates/frames/helioprojective-1.0.0.yaml
ayshih: The schema for `Helioprojective` itself doesn't change, but I updated the example it shows the breakdown for the `observer` attribute (`HeliographicStonyhurst`).  I guess I neglected to update the example for `Heliocentric`.

sunpy/visualization/wcsaxes_compat.py
ayshih: Hmm, I can see that formally speaking, but no sane user ought to be calling this function directly...  How about I put an entry under `trivial`?

dstansby: I was thinking that one could use it to overlay grids at different heights or from different times now, but maybe not. Happy to go without or with a trivial entry.

sunpy/coordinates/frames.py
ayshih: The wording, which I did not modify from the existing wording, is presumably to distinguish between the "physical (length)" radius and the "apparent (angle)" radius.

ayshih: I'll change it to "The radius of the Sun in length units."

sunpy/coordinates/frames.py
ayshih: Clarified

sunpy/coordinates/frames.py
ayshih: Greatly expanded the description

sunpy/coordinates/frames.py
ayshih: Now explained earlier

changelog/5211.feature.rst
ayshih: Reworder

sunpy/visualization/wcsaxes_compat.py
ayshih: New entry under `trivial`.

> I was thinking that one could use it to overlay grids at different heights or from different times now

Yes, that's certainly true, but I'd invariably tell a user to use `WCSAxes.get_coords_overlay()` directly instead of using this function buried under `wcsaxes_compat`.

ayshih:
@dstansby All comments addressed

dstansby (approved):

PR #5213, opened 2021-04-15
Fix planet locations example

### LINKED ISSUE #5212: Example gallery "Getting the location of the planets" seems broken RTD
Blank plot since 2.1.2 https://docs.sunpy.org/en/latest/generated/gallery/units_and_coordinates/planet_locations.html#sphx-glr-generated-gallery-units-and-coordinates-planet-locations-py

### LINKED ISSUE #5214: Example gallery "Getting the observer location from a Map" seems broken RTD
Black plot since 2.1.2 https://docs.sunpy.org/en/latest/generated/gallery/units_and_coordinates/getting_observer_location.html#sphx-glr-generated-gallery-units-and-coordinates-getting-observer-location-py

dstansby:
I think `plt.polar` either wasn't working or wasn't autoscaling, but it's better to use not-pyplot anyway.

Fixes #5212
Fixes #5214


nabobalis (approved):

hayesla:
there is also this one https://docs.sunpy.org/en/latest/generated/gallery/map_transformations/reprojection_aia_euvi_mosaic.html

dstansby:
> there is also this one https://docs.sunpy.org/en/latest/generated/gallery/map_transformations/reprojection_aia_euvi_mosaic.html which looks like it also needs this update

Yep, that's here too.

Issue #5218, opened 2021-04-16
Add a `warn_sunpy` function


dstansby:
Whenever I'm adding warnings, I tend to forget to add `SunpyUserWarning` as the type, and I'm not the only one: https://github.com/sunpy/sunpy/pull/5211#discussion_r614787587

It would be nice to

- [ ] Add a `_warn_user()` function to `sunpy/util/exceptions.py` that takes a single message and emits a `SunpyUserWarning`, at an appropriate stacklevel.
- [ ] Replace all calls to `warnings.warn` with this in the main library. 
- [ ] Set up linting to raise an error if `warnings.warn` is used anywhere apart from `sunpy.util.exceptions` (this might be tricky?)

ayshih:
> I'm not the only one

Heh, to be fair, in that PR all I did was relocate code that _you_ wrote (#4620).  So, it wasn't so much my forgetting as my not noticing your forgetting. =)

dstansby:
🤦 apologies!

samriddhi-dev:
Is this done? Or can I take a try in it?

PR #5221, opened 2021-04-16
Add helper `_warns` function

dstansby:
This PR:
- Creates a new `_warn()` function, that raises a warning. The default warning is `SunpyUserWarning`.
- Enforces use of this by banning the import of `warnings` (via `pre-commit`)

I am very open to any and all opinions on this, including whether it's a good idea or not. For what it's worth I caught a few cases where we weren't using `SunpyUserWarning` before.

Fixes https://github.com/sunpy/sunpy/issues/5218


.pre-commit-config.yaml
```diff
@@ -77,6 +77,11 @@ repos:
       - id: isort
         args: ['--sp','setup.cfg']
         exclude: ".*(.fits|.fts|.fit|.txt|tca.*|extern.*|.rst|.md|cm/__init__.py|sunpy/extern|docs/conf.py)$"
+      - id: isort
```
nabobalis, line 80: Why a second isort one?

dstansby: So the exclude list can be distinct.

nabobalis: Do we want to enforce this? Could there not be a case were we would want warnings in future?

sunpy/physics/differential_rotation.py
```diff
@@ -447,6 +447,7 @@ def _warp_sun_coordinates(xy, smap, new_observer, **diff_rot_kwargs):
     See :func:`~sunpy.coordinates.transform_with_sun_center`.
     """
     # Suppress NaN warnings in coordinate transforms
+    import warnings  # isort:skip
```
nabobalis, line 450: I would rather remove the extra pre-commit isort than adding isort:skip for this.

nabobalis (approved):

nabobalis:
Probably worth a trivial changelog entry?

nabobalis:
Just needs a rebase due to the WCS Axes helper PR.

ayshih:
Some thoughts:

- Banning the import of `warnings` seems extreme to me.
- There ought to be developer documentation for how to overcome the ban when a developer wants to use, e.g., `warnings.simplefilter()`.
- I don't get why this function is private instead of public.  I don't like that the coding standard says that a developer must use a private function.
- My inclination would be to have explicit convenience functions for the different warning types – i.e., `warn_user()`, `warn_metadata()`, `warn_deprecation()` – and avoid the optional argument of warning type.
- Not your fault, but why is `GenericMap` not consistent about its use of `SunpyMetadataWarning` and `SunpyUserWarning` for metadata issues?
- The rest of the development team needs to chime in on this one.

nabobalis:
> * Banning the import of `warnings` seems extreme to me.

I agree with this. I also do not like having repeated pre-commit (being picky I know). 

> * I don't get why this function is private instead of public.  I don't like that the coding standard says that a developer must use a private function.

I have little issue if its "public" or "private". I will leave this up for others to decide.

> * My inclination would be to have explicit convenience functions for the different warning types – i.e., `warn_user()`, `warn_metadata()`, `warn_deprecation()` – and avoid the optional argument of warning type.

I do like this, I just wonder about basically wrapping the same base function 3 times. It would be nice to not have to import both warn and a non-standard warning type.

dstansby:
> Some thoughts:
> 
>     * Banning the import of `warnings` seems extreme to me.

The goal is to prevent accidental use of `warnings.warn` without an appropriate sunpy-specific warning being raised, and I'm not sure how to achieve this without (soft-)banning `import warnings`.

>     * There ought to be developer documentation for how to overcome the ban when a developer wants to use, e.g., `warnings.simplefilter()`.

I shall add.

>     * I don't get why this function is private instead of public.  I don't like that the coding standard says that a developer must use a private function.

Because it's not designed to be used by users, and by keeping it private we can modify the API or behaviour as we want. What's the issue with requiring developers to use a private function?

>     * My inclination would be to have explicit convenience functions for the different warning types – i.e., `warn_user()`, `warn_metadata()`, `warn_deprecation()` – and avoid the optional argument of warning type.

I did think about that, and in the end went the other way, but given I'm on the fence and yourself and @nabobalis like this way I will change to more explicit function names.

>     * Not your fault, but why is `GenericMap` not consistent about its use of `SunpyMetadataWarning` and `SunpyUserWarning` for metadata issues?

🤷 , but when I get a spare moment I'm happy to fix in another PR.

>     * The rest of the development team needs to chime in on this one.

Yes, it should go through at least a round of discussion at a weekly meeting.

dstansby:
Meeting decisions:
- Make public API, but it's in `util` so we can change whenever we want
- Make `warn_user`, `warn_deprecated` etc. functions

nabobalis:
Can you copy the commit from this https://github.com/sunpy/sunpy/pull/5367 into the PR when you get a chance to restart working on this?

dstansby:
👍  will do. 😭 so many conflicting files here now!

setup.cfg
```diff
@@ -119,7 +119,6 @@ remote_data_strict = True
 junit_family=xunit1
 filterwarnings =
     error
-    error::sunpy.util.exceptions.SunpyDeprecationWarning
```
nabobalis, line 122: Do we need to remove this? In fact, did we need this here?

sunpy/map/mapbase.py
```diff
@@ -513,6 +512,7 @@ def wcs(self):
         """
         The `~astropy.wcs.WCS` property of the map.
         """
+        import warnings  # isort:skip
```
nabobalis, line 515: Does this need the isort:skip?

sunpy/net/helio/hec.py
```diff
@@ -149,7 +148,7 @@ def search(self, *args, **kwargs):
         table = qrdict.get('table_name', None)
         if table:
             if isinstance(table, bytes):
-                warn('type `bytes` for table_name is deprecated, use `str` instead.', SunpyDeprecationWarning)
+                warn_deprecated('type `bytes` for table_name is deprecated, use `str` instead.')
```
nabobalis, line 151: Should this be removed instead? I forgot what release this was added into.

sunpy/physics/differential_rotation.py
nabobalis, line 450: Does this need the isort:skip?

.pre-commit-config.yaml
```diff
@@ -91,3 +91,9 @@ repos:
         exclude: ".*(.fits|.fts|.fit|.txt|tca.*)$"
       - id: mixed-line-ending
         exclude: ".*(.fits|.fts|.fit|.txt|tca.*)$"
+  - repo: https://github.com/devanshshukla99/pre-commit-hook-prohibit-string
+    rev: v1.2
+    hooks:
+      - id: prohibit-string
+        args: [--prohibit-string, "from warnings import warn,warnings.warn"]
+        exclude: "sunpy/version.py|sunpy/util/decorators.py|sunpy/util/exceptions.py"
```
nabobalis, line 99: What is in the version file?

nabobalis (approved):

nabobalis:
Probably worth a trivial changelog? Maybe it needs a mention in our dev docs somewhere?

setup.cfg
dstansby: The above line errors by default, so there's not much point in this line. I think it's a hangover from when we didn't error by default on everything.

sunpy/map/mapbase.py
dstansby: No, I'll remove them

sunpy/net/helio/hec.py
dstansby: 🤷 I don't know enough about the warning to change it.

.pre-commit-config.yaml
dstansby: The top of the file is
```python
# NOTE: First try _dev.scm_version if it exists and setuptools_scm is installed
# This file is not included in sunpy wheels/tarballs, so otherwise it will
# fall back on the generated _version module.
try:
    try:
        from ._dev.scm_version import version
    except ImportError:
        from ._version import version
except Exception:
    import warnings

    warnings.warn(
        f'could not determine {__name__.split(".")[0]} package version; this indicates a broken installation'
    )
    del warnings

    version = '0.0.0'
```

sunpy/util/exceptions.py
```diff
@@ -55,3 +57,51 @@ class SunpyPendingDeprecationWarning(PendingDeprecationWarning, SunpyWarning):
     """
     A warning class to indicate a soon-to-be deprecated feature.
     """
+
+
+def warn_meta(msg, stacklevel=1):
```
Cadair, line 62: I think it's marginally clearer to say metadata?

```suggestion
def warn_metadata(msg, stacklevel=1):
```

Cadair (approved):
Looks good to me, I think I would rather `warn_metadata` but I am not going to die on that hill either.

sunpy/util/exceptions.py
dstansby: 👍 there's a few minor things I need to fix anyway, so I can make this change too

sunpy/physics/differential_rotation.py
dstansby: Nope, fixed

PR #5222, opened 2021-04-16
[2.1] Fixing visibility issues with the grid overlays for SunPy 2.x

ayshih:
There is an issue with plotting a grid overlay for maps with a custom solar radius (e.g., AIA).  We can't use the solution of #5211 for SunPy 2.x.  So, this PR papers over the issue by directly changing the tolerance in WCSAxes.  The grid overlays will still be slightly inaccurate, but at least all of the lines will be visible.


ayshih:
before
after

Cadair:
I can close my eyes and merge this I guess.

ayshih:
Weird, this PR somehow made the unit annotation disappear?

Before:
baseline (1)
After:
result (2)
Difference:
result-failed-diff

ayshih:
I should point out that the disappearing grid lines currently occur only with Astropy dev, and not yet with any Astropy release.

nabobalis (approved):

PR #5224, opened 2021-04-18
PyInstaller Support for SunPy

### LINKED ISSUE #5207: Cannot run PyInstaller executables importing SunPy
Tested with the simple script 
```
from sunpy.coordinates.sun import earth_distance
print(earth_distance())
```
then did `pyinstaller test1.py` and it finished successfully. But starting executable I got the following:
```
  File "test1.py", line 1, in <module>
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "PyInstaller/loader/pyimod03_importers.py", line 531, in exec_module
  File "sunpy/__init__.py", line 51, in <module>
  File "sunpy/__init__.py", line 41, in _get_bibtex
FileNotFoundError: [Errno 2] No such file or directory: '/home/yakob/projects/sdc/grisview.dist/tmp/dist/test1/sunpy/CITATION.rst'
[138952] Failed to execute script test1
```
Checking astropy and sunpy folders in PyInstaller dist folder, looks like sunpy is really missing files (there is only `io` folder). Copying `CITATION.rst` to the folder leads to another missing file error. Tested scripts using AstroPy and they work well with PyInstaller. Done on a 'clean' venv environment (and conda' one as well).
Python 3.8.5
AstroPy 4.2.1
SunPy 2.1.4
OS: Kubuntu 20.04

jeffreypaul15:
Fixes #5207 
Based on AstroPy's hook and testing script : https://github.com/astropy/astropy/blob/main/.pyinstaller/run_astropy_tests.py
I've written a similar hook for SunPy, although the testing script is quite different from the one over at AstroPy. 

The hook is at `sunpy/.pyinstaller/hooks/hook-sunpy.py` which can be loaded by `--additional-hooks-dir` while running PyInstaller.

Instead of having to create a separate 'sunpy_tests' directory locally (astropy's version),
the `executable` created from PyInstaller runs :
`pytest -p no:warnings --doctest-rst -m "not mpl_image_compare" --pyargs sunpy`
which runs all the tests by default through the executable.

Support will have to be added upstream to [pyinstaller-hooks-contrib](https://github.com/pyinstaller/pyinstaller-hooks-contrib)


nabobalis:
What needs to be added upstream?

jeffreypaul15:
> What needs to be added upstream?

The Hook support has to be added to PyInstaller directly

nabobalis:
Would that change this PR? Will you be doing that?

jeffreypaul15:
> Would that change this PR? Will you be doing that?

It would remove the hook file from here as it would be added to PyInstaller directly. If this PR is implemented correctly, sure I would love to make the PR over at PyInstaller.

nabobalis:
So this PR adds support but in theory we could move that support upstream and not need this PR? Or we would still need something here?

jeffreypaul15:
> So this PR adds support but in theory we could move that support upstream and not need this PR? Or we would still need something here?

We would still need the test script that is implemented in this PR and the tox environment for it, only `hook-sunpy.py` would be moved to PyInstaller.

tox.ini
```diff
@@ -179,3 +179,34 @@ install_command = pip install --no-deps {opts} {packages}
...
+commands =
+    pyinstaller --onefile run_sunpy_tests.py \
+                --distpath . \
+                --additional-hooks-dir hooks \
+                --exclude-module tkinter \
+                --hidden-import pytest \
+                --hidden-import pytest_doctestplus.plugin \
+                --hidden-import pytest_mpl.plugin
+
+    ./run_sunpy_tests
```
nabobalis, line 212: How much of this do we need?

jeffreypaul15: What do you mean? I shall remove the comments, I forgot to do that.

nabobalis: Each command, the mpl version, the pyinstaller version?

jeffreypaul15: The matplotlib issue still persists with PyInstaller so that is required.
Do we need to test pyinstaller with the dev version? Why not just use the latest working commit?

jeffreypaul15:
Astropy runs the PyInstaller tests on their github daily workflow, would we need to do that as well?

nabobalis:
Yes, we need to make sure this works on our CI somehow.

tox.ini
```diff
@@ -179,3 +179,30 @@ install_command = pip install --no-deps {opts} {packages}
 commands =
     conda list
     {env:PYTEST_COMMAND} {posargs}
+
+[testenv:pyinstaller]
+description = check that sunpy can be included in a pyinstaller bundle
+changedir = .pyinstaller
+deps =
+    # The developer version of PyInstaller doesn't work yet because it
```
Cadair, line 187: Is this copied out of astropy? Is this still true for us, this PR linked here is pretty old.

tox.ini
```diff
@@ -179,3 +179,30 @@ install_command = pip install --no-deps {opts} {packages}
...
+commands =
+    pyinstaller --onefile run_sunpy_tests.py \
+                --distpath . \
+                --additional-hooks-dir hooks \
+                --exclude-module tkinter \
+                --hidden-import pytest \
+                --hidden-import pytest_doctestplus.plugin \
+                --hidden-import pytest_mpl.plugin
+
+    ./run_sunpy_tests
```
Cadair, line 208: Rather than making a separate script can we not do this?

```suggestion
    python -m pytest -p no:warnings --doctest-rst -m "not mpl_image_compare" --pyargs sunpy
```

tox.ini
```diff
@@ -179,3 +179,30 @@ install_command = pip install --no-deps {opts} {packages}
...
+    net
+    all
+commands =
+    pyinstaller --onefile run_sunpy_tests.py \
+                --distpath . \
+                --additional-hooks-dir hooks \
+                --exclude-module tkinter \
+                --hidden-import pytest \
+                --hidden-import pytest_doctestplus.plugin \
+                --hidden-import pytest_mpl.plugin
```
Cadair, line 206: Are these just here to make it possible to run the tests?

tox.ini
```diff
@@ -179,3 +179,30 @@ install_command = pip install --no-deps {opts} {packages}
...
+    # Matplotlib to an older version for now - see the following issue
+    # for more details: https://github.com/pyinstaller/pyinstaller/issues/5004
+    matplotlib<3.3
+extras =
+    tests
+    net
+    all
+commands =
+    pyinstaller --onefile run_sunpy_tests.py \
+                --distpath . \
```
Cadair, line 201: If my suggestion about the test command works we wouldn't need this?
```suggestion
    pyinstaller --distpath . \
```

tox.ini
```diff
@@ -179,3 +179,30 @@ install_command = pip install --no-deps {opts} {packages}
...
+    # so for now we pin to the latest working commit.
+    git+https://github.com/pyinstaller/pyinstaller.git@c111ec5c#egg=PyInstaller
+    # PyInstaller is not yet compatible with Matplotlib 3.3 so we pin
+    # Matplotlib to an older version for now - see the following issue
+    # for more details: https://github.com/pyinstaller/pyinstaller/issues/5004
+    matplotlib<3.3
+extras =
+    tests
+    net
+    all
```
Cadair, line 198: We don't need these here do we? All we are running from tox is pyinstaller? or does pyinstaller copy from your current env?

Cadair (commented):
Thanks for working on this @jeffreypaul15

tox.ini
jeffreypaul15: Yeah, I took reference from AstroPy, I think this wouldn't be a problem for Sunpy though.

tox.ini
jeffreypaul15: Correct me if I'm wrong, but we're creating the executable and making it run all the rests right (to ensure that the executable created for sunpy is working properly)?  What you've suggested would just execute the python statement in the tox environment, which is essentially just testing normally.

tox.ini
jeffreypaul15: This is to create the executable file

tox.ini
jeffreypaul15: Doing this would create an entire directory of files, as I mentioned earlier we want to test the executable for SunPy. We're making an `exe` and then running it.

tox.ini
jeffreypaul15: Tox needs these to ensure that the tests run fine by itself, PyInstaller then copies the current environment (tox) and creates the executable based on the environment it was running on.

nabobalis:
If we could get the hook added upstream first, then we can deal with this PR?

jeffreypaul15:
> If we could get the hook added upstream first, then we can deal with this PR?
Sure, shall I make a PR regarding that?

nabobalis:
Sure.

jeffreypaul15:
> Sure.

I've made the PR over at https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/112
I'm not sure how long it'll take for them to review it though as it has been a while since anyone has written any hooks.

tox.ini
bwoodsend: The matplotlib 3.3 incompatibility should have been cleared up in pyinstaller/pyinstaller#5006 and then 3.4 compatibility in pyinstaller/pyinstaller#5568 which is included in PyInstaller 4.3 (which released a few days ago). If you're still seeing breakages then that's quite worrying...

.pyinstaller/run_sunpy_tests.py
```diff
@@ -0,0 +1,4 @@
+import os
+import sys
+
+sys.exit(os.system('pytest -p no:warnings --doctest-rst -m "not mpl_image_compare" --pyargs sunpy'))
```
bwoodsend, line 4: No no no! This will just run regular unfrozen pytest so it's not testing anything except `os.system()`! Use `pytest.main()` instead.
```suggestion
import pytest

# Set some default args.
args = ['-p', 'no:warnings', '--doctest-rst', '-m',
        'not mpl_image_compare. '--pyargs sunpy']
# And allow additional arguments to be supplied at runtime.
args += sys.argv[1:]

pytest.main(args)
```

However, without an `import sunpy` in there, PyInstaller will have no reason to think that `sunpy` is a dependency so it won't collect it (or run the hook) unless you force it to with a `--hiddenimport=sunpy`. 

That's still probably not the end of the story though - it depends: Does `import sunpy` load everything so that you can access any public corner without additional import statements or are you expected to use `from sunpy.foo.bar import xyz` to pull the bits you need out? If it's the latter then PyInstaller will consider `sunpy.foo.bar` to be unneeded unless you add it as a `hiddenimport` or an explicit reference in `run_sunpy_tests.py` too. (This won't apply if you opt to keep the `collect_submodules("sunpy")` in the hook.)

It's quite likely that you'll have to have more than one test script, each importing and testing a different corner of `sunpy`, and each using only the optional dependencies that that corner requires.

jeffreypaul15:
> Sure.

Another issue with this current testing methodology is : 
```python
from pkg_resources import get_distribution
get_distribution('sunpy')
```
Doesn't work with PyInstaller, `pkg_resources` is required for `sunpy.util.sysinfo` and this causes the failure of testing the executable through PyInstaller.

The https://github.com/sunpy/sunpy/commit/0df8534f3423f6fe9961a59a93fdbeccad0fd945#diff-0c2fc003b7a8d4cda00792118c87f10505af98697c012a80656f63cb926c88a6 update to sysinfo.py is a possible issue with pyinstaller and `pkg_resources` so I'm not sure we can use PyInstaller with SunPy till there's a workaround for this.

Although, PyInstaller can be used with any current stable version of SunPy just before from `get_distribution` from `pkg_resources` was added.

jeffreypaul15:
> Oh, that one is quite an easy one. `get_distribution('sunpy')` requires the package metadata (that `sunpy-x.y.z.dist-info` folder that pip creates whenever you install sunpy) to be included. Just put in `hook-sunpy.py`:
> 
> ```python
> from PyInstaller.utils.hooks import copy_metadata
> datas = copy_metadata("sunpy")
> ```
> 
> Of course, as you already have a `datas = collect_data_files("sunpy")` in there you'd want to combine the two:
> 
> ```python
> datas += copy_metadata("sunpy")
> ```

Thanks for this.

.pyinstaller/run_sunpy_tests.py
nabobalis: Thanks for your help on this @bwoodsend.

jeffreypaul15:
@nabobalis what would we want to do here? We can either skip the tests dependencies while making the hook but this complicates testing, or we could have a separate internal hook within the repository for testing and have the main hook present upstream without the original test dependencies.

nabobalis:
> @nabobalis what would we want to do here? We can either skip the tests dependencies while making the hook, but this complicates testing, or we could have a separate internal hook within the repository for testing and have the main hook present upstream without the original test dependencies.

Whatever makes it easier to setup and to ensure that hook works and the offline tests pass.

jeffreypaul15:
> > @nabobalis what would we want to do here? We can either skip the tests dependencies while making the hook, but this complicates testing, or we could have a separate internal hook within the repository for testing and have the main hook present upstream without the original test dependencies.
> 
> Whatever makes it easier to setup and to ensure that hook works and the offline tests pass.

Offline tests require certain imports like `sunpy.net.hek.tests` which aren't needed for when a user is using sunpy. 
Like mentioned above, including all of these imports would make all the tests run fine but it increases the optional dependencies while making the executable. We have around 154 such imports which aren't required (other than testing).
Astropy has gone with the approach of having to include everything in the hook and make the tests easier so this PR was based on the same idea, although https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/112#discussion_r617836049 is a reason to reconsider that.

nabobalis:
> Offline tests require certain imports like `sunpy.net.hek.tests` which aren't needed for when a user is using sunpy.

Is this a problem? 

> Like mentioned above, including all of these imports would make all the tests run fine but it increases the optional dependencies while making the executable. We have around 154 such imports which aren't required (other than for testing).

154 imports? 

> Astropy has gone with the approach of having to include everything in the hook and make the tests easier so this PR was based on the same idea, although [pyinstaller/pyinstaller-hooks-contrib#112 (comment)](https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/112#discussion_r617836049) is a reason to reconsider that.

I don't see an issue with that comment? The main sunpy import doesn't do much if I recall.

jeffreypaul15:
> 154 imports only for tests?

154 such files that are required for offline tests to run with PyInstaller.

> Is this a problem?

Yes, because we only need these files for testing, we don't need them if we're not running any tests.

nabobalis:
> > 154 imports only for tests?
> 
> 154 such files that are required for offline tests to run with PyInstaller.

Ah I see.

> > Is this a problem?
> 
> Yes, because we only need these files for testing, we don't need them if we're not running any tests.
> 
> So the question is whether to have all of these files included in the SunPy hook

Would having two hooks, one for the install and one to enable testing work betteR?

jeffreypaul15:
> > 154 imports only for tests?
> 
> 154 such files that are required for offline tests to run with PyInstaller.
> 
> > Is this a problem?
> 
> Yes, because we only need these files for testing, we don't need them if we're not running any tests.



> > > 154 imports only for tests?
> > 
> > 
> > 154 such files that are required for offline tests to run with PyInstaller.
> 
> Ah I see.
> 
> > > Is this a problem?
> > 
> > 
> > Yes, because we only need these files for testing, we don't need them if we're not running any tests.
> > So the question is whether to have all of these files included in the SunPy hook
> 
> Would having two hooks, one for the install and one to enable testing work betteR?

Yeah, that's what I thought would be work but I'm not sure upstream support can be added for 2 hooks though. We'd need to have the hook for tests locally if that's the case.

nabobalis:
Then we hook it locally for the tests then.

jeffreypaul15:
> Then we hook it locally for the tests then.

Yeah, there's another issue with line https://github.com/sunpy/sunpy/blob/53a59df1b4d69503d79a47ead5705b0999aba735/sunpy/util/sysinfo.py#L62 and PyInstaller which has to be fixed first.

nabobalis:
Fixed how?

jeffreypaul15:
> Fixed how?

I'm not too sure, I've mentioned it over at the PR, the docs don't seem to have any info on stuff like this

nabobalis:
What is wrong with that line?

jeffreypaul15:
> What is wrong with that line?

`pkg_resources.UnknownExtra: sunpy 2.1.dev1336+g1f4d94597.d20210425 has no such extra feature 'all'`

Working with pkg_resources and PyInstaller, I suppose there must be a workaround by using `copy_metadata` like mentioned above

nabobalis:
> > What is wrong with that line?
> 
> `pkg_resources.UnknownExtra: sunpy 2.1.dev1336+g1f4d94597.d20210425 has no such extra feature 'all'`
> 
> Working with pkg_resources and PyInstaller, I suppose there must be a workaround by using `copy_metadata` like mentioned above.

Then we do that?

jeffreypaul15:
> > What is wrong with that line?
> 
> `pkg_resources.UnknownExtra: sunpy 2.1.dev1336+g1f4d94597.d20210425 has no such extra feature 'all'`
> 
> Working with pkg_resources and PyInstaller, I suppose there must be a workaround by using `copy_metadata` like mentioned above.

`copy_metadata` by itself doesn't seem to fix it, I think I've got to dig a little deeper

jeffreypaul15:
@nabobalis Can we replace these lines with from sunpy.coordinates import whatever or from sunpy import sun, instead offrom .. or ... import whatever?
https://github.com/sunpy/sunpy/blob/9e3a51a12fe237aabea4e191f1ccd39c2aa5cc9b/sunpy/coordinates/tests/test_frames.py#L18

jeffreypaul15:
As of now, I've gotten the tests to run successfully with a tox venv, although these lines had to be removed because they seemed to have an issue with PyInstaller and even after looking everywhere couldn't find a fix for it.

- [ ] `extra_reqs = get_distribution("sunpy").requires(extras=["all"])` : extras['all'] from setup.py isn't being loaded by PyInstaller
- [ ] `asdf = pytest.importorskip('asdf', '2.0')` :  asdf isn't being recognized by PyInstaller and version gives me an unknown module error
- [ ] `(LooseVersion(asdf.__version__) < LooseVersion("2.3.1")` : Version issue, here as well.

By commenting these lines and replacing the 
```python
from ... import sun
```
with 
```python
from sunpy import sun
```
Most of the rests, about 95% of them seem to work

nabobalis:
> @nabobalis Can we replace these lines with from sunpy.coordinates import whatever or from sunpy import sun, instead offrom .. or ... import whatever?
> https://github.com/sunpy/sunpy/blob/9e3a51a12fe237aabea4e191f1ccd39c2aa5cc9b/sunpy/coordinates/tests/test_frames.py#L18

Those relative imports should be absolute ones.

nabobalis:
> * [ ]  `extra_reqs = get_distribution("sunpy").requires(extras=["all"])` : extras['all'] from setup.py isn't being loaded by PyInstaller

This is kind of important for the util function to work correctly. I guess we could replace all with a list of every group of extras instead.

> * [ ]  `asdf = pytest.importorskip('asdf', '2.0')` :  asdf isn't being recognized by PyInstaller and version gives me an unknown module error
> * [ ]  `(LooseVersion(asdf.__version__) < LooseVersion("2.3.1")` : Version issue here as well.

These seem like asdf issues, maybe it would be better to bump the min version of asdf and remove these (if that works). It is already set to, "asdf>=2.6".

jeffreypaul15:
> This is kind of important for the util function to work correctly. I guess we could replace all with a list of every group of extras instead.

For now, astropy gathers all the tests and creates a separate folder called `astropy_tests` to run with PyInstaller, I've done the same thing with the script. 
The only tests failing are the ones involving pytest-mocker : fixture 'mocker' not found
PyInstaller seems to have this issue as well.

bwoodsend:
>  The only tests failing are the ones involving pytest-mock : fixture 'mocker' not found
PyInstaller seems to have this issue as well, pytest-mock doesn't seem to be getting detected.

Pytest plugins are loaded via `pkg_resources.iter_entry_points("pytest11")` which requires the package metadata for all pytest plugins you wish to use. I wrote a helper to collect them easily [`collect_entry_point()`](https://pyinstaller.readthedocs.io/en/stable/hooks.html#PyInstaller.utils.hooks.collect_entry_point).

jeffreypaul15:
> > The only tests failing are the ones involving pytest-mock : fixture 'mocker' not found
> > PyInstaller seems to have this issue as well, pytest-mock doesn't seem to be getting detected.
> 
> Pytest plugins are loaded via `pkg_resources.iter_entry_points("pytest11")` which requires the package metadata for all pytest plugins you wish to use. Again, PyInstaller needs to be hand held through this. I wrote a helper to collect them easily [`collect_entry_point()`](https://pyinstaller.readthedocs.io/en/stable/hooks.html#PyInstaller.utils.hooks.collect_entry_point).

Thanks for the help, this solved most of the issues.

jeffreypaul15:
> This is kind of important for the util function to work correctly. I guess we could replace all with a list of every group of extras instead.

https://github.com/pyinstaller/pyinstaller/issues/3033 is still an open issue regarding this. So we'll need to have a workaround to get SunPy to work with PyInstaller.

> * [ ]   `asdf = pytest.importorskip('asdf', '2.0')` :  asdf isn't being recognized by PyInstaller and version gives me an unknown module error
> * [ ]   `(LooseVersion(asdf.__version__) < LooseVersion("2.3.1")` : Version issue here as well.

If minimun version of asdf is set to, "asdf>=2.6", can we get rid of these checks?

nabobalis:
> > This is kind of important for the util function to work correctly. I guess we could replace all with a list of every group of extras instead.
> 
> [pyinstaller/pyinstaller#3033](https://github.com/pyinstaller/pyinstaller/issues/3033) is still an open issue regarding this. So we'll need to have a workaround to get SunPy to work with PyInstaller.

Fair enough.

> > * [ ]    `asdf = pytest.importorskip('asdf', '2.0')` :  asdf isn't being recognized by PyInstaller and version gives me an unknown module error
> > * [ ]    `(LooseVersion(asdf.__version__) < LooseVersion("2.3.1")` : Version issue here as well.
> 
> If minimun version of asdf is set to, "asdf>=2.6", can we get rid of these checks?

I don't see why not.

.pyinstaller/hooks/hook-sunpy.py
```diff
@@ -0,0 +1,25 @@
+#-----------------------------------------------------------------------------
```
nabobalis, line 1: Do we need this copyright?

.pyinstaller/hooks/hook-sunpy.py
```diff
@@ -0,0 +1,25 @@
...
+from PyInstaller.utils.hooks import collect_data_files, collect_submodules, copy_metadata, collect_entry_point
+
+
+datas, hiddenimports = collect_entry_point("pytest11")
+
+datas += collect_data_files("sunpy")
+datas += collect_data_files("dask")
+datas += collect_data_files("drms")
+datas += copy_metadata("sunpy")
+datas += [('/home/jeffrey/.local/lib/python3.8/site-packages/astroquery/CITATION', 'astroquery')]
```
nabobalis, line 19: I assume this has to change?

jeffreypaul15:
@nabobalis I've added the new test hook which runs all tests except for the ones I've skipped because of an issue with PyInstaller.
After the workaround for 

> - [ ] `extra_reqs = get_distribution("sunpy").requires(extras=["all"])` : extras['all'] from setup.py isn't being loaded by PyInstaller

Is figured out then I shall make a PR upstream with the required imports for a hook without tests.

.pyinstaller/hooks/hook-sunpy.py
jeffreypaul15: Oh yeah, sorry I shall make that change.

sunpy/io/special/asdf/tags/tests/test_coordinate_frames.py
```diff
@@ -63,10 +63,7 @@ def test_hgc_100():
 
 # Skip these two tests on windows due to a weird interaction with atomicfile
 # and tmpdir
-skip_windows_asdf = pytest.mark.skipif(
-    (LooseVersion(asdf.__version__) < LooseVersion("2.3.1")
-     and platform.system() == 'Windows'),
-    reason="See https://github.com/spacetelescope/asdf/pull/632")
+skip_windows_asdf = pytest.mark.skipif(platform.system() == 'Windows', reason="See https://github.com/spacetelescope/asdf/pull/632")
```
nabobalis, line 66: Do we still need to skip this?

jeffreypaul15: Oh yeah this PR seems to have been merged, I don't think we need a skip

.pyinstaller/hooks/hook-sunpy.py
bwoodsend: We (as in PyInstaller) do to appease GPL. You guys shouldn't because the BSD license doesn't mandate that you put a copyright header on every file. And it's your file anyway so it should be under your copyright anyway. You can (should?) get rid of it here.

nabobalis: Thanks @bwoodsend, I think we will remove it and have this file under our BSD licence instead.

tox.ini
nabobalis: Ideally if we can test with a released version that would be better.

tox.ini
```diff
@@ -179,3 +179,30 @@ install_command = pip install --no-deps {opts} {packages}
 commands =
     conda list
     {env:PYTEST_COMMAND} {posargs}
+
+[testenv:pyinstaller]
```
nabobalis, line 183: This needs adding to the azure templates file so it can run.

tox.ini
jeffreypaul15: Oh yeah sorry I undid these changes accidentally.

sunpy/util/sysinfo.py
```diff
@@ -59,10 +59,10 @@ def system_info():
     """
     base_reqs = get_distribution("sunpy").requires()
     base_reqs = {base_req.name.lower() for base_req in base_reqs}
-    extra_reqs = get_distribution("sunpy").requires(extras=["all"])
```
jeffreypaul15, line 62: @nabobalis Here's where we'll need to figure a workaround, for now I've removed it for PyInstaller's tests to pass.

nabobalis: What about replacing the "all" with a list of every section?

bwoodsend: I'm working on that bug (pyinstaller/pyinstaller#5774) so in theory you should just be able to twiddle your thumbs for a few days...

bwoodsend: It's merged. We're planning on a 4.3.1 patch release soon so you shouldn't have to install from Github for long. It'd probably be a good idea to stick a PyInstaller version check at the top of the hook though...

changelog/5224.feature.rst
```diff
@@ -0,0 +1 @@
+Added hook file and tests for using PyInstaller with SunPy.
```
nabobalis, line 1:
```suggestion
Added hook file and tests for using PyInstaller with sunpy.
```

.pyinstaller/run_sunpy_tests.py
```diff
@@ -0,0 +1,81 @@
...
+        'test_missing_dependencies_by_extra',
+        'test_hgc_100',
+        'test_missing_dependencies_by_extra',
+        'test_basic',
+        'test_data_manager',
+        'test_file_tampered',
+        'test_download_cache',
+        'test_skip_all',
+        'test_override_file',
+        'test_same_file_id_different_module']
```
nabobalis, line 34: Can any of these be removed now?

.pyinstaller/run_sunpy_tests.py
```diff
@@ -0,0 +1,81 @@
+import os
+import sys
+import shutil
+
+import pytest
+
+# Skipping these tests that take check the name of the current module (which ends up starting with
+# sunpy_tests rather than sunpy).
+# asdf path issue with PyInstaller as well.
```
nabobalis, line 9: Is this an asdf issue or a sunpy issue?

sunpy/io/special/asdf/tags/tests/test_genericmap.py
```diff
@@ -2,14 +2,11 @@
 isort:skip_file.
 """
 # flake8: noqa: E402
```
nabobalis, line 4: Remove this.

.pyinstaller/run_sunpy_tests.py
jeffreypaul15: From the script, these tests either use current directory's file name which is 'sunpy_tests' instead of 'sunpy' which resutls in an assertion error.
Or this is an issue with asdf, the asdf pathing seems to work fine when run without PyInstaller

tox.ini
nabobalis: Do we still need all of these now?

all, encompasess net. can we get away with just test?

.pyinstaller/run_sunpy_tests.py
jeffreypaul15: An issue specifically with asdf and PyInstaller, works fine when used without. 
The other tests take the wrong directory into consideration.

tox.ini
jeffreypaul15: All doesn't seem to load Beautiful Soup which is part of net, this gave me an error locally I might need to check what else is required from tests and add them individually

sunpy/util/sysinfo.py
nabobalis: Thanks again @bwoodsend

tox.ini
nabobalis: It should, I wonder what the cause is.

jeffreypaul15: My apologies, must've been some local issue, we don't need net

.pyinstaller/hooks/hook-sunpy.py
```diff
@@ -0,0 +1,14 @@
...
+
+datas += collect_data_files("sunpy")
+datas += collect_data_files("dask")
+datas += collect_data_files("drms")
+datas += copy_metadata("sunpy")
+
+hiddenimports += collect_submodules('sunpy')
+hiddenimports += collect_submodules('numpy.distutils')
+hiddenimports += collect_submodules('distutils')
+hiddenimports += ['skimage.filters.rank.core_cy_3d']
```
Cadair, line 13: Say what now? Why is this here?

.pyinstaller/run_sunpy_tests.py
Cadair: Is this just the tests, or will asdf fail to find the schemas and break when running user code?

.pyinstaller/hooks/hook-sunpy.py
jeffreypaul15: I encountered these missing imports when running the SunPy's tests for PyInstaller, adding them fixed it.

.pyinstaller/run_sunpy_tests.py
jeffreypaul15: Just these tests from what I've noticed. I haven't checked it thoroughly though, is there any way we can?

.pyinstaller/hooks/hook-sunpy.py
Cadair: I am curious as to how we got here in the import stack, I didn't realise we had even an optional transitive dependency on skimage.

jeffreypaul15:
An update to Pyinstaller caused a bootloader error with the tox environment.

.pyinstaller/hooks/hook-sunpy.py
jeffreypaul15: https://github.com/sunpy/sunpy/blob/f1c45df9f555ca34f5846ddc615e2c6bd5d25c0f/sunpy/image/coalignment.py#L24
Which leads to a hidden import in skimage is the reason.
This is essential I suppose so I don't think we need it just for the tests.

sunpy/coordinates/tests/test_frames.py
```diff
@@ -13,10 +13,15 @@
...
+from sunpy.coordinates.frames import (
+    Heliocentric,
+    HeliographicCarrington,
+    HeliographicStonyhurst,
+    Helioprojective,
+)
 from sunpy.time import parse_time
 from sunpy.util.exceptions import SunpyUserWarning
-from ... import sun
-from ..frames import Heliocentric, HeliographicCarrington, HeliographicStonyhurst, Helioprojective
```
nabobalis, line 19: Can you pull these import changes and the test changes out of this PR and into another one?

jeffreypaul15: Sure.

nabobalis:
This needs a rebase or main merged in.

Is this complete otherwise?

jeffreypaul15:
> This needs a rebase or main merged in.
> 
> Is this complete otherwise?

Yeah, I for now till the ASDF paths are eventually sorted out with PyInstaller, the local hook should work for anyone trying to use PyInstaller for SunPy. Should I add a separate file for non test hooks? I think that would be better though.

nabobalis:
> > This needs a rebase or main merged in.
> > Is this complete otherwise?
> 
> Yeah, I for now till the ASDF paths are eventually sorted out with PyInstaller, the local hook should work for anyone trying to use PyInstaller for SunPy. Should I add a separate file for non test hooks? I think that would be better though.

So what does this mean for ASDF? IF someone were to open an ASDF file it would error in sunpy?

Sure add a file for the non-test hooks.

jeffreypaul15:
> So what does this mean for ASDF? IF someone were to open an ASDF file it would error in sunpy?

Turns out ASDF has an issue with sunpy not just with the tests. I'm not sure as to why this happens though

nabobalis:
> > So what does this mean for ASDF? IF someone were to open an ASDF file it would error in sunpy?
> 
> Turns out ASDF has an issue with sunpy not just with the tests. I'm not sure as to why this happens though

Could this be an upstream ASDF issue?

jeffreypaul15:
> > > So what does this mean for ASDF? IF someone were to open an ASDF file it would error in sunpy?
> > 
> > 
> > Turns out ASDF has an issue with sunpy not just with the tests. I'm not sure as to why this happens though
> 
> Could this be an upstream ASDF issue?

I think so, a separate hook needs to be added for asdf too.

nabobalis:
> > Could this be an upstream ASDF issue?
> 
> This seems like an issue with the dev version of PyInstaller. The latest stable release of pyinstaller works fine with the tests but it has the `extras=['all']` issue, now that I'm installing the dev version from git the asdf path issue persists. I think we might need to wait for the next stable release for this.

Won't the next stable release be the dev version and then have the same issue?

jeffreypaul15:
> > > Could this be an upstream ASDF issue?
> > 
> > 
> > This seems like an issue with the dev version of PyInstaller. The latest stable release of pyinstaller works fine with the tests but it has the `extras=['all']` issue, now that I'm installing the dev version from git the asdf path issue persists. I think we might need to wait for the next stable release for this.
> 
> Won't the next stable release be the dev version and then have the same issue?

I'm assuming that they'd fix it before the release, should I report it to them? I can't seem to find the change that caused it.

bwoodsend:
I suspect that this asdf issue is in part due to pyinstaller/pyinstaller#5774. asdf makes use of `pkg_resources.require("pyyaml")` which requires all metadata to be included recursively for `pyyaml` and all its dependencies. For some reason we haven't worked out, before pyinstaller/pyinstaller#5774 `pkg_resources.require("pyyaml")` did nothing. asdf also requires its own metadata, both for a version and for an entry point. And it's full of yaml files so it needs those collected too. I can get a hello world asdf script running with:
```
pyinstaller --collect-data asdf --copy-metadata asdf --copy-metadata pyyaml --copy-metadata numpy --copy-metadata semantic-version --copy-metadata pyrsistent --copy-metadata six --copy-metadata attrs --copy-metadata setuptools test.py
```
although we (PyInstaller) will want to stream line it into something more sane. I'm adding a recursive option to copy metadata to deal with `pkg_resources.require("...")` issues. And I'm experimenting with scanning bytecode for uses of `pkg_resources` and `importlib.metatdata` to see if we can automate metadata completely.

nabobalis:
Where did we get this again? I forgot.

jeffreypaul15:
> Where did we get with this again? I forgot.

Last I checked, I think the asdf issue still persists, we would be able to use sunpy without asdf as of this current PR

nabobalis:
Will need a rebase due to the file conflict.

jeffreypaul15:
> Will need a rebase due to the file conflict.

On it.

Cadair (approved):

nabobalis:
What will getting https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/112 merged in do for this PR or will it change nothing?

jeffreypaul15:
@Cadair 
The multiple .fits, .rst, .csv and .txt files are are loaded when we perform `collect_data_files("sunpy")`.
77/134 of these files are required for testing only, I'll open a PR over at PyInstaller for the hook without these files.

jeffreypaul15:
> What will getting [pyinstaller/pyinstaller-hooks-contrib#112](https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/112) merged in do for this PR or will it change nothing?

It won't really do much, since I'm going to open another PR which only contains the required files for sunpy to run without tests

nabobalis:
> > What will getting [pyinstaller/pyinstaller-hooks-contrib#112](https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/112) merged in do for this PR or will it change nothing?
> 
> It won't really do much, since I'm going to open another PR which only contains the required files for sunpy to run without tests.
> We do however need to keep in mind that ASDF won't work either way

Another PR here or upstream?

jeffreypaul15:
> Another PR here or upstream?

Upstream

nabobalis:
> > Another PR here or upstream?
> 
> Upstream

Why another one? Is the one you opened not enough?

jeffreypaul15:
> > > Another PR here or upstream?
> > 
> > 
> > Upstream
> 
> Why another one? Is the one you opened not enough?

I closed that one, thought it would be better to create a new one instead.

nabobalis:
So outside of that and asdf, this PR is good to go?

Can you open an issue on the asdf repo about pyinstaller please?

jeffreypaul15:
> So outside of that and asdf, this PR is good to go?

It should be.

> Can you open an issue on the asdf repo about pyinstaller please?

Sure

nabobalis (approved):

PR #5227, opened 2021-04-19
Replace commas with "&" in fido.search() in docs

### LINKED ISSUE #5226: Prefer `Fido.search(a1 & a2)` to `Fido.search(a1, a2)` in docs?
As far as I know `Fido.search(a1, a2)` will implicitly combine all the attribute arguments using the `&` operator. I think it would be clearer if we used `Fido.search(a1 & a2)` in the docs instead, as this is more explicit, and conveys the concept of combining different attributes in a single search.

FinityEx:
### Description

Replaced commas with ampersand in every file containig fido.search() in documentaion. From fido.search(arg1, arg2, arg3) to fido.search(arg1 & arg2 & arg3) as per request.

Fixes #5226

### DIFF
--- a/changelog/5227.trivial.rst
+++ b/changelog/5227.trivial.rst
@@ -1,2 +1,1 @@
-Replaced ',' with '&' in fido.search() in following doc files: docs/code_ref/net.rst, docs/guide/acquiring_data/fido.rst, docs/guide/acquiring_data/hek.rst, docs/guide/acquiring_data/jsoc.rst, docs/guide/tour.rst, docs/whatsnew/0.8.rst, docs/whatsnew/0.9.rst, docs/whatsnew/2.0.rst, 
-docs/whatsnew/2.1.rst.
+Replaced ',' with '&' in fido.search() in multiple files.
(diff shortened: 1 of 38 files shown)


FinityEx:
> Are there any other places where we might need to make this change?

Not within docs folder, no.

nabobalis:
What about outside the docs folder?

FinityEx:
Apart from actual code files, there's no fido.search() anywhere else

nabobalis:
So there other examples we can change in the code files?

FinityEx:
Plenty of them, actually. Most of them in sunpy/sunpy/net and sunpy/examples folders, some test files etc. 
Should those be changed as well?

nabobalis:
I don't see why not, at least in any examples. It's not worth changing the tests.

changelog/5227.trivial.rst
```diff
@@ -0,0 +1,2 @@
+Replaced ',' with '&' in fido.search() in following doc files: docs/code_ref/net.rst, docs/guide/acquiring_data/fido.rst, docs/guide/acquiring_data/hek.rst, docs/guide/acquiring_data/jsoc.rst, docs/guide/tour.rst, docs/whatsnew/0.8.rst, docs/whatsnew/0.9.rst, docs/whatsnew/2.0.rst,
```
dstansby, line 1: I don't think there needs to be a list of filenames here, just the first bit of the changelog entry is fine.

FinityEx:
Changed all files except tests, changed changelog as well. I don't know why my newest commit comment is cut in half, though. Is it because of double quotes use?

nabobalis:
Hi @FinityEx, sorry for the late response.

Do you mind doing two changes to this PR?

1. You will need to rebase or merge in master due to the file conflicts
2. We want to line break before the `&` if a line is too long instead of line breaking after.

PR #5081, opened 2021-03-12
Context manager to handle differential rotation automatically in WCSAxes

ayshih:
Working with `RotatedSunFrame` is sometimes confusing – even for me! – and plotting exactly what you want with WCSAxes methods can be tricky or, in some cases, impossible.  This PR provides a context manager that enables a WCSAxes instance to automatically account for solar differential rotation, based on the difference between the `obstime` of supplied coordinates or coordinate frames and the `obstime` of the WCSAxes's coordinate frame.  All use of `RotatedSunFrame` is hidden from the user.

To do:
- [x] Fix `GenericMap.draw_grid()`
- [x] Fix `GenericMap.draw_limb()`
- [ ] Add tests
- [ ] Add examples
- [ ] Figure out what changes in WCSAxes would make this code less hacky

## plot_coord()
Here, I create a coordinate for disk center four days before an AIA image.  When I don't use the context manager, `plot_coord()` plots the coordinate as seen from the change in observer location (i.e., Earth's motion along its orbit).  When I use the context manager, four days of differential rotation are automatically accounted for, in addition to change in observer location.
```python
import matplotlib.pyplot as plt
import astropy.units as u

import sunpy.map
from sunpy.coordinates import Helioprojective, transform_with_sun_center
from sunpy.data.sample import AIA_171_IMAGE
from sunpy.visualization import diffrot_axes

aiamap = sunpy.map.Map(AIA_171_IMAGE)
old_disk_center = Helioprojective(0*u.arcsec, 0*u.arcsec, observer='earth', obstime=aiamap.date - 4*u.day)

fig = plt.figure()
ax = plt.subplot(projection=aiamap)
aiamap.plot(clip_interval=(1., 99.95)*u.percent)

ax.plot_coord(old_disk_center, 'bx', label="Without diffrot_axes")

with transform_with_sun_center(), diffrot_axes(ax):
    ax.plot_coord(old_disk_center, 'g+', label="With diffrot_axes")

plt.legend()

plt.show()
```
Figure_1

## get_transform()
Same here, but for `get_transform()`, here used for patches.
```python
from astropy.visualization.wcsaxes import Quadrangle

old_hpc = old_disk_center.replicate_without_data()

fig = plt.figure()
ax = plt.subplot(projection=aiamap)
aiamap.plot(clip_interval=(1., 99.95)*u.percent)

q1 = Quadrangle((0, 0)*u.arcsec, 100*u.arcsec, 500*u.arcsec,
[...]

PR #5237, opened 2021-04-21
Overhaul `GenericMap.draw_limb()` for transformations and to show visibility

ayshih:
(pulled out of #5081)

This PR enhances `GenericMap.draw_limb()` so that the solar limb can be drawn correctly on a WCSAxes based on a different WCS (e.g., plot the EUVI limb on an AIA map).  The limb will not necessarily be a circle, and the non-visible part of the limb is dotted rather than solid.

The PR includes a figure test that generates the following plot:
Figure_1

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -2079,7 +2079,7 @@ class GenericMap(NDData):
             is_visible = limb_in_axes.spherical.distance <= reference_distance
         else:
             # If the axes has no observer, the entire limb is considered visible
-            is_visible = np.ones_like(limb_in_axes.spherical.distance, np.bool)
+            is_visible = np.ones_like(limb_in_axes.spherical.distance, bool, subok=False)
 
         # Create the Polygon for the near side of the Sun (using a solid line)
         if 'linestyle' not in kwargs:
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -2013,7 +2013,9 @@ class GenericMap(NDData):
 
         To avoid triggering Matplotlib auto-scaling, these patches are added as
         artists instead of patches.  One consequence is that the plot legend is not
-        populated automatically when the limb is specified with a text label.
+        populated automatically when the limb is specified with a text label.  See
+        :ref:`sphx_glr_gallery_text_labels_and_annotations_custom_legends.py` in
+        the Matplotlib documentation for examples of creating a custom legend.
         """
         # Put import here to reduce sunpy.map import time
         from matplotlib import patches
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -2019,7 +2019,6 @@ class GenericMap(NDData):
         """
         # Put import here to reduce sunpy.map import time
         from matplotlib import patches
-        from matplotlib.path import Path
 
         # Don't use _check_axes() here, as drawing the limb works fine on none-WCSAxes,
         # even if the image is rotated relative to the axes
@@ -2081,21 +2080,23 @@ class GenericMap(NDData):
             # If the axes has no observer, the entire limb is considered visible
             is_visible = np.ones_like(limb_in_axes.spherical.distance, bool, subok=False)
 
+        # Identify discontinuities in the limb
+        # Use the same approach as astropy.visualization.wcsaxes.grid_paths.get_lon_lat_path()
+        step = np.sqrt((vertices[1:, 0] - vertices[:-1, 0]) ** 2 +
+                       (vertices[1:, 1] - vertices[:-1, 1]) ** 2)
+        continuous = np.concatenate([[True, True], step[1:] < 100 * step[:-1]])
+
         # Create the Polygon for the near side of the Sun (using a solid line)
         if 'linestyle' not in kwargs:
             c_kw['linestyle'] = '-'
         visible = patches.Polygon(vertices, **c_kw)
-        visible_codes = visible.get_path().codes
-        visible_codes[:-1][~is_visible] = Path.MOVETO
-        visible_codes[-1] = Path.MOVETO if not is_visible[0] else Path.LINETO
+        _modify_polygon_visibility(visible, is_visible & continuous)
 
         # Create the Polygon for the far side of the Sun (using a dotted line)
         if 'linestyle' not in kwargs:
             c_kw['linestyle'] = ':'
         hidden = patches.Polygon(vertices, **c_kw)
-        hidden_codes = hidden.get_path().codes
-        hidden_codes[:-1][is_visible] = Path.MOVETO
-        hidden_codes[-1] = Path.MOVETO if is_visible[0] else Path.LINETO
+        _modify_polygon_visibility(hidden, ~is_visible & continuous)
 
         # Add both patches as artists rather than patches to avoid triggering auto-scaling
         axes.add_artist(visible)
@@ -2575,3 +2576,12 @@ def _figure_to_base64(fig):
     buf = BytesIO()
     fig.savefig(buf, format='png', facecolor='none')  # works better than transparent=True
     return b64encode(buf.getvalue()).decode('utf-8')
+
+
+def _modify_polygon_visibility(polygon, keep):
+    # Put import here to reduce sunpy.map import time
+    from matplotlib.path import Path
+
+    polygon_codes = polygon.get_path().codes
+    polygon_codes[:-1][~keep] = Path.MOVETO
+    polygon_codes[-1] = Path.MOVETO if not keep[0] else Path.LINETO
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -31,7 +31,7 @@ import sunpy.coordinates
 import sunpy.io as io
 import sunpy.visualization.colormaps
 from sunpy import config, log
-from sunpy.coordinates import HeliographicCarrington, HeliographicStonyhurst, Helioprojective, get_earth, sun
+from sunpy.coordinates import HeliographicCarrington, HeliographicStonyhurst, get_earth, sun
 from sunpy.coordinates.utils import get_rectangle_coordinates
 from sunpy.image.resample import resample as sunpy_image_resample
 from sunpy.image.resample import reshape_image_to_4d_superpixel
@@ -2011,7 +2011,7 @@ class GenericMap(NDData):
         If the limb is a true circle, ``visible`` will instead be
         `~matplotlib.patches.Circle` and ``hidden`` will be ``None``.
 
-        To avoid trigger Matplotlib auto-scaling, these patches are added as generic
+        To avoid triggering Matplotlib auto-scaling, these patches are added as
         artists instead of patches.  One consequence is that the plot legend is not
         populated automatically when the limb is specified with a text label.
         """
@@ -2036,19 +2036,13 @@ class GenericMap(NDData):
             transform = axes.transData
         else:
             radius = self.rsun_obs.to_value(u.deg)
-
-            if isinstance(self.coordinate_frame, Helioprojective):
-                frame = self.coordinate_frame
-            else:
-                frame = Helioprojective(observer=self.observer_coordinate,
-                                        obstime=self.date, rsun=self.rsun_meters)
-            transform = axes.get_transform(frame)
+            transform = axes.get_transform('world')
 
         # transform is always passed on as a keyword argument
         c_kw.setdefault('transform', transform)
 
         # If not WCSAxes or if the map's frame matches the axes's frame, we can use Circle
-        if not is_wcsaxes or frame == axes._transform_pixel2world.frame_out:
+        if not is_wcsaxes or self.coordinate_frame == axes._transform_pixel2world.frame_out:
             c_kw.setdefault('radius', radius)
 
             circ = patches.Circle([0, 0], **c_kw)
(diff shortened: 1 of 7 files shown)


sunpy/map/mapbase.py
```diff
@@ -1963,50 +1963,122 @@ def draw_grid(self, axes=None, grid_spacing: u.deg = 15*u.deg, annotate=True, **
...
 
         Notes
         -----
-        Keyword arguments are passed onto `matplotlib.patches.Circle`.
+        Keyword arguments are passed onto the patches.
+
+        If the limb is a true circle, ``visible`` will instead be
+        `~matplotlib.patches.Circle` and ``hidden`` will be ``None``.
+
+        To avoid trigger Matplotlib auto-scaling, these patches are added as generic
```
dstansby, line 1999:
```suggestion
        To avoid triggering Matplotlib auto-scaling, these patches are added as
```

sunpy/map/mapbase.py
```diff
@@ -1963,50 +1963,122 @@ def draw_grid(self, axes=None, grid_spacing: u.deg = 15*u.deg, annotate=True, **
...
+        artists instead of patches.  One consequence is that the plot legend is not
+        populated automatically when the limb is specified with a text label.
```
dstansby, line 2001: Maybe add a ref to https://matplotlib.org/stable/gallery/text_labels_and_annotations/custom_legends.html in case someone wants to add a legend for these?

sunpy/map/mapbase.py
```diff
@@ -1963,50 +1963,122 @@ def draw_grid(self, axes=None, grid_spacing: u.deg = 15*u.deg, annotate=True, **
...
-        else:
-            radius = self.rsun_obs.value
-        c_kw = {'radius': radius,
-                'fill': False,
+        c_kw = {'fill': False,
                 'color': 'white',
-                'zorder': 100,
-                'transform': transform
-                }
+                'zorder': 100}
```
dstansby, line 2015: Whilst the artist/patch is being changed, I would recommend setting this to 1, which is default for `Patches`: https://matplotlib.org/stable/gallery/misc/zorder_demo.html. This should be adequate to show above a `WCSAxes`, which is treated by Matplotlib as a single `Artist`:

https://github.com/astropy/astropy/blob/5f8866e41164293e463a3cc53fadb5566ba488cc/astropy/visualization/wcsaxes/core.py#L30

for which the default zorder is 0:

https://github.com/matplotlib/matplotlib/blob/ba373b8862d9d458a118db04b6922f67e1115912/lib/matplotlib/artist.py#L97

sunpy/map/mapbase.py
```diff
@@ -1963,50 +1963,122 @@ def draw_grid(self, axes=None, grid_spacing: u.deg = 15*u.deg, annotate=True, **
...
+        segments, represented by solid and dotted lines, respectively.
 
         Parameters
         ----------
-        axes: `~matplotlib.axes` or None
-            Axes to plot limb on or None to use current axes.
+        axes : `~matplotlib.axes` or ``None``
+            Axes to plot limb on or ``None`` to use current axes.
+        resolution : `int`
+            The number of points to use to represent the limb.
```
dstansby, line 1981:
```suggestion
            The number of segments to use to represent the limb.
```
I think, because you use `resolution+1` in `linspace` below.

sunpy/map/mapbase.py
```diff
@@ -1963,50 +1963,122 @@ def draw_grid(self, axes=None, grid_spacing: u.deg = 15*u.deg, annotate=True, **
...
+
+            circ = patches.Circle([0, 0], **c_kw)
+            axes.add_artist(circ)
+
+            return circ, None
+
+        # Otherwise, we use Polygon to be able to distort the limb
+
+        theta = np.linspace(0, 2*np.pi, resolution+1)[:-1] << u.rad
+        # Use the Cartesian approximation of helioprojective coordinates because the Sun is small
```
dstansby, line 2047: Can you not call a transform directly on the `r, theta` coordinate values somehow?

sunpy/map/mapbase.py
```diff
@@ -1963,50 +1963,122 @@ def draw_grid(self, axes=None, grid_spacing: u.deg = 15*u.deg, annotate=True, **
...
+        Tx, Ty = radius * np.cos(theta), radius * np.sin(theta)
+
+        # Perform the round-trip transformation world->pixel->world
```
dstansby, line 2050:
```suggestion
        # Perform the round-trip transformation map_world->map_pixel->axes_world
```
Or similar, to be clear that this is to transform from the map coordinates to the coordinates used for the axes.

sunpy/map/mapbase.py
```diff
@@ -1963,50 +1963,122 @@ def draw_grid(self, axes=None, grid_spacing: u.deg = 15*u.deg, annotate=True, **
...
+        true_vertices = np.array([Tx, Ty]).T
+        apparent_vertices = transform.inverted().transform(transform.transform(true_vertices))
+
+        # Points that move by less than 1 arcmin are intepreted to be on the "near" side of the Sun
```
dstansby, line 2054:
```suggestion
        # Points that move by less than 10 arcsec are interpreted to be on the "near" side of the Sun
```

dstansby, line 2054: I'm struggling to get how this logic works. If I'm plotting the SDO limb from the POV of STEREO, at a 45 deg observer difference, don't all the points move? I think I'm probably not understanding exactly what the transform chain above is doing.

dstansby (commented):
My big question is, does this work with CAR projections (ie. synoptic maps)? It would be amazing if it did...

ayshih:
> Can you now also delete lots of https://sunpy--5237.org.readthedocs.build/en/5237/generated/gallery/units_and_coordinates/AIA_limb_STEREO.html#sphx-glr-generated-gallery-units-and-coordinates-aia-limb-stereo-py?

Ha!  Definitely!  I had forgotten that example even existed...

sunpy/map/mapbase.py
ayshih: I should be able to link to that through a Sphinx reference, right?

sunpy/map/mapbase.py
ayshih: It isn't.  In full detail, the round-trip transformation is `map_world(2D)`->`map_world(3D)`->`axes_world(3D)`->`axes_world(2D)`->`axes_pixel`->`axes_world(2D)`->`axes_world(3D)`->`map_world(3D)`->`map_world(2D)`.

sunpy/map/mapbase.py
ayshih: See https://github.com/sunpy/sunpy/pull/5237#discussion_r618488272

The load-bearing step is `axes_world(2D)`->`axes_world(3D)`.  If the initial point is on the far side of the Sun for the axes's observer, this step will not be the inverse of the earlier `axes_world(3D)`->`axes_world(2D)`, and thus the round-trip transformation fails.

sunpy/map/mapbase.py
dstansby: Ah, so to make sure I understand this essentially:
- Projects all the 3D limb coordinates on to the axes-observer-facing side of the Sun
- Checks if any of the projected coordinates have moved, and if so labels them as far-side

I presume there's good reasons for doing this instead of transforming to axes-HPC, and testing `transformed_coord.distance >= coord.rsun`? I think this way would be easier to follow in the code, and possibly involve less transforms so be better performance, but may have other issues I haven't thought of.

ayshih:
I'm kicking this back to draft.  Spurred by @dstansby 's questions, I think I can refactor this so that it performs better (speed and accuracy).

ayshih:
@dstansby Take a look at this new implementation

sunpy/map/mapbase.py

sunpy/map/mapbase.py
ayshih: It's actually both points and segments.  (There's a `[:-1]` in the later line.)  Because I use the word "segments" earlier in the docstring, I'd rather say "points" here.

sunpy/map/mapbase.py
ayshih: Moot

sunpy/map/mapbase.py
ayshih: Moot

sunpy/map/mapbase.py
ayshih: Moot

sunpy/map/mapbase.py
ayshih: I don't object to this change, but I'm going to kick this to a future PR.  It breaks a handful of figure tests (particularly with `CompositeMap`), and I don't want to deal with that all in this PR.

ayshih:
For fun:
download (20)
The red limb is for a different observer at ~1 AU.  The green limb is if that same observer is at 1/100th the distance (height of ~1 solar radius).

sunpy/map/mapbase.py

ayshih:
> My big question is, does this work with CAR projections (ie. synoptic maps)? It would be amazing if it did...

Works now!
hgc

dstansby:
> > My big question is, does this work with CAR projections (ie. synoptic maps)? It would be amazing if it did...
> 
> Works now!

This is amazing.

sunpy/map/mapbase.py
dstansby: Ah I forgot for a closed path `npoints == nsegments`.

dstansby (approved):

nabobalis:
Needs a rebase to solve the file conflict.

sunpy/map/mapbase.py
```diff
@@ -1978,50 +1978,130 @@ def draw_grid(self, axes=None, grid_spacing: u.deg = 15*u.deg, annotate=True, **
...
+
+        # Determine which points are visible
+        if hasattr(axes_frame, 'observer'):
+            # The reference distance is the distance to the limb for the axes observer
+            rsun = getattr(axes_frame, 'rsun', self.rsun_meters)
+            reference_distance = np.sqrt(axes_frame.observer.radius**2 - rsun**2)
+            is_visible = limb_in_axes.spherical.distance <= reference_distance
+        else:
+            # If the axes has no observer, the entire limb is considered visible
+            is_visible = np.ones_like(limb_in_axes.spherical.distance, np.bool)
```
dstansby, line 2082:
```suggestion
            is_visible = np.ones_like(limb_in_axes.spherical.distance.value, np.bool)
```
I had to make this small change to make this PR work on a Carrington map. Without it `is_visible` is a boolean array with dimensions `'m``, so the `~is_visible` call below errors out.

ayshih: Thanks for catching this bug.  I should have included `subok=False`, which should be slightly better performing that your suggestion.

However, I'm a little confused.  If you're using Carrington coordinates for your WCSAxes, it shouldn't even fall into this code branch.  Are you using something other than Carrington coordinates?

dstansby: I think it's because I manually created a header without observer information using the `header_helper`. Full code is [here](https://solar-orbiter-python.readthedocs.io/en/latest/_examples_build/remote_sensing/eui_and_aia.html#sphx-glr-examples-build-remote-sensing-eui-and-aia-py) if it's helpful.

dstansby (approved):

sunpy/map/mapbase.py
ayshih: Ah, yes, your WCS is in Stonyhurst heliographic coordinates, not in Carrington heliographic coordinates, so no observer.

ayshih: Added a figure test

ayshih:
I've now also fixed the closing of the limb across the longitude wrap of heliographic maps.  That is, there will no longer be a horizontal line connecting the limb at the left edge (smallest longitude) with the limb at the right edge (greatest longitude).

Someone else review this PR!

sunpy/map/mapbase.py
```diff
@@ -1978,24 +1978,44 @@ def draw_grid(self, axes=None, grid_spacing: u.deg = 15*u.deg, annotate=True, **
...
 
         Returns
         -------
-        circ: list
-            A list containing the `~matplotlib.patches.Circle` object that
-            has been added to the axes.
+        visible : `~matplotlib.patches.Polygon`
+            The patch added to the axes for the visible part of the limb (i.e., the
+            "near" side of the Sun).
+        hidden : `~matplotlib.patches.Polygon`
```
Cadair, line 2003:
```suggestion
        hidden : `~matplotlib.patches.Polygon` or None
```

sunpy/map/mapbase.py
```diff
@@ -1978,24 +1978,44 @@ def draw_grid(self, axes=None, grid_spacing: u.deg = 15*u.deg, annotate=True, **
...
+            The patch added to the axes for the hidden part of the limb (i.e., the
+            "far" side of the Sun).
 
         Notes
         -----
-        Keyword arguments are passed onto `matplotlib.patches.Circle`.
+        Keyword arguments are passed onto the patches.
+
+        If the limb is a true circle, ``visible`` will instead be
+        `~matplotlib.patches.Circle` and ``hidden`` will be ``None``.
```
Cadair, line 2012: I assume Circle is a subclass of Polygon?

Cadair (approved):
This is lovely. :heart_eyes:

sunpy/map/mapbase.py
ayshih: Done

dstansby (approved):
Doc build warnings are unrelated (can't find a couple of intersphinx inventories)

PR #5241, opened 2021-04-22
Save original copy of metadata

dstansby:
`sunpy.util.MetaDict` now saves a copy of it's initial input metadata. This original copy, along with new helper methods, allows one to query any changes, additions, or removals from the metadata. See the included example for an example.

xref https://github.com/sunpy/sunpy/issues/4194; this goes a little way to making progress on that issue.

### DIFF
--- a/examples/map/map_metadata_modification.py
+++ b/examples/map/map_metadata_modification.py
@@ -1,0 +1,40 @@
+"""
+=========================
+Map metadata modification
+=========================
+
+How to query map metadata for changes.
+
+sunpy has a series of map sources that can fix common issues with the original metadata
+stored in FITS files. A copy of the original (unaltered) metadata is stored, so
+any changes that sunpy (or the user) subsequently makes to the metadata can be
+easily queried.
+
+In the example below, we load a HMI sample image, and query the metadata for
+any added, removed, or modified items.
+"""
+import astropy.units as u
+
+import sunpy.map
+from sunpy.data.sample import HMI_LOS_IMAGE
+
+###############################################################################
+# Start by creating a map.
+hmimap = sunpy.map.Map(HMI_LOS_IMAGE)
+
+###############################################################################
+# Now query the ``.meta`` attribute for any changes. We can see that nothing
+# has been added or removed, but the 'bunit' key has been updated from "Gauss"
+# to "G" to make it FITS standard compliant.
+print("Added items:", hmimap.meta.added_items)
+print("Removed items:", hmimap.meta.removed_items)
+print("Modified items:", hmimap.meta.modified_items)
+
+###############################################################################
+# If we modify the map in a way that updates the metadata, these properties
+# allow use to easily see what has changed. As an exmaple, lets rotate the map
+# by 90 degrees, and see what has been updated.
+hmimap = hmimap.rotate(90 * u.deg)
+print("Added items:", hmimap.meta.added_items)
+print("Removed items:", hmimap.meta.removed_items)
+print("Modified items:", hmimap.meta.modified_items)
--- a/sunpy/util/metadata.py
+++ b/sunpy/util/metadata.py
@@ -2,10 +2,13 @@
 This module provides a generalized dictionary class that deals with header
 parsing, normalization, and maintaining coherence between keys and keycomments.
 """
-from collections import OrderedDict
+from collections import OrderedDict, namedtuple
 from collections.abc import Mapping
 
 __all__ = ['MetaDict']
+
+ModifiedItem = namedtuple('ModifiedItem', ['original', 'current'])
+ModifiedItem.__repr__ = lambda t: f"(original={t.original}, current={t.current})"
 
 
 class MetaDict(OrderedDict):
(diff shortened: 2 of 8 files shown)


Cadair (approved):
Other than the fact codecov is sad I think this looks good.

Cadair:
Oh also a real doc fail: https://dev.azure.com/sunpy/sunpy/_build/results?buildId=16117&view=logs&j=090c8010-07b3-5084-af5f-2960d012d632&t=d581a89c-705b-5160-02b7-5f27e289c243&l=283

dstansby:
The doc build fail seems to be a remote issue now. I had to do a bit of contortion to add custom doc for `.meta`, but I think this is the best way.

Cadair (approved):

sunpy/util/metadata.py
```diff
@@ -48,11 +57,46 @@ def __init__(self, *args):
...
+
+    @property
+    def modified_items(self):
+        """
+        Items modified since initialisation.
+
+        This is a mapping from ``key`` to ``[initial_value, current_value]``.
+        """
+        keys = set(self).intersection(set(self.original_meta))
+        return {k: [self.original_meta[k], self[k]] for k in keys if
```
ayshih, line 97: I strongly recommend making these named tuples

examples/map/map_metadata_modification.py
```diff
@@ -0,0 +1,29 @@
...
+
+In the example below, we load a HMI sample image, and query the metadata for
+any added, removed, or modified items.
+"""
+import sunpy.map
+from sunpy.data.sample import HMI_LOS_IMAGE
+
+###############################################################################
+# Start by creating a map
+hmimap = sunpy.map.Map(HMI_LOS_IMAGE)
```
ayshih, line 21: `SWAP_LEVEL1_IMAGE` might be a better example since it has an added item.  (Although _why_ it adds `obsrvtry` I don't know...)

ayshih (commented):
This looks really neat!  My main concern is that you lose the record of the original metadata with a `.rotate()` call, and I haven't tracked down why.

Cadair:
> This looks really neat! My main concern is that you lose the record of the original metadata with a .rotate() call, and I haven't tracked down why.

something something deepcopy?

dstansby:
Thanks for catching `.rotate()` the problem was `.copy()` in the end (not `copy.copy` or `copy.deepcopy`, both of which I added tests for anyway). This also gave me the idea to use `.rotate()` in the example, which I think shows off all three new properties well.

dstansby:
Forgot to add, will update to namedtuples but don't have time right now.

examples/map/map_metadata_modification.py
dstansby: I stuck with this map, but rotated it which demonstrates all three new meta properties.

dstansby:
I'm trying to use a `namedtuple`, but am slightly confused what the best way to do so is. I've tried doing {key: namedtuple} so printing `modified_items` looks like
```python
Modified items: {'bunit': ModifiedValue(initial_value='Gauss', current_value='G'), 'crval1': ModifiedValue(initial_value=-4.234319825764032, current_value=-4.234319825656712), 'crpix2': ModifiedValue(initial_value=512.5, current_value=512.4999999999261), 'crpix1': ModifiedValue(initial_value=512.5, current_value=512.4999999999914)}
```
Is this the right approach?

ayshih:
That's essentially what I had in mind, but I'd tweak it to be shorter because the context is clear:

- `initial_value` -> `original`
- `current_value` -> `current`
- Replace `ModifiedValue.__repr__()` so that the string `ModifiedValue` isn't displayed (e.g., set to `lambda t: f"(original={t.original}, current={t.current})`)

examples/map/map_metadata_modification.py
```diff
@@ -0,0 +1,39 @@
...
+
+In the example below, we load a HMI sample image, and query the metadata for
+any added, removed, or modified items.
+"""
+import astropy.units as u
+import sunpy.map
+from sunpy.data.sample import HMI_LOS_IMAGE
+
+###############################################################################
+# Start by creating a map
```
nabobalis, line 21:
```suggestion
# Start by creating a map.
```

sunpy/util/metadata.py
```diff
@@ -21,18 +21,27 @@ class MetaDict(OrderedDict):
     `MetaDict`, it will also be removed from the keycomments dictionary.
     Additionally, any extraneous keycomments will be removed when the
     `MetaDict` is instantiated.
+
+    Parameters
+    ----------
+    save_original : bool
```
nabobalis, line 27:
```suggestion
    save_original : bool, optional
```

sunpy/util/metadata.py
```diff
@@ -21,18 +21,27 @@ class MetaDict(OrderedDict):
...
+        If `True`, a copy of the original metadata will be saved to the
```
nabobalis, line 28:
```suggestion
        If `True` (the default), a copy of the original metadata will be saved to the
```

examples/map/map_metadata_modification.py
```diff
@@ -0,0 +1,39 @@
...
+sunpy has a series of map sources that can fix common issues with the original metadata
+stored in FITS files. A copy of the original (unaltered) metadata is stored, so
+any changes that sunpy (or the user) subsequently makes to the metadata can be
+easily queried.
+
+In the example below, we load a HMI sample image, and query the metadata for
+any added, removed, or modified items.
+"""
+import astropy.units as u
+import sunpy.map
```
nabobalis, line 17:
```suggestion

import sunpy.map
```

nabobalis:
I think that png image should not be there? (We should add that to the git ignore.

dstansby:
Thanks for catching; I've force pushed it away, and added it to .gitignore.

nabobalis (approved):

PR #5244, opened 2021-04-22
WCSAxes warnings for 3.0

dstansby:
This is currently a draft whilst I check that comparing WCSs is working fine. When I've done that the error will be converted to a warning.

### DIFF
--- a/changelog/5244.removal.rst
+++ b/changelog/5244.removal.rst
@@ -1,4 +1,4 @@
-Plotting a `~sunpy.map.GenericMap` on an `~matplotlib.axes.Axes` that is not
+Using `~sunpy.map.GenericMap` plotting methods on an `~matplotlib.axes.Axes` that is not
 a `~astropy.visualization.wcsaxes.WCSAxes` is deprecated. This previously
 raised a warning, but is now formally deprecated, and will raise an error in
 sunpy 3.1.
--- a/changelog/5244.removal.rst
+++ b/changelog/5244.removal.rst
@@ -1,0 +1,4 @@
+Plotting a `~sunpy.map.GenericMap` on an `~matplotlib.axes.Axes` that is not
+a `~astropy.visualization.wcsaxes.WCSAxes` is deprecated. This previously
+raised a warning, but is now formally deprecated, and will raise an error in
+sunpy 3.1.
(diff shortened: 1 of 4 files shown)


sunpy/map/mapbase.py
```diff
@@ -2431,25 +2431,33 @@ def _check_axes(self, axes, allow_non_wcsaxes=False):
...
                 if not np.array_equal(self.rotation_matrix, np.identity(2)):
                     warnings.warn("The axes of this map are not aligned to the pixel grid. "
                                   "Plot axes may be incorrect.",
                                   SunpyUserWarning)
                 warnings.warn("WCSAxes not being used as the axes object for this plot. "
-                              "Plots may have unexpected behaviour. To fix this pass "
-                              "'projection=map' when creating the axes",
-                              SunpyUserWarning)
+                              "Support for this is deprecated, and will be removed in sunpy 3.1. "
+                              "To fix this pass 'projection=map' when creating the axes.",
```
ayshih, line 2453: Can you reword this?  I feel like a user might literally type `projection=map` and be confused why passing the function `map` doesn't achieve the desired result.

Cadair:
Nice! Cleaner than I expected :smile:

dstansby:
Okay, this is ready for review.

The only gallery example that will raise the new "WCS not the same" warning is [Fading between maps](https://docs.sunpy.org/en/stable/generated/gallery/plotting/fading_between_maps.html#sphx-glr-generated-gallery-plotting-fading-between-maps-py), which is expected as two different maps are being plotted on top of each other.

There is another example that shouldn't raise the warning, but that will hopefully be fixed in https://github.com/sunpy/sunpy/pull/5246

changelog/5244.trivial.rst
```diff
@@ -0,0 +1,6 @@
+Plotting a `~sunpy.map.GenericMap` on a
```
ayshih, line 1:
```suggestion
Calling :meth:`sunpy.map.GenericMap.plot` on a
```

changelog/5244.removal.rst
```diff
@@ -0,0 +1,4 @@
+Plotting a `~sunpy.map.GenericMap` on an `~matplotlib.axes.Axes` that is not
```
ayshih, line 1:
```suggestion
Using `~sunpy.map.GenericMap` plotting methods on an `~matplotlib.axes.Axes` that is not
```

ayshih (commented):
Both changelog entries start with "Plotting a GenericMap on...", but one entry applies just to `plot()` and one applies to all the plotting methods.

Cadair (approved):
Awesome, thanks.

nabobalis (approved):

Issue #3387, opened 2019-10-02
Remove the default observer of Earth for a Map without a defined observer?


ayshih:
(This is where to continue discussion from #3118 and other places.)

Currently, Maps without a defined observer will use a default observer of Earth.  (Curiously, this default doesn't rely on any default-observer behavior in the coordinates machinery, so the proposed change in #3386 isn't relevant.)  Should this be changed?

wtbarnes:
I'm guessing this choice has not been very contentious given the lack of discussion here in the last 6.5 years.

What would be the proposed alternative? Setting `None`? In the past, Earth was a sensible default, but as the number of observatories off the Sun-Earth line grows, it may make less and less sense. Maybe it would be more feasible to set this to `None` and have a default assumption of Earth for individual sources if they are known to be missing their observer keywords?

ayshih:
Yes.  It is already possible through private API to specify a source-specific default observer, so I'd remove the generic default observer.

PR #5246, opened 2021-04-23
Only set WCS observer if map frame has observer attribute

dstansby:
TL;DR: I think for a Stonyhurst frame, observer information shouldn't be set on a WCS.

Story time!

As part of https://github.com/sunpy/sunpy/pull/5244, warnings will be raised if plotting when `map.wcs != axes.wcs`. More precisely, this comparison is done with the line `axes.wcs.wcs.compare(self.wcs.wcs, tolerance=0.01)`. This will compare every WCS keyword, and evaluate true iff all keywords are identical to within an absolute tolerance of 0.01.

In the `reprojection_aia_euvi_mosaic.py` example, the following happens:

1. Create a FITS header in a Stonyhurst frame using `header_helper`. Because the Stonyhurst frame does (in general) not have an observer property, no observer keywords are set in the header. I shall call the WCS from this `wcs_header`.
2. Using `reproject`, create new maps from this header. I shall call the WCS from this `wcs_map`.
3. Create an axes with `wcs_header`
4. Plot the map on this axes

However, `map.wcs` sees there is no observer information in the header, and falls back to assuming the observer is the Earth, and sets the observer keywords in `wcs_map`. This leads to `wcs_header.compare(map_wcs, tolerance=0.01)` failing.

I *think* the solution to this is to only set the observer information in `GenericMap.wcs` if the coordinate frame of the map has an observer attribute. @ayshih should check this.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -552,24 +552,21 @@ class GenericMap(NDData):
         if w2.wcs.ctype[1].lower() in ("solar-y", "solar_y"):
             w2.wcs.ctype[1] = 'HPLT-TAN'
 
-        # Set observer coordinate information
-        #
-        # Clear all the aux information that was set earlier. This is to avoid
-        # issues with maps that store multiple observer coordinate keywords.
-        # Note that we have to create a new WCS as it's not possible to modify
-        # wcs.wcs.aux in place.
-        header = w2.to_header()
-        for kw in ['crln_obs', 'dsun_obs', 'hgln_obs', 'hglt_obs']:
-            header.pop(kw, None)
-        w2 = astropy.wcs.WCS(header)
+        if hasattr(astropy.wcs.utils.wcs_to_celestial_frame(w2), 'observer'):
+            # Set observer coordinate information
+            #
+            # Clear all the aux information that was set earlier. This is to avoid
+            # issues with maps that store multiple observer coordinate keywords.
+            # Note that we have to create a new WCS as it's not possible to modify
+            # wcs.wcs.aux in place.
+            header = w2.to_header()
+            for kw in ['crln_obs', 'dsun_obs', 'hgln_obs', 'hglt_obs']:
+                header.pop(kw, None)
+            w2 = astropy.wcs.WCS(header)
 
-        # Get observer coord, and transform if needed
-        obs_coord = self.observer_coordinate
-        if not isinstance(obs_coord.frame, (HeliographicStonyhurst, HeliographicCarrington)):
-            obs_coord = obs_coord.transform_to(HeliographicStonyhurst(obstime=self.date,
-                                                                      rsun=self.rsun_meters))
-
-        sunpy.coordinates.wcs_utils._set_wcs_aux_obs_coord(w2, obs_coord)
+            # Get observer coord, and set the aux information
+            obs_coord = self.observer_coordinate
+            sunpy.coordinates.wcs_utils._set_wcs_aux_obs_coord(w2, obs_coord)
 
         # Validate the WCS here.
         w2.wcs.set()
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -25,7 +25,7 @@ import sunpy.coordinates
 import sunpy.data.test
 import sunpy.map
 import sunpy.sun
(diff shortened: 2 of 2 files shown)


ayshih (commented):
I'd like to slightly expand the scope of this PR.  Ideally, `GenericMap.observer_coordinate` should really be returning `None` if the map's coordinate frame is not an observer-based frame.  However, it's a bit tricky to fix in our current implementation because `GenericMap.observer_coordinate` is used as if it were a low-level property: `GenericMap.wcs` calls `GenericMap.observer_coordinate` (and `GenericMap.coordinate_frame` calls `GenericMap.wcs`).

I suggest:

- Rename the existing `observer_coordinate` property to some private name (and have the `wcs` property continue to use it)
- Create a new, high-level `observer_coordinate` property, which would simply be `getattr(self.coordinate_frame, 'observer', None)`

Thoughts?

dstansby:
I had a bit of a think about the property call stack, and in the spirit of https://github.com/sunpy/sunpy/pull/5199 it sounds like we should be aiming for a world where

1. Map properties are drawn directly from the metadata
2. The `.wcs` should be constructed from these map properties (not directly from the metadata)

`.observer_coordinate` currently abides by 1., so I am reluctant to change it to being drawn from `.wcs` (via. `.coordinate_frame`). The issue is it defaults out to `get_earth()` instead of `None`, so by inspecting `.observer_coordinate` there is no way `.wcs` can currently know whether to set the `.observer_coordinate` or not.

I think the long term solution is therefore to make `.observer_coordinate` return `None` instead of `get_earth()`, probably after a deprecation period to avoid users tripping up.

Cadair:
I agree with what @dstansby said, I would rather keep the properties as the canonical metadata extraction point.

We could make the observer return conditional on the frame type though, i.e. return `None` by default if the frame is Stonyhurst?

dstansby:
A slight aside, but relevant for this discussion: given `HeliographicCarrington` has an `observer` to take into account light travel times, why doesn't `HeliographicStonyhurst`?

ayshih:
> I agree with what @dstansby said, I would rather keep the properties as the canonical metadata extraction point.

They're _all_ properties.  How do you tell which properties are the "canonical" ones, and which properties can depend on other properties?  That is, why does `.observer_coordinate` have to be property to draw directly from metadata?

> I think the long term solution is therefore to make `.observer_coordinate` return `None` instead of `get_earth()`, probably after a deprecation period to avoid users tripping up.

xref #3387

ayshih:
> A slight aside, but relevant for this discussion: given `HeliographicCarrington` has an `observer` to take into account light travel times, why doesn't `HeliographicStonyhurst`?

Reasons™

`HeliographicCarrington` (HGC) is intended for features on the surface of the Sun.  In that situation, the light travel time is well defined, as is the motion of the solar feature over the light travel time (albeit as a rigid rotator instead of a differential rotator).  This allows for the coordinate-feature correspondence to be maintained for observers at different Sun-observer distances.

`HeliographicStonyhurst` (HGS) is a general coordinate system, used for the location of any type of object anywhere in the heliosphere (or beyond), with no attempt to correct for light travel time.  This means that it is technically inaccurate to overlay HGS maps of the solar surface from two different observers, but the inaccuracy will be small if their Sun-observer distances are very similar (e.g., ~1 AU).  When combining AIA images with EUI images, it'd better to reproject to HGC.

dstansby:
I'm going to put aside large scale design discussion for now, and write down the implementation loop currently spinning in my head...

- To know if a coordinate frame has an observer, we want to do `hasattr(frame, 'observer')`
- To get `frame` we use `astropy.wcs.utils.wcs_to_celestial_frame(wcs)`
- which means we need to construct `wcs`
- which means we need to know if the map coordinate frame has an observer (so we can set the observer keywords on the `wcs`)

If anyone can suggest where this vicious cycle can be disrupted I would be very grateful (and apologies if there is already a solution above somewhere)

ayshih:
As I was fixing a different bug with Stonyhurst heliographic maps, I realized it was very related to the bug driving this PR.  So, I built PR #5478 off of this one.

dstansby:
I've kind of lost track of this PR, so if you want to take over (or just make it part of another PR) then that would be very welcome.

ayshih:
Merged in with #5478

PR #3109, opened 2019-05-15
Remove basic_plot from Map.peek

Cadair:
:fire: :fire:

PR #5255, opened 2021-04-26
Leverage `pcolormesh()` for GenericMap/CompositeMap plotting

ayshih:
This PR makes two significant enhancements:

- `GenericMap.plot()` now has an optional keyword `autoalign`.  If `autoalign=True`, the method will use `pcolormesh()` instead of `imshow()` to plot the image.  When plotting a single map onto its own WCS, there's no benefit to doing this.  However, this enables plotting a map onto a _different_ WCS, without necessarily needing to use `reproject`.  The simpler the transformation (e.g., just a difference in rotation angle), the better.
- `CompositeMap.plot()` now calls `GenericMap.plot()` and `GenericMap.draw_contours()` instead of having its own plotting code.  It sets `autoalign=True` so that image alignment is automatically taken care of.  This approach to fixing `CompositeMap` plotting is much better than the one in #5208


ayshih:
I changed the keyword name from `use_mesh` to `different_wcs`

ayshih:
If you wrap the plotting code with:
```python
with Helioprojective.assume_spherical_screen(m1s.observer_coordinate):
    ...
```
you get:
download (22)
which is mostly better.

ayshih:
Contours version:
```python
c12.set_levels(1, np.arange(0, 100, 10) << u.percent)

fig = plt.figure(figsize=(12, 5), dpi=100)

ax = fig.add_subplot(121, projection=m1s)
c12.plot(title='Composite using 171 WCSAxes')
ax = fig.add_subplot(122, projection=m2s)
c12.plot(title='Composite using 193 WCSAxes')

plt.show()
```
download (23)

ayshih:
Starting from [this example](https://docs.sunpy.org/en/stable/generated/gallery/map_transformations/reprojection_spherical_screen.html), here's a comparison between using `reproject` and using `pcolormesh`:
```python
fig = plt.figure(figsize=(10, 5))

ax1 = fig.add_subplot(121, projection=outmap_screen_off_disk)
outmap_screen_off_disk.plot(title='Using reproject')

with Helioprojective.assume_spherical_screen(aia_map.observer_coordinate,
                                             only_off_disk=True):
    ax2 = fig.add_subplot(122, projection=outmap_screen_off_disk)
    aia_map.plot(different_wcs=True, title='Using pcolormesh')  # Note that this is *not* the reprojected map!

    plt.show()
```
download (25)

ayshih:
Trying to fix `basic_plot` functionality in `CompositeMap.peek()` seems pointless, so I have removed it.  It was removed from `GenericMap.peek()` two years ago (#3109).

Cadair:
I like this! It's cool that all our plotting will handle different data at last.

I have two main thoughts:

1) How bad is the performance here?
2) What would be the disadvantages of auto-detecting if the axes on which the map was being plotted wasn't the same WCS? We could then either drop the `different_wcs` kwarg or make its default `'auto'`?

ayshih:
> 1. How bad is the performance here?

You can definitely feel the performance hit of what is effectively a reprojection, but qualitatively it doesn't feel noticeably worse than using `reproject` or `.rotate()`.

> 2. What would be the disadvantages of auto-detecting if the axes on which the map was being plotted wasn't the same WCS? We could then either drop the `different_wcs` kwarg or make its default `'auto'`?

Implementation-wise, there's no real disadvantage.  However, I think it'd be prudent to put off auto-detection until 3.1, given that this functionality has yet to be put through its paces by the user base.

Cadair:
> Implementation-wise, there's no real disadvantage. However, I think it'd be prudent to put off auto-detection until 3.1, given that this functionality has yet to be put through its paces by the user base.

With the feedback we get about people saying that they expect this to work "out of the box" makes me wonder why we wouldn't do this? Will people even know to use the kwarg? I am happy to leave it a release if you think it's prudent. :shrug:

ayshih:
> Will people even know to use the kwarg?

The user will get a warning about a mismatched WCS, and the warning suggests using the `different_wcs` kwarg if the mismatch is intentional.  So, it depends on whether users read warning messages. =)

> I am happy to leave it a release if you think it's prudent. 🤷

@dstansby or anyone else: Thoughts?

dstansby:
> The user will get a warning about a mismatched WCS, and the warning suggests using the `different_wcs` kwarg if the mismatch is intentional. So, it depends on whether users read warning messages. =)

I have both annecdotal and recent twitter evidence that sunpy emits so many warnings that some users filter out all warnings. I have a slight preference for doing the 'correct' thing (turning it on by default) and not raising a warning.

> You can definitely feel the performance hit of what is effectively a reprojection, but qualitatively it doesn't feel noticeably worse than using reproject or .rotate().

Is the performance hit a one off when the figure is initially rendered, or is it a continuous performance hit every time the user pans/zooms the axes?

ayshih:
> Is the performance hit a one off when the figure is initially rendered, or is it a continuous performance hit every time the user pans/zooms the axes?

Every time

ayshih:
From the dev meeting:

- Won't be the default behavior for the 3.0 release
- Change the name of the keyword from `different_wcs` to `autoalign`, plus a hook for future expandability (e.g., using `reproject` instead of `pcolormesh`)
- Add a gallery example

ayshih:
Test failure is unrelated.  Ready for review!

changelog/5255.breaking.1.rst
```diff
@@ -0,0 +1 @@
+:meth:`sunpy.map.CompositeMap.plot` now internally calls :meth:`sunpy.map.GenericMap.plot` and :meth:`sunpy.map.GenericMap.draw_contours`, which may affect the plot output of existing user code.
```
dstansby, line 1: Would be good to indicate how big these effects are (ie. are they minor changes at a pixel level that users don't need to worry about, or major changes)

examples/map_transformations/autoalign_aia_hmi.py
```diff
@@ -0,0 +1,66 @@
...
+# "upside down" relative to the AIA image.
+
+fig = plt.figure(figsize=(12, 5))
+ax1 = fig.add_subplot(1, 2, 1, projection=map_aia)
+map_aia.plot(axes=ax1, clip_interval=(1, 99.9)*u.percent)
+ax2 = fig.add_subplot(1, 2, 2, projection=map_hmi)
+map_hmi.plot(axes=ax2)
+
+######################################################################
+# Set ``autoalign=True`` allows plotting the HMI image onto axes
```
dstansby, line 41:
```suggestion
# Setting ``autoalign=True`` allows plotting the HMI image onto axes
```

sunpy/map/compositemap.py
```diff
@@ -342,7 +341,8 @@ def draw_grid(self, index=None, axes=None, grid_spacing: u.deg = 20*u.deg, **kwa
 
     def plot(self, axes=None, annotate=True,
              title="SunPy Composite Plot", **matplot_args):
-        """Plots the composite map object using matplotlib
+        """Plots the composite map object by calling :meth:`~sunpy.map.GenericMap.plot`
+        or :meth:`~sunpy.map.GenericMap.draw_contours`.
```
dstansby, line 345: What actually decides whether the map is plotted or contoured?

sunpy/map/compositemap.py
```diff
@@ -384,40 +390,28 @@ def plot(self, axes=None, annotate=True,
...
-
-                # Check for the presence of masked map data
-                if m.mask is None:
-                    ret.append(axes.imshow(m.data, **params))
-                else:
-                    ret.append(axes.imshow(np.ma.array(np.asarray(m.data), mask=m.mask), **params))
+                # Check if any linewidth argument is provided, if so, then delete it from params.
+                for item in ['linewidth', 'linewidths', 'lw']:
+                    if item in matplot_args:
+                        del params[item]
```
dstansby, line 405: The silent deleting of these args could confuse a user trying to specify them - perhaps add a warning, or if we don't want more warnings a note in the docstring should do.

sunpy/map/compositemap.py
```diff
@@ -430,8 +424,7 @@ def plot(self, axes=None, annotate=True,
         return ret
 
     @peek_show
-    def peek(self, colorbar=True, basic_plot=False, draw_limb=True,
```
dstansby, line 433: In the interests of not breaking API without a warning, should we deprecate `basic_plot` in 3.0 and then remove in 3.1?

sunpy/map/mapbase.py
```diff
@@ -2460,9 +2487,30 @@ def plot(self, annotate=True, axes=None, title=True,
...
+        else:
+            data = np.ma.array(np.asarray(self.data), mask=self.mask)
+
+        if autoalign == 'pcolormesh':
+            axes.set_aspect(getattr(imshow_args, 'aspect', 1))
+
+            # Remove imshow keyword arguments that are not accepted by pcolormesh
+            for item in ['aspect', 'extent', 'interpolation', 'origin']:
+                if item in imshow_args:
+                    del imshow_args[item]
```
dstansby, line 2500: Similar to above, do we want to warn if these keyword arguments are removed?

sunpy/map/mapbase.py
```diff
@@ -2364,11 +2364,11 @@ def peek(self, draw_limb=False, draw_grid=False,
         return figure
 
     @u.quantity_input
-    def plot(self, annotate=True, axes=None, title=True,
+    def plot(self, annotate=True, axes=None, title=True, autoalign=False,
```
dstansby, line 2367: Should this be set to `None` by default, which doesn't autoalign and raises a warning (on different WCS), and then if the user manually specifies `False` the warning can be disabled as they've made a deliberate choice to not align the maps.

sunpy/map/mapbase.py
```diff
@@ -2560,7 +2608,8 @@ def _check_axes(self, axes, allow_non_wcsaxes=False, warn_different_wcs=False):
                                 "to this map when creating the axes.")
         elif warn_different_wcs and not axes.wcs.wcs.compare(self.wcs.wcs, tolerance=0.01):
             warnings.warn('The map world coordinate system (WCS) is different from the axes WCS. '
-                          'The map data axes may not correctly align with the coordinate axes.',
+                          'The map data axes may not correctly align with the coordinate axes. '
+                          'If this difference in WCS is intended, specify `different_wcs=True`.',
```
dstansby, line 2612:
```suggestion
                          'If this difference in WCS is intended, specify `autoalign=True`.',
```
Even with this fix I'm a bit confused by this sentence. Changing the message to `autoalign=False` in tandem with my comment above on setting `autoalign=None` by default might work.

examples/map_transformations/autoalign_aia_hmi.py
```diff
@@ -0,0 +1,66 @@
+"""
+==============================
+Auto-Aligning AIA and HMI Data
```
Cadair, line 3: I think we should specify in the title that this is just a plotting thing or else it's a little oversold on the examples index.
```suggestion
Auto-Aligning AIA and HMI Data in Plots
```

examples/map_transformations/autoalign_aia_hmi.py
```diff
@@ -0,0 +1,66 @@
...
+
+fig = plt.figure(figsize=(12, 5))
+ax1 = fig.add_subplot(1, 2, 1, projection=map_aia)
+map_aia.plot(axes=ax1, clip_interval=(1, 99.9)*u.percent)
+ax2 = fig.add_subplot(1, 2, 2, projection=map_hmi)
+map_hmi.plot(axes=ax2)
+
+######################################################################
+# Set ``autoalign=True`` allows plotting the HMI image onto axes
+# defined by the AIA reference frame.  We also need to manually set
```
Cadair, line 42: I think that we should emphasise somewhere that to make this work you have to set the projection to the aia map. It wasn't immediately obvious to me on a first read. Maybe that needs to be here, or maybe an inline comment in the code?

examples/map_transformations/autoalign_aia_hmi.py
```diff
@@ -0,0 +1,66 @@
...
+# the plot limits because Matplotlib gets confused by the off-disk
+# parts of the image.  Note that the HMI image now has the same
+# orientation as the AIA image.
+
+fig = plt.figure(figsize=(12, 5))
+ax1 = fig.add_subplot(1, 2, 1, projection=map_aia)
+map_aia.plot(axes=ax1, clip_interval=(1, 99.9)*u.percent)
+ax2 = fig.add_subplot(1, 2, 2, projection=map_aia)
+map_hmi.plot(axes=ax2, autoalign=True, title='HMI image in AIA reference frame')
+ax2.dataLim = ax1.dataLim
```
Cadair, line 52: Is setting `dataLim` the approved way of doing this? or would `ax2.axis(ax1.axis())` be better? I haven't ever seen this one before.

examples/map_transformations/autoalign_aia_hmi.py
```diff
@@ -0,0 +1,66 @@
...
+
+######################################################################
+# We can directly plot them over one another, by setting the
+# transparency of the HMI plot.
+
+fig = plt.figure()
+ax1 = fig.add_subplot(1, 1, 1, projection=map_aia)
+map_aia.plot(axes=ax1, clip_interval=(1, 99.9)*u.percent)
+map_hmi.plot(axes=ax1, autoalign=True, alpha=0.5)
+plt.title('HMI overlaid on AIA')
```
Cadair, line 62: Rather than cracking and using plt right at the end? :grin: 
```suggestion
ax1.set_title('HMI overlaid on AIA')
```

sunpy/map/mapbase.py
```diff
@@ -2460,9 +2487,30 @@ def plot(self, annotate=True, axes=None, title=True,
                 norm.vmax = imshow_args.pop('vmax')
 
         if self.mask is None:
-            ret = axes.imshow(self.data, **imshow_args)
+            data = self.data
+        else:
+            data = np.ma.array(np.asarray(self.data), mask=self.mask)
+
+        if autoalign == 'pcolormesh':
+            axes.set_aspect(getattr(imshow_args, 'aspect', 1))
```
Cadair, line 2495: I assume you mean this?
```suggestion
            axes.set_aspect(imshow_args.get('aspect', 1))
```

sunpy/map/mapbase.py
Cadair: How about this:
```suggestion
                          'To automatically transform the data to the coordinate axes, specify `autoalign=True`.',
```

sunpy/map/compositemap.py
Cadair: If I remember correctly you set it with a different part of the composite map API, i.e. set the type for each map in the collection.

sunpy/map/compositemap.py
Cadair: I would be happy with a note in the docstring.

sunpy/map/mapbase.py
Cadair: I am not really in favour of throwing warnings here.

dstansby: How about raising errors? From memory that's the approach `WCSAxes` takes in some similar situations. I'm not too fussed, but either way this should definitely be mentioned in the docstring.

changelog/5255.breaking.1.rst
ayshih: Impossible to say definitively.  For composite maps that are helioprojective and already co-oriented, the difference will be negligible.  For other composite maps, the changes could be drastic. (because the previous compositing was incorrect).

sunpy/map/compositemap.py
ayshih: If `.levels` is set for a given map, that map will be contours instead of an images

sunpy/map/compositemap.py
ayshih: The purpose of this deleting is so that someone can specify `.plot(linewidth=2)` and that `linewidth` argument applies only to the maps that are to be plotted as contours, because that's what a user wants.  You can't pass on the `linewidth` argument for the maps that are to be plotted as images because it does bad things.  (For imshow, it would error.  For pcolormesh, it adds an outline to each quadilateral.)

sunpy/map/mapbase.py
ayshih: All of these arguments do not exist for `pcolormesh()` compared to `imshow()`.
* I handle `aspect` separately earlier, so removing it here is fine.
* `origin` and `extent` don't matter.  We don't reasonably expect any user to set any of these, but we set them internally.  (Ideally, the code should be reversed: they shouldn't be in `plot_settings` and instead be _added_ by `plot()` if `imshow()` is going to be used.)
* `interpolation` is an exception, because a user might ask for that, but `pcolormesh()` doesn't support it.  We can raise a warning if a user specifically asks for an interpolation method other than `'none'` or `'nearest'` so they know that they won't be getting what they ask for.

sunpy/map/mapbase.py
ayshih: I'm not keen on this suggestion.  We didn't have a way to disable the warning prior to the PR, and I don't know why we'd want to provide users with a way to turn off the warning.  If a user really wants to disable a warning, there's always `filterwarnings()`.

sunpy/map/compositemap.py
ayshih: Yes, formally there should be a deprecation period, but it would take a bit of effort to get `basic_plot` working with these new internals.

@Cadair: remove outright or deprecate first?

sunpy/map/mapbase.py
ayshih: Ha!  Yeah...

sunpy/map/compositemap.py
Cadair: I vote just remove it. I don't think it's worth Albert's definition of "a bit of effort" to support this. I can't image many people are using it on compmap and we pulled it from GenericMap a while ago.

examples/map_transformations/autoalign_aia_hmi.py
ayshih: \<shrug\>  Your way is probably more approved, so I'll change it.

sunpy/map/compositemap.py
dstansby: Got it, sorry for the confusion.

sunpy/map/compositemap.py
dstansby: 👍

sunpy/map/mapbase.py
dstansby: 👍

sunpy/map/mapbase.py
dstansby: That looks good to me. I thought the original message was saying "If you are knowingly using imshow to plot these two maps, even though they have different WCSes...", so I think your suggestion is much clearer.

examples/map_transformations/autoalign_aia_hmi.py
ayshih: Fixed

sunpy/map/compositemap.py
ayshih: Added more to the docstring

sunpy/map/compositemap.py
ayshih: Note added to docstring

sunpy/map/mapbase.py
ayshih: Warning added for `interpolation`

sunpy/map/mapbase.py
ayshih: Accepted @Cadair's suggestion

examples/map_transformations/autoalign_aia_hmi.py
ayshih: Added sentence

examples/map_transformations/autoalign_aia_hmi.py
ayshih: Renamed

examples/map_transformations/autoalign_aia_hmi.py
ayshih: Done

Cadair (approved):

dstansby (approved):

PR #5257, opened 2021-04-27
MDIMap uses CONTENT keyword to identify measurement

### LINKED ISSUE #5256: MDIMap uses deprecated keyword to identify measurement
### Description

MDIMap uses the keyword `DPC_OBSR` to determine the measurement property. However, this keyword is deprecated and MDIMap therefore incorrectly labels all MDI data as continuum intensity data. Instead, MDIMap should behave the same as HMIMap and use the keyword `CONTENT` to determine the measurement property.

This is the relevant code from [MDIMap](https://docs.sunpy.org/en/stable/_modules/sunpy/map/sources/soho.html#MDIMap):
```python
    @property
    def measurement(self):
        """
        Returns the type of data in the map.
        """
        return "magnetogram" if self.meta.get('dpc_obsr', " ").find('Mag') != -1 else "continuum"
```
It should look the same as the code from [HMIMap](https://docs.sunpy.org/en/stable/_modules/sunpy/map/sources/sdo.html#HMIMap):

```python
    @property
    def measurement(self):
        """
        Returns the measurement type.
        """
        return self.meta.get('content', '').split(" ")[0].lower()
```

The MDI magnetic field data is labelled as continuum intensity data. The MDI magnetic field data should be labelled as magnetic field data.

### Steps to Reproduce

```python
import sunpy.map
import drms

# Query the data (using a registered e-mail address)
email = 'your@email.address'
c = drms.Client(debug=True, verbose=True, email=email)
mdi_query_string = 'mdi.fd_M_96m_lev182[2010.10.15_19:12:00]'

# Export the data
r = c.export(mdi_query_string, method='url', protocol='fits') 
fits_url_mdi = r.urls['url'][0]

# Plot the data
mdi_map = sunpy.map.Map(fits_url_mdi)  
mdi_map.plot()
```

### System Details

- SunPy Version: 2.1.4
- Astropy Version: 4.2.1
- Python Version: 3.9.4
- OS information: Mac OS X 10.15.7 (i386)

### Additional Notes
The MDI data was reprocessed at the end of the MDI mission, and the magnetic field data was ingested into JSOC in 2012. So MDI data downloaded before 2012 will indeed have the keyword `DPC_OBSR` and will not have the keyword `CONTENT`. However, I suggest we do not support versions of MDI data created before 2012.

Thanks to @pauljwright for pointing this out!

mbobra:
### Description

MDIMap uses the keyword CONTENT, instead of the deprecated keyword DPC_OBSR, to determine the measurement property.  This fix is not exactly right, because now the title says "MDI MDI Full Disk 96m Magnetogram MDI Full Disk 96m Magnetogram". But at least it does not say "MDI continuum continuum".

I also added an updated version of the MDI data to sunpy/sunpy/data/test/.  I'm not sure if I need to do anything else besides adding the fits files to the directory (e.g. modify the test suite).

Here is the code to download the magnetic field data:
```python

# Query the data (using a registered e-mail address)
jsoc_email = 'your@email.address'

results = Fido.search(
    a.Time('2010/10/15 19:00:00', '2010/10/15 19:13:00'),
    a.jsoc.Series.mdi_fd_m_96m_lev182,
    a.jsoc.Notify(jsoc_email),
)

downloaded_files = Fido.fetch(results)
```

And here is the code to download the continuum intensity data:
```python

# Query the data (using a registered e-mail address)
jsoc_email = 'your@email.address'

results = Fido.search(
    a.Time('2010/10/15 23:00:00', '2010/10/15 23:00:30'),
    a.jsoc.Series.mdi_fd_ic,
    a.jsoc.Notify(jsoc_email),
)

downloaded_files = Fido.fetch(results)
```

Fixes #5256


nabobalis:
The fits files will need to be resized to <100 kb ideally. 1MB is too large

wtbarnes:
I would lobby for using the plaintext header + dummy data approach that has been used in the EUI and HMI synoptic test data

nabobalis:
> I would lobby for using the plaintext header + dummy data approach that has been used in the EUI and HMI synoptic test data

Yeah lets do this.

mbobra:
I don't know how to make a plaintext header plus dummy data, so I downsampled the data by a factor of 16. Now `mdi.fd_Ic.20101015_230100_TAI.data.fits` is 20 kb and `mdi.fd_M_96m_lev182.20101015_191200_TAI.data.fits` is 33kb.

I also figured out the problem with the title. [Line 181](https://github.com/sunpy/sunpy/blob/416fe9734ebb4e42d48033c193c8a6174c5fa529/sunpy/map/sources/soho.py#L181) of map/sources/soho.py hacked the standard nickname convention in the [base_name attribute](https://github.com/sunpy/sunpy/blob/416fe9734ebb4e42d48033c193c8a6174c5fa529/sunpy/map/mapbase.py#L699) of sunpy/map/mapbase.py. So I removed it and I also defined the instrument property as MDI (instead of the detector).

But for some reason the checks are still failing 😭

nabobalis:
The MDI tests I think are done by filename, so you will need to update that.

dstansby (commented):
There's an example here of how to use the header + dummy data approach: https://github.com/sunpy/sunpy/pull/5210/files - if you have astropy installed, you should have a command line tool called `fitsheader` that will print the header of a fits file to a terminal for easy copy/pasting. I think it is best to go this route so we avoid adding more binary files to the repository.

Cadair (approved):

nabobalis:
Looks we have a bigger can of worms to deal with the mdi file changes.

mbobra:
@nabobalis @wtbarnes 

> We might need the nickname to be added back for the jp2 only if it is missing in the first place. Assuming that is what is the cause of the title missing from the jp2 test.

No, the nickname is still there -- it now uses the standard definition in mapbase.py rather than a non-standard definition from MDIMap.

> The other test failures are:
> FAILED ../../.tox/py39/lib/python3.9/site-packages/sunpy/database/tests/test_tables.py::test_entries_from_dir_recursively_false
> FAILED ../../.tox/py39/lib/python3.9/site-packages/sunpy/map/sources/tests/test_mdi_source.py::test_measurement
> FAILED ../../.tox/py39/lib/python3.9/site-packages/sunpy/database/tests/test_tables.py::test_entries_from_dir_recursively_true
> FAILED ../../.tox/py39/lib/python3.9/site-packages/sunpy/map/sources/tests/test_mdi_source.py::test_observer
> FAILED ../../.tox/py39/lib/python3.9/site-packages/sunpy/map/sources/tests/test_mdi_source.py::test_carrington
> 
> The database two are just we need to update the number of test files it finds.
Okay, I updated the numbers in test_tables.py.

> The others are in the mdi source file and I guess just need to be updated to account for header changes?
Yes, I updated the keyword information in the three functions (test_measurement, test_observer, and test_carrington) in test_mdi_source.py.

But somehow it is still failing, and I don't understand why.

nabobalis:
The remaining two:
```
________________________________ test_observer _________________________________
[gw2] linux -- Python 3.9.4 /home/vsts/work/1/s/.tox/py39/bin/python

mdi = <sunpy.map.sources.soho.MDIMap object at 0x7ff258867310>
SunPy Map
---------
Observatory:		 SOHO
Instrument:		 MDI
Det...2768, -32768, ..., -32768, -32768, -32768],
       [-32768, -32768, -32768, ..., -32768, -32768, -32768]], dtype=int16)

    def test_observer(mdi):
>       assert isinstance(mdi.observer_coordinate.frame.name, 'heliographic_stonyhurst')
E       TypeError: isinstance() arg 2 must be a type or tuple of types

../../.tox/py39/lib/python3.9/site-packages/sunpy/map/sources/tests/test_mdi_source.py:145: TypeError
```

I think this should be an assert that they are `==` and not and `isinstance`

```
__________________ test_map_sequence_animator_wcs_update_plot __________________
[gw1] linux -- Python 3.9.4 /home/vsts/work/1/s/.tox/py39/bin/python
ValueError: operands could not be broadcast together with shapes (64,64) (128,128) 

The above exception was the direct cause of the following exception:

test_map_sequence = <sunpy.map.mapsequence.MapSequence object at 0x7f4be20c5400>
MapSequence of 2 elements, with maps from AIAMap, MDIMap

    def test_map_sequence_animator_wcs_update_plot(test_map_sequence):
        map_animator = MapSequenceAnimator(test_map_sequence)
        map1 = map_animator.im.get_array()
        map_animator.updatefig(1, map_animator.im, map_animator.sliders[0]._slider)
        map2 = map_animator.im.get_array()
>       assert np.all(map1.data != map2.data)
E       DeprecationWarning: elementwise comparison failed; this will raise an error in the future.

```
This I think we need to change this to be `not np.testing.assert_allclose` maybe? Seems like a change in numpy is forcing us to make a change here.

mbobra:
Thanks for your help @nabobalis @wtbarnes @dstansby!

changelog/5257.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+Added MDI continuum intensity and line-of-sight magnetic field test data. These data were reprocessed at the end of the MDI mission and are available through JSOC. For a complete list of all the MDI observables, see `here <http://soi.stanford.edu/mdi/observables/>`__.
+Changed files with hardcoded MDI fits file names.
+Changed test_mdi_source to test on the instrument name, not the observable, and modified values for observer location, distance from SoHO to the Sun, and Carrington latitude and longitude to use keyword information, not hardcoded values.
+Changed MDIMap to use the CONTENT keyword to identify the measurement, similar to HMIMap, and removed the special-case nickname. Now MDIMap defaults to nickname in mapbase.py.
```
nabobalis, line 4:
```suggestion
Changed `MDIMap` to use the "CONTENT" keyword to identify the measurement, similar to `HMIMap`, and removed the special-case nickname. This fixes the broken title on plots.
```
I don't think we need to worry about the rest in the changelog entry.

nabobalis (approved):

PR #5261, opened 2021-04-28
Test data cleanup

### LINKED ISSUE #5247: Replace test image data with header strings + dummy data
We have a significant number of images in FITS files in `sunpy/data/test`. For testing purposes, we really only need the actual images for figure tests. In many cases, just having the header plus some dummy data with the correct shape is sufficient. As an example, see https://github.com/sunpy/sunpy/blob/9402b5e01cae60b0e61e02f389bef491d5db5828/sunpy/map/sources/tests/test_hmi_synoptic_source.py#L13.

I propose replacing those images not needed for the figure tests with a plaintext file (or FITS file) containing just the header information and then adding a helper function, call it `dummy_map_from_header` that takes in a path to these header files and returns you a test dummy map.

wtbarnes:
Fixes #5247 

This is an attempt to replace all of our test image data with test data comprised only of a plaintext header and a dummy data array that is generated on the fly. The exceptions to this are where these images are used for figure tests.

Currently, the `get_dummy_map_from_header` function returns a map constructed from this header. However, a better approach is probably to return a path to a temporary file that is generated on the fly from astropy (with as few fixes as possible to mimic the actual data products).

TODO:

- [ ] Find which files can be replaced (i.e. those not used by `@figure_test`
- [ ] Generate header files for all of these files and remove the FITS image files
- [ ] Fixture which autogenerates temporary file from data and header pair


nabobalis:
Just as a note for future: when you are done, can you copy and paste the pre commit changes from this PR https://github.com/sunpy/sunpy/pull/5197?

sunpy/data/test/__init__.py
```diff
@@ -60,3 +64,48 @@ def test_data_filenames():
...
+
+    See Also
+    --------
+    get_dummy_map_from_header
+    """
+    hdu = kwargs.get('hdu', 0)
+    with astropy.io.fits.open(image_path) as hdul:
+        header = hdul[hdu].header
+    with open(header_path, 'w') as f:
+        f.write(header.tostring())
```
dstansby, line 111: Would it be possible to write this with newlines between each key? I've found it helpful recently to be able to quickly scan FITS headers in the tests to see if they have e.g. any observer information, or if that info is in non-standard keys, which is much easier to do when the header isn't a single line of text.

wtbarnes:
List of files to replace with headers:

*  `EIT/*.fits

List of files to used for figure tests that CANNOT be replaced:

* `aia_171_level1.fits`
* `heliographic_phase_map.fits.gz`


These `.fits` files will be replaced with `.header` files

wtbarnes:
I've now run into a problem where I've really broken the map factory tests. 

We want to test different file name and glob patterns, but by removing the actual FITS files, I've made this impossible. I need to come up with some way to, on the fly, create a tmpdir of dummy FITS files generated on the fly from these header files. This should be possible with a pytest fixture, though I'm not completely sure how.

dstansby:
Perhaps we could just run the factory tests using the same FITS files that are going to stay for figure tests?

wtbarnes:
Things that are still broken:

- map factory tests
- database tests

As I noted above, there are several tests that ingest a directory of fits files (e.g. the map factory) and its not yet clear how to imitate that. I expect that all of these tests will fail for now.

sunpy/data/test/__init__.py
wtbarnes: I've written them out in a much more human-readable format

dstansby:
Closing in favour of https://github.com/sunpy/sunpy/pull/6065

PR #5276, opened 2021-04-30
Fix map source doc when source has no docstring

dstansby:
I have a map source in `pfsspy` that has no docstring (yes I know I should really add one...), which raised an error with the new doc tricks we do with map sources. This should fix it.


nabobalis:
I would rather we enforce a docstring on a map source but I am happy to approve and merge this.

nabobalis (approved):

PR #5279, opened 2021-04-30
ignore gallery warnings on release version of docs only

### LINKED ISSUE #1917: Tests which have hard coded numbers for web services fail often
There are various tests which expect a fixed number of results to come back from a webservice. This is especially the case in the database module. #1902 bodged this by expecting any change `>1`, a more proper workaround should really be developed.

### LINKED ISSUE #4255: SunpyDeprecationWarnings should cause the doc build to fail
This is especially true in the gallery. I tried to set a filter in the `conf.py` file, but that didn't seem to propagate into the gallery.

nabobalis:
On release versions, tagged basically, we will hide all warnings in the gallery. On master or any dev version they will be there and we can still fix them.

Fixes #4255
Fixes #1917 by restoring the check.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -59,6 +59,11 @@ release = __version__
 sunpy_version = Version(__version__)
 is_release = not(sunpy_version.is_prerelease or sunpy_version.is_devrelease)
 
+# We want to ignore all warnings in a release version.
+if is_release:
+    import warnings
+    warnings.simplefilter("ignore")
+
 # -- SunPy Sample Data and Config ----------------------------------------------
 
 # We set the logger to debug so that we can see any sample data download errors
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,19 +1,15 @@
 """
 Configuration file for the Sphinx documentation builder.
-
-isort:skip_file
 """
-# flake8: NOQA: E402
-
 # -- stdlib imports ------------------------------------------------------------
 import os
 import sys
 import datetime
+import warnings
 from pkg_resources import get_distribution
 from packaging.version import Version
 
 # -- Check for dependencies ----------------------------------------------------
-
 doc_requires = get_distribution("sunpy").requires(extras=("docs",))
 missing_requirements = []
 for requirement in doc_requires:
@@ -29,7 +25,6 @@ if missing_requirements:
     sys.exit(1)
 
 # -- Read the Docs Specific Configuration --------------------------------------
-
 # This needs to be done before sunpy is imported
 on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 if on_rtd:
@@ -40,16 +35,15 @@ if on_rtd:
     os.environ['HIDE_PARFIVE_PROGESS'] = 'True'
 
 # -- Non stdlib imports --------------------------------------------------------
-
 import ruamel.yaml as yaml  # NOQA
 from sphinx_gallery.sorting import ExplicitOrder  # NOQA
 from sphinx_gallery.sorting import ExampleTitleSortKey  # NOQA
 
 import sunpy  # NOQA
 from sunpy import __version__  # NOQA
+from sunpy.util.exceptions import SunpyDeprecationWarning, SunpyPendingDeprecationWarning  # NOQA
 
 # -- Project information -------------------------------------------------------
-
 project = 'SunPy'
 author = 'The SunPy Community'
 copyright = '{}, {}'.format(datetime.datetime.now().year, author)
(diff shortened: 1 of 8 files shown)


docs/conf.py
```diff
@@ -29,27 +25,21 @@
...
 
 # -- Read the Docs Specific Configuration --------------------------------------
-
 # This needs to be done before sunpy is imported
 on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 if on_rtd:
-    os.environ['SUNPY_CONFIGDIR'] = '/home/docs/'
-    os.environ['HOME'] = '/home/docs/'
-    os.environ['LANG'] = 'C'
-    os.environ['LC_ALL'] = 'C'
```
nabobalis, line 39: Restore this.

docs/guide/plotting.rst
```diff
@@ -200,41 +200,6 @@ then used to modify the plot:
     plt.colorbar()
     plt.show()
 
-It is possible to create the same plot, explicitly not using
-`~astropy.visualization.wcsaxes`, however, this will not have the features of
-`~astropy.visualization.wcsaxes` which include correct representation of
-rotation and plotting in different coordinate systems.
```
nabobalis, line 206: If we plan to not support non-wcsaxes, probably don't need this example anymore.

docs/conf.py
```diff
@@ -59,8 +49,17 @@
 sunpy_version = Version(__version__)
 is_release = not(sunpy_version.is_prerelease or sunpy_version.is_devrelease)
 
+# We want to ignore all warnings in a release version.
+if is_release:
+    warnings.simplefilter("ignore")
+warnings.filterwarnings("error", category=SunpyDeprecationWarning)
+warnings.filterwarnings("error", category=SunpyPendingDeprecationWarning)
+warnings.filterwarnings("error", category=SunpyUserWarning)
```
nabobalis, line 57: Remove this.

docs/whatsnew/index.rst
```diff
@@ -6,6 +6,7 @@ Release History
    :maxdepth: 1
 
    changelog
+   3.1
```
nabobalis, line 9: This will need removing on the backport.

docs/conf.py
```diff
@@ -59,8 +53,16 @@
 sunpy_version = Version(__version__)
 is_release = not(sunpy_version.is_prerelease or sunpy_version.is_devrelease)
 
+# We want to ignore all warnings in a release version.
+if is_release:
+    warnings.simplefilter("ignore")
+warnings.filterwarnings("error", category=SunpyDeprecationWarning)
```
nabobalis, line 59: I wonder if I want this to be in an else block.

Cadair: errr, I would say no, just because we really don't want to let a release out which is throwing our own deprecation warnings.

docs/guide/plotting.rst
Cadair: Well, people might want to do this still, perhaps we could point them at something in the example gallery here instead?

Cadair: (This example should have been purged before 3.0 tho)

Cadair:
I am :+1: on this PR, although I would like to know what @dstansby thinks as he initially wasn't in favour and also other opinions as well.

docs/guide/plotting.rst
nabobalis: Do we have one?

dstansby: The example I updated in https://github.com/sunpy/sunpy/pull/5284 doesn't use `WCSAxes`.

dstansby:
I am neutral about doing this, so I won't approve or block. I would say, having been through all the examples fixing some of the warnings (https://github.com/sunpy/sunpy/pull/5286), that there are actually very few warnings currently, so I think this change should be made with a theoretical (ie. galleries should never show warnings) instead of a practical (ie. our gallery is not usable without this change) hat on.

docs/guide/plotting.rst
nabobalis: Added a ref to it.

Cadair:
> so I think this change should be made with a theoretical (ie. galleries should never show warnings) instead of a practical (ie. our gallery is not usable without this change) hat on.

I am of the opinion that showing warnings in the gallery (unless explicitly called out in the example) hinders whatever lesson the example is trying to teach by adding more noise which the reader has to parse, when the warning almost certainly isn't about what the example is trying to convey.

That being said, the examples generally shouldn't be doing things which raise warnings, so ideally we shouldn't need this, but I think practically we do.

Cadair (approved):

PR #5281, opened 2021-04-30
Context manager to handle differential rotation automatically in transformations

ayshih:
Working with `RotatedSunFrame` is sometimes confusing – even for me!  This PR provides a context manager such that "all" transformations with a change in `obstime` will automatically apply solar differential rotation for the difference in time.

```python
>>> import astropy.units as u

>>> import sunpy.map
>>> from sunpy.coordinates import Helioprojective, propagate_with_solar_surface
>>> from sunpy.data.sample import AIA_171_IMAGE

>>> aiamap = sunpy.map.Map(AIA_171_IMAGE)

# HPC frame and disk center for 4 days in the past for an Earth observer

>>> past_frame = Helioprojective(observer='earth', obstime=aiamap.date - 4*u.day)
>>> disk_center = SkyCoord(0*u.arcsec, 0*u.arcsec, frame=past_frame)

# Without the context manager, the past disk center is slightly to the east at the present time
# because its position is changed due only to the observer having moved in heliographic longitude

>>> print(disk_center.transform_to(aiamap.coordinate_frame))
<SkyCoord (Helioprojective: obstime=2011-06-07T06:33:02.770, rsun=696000.0 km, observer=<HeliographicStonyhurst Coordinate (obstime=2011-06-07T06:33:02.770, rsun=696000.0 km): (lon, lat, radius) in (deg, deg, m)
    (-0.00406308, 0.04787238, 1.51846026e+11)>): (Tx, Ty, distance) in (arcsec, arcsec, AU)
    (-59.8859642, -7.85599737, 1.01036752)>

# With the context manager, the disk center has moved to the west after 4 days of solar rotation

>>> with propagate_with_solar_surface():
...     print(disk_center.transform_to(aiamap.coordinate_frame))
<SkyCoord (Helioprojective: obstime=2011-06-07T06:33:02.770, rsun=696000.0 km, observer=<HeliographicStonyhurst Coordinate (obstime=2011-06-07T06:33:02.770, rsun=696000.0 km): (lon, lat, radius) in (deg, deg, m)
    (-0.00406308, 0.04787238, 1.51846026e+11)>): (Tx, Ty, distance) in (arcsec, arcsec, AU)
    (761.80116918, -7.86144802, 1.01226939)>
```

This context manager can also be used to achieve all of the plotting stuff in #5081 with an equally clean API, but now without having to hack WCSAxes.

## To do
- [x] Add test(s)
- [x] Add example(s)

### DIFF
--- a/docs/whatsnew/3.1.rst
+++ b/docs/whatsnew/3.1.rst
@@ -13,6 +13,7 @@ The SunPy project is pleased to announce the 3.1 release of the sunpy package.
 On this page, you can read about some of the big changes in this release:
 
 * :ref:`whatsnew-3.1-python`
+* :ref:`whatsnew-3.1-diffrot-context`
 
 SunPy 3.1 also includes a large number of smaller improvements and bug fixes, which are described in the :ref:`changelog`.
 
@@ -80,6 +81,18 @@ are not written to provide valid results for times before 1972-01-01.
 
 This format is equivalent to `~astropy.time.TimeUnixTai`, except that the epoch
 is 12 years earlier.
+
+.. _whatsnew-3.1-diffrot-context:
+
+Propagating solar-surface coordinates in time
+=============================================
+
+There is now an easy-to-use context manager (:func:`~sunpy.coordinates.propagate_with_solar_surface`) to enable coordinate transformations to take solar rotation into account.
+Normally, a coordinate refers to a point in inertial space, so transforming it to a different observation time does not move the point at all.
+Under this context manager, a coordinate will be treated as if it were referring to a point on the solar surface.
+Coordinate transformations with a change in observation time will automatically rotate the point in heliographic longitude for the time difference, with the amount of rotation depending on the specified differential-rotation model.
+
+.. minigallery:: sunpy.coordinates.propagate_with_solar_surface
 
 Contributors to this Release
 ============================
--- a/docs/whatsnew/3.1.rst
+++ b/docs/whatsnew/3.1.rst
@@ -13,6 +13,7 @@ The SunPy project is pleased to announce the 3.1 release of the sunpy package.
 On this page, you can read about some of the big changes in this release:
 
 * :ref:`whatsnew-3.1-python`
+* :ref:`whatsnew-3.1-map-datetime`
 * :ref:`whatsnew-3.1-diffrot-context`
 
 SunPy 3.1 also includes a large number of smaller improvements and bug fixes, which are described in the :ref:`changelog`.
@@ -36,7 +37,7 @@ We have bumped the minimum version of several packages we depend on; these are t
 - astropy >= 4.2
 - numpy >= 1.17.0
 
-.. _whatsnew-3.1-contributors:
+.. _whatsnew-3.1-map-datetime:
 
 Changes to map date/time handling
 =================================
@@ -93,6 +94,8 @@ Under this context manager, a coordinate will be treated as if it were referring
 Coordinate transformations with a change in observation time will automatically rotate the point in heliographic longitude for the time difference, with the amount of rotation depending on the specified differential-rotation model.
 
 .. minigallery:: sunpy.coordinates.propagate_with_solar_surface
+
+.. _whatsnew-3.1-contributors:
 
 Contributors to this Release
 ============================
--- a/sunpy/coordinates/transformations.py
+++ b/sunpy/coordinates/transformations.py
@@ -135,7 +135,7 @@ def transform_with_sun_center():
         old_ignore_sun_motion = _ignore_sun_motion  # nominally False
 
         if not old_ignore_sun_motion:
-            log.debug("Ignore the motion of the center of the Sun for transformations")
+            log.debug("Ignoring the motion of the center of the Sun for transformations")
         _ignore_sun_motion = True
         yield
     finally:
@@ -215,13 +215,13 @@ def propagate_with_solar_surface(rotation_model='howard'):
 
             old_autoapply_diffrot = _autoapply_diffrot  # nominally False
 
-            log.debug("Enable automatic solar differential rotation "
+            log.debug("Enabling automatic solar differential rotation "
                       f"('{rotation_model}') for any changes in obstime")
             _autoapply_diffrot = rotation_model
             yield
         finally:
             if not old_autoapply_diffrot:
-                log.debug("Disable automatic solar differential rotation "
+                log.debug("Disabling automatic solar differential rotation "
                           "for any changes in obstime")
             _autoapply_diffrot = old_autoapply_diffrot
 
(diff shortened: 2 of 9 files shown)


sunpy/coordinates/transformations.py
```diff
@@ -136,6 +141,21 @@ def transform_with_sun_center():
...
+        try:
+            global _apply_diffrot
+
+            old_apply_diffrot = _apply_diffrot  # nominally False
+
+            log.debug("Apply solar differential rotation for any changes in obstime")
+            _apply_diffrot = True
+            yield
+        finally:
+            _apply_diffrot = old_apply_diffrot
```
Cadair, line 156: Add another debug here so you know when it turns off? Might make the debug output a little easier to parse?

sunpy/coordinates/transformations.py
```diff
@@ -260,7 +280,11 @@ def _transform_obstime(frame, obstime):
     # Transform to the new obstime using the appropriate loopback transformation
     new_frame = frame.replicate(obstime=obstime)
     if frame.obstime is not None:
-        return frame.transform_to(new_frame)
+        if _apply_diffrot:
+            from .metaframes import RotatedSunFrame  # avoid a circular import
+            return RotatedSunFrame(base=frame, rotated_time=obstime).transform_to(new_frame)
+        else:
+            return frame.transform_to(new_frame)
```
Cadair, line 287: You don't really need the else, here. :shrug: 
```suggestion
        return frame.transform_to(new_frame)
```

Cadair (commented):
Well this is certainly nicer than hacking at the internal transform machinery of WCSAxes.

ayshih:
> Well this is certainly nicer than hacking at the internal transform machinery of WCSAxes.

Heh, unfortunately, this is still very much a "draft" PR because it doesn't completely work.  HPC->HPC works great.  However, I haven't figured out the best way for HGS->HGS to work.  `RotatedSunFrame` transformations internally call the HGS->HGS loopback, so trying to add auto-propagation to the HGS->HGS loopback results in an infinite recursion...

sunpy/coordinates/transformations.py
ayshih: Good suggestion, and the same for `transform_with_sun_center()`.

ayshih:
I fixed the HGS->HGS loopback, and curiously, now the implementation doesn't actually use `RotatedSunFrame` directly; instead they share underlying code.

ayshih:
Here's a plotting example.  If one has maps `m1` and `m2` separated by four days:
```python
box1 = SkyCoord([-500, -50]*u.arcsec, [-500, -100]*u.arcsec, frame=m1.coordinate_frame)
box2 = SkyCoord([-450, -100]*u.arcsec, [-350, -50]*u.arcsec, frame=m2.coordinate_frame)

fig = plt.figure(figsize=(12, 5))

fig.add_subplot(121, projection=m1)
m1.plot(vmax=3000)
m1.draw_quadrangle(box1, color='blue')
m1.draw_quadrangle(box2, color='green')

fig.add_subplot(122, projection=m2)
m2.plot(vmax=3000)
m2.draw_quadrangle(box1, color='blue')
m2.draw_quadrangle(box2, color='green')

with propagate_with_solar_surface():
    plt.show()
```
download (28)

ayshih:
For some more fun, I submap the blue box from the first map, and then place it on top of the second map using the newly available `autoalign` functionality:
```python
m1s = m1.submap(box1)
m1s.plot_settings['cmap'] = 'gray'

fig = plt.figure()

fig.add_subplot(projection=m2)
m2.plot(vmax=3000)
m1s.plot(autoalign=True)
m2.draw_quadrangle(box1, edgecolor='blue')

with propagate_with_solar_surface():
    plt.show()
```
download (29)

ayshih:
This PR also means that we no longer need to rely on a "hack" to be able to combine reprojection with differential rotation.  We currently have to add the `RotatedSun*` frame to the `WCS` instance and rely on [SunPy-specific code](https://github.com/sunpy/sunpy/blob/3b542f33a499c96ea8c648bddb54be3d2534a921/sunpy/coordinates/wcs_utils.py#L100-L101) to short-circuit the creation of the target coordinate frame when performing the reprojection.  With this PR, there's no need to use `RotatedSunFrame` at all, and the reprojection to an unmodified `WCS` instance under the context manager performs exactly as desired.

sunpy/coordinates/transformations.py
```diff
@@ -109,31 +128,102 @@ def transform_with_sun_center():
...
+         (82.09298036,  30., 1.),
+         (70.24182965,  60., 1.)]>
+    >>> with propagate_with_solar_surface(rotation_model='rigid'):
+    ...     print(meridian.transform_to(out_frame))
+    <SkyCoord (HeliocentricInertial: obstime=2021-09-21T00:00:00.000): (lon, lat, distance) in (deg, deg, AU)
+        [(85.1064, -60., 1.), (85.1064, -30., 1.),
+         (85.1064,   0., 1.), (85.1064,  30., 1.),
+         (85.1064,  60., 1.)]>
+    """
+    with transform_with_sun_center():
```
Cadair, line 212: :grinning: it's context managers all the way down.

Cadair (approved):
Looks lovely to me

Cadair:
Could do with a what's new entry.

ayshih:
> Could do with a what's new entry.

Done

docs/whatsnew/3.1.rst
```diff
@@ -81,5 +83,17 @@ are not written to provide valid results for times before 1972-01-01.
...
 
+.. _whatsnew-3.1-diffrot-context:
+
+Propagating solar-surface coordinates in time
+=============================================
+
+There is now an easy-to-use context manager (:func:`~sunpy.coordinates.propagate_with_solar_surface`) to enable coordinate transformations to take solar rotation into account.
+Normally, a coordinate refers to a point in inertial space, so transforming it to a different observation time does not move the point at all.
+Under this context manager, a coordinate will be treated as if it were referring to a point on the solar surface.
+Coordinate transformations with a change in observation time will automatically rotate the point in heliographic longitude for the time difference, with the amount of rotation depending on the specified differential-rotation model.
```
Cadair, line 94: Wanna add the mini gallery here?

ayshih: Done

nabobalis (approved):

nabobalis:
Do we have a gallery example to showcase the two examples you posted above?

ayshih:
Let's get this PR in first (which has a basic example, just not the more spiffy examples above)

nabobalis:
> Let's get this PR in first (which has a basic example, just not the more spiffy examples above)

Ok.

PR #5290, opened 2021-05-03
rename master to main

nabobalis:
This in prep of renaming our master branch to main.

I went and found where I could where we reference the sunpy master branch and changed it.

Doc fail fixed in another PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ skip-check: &skip-check
   command: bash .circleci/early_exit.sh
 
 merge-check: &merge-check
-  name: Check if we need to merge upstream master
+  name: Check if we need to merge upstream main
   command: |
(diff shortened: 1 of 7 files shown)


Cadair:
Things we need to do before merging this and renaming the branch (that I can think of) are:

* Email sunpy-dev and @room the main matrix room.
* Check all affiliated packages, website etc repos for anywhere else relying on "master".

Things to do afterwards:

* Update RTD so "latest" still works, this apparently does not happen automatically.

nabobalis:
> Things we need to do before merging this and renaming the branch (that I can think of) are:
> 
> * Email sunpy-dev and @room the main matrix room.

Email sent. Only thing left to do it email the commands to make the change locally.

> * Check all affiliated packages, website etc repos for anywhere else relying on "master".

We use it alot in the tox files for the pins for tox-pypi-filter.

Cadair (approved):

PR #5294, opened 2021-05-03
use release sphinx-gallery for rtd

nabobalis:
Probs time to move back to this.

### DIFF
--- a/rtd-environment.yml
+++ b/rtd-environment.yml
@@ -6,4 +6,4 @@ dependencies:
   - pip
   - python=3.9
   - pip:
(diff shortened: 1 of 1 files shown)


dstansby (approved):

dstansby:
The doc build fail is real, but don't know if it's caused by this PR:

```
checking consistency... /home/vsts/work/1/s/docs/whatsnew/3.1.rst: WARNING: document isn't included in any toctree
```

nabobalis:
That is a real error and is fixed in another open PR.

Issue #5289, opened 2021-05-02
Some Map methods incorrectly re-define the reference pixel/coordinate


ayshih:
The reference pixel/coordinate of a Map is intimately tied to the WCS projection.  Switching from one pixel/coordinate to another pixel/coordinate will change the pixel->world mapping for the rest of the pixels in nearly all situations.  Both `GenericMap.resample()` and `GenericMap.superpixel()` re-define the reference pixel/coordinate to be the center of the Map.  This should not be done.  Instead, the new pixel location for the original reference coordinate should be calculated.  That is, `CRVAL1/2` should be preserved, and only `CRPIX1/2` should be changed.

There may be other places in our codebase that also make the same mistake.

(Incidentally, WCS FITS does have the capacity to decouple the reference coordinate from being the origin of the projection through `PVi_1`/`PVi_2` keywords, but I don't know if anyone actually uses those.  See [Section 2.5 of Calabretta and Greisen 2002](https://www.aanda.org/articles/aa/pdf/2002/45/aah3860.pdf#page=4).)

jeffreypaul15:
Correct me if I'm wrong, but should `crval1` and `crval2` in these two methods be the same as the original map?

Cadair:
Yes. CRPIX is the pixel location of the reference pixel, and CRVAL is the coordinate of the reference pixel in the coordinate system described by the WCS. So What needs to happen is that CRVAL shouldn't change but CRPIX should change (even if it references a pixel out of the actual array) so that the coordinates are correct.

jeffreypaul15:
So changing this line :
https://github.com/sunpy/sunpy/blob/e99c7db009f990bfc93dd11b451db9669d862cda/sunpy/map/mapbase.py#L1419
to 
```python
new_meta['crval1'] = self.meta['crval1`]
new_meta['crval2'] = self.meta['crval2`]
``` 
Should work right? Atleast for `GenericMap.resample()`?

dstansby:
I suspect `GenericMap.rotate()` might need the same treatment too.

ayshih:
> I suspect `GenericMap.rotate()` might need the same treatment too.

To my checking, both `submap()` and `rotate()` are fine.

ayshih:
@jeffreypaul15 You simply need to remove those lines because the CRVAL values are already copied.  That fixes CRVAL, but doesn't fix CRPIX.  The non-trivial effort is to correctly calculate the new CRPIX values.

PR #5295, opened 2021-05-03
Fix metadata computations in superpixel()

dstansby:
Fixes half of https://github.com/sunpy/sunpy/issues/5289. Whilst I was at it I did a bit of refactoring to make the code nicer. Also none of the current tests actually failed on the change, so I added a few new tests. More suggestions for tests are very welcome.


sunpy/map/mapbase.py
```diff
@@ -1906,20 +1910,15 @@ def superpixel(self, dimensions: u.pixel, offset: u.pixel = (0, 0)*u.pixel, func
...
-        new_meta['crval1'] = lon.to(self.spatial_units[0]).value + 0.5 * \
-            (offset[0]*self.scale[0]).to(self.spatial_units[0]).value
-        new_meta['crval2'] = lat.to(self.spatial_units[1]).value + 0.5 * \
-            (offset[1]*self.scale[1]).to(self.spatial_units[1]).value
+            new_meta['CD1_1'] *= scale_factor[0]
+            new_meta['CD2_1'] *= scale_factor[0]
+            new_meta['CD1_2'] *= scale_factor[1]
+            new_meta['CD2_2'] *= scale_factor[1]
+        new_meta['crpix1'] = ((self.meta['crpix1'] - 0.5) / scale_factor[0]) + 0.5 + offset[0]
+        new_meta['crpix2'] = ((self.meta['crpix2'] - 0.5) / scale_factor[1]) + 0.5 + offset[1]
```
ayshih, line 1921: This doesn't look to be the correct handling of `offset`...  `offset` is in original-map pixels, but you shift `CRPIX` as if it is in  resampled-map pixels

dstansby: Thanks; I think it should be fixed now.

dstansby:
The new figure test images don't look right; this needs revising to fix them.

sunpy/map/tests/test_plotting.py
```diff
@@ -167,7 +167,7 @@ def test_plot_masked_aia171_nowcsaxes(aia171_test_map_with_mask):
 
 @figure_test
 def test_plot_aia171_superpixel(aia171_test_map):
-    aia171_test_map.superpixel((9, 7) * u.pix, offset=(4, 4) * u.pix).plot()
+    aia171_test_map.superpixel((3, 2) * u.pix, offset=(4, 4) * u.pix).plot()
```
dstansby, line 170: I changed this to check that the changes looked fine on a slightly higher resolution map.

dstansby:
Here's some before and after checks to verify this is now working as intended. I have constructed a WCS in a Carrington frame, with the reference pixel not on the equator. This is similar to HMI SHARPS projections, which I think the original issue could have been motivated by. When using `superpixel`, the data should change but the coordinate overlay should not. Before this change the coordinate overlay did change, but after the coordinates do not change, as expected.


**Before**

before

**After**

after

```python

import astropy.units as u
import numpy as np

import sunpy.map

# Create map
data = np.random.rand(20**2).reshape((20, 20))
obstime = '2020-01-01'
observer = SkyCoord(0*u.deg, 0*u.deg, frame='heliographic_stonyhurst',
                    obstime=obstime)
ref_coord = SkyCoord(120*u.deg, -70*u.deg, frame='heliographic_carrington',
                     obstime=obstime, observer=observer)
ref_pix = [0, 0] * u.pix
scale = [2, 1] * u.deg / u.pix
smap = sunpy.map.Map(data, header)

superpix = smap.superpixel([2, 2] * u.pix)

fig = plt.figure()
ax = plt.subplot(121, projection=smap)
smap.plot()
ax.coords.grid(True, color='tab:red', ls='solid', lw=2, alpha=1)

ax = plt.subplot(122, projection=superpix)
superpix.plot()
ax.coords.grid(True, color='tab:red', ls='solid', lw=2, alpha=1)

plt.show()
```

dstansby:
This is ready to review. The circleci failures are because I've added a new test, and the figure tests error when trying to download a baseline image that doesn't exist.

nabobalis:
I still think the hash might be wrong.

dstansby:
I don't have time to finish this off by the end of the week; still to do:

- Update the new figure test image hash to match the correct value generated by circleci

dstansby:
Right, I've *finally* (grumble grumble this figure test stuff used to be fine) sorted out the hashes.

Cadair:
Took me 10 mins with a bit of paper, but I made sure the maths was right :rofl:

PR #5301, opened 2021-05-04
Minor `superpixel()` fixes

dstansby:
1. `reshape_image_to_4d_superpixel()` internally uses `int()` to round `dimensions` and `offset`, but this rounding wasn't applied to the actual values given to `superpixel()`, which could result in fractional values in metadata. Now the rounding is done explicitly in `superpixel()`, and mentioned in the docstring.
2. Change some `.value` calls to `.to_value()`, to make different units (e.g. milli-pixels) work properly.

This is pulled out of https://github.com/sunpy/sunpy/pull/5295 for easier review.


Cadair (approved):
Seems uncontroversial

nabobalis (approved):

nabobalis:
Looks like some online servers are having issues.

PR #4079, opened 2020-05-01
Linting 3: autopep8

nabobalis:
This shows up the stages of autopep8.

From the base command, to the first a flag, to the second a flag.

PR #5305, opened 2021-05-04
Remove runtime use of assert_quantity_allclose

Cadair:
Introduces a runtime dep on pytest, so must be avoided

### DIFF
--- a/changelog/5305.bugfix.rst
+++ b/changelog/5305.bugfix.rst
@@ -1,0 +1,2 @@
+Remove runtime use of `astropy.tests.helper.assert_quantity_allclose` which
+introduces a runtime dependancy on `pytest`.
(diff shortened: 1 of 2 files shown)


nabobalis (approved):

nabobalis:
On a side note based on pep8, is that an error we want to check for/fix?

nabobalis (approved):

Cadair:
> On a side note based on pep8, is that an error we want to check for/fix?

Given Albert has chosen to use that style in coordinates I wanna say no.

ayshih:
> > On a side note based on pep8, is that an error we want to check for/fix?
> 
> Given Albert has chosen to use that style in coordinates I wanna say no.

xref: https://github.com/sunpy/sunpy/pull/4079#issuecomment-623247256

ayshih:
xref for how this got in and past reviewers (including me): #4879

nabobalis:
This was not backported to 2.1 or 2.0 it seems so I don't think we need to backport this.

ayshih:
Agreed

PR #5308, opened 2021-05-05
update rtd again

nabobalis:
Just tidy up the config.


rtd-environment.yml
```diff
@@ -5,5 +5,3 @@ dependencies:
   - openjpeg
   - pip
   - python=3.9
-  - pip:
-      - sphinx-gallery
```
nabobalis, line 9: Since rtd installs this via pip install sunpy, we don't need to actually do this here anymore at all.

PR #5313, opened 2021-05-06
removed this utter garbage and disgrace

nabobalis:


Cadair:
I kinda like this, making it into a super small sphinx directive would solve the import time issue, but is that worth it?

nabobalis:
> I kinda like this, making it into a super small sphinx directive would solve the import time issue, but is that worth it?

No.

PR #5314, opened 2021-05-06
Re-enable the JPL HORIZONS consistency unit test

ayshih:
Now that astropy/astropy#11601 has made the JPL DE440s ephemeris available, we can re-enable the JPL HORIZONS consistency unit test that was disabled in #5203.  The test will be skipped if the installed version of Astropy is not aware of DE440s.

### DIFF
--- a/changelog/5314.trivial.rst
+++ b/changelog/5314.trivial.rst
@@ -1,0 +1,1 @@
+Re-enabled the unit test to check for coordinates consistency with JPL HORIZONS when the matching ephemeris can be specified.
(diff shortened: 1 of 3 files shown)


Cadair:
I opened a PR to your fork as I have a desire to see all test classes in existence burnt and replaced with pytest fixtures.

ayshih:
PR merged, with a small fix

Cadair:
I thought putting the mark on the fixture means that all tests which use the fixture have the mark?

Cadair (approved):

ayshih:
> well now I know.

Heh, I didn't know either until I tested it today.  It isn't caught by our CI because we don't have `astroquery` installed except on `remote-data=any` builds, so the `pytest.importorskip()` call skips the test before any online access.

PR #5315, opened 2021-05-07
Support observers defined in obsgeo in wcs_utils

Cadair:
### Description

Adds support for reading observer locations out of obsgeo for ground based observers.

### DIFF
--- a/sunpy/coordinates/tests/test_wcs_utils.py
+++ b/sunpy/coordinates/tests/test_wcs_utils.py
@@ -3,7 +3,7 @@ import numpy as np
 import pytest
 
 import astropy.units as u
-from astropy.coordinates import ITRS, BaseCoordinateFrame
+from astropy.coordinates import ITRS, BaseCoordinateFrame, SkyCoord
 from astropy.coordinates.earth import EarthLocation
 from astropy.tests.helper import assert_quantity_allclose
 from astropy.time import Time
@@ -275,7 +275,8 @@ def dkist_location():
     return EarthLocation(*(-5466045.25695494, -2404388.73741278, 2242133.88769004) * u.m)
 
 
-def test_obsgeo_cartesian(dkist_location):
+@pytest.mark.remote_data
+def test_obsgeo_frame_mapping_cartesian(dkist_location, caplog):
 
     obstime = Time("2021-05-21T03:00:00")
     wcs = WCS(naxis=2)
@@ -289,8 +290,11 @@ def test_obsgeo_cartesian(dkist_location):
 
     assert frame.observer == SkyCoord(dkist_location.get_itrs(obstime)).transform_to('heliographic_stonyhurst').frame
 
+    assert not caplog.records
 
-def test_obsgeo_spherical(dkist_location):
+
+@pytest.mark.remote_data
+def test_frame_mapping_obsgeo_spherical(dkist_location, caplog):
 
     obstime = Time("2021-05-21T03:00:00")
     location = dkist_location.get_itrs(obstime)
@@ -305,6 +309,8 @@ def test_obsgeo_spherical(dkist_location):
     assert frame.observer is not None
 
     assert frame.observer == SkyCoord(location).transform_to('heliographic_stonyhurst').frame
+
+    assert not caplog.records
 
 
 # TODO: Remove all these tests after Astropy 5.0 when we can just import obsgeo_to_frame.
--- a/sunpy/coordinates/tests/test_wcs_utils.py
+++ b/sunpy/coordinates/tests/test_wcs_utils.py
@@ -317,7 +317,8 @@ def test_frame_mapping_obsgeo_spherical(dkist_location, caplog):
 def test_obsgeo_cartesian(dkist_location):
     obstime = Time("2021-05-21T03:00:00")
     wcs = WCS(naxis=2)
-    wcs.wcs.obsgeo = dkist_location.to_value(u.m).tolist() + [0, 0, 0]
+    # tolist() returns a tuple.
+    wcs.wcs.obsgeo = list(dkist_location.to_value(u.m).tolist()) + [0, 0, 0]
     wcs.wcs.dateobs = obstime.isot
 
     frame = obsgeo_to_frame(wcs.wcs.obsgeo, obstime)
--- a/sunpy/coordinates/tests/test_wcs_utils.py
+++ b/sunpy/coordinates/tests/test_wcs_utils.py
@@ -270,12 +270,44 @@ def test_self_observer():
     assert u.allclose(wcs.wcs.aux.dsun_obs, frame.radius.to_value(u.m))
 
 
-# TODO: Remove all these tests after Astropy 5.0 when we can just import obsgeo_to_frame.
 @pytest.fixture
 def dkist_location():
     return EarthLocation(*(-5466045.25695494, -2404388.73741278, 2242133.88769004) * u.m)
 
 
+def test_obsgeo_cartesian(dkist_location):
+
+    obstime = Time("2021-05-21T03:00:00")
+    wcs = WCS(naxis=2)
+    wcs.wcs.ctype = ['HPLT', 'HPLN']
+    wcs.wcs.obsgeo = list(dkist_location.to_value(u.m).tolist()) + [0, 0, 0]
+    wcs.wcs.dateobs = obstime.isot
+
+    frame = solar_wcs_frame_mapping(wcs)
+
+    assert frame.observer is not None
+
+    assert frame.observer == SkyCoord(dkist_location.get_itrs(obstime)).transform_to('heliographic_stonyhurst').frame
+
+
+def test_obsgeo_spherical(dkist_location):
+
+    obstime = Time("2021-05-21T03:00:00")
+    location = dkist_location.get_itrs(obstime)
+    loc_sph = location.spherical
+    wcs = WCS(naxis=2)
+    wcs.wcs.ctype = ['HPLT', 'HPLN']
+    wcs.wcs.obsgeo = [0, 0, 0] + [loc_sph.lon.value, loc_sph.lat.value, loc_sph.distance.value]
+    wcs.wcs.dateobs = obstime.isot
+
+    frame = solar_wcs_frame_mapping(wcs)
+
+    assert frame.observer is not None
+
+    assert frame.observer == SkyCoord(location).transform_to('heliographic_stonyhurst').frame
+
+
+# TODO: Remove all these tests after Astropy 5.0 when we can just import obsgeo_to_frame.
 def test_obsgeo_cartesian(dkist_location):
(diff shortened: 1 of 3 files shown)


dstansby:
Is the `OBSGEO` formulation solar physics specific? If not I would think the code to extract an obsgeo `SkyCoord` from a `WCS` should go upstream into astropy so it is more widely available.

Cadair:
It's not, but the parsing it as the observer to the sunpy coordinate frames is specific to us. I agree that the weird struct parsing should probably go upstream as a helper function somewhere, but I am not exactly sure where.

There are also some astropy frames that should really parse this as the observer location, but the astropy frames are all a bit of a mess with regards to observers.

dstansby (changes requested):
Thanks for sending this upstream too - since there is review going on upstream in astropy, I'm going to block this until that PR is merged. Then the "astropy accepted" version can be vendored into this PR.

sunpy/coordinates/wcs_utils.py
```diff
@@ -74,6 +142,19 @@ def solar_wcs_frame_mapping(wcs):
                              attrs[2] * u.m,
                              **kwargs)
 
+    # Read the observer out of obsgeo for ground based observers
+    if observer is None:
```
dstansby, line 146: Beyond the scope of this PR, but would be nice to put all the logic for getting an observer out of a WCS into its own function.

sunpy/coordinates/wcs_utils.py
```diff
@@ -74,6 +142,19 @@ def solar_wcs_frame_mapping(wcs):
...
+        try:
+            observer = obsgeo_to_frame(wcs.wcs.obsgeo, dateobs)
+            observer = SkyCoord(observer, rsun=rsun)
+        except ValueError as e:
+            # The helper function assumes you know the obsgeo coords you are
+            # parsing are good, we are not sure, so catch the error.
+
+            # This approach could lead to an invalid observer (i.e. one of the
+            # coords being NaN), but only if the WCS has been constructed like that.
+            log.debug(f"Could not parse obsgeo coordinates from WCS:\n{e}")
```
dstansby, line 156: Will this emit a message even if there is no observer defined on the WCS? Perhaps an additional check should be added to L146 to see if any `wcs.wcs.obsgeo` values are not their default values?

sunpy/coordinates/tests/test_wcs_utils.py
```diff
@@ -265,3 +268,103 @@ def test_self_observer():
...
+    return EarthLocation(*(-5466045.25695494, -2404388.73741278, 2242133.88769004) * u.m)
+
+
+@pytest.mark.remote_data
+def test_obsgeo_frame_mapping_cartesian(dkist_location, caplog):
+
+    obstime = Time("2021-05-21T03:00:00")
+    wcs = WCS(naxis=2)
+    wcs.wcs.ctype = ['HPLT', 'HPLN']
+    wcs.wcs.obsgeo = list(dkist_location.to_value(u.m).tolist()) + [0, 0, 0]
```
dstansby, line 284:
```suggestion
    wcs.wcs.obsgeo[:3] = list(dkist_location.to_value(u.m).tolist())
```
is clearer, and should keep the default zero values in the second half of the array.

sunpy/coordinates/tests/test_wcs_utils.py
```diff
@@ -265,3 +268,103 @@ def test_self_observer():
...
+
+@pytest.mark.remote_data
+def test_frame_mapping_obsgeo_spherical(dkist_location, caplog):
+
+    obstime = Time("2021-05-21T03:00:00")
+    location = dkist_location.get_itrs(obstime)
+    loc_sph = location.spherical
+    wcs = WCS(naxis=2)
+    wcs.wcs.ctype = ['HPLT', 'HPLN']
+    wcs.wcs.obsgeo = [0, 0, 0] + [loc_sph.lon.value, loc_sph.lat.value, loc_sph.distance.value]
```
dstansby, line 304:
```suggestion
    wcs.wcs.obsgeo[3:] = [loc_sph.lon.to_value(u.deg), loc_sph.lat.to_value(u.deg), loc_sph.distance.to_value(u.m)]
```

sunpy/coordinates/tests/test_wcs_utils.py
```diff
@@ -265,3 +268,103 @@ def test_self_observer():
...
+    assert frame.observer == SkyCoord(location).transform_to('heliographic_stonyhurst').frame
+
+    assert not caplog.records
+
+
+# TODO: Remove all these tests after Astropy 5.0 when we can just import obsgeo_to_frame.
+def test_obsgeo_cartesian(dkist_location):
+    obstime = Time("2021-05-21T03:00:00")
+    wcs = WCS(naxis=2)
+    wcs.wcs.obsgeo = list(dkist_location.to_value(u.m).tolist()) + [0, 0, 0]
```
dstansby, line 320:
```suggestion
    wcs.wcs.obsgeo[:3] = dkist_location.to_value(u.m).tolist()
```
Presumably `list(a.tolist())` wasn't achieving anything?

sunpy/coordinates/tests/test_wcs_utils.py
```diff
@@ -265,3 +268,103 @@ def test_self_observer():
...
+    assert frame.z == dkist_location.z
+
+
+def test_obsgeo_spherical(dkist_location):
+    obstime = Time("2021-05-21T03:00:00")
+    dkist_location = dkist_location.get_itrs(obstime)
+    loc_sph = dkist_location.spherical
+
+    wcs = WCS(naxis=2)
+    wcs.wcs.obsgeo = [0, 0, 0] + [loc_sph.lon.value, loc_sph.lat.value, loc_sph.distance.value]
```
dstansby, line 337:
```suggestion
    wcs.wcs.obsgeo[3:] = [loc_sph.lon.to_value(u.deg), loc_sph.lat.to_value(u.deg), loc_sph.distance.to_value(u.m)]
```

sunpy/coordinates/tests/test_wcs_utils.py
```diff
@@ -265,3 +268,103 @@ def test_self_observer():
...
+
+    frame = obsgeo_to_frame(wcs.wcs.obsgeo, obstime)
+
+    assert isinstance(frame, ITRS)
+    assert u.allclose(frame.x, dkist_location.x)
+    assert u.allclose(frame.y, dkist_location.y)
+    assert u.allclose(frame.z, dkist_location.z)
+
+
+def test_obsgeo_infinite(dkist_location):
```
dstansby, line 348: Which part of this test is 'infinite'?

dstansby (approved):
Minor comments, but looks 👍 overall

sunpy/coordinates/tests/test_wcs_utils.py
Cadair: It doesn't work, wcslib expects a 6 element array to be parsed to it. It doesn't default to an empty array it defaults to an unset pointer. (null)

sunpy/coordinates/tests/test_wcs_utils.py
Cadair: well 'nonfinite' I guess. (Technically I think the floating point spec teats NaN and Inf as 'infinite').

sunpy/coordinates/tests/test_wcs_utils.py
dstansby: 😢

sunpy/coordinates/tests/test_wcs_utils.py
dstansby: Would recommend `infinite` > `nonfinite` then to avoid confusion.

sunpy/coordinates/tests/test_wcs_utils.py
Cadair: lol quite.

sunpy/coordinates/wcs_utils.py
Cadair: It will. The issue is that the detection logic is frail and a little involved so I didn't want to duplicate it here. I added this log only as a way of debugging it if you think it should be working and isn't. I guess just adding a simple `is not None` would wipe out most of the cases where you don't want to hit that. That being said there are cases in the sunpy test suite where it's not None it's all NaNs. So there are just way too many cases to catch and I didn't want to replicate the checks in the upstream function in here.

sunpy/coordinates/tests/test_wcs_utils.py
Cadair:
```suggestion
def test_obsgeo_nonfinite(dkist_location):
```

sunpy/coordinates/tests/test_wcs_utils.py
Cadair: > Presumably `list(a.tolist())` wasn't achieving anything?

Turns out it was:

```
>       wcs.wcs.obsgeo = dkist_location.to_value(u.m).tolist() + [0, 0, 0]
E       TypeError: can only concatenate tuple (not "list") to tuple
```

Thanks for that one numpy, why does `to_list()` return a `tuple`??!

sunpy/coordinates/tests/test_wcs_utils.py
```diff
@@ -265,3 +268,103 @@ def test_self_observer():
...
+    assert frame.observer == SkyCoord(location).transform_to('heliographic_stonyhurst').frame
+
+    assert not caplog.records
+
+
+# TODO: Remove all these tests after Astropy 5.0 when we can just import obsgeo_to_frame.
+def test_obsgeo_cartesian(dkist_location):
+    obstime = Time("2021-05-21T03:00:00")
+    wcs = WCS(naxis=2)
+    wcs.wcs.obsgeo = dkist_location.to_value(u.m).tolist() + [0, 0, 0]
```
Cadair, line 320:
```suggestion
    # tolist() returns a tuple.
    wcs.wcs.obsgeo = list(dkist_location.to_value(u.m).tolist()) + [0, 0, 0]
```

sunpy/coordinates/wcs_utils.py
```diff
@@ -74,6 +142,19 @@ def solar_wcs_frame_mapping(wcs):
                              attrs[2] * u.m,
                              **kwargs)
 
+    # Read the observer out of obsgeo for ground based observers
+    if observer is None:
+        try:
+            observer = obsgeo_to_frame(wcs.wcs.obsgeo, dateobs)
+            observer = SkyCoord(observer, rsun=rsun)
+        except ValueError as e:
```
nabobalis, line 150: So if the value error happens, we assume this WCS is not ground based?

Cadair: yeah, this is also the last thing we check, so if this doesn't work we assume there isn't an observer defined.

nabobalis (approved):

Cadair:
:shipit:

Issue #5300, opened 2021-05-03
Error pulling GOES/XRS data on Windows?


jwreep:
### Description

I get an error (TypeError: cannot unpack non-iterable NotImplementedType object) pulling GOES/XRS data from SunPy on Windows using the code below.  This works with no issues on MacOS.  

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-9-400b67f69c2d> in <module>
      6 files = Fido.fetch(result)
      7 
----> 8 goes = timeseries.TimeSeries(files, concatenate=True, )
      9 goes = goes.truncate('2013-11-05 22:00','2013-11-05 23:00')

C:\ProgramData\Anaconda3\lib\site-packages\sunpy\timeseries\timeseries_factory.py in __call__(self, silence_errors, *args, **kwargs)
    404         for filepath in filepaths:
    405             try:
--> 406                 new_ts = self._check_registered_widgets(filepath=filepath, **kwargs)
    407                 new_timeseries.append(new_ts)
    408             except (NoMatchError, MultipleMatchError, ValidationFunctionError):

C:\ProgramData\Anaconda3\lib\site-packages\sunpy\timeseries\timeseries_factory.py in _check_registered_widgets(self, **kwargs)
    527         units = kwargs.pop('units', None)
    528         if filepath:
--> 529             data, meta, units = WidgetType._parse_file(filepath)
    530 
    531         # Now return a TimeSeries from the given file.

TypeError: cannot unpack non-iterable NotImplementedType object
```

### Expected vs Actual behavior

The files download on MacOS and can be analyzed without issue.

### Steps to Reproduce

```python
from sunpy.time import parse_time
from sunpy import timeseries
from sunpy.net import Fido, attrs as a

result = Fido.search(
    a.Time('2013/11/04 06:00', '2013/11/06 00:00'),
    a.Instrument.xrs,
) 
files = Fido.fetch(result)

goes = timeseries.TimeSeries(files, concatenate=True, ) 
goes = goes.truncate('2013-11-05 22:00','2013-11-05 23:00')
```

### System Details

==============================
SunPy Installation Information
==============================

#######
General
#######
Time : Monday, 03. May 2021 09:03PM UT
System : Windows
Processor : AMD64 Family 23 Model 113 Stepping 0, AuthenticAMD
Arch : 64bit
SunPy : 2.1.4
[...]

PR #5322, opened 2021-05-10
pin a min version of h5py

nabobalis:
I don't think this needs a changelog? 

Fixes https://github.com/sunpy/sunpy/issues/5300

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -67,6 +67,9 @@ net =
   zeep>=3.4.0
 timeseries =
   h5netcdf>=0.8.1
+  # While a not direct dependacy
+  # We need to raise this to ensure the goes netcdf files open.
+  h5py>=3.1.0
   matplotlib>=3.1.0
   pandas>=0.24.0
 visualization =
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -67,6 +67,9 @@ net =
   zeep>=3.4.0
 timeseries =
   h5netcdf>=0.8.1
+  # While a not direct dependacy
+  # We need to raise this to ensure the goes netcdf files open.
+  h5py>=3.1.0
```
nabobalis, line 72: So I wanted to pin to a >2 version but in 3.1 they fixed a bug with opening data with big endian and I felt that was worth pinning to as a min.

setup.cfg
```diff
@@ -67,6 +67,9 @@ net =
   zeep>=3.4.0
 timeseries =
   h5netcdf>=0.8.1
+  # While a not direct dependacy
```
nabobalis, line 70:
```suggestion
  # While a not direct dependency
```

Cadair:
How old is the 3.1 release ooi?

nabobalis:
> How old is the 3.1 release ooi?

Released: Nov 6, 2020

nabobalis:
This has been manually added to 2.1

PR #5348, opened 2021-05-13
Remove warnings for custom attributes set on WCS

dstansby:
I think just ignoring them is fine now, but open to erroring if we detect these attributes instead now?

### DIFF
--- a/changelog/5348.removal.rst
+++ b/changelog/5348.removal.rst
@@ -1,0 +1,2 @@
+sunpy now errors if the unused ``.rsun`` or ``.heliographic_observer``
+attributes are set on a `~astropy.wcs.WCS`.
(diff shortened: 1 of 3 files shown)


Cadair (changes requested):
I am a little conflicted on this one. I need to have a think about it so this is a request to not merge it until I have had time to do so.

dstansby:
Decision from weekly meeting:
- Error in 3.1/4.0
- Remove code altogether (including error) in 4.1

Cadair (approved):

Cadair:
looks good but the figure tests have exploded.

dstansby:
See https://github.com/sunpy/sunpy/issues/5372 and associated rabbit hole re. figure tests

nabobalis (approved):

PR #5354, opened 2021-05-15
Remove support for plotting maps to non-WCSAxes

dstansby:


nabobalis (changes requested):
Just needs that test file removed.

nabobalis (approved):

PR #5355, opened 2021-05-16
Remove deprecated net API

dstansby:
I think this only leaves a deprecation in `net/helio/hec.py` that needs sorting out, but I wasn't sure what to do there so haven't included it here.


sunpy/net/vso/vso.py
```diff
@@ -201,14 +201,7 @@ def search(self, *query, response_format=None):
...
         if response_format is None:
-            response_format = "legacy"
-            warnings.warn("The default response format from the VSO client will "
-                          "be changing to 'table' in version 3.1. "
-                          "To remove this warning set response_format='legacy' "
-                          "to maintain the old behaviour or response_format='table'"
-                          " to use the new behaviour.",
-                          SunpyDeprecationWarning,
-                          stacklevel=2)
+            response_format = "table"
```
nabobalis, line 204: Stupid question, should we not remove `response_format` entirely?

nabobalis:
I will look at the one in hec.

nabobalis:
I think the hec can be removed with a removal changelog, I think table will error if its bytes

nabobalis:
> I think the hec can be removed with a removal changelog, I think table will error if its bytes

I removed 
```
-        if table:
-            if isinstance(table, bytes):
-                warn('type `bytes` for table_name is deprecated, use `str` instead.', SunpyDeprecationWarning)
-            table = str.encode(table)
-
```
the code seems to work fine still.

nabobalis:
I have gone down a rabbit hole and will push a commit soon(TM)

sunpy/database/tests/test_tables.py
```diff
@@ -464,111 +459,66 @@ def test_entry_from_query_results_with_none_wave_and_default_unit(
     entries = list(entries_from_query_result(qr_with_none_waves, 'nm'))
     assert len(entries) == 10
     expected = [
-        DatabaseEntry(
```
nabobalis, line 467: This might be a mistake in future.

nabobalis:
@Cadair you will want to review this since it touches the tables hack done for 2.1.

nabobalis:
What I removed needs changelogs and should block this PR until I add them.

nabobalis:
Need to fix those fails I introduced.

sunpy/net/vso/vso.py
```diff
@@ -158,19 +159,11 @@ def make(self, atype, **kwargs):
         obj = self.api.get_type(f"VSO:{atype}")
         return obj(**kwargs)
 
-    def search(self, *query, response_format=None):
```
Cadair, line 161: My plan wasn't to remove this keyword in this release, it was to keep it and swap the default value over, so it defaults to the new with the option of reverting to the old.

nabobalis: Any reason?

Cadair: Just that this is one of the oldest parts of sunpy and I expect people are still relying on it. (Also I expected that database would need more hacking)

nabobalis: Fine, I will remove my last commit.

nabobalis (approved):

changelog/5355.breaking.rst
```diff
@@ -0,0 +1 @@
+The default response format from the VSO client is now a table.
```
Cadair, line 1: Shouldn't this be a removal? didn't we deprecate the previous default?

nabobalis: We did, this should removal.

Issue #5356, opened 2021-05-17
Improve error message when trying to specify `observer='self'` without a radial coordinate


dstansby:
It is possible to construct a `HeliographicCarrington(lon, lat, observer='self')` coordinate without specifying the radial distance. This means that the observer is not well defined however, which currently silently passes at initiation, but can cause runtime errors that are not user-friendly (see below for an example).

So, I think trying to create `HeliographicCarrington(lon, lat, observer='self')` should probably error when initiated? @ayshih do you have any thoughts?

```python
from astropy.coordinates import SkyCoord
import astropy.units as u
import sunpy.map

obstime = '2000-1-1'
shape = [180, 360]

frame_out = SkyCoord(
    0 * u.deg, 0 * u.deg, obstime=obstime,
    frame="heliographic_carrington", observer='self')
# Construct header
header = sunpy.map.make_fitswcs_header(
    shape, frame_out,
    scale=[360 / shape[0],
           180 / shape[1]] * u.deg / u.pix,
    reference_pixel=[(shape[0] / 2) - 0.5,
                     (shape[1] / 2) - 0.5] * u.pix,
    projection_code="CEA")
```

```python
Traceback (most recent call last):
  File "/Users/dstansby/miniconda3/envs/dev/lib/python3.9/site-packages/astropy/units/quantity.py", line 740, in to_value
    scale = self.unit._to(unit)
  File "/Users/dstansby/miniconda3/envs/dev/lib/python3.9/site-packages/astropy/units/core.py", line 950, in _to
    raise UnitConversionError(
astropy.units.core.UnitConversionError: 'Unit(dimensionless)' is not a scaled version of 'Unit("m")'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/dstansby/github/sunpy/test.py", line 12, in <module>
    header = sunpy.map.make_fitswcs_header(
  File "/Users/dstansby/miniconda3/envs/dev/lib/python3.9/site-packages/astropy/units/decorators.py", line 252, in wrapper
    return_ = wrapped_function(*func_args, **func_kwargs)
  File "/Users/dstansby/github/sunpy/sunpy/map/header_helper.py", line 128, in make_fitswcs_header
    meta_wcs = _get_wcs_meta(coordinate, projection_code)
  File "/Users/dstansby/github/sunpy/sunpy/map/header_helper.py", line 197, in _get_wcs_meta
[...]

PR #5358, opened 2021-05-17
Error on 2D coordinate with observer='self'

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/5356.  This is kind of breaking in that it now errors earlier than it would have done if you're using the faulty coordinate for anything, but is a better error, so I'm not sure where to go re. backports.

### DIFF
--- a/changelog/5358.bugfix.rst
+++ b/changelog/5358.bugfix.rst
@@ -1,0 +1,3 @@
+Constructing a coordinate in the ``'heliographic_carrington'`` frame with ``observer='self'``
+but without specifying ``radius`` now raises an error upon creation. This is because a full
+3D coordinate is needed to fully define the observer.
--- a/changelog/5358.bugfix.rst
+++ b/changelog/5358.bugfix.rst
@@ -1,3 +1,2 @@
-Constructing a coordinate in the ``'heliographic_carrington'`` frame with ``observer='self'``
-but without specifying ``radius`` now raises an error upon creation. This is because a full
-3D coordinate is needed to fully define the observer.
+Constructing a 2D coordinate in the `~sunpy.coordinates.frames.HeliographicCarrington` frame with ``observer='self'`` now raises an error upon creation.
+When specifying ``observer='self'``, the ``radius`` coordinate component serves as the Sun-observer distance that is necessary to fully define the Carrington heliographic coordinates.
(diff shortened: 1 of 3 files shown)


sunpy/coordinates/frames.py
```diff
@@ -325,6 +328,12 @@ class HeliographicCarrington(BaseHeliographic):
 
     observer = ObserverCoordinateAttribute(HeliographicStonyhurst)
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.observer == 'self' and self._is_2d:
```
adrn, line 333: As I mentioned in Slack, the error here is because, if `self.observer` is a frame instance, the == calls `is_equivalent_frame` on the string "self", which is raising an error
```suggestion
        if not isinstance(self.observer, BaseCoordinateFrame) and coord.self.observer == 'self' and self._is_2d:
```

sunpy/coordinates/frames.py
```diff
@@ -173,6 +173,11 @@ class BaseHeliographic(SunPyBaseCoordinateFrame):
 
     rsun = QuantityAttribute(default=_RSUN, unit=u.km)
 
+    @property
+    def _is_2d(self):
+        return (self._data is not None and self._data.norm().unit is u.one
+                and u.allclose(self._data.norm(), 1*u.one))
+
```
Cadair, line 180: This could go on the sunpy base class, then HPC could also use it?

sunpy/coordinates/tests/test_frames.py
```diff
@@ -427,3 +427,9 @@ def test_skycoord_hpc(args, kwargs):
         hgs = sc.transform_to("heliographic_stonyhurst")
 
     assert isinstance(hgs.frame, HeliographicStonyhurst)
+
+
+def test_hpc_incomplete_observer():
```
ayshih, line 432:
```suggestion
def test_hgc_incomplete_observer():
```

changelog/5358.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Constructing a coordinate in the ``'heliographic_carrington'`` frame with ``observer='self'``
+but without specifying ``radius`` now raises an error upon creation. This is because a full
+3D coordinate is needed to fully define the observer.
```
ayshih, line 3:
```suggestion
Constructing a 2D coordinate in the `~sunpy.coordinates.frames.HeliographicCarrington` frame with ``observer='self'`` now raises an error upon creation.
When specifying ``observer='self'``, the ``radius`` coordinate component serves as the Sun-observer distance that is necessary to fully define the Carrington heliographic coordinates.
```

ayshih (approved):

Cadair (approved):

Issue #5357, opened 2021-05-17
Axes limits on NOAA SWPC plot in docs are not ideal


wtbarnes:
https://docs.sunpy.org/en/stable/api/sunpy.timeseries.sources.NOAAIndicesTimeSeries.html#sunpy.timeseries.sources.NOAAIndicesTimeSeries.peek

![image](https://user-images.githubusercontent.com/8676141/118529819-eb017380-b711-11eb-88fa-1495043a3daa.png)

The axes limits on the date axes squish the whole plot to a small section on the right. I'm not sure if this is the expected output, but it looks odd.

jeffreypaul15:
That dataframe in the example is set for all of the `types` in the file and `sunsopt number = -1` for all the values to match the dataframe size is what I'm assuming, with `type = sunspot SWO` has rows `3268` with `-1.00` occupying `2976/3268` values.
removing `axes.set_ylim(0)` should change it right? But this is based on the assumption that `-1` is not set to keep size consistent with the other types in the dataframe

![image](https://user-images.githubusercontent.com/50923538/118614542-80b60500-b7dd-11eb-8292-3c8db0564f3f.png)

Like @nabobalis mentioned at #5200, maybe having a combined plot would be better for this example?
![image](https://user-images.githubusercontent.com/50923538/117732792-1bc34380-b20e-11eb-9e3d-94fb33b38965.png)

jeffreypaul15:
That dataframe in the example is set for all of the `types` in the file and `sunsopt number = -1` for all the values to match the dataframe size is what I'm assuming, with `type = sunspot SWO` has rows `3268` with `-1.00` occupying `2976/3268` values.
removing `axes.set_ylim(0)` should change it right? But this is based on the assumption that `-1` is not set to keep size consistent with the other types in the dataframe.

![image](https://user-images.githubusercontent.com/50923538/118614542-80b60500-b7dd-11eb-8292-3c8db0564f3f.png)

Or removing `-1.00` from the dataframe before plotting ?
![image](https://user-images.githubusercontent.com/50923538/118616248-25851200-b7df-11eb-8d30-16a49d3d7f3c.png)

Or what mentioned over at #5200, maybe having a combined plot would be better for this example?
![image](https://user-images.githubusercontent.com/50923538/117732792-1bc34380-b20e-11eb-9e3d-94fb33b38965.png)

dstansby:
[...]

PR #5363, opened 2021-05-21
Convert -1 sunspot values to NaN

dstansby:
The NOAA sources use -1 as a fill value, so convert these values to `NaN` to make for nicer plotting. I think this goes halfway to fixing https://github.com/sunpy/sunpy/issues/5357 - to fix the remainder of that issue involves working out how pandas plotting automatically sets limits.

### DIFF
--- a/changelog/5363.breaking.rst
+++ b/changelog/5363.breaking.rst
@@ -1,0 +1,2 @@
+Fill values are now set to `np.nan` in `sunpy.timeseries.sources.noaa` file
+parsers. They were previously set to a fill value of ``-1``.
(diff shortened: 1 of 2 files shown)


nabobalis:
I thought this might count as bugfix but I am not sure if this should be a "breaking" change?

dstansby:
I put breaking, and thought we probably shouldn't backport this as it's a behaviour change?

nabobalis (approved):

nabobalis:
Yeah that makes more sense to me.

Issue #5362, opened 2021-05-19
onebox-compatible documentation for promoting SunPy's features and updates


WaaallEEE:
Hi All, 

If you are interested in promoting current, new features and your overall development work, [helionauts.org](helionauts.org) can increase your visibility. As we work toward a Github integration in HelioNauts, our forum will make it easy to direct SunPy requests toward your Github page. One way for us to do that is to make things more visually engaging and so my request here would be to have your webpages (or some of them) compatible with "Onebox". [Onebox](https://meta.discourse.org/t/rich-link-previews-with-onebox/98088) checks the site for oEmbed or Open Graph tags to give an url-preview similar to what is used by Slack. I have no idea how that can be done in Sphinx but I have asked Onebox's dev team to give me some guidelines. Here's an [ongoing conversation](https://meta.discourse.org/t/what-makes-a-site-compatible-with-onebox-previews/190949) with them.
You will gain a LOT more visibility and not just you but also other helionauts will be able to post updates and news about SunPy. (e.g. that SunPy 3 is out?... ;-) )

nabobalis:
Thanks for the report, I will have a look at fixing this and ensure it works for our docs and website.

nabobalis:
I am testing a fix and using opengraph.xyz, to test this. @WaaallEEE could you try https://sunpy--5366.org.readthedocs.build/en/5366/ to see if its fixed for you?

WaaallEEE:
> I am testing a fix and using opengraph.xyz, to test this. @WaaallEEE could you try https://sunpy--5366.org.readthedocs.build/en/5366/ to see if its fixed for you?

@nabobalis It does not seem to work, still getting the error `oEmbed / OpenGraph tags could not be found:description, image`.

nabobalis:
As far as I am aware every tag is defined and in the html file:

```html
<!-- HTML Meta Tags -->
<title>SunPy Documentation</title>
<meta name="description" content="SunPy Documentation">

<!-- Facebook Meta Tags -->
<meta property="og:description" content="SunPy Documentation">
<meta property="og:image" content="https://raw.githubusercontent.com/sunpy/sunpy-logo/master/generated/sunpy_logo_word.png" />
<meta property="og:site_name" content="SunPy Documentation" />
<meta property="og:title" content="SunPy Documentation">
<meta property="og:type" content="website">
<meta property="og:url" content="https://docs.sunpy.org/en/stable//index.html" />

<!-- Twitter Meta Tags -->
[...]

PR #5366, opened 2021-05-22
Use sphinxext-opengraph to fix og tags

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/5362 

Will need rolling out to each package and the template. 

I tested this on our own discourse and got:
one

Issue is that the image on the gallery examples are relative and don't work.
`<meta property="og:image" content="../../../_images/sphx_glr_downloading_lascoC2_001.png">`
Unsure if there is a setting to change with the gallery.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -84,6 +84,9 @@ rst_epilog = """
 """
 
 # -- General configuration -----------------------------------------------------
+# sphinxext-opengraph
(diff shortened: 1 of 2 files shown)


nabobalis:
Not fixed, I will need to patch the theme more or give up and use the sphinx extension.

dstansby (approved):
Seems 👍  - don't know if you wanted to change anything else? If not feel free to merge

nabobalis:
I think I will follow it up if I anything changes upstream with the sphinx-gallery and the first image setting.

PR #5367, opened 2021-05-26
Adds a pre-commit hook for prohibiting a string

### LINKED ISSUE #5361: Add a pre-commit hook which fails on "prohibited strings"
This is inspired by #5221 which uses isort to add a pre-commit hook which prevented import of the `warnings` module.

A more focused solution to this specific case would be to prohibit the existence of both the `warnings.warn` and `from warnings import warn` strings in the code. This concept could be generalised to other strings say if we wanted to stop people doing `from astropy import units as u`.

devanshshukla99:
### Description

Fixes #5361 

This PR adds a pre-commit hook for banning use of certain strings in the package;

<hr>

Sample terminal output from banning ``from warnings import warn``:

```console
Check for prohibited strings.............................................Failed
- hook id: prohibit-string
- exit code: 1

tools/hektemplate.py:40: Prohibited string (from warnings import warn)
sunpy/net/helio/hec.py:6: Prohibited string (from warnings import warn)
sunpy/data/data_manager/cache.py:4: Prohibited string (from warnings import warn)

```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -91,3 +91,8 @@ repos:
         exclude: ".*(.fits|.fts|.fit|.txt|tca.*)$"
       - id: mixed-line-ending
         exclude: ".*(.fits|.fts|.fit|.txt|tca.*)$"
+  - repo: https://github.com/devanshshukla99/pre-commit-hook-prohibit-string
+    rev: v1.2
+    hooks:
+      - id: prohibit-string
+        args: [--prohibit-string, "from warnings import warn,warnings.warn,from astropy import units as u"]


nabobalis:
Can you add the strings we want to avoid?

devanshshukla99:
> Can you add the strings we want to avoid?

Sure, currently ``from warnings import warn`` and ``warnings.warn`` are banned;

nabobalis:
Also "from astropy import units as u"

nabobalis:
So we will need a way to ignore or exclude a file from the check as well. Is that possible?

devanshshukla99:
Yes, that particular ``file`` or ``pattern`` can be mentioned under ``exclude`` in the config file, like:
```console
    hooks:
      - id: prohibit-string
        args: [--prohibit-string, "from warnings import warn,warnings.warn,from astropy import units as u"]
        exclude: "sunpy/map/mapbase.py"
```

nabobalis:
That works, this PR is blocked by #5221 being completed first.

devanshshukla99:
> That works, this PR is blocked by #5221 being completed first.

Maybe we can implement the infrastructure here and enforce the ban via 5221?

nabobalis:
I guess we could add this commit to that PR instead.

devanshshukla99:
> I guess we could add this commit to that PR instead.

Sounds good👍, but I'm not sure how to do that exactly?

nabobalis:
David can do that in his PR.

dstansby:
I've copied the commit across to my PR, and it works great, good work @devanshshukla99!

devanshshukla99:
maybe we should close this then?

PR #2923, opened 2019-02-03
Added support for mocking online tests in test_norh.py

yashrsharma44:
Added some tests which mocks the online tests for `test_norh.py`.
1. Added support for mocking in test_norh.py
2. Added a function which cross checks the data used for mocking
3. Hypothesis test left out of mocking due to different possibilities of data

Partial work for #2874

Issue #2874, opened 2018-12-17
Work towards mocking all the online tests in SunPy


Cadair:

Cadair:
I believe @Chicrala is taking `sunpy/net/dataretriever/test_norh.py` and `sunpy/net/dataretriever/test_fermi.py`

Cadair:
I generated this report on the tests which are covered by our online tests but not our offline ones: https://pste.eu/p/pXZS.html

Chicrala:
@Cadair  Is this link broken or it actually should display an error "404 not found."?

Cadair:
boo it must have deleted it after some amount of time :woman_facepalming:

yashrsharma44:
Hey, I was interested in working some of the portions. Can someone explain the progress, and help me out with the parts that I can work upon?

nabobalis:
There is not a fixed process. The idea is to replicate any test that has a `@pytest.mark.remote_data` decorator  and use the mock library (linked in the original post) to replace the online response with a offline version.

Cadair:
Also for the record, it's worth pointing out that there should still be some online tests to detect changes to the remote services etc. The objective of this issue is to ensure that as much of the sunpy code is tested offline as possible.

Chicrala:
@Cadair, I was discussing that with @nabobalis and we were wondering that maybe the best approach would be to ''duplicate'' the tests with an offline version of them.

Cadair:
The best thing to do is to test the mocks, i.e. if you mock the response to an API call, you want a test to check that that response dosen't change.

darthgera123:
are all tests mocked or are some left?

Chicrala:
Do you want to do fermi? I'm still struggling with norh.

darthgera123:
cool ill do fermi

nabobalis:
test_norh was is "done" here #2923

Chicrala:
Yeah that's the guy who's currently helping me to understand how to make those tests. I was struggling too much to figure it out. So norh is closed?

nabobalis:
Not yet, the pull request is not merged in. It needs another review but it will be assuming that PR is ready.

nabobalis:
I will try to keep track of the PRs sent in for this issue and update the main post.

darthgera123:
fermi is open right?

nabobalis:
It should be.

darthgera123:
@yashrsharma44  im almost done with mocking test_fermi

[...]

PR #5368, opened 2021-05-26
Added support for mocking online tests in norh

jeffreypaul15:
Addresses `norh.py` from #2874 and a partial continuation of #2923
The implementation is a bit different from #2923 and now runs most of the tests offline.

### DIFF
--- a/sunpy/net/dataretriever/sources/tests/test_norh.py
+++ b/sunpy/net/dataretriever/sources/tests/test_norh.py
@@ -22,14 +22,15 @@ def LCClient():
 
 
 def create_url(date, wavelength):
-    date = date.isoformat()
+    year = date.strftime("%Y")
+    month = date.strftime("%m")
+    day = date.strftime("%d")
     if wavelength == 34 * u.GHz:
         freq = 'tcz'
     elif wavelength == 17 * u.GHz:
         freq = 'tca'
-    date = date.split('T')[0].split('-')
-    value_ = date[0][-2:] + date[1] + date[2]
-    base = f"ftp://solar-pub.nao.ac.jp/pub/nsro/norh/data/tcx/{date[0]}/{date[1]}/{freq}{value_}"
+    value_ = year[-2:] + month + day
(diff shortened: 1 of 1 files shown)


sunpy/net/dataretriever/sources/tests/test_norh.py
```diff
@@ -18,7 +21,45 @@ def LCClient():
...
 
-@pytest.mark.remote_data
+def create_url(date, wavelength):
+    date = date.isoformat()
+    if wavelength == 34 * u.GHz:
+        freq = 'tcz'
+    elif wavelength == 17 * u.GHz:
+        freq = 'tca'
+    date = date.split('T')[0].split('-')
+    value_ = date[0][-2:] + date[1] + date[2]
```
dstansby, line 31: Would it be possible to do this with `.strftime()` - https://strftime.org/ is a good reference. If so, it would be easier to understand what's going on here (ie. which bits of the date/time are being extracted where)

sunpy/net/dataretriever/sources/tests/test_norh.py
```diff
@@ -105,29 +154,8 @@ def test_client_repr(LCClient):
...
-        'Source': 'NAOJ',
-        'Provider': 'NRO',
-        'Wavelength': wave,
-        'url': 'ftp://solar-pub.nao.ac.jp/pub/nsro/norh/data/tcx/2016/01/tca160101'
-    }
-    results = QueryResponse([obj], client=LCClient)
-    return results
-
-
 def test_show(LCClient):
```
nabobalis, line 157:
```suggestion
def test_show():
```

sunpy/net/dataretriever/sources/tests/test_norh.py
jeffreypaul15: Yes, this makes more sense. I forgot this existed :sweat_smile:

jeffreypaul15:
@nabobalis Is this the right way to go about this? If yes then shall I open other PRs for the other sources?

nabobalis:
I mean this is ok for now, we really need a better way to mock these tests using vcrpy or other tools.

I would suggest not working on the other sources and working on pyvista.

nabobalis (approved):

PR #5381, opened 2021-06-03
Fixed bugs in transformations involving a change in `obstime`

ayshih:
All transformations between SunPy frames that involve a change in `obstime` invariably depend on the HCRS<->HGS transformations to be correct for a change in `obstime`.  This PR fixes two bugs (including tests to confirm):
* Way back in #3246, I mistakenly shifted the origin in the opposite direction.  This affected both HCRS->HGS and HGS->HCRS.
* In #4613, I accidentally dropped applying a rotation to the origin shift for HGS->HCRS.  This inaccuracy should have been apparent because it meant that HGS->HCRS no longer reversed HCRS->HGS.


ayshih:
Note that since these bugs are both related to shifting of the origin, these bugs were not apparent when using the `transform_with_sun_center()` context manager.

nabobalis (approved):

PR #5384, opened 2021-06-04
Use MapFactory path parsing code in TimeseriesFactory

dstansby:
This extracts a lot of generic path parsing logic from the `MapFactory`, and uses it in the `TimeSeriesFactory`. Part of my attempts to overhaul the `TimeseriesFactory`, with the end goal of making it possible to plug in file readers (CDF files is my use case). But for now this is just refactoring that makes the code better.

### DIFF
--- a/sunpy/timeseries/tests/test_timeseries_factory.py
+++ b/sunpy/timeseries/tests/test_timeseries_factory.py
@@ -450,10 +450,10 @@ class TestTimeSeries:
 
     def test_invalid_filepath(self):
         invalid_filepath = os.path.join(filepath, 'invalid_filepath_here')
-        with pytest.raises(NoMatchError):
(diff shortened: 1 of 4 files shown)


sunpy/timeseries/tests/test_timeseries_factory.py
```diff
@@ -450,10 +450,10 @@ def test_invalid_manual_data(self):
 
     def test_invalid_filepath(self):
         invalid_filepath = os.path.join(filepath, 'invalid_filepath_here')
-        with pytest.raises(NoMatchError):
+        with pytest.raises(ValueError, match='Did not find any files'):
```
dstansby, line 453: These should not have been `NoMatchError`, which is reserved for not finding a matching source (not not finding any files).

nabobalis (approved):

PR #5390, opened 2021-06-04
Added `resample` method in `GenericTimeSeries`

### LINKED ISSUE #5378: Add `resample` method to TimeSeries
### Description

One useful thing when dealing with timeseries is to resample the data. Pandas provide functionality for this, so it would be nice to include it as a method on TimeSeries

devanshshukla99:
### Description

Fixes #5378

### DIFF
--- a/sunpy/timeseries/timeseriesbase.py
+++ b/sunpy/timeseries/timeseriesbase.py
@@ -445,6 +445,33 @@ class GenericTimeSeries:
         object._sanitize_units()
         return object
 
+    def resample(self, rule, **kwargs):
+        """
+        Returns a resampled time series.
+
+        Parameters
+        ----------
+        rule : `~pandas.tseries.offsets.DateOffset`, `~pandas.Timedelta` or `str`
+            The offset string or object representing target conversion.
+
+        Returns
+        -------
+        `~sunpy.timeseries.TimeSeries`
+            A resampled `~sunpy.timeseries.TimeSeries`.
+
+        Notes
+        -----
+        This method is a wrapper around `~pandas.DataFrame.resample`; all additional
+        keyword arguments are passed to this method.
+        """
+
+        object = GenericTimeSeries(self._data.resample(rule, **kwargs).pad(),
+                                   meta=TimeSeriesMetaData(copy.copy(self.meta.metadata)),
+                                   units=copy.copy(self.units))
+        object._sanitize_metadata()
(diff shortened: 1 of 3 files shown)


sunpy/timeseries/timeseriesbase.py
```diff
@@ -445,6 +445,33 @@ def concatenate(self, others, same_source=False, **kwargs):
...
+        `~sunpy.timeseries.TimeSeries`
+            A resampled `~sunpy.timeseries.TimeSeries`.
+
+        Notes
+        -----
+        This method is a wrapper around `~pandas.DataFrame.resample`; all additional
+        keyword arguments are passed to this method.
+        """
+
+        object = GenericTimeSeries(self._data.resample(rule, **kwargs).pad(),
```
nabobalis, line 468: What does the pad do in this context?

devanshshukla99: It'll fill the missing values with `NaN`

dstansby (changes requested):
There are many other operations one might want to do after resampling a `DataFrame` other than `pad()`, see the docs [here](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.resample.html). I think we need to have quite a bit of discussion over how we are going to implement this and what the API will be. Thanks for opening the PR, and sorry for not indicating this on the original issue. I'll update the original issue with a comment when it has been discussed, but for now I'm going to block this from being merged.

nabobalis:
I guess the ugly thing would be a string keyword to parse for the method?

devanshshukla99:
> There are many other operations one might want to do after resampling a `DataFrame` other than `pad()`, see the docs [here](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.resample.html). I think we need to have quite a bit of discussion over how we are going to implement this and what the API will be. Thanks for opening the PR, and sorry for not indicating this on the original issue. I'll update the original issue with a comment when it has been discussed, but for now I'm going to block this from being merged.

was debating myself too, but got swayed towards adding basic func first,
anyway do let me know👍

dstansby:
Thanks for working on this! We decided at the development meeting that we will be transitioning to using astropy TimeSeries at some point, so we want to hold off on implementing any new methods on sunpy TimeSeries until we do that. As such I'm going to close this PR; thanks again for opening it.

PR #5391, opened 2021-06-04
Properly specify `rsun` in the output header when reprojecting an AIA map

ayshih:
Some of the slightly odd results when reprojecting an AIA map (to another HPC frame) is because we need to create a target WCS header that has the correct `rsun_ref`.  We can do that by specifying the `rsun` frame attribute on the reference coordinate provided to `make_fitswcs_header()`.


ayshih:
For example, before:
before1
After:
after1

ayshih:
Also, before:
before2
After:
after2

dstansby (approved):

PR #5392, opened 2021-06-04
Fixed bug where ``GenericMap`` breaks with keyword-arguments

### LINKED ISSUE #4170: Cannot pass sortby keyword argument through to MapSequence from Map
`MapSequence` accepts a `sortby` keyword argument that can be used to optionally sort the maps in the sequence in different ways. However, if you try to pass this kwarg in through `Map`, eg.

```python
sunpy.map.Map(file_list, sequence=True, sortby=None)
```

you get the following exception,

```python-traceback
TypeError: __init__() got an unexpected keyword argument 'sortby'
```

Presumably, this is because `Map` does not properly deal with the `sortby` kwarg before it can be passed to `MapSequence`.

devanshshukla99:
### Description

Fixes #4170 

The parent class of `GenericMap`, `NDData` doesn't support keyword-arguments; that's why an exception was raised with `sortby` kwarg.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -3,6 +3,7 @@ Map is a generic Map class from which all other Map classes inherit from.
 """
 import copy
 import html
+import inspect
 import numbers
 import textwrap
 import webbrowser
@@ -215,7 +216,9 @@ class GenericMap(NDData):
             warn_user("This file contains more than 2 dimensions. "
                       "Data will be truncated to the first two dimensions.")
 
-        super().__init__(data, meta=MetaDict(header), **kwargs)
+        params = list(inspect.signature(NDData).parameters)
+        nddata_kwargs = {x: kwargs.pop(x) for x in params & kwargs.keys()}
+        super().__init__(data, meta=MetaDict(header), **nddata_kwargs)
 
         # Correct possibly missing meta keywords
         self._fix_date()
--- a/changelog/5392.bugfix.rst
+++ b/changelog/5392.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug where `~sunpy.map.GenericMap` used to break with keyword arguments.
--- a/sunpy/map/tests/test_map_factory.py
+++ b/sunpy/map/tests/test_map_factory.py
@@ -60,6 +60,11 @@ class TestMap:
         sequence = sunpy.map.Map(a_list_of_many, sequence=True)
         assert isinstance(sequence, sunpy.map.MapSequence)
 
+    def test_mapsequence_sortby(self):
+        # Test making a MapSequence with sortby kwarg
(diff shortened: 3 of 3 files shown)


nabobalis:
Was there no previous test with sortby or derotate for mapsequence?

devanshshukla99:
mapsequence has one test for `derotate`, which passes;
the exception occurs when supplying kwargs to `MapSequence` through `Map`

sunpy/map/mapbase.py
```diff
@@ -201,7 +201,7 @@ def __init_subclass__(cls, **kwargs):
         if hasattr(cls, 'is_datasource_for'):
             cls._registry[cls] = cls.is_datasource_for
 
-    def __init__(self, data, header, plot_settings=None, **kwargs):
+    def __init__(self, data, header, plot_settings=None, mask=None, **kwargs):
```
nabobalis, line 204: What happens to kwargs now?

dstansby (changes requested):
This isn't the right fix - we still want to hand keyword arguments to `NDData`. I think the correct fix is to

1. Pop the the `sortby` argument from `kwargs` (`sortby = kwargs.pop('sortby', None)`)
2. Pass `sortby` into `MapSequence`, if a `MapSequence` has been requested

devanshshukla99:
> This isn't the right fix - we still want to hand keyword arguments to `NDData`. I think the correct fix is to
> 
> 1. Pop the the `sortby` argument from `kwargs` (`sortby = kwargs.pop('sortby', None)`)
> 2. Pass `sortby` into `MapSequence`, if a `MapSequence` has been requested

Makes sense.

But wouldn't it be better to pop out all kwargs for `NDData` into a seperate var `nd_kwargs`?

dstansby:
> > This isn't the right fix - we still want to hand keyword arguments to `NDData`. I think the correct fix is to
> > 
> > 1. Pop the the `sortby` argument from `kwargs` (`sortby = kwargs.pop('sortby', None)`)
> > 2. Pass `sortby` into `MapSequence`, if a `MapSequence` has been requested
> 
> Makes sense.
> 
> But wouldn't it be better to pop out all kwargs for `NDData` into a seperate var `nd_kwargs`? This would make sure any other kwarg, for example (looks distant but anyway) `filetype`, are also supported.

Yes, it would be good to do that. The clever way would be to use  [`inspect.signature`](https://docs.python.org/3/library/inspect.html#inspect.signature) to work out what the keyword arguments are for `NDData`, and pop them from the kwargs given to `Map`.

sunpy/map/mapbase.py
devanshshukla99: slipped right through my mind, should be fixed now👍

dstansby (approved):

nabobalis (approved):

Issue #4880, opened 2021-01-18
Update FITS time handling to FITS 4


dstansby:
I'm spinning this out of https://github.com/sunpy/sunpy/issues/4716, as it's a big enough and self contained task to warrant its own issue. 

https://fits.gsfc.nasa.gov/standard40/fits_standard40aa-le.pdf is the FITS 4 standards document, with updates to how times are handled in section 9. Making sure we conform to this (and maintain backwards compatibility) will be non trivial, and as such should probably be done with someone who
- is very familiar with sunpy map
- has sat down and read the whole of section 9 of https://fits.gsfc.nasa.gov/standard40/fits_standard40aa-le.pdf

Kate028:
Hey @dstansby , i was going through the section 9 of the mentioned pdf. Could you please let me know if this intended at experts or i try working on it?

PR #5394, opened 2021-06-08
Use DATE-BEG/DATE-END/XPOSURE FITS keywords

dstansby:
This is the recommended way to access image observation time and exposure time in FITS 4. As far as I can tell reading the time part of the FITS 4 standard this now brings `sunpy` up to speed with FITS 4.

TODO:
- [x] Decide whether to backport this 
- [x] Add tests
- [x] Add changelog

### DIFF
--- a/changelog/5394.feature.1.rst
+++ b/changelog/5394.feature.1.rst
@@ -1,0 +1,2 @@
+`sunpy.map.GenericMap.date` now reads from the 'DATE-BEG' FITS keyword if
+present, and if not falls back on the 'DATE-OBS' keyword.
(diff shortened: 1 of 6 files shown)


Cadair:
We should *not* backport this in my opinion, also if we do this we should do the other things we need to do to say we "support FITS 4" in the next release of sunpy.

dstansby:
Just adding support for XPOSURE too, and had the thought that maybe we should backport supporting the new keywords *only* when the old keywords aren't present to 3.0, and then switch to defaulting to new keywords in 3.1.

wtbarnes:
See also #4716 and #4880

dstansby:
Notes from call:

- we should have properties that mirror each of 'DATE-BEG', 'DATE-END', 'DATE-AVG'
- `.date` should be the center point of the observation
- `.date` should be, in order:
- 1) date-avg, 2) average of date-beg, date-end, 3) date-obs, 4) date-beg, 5) date-end

dstansby:
I'm going to split this into a couple of PRs.

PR #5395, opened 2021-06-09
Fixed two bugs associated with the `rsun` attribute for `Heliographic*` frames

ayshih:
#5211 added the `rsun` frame attribute to `Heliographic*` frames.  This PR:

- Properly sets that frame attribute when instantiating a frame from a WCS that has `rsun_ref` set.
- Fixed a bug in the HCC->HGS transformation where the returned HGS frame would fall back to the default `rsun` value regardless of the definition in the target frame.  This bug is most apparent when transforming a SkyCoord from HPC to HGS, which normally tries to propagate the `rsun` value, but instead sees the `rsun` value reset to the default.

PR #5399, opened 2021-06-11
Update devdeps figure hashes

dstansby:
I think some of the devdeps figure hashes need updating for recent coordinate fixes, so here's an empty commit to see if the tests pass or the hashes do indeed need updating.


ayshih:
I think there will need to be a bisection for some change in Matplotlib.  PR #5395's run of devdeps from two days ago does not show these differences in the figure tests:
https://69329-2165383-gh.circle-artifacts.com/0/.tmp/py37-figure-devdeps/figure_test_images/fig_comparison.html

dstansby:
I'd be tempted to say the differences are all so small as to be not a problem. Looking at the commit log (https://github.com/matplotlib/matplotlib/commits/master), it might be https://github.com/matplotlib/matplotlib/pull/20363, but MPL didn't change any of their test images in that PR, so it might be an interaction with that and `WCSAxes` (at a very rough guess)>

ayshih:
It looks like they relaxed a tolerance for an image check.

Yes, we could simply update our hashes, _but_ that Matplotlib PR may also be responsible for this contour craziness in #5398:
result
I don't know whether fixing this plot would require an upstream fix/revert to Matplotlib, but if it does, we may end up reverting the hashes in short order.

dstansby:
Zooming into one of the figure tests that changed, the lines are still in the same place. So I think this is probably some sort of change in the aliasing of paths, and it's safe to change the hashes and merge once https://github.com/sunpy/sunpy/pull/5398 is merged.

dstansby:
This should be good to go now - I've confirmed that all the changes are just in the way lines are drawn, and not in their location.

nabobalis (approved):

PR #5411, opened 2021-06-19
Use singledispatch in TimeSeriesFactory

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/2384. This is just refactoring, but makes it easier to see which bits of code handle the three input types of:

- data, header, units tuple
- file (including URL pointing to a file)
- `GenericTimeseries`

### DIFF
--- a/sunpy/timeseries/timeseries_factory.py
+++ b/sunpy/timeseries/timeseries_factory.py
@@ -5,6 +5,7 @@ import os
 import copy
 import pathlib
 from collections import OrderedDict
+from urllib.request import Request
 
 import numpy as np
 import pandas as pd
@@ -28,6 +29,7 @@ from sunpy.util.datatype_factory_base import (
     NoMatchError,
     ValidationFunctionError,
 )
+from sunpy.util.functools import seconddispatch
 from sunpy.util.io import is_url, parse_path, possibly_a_path
 from sunpy.util.metadata import MetaDict
 from sunpy.util.net import download_file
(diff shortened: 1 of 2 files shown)


nabobalis:
I feel like this is worth a trivial changelog entry.

sunpy/timeseries/timeseries_factory.py
```diff
@@ -238,8 +236,7 @@ def _sanitise_args(self, args):
         Sanitise a list of args so that a single argument corresponds to either:
 
         - A (data, header, units) tuple
-        - A path-like string (e.g. a filename, directory, glob etc.)
-        - A URL
```
nabobalis, line 242: Is this still not true?

dstansby: I updated this to mirror what `MapFactory` does, ie. convert URLs to a request and then use that to dispatch on.

sunpy/timeseries/timeseries_factory.py
```diff
@@ -278,6 +278,11 @@ def _sanitise_args(self, args):
                             args.pop(j)
 
                 args[i] = (data, meta, units)
+
+            elif isinstance(arg, str) and is_url(arg):
+                args[i] = Request(path)
```
wtbarnes, line 283: Based on the test failures, I think this should be `arg` and not `path`.

```suggestion
                args[i] = Request(arg)
```

wtbarnes (changes requested):
Needs a changelog and I think my suggested fix should fix the tests/doc builds

sunpy/timeseries/timeseries_factory.py
```diff
@@ -237,10 +236,10 @@ def _sanitise_args(self, args):
         """
         Sanitise a list of args so that a single argument corresponds to either:
 
-        - A (data, header, units) tuple
-        - A path-like string (e.g. a filename, directory, glob etc.)
-        - A URL
-        - A GenericTimeSeries object
+        - (data, header, units) tuple.
+        - path-like `pathlib.Path` (e.g. a filename, directory, glob etc.).
+        - `urllib.Request`.
```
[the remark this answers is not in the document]
dstansby: Thanks, fixed.

sunpy/timeseries/timeseries_factory.py
dstansby: Woop, thanks for catching, should be fixed now.

wtbarnes (approved):

wtbarnes:
Looks good to me (provided the tests pass!)

PR #5416, opened 2021-06-21
Update handling of solar radius in sunpy.map

dstansby:
This cleans up handling of `rsun_obs` and `rsun_meters` so:

- The place where we choose a default value (if needed) is `rsun_meters`.
- `rsun_meters` can now rely on `rsun_obs` metadata to calculate a value, if present
- The warning has been converted to a log message, fixes https://github.com/sunpy/sunpy/issues/5285

PR #5417, opened 2021-06-22
Add function to get limb coordinates

dstansby:
This pulls out the code for calculating limb coordinates from `draw_limb`. The motivation for this is drawing the limb in `sunkit-pyvista`, so we want the coordinates without calling `draw_limb`.

I chose to use `rsun` (physical distance) instead of `rsun_obs`, as the latter can vary for a varying observer-sun distance, whereas (for a similar observation) `rsun` will not depend on observer-sun distance. As such the logic has changed, but I think I got the trigonometry right.

This relies on https://github.com/sunpy/sunpy/pull/5416 to make sure `dsun` has a fallback on `rsun_obs`.

### DIFF
--- a/sunpy/coordinates/utils.py
+++ b/sunpy/coordinates/utils.py
@@ -7,9 +7,10 @@ import numpy as np
 import astropy.units as u
 from astropy.coordinates import BaseCoordinateFrame, SkyCoord
 
-from sunpy.coordinates import Heliocentric, get_body_heliographic_stonyhurst
+from sunpy.coordinates import Heliocentric, HeliographicStonyhurst, get_body_heliographic_stonyhurst
+from sunpy.sun import constants
 
-__all__ = ['GreatArc', 'get_rectangle_coordinates', 'solar_angle_equivalency']
+__all__ = ['GreatArc', 'get_rectangle_coordinates', 'solar_angle_equivalency', 'get_limb_coordinates']
 
 
 class GreatArc:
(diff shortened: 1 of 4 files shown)


sunpy/coordinates/utils.py
```diff
@@ -7,7 +7,8 @@
 import astropy.units as u
 from astropy.coordinates import BaseCoordinateFrame, SkyCoord
 
-from sunpy.coordinates import Heliocentric, get_body_heliographic_stonyhurst
+from sunpy.coordinates import Heliocentric, HeliographicStonyhurst, get_body_heliographic_stonyhurst
+from sunpy.sun import constants
 
 __all__ = ['GreatArc', 'get_rectangle_coordinates', 'solar_angle_equivalency']
```
nabobalis, line 13:
```suggestion
__all__ = ['GreatArc', 'get_rectangle_coordinates', 'solar_angle_equivalency', 'get_limb_coordinates']
```

wtbarnes (commented):
Looks good to me, but looks like there are conflicts that need to be resolved.

wtbarnes (approved):

sunpy/coordinates/utils.py
```diff
@@ -436,3 +437,36 @@ def solar_angle_equivalency(observer):
               lambda x: np.arctan(x/sun_observer_distance))]
 
     return equiv
+
+
+@u.quantity_input
```
ayshih, line 442: This `quantity_input` decorator isn't doing anything.  Presumably you intended to add a unit annotation for the `rsun` argument?

ayshih (changes requested):
Looks good aside from my one comment

sunpy/coordinates/utils.py
dstansby: Thanks - I either missed this or assumed that setting a default keyword value to a quantity was sufficient. Should be fixed now.

ayshih (approved):
Looks good to me!

PR #5419, opened 2021-06-23
Fixed bugged behavior of `all_corner_coords_from_map()` for non-square maps

ayshih:
#4776 introduced `all_corner_coords_from_map()`, but that function returns incorrect values for non-square maps.  The existing unit test uses a non-square map, but doesn't actually check any of the return values.

### DIFF
--- a/sunpy/map/maputils.py
+++ b/sunpy/map/maputils.py
@@ -34,7 +34,8 @@ def all_pixel_indices_from_map(smap):
         A `numpy.array` with the all the pixel indices built from the
         dimensions of the map.
     """
-    return np.meshgrid(*[np.arange(v.value) for v in smap.dimensions]) * u.pix
+    y, x = np.indices(smap.data.shape)
+    return [x, y] * u.pix
 
 
 def all_coordinates_from_map(smap):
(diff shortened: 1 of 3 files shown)


sunpy/map/maputils.py
```diff
@@ -60,11 +61,9 @@ def all_corner_coords_from_map(smap):
     """
     Returns the coordinates of the pixel corners in a map.
     """
-    nx, ny = smap.data.shape
-    xpix = np.arange(0, nx + 1, 1) - 0.5
-    ypix = np.arange(0, ny + 1, 1) - 0.5
```
ayshih, line 65: Incidentally, I'll note that `nx`, `ny`, `xpix`, and `ypix` were all incorrectly named – the `x` and `y` are swapped – but `meshgrid()` is also used weirdly, which cancels out the swap (from #4776).  Bad @dstansby! =P

ayshih: Yikes, no, it doesn't cancel out.  The existing code is bugged!

dstansby: 😬 thanks for fixing this

dstansby (approved):

PR #5429, opened 2021-06-29
Avoid a pytest import in the coordinates asdf tag

Cadair:
I really hope this is the last one...

closes https://github.com/DKISTDC/dkist/issues/106


nabobalis:
Can you check if we missed any other places?

nabobalis (approved):

Cadair:
I can't see any from grepping, but then I thought that last time I fixed this.

PR #5431, opened 2021-06-30
Add info urls to client results

dstansby:
I think we should enforce this as an `abstractproperty`, but this would be a breaking change for now. So, thoughts on making this compulsory somehow eventually?

Still needs:

- [x] Docs
- [ ] Changelog
- [x] Tests (this will just be fixing current doctests)

### DIFF
--- a/changelog/5431.feature.rst
+++ b/changelog/5431.feature.rst
@@ -1,0 +1,6 @@
+Printing the response from a `~sunpy.net.Fido` query now includes the URL where
+the data files are sourced from.
+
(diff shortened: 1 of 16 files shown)


Cadair:
I like the idea of a structured (mandated) property for a reference URL. I think we can go a step or two further in terms of what we are putting in the repr of `UnifiedResponse`. I think we should include the full import path of the class, and hyperlink it to the docs in the `_html_repr_` as well as probably including the first paragraph of the docstring and now info url.

nabobalis (approved):

nabobalis:
The online tests look like the same ones failing on main.

PR #5432, opened 2021-07-01
Fix plot labels when plotting across different WCSes

dstansby:
This fixes the label handling to take the coordinate frame from the axes being plotted on, instead of the map. It also drops units from the label string if using WCSAxes, as units are indicated in the tick labels, and zooming to different tick label levels (e.g. from degrees to arc-minutes) can result in these labels being invalid.


sunpy/visualization/visualization.py
```diff
@@ -42,14 +42,18 @@ def axis_labels_from_ctype(ctype, unit):
     """
     ctype_short = ctype[:4]
 
-    labels = {'HGLN': f'Heliographic Longitude [{unit}]',
-              'CRLN': f'Carrington Longitude [{unit}]',
-              'HPLN': f'Helioprojective Longitude (Solar-X) [{unit}]',
-              'SOLX': f'Heliocentric X [{unit}]',
+    labels = {'HGLN': f'Heliographic Longitude',
```
hayesla, line 45: is there a reason why you have removed the units here (i.e. arcsec or deg etc)?

dstansby: From the changelogs I will push in a bit, along with updated figure test hashes:
```rst
:meth:`sunpy.map.GenericMap.plot` no longer adds a unit string to the axis
labels if the axes being plotted on is a WCSAxes. For a WCSAxes, angular units
are indicated in the tick labels, and automatically change when the zoom level
changes from e.g. degrees to arc-minutes.
```

dstansby: And I've just added to the end of that:
```rst
This could previously lead to situations where the axis label units were incorrect.
```

hayesla (approved):

nabobalis (approved):

PR #5438, opened 2021-07-02
Fix futurewarning in timeseries

dstansby:
Should fix the py37 macos build.


nabobalis:
Should we not error on futurewarnings?

dstansby:
Yes, we do - this is removing one of those warnings.

nabobalis:
We can do it in another PR.

PR #5448, opened 2021-07-07
Fixed the definition of `lonpole` by the FITS WCS header helper

ayshih:
~I believe this fixes the handling of FITS WCS `lonpole`.  In short, the `lonpole` is set in `WCS` headers based on the default reference coordinate, but if we change the latitude of the reference coordinate (`crval2`), `lonpole` may need to be changed at the same time.~

This PR fixes the definition of `lonpole` by the FITS WCS header helper

This PR also overhauls the suite of unit tests for the header helper because they were a bit of a mess

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -546,6 +546,7 @@ class GenericMap(NDData):
         # array with sequence error.
         w2.wcs.cdelt = u.Quantity(self.scale)
         w2.wcs.crval = u.Quantity([self._reference_longitude, self._reference_latitude])
+        w2.wcs.lonpole = 180. if w2.wcs.crval[1] < w2.wcs.theta0 else 0.
         w2.wcs.ctype = self.coordinate_system
         w2.wcs.pc = self.rotation_matrix
         # FITS standard doesn't allow both PC_ij *and* CROTA keywords
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -546,7 +546,6 @@ class GenericMap(NDData):
         # array with sequence error.
         w2.wcs.cdelt = u.Quantity(self.scale)
         w2.wcs.crval = u.Quantity([self._reference_longitude, self._reference_latitude])
-        w2.wcs.lonpole = 180. if w2.wcs.crval[1] < w2.wcs.theta0 else 0.
         w2.wcs.ctype = self.coordinate_system
         w2.wcs.pc = self.rotation_matrix
         # FITS standard doesn't allow both PC_ij *and* CROTA keywords
--- a/changelog/5448.bugfix.rst
+++ b/changelog/5448.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed the incorrect value for the FITS WCS ``LONPOLE`` keyword when using :func:`~sunpy.map.make_fitswcs_header` for certain combinations of WCS projection and reference coordinate.
--- a/sunpy/map/header_helper.py
+++ b/sunpy/map/header_helper.py
@@ -175,6 +175,12 @@ def _set_transform_params(meta_wcs, coordinate, reference_pixel, scale, shape):
     meta_wcs['crval1'], meta_wcs['crval2'] = (coordinate.spherical.lon.to_value(meta_wcs['cunit1']),
                                               coordinate.spherical.lat.to_value(meta_wcs['cunit2']))
 
+    # When the native latitude of the fiducial point is 0 degrees, which is typical for cylindrical
+    # projections, the correct value of `lonpole` depends on whether the world latitude of the
+    # fiducial point is greater than or less than its native latitude.
+    meta_wcs['LONPOLE'] = 180. if meta_wcs['crval2'] < meta_wcs['theta0'] else 0.
+    del meta_wcs['theta0']  # remove the native latitude of the fiducial point
+
     # Add 1 to go from input 0-based indexing to FITS 1-based indexing
     meta_wcs['crpix1'], meta_wcs['crpix2'] = (reference_pixel[0].to_value(u.pixel) + 1,
                                               reference_pixel[1].to_value(u.pixel) + 1)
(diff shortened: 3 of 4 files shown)


ayshih:
There's a real failing figure test with a Helioviewer JPEG2000 file.  I haven't looked into it, but my guess is that it's related to the fact that the Helioviewer files preserve the original FITS header information even though it may no longer be consistent with the data (cf. https://github.com/sunpy/sunpy/issues/4553#issuecomment-706196247).

sunpy/map/mapbase.py
```diff
@@ -546,6 +546,7 @@ def wcs(self):
         # array with sequence error.
         w2.wcs.cdelt = u.Quantity(self.scale)
         w2.wcs.crval = u.Quantity([self._reference_longitude, self._reference_latitude])
+        w2.wcs.lonpole = 180. if w2.wcs.crval[1] < w2.wcs.theta0 else 0.
```
dstansby, line 549: Why are we overwriting the value if it's given in the header here? And doesn't `wcslib` take care of setting the appropriate default value if it isn't given in the header?

ayshih: Well, hmm.  Presumably the reason this entire code block exists is that we don't want to create the WCS from the header values directly.  The WCS is created above *without* using header values, and then all of these lines set WCS values based on our properties.  If you manually change the reference coordinate, you have to manually update `lonpole` to be the appropriate default value.

To fully follow this model, I can create a `GenericMap._lonpole` (private) property that would use the `lonpole` keyword if it is in the header, otherwise return the appropriate default.

dstansby: To expand on my comment, it seems like `wcslib` automatically sets LONPOLE if it's not in the header. So this might be an upstream `wcslib` issue, or an issue with a particular data source. Do you have a self contained example that motivated this PR?
```python
from sunpy.map import Map

m = Map(AIA_193_IMAGE)
m.meta.pop('lonpole', None)
print('lonpole' in m.meta)  # False
print(m.wcs.to_header()['LONPOLE'])  # 180
```

ayshih: This is an issue with how we construct the WCS.

The following line creates the WCS instance without using many of the header values:
https://github.com/sunpy/sunpy/blob/8d24c7ae268dfcb6d05f52193847333e9a166b5a/sunpy/map/mapbase.py#L526
The created WCS instance has `LONPOLE` set to its appropriate default value for the _default_ value of the reference coordinate (e.g., (0, 0) in Stonyhurst heliographic coordinates).

Then, the following line (potentially) modifies the reference coordinate in the WCS instance to something other than the default reference coordinate (for example, to one with a negative latitude in Stonyhurst heliographic coordinates):
https://github.com/sunpy/sunpy/blob/8d24c7ae268dfcb6d05f52193847333e9a166b5a/sunpy/map/mapbase.py#L548

However, the value for `LONPOLE` has _already been set_ in the WCS instance, and so now you're in the situation where `LONPOLE` may no longer be at its appropriate default value for the non-default reference coordinate.  Note that the user may have intentionally set `LONPOLE` to a non-default value, so I don't fault `wcslib` for not automatically updating `LONPOLE` when `CRVAL` is updated.

ayshih: This is the easiest way to show this.  Again, I don't see this as a bug in `wcslib`.

```python
>>> import astropy.units as u
>>> from astropy.wcs import WCS
>>> import sunpy.map

>>> header1 = sunpy.map.make_fitswcs_header(
...     (180, 360),
...     SkyCoord(0*u.deg, 10*u.deg, frame='heliographic_stonyhurst', obstime='2021-07-21'),
...     scale=(1, 1)*u.deg/u.pix,
...     projection_code='CAR'
... )
>>> del header1['lonpole']  # before PR #5448, this value may be incorrect
>>> wcs1 = WCS(header1)
>>> print(wcs1.wcs.crval)
>>> print(wcs1.wcs.lonpole)  # default value is 0 for HGS reference latitude >= 0
[ 0. 10.]
0.0

>>> header2 = sunpy.map.make_fitswcs_header(
...     (180, 360),
...     SkyCoord(0*u.deg, -10*u.deg, frame='heliographic_stonyhurst', obstime='2021-07-21'),
...     scale=(1, 1)*u.deg/u.pix,
...     projection_code='CAR'
... )
>>> del header2['lonpole']  # before PR #5448, this value may be incorrect
>>> wcs2 = WCS(header2)
>>> print(wcs2.wcs.crval)
>>> print(wcs2.wcs.lonpole)  # default value is 180 for HGS reference latitude < 0
[  0. -10.]
180.0

>>> wcs1a = wcs1.copy()
>>> wcs1a.wcs.crval = wcs2.wcs.crval  # modify the reference latitude to be < 0
>>> print(wcs1a.wcs.crval)
>>> print(wcs1a.wcs.lonpole)  # LONPOLE is still at 0 instead of being changed to 180
[  0. -10.]
0.0
```

dstansby: I agree re. the fixes to `make_fitswcs_header`. I don't understand
> The following line creates the WCS instance without using many of the header values:
> https://github.com/sunpy/sunpy/blob/8d24c7ae268dfcb6d05f52193847333e9a166b5a/sunpy/map/mapbase.py#L526

This uses all the information in the FITS header. We then overwrite some of that information using the map properties, but if there is a reference coordinate in the FITS header it is set in the line above, and `wcslib` proceeds to set the correct default `LONPOLE` value. So I'm still not convinced that a change needs to be made in `GenericMap.wcs`. To illustrate this, on the current main branch:
```python
import astropy.units as u
from astropy.wcs import WCS
import sunpy.map
import numpy as np

shape = (180, 360)
header1 = sunpy.map.make_fitswcs_header(
    (180, 360),
    SkyCoord(0*u.deg, -10*u.deg, frame='heliographic_stonyhurst', obstime='2021-07-21'),
    scale=(1, 1)*u.deg/u.pix,
    projection_code='CAR'
)
print(header1['lonpole'])  # 0 (incorrect, but fixed by #5448)
del header1['lonpole']
wcs1 = WCS(header1)
print(wcs1.wcs.lonpole)  # 180 (correct)

data = np.ones(shape)
m = sunpy.map.Map(data, header1)
print(m.wcs.wcs.lonpole)  # 180 (correct)
```

ayshih: Hmm, okay, that's convincing.  I suspect that I must have been testing with a header with a _wrong_ value for `LONPOLE`, as opposed to a missing value, and I got confused and misinterpreted the behavior of this code block.  I'll take out this line.  Thanks!

dstansby (approved):

PR #5449, opened 2021-07-08
Add new date properties to Map

dstansby:
See https://github.com/sunpy/sunpy/pull/5394#issuecomment-875751094, which summarises a recent chat on the dev call.

Still needs:

- [x] Tests
- [x] Changelog entry

### DIFF
--- a/docs/whatsnew/3.1.rst
+++ b/docs/whatsnew/3.1.rst
@@ -34,5 +34,35 @@ We have bumped the minimum version of several packages we depend on; these are t
 
 .. _whatsnew-3.1-contributors:
 
+Changes to map date/time handling
+=================================
+
+New date properties
+-------------------
+The properties `~sunpy.map.GenericMap.date_start`,
+`~sunpy.map.GenericMap.date_end`, and `~sunpy.map.GenericMap.date_average` have
+been added to be drawn from the relevant FITS metadata, if present in the map
+header. These are from new keywords defined in version 4 of the FITS standard,
+which have precise meanings compared to the previously ill-defined DATE-OBS.
+
+Changes to `~sunpy.map.GenericMap.date`
+---------------------------------------
+`sunpy.map.GenericMap.date` now looks for more metadata than just DATE-OBS.
+This property can return any one of the new properties (see above) depending
+on the metadata present in the map. It now draws from, in order of preference:
+
+1. The DATE-OBS FITS keyword
+2. `~sunpy.map.GenericMap.date_average`
+3. `~sunpy.map.GenericMap.date_start`
+4. `~sunpy.map.GenericMap.date_end`
+5. The current time.
+
+If DATE-OBS is present alongside DATE-AVG or DATE-BEG and DATE-END, this results
+in a behaviour change to favour the new (more precisely defined) keywords.
+It is recommended
+to use `~sunpy.map.GenericMap.date_average`,
+`~sunpy.map.GenericMap.date_start`, or `~sunpy.map.GenericMap.date_end`
+instead if you need one of these specific times.
+
 Contributors to this Release
 ============================
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -730,16 +730,13 @@ class GenericMap(NDData):
     def nickname(self, n):
         self._nickname = n
 
-    @property
-    def date(self):
-        """
-        Image observation time.
+    def _get_date(self, key):
+        time = self.meta.get(key, None)
+        if time is None:
+            return
 
-        This is taken from the 'DATE-OBS' FITS keyword.
-        """
-        time = self.meta.get('date-obs', None)
         # Get the time scale
-        if time is not None and 'TAI' in time:
+        if 'TAI' in time:
             # SDO specifies the 'TAI' scale in their time string, which is parsed
             # by parse_time(). If a different timescale is also present, warn the
             # user that it will be ignored.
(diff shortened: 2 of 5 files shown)


changelog/5449.rst
```diff
@@ -0,0 +1,9 @@
+`sunpy.map.GenericMap.date` now looks for more metadata than just DATE-OBS.
+This property can return any one of the new properties (see above) depending
```
wtbarnes, line 2:
```suggestion
This property can return any one of the new properties depending
```
What is above? Is this referencing the previous changelog?

docs/whatsnew/3.1.rst
```diff
@@ -34,5 +34,27 @@ We have bumped the minimum version of several packages we depend on; these are t
...
+New date properties
+-------------------
+The properties `~sunpy.map.GenericMap.date_beg`,
+`~sunpy.map.GenericMap.date_end`, and `~sunpy.map.GenericMap.date_avg` have
+been added to be drawn from the relevant FITS metadata, if present in the map
+header.
+
+Changes to `~sunpy.map.GenericMap.date`
+---------------------------------------
+`sunpy.map.GenericMap.date` now looks for more metadata than just DATE-OBS.
```
wtbarnes, line 49: Should we reference the FITS-4 standard in here somewhere to explain why we are making this change?

sunpy/map/mapbase.py
```diff
@@ -752,6 +749,62 @@ def date(self):
...
+        5. The current time
+        """
+        time = self.date_avg
+        time = time or self._get_date('date-obs')
+        time = time or self.date_beg
+        time = time or self.date_end
+
         if time is None:
             if self._default_time is None:
                 warn_metadata("Missing metadata for observation time, "
```
wtbarnes, line 810: This warning message now feels incomplete as there are other ways to suppress this warning than by setting 'DATE-OBS'.

sunpy/map/tests/test_mapbase.py
```diff
@@ -135,8 +134,30 @@ def test_nickname_set(generic_map):
...
 
-def test_date(generic_map):
-    assert isinstance(generic_map.date, Time)
+date_dict = {'DATE-AVG': parse_time('2020-01-01'),
+             'DATE-OBS': parse_time('2020-02-01'),
+             'DATE-BEG': parse_time('2020-03-01'),
+             'DATE-END': parse_time('2020-03-03')}
+date_begend = date_dict['DATE-BEG'] + (date_dict['DATE-END'] - date_dict['DATE-BEG']) / 2
+
+
```
wtbarnes, line 143: What about the case where they are all missing?

wtbarnes (commented):
Just a few minor suggestions, but this all looks good to me.

A lingering worry that came up in the dev call last week: this is potentially a breaking change. If a map has both `DATE-OBS` and `DATE-AVG` (or equivalently a `DATE-BEG` and a `DATE-END`), a release following this PR will return the latter for `.date`without any sort of warning as to the change in behavior whereas 3.0 returns the former. I'm not completely sure how we would handle a deprecation here. Maybe switching the order of `date_avg` and `date-obs` in `.date` and then adding a deprecation warning that we will prefer `date-avg` in 4.1?

changelog/5449.rst
dstansby: Woops, this is hungover from when I just copied and pasted from the what's new.

dstansby: Should be fixed now.

sunpy/map/mapbase.py
dstansby: I've swapped to DATE-AVG as the recommended way to insert a date, does that seem better?

docs/whatsnew/3.1.rst
dstansby: I've updated the changelog and what's new with a bit of motivation.

sunpy/map/tests/test_mapbase.py
dstansby: I've added a test for this.

dstansby:
> Just a few minor suggestions, but this all looks good to me.
> 
> A lingering worry that came up in the dev call last week: this is potentially a breaking change. If a map has both `DATE-OBS` and `DATE-AVG` (or equivalently a `DATE-BEG` and a `DATE-END`), a release following this PR will return the latter for `.date`without any sort of warning as to the change in behavior whereas 3.0 returns the former. I'm not completely sure how we would handle a deprecation here. Maybe switching the order of `date_avg` and `date-obs` in `.date` and then adding a deprecation warning that we will prefer `date-avg` in 4.1?

I would argue it's a bugfix? But maybe not, perhaps we can discuss this point again on this weeks' call.

sunpy/map/mapbase.py
```diff
@@ -752,15 +749,71 @@ def date(self):
...
+        return parse_time(time, scale=timesys.lower())
+
+    @property
+    def date_beg(self):
+        """
+        Time of the beginning of the image acquisition.
+
+        Taken from the DATE-BEG FITS keyword.
+        """
+        return self._get_date('date-beg')
```
Cadair, line 761: I think given the following quote from Section 4.4.2.2 of the FITS 4 spec:

> When the format with a four-digit year is used, [...]. The value of DATE-OBS shall be assumed to refer to the start of an observation, unless another interpretation is clearly explained in the comment field.

We could change this to:

```suggestion
        return self._get_date('date-beg', self._get_date('date-obs'))
```

and remove the special casing from `.date`?

dstansby: If we're not going to check the comment field (which I think is too complicated for us to do and explain in a docstring), then I don't think we can be sure that `DATE-OBS` is the start of the observation, so I'm hesitant make this change.

sunpy/map/mapbase.py
```diff
@@ -752,15 +749,71 @@ def date(self):
             # UTC is the FITS standard default
             timesys = self.meta.get('timesys', 'UTC')
 
+        return parse_time(time, scale=timesys.lower())
+
+    @property
+    def date_beg(self):
```
Cadair, line 755: I see no reason to tie ourselves to the weird short FITS names?

```suggestion
    def date_start(self):
```

sunpy/map/mapbase.py
```diff
@@ -752,15 +749,71 @@ def date(self):
...
+    def date_end(self):
+        """
+        Time of the end of the image acquisition.
+
+        Taken from the DATE-END FITS keyword.
+        """
+        return self._get_date('date-end')
+
+    @property
+    def date_avg(self):
```
Cadair, line 773:
```suggestion
    def date_average(self):
```

or maybe

```suggestion
    def date_center(self):
```

sunpy/map/mapbase.py
dstansby: That sounds good to me.

sunpy/map/mapbase.py
```diff
@@ -730,16 +730,13 @@ def nickname(self):
...
+        time = self.meta.get(key, None)
+        if time is None:
+            return
 
-        This is taken from the 'DATE-OBS' FITS keyword.
-        """
-        time = self.meta.get('date-obs', None)
         # Get the time scale
-        if time is not None and 'TAI' in time:
+        if 'TAI' in time:
```
Cadair, line 739: I think this bit of logic should remain solely in `.date`.

dstansby:
Some more decisions:
- We want to prioritise DATE-OBS in `.date`
- Update the docstring of .date to recommend using the other (new) properties
- Move the TAI logic just to .date
- (in another PR) use date_average instead of .date when constructing the observer coordinate.

sunpy/map/mapbase.py
dstansby: I think worth sticking with `average`, given this is the wording in FITS and this doesn't always have to be the center of the observation interval?

sunpy/map/mapbase.py
dstansby: Thinking about this, it should probably live in map sources? But given it's just raising a warning, I think worth keeping it for all the date fields?

wtbarnes (approved):

PR #5451, opened 2021-07-08
fix order of arguments in scale input

### LINKED ISSUE #5445: Order of arguments in `scale` input to `make_fitswcs_header` is wrong in AIA/EUVI reprojection example
In [the AIA/EUVI reprojection mosaic example](https://docs.sunpy.org/en/latest/generated/gallery/map_transformations/reprojection_aia_euvi_mosaic.html), the ordering of the `scale` argument to `make_fitswcs_header` is incorrect. The ordering should be Cartesian (lon, lat) [according to the `make_fitswcs_header` docstring](https://docs.sunpy.org/en/stable/api/sunpy.map.make_fitswcs_header.html#sunpy.map.make_fitswcs_header), but in this example, the order is according to the array index. This actually has no effect on the example output as the scale in both directions is the same (1 deg/pix), but is potentially confusing and conflicts with the function docstring.

RhnSharma:
### Description
Fix order of arguments in scale input to make_fitswcs_header in reprojection example.

Fixes #5445
Hi @wtbarnes @ayshih, how does this look?
Let me know if this needs any changes.
Thanks

### DIFF
--- a/examples/map_transformations/reprojection_aia_euvi_mosaic.py
+++ b/examples/map_transformations/reprojection_aia_euvi_mosaic.py
@@ -95,8 +95,8 @@ header = sunpy.map.make_fitswcs_header(shape_out,
                                        SkyCoord(0, 0, unit=u.deg,
                                                 frame="heliographic_stonyhurst",
                                                 obstime=maps[0].date),
-                                       scale=[180 / shape_out[0],
-                                              360 / shape_out[1]] * u.deg / u.pix,
+                                       scale=[360 / shape_out[1],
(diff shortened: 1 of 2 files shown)


wtbarnes (approved):
The test failures are unrelated. It looks like the NoRH data source is unavailable for some reason. This looks good to me, though someone who either wrote the examples or has played around more with `make_fitswcs_header` should probably double-check me here just to make sure my original issue was not in error.

dstansby (approved):

Issue #5454, opened 2021-07-12
`Map.date` should default to `None` instead of "now"


wtbarnes:
Following  #5394 and now #5449, there was discussion around what the appropriate default value of `.date` should be.

Currently, `Map.date` defaults to `None` if there is no `DATE-OBS` keyword present in the header (though see #5449 for how this logic will change). This is potentially confusing to users as it can lead to coordinate frames being instantiate at time "now" when this may not be the intended behavior. With `None`, coordinate frames can still be created, but these frames cannot be transformed to another frame without an obstime being *explicitly* specified. 

This would be a breaking change and thus we should think carefully about when to implement it.

dstansby:
I'm 👍 for this - deprecate in 3.1, change behaviour in 4.1?

ayshih:
Some comments:

> This is potentially confusing to users as it can lead to coordinate frames being instantiate at time "now" when this may not be the intended behavior.

A warning is raised with the current behavior, so in principle the user is aware of this being done.  Do you envision still raising a warning with this new behavior?  If there's no warning, a user might get confused in the situation with non-standard date keywords.  If there's a warning, it would be difficult for a user who wants `obstime=None` to construct a header that wouldn't emit that warning.

> With `None`, coordinate frames can still be created, but these frames cannot be transformed to another frame without an obstime being _explicitly_ specified.

Well...  For historical reasons, I think there are still some coordinate transformations that may allow a transformation from `obstime=None` in the source frame to `obstime=<something>` in the destination frame, by assuming that both frames are at the same `obstime`.  I'd probably need to sort that out.  `obstime=None` hasn't exactly been thoroughly tested.

wtbarnes:
> A warning is raised with the current behavior, so in principle the user is aware of this being done. Do you envision still raising a warning with this new behavior? If there's no warning, a user might get confused in the situation with non-standard date keywords. If there's a warning, it would be difficult for a user who wants obstime=None to construct a header that wouldn't emit that warning.

Hmm. That's a fair point. I guess I would tentatively say that there would be no warning and that the user should only be made aware of the fact that `date` is `None` when they try to do something that requires it to be defined (e.g. a coordinate transformation) either through a warning or ideally an exception. I'm not completely sure what you mean by non-standard date keywords. Ideally these would be handled by the map source? (e.g. `AIAMap` which uses `DATE__OBS`).

[...]

PR #5457, opened 2021-07-12
See what happens when None is returned for .date

dstansby:
**DON'T MERGE**

I thought we should see how much of our docs/tests breaks if we do return `None` instead of defaulting to 'now' for `.date`. xref https://github.com/sunpy/sunpy/issues/5454


dstansby:
Looks like (apart from some unrelated servers being down) everything is fine?

wtbarnes:
I think the second point in this comment is worth considering: https://github.com/sunpy/sunpy/issues/5454#issuecomment-878509461. This may not be caught by any of our tests.

dstansby:
This was just an experiment to check what happened with this change, so I'm going to close this, with the conclusion that nothing major broke.

PR #5458, opened 2021-07-13
Added additional time format

SyleKu:
Fits files such as "**_IKARUS_19780529_101651i.fit_**" have a time format of "**_%d/%m/%yT%H:%M:%S_**". With this commit also these kind of time format will be considered.


dstansby:
Thanks for opening the PR. The reason why we don't have this format is because DD/MM/YY and MM/DD/YY are ambiguous, so we can't merge this PR. Is there an issue you're having with reading in specific files? If so, feel free to ask a question over at our discourse, and we can try and help find a solution: https://community.openastronomy.org/c/ask-questions/15

PR #5459, opened 2021-07-14
Fix vso.fetch() docstring

dstansby:
Should fix the doc build, not sure how the docstring ended up like this!


nabobalis (approved):

Cadair (approved):

Cadair:
doc build looks like it's an example error.

Issue #5450, opened 2021-07-08
Wrong date being used for EUI observer coordinate


dstansby:
Currently in sunpy 3.0.x, the date/time of `m.observer_coordinate` is taken from `m.date`, which is taken from the DATE-OBS FITS keyword. For Solar Orbiter (and possibly other missions), the coordinate metadata is valid at the time given by DATE-AVG however, which is different to DATE-OBS (see 3.1.1.10 of the [Metadata Definition for Solar Orbiter Science Data](https://issues.cosmos.esa.int/solarorbiterwiki/display/SOSP/Metadata+Definition+for+Solar+Orbiter+Science+Data)).

This will be fixed in 3.1 by https://github.com/sunpy/sunpy/pull/5449/files, but we should issue a fix for the 3.0.x branch in the EUI map source.

wtbarnes:
Should `.date` on the EUI source just default to 'DATE-AVG'? That seems like the easiest fix here.

dstansby:
Yep, I think that's the best thing to do.

PR #5462, opened 2021-07-16
Fix EUI observer coordinate date

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/5450. Currently I've just passed a custom value to `.observer_coordinate` property, which means that `.observer_coordinate.obstime != .time`, to avoid changing the return value of `.time`.

### DIFF
--- a/changelog/5462.bugfix.rst
+++ b/changelog/5462.bugfix.rst
@@ -1,0 +1,5 @@
+The date returned by `~sunpy.map.GenericMap.date` for Solar Orbiter/EUI maps
+has been adjusted to be taken from the DATE-AVG keyword
+(the middle of the image acquisition period), instead of the DATE-OBS
+keyword (the beginning of the image acquisition period). This means the observer
+coordinate now has the correct date.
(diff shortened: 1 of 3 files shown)


ayshih:
Does this mean that `.coordinate_frame.obstime != .coordinate_frame.observer.obstime`?  (This isn't necessarily a problem, but is better avoided if possible.)

dstansby:
> Does this mean that `.coordinate_frame.obstime != .coordinate_frame.observer.obstime`? (This isn't necessarily a problem, but is better avoided if possible.)

Yes, it does, because `.coordinate_frame` draws a date from `.wcs`, which in turn draws from `.date`.

The other way to go with this is to override `.date` to return DATE-AVG instead of DATE-OBS, which might be better for overall consistency? (and was suggested by @wtbarnes here: https://github.com/sunpy/sunpy/issues/5450#issuecomment-876524412)

wtbarnes:
I'm more inclined to override `.date`. This mostly matters for bugfix releases on `3.0.x` though as #5449 will solve this problem by returning `DATE-AVG` from `.date` anyway.

wtbarnes (approved):

PR #5464, opened 2021-07-16
Add plot settings for HMI synoptic map

wtbarnes:
### Description

This PR sets the default colormap for HMI synoptic maps to 'hmimag' and the default normalization to be linear between -1.5e3 and +1.5e3 (the range that this particular colorbar is intended for).

PR #5467, opened 2021-07-18
More cleanup to HMI synoptic map example

dstansby:
- After https://github.com/sunpy/sunpy/pull/5464, no longer need to specify cmap/norm
- Clean up title (a single Carrington rotation is refers to the whole interval of that rotation, so don't do N - N+1 in the title)
- Fix the unit attribute, since Mx isn't in the list of units the IAU guide allows (so astropy doesn't recognise it)

### DIFF
--- a/sunpy/map/sources/tests/test_hmi_synoptic_source.py
+++ b/sunpy/map/sources/tests/test_hmi_synoptic_source.py
@@ -123,7 +123,7 @@ def test_measurement(hmi_synoptic):
 
 
 def test_unit(hmi_synoptic):
-    # Check that the deafult unit of Mx/cm**2 is correctly replaced with a
+    # Check that the default unit of Mx/cm**2 is correctly replaced with a
     # FITS standard unit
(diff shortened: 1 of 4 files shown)


dstansby:
Should probably backport the unit fix to 3.0.x

sunpy/map/sources/tests/test_hmi_synoptic_source.py
```diff
@@ -119,3 +120,11 @@ def test_observatory(hmi_synoptic):
 def test_measurement(hmi_synoptic):
     """Tests the measurement property of the HMISynopticMap object."""
     assert hmi_synoptic.measurement == "carrington"
+
+
+def test_unit(hmi_synoptic):
+    # Check that the deafult unit of Mx/cm**2 is correctly replaced with a
```
nabobalis, line 126:
```suggestion
    # Check that the default unit of Mx/cm**2 is correctly replaced with a
```

nabobalis (approved):

PR #5470, opened 2021-07-19
Added `GenericMap.reproject_to()` to facilitate reprojection

ayshih:
I find the current reprojection workflow to be a bit cumbersome:

- Having to manually create the WCS if starting with a header
- Having to manually call the reprojection function
- Having to manually create the new map
- Having to copy over the plot settings

I wrote a helper method for `GenericMap` that does this all for a user.

```python
import astropy.units as u

import sunpy.map
from sunpy.data.sample import AIA_171_IMAGE

aia = sunpy.map.Map(AIA_171_IMAGE)
header = sunpy.map.make_fitswcs_header(
    (180, 360),
    SkyCoord(0*u.deg, 0*u.deg, frame='heliographic_stonyhurst', obstime=aia.date, rsun=aia.rsun_meters),
    scale=[1, 1]*u.deg/u.pix,
    projection_code='CAR'
)

aia_interp = aia.reproject_to(header)
```

You can easily specify a different algorithm:
```python
aia_adaptive = aia.reproject_to(header, algorithm='adaptive')
```

If you supply a `WCS` instance instead of a header, or the supplied header does not have `NAXISi` keywords, you need to additionally specify the pixel dimensions of the output map via the keyword argument `shape_out`:
```python
from astropy.wcs import WCS
aia.reproject_to(WCS(header), shape_out=(180, 360))
```


examples/map_transformations/reprojection_heliographic_stonyhurst.py
```diff
@@ -29,36 +27,29 @@
...
-# a output image, specified by a different WCS object. Therefore we need to
-# build a `~astropy.wcs.WCS` object describing the output we desire.
-# To do this we use the `sunpy.map.make_fitswcs_header` which assists us in
-# constructing this World Coordinate System (WCS) object.
-# Here we create a WCS based on a heliographic
-# Stonyhurst reference coordinate and with the CAR (plate carree) projection.
+# Reproject works by transforming an input image to a desired World Coordinate
+# System (WCS) projection. Here we use :func:`sunpy.map.make_fitswcs_header`
+# to create a FITS WCS header based on a Stonyhurst heliographic reference
+# coordinate and the CAR (plate carree) projection.
```
dstansby, line 33:
```suggestion
# coordinate and the CAR (plate carrée) projection.
```

sunpy/map/mapbase.py
```diff
@@ -2501,6 +2502,68 @@ def _check_axes(self, axes, warn_different_wcs=False):
...
+        output map.  The ordering of these dimensions is in NumPy order (i.e.,
+        secondary axis and then primary axis).
+        """
+        try:
+            import reproject
+        except ImportError:
+            raise ImportError("This method requires the optional package `reproject`.")
+
+        if not isinstance(output_projection, astropy.wcs.WCS):
+            output_projection = Header(output_projection)
```
dstansby, line 2544: Why `Header` and not `WCS`?

sunpy/map/mapbase.py
```diff
@@ -2501,6 +2502,68 @@ def _check_axes(self, axes, warn_different_wcs=False):
...
+        functions = {'interp': reproject.reproject_interp,
+                     'adaptive': reproject.reproject_adaptive,
+                     'exact': reproject.reproject_exact}
+        if algorithm not in functions:
+            raise ValueError(f"The specified algorithm must be one of: {list(functions.keys())}")
+        func = functions[algorithm]
+
+        # Block the `return_footpoint` keyword argument from being passed on
+        if 'return_footprint' in reproject_args:
+            del reproject_args['return_footprint']
```
dstansby, line 2556:
```suggestion
        reproject_args.pop('return_footprint', None)
```
is one less line?

sunpy/map/mapbase.py
```diff
@@ -2501,6 +2502,68 @@ def _check_axes(self, axes, warn_different_wcs=False):
...
+
+        # Reproject the array
+        output_array = func(self, output_projection, return_footprint=False, **reproject_args)
+
+        # Create and return a new GenericMap
+        if isinstance(output_projection, astropy.wcs.WCS):
+            output_projection = output_projection.to_header()
+        outmap = GenericMap(output_array, output_projection, plot_settings=self.plot_settings)
```
dstansby, line 2564: Do we not want to feed the new map through the map factory so it picks up any source specific overridden properties that correctly interpret non-FITS compliant metadata?

dstansby: And should we also be copying across non-WCS metadata from the original map? (e.g., exposure time, detector name etc.)

sunpy/map/mapbase.py
ayshih: It's because the header can contain the pixel dimensions, but that information is not preserved in the `WCS` instance.

sunpy/map/mapbase.py
ayshih: I know that what you suggest is a code pattern that some use, but I don't like it.  I have a philosophical objection to retrieving a value with no intention of using it.

Regardless, it's moot.  I plan to add a message alerting the user when the `return_footprint` keyword is quashed, and so the `if...del` code pattern has to be used.

sunpy/map/mapbase.py
ayshih: These are thorny considerations.  So far, I have simply approached it the same way as our existing gallery examples: don't use a source-specific subclass, and don't copy over non-WCS metadata.  If the non-WCS metadata is merged with the output header, it might work as intended with a source-specific subclass, but it could also blow up (e.g., if that subclass makes fixes to WCS information).

Cadair:
I want to say that we currently have a GSOC student working on this in ndcube. I am therefore :-1:  on this PR and suggest that we turn this into comments on https://github.com/sunpy/ndcube/pull/434 ?

sunpy/map/mapbase.py
Cadair: It is. It is preserved as `WCS.pixel_shape` https://docs.astropy.org/en/stable/api/astropy.wcs.wcsapi.BaseLowLevelWCS.html#astropy.wcs.wcsapi.BaseLowLevelWCS.pixel_shape

dstansby:
Perhaps we should chat about this on a weekly call. I'm in favour of adding this API to sunpy though, even if internally it gets replaced by an NDCube implementation.

Cadair:
I would much rather we get this directly via inheritance with no wrapper than anything else, I don't need another API to have to sugar coat when we inherit from ndcube (which hopefully will be the next release or the one after).

sunpy/map/mapbase.py
ayshih: Ugh.  You are correct that it is available in the WCS instance, but that means `reproject` has a bug/feature in that it does not make use of `.pixel_shape` (or rather, `.array_shape` for the NumPy ordering of dimensions).  `reproject` shouldn't require that `shape_out` be specified if `.array_shape` exists.

ayshih: Okay, fixed/worked around.  @Cadair can explain whether this is a bug or a feature in `reproject`.

examples/map_transformations/reprojection_heliographic_stonyhurst.py
ayshih: Fiixed

sunpy/map/mapbase.py
ayshih: Heh, I came up with a cleaner approach.  Why remove it at all if I'm always going to add it back in later?

sunpy/map/mapbase.py
Cadair: That's probably a feature request for reproject if you fancy opening it there.

ayshih:
We should definitely discuss this at a dev meeting.

I don't see how you can get away from having a wrapper in core given how we currently implement metadata and subclass instantiation, but the sticking points will also be overhauled too.  If the goal is direct inheritance, you're going to want the NDCube API to be a superset of the simple API in this PR, and aside from algorithm selection, it basically already is.

I'd to see this get into core before the NDCube future, not only because people will get value out of it immediately, but also because we might get feedback that could inform the transition to NDCube.

Cadair:
> We should definitely discuss this at a dev meeting.

:+1: 

> I don't see how you can get away from having a wrapper in core given how we currently implement metadata and subclass instantiation, but perhaps the sticking points will also be overhauled with the NDCube transition. If the goal is direct inheritance, you're going to want the NDCube API to be a superset of the simple API in this PR, and aside from algorithm selection, it basically already is.

I can't think of a reason for this off the top of my head, what are you thinking of?

Cadair:
> aside from algorithm selection, it basically already is.

This is on the top of our todo list for the GSOC project, so we are aiming to have it in by the 2.0 release of ndcube.

ayshih:
The API is now a compatible subset of the API for NDCube's `reproject_to()`.  When `Map` transitions to using `NDCube`, there should be no deprecation period needed.

Review away!

setup.cfg
```diff
@@ -78,6 +78,7 @@ tests =
   pytest-intercept-remote>=1.2
   pytest-mock
   pytest-mpl>=0.12 # First version to support our figure tests
+  reproject
```
Cadair, line 81: Should we add reproject as a map dependency at this point? If not should we add it's own extra? i.e. do we think that `install sunpy[all]` should bring in reproject? reproject [has minimal requirements](https://github.com/astropy/reproject/blob/master/setup.cfg#L19-L23) (healpix has no requirements tho is all C) though it is a bunch of C/Cython to compile.

ndcube [added it as an optional](https://github.com/sunpy/ndcube/blob/main/setup.cfg#L54)

sunpy/map/mapbase.py
```diff
@@ -2584,6 +2587,74 @@ def _check_axes(self, axes, warn_different_wcs=False):
...
+
+        See the respective documentation for these functions for additional keyword
+        arguments that are allowed.
+
+        .. minigallery:: sunpy.map.GenericMap.reproject_to
+        """
+        try:
+            import reproject
+        except ImportError:
+            raise ImportError("This method requires the optional package `reproject`.")
```
Cadair, line 2628:
```suggestion
        except ImportError as exc:
            raise ImportError("This method requires the optional package `reproject`.") from exc
```

sunpy/map/mapbase.py
```diff
@@ -2584,6 +2587,74 @@ def _check_axes(self, axes, warn_different_wcs=False):
...
+
+        if not isinstance(target_wcs, astropy.wcs.WCS):
+            target_wcs = astropy.wcs.WCS(target_wcs)
```
Cadair, line 2631: This is not the best way around. It would be better to check if it's a header-like object (`isinstance(target_wcs, collections.Mapping)`).

https://github.com/sunpy/ndcube/blob/2dff8c1d90f1c43b1eafc3be4c0bd86ad04b8957/ndcube/ndcube.py#L783-L784

sunpy/map/mapbase.py
```diff
@@ -2584,6 +2587,74 @@ def _check_axes(self, axes, warn_different_wcs=False):
...
+        # reproject does not automatically grab the array shape from the WCS instance
+        if target_wcs.array_shape is not None:
+            reproject_args.setdefault('shape_out', target_wcs.array_shape)
+
+        # Reproject the array
+        output_array = func(self, target_wcs, **reproject_args)
+
+        # Create and return a new GenericMap
+        outmap = GenericMap(output_array, target_wcs.to_header(),
+                            plot_settings=self.plot_settings)
```
Cadair, line 2655: hmmm this is a little quirk between ndcube and map. This makes more sense for map, but ndcube does `type(self)()`

sunpy/map/mapbase.py
```diff
@@ -2584,6 +2587,74 @@ def _check_axes(self, axes, warn_different_wcs=False):
...
+                     'adaptive': reproject.reproject_adaptive,
+                     'exact': reproject.reproject_exact}
+        if algorithm not in functions:
+            raise ValueError(f"The specified algorithm must be one of: {list(functions.keys())}")
+        func = functions[algorithm]
+
+        # Block the `return_footpoint` keyword argument from being passed on
+        if reproject_args.get('return_footprint', False):
+            log.warn("This method does not return the footprint, so `return_footprint` is ignored.")
+        reproject_args['return_footprint'] = False
```
Cadair, line 2644: This is also different to ndcube. `NDCube.reproject_to` will optionally return `new_cube, footprint`.

When using `exact` the footprint can be quite useful. I don't feel like there's a good reason not to optionally return the footprint to match ndcube?

Cadair (changes requested):
Generally looks good. There's a few places where it diverges from the ndcube choices.

sunpy/map/mapbase.py
ayshih: Why is that better?  I need a `WCS` instance for the rest of this code.  I don't care the type of the input as long as the `WCS()` constructor works.

sunpy/map/mapbase.py
ayshih: There's no good reason not to; I just didn't implement it for simplicity.  This is the primary reason that I said that the API was a subset, because it does less than the NDCube version.  But, yes, I can implement it the same way.

ayshih: Functionality and test added

sunpy/map/mapbase.py
ayshih: Fixed

setup.cfg
ayshih: <shrug>

sunpy/map/mapbase.py
Cadair: Because reproject supports WCSes which aren't of type `astropy.wcs.WCS`. Sure *for now* map doesn't support that but why not write it the way around which is more correct and future proof?

Cadair: Given we will be adopting the ndcube version before we support any other type of WCS this is a real irrelevant nit.

ayshih: Ah, I see what you mean.  I was approaching it from the perspective that I needed it to be `astropy.wcs.WCS` to rely on the existence of `.array_shape` (and, as currently written, `.to_header()`), and that maybe the constructor would accept other inputs than `Mapping` in the future.  Sounds like we can deal with the future possibilities in the future.

ayshih:
Added a whatsnew

sunpy/map/mapbase.py
Cadair: `.array_shape` is an APE 14 property, so that one is fine. `to_header` is `wcs.WCS` specific obviously.

setup.cfg
Cadair: After discussing this at the community meeting today, I am pro adding reproject to the map extra.

wtbarnes:
I am fine with adding `reproject` as an optional dependency under `[map]` or potentially under its own optional dependency category if we don't want to be as tied to it as a core dependency. But as was discussed at the community meeting today, `reproject` is becoming increasingly important, particularly in our example gallery. 

At the least, I am against it being a "hidden" optional dependency (a la `astroquery`).

setup.cfg
ayshih: Done

sunpy/map/tests/test_reproject_to.py
```diff
@@ -0,0 +1,99 @@
...
+    aia171_test_map.reproject_to(hpc_header, algorithm='exact').plot()
+
+
+@figure_test
+def test_reproject_to_hpc_adaptive(aia171_test_map, hpc_header):
+    aia171_test_map.reproject_to(hpc_header, algorithm='adaptive').plot()
+
+
+def test_return_footprint(aia171_test_map, hpc_header):
+    pytest.importorskip("reproject")
```
dstansby, line 80: Should this whole test file have this skip instead of just this function?

sunpy/map/tests/test_reproject_to.py
```diff
@@ -0,0 +1,99 @@
...
+
+        return_with_footprint = aia171_test_map.reproject_to(hpc_header, return_footprint=True)
+        assert len(return_with_footprint) == 2
+        assert isinstance(return_with_footprint[0], sunpy.map.GenericMap)
+        assert isinstance(return_with_footprint[1], np.ndarray)
+
+
+def test_invalid_inputs(aia171_test_map, hpc_header):
+    pytest.importorskip("reproject")
+    with pytest.raises(ValueError):
```
dstansby, line 98: Would be good practice to add `matches='part of the error string'` here

dstansby:
How do we see the newly added test images? They don't seem to be showing up on the cricleCI artefacts.

docs/whatsnew/3.1.rst
```diff
@@ -81,5 +82,16 @@ are not written to provide valid results for times before 1972-01-01.
 This format is equivalent to `~astropy.time.TimeUnixTai`, except that the epoch
 is 12 years earlier.
 
+.. _whatsnew-3.1-reproject-to:
+
+Convenient reprojection of maps
+===============================
+
+`~sunpy.map.Map` objects now have the :meth:`~sunpy.map.GenericMap.reproject_to` method to easily reproject the map to a new WCS.
+The returned map will be of type `~sunpy.map.GenericMap`, with no metadata preserved from the original map, so copy over any desired metadata from the original map.
```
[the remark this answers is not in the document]
ayshih: Well, no, nothing is copied over.  You're reprojecting onto a new WCS, so that will be the WCS metadata.  None of the old WCS metadata is preserved.

sunpy/map/tests/test_reproject_to.py
ayshih: I don't know if you can skip a whole file, but these checks are actually now a bit superfluous now that we've agreed to make `reproject` an extra dependency

docs/whatsnew/3.1.rst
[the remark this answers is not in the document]
dstansby: D'oh what was I thinking?

ayshih:
test_reproject_to_hgs:
sunpy map tests test_reproject_to test_reproject_to_hgs
test_reproject_to_hgs_wcs:
sunpy map tests test_reproject_to test_reproject_to_hgs_wcs
test_reproject_to_hpc:
sunpy map tests test_reproject_to test_reproject_to_hpc
test_reproject_to_hpc_adaptive:
sunpy map tests test_reproject_to test_reproject_to_hpc_adaptive
test_reproject_to_hpc_exact:
sunpy map tests test_reproject_to test_reproject_to_hpc_exact
test_reproject_to_hpc_interpolation:
sunpy map tests test_reproject_to test_reproject_to_hpc_interpolation

dstansby:
Could a couple of those images be removed by using [`check_figures_equal`](https://matplotlib.org/stable/api/testing_api.html?highlight=check_figures_equal#matplotlib.testing.decorators.check_figures_equal)? In particular:

- `test_reproject_to_hgs_wcs` has same output as `test_reproject_to_hgs`
- `test_reproject_to_hpc_interpolation` has same output as `test_reproject_to_hpc`

Might be good to reduce duplication/bloat in the figure test images repo?

ayshih:
Heh, I didn't even realize `check_figures_equal()` existed.  I'll give that a try

sunpy/map/tests/test_reproject_to.py
ayshih: Fixed

sunpy/map/tests/test_reproject_to.py
ayshih: Removed

sunpy/map/mapbase.py
```diff
@@ -2578,6 +2581,85 @@ def _check_axes(self, axes, warn_different_wcs=False):
 
         return axes
 
+    def reproject_to(self, target_wcs, algorithm='interpolation', return_footprint=False,
+                     **reproject_args):
+        """
+        Reproject the map to a different WCS
```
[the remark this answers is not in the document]
ayshih: Added

ayshih:
Okay, I've re-implemented two of the tests to check for image equality

dstansby (approved):

dstansby (approved):

nabobalis (approved):

sunpy/map/tests/test_reproject_to.py
```diff
@@ -0,0 +1,123 @@
...
+                 obstime=aia171_test_map.date,
+                 observer=new_observer,
+                 rsun=aia171_test_map.coordinate_frame.rsun),
+        scale=u.Quantity(aia171_test_map.scale),
+        projection_code='TAN'
+    )
+
+
+@figure_test
+@pytest.mark.filterwarnings('ignore:Missing metadata')  # TODO: fix bug for HGS maps
```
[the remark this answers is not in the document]
ayshih: Yup, looks like the fix in #5478 makes this filtering unnecessary.

Cadair (approved):

PR #5477, opened 2021-07-22
`GenericMap.plot()` can now plot to a figure

ayshih:
This PR slightly streamlines the non-pyplot workflow for plotting Maps by removing the need to explicitly create the axes for the usual use case of plotting to the Map's own WCS.

Before this PR:
```python

fig = Figure()
ax = fig.add_subplot(projection=aiamap)
aiamap.plot(axes=ax)
```

After this PR:
```python

fig = Figure()
aiamap.plot(figure=fig)
```

Earth-shattering, I know.  Thoughts?

### DIFF
--- a/changelog/5477.feature.rst
+++ b/changelog/5477.feature.rst
@@ -1,0 +1,2 @@
+It is now possible to explicitly specify the target `~matplotlib.figure.Figure` instance when using :meth:`sunpy.map.GenericMap.plot`, via the ``figure`` keyword argument.
+This is useful when not using the `matplotlib.pyplot` workflow, because there may not be a notion of a "current" figure, and saves having to manually create the `~matplotlib.axes.Axes` in common use cases.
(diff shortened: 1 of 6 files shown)


dstansby:
I think I'm 👎 on this change. There's no precedent for API to plot directly to a figure (and create an Axes in the process) in Matplotlib, so I think it's better requiring users to explicitly create an Axes to plot on.

ayshih:
I mean, doesn't `pyplot`-based API do that all the time?
```python
plt.figure()
aiamap.plot()  # plots directly to the current pyplot figure, and creates an Axes in the process
```
(Heck, because it's `pyplot`, you can sometimes get away with not calling `plt.figure()` explicitly, and then `.plot()` is creating the `Figure` as well.)  How is that less objectionable than plotting directly to an `Axes`-less `Figure`?

dstansby:
I'm against mixing the implicit (pyplot) and explicit (`Figure()`) API. I'm also against adding API that implies maps are being plotted to a `Figure`, when in fact they are always plotted to an `Axes`, and don't think the saving on one line of user-code is worth the potential confusion here.

ayshih:
Based on discussion at the 2021-09-15 dev meeting, I'm closing this until someone other than me expresses an interest in having this functionality.

PR #5478, opened 2021-07-22
Better handling of metadata/properties for Stonyhurst heliographic maps

ayshih:
- Copied from #5246: Don't set observer information for the WCS returned for Stonyhurst heliographic maps
- Disrupt the "vicious cycle" for our implementation of `.observer_coordinate` and `.wcs` (see https://github.com/sunpy/sunpy/pull/5246#issuecomment-828575070) by defining a private function to determine the frame class from `CTYPEi` rather than having to build the entire frame instance from `.wcs`.
- Fixed the `.rsun_meters` property so that it does not use `.dsun` unless it needs to, because `.dsun` will use `.observer_coordinate`.  For Stonyhurst heliographic maps without observer information, that will emit a irrelevant warning that an Earth-based observer is being assumed.

### DIFF
--- a/sunpy/coordinates/wcs_utils.py
+++ b/sunpy/coordinates/wcs_utils.py
@@ -153,7 +153,7 @@ def solar_wcs_frame_mapping(wcs):
     if rsun is not None:
         frame_args['rsun'] = rsun
 
-    frame_class = _frame_class_from_ctypes(wcs.wcs.ctype)
+    frame_class = _sunpy_frame_class_from_ctypes(wcs.wcs.ctype)
 
     if frame_class:
         if frame_class == HeliographicStonyhurst:
@@ -164,7 +164,7 @@ def solar_wcs_frame_mapping(wcs):
         return frame_class(**frame_args)
 
 
-def _frame_class_from_ctypes(ctypes):
+def _sunpy_frame_class_from_ctypes(ctypes):
     # Truncate the ctype to the first four letters
     ctypes = {c[:4] for c in ctypes}
 
--- a/sunpy/coordinates/wcs_utils.py
+++ b/sunpy/coordinates/wcs_utils.py
@@ -153,22 +153,31 @@ def solar_wcs_frame_mapping(wcs):
     if rsun is not None:
         frame_args['rsun'] = rsun
 
+    frame_class = _frame_class_from_ctypes(wcs.wcs.ctype)
+
+    if frame_class:
+        if frame_class == HeliographicStonyhurst:
+            frame_args.pop('observer', None)
+        if frame_class == Heliocentric:
+            frame_args.pop('rsun', None)
+
+        return frame_class(**frame_args)
+
+
+def _frame_class_from_ctypes(ctypes):
     # Truncate the ctype to the first four letters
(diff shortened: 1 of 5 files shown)


sunpy/coordinates/wcs_utils.py
```diff
@@ -153,22 +153,31 @@ def solar_wcs_frame_mapping(wcs):
...
+    mapping = {
+        Helioprojective: {'HPLN', 'HPLT'},
+        HeliographicStonyhurst: {'HGLN', 'HGLT'},
+        HeliographicCarrington: {'CRLN', 'CRLT'},
+        Heliocentric: {'SOLX', 'SOLY'},
+    }
+
+    for frame_class, ctype_pair in mapping.items():
+        if ctype_pair <= ctypes:
+            return frame_class
```
Cadair, line 180: I need to think about this more, but this limits us to only our frames, where as at the moment there's no reason you couldn't make a map in equatorial coordinates or anything else in astropy, for that reason I am not really in favour of it.

ayshih: Well, this private function is used in two places:

- In `wcs_utils.py`, it merely breaks out the behavior that is currently in `solar_wcs_frame_mapping()`.  That is, `solar_wcs_frame_mapping()` didn't return a frame for such non-SunPy frames, and it still doesn't.
- In `mapbase.py`, it's used to determine whether we should include the observer information in the WCS instance.  What I can do is change this PR so that  `GenericMap.wcs` includes the observer information unless both the `CTYPEi` corresponds to a SunPy frame class and that frame class does not have the `observer` frame attribute.   Thus, the behavior for non-SunPy frames is unchanged.

ayshih: > What I can do is change this PR so that  `GenericMap.wcs` includes the observer information unless both the `CTYPEi` corresponds to a SunPy frame class and that frame class does not have the `observer` frame attribute.   Thus, the behavior for non-SunPy frames is unchanged.

Done

sunpy/map/mapbase.py
```diff
@@ -564,24 +564,21 @@ def wcs(self):
...
-                                                                      rsun=self.rsun_meters))
-
-        sunpy.coordinates.wcs_utils._set_wcs_aux_obs_coord(w2, obs_coord)
+        # Set observer coordinate information except when we know it is not appropriate (e.g., HGS)
+        sunpy_frame = sunpy.coordinates.wcs_utils._sunpy_frame_class_from_ctypes(w2.wcs.ctype)
+        if sunpy_frame is None or hasattr(sunpy_frame, 'observer'):
+            # Clear all the aux information that was set earlier. This is to avoid
+            # issues with maps that store multiple observer coordinate keywords.
+            # Note that we have to create a new WCS as it's not possible to modify
+            # wcs.wcs.aux in place.
```
Cadair, line 573: Unrelated but when we are 4.2+ we can drop this and rely on: https://github.com/astropy/astropy/pull/11166

Cadair: Have added this to the tracking issue.

Cadair (approved):
Looks ok to me.

PR #5480, opened 2021-07-22
Deprecate 'neighbor' as a resample method

dstansby:
As far as I can tell this is just the same as "nearest", and we don't need to roll our own code since scipy handles this fine?

### DIFF
--- a/sunpy/image/tests/test_resample.py
+++ b/sunpy/image/tests/test_resample.py
@@ -8,6 +8,7 @@ import astropy.units as u
 import sunpy.data.test
 import sunpy.map
 from sunpy.image.resample import reshape_image_to_4d_superpixel
+from sunpy.util.exceptions import SunpyDeprecationWarning
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


nabobalis (approved):

sunpy/image/tests/test_resample.py
```diff
@@ -39,7 +40,9 @@ def resample_method(aia171_test_map, method):
 
 
 def test_resample_neighbor(aia171_test_map):
-    resample_method(aia171_test_map, 'neighbor')
+    with pytest.warns(SunpyDeprecationWarning,
+                      match='Using "neighbor" as a method for resampling is deprecated.'):
+        resample_method(aia171_test_map, 'neighbor')
```
wtbarnes, line 45: Could you maybe add a test here that ensures that `"nearest"` and `"neighbor"` do give equivalent results? It can then be removed once the deprecation period has passed.

wtbarnes (commented):
This looks fine to me. I'm all for removing old redundant code. I think it would be good to have one extra test that ensures that `nearest` and `neighbor` give equivalent results

PR #5374, opened 2021-05-30
Known Issues Page

nabobalis:
I wasn't really sure what angle to take with it.

Anyone have any ideas?

PR #5489, opened 2021-08-06
Add new time format for parsing TAI seconds since 1958-01-01

wtbarnes:
### Description

This PR adds a new `astropy.time.Time` format for parsing the number of TAI (leap) seconds since 1958-01-01 00:00:00. This  was originally used in the SOHO CDS timing software (from what I can tell, see https://hesperia.gsfc.nasa.gov/ssw/gen/idl/time/aaareadme.txt) and is the format returned by the [`anytim2tai` SSW routine](https://hesperia.gsfc.nasa.gov/ssw/gen/idl/time/anytim2tai.pro).

I should say I am not dead set on the name for this time format. 'tai_cds' is a bit awkward and cryptic.

Comparisons between `astropy.time.Time` and `anytim2tai` show that after 1972-01-01, the difference between UTC and TAI is the same for both methods. However, between 1958-01-01 and 1972-01-01, the difference as returned  by `anytim2tai` is 9 s (jumping to 10 at 1972-01-01) while for `astropy.time.Time`, the difference increases roughly linearlly from 0 to 10. While this difference is not critical for missions after 1972, this should still be documented on our "Known Issues" page (see #5374).

image


dstansby (changes requested):
The oldestdeps build failure looks like it's caused by this PR.

wtbarnes:
Yeah, it looks like the `erfa` package was not an explicit dependency of astropy then?

wtbarnes:
So prior to astropy 4.2, PyERFA was included in astropy as a private module `_erfa`: https://docs.astropy.org/en/stable/whatsnew/4.2.html#removed-erfa-module.

wtbarnes:
I'm not sure why my import on the test side wasn't just the same as my import in the actual code for the time format class. I guess the question is, does 

```python
from astropy.coordinates.sky_coordinate import erfa
```

import the same thing as 

```python
from astropy.time.formats import erfa
```

Presumably the answer is yes? They are both going through astropy.

ayshih:
I'd like significant changes to the names and documentation:

- I propose the scale be named `tai_seconds` or something similar.
- The documentation should not state that CDS originated this scale.  Even if that's true – and I'm not sure that it is – it should be relegated to a note rather than as part of the name or as motivation.
- The documentation should not discuss `anytim2tai.pro` as if it were a reference implementation.  The differences with `anytim2tai.pro` should either be moved to docstring notes or to Known Issues.  `anytim2tai.pro` is flawed in two ways for pre-1972 times:
  - UT and TAI were synchronized by definition on 1958 Jan 1 00:00:00.  `anytim2tai.pro` should return 0 at that time.
  - Even if `anytim2tai.pro` is intentionally choosing ignoring the difference in lengths of UT seconds and TAI seconds prior to 1972, it should be returning 10 for pre-1972 times, not 9.  Its implementation acts as if there was a leap second added on 1972 Jan 1, but there was not.

wtbarnes:
> * I propose the scale be named `tai_seconds` or something similar.

Done. I've now renamed it to `TimeTAISeconds` with the format string `'tai_seconds'`

> * The documentation should not state that CDS originated this scale.  Even if that's true – and I'm not sure that it is – it should be relegated to a note rather than as part of the name or as motivation.

Fair. I've moved this to a note

> * The documentation should not discuss `anytim2tai.pro` as if it were a reference implementation.  The differences with `anytim2tai.pro` should either be moved to docstring notes or to Known Issues.  `anytim2tai.pro` is flawed in two ways for pre-1972 times:
>   
>   * UT and TAI were synchronized by definition on 1958 Jan 1 00:00:00.  `anytim2tai.pro` should return 0 at that time.
>   * Even if `anytim2tai.pro` is intentionally choosing to ignore the difference in lengths of UT seconds and TAI seconds prior to 1972, it should be returning 10 for pre-1972 times, not 9.  Its implementation acts as if there was a leap second added on 1972 Jan 1, but there was not.

Fair point. I did not intend to make it sound like a reference implementation, but I see how it does. I've shortened this to a note and will open a PR to add this to the known issues page.

changelog/5489.feature.rst
```diff
@@ -0,0 +1,4 @@
+Add ``sunpy.time.TimeTAICDS`` to support parsing the TAI time format originally used in the CDS
```
samaloney, line 1:
```suggestion
Add ``sunpy.time.TimeTAISeconds`` to support parsing the TAI time format originally used in the CDS
```

samaloney (approved):

wtbarnes:
I've now called this `TimeUnixTai58` since it is exactly equivalent to `astropy.time.TimeUnixTai`, except that the epoch is 1958-01-01 instead of 1970-01-01. The name is kind of ugly, but is consistent with the astropy convention.

sunpy/time/timeformats.py
```diff
@@ -0,0 +1,75 @@
...
+
+class TimeUnixTai58(TimeFromEpoch):
+    """
+    Seconds from 1958-01-01 00:00:00, including leap seconds.
+
+    This is equivalent to `~astropy.time.TimeUnixTai`, except that the
+    epoch is 12 years earlier.
+
+    .. note:: For dates on and after 1972-01-01 00:00:00 UTC, this format is
+              equivalent to that returned by the `anytim2tai` routine in SSW.
```
dstansby, line 43:
```suggestion
              equivalent to that returned by the ``anytim2tai`` routine in SSW.
```

sunpy/time/timeformats.py
```diff
@@ -0,0 +1,75 @@
...
+              For dates between 1958-01-01 00:00:00 UTC and 1971-12-31 23:59:59 UTC,
+              `anytim2tai` returns a constant difference of 9 s difference between
```
dstansby, line 45:
```suggestion
              ``anytim2tai`` returns a constant difference of 9 s difference between
```

dstansby (commented):
Should fix the doc build

sunpy/time/time.py
```diff
@@ -13,7 +13,7 @@
 from astropy.time import Time, TimeDelta
 
 # This is not called but imported to register it
-from sunpy.time.utime import TimeUTime  # NOQA
+from sunpy.time.timeformats import TimeUTime  # NOQA
```
dstansby, line 16:
```suggestion
from sunpy.time.timeformats import *  # NOQA
```
Do we not want to register the new format too here?

wtbarnes: Yes, but I'm a bit confused as to why this is needed here. I'm already importing it in `time/__init__.py` which should register it already?

dstansby: 👍 makes sense

sunpy/time/tests/test_unixtai58.py
```diff
@@ -0,0 +1,50 @@
+import pytest
+
+from astropy.time import Time, TimeUnixTai
+from astropy.time.formats import erfa
+
+# This registers the TimeTAICDS format with astropy
+from sunpy.time import TimeUnixTai58  # NOQA
```
dstansby, line 7:
```suggestion
import sunpy.time  # NOQA
```
Should work too, and is a good test that we're importing TimeUnixTai58 into the `sunpy.time` namespace? This isn't a blocking suggestion though.

sunpy/time/tests/test_unixtai58.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    Test that the offset between TAI and UTC is 10 s in 1972-01-01, when UTC
+    was defined
+    """
+    t1 = Time('1972-01-01 00:00:00', scale='tai', format='iso')
+    t2 = Time('1972-01-01 00:00:00', scale='utc', format='iso')
+    assert t2.unix_tai_58 - t1.unix_tai_58 == 10.0
+
+
+@pytest.mark.parametrize('time', [
```
dstansby, line 31: This would be a good place to use `hypothesis` to generate test data. Not a blocking suggestion though.

dstansby (approved):

ayshih (changes requested):
The name must be changed.  "Unix" should not be used here.  `TimeUnixTai` is named that way because Unix time is TAI seconds from 1970.  `TimeUnixTai1958` is self-contradicting.

sunpy/time/tests/test_unixtai58.py
```diff
@@ -0,0 +1,50 @@
...
+    Time('1958-01-01T00:00:00', format='isot', scale='tai'),
+    Time('1972-01-01T00:00:00', format='isot', scale='tai'),
+    Time('2015-10-25T05:24:08', format='isot', scale='tai'),
+    Time('2018-09-17T19:46:25', format='isot', scale='tai'),
+])
+def test_unix_tai_consistency(time):
+    diff_unix_tai = Time(TimeUnixTai.epoch_val, scale='tai', format=TimeUnixTai.epoch_format,).tai_seconds
+    assert time.unix_tai + diff_unix_tai == time.tai_seconds
```
ayshih:
```suggestion
```
This test appears to have negligible added value, because you're basically checking that `TimeFromEpoch` really means time from an epoch.  Essentially the only conceivable way that it could fail is if someone accidentally changes the `epoch_scale` of `TimeUnixTai` from TAI to something else, and that would presumably cause test failures in Astropy.

ayshih (commented):
Thanks for merging in wtbarnes/sunpy#2!

I've exchanged emails with Bill Thompson and Dominic Zarro, and Bill plans to add disclaimers to the SSW routines to make clear that they do not give valid results for pre-1972 dates.  For example, ``utc2tai.pro`` already has such a disclaimer.

sunpy/time/tests/test_taiseconds.py
```diff
@@ -0,0 +1,50 @@
...
+
+@pytest.mark.parametrize('time', [
+    Time('1958-01-01T00:00:00', format='isot', scale='tai'),
+    Time('1972-01-01T00:00:00', format='isot', scale='tai'),
+    Time('2015-10-25T05:24:08', format='isot', scale='tai'),
+    Time('2018-09-17T19:46:25', format='isot', scale='tai'),
+])
+def test_unix_tai_consistency(time):
+    diff_unix_tai = Time(TimeUnixTai.epoch_val, scale='tai', format=TimeUnixTai.epoch_format,).tai_seconds
+    assert time.unix_tai + diff_unix_tai == time.tai_seconds
```
ayshih, line 50:
```suggestion
```
This test appears to have negligible added value, because you're basically checking that TimeFromEpoch really means time from an epoch. Essentially the only conceivable way that it could fail is if someone accidentally changes the epoch_scale of TimeUnixTai from TAI to something else, and that would presumably cause test failures in Astropy.

ayshih (approved):

nabobalis (changes requested):
Should have a what new section added before merge.

sunpy/time/tests/test_taiseconds.py
wtbarnes: Fair enough. I've removed this.

sunpy/time/time.py
wtbarnes: Fixed

docs/whatsnew/3.1.rst
```diff
@@ -64,5 +64,19 @@ to use `~sunpy.map.GenericMap.date_average`,
 `~sunpy.map.GenericMap.date_start`, or `~sunpy.map.GenericMap.date_end`
 instead if you need one of these specific times.
 
+Addition of new time format `~sunpy.time.formats.TimeTaiSeconds`
+----------------------------------------------------------------
+The `~sunpy.time.formats.TimeTaiSeconds` format is the number of
```
dstansby, line 69:
```suggestion
The new `~sunpy.time.formats.TimeTaiSeconds` format is the number of
```

nabobalis (approved):

nabobalis:
Can we get a rebase to check the CI but the docs failed on:

```
/home/vsts/work/1/s/docs/whatsnew/3.1.rst:67: WARNING: py:obj reference target not found: sunpy.time.formats.TimeTaiSeconds
/home/vsts/work/1/s/docs/whatsnew/3.1.rst:69: WARNING: py:obj reference target not found: sunpy.time.formats.TimeTaiSeconds
```

docs/whatsnew/3.1.rst
```diff
@@ -64,5 +64,19 @@ to use `~sunpy.map.GenericMap.date_average`,
 `~sunpy.map.GenericMap.date_start`, or `~sunpy.map.GenericMap.date_end`
 instead if you need one of these specific times.
 
+Addition of new time format `~sunpy.time.formats.TimeTaiSeconds`
+----------------------------------------------------------------
```
dstansby, line 68:
```suggestion
Addition of new time format `~sunpy.time.TimeTaiSeconds`
--------------------------------------------------------
```

docs/whatsnew/3.1.rst
```diff
@@ -64,5 +64,19 @@ to use `~sunpy.map.GenericMap.date_average`,
...
+The new `~sunpy.time.formats.TimeTaiSeconds` format is the number of
```
dstansby, line 69:
```suggestion
The new `~sunpy.time.TimeTaiSeconds` format is the number of
```

dstansby (commented):
Should fix the doc build

PR #5490, opened 2021-08-09
Fixed a units issue when determining the proper value for ``lonpole``

ayshih:
There's a bug in #5448 where I didn't make sure that I was comparing degrees to degrees when determining the proper value for ``lonpole``.  This affects `Helioprojective` maps when the reference coordinate's latitude is greater than 90 arcseconds.

Ping @wtbarnes


wtbarnes (approved):

dstansby (commented):
Looks 👍  - is it worth adding a test with `CUNIT` set to degrees to check that the original issue doesn't crop up again in the future?

PR #5497, opened 2021-08-11
Fixed the missing docstrings for properties dynamically created by Astropy's `BaseCoordinateFrame`

ayshih:
While we wait for astropy/astropy#12005 to come out in a release, here's an analogous fix in our repo so that our doc builds pass.


ayshih:
This PR works as it should, but the doc build is still failing because of a different reason.

nabobalis (approved):

nabobalis:
Yeah that can be fixed in another PR.

PR #5501, opened 2021-08-12
No more modify meta

dstansby:
Replaces https://github.com/sunpy/sunpy/pull/5199 - see that PR for motivation and discussion.


wtbarnes:
There are also still some modification to header values in `io`, e.g. the fixing of the `WAVEUNIT` keyword: https://github.com/sunpy/sunpy/blob/a319336bb2deb7d495e52614970a4dd26a4b31fd/sunpy/io/fits.py#L143

I don't know if you want to address that here as well?

dstansby:
Thanks - I hadn't spotted that, but I will try to remove that in this PR too.

wtbarnes:
It came about when I was constructing the EUI headers (which are for some reason missing the `WAVEUNIT` key). When you pass a file into the map factory, it passes through `io` and gets the wave unit fix. But if you just pass the data header pair, this fix is not applied and things blow up when you try to access the wavelength property.

dstansby:
I had a quick ctrl-F and `WAVEUNIT` doesn't seem to be in the FITS spec anywhere, so I guess it's just a commonly used but not official keyword?

wtbarnes:
Oh there are going to be some conflicts between this and #5261 😅

PR #5503, opened 2021-08-15
Allow hdu_type argument of sunpy.io.fits.write to be a HDU instance

### LINKED ISSUE #4908: Allow passing keyword arguments to the HDU object used in GenericMap.save
Currently we allow specifying the HDU object to `GenericMap.save` (which ends up in `sunpy.io.fits.write`), but not any way of specifying arguments to this HDU, for example the `quantize_level=` kwarg to the `fits.CompImageHDU`.

The offending line is here: https://github.com/sunpy/sunpy/blob/0757ca6f8b/sunpy/io/fits.py#L185

The difficulty here is we are already using `**kwargs` as arguments to `fits.writeto`, so we either need a `hdu_kwargs=` argument (which doesn't work well with other writers) or we need to work out a way of allowing passing *instances* of a HDU object into `write`. Which would mean something like: `aia.save("test.fits", hdu_type=fits.CompImageHDU(quantize_level=32))` would work.

The only worry with that idea to me is what happens if the HDU instance already has data or a header set?

ConorMacBride:
### Description
The `sunpy.map.map_base.GenericMap.save` method allows a `hdu_type` argument to be passed to `sunpy.io.fits.write`. Currently it only supports a HDU *object*. This PR adds support for `hdu_type` being a HDU *instance*, so the user can control its initialisation kwargs. 

I'm not sure if updating the data after HDU initialisation is supported by Astropy? I've noticed an inconsistency with how headers are updated by the HDU's data setter method. For example updating the data in an `ImageHDU` updates NAXIS etc. immediately, however updating the data in a `CompImageHDU` seems to only update the header before writing to file. (I think I'll open an issue in astropy for this.)

Fixes #4908


sunpy/io/fits.py
```diff
@@ -183,7 +185,20 @@ def write(fname, data, header, hdu_type=None, **kwargs):
     if not hdu_type:
         hdu_type = fits.PrimaryHDU
 
-    hdu = hdu_type(data=data, header=fits_header)
+    if isinstance(hdu_type, type):
+        hdu = hdu_type(data=data, header=fits_header)
+    else:  # HDU already initialised
```
Cadair, line 190: I think this is better done the other way around.

```python
if isinstance(hdu_type, (fits.PrimaryHDU, fits.ExtensionHDU)):
```
should catch all the cases of an already instantiated HDU, and that leaves the alternative a little more flexible.

sunpy/map/mapbase.py
```diff
@@ -1384,9 +1384,11 @@ def save(self, filepath, filetype='auto', **kwargs):
...
         filetype : str
             'auto' or any supported file extension.
-        hdu_type: None, `~astropy.io.fits.CompImageHDU`
-            `None` will return a normal FITS file.
-            `~astropy.io.fits.CompImageHDU` will rice compress the FITS file.
+        hdu_type : type or `~astropy.io.fits.CompImageHDU`, optional
+            By default, a normal FITS file is written.
+            `~astropy.io.fits.CompImageHDU` will Rice compress the FITS file.
+            If a type is given, a new HDU of this type will be created.
+            If a HDU instance is given, its data and header will be updated.
```
Cadair, line 1391:
```suggestion
        hdu_type : `~astropy.io.fits.ExtensionHDU` instance or class, optional
            By default, a FITS file is written with the map in it's primary HDU.
            If a type is given, a new HDU of this type will be created.
            If a HDU instance is given, its data and header will be updated from the map.
            Then that HDU instance will be written to the file.
```

sunpy/io/fits.py
```diff
@@ -161,9 +161,11 @@ def write(fname, data, header, hdu_type=None, **kwargs):
...
     header : `dict`
         A header dictionary.
-    hdu_type: `None` or `~astropy.io.fits.CompImageHDU`
-        `None` will return a normal FITS files.
-        `~astropy.io.fits.CompImageHDU` will rice compress the FITS file.
+    hdu_type : type or `~astropy.io.fits.CompImageHDU`, optional
+        By default, a normal FITS file is written.
+        `~astropy.io.fits.CompImageHDU` will Rice compress the FITS file.
+        If a type is given, a new HDU of this type will be created.
+        If a HDU instance is given, its data and header will be updated.
```
Cadair, line 168: Same comment as the map docstring

Cadair (changes requested):
This looks great, just a couple of little nits and then we can merge it.

sunpy/io/fits.py
```diff
@@ -183,7 +185,20 @@ def write(fname, data, header, hdu_type=None, **kwargs):
     if not hdu_type:
         hdu_type = fits.PrimaryHDU
 
-    hdu = hdu_type(data=data, header=fits_header)
+    if isinstance(hdu_type, (fits.PrimaryHDU, fits.hdu.base.ExtensionHDU)):
+        hdu = hdu_type  # HDU already initialised
+
+        # Merge `header` into HDU's header
+        # Values in `header` take priority, including cards such as
+        # 'SIMPLE' and 'BITPIX'.
```
Cadair, line 193: I think cards like SIMPLE and BITPIX will be "corrected" to match the file requirements at write time, which is fine.

Cadair (approved):

sunpy/io/fits.py
```diff
@@ -161,9 +161,11 @@ def write(fname, data, header, hdu_type=None, **kwargs):
         n-dimensional data array.
     header : `dict`
         A header dictionary.
-    hdu_type: `None` or `~astropy.io.fits.CompImageHDU`
-        `None` will return a normal FITS files.
-        `~astropy.io.fits.CompImageHDU` will rice compress the FITS file.
+    hdu_type : `~astropy.io.fits.hdu.base.ExtensionHDU` instance or class, optional
+        By default, a FITS file is written with the map in its primary HDU.
+        If a type is given, a new HDU of this type will be created.
+        If a HDU instance is given, its data and header will be updated from the map.
```
ayshih, line 167: This sentence suggests that providing an HDU instance with data is acceptable and will simply result in its data being replaced, rather than raising an error as is currently coded.  Is it problematic to accept an HDU instance with data?  Regardless the documentation needs to match the behavior.

ConorMacBride: It shouldn't be problematic to accept a HDU instance with data. I just thought it would be unlikely that someone would want to do that and if they did, they probably didn't mean to. Thinking this through more, it may be best to just allow for existing data to be present (i.e. implement as currently documented). What do you think?

ayshih: My inclination is to support passing in HDU instances that already have data, although maybe there ought to a be a `deepcopy()` in that case.

wtbarnes (commented):
Everything looks good to me, but could use a rebase to fix the figure tests.

sunpy/io/fits.py
ConorMacBride: I've now removed the requirement for no data. Actually, it seems that some data *needs* to be set during initialisation of `CompImageHDU` for the compression settings to be remembered. I opened an issue at https://github.com/astropy/astropy/issues/12216.

Also, I tried wrapping the HDU instance in `copy.deepcopy(...)`, however that causes the tests to fail due to differences in the HDU's header. I think this is an upstream issue and may be related to https://github.com/astropy/astropy/pull/12058 so it might be best if we don't create a copy of the HDU. I think the documentation is clear enough that the HDU instance will simply be updated without copying.

PR #5516, opened 2021-08-27
Enable user to override slider labels in ArrayAnimatorWCS.

DanRyanIrish:
### Description

Currently, is a user tries to set the slider labels for an `ArrayAnimatorWCS` instance, an error is raised saying multiple kwarg values for `slider_labels` received.  This PR resolves this bug by giving priority to user inputs.

Fixes #<Issue Number>


dstansby (changes requested):
Looks good, but
- Needs a test to check that the slider labels are used when supplied by the user
- That "axis_ranges" is ignored in kwargs needs documenting in the docstring

PR #5521, opened 2021-09-03
Only pass line property kwargs to a contour when plotting CompositeMap

### LINKED ISSUE #5455: No way to specify line colors for a CompositeMap plot that includes contours
If I want to specify the line colours of a contour map using `CompositeMap`, I think the correct code is:

```python

import astropy.units as u

import sunpy.data.sample
import sunpy.map

aia_map = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
hmi_map = sunpy.map.Map(sunpy.data.sample.HMI_LOS_IMAGE)
comp_map = sunpy.map.Map(aia_map, hmi_map, composite=True)

comp_map.set_levels(index=1, levels=[-1000, -500, -250, 250, 500, 1000] * u.G)

fig = plt.figure()
ax = plt.subplot(projection=aia_map)
comp_map.plot(colors='black')
plt.show()
```

This fails however, as the keyword arguments given to `comp_map.plot()` are handed to *both* calls to `plot()` and `draw_contours()`, and an error is raised when `.plot(colors='black')` is passed (see below). The fix here is to work out a way to hand both `plot` and `draw_contours` keyword arguments separately to `CompositeMap.plot()`.

```python
  File "/Users/dstansby/github/sunpy/test.py", line 19, in <module>
    comp_map.plot(linewidths=1, colors='black')
  File "/Users/dstansby/github/sunpy/sunpy/map/compositemap.py", line 419, in plot
    ret.append(m.plot(**params))
  File "/Users/dstansby/github/astropy/astropy/units/decorators.py", line 253, in wrapper
    return_ = wrapped_function(*func_args, **func_kwargs)
  File "/Users/dstansby/github/sunpy/sunpy/map/mapbase.py", line 2467, in plot
    ret = axes.pcolormesh(data, **imshow_args)
  File "/Users/dstansby/miniconda3/envs/dev/lib/python3.9/site-packages/matplotlib/__init__.py", line 1361, in inner
    return func(ax, *map(sanitize_sequence, args), **kwargs)
  File "/Users/dstansby/miniconda3/envs/dev/lib/python3.9/site-packages/matplotlib/axes/_axes.py", line 6192, in pcolormesh
    collection = mcoll.QuadMesh(Nx - 1, Ny - 1, coords,
  File "/Users/dstansby/miniconda3/envs/dev/lib/python3.9/site-packages/matplotlib/collections.py", line 1996, in __init__
    super().__init__(**kwargs)
  File "/Users/dstansby/miniconda3/envs/dev/lib/python3.9/site-packages/matplotlib/_api/deprecation.py", line 431, in wrapper
    return func(*inner_args, **inner_kwargs)
  File "/Users/dstansby/miniconda3/envs/dev/lib/python3.9/site-packages/matplotlib/collections.py", line 217, in __init__
    self.update(kwargs)
  File "/Users/dstansby/miniconda3/envs/dev/lib/python3.9/site-packages/matplotlib/artist.py", line 1062, in update
    raise AttributeError(f"{type(self).__name__!r} object "
AttributeError: 'QuadMesh' object has no property 'colors'
```

ConorMacBride:
### Description

This PR filters the matplotlib kwargs passed to a `CompositeMap.plot()` such that kwargs related to line style and line colour are only used when plotting contours. (Line width kwargs were already filtered.) Incorrect contour kwargs such as `linestyle`, `ls`, `color` and `c` are also filtered such that matplotlib raises a warning instead of an error.

Fixes #5455

### DIFF
--- a/sunpy/map/compositemap.py
+++ b/sunpy/map/compositemap.py
@@ -371,8 +371,11 @@ class CompositeMap:
 
         Notes
         -----
-        If a line-width Matplotlib argument (``linewidth``, ``linewidths``, or ``lw``)
-        is provided, it will apply only to those maps that are plotted as contours.
+        Matplotlib arguments for setting the line width (``linewidths``), line style
+        (``linestyles``) or line color (``colors``) will apply only to those maps
+        that are plotted as contours. Similar arguments (i.e., ``linewidth``,
+        ``lw``, ``color``, ``c``, ``linestyle`` and ``ls``) will also be ignored
+        for maps plotted as images.
 
         If a transformation is required to overlay the maps with the correct
         alignment, the plot limits may need to be manually set because
@@ -404,8 +407,11 @@ class CompositeMap:
             # The request to show a map layer rendered as a contour is indicated by a
             # non False levels property.
             if m.levels is False:
-                # Check if any linewidth argument is provided, if so, then delete it from params.
-                for item in ['linewidth', 'linewidths', 'lw']:
+                # Check if any linewidth, color, or linestyle arguments are provided,
+                # if so, then delete them from params.
+                for item in ['linewidth', 'linewidths', 'lw',
+                             'color', 'colors', 'c',
+                             'linestyle', 'linestyles', 'ls']:
                     if item in matplot_args:
                         del params[item]
 
--- a/sunpy/map/tests/test_compositemap.py
+++ b/sunpy/map/tests/test_compositemap.py
@@ -51,6 +51,27 @@ def test_plot_composite_map_linewidths(composite_test_map):
     composite_test_map.plot(linewidths=0.5)
 
 
+@figure_test
+def test_plot_composite_map_linestyles(composite_test_map):
+    composite_test_map.set_levels(1, np.arange(-75, 76, 25) << u.percent)
+    composite_test_map.plot(linestyles='--')
+
+
+@figure_test
+def test_plot_composite_map_colors(composite_test_map):
+    composite_test_map.set_levels(1, np.arange(-75, 76, 25) << u.percent)
+    composite_test_map.plot(colors='red')
+
+
+def test_plot_composite_map_mplkwargs(composite_test_map):
+    composite_test_map.set_levels(1, np.arange(-75, 76, 25) << u.percent)
+    for k, v in [('linewidth', 0.5), ('lw', 0.5), ('color', 'red'),
+                   ('c', 'red'), ('linestyle', '--'), ('ls', '--')]:
+        with pytest.warns(UserWarning, match=f"The following kwargs were not "
+                                             f"used by contour: '{k}'"):
+            composite_test_map.plot(**{k: v})
+
+
 def test_remove_composite_map(composite_test_map):
     composite_test_map.remove_map(0)
     with pytest.raises(IndexError):
--- a/sunpy/tests/figure_hashes_mpl_332_ft_261_astropy_42.json
+++ b/sunpy/tests/figure_hashes_mpl_332_ft_261_astropy_42.json
@@ -2,6 +2,8 @@
   "sunpy.map.tests.test_compositemap.test_plot_composite_map": "8160a5896607f62b7e7fa316236f565f54bd7a08f511808e98c9ad329678d21a",
   "sunpy.map.tests.test_compositemap.test_plot_composite_map_contours": "909ac12088dde3c15d0c7908f08b4b3c0da23e2a4966ed1ef8e705adb8727d2a",
   "sunpy.map.tests.test_compositemap.test_plot_composite_map_linewidths": "27aa1ac78d4e49f731bf1e43f4d2b398fe8a84d53ed9234794c7954fda54c45a",
+  "sunpy.map.tests.test_compositemap.test_plot_composite_map_linestyles": "aa77398390f40d90f791a3fd45ff8d92b7fe98cc8332d9828acc31e7eddc5e88",
+  "sunpy.map.tests.test_compositemap.test_plot_composite_map_colors": "ca7af0e1ee21df8b76bac76f27ac934c452f7906d4bc9000f1faa1dde742aa62",
   "sunpy.map.tests.test_compositemap.test_set_alpha_composite_map": "b9d729ff67b0307c037ac62f97c69bb80d22ad6679a0e56352075845d4ef732b",
   "sunpy.map.tests.test_compositemap.test_peek_composite_map": "bd1df27446d420f3e8cb6fa07daff88b77dd0f0441e4acf83ee48341a9e3218f",
   "sunpy.map.tests.test_mapsequence.test_norm_animator": "a400eef26fdd915ccf12769031bd8dadedbf8f6a93e31cd5982bffe2484702ee",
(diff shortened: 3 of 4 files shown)


dstansby:
There should be a cleverer way to do this, using the `inspect` module in the standard library. In particular, [`inspect.signature`](https://docs.python.org/3/library/inspect.html#introspecting-callables-with-the-signature-object) can be used to get the acceptable keyword arguments for a function or method. You should be able to use this to get the keyword arguments accepted by `GenericMap.plot` and `plt.imshow`, and then delete any keyword arguments not in that list.

I think we should also raise a warning that lists the keys being deleted.

ayshih:
> [`inspect.signature`](https://docs.python.org/3/library/inspect.html#introspecting-callables-with-the-signature-object) can be used to get the acceptable keyword arguments for a function or method.

You can't, because many of the keyword arguments aren't explicit in the signatures, but rather largely lumped in `**kwargs` (possibly to be passed on to lower-level functions).

dstansby:
Ah, I only went two steps down the `**kwargs` rabbit hole 😄 . It seems like it would be possible but for the chain ending at "`Artist` properties" here: https://matplotlib.org/stable/api/image_api.html#matplotlib.image.AxesImage, and I'm not sure how one could query for them.

ConorMacBride:
I think the following code works although it may be making too many assumptions about the inner workings of Matplotlib. It uses `inspect` to get valid keywords from the signature of various functions and methods. As you say, not all valid kwargs appear in a signature, however you can query the presence of callable `set_...` methods, which is how it's done internally (see https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/artist.py#L1058-L1061).

As you can see from the output of the code, we would have to do more filtering. For example `linestyles` and `linewidths` are valid for both `pcolormesh` and `contour`.

```py
import inspect

from matplotlib.image import AxesImage, _ImageBase
from matplotlib.contour import ContourSet, QuadContourSet
from matplotlib.collections import QuadMesh, Collection


def get_keywords(func) -> set:
    """Returns a set of keyword names from `func`'s docstring.
    Recursive if `func` is a list of functions and methods."""
    if isinstance(func, list):
        keywords = set()
        for f in func:
            keywords.update(get_keywords(f))
        return keywords
    sig = inspect.signature(func)
    keywords = {param.name for param in sig.parameters.values()
                if param.default is not inspect.Parameter.empty}
    return keywords


def get_set_methods(obj) -> set:
    """Returns a set of keyword names that can be handled by 
    an object's `set_...` methods."""
    # See https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/artist.py#L1058-L1061
    return {
        m[4:] for m in dir(obj)
        if m.startswith('set_') and callable(getattr(obj, m, None))
    }


# Valid keyword arguments for each plotting method
imshow_kwargs = get_keywords(
    [plt.Axes.imshow, AxesImage.__init__, _ImageBase.__init__]
) | get_set_methods(AxesImage)

pcolormesh_kwargs = get_keywords(
    [plt.Axes.pcolormesh, QuadMesh.__init__, Collection.__init__]
) | get_set_methods(QuadMesh)

contour_kwargs = get_keywords(
    [ContourSet.__init__, QuadContourSet._process_args]
)

# Analysis

print("Valid image kwargs:\n\t", imshow_kwargs | pcolormesh_kwargs, '\n')

print("-----\n")
    
print("Valid image kwargs that...\n")
print("...would raise error for pcolormesh:\n\t", imshow_kwargs - pcolormesh_kwargs, '\n')
print("...would raise error for imshow:\n\t", pcolormesh_kwargs - imshow_kwargs, '\n')
print("...would raise error for contour:\n\t", (imshow_kwargs & pcolormesh_kwargs) - contour_kwargs, '\n')

print("-----\n")

print("Valid contour kwargs:\n\t", contour_kwargs, '\n')

print("-----\n")
    
print("Valid contour kwargs that...\n")
print("...would raise error for pcolormesh:\n\t", contour_kwargs - pcolormesh_kwargs, '\n')
print("...would raise error for imshow:\n\t", contour_kwargs - imshow_kwargs, '\n')

print("-----\n")

print("Valid kwargs that are valid for (imshow AND/OR pcolormesh) AND contour:")
print("\t", (imshow_kwargs | pcolormesh_kwargs) & contour_kwargs)
```

ConorMacBride:
Here's the output of the code above:
```
Valid image kwargs:
	 {'clip_box', 'zorder', 'aspect', 'resample', 'picker', 'clip_on', 'urls', 'interpolation', 'offsets', 'figure', 'ls', 'transOffset', 'gid', 'capstyle', 'clim', 'clip_path', 'visible', 'antialiaseds', 'url', 'label', 'norm', 'transform', 'filternorm', 'aa', 'shading', 'filterrad', 'cmap', 'agg_filter', 'linewidth', 'fc', 'vmax', 'data', 'antialiased', 'dashes', 'linewidths', 'vmin', 'linestyle', 'path_effects', 'array', 'alpha', 'facecolor', 'ec', 'origin', 'animated', 'snap', 'linestyles', 'offset_position', 'in_layout', 'pickradius', 'sketch_params', 'color', 'edgecolor', 'rasterized', 'facecolors', 'hatch', 'joinstyle', 'extent', 'contains', 'lw', 'paths', 'edgecolors'} 

-----

Valid image kwargs that...

...would raise error for pcolormesh:
	 {'filterrad', 'interpolation', 'extent', 'filternorm', 'aspect', 'resample', 'origin'} 

...would raise error for imshow:
	 {'linewidth', 'fc', 'dashes', 'antialiased', 'linewidths', 'urls', 'offsets', 'linestyle', 'facecolor', 'ls', 'ec', 'transOffset', 'capstyle', 'linestyles', 'offset_position', 'paths', 'pickradius', 'color', 'edgecolor', 'antialiaseds', 'facecolors', 'hatch', 'joinstyle', 'lw', 'aa', 'shading', 'edgecolors'} 

...would raise error for contour:
	 {'clip_box', 'zorder', 'agg_filter', 'data', 'picker', 'clip_on', 'figure', 'array', 'path_effects', 'gid', 'animated', 'snap', 'clim', 'clip_path', 'in_layout', 'visible', 'sketch_params', 'rasterized', 'url', 'label', 'contains'} 

-----

Valid contour kwargs:
	 {'cmap', 'levels', 'linestyles', 'filled', 'corner_mask', 'vmax', 'nchunk', 'antialiased', 'locator', 'extend', 'linewidths', 'hatches', 'vmin', 'colors', 'norm', 'extent', 'alpha', 'transform', 'origin'} 

-----

Valid contour kwargs that...

...would raise error for pcolormesh:
	 {'origin', 'hatches', 'levels', 'colors', 'filled', 'extent', 'corner_mask', 'extend', 'nchunk', 'locator'} 

...would raise error for imshow:
	 {'locator', 'linewidths', 'hatches', 'levels', 'linestyles', 'colors', 'filled', 'corner_mask', 'extend', 'nchunk', 'antialiased'} 

-----

Valid kwargs that are valid for (imshow AND/OR pcolormesh) AND contour:
	 {'origin', 'linewidths', 'cmap', 'vmin', 'linestyles', 'norm', 'extent', 'alpha', 'transform', 'vmax', 'antialiased'}
```

dstansby (changes requested):
For the sake of getting a fix in, how about we merge (and backport) this PR as is, and then we can think about taking the more complicated approach. Could you update the hashes for the devdeps figure test? You can find the correct ones in the build output here: https://app.circleci.com/pipelines/github/sunpy/sunpy/14270/workflows/d3b7dbf3-c086-4a16-baad-e14adbf7ec92/jobs/70694

ConorMacBride:
@dstansby Great! I've updated the figure hashes and added a changelog entry. When this is merged I can open a draft PR showing how it could be implemented.

dstansby (approved):

nabobalis (approved):

nabobalis:
Thanks for the PR @ConorMacBride

PR #5526, opened 2021-09-16
[Backport 3.0] Only pass line property kwargs to a contour when plotting CompositeMap

ConorMacBride:
Backport #5521 

(Not sure if I'm supposed to do manual backports so feel free to ignore if you have your own process!)


ConorMacBride:
It needed to be manual because the order of some existing figure hashes were different.

Cadair:
Thanks for doing the backport, that's really helpful :smile:

nabobalis:
Thanks again @ConorMacBride

PR #5527, opened 2021-09-17
Added a comparison of differential-rotation models

ayshih:
- Added a comparison of differential-rotation models as a gallery example
- Added a reference to it in the `diff_rot()` docstring using `minigallery`

### DIFF
--- a/examples/differential_rotation/comparing_rotation_models.py
+++ b/examples/differential_rotation/comparing_rotation_models.py
@@ -54,12 +54,12 @@ for model in ['howard', 'snodgrass', 'allen', 'rigid']:
                                                        rotation_model=model))
 
 ##############################################################################
-# Plot all four differential-rotated meridians over the original plot.  To
-# plot each meridian, we use :meth:`~sunpy.map.GenericMap.draw_quadrangle` to
-# follow a line of constant longitude in the original frame.  (If one were to
-# instead use :meth:`astropy.visualization.wcsaxes.WCSAxes.plot_coord`,
-# ``meridian`` would need to include points at intermediate latitudes between
-# the two poles to render as desired.)
+# Finally, we plot the differentially rotated meridians over the map, using
+# :meth:`~sunpy.map.GenericMap.draw_quadrangle` to conveniently draw a line
+# of constant longitude in the original frame between two endpoints.  (One
+# could instead use :meth:`astropy.visualization.wcsaxes.WCSAxes.plot_coord`,
+# but then ``meridian`` would need to consist of a sequence of many points
+# spanning all latitudes between the two poles to render as desired.)
 # Note that the "rigid" model appears as the meridian again as expected for a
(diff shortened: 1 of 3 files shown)


ayshih:
sphx_glr_comparing_rotation_models_001

examples/differential_rotation/comparing_rotation_models.py
```diff
@@ -0,0 +1,76 @@
...
+
+import sunpy.map
+from sunpy.coordinates import HeliographicStonyhurst, RotatedSunFrame
+from sunpy.data.sample import AIA_335_IMAGE
+from sunpy.sun.constants import sidereal_rotation_rate
+
+##############################################################################
+# First, we use an AIA observation primarily as a pretty background.  We also
+# define a coordinate rectangle with zero width in the longitude direction to
+# represent the meridian.
```
dstansby, line 26:
```suggestion
# First, we use an AIA observation primarily as a pretty background.  We also
# define a line at zero degrees longitude to represent the meridian.
```
Seems simpler? Perhaps there's a reason for using rectangle instead?

dstansby: Ah, now I've made it to the end of the example I see why. I think it might be clearer to use an array of latitudes, and `plot_coord` at the end? Using `draw_quadrangle` is neat in this context, but I don't think very intuitive.

dstansby (approved):
Nice example! I've left one non-blocking comment/suggestion, but otherwise looks great.

examples/differential_rotation/comparing_rotation_models.py
ayshih: We already have an example showing the use of `plot_coord()` and `RotatedSunFrame`, so I doubly like using `draw_quadrangle()` here instead.  I've revised the text, so take a look at the new description.

examples/differential_rotation/comparing_rotation_models.py
```diff
@@ -0,0 +1,81 @@
...
+# model.
+
+rotated_meridian = {}
+for model in ['howard', 'snodgrass', 'allen', 'rigid']:
+    rotated_meridian[model] = SkyCoord(RotatedSunFrame(base=meridian,
+                                                       duration=sidereal_period,
+                                                       rotation_model=model))
+
+##############################################################################
+# Plot all four differential-rotated meridians over the original plot.  To
```
wtbarnes, line 57: I think this sentence could be removed. It reads a bit awkward and the next one pretty much says the same.

wtbarnes (approved):
Just a minor comment but not blocking.

examples/differential_rotation/comparing_rotation_models.py
ayshih: Okay, I've polished the text a bit.

nabobalis (approved):

PR #5529, opened 2021-09-18
Use newer data for stereo examples

dstansby:
Should fix the doc build

### DIFF
--- a/examples/units_and_coordinates/AIA_limb_STEREO.py
+++ b/examples/units_and_coordinates/AIA_limb_STEREO.py
@@ -17,18 +17,17 @@ import sunpy.map
 from sunpy.net import Fido
 from sunpy.net import attrs as a
 
-##############################################################################
-# The first step is to download some data, we are going to get an image from
-# early 2011 when the STEREO spacecraft were roughly 90 deg separated from the
-# Earth.
+###############################################################################
+# The first step is to download some data. We download an image from STEREO-A
+# and an image from SDO, which are separated in longitude.
 
-stereo = (a.Source('STEREO_B') &
+stereo = (a.Source('STEREO_A') &
           a.Instrument("EUVI") &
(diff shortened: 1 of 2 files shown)


examples/units_and_coordinates/AIA_limb_STEREO.py
```diff
@@ -22,13 +22,13 @@
 # early 2011 when the STEREO spacecraft were roughly 90 deg separated from the
```
nabobalis, line 22: Will need to update the example text as well to match.

nabobalis (approved):

dstansby:
https://sunpy--5529.org.readthedocs.build/en/5529/generated/gallery/index.html#coordinates-times-and-units if anyone wants to check the examples are still fine - I've taken a look and they seem to be to me.

ayshih (changes requested):
Stuff that ought to be changed in this PR:

- Plotting this new AIA image would greatly benefit from explicitly setting `vmin`.

Stuff that would be nice to also get in, but can be a separate PR:

- `AIA_limb_STEREO.py`
  - The ~dozen lines near at the end to annotate the HPC grid as seen by SDO on the STEREO image don't do anything because that HPC grid doesn't actually intersect with the edges of the plot area, so no ticks, tick labels, or axis labels are shown.  One way to "fix" this is to zoom in so that the plot is no longer full-disk.
- `SDO_to_STEREO_Coordinate_Conversion.py`
  - This example needs some re-thinking.  It's really about submapping the same region on two maps from different observers, and the explicit coordinate conversions are all superfluous (now).
  - I – of course – heartily recommend using `draw_quadrangle()` for the plotting of boxes:
Figure_4
Figure_6

nabobalis:
> Oh rats, merged while I was typing this. Okay, I'll start a new PR

I opened https://github.com/sunpy/sunpy/issues/5533 to keep track of your suggestions.

PR #5530, opened 2021-09-20
Dynamic Matplotlib argument lists for filtering plotting setting in `CompositeMap.plot`

ConorMacBride:
### Description
This PR builds upon the fix implemented in #5521 by dynamically generating a list of accepted kwargs for each possible Matplotlib plotting method and then filtering the kwargs passed by `CompositeMap.plot` to each plotting method, instead of implementing exceptions for specific arguments.

### DIFF
--- a/sunpy/map/compositemap.py
+++ b/sunpy/map/compositemap.py
@@ -388,11 +388,15 @@ class CompositeMap:
 
         Notes
         -----
-        Matplotlib arguments for setting the line width (``linewidths``), line style
-        (``linestyles``) or line color (``colors``) will apply only to those maps
-        that are plotted as contours. Similar arguments (i.e., ``linewidth``,
-        ``lw``, ``color``, ``c``, ``linestyle`` and ``ls``) will also be ignored
-        for maps plotted as images.
+        Images are plotted using either `~matplotlib.axes.Axes.imshow` or
+        `~matplotlib.axes.Axes.pcolormesh`, and contours are plotted using
+        `~matplotlib.axes.Axes.contour`.
+        The Matplotlib arguments accepted by the plotting method are passed to it.
+        If any Matplotlib arguments are not used by any plotting method,
+        a ``TypeError`` will be raised.
+        For the full list of arguments accepted by each plotting method see
+        ``ACCEPTED_IMSHOW_KWARGS``, ``ACCEPTED_PCOLORMESH_KWARGS``
+        and ``ACCEPTED_CONTOUR_KWARGS``.
 
         If a transformation is required to overlay the maps with the correct
         alignment, the plot limits may need to be manually set because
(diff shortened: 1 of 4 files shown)


sunpy/map/compositemap.py
```diff
@@ -3,18 +3,35 @@
...
+from sunpy.util import expand_list, get_keywords, get_set_methods
 from sunpy.visualization import axis_labels_from_ctype, peek_show, wcsaxes_compat
 
 __all__ = ['CompositeMap']
 
 __author__ = "Keith Hughitt"
 __email__ = "keith.hughitt@nasa.gov"
 
+# Valid keyword arguments for each plotting method
+VALID_IMSHOW_KWARGS = ACCEPTED_IMSHOW_KWARGS = get_keywords(
```
nabobalis, line 22: Do we need both valid and accepted?

ConorMacBride: No, only the `ACCEPTED_*_KWARGS` are used. Just added both for verbosity, as for example `VALID_PCOLORMESH_KWARGS` and `ACCEPTED_PCOLORMESH_KWARGS` are different. Maybe the `VALID_*_KWARGS` variables shouldn't be created as they're neither used nor documented.

nabobalis: If we can get away with only one of them, I would prefer that.

sunpy/util/util.py
```diff
@@ -215,3 +217,57 @@ def get_width():
...
+    sig = inspect.signature(func)
+    keywords = {param.name for param in sig.parameters.values()
+                if param.default is not inspect.Parameter.empty}
+    return keywords
+
+
+def get_set_methods(obj):
+    """
+    Returns a set of keyword names that can be handled by
+    an object's `set_...` methods.
```
nabobalis, line 252:
```suggestion
    an object's ``set_...`` methods.
```

sunpy/util/util.py
```diff
@@ -215,3 +217,57 @@ def get_width():
     else:
         width, _ = get_terminal_size()
     return width
+
+
+def get_keywords(func):
+    """
+    Returns a set of keyword names from `func`'s signature.
+    Recursive if `func` is a list of functions and methods.
```
nabobalis, line 225:
```suggestion
    Recursive if ``func`` is a list of functions and methods.
```

sunpy/util/util.py
```diff
@@ -215,3 +217,57 @@ def get_width():
     else:
         width, _ = get_terminal_size()
     return width
+
+
+def get_keywords(func):
+    """
+    Returns a set of keyword names from `func`'s signature.
```
nabobalis, line 224:
```suggestion
    Returns a set of keyword names from ``func``'s signature.
```

PR #5532, opened 2021-09-20
SunPy > sunpy

dstansby:
This is a (manual) find-and-replace to replace lots of instances of `SunPy` with `sunpy`, and where appropriate clarify with either 'sunpy core package' or 'SunPy project.


nabobalis (approved):

nabobalis:
Example fails only. I will merge this. 

Do we want to backport this? I am worried it will need a manual backport.

dstansby:
Seems like a hassle to backport, and I don't think it's important enough to be worth it.

PR #5414, opened 2021-06-20
Add draw_limb function

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/5253, and includes an example to demonstrate the new functionality.

PR #5542, opened 2021-09-27
Move peek_show and axis_labels_from_ctype to utils

dstansby:
In view of `sunpy.visualization` hopefully getting some useful user-facing functionality in 3.1 (https://github.com/sunpy/sunpy/pull/5414), I thought it would be worth deprecating `peek_show` and `axis_labels_from_ctype`, moving them into `sunpy.util.visualization` so we can reserve the right to modify or break their API without having to provide support to users.


ayshih:
Huh.  I didn't know – or had completely forgotten – that `sunpy.util` is exempt from API stability.

In the case of these two functions, why not just make them private within `sunpy.visualization`?  Why keep them exposed at all?

dstansby:
> In the case of these two functions, why not just make them private within `sunpy.visualization`?

No particular reason, `sunpy.util.vizualisation` seemed nicer to me than (e.g.) `sunpy.vizualisation.util`, as then there's a nice divide between private helpers (`sunpy.util`) and user-facing stuff. Very happy to go the other way though.

> Why keep them exposed at all?

What do you mean by exposed?

dstansby:
https://docs.sunpy.org/en/stable/code_ref/stability.html says of `sunpy.util` 

> This sub-module is not user facing and can and will change between versions.

 so I always assumed it was a bit of a dumping ground for stuff like this, we didn't bother underscoring anything in `sunpy.util`.

Maybe worth chatting on the weekly call about what to do with `sunpy.util`? Having a scan down the docs it currently seems to be a mix of

- User facing stuff worth documenting (e.g. `MetaDict`, warnings, exceptions)
- Stuff that should probably disappear from the docs (like the stuff in this PR, e.g., `dict_keys_same`, `get_width`, `unique`...)

PR #5551, opened 2021-09-29
Document how to write private code and util/utils.py

dstansby:
I think this covers what we agreed on the call today? Opinions/suggestions for changes and additions still very welcome.

### DIFF
--- a/docs/dev_guide/contents/code_standards.rst
+++ b/docs/dev_guide/contents/code_standards.rst
@@ -51,11 +51,26 @@ Coding Style/Conventions
 
 Private code
 ============
+
 It is often useful to designate code as private, which means it is not part of the user facing API, only used internally by sunpy, and can be modified without a deprecation period.
-Any classes, functions, or variables that are private should live in:
+Any classes, functions, or variables that are private should either:
+
+- Have an underscore as the first character of their name, e.g., ``_my_private_function``.
+- If you want to do that to entire set of functions in a file, name the file with a underscore as the first character, e.g., ``_my_private_file.py``.
+
+If these might be useful for other packages within the sunpy ecosphere, they should be made public.
+
+Utilities in sunpy
+==================
+
+Within ``sunpy``, it might be useful to have a set of utility classes or functions that are used by internally to help with certain tasks or to provide a certain level of abstraction.
+These should be placed either:
 
 - ``sunpy.{subpackage}.utils.py``, if it is only used within that sub-package.
-- ``sunpy.utils`` if it is used across multiple sub-packages.
+- ``sunpy.util`` if it is used across multiple sub-packages.
+
+These can be private (see section above) or public.
+The decision is up to the developer, but if these might be useful for other packages within the sunpy ecosphere, they should be made public.
 
 Formatting
 ==========
--- a/docs/dev_guide/contents/code_standards.rst
+++ b/docs/dev_guide/contents/code_standards.rst
@@ -49,6 +49,14 @@ Coding Style/Conventions
 
 * General utilities necessary for but not specific to the package should be placed in the ``sunpy.utils`` module.
 
+Private code
+============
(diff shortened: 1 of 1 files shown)


docs/dev_guide/contents/code_standards.rst
```diff
@@ -49,6 +49,13 @@ Coding Style/Conventions
 
 * General utilities necessary for but not specific to the package should be placed in the ``sunpy.utils`` module.
```
ayshih, line 50:
```suggestion
* General utilities necessary for but not specific to the package should be placed in the ``sunpy.util`` module.
```

docs/dev_guide/contents/code_standards.rst
```diff
@@ -49,6 +49,13 @@ Coding Style/Conventions
...
 
+Private code
+============
+Any classes, functions, or variables that are private (ie. not part of the user facing API) should always be pre-pended with an underscore and should live in:
+
+- a sub-package, if it is only used within that sub-package.
+- ``sunpy.utils`` if it is used across multiple sub-packages.
```
ayshih, line 57:
```suggestion
- ``sunpy.util`` if it is used across multiple sub-packages.
```

ayshih:
Hmm.  Would such guidance mean that the transformation functions in `sunpy.coordinates.transformations`, none of which are directly called by users, should be prepended with an underscore?  Would such guidance mean that we need to start prepending underscores to everything in `sunpy.util`?

dstansby:
Yes, it would. My own thinking behind this is:
- Underscored stuff doesn't show up in tab auto-completion (e.g. in a jupyter notebook)
- If you're using `dir()` to see what's available in a module, an underscore makes it clear that certain functions/classes/methods are not intended for user use

Cadair:
I am not sure I am in favour of a policy which is this blunt. I see the appeal, but at the same time it feels redundant to not have a way to signal "this whole file isn't public API".

One potential solution could be to rename `sunpy/coordinates/transformations.py` to `sunpy/coordinates/_transformations.py` but that too feels weird.

Cadair:
Yeah I was mostly talking about the underscore thing. I think it's probably worth sorting out some way of indicating that certain sub-subpackages are not subject to stability promises though.

dstansby:
How about now? I've removed underscoring. Because this means it's unclear what denotes private/public API in sub-packages, I've suggested normalising this to living in `sunpy.{subpackage}.utils.py`.

nabobalis:
Do we want to talk about the future of utils and if we want to actually make that develop facing and therefore we will maintain its API like we do for user code?

dstansby:
I've lost track of what we decided (if anything). If someone else wants to take my branch and run feel free.

nabobalis:
> I've lost track of what we decided (if anything). If someone else wants to take my branch and run feel free.

I've pushed some changes. Lets see how it goes.

docs/dev_guide/contents/code_standards.rst
```diff
@@ -49,6 +49,29 @@ Coding Style/Conventions
 
 * General utilities necessary for but not specific to the package should be placed in the ``sunpy.utils`` module.
```
nabobalis:
```suggestion
```

docs/dev_guide/contents/code_standards.rst
```diff
@@ -47,7 +47,30 @@ Coding Style/Conventions
 
 * ``__init__.py`` files for modules should not contain any significant implementation code. ``__init__.py`` can contain docstrings and code for organizing the module layout.
 
-* General utilities necessary for but not specific to the package should be placed in the ``sunpy.utils`` module.
+* General utilities necessary for but not specific to the package should be placed in the ``sunpy.util`` module.
```
nabobalis, line 50:
```suggestion
```

Cadair (approved):

Issue #5543, opened 2021-09-27
Rotating a map results in a different `date`


dstansby:
If you take a map without a date defined, sunpy will assume the current date/time as the date for the map. If you then rotate the map, a new updated date is assigned to the rotated map. This should be changed so the assumed date for the initial map is propagated to the new map.

```python
import numpy as np
from sunpy.map import Map

meta = {'crval1': 0,
    'crval2': 0,
    'cdelt1': 2.0,
    'cdelt2': 2.0,
    'ctype1': 'HPLN-TAN',
    'ctype2': 'HPLT-TAN',
    'naxis': 2,
    'naxis1': 129,
    'naxis2': 129,
    'cunit1': 'arcsec',
    'cunit2': 'arcsec',
    'crpix1': 64.5,
    'crpix2': 64.5,
    'pc1_1': -1.0,
    'pc1_2': 0.0,
    'pc2_1': 0.0,
    'pc2_2': 1.0,
#     'date-obs': '2020-01-01 00:00:00'
}
m = Map(np.ones((129, 129)), meta)
mr = m.rotate()

print(m.date)
print(mr.date)
```

```python
2021-09-27 15:56:11.658103
2021-09-27 15:56:12.343774
```

ayshih:
The same consideration applies to other methods that derive a new map (e.g., `.submap()`).

> This should be changed so the assumed date for the initial map is propagated to the new map.

I'm actually slightly against this because there are appropriate warnings emitted about the observation time needing to be assumed.  If a user has a map without an observation time in the header, it's arguably correct for any derived map to also lack the observation time in the header.  It's not clear to me that baking in *our* assumption is the correct approach.

Of course, if this behavior is not changed, it should be documented as a known issue.

dstansby:
This is also probably redundant if/when we do https://github.com/sunpy/sunpy/issues/5454

samaloney:
Yea I think this was a case of user error but it would be nice if a more useful error could be raised rather than the current silent fail.

ayshih:
> rather than the current silent fail

To be a bit snarky, it currently isn't a "silent" fail:
```python
>>> mr = m.rotate()
sunpy\sunpy\map\mapbase.py:809: SunpyMetadataWarning: Missing metadata for observation time, setting observation time to current time. Set the 'DATE-AVG' FITS keyword to prevent this warning.
  warn_metadata("Missing metadata for observation time, "
[...]

PR #5553, opened 2021-09-30
Fixed a bug when rotating a Map missing observation-time metadata

ayshih:
Fixed a bug reported by @samaloney where rotating a Map that is missing observation-time metadata can result in an incorrect reference coordinate, potentially even NaNs if the reference coordinate is off-disk in a helioprojective frame.

When observation-time metadata is missing, the Map will assume "now" for the observation time, where "now" is the current time when the `.date` property is accessed.  (At least for now; see #5454.)

The bug occurs because `.rotate()` internally creates a temporary map using the (incomplete) metadata to aid in a WCS calculation, which triggers another assumption for the observation time, but the "now" for this temporary map will invariably be different from the "now" of the original map.  This triggers an unintended coordinate transformation of the reference coordinate across a change in observation time.

This PR fixes the bug by using a WCS instance for the calculation rather than bothering with internally creating an map.

Aside: This PR does not add any observation-time metadata to the rotated map.  See #5543 for discussion on whether that should be done.


sunpy/map/mapbase.py
```diff
@@ -1573,12 +1573,12 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=4, scale=1.0,
 
         pixel_array_center = (np.flipud(new_data.shape) - 1) / 2.0
 
-        # Create a temporary map so we can use it for the data to pixel calculation.
-        temp_map = self._new_instance(new_data, new_meta, self.plot_settings)
+        # Create a temporary WCS so we can use it for the data to pixel calculation.
+        new_wcs = self.wcs.deepcopy()
+        new_wcs.wcs.crpix += [pad_x, pad_y]
```
dstansby, line 1578:
```suggestion
        new_wcs = WCS(new_meta)
```
Since the padding is already added to `new_meta` above, this avoids having to add it again to the WCS?

ayshih: Nope, that doesn't work because creating the `WCS` instance that way doesn't inject our logic that isn't added to the metadata (e.g., assumptions about observation time and observer location).  Regardless, it turns out this WCS calculation isn't needed.

ayshih:
Upon further review, the WCS calculation is vestigial and unnecessary.  We're asking the `WCS` for the pixel location of the reference pixel, but that should simply return the CRPIX1/2 values that we just set.  I can't think of any pathological situation where that wouldn't be true.

It looks like the figure tests are failing because removing the WCS calculation removes numerical-precision inaccuracy.

sunpy/map/mapbase.py
```diff
@@ -1566,19 +1566,12 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=4, scale=1.0,
...
 
         pixel_array_center = (np.flipud(new_data.shape) - 1) / 2.0
 
-        # Create a temporary map so we can use it for the data to pixel calculation.
-        temp_map = self._new_instance(new_data, new_meta, self.plot_settings)
-
-        # Convert the axis of rotation from data coordinates to pixel coordinates
-        pixel_rotation_center = u.Quantity(temp_map.world_to_pixel(self.reference_coordinate)).value
-        del temp_map
+        pixel_rotation_center = u.Quantity(self.reference_pixel).value + [pad_x, pad_y]
```
dstansby, line 1574:
```suggestion
        pixel_rotation_center = u.Quantity(self.reference_pixel).to_value(u.pix) + [pad_x, pad_y]
```

dstansby (approved):
Just one minor suggestion - I'm always very wary of doing a bare `.value` on a quantity without specifying the desired units.

sunpy/map/mapbase.py
ayshih: Conceptually, sure, but this is not arbitrary input; we control the property `.reference_pixel`, and the units will always be pixels.  As a similar situation, I am fine with a bare `.value` on `.rsun_meters` because the units are a promise.

dstansby:
Looks like some of the test images changed, which suggests this (ie. correctly taking into account the padding) might be a bigger fix than just fixing the missing time issue?

ayshih:
Yup, as I said above, doing the detour through the WCS calculation produces numerical-precision issues.  For example, in a case I just run, CRPIX1/2 that should be a clean (700.5, 700.5), but ends up being (700.5000000000026, 700.4999999999955).  I need to update the hashes and add another bugfix changelog entry.

ayshih:
Ready for review again

nabobalis (approved):

nabobalis:
Figure hashes conflict with main.

ayshih:
> Figure hashes conflict with main.

Should be fixed now

nabobalis:
Doc build failure is in main (woops)

PR #5556, opened 2021-09-30
Adding .satellite_number as a XRSTimeSeries property.

### LINKED ISSUE #5547: Adding GOES SatelliteNumber to XRSTimeSeries when reading cdf files with TimeSeries
There is a dedicated field for satellite number in the goes metadata at XRSTimeSeries.meta.metas[0]['platform'] but this, like many of these fields is blank for all but the most modern data. Currently you get the GOES satellite number by parsing a number of othjer fields or the filenames themselves. 

Since SatelliteNumber is a generally useful field, should this be something that is parsed by TimeSeries when loading the data and is accessible directly from the XRSTimeSeries rather than having users construct their own way of harvesting the information.

alasdairwilson:
### Description
Adds  XRSTimeSeries.satellite_number, populated by scraping the GOES meta data, as a useful reference. The GOES metadata is extremely inconsistent in formatting and completeness so no guarantee the cases are exhaustive but testing it seems to cover all the data. 

Fixes #5547

### DIFF
--- a/sunpy/timeseries/sources/goes.py
+++ b/sunpy/timeseries/sources/goes.py
@@ -5,6 +5,7 @@ import datetime
 from pathlib import Path
 from collections import OrderedDict
 from distutils.version import LooseVersion
+from sunpy.net.dataretriever.attrs.goes import SatelliteNumber
 
 import h5netcdf
 import matplotlib.dates
@@ -18,6 +19,7 @@ from astropy.time import Time, TimeDelta
 
 import sunpy.io
 from sunpy import log
+from sunpy.extern import parse
 from sunpy.io.file_tools import UnrecognizedFileTypeError
 from sunpy.time import is_time_in_given_format, parse_time
 from sunpy.timeseries.timeseriesbase import GenericTimeSeries
@@ -115,6 +117,37 @@ class XRSTimeSeries(GenericTimeSeries):
         axes.fmt_xdata = matplotlib.dates.DateFormatter("%H:%M")
         return axes
 
+    @property
+    def satellite_number(self):
+        """
+        Parses the various metafields to extract GOES satellite number.
+        """
+        #various pattern matches for the meta fields.
+        pattern_old = ("{}go{SatelliteNumber:02d}{}{month:2d}{day:2d}.fits{}")
+        pattern_new = ("{}sci_gxrs-l2-irrad_g{SatelliteNumber:02d}_d{year:4d}{month:2d}{day:2d}_{}.nc{}")
+        pattern_r = ("{}sci_xrsf-l2-flx1s_g{SatelliteNumber:02d}_d{year:4d}{month:2d}{day:2d}_{}.nc{}")
+        pattern_telescop=("GOES {SatelliteNumber:02d}")
+        pattern_inst=("{}GOES 1-{SatelliteNumber:02d} {}")
+
+        try: 
+            id = self.meta.metas[0]['id']
+            parsed=parse(pattern_r,str(id))
+            if parsed==None:
+                parsed=parse(pattern_new,str(id))
+                if parsed==None:
+                    parsed=parse(pattern_old,str(id))
+                    if parsed==None:
+                        id=self.meta.metas[0]['Instrument']
+                        parsed=parse(pattern_inst,str(id))
+        except KeyError:
+            try:
+                id=self.meta.metas[0]['TELESCOP']
+                parsed=parse(pattern_telescop,str(id))
+            except KeyError:
+                print("Error: Satellite Number not found in metadata")
+                return -1                   
+        return parsed['SatelliteNumber']
+       
     @peek_show
     def peek(self, title="GOES Xray Flux", **kwargs):
         """


sunpy/timeseries/sources/goes.py
```diff
@@ -115,6 +117,37 @@ def plot(self, axes=None, **kwargs):
...
+            id = self.meta.metas[0]['id']
+            parsed=parse(pattern_r,str(id))
+            if parsed==None:
+                parsed=parse(pattern_new,str(id))
+                if parsed==None:
+                    parsed=parse(pattern_old,str(id))
+                    if parsed==None:
+                        id=self.meta.metas[0]['Instrument']
+                        parsed=parse(pattern_inst,str(id))
+        except KeyError:
```
nabobalis, line 142: What line of code raises the `KeyError`?

nabobalis:
Thanks for the PR @alasdairwilson.

Could you add some unit tests for this inside `https://github.com/sunpy/sunpy/blob/main/sunpy/timeseries/tests/test_timeseries_factory.py` just to check the string parsing?

Also a changelog entry (following https://github.com/sunpy/sunpy/blob/main/changelog/README.rst). This would be a feature entry.

sunpy/timeseries/sources/goes.py
alasdairwilson: 133/140 for the first and 144 for second, basically anywhere there is an id =

sunpy/timeseries/sources/goes.py
```diff
@@ -115,6 +116,37 @@ def plot(self, axes=None, **kwargs):
...
+                    parsed = parse(pattern_old, str(id))
+                    if parsed is None:
+                        id = self.meta.metas[0]['Instrument']
+                        parsed = parse(pattern_inst, str(id))
+        except KeyError:
+            try:
+                id = self.meta.metas[0]['TELESCOP']
+                parsed = parse(pattern_telescop, str(id))
+            except KeyError:
+                print("Error: Satellite Number not found in metadata")
```
dstansby, line 146:
```suggestion
                log.debug('Satellite Number not found in metadata')
```
Instead of printing we have a custom log system, that allows users to customise what they see.

sunpy/timeseries/sources/goes.py
```diff
@@ -115,6 +116,37 @@ def plot(self, axes=None, **kwargs):
...
+                return -1
```
dstansby, line 147:
```suggestion
                return
```
I think it's better to return `None` (which is what a bare `return` does) rather than -1.

sunpy/timeseries/tests/test_timeseries_factory.py
```diff
@@ -150,6 +150,21 @@ def test_implicit_new_goes16(self):
...
+
+    def test_implicit_new_goes15_satno(self):
+        # Test a GOES TimeSeries for satellite number
+        ts_goes = sunpy.timeseries.TimeSeries(new_goes15_filepath)
+        assert ts_goes.satellite_number == 15
+
+    def test_implicit_new_goes16_satno(self):
+        # Test a GOES TimeSeries for satellite number
+        ts_goes = sunpy.timeseries.TimeSeries(new_goes16_filepath)
+        assert ts_goes.satellite_number == 17
```
dstansby, line 166: Is it expected that the satellite number is 17, even though (according to pre-exising code in sunpy!) this is a GOES 16 file?

alasdairwilson: Yeah that is actually a goes 17 file despite the name, I noticed this when designing the test, was deciding whether to raise a separate issue but it really doesn't effect anything other than this new test.

`>>> goes=ts.TimeSeries(new_goes16_filepath)
>>> goes.satellite_number
17
>>> goes.meta.metas[0]['platform']
b'g17'
>>> goes.meta.metas[0]['id']
b'sci_xrsf-l2-flx1s_g17_d20201016_v2-0-1.nc'`

dstansby:
This looks great now 👍, just code style things need fixing - see 
 https://dev.azure.com/sunpy/sunpy/_build/results?buildId=20065&view=logs&jobId=673f7616-04ea-52e2-247c-12c8f27bc080&j=673f7616-04ea-52e2-247c-12c8f27bc080&t=4088a903-e138-5bca-3a5d-42ba61a9adf0 for info. https://docs.sunpy.org/en/latest/dev_guide/contents/code_standards.html#formatting has instructions on how to run the checks locally, and fix them.

sunpy/timeseries/tests/test_timeseries_factory.py
```diff
@@ -150,6 +150,21 @@ def test_implicit_new_goes16(self):
...
+        assert ts_goes.satellite_number == 15
+
+    def test_implicit_new_goes15_satno(self):
+        # Test a GOES TimeSeries for satellite number
+        ts_goes = sunpy.timeseries.TimeSeries(new_goes15_filepath)
+        assert ts_goes.satellite_number == 15
+
+    def test_implicit_new_goes16_satno(self):
+        # Test a GOES TimeSeries for satellite number
+        ts_goes = sunpy.timeseries.TimeSeries(new_goes16_filepath)
```
nabobalis, line 165: Since this is actually a goes 17 file, I wonder if we should rename the variables while we are here?

Cadair (changes requested):
I have a high level concern here.

In Map we currently have a policy that the API of all Map sources is defined by the `GenericMap` class, and source specific subclasses only change the implementation of that API not add new API.

If we carry that policy over to TimeSeries this change would violate that principle. I think this needs discussing.

Cadair:
Discussion at the community meeting today seemed to converge on instead of `satellite_number` make a property on `GenericTimeSeries` called `observatory` and then populate it on goes with something like `GOES <number>`.

PR #5557, opened 2021-10-01
Improvements to timeseries plotting

dstansby:
- Allow a subset of columns to be plotted
- If all the units of the columns being plotted are the same, add a y-label to the plot


sunpy/timeseries/timeseriesbase.py
```diff
@@ -470,19 +472,30 @@ def plot(self, axes=None, **plot_args):
...
+            columns = self._data.columns
+
+        axes = self._data[columns].plot(ax=axes, **plot_args)
 
+        units = set([self.units[col] for col in columns])
+        if len(units) == 1:
+            # If units of all columns being plotted are the same, add a unit
+            # label to the y-axis.
+            unit = u.Unit(list(units)[0])
+            axes.set_ylabel(unit.to_string())
```
hayesla, line 485: could this not just be:
```suggestion
            axes.set_ylabel(units[0].to_string())
```

hayesla (approved):

sunpy/timeseries/timeseriesbase.py
dstansby: It's not possible to index a `set` (which I used to get unique values), so I had to put it in a list first. There might be another way of making this nicer though that I missed...

PR #5435, opened 2021-07-02
Add CDF reader to TimeSeries

dstansby:
Fixes https://github.com/heliopython/heliopy/issues/949

PR #5558, opened 2021-10-01
Add CDAWeb client

dstansby:
This adds data searching/access for CDAWeb, which is NASA's main repository for in-situ space physics datasets (including Solar Orbiter, Parker Solar Probe, and many others). With https://github.com/sunpy/sunpy/pull/5435 this makes it super easy to download/load/visualise in-situ data.

Currently depends on https://github.com/sunpy/sunpy/pull/5557


examples/acquiring_data/search_cdaweb.py
```diff
@@ -0,0 +1,44 @@
...
+
+How to download data from the Coordinated Data Analysis Web (CDAWeb).
+
+CDAWeb stores data from from current and past space physics missions, and is
+full of heliospheric insitu datasets.
+"""
+
+from sunpy.net import Fido
+from sunpy.net import attrs as a
+from sunpy.net.cdaweb.attrs import Dataset
```
Cadair, line 14: You shouldn't need to do this, you can have the client register the attrs namespace so this becomes `a.cdaweb.Dataset`, 
like this:

https://github.com/sunpy/sunpy/blob/71b09358db2a08fa2aec34a9798c1cc7ed6368ac/sunpy/net/jsoc/jsoc.py#L806-L808

though why this isn't in the Fido client dev docs is a mystery to me.

dstansby: Thanks - FWIW I copy and pasted across code I wrote ~1 year ago and adapted it for sunpy, so haven't been following up to date Fido client docs. I'll take a look through the docs though and see if there's anything else I should update/change.

Cadair:
wanna rebase to see if the tests pass?

dstansby:
It is currently rebased 🙈

Issue #5528, opened 2021-09-18
Simple AIA Fido request failing


dstansby:
```python
import astropy.units as u

from sunpy.net import Fido
from sunpy.net import attrs as a

aia = (a.Instrument.aia &
       a.Sample(24 * u.hour) &
       a.Time('2011-01-01', '2011-01-02'))

wave = a.Wavelength(301 * u.Angstrom, 310 * u.Angstrom)
result = Fido.search(wave & aia)
```
returns no results, and raises the warning:

```
WARNING: SunpyUserWarning: VSO-C500 :soap:Server : ERROR:  relation "vso.aia_synoptic2" does not exist
LINE 12: FROM vso.aia_synoptic2
              ^
 [sunpy.net.vso.vso]
```

This is what's causing our doc build to fail at the moment. Running a query for AIA data on the same date on the VSO website also fails, so I suspect this is a VSO issue?

wtbarnes:
Per the chat today, this is apparently because of some hardware maintenance at the VSO last week. https://matrix.to/#/!MeRdFpEonLoCwhoHeT:matrix.org/$1632069913112438Mwcdv:matrix.org?via=matrix.org&via=openastronomy.org&via=cadair.com

ejm4567:
Actually, the issue is there is no AIA NRT data prior to 5/11/12. Hence the error message.  JSOC gets pulled in because it matches the time and instrument parts of the query.

We've added this use case to our ticket for improving error messages in the VSO, so you'll get a more informative error message in the future.

nabobalis:
Thanks @ejm4567.

wtbarnes:
@ejm4567 The above `Fido` query should not be asking for NRT data. Is there some reason that it might be hitting the NRT data?

Cadair:
here's the full debug output:
```
DEBUG: VSO Request:
 <soap-env:Envelope xmlns:soap-env="http://schemas.xmlsoap.org/soap/envelope/" xmlns:VSO="http://virtualsolar.org/VSO/VSOi">
  <soap-env:Body>
    <VSO:Query>
      <body>
        <block>
          <time>
            <start>20110101000000</start>
            <end>20110102000000</end>
          </time>
          <instrument>AIA</instrument>
          <wave>
            <wavemin>301.0</wavemin>
            <wavemax>310.0</wavemax>
            <waveunit>Angstrom</waveunit>
          </wave>
          <sample>86400.0</sample>
        </block>
      </body>
[...]

PR #5575, opened 2021-10-11
Fix remote tests

dstansby:
Due to https://github.com/sunpy/sunpy/issues/5528 these are failing. Since we're near a release and we should have our CI working, change some tests to a query that works with the VSO at the moment.


Cadair:
I have been told the VSO are still working on fixing this, but given we wont need to change this back I see no reason not to merge this. :+1:

PR #5583, opened 2021-10-11
verbose figure test builds to avoid timeout

nabobalis:
Might need more flags?


Cadair:
What was taking that long? compiling mpl?

nabobalis:
Don't know, there was no output to know.

PR #5584, opened 2021-10-13
Fix an asdf schema bug when saving a HPC or HCC frame with a HGS observer

Cadair:


Cadair:
Ah yes, good shout, fixed and test added.

PR #5594, opened 2021-10-14
Fix version number on RTD

dstansby:
### Description

Fixes #<Issue Number>


dstansby:
@Cadair should I fix or close this?

dstansby:
I think this is fixed? I just tagged `3.1.0` locally, checked it was a dirty checkout, and `conf.py` was reporting a version of '3.1.0`.

PR #5597, opened 2021-10-14
Testing OpenAstronomy/azure-pipelines-templates PR

ConorMacBride:
#### ⚠️ Do not merge!
This PR is to test the upgraded Codecov uploader method in https://github.com/OpenAstronomy/azure-pipelines-templates/pull/68 before merging it upstream.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -20,6 +20,7 @@ resources:
     endpoint: sunpy
     name: ConorMacBride/azure-pipelines-templates
     ref: upgrade-codecov
+# (trigger ci)
 
 trigger:
   branches:


ConorMacBride:
Coverage seems to have dropped mostly in areas where plotting is needed, and possibly some tests that rely on remote data. Maybe the reports uploaded by circleci did not get included? The bash uploader is still being used for circleci and that was able to upload successfully, but maybe codecov don't allow mixing reports from different upload methods?

ConorMacBride:
Yeah, there are definitely Azure jobs missing, however they are all apparently uploaded successfully in the Azure logs. Codecov seem to be having a lot of issues recently with uploading reports. These seem to be getting fixed, so I pushed the update for circleci to trigger a new run to see if it's fixed.

ConorMacBride:
> circle didn't upload: https://app.circleci.com/pipelines/github/sunpy/sunpy/14680/workflows/db1f1816-9b48-4703-8a23-83da0332f078/jobs/71589?invite=true#step-108-33

Uploads are now working with the path to the report given to the codecov program.

ConorMacBride:
I've notice another potential issue with the template. I'll fix that, then I will rebase and push to do a final check here.

ConorMacBride:
I ran the CI again without changing anything meaningful and it has worked. I think the issue before must have been internal to codecov. Even though their upload URL returned a positive response, all the reports did not appear on their website, and the reports that did did not cause the codecov check to appear in Github.

Issue #5560, opened 2021-10-04
read_file stopped working for URLs (Windows)


Laubeee:
### Description

`read_file` used to work perfectly for reading from network, but since 3.0.1 it doesn't work anymore (downgrading to 3.0.0 solved the problem). See also [my comment on the respective commit](https://github.com/sunpy/sunpy/commit/93b21b68b2ec3dcf08bdeb4e7110359bd377a79e#commitcomment-57477187) including a possible solution.
It is probably a windows-only problem due to pathlib replacing / with \ (which then results in an invalid URL)

### Steps to Reproduce

```python
from sunpy.io import read_file
read_file("https://hesperia.gsfc.nasa.gov/rhessi_extras/imagecube_fits/2015/12/20/20151220_2228_2248/hsi_imagecube_clean_20151220_2228_13tx3e.fits")
```

### System Details

- SunPy Version: 3.0.1
- Astropy Version: 4.3.1
- Python Version: 3.9
- OS information: Win10

PR #5601, opened 2021-10-18
Fix read_file for URLs

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/5560


Cadair:
As you say, this should be tested. What is this stack calling out to to download the file?

nabobalis:
> I don't have much desire to look into mocking a test at the moment - if someone else wants to take a stab at it feel free to push to my branch.

I had a go.

PR #5608, opened 2021-10-19
Register CDAWeb Dataset attr values

dstansby:
Add a helper function to download all (>2000!) of the CDAWeb Dataset attributes, and register them with the attr. I figured this was worth a new PR. Depends on https://github.com/sunpy/sunpy/pull/5558.


nabobalis:
Can we get the json file to be one line instead like the others?

Issue #5548, opened 2021-09-29
Add `GenericMap` method for normalizing to exposure time


wtbarnes:
A common operation with imaging instruments is to normalize them to the exposure time (e.g. converting from DN to DN/s). `GenericMap` could provide a method for dividing by the exposure time (which we already expose as property `.exposure_time`), adjusting the metadata appropriately (dividing `'BUNIT'` by `'s'`, setting the exposure time to 1), and emitting a new map. 

There is already an implementation of this in `aiapy` ( https://gitlab.com/LMSAL_HUB/aia_hub/aiapy/-/blob/master/aiapy/calibrate/prep.py#L204), but there's nothing AIA-specific about this and could easily be used with any map that holds an exposure time.

Bracciata:
Tackling this.

PR #5614, opened 2021-10-20
Overload arithmetic operators on `GenericMap`

wtbarnes:
This is a first attempt at providing basic, unit-aware arithmetic operations on `Map`. This means you can do things like background subtraction or exposure time normalization (e.g. #5548) through the usual arithmetic operations without having to manually rebuild the map each time.

These operations are explicitly limited to array-like quantities. Operations between maps are not allowed.

See also sunpy/ndcube#74

For example, the following is now possible with this PR,

```python
>>> import sunpy.map
>>> import sunpy.data.sample
>>> m = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
>>> m + 10*m.unit  # addition and subtraction only work with appropriate units
>>> m - 100*m.unit
>>> 10 * m
>>> m * 5
>>> m / m.exposure_time  # note the resulting units on the new map are now ct / s
>>> 1 / m
>>> -m
>>> m**2
```

Once this is merged, #5611 and #5548 can be closed as they will no longer be relevant.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -178,12 +178,13 @@ class GenericMap(NDData):
     _registry = dict()
     # This overrides the default doc for the meta attribute
     meta = MetaData(doc=_meta_doc, copy=False)
-    # This is set so that the Quantity non-reflected operators will return
-    # NotImplemented instead of an error, which allows the GenericMap reflected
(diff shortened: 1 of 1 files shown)


wtbarnes:
Should also support `__pow__`

PR #5621, opened 2021-10-22
copy astropy wcs setter

nabobalis:
Fixes devdeps.


nabobalis:
> Erm, while this fixes the build error, this approach is bizarre. This setter code lets someone assign something to `.wcs`, but the value that is set is completely ignored by our overridden `.wcs` property (which generates the WCS instance on the fly). In fact, your setter code is not really any different from simply being `pass`. I don't think we want to let someone set `.wcs` and have it apparently do nothing.
> 
> I don't know a suggestion for how to fix the build error, but I don't think we want to do it this way.

Change it to only work if its not set and raise an error otherwise?

Issue #5630, opened 2021-10-26
Fail early if there are non-example gallery doc errors on CI


dstansby:
Currently azure runs the doc build in one go. Building the example gallery takes up a lot of time as part of the build, and often fails due to remote server issues. It would be good to change this to:

- Run the doc build **without** the example gallery, and fail early if e.g. any links are broken
- Then build the gallery in a second stage

This has the advantages of:
- Easily seeing if the doc build failed because of the example gallery or something else
- Reduce build/CI time if a non-gallery part of the doc build fails

dstansby:
Can't remember where we changed this, but we did.

PR #5634, opened 2021-10-26
Allowed fail mode for releases

Cadair:
Also fast fail for docs (#5630)


dstansby:
FYI I've backported 2afceea7f63d75928aafb77875998fa4c20de533 to 3.0 in https://github.com/sunpy/sunpy/pull/5693

PR #5655, opened 2021-10-29
Make `rsun` a keyword argument to get_observer_meta

dstansby:
Given we accept `rsun=None` as input anyway, I think it makes sense to make this an optional keyword argument.

### DIFF
--- a/sunpy/map/header_helper.py
+++ b/sunpy/map/header_helper.py
@@ -229,7 +229,7 @@ def _get_wcs_meta(coordinate, projection_code):
 
 
 @u.quantity_input
-def get_observer_meta(observer, rsun: (u.Mm, None)):
+def get_observer_meta(observer, rsun: (u.Mm, None) = None):
     """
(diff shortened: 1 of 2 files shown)


sunpy/map/header_helper.py
```diff
@@ -238,17 +238,14 @@ def get_observer_meta(observer, rsun: (u.Mm, None)):
     coordinate : `~astropy.coordinates.BaseCoordinateFrame`
         The coordinate of the observer, must be transformable to Heliographic
         Stonyhurst.
-    rsun : `astropy.units.Quantity`
+    rsun : `astropy.units.Quantity`, optional
```
wtbarnes, line 241: Could you add a note here about the consequences of leaving this as `None`? e.g. "if `rsun` is None, the ``rsun_ref`` and ``rsun_obs`` keys will not be set.

dstansby: 👍  done

nabobalis (approved):

PR #5663, opened 2021-11-02
Remove old workaround in QueryResponseTable

dstansby:
This can go now we depend on astropy >= 4.2. Probably not worth a backport or a changelog entry.


sunpy/net/base_client.py
```diff
@@ -172,15 +172,6 @@ class QueryResponseTable(QTable):
     display_keys = TableAttribute(default=slice(None))
     hide_keys = TableAttribute()
 
-    # This is a work around for https://github.com/astropy/astropy/pull/11217
-    # TODO Remove when min astropy version is > 4.2.1
```
Cadair, line 176: I think we need to change the version in setup.cfg to be 4.2.1 rather than 4.2.0.

Cadair:
We could back port this to 3.1?

nabobalis:
I think its fine to not backport.

nabobalis (approved):

PR #5668, opened 2021-11-02
Cache test data in Azure Pipelines

### LINKED ISSUE #4831: Cache test data during remote data CI
I've just noticed that (among other files) the tests download a 77.5MB Lyra file. Even with Azure downloading it a 3MB/s this eats up 30s of test time every time the remote tests are run. It would be good if there was a way to cache the remote data used during the remote data tests.

ConorMacBride:
### Description
This configures caching for data downloaded/generated during Azure Pipelines jobs.

Fixes #4831

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -109,6 +109,11 @@ stages:
             pytest: false
 
           - linux: py38-online
+            cache_dirs:
+            - key: 'sample_data'
+              path: .tox/py38-online/sample_data
+            - key: 'working_dir'
+              path: .tox/py38-online/tmp/sunpy
 
 
   - ${{ if or(eq(variables['Build.Reason'], 'Schedule'), eq(variables['Build.Reason'], 'Manual')) }}:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -107,6 +107,15 @@ stages:
           - linux: build_docs_ci
             posargs: " "
             pytest: false
+            cache_dirs:
+            - key: 'sample_data'
+              path: .tox/build_docs_ci/sample_data
+            - key: 'working_dir'
+              path: .tox/build_docs_ci/tmp/sunpy
+            - key: 'build'
+              path: docs/_build
+            - key: 'generated'
+              path: docs/generated
 
           - linux: py38-online
             cache_dirs:


ConorMacBride:
Just trying one job for now to see how it works. I'm thinking you probably won't want or need caching on cron jobs such as the `py38-hypothesis` (which also uses remote data) as nobody is waiting for those to complete?

The `Installing openjpeg with brew` task takes about 30 seconds to complete. Brew [supports caching](https://docs.brew.sh/External-Commands#other-executable-scripts) so if you think this would be worthwhile I could add an option to cache brew upstream?

I'll also look into caching the `build_docs_ci` job here. In addition to the `sample_data` and `working_dir` caches, I need to work out which directories under `docs` should be cached.

Once a cache is created it will only be cleared after 7 days of not being used, and existing caches cannot be updated with new files that appear in subsequent jobs. This could potentially cause issues. Additional identifiers can be added to the `key` such that the cache can be replaced more regularly. For example, we could add a pull request identifier to the `key`, such that the PR number would act as a "version number" for the cache.

azure-pipelines.yml
```diff
@@ -109,6 +109,11 @@ stages:
             pytest: false
 
           - linux: py38-online
+            cache_dirs:
+            - key: 'sample_data'
+              path: .tox/py38-online/sample_data
```
Cadair, line 114: The sample data shouldn't be needed for anything other than docs (I hope).

azure-pipelines.yml
```diff
@@ -109,6 +109,11 @@ stages:
...
+            - key: 'working_dir'
+              path: .tox/py38-online/tmp/sunpy
```
Cadair, line 116: I am not sure we want to cache this for the test builds, it might mean the tests skip downloads when they should test them etc.

Cadair:
Thanks for this! :partying_face: 

Will these caches get restored over all jobs? all jobs per-PR or what?

azure-pipelines.yml
Cadair: This is of course the point of #4831, I would expect that at least a few of our tests actually want to test transferring data, and not just parfive's skip-if-exist code :wink:

Cadair:
https://docs.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops#cache-isolation-and-security

Is interesting, so PRs can access the `main` branch caches but not write to them.

azure-pipelines.yml
ConorMacBride: My `sunpy/.tox/py39-online/sample_data` has 65M of data so does that not mean it's spending time downloading the sample data? (And therefore can be cached.) If it's not needed, should the `online` tox environment be configured to not download sample data?

Cadair: Interesting.

azure-pipelines.yml
ConorMacBride: Yeah, that makes sense. If there are specific tests that need to test downloading (and not just *use* the data) maybe there is a way to specify a different download directory for those tests, tell it to overwrite exiting, or simply delete certain files from the cached directory at the beginning of the test.

ConorMacBride:
I think that means that on the first pipeline run of a PR, caches would be restored from the `main` branch's scope, and then at the end, the cache would be saved to the PR branch's scope. I assume that extra files (not present in the `main` branch caches) that appear during the first PR pipeline run will then be cached in the PR branch's scope and accessible in future runs of the pipeline in that PR. (This is assuming that cache keys only have to be unique within the same scope.)

azure-pipelines.yml
Cadair: Ah I was incorrect, our narrative documentation uses sample data, which is tested in the online build, so yep, this would be very useful here and in the docs build (they should share the same cache id, and very rarely need changing).

Cadair:
> I'll also look into caching the build_docs_ci job here. In addition to the sample_data and working_dir caches, I need to work out which directories under docs should be cached.

This probably isn't super trivial. I would think `_build` and `generated` but also probably some others.

azure-pipelines.yml
Cadair: Yeah, either setting `overwrite=True` or `path=tmpdir` to `Fido.fetch` would achieve those goals, but we would need to audit all the tests using Fido.fetch or other things making use of the default download dir to make sure we aren't footgunning ourselves.

dstansby: Shall we just put a cache on the doc build for now pending resolution of this?

dstansby (changes requested):

Cadair:
If you don't mind @ConorMacBride I am going to push some changes to this for the doc build stuff I have in mind, it's more complex to explain than do :D

azure-pipelines.yml
Cadair: We can still cache the sample data, probably just not the working dir for now.

ConorMacBride:
> If you don't mind @ConorMacBride I am going to push some changes to this for the doc build stuff I have in mind, it's more complex to explain than do :D

@Cadair Sure, no problem!

dstansby:
Given https://github.com/sunpy/sunpy/issues/5913 I'm going to close this.

PR #5669, opened 2021-11-03
Move automodapi files into generated

Cadair:
This is an idea I had while looking at #5668 

It moves the generated files made by automodapi into the generated directory which is also used by sphinx-gallery and sphinx itself (for the modules tree).

**The major issue with this, is that it changes the URLs. Which means that some links that use our `/latest/` path will break.**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -211,7 +211,6 @@ sunpy/cython_version.py
 sunpy/_version.py
 docs/_build
 docs/generated
-docs/api/
 docs/whatsnew/latest_changelog.txt
 examples/**/*.asdf
 examples/**/*.csv
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -113,6 +113,9 @@ extensions = [
     "sphinxext.opengraph",
 ]
 
+# Set automodapi to generate files inside the generated directory
+automodapi_toctreedirnm = "generated/api"
+
 # Add any paths that contain templates here, relative to this directory.
 # templates_path = ['_templates']
 


Cadair:
I think we could maybe use [read the docs redirects](https://docs.readthedocs.io/en/stable/user-defined-redirects.html#page-redirects) to cover up the URL issues. i.e. redirect `/api/` to `/generated/api/` at the point we merge this for `latest` and then when we release 4.0 do the same for stable. I would need to check that would actually work before we try it though.

ConorMacBride:
I think it would be good to have all the "generated" files in the same directory to make caching and cleaning between builds simpler. However, having `/generated/` in the URL seems unnecessary. I wonder if it's possible to have `docs/generated/gallery` and `docs/generated/api` but as `https://docs.sunpy.org/en/stable/gallery/` and `https://docs.sunpy.org/en/stable/api/` in the URL. (And do the redirect from the gallery instead.)

Cadair:
I don't think pulling it from the URL path is likely to be possible, at least not without some deep sphinx trickery.

I can't seem to get the redirects working as promised in the RTD docs, I shall reach out to them for help.

Cadair:
I have now figured this out, I setup the redirect the wrong way around for testing, so `/generated/api/` now redirects to `/api/` on latest.

nabobalis:
What will happen to the 3.1 and 3.0 branches or the 4.0 branches in future?

Cadair:
Well URLs with specific versions in wont break, the info will just be in a different place in 4.0 vs 3.1, the issue is when you have a URL like latest or stable and it is different one moment to the next.

nabobalis (approved):

Cadair:
I have setup the redirects for latest and stable.

PR #5672, opened 2021-11-04
Pin glymur to fix windows tests

dstansby:
Should fix the windows tests. Glymur 0.9.5 seems to have introduced a new dependency that doesn't play well with windows.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -50,7 +50,7 @@ image =
   scikit-image>=0.16.0
   scipy>=1.3.0
 jpeg2000 =
-  glymur>=0.8.18,!=0.9.0,<0.9.5
+  glymur>=0.8.18,!=0.9.0,!=0.9.5
 map =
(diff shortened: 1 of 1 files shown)


Cadair:
We should open an upstream issue about this.

Cadair (dismissed):

setup.cfg
```diff
@@ -50,7 +50,7 @@ image =
   scikit-image>=0.16.0
   scipy>=1.3.0
 jpeg2000 =
-  glymur>=0.8.18,!=0.9.0
+  glymur>=0.8.18,!=0.9.0,<0.9.5
```
Cadair, line 53: hmm, we could do this:
```suggestion
  glymur>=0.8.18,!=0.9.0,!=0.9.5
```
and hope that it's fixed in the next release? (Which I would say is probable).

Cadair (approved):

PR #5680, opened 2021-11-05
Use default MPL colour cycle when plotting XRS data

dstansby:
This looks much better in my opinion, and will also respect any custom MPL colour cycles people have going on in their themes.

### DIFF
--- a/sunpy/timeseries/sources/goes.py
+++ b/sunpy/timeseries/sources/goes.py
@@ -87,10 +87,10 @@ class XRSTimeSeries(GenericTimeSeries):
         self._validate_data_for_plotting()
         dates = matplotlib.dates.date2num(parse_time(self.to_dataframe().index).datetime)
         axes.plot_date(
-            dates, self.to_dataframe()["xrsa"], "-", label=r"0.5--4.0 $\AA$", color="blue", lw=2, **kwargs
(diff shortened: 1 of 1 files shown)


ayshih:
The discussion to be had is whether the plotting should strive to match the standard GOES/XRS plots we've been looking at for decades.  After all, for Maps we copy the colormaps that each instrument uses.

dstansby:
Sorry, that's my ignorance 😄 . Perhaps a good halfway house would be to use `tab:red` and `tab:blue` (ie. the red/blue in the default MPL colour cycle)? Happy to keep it as is though and I can add a comment to stop other punters coming along with a version of this PR in the future.

ayshih:
> Perhaps a good halfway house would be to use `tab:red` and `tab:blue` (ie. the red/blue in the default MPL colour cycle)?

That'd be similar to what you did with the RHESSI colors (#4330).  However, the GOES/XRS colors are far more broadly known that there may be far more resistance to change.

Even if we do stick with the traditional colors, it'd be good to make it easier for users to use different colors.

dstansby:
I'm going to close this since I think https://github.com/sunpy/sunpy/pull/5680#issuecomment-962169521 is a good argument to leave them as is.

PR #5681, opened 2021-11-06
Speed up VSO time parsing in some cases

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/5107. This PR updates the time parsing so that an attempt is made at making a vectorised call to `parse_time`. If this isn't possible, it falls back to individual `parse_time` calls as before.

I've tested this locally with a search that does benefit from this, and one that doesn't (to check it's not a regression in that case):
```python
from sunpy.net import Fido

# Doesn't benefit from improvements
# 16 seconds before
# 16 seconds after
Fido.search(a.Time('2012/1/1', '2013/1/1'), a.Instrument('EIT'))

# Benefits from improvements
# 30 seconds before
# 21 seconds after
Fido.search(a.Time('2020/01/28', '2020/01/29'), a.Instrument('WISPR'))
```

### DIFF
--- a/sunpy/net/vso/table_response.py
+++ b/sunpy/net/vso/table_response.py
@@ -4,10 +4,12 @@ Classes and helper functions for VSO responses.
 from collections import defaultdict
 from collections.abc import Mapping
 
+import numpy as np
 from zeep.helpers import serialize_object
 
 import astropy.units as u
 from astropy.table import TableAttribute
+from astropy.time import Time
 
 from sunpy.net.base_client import QueryResponseTable
 from sunpy.time import parse_time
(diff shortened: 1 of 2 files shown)


dstansby:
Note to self when I get back to this, `Time(val=0, val2=np.nan, format='mjd')` is the best way to create a masked time at the moment in astropy.

Cadair (approved):

Cadair:
Would be nice to see the online build go green :nauseated_face:

sunpy/net/vso/table_response.py
```diff
@@ -91,17 +93,38 @@ def from_zeep_response(cls, response, *, client, _sort=True):
...
 
-        # Reorder the columns to put the most useful ones first.
         data = cls(data, client=client)
+        # Parse times
+        for col in data.colnames:
+            if col.endswith('Time'):
+                try:
+                    # Try to use a vectorised call to parse_time
+                    data[col] = parse_time(data[col])
+                except Exception:
```
wtbarnes, line 110: Can we catch a more specific exception here?

wtbarnes (approved):
LGTM. My question/suggestion is not blocking.

sunpy/net/vso/table_response.py
dstansby: I think it's worth catching everything and falling back on the non-vectorised approach. If there's something very wrong (like an individual time can't be parsed, then an appropriate error will be raised in the except block.

PR #5686, opened 2021-11-10
Don't use remote data for difference image example

dstansby:
This is one of the examples that regularly fails on CI. Instead of downloading some maps, we can use two cutouts that are already in the sample data. I also added a colorbar.


dstansby:
Here's the new image
sphx_glr_difference_images_002

Cadair (approved):

PR #5687, opened 2021-11-10
Force `_new instance` to preserve plot settings 

### LINKED ISSUE #5667: `.submap()` doesn't preserve `.plot_settings['norm']`
Creating a submap with `GenericMap.submap()` does not preserve any custom set norm (ie. `.plot_settings['norm']`) in the new map.

wtbarnes:
Fixes #5667 

This PR explicitly updates the plot settings when creating a new map instance through `_new_instance`. This is needed because we explicitly set the plot settings in the constructors of some map sources and those constructors ignore the presence of the `plot_settings` kwarg.

I've only backported this to 3.1 because I've also included fixes on the arithmetic operations. If we really want to backport this to 3.0, I can separate these out into a separate PR and backport this one to 3.0 as well.

### DIFF
--- a/changelog/5687.bugfix.rst
+++ b/changelog/5687.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed a bug where custom values in the ``plot_settings`` dictionary were not being propagated
+to new map instances created when calling map methods (e.g. ``.submap``).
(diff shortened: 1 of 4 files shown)


nabobalis:
Do a manual backport to 3.0?

dstansby (approved):

Cadair (approved):

Issue #5670, opened 2021-11-03
read_cdf has issues if unit string not understood


hayesla:
### Description

The new functionality for reading CDF files is great! However I've come across an issue when testing with some SolO EPD files

e..g:

```python
>>> read_cdf('solo_l2_epd-step-hcad_20210728_v01.cdf')

ValueError: 'seconds' did not parse as unit: At col 0, seconds is not a valid unit. Did you mean second? If this is meant to be a custom unit, define it with 'u.def_unit'. To have it recognized inside a file reader or other code, enable it with 'u.add_enabled_units'. For details, see https://docs.astropy.org/en/latest/units/combining_and_defining.html
```
This lies in [here](https://github.com/sunpy/sunpy/blob/main/sunpy/io/cdf.py#L65) , where if units are given in the metadata it tried to pass the string to `u.Unit`, however it seems the case that its not always recognized i.e. `seconds` vs `second`. Similarly, there are also units such as `counts/s` and `particles / (s cm^2 sr MeV)` that are also an issue for this example file. 

### System Details
- SunPy Version: 3.1.0
- Astropy Version: 4.3.1

### Steps to reproduce
```python
from sunpy.net import Fido
from sunpy.net import attrs as a
from sunpy.io.cdf import read_cdf

trange = a.Time('2021/07/28', '2021/07/28')
dataset = a.cdaweb.Dataset('SOLO_L2_EPD-STEP-HCAD')
result = Fido.search(trange, dataset)
file = Fido.fetch(result[0])
read_cdf(file[0])
```

dstansby:
I ran into this yesterday too! The issue is that there is no standard within CDF files on how to format the units and so they can be arbitrary strings 😢

For now the workaround is to define your own units in `astropy`:
```python
u.add_enabled_units([u.def_unit('seconds', represents=u.s), ...])
```

Do you have any thoughts on a more long term solution? Perhaps we could have a dictionary that converts from known unit strings to units? This is what I did in `heliopy`:
https://github.com/heliopython/heliopy/blob/384ac8eff05b8af1b2e47cb9a037ec10510e3b0e/heliopy/data/helper.py#L143-L147

hayesla:
yeah such a pain that there's no standard! 

I think something like a dictionary like what you had in `heliopy` probably makes the most sense - maybe one that just registers known unit strings to astropy units using the `u.add_enabled_units`?

Cadair:
[...]

PR #5692, opened 2021-11-11
Add dictionary of CDF unit conversions

dstansby:
Because CDF files do not have a standard for unit strings (😭😭😭😭😭), there are many CDF files available that do not have unit strings that astropy understands. I *think* this is the best solution - we just keep a mapping of obvious unit strings to their units. The list here is ported over from heliopy where I took the same approach.

Comments and disagreement welcome - but I can't think of a better way to do this from a reducing user friction point of view.

Fixes https://github.com/sunpy/sunpy/issues/5670

### DIFF
--- a/sunpy/io/cdf.py
+++ b/sunpy/io/cdf.py
@@ -6,6 +6,7 @@ import astropy.units as u
 
 from sunpy import log
 from sunpy.timeseries import GenericTimeSeries
+from sunpy.util.exceptions import warn_user
 
 __all__ = ['read_cdf']
 
@@ -63,11 +64,18 @@ def read_cdf(fname):
 
             data = cdf.varget(var_key)
             # Get units
-            unit = attrs['UNITS']
-            if unit in ['None', '', 'unitless']:
-                unit = u.dimensionless_unscaled
-            else:
-                unit = u.Unit(unit)
+            unit_str = attrs['UNITS']
+            try:
+                unit = u.Unit(unit_str)
+            except ValueError:
+                if unit_str in _known_units:
+                    unit = _known_units[unit_str]
+                else:
+                    warn_user(f'astropy did not recognize units of "{unit_str}". '
+                              'Assigning dimensionless units. '
+                              'If you think this unit should not be dimensionless, '
+                              'please raise an issue at https://github.com/sunpy/sunpy/issues')
+                    unit = u.dimensionless_unscaled
 
             if data.ndim == 2:
                 # Multiple columns, give each column a unique label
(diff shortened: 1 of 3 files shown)


sunpy/io/cdf.py
```diff
@@ -63,11 +64,18 @@ def read_cdf(fname):
...
-            else:
-                unit = u.Unit(unit)
+            unit_str = attrs['UNITS']
+            try:
+                unit = u.Unit(unit_str)
+            except ValueError as e:
+                if unit_str in _known_units:
+                    unit = _known_units[unit_str]
+                else:
+                    raise warn_user(f'astropy did not recognize units of "{unit_str}". '
```
hayesla, line 74:
```suggestion
                    warn_user(f'astropy did not recognize units of "{unit_str}". '
```
I dont think you need `raise` here - it causes an error if its there

sunpy/io/cdf.py
```diff
@@ -63,11 +64,18 @@ def read_cdf(fname):
 
             data = cdf.varget(var_key)
             # Get units
-            unit = attrs['UNITS']
```
hayesla, line 66: should this be kept? If there are data that have no units (and shouldnt have them) then this makes sense, and then you done have the warnings thrown at you?

sunpy/io/cdf.py
```diff
@@ -82,3 +90,81 @@ def read_cdf(fname):
...
+                '1/(cm**2-s-sr-MeV)': 1 / (u.cm**2 * u.s * u.sr * u.MeV),
+                '1/(cm**2-s-sr-MeV/nuc.)': 1 / (u.cm**2 * u.s * u.sr * u.MeV),
+                '1/(cm^2 sec ster MeV)': 1 / (u.cm**2 * u.s * u.sr * u.MeV),
+
+                '1/(cm**2-s-sr)': 1 / (u.cm**2 * u.s * u.sr),
+                '1/(SQcm-ster-s)': 1 / (u.cm**2 * u.s * u.sr),
+                '1/(SQcm-ster-s)..': 1 / (u.cm**2 * u.s * u.sr),
+
+                'Counts/256sec': 1 / (256 * u.s),
+                'Counts/hour': 1 / u.hr,
```
hayesla, line 169:
```suggestion
                'counts/s': 1 / u.s,
```

sunpy/io/cdf.py
```diff
@@ -82,3 +90,81 @@ def read_cdf(fname):
...
+                }
```
hayesla, line 170:
```suggestion
                'counts / s': 1/u.s
                }
```

sunpy/io/cdf.py
dstansby: Woops, thanks!

sunpy/io/cdf.py
dstansby: The line below is the same, but I've just renamed `unit` > `unit_str`?

nabobalis (approved):

hayesla (approved):

Cadair (approved):

Cadair (changes requested):
This should have a test or two?

PR #5707, opened 2021-11-18
Fix horizons deprecation warning

dstansby:
This fixes a deprecation warning with `astroquery` 0.4.4, where the default `id_type` changed in horizons queries. I've changed our default `id_type` to `None` to match. To keep back compatibility in sunpy, if `astroquery < 0.4.4` is being used `id_type= None` gets translated to `id_type='majorbody'`, the old default.

I'm unsure what to do re. backporting here.


nabobalis:
I think it would be safe to backport?

nabobalis (approved):

PR #5708, opened 2021-11-18
Register REMOTE_DATA option

dstansby:
See https://github.com/matplotlib/matplotlib/issues/21668

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -49,6 +49,11 @@ project = 'SunPy'
 author = 'The SunPy Community'
 copyright = '{}, {}'.format(datetime.datetime.now().year, author)
 
+
+# Register remote data option with doctest
+import doctest  # NOQA
+REMOTE_DATA = doctest.register_optionflag('REMOTE_DATA')
+
 # The full version, including alpha/beta/rc tags
 release = __version__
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -49,6 +49,11 @@
 author = 'The SunPy Community'
 copyright = '{}, {}'.format(datetime.datetime.now().year, author)
 
+
+# Register remote data option with doctest
```
nabobalis, line 53: Does this need to go upstream to pytest-remote-data?

dstansby: If anywhere, to https://github.com/astropy/pytest-doctestplus - see discussion on https://github.com/matplotlib/matplotlib/issues/21668

nabobalis: Right that one!

nabobalis (approved):

nabobalis:
Some of those online fails are real but I guess we can deal with them in another PR.

dstansby:
> Some of those online fails are real but I guess we can deal with them in another PR.

Here's one I made earlier: https://github.com/sunpy/sunpy/pull/5707

dstansby:
Going to merge as this unblocks the doc build.

PR #5713, opened 2021-11-19
Remove all usage of distutils as it is now deprecated

Cadair:
### Description

Fixes #<Issue Number>

### DIFF
--- a/changelog/5713.trivial.rst
+++ b/changelog/5713.trivial.rst
@@ -1,0 +1,1 @@
+Added `packaging <https://pypi.org/project/packaging/>`__ as a core depedency as distutils is now deprecated.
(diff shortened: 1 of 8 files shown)


dstansby:
I'm guessing this is worth backporting? It does introduce a new runtime dependency though...

Cadair:
We could get away without backporting it, but it's probably a good idea. It should be noted that astropy has also introduced this runtime dep as well.

dstansby (dismissed):
Should probably add a changelog to notify users of the new dependency, even if it is likely that they already have it installed.

nabobalis (approved):
Changelog has been added.

nabobalis:
Backported to 3.0 manually.

PR #5714, opened 2021-11-19
max conns to JSOC is always 1

nabobalis:


wtbarnes:
It seems a bit extreme to *force* only 1 max_split and max_conn. Is it possible to just make the defaults 1?

nabobalis:
JSOC say it should only be one.

changelog/5714.bugfix.rst
```diff
@@ -0,0 +1 @@
+Only open one connection to the JSOC for any download.
```
Cadair, line 1: We should explain why we are making this change here (i.e. JSOC told us to do this), also we should make it clear that parameters are overriden, especially as we are backporting it as well.

sunpy/net/jsoc/jsoc.py
```diff
@@ -524,11 +524,13 @@ def get_request(self, requests, path=None, overwrite=False, progress=True,
...
-                      f"Changing the number of parallel connections to {2 * self.default_max_conn}.")
-            kwargs['max_splits'] = 2
-            downloader.max_conn = self.default_max_conn
+        # Private communication from JSOC say we should not use more than one connection.
+        if downloader.max_conn * kwargs['max_splits'] > 1:
+            warn_user(
+                "JSOC does not support more than 1 parallel connection, changing the number to 1."
+            )
+            kwargs['max_splits'] = 1
+            downloader.max_conn = 1
```
Cadair, line 533: This is a problem, as it affects downloads from all clients using this downloader not just JSOC.

I am pro setting the splits to 1, I am less keen about doing this.

Cadair (changes requested):
I would like to consider the knock-on impacts this has on mixed-client downloads and the general poor experience for people who are now used to parallel downloads even when just downloading from JSOC.

changelog/5714.bugfix.rst
nabobalis: Changed.

sunpy/net/jsoc/jsoc.py
nabobalis: Changed.

sunpy/net/jsoc/jsoc.py
```diff
@@ -437,7 +437,7 @@ def fetch(self, jsoc_response, path=None, progress=True, overwrite=False,
 
         return self.get_request(responses, path=path, overwrite=overwrite,
                                 progress=progress, downloader=downloader,
-                                wait=wait, max_conn=max_conn, **defaults)
+                                wait=wait, max_conn=1, **defaults)
```
Cadair, line 440: revert this, it's redundant.

sunpy/net/jsoc/jsoc.py
```diff
@@ -522,13 +523,8 @@ def get_request(self, requests, path=None, overwrite=False, progress=True,
...
             dl_set = False
-            downloader = Downloader(progress=progress, overwrite=overwrite, max_conn=max_conn)
-
-        if downloader.max_conn * kwargs['max_splits'] > 10:
-            warn_user("JSOC does not support more than 10 parallel connections. " +
-                      f"Changing the number of parallel connections to {2 * self.default_max_conn}.")
-            kwargs['max_splits'] = 2
-            downloader.max_conn = self.default_max_conn
+            # Private communication from JSOC say we should not use more than one connection.
+            downloader = Downloader(progress=progress, overwrite=overwrite, max_conn=1)
```
Cadair, line 527: While this change is good in isolation, it has no effect on queries passed down from a Fido search.

Cadair:
I am treating this as a bugfix for backporting, if we merge this then a second PR actually removing all these kwargs needs opening.

As this is (with out the max_conn setting through Fido) I am :+1: on this, I have no issue with completely disabling chunked downloads for JSOC at their behest.

That being said, I do wonder if there is any point in the max_conn changes in this PR, as it only affects people directly using the JSOCClient, and it's easy to override even for them by passing through a new `Downloader` object. Why don't we just remove that option in a not-backport PR by pulling the keyword arg completely?

Cadair:
To be clear, the proper long term fix here is https://github.com/Cadair/parfive/issues/39 where the JSOCClient can say to parfive "only one connection at once to jsoc.stamford.edu please" and it can have 10,000 open connections to somewhere else at the same time for all JSOCClient cares.

wtbarnes:
> To be clear, the proper long term fix here is [Cadair/parfive#39](https://github.com/Cadair/parfive/issues/39) where the JSOCClient can say to parfive "only one connection at once to jsoc.stanford.edu please" and it can have 10,000 open connections to somewhere else at the same time for all JSOCClient cares.

Is there any reason to not wait on that fix then and just pin to the latest version of parfive or higher?

sunpy/net/jsoc/jsoc.py
nabobalis: So what should I do?

nabobalis:
> That being said, I do wonder if there is any point in the max_conn changes in this PR, as it only affects people directly using the JSOCClient, and it's easy to override even for them by passing through a new `Downloader` object. Why don't we just remove that option in a not-backport PR by pulling the keyword arg completely?

You mean remove the keyword in the fetch for jsoc?

changelog/5714.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Due to consultation with JSOC, we will now limit all downloads to the JSOC to one connection.
+This will override all user settings passed to the downloader.
```
dstansby, line 2:
```suggestion
In consultation with JSOC, we now limit all JSOC downloads to one connection.
This will override all connection user settings passed to the downloader.
```

sunpy/net/fido_factory.py
```diff
@@ -385,10 +385,23 @@ def fetch(self, *query_results, path=None, max_conn=5, progress=True,
         if "wait" in kwargs:
             raise ValueError("wait is not a valid keyword argument to Fido.fetch.")
 
+        # TODO: Hack for JSOC to avoid more than one connection.
+        # Remove when parfive allows us to special case URLS.
+        is_jsoc_only = False
```
dstansby, line 390: Does this mean a query that returns non-JSOC and JSOC results will still use more than one connection to download the JSOC results?

sunpy/net/jsoc/jsoc.py
```diff
@@ -522,13 +523,8 @@ def get_request(self, requests, path=None, overwrite=False, progress=True,
...
         if not downloader:
             dl_set = False
-            downloader = Downloader(progress=progress, overwrite=overwrite, max_conn=max_conn)
-
-        if downloader.max_conn * kwargs['max_splits'] > 10:
-            warn_user("JSOC does not support more than 10 parallel connections. " +
-                      f"Changing the number of parallel connections to {2 * self.default_max_conn}.")
-            kwargs['max_splits'] = 2
-            downloader.max_conn = self.default_max_conn
+            # Private communication from JSOC say we should not use more than one connection.
```
dstansby, line 526: Can we still log a message to `DEBUG` if a user is trying to pass a different `max_conn`?

sunpy/net/fido_factory.py
```diff
@@ -385,10 +385,23 @@ def fetch(self, *query_results, path=None, max_conn=5, progress=True,
...
+        for query_result in query_results:
+            if isinstance(query_result, UnifiedResponse):
+                is_jsoc_only = all([result.client.__class__.__name__ ==
+                                    "JSOCClient" for result in query_result])
```
Cadair, line 394: Why not use `isinstance` or `issubclass` here?

sunpy/net/fido_factory.py
```diff
@@ -385,10 +385,23 @@ def fetch(self, *query_results, path=None, max_conn=5, progress=True,
...
+            elif isinstance(query_result, QueryResponseTable):
+                is_jsoc_only = all([result.table.client.__class__.__name__ ==
+                                    "JSOCClient" for result in query_result])
```
Cadair, line 397: same here

sunpy/net/jsoc/jsoc.py
Cadair: Something higher than that is probably justified, if someone tries to set a thing and we ignore it we should print a message which is visible by default (which means INFO or WARNING iirc).

sunpy/net/fido_factory.py
Cadair: yes.

sunpy/net/jsoc/jsoc.py
```diff
@@ -475,7 +475,8 @@ def get_request(self, requests, path=None, overwrite=False, progress=True,
         """
         c = drms.Client()
 
-        kwargs['max_splits'] = kwargs.get('max_splits', 2)
+        # Private communication from JSOC say we should not use more than one connection.
+        kwargs['max_splits'] = 1
```
Cadair, line 479: We should throw a INFO log here as well if the kwarg is passed.

sunpy/net/fido_factory.py
nabobalis: I don't know.  That makes more sense.

sunpy/net/jsoc/jsoc.py
nabobalis: Fixed.

sunpy/net/jsoc/jsoc.py
nabobalis: Fixed.

sunpy/net/fido_factory.py
nabobalis: I added more of a comment here.

sunpy/net/jsoc/jsoc.py
```diff
@@ -522,13 +525,10 @@ def get_request(self, requests, path=None, overwrite=False, progress=True,
...
-            downloader = Downloader(progress=progress, overwrite=overwrite, max_conn=max_conn)
-
-        if downloader.max_conn * kwargs['max_splits'] > 10:
-            warn_user("JSOC does not support more than 10 parallel connections. " +
-                      f"Changing the number of parallel connections to {2 * self.default_max_conn}.")
-            kwargs['max_splits'] = 2
-            downloader.max_conn = self.default_max_conn
+            # Private communication from JSOC say we should not use more than one connection.
+            if max_conn != self.default_max_conn:
+                log.info(f"max_conn keyword was passed and set to 1.")
```
Cadair, line 530: This message isn't very clear. Something like "Setting max parallel downloads to 1 for JSOC client." would be better

Cadair (approved):

sunpy/net/jsoc/jsoc.py
nabobalis: Fixed.

nabobalis:
I will manually backport to each.

PR #5719, opened 2021-11-20
Fix example in docbuild

nabobalis:
Is it a fix? Who knows.

### DIFF
--- a/sunpy/timeseries/sources/eve.py
+++ b/sunpy/timeseries/sources/eve.py
@@ -84,8 +84,9 @@ class ESPTimeSeries(GenericTimeSeries):
             axes = self.to_dataframe().plot(subplots=True, sharex=True, **kwargs)
         for i, ax in enumerate(axes):
             if predefined_axes:
-                ax.plot(self._data[self._data.columns[i]], color=colors[i %
-                        len(colors)], label=self._data.columns[i])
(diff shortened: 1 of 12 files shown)


sunpy/timeseries/sources/eve.py
```diff
@@ -75,15 +75,17 @@ def plot(self, axes=None, **kwargs):
...
         predefined_axes = False
         if isinstance(axes, np.ndarray):
             predefined_axes = True
         elif axes is None:
             axes = self.to_dataframe().plot(subplots=True, sharex=True, **kwargs)
         for i, ax in enumerate(axes):
             if predefined_axes:
-                ax.plot(self._data[self._data.columns[i]], color=colors[i%len(colors)], label=self._data.columns[i])
+                ax.plot(self._data[self._data.columns[i]], color=colors[i %
+                        len(colors)], label=self._data.columns[i])
```
Cadair, line 88: That is a truly horrifying place for a line break. PEP 8 does say that you shouldn't line break around operators iirc.

Cadair:
```suggestion
                ax.plot(self._data[self._data.columns[i]],
                        color=colors[i % len(colors)],
                        label=self._data.columns[i])
```

Cadair (approved):

PR #5720, opened 2021-11-23
Time formatting is ugly on some timeseries plots

nabobalis:
See post below for before and after.

- [x] Changelog
- [x] Figure hashes


sunpy/timeseries/sources/eve.py
```diff
@@ -75,21 +75,26 @@ def plot(self, axes=None, **kwargs):
...
         predefined_axes = False
         if isinstance(axes, np.ndarray):
             predefined_axes = True
         elif axes is None:
             axes = self.to_dataframe().plot(subplots=True, sharex=True, **kwargs)
         for i, ax in enumerate(axes):
             if predefined_axes:
-                ax.plot(self._data[self._data.columns[i]], color=colors[i%len(colors)], label=self._data.columns[i])
+                ax.plot(self._data[self._data.columns[i]], color=colors[i %
+                        len(colors)], label=self._data.columns[i])
```
nabobalis, line 88: We should change the line length in setup.cfg

wtbarnes: This is a really weird line break. Is that from pre-commit?

sunpy/timeseries/sources/eve.py
```diff
@@ -75,21 +75,26 @@ def plot(self, axes=None, **kwargs):
...
                 plt.xticks(rotation=30)
             ax.set_ylabel(names[i])
             ax.legend(loc="upper right")
         axes[-1].set_xlim(self.to_dataframe().index[0], self.to_dataframe().index[-1])
-        axes[-1].set_xlabel("Time (UT) " + str(self.to_dataframe().index[0])[0:11])
-        axes[-1].xaxis.set_major_formatter(dates.DateFormatter("%H:%M"))
+        # TODO: Work out a way to set this based on the timespan of the data.
```
wtbarnes, line 93: Maybe this is a dumb question, but why aren't we using `astropy.visualization.time_support`? Is this incompatible with pandas?

sunpy/timeseries/sources/lyra.py
```diff
@@ -77,15 +78,17 @@ def plot(self, axes=None, names=3, **kwargs):
...
         predefined_axes = False
         if isinstance(axes, np.ndarray):
             predefined_axes = True
         elif axes is None:
             axes = self.to_dataframe().plot(subplots=True, sharex=True, **kwargs)
         for i, name in enumerate(self.to_dataframe().columns):
             if predefined_axes:
-                axes[i].plot(self._data[self._data.columns[i]], color=colors[i%len(colors)], label=self._data.columns[i])
+                axes[i].plot(self._data[self._data.columns[i]], color=colors[i %
+                             len(colors)], label=self._data.columns[i])
```
wtbarnes, line 91: As above, this is a terrible place to break a line

examples/acquiring_data/goes_xrs_example.py
```diff
@@ -89,6 +89,12 @@
 fig, ax = plt.subplots()
 ax.plot(goes_flare.index, np.gradient(goes_flare.quantity("xrsb")))
 ax.set_ylabel("Flux (Wm$^{-2}$$s^{-1}$)")
+# You may notice that the time axis in the above example is not really well laid out.
+# We can do the following to fix that, this is done by the ``plot`` method.
```
wtbarnes, line 93: What do you mean by "this is done by the plot method"? I see that this code is now included in the `plot` methods for each source. I wonder if we should be less opinionated in `plot()` and then just have this example here showing how to fix the labels in the case where they're not aesthetically pleasing?

sunpy/timeseries/sources/eve.py
nabobalis: This is autopep8.

sunpy/timeseries/sources/eve.py
nabobalis: Not a clue, never used it.

examples/acquiring_data/goes_xrs_example.py
nabobalis: Not sure really.

dstansby (changes requested):
It's hard to review this without knowing what the issues it's fixing are - @nabobalis could you open (or even better, get the original reporter if there is someone else?) an issue describing the issue.

dstansby:
> > It's hard to review this without knowing what the issues it's fixing are - @nabobalis could you open (or even better, get the original reporter if there is someone else?) an issue describing the issue.
> 
> Someone pointed out to me that in the timeseries example with peek, for some sources the x-axis label and formatting is wrong and ugly.
> 
> This attempts to fix that.

Which sources? Can you provide screenshots of the 'ugly' formatting?

nabobalis:
> > > It's hard to review this without knowing what the issues it's fixing are - @nabobalis could you open (or even better, get the original reporter if there is someone else?) an issue describing the issue.
> > 
> > 
> > Someone pointed out to me that in the timeseries example with peek, for some sources the x-axis label and formatting is wrong and ugly.
> > This attempts to fix that.
> 
> Which sources? Can you provide screenshots of the 'ugly' formatting?

See each image in this example https://docs.sunpy.org/en/stable/generated/gallery/acquiring_data/goes_xrs_example.html#sphx-glr-generated-gallery-acquiring-data-goes-xrs-example-py

nabobalis:
What I've done here, needs more testing (longer timeseries mainly) and might not be the nicest style or code. I am looking for feedback on that.

sunpy/timeseries/sources/eve.py
Cadair, line 88:
```suggestion
                ax.plot(self._data[self._data.columns[i]],
                        color=colors[i % len(colors)],
                        label=self._data.columns[i])
```

sunpy/timeseries/sources/lyra.py
Cadair, line 91:
```suggestion
                axes[i].plot(self._data[self._data.columns[i]],
                             color=colors[i % len(colors)],
                             label=self._data.columns[i])
```

wtbarnes:
Differences before and after this pr:
image

sunpy/timeseries/sources/eve.py
nabobalis: We don't need it at all it seems. I have removed the whole dates conversion.

examples/acquiring_data/goes_xrs_example.py
nabobalis: I removed all this.

sunpy/timeseries/sources/goes.py
```diff
@@ -66,14 +65,16 @@ class XRSTimeSeries(GenericTimeSeries):
     if h5netcdf_version >= packaging.version.parse("0.10"):
         _netcdf_read_kw['decode_vlen_strings'] = True
 
-    def plot(self, axes=None, **kwargs):
+    def plot(self, axes=None, channels=["xrsa", "xrsb"], **kwargs):
```
nabobalis, line 68: I might have decided this was a good idea.

nabobalis:
You can see the differences here:
https://73000-2165383-gh.circle-artifacts.com/0/.tmp/py38-figure/figure_test_images/fig_comparison.html
and
https://72999-2165383-gh.circle-artifacts.com/0/.tmp/py39-figure-devdeps/figure_test_images/fig_comparison.html

dstansby:
If you want to backport this, the new keyword argument will need to be removed as it's a new feature. Perhaps better to put that in another PR?

nabobalis:
> If you want to backport this, the new keyword argument will need to be removed as it's a new feature. Perhaps better to put that in another PR?

Fair. Maybe I won't backport it.

examples/acquiring_data/goes_xrs_example.py
```diff
@@ -86,10 +81,7 @@
 # flares.
 
 goes_flare = goes_15.truncate("2015-06-21 09:35", "2015-06-21 10:30")
-fig, ax = plt.subplots()
-ax.plot(goes_flare.index, np.gradient(goes_flare.quantity("xrsb")))
```
hayesla, line 90: this is plotting the derivative here not just the flare so `goes_flare.plot()` isn't the same thing.

sunpy/timeseries/sources/goes.py
hayesla: this is a good idea - I guess the only concern is that it isn't consistent across the other `timeseries` sources. Could we make it more generic and have "columns" and then it can be extended to all sources?

sunpy/timeseries/sources/goes.py
```diff
@@ -161,7 +157,7 @@ def peek(self, title="GOES Xray Flux", **kwargs):
         Parameters
         ----------
         title : `str`, optional
-            The title of the plot. Defaults to "GOES Xray Flux".
+            The title of the plot. Defaults to "GOES x-ray flux".
```
hayesla, line 160:
```suggestion
            The title of the plot. Defaults to "GOES X-ray flux".
```

examples/acquiring_data/goes_xrs_example.py
nabobalis: Oh I did not notice that.

nabobalis: Fixed.

sunpy/timeseries/sources/goes.py
nabobalis: I am not sure how to make this more general based on the timeseries but it is something we should look at.

dstansby: Yeah, I added this functionality for `GenericTimeseries`, so it would be nice to roll it out to all the sources. I'll open an issue and try and tackle it for 4.0.

dstansby, line 68:
```suggestion
    def plot(self, axes=None, columns=["xrsa", "xrsb"], **kwargs):
```

examples/acquiring_data/goes_xrs_example.py
```diff
@@ -72,11 +72,7 @@
 # channel is known as the "xrsa" channel and the 1-8 angstrom channel is known
 # as the "xrsb" channel.
 
-fig, ax = plt.subplots()
-ax.plot(goes_15.index, goes_15.quantity("xrsb"))
-ax.set_ylabel("Flux (Wm$^{-2}$)")
-ax.set_xlabel("Time")
-fig.autofmt_xdate()
+goes_15.plot(channels=["xrsb"])
```
dstansby, line 75:
```suggestion
goes_15.plot(columns=["xrsb"])
```

dstansby: This makes it a consistent kwarg name with `Generictimesereis`

sunpy/timeseries/sources/goes.py
```diff
@@ -66,14 +65,16 @@ class XRSTimeSeries(GenericTimeSeries):
...
-    def plot(self, axes=None, **kwargs):
+    def plot(self, axes=None, channels=["xrsa", "xrsb"], **kwargs):
         """
         Plots the GOES XRS light curve from a pandas dataframe.
 
         Parameters
         ----------
         axes : `matplotlib.axes.Axes`, optional
             The axes on which to plot the TimeSeries. Defaults to current axes.
+        channels : `list` of {'xrsa', 'xrsb'}, optional
```
dstansby, line 76:
```suggestion
        columns : `list` of {'xrsa', 'xrsb'}, optional
```

sunpy/timeseries/sources/goes.py
```diff
@@ -86,19 +87,19 @@ def plot(self, axes=None, **kwargs):
...
-        axes.plot_date(
-            dates, self.to_dataframe()["xrsa"], "-", label=r"0.5--4.0 $\AA$", color="blue", lw=2, **kwargs
-        )
-        axes.plot_date(
-            dates, self.to_dataframe()["xrsb"], "-", label=r"1.0--8.0 $\AA$", color="red", lw=2, **kwargs
-        )
-
+        plot_settings = {"xrsa": ["blue", r"0.5--4.0 $\AA$"], "xrsb": ["red", r"1.0--8.0 $\AA$"]}
+        data = self.to_dataframe()
+        for channel in channels:
```
dstansby, line 92:
```suggestion
        for channel in columns:
```

dstansby (commented):
This looks 👍, modulo renaming the argument to columns

nabobalis:
> This looks 👍, modulo renaming the argument to columns

Fixed.

sunpy/timeseries/sources/goes.py
nabobalis: Ah I forgot about that. I have renamed channels to columns.

nabobalis:
The bot HTML figure artifact link is not showing up as a check? @Cadair

nabobalis:
For 3.0, I will manually backport just the axis labels changes.

Cadair:
Yeah it looks like Giles bot is in trouble.

dstansby (changes requested):
👍 but figure hashes need updating

nabobalis:
I've updated them, hopefully it will pass now.

dstansby (approved):

dstansby:
This can't be backported because it has an API addition (the new columns argument), so I've removed the backport tag. If someone wants to backport the non-API bits of this as a bugfix then it will have to be done manually.

hayesla (approved):

nabobalis:
I've backported this to 3.1 minus the keyword change. 

In 3.0 we still use peek and plot does not exist, so I will not be back porting it to the LTS.

Issue #3342, opened 2019-09-09
Logger configuration issues


ayshih:
Our logger is based on `AstropyLogger`, and we used its method `_set_defaults()` to configure our logger.  However, since `_set_defaults()` refers to Astropy's configuration variables, our implementation takes the "interesting" approach of overwriting Astropy's configuration values with SunPy's configuration values.  Philosophical issues aside, this doesn't immediately cause problems because the respective loggers are typically configured solely upon import, with the Astropy logger configured before the SunPy logger, and so Astropy typically won't be affected by the overwritten variables.  However, there is still the potential for confusion, if not outright issues (e.g., if the Astropy logger is re-instantiated).  At the very least, the Astropy's configuration values should be restored after our logger is instantiated.

The implementation also has a bug where if a SunPy configuration value is not provided, the configuration value that is actually used is whatever Astropy's value is, which may not be the default.

Finally, the existing configuration tests for our logger aren't very good.  They only check for consistency between the logger's settings and the configuration values, and not whether changing the configuration values actually changes the logger's settings.

wtbarnes:
Is the ask here to build our own logger rather than use the astropy logger?

Cadair:
I have learnt some more about loggers in the last 7! years, the DKIST implementation of the logger is better and less tied to the astropy machinery

PR #2980, opened 2019-03-12
Adding Logger to SunPy

ehsteve:
### Description
This PR addresses issue #1859 by adding a logger based on Astropy's logger. It adds the ability to configure the logger within the sunpy configuration file.

Fixes #1859.

To do
- [x] add supporting documentation for users
- [x] add supporting documentation and standards for developers
- [x] add changelog
- [x] add tests

PR #5722, opened 2021-11-29
Fixed our logger intercepting Astropy warnings

ayshih:
We crib Astropy's logger (#2980), which causes some issues (see #3342).  We recently became aware that our logger captures Astropy warnings and prevents them from being handled elsewhere, which can cause test failures in other packages if SunPy gets imported (see astropy/astropy#12425).  Surprisingly, an implicit import of SunPy can occur if `asdf` is used, so a user may not even be aware that SunPy has been imported.

The long-term solution is to build our own logger.  In the meantime, this PR creates a lightweight subclass of `AstropyLogger` that overrides the warning-capturing code to capture `SunpyWarning` instead of `AstropyWarning`.  This resolves the issue with our logger capturing Astropy warnings, but does not resolve issues noted in #3342.

This PR also changes `SunpyWarning` so that it does not inherit from `AstropyWarning`, which had been the easiest way to crib Astropy's logger, but is no longer necessary given the custom warning-capturing code.  Furthermore, this prevents the issue that the Astropy logger could capture SunPy warnings.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -106,7 +106,7 @@ repos:
     hooks:
       - id: prohibit-string
         args: [--prohibit-string, "from warnings import warn,warnings.warn"]
-        exclude: "sunpy/version.py|sunpy/util/decorators.py|sunpy/util/exceptions.py|sunpy/extern/"
+        exclude: "sunpy/version.py|sunpy/util/decorators.py|sunpy/util/exceptions.py|sunpy/extern/|sunpy/util/tests/test_logger.py"
 
 ci:
   autofix_prs: false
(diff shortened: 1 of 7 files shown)


Cadair:
Appears sane enough from a quick glance.

ayshih:
Now actually tested!  Ready for review

Cadair (approved):

dstansby:
Do we still want to backport this to 3.0.x? I've just put out 3.0.3 without this as it didn't have a manual backport label (sorry!), but it can go in 3.0.4.

nabobalis:
I removed the label for 3.0 backport. I think we can leave it for 3.0.x unless someone has an objection?

PR #5742, opened 2021-12-03
Get scikit-image 0.19.0 working

ayshih:
The just-released `scikit-image` 0.19 has deprecated the `multichannel` keyword argument for `rescale()`, which breaks our CI build.  We were specifying the default value of `multichannel` anyway, so we can simply remove it where used.

`scikit-image` 0.19 also has different behavior when filling in constant values beyond the image when shrinking.  This PR is currently testing possible solutions.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,9 +48,7 @@ dask =
 database =
   sqlalchemy>=1.3.4
 image =
-  # TODO: remove the <0.19 pin when discrepenies in
-  # sunpy/image/tests/test_transform.py::test_scale have been resolved
-  scikit-image>=0.16.0,<0.19
+  scikit-image>=0.16.0
   scipy>=1.3.0
 jpeg2000 =
   glymur>=0.8.18,!=0.9.0,!=0.9.5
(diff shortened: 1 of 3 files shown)


nabobalis (approved):

ayshih:
Ugh, real failures, I'll have to look into this later

dstansby:
I had a look on the `scikit-image` side, and found that the output of `rescale` is different between versions 0.18.3 and 0.19.0. I've opened an upstream report here: https://github.com/scikit-image/scikit-image/issues/6093

nabobalis:
Looks like upstream will be patching this back. 

So we can close this PR and wait for the bugfix release?

dstansby:
Sounds good - we just need to remember to change the pin from `<` to `!=` when they have put out a patch.

ayshih:
Re-opening because upstream is not going to revert after all (see https://github.com/scikit-image/scikit-image/issues/6093#issuecomment-994292097)

nabobalis:
Should we close this in favor of https://github.com/sunpy/sunpy/pull/5867?

PR #5749, opened 2021-12-06
Always run figure tests without comparison

### LINKED ISSUE #5541: Allow figure tests to be run without figure comparison at the end
Currently it's only possible to run figure tests in a specific environment, to make sure we have the correct version of packages to do the image comparison at the end. It would be helpful to be able to run the figure tests in other environments to check the code works, without doing the figure test at the end.

e.g., if I try and run a figure test locally that would fail because of an exception being raise (before a test image is even generated) I get:

```python
$ pytest sunpy/map/tests/test_plotting.py::test_draw_limb_heliographic_stonyhurst
/Users/dstansby/miniconda3/envs/dev/lib/python3.9/site-packages/pep8.py:110: FutureWarning: Possible nested set at position 1
  EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')
========================================================= test session starts ==========================================================
platform darwin -- Python 3.9.7, pytest-6.2.3, py-1.9.0, pluggy-0.13.1
Matplotlib: 3.5.0.dev2088+g876022e616
Freetype: 2.6.1
rootdir: /Users/dstansby/github/sunpy, configfile: setup.cfg
plugins: asdf-2.7.1, arraydiff-0.3, doctestplus-0.8.0, remotedata-0.3.2, hypothesis-6.0.0, filter-subpackage-0.1.1, mock-3.3.1, cov-2.10.1, profiling-1.7.0, openfiles-0.5.0, timeout-1.4.2, astropy-header-0.1.2, pep8-1.0.6, mpl-0.12
collected 1 item / 1 deselected                                                                                                        

======================================================== 1 deselected in 0.91s =========================================================
```
I would like this to error and give me an error message instead.

ConorMacBride:
### Description
- Removed the default pytest option that skips all tests marked with `mpl_image_compare`.
- AFAIK all `@figure_test` tests only use data packaged in `sunpy/data/test`, unless they have been manually marked as `remote_data` also. So the figure test helper no longer adds the `remote_data` mark by default.

Therefore, figure tests are now always run, however, the hashes/baselines are still only checked if the tox env has `-figure`. This change should add no more than ~10 seconds to each CI job.

Fixes #5541


Cadair:
I thought the global `remote-data` flag was so that remote data didn't get in the way of downloading the reference images. Maybe this isn't an issue any more as the circle CI tests seem to pass, if it turns out to be an issue later on then we can deal with it then.

dstansby (changes requested):
👍  - it looks like the `oldestdeps` tests need fixing with to catch a numpy warning, but I like that we're now smoke testing the figure tests across a wide range of platforms/python versions

Cadair (approved):

Cadair:
not exactly sure what's going in with codecov?

Cadair:
:clap:

ConorMacBride:
Maybe the force push confused it? (I forgot to ignore both `greater` and `less`.)

PR #5751, opened 2021-12-08
Handle empty CDF files in CDF reader

dstansby:
For some reason some of the Solar Orbiter particle data includes files that are empty with no data. This PR fixes the CDF reader to skip over these files without erroring and adds logging at the DEBUG level.


nabobalis (approved):

hayesla (approved):

nabobalis:
Test for good measure?

dstansby:
Yeah, I probably should.... unfortunately there's no easy way to write a custom CDF file, so it might have to be a remote test 😢

nabobalis:
> Yeah, I probably should.... unfortunately there's no easy way to write a custom CDF file from python, so it might have to be a remote test 😢

If the remote file is small, we can add it to the test data?

dstansby:
> > Yeah, I probably should.... unfortunately there's no easy way to write a custom CDF file from python, so it might have to be a remote test 😢
> 
> If the remote file is small, we can add it to the test data?

I guess by virtue of it not actually having any data in it it will be nice and small 🤣 . Thanks for the suggestion, I will do that.

nabobalis (approved):

PR #5752, opened 2021-12-08
Tox clean up

nabobalis:
Cleaned up tox file comments and removed pytest-logger and the `-s` flag.

### DIFF
--- a/sunpy/tests/helpers.py
+++ b/sunpy/tests/helpers.py
@@ -14,8 +14,7 @@ from astropy.wcs.wcs import FITSFixedWarning
 
 import sunpy.map
 
-__all__ = ['skip_windows', 'skip_glymur', 'skip_ana', 'skip_32bit',
-           'warnings_as_errors', 'asdf_entry_points']
+__all__ = ['skip_windows', 'skip_glymur', 'skip_ana', 'warnings_as_errors', 'asdf_entry_points']
 
 # SunPy's JPEG2000 capabilities rely on the glymur library.
 # First we check to make sure that glymur imports correctly before proceeding.
@@ -45,9 +44,6 @@ else:
 skip_windows = pytest.mark.skipif(platform.system() == 'Windows', reason="Windows.")
 skip_glymur = pytest.mark.skipif(SKIP_GLYMUR, reason="Glymur can not be imported.")
 skip_ana = pytest.mark.skipif(SKIP_ANA, reason="ANA is not available.")
-skip_32bit = pytest.mark.skipif(SKIP_32, reason="Fails on a 32 bit system.")
-
-
 # Skip if the SunPy ASDF entry points are missing.
 asdf_entry_points = pytest.mark.skipif(not list(pkg_resources.iter_entry_points('asdf_extensions', 'sunpy')),
                                        reason="No SunPy ASDF entry points.")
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -50,7 +50,7 @@ database =
 image =
   # TODO: remove the <0.19 pin when discrepenies in
   # sunpy/image/tests/test_transform.py::test_scale have been resolved
-  scikit-image>=0.16.0,<0.19
+  scikit-image>=0.16.0
```
nabobalis, line 53: Just to check it works on the CI for the reporting.

tox.ini
```diff
@@ -81,28 +59,20 @@ deps =
...
-
-    # These are specific online extras we use to run the online tests.
+    # These are specific extras we use to run in specific factors
     online: pytest-rerunfailures
     online: pytest-timeout
     online: astroquery
     hypothesis: astroquery
-
-    # We have to pin this to use a wheel for manylinux to compile
-    32bit: cryptography<3.0
```
nabobalis, line 92: We don't build 32bit anymore do we?

sunpy/tests/helpers.py
```diff
@@ -45,9 +44,6 @@
 skip_windows = pytest.mark.skipif(platform.system() == 'Windows', reason="Windows.")
 skip_glymur = pytest.mark.skipif(SKIP_GLYMUR, reason="Glymur can not be imported.")
 skip_ana = pytest.mark.skipif(SKIP_ANA, reason="ANA is not available.")
-skip_32bit = pytest.mark.skipif(SKIP_32, reason="Fails on a 32 bit system.")
```
nabobalis, line 48: We don't run anything on 32bit anymore?

ConorMacBride:
I'm just thinking we may need the `-s` to always output the `A summary of the failed tests can be found at: ..../fig_comparison.html`, at least once I sort out https://github.com/matplotlib/pytest-mpl/pull/108. i.e. we want to always compare to baseline images even if the hashes match and the tests pass.

Cadair:
What's making that print? We can probably use pytest to output it with a bit of work?

ConorMacBride:
When we add the "always compare" option to see how the images differ to the baseline, we'll need `-s` (when the figure tests all pass) to see the URL.

Yeah, there is probably a pytest function for showing messages. It's currently just a print().

nabobalis:
To see the url on circleci?

ConorMacBride:
Yes. However, just remove `-s`, as I'll hopefully find a better way to show the URL.

ConorMacBride:
The URL is reported by `pytest-mpl`. By default, it writes its to a random temporary directory.

nabobalis:
At least in our case, circleci reports the URL and we can access it directly from the artefacts page. 
So we can worry about losing that down the road.

ConorMacBride:
That's right! I forgot circleci was also showing the output URL. I was just thinking about `pytest-mpl` more generally. Never mind.

nabobalis:
Yeah, that will need addressing in pytest-mpl.

nabobalis:
I am going to add this to my ever growing list of manual backports.

nabobalis:
This probably doesn't even need backporting.

wtbarnes (approved):

nabobalis:
There might be a pyinstaller issues with scikit-image but we shall see with the next cronjob.

PR #5764, opened 2021-12-17
Stop silently assuming stuff in GenericMap

dstansby:
This PR removes what I think are the remaining 'silent' assumptions we make about missing metadata. I am very pro the deprecation warnings I've put in saying that we will raise errors or return `None` in the future. I feel like I must have tried to do this before, so perhaps there are arguments not to do this that I am forgetting?

Note that for at least `.scale` it is desirable for us to return `None` when only a CD_ij matrix is provided (when we support this), as we can't know what `.scale` is in that instance.

Depends on https://github.com/sunpy/sunpy/pull/5763

### DIFF
--- a/changelog/5763.bugfix.rst
+++ b/changelog/5763.bugfix.rst
@@ -1,0 +1,2 @@
+:func:`sunpy.map.make_fitswcs_header` now includes a PC_ij matrix in the returned
+header if no rotation is specified.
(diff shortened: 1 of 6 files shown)


dstansby:
So it turns out the FITS standard does have default values for a lot of these, some of which we are not applying correctly... (see https://fits.gsfc.nasa.gov/standard40/fits_standard40aa-le.pdf, page 31). I'll close this and open another PR in the future.

PR #5766, opened 2021-12-20
Fix CROTA > PC rotation matrix calculation

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/5762. The scale ratio was the wrong way round...


Cadair:
Oh my goodness.

sunpy/map/mapbase.py
```diff
@@ -79,33 +79,20 @@
...
 * If a CDi_j matrix is provided it is assumed that it can be converted to a
     PCi_j matrix and CDELT keywords as described in
-    `Greisen & Calabretta (2002) <https://doi.org/10.1051/0004-6361:20021327>`_
-
-* The 'standard' FITS keywords that are used by this class are the PCi_j
-    matrix and CDELT, along with the other keywords specified in the WCS
-    papers. All subclasses of this class must convert their header
-    information to this formalism. The CROTA to PCi_j conversion is done in
-    this class.
+    `Greisen & Calabretta (2002) <https://doi.org/10.1051/0004-6361:20021326>`_
```
Cadair, line 90: Should this part say we need CDELT? is that what the reference says to do?

sunpy/map/mapbase.py
```diff
@@ -1327,7 +1314,7 @@ def _rotation_matrix_from_crota(self, crota_key='CROTA2'):
         If the specified key isn't present in the metadata, a default rotation
         of 0deg is returned.
         """
-        lam = self.scale[0] / self.scale[1]
+        lam = self.scale[1] / self.scale[0]
```
Cadair, line 1317:
```suggestion
        # lam = lat / lon
        lam = self.scale[1] / self.scale[0]
```

Cadair (approved):
Thanks for finding and untangling this.

Cadair (changes requested):
Oh wait, hang on.

Do we mandate that CTYPE1 is lon and CTYPE2 is lat? if not (which I think is the case) don't we need to make the ratio dependent on CTYPE?

dstansby:
No, `.scale[0]` is always CDELT1 and `.scale[1]` is always CDELT2.

sunpy/map/mapbase.py
dstansby: This isn't always lat / lon, because whether axis 1/2 is lat/lon depends on the values of CTYPE{1,2}.

sunpy/map/mapbase.py
```diff
@@ -79,33 +79,15 @@
...
-the meta data. The first and most extensive assumption is that it is
-FITS-like WCS information as defined in the FITS WCS papers.
-
-Within this scope it also makes some other assumptions.
-
-* In the case of APIS convention headers where the CROTAi/j arguments are
-    provided it assumes that these can be converted to the standard PCi_j
-    notation using equations 32 in Thompson (2006).
-
-* If a CDi_j matrix is provided it is assumed that it can be converted to a
```
dstansby, line 92: This isn't an assumption we make, it's just the definition of the relation between CD_ij, CDELTi, and PC_ij, so I've dropped this bullet.

sunpy/map/mapbase.py
Cadair: (Carrying on from my blocking review comment here).

I don't think that's actually True. I think the rotation being in CROTA2 is actually only valid if latitude is the second axis (i.e. CTYPE2) see this relevant snippet:

2021-12-20-165325_grim

from [(Calbretta & Grisen 2002) (Paper II)[https://www.aanda.org/articles/aa/pdf/2002/45/aah3860.pdf].

wtbarnes:
> Oh wait, hang on.
> 
> Do we mandate that CTYPE1 is lon and CTYPE2 is lat? if not (which I think is the case) don't we need to make the ratio dependent on CTYPE?

Per section 6.1 of Calabretta and Greisen (2002), "we have used subscript 1 for the longitude axis, 2 for latitude." Does this mean that lambda is defined just as `CDELT2/CDELT1` or as "the CDELT corresponding to lat" / "the CDELT corresponding to lon". If the latter, I think we do need the `CTYPE` check as we have no guarantee that 2->lat and 1->lon.

sunpy/map/mapbase.py
dstansby: If you do the swaps 1 > 2 and 2 > 1 in eq 186, you still get lambda = CDELT2 / CDELT1, so I think what I've done is fine. This can be checked with the following code, that with this PR has expected axes limits as explained in the comments:
```python
import numpy as np
import sunpy.map

# Test non-zero crota and unequal CDELT{1,2}
n = 10
data = np.random.rand(n, n)
header = {
    'CRVAL1': 0,
    'CRVAL2': 0,
    'CRPIX1': 0.5,
    'CRPIX2': 0.5,
    'NAXIS1': n,
    'NAXIS2': n,
    'CUNIT1': 'arcsec',
    'CUNIT2': 'arcsec',
    'CTYPE1': 'HPLT-TAN',  # lat
    'CTYPE2': 'HPLN-TAN',  # lon
    'CDELT1': 1,  # total axis length = 10 (lat)
    'CDELT2': 2,  # total axis length = 20 (lon)
    'CROTA2': 90,
    # Rotating by 90 degrees makes the transformation lon > lat, lat > -lon
    # So new axis limits should be:
    # lon: 0 - 10
    # lat: 0 - -20
    'DATE-OBS': '2020-01-01 00:00:00'
}
sunpy.map.Map(data, header).plot()
ax = plt.gca()
# Swap axis labels
if header['CROTA2'] == 90:
    c0 = ax.coords[0]
    c1 = ax.coords[1]
    c0.set_ticks_position('l')
    c0.set_ticklabel_position('l')
    c0.set_axislabel_position('l')
    c1.set_ticks_position('b')
    c1.set_ticklabel_position('b')
    c1.set_axislabel_position('b')

plt.show()
```

dstansby:
> Per section 6.1 of Calabretta and Greisen (2002), "we have used subscript 1 for the longitude axis, 2 for latitude." Does this mean that lambda is defined just as `CDELT2/CDELT1` or as "the CDELT corresponding to lat" / "the CDELT corresponding to lon"? If the latter, I think we do need the `CTYPE` check as we have no guarantee that 2->lat and 1->lon.

As explained in https://github.com/sunpy/sunpy/pull/5766#discussion_r772551807, lambda = CDELT2 / CDELT1 is unrelated to which axes you label as latitude/longitude.

dstansby:
Another way to see that `lambda` is independent of CTYPE is the below flowchart from Paper II. `lambda` is involved only in the transformation from pixel to projection plane coordinates, which does not involve the CTYPE information.

Screenshot 2021-12-20 at 17 47 40

Cadair:
My comments now keep jumping around between the thread and here, but here goes.

---

To be clear upfront, I don't actually think this is an issue but I think we should all come to a shared understanding here because it's a pain.

I think that Fig 1 @dstansby copied in above doesn't include the conversion from the older AIPS convention as they are trying to describe the standard and not the older headers at that point, so I am not really sure it's all that relevant.

I suspect that it is λ = lat / lon, mainly because of the following two lines from section 6.1 in Paper 2.

> In the AIPS convention, CROTA i assigned to the **_latitude_** axis was used to define a bulk rotation of the image plane.

> where we have used subscript 1 for the longitude axis, 2 for latitude

That being said, I think that with my very rusty (and frankly always poor) linear algebra, it looks like equation 187 and 186 work perfectly happily irrespective of what CTYPE is. Where I think the issue is, is that if you wanted `CTYPE1` to be latitude, then you would have to use `CROTA1` and not `CROTA2` to describe the rotation. Making _ρ_ in equation 187 `CROTA1`. I don't then think this would actually change the calculation of λ but, we don't support it anyway!

Cadair (approved):

Cadair:
I would like someone else to check this over before we merge it. @ayshih if you could have a look that would be great.

PR #5770, opened 2021-12-30
Update warning ignores in setup.cfg

dstansby:
Should fix CI. Also remove a bunch of warnings that are fixed in released versions of other packages.

### DIFF
--- a/docs/code_ref/coordinates/rotatedsunframe.rst
+++ b/docs/code_ref/coordinates/rotatedsunframe.rst
@@ -3,6 +3,13 @@
 Differential rotation using coordinate frames
 *********************************************
 
+..
+  >>> # Due to small differences depending on different processors in numpy 1.22,
+  >>> # reduce precision at which results are printed. See https://github.com/matplotlib/matplotlib/pull/21634#issuecomment-1004200517
+  >>> # for the likely reason this is needed.
+  >>> import numpy as np
+  >>> np.set_printoptions(precision=6)
+
 Normally, coordinates refer to a point in inertial space (relative to the barycenter of the solar system).
 Transforming to a different observation time does not move the point itself, but if the coordinate frame origin and/or axis change with time, the coordinate representation is updated to account for this change.
 In solar physics an example of a frame that changes with time is the `~sunpy.coordinates.frames.HeliographicStonyhurst` (HGS) frame.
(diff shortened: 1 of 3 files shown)


nabobalis (approved):

dstansby:
I'm a bit confused as to why this fixes the core tests, but all the other tests are still failing on the warning I think I've filtered 😕

ConorMacBride:
The warning now only seems to get raised during py37 tests. I think it might be getting raised before the actual tests are run, so possibly not handled. Maybe raised by a pytest plugin?

The warning is raised within astropy and that bug was fixed 4 months ago, so it may not have made it into a py37 release of astropy.

nabobalis:
Ignore it and move on?

ConorMacBride:
I moved the ignored warning to the top so it is registered before any astropy warning classes are imported by pytest. So this should work with the py37 test now.

dstansby:
Thanks! Looks like there's now a failed test due to a new numpy release 🤦 . Seems to just be changes in the precision that floats in a doctest are printed. I can't repro locally, but have copied the changes from the CI.

ConorMacBride:
Looks like the py39 and py37 tests are producing different numpy outputs which are not equivalent (due to https://github.com/astropy/pytest-doctestplus/issues/92). I wonder if that issue is worth working on for this or just replace the numbers with ellipsis?

setup.cfg
```diff
@@ -133,30 +133,14 @@ filterwarnings =
...
-    ignore:The loop argument is deprecated
+    # Fixed in https://github.com/h5netcdf/h5netcdf/pull/121
+    ignore:distutils Version classes are deprecated
     # This is due to dependencies building with a numpy version different from
     # the local installed numpy version, but should be fine
     # See https://github.com/numpy/numpy/issues/15748#issuecomment-598584838
     ignore:numpy.ufunc size changed:RuntimeWarning
     ignore:numpy.ndarray size changed:RuntimeWarning
-    # See https://github.com/mvantellingen/python-zeep/issues/956
-    ignore:defusedxml.lxml is no longer supported:DeprecationWarning
```
ConorMacBride, line 148: This is still needed for `py37-oldestdeps`.

setup.cfg
```diff
@@ -133,30 +133,14 @@ filterwarnings =
...
-    # See https://github.com/spacetelescope/asdf/issues/789
-    ignore:direct construction of AsdfSchemaFile has been deprecated
-    ignore:direct construction of AsdfSchemaItem has been deprecated
-    ignore:direct construction of AsdfSchemaExampleItem has been deprecated
-    # See https://github.com/scikit-image/scikit-image/issues/4848
-    ignore:Converting `np.inexact` or `np.floating` to a dtype is deprecated
     # See https://github.com/astropy/astropy/pull/10570
-    ignore:leap-second auto-update failed
-    # See https://github.com/astropy/extension-helpers/issues/23
     ignore:Distutils was imported before Setuptools
```
ConorMacBride, line 144: I think the wrong comment was removed here?

docs/code_ref/coordinates/rotatedsunframe.rst
```diff
@@ -201,7 +201,7 @@ We specify a range of durations from -5 days to +5 days, stepping at 1-day incre
 
 Let's convert to the base coordinate frame to reveal the motion of the active region over time::
 
-  >>> ar.transform_to(ar.base)
+  >>> ar.transform_to(ar.base)  # doctest: +FLOAT_CMP
```
nabobalis, line 204: I thought we have this set in setup.cfg?

dstansby: Thanks for reminding - this change didn't do anything anyway, because the differences between different CI builds are greater than floating point precision.

dstansby:
The best option I could think of was reducing the precision in the doctest that was failing. I've added an explanatory comment inline about this. Hopefully this should fix the test, and is at least a short term fix for the CI.

ConorMacBride:
I think the adjustment needs to be reset at the end of the file: https://github.com/astropy/pytest-doctestplus/issues/92#issuecomment-567765010

PR #5776, opened 2022-01-03
Continue with build if sunpy.io.ana fails to build

dstansby:
Since ANA support is far from a big part of sunpy, I think it makes sense to just carry on installing if for some reason the `sunpy.io.ana` compilation fails. I'm not sure how to test this...

### DIFF
--- a/sunpy/io/setup_package.py
+++ b/sunpy/io/setup_package.py
@@ -8,10 +8,12 @@ from setuptools import Extension
 
 
 def get_extensions():
+    exts = []
 
     if get_compiler() == 'msvc':
-        return list()
-    else:
+        return exts
+
+    try:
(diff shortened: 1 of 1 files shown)


nabobalis:
> Since ANA support is far from a big part of sunpy, I think it makes sense to just carry on installing if for some reason the `sunpy.io.ana` compilation fails. I'm not sure how to test this...

We add a build where we don't add a compiler to the env and see if it errors? 
Might be easier to don circle than azure.

Cadair:
I don't think this actually encapsulates the compilation step? My understanding is that this function is called by extension-helpers to collect the `Extension` classes for all subpackages etc in the package and then they are all compiled together during the build step.

Issue #5783, opened 2022-01-04
`.superpixel()` and `.resample()` do not correctly modify metadata for maps with non-zero rotation


ayshih:
`.superpixel()` and `.resample()` modify the data in pixel space, and they correspondingly modify the coordinate metadata as if the pixel axes are aligned with the coordinate axes.  If a map has non-zero rotation, this does not return the correct result unless the resampling is the same in both directions (which, admittedly, is the typical use case).

Here's an example:
```python
import matplotlib.pyplot as plt
import astropy.units as u
import sunpy.map
from sunpy.data.sample import AIA_171_IMAGE

rot = sunpy.map.Map(AIA_171_IMAGE).rotate(30*u.deg)
rotsp = rot.superpixel([2, 1]*u.pix)

fig = plt.figure(figsize=(10, 4))

fig.add_subplot(121, projection=rot)
rot.plot(clip_interval=(1, 99.5)*u.percent, title='Original')

fig.add_subplot(122, projection=rotsp)
# Using aspect to visually correct for the rectangular pixels
rotsp.plot(clip_interval=(1, 99.5)*u.percent, title='After superpixel to rectangular pixels', aspect=0.5)

plt.show()
```
![superpixel bug](https://user-images.githubusercontent.com/991759/148115511-afd82d85-d6b0-411a-87e8-f2eddc53eb2f.png)

PR #5786, opened 2022-01-04
Fix superpixel() metadata computations

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/5783.

If using the `PC_ij` or `CD_ij` formalisms to describe the transformation, the scaling must be applied to the PC/CD matrices, *not* the CDELT values. This is because the PC/CD matrices multiply pixel coordinates before the CDELT scaling is applied. If the new `superpixel()` dimensions are square (e.g. `[2, 2]`), the scalar factor can be taken through to the other side of the PC/CD matrix multiplication which is why we didn't notice this, but for non-square matrices this is not the case.

### DIFF
--- a/changelog/5786.bugfix.rst
+++ b/changelog/5786.bugfix.rst
@@ -1,5 +1,5 @@
 Maps created from :meth:`~sunpy.map.GenericMap.resample` and
-:meth:`~sunpy.map.GenericMap.superpixel` have bene fixed in the case where
+:meth:`~sunpy.map.GenericMap.superpixel` have been fixed in the case where
 the resampling was not square, and the PCi_j matrix (often a rotation matrix)
 was not a multiple of the identity matrix. When the PCi_j or CDi_j formalisms
 are used in the metadata these are now correctly modified, and the CDELT values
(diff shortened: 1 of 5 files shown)


changelog/5786.bugfix.rst
```diff
@@ -0,0 +1,6 @@
+Maps created from :meth:`~sunpy.map.GenericMap.resample` and
+:meth:`~sunpy.map.GenericMap.superpixel` have bene fixed in the case where
```
nabobalis, line 2:
```suggestion
:meth:`~sunpy.map.GenericMap.superpixel` have been fixed in the case where
```

wtbarnes (approved):

dstansby:
Since this is a bugfix with one approval I'm going to merge.

ayshih:
None of us appreciated at the time that these changes would result in non-intuitive behaviors (#6570).  When using the PCij formalism, one should put the scaling factors in the CDELTi values, even for non-square pixels.  What was missing was that the off-diagonal elements of the PCij matrix need to be modified.

PR #5788, opened 2022-01-05
Use `--mpl-results-always`

ConorMacBride:
### Description
Uses the new `--mpl-results-always` option of pytest-mpl so all tests (including passing tests) are shown in the HTML output during figure tests.

The pytest-mpl git repository is installed as the new option is not in a release yet.

### DIFF
--- a/docs/code_ref/known_issues.rst
+++ b/docs/code_ref/known_issues.rst
@@ -22,3 +22,15 @@ There is not anything we can do to fix this on our side.
 `The specific section of the SDO users guide explaining this <https://www.lmsal.com/sdodocs/doc/dcur/SDOD0060.zip/zip/entry/sdoguidese4.html#x9-240004.2.4>`__.
 
 `Reference issue <https://github.com/sunpy/sunpy/issues/5447>`__.
+
+``scikit-image`` versions used to rotate/transform maps
+=======================================================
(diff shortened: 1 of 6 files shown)


ConorMacBride:
The Matplotlib comparison code doesn't save the diff image if the diff is zero, so this is expected. The URL of the diff image is still included in the HTML though. I have a fix for that in a PR I'll open soon.

Cadair:
Ok, I suspected that was the case, all good then. If we can merge #5789 I can re-test the comparison status here again.

Cadair:
ok, giles works again and the pytest-mpl patch seems to work fine, so I will make a release of pytest-mpl when I get 5.

dstansby:
What's the status here? Is it still waiting on something else upstream?

ConorMacBride:
Yes, waiting for the interactive HTML to be merged. Better to not merge this yet as in the current HTML report the tests are sorted by the order collected by pytest, so failing tests can be buried below passing tests.

ConorMacBride:
With the Scipy deprecation warning during test collection the run fails (as expected) and the report is empty (as expected): https://74589-2165383-gh.circle-artifacts.com/0/.tmp/py38-figure/figure_test_images/fig_comparison.html

ConorMacBride:
@Cadair rc is working here

PR #5793, opened 2022-01-05
update rtd config to remove legacy or deprecated options

nabobalis:
Need to wait for RTD to build to check.


Cadair (approved):

dstansby (changes requested):
It looks like the `docs` requirements need updating to include `matplotlib`

nabobalis:
> It looks like the `docs` requirements need updating to include `matplotlib`

Done.

dstansby:
Then we should probably update docs, instead of leaving it as dev?

nabobalis:
> Then we should probably update the docs requirements, instead of leaving it as dev?

Will do.

PR #5806, opened 2022-01-06
Parse HEK time columns

dstansby:
From https://www.lmsal.com/hek/VOEvent_Spec.html there are only three columns that are time-like. This PR parses each of those columns into `astropy.time.Time` if it is present in the result.

Also added a very basic caching mechanism to avoid running the same HEK query multiple times. 

Fixes partially https://github.com/sunpy/sunpy/issues/5006.

### DIFF
--- a/changelog/5806.feature.rst
+++ b/changelog/5806.feature.rst
@@ -1,0 +1,3 @@
+The 'event_endtime', 'event_starttime' and 'event_peaktime' columns in a HEK
+query are now returned as `~astropy.time.Time` objects. Previously they were
+timestamp strings.
--- a/docs/whatsnew/4.0.rst
+++ b/docs/whatsnew/4.0.rst
@@ -23,6 +23,11 @@ We have bumped the minimum version of several packages we depend on; these are t
 
 - python >= 3.8
 
+Improved return types of HEK queries
+====================================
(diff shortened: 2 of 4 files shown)


Cadair:
Don't suppose there is a machine readable version of that spec we could use to automatically do this? ;)

dstansby:
Yeah, I just realised we already played this game... this time round I did check the HEK website and associated docs, and https://www.lmsal.com/hek/VOEvent_Spec.html seems to be the best we have. So no, no machine readable spec 😢 .

Happy to close this if we want to encourage HEK to make a readable spec but:
- I think that might take a while...
- This is quite a nice improvement
- I'm under the impression that the current spec as written on the above web page is stable

Cadair:
I wonder if @nabobalis could get us any inside information on this which might make it clearer if this is the best way forward or not.

nabobalis:
> I wonder if @nabobalis could get us any inside information on this which might make it clearer if this is the best way forward or not.

I will ask around.

dstansby:
To sum up conversations on element (@nabobalis, correct me if I'm wrong), https://www.lmsal.com/hek/VOEvent_Spec.html is all there is in terms of a spec, and there is no information on which fields are time-like. So the best approach to take for now is manually parsing them, and we should go ahead with this PR.

nabobalis (approved):

Issue #4715, opened 2020-12-09
Automate Whats New creation on Pull Requests


Cadair:
We should have PRs include their whats new entries, would be good to have a system like towncrier, where they are added as fragments to a dir per release. Probably easiest to do as a custom sphinx extensions including all files in a dir.

nabobalis:
Additionally, a bot check with label.

godslayer201:
I would like to work on it @Cadair @nabobalis , may I please proceed?

kakirastern:
Welcome @godslayer201 to your 1st SunPy 🌞 contribution

godslayer201:
Thank you @kakirastern :) !

kakirastern:
Welcome 😀

nabobalis:
Hi @godslayer201, welcome aboard! 

We strongly suggest that new people do not work on these type of infrastructure issues and pick up a "Good First Issue" but at the end of the day the choice is yours.

PR #5811, opened 2022-01-11
Add GitHub Actions Changelog Bot

### LINKED ISSUE #5415: Make a bot to automatically rename changelog files
It's a bit tedious to write some code, commit, open a PR, and then have to go back to rename the changelog entries to match the PR number after the PR has been opened. Instead it would be super nice if we had a bot that automatically renamed newly added changelog entries of the form `xxxx.breaking.rst` etc. to the correct PR number once the PR had been opened.

ConorMacBride:
### Description
This is a new GitHub Action that commits changelog entries that are in the PR message. This still needs a lot of testing before merging.

### Changelog [feature]
```
Added a new GitHub Action for adding changelog entries.

The bot runs every time the PR message is updated.
```

Fixes #5415

### DIFF
--- a/tools/write_changelog_entry.py
+++ b/tools/write_changelog_entry.py
@@ -1,6 +1,7 @@
 import os
 import re
 import glob
+import subprocess
 
 
 def inc(obj, key):
@@ -11,12 +12,13 @@ def inc(obj, key):
         obj[key] = 1
 
 
-# Load PR number and comment
+# Load PR number, comment and bot name
 try:
     pr = os.environ['PR_NUMBER']
     body = os.environ['PR_BODY']
+    bot = os.environ['BOT_AUTHOR']
 except KeyError:
-    raise ValueError('PR_NUMBER and PR_BODY env vars must be set.')
+    raise ValueError('PR_NUMBER, PR_BODY and BOT_AUTHOR env vars must be set.')
 assert str(int(pr)) == pr
 body = body.replace('\\r', '').replace('\\n', '\n')
 
@@ -25,6 +27,20 @@ body = body.replace('\\r', '').replace('\\n', '\n')
 path = os.path.join(os.getenv('CHANGELOG_DIR', 'changelog'))
 if not os.path.exists(path):
     raise ValueError(f"CHANGELOG_DIR path '{path}' must exist.")
+
+# Check if bot has control of the changelogs
+existing = glob.glob(os.path.join(path, f"{pr}.*.rst"))
+for entry in existing:
+    author = subprocess.check_output(["git", "log", "-s", "-n1", "--pretty=%an <%ae>", entry], text=True).strip()
+    if author != bot:
+        print(f"{bot!r} is not modifying any changelogs because")
+        print(f"{entry!r} was last modified by {author!r}.")
+        exit(0)
+
+# Remove existing
+for entry in existing:
+    os.remove(entry)
+    print(f"Deleted existing changelog: {entry}")
 
 # Search for changelog entries
 #     markdown heading of "Changelog [category]"
@@ -54,12 +70,9 @@ for category, text in results:
     if len(text) > 0:  # ignore empty messages
         entries += [(filename, text)]
 
-for existing in glob.glob(os.path.join(path, f"{pr}.*.rst")):
-    os.remove(existing)
-    print(f"Deleted exiting changelog: {existing}")
-
 if len(entries) == 0:
-    raise ValueError("No new changelog entries found.")
+    print("No new changelog entries found.")
+    exit(0)
 
 for filename, text in entries:
     full_path = os.path.join(path, filename)
--- a/tools/write_changelog_entry.py
+++ b/tools/write_changelog_entry.py
@@ -1,6 +1,6 @@
-import glob
 import os
 import re
+import glob
 
 
 def inc(obj, key):
--- a/tools/write_changelog_entry.py
+++ b/tools/write_changelog_entry.py
@@ -1,0 +1,68 @@
+import glob
+import os
+import re
+
+
+def inc(obj, key):
+    """Increment a dictionary[key] value."""
+    try:
+        obj[key] += 1
+    except KeyError:
+        obj[key] = 1
+
+
+# Load PR number and comment
+try:
+    pr = os.environ['PR_NUMBER']
+    body = os.environ['PR_BODY']
+except KeyError:
+    raise ValueError('PR_NUMBER and PR_BODY env vars must be set.')
+assert str(int(pr)) == pr
+body = body.replace('\\r', '').replace('\\n', '\n')
+
+# Get changelog directory
+# TODO: get directory from pyproject.toml
+path = os.path.join(os.getenv('CHANGELOG_DIR', 'changelog'))
+if not os.path.exists(path):
+    raise ValueError(f"CHANGELOG_DIR path '{path}' must exist.")
+
+# Search for changelog entries
+#     markdown heading of "Changelog [category]"
+#     followed by a message in the next ``` code block.
+body = re.sub("(<!--.*?-->)", "", body, flags=re.DOTALL)
+regex = r"[#]+[ ]*Changelog[ ]*\[([a-zA-Z]+)\]*[^`]*```([^`]*)```"
+results = re.findall(regex, body)
+
+# Count entries per category
+count = {}
+for category, _ in results:
+    inc(count, category)
+
+# Generates (filename, message) pairs
+entries = []  # list of changelogs
+n = {}  # counter for each category
+for category, text in results:
+
+    if count[category] > 1:  # include filename [.COUNTER]
+        inc(n, category)
+        counter = f'.{n[category]}'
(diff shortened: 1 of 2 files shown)


ConorMacBride:
It's working except for the following error when trying to push the change to the PR branch. It may not be possible to get this working because if it was surely anyone could open a PR with a new GitHub Action to push whatever they want to `sunpy/sunpy`. Might need to create this as a GitHub App instead.

```
remote: Permission to ConorMacBride/sunpy.git denied to github-actions[bot].
42
fatal: unable to access 'https://github.com/ConorMacBride/sunpy/': The requested URL returned error: 403
```

ConorMacBride:
This should probably be added to https://github.com/OpenAstronomy/baldrick instead as an optional plugin. Should be able to implement #4715 in the same plugin also.

nabobalis:
> There already is a PR! [OpenAstronomy/baldrick#95](https://github.com/OpenAstronomy/baldrick/pull/95)

Does this mean we should close this PR?

ConorMacBride:
They both work in totally different ways and the other PR is a work in progress (and 2 years old). It's worth considering/combining both when expanding the packaging infrastructure.

Cadair:
Wanna merge this and test it here, or you wanna test it on a different repo?

ConorMacBride:
If you like. Are you generally okay with the implementation? Possible downsides: It commits to the PR branch every time a changelog is added, removed or updated, possibly resulting in multiple commits if there are lots of changes. It will also delete any manually committed changelogs for the PR number (unless the no changelog needed tag is added). And it has the issue with people needing to pull the changes back to their local branch.

ConorMacBride:
We could instead create an "Update Changelog" label which, when added, triggers this workflow. (And the workflow would also remove the label, so it can be added again later.) The current behaviour would be very bad for existing PRs as they would get their changelogs deleted when we add labels such as no backport.

tools/write_changelog_entry.py
```diff
@@ -0,0 +1,68 @@
...
+# TODO: get directory from pyproject.toml
+path = os.path.join(os.getenv('CHANGELOG_DIR', 'changelog'))
+if not os.path.exists(path):
+    raise ValueError(f"CHANGELOG_DIR path '{path}' must exist.")
+
+# Search for changelog entries
+#     markdown heading of "Changelog [category]"
+#     followed by a message in the next ``` code block.
+body = re.sub("(<!--.*?-->)", "", body, flags=re.DOTALL)
+regex = r"[#]+[ ]*Changelog[ ]*\[([a-zA-Z]+)\]*[^`]*```([^`]*)```"
```
Cadair, line 33: If we don't find anything in body of the PR can we terminate the action from here?

tools/write_changelog_entry.py
```diff
@@ -0,0 +1,68 @@
...
+        counter = f'.{n[category]}'
+    else:
+        counter = ''
+
+    filename = f"{pr}.{category}{counter}.rst"
+    text = text.strip()
+    if len(text) > 0:  # ignore empty messages
+        entries += [(filename, text)]
+
+for existing in glob.glob(os.path.join(path, f"{pr}.*.rst")):
```
Cadair, line 57: I am trying to think of a way we can detect that the action added the file rather than just nuking all changelogs.

Could we use git to pull the commit message of the last commit editing the file and see if we have a unique string in that message? That means that if the last commit wasn't the changelog bot then we don't delete it and instead we error?

tools/write_changelog_entry.py
```diff
@@ -0,0 +1,68 @@
...
+    os.remove(existing)
+    print(f"Deleted exiting changelog: {existing}")
+
+if len(entries) == 0:
+    raise ValueError("No new changelog entries found.")
```
Cadair, line 62: If we did this above the deletion then we wouldn't remove entries in PRs without the magic comment?

Cadair:
The other option is rather than using a magic heading in the top comment we could require you to @ a bot in a comment to make a change to the log entry? i.e. 

```
@SunPyBot add changelog bugfix

adlkjaskljdklajklj
```

tools/write_changelog_entry.py
ConorMacBride: Currently, if you remove the a Changelog from the main PR comment it deletes that Changelog from the PR branch. If we set it to exit here it would make deleting changelogs inconsistent, I.e: if you have two Changelog entries (feature and bugfix) and remove the feature Changelog from the comment it will be deleted in the PR branch, however, if you then remove the bugfix Changelog (so no regex match) it won't be deleted from the PR branch.

Cadair: Hmmm, that does make sense, but I don't think we should be deleting files people are adding manually either. If we can't delete entries from the PR description because of that, then I think that's a price we would have to pay. Unless we can come up with some other idea.

tools/write_changelog_entry.py
ConorMacBride: Could run e.g. `git log -s -n1 --pretty='format:%an %ae%n' changelog/6032.feature.rst` on each file before deleting, and only delete if it's the SunPyBot?

ConorMacBride:
I like the magic heading in the top comment because it makes it easier to update existing entries (rather than copying and pasting) and edit all the changelogs in the one place. Maybe the bot comment is more intuitive though?

tools/write_changelog_entry.py
ConorMacBride: Actually, it would be much simpler (and with less surprises) if this workflow can only commit if *all* the existing changelog entries for the PR were last edited by the bot. (Because otherwise it's complicated to handle multiple changelogs of the same type where some are controlled by the bot and some are manually controlled.)

Cadair:
I like the idea of the top comment, but it's a stateful interface i.e. handling deletes and edits etc, when the nature of actions is stateless, it's kinda hard to graft the stateful UI onto the stateless actions. Where as responding to specific commands in a comment means that the UI is also stateless.

tools/write_changelog_entry.py
Cadair: That would be fine with me as well.

ConorMacBride:
That makes sense. The comment should work well then. For the API design:

```
@SunPyBot add changelog <changelog type>

<changelog entry>
```

If no existing files match `<pr number>.<changelog type>(.(0-9)+)?.rst`, a new file `<pr number>.<changelog type>.rst` containing `<changelog entry>` will be created.

If `<pr number>.<changelog type>.rst` already exists, it will be renamed `<pr number>.<changelog type>.1.rst`. The changelog will be created as a new `<pr number>.<changelog type>.<n>.rst` file containing `<changelog entry>` where `<n>` is the lowest integer that doesn't already exist as a filename.

---

```
@SunPyBot update changelog <changelog type> [<n>]

<changelog entry>
```

This will update the contents of `<pr number>.<changelog type>[.<n>].rst` to be `<changelog entry>`. If `<pr number>.<changelog type>[.<n>].rst` does not exist it will error. If multiple changelogs of `<changelog type>` exist, `<n>` must be specified.

---

```
@SunPyBot delete changelog <changelog type> [<n>]
```

This will delete the file `<pr number>.<changelog type>[.<n>].rst`. If `<pr number>.<changelog type>[.<n>].rst` does not exist it will error. If multiple changelogs of `<changelog type>` exist, `<n>` must be specified.

---

The workflow will react to the comment with :+1: on success and :-1: on error.

tools/write_changelog_entry.py
ConorMacBride: Which would you prefer? Bot command or top comment? Top comment would be easiest to implement (as that's what we have already), although happy to switch to bot command if you would prefer.

Cadair:
Sounds great!

ConorMacBride:
Need a way to ignore comments from people who do not have write access to the PR branch...

tools/write_changelog_entry.py
ConorMacBride: I can't find a reliable way to check if the comment author should have permission to push to the PR branch (as the workflow will always have permission). So, if you don't mind, I'll just keep it as updating the top comment so permissions can be enforced. I'll add the check mentioned above.

ConorMacBride:
The [git-auto-commit-action documentation](https://github.com/stefanzweifel/git-auto-commit-action#advanced-uses) says that a `pull_request_target` event will run on the fork so the fork needs to have GitHub Actions enabled for their workaround to work. However, this is not what the GitHub documentation says and for example [this workflow](https://github.com/angular/angular/blob/8b1a2f104864afd2aa7b7c938c01dcf1d973fd6f/.github/workflows/dev-infra.yml) on a PR from a fork does [run on the PR target repo](https://github.com/angular/angular/runs/5889490653?check_suite_focus=true#step:2:457) and not on the fork. The difference is, the version of the workflow that is run is the version in the `main` branch and not the PR branch.

How about we merge this so we can test it?

nabobalis (approved):

nabobalis:
> How about we merge this so we can test it?

I am pro this.

Issue #5813, opened 2022-01-12
citation links


AlexPietrow:
### Description

Hi, maybe it would be possible to post a link to the citing sunpy page on the git readme? 
It took me a while to find this page, so a link might make it easier for people. :) https://sunpy.org/about

nabobalis:
Will do!

PR #5814, opened 2022-01-12
Added citation info to readme. 

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/5813


ConorMacBride:
Do we want to add a link to the sidebar? https://github.blog/2021-08-19-enhanced-support-citations-github/

nabobalis:
Yeah I can add that.

nabobalis:
Found one.

nabobalis:
I shoved the bibtex citation into https://citation-js.github.io/cff-generator/ to create the file.

nabobalis:
The check_skip check is failing?

ConorMacBride:
Yeah, I'm taking a look. I think the default git fetch may not be fetching enough.

ConorMacBride:
Found the problem. GitHub now seem to be fetching with `--depth=1` by default. Should be able to override in the upstream template. (This should not be broken for PRs from forks as I run an extra git fetch for those already.)

dstansby (approved):

PR #5816, opened 2022-01-13
Make horizons test more reliable

dstansby:
This test has been flaky for a while on our remote-data CI run. This PR changes it to generate N times and make a single remote query to horizons, instead of generating lots of single times and making N queries to horizons.

### DIFF
--- a/sunpy/coordinates/tests/strategies.py
+++ b/sunpy/coordinates/tests/strategies.py
@@ -2,6 +2,7 @@
 Provide a set of hypothesis strategies for various coordinates-related tests.
 """
 import hypothesis.strategies as st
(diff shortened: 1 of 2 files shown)


ConorMacBride:
In case you didn't know (I only found this recently): Azure Pipelines has some nice tools for monitoring individual tests over time. See: https://dev.azure.com/sunpy/sunpy/_test/analytics?definitionId=4&contextType=build

I've been then filtering that page with branch: main, 3.0 and 3.1, and test run: py38-online

It would be good if there was a way it could notify us if a specific test hasn't passed in so many days.

dstansby:
Ooh very nice, I have not come across that before!

nabobalis (approved):

PR #5817, opened 2022-01-15
Warning when an off-disk helioprojective coordinate is converted from 2D to 3D

### LINKED ISSUE #5812: Add a helpful warning message when converting a 2D `Helioprojective` coordinate will return all NaNs
If a 2D `Helioprojective` coordinate is off-disk, the conversion from 2D to 3D that is required for any non-trivial coordinate transformation will return all NaNs, unless something like `Helioprojective.assume_spherical_screen()` is used.  This is a recurring source of confusion for our users, particularly because they also sometimes fail to realize that they are asking for a non-trivial coordinate transformation.

We can add a helpful warning message when `Helioprojective.make_3d()` is going to turn a coordinate array that is not all NaNs into a coordinate array that is all NaNs*.  It would inform the user that a non-trivial transformation may have been requested and that a user may wish to use `Helioprojective.assume_spherical_screen()`.

\* We would not emit the warning if only some of the elements of a coordinate array turn into NaNs, because this happens all the time in mass transformations (e.g., with `reproject`).

See #5598 for a somewhat related issue

thewindsofwinter:
This pull request warns a user when a conversion to Helioprojective results in a distance array with all NaNs. It uses a simple, one-line test resulting in a short warning to user (further recommending `assume_spherical_screen` as a potential fix)

Fixes #5812 

Additional Info:
 - [X] Linting: I linted the file using PEP8 standards. My small snippet conforms to it but the overall document doesn't (some lines are rather long). I haven't edited anything other than what I added, though. EDIT: pycodestyle gave me a bunch of E501s but didn't notice the E127, whoops 😅 fixed.
 - [ ] Testing: I was unable to test the code as tox kept throwing errors on my laptop. Looks relatively straightforward, though, so hopefully it will work (and others hopefully will have better luck testing if it doesn't.

Thoughts on including a checklist for contributers in pull request template? That way we can try to ensure code follows standards set forth in the guidelines. Something like [this](https://github.com/gitkraken/vscode-gitlens/blob/main/.github/PULL_REQUEST_TEMPLATE.md) but of course customized for sunpy. If needed I can create an issue, make quick edits to this template, and send in another PR after this.

Thanks!

### DIFF
--- a/sunpy/coordinates/frames.py
+++ b/sunpy/coordinates/frames.py
@@ -4,6 +4,7 @@ Common solar physics coordinate systems.
 This submodule implements various solar physics coordinate frames for use with
 the `astropy.coordinates` module.
 """
+import traceback
 from contextlib import contextmanager
 
 import numpy as np
@@ -587,7 +588,9 @@ class Helioprojective(SunPyBaseCoordinateFrame):
 
             d = np.fmin(d, dd) if self._spherical_screen['only_off_disk'] else dd
 
-        if np.all(np.isnan(d)) and np.any(np.isfinite(cos_alpha)):
+        stack_trace = traceback.format_stack()
+        # _draw_grid is the specific function called in WCSAxes
+        if not any(["_draw_grid" in string for string in stack_trace]) and np.all(np.isnan(d)) and np.any(np.isfinite(cos_alpha)):
             warn_user("The conversion of these 2D helioprojective coordinates to 3D is all NaNs "
                       "because off-disk coordinates need an additional assumption to be mapped to calculate distance from the observer. "
                       "Consider using the context manager `Helioprojective.assume_spherical_screen()`.")
--- a/sunpy/coordinates/frames.py
+++ b/sunpy/coordinates/frames.py
@@ -4,6 +4,7 @@ Common solar physics coordinate systems.
 This submodule implements various solar physics coordinate frames for use with
 the `astropy.coordinates` module.
 """
+import re
 import traceback
 from contextlib import contextmanager
 
@@ -588,12 +589,17 @@ class Helioprojective(SunPyBaseCoordinateFrame):
 
             d = np.fmin(d, dd) if self._spherical_screen['only_off_disk'] else dd
 
+        # This warning can be triggered in specific draw calls when plt.show() is called
+        # we can not easily prevent this, so we check the specific function is being called
+        # within the stack trace.
         stack_trace = traceback.format_stack()
-        # _draw_grid is the specific function called in WCSAxes
-        if not any(["_draw_grid" in string for string in stack_trace]) and np.all(np.isnan(d)) and np.any(np.isfinite(cos_alpha)):
+        matching_string = 'wcsaxes.*_draw_grid'
+        bypass = any([re.search(matching_string, string) for string in stack_trace])
+        if not bypass and np.all(np.isnan(d)) and np.any(np.isfinite(cos_alpha)):
             warn_user("The conversion of these 2D helioprojective coordinates to 3D is all NaNs "
-                      "because off-disk coordinates need an additional assumption to be mapped to calculate distance from the observer. "
-                      "Consider using the context manager `Helioprojective.assume_spherical_screen()`.")
+                      "because off-disk coordinates need an additional assumption to be mapped to "
+                      "calculate distance from the observer. Consider using the context manager "
+                      "`Helioprojective.assume_spherical_screen()`.")
 
         return self.realize_frame(SphericalRepresentation(lon=lon,
                                                           lat=lat,
--- a/sunpy/coordinates/frames.py
+++ b/sunpy/coordinates/frames.py
@@ -587,6 +587,13 @@ class Helioprojective(SunPyBaseCoordinateFrame):
 
             d = np.fmin(d, dd) if self._spherical_screen['only_off_disk'] else dd
 
+        # If we have all NaNs in d, warn user
+        if np.all(np.isnan(d)) and np.any(np.isfinite(cos_alpha)):
+            warn_user("The conversion of these 2D helioprojective coordinates to 3D is all NaNs "
+                      "because off-disk coordinates need an additional assumption to be mapped to "
+                      "3D. Consider using the context manager "
+                      "`Helioprojective.assume_spherical_screen()`.")
+
         return self.realize_frame(SphericalRepresentation(lon=lon,
                                                           lat=lat,
                                                           distance=d))
--- a/sunpy/coordinates/frames.py
+++ b/sunpy/coordinates/frames.py
@@ -587,12 +587,10 @@ class Helioprojective(SunPyBaseCoordinateFrame):
 
             d = np.fmin(d, dd) if self._spherical_screen['only_off_disk'] else dd
 
-        # If we have all NaNs in d, warn user
         if np.all(np.isnan(d)) and np.any(np.isfinite(cos_alpha)):
             warn_user("The conversion of these 2D helioprojective coordinates to 3D is all NaNs "
-                      "because off-disk coordinates need an additional assumption to be mapped to "
-                      "3D. Consider using the context manager "
-                      "`Helioprojective.assume_spherical_screen()`.")
+                      "because off-disk coordinates need an additional assumption to be mapped to calculate distance from the observer. "
+                      "Consider using the context manager `Helioprojective.assume_spherical_screen()`.")
 
         return self.realize_frame(SphericalRepresentation(lon=lon,
                                                           lat=lat,
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -7,6 +7,7 @@ import html
 import inspect
 import numbers
 import textwrap
+import warnings
 import itertools
 import webbrowser
 from tempfile import NamedTemporaryFile
@@ -46,6 +47,8 @@ from sunpy.util.decorators import (
     check_arithmetic_compatibility,
 )
 from sunpy.util.exceptions import warn_metadata, warn_user
+from sunpy.util.decorators import cached_property_based_on, check_arithmetic_compatibility
+from sunpy.util.exceptions import SunpyUserWarning, warn_metadata, warn_user
 from sunpy.util.functools import seconddispatch
 from sunpy.util.util import _figure_to_base64
 from sunpy.visualization import axis_labels_from_ctype, peek_show, wcsaxes_compat
@@ -2154,18 +2157,18 @@ class GenericMap(NDData):
         if not axes:
             axes = wcsaxes_compat.gca_wcs(self.wcs)
         is_wcsaxes = wcsaxes_compat.is_wcsaxes(axes)
-
         if is_wcsaxes:
-            # TODO: supply custom limb radius
-            return sunpy.visualization.limb.draw_limb(
-                axes, self.observer_coordinate, resolution=resolution,
-                rsun=self.rsun_meters, **kwargs)
+            with warnings.catch_warnings():
+                warnings.filterwarnings("ignore", category=SunpyUserWarning)
+                # TODO: supply custom limb radius
+                return sunpy.visualization.limb.draw_limb(
+                    axes, self.observer_coordinate, resolution=resolution,
+                    rsun=self.rsun_meters, **kwargs)
         else:
             warn_deprecated("Drawing the limb on a non-WCS Axes is deprecated, "
                             "and will be removed in sunpy 4.1.")
             # If not WCSAxes use a Circle
             c_kw.setdefault('radius', self.rsun_obs.value)
-
             circ = patches.Circle([0, 0], **c_kw)
             axes.add_artist(circ)
             return circ, None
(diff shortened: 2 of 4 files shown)


nabobalis:
> * [ ]  Testing: I was unable to test the code as tox kept throwing errors on my laptop. Looks relatively straightforward, though, so hopefully it will work (and others hopefully will have better luck testing if it doesn't.

What errors?

> Thoughts on including a checklist for contributers in pull request template? That way we can try to ensure code follows standards set forth in the guidelines. Something like [this](https://github.com/gitkraken/vscode-gitlens/blob/main/.github/PULL_REQUEST_TEMPLATE.md?rgh-link-date=2022-01-15T05%3A30%3A53Z) but of course customized for sunpy. If needed I can create an issue, make quick edits to this template, and send in another PR after this.

We can look into adding something similar, yes.

thewindsofwinter:
> > * [ ]   Testing: I was unable to test the code as tox kept throwing errors on my laptop. Looks relatively straightforward, though, so hopefully it will work (and others hopefully will have better luck testing if it doesn't.
> 
> What errors?

Import errors from the environment itself, and I'm too worried to edit it since this was something conda itself generated

  File "$HOME\.conda\envs\sunpy-dev\lib\ctypes\__init__.py", line 8, in <module>
    from _ctypes import Union, Structure, Array
ImportError: DLL load failed while importing _ctypes: The specified module could not be found.

nabobalis:
>  File "$HOME\.conda\envs\sunpy-dev\lib\ctypes\__init__.py", line 8, in <module>
>    from _ctypes import Union, Structure, Array
> ImportError: DLL load failed while importing _ctypes: The specified module could not be found.

I can't say I've ever seen this error but can you create a new Python environment or reinstall conda?

thewindsofwinter:
> > File "$HOME.conda\envs\sunpy-dev\lib\ctypes__init__.py", line 8, in 
> > from _ctypes import Union, Structure, Array
> > ImportError: DLL load failed while importing _ctypes: The specified module could not be found.
> 
> I can't say I've ever seen this error but can you create a new Python environment or reinstall conda?

I'll try again then!

sunpy/coordinates/frames.py
```diff
@@ -587,6 +587,15 @@ def make_3d(self):
...
             d = np.fmin(d, dd) if self._spherical_screen['only_off_disk'] else dd
 
+        # If we have all NaNs in d, warn user
+        if np.all(np.isnan(d)):
+            raise SunpyUserWarning("The coordinate array produced by this "
+                                   "method appears to be all NaN. This may "
+                                   "occur when a non-trivial coordinate "
+                                   "transformation is requested using "
+                                   "`Helioprojective.make_3d()`: try using "
+                                   "`assume_spherical_screen()` instead.")
```
ayshih, line 597:
```suggestion
            warn_user("The conversion of these 2D helioprojective coordinates to 3D is all NaNs "
                      "because off-disk coordinates need an additional assumption to be mapped to "
                      "3D. Consider using the context manager "
                      "`Helioprojective.assume_spherical_screen()`.")
```
You shouldn't use `raise` for warnings (which is a mistake I've made as well).  We have a helper function (`warn_user()`), which is imported from `sunpy.util.exceptions`.

sunpy/coordinates/frames.py
```diff
@@ -587,6 +587,15 @@ def make_3d(self):
 
             d = np.fmin(d, dd) if self._spherical_screen['only_off_disk'] else dd
 
+        # If we have all NaNs in d, warn user
+        if np.all(np.isnan(d)):
```
ayshih, line 591:
```suggestion
        if np.all(np.isnan(d)) and np.any(np.isfinite(cos_alpha)):
```
This added check makes sure that we are emitting the warning only when the input is finite, and not when the user has provided NaNs in the first place.

ayshih (changes requested):
And needs a unit test, of course

dstansby (changes requested):
It looks like there a couple of places in the tests where this warning is now (correctly!) being thrown, that need catching. Could you have a look at the test output, and add `pytest.warns(...)` blocks where the warnings are being thrown, to make the tests pass again?

ayshih:
Convenient: the unit test already exists!  As @dstansby said, you should add a `pytest.warns()` to the the test `test_hpc_distance_off_limb`.  For the test `test_skycoord_hpc`, it might make sense to instead add to the existing warning filtering in that test because only 1 of the 7 parameterized arguments actually emits the warning.

nabobalis:
Looks like a test is still failing?

thewindsofwinter:
Yes, sorry, I've been busy as of late. I'm going to fix this once I get some spare time (unless someone else does before then)

nabobalis:
> Yes, sorry, I've been busy as of late. I'm going to fix this once I get some spare time (unless someone else does before then)

That is no problem, real life is more important than PRs

sunpy/map/tests/test_plotting.py
```diff
@@ -79,28 +80,58 @@ def test_peek_aia171(aia171_test_map):
 
 @figure_test
 def test_peek_grid_aia171(aia171_test_map):
-    aia171_test_map.peek(draw_grid=True)
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message=(
+            'The conversion of these 2D helioprojective coordinates to '
+            '3D is all NaNs because off-disk coordinates need an additional assumption to be mapped to 3D. '
+            'Consider using the context manager `Helioprojective.assume_spherical_screen()'
+        ), category=SunpyUserWarning)
```
nabobalis, line 88: Locally if I use pytest.raises, the test error. 

I don't know why so I did this.

sunpy/coordinates/frames.py
```diff
@@ -587,6 +587,12 @@ def make_3d(self):
 
             d = np.fmin(d, dd) if self._spherical_screen['only_off_disk'] else dd
 
+        if np.all(np.isnan(d)) and np.any(np.isfinite(cos_alpha)):
+            warn_user("The conversion of these 2D helioprojective coordinates to 3D is all NaNs "
+                      "because off-disk coordinates need an additional assumption to be mapped to "
+                      "3D. Consider using the context manager "
```
Cadair, line 593: Maybe something like this would clarify this a little?

```suggestion
                      "because off-disk coordinates need an additional assumption to be mapped to calculate distance from the observer."
                      "Consider using the context manager "
```

sunpy/map/tests/test_plotting.py
```diff
@@ -79,28 +80,48 @@ def test_peek_aia171(aia171_test_map):
 
 @figure_test
 def test_peek_grid_aia171(aia171_test_map):
-    aia171_test_map.peek(draw_grid=True)
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message=(
+            'The conversion of these 2D helioprojective coordinates to '
+        ), category=SunpyUserWarning)
+        aia171_test_map.peek(draw_grid=True)
```
Cadair, line 87: I assume these are rasing this warning because wcsaxes is making multiple calls to the transform stack for each contour line and *some* of them are all NaN?

I am not a fan of the map plotting methods rasing this warning to the user, as there is nothing they can do to stop it. (I.e. the current behaviour is correct and using assume spherical screen would be incorrect). If everything is working as intended maybe we need to silence this warning inside our plotting methods?

nabobalis: I can't work out how to silence it.

ayshih: Hmm, this is an interesting pickle.  The reason that you can't squelch the warnings this way is because our logger overrides the normal `warnings` behavior such that any `SunpyUserWarning` gets directed to the logger instead of being handled by the normal filtering machinery.  I don't immediately know what the solution is...

nabobalis: I forgot completely/didn't know that we do that with our logger.

ayshih: I imagine that the brute-force solution should work: check if our logger is currently set to capture warnings, and if so, disable it temporarily before entering the `catch_warnings()` context.  You probably should create a convenience context manager for this; I can see this being useful in future areas as well.

That said, I have a suspicion that there may be a problem with some code.  This PR's code is supposed to emit the warning only if an entire line is all NaNs.  But why would any grid line be all NaNs?  By definition every line should all be on disk.  Points at the limb can sometimes be interpreted as off-disk due to numerical precision, but that can't be why an entire line would be all NaNs.  I'm missing something...

ayshih: Okay, I diagnosed this completely wrong: the whole logger business is a red herring.  The good news is that it is possible to filter out the warnings even when the logger is set to capture warnings.  The bad news is that it is highly nontrivial to filter out the warnings within SunPy calls.  That's because the warnings aren't generated by the `draw_grid()` call, which simply instantiates the grid overlay, but rather by the rendering call (e.g., `plt.show()`).  For non-`peek()` plotting, we don't control when rendering happens, so we can't wrap it in a context manager.  The only solution that I can think of is to monkey-patch `WCSAxes`...

ayshih: (Incidentally, the lines of all NaNs are the grid overlay trying to figure out the extent of the map in world coordinates, not any of the grid lines themselves.)

sunpy/coordinates/frames.py
```diff
@@ -587,6 +588,11 @@ def make_3d(self):
 
             d = np.fmin(d, dd) if self._spherical_screen['only_off_disk'] else dd
 
+        if not any(["draw_grid" in string for string in traceback.format_stack()]) and np.all(np.isnan(d)) and np.any(np.isfinite(cos_alpha)):
```
nabobalis, line 591: Well....

nabobalis: Nope, it doesn't.

sunpy/map/mapbase.py
```diff
@@ -2363,7 +2361,7 @@ def peek(self, draw_limb=False, draw_grid=False,
 
         if draw_limb:
             self.draw_limb(axes=axes)
-
+
```
nabobalis, line 2364:
```suggestion
```

sunpy/coordinates/frames.py
```diff
@@ -587,6 +588,13 @@ def make_3d(self):
 
             d = np.fmin(d, dd) if self._spherical_screen['only_off_disk'] else dd
 
+        stack_trace = traceback.format_stack()
+        # _draw_grid is the specific function called in WCSAxes
+        if not any(["_draw_grid" in string for string in stack_trace]) and np.all(np.isnan(d)) and np.any(np.isfinite(cos_alpha)):
```
Cadair, line 593: I haven't used `stack_trace` before but can we not put a fully qualified path here or something else mentioning `wcsaxes` in some way?

sunpy/map/mapbase.py
```diff
@@ -2362,7 +2360,6 @@ def peek(self, draw_limb=False, draw_grid=False,
 
         if draw_limb:
             self.draw_limb(axes=axes)
```
Cadair, line 2362:
```suggestion
            self.draw_limb(axes=axes)

```

sunpy/coordinates/frames.py
nabobalis: You mean check for the file name and etc?

Cadair: yeah, or class at least?

nabobalis: I can add part of the file path.

nabobalis: Done it, but it's probably worse now.

nabobalis: Nope, broke it.

sunpy/coordinates/frames.py
```diff
@@ -587,6 +589,13 @@ def make_3d(self):
 
             d = np.fmin(d, dd) if self._spherical_screen['only_off_disk'] else dd
 
+        stack_trace = traceback.format_stack()
+        matching_string = 'wcsaxes.*_draw_grid'
```
nabobalis, line 593: Since the longer string is `astropy/visualization/wcsaxes/coordinate_helpers.py", line 567, in _draw_grid`, to check for the method name and it being called in wcsaxes

sunpy/coordinates/frames.py
```diff
@@ -587,6 +589,13 @@ def make_3d(self):
...
+        if not any([re.search(matching_string, string) for string in stack_trace]) and np.all(np.isnan(d)) and np.any(np.isfinite(cos_alpha)):
```
Cadair, line 594: Why don't you make this line a lot shorter by doing the regex on the line above?

sunpy/coordinates/frames.py
```diff
@@ -587,6 +589,13 @@ def make_3d(self):
...
+            warn_user("The conversion of these 2D helioprojective coordinates to 3D is all NaNs "
+                      "because off-disk coordinates need an additional assumption to be mapped to calculate distance from the observer. "
```
Cadair, line 596: also line length

nabobalis: Fixed.

sunpy/coordinates/frames.py
nabobalis: Fixed.

Cadair (approved):

ayshih (approved):
LGTM

PR #5818, opened 2022-01-17
Fixed a warning that was always being raised as an error

ayshih:
Way back in #4724, I accidentally raised a warning instead of properly using `warnings.warn()`.  Since our CI tests always elevate warnings to errors, nobody noticed.

In principle this could be backported to earlier than 3.1, but then I couldn't use `warn_user()`, so the backport is not straightforward.


dstansby:
I've tagged to backport to 3.0, and am happy taking responsibility for doing that manually.

dstansby (approved):

PR #5827, opened 2022-01-21
Prevent JSOC tests from being run in parallel

dstansby:
Today's CI failure was
```python
E       drms.exceptions.DrmsExportError: User jsoc@cadair.com has 1 pending export requests (JSOC_20220121_387); please wait until at least one request has completed before submitting a new one. [status=7]

../../.tox/py38-online/lib/python3.8/site-packages/drms/client.py:185: DrmsExportError
```

This PR:
- Uses `pytest-xdist` by default, with the auto option (this was already being used on CI)
- Marks tests that download JSOC data as being in the same group. This ensures they are allocated the same worker and are run in serial, hopefully avoiding pending export errors.

I don't think we should backport a new test dependency, but am open to other opinions on that.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -83,6 +83,7 @@ tests =
   pytest-doctestplus>=0.5 # We require the newest version of doctest plus to use +IGNORE_WARNINGS
   pytest-mock
   pytest-mpl>=0.12 # First version to support our figure tests
+  pytest-xdist
 docs =
   astroquery
   jplephem
@@ -109,7 +110,7 @@ testpaths = "sunpy" "docs"
 norecursedirs = ".tox" "build" "docs[\/]_build" "docs[\/]generated" "*.egg-info" "examples" "sunpy[/\]_dev" ".jupyter" ".history" "tools" "sunpy[\/]extern" ".pyinstaller"
 doctest_plus = enabled
 doctest_optionflags = NORMALIZE_WHITESPACE FLOAT_CMP ELLIPSIS
-addopts = --doctest-rst --doctest-ignore-import-errors -p no:unraisableexception -p no:threadexception
+addopts = --dist no --doctest-rst --doctest-ignore-import-errors -p no:unraisableexception -p no:threadexception
 asdf_schema_tests_enabled = true
 asdf_schema_root = sunpy/io/special/asdf/schemas/
 mpl-results-path = figure_test_images
--- a/sunpy/net/jsoc/tests/test_jsoc.py
+++ b/sunpy/net/jsoc/tests/test_jsoc.py
@@ -15,6 +15,9 @@ import sunpy.net.attrs as a
 from sunpy.net.jsoc import JSOCClient, JSOCResponse
 from sunpy.util.exceptions import SunpyUserWarning
 
+# Ensure all JSOC tests are run on the same parallel worker
+pytestmark = pytest.mark.xdist_group(name="jsoc")
+
 
 @pytest.fixture
 def client():
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -72,7 +72,7 @@ stages:
         submodules: false
         coverage: codecov
         toxdeps: tox-pypi-filter
-        posargs: -n=4
+        posargs: -n auto --dist loadgroup
         libraries:
           apt:
             - libopenjp2-7
@@ -97,7 +97,7 @@ stages:
         submodules: false
         coverage: codecov
         toxdeps: tox-pypi-filter
-        posargs: -n=4
+        posargs: -n auto --dist loadgroup
         libraries:
           apt:
             - libopenjp2-7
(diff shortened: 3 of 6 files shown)


ConorMacBride (commented):
Makes sense to ensure that two requests are not made before another finishes. Should we also stagger the daily scheduled runs to reduce the chance of this happening across runs?

I'm not sure about making parallel runs the default. Output from a parallel run is (slightly) more difficult to follow and, I haven't tested it, but maybe `breakpoint()` won't work in that case. It should be possible to [dynamically include](https://docs.pytest.org/en/6.2.x/example/simple.html#dynamically-adding-command-line-options) `--dist loadgroup` in `conftest.py` if `pytest-xdist` is installed and `-n` is set.

We can add a helper function that only applies the `xdist_mark` group if xdist is actually installed. That way we can backport the groups without adding xdist as a dependency. Although, I think it makes sense to add `pytest-xdist` as a dependency, and for the same reasons, maybe we should make `tox` a dependency also?

nabobalis:
We can add pytest-xdist as a test dep going backwards.

dstansby:
If someone else wants to take on backporting this feel free. Note that this will involve checking if dependencies need to be updated for wheel building/testing and conda building. If something needs updating with conda please open an issue on the conda-forge repo so I can update it at the next release.

> Makes sense to ensure that two requests are not made before another finishes. Should we also stagger the daily scheduled runs to reduce the chance of this happening across runs?

👍 

> I'm not sure about making parallel runs the default. Output from a parallel run is (slightly) more difficult to follow and, I haven't tested it, but maybe `breakpoint()` won't work in that case. It should be possible to [dynamically include](https://docs.pytest.org/en/6.2.x/example/simple.html#dynamically-adding-command-line-options) `--dist loadgroup` in `conftest.py` if `pytest-xdist` is installed and `-n` is set.

👍 - parallel is now disabled by default. I took the easy route of manually putting --dist loadgroup in the pipelines configuration.

> We can add a helper function that only applies the `xdist_mark` group if xdist is actually installed. That way we can backport the groups without adding xdist as a dependency. Although, I think it makes sense to add `pytest-xdist` as a dependency, and for the same reasons, maybe we should make `tox` a dependency also?

I think this is too much of a hassle.

sunpy/net/jsoc/tests/test_jsoc.py
```diff
@@ -15,6 +15,9 @@
 from sunpy.net.jsoc import JSOCClient, JSOCResponse
 from sunpy.util.exceptions import SunpyUserWarning
 
+# Ensure all JSOC tests are run on the same parallel worker
+pytest.mark.xdist_group(name="jsoc")
```
ConorMacBride, line 19: See [pytest docs](https://docs.pytest.org/en/6.2.x/example/markers.html#marking-whole-classes-or-modules). Running `pytest --co -m "xdist_group"` shows this marks the whole module.
```suggestion
pytestmark = pytest.mark.xdist_group(name="jsoc")
```

ConorMacBride (changes requested):
Just one update for the module level mark.

When the figure tests were using `-n auto` they used 18 threads and a thread crashed, failing the tests. So we should probably wait for #5830 to be merged so we can test Azure with `-n auto` (or just change `auto` to `4`?). The Core Tests stage is still running with `-n=4` which is probably fine.

The conda recipe will need `pytest-xdist` due to the `--dist` option and the `xdist_group` mark. I'll open an issue there.

ConorMacBride (approved):

nabobalis (approved):

nabobalis:
> Makes sense to ensure that two requests are not made before another finishes. Should we also stagger the daily scheduled runs to reduce the chance of this happening across runs?

You want to open a PR changing this?

nabobalis:
Do we also want to change the core tests to use -n auto too?

ConorMacBride:
Yeah, I can open a PR for the scheduled runs. I can't see any difference in the times with 2 vs 4. Maybe just change all to auto for consistency?

nabobalis:
Works for me.

setup.cfg
```diff
@@ -109,7 +110,7 @@ testpaths = "sunpy" "docs"
 norecursedirs = ".tox" "build" "docs[\/]_build" "docs[\/]generated" "*.egg-info" "examples" "sunpy[/\]_dev" ".jupyter" ".history" "tools" "sunpy[\/]extern" ".pyinstaller"
 doctest_plus = enabled
 doctest_optionflags = NORMALIZE_WHITESPACE FLOAT_CMP ELLIPSIS
-addopts = --doctest-rst --doctest-ignore-import-errors -p no:unraisableexception -p no:threadexception
+addopts = --dist no --doctest-rst --doctest-ignore-import-errors -p no:unraisableexception -p no:threadexception
```
Cadair, line 113: Why did you add `--dist no` here? What's it achieving that isn't by default? Don't you have to specify `-n` to enable pytest-xdist?

Cadair: Why do we want to ensure pytest-xdist is installed? To be clear I don't care that it's here, I just want to understand why.

nabobalis: Not a clue. At least this way, it ensures its installed and listed as a test deps instead of only working via tox/CI.

Issue #5422, opened 2021-06-25
TimeSeries objects can not be created from data with leap seconds as causes a value error


samaloney:
### Description

TimeSeries object can not handle dates with leap seconds as during creation the astropy `Time` objects (which support leap seconds) are converted as the panda df dateformat  (doesn't support leap seconds) and results in a value error.

Original issue was from a mailing list user Alan Wang

### Expected vs Actual behavior

A TimeSeries object to be created without an error

### Steps to Reproduce

```python
from sunpy.timeseries import TimeSeries
from sunpy.net import Fido,  attrs as a

tstart = "2015-07-01 01:00"
tend = "2015-07-01 23:00"
result = Fido.search(a.Time(tstart, tend), a.Instrument("XRS"))
result_goes15 = Fido.search(a.Time(tstart, tend), a.Instrument("XRS"), a.goes.SatelliteNumber(15))
file_goes15 = Fido.fetch(result_goes15)
goes_15 = TimeSeries(file_goes15)
```

Stack Trace
```
ValueError                                Traceback (most recent call last)
<ipython-input-26-504ac605a035> in <module>
      7 result_goes15 = Fido.search(a.Time(tstart, tend), a.Instrument("XRS"), a.goes.SatelliteNumber(15))
      8 file_goes15 = Fido.fetch(result_goes15)
----> 9 goes_15 = TimeSeries(file_goes15)

~/Projects/sunpydev/sunpy/timeseries/timeseries_factory.py in __call__(self, silence_errors, *args, **kwargs)
    381         for filepath in filepaths:
    382             try:
--> 383                 new_ts = self._check_registered_widgets(filepath=filepath, **kwargs)
    384                 new_timeseries.append(new_ts)
    385             except (NoMatchError, MultipleMatchError, ValidationFunctionError):

~/Projects/sunpydev/sunpy/timeseries/timeseries_factory.py in _check_registered_widgets(self, **kwargs)
    504         units = kwargs.pop('units', None)
    505         if filepath:
--> 506             data, meta, units = WidgetType._parse_file(filepath)
    507
    508         # Now return a TimeSeries from the given file.

~/Projects/sunpydev/sunpy/timeseries/sources/goes.py in _parse_file(cls, filepath)
    153         """
    154         if sunpy.io.detect_filetype(filepath) == "hdf5":
--> 155             return cls._parse_netcdf(filepath)
    156         try:
    157             hdus = sunpy.io.read_file(filepath)

[...]

Issue #4622, opened 2020-11-05
Data access model for TimeSeries: data versus to_dataframe(). Why?


wtbarnes:
In the [`TimeSeries` docs](https://docs.sunpy.org/en/latest/guide/data_types/timeseries.html), we tell people to access the columns of a TS using `.data` and the appropriate key name (like a `dict`). When you actually do this, you get a fun warning,

```python-traceback
WARNING: SunpyUserWarning: Using .data to access the dataframe is discouraged; use .to_dataframe() instead. [sunpy.timeseries.timeseriesbase]
```

This is obviously contradictory. Additionally, I find the use of `to_dataframe()` quite cumbersome. Or really even `.data`. I guess this sort of follows from the `Map` model, but I feel that TS, unlike `Map`, is inherently multi-dimensional and so it would be useful to be able to access these multiple dimensions as attributes directly on the TS object (even if they are just aliases to the underlying columns on `data`).

If we do actually want to discourage this, I think we should put a deprecation warning on it and remove it after the appropriate period. If we have no plans to remove it, I think we should remove the existing warning.

I suppose the larger question is: **why are we even bothering with providing a separate data type if we are telling users they need to cast it to a whole other data type (a `pandas.DataFrame`) in order to pull data out of it???** I recognize that TS is primarily useful as a factory and for pulling and combining files into a single data structure, including the metadata handling.

#3954 is also likely relevant here.

I fully realize this is sort of an open-ended rant and I don't mean to be overly critical. I'm not a frequent-user of TS so maybe I'm missing something here.

Cadair:
The main reason to encourage the use of `to_dataframe` was to insulate users against possible changes of type for `.data`.

Direct access to columns or similar sounds like a potentially good idea.

dstansby:
I am going to put it out there that `TimeSeries` using a `DataFrame` to store the data underneath is past the point of no return.

[...]

Issue #3954, opened 2020-04-01
Should sunpy.timeseries.TimeSeries use astropy or pandas underneath?


dstansby:
This is an issue to discuss whether `sunpy.timeseries.TimeSeries` should use `pandas.DataFrame` or `astropy.timeseries.TimeSeries` underneath.

Some of my thoughts/opinions:
- If we move to `astropy.timeseries.TimeSeries`, should we just remove the SunPy version of a TimeSeries altogether, as a timeseries of data is not a solar physics specific concept?
- Personally I regularly use the underlying pandas methods (e.g. `tseries.data.dropna()`), which I would lose access to if this change was made. This is potentially a very large change, and would mean losing out on a lot of functionality that is in pandas but not astropy.
- `astropy.timeseries.TimeSeries` is not currently stable (according to the astropy docs: https://docs.astropy.org/en/stable/timeseries)

It would be good to get someone who knows more about why astropy Timeseries was developed to briefly list the advantages it has over a pandas DataFrame (@Cadair?).

samaloney:
What about NDCube? A time series is just 1xn cube or maybe a combination of 1xn cubes?

dstansby:
I can't find it at the moment, but it should be documented somewhere that we decided during the 2021 dev meeting to use astropy underneath.

dstansby:
We had more discsussions on this at the 2022 coordination meeting, and decided that astropy TimeSeries might not be the right answer. After a bit of back and forth on different options, we also decided this needed some proper thought and scoping, with requirements written up properly and a choice of data container made from that.

I have applied for a NumFocus small development grant to do the scoping work, here is the [submitted application](https://docs.google.com/document/d/1bS6mFd3XJ7cbHvgpPfdVmFLFErLOEYqeeN_ndeLQNY4/edit#). Should hear back in mid-October on whether the proposal was succesful.

hayesla:
this sounds great @dstansby - fingers crossed!

wtbarnes:
Given that this discussion is still ongoing, could someone please link the findings of the aforementioned SDG work?

wtbarnes:
As part of the aforementioned successful SDG, @dstansby wrote up a very nice assessment of the different possible data structures for backing `TimeSeries`: https://sunpy.org/posts/2023/2023-05-10-timeseries_sdg/. This should be a useful reference when deciding on the future of `TimeSeries`.

Cadair:
[...]

PR #5834, opened 2022-01-23
Refactor TimeSeries to use astropy TimeSeries underneath

dstansby:
This PR is the minimal set of changes to use astropy TimeSeries instead of pandas DataFRame under the hood of our `GenericTimeSeries`. For ease of review there are a few PRs I plan to build on top of this, but I will open them later. This PR should not change any user-facing functionality.

See https://github.com/sunpy/sunpy/issues/3954 for the discussion/decision to do this.

These changes are tested by all the existing `TimeSeries` tests, and I have added a couple of extra ones for a new deprecation warning and leapseconds.

Fixes https://github.com/sunpy/sunpy/issues/5422
Fixes https://github.com/sunpy/sunpy/issues/4622
Fixes https://github.com/sunpy/sunpy/issues/3954

TODO:
- [ ] Fix `to_array()` to not go via. pandas

### DIFF
--- a/changelog/5834.feature.rst
+++ b/changelog/5834.feature.rst
@@ -1,0 +1,1 @@
+`~sunpy.timeseries` now has support for times that fall within a leapsecond.
--- a/docs/whatsnew/4.0.rst
+++ b/docs/whatsnew/4.0.rst
@@ -93,7 +93,15 @@ As part of this the following API has been deprecated:
   :meth:`sunpy.timeseries.GenericTimeSeries.to_table` or
   :meth:`sunpy.timeseries.GenericTimeSeries.to_dataframe` instead.
 
-Improvements to map and image rotation
+Calling :meth:`sunpy.timeseries.GenericTimeSeries.concatenate` without
+specifying the new method argument as ``method='astropy'`` is also deprecated.
+The default value is ``method='pandas'``, which reproduces previous default
+behaviour, but will be removed in sunpy 4.1. The only difference between the
+two methods is that ``method='astropy'`` will not remove duplicate rows.
+
+This change allows sunpy 4.0 to be backward compatible with sunpy 3.1,
+whilst raising a warning that allows us to break this compatability in sunp
+4.1.
 
 Improved return types of HEK queries
 ====================================
--- a/docs/whatsnew/4.0.rst
+++ b/docs/whatsnew/4.0.rst
@@ -78,6 +78,22 @@ Also, there is now the option to rotate using `OpenCV <https://opencv.org>`__.
 The rotation function to use can be selected via the ``method`` keyword argument.
 New rotation functions beyond these three can be added using the new decorator :func:`~sunpy.image.transform.add_rotation_function`.
 
+Reworked internals of TimeSeries
+================================
+`~sunpy.timeseries.GenericTimeseries` has been re-worked to use astropy time
+handling underneath. This allows users to take full advantage of astropy time
+handling (e.g. different time systems, support for leapseconds) when using
+`~sunpy.timeseries.GenericTimeseries`.
+
+As part of this the following API has been deprecated:
+- `sunpy.timeseries.GenericTimeseries.index`. Use
+  `sunpy.timeseries.GenericTimeseries.times` instead.
+- `sunpy.timeseries.GenericTimeseries.data`. Use
+  :meth:`sunpy.timeseries.GenericTimeseries.to_table` or
+  :meth:`sunpy.timeseries.GenericTimeseries.to_dataframe` instead.
+
+Improvements to map and image rotation
+
 Improved return types of HEK queries
 ====================================
 The 'event_endtime', 'event_starttime' and 'event_peaktime' columns in a HEK
--- a/docs/whatsnew/4.0.rst
+++ b/docs/whatsnew/4.0.rst
@@ -80,17 +80,18 @@ New rotation functions beyond these three can be added using the new decorator :
 
 Reworked internals of TimeSeries
 ================================
-`~sunpy.timeseries.GenericTimeseries` has been re-worked to use astropy time
+`sunpy.timeseries.GenericTimeSeries` has been re-worked to use astropy time
 handling underneath. This allows users to take full advantage of astropy time
 handling (e.g. different time systems, support for leapseconds) when using
-`~sunpy.timeseries.GenericTimeseries`.
+`sunpy.timeseries.GenericTimeSeries`.
 
 As part of this the following API has been deprecated:
-- `sunpy.timeseries.GenericTimeseries.index`. Use
-  `sunpy.timeseries.GenericTimeseries.times` instead.
-- `sunpy.timeseries.GenericTimeseries.data`. Use
-  :meth:`sunpy.timeseries.GenericTimeseries.to_table` or
-  :meth:`sunpy.timeseries.GenericTimeseries.to_dataframe` instead.
+
+- `sunpy.timeseries.GenericTimeSeries.index`. Use
+  `sunpy.timeseries.GenericTimeSeries.times` instead.
+- `sunpy.timeseries.GenericTimeSeries.data`. Use
+  :meth:`sunpy.timeseries.GenericTimeSeries.to_table` or
+  :meth:`sunpy.timeseries.GenericTimeSeries.to_dataframe` instead.
 
 Improvements to map and image rotation
 
--- a/sunpy/timeseries/sources/eve.py
+++ b/sunpy/timeseries/sources/eve.py
@@ -5,11 +5,12 @@ from collections import OrderedDict
 
 import matplotlib.pyplot as plt
 import numpy as np
-from pandas import DataFrame, to_datetime
+from pandas import to_datetime
 from pandas.io.parsers import read_csv
 
 import astropy.units as u
 from astropy.time import TimeDelta
+from astropy.timeseries import TimeSeries
 
 import sunpy.io
 from sunpy.time import parse_time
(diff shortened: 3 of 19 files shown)


changelog/5834.feature.rst
```diff
@@ -0,0 +1 @@
+`~sunpy.timeseries` now has support for times that fall within a leapsecond.
```
Cadair, line 1: :clap: I love this

sunpy/timeseries/timeseriesbase.py
```diff
@@ -623,20 +643,8 @@ def to_table(self, **kwargs):
...
-        table.add_column(index_col, index=0)
-
-        # Add in units.
-        for key in self.units:
-            table[key].unit = self.units[key]
-
-        # Output the table
-        return table
+        # TODO: deprecate this
+        return Table(data=self._data, units=self.units)
```
Cadair, line 647: Can't we just return `.data` here? Timeseries is a subclass of Table?

dstansby: I think I tried that and some tests failed.

sunpy/timeseries/sources/eve.py
```diff
@@ -95,6 +96,7 @@ def plot(self, axes=None, **kwargs):
         formatter = mdates.ConciseDateFormatter(locator)
         axes[-1].xaxis.set_major_locator(locator)
         axes[-1].xaxis.set_major_formatter(formatter)
+        axes[-1].set_xlabel('')
```
nabobalis, line 99: What was it setting before?

sunpy/timeseries/timeseries_factory.py
```diff
@@ -272,7 +268,10 @@ def _sanitise_args(self, args):
                     meta.update(new_meta)
                 elif isinstance(data, np.ndarray):
                     # We have a numpy ndarray. We assume the first column is a dt index
-                    data = pd.DataFrame(data=data[:, 1:], index=Time(data[:, 0]))
+                    data = AstroTimeSeries(data=data[:, 1:], time=Time(data[:, 0]))
```
nabobalis, line 271: Codecov says this isn't hit?

sunpy/timeseries/timeseriesbase.py
```diff
@@ -92,6 +75,11 @@ def __init_subclass__(cls, **kwargs):
 
     # kwargs are not used here but are passed in for sources.
     def __init__(self, data, meta=None, units=None, **kwargs):
+        if isinstance(data, pd.DataFrame):
+            warn_deprecated('Passing a DataFrame to GenericTimeSeries is deprecated. '
+                            'Pass to sunpy.timeseries.TimeSeries instead.')
```
hayesla, line 80: maybe I'm missing something - should this be `pass a astropy.timeseries.TimeSeries`?

sunpy/timeseries/timeseriesbase.py
```diff
@@ -152,29 +140,31 @@ def columns(self):
         """
         A list of all the names of the columns in the data.
         """
-        return list(self._data.columns.values)
+        names = self._data.colnames
+        names.remove('time')
+        return names
 
     @property
     def index(self):
```
hayesla, line 148: more of a general comment for discussion - should we deprecate index? and have it as `ts.time` instead?

sunpy/timeseries/timeseriesbase.py
```diff
@@ -428,9 +442,13 @@ def concatenate(self, others, same_source=False, **kwargs):
         # Concatenate the metadata and data.
         kwargs["sort"] = kwargs.pop("sort", False)
         meta = self.meta.concatenate([series.meta for series in others])
+        # TODO: refactor this to avoid going via DataFrames
```
hayesla, line 445: just curious - by going via data frames does this do anything in regards to leap seconds? i.e. will it error if leap seconds are in data? or does it get rid of them?

hayesla (commented):
So far this looks great, and haven't managed to properly break anything yet! 

Something that could be an issue in terms of a somewhat breaking change is that now the `ts.index` returns an astropy time object rather than a `DatetimeIndex`. This doesnt play nicely with matplotlib. 

For example this now won't work with this PR:
```python
goes = ts.TimeSeries(GOES_XRS_TIMESERIES)
plt.plot(goes.index, goes.quantity("xrsa"))
```
you'll have to do

```python
goes = ts.TimeSeries(GOES_XRS_TIMESERIES)
plt.plot(goes.index.datetime, goes.quantity("xrsa"))
```
Building on this - would it make sense to follow from astropy timeseries and have a `.time` attribute (which basically would be `.index`), which would then just make more sense and could follow from astropy docs? Something to consider.

hayesla:
also just a comment - why do we still have a user warning    `"Using .data to access the dataframe is discouraged; use .to_dataframe() instead."`. Should this be ok now? like astropy timeseries allows for .data?

hayesla:
also I do remember we had issues with a file once that had leap seconds - does anyone remember what that was? it would be a good test case

sunpy/timeseries/timeseriesbase.py
dstansby: No, this is correct. I've deprecated `GenericTimeSeries(DataFrame)`, instead telling users to use the factory and do `TimeSeries(DataFrame)`.

sunpy/timeseries/timeseriesbase.py
dstansby: 👍 after reading your comment above I independently came to this conclusion

dstansby:
> also just a comment - why do we still have a user warning `"Using .data to access the dataframe is discouraged; use .to_dataframe() instead."`. Should this be ok now? like astropy timeseries allows for .data?

I don't think it should be okay, since we want to discourage users from casting to dataframes now. In fact, I will deprecate `.data` in this PR.

sunpy/timeseries/timeseriesbase.py
dstansby: The reason is that `self._data` doesn't have any units attached. I've deliberately done that to make this PR easier to review, but have another PR in the works to move the units on to self._data. At that point we can return ._data.

sunpy/timeseries/timeseriesbase.py
dstansby: It will error, but you couldn't even create a `TimeSeries` with leapseconds before this PR, so I don't think this is a deal breaker for this PR? I've opened https://github.com/sunpy/sunpy/issues/5873 to track testing all the methods with leapsecond data.

dstansby:
I think I've addressed all the comments - reviewers please mark as resolved if I have above, and re-review 😄

sunpy/timeseries/timeseriesbase.py
```diff
@@ -25,6 +27,7 @@
 u.add_enabled_units([det])
 
 TIME_FORMAT = config.get("general", "time_format")
+_builtin_int = int
```
nabobalis, line 30: What is the purpose of this?

nabobalis (approved):

nabobalis (changes requested):
Just needs a whatsnew entry for me.

nabobalis (approved):

nabobalis:
Could this get a rebase due to the file conflicts?

dstansby:
Yes, thanks for the nudge!

sunpy/timeseries/timeseriesbase.py
```diff
@@ -432,9 +451,13 @@ def concatenate(self, others, same_source=False, **kwargs):
         # Concatenate the metadata and data.
         kwargs["sort"] = kwargs.pop("sort", False)
         meta = self.meta.concatenate([series.meta for series in others])
+        # TODO: refactor this to avoid going via DataFrames
```
Cadair, line 454: Merging this with this hack in place isn't something I am particularly happy about. Not without some substantive justification anyway.

Cadair (changes requested):
Noticed one show stopper, will review more soon.

sunpy/timeseries/timeseriesbase.py
dstansby: Why not?

docs/whatsnew/4.0.rst
```diff
@@ -25,17 +25,18 @@ We have bumped the minimum version of several packages we depend on; these are t
 
 Reworked internals of TimeSeries
 ================================
-`~sunpy.timeseries.GenericTimeseries` has been re-worked to use astropy time
+``sunpy.timeseries.GenericTimeseries`` has been re-worked to use astropy time
```
dstansby, line 28: This linking should be fine, it just needs `Timeseries` > `TimeSeries`. @nabobalis can you fix this? I don't have time at the moment.

sunpy/timeseries/timeseriesbase.py
Cadair: Mainly because it destroys all the benefits of being backed by Astropy Time if you cycle out and back in again.

Cadair:
So that I understand the scope a little better here, can you explain what you are planning on doing in followups and what's in here? I see you have #5875 open, which feels pretty fundamental, but I worry that half-doing this (i.e. not all the methods work etc) is going to be a net-negative for the end users.

dstansby:
#5875 is a re-factor that shouldn't change any user-facing functionality, so I split it out for ease of PR review.

To be honest it's been so long since I looked at this that I can't remember what the plan for followups was... In my head (this should be documented in an issue...) I think the very long term goal should be replicating the API of `astropy.timeseries.TimeSeries`, with the metadata handling bolted on top (that's the only reason we're not just direction users to astropy right?)

docs/whatsnew/4.0.rst
nabobalis: Sure.

dstansby:
I've gone back down the rabbit hole, and this isn't going to be finished for 4.0.

dstansby:
As discussed at the coordination meeting 2022, due to recent changes in the data-container landscape (in particular NDCube 2.0, xarray updates) we're going to put this on hold pending a review of whether changing to `astropy.timeseries.TimeSeries` is the right way to go.

PR #5835, opened 2022-01-23
Raise exception if download errors during database fetch

### LINKED ISSUE #5826: Fixing flaky tests in `sunpy.database`
### ~Make errors from `Database.fetch` available?~
~I'm trying to investigate some flaky remote data tests. A lot of the tests rely on asserting the length of the database after fetching. Would it be useful (and make sense) if there was some way to access errors from the `Downloader` in a `Database` object? If we can access the error we should be able to print which servers are failing in the test log.  In the code below `paths._errors` should contain any errors.~

~https://github.com/sunpy/sunpy/blob/bf3a54f6efd104faad4f4bb7347ba542cb2f16a2/sunpy/database/database.py#L412-L414~

### Possible bug if only some downloads fail
Also, in the above code, could there ever be a situation where `paths` and `query_result` are not paired correctly by `zip`? If there is an error downloading a file, it will not appear in `paths`. I've only seen `paths=[]` which is fine, but could you get `len(paths)==1` and `len(query_result)==2`?

---
### Specific server that's not responding consistently
From multiple networks I'm finding that `sdo7.nascom.nasa.gov` is taking a long time to respond but eventually gives a 500 internal server error. The response provides an email address to report the error. Any point doing that? See example:

```sh
curl "https://sdo7.nascom.nasa.gov/cgi-bin/drms_export.cgi?series=aia__lev1;compress=rice;record=335_1123200039-1123200039"
```

ConorMacBride:
If there are download errors during a database fetch, metadata for expected downloads may not matched to downloaded files correctly. (Due to `zip()` with unequal lengths.) I can't see a good way to handle this, so raising an exception is probably the best approach.

Fixes #5826

A better fix would be for the `parfive.Results` object to include errors in the list rather than just the `errors` attribute, however this would probably cause problems elsewhere.

## Code to reproduce error
Notice how the first time it's run it gives (NSO, JSOC, JSOC). When it's run for a second time the NSO file, which was already downloaded via FTP, fails (possibly due to a bug in parfive). (I was planning to force it to fail by editing my `/etc/hosts` so this was handy 😆 ) However, it returns (NSO, NSO) instead of (JSOC, JSOC), as the first expected result was the NSO file but the first (and only) element of `paths` is the JSOC file. This PR will just raise a new `PartialFetchError` exception in such a case.

```python
from sunpy.database import Database
from sunpy.net import attrs

db = Database('sqlite:///', default_waveunit='Angstrom')

download_query = [
    attrs.Time('2012-08-05', '2012-08-05 00:01:05'),
    attrs.Instrument('HMI'),
    attrs.Physobs.los_magnetic_field
]

# This returns: (provider, url)
# NSO, ftp://solis.nso.edu/HMI/x73/201208/x9x73120805/x9x73120805t000000_map-err_dim-180_source-SDO-HMI_type-m-720s.fits.gz
# JSOC, http://netdrms01.nispdc.nso.edu/cgi-bin/netdrms/drms_export.cgi?series=hmi__M_45s;compress=rice;record=13739522-13739522

db.fetch(*download_query, path=".")

for entry in db:
    print(entry.provider, entry.path)

print(db)
```

### DIFF
--- a/sunpy/database/database.py
+++ b/sunpy/database/database.py
@@ -121,6 +121,23 @@ class TagAlreadyAssignedError(Exception):
         return errmsg.format(self.database_entry, self.tag_name)
 
 
+class PartialFetchError(Exception):
+    """This exception is raised if the number of files returned by the
+    downloader does not match the number expected.
+
+    """
+
+    def __init__(self, paths, expected):
+        self.paths = paths
+        self.expected = int(expected)
+
+    def __str__(self):
+        return (f"The downloader returned {len(self.paths)} file(s) "
+                f"but {self.expected} file(s) were expected.\n"
+                f"Successful downloads: {self.paths}\n"
+                f"Errors: {self.paths.errors}")
+
+
 def split_database(source_database, destination_database, *query_string):
     """
     Queries the source database with the query string, and moves the
@@ -410,6 +427,10 @@ class Database:
             self.remove(temp)
 
         paths = client.fetch(query_result, path)
+
+        # Unable to handle partial success
+        if len(paths) != len(query_result):
+            raise PartialFetchError(paths, len(query_result))
 
         for (path, block) in zip(paths, query_result):
             qr_entry = tables.DatabaseEntry._from_query_result_block(block)
--- a/sunpy/database/database.py
+++ b/sunpy/database/database.py
@@ -122,18 +122,18 @@ class TagAlreadyAssignedError(Exception):
 
 
 class PartialFetchError(Exception):
-    """This exception is raised if the number of files returned by the
+    """
+    This exception is raised if the number of files returned by the
     downloader does not match the number expected.
-
     """
 
-    def __init__(self, paths, expected):
+    def __init__(self, paths, n_expected):
         self.paths = paths
-        self.expected = int(expected)
+        self.n_expected = int(n_expected)
 
     def __str__(self):
         return (f"The downloader returned {len(self.paths)} file(s) "
-                f"but {self.expected} file(s) were expected.\n"
+                f"but {self.n_expected} file(s) were expected.\n"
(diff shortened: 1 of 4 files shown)


ConorMacBride:
If you think this is the best approach I can add a changelog entry. Don't think testing this would be straightforward though.

sunpy/database/database.py
```diff
@@ -121,6 +121,23 @@ def __str__(self):
         return errmsg.format(self.database_entry, self.tag_name)
 
 
+class PartialFetchError(Exception):
+    """This exception is raised if the number of files returned by the
+    downloader does not match the number expected.
+
+    """
```
dstansby, line 128:
```suggestion
    """
    This exception is raised if the number of files returned by the
    downloader does not match the number expected.
    """
```

sunpy/database/database.py
```diff
@@ -121,6 +121,23 @@ def __str__(self):
...
+
+    def __init__(self, paths, expected):
```
dstansby, line 130: `n_expected` might be a clearer variable name - I originally thought expected might be a list of expected paths or similar.

dstansby (approved):
👍 with two optional suggestions

dstansby:
Oh, and
- Needs a changelog entry
- pre-commit needs fixing

nabobalis (approved):

nabobalis:
> If you think this is the best approach I can add a changelog entry. Don't think testing this would be straightforward though.

Would mocking `paths = client.fetch(query_result, path)` to return a empty list not be enough?

ConorMacBride:
Yeah, I will add that.

nabobalis:
Online fail looks like a remote one off. Actually raised the new error from this PR.

PR #5844, opened 2022-01-25
Make codecov report on failing CI

dstansby:
This will let codecov report if a CI run fails, which can be helpful for checking coverage even if an unreleated remote data test fails.


Cadair (approved):

ConorMacBride (approved):
Nice!

PR #5851, opened 2022-01-26
remove pkg_resources

### LINKED ISSUE #5846: Replace use of pkg_resources with importlib.metadata
Setuptools' `pkg_resources` is no longer recommended and from Python 3.7+ we should be using `importlib.metadata`: https://docs.python.org/3/library/importlib.metadata.html#distribution-versions

nabobalis:
Man this PR sucks.

Fixes #5846

### DIFF
--- a/sunpy/tests/figure_hashes_mpl_dev_ft_261_astropy_dev_animators_100.json
+++ b/sunpy/tests/figure_hashes_mpl_dev_ft_261_astropy_dev_animators_100.json
@@ -1,0 +1,1 @@
+figure_hashes_mpl_dev_ft_261_astropy_dev_animators_dev.json
--- a/sunpy/tests/figure_hashes_mpl_dev_ft_261_astropy_dev_animators_100.json
+++ b/sunpy/tests/figure_hashes_mpl_dev_ft_261_astropy_dev_animators_100.json
@@ -1,1 +1,0 @@
-figure_hashes_mpl_dev_ft_261_astropy_dev_animators_dev.json
--- a/sunpy/tests/helpers.py
+++ b/sunpy/tests/helpers.py
@@ -2,12 +2,14 @@ import sys
 import platform
 import warnings
 from pathlib import Path
+from platform import python_version
 from functools import wraps
+from importlib.metadata import entry_points
 
 import matplotlib as mpl
 import matplotlib.pyplot as plt
-import pkg_resources
 import pytest
+from packaging.version import Version
 
 import astropy
 from astropy.wcs.wcs import FITSFixedWarning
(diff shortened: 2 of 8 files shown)


sunpy/tests/figure_hashes_mpl_dev_ft_261_astropy_dev_animators_100.json
```diff
@@ -0,0 +1 @@
+figure_hashes_mpl_dev_ft_261_astropy_dev_animators_dev.json
```
nabobalis, line 1: This is due to the master version of mpl_animators returns 1.0.0 as a version for now.

sunpy/util/sysinfo.py
```diff
@@ -2,14 +2,41 @@
...
-from pkg_resources import get_distribution
+from collections import defaultdict
+from importlib.metadata import version, requires, distribution
 
 import sunpy.extern.distro as distro
 
 __all__ = ['system_info', 'find_dependencies', 'missing_dependencies_by_extra']
 
 
+def get_requirements(package):
```
nabobalis, line 13: I hate this.

Cadair:
https://packaging.pypa.io/en/latest/requirements.html might be useful?

nabobalis:
> [packaging.pypa.io/en/latest/requirements.html](https://packaging.pypa.io/en/latest/requirements.html) might be useful?

I was going to say this means a new dep but its already there.

nabobalis:
Only really removed the regex

sunpy/util/sysinfo.py
```diff
@@ -2,14 +2,42 @@
...
+    Returns
+    -------
+    `dict`
+        A dictionary of requirements with keys being the extra requirement group names.
+    """
+    requirements: list = requires(package)
+    requires_dict = defaultdict(list)
+    for requirement in requirements:
+        package = Requirement(requirement).name
+        if ";" in requirement:
```
Cadair, line 33: you can use `Requirement().marker` to extract what's after the `;` which at least marginally cleans this up (as you can do `if req.marker` rather than `if ';' in`.

This is still upsetting though.

sunpy/util/sysinfo.py
```diff
@@ -2,14 +2,42 @@
...
+            group = requirement.split("extra == ")[1].strip('"').strip("'").strip()
```
Cadair, line 34: because string parsing is a pain, we should also guard this with `if extra in marker` as you can have other stuff like python version restrictions in the marker.

Cadair: I have no idea how the Python version and extra thing would combine. Euuugghhh this is gross, maybe we don't want to do this.

nabobalis: Don't we need to tho?

nabobalis: Regardless, we are getting this PR in, I will do some fixes.

sunpy/util/sysinfo.py
nabobalis: Its all upsetting.

sunpy/util/sysinfo.py
```diff
@@ -2,14 +2,42 @@
...
+
+    Returns
+    -------
+    `dict`
+        A dictionary of requirements with keys being the extra requirement group names.
+    """
+    requirements: list = requires(package)
+    requires_dict = defaultdict(list)
+    for requirement in requirements:
+        package_name, package_marker = Requirement(requirement).name, Requirement(requirement).marker
```
Cadair, line 32:
```suggestion
        req = Requirement(requirement)
        package_name, package_marker = req.name, req.marker
```

sunpy/util/sysinfo.py
```diff
@@ -2,14 +2,43 @@
...
+    `dict`
+        A dictionary of requirements with keys being the extra requirement group names.
+    """
+    requirements: list = requires(package)
+    requires_dict = defaultdict(list)
+    for requirement in requirements:
+        req = Requirement(requirement)
+        package_name, package_marker = req.name, req.marker
+        if package_marker and "extra ==" in str(package_marker):
+            group = str(package_marker).split("extra == ")[1].strip('"').strip("'").strip()
```
nabobalis, line 35: What if there is a Python version or something else in the extra?

sunpy/tests/helpers.py
```diff
@@ -41,24 +43,33 @@
...
+    asdf_entry_points = pytest.mark.skipif(
+        not all(
+            [
+                enter_point.name == "sunpy"
+                for enter_point in entry_points()["asdf_extensions"]
+            ]
+        ),
+        reason="No SunPy ASDF entry points.",
+    )
+new_hash_library = {}
```
dstansby, line 64: I think it makes sense to leave this next to  `get_hash_library_name()`?

dstansby (approved):

sunpy/tests/helpers.py
ConorMacBride: This is the only reference to `new_hash_library` so maybe just remove it? Probably left by mistake when the hash testing was moved to pytest-mpl.

sunpy/util/sysinfo.py
```diff
@@ -2,14 +2,43 @@
...
+    ----------
+    package : str
+        Package you want requirements for.
+
+    Returns
+    -------
+    `dict`
+        A dictionary of requirements with keys being the extra requirement group names.
+    """
+    requirements: list = requires(package)
```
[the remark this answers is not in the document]
nabobalis: Its there to stop my editor from telling me the type is wrong. It should be fixed in python really,

sunpy/tests/helpers.py
nabobalis: Removed.

sunpy/util/sysinfo.py
nabobalis: Doesnt seem to matter

PR #5854, opened 2022-01-27
added graphviz to rtd

nabobalis:
Check RTD's graphs actually render before merge.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -5,6 +5,7 @@ build:
     python: "3.9"
   apt_packages:
     - libopenjp2-7
+    - graphviz
 
 sphinx:
   builder: html


ConorMacBride:
Also, the links in the inheritance diagrams 404 as they exclude the build version in the URL. Other packages such as AstroPy [have the same problem](https://docs.astropy.org/en/stable/visualization/index.html#class-inheritance-diagram), although for some reason it works [here](https://mcalf.macbride.me/en/stable/code_ref/models.html#class-inheritance-diagram). I compared the configs a while ago but couldn't see any meaningful difference.

ayshih:
> Also, the links in the inheritance diagrams...

That bug has been around for a *long* time (>5 years).  See #2097, and the links therein to issues on other repos.

Cadair (approved):

PR #5867, opened 2022-02-08
Overhaul of `affine_transform()` and `GenericMap.rotate()`

### LINKED ISSUE #3718: `image.transform.affine_transform()` can produce values outside of the range of the input array (e.g., more negative)
When a map is rotated using `map.rotate` It leads to a general yellow-washing of the new map.
map without rotation and scaling :
map
map without rotation and scaling = 1.5 :
Scale1 5

map with rotation = 45 degree and scaling = 4 :
rot45_sc4

map with rotation 45 degree and  scaling = 2 :
Figure_1

map without rotation degree and scaling = 2 :
Figure_2

map with rotation 45 degree and no scaling :
45sc0

I observed that the yellowness vanishes around scale = 1.812 without rotation, and on different thresholds for different rotations.

ayshih:
This PR makes major changes to `sunpy.image.transform.affine_transform()` and `sunpy.map.GenericMap.rotate()`:
- By default (but controllable by a keyword), clips the output image to the range of the input image.  Previously, using `scikit-image` would clip and using `scipy` would not clip.  Fixes #3718.
- Preserves NaNs in the input image through to the output image, with added NaNs around each NaN pixel based on the order of the interpolation.
- Switches the default transformation method to `scipy` due to changes in `scikit-image` (https://github.com/sunpy/sunpy/pull/5742#issuecomment-994329299).
- Fixed the issues with `scikit-image` 0.19
- On GenericMap, changed the math convenience methods (`max()`, `mean()`, `min()`, and `std()`) so that they ignore NaNs
- Changed the default behavior of rotations so that pixels not in the original image are filled with NaNs rather than with zeros
- Created a public decorator `add_rotation_function()` to add rotation functions


ayshih:
This is no longer a slash-and-burn PR that removes `scikit-image` entirely.  With #5916, I switched the default method to `scipy`.

ayshih:
Figure test for clipping showing the over-ringing when `clip=False`:
clipping

Figure test for NaNs showing the preservation and expansion of the NaNs:
nans

Note that the bottom right image (high-order scikit-image rotation) is different from the lower orders.  This was run on scikit-image 0.18.3, and this inconsistency was fixed with scikit-image 0.19.0.  Unfortunately for us, they fixed it in the direction to have the higher orders match the lower orders, and not the other way.

Cadair:
This looks great, you got more you want to do on it?

In the context of #4502 can we clearly document the expected behaviour of this `affine_transform` method with regards to clipping and NaNs etc, so that if other people want to implement other methods backed by opencv or whatever they have a target (and even a test?!)

ayshih:
My hit list, some of which will likely be deferred to future PRs:
- [ ] Simplify the normalization code for the scikit-image implementation
- [ ] Check for differences between scipy and scikit-image when transforming integer arrays
- [ ] Break out each of the rotation implementations into private functions to define the private API for future rotation implementations
- [ ] Look into supporting NaN as a `missing` value

ayshih:
Latest versions of figure tests:
clipping
nans

sunpy/image/transform.py
```diff
@@ -167,7 +107,175 @@ def _get_transform_method(method, use_scipy):
...
+    # Convert the image back to its original range
+    if im_max > 0:
+        rotated_image *= im_max
+    rotated_image += im_min
+
+    return rotated_image
+
+
+_supported_rotation_methods = {'scipy': _rotation_scipy,
+                               'skimage': _rotation_skimage}
```
Cadair, line 281: Rather than having this dict be manually populated, you could have a decorator build it automatically? :D

sunpy/image/transform.py
```diff
@@ -167,7 +107,175 @@ def _get_transform_method(method, use_scipy):
...
+def _rotation_skimage(image, matrix, shift, order, missing, clip):
+    """
+    Rotation using scikit-image
+
+    We handle clipping externally because the clipping behavior of
+    :func:`skimage.transform.warp` is different between orders in {0, 1, 3} and
+    orders in {2, 4, 5}.
+
+    We handle NaN for `missing` externally because :func:`skimage.transform.warp`
+    does not handle it correctly for orders in {2, 4, 5}.
```
Cadair, line 243: These docstrings are useful but are in these hidden functions. If we had a `@sunpy_affine_transform` decorator, the other thing it could do is take the docstring out of here and put it in the `affine_transform` docstring to have details about why you might want to pick one over the other.

sunpy/image/transform.py
```diff
@@ -167,7 +101,176 @@ def _get_transform_method(method, use_scipy):
...
         except ImportError:
-            warn_user("scikit-image could not be imported. Image rotation will use scipy.")
-            method = 'scipy'
+            raise ImportError("scikit-image must be installed to be usable for rotation.")
 
     return method
+
+
+def add_rotation_function(name, *, handles_clipping=False,
+                          handles_image_nans=False, handles_nan_missing=False):
```
Cadair, line 110: How would you feel about removing the default values for these kwargs, so that every call of the decorator is explicit about what it handles. I feel like while this solution is cleaner, it's lacking some of this information when looking at the implementations as you are defaulting to this decorator doing stuff to the output?

sunpy/image/transform.py
```diff
@@ -40,11 +37,14 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
         Defaults to `True` i.e., recenter to the center of the array.
     missing : `float`, optional
         The value to replace any missing data after the transformation.
-    method : {``'skimage'``, ``'scipy'``}, optional
-        Transform function to use. Currently
-        :func:`scipy.ndimage.affine_transform` and
-        :func:`skimage.transform.warp` are supported.
-        Defaults to 'skimage', unless scikit-image can't be imported.
+        Defaults to `numpy.nan`.
+    method : {{{rotation_function_names}}}, optional
```
Cadair, line 41: omg the attack of the curly braces!

sunpy/image/transform.py
```diff
@@ -53,29 +53,8 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
...
-    In the case of `skimage.transform.warp`, the image is normalized to [0, 1]
-    before passing it to the function. It is later rescaled back to the original range.
-
-    Although this function is analogous to the IDL's ``rot`` function, it does not
-    use the same algorithm as the IDL ``rot`` function.
-    IDL's ``rot`` calls the `POLY_2D <https://www.harrisgeospatial.com/docs/poly_2d.html>`__
-    method to calculate the inverse mapping of original to target pixel
-    coordinates. This is a polynomial geometrical transformation.
-    Then optionally it uses a bicubic convolution interpolation
-    algorithm to map the original to target pixel values.
```
Cadair, line 78: I feel like this little bit of info should be captured *somewhere*. It's not just that our implementation of `rotate` is different from `rot` it's that it's actually using a completely different algorithm?

sunpy/image/transform.py
```diff
@@ -167,7 +101,176 @@ def _get_transform_method(method, use_scipy):
...
+
+            # If missing cannot be used directly, use a value in the input range of the image
+            missing_to_use = (missing if handles_nan_missing or not np.isnan(missing)
+                              else np.nanmin(image))
+
+            if not handles_image_nans:
+                isnan = np.isnan(image)
+                has_nans = np.any(isnan)
+            else:
+                has_nans = False
```
Cadair, line 165: Apparently this line isn't covered, I thought we had one rotation method which handled NaNs?

sunpy/image/transform.py
```diff
@@ -167,7 +101,176 @@ def _get_transform_method(method, use_scipy):
...
+        def wrapper(image, matrix, shift, order, missing, clip):
+            clip_to_use = clip if handles_clipping else False
+
+            # If missing cannot be used directly, use a value in the input range of the image
+            missing_to_use = (missing if handles_nan_missing or not np.isnan(missing)
+                              else np.nanmin(image))
+
+            if not handles_image_nans:
+                isnan = np.isnan(image)
+                has_nans = np.any(isnan)
```
Cadair, line 163: The variable name `has_nans` confused me on the `image_to_use` line, because I didn't realise it was dependent on `handles_image_nans`.

sunpy/image/transform.py
```diff
@@ -167,7 +101,176 @@ def _get_transform_method(method, use_scipy):
...
+    handles_clipping : `bool`
+        Specifies whether the rotation function can internally perform clipping
+    handles_image_nans : `bool`
+        Specifies whether the rotation function can internally handle NaNs in the
+        input image
+    handles_nan_missing : `bool`
+        Specifies whether the rotation function can internally handle NaN as the
+        missing value
+
+    Other Parameters
```
Cadair, line 132: I think we could use some documentation on the output requirements, i.e. clipping, NaNs etc with reference to the fact that this decorator can handle those things for you? i.e. If I want to write a new function when should I use the keyword args to the decorator, what if they slow my implementation down etc.

Cadair (commented):
I think this looks great, I would be pretty happy to :shipit:  as it is.

What else do you want to do on it before marking it as ready?

sunpy/image/transform.py
```diff
@@ -167,7 +101,176 @@ def _get_transform_method(method, use_scipy):
...
+                    # Otherwise, check if `missing` should be considered part of the input range
+                    lower = np.nanmin([np.max([missing, np.nanmin(rotated_image)]), np.nanmin(image)])
+                    upper = np.nanmax([np.min([missing, np.nanmax(rotated_image)]), np.nanmax(image)])
+                    rotated_image.clip(lower, upper, out=rotated_image)
+
+            return rotated_image
+
+        _rotation_function_library[name] = wrapper
+
+        affine_transform.__doc__ += (f"\n    Notes for the ``{name}`` rotation method:"
```
Cadair, line 207: Making the sections stand out a little more in the built docs seems like a good idea: https://sunpy--5867.org.readthedocs.build/en/5867/generated/api/sunpy.image.transform.affine_transform.html
```suggestion
        affine_transform.__doc__ += (f"\n    **Notes for the ``{name}`` rotation method:**"
```

ayshih:
> What else do you want to do on it before marking it as ready?

Nothing, really.  Not counting addressing your comments, all I have left is the what's new entry.

sunpy/image/transform.py
ayshih: Good point.  If we were adding a ton of these, I think cleaner might be better than explicit, but that's not going to be the case here.  I'll make the change.

sunpy/image/transform.py
ayshih: I don't think it should be in the docstring of this function.  If anywhere, it should be on the known-issues page as a difference from IDL.

Regardless, I'm not versed enough in the algorithms being used to judge whether this written description of IDL's ``rot`` function is even accurate, or whether it is in fact true that there is no way to match it with any of the `scipy` or `scikit-image` interpolation orders.

sunpy/image/transform.py
ayshih: Yeah, I should rename that variable

sunpy/image/transform.py
ayshih: I believe the current state is that `scikit-image` can handle image NaNs for some interpolation orders, but (unsurprisingly) not the orders where it calls `scipy` code instead of using native code.  As currently designed, the `handles_image_nans` can't be order-dependent, so I set it to `False` for `scikit-image`.

sunpy/image/transform.py
ayshih: Yeah, I can add more description.  In particular, the additional convolution if we have to externally handle image NaNs should be noted.

sunpy/image/transform.py
Cadair: This was my understanding at the point @SolarDrew and I went all the way down this particular :rabbit: :hole:, I can no longer vouch for it's veracity.

I wouldn't say it's a known issue, I would say using an affine transform vs a warp is more appropriate to the WCS papers but I am not that fussed if it's easier to get rid of it then go for it.

sunpy/image/transform.py
ayshih: Fixed

sunpy/image/transform.py
ayshih: Some more description added, but perhaps you still want more

sunpy/image/transform.py
ayshih: Done

sunpy/image/transform.py
ayshih: Renamed

ayshih:
Ready for (re-)review

docs/whatsnew/4.0.rst
```diff
@@ -23,6 +23,17 @@ We have bumped the minimum version of several packages we depend on; these are t
...
+Improvements to map and image rotation
+======================================
+There have been significant improvements to :meth:`sunpy.map.GenericMap.rotate` and :func:`sunpy.image.transform.affine_transform`.
+Most notably, there is now comprehensive support for having not-a-number (NaN) values in pixels, with NaNs now preserved across image rotation.
+Accordingly, the default behavior for these functions has been changed to use NaN for the pixels in the output image that are beyond the extent of the input image.
+To obtain the prior behavior, which used zero by default for such pixels, the ``missing`` needs to be explicitly set to zero.
+
+The default rotation function used internally has been switched to :func:`scipy.ndimage.affine_transform` from :func:`skimage.transform.warp` due to changes in the latest release of `skimage`, although the differences between the rotation results are small.
+The rotation function to use can be selected via the ``method`` keyword argument.
+New rotation functions beyond these two can be added using the decorator :func:`~sunpy.image.transform.add_rotation_function`.
```
Cadair, line 35:
```suggestion
New rotation functions beyond these two can be added using new the decorator :func:`~sunpy.image.transform.add_rotation_function`.
```

docs/conf.py
```diff
@@ -153,6 +153,9 @@
 # Disable google style docstrings
 napoleon_google_docstring = False
 
+# Disable the use of param, which prevents a distinct "Other Parameters" section
+napoleon_use_param = False
```
Cadair, line 157: Does this have any effect on how the docs are rendered?

sunpy/image/transform.py
Cadair: It still feels like it needs some kind of summary for someone coming at this cold. This docstring covers all the technical details but if we threw it at someone who doesn't have any context I don't think they would be able to parse this and write a function and be confident they have handled everything correctly.

Perhaps the best place for this isn't here it's in the dev guide?

docs/conf.py
Cadair: Seemingly not.

docs/whatsnew/4.0.rst
ayshih: Done

docs/conf.py
ayshih: As I understand it, `napoleon_use_param=False` bakes in the usual formatting for the parameter lists, while `napoleon_use_param=True` uses the reStructuredText directive `:param:`, which allows for a theme to specify a non-default formatting.  Since our docs just use the default formatting of parameter lists, there is no difference in the rendered output.

Cadair (approved):

.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       submodules: false
       coverage: codecov
       toxdeps: tox-pypi-filter
-      posargs: -n auto
+      posargs: -n 1
```
nabobalis, line 38: I think you want to remove this fully to bypass xdist setting up workers.

ayshih:
The remaining test failures should be unrelated

nabobalis (approved):

PR #5892, opened 2022-02-18
remove online build, revert in future

nabobalis:
I already pushed to the 3.0 and 3.1 branches.


dstansby (dismissed):
I think we should keep the rest of the GH actions for now until we know for good that azure is working (it isn't on the latest run of this PR)

ConorMacBride (approved):

Issue #5896, opened 2022-02-21
Only run Github actions on the sunpy repo


dstansby:
### Provide a general description of the issue or problem.

Currently Github actions will run on any forks of the sunpy repo too. This is probably unnecessary and wasteful of CI resources, so we should restrict it to just run on PRs/commits made to `sunpy/sunpy`.

Issue #5895, opened 2022-02-21
Add a github actions docs build


dstansby:
### Provide a general description of the issue or problem.

Since we're not running the online Azure builds, we should add a doc build to the github actions run. This involves:

- Making a new template file (probably `python-docs-workflow.yml`) that runs a generic doc build
- Adding a new job to `test.yml` that uses this to build the docs

PR #5897, opened 2022-02-21
more actions

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/5896 
Fixes https://github.com/sunpy/sunpy/issues/5895

### DIFF
--- a/.github/workflows/python-test-workflow.yml
+++ b/.github/workflows/python-test-workflow.yml
@@ -1,5 +1,4 @@
-name: Python test workflow
-
+name: Python testing workflow
 
 on:
   workflow_call:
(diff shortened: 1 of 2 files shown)


.github/workflows/python-test-workflow.yml
```diff
@@ -31,7 +30,12 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install tox tox-pypi-filter
+    - name: Build docs
+      if: contains(inputs.tox-env, 'build_docs')
```
nabobalis, line 34: This is not pretty but it works as a solution for now.

.github/workflows/python-test-workflow.yml
```diff
@@ -31,7 +30,12 @@ jobs:
...
+      run: |
+        tox -e ${{ inputs.tox-env }}
     - name: Run tests
+      if: startsWith(inputs.tox-env, 'py')
```
nabobalis, line 38: This is not pretty but it works as a solution for now.

ConorMacBride (approved):

dstansby:
Why did you get rid of the windows and MacOS builds?

dstansby:
Also I'm not sure there's much point having a CRON build given there's no notification in element for it?

nabobalis:
> Why did you get rid of the windows and MacOS builds?

Azure has them?

nabobalis:
> Also I'm not sure there's much point having a CRON build given there's no notification in element for it?

So we can at least check on the odd occasion that the online and docbuild pass on main?

dstansby:
Ah yeah, I forgot about Azure 😆

nabobalis:
> Ah yeah, I forgot about Azure 😆

If we decide to forget about azure fully, then I will add them back.

PR #5906, opened 2022-02-22
Pin pytest in oldestdeps tests

dstansby:
To help external packagers (e.g. Debian) we should also make sure we support older versions of pytest. I've pinned following our dependency support policy (24 months prior).

xref https://github.com/sunpy/sunpy/issues/5904

PR #5910, opened 2022-02-22
Add GH actions oldestdeps workflow

dstansby:
Putting this in a separate PR to #5906 so we can get it working before trying to pin pytest and pytest-xdist. Before merging we should make the new job run after the linux-py310 build.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,6 +28,12 @@ jobs:
       os: ubuntu-latest
       python-version: '3.10'
       tox-env: py310
+  linux-py38-oldestdeps:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -28,6 +28,12 @@ jobs:
       os: ubuntu-latest
       python-version: '3.10'
       tox-env: py310
+  linux-py38-oldestdeps:
+    uses: ./.github/workflows/python-test-workflow.yml
+    with:
```
nabobalis, line 33:
```suggestion
    needs: linux-py310
    with:
```

nabobalis (approved):

nabobalis:
Backported manually.

PR #5915, opened 2022-02-24
Updating parsing GOES/XRS time array so that it can ignore leap seconds

### LINKED ISSUE #5914: GOES/XRS time issue
### Describe the bug

it looks like the way we currently are registering the time array for the GOES XRS data is including leap seconds, however in the data file it ignores it, meaning that the data is currently off by ~27 seconds for GOES 13, 14, 15 for the reprocessed data (as the time is calculated from seconds from 1970-01-01). Its not much of an issue for GOES 16 and 17 as the time is calculated from seconds since 2000-01-01 12:00:00 and hence there are not many leap seconds. 

it is not an issue with the older GOES files (i.e. the fits files from SDAC) see plot:

Screen Shot 2022-02-24 at 12 23 45

I'll make a PR now to fix this

### To Reproduce

_No response_

### What happened?

_No response_

### Expected behavior

_No response_

### Screenshots

_No response_

### System Details

_No response_

### Installation method

_No response_

hayesla:
## PR Description

fixes the fact the the GOES netcdf files define their times ignoring leap seconds 

Fixes #5914 

## PR Checklist

Need to run some proper tests, but this seems to fix the issue


dstansby (dismissed):
Needs a changelog entry. Maybe a test too? If we currently test loading one of these files just a `assert goes.index[0] == #whatever the value is` would be fine, so we can make sure it is correctly parsed under future changes.

ayshih:
Setting the scale to TAI should not "ignore" leap seconds; TAI counts every second.  Can you provide an example of the problem?

hayesla:
hmm it seems to - but I'm still a little confused

see here:


```python

>>> Time("1970-01-01 00:00:00.0") + TimeDelta(1505001601.255, format="sec")
<Time object: scale='utc' format='iso' value=2017-09-09 23:59:32.255>

>>> Time("1970-01-01 00:00:00.0", scale="tai") + TimeDelta(1505001601.255, format="sec")
 <Time object: scale='tai' format='iso' value=2017-09-10 00:00:01.255>

```

ayshih:
Ah, I understand what you are doing now, and it's a bit of computational trickery.  It empirically works, but the `times` instance has the incorrect scale, and a later line calls `times.datetime` without being explicit about the scale that `times` is in, which causes the leap seconds to seemingly vanish.

The proper way to do what you want should be:
```python
# Perform the time addition in UTime format to ignore leap seconds
times = Time(parse_time(start_time_str).utime + d["time"], format='utime')
```

hayesla:
still figuring out a test for this as the test files for the GOES netcdf files that I truncated changed the metadata so it time array starts from the start time of the day

ayshih (changes requested):
~Blocking this for now, see https://github.com/sunpy/sunpy/issues/5914#issuecomment-1050852291~

hayesla:
this test requires #5917 to be merged first

sunpy/timeseries/sources/goes.py
```diff
@@ -257,12 +257,13 @@ def _parse_netcdf(filepath):
...
                 xrsa = np.array(d["a_flux"])
                 xrsb = np.array(d["b_flux"])
+
                 start_time_str = d["time"].attrs["units"]
                 if not isinstance(start_time_str, str):
                     # For h5netcdf<0.14
                     start_time_str = start_time_str.astype(str)
                 start_time_str = start_time_str.lstrip("seconds since").rstrip("UTC")
-                times = parse_time(start_time_str) + TimeDelta(d["time"], format="sec")
+                times = Time(parse_time(start_time_str).utime + d["time"], format="utime")
```
ayshih, line 266:
```suggestion
                # Perform the time addition in UTime format to ignore leap seconds
                times = Time(parse_time(start_time_str).utime + d["time"], format="utime")
```

sunpy/timeseries/sources/goes.py
```diff
@@ -271,7 +272,7 @@ def _parse_netcdf(filepath):
                     # For h5netcdf<0.14
                     start_time_str = start_time_str.astype(str)
                 start_time_str = start_time_str.lstrip("seconds since")
-                times = parse_time(start_time_str) + TimeDelta(d["time"], format="sec")
+                times = Time(parse_time(start_time_str).utime + d["time"], format="utime")
```
ayshih, line 275:
```suggestion
                # Perform the time addition in UTime format to ignore leap seconds
                times = Time(parse_time(start_time_str).utime + d["time"], format="utime")
```

changelog/5915.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix parsing of the GOES/Xrs netcdf files to ignore leap seconds/
```
ayshih, line 1:
```suggestion
Fix parsing of the GOES/XRS netcdf files to ignore leap seconds.
```

ayshih (approved):
I think it'd be helpful to add some comments lest some well-meaning person in the future try to "simplify" the code.

nabobalis (approved):

PR #5916, opened 2022-02-25
Add method argument to rotate and affine_transform

dstansby:
This PR replaces the `use_scipy` argument with the `method` argument, allowing us more flexibility in how `affine_transform` methods are specified/implemented in the future.

This is pulled out of https://github.com/sunpy/sunpy/pull/4502/files for easier review/discussion.

### DIFF
--- a/sunpy/image/transform.py
+++ b/sunpy/image/transform.py
@@ -6,13 +6,13 @@ import numbers
 import numpy as np
 import scipy.ndimage
 
-from sunpy.util.exceptions import warn_user
+from sunpy.util.exceptions import warn_deprecated, warn_user
 
 __all__ = ['affine_transform']
 
 
 def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
-                     recenter=False, missing=0.0, use_scipy=False):
+                     recenter=False, missing=0.0, use_scipy=None, *, method='skimage'):
     """
     Rotates, shifts and scales an image.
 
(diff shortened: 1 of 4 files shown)


sunpy/image/transform.py
```diff
@@ -148,3 +147,27 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
...
     return rotated_image
+
+
+def _get_transform_method(method, use_scipy):
+    # This is re-used in affine_transform and GenericMap.rotate
+    supported_methods = {'scipy', 'skimage'}
+    if method not in supported_methods:
+        raise ValueError(f'Method {method} not in supported methods: {supported_methods}')
+
+    if use_scipy is not None:
```
Cadair, line 158: https://docs.astropy.org/en/stable/api/astropy.utils.decorators.deprecated_renamed_argument.html#astropy.utils.decorators.deprecated_renamed_argument might be easier?

nabobalis (approved):

sunpy/image/transform.py
dstansby: Ooh fancy, I'll give it a go!

dstansby: This wouldn't quite work, because someone could hand `use_scipy` as a positional argument and the decorator wouldn't handle that situation.

Cadair: You could probably make it work from a quick read of the docstring, but it's not worth it. What you already have works.

Cadair (approved):

PR #5921, opened 2022-02-26
Fix cdaweb doctest

dstansby:
This should protect against future changes in the remote data too.


ConorMacBride (approved):

dstansby:
Woops, this doesn't need to be backported to 3.0.x

Issue #5874, opened 2022-02-10
Refactor `GenericMap.contour` to not use `scikit-image`


wtbarnes:
### Provide a general description of the issue or problem.

Currently, `GenericMap.contour` uses `skimage.measure.find_contours` to calculate the contours (in pixel space) of a map. If we want to remove our dependency on `scikit-image`, we will need to find an alternate way of doing this.

dstansby:
From memory there isn't a better function outside scikit-image to find contours. Matplotlib can do it, but calculating contours is tightly coupled with having a live Axes: https://matplotlib.org/stable/api/contour_api.html#matplotlib.contour.QuadContourSet

dstansby:
From memory the matplotlib contour code is written in C, so probably not that easy to do.

nabobalis:
What about the skiimage one?

wtbarnes:
It sounds like this is a good candidate for an optional dependency then. We move it from the `image` optional dependency option to the `map` option.

dstansby:
In excellent news the contouring code used by Matplotlib is being moved to a new package: https://github.com/contourpy/contourpy, so I think it might be worth re-opening this and replacing `scikit-image` with `contourpy` in the future.

dstansby:
Since we seem to have agreed it would be good to use `contourpy` for our contouring, so it is consitent with what Matplotlib does when drawing contours, I've re-opened this. I think we should wait for https://github.com/matplotlib/matplotlib/pull/22567 to be in a Matplotlib release before doing this. There is a proof of concept in https://github.com/sunpy/sunpy/pull/5923.

Prtm2110:
The pull request at https://github.com/matplotlib/matplotlib/pull/22567 has been merged. Shall I create a new pull request to replace the current contour import with ContourPy?

nabobalis:
Sure.

PR #5923, opened 2022-02-27
Proof of concept use contourpy for contouring

dstansby:
See https://github.com/sunpy/sunpy/issues/5874. It didn't take me long to throw this together so thought I'd open a draft PR for discussion. Things we should consider (this is non-exhaustive and off the top of my head; **please add to this**):

- Should we also change `.draw_contour()` to generate the contours through `contourpy`, to make them consistent with the results of `contour()`?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -55,6 +55,7 @@ image =
 jpeg2000 =
   glymur>=0.8.18,!=0.9.0,!=0.9.5
 map =
+  contourpy>=1.0.1
   matplotlib>=3.2.0
   mpl-animators>=1.0.0
   scipy>=1.3.0
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -2540,7 +2540,7 @@ class GenericMap(NDData):
         """
         Returns coordinates of the contours for a given level value.
 
-        For details of the contouring algorithm see `skimage.measure.find_contours`.
+        For details of the contouring algorithm see `contourpy`.
 
         Parameters
         ----------
@@ -2549,7 +2549,7 @@ class GenericMap(NDData):
             is not `None`, this must be a `~astropy.units.Quantity` with units
             equivalent to the map data units.
         kwargs :
-            Additional keyword arguments are passed to `skimage.measure.find_contours`.
+            Additional keyword arguments are passed to `contourpy.contour_generator`.
 
         Returns
         -------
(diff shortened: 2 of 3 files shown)


nabobalis (approved):

Cadair:
> to make them consistent with the results of contour()?

isn't contourpy the same code mpl uses, so wont it be the same anyway?

dstansby:
> > to make them consistent with the results of contour()?
> 
> isn't contourpy the same code mpl uses, so wont it be the same anyway?

There are three different algorithms:

- MPL to 2014
- MPL 2014 - present
- A new one which is the default and better because it has been written from scratch

I don't think there's any major differences between them, but if we use the default new algorithm for `.contour()` and the `Axes.contour()` default for drawing contours we might end up with slightly different contours.

dstansby:
For reference, here's the Matplotlib issue and pull request:

https://github.com/matplotlib/matplotlib/issues/22529
https://github.com/matplotlib/matplotlib/pull/22567

Cadair:
Happy to change how we do it for now and drop back to the new mpl behaviour in ~18 months when we can depend on it in mpl?

setup.cfg
```diff
@@ -55,6 +55,7 @@ image =
 jpeg2000 =
   glymur>=0.8.18,!=0.9.0,!=0.9.5
 map =
+  contourpy>=1
```
ianthomas23, line 58: This should now be `contourpy>=1.0.1`. I had an unnecessary `numpy` version pin in `contourpy` which might cause problems, the new release fixes this.

ianthomas23:
Contourpy dev here. Do you folks want mpl's `contour/contourf` to have a new kwarg for the contourpy algorithm name? I have been assuming that mpl will switch to using the new algorithm asap and that a kwarg isn't needed as it means explaining the algorithm options to end users who mostly just want it to work. But if there is a use case for the kwarg being there, I am happy to change my mind.

Cadair:
:wave: 

If mpl switches to the new algorithm then I think it has the same effect for us, when we depend on that version of mpl we can say they do the same thing, so I think that approach is fine with us :)

nabobalis:
If we plan to work around the main issue that we have with scikit-image, do we just wait for MPL to depend on contourpy and we close this PR?

dstansby:
I'm not bothered whether this makes it into 4.0; the only good reason to rush seems to be removing the `scikit-image` dependency, which I don't think will happen for 4.0, so I'm going to re-milestone this but leave it open.

nabobalis:
I am happy to get this into 4.0 if we decide that removing scikit-image is important enough, if not, waiting for MPL to do the switch is just easier for us? (We would still need to update the contour function tho)

dstansby:
I think we should hold off on this until MPL makes the switch that won't be before we feature freeze 4.0, so I'm going to close this PR and we can track this at https://github.com/sunpy/sunpy/issues/5874

PR #5939, opened 2022-03-02
Move Publishing to GitHub Actions

ConorMacBride:
## PR Description
This moves the building and publishing of sdist and wheels from Azure Pipelines to GitHub Actions.

I've added a build for `macosx_universal2` wheels. These include both the Intel and Apple Silicon (`arm64`) binaries, however, we should continue to build `macosx_x86_64` to support older versions of pip. However, the `arm64` binaries are not tested.

I removed `CIBW_MANYLINUX_X86_64_IMAGE` and `CIBW_MANYLINUX_I686_IMAGE` as `manylinux2014` is now the default for these architectures.

- [x] Setup a protected [GitHub environment](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) with the PyPI token and Matrix bot token
- [x] Document everything
- [x] **Before merging** update commented out code in `publish.yml` 
- [ ] When backporting, update the tox environments and build targets to match current `azure-pipelines.yml` on those branches
- [x] Update name and profile picture of Matrix bot

## PR Checklist


.github/workflows/publish.yml
```diff
@@ -0,0 +1,40 @@
...
+      - 'main'
+      - '*backport*'
+    tags:
+      - 'v*'
+      - '!*dev*'
+      - '!*pre*'
+      - '!*post*'
+  pull_request:
+    branches:
+      - 'main'
```
dstansby, line 15: I'm guessing this will go once we've checked the build is working in this PR?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,40 @@
...
+#      - '.github/workflows/publish.yml'
+  # Allow manual runs through the web UI
+  workflow_dispatch:
+  schedule:
+    #        ┌───────── minute (0 - 59)
+    #        │ ┌───────── hour (0 - 23)
+    #        │ │ ┌───────── day of the month (1 - 31)
+    #        │ │ │ ┌───────── month (1 - 12 or JAN-DEC)
+    #        │ │ │ │ ┌───────── day of the week (0 - 6 or SUN-SAT)
+    - cron: '0 3 * * *'  # Every day at 03:00 UTC
```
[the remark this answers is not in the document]
ConorMacBride: Probably not! What schedule would you like, if any? It's taking over an hour to build. Probably because it needs to build scikit-image from source quite often.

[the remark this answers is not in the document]
nabobalis: We would probably never schedule it?

.github/workflows/publish.yml
ConorMacBride: Yes, that will be removed before merging. Do we want to include the lines below which should always build on PRs if they modify this workflow file?

.github/workflows/publish.yml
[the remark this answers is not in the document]
ConorMacBride: Okay. Currently it will run for an hour every time we merge a backport. Is that necessary? We could just test the build daily and trigger a manual build when we think it's necessary?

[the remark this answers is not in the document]
nabobalis: Yeah a daily build and trigger it manually seems like a good idea.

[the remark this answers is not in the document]
ConorMacBride: ["Scheduled workflows run on the latest commit on the default or base branch."](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onschedule) which means we can't have scheduled jobs on the backport branches!!! (Confirmed by checking our asv workflow.) There must be some way to have a bot manually trigger a run with the GitHub API but that's out of the scope of this PR I think. I'll just remove the schedule for now.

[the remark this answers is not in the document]
nabobalis: Oh we never want to run this on backport PRs.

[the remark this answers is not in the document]
dstansby: I'd say only run it weekly if we can get it reporting to the element channel?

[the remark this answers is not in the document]
ConorMacBride: I mean run after _merging_ backport PRs?

.github/workflows/publish.yml
dstansby: That sounds like a good idea

.github/workflows/publish.yml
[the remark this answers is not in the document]
nabobalis: Oh, nah. I think only on cron and manually. On tag we want to release after that.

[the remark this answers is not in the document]
ConorMacBride: The new `scheduled_builds.yml` workflow will hopefully take care of the cron.

.github/workflows/publish.yml
Cadair: oh that's a cool little feature.

.github/workflows/publish.yml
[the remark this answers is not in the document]
Cadair: We currently build wheels on every merge to a backport branch.

I am a little concerned that making it weekly on backport branches means that someone doing a release has another step to check before tagging, which is "make sure a wheel build has passed on the last commit". I am *very* against anything which adds steps to the release pipeline that isn't automated.

I propose we maintain the status quo of building wheels on every merge to a backport branch (which aren't *that* frequent) and add a request to sunpy/sunpy-project#2 for more automation of bug fix releases. [I think I can probably make a whole bug fix release be "open an issue and talk to a bot" with actions.]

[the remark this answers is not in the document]
Cadair: Yep, that is what I think :+1:

.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
...
+    branches:
+      - 'main'
+#    paths:
+#      - '.github/workflows/publish.yml'
+  # Allow manual runs through the web UI
+  workflow_dispatch:
+
+jobs:
+  publish:
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@main
```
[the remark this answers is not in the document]
ConorMacBride: According to the [documentation](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow) they must be located at the root of the `.github/workflows` directory.

.github/workflows/scheduled_builds.yml
```diff
@@ -0,0 +1,17 @@
...
+    #        │ │ │ ┌───────── month (1 - 12 or JAN-DEC)
+    #        │ │ │ │ ┌───────── day of the week (0 - 6 or SUN-SAT)
+    - cron: '0 3 * * 1'  # Every Monday at 03:00 UTC
+
+jobs:
+  dispatch_workflows:
+    runs-on: ubuntu-latest
+    steps:
+      - run: gh workflow run publish.yml --ref 3.1
+      - run: gh workflow run publish.yml --ref 3.0
```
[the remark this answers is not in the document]
ConorMacBride: I've now added a cron for `main` and also added the Matrix notifications.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
+name: Build and publish
```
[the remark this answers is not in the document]
ConorMacBride: I've now changed it to a ci.yml workflow that does both. Reusable actions cannot call other reusable actions so the only way I can see is to have one big workflow for all CI/CD.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,128 @@
...
+      default_python: '3.8'
+      submodules: false
+      coverage: codecov
+      toxdeps: tox-pypi-filter
+      posargs: -n auto --dist loadgroup
+      libraries: |
+        apt:
+          - libopenjp2-7
+      envs: |
+        - linux: build_docs_ci
```
nabobalis, line 64: Is this a new tox env?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,128 @@
...
+      coverage: codecov
+      toxdeps: tox-pypi-filter
+      posargs: -n auto
+      libraries: |
+        apt:
+          - libopenjp2-7
+      envs: |
+        - linux: base_deps
+        - linux: py39-devdeps
+        - linux: pyinstaller
```
nabobalis, line 88:
```suggestion
```
Just to preempt us removing it.

.github/workflows/ci.yml
ConorMacBride: Not new to this PR but relatively new.

nabobalis: I must have missed/forgot about it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,128 @@
...
+      submodules: false
+      coverage: codecov
+      toxdeps: tox-pypi-filter
+      posargs: -n auto
+      libraries: |
+        apt:
+          - libopenjp2-7
+      envs: |
+        - linux: base_deps
+        - linux: py39-devdeps
```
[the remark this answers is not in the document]
ConorMacBride: `base_deps` takes ~2 min although `py39-devdeps` takes about ~9 min. It would be good if we didn't have to build wheels on every run. Astropy have a [nightly build](https://dev.azure.com/astropy-project/astropy/_packaging?_a=feed&feed=nightly%40Local) available through an alternative PyPI repository. Maybe Matplotlib has something similar? (The other devdeps are pure python.)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,127 @@
...
+        apt:
+          - libopenjp2-7
+        brew:
+          - openjpeg
+      envs: |
+        - macos: py38
+        - windows: py310
+        - linux: py38-oldestdeps
+        - linux: py39-conda
+          libraries: ''
```
ConorMacBride, line 49: Why does macOS and Linux (apparently) need a JP2 library while Windows and conda on Linux does not?

nabobalis: If I recall, at the time, we couldn't install the JP2 library on windows so it was skipped (along with those tests).
For conda, JP2 is a dependency of the library we use and is installed via conda-forge.

.github/workflows/ci.yml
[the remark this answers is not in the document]
nabobalis: I am not saying anything about the wheels but I feel like running at least devdeps would be good on a PR (basedeps could probably stay cron only or be removed now).

ConorMacBride (commented):
This is now ready for final review.

.github/actions/matrix/action.yml
```diff
@@ -0,0 +1,41 @@
+name: 'Notify Matrix'
```
Cadair, line 1: I linked an existing action for this in the issue, can we not use that one?

ConorMacBride: Actually, that action adds [a lot of unnecessary information to the message](https://github.com/KrazyKirby99999/matrix-commit/blob/cae66ad84edd6a0e5ce3e3a5e0532127d9a6a236/matrix-commit/action.py#L28-L33), so keep as is?

Cadair: Right, ok, I will probably work on a better version with the start of the OSTFL grant so this will do for now :+1:

nabobalis (approved):

nabobalis:
We can leave backporting it to a future PR.

nabobalis:
> * [ ]  Setup a protected [GitHub environment](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) with the PyPI token and Matrix bot token

I assume someone with admin access will need to do that?

> * [ ]  When backporting, update the tox environments and build targets to match current `azure-pipelines.yml` on those branches

Manual backport required then?

> * [ ]  Update name and profile picture of Matrix bot

Was this done?

Cadair (commented):
I have added an environment with the pypi token call `Release` it matches on all branches named `?.?` (i.e. one character, a `.`, then another single character) we will have to update it when we hit v10 ;)

I also added a new matrix account to the sunpy org settings so that all repos can use it, I also added the homeserver address as a secret to the org as well, because then someone can change the matrix account being used for all repos by just editing the org secrets.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,128 @@
...
+      toxdeps: tox-pypi-filter
+      posargs: -n auto
+      libraries: |
+        apt:
+          - libopenjp2-7
+      envs: |
+        - linux: base_deps
+        - linux: py39-devdeps
+
+  publish:
```
Cadair, line 89: Rather than putting publish here, I think it would be better to put it in it's own workflow, which is triggered upon completion of the CI one, like this: https://github.com/DKISTDC/dkist/blob/main/.github/workflows/publish.yml#L3-L8

This will mean that it doesn't show up as a skipped check in all PRs.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,128 @@
...
+        - linux: build_docs_ci
+          posargs: ''
+          pytest: false
+          libraries:
+            apt:
+              - libopenjp2-7
+              - graphviz
+        - linux: py38-online
+
+  cron:
```
Cadair, line 73: Same with the cron-only builds, I think they should be in their own workflow. We can also copy [this trick](https://github.com/astropy/astropy/blob/main/.github/workflows/ci_cron_daily.yml#L30) I came up with for Astropy where by adding the "Extra CI" label to the PR triggers the cron build. This means that if you are modifying something which affects the builds we only cron you can test it on the PR before merging.

---
. o O (We could also do this semi-autonomously for online tests?!)

.github/workflows/ci.yml
Cadair: If we have a way to install the jp2 lib on windows we should and then we can run the tests.

Cadair (approved):
I have a couple of suggestions but they are probably best done in follow up PRs anyway.

.github/workflows/ci.yml
ConorMacBride: @Cadair Yeah, it would be good to remove all the skipped messages. It seems that in the DKIST repo the publish workflow will run even if the CI workflow has failures. [It can be configured to skip if failures like this](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow), although sunpy would need a way to ignore the status of the online tests when publishing which doesn't seem straightforward. However, [`Note: This event will only trigger a workflow run if the workflow file is on the default branch.`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) sounds like this might not work for publishing on the 3.0 and 3.1 branches?

.github/workflows/ci.yml
ConorMacBride: @Cadair I can add the "Extra CI" label to trigger the cron test and publish test in a PR. Do you want to only run the online tests in a PR if an "Online CI" tag is added? (But always run online tests otherwise, i.e. merging PRs, tags and weekly cron.)

Cadair: I think we should keep the online build as default at the moment, and think about if that's something we can do in a bit more of a smart way.

Cadair: Running devdeps on the every PR seems like a waste of :bike: to me, especially with all the compiling.

ConorMacBride: Yeah, it takes ~9 min to run currently. Maybe we should be using astropy's nightly PyPI repository? And for things which don't have that we could have a daily workflow which builds and caches wheels. Maybe just keep the cron stage as-is for now?

PR #5943, opened 2022-03-08
Speeding up Fermi GBM data parsing for TimeSeries

### LINKED ISSUE #5942: Parsing GBM files for GBMSummaryTimeSeries takes too long
### Describe the performance issue

When reading in the Fermi GBM CSPEC and CTIME fits files to create a timeseries, the parsing takes a very long time. 
This can be up to 3 minutes for the CTIME files 😱 which have a very high cadence.

This is due to t he nested loop here:

https://github.com/sunpy/sunpy/blob/main/sunpy/timeseries/sources/fermi_gbm.py#L215

It loops over the time index which is really inefficient.

### To Reproduce

```python
>>> from sunpy import timeseries as ts
>>> import time

>>> t1 = time.time()
>>> gbm_cspec = ts.TimeSeries("glg_cspec_n0_220120_v00.pha")
>>> print("Time taken to read file", time.time() - t1)
Time taken to read file 11.445274829864502

>>> t1 = time.time()
>>> gbm_ctime = ts.TimeSeries("glg_ctime_n0_220120_v00.pha")
>>> print("Time taken to read file", time.time() - t1)
Time taken to read file 188.96947813034058 😱 
```

### Proposed fix

array divisions rather than looping, I'll make a PR now

hayesla:
## PR Description

This fixes #5942 for which reading in fermi GBM files was taking an age and a half

## PR Checklist


hayesla:
for context - it now takes ~0.2 and ~2s to read the CSPEC and CTIME files, respectively compared to ~12, 200s beforehand.

dstansby (changes requested):
Looking at the figure images that have changed (https://75296-2165383-gh.circle-artifacts.com/0/.tmp/py38-figure/figure_test_images/fig_comparison.html) it seems like the data being returned in the 800-2000 keV range is different now? If that's bugfix it should be documented in the changelog; if it's a newly introduced bug it should be fixed 😄

hayesla:
yes sorry I forgot to mention, I've slightly changed the energy edges used for binning (instead of using np.searchsort) and that's the differences here - but it is what we want. I'll update the changelog now! :)

dstansby (approved):
Given this fixes existing behaviour, approving. It seems odd to me that we're doing this kind of inaccurate reduction to the data though, instead of just returning what's in the file...

ayshih (changes requested):
Hmm...  The speed-up is great, but I'm a little uncomfortable with this change being a backported bugfix because it does change the approach to binning the data (which wasn't necessary for the speed-up).  While the prior approach wasn't documented and could be considered nonintuitive, this PR's approach also isn't documented and could also be considered nonintuitive.  At the very least, a note should be added to class's docstring to describe what precisely is being done.

Plus, someone should implement a proper rebinning that splits the data bins, and then the binning approach will change again.

hayesla:
ok yes, I agree - how about I change this PR to just be a speed up, and then another one that properly deals with the binning?

nabobalis:
How about we backport a separate PR with just the speed fix and leave the bin changes for 4.0?

nabobalis:
We can create a seperate changelog for the bin changes and label that breaking for 4.0?

ayshih (changes requested):
The dosctring for `GBMSummaryTimeSeries` ought to elaborate about the rebinning approach.  The existing statement "Note that the data is re-binned from the original 128..." does not have enough detail to discriminate between the old approach and the new approach.  Given the lack of detail, a user might reasonably assume a rebinning approach that includes splitting bins, which is neither the old or the new approaches.

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -31,7 +31,9 @@ class GBMSummaryTimeSeries(GenericTimeSeries):
 
     This summary lightcurve makes use of the CSPEC (daily version) data set which consists of the counts
     accumulated every 4.096 seconds in 128 energy channels for each of the 14 detectors.
+
     Note that the data is re-binned from the original 128 into the following 8 pre-determined energy channels.
+    The rebinning method is ...
```
nabobalis, line 36: TODO: FIX THIS

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -205,25 +207,21 @@ def _bin_data_for_summary(energy_bins, count_data):
     count_data : `numpy.ndarray`
         The array of count data to rebin.
     """
-    # find the indices corresponding to some standard summary energy bins
+
+    # list of energy bands to sum between
     ebands = [4, 15, 25, 50, 100, 300, 800, 2000]
```
ayshih, line 212:
```suggestion
    ebands = [4, 15, 25, 50, 100, 300, 800, 2000]
    indices = [np.searchsorted(energy_bins['e_max'], e) for e in ebands]
```

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -205,25 +207,21 @@ def _bin_data_for_summary(energy_bins, count_data):
...
-                 (energy_bins['e_max'][indices[j]] -
-                  energy_bins['e_min'][indices[j - 1]])))
-
-        summary_counts.append(counts_in_bands)
-
-    return summary_counts
+    for i in range(len(ebands)-1):
+        # find the indices corresponding to some standard summary energy bins
+        ind_start = np.argmin(np.abs(energy_bins["e_min"]-ebands[i]))
+        ind_end = np.argmin(np.abs(energy_bins["e_max"]-ebands[i+1]))
```
ayshih, line 218:
```suggestion
    for ind_start, ind_end in zip(indices[:-1], indices[1:]):
```

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -205,25 +207,21 @@ def _bin_data_for_summary(energy_bins, count_data):
...
+        # sum the counts in the energy bands, and find counts/s/keV
+        summed_counts = np.sum(count_data["counts"][:, ind_start:ind_end+1], axis=1)
```
ayshih, line 220:
```suggestion
        summed_counts = np.sum(count_data["counts"][:, ind_start:ind_end], axis=1)
```

changelog/5943.breaking.rst
```diff
@@ -0,0 +1 @@
+The bin edge indices for `.GBMSummaryTimeSeries` have changed such that it finds the closest energy bin to chosen energy bands.
```
ayshih, line 1:
```suggestion
```
Delete this file

sunpy/timeseries/sources/fermi_gbm.py
ayshih:
```suggestion
    The rebinning method treats the counts in each of the original 128 channels as
    all having the energy of the maximum energy of that channel.  For example, the
    counts in an 13.9--15.1 keV original channel would be accumulated into the
    15--25 keV rebinned channel.
```

ayshih (changes requested):
Here are the changes for a speed improvement alone, without modifying the rebinning scheme, so it is safe to backport.  Given that the speed improvement is ~2 orders of magnitude, the changelog entry should probably be `bugfix` instead of `trivial`.

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -205,25 +210,19 @@ def _bin_data_for_summary(energy_bins, count_data):
...
-                 (energy_bins['e_max'][indices[j]] -
-                  energy_bins['e_min'][indices[j - 1]])))
-
-        summary_counts.append(counts_in_bands)
-
-    return summary_counts
+    for ind_start, ind_end in zip(indices[:-1], indices[1:]):
+        # sum the counts in the energy bands, and find counts/s/keV
+        summed_counts = np.sum(count_data["counts"][:, ind_start:ind_end], axis=1)
+        energy_width = (energy_bins["e_max"][ind_end] - energy_bins["e_min"][ind_start])
```
ayshih, line 222:
```suggestion
        energy_width = (energy_bins["e_max"][ind_end - 1] - energy_bins["e_min"][ind_start])
```

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -31,7 +31,12 @@ class GBMSummaryTimeSeries(GenericTimeSeries):
 
     This summary lightcurve makes use of the CSPEC (daily version) data set which consists of the counts
     accumulated every 4.096 seconds in 128 energy channels for each of the 14 detectors.
+
     Note that the data is re-binned from the original 128 into the following 8 pre-determined energy channels.
+    The rebinning method treats the counts in each of the original 128 channels as
+    all having the energy of the maximum energy of that channel.  For example, the
+    counts in an 13.9--15.1 keV original channel would be accumulated into the
+    15--25 keV rebinned channel.
```
ayshih, line 39:
```suggestion
    all having the energy of the average energy of that channel.  For example, the
    counts in an 14.5--15.6 keV original channel would all be accumulated into the
    15--25 keV rebinned channel.
```

ayshih (changes requested):
Add another `bugfix` changelog entry:
```
Fixed bugs with the rebinning and per-keV normalization for Fermi/GBM summary lightcurves.
```

ayshih:
It turns out that the per-keV calculation is currently wrong, so since we have to bugfix the lightcurves already, I'm now on board with rolling in the better rebinning at the same time.  Let's backport this

changelog/5943.bugfix.1.rst
```diff
@@ -0,0 +1 @@
+Speed up parsing Fermi/GBM files when binning the count data.
```
ayshih, line 1:
```suggestion
Fixed the unintentionally slow parsing of Fermi/GBM files (`~sunpy.timeseries.sources.GBMSummaryTimeSeries`).
```

changelog/5943.bugfix.2.rst
```diff
@@ -0,0 +1 @@
+Fixed bugs with the rebinning and per-keV calculation for Fermi/GBM summary lightcurves.
```
ayshih, line 1:
```suggestion
Fixed bugs with the rebinning and per-keV calculation for Fermi/GBM summary lightcurves (`~sunpy.timeseries.sources.GBMSummaryTimeSeries`).
```

ayshih (approved):
Tweaks to the changelog entries, otherwise looks good (but, I wrote most of it...)

nabobalis:
Doc fails are examples that all I can see.

Issue #5908, opened 2022-02-22
Replace FILLVAL with np.nan in read_cdf()


jgieseler:
### Describe the feature

Previously, Heliopy replaced the `FILLVAL` defined for a variable in a CDF file with `np.nan` (cf. https://github.com/heliopython/heliopy/blob/0.15.4/heliopy/data/util.py#L825). Is there a reason why this is not done in SunPy's `read_cdf()`? (Probably @dstansby)

### Proposed solution

Adopt old Heliopy code from https://github.com/heliopython/heliopy/blob/0.15.4/heliopy/data/util.py#L825

dstansby:
No reason, other than I probably forgot. We should definitely do this.

dstansby:
Had a quick think about this, and instead of filling with `NaN` we should try and used masked arrays and set the mask. That way we're not limited to applying fillval to just floating point columns.

jgieseler:
Yes, that would solve the problem of assigning `np.nan` to an integer column. I think I can take care of it.

dstansby:
I'm just working on this, but it doesn't look like our current CDF test file has any masked values. @jgieseler can you recommend a small (in filesize) CDF file that has some FILLVAL values that we can use to test with?

dstansby:
Just coming back to this, and sadly it looks like pandas doesn't have support for masking data 😢 However, once https://github.com/sunpy/sunpy/pull/5834 is merged we can mask the values in the astropy Timeseries (see https://docs.astropy.org/en/stable/timeseries/masking.html)

dstansby:
The issue is that `NaN` is a floating point number, and there's no equivalent for integer data types. In `heliopy` I hacked this by converting all integer data type columns to floating point first, but that was a Bad Idea because it turns out floating point numbers take up much more space than integers (who knew?!). So the approach in that PR works fine for floating point columns, but I'd like it to work more generally which requires support for properley masking data.

dstansby:
I'm going to re-open this as we only fixed it for floating point data types.

nabobalis:
Do the integer data have fill values too?

If so, we shouldn't need to worry about DTYPE and just replace it wholesale.

dstansby:
They do, but `NaN` is a floating point number, with no integer equivalent, so we can't set the masked values to `NaN` for integer data.

nabobalis:
They don't have a seperate fill value for INT data stored in the metadata?

dstansby:
[...]

PR #5956, opened 2022-03-12
Mask values when reading CDF files

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/5908. Still needs tests and a changelog.

I would say lets not backport this as it's a change in behaviour (not really a bugfix), but open to being swayed on that.


dstansby:
This is waiting on https://github.com/sunpy/sunpy/pull/5834, which will make masked values possible.

dstansby:
@jgieseler it's looking like it's going to be a while (~1 year) until we settle on a long term storage container for sunpy TimeSeries, do you think it's valueable to convert only floating point values to `NaN` in the meantime?

jgieseler:
I guess this would be better than nothing for now, especially if the effort to include it isn't too big (which I guess is the case).

I would say that at least for charged particle intensities, usually the level 2 science data should never be integers. Only lower level data like the actual counts or something like status information should be integers, I would say. So this workaround should already cover the most important cases. At least for my field of data files, don't know how it is in other areas...

Issue #5425, opened 2021-06-28
Move `sunpy.image.coalignment` to sunkit-image?


Cadair:
We don't rely on this anywhere inside core I don't think, so maybe move it out?

Cadair:
@wafels just because he wrote it.

dstansby:
I'm 👍 to doing this

wtbarnes:
Do we deprecate first? Or do we just import from `sunkit-image` and add `sunkit-image` to the optional dependencies?

Cadair:
I would rather not depend on sunkit-image, so I say deprecate and not remove.

wafels:
What about the rest of the sunpy/image ?

wtbarnes:
I would say no. The rest of image is basically resample and affine transform, both of which `Map` heavily depends on.

PR #5957, opened 2022-03-13
Deprecate image.coalignment

wtbarnes:
`image.coalignment` has been moved to `sunkit-image` (See sunpy/sunkit-image#78).
This PR deprecates the `coalignment` module.
This deprecation will be included in 4.0 and thus we can remove the module in 4.1

Addresses #5425.

## PR Checklist

- [ ] I have added a `Fixes #XXXX -` or `Closes #XXXX -` comment to auto-close the issue that your PR addresses

### DIFF
--- a/changelog/5957.deprecation.rst
+++ b/changelog/5957.deprecation.rst
@@ -1,3 +1,3 @@
-Deprecate the `~sunpy.image.coalignment` as the code has now been moved to
+Deprecate `~sunpy.image.coalignment` as the code has now been moved to
 `~sunkit_image.coalignment`.
 This module will be removed in sunpy 4.1.
--- a/changelog/5957.deprecation.rst
+++ b/changelog/5957.deprecation.rst
@@ -1,0 +1,3 @@
+Deprecate the `~sunpy.image.coalignment` as the code has now been moved to
+`~sunkit_image.coalignment`.
+This module will be removed in sunpy 4.1.
--- a/changelog/5957.deprecation.rst
+++ b/changelog/5957.deprecation.rst
@@ -1,3 +1,3 @@
-Deprecate `~sunpy.image.coalignment` as the code has now been moved to
-`~sunkit_image.coalignment`.
+Deprecate `sunpy.image.coalignment` as the code has now been moved to
+`sunkit_image.coalignment` with an identical API.
 This module will be removed in sunpy 4.1.
--- a/setup.cfg
+++ b/setup.cfg
@@ -177,6 +177,8 @@ filterwarnings =
     # These should be removed when our scikit-image pinning is removed
     ignore:Please use `gaussian_filter` from the `scipy.ndimage` namespace:DeprecationWarning
     ignore:Please use `laplace` from the `scipy.ndimage` namespace:DeprecationWarning
+    # Remove this once we remove sunpy.image.coalignment in v4.1
+    ignore:sunpy.image.coalignment is deprecated and will be removed in sunpy 4.1.
 
 [pycodestyle]
 max_line_length = 110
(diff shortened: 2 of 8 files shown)


nabobalis (approved):

wtbarnes:
The doc fail is real. I've messed up the use of `.. warning::` in the module docstring.

wtbarnes:
I just ended up slapping the deprecated decorator on every function.

changelog/5957.deprecation.rst
```diff
@@ -0,0 +1,3 @@
+Deprecate `~sunpy.image.coalignment` as the code has now been moved to
```
dstansby, line 1: You should mention here that all functionality has been replicated with the same API in `sunkit_image.coalignment`. (if it has, I hope it has...) Also mention the motivation for doing this? Users are likely to grumble less if we explain to them why we're making their lives harder.

Also remove the tidles, or this will show up as "Deprecated coalignment, code has been moved to coalignment"!!

sunpy/image/coalignment.py
```diff
@@ -1,4 +1,8 @@
 """
+.. deprecated:: 4.0
+    Use `~sunkit_image.coalignment` instead.
```
dstansby, line 3: Get rid of the tilde

dstansby (changes requested):

setup.cfg
```diff
@@ -177,6 +177,8 @@ filterwarnings =
     # These should be removed when our scikit-image pinning is removed
     ignore:Please use `gaussian_filter` from the `scipy.ndimage` namespace:DeprecationWarning
     ignore:Please use `laplace` from the `scipy.ndimage` namespace:DeprecationWarning
+    # Remove this in 4.1 when we remove the image.coalignment module.
+    ignore:The .* function is deprecated and may be removed in version 4.1.\s+Use `~sunkit_image.coalignment.* instead.:sunpy.util.exceptions.SunpyDeprecationWarning
```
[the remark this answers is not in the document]
wtbarnes: I'm now using `filterwarnings`. I have to admit I didn't see the utility in your suggestion at first, but switching to explicitly marking each with `filterwarnings` has revealed that we are actually using `apply_shifts` (one of the functions I'm deprecating here) in `physics.solar_rotation.mapsequence_solar_derotate`.

changelog/5957.deprecation.rst
wtbarnes: Yes, the API is identical. I've added a "What's New" entry that gives some explanation as to the motivation as I figured this would be more visible to users than a changelog.

dstansby:
I guess the pragmatic option is move the required code but make it private, and then we can make a decision on what to do about `mapsequence_solar_derotate` independently of this PR?

wtbarnes:
> I guess the pragmatic option is move the required code but make it private, and then we can make a decision on what to do about `mapsequence_solar_derotate` independently of this PR?

That seems sensible. I had kind of forgotten that `physics.solar_rotation` existed...

dstansby (changes requested):
One of the remote data test fails is real: https://github.com/sunpy/sunpy/runs/5557145861?check_suite_focus=true#step:7:5237 and I've just rebooted the figure tests to check them (I think the fail was the MPL issue that's now fixed)

wtbarnes:
> One of the remote data test fails is real: https://github.com/sunpy/sunpy/runs/5557145861?check_suite_focus=true#step:7:5237 and I've just rebooted the figure tests to check them (I think the fail was the MPL issue that's now fixed)

Ah thanks for catching that. I've ignored the warnings in the two doctests. We'll also need to remove that section of the guide once the coalignment stuff actually gets removed.

dstansby (approved):

nabobalis (approved):

PR #5958, opened 2022-03-13
Fix some problems around `warn_deprecated`

wtbarnes:
This PR fixes two minor issues

* There is a code path in `deprecate_positional_args_since` that calls `warn_deprecated` but the import for it was missing. I guess we've never hit this code path?
* The dev guide was using `warnings.warn` instead of `warn_deprecated` which contradicts our pre-commit config (https://github.com/sunpy/sunpy/blob/3d3cfcf2669e0435e4f6a037fb2cf5314f10d7f4/.pre-commit-config.yaml#L108). The example now uses `warn_deprecated`.

### DIFF
--- a/docs/dev_guide/contents/api.rst
+++ b/docs/dev_guide/contents/api.rst
@@ -44,15 +44,15 @@ The same applies if you want to change the default value of a keyword argument f
 
 .. code-block:: python
 
+    from sunpy.util.exceptions import warn_deprecated
+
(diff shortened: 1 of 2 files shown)


nabobalis (approved):

nabobalis:
I forgot if this needs to go back to 3.0 (I suspect not but I put the label on anyway).

PR #5971, opened 2022-03-16
Removed broken online test

nabobalis:
This is easier than fixing it.

Also updated the changelog entries.


wtbarnes (approved):
Oops. Thanks for fixing my change logs haha.

PR #5977, opened 2022-03-18
Rename shift method

### LINKED ISSUE #5964: Rename `GenericMap.shift` to `GenericMap.shift_pointing`

This is a useful method that I only recently realized existed. However, I think the name is misleading. If I have a map `m` and I call `m.shift(dx,dy)`, I would naively expect the actual image data to be shifted. However, this function is not doing that. It is shifting the pointing by applying shift to the reference coordinate, thereby shifting the pointing.

I would recommend that we

* rename this function `shift_pointing`
* make `shift` an alias of `shift_pointing`
* deprecate `shift` in 4.0 so we can remove it in 4.1

Additionally, we also keep track of the modifications to the pointing by incrementing `GenericMap.shifted_value` every time `shift` is called. However, I'm not sure how useful this actually is as soon you write this map to disk, all record of this shift is lost. I'd also be inclined to deprecate `shifted_value` but I'm happy to hear arguments against it.

wtbarnes:
## PR Description

Fixes #5964.

`Map.shift` is now `Map.shift_reference_coord` and `Map.shifted_value` is deprecated in favor of using `Map.meta.modified_items`.

## PR Checklist


docs/whatsnew/4.0.rst
```diff
@@ -55,6 +55,21 @@ Second, the module has seen little development in several years and by moving
...
+Deprecation of `~sunpy.map.GenericMap.shift` and `~sunpy.map.GenericMap.shifted_value`
+======================================================================================
+The method `sunpy.map.GenericMap.shift` has been renamed
+`sunpy.map.GenericMap.shift_reference_coord` and `~sunpy.map.GenericMap.shift` has been
+deprecated.
+The method has been renamed to make it clear that it is the reference coordinate that is
+being shifted and not the image itself.
+Additionally, the `sunpy.map.GenericMap.shifted_value` property, which keeps track of
+the shifts applied by `~sunpy.map.GenericMap.shift`, has been deprecated.
+Users should instead use the ``CRVAL1`` and ``CRVAL2`` keys  in
```
dstansby, line 67:
```suggestion
Users should instead use the ``CRVAL1`` and ``CRVAL2`` keys in
```

sunpy/map/tests/test_mapbase.py
```diff
@@ -495,40 +500,44 @@ def test_shift_applied(generic_map):
                                 generic_map.reference_coordinate.Ty)
     x_shift = 5 * u.arcsec
     y_shift = 13 * u.arcsec
-    shifted_map = generic_map.shift(x_shift, y_shift)
+    shifted_map = generic_map.shift_reference_coord(x_shift, y_shift)
```
dstansby, line 503: Instead of using the new method, I would keep the tests as `shift()` here (and filter/catch the warnings) so we're still testing both methods until the deprecated one is removed.

wtbarnes: Ah that's a good point. I'll fix that.

sunpy/map/mapbase.py
```diff
@@ -975,15 +975,32 @@ def center(self):
...
+            '`sunpy.map.GenericMap.shifted_value` is deprecated and will be removed in sunpy 4.1. '
+            'Use `sunpy.map.GenericMap.meta.modified_items` to see how the '
+            'reference coordinate has been modified.'
+        )
         return self._shift
 
+    @deprecated('4.0', alternative='`sunpy.map.GenericMap.shift_reference_coord`')
     @u.quantity_input
     def shift(self, axis1: u.deg, axis2: u.deg):
+        GenericMap.shift.__doc__ = f"{GenericMap.shift_reference_coord.__doc__}"
```
dstansby, line 988:
```suggestion
        GenericMap.shift.__doc__ = GenericMap.shift_reference_coord.__doc__
```
The format string isn't actually doing anything??

sunpy/map/mapbase.py
```diff
@@ -975,15 +975,32 @@ def center(self):
...
+    def shift_reference_coord(self, axis1: u.deg, axis2: u.deg):
         """
         Returns a map shifted by a specified amount to, for example, correct
         for a bad map location. These values are applied directly to the
-        `~sunpy.map.GenericMap.reference_coordinate`. To check how much shift
-        has already been applied see `~sunpy.map.GenericMap.shifted_value`
+        `~sunpy.map.GenericMap.reference_coordinate`. To check how much the
+        reference coordinate has been modified, see
+        `~sunpy.map.GenericMap.meta.modified_items['CRVAL1']` and
+        `~sunpy.map.GenericMap.meta.modified_items['CRVAL2']`.
```
dstansby, line 1003:
```suggestion
        ``sunpy.map.GenericMap.meta.modified_items['CRVAL1']`` and
        ``sunpy.map.GenericMap.meta.modified_items['CRVAL2']``.
```
Since these are code snippets, not sphinx links

dstansby (changes requested):
There are some doc build warnings that need fixing; I think I suggested the fix for one of them, but there might be others.

dstansby:
(otherwise looks 👍 )

sunpy/map/mapbase.py
```diff
@@ -981,15 +981,32 @@ def center(self):
...
     def shift(self, axis1: u.deg, axis2: u.deg):
+        GenericMap.shift.__doc__ = GenericMap.shift_reference_coord.__doc__
+
+        new_map = self.shift_reference_coord(axis1, axis2)
+        new_map._shift = SpatialPair(self.shifted_value[0] + axis1,
+                                     self.shifted_value[1] + axis2)
+        return new_map
+
+    @u.quantity_input
+    def shift_reference_coord(self, axis1: u.deg, axis2: u.deg):
```
Cadair, line 1002: I think this is now a pretty unique method? We don't have any methods for modifying other metadata?

wtbarnes:
Fix the asdf stuff here: https://github.com/sunpy/sunpy/blob/main/sunpy/io/special/asdf/tags/map/generic_map.py#L35

Optionally read `shift` but do not write it.

sunpy/map/mapbase.py
wtbarnes: 🤷  you're not wrong. I think it is still useful though. I'm not sure what rationale we would have for removing it wholesale.

sunpy/io/special/asdf/tags/map/generic_map.py
```diff
@@ -19,7 +18,7 @@ class GenericMapType(SunPyType):
     def from_tree(cls, node, ctx):
         # Use the factory here to get the correct subclass back
         out_map = sunpy.map.Map(np.asarray(node['data']), node['meta'])
-        out_map.shift(*node['shift'])
+        out_map.shift_reference_coord(*node['shift'])
```
Cadair, line 21: You need to guard this as new files wont have shift in the node.
```suggestion
        if 'shift' in node:
            out_map.shift_reference_coord(*node['shift'])
```

Cadair, line 21: Does calling this when the shift is 0 do anything? If it does (i.e. leads to modifying the keys with the same value) it would also be good to check if the shift value in the file is 0 so that we don't manipulate the meta without reason.

wtbarnes: 🤦 thanks!

wtbarnes: No, there's no special casing for 0 shift. I could special case it here or in the method itself?

I'm also now quite confused as this method is being called as if it's an in-place operation but the function returns a new map with the shift applied. I'm not sure that an asdf roundtrip test with non-zero shift would've ever passed.

wtbarnes: Kind of. It increments the `_shift` attribute and then returns a new map with that incremented attribute.

https://github.com/sunpy/sunpy/blob/fa2c427e6a54493ac39d9536a431504ccc04d014/sunpy/map/mapbase.py#L1017

Cadair: heh, then yeh it probably never worked :facepalm:

dstansby (approved):
I haven't reviewed the ASDF changes (not enough time to review that bit properley), but the rest looks 👍

nabobalis:
Needs a rebase.

nabobalis (approved):

sunpy/io/special/asdf/tags/map/generic_map.py
```diff
@@ -19,7 +18,8 @@ class GenericMapType(SunPyType):
     def from_tree(cls, node, ctx):
         # Use the factory here to get the correct subclass back
         out_map = sunpy.map.Map(np.asarray(node['data']), node['meta'])
-        out_map.shift(*node['shift'])
+        if 'shift' in node:
+            out_map = out_map.shift_reference_coord(*node['shift'])
```
Cadair, line 22: Ideally we would test this by making a file before this change with a shift and asserting that the shift gets applied after it.

wtbarnes: I would argue that's not in scope for this PR.

sunpy/io/special/asdf/tests/test_genericmap.py
```diff
@@ -12,13 +12,13 @@
...
-
     # Test the meta by force!
     for ok, ov in old.meta.items():
         assert ok in new.meta
         assert new.meta[ok] == ov
-
-    assert u.allclose(old.shifted_value, new.shifted_value)
+    if 'crval1' in old.meta.modified_items and 'crval1' in new.meta.modified_items:
+        np.testing.assert_allclose(old.meta.modified_items["crval1"], new.meta.modified_items["crval1"])
+        np.testing.assert_allclose(old.meta.modified_items["crval2"], new.meta.modified_items["crval2"])
```
nabobalis, line 21: Unsure if this works, I don't think we actually hit this code in any real way.

sunpy/io/special/asdf/converters/generic_map.py
```diff
@@ -30,7 +29,8 @@ def to_yaml_tree(self, smap, tag, ctx):
         node = {}
         node["data"] = np.asarray(smap.data)
         node["meta"] = dict(smap.meta)
-        node["shift"] = u.Quantity(smap.shifted_value)
+        if 'crval1' in smap.meta.modified_items:
```
nabobalis, line 32: Not sure if this check is enough

sunpy/io/special/asdf/tags/map/generic_map.py
Cadair: I did it. Code without tests shall not pass :p

Cadair:
pre-commit.ci autofix

Cadair (approved):
I approve the non-asdf parts and the asdf parts, but I wrote those so someone else should review this properly.

sunpy/map/mapbase.py
```diff
@@ -1010,15 +1011,32 @@ def center(self):
...
+            '`sunpy.map.GenericMap.shifted_value` is deprecated and will be removed in sunpy 4.1. '
+            'Use ``sunpy.map.GenericMap.meta.modified_items`` to see how the '
+            'reference coordinate has been modified.'
+        )
         return self._shift
 
+    @deprecated('4.0', alternative='`sunpy.map.GenericMap.shift_reference_coord`')
     @u.quantity_input
     def shift(self, axis1: u.deg, axis2: u.deg):
+        GenericMap.shift.__doc__ = GenericMap.shift_reference_coord.__doc__
```
ConorMacBride, line 1024: I don't think updating the docstring here has any effect. Should it be updated at the bottom of the file instead? [Here's it in the docs.](https://sunpy--5977.org.readthedocs.build/en/5977/generated/api/sunpy.map.GenericMap.html#sunpy.map.GenericMap.shift)

sunpy/io/special/asdf/converters/generic_map.py
```diff
@@ -12,25 +11,36 @@ class GenericMapConverter(Converter):
...
 
         # Use the factory here to get the correct subclass back
         out_map = sunpy.map.Map(np.asanyarray(node["data"]), node["meta"])
-        out_map = out_map.shift(*node["shift"])
-
         out_map.mask = node.get("mask")
         out_map.uncertainty = node.get("uncertainty")
         out_map._unit = node.get("unit")
+
+        # Old files can have a shift that needs applying
```
ConorMacBride, line 34: The existing shift method set the `shift` value *and* updated the Map `meta`. So was `from_yaml_tree` shifting it twice? Or maybe the `shift` was `0º` in this case?

ConorMacBride (approved):

sunpy/io/special/asdf/converters/generic_map.py
ConorMacBride: `GenericMap.shift` does:

```py
new_meta['crval1'] = ((self._reference_longitude + axis1).to(self.spatial_units[0])).value
new_meta['crval2'] = ((self._reference_latitude + axis2).to(self.spatial_units[1])).value
```

and

```py
new_map._shift = SpatialPair(self.shifted_value[0] + axis1,
                             self.shifted_value[1] + axis2)
```

If you then save to ASDF and run `sunpy.io.special.asdf.converters.generic_map.from_yaml_tree` it does:

```py
out_map = sunpy.map.Map(np.asanyarray(node["data"]), node["meta"])
```

which already includes the shift (as `crval1` and `crval2` were already updated by `GenericMap.shift`), however it also does:

```py
out_map = out_map.shift_reference_coord(*node['shift'])
```

Which shifts it again?

sunpy/map/mapbase.py
ConorMacBride: I tried that but it doesn't update the docstring. Inside the method the docstring must be a string literal. The code inside the method isn't run unless it is called and doing `help(GenericMap.shift)` doesn't do that.

nabobalis: Thanks for checking, to the end of the file it is.

changelog/6052.breaking.rst
```diff
@@ -1 +1 @@
-The `.JSOCClient` and every `.GenericClient` was passing all ``**kwargs`` to `parfive.Downloader.enqueue_file`, this was unintended and has been removed.
+The `.JSOCClient` and every `sunpy.net.dataretriever.GenericClient` was passing all ``**kwargs`` to `parfive.Downloader.enqueue_file`, this was unintended and has been removed.
```
nabobalis, line 1: Ah, so that didn't link the first time.

sunpy/data/test/_generate_asdf_test.py
```diff
@@ -0,0 +1,23 @@
+"""
```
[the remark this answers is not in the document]
Cadair: done

changelog/6052.breaking.rst
Cadair: It was throwing warnings, because of the ambiguous references.

nabobalis (approved):
Assuming how we deal with the shifts are correct, LGTM

sunpy/io/special/asdf/converters/generic_map.py
Cadair: So both these things don't happen together.

We no longer save the shift explicitly (see the `to_yaml_tree` method), it is just folded in to `meta`. What this code supports is loading files which were written with sunpy <4.0 (genericmap tag version 1.0.0), where the shift is explicitly saved in the file and we need to apply it on load.

PR #5979, opened 2022-03-19
Fix uncaptured image.coalignment warnings during release tests

ConorMacBride:
The `@pytest.mark.filterwarnings(DEP_WARNING)` does not apply when testing sdist and wheels as the pytest warnings plugin is disabled. This updates how the warning is captured.

### DIFF
--- a/sunpy/image/tests/test_coalignment.py
+++ b/sunpy/image/tests/test_coalignment.py
@@ -1,5 +1,4 @@
 import os
-import warnings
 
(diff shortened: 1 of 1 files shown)


ConorMacBride:
Release stage passed here: https://dev.azure.com/sunpy/sunpy/_build/results?buildId=23794 (Manual run)

wtbarnes (approved):
Thanks for catching this!

Issue #5994, opened 2022-03-22
Remove Mx/cm2 to G conversions in Map.


nabobalis:
### Describe the feature

I have submitted a PR (https://github.com/astropy/astropy/pull/12975) to astropy to add Maxwell as a CGS unit. When we finally depend on that version (assuming it gets accepted), we should leave the units alone and remove any places where do do this conversion. 

In the mean time, we could add a custom unit to sunpy and use that in the mean time?

dstansby:
+1 to adding a custom unit

PR #5998, opened 2022-03-23
Temporary addition of a Maxwell unit 

nabobalis:
Added maxwells and the workarounds needed to get it to work.

Fixes https://github.com/sunpy/sunpy/issues/5994

### DIFF
--- a/sunpy/map/sources/sdo.py
+++ b/sunpy/map/sources/sdo.py
@@ -198,8 +198,8 @@ class HMISynopticMap(HMIMap):
         unit_str = self.meta.get('bunit', None)
         if unit_str is None:
             return
-		# Maxwells aren't in the IAU unit style manual and therefore not a valid FITS unit
-		# The mapbase unit property forces this validation, so we must override it to prevent it.
+        # Maxwells aren't in the IAU unit style manual and therefore not a valid FITS unit
+        # The mapbase unit property forces this validation, so we must override it to prevent it.
         return u.Unit(unit_str)
 
     @classmethod
--- a/sunpy/map/sources/sdo.py
+++ b/sunpy/map/sources/sdo.py
@@ -193,6 +193,13 @@ class HMISynopticMap(HMIMap):
         else:
             return date
 
+    @property
+    def unit(self):
+        unit_str = self.meta.get('bunit', None)
+        if unit_str is None:
+            return
+        return u.Unit(unit_str)
+
     @classmethod
     def is_datasource_for(cls, data, header, **kwargs):
         """
--- a/sunpy/map/sources/sdo.py
+++ b/sunpy/map/sources/sdo.py
@@ -193,15 +193,6 @@ class HMISynopticMap(HMIMap):
         else:
             return date
 
-    @property
-    def unit(self):
-        unit_str = self.meta.get('bunit', None)
-        if unit_str == 'Mx/cm^2':
-            # Maxwells aren't in the IAU unit sytle manual, so replace with Gauss
-            return u.Unit('G')
-        else:
-            return super().unit
-
     @classmethod
     def is_datasource_for(cls, data, header, **kwargs):
         """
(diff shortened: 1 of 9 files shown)


sunpy/map/sources/soho.py
```diff
@@ -279,6 +279,13 @@ def spatial_units(self):
 
         return SpatialPair(u.Unit(cunit1), u.Unit(cunit2))
 
+    @property
+    def unit(self):
```
dstansby, line 283: This whole property seems like it's not doing anything - is there a reason you can't remove it?

nabobalis: I want to avoid it calling the super version as it will error from the fits validation unit check.

dstansby: What's the error you're getting? And if it's erroring now, won't it also error when the unit is added to astropy, so we'll have to fix the `._parse_fits_unit` method at some point any way?

nabobalis: The error is that Maxwell is not a valid FITS unit. So to avoid it being called, I override the unit on the map source.

My impression and Stuart (@cadair) will correct me that the code in mapbase is only going to allow FITS compliance and anything else has to go into the map source.

nabobalis: Will do.

sunpy/map/sources/sdo.py
```diff
@@ -196,11 +196,9 @@ def date(self):
...
     def unit(self):
         unit_str = self.meta.get('bunit', None)
-        if unit_str == 'Mx/cm^2':
-            # Maxwells aren't in the IAU unit sytle manual, so replace with Gauss
-            return u.Unit('G')
-        else:
-            return super().unit
+        if unit_str is None:
+            return
+        return u.Unit(unit_str)
```
nabobalis, line 201:
```suggestion
		# Maxwells aren't in the IAU unit style manual and therefore not a valid FITS unit
		# The mapbase unit property forces this validation, so we must override it to prevent it.
        return u.Unit(unit_str)
```

sunpy/map/sources/soho.py
```diff
@@ -279,6 +279,13 @@ def spatial_units(self):
...
+        bunit = self.meta.get('bunit', None)
+        if bunit is None:
+            return
+        return u.Unit(bunit)
```
nabobalis, line 287:
```suggestion
		# Maxwells aren't in the IAU unit style manual and therefore not a valid FITS unit
		# The mapbase unit property forces this validation, so we must override it to prevent it.
        return u.Unit(bunit)
```

sunpy/map/sources/soho.py
```diff
@@ -279,6 +279,15 @@ def spatial_units(self):
 
         return SpatialPair(u.Unit(cunit1), u.Unit(cunit2))
 
+    @property
+    def unit(self):
+        bunit = self.meta.get('bunit', None)
+        if bunit is None:
+            return
+		# Maxwells aren't in the IAU unit style manual and therefore not a valid FITS unit
+		# The mapbase unit property forces this validation, so we must override it to prevent it.
```
nabobalis, line 288:
```suggestion
        # Maxwells aren't in the IAU unit style manual and therefore not a valid FITS unit
        # The mapbase unit property forces this validation, so we must override it to prevent it.
```

sunpy/map/sources/sdo.py
```diff
@@ -196,11 +196,11 @@ def date(self):
...
         unit_str = self.meta.get('bunit', None)
-        if unit_str == 'Mx/cm^2':
-            # Maxwells aren't in the IAU unit sytle manual, so replace with Gauss
-            return u.Unit('G')
-        else:
-            return super().unit
+        if unit_str is None:
+            return
+		# Maxwells aren't in the IAU unit style manual and therefore not a valid FITS unit
+		# The mapbase unit property forces this validation, so we must override it to prevent it.
```
nabobalis, line 202:
```suggestion
        # Maxwells aren't in the IAU unit style manual and therefore not a valid FITS unit
        # The mapbase unit property forces this validation, so we must override it to prevent it.
```

nabobalis:
**NOTE TO MAINTAINERS: SQUASH MERGE**

PR #6014, opened 2022-03-29
Update cron schedule and add GitHub CLI token

ConorMacBride:
This PR updates the schedule of the cron to daily at 07:00 UTC, which is what I think we agreed on Matrix. There was also an 
issue with authentication on the scheduled workflow run yesterday. The GitHub CLI needs a `GITHUB_TOKEN` to be present as an 
env variable. Do we need to create a new GitHub account for this purpose? Ideally it should only have permission to dispatch 
workflows. (This PR was opened from the GitHub CLI!)

- ~[ ] Add `GITHUB_TOKEN` secret to the `main` branch environment~

### DIFF
--- a/.github/workflows/scheduled_builds.yml
+++ b/.github/workflows/scheduled_builds.yml
@@ -7,7 +7,7 @@ on:
     #        │ │ ┌───────── day of the month (1 - 31)
     #        │ │ │ ┌───────── month (1 - 12 or JAN-DEC)
     #        │ │ │ │ ┌───────── day of the week (0 - 6 or SUN-SAT)
-    - cron: '0 3 * * 1'  # Every Monday at 03:00 UTC
+    - cron: '0 7 * * *'  # Every day at 07:00 UTC
 
 jobs:
   dispatch_workflows:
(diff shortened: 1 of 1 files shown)


nabobalis (approved):

.github/workflows/scheduled_builds.yml
```diff
@@ -16,3 +16,5 @@ jobs:
       - run: gh workflow run ci.yml --ref main
       - run: gh workflow run ci.yml --ref 3.1
       - run: gh workflow run ci.yml --ref 3.0
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
Cadair, line 20: I am surprised the built in actions token isn't sufficient for this? Do we really need to add a new one?

ConorMacBride: That's what the docs say: https://docs.github.com/en/actions/using-workflows/using-github-cli-in-workflows

Cadair: But there already is a default token on all actions no?

Cadair: Basically I am saying, we need this line, but to do nothing else.

ConorMacBride: You're right! I hadn't realised there was a default token. It has write permissions for Actions on main: https://github.com/sunpy/sunpy/runs/5791457157?check_suite_focus=true#step:1:14 So I shall merge.

PR #6017, opened 2022-03-31
Adds a function to get linearly spaced astropy time object in TimeRange

### LINKED ISSUE #5377: Functionality to pass a dt (sampling frequency) to TimeRange and it would return list of astropy time objects
### Description

It would be nice if you could pass a start and end time and a sampling frequency to TimeRange and it would return a list of astropy time objects. 

The list could be an attribute of the TimeRange class

E.g. an equivalent to something like np.linspace but for a timerange

Krish2208:
## PR Description

Adds a function `range` to TimeRange Class which returns a linearly spaced astropy time object, as mentioned in the issue #5377 

Fixes #5377 

## PR Checklist

- [ ] Changes pass pytest style unit tests (and `pytest` passes).
- [ ] Changes include any required corresponding changes to the documentation
- [ ] I have added a `Fixes #XXXX -` or `Closes #XXXX -` comment to auto-close the issue that your PR addresses

### DIFF
--- a/changelog/6017.feature.rst
+++ b/changelog/6017.feature.rst
@@ -1,1 +1,1 @@
-Added a range function to TimeRange class that returns a linearly spaced time object.
+Added a range function to `TimeRange` that returns a linearly spaced time object.
--- a/changelog/6017.feature.rst
+++ b/changelog/6017.feature.rst
@@ -1,1 +1,0 @@
-Added a range function to `.TimeRange` that returns a linearly spaced time object.
--- a/changelog/6017.feature.rst
+++ b/changelog/6017.feature.rst
@@ -1,0 +1,1 @@
+Added a range function to TimeRange class that returns a linearly spaced time object.
--- a/changelog/6017.feature.rst
+++ b/changelog/6017.feature.rst
@@ -1,1 +1,1 @@
-Added a range function to `TimeRange` that returns a linearly spaced time object.
+Added a range function to `.TimeRange` that returns a linearly spaced time object.
--- a/sunpy/time/tests/test_timerange.py
+++ b/sunpy/time/tests/test_timerange.py
@@ -302,7 +302,7 @@ def test_range():
     linspace_range2 = tr2.range(TimeDelta(60*60, format='sec'))
     assert tr._t1 == linspace_range[0]
     assert tr._t1 == linspace_range2[0]
-    
+
     # Check time in the range
     assert tr._t1 + TimeDelta(5, format='sec') in linspace_range
     assert tr._t1 + TimeDelta(61, format='sec') not in linspace_range
--- a/sunpy/time/tests/test_timerange.py
+++ b/sunpy/time/tests/test_timerange.py
@@ -292,3 +292,10 @@ def test_intersects():
     tr2 = sunpy.time.TimeRange('2020-01-01', '2020-01-04')
     assert tr1.intersects(tr2)
     assert tr2.intersects(tr1)
+
+def test_range():
+    tr = sunpy.time.TimeRange('2020-01-01', '2020-01-02')
+    linspace_range = tr.range(2)
+    assert tr._t1 == linspace_range[0]
+    assert tr._t2 == linspace_range[1]
+    assert len(linspace_range.value)==2
--- a/sunpy/time/tests/test_timerange.py
+++ b/sunpy/time/tests/test_timerange.py
@@ -292,28 +292,3 @@ def test_intersects():
     tr2 = sunpy.time.TimeRange('2020-01-01', '2020-01-04')
     assert tr1.intersects(tr2)
     assert tr2.intersects(tr1)
-
-
-def test_range():
-    t = Time('2020-01-02')
-    tr = sunpy.time.TimeRange('2020-01-01T00:00:00', '2020-01-01T00:01:00')
-    tr2 = sunpy.time.TimeRange('2020-01-01T00:00:00', '2020-01-05T00:00:00')
-    linspace_range = tr.range(TimeDelta(1, format='sec'))
-    linspace_range2 = tr2.range(TimeDelta(60*60, format='sec'))
-    assert tr._t1 == linspace_range[0]
-    assert tr._t1 == linspace_range2[0]
-
-    # Check time in the range
-    assert tr._t1 + TimeDelta(5, format='sec') in linspace_range
-    assert tr._t1 + TimeDelta(61, format='sec') not in linspace_range
-    assert t in linspace_range2
-
-    # Check the length of the range
-    assert len(linspace_range.value) == 60
-    assert len(linspace_range2.value) == 96
-
-    # Check the raised exceptions
-    with pytest.raises(ValueError):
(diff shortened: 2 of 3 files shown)


nabobalis:
Can I get an example of how someone uses this?

Krish2208:
> Can I get an example of how someone uses this?



```
tr = TimeRange('2022/03/31 12:00', '2022/03/31 13:00')
range_list = tr.range_list(5)
print(range_list)
```

This would print a list of 5 linearly spaced astropy time objects

`[<Time object: scale='utc' format='isot' value=2022-03-31T12:00:00.000>, <Time object: scale='utc' format='isot' value=2022 03-31T12:15:00.000>, <Time object: scale='utc' format='isot' value=2022-03-31T12:30:00.000>, <Time object: scale='utc' format='isot' value=2022-03-31T12:45:00.000>, <Time object: scale='utc' format='isot' value=2022-03-31T13:00:00.000>]`

sunpy/time/timerange.py
```diff
@@ -460,3 +461,20 @@ def intersects(self, other):
             int_second = self
 
         return int_second.start <= int_first.end
+
+    def range_list(self, frequency):
```
nabobalis, line 465: Could we not name this range?

sunpy/time/timerange.py
```diff
@@ -460,3 +461,20 @@ def intersects(self, other):
...
+        """
+        Return a Linearly Spaced List of Astropy Time Objects between the passed TimeRange.
```
nabobalis, line 467:
```suggestion
        Return a linearly spaced time range.
```

sunpy/time/timerange.py
```diff
@@ -460,3 +461,20 @@ def intersects(self, other):
...
+
+        Parameters
+        ----------
+        frequency : `int`
+            The number of objects needed
```
nabobalis, line 472:
```suggestion
        steps : `int`
            The number of steps between the start and end time.
```

sunpy/time/timerange.py
```diff
@@ -460,3 +461,20 @@ def intersects(self, other):
...
+        """
+        if frequency<=0:
+            raise ValueError("The number of objects needed cannot be 0 or less")
```
nabobalis, line 475: I am not sure if we need this check?

sunpy/time/timerange.py
```diff
@@ -460,3 +461,20 @@ def intersects(self, other):
...
+        end = self._t2
+        start = self._t1
+        delta = (end-start)
```
nabobalis, line 478: I don't think we need these defined as variables?

sunpy/time/timerange.py
```diff
@@ -460,3 +461,20 @@ def intersects(self, other):
...
+        time_list = list(start + delta * np.linspace(0,1,frequency))
```
nabobalis, line 479: It should be possible to return one astropy.time.Time object that has a range of values, I would personally prefer that instead of a list.

nabobalis:
> > Can I get an example of how someone uses this?
> 
> ```
> tr = TimeRange('2022/03/31 12:00', '2022/03/31 13:00')
> range_list = tr.range_list(5)
> print(range_list)
> ```
> 
> This would print a list of 5 linearly spaced astropy time objects
> 
> `[<Time object: scale='utc' format='isot' value=2022-03-31T12:00:00.000>, <Time object: scale='utc' format='isot' value=2022 03-31T12:15:00.000>, <Time object: scale='utc' format='isot' value=2022-03-31T12:30:00.000>, <Time object: scale='utc' format='isot' value=2022-03-31T12:45:00.000>, <Time object: scale='utc' format='isot' value=2022-03-31T13:00:00.000>]`

Thanks!

nabobalis:
Just now to add some unit tests and a changelog entry!

changelog/6017.feature.rst
```diff
@@ -0,0 +1 @@
+Added a range function to TimeRange class that returns a linearly spaced time object.
```
nabobalis, line 1:
```suggestion
Added a range function to 	`.TimeRange` that returns a linearly spaced time object.
```

sunpy/time/tests/test_timerange.py
```diff
@@ -292,3 +292,10 @@ def test_intersects():
     tr2 = sunpy.time.TimeRange('2020-01-01', '2020-01-04')
     assert tr1.intersects(tr2)
     assert tr2.intersects(tr1)
+
+def test_range():
```
nabobalis, line 296: Can I get a few more tests, with larger ranges, what happens if you pass in a negative one or 0 even.

nabobalis:
I looked over the original issue, and the request was for a sampling frequency. 

I wonder if passing a astropy.time.TimeDelta would be a better idea than a int?

Krish2208:
> I looked over the original issue, and the request was for a sampling frequency.
> 
> I wonder if passing a astropy.time.TimeDelta would be a better idea than a int?

So we need a something like 
When we pass `TimeDelta(300, format="sec")` into the function, we should get a time object with 5 min difference between consecutive values?

nabobalis:
> So we need a something like When we pass `TimeDelta(300, format="sec")` into the function, we should get a time object with 5 min difference between consecutive values?

Yeah,  that sounds right.
 
> Shall I add another method for this or do I need to make changes to the `range()` method?

I think it should be a change to the range function.

changelog/6017.feature.rst
```diff
@@ -0,0 +1 @@
+Added a range function to `TimeRange` that returns a linearly spaced time object.
```
nabobalis, line 1:
```suggestion
Added a range function to `.TimeRange` that returns a linearly spaced time object.
```

dstansby (dismissed):
Thanks for working on this; I'm personally not sure we should implement this ourselves, but instead direct users using `np.arange(TimeRange.start, TimeRange.end, delta)`. This:
- Avoids us creating new API
- Is a more general solution, that users can use with any two times

As such I'm going to block this for now until there's been a wider developer discussion. I'd say if someone else agrees with me we should close this without merging (sorry), and turn https://github.com/sunpy/sunpy/issues/5377 into a request to add `np.arange(TimeRange.start, TimeRange.end, delta)` to the docs somewhere.

I'm willing to be swayed on my thoughts though if other developers think we should add this?

nabobalis:
I have to agree with dstansby, having thought about it and shown that anyone can do this with `np.arange(TimeRange.start, TimeRange.end, delta)`, I think maybe a mention in the documentation of time range as an example would be a better idea than a new function.

We can turn this PR into that?

Krish2208:
I added an example can you please check it.
Sorry for the delay.

sunpy/time/timerange.py
```diff
@@ -65,6 +65,16 @@ class TimeRange:
     True
     >>> time2 in time_range
     False
+    >>> import numpy as np
```
nabobalis, line 68:
```suggestion

    >>> import numpy as np
```

nabobalis (approved):

nabobalis:
Thanks for the PR @Krish2208

PR #6023, opened 2022-04-01
Use astropy and matplotlib nightly feeds for devdeps CI

### LINKED ISSUE #6019: Use Matplotlib nightly wheels in devdeps testing
### Describe the feature

I think Matplotlib now have nightly wheels: https://github.com/matplotlib/matplotlib/pull/22733#issuecomment-1085269320. I haven't read the details, but thought I'd leave a note here that it would be good to use the nightly wheels to save a bit of time (and energy!) in our `devdeps` CI by using these instead of compiling from source.

### Proposed solution

_No response_

ConorMacBride:
This PR updates the tox configuration to search for the latest pre-release from the Astropy and Matplotlib nightly PyPI repositories when installing `astropy` and `matplotlib` in any of the `-devdeps` environments. This should significantly speed up the devdeps cron test on GitHub Actions and the devdeps figure tests on CircleCI. The remaining dev packages (`asdf` and `mpl-animators`) seem to be pure python so should't be a significant bottleneck.

The Astropy feed is at: https://dev.azure.com/astropy-project/astropy/_artifacts/feed/nightly/PyPI/astropy/5.1.dev781/overview

The Matplotlib feed is at: https://anaconda.org/scipy-wheels-nightly/matplotlib

Closes #6019

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,9 @@ jobs:
           - openjpeg
       envs: |
         - macos: py38
-        - windows: py310
+        - windows: py39
+        - linux: base_deps
+        - linux: py39-devdeps
         - linux: py38-oldestdeps
         - linux: py39-conda
           libraries: ''
@@ -83,8 +85,7 @@ jobs:
         apt:
           - libopenjp2-7
       envs: |
-        - linux: base_deps
-        - linux: py39-devdeps
+        - windows: py310
 
   publish:
     # Build wheels when pushing to any branch except main
--- a/tox.ini
+++ b/tox.ini
@@ -31,10 +31,14 @@ setenv =
     devdeps,build_docs,online: HOME = {envtmpdir}
     HIDE_PARFIVE_PROGESS = True
     NO_VERIFY_HELIO_SSL = 1
+    devdeps: PIP_EXTRA_INDEX_URL = https://pkgs.dev.azure.com/astropy-project/astropy/_packaging/nightly/pypi/simple/ https://pypi.anaconda.org/scipy-wheels-nightly/simple
+pip_pre =
+    devdeps: true
+    !devdeps: false
 deps =
     # devdeps is intended to be used to install the latest developer version of key dependencies.
-    devdeps: git+https://github.com/astropy/astropy
-    devdeps: git+https://github.com/matplotlib/matplotlib
+    devdeps: astropy
+    devdeps: matplotlib
     devdeps: git+https://github.com/asdf-format/asdf
     devdeps: git+https://github.com/sunpy/mpl-animators
     # Oldest deps we pin against
--- a/tox.ini
+++ b/tox.ini
@@ -32,13 +32,10 @@ setenv =
     HIDE_PARFIVE_PROGESS = True
     NO_VERIFY_HELIO_SSL = 1
     devdeps: PIP_EXTRA_INDEX_URL = https://pkgs.dev.azure.com/astropy-project/astropy/_packaging/nightly/pypi/simple/ https://pypi.anaconda.org/scipy-wheels-nightly/simple
-pip_pre =
-    devdeps: true
-    !devdeps: false
 deps =
     # devdeps is intended to be used to install the latest developer version of key dependencies.
-    devdeps: astropy
-    devdeps: matplotlib
+    devdeps: astropy>=0.0.dev0
+    devdeps: matplotlib>=0.0.dev0
     devdeps: git+https://github.com/asdf-format/asdf
     devdeps: git+https://github.com/sunpy/mpl-animators
     # Oldest deps we pin against


nabobalis (approved):

ConorMacBride:
Will check the errors in more detail tomorrow, but the correct pre-release astropy and matplotlib versions have been installed: https://output.circle-artifacts.com/output/job/3acfb9fb-115d-41dd-993c-32fc68ac3d2f/artifacts/0/.tmp/py39-figure-devdeps/figure_test_images/figure_version_info.txt

nabobalis:
If its a good speed up, is it time to run the devdeps on a PR now?

tox.ini
```diff
@@ -31,10 +31,14 @@ setenv =
     devdeps,build_docs,online: HOME = {envtmpdir}
     HIDE_PARFIVE_PROGESS = True
     NO_VERIFY_HELIO_SSL = 1
+    devdeps: PIP_EXTRA_INDEX_URL = https://pkgs.dev.azure.com/astropy-project/astropy/_packaging/nightly/pypi/simple/ https://pypi.anaconda.org/scipy-wheels-nightly/simple
+pip_pre =
+    devdeps: true
+    !devdeps: false
```
nabobalis, line 37: The only issue I can imagine is that it might pull in more devdeps than we expect from pypi?

ConorMacBride: This tox option just adds `--pre` to the pip command. I'd need to check if that applies to the dependencies of explicitly requested packages also. It only has an effect when installing the tox reps and definitely has no effect when installing sunpy.

ConorMacBride:
Also note that when we request the git repository directly we are guaranteed to install the latest version. If astropy or matplotlib were to stop publishing to the URLs added to tox.ini, the sunpy devdeps tests would always run against the last published nightly release and eventually start running against the latest release on the default PyPI repository, and no error would be produced.

tox.ini
dstansby: > The only issue I can imagine is that it might pull in more devdeps than we expect from pypi?

This might be a good thing? Perhaps we can see how it goes for a bit, and we get more errors than expected have a re-think?

dstansby:
There doesn't seem to be a GH actions devdeps build - am I missing something?

nabobalis:
> There doesn't seem to be a GH actions devdeps build - am I missing something?

Needs to be added to the PR CI, for now we have the figure devdeps one.

tox.ini
ConorMacBride: It is installing pre-releases of astropy and matplotlib dependents: `greenlet`, `numpy`, `pandas` and `scipy`. And the pre-release `numpy` seems to be causing the test fails.

nabobalis: If its a pre-release numpy causing the issue, I feel like we should inform someone upstream?

tox.ini
```diff
@@ -31,10 +31,11 @@ setenv =
     devdeps,build_docs,online: HOME = {envtmpdir}
     HIDE_PARFIVE_PROGESS = True
     NO_VERIFY_HELIO_SSL = 1
+    devdeps: PIP_EXTRA_INDEX_URL = https://pkgs.dev.azure.com/astropy-project/astropy/_packaging/nightly/pypi/simple/ https://pypi.anaconda.org/scipy-wheels-nightly/simple
 deps =
     # devdeps is intended to be used to install the latest developer version of key dependencies.
-    devdeps: git+https://github.com/astropy/astropy
-    devdeps: git+https://github.com/matplotlib/matplotlib
+    devdeps: astropy>=0.0.dev0
+    devdeps: matplotlib>=0.0.dev0
```
nabobalis, line 38: That is some magic.

ConorMacBride: Yeah! Found here: https://pip.pypa.io/en/stable/cli/pip_install/#pre-release-versions

nabobalis (approved):

ConorMacBride:
Given https://github.com/sunpy/sunpy/pull/5939#discussion_r833466590, maybe @Cadair should review before merging? The devdeps test only takes 5 min now, and in comparison to the py310 windows test which takes 15 min, seems fine to run in a PR. (Although we should look into what windows wheels are missing.)

nabobalis:
> Given https://github.com/sunpy/sunpy/pull/5939#discussion_r833466590, maybe @Cadair should review before merging? The devdeps test only takes 5 min now, and in comparison to the py310 windows test which takes 15 min, seems fine to run in a PR. (Although we should look into what windows wheels are missing.)

What if we swap them around, that could be faster?

ConorMacBride:
Yeah, should probably move py310 testing to the cron stage. When we remove the scikit-image pin we'll have py310 wheels and no longer have to build it from source. Also, `astropy-healpix` and `reproject` have platform specific py39 wheels but none for py310 yet, although they are probably not slowing things down?

nabobalis:
I was wondering if we just switched to a linux 310 job and ran that on a PR instead of having a cron stage for just that?

ConorMacBride:
The pinned scikit-image doesn't have py310 linux wheels either.

nabobalis:
> As the person who maintains the Astropy nightly wheels, this is my major concern. We would have no way of knowing that the version of astropy we are testing against was suddenly a month old...

Are they not updated automatically?

Cadair:
Yeah, but if something goes wrong in the build pipeline or whatever then it can break and nobody (me) notices for a while.

.github/workflows/ci.yml
```diff
@@ -43,7 +43,9 @@ jobs:
           - openjpeg
       envs: |
         - macos: py38
-        - windows: py310
+        - windows: py39
+        - linux: base_deps
+        - linux: py39-devdeps
```
Cadair, line 48: I still maintain we don't need or want to be running this on PRs. For the majority of contributions it's not a relevant test.

.github/workflows/ci.yml
```diff
@@ -83,8 +85,7 @@ jobs:
         apt:
           - libopenjp2-7
       envs: |
-        - linux: base_deps
-        - linux: py39-devdeps
+        - windows: py310
```
Cadair, line 88:
```suggestion
        - linux: py310
```

Windows is slower anyway skimage issues aside.

Cadair (approved):

.github/workflows/ci.yml
nabobalis: I will fix this in my open PR

.github/workflows/ci.yml
nabobalis: I will fix this in my open PR

PR #6028, opened 2022-04-05
fix doc build

nabobalis:
Should fix a few issues I introduced into the doc build.


changelog/6011.bugfix.rst
```diff
@@ -1 +1 @@
-Fixed `.system_info` so it returns the extra group when an optional dependency is missing.
+Fixed `~sunpy.util.sysinfo.system_info` so it returns the extra group when an optional dependency is missing.
```
nabobalis, line 1:
```suggestion
Fixed `sunpy.system_info` so it returns the extra group when an optional dependency is missing.
```

dstansby (approved):

PR #6041, opened 2022-04-06
Add environment for release token

### LINKED ISSUE #6040: PyPi uploading not working on GH actions
### Describe the bug

See https://github.com/sunpy/sunpy/runs/5848591278?check_suite_focus=true - it looks like an authentication issue.

### To Reproduce

import sunpy

sunpy.map.Map(...)
etc

### What happened?

_No response_

### Expected behavior

_No response_

### Screenshots

_No response_

### System Details

_No response_

### Installation method

_No response_

Cadair:
I think this should fix #6040

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,6 +95,7 @@ jobs:
         github.event_name == 'workflow_dispatch' )
     needs: [test]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@main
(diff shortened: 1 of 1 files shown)


ConorMacBride:
Environment is only available at the job level (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions). Will need to update the workflow upstream

Cadair:
Yeah I can mint a new token and do that.

PR #6046, opened 2022-04-07
Fixed outdated statement in map-rotation example

ayshih:
`GenericMap.rotate()` does compound rotations (now).  Maybe we should fact-check the rest of our examples...


nabobalis (approved):

ayshih:
I called the statement "out-dated", but I can't seem to find a SunPy version when it was ever true.  So the statement might have simply been wrong at first writing.

nabobalis:
Doc fails are examples and unrelated to the PR.

PR #6049, opened 2022-04-07
Don't bother running online CI when publishing

dstansby:
Publishing doesn't require these builds to pass, so we might as well save ourselves ~1h of actions runtime by not running them when publishing. I think I got the if statement right, but would appreciate someone else checking...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,9 +52,7 @@ jobs:
 
   online:
     # Don't run online tests for tagged releases
-    if: |
-      github.event_name == 'pull_request' &&
-          !startsWith(github.event.ref, 'refs/tags/v')
+    if: !startsWith(github.event.ref, 'refs/tags/v')
     needs: [core]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@main
     with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,8 +51,7 @@ jobs:
           libraries: ''
 
   online:
-    # Don't run online tests for tagged releases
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -49,6 +49,10 @@ jobs:
           libraries: ''
 
   online:
+    # Don't run online tests for tagged releases
+    if: |
+      github.event_name == 'pull_request' &&
```
Cadair, line 54: I am going to have to delegate checking this to @ConorMacBride but I think you just need the not a tag check?

.github/workflows/ci.yml
```diff
@@ -49,6 +49,10 @@ jobs:
...
+          !startsWith(github.event.ref, 'refs/tags/v')
```
ConorMacBride, line 55: Including the PR check would mean online tests would only run during PRs, and not during merges or crons.
```suggestion
    if: !startsWith(github.event.ref, 'refs/tags/v')
```

ConorMacBride (approved):

.github/workflows/ci.yml
```diff
@@ -49,6 +49,8 @@ jobs:
           libraries: ''
 
   online:
+    # Don't run online tests for tagged releases
```
ConorMacBride, line 52: Comments must not be allowed there?
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -51,6 +51,7 @@ jobs:
           libraries: ''
 
   online:
+    if: "!startsWith(github.event.ref, 'refs/tags/v')"
```
nabobalis, line 54: Google says you have to wrap the entire thing in qoutes for it to work.

PR #6054, opened 2022-04-08
Test changelog bot

ConorMacBride:
## PR Description
This PR tests the changelog bot and show never be merged.

### Changelog [bugfix]
```
Added automatic conversion of some common but non-standard unit strings in CDF
files to astropy unit objects. If sunpy does not recognise the unit string for
a particular column, units of ``u.dimensionless_unscaled`` are applied to that
column and a warning raised. Testing.

If you think a given unit should not be dimensionless and support should be
added for it in sunpy, please raise an issue at
https://github.com/sunpy/sunpy/issues. Test.

But if someone were to edit.
```

### Changelog [bugfix]

You can include text here if you like and it'll not be in the changelog. No change.

```
I fixed another bug. Updated.
```

## A new feature

This is a new feature

#### Changelog[feature]
```
A `feature` was also added.
```


ConorMacBride:
Need to fix the regex to allow `` ` `` inside the changelog.

ConorMacBride:
Nice!

nabobalis:
That is cool.

ConorMacBride:
The problem is, the commit triggers the CI to run (except GitHub Actions as the `GITHUB_TOKEN` used to commit doesn't allow actions to be triggered to prevent recursion). I wonder is it enough to add `[skip ci]` to the commit message.

nabobalis:
I think maybe a skip ci should be added to the changelog tweak, I think RTD ignores it anyway?

ConorMacBride:
That last edit should have deleted `changelog/6054.bugfix.rst`

nabobalis:
Are these fixes to be done in the actions script?

ConorMacBride:
Yes

ConorMacBride:
Some of the issues are not fixed:

1. Allow `` ` `` in changelog entry
2. Ensure no `\r` characters are in the changelog entry (I wasn't able to recreate this bug locally so this fix may not work. For some reason it was doing `\r\n` in GitHub Actions `ubuntu-latest` and not on macOS.)
3. Make the bot appear as SunPyBot instead of GitHub Actions

Issues 1 and 2 are fixed when I test locally. I might try debugging this on Ubuntu tomorrow rather than macOS.

Issue 3 needs to have the SunPyBot email address set correctly. Or just keep the GitHub Actions user rather than spoofing the real SunPyBot user account?

nabobalis:
> Issues 1 and 2 are fixed when I test locally. I might try debugging this on Ubuntu tomorrow rather than macOS.

Change the bot to use a mac os image  (I guess it would be slower?)?

> Issue 3 needs to have the SunPyBot email address set correctly. Or just keep the GitHub Actions user rather than spoofing the real SunPyBot user account?

I would prefer if it looked sunpy-like but if its a pain, leave it as is.

Cadair:
I added bot@sunpy.org to the @SunPyBot github account, so commits should show up as that user now. Is the plan to make this work well enough and then rewrite the UI as discussed in the original PR or what?

ConorMacBride:
> I added [bot@sunpy.org](mailto:bot@sunpy.org) to the @SunPyBot github account, so commits should show up as that user now. Is the plan to make this work well enough and then rewrite the UI as discussed in the original PR or what?

Thanks. To change it to the bot command UI, the workflow and script would need to be mostly  rewritten, although that can be done.

I think this creates a lot of noise in the [Actions tab](https://github.com/sunpy/sunpy/actions) as every time a PR is tagged etc this workflow will run. If we switch to the comment trigger it will mean a workflow will be run *every time* someone comments. Another thing I don't like is that when the workflow commits the changelog, it clears the existing checks from the PR and only the four (skipped figure-tests, changelog: found, RTD, and pre-commit.ci) checks appear. Maybe it would be better to just ask people to name the changelogs `pr.bugfig.rst` (or something) and then a workflow runs on `push` to the `main` branch that renames the files to have the merged PR number? (Also, I just disabled the changelog workflow as I noticed that the Python script being run was the version from the PR, so it gives arbitrary code execution with access to the writeable `GITHUB_TOKEN`.)

Cadair:
I am not that worried about the noise on the actions tab tbh, you can filter by workflow so you can just ignore it etc.

The thing I really like about an actions based workflow is not getting the number right it's not having to commit to the PR at all, especially useful when any maintainer can add a changelog entry by just commenting etc.

Not having the CI run on the changlog commit is fine imo, before merging it it's up to the maintainer to check the CI status on the previous commits.

If people agree with you and this isn't the way to go the alternative is the baldrick esque route where a GH app does the work.

nabobalis:
I am going to close this for now. Reopen it when we are ready to give it another go!

PR #6056, opened 2022-04-09
Adds columns kwarg to plot methods

### LINKED ISSUE #5729: Add `column` kwarg to all timeseries source `plot` methods
In sunpy 3.1 I added the ability to plot a subset of the columns in a `GenericTimeseries`: https://docs.sunpy.org/en/stable/api/sunpy.timeseries.GenericTimeSeries.html#sunpy.timeseries.GenericTimeSeries.plot

It would be good to roll a similar change out to all the timeseries sources.

Krish2208:
## PR Description
Adds column kwarg to timeseries source plot methods. 
Fixes #5729


sunpy/timeseries/sources/eve.py
```diff
@@ -78,15 +80,18 @@ def plot(self, axes=None, **kwargs):
...
+            axes = self.to_dataframe()[column].plot(subplots=True, sharex=True, **kwargs)
         for i, ax in enumerate(axes):
             if predefined_axes:
-                ax.plot(self._data[self._data.columns[i]],
-                        color=colors[i % len(colors)],
-                        label=self._data.columns[i])
+                if i in column:
+                    ax.plot(self._data[self._data.columns[i]],
+                            color=colors[i % len(colors)],
+                            label=self._data.columns[i])
```
nabobalis, line 94: How does this work?

Krish2208: Sorry, my bad I messed up there. I am on it...

Krish2208:
Sorry, I was unable to work due to certain circumstances.
Earlier I forget to render the names according to the column so I added that too.

sunpy/timeseries/sources/eve.py
```diff
@@ -74,22 +76,25 @@ def plot(self, axes=None, **kwargs):
...
         predefined_axes = False
+        if column is None:
+            column = self.to_dataframe().columns
         if isinstance(axes, np.ndarray):
             predefined_axes = True
         elif axes is None:
-            axes = self.to_dataframe().plot(subplots=True, sharex=True, **kwargs)
-        for i, ax in enumerate(axes):
+            axes = self.to_dataframe()[column].plot(subplots=True, sharex=True, **kwargs)
+        for i, name in enumerate(self.to_dataframe()[column]):
```
nabobalis, line 90: is `name` used somewhere?

sunpy/timeseries/sources/eve.py
```diff
@@ -74,22 +76,25 @@ def plot(self, axes=None, **kwargs):
...
         """
         self._validate_data_for_plotting()
-        names = ("Flux \n 0.1-7nm", "Flux \n 18nm", "Flux \n 26nm", "Flux \n 30nm", "Flux \n 36nm")
+        column_names = {"QD": "Flux \n 0.1-7nm", "CH_18": "Flux \n 18nm",
+                        "CH_26": "Flux \n 26nm", "CH_30": "Flux \n 30nm", "CH_36": "Flux \n 36nm"}
         colors = ('tab:blue', 'tab:orange', 'tab:green', 'tab:red', 'tab:purple',
                   'tab:brown', 'tab:pink', 'tab:gray', 'tab:olive', 'tab:cyan')
         predefined_axes = False
+        if column is None:
+            column = self.to_dataframe().columns
```
nabobalis, line 85: I am not sure if this is done in all the other functions but we should cast to a dataframe once.

sunpy/timeseries/sources/fermi_gbm.py
```diff
@@ -59,14 +59,16 @@ class GBMSummaryTimeSeries(GenericTimeSeries):
     # Class attribute used to specify the source class of the TimeSeries.
     _source = 'gbmsummary'
 
-    def plot(self, axes=None, **kwargs):
+    def plot(self, axes=None, column=None, **kwargs):
```
nabobalis, line 62: The original issue says its `columns` and not `column`, can you check which one is corrrect?

nabobalis:
It will be easier if you try to run the time series tests or even the sample data example (I think we plot most of the timeseries there). It should make debugging easier.

sunpy/timeseries/sources/fermi_gbm.py
Krish2208: In the title of the issue it is `column`.

sunpy/timeseries/sources/eve.py
Krish2208: Sorry, I don't follow, what exactly is required?

sunpy/timeseries/sources/eve.py
Krish2208: I thought of using it to set the label but I was uncertain about it.
Okk I will make the changes accordingly

Krish2208:
I don't know what's the problem but when I run the tests it takes the installed `sunpy` package and not the local development source.

sunpy/timeseries/sources/fermi_gbm.py
nabobalis: Can you check this again, what if the title was wrong?

sunpy/timeseries/sources/eve.py
nabobalis: There are multiple calls to convert the time series to a data frame. This should only happen once.

nabobalis:
> I don't know what's the problem but when I run the tests it takes the installed `sunpy` package and not the local development source.

Did you install the local development version?

Krish2208:
Ohh I didn't do it earlier, now I did it, thank you.

sunpy/timeseries/sources/rhessi.py
```diff
@@ -192,12 +196,14 @@ def peek(self, title="RHESSI Observing Summary Count Rate", **kwargs):
         ----------
         title : `str`
             The title of the plot. Defaults to "RHESSI Observing Summary Count Rate".
+        columns : list[str], optional
+            If provided, only plot the specified columns.
```
nabobalis, line 200: This is in the wrong order.

nabobalis:
Do all the plots still look the same then?

I would also like some unit tests just to check the columns code is at least passed and doesn't error. They do not need to be new figure tests.

Krish2208:
> Do all the plots still look the same then?

I didn't add columns to some of the plot methods as

- `NoRHTimeSeries` it only have one column, so I thought that `columns` kwarg won't be neccessary.
- `NOAAPredictIndicesTimeSeries` & `NOAAIndicesTimeSeries` because right now also they are plotting specific columns.

If any changes are required, I will do that.

> I would also like some unit tests just to check the columns code is at least passed and doesn't error. They do not need to be new figure tests.

Ok, I will write some unit tests.

nabobalis:
> I didn't add columns to some of the plot methods as
> 
> * `NoRHTimeSeries` it only have one column, so I thought that `columns` kwarg won't be neccessary.
> * `NOAAPredictIndicesTimeSeries` & `NOAAIndicesTimeSeries` because right now also they are plotting specific columns.
> 

I don't see the harm in adding the keyword and just doing nothing with it.

changelog/6056.feature.rst
```diff
@@ -0,0 +1 @@
+Adds `columns` kwarg to plot method to timeseries sources.
```
nabobalis, line 1:
```suggestion
Added a ``columns`` keyword to each plot method for all `sunpy.timeseries.GenericTimeSeries` sources.
```

sunpy/timeseries/sources/eve.py
```diff
@@ -57,14 +57,16 @@ class ESPTimeSeries(GenericTimeSeries):
...
-    def plot(self, axes=None, **kwargs):
+    def plot(self, axes=None, columns=None, **kwargs):
         """
         Plots the EVE ESP Level 1 timeseries data.
 
         Parameters
         ----------
         axes : numpy.ndarray of `matplotlib.axes.Axes`, optional
             The axes on which to plot the TimeSeries.
+        columns : list[str], optional
```
nabobalis, line 68: Note to self: Check how this renders on RTD.

sunpy/timeseries/sources/eve.py
```diff
@@ -215,8 +222,8 @@ def peek(self, column=None, **kwargs):
 
         Parameters
         ----------
-        column : `str`, optional
-            The column to display. Defaults to ``None``, so it will display all.
+        columns : `str`, optional
```
nabobalis, line 225: So this is a string but the plot method wants a list of strings, we should unify this somehow.

sunpy/timeseries/sources/goes.py
```diff
@@ -146,7 +146,7 @@ def observatory(self):
         return None
 
     @peek_show
-    def peek(self, title="GOES X-ray flux", **kwargs):
+    def peek(self, columns=["xrsa", "xrsb"], title="GOES X-ray flux", **kwargs):
```
nabobalis, line 149: Why could this be set to None and then have it plot the two columns by default?

sunpy/timeseries/sources/noaa.py
```diff
@@ -67,6 +67,8 @@ def plot(self, axes=None, plot_type='sunspot SWO', **kwargs):
             The axes on which to plot the TimeSeries.
         plot_type : {``"sunspot SWO"``, ``"sunspot RI"``, ``"sunspot compare"``, ``"radio"``, ``"geo"``}, optional
             The type of plot required. Defaults to ``"sunspot SWO"``.
+        columns : *
```
nabobalis, line 70: This shouldn't be a *

sunpy/timeseries/tests/test_timeseriesbase.py
```diff
@@ -951,6 +951,39 @@ def test_timeseries_array():
         ts = sunpy.timeseries.TimeSeries(data, {})
         assert isinstance(ts, sunpy.timeseries.GenericTimeSeries)
 
+# =============================================================================
+# Test Source Plot columns keyword argument
+# =============================================================================
+
+
```
nabobalis, line 958: Are these the only 4 sources that were missing?

nabobalis:
Can you fix the failing tests?

nabobalis (approved):

nabobalis:
Thanks for the PR @Krish2208, sorry it took so long!

Issue #5924, opened 2022-02-27
Move cibuildwheel config to `pyproject.toml`


ConorMacBride:
### Describe the feature

Currently the specific wheels that cibuildwheel generates are configured within the CI pipeline files:
https://github.com/sunpy/sunpy/blob/main/azure-pipelines.yml#L3-L4
https://github.com/sunpy/sunpy/blob/main/azure-pipelines.yml#L139-L142

I think it would be good to move as much cibuildwheel configuration as possible out of the CI configuration files, so it would make the build pipeline more reproducible across CI platforms. This way we could run a basic `cibuildwheel` command within a Linux or MacOS runner and it would just build everything we need.

This is also to reduce the complexity of the GitHub Actions reusable workflow, as the lack of support for lists as a workflow input makes it challenging to customise the cibuildwheel environment on a per-job basis. (The only way I could achieve a simple "for loop" was this: https://github.com/OpenAstronomy/github-actions-workflows/pull/4/files#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7R42)

### Proposed solution

Instead of using CIBW env vars, move these to `pyproject.yml` as documented here: https://cibuildwheel.readthedocs.io/en/stable/options/#configuration-file

ConorMacBride:
I opened this issue before the current GitHub Action workflows were developed as I thought it would make developing the workflows simpler, however the workflow was subsequently designed to work without the solution proposed in this issue. 

Anyway, there doesn't seen to be a straightforward way to implement this (as we saw in #6063). So I shall close, but feel free to reopen if this is something you'd like.

PR #6063, opened 2022-04-12
Move some cibuildwheel configuration to pyproject.toml

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/5924. I don't think there's an easy way to move the platform/python configuration because of the way we're using the OpenAstronomy actions/the way those actions are set up, but happy to  be corrected on that @ConorMacBride ?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,8 +97,6 @@ jobs:
     needs: [test]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@main
     with:
-      test_extras: 'dev'
-      test_command: 'pytest -p no:warnings --doctest-rst -m "not mpl_image_compare" --pyargs sunpy'
       submodules: false
       targets: |
         - cp3{8,9,10}-manylinux*_x86_64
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -97,8 +97,6 @@ jobs:
     needs: [test]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@main
     with:
-      test_extras: 'dev'
-      test_command: 'pytest -p no:warnings --doctest-rst -m "not mpl_image_compare" --pyargs sunpy'
```
ConorMacBride, line 101: If we remove these, the `sdist` won't be tested. However, if we keep these, cibuildwheel will use these over the values set in `pyproject.toml`.

I do like the idea of including the packaging test commands inside the package (for reproducible builds). If this is the goal we could add a custom entry into `pyproject.toml` and parse it ourselves in a separate job during the workflow run. (Or we could not prune `.github` in the manifest.)

ConorMacBride (changes requested):

PR #6064, opened 2022-04-12
Fixed scikit-image rotation of big-endian images

ayshih:
For a long time, our rotation code has crashed on images stored in big-endian byte order when using `scikit-image` for certain interpolation orders (1 and 3).  With @nabobalis pointing out that our HMI sample data is stored in big-endian byte order, I have finally fixed this bug.

This can be backported, but will have to be done manually.

### DIFF
--- a/sunpy/image/transform.py
+++ b/sunpy/image/transform.py
@@ -1,6 +1,7 @@
 """
 Functions for geometrical image transformation and warping.
 """
+import sys
 import numbers
 from functools import wraps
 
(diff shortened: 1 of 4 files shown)


sunpy/image/transform.py
```diff
@@ -304,6 +305,10 @@ def _rotation_skimage(image, matrix, shift, order, missing, clip):
         adjusted_image /= im_max
         adjusted_missing /= im_max
 
+    # Swap the byte order if it is non-native (e.g., big-endian on a little-endian system)
+    if adjusted_image.dtype.byteorder == ('>' if sys.byteorder == 'little' else '<'):
+        adjusted_image = adjusted_image.byteswap().newbyteorder()
```
Cadair, line 310: This does a full in mem copy right? That might want to be noted in the docstring?

ayshih: Note added

nabobalis (approved):

Cadair (approved):

PR #5981, opened 2022-03-21
Import ABC from collections.abc for Python 3.10 compatibility.

tirkarthi:
## PR Description

Ref : https://github.com/python/cpython/pull/23754

ABC aliases were removed from collections in favor of collections.abc

## PR Checklist

- [ ] I have followed the guidelines in the [Contributing document](https://docs.sunpy.org/en/latest/dev_guide/contents/newcomers.html)
- [ ] Changes follow the coding style of this project
- [ ] Changes have been formatted and linted
- [ ] Changes pass pytest style unit tests (and `pytest` passes).
- [ ] Changes include any required corresponding changes to the documentation
- [ ] Changes have been rebased and squashed to the minimal number (typically 1) of relevant commits
- [ ] Changes have a descriptive commit message with a short title
- [ ] I have added a `Fixes #XXXX -` or `Closes #XXXX -` comment to auto-close the issue that your PR addresses

PR #6066, opened 2022-04-12
Fix and tidy tests

ConorMacBride:
This PR tidies up and fixes some issues with the tests:

- [Tests should start with test](https://github.com/sunpy/sunpy/commit/7d1795031a7650ad4cadd237fd75335921106a71)
These tests were not being run

- [Fix duplicate test names](https://github.com/sunpy/sunpy/commit/1d8b6d1ccc6c70a3bf4d1d3df41321755d3da3d4)
MDI test had a duplicate name (renamed) and plotting test was a duplicate test (deleted)

- [One statement per pytest context manager](https://github.com/sunpy/sunpy/commit/b85700b766d8e9447cfc8ec04cb4c88f2206632d)
In these cases, the first statement was raising the error/warning and the subsequent statements were not being run

- [Remove redundant remote_data mark from fixtures](https://github.com/sunpy/sunpy/commit/77b865d1da9edcc8a7bb41e233a1d364becc20eb)
["Marks can only be applied to tests, having no effect on fixtures."](https://docs.pytest.org/en/6.2.x/mark.html)

- [Fix contour test](https://github.com/sunpy/sunpy/commit/1733a2e3c8ec13114f0d327a12b831117c83f188)
This test wasn't returning any values so the assert statement in the for loop was not running. I have moved the percentage from 100% to 50% and there are now values to check.

Most of these issues were found by running the offline tests but generating a coverage report for all files including the test files. This shows which tests and which parts are actually being run. The rest of the issues were found with regex.

There are also quite a few fixtures and functions in the test files which are not being used, although I didn't delete any in case they are useful in the future or they are needed for online tests etc. I think it would be useful to include the test files in the coverage as we can more easily spot issues where assertions are not running. [I see that matplotlib includes test files in the coverage for example.](https://codecov.io/gh/matplotlib/matplotlib/tree/main/lib/matplotlib/tests) What do you think?

### DIFF
--- a/sunpy/database/tests/test_database.py
+++ b/sunpy/database/tests/test_database.py
@@ -295,7 +295,7 @@ def test_tag_undo(database):
     assert len(entry.tags) == 1
 
 
-def remove_nonexisting_tag(database):
+def test_remove_nonexisting_tag(database):
     with pytest.raises(NoSuchTagError):
         database.remove_tag('foo')
 
@@ -367,7 +367,7 @@ def test_star_undo(database):
     assert entry.starred
 
 
-def unstar_entry(database):
+def test_unstar_entry(database):
     entry = DatabaseEntry()
     assert not entry.starred
     database.star(entry)
--- a/sunpy/database/tests/test_database.py
+++ b/sunpy/database/tests/test_database.py
@@ -296,8 +296,10 @@ def test_tag_undo(database):
 
 
 def test_remove_nonexisting_tag(database):
+    entry = DatabaseEntry()
+    database.add(entry)
     with pytest.raises(NoSuchTagError):
-        database.remove_tag('foo')
+        database.remove_tag(entry, 'foo')
 
 
 def test_remove_tag(filled_database):
--- a/sunpy/io/_fits.py
+++ b/sunpy/io/_fits.py
@@ -37,6 +37,7 @@ import sys
 import math
 import traceback
 import collections
+import collections.abc
 
 from astropy.io import fits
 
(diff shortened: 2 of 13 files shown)


nabobalis (approved):

nabobalis:
Looks like:
```
FAILED ../../.tox/py310/lib/python3.10/site-packages/sunpy/database/tests/test_database.py::test_remove_nonexisting_tag
FAILED ../../.tox/py310/lib/python3.10/site-packages/sunpy/io/tests/test_fits.py::test_read_hdus[/home/runner/work/sunpy/sunpy/.tox/py310/lib/python3.10/site-packages/sunpy/data/test/hsi_image_20101016_191218.fits-hdus2-2]
FAILED ../../.tox/py310/lib/python3.10/site-packages/sunpy/io/tests/test_fits.py::test_read_hdus[/home/runner/work/sunpy/sunpy/.tox/py310/lib/python3.10/site-packages/sunpy/data/test/hsi_image_20101016_191218.fits-hdus3-2]
FAILED ../../.tox/py310/lib/python3.10/site-packages/sunpy/net/vso/tests/test_attrs.py::test_wave_inputQuantity
```
failed, the database one, I wonder if it ever worked. We could probably just ignore that one.

ConorMacBride:
The failing fits tests should have been fixed by #5981. The deprecation PR must have be started before that was merged.

nabobalis:
> There are also quite a few fixtures and functions in the test files which are not being used, although I didn't delete any in case they are useful in the future or they are needed for online tests etc. I think it would be useful to include the test files in the coverage as we can more easily spot issues where assertions are not running. [I see that matplotlib includes test files in the coverage for example.](https://codecov.io/gh/matplotlib/matplotlib/tree/main/lib/matplotlib/tests) What do you think?

My memory (assuming its accurate) was that I wanted to include test file coverage for this reason but was overruled.

ConorMacBride:
Maybe I'll open a separate issue/PR so there can be more discussion. It would definitely be useful. If the issue is about mixing the two different types of coverage we may be able to use [codecov flags](https://docs.codecov.com/docs/flags) to separate them within the codecov UI.

nabobalis:
The online and doc builds are gonna fail for a while

PR #6067, opened 2022-04-12
removed more jsoc tests

nabobalis:
I also got:

```
The 'asyncio_mode' default value will change to 'strict' in future, please explicitly use 'asyncio_mode=strict' or 'asyncio_mode=auto' in pytest configuration file.:DeprecationWarning
```
So I patched that.

### DIFF
--- a/.github/workflows/cancel_ci.yml
+++ b/.github/workflows/cancel_ci.yml
@@ -2,7 +2,7 @@ name: Cancel duplicate workflows
 
 on:
   workflow_run:
-    workflows: ["CI"]
+    workflows: ["CI", "Automated tests", "Scheduled builds", "asv-benchmarks-daily"]
     types:
       - requested
 
@@ -14,3 +14,6 @@ jobs:
       uses: styfle/cancel-workflow-action@0.9.1
       with:
         workflow_id: ${{ github.event.workflow.id }}
+        ignore_sha: true
+        all_but_latest: true
+        access_token: ${{ github.token }}
--- a/sunpy/conftest.py
+++ b/sunpy/conftest.py
@@ -25,12 +25,19 @@ else:
 remotedata_spec = importlib.util.find_spec("pytest_remotedata")
 HAVE_REMOTEDATA = remotedata_spec is not None
 
+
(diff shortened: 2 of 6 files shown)


.github/workflows/cancel_ci.yml
```diff
@@ -14,3 +14,6 @@ jobs:
       uses: styfle/cancel-workflow-action@0.9.1
       with:
         workflow_id: ${{ github.event.workflow.id }}
+        ignore_sha: true
+        all_but_latest: true
+        access_token: ${{ github.token }}
```
nabobalis, line 19: This one isn't needed but yolo.

nabobalis:
Online fails are now HEC.

sunpy/net/jsoc/tests/test_jsoc.py
```diff
@@ -15,6 +14,7 @@
 from sunpy.net.jsoc import JSOCClient, JSOCResponse
 from sunpy.util.exceptions import SunpyUserWarning
 
+TEST_EMAIL = "nabil.freij@gmail.com"
```
wtbarnes, line 17: can you just put this in conftest so we don't have to change it everywhere if we want to change it to something less-Nabil-specific later?

wtbarnes (approved):

sunpy/net/jsoc/tests/test_jsoc.py
nabobalis: Sure.

sunpy/conftest.py
```diff
@@ -33,6 +34,12 @@
 console_logger.setLevel('INFO')
 
 
+@pytest.fixture(scope='session', autouse=True)
+def jsoc_test_email():
+    # JSOC Test email
+    pytest.jsoc_test_email = "nabil.freij@gmail.com"
```
Cadair, line 40: Is this really a recommended pattern, to just attach an attribute to an imported module? This seems real gross.

ConorMacBride: I'll change this in a PR

nabobalis: It was more fun this way.

PR #6073, opened 2022-04-13
Example and dev doc for new rotate

nabobalis:
Have you ever wanted more text? Well this PR delivers on that.

TODO:
- [x] Check they render nice.

### DIFF
--- a/docs/dev_guide/contents/map_rotate_custom.rst
+++ b/docs/dev_guide/contents/map_rotate_custom.rst
@@ -4,21 +4,20 @@
 Adding a new rotation method
 ****************************
 
-If you want to add a new rotation method that can be used by `sunpy.map.GenericMap.rotate`, you can do the following:
+It is possible to select from a number of rotation methods when using :func:`sunpy.image.transform.affine_transform` and :meth:sunpy.map.GenericMap.rotate`.
+You can add a custom rotation method using the decorator :func:`sunpy.image.transform.add_rotation_function`:
 
 .. code-block:: python
 
     from sunpy.image.transform import add_rotation_function
 
-    @add_rotation_function("my_rotate", handles_clipping=False, handles_image_nans=False, handles_nan_missing=False)
+    @add_rotation_function("my_rotate", allowed_orders={0, 1, 3},
+                           handles_clipping=False, handles_image_nans=False, handles_nan_missing=False)
     def _rotation_mine(image, matrix, shift, order, missing, clip):
         # Rotation code goes here
         return rotated_image
 
-The best explanation of the arguments is found from the detailed function documentation:
-
-.. autofunction:: sunpy.image.transform.add_rotation_function
-   :noindex:
+See the docstring for :func:`~sunpy.image.transform.add_rotation_function` for a detailed explanation of each of the decorator parameters and each of the required input parameters to the rotation function
 
 Then you can do:
 
@@ -31,16 +30,5 @@ Then you can do:
 
     rot_map = hmi_map.rotate(order=3, recenter=True, method="my_rotate")
 
-It is important to know what keywords from `sunpy.map.GenericMap.rotate` are used by the underlying rotation function.
 
-* ``angle`` - This is passed to `sunpy.image.transform.affine_transform` as a rotation matrix.
-* ``rmatrix`` - This is passed to `sunpy.image.transform.affine_transform`
-* ``order`` - This is passed to `sunpy.image.transform.affine_transform`
-* ``scale`` - This is passed to `sunpy.image.transform.affine_transform`
-* ``recenter`` - This is passed to `sunpy.image.transform.affine_transform`
-* ``missing`` - This is passed to `sunpy.image.transform.affine_transform`
-* ``use_scipy`` - This is passed to `sunpy.image.transform.affine_transform`, it is deprecated and removal is planned for sunpy 4.1, you should use ``method`` instead.
-* ``method`` - This is passed to `sunpy.image.transform.affine_transform`
-* ``clip`` - This is passed to `sunpy.image.transform.affine_transform`
-
-The methods that ``sunpy`` use are defined in `sunpy/image/transform.py <https://github.com/sunpy/sunpy/blob/main/sunpy/image/transform.py>`__`.
+The available rotation methods are all implemented using the :func:`~sunpy.image.transform.add_rotation_function` decorator, so you can look in `sunpy/image/transform.py <https://github.com/sunpy/sunpy/blob/main/sunpy/image/transform.py>`__ for examples of how to use this decorator.
--- a/docs/dev_guide/contents/map_rotate_custom.rst
+++ b/docs/dev_guide/contents/map_rotate_custom.rst
@@ -31,6 +31,19 @@ Then you can do:
 
     rot_map = hmi_map.rotate(order=3, recenter=True, method="my_rotate")
 
+It is important to know what keywords from `sunpy.map.GenericMap.rotate` are used by the rotation function.
+
+
+* ``angle`` - This is passed to `sunpy.image.transform.affine_transform` as a rotation matrix.
+* ``rmatrix`` - This is passed to `sunpy.image.transform.affine_transform`
+* ``order`` - This is passed to `sunpy.image.transform.affine_transform`
+* ``scale`` - This is passed to `sunpy.image.transform.affine_transform`
+* ``recenter`` - This is passed to `sunpy.image.transform.affine_transform`
+* ``missing`` - This is passed to `sunpy.image.transform.affine_transform`
+* ``use_scipy`` - This is passed to `sunpy.image.transform.affine_transform`
+* ``method`` - This is passed to `sunpy.image.transform.affine_transform`
+* ``clip`` - This is passed to `sunpy.image.transform.affine_transform`
+
 Below is a non-functioning OpenCV2 rotation method as an example:
 
 .. code-block:: python
--- a/docs/dev_guide/contents/map_rotate_custom.rst
+++ b/docs/dev_guide/contents/map_rotate_custom.rst
@@ -4,7 +4,7 @@
 Adding a new rotation method
 ****************************
 
-It is possible to select from a number of rotation methods when using :func:`sunpy.image.transform.affine_transform` and :meth:sunpy.map.GenericMap.rotate`.
+It is possible to select from a number of rotation methods when using :func:`sunpy.image.transform.affine_transform` and :meth:`sunpy.map.GenericMap.rotate`.
 You can add a custom rotation method using the decorator :func:`sunpy.image.transform.add_rotation_function`:
 
 .. code-block:: python
--- a/docs/dev_guide/contents/map_rotate_custom.rst
+++ b/docs/dev_guide/contents/map_rotate_custom.rst
@@ -15,7 +15,7 @@ If you want to add a new rotation method that can be used by `sunpy.map.GenericM
         # Rotation code goes here
         return rotated_image
 
-The best explanation of what the arguments mean is:
+The best explanation of the arguments is found from the detailed function documentation:
 
 .. autofunction:: sunpy.image.transform.add_rotation_function
    :noindex:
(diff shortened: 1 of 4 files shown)


docs/dev_guide/contents/map_rotate_custom.rst
```diff
@@ -0,0 +1,90 @@
...
+****************************
+
+If you want to add a new rotation method that can be used by `sunpy.map.GenericMap.rotate`, you can do the following:
+
+.. code-block:: python
+
+    from sunpy.image.transform import add_rotation_function
+
+    @add_rotation_function("my_rotate",
+                        handles_clipping=False, handles_image_nans=False, handles_nan_missing=False)
```
Cadair, line 14: Is the indentation really out on these lines like it renders on GH or is GH having a fit?

docs/dev_guide/contents/map_rotate_custom.rst
```diff
@@ -0,0 +1,90 @@
...
+    from sunpy.map import Map
+    from sunpy.data import sample
+
+    hmi_map = sunpy.map.Map(sample.HMI_LOS_IMAGE)
+
+    rot_map = hmi_map.rotate(order=3, recenter=True, method="my_rotate")
+
+There are several things you will want to be aware of when adding a new rotation method.
+
+* The `.add_rotation_function` means you can handle nans, clipping, and missing values in the image array
```
Cadair, line 32: "you can handle" here is ambiguous, does the function have to handle it or the decorator. The decorator isn't "me".

I think we need to describe the expected inputs and outputs of a rotation function first. Things like you shouldn't modify the input array, the function + decorator must handles NaNs, handle missing values, clipping etc. Then after the expectations of the function are setup then we can say "and there is this decorator which can do these things for you".

Basically I am saying drop the list and describe the function, then introduce the details of the decorator as an aid.

docs/dev_guide/contents/map_rotate_custom.rst
```diff
@@ -0,0 +1,90 @@
...
+  before it is passed to your rotation function via the boolean flags.
+  The documentation in `.add_rotation_function` explain how these work.
+* The arguments into the rotation method must be ``image, matrix, shift, order, missing, clip``.
+  There is no variation allowed.
+* The return value of the rotation method must the rotated image array.
+* You must describe the rotation method in the docstring of the function using unnumbered bullet points.
+
+Below is a semi-functioning OpenCV2 rotation method as an example:
```
Cadair, line 40: You probably need to elaborate on semi-functioning or people are going to ask why this isn't just in the code.

examples/map/compare_rotation_results.py
```diff
@@ -0,0 +1,106 @@
...
+fig0, ax0 = plt.subplots()
+img0 = ax0.imshow(skimage_map.data - scipy_map.data, vmin=-50, vmax=50, cmap=plt.cm.seismic)
+fig0.colorbar(img0)
+ax0.set_title("Raw Difference: scikit-image vs scipy")
+
+plt.show()
+
+###############################################################################
+# We can repeat this but for AIA data, using a 171 sample image.
+# Since there is no rotation in the FITS header, it will just apply the same rotation to the image.
```
Cadair, line 79: If there isn't a rotation in the header this isn't achieving anything other than convolving the image a bunch of times? Wouldn't it be better to rotate the image by some arbitrary angle for testing purposes?

examples/map/compare_rotation_results.py
```diff
@@ -0,0 +1,106 @@
...
+# This essentially means that the average difference between every pixel in the
+# ``skimage`` and ``scipy`` images is about 15\%, in the final rotated image.
+
+###############################################################################
+# Now for a visual comparison, the raw differences, that should highlight the differences.
+
+fig0, ax0 = plt.subplots()
+img0 = ax0.imshow(skimage_map.data - scipy_map.data, vmin=-50, vmax=50, cmap=plt.cm.seismic)
+fig0.colorbar(img0)
+ax0.set_title("Raw Difference: scikit-image vs scipy")
```
Cadair, line 73: You should differentiate the two titles of the two plots.

docs/dev_guide/contents/map_rotate_custom.rst
nabobalis: That looks like a rendering on the diff.

examples/map/compare_rotation_results.py
nabobalis: 30 degrees it is

docs/dev_guide/contents/map_rotate_custom.rst
nabobalis: How about I just replicate the very detailed doc string that Albert wrote?

nabobalis: Done.

docs/dev_guide/contents/map_rotate_custom.rst
```diff
@@ -0,0 +1,84 @@
...
+    def _rotation_mine(image, matrix, shift, order, missing, clip):
+        # Rotation code goes here
+        return rotated_image
+
+The best explanation of what the arguments mean is:
+
+.. autofunction:: sunpy.image.transform.add_rotation_function
+   :noindex:
+
+Then you can do:
```
dstansby, line 23: Can you add a sentence that explains how keyword arguments are passed from `rotate()` to the custom rotate function?

docs/dev_guide/contents/map_rotate_custom.rst
```diff
@@ -0,0 +1,84 @@
...
+.. code-block:: python
+
+    from sunpy.map import Map
+    from sunpy.data import sample
+
+    hmi_map = sunpy.map.Map(sample.HMI_LOS_IMAGE)
+
+    rot_map = hmi_map.rotate(order=3, recenter=True, method="my_rotate")
+
+Below is a non-functioning OpenCV2 rotation method as an example:
```
dstansby, line 34: Why non-funcitoning? Can we make it functioning without much effort?

examples/map/compare_rotation_results.py
```diff
@@ -0,0 +1,107 @@
...
+
+###############################################################################
+# This essentially means that the average difference between every pixel in the
+# ``skimage`` and ``scipy`` images is about 15\%, in the final rotated image.
+
+###############################################################################
+# Now for a visual comparison, the raw differences, that should highlight the differences.
+
+fig0, ax0 = plt.subplots()
+img0 = ax0.imshow(skimage_map.data - scipy_map.data, vmin=-50, vmax=50, cmap=plt.cm.seismic)
```
dstansby, line 73: The differences are washed out using this norm - could you use [`SymLogNorm`](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.SymLogNorm.html) instead to make these images look better?

dstansby (changes requested):

docs/dev_guide/contents/map_rotate_custom.rst
nabobalis: It lacks proper handling of the shift (based off the comments from the original PR) and someone who understands the differences in the affine transform in openCV2 could make it functioning with some effort.

examples/map/compare_rotation_results.py
nabobalis: Sure.

docs/dev_guide/contents/map_rotate_custom.rst
nabobalis: Fixed maybe?

examples/map/compare_rotation_results.py
nabobalis: Fixed maybe?

dstansby (approved):

ayshih (changes requested):
I'm going to block this because I find including the "non-functioning" OpenCV code to be unpalatable.  I'm working on figuring out how to fix it.

nabobalis:
I rebased this on top of the opencv2 PR so I can update the example.

docs/dev_guide/contents/map_rotate_custom.rst
```diff
@@ -0,0 +1,46 @@
...
+
+* ``angle`` - This is passed to `sunpy.image.transform.affine_transform` as a rotation matrix.
+* ``rmatrix`` - This is passed to `sunpy.image.transform.affine_transform`
+* ``order`` - This is passed to `sunpy.image.transform.affine_transform`
+* ``scale`` - This is passed to `sunpy.image.transform.affine_transform`
+* ``recenter`` - This is passed to `sunpy.image.transform.affine_transform`
+* ``missing`` - This is passed to `sunpy.image.transform.affine_transform`
+* ``use_scipy`` - This is passed to `sunpy.image.transform.affine_transform`, it is deprecated and removal is planned for sunpy 4.1, you should use ``method`` instead.
+* ``method`` - This is passed to `sunpy.image.transform.affine_transform`
+* ``clip`` - This is passed to `sunpy.image.transform.affine_transform`
```
ayshih, line 44:
```suggestion
```
How is this "important to know"?  I presume you added this in response to @dstansby's comment (https://github.com/sunpy/sunpy/pull/6073/files#r851439078), but I don't understand the value in this description.  Why would a developer care how parameters are passed along?

docs/dev_guide/contents/map_rotate_custom.rst
```diff
@@ -0,0 +1,46 @@
...
+
+The methods that ``sunpy`` use are defined in `sunpy/image/transform.py <https://github.com/sunpy/sunpy/blob/main/sunpy/image/transform.py>`__`.
```
ayshih, line 46:
```suggestion
The available rotation methods are all implemented using the :func:`~sunpy.image.transform.add_rotation_function` decorator, so you can look in `sunpy/image/transform.py <https://github.com/sunpy/sunpy/blob/main/sunpy/image/transform.py>`__ for examples of how to use this decorator.
```

ayshih (changes requested):
The fact that the example uses SMAPE to compare the results is deeply unsatisfying.  SMAPE horrendously emphasizes discrepancies where pixel values are near zero, which is not where most users of AIA data care about.  (Users of HMI data might care more.)

Any image-averaged metric is also problematic because it disregards the structure/coherence of the differences.  Here's the image comparison if I take the final image in the gallery example, and instead plot it on a linear scale from -10 to 10:
rotation comparison
No matter what an image-averaged metric may say, scipy and scikit-image are much more consistent to each other than to opencv.

nabobalis:
> The fact that the example uses SMAPE to compare the results is deeply unsatisfying. SMAPE horrendously emphasizes discrepancies where pixel values are near zero, which is not where most users of AIA data care about. (Users of HMI data might care more.)

So would you prefer me to remove that metric completely and just have the raw subscrtaction with a linear scale?

examples/map/compare_rotation_results.py
```diff
@@ -0,0 +1,75 @@
...
+
+img2 = ax2.imshow(scipy_map.data - cv2_map.data, cmap='RdBu_r', vmin=-10, vmax=10)
+ax2.set_title("HMI Difference: scipy vs opencv2")
+fig.colorbar(img2, ax=ax2)
+
+plt.show()
+
+###############################################################################
+# We can repeat this but for AIA data, using a 171 sample image.
+# Since there is no rotation in the FITS header, we will rotate by 30 degrees
```
ayshih, line 53:
```suggestion
# We will rotate it by the large amount of 30 degrees.
```
Funnily enough, our sample image *does* have a rotation angle, but it's very small.

ayshih (changes requested):
Small things, but the bigger thing is that you need to add `opencv` to the docs build.

nabobalis:
As far as I can see, doc build failed only due to a remote data issue.
Online test suite fails have nothing to do with the PR.

docs/dev_guide/contents/map_rotate_custom.rst
```diff
@@ -0,0 +1,34 @@
+.. _map_rotate_custom:
+
+****************************
+Adding a new rotation method
+****************************
+
+It is possible to select from a number of rotation methods when using :func:`sunpy.image.transform.affine_transform` and :meth:sunpy.map.GenericMap.rotate`.
```
ayshih, line 7:
```suggestion
It is possible to select from a number of rotation methods when using :func:`sunpy.image.transform.affine_transform` and :meth:`sunpy.map.GenericMap.rotate`.
```

ayshih (approved):
Just one missing backtick, otherwise good to go

docs/dev_guide/contents/map_rotate_custom.rst
nabobalis: I wonder why that didnt raise a warning I could find in the doc build.

ayshih: Heh, I wondered the same thing.  I guess it was so confused by it that it simply interpreted it as straight text.

Issue #5963, opened 2022-03-15
Cannot draw limb on non-WCSAxes instance.


wtbarnes:
### Describe the bug

We explicitly allow for drawing the limb even the axis is not a `WCSAxes` instance,

https://github.com/sunpy/sunpy/blob/a5c5be9504fef613cbe4a3957e587bc797ffc07e/sunpy/map/mapbase.py#L2144-L2150

However, there is undefined variable `c_kw` in this block that throws a `NameError`. I'm assuming that this is supposed to be kwargs for `patches.Circle`, but I'm not sure what the intended value is supposed to be.

### To Reproduce

```python
import sunpy.map
import sunpy.data.sample
import matplotlib.pyplot as plt

m = sunpy.map.Map(sunpy.data.sample.AIA_094_IMAGE)

fig = plt.figure()
ax=fig.add_subplot(111)
m.draw_limb(axes=ax)
```
### System Details

- sunpy version: 3.1.4

ayshih:
It got broken in #5414

ayshih:
All that should need to be done is to restore these lines for this specific code branch:
```python
c_kw = {'fill': False,
        'color': 'white',
        'zorder': 100}
c_kw.update(kwargs)
```

wtbarnes:
I'm sure I'm just missing something, but what is the use case for plotting the limb on a non-WCS axis and how would one do it? My attempt at coming up with a figure test is clearly wrong,

```python
import sunpy.map
import sunpy.data.sample
import matplotlib.pyplot as plt

m = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)

fig = plt.figure()
ax=fig.add_subplot(111)
ax.imshow(m.data,**m.plot_settings)
m.draw_limb(axes=ax)
plt.show()
```

<img width="623" alt="image" src="https://user-images.githubusercontent.com/8676141/158459115-9d94589d-cca7-4d47-95f6-c2206d40d504.png">

I'm assuming this is just a case of passing the right `transform` to `patches.Circle` via `draw_limb`.

dstansby:
From memory, you can do `map.plot(axes=not_a_wcs_axes)` and it should work?

Cadair:
I thought we deprecated support for non-wcsaxes?

Cadair:
or did I just wish that had happened?

wtbarnes:
```python
fig = plt.figure()
ax=fig.add_subplot(111)
[...]

PR #6079, opened 2022-04-15
Deprecate draw_limb on non-WCS axes

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/5963. Plotting a map on non-wcs Axes has been unsupported for a while, so deprecated doing `draw_limb()` on a non-wcs Axes and we can remove in 4.1.

I also cleaned up a few code paths that are never hit because we force the use of WCSAxes.


nabobalis:
Do we have any other methods we need to check for this as well?

nabobalis (approved):

Issue #5894, opened 2022-02-19
Improve the performance of `CompositeMap` plotting


ayshih:
As of #5255, `CompositeMap` internally uses the `autoalign=True` setting for Map plotting to ensure that all layers are aligned.  However, plotting with `autoalign=True` is slow (because it uses Maplotlib's `pcolormesh()`), which is especially awful in interactive Matplotlib windows.  `autoalign=True` is actually only needed when a given layer's WCS does not match the axes's WCS, so we should set `autoalign=False` when autoalignment isn't needed.

vmpyr:
Hi, I'd like to work on this issue if no one else is. This is my first issue ever in open-source, so I hope it's okay for a beginner

nabobalis:
> Hi, I'd like to work on this issue if no one else is. This is my first issue ever in open-source, so I hope it's okay for a beginner

I don't think this issue is a good place to start.

ooprathamm:
Hi , I have made a pull request regarding this issue  #6085

PR #6085, opened 2022-04-16
Improve the performance of CompositeMap plotting

ooprathamm:
Partial attempt at fixing #5894

### DIFF
--- a/sunpy/map/compositemap.py
+++ b/sunpy/map/compositemap.py
@@ -439,8 +439,12 @@ class CompositeMap:
             # non False levels property.
             if m.levels is False:
                 # We tell GenericMap.plot() that we need to autoalign the map
-                if wcsaxes_compat.is_wcsaxes(axes):
+                
+                if wcsaxes_compat.gca_wcs(m.wcs)==axes.wcs:
+                    params['autoalign'] = False
+                elif wcsaxes_compat.is_wcsaxes(axes):
                     params['autoalign'] = True
+                   
 
                 # Filter `matplot_args`
                 if params.get('autoalign', None) in (True, 'pcolormesh'):
@@ -523,7 +527,7 @@ class CompositeMap:
             raise TypeError("draw_grid should be bool, int, long or float")
 
         return figure
-
+ 
 
 class OutOfRangeAlphaValue(ValueError):
     """Exception to raise when an alpha value outside of the range 0-1 is
--- a/sunpy/map/tests/test_compositemap.py
+++ b/sunpy/map/tests/test_compositemap.py
@@ -11,6 +11,8 @@ import sunpy.data.test
 import sunpy.map
 from sunpy.tests.helpers import figure_test
 
+import test_plotting
+
 # Ignore missing metadata warnings
 pytestmark = [pytest.mark.filterwarnings('ignore:Missing metadata for observer'),
               pytest.mark.filterwarnings(r'ignore:Unable to treat `\.meta` as a FITS header')]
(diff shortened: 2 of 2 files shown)


sunpy/map/tests/test_compositemap.py
```diff
@@ -11,6 +11,8 @@
 import sunpy.map
 from sunpy.tests.helpers import figure_test
 
+import test_plotting
```
nabobalis, line 14: I don't think you need this import.

sunpy/map/compositemap.py
```diff
@@ -439,8 +439,12 @@ def plot(self, axes=None, annotate=True,
             # non False levels property.
             if m.levels is False:
                 # We tell GenericMap.plot() that we need to autoalign the map
-                if wcsaxes_compat.is_wcsaxes(axes):
+                
+                if wcsaxes_compat.gca_wcs(m.wcs)==axes.wcs:
+                    params['autoalign'] = False
+                elif wcsaxes_compat.is_wcsaxes(axes):
                     params['autoalign'] = True
```
nabobalis, line 446: How does this speed up the plotting? I am not familiar with CompositeMap plotting

sunpy/map/compositemap.py
```diff
@@ -523,7 +527,7 @@ def peek(self, colorbar=True, draw_limb=True, draw_grid=False, **matplot_args):
             raise TypeError("draw_grid should be bool, int, long or float")
 
         return figure
-
+
```
nabobalis, line 530:
```suggestion

```

ayshih (changes requested):
I marked this PR as "draft" since it is not finished, here's stuff that needs to be fixed in it:

- You access `axes.wcs` before you check if `axes` is a `WCSAxes` instance.
- Calling `gca_wcs()` doesn't make sense because (1) it will return the current `pyplot` axes if one exists, and thus may not correspond to `m.wcs` at all, and (2) if a user hasn't been using `pyplot`, it will create a new figure and axes through `pyplot`.
- You (attempt to) import `test_plotting.py`, but I don't know why.
- Your unit test appears to be adapted from `test_plot_autoalign()` from `test_plotting.py`, but I suspect you did not run your unit test at all because your code tries to access attributes and methods that aren't available for `CompositeMap`.  Plus, you copied over the routine, but it would work as intended only if it is also marked as a figure test.
- Even if your unit test did actually run, it doesn't test the functionality that this PR is trying to add because you are trying to plot the composite map onto a non-matching WCS.  That does and should continue to use `autoalign=True`.  A proper test would likely not be a figure test – because the figure output would be very similar and possibly even equal – but instead would probably have to check whether a `matplotlib.collections.QuadMesh` was attached to the plot when a single composite-map layer is plotted onto a matching WCS.

sunpy/map/compositemap.py
ayshih: See #5894.  We get improved performance if we set `autoalign=False` when the map layer is being plotted onto axes with a matching WCS because then we use the faster `imshow()` instead of the slower `pcolormesh()`.

nabobalis: Ah, so the idea is to bypass using pcolormesh when it makes sense?

ayshih: Yup, precisely, because it's glacial in an interactive plot.

nabobalis:
Hi @ooprathamm, thanks for the PR.

Hopefully you will have time in the future to update this PR using the comments I and ayshih left!

nabobalis:
Thank you for the PR @ooprathamm, if you have a chance to work on this in figure, please do reopen this PR or a new one!

PR #6086, opened 2022-04-18
Switch helioviewer URL order

dstansby:
api.helioviewer.org is the one in the official docs (e.g. https://api.helioviewer.org/docs/v2/api/api_groups/jpeg2000.html#example-binary-jpeg2000-image-data), so use that first. This should prevent some test warnings/errors because https://helioviewer-api.ias.u-psud.fr/ is currently down.


nabobalis:
We originally set it the other way as helioviewer was down.

nabobalis (approved):

PR #6088, opened 2022-04-19
Add fix for dask html repr bug. Add regression tests

### LINKED ISSUE #5234: Map HTML repr crashes when data array is a Dask array
If your map is backed by a Dask array (which `Map` does *nominally* support), the nifty HTML repr freaks out.

```python
import sunpy.data.sample
import sunpy.map
import dask.array
m = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
m_dask = sunpy.map.Map(dask.array.from_array(m.data),m.meta)
m_dask._repr_html_()
```

gives

```python traceback
<ipython-input-17-89689641455e> in <module>
----> 1 m_dask._repr_html_()

~/Documents/codes/sunpy/sunpy/map/mapbase.py in _repr_html_(self)
    356             FigureCanvasBase(fig)
    357         ax = fig.subplots()
--> 358         values, bins, patches = ax.hist(finite_data.ravel(), bins=100)
    359         norm_centers = norm(0.5 * (bins[:-1] + bins[1:])).data
    360         for c, p in zip(norm_centers, patches):

~/miniconda3/envs/sunpy-dev/lib/python3.8/site-packages/matplotlib/__init__.py in inner(ax, data, *args, **kwargs)
   1445     def inner(ax, *args, data=None, **kwargs):
   1446         if data is None:
-> 1447             return func(ax, *map(sanitize_sequence, args), **kwargs)
   1448
   1449         bound = new_sig.bind(ax, *args, **kwargs)

~/miniconda3/envs/sunpy-dev/lib/python3.8/site-packages/matplotlib/axes/_axes.py in hist(self, x, bins, range, density, weights, cumulative, bottom, histtype, align, orientation, rwidth, log, color, label, stacked, **kwargs)
   6569
   6570         # Massage 'x' for processing.
-> 6571         x = cbook._reshape_2D(x, 'x')
   6572         nx = len(x)  # number of datasets
   6573

~/miniconda3/envs/sunpy-dev/lib/python3.8/site-packages/matplotlib/cbook/__init__.py in _reshape_2D(X, name)
   1373
   1374     # Iterate over list of iterables.
-> 1375     if len(X) == 0:
   1376         return [[]]
   1377

TypeError: 'float' object cannot be interpreted as an integer
```

warussel:
## PR Description

Fixes #5234

Adds support for dask arrays using the _repr_html_ method. Checks if a map data source is a dask array. If so, uses a bit flag to either
a) pull the array into memory and plot it such as for the quicklook method or
b) use the dask _repr_html_ method to display useful information without computing the dask array. 
Adds compute_dask bit flag to _repr_html_ parameters. Default value is set to False. 
Adds unit test to test_mapbase.py to check that both methods work with a dask input array. 

Testing results:
_html_repr_

Screen Shot 2022-04-18 at 11 37 17 PM

quicklook
Screen Shot 2022-04-18 at 11 40 21 PM


sunpy/map/mapbase.py
```diff
@@ -17,6 +17,7 @@
 from matplotlib import cm
 from matplotlib.backend_bases import FigureCanvasBase
 from matplotlib.figure import Figure
+from dask.array import Array as DaskArray
```
nabobalis, line 20: I think this should probably be in the `_repr_html_` function, otherwise we will have to add dask to the main dependencies.

sunpy/map/tests/test_mapbase.py
```diff
@@ -1181,6 +1181,27 @@ def test_quicklook(aia171_test_map):
         assert aia171_test_map._repr_html_() in html_string
 
 
+# requires dask array to run properly
+def test_dask_array(aia171_test_map):
+    dask_array = pytest.importorskip('dask.array')
```
nabobalis, line 1186: Do we need to have this? I feel like we run our test suite with all our optional dependencies installed.

wtbarnes: In setup.cfg, the dask dependency is under its own extras heading and is not included under the test heading so I would say this is the right approach. This is also how the tests are setup for dask arrays in the map factory so I think this is the best approach.

sunpy/map/tests/test_mapbase.py
```diff
@@ -1181,6 +1181,27 @@ def test_quicklook(aia171_test_map):
...
+
+    da = dask_array.from_array(aia171_test_map.data, chunks=(1, 1))
+    pair_map = sunpy.map.Map(da, aia171_test_map.meta)
+
+    # Check that _repr_html_ functions for a dask array
+    html_string = pair_map._repr_html_()
+    assert isinstance(html_string, str)
```
wtbarnes, line 1193: Wouldn't this test pass on our main branch right now? The issue isn't necessarily whether it functions, but whether it pulls the data array into memory.

sunpy/map/mapbase.py
wtbarnes: The odds of using dask array somewhere else in mapbase in the future are high. I would just catch the import error here and set a flag, e.g.

```suggestion
DASK_INSTALLED = False
try:
    from dask.array import Array as DaskArray
    DASK_INSTALLED = True
except ImportError:
    pass
```

sunpy/map/mapbase.py
```diff
@@ -317,6 +318,26 @@ def _repr_html_(self):
         count_nan = np.isnan(self.data).sum()
         count_inf = np.isinf(self.data).sum()
 
+        # If data is from a dask array, pull data into memory for quicklook() method
+        # For basic html representation, only use dask html representation without computing data
+        if isinstance(finite_data, DaskArray):
```
wtbarnes, line 323: See above comment too:

```suggestion
        if DASK_INSTALLED and isinstance(finite_data, DaskArray):
```

wtbarnes (changes requested):
Thanks @warussel! Just a few minor suggestions regarding the optional dependency imports and the test.

sunpy/map/tests/test_mapbase.py
wtbarnes: One option that might be a little hacky: you could call `_repr_html` with `compute_dask=True` and `False` and then assert that they are unequal for the case of the map that holds a Dask array.

warussel:
Thanks for the reviews @nabobalis @wtbarnes. I will make changes and also add changelog.

sunpy/map/mapbase.py
warussel: Added changes as suggested

sunpy/map/tests/test_mapbase.py
warussel: Changed test as suggested

changelog/6088.bugfix.rst
```diff
@@ -0,0 +1 @@
+Add fix for dask arrays in map html representation.
```
nabobalis, line 1:
```suggestion
Fixed a crash for dask arrays when displaying the `~sunpy.map.GenericMap` html representation.
```

sunpy/map/mapbase.py
```diff
@@ -17,6 +17,12 @@
 from matplotlib import cm
 from matplotlib.backend_bases import FigureCanvasBase
 from matplotlib.figure import Figure
+DASK_INSTALLED = False
+try:
+    from dask.array import Array as DaskArray
+    DASK_INSTALLED = True
+except ImportError:
+    pass
```
nabobalis, line 25:
```suggestion
try:
    from dask.array import Array as DaskArray
    DASK_INSTALLED = True
except ImportError:
    DASK_INSTALLED = False
```
I personally prefer this.

sunpy/map/mapbase.py
```diff
@@ -317,6 +323,26 @@ def _repr_html_(self):
         count_nan = np.isnan(self.data).sum()
         count_inf = np.isinf(self.data).sum()
 
+        # If data is from a dask array, pull data into memory for quicklook() method
+        # For basic html representation, only use dask html representation without computing data
+        if DASK_INSTALLED and isinstance(finite_data, DaskArray):
+            if compute_dask:
+                finite_data = finite_data.compute()
```
nabobalis, line 330:
```suggestion
            # This will fetch the entire data array into memory.
            if compute_dask:
                finite_data = finite_data.compute()
```

sunpy/map/mapbase.py
```diff
@@ -317,6 +323,26 @@ def _repr_html_(self):
         count_nan = np.isnan(self.data).sum()
         count_inf = np.isinf(self.data).sum()
 
+        # If data is from a dask array, pull data into memory for quicklook() method
+        # For basic html representation, only use dask html representation without computing data
```
nabobalis, line 327:
```suggestion
```

sunpy/map/tests/test_mapbase.py
```diff
@@ -1181,6 +1181,21 @@ def test_quicklook(aia171_test_map):
         assert aia171_test_map._repr_html_() in html_string
 
 
+# requires dask array to run properly
```
nabobalis, line 1184:
```suggestion
```

sunpy/map/tests/test_mapbase.py
```diff
@@ -1181,6 +1181,21 @@ def test_quicklook(aia171_test_map):
...
+def test_dask_array(aia171_test_map):
+    dask_array = pytest.importorskip('dask.array')
+
```
nabobalis, line 1187:
```suggestion
```

sunpy/map/tests/test_mapbase.py
```diff
@@ -1181,6 +1181,21 @@ def test_quicklook(aia171_test_map):
...
+    da = dask_array.from_array(aia171_test_map.data, chunks=(1, 1))
+    pair_map = sunpy.map.Map(da, aia171_test_map.meta)
+
+    # Check that _repr_html_ functions for a dask array
+    html_dask_repr = pair_map._repr_html_()
+
```
nabobalis, line 1193:
```suggestion
```
I don't think we need this line?

warussel: oops that was leftover. Thanks for the catch

sunpy/map/tests/test_mapbase.py
warussel: made changes

sunpy/map/tests/test_mapbase.py
warussel: made changes as requested

sunpy/map/mapbase.py
warussel: made changes as requested

sunpy/map/mapbase.py
warussel: made changes as requested

sunpy/map/mapbase.py
warussel: I like this as well. Made changes as requested.

changelog/6088.bugfix.rst
warussel: made changes as requested. Added backticks

nabobalis (approved):

wtbarnes (approved):

nabobalis:
Thanks for the PR @warussel

nabobalis:
pre-commit.ci autofix

Issue #5784, opened 2022-01-04
`.rotate()` does not correctly de-rotate maps with a PCij matrix that is not a pure rotation


ayshih:
`GenericMap.rotate()` is currently bugged when it uses a rotation matrix that is not a pure rotation.  While we could simply slap a warning to dissuade a user from explicitly providing such a matrix, a user can implicitly encounter this issue when de-rotating a map (i.e., calling `.rotate()` without any arguments) that has a PCij matrix that is not a pure rotation.  One example of a map with such a PCij matrix is one with rectangular pixels that are not aligned with the coordinate axes.

```python
import numpy as np
import matplotlib.pyplot as plt
from skimage.data import camera

import sunpy.map

data = np.flipud(camera().astype('float'))[::2]
header = {
    'CRVAL1': 0,
    'CRVAL2': 0,
    'CRPIX1': 0.5,
    'CRPIX2': 0.5,
    'NAXIS1': data.shape[0],
    'NAXIS2': data.shape[1],
    'CUNIT1': 'arcsec',
    'CUNIT2': 'arcsec',
    'CTYPE1': 'HPLN-TAN',
    'CTYPE2': 'HPLT-TAN',
    'CDELT1': 1,
    'CDELT2': 2,  # note != CDELT1, so rectangular pixels
    'CROTA2': 30,  # CROTA != 0 with rectangular pixels results in a PCij matrix that is not a pure rotation
    'DATE-OBS': '2020-01-01 00:00:00'
}
base = sunpy.map.Map(data, header)
rot = base.rotate()  # uses the PCij matrix that is not a pure rotation

plt.figure()
base.plot()
plt.gca().coords.grid(color='red', lw=1)
```
![base](https://user-images.githubusercontent.com/991759/148321906-0a9085d0-1d97-48b4-a2d1-a0a7c49c0513.png)

```python
plt.figure()
rot.plot()
plt.gca().coords.grid(color='red', lw=1)
```
![rot](https://user-images.githubusercontent.com/991759/148321919-e1fe0fb4-5cb4-4004-b90e-b9f15744966c.png)
There are a couple of problems with this output map.

PR #6089, opened 2022-04-19
Add OpenCV as an option for rotation

### LINKED ISSUE #4478: OpenCV as alternative library for map.image.affine_transform

ayshih:
* Following #5867, this PR adds OpenCV as an option for rotating images or maps (by specifying `method='cv2'`).  This closes #4478.  See also #4502 and #6071.
* Changed the default interpolation order for `GenericMap.rotate()` from 4 to 3
* Make the definition of supported values for ``order`` part of the rotation-function registration

To do:

- [x] Figure out some sane way to link to the documentation for `cv2.warpAffine()`
- [x] ~Fix the bug in `pytest-mpl` for generating the HTML summary when an image comparison fails due to a mismatch in image dimensions~ Bug is present in `pytest-mpl` 0.14.0, but is not present in the dev branch.


sunpy/image/transform.py
```diff
@@ -325,6 +325,58 @@ def _rotation_skimage(image, matrix, shift, order, missing, clip):
...
+
+    # Swap the byte order if it is non-native (e.g., big-endian on a little-endian system)
+    if adjusted_image.dtype.byteorder == ('>' if sys.byteorder == 'little' else '<'):
+        adjusted_image = adjusted_image.byteswap().newbyteorder()
+
+    # missing must be a Python float, not a NumPy float
+    missing = float(missing)
+
+    return cv2.warpAffine(adjusted_image.T, trans, adjusted_image.shape,
+                          flags=_CV_ORDER_FLAGS[order] | cv2.WARP_INVERSE_MAP,
```
nabobalis, line 376: What does `_CV_ORDER_FLAGS[order] | cv2.WARP_INVERSE_MAP` mean for the final input to the function?

sunpy/image/transform.py
```diff
@@ -325,6 +325,58 @@ def _rotation_skimage(image, matrix, shift, order, missing, clip):
     return rotated_image
 
 
+@add_rotation_function("cv2",
+                       handles_clipping=False, handles_image_nans=False, handles_nan_missing=False)
+def _rotation_skimage(image, matrix, shift, order, missing, clip):
```
cbard, line 330: Should this be named `_rotation_cv2` instead?

cbard:
Hey @ayshih ! Thanks for digging out my old PR and incorporating it into your `rotate` redesign!

sunpy/image/transform.py
ayshih: Ha, whoops, yes.  It technically doesn't matter because the function is never called by name, even internally, so the function names could all be the same, generic name.  But certainly a mismatched name is confusing!

sunpy/image/transform.py
ayshih: They're independent bits in the bitfield.  Bits 0-3 specify the interpolation algorithm, and bit 4 specifies whether the mapping is an inverse map.

ayshih:
Rats, my implementation isn't working correctly on non-square pixels...  Stay tuned

ayshih:
To be clear, I mean non-square pixels that are not aligned with the coordinate axes, that is a PCij matrix that is not a pure rotation (see #5784/#5803)

setup.cfg
```diff
@@ -77,6 +77,7 @@ visualization =
 tests =
   hypothesis>=6.0.0  # Included in pytest-astropy. 6.0 is the first version to support disabling function-scoped fixture warning
   jplephem  # For some coordinates tests
+  opencv-python
```
Cadair, line 80: Does this provide binary wheels with opencv included?

sunpy/image/transform.py
```diff
@@ -325,6 +325,60 @@ def _rotation_skimage(image, matrix, shift, order, missing, clip):
     return rotated_image
 
 
+@add_rotation_function("cv2",
+                       handles_clipping=False, handles_image_nans=False, handles_nan_missing=False)
```
Cadair, line 329: Sad

ayshih:
The bug with rotating by a PCij matrix that is not a pure rotation has been fixed.  It's neat that parametrizing a figure test works.

The figure-test CI builds are currently failing due to a bug that is present in the latest stable release of `pytest-mpl` (0.14.0).

ayshih:
There is a quirk in my implementation here where I allow for `order=4` even though OpenCV doesn't implement biquartic, and simply have it use bicubic (i.e., `order=4` and `order=3` are equivalent).  The reason is that I want OpenCV rotation to Just Work with the default for `GenericMap.rotate()`, which is `order=4`.

This may be an opportune time to reconsider the change in #1185 (by me, for SunPy 0.7!), and perhaps revert back to a default of `order=3`.  I haven't yet investigated, but maybe the switch in #5867 to defaulting to SciPy mitigates the concerns that had been raised with `order=3` as the default.

ayshih:
I went ahead and switched `GenericMap.rotate()` to default to `order=3`.

sunpy/image/transform.py
```diff
@@ -325,8 +346,58 @@ def _rotation_skimage(image, matrix, shift, order, missing, clip):
     return rotated_image
 
 
+@add_rotation_function("cv2", allowed_orders={0, 1, 3},
```
Cadair, line 349: Why `'cv2'` and not `'opencv'`?

ayshih: \<shrug\>  Extrapolating from the single point of `'skimage'` instead of `'scikit-image'`, it seems that we are preferring the import name.

changelog/6089.feature.rst
```diff
@@ -0,0 +1,2 @@
+Added the option to rotate using `OpenCV <https://opencv.org>`__ when using :func:`sunpy.image.transform.affine_transform` or :meth:`sunpy.map.GenericMap.rotate` by specifying ``method=='cv2'``.
```
nabobalis, line 1:
```suggestion
Added the option to rotate using `OpenCV <https://opencv.org>`__ when using :func:`sunpy.image.transform.affine_transform` or :meth:`sunpy.map.GenericMap.rotate` by specifying ``method='cv2'``.
```

sunpy/image/transform.py
nabobalis: I also like "cv2" over "opencv".

changelog/6089.feature.rst
ayshih: Fixed

ayshih:
I added an initial commit that preserves @cbard's authorship at some level.

I added (potentially objectionable) scraping of the OpenCV docs to construct the link to their docstring for `cv2.warpAffine()` for their latest stable version.  The link can't be styled exactly the same as intersphinx links, lest it trigger an intersphinx warning due to not having an inventory for `cv2`.

ayshih:
I'll point out that even though the figure-test CI builds are crashing (with `pytest-mpl` 0.14.0), that's only because the HTML summary page can't be generated when there is mismatch in image dimensions between the baseline and the new images.  Once this PR is merged, its images will become the new baseline, so there won't be a persisting crashing of the figure-test CI builds.

setup.cfg
nabobalis: Its a large wheel download so I am going to say yes.

nabobalis (approved):

docs/conf.py
```diff
@@ -253,6 +253,39 @@
...
+                           'html.parser')
+    warpAffine = all_cv.find("a", string="warpAffine").attrs['href'][6:]  # strip leading "../../"
+
+    # Construct the full URL for warpAffine
+    warpAffine_full = f"{base_url}/{version}/{warpAffine}"
+except:
+    # In the event of any failure (e.g., no network connectivity)
+    warpAffine_full = ""
+
+rst_epilog = f"""
```
dstansby, line 283: Putting this at the end of *every* .rst file seems a bit much - I presume there's not an easy way to put this just on the pages that it's needed?

changelog/6089.breaking.rst
```diff
@@ -0,0 +1 @@
+Changed the default interpolation order for :meth:`sunpy.map.GenericMap.rotate` from 4 (typically biquartic) to 3 (typically bicubic).
```
dstansby, line 1: Can this get another sentence explaining the motivation for this breaking change?

docs/whatsnew/4.0.rst
```diff
@@ -31,8 +31,12 @@ Accordingly, the default behavior for these functions has been changed to use Na
 To obtain the prior behavior, which used zero by default for such pixels, the ``missing`` needs to be explicitly set to zero.
 
 The default rotation function used internally has been switched to :func:`scipy.ndimage.affine_transform` from :func:`skimage.transform.warp` due to changes in the latest release of `skimage`, although the differences between the rotation results are small.
+Since :func:`scipy.ndimage.affine_transform` has more consistent results between biquartic interpolation (``order=4``) and bicubic interpolation (``order=3``), we have changed the default interpolation order for :meth:`sunpy.map.GenericMap.rotate` to bicubic interpolation for improved performance under default settings.
```
dstansby, line 34: I don't understand this sentence, in particular:
> has more consistent results between biquartic interpolation (``order=4``) and bicubic interpolation

What does having consistent results between two methods mean?

sunpy/image/tests/test_transform.py
```diff
@@ -285,10 +285,12 @@ def test_clipping(rot30):
     image = np.ones((20, 20))
     image[4:-4, 4:-4] = 2
 
-    fig = Figure(figsize=(12, 4))
-    axs = fig.subplots(2, 5)
+    num_methods = len(_rotation_registry.keys())
 
-    for i, method in enumerate(['scipy', 'skimage']):
+    fig = Figure(figsize=(12, 2*num_methods))
+    axs = fig.subplots(num_methods, 5)
```
dstansby, line 291:
```suggestion
    axs = fig.subplots(nrows=num_methods, ncols=5)
```

sunpy/image/tests/test_transform.py
```diff
@@ -319,32 +319,38 @@ def test_nans(rot30):
...
 
-    fig = Figure(figsize=(16, 4))
-    axs = fig.subplots(2, 7)
+    num_methods = len(_rotation_registry.keys())
 
-    axs[0, 0].set_title('Original (NaNs are white)')
-    axs[0, 0].imshow(image_with_nans, vmin=-1.1, vmax=1.1)
-    axs[1, 0].imshow(image_with_nans, vmin=-1.1, vmax=1.1)
+    fig = Figure(figsize=(16, 2*num_methods))
+    axs = fig.subplots(num_methods, 7)
```
dstansby, line 325:
```suggestion
    axs = fig.subplots(nrows=num_methods, ncols=7)
```

sunpy/image/tests/test_transform.py
```diff
@@ -319,32 +319,38 @@ def test_nans(rot30):
...
     return fig
 
 
 @pytest.mark.filterwarnings("ignore:.*bug in the implementation of scikit-image")
-@pytest.mark.parametrize('method', _rotation_function_registry.keys())
+@pytest.mark.parametrize('method', _rotation_registry.keys())
 @pytest.mark.parametrize('order', range(6))
 def test_endian(method, order, rot30):
+    if order not in _rotation_registry[method].allowed_orders:
+        return
```
dstansby, line 352: Do a `pytest.skip` with a suitable message here instead?

sunpy/image/transform.py
```diff
@@ -115,7 +116,8 @@ def _get_transform_method(method, use_scipy):
     return method
 
 
-def add_rotation_function(name, handles_clipping, handles_image_nans, handles_nan_missing):
+def add_rotation_function(name, allowed_orders,
```
dstansby, line 119:
```suggestion
def add_rotation_function(name, *, allowed_orders,
```
We should definitely make the rest of these arguments keyword only

dstansby (approved):
Looks good - I've left some comments that shouldn't block feature freeze, so if they aren't read/replied to by this afternoon I will merge this and open a new issue with my comments.

docs/conf.py
Cadair: It's not particularly an issue putting it everywhere, and putting it everywhere means that you don't have to do anything to use it.

sunpy/image/transform.py
Cadair: I really don't like "cv2" it's completely impenetrable to people who haven't used opencv before, unlike opencv it doesn't give you sensible things when you search for it (unlike skimage which is the first result). While it might be the module name, I would argue that whoever decided that you should do `import cv2` was also not of sound mind.

Cadair (changes requested):
I am really struggling to get past `cv2` over `opencv`, and the arguments to `add_rotation_function` should be keyword only arguments.

docs/conf.py
ayshih: A different approach would be to programmatically add it to the specific docstring where it is needed, but that would be fiddly if we ever rename the function or want to use the shortcut in more places.  If this showed up in the rendered pages, it'd of course be untenable, but it's essentially invisible to the reader.

docs/whatsnew/4.0.rst
ayshih: This is saying that scipy's bicubic looks more like scipy's biquartic than skimage's bicubic looks like skimage's biquartic.  skimage's bicubic is fast, but can have noticeably poor quality.  That's why, when skimage was the default, we switched to biquartic as the default in #1185.

sunpy/image/tests/test_transform.py
ayshih: Hmm, yes, that could be done, but wouldn't that just add noise to the short summary of every test run?  There'd always be skips reported because these orders are simply not supported.  I'm inclined to consider them passes rather than skips.

sunpy/image/transform.py
ayshih: Note that these parameters are not keyword parameters and are non-optional, so there's no such thing as making them keyword-only.  These parameters are all positional – as suggested by @Cadair! – so that every function is required to state `True`/`False` for each of these functionalities, with no default value.

All usages of the decorator still name the parameters as if they were keyword parameters, both for clarity and to seamlessly support if we ever do make them optional.

sunpy/image/transform.py
ayshih: If I recall correctly, the reason that the package is called `cv2` is because they changed the API from their original `cv` package.  It has nothing to do with the version of OpenCV.

If we want, we could actually support both `'cv2'` and `'opencv'`.  We could also support both `'skimage'` and `'scikit-image'`.  Why not!

sunpy/image/transform.py
ayshih: Never mind.  I had an incomplete understanding of how the star worked.

ayshih: Fixed

sunpy/image/tests/test_transform.py
ayshih: Changed

sunpy/image/tests/test_transform.py
ayshih: Changed

docs/whatsnew/4.0.rst
ayshih: Reworded

ayshih:
As discussed on Element, the rotation method names have been changed from `'skimage'` and `'cv2'` to `'scikit-image'` and `'opencv'`.  Since the `method` keyword was only recently introduced (#5916) and hasn't actually been released, I did not add a changelog entry for the change to the string used to select `scikit-image`.  I wanted to modify that PR's changelog entry anyway because it explicitly stated that `scikit-image` was the default, but that is changed by this PR.

sunpy/image/tests/test_transform.py
nabobalis: I am with Albert on this, adding more skip test results at the end of the test output is not something I support especially since we might never be able to unskip them.

nabobalis (approved):

cbard:
Thanks to everyone here for their work in getting OpenCV into Sunpy! I also want to acknowledge Raphael Attie (@WaaallEEE), whose work during the HelioHackweek2020 demonstrated the utility of OpenCV as a worthwhile alternative to skimage and scipy and led me to create the original PR #4502.

PR #6090, opened 2022-04-21
Capture warnings in new test

ConorMacBride:
This fixes the warning capture in wheel tests (i.e. pytest with `-p no:warnings`) for a new test that was added. 

Also removes [a configuration option for `pytest-asyncio`](https://github.com/pytest-dev/pytest-asyncio#modes) which was causing a `PytestConfigWarning: Unknown config option: asyncio_mode` warning. I suspect it was added originally because someone also had `pytest-asyncio` installed in their SunPy environment, although we don't depend on it.

### DIFF
--- a/sunpy/coordinates/tests/test_frames.py
+++ b/sunpy/coordinates/tests/test_frames.py
@@ -176,13 +176,13 @@ def test_hpc_distance_off_limb():
     assert hpc1.Tx == 1500 * u.arcsec
     assert hpc1.Ty == 0 * u.arcsec
 
-    with pytest.raises(SunpyUserWarning, match="is all NaNs"):
+    with pytest.warns(SunpyUserWarning, match="is all NaNs"):
         hpc2 = hpc1.make_3d()
-        assert isinstance(hpc2._data, SphericalRepresentation)
-        # Check the attrs are correct
-        assert hpc2.Tx == 1500 * u.arcsec
-        assert hpc2.Ty == 0 * u.arcsec
-        assert_quantity_allclose(hpc2.distance, u.Quantity(np.nan, u.km))
+    assert isinstance(hpc2._data, SphericalRepresentation)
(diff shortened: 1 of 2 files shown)


sunpy/coordinates/tests/test_frames.py
```diff
@@ -176,13 +176,13 @@ def test_hpc_distance_off_limb():
     assert hpc1.Tx == 1500 * u.arcsec
     assert hpc1.Ty == 0 * u.arcsec
 
-    with pytest.raises(SunpyUserWarning, match="is all NaNs"):
+    with pytest.warns(SunpyUserWarning, match="is all NaNs"):
```
Cadair, line 179: This happens *all the time* is there a way we can make this error on our regular test suite?

Cadair (approved):

sunpy/coordinates/tests/test_frames.py
ConorMacBride: We tell pytest to promote warnings to errors so technically a `SunpyUserWarning` (a subclass of `Exception`) is *raised*. We could create a wrapper for (or monkeypatch?) `pytest.raises` such that it always raises an error if the class is a subclass of `Warning`, although that would require people to actually use the pytest wrapper.

Cadair: Yeah, that's the issue. I guess the easy way would be to make one of our test jobs not throw warnings as errors?

nabobalis:
> Also removes [a configuration option for `pytest-asyncio`](https://github.com/pytest-dev/pytest-asyncio#modes) which was causing a `PytestConfigWarning: Unknown config option: asyncio_mode` warning. I suspect it was added originally because someone also had `pytest-asyncio` installed in their SunPy environment, although we don't depend on it.

Guilty.

nabobalis (approved):

PR #6098, opened 2022-04-23
Use VCR to mock some online tests

dstansby:
This is a proof of concept revival of https://github.com/sunpy/sunpy/pull/4010, since it seems like using `vcrpy` for some of the remote data tests is a good idea. Currently just applied to two tests as a proof of concept, but I will go through and see which other tests we can add this to, and how much that brings down the runtime of the remote data tests.

Notes:
- As it stands, this knocks ~4 mintues of the online tests runtime
- We should check that the non-online build test times aren't imapacted too much by including the `vcrpy` tests. If they are, we can ignore the `vcrpy` tests on them.
- Currently the `vcrpy` cassettes are 27MB in size. If we go ahead I think we will need to use a new repository to put these in, similar to the `sunpy-figure-images` repository.


nabobalis:
> * Currently the `vcrpy` cassettes are 27MB in size. If we go ahead I think we will need to use a new repository to put these in, similar to the `sunpy-figure-images` repository.

Makes sense, I would like to prevent adding all these files to the main repo history. We could add them but then, don't package them to the sdist and make sure we skip those tests?

dstansby:
> Makes sense, I would like to prevent adding all these files to the main repo history. We could add them but then, don't package them to the sdist and make sure we skip those tests?

Are you suggesting we do put them in `sunpy/sunpy`, but make sure they're not packaged? Or putting them in a separate repository? I'm actually open to putting them in `sunpy/sunpy` as they're not too large (and ~2/3 of the 25MB are the SUVI files).

> I assume the goal is to mock every online test, never run them on PRs and then have an online cron job that tests the real servers?

To start with I think mocking every online test which doesn't download a data file is a good goal. I think I've done all those in this PR.

And yes, having a CRON job that tests the real servers, and automatically opens PRs to update the VCR cassettes if anything changes would be a great addition 👍

nabobalis:
> Are you suggesting we do put them in `sunpy/sunpy`, but make sure they're not packaged? Or putting them in a separate repository? I'm actually open to putting them in `sunpy/sunpy` as they're not too large (and ~2/3 of the 25MB are the SUVI files).

I am suggesting every possible combo. I am not sure what one I like more.

Cadair:
We can't put something that large in the main git history, we could use lfs, but I am reluctant to add lfs to the main repo due to contribution complexity. I am pro putting them in a second repo (also using lfs).

dstansby:
Sounds good. I'll probably hack away at this sporadically over a couple of months, but if anyone else wants to have a go feel free, but please drop a comment here so I know 😃

sunpy/net/cdaweb/test/test_cdaweb.py
```diff
@@ -5,17 +5,14 @@
 from sunpy.net.cdaweb.attrs import Dataset
 
 
-@pytest.mark.remote_data
+@pytest.mark.vcr()
```
Cadair, line 8: Why do we need the `()` on this, that's not a normal requirement for pytest marks?

Cadair:
Where is the size coming from here? They look like yaml files?

Figuring out how to get hold of the cassettes not in the main repo and skip the tests conditionally on it existing should be fun...

dstansby:
They are all yaml files, but the GOES SUVI requests return a lot of filenames, ~17k for the largest file which equates to 2.4MB of yaml 🙀

Cadair:
Wowzer. Could we compress them?

dstansby:
Yeah, we definitely could. Compressing everything I added here takes the total size down to 2.3MB. I guess then we lose having a nice text diff if/when the requests change, but 🤷

Cadair:
It also means git treating them as binary which means any change to the file makes a whole new copy, so not great for the history size either. I think we are stuck with another repo for this (though we wouldn't have to use git lfs as they are text).

samaloney:
It could be useful to see the diff of the cassettes too so a separate repo (without ifs) would be nice from that perspective?

dstansby:
I'm prototyping some CI in an external repository to keep these up to date - https://github.com/dstansby/sunpy-vcr-cassettes. Will update here when that's in good shape.

sunpy/net/cdaweb/test/test_cdaweb.py
nabobalis: You can normally add them and they behave the same as if they weren't there.

dstansby:
https://github.com/dstansby/sunpy-vcr-cassettes now runs all the VCR tests - it's currently pointing at one of my branches that has vcrpy enabled. See https://github.com/dstansby/sunpy-vcr-cassettes/commit/3b80ce51eb1b6cfb43bfb2fc0a39c45fd84515b0 as a sample commit where the cassettes are updated (for simplicity there's just one cassette).

The planned workflow in `sunpy` core for a new test is:
1. Create a new remote test in `sunpy`. Mark it as `@pytest.mark.remote_data` and `@pytest.mark.vcr`.
2. When the tests are run the new remote test will run as normal.

For existing tests I propose keeping the `@pytest.mark.remote_data`, and the workflow it is:
1. `sunpy` will clone the cassettes stored in `sunpy-vcr-cassettes`**TODO**
2. When a test with an existing cassette is run, it will use this instead of opening a remote connection **TODO**

For updating cassettes the workflow is:
1. `sunpy-vcr-cassettes` will run CI on a CRON schedule to update the cassettes from `sunpy@main` on a weekly schedule
2. If any of the pre-existing cassettes change, the CI will automatically open an issue on `sunpy-vcr-cassettes` detailing the change. **TODO**
3. We can either close these issues or take action as appropriate and depending on the change.

As you can see there's still a couple of todo items, but I'm posting this to save my thoughts, and see if anyone has any comments on my proposals.

nabobalis:
We discussed this on the community meeting, the plan looks good.

dstansby:
Replaced by https://github.com/sunpy/sunpy/pull/6167

PR #6099, opened 2022-04-24
Significant improvements in rotation performance

ayshih:
These are performance-related improvements to the decorator added in #5867.  These have the most noticeable impact when using `method='opencv'` (added in #6089), which was far slower than it should be.

- The external handling of image NaNs and `missing=np.nan` each requires an additional rotation.  The decorator currently uses `scipy.ndimage.affine_transform()` for these additional rotations regardless of the wrapped rotation function.  Now, the decorator re-uses the rotation function: e.g., `method='opencv'` uses `cv2.warpAffine()` for rotating the image *and* for the external handling.
- The external handling of image NaNs (which includes map rotations with `missing=np.nan`) requires a convolution.  The decorator currently uses `scipy.signal.convolve2D()`, but OpenCV's convolution (`cv2.filter2D()`) can be dramatically faster (and gives the identical result).  Now, the decorator checks if OpenCV is installed, and if so, uses its convolution.
- Added some performance notes to the rotation docstrings about the computational cost of NaN handling.
- Added debug log output to rotation to easily check how long each step takes


ayshih:
> Now, the decorator checks if OpenCV is installed, and if so, uses its convolution.

This one's a little weird in the context of OpenCV not even being an optional dependency.  If a user wants to speed up their `method='scipy'` rotation when NaNs need to be handled, they can install OpenCV for the faster convolution.  I don't know how/where to describe this.

nabobalis:
> > Now, the decorator checks if OpenCV is installed, and if so, uses its convolution.
> 
> This one's a little weird in the context of OpenCV not even being an optional dependency. If a user wants to speed up their `method='scipy'` rotation when NaNs need to be handled, they can install OpenCV for the faster convolution. I don't know how/where to describe this.

Would a note in the decorator function docstring be enough?

sunpy/image/transform.py
```diff
@@ -22,6 +24,7 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
 
     This function supports NaN values in the input image and supports using NaN for
     pixels in the output image that are beyond the extent of the input image.
+    Handling NaN values in the input image requires additional computation time.
```
Cadair, line 27: I am kind of amazed that no affine transform implementation just handles nans :(

Cadair (approved):
Seems sensible to me. :+1: to what Nabil said about putting a note in the docstring about opencv, I wouldput it in both `GenericMap` and `affine_transform`.

Should we get this into 4.0?

Cadair:
Ready for review? ;)

sunpy/image/transform.py
ayshih: The scikit-image interpolation orders that are handled by native Cython code – nearest-neighbor, bi-linear, and bi-cubic – do handle image NaNs.  However, the current implementation is an all-or-nothing approach; since half of the allowed orders do not handle NaNs (the orders that call SciPy code), we can't mark scikit-image as `handles_image_nans=True`.

SciPy's `affine_transform()` is actually two steps: a spline pre-filtering step and a polynomial interpolation.  The polynomial interpolation handles image NaNs fine.  The problem is the spline pre-filtering, which turns the entire image to NaNs if any pixel is NaN.  And, apparently, that is formally correct: when it comes to spline pre-filtering, every pixel in the image contributes at some level to the resulting spline interpolation, no matter how far away that pixel is.

ayshih:
Ready for review, and this should definitely go into 4.0.  All this changes are fixes to the very recently merged #5867 and #6089, so I don't know if any of this warrants new changelog entries.

nabobalis (approved):

PR #6101, opened 2022-04-24
Improve performance of timerange extraction in scraper

dstansby:
I was digging into why some SUVI tests take so long to run (even with `vcrpy`, ie. ignoring the remote response times), and when there are a lot of URLs to parse, `get_timerange_from_exdict` is the bottleneck, as it is run on every URL. Most of the time was spent in `astropy.time.Time` logic, but this is unnesseary as this code doesn't use any special features of `Time`, so:

- switch to just using `datetime`
- pre-compute the deltas
- re-order the many `if` statements so they're not all executed if seconds are specified in the timestamp


samaloney:
This looks great - I have a vague recollection of something about leap seconds which was why we use `astropy.time.Time` but I can't remember or see how it would make a difference.

nabobalis (approved):

nabobalis:
pre-commit.ci autofix

nabobalis:
The SRS remote tests are failing on main. Otherwise the other online fails are flaky ones.

PR #1750, opened 2016-04-25
First pass at allowing fits memmapping in io and map.

Cadair:
This is one approach to solving the files open problem see #1745

PR #6110, opened 2022-04-28
Memmap tests for io

### LINKED ISSUE #1774: Add tests for memory mapping in sunpy.io
We forgot to add tests in the #1750

akash5100:
## PR Description
In the PR [#1750](https://github.com/sunpy/sunpy/pull/1750), to make SunPy use memory mapping for access to files, changes in the read functions of the `sunpy.io` were made. This PR adds test for those read functions, which are:
- ana.py
- file_tools.py
- fits.py
- jp2.py

Merging this PR:
* closes #1774

### DIFF
--- a/sunpy/io/tests/test_ana.py
+++ b/sunpy/io/tests/test_ana.py
@@ -79,7 +79,7 @@ def test_f32c():
 
 @skip_ana
 def test_read_memmap():
-    # Test to check memmap is passed to the reader
+    # Test to check that passing memmap=True doesn't raise an error
     afilename = tempfile.NamedTemporaryFile().name
     ana.write(afilename, img_f32, 'testcase', 0)
     img_f32c_rec = ana.read(afilename, memmap=True)
--- a/sunpy/io/tests/test_ana.py
+++ b/sunpy/io/tests/test_ana.py
@@ -3,6 +3,7 @@ import tempfile
 
 import numpy as np
 import pytest
+import mmap
 
 from sunpy.io import ana
 from sunpy.tests.helpers import skip_ana
@@ -84,3 +85,7 @@ def test_read_memmap():
     ana.write(afilename, img_f32, 'testcase', 0)
     img_f32c_rec = ana.read(afilename, memmap=True)
     assert np.sum(img_f32c_rec[0][0] - img_f32) == 0
+
+    # Test to check that passing memmap=False doesn't raise an error
+    img_f32c_rec = ana.read(afilename, memmap=False)
+    assert not isinstance(img_f32c_rec, mmap.mmap)
--- a/sunpy/io/tests/test_ana.py
+++ b/sunpy/io/tests/test_ana.py
@@ -79,6 +79,7 @@ def test_f32c():
 
 @skip_ana
 def test_read_memmap():
+    # Test to check memmap is passed to the reader
     afilename = tempfile.NamedTemporaryFile().name
     ana.write(afilename, img_f32, 'testcase', 0)
     img_f32c_rec = ana.read(afilename, memmap=True)
--- a/sunpy/io/tests/test_ana.py
+++ b/sunpy/io/tests/test_ana.py
@@ -75,3 +75,12 @@ def test_f32c():
     afilename = tempfile.NamedTemporaryFile().name
     with pytest.raises(RuntimeError):
         ana.write(afilename, img_f32, 'testcase', 1)
+
+
+@skip_ana
+def test_memmap():
+    # Test if float 32 compressed functions
+    afilename = tempfile.NamedTemporaryFile().name
+    ana.write(afilename, img_f32, 'testcase', 0)
+    img_f32c_rec = ana.read(afilename, memmap=True)
+    assert np.sum(img_f32c_rec[0][0] - img_f32) == 0
--- a/sunpy/io/tests/test_ana.py
+++ b/sunpy/io/tests/test_ana.py
@@ -79,7 +79,6 @@ def test_f32c():
 
 @skip_ana
 def test_memmap():
-    # Test if float 32 compressed functions
     afilename = tempfile.NamedTemporaryFile().name
     ana.write(afilename, img_f32, 'testcase', 0)
     img_f32c_rec = ana.read(afilename, memmap=True)
--- a/sunpy/io/tests/test_ana.py
+++ b/sunpy/io/tests/test_ana.py
@@ -78,7 +78,7 @@ def test_f32c():
 
 
 @skip_ana
-def test_memmap():
+def test_read_memmap():
     afilename = tempfile.NamedTemporaryFile().name
     ana.write(afilename, img_f32, 'testcase', 0)
     img_f32c_rec = ana.read(afilename, memmap=True)
--- a/sunpy/io/tests/test_ana.py
+++ b/sunpy/io/tests/test_ana.py
@@ -1,9 +1,9 @@
 
+import mmap
 import tempfile
 
 import numpy as np
 import pytest
-import mmap
 
 from sunpy.io import ana
 from sunpy.tests.helpers import skip_ana
--- a/sunpy/io/tests/test_fits.py
+++ b/sunpy/io/tests/test_fits.py
@@ -184,6 +184,3 @@ def test_read_memmap():
     # Create memmap array with data and header
     mm = np.memmap(AIA_171_IMAGE, dtype=data.dtype, mode='r', shape=data.shape)
     assert isinstance(mm, np.memmap)
-
-
-test_read_memmap()
--- a/sunpy/io/tests/test_fits.py
+++ b/sunpy/io/tests/test_fits.py
@@ -1,3 +1,4 @@
+import mmap
 from pathlib import Path
 from collections import OrderedDict
 
@@ -178,10 +179,13 @@ def test_warn_longkey():
     assert 'BADLONGKEY' not in fits.keys()
 
 
-def test_read_memmap():
-    # Test that memmap=False gives the same data as memmap=True
-    data, header = sunpy.io.read_file(AIA_171_IMAGE, memmap=False)[0]
-    data_memmap, header_memmap = sunpy.io.read_file(AIA_171_IMAGE, memmap=True)[0]
+def is_memmap(data):
+    if data.base is None:
+        return False
+    return isinstance(data.base, mmap.mmap) or is_memmap(data.base)
 
-    assert np.all(data == data_memmap)
-    assert np.all(header == header_memmap)
+
+def test_read_memmap():
+    # Check that memmap is passed as an argument to read function doesnot raise an error.
+    data, header = sunpy.io._fits.read(AIA_171_IMAGE, memmap=True)[0]
+    assert is_memmap(data)
--- a/sunpy/io/tests/test_fits.py
+++ b/sunpy/io/tests/test_fits.py
@@ -180,7 +180,7 @@ def test_warn_longkey():
 
 def test_read_memmap():
     # Test that memmap is passed correctly to the FITS reader
-    # Expecting ValueError: not enough values to unpack (expected 2, got 1)
-    with pytest.raises(ValueError):
-        data, header = sunpy.io._fits.read(AIA_171_IMAGE, memmap=True)
-        assert isinstance(data, np.memmap)
+    data, header = sunpy.io._fits.read(AIA_171_IMAGE, memmap=True)[0]
+    # Create memmap array
+    mm = np.memmap(AIA_171_IMAGE, dtype='float32', mode='r', shape=data.shape)
+    assert isinstance(mm, np.memmap)
--- a/sunpy/io/tests/test_fits.py
+++ b/sunpy/io/tests/test_fits.py
@@ -182,14 +182,10 @@ def test_warn_longkey():
 def test_read_memmap():
     # Check that the FITS reader can read a memmap without memmap argument by default
     data, header = sunpy.io._fits.read(TEST_AIA_IMAGE)[0]
-    if data.base is None:
-        assert False
     assert isinstance(data.base, mmap.mmap)
 
     # Check that memmap=True does the same thing
     data, header = sunpy.io._fits.read(TEST_AIA_IMAGE, memmap=True)[0]
-    if data.base is None:
-        assert False
     assert isinstance(data.base, mmap.mmap)
 
     # Check that memmap=False doesn't do memory mapping
--- a/sunpy/io/tests/test_fits.py
+++ b/sunpy/io/tests/test_fits.py
@@ -180,8 +180,12 @@ def test_warn_longkey():
 
 
 def test_read_memmap():
-    # Check that memmap is read correctly from a file
+    # Check that memmap=True does memory map to the file
     data, header = sunpy.io._fits.read(TEST_AIA_IMAGE, memmap=True)[0]
     if data.base is None:
         assert False
-    assert isinstance(data.base, mmap.mmap) or isinstance(data.base, np.memmap)
+    assert isinstance(data.base, mmap.mmap)
+
+    # Check that memmap=False doesn't do memory mapping
+    data, header = sunpy.io._fits.read(TEST_AIA_IMAGE, memmap=False)[0]
+    assert data.base is None
--- a/sunpy/io/tests/test_fits.py
+++ b/sunpy/io/tests/test_fits.py
@@ -179,7 +179,9 @@ def test_warn_longkey():
 
 
 def test_read_memmap():
-    # Test that memmap is passed correctly to the FITS reader
-    data, header = sunpy.io._fits.read(AIA_171_IMAGE, memmap=True)[0]
-    assert isinstance(data, np.ndarray)
-    assert isinstance(header, sunpy.io.header.FileHeader)
+    # Test that memmap=False gives the same data as memmap=True
+    data, header = sunpy.io.read_file(AIA_171_IMAGE, memmap=False)[0]
+    data_memmap, header_memmap = sunpy.io.read_file(AIA_171_IMAGE, memmap=True)[0]
+
+    assert np.all(data == data_memmap)
+    assert np.all(header == header_memmap)
--- a/sunpy/io/tests/test_fits.py
+++ b/sunpy/io/tests/test_fits.py
@@ -1,4 +1,3 @@
-from cgi import print_directory
 from pathlib import Path
 from collections import OrderedDict
 
--- a/sunpy/io/tests/test_fits.py
+++ b/sunpy/io/tests/test_fits.py
@@ -19,7 +19,7 @@ EIT_195_IMAGE = get_test_filepath('EIT_header/efz20040301.000010_s.header')
 AIA_171_IMAGE = get_test_filepath('aia_171_level1.fits')
 SWAP_LEVEL1_IMAGE = get_test_filepath('SWAP/resampled1_swap.header')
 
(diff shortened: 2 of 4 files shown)


akash5100:
I know many things are left to do to make this PR complete, just wanted to ask am I going the right way?

sunpy/io/tests/test_ana.py
```diff
@@ -75,3 +75,11 @@ def test_f32c():
     afilename = tempfile.NamedTemporaryFile().name
     with pytest.raises(RuntimeError):
         ana.write(afilename, img_f32, 'testcase', 1)
+
+
+@skip_ana
+def test_memmap():
+    afilename = tempfile.NamedTemporaryFile().name
+    ana.write(afilename, img_f32, 'testcase', 0)
+    img_f32c_rec = ana.read(afilename, memmap=True)
```
nabobalis, line 84: Does the ana reader support memmap?

nabobalis:
> I know many things are left to make this PR complete, just wanted to ask am I going the right way?

I would say on the surface, yes.

sunpy/io/tests/test_ana.py
akash5100: In the file ana.py with PR #1750, the following line was added
```py
def read(filename, debug=False, **kwargs):```
So I thought I have to add a test for that.

Comment (nabobalis) [reply sunpy/io/tests/test_ana.py:84]: What happens to those kwargs in the read function?

**Comment by akash5100** [comment]
Do I need a changelog file for adding tests? 
If yes, please tell me which type of changelog.

File: sunpy/io/tests/test_ana.py (hunk 1 above)
Comment (akash5100) [reply sunpy/io/tests/test_ana.py:84]: So when `memmap` is passed as a Keyword Arguments in the read function, this test will make sure that everything will stay fine using `assert`.

Comment (nabobalis) [reply sunpy/io/tests/test_ana.py:84]: Where do they end up?

**Comment by nabobalis** [comment]
> Do I need a changelog file for adding tests? If yes, please tell me which type of changelog.

No, they don't need a changelog.

File: sunpy/io/tests/test_ana.py (hunk 1 above)
Comment (akash5100) [reply sunpy/io/tests/test_ana.py:84]: Sorry, I don't understand what you mean?

Comment (akash5100) [reply sunpy/io/tests/test_ana.py:84]: In the docs here, https://docs.sunpy.org/en/stable/code_ref/io.html?highlight=io#id3
I don't find anything related to memmap, so do you think `ana.py` doesn't need this test?

Comment (akash5100) [reply sunpy/io/tests/test_ana.py:84]: > In the docs here, https://docs.sunpy.org/en/stable/code_ref/io.html?highlight=io#id3 I don't find anything related to memmap, so do you think `ana.py` doesn't need this test?

cc @Cadair

Comment (nabobalis) [reply sunpy/io/tests/test_ana.py:84]: ANA most likely does not support memmap (you would have to check), but it's good to see that it doesn't error with the keyword set. 

You will need to replicate this for all the other readers and file types as well.

Comment (akash5100) [reply sunpy/io/tests/test_ana.py:84]: In the PR [#1750](https://github.com/sunpy/sunpy/pull/1750) read function of 4 files was changed, they are:
- ana.py
- file_tools.py
- fits.py
- jp2.py

So, I will be writing tests for these four read functions.

Comment (ConorMacBride) [reply sunpy/io/tests/test_ana.py:84]: The documentation for [`sunpy.io.read_file`](https://docs.sunpy.org/en/latest/generated/api/sunpy.io.read_file.html#sunpy.io.read_file) says that `memmap` "is currently only supported by the FITS reader", i.e., `sunpy.io._fits.read` for which you have added a test which should work very nicely! (Although that test is currently erroring, maybe double check the type of object that `sunpy.io._fits.read` returns to see why it says [`not enough values to unpack (expected 2, got 1)`](https://github.com/sunpy/sunpy/runs/6231810970?check_suite_focus=true#step:8:5137).)

The rest of the changes to the read functions in that PR were to allow `memmap=` to be passed to the [`sunpy.io.read_file`](https://docs.sunpy.org/en/latest/generated/api/sunpy.io.read_file.html#sunpy.io.read_file) function and not have an error raised by a `sunpy.io.*.read` function if `memmap=` was passed but wasn't supported in it. 

So I guess the rest of the tests should just be to see that it doesn't raise an error if `memmap` is passed but unsupported. Although maybe sunpy should be presenting the user with a warning instead?

Comment (akash5100) [reply sunpy/io/tests/test_ana.py:84]: > Although maybe sunpy should be presenting the user with a warning instead?

How do we know that the file supports memmap?

Comment (nabobalis) [reply sunpy/io/tests/test_ana.py:84]: You should be able to check each reader and see if they support memory mapping. As Conor has said, I would assume only the FITS reader supports it.

File: sunpy/io/tests/test_fits.py [hunk 2]
```diff
@@ -176,3 +177,11 @@ def test_warn_longkey():
                                'goodkey': 'test'})
     assert 'GOODKEY' in fits.keys()
     assert 'BADLONGKEY' not in fits.keys()
+
+
+def test_read_memmap():
+    # Test that memmap is passed correctly to the FITS reader
+    # Expecting ValueError: not enough values to unpack (expected 2, got 1)
```
ConorMacBride, line 184: The ValueError should be due to a bug in the test function itself, as using memmap is not expected to produce an error.

sunpy/io/tests/test_fits.py
```diff
@@ -1,5 +1,6 @@
 from pathlib import Path
 from collections import OrderedDict
+from re import A
```
ConorMacBride, line 3: Don't think this import is used?

sunpy/io/tests/test_fits.py
akash5100: We need 2 parameters in the `sunpy.io._fits.read()` function,

filepath : `str`
The fits file to be read.

I passed `AIA_193_JP2`.

Can you help me, I don't know how to get the second parameter which is:

hdus : `int` or iterable
The HDU indexes to read from the file.

ConorMacBride: The second parameter is optional. (It's for selecting a specific array of data inside the FITS file if it contains multiple but you only need some of them.) The unpack error is occurring when Python tries to unpack the output of the read function in the test: e.g. `a, b = [1, 2]` works but `a, b = [[1, 2]]` produces an error because the outer object has length 1.

akash5100: Done!
Do the assertion test looks good?

sunpy/io/tests/test_fits.py
```diff
@@ -176,3 +176,11 @@ def test_warn_longkey():
                                'goodkey': 'test'})
     assert 'GOODKEY' in fits.keys()
     assert 'BADLONGKEY' not in fits.keys()
+
+
+def test_read_memmap():
+    # Test that memmap is passed correctly to the FITS reader
+    data, header = sunpy.io._fits.read(AIA_171_IMAGE, memmap=True)[0]
+    # Create memmap array
+    mm = np.memmap(AIA_171_IMAGE, dtype='float32', mode='r', shape=data.shape)
```
ConorMacBride, line 185: The test should verify that passing memmap=True to the read function opens the file as a memmap. By calling np.memmap explicitly, I don't think this actually tests the memmap functionally of the read function. Was the "data" variable not a subclass of some memmap class?

akash5100: data = <class 'numpy.ndarray'>
header = <class 'sunpy.io.header.FileHeader'>

sunpy/io/tests/test_fits.py
```diff
@@ -176,3 +176,11 @@ def test_warn_longkey():
...
     assert 'GOODKEY' in fits.keys()
     assert 'BADLONGKEY' not in fits.keys()
+
+
+def test_read_memmap():
+    # Test that memmap is passed correctly to the FITS reader
+    data, header = sunpy.io._fits.read(AIA_171_IMAGE, memmap=True)[0]
+    # Create memmap array with data and header
+    mm = np.memmap(AIA_171_IMAGE, dtype=data.dtype, mode='r', shape=data.shape)
+    assert isinstance(mm, np.memmap)
```
nabobalis, line 186: I would suggest looking at this test and asking yourself, what are you testing?

akash5100: Sorry for that, is it good now?

sunpy/io/tests/test_fits.py
```diff
@@ -176,3 +176,10 @@ def test_warn_longkey():
                                'goodkey': 'test'})
     assert 'GOODKEY' in fits.keys()
     assert 'BADLONGKEY' not in fits.keys()
+
+
+def test_read_memmap():
+    # Test that memmap is passed correctly to the FITS reader
+    data, header = sunpy.io._fits.read(AIA_171_IMAGE, memmap=True)[0]
+    assert isinstance(data, np.ndarray)
```
nabobalis, line 184: So this isnt a memory mapped array?

sunpy/io/tests/test_fits.py
nabobalis: Yes, much better.

sunpy/io/tests/test_fits.py
akash5100: I thought, data is a ndarray together with io.header it makes memmap.

nabobalis: data, header pair is what the reader returns. I had assumed that a correctly memory mapped array would have a different type?

akash5100: I think, we can use [np.all](https://numpy.org/doc/stable/reference/generated/numpy.all.html) to check whether both gives same result, 
So, the test will be like:
 ```py
    # Test that memmap=False gives the same data as memmap=True
    data, header = sunpy.io.read_file(AIA_171_IMAGE, memmap=False)[0]
    data_memmap, header_memmap = sunpy.io.read_file(AIA_171_IMAGE, memmap=True)[0]
    assert np.all(data == data_memmap)
    assert np.all(header == header_memmap)
```

Comment (ayshih) [reply sunpy/io/tests/test_fits.py:184]: Testing for value equality is not a useful test.  The way that memory mapping works in `astropy.io.fits` is that a `mmap.mmap` instance is created and then passed as a buffer to `np.ndarray()`.  What you need to do is to find and check the base buffer.  Here's a simple utility to check whether a `mmap.mmap` instance is being used:
```python
def is_memmap(data):
    if data.base is None:
        return False
    return isinstance(data.base, mmap.mmap) or is_memmap(data.base)
```
Be aware that reading in a FITS file that uses image compression or image scaling will – by default – defeat memory mapping and create the whole data array in memory.  For example, you need to open `AIA_171_IMAGE` with `disable_image_compression=True` to actually use memory mapping.  (To disable image scaling, you'd pass in `do_not_scale_image_data=True`.)

As an aside, you shouldn't use sample data like `AIA_171_IMAGE` for unit tests.  We have specific test files.

nabobalis: Just to note (as I made the same assumption), the `AIA_171_IMAGE` is actually a test file.

ayshih: Oh, sorry, okay.  The variable shouldn't be named that to avoid confusion for code maintenance.

nabobalis: It does need to be renamed.

akash5100: > To disable image compression, you'd pass in `disable_image_compression=True`. To disable image scaling, you'd pass in `do_not_scale_image_data=True`

Pass them to the `sunpy.io._fits.read()` or `sunpy.io.read_file()` function? (I don't find anything related to these arguments in the sunpy documentation)

ayshih: \<sigh\>  Apparently neither.  For some reason, `sunpy.io._fits.read()` (which is called by `sunpy.io.read_file()`) fails to pass `kwargs` on to the `astropy.io.fits.open()` call.  

Fortunately, you don't need to disable either for the AIA test file, so it's not necessary for this PR.

nabobalis: I have a PR for this issue: https://github.com/sunpy/sunpy/pull/6123

akash5100: > def is_memmap(data):

Can I add this function somewhere and then import it into the test files?

nabobalis: Yes that is ok.

akash5100: > It does need to be renamed and we can do it in this PR.


- `RHESSI_IMAGE` to `TEST_RHESSI_IMAGE`
- `AIA_171_IMAGE` to `TEST_AIA_IMAGE`
- `EIT_195_IMAGE` to `TEST_EIT_HEADER`
- `SWAP_LEVEL1_IMAGE` to `TEST_SWAP_HEADER` 

Should I do these renames?

nabobalis: Sure

sunpy/io/tests/test_fits.py
```diff
@@ -176,3 +177,11 @@ def test_warn_longkey():
...
     assert 'GOODKEY' in fits.keys()
     assert 'BADLONGKEY' not in fits.keys()
+
+
+def test_read_memmap():
+    # Check that memmap is read correctly from a file
+    data, header = sunpy.io._fits.read(TEST_AIA_IMAGE, memmap=True)[0]
+    if data.base is None:
+        assert False
+    assert isinstance(data.base, mmap.mmap) or isinstance(data.base, np.memmap)
```
nabobalis, line 187: Do we need both asserts and the `or`?

sunpy/io/tests/test_jp2.py
```diff
@@ -36,3 +36,12 @@ def test_read_file():
...
     assert isinstance(map_, GenericMap)
+
+
+@skip_glymur
+def test_read_memmap():
+    """
+    Passing memmap as keyword argument should not raise an error.
+    """
+    map_ = Map(AIA_193_JP2, memmap=True)
+    assert isinstance(map_, GenericMap)
```
nabobalis, line 47: Do we not want to check the data is memory mapped?

ConorMacBride: Only the FITS reader supports the `memmap` kwarg, so whether the data is actually memory mapped or not doesn't depend on the value of `memmap` passed.

nabobalis: Right, I for some reason though this was a fits file.

sunpy/io/tests/test_fits.py
```diff
@@ -176,3 +177,11 @@ def test_warn_longkey():
                                'goodkey': 'test'})
     assert 'GOODKEY' in fits.keys()
     assert 'BADLONGKEY' not in fits.keys()
+
+
+def test_read_memmap():
+    # Check that memmap is read correctly from a file
+    data, header = sunpy.io._fits.read(TEST_AIA_IMAGE, memmap=True)[0]
```
ConorMacBride, line 184: As the default for astropy's fits reader is to use memmap, this test would also pass if `memmap=True` was not included in this function call. Should we also be testing that `memmap=False` *doesn't* do memory mapping?

ayshih: Yes, we definitely should test `memmap=False` as well.

sunpy/io/tests/test_fits.py
ayshih: We don't need to check `isinstance(data.base, np.memmap)`.  `astropy.io.fits.open()` uses only `mmap.mmap`, so this unit test doesn't need to anticipate a possible future where `np.memmap` might be used.

ayshih:
> ANA most likely does not support memmap (you would have to check), but it's good to see that it doesn't error with the keyword set.

> Although maybe sunpy should be presenting the user with a warning instead?

I'd go one step further.  All readers that do not support memory mapping (i.e., everything but FITS) *should error* if `memmap=True` is passed in.

I don't understand why any of the non-FITS readers have `**kwargs` in the call signature for `read()`.  They don't use them at all, and it prevents any visibility of whether `read_file()` is passing in inappropriate keywords.  @Cadair: can you defend this pattern from #1750?

sunpy/io/tests/test_fits.py
akash5100: Sure

sunpy/io/tests/test_fits.py
akash5100: Sure.

sunpy/io/tests/test_ana.py
```diff
@@ -75,3 +75,12 @@ def test_f32c():
...
         ana.write(afilename, img_f32, 'testcase', 1)
+
+
+@skip_ana
+def test_read_memmap():
+    # Test to check memmap is passed to the reader
+    afilename = tempfile.NamedTemporaryFile().name
+    ana.write(afilename, img_f32, 'testcase', 0)
+    img_f32c_rec = ana.read(afilename, memmap=True)
+    assert np.sum(img_f32c_rec[0][0] - img_f32) == 0
```
nabobalis, line 86: Can we check that this array is not memory mapped?

sunpy/io/tests/test_fits.py
```diff
@@ -176,3 +177,21 @@ def test_warn_longkey():
                                'goodkey': 'test'})
     assert 'GOODKEY' in fits.keys()
     assert 'BADLONGKEY' not in fits.keys()
+
+
+def test_read_memmap():
+    # Check that the FITS reader can read a memmap without memmap argument by default
+    data, header = sunpy.io._fits.read(TEST_AIA_IMAGE)[0]
+    if data.base is None:
+        assert False
```
nabobalis, line 186: Do we need this?

sunpy/io/tests/test_fits.py
```diff
@@ -176,3 +177,21 @@ def test_warn_longkey():
...
+    assert isinstance(data.base, mmap.mmap)
+
+    # Check that memmap=True does the same thing
+    data, header = sunpy.io._fits.read(TEST_AIA_IMAGE, memmap=True)[0]
+    if data.base is None:
+        assert False
```
nabobalis, line 192: Do we need this?

sunpy/io/tests/test_fits.py
akash5100: Oh, we don't need that, whenever the `data.base` is `None` the test will fail.

sunpy/io/tests/test_ana.py
akash5100: One question arises, Do we need memmap testing for `ana` and `jp2` readers if only the FITS reader supports it?

nabobalis: We should at at least check the keyword doesn't error and the array is not a memory map.

nabobalis:
> I don't understand why any of the non-FITS readers have `**kwargs` in the call signature for `read()`. They don't use them at all, and it prevents any visibility of whether `read_file()` is passing in inappropriate keywords. @Cadair: can you defend this pattern from #1750?

I assume it's the same principle as our Fido Clients, we don't know what extra keywords need to be passed internally that only certain readers might need. 

I wonder if we should make the entire io module private and then we can change all of this internally.

sunpy/io/tests/test_ana.py
akash5100: I added the following, is it correct?

nabobalis: Yeah

akash5100:
pre-commit.ci autofix

nabobalis:
Thanks for the PR @akash5100

PR #6123, opened 2022-05-05
fix kwarg for read

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/6122

PR #6125, opened 2022-05-05
update read function docstrings

nabobalis:
See https://github.com/sunpy/sunpy/pull/6123#pullrequestreview-963797603


ayshih:
Note that the first note under "Notes" for both `read_file()` and `write_file()` already notes that keywords are passed on, so those notes can be removed as redundant notes.

Issue #6094, opened 2022-04-22
Automate updating vendored code in `sunpy/extern`


dstansby:
### Describe the feature

I'm currently manually checking for updates of vendored code in `/sunpy/extern`. It should be pretty easy to automate this in a simple script in the `/tools` folder that downloads the latest version of the files, and prints which files have been updated (and possibly the version number bumps?)

akash5100:
Hi @dstansby, I would like to work on this.
I don't understand which part of `/sunpy/extern` must be automated, can you guide me through this?
All that I understood is, that we are going to automate the process of updating versions, with the help of a script that will be in `/tools`. 
I have a question, when will this script in `/tools` will be executed?

nabobalis:
> Hi @dstansby, I would like to work on this. I don't understand which part of `/sunpy/extern` must be automated, can you guide me through this? All that I understood is, that we are going to automate the process of updating versions, with the help of a script that will be in `/tools`.

Yes that is correct, what would you like to know in addition to this? 

> I have a question, when will this script in `/tools` will be executed?

It will be run by the person doing a sunpy release.

akash5100:
which files in `/extern` should be automated?

nabobalis:
> which files in `/extern` should be automated?

Every python file expect `modest_image.py`, I don't think that has a proper place to update it. The licences should be fine to leave alone.

akash5100:
Automate these changes?
https://github.com/sunpy/sunpy/commit/5ba9c9141e3364529138f8451225a97bb195851d
https://github.com/sunpy/sunpy/commit/9f5427cc4939db023a36092b6bb43cd3505249d8
https://github.com/sunpy/sunpy/commit/9ab27122dd96b69f88831c54106ed251f6bdee09

These commits also have changes that are different from changing versions (I mean, some changes change function), how are we going to do that?

What should I use to download the latest files and check the versions?

nabobalis:
> Automate these changes? [5ba9c91](https://github.com/sunpy/sunpy/commit/5ba9c9141e3364529138f8451225a97bb195851d) [9f5427c](https://github.com/sunpy/sunpy/commit/9f5427cc4939db023a36092b6bb43cd3505249d8) [9ab2712](https://github.com/sunpy/sunpy/commit/9ab27122dd96b69f88831c54106ed251f6bdee09)
> 

Yes to create a script that will for the packages we have in the `sunpy/extern` folder, update them from their canonical source.

[...]

PR #6127, opened 2022-05-06
Adds a script for updating vendored code in `sunpy/extern`

akash5100:
# DRAFT
## PR Description
The PR will close 
* #6094 

This PR will add a script in the `sunpy/tools` that will be used to update vendor code automatically, by the person doing a release. Executing this script will update all python files in `sunpy/sunpy/extern`.

## Approach

- [ ] Get a list of all the packages imported in sunpy/extern
- [ ] Search for the source distribution of each package (TBD, by Github API or PyPI)
- [ ] Download the source distribution in `USER/home/sunpy/extern` (A temporary directory to store files)
- [ ] Extract the source distribution
- [ ] Update the python files in `sunpy/extern` (Need guidance)

### DIFF
--- a/tools/update_extern.py
+++ b/tools/update_extern.py
@@ -1,0 +1,56 @@
+#!/bin/env python
+
+"""
+This script is used to update the files in `sunpy/extern` while making new releases.
+"""
+
+# TODO:
+# 1. Get list of all the packages imported in sunpy/extern
+# 2. Search for the source distribution of each package in `USER/home/sunpy/extern`
+# 3. Extract the source distribution to a temporary directory
+# 4. Update the python files in `sunpy/extern`
+
+
+import ast
+from pathlib import Path
+
+# get parent directory of sunpy
+SUNPY_DIR = Path(__file__).parent.parent
+
+
+def visit_import(node):
+    for name in node.names:
+        modules.append(name.name.split(".")[0])
+
+
+def visit_from(node):
+    # if node.module is missing it's a "from . import ..." statement
+    # if level > 0 it's a "from .submodule import ..." statement
+    if node.module is not None and node.level == 0:
+        modules.append(node.module.split(".")[0])
+
+
+# get all the modules imported in sunpy/extern
+file_names = ["appdirs", "distro", "inflect", "parse"]
+
+modules = list()
+
+for i in file_names:
+    with open(SUNPY_DIR/f"sunpy/extern/{i}.py", "r") as f:
+        # print(f"Reading {i}.py")
+        tree = ast.parse(f.read())
+        for node in ast.walk(tree):
+            if isinstance(node, ast.Import):
+                visit_import(node)
+            elif isinstance(node, ast.ImportFrom):
+                visit_from(node)
+
+# remove python modules from the list
+modules = [i for i in modules if i not in ['platform', 'warnings', 're', 'ctypes', 'os',
+                                           'shlex', 'functools', 'ast',
+                                           'datetime', 'logging', 'sys', 'com', 'decimal', 'argparse', '__future__', 'json', 'typing', 'subprocess', 'array']]
+modules = set(modules)
+print(modules)
+
+# Output:
+# {'win32api', 'win32com', 'winreg', '_winreg'}
--- a/tools/update_extern.py
+++ b/tools/update_extern.py
@@ -5,10 +5,10 @@ This script is used to update the files in `sunpy/extern` while making new relea
 """
 
 # TODO:
-# 1. Get list of all the packages imported in sunpy/extern
-# 2. Search for the source distribution of each package in `USER/home/sunpy/extern`
-# 3. Extract the source distribution to a temporary directory
-# 4. Update the python files in `sunpy/extern`
+# [x] Get list of all the packages imported in sunpy/extern
+# [] Search for the source distribution of each package in `USER/home/sunpy/extern`
+# [] Extract the source distribution to a temporary directory
+# [] Update the python files in `sunpy/extern`
 
 
 import ast


ConorMacBride (commented):
Hi @akash5100, thank you for working on this PR! Looking through your script, it seems that you are attempting to update the *dependencies* of the packages in `sunpy/sunpy/extern` rather than the packages (`sunpy/sunpy/extern/*.py`) themselves. 

As an example, updating [`sunpy/sunpy/extern/appdirs.py`](https://github.com/sunpy/sunpy/blob/main/sunpy/extern/appdirs.py) should download the latest *released* version of the upstream [ActiveState/appdirs `appdirs.py`](https://github.com/ActiveState/appdirs/blob/7af32e0b1fe57070ae8b5a717cdaebc094449518/appdirs.py). (Either using the GitHub API or `pip download appdirs` for example.) Then the script should replace the old `appdirs.py` file with the upstream file (and possibly run SunPy's `precommit` hooks to fix any PEP8 issues etc. not enforced upstream).

akash5100:
Oh! I misunderstood, thank you for the example @ConorMacBride.

akash5100:
I think this script will fail in 2 cases (maybe more):
1. If you've ever had to use the GitHub REST APIs, you may be familiar with [GitHub's API Rate Limits](https://docs.github.com/en/free-pro-team@latest/rest/overview/resources-in-the-rest-api#rate-limiting):
This means that if you call a GitHub API more than 5,000 times within a 60-minute window, even if you are authenticated, the GitHub APIs will return 403 FORBIDDEN.

2. If the packages are released without a version tag.

nabobalis:
> I think this script will fail in 2 cases (maybe more):
> 
> 1. If you've ever had to use the GitHub REST APIs, you may be familiar with [GitHub's API Rate Limits](https://docs.github.com/en/free-pro-team@latest/rest/overview/resources-in-the-rest-api#rate-limiting):
>    This means that if you call a GitHub API more than 5,000 times within a 60-minute window, even if you are authenticated, the GitHub APIs will return 403 FORBIDDEN.

I don't think we have to worry about that.

> 2. If the packages are released without a version tag.

This might be an issue. 
 
> edit: I don't know much about release of a package, but If `pip download <package>` always gives us the latest version, it would be the solution.

What does `pip download` do?

tools/update_extern.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    # if dst already exists, remove it
+    if os.path.exists(dst):
+        print(f"Removing {dst}")
+    # move the files
+    shutil.move(src, dst)
+    print(f"Moving {src} to {dst}")
+
+
+def extract_appdirs(destination):
```
nabobalis, line 86: Do we need one function per library?

tools/update_extern.py
```diff
@@ -0,0 +1,169 @@
...
+# List of packages to update
+PACKAGES = ["appdirs", "distro", "inflect", "parse"]
+
+# Dictionary to store the authors of the packages
+AUTHORS = {
+    "appdirs": "ActiveState",
+    "distro": "python-distro",
+    "inflect": "jaraco",
+    "parse": "r1chardj0n3s",
+}
```
nabobalis, line 26: Does this need to be seperate?

tools/update_extern.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+def get_latest_version(package):
+    """
+    Get the latest version of the package.
+    """
+    url = f"https://api.github.com/repos/{AUTHORS[package]}/{package}/releases/latest"
```
nabobalis, line 33: What about using tags instead?

tools/update_extern.py
akash5100: Earlier I thought the same thing. But the problem is, that the file which needs to be extracted from the source distribution is not the same for every package. 
For example, in `appdirs`, we need `appdirs/appdirs.py` but in `inflect`, we need `inflect/inflect/__init__.py`.
for `distro` we need `src/distro/distro.py`, for `parse` we need `/parse.py`. 
I thought that having a different function to extract the required `.py` for each package will be better than having a single complicated function.

If there is any better way we can work out, I will do it.

tools/update_extern.py
akash5100: I am not sure what you mean, but this function returns the latest `version tag` of the package.
```py
return response.json()["tag_name"]
```
For example, using this endpoint: https://api.github.com/repos/sunpy/sunpy/releases/latest we can get `"tag_name": "v4.0.0"`

nabobalis: Sounds good.

tools/update_extern.py
akash5100: I am not sure, but this makes accessing authors easy otherwise we have to hardcode the author's name.

ConorMacBride: Rather than having to write the package names twice, you can access them using the keys of the authors dictionary. My thinking is making sure it's easy to update this script when we add a new package to the extern directory.

akash5100: Sure, I will do that.

tools/update_extern.py
```diff
@@ -0,0 +1,169 @@
...
+
+# List of packages to update
+PACKAGES = ["appdirs", "distro", "inflect", "parse"]
+
+# Dictionary to store the authors of the packages
+AUTHORS = {
+    PACKAGES[0]: "ActiveState",
+    PACKAGES[1]: "python-distro",
+    PACKAGES[2]: "jaraco",
+    PACKAGES[3]: "r1chardj0n3s",
```
ConorMacBride, line 25: Having the keys as part of the `PACKAGES` list still means two updates if adding another package as you would have to update the `PACKAGES` list and add a `PACKAGES[4]` key to the `AUTHORS` dictionary. What I'm thinking is that, since we have single file packages maybe we could have a dictionary such as:

```py
PACKAGES = {
    "inflect": ["jaraco", "inflect/__init__.py"],
    ...
}
```
This tells us the package name, the author, and the file within the package that we want to copy to `sunpy/extern/<package name>.py`. If that works, it should mean that we could remove the package specific code from the script (i.e. remove `update_extern` and move that logic to `move_file`), and if we ever want to vendor a new package into extern it should be as simple as adding it to the `PACKAGES` dictionary and running this script.

tools/update_extern.py
```diff
@@ -0,0 +1,166 @@
...
+    "appdirs": "ActiveState",
+    "distro": "python-distro",
+    "inflect": "jaraco",
+    "parse": "r1chardj0n3s",
+}
+
+
+def package_exists():
+    """
+    Check if the package exists.
```
nabobalis, line 28: If it does, delete ti and redownload it.

tools/update_extern.py
akash5100: I hadn't read this comment before the push, but I tried to use a single dictionary.

tools/update_extern.py
```diff
@@ -0,0 +1,166 @@
...
+        # download the package
+        download_package(url, package)
+
+        # Open the zip file
+        unzip(f"extern_pkg/{package}.zip")
+
+    folders = list()
+    for root, dirs, files in os.walk("extern_pkg"):
+        folders = dirs
+        break
```
nabobalis, line 156: What does this do?

tools/update_extern.py
```diff
@@ -0,0 +1,166 @@
+#!/bin/env python
+
```
nabobalis, line 2:
```suggestion
```

tools/update_extern.py
```diff
@@ -0,0 +1,166 @@
...
+"""
+This script is used to update the files in `sunpy/extern` while making new releases.
```
nabobalis, line 4:
```suggestion
Updates all the libraries in ``sunpy/extern``
```

tools/update_extern.py
```diff
@@ -0,0 +1,166 @@
...
+"""
+
+import os
+import shutil
+from zipfile import ZipFile
+import requests
+from pathlib import Path
+
+# get parent directory of sunpy
```
nabobalis, line 13:
```suggestion
```

tools/update_extern.py
```diff
@@ -0,0 +1,166 @@
...
+SUNPY_DIR = Path(__file__).parent.parent
```
nabobalis, line 14: Why do we need both parents?

tools/update_extern.py
```diff
@@ -0,0 +1,166 @@
...
+    response = requests.get(url)
+    return response.json()["tag_name"]
+
+
+def get_download_url(author, package, version):
+    """
+    Get the download url for the package.
+    """
+    url = f"https://github.com/{author}/{package}/archive/refs/tags/{version}.zip"
+    return url
```
nabobalis, line 55: This seems like it can be one function or one block of code?

tools/update_extern.py
```diff
@@ -0,0 +1,166 @@
...
+        break
+
+    # Sort the packages in alphabetical order
+    folders.sort()
+
+    # Extract the files
+    for folder in folders:
+        update_extern(folder)
+
+    # Remove the temporary directory
```
nabobalis, line 165:
```suggestion
```

tools/update_extern.py
```diff
@@ -0,0 +1,166 @@
...
+
+    folders = list()
+    for root, dirs, files in os.walk("extern_pkg"):
+        folders = dirs
+        break
+
+    # Sort the packages in alphabetical order
+    folders.sort()
+
+    # Extract the files
```
nabobalis, line 161:
```suggestion
```

tools/update_extern.py
```diff
@@ -0,0 +1,166 @@
...
+
+        # Open the zip file
+        unzip(f"extern_pkg/{package}.zip")
+
+    folders = list()
+    for root, dirs, files in os.walk("extern_pkg"):
+        folders = dirs
+        break
+
+    # Sort the packages in alphabetical order
```
nabobalis, line 158:
```suggestion
```

tools/update_extern.py
```diff
@@ -0,0 +1,166 @@
...
+    # Check if the packages exist
+    if not package_exists():
+        print("Exiting...")
+        exit()
+
+    for package in PACKAGES:
+        # get the url of the package
+        url = get_download_url(PACKAGES[package], package, get_latest_version(PACKAGES[package], package))
+
+        # download the package
```
nabobalis, line 147:
```suggestion
```

tools/update_extern.py
```diff
@@ -0,0 +1,166 @@
...
+        download_package(url, package)
+
+        # Open the zip file
```
nabobalis, line 150:
```suggestion
```

tools/update_extern.py
```diff
@@ -0,0 +1,166 @@
...
+            # Move the file to the sunpy/extern directory
+            move_files(f"extern_pkg/{destination}/parse.py", SUNPY_DIR / "sunpy" / "extern" / "parse.py")
+
+    # Remove the directory
+    shutil.rmtree(f"extern_pkg/{destination}")
+
+
+if __name__ == "__main__":
+
+    # Check if the packages exist
```
nabobalis, line 138:
```suggestion
```

tools/update_extern.py
```diff
@@ -0,0 +1,166 @@
...
+    """
+    # if dst already exists, remove it
+    if os.path.exists(dst):
+        print(f"Removing {dst}")
+    # move the files
+    shutil.move(src, dst)
+    print(f"Moving {src} to {dst}")
+
+
+def update_extern(destination):
```
nabobalis, line 98: This function as it stands is too complex, we need to simplify it.

tools/update_extern.py
```diff
@@ -0,0 +1,166 @@
...
+    # Remove the directory
+    shutil.rmtree(f"extern_pkg/{destination}")
+
+
+if __name__ == "__main__":
+
+    # Check if the packages exist
+    if not package_exists():
+        print("Exiting...")
+        exit()
```
nabobalis, line 141: I don't see why we need to check this?

tools/update_extern.py
nabobalis: I am with Conor on this, that we we store the information we need in the dict, the author, the files we need to copy over and it should make the script simpler.

tools/update_extern.py
akash5100: Okay, I will explain what I tried to do.

1. we check that this [dict](https://github.com/sunpy/sunpy/pull/6127/files#diff-bb5c4cdf80ea20b4ae8f8db4b6d8d12e88e1d32f16f2b545ca551accf32f70d4R18) has correct keys and value. If the key-value pair are wrong, show error to the user and exit. 
 ```py
 # Check if the packages exist
    if not package_exists():
        print("Exiting...")
        exit()
```

2. we make a download URL using `get_download_url()` and use that URL to download `.zip` with `download_package()` then unzip it in a temporary directory 
```py
    for package in PACKAGES:
        # get the url of the package
        url = get_download_url(PACKAGES[package], package, get_latest_version(PACKAGES[package], package))

        # download the package
        download_package(url, package)

        # Open the zip file
        unzip(f"extern_pkg/{package}.zip")
```

3. we get names of all the `top-folder` inside that temporary directory in a `list`.
for example,
```
Temporary Directory
├──  appdirs-1.4.4
│  ├──  appdirs.py
│   ...
├──  distro-1.7.0
│  ├──  CHANGELOG.md
│   ...
├──  inflect-5.6.0
│  ├──  CHANGES.rst
│   ...
└──  parse-1.19.0
   ├──  LICENSE
   ├──  parse.py
    ...
```
In this dir, we store to names of packages in a list. 
So, the list `folders` will be 
`folders = ["appdirs-1.4.4","distro-1.7.0","inflect-5.6.0","parse-1.19.0"]`

```py
    folders = list()
    for root, dirs, files in os.walk("extern_pkg"):
        folders = dirs
        break
```
And we don't need to sort this. (This was something related to old code where I used PACKAGES and AUTHORS, but now it's a single dictionary, so I will remove this in the next commit.)
```py
    # Sort the packages in alphabetical order
    folders.sort()
```

Now we use that `list` to extract (move) the required python files, that will be done with `update_extern()` function.

```py
    # Extract the files
    for folder in folders:
        update_extern(folder)
```
4. Finally remove the temporary directory.
```py
    # Remove the temporary directory
    shutil.rmtree("extern_pkg")
```

ConorMacBride: For `1.` it might be less code to verify the HTTP response code of the request that's already in `get_latest_version` instead, although as @nabobalis pointed out we should merge some of those functions.

For `3.` it may simplify things if we extracted directly into the `sunpy/extern` directory. For example, using [`extract`](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.extract) instead of `extractall` you should be able to extract the single file only with `_______/sunpy/extern` as the path. Although you may need to delete the file if it exists, e.g.:

```py
def update_package(package_name, author, src):
    zip_filename = download_package(author, package_name)
    dest = SUNPY_EXTERN_DIR / f"{package_name}.py"
    if os.path.exists(dest):
        os.remove(dest)
    with ZipFile(zip_filename, "r") as zip_file:
        zip_file.extract(src, path=SUNPY_EXTERN_DIR)


for package_name, (author, src) in PACKAGES.items():
    update_package(package_name, author, src)
```

nabobalis: Yeah this is what we should do, to reduce the amount of code and the complexity of it.

tools/update_extern.py
akash5100: Yes, I will do that.

tools/update_extern.py
akash5100: Sorry for the late reply, I was busy with college assignments. 

If we try to unzip particular files that would be great, and I wrote it!

But what if the `.py` file we unzipped is named `__init__.py`? 
For example, In the case of `inflect` that would cause to replace the different file in `sunpy/extern`
So, can we extract it in the same temporary directory, rename and then move them?

akash5100: I doubt if it is possible to rename files during their extraction.

ConorMacBride: Yeah, good point! That should work, if you extract without the path you should then be able to move the file separately to the extern directory. Updated example from above:
```py
def update_package(package_name, author, src):
    zip_filename = download_package(author, package_name)
    with ZipFile(zip_filename, "r") as zip_file:
        ext = zip_file.extract(src)
    dest = SUNPY_EXTERN_DIR / f"{package_name}.py"
    if os.path.exists(dest):
        os.remove(dest)
    shutil.move(ext, dest)


for package_name, (author, src) in PACKAGES.items():
    update_package(package_name, author, src)
```

akash5100: > 
```py
def update_package(package_name, author, src):
    zip_filename = download_package(author, package_name)
    with ZipFile(zip_filename, "r") as zip_file:
        ext = zip_file.extract(src)
    dest = SUNPY_EXTERN_DIR / f"{package_name}.py"
    if os.path.exists(dest):
        os.remove(dest)
    shutil.move(ext, dest)
for package_name, (author, src) in PACKAGES.items():
    update_package(package_name, author, src)
```

Do I have to use this exact or this is just an example?

ConorMacBride: No, it's just an example. It would probably need some other things to make it work properly.

akash5100: Yes, for some reason I thought you want me to use this code.😆

nabobalis: I don't see why it wouldn't be simple enough to extract it and then rename it? We can also add that info to the DICT or a named tuple.

akash5100: Because for example, in case of inflect the file name is __init__.py and not inflect.py

akash5100: > Right. But we already have names for them in sunpy.

Yes, doing this will give us the same result, because `__init__.py` will be renamed as `inflect.py` & `distro.py`, `parse.py`, and `appdirs.py` will remain the same.

nabobalis: > > Right. But we already have names for them in sunpy.
> 
> Yes, doing this will give us the same result, because `__init__.py` will be renamed as `inflect.py` & `distro.py`, `parse.py`, and `appdirs.py` will remain the same.

Sure, is that a problem?

akash5100: > Sure, is that a problem?

No, I will just do these changes and then commit them.

ConorMacBride (commented):
I think we nearly have a script that'll automate the tedious process of updating the vendored code! I just made a couple of suggestions to make the script easier to update if we add new packages. Thanks @akash5100!

tools/update_extern.py
```diff
@@ -0,0 +1,116 @@
...
+
+SUNPY_DIR = Path(__file__).parent.parent
+
+
+PACKAGES = {
+    "appdirs": ["ActiveState", "appdirs.py"],
+    "distro": ["python-distro", "src/distro/distro.py"],
+    "inflect": ["jaraco", "inflect/__init__.py"],
+    "parse": ["r1chardj0n3s", "parse.py"],
+}
```
ConorMacBride, line 19: Hi @akash5100, sorry I have just realised in the original issue I forgot to mention that the `modest_image.py` file needs to be updated (as it was buried within the `glue` package). I tried including it in the list of packages although currently the script would have copied it to extern as `glue.py` rather than `modest_image.py`. 

I think the following convention for the dictionary of packages might make more sense going forward, i.e. keys are unique names for the files in the extern directory, and the values are lists of where to find the file: [user, repository, path]. That way it would be possible to update two separate files taken from the same GitHub repository.
```suggestion
PACKAGES = {
    "appdirs": ["ActiveState", "appdirs", "appdirs.py"],
    "distro": ["python-distro", "distro", "src/distro/distro.py"],
    "inflect": ["jaraco", "inflect", "inflect/__init__.py"],
    "modest_image": ["glue-viz", "glue", "glue/external/modest_image.py"],
    "parse": ["r1chardj0n3s", "parse", "parse.py"],
}
```

I do still think we should simplify the script by instead updating one file at a time. That way we wouldn't need to walk through directories to find the file we want to copy across, as we can keep track of everything inside the function based on the return values (see how I got the path to the extracted file in https://github.com/sunpy/sunpy/pull/6127#discussion_r872409761). What I'm thinking is a function:

```py
def download_github_file(user, repo, src, dest):
```

where `user` and `repo` are the GitHub username and repository where the file is present, `src` is the path to the file within the GitHub repository, and `dest` is where to download the file to (e.g. `/Users/username/sunpyrepo/sunpy/extern/name.py`). This function could then be called in a `for` loop for each entry in the `PACKAGES` dictionary as shown in https://github.com/sunpy/sunpy/pull/6127#discussion_r872409761. 

Sorry for the last minute change, although I think this is a great opportunity to create a function that can grab a file from a GitHub repository and place it in any directory you like!! Therefore the function itself could be used by other projects and not just SunPy! 😀

nabobalis: I actually think in this case, we can remove `modest_image.py`, I can't find where it is imported anymore. I suspect it was needed when we had the animation source code in sunpy but now we dont.

tools/update_extern.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def download_package(package: str):
+    """
+    Download the latest version of package using Github release tags.
+
+    Args:
+        package: The name of the package to download.
+    """
+    for package in PACKAGES:
```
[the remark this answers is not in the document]
akash5100: Oh yes, That's a mistake by me.

tools/update_extern.py
Cadair: I think modest image was only used by the animators and it's now in `mpl-animators` and I forgot to remove it from here.

akash5100:
Sorry for working on this late, I was on a vacation; I think the script is almost done.

The only thing left is, renaming the extracted file to `PACKAGES[package]` and I will do it asap.

tools/update_extern.py
```diff
@@ -0,0 +1,119 @@
+"""
+Updates all the libraries in ``sunpy/extern``
+"""
+
+import os
+import shutil
+from zipfile import ZipFile
+import requests
```
nabobalis, line 8: Can we not just use Parfive (the sunpy downloader instead of having to use requests and tqdm?

tools/update_extern.py
```diff
@@ -0,0 +1,119 @@
...
+from tqdm import tqdm
+
+SUNPY_DIR = Path(__file__).parent.parent
+
+# "file_name": ["user", "repository", "path_to_file"]
+PACKAGES = {
+    "appdirs": ["ActiveState", "appdirs", "appdirs.py"],
+    "distro": ["python-distro", "distro", "src/distro/distro.py"],
+    "inflect": ["jaraco", "inflect", "inflect/__init__.py"],
+    "modest_image": ["glue-viz", "glue", "glue/external/modest_image.py"],
```
nabobalis, line 19:
```suggestion
```
We don't need this.

tools/update_extern.py
```diff
@@ -0,0 +1,119 @@
...
+import os
+import shutil
+from zipfile import ZipFile
+import requests
+from pathlib import Path
+from tqdm import tqdm
+
+SUNPY_DIR = Path(__file__).parent.parent
+
+# "file_name": ["user", "repository", "path_to_file"]
```
nabobalis, line 14:
```suggestion
# "package_name": ["user", "repository", "path_to_file"]
```
?

nabobalis:
> Sorry for working on this late, I was on a vacation; I think the script is almost done.

It is fine, real life is more important!

tools/update_extern.py
nabobalis: You can see how to use it here: https://github.com/Cadair/parfive

ConorMacBride: Yeah, `requests` is a third party packages that SunPy does not directly depend on, so ideally it would be better to use `parfive` to download the zip file and [`urllib.request`](https://docs.python.org/3/library/urllib.request.html#module-urllib.request) from the Python Standard Library for getting the HTTP response to parse as JSON.

tools/update_extern.py
```diff
@@ -0,0 +1,111 @@
...
+    Download the latest version of package using Github release tags.
+
+    Args:
+        package: The name of the package to download.
+    """
+    print(f"Checking {user}/{repo}")
+    response = requests.get(f"https://api.github.com/repos/{user}/{repo}")
+    if response.status_code != 200:
+        print(f"{user}/{repo} does not exist.")
+        exit()
```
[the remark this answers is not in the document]
nabobalis: Is this something we should hit anyway?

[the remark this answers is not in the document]
ConorMacBride: Probably not. If the package doesn't exist the download will fail anyway.

akash5100:
pre-commit.ci autofix

ConorMacBride (approved):
Thanks @akash5100! I tried this script and it works brilliantly! 🎉 So much faster than downloading each file manually and copying across.

tools/update_extern.py
```diff
@@ -0,0 +1,84 @@
...
+    if response.status != 200:
+        raise ValueError(f"{user}/{repo} does not exist.")
+
+    url = f"https://api.github.com/repos/{user}/{repo}/releases/latest"
+    response = urllib.request.urlopen(url)
+    if response.status != 200:
+        url = f"https://api.github.com/repos/ActiveState/appdirs/tags"
+        response = urllib.request.urlopen(url)
+        response = json.load(response)
+        version = response[0]["name"]
```
nabobalis, line 43: Why is this here?

tools/update_extern.py
```diff
@@ -0,0 +1,84 @@
...
+        version = response["tag_name"]
+    url = f"http://github.com/{user}/{repo}/archive/refs/tags/{version}.zip"
+
+    if not os.path.exists("extern_pkg"):
+        os.mkdir("extern_pkg")
+
+    dl = Downloader()
+    dl.enqueue_file(url, path="extern_pkg", filename=f"{repo}.zip")
+    files = dl.download()
+    return f"extern_pkg/{repo}.zip"
```
nabobalis, line 55: Can't you return files here?

tools/update_extern.py
```diff
@@ -0,0 +1,84 @@
...
+    dest = Path(dest)
+    if dest.exists() and dest.is_file():
+        os.remove(dest)
+    shutil.move(ext, dest)
+    os.remove(zip_file)
+    shutil.rmtree("extern_pkg")
+
+
+if __name__ == "__main__":
+
```
nabobalis, line 81:
```suggestion
```

tools/update_extern.py
```diff
@@ -0,0 +1,84 @@
...
+def download_github_file(user: str, repo: str, src: Path, dest: Path):
+    """
+    Download a file from Github.
+    Args:
+        user: Github user name
+        repo: Github repository name
+        src: Path to file in the package
+        dest: Path where the file should be extracted
+    """
+
```
nabobalis, line 67:
```suggestion
```

tools/update_extern.py
```diff
@@ -0,0 +1,84 @@
...
+
+def download_github_file(user: str, repo: str, src: Path, dest: Path):
+    """
+    Download a file from Github.
+    Args:
+        user: Github user name
+        repo: Github repository name
+        src: Path to file in the package
+        dest: Path where the file should be extracted
```
nabobalis, line 65:
```suggestion
```

tools/update_extern.py
```diff
@@ -0,0 +1,84 @@
...
+}
+
+
+def download_package(user: str, repo: str):
+    """
+    Download the latest version of package using Github release tags.
+
+    Args:
+        user: Github user name
+        repo: Github repository name
```
nabobalis, line 30:
```suggestion
```

nabobalis:
I think we also should copy over the latest licence file for each package, hopefully its stored in the dist.

tools/update_extern.py
ConorMacBride: Oh, I missed that

tools/update_extern.py
```diff
@@ -0,0 +1,84 @@
...
+        response = urllib.request.urlopen(url)
+        response = json.load(response)
+        version = response[0]["name"]
+    else:
+        response = json.load(response)
+        version = response["tag_name"]
+    url = f"http://github.com/{user}/{repo}/archive/refs/tags/{version}.zip"
+
+    if not os.path.exists("extern_pkg"):
+        os.mkdir("extern_pkg")
```
nabobalis, line 50: Should this not be in the main part at the bottom?

tools/update_extern.py
```diff
@@ -0,0 +1,84 @@
...
+        dest: Path where the file should be extracted
+    """
+
+    zip_file = download_package(user, repo)
+    with ZipFile(zip_file, "r") as f:
+        folder = Path(f.namelist()[0]).parts[0]
+        ext = f.extract(f"{folder}/{src}", "extern_pkg")
+    dest = Path(dest)
+    if dest.exists() and dest.is_file():
+        os.remove(dest)
```
nabobalis, line 74: Do we need to remove it? Does `shutil.move` not just override it?

tools/update_extern.py
```diff
@@ -0,0 +1,84 @@
...
+    shutil.move(ext, dest)
+    os.remove(zip_file)
+    shutil.rmtree("extern_pkg")
+
+
+if __name__ == "__main__":
+
+    for package, (user, repo, src) in PACKAGES.items():
+        dest = SUNPY_EXTERN_DIR / f"{package}.py"
+        download_github_file(user, repo, src, dest)
```
nabobalis, line 84:
```suggestion
        download_github_file(user, repo, src, SUNPY_EXTERN_DIR / f"{package}.py"
)
```

tools/update_extern.py
ConorMacBride: I think they all have releases so the tags code isn't run. But we should probably keep it anyway.

nabobalis: But this is only for one package? It isn't checking the tags for each project that I can see.

Maybe we should fallback to tags if no release is found for each project?

tools/update_extern.py
ConorMacBride: It raises an error if run on Windows I think.

tools/update_extern.py
```diff
@@ -0,0 +1,84 @@
...
+        response = json.load(response)
+        version = response[0]["name"]
+    else:
+        response = json.load(response)
+        version = response["tag_name"]
+    url = f"http://github.com/{user}/{repo}/archive/refs/tags/{version}.zip"
+
+    if not os.path.exists("extern_pkg"):
+        os.mkdir("extern_pkg")
+
```
ConorMacBride, line 51: Just so we can double check we are not downloading a prerelease version:
```suggestion

    print(f"Downloading {user}/{repo}:refs/tags/{version}")
```

tools/update_extern.py
ConorMacBride: It is doing that, just `f"https://api.github.com/repos/ActiveState/appdirs/tags"` should be `f"https://api.github.com/repos/{user}/{repo}/tags"`. I tried `raise ValueError()` in that code branch and the error wasn't raised so it found a release for all.

tools/update_extern.py
ConorMacBride: Should probably use a `tempfile.mkdtemp` instead inside the `download_package` function and then inside `download_github_file` do `ext = f.extract(f"{folder}/{src}", zip_file.parent)`.

akash5100:
> I think we also should copy over the latest license file for each package, which should be stored in the dist.

If we have a list of files that we need to extract, we can do this in the same function, but I need to figure out how we going to rename stuff.

> If we have a list of files that we need to extract

Something like this:
```py
PACKAGES = {
    "appdirs": ["ActiveState", "appdirs", ["appdirs.py", "LICENSE.txt" ],
}
```

Otherwise, without any change in the current function, we can make a new dictionary of LICENSE and call it once again? (But this will repeat downloading, extracting, deleting)

nabobalis:
We just rename it to `<app_name>_licence`?

tools/update_extern.py
nabobalis: If we need to fall back to tags, would using the tags not be a better place to start?

tools/update_extern.py
nabobalis: Fair enough.

akash5100:
> We just rename it to `<app_name>_licence`?

Adding this does the work without changing the function, does this looks good?
```py
LICENSES = {
    "appdirs_license.txt": ["ActiveState", "appdirs", "LICENSE.txt"],
    "distro_license.rst": ["python-distro", "distro", "LICENSE"],
    "inflect_license.txt": ["jaraco", "inflect", "LICENSE"],
    "parse_license.txt": ["r1chardj0n3s", "parse", "LICENSE"],
}

for package, (user, repo, src) in LICENSES.items():
    download_github_file(user, repo, src, SUNPY_EXTERN_DIR / package)
```

ConorMacBride:
> We just rename it to `<app_name>_licence`?

Might need to update the dictionary keys to be `appdirs.py` etc, and then add the licences as new "packages" `appdirs_licence.txt` etc to the dictionary.

akash5100:
Actually, we can add `.py` in the keys instead of formatting the string while calling the function, still works the same.
```py 
PACKAGES = {
    "appdirs.py": ["ActiveState", "appdirs", "appdirs.py"],
    "distro.py": ["python-distro", "distro", "src/distro/distro.py"],
    "inflect.py": ["jaraco", "inflect", "inflect/__init__.py"],
    "parse.py": ["r1chardj0n3s", "parse", "parse.py"],
}
```
and then
```py
    for package, (user, repo, src) in PACKAGES.items():
        download_github_file(user, repo, src, SUNPY_EXTERN_DIR / package)
```

If we have the `LICENSES` dict, we can just replace `PACKAGES.items()` to `LICENSES.item()`

ConorMacBride:
Oh! We don't have to download the whole repo! We should be able to just download the single file: `https://raw.githubusercontent.com/{user}/{repo}/{version}/{src}` (e.g. https://raw.githubusercontent.com/ActiveState/appdirs/1.4.4/appdirs.py)

Then we can keep `download_github_file` as a function which only downloads one file.

akash5100:
Nice, no more zips

ConorMacBride:
In GitHub if you navigate to a file with a particular tag selected e.g. https://github.com/sunpy/sunpy/blob/v4.0.0/setup.py you can see this URL by clicking the "Raw" button. I think we should be able to rely on this? I have seen this type of URL used directly before. What do you think @nabobalis?

nabobalis:
> In GitHub if you navigate to a file with a particular tag selected e.g. https://github.com/sunpy/sunpy/blob/v4.0.0/setup.py you can see this URL by clicking the "Raw" button. I think we should be able to rely on this? I have seen this type of URL used directly before. What do you think @nabobalis?

Sounds good

akash5100:
> Then we can keep `download_github_file` as a function which only downloads one file.

We get bytes as a response, Should I write the byte in the required file?
for example: write the response directly in `sunpy/extern/appdirs.py`

nabobalis:
> > Then we can keep `download_github_file` as a function which only downloads one file.
> 
> We get bytes as a response, Should I write the byte in the required file? for example: write the response directly in `sunpy/extern/appdirs.py`

Sure.

tools/update_extern.py
```diff
@@ -0,0 +1,55 @@
...
+    "parse_license.txt": ["r1chardj0n3s", "parse", "LICENSE"],
+}
+
+
+def download_github_file(user: str, repo: str, src: Path, dest: Path):
+    """
+    Download a file from Github.
+    """
+    print(f"Checking {user}/{repo}")
+    response = urllib.request.urlopen(f"https://api.github.com/repos/{user}/{repo}")
```
nabobalis, line 32: So we need this block of code?

Cadair (approved):
Seems sane to me. Thanks!

PR #6133, opened 2022-05-06
Use new js action for matrix notifications

Cadair:
This simplifies the yaml and also sends reactions which tell you what jobs failed.

I am currently testing this out on ndcube, so let's see if that passes and then merge here.


nabobalis (approved):

.github/workflows/ci.yml
```diff
@@ -111,26 +111,14 @@ jobs:
     secrets:
       pypi_token: ${{ secrets.pypi_token }}
 
-  notify_pass:
+  notify:
     if: github.event_name == 'workflow_dispatch'
```
ConorMacBride, line 115: Needs `always()` otherwise it would only message on success:
```suggestion
    if: always() && github.event_name == 'workflow_dispatch'
```

nabobalis: So actions will not run this without this?

ConorMacBride: We want the `notify` job to run even if its dependants (`[publish, online, cron]`) fail. See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-not-requiring-successful-dependent-jobs

Cadair: Ah yes, I fixed this in ndcube but forgot to put it here as well.

ConorMacBride (approved):

PR #6136, opened 2022-05-09
theme broke the conf

nabobalis:
Fix for the doc build.


nabobalis:
This will be broken on the 3.0/3.1 builds, but if we dont really plan to do a release, it wont matter.

Cadair:
https://github.com/sunpy/sunpy/runs/6356705018?check_suite_focus=true#step:8:761 balls

Cadair:
We should adjust the theme test build to build in parallel

nabobalis:
For now, we pulled the theme version causing a problem and will do this again in future.

PR #6137, opened 2022-05-10
Add option to run cron and publish stages in PRs

### LINKED ISSUE #6130: Add label for wheel building on PRs
### Describe the feature

There are some PRs (e.g. https://github.com/sunpy/sunpy/pull/6129) where it is helpful to build wheels to check that the code change fixes or doesn't break wheel building. Matplotlib recently implemented the option to build wheels by adding a label to an issue: https://github.com/sunpy/sunpy/pull/6129 - it would be nice to replicate that here.

### Proposed solution

_No response_

ConorMacBride:
This updates the CI workflow to also run the `cron` stage if a `Run cron` tag is added to (or already present in) the PR. And similarly run the `publish` stage on `Run publish` tag.

As there doesn't seem to be a way to filter triggers based on specific tags, I've added a filtering conditional to the `cron` stage to skip if a different tag was added.

Can you think of better names for the tags?

Closes #6130

### Notes:
These are the defaults, so had to state them to also include `labeled`:
```
      - opened
      - synchronize
      - reopened
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,11 +11,6 @@ on:
       - '!*pre*'
       - '!*post*'
   pull_request:
-    types:
-      - opened
-      - synchronize
-      - reopened
-      - labeled
   # Allow manual runs through the web UI
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


nabobalis (approved):

nabobalis:
Names look fine to me.

Cadair:
I would maybe use `Run cron CI` or something as in isolation a `Run cron` label is a little weird.

Looks good though (it does annoy me that you end up with the skipped checks in the PR).

nabobalis:
Is the reason I can't merge this PR related to that?

ConorMacBride:
Yeah, the skipped tests are not good. I think we should remove the `labeled` event listener and only run the additional checks on pushes after labelling. For most cases this should be sufficient as we can just remember to add the label while creating the PR. (Alternatively, closing and reopening the PR would trigger the extra CI.) That way we aren't running (and skipping) the CI workflow every time we update labels.

nabobalis:
Sounds good.

nabobalis:
Manually backported.

PR #6153, opened 2022-05-16
Adds support for saving maps as jp2 files

dgarciabriseno:
## PR Description

Uses glymur and numpy to save images as jp2 files when executing `mapfile.save("outfile.jp2")`

For some time this has simply raised a "Not implemented" exception.

### DIFF
--- a/changelog/6153.feature.rst
+++ b/changelog/6153.feature.rst
@@ -1,0 +1,1 @@
+Add support for saving maps as jp2 files.
--- a/changelog/6153.feature.rst
+++ b/changelog/6153.feature.rst
@@ -1,1 +1,1 @@
-Add support for saving maps as jp2 files.
+Add support for saving `~sunpy.map.GenericMap` as JPEG 2000 files.
--- a/sunpy/io/jp2.py
+++ b/sunpy/io/jp2.py
@@ -169,7 +169,7 @@ def write(fname, data, header, **kwargs):
     jp2_data = np.uint8(data)
     jp2 = Jp2k(tmpname, jp2_data, **kwargs)
 
-    # Append the XML data to the headerp information
+    # Append the XML data to the header information
     meta_boxes = jp2.box
     target_index = len(meta_boxes) - 1
     fits_box = generate_jp2_xmlbox(header)
--- a/sunpy/io/jp2.py
+++ b/sunpy/io/jp2.py
@@ -1,6 +1,7 @@
 """
 This module provides a JPEG 2000 file reader.
 """
+import os
 from xml.etree import cElementTree as ET
 
 from sunpy.io.header import FileHeader
@@ -71,6 +72,101 @@ def get_header(filepath):
 
     return [FileHeader(pydict)]
 
+def header_to_xml(header):
+    """
+    Converts image header metadata into an XML Tree that can be inserted into
+    a JP2 file header.
+
+    Parameters
+    ----------
+    header : `MetaDict`
+        A header dictionary.
+    """
+    # glymur uses lxml and will crash if trying to use
+    # python's builtin xml.etree
+    import lxml.etree as ET
+
+    fits = ET.Element("fits")
+
+    already_added = set()
+    for key in header:
+        # Some headers span multiple lines and get duplicated as keys
+        # header.get will appropriately return all data, so if we see
+        # a key again, we can assume it was already added to the xml tree.
+        if (key in already_added):
+            continue
+
+        # Add to the set so we don't duplicate entries
+        already_added.add(key)
+
+        el = ET.SubElement(fits, key)
+        data = header.get(key)
+        if type(data) == bool:
+            data = "1" if data else "0"
+        else:
+            data = str(data)
+
+        el.text = data
+
+    return fits
+
+def generate_jp2_xmlbox(header):
+    """
+    Generates the JP2 XML box to be inserted into the jp2 file.
+
+    Parameters
+    ----------
+    header : `MetaDict`
+        A header dictionary.
+    """
+    # glymur uses lxml and will crash if trying to use
+    # python's builtin xml.etree
+    import lxml.etree as ET
+    from glymur import jp2box
+
+    header_xml = header_to_xml(header)
+    meta = ET.Element("meta")
+    meta.append(header_xml)
+    tree = ET.ElementTree(meta)
+    return jp2box.XMLBox(xml=tree)
+
+def get_tmp_jp2_file_name(filename):
+    """
+    Returns a temporary file name for jp2 files since creating
+    a JP2 file with the correct header takes 2 passes.
+    """
+    return filename + ".tmp.jp2"
 
 def write(fname, data, header):
-    raise NotImplementedError("No jp2 writer is implemented.")
+    """
+    Take a data header pair and write a JP2 file.
+
+    Parameters
+    ----------
+    fname : `str`
+        File name, with extension.
+    data : `numpy.ndarray`
+        n-dimensional data array.
+    header : `dict`
+        A header dictionary.
+    """
+    import numpy as np
+    from glymur import Jp2k
+
+    # Create an initial jp2 file with the given data
+    tmpname = get_tmp_jp2_file_name(fname)
+    jp2_data = np.uint8(data)
+    jp2 = Jp2k(tmpname, jp2_data)
+
+    # Append the XML data to the headerp information
+    meta_boxes = jp2.box
+    target_index = len(meta_boxes) - 1
+    fits_box = generate_jp2_xmlbox(header)
+    meta_boxes.insert(target_index, fits_box)
+
+    # Rewrite the jp2 file with the xml data in the header
+    jp2.wrap(fname, boxes=meta_boxes)
+
+    # Remove the initial temporary file
+    os.remove(tmpname)
+
--- a/sunpy/io/jp2.py
+++ b/sunpy/io/jp2.py
@@ -164,8 +164,8 @@ def write(fname, data, header, **kwargs):
 
     Notes
     -----
-    Saving as a jpeg 2000 will modify the provided
-    data array to support the jpeg 2000 format.
+    Saving as a JPEG2000 will cast the data array to
+    uint8 values to support the JPEG2000 format.
     """
     from glymur import Jp2k
 
--- a/sunpy/io/jp2.py
+++ b/sunpy/io/jp2.py
@@ -147,7 +147,7 @@ def generate_jp2_xmlbox(header):
     return jp2box.XMLBox(xml=tree)
 
 
-def write(fname, data, header):
+def write(fname, data, header, **kwargs):
     """
     Take a data header pair and write a JP2 file.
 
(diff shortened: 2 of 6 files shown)


sunpy/io/tests/test_jp2.py
```diff
@@ -1,12 +1,14 @@
 import numpy as np
+import sunpy.io.jp2 as jp2
```
nabobalis, line 2: Below we do` from sunpy.io.jp2 import get_header, read`, these imports should be unified.

Either adding write there or changing all the others to be `jp2.read` etc

sunpy/io/tests/test_jp2.py
```diff
@@ -30,3 +32,10 @@ def test_read_memmap():
     # Keyword is not passed in the function call
     data, _ = read(AIA_193_JP2, memmap=False)[0]
     assert data.base is not None
+
+def test_simple_write(tmpdir):
+    data, header = _fits.read(TEST_AIA_IMAGE)[0]
+    outfile = tmpdir / "test.jp2"
+    jp2.write(str(outfile), data, header)
+    assert outfile.exists()
+
```
nabobalis, line 41: Would a data or header == the fits one (after conversion) be possible?

sunpy/io/tests/test_jp2.py
dgarciabriseno: I'll change others to jp2.get_header and jp2.read

read/write alone are too ambiguous so I added the extra import.

sunpy/io/jp2.py
```diff
@@ -71,6 +72,101 @@ def get_header(filepath):
...
+def header_to_xml(header):
+    """
+    Converts image header metadata into an XML Tree that can be inserted into
+    a JP2 file header.
+
+    Parameters
+    ----------
+    header : `MetaDict`
+        A header dictionary.
+    """
```
nabobalis, line 84: This will need a return section too.

sunpy/io/jp2.py
```diff
@@ -71,6 +72,101 @@ def get_header(filepath):
...
+    # glymur uses lxml and will crash if trying to use
+    # python's builtin xml.etree
+    import lxml.etree as ET
```
nabobalis, line 87: Will `lxml` be installed on a users PC? I forget if we pull it in from a dependency.

Cadair (commented):
Thanks for the PR @dgarciabriseno it looks good :smile:

sunpy/io/jp2.py
Cadair: iirc Glymur has it as a dep, but if we are importing it directly perhaps we should add it to our deps as well?

sunpy/io/jp2.py
```diff
@@ -71,6 +72,101 @@ def get_header(filepath):
...
+    from glymur import jp2box
+
+    header_xml = header_to_xml(header)
+    meta = ET.Element("meta")
+    meta.append(header_xml)
+    tree = ET.ElementTree(meta)
+    return jp2box.XMLBox(xml=tree)
+
+def get_tmp_jp2_file_name(filename):
```
nabobalis, line 133: Does this need to be a function?

sunpy/io/tests/test_jp2.py
dgarciabriseno: not for data, the uint8 conversion for glymur changes what the data array looks like.

For header it seems like it would be possible, but not as trivial as it sounds because the "read" functions seem to read differently, which may be an error on its own. It seems like _fits.read creates a header out of strings and jp2.read creates a header out of real values. I don't know where the casts happen between reading/writing.

I tried to compare the headers and found that
- _fits.read header will put in `'nan'` but jp2.read will use `nan` (str vs. real nan)
- _fits.read header has `'001.197'` and jp2.read gives `1.197`

```python
FAILED test_jp2.py::test_simple_write - AssertionError: assert '001.197' == 1.197
```

sunpy/io/jp2.py
dgarciabriseno: Yes because it's a dependency of glymur.
Should I also add it as a dependency for sunpy?

sunpy/io/jp2.py
dgarciabriseno: Does anything need to be a function?
I can remove it, you're right it doesn't really add much. Came from writing the main code first and coming back to this. In my mind I was going to save it to the /tmp directory but didn't want to write that yet so I created the function, but then I ended up not doing that.

sunpy/io/jp2.py
nabobalis: We should, do we have an io group to add it to in setup.cfg?

sunpy/io/tests/test_jp2.py
dgarciabriseno: Added a read back and sanity check on the header, let me know if you think that's sufficient

changelog/6153.feature.rst
```diff
@@ -0,0 +1 @@
+Add support for saving maps as jp2 files.
```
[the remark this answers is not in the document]
dgarciabriseno: Updated, thanks

sunpy/io/jp2.py
dgarciabriseno: Added the return sections

sunpy/io/jp2.py
```diff
@@ -71,6 +72,101 @@ def get_header(filepath):
...
+    ----------
+    fname : `str`
+        File name, with extension.
+    data : `numpy.ndarray`
+        n-dimensional data array.
+    header : `dict`
+        A header dictionary.
+    """
+    import numpy as np
+    from glymur import Jp2k
```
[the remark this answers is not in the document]
dgarciabriseno: I moved numpy to the top but left glymur in the function. It's not the fastest import which is probably why it's also imported inside the read function already instead of the top of the file.

sunpy/io/jp2.py
```diff
@@ -71,6 +72,101 @@ def get_header(filepath):
 
     return [FileHeader(pydict)]
 
+def header_to_xml(header):
```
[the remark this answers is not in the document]
nabobalis: It might be easier to install pre-commit and run it locally. It has failed with a bunch of little style checks and doing it all automatically will be easier.

sunpy/io/jp2.py
```diff
@@ -72,5 +74,107 @@ def get_header(filepath):
...
+    from glymur import jp2box
+
+    header_xml = header_to_xml(header)
+    meta = ET.Element("meta")
+    meta.append(header_xml)
+    tree = ET.ElementTree(meta)
+    return jp2box.XMLBox(xml=tree)
+
+
 def write(fname, data, header):
```
nabobalis, line 149: Can you check if other writes have a ``**kwargs``? If so, we should try to follow that for the time being.

sunpy/io/jp2.py
```diff
@@ -72,5 +74,107 @@ def get_header(filepath):
...
+    fname : `str`
+        File name, with extension.
+    data : `numpy.ndarray`
+        n-dimensional data array.
+    header : `dict`
+        A header dictionary.
+    """
+    from glymur import Jp2k
+
+    # Create an initial jp2 file with the given data
```
nabobalis, line 164:
```suggestion
```

sunpy/io/jp2.py
```diff
@@ -72,5 +74,107 @@ def get_header(filepath):
...
+    tmpname = fname + "tmp.jp2"
+    jp2_data = np.uint8(data)
+    jp2 = Jp2k(tmpname, jp2_data)
+
+    # Append the XML data to the headerp information
```
nabobalis, line 169:
```suggestion
```

sunpy/io/jp2.py
```diff
@@ -72,5 +74,107 @@ def get_header(filepath):
...
+    meta_boxes = jp2.box
+    target_index = len(meta_boxes) - 1
+    fits_box = generate_jp2_xmlbox(header)
+    meta_boxes.insert(target_index, fits_box)
+
+    # Rewrite the jp2 file with the xml data in the header
```
nabobalis, line 175:
```suggestion
```

sunpy/io/jp2.py
```diff
@@ -72,5 +74,107 @@ def get_header(filepath):
...
+    jp2.wrap(fname, boxes=meta_boxes)
+
+    # Remove the initial temporary file
```
nabobalis, line 178:
```suggestion
```

sunpy/io/tests/test_jp2.py
```diff
@@ -1,32 +1,45 @@
 import numpy as np
+import sunpy.io.jp2 as jp2
```
nabobalis, line 2:
```suggestion
from sunpy.io import jp2
```

sunpy/io/tests/test_jp2.py
nabobalis: It works for me!

nabobalis:
The failing test is:

`sunpy/util/tests/test_sysinfo.py::test_find_dependencies`

You will need to add the new depciecny (lxml) to the sorted list at last assert statement.

sunpy/io/jp2.py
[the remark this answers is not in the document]
dgarciabriseno: Ok

sunpy/io/jp2.py
dgarciabriseno: Why do you want to remove the comments?

nabobalis: I don't think they are useful.

dgarciabriseno: They describe the process that's being done rather than trying to figure out what's going on inside glymur.
Like the fact that Jp2k and jp2.wrap actually write the file.
And that the these arbitrary "boxes" are part of the file's header.

nabobalis: Then the middle two comments could stay?

dgarciabriseno: Sure

dgarciabriseno: Updated

nabobalis:
Thanks for the PR and fixing all our nitpicking comments!

nabobalis (approved):

dgarciabriseno:
No problem, anything else need to be updated for the last failing checks?

Cadair:
Yeah, we need to do something about this test fail: https://github.com/sunpy/sunpy/runs/6457750672?check_suite_focus=true#step:8:5138

It's under windows, where we don't have openJPEG installed on the CI worker, so I think the quickest fix is to skip that test under windows. To do that you can use the `@sunpy.tests.helpers.skip_windows` decorator.

Cadair:
If we want to test openjpeg on windows we would have to build a [Chocolatey](https://docs.chocolatey.org/en-us/) package for it.

sunpy/io/tests/test_jp2.py
Cadair: Have we documented the fact that the data is modified by saving it to jp2?

Cadair (changes requested):
I think we need to put a note somewhere in the docs (where people will actually see it, so `map.save` as well as `jp2.write`) that the data isn't preserved when saving to jp2.

Otherwise I think this looks good.

sunpy/io/tests/test_jp2.py
nabobalis: It is done now.

nabobalis:
Thanks @dgarciabriseno

nabobalis:
Thanks for the PR @dgarciabriseno and dealing with our many small comments!

dgarciabriseno:
Thanks! I'll be using this feature later on in helioviewer.

PR #6162, opened 2022-05-18
Split docs into gallery and non-gallery builds

ConorMacBride:
So we can more easily distinguish between a tox doc build failing due to remote data or formatting, this PR separates the two steps that ran inside the `build_docs_ci` tox environment into two separate jobs. Caching is used between the jobs within the same GitHub Actions run.

- [ ] Delete commit where I point to my fork of the workflow (see https://github.com/OpenAstronomy/github-actions-workflows/pull/46)


tox.ini
```diff
@@ -88,27 +88,18 @@ commands =
     figure: python -c "import sunpy.tests.helpers as h; print(h.get_hash_library_name())"
     figure: {env:PYTEST_COMMAND} -m "mpl_image_compare" --mpl --remote-data=any --mpl-generate-summary=html --mpl-baseline-path=https://raw.githubusercontent.com/sunpy/sunpy-figure-tests/sunpy-master/figures/{envname}/ {posargs}
 
-[testenv:build_docs]
+[testenv:build_docs-{,nogallery}]
```
Cadair, line 91: can I suggest flipping this logic so it's `-gallery` to opt-in rather than opt out?

ConorMacBride: Sure. I just did it like that for backwards compatibility.

Cadair:
it looks like the `-nogallery` ran the gallery anyway?! https://github.com/sunpy/sunpy/runs/6492479475?check_suite_focus=true#step:9:155

Cadair:
and the online build_docs is running the tests :eyes:

ConorMacBride:
> and the online build_docs is running the tests 👀

I did `build_docs-{,nogallery}` instead of `build_docs{,-nogallery}`. Will fix.

ConorMacBride:
> it looks like the `-nogallery` ran the gallery anyway?! https://github.com/sunpy/sunpy/runs/6492479475?check_suite_focus=true#step:9:155

This is expected as it parses without executing https://sphinx-gallery.github.io/stable/configuration.html?highlight=plot_gallery#building-without-executing-examples

Cadair:
> > it looks like the `-nogallery` ran the gallery anyway?! https://github.com/sunpy/sunpy/runs/6492479475?check_suite_focus=true#step:9:155
> 
> This is expected as it parses without executing https://sphinx-gallery.github.io/stable/configuration.html?highlight=plot_gallery#building-without-executing-examples

I think normally it prints something which says it's skipping running them?

Cadair:
It seems to have worked this time :partying_face:

That's really cool :)

ConorMacBride:
Should we be caching `docs/generated` also?

Cadair:
The two builds are roughly the same total execution time as the last combined doc build on main, so I don't think we are losing anything of significance here.

nabobalis:
Should we update the dev docs on this now?

tox.ini
```diff
@@ -27,6 +27,7 @@ description =
 setenv =
     MPLBACKEND = agg
     SUNPY_SAMPLEDIR = {env:SUNPY_SAMPLEDIR:{toxinidir}/.tox/{envname}/sample_data/}
+    ci: SUNPY_SAMPLEDIR = {toxinidir}/docs/sample_data/
```
nabobalis, line 30: That is where the sample data gets downloaded to in the CI?!

nabobalis (approved):

ConorMacBride:
Don't merge yet. I need to fix the build_docs-gallery-ci environment. It's running pytest instead of sphinx

tox.ini
Cadair: Why can't we just cache the default directory? or set this env var inside the github actions workflow?

ConorMacBride: In tox the default (from the previous line) is e.g. `.tox/build_docs/sample_data/` or `.tox/build_docs-gallery/sample_data/` so we can't cache and restore easily as they are different directories. So, yes, *in the docs builds only* I force it to be `docs/sample_data/` instead. Unless there would be a better common directory?

We can't pass env vars from a workflow to a reusable workflow. https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations

nabobalis:
> Don't merge yet. I need to fix the build_docs-gallery-ci environment. It's running pytest instead of sphinx

When it's ready, you should be the one who merges it.

tox.ini
Cadair: Why don't we change the line above so all the envs use the same sample data in `.tox/sample_data` then we can restore the same cache for the online builds later etc?

Cadair: The existing files shouldn't ever change.

ConorMacBride: Okay I'll make `.tox/sample_data` the default.

ConorMacBride:
I tried caching the sample data in the online test such that it's only restored if _sample.py and sample.py haven't been modified:
```yaml
- linux: py39-online
  cache-path: .tox/sample_data/
  cache-key: sample-${{ hashFiles('sunpy/data/_sample.py', 'sunpy/data/sample.py') }}
```

However, it doesn't work as the repo is not checked out at this stage. I'll remove this for now, but I've opened an issue with a workaround: https://github.com/OpenAstronomy/github-actions-workflows/issues/47

.github/workflows/ci.yml
```diff
@@ -50,9 +50,29 @@ jobs:
...
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@main
+    with:
+      default_python: '3.8'
+      submodules: false
+      pytest: false
+      toxdeps: tox-pypi-filter
+      cache-path: |
+        docs/_build/
+        docs/generated/
+        .tox/sample_data/
```
Cadair, line 64: Can we only restore one cache per build? Having this cached like this means that the second build doesn't need to download it, but ideally  the first build could restore it from a cache as well.

ConorMacBride: Our reusable workflow only supports one cache per build. Otherwise we would need GitHub Actions to support `for` loops, or we could implement a fixed number of caches with `cache1-path`, `cache2-path`, and `cache3-path` etc. instead of just the one `cache-path`.

However, we could do:
```yaml
cache-path: |
  docs/_build/
  docs/generated/
  .tox/sample_data/
cache-key: docs-${{ github.run_id }}
cache-restore-keys: docs-
```

AFAIK in the first docs build this will restore the most recent cache with a key matching `docs-*` and then save a new cache with the `docs-${{ github.run_id }}` at the end of that build. Then in the docs gallery build it would restore `docs-${{ github.run_id }}`. I really don't think we should do this as we want to build the docs with clean `_build/` and `generated/` directories, and we can't restore the `sample_data/` directory only from the less specific cache.

ConorMacBride: In the Azure template we could just iterate through a YAML list 😞 https://openastronomy-azure-pipelines.readthedocs.io/en/latest/run_tox_env.html#caching

Cadair: Yeah well BOOOoooOoOOoo

ConorMacBride:
I can double check the CI logs later and merge.

Cadair:
The build logs on the two docs jobs look good to me.

ConorMacBride:
Yes, they both look to be working properly

ConorMacBride:
Why hasn't the backport bot run on this?

nabobalis:
It hasn't run on another PR, it might be down/having problems.

nabobalis:
Cherry-picked to the 4.0 branch

Issue #6164, opened 2022-05-18
Fix oldest deps CI build


nabobalis:
### Provide a general description of the issue or problem.

Since I merged in https://github.com/sunpy/sunpy/pull/5951 the oldest deps build is failing. We should fix it.

Also address https://github.com/sunpy/sunpy/pull/5951#discussion_r874850237

Issue #5904, opened 2022-02-22
[3.1.4] Type error because of empty pytest.warns


olebole:
### Describe the bug

While packagingsunpy 3.1.4 for Debian, I observerd a failure originating in a wrong call of `pytest.warns`.

### To Reproduce

Run the test suite on version 3.1.4.

### What happened?

I observed the following failure:
```
__________________ test_check_for_nonfinite_entries _____________________
    def test_check_for_nonfinite_entries():
        with warnings.catch_warnings(record=True) as warning_list:
            a = np.zeros((3, 3))
            b = np.ones((3, 3))
            check_for_nonfinite_entries(a, b)
    
        assert len(warning_list) == 0
    
        for i in range(0, 9):
            for non_number in [np.nan, np.inf]:
                a = np.ones(9)
                a[i] = non_number
                b = a.reshape(3, 3)
    
                with pytest.warns(SunpyUserWarning,
                                  match='The layer image has nonfinite entries.') as warning_list:
                    check_for_nonfinite_entries(b, np.ones((3, 3)))
    
                assert len(warning_list) == 1
    
                with pytest.warns(SunpyUserWarning,
                                  match='The template image has nonfinite entries.') as warning_list:
                    check_for_nonfinite_entries(np.ones((3, 3)), b)
    
                assert len(warning_list) == 1
    
>               with pytest.warns() as warning_list:
E               TypeError: warns() missing 1 required positional argument: 'expected_warning'

sunpy/image/tests/test_coalignment.py:104: TypeError
```

### Expected behavior

The tests should all pass. However, checking the [API of `pytest.warns`](https://docs.pytest.org/en/6.2.x/reference.html#pytest-warns), I would expect this error; I am wondering why this did not happen in the Github CI.

### System Details

* Sunpy 3.1.4
* Astropy 5.0.1
* pytest 6.2.5
* Python 3.9 and 3.10

### Installation method

pip

dstansby:
This is probably because we are now using `pytest>=7` to test, and I don't think we have or document a minimum pytest dependency for running the tests. Would it be helpful for debian if we maintained compatability with pytest==6 for a while?

olebole:
I don't know about the schedule of pytest in Debian. Since this is the only failure, I would propose to keep this one Py6 compatible. And checking the [pytest 7.0 API docs](https://docs.pytest.org/en/7.0.x/reference/reference.html#pytest-warns), I would assume that the `expected_warning` is required? At least, this one
```diff
[...]

PR #6165, opened 2022-05-19
Bump min versions to deal with oldest deps failure due to merged PR

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/6164

The timeseries HTML PR ended up breaking out oldest deps builds due to pytest and pytest-xdist. Bumping that fixed that problem but it exposed that the timeseries HTML repr items need MPL 3.4 as a min. So I bumped that too.

Since this is for 4.1, in ~5 months, this feels like the easy solution to this problem.
Decided to then update the MPL version for the figure tests and the image comparison page shows some fun diffs.

In addition, this PR redoes the mocking for each HTML/WEBPAGE test to use pytest-mock instead.
I can do the rest but I wanted to double check we want to do this before I did so.


sunpy/map/tests/test_mapsequence.py
```diff
@@ -169,21 +168,18 @@ def test_repr_html(mapsequence_all_the_same):
         assert m._repr_html_() in html_string
 
 
-def test_quicklook(mapsequence_all_the_same):
-    with mock.patch('webbrowser.open_new_tab') as mockwbopen:
-        mapsequence_all_the_same.quicklook()
-
+def test_quicklook(mocker, mapsequence_all_the_same):
+    mockwbopen = mocker.patch('webbrowser.open_new_tab')
```
nabobalis, line 172: Using mocker we can do this, if we want to replace all of the mocks, we would need to follow this pattern inside the  test as far as I can tell.

setup.cfg
```diff
@@ -54,7 +54,7 @@ image =
 jpeg2000 =
   glymur>=0.8.18,!=0.9.0,!=0.9.5
 map =
-  matplotlib>=3.2.0
+  matplotlib>=3.4.0
```
Cadair, line 57: According to our [Dependancy Support Policy](https://docs.sunpy.org/en/latest/dev_guide/contents/dependencies.html) we can't do this.

Matplotlib 3.4 was released in March 2021 meaning we can't depend on it until March 2023 which means 5.0 not 4.1.

What does the html repr rely on in mpl 3.4?

setup.cfg
```diff
@@ -83,8 +83,8 @@ tests =
   pytest-doctestplus>=0.5 # We require the newest version of doctest plus to use +IGNORE_WARNINGS
   pytest-mock
   pytest-mpl>=0.12 # First version to support our figure tests
-  pytest-xdist>=1.27
-  pytest>=5.4
+  pytest-xdist>=2.0
+  pytest>=6.0
```
Cadair, line 87: Didn't we have a conversation recently about pytest versions because of debain or something?

Ah it was this: https://github.com/sunpy/sunpy/issues/5904 looks fine then.

Cadair (changes requested):
I don't think we can bump the mpl version unless we *really* need to as it would be a breach of our dependency policy (by 5 months).

dstansby (changes requested):
Agreed that we can't bump Matplotlib to 3.4

sunpy/map/tests/test_mapsequence.py
nabobalis: Its ok.

nabobalis:
Ok, I will undo the pin and work out why its failing and if I can fix it.

setup.cfg
ConorMacBride: If I remember correctly supylabel was only added in 3.4?

nabobalis: Yeah that is the problem and why I originally bumped it to 3.4 

Now I need to figure a work around.

nabobalis: Delete the line, sorted

nabobalis:
I did accidently bump to 3.3, I think thats ok?

Cadair:
3.3 is fine for 4.1 :+1:

nabobalis:
Can I do a follow up PR bumping all the packages and adding to the whatsnew?

Cadair (approved):

ConorMacBride (approved):

tox.ini
```diff
@@ -58,16 +58,16 @@ deps =
...
+    oldestdeps: pytest-xdist<2.1
     # These are specific extras we use to run in specific factors
     online: pytest-rerunfailures
     online: pytest-timeout
     online: astroquery
     hypothesis: astroquery
     # Figure tests need a tightly controlled environment
     figure-!devdeps: astropy==4.2.1
-    figure-!devdeps: matplotlib==3.3.2
+    figure-!devdeps: matplotlib==3.5.2
```
dstansby, line 70: Now this has been bumped, does the `.json` file at https://github.com/sunpy/sunpy/tree/main/sunpy/tests need renaming for the new version?

dstansby (changes requested):

tox.ini
dstansby: My bad, I see this has been done

dstansby (approved):

PR #6166, opened 2022-05-19
Add an environment variable which can skip compiling the ANA extension

Cadair:
This is to help out in https://github.com/holzschu/Carnets/issues/226

### DIFF
--- a/sunpy/io/setup_package.py
+++ b/sunpy/io/setup_package.py
@@ -9,7 +9,7 @@ from setuptools import Extension
 
 def get_extensions():
 
-    if get_compiler() == 'msvc':
+    if get_compiler() == 'msvc' or os.environ.get("SUNPY_NO_BUILD_ANA_EXTENSION", None):
         return list()
     else:
         cfg = defaultdict(list)
--- a/changelog/6166.trivial.rst
+++ b/changelog/6166.trivial.rst
@@ -1,0 +1,2 @@
+Add an environment variable ``SUNPY_NO_BUILD_ANA_EXTENSION`` which when present
+will cause sunpy to not compile the ANA C extension when building from source.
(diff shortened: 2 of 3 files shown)


sunpy/io/setup_package.py
```diff
@@ -9,7 +9,7 @@
 
 def get_extensions():
 
-    if get_compiler() == 'msvc':
+    if get_compiler() == 'msvc' or os.environ.get("SUNPY_NO_BUILD_ANA_EXTENSION"], None):
```
holzschu, line 12: I think you have an extra "]" here.

ConorMacBride (approved):
I ran `python -m build --wheel .` on this PR and it built `sunpy-4.1.dev95+g7d4ebe66c-cp39-cp39-macosx_10_9_x86_64.whl`, while running `SUNPY_NO_BUILD_ANA_EXTENSION=1 python -m build --wheel .` built `sunpy-4.1.dev95+g7d4ebe66c-py3-none-any.whl` so this seems to work.

I guess sunpy isn't going to add more extensions anytime soon, but would it be better to use `SUNPY_NO_BUILD_EXTENSIONS` and check in `setup.py` instead?

nabobalis:
> I ran `python -m build --wheel .` on this PR and it built `sunpy-4.1.dev95+g7d4ebe66c-cp39-cp39-macosx_10_9_x86_64.whl`, while running `SUNPY_NO_BUILD_ANA_EXTENSION=1 python -m build --wheel .` built `sunpy-4.1.dev95+g7d4ebe66c-py3-none-any.whl` so this seems to work.
> 
> I guess sunpy isn't going to add more extensions anytime soon, but would it be better to use `SUNPY_NO_BUILD_EXTENSIONS` and check in `setup.py` instead?

I feel like we would want:
1. Per compiled flag, 
2. Overall complied flag

Doesn't have to be this PR.

nabobalis (approved):

Cadair:
I would rather not promise that any future C extensions will be optional.

nabobalis:
Cherry-picked to the 4.0 branch

PR #6167, opened 2022-05-20
Add vcrpy testing config

dstansby:
This PR enables `vcrpy` testing with remote data tests. It includes:

- Change the `online` tox environment to download cassettes from https://github.com/dstansby/sunpy-vcr-cassettes before running the online tests
- A new tox environment to re-generate all the cassettes
- A new pytest fixutre to configure where the generated cassettes are stored
- Add dev docs for the above

If/when this is merged, we can move https://github.com/dstansby/sunpy-vcr-cassettes to the `sunpy` org. See https://github.com/dstansby/sunpy-vcr-cassettes#workflow for how the cassettes are automaticaly updated.


tox.ini
```diff
@@ -79,14 +82,16 @@ extras =
 commands =
     pip freeze --all --no-input
     oldestdeps: python -c "import astropy.time; astropy.time.update_leap_seconds()"
-    !online-!hypothesis-!figure: {env:PYTEST_COMMAND} {posargs}
-    online: {env:PYTEST_COMMAND} --hypothesis-show-statistics --reruns 2 --reruns-delay 15 --timeout=180 --remote-data=any {posargs}
+    !online-!hypothesis-!figure-!vcr: {env:PYTEST_COMMAND} {posargs}
+    online: git clone "https://github.com/dstansby/sunpy-vcr-cassettes" {envtmpdir}/sunpy-vcr-cassettes
```
ConorMacBride, line 86: Will we want to use the GitHub Actions cache for `{envtmpdir}/sunpy-vcr-cassettes`? Using the dynamic `{envtmpdir}` might make it more tricky to restore the cache.

ConorMacBride (approved):
This looks brilliant! It's nice that this just works without writing the tests especially for it 👍

docs/dev_guide/contents/tests.rst
```diff
@@ -87,6 +87,15 @@ or::
 
     $ pytest --remote-data=any
 
+
+Mocked remote data tests
+------------------------
+Some of our remote data tests are run using `vcrpy <https://vcrpy.readthedocs.io/en/latest/>`_, which saves a copy of any HTTP requests made within the test so a remote request doesn't have to be run on subsequent test runs.
```
nabobalis, line 93:
```suggestion
Some of our remote data tests are run using `vcrpy <https://vcrpy.readthedocs.io/en/latest/>`__, which saves a copy of any HTTP requests made within the test so a remote request doesn't have to be run on subsequent test runs.
```

docs/dev_guide/contents/tests.rst
```diff
@@ -87,6 +87,15 @@ or::
...
+To use ``vcrpy`` on a test add the ``@pytest.mark.vcrpy()`` decorator to the test, and **keep the ``@pytest.mark.remote_data`` decorator**.
+These should only be used to test search requests, and **not** data download requests, to avoid storing large files in the git repository.
```
nabobalis, line 95: How bad is it if we were to try to download the data?

sunpy/net/cdaweb/test/test_cdaweb.py
```diff
@@ -27,3 +25,4 @@ def test_download():
     assert len(result[0]) == 4
     files = Fido.fetch(result)
     assert len(files.errors) == 0
+    assert len(files) == 4
```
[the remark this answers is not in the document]
nabobalis: I think it should renamed.

docs/dev_guide/contents/tests.rst
```diff
@@ -87,6 +87,15 @@ or::
 
     $ pytest --remote-data=any
 
+
+Mocked remote data tests
+------------------------
+Some of our remote data tests are run using `vcrpy <https://vcrpy.readthedocs.io/en/latest/>`_, which saves a copy of any HTTP requests made within the test so a remote request doesn't have to be run on subsequent test runs.
+To use ``vcrpy`` on a test add the ``@pytest.mark.vcrpy()`` decorator to the test, and **keep the ``@pytest.mark.remote_data`` decorator**.
```
nabobalis, line 94: Does it need  the `()`? I would like to avoid if it we can.

tox.ini
dstansby: We could hard code a path that isn't {envtmpdir}. Is the suggestion to cache the whole `sunpy-vcr-cassettes` git repository, and then just run a `git pull` (or something similar) to update it instead of doing a full clone each time? If so I think that's a good idea.

docs/dev_guide/contents/tests.rst
dstansby: The cassette files end up being as large as the downloaded files. So mileage may vary depending on the data being downloaded. Perhaps lets start with non-binary files, and then we can have a discussion about pros/cons of including the binary data?

docs/dev_guide/contents/tests.rst
dstansby: Any reason apart for aesthetic reasons? I just copied the `pytest-vcr` docs.

tox.ini
```diff
@@ -79,14 +82,16 @@ extras =
 commands =
     pip freeze --all --no-input
     oldestdeps: python -c "import astropy.time; astropy.time.update_leap_seconds()"
-    !online-!hypothesis-!figure: {env:PYTEST_COMMAND} {posargs}
-    online: {env:PYTEST_COMMAND} --hypothesis-show-statistics --reruns 2 --reruns-delay 15 --timeout=180 --remote-data=any {posargs}
```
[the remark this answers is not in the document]
dstansby: Probably, I'll put this back

docs/dev_guide/contents/tests.rst
nabobalis: We can punt it to another PR/issue for the future, I just wondered if we could use this for the .

In theory if we never change the times, the data should be same, so I can't imagine it being too much of a problem for a throw away git repo to add it to the history.

docs/dev_guide/contents/tests.rst
nabobalis: Purely aesthetic reason, it would match the our other decorators that have no args, mainly the remote data one.

docs/dev_guide/contents/tests.rst
Cadair: We should probably avoid this for sanity's sake, we shouldn't need very many tests actually downloading files. We can either pick downloads which are very small for testing fetch or we can make those tests actually online, or we can vcr the query and fetch the files?!

nabobalis: Yeah I agree, we would make sure we review our download tests and ensure we arent wasting our time with them.

But the ones that remain, maybe we can try to use vcrpy on them?

Cadair: I am fine with doing that iff we use git-lfs in the cassettes repo.

tox.ini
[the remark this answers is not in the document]
dstansby: Actually I've changed my mind - this will just re-run *any* test that fails, which if a bunch of stuff fails for non-remote reasons is a waste of test time. I'm going to leave this out, and if there are specific remote tests that need retrying we can mark them as such in the future.

tox.ini
dstansby: I think I'll leave this for now as it's not very trivial. We can alwasy add it later if/when the `sunpy-vcr-cassettes` repo grows in size.

docs/dev_guide/contents/tests.rst
dstansby: Link seems to work fine with one underscore?

docs/dev_guide/contents/tests.rst
dstansby: I'm going to leave the brackets as it's what's shown in the `pytest-vcrpy` docs.

tox.ini
ConorMacBride: I forgot that the cache would have to interact with git. Maybe just leave as-is for now. If we change it to git pull we would also need to have it git clone if the pull fails. We can change the path and add caching later.

docs/dev_guide/contents/tests.rst
nabobalis: It does but if I recall RST convention for links in the middle of sentences should have a double underscore.

nabobalis (approved):

docs/dev_guide/contents/tests.rst
Cadair: It's generally a good idea for external URLs to use double underscore.

IIRC the difference between a single and a double is how rst handles repeats of links to the same target. If you link to the same external URL twice when using a single underscore it can cause issues.

nabobalis:
@Cadair you can merge when you are happy with it.

Cadair (changes requested):
It looks good, I think before we merge this we might as well migrate the repo into the sunpy org. If you transfer it to me, I shall move it in.

I see that we have only added this to the cdaweb and DR tests, I would be really interested to know if this works *at all* with VSO as it's SOAP requests. Do you fancy giving that a try? (Not that I think it's a deal breaker, but I think it would be good to know if this is likely to turn into a complete solution for our online tests or not before we adopt the infrastructure required to run it.

(Requesting changes just to put it in a holding pattern while we migrate the repo)

nabobalis:
Changelog and backport?

dstansby:
👍 to changelog, 👎 to backport.

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    py{37,38,39,310}{,-oldestdeps,-devdeps,-online,-figure,-hypothesis,-conda}
+    py{37,38,39,310}{,-oldestdeps,-devdeps,-online,-figure,-hypothesis,-conda,-vcr}
```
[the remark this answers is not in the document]
dstansby: I created a new environment called just `vcr`

dstansby:
> I see that we have only added this to the cdaweb and DR tests, I would be really interested to know if this works _at all_ with VSO as it's SOAP requests. Do you fancy giving that a try? (Not that I think it's a deal breaker, but I think it would be good to know if this is likely to turn into a complete solution for our online tests or not before we adopt the infrastructure required to run it.

I've updated all the tests I think can use `vcr` at the moment, which includes some VSO tests; see https://github.com/dstansby/sunpy-vcr-cassettes/commit/2393268b164c18c0b95b7c740e661b7e7f8a37f9 for the newly recorded cassettes.

> (Requesting changes just to put it in a holding pattern while we migrate the repo)

I'm away for a week or so, so migth work on this at odd times. Given the transfer timeout, we can try and do it next week during sensible working hours 😄

dstansby:
🎉 as evidence of this working, even though there are pre-existing remote data fails the remote tests took 10 minutes to run, vs. 20 minutes on the previous daily CI job.

tox.ini
```diff
@@ -79,8 +83,9 @@ extras =
 commands =
     pip freeze --all --no-input
     oldestdeps: python -c "import astropy.time; astropy.time.update_leap_seconds()"
-    !online-!hypothesis-!figure: {env:PYTEST_COMMAND} {posargs}
-    online: {env:PYTEST_COMMAND} --hypothesis-show-statistics --reruns 2 --reruns-delay 15 --timeout=180 --remote-data=any {posargs}
+    !online-!hypothesis-!figure-!vcr: {env:PYTEST_COMMAND} {posargs}
+    online: git clone "https://github.com/dstansby/sunpy-vcr-cassettes" {envtmpdir}/sunpy-vcr-cassettes
```
nabobalis, line 87:
```suggestion
    online: git clone "https://github.com/sunpy/sunpy-vcr-cassettes" {envtmpdir}/sunpy-vcr-cassettes
```

nabobalis:
Rebased, I think we can merge when things pass

docs/dev_guide/contents/tests.rst
```diff
@@ -119,6 +119,15 @@ or::
...
 
+
+Mocked remote data tests
+------------------------
+Some of our remote data tests are run using `vcrpy <https://vcrpy.readthedocs.io/en/latest/>`__, which saves a copy of any HTTP requests made within the test so a remote request doesn't have to be run on subsequent test runs.
+To use ``vcrpy`` on a test add the ``@pytest.mark.vcrpy()`` decorator to the test, and keep the ``@pytest.mark.remote_data`` decorator.
+These should only be used to test search requests, and **not** data download requests, to avoid storing large files in the git repository.
+Every week the ``sunpy-vcr-cassettes`` repository will update the cassettes from the sunpy core main branch.
+After this is done, the sunpy core remote data tests will use these cassettes to avoid remote requests in the future.
+
```
Cadair, line 130: I think we are missing some info on how all this works, i.e. how someone might use this without tox here. I just had to scroll through this PR for a minute or two to work it out.

docs/dev_guide/contents/tests.rst
```diff
@@ -119,6 +119,15 @@ or::
...
     $ pytest --remote-data=any
 
+
+Mocked remote data tests
+------------------------
+Some of our remote data tests are run using `vcrpy <https://vcrpy.readthedocs.io/en/latest/>`__, which saves a copy of any HTTP requests made within the test so a remote request doesn't have to be run on subsequent test runs.
+To use ``vcrpy`` on a test add the ``@pytest.mark.vcrpy()`` decorator to the test, and keep the ``@pytest.mark.remote_data`` decorator.
+These should only be used to test search requests, and **not** data download requests, to avoid storing large files in the git repository.
+Every week the ``sunpy-vcr-cassettes`` repository will update the cassettes from the sunpy core main branch.
+After this is done, the sunpy core remote data tests will use these cassettes to avoid remote requests in the future.
```
Cadair, line 129: If a new test is added which is missing a cassette it will do the request until such time as the cassette is added?

Cadair:
Also I think we need to close https://github.com/sunpy/sunpy-vcr-cassettes/issues/2 before we can merge this.

My main worry with this PR is that something is going to change in one of the remote servers and we aren't going to notice for ages, which means we wont patch it and our users will suffer.

I am not sure what a good solution for monitoring that is, if the online build was more reliable then we could use the cassettes build failure as a notification we need to look at it, but knowing how often our online suite borks at the moment I am not sure that's going to work. Anyone got any better ideas?

Also, what's the advantage of rebuilding the cassettes on the cassettes repo rather than in the regular sunpy cron CI jobs?

docs/dev_guide/contents/tests.rst
dstansby: Yes

Cadair: That's nice.

dstansby:
I've lost track of this and any context that was previously in my head, and am not planning on returning to it in the near future. If anyone else wants to pick this up and run with it feel free! I'll close this PR but leave the branch around.

PR #6169, opened 2022-05-21
Fix tox conda env spec

dstansby:
Not sure whether this is intended or a bug in `tox`. Either way this should fix running the tox job. I've tested this locally and it works.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -121,7 +121,7 @@ commands =
     python -c "import sunpy; sunpy.self_test()"
 
 # Requires tox-conda
-[testenv:conda]
+[testenv:py{37,38,39,310}-conda]
 pypi_filter =
 extras =
 deps =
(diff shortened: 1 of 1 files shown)


nabobalis (approved):

Cadair (approved):
Pretty sure this is "expected behaviour" in tox 3.

tox.ini
```diff
@@ -121,7 +121,7 @@ commands =
     python -c "import sunpy; sunpy.self_test()"
 
 # Requires tox-conda
-[testenv:conda]
+[testenv:py{37,38,39,310}-conda]
```
nabobalis, line 124:
```suggestion
[testenv:py{38,39,310}-conda]
```
Did we drop 3.7?

nabobalis:
The label doesn't seem to do anything?

nabobalis:
Oh its there now after the commit update

nabobalis (approved):

nabobalis:
Oh man, I wonder if we want a `skip_conda` flag.

PR #6170, opened 2022-05-21
Build aarch64 wheels

### LINKED ISSUE #5772: Add support to release linux aarch64 wheels
On aarch64, `pip install sunpy` build the wheels from source code and then installs it. It requires the user to have a development environment installed on their system. Also, it takes more time to build the wheels than downloading and extracting the wheels from PyPI. 

I built the aarch64 wheels successfully on azure pipeline but the job gets timed out during testing due to emulation.  

@nabobalis, Please let me know your interest in releasing aarch64 wheels. I can help with this.

ConorMacBride:
Closes #5772

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,6 +143,7 @@ jobs:
         - cp3{8,9,10}-manylinux*_x86_64
         - cp3{8,9,10}-macosx_x86_64
         - cp3{8,9,10}-macosx_arm64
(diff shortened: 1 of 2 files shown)


ConorMacBride:
It still ran `pytest` because cibuildwheel always uses the env var if set. Anyway, just installing sunpy and its dependencies took far to long to install (maybe 40 min?) so still not feasible.

PR #6171, opened 2022-05-21
Use Python Limited API

ConorMacBride:
Not sure if we want to do this. Basically, it compiles the extensions in such a way that they are compatible with all the Python versions, so we only need one wheel per OS. It will still test the wheel across all supported Python version so there won't be much of a speed up when publishing.

The ANA tests passed for me locally on macOS. Will see if all the tests pass in the CI. This is assuming the tests have good coverage of the C extension.

This is based on https://github.com/astrofrog/fast-histogram/pull/56. See also [PEP 384 – Defining a Stable ABI](https://peps.python.org/pep-0384/).

### DIFF
--- a/changelog/6171.trivial.rst
+++ b/changelog/6171.trivial.rst
@@ -1,0 +1,2 @@
+``sunpy`` now uses the `Limited Python API <https://docs.python.org/3/c-api/stable.html>`__.
+Therefore, one binary distribution (wheel) per platform is now published and it is compatible with all Python versions ``sunpy`` supports.
--- a/setup.cfg
+++ b/setup.cfg
@@ -110,6 +110,10 @@ asdf.resource_mappings =
 asdf.extensions =
   sunpy = sunpy.io.special.asdf.entry_points:get_extensions
 
+[bdist_wheel]
+# the Py_LIMITED_API version hex in _pyana.c should match the version specified here
+py_limited_api = cp38
+
 [tool:pytest]
 testpaths = "sunpy" "docs"
 norecursedirs = ".tox" "build" "docs[\/]_build" "docs[\/]generated" "*.egg-info" "examples" "sunpy[/\]_dev" ".jupyter" ".history" "tools" "sunpy[\/]extern"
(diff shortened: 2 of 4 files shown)


nabobalis:
Is there a downside?

ConorMacBride:
Okay. This PR adds to sunpy what the extension-helpers PR would add automatically. (I accidentally set the `Py_LIMITED_API` macro for `cp36` instead of `cp38`, which I've now fixed here.)

setup.cfg
```diff
@@ -110,6 +110,9 @@ asdf.resource_mappings =
 asdf.extensions =
   sunpy = sunpy.io.special.asdf.entry_points:get_extensions
 
+[bdist_wheel]
+py_limited_api = cp38
```
Cadair, line 114: Put a comment here saying if you change this you need to change the line in `_pyana.c` or else we *will* forget.

nabobalis (approved):

Cadair:
If we rebase this we can see the CI run?

ConorMacBride:
Does this need a changelog?

nabobalis:
Up to you, we could add a trivial one?

changelog/6171.trivial.rst
```diff
@@ -0,0 +1 @@
+SunPy now uses the `Limited Python API <https://docs.python.org/3/c-api/stable.html>`__. Therefore, one binary distribution (wheel) per platform is now published and it is compatible with all Python versions SunPy supports.
```
nabobalis, line 1:
```suggestion
``sunpy`` now uses the `Limited Python API <https://docs.python.org/3/c-api/stable.html>`__.
Therefore, one binary distribution (wheel) per platform is now published and it is compatible with all Python versions ``sunpy`` supports.
```

setup.cfg
nabobalis: Silly question, how does one get the version hex? Is it just a conversion of the number to hexadecimal?

ConorMacBride: It's defined here: https://docs.python.org/3.12/c-api/apiabiversion.html

And more general details here: https://docs.python.org/3/c-api/stable.html#c.Py_LIMITED_API

PR #6175, opened 2022-05-23
Publish pure Python wheels

ConorMacBride:
Based on discussions in #5772, this adds a pure Python wheel to the publishing workflow. (I think this PR should probably close that issue?) The pure Python wheel would be installed (by pip) when a platform specific compiled wheel is not available. 

For example, by default pip on aarch64 will install the pure Python wheel (without the `sunpy.io.ana` compiled extension) rather than compiling from the sdist. And for macOS, it would continue to use the specific compiled wheel rather than the pure Python wheel.

Due to how we detect whether to build pure python or not on GitHub Actions, this means the `ana` test no longer run when testing the sdist on GitHub Actions. I don't think this is an issue.

- [x] Update documentation
- [x] Document how to install sdist instead of pure wheel (pip `--no-binary :all:` option)
- [x] Improve `ana.py` import errors https://github.com/sunpy/sunpy/blob/5a351a8e6161cab6bc4cb7974d0d8a93ad92a3ef/sunpy/io/ana.py#L51

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,6 +143,7 @@ jobs:
     needs: [test]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@main
     with:
+      sdist: false
       test_extras: 'dev'
       test_command: 'pytest -p no:warnings --doctest-rst -m "not mpl_image_compare" --pyargs sunpy'
       submodules: false
@@ -153,9 +154,21 @@ jobs:
     secrets:
       pypi_token: ${{ secrets.pypi_token }}
 
+  publish_pure:
+    needs: [publish]
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@main
+    with:
+      test_extras: 'dev'
+      test_command: 'pytest -p no:warnings --doctest-rst -m "not mpl_image_compare" --pyargs sunpy'
+      submodules: false
+      env: |
+        SUNPY_NO_BUILD_ANA_EXTENSION: 1
+    secrets:
+      pypi_token: ${{ secrets.pypi_token }}
+
   notify:
     if: always() && github.event_name == 'workflow_dispatch'
-    needs: [publish, online, cron]
+    needs: [publish_pure, online, cron]
     runs-on: ubuntu-latest
     steps:
       - uses: Cadair/matrix-notify-action@main
--- a/changelog/6175.feature.rst
+++ b/changelog/6175.feature.rst
@@ -1,0 +1,6 @@
+A pure Python ``sunpy`` wheel is now published on PyPI with each release.
+``pip`` will now default to installing the pure Python wheel instead of the source distribution on platforms other than Linux (x86-64) and macOS (x86-64 and ARM64).
+This should mean simpler and faster installs on such platforms, which includes the Raspberry Pi as well as some cloud computing services.
+
+This wheel does not contain the ``sunpy.io.ana`` compiled extension.
+If you need this extension (not available on Windows) you can install the ``sunpy`` source distribution with ``pip install --no-binary sunpy "sunpy[all]"``.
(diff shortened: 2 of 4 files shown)


nabobalis:
Worth backporting to 4.0?

ConorMacBride:
It would be nice to backport it. Although if someone was running `4.0.0` on e.g. aarch64, they would have compiled from source, however, when they upgrade to `4.0.1` they would get the pure Python wheel which doesn't include `ana`. So from the point of view of the user, a feature is removed by a bugfix release. That's an unlikely case so maybe we should backport??

sunpy/io/setup_package.py
```diff
@@ -9,7 +9,11 @@
 
 def get_extensions():
 
-    if get_compiler() == 'msvc' or os.environ.get("SUNPY_NO_BUILD_ANA_EXTENSION", None):
+    if (
+        get_compiler() == 'msvc' or
+        os.environ.get("SUNPY_NO_BUILD_ANA_EXTENSION", None) or
+        "/OpenAstronomy/build-python-dist/" in os.environ.get("GITHUB_ACTION_PATH", "")
```
Cadair, line 15: Is this because we can't pass through env vars to the action workflows?

ConorMacBride: Unfortunately yes. It would be better if we could avoid GitHub Actions specific code inside the package

Cadair: Why can't we pass env vars through again?

ConorMacBride: It's in the list of limitations and I have tried before and it doesn't get passed through. (They do get passed into actions though.) https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations

Cadair: Can we hack round that in the workflow? Add it as a custom var which gets parsed by our python script?

ConorMacBride: Yes. We could add a step that runs this bash command for each line in a string of env vars. https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

ConorMacBride: I've created an issue https://github.com/OpenAstronomy/github-actions-workflows/issues/52

nabobalis:
> @nabobalis this needs the workflow to support an input for passing in env vars. Do you have any thoughts on the syntax here? [OpenAstronomy/github-actions-workflows#56 (comment)](https://github.com/OpenAstronomy/github-actions-workflows/pull/56#issuecomment-1425089211)

👍

ConorMacBride:
This appears to be successfully building the pure Python wheel, with just an env var set!

nabobalis:
I think this is worth adding a changelog.

ConorMacBride:
`feature` or `trivial` changelog?

nabobalis:
I consider this a new feature.

ConorMacBride:
I tagged to close #5772 as I think aarch64 wheels without the ana extension would be sufficient for almost all users.

nabobalis (approved):

nabobalis:
> I tagged to close #5772 as I think aarch64 wheels without the ana extension would be sufficient for almost all users.

Makes sense.

About your last bullet point. Should we link to a webpage? I think trying to explain how to fix that issue in a message might be impossible.

ConorMacBride:
Good idea! I changed the message to:
> C extension for ANA is missing. For more details see: https://docs.sunpy.org/en/stable/installation.html#installing-without-conda

Cadair:
If we are putting that link in that error we should add a comment to the page about what platforms we don't have ana on?

ConorMacBride:
I've reworded the info on the Python wheels in the installation guide. I've stated what platforms have a complied wheel, and made it clear that all other Python implementations and platforms will need to compile from source if they want the extension.

nabobalis (approved):

Cadair (approved):

Cadair:
Thanks @ConorMacBride !

PR #6178, opened 2022-05-25
Pin hypothesis

ConorMacBride:


ConorMacBride:
Pinning fixes the CI so we should merge this. However, I'll open an issue later with the specific example(s) that are failing as hypothesis is not the problem, it's just the new version of hypothesis brings the issue to light.

nabobalis (approved):

PR #6184, opened 2022-05-26
Fixed various issues with tests that were revealed by new `hypothesis` float sampling

### LINKED ISSUE #6180: Edge case when transforming from `NorthOffsetFrame` to `'heliographic_stonyhurst'`
### Describe the bug

Hypothesis found a falsifying example for `sunpy/coordinates/tests/test_offset_frame.py::test_transformation`, which is "Test[ing] that the north pole in the new frame transforms back to the given north argument."

@ayshih could you take a look please?

(This issue has come to light as hypothesis have recently changes how they deal with the floats strategy: https://github.com/HypothesisWorks/hypothesis/releases/tag/hypothesis-python-6.46.8)

### To Reproduce

```py
import astropy.units as u
from astropy.coordinates import SkyCoord, Longitude, Latitude
from sunpy.coordinates import NorthOffsetFrame


# From sunpy/coordinates/tests/test_offset_frame.py
def test_transform(lon, lat):
    """
    Test that the north pole in the new frame transforms back to the given
    north argument.
    """
    north = SkyCoord(lon=lon, lat=lat, frame='heliographic_stonyhurst')
    off = NorthOffsetFrame(north=north)
    t_north = SkyCoord(lon=0*u.deg, lat=90*u.deg, frame=off)
    t_north = t_north.transform_to('heliographic_stonyhurst')

    # From sunpy/coordinates/tests/helpers.py
    difference = Longitude(north.lon - t_north.lon, wrap_angle=180 * u.deg)
    return difference


# Falsifying example
difference = test_transform(Longitude(0, u.deg), Latitude(-90, u.deg))
print(f"Difference should be 0 deg but is {difference}.")
```

### What happened?

```
Difference should be 0 deg but is 116.565051177078 deg.
```

### Expected behavior

```
Difference should be 0 deg but is 0 deg.
```

### Screenshots

_No response_

### System Details

```
==============================
sunpy Installation Information
==============================

General
#######
OS: Mac OS 12.4
Arch: 64bit, (i386)
sunpy: 4.0.0
Installation path: /Users/conor/miniconda3/envs/astropy-old/lib/python3.10/site-packages/sunpy-4.0.0.dist-info

Required Dependencies
#####################
astropy: 5.0.4
numpy: 1.22.4
packaging: 21.3
parfive: 1.5.1

Optional Dependencies
#####################
asdf: Missing asdf>=2.8.0; extra == "asdf"
asdf-astropy: Missing asdf-astropy>=0.1.1; extra == "asdf"
beautifulsoup4: Missing beautifulsoup4>=4.8.0; extra == "net"
cdflib: Missing cdflib!=0.4.0,>=0.3.19; extra == "timeseries"
dask: Missing dask[array]>=2.0.0; extra == "dask"
drms: Missing drms>=0.6.1; extra == "net"
glymur: Missing glymur!=0.9.0,!=0.9.5,>=0.8.18; extra == "jpeg2000"
h5netcdf: Missing h5netcdf>=0.8.1; extra == "timeseries"
h5py: Missing h5py>=3.1.0; extra == "timeseries"
matplotlib: Missing matplotlib>=3.2.0; extra == "map" or "timeseries" or "visualization"
mpl-animators: Missing mpl-animators>=1.0.0; extra == "map" or "visualization"
pandas: Missing pandas>=1.0.0; extra == "timeseries"
python-dateutil: Missing python-dateutil>=2.8.0; extra == "net"
reproject: Missing reproject; extra == "docs" or "map"
scikit-image: Missing scikit-image>=0.16.0; extra == "image"
scipy: Missing scipy>=1.3.0; extra == "image" or "map"
sqlalchemy: Missing sqlalchemy>=1.3.4; extra == "database"
tqdm: 4.64.0
zeep: Missing zeep>=3.4.0; extra == "net"
```

Also broken with astropy `5.1.0`.

### Installation method

pip

ayshih:
Fixes #6180 and reverts #6178

* Now that `hypothesis` fully samples the float range, it revealed that a longitude check in a test did not account for [gimbal lock](https://en.wikipedia.org/wiki/Gimbal_lock) at the poles (latitude of +/-90 degrees), where differences in longitude are meaningless.  This PR switches to using `SkyCoord.separation()`, which is more straightforward.
* `hypothesis` also revealed that an equality check of numbers that could be zero was missing the specification of the absolute tolerance.
* It seems that `numpy.linalg.det()` is incorrectly computing the determinant of a 2x2 matrix with underflow floats (thanks to `hypothesis`) under some confluence of NumPy version, LAPACK compilation, OS, etc.  I don't want to go down that rabbit hole, and it's straightforward to manually compute the determinant of a 2x2 matrix, so I switched to doing that.

### DIFF
--- a/sunpy/map/tests/strategies.py
+++ b/sunpy/map/tests/strategies.py
@@ -1,5 +1,3 @@
-import warnings
-
 import hypothesis.strategies as st
 import numpy as np
 from hypothesis import assume
@@ -19,10 +17,8 @@ def matrix_meta(draw, key):
         float, (2, 2),
         elements=st.floats(min_value=-1, max_value=1, allow_nan=False))
     )
-    # Make sure matrix isn't singular
-    with warnings.catch_warnings():
-        warnings.filterwarnings(action='ignore', category=RuntimeWarning, module='numpy.linalg')
-        assume(np.abs(np.linalg.det(arr)) > 1e-8)
+    # Make sure matrix isn't singular by manually computing the determinant
+    assume(np.abs(arr[1, 1]*arr[0, 0] - arr[0, 1]*arr[1, 0]) > 1e-8)
     return {f'{key}1_1': arr[0, 0],
             f'{key}1_2': arr[0, 1],
             f'{key}2_1': arr[1, 0],
--- a/setup.cfg
+++ b/setup.cfg
@@ -77,7 +77,7 @@ visualization =
   matplotlib>=3.3.0
   mpl-animators>=1.0.0
 tests =
-  hypothesis==6.46.7  # >=6.0.0  # Included in pytest-astropy. 6.0 is the first version to support disabling function-scoped fixture warning
+  hypothesis>=6.0.0  # Included in pytest-astropy. 6.0 is the first version to support disabling function-scoped fixture warning
   jplephem  # For some coordinates tests
   opencv-python
   pytest-astropy>=0.8  # 0.8 is the first release to include filter-subpackage
--- a/sunpy/coordinates/tests/test_metaframes.py
+++ b/sunpy/coordinates/tests/test_metaframes.py
@@ -214,7 +214,7 @@ def test_rotatedsun_transforms(frame, lon, lat, obstime, rotated_time1, rotated_
     desired_delta_lon1 = diff_rot((rotated_time1 - obstime).to(u.day), lat)
 
     assert_longitude_allclose(result1.lon, rsf1.lon + desired_delta_lon1, atol=1e-5*u.deg)
-    assert_quantity_allclose(base.lat, result1.lat)
+    assert_quantity_allclose(base.lat, result1.lat, atol=1e-10*u.deg)
     # Use the `spherical` property since the name of the component varies with frame
     assert_quantity_allclose(base.spherical.distance, result1.spherical.distance)
 
@@ -225,7 +225,7 @@ def test_rotatedsun_transforms(frame, lon, lat, obstime, rotated_time1, rotated_
     desired_delta_lon2 = -diff_rot((rotated_time2 - obstime).to(u.day), lat)
 
     assert_longitude_allclose(result2.lon, rsf2.lon + desired_delta_lon2, atol=1e-5*u.deg)
-    assert_quantity_allclose(base.lat, result2.lat)
+    assert_quantity_allclose(base.lat, result2.lat, atol=1e-10*u.deg)
     # Use the `spherical` property since the name of the component varies with frame
     assert_quantity_allclose(base.spherical.distance, result2.spherical.distance)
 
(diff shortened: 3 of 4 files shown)


nabobalis (approved):

nabobalis:
Or not :P

sunpy/map/tests/strategies.py
```diff
@@ -19,10 +17,8 @@ def matrix_meta(draw, key):
         float, (2, 2),
         elements=st.floats(min_value=-1, max_value=1, allow_nan=False))
     )
-    # Make sure matrix isn't singular
-    with warnings.catch_warnings():
-        warnings.filterwarnings(action='ignore', category=RuntimeWarning, module='numpy.linalg')
-        assume(np.abs(np.linalg.det(arr)) > 1e-8)
```
nabobalis, line 25: We assume somehow the combo of numpy and hypo broke the test on macos?

nabobalis (approved):

sunpy/map/tests/strategies.py
ayshih: Yes.  I believe it also broke under `ubuntu-latest` once, but succeeded on a later run.  (Are we not caching hypothesis failures in any way?)

nabobalis: Ah, I only saw the MacOS fail.

I don't think we are caching anything to do with hypothesis, we could look into it?

ayshih: `hypothesis` naturally caches when run locally, but I don't know how hard it is to set up in the CI.

nabobalis: Conner added support for caching items between builds (like our documentation build).

ConorMacBride (approved):

ayshih:
> It seems that `numpy.linalg.det()` is incorrectly computing the determinant of a 2x2 matrix with underflow floats

For the record, in case anyone wants to investigate, it appears that on the `macos` CI build, and potentially on others as well:
```python
np.abs(np.linalg.det(np.array([[5e-324, 5e-324], [5e-324, 5e-324]])))
```
was evaluating to a value greater than 1e-8 rather than to zero.  I could not reproduce locally, so I suspect it depends on how LAPACK is compiled because `np.linalg.det()` calls a LAPACK routine.

ConorMacBride:
Yeah, on numpy 1.22.4 on macOS I'm getting:
```py
>>> np.abs(np.linalg.det(np.array([[5e-324, 5e-324], [5e-324, 5e-324]])))
/Users/conor/miniconda3/envs/astropy-old/lib/python3.10/site-packages/numpy/linalg/linalg.py:2146: RuntimeWarning: overflow encountered in det
  r = _umath_linalg.det(a, signature=signature)
inf
```
I guess numpy would consider this a bug?

ayshih:
Yikes!  Infinity is not the same as zero.  If you haven't already, I suggest you report that to NumPy!

PR #6195, opened 2022-06-04
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/stale_bot.yml
+++ b/.github/workflows/stale_bot.yml
@@ -9,8 +9,14 @@ on:
 #            │ │ │ │ ┌───────── day of the week (0 - 6 or SUN-SAT)
     - cron: '0 5 * * *' # Every day at 05:00 UTC
 
+permissions:
+  contents: read
+
 jobs:
   stale:
+    permissions:
+      issues: write  # for actions/stale to close stale issues
+      pull-requests: write  # for actions/stale to close stale PRs
     runs-on: ubuntu-latest
     steps:
     - uses: actions/stale@v3
--- a/.github/workflows/changelog_bot.yml
+++ b/.github/workflows/changelog_bot.yml
@@ -10,8 +10,13 @@ env:
   BOT_USER_NAME: 'SunPyBot'
   BOT_USER_EMAIL: 'bot@sunpy.org'
 
+permissions:
+  contents: read
+
 jobs:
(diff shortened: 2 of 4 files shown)


.github/workflows/stale_bot.yml
```diff
@@ -9,8 +9,14 @@ on:
...
     - cron: '0 5 * * *' # Every day at 05:00 UTC
 
+permissions:
+  contents: read
+
 jobs:
   stale:
+    permissions:
+      issues: write  # for actions/stale to close stale issues
+      pull-requests: write  # for actions/stale to close stale PRs
```
ConorMacBride, line 19: Thanks for the PR! To reduce the amount of workflow-level configuration I think we should instead [set the token to be read-only by default in the repository settings](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-the-default-github_token-permissions). That way we would only need to add these three lines. (We don't need to update the changelog workflow as it is disabled.)

nabobalis: I have changed that to be read only on the settings page.

naveensrinivasan: Do you need anything from me? Thanks

ConorMacBride: If you like you can edit the commit to only include these three lines, or we can update it?

naveensrinivasan: I can do that.

ConorMacBride: Thanks!

Issue #5632, opened 2021-10-26
Make sure example gallery is using the smallest possible set of data


dstansby:
The example gallery currently takes a long time to build. One way we could speed this up is by making sure all of the gallery examples use a minimal set of data. As an example, both
- https://docs.sunpy.org/en/stable/generated/gallery/units_and_coordinates/SDO_to_STEREO_Coordinate_Conversion.html#sphx-glr-generated-gallery-units-and-coordinates-sdo-to-stereo-coordinate-conversion-py
- https://docs.sunpy.org/en/stable/generated/gallery/map_transformations/reprojection_different_observers.html#sphx-glr-generated-gallery-map-transformations-reprojection-different-observers-py

could use the same images and Fido search, but currently do not.

dstansby:
I've had a look at adding some STEREO data, and if we just add an image at the same time as the existing AIA data it doesn't play well with reprojection, because the AIA maps have some very high pixel values (see discussion in https://github.com/sunpy/sunpy/issues/2726). So, options are:

- We choose another interval and completely re-do the sample data. This could allow us to choose an interval with Solar Orbiter/EUI and PSP/WISPR images which might be nice...
- We add two new maps: a STEREO and AIA map taken at the same time, but a different time to existing sample data

Cadair:
We are never going to be able to have a single event with all the past and future satellites / instruments running. i.e. RHESSI, both stereos and PSP etc are mutually exclusive at this point. So wholesale redoing all the sample data isn't going to work, as we should keep a RHESSI & AIA pair for composite map contours over image example etc.

While I think we should keep our sample data a) not too large and b) of some kind of interesting events, I would not be adverse to adding a new event or interesting feature to the set to cover new instruments etc.

ayshih:
> it doesn't play well with reprojection, because the AIA maps have some very high pixel values (see discussion in #2726)

It doesn't play well reprojection in a way that isn't mitigated by the `clip_interval` keyword?  I'll note that #2726 predates the addition of that functionality.

dstansby:
The issue is with using the `match_background` setting in `reproject`. I tried masking out the high values pixels from the AIA maps, but the combined reprojected AIA + EUVI map still didn't look right. I lost my test code for this, but will try and put something together again.

dstansby:
In `reprojection_aia_euvi_mosaic.py`, trying to use an EUVI map at the same time as the AIA map in the sample data gives:
[...]

Issue #3514, opened 2019-11-16
Use sample data in examples when not demonstrating download


Cadair:
It's very annoying when examples fail to download data.

nabobalis:
This means an expansion of the sample data.

Raahul-Singh:
Hi I would like to go at it.
It says package intermediate. What parts of the package do I need to be aware of? :sweat_smile:

nabobalis:
In this case you need to be aware of how the sample data is downloaded, where it's downloaded from and how to update it.

dstansby:
Currently examples that use `Fido` are:

```
examples/units_and_coordinates/AIA_limb_STEREO.py
examples/units_and_coordinates/SDO_to_STEREO_Coordinate_Conversion.py
examples/units_and_coordinates/venus_transit.py
examples/map/difference_images.py
examples/plotting/magnetogram_active_regions.py
examples/plotting/solar_cycle_example.py
examples/map_transformations/reprojection_aia_euvi_mosaic.py
examples/map_transformations/reprojection_different_observers.py
examples/time_series/skip_timeseries_example.py
examples/time_series/goes_hek_m25.py
examples/time_series/timeseriesmetadata_example.py
examples/acquiring_data/downloading_hmi.py
examples/acquiring_data/goes_xrs_example.py
examples/acquiring_data/querying_the_GOES_event_list.py
examples/acquiring_data/searching_vso.py
examples/acquiring_data/downloading_cutouts.py
examples/acquiring_data/fido_metadata_queries.py
examples/acquiring_data/search_cdaweb.py
examples/acquiring_data/downloading_lascoC2.py
```

Cadair:
The ones under "acquiring data" are probably always going to have to keep using Fido? other than that, we should try and cut out as many as we can.

dstansby:
I think #6197 has now fixed the most regular failed downloads here. A quick search of the examples now reveals:
```
% grep -r 'Fido.fetch' examples 
examples/units_and_coordinates/venus_transit.py:files = Fido.fetch(result)
examples/plotting/solar_cycle_example.py:f_noaa_indices = Fido.fetch(result)
examples/plotting/solar_cycle_example.py:f_noaa_predict = Fido.fetch(result)
examples/time_series/skip_timeseries_example.py:goes_files = Fido.fetch(goes)
examples/time_series/goes_hek_m25.py:files = Fido.fetch(results)
[...]

PR #6197, opened 2022-06-05
Use STEREO sample data in example gallery

dstansby:
Our doc build consistently fails on downloading STEREO data. I've added some new sample data here: https://github.com/sunpy/sample-data/pull/26, so hopefully this doesn't happen as often.

Part of fixing https://github.com/sunpy/sunpy/issues/3514 and https://github.com/sunpy/sunpy/issues/5632

### DIFF
--- a/changelog/6197.doc.rst
+++ b/changelog/6197.doc.rst
@@ -1,0 +1,2 @@
+Updated gallery examples that use STEREO data to use sample data instead
+of searching for and downloading data via Fido.
--- a/examples/map_transformations/reprojection_aia_euvi_mosaic.py
+++ b/examples/map_transformations/reprojection_aia_euvi_mosaic.py
@@ -3,12 +3,11 @@
 Creating a Full Sun Map with AIA and EUVI
 =========================================
 
-With SDO/AIA and STEREO/A and STEREO/B, it is now possible (given specific dates)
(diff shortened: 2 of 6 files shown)


nabobalis:
Hmm:

```
ERROR: Failed to download STEREO_B_195_Jun2012 from all mirrors,the file will not be available. [sunpy.data._sample]
ERROR: Failed to download STEREO_A_195_Jun2012 from all mirrors,the file will not be available. [sunpy.data._sample]
```
I wonder why its having issues.

dstansby:
Woops, I renamed the extensions in the sample data repo, and forgot to rename them here. Should be fixed now.

nabobalis (approved):

nabobalis:
The examples look good to me.

Cadair (commented):
Well I never submitted this comment.

I think we should make these keys all caps.

Cadair:
Also as we backported this we should probably have mentioned the new sample data files in the changelog, as now if people import our sample data in their scripts they are getting new files downloaded when they update to the next bugfix.

nabobalis:
> Also I would argue that adding sample data didn't really fall under the intention of the "docfix" label and this should have had two reviews, but whatever.

We added the sample data in the sample data repo, adding it to sunpy isn't a two person review.

Issue #6198, opened 2022-06-06
EUI jp2 images cannot be mapped


dgarciabriseno:
### Describe the bug

Attempting to parse an EUI jp2 file from https://www.sidc.be/EUI/data/releases/202204_release_5.0/L3/ results in an error. Traceback below.

### To Reproduce

Download https://www.sidc.be/EUI/data/releases/202204_release_5.0/L3/2022/02/07/solo_L3_eui-fsi174-image_20220207T050550186_V01.jp2
or any other jp2 file
Then run:

```python
from sunpy.map import Map
Map("solo_L3_eui-fsi174-image_20220207T050550186_V01.jp2")
```

### What happened?

```
>>> Map("solo_L3_eui-fsi174-image_20220207T050550186_V01.jp2")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.8/dist-packages/sunpy/map/map_factory.py", line 331, in __call__
    new_map = self._check_registered_widgets(data, meta, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/sunpy/map/map_factory.py", line 381, in _check_registered_widgets
    return WidgetType(data, meta, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/sunpy/map/sources/solo.py", line 37, in __init__
    self.plot_settings['cmap'] = self._get_cmap_name()
  File "/usr/local/lib/python3.8/dist-packages/sunpy/map/mapbase.py", line 1396, in _get_cmap_name
    str(int(self.wavelength.to('angstrom').value)))
  File "/usr/local/lib/python3.8/dist-packages/astropy/units/quantity.py", line 848, in to
    value = self._to_value(unit, equivalencies)
  File "/usr/local/lib/python3.8/dist-packages/astropy/units/quantity.py", line 802, in _to_value
    return self.unit.to(unit, self.view(np.ndarray),
  File "/usr/local/lib/python3.8/dist-packages/astropy/units/core.py", line 1135, in to
    return self._get_converter(Unit(other),
  File "/usr/local/lib/python3.8/dist-packages/astropy/units/core.py", line 1066, in _get_converter
    raise exc
  File "/usr/local/lib/python3.8/dist-packages/astropy/units/core.py", line 1051, in _get_converter
    return self._apply_equivalencies(
  File "/usr/local/lib/python3.8/dist-packages/astropy/units/core.py", line 1027, in _apply_equivalencies
[...]

PR #6199, opened 2022-06-06
Allow GenericMap to automatically set EUI map colormap

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/6198. GenericMap has a try... except... block when setting the colormap that prevents the error in that issue being raised.

Not sure the best way to test this without adding a jp2 file to the test data? Personally I'm happy this without a test, but also understand if another dev wants one.

### DIFF
--- a/sunpy/map/sources/solo.py
+++ b/sunpy/map/sources/solo.py
@@ -34,7 +34,6 @@ class EUIMap(GenericMap):
     def __init__(self, data, header, **kwargs):
         super().__init__(data, header, **kwargs)
         self._nickname = self.detector
-        self.plot_settings['cmap'] = self._get_cmap_name()
         self.plot_settings['norm'] = ImageNormalize(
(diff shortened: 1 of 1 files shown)


nabobalis (approved):

nabobalis:
I think its ok without a test.

wafels:
Does this fix will mean that Map users loading in EUI JP2 files will have to manually set the colormap?

nabobalis:
I don't think so as main class (GenericMap) already handles this setting the colormap.

dstansby:
> Does this fix will mean that Map users loading in EUI JP2 files will have to manually set the colormap?

It defaults to a grey colourmap if the wavelength attribute isn't in the metadata, so it works, but if you want the 'official' EUI colormap you have to set it yourself. I think this is all sunpy should do, and it's EUI's responsibility to populate their wavelength metadata.

Issue #5974, opened 2022-03-17
Using 'pip install "sunpy[all]"' does not include the packages needed to run the test suite.


wafels:
### Describe the bug

The SunPy installation documentation states

`pip install "sunpy[all]"`

will install a fully working SunPy installation.

However if you try to run the test suite later on, the following response is obtained

```
C:\Users\xxx>python
Python 3.9.10 (tags/v3.9.10:f2f3f53, Jan 17 2022, 15:14:21) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import sunpy
>>> sunpy.self_test()

Starting sunpy self test...
Checking for packages needed to run sunpy:
All required and optional sunpy dependencies are installed.
You do not have all the required dependencies installed to run the sunpy test suite.
['hypothesis', 'pytest-mock', 'pytest-astropy', 'pytest-mpl', 'pytest-doctestplus', 'jplephem']
If are using conda, you will want to run `conda install <package name>`
Otherwise you will want run `pip install "sunpy[all,tests]"`
>>>
```

This situation is confusing for a new user - why doesn't `all` also include the code to run the test suite - the very next section in the installation docs? It would be easy for a new user to think they had done something wrong. Although the user is told what to do next in order to run the tests this could be missed.

### To Reproduce

A fresh install is probably needed so that the test-related packages are not already present.

### Expected behavior

Installing using the `all` option should also include the packages required to perform the tests.

### Installation method

pip

dstansby:
> The problem is all conveys to the user that you're getting everything you need to have a fully working SunPy installation.

This is true

> Except it doesn't mean everything since it doesn't also include the packages you need to run the test suite in SunPy.

In my opinion, fully working installation != you have to be able to run the tests

Do you have any locations in the docs where we could make it clearer that you have to install the `test` extra to get test dependencies? Otherwise I think we should close this as "wontfix".

nabobalis:
I still think removing mentions of self_test in our user doc is something we want to do.

nabobalis:
We removed it from the installation documentation, so I think this is solved.

PR #6201, opened 2022-06-06
Move self_test() instructions from the installation to dev docs. 

dstansby:
I feel like in the distant past we have discussed doing this. See https://github.com/sunpy/sunpy/issues/5974 for the latest related issue where `self_test()` didn't work as expected for a user, but their install was fine. 

As I see it maintaining `self_test()` is a burden that we don't currently do, and having `self_test()` has minimal benefit. I know of no other python package that has this, and we have had many several negative `self_test()` failures from users in the past.

This should probably get a discussion at one of the weekly dev meetings.

### DIFF
--- a/docs/dev_guide/contents/tests.rst
+++ b/docs/dev_guide/contents/tests.rst
@@ -77,6 +77,38 @@ If a test errors, you can use ``pdb`` to create a debugging session at the momen
 
     $ pytest --pdb
 
+``self_test``
+=============
+
+Another method is to use :func:`sunpy.self_test`::
+
+    import sunpy
+    sunpy.self_test()
+
+You will see something like the following in your terminal::
+
+    Starting sunpy self test...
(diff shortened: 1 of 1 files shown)


nabobalis:
I agree, we should discuss this in the weekly meeting. 

I am in favour of removing it from the documentation, with removing it wholesale in future versions of sunpy.

In fairness, numpy does have their own version: https://numpy.org/doc/stable/reference/testing.html tho I don't support adding that to sunpy.

ayshih:
Astropy also has one: https://docs.astropy.org/en/latest/development/testguide.html#astropy-test

dstansby (commented):
I approve

dstansby:
@nabobalis feel free to approve and merge yourself, since this is my PR I can't do that, but I approve your changes

nabobalis (approved):

PR #6204, opened 2022-06-06
Fix mistake in plotting example

nabobalis:
Since we don't run this block, we never caught this error.

I also made the note part, a proper note.

### DIFF
--- a/docs/guide/plotting.rst
+++ b/docs/guide/plotting.rst
@@ -223,11 +223,16 @@ following ::
 when plotting on an `~astropy.visualization.wcsaxes.WCSAxes` axes, it will by
 default plot in pixel coordinates, you can override this behavior and plot in
 'world' coordinates by getting the transformation from the axes with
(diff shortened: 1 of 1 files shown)


wtbarnes (approved):

wtbarnes:
Our user guide needs some TLC 😅

nabobalis:
> Our user guide needs some TLC 😅

Yeah, it really does.

ayshih:
As a coding pattern, I prefer `.to_value(u.deg)` to `.to(u.deg).value` to avoid the creation of an intermediate object.

nabobalis:
> As a coding pattern, I prefer `.to_value(u.deg)` to `.to(u.deg).value` to avoid the creation of an intermediate object.

I had no idea that was possible. I will make that change.

nabobalis:
> As a coding pattern, I prefer `.to_value(u.deg)` to `.to(u.deg).value` to avoid the creation of an intermediate object.

Addressed by https://github.com/sunpy/sunpy/pull/6206

PR #6217, opened 2022-06-07
Overrride EUI waveunit

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/6198 - see https://issues.cosmos.esa.int/solarorbiterwiki/display/SOSP/Metadata+Definition+for+Solar+Orbiter+Science+Data for the ESA documentation that says this should be Angstrom.

I checked loading a .jp2 file with this PR, and it correctly loads the right colormap 🎉

### DIFF
--- a/sunpy/map/sources/solo.py
+++ b/sunpy/map/sources/solo.py
@@ -59,9 +59,10 @@ class EUIMap(GenericMap):
 
     @property
     def waveunit(self):
-        # EUI JP2000 files don't have this in their metadata,
-        # the FITS files are ok, so we check if it exists, and if not,
-        # the metadata spec wavelengths are always in Angstrom
+        # EUI JP2000 files do not have the WAVEUNIT key in the metadata.
+        # However, the FITS files do.
+        # The EUI metadata spec says the WAVELNTH key is always expressed
+        # in Angstroms so we assume this if the WAVEUNIT is missing.
         return super().waveunit or u.Angstrom
 
     @property
--- a/sunpy/map/sources/solo.py
+++ b/sunpy/map/sources/solo.py
@@ -59,9 +59,10 @@ class EUIMap(GenericMap):
 
     @property
     def waveunit(self):
-        # EUI maps don't have this in their metadata, but according to their
-        # metadata spec wavelengths are always in Angstrom
-        return u.Angstrom
+        # EUI JP2000 files don't have this in their metadata,
+        # the FITS files are ok, so we check if it exists, and if not,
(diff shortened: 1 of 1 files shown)


nabobalis:
Do we need to add back the colormap line?

dstansby:
> Do we need to add back the colormap line?

No

Cadair (changes requested):
If this is only an issue with jp2 files and not FITS files this probably isn't the correct solution.

sunpy/map/sources/solo.py
```diff
@@ -57,6 +57,13 @@ def date(self):
         timesys = self.meta.get('timesys')
         return parse_time(t, scale=timesys.lower())
 
+    @property
+    def waveunit(self):
+        # EUI JP2000 files don't have this in their metadata,
+        # the FITS files are ok, so we check if it exists, and if not,
+        # the metadata spec wavelengths are always in Angstrom
```
wtbarnes, line 64:
```suggestion
        # EUI JP2000 files do not have the WAVEUNIT key in the metadata.
        # However, the FITS files do.
        # The EUI metadata spec says the WAVELNTH key is always expressed
        # in Angstroms so we assume this if the WAVEUNIT is missing.
```

wtbarnes (approved):

Cadair (approved):
Happy with this now it defaults to extracting and falls back to Angstrom

bogdanni:
If this keyword is added in the next EUI data release, would this representation be OK?
```
WAVEUNIT= 'Angstrom' / Wavelength unit: angstrom
```
Note the capitalization guided by the FITS standard, where 'Angstrom' is the unit and 'angstrom' is the meaning.

bogdanni:
If this keyword is added in the next EUI data release, would this representation be OK?
```
WAVEUNIT= 'Angstrom' / Wavelength unit: angstrom
```
Note the capitalization guided by the FITS standard, where 'Angstrom' is the unit and 'angstrom' is the meaning.

nabobalis:
I would have to double check, but I am pretty sure that would work with `extract_waveunit`

Cadair:
@bogdanni that would work. To be clear though what you have at the moment is perfectly spec compliant and works great for the EUI FITS files, it's just that the JPEG2000 files made by helioviewer (I assume) strip out the FITS comment with the unit in.

bogdanni:
The Helioviewer XML metadata in JP2 files cannot transcribe the FITS header comments, so this is an explicit solution.

PR #6222, opened 2022-06-08
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


Cadair (approved):
Looks useful, thanks!

ConorMacBride:
Maybe I'm missing an option but it doesn't seem possible to prevent Dependabot from running on forks? (See https://github.com/dependabot/dependabot-core/issues/2804) Ideally we would configure, in this repo, Dependabot to not run on forks by default. However, there doesn't even seem to be an option in the fork repo to disable it from running. Do you know about this @naveensrinivasan?

naveensrinivasan:
> Maybe I'm missing an option but it doesn't seem possible to prevent Dependabot from running on forks? (See [dependabot/dependabot-core#2804](https://github.com/dependabot/dependabot-core/issues/2804)) Ideally we would configure, in this repo, Dependabot to not run on forks by default. However, there doesn't even seem to be an option in the fork repo to disable it from running. Do you know about this @naveensrinivasan?

AFAIK I don't think so.

Issue #6134, opened 2022-05-06
Improve generation of changelog for GitHub release


dstansby:
### Describe the feature

When creating a GitHub release the changelog needs to be generated from our own changelog. The steps for this are:

1. Run `pandoc --wrap=none -t markdown_strict CHANGELOG.rst`
2. Manually extract the relevant bit (the topmost section)
3. Paste into the GH release

I think 2. could be easily automated with a new small script in the `tools` directory.

Cadair:
We could make a GitHub action which does all of this on a new tag (not on main).

* Extract the top section of the changelog
* Convert to markdown
* Publish Github release with content and name equal to the tag.

ankitkhushwaha:
@nabobalis i think with #6240, closes this issue.

nabobalis:
The script is good but the goal would be also add that as an automation step to the release workflow.

PR #6240, opened 2022-06-09
Add script to generate github markdown changelog

dstansby:
Fixes ish https://github.com/sunpy/sunpy/issues/6134. A bit of a ridiculous script on many levels, but it works to only print the most recent release changelog.

### DIFF
--- a/tools/generate_github_changelog.py
+++ b/tools/generate_github_changelog.py
@@ -1,4 +1,10 @@
+"""
+Script to render to terminal, the changelog entry for the latest version in markdown so it can be copied and pasted for a GitHub release.
+
+It requires pandoc to be installed on your system.
+"""
+
 import subprocess
 
-res = subprocess.run(['pandoc', '--wrap=none', '-t', 'markdown_strict', 'CHANGELOG.rst'], capture_output=True)
+res = subprocess.run(['pandoc', '--wrap=none', '-t', 'markdown_strict', str(Path(__file__).parent.parent / "CHANGELOG.rst")], capture_output=True)
 print(res.stdout.decode('ascii').split('\n# ')[0])
--- a/tools/generate_github_changelog.py
+++ b/tools/generate_github_changelog.py
@@ -6,5 +6,5 @@ It requires pandoc to be installed on your system.
 import subprocess
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


nabobalis (approved):

ConorMacBride (commented):
Maybe we should add some code to get the path to Changelog relative to the script? Something like `Path(__file__).parent.parent / "CHANGELOG.rst"`

tools/generate_github_changelog.py
```diff
@@ -0,0 +1,4 @@
+import subprocess
+
+res = subprocess.run(['pandoc', '--wrap=none', '-t', 'markdown_strict', 'CHANGELOG.rst'], capture_output=True)
```
nabobalis, line 3:
```suggestion
res = subprocess.run(['pandoc', '--wrap=none', '-t', 'markdown_strict', str(Path(__file__).parent.parent / "CHANGELOG.rst")], capture_output=True)
```

tools/generate_github_changelog.py
```diff
@@ -0,0 +1,4 @@
+import subprocess
```
nabobalis, line 1:
```suggestion
"""
Script to render to terminal, the changelog entry for the latest version in markdown so it can be copied and pasted
for a GitHub release.

It requires pandoc to be installed on your system.
"""

import subprocess
```

nabobalis:
Locally pandoc is not converting the titles for me from `=====` to `#`, I assume it's a me problem tho.

nabobalis (approved):

ConorMacBride:
> Locally pandoc is not converting the titles for me from `=====` to `#`, I assume it's a me problem tho.

That may still be valid markdown https://www.markdownguide.org/basic-syntax/#alternate-syntax

nabobalis:
> > Locally pandoc is not converting the titles for me from `=====` to `#`, I assume it's a me problem tho.
> 
> That may still be valid markdown [markdownguide.org/basic-syntax/#alternate-syntax](https://www.markdownguide.org/basic-syntax/#alternate-syntax)

Wait, those were/are valid headings?

nabobalis:
If I do `gfm` as the output format, it does convert them for me.

ConorMacBride:
Probably best to switch it to gfm since we want to use it in GitHub primarily

nabobalis:
I think I will merge this now?

PR #6242, opened 2022-06-10
Use sample data for Venus exapmle

dstansby:
The Venus gallery example regularly fails on our doc gallery build. Instead of making the sample data bigger, I've used pooch (https://www.fatiando.org/pooch/latest/) to set up a new sample data registry for data that is only used in the example gallery.

This does add an extra dependency that users will need for running the gallery example, so that needs documenting properley if we go with this. Thoughts/opinions on whether this is a sensible approach very welcome.


ayshih:
> The Venus gallery example regularly fails on our doc gallery build.

It does?  More so than any other Fido grab of AIA data?

dstansby:
> It does? More so than any other Fido grab of AIA data?

yep, it seems to. It failed today on both 4.0 and main, and I think has been failing for a few days now. I don't know what else in the gallery grabs AIA data though? The cutout example has also been failing as regularly as the Venus example.

nabobalis:
We should also the VSO about it.

nabobalis:
**CHECK THE DOC BUILD IS OK BEFORE MERGE**

PR #6243, opened 2022-06-10
Support upcoming parfive 2.0 release

Cadair:
Currently just testing out the rc with our CI.

### DIFF
--- a/sunpy/util/parfive_helpers.py
+++ b/sunpy/util/parfive_helpers.py
@@ -43,4 +43,10 @@ else:
         def __init__(self, *args, **kwargs):
             if "config" not in kwargs:
                 kwargs["config"] = config
+
+            # This is to make our sample data download code work on 1.x and 2.x
+            # when we depend on 2+ we should remove this.
+            headers = kwargs.pop("headers", {})
+            kwargs["config"].headers = {**headers, **kwargs["config"].headers}
+
             super().__init__(*args, **kwargs)
--- a/sunpy/util/parfive_helpers.py
+++ b/sunpy/util/parfive_helpers.py
@@ -4,27 +4,43 @@ from functools import wraps
 
 import aiohttp
 import parfive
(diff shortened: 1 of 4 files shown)


nabobalis:
```
E           parfive.utils.ParfiveFutureWarning: The headers keyword argument to Downloader is deprecated. Please instead pass a SessionConfig object to the config keyword argument.
```

Cadair:
yeah well I should have seen that coming...

nabobalis:
That ignore didnt work for some reason

Cadair:
I think we should merge this PR to main and 4.0, then get a bugfix release of 4.0 out. Then once parfive 2.0 has made it into the wild we can bump the minimum version on main to 2.0 if we want to.

nabobalis (approved):

sunpy/util/parfive_helpers.py
```diff
@@ -4,27 +4,49 @@
...
+    class Downloader(parfive.Downloader):
+        @wraps(parfive.Downloader.__init__)
+        def __init__(self, *args, **kwargs):
+            if "config" not in kwargs:
+                kwargs["config"] = config
+
+            # This is to make our sample data download code work on 1.x and 2.x
+            # when we depend on 2+ we should remove this.
+            headers = kwargs.pop("headers", {})
+            kwargs["config"].headers = {**headers, **kwargs["config"].headers}
```
ConorMacBride, line 50: Are the two dictionaries here the right way around?

Cadair: It should be overriding the `headers` argument with the one in `config`?

Cadair: I think you are right this should be the other way around, not that it matters in practise.

```suggestion
            kwargs["config"].headers = {**kwargs["config"].headers, **headers}
```

ConorMacBride (approved):

PR #6247, opened 2022-06-11
GOES: Fixes an issue where timestamps containing leap seconds could not be converted to Python datetimes

mariusgiger:
## PR Description

- Fixes an issue where timestamps containing leap seconds could not be converted to Python datetimes (XY is within a leap second but datetime does not support leap seconds) #5422
- Exposes GOES quality flags as it is not possible to decide whether a given value is faulty otherwise (refer to https://satdat.ngdc.noaa.gov/sem/goes/data/science/xrs/GOES_13-15_XRS_Science-Quality_Data_Readme.pdf)

Example of faulty values (See X class values):
download

### DIFF
--- a/sunpy/timeseries/sources/goes.py
+++ b/sunpy/timeseries/sources/goes.py
@@ -263,6 +263,8 @@ class XRSTimeSeries(GenericTimeSeries):
             if "a_flux" in d.variables:
                 xrsa = np.array(d["a_flux"])
                 xrsb = np.array(d["b_flux"])
+                quality_xrsb = np.array(d['b_flags'])
+                quality_xrsa = np.array(d['a_flags'])
 
                 start_time_str = d["time"].attrs["units"]
                 if not isinstance(start_time_str, str):
@@ -274,6 +276,9 @@ class XRSTimeSeries(GenericTimeSeries):
             elif "xrsa_flux" in d.variables:
                 xrsa = np.array(d["xrsa_flux"])
                 xrsb = np.array(d["xrsb_flux"])
+                quality_xrsb = np.array(d['xrsa_flags'])
+                quality_xrsa = np.array(d['xrsb_flags'])
+
                 start_time_str = d["time"].attrs["units"]
                 if not isinstance(start_time_str, str):
                     # For h5netcdf<0.14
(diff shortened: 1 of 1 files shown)


mariusgiger:
Likely conflicting with https://github.com/sunpy/sunpy/pull/5834

hayesla:
Hi @mariusgiger - I haven't properly looked into this yet, but was this not fixed by https://github.com/sunpy/sunpy/pull/5915
 or is this another issue?

mariusgiger:
> Hi @mariusgiger - I haven't properly looked into this yet, but was this not fixed by #5915 or is this another issue?

Hi @hayesla, thanks for writing back to me. Unfortunately, this did not fix it. The issue can be reproduced with the following example using Sunpy v4.0.1:

```
from sunpy.net import Fido

t_start = "2015-06-30 00:00"
t_end = "2015-07-01 00:00:00"
search_result = Fido.search(
df = goes_ts.to_dataframe()
print(df)
```

which will result in 

```
ValueError: Time (array(2015, dtype=int32), array(6, dtype=int32), array(30, dtype=int32), array(23, dtype=int32), array(59, dtype=int32), array(60, dtype=int32), array(965000, dtype=int32)) is within a leap second but datetime does not support leap seconds
```

I am not well versed in Astropy Times, so there might be a better way to drop/convert the leap seconds. 

Cheers,
Marius

dstansby (changes requested):
Thanks for opening this PR. Sadly I don't think this is the right way to go - we shouldn't be dropping data from the file before returning it to the user! Unfortunately this means storing our data using not pandas, which there are plans to do, but it's a bit complicated which is why we haven't done it yet (see https://github.com/sunpy/sunpy/pull/5834 for work in progress)

mariusgiger:
@dstansby thanks for the reply! I agree, however it is currently not working - so depending on when the refactoring is done, you might still want to consider to merge this. 

Regarding the second point, it would be great to have access to the quality flags in order to filter the timeseries by quality as there are quite a few corrupt data points.

mariusgiger:
Ok thanks for the feedback - I have moved access to the quality flags to another PR #6260. @dstansby it would be great if `sunpy.timeseries.GenericTimeSeries.to_dataframe` would not fail in the future if the underlying Astropy TimeSeries contained leap seconds (for ease of use). Do you think this is achievable? Maybe a filtering mechanism as was proposed here could be used.

nabobalis:
The issue here is not `to_dataframe`, it fails in the loading of the file. So we will need to patch that first. 

Then for `to_dataframe`, I think we need to add keyword that will allow a user to get dataframe out without leap seconds.

nabobalis:
> @nabobalis yes sure this is the case currently, however in the future the file loading will no longer involve pandas if I got it correctly.

That is true, hopefully for 4.1
 
> Such a flag would be great 👍

We will make sure add such a flag for the refactor of timeseries.

nabobalis:
I have what I hope is a solution to your problem @mariusgiger here: https://github.com/sunpy/sunpy/pull/6262 
Your example code now works.

It will take a few days for it to be reviewed and released in a bugfix release.

PR #6250, opened 2022-06-13
update dev hashes

nabobalis:
Looks like mpl dev changes have moved titles to the left slightly.


ConorMacBride (approved):
Dashes are now longer. Not sure where this change has happened, but it looks fine. Maybe the `--` should now be removed from [the GOES time series legend](https://github.com/sunpy/sunpy/blob/7ecdcba2b8e451d5a2800c820752ba858ab84deb/sunpy/timeseries/sources/goes.py#L94) since it's now longer [(see figure test)](https://output.circle-artifacts.com/output/job/07f152d0-3643-4be9-9217-8563c5645bd8/artifacts/0/.tmp/py39-figure-devdeps/figure_test_images/fig_comparison.html?q=test_goes_peek)? There's also dates in other figure tests which don't have the longer dash, [this test for example](https://output.circle-artifacts.com/output/job/07f152d0-3643-4be9-9217-8563c5645bd8/artifacts/0/.tmp/py39-figure-devdeps/figure_test_images/fig_comparison.html?q=test_map_sequence_plot_custom_cmap_norm)?

dstansby (changes requested):
I reckon this *might* be a Matplotlib bug (I think it's at least a Matplotlib change). I'll block until I've done a bit more investigation on this (which I'm doing now)

dstansby:
Yep, pretty sure this is a bug: https://github.com/matplotlib/matplotlib/issues/23268 - I think we should hold tight and see if a fix comes in from Matplotlib before changing these hashes.

nabobalis:
Since this is a bug, im turning off this PR.

Issue #5969, opened 2022-03-16
Add options to `GenericMap.draw_grid()` for emphasizing the prime meridian and/or the equator


ayshih:
In addition to the usual dotted heliographic grid lines that are drawn by `GenericMap.draw_grid()`, let's add options to highlight (e.g., as solid lines) the prime meridian and/or the equator.  This would be particularly useful when comparing observations between two observer locations with a significant difference in heliographic longitude.

dstansby:
Instead of adding more arguments to `draw_grid()`, how about `draw_equator()` and `draw_meridian()` methods? These could then be used independently of `draw_grid()`, and avoids complicating the signature of `draw_grid()`.

tanishy7777:
is this issue open to work on?

nabobalis:
Looks like a  PR solved this but it was never closed

PR #6251, opened 2022-06-14
Add draw_equator and draw_prime_meridian to visualization

mwhv2:
This PR adds two plotting methods for maps, draw_equator and draw_prime_meridian, that resolves Issue #5969. I have also added two corresponding figure tests for each function. 

Attached are a few pictures showing examples for a couple maps and a submap. 
example1
example2
example3

### DIFF
--- a/sunpy/visualization/draw.py
+++ b/sunpy/visualization/draw.py
@@ -63,7 +63,8 @@ def equator(axes, *, rsun: u.m = R_sun, resolution=500, **kwargs):
     return visible, hidden
 
 
-def prime_meridian(axes, rsun: u.m = R_sun, resolution=500, **kwargs):
+@u.quantity_input
+def prime_meridian(axes, *, rsun: u.m = R_sun, resolution=500, **kwargs):
     """
     Draws the solar prime meridian (zero Carrington longitude) as seen by the
     axes observer. Hidden parts are drawn as a dotted line.


Cadair (changes requested):
I haven't looked over this yet, but I am going to put a blocker on it because I want us to think about the proliferation of `draw_` methods on map, we are clearly in a situation where we have a never ending list of helpers we could write and we should think about the shape of the api.

mwhv2:
No worries! I understand the concerns.

nabobalis:
Hi @mwhv2, I think we decided at the weekly meeting that these methods should be moved into the visualization subpackage instead as they only need a `ax` to draw instead of a sunpy Map. 

@Cadair or @ayshih will correct me if I am wrong.

mwhv2:
Okay, should these methods be included alongside current functions within the visualization subpackge (e.g., peek_show and axis_labels_from_ctype)? Or in a new/standalone file within visualization?

nabobalis:
> Okay, should these methods be included alongside current functions within the visualization subpackge (e.g., peek_show and axis_labels_from_ctype)? Or in a new/standalone file within visualization?

Good question, I feel like a new file, `draw_limb` has its own file for example.

nabobalis:
> To be clear, is the preference that these functions (within visualizations) have no dependence on a Map?

Yes, the goal is to avoid adding more items to the API for map and draw a separation between the Mpa class and its draw methods.

> Currently, a new module with these functions (as they are written) will not require a "import sunpy.map" statement, but both functions do require a Map as the primary argument. Is the intent that these functions only require axes and observer arguments (like `draw_limb`)?

The intent is to avoid needing to use a Map to draw these as I think a WCS axes and maybe the observer is enough to draw these so in fact they could be more general than being just attached to the Map class.

nabobalis:
Can this rebased onto current main please?

nabobalis:
> Okay, I have rebased onto main. Sorry for the mess (I am not entirely certain what happened in regards to timeseries files changing).

That is no problem!

> Is there a better practice for making cleaner PRs (e.g., creating a new branch for a PR)?

Typically yes, it is common to make a new branch per PR to isolate changes while keeping main up to date with the remote (in this case the sunpy repo) version.

sunpy/visualization/tests/test_draw.py
```diff
@@ -0,0 +1,20 @@
...
+Test draw
+"""
+import pytest
+
+from sunpy.tests.helpers import figure_test
+from sunpy.visualization import draw
+from sunpy.map.tests.conftest import aia171_test_map
+
+
+@figure_test
```
nabobalis, line 11: You will need to add hashes to the figure hashes file for these to work correctly.

If you see https://app.circleci.com/pipelines/github/sunpy/sunpy/18331/workflows/27b6a082-24ed-4fc2-ad0e-7e06599b0594/jobs/79348/parallel-runs/0/steps/0-107

The artefacts with the hashes are there too.

sunpy/visualization/tests/test_draw.py
```diff
@@ -0,0 +1,20 @@
+"""
+Test draw
+"""
```
nabobalis, line 3:
```suggestion
```

sunpy/visualization/draw.py
```diff
@@ -0,0 +1,106 @@
...
+    rsun = axes_frame.rsun
+    reference_dist = np.sqrt(observer.radius**2 - rsun**2)
+    is_visible = lat0.spherical.distance <= reference_dist
+    vis = np.where(is_visible == True)
+    visible = lat0[vis]
+
+    equator = Line2D(visible.Tx.deg, visible.Ty.deg,
+                     **c_kw)
+    axes.add_line(equator)
+    return equator
```
nabobalis, line 64: Do these functions need to return anything?

sunpy/visualization/draw.py
```diff
@@ -0,0 +1,106 @@
+"""
+This module provides functions that draw on Map plots.
```
nabobalis, line 2:
```suggestion
This module provides functions that draw on WCSAxis.
```
I think this is more correct?

changelog/6251.feature.rst
```diff
@@ -0,0 +1 @@
+Add draw module which includes draw_equator and draw_prime_meridian functions to `sunpy.visualization`.
```
nabobalis, line 1:
```suggestion
Created the `sunpy.visualization.draw` module which includes `.draw_equator` and `.draw_prime_meridian` functions.
```
I forget if this will link correctly.

sunpy/visualization/draw.py
```diff
@@ -0,0 +1,106 @@
...
+        The number of points used to represent the equator.
+
+    Examples
+    --------
+    >>> import sunpy.map
+    >>> from sunpy.visualization import draw
+    >>> import sunpy.data.sample   # doctest: +REMOTE_DATA
+    >>> aia = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)   # doctest: +REMOTE_DATA
+    >>> aia.plot()   # doctest: +SKIP
+    >>> draw.draw_equator(aia._check_axes(None))   # doctest: +SKIP
```
nabobalis, line 34: This is how someone has to get the axes from the plot method?

sunpy/visualization/draw.py
ayshih: I think it is consistent practice to do so

nabobalis: Thanks!

changelog/6251.feature.rst
ayshih:
```suggestion
Created the `sunpy.visualization.draw` module which includes :func:`~sunpy.visualization.draw_equator` and :func:`~sunpy.visualization.draw_prime_meridian` functions.
```

sunpy/visualization/draw.py
ayshih: Well, they can, but they shouldn't (because it is a private method).  If we have some reason to avoid showing the explicit creation of axes in the docstring example, the best option is to call `plt.gca()` to retrieve the current axes.

sunpy/visualization/draw.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    if not wcsaxes_compat.is_wcsaxes(axes):
```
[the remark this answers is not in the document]
ayshih: Well, as functions rather than Map methods, it's (slightly) more likely that these functions would be called on axes not created by a Map method, so I think it's fine to keep a check for WCSAxes.

sunpy/visualization/draw.py
```diff
@@ -0,0 +1,106 @@
...
+    c_kw = {'linewidth': 1,
+            'color': 'white',
+            'zorder': 100}
+    c_kw.update(kwargs)
+
+    transform = axes.get_transform(axes_frame)
+    c_kw.setdefault('transform', transform)
+
+    lat = 0*u.deg
+    lat0 = SkyCoord(np.linspace(-90, 90, resolution) * u.deg,
```
ayshih, line 50:
```suggestion
    lat0 = SkyCoord(np.linspace(-180, 180, resolution) * u.deg,
```
You need to construct the whole equator because you don't know where the observer is.

sunpy/visualization/draw.py
```diff
@@ -0,0 +1,106 @@
...
+
+from sunpy.coordinates import HeliographicStonyhurst
+from sunpy.visualization import wcsaxes_compat
+
+__all__ = ["draw_equator", "draw_prime_meridian"]
+
+
+def draw_equator(axes, resolution=300, **kwargs):
+    """
+    Draws the equator as seen by the axes observer.
```
ayshih, line 18:
```suggestion
    Draws the solar equator as seen by the axes observer.
```

sunpy/visualization/tests/test_draw.py
```diff
@@ -0,0 +1,20 @@
...
+
+from sunpy.tests.helpers import figure_test
+from sunpy.visualization import draw
+from sunpy.map.tests.conftest import aia171_test_map
+
+
+@figure_test
+def test_draw_equator_aia171(aia171_test_map):
+    aia171_test_map.plot()
+    draw.draw_equator(aia171_test_map._check_axes(None))
```
ayshih, line 14: Definitely explicitly create the axes here and below

sunpy/visualization/draw.py
```diff
@@ -0,0 +1,106 @@
...
+                    frame=HeliographicStonyhurst)
+    lat0 = lat0.transform_to(axes_frame)
+
+    rsun = axes_frame.rsun
+    reference_dist = np.sqrt(observer.radius**2 - rsun**2)
+    is_visible = lat0.spherical.distance <= reference_dist
+    vis = np.where(is_visible == True)
+    visible = lat0[vis]
+
+    equator = Line2D(visible.Tx.deg, visible.Ty.deg,
```
ayshih, line 61: The equator should probably be a `Polygon` instead so that it will close the endpoints together.  See `draw_limb()`.

ayshih (changes requested):
Thanks for the PR!

Additional comments:
- `draw_prime_meridian()` is not currently plotting the "prime" meridian, but rather the central meridian as observed by Earth.  I think the proper definition of the "prime" meridian is where the Carrington longitude is zero, rather than the Stonyhurst longitude.
- As with `draw_limb()`, these functions should probably accept the solar radius as an option argument, which should be provided to the `HeliographicStonyhurst` coordinates that you create
- As with `draw_limb()`, these functions should check for whether the frame has an observer attribute before calculating visibility.  These functions ought to be usable on heliographic maps, which have no observer.
- On that subject, there should be a figure test that uses these on a heliographic map
- As with `sunpy.visualization.limb.draw_limb()`, we should probably import these into the `sunpy.visualization` namespace
- There is a bit too much code duplication between these two functions.  The shared code should be moved to a private function.
- Eventually, this module should be combined with `limb.py`, and there would be further opportunity for sharing code
- It can be outside of this PR, but we will want a gallery example showcasing these functions

sunpy/visualization/draw.py
```diff
@@ -0,0 +1,106 @@
...
+    >>> import sunpy.map
+    >>> from sunpy.visualization import draw
+    >>> import sunpy.data.sample   # doctest: +REMOTE_DATA
+    >>> aia = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)   # doctest: +REMOTE_DATA
+    >>> aia.plot()   # doctest: +SKIP
+    >>> draw.draw_equator(aia._check_axes(None))   # doctest: +SKIP
+    """
+    if not wcsaxes_compat.is_wcsaxes(axes):
+        raise ValueError('axes must be a WCSAxes')
+    axes_frame = axes._transform_pixel2world.frame_out
```
Cadair, line 38: We can avoid using this private attribute by doing the following:
```suggestion
    from astropy.visualization.wcsaxes.wcsapi import wcsapi_to_celestial_frame
    axes_frame = wcsapi_to_celestial_frame(axes.wcs)
```

sunpy/visualization/draw.py
```diff
@@ -0,0 +1,106 @@
...
+    observer = axes_frame.observer
+
+    c_kw = {'linewidth': 1,
+            'color': 'white',
+            'zorder': 100}
+    c_kw.update(kwargs)
+
+    transform = axes.get_transform(axes_frame)
```
Cadair, line 46: This is synonymous with this for what's it's worth.
```python
    transform = axes.get_transform("world")
```

mwhv2:
I want to double check my understanding in regards to observers before I make too many more edits. As I follow `sunpy.visualization.limb.draw_limb`, this function cannot work without an observer. In the case of a heliographic Map, the `sunpy.map.GenericMap.observer_coordinate` method automatically assigns an Earth-based observer to the Map that is passed into the `sunpy.map.GenericMap.draw_limb` wrapper before plotting the solar limb.

Currently, the corrected prime_meridian function (using a Carrington longitude of 0) requires an observer. If the above process is correct, should I require an observer as an argument for the prime_meridian function? Or, have, for example, try and except statements check for an observer and assign an Earth-based observer if one is not given?

ayshih:
> As I follow sunpy.visualization.limb.draw_limb, this function cannot work without an observer.

To clarify, there are can be two observers that are relevant.  The observer passed to `draw_limb()` defines the apparent limb in 3D space.  The observer – if any – of the WCSAxes defines how that limb should be plotted, including indicating whether part of the limb for the first observer is hidden from the second observer.  These observers need not be the same.  See the logical branching in `draw_limb()` depending on whether the WCSAxes has an observer:
https://github.com/sunpy/sunpy/blob/8a241fd16d7c9bbc0839436b8a8d9a6f7e1ba8a8/sunpy/visualization/limb.py#L103-L113

> In the case of a heliographic Map, the sunpy.map.GenericMap.observer_coordinate method automatically assigns an Earth-based observer to the Map that is passed into the sunpy.map.GenericMap.draw_limb wrapper before plotting the solar limb.

This is true, but is not the use case that I meant.  If you look at [this example](https://docs.sunpy.org/en/stable/generated/gallery/plotting/limb_plotting.html), there is a heliographic Map, and then `draw_limb()` is called for two different observers.  `draw_equator()` should work on both Stonyhurst and Carrington heliographic maps, and draw the entire 360 degrees of longitude.  If it's an Carrington heliographic map, the `SkyCoord` for the equator needs to be created in that same coordinate frame to make sure that the observer is properly taken into account.

> Currently, the corrected prime_meridian function (using a Carrington longitude of 0) requires an observer. If the above process is correct, should I require an observer as an argument for the prime_meridian function? Or, have, for example, try and except statements check for an observer and assign an Earth-based observer if one is not given?

Upon further thought, `draw_prime_meridian()` should require an observer on the WCSAxes, and should neither accept an observer as input nor assume an observer.  If the WCSAxes does not have an observer – e.g., a Stonyhurst heliographic map – an informative error message should be raised.

mwhv2:
For the purpose of testing prime meridian, how can I add an observer to the axes of a heliographic map?

mwhv2:
In the above update, I have borrowed heavily from `sunpy.visualization.limb` so that these modules can be easily combined in a future PR (which I would be happy to do). Both equator and prime_meridian now share a plotting function which draws both as `matplotlib.patches.Polygon` objects. Like `draw_limb`, the hidden part of both equator and prime_meridian is shown as a dotted line. 

I have also renamed the functions so they can be called more easily (i.e., `draw.equator` and `draw.prime_meridian`). There is an added test for a heliographic map. I will update the fake figure hashes after CircleCi has run.

nabobalis (commented):
The code looks good to me but I lack the expertise on coordinates to actually approve this PR.

changelog/6251.feature.rst
```diff
@@ -0,0 +1,3 @@
+Created the `sunpy.visualization.draw` module which includes 
+:func:`~sunpy.visualization.equator` and 
+:func:`~sunpy.visualization.prime_meridian` functions.
```
ayshih, line 3:
```suggestion
:func:`~sunpy.visualization.draw.equator` and 
:func:`~sunpy.visualization.draw.prime_meridian` functions.
```

sunpy/visualization/draw.py
```diff
@@ -0,0 +1,163 @@
...
+                    np.ones(resolution) * lat, radius=rsun,
+                    frame=HeliographicStonyhurst(obstime=axes_frame.obstime))
+    visible, hidden = _plot_vertices(lat0, axes, axes_frame, rsun,
+                                     **kwargs)
+    return visible, hidden
+
+
+def prime_meridian(axes, rsun: u.m = R_sun, resolution=500, **kwargs):
+    """
+    Draws the solar prime meridian as seen by the axes observer.
```
ayshih, line 67:
```suggestion
    Draws the solar prime meridian (zero Carrington longitude) as seen by the axes observer.
```

sunpy/visualization/draw.py
```diff
@@ -0,0 +1,163 @@
...
+    lon0 = SkyCoord(np.ones(resolution) * lon,
+                    np.linspace(-90, 90, resolution) * u.deg, radius=rsun,
+                    frame=HeliographicCarrington(observer=observer,
+                                                 obstime=axes_frame.obstime))
+    visible, hidden = _plot_vertices(lon0, axes, axes_frame, rsun,
+                                     pm_dis=True, **kwargs)
+    return visible, hidden
+
+
+def _plot_vertices(coord, axes, frame, rsun, pm_dis=False, **kwargs):
```
ayshih, line 105:
```suggestion
def _plot_vertices(coord, axes, frame, rsun, close_path=True, **kwargs):
```
Rename this keyword to be more general

sunpy/visualization/draw.py
```diff
@@ -0,0 +1,163 @@
...
+        # If the axes has no observer, the entire limb is considered visible
+        is_visible = np.ones_like(coord.spherical.distance, bool, subok=False)
+
+    # Identify discontinuities. Uses the same approach as
+    # astropy.visualization.wcsaxes.grid_paths.get_lon_lat_path()
+    step = np.sqrt((vertices[1:, 0] - vertices[:-1, 0]) ** 2 +
+                   (vertices[1:, 1] - vertices[:-1, 1]) ** 2)
+    continuous = np.concatenate([[True, True], step[1:] < 100 * step[:-1]])
+    # The prime meridian is not continuous, so set the endpoints equal to False
+    if pm_dis is True:
```
ayshih, line 139:
```suggestion
    if not close_path is True:
```

sunpy/visualization/draw.py
```diff
@@ -0,0 +1,163 @@
...
+        raise AttributeError('an observer is required on the WCSAxes')
+    observer = axes_frame.observer
+
+    lon = 0*u.deg
+    lon0 = SkyCoord(np.ones(resolution) * lon,
+                    np.linspace(-90, 90, resolution) * u.deg, radius=rsun,
+                    frame=HeliographicCarrington(observer=observer,
+                                                 obstime=axes_frame.obstime))
+    visible, hidden = _plot_vertices(lon0, axes, axes_frame, rsun,
+                                     pm_dis=True, **kwargs)
```
ayshih, line 101:
```suggestion
                                     close_path=False, **kwargs)
```

mwhv2:
pre-commit.ci autofix

sunpy/visualization/tests/test_draw.py
```diff
@@ -0,0 +1,39 @@
+import matplotlib.pyplot as plt
+import pytest
+
+from astropy.wcs import WCS
+
+from sunpy.map.tests.conftest import aia171_test_map, heliographic_test_map  # NoQA
```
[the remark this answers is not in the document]
ConorMacBride: Maybe those fixtures are not being loaded as they are defined in the conftest for map and not visualisation. Although not sure why one would work.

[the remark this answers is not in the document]
mwhv2: Yeah, the available fixtures listed in the [error](https://app.circleci.com/pipelines/github/sunpy/sunpy/18639/workflows/4287f8c5-cf32-4fc1-804a-c6aa36953448/jobs/79983?invite=true#step-107-355) seem to include everything in the main conftest.py file. Should I just remove the aia171_test_map and leave the heliographic_test_map import included for now?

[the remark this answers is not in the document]
nabobalis: Can you just re-define them in this test file. I think we can worry about that later on.

sunpy/visualization/tests/test_draw.py
```diff
@@ -0,0 +1,57 @@
...
+    fig = plt.figure()
+    axes = fig.add_subplot(projection=heliographic_test_map)
+    heliographic_test_map.plot()
+    draw.equator(axes, color="blue")
+    draw.prime_meridian(axes, color="red")
+
+
+def test_prime_meridian_error():
+    axes = plt.subplot(projection=WCS())
+    with pytest.raises(AttributeError):
```
ayshih, line 56:
```suggestion
    with pytest.raises(ValueError, match="does not have an observer"):
```

wtbarnes:
Once this is rebased, you'll see there is a 4.1 What's New file. There should be an entry for these two new draw methods.

Cadair (approved):
Haven't gone over it in great detail, but it looks good to me.

ayshih (approved):
Code looks good to go

It sounds like @wtbarnes wants to block this PR until a "What's New?" entry is added, but I'd be fine with deferring that entry until the subsequent PR that moves `draw_limb()` into the `sunpy.visualization.draw` module.  I see that consolidation as a blocker for 4.1.

mwhv2:
Yes, once I have identified all the usages of `sunpy.visualization.limb` I will combine the two modules and make the necessary updates (including the What's New mention). I also plan to include an example for the example gallery showcasing the draw methods.

nabobalis:
> Yes, once I have identified all the usages of `sunpy.visualization.limb` I will combine the two modules and make the necessary updates (including the What's New mention). I also plan to include an example for the example gallery showcasing the draw methods.

Thanks! I will merge it then.

mwhv2:
Thank you @ayshih, @Cadair, and @nabobalis for all the help on this PR!

PR #6260, opened 2022-06-16
Exposes GOES quality flags

### LINKED ISSUE #6259: Expose GOES quality flags
### Describe the feature

Currently, it is not possible to differentiate between corrupt and normal values when using the GOES `XRSTimeSeries`. As there are quite a few corrupt values which can be seen in the following image (see number of X-class values), it would be great to expose the flags indicating whether a given value is corrupt or not as well.
173183771-b030c0e6-6897-4752-a341-e17b3e054658
(also refer to https://satdat.ngdc.noaa.gov/sem/goes/data/science/xrs/GOES_13-15_XRS_Science-Quality_Data_Readme.pdf)

### Proposed solution

Exposes GOES quality flags (xrsa_quality, xrsb_quality) as part of the `XRSTimeSeries` in order to filter for invalid values. An example could look as follows:

```
from sunpy.net import Fido

t_start = "2014-01-01 00:00"
t_end = "2014-01-31 00:00"
search_result = Fido.search(
df = goes_ts.to_dataframe()
df_clean = df[(df["xrsb_quality"] == 0)]
```

A PR will follow

mariusgiger:
## PR Description

Resolves #6259 by exposing GOES quality flags to the user

TODO how to extract quality flags from the FITS header?

### DIFF
--- a/examples/acquiring_data/goes_xrs_example.py
+++ b/examples/acquiring_data/goes_xrs_example.py
@@ -69,7 +69,7 @@ goes_15.peek()
 
 #############################################################
 # The resulting `~sunpy.timeseries.TimeSeries` can be filtered by GOES quality flags. For more information
-# refer to the `GOES Data Guide <https://satdat.ngdc.noaa.gov/sem/goes/data/science/xrs/GOES_13-15_XRS_Science-Quality_Data_Readme.pdf>`.
+# refer to the `GOES Data Guide <https://satdat.ngdc.noaa.gov/sem/goes/data/science/xrs/GOES_13-15_XRS_Science-Quality_Data_Readme.pdf>`__.
 
 df = goes_15.to_dataframe()
 df = df[(df["xrsa_quality"] == 0) & (df["xrsb_quality"] == 0)]
--- a/examples/acquiring_data/goes_xrs_example.py
+++ b/examples/acquiring_data/goes_xrs_example.py
@@ -73,7 +73,7 @@ goes_15.peek()
 
 df = goes_15.to_dataframe()
 df = df[(df["xrsa_quality"] == 0) & (df["xrsb_quality"] == 0)]
-goes_15.data = df
+goes_15 = ts.TimeSeries(df, goes_15.meta, goes_15.units)
 
 ###############################################################
 # We can also pull out the individual GOES chanels and plot. The 0.5-4 angstrom
--- a/examples/acquiring_data/goes_xrs_example.py
+++ b/examples/acquiring_data/goes_xrs_example.py
@@ -67,6 +67,14 @@ file_goes15 = Fido.fetch(result_goes15)
 goes_15 = ts.TimeSeries(file_goes15)
 goes_15.peek()
 
+#############################################################
+# The resulting `~sunpy.timeseries.TimeSeries` can be filtered by GOES quality flags. For more information
+# refer to the `GOES Data Guide <https://satdat.ngdc.noaa.gov/sem/goes/data/science/xrs/GOES_13-15_XRS_Science-Quality_Data_Readme.pdf>`.
+
+df = goes_15.to_dataframe()
+df = df[(df["xrsa_quality"] == 0) & (df["xrsb_quality"] == 0)]
+goes_15.data = df
+
 ###############################################################
 # We can also pull out the individual GOES chanels and plot. The 0.5-4 angstrom
 # channel is known as the "xrsa" channel and the 1-8 angstrom channel is known
--- a/sunpy/timeseries/sources/goes.py
+++ b/sunpy/timeseries/sources/goes.py
@@ -218,6 +218,7 @@ class XRSTimeSeries(GenericTimeSeries):
                 raise ValueError("Date not recognized")
             xrsb = hdulist[2].data['FLUX'][0][:, 0]
             xrsa = hdulist[2].data['FLUX'][0][:, 1]
+            # TODO how to extract quality flags from HDU?
             seconds_from_start = hdulist[2].data['TIME'][0]
         elif 1 <= len(hdulist) <= 3:
             start_time = parse_time(header['TIMEZERO'], format='utime')
@@ -263,6 +264,8 @@ class XRSTimeSeries(GenericTimeSeries):
             if "a_flux" in d.variables:
                 xrsa = np.array(d["a_flux"])
                 xrsb = np.array(d["b_flux"])
+                xrsa_quality = np.array(d['a_flags'])
+                xrsb_quality = np.array(d['b_flags'])
 
                 start_time_str = d["time"].attrs["units"]
                 if not isinstance(start_time_str, str):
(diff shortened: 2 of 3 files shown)


sunpy/timeseries/sources/goes.py
```diff
@@ -285,10 +291,13 @@ def _parse_netcdf(filepath):
...
                 raise ValueError(f"The file {filepath} doesn't seem to be a GOES netcdf file.")
 
-        data = DataFrame({"xrsa": xrsa, "xrsb": xrsb}, index=times.datetime)
+        data = DataFrame({"xrsa": xrsa, "xrsb": xrsb, "xrsa_quality": xrsa_quality,
+                         "xrsb_quality": xrsb_quality}, index=times.datetime)
         data = data.replace(-9999, np.nan)
         units = OrderedDict([("xrsa", u.W/u.m**2),
-                             ("xrsb", u.W/u.m**2)])
+                             ("xrsb", u.W/u.m**2),
+                             ("xrsa_quality", int),
```
mariusgiger, line 299: Is it better to name this xrsa_flags as well?

sunpy/timeseries/sources/goes.py
```diff
@@ -218,6 +218,7 @@ def _parse_hdus(cls, hdulist):
                 raise ValueError("Date not recognized")
             xrsb = hdulist[2].data['FLUX'][0][:, 0]
             xrsa = hdulist[2].data['FLUX'][0][:, 1]
+            # TODO how to extract quality flags from HDU?
```
nabobalis, line 221: Do you have a FITS file to test with? I would hope it's one of the array rows in a HDU.

sunpy/timeseries/sources/goes.py
```diff
@@ -274,6 +277,9 @@ def _parse_netcdf(filepath):
             elif "xrsa_flux" in d.variables:
                 xrsa = np.array(d["xrsa_flux"])
                 xrsb = np.array(d["xrsb_flux"])
+                xrsa_quality = np.array(d['xrsa_flags'])
+                xrsb_quality = np.array(d['xrsb_flags'])
```
nabobalis, line 281: Are there other columns that would be useful to have extracted too?

sunpy/timeseries/sources/goes.py
mariusgiger: When looking at the FITS header the following information is contained,
e.g. for https://umbra.nascom.nasa.gov/goes/fits/2019/go1520191210.fits:

```
import sunpy.io
from sunpy.util.metadata import MetaDict
from collections import OrderedDict

filepath= "./go1520191210.fits"
hdulist = sunpy.io.read_file(filepath)
header = MetaDict(OrderedDict(hdulist[3].header))
print(header)
print(hdulist[3].data)
>FITS_rec([(0.55999994, [0., 0.])],
         dtype=(numpy.record, [('Time', '>f4'), ('Status', '>f4', (2,))]))
```

The following doc is provided for `Status`, however it is not clear to me how the status maps to xrsa and xrsb (e.g. in netcdf format there is a value for each channel and measurement):

Convert status words to long integer then examine bits.Status 
1 - octal mask, description (incomplete list):
- 1000, Sun eclipsed by MoonStatus 
2 - octal mask, description (incomplete list):
- 1, X-Ray detector off
- 2, X-Ray detector being calibrated
- 4, X-Ray Transient
- 10, X-Ray short channel saturation
- 20, X-Ray long channel range change
- 40, X-Ray short channel range change
- 200, X-Ray long channel saturation')

Does someone know how to extract this information?

sunpy/timeseries/sources/goes.py
mariusgiger: other variables in netcdf - not sure if any of those should also be exposed:

```
xrsa1_flux: ('time',) float32
xrsa2_flux: ('time',) float32
xrsb1_flux: ('time',) float32
xrsb2_flux: ('time',) float32
xrsa_primary_chan: ('time',) uint8
xrsb_primary_chan: ('time',) uint8
xrsa1_flags: ('time',) uint16
xrsa2_flags: ('time',) uint16
xrsb1_flags: ('time',) uint16
xrsb2_flags: ('time',) uint16
corrected_current_xrsa2: ('time', 'quad_diode') float32
corrected_current_xrsb2: ('time', 'quad_diode') float32
au_factor: ('time',) float32
roll_angle: ('time',) float32
yaw_flip_flag: ('time',) uint8
```

sunpy/timeseries/sources/goes.py
mariusgiger: the value is a flag based on the following mask:

```
[0,   1,   2,   4,   8,  16,  32,  64, 128, 256, 512]
[good_data, eclipse, particle_spike, calibration, off_point, temperature_error, data_quality_error, pointing_error, invalid_mode, missing_data, L0_error]
```

any value different from 0 means something is not correct

e.g. 72 = 64 + 8 = pointing_error + off_point

mariusgiger:
@wtbarnes yes this could also be an approach - the question is which values to mask without hiding something from the user that should actually be available.

One example could be:

```
ts['xrsb'].mask = ts['xrsb_flags'] == 0
```

dstansby:
It's not possible to mask data in pandas as far as I know. I think we should limit the scope of this PR to returning the quality flags, and backporting as a bugfix. @mariusgiger could you add a short changelog entry in the `changelog` directory (there's instructions on what to do in there? Once that's done I'm happy to approve this.

dstansby:
It's not possible to mask data in pandas as far as I know. I think we should limit the scope of this PR to returning the quality flags, and backporting as a bugfix. @mariusgiger could you add a short changelog entry in the `changelog` directory (there's instructions on what to do in there? Once that's done I'm happy to approve this.

sunpy/timeseries/sources/goes.py
dstansby: I think this is beyond scope of this PR - could someone open an issue so we can keep track of this though?

dstansby: I think this is beyond scope of this PR - could someone open an issue so we can keep track of this though?

sunpy/timeseries/sources/goes.py
dstansby: I think we can leave these, unless someone else requests one of them with a reason.

sunpy/timeseries/sources/goes.py
nabobalis: I would rather we didn't add support only for one file type but I don't want to block this PR on that.

examples/acquiring_data/goes_xrs_example.py
```diff
@@ -67,6 +67,14 @@
 goes_15 = ts.TimeSeries(file_goes15)
 goes_15.peek()
 
+#############################################################
+# The resulting `~sunpy.timeseries.TimeSeries` can be filtered by GOES quality flags. For more information
+# refer to the `GOES Data Guide <https://satdat.ngdc.noaa.gov/sem/goes/data/science/xrs/GOES_13-15_XRS_Science-Quality_Data_Readme.pdf>`__.
+
+df = goes_15.to_dataframe()
+df = df[(df["xrsa_quality"] == 0) & (df["xrsb_quality"] == 0)]
+goes_15.data = df
```
dstansby, line 76:
```suggestion
goes_15 = ts.TimeSeries(df, goes_15.meta, goes_15.units)
```
This is the best way to do this - if we change the internals of how `TimeSeries` works in the future this will avoid relying `.data` being a dataframe.

dstansby (changes requested):
Other than my single comment, looks 👍

dstansby (approved):

nabobalis:
Thanks for the PR @mariusgiger, I have created a tracking issue about the FITS files.

PR #6262, opened 2022-06-16
Goes parse netcdf and leap seconds 

nabobalis:
Failing example that now passes
```python
from sunpy.net import Fido

t_start = "2015-06-30 00:00"
t_end = "2015-07-01 00:00:00"
search_result = Fido.search(
```


sunpy/timeseries/sources/goes.py
```diff
@@ -257,39 +257,34 @@ def _parse_netcdf(filepath):
...
+            try:
                 # Perform the time addition in UTime format to ignore leap seconds
-                times = Time(parse_time(start_time_str).utime + d["time"], format="utime")
-
-            else:
-                raise ValueError(f"The file {filepath} doesn't seem to be a GOES netcdf file.")
+                times = Time(parse_time(start_time_str).utime + h5nc["time"], format="utime").datetime
+            except ValueError:
+                # If that fails, we can time addition as seconds since epoch.
+                times = Time(parse_time(start_time_str) + TimeDelta(h5nc["time"], format="sec")).datetime
```
nabobalis, line 278: This was the old way of avoiding leap seconds but it works for this case.

nabobalis:
> Why are there now two different ways to parse the times here? Do they produce different results for identical data? If not, why not just use the second newly added method?

Yes, they produce times that are about 26 seconds for each time stamp.

dstansby:
> > Why are there now two different ways to parse the times here? Do they produce different results for identical data? If not, why not just use the second newly added method?
> 
> Yes, they produce times that are about 26 seconds for each time stamp.

Well we can't have code that produces completely different times depending on if your data has a leapsecond in it or not...

nabobalis:
> > > Why are there now two different ways to parse the times here? Do they produce different results for identical data? If not, why not just use the second newly added method?
> > 
> > 
> > Yes, they produce times that are about 26 seconds for each time stamp.
> 
> Well we can't have code that produces completely different times depending on if your data has a leapsecond in it or not...

Yeah, so if I do:

```
 Time(parse_time(start_time_str).unix_tai + h5nc["time"], format="unix_tai").datetime
```
instead, it works for both the example code that was failing and our tests. Issue is, it changes the times checked for in our tests.

nabobalis:
The difference is that there have been 27 leap seconds and that's the difference. 

So do we want to avoid having any leap seconds or not? I am not sure I can make that decision.

dstansby:
https://github.com/sunpy/sunpy/issues/5914 has some past context on this.

nabobalis:
That is helpful, I will adjust my code.

sunpy/timeseries/sources/goes.py
```diff
@@ -249,50 +250,43 @@ def _parse_netcdf(filepath):
...
+            start_time_str = start_time_str.lstrip("seconds since").rstrip("UTC").strip()
+            times = Time(parse_time(start_time_str).unix + h5nc["time"], format="unix")
 
-        data = DataFrame({"xrsa": xrsa, "xrsb": xrsb, "xrsa_quality": xrsa_quality,
-                         "xrsb_quality": xrsb_quality}, index=times.datetime)
+        try:
+            times = times.datetime
+        except ValueError:
+            warn_user("There is a leap second in the data, this has been subtracted by 1 second to work around this")
+            idx = np.argwhere((np.char.find(times.isot, ":60.") != -1) == True).flatten()[0]
```
nabobalis, line 277: I am assuming there is only one leap second in this file.

wtbarnes, line 277: Please give a few comments as to what each step of this line means/does

sunpy/timeseries/sources/goes.py
```diff
@@ -249,50 +250,43 @@ def _parse_netcdf(filepath):
...
+                start_time_str = start_time_str.decode("utf-8")
+            start_time_str = start_time_str.lstrip("seconds since").rstrip("UTC").strip()
+            times = Time(parse_time(start_time_str).unix + h5nc["time"], format="unix")
 
-        data = DataFrame({"xrsa": xrsa, "xrsb": xrsb, "xrsa_quality": xrsa_quality,
-                         "xrsb_quality": xrsb_quality}, index=times.datetime)
+        try:
+            times = times.datetime
+        except ValueError:
+            warn_user("There is a leap second in the data, this has been subtracted by 1 second to work around this")
```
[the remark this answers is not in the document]
nabobalis: We currently have no way of passing args down to this function. I can see about adding it. But since there are 25 or so leap seconds since 1971, it shouldn't be too many messages.

sunpy/timeseries/sources/goes.py
nabobalis: I added some comments and I removed the flatten and index bits

sunpy/timeseries/sources/goes.py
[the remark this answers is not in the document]
nabobalis:
```
Files Downloaded: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:00<00:00,  9.30file/s]
WARNING: SunpyUserWarning: There is a leap second present in: sci_gxrs-l2-irrad_g15_d20150630_v0-0-0.nc, 1 second has been subtracted to allow its conversion into a Python datetime. The leap second timestamp was at index: [[42169]] - [['2015-06-30T23:59:60.428']] [sunpy.timeseries.sources.goes]
WARNING: SunpyUserWarning: There is a leap second present in: sci_gxrs-l2-irrad_g13_d20150630_v0-0-0.nc, 1 second has been subtracted to allow its conversion into a Python datetime. The leap second timestamp was at index: [[42188]] - [['2015-06-30T23:59:60.965']] [sunpy.timeseries.sources.goes]
```
Is how the message looks right now.

sunpy/timeseries/sources/goes.py
```diff
@@ -222,50 +223,52 @@ def _parse_netcdf(filepath):
...
+        except ValueError:
+            # We make an assumption here that there is only 1 leap second in the file.
+            # We do not make the assumption that the leap second occurs at the end of the file.
+            # Therefore, we need to find it:
+            # To do so, we convert the times to isot strings, use numpy to find the the leap second string,
+            # then use that to workout the index of the leap timestamp.
+            idx = np.argwhere((np.char.find(times.isot, ":60.") != -1) == True)
+            warn_user(
+                f"There is a leap second present in: {Path(filepath).name}, "
+                "1 second has been subtracted to allow its conversion into a Python datetime. "
```
dstansby, line 257: This is currently a bit ambiguous - has 1 second been subtracted from every timestamp or just the leapsecond timestamp?

dstansby (changes requested):
As well as my comment above

- can we add a file that has a leapsecond as a test file (how big would it be?) and add a test for this new code path?
- does this result in the time ordering of data being changed, ie. if there was data at `...:59.8, :60.2` does this data now get timestamps of `:59.8, 59.2`? If so I think this is sub-optimal and a better solution is flooring the timestamps at `59.999999` or whatever the maximum timestamp is

nabobalis:
> * can we add a file that has a leapsecond as a test file (how big would it be?) and add a test for this new code path?

I will need to check and add one, if not, I can see about faking one. 

> * does this result in the time ordering of data being changed, ie. if there was data at `...:59.8, :60.2` does this data now get timestamps of `:59.8, 59.2`? If so I think this is sub-optimal and a better solution is flooring the timestamps at `59.999999` or whatever the maximum timestamp is

In theory yes, I think in practice the time step is larger so this won't occur. I can floor the timestamp instead?

sunpy/timeseries/sources/goes.py
nabobalis: I changed the message, hopefully it's more clear.

nabobalis:
Grumble, my test file is broken. It opens locally, so I have a version mismatch I have to address.

sunpy/timeseries/sources/goes.py
```diff
@@ -222,50 +223,52 @@ def _parse_netcdf(filepath):
...
+            start_time_str = start_time_str.lstrip("seconds since").rstrip("UTC").strip()
+            times = Time(parse_time(start_time_str).unix + h5nc["time"], format="unix")
 
-        data = DataFrame({"xrsa": xrsa, "xrsb": xrsb, "xrsa_quality": xrsa_quality,
-                         "xrsb_quality": xrsb_quality}, index=times.datetime)
+        try:
+            times = times.datetime
+        except ValueError:
+            # We make an assumption here that there is only 1 leap second in the file.
+            # We do not make the assumption that the leap second occurs at the end of the file.
```
nabobalis, line 250: I think we could make this assumption easily enough.

dstansby:
I think flooring the timestamp is the best way to go here to preserve the time-ordering of the data.

nabobalis:
> I think flooring the timestamp is the best way to go here to preserve the time-ordering of the data.

Done

dstansby (approved):

sunpy/timeseries/sources/goes.py
```diff
@@ -222,50 +223,52 @@ def _parse_netcdf(filepath):
...
-            else:
-                raise ValueError(f"The file {filepath} doesn't seem to be a GOES netcdf file.")
+        with h5netcdf.File(filepath, mode="r", **XRSTimeSeries._netcdf_read_kw) as h5nc:
+            header = MetaDict(OrderedDict(h5nc.attrs))
+            flux_name = h5nc.variables.get("a_flux") or h5nc.variables.get("xrsa_flux")
+            if flux_name is None:
+                raise ValueError(f"No flux data (either a_flux or xrsa_flux) found in file: {filepath}")
+            flux_name = flux_name.name
+            xrsa = np.array(h5nc[flux_name])
+            xrsb = np.array(h5nc[flux_name.replace("a", "b")])
```
mariusgiger, line 236: introduce a second variable for xrsb instead of the replacement, same for the flags as it makes it difficult to read the code

sunpy/timeseries/sources/goes.py
```diff
@@ -222,50 +223,52 @@ def _parse_netcdf(filepath):
...
-                         "xrsb_quality": xrsb_quality}, index=times.datetime)
+        try:
+            times = times.datetime
+        except ValueError:
+            # We make an assumption here that there is only 1 leap second in the file.
+            # We do not make the assumption that the leap second occurs at the end of the file.
+            # Therefore, we need to find it:
+            # To do so, we convert the times to isot strings, use numpy to find the the leap second string,
+            # then use that to workout the index of the leap timestamp.
+            idx = np.argwhere((np.char.find(times.isot, ":60.") != -1) == True)
```
mariusgiger, line 254: as you only expect one index to be returned here, already make sure that it is a number and not an array of arrays

mariusgiger (approved):
Looks good to me, two small suggestions

sunpy/timeseries/sources/goes.py
nabobalis: This does not return a number but an nested array with an number. Luckily when I use it to index the original array, it handles it.

sunpy/timeseries/sources/goes.py
nabobalis: I have created one for xrsb.

sunpy/timeseries/sources/goes.py
nabobalis: Oh you mean, I should assert it has one value.

nabobalis: This should be fixed now.

mariusgiger: theoretically ValueError could happen for a number of reasons (not just because of leap seconds), being a bit defensive is therefore better ;) currently it would have failed for the following statement if there was no index found `times[idx].isot.tolist()[0][0]` with your fix it should be clearer - when revisiting the code `tolist()[0][0][:17]` might still bear some mysteries for future readers that you might want to improve. e.g.

```
idx = idx[0][0]
leap_second = times[idx]
rounded_leap_second = leap_second[:17] + "59.999"
times[idx] = rounded_leap_second
```

not sure, maybe too verbose

nabobalis: That is fair, at least for getting this out sooner, I probably won't make more changes. 

This code will also be deleted when we transition to astropy.times for Timeseries for 4.1 so I am ok with it being a bit rough.

nabobalis:
Backported manually.

Issue #6270, opened 2022-06-19
Switch from `sphinx-panels` to `sphinx-design`


dstansby:
### Provide a general description of the issue or problem.

https://sphinx-panels.readthedocs.io/en/latest/index.html says

> This repository is not actively maintained. Use [sphinx-design](https://github.com/executablebooks/sphinx-design) instead! See [the migration guide](https://sphinx-design.readthedocs.io/en/latest/get_started.html#migrating-from-sphinx-panels) and [this github issue](https://github.com/executablebooks/sphinx-design/issues/51) for more information.

PR #6271, opened 2022-06-19
sphinx_panels -> sphinx_design

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/6270

The CSS isn't the same, something we will need to tweak I think in our theme. 

Untitled

Built with Sphinx 5.0.2 as well now.

### DIFF
--- a/docs/dev_guide/index.rst
+++ b/docs/dev_guide/index.rst
@@ -12,54 +12,65 @@ Developer's Guide
 
 This section contains the various guidelines to be followed by anyone working on sunpy.
 
-.. panels::
+.. grid:: 1 2 2 2
+    :gutter: 3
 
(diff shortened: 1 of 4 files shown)


docs/dev_guide/index.rst
```diff
@@ -12,54 +12,65 @@ Developer's Guide
...
+    .. grid-item-card::
+        :class-card: card
+
+        Getting started
+        ^^^^^^^^^^^^^^^
+
+        .. toctree::
+            :maxdepth: 3
+
+            contents/newcomers
```
dstansby, line 27: Having this at a depth of 3 looks odd, but beyond scope of this PR

dstansby (approved):

PR #6281, opened 2022-06-22
tweak goes plotting code to avoid quality flags

nabobalis:
The example was failing due to us trying to plot all the columns.

### DIFF
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -1521,7 +1521,7 @@ def test_map_arithmetic_multiplication_division(aia171_test_map, value):
     check_arithmetic_value_and_units(new_map, value * aia171_test_map.quantity)
     new_map = aia171_test_map / value
     check_arithmetic_value_and_units(new_map, aia171_test_map.quantity / value)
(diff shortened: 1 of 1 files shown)


sunpy/timeseries/sources/goes.py
```diff
@@ -89,6 +89,8 @@ def plot(self, axes=None, columns=None, **kwargs):
         axes, columns = self._setup_axes_columns(axes, columns)
         plot_settings = {"xrsa": ["blue", r"0.5--4.0 $\AA$"], "xrsb": ["red", r"1.0--8.0 $\AA$"]}
         data = self.to_dataframe()
+        # We do not want to plot the quality flags in the light curve.
+        columns = [col for col in columns if col not in ["xrsa_quality", "xrsb_quality"]]
         for channel in columns:
```
dstansby, line 94: Sorry, my bad for merging that PR without checking... Might be a bit nicer to do this instead?
```suggestion
        for channel in ['xrsa', 'xrsb']:
```

nabobalis: That is so much simpler, I don't know why I didn't do that.

nabobalis:
I reverted the tweak since we need to account for if a user passes in a column list (as one of the test does).

PR #6282, opened 2022-06-23
numpy warning changed

nabobalis:


dstansby (dismissed):
This needs to be be adapted to work with both old/new messages to allow our oldestdeps to pass. I think that should be easy enough to do with regex

sunpy/map/tests/test_mapbase.py
```diff
@@ -1521,7 +1521,7 @@ def test_map_arithmetic_multiplication_division(aia171_test_map, value):
     check_arithmetic_value_and_units(new_map, value * aia171_test_map.quantity)
     new_map = aia171_test_map / value
     check_arithmetic_value_and_units(new_map, aia171_test_map.quantity / value)
-    with pytest.warns(RuntimeWarning, match='divide by zero encountered in true_divide'):
+    with pytest.warns(RuntimeWarning, match='divide by zero encountered in divide'):
```
Cadair, line 1524: I mean the stupidest solution is just this right?

```suggestion
    with pytest.warns(RuntimeWarning, match='divide by zero encountered in'):
```

PR #6284, opened 2022-06-23
Run figure tests with astropy v5.1.0

ConorMacBride:
Astropy versions before v5.1 are not compatible with numpy v1.23.0 (released yesterday). So the figure tests, which are pinned to astropy v4.2.1, are failing: https://app.circleci.com/pipelines/github/sunpy/sunpy/18426/workflows/04755d06-2562-4ef4-b1cf-2d6b41d79fdf/jobs/79538

Astropy v4.2.1 was released in April 2021 so probably a good time for an upgrade anyway, and pinning numpy instead seems counterproductive.

- [x] Review and update hashes


nabobalis:
UPDATE ALL THE HASHES

ConorMacBride:
We have one visible difference which seems fine and 6 with just hashes different. What's very strange is we have two tests where the hashes match but the images differ!!! That should be impossible, so I guess the baseline update workflow isn't working?

ConorMacBride:
It's configured to not update any baselines if either `figure` or `figure-devdeps` fails. Is that what we want?

nabobalis:
It should update the successful ones regardless of if the other one fails, can we update it?

ConorMacBride:
That should work

nabobalis:
Oh we should also update the images: https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034

nabobalis:
Oh those new images might be a problem now:

```
ERROR: InterpreterNotFound: python3.9
```

nabobalis:
Maybe we should just make them both use the same python version.

nabobalis (approved):

ConorMacBride:
The dev figures have an upstream matplotlib bug

nabobalis:
> The dev figures have an upstream matplotlib bug

Not anymore.

PR #6286, opened 2022-06-24
[4.0] Merge pull request #6284 from ConorMacBride/upgrade-astropy

nabobalis:


ConorMacBride:
The figure tests are not relevant on this branch, but maybe we should keep these files in sync anyway. You seem to have reverted the dev hashes though?

nabobalis:
> The figure tests are not relevant on this branch, but maybe we should keep these files in sync anyway. You seem to have reverted the dev hashes though?

Some of the timeseries changes were never backported, so I had to revert some hashes to make them inline with the branch.

PR #6295, opened 2022-06-24
Add support for downloading sample data from more than two mirrors

Cadair:


ConorMacBride (approved):
This looks great :+1: I tested by running `sunpy.data.download_sample_data(overwrite=True)` with the GitHub URLs edited to be 404, and with the GitHub URLs as is and it works as intended. `data.sunpy.org` seemed a good bit faster than GitHub.

dstansby (approved):

Cadair:
> data.sunpy.org seemed a good bit faster than GitHub.

yeah I noticed that as well.

Issue #6299, opened 2022-06-24
Error in GOES gallery example


dstansby:
### Describe the bug

The GOES gallery example is broken, with the following error traceback on GH actions. This should be fixed,.

```python
/home/runner/work/sunpy/sunpy/examples/time_series/goes_hek_m25.py failed leaving traceback:
Traceback (most recent call last):
  File "/home/runner/work/sunpy/sunpy/examples/time_series/goes_hek_m25.py", line 39, in <module>
    goes.plot()
  File "/home/runner/work/sunpy/sunpy/.tox/build_docs-gallery/lib/python3.8/site-packages/sunpy/timeseries/sources/goes.py", line 94, in plot
    data.index, data[channel], "-", label=plot_settings[channel][1], color=plot_settings[channel][0], lw=2, **kwargs
KeyError: 'xrsa_quality'
```

PR #6302, opened 2022-06-24
Fix GOES XRS plotting

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/6299


sunpy/timeseries/sources/goes.py
```diff
@@ -89,7 +89,7 @@ def plot(self, axes=None, columns=None, **kwargs):
         axes, columns = self._setup_axes_columns(axes, columns)
         plot_settings = {"xrsa": ["blue", r"0.5--4.0 $\AA$"], "xrsb": ["red", r"1.0--8.0 $\AA$"]}
         data = self.to_dataframe()
-        for channel in columns:
+        for channel in ['xrsa', 'xrsb']:
```
wtbarnes, line 92: This is not the right fix. It ignores any columns specifed in the `columns` kwarg (as evidenced by the failing test).

wtbarnes (changes requested):

wtbarnes (approved):

PR #6303, opened 2022-06-24
Move timeseries plotting tests to their source test file

dstansby:
In my continuing quest to clean up the mess that is `timeseries`, this moves figure tests to their respective source test file. This means it's easy to see which ones are being tested and which ones aren't. I also put all the test fixtures in a `conftest.py` file so they can be shared across timeseries tests.


test.py
```diff
@@ -0,0 +1,7 @@
+import sunpy.timeseries
```
nabobalis, line 1: I assume this should be copied somewhere else?

nabobalis (approved):

nabobalis:
Will need a rebase ontop of main due to file hash conflicts.

nabobalis (approved):

PR #6311, opened 2022-06-28
Fixed a bug when working with an observer defined in HEEQ

ayshih:
As [reported by @samaloney](https://github.com/samaloney/stixpy/pull/39#discussion_r908321543), trying to work with an observer defined in `HeliographicStonyhurst` with a Cartesian representation (i.e., equivalent to HEEQ) causes transformations to error.  This is due to the observer-equality check assuming that an `HeliographicStonyhurst` observer is already be in spherical representation.  There are other places in our code that also assume that the observer is already in spherical representation.

This PR fixes those bugs by converting a `HeliographicStonyhurst` observer to the default representation (i.e., spherical).

### DIFF
--- a/changelog/6311.bugfix.rst
+++ b/changelog/6311.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed bugs when working with a coordinate frame where the observer is specified in `~sunpy.coordinates.frames.HeliographicStonyhurst` and Cartesian representation, which is equivalent to Heliocentric Earth Equatorial (HEEQ).
+Now, the observer will always be converted to spherical representation.
--- a/changelog/6311.bugfix.rst
+++ b/changelog/6311.bugfix.rst
@@ -1,2 +1,2 @@
-Fixed bugs when working with a coordinate frame where the observer is specified in `~sunpy.coordinates.frames.HeliographicStonyhurst` and Cartesian representation, which is equivalent to Heliocentric Earth Equatorial (HEEQ).
-Now, the observer will always be converted to spherical representation.
+Fixed bugs when working with a coordinate frame where the observer is specified in `~sunpy.coordinates.frames.HeliographicStonyhurst` with a Cartesian representation, which is equivalent to Heliocentric Earth Equatorial (HEEQ).
+Now, the observer will always be converted to spherical representation when the coordinate frame is created.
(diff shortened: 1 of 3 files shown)


changelog/6311.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed bugs when working with a coordinate frame where the observer is specified in `~sunpy.coordinates.frames.HeliographicStonyhurst` and Cartesian representation, which is equivalent to Heliocentric Earth Equatorial (HEEQ).
+Now, the observer will always be converted to spherical representation.
```
dstansby, line 2:
```suggestion
Now, the observer will always be converted to spherical representation when the coordinate is created.
```
If I'm interpreting this correctly, this is a slightly more helpful message.

changelog/6311.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed bugs when working with a coordinate frame where the observer is specified in `~sunpy.coordinates.frames.HeliographicStonyhurst` and Cartesian representation, which is equivalent to Heliocentric Earth Equatorial (HEEQ).
```
dstansby, line 1:
```suggestion
Fixed bugs when working with a coordinate frame where the observer is specified in `~sunpy.coordinates.frames.HeliographicStonyhurst` with a Cartesian representation, which is equivalent to Heliocentric Earth Equatorial (HEEQ).
```

sunpy/coordinates/tests/test_frameattributes.py
```diff
@@ -116,6 +116,17 @@ def test_observer_not_hgs_astropy(oca):
     assert converted
 
 
+def test_observer_not_default_representation(oca):
```
dstansby, line 119: Am I right in thinking that this doesn't actually run any code that would have caused the bug beforehand? If so it would be good to add a test for a workflow a user would use that would have errored before this fix.

Cadair: It does, it just directly calls the converter function, which is automatically called when assigning an observer attribute to a sunpy frame. (Via a lot of descriptors in astropy)

changelog/6311.bugfix.rst
ayshih: Edited

sunpy/coordinates/tests/test_frameattributes.py
ayshih: As @Cadair says, the test I wrote does fail on bugged code, but it's not how a user normally interfaces with the `observer` frame attribute.  I've added another test that is more like user code.

dstansby (approved):

Issue #6300, opened 2022-06-24
Error removing Fido results in gallery examples


dstansby:
### Describe the bug

Several of the gallery examples are failing with tracebacks like the ones below. These should be fixed. From GH actions the failing examples are:

- `time_series/timeseriesmetadata_example.py`
- `plotting/solar_cycle_example.py`
- `acquiring_data/goes_xrs_example.py`

```python
 /home/runner/work/sunpy/sunpy/examples/time_series/timeseriesmetadata_example.py failed leaving traceback:
Traceback (most recent call last):
  File "/home/runner/work/sunpy/sunpy/examples/time_series/timeseriesmetadata_example.py", line 21, in <module>
    norh_res = Fido.search(a.Time("2010-11-03", "2010-11-03 23:59:59"), a.Instrument.norh,
  File "/home/runner/work/sunpy/sunpy/.tox/build_docs-gallery/lib/python3.8/site-packages/sunpy/net/fido_factory.py", line 322, in search
    results.remove(vres)
  File "/home/runner/work/sunpy/sunpy/.tox/build_docs-gallery/lib/python3.8/site-packages/astropy/table/table.py", line 3426, in __eq__
    return self._rows_equal(other)
  File "/home/runner/work/sunpy/sunpy/.tox/build_docs-gallery/lib/python3.8/site-packages/astropy/table/table.py", line 3474, in _rows_equal
    result = self.as_array() == other
TypeError: Cannot compare structured or void to non-void arrays.
```

nabobalis:
What is going on is that when you do a Fido search and an entry is empty as can be for the VSO, we go to remove it.

However, now with numpy 1.23.0, the comparison errors when one is empty. 
This means that certain generic client searches will not work with Fido at the moment.

Cadair:
Is there an astropy bug for this behaviour change?

nabobalis:
I didn't check

dstansby:
This might be the upstream bug: https://github.com/astropy/astropy/issues/13421 (it's a slightly different error message, but fixing that bug might fix the bug here too?)

PR #6317, opened 2022-06-30
Pin maximum numpy version

dstansby:
This is a stopgap fix for https://github.com/sunpy/sunpy/issues/6300. We should probably fix that issue properley before doing any more releases.


nabobalis:
Closed in favour of https://github.com/sunpy/sunpy/pull/6318

PR #6318, opened 2022-07-02
Fix filtering of empty VSO results w/ numpy 1.23

dstansby:
Should fix https://github.com/sunpy/sunpy/issues/6300, by avoiding use of `.remove()`. I think the new code is a bit nicer anyway.

### DIFF
--- a/sunpy/net/fido_factory.py
+++ b/sunpy/net/fido_factory.py
@@ -315,11 +315,7 @@ class UnifiedDownloaderFactory(BasicRegistrationFactory):
         # client generated results, we drop the empty VSO results for tidiness.
         # This is because the VSO _can_handle_query is very broad because we
         # don't know the full list of supported values we can search for (yet).
-        if len(results) > 1:
-            vso_results = list(filter(lambda r: isinstance(r, vso.VSOQueryResponseTable), results))
-            for vres in vso_results:
-                if len(vres) == 0:
-                    results.remove(vres)
+        results = [r for r in results if not isinstance(r, vso.VSOQueryResponseTable) or len(r) > 0]
 
         return UnifiedResponse(*results)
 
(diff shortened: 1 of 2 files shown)


nabobalis (approved):

sunpy/net/fido_factory.py
```diff
@@ -315,11 +315,7 @@ def search(self, *query):
         # client generated results, we drop the empty VSO results for tidiness.
         # This is because the VSO _can_handle_query is very broad because we
         # don't know the full list of supported values we can search for (yet).
-        if len(results) > 1:
-            vso_results = list(filter(lambda r: isinstance(r, vso.VSOQueryResponseTable), results))
-            for vres in vso_results:
-                if len(vres) == 0:
-                    results.remove(vres)
+        results = [r for r in results if not isinstance(r, vso.VSOQueryResponseTable) or len(r) > 0]
```
Cadair, line 318: I don't think this logic is correct at all?

The logic here is if more than one results table is returned and one of those tables is a VSO table and the VSO table is empty then drop the VSO table from this list.

I think this logic has broken searches where you only have VSO results?

dstansby: > I think this logic has broken searches where you only have VSO results?

As long as there's some results I think this is still fine (certainly no tests are failing...) because any table with `len > 0` is kept. It might break previous behaviour when nothing returns any results. I shall add a test for that in another PR.

dstansby: Before this PR, a search with no results from anywhere would return 
```
Results from 1 Provider:

0 Results from the VSOClient:
Source: http://vso.stanford.edu/cgi-bin/search

<No columns>
```

After this PR it's just 
```
Results from 0 Providers:
```

If I was writing this from scratch I'd definitely go for the second option, but for backwards compatability should I open a new PR to return an empty VSO table?

nabobalis: I think its fine, the new behavior is better than returning an empty VSO table.

Cadair: Ok that works.

Cadair: The comment above is now slightly incorrect though. Should probably be reworded so it's the same as the logic.

PR #6327, opened 2022-07-08
Deprecate GenericTimeSeries.index

dstansby:
The goal here is to deprecate `.index`, which returns a pandas date-time array, and replace it with `.time` which returns astropy times. This new API is the same as `astropy.timeseries.TimeSeries`.

This is pulled out of https://github.com/sunpy/sunpy/pull/5834/files for easier review.

### DIFF
--- a/examples/acquiring_data/goes_xrs_example.py
+++ b/examples/acquiring_data/goes_xrs_example.py
@@ -30,9 +30,13 @@ available from each satellite. Similarly there are times when GOES 16 and 17 ove
 import matplotlib.pyplot as plt
 import numpy as np
 
+from astropy.visualization import time_support
+
 from sunpy import timeseries as ts
 from sunpy.net import Fido
 from sunpy.net import attrs as a
+
+time_support()
 
 #############################################################
 # Lets first define our start and end times and query using the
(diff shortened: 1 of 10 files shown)


examples/acquiring_data/goes_xrs_example.py
```diff
@@ -30,10 +30,14 @@
 import matplotlib.pyplot as plt
 import numpy as np
 
+from astropy.visualization import time_support
+
 from sunpy import timeseries as ts
 from sunpy.net import Fido
 from sunpy.net import attrs as a
 
+time_support()
```
nabobalis, line 39:
```suggestion
# Enable support for plotting astropy times
time_support()
```

I think we want to decide on how/where we include this call in our examples.

nabobalis (approved):
Other than the examples comment, LGTM

examples/acquiring_data/goes_xrs_example.py
dstansby: Good point - what are your thoughts? seems like two options are
1. Under imports
2. Before first plot

I think I'd go with 2. personally, but am not too fussed.

nabobalis: I think 2 as well.

dstansby: 👍 I updated it to this and documented it in the dev docs

Issue #6315, opened 2022-06-28
Expose `processing_level` for XRT data


wtbarnes:
According to the Appendix of the [XRT Analysis Guide](https://xrt.cfa.harvard.edu/resources/documents/XAG/XAG.pdf), `DATA_LEV` keyword is used to denote the processing level (rather than the usual `LVL_NUM` key). However, our `XRTMap` source does not account for this and thus `.processing_level` returns `None`. The `processing_level` property should be special-cased to account for this keyword.

Issue #3650, opened 2020-01-07
Remove authorship information from the code itself?


ayshih:
Let's have a discussion about whether to remove authorship information from the code itself (e.g., `__author__` and `__email__`).  (See [triggering comment thread](https://github.com/sunpy/sunpy/pull/3646#discussion_r363508066).)

Some reasons to remove it:

- It's not currently commonly available across most of our code.
- The stuff that's there isn't necessarily accurate because it hasn't been maintained well.
- Git history can be a more accurate way to track authorship, particularly for code that involves a variety of contributors.
- I don't know whether users are even typically aware of this metadata variables.

Some reasons to keep it:

- When a code file is essentially the work of a single person, it's nice to acknowledge that person.
- `sunpy` as a downloaded package doesn't have its Git history, so no authorship information would be available via Git.
- Git history can be misleading about the authorship that actually matters if there are swathes of inconsequential edits (e.g., for style).
- I think the average (non-developer) user may be even less knowledgeable about using Git to get authorship information than looking for metadata variables.

If we do choose to keep this authorship information, it *needs* to be maintained.  Perhaps this decision could be made on a per-subpackage basis?

Discuss.

nabobalis:
If it's being stored anywhere its stored in a nice rst doc file somewhere.

Cadair:
I am in favour of removing it. Even in most cases when a file started off as the work on one or two people it rarely stays like that for very long and then the comments never get updated.

wtbarnes:
I'm also in favor of removing it. At best it is redundant as authorship is already recorded in the git history and at worst it is inaccurate. It also has to be maintained by hand. 

Whichever way we decide, I don't think we should do this on a subpackage level. It should be consistent across the entire code base.

Before we make a final decision though, I think we should get the opinions of those whose names we would be removing. Searching both `__author__` and `__authors__`, it looks like that's

- [ ] @ayshih 
- [ ] @khughitt 
- [ ] @wafels 
- [ ] @dpshelio 
- [ ] @ehsteve 
- [ ] @Hypnus1803 
- [ ] @mjm159,

ayshih:
> If it's being stored anywhere it belongs in a nice rst doc file.

That doesn't seem practical in terms of format or maintenance.  Authorship is naturally coupled with individual code files, so creating and maintaining a separate list of authorships is even more work.

> Even in most cases when a file started off as the work on one or two people it rarely stays like that for very long and then the comments never get updated.

[...]

PR #6328, opened 2022-07-08
Update XRTMap to handle the missing metadata for the observer

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/6315 as well

TODO:
- [x] Changelog


sunpy/map/sources/tests/test_xrt_source.py
```diff
@@ -13,45 +13,49 @@
...
+    assert (xrt_map.filter_wheel2_measurements ==
             ["Open", "Al_mesh", "Al_thick", "Be_thick", "Gband", "Ti_poly"])
 
 
-def test_wcs(xrt):
+def test_wcs(xrt_map):
     # Smoke test that WCS is valid and can transform from pixels to world coordinates
-    with pytest.warns(SunpyMetadataWarning, match='Missing metadata for observer'):
-        xrt.pixel_to_world(0*u.pix, 0*u.pix)
+    with pytest.warns(SunpyMetadataWarning, match='the following metadata is missing: dsun_obs'):
```
nabobalis, line 60: Level 0 files miss this but the Level 1 files do have it.

sunpy/map/sources/hinode.py
```diff
@@ -36,6 +36,9 @@ class XRTMap(GenericMap):
                                   "Be_thick", "Gband", "Ti_poly"]
 
     def __init__(self, data, header, **kwargs):
+        # https://community.openastronomy.org/t/sunpymetadatawarnings-when-using-hinode-xrt-data/393/7
+        header["HGLN_OBS"] = 0
+        header["HGLT_OBS"] = 0
```
nabobalis, line 41: I could add more of a comment but I feel like the link is good enough?

dstansby (changes requested):
Since https://github.com/sunpy/sunpy/pull/5501 our approach to source specific fixes has been to override map properties, and not to modify metadata, so this PR should be updated to take that approach (presumably to override `.observer_coordinate` property, but that's off the top of my head without checking) .

dstansby:
I am also of the view that whilst this might be a fine approach to take, we should:
1. Quantify and document how big the potential error is in setting these values to zero (probably via asking on the OpenAstronomy thread)
2. Possibly still emit some warning (but a single one that's source specific instead of the current ones) because we are setting innacurate data that could trip a user up depending on how accurate they need the observer coordinate information for their analysis.

Cadair:
I think you will want to override the property below `.observer_coordinate` the one that reads the groups of keys out of the header.

nabobalis:
I think I override it correctly?

sunpy/map/sources/hinode.py
```diff
@@ -1,8 +1,8 @@
 """Hinode XRT and SOT Map subclass definitions"""
-from sunpy.map import GenericMap
+import astropy.units as u
+from astropy.coordinates import Latitude, Longitude
 
-__author__ = ["Jack Ireland, Jose Ivan Campos-Rozo, David Perez-Suarez"]
-__email__ = "jack.ireland@nasa.gov"
```
dstansby, line 5: Should probably not remove this until we resolve https://github.com/sunpy/sunpy/issues/3650

dstansby (changes requested):
- I think you're going to need to override `.observer_coordinate` too to prevent the warning.
- I'm stil +0.5 on emitting some sort of warning here, but there should defintitely be a clear explanation of what we're assuming and why, and this is where a link to the forum post can go instead of being hidden in the comments
- We need to allow for the case where the metadata is present in the files and not assume 0 lat/lon in that case, e.g. if a user has manually set these values in the files before loading them. This will need a test too.

ayshih (changes requested):
- You'll need to override `.observer_coordinate` property (and not the `.heliographic_latitude`/`.heliographic_longitude` properties).  I don't think there's a way to fix this by overriding `._supported_observer_coordinates` property, because there is no keyword to use for the heliographic longitude.  The overridden `.observer_coordinate` property should check if it needs to apply the assumption, since there may be user-added observer metadata.
- The heliographic latitude should use the value in `SOLAR_B0`.  It should not be set to 0 degrees.
- This assumption should be stated in the `XRTMap` docstring.  I don't think it's necessary to quantify the uncertainty.  I also don't think any warning should be emitted.

ayshih:
Hinode is in low-Earth orbit (LEO) with an altitude of ~600 km.  Thus, it is always <1.1 Earth radii from Earth center.  (For comparison, SDO is ~6.5 Earth radii from Earth center.)  This amount of offset is negligible for normal imaging of the Sun.  The exception cases are when there is an intermediate body – Venus or the Moon – occulting part of the Sun.

nabobalis:
Ok, I updated it and rebased the PR.

I still think I did it wrong especially when it comes to settings the radius in the creation of the new skycoord.

The code isn't pretty either and I need to add a test to make sure if someone provides the correct metadata, it will not be overridden.

sunpy/map/sources/hinode.py
nabobalis: I think it should really go and we should decide on #3650.

sunpy/map/sources/hinode.py
```diff
@@ -36,24 +46,43 @@ class XRTMap(GenericMap):
                                   "Be_thick", "Gband", "Ti_poly"]
 
     def __init__(self, data, header, **kwargs):
-        super().__init__(data, header, **kwargs)
-
-        fw1 = header.get('EC_FW1_')
+        fw1 = header.get('EC_FW1_', '')
```
nabobalis, line 49: Maybe we want this code to error if the keyword is missing wholesale?

sunpy/map/sources/hinode.py
```diff
@@ -36,24 +46,43 @@ class XRTMap(GenericMap):
...
     @property
     def _timesys(self):
         if self.meta.get('timesys', '').upper() == 'UTC (TBR)':
             return 'UTC'
-        else:
-            return super()._timesys
+        return super()._timesys
+
+    @property
+    # @cached_property_based_on('_meta_hash')
```
nabobalis, line 65: I was unsure if I should restore this behaviour here.

sunpy/map/sources/hinode.py
```diff
@@ -21,6 +24,13 @@ class XRTMap(GenericMap):
 
     Hinode was launched on 22 September 2006 into a sun-synchronous orbit.
 
+    .. note::
+
+        If the required parameters for observator are not provided,
```
nabobalis, line 29: TOOD: Check this renders nicely.

sunpy/map/sources/hinode.py
dstansby: I agree, but given there's still disagreement lets resolve this in a different PR.

sunpy/map/sources/hinode.py
dstansby: You should be able to use
```
Notes
-----
```
To create a notes heading

sunpy/map/sources/hinode.py
dstansby: I can't see why not?

ayshih:
> I don't think there's a way to fix this by overriding `._supported_observer_coordinates` property, because there is no keyword to use for the heliographic longitude.

Whoops, I refreshed myself on how `._supported_observer_coordinates` actually works, and you should override that instead of `.observer_coordinate`.  I'll put up an (untested) code suggestion.

sunpy/map/sources/hinode.py
```diff
@@ -36,24 +46,43 @@ class XRTMap(GenericMap):
...
+            if "heliographic_stonyhurst" in missing_meta:
+                if "hgln_obs" in missing_meta["heliographic_stonyhurst"]:
+                    extra_kwargs = {"lon": 0,
+                                    "lat": self.meta.get("SOLAR_B0", sun.B0(self.meta.get("DATE_OBS")).to_value(u.deg)),
+                                    "radius": constants.radius
+                                    }
+                    sc = SkyCoord(obstime=self.date, **{**kwargs, **extra_kwargs})
+                    sc = sc.heliographic_stonyhurst
+                    return SkyCoord(sc.replicate(rsun=self._rsun_meters(sc.radius)))
+        return super().observer_coordinate
```
ayshih, line 85:
```suggestion
    @property
    def _supported_observer_coordinates(self):
        return (super()._supported_observer_coordinates
                + [(('solar_b0', 'dsun_obs'), {'lon': 0*u.deg,
                                               'lat': self.meta.get('solar_b0'),
                                               'radius': self.meta.get('dsun_obs'),
                                               'unit': (u.deg, u.deg, u.m),
                                               'frame': "heliographic_stonyhurst"})])
```
Note that unlike our other map sources, we add this option with the assumption at the end of the list so that we'll automatically use observer metadata if it already exists.

ayshih (changes requested):

sunpy/map/sources/hinode.py
```diff
@@ -21,6 +24,13 @@ class XRTMap(GenericMap):
 
     Hinode was launched on 22 September 2006 into a sun-synchronous orbit.
 
+    .. note::
+
+        If the required parameters for observator are not provided,
+        default values will be used for ``HGLT_OBS`` and ``HGLN_OBS``.
+        ``HGLN_OBS`` will be set to 0 degrees.
+        ``HGLT_OBS`` will be set to the B0 value.
```
ayshih, line 32:
```suggestion
    Notes
    -----
    XRT files do not normally specify the heliographic longitude of the spacecraft,
    and the default assumption is that the spacecraft is at zero Stonyhurst
    heliographic longitude (i.e., the same longitude as Earth).  This assumption is
    safe for nearly all analyses due to Hinode's orbital altitude of only ~600 km.
    This assumption is not made if ``HGLN_OBS`` and ``HGLT_OBS`` values have been
    explicitly added to the metadata.
``````

File: sunpy/map/sources/hinode.py (hunk 3 above)
Comment (nabobalis) [reply sunpy/map/sources/hinode.py:5]: If we don't decide tomorrow about this, I will undo these changes.

File: sunpy/map/mapbase.py [hunk 9]
```diff
@@ -1213,7 +1213,7 @@ def observer_coordinate(self):
                 # representation)
                 return SkyCoord(sc.replicate(rsun=self._rsun_meters(sc.radius)))
 
-            elif any(meta_list) and not set(keys).isdisjoint(self.meta.keys()):
+            elif not any(meta_list) and set(keys).isdisjoint(self.meta.keys()):
```
nabobalis, line 1216: I think this is correct?

nabobalis: Clearly not.

nabobalis: I think this line should be dropped and we work out the missing keys all time instead?

ayshih:
(or add a newline to the end of the warning message)

nabobalis:
Will do, once I fix my broken local env, I will make sure it works. This time.

sunpy/map/mapbase.py
nabobalis: Tweaked now.

sunpy/data/_sample.py
```diff
@@ -93,7 +93,7 @@ def _retry_sample_data(results, new_url_base):
     dl = Downloader(overwrite=True, progress=True, headers={'Accept-Encoding': 'identity'})
 
     for err in results.errors:
-        file_name = err.filepath_partial().name
+        file_name = err.url.split("/")[-1]
```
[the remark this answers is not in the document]
nabobalis: This will probably conflict with the open PR for parfive.

sunpy/map/sources/tests/test_xrt_source.py
```diff
@@ -13,45 +13,58 @@
...
+    assert xrt_map.measurement == measurement
 
 
-def test_wheel_measurements(xrt):
+def test_level_number(xrt_map):
+    assert xrt_map.processing_level == 0
+
+
+def test_heliographic_longitude(xrt_map):
+    with pytest.warns(SunpyMetadataWarning, match='Missing metadata for observer'):
```
dstansby, line 49: I thought the point was to get rid of these warnings?

sunpy/map/sources/hinode.py
```diff
@@ -36,24 +47,31 @@ class XRTMap(GenericMap):
...
+    @property
+    def _supported_observer_coordinates(self):
+        # Assume observer is at zero Stonyhurst heliographic longitude if not otherwise specified
+        # https://community.openastronomy.org/t/sunpymetadatawarnings-when-using-hinode-xrt-data/393/7 for more information
+        return (super()._supported_observer_coordinates
+                + [(('solar_b0', 'dsun_obs'), {'lon': 0*u.deg,
+                                               'lat': self.meta.get('solar_b0'),
+                                               'radius': self.meta.get('dsun_obs'),
+                                               'unit': (u.deg, u.deg, u.m),
+                                               'frame': "heliographic_stonyhurst"})])
```
dstansby, line 74: Is `'solar_b0'` in the meadata, so we don't actually need to calculate it ourselves?

dstansby (changes requested):
Wasn't the point of this PR to get rid of the warnings? If so there shouldn't be an `pytest.warns` for missing metadata in the tests any more.

This should also get a new test to check that manually entered metadata takes preference over the new assumption.

Otherwise looks 👍

sunpy/map/sources/hinode.py
nabobalis: In the level 1 data it should be, we don't have a level 1 header at the moment.

sunpy/map/sources/tests/test_xrt_source.py
nabobalis: We use a level 0 header from what I can see and that will raise this warning. 

In a level 1 header (which I still need to source), it should not raise this warning.

dstansby: Are we not trying to get rid of the warning for the L0 header too?

nabobalis: I think we would need to calculate solar_b0 if its missing for the level 0 headers but I think people are meant to use the level 1 files.

ayshih: I am slightly opposed to suppressing the warning for L0 files

dstansby: Why? Doesn't the same logic for suppressing the warning for L1 files apply for L0 files?

ayshih: Here's the distinction that I'm drawing.  When generating the L1 files, the XRT team has decided that specifying `SOLAR_B0` and `DSUN_OBS` is sufficient observer metadata, and that the omission of `HGLN_OBS` shall be interpreted as `HGLN_OBS = 0`.  This PR implements that specific handling of an intentionally omitted keyword.  In contrast, the L0 files are lacking observer metadata entirely, so to me are still a bit raw rather than "analysis-ready".  Since our assuming that the observer is at Earth center injects information rather than fixes how we interpret existing information, it feels more dramatic and still deserving of a warning, even if the answer turns out to be the same as what the XRT team does when promoting L0 to L1.

changelog/6333.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+If an XRT file does not specify the heliographic longitude of the spacecraft,
```
dstansby, line 1:
```suggestion
If a level 1 XRT file does not specify the heliographic longitude of the spacecraft,
```

changelog/6333.bugfix.rst
```diff
@@ -0,0 +1,3 @@
...
+the default assumption is that the spacecraft is at zero Stonyhurst
+heliographic longitude (i.e., the same longitude as Earth).
```
dstansby, line 3:
```suggestion
heliographic longitude (i.e., the same longitude as Earth) is silently assumed.
```

sunpy/map/sources/hinode.py
```diff
@@ -21,6 +23,15 @@ class XRTMap(GenericMap):
 
     Hinode was launched on 22 September 2006 into a sun-synchronous orbit.
 
+    Notes
+    -----
+    XRT files do not normally specify the heliographic longitude of the spacecraft,
+    and the default assumption is that the spacecraft is at zero Stonyhurst
+    heliographic longitude (i.e., the same longitude as Earth). This assumption is
```
dstansby, line 30:
```suggestion
    so sunpy silently assumes is that the spacecraft is at zero Stonyhurst
    heliographic longitude (i.e., the same longitude as Earth) for L1 files. This assumption is
```

dstansby (changes requested):
If I'm correct we haven't added a test for this - I think we should add a L1 header and test map, and add a test to
- check the warning isn't being raised.
- check that if the longitude is in the file, that value is used instead of the deafult 0 deg value we've added

See also my suggestions to make the docstring/changelog clearer.

nabobalis:
Updated the test header to a Level 1 header and we don't have any errors raised with it as things stand with this PR.

changelog/6333.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+If a level 1 XRT file does not specify the heliographic longitude of the spacecraft,
+the default assumption is that the spacecraft is at zero Stonyhurst
+heliographic longitude (i.e., the same longitude as Earth) is silently assumed.
```
dstansby, line 3:
```suggestion
If a level 1 XRT file does not specify the heliographic longitude of the spacecraft,
a silent assumption is made that the spacecraft is at zero Stonyhurst
heliographic longitude (i.e., the same longitude as Earth).
```

dstansby (approved):

changelog/6333.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+If a level 1 XRT file does not specify the heliographic longitude of the spacecraft,
+a silent assumption is made that the spacecraft is at zero Stonyhurst
+heliographic longitude (i.e., the same longitude as Earth) is silently assumed.
```
ayshih, line 3:
```suggestion
heliographic longitude (i.e., the same longitude as Earth).
```
Heh, I don't know what you guys are doing, but https://github.com/sunpy/sunpy/pull/6328/commits/df075b73b5c1b373423bbf70777b0176546030a1 should not have been committed. =P

sunpy/map/sources/hinode.py
```diff
@@ -21,6 +23,15 @@ class XRTMap(GenericMap):
 
     Hinode was launched on 22 September 2006 into a sun-synchronous orbit.
 
+    Notes
+    -----
+    XRT files do not normally specify the heliographic longitude of the spacecraft,
+    so sunpy silently assumes is that the spacecraft is at zero Stonyhurst
+    heliographic longitude (i.e., the same longitude as Earth) for L1 files. This assumption is
```
ayshih, line 30:
```suggestion
    so sunpy silently assumes that the spacecraft is at zero Stonyhurst heliographic
    longitude (i.e., the same longitude as Earth) for L1 files. This assumption is
```

ayshih (approved):
Some docstring tweaks, otherwise looks good to me

PR #6333, opened 2022-07-14
parfive 2.0.0 whatsnew entry

nabobalis:


Cadair (dismissed):
This needs to remove the compatibility layer from utils/parfive.

nabobalis:
> This needs to remove the compatibility layer from utils/parfive.

fixed

sunpy/util/parfive_helpers.py
```diff
@@ -16,37 +15,12 @@
...
-            headers = kwargs.pop("headers", {})
-            kwargs["config"].headers = {**kwargs["config"].headers, **headers}
-
-            super().__init__(*args, **kwargs)
+class Downloader(parfive.Downloader):
+    @wraps(parfive.Downloader.__init__)
+    def __init__(self, *args, **kwargs):
+        if "config" not in kwargs:
+            kwargs["config"] = config
+        super().__init__(*args, **kwargs)
```
Cadair, line 26: Do we need this? I should check the built docs.

```suggestion
        super().__init__(*args, **kwargs)
        
Downloader.__doc__ = parfive.Downloader.__doc__
```

nabobalis: Let us find out.

dstansby (changes requested):
This violates [our dependency policy](https://docs.sunpy.org/en/latest/dev_guide/contents/dependencies.html), which says:

> Everything else [ie. not Python or astropy]: Released in the prior 24 months from the anticipated release date.

So I think this is a pretty hard block?

nabobalis:
We can just add parfive to that list.

nabobalis:
I can do that in this PR?

dstansby:
I really think we need to stick by our dependency support policiy, which means supporting back to `parfive` 1.2 in `sunpy` 4.1. If we want to violate that policy it needs:
- A good reason (which I don't see here - it wouldn't be too hard to support parfive 1 and 2 right?)
- Discussion and agreement among sunpy developers that we're going to make an exception to that policty

nabobalis:
At the weekly meeting, we decided we will do this for parfive 2.0.0 with no modification to our policy.

This is a one time deal for this version of parfive.

Cadair:
I am a little torn on this, the purist in me wants to follow the dependency policy, however I think this particular case is worth an exemption.

Most users wont even know what parfive is let alone that there's a release they should upgrade to, but this release of parfive comes with major usability improvements (removal of incomplete files and major error reporting upgrades). I think it's worth exempting this from our dependency policy because it will have a dramatic impact on how people experience the sunpy downloader. (If upgrading sunpy in place then doing this will force a parfive upgrade which will improve the experience).

We also discussed this at the community meeting today and the general consensus was to give the 2.0 release of parfive a pass on our dep policy.

dstansby:
Hmm, I'm still not sold on this bump, given users can always upgrade to parfive 2.0 if they want to get the new functionality (and we can even recommend this in the sunpy release notes). One of the major reasons to have a dependency policy is to allow users and downstream consumers to plan, and I think we erode trust if we violate it.

> dramatic impact on how people experience the sunpy downloader

I think dramatic is a bit extreme? `parfive` 1 works fine, but `parfive` 2 has some nice usability improvements.

Given 4.1 isn't being released until after the dev meeting, could I request we make a decision on this at the dev meeting?

Cadair:
You mean the coordination meeting?

dstansby:
Woops, yes.

dstansby:
I *think* we decided to still support parfive 1 at the coordination meeting and follow the dependency policy, but my mind is a bit fuzzy so if anyone disagrees with that perhaps we can discuss at a weekly meeting and document our decision as a comment here?

nabobalis:
This turns into a please update parfive to 2.0 in the whatsnew.

nabobalis:
This has now been updated to mention parfive 2.0.0 in the whatsnew, with no change to the min version of parfive. As thing stand, sunpy still supports both.

Cadair:
I think this is also worth calling out in the release announcements and stuff (i.e. when we give out pip install commands also add parfive)?

nabobalis:
> I think this is also worth calling out in the release announcements and stuff (i.e. when we give out pip install commands also add parfive)?

:+1:

Cadair (approved):

PR #6339, opened 2022-07-18
Check that adding .T to rotate fails a test

dstansby:
We discovered on Element chat (thanks `eimason`, https://matrix.to/#/!MeRdFpEonLoCwhoHeT:matrix.org/$16581713051370OFAQS:openastronomy.org?via=matrix.org&via=openastronomy.org&via=cadair.com onwards) that a bug found it's way into sunpy 3.1.7 during the backport of #5803 in #6000.

This PR is a test to see if the error that was introduced fails at least one of our tests. If it doesn't, we should definitely add a test that fails with this change...


dstansby:
👍 all good, lots of figure tests fail

PR #6340, opened 2022-07-19
[3.1] Fix really bad bug in .rotate

nabobalis:
Yes the branch name is wrong.


setup.cfg
```diff
@@ -48,9 +48,7 @@ dask =
 database =
   sqlalchemy>=1.3.4
 image =
-  # TODO: remove the <0.19 pin when discrepenies in
-  # sunpy/image/tests/test_transform.py::test_scale have been resolved
-  scikit-image>=0.16.0,<0.19
+  scikit-image>=0.16.0
```
nabobalis, line 51: Either this or we need to < numpy to below 1.23.0

tox.ini
```diff
@@ -66,7 +66,7 @@ deps =
     online: astroquery
     hypothesis: astroquery
     # Figure tests need a tightly controlled environment
-    figure-!devdeps: astropy==4.2.0
+    figure-!devdeps: astropy==5.0.1
```
nabobalis, line 69: 4.2.0 and recent numpys arent happy, so I decided to bump this instead.

setup.cfg
```diff
@@ -182,6 +181,10 @@ filterwarnings =
     ignore:FLIP_TOP_BOTTOM is deprecated and will be removed in Pillow 10:DeprecationWarning
     # Deprecation warning added in ImageIO 2.16.2, called by scikit-image
     ignore:Starting with ImageIO v3 the behavior of this function will switch:DeprecationWarning
+    ignore:overflow encountered in det:RuntimeWarning
+    ignore:`multichannel` is a deprecated argument name for `rescale`:FutureWarning
+    ignore:divide by zero encountered in double_scalars:RuntimeWarning
+    ignore:All-NaN slice encountered:RuntimeWarning
```
nabobalis, line 187: I am ignoring these but I don't know if they are important.

setup.cfg
```diff
@@ -24,6 +24,7 @@ classifiers =
   Programming Language :: Python :: 3.7
   Programming Language :: Python :: 3.8
   Programming Language :: Python :: 3.9
+  Programming Language :: Python :: 3.10
```
dstansby, line 27: Did we already support `3.10` with `3.1.x` and just miss this?

setup.cfg
```diff
@@ -182,6 +180,10 @@ filterwarnings =
     ignore:FLIP_TOP_BOTTOM is deprecated and will be removed in Pillow 10:DeprecationWarning
     # Deprecation warning added in ImageIO 2.16.2, called by scikit-image
     ignore:Starting with ImageIO v3 the behavior of this function will switch:DeprecationWarning
+    ignore:`multichannel` is a deprecated argument name for `rescale`:FutureWarning
+    ignore:divide by zero encountered in double_scalars
+    ignore:All-NaN slice encountered
+    ignore:ASDF functionality for astropy is being moved out of the astropy package
```
dstansby, line 186: Can we try and add these to individual tests instead of globally? That way it makes it easy to see that they're expected warnings in context.

sunpy/data/_sample.py
```diff
@@ -7,6 +7,7 @@
 from sunpy.util.parfive_helpers import Downloader
 
 _BASE_URLS = (
+    'https://github.com/sunpy/data/raw/main/sunpy/v1/',
```
dstansby, line 10: Does sample data work without making these changes? If so I think we should either leave the sample data alone, or if making these changes put them in a single commit with a meaningful commit message.

sunpy/util/parfive_helpers.py
```diff
@@ -4,27 +4,49 @@
 
 import aiohttp
 import parfive
+from packaging.version import Version
 from parfive import Results
 
 import sunpy
 
-__all__ = ['Downloader', 'Results']
```
dstansby, line 11: To avoid making lots of changes, I think we should just pin the major version of parfive to `<2` if it's not compatible with `3.1.x`.

dstansby (commented):
Given we're doing this I think it's prudent to make the set of changes absoutely minimal to avoid us tripping up on anything else, hence a lot of my review comments. Since we want to be careful I also think it pays to have a clean and clear commit history - could you clean it up so there's one change/commit with meaninful messages?

setup.cfg
```diff
@@ -116,7 +115,6 @@ asdf_schema_tests_enabled = true
 asdf_schema_root = sunpy/io/special/asdf/schemas/
 mpl-results-path = figure_test_images
 mpl-use-full-test-name = True
-asyncio_mode = auto
```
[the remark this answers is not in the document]
ConorMacBride: I think this ini option was added by mistake because someone had an extra non-sunpy pytest plugin installed which gave a warning if this ini option was missing.

sunpy/data/_sample.py
ConorMacBride: Without these sample data changes it should fall back to the data.sunpy.org endpoint when we rename the sample data repo.

Cadair: Given we are doing this, I would rather add this URL just to save me a few $ on the AWS bill. +1 on the history tho.

setup.cfg
```diff
@@ -48,9 +49,7 @@ dask =
 database =
   sqlalchemy>=1.3.4
 image =
-  # TODO: remove the <0.19 pin when discrepenies in
-  # sunpy/image/tests/test_transform.py::test_scale have been resolved
-  scikit-image>=0.16.0,<0.19
+  scikit-image>=0.16.0
```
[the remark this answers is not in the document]
dstansby: How so? Shouldn't we be able to run fine with the minimum version (0.16.0) here?

nabobalis:
The remaining CI fails are I think ok. We will bypass those builds for the release.

setup.cfg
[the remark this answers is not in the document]
nabobalis: Ill remove it again.

ConorMacBride:
Do these image differences all look fine? https://output.circle-artifacts.com/output/job/e250d8fb-ebf1-49cf-a9d4-7a61c8093682/artifacts/0/.tmp/py38-figure/figure_test_images/fig_comparison.html?f=image-diff&c=and

I realise the baselines for for 4.0, but are different croppings of rotations expected?

Also, I assume the oldestdeps failure is fine? https://github.com/sunpy/sunpy/runs/7417596641?check_suite_focus=true (This test was modified in #5867)

nabobalis:
The figure comparisions are with main I assume? 

I think that oldest deps failure is ok.

ConorMacBride:
> The figure comparisions are with main I assume?

Yes, that's what I meant sorry

> I think that oldest deps failure is ok.

Great!

nabobalis:
We will need several people to double check the figure tests figures are ok.

dstansby (approved):
Thanks for putting this together! I think the figure test differences are fine, and all the code changes look fine too. This should definnitely get a second approval before merging.

changelog/6340.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed :meth:`sunpy.map.GenericMap.rotate` incorrectly transposing twice.
```
ayshih, line 1:
```suggestion
Fixed an erroneous transpose of the image data by :meth:`sunpy.map.GenericMap.rotate`.
```
Not only does "transposing twice" not really accurately describe the bug, it reads silly since transposing twice would be the same as a no-op.

sunpy/image/tests/test_transform.py
```diff
@@ -67,14 +67,14 @@ def test_rotation(original, angle, k):
     expected = np.rot90(original, k=k)
 
     # Run the tests at order 4 as it produces more accurate 90 deg rotations
```
ayshih, line 69:
```suggestion
```

ayshih (approved):
Minor fixes needed, otherwise looks good to me

changelog/6340.bugfix.rst
nabobalis: Fair point!

nabobalis:
Thanks everyone, when the CI passes again, I will merge.

wtbarnes:
Since we're here, can we also pin parfive to `2.0`? I'm getting this exception with `2.0.x` when doing `Fido.fetch` on 3.1.7,

```python traceback
Input In [4], in <cell line: 1>()
----> 1 file = Fido.fetch(q)

File ~/anaconda/envs/eimason-sunpy/lib/python3.9/site-packages/sunpy/net/fido_factory.py:397, in UnifiedDownloaderFactory.fetch(self, path, max_conn, progress, overwrite, downloader, *query_results, **kwargs)
    395         max_conn = 1
    396         kwargs['max_splits'] = 1
--> 397     downloader = Downloader(max_conn=max_conn, progress=progress, overwrite=overwrite)
    398 elif not isinstance(downloader, parfive.Downloader):
    399     raise TypeError("The downloader argument must be a parfive.Downloader instance.")

File ~/anaconda/envs/eimason-sunpy/lib/python3.9/site-packages/sunpy/util/parfive_helpers.py:30, in Downloader.__init__(self, *args, **kwargs)
     27 if headers:
     28     kwargs['headers'] = headers
---> 30 super().__init__(*args, **kwargs)

TypeError: __init__() got an unexpected keyword argument 'headers'
```

nabobalis:
I thought I already pinned it?

dstansby (approved):

dstansby:
This is good, but shouldn't it be targeting the `3.1` branch? Currently targeting `3.0`, and I can't see `3.1` in the list of options...

nabobalis:
This is the 3.1 branch.

dstansby:
Eh, why's it called `3.0` then?

PR #6343, opened 2022-07-28
Wholesale refactor of the map source tests

nabobalis:
Moved all the tests for the sources to follow how they are laid out under the source folder aka per mission/satellite/observatory. 

The only thing that has changed from current main is that:

1. Added tests for Yohkoh
2. This has lead to two changes to that map source file. 

Otherwise everything else is just moving tests and some doc tweaks.


sunpy/map/sources/yohkoh.py
```diff
@@ -57,23 +57,22 @@ def detector(self):
...
-        # the Sun from the spacecraft.  Note that the small angle approximation
+        # the Sun from the spacecraft. Note that the small angle approximation
         # is used, and the solar radius stored in SXT FITS files is in arcseconds.
-
         if 'solar_r' in self.meta:
             dsun = constants.radius / (np.deg2rad(self.meta['solar_r'] / 3600.0))
         else:
             dsun = constants.au
-        return self.meta.get('dsun_obs', dsun)
+        return self.meta.get('dsun_obs', dsun) * u.m
```
nabobalis, line 75: I had to do this.

sunpy/map/sources/yohkoh.py
```diff
@@ -89,8 +88,15 @@ def wavelength(self):
...
     def is_datasource_for(cls, data, header, **kwargs):
-        """Determines if header corresponds to an SXT image"""
+        """
+        Determines if header corresponds to an SXT image.
+        """
         return header.get('instrume') == 'SXT'
+
+    @property
+    def date(self):
+        return Time(self.meta.get('DATE_OBS'))
```
nabobalis, line 102: I had to do this.

sunpy/map/sources/yohkoh.py
ayshih: Er, neither is correct.  `self.meta.get('dsun_obs')` is a bare value (in meters), while `dsun` is already a `Quantity`.

dstansby:
General thoughts on this PR:
- Can we get rid of the whitespace changes in the header files? They seem unnesseseary and make the PR harder to review
- Can you put any behaviour changes or bugfixes in individual PRs, leaving this on purely for refactoring? That way we can much more easily keep track of changelogs/backports etc.

Cadair:
The whitespace changes in the header files might actually be bad (I haven't looked closely) but a real FITS header always has 80 char per card padded with whitespace. So I would rather keep them as close to "copied out of file" as possible.

sunpy/map/sources/yohkoh.py
nabobalis: I will fix this.

nabobalis:
> The whitespace changes in the header files might actually be bad (I haven't looked closely) but a real FITS header always has 80 char per card padded with whitespace. So I would rather keep them as close to "copied out of file" as possible.

I can undo the ones I changed but the ones I created don't have them.

nabobalis:
> General thoughts on this PR:
> 
> * Can we get rid of the whitespace changes in the header files? They seem unnesseseary and make the PR harder to review

Will do

> * Can you put any behaviour changes or bugfixes in individual PRs, leaving this on purely for refactoring? That way we can much more easily keep track of changelogs/backports etc.

The only bugfix I can see is the yohkoh if that even counts.

dstansby:
Yeah, it was the changes you pointed out in your own comments that I thought deserved separate PR(s?)

nabobalis:
I would rather they stay here since they are one line changes.

sunpy/map/sources/yohkoh.py
nabobalis: I suspect this is a bug in mapbase but this was quicker.

sunpy/map/sources/yohkoh.py
nabobalis: What do you think about the comment above it? 

If that is true, we don't in fact want to return the meta keyword value at all?

nabobalis:
I will come back for these header files in the future.

sunpy/map/sources/yohkoh.py
```diff
@@ -57,23 +56,22 @@ def detector(self):
...
-        # the Sun from the spacecraft.  Note that the small angle approximation
+        # the Sun from the spacecraft. Note that the small angle approximation
         # is used, and the solar radius stored in SXT FITS files is in arcseconds.
-
         if 'solar_r' in self.meta:
             dsun = constants.radius / (np.deg2rad(self.meta['solar_r'] / 3600.0))
         else:
             dsun = constants.au
-        return self.meta.get('dsun_obs', dsun)
+        return dsun
```
nabobalis, line 74: To align with the comment, I decided to calculate the value and return that instead.

ayshih: Well...  The test header has a usable (and presumably) correct value for `DSUN_OBS`...

ayshih: I think we should get rid of this overridden `dsun` property entirely.  As currently implemented, `.dsun` is not used to construct the observer coordinate, so the change of inconsistency is high.  If `DSUN_OBS` is missing, we need `.dsun` to give the same answer as `.observer_coordinate.radius`.  If `DSUN_OBS` is present but wildly wrong, that error needs to be addressed during the creation of the observer coordinate, not in this property.

ayshih:
Bleh, these values in the SXT header ought to be equal:
```
SOLAR_B0=              3.06603
HGLT_OBS=              0.00000
```
`HGLT_OBS` is clearly incorrect.  We'll need to override `._supported_observer_coordinates()`.

ayshih:
Something like this:
```python
    @property
    def _supported_observer_coordinates(self):
        return [(('hgln_obs', 'solar_b0', 'dsun_obs'), {'lon': self.meta.get('hgln_obs'),
                                                        'lat': self.meta.get('solar_b0'),
                                                        'radius': self.meta.get('dsun_obs'),
                                                        'unit': (u.deg, u.deg, u.m),
                                                        'frame': "heliographic_stonyhurst"})]
            + super()._supported_observer_coordinates
```

dstansby:
I'm going to request again that any bug fixes or behaviour changes go in individual PRs with appropriate names; the overheads trying to review a PR that contains refactoring and changes are large, make it less likely for reviewers to review code, and much easier for bugs to slip through the cracks during review.

nabobalis:
The only code changes are in the Yohkoh map file. That I think is isolated enough to not need pulling out.

nabobalis:
I think since we need to backport this and we didn't backport the test files changes, I am wondering if it's time to backport those now.

sunpy/map/sources/yohkoh.py
nabobalis: This is all gone now and using your suggested fixes.

sunpy/map/mapbase.py
```diff
@@ -786,10 +786,9 @@ def nickname(self, n):
         self._nickname = n
 
     def _get_date(self, key):
-        time = self.meta.get(key, None)
-        if time is None:
+        time = self.meta.get(key)
+        if not time:
```
nabobalis, line 790: This might be a bug?

sunpy/map/mapbase.py
```diff
@@ -786,10 +786,9 @@ def nickname(self, n):
...
             return
-
```
dstansby, line 792: What's the context for this change?

dstansby (changes requested):
I'm going to mark this as changes again because:
- We haven't agreed to apply black to the codebase, and there's lots of that in this PR. @nabobalis if you're intent on doing this, please can we have clearly labelled commit messages that *only* apply black and nothing else?
- I'm going to insist that any changes to actual code that aren't refactoring go in their own PRs for visibility and ease of review. In this case I think that's just the Yokoh changes

dstansby:
Sorry, I should have said codestyle changes in general, not just black.

nabobalis:
I can remove actual code changes to a specific PR but the rest of the PR is a code refactor, it is going to look different and there is no simple way to commit just that

dstansby:
I'm not against codestyle changes in general, but it's really tricky to review when they're mixed in with other changes. So commits specifically for code style updates that have "Codestyle updates" or simliar in the commit message, and are *just* codestyle updates make it much easier to review stuff (and bisect any issues in the future).

sunpy/map/mapbase.py
nabobalis: The key could be empty but not None.

dstansby:
Just coming back to this, and I think it's actually better to have one test file per source, so it's easier to check that each map source is being tested in the same way. Long term it might be nicer to parametrize all the map source tests in one file. So I'd advocate for closing this without merging, and if there's small bits that aren't re-organising tests putting them in new PRs, thoughts?

dstansby:
Sorry, I don't follow - do the same what for all map sources?

nabobalis:
Either the map sources should be one class per file or the tests should be collected together to match the map sources files

dstansby:
My preference would be single file per spacecraft, which would mean merging the test files. If we go this way, tests for each map class should live within their own test class.

dstansby:
Worth closing this PR and redoing it in a new (or several new) PR if it's still relevant?

nabobalis:
I will handle it.

PR #6362, opened 2022-08-15
Implement a cgi alive test when determining what VSO endpoint to use

### LINKED ISSUE #6361: When deciding which VSO endpoint to connect to we should determine if the soap address in the WDSL is accessible
By default we access the WDSL at http://docs.virtualsolar.org/WSDL/VSOi_rpc_literal.wsdl but the soap endpoint is http://vso.nso.edu/cgi-bin/VSO/PROD/vsoi_wsdl.cgi we should test both somehow

Cadair:
fixes #6361 

I have no idea how to test this, as we are testing that a url in a remote file is accessible. I guess the only way would be to trick the function into downloading a pre-faked WSDL file.

### DIFF
--- a/sunpy/net/vso/tests/test_vso.py
+++ b/sunpy/net/vso/tests/test_vso.py
@@ -1,4 +1,7 @@
+from functools import partial
+from xml.etree import ElementTree
 from urllib.error import URLError, HTTPError
+from urllib.request import urlopen
 
 import pytest
 from parfive import Results
@@ -12,6 +15,7 @@ from sunpy.net.vso import attrs as va
 from sunpy.net.vso.legacy_response import QueryResponse
 from sunpy.net.vso.table_response import VSOQueryResponseTable, iter_sort_response
 from sunpy.net.vso.vso import (
+    DEFAULT_URL_PORT,
     VSOClient,
     build_client,
     check_cgi_connection,
@@ -27,8 +31,16 @@ class MockQRRecord:
     """
     Used to test sunpy.net.vso.QueryResponse.build_table(...)
     """
-    def __new__(cls, start_time=None, end_time=None, size=0, source='SOHO', instrument='aia',
-                extent=None, fileid="spam"):
+    def __new__(
+        cls,
+        start_time=None,
+        end_time=None,
+        size=0,
+        source='SOHO',
+        instrument='aia',
+        extent=None,
+        fileid="spam"
+    ):
         return MockObject(size=size,
                           time=MockObject(start=start_time, end=end_time),
                           source=source,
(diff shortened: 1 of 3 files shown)


sunpy/net/vso/vso.py
```diff
@@ -61,12 +61,37 @@ def check_connection(url):
...
+    This is weird enough that it probably satisfies us for this check.
+    """
+    try:
+        return urlopen(url).getcode() == 411
+    except HTTPError as e:
+        if e.code == 411:
+            return True
+        warn_user(f"Connection to {url} failed with error {e}. Retrying with different url and port.")
+        return None
+    except (socket.error, socket.timeout, HTTPError, URLError) as e:
```
Cadair, line 77:
```suggestion
    except (socket.error, socket.timeout, URLError) as e:
```

sunpy/net/vso/vso.py
```diff
@@ -61,12 +61,37 @@ def check_connection(url):
...
 
 
+def check_cgi_connection(url):
+    """
+    At the moment there is no way to make an "are you alive" request to the
+    cgi, so we just hit it with a HTTP get and it gives us back a 411 response.
+    This is weird enough that it probably satisfies us for this check.
+    """
+    try:
+        return urlopen(url).getcode() == 411
```
alasdairwilson, line 71:
```suggestion
        return urlopen(url, timeout=10).getcode() == 411
```

We probably don't want to wait the 5 minute default timeout?

alasdairwilson:
Thanks for this, very edge case but frustrating to encounter. 👍

sunpy/net/vso/vso.py
```diff
@@ -40,33 +40,63 @@
...
+    At the moment there is no way to make an "are you alive" request to the
+    cgi, so we just hit it with a HTTP get and it gives us back a 411 response.
+    This is weird enough that it probably satisfies us for this check.
+    """
+    try:
+        return urlopen(url, timeout=15).getcode() == 411
+    except HTTPError as e:
+        if e.code == 411:
+            return True
+        warn_user(f"Connection to {url} failed with error {e}. Retrying with different url and port.")
```
dstansby, line 80: I think this error message is outside the scope of this function, as whatever calls it could do something else instead of retyring with a different url and port?

sunpy/net/vso/vso.py
```diff
@@ -40,33 +40,63 @@
...
+        return None
+    except (socket.error, socket.timeout, URLError) as e:
+        warn_user(f"Connection to {url} failed with error {e}. Retrying with different url and port.")
```
dstansby, line 83: Similarly for this message

sunpy/net/vso/vso.py
```diff
@@ -40,33 +40,63 @@
...
-        return urlopen(url).getcode() == 200
+        return urlopen(url, timeout=15).getcode() == 200
     except (socket.error, socket.timeout, HTTPError, URLError) as e:
         warn_user(f"Connection to {url} failed with error {e}. Retrying with different url and port.")
         return None
 
 
+def check_cgi_connection(url):
+    """
+    At the moment there is no way to make an "are you alive" request to the
```
dstansby, line 71: Could you quickly document in the docstring what returning `None` means here, as opposed to `True` or `False`?

dstansby (changes requested):

sunpy/net/vso/vso.py
Cadair: These messages are here because they need the exception context to be able to display the error to the user (as the exception is not raised).

dstansby: Doesn't returning `False` provide the context for calling code to raise these warning messages instead (if they are indeed retrying with a different URL)?

sunpy/net/vso/vso.py
Cadair: I made it return False, I was just copying the previous implementation, but anything Falsey works.

dstansby:
Nudge - I'm happy to try and finish this up @Cadair if you don't have time?

sunpy/net/vso/vso.py
Cadair: The calling context wouldn't have the exception instance which has the reason why that URL erred out.

nabobalis:
I wasn't sure what was left for this but I rebased and added some mock tests for the new functions.

I think a better mocked test would be to make the first urlopen fail then see if it connects to a second URL later on as a real remote test.

Cadair:
The online test fails are unrelated

sunpy/net/vso/tests/test_vso.py
```diff
@@ -9,17 +14,33 @@
...
 
 
 class MockQRRecord:
     """
     Used to test sunpy.net.vso.QueryResponse.build_table(...)
     """
-    def __new__(cls, start_time=None, end_time=None, size=0, source='SOHO', instrument='aia',
-                extent=None, fileid="spam"):
+    def __new__(
+        cls,
```
nabobalis, line 35: What did this?

Cadair: I did

nabobalis (approved):

nabobalis:
LGTM, shall we merge?

PR #6366, opened 2022-08-16
pin version of jsonschema

nabobalis:


ConorMacBride:
Should we just wait for the fix in asdf? https://github.com/asdf-format/asdf/issues/1172

nabobalis:
My feeling is that https://github.com/python-jsonschema/jsonschema/issues/981 is going to take some time, so we might as well restore our CI until then.

nabobalis:
Reverting this PR tomorrow or in 6 hours won't take too much effort either.

ConorMacBride:
Yeah, it might take some time. I didn't see the jsonschema issue

ConorMacBride (approved):

nabobalis:
Thanks.

Cadair:
My understanding of test package pins was that there was only one copy on main that everything uses so this wouldn't have needed back porting. However, I see that's not true. If we change the tox.ini file to use the URL version of the file (like we do in ndcube) then we only need to maintain a single version of this?

nabobalis:
Right, we should do that.

WilliamJamieson:
ASDF 2.12.1 and 2.11.2 have been released which pin jsonschema inside them. You may want to update your pins to use them now.

nabobalis:
> ASDF 2.12.1 and 2.11.2 have been released which pin jsonschema inside them. You may want to update your pins to use them now.

Thanks

PR #6370, opened 2022-08-17
Update .test_package_pins.txt

nabobalis:
Asdf pins it now.

### DIFF
--- a/.test_package_pins.txt
+++ b/.test_package_pins.txt
@@ -1,1 +1,1 @@
-jsonschema<4.10.0
+


ConorMacBride (approved):
Great! We should probably update the tox config to use the URL rather than the file in this PR (and also backport)? https://github.com/sunpy/ndcube/blob/485a0551b35f1113398ceb9b0fcd38807288b488/tox.ini#L13

nabobalis:
I made the tox file on 4.0 point to the file on main. So no need to backport.

nabobalis:
Do we need to change the one for main to point to the URL, the file is in the repo and I think would be useful to point to that.

ConorMacBride:
I missed the update to 4.0. That should be fine then. A benefit of updating the main branch to the url would be so people don't have to pull to their local sunpy repo to get the latest tox pins, but that's an unlikely issue.

nabobalis:
> I missed the update to 4.0. That should be fine then. A benefit of updating the main branch to the url would be so people don't have to pull to their local sunpy repo to get the latest tox pins, but that's an unlikely issue.

This is fair, I think if wanto make sure it works on a PR tho to main, we would need it to point to the local file.

PR #6385, opened 2022-08-25
Remove tick rotation and grids from timeseries plotting

dstansby:
Part of my ongoing quest to simplify timeseries plotting, this removes:
- Rotating tick labels
- Drawing grids

From timeseries source files. I also put in a what's new entry to cover a number of PRs that have already or will simplify timeseries plotting.


sunpy/timeseries/sources/lyra.py
```diff
@@ -123,11 +120,7 @@ def peek(self, title=None, columns=None, names=3, **kwargs):
         axes = self.plot(columns=columns, names=names, **kwargs)
         if title is None:
             title = "LYRA ({0:{1}})".format(self.to_dataframe().index[0], TIME_FORMAT)
-        axes[0].set_title(title)
-        fig = axes[-1].get_figure()
-        fig.subplots_adjust(left=0.17, top=0.94, right=0.94, bottom=0.15)
-        fig = axes[0].get_figure()
-        fig.subplots_adjust(left=0.17, top=0.94, right=0.94, bottom=0.15)
+            axes[0].set_title(title)
```
nabobalis, line 123:
```suggestion
          axes[0].set_title(title)
```
Unless the title is already set?

wtbarnes:
I have no objections to any of this, but could I just ask for a bit more clarification about what the motivation here is? Is this just a simplification of the plotting code and an attempt to be less opinionated about how we make these plots?

nabobalis:
I guess the idea is to make all plot methods as simple and un-opinionated as possible? 

I forgot why we did those to the plot methods but I suspect it was the fact that the default plots (the sample data example timeseries for example) looked kind of bad and we wanted to avoid people thinking sunpy makes ugly plots.

nabobalis (approved):

nabobalis:
> I have no objections to any of this, but could I just ask for a bit more clarification about what the motivation behind this PR is? Is this just a simplification of the plotting code and an attempt to be less opinionated about how we make these plots?

It also makes them the same, we only rotated some of the axis labels and some had grids. Now all the current timeseries plotting looks the same.

dstansby:
> I have no objections to any of this, but could I just ask for a bit more clarification about what the motivation behind this PR is? Is this just a simplification of the plotting code and an attempt to be less opinionated about how we make these plots?

Yes, exactly. If nothing else for maintainability purposes I'm trying to slim the plotting code down so source-specific customization is a minimum. I think it's also good from a user point of view to make consistent and minimally-altered-from-the-defaults plots across sources.

PR #6386, opened 2022-08-25
Don't run CI if only a changelog has changed

dstansby:
It's pretty common to make a PR, and then add a changelog in a second commit (since you need the PR number before you make a changelog).

This PR stops the CI from running on a commit that's just a changelog being added, to prevent our CI being clogged up by unneeded runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,8 @@ name: CI
 
 on:
   push:
+    paths-ignore:
+        - 'changelog/**'
     branches:
       - '*'
       - '!*backport*'
--- a/changelog/6386.trivial.rst
+++ b/changelog/6386.trivial.rst
@@ -1,0 +1,1 @@
+CI is no longer run on commits when only a changelog is modified.


dstansby:
I will test this by pushing a changelog-only commit in a bit

nabobalis:
The CI should be canceled on the previous commit which means if it doesn't run fully before push of the changelog, we don't know if the past commit passes to merge.

dstansby:
> The CI should be canceled on the previous commit which means if it doesn't run fully before push of the changelog, we don't know if the past commit passes to merge.

Boooooo this is indeed true

.github/workflows/ci.yml
```diff
@@ -2,6 +2,8 @@ name: CI
 
 on:
   push:
+    paths-ignore:
+        - 'changelog/**'
```
ConorMacBride, line 6:
```suggestion
      - 'changelog/**'
```

ConorMacBride (commented):
This looks fine and it would usually be good to skip in this case. However, it does mean a changelog can be merged without having its references tested in the offline docs build.

nabobalis:
> This looks fine and it would usually be good to skip in this case. However, it does mean a changelog can be merged without having its references tested in the offline docs build.

So it won't auto cancel the previous commit build?

ConorMacBride:
Currently committing just a changelog will auto cancel the previous GitHub Actions jobs. Maybe just close this PR?

nabobalis:
If we can work out how to prevent that in future, that would be good.

PR #6388, opened 2022-08-26
Fix the ordering of the API docs

Cadair:
I don't know why we don't have the top level package at the top?


nabobalis (approved):

nabobalis:
Just want to make sure the doc build works as adding a interlink could go horribly.

PR #6384, opened 2022-08-24
Make sure data examples achieve a single task

dstansby:
As part of the documentation reorganisation we will have the four types of documentation described at https://diataxis.fr/. One of these is "How-to guides", which are 
> recipes, directions that guide the reader through the steps to achieve a specific end.

We already have these guides to some extent in our example gallery.

This PR is the first part of work to make sure the files in our example gallery achieve **one** specific task. I don't think any of the delted content in this PR needs to be moved elsewhere as it's covered by other examples, but in PRs for the other example categories we may want to move some stuff to entries in a Tutorials section.

Fixes #5888

PR #6393, opened 2022-08-30
Make sure non-data examples each achieve a single task

dstansby:
This is a follow on to https://github.com/sunpy/sunpy/pull/6384 to finish making sure our examples are only scoped to achieving a single task.

The main change is either moving examples between folders, or moving some examples out to new folders `examples_tutorials` and `examples_explanation` depending on which "quadrant" of the new doc structure they belong in. A future PR can then integrate these into whatever method we use for writing the tutorials and the explanations.


dstansby:
Going to close this as there's lots of conflicts

PR #6394, opened 2022-08-31
Use WCS API in `contains_coordinate`

wtbarnes:
This PR makes a small change to the internals of `contains_coordinate` to use the API of `GenericMap.wcs` rather than that on `GenericMap` directly. The advantage of this is that functions like `contains_coordinate` and `extract_along_coord` can be used with objects with a `.wcs` property with two celestial axes (e.g. an `NDCube`) and not just `GenericMap`.

### DIFF
--- a/sunpy/map/maputils.py
+++ b/sunpy/map/maputils.py
@@ -404,9 +404,11 @@ def contains_coordinate(smap, coordinates):
         returns a boolean arrary.
     """
     # Dimensions of smap
-    xs, ys = smap.dimensions
+    ys, xs = smap.wcs.array_shape * u.pix
     # Converting coordinates to pixels
-    xc, yc = smap.world_to_pixel(coordinates)
+    xc, yc = smap.wcs.world_to_pixel(coordinates)
+    xc = xc * u.pix
+    yc = yc * u.pix
     point5pix = 0.5 * u.pix
     return ((xc >= -point5pix) &
             (xc <= xs - point5pix) &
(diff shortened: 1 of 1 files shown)


dstansby (approved):
Happy to approve since this is a trivial change, but if we explicitly want to support the use case of `NDCube` this should get docs and a test to prevent it breaking in the future.

sunpy/map/maputils.py
```diff
@@ -404,14 +404,13 @@ def contains_coordinate(smap, coordinates):
...
+    ys, xs = smap.wcs.array_shape
     # Converting coordinates to pixels
-    xc, yc = smap.world_to_pixel(coordinates)
-    point5pix = 0.5 * u.pix
-    return ((xc >= -point5pix) &
-            (xc <= xs - point5pix) &
-            (yc >= -point5pix) &
-            (yc <= ys - point5pix))
+    xc, yc = smap.wcs.world_to_pixel(coordinates)
+    return ((xc >= -0.5) &
```
Cadair, line 410: Hasn't this changed the return from being Quantity with units of pix to floats?

Cadair (changes requested):
Definitely pro this, and generally removing all our internal use of `GenericMap.world_to_pixel` and `pixel_to_world` in favour of the `.wcs` alternatives.

sunpy/map/maputils.py
dstansby: The return is `bool` no? (adding this to my log of "reasons to add typing" 😉 )

wtbarnes: Yes it returns a boolean array

Cadair (approved):

sunpy/map/maputils.py
Cadair: I mean I could have read the docstring as well :rofl:

PR #6405, opened 2022-09-08
Prevent double CI on branch and open PR

nabobalis:
It happens sometimes. 

I plan to roll this out to each repo with actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,8 @@ name: CI
 on:
   push:
     branches:
-      - '*'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -3,7 +3,8 @@ name: CI
 on:
   push:
     branches:
-      - '*'
+      - 'main'
+      - '*.*'
```
nabobalis, line 7: This should only trigger for release branches as well.

dstansby (approved):

PR #6406, opened 2022-09-09
Make all kwargs to `reproject_to` kwarg only

Cadair:
While working on fixing compatibility with reproject_to in ndcube I have come to the conclusion that we really should be kwarg only here, so that the ordering of the arguments inside or outside of `**reproject_args` is not relevant to our API compatibility.

### DIFF
--- a/changelog/6406.deprecation.rst
+++ b/changelog/6406.deprecation.rst
@@ -1,0 +1,1 @@
+The ``algorithm``, ``return_footprint`` and any keyword arguments passed through to the underlying `reproject` functions now must be passed as keyword arguments.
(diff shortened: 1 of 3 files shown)


changelog/6406.deprecation.rst
```diff
@@ -0,0 +1 @@
+The ``algorithm``, ``return_footprint`` and any keyword arguments passed through to the underlying `reproject` functions now must be passed as keyword arguments.
```
dstansby, line 1:
```suggestion
Passing the ``algorithm``, ``return_footprint`` arguments as positional arguments is deprecated. Pass them as keyword arguments (e.g. ``..., return_footprint=True, ...``) instead.
```

dstansby (changes requested):
This should also get a quick test for the deprecations.

nabobalis (approved):

PR #6415, opened 2022-09-20
Add helper for creating Carrington header

dstansby:
This adds a helper function to make it easier to create a FITSWCS header for a Carrington coordinate frame. Even though we already have the header helper function, reprojecting to whole Sun maps is a common enough use case that I think it deserves its own function.

To see how the API is simpler, look at the diff between the example I've renamed and modified to use a Carrington (instead of Stonyhurst) system. I am also imagining there is room for creating other commonly used headers. The one that jumps to mind at the moment is a something to reproject into a polar system like this: http://swhv.oma.be/user_manual/projection_2.png

There was some discussion about whether to put this in `sunkit-image` (https://github.com/sunpy/sunkit-image/pull/93), but @Cadair suggested this *might* belong in core, and I'm inclined to agree.

I would appreciate some either positive or negative feedback on whether this should be added as new API!

If feedback is posistive this still needs:
- A changelog entry
- A what's new entry
- Some more tests? Suggestions welcome...

### DIFF
--- a/docs/guide/data_types/maps.rst
+++ b/docs/guide/data_types/maps.rst
@@ -492,7 +492,7 @@ See this :ref:`sphx_glr_generated_gallery_map_map_from_numpy_array.py` for a bri
 
 The keys required for the header information follow the `FITS standard <https://fits.gsfc.nasa.gov/fits_dictionary.html>`__.
 **sunpy** provides a Map header helper function to assist in creating a header that contains the correct meta information.
-This includes a `~sunpy.map.header_helper.meta_keywords` function that will return a `dict` of the meta keywords used when creating a Map.
+This includes a :func:`~sunpy.map.header_helper.meta_keywords` function that will return a `dict` of the meta keywords used when creating a Map.
 
 .. code-block:: python
 
@@ -504,12 +504,12 @@ This includes a `~sunpy.map.header_helper.meta_keywords` function that will retu
      'crval1': 'Coordinate value at reference point on naxis1 **required'
      ...
 
-The utility function `~sunpy.map.header_helper.make_fitswcs_header` will return a header with the appropriate FITS keywords once the Map data array and an `astropy.coordinates.SkyCoord` or `sunpy.coordinates.frames` is provided.
+The utility function :func:`~sunpy.map.header_helper.make_fitswcs_header` will return a header with the appropriate FITS keywords once the Map data array and an `astropy.coordinates.SkyCoord` or `sunpy.coordinates.frames` is provided.
 The `astropy.coordinates.SkyCoord` is defined by the user and contains information on the reference frame, reference coordinate, and observer location.
 This function returns a `sunpy.util.MetaDict`.
 The `astropy.coordinates.SkyCoord` or `sunpy.coordinates.frames` must contain an observation time.
 
-The `~sunpy.map.header_helper.make_fitswcs_header` function also takes optional keyword arguments including ``reference_pixel`` and ``scale`` that describe the pixel coordinate at the reference coordinate (defined by the `~astropy.coordinates.SkyCoord`) and the spatial scale of the pixels, respectively.
+The :func:`~sunpy.map.header_helper.make_fitswcs_header` function also takes optional keyword arguments including ``reference_pixel`` and ``scale`` that describe the pixel coordinate at the reference coordinate (defined by the `~astropy.coordinates.SkyCoord`) and the spatial scale of the pixels, respectively.
 If neither of these are given their values default to the center of the data array and 1 arcsec, respectively.
 
 Here's an example of creating a header from some generic data and an `astropy.coordinates.SkyCoord`:
@@ -598,8 +598,8 @@ Here's another example of passing ``reference_pixel`` and ``scale`` to the funct
     rsun_obs: 965.3829548285768
 
 As we can see, a list of WCS and observer meta information is contained within the generated headers, however we may want to include other meta information including the observatory name, the wavelength and waveunit of the observation.
-Any of the keywords listed in ``header_helper.meta_keywords`` can be passed to the `~sunpy.map.header_helper.make_fitswcs_header` and will then populate the returned MetaDict header.
-Furthermore, the following observation keywords can be passed to the `~sunpy.map.header_helper.make_fitswcs_header` function: ``observtory``, ``instrument``, ``telescope``, ``wavelength``, ``exposure``.
+Any of the keywords in the dictionary returned by :func:`~sunpy.map.header_helper.meta_keywords` can be passed to the :func:`~sunpy.map.header_helper.make_fitswcs_header` and will then populate the returned MetaDict header.
+Furthermore, the following observation keywords can be passed to the `~sunpy.map.header_helper.make_fitswcs_header` function: ``observatory``, ``instrument``, ``telescope``, ``wavelength``, ``exposure``.
 
 An example of creating a header with these additional keywords:
 
--- a/docs/guide/data_types/maps.rst
+++ b/docs/guide/data_types/maps.rst
@@ -492,11 +492,11 @@ See this :ref:`sphx_glr_generated_gallery_map_map_from_numpy_array.py` for a bri
 
 The keys required for the header information follow the `FITS standard <https://fits.gsfc.nasa.gov/fits_dictionary.html>`__.
 **sunpy** provides a Map header helper function to assist in creating a header that contains the correct meta information.
-This includes a `~sunpy.map.meta_keywords` function that will return a `dict` the meta keywords used when creating a Map.
+This includes a `~sunpy.map.header_helper.meta_keywords` function that will return a `dict` the meta keywords used when creating a Map.
 
 .. code-block:: python
 
-    >>> from sunpy.map import meta_keywords
+    >>> from sunpy.map.header_helper import meta_keywords
 
     >>> meta_keywords() # doctest: +SKIP
     {'cunit1': 'Units of the coordinate increments along naxis1 e.g. arcsec **required',
@@ -504,12 +504,12 @@ This includes a `~sunpy.map.meta_keywords` function that will return a `dict` th
      'crval1': 'Coordinate value at reference point on naxis1 **required'
      ...
 
-The utility function `~sunpy.map.make_fitswcs_header` will return a header with the appropriate FITS keywords once the Map data array and an `astropy.coordinates.SkyCoord` or `sunpy.coordinates.frames` is provided.
+The utility function `~sunpy.map.header_helper.make_fitswcs_header` will return a header with the appropriate FITS keywords once the Map data array and an `astropy.coordinates.SkyCoord` or `sunpy.coordinates.frames` is provided.
 The `astropy.coordinates.SkyCoord` is defined by the user and contains information on the reference frame, reference coordinate, and observer location.
 This function returns a `sunpy.util.MetaDict`.
 The `astropy.coordinates.SkyCoord` or `sunpy.coordinates.frames` must contain an observation time.
 
-The `~sunpy.map.make_fitswcs_header` function also takes optional keyword arguments including ``reference_pixel`` and ``scale`` that describe the pixel coordinate at the reference coordinate (defined by the `~astropy.coordinates.SkyCoord`) and the spatial scale of the pixels, respectively.
+The `~sunpy.map.header_helper.make_fitswcs_header` function also takes optional keyword arguments including ``reference_pixel`` and ``scale`` that describe the pixel coordinate at the reference coordinate (defined by the `~astropy.coordinates.SkyCoord`) and the spatial scale of the pixels, respectively.
 If neither of these are given their values default to the center of the data array and 1 arcsec, respectively.
 
 Here's an example of creating a header from some generic data and an `astropy.coordinates.SkyCoord`:
@@ -517,13 +517,15 @@ Here's an example of creating a header from some generic data and an `astropy.co
 .. code-block:: python
 
     >>> import numpy as np
-    >>> import astropy.units as u
-    >>> from sunpy.coordinates import frames
     >>> from astropy.coordinates import SkyCoord
+    >>> import astropy.units as u
+    >>>
+    >>> from sunpy.coordinates import frames
+    >>> from sunpy.map.header_helper import make_fitswcs_header
 
     >>> data = np.arange(0,100).reshape(10,10)
     >>> coord = SkyCoord(0*u.arcsec, 0*u.arcsec, obstime = '2013-10-28', observer = 'earth', frame = frames.Helioprojective)
-    >>> header = sunpy.map.make_fitswcs_header(data, coord)
+    >>> header = make_fitswcs_header(data, coord)
     >>> for key, value in header.items():
     ...     print(f"{key}: {value}")
     wcsaxes: 2
@@ -562,7 +564,7 @@ Here's another example of passing ``reference_pixel`` and ``scale`` to the funct
 
 .. code-block:: python
 
-    >>> header = sunpy.map.make_fitswcs_header(data, coord,
+    >>> header = make_fitswcs_header(data, coord,
     ...                                        reference_pixel=u.Quantity([5, 5]*u.pixel),
     ...                                        scale=u.Quantity([2, 2] *u.arcsec/u.pixel))
     >>> for key, value in header.items():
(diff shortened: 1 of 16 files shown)


sunpy/map/header_helper.py
```diff
@@ -361,3 +361,52 @@ def _set_instrument_meta(meta_wcs, instrument, telescope, observatory, wavelengt
     'CD2_2':
     'Matrix element CDi_j describing the rotation required to align solar North with the top of the image.'
 }
+
+
+def carrington_header(date, observer_coordinate, *, shape_out, projection_code="CAR"):
```
ayshih, line 366:
```suggestion
def carrington_header(date, observer_coordinate, *, dimensions, projection_code="CAR"):
```
I suggest changing the name for this keyword (here and thereafter) to make it clear that it follows our "dimensions" ordering rather than the NumPy "shape" ordering

sunpy/map/header_helper.py
```diff
@@ -361,3 +361,52 @@ def _set_instrument_meta(meta_wcs, instrument, telescope, observatory, wavelengt
...
+    Longitude/Latitude.
+
+    Parameters
+    ----------
+    date :
+        Date for the output header.
+    observer_coordinate :
+        Observer coordinate for the output header.
+    shape_out : [int, int]
+        Output map shape, number of pixels in (latitude, longitude).
```
ayshih, line 381: This description is incorrect

sunpy/map/header_helper.py
```diff
@@ -361,3 +361,52 @@ def _set_instrument_meta(meta_wcs, instrument, telescope, observatory, wavelengt
...
+    )
+
+    if projection_code == "CAR":
+        scale = [360 / int(shape_out[0]), 180 / int(shape_out[1])] * u.deg / u.pix
+    elif projection_code == "CEA":
+        # Since, this map uses the cylindrical equal-area (CEA) projection,
+        # the spacing needs to be to 180/pi times the sin(latitude)
+        # spacing
+        # Reference: Section 5.5, Thompson 2006
+        scale = [360 / int(shape_out[0]), 180 / int(shape_out[1]) / (np.pi / 2)] * u.deg / u.pix
```
ayshih, line 408:
```suggestion
        scale = [360 / int(shape_out[0]), (180 / np.pi) * (2 / int(shape_out[1]))] * u.deg / u.pix
```
If you want to keep things a bit broken out for clarity, you ought to group the elements more logically.  The `180 / np.pi` is the conversion from radians to degrees, and the `2 / int(shape_out[1])` is the spacing of sin(latitude) bins.  Note how my suggested arrangement corresponds with the preceding comment.

ayshih (commented):
I worry about potential confusion that this helper accepts the dimensions/shape in (lon, lat), whereas `make_fitswcs_header` accepts it in (lat, lon) (when supplying a shape rather than an instantiated array).  We ought to discuss whether we are comfortable with this seeming inconsistency.

sunpy/map/header_helper.py
dstansby: Since I swapped the dimensions round to be consistent with the general header helper, I think this is fine now.

sunpy/map/header_helper.py
dstansby: Since I swapped the dimensions round to be consistent with the general header helper, I think this is fine now.

sunpy/map/header_helper.py
```diff
@@ -361,3 +362,54 @@ def _set_instrument_meta(meta_wcs, instrument, telescope, observatory, wavelengt
     'CD2_2':
     'Matrix element CDi_j describing the rotation required to align solar North with the top of the image.'
 }
+
+
+def carrington_header(date, observer_coordinate, shape, *, projection_code="CAR"):
```
Cadair, line 367: I would like to bikeshed this a little.

What would be the pros/cons of making this into a generic heliographic helper, where it can do stonyhurst as well?

It would make the API slightly more complex, i.e adding a new keyword argument for frame, but on the pro side it would allow people to reuse the pixel size calculation for stonyhurst frames as well?

sunpy/map/header_helper.py
```diff
@@ -361,3 +362,54 @@ def _set_instrument_meta(meta_wcs, instrument, telescope, observatory, wavelengt
...
+    elif projection_code == "CEA":
+        # Using the cylindrical equal-area (CEA) projection,
+        # scale needs to be to 180/pi times the sin(latitude) spacing
+        # See Section 5.5, Thompson 2006
+        scale = [
+            360 / int(shape[1]),
+            (180 / np.pi) / (int(shape[0]) / 2)
+        ] * u.deg / u.pix
+
+    # Header helper expects shape to be in [y, x] order, but scale in [x, y]...
```
Cadair, line 413: This is now out of date?
```suggestion
```

sunpy/map/header_helper.py
dstansby: What's your proposed API, including a new function name? I'm not against making it more general in theory, but think there is value in having a simple and recognisable function name.

Perhaps this is where some sort of user testing/questioning would be valuable?

sunpy/map/header_helper.py
dstansby: Good catch - will remove alonside any bikeshedding that gets done above 👍

sunpy/map/header_helper.py
wtbarnes, line 367: Could we change the name here to be consistent with `make_fitswcs_header`? 
```suggestion
def make_carrington_header(date, observer_coordinate, shape, *, projection_code="CAR"):
```

dstansby: Good shout, done

Cadair: I was thinking something like:
```python
def make_heliographic_header(date, observer_coordinate, shape, *, frame='heliographic_carrington', projection_code="CAR")
```
where frame is then passed to SkyCoord to make the reference coordinate?

dstansby: Hmm, I would say that "Carrington" is a more recognisable name than "Heliographic" to a general solar physicist, so I'm pro keeping it as Carrington. If there's demand for a Stonyhurst version then we could add `make_stonyhurst_header` and de-duplicate the interal logic instead of unifying the API?

dstansby: Decided to go with `heliographic` in the community meeting today

docs/guide/data_types/maps.rst
```diff
@@ -492,38 +492,40 @@ See this :ref:`sphx_glr_generated_gallery_map_map_from_numpy_array.py` for a bri
 
 The keys required for the header information follow the `FITS standard <https://fits.gsfc.nasa.gov/fits_dictionary.html>`__.
 **sunpy** provides a Map header helper function to assist in creating a header that contains the correct meta information.
-This includes a `~sunpy.map.meta_keywords` function that will return a `dict` the meta keywords used when creating a Map.
+This includes a `~sunpy.map.header_helper.meta_keywords` function that will return a `dict` the meta keywords used when creating a Map.
```
wtbarnes, line 495:
```suggestion
This includes a `~sunpy.map.header_helper.meta_keywords` function that will return a `dict` of the meta keywords used when creating a Map.
```

docs/guide/data_types/maps.rst
```diff
@@ -492,38 +492,40 @@ See this :ref:`sphx_glr_generated_gallery_map_map_from_numpy_array.py` for a bri
...
 If neither of these are given their values default to the center of the data array and 1 arcsec, respectively.
 
 Here's an example of creating a header from some generic data and an `astropy.coordinates.SkyCoord`:
 
 .. code-block:: python
 
     >>> import numpy as np
+    >>> from astropy.coordinates import SkyCoord
     >>> import astropy.units as u
+    >>>
```
wtbarnes, line 522:
```suggestion
```

docs/guide/data_types/maps.rst
```diff
@@ -596,14 +598,14 @@ Here's another example of passing ``reference_pixel`` and ``scale`` to the funct
     rsun_obs: 965.3829548285768
 
 As we can see, a list of WCS and observer meta information is contained within the generated headers, however we may want to include other meta information including the observatory name, the wavelength and waveunit of the observation.
-Any of the keywords listed in ``header_helper.meta_keywords`` can be passed to the `~sunpy.map.make_fitswcs_header` and will then populate the returned MetaDict header.
-Furthermore, the following observation keywords can be passed to the `~sunpy.map.make_fitswcs_header` function and will be translated to the FITS standard: ``observtory``, ``instrument``, ``telescope``, ``wavelength``, ``exposure``.
+Any of the keywords listed in ``header_helper.meta_keywords`` can be passed to the `~sunpy.map.header_helper.make_fitswcs_header` and will then populate the returned MetaDict header.
+Furthermore, the following observation keywords can be passed to the `~sunpy.map.header_helper.make_fitswcs_header` function and will be translated to the FITS standard: ``observtory``, ``instrument``, ``telescope``, ``wavelength``, ``exposure``.
```
wtbarnes, line 602:
```suggestion
Furthermore, the following observation keywords can be passed to the `~sunpy.map.header_helper.make_fitswcs_header` function: ``observatory``, ``instrument``, ``telescope``, ``wavelength``, ``exposure``.
These will then be translated to the corresponding FITS keyword according to the FITS standard.
```

wtbarnes (approved):

sunpy/map/header_helper.py
```diff
@@ -361,3 +362,70 @@ def _set_instrument_meta(meta_wcs, instrument, telescope, observatory, wavelengt
     'CD2_2':
     'Matrix element CDi_j describing the rotation required to align solar North with the top of the image.'
 }
+
+
+def make_heliographic_header(date, observer_coordinate, shape, frame, *, projection_code="CAR"):
```
Cadair, line 367: How would you feel about this? I think it makes calling this a lot clearer if the `frame=` is there.
```suggestion
def make_heliographic_header(date, observer_coordinate, shape, *, frame, projection_code="CAR"):
```

Cadair (commented):
I think this is good. I assume that the changes to use `map.header_helper` everywhere are just a case of documentation rather than API changes?

sunpy/map/header_helper.py
dstansby: 👍  done

dstansby:
> I assume that the changes to use map.header_helper everywhere are just a case of documentation rather than API changes?

Yes, I've created a new listing in the docs for `sunpy.map.header_helper` and changed examples to use that namespace, but the functions are still available in the `sunpy.map` namespace.

Cadair:
Looks good apart from the very real doctest fail: https://github.com/sunpy/sunpy/actions/runs/3318989623/jobs/5483584914#step:9:5186

Cadair (approved):
@ayshih I would like your final review on this please.

docs/whatsnew/4.1.rst
```diff
@@ -33,6 +33,11 @@ between the consecutive coordinates, in pixel space, and then indexing the data
 array of the map at those points.
 See :ref:`sphx_glr_generated_gallery_units_and_coordinates_map_slit_extraction.py` for an example.
 
+Easier reprojecting to Carrington and Stonyhurst coordinates
+============================================================
+``sunpy.map.header_helper`` has a new `~sunpy.map.header_helper.make_heliographic_header` function that simplifies creating map headers that span the whole solar surface in Carrington or Stonyhurst coordinates.
+When used in combination with `sunpy.map.GenericMap.reproject_to` this greatly simplifies reprojecting helioprojective maps into heliographic full-Sun maps.
+
```
ayshih, line 40: Minigallery!

sunpy/map/header_helper.py
```diff
@@ -375,3 +376,70 @@ def _set_instrument_meta(meta_wcs, instrument, telescope, observatory, wavelengt
...
+    --------
+    >>> from sunpy.map.header_helper import make_heliographic_header
+    >>> from sunpy.coordinates import get_earth
+    >>>
+    >>> date = '2020-01-01 12:00:00'
+    >>> observer = get_earth(date)
+    >>> header = make_heliographic_header(date, observer, [90, 180], frame='carrington')
+    >>> header
+    MetaDict([('wcsaxes', 2), ('crpix1', 90.5), ('crpix2', 45.5), ('cdelt1', 2.0), ('cdelt2', 2.0), ('cunit1', 'deg'), ('cunit2', 'deg'), ('ctype1', 'CRLN-CAR'), ('ctype2', 'CRLT-CAR'), ('crval1', 0.0), ('crval2', 0.0), ('lonpole', 0.0), ('latpole', 90.0), ('mjdref', 0.0), ('date-obs', '2020-01-01T12:00:00.000'), ('rsun_ref', 695700000.0), ('dsun_obs', 147096975776.97), ('hgln_obs', 0.0), ('hglt_obs', -3.0011725838606), ('naxis', 2), ('naxis1', 180), ('naxis2', 90), ('pc1_1', 1.0), ('pc1_2', -0.0), ('pc2_1', 0.0), ('pc2_2', 1.0), ('rsun_obs', 975.5398432033492)])
+    """
```
ayshih, line 416: Minigallery!

sunpy/map/header_helper.py
```diff
@@ -375,3 +376,70 @@ def _set_instrument_meta(meta_wcs, instrument, telescope, observatory, wavelengt
     'CD2_2':
     'Matrix element CDi_j describing the rotation required to align solar North with the top of the image.'
 }
+
+
+def make_heliographic_header(date, observer_coordinate, shape, *, frame, projection_code="CAR"):
+    """
+    Construct a FITS-WCS header for a heliographic (Carrington or Stonyhurst) coordinate frame.
```
ayshih, line 383:
```suggestion
    Construct a FITS-WCS header for a full-Sun heliographic (Carrington or Stonyhurst) coordinate frame.
```

sunpy/map/header_helper.py
```diff
@@ -375,3 +376,70 @@ def _set_instrument_meta(meta_wcs, instrument, telescope, observatory, wavelengt
...
+
+    The date-time and observer coordinate of the new coordinate frame
+    are taken from the input map. The resulting WCS covers the full surface
+    of the Sun, and has a reference coordinate at (0, 0) degrees Longitude/Latitude.
+
```
ayshih, line 388: Add a note saying that if more customization of the WCS is desired, use `make_fitswcs_header()` instead.

ayshih (approved):
The code looks fine to me; I just have a few documentation suggestions.  Also, you may as well update the "Creating a Full Sun Map with AIA and EUVI" to use this new function.

ayshih (changes requested):
Hold!  There's a problem...

sunpy/map/header_helper.py
```diff
@@ -375,3 +376,70 @@ def _set_instrument_meta(meta_wcs, instrument, telescope, observatory, wavelengt
...
+    if frame not in valid_frames:
+        raise ValueError(f"frame must be one of {valid_frames}")
+
+    frame_out = SkyCoord(
+        0 * u.deg,
+        0 * u.deg,
+        frame=f"heliographic_{frame}",
+        obstime=date,
+        observer=observer_coordinate,
+    )
```
ayshih, line 431:
```suggestion
        rsun=getattr(observer_coordinate, "rsun", None),
    )
```
`rsun` needs to be carried through.  Currently your "Creating Carrington maps" example is discarding data near the limb.

The fact that (a non-default) `rsun` is specified via a frame attribute on the `observer_coordinate` input should probably be a note in the docstring (of both header helpers).

ayshih (dismissed):
Important bug fix

sunpy/map/header_helper.py
Cadair: Yes!

sunpy/map/header_helper.py
dstansby: I've added a See Also section for this.

sunpy/map/header_helper.py
dstansby: Done

dstansby:
Hmm, I've tried adding a minigallery, but `sphinx-gallery` doesn't seem to be picking up the example modified to use `make_heliographic_header` ([here's a link to the example in the latest RTD build of this PR](https://sunpy--6415.org.readthedocs.build/en/6415/generated/gallery/map_transformations/reprojection_carrington.html#sphx-glr-generated-gallery-map-transformations-reprojection-carrington-py)). The issue is not with linking the minigallery, as a minigallery page is generated in `_build/generated/modules`, but the page is empty for `make_heliographic_header`. Does anyone have any ideas why this could be?

Cadair:
So I added a commit which makes it work, but it uses the `map.` import not the header_helper one.

ayshih:
That's just the way it is.  Depending on how it is imported, the backreference can be constructed differently.  So you either specify every possibility in the `minigallery` call, or you look at the generated files to see if only one possibility happens to cover what you want.

dstansby:
Thanks for the investigation/solution! Can we merge this now?

Cadair:
The major drawback in my solution is the fact that we have to import it from the `sunpy.map` namespace in the example not `sunpy.map.header_helper`

ayshih:
Huh, what?  Okay, I need to review =)

examples/map_transformations/reprojection_carrington.py
```diff
@@ -24,32 +22,23 @@
...
-                                       frame_out,
-                                       scale=(360 / shape_out[1],
-                                              180 / shape_out[0]) * u.deg / u.pix,
-                                       projection_code="CAR")
+# System (WCS) projection. Here we use :func:`sunpy.map.header_helper.make_heliographic_header`
+# to create a FITS WCS header based on a heliographic Carrington reference
+# coordinate.
+
+shape = (720, 1440)
+carr_header = make_heliographic_header(aia_map.date, aia_map.observer_coordinate, shape, frame='carrington')
```
hayesla, line 33: this is very nice

examples/map_transformations/reprojection_carrington.py
```diff
@@ -24,32 +22,23 @@
 
 plt.figure()
 aia_map.plot()
-plt.show()
```
nabobalis, line 27: Do we not need a show at each point?

ayshih:
> So you either specify every possibility in the `minigallery` call, or you look at the generated files to see if only one possibility happens to cover what you want.

So, it's essentially what I said earlier.  When you import `sunpy.map.header_helper.make_heliographic_header` in an example, the backreference that gets built is `sunpy.map.make_heliographic_header`, so that's what you have to specify the minigallery for.  (Maybe I'll see if that can be fixable in `sphinx-gallery`...)  There's no need to also modify the import.

examples/map_transformations/reprojection_carrington.py
nabobalis: I don't want to block this PR on that, so we can ignore my comment and punt it to a weekly meeting or the doc sprint.

ayshih: Created #6561 for continued discussion

ayshih (approved):
It was a surprisingly torturous road for this PR, but at last it's good to go!

Issue #6244, opened 2022-06-10
Use `pooch` to manage sample data?


dstansby:
### Describe the feature

[pooch](https://www.fatiando.org/pooch/latest/sample-data.html#intermediate) is a not-very-new but newer-than-sunpy package that was built for managing package sample data. It includes a lot of the functionality that we currently have custom implementations of for registring and downloading the data. I would like to propose we use `pooch` to manage our sample data in `4.1` because:

- It removes the need for us to maintain the sample data download code
- It adds the possiblity of users downloading individual files on demand, instead of downloading all the sample data when they only want one file
- This in turn means we don't need to worry as much about adding new files to the sample data as/when we need them

See https://github.com/sunpy/sunpy/pull/6242 for how this works in practice with one new file. In terms of implementation I would propose to;
- Keep `sunpy.data.sample` for backwards compatability, but with a deprecation notice in `4.1`, and remove in `5.1`
- Add the new functionality in a `sunpy.sample_data` module in `4.1`
- Switch all the examples to using the new functionality in `4.1`

Thoughts/opinions/reasons this is a bad idea? Maybe one day I'll make a call so this can be discussed, or we could discuss (and maybe implement?) at the in-person dev meeting.

Cadair:
It's *another* dependency, but pooch itself is pretty light on the deps and it would mean we could move parfive into the `net` extra.

My main question is why make a `sunpy.sample_data` can't we inplace swap this out? As long as the constant names keep working then the rest of the API is pretty "hidden".

dstansby:
I couldn't think of a way to keep the same API. With pooch the workflow is to define a function that gets the data and returns whatever object you want, and then call that function when you want the data. Currently the workflow is to get a string, and assume that points to an existing file. I'm not sure how you would trigger code to download a file when that string is accessed.

dstansby:
I had a bit of a search around today, and it looks like we might be able to use some Python magic to achieve the same API: https://mail.python.org/pipermail/python-ideas/2012-May/014969.html

> There is actually a hack that is occasionally used and recommended: a module can define a class with the desired functionality, and then at the end, replace itself in sys.modules with an instance of that class (or with the class, if you insist, but that's generally less useful). E.g.:

If we replace `sunpy.data.sample` with a class, we can impelment a custom getattr on the class to have the data download whenever the sample data file is accessed as a string.

ayshih:
[...]

PR #6426, opened 2022-09-27
Download sample data files on demand

ayshih:
This PR changes the way that sample data is downloaded.  Rather than downloading all the files when `sunpy.data.sample` is imported, the download is deferred until a sample file is actually requested, and then only that file is downloaded:
```python
>>> import sunpy.data.sample
>>> import sunpy.map
>>> aiamap = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
Files Downloaded: 100%|████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00,  1.61file/s]
```
It is still possible to trigger the download of the entire set – e.g., to take advantage of when internet access is available by calling `sunpy.data.sample.download_all()`.

See discussion in #6244

To consider:
~- Merge `_sample.py` into `sample.py`?~ Deferred to future cleanup (e.g., when `download_sample_data()` is removed)

### DIFF
--- a/sunpy/data/_sample.py
+++ b/sunpy/data/_sample.py
@@ -4,6 +4,7 @@ from urllib.parse import urljoin
 
 from sunpy import log
 from sunpy.util.config import _is_writable_dir, get_and_create_sample_dir
+from sunpy.util.decorators import deprecated
 from sunpy.util.parfive_helpers import Downloader
 
 _BASE_URLS = (
@@ -122,6 +123,19 @@ def _handle_final_errors(results):
         )
 
 
+def _get_sampledata_dir():
+    # Workaround for tox only. This is not supported as a user option
+    sampledata_dir = os.environ.get("SUNPY_SAMPLEDIR", False)
+    if sampledata_dir:
+        sampledata_dir = Path(sampledata_dir).expanduser().resolve()
+        _is_writable_dir(sampledata_dir)
+    else:
+        # Creating the directory for sample files to be downloaded
+        sampledata_dir = Path(get_and_create_sample_dir())
+    return sampledata_dir
+
+
+@deprecated('4.1', alternative='`sunpy.data.sample.download_all`')
 def download_sample_data(overwrite=False):
     """
     Download all sample data at once. This will overwrite any existing files.
@@ -131,14 +145,7 @@ def download_sample_data(overwrite=False):
     overwrite : `bool`
         Overwrite existing sample data.
     """
-    # Workaround for tox only. This is not supported as a user option
-    sampledata_dir = os.environ.get("SUNPY_SAMPLEDIR", False)
-    if sampledata_dir:
-        sampledata_dir = Path(sampledata_dir).expanduser().resolve()
-        _is_writable_dir(sampledata_dir)
-    else:
-        # Creating the directory for sample files to be downloaded
-        sampledata_dir = Path(get_and_create_sample_dir())
+    sampledata_dir = _get_sampledata_dir()
 
     already_downloaded = []
     to_download = []
@@ -166,3 +173,31 @@ def download_sample_data(overwrite=False):
             _handle_final_errors(results)
 
     return results + already_downloaded
+
+
+def _get_sample_files(filename_list, no_download=False, force_download=False):
+    sampledata_dir = _get_sampledata_dir()
+
+    fullpath = [sampledata_dir/fn for fn in filename_list]
+
+    if no_download:
+        fullpath = [fp if fp.exists() else None for fp in fullpath]
+    else:
+        to_download = zip(filename_list, fullpath)
+        if not force_download:
+            to_download = [(fn, fp) for fn, fp in to_download if not fp.exists()]
+
+        if to_download:
+            results = _download_sample_data(_BASE_URLS[0], to_download, overwrite=force_download)
+
+            # Something went wrong
+            if results.errors:
+                for next_url in _BASE_URLS[1:]:
+                    results = _retry_sample_data(results, next_url)
+                    if not results.errors:
+                        break
+                else:
+                    _handle_final_errors(results)
+                    raise RuntimeError
+
+    return fullpath
--- a/sunpy/data/_sample.py
+++ b/sunpy/data/_sample.py
@@ -176,21 +176,47 @@ def download_sample_data(overwrite=False):
 
 
 def _get_sample_files(filename_list, no_download=False, force_download=False):
+    """
+    Returns a list of disk locations corresponding to a list of filenames for
+    sample data, downloading the sample data files as necessary.
+
+    Parameters
+    ----------
+    filename_list : `list` of `str`
+        List of filenames for sample data
+    no_download : `bool`
+        If ``True``, do not download any files, even if they are not present.
+        Default is ``False``.
+    force_download : `bool`
+        If ``True``, download all files, and overwrite any existing ones.
+        Default is ``False``.
+
+    Returns
+    -------
+    `list` of `pathlib.Path`
+        List of disk locations corresponding to the list of filenames.  An entry
+        will be ``None`` if ``no_download == True`` and the file is not present.
+
+    Raises
+    ------
+    RuntimeError
+        Raised if any of the files cannot be downloaded from any of the mirrors.
+    """
     sampledata_dir = _get_sampledata_dir()
 
-    fullpath = [sampledata_dir/fn for fn in filename_list]
+    fullpaths = [sampledata_dir/fn for fn in filename_list]
 
     if no_download:
-        fullpath = [fp if fp.exists() else None for fp in fullpath]
+        fullpaths = [fp if fp.exists() else None for fp in fullpaths]
     else:
-        to_download = zip(filename_list, fullpath)
+        to_download = zip(filename_list, fullpaths)
         if not force_download:
             to_download = [(fn, fp) for fn, fp in to_download if not fp.exists()]
 
         if to_download:
             results = _download_sample_data(_BASE_URLS[0], to_download, overwrite=force_download)
 
-            # Something went wrong
+            # Try the other mirrors for any download errors
             if results.errors:
                 for next_url in _BASE_URLS[1:]:
                     results = _retry_sample_data(results, next_url)
@@ -200,4 +226,4 @@ def _get_sample_files(filename_list, no_download=False, force_download=False):
                     _handle_final_errors(results)
                     raise RuntimeError
 
-    return fullpath
+    return fullpaths
--- a/sunpy/data/sample.py
+++ b/sunpy/data/sample.py
@@ -9,8 +9,8 @@ Summary variables
    :widths: auto
 
    * - ``file_dict``
-     - Dictionary of all sample shortnames and, if downloaded, file locations
-       on disk
+     - Dictionary of all sample shortnames and, if downloaded, corresponding
+       file locations on disk (otherwise, ``None``)
    * - ``file_list``
      - List of disk locations for sample data files that have been downloaded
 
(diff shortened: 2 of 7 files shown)


nabobalis:
> * [ ]  Deprecate/remove `download_sample_data()`?

I can't recall if that function was Public, is so, we should deprecate it. 
It might still be nice to be able to download all of before a CI run.

> * [ ]  Merge `_sample.py` into `sample.py`?

Why not.

Cadair:
>  Deprecate/remove download_sample_data()?

Given this:

> by accessing either file_dict or file_list under sunpy.data.sample.

I would rather keep the API as a nice way for people to download the whole set?

dstansby:
> I would rather keep the API as a nice way for people to download the whole set?

I'm also 👍 to keeping `download_sample_data()`.

ayshih:
Okay, I'm on board with retaining a function to download everything, but I think I'll reimplement it and give it a new name (and deprecate the old function, even though no one was actually calling it directly) because I think it's too silly to tell people to call `sunpy.data.sample.download_sample_data()`.

dstansby:
Would importing it into the `sunpy` namespace work instead: `sunpy.download_sample_data()`?

sunpy/data/_sample.py
```diff
@@ -122,6 +123,19 @@ def _handle_final_errors(results):
         )
 
 
+def _get_sampledata_dir():
```
nabobalis, line 126: This came up in several matrix discussions.

I need to check if we still need this anymore or if dropping this and making tox work properly with the config is better.

sunpy/data/_sample.py
```diff
@@ -166,3 +173,31 @@ def download_sample_data(overwrite=False):
             _handle_final_errors(results)
 
     return results + already_downloaded
+
+
+def _get_sample_files(filename_list, no_download=False, force_download=False):
+    sampledata_dir = _get_sampledata_dir()
+
+    fullpath = [sampledata_dir/fn for fn in filename_list]
```
dstansby, line 181:
```suggestion
    fullpaths = [sampledata_dir/fn for fn in filename_list]
```
Since it's multiple paths?

sunpy/data/_sample.py
```diff
@@ -166,3 +173,31 @@ def download_sample_data(overwrite=False):
...
+
+    if no_download:
+        fullpath = [fp if fp.exists() else None for fp in fullpath]
```
dstansby, line 184: It seems a bit un-intuitive that this returns a list that can include `None` entries - can you include a quick description of this behaviour in a short docstring to avoid future developer confusion?

sunpy/data/_sample.py
```diff
@@ -166,3 +173,31 @@ def download_sample_data(overwrite=False):
...
+    else:
+        to_download = zip(filename_list, fullpath)
+        if not force_download:
+            to_download = [(fn, fp) for fn, fp in to_download if not fp.exists()]
+
+        if to_download:
```
dstansby, line 190:
```suggestion
        if len(to_download):
```
To make it clear that `to_download` is an iterable and not a `bool`.

ayshih: I understand your perspective, but I'll point out that the PEP 8 recommendation – which I don't always agree with – is to not have the additional `len()` when checking for empty sequences.

sunpy/data/_sample.py
wtbarnes, line 184:
```suggestion
        fullpaths = [fp if fp.exists() else None for fp in fullpath]
```
To be consistent with above

sunpy/data/_sample.py
```diff
@@ -166,3 +173,31 @@ def download_sample_data(overwrite=False):
...
+            if results.errors:
+                for next_url in _BASE_URLS[1:]:
+                    results = _retry_sample_data(results, next_url)
+                    if not results.errors:
+                        break
+                else:
+                    _handle_final_errors(results)
+                    raise RuntimeError
+
+    return fullpath
```
wtbarnes, line 203:
```suggestion
    return fullpaths
```

sunpy/data/_sample.py
```diff
@@ -166,3 +173,31 @@ def download_sample_data(overwrite=False):
...
+        fullpath = [fp if fp.exists() else None for fp in fullpath]
+    else:
+        to_download = zip(filename_list, fullpath)
+        if not force_download:
+            to_download = [(fn, fp) for fn, fp in to_download if not fp.exists()]
+
+        if to_download:
+            results = _download_sample_data(_BASE_URLS[0], to_download, overwrite=force_download)
+
+            # Something went wrong
```
wtbarnes, line 193: Could you make this comment a bit more descriptive, e.g. say at a high level what happens when something goes wrong

sunpy/data/sample.py
```diff
@@ -1,35 +1,62 @@
...
-    _key = _SAMPLE_FILES.get(name, None)
-    if _key:
-        setattr(sys.modules[__name__], _key, str(f))
-        file_dict.update({_key: f})
 
-# Sort the entries in the dictionary
-file_dict = dict(sorted(file_dict.items()))
+def __getattr__(name):
+    if name in _SAMPLE_DATA:
+        return _get_sample_files([_SAMPLE_DATA[name]])[0]
```
wtbarnes, line 42: Can you add a comment saying why the `[0]` is needed?

sunpy/data/sample.py
```diff
@@ -1,35 +1,62 @@
...
+    __doc__ += f'   * - ``{_keyname}``\n     - {_filename}\n'
+
+
+__all__ = list(sorted(_SAMPLE_DATA.keys())) + ['download_all', 'file_dict', 'file_list']
 
-from ._sample import _SAMPLE_FILES, download_sample_data
 
-files = download_sample_data()
+def __dir__():
+    return __all__
```
wtbarnes, line 37: Consider adding a comment that links to [PEP 562](https://peps.python.org/pep-0562/), the proposal for allowing module-level `__dict__` and `__getattr__`. This is esoteric enough that a clarifying comment and helpful link would be beneficial to future readers.

sunpy/data/sample.py
```diff
@@ -1,35 +1,62 @@
...
-for f in files:
-    name = Path(f).name
-    _key = _SAMPLE_FILES.get(name, None)
-    if _key:
-        setattr(sys.modules[__name__], _key, str(f))
-        file_dict.update({_key: f})
 
-# Sort the entries in the dictionary
-file_dict = dict(sorted(file_dict.items()))
+def __getattr__(name):
```
wtbarnes, line 40: Same as above. A link to PEP 562 would be nice.

wtbarnes (approved):
I just left a few comments asking for some clarifying comments, but otherwise this looks good to me. I think this is a really great idea so thanks for implementing it.

sunpy/data/_sample.py
nabobalis: This comment was a hold over from the old code, I think it should go.

sunpy/data/_sample.py
nabobalis: Is there a reason to check the length of it instead of just doing a plain if?

sunpy/data/sample.py
nabobalis: To escape the container?

sunpy/data/_sample.py
ayshih: Changed

sunpy/data/_sample.py
ayshih: Added in two places

sunpy/data/_sample.py
ayshih: I think @dstansby isn't particularly comfortable with the falsiness of empty sequences, which is a fair position

sunpy/data/_sample.py
ayshih: Yup, changed

sunpy/data/_sample.py
ayshih: Comment updated for clarity

sunpy/data/sample.py
ayshih: `_get_sample_files()` accepts a list and returns a list, and that's now made clear in a docstring.  That's why the single argument is turned into a one-element list, and the return is de-listed.

sunpy/data/sample.py
ayshih: Done

sunpy/data/sample.py
ayshih: Done

ayshih:
All comments have been addressed.  There is now a rather detailed docstring for a private function. =)  Someone can tell me why `autoflake` is complaining because it doesn't complain locally on my machine.

The only "negative" that I have encountered is that tab completion within `sunpy.data.sample` (e.g., in IPython) will trigger downloads.  That is, if you type `sunpy.data.sample.AIA_1` and then hit \<Tab\>, the first matching shortname (i.e., `AIA_131_IMAGE`) will immediately be downloaded.  (I wonder if this is unintended behavior in IPython, because I don't know why it's accessing any matching entry at all, and if it has a good reason to do so, why it is not accessing all of them.)  Interestingly, although the progress bar does get displayed in IPython, *nothing* gets displayed in Jupyter Notebook.  I don't think this curious behavior should hold back this PR.

sunpy/data/_sample.py
nabobalis: Is it bad practice? I use it all the time in my code.

nabobalis:
> Interestingly, although the progress bar does get displayed in IPython, _nothing_ gets displayed in Jupyter Notebook. I don't think this curious behavior should hold back this PR.

I assume this is a parfive issue?

ayshih:
Sorry, to be clear, by "curious behavior" I meant the triggering of downloading by tab completion, not that the parfive output can go into a void when using Jupyter Notebook.

sunpy/data/_sample.py
ayshih: As I said, PEP 8 recommends not using `len()`, so it's certainly not "bad practice"

dstansby: 👍 happy to keep this and follow PEP8

dstansby (approved):
Looks great now 👍 Might want to open separate issues about the autocomplete/notebook stuff to keep track?

I fixed pre-commit - see the code comment in the latest commit I pushed. Someone else should review that commit, but anyone (including Albert) can merge if they're happy with my commit.

Cadair:
I would like to review this.

ayshih:
> I fixed pre-commit - see the code comment in the latest commit I pushed.

Thanks!  I understand now how I was running `autoflake` incorrectly.

> I would like to review this.

Sure, review away!

I'll point out with amusement that once @Cadair reviews this PR, it will have been (self-)reviewed by five of our most active developers.  I'll cheekily wonder out loud how many bugs would still get through if all of our PRs received this much review attention. =P

sunpy/data/sample.py
```diff
@@ -1,35 +1,70 @@
...
 
-# Sort the entries in the dictionary
-file_dict = dict(sorted(file_dict.items()))
+# See PEP 562 (https://peps.python.org/pep-0562/) for module-level __getattr__()
+def __getattr__(name):
+    if name in _SAMPLE_DATA:
+        return _get_sample_files([_SAMPLE_DATA[name]])[0]
+    elif name == 'file_dict':
+        return {k: _get_sample_files([_SAMPLE_DATA[k]], no_download=True)[0]
+                for k in _SAMPLE_DATA.keys()}
```
Cadair, line 53:
```suggestion
        return {k: _get_sample_files([v], no_download=True)[0]
                for k,v in _SAMPLE_DATA.items()}
```

Cadair, line 53: If this is `no_download=True` why does it need to be inside `__getattr__`?

Cadair (approved):
This is certainly a new bit of Python magic for us. A couple of comments.

Cadair:
> > I fixed pre-commit - see the code comment in the latest commit I pushed.
> 
> Thanks! I understand now how I was running `autoflake` incorrectly.

Remember you can always get the pre-commit.ci bot to fix stuff for you if you want.


> I'll point out with amusement that once @Cadair reviews this PR, it will have been (self-)reviewed by five of our most active developers. I'll cheekily wonder out loud how many bugs would still get through if all of our PRs received this much review attention. =P

I mostly wanted to review it to internalise the new module level magic.

sunpy/data/sample.py
ayshih: Whoops, yeah, my additional contortion is no longer necessary after a recent edit

ayshih: The behavior of `no_download=True` is such that the dictionary entries will have file locations for those sample files that have been downloaded and have `None` otherwise.  The reason to have `file_dict` dynamically built in `__getattr__` is so that it acts like a module-level property that returns an up-to-date dictionary upon request.

In principle, `file_dict` could be a normal dictionary and `_get_sample_files()` could have an additional step to update an entry upon every successful download.  However, that's actually a bit of a pain to code, is even more of a pain when having to support the deprecated `download_sample_data()`, and can't stay updated when sample files are removed through other means or are downloaded in a parallel Python kernel.

ayshih:
> Remember you can always get the pre-commit.ci bot to fix stuff for you if you want.

Heh, I continue to resist our robot overlords... =P

sunpy/data/sample.py
ayshih: Actually, I've now decided that I want sorting back.  This is now changed to:
```python
        return dict(sorted(zip(_SAMPLE_DATA.keys(),
                               _get_sample_files(_SAMPLE_DATA.values(), no_download=True))))
```

Cadair: Ah ok, I missed that interaction. Makes sense. :+1:

PR #6433, opened 2022-09-29
Remove solar rotation code

dstansby:
Deprecated in 4.0, so time to remove for 4.1.


wtbarnes:
Should we start (if we don't have one already) a "What's New" section for things that have been removed in 4.1?

wtbarnes:
Also the doc build failures are real. I believe you'll have to change any mentions of `sunpy.physics.solar_rotation` to ``sunpy.physics.solar_rotation`` now that that subpackage does not exist.

nabobalis (approved):

PR #6437, opened 2022-09-29
Remove `shift` and `shifted_value`

wtbarnes:
Since we deprecated `GenericMap.shift` and `GenericMap.shifted_value` in 4.0, it's time to remove them in 4.1


dstansby (changes requested):
I think there's some tests that need removing here too

Issue #6395, opened 2022-09-02
Passing an empty UnifiedResponse object to Fido.fetch returns a directory from parfive


alasdairwilson:
### Describe the bug

If you conduct a query with JSOCClient which returns no results then you get back an empty UnifiedResponse object, unlike the behaviour of say VSOClient which will not return anything.

```Fido.fetch``` will happily take this results object and "download"  and the resulting ```parfive.Results``` object that is returned is not empty, instead it returns the base download directory (either set in sunpy or if using the 'path=' you get that directory)

### To Reproduce

For example, we can guarantee an empty result with a future date:

```
from sunpy.net import Fido, attrs as a, attr
query = a.Time('2022-12-12T00:00:00','2022-12-12T0:01:00') \
                 & a.jsoc.Series.hmi_b_720s \
                 & a.jsoc.Notify('solarmonitor.info@gmail.com')
res = Fido.search(query)
```

Gives an empty response:

```
<sunpy.net.fido_factory.UnifiedResponse object at 0x000001E5AD515610>
Results from 1 Provider:

0 Results from the JSOCClient:res = Fido.search(query)
Source: http://jsoc.stanford.edu

<No columns>
``` 
Which we can fetch with Fido. 
```               
res = Fido.search(query)
f = Fido.fetch(res)
```
Outputs:
```
<parfive.results.Results object at 0x000001E599777A90>
['C:\\Users\\Alasdair\\sunpy\\data']
```

Which will obviously break anything which the parfive result is passed to.

Perhaps a manifestation of two bugs? 

1. JSOCClient should not be returning empty responses? 
2. parfive should not be returning anything if it is passed an empty response.

### System Details

sunpy: 4.0.4
python: 3.9
astropy: 5.1
numpy: 1.23.2
packaging: 21.3
parfive: 2.0.1

### Installation method

pip

PR #6449, opened 2022-10-07
Fix download result for empty JSOC searches

dstansby:
See changelog entry and test for context.

Fixes https://github.com/sunpy/sunpy/issues/6395

### DIFF
--- a/sunpy/net/jsoc/tests/test_jsoc.py
+++ b/sunpy/net/jsoc/tests/test_jsoc.py
@@ -357,13 +357,17 @@ def test_check_request_keywords(client):
     assert len(responses) == 7
 
 
-@pytest.mark.remote_data
 def test_empty_response_fetch(client):
     # Check that results from downloading an empty response are empty
-    response = client.search(
-        a.Time('1990-12-12T00:00:00', '1990-12-12T0:01:00'),
-        a.jsoc.Series.hmi_b_720s
-    )
+    #
+    # Could do a search like below, but to save a JSOC query in the tests
+    # just test against an empty response.
+    #
+    # response = client.search(
+    #     a.Time('1990-12-12T00:00:00', '1990-12-12T0:01:00'),
+    #    a.jsoc.Series.hmi_b_720s
+    # )
+    response = JSOCResponse()
     assert len(response) == 0
     result = client.fetch(response)
     assert len(result) == 0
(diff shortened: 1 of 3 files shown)


sunpy/net/jsoc/tests/test_jsoc.py
```diff
@@ -355,3 +355,15 @@ def test_check_request_keywords(client):
         a.Time('2020/1/1T1:00:36', '2020/1/1T01:00:38'),
         a.jsoc.Series('aia.lev1_euv_12s'), a.jsoc.Keyword("QUALITY") < 2, a.jsoc.Keyword("EXPTIME") > 2)
     assert len(responses) == 7
+
+
+@pytest.mark.remote_data
+def test_empty_response_fetch(client):
+    # Check that results from downloading an empty response are empty
+    response = client.search(
+        a.Time('2022-12-12T00:00:00', '2022-12-12T0:01:00'),
```
dstansby, line 364:
```suggestion
        a.Time('1990-12-12T00:00:00', '1990-12-12T0:01:00'),
```

nabobalis (approved):

sunpy/net/jsoc/tests/test_jsoc.py
```diff
@@ -355,3 +355,15 @@ def test_check_request_keywords(client):
...
+
+
+@pytest.mark.remote_data
+def test_empty_response_fetch(client):
+    # Check that results from downloading an empty response are empty
+    response = client.search(
+        a.Time('1990-12-12T00:00:00', '1990-12-12T0:01:00'),
+        a.jsoc.Series.hmi_b_720s
+    )
+    assert len(response) == 0
```
Cadair, line 367: Can we not mock a response here to test this behaviour in an offline test?

Something like this should work:
```suggestion
    response = JSOCResponse()
```

dstansby: We could, but I think it's safer here to test the code in the way a user would interact with it.

Cadair: I don't think anything that actually involves interacting with JSOC servers is a good test :stuck_out_tongue: 

If we can avoid having to make a call to JSOC then the test is going to be a lot more reliable and less likely to be ignored as a "transient failure".

If you want to have both then ok, but I would rather have one which is offline as well.

dstansby: 👍  - I've done as suggested, but left the search in a comment so it's easy to enable an end-to-end test in the future (e.g. if the `vcrpy` stuff ever gets sorted).

Issue #6177, opened 2022-05-25
Issue with querying CDAWeb when searched over time when no data exists


hayesla:
### Describe the bug

When using the `cdaweb.Dataset` attribute, it errors when you search over a time period for which no data exists between the start and end time. If you include a time period when some data exists it works, and only returns the data available (even if for the remaining time period is missing.

For example, with the Solar Orbiter MAG observations, data is only available up til 28-Feb-2022, and if you search for times after this (e.g. between 1-Mar-2022 and 2-Mar-2022) you get an error rather than it just returning no results. However if you search between 28-Feb-2022 and 2-Mar-2022 it will return one file and not error.

### To Reproduce

```python
from sunpy.net import Fido, attrs as a 
result = Fido.search(a.Time('2022/03/01', '2022/03/02'), a.cdaweb.Dataset('SOLO_L2_MAG-RTN-NORMAL-1-MINUTE'))
```

### What happened?

```python
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
/var/folders/wm/gysdzbp10qv1kcm4zxb5q95xkjzv5q/T/ipykernel_97579/3888184350.py in <module>
      1 from sunpy.net import Fido, attrs as a
----> 2 result = Fido.search(a.Time('2022/03/01', '2022/03/02'), a.cdaweb.Dataset('SOLO_L2_MAG-RTN-NORMAL-1-MINUTE'))

~/opt/miniconda3/lib/python3.9/site-packages/sunpy/net/fido_factory.py in search(self, *query)
    310         """
    311         query = attr.and_(*query)
--> 312         results = query_walker.create(query, self)
    313 
    314         # If we have searched the VSO but no results were returned, but another

~/opt/miniconda3/lib/python3.9/site-packages/sunpy/net/attr.py in create(self, *args, **kwargs)
    611         Call the create function(s) matching the arguments to this method.
    612         """
--> 613         return self.createmm(self, *args, **kwargs)
    614 
    615     def apply(self, *args, **kwargs):

~/opt/miniconda3/lib/python3.9/site-packages/sunpy/util/functools.py in wrapper(*args, **kwargs)
     16 
     17     def wrapper(*args, **kwargs):
---> 18         return dispatcher.dispatch(args[1].__class__)(*args, **kwargs)
     19 
     20     wrapper.dispatch = dispatcher.dispatch

[...]

PR #6450, opened 2022-10-07
Fix response for no CDAWEB results

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/6177


sunpy/net/cdaweb/cdaweb.py
```diff
@@ -94,17 +94,21 @@ def _do_search(self, query):
...
-        etimes = [f['EndTime'] for f in response['FileDescription']]
-        urls = [f['Name'] for f in response['FileDescription']]
-
-        return astropy.table.QTable(
-            {'Dataset': [query['dataset']] * len(stimes),
-             'Start time': Time.strptime(stimes, '%Y-%m-%dT%H:%M:%S.%fZ').iso,
-             'End time': Time.strptime(etimes, '%Y-%m-%dT%H:%M:%S.%fZ').iso,
-             'URL': urls})
+            # No results
+            return astropy.table.QTable(
```
Cadair, line 98: This shouldn't be returning a QTable at all it should be returning a QueryResponseTable (which is a subclass of QTable).

dstansby: This is an internal helper method, isn't it fine that this returns whatever as long as `search` returns the `QueryResponseTable` (which it does above)?

nabobalis (approved):

wtbarnes (approved):

dstansby:
Woops, I've fixed the changelog name. Accidentally closed and re-opened this PR, will merge when CI passes....

sunpy/net/cdaweb/cdaweb.py
nabobalis: I think if we are worried about this, we can follow up in a new PR but it works and I don't see the issue with it being a Qtable.

PR #6451, opened 2022-10-09
Fix for GOES 13,14,15 files to correctly parse the satellite number

### LINKED ISSUE #6417: GOES XRS observatory attribute not parsing correctly
### Describe the bug

another GOES/XRS issue ...

For the GOES 13,14, 15, from what I can tell the meta information within the netcdf file does not include the GOES Satellite Number but rather gives 'GOES 1-15 X-ray Irradiance Sensor (XRS)'. The way we are parsing out the `observatory` attribute is just to read this as GOES-15 but this is the same for GOES 13 and 14. 

Hence if you read in the GOES 13 or 14 data the `observatory` attribute will give you `GOES-15`.
This is an issue for the temperature/emission measure calculation (that will very soon be PR'd to sunkit-instruments) needs to know the satellite to use the correct response. 

(not an issue for GOES 16 and 17).

### To Reproduce

```python
>>> goes_15 = ts.TimeSeries("sci_gxrs-l2-irrad_g15_d20190301_v0-0-0.nc")
>>> goes_14 = ts.TimeSeries("sci_gxrs-l2-irrad_g14_d20190301_v0-0-0.nc")


>>> goes_15.observatory
'GOES-15'
>>> goes_14.observatory 
'GOES-15'
```

I think the only way to go around this is to take the satellite number from the file name itself (for GOES 13, 14, 15).

hayesla:
## PR Description

This fixes #6417 

For the netcdf GOES-XRS files, there is no information in the metadata as to what GOES satellite number the data is from. This needs to be now parsed from the filename. For the GOES 16, 17 the filename is actually within the metadata which is used to parse out the satellite number. Thats what is now done here, upon parsing the netcdf file, the filename is added to the `id` attribute within the metadata. 

For example this now works:

```python
>>> res = Fido.search(a.Time("2020-01-01", "2020-01-01"), a.Instrument("XRS"))
>>> ff = Fido.fetch(res)
>>> goes_ts = ts.TimeSeries(ff)
>>> for g in goes_ts:
            print(g.observatory)
GOES-15
GOES-16
GOES-17
GOES-14
```


nabobalis:
Do we not check for this in our source tests for GOES?

hayesla:
so yes there is a test - the issue was that for GOES 13, 14 is said it was GOES-15 and the[ test uses the GOES-15 data](https://github.com/sunpy/sunpy/blob/main/sunpy/timeseries/sources/tests/test_goes.py#L48).
there's also a test for GOES-17. So we could add one, but would need to add another test file

hayesla:
actually this also needed to be fixed for the 1minute average data, which I've now also added. I'll get on making a test for this

sunpy/timeseries/sources/goes.py
```diff
@@ -228,6 +229,8 @@ def _parse_netcdf(filepath):
         """
         with h5netcdf.File(filepath, mode="r", **XRSTimeSeries._netcdf_read_kw) as h5nc:
             header = MetaDict(OrderedDict(h5nc.attrs))
+            if len(header["id"].strip()) == 0:
+                header.update({"id": Path(filepath).name})
```
dstansby, line 233: In `map` we take the approach of not modifying any information in the header, but instead overriding properties. I think it would be good to follow that pattern in `timeseries`, if that's possible in this case (ie. with a custom `.observatory` property)?

hayesla: yeah thats a good point. I was only modifying it here if it was empty as the filename is in the 'id' meta for the GOES-16 and 17 satellites, but not for 13, 14, 15 for some reason. 

I'll change this to another name

nabobalis:
I think this will need rebasing due to the new GOES test files?

hayesla:
I've also renamed the goes test files in this PR to preserve the filename structure.

hayesla:
will that need a separate changelog or can I just include it in the current one?

dstansby:
🤷 I wouldn't bother at all to be honest, changing test filenames is only an internal change.

changelog/6451.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix a bug to parse the GOES `observatory` number in XRSTimeSeries for GOES 13, 14, 15 and for the 1minute GOES-R data.
```
nabobalis, line 1:
```suggestion
Fix a bug to parse the GOES ``observatory`` number in `~.XRSTimeSeries `for GOES 13, 14, 15 and for the 1 minute GOES-R data.
```

changelog/6451.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix a bug to parse the GOES ``observatory`` number in `~.XRSTimeSeries `for GOES 13, 14, 15 and for the 1 minute GOES-R data.
```
nabobalis, line 1:
```suggestion
Fix a bug to parse the GOES "observatory" number in `~.XRSTimeSeries` for GOES 13, 14, 15 and for the 1 minute GOES-R data.
```

nabobalis (approved):

PR #6460, opened 2022-10-10
Tidy GOES tests and adding new XRS files for testing

hayesla:
## PR Description

This PR tidies up some of the GOES-XRS timeseries tests. It adds two new files to cover all the files supported by XRSTimeSeries. I've also removed some redundant tests that were duplicated.

This will fail some some that will be fixed by #6451 (not too sure what to do in this regard).

Also I used `pytest.mark.parameterize` to test all files - I'm not too sure if this is the best way but only way I could think of.


dstansby:
> This will fail some some that will be fixed by https://github.com/sunpy/sunpy/pull/6451 (not too sure what to do in this regard).

The tests could be marked with [xfail](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail), and then the marker can be removed in the PR that fixes the tests?

> Also I used pytest.mark.parameterize to test all files - I'm not too sure if this is the best way but only way I could think of.

Seems like the best way to me 👍

dstansby (approved):

nabobalis (approved):

nabobalis:
@MeeseeksDev backport 4.0

Issue #5217, opened 2021-04-16
Scraper acting unexpectedly for different timeranges passed


hayesla:
### Description
I came across something very off with scraper when using it to scrape files that only have a `%Y` in the pattern of interest.

For example, I'm currently looking at scraping files from this pattern, and it works fine for this timerange.

```python

>>> file_pattern_ngdc = ("https://www.ngdc.noaa.gov/stp/space-weather/solar-data/solar-features/solar-flares/x-rays/goes/xrs/goes-xrs-report_%Y.txt")
>>> file_scraper_ngdc = scraper.Scraper(file_pattern_ngdc)


>>> trange = TimeRange("2013-01-01", "2013-12-31")
>>> file_scraper_ngdc.filelist(trange)
['https://www.ngdc.noaa.gov/stp/space-weather/solar-data/solar-features/solar-flares/x-rays/goes/xrs/goes-xrs-report_2013.txt']
```
But like if I slightly change the date (it shouldn't matter right - its should only look at 2013?) this returns nothing, whereas it should return the same file.
```python
>>> trange = TimeRange("2013-05-01", "2013-12-31")
>>> file_scraper_ngdc.filelist(trange)
[]

```
am I going mad? I think this is a bug withing scraper

again more of a need for #4888

hayesla:
so this issue lies in Scraper.[_check_timerange](https://github.com/sunpy/sunpy/blob/master/sunpy/util/scraper.py#L375).

if the url pattern only depends %Y, it extracts the datetime as the first day of the month in [_extractDateURL](https://github.com/sunpy/sunpy/blob/master/sunpy/util/scraper.py#L183). Then in the `_check_timerange` function, it checks that this time lies between the start and end of time. However if it only depends on %Y it should really only check that.

samaloney:
🤦

dpshelio:
My reasoning probably was that the data you obtain should be within the time range you ask for. For example, if I ask for a dataset between 10:50 and 10:51, I should get the ones in that time range and no before... even if there's an image with an observation time from 10:49 to 10:51. And probably the same logic appears when asking for a year. However, I would say that returning anything that observed in the time range one asks for should be desirable. The problem is that most of the time by looking at the filename you don't know how long does the observation lasts, so the image from 10:49 could just be from 10:49:00 to 10:49:12 and therefore that shouldn't be returned.

PR #6472, opened 2022-10-14
Fix scraper file timerange

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/5217, by making sure that the code determining the timerange a given file represents extends across the smallest interval specified in the filename.


nabobalis:
Or is this a bugfix?

dstansby:
I'm not sure - since this behaviour wasn't clearly previously documented, I'd say we shouldn't backport this and brand it as a new feature?

nabobalis (approved):

samaloney (commented):
Nice, perhaps an offline test for `_check_timerange` function could be a useful addition?

samaloney (approved):

Cadair (approved):

dstansby:
I've check the test failures and they look unrelated, so merging

Issue #6445, opened 2022-10-05
Provide guidance on setting up a development environment with all dependencies installed through `conda`?


ayshih:
Currently, our developer's guide tells users to use `conda` to set up an environment and install `pip`, and then use `pip` rather than `conda` to install all of `sunpy`'s dependencies.  If a user subsequently uses `conda` to update any dependencies in that environment, that environment may become broken for further `conda` use.

We could consider providing guidance on how to best use a local repo for `sunpy` but still use  `conda` for all of the dependencies, e.g.:
```
conda install sunpy
conda remove --force sunpy
pip install --no-build-isolation --no-deps -e .
```

nabobalis:
The reason I didn't add that and said to use pip is that using `conda install sunpy`, you still miss out on some testing dependencies we do not add to the conda recipe. 

Tox being the main one, but if you wanted to use pytest directly, there are a set of other packages to install. 

We could add those instructions with a note telling people about that fact?

Cadair:
In my opinion it's cleaner to install everything into a conda env with pip *and then remember you did so*. I get that that last part can be tricky though.

The other reason I am not sure this is a good idea is what Nabil said about the dev and testing dependencies.

ayshih:
I'll note that there are packages that are significantly more straightforward to install via `conda` than other means.  As such, I myself do precisely what I've suggested in order to sustain the mix of `conda` and `pip`.  (Ideally, `conda develop` would be useful in this regard, but apparently that command is not recommended for use by `conda` folks.)  It just requires one to remember to `conda remove --force` `sunpy`/`astropy`/whatever whenever `conda` installs a release version to satisfy a dependency, which masks the editable install.

> The reason I didn't add that and said to use pip is that using `conda install sunpy`, you still miss out on some testing dependencies we do not add to the conda recipe.

It'd be straightforward to have a file in our repo (e.g., `conda_dev_dependencies.txt`) that a user could pass to `conda install --file` in order to install those additional dependencies.  In fact, we essentially already have such a list in our `tox.ini`, so we just have to move it out to a separate file (cleaning it up at the same time), and then have `tox.ini` make use of that file.

> In my opinion it's cleaner to install everything into a conda env with pip _and then remember you did so_. I get that that last part can be tricky though.

[...]

PR #6479, opened 2022-10-19
Switch tox-conda to use a conda environment file

ayshih:
This PR itself is (currently) only an internal change, but the longer-term objective is to provide an (advanced) installation guide for a developer that wants to have an editable install of `sunpy` (and other specific packages) within an environment where all of the other packages are managed by `conda` (see #6445).

### DIFF
--- a/sunpy-dev-env.yml
+++ b/sunpy-dev-env.yml
@@ -1,0 +1,67 @@
+channels:
+  - conda-forge
+
+dependencies:
+  # Required
+  - asdf
+  - asdf-astropy
+  - astropy
+  - beautifulsoup4
+  - cdflib
+  - drms
+  - glymur
+  - h5netcdf
+  - matplotlib-base
+  - mpl_animators
+  - numpy
+  - pandas
+  - parfive
+  - python
+  - python-dateutil
+  - python_abi
+  - reproject
+  - scikit-image
+  - scipy
+  - setuptools
+  - sqlalchemy
+  - tqdm
+  - zeep
+
+  # Optional
+  - astroquery
+  - jplephem
+  - opencv
+
+  # Testing
+  - hypothesis
+  - pytest
+  - pytest-astropy
+  - pytest-cov
+  - pytest-doctestplus
+  - pytest-mock
+  - pytest-mpl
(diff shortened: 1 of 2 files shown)


sunpy-dev-env.yml
```diff
@@ -0,0 +1,67 @@
...
+
+  # Development
+  - pre-commit
+ 
+  # Not installable through conda
+  - pip:
+    - sphinx-bootstrap-theme
+    - sphinx-changelog
+    - sphinxext-opengraph
+    - sunpy-sphinx-theme
```
nabobalis, line 67: We could package some of them up?

tox.ini
```diff
@@ -122,47 +122,9 @@ commands =
...
-    scipy
-    sphinx
-    sqlalchemy
-    towncrier
-    zeep
-conda_channels = conda-forge
-install_command = pip install --no-deps {opts} {packages}
+conda_env = sunpy-dev-env.yml
+install_command = pip install --no-deps --no-build-isolation {opts} {packages}
+allowlist_externals = conda
```
nabobalis, line 127:
```suggestion
whitelist_externals = conda
```
See if this works?

ayshih: Ah, I see, it should probably change the other way actually, because `whitelist_externals` is deprecated.  That is, both places should use `allowlist_externals`.

nabobalis: Oh I didn't realize it was deprecated. We should change that to be correct.

The CI passed tho!

tox.ini
```diff
@@ -122,47 +122,9 @@ commands =
...
-    scipy
-    sphinx
-    sqlalchemy
-    towncrier
-    zeep
-conda_channels = conda-forge
-install_command = pip install --no-deps {opts} {packages}
+conda_env = sunpy-dev-env.yml
+install_command = pip install --no-deps --no-build-isolation {opts} {packages}
+whitelist_externals = conda
```
nabobalis, line 127:
```suggestion
allowlist_externals = conda
```
This needs to be changed near the top of the file as well.

nabobalis (approved):

PR #6480, opened 2022-10-19
Hinode map source fixes

wtbarnes:
This PR makes two small changes:

- Fix the `detector` property of `SOTMap` to actually return "SOT". It previously was hardcoded to "XRT", presumably from when the map source was copied over
- Change the default normalization of `XRTMap` to have a `LogStretch` rather than no stretch at all

With the current normalization (no stretch),

image

With this PR (using `LogStretch`),

image


ayshih:
Any particular reason you chose `LogStretch` instead of `AsinhStretch`?  I think we predominantly – although not exclusively – use `AsinhStretch` for similar imagery.

wtbarnes:
> Any particular reason you chose `LogStretch` instead of `AsinhStretch`? I think we predominantly – although not exclusively – use `AsinhStretch` for similar imagery.

I experimented with both. This is `AsinhStretch` with the default "stretch" parameter (`a`, I don't know what it's actually called),

image

You can see larger range of intensities with a smaller value of a (e.g. 1e-3),

image

But this to me looks basically equivalent to `LogStretch`.

wafels:
Helioviewer JP2 files implement a log stretch for all filter wheel combinations. I'm pretty sure this was decided in conjunction with the XRT team.

nabobalis (approved):

ayshih (approved):

ayshih:
Er, aside from maybe the backport...

wtbarnes:
> Helioviewer JP2 files implement a log stretch for all filter wheel combinations. I'm pretty sure this was decided in conjunction with the XRT team.

I did a by-eye comparison with the equivalent HV images and they looked similar (with `LogStretch`) so that is good to know.

Issue #6163, opened 2022-05-18
Setting `ylim` on `TimeSeries.plot(...)` for GOES data does not adjust the flare classification axis


wtbarnes:
### Describe the bug

When plotting a GOES XRS timeseries, a label for the associated flare classification is automatically added to the right axis. When you adjust the limits on the intensity axis, the corresponding limits are not applied to the flare-classification axis such that the correspondence between the two is not correct.

### To Reproduce

```python
from sunpy.timeseries import TimeSeries
from sunpy.data.sample import GOES_XRS_TIMESERIES
import matplotlib.pyplot as plt
ts = TimeSeries(GOES_XRS_TIMESERIES)
fig = plt.figure()
ax=fig.add_subplot(111)
ts.plot(axes=ax)
ax.set_ylim(1e-7,1e-5)
plt.show()
```

### What happened?

<img width="638" alt="image" src="https://user-images.githubusercontent.com/8676141/169153851-2b025675-92a8-4f4d-bc17-bd1e1b1018e1.png">

This is because we are not simultaneously adjusting the limits on the `twinx` axis that is added to display the labels. I'm not really sure if there is a sensible fix for this. Maybe we could return both axes from `plot` so that one could easily set the limits on both? It would seem weird to do this for only some sources and not others though. If there's a way to check (and return) the "twin" from an axes instance, that could work as well.

dstansby:
This is a Matplotlib bug I think? The following example shows the same behaviour, where manually changing the y limits of an axes doesn't change the y limits of any twinned axes:
```python
import matplotlib.pyplot as plt

fig = plt.figure()
ax1 = fig.add_subplot(111)
ax2 = ax1.twinx()
ax1.set_ylim(0, 2)
plt.show()
```

wtbarnes:
Ah interesting. Either way, it's a bit strange that in plot we create an axis that we don't return to the user. Even if with this bug, if we gave them the twin axis back, they could just set the y-lim the same and all would be (presumably) well.

PR #6486, opened 2022-10-19
Share y-axis limits on GOES plots

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/6163. A bit of a hacky fix, but as far as I can tell the only way to share limits in Matplotlib without having the same locator/formatter on both Axis instances.


hayesla:
I wonder could we simplify this part of the code completely and now use twinx but rather just ax.text it, for example, if we remove the `ax2 = axes.twinx()` part to label the GOES classes we could do:
```python
    values = [2.5e-8, 2.5e-7, 2.5e-6, 2.5e-5, 2.5e-4] # the 2.5 to plot in centre
    labels = ['A', 'B', 'C', 'M', 'X']
    for value, label in zip(values, labels):
        axes.text(1.02, value, label, transform=axes.get_yaxis_transform())
```

and seems to work well:
example_goes

and solves the issue with the ylims form #6163 (i.e. `ax.set_ylim(1e-7,1e-5)`)

example_goes_zoom

thoughts?

dstansby:
A cunning plan! That's much better, I've implemented it.

changelog/6486.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+The right-hand y-axis of GOES timeseries plots with labelled flare classes
```
hayesla, line 1:
```suggestion
The right-hand y-axis of the GOES-XRS timeseries plots with labelled flare classes
```

hayesla (approved):

dstansby:
Merging as a bugfix with one approval - I only updated the figure hashes since the positive review.

PR #6490, opened 2022-10-20
Fix tests that did not account for the Windows path separator

ayshih:
No has pointed out that a few of the tests fail on Windows because our CI doesn't run the online tests on Windows...


nabobalis (approved):

nabobalis:
I assume these pass locally?

ayshih:
Yes, the fixed tests passed on my Windows machine

PR #6491, opened 2022-10-20
Added graphviz to the conda environment file

ayshih:
Oops, left out `graphviz` in #6479.  `graphviz` is in fact one of the primary reasons that we'd want to utilize `conda`.  I also added two `pytest` extensions that are used by the test commands in our `tox` builds.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -125,6 +125,9 @@ deps =
 conda_env = sunpy-dev-env.yml
 install_command = pip install --no-deps --no-build-isolation {opts} {packages}
 allowlist_externals = conda
+commands_pre =
+#   Remove pytest-rerunfailures to avoid appearing to access the internet when pytest-xdist is used
+    conda remove -y -q --force pytest-rerunfailures
 commands =
     conda list
     {env:PYTEST_COMMAND} {posargs}
--- a/sunpy-dev-env.yml
+++ b/sunpy-dev-env.yml
@@ -40,10 +40,13 @@ dependencies:
   - pytest-doctestplus
   - pytest-mock
   - pytest-mpl
+  - pytest-rerunfailures
+  - pytest-timeout
   - pytest-xdist
   - tox
 
   # Documentation
+  - graphviz
   - ruamel.yaml
   - sphinx
   - sphinx-automodapi


nabobalis (approved):

nabobalis:
Huh:

```
INTERNALERROR> OSError: An attempt was made to connect to the internet by a test that was not marked `remote_data`. The requested host was:
Internet access disabled
```

ayshih:
This use of the network stack is set up at configuration, so it doesn't matter if the rerun option is not used.

tox.ini
```diff
@@ -125,6 +125,8 @@ deps =
 conda_env = sunpy-dev-env.yml
 install_command = pip install --no-deps --no-build-isolation {opts} {packages}
 allowlist_externals = conda
+# Remove pytest-rerunfailures to avoid appearing to access the internet when pytest-xdist is used
```
nabobalis, line 128: Can we not move this into the command block or will tox error?

ayshih:
I'm trying out the different approach of uninstalling `pytest-rerunfailures` prior to running the test c9mmand.

tox.ini
ayshih: I don't know: I still don't run tox locally, and I don't see guidance in the tox documentation.  Maybe it'd work fine?

nabobalis: Want me to try?

nabobalis:
Failure is due to `pytest-remotedata-0.3.2`, 0.3.3 fixed the deprecation warning but isn't on conda-forge?

nabobalis:
Wait, https://github.com/conda-forge/pytest-remotedata-feedstock/blob/main/recipe/meta.yaml is up to date.

ayshih:
Yeah, something about the removal is trigger a re-solve, and consequently a regression in `pytest-remotedata`...

ayshih:
No, no, I can get this to work. =)

nabobalis:
I can run any tox configs locally if you would like.

ayshih:
I thought making sure the solver prioritized `conda-forge` was a good thing, but apparently not!  Let's try skipping the solver entirely...

ayshih:
Yay, passing again

nabobalis (approved):

PR #6494, opened 2022-10-21
Fix the incorrect Python version for tox-conda test build

ayshih:
It turns out that when tox-conda creates a fresh environment from a YAML file (see #6479 and #6491), and that YAML file does not explicitly specify the Python version, the latest Python version is used (i.e., 3.10 currently).  For example, `py39-conda` tox build ends up having Python 3.10 instead of the intended 3.9.

Let's see if this PR works.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -122,10 +122,17 @@ commands =
 pypi_filter =
 extras =
 deps =
-conda_env = sunpy-dev-env.yml
+conda_deps =
+#   These packages are needed to install sunpy even with --no-deps
+    extension-helpers
+    numpy
+    setuptools-scm
+conda_channels = conda-forge
 install_command = pip install --no-deps --no-build-isolation {opts} {packages}
 allowlist_externals = conda
 commands_pre =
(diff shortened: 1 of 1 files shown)


nabobalis:
We need to check if `tox-conda` is still being supported. Last time I checked a long time ago, it seemed dead.

ayshih:
There may not be a whole lot of activity in `tox-conda`, but I don't think it's dead dead.  There's currently an open PR (which is actually somewhat relevant to my headache here) with active engagement as of this week.

nabobalis:
> There may not be a whole lot of activity in `tox-conda`, but I don't think it's dead dead. There's currently an open PR (which is actually somewhat relevant to my headache here) with active engagement as of this week.

Thats good to know

ayshih:
This solution is admittedly a bit clunky, but fortunately I believe that it can be reverted if tox-dev/tox-conda#137 is accepted

nabobalis (approved):

PR #6497, opened 2022-10-21
Move GOES XRS tutorial to the guide

dstansby:
I think I moved this to a `tutorials` folder in a previous PR with the intention of putting it in a tutorials section - I guess we already have that as the Guide, so I've moved it to there so we don't lose it when 4.1 is released!


nabobalis:
One odd docfailure:

```
/home/runner/work/sunpy/sunpy/docs/guide/acquiring_data/goes_xrs.rst:155: WARNING: py:obj reference target not found: concatenate=True
```

dstansby:
I'm struggling on getting the plotting working here, so in the interests of getting 4.1 out the door am going to put back the example gallery example in another PR.

PR #6499, opened 2022-10-21
Add unit to header helper

### LINKED ISSUE #6219: `make_fitswcs_header` should have an option to set the unit
### Describe the feature

`make_fitswcs_header` provides kwargs for setting several non-WCS keywords such as wavelength, exposure time, instrument, etc. We should also provide the ability to set the unit by setting the `"BUNIT"` key. 

### Proposed solution

Add a `unit=...` kwarg that takes in either an `astropy.Unit` or a `str` that is then used to set the `"BUNIT"` key in the new header.

wtbarnes:
Fixes #6219 

This also adds some tests for other instrument metadata options that, as far as I can tell, we were not testing before.

### DIFF
--- a/changelog/6499.feature.rst
+++ b/changelog/6499.feature.rst
@@ -1,0 +1,2 @@
+`~sunpy.map.make_fitswcs_header` now includes the keyword argument `unit` for setting the
+``BUNIT`` FITS keyword in the resulting header.
--- a/sunpy/map/header_helper.py
+++ b/sunpy/map/header_helper.py
@@ -123,7 +123,7 @@ def make_fitswcs_header(data, coordinate,
         if unit is None:
             unit = data.unit
         else:
-            log.info("Overwriting Quantity's current unit with specified unit.")
+            log.info("Overwriting data's current unit with specified unit.")
 
     meta_wcs = _get_wcs_meta(coordinate, projection_code)
 
--- a/sunpy/map/header_helper.py
+++ b/sunpy/map/header_helper.py
@@ -37,7 +37,8 @@ def make_fitswcs_header(data, coordinate,
                         observatory=None,
                         wavelength: u.angstrom = None,
                         exposure: u.s = None,
-                        projection_code="TAN"):
+                        projection_code="TAN",
+                        unit=None):
     """
     Function to create a FITS-WCS header from a coordinate object
     (`~astropy.coordinates.SkyCoord`) that is required to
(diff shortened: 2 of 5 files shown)


changelog/6499.feature.rst
```diff
@@ -0,0 +1,2 @@
+`~sunpy.map.make_fitswcs_header` now includes the keyword argument `unit` for setting the
```
dstansby, line 1:
```suggestion
`~sunpy.map.make_fitswcs_header` now includes the keyword argument ``unit`` for setting the
```

dstansby (approved):
Once the doc builds pass (should do with my fix), 👍

sunpy/map/header_helper.py
```diff
@@ -37,7 +37,8 @@ def make_fitswcs_header(data, coordinate,
                         observatory=None,
                         wavelength: u.angstrom = None,
                         exposure: u.s = None,
-                        projection_code="TAN"):
+                        projection_code="TAN",
+                        unit=None):
```
Cadair, line 41: Just to complicate this, what if `data` is a quantity? Should we use that unit?

wtbarnes: I like that idea...but I wonder if it would be confusing given the fact that you cannot initiate a map with a `Quantity` (or at least, we don't explicitly support that).

sunpy/map/header_helper.py
```diff
@@ -116,6 +119,12 @@ def make_fitswcs_header(data, coordinate,
     else:
         shape = data
 
+    if hasattr(data, "unit"):
+        if unit is None:
+            unit = data.unit
+        else:
+            log.info("Overwriting Quantity's current unit with specified unit.")
```
Cadair, line 126:
```suggestion
            log.info("Overwriting data's current unit with specified unit.")
```

Cadair (approved):

changelog/6499.feature.rst
```diff
@@ -0,0 +1,2 @@
+`~sunpy.map.make_fitswcs_header` now includes the keyword argument ``unit`` for setting the
```
Cadair, line 1: Oh we should add another changelog entry for the data unit

sunpy/map/header_helper.py
wtbarnes: Ok, I've added the option to do this as well. It's pretty trivial. I've taken the NDCube approach of the `unit=` kwarg taking precedence if units are specified on `data` as well. This gets logged to `INFO` if that's the case.

sunpy/map/header_helper.py
wtbarnes: https://github.com/astropy/astropy/blob/1ff5c4e6c89db611248427fb961210b70732155e/astropy/nddata/nddata.py#L193

🤷

sunpy/map/header_helper.py
wtbarnes: @dstansby you may want to have another peek at this since you approved before this addition.

nabobalis (approved):

PR #6502, opened 2022-10-24
Optionally preserve metadata in `reproject_to`

### LINKED ISSUE #5627: `reproject_to` does not preserve metadata
This is not a bug as I'm quite sure this is the intended behavior, but when using the new `reproject_to` method, the resulting map does not have much of the associated metadata of the original map as the resulting map is reconstructed from a data-header pair where the header is reconstructed only from the WCS.

For example, after reprojecting a map, the resulting map has no  units,

```python
>>> import sunpy.map
>>> import sunpy.data.sample
>>> m = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
>>> m.unit
Unit("ct")
>>> m.reproject_to(m.wcs).unit is None
True
```

One possible solution would be to instead create a header using the header helper and then creating the resulting map from a `(data,header)` pair instead.

Furthermore, I just noticed that this returns a `GenericMap` instance explicitly. Given the approach described above, could we instead return map with `_new_instance`?

wtbarnes:
Fixes #5627 

This uses the header helper to optionally pass through a subset of the instrument keywords when using `reproject_to` and also ensures that the new map is the same subclass as the old map. As discussed in the above issue, this is not always a correct assumption so this False by default.

I'm not completely sure about the best way to do this. This should also wait on #6499 so that the unit can be easily passed through to the new instance.

## TODOs 

- [x] Incorporate changes from #6499 
- [ ] ~~Incorporate changes from #6503~~
- [x] Tests
- [x] Changelog
- [x] docstring

### DIFF
--- a/changelog/6502.feature.rst
+++ b/changelog/6502.feature.rst
@@ -1,0 +1,2 @@
+`~sunpy.map.GenericMap.reproject_to` now optionally preserves a subset of the metadata from the input map
+if ``preserve_meta=True``.
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -50,7 +50,6 @@ from sunpy.util.decorators import (
     add_common_docstring,
     cached_property_based_on,
     check_arithmetic_compatibility,
-    deprecate_positional_args_since,
 )
 from sunpy.util.exceptions import warn_metadata, warn_user
 from sunpy.util.functools import seconddispatch
@@ -2731,18 +2730,12 @@ class GenericMap(NDData):
         # Reconstruct header
         target_header = MetaDict(target_wcs.to_header())
         if preserve_meta:
-            from .header_helper import _set_instrument_meta
-            target_header = _set_instrument_meta(target_header,
-                                                 self.instrument,
-                                                 self.meta.get('telescop', None),
-                                                 self.observatory,
-                                                 self.wavelength,
-                                                 self.exposure_time,
-                                                 self.unit)
+            original_header = copy.deepcopy(self.meta)
+            original_header.update(target_header)
+            target_header = original_header
 
         # Create and return a new GenericMap
-        outmap = GenericMap(output_array, target_header,
-                            plot_settings=self.plot_settings)
+        outmap = self._new_instance(output_array, target_header, plot_settings=self.plot_settings)
 
         if return_footprint:
             return outmap, footprint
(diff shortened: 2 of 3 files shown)


ayshih:
> the new map is the same subclass as the old map

I recognize the intent, but now that we no longer modify/fix the metadata on instantiation (#5501), I think you'll have to figure out how to support those Map subclasses that construct their properties from non-standard keywords (or from standard keywords in a non-standard way).  Skipping the instantiation was previously possible, but now the subclass properties will need to know whether they ought to be reading the metadata in the source-specific way or in the default way.

wtbarnes:
Oh good point. I had not even thought about that...

Thinking about it a little more, I think it's fine to just do away with instantiating through the specific subclass and just route the new map creation through `GenericMap`. As long as the needed metadata is accessible (e.g. `wavelength` in David's example in #5627), then that's the only thing that matters. The purpose of the source-specific subclasses is to normalize the representation of the metadata. If we've already done that via the header helper, then there's no need to use the source specific subclass.

Cadair (changes requested):
this significantly deviates the behaviour of this method away from ndcube which I am not a massive fan of. I need to think about this in more detail.

ayshih:
Some thoughts:

- The name of the keyword should probably be changed to something less seemingly comprehensive than `preserve_meta`, e.g., `preserve_core_meta` or `preserve_basic_meta`.
- Of the metadata that is being preserved, `unit` gives me pause, as I've mentioned before.
- As @Cadair notes, there continues to be a disparity between the two `reproject_to()` methods.  Amusingly, you would have moved our `reproject_to()` closer to NDCube's had you kept the instantiation of the subclass.

wtbarnes:
Another option that would address the NDCube API consistency issue: `preserve_meta` would preserve *every* non-WCS keyword (or any keyword that is not included from the new header) from the old header and then instatiate the new map through `_new_instance`.

This makes me a little nervous because there are many extra keywords in headers that then may become stale (e.g. ones related to statistics on the data), but this happens with our other methods like resample, rotate, etc and the properties exposed by `GenericMap` does not depend on them. We would just have to make absolutely sure that there are no stale keywords left around that are used by `GenericMap`

ayshih (commented):
Doesn't your current implementation fail to strip out all WCS keywords from the original map header before merging in WCS keywords from the target header?  It seems like it'd be easy to end up with a output map header that contains inconsistent metadata (e.g., a PCij matrix not matching a CROTA2 value).

wtbarnes:
I think I've been convinced that this is just not a viable option because of the way we treat our metadata, both in terms of our use of source-specific classes and our decision to not modify the underlying metadata. There is no way to reconcile this with propagating `type(class)` when creating a reprojected map. This works fine in NDCube because of the strict separation between WCS and non-WCS metadata.

Issue #6505, opened 2022-10-25
Tests failing with pytest 7.2.0


dstansby:
### Describe the bug

With `pytest` 7.2.0 the following tests are failing:

```python
 FAILED ../../.tox/py310/lib/python3.10/site-packages/sunpy/map/tests/test_map_factory.py::test_data_filenames
FAILED ../../.tox/py310/lib/python3.10/site-packages/sunpy/io/tests/test_fits.py::test_data_filenames
FAILED ../../.tox/py310/lib/python3.10/site-packages/sunpy/map/sources/tests/test_eit_source.py::test_data_filenames
FAILED ../../.tox/py310/lib/python3.10/site-packages/sunpy/physics/tests/test_differential_rotation.py::test_differential_rotation
FAILED ../../.tox/py310/lib/python3.10/site-packages/sunpy/timeseries/tests/test_timeseries_factory.py::test_data_filenames
```

with messages like:
```python
which will be an error in a future version of pytest.  Did you mean to use `assert` instead of `return`?
```

I'm guessing these tests are unintentionally returning something?

I'm going to fix this by pinning `pytest<7.2` on the 4.0 branch to get 4.0.6 out, but when a fix is merged to main and backported we should remember to remove that pytest pin on the 4.0 branch.

### To Reproduce

import sunpy

sunpy.map.Map(...)
etc

dstansby:
We're now getting this mysterious warning with the new release of `pytest-xdist` (https://github.com/pytest-dev/pytest-xdist/pull/828 is the PR where the warning came in I think):

```python
INTERNALERROR>     warnings.warn(warning, stacklevel=stacklevel)
INTERNALERROR> DeprecationWarning: The --rsyncdir command line argument and rsyncdirs config variable are deprecated.
INTERNALERROR> The rsync feature will be removed in pytest-xdist 4.0.
```

ConorMacBride:
The `test_data_filenames` errors have been fixed by #6507.

`test_differential_rotation` needs https://github.com/astropy/pytest-arraydiff/pull/36 to be released, however, I don't think we are actually running pytest-arraydiff. It needs to be enabled with `--arraydiff`. Do we want to enable it? When I do that the test fails with:
```
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 10 / 16384 (0.061%)
Max absolute difference: 0.001
Max relative difference: 5.55484577e-06
```

nabobalis:
[...]

PR #6507, opened 2022-10-25
rename test_data_filenames to get_test_data_filenames

nabobalis:
xref https://github.com/sunpy/sunpy/issues/6505 

Pytest 7.2 now raises a warning if a function it considers a test is returning anything. 

This happens due to `test_data_filenames` and the fact that pytest looks for `test*` functions.

We could exclude it or just in this case, rename it.
I don't consider this a public API change, this function is only used to get the sunpy test data. Other packages should not be using this.

### DIFF
--- a/docs/dev_guide/contents/code_standards.rst
+++ b/docs/dev_guide/contents/code_standards.rst
@@ -190,7 +190,7 @@ Data and Configuration
 ======================
 
 * We store test data in ``sunpy/data/test`` as long as it is less than about 100 kB.
-  These data should always be accessed via the :func:`sunpy.data.test.get_test_filepath` and :func:`sunpy.data.test.test_data_filenames` functions.
+  These data should always be accessed via the :func:`sunpy.data.test.get_test_filepath` and :func:`sunpy.data.test.get_test_data_filenames` functions.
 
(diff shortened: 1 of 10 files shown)


ConorMacBride:
You just beat me to it! `pytest-arraydiff` needs updated to support pytest 7.2. I'll open a PR there to convert it to work similar to pytest-mpl.

nabobalis:
> You just beat me to it! `pytest-arraydiff` needs updated to support pytest 7.2. I'll open a PR there to convert it to work similar to pytest-mpl.

Should I stick with renaming or should I just skip it?

ConorMacBride:
I think renaming would be best? As you see in #6505, it will probably show up four times as being skipped.

nabobalis:
If anyone has a better name for the function as well, I am happy to rename it more.

ConorMacBride:
I was going to name it `get_test_data_filenames`. I thought `get` would work since it needs to be called. It's fine as is though.

nabobalis:
> I was going to name it `get_test_data_filenames`. I thought `get` would work since it needs to be called. It's fine as is though.

That is a better name, I will do that.

ConorMacBride (approved):

Issue #6510, opened 2022-10-26
Downloading cutouts example failing on gallery docs build


dstansby:
### Describe the bug

In the gallery docs build the "downloading AIA cutouts" example is failing very often at the moment with the error message:
```python
 generating gallery for generated/gallery/acquiring_data... [ 80%] downloading_cutouts.py
WARNING: /home/runner/work/sunpy/sunpy/examples/acquiring_data/downloading_cutouts.py failed to execute correctly: Traceback (most recent call last):
  File "/home/runner/work/sunpy/sunpy/examples/acquiring_data/downloading_cutouts.py", line 34, in <module>
    amap = sunpy.map.Map(files)
  File "/home/runner/work/sunpy/sunpy/.tox/build_docs-gallery/lib/python3.8/site-packages/sunpy/map/map_factory.py", line 340, in __call__
Error:     raise RuntimeError('No maps loaded')
RuntimeError: No maps loaded
```

I would propose we disable this example from running in the gallery for now, unless someone can come up with a quick fix. Whenver I try and run this example locally it seems to work fine.

### To Reproduce

import sunpy

sunpy.map.Map(...)
etc

dstansby:
This is happening again...

Issue #6146, opened 2022-05-13
Use the `cmap` and `norm` kwargs in our example gallery animations


wtbarnes:
### Provide a general description of the issue or problem.

We have a few gallery examples that show animations. Previously, to get each time step in the animation to have the same norm, this had to be set manually for each one in `plot_settings`. Now, you can pass `norm` directly to `.plot()`. Those examples should be cleaned up to reflect this

Effected examples:

- https://docs.sunpy.org/en/stable/generated/gallery/acquiring_data/downloading_cutouts.html#sphx-glr-generated-gallery-acquiring-data-downloading-cutouts-py
- https://docs.sunpy.org/en/stable/generated/gallery/map/difference_images.html#sphx-glr-generated-gallery-map-difference-images-py

PR #6511, opened 2022-10-26
Remove HMI download and update cutout example to remove first image download

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/6510
Fixes https://github.com/sunpy/sunpy/issues/6146

### DIFF
--- a/examples/acquiring_data/downloading_cutouts.py
+++ b/examples/acquiring_data/downloading_cutouts.py
@@ -6,7 +6,6 @@ Requesting cutouts of AIA images from the JSOC
 This example shows how to request a cutout of a series of
 AIA images from the JSOC.
 """
-# sphinx_gallery_thumbnail_number = 2
 import os
 
 import matplotlib.pyplot as plt
@@ -32,6 +31,7 @@ query = Fido.search(
 )
 file = Fido.fetch(query)
 amap = sunpy.map.Map(file)
+
 #####################################################
 # Next, we will use the coordinate frame from this map to define the top right and bottom
 # left coordinates we want for the cutout request.
--- a/examples/acquiring_data/downloading_cutouts.py
+++ b/examples/acquiring_data/downloading_cutouts.py
@@ -15,29 +15,18 @@ from astropy.coordinates import SkyCoord
 from astropy.time import Time
 from astropy.visualization import ImageNormalize, SqrtStretch
 
+import sunpy.coordinates  # NOQA
 import sunpy.map
 from sunpy.net import Fido
 from sunpy.net import attrs as a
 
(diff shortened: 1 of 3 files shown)


examples/acquiring_data/downloading_cutouts.py
```diff
@@ -23,15 +23,15 @@
 #####################################################
 # First, query a full frame AIA image.
 
-t0 = astropy.time.Time('2012-09-24T14:56:03', scale='utc', format='isot')
+start_time = Time('2022-01-01T00:00:00', scale='utc', format='isot')
```
nabobalis, line 26: I am hoping that a more recent set of data might help with the failure of the download.

wtbarnes: I am 👎 on this change. As I said on the call yesterday, the date and coordinates of the cutout were chosen specifically to follow an active region observed on that date. This is often what people use the cutout service for. The example is less compelling with a random date/cutout.

nabobalis: Fair, I will update the time and crop region to be an AR, would that be ok?

nabobalis: I have updated the crop to follow an AR, it has some nice loops.

nabobalis: Also should display now on the gallery page.

dstansby (approved):

dstansby:
This should get two reviews!

nabobalis:
Gallery thumbnail is now back.

examples/acquiring_data/downloading_cutouts.py
```diff
@@ -74,13 +74,15 @@
...
 files.sort()
 
 #####################################################
 # Now that we've downloaded the files, we can create
 # a `~sunpy.map.MapSequence` from them and animate
 # them.
+
 sequence = sunpy.map.Map(files, sequence=True)
 # Make sure the colorbar limits are the same for each image
 for each_map in sequence:
```
wtbarnes, line 88: Since we're here, can you remove this for loop and just put the `norm=` directly in the `sequence.plot` call? When I first wrote this example, that didn't work.

nabobalis: Give me a minute.

wtbarnes (approved):
One peripheral change, but otherwise looks good to me.

examples/acquiring_data/downloading_cutouts.py
nabobalis: Removed.

PR #6512, opened 2022-10-27
Python 3.11 Support

### LINKED ISSUE #6070: “Dead batteries” in SunPy

[PEP 594–Removing dead batteries from the standard library](https://peps.python.org/pep-0594/) identifies a number of modules in the standard library to be deprecated and eventually removed. We rely on several of these in SunPy:

- cgi (via par five)
- xdrlib (in sunpy.io.special.read_genx)

Both of these have no alternative package listed and are scheduled for deprecation in 3.11 and removal in 3.13. There may be others that I’ve missed. Please edit the list above if so. 

Ideally prior to deprecation and certainly before removal, we need to think about how to deal with these changes in the standard library.

As a first cut, we should identify what functionality we are using exactly from these modules. If it is only one function, is it something we can just vendor? If it is larger, is there some alternative package we can rely on in its place?

This is more of a longer term discussion but the sooner we start thinking about this the better.

Cadair:
Fixes #6070 

This pull request adds support for Python 3.11. Currently I am only testing 3.11 with devdeps, because there isn't a reproject release which supports 3.11 but also probably other packages as well.

The main changes in this pull request are:

* Vendor in `cgi.parse_header` and use it in vso and our docs.
* Vendor in `xdrlib.Unpacker`
* Fix `sunpy.util.metadata.MetaDict.popitem` after the 3.11 implementation moved away from using `__del__`.

There are no h5py wheels for 3.11 yet, which is a little problem here, but they are close: https://github.com/h5py/h5py/pull/2165

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,13 +145,11 @@ jobs:
       test_extras: 'dev'
       test_command: 'pytest -p no:warnings --doctest-rst -m "not mpl_image_compare" --pyargs sunpy'
       submodules: false
-      # This can be removed when there are binaries for h5py on 3.11
-      libraries: 'libhdf5-dev'
       targets: |
         - cp3{8,9,10}-manylinux*_x86_64
         - cp3{8,9,10}-macosx_x86_64
         - cp3{8,9,10}-macosx_arm64
-        # Split these out as building h5py is slower, can be reintegrated when binaries for h5py exist
+        # Split these out for more rapid debugging
         - cp311-manylinux*_x86_64
         - cp311-macosx_x86_64
         - cp311-macosx_arm64
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,11 +44,14 @@ jobs:
       libraries: |
         apt:
           - libopenjp2-7
+          # This can be removed when there are binaries for h5py on 3.11
+          - libhdf5-dev
         brew:
           - openjpeg
       envs: |
-        - macos: py38
+        - linux: py311
         - windows: py39
+        - macos: py38
         - linux: py38-oldestdeps
 
   docs:
@@ -115,9 +118,11 @@ jobs:
       libraries: |
         apt:
           - libopenjp2-7
+          # This can be removed when there are binaries for h5py on 3.11
+          - libhdf5-dev
       envs: |
         - linux: base_deps
-        - linux: py39-devdeps
+        - linux: py311-devdeps
         - linux: py39-conda
           libraries: ''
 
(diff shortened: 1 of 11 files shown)


nabobalis:
Don't we need to?

- Add 11 here: https://github.com/sunpy/sunpy/blob/main/.github/workflows/ci.yml#L143 
- Add 11 here: https://github.com/sunpy/sunpy/blob/main/setup.cfg#L32

nabobalis (approved):

.github/workflows/ci.yml
```diff
@@ -44,12 +44,16 @@ jobs:
...
           - libopenjp2-7
+          - libhdf5-dev
         brew:
           - openjpeg
       envs: |
         - macos: py38
         - windows: py39
         - linux: py38-oldestdeps
+        # This job should be put back into cron when we can have a non-devdeps
+        # 3.11 job
```
dstansby, line 55:
```suggestion
        # When a 3.11 compatible release of reproject is made the devdeps job in cron should be re-enabled
        # and this job should be changed to a normal non-devdeps job.
```

dstansby: To be a bit more explicit

changelog/6512.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+Add support for Python 3.11.
+
+The deprecated `cgi.parse_header` is now available as
+`sunpy.util.net.parse_header`.
```
dstansby, line 4: I'm not sure we want to be advertising this publicly in a changelog since the new vendored code isn't public API? I'd suggest just removing the changelog altogether.

dstansby (changes requested):
I think the current changelog entry should be removed (see my inline comment), and a new one added that advertises support for 3.11. Otherwise looks 👍

changelog/6512.bugfix.rst
Cadair: We document it in the Fido dev guide so it's public api to my mind.

.github/workflows/ci.yml
Cadair: It's probably not just reproject (although it definitely *is* reproject). I have just tagged a reproject bugfix release with 3.11 support, so we can find out shortly lol.

Cadair: I have now removed this comment after a release of reproject.

Cadair:
I think this is now good, but I have added *all the CI* to make sure (particularly the publish stuff).

dstansby (approved):

wtbarnes:
Should this get a What's New entry as well?

Cadair:
> Should this get a What's New entry as well?

I guess it should be mentioned in there, but it is getting backported as well.

PR #6518, opened 2022-10-28
[4/0] Backport #6511

nabobalis:
I did not remove the old HMI example since people have linked to it recently and breaking that URL on this LTS release I don't feel is ok.


examples/acquiring_data/downloading_hmi.py
```diff
@@ -1,67 +0,0 @@
-"""
```
nabobalis, line 1: Unsure how I feel about deleting this in the backport.

I am tempted to restore it.

PR #6521, opened 2022-10-30
Fixing issues related to tox-conda

ayshih:
For our cron CI builds between the 25th and the 27th this past week, `pytest` went from 7.1.3 to 7.2.0 and `tox` went from 3.26.0 to 3.27.0, but also `tox-conda` went from 0.9.2 to 0.10.0.  Probably due to the new `tox-conda` release, but maybe due to something else, our `py39-conda` build is once again running Python 3.10 instead of the intended 3.9.  This PR reverts my workaround in #6494, and as predicted in that PR, the new `tox-conda` release no longer needs to be worked around.

Also, adding `pytest-arraydiff` as a direct dependency (it was already an indirect dependency through `pytest-astropy`)

Also, pinning `pytest` and `pytest-xdist` a la #6508 (see also #6505)

Also, adding `hvpy` as a dependency, via `pip` until it is available on `conda-forge`.  `hvpy` has a dependency (`pydantic`) that is not otherwise installed in the conda environment, so temporarily explicitly install `pydantic`.

### DIFF
--- a/sunpy-dev-env.yml
+++ b/sunpy-dev-env.yml
@@ -35,6 +35,7 @@ dependencies:
   # Testing
   - hypothesis
   - pytest
+  - pytest-arraydiff
   - pytest-astropy
   - pytest-cov
   - pytest-doctestplus
@@ -44,9 +45,11 @@ dependencies:
   - pytest-timeout
   - pytest-xdist
   - tox
+  - tox-conda
 
   # Documentation
   - graphviz
+  - hvpy
   - ruamel.yaml
   - sphinx
   - sphinx-automodapi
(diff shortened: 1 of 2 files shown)


sunpy-dev-env.yml
```diff
@@ -35,6 +35,7 @@ dependencies:
   # Testing
   - hypothesis
   - pytest
+  - pytest-arraydiff
```
ConorMacBride, line 38: This also needs to be enabled by passing `--arraydiff` to the pytest command.

ayshih: I'm adding it to the environment file in anticipation (and it doesn't matter because it's already installed anyway due to `pytest-astropy`).  Actually enabling it our tox commands requires me to fix that one test.

nabobalis (approved):

PR #6523, opened 2022-10-31
Re-enabled the use of pytest-arraydiff

ayshih:
As discussed in #6505, at some point we unintentionally omitted enabling the array comparison of `pytest-arraydiff`, via the `--arraydiff` option.  Thus, we missed that our one array-comparison test should have been failing and needed to be updated.  This PR fixes both things.


dstansby (approved):

Cadair:
I should have approved this first, but take me merging it as an approval :see_no_evil:

Issue #6530, opened 2022-10-31
Remove `use_scipy` argument in `transform.py`


dstansby:
### Provide a general description of the issue or problem.

In `sunpy.image.transform.py`: 

```python
warn_deprecated("The 'use_scipy' argument is deprecated. "
                        "Specify the rotation method to the 'method' "
                        "keyword argument instead.")
```

This was deprecated in 4.0 so should be removed in 4.1

PR #6538, opened 2022-10-31
Remove use_scipy argument in rotate functions

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/6530

### DIFF
--- a/changelog/6538.removal.rst
+++ b/changelog/6538.removal.rst
@@ -1,0 +1,1 @@
+:meth:`sunpy.image.resample` will no longer accept "neighbour" as an interpolation method.
--- a/changelog/6538.removal.rst
+++ b/changelog/6538.removal.rst
@@ -1,1 +1,1 @@
-:meth:`sunpy.image.resample` will no longer accept "neighbour" as an interpolation method.
+:meth:`sunpy.image.transform.affine_transform` and :func:`sunpy.map.GenericMap.rotate` will no longer accept the ``use_scipy`` keyword.
--- a/changelog/6538.removal.rst
+++ b/changelog/6538.removal.rst
@@ -1,1 +1,1 @@
-:meth:`sunpy.image.transform.affine_transform` and :func:`sunpy.map.GenericMap.rotate` will no longer accept the ``use_scipy`` keyword.
+:meth:`sunpy.image.transform.affine_transform` and :func:`sunpy.map.GenericMap.rotate` no longer accepts the ``use_scipy`` keyword.
--- a/sunpy/image/transform.py
+++ b/sunpy/image/transform.py
@@ -12,13 +12,13 @@ import scipy.ndimage
 from scipy.signal import convolve2d
 
 from sunpy import log
-from sunpy.util.exceptions import warn_deprecated, warn_user
+from sunpy.util.exceptions import warn_user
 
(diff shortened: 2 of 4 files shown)


sunpy/image/transform.py
```diff
@@ -90,35 +90,19 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
...
+    method = _get_transform_method(method)
 
     # Transform the image using the appropriate function
     rotated_image = _rotation_registry[method].function(image, rmatrix, shift, order, missing, clip)
 
     return rotated_image
 
 
-def _get_transform_method(method, use_scipy):
+def _get_transform_method(method):
```
nabobalis, line 101: I didn't really want to remove this function as it was more "work" but as it stands it doesn't do anything that can't be copy and pasted into each method.

dstansby (approved):

changelog/6538.removal.rst
```diff
@@ -0,0 +1 @@
+:meth:`sunpy.image.transform.affine_transform` and :func:`sunpy.map.GenericMap.rotate` will no longer accept the ``use_scipy`` keyword.
```
wtbarnes, line 1:
```suggestion
:meth:`sunpy.image.transform.affine_transform` and :func:`sunpy.map.GenericMap.rotate` no longer accepts the ``use_scipy`` keyword.
```

sunpy/image/transform.py
```diff
@@ -90,35 +90,19 @@ def affine_transform(image, rmatrix, order=3, scale=1.0, image_center=None,
...
-        if use_scipy is True and method != 'scipy':
-            warn_user(f"Using scipy instead of {method} for rotation.")
-            method = 'scipy'
-
-    if method == 'scikit-image':
-        try:
-            import skimage  # NoQA
-        except ImportError:
-            raise ImportError("scikit-image must be installed to be usable for rotation.")
-
```
wtbarnes, line 121: Why move this block out of here? Isn't the point of the `_get_method` function to do validation of the choice of method all in one place?

sunpy/image/transform.py
```diff
@@ -336,7 +320,10 @@ def _rotation_skimage(image, matrix, shift, order, missing, clip):
     * Does not pass NaN as ``missing`` to :func:`~skimage.transform.warp` due to
       inconsistent handling across interpolation orders
     """
-    import skimage.transform
+    try:
+        import skimage.transform
+    except ImportError:
+        raise ImportError("The scikit-image package is required to use this rotation method.")
```
wtbarnes, line 326: See comment above

wtbarnes (approved):
Just a few comments, but nothing to block on. I'm just a bit confused by the reorganization of things.

sunpy/image/transform.py
nabobalis: The rotation method for opencv check for the import inside of them and I think they should handle it.

The same goes for cupy, if it was done here, someone would need to PR it to core sunpy.

wtbarnes: Ah ok, thanks. That makes sense. So `_get_method` should just check the registry and return the desired method.

nabobalis: That is what it does now only, before it also did the import check only for scikit image.

Issue #6528, opened 2022-10-31
Remove `GenericTimeSeries.index`


dstansby:
### Provide a general description of the issue or problem.

`GenericTimeSeries.index` was deprecated in 4.0, so we should remove it in 4.1

dstansby:
My bad, looks like we deprecated this in 4.1...

PR #6539, opened 2022-10-31
Remove GenericTimeSeries.index

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/6528

Things will break.

- [x] Fix that example.


sunpy/timeseries/timeseriesbase.py
```diff
@@ -313,7 +304,7 @@ def _repr_html_(self):
         for i in range(len(self.columns)):
             if len(self.columns) == 1:
                 axs.plot(
-                    dat.index,
+                    da.indext,
```
wtbarnes, line 307: Is this a typo?

changelog/6327.deprecation.rst
```diff
@@ -1,3 +1,3 @@
-Using `sunpy.timeseries.GenericTimeSeries.index` is deprecated.
```
wtbarnes, line 1: The filename is wrong here.

wtbarnes (commented):
There were some problems with the doc build, maybe your most recent commits fix those? I think they were mostly just broken references.

Are there other things you're expecting to break besides that example?

sunpy/timeseries/timeseriesbase.py
nabobalis: Damn it, how did I miss that.

changelog/6327.deprecation.rst
nabobalis: This is an old changelog we didnt remove?!

nabobalis, line 1: Why is this changelog still on this branch?

nabobalis:
> There were some problems with the doc build, maybe your most recent commits fix those? I think they were mostly just broken references.

Those refs should be fixed now.
I am confused about that double changelog.

> Are there other things you're expecting to break besides that example?

No.

nabobalis:
I don't think we did deprecate this in 4.0, I think the only thing that should last here is the example.

dstansby:
My bad, we only deprecated this in 4.1. I think this PR can be closed then? Sorry again...

nabobalis:
The example!

PR #6541, opened 2022-10-31
Update vendored code for 4.1

dstansby:


nabobalis (approved):

nabobalis:
We need to check why the cronjob(s) failed before merge.

nabobalis:
Base deps failed with:

ModuleNotFoundError: No module named 'pydantic'

That shouldn't be needed. 

Conda failed with:

DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.

Due to the pinned versions of test dependencies, arraydiff being the problem here.
This should be fixed separately. 

Online and gallery are known fails.


None of these fails are due to the vendored changes so we can merge without a problem.

dstansby:
The fail is real, it looks like `inflect` gained `pydantic` as a dependency. I think we should just update our vendorded version of `inflect` to the first version that didn't have a `pydantic` dep.

nabobalis:
We will need to think about how to deal with this in the future.

Cadair:
> We will need to think about how to deal with this in the future.

We should open an issue.

nabobalis (approved):

Cadair (approved):

PR #6542, opened 2022-10-31
Update stability matrix for 4.1

dstansby:
@sunpy/sunpy-developers I will give you until the end of the week to comment and suggest other changes!

### DIFF
--- a/docs/code_ref/sunpy_stability.yaml
+++ b/docs/code_ref/sunpy_stability.yaml
@@ -9,7 +9,7 @@ database:
   comments: Pending deprecation. See https://community.openastronomy.org/t/deprecating-sunpy-database/495/2 for community discussion.
 image:
   status: stable
-  comments: While it has not changed recently, it could be subject to major revisions in future.
+  comments: Any future changes should largely maintain backwards compatibility.
 io:
   status: mature
   comments: Should see no major changes.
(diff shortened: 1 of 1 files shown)


docs/code_ref/sunpy_stability.yaml
```diff
@@ -5,10 +5,10 @@ data:
   status: mature
```
nabobalis, line 5: I wonder since we added a new method and will remove the old one. Maybe this should be stable?

docs/code_ref/sunpy_stability.yaml
```diff
@@ -5,10 +5,10 @@ data:
...
   comments: Should see no major changes.
 database:
-  status: dev
-  comments: While it has not changed recently, it could be subject to major revisions in future.
+  status: pendingdep
+  comments: Pending deprecation. See https://community.openastronomy.org/t/deprecating-sunpy-database/495/2 for community discussion.
 image:
-  status: dev
+  status: stable
   comments: While it has not changed recently, it could be subject to major revisions in future.
```
Cadair, line 12: If we are happy with what's there should we not change this description?

dstansby: 👍 I forgot, will do

wtbarnes (approved):

Cadair (approved):

PR #6547, opened 2022-11-01
Update conda env with sphinx-copybutton

ayshih:
I missed that `sphinx-copybutton` was recently added as a dependency

### DIFF
--- a/sunpy/map/sources/tests/test_mdi_source.py
+++ b/sunpy/map/sources/tests/test_mdi_source.py
@@ -64,7 +64,6 @@ def test_unit(mdi):
     assert mdi.unit == u.dimensionless_unscaled
 
 
-@pytest.mark.filterwarnings("error")
 def test_synoptic_source(mdi_synoptic):
     assert isinstance(mdi_synoptic, MDISynopticMap)
     # Check that the WCS is valid
--- a/sunpy-dev-env.yml
+++ b/sunpy-dev-env.yml
@@ -52,6 +52,7 @@ dependencies:
   - ruamel.yaml
   - sphinx
   - sphinx-automodapi
+  - sphinx-copybutton
   - sphinx-gallery
   - sphinx-design
   - towncrier


nabobalis:
We have too many dependencies!

nabobalis (approved):

nabobalis:
The conda builds started failing with:

E       DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.


What I don't understand is why now. I swear the last conda PR I merged passed.

ayshih:
It did pass.  But then afterwards we enabled `--arraydiff`, and we didn't run the conda build on that PR (because why should that fail!)

ayshih:
Aren't we filtering out this warning? (https://github.com/sunpy/sunpy/blob/87bcf39d7551ddd48e0f22419aa39cc5f229e378/setup.cfg#L158)  Something is defeating the filtering?

ayshih:
Why is this line?
https://github.com/sunpy/sunpy/blob/87bcf39d7551ddd48e0f22419aa39cc5f229e378/sunpy/map/sources/tests/test_mdi_source.py#L67

ayshih:
It's the only such line that I see in our entire codebase.  I'm killing it.

ayshih:
Yay, conda build fixed

ayshih:
Hmm.  I guess this isn't a problem on recent versions of `pytest` but for whatever reason, `conda` is solving the pin `pytest<7.2` with `pytest` 6.2.5.  That's why only the `conda` run was erroring.

sunpy/map/sources/tests/test_mdi_source.py
```diff
@@ -64,7 +64,6 @@ def test_unit(mdi):
     assert mdi.unit == u.dimensionless_unscaled
 
 
-@pytest.mark.filterwarnings("error")
```
nabobalis, line 67: ?!?

nabobalis (approved):

Issue #6550, opened 2022-11-01
`get_horizons_coord` failing


dstansby:
### Describe the bug

`get_horizons_coord` is failing on the CI. The example build has this traceback:
```python
 /home/runner/work/sunpy/sunpy/examples/units_and_coordinates/getting_lasco_observer_location.py failed leaving traceback:
Traceback (most recent call last):
  File "/home/runner/work/sunpy/sunpy/examples/units_and_coordinates/getting_lasco_observer_location.py", line 69, in <module>
    soho = get_horizons_coord('SOHO', lasco_map.date)
  File "/home/runner/work/sunpy/sunpy/.tox/build_docs-gallery/lib/python3.8/site-packages/sunpy/coordinates/ephemeris.py", line 317, in get_horizons_coord
    raise e
  File "/home/runner/work/sunpy/sunpy/.tox/build_docs-gallery/lib/python3.8/site-packages/sunpy/coordinates/ephemeris.py", line 313, in get_horizons_coord
    result = query.vectors()
  File "/home/runner/work/sunpy/sunpy/.tox/build_docs-gallery/lib/python3.8/site-packages/astroquery/utils/class_or_instance.py", line 25, in f
    return self.fn(obj, *args, **kwds)
  File "/home/runner/work/sunpy/sunpy/.tox/build_docs-gallery/lib/python3.8/site-packages/astroquery/utils/process_asyncs.py", line 29, in newmethod
    result = self._parse_result(response, verbose=verbose)
  File "/home/runner/work/sunpy/sunpy/.tox/build_docs-gallery/lib/python3.8/site-packages/astroquery/jplhorizons/core.py", line 1295, in _parse_result
    data = self._parse_horizons(response.text)
  File "/home/runner/work/sunpy/sunpy/.tox/build_docs-gallery/lib/python3.8/site-packages/astroquery/jplhorizons/core.py", line 1152, in _parse_horizons
    raise ValueError(('Ambiguous target name; provide '
ValueError: Ambiguous target name; provide unique id:
  ID#      Name                               Designation  IAU/aliases/other   
  -------  ---------------------------------- -----------  ------------------- 
       10  Sun                                             Sol                  
      -23  Pioneer 10 (spacecraft)                                              
  -399100  Apollo 10 S-IVB (spacecraft)                                         
  -399101  Apollo 10 LM (spacecraft)                       Snoopy               
 


-------------------------------------------------------------------------------
```

### To Reproduce

import sunpy

[...]

PR #6551, opened 2022-11-01
Fix body spec in get_horizons_coord

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/6550

### DIFF
--- a/changelog/6551.bugfix.rst
+++ b/changelog/6551.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed interal logic in :func:`sunpy.coordinates.get_horizons_coord` to adapt to a recent change in how the remote HORIZONS service handles body specifications.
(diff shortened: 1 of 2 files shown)


ayshih (changes requested):
Blocking until I interact with the JPL HORIZONS folks

dstansby:
I definitely think it's a good idea to chat to JPL Horizons, but isn't making this clearer ("Sun" instead of "10") still a change worth making?

dstansby:
Ah, just seen your comment on matrix:

>  ID# is intended to be the robust way to specify the body.

Is the issue with "Sun" that there's a possiblity that the meaning of "Sun" might change in the future, whereas the meaning of 10 won't change because it's a unique immutable ID?

ayshih:
Not literally a change of meaning of "Sun", but rather the potential addition of other bodies to the database that results in the string "Sun" being an ambiguous identifier.

If it takes a while to get a satisfactory response from the HORIZONS folks, we'll make this change temporarily to restore functionality, but the final implementation should be the formal way to unambiguously specify the Sun, whatever that is.

ayshih:
Closing this PR because the Horizons folks have fixed the bug on their side

PR #6557, opened 2022-11-02
Use `XPOSURE` key in `exposure_time` property

### LINKED ISSUE #6513: Use `'XPOSURE'` instead of `'EXPTIME'` key when creating `exposure_time` property
Currently, we pull the exposure time information from the `EXPTIME` key when creating the `exposure_time` property,

https://github.com/sunpy/sunpy/blob/f33efda2b758d3e9791ae4092bd639d260434223/sunpy/map/mapbase.py#L909-L917

However, the FITS-4 standard uses `XPOSURE` to denote this value. We should be using this by default to access the exposure time. We currently have two map sources (WISPR and EUI) that special case this because those FITS headers use `XPOSURE`

A few comments:

- Do we change the default key to `XPOSURE` and special case all of our other sources that use `EXPTIME` (most of them)?
- Alternatively, do we, in `GenericMap` look first for `XPOSURE` and then for `EXPTIME`? This is easiest but is inconsistent with our model of having a single default for such properties and then special casing individual sources. 
- We should also modify the header helper to set the exposure time using `XPOSURE`.

wtbarnes:
Fixes #6513 

When creating the `exposure_time` property on a map, this look first for the FITS 4 compliant keyword `XPOSURE` and then for the FITS 3 keyword `EXP_TIME`. After some discussion at the most recent community meeting, it was decided that we should continue to support both keys by default (i.e. not make them source-specific as both are part of the FITS standard.

### DIFF
--- a/changelog/6557.feature.rst
+++ b/changelog/6557.feature.rst
@@ -1,2 +1,2 @@
 `~sunpy.map.GenericMap.exposure_time` now looks for the exposure time in the ``XPOSURE`` key first
-and then the ``EXP_TIME`` key.
+and then the ``EXPTIME`` key.
--- a/changelog/6557.feature.rst
+++ b/changelog/6557.feature.rst
@@ -1,0 +1,2 @@
+`~sunpy.map.GenericMap.exposure_time` now looks for the exposure time in the ``XPOSURE`` key first
+and then the ``EXP_TIME`` key.
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -914,7 +914,7 @@ class GenericMap(NDData):
         This is taken from the 'XPOSURE' keyword or the 'EXPTIME' FITS keyword,
         in that order.
         """
-        exptime = self.meta.get('xposure', self.meta.get('exptime', None))
+        exptime = self.meta.get('xposure') or self.meta.get('exptime')
         if exptime:
             return exptime * self.timeunit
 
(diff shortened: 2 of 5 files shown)


Cadair (changes requested):
Oh wait, a test.

sunpy/map/mapbase.py
```diff
@@ -911,10 +911,12 @@ def exposure_time(self):
         """
         Exposure time of the image.
 
-        This is taken from the 'EXPTIME' FITS keyword.
+        This is taken from the 'XPOSURE' keyword or the 'EXPTIME' FITS keyword,
+        in that order.
         """
-        if 'exptime' in self.meta:
-            return self.meta['exptime'] * self.timeunit
+        exptime = self.meta.get('xposure', self.meta.get('exptime', None))
```
nabobalis, line 917:
```suggestion
        exptime = self.meta.get('xposure') or self.meta.get('exptime')
```
?

wtbarnes: Won't `get`raise a `KeyError` if the key is missing if there isn't a default?

nabobalis: I thought the default was None

wtbarnes: You're right. I've no idea why I thought it raised an error.

nabobalis (approved):

changelog/6557.feature.rst
```diff
@@ -0,0 +1,2 @@
+`~sunpy.map.GenericMap.exposure_time` now looks for the exposure time in the ``XPOSURE`` key first
+and then the ``EXP_TIME`` key.
```
Cadair, line 2:
```suggestion
and then the ``EXPTIME`` key.
```

Cadair (approved):

PR #6558, opened 2022-11-02
Optionally set detector in header helper

wtbarnes:
Optionally set the `DETECTOR` key using the header helper. In `GenericMap`, we derive the `.detector` property from this key so it make sense to be able to set it using the header helper.

### DIFF
--- a/changelog/6558.feature.rst
+++ b/changelog/6558.feature.rst
@@ -1,0 +1,2 @@
+`~sunpy.map.make_fitswcs_header` now includes the keyword argument ``detector`` for setting the
+``DETECTOR`` FITS keyword in the resulting header.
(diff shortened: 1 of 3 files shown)


nabobalis (approved):

nabobalis:
macos failure looks like a hypothesis issue?

wtbarnes:
> macos failure looks like a hypothesis issue?

I kicked it and it passed

Cadair:
> we derive the .detector property from this key so it make sense to be able to set it using the header helper.

This is a slippery slope no? Don't we have a bunch of other properties we don't support in the header helper?

Cadair (approved):
I am tentatively ok with this, I would like us to have a concrete scope of the things we do and don't support in this API. If it's just "things we have properties for" do we have them all at this point?

nabobalis:
> I am tentatively ok with this, I would like us to have a concrete scope of the things we do and don't support in this API. If it's just "things we have properties for" do we have them all at this point?

Want me to open an issue?

Issue #6563, opened 2022-11-03
Docs do not build properly on case-insensitive filesystems


ayshih:
### Provide a general description of the issue or problem.

When Sphinx is processing a target, say `alice.bob.carol`, it will check whether `alice/bob.py` exists in the filesystem, and if it does, it treats `alice.bob` as a module and `carol` as a something defined in that module.  Makes sense. 
 Unfortunately, this filesystem-check can have false positives on case-insensitive filesystems (primarily Windows) when a class is defined in a module with the same name, and that class is imported a level up.

Here's a illustrative example from our codebase.  The class `CompositeMap` is defined in the file `sunpy.map.compositemap.py`, and then we import `sunpy.map.compositemap.CompositeMap` to one level higher in the namespace: `sunpy.map.CompositeMap`.  We then want to document methods such as `sunpy.map.CompositeMap.plot()`, which is allowed by `sphinx-automodapi`.  The problem is that Sphinx then checks for a file `sunpy/map/CompositeMap.py`, which on a case-insensitive filesystem is the same as `sunpy/map/compositemap.py`, and thus exists.  So, Sphinx tries and fails to find a `plot()` defined (not within a class) in `sunpy/map/compositemap.py`.

I believe that these are all of the classes in our codebase that have this problematic pattern:
```
sunpy.data.data_manager.cache.Cache
sunpy.database.database.Database
sunpy.map.compositemap.CompositeMap
sunpy.map.mapsequence.MapSequence
sunpy.net.scraper.Scraper
sunpy.net.helio.chaincode.Chaincode
sunpy.time.timerange.TimeRange
sunpy.visualization.animator.mapsequenceanimator.MapSequenceAnimator
```

The expedient solution to this problem is to simply rename each of these modules so that its name does not case-insensitive match the class that it defines (e.g., prepend an underscore to the module name).  Each of these classes is imported one level up and documented only there, so the name of the module doesn't really matter.

I am exploring code-based solutions.

ayshih:
> The expedient solution to this problem is to simply rename each of these modules so that its name does not case-insensitive match the class that it defines (e.g., prepend an underscore to the module name). Each of these classes is imported one level up and documented only there, so the name of the module doesn't really matter.

[...]

PR #6564, opened 2022-11-03
Clean up our namespace by making private some modules that are not intended to be imported directly

ayshih:
It is common in our codebase to define classes/functions in a module, but import them one level up and never intend that module to be directly accessed.  Think how `sunpy.map.Map()` is actually defined in `sunpy.map.map_factory.py`, but we never tell people to call `sunpy.map.map_factory.Map()`.  I argue that it'd be cleaner to make the module private – i.e., rename `map_factory.py` to `_map_factory.py` – so that the user won't see `sunpy.map.map_factory` when perusing the public API.  We don't document `sunpy.map.map_factory.Map()` at that location, so making the module private has no effect on documentation.

As a first step, this PR makes private a subset of such modules (and not even `map_factory.py` at this point).  Since these classes are all documented at the higher level, our formal API is not changed.  Under the original module names, there are stub files to signal deprecation for anyone who might be importing them from the lower level, even though they did not need to.

There is one additional change because `sunpy/net/scraper.py` contains more than just the `Scraper` class: there is an undocumented function `sunpy.net.scraper.get_timerange_from_exdict()`.  Since this function is highly tied to `Scraper`, I have moved it to be a static method of `Scraper`.  Thus, it is now available as `sunpy.net.Scraper.get_timerange_from_exdict()` (with deprecation at the old place).

The reason for the selection of these eight modules is motivated by a separate consideration (see below), although it will not fix that problem until the deprecations introduced in this PR are removed (with version 5.1).

Note that this PR is far easier to review by looking at individual commits because then the renaming actions are clearer.

-----
There are eight classes in our codebase that are defined in a module file with the same name in lowercase.  These classes are all imported one level up and documented there.  The consequence is that the docs build on case-insensitive filesystems fails (see #6563 for more detail).


dstansby:
I'm 👎 on trying to sneak this into 4.1 since:

- Deprecations have to be in one LTS before being enforced, so putting this in 4.1 vs. 5.0 won't change when the deprecations expire.
- It's not an urgent user facing bug being fixed
- It's ~2 weeks from feature freeze now (and I'd count deprecating something as a feature-level change)
- I think we (well, someone!) should explore upstream fixes first (see https://github.com/sunpy/sunpy/issues/6563#issuecomment-1302481038) before we change API to work around an issue with how documentation is built

ayshih:
Note: our documented API is not changed at all by this PR.

* Deprecations have to be in one LTS before being enforced, so putting this in 4.1 vs. 5.0 won't change when the deprecations expire.

Sure, but it's not solely about starting the clock.  Because I'm closing off undocumented places for importing things, I want users who use only releases to have as much notice as possible to update their code to the documented places for importing things.

* It's not an urgent user facing bug being fixed

Not urgent, but there are user-facing aspects.  I think that it reduces the potential for user confusion by having only one public place to import things.  Why permit `sunpy.map.compositemap.CompositeMap` instead of funneling everyone to `sunpy.map.CompositeMap`?

> * I think we (well, someone!) should explore upstream fixes first (see [Docs do not build properly on case-insensitive filesystems #6563 (comment)](https://github.com/sunpy/sunpy/issues/6563#issuecomment-1302481038)) before we change API to work around an issue with how documentation is built

There are years-old issues on the Sphinx repo about how it interacts poorly with the case-insensitive filesystems, with zero prospects for them being addressed.  The exact cases are bit different, because here it is related to our use of `sphinx-automodapi`.

ayshih:
I'll add that even if there turns out to be a tractable code solution in `sphinx` or `sphinx-automodapi` – and I don't think one exists – I'd argue that we should still turn these modules private.  These classes are being documented only at the higher level, so there's no reason to give users the impression that importing a module such as `sunpy.map.compositemap` is something they should do.

nabobalis:
> Not urgent, but there are user-facing aspects. I think that it reduces the potential for user confusion by having only one public place to import things. Why permit `sunpy.map.compositemap.CompositeMap` instead of funneling everyone to `sunpy.map.CompositeMap`?

I think this is a good question and one that has broader implications for our entire API namespace.

ayshih:
Of course, there are times that it can be useful to have multiple places to import things.  For example, our coordinate frames are both in `sunpy.coordinates.frames` and `sunpy.coordinates`, and the logical grouping offered by `sunpy.coordinate.frames` can be useful in certain situations.  However, no such justification exists for single-class modules.

dstansby:
I'm definitely 👍 to these changes in principle.

Sorry, I decided to go ahead and branch for 4.1 since we're after feature freeze and slightly behind schedule (full release was meant to be tomorrow, and we still need to generate a rc and test sponsored packages) and I found a spare moment to do the branching.

Cadair:
I am also :-1: on sneaking this into the release, I think it's too late for that, this needs some consideration.

ayshih:
Welp, kicking this PR to draft.  It turns out – obviously, in retrospect – that adding the stub modules for deprecation causes the docs build to fail because there's still a, e.g., `sunpy.map.compositemap`, even if it's just a deprecation helper.  Unless I get any epiphanies, this PR will probably be recast as an API-simplifying PR, because the fixing-the-docs-build-on-Windows goal would be achieved only upon removal (i.e., 5.1).

Cadair:
I haven't had a chance to really look at this in detail yet, but I have to say I am not a fan of the filesystem dog wagging the api tail.

If these changes can be justified as pure cleanup then I am probably on board, but I am not a fan of trying to work around limitations of poor filesystems and sphinx by refactoring our code with no other justification.

ayshih:
Again, the documented API is not changing.  We don't document, and have no use, for the problematic exposed namespace.  Simplifying the exposed namespace is a Good Thing, and the fact that it would allow the docs build to succeed just gives added reason why we should do it.

Cadair (commented):
I am :+1:  on this, but it *definitely* needs developer documentation laying out when new files should be private and when they shouldn't.

ayshih:
Ha!  I had missed that stale bot was gunning this off.  I'll dust this off and re-open if there are no showstoppers.

PR #6571, opened 2022-11-06
Fix the handling of `CDELTi` and `PCi_j` in `resample()` and `superpixel()`

### LINKED ISSUE #6570: `.resample()` and `.superpixel()` modify `CDELTi`/`PCi_j` in a way that results in non-intuitive behavior
The FITS WCS standard allows for "considerable flexibility" in the way that `CDELTi` values and the `PCi_j` matrix are defined.  You can choose any combination that multiplies to the `CDi_j` matrix that you need.  However, for 2D imagery, most users would probably expect that the `PCi_j` matrix be as close to a pure rotation matrix as possible, and the `CDELTi` values to be represent the pixel plate scales.

For whatever reason, when modifying a map that uses the `PCi_j` matrix, both `.resample()` and `.superpixel()` modify only the `PCi_j` matrix and do not modify the `CDELTi` values.  Again, this is technically permitted.  However, this has several non-intuitive consequences:

- `.scale` returns `CDELTi` values, which are no longer the pixel plate scales
- `.rotation_matrix` returns the `PCi_j` matrix, which is no longer close to a pure rotation matrix and contains scaling factors
- A de-rotating `.rotate()` call on the map leads to an unexpected resampling because it uses `.rotation_matrix`, which now has scaling factors in it

I'll work up a PR to fix this.

ayshih:
For a map that uses the PCij matrix, both `GenericMap.resample()` and `GenericMap.superpixel()` put the scaling factors in the `PCi_j` matrix instead of the `CDELTi` values (see #5786), which results in non-intuitive behaviors (see #6570).  This PR puts the scaling factors in the `CDELTi` values instead, the same as if the map uses the `CROTA2` formalism.  The off-diagonal elements of the `PCi_j` matrix need to be modified if the pixels are not square, but the diagonal elements of the `PCi_j` matrix stay unchanged.

Figure hashes need to be updated for the figure tests that test de-rotating a resampled map because the de-rotation no longer undoes the rectangular pixels.

Closes #6570


sunpy/map/mapbase.py
```diff
@@ -1542,20 +1542,13 @@ def resample(self, dimensions: u.pixel, method='linear'):
...
-            new_meta['cd1_1'] *= scale_factor_x
-            new_meta['cd2_1'] *= scale_factor_x
-            new_meta['cd1_2'] *= scale_factor_y
-            new_meta['cd2_2'] *= scale_factor_y
-        if 'cd1_1' not in self.meta and 'pc1_1' not in self.meta:
-            # Using the CROTA2 and CDELT formalism
-            new_meta['cdelt1'] *= scale_factor_x
-            new_meta['cdelt2'] *= scale_factor_y
+            new_meta['pc2_1'] *= scale_factor_x / scale_factor_y
+            new_meta['pc1_2'] *= scale_factor_y / scale_factor_x
```
dstansby, line 1551: To make it clearer that the CDELTi keywords are used in conjunction with the PCi_j keywords, and not the CDi_j keywords when interpreting the metadata, perhpas it would be better to move the code for scaling CDELTi into this `if` block?

sunpy/map/tests/test_mapbase.py
```diff
@@ -686,8 +686,15 @@ def test_resample_metadata(generic_map, sample_method, new_dimensions):
...
-    assert resampled_map.meta['cdelt1'] == generic_map.meta['cdelt1']
-    assert resampled_map.meta['cdelt2'] == generic_map.meta['cdelt2']
+
+    scale_x = generic_map.data.shape[1] / resampled_map.data.shape[1]
+    scale_y = generic_map.data.shape[0] / resampled_map.data.shape[0]
+
+    assert resampled_map.meta['cdelt1'] == scale_x * generic_map.meta['cdelt1']
+    assert resampled_map.meta['cdelt2'] == scale_y * generic_map.meta['cdelt2']
+    assert resampled_map.meta['pc1_2'] == scale_y / scale_x * generic_map.meta['pc1_2']
+    assert resampled_map.meta['pc2_1'] == scale_x / scale_y * generic_map.meta['pc2_1']
```
dstansby, line 696:
```suggestion
    assert resampled_map.meta['pc2_1'] == scale_x / scale_y * generic_map.meta['pc2_1']
    assert resampled_map.meta['pc1_1'] == generic_map.meta['pc1_1']
    assert resampled_map.meta['pc2_2'] == generic_map.meta['pc2_2']
```
Worth adding a check that the pc1_1 and pc2_2 metadata hasn't been changed? Even though it's implicitly checked below, I think for clarity it's worth having an explicit check up here.

dstansby (approved):
Looks 👍  - two suggests to clarify the tests, but feel free to take or leave them

sunpy/map/mapbase.py
ayshih: As far as I can tell, there is no prohibition to have CDELTi when using CDi_j, even though the CDELTi will not be used for the transformation.  The header might be keeping them around to facilitate conversion back to a specific PCi_j formulation.  My thinking is that there is no reason to skip scaling CDELTi.

sunpy/map/tests/test_mapbase.py
ayshih: Sure, done

dstansby (approved):

sunpy/map/tests/test_mapbase.py
```diff
@@ -686,8 +686,17 @@ def test_resample_metadata(generic_map, sample_method, new_dimensions):
     Check that the resampled map has correctly adjusted metadata.
     """
     resampled_map = generic_map.resample(new_dimensions, method=sample_method)
-    assert resampled_map.meta['cdelt1'] == generic_map.meta['cdelt1']
-    assert resampled_map.meta['cdelt2'] == generic_map.meta['cdelt2']
+
+    scale_x = generic_map.data.shape[1] / resampled_map.data.shape[1]
```
Cadair, line 690: We seem to be missing an equivalent test for this for superpixel? In particular I can't see a test which checks superpixel's metadata when using a CD matrix?

Cadair (dismissed):
Looks good to me, I would just like clarity on the superpixel test.

Also I had a quick look at `rotate` to check there isn't a similar issue there but I think it's fine?

sunpy/map/mapbase.py
```diff
@@ -1542,20 +1542,14 @@ def resample(self, dimensions: u.pixel, method='linear'):
         new_meta = self.meta.copy()
 
         # Update metadata
+        for key in {'cdelt1', 'cd1_1', 'cd2_1'} & self.meta.keys():
+            new_meta[key] *= scale_factor_x
+        for key in {'cdelt2', 'cd1_2', 'cd2_2'} & self.meta.keys():
```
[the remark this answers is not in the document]
ayshih: Since you acknowledge that it is a "very personal" preference, I'll softly push back.  I feel that the set operators `&` (instead of `.intersection()`), '|' (instead of `.union()`), and `-` (instead of `.difference()`) are sufficiently intuitive to be used without reservation, and certainly much preferred for more complex set expressions (e.g., `(a | b) - (c & d)`).  In contrast, I'm not going to go to bat for `^`, `<`, etc.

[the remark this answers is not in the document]
Cadair: Certainly for complex operations the brevity wins out. I just think for here if someone who isn't as experience with sets reads this if we use the method then the code "reads" with minimal extra understanding needed. "for key in the intersection of this set with meta keys" where as you can only read it like that with the operator if you know what the operator does.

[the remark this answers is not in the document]
Cadair: (I am not fussed if we change this or not, I just thought it was a mild readability improvement)

sunpy/map/tests/test_mapbase.py
ayshih: It's a good thing you asked for this, because it uncovered another bug: the `NAXIS1`/`NAXIS2` keywords weren't getting updated by `superpixel()`

dstansby (approved):

PR #6572, opened 2022-11-06
Converts some parts of documentation to OSPL style

dpshelio:
Most of the documentation follows already this style (as suggested in #2849).
This completes (I think!) the effort in a semi-automated way (with emacs power).

What it's required now is a mechanism to check that this is kept as such.

Changes on the Changelog files have not been included as these files should not change in the future.


Cadair:
Could we do something like a pre-commit which every line over like 50 char long has to end in a full stop?

nabobalis:
I think we have so many conflicts, that I will close this for now.

Ideally if we can get a pre-commit for this but I imagine parsing text like this that can be quite varied will be a nightmare.

PR #6573, opened 2022-11-07
Fixed the determination of scale and rotation matrix from the CDij matrix

ayshih:
- Fixes a bug with the `.rotation_matrix` property where the PCij matrix was not determined correctly from the CDij matrix in the case of non-square pixels.
- Fixes the `.scale` property to return reasonable values when a map has a CDij matrix and no CDELTi values.  The returned pixel scales are the values that are needed so that each row of the inferred PCij matrix has unity norm.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1273,10 +1273,22 @@ class GenericMap(NDData):
         """
         Image scale along the x and y axes in units/pixel
         (i.e. cdelt1, cdelt2).
+
+        If the CDij matrix is defined but no CDELTi values are explicitly defined,
+        effective CDELTi values are constructed from the CDij matrix.  The effective
+        CDELTi values are chosen so that each row of the PCij matrix has unity norm.
+        This choice is optimal if the PCij matrix is a pure rotation matrix, but may not
+        be as optimal if the PCij matrix includes any skew.
         """
-        # TODO: Fix this if only CDi_j matrix is provided
-        return SpatialPair(self.meta.get('cdelt1', 1.) * self.spatial_units[0] / u.pixel,
-                           self.meta.get('cdelt2', 1.) * self.spatial_units[1] / u.pixel)
+        if 'cd1_1' in self.meta and 'cdelt1' not in self.meta and 'cdelt2' not in self.meta:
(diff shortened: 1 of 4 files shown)


sunpy/map/mapbase.py
```diff
@@ -1271,10 +1271,19 @@ def scale(self):
         """
         Image scale along the x and y axes in units/pixel
         (i.e. cdelt1, cdelt2).
+
+        If the CDij matrix is defined but no CDELTi values are explicitly defined,
+        effective CDELTi values are constructed from the CDij matrix.
```
Cadair, line 1276: I think it would be good to explain *how* here, because as you pointed out in #6570 there's not one singular "correct" way to do this. (Which is why it didn't used to do it)

Cadair (approved):
Looks good, it's nice that we can now have feature parity with the CD matrix. My ability to fully comprehend the FITS spec was clearly not up to scratch when I first implemented this.

sunpy/map/mapbase.py
nabobalis: Do you want this in this PR or another one?

Cadair: This one preferably.

nabobalis (changes requested):
Blocking until Stuart is happy

sunpy/map/mapbase.py
dstansby: +1 to documenting this clearly in this PR.

dstansby:
>  The returned pixel scales are the values that are needed so that each row of the inferred PCij matrix has unity norm.

What's the reasoning for this choice? Does it make the resulting scale have some nice mathematica; properties?

sunpy/map/mapbase.py
ayshih: Elaborated

ayshih:
The reason to make this choice (which is the optional constraint suggested in Paper I, equation 4) is so that if the rows of the CDij matrix are orthogonal, then the rows of the PCij matrix are orthonormal.  That's your best shot at getting to a matrix that is a pure rotation (or reflection), if that is possible.

nabobalis (approved):

nabobalis:
Doc fail is due to another PR that I merged that broke it.

PR #6581, opened 2022-11-10
Fixes issues with `parse_time()` and the handling of excess zeros in the microsecond field

### LINKED ISSUE #6586: Issue with parse_time and formats
### Describe the bug

I have come across this issue when trying to parse a time string using parse time, when it's in this format  "%d-%b-%Y %H:%M:%S.%f",  # Example 04-May-2007 21:08:12.999999 

In this case it works:
```python 
parse_time('30-Mar-2022 17:25:14.500')
<Time object: scale='utc' format='isot' value=2022-03-30T17:25:14.500>
```
but doesn't in this:

```python
parse_time('30-Mar-2022 17:25:14.000')
ValueError: Input values did not match any of the formats where the format keyword is optional: {'datetime': TypeError('Input values for datetime class must be datetime objects'), 'ymdhms': ValueError('input must be dict or table-like'), 'iso': ValueError('Time 30-Mar-2022 17:25:14 does not match iso format'), 'isot': ValueError('Time 30-Mar-2022 17:25:14 does not match isot format'), 'yday': ValueError('Time 30-Mar-2022 17:25:14 does not match yday format'), 'datetime64': TypeError('Input values for datetime64 class must be datetime64 objects'), 'fits': ValueError('Time 30-Mar-2022 17:25:14 does not match fits format'), 'byear_str': ValueError('Time 30-Mar-2022 17:25:14 does not match byear_str format'), 'jyear_str': ValueError('Time 30-Mar-2022 17:25:14 does not match jyear_str format'), 'astropy_time': TypeError('Input values for astropy_time class must all be same astropy Time type.')}
```
Seems to be related to the zero after the decimal point.

### System Details

sunpy 4.0.6  
astropy 5.0

ayshih:
- Fixed overly blunt handling of excess trailing zeros beyond the six digits allowed for the microsecond field (see #445).  Previously, any string input would be stripped of the final period and trailing zeros, even if it was not a microsecond field.  Now, the handling is properly targeted.
- Two time formats were missing for parity with other time formats with respect to with/without microseconds.

Closes #6586

### DIFF
--- a/changelog/6581.bugfix.rst
+++ b/changelog/6581.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug where :func:`~sunpy.time.parse_time` would always disregard the remainder of a time string starting with the final period if it was followed by only zeros, which could affect the parsing of the time string.
--- a/sunpy/time/tests/test_time.py
+++ b/sunpy/time/tests/test_time.py
@@ -27,12 +27,19 @@ def test_parse_time_24_2():
     assert dt.scale == 'utc'
 
 
-def test_parse_time_trailing_zeros():
-    # see issue #289 at https://github.com/sunpy/sunpy/issues/289
-    dt = parse_time('2010-10-10T00:00:00.00000000')
-    assert dt == Time('2010-10-10')
+def test_parse_time_microseconds_excess_trailing_zeros():
+    dt = parse_time('2010-Oct-10 00:00:00.1234560')
+    assert dt == Time('2010-10-10 00:00:00.123456')
     assert dt.format == 'isot'
     assert dt.scale == 'utc'
+
+    # Excess digits beyond 6 digits should error if they are not zeros
+    with pytest.raises(ValueError):
+        dt = parse_time('2010-Oct-10 00:00:00.1234567')
+
+    # An ending run of zeros should still error if they are not a microsecond field
+    with pytest.raises(ValueError):
+        dt = parse_time('10-Oct-2010.0000000')
 
 
 def test_parse_time_tuple():
--- a/sunpy/time/time.py
+++ b/sunpy/time/time.py
@@ -102,18 +102,28 @@ def _regex_parse_time(inp, format):
     match = re.match(format, inp)
     if match is None:
         return None, None
-    try:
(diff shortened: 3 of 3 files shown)


ayshih:
The issue in the wild is that `parse_time("10-Oct-2010 00:00:00.0")` errors, while `parse_time("10-Oct-2010 00:00:00.1")` works fine.  That's because for the first string, the final `.0` gets stripped, and there happens to be a silly omission in the defined time string formats.  The issue in the wild can be fixed by simply adding the missing time string format, but we really shouldn't be always stripping out part of the string without considering whether the stripped digits are actually part of a microsecond field.

changelog/6581.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed a bug where :func:`~sunpy.time.parse_time` would always disregard the remainder of a time string starting with the ifnal period if it was followed by only zeros, which could affect the parsing of the time string.
```
nabobalis, line 1:
```suggestion
Fixed a bug where :func:`~sunpy.time.parse_time` would always disregard the remainder of a time string starting with the finial period if it was followed by only zeros, which could affect the parsing of the time string.
```

sunpy/time/tests/test_time.py
```diff
@@ -27,13 +27,20 @@ def test_parse_time_24_2():
     assert dt.scale == 'utc'
 
 
-def test_parse_time_trailing_zeros():
-    # see issue #289 at https://github.com/sunpy/sunpy/issues/289
-    dt = parse_time('2010-10-10T00:00:00.00000000')
-    assert dt == Time('2010-10-10')
```
nabobalis, line 33: Does this still work?

ayshih: Yes, but I replaced it with a "harder" test

changelog/6581.bugfix.rst
ayshih: Heh, you misspelled your fix of my misspelling =)

nabobalis (approved):

alasdairwilson (approved):

Issue #6519, opened 2022-10-28
Parallelise code to update cdaweb datasets


dstansby:
### Describe the feature

I'm currently running the code that scrapes CDAWeb for currently supported datasets. The code can be found here: https://github.com/sunpy/sunpy/blob/8c0b9054dc25918ee5b6286a93edbbca09cde903/sunpy/net/cdaweb/helpers.py#L118

The code only fires off one request at a time, waiting for it to come back before putting in the next request. This means it takes a long time (>~5 mins) to scrape all the datasets. I think it *should* be possible to send multiple requests at once to speed this up.

lzivadinovic:
I would like to take a look at this.

exitflynn:
I should be able to do this but it'd involve shifting from using the `requests` module to `aiohttp` for making the requests. Would that be okay? 
[`aiohttp`'s exactly for this purpose](https://docs.aiohttp.org/en/stable/http_request_lifecycle.html#aiohttp-request-lifecycle) and is also how requests are handled in parfive (which is where i first discovered it) so it's already a dependency.

nabobalis:
There was an attempt here: https://github.com/sunpy/sunpy/pull/6601

Unclear what the outcome was as I've not been tracking it.

I think we would need to see an implementation that looked like with aiohttp.

exitflynn:
I remember going through that attempt and it looks like the approach they were going with was :-1:'ed because progress (the name of the databases being upgraded) could no longer be printed as it happened. They had another approach in mind which involved using threads and gnu parallel (which is a Linux-only thing?) but there hasn't been any implementation for that in a while. 

However I believe doing things asynchronously using aiohttp would be a *much* more elegant and suitable approach. Here's a PR I came up with to show how the implementation would look like.

exitflynn:
i wonder if we can use Parfive here to download the data?

dstansby:
Fixed by https://github.com/sunpy/sunpy/pull/6601

PR #6601, opened 2022-11-15
Improve performance when fetching data from cdaweb

lzivadinovic:
Reported in issue #6519 fetching data from cdaweb is slow because we are sending one request per instrument. Whit this patch, code will build query containing all instruments and send request.

There is support for that kind of queries in cdas docs where it says that if you add multiple query parameters with same key, it will OR them https://cdaweb.gsfc.nasa.gov/WebServices/REST/#Get_Datasets

Also, i've ditched the idea of running multiple requests because there is a note here https://cdaweb.gsfc.nasa.gov/WebServices/REST/py/cdasws/ saying that calls can be throttled if we have more than 5 calls simultaneously.

Simple test with `time` and running command with my connections gives around 3 minutes faster code execution

```
# Before patch, one request per instrument
python tools/update_attrs_json.py  8.37s user 0.86s system 1% cpu 8:56.05 total

# After patch, only one request
python tools/update_attrs_json.py  2.08s user 0.75s system 0% cpu 6:09.15 total
```


lzivadinovic:
> Thanks a lot for this! Could you remove the changelog entry? (this isn't a user-facing tool) Otherwise I will give this a test later this week, but looks +1

Sure, no problem.

dstansby:
I've just run this, and it might have sped it up a bit, but not significantly. There's also a lack of progress printed now - in the previous version each observatory was printed to the command line when it was being updated, which provided useful feedback that something was happening. Unless we can get this script down to running in <~ 30secs, I think having the feedback is important to reassure that something is happening.

So I think I'm actually 👎 on this (sorry!), unless there's a way to get continual feedback printed during the request being made.

lzivadinovic:
I dont think there is a way to get progress, this is one api request and we wait while backend process it. We dont get any info, simple get and that's it.

We can try threads but we should be limited to 3 or 4 at given time. 

With gnu parallel and 4 connections i can get
```
time parallel -j4 'curl -H "Accept: application/json" {}' < plist >> res  
# plist is list of endpoints with different observatories
12.12s user 3.74s system 7% cpu 3:40.27 total
```
And with single request: 
```
curl -H "Accept: application/json" https....... -o /dev/null  0.33s user 0.25s system 0% cpu 6:27.84 total
```
So yeah, looks like we can do it this way. I can try this in few days.

dstansby (changes requested):
As per https://github.com/sunpy/sunpy/pull/6601#issuecomment-1316834322 I'm going to mark as request changes. I running 5 requests at the same time but still individually querying each group is the way to go, so we can still have some user feedback as the requests are happening.

dstansby:
@lzivadinovic do you think it's best to close this PR, and you can always open a new PR with the other approach? Or are you planning on modifying this PR to use the other approach?

lzivadinovic:
Hi, sorry for very late response. I've added threadpool executor that spawns multiple threads and make multiple parallel request to rest API. It's 'user-friendly' in terms that it will write data its currently fetching, like original solution. Number of connection is limited to N=3 I've read this number somewhere in documentation, that they tolerate that much simultaneous requests from same ip, but this number can be easily changed.

dstansby (approved):
Sorry it took so long, but just tried this out and it's great, thanks a lot! 🚀

PR #6592, opened 2022-11-11
Remove 4.1 changelogs

dstansby:
# Thank you for contributing to SunPy

## 🚨 IMPORTANT 🚨

We know that working on code and submitting pull requests takes effort, and we appreciate your time.
Thank you.

[Please be aware that everyone has to follow our code of conduct](https://sunpy.org/coc)

[Furthermore, you might need to check with your work place if you are allowed to contribute code](https://docs.sunpy.org/en/latest/dev_guide/contents/newcomers.html)

- Please try to create an issue _before_ creating a Pull Request.
- Please use the following Git commit message style
  - Use the future tense ("Adds feature" not "Added feature")
  - Limit the first line to 72 characters or less

We have several automatic features that run when a pull request is open.
They can appear daunting but do not worry about them!
[This a brief explanation of them.](https://docs.sunpy.org/en/latest/dev_guide/contents/pr_review_procedure.html#continuous-integration)

PR #6603, opened 2022-11-16
Remove 4.1 changelogs

dstansby:
Don't know what I managed to do with https://github.com/sunpy/sunpy/pull/6592, but here's a replacement with README.md not deleted.

### DIFF
--- a/changelog/5951.feature.rst
+++ b/changelog/5951.feature.rst
@@ -1,1 +1,0 @@
-Updated and expanded the HTML representation for `~sunpy.timeseries.TimeSeries`.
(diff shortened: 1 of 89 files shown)


nabobalis:
Do we not need to update changelog.rst?

dstansby:
Oh yeah, so we do, thanks for catching! I shall do it in this PR.

Cadair:
The best way to do this in my experience has been to cherry-pick the render changelog commit from the release branch back onto main. Is that what you have done here?

That approach ensures two things: firstly that you get the rendered changelog and two that you only remove changelog fragments that were released in the major version in question.

dstansby:
> The best way to do this in my experience has been to cherry-pick the render changelog commit from the release branch back onto main. Is that what you have done here?

No, it was not, thanks for catching and suggesting the Right Way to do it - new PR at https://github.com/sunpy/sunpy/pull/6609 with the cherry-picked commit

Issue #6464, opened 2022-10-12
Add `PixelPair` to API docs


dstansby:
### Describe the feature

It looks like `PixelPair` isn't in the API docs, see [this search for PixelPair](https://docs.sunpy.org/en/stable/search.html?q=pixelpair&check_keywords=yes&area=default). It would be good to add it.

PR #6620, opened 2022-11-22
Add PixelPair to docs

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/6464

### DIFF
--- a/changelog/6620.doc.rst
+++ b/changelog/6620.doc.rst
@@ -1,0 +1,1 @@
+Added `sunpy.map.PixelPair` to the reference documentation.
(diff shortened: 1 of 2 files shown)


dstansby:
This worked: https://sunpy--6620.org.readthedocs.build/en/6620/generated/api/sunpy.map.PixelPair.html#sunpy.map.PixelPair

The rendered docs aren't perfect, but I think just documenting that the attributes are `x` and `y` is worthwhile here. I had a quick look at the `namedtuple` docs but there was nothing obvious about adding documentation to the attributes.

nabobalis (approved):

Issue #6615, opened 2022-11-18
Make the landing page more friendly


wtbarnes:
When you land on docs.sunpy.org, you're currently greeted with two boxes with bullet points. We should instead adapt what is currently in the brief tour into some sort of introduction into what sunpy can do. The user should be greeted with what sunpy can do, ideally in graphical format. This could be either in the form of actual plots or something more cartoon-y as in the glyph attached below.

![sunpy-summary-slide](https://user-images.githubusercontent.com/8676141/202766188-616336e5-6a28-4d21-9b81-58faccb517a4.svg)

nabobalis:
> Hi, I'd like to fix this issue. Should I make the graphics myself too if that's alright?

What graphics would you create?

> One query, I should be editing [`main`/docs/index.rst](https://github.com/sunpy/sunpy/blob/main/docs/index.rst?rgh-link-date=2023-02-22T08%3A32%3A37Z) right? Does the current [docs.sunpy.org](https://docs.sunpy.org/) reflect the reStructured text?

Yes it does.

Satan-Claws:
> What graphics would you create?

Some graphics like what @wtbarnes made but for each section

> Yes it does.

Are you sure? Because the current index.rst doesn't divide into 2 sections like mentioned but it does happen in [docs.sunpy.org](https://docs.sunpy.org/).

nabobalis:
> > What graphics would you create?
> 
> Some graphics like what @wtbarnes made but for each section

I think we would want to use the one that Will created instead of making a new one.

> > Yes it does.
> 
> Are you sure? Because the current index.rst doesn't divide into 2 sections like mentioned but it does happen in [docs.sunpy.org](https://docs.sunpy.org/).

I am sure, you need to go to https://docs.sunpy.org/en/latest/

Satan-Claws:
> I think we would want to use the one that Will created instead of making a new one.
 
Alright, I'll look into it.

> I am sure, you need to go to https://docs.sunpy.org/en/latest/

Oh sorry sorry, I was looking at the wrong webpage. I'll try to cut the images and put them in and will let you know

Satan-Claws:
@nabobalis I've edited index.rst and I've added the graphics provided by @wtbarnes. Should I create a pull request? Link to my fork https://github.com/Satan-Claws/sunpy/tree/main/docs

nabobalis:
[...]

PR #6616, opened 2022-11-18
Add target audience and how to read doc to the intro tutorial

dstansby:
This adds some narrrative documentation the describes the purpose, target audience, and "how to read" of the introductory tutorial. All up for bikeshedding and feedback, but I think this broadly explains what we discussed at the documentation sprint.

PR #6622, opened 2022-11-22
[doc-sprint] Add some narrative text to landing page

alasdairwilson:
A very quick attempt  (slowed by everything having being renamed when rebasing) at solving the same problem as #6615 but in a different way. This has the same information as the current landing page but presented in a more descriptive manner.  

I envision the flashy, attention grabbing content suggested in that issue as going in a new "what is sunpy" page linked right at the start here such that the first steps section goes "what is sunpy -> installation guide -> tutorial/walkthrough"

I think having laid it out like this it highlights the need for the tutorial to be improved and reorganised.

There is some placeholder code-comments envisioning where links to future envisioned sections would go. The whole thing Very much up for feedback.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -2,49 +2,46 @@
 sunpy core Documentation
 ************************
 
-``sunpy`` is a community-developed, free and open-source solar
-data analysis environment for Python.
+``sunpy`` is a community-developed, free and open-source solar data analysis environment for Python.
+
 
 New Users
-*********
+==========
 
-If you are a first time user or new to Python,
-
+For help installing ``sunpy`` refer to our :ref:`installation guide <installing>`.
 ..
         * :Introduction to sunpy core [new page described in #6615]
 
-* :doc:`installation guide <intro/installation>`
+If you are a first time user or new to Python, our **tutorial** provides a walkthrough on how to use the key features of sunpy core.
+Once you've installed sunpy, start here.
 
-Our **tutorial** provides a walkthrough on how to use the key features of sunpy core, we recommend new users start here.
+.. toctree::
+    :maxdepth: 2
 
-* :doc:`Getting Started <intro/index>`
+    intro/index
 
-  #. :doc:`Data Acquisition <intro/acquiring_data/index>`
-  #. :doc:`Data Types <intro/data_types/index>`
-  #. :doc:`Plotting <intro/plotting>`
-  #. :doc:`Units <intro/units>`
-  #. :doc:`Coordinates <intro/coordinates>`
-  #. :doc:`Time <intro/time>`
+
 
 
 Next Steps
-**********
+==========
 
 * :doc:`Example gallery <generated/gallery/index>` contains short-form guides on accomplishing common tasks using sunpy.
-* :doc:`API reference<reference/index>` contains a technical description of the functionality of sunpy core's modules. It describes how they work and how to use them but assume you have a good understanding of sunpy and the scientific python ecosystem in general.
+* :ref:`API reference<reference>` contains a technical description of the inputs, outputs, and behaviour of each component of sunpy core.
 
 ..
     * [if we have how-tos?] How to guides provide detailed overview on some of the features of sunpy. These assume some level of experience using sunpy.
 ..
     * [we dont have a link to these but we do have some explanation pages] Topic guides (name?) provide
 
-Further Info
-************
 
-* Find out :doc:`what's new <whatsnew/index>` with sunpy.
-* View a list of :doc:`known issues <reference/known_issues>` or report new issues on our `GitHub issue tracker`_.
+Further Info
+============
+
+* Find out :ref:`what's new <whatsnew>` with sunpy.
+* View a list of :ref:`known issues <known_issues>` or report new issues on our `GitHub issue tracker`_.
 * Join our `chat room`_ or post to our `discourse`_ for help.
-* Read the :doc:`Developers guide <dev_guide/index>` to start contributing to the sunpy project.
+* Read the :ref:`Developers guide <dev_guide>` to start contributing to the sunpy project.
 
 
 .. _chat room: https://openastronomy.element.io/#/room/#sunpy:openastronomy.org
@@ -55,7 +52,6 @@ Further Info
     :maxdepth: 1
     :hidden:
 
-    intro/index
     generated/gallery/index
     reference/index
     whatsnew/index
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -9,18 +9,25 @@ New Users
 ==========
 
 For help installing ``sunpy`` refer to our :ref:`installation guide <installing>`.
-..
-        * :Introduction to sunpy core [new page described in #6615]
+
+
+.. todo::
+
+    if #6615 is completed link here
+    Introduction to sunpy core
 
 If you are a first time user or new to Python, our **tutorial** provides a walkthrough on how to use the key features of sunpy core.
 Once you've installed sunpy, start here.
 
-.. toctree::
-    :maxdepth: 2
-
-    intro/index
-
-
+* :ref:`guide`
+    #. :ref:`installing`
+    #. :ref:`units-sunpy`
+    #. :ref:`time-in-sunpy`
+    #. :ref:`coordinates-sunpy`
+    #. :ref:`acquiring_data`
+    #. :ref:`map_guide`
+    #. :ref:`timeseries_guide`
+    #. :ref:`plotting`
 
 
 Next Steps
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -2,42 +2,64 @@
 sunpy core Documentation
 ************************
 
-The sunpy core package is a community-developed, free and open-source solar
+``sunpy`` is a community-developed, free and open-source solar
 data analysis environment for Python.
 
+New Users
```
dstansby, line 8: I think this would be better at one title level lower, to make sure "sunpy core Documentation" stays the biggest at the top.

docs/index.rst
```diff
@@ -2,42 +2,64 @@
...
+*********
+
+If you are a first time user or new to Python,
+
+..
+        * :Introduction to sunpy core [new page described in #6615]
+
+* :doc:`installation guide <intro/installation>`
```
dstansby, line 16:
```suggestion
* :doc:`Installation guide <intro/installation>`
```

docs/index.rst
```diff
@@ -2,42 +2,64 @@
...
+
+Our **tutorial** provides a walkthrough on how to use the key features of sunpy core, we recommend new users start here.
```
dstansby, line 18: But you've just said first time users go to the installation guide 😆 How about:
```suggestion
Our **tutorial** provides a walkthrough on how to use the key features of sunpy core.
Once you've installed sunpy, start here.
```

docs/index.rst
```diff
@@ -2,42 +2,64 @@
...
+
+* :doc:`Getting Started <intro/index>`
```
dstansby, line 20: I'm not a fan of this manual ordering and then a hidden toc tree. How about we move "Installation guide" to a new page that shows up in the side bar, then include that as a toctree above, and turn this getting started into a toctree?

docs/index.rst
```diff
@@ -2,42 +2,64 @@
...
+  #. :doc:`Units <intro/units>`
+  #. :doc:`Coordinates <intro/coordinates>`
+  #. :doc:`Time <intro/time>`
+
+
+Next Steps
+**********
+
+* :doc:`Example gallery <generated/gallery/index>` contains short-form guides on accomplishing common tasks using sunpy.
+* :doc:`API reference<reference/index>` contains a technical description of the functionality of sunpy core's modules. It describes how they work and how to use them but assume you have a good understanding of sunpy and the scientific python ecosystem in general.
```
dstansby, line 34:
```suggestion
* :doc:`API reference<reference/index>` contains a technical description of the inputs, outputs, and behaviour of each component of sunpy core.
```
I don't think API ref requires a good understanding of sunpy? With a bit of knoweldge (but not good understanding) I'd expect to be able to understand API reference entries in isolation.

docs/index.rst
```diff
@@ -2,42 +2,64 @@
...
+..
+    * [we dont have a link to these but we do have some explanation pages] Topic guides (name?) provide
+
+Further Info
+************
+
+* Find out :doc:`what's new <whatsnew/index>` with sunpy.
+* View a list of :doc:`known issues <reference/known_issues>` or report new issues on our `GitHub issue tracker`_.
+* Join our `chat room`_ or post to our `discourse`_ for help.
+* Read the :doc:`Developers guide <dev_guide/index>` to start contributing to the sunpy project.
```
dstansby, line 47: This is excellent 💯

dstansby (commented):
Nice! My general comment is I think we should try and mirror the structure of the landing page with our toctree, and then include toctrees in the actual landing page instead of having a hidden toctree.

Cadair:
> My general comment is I think we should try and mirror the structure of the landing page with our toctree, and then include toctrees in the actual landing page instead of having a hidden toctree.

I am unconvinced by this as a pattern (although I am very open to be proven wrong). The toctree(s) primary function is to determine the layout of the left nav bar. I don't see any reason why the presentation of the content on the landing page has to exactly mirror the links we want to show and the order we want to show them in the sidebar? We might want them to be the same or we might not (now or in the future) so why not keep them separate and keep the option?

docs/index.rst
```diff
@@ -2,42 +2,64 @@
...
+``sunpy`` is a community-developed, free and open-source solar
 data analysis environment for Python.
 
+New Users
+*********
+
+If you are a first time user or new to Python,
+
+..
+        * :Introduction to sunpy core [new page described in #6615]
```
Cadair, line 14: There is actually a sphinx extension which supports todo items and tracking them, might be worth a look: https://www.sphinx-doc.org/en/master/usage/extensions/todo.html ?

docs/index.rst
Cadair: See my comment on the main thread. I actually think there's potential value in the separation.

In this case I think not having installation guide in the sidebar is an improvement (I got rid of it in #6613)

docs/index.rst
Cadair, line 16: We should avoid using the `:doc:` role where possible and use `:ref:` with the reference names of the titles we are referencing. This is so that if we change the path the links still work.

alasdairwilson:
> I am unconvinced by this as a pattern (although I am very open to be proven wrong). The toctree(s) primary function is to determine the layout of the left nav bar. I don't see any reason why the presentation of the content on the landing page has to exactly mirror the links we want to show and the order we want to show them in the sidebar? We might want them to be the same or we might not (now or in the future) so why not keep them separate and keep the option?

So I hadn't realised you had done the same thing in #6613 but we definitely reached the same conclusion. We currently have a table of links (sidebar) and we present the same data in (two) tables on the page, the idea here was to retain the useful sidebar nav but to present the same information in a different way on the page. I agree this retains more flexibility.

alasdairwilson:
So I tried to address the concerns from the code review:

Did all the small style changes form @dstansby and replaced docs with refs for all but the example gallery.

I addressed the concerns over manually populating the tutorial subsections by adding them as a toctree, this means changes to the tutorial will automatically update the landing page.

There  are also 2 additional commits that can be taken or left (or cherry picked for another PR):

One removes the installation guide from the tutorial and instead adds it to the blurb at the beginning of the guide, removing it from the tutorial toctree on the landing page (which also removes it from the sidebar). I am torn on this, think neither way is ideal but we retain the ability to manually populate the sidebar by readding it as a hidden toctree or just removing that last commit which will mean the installation guide appears in the sidebar of /guide and the tutorial toctree on the landing page.

Added Sphinx-ext-todo which, to me looks useful, add a `.. todo::` tag, the same as a `.. note::` tag and it will not render unless you set `todo_include_todos` i.e.  `sphinx-build -D todo_include_todos=1 -b html . _build/html`

This generates this really cool page that I have put at html/todo.html :

image

I think this could be useful, especially during a time when the docs are in flux as reminders of parts that need completing or parts that will need revisited (as in the case) once other tasks are completed.

Cadair:
> One removes the installation guide from the tutorial and instead adds it to the blurb at the beginning of the guide, removing it from the tutorial toctree 

Having the install guide not appear in the menus, either at top level or in the tutorial, seems weird. What's the thinking for not having it in the tutorial toctree?

docs/index.rst
```diff
@@ -2,42 +2,74 @@
...
+
+.. todo::
+
+    if #6615 is completed link here
+    Introduction to sunpy core
+
+If you are a first time user or new to Python, our **tutorial** provides a walkthrough on how to use the key features of sunpy core.
+Once you've installed sunpy, start here.
+
+* :ref:`guide`
```
Cadair, line 22: I thought you had put a toctree here?

alasdairwilson:
Think that's it, sadly the changes in #6616 meant the solution to auto generating a tutorial index on this page no longer works. I suspect there is probably a good solution along the lines of having the layout of the tutorial stored in a separate file and included in both places, preventing them getting out of sync later, but out there but out of scope I think here.

docs/index.rst
alasdairwilson: see my comment in main thread, because of #6616 it now breaks the toctree here

dstansby (approved):

docs/index.rst
```diff
@@ -2,42 +2,74 @@
 sunpy core Documentation
 ************************
 
-The sunpy core package is a community-developed, free and open-source solar
-data analysis environment for Python.
+``sunpy`` is a community-developed, free and open-source solar data analysis environment for Python.
```
nabobalis, line 5: Do we want to enforce all sunpy library references are marked up like this?

docs/index.rst
```diff
@@ -2,42 +2,74 @@
...
+    [Topic guides] discuss key topics and concepts at a fairly high level and provide useful background information and explanation.
+
+
+Further Info
+============
+
+* Find out :ref:`what's new <whatsnew>` with sunpy.
+* View a list of :ref:`known issues <known_issues>` or report new issues on our `GitHub issue tracker`_.
+* Join our `chat room`_ or post to our `discourse`_ for help.
+* Read the :ref:`Developers guide <dev_guide>` to start contributing to the sunpy project.
```
nabobalis, line 56:
```suggestion
* Read the :ref:`Developers guide <dev_guide>` to start contributing to the SunPy project.
```

docs/index.rst
```diff
@@ -2,42 +2,74 @@
...
+``sunpy`` is a community-developed, free and open-source solar data analysis environment for Python.
+
+
+New Users
+==========
+
+For help installing ``sunpy`` refer to our :ref:`installation guide <installing>`.
+
+
+.. todo::
```
nabobalis, line 14: Are we ok with adding todos in files? 

Would an issue not be a better idea?

docs/index.rst
dstansby: I think it looks good here - if we want to disucss enforcing (or not) as a rule, worth opening a new issue?

docs/index.rst
dstansby: I'm 👍 to opening an issue too, so we don't forget.

docs/index.rst
alasdairwilson: I had a look over the docs and its a bit all over the place how sunpy is stylised: ``sunpy``, SunPy, sunpy core (package), the sunpy project etc.

If there are rules to follow or rules to be made then we should do that. I don't care how it is stylised here really.

nabobalis: We need to enforce this down the road but for now, we don't need to block.

Cadair (approved):

dstansby:
:shipit:

PR #6628, opened 2022-11-22
Try unpinning pytest{-xdist}

dstansby:
xref https://github.com/sunpy/sunpy/issues/6505. pytest 7.2.0 seems to be working locally for me, so lets see what CI has to say about that.

### DIFF
--- a/.test_package_pins.txt
+++ b/.test_package_pins.txt
@@ -1,2 +1,2 @@
-pytest<7.2
-pytest-xdist<3.0
+# This file is used to pin test packages across sunpy coordinated packages
+# Do not remove, even if it's empty!
--- a/tox.ini
+++ b/tox.ini
@@ -124,9 +124,6 @@ pypi_filter =
 extras =
 deps =
 conda_env = sunpy-dev-env.yml
-conda_deps =
-    pytest<7.2
-    pytest-xdist<3.0
 conda_channels = conda-forge
 install_command = pip install --no-deps --no-build-isolation {opts} {packages}
 allowlist_externals = conda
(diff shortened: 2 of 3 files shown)


dstansby:
Looks like we're running into https://github.com/pytest-dev/pytest-cov/issues/557

tox.ini
```diff
@@ -11,7 +11,6 @@ requires =
 isolated_build = true
 
 [testenv]
-pypi_filter = file://.test_package_pins.txt
```
nabobalis, line 14: Don't we want this?

.test_package_pins.txt
```diff
@@ -1,2 +0,0 @@
-pytest<7.2
```
nabobalis, line 1: Don't we want to keep this file?

dstansby:
I figured since there was nothing in it the test_package_pins could go - there's no comment in it about what it does, so if it's going to stay (and be empty) could you suggest a comment to put in it describing its purpose?

Cadair:
The file is used by `tox-pypi-filter` as a place for us to exclude packages from test builds for all the sponsored packages. (i.e. [ndcube](https://github.com/sunpy/ndcube/blob/main/tox.ini#L13)) so we shouldn't remove the file even if it's empty as that will cause the link to stop working, and also we want to keep the config in the tox file to make it nice and easy to exclude packages in the future.

Hopefully that provides enough context to document it :see_no_evil:

dstansby:
👍 put back with comment

nabobalis (approved):

nabobalis:
I wonder if we will need to manually backport something for 4.0 later on.

dstansby:
@MeeseeksDev backport to 4.0

Issue #6467, opened 2022-10-12
HGS Ticklabels overlap with plot title in "Cropping a Map" example


wtbarnes:
In [the example on cropping a map](https://docs.sunpy.org/en/stable/generated/gallery/map/submaps_and_cropping.html#sphx-glr-generated-gallery-map-submaps-and-cropping-py), we show an HGS grid overlay on a cropped map (see below). However, the tick labels for the HGS grid on the top overlap with the plot title. 

Interestingly, @ayshih says that similar overlaps with HGS axislabels and ticklables  *do not* happen with the following combination of packages:
- sunpy 3.1.8
- astropy 4.3.1
- matplotlib 3.2.2

Whether this is a regression or not is not clear.

![image](https://user-images.githubusercontent.com/8676141/195428910-84d8f83a-d2b1-44af-a884-231c986ae403.png)

ayshih:
Well, the overlap is different with the versions that @wtbarnes listed (which are the versions that are installable on Google Colab), rather than entirely eliminated:
![download (7)](https://user-images.githubusercontent.com/991759/195433726-a1e2cc68-67e8-469d-99f0-006f51b6b750.png)
It seems that one of the packages used to shift the title to make some (but not necessarily enough) room for labels, while the latest packages aren't trying at all.

ayshih:
Looking at the above images again, it looks like the image I posted has the title properly padded for the tick labels, but not for the axis label.  That's in contrast to the image @wtbarnes posted where the title isn't properly padded for either.

dstansby:
Upstream issue reported at https://github.com/astropy/astropy/issues/14047

PR #6629, opened 2022-11-22
Fix title in SWAP example

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/6467. A nicer fix really needs to be done upstream in astropy, but this should fix our example for now.

### DIFF
--- a/examples/map/submaps_and_cropping.py
+++ b/examples/map/submaps_and_cropping.py
@@ -29,8 +29,14 @@ swap_submap = swap_map.submap(bottom_left, top_right=top_right)
 ###############################################################################
 # Let's plot the results.
 
-plt.figure()
-swap_submap.plot()
+fig = plt.figure()
+image = swap_submap.plot()
 swap_submap.draw_limb()
 swap_submap.draw_grid()
+
+# Make some room and put the title at the top of the figure
+ax = image.axes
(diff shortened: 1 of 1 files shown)


Cadair:
If we can't (yet) fix this upstream, can't we fix it in `GenericMap.plot`?

ayshih:
Two comments:

- The juggling of titles seems a bit silly.  How about we pad the axes title as we do in another example? https://github.com/sunpy/sunpy/blob/e2b2b5a2f1f26699ae8bbdd505f145af90763d53/examples/units_and_coordinates/AIA_limb_STEREO.py#L56-L58
- Putting this in `GenericMap.plot()` doesn't make sense to me, because it depends on whether there are tick labels to avoid.  Plus, as #6467 shows, the amount to pad the title differs with dependency versions.  If we understood the issue well enough such that `GenericMap.plot()` would do what we want under all situations, then we presumably would know enough to fix WCSAxes outright.

dstansby:
> The juggling of titles seems a bit silly. How about we pad the axes title as we do in another example? 

Much better, thanks for the suggestion.

> Putting this in GenericMap.plot() doesn't make sense to me

Agreed - if anyone wants to try tackling this upstream I've opened https://github.com/astropy/astropy/issues/14047.

nabobalis (approved):

PR #6642, opened 2022-12-03
Fix NOAA ind plotting on devdeps

dstansby:
I have no idea why this is failing - the only difference in packages between passing and failing tests on circleCI that could have caused the failure was different commits for Matplotlib dev version. I had a look at Matplotlib commits and couldn't find anything obvious that would casue the failure.

I can't reproduce outside tox, so this is my best guess at a fix...


dstansby:
🤷 seemed to work, so I think this is good for review

nabobalis (approved):

PR #6654, opened 2022-12-09
Tox 4.0 and drop tox-conda

Cadair:


Cadair:
I am working on updating tox-pypi-filter, so there should be a release of that soon and then we can merge this.

tox.ini
```diff
@@ -7,8 +7,8 @@ envlist =
 requires =
     setuptools >=56, !=61.0.0
     pip >= 19.3.1
-    tox-pypi-filter >= 0.12
-isolated_build = true
+    # Currently tox-pypi-filter is not compatible with tox 4
+    #tox-pypi-filter >= 0.12
```
dstansby, line 11: Do we actually use `tox-pypi-filter` for anything? Looks like it was added to the config by @nabobalis in https://github.com/sunpy/sunpy/pull/5161, but that PR didn't actually invoke it using `--pypi-filter`, and it doesn't look like we currently invoke it.

Cadair: It's enabled automatically by it being in the config.

dstansby: 🤔 but the [tox-pypi-filter readme](https://github.com/astrofrog/tox-pypi-filter#readme) doesn't say anything about it enabling anything automatically

dstansby: (not disagreeing, with you, just trying to work out what it's doing and why we can't just stop using it!)

Cadair: tox pypi filter allows us to globally exclude versions of dependencies, or even transitive dependencies which are causing the tests to fail without pinning them down in our package metadata. For example, if lxml which is a dependency of zeep had an issue we can ignore it until a bugfix is released for all our tox tests. Or if there's a specific issue with pytest itself we can ignore it for all packages using our pins file in one commit.

dstansby:
Are we going to wait for `tox-pypi-filter` to be fixed for this PR then?

Cadair:
That was my intention.

dstansby:
👍  - cross posting for reference https://github.com/astrofrog/tox-pypi-filter/pull/8

Cadair:
new release of tox-pypi-filter is out and this should be good to go.

tox.ini
```diff
@@ -7,8 +8,7 @@ envlist =
 requires =
     setuptools >=56, !=61.0.0
     pip >= 19.3.1
-    tox-pypi-filter >= 0.12
-isolated_build = true
```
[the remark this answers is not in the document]
dstansby: If I'm correct here, feel free to self-merge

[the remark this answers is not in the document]
Cadair: yes, indeed. I think it's actually mandated in tox 4, but could be wrong.

Cadair:
Do not merge: the conda fail is real.

nabobalis:
Can we enforce tox>=4.0 or does it not matter?

Cadair:
> Can we enforce tox>=4.0 or does it not matter?

I have as far as the tox config goes: 
https://github.com/sunpy/sunpy/pull/6654/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449R2

dstansby:
Looks like this is now blocked on https://github.com/tox-dev/tox-conda/issues/156

Cadair:
Ok, I hacked it. I made the conda job still use tox 3 as it doesn't look like a fix is likely very fast there. This allows us to use tox 4 when developing locally (`--develop` now works, which speeds up testing *a lot*).

Cadair:
I didn't re-run the CI on the last commit, the last run is here: https://github.com/sunpy/sunpy/actions/runs/4426012546

sunpy-dev-env.yml
```diff
@@ -2,6 +2,13 @@ channels:
   - conda-forge
 
 dependencies:
+  # Build time (taken from pyproject.toml)
+  - setuptools
+  - setuptools-scm
+  - extension-helpers
+  - pip
+  - numpy
+
```
ayshih, line 11: Er, all of these dependencies are already explicitly specified later in the environment file, just not grouped in one place: `numpy` and `setuptools` under "Required" and the rest under "Installation".  Do you really want to have these dependencies specified twice?

nabobalis:
> Ok, I hacked it. I made the conda job still use tox 3 as it doesn't look like a fix is likely very fast there. This allows us to use tox 4 when developing locally (`--develop` now works, which speeds up testing _a lot_).

What is `--develop`?

sunpy-dev-env.yml
Cadair: It didn't work anyway, so I could just pull this change I think.

nabobalis (approved):

nabobalis:
> This is now ready. I have moved us away from tox-conda as it was becoming more trouble than it was worth. With the env file as it is, the actions code to run the tests in conda was simple.

I can roll this out to other packages and we can remove it.

Cadair:
argh, the codecov upload on conda was still broken

nabobalis:
I don't think we need codecov for conda.

PR #6660, opened 2022-12-14
Add code to generate tutorial graph

dstansby:
I wrote this a while ago, and don't want to lose it, so figured saving it in the tools folder was worthwhile.


nabobalis:
pre-commit.ci autofix

nabobalis (approved):

PR #6674, opened 2022-12-29
[4.0] pre-commit updates

nabobalis:
Updated pre-commit from main, fixed issues.
Two backports as well.

### DIFF
--- a/sunpy/net/hek/attrs.py
+++ b/sunpy/net/hek/attrs.py
@@ -34,7 +34,7 @@ class HEKAttr(_attr.AttrComparison):
     """
 
 
-class HEKComparisonParamAttrWrapper(_attr.ComparisonParamAttrWrapper):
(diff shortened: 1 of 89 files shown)


sunpy/net/hek/attrs.py
```diff
@@ -60,6 +60,23 @@ def collides(self, other):
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        # Do this here to not clutter the namespace
+        from sunpy.util.exceptions import SunpyDeprecationWarning, _warn
+
+        name = type(self).__name__
+        _warn(f"`sunpy.net.hek.attrs.{name}` is deprecated, please use `sunpy.net.attrs.{name}`",
+              SunpyDeprecationWarning)
+        super().__init__(*args, **kwargs)
```
nabobalis, line 77: I think this is wrong. I need to double check, I think this was removed in 4.0

PR #6677, opened 2022-12-29
add numpy to the devdeps builds

nabobalis:
We forgot to test numpy in devdeps.


dstansby:
It looks like we don't have a GitHub actions devdeps run at the moment? There's the figure devdeps run, but that only runs the figure tests. 👍 to including numpy, but would be good to enable a GH Actions run with all the tests under devdeps.

ConorMacBride:
The devdeps job only runs in the cron group, not in PRs. Mostly because it's unlikely that a devdeps failure is related to changes in a PR.

dstansby:
Ah, thanks for the reminder! Do we need to close and re-open this to get the CRON CI running? Adding the label doesn't seem to have triggered it

dstansby (approved):

PR #6678, opened 2022-12-29
Use oldestdeps for figure tests

nabobalis:
That time of the year.


nabobalis:
Figure test is failing on:
```
FAILED ../../.tox/py310-figure/lib/python3.10/site-packages/sunpy/map/tests/test_plotting.py::test_draw_limb_heliographic_stonyhurst - RuntimeWarning: invalid value encountered in cast
```

tox.ini
```diff
@@ -68,21 +68,20 @@ deps =
     online: astroquery
     hypothesis: astroquery
     # Figure tests need a tightly controlled environment
-    figure-!devdeps: astropy==5.1.0
-    figure-!devdeps: matplotlib==3.5.2
+    figure-!devdeps: astropy==5.2.0
+    figure-!devdeps: matplotlib==3.6.2
```
dstansby, line 72: As a general testing strategy I think it's worth keeping these pinned to the lowest versions we support (of both astropy and Matplotlib), and then using `devdeps` as the most-recent version test. So I'm 👎 on this bump (unless it's time to bump these in our minimum dependencies too?)

tox.ini
```diff
@@ -68,21 +68,20 @@ deps =
...
     figure-!devdeps: mpl-animators==1.0.0
-    figure-!devdeps: numpy<1.24
```
dstansby, line 74: Again, I think it's worth keeping this as the minimum version we support.

tox.ini
```diff
@@ -68,21 +68,20 @@ deps =
...
 extras =
     all
     tests
 commands =
-    pip freeze --all --no-input
+    !figure: pip freeze --all --no-input
```
dstansby, line 78: I think it's still worth printing this to the log (as well as piping it ot a file later) - I've found it useful recently.

dstansby (changes requested):

tox.ini
nabobalis: These versions have never been the oldest ones we support. We normally bump these when we encounter problems in the build like we had to pin numpy. This removes the pin below.

tox.ini
nabobalis: The base figure tests do not test for min versions, this is not the earliest version we use. This was pinned to avoid the issue with the deprecation warning.

tox.ini
nabobalis: We don't need this as the commands below also cat the file into the logs.
This avoids a repeat of this output.

dstansby:
Is there a good reason not to use the non-devdeps figure tests to check our minimum supported versions of Matplotlib, astropy, numpy etc.?

nabobalis:
> Is there a good reason not to use the non-devdeps figure tests to check our minimum supported versions of Matplotlib, astropy, numpy etc.?

Considering the pains we sometimes get into with the `oldestdeps` build I don't think I am in favour of it.

But nothing is stopping us.

dstansby:
In that case I think we should adopt this strategy - if we have pains on our CI with minimum deps, then we should sort them out or bump them so users can use minimum deps!

dstansby:
I think we decided on a call at some point to keep these as minimum versions we support? In which case I'll close this, and I've opened another PR at https://github.com/sunpy/sunpy/pull/6849 to add a comment to this effect.

PR #6679, opened 2022-12-29
Update test that was marked xfail due to numpy version and trigger warning also based on numpy version

nabobalis:
The test almost always fails and anyone who gets the warning will also most likely raise a error from numpy when you try to pad a nan to a int.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -5,7 +5,6 @@ import re
 import copy
 import html
 import inspect
-import numbers
 import textwrap
 import itertools
 import webbrowser
@@ -1687,20 +1686,13 @@ class GenericMap(NDData):
         unpad_y = -np.min((diff[1], 0))
 
         # Pad the image array
-
-        if issubclass(self.data.dtype.type, numbers.Integral) and (missing % 1 != 0):
-            warn_user("The specified `missing` value is not an integer, but the data "
-                      "array is of integer type, so the output may be strange.")
-
         new_data = np.pad(self.data,
                           ((pad_y, pad_y), (pad_x, pad_x)),
                           mode='constant',
                           constant_values=(missing, missing))
 
         # All of the following pixel calculations use a pixel origin of 0
-
         pixel_array_center = (np.flipud(new_data.shape) - 1) / 2.0
-
         pixel_rotation_center = u.Quantity(self.reference_pixel).value + [pad_x, pad_y]
 
         if recenter:
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -5,6 +5,7 @@ import re
 import copy
 import html
 import inspect
+import numbers
 import textwrap
 import itertools
 import webbrowser
@@ -15,6 +16,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 from matplotlib.backend_bases import FigureCanvasBase
 from matplotlib.figure import Figure
+from packaging import version
 
 try:
     from dask.array import Array as DaskArray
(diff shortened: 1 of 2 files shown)


sunpy/map/mapbase.py
```diff
@@ -1687,20 +1686,13 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=3, scale=1.0,
         unpad_y = -np.min((diff[1], 0))
 
         # Pad the image array
-
-        if issubclass(self.data.dtype.type, numbers.Integral) and (missing % 1 != 0):
-            warn_user("The specified `missing` value is not an integer, but the data "
-                      "array is of integer type, so the output may be strange.")
```
nabobalis, line 1693: So with numpy >= 1.2.0, this warning will be raised and then the pad below will error.

So I think we can remove this now or maybe add a check for older versions of numpy ?

dstansby:
The versions in the xfail check (1.2.0) and message (1.20) don't agree - I think the version check should be fixed to `1.20`? And then we should keep these tests, and check it runs on our mindeps run, as our minimum numpy is currently 1.17

nabobalis:
Ah yeah, even with a bump to astropy 5.1, we aren't clear of numpy 1.20 for a long time.

I will add it back and change the test.

dstansby (changes requested):
Looks like `oldesdeps` is failing 😢

nabobalis:
Stupid typo on my part.

sunpy/map/tests/test_mapbase.py
```diff
@@ -941,9 +941,9 @@ def test_rotate(aia171_test_map):
     assert aia171_test_map_crop_rot.data.shape[0] < aia171_test_map_crop_rot.data.shape[1]
 
 
-@pytest.mark.xfail(version.parse(np.__version__) >= version.parse("1.2.0"),
-                   reason="Numpy >= 1.20.0 doesn't allow NaN to int conversion")
-def test_rotate_with_incompatible_missing_dtype():
+@pytest.mark.skipif(version.parse(np.__version__) >= version.parse("1.2.0"),
```
nabobalis, line 944:
```suggestion
@pytest.mark.skipif(version.parse(np.__version__) >= version.parse("1.20.0"),
```

sunpy/map/tests/test_mapbase.py
```diff
@@ -955,6 +955,18 @@ def test_rotate_with_incompatible_missing_dtype():
         test_map.rotate(order=3, missing=np.nan)
 
 
+@pytest.mark.skipif(version.parse(np.__version__) <= version.parse("1.2.0"),
```
nabobalis, line 958:
```suggestion
@pytest.mark.skipif(version.parse(np.__version__) <= version.parse("1.20.0"),
```

dstansby (changes requested):
Tests are still failing on this

nabobalis:
> Tests are still failing on this

Not sure why they failed, I've done a rebase and they now pass. Will see if oldest deps agrees.

sunpy/map/tests/test_mapbase.py
```diff
@@ -941,9 +941,9 @@ def test_rotate(aia171_test_map):
     assert aia171_test_map_crop_rot.data.shape[0] < aia171_test_map_crop_rot.data.shape[1]
 
 
-@pytest.mark.xfail(version.parse(np.__version__) >= version.parse("1.2.0"),
-                   reason="Numpy >= 1.20.0 doesn't allow NaN to int conversion")
-def test_rotate_with_incompatible_missing_dtype():
+@pytest.mark.skipif(version.parse(np.__version__) >= version.parse("1.20.0"),
```
dstansby, line 944:
```suggestion
@pytest.mark.xfail(version.parse(np.__version__) >= version.parse("1.20.0"),
```
I think we should check this fails instead of just skipping? Similarly for the new test below.

nabobalis: Good point.

PR #6684, opened 2022-12-30
Fix database test

dstansby:
I'm not sure why this has changed remotely, but it has, so update the filename we test against.

### DIFF
--- a/sunpy/database/tests/test_tables.py
+++ b/sunpy/database/tests/test_tables.py
@@ -190,7 +190,7 @@ def test_entries_from_fido_search_result(fido_search_result):
         wavemin=17634850.470588233, wavemax=17634850.470588233,
         instrument='NORH')
     # 2 entries from rhessi
-    assert entries[66].fileid.endswith("/metadata/catalog/hsi_obssumm_20120101_032.fits")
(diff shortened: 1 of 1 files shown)


nabobalis:
I saw on some CI but sometimes it was passing and locally as well. I was quite confused.

nabobalis (approved):

nabobalis:
The test failed with:

```
        # 2 entries from rhessi
>       assert entries[66].fileid.endswith("/metadata/catalog/hsi_obssumm_20120101_016.fits")
E       AssertionError: assert False
E        +  where False = <built-in method endswith of numpy.str_ object at 0x7f0f275d3d50>('/metadata/catalog/hsi_obssumm_20120101_016.fits')
E        +    where <built-in method endswith of numpy.str_ object at 0x7f0f275d3d50> = 'http://hessi.ssl.berkeley.edu/hessidata/metadata/catalog/hsi_obssumm_20120101_032.fits'.endswith
```

The test is sometimes returning the other file?

dstansby:
Boooo. I've updated the test to check the part of the filename that isn't changing (but still contains a timestamp)

dstansby:
🥳

PR #6693, opened 2023-01-05
detect gzipped-FITS files by magic-bytes

### LINKED ISSUE #6692: improve gzipped-FITS files detection
### Describe the feature

Currently the file detection for gzipped-FITS files relies on checking if they are named in the `filename.fits.gz` format. However since this condition isn't a necessity for gzipped files, this approach misses those gzipped-FITS files which do not end with the `.gz` extension.

### Proposed solution

Use magic bytes / file signatures to check whether the file is gzipped or not. If it is, decompress and read its first few bytes to determine whether it's a FITS file.

exitflynn:
## PR Description
Detection for gzipped-FITS files is now done by checking file signatures instead of checking for file extensions.
First a gzip file-signature is checked and if found, the file's first few bytes are decompressed and checked for the FITS file signature.
The PR closes  #6692.


nabobalis:
Hi @exitflynn, thanks for the PR. If you could add a unit test for this in https://github.com/sunpy/sunpy/blob/main/sunpy/io/tests/test_filetools.py 

There should be some gzipped files to use in https://github.com/sunpy/sunpy/tree/main/sunpy/data/test

sunpy/io/file_tools.py
```diff
@@ -214,14 +214,16 @@ def detect_filetype(filepath):
...
-    filepath_rest_ext1, ext1 = os.path.splitext(filepath)
-    _, ext2 = os.path.splitext(filepath_rest_ext1)
-
-    gzip_extensions = [".gz"]
-    fits_extensions = [".fts", ".fit", ".fits"]
-    if (ext1 in gzip_extensions and ext2 in fits_extensions):
-        return 'fits'
+    # Checks for gzip signature. 
+    # If found, decompresses first few bytes and checks for FITS
+    if cdf_magic_number[:4] == '1f8b':
```
ConorMacBride, line 219: I would rather avoid using the `cdf_magic_number` variable here as it's semantically intended for detecting CDF files. In the code above `first80` is intended for FITS. Astropy detect with an extra byte [in this code](https://github.com/astropy/astropy/blob/f1620baa9f0a5228cac0c8bed22f42e4f4243896/astropy/io/fits/file.py#LL98C14-L98C15), so maybe we should check that too?
```suggestion
    if first80[:3] == b"\x1f\x8b\x08":
```
They also support PKZIP and BZIP2, which I don't think sunpy will detect, but maybe they are not as relevant to solar data though?

nabobalis: I've not seen solar data with those but doesn't mean we won't in the future.

sunpy/io/file_tools.py
```diff
@@ -214,14 +214,16 @@ def detect_filetype(filepath):
...
+    # Checks for gzip signature. 
+    # If found, decompresses first few bytes and checks for FITS
+    if first80[:3] == b"\x1f\x8b\x08":
+        import gzip
+        with gzip.open(filepath, 'rb') as fp:
+            first80 = fp.read(80)
+            match = re.match(br"[A-Z0-9_]{0,8} *=", first80)
+            if match is not None:
+                return 'fits'
+
```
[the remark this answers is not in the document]
exitflynn: oh yes, absolutely. that sounds like a great way to go about it.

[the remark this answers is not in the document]
nabobalis: Yeah looking at this code, we should simplify where possible.

exitflynn:
sure. i'll be getting to including PKZIP and BZIP2 detection and the tests tomorrow! (it's quite late here)

nabobalis:
No problem. I also wouldn't add the other two file formats for now.

exitflynn:
Hi, I just noticed that a test for detecting gzipped FITS files [already exists](https://github.com/sunpy/sunpy/blob/main/sunpy/io/tests/test_filetools.py#L50-L59), since i only re-modified how this was implemented.

exitflynn:
also i'm not sure why the CI tests have failed, is it something unrelated to my changes that I can ignore? I tried runing the tests before pushing by running `test_filetools.py` and it didn't raise any issue (this is my first time testing stuff).

sunpy/io/file_tools.py
```diff
@@ -214,14 +214,12 @@ def detect_filetype(filepath):
...
-    _, ext2 = os.path.splitext(filepath_rest_ext1)
-
-    gzip_extensions = [".gz"]
-    fits_extensions = [".fts", ".fit", ".fits"]
-    if (ext1 in gzip_extensions and ext2 in fits_extensions):
-        return 'fits'
+    # Checks for gzip signature. 
+    # If found, decompresses first few bytes and checks for FITS
+    if first80[:3] == b"\x1f\x8b\x08":
+        import gzip
```
Cadair, line 220: I don't think there's a lot to gain by putting this here, we can move it to the top of the file.

Cadair:
> Hi, I just noticed that a test for detecting gzipped FITS files [already exists](https://github.com/sunpy/sunpy/blob/main/sunpy/io/tests/test_filetools.py#L50-L59), since i only re-modified how this was implemented.

Could you add a test where the filextension doesn't match but we still detect it properly? This means we test this new behaviour and we make sure that we don't regress in the future?

Cadair:
> also i'm not sure why the CI tests have failed, is it something unrelated to my changes that I can ignore? I tried runing the tests before pushing by running `test_filetools.py` and it didn't raise any issue (this is my first time testing stuff).

It's not your fault, if you can rebase or merge in main then it should be fixed now #6694

exitflynn:
right now there are two tests for gzipped-fits files, the `test_read_file_fits_gzip()` which tests `filename.fits.gz`, `-.fts.gz` and `-fit.gz` files (namely the ones that were included there when detection was done based on the extension) and the `test_read_file_fits_gzip_not_gz_extension()` which tests `gzip_fits_test.file`. 

I think it'd make more sense to have only one test for gzip-fits files, the latter one since if it passes then there's no need to test for the first one. 

Should I remove the first test (renaming the second one to `test_read_file_fits_gzip()`)? If yes, it'd also make the three test `-fits.gz` files redundant, should i delete those too?

nabobalis:
If you can combine the two tests, I would prefer that.

exitflynn:
in that case, is combining them like that alright?

nabobalis:
Looks fine to me.

nabobalis:
If you could add a bugfix changelog following https://github.com/sunpy/sunpy/blob/main/changelog/README.rst

and if you could also deal with the pre-commit.ci fail.
You can do that either installing pre-commit, running the tox -e codestyle or checking the failure and fixing it manually.

exitflynn:
yep, done! (I mean I think that should do it? I can't figure out what's causing the `codecov/project` check to fail... :thinking: )

changelog/6693.feature.rest
```diff
@@ -0,0 +1,3 @@
+Able to detect gzip-compressed FITS files even if they don't have the .gz extension in filename.
+`~sunpy.io.file_tools.detect_filetype` now looks for the right file signature while checking
```
ConorMacBride, line 2: Should fix the documentation build:
```suggestion
`~sunpy.io.detect_filetype` now looks for the right file signature while checking
```
https://docs.sunpy.org/en/latest/generated/api/sunpy.io.detect_filetype.html#detect-filetype

changelog/6693.feature.rest
```diff
@@ -0,0 +1,3 @@
+Able to detect gzip-compressed FITS files even if they don't have the .gz extension in filename.
```
ConorMacBride, line 1:
```suggestion
Able to detect gzip-compressed FITS files even if they don't have the ``.gz`` extension in the filename.
```

exitflynn:
thanks @ConorMacBride! done 👍

nabobalis (approved):

ConorMacBride (approved):
Thanks @exitflynn!

exitflynn:
:cowboy_hat_face:  Thanks Conor, Nabil and Stuart for your time!

PR #6705, opened 2023-01-11
Fix doc path

nabobalis:
Need to check the output is sane.


nabobalis:
Seems to not make a difference, locally, the path did not render for me.

ConorMacBride:
@nabobalis There's an example in the docs which should probably work https://tox.wiki/en/latest/faq.html#building-documentation-with-sphinx

nabobalis:
I encouthered this issue while adding tox to the website repo.

The raw string was not rendering half of the path for me.

Issue #6706, opened 2023-01-12
Inconsistent results from `_parse_submap_input` when using coordinates versus pixel values


wtbarnes:
### Describe the bug

When parsing possible inputs to `submap`, the function [`_parse_submap_inputs`](https://github.com/sunpy/sunpy/blob/ae2a93f7ed438b1e5a3e7a49231541c246b7c80b/sunpy/map/mapbase.py#L1947) is used to get the pixel coordinates of all corners of the inputs, regardless of whether the inputs are coordinates or quantities, specified in widths and heights, etc.

I would expect then that, for a given set of bottom left and top right corners, whether specified in pixel or world coordinates, that they should return the same set of pixel coordinates. In particular, for the bottom left and top right corners of a map, I would expect that this function should always return `((0,0), (N-1,0), (N-1,N-1),(0,N-1))`, for an image of dimensions (N,N), where the ordering of the result here is (bottom left, bottom_right, top_right, top_left). This is indeed the case for pixel coordinates but is not the case for those same coordinates when specified in terms of their world coordinates. 

Interestingly, in the world coordinate input case, some of the coordinates are off by several pixels.

### To Reproduce

```python
import sunpy.data.sample
import sunpy.map
import astropy.units as u

m = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)

# Get pixel coordinates of corners when specified in world coordinates
corners_coord = m._parse_submap_input(m.bottom_left_coord, m.top_right_coord, None, None)
print('Corners from coordinates')
print(corners_coord)
  
# Convert coordinates to pixel space and pass them through the same function
blc_pix = m.world_to_pixel(m.bottom_left_coord)*u.pix
trc_pix = m.world_to_pixel(m.top_right_coord)*u.pix
corners_pix = m._parse_submap_input(blc_pix,trc_pix,None, None)
print('Corners from pixels')
print(corners_pix)
```

gives the result,

```
Corners from coordinates
(<Quantity [-7.96376298e-11, -9.09494702e-12] pix>, <Quantity [  -2.47269596, 1020.51525106] pix>, <Quantity [1023., 1023.] pix>, <Quantity [1025.47277787,    2.48474914] pix>)
Corners from pixels
[...]

Issue #4896, opened 2021-01-21
Submap has a faulty assumption when given coordinates in a different frame


ayshih:
`GenericMap.submap()` has a faulty assumption when given coordinates in a different frame: it assumes that the bounding box defined by the four corners of the input quadrangle include all other points within that quadrangle.  Here's an illustrative picture of the problem.  For an AIA map in helioprojective, if we define the quadrangle shown in blue using heliographic Stonyhurst coordinates, the `submap()` call will return the region shown in red.
![download](https://user-images.githubusercontent.com/991759/105371498-abe36b80-5bd2-11eb-9a4e-7c52e820ac90.png)

ayshih:
```python
import matplotlib.pyplot as plt

from astropy.coordinates import SkyCoord
import astropy.units as u

import sunpy.map
from sunpy.coordinates.utils import get_rectangle_coordinates
from sunpy.data.sample import AIA_171_IMAGE

m = sunpy.map.Map(AIA_171_IMAGE)

bottom_left = SkyCoord(40*u.deg, -45*u.deg, frame='heliographic_stonyhurst', obstime=m.date)
sm = m.submap(bottom_left, width=10*u.deg, height=90*u.deg)

fig = plt.figure(figsize=(8, 8))
ax = fig.add_subplot(111, projection=m)

sm.plot()
```
![download (1)](https://user-images.githubusercontent.com/991759/105372177-52c80780-5bd3-11eb-888e-e54344cdefcc.png)

Cadair:
I would agree we documented the current algorithm correctly but ideally I think it would return the bounding box containing the lines of constant lon/lat defined by the corners.

jeffreypaul15:
Is this a requirement? I suppose this is possible through astropy's ```Quadrangle``` and ```Numpy```?

Cadair:
From an out of band discussion with @ayshih about looking at implementing this in `NDCube.crop`:

I think there are probably 2 main algorithmic approaches here:

1) Generate the edges of the bounding box in world space at some sampling resolution, transform these to pixel space, select the min/max pixel indices which give a rectangular region in pixel space containing the whole ROI (in pixel space).
2) Convert all the pixel coordinates to world coordinates, perform a logical operation on the world coordinates calculating which ones fall inside the ROI, generate a pixel bounding box which encompasses all points.

[...]

PR #6707, opened 2023-01-12
Allow padding in submap when bounding coordinates exceed original map boundaries

wtbarnes:
This PR makes a change to `submap` that pads maps when the coordinates that define the bounding box are outside of the boundaries of the original map. In other words, maps can expand in either dimension. The current behavior for submap is to clip the boundaries within the existing bounds of the map, i.e. maps cannot expand, they can only shrink.

This also adds a keyword `missing` (similar to that in `rotate`) that is used to fill in any of the expanded pixels.

This PR also fixes an inconsistency in `_parse_submap_input` where the order of the corners returned was not consistent over return types. See #6706.

## TODOs

- [ ] Discuss whether this is a good idea. This is a fairly significant change in behavior for `submap`.
- [ ] Understand why current tests are failing
- [ ] Add tests (including updates to existing ones)
- [ ] Changelog
- [ ] Update docstring to make it clear submap allows for this.
- [ ] Example?

## Example

As an example, the following code snippet creates a new map with the bottom left corner at the center of the original map and a width and height that expands far outside the original FOV,

```python
import sunpy.data.sample
import sunpy.map
import astropy.units as u

m = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)
m_expand = m.submap(m.wcs.world_to_pixel(m.center)*u.pix, 
                                         width=1000*u.pix,
                                         height=1000*u.pix,
                                         missing=m.data.min())
m_expand.peek()
```
image

Similarly, you can expand in all directions,

```python
margin = 1000 * u.pix
m_expand = m.submap(u.Quantity([-margin,-margin]),
                    top_right=u.Quantity([m.dimensions.x-1*u.pix+margin, m.dimensions.y-1*u.pix+margin]),
                    missing=m.data.min())
m_expand.peek()
```
image

Or just in one

```python
m_expand = m.submap(m.wcs.world_to_pixel(m.center)*u.pix,
                    width=1000*u.pix,
                    height=200*u.pix,
                    missing=m.data.min())
m_expand.peek()
```
image

Note that this also has the "fun" consequence that, for a map with even a slightly non-diagonal PCij (e.g. a level 1 AIA image), cropping the image to its bottom left and top right *world* coord results in an expanded map (see #4896),

```python
>>> m_new = m.submap(m.bottom_left_coord, top_right=m.top_right_coord)
>>> print(m_new.dimensions)
PixelPair(x=<Quantity 1028. pix>, y=<Quantity 1024. pix>)
>>> print(m.dimensions)
PixelPair(x=<Quantity 1024. pix>, y=<Quantity 1024. pix>)
```

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1768,7 +1768,8 @@ class GenericMap(NDData):
         return new_map
 
     @u.quantity_input
-    def submap(self, bottom_left, *, top_right=None, width: (u.deg, u.pix) = None, height: (u.deg, u.pix) = None):
+    def submap(self, bottom_left, *, top_right=None, width: (u.deg, u.pix) = None, height: (u.deg, u.pix) = None,
+               missing=0.0):
         """
         Returns a submap defined by a rectangle.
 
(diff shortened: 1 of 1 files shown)


alasdairwilson:
Is there an example of a use case?

If this is controlled by a keyword that defaults to False, so is not a breaking change, then this conceptually seems fine to me.

nabobalis:
I would like to use this in aiapy to pad de-rotated level 1.5 images to be the full 4K by 4K.

dstansby:
At a high level I'm 👍 to this new functionality, but 👎 to putting it under `submap`. "sub" implies that you're getting less of the map back, not more. Now, having said that I don't have a better suggestion 😄 for a name of a method that both crops and expands a map, so perhaps a good way to go would be leave `submap` alone, and have a new `expandmap` or similar method for doing the expanding?

wtbarnes:
I think I've decided that this is not a good idea so I'm closing this. If you really want this kind of functionality, you should just reproject to the appropriate FOV. If someone disagrees and wants to revive this effort, feel free to reopen.

Issue #6709, opened 2023-01-17
failed in reading HMI magnetogram in J2k format 


sageyu123:
### Describe the bug

Python version 3.8.10 to 3.8.16. SunPy version 4.1.0

When using `sunpy.map.Map` to read hmi magnetogram image in j2k format downloaded from helioviewer, the following error is given.

```python
In [9]: sdomap = sunpy.map.Map('2023_01_17__15_49_52_300__SDO_HMI_HMI_magnetogram.jp2')
~/.pyenv/versions/3.8.10/lib/python3.8/site-packages/sunpy/map/mapbase.py:584: RuntimeWarning: overflow encountered in scalar negative
  self.plot_settings['norm'].vmin = -threshold
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[9], line 1
----> 1 sdomap = smap.Map(sdofile)

File ~/.pyenv/versions/3.8.10/lib/python3.8/site-packages/sunpy/map/map_factory.py:331, in MapFactory.__call__(self, composite, sequence, silence_errors, *args, **kwargs)
    328 meta = MetaDict(header)
    330 try:
--> 331     new_map = self._check_registered_widgets(data, meta, **kwargs)
    332     new_maps.append(new_map)
    333 except (NoMatchError, MultipleMatchError,
    334         ValidationFunctionError, MapMetaValidationError) as e:

File ~/.pyenv/versions/3.8.10/lib/python3.8/site-packages/sunpy/map/map_factory.py:381, in MapFactory._check_registered_widgets(self, data, meta, **kwargs)
    378 # Only one is found
    379 WidgetType = candidate_widget_types[0]
--> 381 return WidgetType(data, meta, **kwargs)

File ~/.pyenv/versions/3.8.10/lib/python3.8/site-packages/sunpy/map/sources/sdo.py:113, in HMIMap.__init__(self, data, header, **kwargs)
    110 super().__init__(data, header, **kwargs)
    111 if self.unit is not None and self.unit.is_equivalent(u.T):
    112     # Magnetic field maps, not intensity maps
--> 113     self._set_symmetric_vmin_vmax()
    114 self._nickname = self.detector

File ~/.pyenv/versions/3.8.10/lib/python3.8/site-packages/sunpy/map/mapbase.py:584, in GenericMap._set_symmetric_vmin_vmax(self)
    581 def _set_symmetric_vmin_vmax(self):
    582     # Set symmetric vmin/vmax about zero
[...]

PR #6710, opened 2023-01-17
Fix init issue for JP2 HMI images from helioveiwer

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/6709

Does not scale the images correctly:

Figure_1


alasdairwilson (approved):
Approved (with changelog) I can't see the need for unit test but someone else might.

dstansby (dismissed):
A figure test with a jp2 image would be worth adding here, to make sure we keep the scaling correct in the future.

nabobalis:
> A figure test with a jp2 image would be worth adding here, to make sure we keep the scaling correct in the future.

I have added a test and a figure test: The hash will need to be updated due to my inability to generate it on WSL2. 

The big problem is that I don't know how to create a test JP2 file, so its just the file direct from Helioviewer which is 1MB in side, similar to our AIA JP2 file.

I don't think we should add a 1MB file. So I guess I have to see if I can downsample the JP2 File?

nabobalis:
TODO: Rebase and check the hashes are ok.

dstansby:
Yeah, not convinced we should add a 1MB file - how hard would it be to downsample this to something smaller but still reasonable (maybe 256squared?)

nabobalis:
Resizes images and added the code I used to do it. 

I imagine some CI will break at least for any figure tests.

wtbarnes (approved):

nabobalis:
Manually backported to 4.0

PR #6711, opened 2023-01-18
Raise error if downloader for cache fails

nabobalis:
Noticed in aiapy, if the cache/data manager fails to download a file.

It does not raise the error that happened, and continues as if the file is there. Then errors when it goes to open the file. 

I think this code was before parfive returned the errors and used to raise exceptions.

### DIFF
--- a/sunpy/data/data_manager/downloader.py
+++ b/sunpy/data/data_manager/downloader.py
@@ -48,6 +48,8 @@ class ParfiveDownloader(DownloaderBase):
         directory = path.parent
         downloader.enqueue_file(url, directory, filename)
         try:
-            downloader.download()
+            output = downloader.download()
         except Exception as e:
             raise DownloaderError from e
+        if output.errors:
+            raise DownloaderError(output.errors[0].exception)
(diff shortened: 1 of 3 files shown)


sunpy/data/data_manager/downloader.py
```diff
@@ -48,6 +48,8 @@ def download(self, url, path):
         directory = path.parent
         downloader.enqueue_file(url, directory, filename)
         try:
-            downloader.download()
+            output = downloader.download()
         except Exception as e:
             raise DownloaderError from e
+        if output.errors:
+            raise DownloaderError(output.errors[0].exception)
```
wtbarnes, line 55: Why only the first one?

wtbarnes (approved):

sunpy/data/data_manager/downloader.py
nabobalis: This function should only acquire one file, the return of errors is a list, so I want to escape it here.

dstansby (changes requested):
Would be good to add a test. Either a 'real' URL that doesn't exist, or by mocking the `parfive` downloader.

nabobalis:
> Would be good to add a test. Either a 'real' URL that doesn't exist, or by mocking the `parfive` downloader.

Test has been added.

dstansby (approved):

Issue #6126, opened 2022-05-05
Document that `cmap` and `norm` can be passed to `MapSequence.plot`


wtbarnes:
In #5889, I added the ability to explicitly pass `cmap` and `norm` to `MapSequence.plot`. However, I forgot to document this. It should be explicitly stated in the docstring for `plot` that you can do this and thus override the values in `plot_settings`.

nabobalis:
> Id like to handle this @nabobalis

Sure.

PR #6720, opened 2023-01-20
added docstring

ryuusama09:

## 🚨 IMPORTANT 🚨

Thank you.

## PR Description
added documentation for ``norm`` and ``cmap``

Fixes https://github.com/sunpy/sunpy/issues/6126

### DIFF
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -271,8 +271,12 @@ class MapSequence:
         plot_function : function
             A function to be called as each map is plotted.
             For more information see `sunpy.visualization.animator.MapSequenceAnimator`.
-        norm : can be explicitly used to call mapsequence.plot function
-        cmap : can be explicitly used to call mapsequence.plot function
+        norm : 
+            Normalization used in scaling the plot.
+            This will override the predefined value in ``plot_settings['norm']``.
+        cmap : 
+            Color map to be used in coloring the plot.
+            This will override the predefined value in ``plot_settings['cmap']``.
 
         Returns
         -------
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -346,7 +346,7 @@ class MapSequence:
             ani_data = self.maps
 
         im = ani_data[0].plot(axes=axes, **kwargs)
-      
+
         def updatefig(i, im, annotate, ani_data, removes):
             while removes:
                 removes.pop(0).remove()
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -271,10 +271,10 @@ class MapSequence:
         plot_function : function
             A function to be called as each map is plotted.
             For more information see `sunpy.visualization.animator.MapSequenceAnimator`.
-        norm : 
+        norm : string 
             Normalization used in scaling the plot.
             This will override the predefined value in ``plot_settings['norm']``.
-        cmap : 
+        cmap : string 
             Color map to be used in coloring the plot.
             This will override the predefined value in ``plot_settings['cmap']``.
 
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -274,7 +274,7 @@ class MapSequence:
         norm : astropy.visualization.mpl_normalize.ImageNormalize 
             Normalization used in scaling the plot.
             This will override the predefined value in ``plot_settings['norm']``.
-        cmap : string 
+        cmap : `matplotlib.colors.Colormap` or `str`
             Color map to be used in coloring the plot.
             This will override the predefined value in ``plot_settings['cmap']``.
 
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -271,7 +271,7 @@ class MapSequence:
         plot_function : function
             A function to be called as each map is plotted.
             For more information see `sunpy.visualization.animator.MapSequenceAnimator`.
-        norm : string 
+        norm : astropy.visualization.mpl_normalize.ImageNormalize 
             Normalization used in scaling the plot.
             This will override the predefined value in ``plot_settings['norm']``.
         cmap : string 
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -271,6 +271,8 @@ class MapSequence:
         plot_function : function
             A function to be called as each map is plotted.
             For more information see `sunpy.visualization.animator.MapSequenceAnimator`.
+        norm : can be explicitly used to call mapsequence.plot function
+        cmap : can be explicitly used to call mapsequence.plot function
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


sunpy/map/mapsequence.py
```diff
@@ -271,6 +271,8 @@ def plot(self, axes=None, resample=None, annotate=True,
         plot_function : function
             A function to be called as each map is plotted.
             For more information see `sunpy.visualization.animator.MapSequenceAnimator`.
+        norm : can be explicitly used to call mapsequence.plot function
+        cmap : can be explicitly used to call mapsequence.plot function
```
nabobalis, line 275:
```suggestion
        norm : 
            Normalization used in scaling the plot.
            This will override the predefined value in ``plot_settings['norm']``.
        cmap : 
            Color map to be used in coloring the plot.
            This will override the predefined value in ``plot_settings['cmap']``.
```
The types are missing for both, I don't know what they are off by hand, you will need to find them either via Matplotlib's documentation or get into the code and check their type.

sunpy/map/mapsequence.py
```diff
@@ -271,6 +271,12 @@ def plot(self, axes=None, resample=None, annotate=True,
         plot_function : function
             A function to be called as each map is plotted.
             For more information see `sunpy.visualization.animator.MapSequenceAnimator`.
+        norm :
```
nabobalis, line 274: The type is missing, that needs to be filled in.

sunpy/map/mapsequence.py
```diff
@@ -271,6 +271,12 @@ def plot(self, axes=None, resample=None, annotate=True,
         plot_function : function
             A function to be called as each map is plotted.
             For more information see `sunpy.visualization.animator.MapSequenceAnimator`.
+        norm : 
+            Normalization used in scaling the plot.
+            This will override the predefined value in ``plot_settings['norm']``.
+        cmap :
```
nabobalis, line 277: The type is missing, that needs to be filled in.

nabobalis: I don't know what they are. You will need to work it out by either checking the matplotlib docs or running the code and checking the types manually.

nabobalis:
> sorry but didnt really understand what is `plot_setting` because i couldnt find any documentation for it in `mapsequence.py`

plot_settings is an old plotting API that's kept around for historical reasons.

nabobalis:
Did you mean to close the PR?

nabobalis:
> > Did you mean to close the PR?
> 
> yes ,i will reopen the pr once i successfully find their datatypes , because for now i assumed them to be string

You don't need to close it while you work that out.

nabobalis:
> image
> if this is the colormap we are talking about then it is a regular python object. then apparently i think both ``cmap and norm`` are objects @nabobalis . what do you think about this 

Their type won't be a Python object, notice how most types are not objects in our documentation strings. I would suggest trying to check their type when you run the plot command.

sunpy/map/mapsequence.py
```diff
@@ -271,6 +271,12 @@ def plot(self, axes=None, resample=None, annotate=True,
         plot_function : function
             A function to be called as each map is plotted.
             For more information see `sunpy.visualization.animator.MapSequenceAnimator`.
+        norm : string 
+            Normalization used in scaling the plot.
+            This will override the predefined value in ``plot_settings['norm']``.
+        cmap : string
```
nabobalis, line 277: It will be https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.set_cmap.html

sunpy/map/mapsequence.py
```diff
@@ -271,6 +271,12 @@ def plot(self, axes=None, resample=None, annotate=True,
         plot_function : function
             A function to be called as each map is plotted.
             For more information see `sunpy.visualization.animator.MapSequenceAnimator`.
+        norm : string
```
nabobalis, line 274: It will be https://matplotlib.org/stable/api/cm_api.html#matplotlib.cm.ScalarMappable.set_norm

nabobalis: A Colormap is an object, its type is not.

nabobalis: Notice how they don't say object for the type of the first item in set_norm.

nabobalis: There should be an example in our documentation: https://docs.sunpy.org/en/stable/generated/gallery/index.html

nabobalis: Yes, getting closer. But cmap doesn't have to be a string. I am pretty sure that a norm doesn't have to be just an astropy class either.

nabobalis: > hmm , but a same thing can have more than 1 dtpye ? .

Yes because the function we call to set both items, the colormap and normalization allows two types. As their documentation tells us.

> most of the times norm should be an astropy class because of the way it is used so dont you think this should work ?

It does not have to be an astropy class, therefore saying it can only be one is incorrect.

nabobalis: It is times like this when going to the original function is very useful since otherwise knowing what types are allowed is very tricky.

sunpy/map/mapsequence.py
```diff
@@ -271,6 +271,12 @@ def plot(self, axes=None, resample=None, annotate=True,
         plot_function : function
             A function to be called as each map is plotted.
             For more information see `sunpy.visualization.animator.MapSequenceAnimator`.
+        norm : astropy.visualization.mpl_normalize.ImageNormalize 
+            Normalization used in scaling the plot.
+            This will override the predefined value in ``plot_settings['norm']``.
+        cmap : string
```
nabobalis, line 277:
```suggestion
        cmap : `matplotlib.colors.Colormap` or `str`
```

sunpy/map/mapsequence.py
```diff
@@ -271,6 +271,12 @@ def plot(self, axes=None, resample=None, annotate=True,
         plot_function : function
             A function to be called as each map is plotted.
             For more information see `sunpy.visualization.animator.MapSequenceAnimator`.
+        norm : astropy.visualization.mpl_normalize.ImageNormalize
```
nabobalis, line 274:
```suggestion
        norm : `matplotlib.colors.Normalize` or `str`
```

As https://matplotlib.org/stable/api/cm_api.html#matplotlib.cm.ScalarMappable.set_norm suggests.

I have no idea how a string would work for this one.

nabobalis: You don't need to apologize, you are not on a schedule. 

Changes look good to me.

sunpy/map/mapsequence.py
```diff
@@ -271,6 +271,12 @@ def plot(self, axes=None, resample=None, annotate=True,
...
```
nabobalis, line 274:
```suggestion
        norm : `matplotlib.colors.Normalize` or `str`
```

nabobalis (approved):

Issue #6562, opened 2022-11-03
Turn `sunpy.map.header_helper.meta_keywords()` into a dictionary


ayshih:
Currently, the function `sunpy.map.header_helper.meta_keywords()` does one thing: it returns a (non-copied) dictionary.  If that's all it's going to do, with no apparent need to protect the returned dictionary, it should simply be a dictionary rather than a function.  Why force people to type the parentheses and cause confusion if they accidentally omit them?

dstansby:
~~I guess this way stops users accidentally modifying the dictionary in-place?~~ ha I treid this and it definitely doesn't 😄 . But maybe we could change that by making a copy of the dict in the `meta_keywords()` function?

ayshih:
Yeah, that was my point, but I worded it poorly.  Since the dictionary wasn't being copied, either it doesn't need to be protected and can be exposed directly as a variable, or the lack of a copy is a bug.

Even if the dictionary does need to be protected, I think this should be made into a (now-possible) module-level attribute (that returns a copy) rather than a function that takes no arguments.

wtbarnes:
[PEP 562](https://peps.python.org/pep-0562/) describes module-level attrs which it sounds like may be useful here. I was not aware of these.

ayshih:
> [PEP 562](https://peps.python.org/pep-0562/) describes module-level attrs which it sounds like may be useful here. I was not aware of these.

You say that, but you reviewed my PR for downloading sample data on demand (https://github.com/sunpy/sunpy/pull/6426#pullrequestreview-1132149882), for which module-level attributes are fundamental.  I'm now particularly amused by your request there (https://github.com/sunpy/sunpy/pull/6426#discussion_r988438026) to add a comment pointing to PEP 562 in order to help "future readers", seemingly foreshadowing your own amnesia. =P

ayshih:
_Originally posted by @ayshih in https://github.com/sunpy/sunpy/issues/6722#issuecomment-1404575334_
> Um...  What is the purpose for this dictionary?  Our codebase doesn't appear to use it; does any user?  Can we just remove it?  If it has a purpose, why does it live in `sunpy.map.header_helper`?  All questions I should pose in the original issue, I suppose.

nabobalis:
The only reference I can find is in the map.rst:

[...]

PR #6722, opened 2023-01-20
make meta_keywords a dict

exitflynn:
## PR Description

makes the `sunpy.map.header_helper.meta_keywords()` a dict from the get-go, making it a copy of the `_map_meta_keywords` dict as discussed in #6562

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -61,6 +61,7 @@ For more information or to ask questions about ``sunpy`` or any other SunPy libr
 -  `sunpy documentation <https://docs.sunpy.org/en/stable/>`__
 -  `SunPy Chat`_
 -  `SunPy mailing list <https://groups.google.com/forum/#!forum/sunpy>`__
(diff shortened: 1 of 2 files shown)


exitflynn:
i suppose we'd need a changelog for this right?

nabobalis:
I need to check with @ayshih but this is public API, we can't remove the function without a deprecation cycle.

ayshih:
Um...  What is the purpose for this dictionary?  Our codebase doesn't appear to use it; does any user?  Can we just remove it?  If it has a purpose, why does it live in `sunpy.map.header_helper`?  All questions I should pose in the original issue, I suppose.

nabobalis:
So in that vein, should we just fix the fact it isn't copied and deprecate it? 

We can worry about if we need to keep it around in some form in the future?

exitflynn:
whoops looks like i accidentally closed the PR by force-pushing ☠️🫣

PR #6724, opened 2023-01-21
parallelise cdaweb database updation

### LINKED ISSUE #6519: Parallelise code to update cdaweb datasets

exitflynn:
## PR Description

updates cdaweb databases asynchronously using aiohttp instead of requests. potentially closes #6519.

### DIFF
--- a/sunpy/net/cdaweb/helpers.py
+++ b/sunpy/net/cdaweb/helpers.py
@@ -1,6 +1,7 @@
 import json
 import pathlib
 
+import aiohttp
 import requests
 
 from astropy.table import Table
@@ -116,7 +117,7 @@ def get_datasets(observatory):
     return t
 
 
-def _update_cdaweb_dataset_data():
+async def _update_cdaweb_dataset_data():
     all_obs = get_observatory_groups()
     url = '/'.join([
         _CDAS_BASEURL,
@@ -128,11 +129,13 @@ def _update_cdaweb_dataset_data():
     for group in all_obs['Group']:
         print(f'🛰 Getting datasets for {group}')
         group_url = url + f'?observatoryGroup={group}'
-        response = requests.get(group_url, headers=_CDAS_HEADERS)
+        async with aiohttp.ClientSession() as session:
+            async with session.get(group_url) as response:
+                response = requests.get(group_url, headers=_CDAS_HEADERS)
 
-        datasets = response.json()['DatasetDescription']
-        dataset_ids = {ds['Id']: ds['Label'] for ds in datasets}
-        all_datasets.update(dataset_ids)
+                datasets = response.json()['DatasetDescription']
+                dataset_ids = {ds['Id']: ds['Label'] for ds in datasets}
+                all_datasets.update(dataset_ids)
 
     attr_file = pathlib.Path(__file__).parent / 'data' / 'attrs.json'
     with open(attr_file, 'w') as attrs_file:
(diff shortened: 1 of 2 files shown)


nabobalis:
What's the speed difference now?

exitflynn:
apologies actually, just saw why the original PR ditched parallel downloads, since there's [rate-limiting](https://cdaweb.gsfc.nasa.gov/WebServices/REST/py/cdasws/) and they recommend no more than 4-5 requests at a time. From what i've seen so far, I should still be able to implement a way to limit 4-5 connections at a time per host since there's plenty support for that kind of thing in since it's a common enough usecase. I'll get to it tomorrow and report the time difference then!

exitflynn:
I've been trying to get this working for quite a while now and had been running into a couple of issues. I figured I'd do a quick progress report on 
### what I've managed to implement so far:
1. The requests/downloads during the updation happen concurrently. It takes about 10 minutes to update the data for me now
2. I've also put a [Semaphore](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Semaphore) in [there](https://github.com/exitflynn/sunpy/blob/paralyzed/sunpy/net/cdaweb/helpers.py#L144) which can be used to limit the number of concurrent processes at a moment.

### Problems I've been facing:
- I can't figure out why it seems to go out with a connection lost error by the time it reaches the end of the list. :flushed: 
- Also, despite limiting to 5 connections, some servers error out, and right now the way I'm getting around that is by using a dirty hack which basically involves waiting for 5 secs and trying the same thing again, hoping some request gets completed in the meanwhile and decreases the load there. :clown_face: 
- Aand currently what i've done is limit <=5 connections _from_ a client whereas it should be limited to <=5 connections _to_ a particular server so as to not put too much load there (no problem with >5 connections from the user's client). I'll have to look into how one would implement that. :thinking: 


To test these changes, you'll only need to run `tools/update_attrs_json.py`. :grimacing: :innocent: 

[There are some helpful print statements in there for debugging purposes which I'll remove later on. I also removed updating the VSO and JSOC datasets for the duration of the PR since we're concerned with changes in CDAweb database alone for now (Though this could be implemented in the other databases too if this is something that we end up going out)]
In case anyone has any suggestions about any improvements I can make to my approach here that'd be great!

exitflynn:
I've been trying to see if i can make this (and #6779) work. Health's been acting up a bit as of late. Hence, not much progress, just felt like putting out a quick update on this in that regard 🤠

nabobalis:
> I've been trying to see if i can make this (and #6779) work. Health's been acting up a bit as of late. Hence, not much progress, just felt like putting out a quick update on this in that regard cowboy_hat_face

Your health is more important, I would focus on that first and foremost.

dstansby:
👋 @exitflynn I'm afraid I'm just merged an older PR to fix this issue, so I'll close this one.

exitflynn:
I see, @dstansby. I wasn't able to get much work here done as well due to some health issues as of late, glad to see an improvement is in place!

exitflynn:
I see, @dstansby. I wasn't able to get much work here done as well since due to some health issues and exmas as of late, glad to see an improvement is in place!

PR #6725, opened 2023-01-24
add unit detection for XRT data

### LINKED ISSUE #6716: XRT Maps do not include unit information
XRT maps, from any level (as far as I can tell), do not include a `BUNIT` keyword or any equivalent keyword to encode the unit information.

However, according to Table 1.1 in [the XRT analysis guide](https://xrt.cfa.harvard.edu/resources/documents/XAG/XAG.pdf), the units for the different levels are DN for level 0 and DN/s for level 1. However, there is also a note that level 1 can be DN or DN/s, depending on which options are passed to `xrt_prep`. When level 1 data is divided by the exposure time (to be put in units of DN/s), there is a tag added to the `HISTORY` that indicates that the data have been renormalized. When pulling in the data, the source could check for this tag and set the unit appropriately and otherwise fall back to DN.

exitflynn:
## PR Description

XRT data is now given the `unit` property which is `DN` by default and `DN/sec` if it's been renormalized. Addresses and closes #6716


exitflynn:
Is an outcome like this what we're aiming for?
image
Looks alright to me just that the unit doesn't show up among other properties...

sunpy/map/sources/hinode.py
```diff
@@ -98,6 +98,15 @@ def processing_level(self):
             return
         return int(lvl)
 
+    @property
+    def unit(self):
+        unit = "DN"
+        response = self.meta.get('HISTORY')
+        for i in response:
+            if i[:15] == "XRT_RENORMALIZE":
+                unit = "DN/sec"
```
nabobalis, line 107: Does this not have to be an astropy unit or does a string work?

sunpy/map/sources/hinode.py
```diff
@@ -98,6 +98,15 @@ def processing_level(self):
             return
         return int(lvl)
 
+    @property
+    def unit(self):
+        unit = "DN"
```
nabobalis, line 103: Are we sure we can set this to DN?

nabobalis:
We should add the reference information as to how we know what the unit is in the unit code.

sunpy/map/sources/hinode.py
exitflynn: yes totally, thanks for pointing out

sunpy/map/sources/hinode.py
exitflynn: that was my understanding of the issue but I'm not sure. I would like @wtbarnes to verify. I tried going through the XRT analysis guide, and I think I'm missing accounting for level 2 data?

nabobalis: Where is this guide?

exitflynn: it was in the original issue (#6716) description, here's the [link](https://xrt.cfa.harvard.edu/resources/documents/XAG/XAG.pdf), table 1.1. Will also described things pretty well in the issue description.

nabobalis: I see it, I think we would check for the unit in the metadata, and if non, set it to DN, and also check the history tag to see if it should be DN/S. 

A reference to that document and the table number would be good here as well. If it isn't already in the main doc string.

exitflynn: I had a better look at the document and Will's responses in the original issue, and i think it'd be better to mention section 2.11 of the XRT Analysis Guide, I'll do that :+1:

exitflynn:
does the message work? The XRT Analysis Guide is already (indirectly) linked to in the main doc string

sunpy/map/sources/hinode.py
```diff
@@ -98,6 +98,20 @@ def processing_level(self):
...
         return int(lvl)
 
+    @property
+    def unit(self):
+        # XRT data values are in DN and are converted into DN/s if the data has been normalized.
+        # A tag starting with "XRT_RENORMALIZE" is added to the HISTORY tag in that case.
+        # Ref. Section 2.11 of the XRT Analysis Guide.
+        unit = self.meta.get('BUNIT')
+        if not unit:
+            unit = u.astrophys.dn
```
nabobalis, line 108: I forgot now, I will need to check with the others, but we don't use DN but counts as a result that we can't save out Maps with DN units.

sunpy/map/sources/hinode.py
```diff
@@ -98,6 +98,20 @@ def processing_level(self):
...
+            hvalues = self.meta.get('HISTORY')
+            for value in hvalues:
```
nabobalis, line 110: Do we have to look over them?

nabobalis:
Reads fine to me. It will need a bugfix changelog.

sunpy/map/sources/hinode.py
wtbarnes, line 108:
```suggestion
            unit = u.ct
```

sunpy/map/sources/hinode.py
```diff
@@ -98,6 +98,20 @@ def processing_level(self):
...
+                if "XRT_RENORMALIZE" in value:
+                    unit = u.astrophys.dn / u.second
```
wtbarnes, line 112: Once #6728 is merged, this can become,

```suggestion
            if "XRT_RENORMALIZE" in self.meta.get('HISTORY')
                    unit = u.ct / u.second
```

sunpy/map/sources/hinode.py
```diff
@@ -98,6 +98,20 @@ def processing_level(self):
             return
         return int(lvl)
 
+    @property
+    def unit(self):
+        # XRT data values are in DN and are converted into DN/s if the data has been normalized.
+        # A tag starting with "XRT_RENORMALIZE" is added to the HISTORY tag in that case.
+        # Ref. Section 2.11 of the XRT Analysis Guide.
```
wtbarnes, line 105:
```suggestion
        # See Table 1.1 and Section 2.11 of the XRT Analysis Guide.
```

sunpy/map/sources/hinode.py
```diff
@@ -98,6 +98,20 @@ def processing_level(self):
...
+        unit = self.meta.get('BUNIT')
```
wtbarnes, line 106: We do a bit more parsing of the unit information in the base class `unit` property.
```suggestion
        unit = super().unit
```

wtbarnes (changes requested):
This will have to wait on #6728. This should also get a test for the unit property.

sunpy/map/sources/hinode.py
exitflynn: ooh, I see

exitflynn:
i've made the changes, should I get to writing a test for it now?

nabobalis:
Yes please.

exitflynn:
there's currently only a header file in the examples, can you link to one?

nabobalis:
> there's currently only a header file in the examples, can you link to one?

You mean to a file that has the correct history text?

exitflynn:
yes, i'm unsure of where I can get a file to test on since as of now we have a header file and we do something like a `get_dummy_map_from_header`. We *could* test based on that as well though...

nabobalis:
> yes, i'm unsure of where I can get a file to test on since as of now we have an XRT header file and we do a `get_dummy_map_from_header` for testing. We _could_ test based on this alone though...

Why would we not use that test header?

exitflynn:
oh no i was mistaken, i confused things with some stuff i did while testing locally :sweat_smile:. yes we can do just that :+1:

sunpy/map/sources/hinode.py
exitflynn: hi will, is `u.ct` here the same as `DN`? :eyes:

nabobalis: No but there is a technical reason why we use ct instead of DN.

exitflynn:
(i added the test but it's showing up a few hours in the past because i forgot to fix my system time properly). Also, i think the PR should be complete now!

exitflynn:
i wonder if it should be of concern that units doesn't show up here: 
image

Is this something I should look into?

nabobalis:
Should it?

wtbarnes:
The text summary does not currently include information about the unit. Maybe it should but that should be a separate PR.

changelog/6725.bugfix.rst
```diff
@@ -0,0 +1 @@
+XRT Maps now detect and include the unit property.
```
nabobalis, line 1:
```suggestion
`.XRTMap` now will set the units for level 0 and level 1 files correctly.
```

wtbarnes: This unit checking should work for any processing level, not just 0 or 1. Also, I don't think that intersphinx link is right.

exitflynn:
is a changelog message like this one fine? :grimacing: :eyes:

changelog/6725.bugfix.rst
```diff
@@ -0,0 +1 @@
+`~sunpy.map.sources.XRTMap`` will now set the units for all files correctly.
```
nabobalis, line 1:
```suggestion
`~sunpy.map.sources.XRTMap` will now set the units for all files correctly.
```

changelog/6725.bugfix.rst
```diff
@@ -0,0 +1 @@
+`~sunpy.map.sources.XRTMap` will now set the units for all files correctly.
```
nabobalis, line 1:
```suggestion
`~sunpy.map.sources.XRTMap` will now set the unit for all files correctly.
```

exitflynn:
should be good to go now, i think

nabobalis:
It will need a review before but yes I think its ready to go.

sunpy/map/sources/hinode.py
```diff
@@ -98,6 +98,18 @@ def processing_level(self):
...
 
+    @property
+    def unit(self):
+        # XRT data values are in DN and are converted into DN/s if the data has been normalized.
+        # A tag starting with "XRT_RENORMALIZE" is added to the HISTORY tag in that case.
+        # See Table 1.1 and Section 2.11 of the XRT Analysis Guide.
+        unit = super().unit
+        if not unit:
+            unit = u.ct
+            if "XRT_RENORMALIZE" in self.meta.get('HISTORY'):
```
wtbarnes, line 109:
```suggestion
            history = self.meta.get('HISTORY', '')
            if "xrt_renormalize" in history.lower():
```

wtbarnes: To protect against any capitalization issues. Sorry, should've caught that in my previous suggestion.

sunpy/map/sources/tests/test_xrt_source.py
```diff
@@ -38,6 +38,11 @@ def test_measurement(xrt_map):
     assert xrt_map.measurement == 'Be thin-Open'
 
 
+def unit(xrt_map):
```
wtbarnes, line 41:
```suggestion
def test_unit(xrt_map):
```

wtbarnes: Without the `test_` prefix on the name, I don't think pytest will run this so I don't think this test has actually been run yet...

changelog/6725.bugfix.rst
```diff
@@ -0,0 +1 @@
+`~sunpy.map.sources.XRTMap` will now set the unit for all files correctly.
```
wtbarnes, line 1:
```suggestion
`~sunpy.map.sources.XRTMap` will now set the unit for XRT files if the ``BUNIT`` key is missing.
```

wtbarnes (approved):
Sorry, just a few last things that I should've caught previously.

wtbarnes (dismissed):
Sorry, I meant to request changes, not approve.

sunpy/map/sources/tests/test_xrt_source.py
exitflynn: yes, good catch! :cowboy_hat_face:

exitflynn:
thanks for catching these!

nabobalis (approved):

nabobalis:
Since Will's comments are addressed, I will dismiss and merge this PR.

nabobalis:
Thanks again @exitflynn

exitflynn:
Thank you so much for the reviews @nabobalis and @wtbarnes! :grin:

PR #6728, opened 2023-01-25
Fix `HISTORY` key in test maps created from header files

wtbarnes:
This PR fixes a small bug in the dummy test map creation where, because the maps were created directly from astropy header objects, the HISTORY (and COMMENTS) keys were not being formatted as a single string.

This was discovered in #6725 when checking the HISTORY field for a specific key.

This should have no impact on any of the existing tests.

### DIFF
--- a/sunpy/data/test/__init__.py
+++ b/sunpy/data/test/__init__.py
@@ -13,6 +13,7 @@ import astropy.io.fits
 from astropy.utils.data import get_pkg_data_filename
 
 import sunpy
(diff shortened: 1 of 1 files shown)


nabobalis:
So far the tests are ok.

nabobalis (approved):

wtbarnes:
The figure test failure is unrelated

nabobalis:
Online fails are the same in my eyes.

PR #6731, opened 2023-01-26
fig-devdeps : Have to add a unit to lon.coord_wrap

nabobalis:
Upstream astropy issue was merged https://github.com/astropy/astropy/pull/14050

This should fail again in future as we have 2 examples that make a similar call but we will need to wait for the release of astropy 5.3.

### DIFF
--- a/sunpy/visualization/wcsaxes_compat.py
+++ b/sunpy/visualization/wcsaxes_compat.py
@@ -2,8 +2,10 @@
 This module provides functions to make WCSAxes work in SunPy.
 """
 import matplotlib.pyplot as plt
+from packaging.version import Version
 
 import astropy.units as u
+from astropy import __version__ as astropy_version
 from astropy.visualization import wcsaxes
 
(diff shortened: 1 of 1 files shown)


sunpy/visualization/wcsaxes_compat.py
```diff
@@ -172,7 +172,11 @@ def wcsaxes_heliographic_overlay(axes, grid_spacing: u.deg = 10*u.deg, annotate=
     lon = overlay[0]
     lat = overlay[1]
 
-    lon.coord_wrap = 180
+    from astropy import __version__
+    if __version__ >= '5.3':
```
nabobalis, line 176: I can parse this properly if people prefer.

dstansby (changes requested):
Yeah, would be good to parse the version properley!

nabobalis:
> Yeah, would be good to parse the version properley!

Should be ok now.

Issue #6695, opened 2023-01-05
Add support sqlalchemy 2.0


nabobalis:
The database module uses sqlalchemy and there is a major release coming soon, 2.0. 

This is going to break something in database so we pinned the version for now.
But we should investigate how much work it will be to add support for version 2.0.

_Originally posted by @alasdairwilson in https://github.com/sunpy/sunpy/issues/6694#issuecomment-1372853864_

ConorMacBride:
Might just need to change the imports (https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#declarative-becomes-a-first-class-api) depending on the installed version.

nabobalis:
Fixed by https://github.com/sunpy/sunpy/pull/6749

Issue #6647, opened 2022-12-05
TST: sqlalchemy.exc.InvalidRequestError DatabaseEntry has been deleted


pllim:
### Describe the bug

```
_______________________________________________________________________________ test_clear_database ________________________________________________________________________________

filled_database = <Table length=1>
 id  observation_time_start observation_time_end ... starred download_time size
str1          str3   ...-------- ... ------- ------------- ----
   4                    N/A                  N/A ...      No           N/A  N/A

    def test_clear_database(filled_database):
        assert len(filled_database) == 10
        filled_database.clear()
        assert not filled_database
        assert filled_database.session.query(JSONDump).all() == []
        assert filled_database.session.query(FitsHeaderEntry).all() == []
        assert filled_database.session.query(FitsKeyComment).all() == []
        assert filled_database.session.query(Tag).all() == []
>       filled_database.undo()

.tox/py310-sunpy/lib/python3.10/site-packages/sunpy/database/tests/test_database.py:731: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.tox/py310-sunpy/lib/python3.10/site-packages/sunpy/database/database.py:1087: in undo
    self._command_manager.undo(n)
.tox/py310-sunpy/lib/python3.10/site-packages/sunpy/database/commands.py:359: in undo
    command.undo()
.tox/py310-sunpy/lib/python3.10/site-packages/sunpy/database/commands.py:98: in undo
    operation.undo()
.tox/py310-sunpy/lib/python3.10/site-packages/sunpy/database/commands.py:161: in undo
    self.session.add(self.entry)
.tox/py310-sunpy/lib/python3.10/site-packages/sqlalchemy/orm/scoping.py:361: in add
    return self._proxied.add(instance, _warn=_warn)
.tox/py310-sunpy/lib/python3.10/site-packages/sqlalchemy/orm/session.py:3137: in add
    self._save_or_update_state(state)
.tox/py310-sunpy/lib/python3.10/site-packages/sqlalchemy/orm/session.py:3167: in _save_or_update_state
[...]

PR #6735, opened 2023-01-27
Unpin and update database for sqla 2.0

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/6695
Fixes https://github.com/sunpy/sunpy/issues/6647

TODO:
- [ ] Remove changelog that mentions pinning sql
- [ ] Fix unittest.


alasdairwilson:
So the failed test was a massive pain to track down the exact root cause. `Database.clear()` generates a bunch of commands on a `CompositeCommand` which are ordered by the type of thing removed (first tags from entries then tags then entries) for some reason adding them back in via undo, in this same order, worked in <2.0 and now not in 2.0. In 2.0 it results in tags being added to id's that do not exist, creating entries with those id's which are then invalid when you try to write the `DatabaseEntry` with the same id back into the database. The whole `make_transient()` error is a red herring in my opinion because actually the sunpy code was correctly setting the` _sa_instance_state.deleted` flag to False before restoring these entries.

I tried for a long time to figure out the cause rooted in the difference between 2.0 and <2.0 SQLa for an elegant solution but I could not but the basic solution is to either sort the list of operations in the Composite operation but I found this, in any non manual ordering, to normally result in the SQL database violating unique constrains as id's are automatically assigned prior to the correct DBentry being readded to the table. The alternative, which I have done at  is to specifically search the list of _operations in a `CompositeOperation.undo` for `RemoveEntry `where `RemoveEntry `is a `DatabaseEntry `and do that First and then undo all other RemoveEntry's (e.g. Tag) and finally undo all other operations (e.g. `RemoveTag`).

I actually dunno the best way to push to this branch Can either merge it at your repo or replace this pull with my branch:

https://github.com/alasdairwilson/sunpy/tree/sql

Tests pass:

https://github.com/alasdairwilson/sunpy/actions/runs/4044537889/jobs/6954812107

nabobalis:
You can create a new PR and I can close this?

nabobalis:
Replaced vby https://github.com/sunpy/sunpy/pull/6749

Issue #6655, opened 2022-12-09
Creating a map from a .fz file causes an error 


samaloney:
### Describe the bug
Originally raised by nawinnga in the sunpy chat room.

Opening and GONG H-alpha fits obtained from Fido causes an error, it tires to read it as an ana file.

### To Reproduce
```
import sunpy
sunpy.map.Map('ftp://gong2.nso.edu/HA/haf/202203/20220318/20220318000050Bh.fits.fz')

ck_synch_hd: error: file does not have the F0 synch pattern (found 0x504d4953 instead)
ana_fzread: error: ck_sync_hd error!
Traceback (most recent call last):
  File "/Users/shane/.virtualenvs/temp/lib/python3.7/site-packages/sunpy/map/map_factory.py", line 161, in _read_file
    pairs = read_file(os.fspath(fname), **kwargs)
  File "/Users/shane/.virtualenvs/temp/lib/python3.7/site-packages/sunpy/io/file_tools.py", line 89, in read_file
    return _readers[readername].read(filepath, **kwargs)
  File "/Users/shane/.virtualenvs/temp/lib/python3.7/site-packages/sunpy/io/ana.py", line 53, in read
    data = _pyana.fzread(filename, debug)
ValueError: In pyana_fzread: could not read ana file, data returned is NULL.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/shane/.virtualenvs/temp/lib/python3.7/site-packages/sunpy/map/map_factory.py", line 318, in __call__
    data_header_pairs = self._parse_args(*args, silence_errors=silence_errors, **kwargs)
  File "/Users/shane/.virtualenvs/temp/lib/python3.7/site-packages/sunpy/map/map_factory.py", line 244, in _parse_args
    data_header_pairs += self._parse_arg(arg, **kwargs)
  File "/Users/shane/.virtualenvs/temp/lib/python3.7/site-packages/sunpy/util/functools.py", line 18, in wrapper
    return dispatcher.dispatch(args[1].__class__)(*args, **kwargs)
  File "/Users/shane/.virtualenvs/temp/lib/python3.7/site-packages/sunpy/map/map_factory.py", line 285, in _parse_url
    pairs = self._read_file(path, **kwargs)
  File "/Users/shane/.virtualenvs/temp/lib/python3.7/site-packages/sunpy/map/map_factory.py", line 164, in _read_file
    raise IOError(msg) from e
[...]

PR #6736, opened 2023-01-27
Reverse file detection logic fall to to extension

samaloney:
## PR Description

This PR reverses the file detection logic so first try to detect the filetype based on the content and then fallback to using extension. I could envisage some performance issues compared the current implementation but this should probably be more robust.

This partially addresses #6655

### DIFF
--- a/sunpy/io/_file_tools.py
+++ b/sunpy/io/_file_tools.py
@@ -81,14 +81,21 @@ def read_file(filepath, filetype=None, **kwargs):
     if filetype is not None:
         return _readers[filetype].read(filepath, **kwargs)
 
-    # Go through the known extensions
-    for extension, readername in _known_extensions.items():
-        if filepath.endswith(extension) or filetype in extension:
-            return _readers[readername].read(filepath, **kwargs)
-
     # If filetype is not apparent from the extension, attempt to detect it
-    readername = _detect_filetype(filepath)
-    return _readers[readername].read(filepath, **kwargs)
+    try:
+        readername = _detect_filetype(filepath)
+    except UnrecognizedFileTypeError:
+        readername = None
+
+    # Go through the known extensions
+    for extension, name in _known_extensions.items():
+        if filepath.endswith(extension) or filetype in extension:
+            readername = name
+
+    if readername is not None:
+        return _readers[readername].read(filepath, **kwargs)
+
+    raise UnrecognizedFileTypeError("The requested filetype is not currently supported by sunpy.")
 
 
 def read_file_header(filepath, filetype=None, **kwargs):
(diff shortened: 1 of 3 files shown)


samaloney:
I can tidy this up if folks think the idea is good

nabobalis:
I think it makes more sense this way.

nabobalis:
@samaloney I reworked this slight, I add to add a None return for a URL.

I think there must be a better way like if we download it in a temporary location.

sunpy/io/file_tools.py
```diff
@@ -198,18 +163,19 @@ def detect_filetype(filepath):
     filetype : `str`
         The type of file.
     """
+    if "http" in filepath or "ftp" in filepath:
```
dstansby, line 166: This seems like a bad idea, what if I have a file at `/home/me/http/mymap.fits`? It would return `None` here because it has "http" in the filepath!

sunpy/io/tests/test_filetools.py
```diff
@@ -60,9 +59,39 @@ def test_read_file_fits_gzip():
         assert np.all(pair[0][0] == np.tile(np.arange(32), (32, 1)).transpose())
 
 
+@pytest.mark.parametrize('fname', ['aia_171_image.fits',
+                                   pathlib.Path('aia_171_image.fits')])
+def test_write_file_fits(fname, tmpdir):
```
dstansby, line 64: Could you add some short one line comments to the top of these test functions explaining what they're testing?

sunpy/io/file_tools.py
```diff
@@ -45,14 +44,31 @@ def __getitem__(self, key):
...
 
-# Map the readers
 _readers = Readers({
     'fits': fits,
     'jp2': jp2,
     'ana': ana
 })
 
 
+def _read(filepath, function_name, filetype=None, **kwargs):
```
dstansby, line 54: This needs a docstring with a short description and a listing of the arguments and their types.

dstansby (changes requested):
@samaloney do you still want to get this in? If so it needs
1. a changelog entry
2. A test to make sure it works. Something similar to the original issue (writing a `.fits` file with a `.fits.gz` extension then trying to read it back in) should work

 I've left some other requests for extra clarification inline too

sunpy/io/file_tools.py
samaloney: Yea very true is guess `startswith('http://')` and `startswith('http://')` would be better but still not fool proof.

sunpy/io/_file_tools.py
```diff
@@ -259,9 +250,3 @@ class ReaderError(ImportError):
     """
     Exception to raise when a reader errors.
     """
-
-
-class InvalidJPEG2000FileExtension(OSError):
```
nabobalis, line 264: This was never used?

nabobalis (approved):

Cadair (approved):

PR #6741, opened 2023-01-27
Bump min Python to 3.9

dstansby:
According to NEP 29, Python 3.8 support drops on Apr 14, 2023. Since sunpy 5.0 release date is 2023-05-06, drop Python 3.8 support.


dstansby:
I've stuck this in https://github.com/sunpy/sunpy/pull/6742

PR #6742, opened 2023-01-27
Bump minimum deps

dstansby:
Since sunpy 5.0 release date is 2023-05-06
- bump minimum astropy version to 5.0, which was released on Nov 19, 2021.
- bump minimum numpy version to 1.21, allowed from Apr 14, 2023 according to NEP29


ConorMacBride (commented):
Should we make the core job run on py311? The HDF5 library dependency in the publish job may no longer be needed as h5py now has cp311 binaries: https://pypi.org/project/h5py/#files

nabobalis:
> Should we make the core job run on py311? The HDF5 library dependency in the publish job may no longer be needed as h5py now has cp311 binaries: [pypi.org/project/h5py/#files](https://pypi.org/project/h5py/#files)

Yeah and we should remove the library install if we don't need it anymore.

dstansby:
I think all the CI failures are un-releated, and being fixed by other open PRs.

nabobalis:
Needs a whats new entry before it should be merged.

.github/workflows/ci.yml
```diff
@@ -146,13 +141,9 @@ jobs:
...
-        - cp3{8,9,10}-macosx_arm64
-        # Split these out for more rapid debugging
-        - cp311-manylinux*_x86_64
-        - cp311-macosx_x86_64
-        - cp311-macosx_arm64
+        - cp3{9,10,11}-manylinux*_x86_64
+        - cp3{9,10,11}-macosx_x86_64
+        - cp3{9,10,11}-macosx_arm64
       # This can be removed when there are binaries for h5py on 3.11
       libraries: libhdf5-dev
```
nabobalis, line 148: Do we need this still?

dstansby: Ah thanks for catching, there was one more instance too and I've removed both.

dstansby:
👍 what's new added

nabobalis (approved):

nabobalis:
Looks like scipy tried to build a wheel on the oldest deps. I assume we should bump that version too?

Issue #5380, opened 2021-06-02
NRT GOES data functionality?


hayesla:
### Description

The NRT GOES 1-minute space weather data is available on the SWPC page as a [json file](https://services.swpc.noaa.gov/json/goes/primary/) updated every minute (I think) for data for up to the past 7 days. It could be useful for some to have a GOES NRT client to get the data and to have TimeSeries functionality.

AthKouloumvakos:
#4346

nabobalis:
> @nabobalis , the issue is pretty old and uncatered , if this issue is still relevant , are there any plans to implement this ?

I assume this issue is still relevant, @hayesla would you agree?

I don't think anyone has started to implement this.

nabobalis:
> till would you want to suggest me any significant issue that needs to be tackled ?

Honestly, I can't think of one, I would suggest just browsing through the issues to see what catches your eye.

nabobalis:
> @nabobalis , how do we want to structure this api as (its params , how will it produce results etc).

You mean how do we get the data and use that in sunpy? If so, https://docs.sunpy.org/en/latest/dev_guide/contents/extending_fido.html should hopefully cover most of the steps.

nabobalis:
> > > @nabobalis , how do we want to structure this api as (its params , how will it produce results etc).
> > 
> > 
> > You mean how do we get the data and use that in sunpy? If so, [docs.sunpy.org/en/latest/dev_guide/contents/extending_fido.html](https://docs.sunpy.org/en/latest/dev_guide/contents/extending_fido.html) should hopefully cover most of the steps.
> 
> so this will be a "scraper" client i suppose ?

Yeah.

nabobalis:
> > > > > @nabobalis , how do we want to structure this api as (its params , how will it produce results etc).
> > > > 
> > > > 
> > > > You mean how do we get the data and use that in sunpy? If so, [docs.sunpy.org/en/latest/dev_guide/contents/extending_fido.html](https://docs.sunpy.org/en/latest/dev_guide/contents/extending_fido.html) should hopefully cover most of the steps.
> > > 
> > > 
> > > so this will be a "scraper" client i suppose ?
> > 
> > 
> > Yeah.
> 
> so in this client , we recieve a json file .

Yeah.

> How exactly do we want to represent this data, because i dont think this is related to attrs or attrs_register. 

I would assume we load the data into TimeSeries with a datetime index and the other columns as data. 

[...]

PR #6744, opened 2023-01-29
Example to read and load GOES XRS NRT JSON data

ryuusama09:
## PR Description

Add an example of reading and loading GOES NRT JSON.


sunpy/net/dataretriever/sources/noaa.py
```diff
@@ -174,3 +174,29 @@ def register_values(cls):
                  attrs.Source: [('SWPC', 'The Space Weather Prediction Center.')],
                  attrs.Provider: [('NOAA', 'The National Oceanic and Atmospheric Administration.')]}
         return adict
+    
+class WeatherDataClient(GenericClient):
```
nabobalis, line 178: This needs a better name, like the other clients.

sunpy/net/dataretriever/sources/noaa.py
```diff
@@ -174,3 +174,29 @@ def register_values(cls):
...
+    """documentation review required from mentor"""
```
nabobalis, line 179: You will need to add the relevant information for this client.

sunpy/net/dataretriever/sources/noaa.py
```diff
@@ -174,3 +174,29 @@ def register_values(cls):
...
+        rowdict['url'] = 'https://services.swpc.noaa.gov/json/goes/primary/differential-electrons-1-day.json'
+        rowdict['Instrument'] = 'NOAA-Predict' # not sure about this , needs help
+        # These results are not dependent on time, but we allow time as a
+        # parameter for easy searching, so remove time from the results table
+        # injected by GenericClient.
+        #this is useless as this data is time invariant 
+        rowdict.pop('Start Time', None)
+        rowdict.pop('End Time', None)
+        return QueryResponse([rowdict], client=self)
+
```
nabobalis, line 202: I think you will need a 

```
register_values
```
method so this client can be searched.

nabobalis:
> fixes #5380

The original issue has two parts, the client and adding support for loading this file in timeseries. 
As a result, this PR alone will not close that issue, so its best to not say it fixes that issue.

nabobalis:
> @nabobalis , here is the raw implementation. The NRT GOES data i suppose as given is time invariant , but there are several types of files that can be retrieved , temporary ive only added the link to one such file. Let me know how the implementation should be

What other files?

sunpy/net/dataretriever/sources/noaa.py
nabobalis: Which are?

sunpy/net/dataretriever/sources/noaa.py
nabobalis: I am not sure I understand this comment.

nabobalis:
> Theyre : image so in one go we want to get all these files ? currently it only gets `differential-electrons-1-day.json`

Unless we know what each file is, I don't think we want to download them all.

sunpy/net/dataretriever/sources/noaa.py
nabobalis: > What does it mean by loading into TimeSeries, do you mean the folder ?

I don't understand why this reply is here but Timeseries is a class to read tabular data such as this.

sunpy/net/dataretriever/sources/noaa.py
nabobalis: No, I don't think this will have that instrument.

nabobalis: Yes I would imagine that is the case. We should work that out.

nabobalis: Right now I don't, if you could try to come up with one that would be great.

nabobalis: Why whose?

nabobalis: But why those values?

nabobalis: Can you research and work out which one is the answer please?

nabobalis: Ok, we will use that.

nabobalis:
> Now we need to finalize which files do we want to fetch. should we leave upto the user to query on this ?. but the file names dont make sense to me 😅

It would be good if we could let them pick the file they wanted.

nabobalis:
> alright ,so heres what we can do , first present them with the file list . let them select the file and add that file name to the base url to fetch that file , if they select none then fetch all files

While that isn't a bad idea, the problem is that Fido and the GenericClient do not work like this. The user input comes from the search input, and not picking items from a list.

nabobalis:
> then we are left with 2 options :
> 
> 1. expect the user to know the names of the files which i hardly think its possible.

Users can not pass in filenames when downloading or searching for files. I would strongly suggest you have a go at searching and downloading a file to see how it all works.

> 2. maintain all the file names in the documentation and pass them as constructor params while calling theclient for eg the file `differential-electrons-1-day.json` is needed then :  `client = NOAAWeatherClient('differential-electrons-1-day.json')`

Users also do not construct clients, it is done in the background at import time.

Unforntally, neither option is open to us here.

nabobalis:
I think for now, should focus on one file.

nabobalis:
> @nabobalis , in the current code that i have pushed , it adds the functionality to add the filename while making the client object

But how does a user get this to work?

hayesla:
Hi @ryuusama09  and @nabobalis  I guess theres a bigger question here - do we want a way to be able to query the NRT data of all the space-weather products provided here? or only the GOES XRS NRT data? 

In terms of a users perspective, you will need to have some attribute for the search such that a user passed something along the lines of `a.Product.NRT`  or something. 

However this still may be best to just be an example in the gallery rather than its own client - what do people think?

hayesla:
so for some more context here and following from the original issue:

The GOES-XRS near real-time (NRT) data is provided here:
https://services.swpc.noaa.gov/json/goes/primary/xrays-7-day.json and updates every minute with a new data point (and deletes an old one). The other files i.e. xrays-1-day.json and xrays-3-day.json are the same, but only have the past day and three days respectively. The reason a use may want this data is 1. for quick space weather monitoring, 2. quickly look at flares over the past few days before the science quality data is available (which is usually a few days) and then it can be accessed via the client we already have.

nabobalis:
> ive mentioned an example in the codebase like :
> 
> > > > from sunpy.net import attrs as a
> > > > from sunpy.net.dataretriever import NOAAWeatherClient
> > > > client = NOAAWeatherClient(fine-name)

This is not how a user downloads data. This will not work as a result.

nabobalis:
> > with the client is tha
> 
> I agree on this with @nabobalis

On what specifically?

nabobalis:
> image this example worked for me locally , is this fine or do you want me to use `Fido` ?

The clients are not designed to be called directly, you have to use Fido.

nabobalis:
> That Searching on the basis of attribute is difficult to implement now . (for me atleast 😅)

In that case, I think we should go about this as an example instead.

nabobalis:
> We need to add a new attribute i think because goes has already been taken by XRSClient image

It should be possible to use the same attribute for instrument.

nabobalis:
At this point, It appears to me, we have too many unsolved questions about creating a new client.

So by following: https://github.com/sunpy/sunpy/issues/5380#issuecomment-1408952614 and creating a new gallery example using the linked snippet is a much more concrete path.

nabobalis:
> So by creating an entire example gallery , we can avoid showing Fido.search() and show how the client can be directly used to query and parse json data ?.

We would avoid any client, and use pandas to download the JSON file.

nabobalis:
> For that i need to update the client column in instrument list , can you suggest me how can it be done ?

That should be covered by the developer documentation on creating clients.

nabobalis:
> I meant we could fetch file in the traditional way like the client does and then play with the file with all sorts of functions

Yeah that would be the example but we can use pandas to download it as the code example does in the linked issue.

nabobalis:
> alright , this is also good enough , avoiding the client. But we can wrap all the methods and other functionalities into one class and help user to avoid such lengthy code :). this class is not a client so it can be implemented into dataretriever as a new file as it is not dependant on `base_client`

What methods and what lengthy code? 
I don't see the need for any new class, just a gallery example.

nabobalis:
> Alright my idea was , lets say the user wanted to mimic the example mentioned in #5380 . Some 10-12 lines of code has to be written , what if we provide a convenient way to the user by just passing in some methods produce the same result. I just mentioned this as an idea 😅. This will reduce code complexity and help write short i feel (upto some extent). But let us do what @hayesla and you are proposing

What 10-12 lines? I don't quite follow.

nabobalis:
> the example which is mentioned [#4346 (comment)](https://github.com/sunpy/sunpy/issues/4346#issuecomment-722368424). I hope you get it here ?

But there is only one line that does the download in that example.

nabobalis:
> I meant except for the download part , other functionalities like let say `show_data_report()` or `count_data_attributes()`. these are just for reference , we can have a good discussion to add such methods. If this does not suite well , then let us proceed with example gallery.

What would `show_data_report()` or `count_data_attributes()` do? 
Would they be methods on a class or a function?

There will no new random class added to net, only network clients live in this area of sunpy.

nabobalis:
> for example `show_data_report()`(class method). there are many then for a particular time , or range of time produce data , or like produce all data collected by satellite-18 or any other functionality that is not yet implemented but is helpful to the user. If this is not so significant we can stop thinking about this. 

Honestly, I don't really understand. I would want to avoid adding methods to a class, especially since I don't know which class or what these methods would do.

We should focus on the gallery example by taking that code snippet and expanding it. 

> I actually feel the nuance in communication over here , so please bear with me 😥

No problem.

nabobalis:
> how the example needs to be expanded ?. explaining every line ?

Yes, I would suggest you have a look at the other examples to see how they are written.
Most of an example is text, and not code. Some examples are pretty short.

I would also like to expand it to some of the other JSON files that are present.

nabobalis:
> also if you dont mind , should i just code the part which i am trying to explain maybe it would explain you better ?

Try doing both and I will help you.

examples/time_series/nooa_json.py
```diff
@@ -0,0 +1,36 @@
+import pandas as pd
```
nabobalis, line 1: Can you make this example look like the others? 

There needs to be a whole skeleton for this example to work.

mytest.py
```diff
@@ -0,0 +1,5 @@
+
```
nabobalis, line 1: Can you remove this?

sunpy/net/dataretriever/sources/noaa.py
```diff
@@ -174,3 +174,69 @@ def register_values(cls):
                  attrs.Source: [('SWPC', 'The Space Weather Prediction Center.')],
                  attrs.Provider: [('NOAA', 'The National Oceanic and Atmospheric Administration.')]}
         return adict
+"""
```
nabobalis, line 177: This should not be here.

nabobalis:
> can you suggest some ways in which user could possibly want to query the nrt data ? @nabobalis

Query as in doing analysis on it? 

I think its ok if we show someone how to download the file, load it into timeseries and plot it.
I would like this done for a selection of the JSON files.

nabobalis:
> yes , except for plotting , are there any querying methods that user would want ?

I think we can leave that up to the users to decide, we can focus on  showing them how to get the data into sunpy.

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,72 @@
...
+
+import pandas as pd 
+from sunpy.time import parse_time
+from sunpy import timeseries as ts 
+from collections import OrderedDict
+from astropy import units as u 
+
+data = pd.read_json("https://services.swpc.noaa.gov/json/goes/primary/xrays-7-day.json")
+
+#################### retrieval ########################
```
nabobalis, line 20: These are not how the comment blocks are created in examples.

nabobalis:
> I think this should be enough to fix the issue ? (both the parts ) @nabobalis

I would really suggest that you review the Files Changed tab more often and making sure that when you make changes, the documentation builds locally and looks nice in the final HTML.

sunpy/net/dataretriever/sources/noaa.py
```diff
@@ -174,3 +174,7 @@ def register_values(cls):
                  attrs.Source: [('SWPC', 'The Space Weather Prediction Center.')],
                  attrs.Provider: [('NOAA', 'The National Oceanic and Atmospheric Administration.')]}
         return adict
+
```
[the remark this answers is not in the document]
nabobalis: No, I mean all these new lines.

examples/time_series/noaa_json.py
nabobalis: You should look over the other examples and see how they are done and replicate that.

nabobalis: We use https://sphinx-gallery.github.io/stable/index.html to create these examples and its important to follow their syntax for that.

nabobalis:
> > the documentation builds locally and looks nice in the final HTML.
> 
> i didnt get this , could explain me what you meant

The documentation is rendered into html. These examples for example end up looking like this: https://docs.sunpy.org/en/stable/generated/gallery/index.html

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,74 @@
+"""
+===================================================================
+Plotting of near real time (NRT) data by fetching it from json file
+====================================================================
```
nabobalis, line 4:
```suggestion
==================================
GOES XRS Near Real Time (NRT) data
==================================
```

nabobalis:
> > ml. These examples for example end up looking like t
> 
> yes while building locally , its showing a fine plot in a pop up window

Sure but just because a plot shows up, doesn't mean the example renders correctly.

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,74 @@
...
+
+How to query and visualize nrt data(json format) using pandas and plot
+it using TimeSeries Library
```
nabobalis, line 7:
```suggestion
Currently, ``sunpy`` has no mechanism to directly download several of the the GOES XRS Near Real Time (NRT) data.

We will showcase an example of how to download it and load it into a `sunpy.timeseries.TimeSeries`.
```

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,74 @@
...
+"""
+
+
```
nabobalis, line 10:
```suggestion
```

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,74 @@
...
+from collections import OrderedDict
+
+import pandas as pd
+
+from astropy import units as u
+
+from sunpy import timeseries as ts
+from sunpy.time import parse_time
+
+data = pd.read_json("https://services.swpc.noaa.gov/json/goes/primary/xrays-7-day.json")
```
nabobalis, line 20:
```suggestion
```
Examples need to explain these type of lines.

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,74 @@
...
+
+############################################
```
nabobalis, line 22: These should all be 79 or 80 characters long (I forgot which) to match our existing examples.

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,74 @@
...
+# over here the json is transformed and stored as a pandas dataframe
+
+data_short = data[data["energy"] == "0.05-0.4nm"]
```
nabobalis, line 25:
```suggestion
# We will start by getting reading the GOES JSON file using `pandas`.
# It allows us to download the file and get it straight into a `pandas.DataFrame`.
# This file updates every minute and contains only the last 7 days worth of data.

goes_json_data = pd.read_json("https://services.swpc.noaa.gov/json/goes/primary/xrays-7-day.json")
```

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,74 @@
...
+
+##################################################################
+# Filtering the data: A new dataframes are created by filtering the original data.
+# The data_short dataframe contains only the rows where the value of the "energy" column is "0.05-0.4nm"
+
+data_long = data[data["energy"] == "0.1-0.8nm"]
```
nabobalis, line 31:
```suggestion
# The aim here is to display the two seperate energy bands that GOES can measure.
# These being: "0.05-0.4nm" and "0.1-0.8nm". Therefore we will need to filter the data.

# This will get us the short wavelength data.
goes_short = data[data["energy"] == "0.05-0.4nm"]
# This will get us the long wavelength data.
goes_long = data[data["energy"] == "0.1-0.8nm"]
```

nabobalis:
Yes, I would be worried if they looked the same.

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,74 @@
...
+
+##################################################################
+# Again a new temporary dataframe is formed ,and the data_long contains only
+# the rows where the value of the "energy" column is "0.1-0.8nm".
+
+time_array = parse_time(data_short["time_tag"])
```
nabobalis, line 37:
```suggestion
# Similar to a `pandas.DataFrame`, `sunpy.timeseries.TimeSeries` requires
# a datetime index which we can get directly and transform into `astropy.time.Time`.

time_array = parse_time(data_short["time_tag"])
```

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,74 @@
...
+
+##################################################
+# The "time_tag" column from the data_short dataframe is passed
+# to the parse_time function to parse the strings into Time objects.
+
+units = OrderedDict([("xrsa", u.W/u.m**2), ("xrsb", u.W/u.m**2)])
```
nabobalis, line 43:
```suggestion
# `sunpy.timeseries.TimeSeries` requires that there are units for data variables.
# To do this, we will create a dictionary that will map the names of the two columns,
# "xrsa" and "xrsb" (the channel names for GOES XRS), to their corresponding physical units, ``u.W/u.m**2``.

units = dict([("xrsa", u.W/u.m**2), ("xrsb", u.W/u.m**2)])
```

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,74 @@
...
+# this plots data for energy range from data_short as xrsa and energy range from dta_long as xrsb
+
+meta = OrderedDict({"instrument": "GOES X-ray sensor", "measurments": "primary", "type": "quicklook"})
+
+#######################################################
+# The meta variable is defined as an OrderedDict that contains information about the data,
+# such as the instrument used to obtain the data and the type of data.
+# it can contain other types of metadata as well
+
+goes_data = pd.DataFrame({"xrsa": data_short["flux"].values, "xrsb": data_long["flux"].values}, index=time_array.datetime)
```
nabobalis, line 57: Do we need to do this step? I feel like surely we could bypass this and go directly to the timeseries call.

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,74 @@
...
+# to the parse_time function to parse the strings into Time objects.
+
+units = OrderedDict([("xrsa", u.W/u.m**2), ("xrsb", u.W/u.m**2)])
+
+#####################################################
+# The units variable is defined as an OrderedDict that maps the names of the two columns,
+# "xrsa" and "xrsb", to the corresponding physical units, u.W/u.m**2.
+# this plots data for energy range from data_short as xrsa and energy range from dta_long as xrsb
+
+meta = OrderedDict({"instrument": "GOES X-ray sensor", "measurments": "primary", "type": "quicklook"})
```
nabobalis, line 50:
```suggestion
# Typically `sunpy.timeseries.TimeSeries` will read metadata from the file,
# however, here we need to define our own metadata and we will keep it fairly simple.

meta = dict({"instrument": "GOES X-ray sensor", "measurements": "primary", "type": "quicklook"})
```

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,74 @@
...
+# using the values of the "flux" columns from the data_short
+# and data_long dataframes, with the parsed time data as the index.
+
+test_ts = ts.TimeSeries(goes_data, meta, units, source="xrs")
+
+####################################################################
+# Creating the time series object: The TimeSeries class from the sunpy.timeseries module is used to create a time series object,
+# test_ts, from the goes_data dataframe, the meta metadata, the units units, and a source string "xrs".
+
+test_ts.peek()
```
nabobalis, line 71:
```suggestion
test_ts.plot()

plt.show()
```

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,74 @@
...
+
+############################################################
+# Plotting the time series: The peek method of the test_ts object is called to plot the time series data.
```
nabobalis, line 74:
```suggestion
```

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,74 @@
...
+# after filtering all the data the final data is merged into goes_data as final dataset
+# Creating the time series data: A new Pandas dataframe, goes_data, is created
+# using the values of the "flux" columns from the data_short
+# and data_long dataframes, with the parsed time data as the index.
+
+test_ts = ts.TimeSeries(goes_data, meta, units, source="xrs")
+
+####################################################################
+# Creating the time series object: The TimeSeries class from the sunpy.timeseries module is used to create a time series object,
+# test_ts, from the goes_data dataframe, the meta metadata, the units units, and a source string "xrs".
```
nabobalis, line 69:
```suggestion
# Now we have created the timeseries, we will finally plot it.
```

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,74 @@
...
+# such as the instrument used to obtain the data and the type of data.
+# it can contain other types of metadata as well
+
+goes_data = pd.DataFrame({"xrsa": data_short["flux"].values, "xrsb": data_long["flux"].values}, index=time_array.datetime)
+
+###########################################################
+# after filtering all the data the final data is merged into goes_data as final dataset
+# Creating the time series data: A new Pandas dataframe, goes_data, is created
+# using the values of the "flux" columns from the data_short
+# and data_long dataframes, with the parsed time data as the index.
```
nabobalis, line 63:
```suggestion
# Now we will create a `sunpy.timeseries.TimeSeries` by passing in the data, the metadata and the units.
```

nabobalis (commented):
Some minor comments

nabobalis:
> so making a check-list : wrote the explanation of file `'noaa_json.py` by taking format reference from `power_spectra_example.py`. removed the client from `noaa.py` checked the validity of example plot

Sounds fair.

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,74 @@
+"""
```
nabobalis, line 1: The file needs renaming to be more specific.

nabobalis: I'm not sure those are specific enough. I'll have a think.

nabobalis: goes_xrs_nrt_data.py ?

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,66 @@
...
+==================================
+GOES XRS Near Real Time (NRT) data
+==================================
+
+Currently, ``sunpy`` has no mechanism to directly download several of the the GOES XRS Near Real Time (NRT) data.
+
+We will showcase an example of how to download it and load it into a `sunpy.timeseries.TimeSeries`.
```
nabobalis, line 8:
```suggestion
We will showcase an example of how to download and load these files into a `sunpy.timeseries.TimeSeries`.
```

examples/time_series/noaa_json.py
```diff
@@ -0,0 +1,72 @@
...
+
+###############################################################################
+# The meta variable is defined as an OrderedDict that contains information about the data,
+# such as the instrument used to obtain the data and the type of data.
+# it can contain other types of metadata as well
+
+goes_data = pd.DataFrame({"xrsa": data_short["flux"].values, "xrsb": data_long["flux"].values}, index=time_array.datetime)
+
+###############################################################################
+# Now we will create a `sunpy.timeseries.TimeSeries` by passing in the data, the metadata and the units.
```
nabobalis, line 63:
```suggestion
# Now we will create a `sunpy.timeseries.TimeSeries` by passing in the data, 
# the metadata and the units.
```

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+# however, here we need to define our own metadata and we will keep it fairly simple.
+
+meta = dict({"instrument": "GOES X-ray sensor", "measurements": "primary", "type": "quicklook"})
+
+###############################################################################
+# The meta variable is defined as an OrderedDict that contains information about the data,
+# such as the instrument used to obtain the data and the type of data.
+# it can contain other types of metadata as well
+
+goes_data = pd.DataFrame({"xrsa": data_short["flux"].values, "xrsb": data_long["flux"].values}, index=time_array.datetime)
```
nabobalis, line 58: So we can't pass these variables directly into Timeseries?

nabobalis:
The pre-commit needs fixing and a changelog will need to be added.

Once those are done, I will request reviews from others.

nabobalis:
> just curious : how did precommit bot break ,was it something from organization's end ?

Break how?

nabobalis:
> As you mentioned the bot needs fixing , so i though it mustve broke in some update

Ah I see, I mean't the check was failing.

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+###############################################################################
+# We will start by getting reading the GOES JSON file using `pandas`.
+# It allows us to download the file and get it straight into a `pandas.DataFrame`.
+# This file updates every minute and contains only the last 7 days worth of data.
+
+goes_json_data = pd.read_json("https://services.swpc.noaa.gov/json/goes/primary/xrays-7-day.json")
+
+###############################################################################
+# The aim here is to display the two separate energy bands that GOES can measure.
+# These being: "0.05-0.4nm" and "0.1-0.8nm". Therefore we will need to filter the data.
```
alasdairwilson, line 27:
```suggestion
# XRS collects data in two energy channels, "0.05-0.4nm" and "0.1-0.8nm".
# We separate these `short` and `long` wavelength readings into two arrays.
```

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
+"""
+==================================
+GOES XRS Near Real Time (NRT) data
+==================================
+
+Currently, ``sunpy`` has no mechanism to directly download several of the the GOES XRS Near Real Time (NRT) data.
```
alasdairwilson, line 6:
```suggestion
Currently, ``sunpy`` has no mechanism to directly download the GOES XRS Near Real Time (NRT) data.
```

A small change but I am also unsure on the phrase "no mechanism to", it is probably fine to stay though.

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+goes_data = pd.DataFrame({"xrsa": data_short["flux"].values, "xrsb": data_long["flux"].values}, index=time_array.datetime)
+
+###############################################################################
+# Now we will create a `sunpy.timeseries.TimeSeries` by passing in the data,
+# the metadata and the units.
+
+test_ts = ts.TimeSeries(goes_data, meta, units, source="xrs")
+
+###############################################################################
+# Now we have created the timeseries, we will finally plot it.
```
alasdairwilson, line 67:
```suggestion
# Finally, we can plot the timeseries.
```

alasdairwilson (commented):
Good work 👍

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+
+import pandas as pd
+
+from astropy import units as u
+
+from sunpy import timeseries as ts
+from sunpy.time import parse_time
+
+###############################################################################
+# We will start by getting reading the GOES JSON file using `pandas`.
```
hayesla, line 19:
```suggestion
# We will start by getting reading the GOES-XRS JSON file using `pandas.read_json`.
```

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+###############################################################################
+# Similar to a `pandas.DataFrame`, `sunpy.timeseries.TimeSeries` requires
+# a datetime index which we can get directly and transform into `astropy.time.Time`.
+
+time_array = parse_time(data_short["time_tag"])
+
+###############################################################################
+# `sunpy.timeseries.TimeSeries` requires that there are units for data variables.
+# To do this, we will create a dictionary that will map the names of the two columns,
+# "xrsa" and "xrsb" (the channel names for GOES XRS), to their corresponding physical units, ``u.W/u.m**2``.
```
hayesla, line 43:
```suggestion
# "xrsa" and "xrsb" (the channel names for GOES XRS), to their corresponding physical flux units, ``u.W/u.m**2``.
```

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
+"""
+==================================
+GOES XRS Near Real Time (NRT) data
```
hayesla, line 3: I would rephase this:

```suggestion
How to create a `sunpy.timeseries.TimeSeries` from the GOES-XRS near real time (NRT). 
```

examples/time_series/goes_xrs_nrt_data.py
hayesla: I agree. I suggested its reworded to something along the lines of:

```suggestion
NOAA provides the GOES-XRS near-real time data of observations of the past 7 days that is updated ever minute in a json file format. Here we shall show an example of how to load this data and create a `sunpy.timeseries.TimeSeries`. Note that the scientific quality data can be  queried and downloaded through Fido.
```

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+import pandas as pd
+
+from astropy import units as u
+
+from sunpy import timeseries as ts
+from sunpy.time import parse_time
+
+###############################################################################
+# We will start by getting reading the GOES JSON file using `pandas`.
+# It allows us to download the file and get it straight into a `pandas.DataFrame`.
```
hayesla, line 20:
```suggestion
# This allows us to download the file and load it straight into a `pandas.DataFrame`.
```

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+# The aim here is to display the two separate energy bands that GOES can measure.
+# These being: "0.05-0.4nm" and "0.1-0.8nm". Therefore we will need to filter the data.
+
+# This will get us the short wavelength data.
+goes_short = data[data["energy"] == "0.05-0.4nm"]
+# This will get us the long wavelength data.
+goes_long = data[data["energy"] == "0.1-0.8nm"]
+
+###############################################################################
+# Similar to a `pandas.DataFrame`, `sunpy.timeseries.TimeSeries` requires
```
hayesla, line 35:
```suggestion
# `sunpy.timeseries.TimeSeries` requires
```

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+
+###############################################################################
+# `sunpy.timeseries.TimeSeries` requires that there are units for data variables.
+# To do this, we will create a dictionary that will map the names of the two columns,
+# "xrsa" and "xrsb" (the channel names for GOES XRS), to their corresponding physical units, ``u.W/u.m**2``.
+
+units = dict([("xrsa", u.W/u.m**2), ("xrsb", u.W/u.m**2)])
+
+###############################################################################
+# Typically `sunpy.timeseries.TimeSeries` will read metadata from the file,
```
hayesla, line 48:
```suggestion
# To create a `sunpy.timeseries.TimeSeries` we can create a meta dictionary that provides the detailed of the data. Typically, `sunpy.timeseries.TimeSeries` reads the metadata directly from the file.
```

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+# such as the instrument used to obtain the data and the type of data.
+# it can contain other types of metadata as well
+
+goes_data = pd.DataFrame({"xrsa": data_short["flux"].values, "xrsb": data_long["flux"].values}, index=time_array.datetime)
+
+###############################################################################
+# Now we will create a `sunpy.timeseries.TimeSeries` by passing in the data,
+# the metadata and the units.
+
+test_ts = ts.TimeSeries(goes_data, meta, units, source="xrs")
```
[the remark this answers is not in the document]
hayesla: yes I agree `goes_ts` works!

hayesla (commented):
Looks great! thanks for creating it, some suggestions on the text and then it should be good to go

nabobalis:
> @nabobalis ,, i suppose the review is done so can we go ahead with merging ?

A review has not taken place. There is no need to rush merge this.

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
+"""
+==================================
+How to create a `sunpy.timeseries.TimeSeries` from the GOES-XRS near real time (NRT).
+==================================
```
nabobalis, line 4:
```suggestion
=======================================================
Creating a TimeSeries from GOES-XRS near real time data
=======================================================
```

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+
+Currently, ``sunpy`` has no mechanism to directly download the GOES XRS Near Real Time (NRT) data.
+
+We will showcase an example of how to download and load these files into a `sunpy.timeseries.TimeSeries`.
```
nabobalis, line 8:
```suggestion
We will showcase an example of how to download and load GOES XRS Near Real Time (NRT) data into a `sunpy.timeseries.TimeSeries`.
```

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+###############################################################################
+# We will start by getting reading the GOES-XRS JSON file using `pandas.read_json`.
+# This allows us to download the file and load it straight into a `pandas.DataFrame`.
+# This file updates every minute and contains only the last 7 days worth of data.
+
+goes_json_data = pd.read_json("https://services.swpc.noaa.gov/json/goes/primary/xrays-7-day.json")
+
+###############################################################################
+# XRS collects data in two energy channels, "0.05-0.4nm" and "0.1-0.8nm".
+# We separate these `short` and `long` wavelength readings into two arrays.
```
nabobalis, line 27:
```suggestion
# We separate these "short" and "long" wavelength readings into two arrays.
```

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+
+# This will get us the short wavelength data.
+goes_short = data[data["energy"] == "0.05-0.4nm"]
+# This will get us the long wavelength data.
+goes_long = data[data["energy"] == "0.1-0.8nm"]
+
+###############################################################################
+# `sunpy.timeseries.TimeSeries` requires
+# a datetime index which we can get directly and transform into `astropy.time.Time`.
```
nabobalis, line 36:
```suggestion
# `sunpy.timeseries.TimeSeries` requires a datetime index which we can get 
# directly and transform into `astropy.time.Time`.
```

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+
+time_array = parse_time(data_short["time_tag"])
+
+###############################################################################
+# `sunpy.timeseries.TimeSeries` requires that there are units for data variables.
+# To do this, we will create a dictionary that will map the names of the two columns,
+# "xrsa" and "xrsb" (the channel names for GOES XRS), to their corresponding physical flux units, ``u.W/u.m**2``.
```
nabobalis, line 43:
```suggestion
# To do this, we will create a dictionary that will map the names of the two columns,
# "xrsa" and "xrsb" (the channel names for GOES XRS), to their corresponding
# physical flux units, ``u.W/u.m**2``.
```

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+
+units = dict([("xrsa", u.W/u.m**2), ("xrsb", u.W/u.m**2)])
+
+###############################################################################
+# To create a `sunpy.timeseries.TimeSeries` we can create a meta dictionary that provides the detailed of the data. Typically, `sunpy.timeseries.TimeSeries` reads the metadata directly from the file.
+# however, here we need to define our own metadata and we will keep it fairly simple.
```
nabobalis, line 49:
```suggestion
# We need to create a metadata dictionary for the data.
# Typically, `sunpy.timeseries.TimeSeries` reads the metadata directly from the file.
# However, here we need to define our own metadata and we will keep it fairly simple.
```

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+
+meta = dict({"instrument": "GOES X-ray sensor", "measurements": "primary", "type": "quicklook"})
+
+###############################################################################
+# The meta variable is defined as an OrderedDict that contains information about the data,
+# such as the instrument used to obtain the data and the type of data.
+# it can contain other types of metadata as well
```
nabobalis, line 56:
```suggestion
#  The final pre-step is create a new `pandas.DataFrame` which we can pass to
# `sunpy.timeseries.TimeSeries` as the data input.
```

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+
+goes_data = pd.DataFrame({"xrsa": data_short["flux"].values, "xrsb": data_long["flux"].values}, index=time_array.datetime)
+
+###############################################################################
+# Now we will create a `sunpy.timeseries.TimeSeries` by passing in the data,
+# the metadata and the units.
+
+goes_ts = ts.TimeSeries(goes_data, meta, units, source="xrs")
```
nabobalis, line 64: Do we need source here?

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,71 @@
...
+
+###############################################################################
+# Finally, we can plot the timeseries.
+
+goes_ts.plot()
```
nabobalis, line 69:
```suggestion
plt.figure()
goes_ts.plot()
```

nabobalis:
We will need a changelog but I think we are getting close to this being finished.

nabobalis:
Example failed due to:

```
Unexpected failing examples:
/home/runner/work/sunpy/sunpy/examples/time_series/goes_xrs_nrt_data.py failed leaving traceback:
  File "/home/runner/work/sunpy/sunpy/examples/time_series/goes_xrs_nrt_data.py", line 68, in <module>
    plt.figure()
NameError: name 'plt' is not defined
```

Can you also add a changelog please?

changelog/6744.feature.rst
```diff
@@ -0,0 +1,2 @@
+In  `examples\time_series` added a new example `goes_xrs_nrt_json.py`
+This example shows  `how the goes_nrt` data can be queried using `matplotlib` and `sunpy.Timseries`
```
nabobalis, line 2:
```suggestion
Added an example (:ref:`sphx_glr_generated_gallery_time_series_goes_xrs_nrt_data.py`) to download GOES NRT data and load it into `~sunpy.timeseries.TimeSeries`.
```

The file should be renamed from `feature` to `doc`.

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -0,0 +1,72 @@
+"""
+=======================================================
+Creating a TimeSeries from GOES-XRS near real time data
+=======================================================
+
+We will showcase an example of how to download and load GOES XRS Near Real Time (NRT) data into a `sunpy.timeseries.TimeSeries`.
```
nabobalis, line 6:
```suggestion
This example will demonstrate how to download and load GOES XRS Near Real Time (NRT) data into a `sunpy.timeseries.TimeSeries`.
```

alasdairwilson (approved):
For me, this is good to go now. Maybe @hayesla wants another look?

nabobalis:
Thanks for the PR @ryuusama09

nabobalis:
> For me, this is good to go now. Maybe @hayesla wants another look?

Yeah, if she has sometime, that would be great. 
I can make any follow up corrections in a new PR.

PR #6758, opened 2023-02-01
Use secrets.GITHUB_TOKEN for Cron jobs

nabobalis:
Not sure if this will help or if the reason a custom token was used was because it needs specific permissions that the standard token lacks.


ConorMacBride:
The standard token has (or at least had) protections to prevent accidental recursive workflow calls, so I don't think this will work.

ConorMacBride:
Still the case: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

PR #6764, opened 2023-02-02
Raise actual error message from failed read

nabobalis:
We don't actually raise the error message when we fail to load a file and its annoying.


wtbarnes:
Should this be backported? It's not really a bug is it?

wtbarnes (approved):

nabobalis:
Unbackported.

PR #6768, opened 2023-02-05
Update functions due to some tests in sunkit-image

nabobalis:
Some minor changes to the API of our internal test functions.

I hit some problems trying to reuse these functions in sunkit-image.

### DIFF
--- a/sunpy/data/test/__init__.py
+++ b/sunpy/data/test/__init__.py
@@ -29,7 +29,7 @@ rootdir = Path(os.path.dirname(sunpy.__file__)) / "data" / "test"
 file_list = glob.glob(os.path.join(rootdir, '*.[!p]*'))
 
 
-def get_test_filepath(filename, **kwargs):
+def get_test_filepath(filename, package="sunpy.data.test", **kwargs):
     """
     Return the full path to a test file in the ``data/test`` directory.
 
@@ -37,6 +37,10 @@ def get_test_filepath(filename, **kwargs):
     ----------
     filename : `str`
         The name of the file inside the ``data/test`` directory.
+    package : `str`
+        The package name, defaults to ``sunpy.data.test``.
+    kwargs : `dict`
+        Passed to `astropy.utils.data.get_pkg_data_filename`.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


sunpy/data/test/__init__.py
```diff
@@ -51,7 +55,7 @@ def get_test_filepath(filename, **kwargs):
     if isinstance(filename, Path):
         # NOTE: get_pkg_data_filename does not accept Path objects
         filename = filename.as_posix()
-    return get_pkg_data_filename(filename, package="sunpy.data.test", **kwargs)
+    return get_pkg_data_filename(filename, package=package, **kwargs)
```
nabobalis, line 58: I wanted to use this function, but could not override package.

Either this is used inside sunpy only and we make all other packages call get_pkg_data_filename or this change needs to happen.

sunpy/data/test/__init__.py
```diff
@@ -103,16 +107,15 @@ def write_image_file_from_header_file(header_file, fits_directory):
...
 
-def get_dummy_map_from_header(filename):
+def get_dummy_map_from_header(filepath):
     """
     Generate a dummy `~sunpy.map.Map` from header-only test data.
 
     The "image" will be random numbers with the correct shape
     as specified by the header.
     """
-    filepath = get_test_filepath(filename)
```
nabobalis, line 113: Due to the above API change, I think we should make this function use the file path instead of having to add more args to this.

sunpy/data/test/__init__.py
```diff
@@ -103,16 +107,15 @@ def write_image_file_from_header_file(header_file, fits_directory):
...
     header = _fits.format_comments_and_history(astropy.io.fits.Header.fromtextfile(filepath))
-    data = np.random.rand(header['NAXIS2'], header['NAXIS1'])
+    data = np.random.randint(low=0, high=100, size=(header['NAXIS2'], header['NAXIS1']))
```
nabobalis, line 118: I thought we should create ints instead and use a small range of values.

sunpy/data/test/__init__.py
wtbarnes: Uh I mean fine, but why not just use `get_pkg_data_filename` directly from astropy?

nabobalis: We could and I can close this PR.

PR #6773, opened 2023-02-08
Hover for numpy, astropy and python

nabobalis:
We don't hover, we hoover.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -204,24 +204,19 @@ intersphinx_mapping = {
 }
 
 # -- Options for hoverxref -----------------------------------------------------
-# adapted from sphinx-hoverxref conf.py
 if os.environ.get("READTHEDOCS"):
-    # Building on Read the Docs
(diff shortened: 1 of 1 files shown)


ConorMacBride (approved):
This worked in RTD

docs/conf.py
```diff
@@ -234,13 +229,13 @@
...
     "mod": "tooltip",
     "obj": "tooltip",
-    #
     # roles with std domain
     "confval": "tooltip",
     "hoverxref": "tooltip",
     "ref": "tooltip",
     "term": "tooltip",
 }
+hoverxref_intersphinx = ["numpy", "astropy", "python"]
```
alasdairwilson, line 238: We use scipy in places, worth adding?

nabobalis: I don't think they mention scipy as supported but I should check as both scipy and numpy are basically the same doc site.

PR #6778, opened 2023-02-10
Tweak settings for hover

nabobalis:
Hopefully should address some of the issues we have with it.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -217,27 +217,7 @@ hoverxref_domains = ["py"]
 hoverxref_mathjax = True
 hoverxref_modal_hover_delay = 500
 hoverxref_tooltip_maxwidth = 600  # RTD main window is 696px
-hoverxref_intersphinx = [
-    "astropy",
-    "matplotlib",
-    "numpy",
-    "python",
-    "scipy",
-    "hvpy",
-    "asdf",
-    "zeep",
-    "sunkit_instruments",
-    "sunkit_image",
-    "sqlalchemy",
-    "skimage",
-    "reproject",
-    "parfive",
-    "pandas",
-    "mpl_animators",
-    "drms",
-    "astroquery",
-    "aiapy",
-]
+hoverxref_intersphinx = list(intersphinx_mapping.keys())
 hoverxref_role_types = {
     # Roles within the py domain
     "attr": "tooltip",
@@ -249,7 +229,7 @@ hoverxref_role_types = {
     "meth": "tooltip",
     "mod": "tooltip",
     "obj": "tooltip",
-    # Roles within the standard domain
+    # Roles within the std domain
     "confval": "tooltip",
     "hoverxref": "tooltip",
     "ref": "tooltip", # Would be used by hoverxref_auto_ref if we set it to True
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -185,11 +185,13 @@ intersphinx_mapping = {
         "https://docs.scipy.org/doc/scipy/reference/",
         (None, "http://www.astropy.org/astropy-data/intersphinx/scipy.inv"),
     ),
-    "matplotlib": ("https://matplotlib.org/stable", None),
     "aiapy": ("https://aiapy.readthedocs.io/en/stable/", None),
+    "asdf": ("https://asdf.readthedocs.io/en/stable/", None),
     "astropy": ("https://docs.astropy.org/en/stable/", None),
     "astroquery": ("https://astroquery.readthedocs.io/en/latest/", None),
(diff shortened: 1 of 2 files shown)


nabobalis:
I want to check its behaviour on RTD first.

Cadair:
Does this work on the preview builds, it seems to not be working for me?

nabobalis:
TODO:

See if it's possible to disable the pop up fully for refs or URLs.

docs/conf.py
```diff
@@ -212,12 +212,18 @@
...
 
-hoverxref_tooltip_maxwidth = 600  # RTD main window is 696px
 hoverxref_auto_ref = True
-hoverxref_mathjax = True
-# hoverxref has to be applied to these
 hoverxref_domains = ["py"]
-
+hoverxref_mathjax = True
+hoverxref_modal_hover_delay = 500
+hoverxref_tooltip_maxwidth = 600  # RTD main window is 696px
```
nabobalis, line 219: The default is 450, if we want to go narrower.

docs/conf.py
```diff
@@ -212,14 +212,34 @@
         # A RTD thing to avoid a CSRF block when docs are using a custom domain
         hoverxref_api_host = "/_"
 
-hoverxref_tooltip_maxwidth = 600  # RTD main window is 696px
-hoverxref_auto_ref = True
-hoverxref_mathjax = True
-# hoverxref has to be applied to these
+hoverxref_auto_ref = False
```
nabobalis, line 215: I had to disable this, I can't get it to not work on gallery examples.

Either I am missing a config value or we need to consider a patch upstream.

nabobalis: I think if we could change: https://github.com/readthedocs/sphinx-hoverxref/blob/b4b5cf4ae95cbb78151509ed46b5b0c5f821ec58/hoverxref/domains.py#L60 to string match instead, we could add the generic gallery text.

Cadair:
I want to see this built before approving, especially as it's hard/impossible to make hover stuff work locally.

docs/conf.py
```diff
@@ -212,14 +212,34 @@
...
 hoverxref_domains = ["py"]
-
+hoverxref_mathjax = True
+hoverxref_modal_hover_delay = 500
+hoverxref_tooltip_maxwidth = 600  # RTD main window is 696px
+hoverxref_intersphinx = [
```
Cadair, line 220: Is this just a list of all our intersphinx mappings or have you curated it?

docs/conf.py
```diff
@@ -212,14 +212,34 @@
...
+    "astropy",
+    "matplotlib",
```
Cadair, line 222: does mpl work as it's not on rtd?

docs/conf.py
```diff
@@ -212,14 +212,34 @@
...
+    "numpy",
+    "python",
+    "scipy",
+    "hvpy",
+    "asdf",
+    "zeep",
```
Cadair, line 228: Do we really need this?

docs/conf.py
```diff
@@ -212,14 +212,34 @@
...
+    "sunkit_instruments",
+    "sunkit_image",
+    "sqlalchemy",
```
Cadair, line 231: and this

docs/conf.py
nabobalis: The first one.

docs/conf.py
```diff
@@ -229,13 +249,12 @@
     "meth": "tooltip",
     "mod": "tooltip",
     "obj": "tooltip",
-    # roles with std domain
+    # Roles within the standard domain
```
Cadair, line 252: the std domain is a *thing*

```suggestion
    # Roles within the std domain
```

docs/conf.py
nabobalis: They special case several of them. Scipy, Numpy and MPL should all work as they support that manually.

docs/conf.py
nabobalis: Why not? Surely we want this to work for links that are API and we link to zeep classes.

docs/conf.py
nabobalis: I will accept after RTD builds.

docs/conf.py
Cadair: If we are going to do that then we can just do `list(intersphinx_mapping.keys())` or whatever, but should we curate it?

nabobalis: Fair, I can do that instead. 

I personally think if we intersphinx, we would want them all to hover when we link to those classes.

Cadair (approved):

PR #6779, opened 2023-02-13
Add PROBA-2 Client

### LINKED ISSUE #5804: Add ability to download Proba-2 in-situ DSLP data
Originally opened by @joncook at https://github.com/heliopython/heliopy/issues/945 - I'm transferring this to sunpy.

Hello @dstansby - as mentioned here are the details for accessing Proba-2 in-situ DSLP data:

An example query:
- http://p2sa.esac.esa.int/p2sa-sl-tap/tap/sync?REQUEST=doQuery&LANG=ADQL&FORMAT=JSON&PHASE=RUN&QUERY=SELECT * FROM p2sa.v_file WHERE  (file_date >= '2014-10-01 00:00:00')   AND   (file_date <= '2014-10-31 00:00:00')   AND   (file_type  in  ('DSLP_DERIVED'))  ORDER BY file_date ASC

There is also the file_type 'DSLP_CALIBRATED' which I presume are some calibration files. Not sure if they are also of interest.

In both cases, this will give you a vo-table json file with the metadata required to then download the files:

```
{
  "metadata": [
    {
      "name": "calibrated",
      "datatype": "BOOLEAN"
    },
    {
      "name": "extension",
      "datatype": "VARCHAR"
    },
    {
      "name": "file_date",
      "datatype": "TIMESTAMP",
      "xtype": "adql:TIMESTAMP"
    },
    {
      "name": "file_name",
      "datatype": "VARCHAR"
    },
    {
      "name": "file_oid",
      "datatype": "INTEGER"
    },
    {
      "name": "file_path",
      "datatype": "VARCHAR"
    },
    {
      "name": "file_size",
      "datatype": "BIGINT"
    },
    {
      "name": "file_type",
      "datatype": "VARCHAR"
    },
    {
      "name": "instrument_name",
      "datatype": "VARCHAR"
    },
    {
      "name": "instrument_oid",
      "datatype": "INTEGER"
    },
    {
      "name": "processing_level",
      "datatype": "VARCHAR"
    }
  ],
  "data": [
    [
      "false",
      "CDF",
      "2014-10-01 00:00:00.0",
      "DERIVED_SLPA_20141001.cdf",
      "136478",
      "\\/p2sa\\/p2saps\\/science\\/dslp\\/dslp_derived\\/2014\\/10",
      "19815408",
      "DSLP_DERIVED",
      "DSLP",
      "3",
      "N\\/A"
    ],
    [
      "false",
      "CDF",
      "2014-10-01 00:00:00.0",
      "DERIVED_SLPB_20141001.cdf",
      "136479",
      "\\/p2sa\\/p2saps\\/science\\/dslp\\/dslp_derived\\/2014\\/10",
      "19788276",
      "DSLP_DERIVED",
      "DSLP",
      "3",
      "N\\/A"
    ],
...
```
I think you can ignore most of it and if you prefer only select on the fields you want in the original query which will reduce the size of the json if you see what I mean.

You can then use the file_oid or the file_name to download the data. The file_name is probably more obvious to the naked eye e.g:

- http://p2sa.esac.esa.int/p2sa-sl-tap/data?retrieval_type=PRODUCT&QUERY=SELECT+*+FROM+p2sa.file+where+file_oid++in+%28136483%29

- http://p2sa.esac.esa.int/p2sa-sl-tap/data?retrieval_type=PRODUCT&QUERY=SELECT+*+FROM+p2sa.file+where+file_name='DERIVED_SLPB_20141001.cdf'

The DSLP data is between 2009 and 2016.

Any doubts, please let me know. You have my contact details. I wouldn't mind having a go at this one but probably won't have time in the next couple of weeks.

Thanks
Jonathan

### Notes

Please be aware that this is a TAP protocol server, this means that the current Clients in sunpy do not work for this directly. https://github.com/sunpy/sunpy-soar/blob/main/sunpy_soar/client.py does implement a client that can interact with this API, so that can be used as an example.

exitflynn:
## PR Description
Adds a client that provides support for downloading Proba-2 in-situ DSLP data and closes #5804.

[This is what an example response from the server looks like.](https://gist.github.com/exitflynn/ffa486170b93091e244ebc4c4e6c2469)


exitflynn:
here's a quick example of how the response looks like which i felt would be relevant: https://gist.github.com/exitflynn/ffa486170b93091e244ebc4c4e6c2469

exitflynn:
i'm all good now and have been able to make some progress over the week on this. the progress hasn't been as fast on this since it's a bit of a learning curve here, however there are some places where i could use some help/input:
1. I'm mainly having trouble with figuring out how to go from a common `requests.json()` to a table response. It'd be tremendously helpful to be pointed to any example in the codebase (/documentation).
2. The data has fields like processing level (you can see that in the [linked](https://gist.github.com/exitflynn/ffa486170b93091e244ebc4c4e6c2469) example response) but since it's always "N/A", should i drop that as an attr?

nabobalis:
> 1. I'm mainly having trouble with figuring out how to go from a common `requests.json()` to a table response. It'd be tremendously helpful to be pointed to any example in the codebase (/documentation).

I am not aware of one in the codebase but you'd need to check each client first.
Otherwise, https://docs.astropy.org/en/stable/table/construct_table.html#examples details how to create tables from various data forms, so hopefully that helps.

> 2. The data has fields like processing level (you can see that in the [linked](https://gist.github.com/exitflynn/ffa486170b93091e244ebc4c4e6c2469) example response) but since it's always "N/A", should i drop that as an attr?

If it is **always** going to be N/A, we should drop it. Do we know if it will always be N/A?

exitflynn:
Thank you for that link to the astropy table docs. It helped tremendously! Since the PROBA2 servers are out of maintenance now, I've made some progress and would like some input...

*forgive the length of the reply it's pretty straightforward :sob:*

the DSLP data is from 2009-11-18 to 2016-03-31. At once, it returns only 1000 queries. Also complex SQL queries involving `GROUP_BY` or aggregate functions of any kind aren't supported so I had to rely on manually going through the data in roughly sets of two years by changing the years in the file_date queries which is how I learned the previous information and that the `processing_level` column is indeed **always N/A.**

A quick script to verify:
```python
for year in range(2009, 2016, 2):
    response = requests.get(f"http://p2sa.esac.esa.int/p2sa-sl-tap/tap/sync?REQUEST=doQuery&LANG=ADQL&FORMAT=JSON&PHASE=RUN&QUERY=SELECT * FROM p2sa.v_file WHERE (file_date >= '{year}-01-01 00:00:00') AND (file_date <= '{year+2}-01-01 00:00:00') AND (file_type in ('DSLP_DERIVED')) ORDER BY file_date ASC")
    # print(response.json())
    data = response.json()['data']
    print(len([row for row in data if row[10] != 'N\\/A']))
```

There are following fields which are supported (in this order, would help with indexes in records if anyone wants to verify):
`calibrated`, `extension`, `file_date`, `file_name`, `file_oid`, `file_path`, `file_size`, `file_type`, `instrument_name`, `instrument_oid`, `processing_level`.

out of which the following have constant values:
`calibrated`: 'false'
`extension`: 'CDF'
`file_type`: 'DSLP_DERIVED'
`instrument_name`: 'DSLP'
`instrument_oid`: 3
`processing_level`: 'N\\/A'


I have the following questions:
- This leaves us with `file_date`, `file_name`, `file_oid`, `file_path` and `file_size` for unique values out of which I think we don't need to make a new attr out of? (except `file_date` which is already done with `a.Time`) Should there be support for any one of these as an attr to search using?
- another point of discussion is which of these should show up in the `QueryResponseTable` response? (since they'll all be redundant and repeating the same value) Currently every field is included.
- also the answer to these questions could depend upon whether we want to use this source for the other instruments' data it provides, namely SWAP and LYRA.

nabobalis:
> I have the following questions:
> 
> * This leaves us with `file_date`, `file_name`, `file_oid`, `file_path` and `file_size` for unique values out of which I think we don't need to make a new attr out of? (except `file_date` which is already done with `a.Time`) Should there be support for any one of these as an attr to search using?

I don't know why anyone would want to search by anything other than date. But I am not familiar with the data.

> * another point of discussion is which of these should show up in the `QueryResponseTable` response? (since they'll all be redundant and repeating the same value) Currently every field is included.

I don't see the harm in adding them to the return. We normally return most information that we get from a provider. 

> * also the answer to these questions could depend upon whether we want to use this source for the other instruments' data it provides, namely SWAP and LYRA.

I think we already have a LYRA client. Whether we need to update it is another question.

nabobalis:
> i think the source of my troubles has been maybe that the server has been down (/buggy) for a while? it was down for maintenance once and i think this has been the issue ever since the server's been back up, all the data requests for example something like http://p2sa.esac.esa.int/p2sa-sl-tap/data?retrieval_type=PRODUCT&QUERY=SELECT+*+FROM+p2sa.file+where+file_name=%27DERIVED_SLPB_20141001.cdf (as mentioned in the original issue's description) return Internal Server Error, which was working at the time of the issue's creation and during January this year as well. I've made other data requests going by the documentation as well and have received similar results.

If the server is changing or broken, it might be better to stop working on this and move on to another issue.

nabobalis:
Thanks again @exitflynn for working on this but between the server changing and the fact that we don't have a TAP client in sunpy for now. It means working on this actually requires some deeper levels sunpy to occur first. 

Due to these factors, I will close this PR

JonCook:
Hello @exitflynn and @nabobalis - I came across this issue after checking the original recently #5804. I work at ESA and am one of the responsible people for the Proba-2 archive which you are trying to access. It was just to clarify that the server certainly isn't buggy or down. It's an operational server running on ESA infrastructure. 

It is simply that for security reasons, you are no longer allowed to use the * in the select clause of a data download request. You should go for something like this:

https://p2sa.esac.esa.int/p2sa-sl-tap/data?retrieval_type=PRODUCT&QUERY=SELECT+file_path,+file_name+FROM+p2sa.v_file+%20WHERE+file_name+=%27DERIVED_SLPB_20141001.cdf%27

I agree the error message in this case is not very helpful, but please check the examples at:
* http://p2sa.esac.esa.int/p2sa/#aio

Or please do not hesitate to reach out for help,

Hoping you can now make progress on integrating this data into sunpy

Many Thanks,
Jonathan

nabobalis:
Ah that would explain it, my unfamiliarly with TAP was the problem.

Thanks for explaining that @JonCook

JonCook:
No problem @nabobalis - just to add Proba2 data is also available via the virtual observatory (via ESA TAP) e.g
* https://sdac.virtualsolar.org/cgi/search?time=1&instrument=1&version=current&build=1

Therefore maybe it already appears via SunPy when searching the VSO?

PR #6780, opened 2023-02-16
Add pandas to the devdeps

nabobalis:
See if we break due to the upcoming major 2.0 release.


dstansby (approved):
LGTM - I checked that a dev version of pandas was installed on the devdeps build.

PR #6771, opened 2023-02-08
Add Example to plot a the path of PSP around the Sun.

Jett-Code:
## PR Description

Fixes #6301

PR #6781, opened 2023-02-16
Add example for reproducing the "Where is STEREO Today?" plot

### LINKED ISSUE #6301: Reproduce the "Where is STEREO?" plot in SunPy
### Describe the feature

The STEREO webpage has a handy figure showing where STEREO and other spacecraft are: https://stereo-ssc.nascom.nasa.gov/where.shtml . Given the existing examples in the SunPy gallery it seems that this should be reproducible using SunPy.

### Proposed solution

The solution would use a mix of existing examples that use the ability to get locations of spacecraft and planets.

ayshih:
Although we already closed #6301 with #6771, this PR adds an example that literally reproduces the "Where is STEREO Today?" plot as requested by #6301.  Since this example is complex and particular, it is not intended to be a tutorial or a how-to guide.  I have put it into a newly created "Showcase" category of examples, which I envision could house other complex and particular examples.


ayshih:
Bleh, the axes text color isn't coming out correct on Matplotlib 3.7.0, but is correct on Matplotlib 3.6.1.  I'll have to track that down.

nabobalis:
Got to admit, I do like the concept of a showcase.

ayshih:
The example [HMI Showcase: Cutout](https://docs.sunpy.org/en/stable/generated/gallery/plotting/hmi_cutout.html) would also be a good candidate to move to this new "Showcase" category

nabobalis:
> The example [HMI Showcase: Cutout](https://docs.sunpy.org/en/stable/generated/gallery/plotting/hmi_cutout.html) would also be a good candidate to move to this new "Showcase" category

Lets move it.

Jett-Code:
> Although we already closed #6301 with #6771, this PR adds an example that literally reproduces the "Where is STEREO Today?" plot as requested by #6301. Since this example is complex and particular, it is not intended to be a tutorial or a how-to guide. I have put it into a newly created "Showcase" category of examples, which I envision could house other complex and particular examples.

Since the connection with STEREO-B was lost so is it good to show the location of STEREO-B in the plot?

nabobalis:
> Since the connection with STEREO-B was lost so is it good to show the location of STEREO-B in the plot?

The goal is to replicate the figure as close as possible and that includes STEREO-B despite that fact its lost to us.

nabobalis:
`<unknown>:837: WARNING: undefined label: 'sphx_glr_generated_gallery_plotting_hmi_cutout.py'`  a reference somewhere that needs updating due to the move.

examples/showcase/README.txt
```diff
@@ -0,0 +1,4 @@
+Showcase
+========
+
+Examples that use an advanced combination of capabilities in sunpy
```
dstansby, line 4:
```suggestion
Examples that use an advanced combination of capabilities in sunpy.
```

examples/showcase/where_is_stereo.py
```diff
@@ -0,0 +1,154 @@
...
+stereo_a = get_horizons_coord('STEREO-A', obstime)
+stereo_b = get_horizons_coord('STEREO-B', obstime)
+
+missions = ['Parker Solar Probe', 'Solar Orbiter', 'BepiColombo']
+mission_labels = {'Parker Solar Probe': 'PSP', 'Solar Orbiter': 'SO', 'BepiColombo': 'BEPICOLOMBO'}
+periods = [100, 180, 120]  # may need to be updated as orbits evolve
+mission_coords = {mission: get_horizons_coord(mission, {'start': obstime,
+                                                        'stop': obstime + period * u.d,
+                                                        'step': '1d'})
+                  for mission, period in zip(missions, periods)}
```
dstansby, line 51:
```suggestion
# Times to sample each orbit at from today, chosen to span at least a full
# orbit of every body plotted
times = np.arange(1000) * u.day
planets = ['Mercury', 'Venus', 'Earth', 'Mars']
planet_coords = {planet: get_body_heliographic_stonyhurst(planet,
                                                          obstime + times)
                 for planet in planets}

stereo_a = get_horizons_coord('STEREO-A', obstime)
stereo_b = get_horizons_coord('STEREO-B', obstime)

missions = ['Parker Solar Probe', 'Solar Orbiter', 'BepiColombo']
mission_labels = {'Parker Solar Probe': 'PSP', 'Solar Orbiter': 'SO', 'BepiColombo': 'BEPICOLOMBO'}
mission_coords = {mission: get_horizons_coord(mission, {'start': obstime,
                                                        'stop': obstime + times,
                                                        'step': '1d'})
                  for mission in missions}
```

I got very confused about periods for a while here, so would suggest the above to simplify and clarify things.

examples/showcase/where_is_stereo.py
```diff
@@ -0,0 +1,154 @@
...
+    ax.plot(*coord_to_heexy(coord), linestyle='dashed', color='gray')
+
+    if planet == 'Earth':
+        color, markersize = 'lime', 10
+    else:
+        color, markersize = 'gray', None
+
+    x, y = coord_to_heexy(coord[0])
+    ax.plot(x, y, 'o', markersize=markersize, color=color)
+    ax.text(x + (0.1 if planet == 'Earth' else 0.05), y, planet, color=color)
```
dstansby, line 119:
```suggestion
    if planet == 'Earth':
        color, markersize, offset = 'lime', 10, 0.1
    else:
        color, markersize, offset = 'gray', None, 0.05

    x, y = coord_to_heexy(coord[0])
    ax.plot(x, y, 'o', markersize=markersize, color=color)
    ax.text(x + offset, y, planet, color=color)
```

examples/showcase/where_is_stereo.py
```diff
@@ -0,0 +1,154 @@
...
+
+# sphinx_gallery_defer_figures
+
+##############################################################################
+# Draw the STEREO spacecraft (sans orbits), as well as Sun-STEREO lines.
```
dstansby, line 124:
```suggestion
# Draw the STEREO spacecraft (without orbits), as well as Sun-STEREO lines.
```

dstansby (commented):
Some suggestions to make code a bit clearer - otherwise looks 👍 , and I like the idea of a showcase gallery section

examples/showcase/where_is_stereo.py
ayshih: Weird, I don't normally ever use the word "sans", so I don't know why I did in this case...

examples/showcase/where_is_stereo.py
ayshih: Hmm, I don't like the idea of plotting multiple orbits just to make the code simpler.  It won't matter much for planets, but it could look bad for spacecraft.

An alternative approach that'd be more robust than hard-coding orbital periods, but would mean even more code, is to grab 1000 days as you have done here, but then figure out how many elements make up the first orbit (by looking for when the ecliptic longitude passes the starting point) and then plotting just that first orbit.

dstansby: Ah fair enough, I'm guessing it would look bad because of the dashed lines? In that case hardcoding is probably best, but maybe with a clarifying comment along the lines of `# Orbital period in days` above where `periods` is defined?

ayshih: > Ah fair enough, I'm guessing it would look bad because of the dashed lines?

Not just that, but orbital precession means that successive orbits of spacecraft – most notably BepiColombo – do not lie on top of each other.

I've added a function to automatically determine the first full orbit in a trajectory.

examples/showcase/where_is_stereo.py
ayshih: Done

examples/showcase/where_is_stereo.py
ayshih: Done

examples/showcase/where_is_stereo.py
```diff
@@ -0,0 +1,171 @@
...
+def coord_to_heexy(coord):
+    coord = coord.transform_to(hee_frame)
+    coord.representation_type = 'cartesian'
+    return coord.y.to_value('AU'), coord.x.to_value('AU')
+
+##############################################################################
+# Set Matplotlib settings to the desired appearance and initialize the axes.
+
+
+mpl.rcParams.update({'figure.facecolor': 'black',
```
Cadair, line 85: Oh god, why are we making it so ugly! :laughing:

dstansby (approved):

nabobalis (approved):

PR #6786, opened 2023-02-19
Added a line to switch astropy ephemeris via astropy.coordinates.solar_system_ephemeris

Jett-Code:
## PR Description

 Fixes #3114

PR #6797, opened 2023-02-24
Wrong ref type for class solar_system_ephemeris

nabobalis:
I didn't notice the doc failure from when I merged: https://github.com/sunpy/sunpy/pull/6786

### DIFF
--- a/sunpy/coordinates/ephemeris.py
+++ b/sunpy/coordinates/ephemeris.py
@@ -64,7 +64,7 @@ def get_body_heliographic_stonyhurst(body, time='now', observer=None, *, include
     apparent location of the body will shift in tandem with the Sun.
 
     For planets other than Earth, Astropy's built-in ephemeris is not as accurate as JPL
-    ephemerides, so one can use `~astropy.coordinates.solar_system.solar_system_ephemeris` to switch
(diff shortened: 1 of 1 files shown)


sunpy/coordinates/ephemeris.py
```diff
@@ -64,7 +64,7 @@ def get_body_heliographic_stonyhurst(body, time='now', observer=None, *, include
     apparent location of the body will shift in tandem with the Sun.
 
     For planets other than Earth, Astropy's built-in ephemeris is not as accurate as JPL
-    ephemerides, so one can use :func:`astropy.coordinates.solar_system_ephemeris` to switch
+    ephemerides, so one can use `~astropy.coordinates.solar_system.solar_system_ephemeris` to switch
```
ayshih, line 67:
```suggestion
    ephemerides, so one can use `astropy.coordinates.solar_system_ephemeris` to switch
```

ayshih (commented):
Adding the `:func:` was my mistake, but I feel the full namespace should be shown (no `~`) and that the reference should point to one level up because that it where it is documented on the Astropy side.

ayshih (approved):
Docs are good now

PR #6803, opened 2023-02-27
Update ASV environment 

nabobalis:
Forgot to account for opencv.
Updated the tag to point to 3.0 instead of 1.0

### DIFF
--- a/.github/workflows/asv-regular.yml
+++ b/.github/workflows/asv-regular.yml
@@ -14,7 +14,7 @@ on:
 jobs:
   asv-run:
     if: ${{ github.repository == 'sunpy/sunpy' }}
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout sunpy repo
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/asv-regular.yml
```diff
@@ -33,13 +33,14 @@ jobs:
           architecture: "x64"
       - name: Install dependencies
         run: |
+          sudo apt-get install -y libopencv-dev
```
nabobalis, line 36: Not sure if we need this but there is no harm installing it.

.github/workflows/asv-regular.yml
```diff
@@ -33,13 +33,14 @@ jobs:
...
           python -m pip install --upgrade pip
           pip install asv virtualenv
       - name: Add machine info for ASV
         run: asv machine --machine GH-Actions --os ubuntu-20.04 --arch x64 --cpu "2-core unknown" --ram 7GB
       - name: Run benchmarks for commits since v1.1
-        # On main the release of v1.1 is the 2.0dev tag
-        run: taskset -c 0 asv run --skip-existing-successful v2.0.dev..
+        # On main the release of v3.0 is the 4.0dev tag
```
ayshih, line 42: I'm confused: Doesn't the `v4.0dev tag` point to 3.1, not 3.0?

nabobalis: Oh right I'm wrong. That's not how it works.

 I'll fix it

ayshih (approved):
It looks fine to me, so go go go?

nabobalis:
Let us see how this one goes: https://github.com/sunpy/sunpy/actions/runs/4289504002

Issue #6801, opened 2023-02-27
Great Arc example should use `extract_along_coord`


wtbarnes:
[The example showing how to construct a Great Arc](https://docs.sunpy.org/en/stable/generated/gallery/plotting/great_arc_example.html#sphx-glr-generated-gallery-plotting-great-arc-example-py) also shows how to get the intensity along that arc. Since this example was written, we added a function for extracting the value of a map along a coordinate as shown in [this example](https://docs.sunpy.org/en/stable/generated/gallery/units_and_coordinates/map_slit_extraction.html#sphx-glr-generated-gallery-units-and-coordinates-map-slit-extraction-py). The great arc should use that functionality rather than implementing the intensity extraction itself.

davnish:
Should I create a PR for this issue? Following the example gallery dev guidelines.

PR #6816, opened 2023-03-01
Use extract_along_coord in arc example

dstansby:
Does what it says in the title. Fixes https://github.com/sunpy/sunpy/issues/6801

### DIFF
--- a/examples/plotting/great_arc_example.py
+++ b/examples/plotting/great_arc_example.py
@@ -7,15 +7,16 @@ How to define and draw a great arc on an image of the
 Sun, and to extract intensity values along that arc.
 """
 import matplotlib.pyplot as plt
-import numpy as np
 
 import astropy.units as u
 from astropy.coordinates import SkyCoord
+from astropy.visualization import quantity_support
 
 import sunpy.map
 from sunpy.coordinates.utils import GreatArc
 from sunpy.data.sample import AIA_171_IMAGE
 
+quantity_support()
 ###############################################################################
 # We start with the sample data.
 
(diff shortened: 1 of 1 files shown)


examples/plotting/great_arc_example.py
```diff
@@ -7,7 +7,6 @@
 Sun, and to extract intensity values along that arc.
 """
 import matplotlib.pyplot as plt
-import numpy as np
 
 import astropy.units as u
 from astropy.coordinates import SkyCoord
```
wtbarnes, line 12:
```suggestion
from astropy.coordinates import SkyCoord
from astropy.visualization import quantity_support
```

examples/plotting/great_arc_example.py
```diff
@@ -43,28 +42,18 @@
...
 ###############################################################################
 # Plot the intensity along the arc from the start to the end point.
 
-plt.plot(angles, intensity_along_arc)
-plt.xlabel(f'Distance along arc [{angles.unit}]')
-plt.ylabel(f'Intensity [{m.unit}]')
+fig, ax = plt.subplots()
+ax.plot(separation, intensity)
+ax.set_xlabel(f'Separation from start of arc [{separation.unit}]')
+ax.set_ylabel(f'Intensity [{m.unit}]')
```
wtbarnes, line 57: Following our example gallery style guidelines for plotting: https://docs.sunpy.org/en/latest/dev_guide/contents/example_gallery.html#contribution-guidelines
```suggestion
plt.figure()
with quantity_support():
    plt.plot(separation, intensity)
```

wtbarnes (changes requested):

examples/plotting/great_arc_example.py
wtbarnes: Our example gallery dev guidelines say that you shouldn't create an axes instance if you don't need to. I would say that adding axes labels does not necessitate the creation of an axes instance.

dstansby: I've opened https://github.com/sunpy/sunpy/pull/6822 to discuss this, and in the meantime removed the `Axes` object in this PR.

examples/plotting/great_arc_example.py
dstansby: I've put in quantity support, but left the manual axes labelling in because I think it adds useful information to the plot

examples/plotting/great_arc_example.py
```diff
@@ -43,28 +44,18 @@
...
 
 ###############################################################################
 # Plot the intensity along the arc from the start to the end point.
 
-plt.plot(angles, intensity_along_arc)
-plt.xlabel(f'Distance along arc [{angles.unit}]')
+plt.figure()
+plt.plot(separation, intensity)
+plt.xlabel(f'Separation from start of arc [{separation.unit}]')
 plt.ylabel(f'Intensity [{m.unit}]')
```
wtbarnes, line 59:
```suggestion
plt.ylabel(f'Intensity [{intensity.unit}]')
```

examples/plotting/great_arc_example.py
wtbarnes: That makes sense. But then I don't think importing quantity_support is needed.

wtbarnes (approved):

nabobalis (approved):

Issue #6815, opened 2023-03-01
Explain why `extract_along_coords` returns coordinates


dstansby:
### Describe the feature

[`extract_along_coord`](https://docs.sunpy.org/en/stable/generated/api/sunpy.map.extract_along_coord.html#sunpy.map.extract_along_coord) takes coordinates to get the value of a map at, but also returns a set of coordinates. I presume these are not the same coordinates, but the docstring does not explain what they are or why they're useful. This is a request to add input and output parameter documentation to the docstring of `extract_along_coord`.

Cadair:
The returned coordinates are the coordinates for the pixels from which the values were taken, as the function walks the pixels. Also it returns the same number of coordinates as values, so you can plot them or whatever you want.

dstansby:
Had a play around, and I think putting a version of this in the docstring would help:

![extract_along_coord](https://user-images.githubusercontent.com/6197628/222276584-a195b528-96e5-43df-986e-f0d2697e9095.png)


```python
import matplotlib.pyplot as plt

import astropy.units as u
from astropy.coordinates import SkyCoord

import sunpy.map
from sunpy.data.sample import AIA_171_IMAGE

m = sunpy.map.Map(AIA_171_IMAGE)

coords = SkyCoord(
    Tx=[715, 720, 731] * u.arcsec,
    Ty=[-360, -350, -347] * u.arcsec,
    frame=m.coordinate_frame,
)

intens, intens_coords = sunpy.map.extract_along_coord(m, coords)

fig = plt.figure()
ax = fig.add_subplot(111, projection=m)
m.plot(ax)
ax.plot_coord(coords, marker='o', color='black', label="Input coords")
ax.plot_coord(intens_coords, marker='s', color='tab:purple', label="Sampled coords")
ax.set_ylim(360, 360 + 10)
ax.set_xlim(807, 807 + 10)
ax.legend()
plt.show()
```

dstansby:
🤔 also the picture above is now making me suspicious about how `extract_along_coord` works, there are clearly some pixels the black line goes through that aren't sampled in the output...

ayshih:
Some pixels will always be "skipped", but it appears that the implementation of Bresenham's line algorithm is slightly bugged, because it's not picking the correct pixels.

ayshih:
See #6817 for the bug fix.  The correct result is:
[...]

PR #6817, opened 2023-03-02
Fixed a bug with the wrong pixels being used by `extract_along_coord()`

ayshih:
`extract_along_coord()` uses Bresenham's line algorithm to select the closest pixels to the line, but there is a bug in the implementation: due to the wrong use of `>` instead of `>=`, some wrong pixels were being used ~when the slope was exactly +/-2 or +/-1/2~ for virtually all slopes.

The following images are taken from the discussion in #6815:

Before this PR (two incorrect pixels were chosen):
222276584-a195b528-96e5-43df-986e-f0d2697e9095

After this PR:
222304190-fec51d51-6f3b-4697-bc18-50da21108e94

### DIFF
--- a/changelog/6817.bugfix.rst
+++ b/changelog/6817.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug with :func:`~sunpy.map.extract_along_coord` where some of the pixels sampled were not in fact the closest pixels to the coordinate line segments.
--- a/sunpy/map/maputils.py
+++ b/sunpy/map/maputils.py
@@ -443,10 +443,10 @@ def _bresenham(*, x1, y1, x2, y2):
         if (x == x2) and (y == y2):
             break
         e2 = 2 * err
-        if e2 > -dy:
+        if e2 >= -dy:
             err = err - dy
             x += sx
-        if e2 < dx:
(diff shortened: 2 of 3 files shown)


nabobalis:
Do we want to add a figure test?

ayshih:
Yeah, I certainly plan to add a (non-figure) test

ayshih:
Upon writing the comprehensive unit test, it became clear that wrong pixels were being picked for significantly more slopes than I had initially thought.  Anyway, all fixed now.

nabobalis (approved):

ayshih:
Hold pending some investigation

ayshih:
I was mistaken, so I am closing this PR.  This "bug" is simply a difference in choice when here are two pixels equally close to the line connecting the center of the start pixel and the center of the end pixel.  The reason why it appears to fix the plots in #6815 is because the coordinates are not exactly at the centers of pixels, so one choice appears to match better than the other choice.

ayshih:
If the coordinates were at pixel centers, the PR doesn't make a necessary change.  In the plot below, the solid purple line is the current implementation, and the dashed purple line is after this PR.
Figure_1

PR #6822, opened 2023-03-06
Always create Axes objects in examples

dstansby:
Our rules of thumb for gallery examples exist to "minimize verbosity and maintain consistency with the other gallery examples:". In some places consistency and verbosity are a trade off, including which flavour of the Matplotlib API we are calling.

Broadly the two options are:
1. Use `pyplot` (e.g. `plt.xlabel`) where possible, and do not create Axes objects
2. Use `Axes` all the time, even when it would be possible to use `pyplot`

1. is a bit less verbose than 2., but this comes at the cost of having two ways of doing stuff in our gallery (e.g. `plt.xlabel()` vs. `ax.set_xlabel()`. In my opinion consistency is more important than minimizing verbosity here, so I am proposing that we always create an `Axes` object.

I've converted one gallery example as an example of what this change would look like. Thoughts?

### DIFF
--- a/docs/dev_guide/contents/example_gallery.rst
+++ b/docs/dev_guide/contents/example_gallery.rst
@@ -37,7 +37,7 @@ Contribution Guidelines
 
   * Always create a figure instance using ``plt.figure()`` prior to creating a plot.
 
-  * Only create an axes instance if it is explicitly needed later in the example (e.g. when overplotting a coordinate on a map using ``ax.plot_coord``).
+  * Always create an Axes instance, and where possible use this to modify the plot instead of using the ``pyplot`` interface (for example use ``ax.set_label()``` instead of ``plt.xlabel()``).
 
   * If an axes instance is created, it should be explicitly passed as a keyword argument wherever possible (e.g. in `~sunpy.map.GenericMap.plot` or `~sunpy.map.GenericMap.draw_grid`).
 
--- a/docs/dev_guide/contents/example_gallery.rst
+++ b/docs/dev_guide/contents/example_gallery.rst
@@ -37,7 +37,7 @@ Contribution Guidelines
 
   * Always create a figure instance using ``plt.figure()`` prior to creating a plot.
 
-  * Always create an Axes instance, and where possible use this to modify the plot instead of using the ``pyplot`` interface (for example use ``ax.set_label()``` instead of ``plt.xlabel()``).
+  * Always create an Axes instance, and where possible use this to modify the plot instead of using the ``pyplot`` interface (for example use ``ax.set_xlabel()``` instead of ``plt.xlabel()``).
 
   * If an axes instance is created, it should be explicitly passed as a keyword argument wherever possible (e.g. in `~sunpy.map.GenericMap.plot` or `~sunpy.map.GenericMap.draw_grid`).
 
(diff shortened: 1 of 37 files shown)


ayshih:
I usually favor avoiding the `pyplot` interface, but I'll point out that the non-`pyplot` alternative to a plain `plt.colorbar()` call is notably more clunky.

Cadair:
I am generally in favour of this, it generally works nicer with WCSAxes and our custom plotting wrappers.

wtbarnes:
I don't really have an opinion here as long as we are consistent across the whole gallery.

examples/map/image_bright_regions_gallery_example.py
```diff
@@ -35,8 +35,9 @@
 
 aiamap_mask.mask = mask
 
-plt.figure()
-aiamap_mask.plot()
+fig = plt.figure()
+ax = fig.add_subplot(projection=aiamap_mask)
```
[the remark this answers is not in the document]
dstansby: I think we could do `fig, ax = plt.subplots(subplot_kw={'projection': aiamap})` instead?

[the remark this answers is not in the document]
Cadair: yeah that's worse :rofl: 

I more meant having to manually specify the projection at all, but I guess "explicit is better than implicit".

docs/dev_guide/contents/example_gallery.rst
```diff
@@ -37,7 +37,7 @@ Contribution Guidelines
 
   * Always create a figure instance using ``plt.figure()`` prior to creating a plot.
 
-  * Only create an axes instance if it is explicitly needed later in the example (e.g. when overplotting a coordinate on a map using ``ax.plot_coord``).
+  * Always create an Axes instance, and where possible use this to modify the plot instead of using the ``pyplot`` interface (for example use ``ax.set_label()``` instead of ``plt.xlabel()``).
```
wtbarnes, line 40:
```suggestion
  * Always create an Axes instance, and where possible use this to modify the plot instead of using the ``pyplot`` interface (for example use ``ax.set_xlabel()``` instead of ``plt.xlabel()``).
```

examples/computer_vision_techniques/off_limb_enhance.py
```diff
@@ -49,18 +49,19 @@
 # Let's plot the results using LaTeX for all the text.
 
 fontsize = 14
-plt.plot(rsun_array, y, label='data')
+fig, ax = plt.subplots()
```
wtbarnes, line 52:
```suggestion
fig = plt.figure()
ax = fig.add_subplot()
```

My initial reaction was to make this replacement everywhere, but I don't think the added verbosity is worth the relatively minor gain in consistency.

wtbarnes (approved):
I initially started replacing `plt.subplots` with `fig = plt.figure(); ax = fig.add_subplot()` but ran out of steam. I don't particularly like the inconsistency in the ways we create the figure, axes objects, but I do not think it is worth the added verbosity.

examples/computer_vision_techniques/off_limb_enhance.py
nabobalis: We value consistency!

nabobalis (approved):

PR #6828, opened 2023-03-09
Fix numpy deprecation warnings

dstansby:
Should fix the figure devdeps runs.


ayshih:
We probably need to add the deprecation warning to our ignore list because h5py is triggering them in our cron devdeps build

PR #6836, opened 2023-03-11
Ruff

Cadair:
I don't quite know how I ended up here...


pyproject.toml
```diff
@@ -83,3 +83,81 @@ test-requires = ["opencv_python==4.6.0.66"]
       directory = "trivial"
       name = "Internal Changes"
       showcontent = true
+
+
+[tool.ruff]
+target-version = "py38"
```
nabobalis, line 89:
```suggestion
target-version = "py39"
```

ayshih:
I also don't like the changes to `_supported_observer_coordinates()` in the map sources

nabobalis:
> I also don't like the changes to `_supported_observer_coordinates` in the map sources

I assume that is a side effect of ignoring the line length which looks like it has  TODO attached to?

ayshih:
Sure.  It's amusing, albeit understandable, that telling an automatic code formatter to ignore line length gives it license to make heinously long lines.

Cadair:
I don't think ruff autofixes line too long anyway.

nabobalis:
With the three ruff PRS merged, rebasing this might be a pain and a fresh PR with extra config should be opened.

PR #6837, opened 2023-03-12
Fits reading emits a warning when the data array is not actually memory mapped

### LINKED ISSUE #6141: FITS reading should emit a warning when memory mapping is explicitly requested but not successful
Our FITS reading currently has the default behavior of attempting to memory map (see #1745), but this memory mapping can silently "fail" and the returned data array will not be memory mapped in certain situations (e.g., the data array needed to be decompressed or scaled).  I don't have a problem with the default behavior being silent.

However, if a user explicitly passes in `memmap=True`, I suggest emitting a warning if the returned data array is not actually memory mapped.  The warning can also suggest to the user to use `disable_image_compression=True` or `do_not_scale_image_data=True` as applicable to preserve memory mapping.

See also https://github.com/sunpy/sunpy/pull/6110#discussion_r865083794

davnish:
This adds a warning to the read function in `sunpy.io`. This warning is emitted when  `memmap=True` and the data array is not actually memory mapped.

Merging this PR:
Closes #6141


ayshih:
Thanks for the PR!  There are issues with how your use of metadata keywords, but it's moot because you should instead explicitly check whether the data array is memory mapped.  Please see https://github.com/sunpy/sunpy/pull/6110#discussion_r865083794 for one way to do so.

I'm going to switch this PR to draft status so that it will not be critically reviewed before it is ready (e.g., with unit tests).  We continue to pay attention to all PRs as they are developed, even draft PRs.

davnish:
Thanks for the helping me @ayshih. I have updated the approach. Will this approach work?

ayshih (changes requested):
You need to add one or more unit tests to verify that this PR works as intended.  The code I wrote in https://github.com/sunpy/sunpy/pull/6110#discussion_r865083794 uses recursion because in general you could be inspecting a view of a non-memory-mapped array, and then `base` would not be `None`, but that doesn't mean memory mapping is being used.

Also, the warning to the user shouldn't blindly suggest using `disable_image_compression=True` or `do_not_scale_image_data=True`.  The suggestion should be tailored to why we believe the array was not memory mapped.  That is, don't suggest to a user that they turn off image compression when the image wasn't actually compressed in the first place.

davnish:
@ayshih Is this okay? I have added the `isinstance` function to check whether `data.base` is `mmap.mmap` object or not.
I have also tailored the warnings to be more specific. 

Also, how should I approach testing on this? I did some test on `TEST_AIA_IMAGE` and `TEST_RHESSI_IMAGE` and the read function works properly on them.

ayshih:
> Also, how should I approach testing on this?

In our sample data set (as opposed to our test data set), `AIA_171_IMAGE` is compressed and `SWAP_LEVEL1_IMAGE` is scaled, so neither should successfully memory map.  We could add small test files to our test data set, but for now you can proceed with creating online-required unit tests.

Additionally, be aware that:

- `AIA_171_IMAGE` has the compressed data as the second HDU, so your code will error as currently written because it assumes that the first HDU has a data array.
- For seeing whether the data array has been rescaled, your check for the `BSCALE` keyword will not work because the `BSCALE` keyword is removed by Astropy when it rescales the data.  You need to instead check whether `data._orig_bzero !=0` or `data._orig_bscale != 1` or `data._blank is not None`.
- For seeing whether the data array needed to be uncompressed, your check for the `BZERO` keyword doesn't make sense because that is not a keyword relevant to compression.  You should be able to instead simply check whether the HDU is of type `CompImageHDU`.

sunpy/io/_fits.py
```diff
@@ -99,6 +100,15 @@ def read(filepath, hdus=None, memmap=None, **kwargs):
                 message += repr(e)
                 warn_user(message)
 
+            if memmap == True and not (isinstance(pairs[0].data.base, mmap.mmap) and pairs[0].data.base is not None):
+                try:
+                    if pairs[0].header['BZERO'] != 0:
+                        warn_user("Data array is not memory mapped. Use 'disable_image_compression=True' to preserve memory mapping.")
```
Cadair, line 106: You can preserve the memmap, but it would be entirely useless as you wont be able to see the image any more? I don't think this is a good warning to throw.

sunpy/io/_fits.py
```diff
@@ -99,6 +100,15 @@ def read(filepath, hdus=None, memmap=None, **kwargs):
                 message += repr(e)
                 warn_user(message)
 
+            if memmap == True and not (isinstance(pairs[0].data.base, mmap.mmap) and pairs[0].data.base is not None):
+                try:
+                    if pairs[0].header['BZERO'] != 0:
```
Cadair, line 105: Is this actually a reliable way to detect a compressed image header?

ayshih: No, it isn't for two reasons.  See https://github.com/sunpy/sunpy/pull/6837#issuecomment-1479628183

sunpy/io/_fits.py
Cadair: All that's true, but I don't think that this warning is helpful as written for someone who doesn't have a good understanding of FITS internals. I can just see someone doing that and then being like waaaat

ayshih: The easy "fix" would be to elaborate in the warning, e.g.:
> Data array is not memory mapped because it is stored in the file using compression. Specify `disable_image_compression=True` to preserve memory mapping, but only if you do not need the data array to be decompressed.

davnish:
Do I have to create a different PR for the tests?

davnish:
How are these changes?
Do I have to create a seperate PR for the tests?
I tested on `AIA_171_IMAGE` and `SWAP_LEVEL1_IMAGE`. Works fine on `AIA_171_IMAGE` it throws the warning, but when I read `SWAP_LEVEL1_IMAGE` and `memmap = True` it already throws a `ValueError: Cannot load a memory-mapped image: BZERO/BSCALE/BLANK header keywords present. Set memmap=False.`

nabobalis:
> Do I have to create a seperate PR for the tests?

No, they should be in this PR.

davnish:
@ayshih can you review this again?

PR #6843, opened 2023-03-13
Use type hints with single dispatch in time.py

### LINKED ISSUE #3201: Use type hints with single dispatch in time
As of python 3.7 single dispatch supports registration by type hint. We should consider using this when we are 3.7+

Telomelonia:
## PR Description
Adds type annotation(type-hinting) and `@singledispatch` for few function 

## TODO
- [ ] changelog
- [ ] single dispatch for `is_time` 
- [ ] single dispatch for `is_time_in_given_format `

Fixes #3201

### DIFF
--- a/sunpy/time/time.py
+++ b/sunpy/time/time.py
@@ -71,7 +71,7 @@ TIME_FORMAT_LIST = [
 _ONE_DAY_TIMEDELTA = TimeDelta(1 * u.day)
 
 
-def is_time_equal(t1, t2):
+def is_time_equal(t1 :u.nanosecond, t2 :u.nanosecond) -> bool:
     """
     Work around for https://github.com/astropy/astropy/issues/6970.
 
@@ -296,7 +296,7 @@ def _variables_for_parse_time_docstring():
 
 
 @add_common_docstring(**_variables_for_parse_time_docstring())
-def parse_time(time_string, *, format=None, **kwargs):
+def parse_time(time_string : str, *, format:str=None, **kwargs) -> astropy.time.Time:
     """
     Takes a time input and will parse and return a `astropy.time.Time`.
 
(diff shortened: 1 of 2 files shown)


sunpy/time/time.py
```diff
@@ -296,7 +296,7 @@ def _variables_for_parse_time_docstring():
 
 
 @add_common_docstring(**_variables_for_parse_time_docstring())
-def parse_time(time_string, *, format=None, **kwargs):
+def parse_time(time_string : str, *, format:str=None, **kwargs) -> astropy.time.Time:
```
nabobalis, line 299: I don't think the first type is correct.

sunpy/time/time.py
```diff
@@ -71,7 +71,7 @@
 _ONE_DAY_TIMEDELTA = TimeDelta(1 * u.day)
 
 
-def is_time_equal(t1, t2):
+def is_time_equal(t1 :u.nanosecond, t2 :u.nanosecond) -> bool:
```
nabobalis, line 74: Are these always nanoseconds?

sunpy/time/time.py
Telomelonia: uhh yes I understood the issue incorrectly, will work on that

sunpy/time/time.py
Telomelonia: Hmm... true I need to consider multiple registers and make methods for those

Telomelonia:
> The goal of the original issue is to use [`functools.singledispatch`](https://docs.python.org/3/library/functools.html#functools.singledispatch) to automatically switch between the different functions used to parse times. This doesn't involve adding any type hinting. I recommend looking at the documentation first (linked in the previous sentence), and then trying to apply it to the `parse_time` parts of `time.py`. If you have any questions feel free to ask them here or in the original issue.

Yeah silly me, I mistook the issue and started to apply type hints will make changes ✌️

nabobalis:
Hi @Telomelonia, I am going to close this pull request. If you have time to come back to this, please reopen it.

PR #1981, opened 2017-02-01
added method to get entry from file using hdu_index

abit2:
https://github.com/sunpy/sunpy/issues/1970

Ping @Cadair @dpshelio

PR #6845, opened 2023-03-14
Adds a helper function to open fits file based on filename and HDU

### LINKED ISSUE #1970: Helper function to open fits file based on filename and HDU
After #1952 is merged, the user will still have to open the file and then extract the particular desired HDU.

A helper function that would return the data and the header of that particular HDU would be useful.

Telomelonia:
## PR Description

As @abit2 solved the issue in PR #1981 but was closed I think it was the right fix and wanted to continue his work.

Get a entry by its unique index number. If an entry with the given index does not exist, :exc:`sunpy.database.HDUEntryNotFoundError` is raised.

Fixes #1970 
## TODO
- [ ] changelog
## Query for reviewer
- [ ] I came across a function that performs a task similar to this helper function and wanted to clarify. Check [this](https://github.com/sunpy/sunpy/blob/df66b46e224199de12ea623aafb20ecc6a10c94f/sunpy/database/database.py#L637).

### DIFF
--- a/sunpy/database/database.py
+++ b/sunpy/database/database.py
@@ -659,7 +659,7 @@ class Database:
     
     def get_entry_by_hdu_index(self, file_name, hdu_index=0):
         """
-        Get a hdu index entry by its unique index number. If an entry with the
+        Get a entry by its unique index number. If an entry with the
         given index does not exist, :exc:`sunpy.database.HDUEntryNotFoundError` is
         raised.
         """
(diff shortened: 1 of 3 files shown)


nabobalis:
Thanks for the PR @Telomelonia but as I was saying on the issue. The future of database is pretty much on life support so there is no point in fixing issues for that. 

I will be closing this PR and any open issues for database.

Thanks again.

Telomelonia:
No problem 😊 ... I didn't know about the database issue I'll keep that in mind for future contribution ✌️

Issue #6585, opened 2022-11-10
"The Reprojecting Images to Different Observers example needs to explain the different rsun issue" "Curerently we deal with the fact that STEREO and SDO use different RSun constants in this example (https://docs.sunpy.org/en/stable/generated/gallery/map_transformations/reprojection_different_observers.html) with no explanation at all


Cadair:

ayshih:
Do users truly "keep tripping over it"?  The reason why @DanRyanIrish was confused by the first part of the example was not because `rsun` was being set, but rather because `observer` didn't seem to need to be set.  That's because `map_aia.reference_coordinate` already contains the desired observer, which is non-obvious to non-experts.

Cadair:
Danny isn't the first person I have had in chat asking about it. I certainly feel like it's worth a "STEREO and SDO don't use the same rsun" line.

wtbarnes:
Thinking in terms of our potential diátaxis refactor, I think this is more of an *explanation* and should go in a separate doc that we link to.

Cadair:
That would be fine (and I do say that), but we do need to explain it somewhere.

ayshih:
What I just wrote on Element:
> It's because of the (default) round-trip-check requirement for reproject. The idea is that you should trust the reprojection only if the 2D coordinates from each observer both resolve to the same 3D point (within a pixel). When `rsun` is different, that criterion fails towards the limb, which [results in NaNs, and] in some cases results in funny-looking banding as it fails and then succeeds and then fails again.

PR #6853, opened 2023-03-16
Change "Reprojecting Images to Different Observers"  example to avoid using custom wcs header until necessary

alasdairwilson:
This changes the Reprojecting Images to Different Observers example to avoid making a custom wcs until necessary (doesn't use it for euvi but does for mars). 

I think there has been repeated confusion amongst users that see this example and assume a custom output wcs is necessary.

To avoid the reprojected image having NaNs on the limb, we still force the rsun for the two maps to match before reprojecting.
 
This however only partly deals with #6585 , I think a diataxis style explanation page is still needed, this example is now begging for a: "for a full explanation of why this is necessary go here" kind of deal but I am not sure I am the man for that page. I can definitely start it if people want, get the ball rolling (in a new PR,) but will probably need help from more savy reprojecters to weigh in.

### DIFF
--- a/examples/map_transformations/reprojection_different_observers.py
+++ b/examples/map_transformations/reprojection_different_observers.py
@@ -63,26 +63,19 @@ plt.legend([limb_aia[0], limb_euvi[0]],
            ['Limb as seen by AIA', 'Limb as seen by EUVI A'])
 
 ######################################################################
-# We now need to construct an output WCS. We build a custom header using
-# :func:`sunpy.map.header_helper.make_fitswcs_header` but we use a lot of the ``map_aia``
-# properties to do it.
+# Due to a mismatch between the solar radius as defined in ``map_aia``
+# and ``map_euvi``, our reprojection will not behave correctly on pixels
+# very near the limb. This can be prevented by equating the two radii.
 
-out_header = sunpy.map.make_fitswcs_header(
-    out_shape,
-    map_aia.reference_coordinate.replicate(rsun=map_euvi.reference_coordinate.rsun),
-    scale=u.Quantity(map_aia.scale),
-    instrument="EUVI",
-    observatory="AIA Observer",
-    wavelength=map_euvi.wavelength
-)
+map_euvi.meta['rsun_ref'] = map_aia.meta['rsun_ref']
 
 ######################################################################
-# We can now reproject the EUVI map to this output WCS header.
-# The :meth:`~sunpy.map.GenericMap.reproject_to` defaults to using
+# We can reproject the EUVI map to the AIA observer wcs using
+# :meth:`~sunpy.map.GenericMap.reproject_to`. This method defaults to using
 # the fast :func:`reproject.reproject_interp` algorithm, but a different
 # algorithm can be specified (e.g., :func:`reproject.reproject_adaptive`).
 
-outmap = map_euvi.reproject_to(out_header)
+outmap = map_euvi.reproject_to(map_aia.wcs)
 
 ######################################################################
 # We can now plot the STEREO/EUVI image as seen from the position of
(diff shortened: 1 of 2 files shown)


examples/map_transformations/reprojection_different_observers.py
```diff
@@ -63,26 +63,19 @@
            ['Limb as seen by AIA', 'Limb as seen by EUVI A'])
 
 ######################################################################
-# We now need to construct an output WCS. We build a custom header using
-# :func:`sunpy.map.header_helper.make_fitswcs_header` but we use a lot of the ``map_aia``
-# properties to do it.
+# Due to a mismatch between the solar radius as defined in ``map_aia``
+# and ``map_euvi``, our reprojection will not behave correctly on pixels
+# very near the limb. This can be prevented by equating the two radii.
```
dstansby, line 68:
```suggestion
# very near the limb. This can be prevented by assuming that the emission
# observed by EUVI was emitted at the same radius from the centre of the Sun
# as the emission observed by AIA.
```

dstansby: I think this is a bit clearer on exactly what the line of code means in terms of assumption

examples/map_transformations/reprojection_different_observers.py
```diff
@@ -63,26 +63,19 @@
...
-    instrument="EUVI",
-    observatory="AIA Observer",
-    wavelength=map_euvi.wavelength
-)
+map_euvi.meta['rsun_ref'] = map_aia.meta['rsun_ref']
 
 ######################################################################
-# We can now reproject the EUVI map to this output WCS header.
-# The :meth:`~sunpy.map.GenericMap.reproject_to` defaults to using
+# We can reproject the EUVI map to the AIA observer wcs using
```
dstansby, line 73:
```suggestion
# We can reproject the EUVI map to the AIA world coordinate system (WCS) using
```

examples/map_transformations/reprojection_different_observers.py
alasdairwilson: I agree that it is clearer as to what an assumption about rsun means in general but I think there might be a danger there that readers leave with the idea that the mismatch between EUVI (or really the sunpy default rsun value) and AIA rsun is because of different emission heights.

dstansby: To make it all clear, how about:
> Data providers can set the radius at which emission in the map is assumed to have come from. Most maps set this to the standard photospheric radius (including EUI maps), but some maps (including AIA maps) set it to a slightly larger value. This mismatch means our reprojection won't work correctly on pixels near the limb. This can be prevented by assuming that the emission observed by EUVI was emitted at the same radius from the centre of the Sun as the emission observed by AIA.

? I think it's not a bad thing if users go away understanding that emission can come from different heights, and there has to be an assumption about those heights somewhere?

alasdairwilson: I like this paragraph, I think I was being stupid: my thought was this needs an explanation page for what rsun is, how it can have different values and why that can cause problems with reproject.

Now I have realised that this text can go in here now and when the page is made, be removed or reduced in place of a link, will make the change.

dstansby:
> Instead of suggesting to a user to modify the EUVI metadata (`map_euvi.meta['rsun_ref']`), what if we suggested to the user:
> 
> ```python
> map_aia.wcs.wcs.aux.rsun_ref = sunpy.sun.constants.radius.to_value('m')
> ```

What's the motivation for doing this instead of modifying the metadata? Because `GenericMap.wcs` is a property (not an attribute), this change will not persist if any of the other metadata is changed and `.wcs` is re-cached. I think it's safer and cleaner to change the metadata directly, since the current model of `GenericMap` is using the metadata as ground truth for deriving the properties.

dstansby:
Removing this from 5.0 milestone, we can backport this to a bugfix release if it doesn't get in 5.0.0

alasdairwilson:
> > Instead of suggesting to a user to modify the EUVI metadata (`map_euvi.meta['rsun_ref']`), what if we suggested to the user:
> > ```python
> > map_aia.wcs.wcs.aux.rsun_ref = sunpy.sun.constants.radius.to_value('m')
> > ```
> 
> What's the motivation for doing this instead of modifying the metadata? Because `GenericMap.wcs` is a property (not an attribute), this change will not persist if any of the other metadata is changed and `.wcs` is re-cached. I think it's safer and cleaner to change the metadata directly, since the current model of `GenericMap` is using the metadata as ground truth for deriving the properties.

So I agree, and without any opposing argument: I am going to revert this change and the surrounding text to represent that change.

In addition to your point, it makes sense because it is not a generic change, if the maps were reversed then setting the map_eui metadata to the default value would not fix it (since it is already the default sunpy value) whereas equating the two meta values will always work.

nabobalis (approved):

PR #1846, opened 2016-06-29
Timeseries

Alex-Ian-Hamilton:
Early implementation of the TimeSeries class and factory.
Basic gallery tutorial included.

PR #1097, opened 2014-07-15
Added tools for Fermi/GBM data, and added GBM as a lightcurve source

aringlis:
Added basic support for Fermi/GBM as a lightcurve source, and some additional tools in sunpy/instr for Fermi. In summary:

  - Added support for downloading Fermi/GBM quicklook (for now binning is fixed into ~7 channels) data for NaI detectors. BGO is not included yet.
  - Added tools to calculate NaI detector angles vs Sun at any time or for a timerange. This info is useful for choosing which detector to download data for.

This is still work in progress. Todo:
  ~~- more cleanup~~
  ~~- add tests~~
  ~~- deal with long timeranges in detector angle queries.~~

It would be possible to add support for arbitrary binning of the GBM data here (the native binning is in 128 channels), but this is a grey area between LC and Spectrum, so it's not clear to me how to proceed there. I'll raise this as a separate issue.

PR #6858, opened 2023-03-22
Remove un-used fermi gbm code

dstansby:
This code isn't tested, isn't used anywhere else, and after having a quick look through git blame doesn't appear to ever have been used...


ayshih:
Out of curiosity, I tracked down the history of this function.  It originated in commit 5675199 in PR #1097 (sunpy v0.6).  The code was copied to a new file in commit 2e59759 of PR #1846 (sunpy v0.8) with the change from `LightCurve` to `TimeSeries`.  And then code that uses the function was removed in commit 88d03db of that same monster PR, but the function itself was not removed at the same time.

PR #6859, opened 2023-03-22
Bump min version of h5netcdf

dstansby:
As per our dependency support policy. This allows some h5netcdf version-dependent code to be removed from GOES timeseries parsing.


changelog/xxxx.feature.rst
nabobalis, line 1: Just rename 6859 to and it LGTM

Cadair (approved):

PR #6876, opened 2023-03-22
Download hvpy files to the sunpy download dir

Cadair:


dstansby (approved):
but I am sad that `get_and_create_download_dir()` does not return a `Path`

Cadair:
> but I am sad that get_and_create_download_dir() does not return a Path

me too.

Issue #6864, opened 2023-03-22
Improve "Coordinates" section of the tutorial


dstansby:
### Describe the feature

Things that I've noticed that could be improved at a glance are:

- Move "Using Coordinates with Maps" to the map section of the tutorial (at the point of coordinates in the tutorials we haven't introduced the concept of maps)

There might be other improvements that can be made too 🧭

dstansby:
I'm going to de-milestone this, since we can always backports improvements to the tutorials

wtbarnes:
I do not agree. The refactor of the documentation is our biggest improvement for 5.0. There is not any point in pushing something out the door that is incomplete.

dstansby:
I don't think pushing out the coordinate tutorial in it's current state would be considered incomplete? I think there's incremental improvements to be made, and given they can be backported and no-one opened a PR before feature freeze that we shouldn't hold up 5.0 for it.

PR #6881, opened 2023-03-23
Move content from coordinate reference to guide

dstansby:
This PR moves existing content from the `sunpy.coordinates` reference section to the new guides section, without changing any of the content.


docs/reference/coordinates/index.rst
```diff
@@ -85,247 +83,6 @@ and `Franz & Harper (2002) <https://doi.org/10.1016/S0032-0633(01)00119-2>`_
 (and `corrected version <https://www2.mps.mpg.de/homes/fraenz/systems/systems3art/systems3art.html>`_).
```
Cadair, line 84: I feel like we should link to the guide pages here?

docs/reference/coordinates/index.rst
```diff
@@ -1,5 +1,3 @@
-.. _sunpy-coordinates:
```
Cadair, line 1: I think this section still needs an anchor

docs/guide/index.rst
```diff
@@ -17,3 +17,9 @@ Although there are code snippets in various bits of the guides, these are to hel
 
    installation
    logger
+
+
+.. toctree::
+   :maxdepth: 2
```
Cadair, line 23: Why another toctree?

docs/reference/coordinates/index.rst
nabobalis: Fixed.

docs/reference/coordinates/index.rst
nabobalis: Fixed.

docs/guide/index.rst
nabobalis: Removed.

docs/guide/coordinates/index.rst
```diff
@@ -0,0 +1,324 @@
...
+The `sunpy.coordinates` sub-package contains:
+
+* A robust framework for working with solar-physics coordinate systems
+* Functions to obtain the locations of solar-system bodies (`sunpy.coordinates.ephemeris`)
+* Functions to calculate Sun-specific coordinate information (`sunpy.coordinates.sun`)
+
+The SunPy coordinate framework extends the
+:ref:`Astropy coordinates framework <astropy:astropy-coordinates>`.
+
+This page contains a short overview of how coordinates work in sunpy.
```
ayshih, line 16:
```suggestion
This page contains an overview of how coordinates work in sunpy.
```

docs/guide/coordinates/index.rst
```diff
@@ -0,0 +1,324 @@
...
+For more detailed information see the following pages:
```
ayshih, line 17:
```suggestion
See the following pages for detailed discussion of specific topics:
```

docs/guide/coordinates/index.rst
```diff
@@ -0,0 +1,324 @@
...
+provided, and repopulate the final frame with them. If you were to do
+transformations using the Frames alone this would not happen.
+
+The most important implication for this in `sunpy.coordinates` is the ``rsun``
+parameter in the projective frames. If you create a projective frame with a
+``rsun`` attribute, if you convert back to a projective frame it will be set
+correctly. It should also be noted that, if you create a Heliographic frame and
+then transform to a projective frame with an ``rsun`` attribute, it will not
+match the ``radius`` coordinate in the Heliographic frame. This is because you may
+mean to be describing a point above the defined 'surface' of the Sun.
```
ayshih, line 324:
```suggestion
```
This paragraph is obsolete because `HeliographicStonyhurst` has had a `rsun` frame attribute for quite a while now.

docs/guide/coordinates/index.rst
```diff
@@ -0,0 +1,324 @@
...
+coordinate frames are created using map.
+
+In the case of `~sunpy.coordinates.frames.HeliographicCarrington`, one can specify ``observer='self'`` to indicate that the coordinate itself should be used as the observer for defining the coordinate frame.
+
+It is possible to convert from a `~sunpy.coordinates.frames.Helioprojective`
+frame with one observer location to another
+`~sunpy.coordinates.frames.Helioprojective` frame with a different observer
+location, by converting through `~sunpy.coordinates.frames.HeliographicStonyhurst`, this
+does involve making an assumption of the radius of the Sun to calculate the
+position on the solar sphere. The conversion can be performed as follows::
```
ayshih, line 261:
```suggestion
It is possible to convert from a `~sunpy.coordinates.frames.Helioprojective` frame with one observer location to another `~sunpy.coordinates.frames.Helioprojective` frame with a different observer location.
The transformation requires the coordinate to be 3D, so if the initial coordinate is only 2D, the default assumption maps that 2D coordinate to the surface of the Sun, as defined by the ``rsun`` frame attribute.
The conversion can be performed as follows::
```

docs/guide/coordinates/index.rst
```diff
@@ -0,0 +1,324 @@
...
+  >>> hpc2 = hpc1.transform_to(stereo.coordinate_frame)  # doctest: +SKIP
+
+
+Design of the Coordinates Sub-Package
+=====================================
+
+This sub-package works by defining a collection of ``Frames``
+(`sunpy.coordinates.frames`), which exists on a transformation graph, where the
+transformations between the coordinate frames are then defined and registered
+with the transformation graph (`sunpy.coordinates.transformations`). It is also
```
ayshih, line 285:
```suggestion
with the transformation graph (`sunpy.coordinates`). It is also
```
The transformation graph isn't shown in `sunpy.coordinates.transformations`, which will eventually be turned into a private module anyway.

docs/reference/coordinates/index.rst
```diff
@@ -84,246 +84,7 @@ see `Thompson (2006) <https://doi.org/10.1051/0004-6361:20054262>`_
 and `Franz & Harper (2002) <https://doi.org/10.1016/S0032-0633(01)00119-2>`_
 (and `corrected version <https://www2.mps.mpg.de/homes/fraenz/systems/systems3art/systems3art.html>`_).
```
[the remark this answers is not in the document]
wtbarnes: Ha I had the exact opposite reaction when looking at the table. I would say it is solidly "reference" material, and very much not "how-to-guide" material. From [the diátaxis on how-to guides](https://diataxis.fr/how-to-guides/), how-to guide characteristics include:
- action and only action
- no digression, explanation, teaching

wtbarnes:
I've already left several comments in a similar vein, but I do not think that the components of the coordinates API docs should be wholesale copied to the "guide" section. FWIW, I'm treating the "guide" section of a "how-to guide" section in the language of the diátaxis framework.

By section, my suggested reorganization in terms of the diátaxis sections is,

- Supported Coordinate Systems --> Reference
- Getting Started --> Tutorial
- Transforming Between Coordinate Frames --> Tutorial
- Observer Location Information  --> How-to Guide
- Design of the Coordinates Subpackage --> Explanation
- More Detailed Information --> Explanation

ayshih:
> FWIW, I'm treating the "guide" section of a "how-to guide" section in the language of the diátaxis framework.

That's precisely a point that I'm unclear about.  Is that what we are aiming for?

docs/reference/coordinates/index.rst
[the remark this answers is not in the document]
ayshih: I disagree.  If the table had the classes listed in the first column, sorted by the class names, and the other information in subsequent columns, it'd be a reference table.  However, as formatted, sorted, and intended to be used, I argue that it is a bare how-to guide.  Perhaps it'd be clearer if the column titles were renamed:

- "Coordinate system" -> "If you want to use the coordinate system..."
- "Abbreviation" -> "also abbreviated as..."
- "SunPy/Astropy equivalent" -> "use the class..."
- "Notes" -> "but be aware that..."

alasdairwilson:
> Is that what we are aiming for?

That is the intention of that new guide section.

docs/reference/coordinates/index.rst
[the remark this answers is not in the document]
wtbarnes: Ok I can see the logic in how this *could* be used as a how-to guide and if that was how it was intended to be used, we should promote it as such. A relabeling of the columns as suggested would definitely make that more clear. Maybe breaking it out of a table and/or adding a title and description akin to "coordinate transform recipes" might also help?

I still think this could be useful piece of reference material as well. During the course of trying to understand what coordinate systems sunpy supports, a user may want to be able to look up what the supported coordinate systems are and a table is a natural way to present that. They should not need to go to a how-to guide to find that information.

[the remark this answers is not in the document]
Cadair: I think that as it is currently written this table should be a reference. If we want to reformat it later and add it to the guide then we should, but I think we should try and keep large copy editing to followup PRs?

dstansby:
Oh no, I'm intending the guide to be the Explanation part of the docs

Cadair:
> Oh no, I'm intending the guide to be the Explanation part of the docs

:facepalm:

So to be clear:

| Our Docs Section | Diátaxis Type |
| ------------------------- | -------------------- |
| Reference / API    | Reference       |
| Tutorial                  | Tutorials          |
| Guide                     | Explanation    |
| TBD                        | How-To Guides |

Is this what we are all thinking?

wtbarnes:
> Is this what we are all thinking?

No...we should label the sections in a way that tells users how we expect them to use that section of the guide. So "Explanation" should be "Explanation", "How-to-Guides" should be "How-to-Guides".

docs/reference/coordinates/index.rst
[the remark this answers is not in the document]
dstansby: I agree this seems like reference, so I'll leave it here in this PR. Open to someone else convincing us that it should be moved to another section in a further PR.

docs/reference/coordinates/index.rst
```diff
@@ -84,246 +84,7 @@ see `Thompson (2006) <https://doi.org/10.1051/0004-6361:20054262>`_
...
-
-.. toctree::
-   :maxdepth: 1
-
-   carrington
-   rotatedsunframe
-   velocities
-   wcs
-   other_api
+The :ref:`coordinates guide <sunpy-coordinates>` provides in depth discussion of the structure and concepts underlying the coordinates framework.
```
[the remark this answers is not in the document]
dstansby: I've moved this up the page so it's above the table

docs/guide/coordinates/index.rst
```diff
@@ -0,0 +1,324 @@
...
+
+For a description of these coordinate systems,
+see `Thompson (2006) <https://doi.org/10.1051/0004-6361:20054262>`_
+and `Franz & Harper (2002) <https://doi.org/10.1016/S0032-0633(01)00119-2>`_
+(and `corrected version <https://www2.mps.mpg.de/homes/fraenz/systems/systems3art/systems3art.html>`_).
+
+
+Getting Started
+===============
+
```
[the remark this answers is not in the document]
dstansby: Is this something you'd like me to do in this PR?

dstansby:
@wtbarnes I think you should re-review this assuming that the "Guide" as it's currently called is the "Explanation" part of diataxis. If you would like me to change "Guide" > "Explanation" first in this PR, let me know and I'd be happy to do it (but don't want to do it before someone asks for it!)

docs/reference/coordinates/index.rst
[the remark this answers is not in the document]
ayshih: Well, even if we can't agree on where the table should be if were to be in only one place, don't implement the two tables with duplicated RST, because that's just annoying from a maintenance perspective.  The table should live as a separate RST snippet that is inserted in both places using the `include` directive.

docs/guide/coordinates/index.rst
```diff
@@ -0,0 +1,324 @@
...
+  >>> hpc2 = hpc1.transform_to(hpc_out)
+
+An example with two maps, named ``aia`` and ``stereo``::
+
+  >>> hpc1 = SkyCoord(0*u.arcsec, 0*u.arcsec, frame=aia.coordinate_frame)  # doctest: +SKIP
+  >>> hpc2 = hpc1.transform_to(stereo.coordinate_frame)  # doctest: +SKIP
+
+
+Design of the Coordinates Sub-Package
+=====================================
```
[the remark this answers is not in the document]
wtbarnes: Ok in my revised understanding of what is meant by "Guide" in the context of this PR, I think it makes sense to leave this as is.

wtbarnes:
> Also very open to renaming "Guide" if it's causing confusion

To recap the discussion from the dev meeting last week, the consensus was that "guide" is too generic and that "Topic Guide" or "Explanation" would be preferable.

wtbarnes:
> @wtbarnes I think you should re-review this assuming that the "Guide" as it's currently called is the "Explanation" part of diataxis. If you would like me to change "Guide" > "Explanation" first in this PR, let me know and I'd be happy to do it (but don't want to do it before someone asks for it!)

With a clearer understanding of the labeling being used here, the last three items in my suggested reorganization above are already done. That leaves the following,

1. Supported Coordinate Systems --> Reference
2. Getting Started --> Tutorial
3. Transforming Between Coordinate Frames --> Tutorial
4. Observer Location Information --> How-to Guide

We are still debating item 1 (see above comments). I'm now inclined to leave items 2 and 3 to #6864. We do not yet have a how-to guide section so item 4 should wait until that exists. All that being said, in the interesting of not completely blowing up the scope of this PR more than I already have, I think this PR should do the following,

- [x] Separate material from coordinates reference material that is not reference
- [ ] Decide on "Explanation" or "Topic Guide" and rename the "Guide" to that
- [ ] Settle the debate on the "supported coordinate systems" table

docs/guide/coordinates/index.rst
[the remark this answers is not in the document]
wtbarnes: No, I don't think so. See my comment below for reasoning.

alasdairwilson:
I personally am a fan of "Topic Guides" as a title: Despite explanations being the diataixis name for that data type, the title should best represent what can be found in that section to an end user.

As for the other todo item, I think that table should go in reference even if there is a place in the docs for a duplication of the content in a different format, e.g. a "Topic Guide" illustrating the difference between the sunpy frames, or a tutorial on frame transformations.

dstansby:
Thanks for the clear comments! I think I've addressed them, and opened two new issues to track the ones that I won't tackle in this PR (https://github.com/sunpy/sunpy/issues/6913 and https://github.com/sunpy/sunpy/issues/6914)

wtbarnes (approved):
Thanks for this @dstansby.

alasdairwilson (approved):

PR #6882, opened 2023-03-28
Change the output of `sample_at_coords()` to have units

ayshih:
For some reason, `sunpy.map.maputils.sample_at_coords()` doesn't return a `Quantity` when sampling pixel values.  This PR fixes that oversight, although this is an API change that could break code in the wild without any practical way to have a deprecation period.

### DIFF
--- a/changelog/6882.breaking.rst
+++ b/changelog/6882.breaking.rst
@@ -1,0 +1,1 @@
+Changed the output of :func:`sunpy.map.sample_at_coords` to return the sampled values as `~astropy.units.Quantity` with the appropriate units instead of merely numbers.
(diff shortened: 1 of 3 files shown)


wtbarnes (approved):
I'm ok with this being a breaking change, especially given that `Quantity` objects behave like numpy arrays in most scenarios.

nabobalis (approved):

nabobalis:
Needs a rebase but good to me. Happy to break due to the reason Will has given.

ayshih:
Oh, whoops, I figured this would get merged before #6840, so I didn't mention in #6840 that it had been rebased on top of this PR.  So, the changes here, including the changelog entry, are already in with the merging of #6840.  I don't think it makes sense to rebase this PR, which would wipe out the single commit here and make it more difficult for future readers to understand.  I'll just close this PR, and we'll live with the weirdness of having our changelog point to a PR that wasn't directly merged.

PR #6887, opened 2023-03-29
Adds support in SunPy for the VSO REST/TAP Data Providers

ejm4567:
## PR Description

We have added support in the VSO to handle new Data Providers that use REST or TAP services as their transport mechanism. To accomplish this change, while maintaining the current codebase in the VSO for SOAP, we had to
generate an Apache sessionID in the VSO for REST/TAP queries, and return it to the client in the QueryResponse object in the column Info[Required]. For the case of SunPy, we changed the WSDL to convert Info from a simple string to a complexType. We chose Info since it is already in our Data Model. In the future, as we extend the VSO Data Model, additional data may well end up in Info, hence a complexType is a better data type to use in the long run than a simple string.

To have SunPy handle Info as a complexType, we have added Info in net/_attrs.py and net/vso/attrs.py.
We've also added a small bit of code in net/vso/vso.py to conditionally return Info[Required] if it exists for data
requests. This change will be transparent to Users, and they will not need to explicitly return Info[Required].
SunPy does nothing with the Apache session ID, it simply returns it back to the VSO for processing.

We've also added unit tests for SWAP & LYRA in Proba2 to test for existence of Info[Required] and the expected amount of data returned by Proba2. Finally, to cover all bases, we've added a simple SOAP query for AIA data where Info[Required] should _not_ exist.


changelog-VSO-REST-TAP
Cadair, line 1: This file needs to be renamed to `changelog/6887.feature.rst`

sunpy/net/_attrs.py
```diff
@@ -286,3 +286,13 @@ class ExtentType(SimpleAttr):
     """
     The type of Extent; for example, "FULLDISK", "SYNOPTIC", "LIMB", etc.
     """
+    
+class Info(SimpleAttr):
```
Cadair, line 290: Am I correct in my understanding that this isn't a thing the user can search on? If that's the case we don't need this as attrs are only for query input.

sunpy/net/vso/tests/test_vso.py
```diff
@@ -359,3 +359,49 @@ def test_iris_filename(client):
...
+@pytest.mark.remote_data
+def test_table_noinfo_required(client):
+    res = client.search(a.Time('2017/12/17 00:00:00', '2017/12/17 06:00:00'), a.Instrument('aia'),
+                        a.Wavelength(171 * u.angstrom),
+                        response_format="table")
+    has_info = False
+    res_keys = res.keys()
+    if 'Info Required' in res_keys:
+       has_info = True
+    assert not has_info
```
Cadair, line 372: I think this is better written as:

```suggestion
    assert 'Info Required' in res.keys()
```

sunpy/net/vso/vso.py
```diff
@@ -419,6 +419,14 @@ def fetch(self, query_response, path=None, methods=None, site=None,
 
     def make_getdatarequest(self, response, methods=None, info=None):
         """ Make datarequest with methods from response. """
+        
+        #    Pass back the Apache session ID to the VSO if it exists in the response
+
+        for item in response:
+           for k in item.keys():
+              if k == 'Info Required':
+                 info['required'] = item['Info Required']
```
Cadair, line 428: The indentation is wrong in this section, indentation should always be 4 spaces.

Cadair, line 428: A better way to write this would be:
```suggestion
        for item in response:
           info_required = item.get("Info Required", None)
           if info_required is not None:
                 info['required'] = item['Info Required']
```
as it prevents you looping through the keys.

This is assuming you can't just set the required value, i.e. will VSO ignore it if it's there and it's `None`?

sunpy/net/vso/tests/test_vso.py
```diff
@@ -359,3 +359,49 @@ def test_iris_filename(client):
...
+    res = client.search(a.Time('2020/02/15 00:00:00', '2020/02/15 20:00:00'), a.Instrument('swap'), a.Provider('ESA'), a.Source('PROBA2'),
+                        response_format="table")
+    has_info = False
+    res_keys = res.keys()
+    if 'Info Required' in res_keys:
+       has_info = True
+    assert has_info
+
+@pytest.mark.remote_data
+def test_table_length_test_swap(client):
```
Cadair, line 385: Does this test actually test anything in the sunpy code base? You haven't modified anything in search to make REST/TAP work?

I am a little nervous about tests which assert the number of results like this as that number can change without any modification to sunpy.

changelog-VSO-REST-TAP
Cadair, line 1: Our changelog is written for our users, so could you please rewrite this just focusing on what new functionality is being provided? i.e. "Add support for downloading files from the VSO using new REST or TAP data providers, i.e. Lyra and SWAP"? or whatever is appropriate.

Cadair (changes requested):
Hi @ejm4567 Thank you so much for the PR, it's super exciting to have the VSO team help maintain our client :heart:.

I have left a bunch of inline comments, please don't hesitate to comment back if you don't understand them or I am talking nonsense.

My big question for you is what happens if someone tries to download these data with the current version of our client? Will it just not work, will VSO error? 

I assume people can already search for these records, and that's why our online test build has started failing?

I ask these questions primarily because we need to decide if we are going to backport this change to already released versions of sunpy or if this is a feature that should only be available in 5.0.

Other comments:

* `Info Required` should be added to this ignore list, so we don't display it to users by default: https://github.com/sunpy/sunpy/blob/495ea6bde7b32a6a508c1b95a3f71155b130b859/sunpy/net/vso/table_response.py#L53
* This needs a test of fetching some data to make sure that the code in `make_getdatarequest` actually works as intended.

changelog-VSO-REST-TAP
ejm4567: OK Will do.  That part of the PR guideline was a little unclear to me.

sunpy/net/vso/tests/test_vso.py
ejm4567: Nice one-liner, and the response_format="table" was redundant. Tests updated.

sunpy/net/_attrs.py
ejm4567: Correct. Currently Info is not searchable in the VSO. But with a new Data Model that we are currently working  on, we may well make it searchable. I did find that if I did not have entries in both net/_attrs.py and net/vso/attrs.py for Info, I got errors when executing a TAP query.

sunpy/net/vso/tests/test_vso.py
ejm4567: The length tests in SunPy indirectly test that the VSO indeed returns the expected amount of data from specific TAP queries. I have similar tests in the backend Perl code testing the main CGI that the WSDL points to. We can of course take out the length tests if you prefer. Given that both the SWAP and LYRA tests are for dates back in 2020, I would think they are most likely done with any calibration and that the no. of files returned should be stable.

changelog-VSO-REST-TAP
ejm4567: Done.

sunpy/net/_attrs.py
Cadair: > we may well make it searchable

I think I would rather add this at the point you do make this searchable, as at the moment it might just confuse people.

> I got errors when executing a TAP query

I would be interested to see what they are, as I think I can search for the TAP backed things with released sunpy?

Cadair:
@ejm4567 I think the main thing that's left now is a test which actually confirms we can download a singe file with `.fetch()`.

sunpy/net/_attrs.py
ejm4567: After testing removing Info from both net/_attrs.py and net/attrs.py, I found that all tests were fine. So I was mistaken about the need for Info in those files. We can add it back once we make Info, and other fields searchable in the VSO in Data Model 2.0

sunpy/net/vso/tests/test_vso.py
```diff
@@ -359,3 +359,32 @@ def test_iris_filename(client):
...
+def test_table_has_info_required_lyra(client):
+    res = client.search(a.Time('2020/02/15 00:00:00', '2020/02/17 20:00:00'), a.Instrument('lyra'), a.Provider('ESA'), a.Source('PROBA2'))
+    assert 'Info Required' in res.keys() and len(res) > 0
+  
+@pytest.mark.remote_data
+def test_fetch_swap(client):
+    res = client.search(a.Time('2020/02/15 00:00:00', '2020/02/15 20:00:00'), a.Instrument('swap'), a.Provider('ESA'), a.Source('PROBA2'))
+    tmp_dir = '/tmp'
+    files = client.fetch(res[0:1], path=tmp_dir)
+    assert len(files) == 1
```
Cadair, line 383: You should use the [`tmp_path`](https://docs.pytest.org/en/latest/how-to/tmp_path.html#the-tmp-path-fixture) pytest fixture here so that it is automatically cleaned up and platform independent.
```suggestion
def test_fetch_swap(client, tmp_path):
    res = client.search(a.Time('2020/02/15 00:00:00', '2020/02/15 20:00:00'), a.Instrument('swap'), a.Provider('ESA'), a.Source('PROBA2'))
    tmp_dir = '/tmp'
    files = client.fetch(res[0:1], path=tmp_path)
    assert len(files) == 1
```

and the same for the other fetch test as well.

Cadair:
Just that one tweak and I think we are good to go @ejm4567 :partying_face:

Cadair:
Ah there is one other issue, which is that this LyraClient doctest has failed: https://github.com/sunpy/sunpy/actions/runs/4599077683/jobs/8124271759?pr=6887#step:10:5298

We need to decide what we are going to do about the fact that we have our pre-existing client with some overlap.

Cadair (commented):
I think the diff is good, if we can resolve the conflicts and the tests pass then we are good to go from me.

nabobalis:
I opened https://github.com/ejm4567/sunpy/pull/4 hopefully to get this PR closer to the finish line. 

It has been a long time since I did an actual merge, so I probably screwed up.

Cadair (approved):

Cadair:
Thanks a lot @ejm4567 I think this is ready to go.

@sunpy/sunpy-developers  This needs a second review and squash merging.

PR #6889, opened 2023-03-30
Marks sunpy.io readers depriciated

gmrpr321:
Marks the following modules in sunpy.io private
1. _cdf.py
2._file_tools.py
3._fits.py
4._jp2.py


nabobalis:
I think the changes your other pull request made seemed to be more encompassing?

PR #6891, opened 2023-03-31
fixes #6373 :makes sunpy.io readers private 

### LINKED ISSUE #6373: Private more parts of sunpy.io

Large parts of `sunpy.io` are used only and written for `Map` and `Timeseries`, they are not meant to be used by users to open files.

These are:

1. JPEG2000
2. CDF
3. FITS (this is already done)
4. ASDF
5. File tools

The readers we do support for users are:
1. ANA
2. GENX
3. SRS 

These should not be private.

gmrpr321:
Marks the following readers from sunpy.io private 
1.JPEG2000 
2.CDF
 3.File tools


sunpy/io/tests/_test_filetools.py
nabobalis, line 1: This should not change.

nabobalis:
I would recommend looking at the pull request that did this for the FITS reader.
We need to follow that example.

PR #6820, opened 2023-03-05
make submap() handle both 'BaseCoordinateFrame' and 'SkyCoord'

OussCHE:
## PR Description

As mentioned in the discussion of the #6252 issue and following the advice of @ayshih ,I modified the function so that `submap()` can handle `BaseCoordinateFrame` objects. 

However, I wanted to add a test to make sure that the functionality results in the same `AIAMAP` object,but I didn't find a way to compare the resulting objects (I compared them in debugging and using the code bellow, they look the same for me).

I wrote a snippet code that should do a big part of the job in comparing between the results(and to be used in the test). I can write the test(or modify my code) if you give me some feedback.

    import sunpy.map
    from sunpy.data.sample import AIA_171_IMAGE
    from astropy.coordinates import SkyCoord
    import numpy as np
    aia = sunpy.map.Map(AIA_171_IMAGE)
    bottom_left = aia.bottom_left_coord.frame
    top_right = aia.top_right_coord.frame
    basecoordianteframe_value = aia.submap(bottom_left, top_right=top_right)

    bottom_left =SkyCoord(aia.bottom_left_coord.frame)
    top_right = SkyCoord(aia.top_right_coord.frame)
    skycoord_value = aia.submap(bottom_left, top_right=top_right)
    
    np.testing.assert_equal(basecoordianteframe_value.data,skycoord_value.data)

Cheers

PR #6897, opened 2023-04-04
Fixed backticks in a changelog entry

ayshih:
I missed this in the changelog entry for #6820.  That changelog entry hasn't gotten into any release yet, so this PR doesn't need its own changelog entry.


nabobalis:
This was released for both 4.0 and 4.1

nabobalis (approved):

Issue #5388, opened 2021-06-04
Display data units to the the map quick view plots and colorbar ?


hayesla:
### Description

As suggested from workshop, it would be nice to have the data units (e.g. cts, etc) printed on the .peek() histogram (pixel value plot) and maybe added to the colorbar of a map when plotted

or:

The units of a map are accessible via .unit. This is included in the peek colorbar, but not in the fancy html repr. It would be nice to include this too, maybe just as an additional entry in the table on the left.

himanshu007-creator:
Hi , i would like to work on this issue. I know python, but u have never used sunpy. With guidance , i will be able to solve this issue

nabobalis:
Hi @himanshu007-creator, welcome.

We will help you as much as we can.

jiaqing23:
Hello @nabobalis, seems like no one is working on this issue. I would like to try it out.

Shikharishere:
Is this issue still open?

nabobalis:
> Is this issue still open?

Hello, I think what is left is to check that the unit appears on the colorbar and if no, add it and see if it looks nice. Otherwise this issue is finished.

dstansby:
Just checked some of the figure tests with colorbars, and units appear below the colobar, so I'll close this.

PR #6902, opened 2023-04-04
display data unit in html repr

jiaqing23:
## PR Description

xref #5388, data unit into html repr of `sunpy.map.Map`

|image|image|
|-|-|

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -401,7 +401,7 @@ class GenericMap(NDData):
         ax.semilogy()
         # Explicitly set the power limits for the X axis formatter to avoid text overlaps
         ax.xaxis.get_major_formatter().set_powerlimits((-3, 4))
-        ax.set_xlabel('Pixel value in linear bins')
+        ax.set_xlabel(f"Pixel value{' (' + str(self.unit) + ')' if self.unit else ''} in linear bins")
         ax.set_ylabel('# of pixels')
         ax.set_title('Distribution of pixel values [click for cumulative]')
         hist_src = _figure_to_base64(fig)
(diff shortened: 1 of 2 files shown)


sunpy/map/mapbase.py
```diff
@@ -401,7 +401,7 @@ def _repr_html_(self, compute_dask=False):
         ax.semilogy()
         # Explicitly set the power limits for the X axis formatter to avoid text overlaps
         ax.xaxis.get_major_formatter().set_powerlimits((-3, 4))
-        ax.set_xlabel('Pixel value in linear bins')
+        ax.set_xlabel(f'Pixel value{" (" + str(self.unit) +") " if self.unit is not None else " "}in linear bins')
```
nabobalis, line 404: Can we not tidy this string up a bit?

jiaqing23: Was thinking to create an temp variable:
```
display_unit = f"({str(self.unit)})" if self.unit is not None else " "
ax.set_xlabel('Pixel value' + display_unit + 'in linear bins') 
```

nabobalis: That sounds worse?

jiaqing23: This looks better:
```
display_unit = f' ({self.unit})' if self.unit is not None else ''
ax.set_xlabel(f'Pixel value{unit_string} in linear bins')
```

nabobalis: I don't think we need a variable for this.

jiaqing23: Shorten the code: 

ax.set_xlabel(f"Pixel value{' (' + self.unit + ')' if self.unit else ''} in linear bins")

nabobalis: Sure

jiaqing23: Updated the code

nabobalis (approved):

nabobalis:
Thanks @jiaqing23

Issue #6812, opened 2023-03-01
Add "imshow" as an option for autoalign plotting of maps


ayshih:
For autoalign plotting of maps, our current implementation uses `pcolormesh()`.  However, if the pixel-to-pixel transformation between the map's WCS and the axes's WCS is affine (i.e., just scaling and rotation), using `imshow(transform=...)` would have much better performance than `pcolormesh()`.  The trick is that we'd have to add a clip path to the image produced by `imshow()` because it can unnecessarily extrapolate data.

We should add `imshow()` as an option for autoalign plotting.

aritrasinha108:
@nabobalis I am on it

nabobalis:
> @nabobalis I am on this issue

Are you sure? This probably won't be a simple issue to tackle.

aritrasinha108:
@nabobalis I have read through the issue a bit and the PR [#5255](https://github.com/sunpy/sunpy/pull/5255) which is on the similar lines. I think I just need to look into how we can use `imshow` here.

ayshih:
Look at the middle panel of the figure in [this comment](https://github.com/astropy/astropy/issues/14420#issuecomment-1443864286), and then the addition of a clip path in [this later comment](https://github.com/astropy/astropy/issues/14420#issuecomment-1447197665).

Mihan786Chistie:
Ok I understood now why we need imshow() function in cases like only scaling and rotation, as it has a much better performance than pcolormesh() function and also we can't use imshow() function for every case as it might not handle to clip path correctly as pcolormesh() function does

Mihan786Chistie:
So when autoalign is set True, what are the conditions for to plot using pcolormesh() function and to use imshow() function?

Prtm2110:
Hello, what will be the clip path for a map object? Any resources on this will be helpful.
```
w = data.shape[1]
h = data.shape[0]
path = mp.path.Path([[-0.5,-0.5], [w-0.5,-0.5], [w-0.5,h-0.5], [-0.5,h-0.5], [-0.5,-0.5]])
```
This clip path from https://github.com/astropy/astropy/issues/14420#issuecomment-1443864286 don't seem to be general for all?

ayshih:
[...]

PR #6904, opened 2023-04-04
Adds 'imshow' as an option for autoalign plotting of maps

Mihan786Chistie:
#6812


ayshih:
This is the correct start in terms of interface (i.e., specifying `"imshow"` to the keyword argument), but you need to:

- Call `imshow()` with the `transform=` keyword so that the plot is actually transformed as intended
- Add a clip path to the image artist returned by `imshow()`

Please take another look at [the middle panel of this example script](https://github.com/astropy/astropy/issues/14420#issuecomment-1447197665), because our internals would need to do something similar.

Moving this PR to draft so that it is not critically reviewed until it is ready.

nabobalis:
Hi @Mihan786Chistie , I am going to close this pull request. If you have time to come back to this, please reopen it.

PR #6921, opened 2023-04-16
Now pass max_splits into the downloader for JSOC only fetches

nabobalis:
We never prevented more than one connection to the JSOC and I think this is why we now hit the rate limit.

Hopefully fixes https://github.com/sunpy/sunpy/issues/6510 again

### DIFF
--- a/changelog/6921.bugfix.rst
+++ b/changelog/6921.bugfix.rst
@@ -1,0 +1,1 @@
+Pass in "max_splits" to Parfive to prevent multi connections to JSOC for JSOC only queries.
--- a/sunpy/net/fido_factory.py
+++ b/sunpy/net/fido_factory.py
@@ -385,6 +385,7 @@ class UnifiedDownloaderFactory(BasicRegistrationFactory):
         # Avoid more than one connection for JSOC only requests.
         from sunpy.net.jsoc import JSOCClient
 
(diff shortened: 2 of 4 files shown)


dstansby (changes requested):
Amazing, thanks! Could you stick in a bugfix changelog entry, and then this looks good to go.

Cadair (changes requested):
I don't think this is the correct approach. Passing kwargs to both `Downloader` and the `client.fetch` method sounds fraught with trouble.

I think the answer is to not pass `**kwargs` but to pass `max_splits` explicitly.

nabobalis:
Added a changelog and only now pass in max_spiits.

nabobalis:
> Am I mistaken in thinking we aren't doing anything with this either: [`main`/sunpy/net/jsoc/jsoc.py#L485](https://github.com/sunpy/sunpy/blob/main/sunpy/net/jsoc/jsoc.py?rgh-link-date=2023-04-18T07%3A38%3A35Z#L485)

Yeah we didnt, should be fixed now.

Cadair (approved):
Looks good.

We should fix this test as well: https://github.com/sunpy/sunpy/actions/runs/4733988353/jobs/8402592079?pr=6921#step:10:5325

nabobalis:
I fixed that in another PR just to be tricky.

PR #6923, opened 2023-04-16
Simplify ways to get the solar angular radius

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/4516. This deprecates the previous two ways we had to calculate the angular radius, and replaces it with a new function. Before the inputs to the two functions were:

1. Time (to get angular radius seen by earth)
2. Coordinate (but had to be helioprojective, and took the observer attribute)

The new function takes a coordinate as input, so is much more generic. See the what's new entry for how to migrate from the old functions to the new.

### DIFF
--- a/docs/whatsnew/5.0.rst
+++ b/docs/whatsnew/5.0.rst
@@ -42,3 +42,26 @@ Deprecation of IO readers
 ``sunpy.io.cdf``, ``sunpy.io.file_tools`` and ``sunpy.io.jp2`` sub-modules have been deprecated, and will be removed in version 5.1.
 
 This because they are designed for internal use only, and removing it from the public API gives the developers more flexibility to modify it without impacting users.
+
+Simplified way to get the solar angular radius
+==============================================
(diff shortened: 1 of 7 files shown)


ayshih (commented):
I'm not a fan of deprecating `sunpy.coordinates.sun.angular_radius` and creating a new function.  Why not simply add a keyword argument `observer=...` to the existing function?

dstansby:
The motivation is to keep the API simple.

ayshih:
I understand why the separated API may be preferred by a developer, but I think it makes actual usage more cumbersome without any clear added value.  Plus, it introduces a new calling paradigm that is incompatible with the other dozen functions in `sunpy.coordinates.sun`.  We can instead simply extend the existing function (thus not disrupting everyone already using this function) and avoid transitioning to an unnecessary longer name for the function (none of the functions in `sunpy.coordinates.sun` should be prefixed with `solar_...`).

dstansby:
Closing as I'm don't have bandwidth to finish this off.

PR #6937, opened 2023-04-18
Only run PR CI on PRs targeting main

dstansby:
Since we don't wait for CI on backport PRs anyway, doesn't seem to be any point running it.


nabobalis:
We do want to run it, there was no need to do it for 6 separate PRs at the same time when we can do it for only 2.

dstansby:
Isn't the point that we still run it on the branch, but just not on the PRs? This PR keeps CI running on commits to the branch itself.

Issue #6940, opened 2023-04-19
Drop parfive < 2


Cadair:
### Provide a general description of the issue or problem.

After the 5.0 release we can drop parfive 1.x as 2.0 was released in [July 2022](https://pypi.org/project/parfive/#history)

dstansby:
Our policy is 24 months, so I don't think we can drop it until July 2024?

dstansby:
Ah, actually I guess the relevant date is Sep 24th 2021, which is when 1.5.0 was released, so we probably can drop support for parfive 1.x in sunpy 5.1

nabobalis:
I want to drop this for 5.0. We already have hit a download problem with the JSOC and supporting both versions of parfive was making that more difficult than it needs to be.

Issue #6308, opened 2022-06-28
Customise our user agent on CI


Cadair:
Looking through the VSO user agent statistics it would be good to be able to exclude CI jobs.

Cadair:
Hello @Vickyaryan33 we don't assign issues to people but you are free to work on it, please open a draft PR when you have something to share.

jayantbhakar:
Hi, I would like to work on this issue. Can you add some more information here, to better comprehend the CI jobs to be excluded, maybe some screenshots or output of the response statistics would do. That would help a lot. Thanks.

nabobalis:
The CI job to be exclude isn't on our side. It is to add more information to https://github.com/sunpy/sunpy/blob/main/sunpy/util/parfive_helpers.py#L15 to indicate if the sunpy test suite has made the server call or if it is a user.

PR #6942, opened 2023-04-20
Drop parfive less than 2

nabobalis:
Also a hack for adding the words CI-test to the user agent headers for our tests.

Fixes https://github.com/sunpy/sunpy/issues/6308
Fixes https://github.com/sunpy/sunpy/issues/6940

I adjusted the previous changelog for updated versions instead of a new one.

### DIFF
--- a/changelog/6742.feature.rst
+++ b/changelog/6742.feature.rst
@@ -3,7 +3,6 @@ The minimum required versions of several core dependencies have been updated:
 - Python 3.9
 - astropy 5.0.1
 - numpy 1.21.0
-- parfive 2.0.0
 
 The minimum required versions of these optional dependencies has also been updated:
 
--- a/changelog/6742.feature.rst
+++ b/changelog/6742.feature.rst
@@ -3,6 +3,7 @@ The minimum required versions of several core dependencies have been updated:
 - Python 3.9
 - astropy 5.0.1
 - numpy 1.21.0
+- parfive 2.0.0
 
 The minimum required versions of these optional dependencies has also been updated:
 
--- a/sunpy/conftest.py
+++ b/sunpy/conftest.py
@@ -115,6 +115,16 @@ def undo_config_dir_patch():
 
 
 @pytest.fixture(scope='session', autouse=True)
+def parfive_header_test(request):
+    """
+    Add a keyword to tell parfive this is a test run.
+    """
+    os.environ["PARFIVE_SUNPY_TESTS"] = "True"
+    yield
+    del os.environ["PARFIVE_SUNPY_TESTS"]
+
+
+@pytest.fixture(scope='session', autouse=True)
 def hide_parfive_progress(request):
     """
     Set the PARFIVE_HIDE_PROGESS to hide the parfive progress bar in tests.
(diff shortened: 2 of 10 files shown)


sunpy/conftest.py
```diff
@@ -114,6 +114,16 @@ def undo_config_dir_patch():
...
 
 
+@pytest.fixture(scope='session', autouse=True)
+def parfive_header_test(request):
+    """
+    Add a keyword to tell parfive this is a test run.
+    """
+    os.environ["PARFIVE_SUNPY_TESTS"] = "True"
+    yield
+    del os.environ["PARFIVE_SUNPY_TESTS"]
```
nabobalis, line 124: Was not sure how to do this, so I did this as it was quick.

dstansby: Seems legit to me 👍

dstansby (changes requested):
- The `oldestdeps` version pin needs updating in `tox.ini`
- There should be a note in the changelog and in the what's new that says something like "We are aware the change in the parfive minimum version is a release earlier than our dependency policy allows for, but due to significant issues that parfive v2 solves we have decided to bump it now"
- Can you explain a bit more about why you've tweaked the user agent for the tests?

sunpy/util/parfive_helpers.py
```diff
@@ -14,36 +14,16 @@
 parfive_version = Version(parfive.__version__)
 sunpy_headers = {
     "User-Agent": f"sunpy/{sunpy.__version__} parfive/{parfive.__version__} "
-    f"aiohttp/{aiohttp.__version__} python/{sys.version[:5]}"
+    f"aiohttp/{aiohttp.__version__} python/{sys.version[:5]} "
 }
+if os.environ.get("PARFIVE_SUNPY_TESTS"):
```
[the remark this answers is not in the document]
nabobalis: I thought the `get` defaults to a None if the var is missing?

changelog/6742.feature.rst
```diff
@@ -3,6 +3,7 @@ The minimum required versions of several core dependencies have been updated:
 - Python 3.9
 - astropy 5.0.1
 - numpy 1.21.0
+- parfive 2.0.0
```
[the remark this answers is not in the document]
nabobalis: Fixed.

sunpy/util/parfive_helpers.py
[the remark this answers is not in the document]
dstansby: Huh, so it does, my mistake!

nabobalis:
> * Can you explain a bit more about why you've tweaked the user agent for the tests?

The motivation is to help the VSO or other providers to be able to see if requests coming from sunpy come either from our CI or actual users.

This is used to be more of a problem back in the day when we ran a larger online test suite but I think it would still be nice for data providers to be able to parse logs and know if our test suite is causing a problem or if its users.

dstansby (approved):

sunpy/conftest.py
Cadair: Yeah this seems good, but why don't we make this a generic environment var for being in our test suite? We might have other non-parfive uses for it in the future?

nabobalis: Was not sure what to call it, so I made one up.

Cadair (approved):

PR #6943, opened 2023-04-20
Use 'test' extra to isntall for tests on cibuildwheel

dstansby:
After `cairo` was introduced as a documentation dependency, [the wheel build started failing](https://github.com/sunpy/sunpy/actions/runs/4751488173/jobs/8440901890). We should only need to install the test (not the full dev) deps for tests, so change that and hopefully it fixes the wheel build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -174,7 +174,7 @@ jobs:
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@main
     with:
       sdist: false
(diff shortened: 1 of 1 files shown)


nabobalis (approved):

dstansby:
Hmm, for some reason the CRON CI doesn't want to run

dstansby:
I'm bored of trying to get publish CI to work 😭 , going to merge and we can see tomorrow if the publish CI is happy

PR #6949, opened 2023-04-23
Pass a scalar to skimage contour

dstansby:
Should fix the `oldestdeps` tests (e.g. see https://github.com/sunpy/sunpy/actions/runs/4771491361/jobs/8483459723 for a failing run). Also add an error if more than one level is given to `contour`, as it can only do one level at a time anyway.


nabobalis (approved):

nabobalis:
All cron and oldest deps CI is passing.

Cadair (approved):

PR #6950, opened 2023-04-23
Fix extras during cibuildwheel

dstansby:
Mistake came in as part of https://github.com/sunpy/sunpy/pull/6943.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -174,7 +174,7 @@ jobs:
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@main
     with:
       sdist: false
-      test_extras: 'test'
+      test_extras: 'all,tests'
       test_command: 'pytest -p no:warnings --doctest-rst -m "not mpl_image_compare" --pyargs sunpy'
       submodules: false
       targets: |


.github/workflows/ci.yml
```diff
@@ -174,7 +174,7 @@ jobs:
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@main
     with:
       sdist: false
-      test_extras: 'test'
+      test_extras: 'tests'
```
nabobalis, line 177:
```suggestion
      test_extras: 'all,tests'
```
maybe this will work?

dstansby: I don't think so, the idea here is not to install `all` because we don't want to pull in doc dependencies when testing.

nabobalis: all should not pull in docs. dev should.

dstansby: Ah, I didn't realise, thanks!

nabobalis: It does mean we need to be more careful now with using `dev` as it can break CI or maybe fail to install on peoples machines.

nabobalis (approved):

PR #6951, opened 2023-04-23
Remove setup.py

dstansby:
Two misc cleanups to `setup.py`:

- Specify the extras using `setup.cfg`, instead of using python code
- Use `pyproject.toml` for `setuptools_scm`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,3 @@
+from setuptools import setup
+
+setup()
--- a/setup.py
+++ b/setup.py
@@ -1,33 +1,10 @@
 #!/usr/bin/env python
 from setuptools import setup  # isort:skip
 import os
-from itertools import chain
 
 from extension_helpers import get_extensions
 
-try:
-    # Recommended for setuptools 61.0.0+
-    # (though may disappear in the future)
-    from setuptools.config.setupcfg import read_configuration
-except ImportError:
-    from setuptools.config import read_configuration
-
-################################################################################
-# Programmatically generate some extras combos.
-################################################################################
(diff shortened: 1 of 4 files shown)


nabobalis (approved):

setup.py
```diff
@@ -1,33 +1,6 @@
...
-ex_extras = dict(filter(lambda i: i[0] not in exclude_keys, extras.items()))
-# Concatenate all the values together for 'all'
-extras['all'] = list(chain.from_iterable(ex_extras.values()))
-
-setup(
-    extras_require=extras,
-    use_scm_version={'write_to': os.path.join('sunpy', '_version.py')},
-    ext_modules=get_extensions(),
-)
+setup(ext_modules=get_extensions())
```
Cadair, line 6: We don't even need this, you can do:
```
[extension-helpers]
use_extension_helpers = true
```

https://extension-helpers.readthedocs.io/en/latest/using.html

Cadair:
I think it's probably best if we keep a barebones setup.py file around for a little while longer. There's a discussion over on astropy with some reasons. I don't really see any downsides to doing so if we have no code or config in there.

dstansby:
👍 , I'm guessing https://github.com/astropy/astropy/pull/14361 is the xref for anyone interested

nabobalis:
So we should leave this for 5.1?

dstansby:
I think we can put it in, I just need to put a `setup.py` file back that calls `setup()`.

Cadair (approved):
lgtm, the next thing to do is move all the setup.cfg stuff to pyproject.toml

PR #6952, opened 2023-04-23
Clean cibuildwheel config

dstansby:
- Copy some config to `pyproject.toml`
- Enable testing on Python 3.11
- Remove openCV pin (the issue that caused us to pin it should be fixed now)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,8 +13,8 @@ write_to = "sunpy/_version.py"
 
 # Until we have h5py binaries we can't test wheels on Python 3.11
 [tool.cibuildwheel]
-test-skip = "cp311-*"
-test-requires = ["opencv_python==4.6.0.66"]
+test-command = "pytest -p no:warnings --doctest-rst -m 'not mpl_image_compare' --pyargs sunpy"
+test-extras = ["all,test"]
 
 [ tool.gilesbot ]
   [ tool.gilesbot.circleci_artifacts ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ write_to = "sunpy/_version.py"
 # Until we have h5py binaries we can't test wheels on Python 3.11
 [tool.cibuildwheel]
 test-command = "pytest -p no:warnings --doctest-rst -m 'not mpl_image_compare' --pyargs sunpy"
-test-extras = ["all,test"]
+test-extras = ["all,tests"]
 
 [ tool.gilesbot ]
   [ tool.gilesbot.circleci_artifacts ]


dstansby:
Hmm, for some reason cibuildwheel isn't picking up the config in `pyproject.toml`

.github/workflows/ci.yml
```diff
@@ -174,8 +174,6 @@ jobs:
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@main
     with:
       sdist: false
-      test_extras: 'test'
-      test_command: 'pytest -p no:warnings --doctest-rst -m "not mpl_image_compare" --pyargs sunpy'
```
ConorMacBride, line 178: One of the limitations of the OpenAstronomy workflow is that these two options cannot be specified in `pyproject.toml`. [These four cibuildwheel options](https://github.com/OpenAstronomy/github-actions-workflows/blob/bb12d5420b890a2102eea80fbfbda86e626baec4/.github/workflows/publish.yml#L165-L168) are set as env vars in the workflow, which override the `pyproject.toml` config. You can see [in the build logs](https://github.com/sunpy/sunpy/actions/runs/4779231611/jobs/8496280260#step:6:72) that it skipped testing as the env var was an empty string.

dstansby: 😆 snap, I just opened an issues about this over there

dstansby:
🎉 wheel tests are being run and passing on Python 3.11

nabobalis (approved):

pyproject.toml
```diff
@@ -10,8 +10,8 @@ build-backend = 'setuptools.build_meta'
 
 # Until we have h5py binaries we can't test wheels on Python 3.11
 [tool.cibuildwheel]
-test-skip = "cp311-*"
-test-requires = ["opencv_python==4.6.0.66"]
+test-command = "pytest -p no:warnings --doctest-rst -m 'not mpl_image_compare' --pyargs sunpy"
+test-extras = ["all,test"]
```
ConorMacBride, line 14:
```suggestion
test-extras = ["all,tests"]
```

ConorMacBride (commented):
I've opened https://github.com/OpenAstronomy/github-actions-workflows/pull/127 which should allow the `test_extras` and `test_command` workflow inputs to be removed. Although, we'll still need these inputs for the sdist and pure Python wheel jobs.

Cadair (changes requested):
We need the other build fails to get out of the way for this to run the publish step.

Cadair (approved):
Looks good to me as long as all the CI passes, including publish.

Cadair:
wow, a completely green CI run, that never happens.

Issue #6989, opened 2023-05-03
Sections are at wrong level in the navbar


wtbarnes:
All of the toctree pages are a subheading of the last page. For an example, see https://github.com/sunpy/sunpy/blob/a6c1417367fe63abc7f84c851b05974af2bc7634/docs/tutorial/index.rst?plain=1#L26-L35. This happens at several places throughout our documentation. See screenshots below.

<img width="503" alt="image" src="https://user-images.githubusercontent.com/8676141/235982718-bd6cc780-87c9-4f41-b294-ca60781c9796.png">

<img width="332" alt="image" src="https://user-images.githubusercontent.com/8676141/235982853-87a2a5f0-97c9-45de-818e-19b096af2a2d.png">

wtbarnes:
Also in the coordinates topic guide,

![image](https://user-images.githubusercontent.com/8676141/236215742-fded4bc8-14fa-4aeb-b076-b169b4af9aa5.png)

The more advanced topics are listed inthe toctree and so they show up in the main toctree before the more introductory topics!

PR #6954, opened 2023-04-23
Re-organise some coord docs

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/6913.

A lot of the tutorial content was duplicated. In the guide page, there was also some accessing coordinate components content that felt like a how to guide, so I moved that there. I'm open to a bit of bikeshedding in this PR, but would be good to get this in realtively quickly in time for 5.0.

PR #6991, opened 2023-05-04
Replace some TOC headings with raw html to fix sidebar.

alasdairwilson:
This partially solves #6989

The issue is caused by headings in the tutorial/how to page becoming top level toc entries, by turning these into raw html headings this behaviour is avoided.

The issue also has a comment regarding the coords TOC but this can't be tackled till after #6954

### DIFF
--- a/docs/guide/index.rst
+++ b/docs/guide/index.rst
@@ -7,8 +7,8 @@ Topic Guides
 These guides provide a set of in-depth explanations for various parts of the ``sunpy`` core package.
 They assumes you have some knowledge of what ``sunpy`` is and how it works - if you're starting fresh you might want to check out the :ref:`tutorial` first.
 
-How to use the guides
----------------------
+**How to use the guides**
+
 The guides are designed to be read in a standalone manner, without running code at the same time.
 Although there are code snippets in various bits of the guides, these are to help explanation, and are not structured in a way that they can be run as you're reading each guide.
 
--- a/docs/how_to/index.rst
+++ b/docs/how_to/index.rst
@@ -20,9 +20,7 @@ If you're starting fresh you might want to check out the :ref:`tutorial` first.
    create_a_map
    remote_data_manager
 
-
-Quick Reference
----------------
+**Quick Reference**
 
 The following table is meant as a quick reference.
 For more complete code examples, see the how-to guides above.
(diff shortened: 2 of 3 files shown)


nabobalis (approved):

dstansby (changes requested):
This seems a bit hacky, and not a sustainable solution if/when the pages get modified. What happens if the toctrees are moved above the first heading on the page instead? Maybe then they'll be treated as children of the page they're on, instead of children of a section within that page.

Cadair:
> This seems a bit hacky

oh yes.

> not a sustainable solution if/when the pages get modified

Why specifically? Yes it's weird, but it's just inline content, we aren't breaking references or anything else with this hack.

> What happens if the toctrees are moved above the first heading on the page instead?

This works fine, but then you either have the whole contents list above the intro which is weird, or you don't have the contents list in the index page at all (i.e. make the toctree hidden).

dstansby:
I was thinking more from the point of view of new contributiors (or old ones) coming along later and wondering why these aren't just RST headings.

Alternative fix, we could just drop the headings altogether? They don't seem super neccesary to have.

Cadair:
My solution was going to be to make them bold.

alasdairwilson:
One solution that came to mind would be to have a hidden toctree for the sidebar and then a manually generated listed that looks like the current section listing but this is hard to maintain because future changes to the toc or ordering of the sections aren't changed in both places automatically.

Despite this being hacky, I think this html solution is maintainable. I think it is clear even by a glance at what the sphinx directive is doing, inserting a heading into the generated html, and it would be pretty much impossible to break them regardless of where the toc on this page is moved. Someone might not be able to work out why regular rst section headings aren't being used but I don't think its that obtuse, certainly not the most egregious patch anyway, (and if they changed them to rst headings they would probably immediately clock the reasoning).

I don't mind the headings going (or going bold) but I do think that the page is better with the headings than without so the question is is the tradeoff of using this directive and perhaps being unclear worth it.

dstansby:
I'd be 👍 just making them bold

wtbarnes:
> I'd be 👍 just making them bold

I'm in this camp too I think. I think the headings are still obvious, especially since they are mostly at the top of the page anyway.

wtbarnes:
Do we want to hold this up until #6954 is merged so it can fix those headings too?

dstansby (approved):

nabobalis:
> Do we want to hold this up until #6954 is merged so it can fix those headings too?

I don't see why not, might as well do it all in one go.

Cadair (approved):

PR #6995, opened 2023-05-05
Prioritise dateavg over dateobs in wcs_frame_mapping

### LINKED ISSUE #6994: `sunpy.coordinates.wcs_utils.solar_wcs_frame_mapping` only checks `dateobs` not `dateavg`
### Describe the bug

When looks for the obstime of a WCS the `solar_wcs_frame_mapping` function should prioritise `dateavg` as it's the more modern reference time for the WCS.

Note that the DKIST data doesn't have `DATE-OBS` so this just fails to extract a obstime at all for DKIST data.

### To Reproduce

_No response_

### Screenshots

_No response_

### System Details

_No response_

### Installation method

_No response_

Cadair:
fixes #6994

### DIFF
--- a/changelog/6995.bugfix.rst
+++ b/changelog/6995.bugfix.rst
@@ -1,0 +1,3 @@
+When converting an `astropy.wcs.WCS` object to a solar coordinate frame the
+``DATE-AVG`` key will be used before the ``DATE-OBS`` key, previously only
+``DATE-OBS`` was checked.
--- a/sunpy/coordinates/tests/test_wcs_utils.py
+++ b/sunpy/coordinates/tests/test_wcs_utils.py
@@ -46,6 +46,25 @@ def test_wcs_frame_mapping_none():
     assert result is None
 
 
+def test_wcs_frame_mapping_dateavg():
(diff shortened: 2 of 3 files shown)


Cadair:
Only question is if we should backport this very subtle and hidden behaviour change?

nabobalis:
> Only question is if we should backport this very subtle and hidden behaviour change?

Let 4.0 and 4.1 rest.

Cadair:
> > Only question is if we should backport this very subtle and hidden behaviour change?
> 
> Let 4.0 and 4.1 rest.

Well, if we decide it's worthy of a backport I would quite like to get it backported so the DKIST data works properly

nabobalis:
> > > Only question is if we should backport this very subtle and hidden behaviour change?
> > 
> > 
> > Let 4.0 and 4.1 rest.
> 
> Well, if we decide it's worthy of a backport I would quite like to get it backported so the DKIST data works properly

Ah, I guess making dkist user tools use 5.0 only might be too soon.

In that case, lets do the full backport.

nabobalis (approved):

PR #7000, opened 2023-05-09
Add topic guide describing ``rsun`` idiosyncracies

alasdairwilson:
This takes some of the content from coords regarding rsun and its involvement in transforms/reprojects expands on it and makes a topic guide.

If anyone else can think of additional content that should be here and feedback on current content extremely welcome.


docs/guide/rsun.rst
```diff
@@ -0,0 +1,62 @@
...
+    >>> from sunpy.data.sample import AIA_193_JUN2012, STEREO_A_195_JUN2012
+    >>> map_aia = Map(AIA_193_JUN2012)
+    >>> map_euvi = Map(STEREO_A_195_JUN2012)
+    >>> outmap1 = map_euvi.reproject_to(map_aia.wcs)
+    >>> map_euvi.meta['rsun_ref'] = map_aia.meta['rsun_ref']
+    >>> outmap2 = map_euvi.reproject_to(map_aia.wcs)
+
+we can see the difference in the appearance of the reprojected maps at the limb.
+
+Transforming between ``helioprojective`` frames
```
Cadair, line 42: Would it make more sense to have this first, as the reproject_to example is technically a special case of this.

docs/guide/rsun.rst
```diff
@@ -0,0 +1,62 @@
...
+:meth:`~sunpy.map.GenericMap.reproject_to` by default enforces a round-trip behavior,
+the idea being that you should only trust the reprojection if the 2D coordinates from each observer both resolve to the same 3D point (within a pixel volume).
+When the values for ``rsun`` are different, that criterion fails towards the limb.
+In other cases, this furthers results in banding as the criterion fails, then succeeds and then fails again.
+
+Changing the value of ``rsun`` can fix this issue:
+Take the example, :ref:`sphx_glr_generated_gallery_map_transformations_reprojection_different_observers.py`
+If we run the reprojection twice, once before fixing the discrepancy in the ``rsun_ref`` metadata and once after:
+
+.. code-block:: python
```
Cadair, line 29: Neither of these examples output anything? Is that deliberate?

Cadair (commented):
Thanks! This is an excellent use of the topic guide.

wtbarnes:
This will conflict with #6954. I would suggest merging that first since it does additional reorganization as well. It maybe that all of the info in that (very incomplete) topic guide is completely redundant compared to the more complete information here.

docs/guide/rsun.rst
```diff
@@ -0,0 +1,70 @@
+.. _topic-guide-rsun:
+
+*********************************
+The role of ``rsun`` in ``sunpy``
+*********************************
```
nabobalis, line 5:
```suggestion
***************************
The role of "rsun" in sunpy
***************************
```

docs/guide/rsun.rst
```diff
@@ -0,0 +1,70 @@
...
+
+When dealing with many solar coordinate systems, ``rsun``, the height at which emission is presumed to originate, can be important.
```
nabobalis, line 7:
```suggestion
When dealing with many solar coordinate systems, "rsun", the height at which emission is presumed to originate, can be important.
```

docs/guide/rsun.rst
```diff
@@ -0,0 +1,70 @@
...
+For example, when loading a fits file with `sunpy.map.Map` a ``map.wcs`` is constructed with a value for ``rsun`` collected for use in coordinate transformations.
```
nabobalis, line 8:
```suggestion
For example, when loading a fits file with `sunpy.map.Map` a ``map.wcs`` is constructed with a value for "rsun" collected for use in coordinate transformations.
```

docs/guide/rsun.rst
```diff
@@ -0,0 +1,70 @@
...
+
+Data providers often include this information for end users via the fits keyword: ``RSUN_REF``,
+where this keyword is not available ``sunpy`` assumes a standard value defined by `sunpy.sun.constants`.
```
nabobalis, line 11:
```suggestion
Data providers often include this information for end users via the fits keyword: "RSUN_REF", where this keyword is not available, ``sunpy`` assumes a standard value defined by `sunpy.sun.constants`.
```

docs/guide/rsun.rst
```diff
@@ -0,0 +1,70 @@
...
+In theory, the radiation detected by a single filter, especially for coronal emission,
+comes from a variety of heights and ``rsun`` is only a loose approximation.
```
nabobalis, line 13:
```suggestion
In theory, the radiation detected by a single filter, especially for coronal emission, comes from a variety of heights and "rsun" is only a loose approximation.
```

docs/guide/rsun.rst
```diff
@@ -0,0 +1,70 @@
...
+In practice, the exact value of ``rsun`` is important when working with coordinates and transforms, it provides 3 dimensional information about the location of a map pixel.
```
nabobalis, line 14:
```suggestion
In practice, the exact value of "rsun" is important when working with coordinates and transforms, it provides 3 dimensional information about the location of a map pixel.
```

docs/guide/rsun.rst
```diff
@@ -0,0 +1,70 @@
...
+When manipulating these coordinates, idiosyncrasies and unexpected behavior can be encountered which can be preempted by an understanding of behaviors surrounding ``rsun``.
```
nabobalis, line 15:
```suggestion
When manipulating these coordinates, idiosyncrasies and unexpected behavior can be encountered which can be preempted by an understanding of behaviors surrounding "rsun".
```

docs/guide/rsun.rst
nabobalis: For me, rsun is not code but a keyword, so I think it should like this and the changes below follow that approach.

docs/guide/rsun.rst
```diff
@@ -0,0 +1,70 @@
...
+
+Transforming between ``helioprojective`` frames
+===============================================
```
nabobalis, line 18:
```suggestion
Transforming between Helioprojective frames
===========================================
```
Not sure if that is correctly titled.

docs/guide/rsun.rst
```diff
@@ -0,0 +1,70 @@
...
+
+  >>> hpc1 = SkyCoord(0*u.arcsec, 0*u.arcsec, frame=aia.coordinate_frame)  # doctest: +SKIP
+  >>> hpc2 = hpc1.transform_to(stereo.coordinate_frame)  # doctest: +SKIP
+
+Reprojecting between frames with different ``rsun``
+===================================================
+
+When the values of ``rsun`` from two wcs instances are different, issues with reprojecting between those frames can be encountered:
+:meth:`~sunpy.map.GenericMap.reproject_to` by default enforces a round-trip behavior,
+the idea being that you should only trust the reprojection if the 2D coordinates from each observer both resolve to the same 3D point (within a pixel volume).
```
nabobalis, line 44:
```suggestion
When the values of "rsun" from two wcs instances are different, issues with reprojecting between those frames can be encountered.
:meth:`~sunpy.map.GenericMap.reproject_to` by default, enforces a round-trip behavior, the idea being that you should only trust the reprojection if the 2D coordinates from each observer both resolve to the same 3D point (within a pixel volume).
```

docs/guide/rsun.rst
```diff
@@ -0,0 +1,70 @@
...
+When the values for ``rsun`` are different, that criterion fails towards the limb.
+In other cases, this furthers results in banding as the criterion fails, then succeeds and then fails again.
+
+Changing the value of ``rsun`` can fix this issue:
+Take the example, :ref:`sphx_glr_generated_gallery_map_transformations_reprojection_different_observers.py`
+If we run the reprojection twice, once before fixing the discrepancy in the ``rsun_ref`` metadata and once after:
```
nabobalis, line 50:
```suggestion
Changing the value of "rsun" can fix this issue.
Take the example :ref:`sphx_glr_generated_gallery_map_transformations_reprojection_different_observers.py`.
If we run the reprojection twice, once before fixing the discrepancy in the "rsun_ref" metadata and once after:
```

nabobalis (commented):
Another really minor review

nabobalis:
It unfortunately conflicts with main as well now.

ayshih (commented):
I don't want to block this PR, but here is the structure and content this topic guide eventually needs to have, which could be deferred to a follow-up PR:

- The title should be something like "Different values for the radius of the Sun"
- There ought to be an introduction that explains the radius of the Sun:
```
Observing in visible wavelengths
The meaning and context of the IAU constant
Observing in wavelengths other than visible (e.g., EUV)
```
-  The second section should describe the `rsun` frame attribute
```
2D coordinates
Converting 2D coordinates to 3D coordinates, with examples
```
- The third section should describe how the radius of the Sun is determined for data files
```
RSUN_REF keyword, or calculated from RSUN_OBS keyword, or default
Why Maps from different data sources might have different `rsun` values for the same emission
Why the precise value of the radius of the Sun typically doesn't matter
```
- The fourth section then goes into reprojection

docs/guide/rsun.rst
```diff
@@ -0,0 +1,69 @@
+.. _topic-guide-rsun:
```
Cadair, line 1:
```suggestion
.. _sunpy-topic-guide-rsun:
```

Cadair (approved):

docs/guide/rsun.rst
nabobalis: I FORGOT ONE?!

docs/guide/rsun.rst
```diff
@@ -0,0 +1,69 @@
+.. _sunpy-topic-guide-rsun:
```
nabobalis, line 1:
```suggestion
.. _sunpy-topic-guide-role-of-rsun:
```
I made some of them quite verbose in my PR and I think that would be nice here as well :P

nabobalis:
We should create a tracking issue for Alberys comment.

nabobalis (approved):

Issue #7003, opened 2023-05-09
Setting rcparams in STEREO example gallery entry globally alters plots in tutorial


wtbarnes:
The ["where is STEREO" plot example gallery](https://docs.sunpy.org/en/latest/generated/gallery/showcase/where_is_stereo.html#sphx-glr-generated-gallery-showcase-where-is-stereo-py) sets the `rcparams` to alter several aspects of the plot appearance. This seems to have the unintentional effect of altering the appearance of the plots in the tutorial as well. For example, see https://sunpy--6944.org.readthedocs.build/en/6944/tutorial/maps.html#basic-plotting. Instead, the rcparams in that example should be changed only in a limited context.

dstansby:
Fixed by

PR #7005, opened 2023-05-11
Reset default rcParams in STEREO example

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/7003


nabobalis:
I think this commit would be better in Wills Map PR since we can confirm it works.

Cadair:
or we merge that and then deal with it.

nabobalis:
Or that.

dstansby:
Whatever, if someone wants to cherry pick it over there feel free

dstansby:
I think the version we already merged is a bit clearer, so I'll close this.

PR #7018, opened 2023-05-23
enabled passing of corrupt files to map

hannahc243:
Fixes https://github.com/sunpy/sunpy/issues/3688 which allows passing of corrupt files to map


hayesla:
Thanks for the PR @hannahc243 and congrats on your first contribution ! 🥳

ayshih (commented):
Thanks for the PR!  Some immediate thoughts:

- Don't add a "bad" test file to the repository, much less one that is named to look like a real file!  Such an approach is bound to cause confusion and breakages, as you've already discovered with the unrelated unit test that you had to separately fix.  Instead, define a pytest fixture that creates a temporary "bad" test file for use in the unit tests.
- Naming the keyword "silence_errors" makes me uneasy because it still emits warnings, so is definitely not silent.  I'd probably name it something like "skip_errors", but others might have better suggestions.

ayshih:
Maybe `allow errors`?

nabobalis:
> * Don't add a "bad" test file to the repository, much less one that is named to look like a real file!  Such an approach is bound to cause confusion and breakages, as you've already discovered with the unrelated unit test that you had to separately fix.  Instead, define a pytest fixture that creates a temporary "bad" test file for use in the unit tests.

I don't see the harm in a bad test file, I was the one who told them to add it, the filename I think we can work on. It would be nice to have a file that is from a failed download, that we can use easily. 

That unrelated unit test I don't think even needs to exist. 

> * Naming the keyword `silence_errors` makes me uneasy because it still emits warnings, so is definitely not silent.  I'd probably name the keyword something like `skip_errors`, but others might have better suggestions.

That `silence_errors` already exists in the Map API, I wanted to avoid adding a new keyword. If you prefer, we can add a new keyword that will handle this behaviour, but since "silence_errors" already does this but for hdu pairs that are passed, extending it to do it for files seems fine to me.

nabobalis:
> Well, then I additionally prefer to change the name of the keyword that already exists in the API.  It doesn't make sense that a keyword called `silence_errors` does not actually silence anything, but instead replaces errors with (non-silent) warnings.  

I was hoping you wouldn't say this but I agree.

changelog/7018.feature.rst
```diff
@@ -0,0 +1 @@
+We added the ability to pass corrupt fits files to map, using the existing keyword argument ``silence_errors``.
```
alasdairwilson, line 1: Maybe something like this for the changelog?:

```suggestion
Added the ability to skip over errors raised when passing invalid fits files to map using the existing keyword argument ``silence_errors``.
```

sunpy/data/test/aia_lev1_211a_2022_03_30t17_21_21_63z_image_lev1.fits
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
```
nabobalis, line 1: We should rename this file to something else: `not_actually_a_fits.fits`

sunpy/io/tests/test_fits.py
```diff
@@ -138,7 +138,8 @@ def test_fitsheader():
     """Test that all test data can be converted back to a FITS header."""
     extensions = ('.fts', '.fits')
     for ext in extensions:
-        test_files = [f for f in get_test_data_filenames() if f.suffix == ext]
+        bad_file = "aia_lev1_211a_2022_03_30t17_21_21_63z_image_lev1.fits"
```
nabobalis, line 141:
```suggestion
        bad_file = "not_actually_fits.fits"
```
When the file has been renamed.

sunpy/map/tests/test_map_factory.py
```diff
@@ -315,3 +315,21 @@ def test_map_fits():
...
     assert isinstance(fits_map, sunpy.map.GenericMap)
     assert fits_map.data.base is not None
+
+def test_map_list_of_files():
+    files = [AIA_171_IMAGE, get_test_filepath('aia_lev1_211a_2022_03_30t17_21_21_63z_image_lev1.fits')]
+    with pytest.warns(SunpyUserWarning, match='Failed to read'):
+        amap = sunpy.map.Map(files, silence_errors=True)
+        assert amap.data.shape == (128,128)
+
+    files = [AIA_171_IMAGE, get_test_filepath('aia_lev1_211a_2022_03_30t17_21_21_63z_image_lev1.fits'), AIA_171_IMAGE]
```
nabobalis, line 325:
```suggestion
    files = [AIA_171_IMAGE, get_test_filepath('not_actually_fits.fits'), AIA_171_IMAGE]
```
When the file has been renamed.

changelog/7018.feature.rst
```diff
@@ -0,0 +1 @@
+Added the ability to skip over errors raised when passing invalid fits files to map using the existing keyword argument ``silence_errors``.
```
hayesla, line 1:
```suggestion
Added the ability to skip over errors raised for invalid fits files when passing a list of files to map using the existing keyword argument ``silence_errors``.
```

hayesla (approved):
Thanks for this PR @hannahc243 !

Looks great, just two suggestions

sunpy/map/map_factory.py
```diff
@@ -97,7 +97,11 @@ def _read_file(self, fname, **kwargs):
         try:
             pairs = read_file(os.fspath(fname), **kwargs)
         except Exception as e:
-            raise OSError(f"Failed to read {fname}\n{e}") from e
+            msg = f"Failed to read {fname}\n{e}"
+            if kwargs.get("silence_errors"):
+                warn_user(msg)
+                return []
+            raise OSError(msg) from e
```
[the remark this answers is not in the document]
nabobalis: I think this is worth adding to my follow up PR renaming the keyword.

nabobalis (approved):

alasdairwilson (approved):

nabobalis:
Thank you for the pull request @hannahc243!

PR #7024, opened 2023-05-24
Add support for Map to load from a generator 

### LINKED ISSUE #7022: Map factory does not support `pathlib.Path.glob`
### Describe the feature

`pathlib.Path.glob` produces a generator for the list of files that match the glob pattern in a particularly directory. However, this not a supported pattern for the map factory. This was surprising to me since we support `pathlib.Path` objects as well as more traditional glob patterns as well as lists.

As an example, I would expect the following to work,

```python
import pathlib
import sunpy.map
sunpy.map.Map(pathlib.Path.home().glob('*.fits'))
```
but instead it throws the following exception
```python-traceback
Cell In [16], line 1
----> 1 sunpy.map.Map(pathlib.Path.home().glob('*.fits'))

File ~/Documents/codes/sunpy/sunpy/map/map_factory.py:254, in MapFactory.__call__(self, composite, sequence, silence_errors, *args, **kwargs)
    228 def __call__(self, *args, composite=False, sequence=False, silence_errors=False, **kwargs):
    229     """ Method for running the factory. Takes arbitrary arguments and
    230     keyword arguments and passes them to a sequence of pre-registered types
    231     to determine which is the correct Map-type to build.
   (...)
    252     as `memmap` for FITS files.
    253     """
--> 254     data_header_pairs = self._parse_args(*args, silence_errors=silence_errors, **kwargs)
    255     new_maps = list()
    257     # Loop over each registered type and check to see if WidgetType
    258     # matches the arguments.  If it does, use that type.

File ~/Documents/codes/sunpy/sunpy/map/map_factory.py:180, in MapFactory._parse_args(self, silence_errors, *args, **kwargs)
    178 for arg in args:
    179     try:
--> 180         data_header_pairs += self._parse_arg(arg, **kwargs)
    181     except NoMapsInFileError as e:
    182         if not silence_errors:

File ~/Documents/codes/sunpy/sunpy/util/functools.py:18, in seconddispatch.<locals>.wrapper(*args, **kwargs)
     17 def wrapper(*args, **kwargs):
---> 18     return dispatcher.dispatch(args[1].__class__)(*args, **kwargs)

File ~/Documents/codes/sunpy/sunpy/map/map_factory.py:195, in MapFactory._parse_arg(self, arg, **kwargs)
    189 @seconddispatch
    190 def _parse_arg(self, arg, **kwargs):
    191     """
    192     Take a factory input and parse into (data, header) pairs.
    193     Must return a list, even if only one pair is returned.
    194     """
--> 195     raise ValueError(f"Invalid input: {arg}")

ValueError: Invalid input: <generator object Path.glob at 0x104dc71b0>
```

### Proposed solution

Add this as a possible input to the map factory.

Jett-Code:
## PR Description
Fixes #7022


nabobalis:
This will need a unit test as well to check that pathlib.glob works.

Jett-Code:
> Looks like this change has failed on the CI, this will need debugging.
> 
> This will need a unit test as well to check that pathlib.glob works.

Could you please guide me on how to write unit tests for this?

Jett-Code:
> You will want to add extra unit tests to https://github.com/sunpy/sunpy/blob/main/sunpy/util/tests/test_util.py to ensure that we skip over strings correctly.

There are already two unit test cases for string, list, and tuples. Do you want me to add extra on top of that?
`lst = ['a', 'b', [], (['c', 'd']), tuple(), ['e']]`

nabobalis:
> There are already two unit test cases for string, list, and tuples. Do you want me to add extra on top of that? `lst = ['a', 'b', [], (['c', 'd']), tuple(), ['e']]`

If the tests for those are already there, then its ok. But what if you just pass in a string?

The other unit tests are failing with these changes, we need to debug this.

sunpy/map/tests/test_map_factory.py
```diff
@@ -86,6 +86,10 @@ def test_patterns(eit_fits_directory):
     # Pathlib
     path = pathlib.Path(AIA_171_IMAGE)
     aiamap = sunpy.map.Map(path)
+
+    maps = sunpy.map.Map(list(path.glob('*.fits')))
```
nabobalis, line 90: Why does this need to be wrapped a in list first?

Jett-Code: Yeah! got it will correct it ASAP

Jett-Code:
>

sunpy/map/tests/test_map_factory.py
Jett-Code: I have got a small question. Which class does this object belong to? list?

nabobalis: Which object?

PaulJWright: Have you had the time to correct this?

PaulJWright (changes requested):

sunpy/util/util.py
```diff
@@ -9,6 +9,7 @@
 from shutil import get_terminal_size
 from itertools import chain, count
 from collections import UserList
```
PaulJWright, line 11: This can be removed as it is no longer needed (causing `pre-commit.ci` to fail)

nabobalis:
Hello @Jett-Code, sorry for not getting back to you after a long time. WOuld you have some time soon to add a few unit tests for this?

Jett-Code:
> Hello @Jett-Code, sorry for not getting back to you after a long time. WOuld you have some time soon to add a few unit tests for this?

Sure, on it!

nabobalis:
Actually I think that one glob test is enough, I don't know why I asked for more.

Could you rebase this and we will get this reviewed!

nabobalis (approved):

sunpy/util/util.py
```diff
@@ -92,12 +94,13 @@ def expand_list(inp):
     ----------
     * https://stackoverflow.com/questions/2185822/expanding-elements-in-a-list/2185971#2185971
     """
-    return [item for item in expand_list_generator(inp)]
+    return list(expand_list_generator(inp))
 
 
 def expand_list_generator(inp):
+    type_union = (str | bytes | np.ndarray | dict)
```
nabobalis, line 101: This has to be wrong.

Jett-Code: Got it!

sunpy/util/util.py
```diff
@@ -92,12 +97,13 @@ def expand_list(inp):
     ----------
     * https://stackoverflow.com/questions/2185822/expanding-elements-in-a-list/2185971#2185971
     """
-    return [item for item in expand_list_generator(inp)]
+    return list(expand_list_generator(inp))
 
 
 def expand_list_generator(inp):
+    type_union = (str | bytes | np.ndarray | dict | WCS | Header)
```
nabobalis, line 104: A lot of extra are Iterable, there has to be a better class to avoid forming this list.

Jett-Code: Instead of checking if the input is an instance of a specific set of types, I can check if it is an instance of the `collections.abc.Collection` class, right?

nabobalis: Maybe that would work?

sunpy/util/util.py
```diff
@@ -92,12 +92,14 @@ def expand_list(inp):
     ----------
     * https://stackoverflow.com/questions/2185822/expanding-elements-in-a-list/2185971#2185971
     """
-    return [item for item in expand_list_generator(inp)]
+    return list(expand_list_generator(inp))
 
 
 def expand_list_generator(inp):
+    exclude_union = (str | bytes)
+    include_union = (list | tuple | Iterator)
```
nabobalis, line 100: Version 3

sunpy/util/util.py
Jett-Code: Are there any ways to find out if it's working or not? I'm confused

nabobalis: The unit tests.

sunpy/util/util.py
Jett-Code: It's good if it's working, right? or I'm missing something

nabobalis: Understanding why it works is what I need to learn.

sunpy/util/util.py
```diff
@@ -92,12 +92,14 @@ def expand_list(inp):
...
     for item in inp:
-        if isinstance(item, tuple | list | UserList):
+        if not isinstance(item, exclude_union) and isinstance(item, include_union):
```
nabobalis, line 102:
```suggestion
        if isinstance(item, include_union):
```
Would this be enough?

Jett-Code: I don't think so, what if someone passes a byte string?

nabobalis: Would that not be caught by this check? 

Might be worth adding some unit tests to check `expand_list_generator` behavior?

Jett-Code: We are checking for the list which in turn checks for the generator as input which solves the problem but here I'm adding an extra check for byte string

nabobalis: But why would a byte string hit either of those categories?

sunpy/util/util.py
```diff
@@ -97,7 +98,7 @@ def expand_list(inp):
 
 def expand_list_generator(inp):
     for item in inp:
-        if isinstance(item, (tuple, list, UserList)):
+        if not isinstance(item, str) and isinstance(item, Iterable):
```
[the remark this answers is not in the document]
Jett-Code: > While I doubt someone will pass a byte string in the context of this MR, as this is a utility function, I believe you want to avoid an issue with byte strings. In the current implementation a byte string wouldn't be caught by `isinstance(item, str)` as a regular string or raw string literal would be.

As suggested by @PaulJWright

sunpy/util/util.py
Jett-Code: > While I doubt someone will pass a byte string in the context of this MR, as this is a utility function, I believe you want to avoid an issue with byte strings. In the current implementation a byte string wouldn't be caught by `isinstance(item, str)` as a regular string or raw string literal would be.

As suggested by @PaulJWright

nabobalis: Sure, but I mean why would a byte string now trigger this line of code?

Jett-Code: It triggers exclude_union which in turn triggers a string or byte

Jett-Code: I guess we don't want any unit tests for byte strings

nabobalis: But in this suggestion, we only check for include_union which is list, tuple and Iterator. 

I would not have assumed a byte string would trigger the last one but I don't know what Iterator fully covers.

nabobalis: Why would we not want a unit test for it?

nabobalis: We should expand the unit test to see what happens if we use a numpy array, a fits header, a astropy WCS, a byte string and maybe some others?

Jett-Code: Yeah, this suggestion doesn't covers exclude_union.

Jett-Code: > Why would we not want a unit test for byte strings?

If we have to remove byte dependency then there is no need of byte test cases

nabobalis: > Yeah, this suggestion doesn't covers exclude_union.

Yeah, since we I don't we would need it.

nabobalis: > > Why would we not want a unit test for byte strings?
> 
> If we have to remove byte dependency then there is no need of byte test cases

I am not sure I follow?

Jett-Code: Ah! Never mind. I'll add unit tests for the data types and objects mentioned above

nabobalis: I updated the test and the code, see what you think.

Jett-Code: Seems good. What about Astropy WCS object and numpy, should I add them?

nabobalis: Those should be covered by the .data and .wcs from the sunpy map.

sunpy/util/tests/test_util.py
```diff
@@ -48,11 +53,20 @@ def test_expand_list():
     lst = [1, 2, 3, [4, 5, 6], 7, (8, 9), ((10, 11), ((12, 13),))]
     assert util.expand_list(lst) == [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13]
 
-
-def test_expand_list_generator():
-    lst = ['a', 'b', [], (['c', 'd']), tuple(), ['e']]
-    assert list(util.expand_list_generator(lst)) == ['a', 'b', 'c', 'd', 'e']
-
+@pytest.mark.filterwarnings("ignore:Invalid 'BLANK' keyword in header")
+def test_expand_list_generator(aia171_test_map):
```
nabobalis, line 57: Oh I should check a file handle?

Jett-Code: Yeah, you can exclude them from being yielded or flattened by adding checks

sunpy/util/util.py
```diff
@@ -92,12 +92,12 @@ def expand_list(inp):
...
     * https://stackoverflow.com/questions/2185822/expanding-elements-in-a-list/2185971#2185971
     """
-    return [item for item in expand_list_generator(inp)]
+    return list(expand_list_generator(inp))
 
 
 def expand_list_generator(inp):
     for item in inp:
-        if isinstance(item, tuple | list | UserList):
+        if isinstance(item, (list | tuple | Iterator)):
```
Jett-Code, line 100: This syntax is not supported for earlier versions of python, should I refactor it?

Jett-Code: We don't want to flattened strings or byte string, that's why we have used `not isinstance(item, (bytes, str))` so that it goes through `yield item` and not through `yield from expand_list_generator(item)`

nabobalis: > This syntax is not supported for earlier versions of python, should I refactor it?

sunpy main supports python 3.10 or above, I don't think this should be a problem.

> We don't want to flattened strings or byte string, that's why we have used `not isinstance(item, (bytes, str))` so that it goes through `yield item` and not through `yield from expand_list_generator(item)`

Oh I didn't add a test with a longer string, I should add that to check.

sunpy/util/tests/test_util.py
```diff
@@ -48,11 +53,23 @@ def test_expand_list():
...
+    zip_generator = zip(['1', '2'], ['3', '4'])
+    first_list = ['a1234', 'b', [], (['c', 'd']), tuple(), ['e'], b'fghj']
+    # The empty list and tuple should be ignored
+    assert list(util.expand_list_generator(first_list)) == ['a1234', 'b', 'c', 'd', 'e', b'fghj']
+    assert list(util.expand_list_generator([aia171_test_map.data, aia171_test_map.wcs, header])) == [aia171_test_map.data, aia171_test_map.wcs, header]
+    assert list(util.expand_list_generator(generator)) == ['c', 'd']
+    assert list(util.expand_list_generator(zip_generator)) == ['1', '3', '2', '4']
+    with open(get_test_filepath('aia_171_level1.fits'), 'rb') as f:
+        with open(get_test_filepath('aia_171_level1.fits'), 'rb') as f2:
+            assert list(util.expand_list_generator([f])) == f2.readlines()
```
nabobalis, line 72: This is what I came up with, I hate it.

Jett-Code: Isn't this a more efficient way of doing the same thing 
```
with open(get_test_filepath('aia_171_level1.fits'), 'rb') as f:
        file_contents = f.readlines()
        assert list(util.expand_list_generator([f])) == file_contents
```

nabobalis: I thought that when you call readlines, it consumes f, so it's now at the end of the file. 

I guess I can just rewind it?

Jett-Code:
Hey, can you help me with failing test cases?

nabobalis:
> Hey, can you help me with failing test cases?

You mean the two failing checks? What would you like help with? 

The gallery one looks like a real failure due to this code. 
Unsure about the online one.

We just need to run those failing examples and debug why they are failing now and see if they are due to the code changes in this pull request.

Jett-Code:
The examples mentioned under unexpected failing examples are working perfectly fine when executed locally, is there any other way of finding the error?

nabobalis:
> Hey, can you help me with failing test cases?

Looks like it was due to me removing UserList, adding it back fixed it (at least the examples)

sunpy/util/tests/test_util.py
```diff
@@ -48,11 +54,41 @@ def test_expand_list():
...
+@pytest.mark.filterwarnings("ignore:Invalid 'BLANK' keyword in header")
+def test_expand_list_generator(aia171_test_map):
+    # We want to ensure that it does not expand:
+    # - strings, bytes, array-like objects, WCS or FITS headers
+    # Only lists, tuples and Generator objects should be expanded
+    header = fits.getheader(get_test_filepath('aia_171_level1.fits'))
+    generator = iter(['c', 'd'])
+    zip_generator = zip(['1', '2'], ['3', '4'])
+    first_list = ['a1234', 'b', [], (['c', 'd']), tuple(), ['e'], b'fghj']
+<<<<<<< HEAD
```
nabobalis, line 66: I assume this was not meant to happen?

nabobalis:
The last set of pushes broke the pull request, can you revert it or fix it please?

sunpy/util/tests/test_util.py
Jett-Code: Yeah, working on merge conflicts

nabobalis: What caused the merge conflicts?

Jett-Code: last commit

Jett-Code:
> The last set of pushes broke the pull request, can you revert it or fix it please?

Is it possible to delete the last commit?

nabobalis:
> > The last set of pushes broke the pull request, can you revert it or fix it please?
> 
> Is it possible to delete the last commit?

Do we not need the changes in that commit?

nabobalis:
The entire commit history is a bit strange right now. Seems to be repeated?

nabobalis:
I pushed from an old branch I had locally. Not sure if it was up to date tho.

Jett-Code:
> I pushed from an old branch I had locally. Not sure if it was up to date tho.

Thanks, let me check

Jett-Code:
I guess everything is working fine now

nabobalis:
> Is there anything else to work on regarding this pr?

No, it needs reviews from the other mainainters before a merge.

sunpy/util/tests/test_util.py
```diff
@@ -48,11 +54,30 @@ def test_expand_list():
     lst = [1, 2, 3, [4, 5, 6], 7, (8, 9), ((10, 11), ((12, 13),))]
     assert util.expand_list(lst) == [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13]
 
-
-def test_expand_list_generator():
-    lst = ['a', 'b', [], (['c', 'd']), tuple(), ['e']]
-    assert list(util.expand_list_generator(lst)) == ['a', 'b', 'c', 'd', 'e']
-
+@pytest.mark.filterwarnings("ignore:Invalid 'BLANK' keyword in header")
+def test_expand_list_generator(aia171_test_map):
```
wtbarnes, line 58: This test could be cleaned up quite a bit using `parametrize`.

wtbarnes (approved):

sunpy/util/tests/test_util.py
nabobalis: Have a look now.

wtbarnes: Beautiful. Thanks!

nabobalis:
Thanks for the PR @Jett-Code, sorry it took so long on our side.

PR #7028, opened 2023-05-25
Fix warnings for doc example build

nabobalis:
Passing this without a unit is now deprecated.


nabobalis:
Another example I need to patch.

dstansby (approved):

Issue #7050, opened 2023-06-07
Add sequence of rst overline/underline syntax to dev guide


alasdairwilson:
#7032 makes rst headings more consistent, we should add the order we use the underlines to the dev guide.

> Also the format of the anchors

Cadair:
Also the format of the anchors

PR #7055, opened 2023-06-08
Form more doc rules [ skip ci ]

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/7050


Cadair (approved):
We should figure out a way to lint all this.

PR #3256, opened 2019-07-01
Add binder support via jupytext

StanczakDominik:
Add binder support via jupytext as per Matrix discussion https://matrix.to/#/!MeRdFpEonLoCwhoHeT:matrix.org/$1561981905806323ldgXZ:matrix.org?via=matrix.org&via=cadair.com&via=maunium.net

this is probably going to take me a few tries, please be patient :)

PR #7062, opened 2023-06-12
Remove binder config

dstansby:
The config is currently not very helpful, because it:

- Launches using the development version of sunpy, not a stable version
- Takes a long time to boot up
- Loads into our examples folder, which contains python files and not jupyter notebook files, so there's no obvious way to run them as a notebook


dstansby:
(or we can just get rid of binder if people don't think it's useful and worth maintaining?)

Cadair:
What's the end user experience of landing on this binder actually like? Do the examples render nicely? I am sceptical of the utility of this.

ayshih:
I forgot this existed.  How is this supposed to work? If I start up a Binder instance – which takes a while – I don't understand what I'm actually able to accomplish with it given that we don't have any Jupyter notebooks in the repository.  Notebooks do get rendered for our online docs, but that's separate from this Binder instance.

It may be moot given the above, but I'll note that switching the install to just `sunpy` means you don't install optional dependencies.  Also, of course, this file is redundant with my (better-maintained) pet conda dev environment yml...

dstansby:
I thought installing sunpy from conda did install all the optional dependencies?

Anyway, since we don't seem to know how this works I think I'll remove the config, and if someone wants to come along with working config and demonstrate how it works then we can add it back later.

ayshih:
Ah, this functionality came in with #3256, and `jupytext` was used to dynamically convert the example scripts to Jupyter notebooks.  Perhaps we just need to add a `jupytext.toml` file to get this working again (see https://jupytext.readthedocs.io/en/latest/formats.html#sphinx-gallery-scripts)?

ayshih:
I don't at all understand how `jupytext` works...  Maybe this functionality was ever only half working?

ayshih (changes requested):
> This needs a changelog entry to "cancel out" the changelog entry from https://github.com/sunpy/sunpy/pull/3256.

dstansby:
👍 I added a changelog entry

nabobalis (approved):

ayshih (approved):
Looks good to me!

PR #7075, opened 2023-06-23
Fix typing errors in sunpy.map

dstansby:
This PR:

- Adds configuration and pre-commit-config for running `mypy`
- Fixes errors in `sunpy.map`.

It would be nice to at least not have any typing errors across the library - whether we want to wholesale add typing is out of scope for this PR.

If people are happy with getting rid of typing errors, I can go through and do this for the rest of the library.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -9,6 +9,7 @@ import numbers
 import textwrap
 import itertools
 import webbrowser
+from typing import Union, Callable
 from tempfile import NamedTemporaryFile
 from collections import namedtuple
 
@@ -62,6 +63,8 @@ from sunpy.visualization.colormaps import cm as sunpy_cm
 TIME_FORMAT = config.get("general", "time_format")
 PixelPair = namedtuple('PixelPair', 'x y')
 SpatialPair = namedtuple('SpatialPair', 'axis1 axis2')
+
+_width_height_type = Union[u.pix, u.deg]
 
 _META_FIX_URL = 'https://docs.sunpy.org/en/stable/code_ref/map.html#fixing-map-metadata'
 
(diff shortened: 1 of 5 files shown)


sunpy/map/sources/soho.py
```diff
@@ -148,6 +148,10 @@ def nickname(self):
         filter = self.meta.get('filter', '')
         return f'{self.instrument}-{self.detector} {filter}'
 
+    @nickname.setter
+    def nickname(self, value):
+        raise RuntimeError("Cannot manually set nickname for LASCOMap")
```
nabobalis, line 153: Do we have this for other sources?

dstansby: This is because GenericMap defines a setter, but LASCOMap doesn't respect that and instead just generates the nickname from the metadata. Perhaps I should put this in a separate PR anyway since it's more of a bugfix.

wtbarnes:
Apologies for my ignorance here, but what constitutes a typing error in this context?

dstansby:
Sorry for lack of context - a typing error here means an error when [`mypy`](https://mypy.readthedocs.io/en/stable/) (the most widely used static type checker) is run on the code. We can gradually make the type checking stricter if desired (e.g. enforcing type hints in a given file/sub-module), but this is the minimum set of changes needed to make `mypy` pass with the default (not very strict) settings.

nabobalis:
I think this now needs rebasing due to the nickname PR merge?

sunpy/map/mapbase.py
```diff
@@ -1776,7 +1779,7 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=3, scale=1.0,
         return new_map
 
     @u.quantity_input
-    def submap(self, bottom_left, *, top_right=None, width: (u.deg, u.pix) = None, height: (u.deg, u.pix) = None):
+    def submap(self, bottom_left, *, top_right=None, width: _width_height_type = None, height: _width_height_type = None):
```
Cadair, line 1782: I feel like this change has had a negative effect on our documentation.

Before:
2023-10-09-144642_grim

After:
2023-10-09-144739_grim

Now I am not saying either is great but it's worse now.

Perhaps we should look into using this: https://pypi.org/project/sphinx-autodoc-typehints/

Cadair (changes requested):
I would like to see some tweaks to our docs to make it a net improvement.

PR #7100, opened 2023-07-14
Add tests for checking whether Map methods are lazy

wtbarnes:
This adds a set of tests that checks whether map methods preserve laziness and that the results are consistent with their eager counterparts. Several are marked as xfail for now because the computations cannot be/are not implemented on the Dask side.

If someone has an idea of how better to use a WCS as an input parameter in parametrize, please let me know.

- [x] Changelog
- [x] Rebase

### DIFF
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -2,6 +2,7 @@
 Test Generic Map
 """
 import re
+import copy
 import tempfile
 import contextlib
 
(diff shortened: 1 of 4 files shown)


dstansby:
Is this ready to be reviewed?

wtbarnes:
> Is this ready to be reviewed?

Yep! Just realized I did in fact add a changelog and rebase but didn't tick the boxes in my above comment.

sunpy/map/tests/test_mapbase.py
```diff
@@ -1653,3 +1654,43 @@ def test_plot_annotate_nonboolean(aia171_test_map):
...
+    ("mean", {}),
+    ("min", {}),
+    pytest.param("reproject_to", {"wcs": aia_wcs}, marks=pytest.mark.xfail(reason="reproject is not dask aware")),
+    pytest.param("resample", {"dimensions": (100, 100)*u.pix}, marks=pytest.mark.xfail()),
+    pytest.param("rotate", {}, marks=pytest.mark.xfail(reason="nanmedian is not implemented in Dask")),
+    ("std", {}),
+    ("superpixel", {"dimensions": (10, 10)*u.pix}),
+    ("submap", {"bottom_left": (100, 100)*u.pixel, "width": 10*u.pixel, "height":10*u.pixel}),
+])
+def test_method_preserves_dask_array(aia171_test_map, aia171_test_dask_map, func, args):
```
dstansby, line 1685:
```suggestion
def test_method_preserves_dask_array(aia171_test_map, aia171_test_dask_map, func, args):
    """
    Check that map methods preserve dask arrays if they are given as input, instead of eagerly
    operating on them and bringing them into memory.
    """
```

dstansby (commented):
Looks great overall 👍  Left two comments/suggestions. My other suggestion is maybe start a new `test_mapbase_dask.py` file with this, as `test_mapbase.py` is getting a bit long?

sunpy/map/tests/test_mapbase.py
```diff
@@ -1653,3 +1654,43 @@ def test_plot_annotate_nonboolean(aia171_test_map):
...
+    dask_array = pytest.importorskip('dask.array')
+    # Check that result array is still a Dask array
+    res_dask = aia171_test_dask_map.__getattribute__(func)(**args)
+    res = aia171_test_map.__getattribute__(func)(**args)
+    result_is_map = isinstance(res, GenericMap)
+    if result_is_map:
+        assert isinstance(res_dask.data, dask_array.Array)
+        assert np.allclose(res_dask.data.compute(), res.data)
```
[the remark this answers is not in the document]
wtbarnes: I made the mistake of moving everything to a separate file before committing these suggestions, but I've made both of these changes. Thanks!

wtbarnes:
> Looks great overall 👍 Left two comments/suggestions. My other suggestion is maybe start a new `test_mapbase_dask.py` file with this, as `test_mapbase.py` is getting a bit long?

Good idea. i've moved all of the Dask-related tests to a separate file now.

dstansby (approved):

sunpy/map/tests/test_mapbase_dask.py
```diff
@@ -0,0 +1,67 @@
...
+import astropy.wcs
+from astropy.io import fits
+from astropy.io.fits.verify import VerifyWarning
+
+import sunpy.map
+from sunpy.data.test import get_test_filepath
+
+
+def test_dask_array(generic_map):
+    dask_array = pytest.importorskip('dask.array')
```
nabobalis, line 16: Can we do this for the entire file instead of in each test?

nabobalis (approved):

sunpy/map/tests/test_mapbase_dask.py
wtbarnes: That I do not know. You can do it with a fixture and then every thing that uses that fixture will be skipped if that fixture is skipped.

wtbarnes: Ok I've cleaned them up so that that `importorskip` is only in one fixture.

nabobalis (approved):

dstansby (approved):

PR #7108, opened 2023-07-20
Add goes-18 as possible value for satno in xrsclient

alasdairwilson:
Currently you can only get XRS data up to 17 and 18 has been in active use for a while.

There is an open question about support for goes 18 data in timeseries due to unresolved issues with the satellite calibration but I don't think that this should stop us supporting it in Fido.

```
result = Fido.search(a.Time(tstart, tend), a.Instrument("XRS"), a.Resolution('flx1s'))

np.unique(result[0]['SatelliteNumber'])

<QueryResponseColumn name='SatelliteNumber' dtype='int32' length=2>
16
18

```

### DIFF
--- a/changelog/7108.trivial.rst
+++ b/changelog/7108.trivial.rst
@@ -1,0 +1,1 @@
+Added missing support to find GOES-18 XRS data in XRSClient.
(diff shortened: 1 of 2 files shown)


changelog/7108.trivial.rst
```diff
@@ -0,0 +1 @@
+Added missing support to find GOES-18 XRS data in XRSClient.
```
nabobalis, line 1:
```suggestion
Added missing support to find GOES-18 XRS data in `~sunpy.net.dataretriever.XRSClient`.
```

nabobalis (approved):

Cadair (approved):

PR #7113, opened 2023-07-26
Internal functions for coordinate transformations are now private API

ayshih:
The various internal functions for coordinate transformations (e.g., `hgs_to_hgc()`) were never intended to be used directly by users.  This PR makes the file private API by renaming `transformations.py` to `_transformations.py`.  There are two user-facing context managers in that file (`transform_with_sun_center()` and `propagate_with_solar_surface()`), but those have always been imported one level up and documented as part of the API there.

Is deprecation needed here?


Cadair:
> Is deprecation needed here?

yes, I think if we are tacitly admitting that it *was* public API then it needs to be deprecated.

Cadair (approved):

nabobalis:
pytest fails with:

```
ERROR ../../.tox/py311/lib/python3.11/site-packages/sunpy/coordinates/transformations.py
ERROR ../../.tox/py311/lib/python3.11/site-packages/sunpy/coordinates/transformations.py
```
The deprecation warning is failing the CI, I think we need to add it to the pytest ignore list in setup.cfg

dstansby (dismissed):
Requesting changes so this isn't merged with broken CI

sunpy/coordinates/transformations.py
```diff
@@ -1,1210 +1,9 @@
...
-          'color: #848484;padding: 0;margin: 0;">\n'\
-          '                <b>' + label + ':</b>\n'\
-          '                    <span class="dot" style="height: 20px;width: 40px;'\
-          'background-color: ' + color + ';border-radius: 50%;border: 1px solid black;'\
-          'display: inline-block;"></span>\n'\
-          '            </p>\n'\
-          '        </li>\n\n\n'
-    return row
+warn_deprecated("The `sunpy.coordinates.transformations` module is deprecated, as it was designed "
+                "for internal use.")
```
dstansby, line 9:
```suggestion
warn_deprecated("The `sunpy.coordinates.transformations` module is deprecated, as it was designed "
                "for internal use. "
                "transform_with_solar_centre() and propagate_with_solar_surface() can be imported from sunpy.coordinates")
```

dstansby (changes requested):
I think this needs a second changelog entry to mention the user facing functions in `sunpy.coordinates.transformations` that have been moved. Even if they are in the `sunpy.coordinates` namespace there could be someone who was importing them from `...transformations`.

wtbarnes:
I think the docs build is still failing…I gave up trying to fix this last week because I have no idea what the problem is and can't reproduce locally.

sunpy/coordinates/transformations.py
ayshih: Done with slightly different text

dstansby (approved):

PR #7118, opened 2023-07-27
Add a Helioprojective method to return coordinate visibility

### LINKED ISSUE #7117: Function to check whether coordinate on disk is visible from another observer
### Describe the feature

it would be very helpful to have a helper to check whether a coordinate from one observer is visible from another. For example, say I have the coordinate of a flare from Solar Orbiter viewpoint and I want to check if its visible from Earth. 

At the moment, it can be misleading to take that coordinate and transform it using `.transform_to(frames.Helioprojective(observer="earth", obstime='xxx')` as it will give you a Tx, Ty value that could be on disk (but larger distance away) (see figure - where the coordinate as seen from solar orbiter, shouldnt be visible from earth, but when you transform that coordinate and plot it, it looks like you are "seeing" through the Sun)

example_issue

### Proposed solution

A function like `is_coordinate_visible(observer, coordinate)` would be helpful, such that you pass it a coordinate from one observer and it will check if it is visible from another passed, and return `True` or  `False`.

Or indeed, maybe a helper such that it will only make the transform if it is visible?
For larger catalogues of solar orbiter events coming online, this is going to catch people up for sure by using the typical 'transform_to' ways etc

Any ideas/suggestions?

ayshih:
This PR adds an `is_visible()` method to `Helioprojective` to return whether the coordinate is visible (i.e., not obscured from the observer assuming that the Sun is an opaque sphere).  The radius of the Sun that is used is the `rsun` frame attribute on the `Helioprojective` coordinate.

Closes #7117

### DIFF
--- a/sunpy/coordinates/frames.py
+++ b/sunpy/coordinates/frames.py
@@ -609,6 +609,85 @@ class Helioprojective(SunPyBaseCoordinateFrame):
                                                           lat=lat,
                                                           distance=d))
 
+    @u.quantity_input
+    def is_visible(self, *, tolerance: u.m = 1*u.m):
+        """
+        Returns whether the coordinate is on the visible side of the Sun.
+
+        A coordinate is visible if it can been seen from the observer (the ``observer``
+        frame attribute) assuming that the Sun is an opaque sphere with a fixed radius
+        (the ``rsun`` frame attribute).  The visible side of the Sun is always smaller
+        than a full hemisphere.
+
+        Parameters
+        ----------
+        tolerance : `~astropy.units.Quantity`
+            The depth below the surface of the Sun that should be treated as
+            transparent.
+
+        Notes
+        -----
+        If the coordinate is 2D, it is automatically deemed visible.  A 2D coordinate
+        describes a look direction from the observer, who would simply see whatever is
+        in "front", and thus cannot correspond to a point hidden from the observer.
+
+        The ``tolerance`` parameter accommodates situations where the limitations of
+        numerical precision would falsely conclude that a coordinate is not visible.
+        For example, a coordinate that is expressly created to be on the solar surface
+        may be calculated to be slightly below the surface, and hence not visible if
+        there is no tolerance.  However, a consequence of the ``tolerance`` parameter
+        is that a coordinate that is formally on the far side of the Sun but is
+        extremely close to the solar limb can be evaluated as visible.  With the
+        default ``tolerance`` value of 1 meter, a coordinate on the surface of the Sun
+        can be up to 11 arcseconds of heliographic longitude past the solar limb and
+        still be evaluated as visible.
+
+        Examples
+        --------
+        >>> import numpy as np
+        >>> import astropy.units as u
+        >>> from astropy.coordinates import SkyCoord
+        >>> from sunpy.coordinates import HeliographicStonyhurst, Helioprojective
+
+        >>> hpc_frame = Helioprojective(observer='earth', obstime='2023-08-03')
+
+        >>> in_front = SkyCoord(0*u.arcsec, 0*u.arcsec, 0.5*u.AU, frame=hpc_frame)
+        >>> print(in_front.is_visible())
+        True
+
+        >>> behind = SkyCoord(0*u.arcsec, 0*u.arcsec, 1.5*u.AU, frame=hpc_frame)
+        >>> print(behind.is_visible())
+        False
+
+        >>> hgs_array = SkyCoord(np.arange(-180, 180, 60)*u.deg, [0]*6*u.deg,
+        ...                      frame='heliographic_stonyhurst', obstime='2023-08-03')
+        >>> print(hgs_array)
+        <SkyCoord (HeliographicStonyhurst: obstime=2023-08-03T00:00:00.000, rsun=695700.0 km): (lon, lat) in deg
+            [(-180., 0.), (-120., 0.), ( -60., 0.), (   0., 0.), (  60., 0.),
+             ( 120., 0.)]>
+        >>> print(hgs_array.transform_to(hpc_frame).is_visible())
+        [False False  True  True  True False]
+        """
+        # If the coordinate is 2D, it must be visible
+        if self._is_2d:
+            return np.ones_like(self.data, dtype=bool)
+
+        # Use a slightly smaller solar radius to accommodate numerical precision
+        solar_radius = self.rsun - tolerance
+
+        data = self.cartesian
+        data_to_sun = self.observer.radius * CartesianRepresentation(1, 0, 0) - data
+
+        # When representing the helioprojective point as true Cartesian, the X value is the
+        # distance from the observer to the point in the sunward direction
+        is_behind_observer = data.x < 0
+        # When comparing heliocentric angles, we compare the sine values and hence avoid calling arcsin()
+        is_beyond_limb = np.sqrt(data.y **2 + data.z **2) / data.norm() > solar_radius / self.observer.radius
+        is_above_surface = data_to_sun.norm() >= solar_radius
+        is_on_near_side = data.dot(data_to_sun) >= 0
+
+        return is_behind_observer | is_beyond_limb | (is_on_near_side & is_above_surface)
+
     _spherical_screen = None
 
     @classmethod
--- a/changelog/7118.feature.rst
+++ b/changelog/7118.feature.rst
@@ -1,0 +1,1 @@
+Added a :meth:`sunpy.coordinates.Helioprojective.is_visible` method to return whether the coordinate is visible (i.e., not obscured from the observer assuming that the Sun is an opaque sphere).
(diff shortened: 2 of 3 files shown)


sunpy/coordinates/frames.py
```diff
@@ -605,6 +605,24 @@ def make_3d(self):
                                                           lat=lat,
                                                           distance=d))
 
+    def is_visible(self):
+        """
+        Returns whether the coordinate is on the visible side of the Sun.
```
Cadair, line 610: Side is probably not the best phrasing given the nuance with the distance we discussed earlier?

Cadair: Also might be worth adding an example of realise frame to change rsun in this docstring?

hayesla: yeah, it would be nice to explain what this could mean (i.e. that it could be on the other side), and checks that its visible from the observer in the HPC frame.

hayesla (commented):
thanks for the quick PR

I had a quick play around with it, but seems to have an issue with the ` is_above_surface = data_to_sun.norm() >=  self.rsun` 

for example if you do:
```python
>>> earth_flare = SkyCoord(0*u.arcsec, 0*u.arcsec, frame=frames.Helioprojective(observer=earth_observer))

>>> earth_flare.is_visible()
array(True)


>>> earth_flare.make_3d().is_visible()
False
```

seems to be that if you look at the difference between `data_to_sun.norm()` and `self.rsun` can be really small even though they should be the same?

```python

>>> test_coord = earth_flare.make_3d()
>>> data = test_coord.cartesian
>>> data_to_sun = test_coord.observer.radius * CartesianRepresentation(1, 0, 0) - data
>>> (data_to_sun.norm() - test_coord.rsun)
−3.522356×10−15AU
```

ayshih:
Yeah, numerical noise is a pain for the inequality checks.  I'll likely add some sort of tolerance parameter so that points ever so slightly below the surface are still considered visible.

hayesla:
yeah like maybe something like a `u.isclose()` check or something?

Cadair:
> In your example above, that's 0.5 millimeters below the surface.

it said *below*.

sunpy/coordinates/frames.py
ayshih: > Also might be worth adding an example of realise frame to change rsun in this docstring?

Definitely not

Cadair: or at least linking out to it, because people wont know it exists.

wtbarnes (approved):

sunpy/coordinates/frames.py
```diff
@@ -605,6 +605,70 @@ def make_3d(self):
...
+        """
+        # If the coordinate is 2D, it must be visible
+        if self._is_2d:
+            return np.ones_like(self.data, dtype=bool)
+
+        data = self.cartesian
+        data_to_sun = self.observer.radius * CartesianRepresentation(1, 0, 0) - data
+
+        is_behind = data.x < 0
+        is_beyond_limb = np.sqrt(1 - (data.x / data.norm()) ** 2) > self.rsun / self.observer.radius
```
dstansby, line 666:
```suggestion
        is_beyond_limb = np.sqrt(self.Tx**2 + self.Ty**2) > np.arcsin(self.rsun / self.observer.radius)
```
I got a bit confused with the logic, and I think my suggestion is a bit clearer (making it epxlicit the angle of the coordinate is being compared to the angle of the solar limb)?

dstansby (changes requested):
I'm slightly confused as to why this seems to return `False` for any coordinate in the half plane opposite to the observer-Sun vector (if I'm interpreting the logic correctly?). What's the motiviation for this? I'd be minded to say a coordinate "behind" the observer is still visible if the observer turns around (or has a > 180 deg FOV).

sunpy/coordinates/frames.py
ayshih: Your suggestion is formally inaccurate because it relies on the small-angle approximation to combine Tx and Ty like that to get the radial angle.

sunpy/coordinates/frames.py
```diff
@@ -605,6 +605,70 @@ def make_3d(self):
...
+             ( 120., 0.)]>
+        >>> print(hgs_array.transform_to(hpc_frame).is_visible())
+        [False False  True  True  True False]
+        """
+        # If the coordinate is 2D, it must be visible
+        if self._is_2d:
+            return np.ones_like(self.data, dtype=bool)
+
+        data = self.cartesian
+        data_to_sun = self.observer.radius * CartesianRepresentation(1, 0, 0) - data
```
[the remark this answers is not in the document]
ayshih: Your suggested comment has it backwards.  `data_to_sun` is the vector from the coordinate to the center of the Sun.

sunpy/coordinates/frames.py
```diff
@@ -605,6 +605,70 @@ def make_3d(self):
...
+        A coordinate is visible if it can been seen from the observer (the ``observer``
+        frame attribute) assuming that the Sun is an opaque sphere with a fixed radius
+        (the ``rsun`` frame attribute).  The visible side of the Sun is always smaller
+        than a full hemisphere.
+
+        Parameters
+        ----------
+        tolerance : `~astropy.units.Quantity`
+            The depth below the surface of the Sun that should be treated as
+            transparent.
```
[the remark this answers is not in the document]
ayshih: Hmm, fair point, and doing either is actually a bit nontrivial.  That said, the default tolerance of 1 m is so small that the difference will be academic in nearly all normal usage.

ayshih:
> I'm slightly confused as to why this seems to return `False` for any coordinate in the half plane opposite to the observer-Sun vector (if I'm interpreting the logic correctly?). What's the motiviation for this? I'd be minded to say a coordinate "behind" the observer is still visible if the observer turns around (or has a > 180 deg FOV).

I think you misread.  If `is_behind` (or `is_behind_observer`) is True, the return of `is_visible()` will be True.

sunpy/coordinates/frames.py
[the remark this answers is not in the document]
dstansby: 👍

sunpy/coordinates/frames.py
dstansby: Could you add a comment on what `data.x` is instead then? I think it would be good to make it easier for someone else to understand (and check) the maths going on here.

sunpy/coordinates/frames.py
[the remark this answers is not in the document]
dstansby: Couldn't you just use `actual_solar_radius - tolerance` as the radius in all the calculations?

dstansby:
> > I'm slightly confused as to why this seems to return `False` for any coordinate in the half plane opposite to the observer-Sun vector (if I'm interpreting the logic correctly?). What's the motiviation for this? I'd be minded to say a coordinate "behind" the observer is still visible if the observer turns around (or has a > 180 deg FOV).
> 
> I think you misread. If `is_behind` (or `is_behind_observer`) is True, the return of `is_visible()` will be True.

Woops, so I did 🙈 . I think renaming `is_behind` to `is_behind_observer` is worth doing anyway.

sunpy/coordinates/frames.py
```diff
@@ -605,6 +605,70 @@ def make_3d(self):
...
+        [False False  True  True  True False]
+        """
+        # If the coordinate is 2D, it must be visible
+        if self._is_2d:
+            return np.ones_like(self.data, dtype=bool)
+
+        data = self.cartesian
+        data_to_sun = self.observer.radius * CartesianRepresentation(1, 0, 0) - data
+
+        is_behind = data.x < 0
```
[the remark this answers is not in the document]
ayshih: Renamed

sunpy/coordinates/frames.py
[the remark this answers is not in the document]
ayshih: I've made this change, and tried to explain the implication a gory note.

sunpy/coordinates/frames.py
ayshih: I added a clarification for `data.x`, and I also changed the expression to `np.sqrt(data.y **2 + data.z **2) / data.norm()` since that's likely easier for future readers.

sunpy/coordinates/frames.py
ayshih: I'm not a fan of adding a reference to `realize_frame()`.

wtbarnes: "side" is further clarified lower down in the docstring so I think the wording here is ok.

wtbarnes (approved):

dstansby (approved):
Modulo Will's comment above, 👍 from me

sunpy/coordinates/frames.py
```diff
@@ -605,6 +605,83 @@ def make_3d(self):
...
+
+        Parameters
+        ----------
+        tolerance : `~astropy.units.Quantity`
+            The depth below the surface of the Sun that should be treated as
+            transparent.
+
+        Notes
+        -----
+        If the coordinate is 2D, it is immediately assumed to be visible.
```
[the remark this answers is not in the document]
ayshih: Okay, I blathered on a bit

PR #7123, opened 2023-07-28
Added an example to make an anaglyph 3D stereoscopic image

ayshih:
Here's a gallery example to make an anaglyph 3D (i.e., red-cyan) image from a stereoscopic observation.  The data is from EUVI and AIA in July 2023, when STEREO-A and SDO were only a few degrees separated.

### DIFF
--- a/changelog/7123.doc.rst
+++ b/changelog/7123.doc.rst
@@ -1,0 +1,1 @@
+Added an example (:ref:`sphx_glr_generated_gallery_showcase_stereoscopic_3d.py`) for how to make an anaglyph 3D (i.e., red-cyan) image from a stereoscopic observation.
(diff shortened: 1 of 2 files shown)


nabobalis:
Any reason to not backport this?

ayshih:
I don't object to backporting

dstansby (approved):
Nice! My only (optional) suggestion would be to make the sphinx gallery thumnail the second image.

ayshih:
> Nice! My only (optional) suggestion would be to make the sphinx gallery thumnail the second image.

I considered it, but because it's a color anaglyph image (as opposed to grayscale), it kinda just looks like a badly processed AIA image at thumbnail size.

wtbarnes (approved):

PR #7130, opened 2023-08-07
Add nice 404 page

nabobalis:
IDea stolen from plasmapy with love.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -119,6 +119,7 @@ extensions = [
     'sphinx_design',
     'sphinx_copybutton',
     'hoverxref.extension',
+    'notfound.extension',
 ]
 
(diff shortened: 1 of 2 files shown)


Cadair (commented):
:+1: this change is a nice improvement as-is.

My main comment would be that I don't think this is the place to do it. Ideally we would want this to work over all pages on sunpy.org, so it should live in the theme. I think we don't have a good way of injecting extensions in the theme's config at the moment though right?

nabobalis:
Yeah we used to do that and maybe we should bring it all back when we refactor the theme. 

We have to currently setup a lot of extensions and if be nice for the theme to do it

Cadair:
open an issue on the theme repo so we don't forget?

PR #7132, opened 2023-08-08
Experiment with building a htmlzip of the docs

Cadair:
I am interested in the build overhead of this, but I think having a downloadable artefact of the whole built docs would be useful generally, and specifically as I would like to use it to build Dash/Zeal docsets from.


Cadair:
Hmm, this doesn't seem to have had any effect on the PR build, but as far as I can tell PRs only build HTML so I think we need to merge this.

nabobalis (approved):
I mean, sure why not. What could go wrong.

Cadair:
@MeeseeksDev backport to 5.0

PR #7134, opened 2023-08-08
Typo correction to the frameless example

nabobalis:
Basically a typo

### DIFF
--- a/examples/map/plot_frameless_image.py
+++ b/examples/map/plot_frameless_image.py
@@ -27,7 +27,7 @@ ax = plt.axes([0, 0, 1, 1])
 ax.set_axis_off()
 
 # Plot the map.
-# Since we are not interested in the exact map coordinates, 
+# Since we are not interested in the exact map coordinates,
 # we can simply use :meth:`~matplotlib.Axes.imshow`.
 norm = smap.plot_settings['norm']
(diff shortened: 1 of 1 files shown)


examples/map/plot_frameless_image.py
```diff
@@ -38,6 +39,7 @@
 # sphinx_gallery_defer_figures
 
 ##############################################################################
-# At this point you could save the figure with ``plt.savefig()`` or show it:
+# At this point you could save the figure with :meth:`~matplotlib.pyplot.savefig`
+# or show it:
```
Cadair, line 43: Why? this should link to savefig no?

nabobalis: DOes it not?

Cadair: it should have done before and wont now? I don't understand why you changed this?

ayshih: I feel like @Cadair is misreading...  It certainly wasn't linking before, and so this fix is in the right direction, but `:meth:` needs to be `:func:`

alasdairwilson (approved):

examples/map/plot_frameless_image.py
nabobalis: Corrected.

PR #7137, opened 2023-08-09
Remove pytest skip in files as it was breaking downstream

nabobalis:
I had this problem in the doc build for sunkit image

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -224,6 +224,8 @@ filterwarnings =
     # https://github.com/h5py/h5py/issues/2243 for upstream issues about this warning being
     # emitted
     ignore:`product` is deprecated as of NumPy 1.25.0:DeprecationWarning
+    # sunpy IO skip here to prevent pytest breaking items downstream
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -224,6 +224,8 @@ filterwarnings =
     # https://github.com/h5py/h5py/issues/2243 for upstream issues about this warning being
     # emitted
     ignore:`product` is deprecated as of NumPy 1.25.0:DeprecationWarning
+    # sunpy IO skip here to prevent pytest breaking items downstream
+    ignore:.*module is deprecated, as it was designed for internal use
```
Cadair, line 228: Add the sunpy deprecation class here to make it clear it's ours?

Cadair (approved):

setup.cfg
nabobalis: sunpy.util.exceptions.SunPyDeprecationWarning should work?

nabobalis:
That figure test fail is a problem

PR #7141, opened 2023-08-10
Fix sunpy.io to update header info after scaling applied when data is accessed

samaloney:
Initial quick fix `_fits.get_header` and `_fis.read_fits` could be refactored to make this a but nicer will do later.


sunpy/io/_fits.py
```diff
@@ -79,11 +79,12 @@ def read(filepath, hdus=None, memmap=None, **kwargs):
         for h in hdulist:
             h.verify('silentfix+warn')
 
-        headers = get_header(hdulist)
         pairs = []
 
-        for i, (hdu, header) in enumerate(zip(hdulist, headers)):
+        for i, hdu in enumerate(hdulist):
             try:
+                hdu.data  # access data to force any scaling BSCALE/BZERO to be applied and keyword will be updated
```
Cadair, line 86: I worry about the performance impact and general side-effects of doing this here. :see_no_evil:

samaloney: Yea I was too but I don't _think_ it's any different to this line which was being called `HDPair(hdu.data, header)` it has the same side effect as far as I can see, `hdu.data` is being accessed and causes astropy.io to load and scale the data, I'll try to confirm either way. Long term if you want to support lazy loading and dask etc this is an issue for sure.

dstansby (changes requested):
I'm going to request changes since I'm not quite sure what this is fixing. It should have a test anyway, and I think this would help review the PR.

nabobalis:
I pushed a unit test thinking I understood what the problem is but turns out I did not.

The test passed before and after the code change. 

@samaloney, what was the original bug?

nabobalis:
I wasn't able to replicate the original issue, so I will close this for now.

Issue #7023, opened 2023-05-24
Customize reference pixel longitude in `sunpy.map.header_helper.make_heliographic_header`


STBadman:
### Describe the feature

`sunpy.map.header_helper.make_heliographic_header` has a hardcoded reference pixel coordinate which assumes the map center is at 0 degrees. This makes sense for the stonyhurst reference frame but when creating maps in the carrington frame, 0 is often at the left hand edge of the map or at an arbitrary roll. 

It is very straightforward to simply introduce a keyword to the function that allows a user to modify the reference pixel longitude from 0. 

### Proposed solution

Add a keyword `map_center_longitude` with an `astropy.units.quantity_input` decorator to `sunpy.map.header_helper.make_heliographic_header`  and pass the keyword to the creation of the reference pixel coordinate inside the function.

See :  https://github.com/sunpy/sunpy/compare/main...STBadman:sunpy:heliographic_headers and similar to [`pfsspy.utils.carr_cea_wcs_header` ](https://github.com/dstansby/pfsspy/blob/main/pfsspy/utils.py)

Please let me know if this seems reasonable and I can request a PR from my fork. Thank you!

ayshih:
I agree that it's less common for Carrington maps to have zero longitude at map center.  I'd be in favor of changing the stock behavior of `make_heliographic_header()` so that Carrington maps go from 0 to 360 degrees in longitude, but there are probably people that would dislike such a change.

I'm not keen on adding a keyword option to `make_heliographic_header()`.  (Others are welcome to voice contrary opinions!)  That convenience function is a stripped-down version of `make_fitswcs_header()`, so adding a keyword option is fighting against the stripping down.  If someone needs an unusual WCS, I'm inclined to just point them to `make_fitswcs_header()`.

Note that it is straightforward to work around the current behavior by simply modifying `CRVAL1`/`CRPIX1` in the returned header.

schonfsj:
I am strongly in support of @STBadman's original suggestion that this should be a user-defined keyword (although I'm not opposed to also changing the default behavior).

[...]

PR #7143, opened 2023-08-14
Create heliographic map header with non-zero longitude for map center

STBadman:
## Allow user to create heliographic map header with custom longitude roll

This PR adds a keyword "map_center_longitude" to `sunpy.map.header_helper.make_heliographic_header` to allow the user to specify the longitude of crpix1 when creating a Carrington map header. This addresses issue [7023](https://github.com/sunpy/sunpy/issues/7023).  The keyword is required to be an astropy.Quantity with units equivalent to radians. User input values satisfying this are automatically wrapped to the range [0,360] deg in the output header. 

 As discussed in [7023](https://github.com/sunpy/sunpy/issues/7023), this extends the utility of `sunpy.map.header_helper.make_heliographic_header` to allow user created maps at an arbirtrary roll angle, rather than just the default (unchanged) which sets the map center to be 0 degrees longitude (as is the case for the Stonyhurst frame).

The PR includes adding the keyword to `sunpy.map.header_helper.make_heliographic_header` , including the astropy units decorator, and includes tests in `sunpy.map.tests.test.test_header_helper.py` that ensure the units input is correctly enforced, and that the keyword supplied by the user correctly maps and wraps to the expected header value of 'crpix1'


changelog/7143.feature.rst
```diff
@@ -0,0 +1 @@
+Add map_center_longitude keyword to `sunpy.map.header_helper.make_heliographic_header`
```
ayshih, line 1:
```suggestion
Add a keyword (``map_center_longitude``) to :func:`~sunpy.map.header_helper.make_heliographic_header` for centering the heliographic map at a longitude other than zero longitude.
```

sunpy/map/header_helper.py
```diff
@@ -384,8 +384,8 @@ def _set_instrument_meta(meta_wcs, instrument, telescope, observatory, detector,
     'Matrix element CDi_j describing the rotation required to align solar North with the top of the image.'
 }
 
-
-def make_heliographic_header(date, observer_coordinate, shape, *, frame, projection_code="CAR"):
+@u.quantity_input
+def make_heliographic_header(date, observer_coordinate, shape, *, frame, projection_code="CAR", map_center_longitude=0.0*u.deg):
```
ayshih, line 388:
```suggestion
def make_heliographic_header(date, observer_coordinate, shape, *, frame, projection_code="CAR",
                             map_center_longitude: u.Quantity['angle'] = 0.0*u.deg):
```
You specified the argument to have a `Quantity` as the default value, but that doesn't actually specify that the argument has to be a `Quantity` or what units are allowed.  You need to specify what unit(s) are permitted so that the input can be validated.

The `UnitsError`s that you were getting were from subsequent code trying to use the provided value, rather than from validation of the inputs to this function.

sunpy/map/header_helper.py
```diff
@@ -405,6 +405,8 @@ def make_heliographic_header(date, observer_coordinate, shape, *, frame, project
         Coordinate frame.
     projection_code : {'CAR', 'CEA'}
         Projection to use for the latitude coordinate.
+    map_center_longitude : `astropy.units.Quantity` (equivalent units to `u.deg`)
+        Heliographic longitude of the center pixel
```
ayshih, line 409:
```suggestion
    map_center_longitude : `~astropy.units.Quantity`
        Heliographic longitude of the map center
```
There is a "center pixel" only if the map has an odd number of rows and columns.

sunpy/map/tests/test_header_helper.py
```diff
@@ -272,3 +272,18 @@ def test_make_heliographic_header_invalid_inputs(aia171_test_map):
 
     with pytest.raises(ValueError, match='frame must be one of'):
         make_heliographic_header(aia171_test_map.date, aia171_test_map.observer_coordinate, [90, 180], frame='blah')
+
+    with pytest.raises(u.UnitsError):
+        make_heliographic_header(aia171_test_map.date, aia171_test_map.observer_coordinate, [90, 180], frame='carrington', map_center_longitude=0)
```
ayshih, line 277:
```suggestion
    with pytest.raises(u.TypeError):
        make_heliographic_header(aia171_test_map.date, aia171_test_map.observer_coordinate, [90, 180], frame='carrington', map_center_longitude=0)

    with pytest.raises(u.UnitsError):
        make_heliographic_header(aia171_test_map.date, aia171_test_map.observer_coordinate, [90, 180], frame='carrington', map_center_longitude=0*.m)
```
With the fix above to correctly use the `quantity_input` decorator, passing something that is not a `Quantity` raises a `TypeError`.

ayshih (changes requested):
Thanks for the PR!  There's some stuff that needs to be fixed

sunpy/map/header_helper.py
STBadman: Got it, thanks

STBadman:
Thank you so much for @ayshih for writing the changes so explicitly.

sunpy/map/header_helper.py
```diff
@@ -384,8 +384,9 @@ def _set_instrument_meta(meta_wcs, instrument, telescope, observatory, detector,
     'Matrix element CDi_j describing the rotation required to align solar North with the top of the image.'
 }
 
-
-def make_heliographic_header(date, observer_coordinate, shape, *, frame, projection_code="CAR"):
+@u.quantity_input
+def make_heliographic_header(date, observer_coordinate, shape, *, frame, projection_code="CAR",
+                             map_center_longitude: u.Quantity['angle'] = 0.0*u.deg):
```
ayshih, line 389:
```suggestion
                             map_center_longitude: u.Quantity[u.deg] = 0.0*u.deg):
```
Specifying a string makes `ruff` unhappy, so let's just specify a specific angle unit instead.  It doesn't affect the behavior at all, just the error message.

ayshih (approved):
Looks good to me!  We need another approval before merging

To whomever clicks the button: squash upon merge

wtbarnes (approved):

PR #7159, opened 2023-08-17
Remove the hard dep on parfive

### LINKED ISSUE #7150: sunpy map has a hard dep on parfive
### Describe the bug

I am trying to use sunpy inside pyiodide and because of the lack of aiohttp https://github.com/pyodide/pyodide/issues/4070 I noticed that we trigger an import of parfive through map_factory. This is unexpected and probably we should think about reworking how `sunpy.data` import

### To Reproduce

_No response_

### Screenshots

_No response_

### System Details

_No response_

### Installation method

_No response_

Cadair:
We deprecated it already, this should practically fix #7150 but I think we need to add some testing infra before we close that.


sunpy/data/data_manager/downloader.py
```diff
@@ -42,6 +40,8 @@ class ParfiveDownloader(DownloaderBase):
     """
 
     def download(self, url, path):
+        from sunpy.util.parfive_helpers import Downloader
```
Cadair, line 43: This is basically a hack. We need to be able to import `sunpy.data` for our tests, but the user API requires parfive. I am open to better suggestions.

tox.ini
```diff
@@ -117,6 +117,15 @@ extras =
 deps =
     astropy
     numpy
-    parfive[ftp]
+    packaging
 commands =
     python -c "import sunpy; sunpy.self_test()"
+    # Test that we can import subpackages which don't have specific extras
+    # TODO: Actually run the tests for these modules
```
Cadair, line 124: In theory (hahaa) we should be able to run the tests for these subpackages with the test and core deps installed.

nabobalis: How does a sunpy test function to do these sound instead of adding lines to tox?

dstansby (changes requested):
Requesting that CI is fixed 😉

nabobalis:
That base deps failed but the check passed. Will need to fix that.

setup.cfg
```diff
@@ -43,7 +43,6 @@ install_requires =
   astropy>=5.0.1
   numpy>=1.21.0
   packaging>=19.0
-  parfive[ftp]>=2.0.0
```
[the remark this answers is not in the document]
alasdairwilson: It has been moved to the net extras but this does further limit what functionality you get from a simple `install sunpy`

setup.cfg
```diff
@@ -54,12 +52,13 @@ asdf =
...
   dask[array]>=2021.4.0
+data =
+  parfive[ftp]>=2.0.0
 image =
   scikit-image>=0.18.0
   # !=1.10.0 due to https://github.com/scipy/scipy/issues/17718
   scipy>=1.7.0,!=1.10.0
 jpeg2000 =
-  glymur>=0.9.1,!=0.9.5
   lxml>=4.8.0
```
nabobalis, line 62: I wonder if we need lxml for this.

nabobalis:
This has become a conflicted mess. 

I will close this and we should start again.

PR #7160, opened 2023-08-18
Reproject AIA instead of HMI in masking example

wtbarnes:
In the ["Masking HMI"](https://docs.sunpy.org/en/stable/generated/gallery/map/masking_hmi.html#sphx-glr-generated-gallery-map-masking-hmi-py) example gallery entry, the HMI data is being reprojected to the AIA frame in order to mask the pixels in the pixel coordinate system of AIA. Because of the range of LOS magnetogram data, reprojection of these data should be avoided. This PR modifies the example to reproject the AIA image instead and work in the HMI pixel frame. It also adds a plot of the distribution of LOS HMI values at the end.


examples/map/masking_hmi.py
```diff
@@ -56,34 +61,43 @@
...
+# We can then plot the largest HMI region.
 
 fig = plt.figure()
 ax = fig.add_subplot(projection=hmi_masked)
 im = hmi_masked.plot(axes=ax, cmap="hmimag", norm=Normalize(-1500, 1500))
 fig.colorbar(im)
 
+###############################################################################
+# Finally, we can plot the distribution of HMI LOS flux for only the unmasked values in the largest region shown above.
+fig = plt.figure()
```
nabobalis, line 97:
```suggestion

fig = plt.figure()
```

nabobalis (approved):

wtbarnes:
I still need a changelog here. Stuart suggested adding an explanation of *why* we avoid reprojecting HMI and reproject AI instead. I think is more appropriate as a topic guide.

dstansby (changes requested):
Requesting changes since this still needs some form of changelog

dstansby (approved):

examples/map/masking_hmi.py
```diff
@@ -56,34 +61,44 @@
...
-# Finally, plot the largest HMI region.
+# We can then plot the largest HMI region.
 
 fig = plt.figure()
 ax = fig.add_subplot(projection=hmi_masked)
 im = hmi_masked.plot(axes=ax, cmap="hmimag", norm=Normalize(-1500, 1500))
 fig.colorbar(im)
 
+###############################################################################
+# Finally, we can plot the distribution of HMI LOS flux for only the unmasked values in the largest region shown above.
```
dstansby, line 96: I may be wrong, but isn't the  data is in units of magnetic field strength (Gauss) and not flux (Webers or similar)?

examples/map/masking_hmi.py
```diff
@@ -56,34 +61,44 @@
...
+
+fig = plt.figure()
+ax = fig.add_subplot()
+ax.hist(hmi_masked.data[~hmi_masked.mask], bins='auto', histtype='step')
+ax.set_ylabel('Number of Pixels')
+ax.set_xlabel(f'LOS Flux [{hmi.unit.to_string(format="latex_inline")}]')
```
dstansby, line 102: Same comment re. unit above.

examples/map/masking_hmi.py
wtbarnes: Good point! I've changed it to "magnetic field".

examples/map/masking_hmi.py
```diff
@@ -30,13 +30,18 @@
...
 
 hmi = sunpy.map.Map(HMI_LOS_IMAGE)
-hmi = hmi.reproject_to(aia.wcs)
-hmi.nickname = 'HMI magnetogram'
+hmi = hmi.submap(aia.bottom_left_coord, top_right=aia.top_right_coord)
+
+###############################################################################
+# We then call the :meth:`~sunpy.map.GenericMap.reproject_to` to reproject the AIA Map
+# to have exactly the same grid as the HMI Map.
+# :ref:`sphx_glr_generated_gallery_map_transformations_reprojection_align_aia_hmi.py` provides more reference.
```
Cadair, line 41: I maintain that this choice is really subtle and important and not calling it out here at all is doing the reader of this example a disservice.

examples/map/masking_hmi.py
```diff
@@ -56,34 +61,44 @@
...
 fig.colorbar(im)
 
+###############################################################################
+# Finally, we can plot the distribution of HMI LOS magnetic field for only the unmasked values in the largest region shown above.
+
+fig = plt.figure()
+ax = fig.add_subplot()
+ax.hist(hmi_masked.data[~hmi_masked.mask], bins='auto', histtype='step')
+ax.set_ylabel('Number of Pixels')
+ax.set_xlabel(f'LOS Magnetic Field [{hmi.unit.to_string(format="latex_inline")}]')
```
Cadair, line 102:
```suggestion
ax.set_xlabel(f'LOS Magnetic Field [{hmi.unit:latex_inline}]')
```

Cadair (changes requested):
I feel like we need the explanation page to go along with this PR.

If we don't want to add explanatory comments inline any more then I am sorry but we have to do the work to add explanation pages in lieu of doing so.

wtbarnes:
> I feel like we need the explanation page to go along with this PR.
> 
> If we don't want to add explanatory comments inline any more then I am sorry but we have to do the work to add explanation pages in lieu of doing so.

I added two sentences to try and address this. I think there is more value in adding the information directly in this example than splitting it off into a separate topic guide, at least with the amount of information I've added here. If there is a more formal explanation we want to add, I think this could go in a topic guide and we could link to it.

Cadair (approved):

PR #7163, opened 2023-08-23
Catch when the `annotate` keyword for `GenericMap.plot()` is not boolean

ayshih:
Twice in the past month, we've had a user call `GenericMap.plot()` assuming that the first argument as `axes` (e.g., `my_map.plot(ax)`), but the first argument is actually `annotate`.  This can result in confusing plotting behavior or outright errors.  This PR catches this common error.

In the future, we can switch `GenericMap.plot()` to accept keyword-only arguments, but that of course would be an API change, so can't be backported.


nabobalis (approved):

dstansby:
👍 to this, I opened a new issue to track making them keyword only

dstansby (approved):

PR #1432, opened 2015-06-03
Major fixes to Map Plotting 

ehsteve:
This includes a number of bug fixes and standardization for plotting of maps including better normalizations for each map. All data is now shown in each map from the minimum to maximum and the normalization is chosen so that this is so.

You can get a preview of my changes in the following [ipython notebook](https://github.com/ehsteve/ipython-notebooks/blob/master/Testing%20Plot%20Fixes%200.6.ipynb)

A build of the documentation can be found [here](https://readthedocs.org/projects/ehsteve-sunpy/)

## Needs to be done
- [x] Get consensus on normalization choices
- [ ] Documentation of normalization choices
- [x] Fix tests
- [x] Address code comments
- [x] Changelog entry

PR #7180, opened 2023-08-30
Improve handling of WISPR maps

svank:
At @Cadair 's suggestion, this PR adds nice default colorbar vmin/vmax values for PSP/WISPR maps. Without this, the bright stars cause the automatic bounds to leave the faint coronal structures mostly invisible.

The range of brightness values in a WISPR image varies significantly through each encounter, so no one set of vmin/vmax defaults will be perfect, but I've found these values to be generally good defaults in my own work.

Also offers nicer names for the two detectors---WISPR is two imagers, an inner and an outer FOV. The FITS `detector` value is `1` or `2` to indicate the source imager, but providing labels of `Inner` and `Outer` for e.g. plot titles is nicer and more in line with how the community seems to refer to the two.

Before:
`wispr_map.peek()`
image

After:
`wispr_map.peek()`
image

### DIFF
--- a/sunpy/map/sources/psp.py
+++ b/sunpy/map/sources/psp.py
@@ -36,7 +36,14 @@ class WISPRMap(GenericMap):
         lvl = self.meta.get('level', None)
         if lvl is None:
             return
-        return int(lvl[1])
+        # Chop off the leading 'L'
+        lvl = lvl[1:]
+        try:
+            return int(lvl)
+        except ValueError:
+            # The int conversion will fail for L2b files, and we should fail
+            # safe if the user chooses to customize this with other values.
+            return lvl
 
     @property
     def name(self):
--- a/sunpy/map/sources/psp.py
+++ b/sunpy/map/sources/psp.py
@@ -31,6 +31,17 @@ class WISPRMap(GenericMap):
     def name(self):
         return 'WISPR ' + super().name
 
+    @property
+    def detector(self):
+        detector = self.meta.get('detector', "")
+        if detector == 1:
+            return "Inner"
+        if detector == 2:
+            return "Outer"
+        # Official data products will only be 1 or 2, but we should fail safe
+        # if users customize this value themselves.
+        return detector
+
     @classmethod
     def is_datasource_for(cls, data, header, **kwargs):
         """Determines if header corresponds to an WISPR image"""
--- a/sunpy/map/sources/psp.py
+++ b/sunpy/map/sources/psp.py
@@ -2,6 +2,9 @@
 Parker Solar Probe subclass definitions.
 """
 
+from astropy.visualization import AsinhStretch
+from astropy.visualization.mpl_normalize import ImageNormalize
+
 from sunpy.map import GenericMap
 
 __all__ = ['WISPRMap']
@@ -20,6 +23,14 @@ class WISPRMap(GenericMap):
     * `WISPR Instrument Page <https://wispr.nrl.navy.mil//>`__
     * `Instrument Paper <https://doi.org/10.1007/s11214-014-0114-y>`__
     """
+
+    def __init__(self, data, header, **kwargs):
+        super().__init__(data, header, **kwargs)
+
+        self.plot_settings['norm'] = ImageNormalize(
+            stretch=AsinhStretch(a=0.001), clip=True,
+            vmin=0)
+
     @property
     def processing_level(self):
         lvl = self.meta.get('level', None)
--- a/sunpy/map/sources/psp.py
+++ b/sunpy/map/sources/psp.py
@@ -17,6 +17,12 @@ class WISPRMap(GenericMap):
     The The Wide-field Imager for Parker Solar Probe (WISPR) is a white light
     telescope onboard the Parker Solar Probe (PSP) spacecraft.
 
+    Notes
+    -----
+    By default, plotting of this map will set the lower bound to zero
+    (i.e., clip out negative values for pixels).  You can change this bound
+    by modifying ``.plot_settings['norm'].vmin``.
+
     References
     ----------
     * `PSP science gateway <https://sppgway.jhuapl.edu//>`__
@@ -28,22 +34,23 @@ class WISPRMap(GenericMap):
         super().__init__(data, header, **kwargs)
 
         self.plot_settings['norm'] = ImageNormalize(
-            stretch=AsinhStretch(a=0.001), clip=True,
-            vmin=0)
+            stretch=AsinhStretch(a=0.001), vmin=0)
 
     @property
     def processing_level(self):
         lvl = self.meta.get('level', None)
         if lvl is None:
             return
-        # Chop off the leading 'L'
-        lvl = lvl[1:]
+        # Chop off the leading 'L' if present
+        if lvl[0] == 'L':
+            lvl = lvl[1:]
         try:
-            return int(lvl)
+            lvl = int(lvl)
         except ValueError:
             # The int conversion will fail for L2b files, and we should fail
             # safe if the user chooses to customize this with other values.
-            return lvl
+            pass
+        return lvl
 
     @property
     def name(self):
--- a/changelog/7180.feature.rst
+++ b/changelog/7180.feature.rst
@@ -1,0 +1,1 @@
+Added new default colormap scalings for WISPR Maps. Plots are now clipped at zero, and `~astropy.visualization.AsinhStretch` is used for the scaling to ensure coronal details are visible despite the much-brighter stars. Parsing of the ``detector`` and ``level`` fields of the FITS headers is also improved.
--- a/sunpy/map/sources/tests/test_wispr_source.py
+++ b/sunpy/map/sources/tests/test_wispr_source.py
@@ -303,8 +303,8 @@ def test_unit(wispr_map):
 
 
 def test_norm_clip(wispr_map):
-    # Tests that the default normalizer has clipping enabled
-    assert wispr_map.plot_settings['norm'].clip
+    # Tests that the default normalizer has clipping disabled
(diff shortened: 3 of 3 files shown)


Cadair:
Thanks a lot @svank I will review it properly next week.

Makes me wonder exactly how much of a pain changing our default units would be (a lot).

ayshih:
I'm hesitant on hard-coding vmin/vmax, and a user could always use `clip_interval`.  How well does `AsinhStretch` work instead of hard-coding vmin/vmax?

dstansby (changes requested):
Looks good to me 👍 

Before this is merged, could you add a changelog entry?

svank:
@dstansby Done, thanks!

dstansby (approved):

ayshih:
I'll reiterate:

> I'm hesitant on hard-coding vmin/vmax, and a user could always use `clip_interval`.

We've (historically) followed the philosophy that showing all of the map data was preferable to showing a "pretty" map with defaults.  It's difficult to go outright "wrong" by showing the user all of the data, but you can confuse users about what's in the data by preemptively clipping out some of the data.  Plus, this avoided arguments about at what exact level to clip for, say, AIA maps.

Admittedly, this interest of data evaluation is partially satisfied (now) by `quicklook()` with its histogram equalization, so maybe it's okay to allow for clipping in the map sources.  However, I feel that this PR should follow the current philosophy, with a potential broad change to this philosophy as a subsequent PR (i.e., that changes the WISPR subclass in tandem with many other subclasses).

ayshih:
(We have followed this philosophy since #1432 and `sunpy` 0.6)

sunpy/map/sources/psp.py
```diff
@@ -31,6 +64,16 @@ def processing_level(self):
...
 
+    @property
+    def detector(self):
+        detector = self.meta.get('detector', "")
+        if detector == 1:
+            return "Inner"
+        if detector == 2:
+            return "Outer"
+        # Should not happen
+        return detector
```
wtbarnes, line 75: If the only allowed cases are 1 or 2, then this should be strictly enforced, i.e. this should fail if detector does not equal 1 or 2. 

```suggestion
        if detector == 1:
            return "Inner"
        elif detector == 2:
            return "Outer"
        else:
            raise ValueError(f'Unrecognized detector type {detector}')
```

wtbarnes:
@ayshih  Do you want to block this until the debate about setting hard limits on the normalization has been settled? If so, I'd suggest explicitly requesting changes on this PR.

sunpy/map/sources/psp.py
```diff
@@ -20,6 +22,37 @@ class WISPRMap(GenericMap):
...
+        }
+
+    def __init__(self, data, header, **kwargs):
+        super().__init__(data, header, **kwargs)
+
+        try:
+            detector = self.detector[0]
+            level = self.processing_level
+            vmin, vmax = self.colormap_presets[level][detector]
+        except KeyError:
```
wtbarnes, line 49: What case is this try-except catching? Are there files that don't have the detector or level set?

ayshih (changes requested):
> @ayshih Do you want to block this until the debate about setting hard limits on the normalization has been settled? If so, I'd suggest explicitly requesting changes on this PR.

Well, if you're going to make me...

\<in booming voice\>

_No map-source subclass shall specify a default value of `vmin`/`vmax` for plotting that can alter the apparent value for any pixel (i.e., via clipping) without providing due notice to the user in the docstring of said map-source subclass.  That notice shall include a justification for the value chosen for `vmin`/`vmax`._

\<deep breath\>

_If the chosen value is time-dependent (e.g., because it depends on instrument calibration), or has the potential to need to be updated in the future, the notice shall cite a person, guide, or publication with the authority (explicit or implied) to represent the standard practice (official or de facto) in the community for analyzing said map source._

sunpy/map/sources/psp.py
svank: Yeah, it's a failsafe for if the detector or level isn't in the collection of presets. For this and your other comment, the official data products should only ever have `1` or `2` for the `detector` entry in the header (I'm not a SOC member that can *guarantee* this, but I don't think they'll unveil a secret third detector any time soon). The `level` entry can also be `L1` or `L2b`---this PR was mostly just copying over snippets from my own WISPR tools, and since I don't work with those levels I didn't have presets ready, but they could be prepared.

Outside the official data products, others on the team produce `LW` and `LT` files using other processing techniques, and they put `LW` in the `level` field, for instance. One could also imagine someone producing a composite from the two imagers and putting some other value in the `detector` field to indicate this (or deleting the field altogether).

So I think either `WISPRMap` should be able to gracefully handle unexpected values in these fields, or `is_datasource_for` should only return `True` if these fields contain expected values. Not sure which is preferable.

svank:
I'm waiting for you maintainers to settle this debate. I think the coronal signal (versus the background stars) is very likely the signal of interest for almost all users, so using the full colorbar for that by default seems sensible, and the fixed vmin/vmax seems to me the best way to do that, but `AsinhStretch` was at least OK and didn't require those fixed values and I see the wisdom in not hiding new philosophies in this little PR.

ayshih:
There isn't actually any debate among the maintainers; there's broad agreement with my above decree.

As it pertains to this PR:

- `vmin`
  - Setting `vmin` to zero should be documented in the class docstring with a brief justification.
  - No reference is required as zero is a special value that is unlikely to require future updating.
- `vmax`
  - Setting `vmax` to a non-zero value is allowed only if the choice is backed by authority or reference.
  - The value, the justification, and the authority/reference should be documented in the class docstring.

In the absence of adequate authority/reference for the setting of `vmax`, I suggest that this PR go forward with `vmin=0` and `AsinhStretch`.

svank:
OK, here's a version with `AsinhStretch`. I spot-checked for L2 and L3 data from both detectors, and I think the plots look reasonable. The apparent background level can vary through an encounter, sometimes very rapidly, as `vmax` floats but I think that's just the way it is.

After @wtbarnes 's comments, I've clarified my intent in `detector()` about handling unexpected values. I also applied similar logic to `processing_level()`, ensuring that the `int` conversion fails safe for non-`int` values (which I know people use for files outside the official data products). At the risk of broadening this PR too much, I introduced one change in behavior, which I think counts as a bug fix. The official data files include `L2b` files, which are the estimated backgrounds subtracted from `L2` files to produce `L3` files. In the FITS headers, the `LEVEL` field is `L2`, `L2b`, or `L3`. Before, `processing_level()` would chop off the leading `L` by returning `int(lvl[1])`, providing a level of `2` for both `L2` and `L2b` files. Now it parses `lvl[1:]` instead, so for `L2` files it will return `2` as an int, the same as before, and it returns `2b` as a string for `L2b` files. If `processing_level()` is meant to be just convenient access to the `LEVEL` field, this ensures we're not accidentally providing the same parsed value for files with different values in the actual header.

sunpy/map/sources/psp.py
```diff
@@ -20,17 +23,43 @@ class WISPRMap(GenericMap):
     * `WISPR Instrument Page <https://wispr.nrl.navy.mil//>`__
     * `Instrument Paper <https://doi.org/10.1007/s11214-014-0114-y>`__
     """
+
+    def __init__(self, data, header, **kwargs):
+        super().__init__(data, header, **kwargs)
+
+        self.plot_settings['norm'] = ImageNormalize(
+            stretch=AsinhStretch(a=0.001), clip=True,
```
ayshih, line 31: Why are you choosing to set `clip=True`?  That is more aggressive than merely setting `vmin`.

sunpy/map/sources/psp.py
```diff
@@ -20,17 +23,43 @@ class WISPRMap(GenericMap):
...
+            vmin=0)
+
     @property
     def processing_level(self):
         lvl = self.meta.get('level', None)
         if lvl is None:
             return
-        return int(lvl[1])
+        # Chop off the leading 'L'
+        lvl = lvl[1:]
```
ayshih, line 40:
```suggestion
        # Chop off the leading 'L' if present
        lvl = lvl[1:] if lvl[0] == 'L' else lvl
```
Since you are allowing for custom levels, it seems odd not to check for whether the character being popped off is 'L' or not.

sunpy/map/sources/psp.py
```diff
@@ -20,17 +23,43 @@ class WISPRMap(GenericMap):
...
+        try:
+            return int(lvl)
+        except ValueError:
+            # The int conversion will fail for L2b files, and we should fail
+            # safe if the user chooses to customize this with other values.
+            return lvl
```
ayshih, line 46:
```suggestion
        try:
            lvl = int(lvl)
        except ValueError:
            # The int conversion will fail for L2b files, and we should fail
            # safe if the user chooses to customize this with other values.
            pass
        return lvl
```
I'm probably just being crotchety, but I don't like seeing the `return` inside a `try...except` here because the `return` will never raise an exception.

ayshih (commented):
Add a note to the docstring before the `References` section a la:
```
Notes
-----
By default, plotting of this map will set the lower bound to zero
(i.e., clip out negative values for pixels).  You can change this bound
by modifying ``.plot_settings['norm'].vmin``.

wtbarnes (approved):
Thanks @svank! This looks good to me.

sunpy/map/sources/psp.py
svank: Good catch, I think this was just left from the initial, more opinionated plot settings.

svank:
@ayshih I've made your requested changes, thanks!

ayshih (approved):
Looks good to me, thanks!

ayshih:
Also, thanks for your patience!

Issue #7184, opened 2023-09-07
`crln_obs` used instead of `hgln_obs` by `astropy.wcs` but not `sunpy.map`


svank:
### Describe the bug

I *think* this is a bug in Sunpy, but I don't think I understand the full context here.

In short, when parsing the WCS information in a FITS header, information is extracted from the FITS header to define the observer location. When creating a WCS directly, it appears to load the Carrington longitude/latitude (`crln_obs`, `crlt_obs`) and convert it to Stonyhurst coordinates, but when using a Sunpy Map, the Stonyhurst lat/lon in the header are used (`hgln_obs`, `hglt_obs`). The latter seems preferable, since as I understand it the Carrington frame has the most variation in how implementations define it. For at least files where `crln_obs` comes from SunSPICE, there's an offset between its definition and Sunpy's, meaning that the observer location is different depending on whether the file is passed to `WCS` directly or passed though `Map` (with `Map` being correct). Maybe the answer is "just use `Map`", but this seems like an inconsistency worth flagging.

I don't know if this is the right place to fix this (if this is a bug), but I've found that [here](https://github.com/sunpy/sunpy/blob/4343e44ac19c4113c3ba9205d242c5bc2e11e4b2/sunpy/coordinates/wcs_utils.py#L131), if I add a `break` statement at the bottom of this loop, passing the header to `WCS` directly behaves the same as `Map`, using `hgln_obs` if present (because the loop stops when `HeliographicStonyhurst` works). If this seems right, I can make a PR.

In more detail:

I'm working with FITS files from WISPR on PSP.
```python
from astropy.io import fits
import astropy.units as u
from astropy.wcs import WCS
import sunpy.coordinates
from sunpy.map import Map
file = '/home/svankooten/psp/WISPR/WISPR_ENC10_L3_FITS/20211117/psp_L3_wispr_20211117T120315_V1_1221.fits'
header = fits.getheader(file)
wcs = WCS(header)
map = Map(file)
```

The `SkyCoord`s produced by `wcs` and `map` (and `map.wcs`, which does the same as `map`) have different observer longitudes:
```python
In [3]: wcs.pixel_to_world(0,0).observer
Out[3]: 
<HeliographicStonyhurst Coordinate (obstime=2021-11-17T12:03:35.901, rsun=695507.968 km): (lon, lat, radius) in (deg, deg, m)
    (-30.60235774, 2.98625488, 3.0456058e+10)>
[...]

PR #7188, opened 2023-09-08
Prioritize hgln_obs over crln_obs when extracting observer information from FITS headers

svank:
Fix as proposed in #7184

Thanks @ayshih for confirming this should be right!

### DIFF
--- a/sunpy/coordinates/tests/test_wcs_utils.py
+++ b/sunpy/coordinates/tests/test_wcs_utils.py
@@ -399,3 +399,51 @@ def test_obsgeo_invalid(obsgeo):
 
     with pytest.raises(ValueError):
         obsgeo_to_frame(obsgeo, None)
+
+
+def test_observer_hgln_crln_priority():
+    """
+    When extracting the observer information from a FITS header, ensure
+    Stonyhurst (HG) coordinates are preferred over Carrington (CR) if present.
+    Going through `Map` and `WCS` seem to follow (at least slightly) different
+    code paths, so test them both.
+    """
+    data = np.ones([6, 6], dtype=np.float64)
+    header = {'CRVAL1': 0,
+              'CRVAL2': 0,
+              'CRPIX1': 5,
+              'CRPIX2': 5,
+              'CDELT1': 10,
+              'CDELT2': 10,
+              'CUNIT1': 'arcsec',
+              'CUNIT2': 'arcsec',
+              'PC1_1': 0,
+              'PC1_2': -1,
+              'PC2_1': 1,
+              'PC2_2': 0,
+              'NAXIS1': 6,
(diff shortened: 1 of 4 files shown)


sunpy/coordinates/wcs_utils.py
```diff
@@ -129,6 +129,7 @@ def solar_wcs_frame_mapping(wcs):
                              attrs[1] * u.deg,
                              attrs[2] * u.m,
                              **kwargs)
+            break
```
dstansby, line 132: Could you add a quick comment above where the `required_attrs` dictionary is defined that the order of entries matters there?

dstansby (changes requested):
Looks 👍 , just one request for a comment, and needs a bugifx changelog entry adding.

svank:
@dstansby Thanks! I've made these changes.

sunpy/coordinates/tests/test_wcs_utils.py
```diff
@@ -399,3 +399,51 @@ def test_obsgeo_invalid(obsgeo):
...
     with pytest.raises(ValueError):
         obsgeo_to_frame(obsgeo, None)
+
+
+def test_observer_hgln_crln_priority():
+    """
+    When extracting the observer information from a FITS header, ensure
+    Stonyhurst (HG) coordinates are preferred over Carrington (CR) if present.
+    Going through `Map` and `WCS` seem to follow (at least slightly) different
+    code paths, so test them both.
```
ayshih, line 409:
```suggestion
    `Map` creates a custom FITS header with a sanitized observer location, so
    we need to also test a directly instantiated `WCS` object.
```

ayshih (changes requested):
Some text edits, otherwise looks good to me!

If I were to be super-picky, the part of the unit test that tests the prioritization in the `Map` implementation should really live under `sunpy.map.tests` and not `sunpy.coordinates.tests`, because the relevant code – how `GenericMap` prioritizes and sanitizes the observer location – all lives in `sunpy/map/mapbase.py`.

Cadair:
>  the part of the unit test that tests the prioritization in the Map implementation should really live under sunpy.map.tests

Given my recent foray into seeing if it's possible to run tests for subpackages without optional dependencies installed, ideally we would not have a direct reliance on map in the coordinates tests.

svank:
What's code review for, if not being picky? I moved the `Map` part of the test to the `map` module---there is still a pre-existing `Map` dependence in `coordinates/tests/test_wcs_utils.py`, but now I avoid further entrenching that.

Text edits are also implemented.

ayshih (approved):
Looks good to me!

ayshih:
Thanks, @svank!

Issue #7179, opened 2023-08-29
Remove internal coordinate-frame docstring code now that we depend on Astropy 5.0


ayshih:
We should be able to revert #5497 now that we depend on Astropy 5.0, and our documentation should not be affected.

TanviPooranmal:
The latest PR on this issue closed months back. So, can I work on this (I know issues are not assigned)?

nabobalis:
@TanviPooranmal go ahead.

PR #7190, opened 2023-09-14
Removed internal coordinate-frame docstring code #7179 

mahinigam:
## PR Description

I have successfully reverted to #5497 as mentioned in issue #7179. I have also formatted frames.py using Black Formatter.
I have just begun contributing to Open Source. Please inform me if I have made any errors, and I will surely rectify them.


nabobalis:
Hello @mahinigam.

 Thanks for the Pull request.

I am rendering this on my phone but a quick glance makes it seem the code was run through black.
Were there any other changes?

I would also suggest if you can undo those style changes as well, that would be great. We are still unsure about whether we should use black on our codebase and as such, accepting a PR before we make that decision is tricky.

mahinigam:
ok, i will the revert the changes i made using black.

PR #7192, opened 2023-09-15
Fix agg warning filter w/ Matplotlib 3.8

dstansby:
Should fix tests with Matplotlib 3.8. I put no changelog needed as this isn't user facing.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -192,6 +192,9 @@ filterwarnings =
     ignore:Distutils was imported before Setuptools
     # Figure tests use agg backend so show doesn't work but we use it in peek
     ignore:Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
+    # Warning changed in Matplotlib 3.8; when that's our min dep, warning filter on line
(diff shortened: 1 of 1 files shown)


dstansby:
I think devdeps testing will resolve when Matplotlib upload a dev wheel that isn't exactly version 3.8.0.

ayshih:
> I think devdeps testing will resolve when Matplotlib upload a dev wheel that isn't exactly version 3.8.0.

Ha!  That's actually kinda hilarious.  I guess the `devdeps` build will always fail immediately following every minor release...

ayshih (approved):

nabobalis (approved):

nabobalis:
> > I think devdeps testing will resolve when Matplotlib upload a dev wheel that isn't exactly version 3.8.0.
> 
> Ha! That's actually kinda hilarious. I guess the `devdeps` build will always fail immediately following every minor release...

Yeah, we have this problem every time. I am unsure if this is a pytest-mpl bug or if its an ours.
One day we should fix this.

PR #7197, opened 2023-09-16
Fix colormap name in map tutorial

dstansby:
I'm guessing this change came in Matplotlib 3.8 because of https://github.com/matplotlib/matplotlib/pull/25479/.


nabobalis (approved):
LGTM. Can we override the name to be nicer again or is that impossible now?

dstansby:
I think the "name" here is the name with which the colormap is registered with Matplotlib, which is different to the "nice" human readable name given to the colormap object. So in short, this is just a Matplotlib bugfix and it's confusing that there are two different names, and there's nothing we need to change on our end.

wtbarnes (approved):

PR #7200, opened 2023-09-16
Remove deprecated extract_along_coord

dstansby:


sunpy/map/tests/test_maputils.py
```diff
@@ -277,38 +275,6 @@ def test_contains_coord(aia171_test_map):
...
-        aia171_test_map.unit
-    )
-    line_discrete_pix = aia171_test_map.world_to_pixel(line_discrete)
-    assert np.allclose(pix_diag[:, 0], line_discrete_pix.x.value)
-    assert np.allclose(pix_diag[:, 1], line_discrete_pix.y.value)
-    assert u.quantity.allclose(intensity_diag, intensity)
-
-
-@pytest.mark.filterwarnings("ignore:The extract_along_coord function is deprecated:sunpy.util.exceptions.SunpyDeprecationWarning")
-def test_extract_along_coord_one_point_exception(aia171_test_map):
```
hayesla, line 298: just looking at this, this should really be added as a test to `pixelate_coord_path`, and similarly with the test below. Maybe they can go in a different pr  though

nabobalis: If they go into a separate PR, we can backport it easier. Do we want to track this with an issue?

dstansby: Why would we need to backport anything here?

nabobalis: If we want to migrate these tests to the current API, I think those should be backported to ensure we don't break anything in the future.

hayesla: could they just be added in a new PR? (also the second test is something that is needed for the pixelate_coord_path I think (see #7205)

ayshih: Aside from the fix that is needed for #7205, these tests do not need to be preserved.

wtbarnes (approved):

Cadair (approved):

PR #7206, opened 2023-09-18
Add a check for out-of-bounds sampling by `sunpy.map.sample_at_coords()`

### LINKED ISSUE #7205: `sunpy.map.sample_at_coords()` should check that passed coordinates are on the map
### Describe the bug

There's not a check whether the passed coordinates to `sunpy.map.pixelate_coord_path` are actually within the bounds of the map. Without this its returning values that are not correct

### To Reproduce

```python

line_coords = SkyCoord([-1324, -1300], [500, 1000], unit=(u.arcsec, u.arcsec),
                       frame=aia_map.coordinate_frame)

intensity_coords = sunpy.map.pixelate_coord_path(aia_map, line_coords)
intensity = sunpy.map.sample_at_coords(aia_map, intensity_coords)
fig = plt.figure(figsize=(10, 4))
ax1 = fig.add_subplot(121, projection=aia_map)
aia_map.plot(axes=ax1, clip_interval=[1, 99.9]*u.percent)
ax1.plot_coord(intensity_coords)
ax1.plot_coord(line_coords[0], marker="o", color="blue", label="start")
ax1.plot_coord(line_coords[1], marker="o", color="green", label="end")
ax1.legend()

ax2 = fig.add_subplot(122)
ax2.plot(angular_separation, intensity)
ax2.set_xlabel("Angular distance along slit [arcsec]")
ax2.set_ylabel(f"Intensity [{aia_map.unit}]")
```
pixelate_coord_path_error

ayshih:
Closes #7205 

This PR adds a check for out-of-bounds sampling by `sunpy.map.sample_at_coords()`.  Currently, if the supplied coordinate corresponds to negative array indices, those indices will then get interpreted as negative indexing (i.e., indexing from the end) rather than an out-of-bounds request.

This omission gained new relevance with the replacement of `extract_along_coord()` by `pixelate_coord_path()`+`sample_at_coords()`, because `extract_along_coord()` had such a check.

### DIFF
--- a/changelog/7206.bugfix.rst
+++ b/changelog/7206.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug with :func:`~sunpy.map.sample_at_coords()` where sampling outside the bounds of the map would sometimes not error and instead return strange pixel values.
--- a/sunpy/map/maputils.py
+++ b/sunpy/map/maputils.py
@@ -140,6 +140,8 @@ def sample_at_coords(smap, coordinates):
     Uses nearest-neighbor interpolation of coordinates in map, as
     it effectively uses array indexing.
 
+    An error is raised if any of the coordinates fall outside the map bounds.
+
     Parameters
     ----------
     smap : `~sunpy.map.GenericMap`
@@ -156,6 +158,9 @@ def sample_at_coords(smap, coordinates):
     --------
     .. minigallery:: sunpy.map.sample_at_coords
     """
+    if not all(contains_coordinate(smap, coordinates)):
+        raise ValueError('At least one coordinate is not within the bounds of the map.')
+
     return u.Quantity(smap.data[smap.wcs.world_to_array_index(coordinates)], smap.unit)
 
 
@@ -580,6 +585,12 @@ def pixelate_coord_path(smap, coord_path, *, bresenham=False):
     -----
     If a pixel intersects the coordinate path at only its corner, it may not be
     returned due to the limitations of floating-point comparisons.
+
+    If part of the coordinate path lies outside of the bounds of the map, this
+    function will still return pixel coordinates that are consistent with the WCS of
+    the map, but attempting to obtain the map data at these pixel coordinates
+    (e.g., using :func:`~sunpy.map.sample_at_coords`) will raise an error, as these
+    pixels are not "real" and have no corresponding data.
 
     Returns
     -------
--- a/sunpy/map/tests/test_maputils.py
+++ b/sunpy/map/tests/test_maputils.py
@@ -219,6 +219,12 @@ def test_data_at_coordinates(aia171_test_map, aia_test_arc):
     assert_quantity_allclose(data[-1], intensity_along_arc[-1])
 
 
+def test_sample_out_of_bounds(aia171_test_map):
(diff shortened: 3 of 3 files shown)


changelog/7206.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed a bug with :func:`~sunpy.map.sample_at_coords()` where sampling outside the bounds of the map would sometimes not error and instead return strange pixel values.
```
hayesla, line 1: is there an ever a case where it would error?

```suggestion
Fixed a bug with :func:`~sunpy.map.sample_at_coords()` where sampling outside the bounds of the map would not error and instead return strange pixel values.
```

sunpy/map/maputils.py
```diff
@@ -581,6 +586,9 @@ def pixelate_coord_path(smap, coord_path, *, bresenham=False):
     If a pixel intersects the coordinate path at only its corner, it may not be
     returned due to the limitations of floating-point comparisons.
 
+    If the coordinate path exceeds the bounds of the map, those returned pixel
```
hayesla, line 589: this sentence is a bit confusing especially if someone was not familiar with what is going on here. Can it be more explicit? for example, saying that it will correspond to pixels that lie outside the bounds of the passed map etc.

changelog/7206.bugfix.rst
ayshih: Yes, the specific bug occurs when the array index is negative, but its absolute value is smaller than or equal to the size of that dimension.  Everywhere else out of bounds, there'd be an error.  To schematically illustrate:
```python
>>> import numpy as np
>>> x = np.arange(10)
>>> x[10]  # raises IndexError
>>> x[-11]  # raises IndexError
>>> x[-1]  # does not raise an error because it is interpreted as indexing from the end
```

ayshih: If you think of the original map as a patch on the sky, there are three patches on the sky of the same size that would not error.  The remaining 4 pi of the sky would error.

hayesla: ok, yes that makes sense - thanks

sunpy/map/maputils.py
```diff
@@ -581,6 +586,9 @@ def pixelate_coord_path(smap, coord_path, *, bresenham=False):
...
+    coordinates will not correspond to actual pixels with data.
```
dstansby, line 590:
```suggestion
    If the coordinate path is at least partly outside the edges of the map an error is raised.
```

wtbarnes (approved):

sunpy/map/maputils.py
dstansby: @hayesla how about my above suggestion?

ayshih: Your suggestion is not correct, as the point of this note is that an error is *not* raised in this situation

dstansby:
```suggestion
    If the coordinate path crosses pixels outside the bounds of the map, the returned pixel
    coordinates will not correspond to actual pixels with data.
```
Ah, how about this instead?

sunpy/map/maputils.py
ayshih: I have expanded this note a bunch

dstansby (approved):

PR #7207, opened 2023-09-18
Fix issue when `transform` keyword passed to `GenericMap.draw_contours()`

### LINKED ISSUE #7191: `GenericMap.draw_contours()`: don't run internal transform code if `transform` keyword is provided
#5398 fixed some issues in `GenericMap.draw_contours()` by internally transforming the map to the axes WCS if the map WCS was different.  However, we should run this code only if the `transform` keyword argument was not provided, because we shouldn't do the internal transformation if the user is providing a specific transformation to use.

hayesla:
Fixes #7191 

I think this now address the issue outlined

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -2350,7 +2350,12 @@ class GenericMap(NDData):
         # We do this instead of using the `transform` keyword argument so that Matplotlib does not
         # get confused about the bounds of the contours
         if self.wcs is not axes.wcs:
-            transform = axes.get_transform(self.wcs) - axes.transData  # pixel->pixel transform
+            if "transform" in contour_args:
+                transform_orig = contour_args["transform"]
+                contour_args.pop("transform")
+            else:
+                transform_orig = self.wcs
+            transform = transform_orig - axes.transData  # pixel->pixel transform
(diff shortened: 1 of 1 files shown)


sunpy/map/mapbase.py
```diff
@@ -2350,7 +2350,12 @@ def draw_contours(self, levels, axes=None, **contour_args):
         # We do this instead of using the `transform` keyword argument so that Matplotlib does not
         # get confused about the bounds of the contours
         if self.wcs is not axes.wcs:
-            transform = axes.get_transform(self.wcs) - axes.transData  # pixel->pixel transform
+            if "transform" in contour_args:
+                transform_orig = contour_args["transform"]
+                contour_args.pop("transform")
```
ayshih, line 2355:
```suggestion
                transform_orig = contour_args.pop("transform")
```

ayshih (commented):
Needs a unit test, of course, which will presumably have to be a figure test

dstansby (changes requested):
Looks like this still needs a test adding.

nabobalis:
Superseded by https://github.com/sunpy/sunpy/pull/7427

Issue #7204, opened 2023-09-18
Add a GONG map class


dstansby:
### Describe the feature

[GONG synotpic maps](https://gong.nso.edu/data/magmap/) have broken FITS metadata, and cannot be loaded using just sunpy core at the moment.

This has been fixed by pfsspy but since it is now archived, we can not ask users to install this every time.

### Proposed solution

pfsspy has a GONG map class to fix some of this metadata: https://github.com/dstansby/pfsspy/blob/f493e2cdc653e541425f01318ec2f35645ee36d0/pfsspy/map.py#L12.

Copying this class over, adding unit tests and a test header are required steps to get this into sunpy core.
 
This will allow users in future to load these type of files.

TrestanSimon:
How should the copyleft conditions of pfsspy's GPL v3 license be handled here? Per [section 5](https://github.com/dstansby/pfsspy/blob/main/LICENSE#L211-L246), modified source versions must be released under the same license. Does this condition apply here?

I have started to adapt the code and add tests, but I have yet to open a PR.

nabobalis:
My license knowledge is limited but I think we can ask the original copyright holder to relicense the code that we add to SunPy.

dstansby:
As the copyright holder of that code (specifically the GONG map class) I'm happy to relicense it under the BSD-2 clause license for sunpy.

PR #7219, opened 2023-09-30
Fix inappropriate x limits in "Flare times on a GOES XRS plot" example

TrestanSimon:
Fixes the x limits in the "Flare times on a GOES XRS plot" example to the time of interest rather than for the entire day. Closes https://github.com/sunpy/sunpy/issues/7218.

PR #7220, opened 2023-10-01
Add gong map class

TrestanSimon:
Adapts the GONG synoptic map class ~~and ADAPT map class~~ originally developed for [pfsspy](https://github.com/dstansby/pfsspy/blob/f493e2cdc653e541425f01318ec2f35645ee36d0/pfsspy/map.py#L12).

Closes https://github.com/sunpy/sunpy/issues/7204.


nabobalis:
@dstansby do you know if we need to add a licence file from pffspy to ensure the licence change?

dstansby:
No, I think my permission is enough.

@TrestanSimon could you update this PR to remove the ADAPT map class? That wasn't written by me, so will need permission of the other author, and it will be much easier to review this pull request anyway if we just keep it to the GONG map class.

TrestanSimon:
It looks like the codecov tests are failing with a ton of indirect changes to the coverage of unrelated files. This was also happening in a previous PR https://github.com/sunpy/sunpy/pull/7219 that I opened, and I was advised to ignore it.

nabobalis:
The codecov checks looks to have passed on the PR, I wouldn't worry.

TrestanSimon:
> The codecov checks looks to have passed on the PR, I wouldn't worry.

Weird... I see that now. GitHub said it was failing earlier, and the [associated page](https://app.codecov.io/gh/sunpy/sunpy/commit/b7794029fe48c1f8ace0b3860d0433552b392da4/indirect-changes) on the codecov website still lists the unusual indirect coverage changes. Thanks again, I will ignore this going forward.

sunpy/map/sources/gong.py
```diff
@@ -0,0 +1,92 @@
...
+    ----------
+    * `GONG Page <https://gong.nso.edu/>`_
+    * `Magnetogram Synoptic Map Images Page <https://gong.nso.edu/data/magmap/>`_
+    * `FITS header keywords <https://gong.nso.edu/data/DMAC_documentation/General/fitsdesc.html>`_
+    * `Instrument Paper (pp. 203–208) <https://inis.iaea.org/collection/NCLCollectionStore/_Public/20/062/20062491.pdf>`_
+    * `GONG+ Documentation <https://gong.nso.edu/data/DMAC_documentation/PipelineMap/GlobalMap.html>`_
+
+
+    """
+    def __init__(self, data, header, **kwargs):
```
Cadair, line 42: I think the way this `__init__` is written isn't the way we want to handle this type of metadata fixing any more. These fixes should be implemented by overloading the properties on the `GenericMap` class rather than editing the `.header` property in the `__init__`.

For example you would "fix" CUNIT1/2 by writing a property like this: https://github.com/sunpy/sunpy/blob/main/sunpy/map/mapbase.py#L1298

Cadair (dismissed):
Thanks a lot for the PR @TrestanSimon 

I think this needs some rewriting to match the way we have adopted doing metadata fixing in core. Let me know if you need me to give you more details on how to write all the properties.

wtbarnes:
Hi @TrestanSimon 👋 Thanks again for your contribution. We'd like to get this into the next release of sunpy (v5.1) which will tentatively be released on 3 November 2023. Do you think you'll have time to address the comments on this PR by then? If not, one of the maintainers is happy to pick this up and finish it off.

TrestanSimon:
I actually have it done and can commit and push in a few minutes. Apologies for the delay.

sunpy/map/sources/gong.py
```diff
@@ -0,0 +1,97 @@
...
+            return u.G
+        return u.Unit(unit_str)
+
+    @property
+    def spatial_units(self):
+        return SpatialPair(u.deg, u.deg)
+
+    @property
+    def observer_coordinate(self):
+        return _earth_obs_coord_meta(self.meta['date-obs'])
```
ayshih, line 75: Um, `observer_coordinate` is supposed to return a `SkyCoord`, not a dictionary...

ayshih (dismissed):

sunpy/map/sources/gong.py
TrestanSimon: I am aware. I mistakenly pushed a work-in-progress version of the code. My final copy of the code is on a different machine I have lost access to (hence the lack of any commits here recently). I thought I had transferred a version with this property addressed before losing access, but I did not.

Regardless, a forthcoming commit should have this implemented properly.

sunpy/map/sources/gong.py
```diff
@@ -0,0 +1,90 @@
...
+                           self.meta['cdelt2'] * 180 / np.pi * self.spatial_units[0] / u.pixel)
+
+    @property
+    def unit(self):
+        unit_str = self.meta.get('bunit', None)
+        if unit_str is None:
+            return
+        elif unit_str == 'Gauss':
+            return u.G
+        return u.Unit(unit_str)
```
wtbarnes, line 68:
```suggestion
```
I don't think we need this property anymore as `u.Unit ('Gauss')` works and the `GenericMap.unit` already covers the None case.

sunpy/map/sources/gong.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @property
+    def spatial_units(self):
+        return SpatialPair(u.deg, u.deg)
+
+    @property
+    def observer_coordinate(self):
+        observer_coord = get_earth(self.date)
```
wtbarnes, line 76: I'm confused as to why we can't just return `get_earth(self.date)` here.

sunpy/map/sources/gong.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @classmethod
+    def is_datasource_for(cls, data, header, **kwargs):
+        return (str(header.get('TELESCOP', '')).endswith('GONG') and
+                str(header.get('CTYPE1', '').startswith('CRLN')))
+
+    @property
+    def date(self):
+        return Time(self.meta.get('date-obs') + ' '
+                    + self.meta.get('time-obs'))
```
wtbarnes, line 52:
```suggestion
        return Time(f"{self.meta.get('date-obs')} {self.meta.get('time-obs')}")
```

sunpy/map/sources/tests/test_gong_synoptic_source.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def test_observatory(gong_synoptic):
+    """Tests the observatory property of the GongSynopticMap object."""
+    assert gong_synoptic.observatory == "NSO-GONG"
+
+
+def test_measurement(gong_synoptic):
+    """Tests the measurement property of the GongSynopticMap object."""
+    assert gong_synoptic.measurement == 676.8
```
wtbarnes, line 31: It's a bit worrying to me that the measurement (which here is just the wavelength) is being returned without a unit. In the above test header, the unit (nm) is included in the key comment, but it does not seem to be being parsed correctly. This is out of scope of this PR though so I'm OK merging and fixing this later.

sunpy/map/sources/tests/test_gong_synoptic_source.py
```diff
@@ -0,0 +1,54 @@
...
+    """Check that accessing the date doesn't raise a warning."""
+    gong_synoptic.date
+
+
+def test_unit(gong_synoptic):
+    assert gong_synoptic.unit == u.G
+    assert gong_synoptic.unit == u.Unit("Mx/cm^2")
+    assert gong_synoptic.unit.to_string() == 'G'
+    gong_synoptic.meta['bunit'] = 'm'
+    assert gong_synoptic.unit == u.m
```
wtbarnes, line 44:
```suggestion
```
I think we can remove this since we don't override the base `unit` property anymore.

wtbarnes (changes requested):
Thanks again for your contribution @TrestanSimon and thanks for your patience during the review. There have been a number of pretty big changes in how metadata is handled since this map source was written, hence all of the comments. Just a few suggested changes and a clarification about the observer coordinate.

sunpy/map/sources/gong.py
TrestanSimon: If we just use ``get_earth(self.date)``, do you think we should test that it is consistent with the metadata?

sunpy/map/sources/tests/test_gong_synoptic_source.py
wtbarnes: See #7245. If that is merged, we can rebase this PR and then change the test to check for `676.8 * u.nm`

sunpy/map/sources/gong.py
wtbarnes: I do not think these GONG maps typically carry around the metadata that specifies the observer location, hence the assumption that the observer is at Earth. Someone with more knowledge of these data products should weigh in though.

sunpy/map/sources/gong.py
```diff
@@ -0,0 +1,58 @@
...
+                str(header.get('CTYPE1', '').startswith('CRLN')))
+
+    @property
+    def date(self):
+        return Time(f"{self.meta.get('date-obs')} {self.meta.get('time-obs')}")
+
+    @property
+    def scale(self):
+        # Instead of the spacing in sin(lat), this should be 180/pi times
+        # that value (see Thompson 2005)
```
wtbarnes, line 48:
```suggestion
        # Since, this map uses the cylindrical equal-area (CEA) projection,
        # the spacing should be modified to 180/pi times the original value
        # Reference: Section 5.5, Thompson 2006
```
Sorry, missed this before. Copying over the same comment as https://github.com/sunpy/sunpy/blob/7821330e32384e3d1ee66a411738935364e1ff8b/sunpy/map/sources/sdo.py#L179-L181

wtbarnes (approved):
Looks great to me, just one minor correction on the comment about the scale correction. Thanks again @TrestanSimon!

dstansby (approved):

PR #7226, opened 2023-10-09
Update GOES tests for new files

nabobalis:
Should fix the online tests for now.

### DIFF
--- a/sunpy/util/exceptions.py
+++ b/sunpy/util/exceptions.py
@@ -9,7 +9,7 @@ import warnings
 __all__ = ["NoMapsInFileError",
            "SunpyWarning", "SunpyUserWarning", "SunpyDeprecationWarning",
            "SunpyPendingDeprecationWarning", "SunpyMetadataWarning",
-           "warn_user", "warn_deprecated", "warn_metadata"]
+           "warn_user", "warn_deprecated", "warn_metadata", "warn_connection"]
 
 
 class NoMapsInFileError(Exception):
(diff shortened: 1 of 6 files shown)


dstansby (approved):

hayesla (approved):
GOES gives again 🙏

nabobalis:
The previous Online CI failed due to a warning that it was using a different endpoint for the VSO. After speaking with @Cadair, we do not want this stop the CI as this isn't a warning that 1) is a problem unless its failed to fallback 2) means we have to run the suite again.

Due to our policy of not using warnings.warn directly, we have to create a new function even if this is getting very silly. In future, we need to refactor this.

sunpy/util/exceptions.py
```diff
@@ -57,6 +56,14 @@ class SunpyPendingDeprecationWarning(PendingDeprecationWarning, SunpyWarning):
     """
 
 
+class SunpyConnectionWarning(SunpyUserWarning):
+    """
+    A warning class to indicate a connection warning.
+
+    This will not fail the CI (via a pytest ignore) as it is not a critical warning.
```
Cadair, line 63:
```suggestion
    This warning should be issued when a recoverable error occurs during a connection to a remote server, such as falling back to a mirror etc.
    
    This will not fail the CI (via a pytest ignore) as it is not a critical warning.
```

dstansby (approved):

nabobalis:
Online test suite failed with:

```
FAILED ../../.tox/py39-online/lib/python3.9/site-packages/sunpy/net/tests/test_fido.py::test_path
FAILED ../../.tox/py39-online/lib/python3.9/site-packages/sunpy/net/vso/tests/test_vso.py::test_path
FAILED ../../docs/tutorial/acquiring_data/index.rst::index.rst
```
These are socket timeouts to the VSO all on the same file it looks like.

PR #7236, opened 2023-10-14
Added segments to be passed to drms so we can get the sums dirs

nabobalis:
TODO:

- [ ] Changelog

### DIFF
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -7,6 +7,7 @@ from pathlib import Path
 
 import drms
 import numpy as np
+import pandas as pd
 import parfive
 from packaging.version import Version
 
@@ -765,11 +766,17 @@ class JSOCClient(BaseClient):
             key = keywords
         log.debug(f"Running following query: {ds}")
         log.debug(f"Requesting following keywords: {key}")
-        result = client.query(ds, key=key, rec_index=isMeta)
-        if result is None or result.empty:
+        result = client.query(ds, key=key, seg=segments_passed, rec_index=isMeta)
+        # DRMS returns a tuple of length 2 if there is segment information
+        if len(result) == 2:
+            query_result, segment_result = result[0], result[1]
+            query_result = pd.concat([query_result, segment_result],axis=1)
+        else:
+            query_result = result
+        if query_result is None or query_result.empty:
             return astropy.table.Table()
         else:
-            return astropy.table.Table.from_pandas(result)
+            return astropy.table.Table.from_pandas(query_result)
 
     @classmethod
     def _can_handle_query(cls, *query):
--- a/sunpy/net/jsoc/tests/test_jsoc.py
+++ b/sunpy/net/jsoc/tests/test_jsoc.py
@@ -53,6 +53,8 @@ def test_return_query_args(client):
                         a.jsoc.Segment('Bp') & a.jsoc.Segment('magnetogram'))
     # Because res.query_args is list that contains dict
     assert 'primekey' in res.query_args[0]
+    assert "magnetogram" in res.keys()
+    assert res["magnetogram"][0].startswith("/SUM")
 
 
 @pytest.mark.remote_data


sunpy/net/jsoc/jsoc.py
```diff
@@ -765,11 +766,17 @@ def _lookup_records(self, iargs):
             key = keywords
         log.debug(f"Running following query: {ds}")
         log.debug(f"Requesting following keywords: {key}")
-        result = client.query(ds, key=key, rec_index=isMeta)
-        if result is None or result.empty:
+        result = client.query(ds, key=key, seg=segments_passed, rec_index=isMeta)
+        # DRMS returns a tuple of length 2 if there is segment information
+        if len(result) == 2:
+            query_result, segment_result = result[0], result[1]
+            query_result = pd.concat([query_result, segment_result],axis=1)
```
nabobalis, line 773: I am up for better suggestions that allow me to bypass importing pandas.

sunpy/net/jsoc/jsoc.py
```diff
@@ -765,11 +766,17 @@ def _lookup_records(self, iargs):
             key = keywords
         log.debug(f"Running following query: {ds}")
         log.debug(f"Requesting following keywords: {key}")
-        result = client.query(ds, key=key, rec_index=isMeta)
-        if result is None or result.empty:
+        result = client.query(ds, key=key, seg=segments_passed, rec_index=isMeta)
+        # DRMS returns a tuple of length 2 if there is segment information
+        if len(result) == 2:
```
nabobalis, line 771: This is not the cleanest code. Needs some improvment.

sunpy/net/jsoc/tests/test_jsoc.py
```diff
@@ -53,6 +53,8 @@ def test_return_query_args(client):
                         a.jsoc.Segment('Bp') & a.jsoc.Segment('magnetogram'))
     # Because res.query_args is list that contains dict
     assert 'primekey' in res.query_args[0]
+    assert "magnetogram" in res.keys()
+    assert res["magnetogram"][0].startswith("/SUM")
```
nabobalis, line 57: Instead of a new test, I just added to one that already was calling segment.

nabobalis:
> Can you explain what this is doing and why?

No and the goal is to get the location of the files on the Stanford machines. These are only returned when you pass in the segments keywords to drms which we support in Fido but never passed to drms. 

I would consider this a bug but I don't think it's so important that it needs to be backported.

sunpy/net/jsoc/tests/test_jsoc.py
Cadair, line 57: How has this changed the results table a user sees?

nabobalis: It adds a new column with the name of the segment.

Cadair:
I can't review this without some more information. 

Why would anyone want to get the sums dirs out of Fido?
How has this affected the results table? (i.e. is this a breaking change?)
Why (and how) is this only if you specify segment to a search (is this a breaking change only if segment is provided??)
Please document this somewhere to explain to me and everyone else wtf it is and why it's useful?

nabobalis:
> Why would anyone want to get the sums dirs out of Fido? 

It can be useful if you want to construct the URLS yourself to avoid an export.

> How has this affected the results table? (i.e. is this a breaking change?) 

It adds a new column to the results table.

> Why (and how) is this only if you specify segment to a search (is this a breaking change only if segment is provided??)

The segment tells the JSOC to return only specific file types but as a result of that, it also tells you the location of the files at the JSOC. You can not pass in segment only, it will error. A query string must be supplied as the segments are a function of the dataset you are asking for. 

> Please document this somewhere to explain to me and everyone else wtf it is and why it's useful?

Where would you like this information?

sunpy/net/jsoc/jsoc.py
wtbarnes: Could you instead check explicitly segments are being passed?

Issue #7066, opened 2023-06-14
Decide and document a conda dependency policy


dstansby:
### Describe the feature

Currently there are some optional pip dependencies we include as a conda dependency, and some we don't (see https://github.com/sunpy/sunpy/pull/7062#issuecomment-1587398401). We should collectively decide on a policy for conda dependencies, document it, and implement it (probably with the next major release, 5.1).

ayshih:
To be clear, we have two types of "optional" dependencies:

- Ones that are included as part of `pip install sunpy[all]` (e.g., `reproject` for sunpy.map)
- Ones that are not included as part of `pip install sunpy[all]`, but are included as part of `pip install sunpy[dev]` (`astroquery` and `jplephem` for sunpy.coordinates, `opencv` for sunpy.image/sunpy.map, and `mplcairo` for a gallery example)

The first category is included in our conda dependencies, while the second category is not.

ayshih:
xref #5974 since it wrestles with the notion of what `pip install sunpy[all]` should actually install

dstansby:
Use cases as I see them are:
- [all]: min set of deps that are needed to run all sunpy public API
- [dev]: min set of deps that are needed to run test suite, build docs, and pre-commit

Thoughts?

Once we've agreed on what each extra is for we can fit deps into each category?

PR #7237, opened 2023-10-16
Add module to use SkyCoord for SPICE computations

ayshih:
At the recent DASH meeting, there was a lot interest expressed in having SunPy support SPICE kernels at some level.  The now-archived `astrospice` package covered a little of what people use SPICE for, but is still missing many things.

This PR wraps the SPICE computations using the API of our coordinates framework (i.e., using `SkyCoord`).  Here's an example using Solar Orbiter's predicted set of SPICE kernels:
```python
>>> import astropy.units as u

>>> from sunpy.coordinates import spice

# Load the meta-kernel and create lots of frames
>>> spice.initialize('solo_ANC_soc-pred-mk.tm')
...
INFO: Installing SOLO_STIX_ILS CK frame (-144851) as 'spice_SOLO_STIX_ILS' [sunpy.coordinates.spice]
...
INFO: Installing SOLO_EUI_FSI_ILS CK frame (-144211) as 'spice_SOLO_EUI_FSI_ILS' [sunpy.coordinates.spice]
...
INFO: Installing SOLO_HEE_NASA dynamic frame (-144982) as 'spice_SOLO_HEE_NASA' [sunpy.coordinates.spice]
...
INFO: Installing SOLO_GSM dynamic frame (-144962) as 'spice_SOLO_GSM' [sunpy.coordinates.spice]
...

# Use SPICE to get the location of Solar Orbiter in a particular SPICE frame
>>> solo = spice.get_body('SOLO', '2024-01-01', spice_frame='SOLO_HEE_NASA')
>>> solo
<SkyCoord (spice_SOLO_HEE_NASA: obstime=2024-01-01T00:00:00.000): (lon, lat, distance) in (deg, deg, km)
    (-17.54823605, 4.15617429, 1.42506511e+08)>

# Transform the location to a different SPICE frame
>>> solo.spice_SOLO_GSM
<SkyCoord (spice_SOLO_GSM: obstime=2024-01-01T00:00:00.000): (lon, lat, distance) in (deg, deg, km)
    (73.47437224, 26.6713468, 45577871.53062959)>

# Retrieve the instrument line-of-sight (boresight) for STIX
>>> stix_ils = SkyCoord(0*u.deg, 0*u.deg, frame='spice_SOLO_STIX_ILS', obstime='2024-01-01')
>>> stix_ils
<SkyCoord (spice_SOLO_STIX_ILS: obstime=2024-01-01T00:00:00.000): (lon, lat) in (deg, deg)
    (0., 0.)>

# Transform the STIX boresight to the instrument frame of EUI FSI
>>> stix_ils.spice_SOLO_EUI_FSI_ILS
<SkyCoord (spice_SOLO_EUI_FSI_ILS: obstime=2024-01-01T00:00:00.000): (lon, lat) in (deg, deg)
    (-0.03, 0.035)>
```

TODO for this PR:

- [x] Better document the existence of `to_helioprojective()`
- [x] Write a unit tests for `get_fov()`
- [x] Write a unit test for `to_helioprojective()`
- [x] Add more implementation-approach details so that other developers can actually understand and maintain this
- [x] Add a way to install a built-in SPICE frame (`install_frame()`)

Deferred to future PRs:

- Create custom test SPICE kernels rather than downloading "live" ones so that at least some of the tests can be run offline
- Figure out whether the transformation of velocities is consistent with how Astropy does it
- Add support for elliptical FOVs (I'd like a "live" instrument kernel that actually defines such an FOV)

### DIFF
--- a/examples/units_and_coordinates/spice.py
+++ b/examples/units_and_coordinates/spice.py
@@ -166,10 +166,11 @@ print(stix_fov)
 
 ###############################################################################
 # More usefully, every coordinate in a SPICE frame has a
-# ``to_helioprojective()`` method that converts the coordinate to
-# `~sunpy.coordinates.Helioprojective` with the ``observer`` at the center of
-# the SPICE frame.  For the 'SOLO_STIX_ILS' frame, the center is Solar Orbiter,
-# which is exactly what we want.
+# :meth:`~sunpy.coordinates.spice.SpiceBaseCoordinateFrame.to_helioprojective`
+# method that converts the coordinate to `~sunpy.coordinates.Helioprojective`
+# with the ``observer`` at the center of te SPICE frame.  For the
+# 'SOLO_STIX_ILS' frame, the center is Solar Orbiter, which is exactly what we
+# want.
 
 print(stix_fov.to_helioprojective())
 
--- a/examples/units_and_coordinates/spice.py
+++ b/examples/units_and_coordinates/spice.py
@@ -1,0 +1,144 @@
+"""
+============================================
+Coordinates computations using SPICE kernels
+============================================
+
+How to use SPICE kernels provided by space missions to perform coordinates
+computations
+
+The `SPICE <https://naif.jpl.nasa.gov/naif/>`__ observation geometry information
+system is being increasingly used by space missions to describe the locations of
+spacecraft and the time-varying orientations of reference frames.  Here are two
+examples of mission SPICE kernels:
+
+* `Solar Orbiter <https://www.cosmos.esa.int/web/spice/solar_orbiter>`__
+* `Parker Solar Probe <https://sppgway.jhuapl.edu/ancil_products>`__
+
+The `sunpy.coordinates.spice` module enables the use of the
+`~astropy.coordinates.SkyCoord` to API to perform SPICE computations such as the
+location of bodies or the transformation of a vector from one coordinate frame
+to another coordinate frame.  Although SPICE kernels can define coordinate
+frames that are very similar to the frames that `sunpy.coordinates` already
+provides, there will very likely be slight differences.  Using
+`sunpy.coordinates.spice` will ensure that the definitions are exactly what the
+mission specifies and that the results are identical to other implementations
+of SPICE (e.g., CSPICE or Icy).
+
+.. note::
+    This example requires the optional dependency `~spiceypy.spiceypy` to be
+    installed.
+
+"""
+import matplotlib.pyplot as plt
+import numpy as np
+from matplotlib.dates import DateFormatter
+
+import astropy.units as u
+from astropy.coordinates import SkyCoord
+
+from sunpy.coordinates import spice
+from sunpy.data import cache
+from sunpy.time import parse_time
+
+###############################################################################
+# Download a small subset (~30 MB) of the Solar Orbiter SPICE kernel set that
+# corresponds to about a day of the mission.
+
+obstime = parse_time('2022-10-12') + np.arange(720) * u.min
+
+kernel_urls = [
+    "ck/solo_ANC_soc-sc-fof-ck_20180930-21000101_V03.bc",
+    "ck/solo_ANC_soc-stix-ck_20180930-21000101_V03.bc",
+    "ck/solo_ANC_soc-flown-att_20221011T142135-20221012T141817_V01.bc",
+    "fk/solo_ANC_soc-sc-fk_V09.tf",
+    "fk/solo_ANC_soc-sci-fk_V08.tf",
+    "lsk/naif0012.tls",
+    "pck/pck00010.tpc",
+    "sclk/solo_ANC_soc-sclk_20231015_V01.tsc",
+    "spk/de421.bsp",
+    "spk/solo_ANC_soc-orbit-stp_20200210-20301120_280_V1_00288_V01.bsp",
+]
+kernel_urls = [f"http://spiftp.esac.esa.int/data/SPICE/SOLAR-ORBITER/kernels/{url}"
+               for url in kernel_urls]
+
+kernel_files = [cache.download(url) for url in kernel_urls]
+
+###############################################################################
+# Initialize `sunpy.coordinates.spice` with these kernels, which will create
+# classes for `~astropy.coordinates.SkyCoord` to use.
+
+spice.initialize(kernel_files)
+
+###############################################################################
+# We can request the location of the spacecraft in any SPICE frame.  Here, we
+# request it in 'SOLO_HEEQ', which is Stonyhurst heliographic coordinates.
+
+spacecraft = spice.get_body('Solar Orbiter', obstime, spice_frame='SOLO_HEEQ')
+print(spacecraft[:4])
+
+###############################################################################
+# Plot the radial distance from the Sun over the time range.
+
+fig = plt.figure()
+ax = fig.add_subplot()
+ax.plot(obstime.datetime64, spacecraft.distance.to('AU'))
+ax.xaxis.set_major_formatter(DateFormatter('%H:%M'))
+ax.set_xlabel('2022 October 12 (UTC)')
+ax.set_ylabel('Radial distance (AU)')
+ax.set_title('Solar Orbiter distance from Sun center')
+
+###############################################################################
+# We can then transform the coordinate to a different SPICE frame.  When
+# specifying the frame for `~astropy.coordinates.SkyCoord`, SPICE frame names
+# should be prepended with ``'spice_'``.  Here, we transform it to 'SOLO_GAE'.
+
+spacecraft_gae = spacecraft.spice_SOLO_GAE
+print(spacecraft_gae[:4])
+
+###############################################################################
+# Plot the radial distance from the Earth over the time range.
+
+fig = plt.figure()
+ax = fig.add_subplot()
+ax.plot(obstime.datetime64, spacecraft_gae.distance.to('AU'))
+ax.xaxis.set_major_formatter(DateFormatter('%H:%M'))
+ax.set_xlabel('2022 October 12 (UTC)')
+ax.set_ylabel('Radial distance (AU)')
+ax.set_title('Solar Orbiter distance from Earth center')
+
+###############################################################################
+# Finally, we can leverage the Solar Orbiter SPICE kernels to look at
+# instrument pointing.  Let's define a coordinate that points directly along
+# the line of sight of the STIX instrument.
+
+stix_ils = SkyCoord(np.repeat(0*u.deg, len(obstime)),
+                    np.repeat(0*u.deg, len(obstime)),
+                    frame='spice_SOLO_STIX_ILS', obstime=obstime)
+print(stix_ils[:4])
+
+###############################################################################
+# We can transform that line of sight to the SPICE frame 'SOLO_SUN_RTN', which
+# is helioprojective coordinates as observed from Solar Orbiter.  The latitude
+# and longitude values of the resulting coordinate are the pitch and yaw
+# offsets from disk center, respectively.
+
+stix_ils_rtn = stix_ils.spice_SOLO_SUN_RTN
+print(stix_ils_rtn[:4])
+
+###############################################################################
+# Plot the pitch/yaw offsets over the time range.
+
+fig, ax = plt.subplots(2, 1)
+ax[0].plot(obstime.datetime64, stix_ils_rtn.lat.to('arcsec'))
+ax[0].xaxis.set_major_formatter(DateFormatter('%H:%M'))
+ax[0].set_xlabel('2022 October 12 (UTC)')
+ax[0].set_ylabel('Pitch offset (arcsec)')
+ax[1].plot(obstime.datetime64, stix_ils_rtn.lon.to('arcsec'))
+ax[1].xaxis.set_major_formatter(DateFormatter('%H:%M'))
+ax[1].set_xlabel('2022 October 12 (UTC)')
+ax[1].set_ylabel('Yaw offset (arcsec)')
+ax[0].set_title('Pointing offset of STIX from disk center')
+
+plt.show()
+
+# sphinx_gallery_thumbnail_number = 3
--- a/examples/units_and_coordinates/spice.py
+++ b/examples/units_and_coordinates/spice.py
@@ -71,6 +71,14 @@ kernel_files = [cache.download(url) for url in kernel_urls]
 spice.initialize(kernel_files)
 
 ###############################################################################
+# The above call automatically installs all SPICE frames defined in the
+# kernels, but you may also want to use one of the built-in SPICE frames (e.g.,
+# inertial frames or body-fixed frames).  Here, we manually install the
+# 'IAU_SUN' built-in SPICE frame for potential later use.
+
+spice.install_frame('IAU_SUN')
+
+###############################################################################
 # We can request the location of the spacecraft in any SPICE frame.  Here, we
 # request it in 'SOLO_HEEQ', which is Stonyhurst heliographic coordinates.
 
--- a/examples/units_and_coordinates/spice.py
+++ b/examples/units_and_coordinates/spice.py
@@ -4,7 +4,7 @@ Coordinates computations using SPICE kernels
 ============================================
 
 How to use SPICE kernels provided by space missions to perform coordinates
-computations
+computations.
 
 The `SPICE <https://naif.jpl.nasa.gov/naif/>`__ observation geometry information
 system is being increasingly used by space missions to describe the locations of
@@ -15,7 +15,7 @@ examples of mission SPICE kernels:
 * `Parker Solar Probe <https://sppgway.jhuapl.edu/ancil_products>`__
 
 The `sunpy.coordinates.spice` module enables the use of the
-`~astropy.coordinates.SkyCoord` to API to perform SPICE computations such as the
+`~astropy.coordinates.SkyCoord` API to perform SPICE computations such as the
 location of bodies or the transformation of a vector from one coordinate frame
 to another coordinate frame.  Although SPICE kernels can define coordinate
 frames that are very similar to the frames that `sunpy.coordinates` already
@@ -92,7 +92,7 @@ ax.set_title('Solar Orbiter distance from Sun center')
 # specifying the frame for `~astropy.coordinates.SkyCoord`, SPICE frame names
 # should be prepended with ``'spice_'``.  Here, we transform it to 'SOLO_GAE'.
 
-spacecraft_gae = spacecraft.spice_SOLO_GAE
+spacecraft_gae = spacecraft.transform_to("spice_SOLO_GAE")
 print(spacecraft_gae[:4])
 
 ###############################################################################
@@ -122,7 +122,7 @@ print(stix_ils[:4])
 # and longitude values of the resulting coordinate are the pitch and yaw
 # offsets from disk center, respectively.
 
-stix_ils_rtn = stix_ils.spice_SOLO_SUN_RTN
+stix_ils_rtn = stix_ils.transform_to("spice_SOLO_SUN_RTN")
 print(stix_ils_rtn[:4])
 
 ###############################################################################
--- a/examples/units_and_coordinates/spice.py
+++ b/examples/units_and_coordinates/spice.py
@@ -110,22 +110,14 @@ ax.set_title('Solar Orbiter distance from Earth center')
 ###############################################################################
 # We can also leverage the Solar Orbiter SPICE kernels to look at instrument
 # pointing.  Let's define a coordinate that points directly along the line of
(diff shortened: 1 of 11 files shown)


dstansby:
Given SPICE is used by a wider set of people (/ communities) than just solar physics, I think it's worth having a conversation about whether putting this in sunpy is the best way forward, compared to a separate package or in astropy core. In particular, having a separate package might encourange a wider set of contributors and eyes on the code from different research communities.

(I appreciate astrospice is archived, but that doesn't stop someone forking it and continuing development)

Cadair:
First off, this is awesome.

astrospice also has helpers for getting kernels etc from the net iirc, which is probably useful additional functionality to go along with this? We could resurrect astrospice if we think that's the best place for this to live, but I am not sure I know enough about SPICE and how hard all this is going to be to maintain that I can make a very qualified decision.

ayshih:
In the long term, a mature version of this bridging module should probably live in Astropy core.  I feel that it is too few code lines and too integrated into the coordinates framework to separate out to its own package (e.g., `astrospice`).  (Downloading SPICE kernels should probably always be a separate package, regardless of where this bridging module lives.)

In the short term, I think maturing this module in SunPy core is the most effective approach.  Development can be agile while also automatically distributing the module to the community where the module will have the most impact, and thus get the relevant feedback.  Solar missions are the special nexus where scientists both are actively using the `SkyCoord` API (which planetary folks aren't really yet) and are actively using SPICE kernels (which the non-solar astrophysicists aren't really yet).

setup.cfg
```diff
@@ -110,6 +110,7 @@ docs =
   astroquery
   hvpy>=1.0.1
   jplephem
+  spiceypy
```
wtbarnes, line 113: Could we add a new optional dependency category, e.g. `spice` or even `spiceypy`? We already do this with `dask`. It seems weird that a user would need to manage this dependency with the `docs` optional dependency category

Cadair: +1 to `[spice]`.

sunpy/coordinates/spice.py
```diff
@@ -0,0 +1,169 @@
...
+    spiceypy.furnsh(kernels)
+
+    for class_num in range(1, 7):
+        frames = spiceypy.kplfrm(class_num)
+        for frame_id in frames:
+            _install_frame(frame_id)
+
+
+# TODO: Add support for array time
+# TODO: Add support for light travel time correction
```
wtbarnes, line 135: If these TODOs don't end up in this first PR, let's make sure to convert these to issues so we don't lose track of them!

sunpy/coordinates/spice.py
```diff
@@ -0,0 +1,170 @@
...
+from sunpy.time.time import _variables_for_parse_time_docstring
+from sunpy.util.decorators import add_common_docstring
+
+__all__ = ['get_body', 'initialize']
+
+
+# Note that this epoch is very slightly different from the typical definition of J2000.0 (in TT)
+_ET_REF_EPOCH = Time('J2000', scale='tdb')
+
+_CLASS_TYPES = ['', 'inertial', 'PCK', 'CK', 'TK', 'dynamic', 'switch']
```
Cadair, line 55: this is a real nit, but this is probably better as a tuple?

sunpy/coordinates/spice.py
```diff
@@ -0,0 +1,170 @@
...
+    pass
+
+
+def _convert_to_et(time):
+    return (time - _ET_REF_EPOCH).to_value('s')
+
+
+def _install_frame(frame_id):
+    frame_name = spiceypy.frmnam(frame_id)
+    # TODO: Sanitize/escape the frame name of special characters
```
Cadair, line 69: Do we need to sanitise it to a valid Python identifier? If so we do something similar in net for the attributes on `Attr`: https://github.com/sunpy/sunpy/blob/main/sunpy/net/attr.py#L241 perhaps we could spin that out into a util function?

sunpy/coordinates/tests/test_spice.py
```diff
@@ -0,0 +1,82 @@
...
+import astropy.units as u
+from astropy.coordinates import SkyCoord, frame_transform_graph
+from astropy.tests.helper import assert_quantity_allclose
+from astropy.utils.data import download_file
+
+from sunpy.coordinates import spice
+
+
+@pytest.mark.remote_data
+def setup_module():
```
Cadair, line 13: can this not be a pytest fixture? I think you can have tests which use a fixture inherit the marks on the fixture one way or another.

You could also make this a module scoped fixture so it is only called once.

dstansby:
> In the long term, a mature version of this bridging module should probably live in Astropy core.

If that's the case, isn't it worth PR'ing this to astropy to get eyes on it from the start from the developers of where we want it to end up? Short term it might be easier for us (sunpy develoeprs) to quickly put something in sunpy, but since this is a major new feature I think it's worth putting the effort in to have design decisions with people outside the solar physics community before jumping in.

sunpy/coordinates/spice.py
ayshih: Heh, if you want to be super-picky, it should actually be a dictionary, since really only values 2 through 5 are possible...

sunpy/coordinates/spice.py
ayshih: The allowed characters for a SPICE frame name are alphanumeric, underscore, plus, and minus.  I don't know what to do with plus or minus characters.  Presumably the least bad choice would be to replace them with underscores, but collisions would be bad.  I'm keeping SPICE frame names in all uppercase, so I could replace them with lowercase p/n characters?

sunpy/coordinates/tests/test_spice.py
Cadair: I think this should be a fixture, even if we have to mark all the tests explicitly.

dstansby:
Discussed briefly on the weekly call, and consesnus was to go ahead and put this in sunpy core

sunpy/coordinates/tests/test_spice.py
ayshih: Done

sunpy/coordinates/spice.py
ayshih: > I'm keeping SPICE frame names in all uppercase, so I could replace them with lowercase p/n characters?

This is what I've done

sunpy/coordinates/spice.py
ayshih: Made a dictionary

sunpy/coordinates/spice.py
ayshih: Heh, I addressed all of my TODOs.  I'm done now, right? =)

sunpy/coordinates/spice.py
```diff
@@ -0,0 +1,217 @@
...
+def _install_frame_by_id(frame_id):
+    frame_name = spiceypy.frmnam(frame_id)
+
+    # Replace plus/minus characters in the SPICE frame name with lowercase 'p'/'n'
+    frame_name = frame_name.translate(str.maketrans("+-", "pn"))
+
+    frame_center, class_num, _ = spiceypy.frinfo(frame_id)
+    frame_center_name = spiceypy.bodc2n(frame_center)
+    log.info(f"Installing {frame_name} {_CLASS_TYPES[class_num]} frame ({frame_id}) "
+             f"as 'spice_{frame_name}'")
```
Cadair, line 82: Does any of this need some error handling? Can you end up in a situation where you request a frame_id which doesn't exist?

sunpy/coordinates/spice.py
```diff
@@ -0,0 +1,217 @@
...
+        et = _convert_to_et(from_spice_coord.obstime)
+        # .copy() is necessary due to a bug (https://github.com/astropy/astropy/issues/15503)
+        matrix = spiceypy.sxform(frame_name, 'J2000', et)[..., :3, :3].copy()
+        shifted_new_pos = from_spice_coord.cartesian.transform(matrix)
+        icrs_offset = spiceypy.spkpos(frame_center_name,
+                                      et, 'J2000', 'NONE', 'SSB')[0] << u.km
+        new_pos = shifted_new_pos + CartesianRepresentation(icrs_offset.T)
+        return to_icrs_frame.realize_frame(new_pos)
+
+    frame_transform_graph._add_merged_transform(spice_frame, ICRS, spice_frame)
```
Cadair, line 112: Random aside, but we are making heavy use of this method in sunpy, should we open a PR to astropy to make it not private?

sunpy/coordinates/spice.py
```diff
@@ -0,0 +1,217 @@
...
+                                      et, 'J2000', 'NONE', 'SSB')[0] << u.km
+        shifted_old_pos = from_icrs_coord.cartesian - CartesianRepresentation(icrs_offset.T)
+        new_pos = shifted_old_pos.transform(matrix)
+        return to_spice_frame.realize_frame(new_pos)
+
+    @frame_transform_graph.transform(FunctionTransformWithFiniteDifference, spice_frame, ICRS)
+    def spice_to_icrs(from_spice_coord, to_icrs_frame):
+        et = _convert_to_et(from_spice_coord.obstime)
+        # .copy() is necessary due to a bug (https://github.com/astropy/astropy/issues/15503)
+        matrix = spiceypy.sxform(frame_name, 'J2000', et)[..., :3, :3].copy()
```
Cadair, line 105: An alternative to `copy()` would be `np.ascontiguiousarray`? it's more explicit as to what you are trying to achieve and it's theoretically possible it's a noop unlike copy (although here I guess it makes no difference)?

sunpy/coordinates/tests/test_spice.py
```diff
@@ -0,0 +1,104 @@
...
+    pck = download_file("https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/pck00011_n0066.tpc")
+
+    # Use frames (FK) from SunSPICE
+    fk = download_file("https://soho.nascom.nasa.gov/solarsoft/packages/sunspice/data/heliospheric.tf")
+
+    spice.initialize([spk, lsk, pck, fk])
+
+    yield
+
+    # The fact that we need to do this suggests that the cache is not persistent across tests...
```
Cadair, line 30: I am not sure I follow you here, what cache?

Cadair (commented):
Other than a few coverage gaps (primarily get_body with an observer by the looks of it) I think this looks good to go.

sunpy/coordinates/spice.py
ayshih: As a private function, it doesn't require any error handling, because the calling code will be handing it valid frame IDs.  And besides, if it somehow receives an invalid frame ID, `spicepy` will raise an error at the first line with `frmnam()`, so that's your validity check right there.

sunpy/coordinates/spice.py
ayshih: I'm pretty sure it makes no difference functionally, but sure, it's clearer.  Just...so...many...characters...

sunpy/coordinates/tests/test_spice.py
ayshih: My hope was that by using Astropy`s `download_file()`, I would be using the same cache that is used when downloading the DE440s JPL ephemeris in other unit tests.  However, I suspect that `pytest` is doing too good of a job of isolation (without setting things to session scope).

sunpy/coordinates/spice.py
ayshih: Changed

sunpy/coordinates/spice.py
Cadair: oh maybe I am getting confused between your geo frames PR and the merged code lol. Either way, I think that if we are going to rely on it we should ask astropy to mark it as public, mainly so someone doesn't change it somehow thinking that's a safe thing to do?

ayshih:
This module is in very good shape, and I added a gallery example, so bring on the reviews!

examples/units_and_coordinates/spice.py
```diff
@@ -0,0 +1,142 @@
...
+from sunpy.data import cache
+from sunpy.time import parse_time
+
+###############################################################################
+# Download a subset (~30 MB) of the Solar Orbiter SPICE kernel set that
+# corresponds to about a day of the mission.
+
+obstime = parse_time('2022-10-12') + np.arange(720) * u.min
+
+kernel_urls = [
```
Cadair, line 49: We clearly need some functionality for getting kernels. Is the functionality in astrospice capable of doing this?

examples/units_and_coordinates/spice.py
```diff
@@ -0,0 +1,142 @@
...
+                    frame='spice_SOLO_STIX_ILS', obstime=obstime)
+print(stix_ils[:4])
+
+###############################################################################
+# We can transform that line of sight to the SPICE frame 'SOLO_SUN_RTN', which
+# is helioprojective coordinates as observed from Solar Orbiter.  The latitude
+# and longitude values of the resulting coordinate are the pitch and yaw
+# offsets from disk center, respectively.
+
+stix_ils_rtn = stix_ils.spice_SOLO_SUN_RTN
```
Cadair, line 123:
```suggestion
stix_ils_rtn = stix_ils.transform_to("spice_SOLO_SUN_RTN")
```

examples/units_and_coordinates/spice.py
ayshih: `astrospice` has some limited capability of doing this, but of course it's at a very early stage of development.  For example, for Solar Orbiter, `astrospice` downloads only the SPK kernel for spacecraft location, and only the predicted kernel at that.  In order to support all things SPICE, `astrospice` would need to be fleshed out to download all of the different types of kernels.

examples/units_and_coordinates/spice.py
```diff
@@ -0,0 +1,142 @@
...
+ax.set_xlabel('2022 October 12 (UTC)')
+ax.set_ylabel('Radial distance (AU)')
+ax.set_title('Solar Orbiter distance from Sun center')
+
+###############################################################################
+# We can then transform the coordinate to a different SPICE frame.  When
+# specifying the frame for `~astropy.coordinates.SkyCoord`, SPICE frame names
+# should be prepended with ``'spice_'``.  Here, we transform it to 'SOLO_GAE'.
+
+spacecraft_gae = spacecraft.spice_SOLO_GAE
```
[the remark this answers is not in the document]
ayshih: Possibly, so sure

examples/units_and_coordinates/spice.py
```diff
@@ -0,0 +1,144 @@
+"""
+============================================
+Coordinates computations using SPICE kernels
+============================================
+
+How to use SPICE kernels provided by space missions to perform coordinates
+computations
```
Cadair, line 7:
```suggestion
computations.
```

examples/units_and_coordinates/spice.py
```diff
@@ -0,0 +1,144 @@
...
+The `SPICE <https://naif.jpl.nasa.gov/naif/>`__ observation geometry information
+system is being increasingly used by space missions to describe the locations of
+spacecraft and the time-varying orientations of reference frames.  Here are two
+examples of mission SPICE kernels:
+
+* `Solar Orbiter <https://www.cosmos.esa.int/web/spice/solar_orbiter>`__
+* `Parker Solar Probe <https://sppgway.jhuapl.edu/ancil_products>`__
+
+The `sunpy.coordinates.spice` module enables the use of the
+`~astropy.coordinates.SkyCoord` to API to perform SPICE computations such as the
```
Cadair, line 18:
```suggestion
`~astropy.coordinates.SkyCoord` API to perform SPICE computations such as the
```

examples/units_and_coordinates/spice.py
```diff
@@ -0,0 +1,144 @@
...
+ax.set_ylabel('Radial distance (AU)')
+ax.set_title('Solar Orbiter distance from Earth center')
+
+###############################################################################
+# Finally, we can leverage the Solar Orbiter SPICE kernels to look at
+# instrument pointing.  Let's define a coordinate that points directly along
+# the line of sight of the STIX instrument.
+
+stix_ils = SkyCoord(np.repeat(0*u.deg, len(obstime)),
+                    np.repeat(0*u.deg, len(obstime)),
```
Cadair, line 115: Am I right in thinking that with astropy/astropy#15121 we wont need to do this any more?

sunpy/coordinates/spice.py
```diff
@@ -0,0 +1,275 @@
+"""
+Bridge module to use the SkyCoord API for SPICE computations
```
Cadair, line 2:
```suggestion
Bridge module to use the SkyCoord API for SPICE computations.
```

sunpy/coordinates/spice.py
```diff
@@ -0,0 +1,275 @@
...
+def _astropy_center_name(spice_center_id):
+    # Use the center ID directly, changing a negative sign to 'n'
+    return f"_spice_center_{str(spice_center_id).replace('-', 'n')}"
+
+
+def _is_2d(data):
+    return data.norm().unit is u.one and u.allclose(data.norm(), 1*u.one)
+
+
+def _install_center_by_id(center_id):
```
Cadair, line 89: Can you explain to me (and maybe in a docstring) a little more about what a "center" is here. I am a little worried that there's a lot of stuff which is :magic_wand: here if you aren't that familiar with spice.

Cadair (approved):
This is awesome, and I am happy for it to go in as it is.

The thing that I think is lacking here is documentation, if you are familiar with SPICE then this PR as it is is probably great for you, but there's no docs on getting started with it for people like me who have no experience with it. Take the statement "We can transform that line of sight to the SPICE frame 'SOLO_SUN_RTN', which is helioprojective coordinates as observed from Solar Orbiter." from the example. How is anyone supposed to know that :rofl:.

Also, along the same lines, there isn't a lot of documentation / notes for developers on why the spice module does certain things etc (like my center comment).

I am fine for these things to be "future work" and opened as issues.

examples/units_and_coordinates/spice.py
samaloney: yea code to download/manage the kernels would be super handy for example give a SO spice repo and meta kernel path code could download and load the spice kernels

sunpy/coordinates/spice.py
ayshih: Fixed

examples/units_and_coordinates/spice.py
ayshih: Fixed

examples/units_and_coordinates/spice.py
ayshih: Fixed

examples/units_and_coordinates/spice.py
ayshih: I believe so

setup.cfg
```diff
@@ -119,6 +122,7 @@ docs =
   sphinx-design
   sphinx-gallery>=0.13.0 # First to allow us to use a different backend while scraping the images
   sphinxext-opengraph
+  %(spice)s  # because importing sunpy.coordinates.spice raises an error if spiceypy is not installed
```
Cadair, line 125: Do we not want this included in `[all]`?

sunpy/coordinates/spice.py
```diff
@@ -0,0 +1,370 @@
...
+            obspos = matrix @ obspos
+        pos += obspos << u.km
+
+    frame_name = 'icrs' if spice_frame == 'J2000' else _astropy_frame_name(spice_frame)
+
+    return SkyCoord(CartesianRepresentation(pos.T), frame=frame_name, obstime=obstime)
+
+
+@add_common_docstring(**_variables_for_parse_time_docstring())
+def get_fov(instrument, time, *, resolution=100):
```
Cadair, line 330: Could [astropy's regions](https://astropy-regions.readthedocs.io/en/stable/index.html) package be useful here?

ayshih: Er, perhaps, marginally, at least for circular FOVs, but wouldn't technically do rectangular FOVs correctly.  Besides, `regions` hasn't had a release for over a year; is it something we'd want to depend on?

setup.cfg
ayshih: It's worth discussing (a la https://github.com/sunpy/sunpy/issues/7066#issuecomment-1591721444).  `spiceypy` is akin to other optional coordinates dependencies (`astroquery`/`jplephem`) that are relevant to only a small subset of users, and those dependencies also aren't included in `[all]`.

sunpy/coordinates/spice.py
ayshih: I added some developer notes

sunpy/coordinates/spice.py
Cadair: Yeah I am not sure what the status of it is, I can ask around. Maybe we should look at what we would want it to do to support this?

nabobalis: If you can ask around, we want to use it for the HEK client in the future.

nabobalis (approved):

PR #7239, opened 2023-10-17
Add coordinate frames that depend on the Earth's magnetic field (MAG, SM, and GSM)

ayshih:
For reasons!

The PR bundles in the latest IGRF model (IGRF-13) for the orientation of Earth's magnetic dipole, and the user can select earlier model iterations, which are downloaded and cached indefinitely.

TODO:
- [x] Frame docstrings
- [x] Lots of accuracy unit tests
- [x] Add ASDF stuff


sunpy/coordinates/frames.py
```diff
@@ -847,3 +856,99 @@ class GeocentricEarthEquatorial(SunPyBaseCoordinateFrame):
...
     equinox = TimeFrameAttributeSunPy(default=_J2000)
+
+
+class BaseMagnetic(SunPyBaseCoordinateFrame):
+    """
+    Base class for frames that rely on the Earth's magnetic model (MAG, SM, and GSM).
+
+    This class is not intended to be used directly and has no transformations defined.
+    """
+    magnetic_model = Attribute(default='igrf13')
```
Cadair, line 867: Does the "igrf13" model get updated every 5 years? So the file at the URL below just gets changed?

ayshih: No, the file `igrf13coeffs.txt` should never change.  In ~2025, there will be a newly available `igrf14coeffs.txt`.

sunpy/coordinates/frames.py
```diff
@@ -847,3 +856,99 @@ class GeocentricEarthEquatorial(SunPyBaseCoordinateFrame):
...
+
+    @property
+    def _lowest_igrf_coeffs(self):
+        if not self.magnetic_model.startswith("igrf"):
+            raise ValueError
+
+        igrf_file = download_file("https://www.ngdc.noaa.gov/IAGA/vmod/coeffs/"
+                                  f"{self.magnetic_model}coeffs.txt", cache=True)
```
wtbarnes, line 875: Can we use the data manager here instead?

sunpy/coordinates/frames.py
Cadair: So what would be our update plan? Change the default model name in a new feature release of sunpy?

Given this I would think the best idea would be to have a way of first checking if we have included the file and then trying to fetch it from online if not. This would let people use newer ones their version does not support as well as older ones we might not have included if they are trying to reproduce some analysis.

I tested to see if you can pass a `file://` url to our data manager / cache but unfortunately not. I think that would be my preferred solution long term (get them from the global cache with the first mirror being a `file://`).

sunpy/coordinates/frames.py
ayshih: Er, no, because this doesn't follow the model of our data manager:

- The required file is one of several possibilities, but which one is actually needed is resolved only at runtime.  I wouldn't want to blanket require all of the possibilities, so I'd have to jump through some hoops with subfunctions in order to avoid downloading what is not needed.
- I want to be able to immediately support `igrf14coeffs.txt` when it is comes out in 2025, without needing to issue a bugfix or greater release to do so.  I can't provide the data manager with a hash to a file that doesn't exist yet.

sunpy/coordinates/frames.py
ayshih: > So what would be our update plan? Change the default model name in a new feature release of sunpy?

Yeah

> Given this I would think the best idea would be to have a way of first checking if we have included the file and then trying to fetch it from online if not.

Now done, and I've bundled the IGRF-13 coefficients, so no downloading is needed for the default model

sunpy/coordinates/frames.py
Cadair: I think the `sunpy.data.cache` part of "data manager" could be used, not the decorator with the hashes.

wtbarnes:
Discussed at dev meeting that this PR is a priority for v5.1 given the discussion at the DASH meeting.

sunpy/coordinates/frames.py
ayshih: Using `cache` now

sunpy/io/special/asdf/resources/manifests/sunpy-1.0.0.yaml
```diff
@@ -44,3 +44,12 @@ tags:
 
   - tag_uri: "tag:sunpy.org:sunpy/coordinates/frames/geocentricearthequatorial-1.0.0"
     schema_uri: "asdf://sunpy.org/sunpy/schemas/geocentricearthequatorial-1.0.0"
+
```
Cadair, line 47: @braingram When is good practice to make a new manifest vs adding stuff to an existing one?

Cadair (commented):
All looks sensible to me from a quick phone review.

sunpy/io/special/asdf/resources/manifests/sunpy-1.0.0.yaml
braingram: tldr; I'd suggest adding a new manifest for these changes
Adding a new manifest (and using it to create a new extension `sunpy-1.10`) has the benefit of informing users of old software that a new version (that supports these new features) is available.
This can be seen by generating 2 different files, one with a new manifest and one with an updated manifest (both containing a `SolarMagnetic` frame in this case). The files differ only in the extension metadata in the asdf header. For the updated manifest the file contains
```yaml
#ASDF 1.0.0
#ASDF_STANDARD 1.5.0
%YAML 1.1
%TAG ! tag:stsci.edu:asdf/
--- !core/asdf-1.1.0
asdf_library: !core/software-1.0.0 {author: The ASDF Developers, homepage: 'http://github.com/asdf-format/asdf',
  name: asdf, version: 3.0.1}
history:
  extensions:
  - !core/extension_metadata-1.0.0
    extension_class: asdf.extension._manifest.ManifestExtension
    extension_uri: asdf://asdf-format.org/core/extensions/core-1.5.0
    software: !core/software-1.0.0 {name: asdf, version: 3.0.1}
  - !core/extension_metadata-1.0.0
    extension_class: asdf.extension._manifest.ManifestExtension
    extension_uri: asdf://sunpy.org/extensions/sunpy-1.0.0
    software: !core/software-1.0.0 {name: sunpy, version: 5.1.0}
c: !<tag:sunpy.org:sunpy/coordinates/frames/solarmagnetic-1.0.0>
  frame_attributes: {magnetic_model: igrf13}
...
```
Using a new manifest, the file differs only by the extension_uri for the new extension:
```yaml
    extension_uri: asdf://sunpy.org/extensions/sunpy-1.1.0
```
Opening either of these files does not work with sunpy main (as the converters are not supported by the existing `1.0.0` extension). However, the warnings that are reported differ for the two files. For the updated manifest, asdf warns:
```
AsdfConversionWarning: tag:sunpy.org:sunpy/coordinates/frames/solarmagnetic-1.0.0 is not recognized, converting to raw Python data structure
```
For the new manifest option, the file gives 2 warnings:
```
AsdfConversionWarning: tag:sunpy.org:sunpy/coordinates/frames/solarmagnetic-1.0.0 is not recognized, converting to raw Python data structure
AsdfWarning: File 'new_manifest.asdf' was created with extension URI 'asdf://sunpy.org/extensions/sunpy-1.1.0' (from package sunpy==5.1.0), which is not currently installed
```
Note that the warning only seen for the new manifest indicates that something (in this case the `sunpy-1.1.0` extension) is not installed. Additionally, if the files are opened with `strict_extension_check` enabled only the one created with the new manifest will produce an error.
Let me know if more details would be helpful.

wtbarnes (changes requested):
This all looks good to me, but the functionality for using alternate magnetic models seems to be broken.

sunpy/coordinates/tests/test_transformations.py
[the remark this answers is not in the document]
ayshih: Yeah, I never added a set of unit tests.  Now done.

sunpy/coordinates/frames.py
```diff
@@ -847,3 +858,186 @@ class GeocentricEarthEquatorial(SunPyBaseCoordinateFrame):
...
+
+        # First look if the file is bundled in package
+        local_file = os.path.join(os.path.dirname(__file__), "data",
+                                  f"{self.magnetic_model}coeffs.txt")
+        if os.path.exists(local_file):
+            return local_file
+
+        # Otherwise download the file and cache it
+        return cache.download("https://www.ngdc.noaa.gov/IAGA/vmod/coeffs/"
+                              f"{self.magnetic_model}coeffs.txt", expiry=None)
```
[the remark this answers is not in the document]
ayshih: I've switched back to using Astropy's `download_file()` instead of our `cache.download()`.  I don't like the expiry behavior of our cache for these IGRF coefficient files

docs/whatsnew/5.1.rst
```diff
@@ -34,3 +34,9 @@ In anticipation of the upcoming `"Great North American Eclipse" <https://en.wiki
 The output can be used to determine the start/end times of partial eclipse and of totality.
 
 .. minigallery:: sunpy.coordinates.sun.eclipse_amount
+
+Coordinate frames related to the Earth's magnetic dipole
+========================================================
+Three new coordinate frames are available: `~sunpy.coordinates.Geomagnetic` (MAG), `~sunpy.coordinates.SolarMagnetic` (SM), and `~sunpy.coordinates.GeocentricSolarMagnetospheric` (GSM).
+These frames depend on the orientation of the Earth's magnetic dipole, as defined by the `International Geomagnetic Reference Field (IGRF) <https://www.ngdc.noaa.gov/IAGA/vmod/igrf.html>`__ model.
+The latest model at the time of this release (IGRF-13) is used by default, but an older model can be selected if desired.
```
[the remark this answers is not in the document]
ayshih: Sure, added

sunpy/coordinates/frames.py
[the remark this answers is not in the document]
ayshih: I'll push back on switching to using Astropy's `get_pkg_data_filename()`.  We're not really using it as intended, including depending on catching the error it raises because it won't find the file on Astropy's data server.  I think the direct code is more straightforward and more readable/maintainable.

I see only one place, at least in core, where we use `get_pkg_data_filename()`, which is in `get_test_filepath()`.  Its usage there arguably makes little sense as well.

wtbarnes (approved):

sunpy/coordinates/frames.py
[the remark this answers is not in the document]
wtbarnes: Fair enough. I don't like using `dirname`, but using `get_pkg_data_filename` is objectively worse here.

nabobalis (approved):

sunpy/io/special/asdf/entry_points.py
```diff
@@ -39,7 +39,7 @@ def get_extensions():
     sunpy_converters = [GenericMapConverter()] + SUNPY_FRAME_CONVERTERS
 
     return [
-        ManifestExtension.from_uri("asdf://sunpy.org/sunpy/manifests/sunpy-1.0.0",
+        ManifestExtension.from_uri("asdf://sunpy.org/sunpy/manifests/sunpy-1.1.0",
```
braingram, line 42: You likely want to add another `ManifestExtension` here (one for `1.0.0` and one for `1.1.0`). The order is also import (newest first so `1.1.0` then `1.0.0`) as asdf will search registered extensions in order.

As this PR is currently written (where `1.0.0` is not installed) a user that opens a file that was created with the `1.0.0` extension will encounter a warning:
```
AsdfWarning: File '...' was created with extension URI 'asdf://sunpy.org/extensions/sunpy-1.0.0' (from package sunpy==5.1.0), which is not currently installed
```

Let me know if more details are helpful. It is fine to register the same converters for both as asdf will filter out converters that don't support any tags in the manifest (as will be the case for `SolarMagnetic` etc and the `1.0.0` extension/manifest).

ayshih: Thanks for pointing this out!  Apparently the one old ASDF file that I happened to try is *so* old that it doesn't trigger this warning, but with more testing I see the behavior you are talking about.  I've made the change as I believe you recommended: 65dcf90

PR #7241, opened 2023-10-17
Remove deprecated sunpy.io.jp2

dstansby:


nabobalis:
precommit fail and it seems a file conflict now with doc/conf.py which I assume git would have been able to merge.

PR #7243, opened 2023-10-18
Let codepsell autofix stuff

dstansby:
This sets `codespell` to fix stuff in place. I think the true positive rate is much > the false positive rate, so much nicer to be able to do `pre-commit.ci autofix` on a PR and have it apply the `codespell` fixes for you, and in the rare case it gets something wrong manually add an exception in the config.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,6 +49,6 @@ repos:
     rev: v2.2.6
     hooks:
       - id: codespell
-        args: ['--config setup.cfg']
+        args: ['--config setup.cfg --write-changes']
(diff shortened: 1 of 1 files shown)


ayshih:
If you're going to do this, move the `git diff` hook to be after `codespell`.

nabobalis (approved):

nabobalis:
I agree with @dstansby, only updates to codespell should make any new changes to the codebase and we have the bot do that for us. 

Any new content should be fixed and hopefully we would spot any typos in a code review.

Cadair (approved):

PR #7245, opened 2023-10-19
Parse more `wavelnth` comments for units

wtbarnes:
This adds yet another possible case when extracting the wavelength unit from the key comment on `WAVELNTH` in `extract_waveunit`. The inspiration for this is #7220, specifically this comment: https://github.com/sunpy/sunpy/pull/7220#discussion_r1364771658


dstansby (changes requested):
Could do with with a quick test adding for the new pattern?

wtbarnes:
> Could do with with a quick test adding for the new pattern?

I agree. #7220 provides a test FITS header that is a good test case for this. For convenience, could we merge this and then I'll add a test for this extra case to #7220?

dstansby (approved):

nabobalis (approved):

Issue #7252, opened 2023-10-23
Allow `clip_interval` and `vmin/vmax` be passed to MapSequence plotting


hayesla:
### Describe the feature

Currently for MapSequence.plot(), you can pass `cmap` and `norm` for plotting a sequence following #5889 . 

However, it may be nice to also be able to pass `clip_interval` and `vmin/vmax` explicitly, such that it follows the same from `Map.plot()`

For example, currently if you do:

``` python
import sunpy.map
import sunpy.data.sample
from astropy.visualization import ImageNormalize

m_seq = sunpy.map.Map([
    sunpy.data.sample.AIA_193_CUTOUT01_IMAGE,
    sunpy.data.sample.AIA_193_CUTOUT02_IMAGE,
], sequence=True)

m_seq.plot(clip_interval=[5, 99.9]*u.percent)
```
the clipping is only applied to the first image of the sequence. This is the same for passing`vmin/vmax`,
### Proposed solution

Implement it in `MapSequence.plot()`

ryuusama09:
@hayesla  , can I get the context of why this feature is requested . It would be easier for me to interpret all these terms !. Thank you

nabobalis:
@ryuusama09, there is an open pull request for this issue, #7253

I would suggest looking at another issue.

dpshelio:
Did #7253 solve this issue?

nabobalis:
Probably

PR #7253, opened 2023-10-23
Adding ability to pass `clip_interval` and `vmin/vmax` to MapSequence.plot()

hayesla:
Addresses #7252

### DIFF
--- a/sunpy/map/tests/test_mapsequence.py
+++ b/sunpy/map/tests/test_mapsequence.py
@@ -235,5 +235,6 @@ def test_save(aia171_test_map, hmi_test_map, tmp_path):
 @figure_test
 def test_map_sequence_plot_clip_interval(aia171_test_map):
     seq = sunpy.map.Map([aia171_test_map, aia171_test_map, aia171_test_map], sequence=True)
-    animation = seq.plot(cmap='Greys', clip_interval=(1,99.5)*u.percent)
+    # We want to blow the image out to make sure it is clipped on the test data
+    animation = seq.plot(clip_interval=(5,75)*u.percent)
     animation._step()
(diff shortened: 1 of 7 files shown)


wtbarnes:
This is great, though the amount of code we need to repeat both here and in `GenericMap.plot` is worrying. I don't know of a good way to avoid that though without an overhaul of how we handle visualization.

This should probably get a figure test as well.

nabobalis:
> This is great, though the amount of code we need to repeat both here and in `GenericMap.plot` is worrying. I don't know of a good way to avoid that though without an overhaul of how we handle visualization.
> 
> This should probably get a figure test as well.

Have my terrible idea.

sunpy/map/tests/test_mapsequence.py
```diff
@@ -230,3 +230,10 @@ def test_save(aia171_test_map, hmi_test_map, tmp_path):
     for k in seq.maps[1].meta:
         assert test_seq.maps[1].meta[k] == seq.maps[1].meta[k]
     assert_quantity_allclose(test_seq.maps[1].data, seq.maps[1].data)
+
+
+@figure_test
+def test_map_sequence_plot_clip_interval(aia171_test_map):
+    seq = sunpy.map.Map([aia171_test_map, aia171_test_map, aia171_test_map], sequence=True)
+    animation = seq.plot(cmap='Greys', clip_interval=(1,99.5)*u.percent)
+    animation._step()
```
nabobalis, line 239: Need to check this works and add the hash

nabobalis (approved):

wtbarnes (approved):
We should probably get another review from someone who didn't write most of the code.

Issue #7249, opened 2023-10-22
`Cache.download()` deletes an expired file before knowing that it can download a new version of the file


ayshih:
### Provide a general description of the issue or problem.

The `download()` method of our `Cache` class deletes an expired file:

https://github.com/sunpy/sunpy/blob/7821330e32384e3d1ee66a411738935364e1ff8b/sunpy/data/data_manager/cache.py#L76-L79

before attempting to download a new version of the file:

https://github.com/sunpy/sunpy/blob/7821330e32384e3d1ee66a411738935364e1ff8b/sunpy/data/data_manager/cache.py#L84

That means that if the download does not succeed for whatever reason, the user now has no copy of the file.  Such a harsh application of expiration would be appropriate if we were dealing with time bombs, but for virtually all use cases, an expired file is just as usable as a new version.

We should not delete the expired file until after successfully downloading a new version of the file.  If the downloading fails, we retain the expired file, and we can emit a warning to the user that they will be working with a potentially stale file.

samish55:
You can check the open PR #7265.

NucleonGodX:
the PR of this issue is drafted for over 2.5months now can I take up this issue?

nabobalis:
That PR was waiting for unit tests to be added. IF you want to help @samish55 finish off the PR, you should contact them first.

NucleonGodX:
> That PR was waiting for unit tests to be added. IF you want to help @samish55 finish off the PR, you should contact them first.

alrighty makes sense, he don't seem to have any particular contact on github, but since its just unit tests left I think @samish55  might address them.

nabobalis:
Sure, there is an open PR which addresses most of this but it needs to be completed.

alasdairwilson:
Closed as completed via #7935

PR #7265, opened 2023-10-26
Prevent `Cache.download()` deleting an expired file before knowing that it can download a new version of the file

samish55:
## PR Description

I've attempted to handle issue #7249. The expired file won't be deleted until the file is successfully downloaded and if the download fails(for some reason) an error warning is emitted about working with an expired file in cache.

Fixes https://github.com/sunpy/sunpy/issues/7249

### DIFF
--- a/sunpy/data/data_manager/cache.py
+++ b/sunpy/data/data_manager/cache.py
@@ -36,14 +36,15 @@ class Cache:
         self._cache_dir = Path(cache_dir)
         self._expiry = expiry if expiry is None else TimeDelta(expiry)
 
-    def download(self, url, namespace='', redownload=False):
+    def download(self, urls, namespace='', redownload=False):
         """
         Downloads the files from the urls.
 
         Parameters
         ----------
-        urls : `str`
-            A single url.
+        urls : `list` of path-like or one path-like
+            A list of urls or a single url.
+            The list is for urls of the same file but from different sources.
         namespace : `str`, optional
             A namespace to be used for the file name.
             Defaults to an empty string.
@@ -56,13 +57,18 @@ class Cache:
         `pathlib.Path`
             Path to the downloaded file.
         """
+        if isinstance(urls, str | Path):
+            urls = [urls]
         # Logic flow
         # 1. If redownload: Download, update cache and return file path
         # 2. If not redownload: Check cache,
         #    i. If present in cache:
         #        - If cache expired, download and add to cache and remove entry from cache
         #        - If cache not expired, return path
-        present = self._get_by_url(url)
+        for url in urls:
+            present = self._get_by_url(url)
+            if present:
+                break
         if present and not redownload and not self._has_expired(present):
             return Path(present['file_path'])
         try:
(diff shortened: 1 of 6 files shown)


sunpy/data/data_manager/cache.py
```diff
@@ -73,23 +73,33 @@ def download(self, urls, namespace='', redownload=False):
...
+            if(details and redownload):
+                # if file is in cache and it has to be redownloaded
                 # then remove the file and delete the details from the storage
                 os.remove(details['file_path'])
                 self._storage.delete_by_key('url', details['url'])
-            else:
-                return Path(details['file_path'])
 
-        file_path, file_hash, url = self._download_and_hash(urls, namespace)
+            file_path, file_hash, url = self._download_and_hash(urls, namespace)
```
Cadair, line 85: I think to achieve the aims of the issue this line needs to be first under the try block. You need to do the download first, and then if the download fails you need a new code path which returns the cached version. I think from looking at your code at the moment you can still end up in a situation where you delete the file and the download fails.

Cadair (commented):
Thanks a lot for the PR @samish55 

I think you need to refactor the logic a little bit to be sure the old file doesn't get deleted.

sunpy/data/data_manager/cache.py
samish55: I've made a few changes you can review them.
I think I've made a lot of uneccessory commits for a single PR, you can ask if you want me to make a new PR and close this.

nabobalis: We can squash merge, that isn't a problem.

nabobalis:
Thanks for the code. I think ideally we need a unit test to make sure this behavior is working.

It is going to be tricky to do as you will need to monkey patch several calls (I think) to fake the download and make it fail.

nabobalis (approved):

sunpy/data/data_manager/cache.py
```diff
@@ -40,56 +39,60 @@ def download(self, urls, namespace='', redownload=False):
         """
         Downloads the files from the urls.
 
-        The overall flow of this function is:
```
nabobalis, line 43: This is now just in the code as a comment instead of in both places.

sunpy/data/data_manager/cache.py
```diff
@@ -40,56 +39,60 @@ def download(self, urls, namespace='', redownload=False):
...
+        # We want to just return the file
+        if cache_details and not redownload and not self._has_expired(cache_details):
+            return Path(cache_details['file_path'])
+        try:
+            file_path, file_hash, url = self._download_and_hash(urls, namespace)
+            # We explicitly check for redownload or expiry:
+            # This isn't needed as it should be caught bt the above if statement
+            # but this is more explicit
+            if cache_details and (redownload or self._has_expired(cache_details)):
+                self._storage.delete_by_key('url', cache_details['url'])
```
nabobalis, line 80: I think there is logic problem here:

1. We do not remove any old files from disk
2. If the file changes on disk, we could remove it here but then it will break logic in the manager as we need to hash the file a second time.

Sauravroy34: how about adding
 ``` 
os.remove(cache_details['file_path'])
```
removing the file and then deleting it?

Comment (nabobalis) [reply sunpy/data/data_manager/cache.py:80]: Will that work?

Comment (Sauravroy34) [reply sunpy/data/data_manager/cache.py:80]: It should since we would remove the file so no we would not have any old file remaining

Comment (nabobalis) [reply sunpy/data/data_manager/cache.py:80]: But what about the need to hash and update the database again?

Comment (Sauravroy34) [reply sunpy/data/data_manager/cache.py:80]: how about we check the file hash in cache and downloaded and only remove if its not same 
```
    try:
        file_path, file_hash, url = self._download_and_hash(urls, namespace)

        if cache_details and not redownload:

            if cache_details['file_hash'] == file_hash:
                return Path(cache_details['file_path'])

            os.remove(cache_details['file_path'])
            self._storage.delete_by_key('url', cache_details['url'])
```

nabobalis: Maybe but we first need a unit test to replicate this and work out the best solution from there.

nabobalis:
Closed in favour of https://github.com/sunpy/sunpy/pull/7935

Thank you @samish55 for taking this complex issue on in the first place.

PR #7267, opened 2023-10-26
Seperate docs into docs and docs-gallery

nabobalis:
Following the problems of installing mplcario, I have separate docs into docs and docs-gallery. Only docs is installed for the dev extra.

PR #7272, opened 2023-10-28
RTD needs docs-gallery

ayshih:
Fixes an oversight in #7267, since of course RTD needs `docs-gallery` dependencies

Apparently RTD still declares a build passing even if a bunch of gallery examples error out


nabobalis (approved):
Woops and thanks

PR #7307, opened 2023-11-18
Update `JSOCClient.fetch` to accomodate breaking changes in `drms`

nabobalis:
Also pins `drms` to `>=0.7.1`.

### DIFF
--- a/changelog/7307.trivial.rst
+++ b/changelog/7307.trivial.rst
@@ -1,6 +1,6 @@
 :meth:`sunpy.net.jsoc.JSOCClient.fetch` called `drms` API that passed a ``progress`` keyword which added extra print statements to the console.
 This has been removed in `drms` 0.7.0, which had breaking API changes within this release.
-As a result, we increased the minimum required version of `drms` to 0.7.0.
+As a result, we increased the minimum required version of `drms` to 0.7.1.
 
 This specifically refers to the following information that was printed to the console by default:
 
--- a/changelog/7307.trivial.rst
+++ b/changelog/7307.trivial.rst
@@ -1,0 +1,3 @@
+Progress from :meth:`sunpy.net.jsoc.JSOCClient.fetch` is now reported through the logger.
+Pinned the ``drms`` requirement to ``>= 0.7`` to account for this breaking API change in ``drms``.
+Note that this change supersedes that in `PR #7308 <https://github.com/sunpy/sunpy/pull/7308>`_.
--- a/changelog/7307.trivial.rst
+++ b/changelog/7307.trivial.rst
@@ -2,10 +2,22 @@
 This has been removed in `drms` 0.7.0, which had breaking API changes within this release.
 As a result, we increased the minimum required version of `drms` to 0.7.0.
 
-This specifically refers to the following information that was printed to the console:
+This specifically refers to the following information that was printed to the console by default:
 
 ``"Export request pending. [id=X, status=X]"``
 ``"Waiting for X seconds..."``
 ``"Request not found on server, X retries left."``
 
 These were handled by `drms` and are now logging messages.
+
+If you want to silence these messages, you can set the logging level to ``WARNING`` or higher.
+
+.. code-block:: python
+
+    import logging
+    drms_logger = logging.getLogger("drms")
+    drms_logger.setLevel(logging.WARNING)
+
+    from sunpy.net import fido, attrs
+
+Note, you have to do it before you import ``fido``.
--- a/changelog/7307.trivial.rst
+++ b/changelog/7307.trivial.rst
@@ -1,3 +1,11 @@
-Progress from :meth:`sunpy.net.jsoc.JSOCClient.fetch` is now reported through the logger.
-Pinned the ``drms`` requirement to ``>= 0.7`` to account for this breaking API change in ``drms``.
-Note that this change supersedes that in `PR #7308 <https://github.com/sunpy/sunpy/pull/7308>`_.
+:meth:`sunpy.net.jsoc.JSOCClient.fetch` called `drms` API that passed a ``progress`` keyword which added extra print statements to the console.
+This has been removed in `drms` 0.7.0, which had breaking API changes within this release.
+As a result, we increased the minimum required version of `drms` to 0.7.0.
+
+This specifically refers to the following information that was printed to the console:
+
+``"Export request pending. [id=X, status=X]"``
+``"Waiting for X seconds..."``
+``"Request not found on server, X retries left."``
+
+These were handled by `drms` and are now logging messages.
--- a/setup.cfg
+++ b/setup.cfg
@@ -70,7 +70,7 @@ map =
   scipy>=1.7.0,!=1.10.0
 net =
   beautifulsoup4>=4.8.0
-  drms>=0.6.1,<0.7.0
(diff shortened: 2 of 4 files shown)


nabobalis:
Need to double check the online build passes first before merge.

nabobalis:
```
FAILED ../../.tox/py39-online/lib/python3.9/site-packages/sunpy/database/tests/test_tables.py::test_entry_from_qr_block_with_missing_physobs
FAILED ../../.tox/py39-online/lib/python3.9/site-packages/sunpy/net/tests/test_fido.py::test_path
FAILED ../../.tox/py39-online/lib/python3.9/site-packages/sunpy/net/vso/tests/test_vso.py::test_path
FAILED ../../docs/tutorial/acquiring_data/index.rst::index.rst
```

These looks like VSO <-> JSOC issues

sunpy/net/jsoc/jsoc.py
```diff
@@ -438,7 +438,10 @@ def fetch(self, jsoc_response, path=None, progress=True, overwrite=False,
         time.sleep(sleep/2.)
 
         for response in responses:
-            response.wait(verbose=progress)
+            if Version(drms.__version__) >= Version('0.7.0'):
```
Cadair, line 441: So now we don't get messages with the export status any more?

nabobalis: You get via logging.info, no keyword anymore.

dstansby (changes requested):
This should have a changelog, explaining the difference in logging/progress bar behaviour with an upgrade to `drms` 0.7 (since users of sunpy won't neccessarily know about drms, we should tell them if the user experience of downloading from JSOC is changed)

sunpy/net/jsoc/jsoc.py
wtbarnes: Seeems sensible to me and is consistent with how we control output in the rest of the package.

changelog/7307.trivial.rst
```diff
@@ -0,0 +1,3 @@
+Progress from :meth:`sunpy.net.jsoc.JSOCClient.fetch` is now reported through the logger.
```
dstansby, line 1: Does this mean no progress bar any more? Does the progress show by default? How do I turn on and off the progress now compared to before? (are all questions that I would probably ask as a user looking at this change)

dstansby, line 1: Does this mean no progress bar any more? Does the progress show by default? How do I turn on and off the progress now compared to before? (are all questions that I would probably ask as a user looking at this change)

wtbarnes: All fair questions. Doing some manual testing, I find that the only way I'm able to control the verbosity of the progress bar is still through the `progress`keyword argument to `JSOCClient.fetch`. For example, the following code snippet still produces a progress bar, even though I've set the log level to "warn" (i.e. exclude "info"),

```python
import sunpy
sunpy.log.setLevel('WARN')

import astropy.units as u
from sunpy.net import jsoc
client = jsoc.JSOCClient()
response = client.search(a.Time('2014-01-01T00:00:00', '2014-01-01T00:00:10'),
                         a.jsoc.Series('aia.lev1_euv_12s'),
                         a.jsoc.PrimeKey('WAVELNTH','171'))
file = client.fetch(response, progress=True)
```
Given this, I'm confused as to what has actually changed on the `drms` side. Maybe I'm misunderstanding something here.

wtbarnes: Resolving this one as the same comment seems to have been posted twice.

wtbarnes: If we cannot figure this out in the next few days, we can close this and just carry on with the `<0.7` pin that we already have. I don't really understand the point of `drms` v0.7 anyway...

nabobalis: Drms with the progress keyword should print:

```
            if verbose:
                print(f"Downloading file {int(i + 1)} of {int(ndata)}...")
                print(f"    record: {di.record}")
                print(f"  filename: {di.filename}")
                print(f"  -> {os.path.relpath(fpath_new)}")

```
These are now logging.info

nabobalis:
I updated the changelog to say what messages were affected. 

Triggering them isn't common, and it won't happen if the request is already complete.

setup.cfg
```diff
@@ -69,7 +69,7 @@ map =
   scipy>=1.7.0,!=1.10.0
 net =
   beautifulsoup4>=4.8.0
-  drms>=0.6.1,<0.7.0
+  drms>=0.7.0
```
nabobalis, line 72:
```suggestion
  drms>=0.7.1
```

sunpy/net/jsoc/jsoc.py
```diff
@@ -547,8 +547,8 @@ def get_request(self, requests, path=None, overwrite=False, progress=True,
 
         if urls:
             if progress:
-                print_message = "{0} URLs found for download. Full request totalling {1}MB"
-                print(print_message.format(len(urls), request._d['size']))
+                log_message = "{0} URLs found for download. Full request totalling {1}MB"
+                log.info(log_message.format(len(urls), request._d['size']))
```
nabobalis, line 551: I changed this as we should really just be dumping to the logger instead of a print. But maybe others like it this way?

wtbarnes, line 551:
```suggestion
                log.info(f"{len(urls)} URLs found for download. Full request totaling {request._d['size']}MB")
```

wtbarnes: I'm in favor of all of these messages being handled by the logger as it gives the user a standardized way to control the output.

wtbarnes (approved):

changelog/7307.trivial.rst
```diff
@@ -0,0 +1,11 @@
...
+This has been removed in `drms` 0.7.0, which had breaking API changes within this release.
+As a result, we increased the minimum required version of `drms` to 0.7.0.
+
+This specifically refers to the following information that was printed to the console:
+
+``"Export request pending. [id=X, status=X]"``
+``"Waiting for X seconds..."``
+``"Request not found on server, X retries left."``
+
+These were handled by `drms` and are now logging messages.
```
dstansby, line 11: It's not clear if these messages are now shown by default or not? And if not, could you add a quick explanation (with code) on how to turn them back on? (or say that it's not possible to see them any more if it isn't possible)

nabobalis: Added more text and code example. Hopefully that is good enough.

dstansby: 👍 liking the code example

changelog/7307.trivial.rst
```diff
@@ -0,0 +1,23 @@
+:meth:`sunpy.net.jsoc.JSOCClient.fetch` called `drms` API that passed a ``progress`` keyword which added extra print statements to the console.
+This has been removed in `drms` 0.7.0, which had breaking API changes within this release.
+As a result, we increased the minimum required version of `drms` to 0.7.0.
```
dstansby, line 3:
```suggestion
As a result, we increased the minimum required version of `drms` to 0.7.1.
```

dstansby (approved):
LGTM, modulo fixing the verison number in the changelog.

PR #7314, opened 2023-11-22
Extremely minor change to plotting lines on a map example

alasdairwilson:
There was some confusion in the chat by a user the other day over this example as swapping the arrays of values for a python list, rather than a ndarray, in this example breaks the plot command due to the conversion to degrees not being able to operate on a list.

Adding the parenthesis before the conversion to degree should allow users to more easily tweek this example for their own needs.


ayshih:
Already fixed (in a different manner) in #7302

PR #7319, opened 2023-11-23
`get_horizons_coord()` no longer uses `astroquery`

### LINKED ISSUE #7080: Re-implement `get_horizons_coord()` to not use `astroquery`
The function `sunpy.coordinates.get_horizons_coord()` internally uses `astroquery` to query JPL Horizons for the location of a wide range of solar-system bodies.  We don't use rely on any other functionality of `astroquery` (not counting the example gallery), nor do we even utilize its support for other parts of the JPL Horizons API.  We can re-implement `get_horizons_coord()` to directly query JPL Horizons in exactly the manner needed, and remove the (optional) dependency on `astroquery`.

x-ref #7066

ayshih:
Fixes #7080

This PR re-implements `get_horizons_coord()` so that the function no longer uses `astroquery`.  Consequently, `astroquery` is no longer a library dependency, although it is still a dependency for building the gallery examples because we use `astroquery` to get star positions in one example.  This new implementation actually has two enhancements: long object names are no longer truncated, and it uses a different HORIZONS API which is in principle more permissive of a long list of times.

To do:

- [ ] Improve error handling


sunpy/coordinates/ephemeris.py
```diff
@@ -1,8 +1,11 @@
 """
 Ephemeris calculations using SunPy coordinate frames
 """
+import re
+import tempfile
+
 import numpy as np
-from packaging import version
+import requests
```
Cadair, line 8: It seems we don't have an explicit dependancy on requests at the moment? (I would be shocked if we don't have a transitive one.)

We could either add it or use urllib.

ayshih: It looks like `requests` is implicit dependency through `zeep` (and also `sphinx`).  I'd prefer to add `requests` as an explicit dependency because we're going to have it anyway and doing a POST via `urllib` is bit more cumbersome.

setup.cfg
```diff
@@ -51,6 +51,8 @@ install_requires =
 asdf =
   asdf-astropy>=0.1.1
   asdf>=2.8.0
+coordinates =
+  requests
```
Cadair, line 55: Do we want to make this a core dep, because at the moment I consider coordinates as one of the main things (if not *the* thing) which you get without installing any extras.

ayshih: Done

sunpy/coordinates/ephemeris.py
```diff
@@ -282,46 +287,97 @@ def get_horizons_coord(body, time='now', id_type=None, *, include_velocity=False
...
         epochs = array_time.tdb.jd.tolist()  # Time must be provided in JD TDB
+        args['TLIST'] = '\n'.join(str(epoch) for epoch in epochs)
+        args['TLIST_TYPE'] = 'JD'
+
+    with tempfile.TemporaryFile('w+t') as query:
+        contents = "!$$SOF\n" + '\n'.join(f"{k}={v}" for k, v in args.items())
+        log.debug(f"File-based JPL HORIZONS query:\n{contents}")
+
+        query.write(contents)
+        query.seek(0)
```
samaloney, line 336: Is it necessary to write this to disk, or put another way is there a benefit? I'm pretty sure you could pass `contents` directly to request as post data, might have to wrap in StringIO or something first.

sunpy/coordinates/ephemeris.py
```diff
@@ -282,46 +287,97 @@ def get_horizons_coord(body, time='now', id_type=None, *, include_velocity=False
...
-    from astroquery.jplhorizons import Horizons
-
-    if id_type is None and version.parse(astroquery.__version__) < version.parse('0.4.4'):
-        # For older versions of astroquery retain default behaviour of this function
-        # if id_type isn't manually specified.
-        id_type = 'majorbody'
+    args = {
+        'EPHEM_TYPE': 'VECTORS',
+        'OUT_UNITS': 'AU-D',
+        'CENTER': '500@10',
```
samaloney, line 296: Maybe add comments to say what AU-D and 500@10 mean?

samaloney (commented):
Looks good

sunpy/coordinates/ephemeris.py
ayshih: Added

sunpy/coordinates/ephemeris.py
ayshih: Heh, whoops, indeed you *can* simply provide a string.  The `requests` documentation isn't particularly clear about all the different forms of data it can accept...

samaloney (approved):
LGTM

sunpy-dev-env.yml
```diff
@@ -53,6 +53,7 @@ dependencies:
   - tox-conda
 
   # Documentation
+  - astroquery
```
nabobalis, line 56: What example still uses it?

nabobalis (approved):

sunpy-dev-env.yml
ayshih: [Identifying stars in a STEREO/SECCHI COR2 coronagraph image](https://docs.sunpy.org/en/stable/generated/gallery/units_and_coordinates/STEREO_SECCHI_starfield.html)

nabobalis:
Codecov left a bunch of comments but the patches passed so I am confused but don't care enough to dig down as to what happened.

PR #7329, opened 2023-12-04
Add Clarification on setting Environment Variable for JSOC Email

### LINKED ISSUE #7327: Add clarifications to examples that use `os.environ["JSOC_EMAIL"]`
We have three examples that have `jsoc_email = os.environ["JSOC_EMAIL"]` (where the environment variable is defined in our `conf.py` for Sphinx).  We tell users that they should use their own (registered) email address instead, but we're not clear about how the line should actually be changed.  One user on Element replaced only `"JSOC_EMAIL"` instead of replacing `os.environ["JSOC_EMAIL"]`.  Assuming that we continue to use this (exposed) environment-variable approach, we should be more clear in the associated description.

The three examples are:
https://github.com/sunpy/sunpy/blob/42087228ff25d7e3fcd4d9c86c8af12c311d5b7d/examples/acquiring_data/downloading_cutouts.py#L38-L44
https://github.com/sunpy/sunpy/blob/42087228ff25d7e3fcd4d9c86c8af12c311d5b7d/examples/acquiring_data/fido_metadata_queries.py#L24-L26
https://github.com/sunpy/sunpy/blob/42087228ff25d7e3fcd4d9c86c8af12c311d5b7d/examples/acquiring_data/querying_and_loading_SHARP_data.py#L23-L26

abastola0:
Closes #7327

### DIFF
--- a/changelog/7329.doc.rst
+++ b/changelog/7329.doc.rst
@@ -1,0 +1,1 @@
+Added Clarification on setting JSOC Email.
--- a/examples/acquiring_data/downloading_cutouts.py
+++ b/examples/acquiring_data/downloading_cutouts.py
@@ -35,11 +35,10 @@ top_right = SkyCoord(150*u.arcsec, 375*u.arcsec, obstime=start_time, observer="e
 cutout = a.jsoc.Cutout(bottom_left, top_right=top_right, tracking=True)
 
 #####################################################
-# Exporting data from the JSOC requires registering your
-# email first. Please replace this with your email
-# address once you have registered.
-# See `this page <http://jsoc.stanford.edu/ajax/register_email.html>`_
-# for more details.
+# Exporting data from the JSOC requires registering your email first.
+# Please replace this with your email address once you have registered
+# like so: jsoc_email = "your_email@example.com"
+# See `this page <http://jsoc.stanford.edu/ajax/register_email.html>`__ for more details.
 
 jsoc_email = os.environ["JSOC_EMAIL"]
 
--- a/examples/acquiring_data/fido_metadata_queries.py
+++ b/examples/acquiring_data/fido_metadata_queries.py
@@ -21,8 +21,11 @@ from sunpy.net import attrs as a
 
 timerange = a.Time('2010/8/1 03:40', '2010/8/1 3:40:10')
 
-# Note that JSOC needs an email address to export the files, if you want to run
-# this, you must supply your own email here.
+# Exporting data from the JSOC requires registering your email first.
+# Please replace this with your email address once you have registered
+# like so: jsoc_email = "your_email@example.com"
+# See `this page <http://jsoc.stanford.edu/ajax/register_email.html>`__ for more details.
+
 jsoc_email = os.environ["JSOC_EMAIL"]
 
 results = Fido.search(timerange,
--- a/examples/acquiring_data/querying_and_loading_SHARP_data.py
+++ b/examples/acquiring_data/querying_and_loading_SHARP_data.py
@@ -21,8 +21,10 @@ from sunpy.net import attrs as a
 # and make use of the search attributes in `sunpy.net.jsoc` that allow us to query the JSOC.
 #
 # Exporting data from the JSOC requires registering your email first.
-# Please replace this with your email address once you have registered.
+# Please replace this with your email address once you have registered
+# like so: jsoc_email = "your_email@example.com"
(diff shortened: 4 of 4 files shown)


docs/conf.py
```diff
@@ -269,6 +269,10 @@
 # -- Sphinx Gallery ------------------------------------------------------------
 # JSOC email os env
 # see https://github.com/sunpy/sunpy/wiki/Home:-JSOC
+# use "export JSOC_EMAIL='your_email@example.com'"
```
nabobalis, line 272: This file does not need to be changed.

examples/acquiring_data/downloading_cutouts.py
```diff
@@ -38,6 +38,9 @@
 # Exporting data from the JSOC requires registering your
 # email first. Please replace this with your email
 # address once you have registered.
+# use "export JSOC_EMAIL='your_email@example.com'"
+# in your terminal to set the environment variable.
+# NO CHANGE REQUIRED HERE!
```
nabobalis, line 43: This should not be what we tell users to change.

They should just change the line to be `jsoc_email="<their_email_address"` and not to export the variable.

docs/conf.py
abastola0: Thanks for the information @nabobalis. Let me fix it.

examples/acquiring_data/downloading_cutouts.py
```diff
@@ -41,7 +41,7 @@
 # See `this page <http://jsoc.stanford.edu/ajax/register_email.html>`_
 # for more details.
 
-jsoc_email = os.environ["JSOC_EMAIL"]
+jsoc_email = "<your email address>"
```
nabobalis, line 44: Did you push in-between commits?

abastola0: yes but i ammended all commits to my previous one with no edit.

nabobalis: I see but this change is now incorrect (the old code has to stay) and it doesn't change the other two examples you did from before.

abastola0: i undo all changes except one line which is why only one change is showing up. All other changes were removed.

nabobalis: Ah I see but this is now incorrect. We need to tell readers of the the examples that they need to replace this line with their email address and make it really clear to prevent any misunderstanding.

abastola0: what do you suggest i do next. shall i remove all changes and sync with origin and add another commit with only the line changing?

nabobalis: I am not sure I follow. We need to restore this line back to normal and add a comment above each of these lines in each example telling people what to change if they want to use their own email address.

abastola0: gotcha. I'll make these changes right now. So i can remove all the os.environ["JSOC_EMAIL"] and instead tell users to edit the example files themselves right? Also they will be replaced with placeholder string for their email addresses.

abastola0: gotcha! Thanks

abastola0: lmk if the latest commit addresses your comments.

docs/conf.py
```diff
@@ -269,6 +269,7 @@
 # -- Sphinx Gallery ------------------------------------------------------------
 # JSOC email os env
 # see https://github.com/sunpy/sunpy/wiki/Home:-JSOC
+# Do not change the email here !
```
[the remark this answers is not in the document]
abastola0: done !

alasdairwilson (dismissed):
👍

alasdairwilson (dismissed):
Can you make the language consistent between the 3 examples?

abastola0:
> Can you make the language consistent between the 3 examples?

sure

nabobalis (dismissed):

alasdairwilson (commented):
Happy with this once the changelog is done properly.

changelog/7329.doc.rst
```diff
@@ -0,0 +1 @@
+Added Clarification on setting Environment Variable for JSOC Email.
```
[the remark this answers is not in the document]
abastola0: Done. Thanks for your patience 😁

nabobalis:
Thank you @abastola0 for the PR.

PR #7334, opened 2023-12-06
Add prettier precommit

nabobalis:
It makes the json files easier to diff, that's for sure.

Thoughts?

### DIFF
--- a/.github/workflows/stale_bot.yml
+++ b/.github/workflows/stale_bot.yml
@@ -2,12 +2,12 @@ name: Stale Bot
 
 on:
   schedule:
-#            ┌───────── minute (0 - 59)
-#            │ ┌───────── hour (0 - 23)
-#            │ │ ┌───────── day of the month (1 - 31)
-#            │ │ │ ┌───────── month (1 - 12 or JAN-DEC)
-#            │ │ │ │ ┌───────── day of the week (0 - 6 or SUN-SAT)
-    - cron: '0 5 * * *' # Every day at 05:00 UTC
+    #        ┌───────── minute (0 - 59)
+    #        │ ┌───────── hour (0 - 23)
+    #        │ │ ┌───────── day of the month (1 - 31)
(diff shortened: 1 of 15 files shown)


.github/workflows/stale_bot.yml
```diff
@@ -2,12 +2,12 @@ name: Stale Bot
 
 on:
   schedule:
-#            ┌───────── minute (0 - 59)
-#            │ ┌───────── hour (0 - 23)
-#            │ │ ┌───────── day of the month (1 - 31)
-#            │ │ │ ┌───────── month (1 - 12 or JAN-DEC)
-#            │ │ │ │ ┌───────── day of the week (0 - 6 or SUN-SAT)
-    - cron: '0 5 * * *' # Every day at 05:00 UTC
+    #            ┌───────── minute (0 - 59)
```
nabobalis, line 5: I have to fix this

dstansby (changes requested):
Can you exclude the autogenerated files in `sunpy/net`? They look fine and diff well as it is, and linting them will be a pain every time we update them. Otherwise 👍 to doing this to the rest of the files

dstansby (commented):
Oh also `zenodo.json` is autogenerated, so that should be excluded too.

nabobalis:
I disagree, we create the two attrs JSON and they should either have the output adjusted to look like the prettier output or they should be linted by the pre-commit.

dstansby:
Why? They're nicely human readable as they are, what's the point in applying an arbitrary linter to them and making our lives harder by having to modify the scripts that generate the files?

nabobalis:
> Why? They're nicely human readable as they are, what's the point in applying an arbitrary linter to them and making our lives harder by having to modify the scripts that generate the files?

I think they are much nicer to read in this form than before.

wtbarnes:
I'm not going to die on this hill, but I don't really see the utility in this.

nabobalis:
In that case, I will close this and never bring it up ever again.

PR #7340, opened 2023-12-07
Removed positional argument support for timeseries peek

nabobalis:
Deprecated since 4.1


wtbarnes (approved):

nabobalis:
MacOs error is 
`E           RuntimeError: set_wakeup_fd only works in main thread of the main interpreter`

PR #7341, opened 2023-12-07
Removal of response_format keyword from VSOClient.search

nabobalis:
We created this as a workaround back in 2.1 days, in theory this was not public API but we never made that clear in anything.


dstansby (changes requested):
This should get a deprecation period? Unless there's a very good reason not to do a deprecation?

nabobalis:
> This should get a deprecation period? Unless there's a very good reason not to do a deprecation?

This was never intended to be user facing API. We introduced this when we adding a breaking change to the VSO client to allow the old behaviour but the old behaviour was only used in the hektovso client for reasons I do not remember.  

This removal changes nothing for any of users and for me that does not need a depreciation period. 

If others agree, I can add one but we should pull this out instead of going through the cycle.

wtbarnes:
> This removal changes nothing for any of users and for me that does not need a depreciation period.

You're probably right, but unfortunately, we have no way to know if that is actually true and technically, this is public-facing API.

> If others agree, I can add one but we should pull this out instead of going through the cycle.

From a pragmatic point of view, I agree with you, but departing from our deprecation policy because it is convenient for us is worrying to me.

nabobalis:
In that case, it will be easier for me to depreciate it in a new PR.

PR #7342, opened 2023-12-08
Double check a non-empty but not None value was provided to jsoc notify

nabobalis:
This catches me out when running examples sometimes, as if you do not index the os.environ dict but use .get, you get a default value. I guess the answer is to fix the example but either way, we should catch this.

### DIFF
--- a/changelog/7342.trivial.rst
+++ b/changelog/7342.trivial.rst
@@ -1,1 +1,1 @@
-`~sunpy.net.jsoc.attrs.Notify` checks for a valid input value.
+`~sunpy.net.jsoc.attrs.Notify` checks that a valid email address has been given as a value.
--- a/changelog/7342.trivial.rst
+++ b/changelog/7342.trivial.rst
@@ -1,0 +1,1 @@
+`~sunpy.net.jsoc.attrs.Notify` checks for a valid input value.
--- a/sunpy/net/jsoc/attrs.py
+++ b/sunpy/net/jsoc/attrs.py
@@ -110,11 +110,11 @@ class Notify(SimpleAttr):
 
     def __init__(self, value):
         super().__init__(value)
-        if value is None:
+        if not value:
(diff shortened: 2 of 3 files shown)


sunpy/net/jsoc/attrs.py
```diff
@@ -110,9 +110,11 @@ class Notify(SimpleAttr):
 
     def __init__(self, value):
         super().__init__(value)
+        if not value:
+            raise ValueError("Notify attribute must contain an email address")
         if value.find('@') == -1:
             raise ValueError("Notify attribute must contain an '@' symbol "
-                             "to be a valid email address")
+                             "to be a valid email address: {value}")
```
nabobalis, line 117:
```suggestion
                             "to be a valid email address")
```

sunpy/net/jsoc/attrs.py
```diff
@@ -110,9 +110,11 @@ class Notify(SimpleAttr):
 
     def __init__(self, value):
         super().__init__(value)
+        if not value:
```
dstansby, line 113:
```suggestion
        if value is None:
```
Is a bit clearer (if `value` is indeed `None` if no email is passed...

changelog/7342.trivial.rst
```diff
@@ -0,0 +1 @@
+`~sunpy.net.jsoc.attrs.Notify` checks for a valid input value.
```
dstansby, line 1:
```suggestion
`~sunpy.net.jsoc.attrs.Notify` checks that a valid email address has been given as a value.
```
Does this have any other higher-level user facing behaviour that needs mentioning? e.g. if I forget to do a JSOC Fido search with an email address, does this now raise a more helpful error?

dstansby (changes requested):
👍 , with a suggestion and question.

dstansby (approved):
Woops, that was meant to be an approval...

wtbarnes (approved):

PR #7343, opened 2023-12-11
Ran the autopep8 linter for pyopensci review

nabobalis:
A requirement for us to pass. This isn't black so I imagine there will be less problems with this adoption. 

With this `flake8 --count sunpy` is 0


ayshih:
As I've ranted before, it frustrates me how tools like autopep8/flake8 insist on spaces around every binary operator even though that's not what [PEP 8 recommends](https://peps.python.org/pep-0008/#other-recommendations).

nabobalis:
> As I've ranted about before, it frustrates me how tools like autopep8/flake8 insist on spaces around every binary operator even though that's not what [PEP 8 recommends](https://peps.python.org/pep-0008/#other-recommendations).

Yeah I agree for units.

What I could do is remove the auto linter, just lint it once and for each place we have at least a unit *, I can remove the space. I can probably do that for most operations.

dstansby:
I agree with Albert that spaces around `1 * u.Unit` look odd. Is there a way to disable that specific flake8 rule and run the rest?

alasdairwilson:
Can we run it with --ignore=E225 then?

sunpy/coordinates/tests/test_metaframes.py
```diff
@@ -88,7 +88,7 @@ def test_class_creation(indirect_fixture):
     assert base_class.__name__ in rot_class.__name__
 
     # Check that the base class is in fact the specified class
-    assert type(rot_frame.base) == base_class
+    assert isinstance(rot_frame.base, base_class)
```
nabobalis, line 91: Need to see if this is ok or if I need to revert it and add the NOQA

ayshih: I'd prefer if this change and the two similar changes in this file were reverted (and this one flagged as NOQA).  This is testing that the frame classes in the RotatedSunFrame machinery are exactly what they are supposed to be, so unexpected subclassing ought to be treated as failures.

sunpy/io/special/genx.py
```diff
@@ -125,7 +126,7 @@ def unpack_string(self):
 
     def unpack_list(self, unpack_item):
         list = []
-        while 1:
+        while True:
             x = self.unpack_uint()
             if x == 0:
                 break
```
ayshih, line 132: This appears to be a prime candidate for the walrus operator!
```python
while (x := self.unpack_uint()) != 0:
```

ayshih, line 132: This appears to be a prime candidate for the walrus operator!
```suggestion
        while (x := self.unpack_uint()) != 0:
```

sunpy/coordinates/tests/test_metaframes.py
nabobalis: Makes sense, I will fix this and add the noqa plus the rule.

sunpy/io/special/genx.py
nabobalis: Added.

sunpy/coordinates/tests/test_metaframes.py
nabobalis: Restored. autopep8 ignores noqas

sunpy/net/_attrs.py
```diff
@@ -64,7 +64,7 @@ def collides(self, other):
         # Use exact type checking here, because otherwise it collides with all
         # subclasses of itself which can have completely different search
         # meanings.
-        return type(other) is type(self)
+        return isinstance(other, type(self))
```
dstansby, line 67: Doesn't the comment above imply that this should stay as `is` and not `istinstance`?

nabobalis: Yes, I will need to fix that.

wtbarnes (approved):

alasdairwilson (approved):
I'm happy with where this is, this isn't exactly something we can't tweak in future either.

setup.cfg
```diff
@@ -237,10 +237,17 @@ filterwarnings =
...
 [flake8]
+ignore =
+    E225,
+    E226,
+    E501,
+    F401,
+    F403,
+    F811,
+    W503,
+    W504,
```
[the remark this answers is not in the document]
nabobalis: wILL DO

dstansby (approved):
Feel free to self merge once the error code descriptions are added

setup.cfg
nabobalis, line 250:
```suggestion
    E225, # missing-whitespace-around-operator
    E226, # missing-whitespace-around-arithmetic-operator
    E501, # line-too-long
    F401, # unused-import
    F403, # undefined-local-with-import-star
    F811, # redefined-while-unused
    W503, # Line break occurred before a binary operator
    W504, # Line break occurred after a binary operator
```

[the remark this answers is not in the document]
nabobalis: Added

nabobalis:
I added comments in the setup.cfg for each rule.

dstansby:
:shipit:

Issue #7215, opened 2023-09-23
Provide a better error message when padding a rotated map of integer data with a missing value of NaN


ayshih:
Rotating a map of integer data using `missing` set to NaN (which is the default value) causes NumPy to error on this call to `numpy.pad()`:
https://github.com/sunpy/sunpy/blob/c7d9712262626704a331620f4b07dd7ee6779a17/sunpy/map/mapbase.py#L1704-L1708
We should wrap this call in a `try...except` so that we can provide a useful error message to users.

Alternatively, we could debate changing the default value for `missing`, but it's probably to keep the default as NaN.

samish55:
I've been trying understand the issue and wanted to ask can't we just add a line of code setting missing to 0 if issubclass(self.data.dtype.type ,numbers.Integeral) .

ayshih:
Yes, of course that could be done.  However, the reason we do not do that (anymore) is that zero is likely a valid pixel value, especially for integer data, and we don't want there to be confusion between a pixel having zero because that is the actual data value and pixel having zero because there was no data.  That is, unlike NaN, zero is not a "safe" value to represent missing data.

We could use map masks instead to denote missing data, and then setting pixels missing data to zero wouldn't be so bad, but our support for map masks is clunky.

NucleonGodX:
@ayshih kindly check #7344

alasdairwilson:
Closed as completed via #7344

PR #7344, opened 2023-12-12
Raise our own error when trying to rotate a map of integer data while using a `missing` value of NaN

NucleonGodX:
try...except added to deal with #7215 issue


nabobalis:
Hi @NucleonGodX, thank you for the pull request but if you could update the title of the pull request.

sunpy/map/mapbase.py
```diff
@@ -1706,11 +1706,17 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=3, scale=1.0,
...
-                          ((pad_y, pad_y), (pad_x, pad_x)),
-                          mode='constant',
-                          constant_values=(missing, missing))
+        try:
+            # Pad the image array 
+            new_data = np.pad(self.data, 
+                      ((pad_y, pad_y), (pad_x, pad_x)), 
+                      mode='constant', 
+                      constant_values=(missing, missing))
+
```
nabobalis, line 1715:
```suggestion
```

sunpy/map/mapbase.py
```diff
@@ -1706,11 +1706,17 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=3, scale=1.0,
...
+        except ValueError :
```
nabobalis, line 1716: Is it always a value error?

sunpy/map/mapbase.py
```diff
@@ -1706,11 +1706,17 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=3, scale=1.0,
...
+            # Handle the exception and provide a useful error message
```
nabobalis, line 1717:
```suggestion
            # Handle the exception and provide a useful error message
```
This comment isn't useful.

sunpy/map/mapbase.py
```diff
@@ -1706,11 +1706,17 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=3, scale=1.0,
         if version.parse(np.__version__) < version.parse("1.20.0") and issubclass(self.data.dtype.type, numbers.Integral) and (missing % 1 != 0):
             warn_user("The specified `missing` value is not an integer, but the data "
                       "array is of integer type, so the output may be strange.")
-        # Pad the image array
-        new_data = np.pad(self.data,
-                          ((pad_y, pad_y), (pad_x, pad_x)),
-                          mode='constant',
-                          constant_values=(missing, missing))
+        try:
+            # Pad the image array
```
nabobalis, line 1710:
```suggestion
```
This was here originally but its not really helpful.

sunpy/map/mapbase.py
```diff
@@ -1706,11 +1706,17 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=3, scale=1.0,
...
+            # Pad the image array 
+            new_data = np.pad(self.data, 
+                      ((pad_y, pad_y), (pad_x, pad_x)), 
+                      mode='constant', 
+                      constant_values=(missing, missing))
+
+        except ValueError :
+            # Handle the exception and provide a useful error message
+            error_message = f"Error occurred during padding as constant values are not numbers"
+            print(error_message)
```
nabobalis, line 1719: Using print is not how we handle errors, can you please find another example in our codebase and replace it with the correct method please?

nabobalis, line 1719: Using print is not how we handle errors, can you please find another example in our codebase and replace it with the correct method please?

sunpy/map/mapbase.py
```diff
@@ -1706,11 +1706,17 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=3, scale=1.0,
...
+        try:
+            # Pad the image array 
+            new_data = np.pad(self.data, 
+                      ((pad_y, pad_y), (pad_x, pad_x)), 
+                      mode='constant', 
+                      constant_values=(missing, missing))
+
+        except ValueError :
+            # Handle the exception and provide a useful error message
+            error_message = f"Error occurred during padding as constant values are not numbers"
```
nabobalis, line 1718: I think this error message needs improvement, I actually do not understand what this means.

sunpy/map/mapbase.py
NucleonGodX: we've set missing=np.nan from that I get value error only in padding

sunpy/map/mapbase.py
NucleonGodX:
```suggestion
            error_message = f"Error occurred during padding as constant values are NaN(not a number)"
            warn_user(error_message)
```

sunpy/map/mapbase.py
```diff
@@ -1706,11 +1706,14 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=3, scale=1.0,
...
-                          mode='constant',
-                          constant_values=(missing, missing))
+        try:
+            new_data = np.pad(self.data, 
+                      ((pad_y, pad_y), (pad_x, pad_x)), 
+                      mode='constant', 
+                      constant_values=(missing, missing))
+        except ValueError :
+            error_message = f"Error occurred during padding as constant values are NaN(not a number)"
+            warn_user(error_message)
```
nabobalis, line 1716: Can you add a unit test to ensure we hit this error please?

NucleonGodX: do you want me to add unit test in this code?

sunpy/map/mapbase.py
```diff
@@ -1706,11 +1706,14 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=3, scale=1.0,
...
-                          ((pad_y, pad_y), (pad_x, pad_x)),
-                          mode='constant',
-                          constant_values=(missing, missing))
+        try:
+            new_data = np.pad(self.data, 
+                      ((pad_y, pad_y), (pad_x, pad_x)), 
+                      mode='constant', 
+                      constant_values=(missing, missing))
+        except ValueError :
+            error_message = f"Error occurred during padding as constant values are NaN(not a number)"
```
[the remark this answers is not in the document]
NucleonGodX:
```suggestion
            error_message = f"Error occurred during padding as the missing is set to NaN by default kindly change the missing to an integer to resolve this error and add padding to the array"
```

sunpy/map/tests/test_mapbase.py
```diff
@@ -1036,12 +1036,12 @@ def test_rotate_with_incompatible_missing_dtype_warning():
...
     test_map = sunpy.map.Map(data, header)
+    
     with pytest.warns(SunpyUserWarning,
                       match="The specified `missing` value is not an integer, but the data "
-                      "array is of integer type, so the output may be strange."):
+                            "array is of integer type, so the output may be strange."):
         test_map.rotate(order=3, missing=np.nan)
 
-
 @pytest.mark.skipif(version.parse(np.__version__) <= version.parse("1.20.0"),
```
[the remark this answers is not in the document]
nabobalis: The same goes for any of the tests with this skipif.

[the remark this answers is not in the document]
NucleonGodX: oh yes, we don't need the skipif anymore

sunpy/map/tests/test_mapbase.py
```diff
@@ -1036,24 +1036,21 @@ def test_rotate_with_incompatible_missing_dtype_warning():
...
     test_map = sunpy.map.Map(data, header)
+    
     with pytest.warns(SunpyUserWarning,
                       match="The specified `missing` value is not an integer, but the data "
-                      "array is of integer type, so the output may be strange."):
+                            "array is of integer type, so the output may be strange."):
         test_map.rotate(order=3, missing=np.nan)
 
-
-@pytest.mark.skipif(version.parse(np.__version__) <= version.parse("1.20.0"),
```
nabobalis, line 1045: Can you remove this on the other test as well?

sunpy/map/mapbase.py
```diff
@@ -1706,11 +1706,14 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=3, scale=1.0,
...
-                          ((pad_y, pad_y), (pad_x, pad_x)),
-                          mode='constant',
-                          constant_values=(missing, missing))
+        try:
+            new_data = np.pad(self.data, 
+                      ((pad_y, pad_y), (pad_x, pad_x)), 
+                      mode='constant', 
+                      constant_values=(missing, missing))
+        except ValueError :
+            error_message = f"Error occurred during padding as the missing is set to NaN by default kindly change the missing to an integer to resolve this error and add padding to the array"
```
nabobalis, line 1715:
```suggestion
            error_message = "Unable to pad the array most likely due to the default `missing` keyword as it is set to NaN. Change the `missing` keyword to a different value."
```

sunpy/map/tests/test_mapbase.py
NucleonGodX: can you please elaborate what you want me to remove?

nabobalis: There were two tests marked with a pytest skipif. This should be removed

NucleonGodX: I think there is only one test_mapbase.py with skipif one is with xfail.

nabobalis: Is there not another one above this test?

NucleonGodX: the one above it is the xfail

nabobalis: Ah I didn't notice. Either way that should removed and the test fixed if needed.

sunpy/map/tests/test_mapbase.py
```diff
@@ -1028,31 +1028,24 @@ def test_rotate(aia171_test_map):
...
-def test_rotate_with_incompatible_missing_dtype_error():
-    data = np.arange(0, 100).reshape(10, 10)
-    coord = SkyCoord(0 * u.arcsec, 0 * u.arcsec, obstime='2013-10-28',
-                     observer='earth', frame=sunpy.coordinates.Helioprojective)
-    header = sunpy.map.make_fitswcs_header(data, coord)
-    test_map = sunpy.map.Map(data, header)
-    with pytest.raises(ValueError, match="cannot convert float NaN to integer"):
-        test_map.rotate(order=3, missing=np.nan)
 
+    if version.parse(np.__version__) >= version.parse("1.20.0"):
```
nabobalis, line 1038: Can you leave these tests as separate unit tests?

This version check is also not needed.

sunpy/map/mapbase.py
```diff
@@ -1706,11 +1706,15 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=3, scale=1.0,
...
-                          mode='constant',
-                          constant_values=(missing, missing))
+        try:
+            new_data = np.pad(self.data, 
+                      ((pad_y, pad_y), (pad_x, pad_x)), 
+                      mode='constant', 
+                      constant_values=(missing, missing))
+        except ValueError :
+            error_message = "Unable to pad the array most likely due to the default `missing` keyword as it is set to NaN. Change the `missing` keyword to a different value."
+            warn_user(error_message)
```
nabobalis, line 1716: I was wrong, this should not be a warning, this should still return the error and stop the code from running.

sunpy/map/tests/test_mapbase.py
NucleonGodX: But the version check is making sure that we are giving warning to under 1.20.0 numpy user, as any version equal to or above it will fail in it so will provide error rather than a warning? without it or xfail pytest will fail

sunpy/map/mapbase.py
NucleonGodX: So should we do like?
error_message=...........
and update unit test upon it?

sunpy/map/tests/test_mapbase.py
nabobalis: The error should be provided to the user. We also do not support any numpy version below 1.21.0.

sunpy/map/mapbase.py
nabobalis: Yes please.

sunpy/map/tests/test_mapbase.py
NucleonGodX: But do we even need the warning test then? Shouldn't we just have one test for error maybe?
And even in the main code, shouldn't we just omit the if condition?
`
           if version.parse(np.__version__) < version.parse("1.20.0") and issubclass(self.data.dtype.type, numbers.Integral) and (missing % 1 != 0):
            warn_user("The specified `missing` value is not an integer, but the data "
                      "array is of integer type, so the output may be strange.")
`

nabobalis: That sounds like a good idea.

nabobalis:
Can you add a changelog and fix the pre-commit please?

NucleonGodX:
"pre-commit.ci autofix"

NucleonGodX:
> Can you add a changelog and fix the pre-commit please?

done

nabobalis:
Looks like the test is failing, could you debug it please?

NucleonGodX:
> Looks like the test is failing, could you debug it please?

could you try now please ?

nabobalis:
> > Looks like the test is failing, could you debug it please?
> 
> could you try now please ?

It is passing now, thanks.

changelog/7344.feature.rst
```diff
@@ -0,0 +1 @@
+Added a new error giving user more context of padding error when running rotate function and removing the numpy version conditions.
```
ayshih, line 1:
```suggestion
When calling :meth:`sunpy.map.GenericMap.rotate` on an integer data array, with ``missing`` set to NaN (the default value), the method will now itself raise an informative error message instead deferring to NumPy to raise the error.
```

sunpy/map/mapbase.py
```diff
@@ -5,7 +5,6 @@
 import copy
 import html
 import inspect
-import numbers
```
ayshih, line 8: Add this line back in

sunpy/map/mapbase.py
```diff
@@ -1703,14 +1701,17 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=3, scale=1.0,
...
+            new_data = np.pad(self.data,
+                              ((pad_y, pad_y), (pad_x, pad_x)),
+                              mode='constant',
+                              constant_values=(missing, missing))
+        except ValueError:
+            error_message = (
+                "Unable to pad the array most likely due to the default `missing` keyword "
+                "as it is set to NaN. Change the `missing` keyword to a different value."
+            )
+            raise ValueError(error_message)
```
ayshih, line 1714:
```suggestion
    if (pad_x > 0 or pad_y > 0) and issubclass(self.data.dtype.type, numbers.Integral) and (missing % 1 != 0):
            raise ValueError("The underlying data is integers, but the fill value for missing "
                             "pixels cannot be cast to an integer, which is the case for the "
                             "default fill value of NaN.  Set the `missing` keyword to an "
                             "appropriate integer value for the data set.")

        new_data = np.pad(self.data,
                          ((pad_y, pad_y), (pad_x, pad_x)),
                          mode='constant',
                          constant_values=(missing, missing))         
```

sunpy/map/mapbase.py
```diff
@@ -1703,14 +1701,17 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=3, scale=1.0,
 
         # Numpy 1.20+ prevents np.pad from padding with NaNs in integer arrays
         # Before it would be cast to 0, but now it raises an error.
```
ayshih, line 1703:
```suggestion
        # Raise an informative error message if trying to pad an integer array with NaNs
```

sunpy/map/tests/test_mapbase.py
```diff
@@ -1028,29 +1027,14 @@ def test_rotate(aia171_test_map):
...
-                    reason="Numpy >= 1.20.0 doesn't allow NaN to int conversion")
 def test_rotate_with_incompatible_missing_dtype_error():
     data = np.arange(0, 100).reshape(10, 10)
     coord = SkyCoord(0 * u.arcsec, 0 * u.arcsec, obstime='2013-10-28',
                      observer='earth', frame=sunpy.coordinates.Helioprojective)
     header = sunpy.map.make_fitswcs_header(data, coord)
     test_map = sunpy.map.Map(data, header)
-    with pytest.raises(ValueError, match="cannot convert float NaN to integer"):
+    with pytest.raises(ValueError, match="Unable to pad the array most likely due to the default `missing` keyword as it is set to NaN. "
+    "Change the `missing` keyword to a different value."):
```
ayshih, line 1037:
```suggestion
    with pytest.raises(ValueError, match="The underlying data is integers, but the fill value for "
                                         "missing pixels cannot be cast to an integer"):
```

ayshih (changes requested):
Apologies, @NucleonGodX!  When I wrote #7215, I didn't realize that the change in NumPy behavior is older than our minimum dependency of  v1.21.  As such, we shouldn't bother with the `try...except` approach, and instead simply detect and error on our side.  See my suggested changes.

Also, please rename `7344.feature.rst` to `7344.bugfix.rst`

sunpy/map/mapbase.py
```diff
@@ -1701,12 +1700,13 @@ def rotate(self, angle: u.deg = None, rmatrix=None, order=3, scale=1.0,
...
         unpad_y = -np.min((diff[1], 0))
 
-        # Numpy 1.20+ prevents np.pad from padding with NaNs in integer arrays
-        # Before it would be cast to 0, but now it raises an error.
-        if version.parse(np.__version__) < version.parse("1.20.0") and issubclass(self.data.dtype.type, numbers.Integral) and (missing % 1 != 0):
-            warn_user("The specified `missing` value is not an integer, but the data "
-                      "array is of integer type, so the output may be strange.")
-        # Pad the image array
+        # Raise an informative error message if trying to pad an integer array with NaNs
+        if (pad_x > 0 or pad_y > 0 or missing % 1 != 0) and issubclass(self.data.dtype.type, numbers.Integral):
```
ayshih, line 1704:
```suggestion
        if (pad_x > 0 or pad_y > 0) and issubclass(self.data.dtype.type, numbers.Integral) and (missing % 1 != 0):
```
The change you made in adc72e0 has incorrect logic.  The error is now getting raised in situations where it shouldn't be raised.

ayshih (changes requested):
Thanks, looks good to me!

nabobalis (approved):

nabobalis:
Thanks for the PR @NucleonGodX

PR #7232, opened 2023-10-11
Add python 3.12 support

dstansby:
Lets see how badly the CI goes...

PR #7351, opened 2023-12-30
Add python 3.12 tests

nabobalis:
Supercedes https://github.com/sunpy/sunpy/pull/7232 

I updated the metadict str/repr to try and conform. 

Will require manual backports. 

**lxml 5.0.0 is breaking out doc build and our oldest deps build. We could pin against it or wait to see if its slow roll out of wheels**


sunpy/time/tests/test_time.py
```diff
@@ -352,7 +353,12 @@ def test_parse_time_list_3():
 
 
 def test_is_time():
-    assert time.is_time(datetime.utcnow()) is True
+    if sys.version_info >= (3, 12):
+        from datetime import UTC
+
+        assert time.is_time(datetime.now(UTC)) is True
+    else:
+        assert time.is_time(datetime.utcnow()) is True
```
nabobalis, line 361: This is easy but not the best way to handle this.

sunpy/util/metadata.py
```diff
@@ -67,8 +67,6 @@ def __init__(self, *args, save_original=True):
         if save_original and not hasattr(self, '_original_meta'):
             self._original_meta = MetaDict(*args, save_original=False)
 
-    def __str__(self):
-        return '\n'.join([f'{key}: {item}' for key, item in self.items()])
```
nabobalis, line 71: Just wondering if this will help or not.

.github/workflows/ci.yml
```diff
@@ -47,8 +47,8 @@ jobs:
         brew:
           - openjpeg
       envs: |
-        - windows: py310
-        - macos: py39
+        - windows: py311
+        - macos: py310
```
dstansby, line 51: Is it worth keeping around a `py39` normal job instead of relying on oldestdeps? I think it is.
```suggestion
        - macos: py310
        - linux: py39
```

setup.cfg
```diff
@@ -60,7 +61,7 @@ image =
   scipy>=1.7.0,!=1.10.0
 jpeg2000 =
   glymur>=0.9.1,!=0.9.5
-  lxml>=4.8.0
+  lxml>=4.8.0,!=5.0.0
```
dstansby, line 64: Is it better to pin this to `<5` for now? Or is the issue just with exactly `5.0.0`?

setup.cfg
```diff
@@ -271,7 +274,6 @@ omit =
   sunpy/cython_version*
   sunpy/*setup*
   sunpy/extern/*
-  sunpy/*/tests/*
```
dstansby, line 274: Why delete these lines?

sunpy/net/tests/strategies.py
```diff
@@ -15,6 +16,11 @@
 from sunpy.net import attrs as a
 from sunpy.time import TimeRange, parse_time
 
+if sys.version_info >= (3, 12):
+    now_year = datetime.datetime.now(datetime.UTC).year
```
dstansby, line 20: Does this not work with Pytbon 3.9? The docs suggest there's support for tzinfo on `now()`: https://docs.python.org/3.9/library/datetime.html#datetime.datetime.now

setup.cfg
nabobalis: We can remove this pin now, they uploaded the wheels needed.

setup.cfg
nabobalis: We should be checking our test coverage of our tests to ensure we don't keep finding tests we are incorrectly asserting.

sunpy/net/tests/strategies.py
nabobalis: I will adjust

setup.cfg
nabobalis: turns out not to be the case, we need to wait for lxml 5.0.1.

5.0.1 (2024-??-??)
Bugs fixed
LP#2046208: Parsing non-BMP Python Unicode strings could fail on macOS.

LP#2044225: When incrementally parsing broken HTML, reporting start events on
missing structural tags failed and could lead to subsequent exceptions.

LP#2045435: Some (not all) issues with stricter C compilers were resolved.

The binary wheels in the 5.0.0 release did not validate cleanly (but installed ok).

wtbarnes (approved):

PR #7352, opened 2023-12-30
Add text from an email from Albert - Lazy edition

nabobalis:
The text from https://groups.google.com/g/sunpy/c/RANTEGwyavE/m/a0XFxgmxAAAJ?utm_medium=email&utm_source=footer&pli=1

Its too verbose but I think we should add something to the example.

### DIFF
--- a/examples/plotting/plotting_blank_map.py
+++ b/examples/plotting/plotting_blank_map.py
@@ -54,34 +54,17 @@ xc = [0, 50, 400] * u.arcsec
 yc = [0, 100, 400] * u.arcsec
 
 ################################################################################
-# Note that the blue dot corresponding to (0, 0) in helioprojective coordinates is
-# not at the intersection of the heliographic equator and the heliographic prime
-# meridian. (0, 0) in helioprojective coordinates *is* at the center of the solar
-# disk, and at the center of the overall image. The reason that the intersection of
-# the two heliographic grid lines is not also at disk center is because the observer
-# (specified as Earth in this example) is at non-zero heliographic latitude (because
-# the Sun's equatorial plane is tilted relative to the ecliptic plane that contains
-# Earth's orbit). Disk center as seen by an observer has the same heliographic
-# latitude as the heliographic latitude of the observer, and this value is known as
-# the B0 angle, and varies over the year.
-#
-# To have no no offset (i.e., the heliographic equator is a horizontal line), you
-# can specify an observer on the Sun's equatorial plane instead of using Earth as
-# the observer.
-
-# sphinx_gallery_defer_figures
-
-observer_sun = SkyCoord(0*u.deg, 0*u.deg, 1*u.AU, obstime='2013-10-28',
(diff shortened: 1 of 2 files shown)


wtbarnes (approved):

examples/plotting/plotting_blank_map.py
```diff
@@ -53,12 +53,35 @@
...
+# (specified as Earth in this example) is at non-zero heliographic latitude (because
+# the Sun's equatorial plane is tilted relative to the ecliptic plane that contains
+# Earth's orbit). Disk center as seen by an observer has the same heliographic
+# latitude as the heliographic latitude of the observer, and this value is known as
+# the B0 angle, and varies over the year.
+#
+# To have no no offset (i.e., the heliographic equator is a horizontal line), you
+# can specify an observer on the Sun's equatorial plane instead of using Earth as
+# the observer.
+
```
ayshih, line 71:
```suggestion
```
I think having this sentence is of low value unless the example actually walks someone through how to do it.  My inclination is to remove the sentence.

examples/plotting/plotting_blank_map.py
```diff
@@ -53,12 +53,35 @@
...
+# sphinx_gallery_defer_figures
+
+observer_sun = SkyCoord(0*u.deg, 0*u.deg, 1*u.AU, obstime='2013-10-28',
+                    frame=frames.HeliographicStonyhurst)
+skycoord_sun = SkyCoord(0*u.arcsec, 0*u.arcsec, obstime='2013-10-28',
+                    observer=observer_sun, frame=frames.Helioprojective)
+
```
ayshih, line 78:
```suggestion
```
What you are doing here could have value, but is completely unsupported by the text.  The sentence above – which I want to remove anyway – is about creating a map with a horizontal-looking equator.  Plotting disk center as seen by a different observer is not the same thing at all.

examples/plotting/plotting_blank_map.py
```diff
@@ -53,12 +53,35 @@
...
 ################################################################################
 # Place and mark coordinates on the plot.
 
 coords = SkyCoord(xc, yc, frame=blank_map.coordinate_frame)
-p = ax.plot_coord(coords, 'o')
-# Set title.
+ax.plot_coord(coords, 'o')
+ax.plot_coord(skycoord_sun, 'o')
```
ayshih, line 84: You don't add any explanatory text, label, or legend for this marker, so I don't understand why it's even being plotted
```suggestion
```

ayshih (commented):
Hmm, I edited down the text from the email, but I'm truly confused about the purpose of the new coordinate you are plotting.

examples/plotting/plotting_blank_map.py
nabobalis: It was just to show the point plotted on the map that aligned with the center but with your other changes, I will remove all of this and leave the text you suggested.

PR #7362, opened 2024-01-04
Update the precommit to have a global exclude

nabobalis:
Simplify the config.


dstansby (changes requested):
I'm not sure we want to be globally excluding these files - what if e.g. someone adds a check that no large files are being added? Instead can we get rid of them where we don't need them in the individual checks? e.g. I'm pretty sure isort isn't going to try and sort a .fits file.

(also please stop updating the pre-commit versions in unreleated PRs, it makes the PRs much harder to review)

PR #7370, opened 2024-01-07
Add Warning for rsun Mismatch in reproject_to() Method

### LINKED ISSUE #7345: Show a warning when calling `reproject_to()` and there is a (minor) mismatch in `rsun`
People continue to get tripped up on a mismatch in `rsun` causing data to be lost in a reprojection (e.g., #7323 or https://community.openastronomy.org/t/derotate-hmi-map-at-carrington-rate-with-postel-projection/840/5).  Something like #7328 can help a user deal with an `rsun` mismatch, but only if they're aware that that's the problem.

We could emit a warning specific to `reproject_to()` such that it alerts a user if the `rsun` values of the map and the output header are not the same, since it might lead to unexpected results.  It should probably come with a tolerance parameter to control what level of `rsun` mismatch is considered a potential mistake.

Mubin17:
## PR Description

Fixes #7345.

This PR adds a warning mechanism in the `reproject_to()` method of the `GenericMap` class within SunPy. The enhancement aims to alert users when a `rsun` discrepancy (greater than 200000 m) exists between the input map and the target WCS during the reprojection process. This warning helps prevent potential data loss or unexpected results resulting from `rsun` mismatches.


sunpy/map/mapbase.py
```diff
@@ -2739,6 +2740,14 @@ def reproject_to(self, target_wcs, *, algorithm='interpolation', return_footprin
         outmap = GenericMap(output_array, target_wcs.to_header(),
                             plot_settings=self.plot_settings)
 
+        # Check rsun mismatch
+        rsun_map = self.meta["rsun_ref"]
+        rsun_target = outmap.meta["rsun_ref"]
+        rsun_diff = abs(rsun_map - rsun_target)
+        if rsun_diff > 200000:
```
nabobalis, line 2747: Can you suggest why you picked this value?

sunpy/map/mapbase.py
```diff
@@ -2739,6 +2740,14 @@ def reproject_to(self, target_wcs, *, algorithm='interpolation', return_footprin
...
+            warnings.warn("Warning: rsun mismatch greater than 200000 m detected. "
```
nabobalis, line 2748: We try to avoid using warnings.warn, can you replace it with the function warn_user instead?

nabobalis:
Thanks for the PR @Mubin17, can you add changelog and a unit test? If you need help with these, we can proivde that.

sunpy/map/mapbase.py
Mubin17: Will do

sunpy/map/mapbase.py
Mubin17:
```python
from sunpy.map import Map
import numpy as np

aiamap = Map("https://git.ias.u-psud.fr/adolliou/report_issue_github/-/raw/main/data/sunpy_reprojection_issue/aia.lev1.171.1497617052.2022-03-17T00_22_03.342Z.image_lev1.fits?ref_type=heads")
hrimap = Map("https://git.ias.u-psud.fr/adolliou/report_issue_github/-/raw/main/data/sunpy_reprojection_issue/solo_L2_eui-hrieuvopn-image_20220317T001800215_V02.fits?ref_type=heads")

aiamap_rsun_ref_val = aiamap.meta["rsun_ref"]
hrimap_rsun_ref_val = hrimap.meta["rsun_ref"]

# Looping through different rsun_ref differences to evaluate has_nan behavior
for i in range(101790, 101800, 1):
    # Modifying aiamap rsun_ref based on the iteration
    aiamap.meta["rsun_ref"] = aiamap_rsun_ref_val - i

    # Calculate the difference in rsun_ref values between the maps
    rsun_diff = abs(aiamap.meta["rsun_ref"] - hrimap_rsun_ref_val)

    # Reprojecting AIA map to HRI map and checking for NaN values
    aia_reprojected = aiamap.reproject_to(hrimap.wcs, algorithm="interpolation")
    has_nan = np.isnan(aia_reprojected.data).any()

    # Print the rsun_diff and has_nan status
    print(f"rsun_diff: {rsun_diff}, has_nan: {has_nan}")
    # aia_reprojected.peek()
```

```python
Output:
rsun_diff: 198210.0, has_nan: True
rsun_diff: 198209.0, has_nan: True
rsun_diff: 198208.0, has_nan: True
rsun_diff: 198207.0, has_nan: True
rsun_diff: 198206.0, has_nan: True
rsun_diff: 198205.0, has_nan: False
rsun_diff: 198204.0, has_nan: False
rsun_diff: 198203.0, has_nan: False
rsun_diff: 198202.0, has_nan: False
rsun_diff: 198201.0, has_nan: False
```

As `rsun_diff` decreased from 198210 to 198201, the `has_nan` transitioned from `True` (indicating data loss) to `False` (no data loss). Notably, at the `rsun_diff` value of 198205, the `has_nan` status changed to `False`, suggesting no data loss beyond this point. Therefore, `rsun_diff` values greater than 198205 consistently caused data loss during reprojection.

Based on the above evaluation I have selected new a threshold of 198205 which I believe is precise compared to 20000.

nabobalis: Hmm, the original issue suggested raising it if the values are different at all. 

@ayshih what are your thoughts on this?

ayshih: Rigorously, there is no special threshold at which a difference in rsun becomes problematic.  @Mubin17: The hidden assumption in your investigation is the pixel size of the original map.  Once the difference in rsun is smaller than the pixel size, it's possible to get through a reprojection without losing any data.

While a constant threshold for rsun difference would not work, there is the intriguing possibility of implementing a heuristic threshold derived from the pixel size of the original map.  Barring that, the safest thing to do for the time being is to raise the warning for any difference in rsun, no matter how small.

Mubin17: @ayshih @nabobalis Thanks for the clarification. I've implemented your feedback, raising the warning for any difference in `rsun`, regardless of its magnitude.

sunpy/map/mapbase.py
```diff
@@ -2739,6 +2739,13 @@ def reproject_to(self, target_wcs, *, algorithm='interpolation', return_footprin
         outmap = GenericMap(output_array, target_wcs.to_header(),
                             plot_settings=self.plot_settings)
 
+        # Check rsun mismatch
+        rsun_map = self.meta["rsun_ref"]
+        rsun_target = outmap.meta["rsun_ref"]
+        if rsun_map != rsun_target:
+            warn_user("rsun mismatch detected. "
```
nabobalis, line 2746: Do we want this to mention how to fix it?

dstansby, line 2746: Could you print the values and the names of the map that the different `rsun` values are coming from so this error message is clearer? Something like `warn_user(f"rusn mismatch detected: {self.name}.rsun_obs={self.rsun_obs}; {outmap.name}.rsun_obs={outmap.rsun_obs}"). This will make it easier for the user to understand which maps have different values and how different those values are.

dstansby (commented):
I've added a suggestion for a clearer message. Otherwise this looks good 👍

sunpy/map/mapbase.py
Mubin17: Sure thats a good idea.

sunpy/map/mapbase.py
```diff
@@ -2739,6 +2739,14 @@ def reproject_to(self, target_wcs, *, algorithm='interpolation', return_footprin
         outmap = GenericMap(output_array, target_wcs.to_header(),
                             plot_settings=self.plot_settings)
 
+        # Check rsun mismatch
+        rsun_map = self.meta["rsun_ref"]
+        rsun_target = outmap.meta["rsun_ref"]
+        if rsun_map != rsun_target:
```
nabobalis, line 2745: Would the walrus operator be useful here?

Mubin17: Did you mean using something like this??
```
if (rsun_map := self.meta["rsun_ref"]) != (rsun_target := outmap.meta["rsun_ref"]):
```

nabobalis: Yeah, would that work?

Mubin17: Yes

Mubin17: I see that pre-commit isn’t happy about local variables `rsun_map` and `rsun_target` which are never use. 
Instead can we simply do this 
```
if self.meta["rsun_ref"] != outmap.meta["rsun_ref"]:
```
as there is no real use of assigning `self.meta["rsun_ref"]` `outmap.meta["rsun_ref"]` to local variables.

nabobalis: Were they not used for the error message?

Mubin17: No. For the error message we make direct calls to `self` and `outmap`.
```python
warn_user("rsun mismatch detected: "
         f"{self.name}.rsun_obs={self.rsun_obs}; {outmap.name}.rsun_obs={outmap.rsun_obs}. "
```

nabobalis: Ah I misread the warn statement.

nabobalis: but then why can not we use the same keywords here instead of defining new variables?

Mubin17: Yes we can simply do something like this:
```python
if self.rsun_meters != outmap.rsun_meters:
   warn_user("rsun mismatch detected: "
            f"{self.name}.rsun_meters={self.rsun_meters}; {outmap.name}.rsun_meters={outmap.rsun_meters}. "
```
Or

nabobalis: I would assume so if that is what we are using to check the difference between two maps?

nabobalis (approved):

wtbarnes (approved):

PR #7372, opened 2024-01-08
remove filter on warnings for release docs

### LINKED ISSUE #7301: User confusion because warnings are hidden in gallery examples for stable docs
No doubt there was a good reason to do this, but it can be confusing to users that warnings are hidden in the gallery examples for stable docs (was done in #5279):
https://github.com/sunpy/sunpy/blob/bc8f08fe912adb1396e1a0b976ae98bdc32170a8/docs/conf.py#L59-L61

That means when a user runs the example on their own machine, and they see warning messages that do not appear on the rendered example, the natural reaction is that the user thinks something is not set up properly on their end.

Can we stop hiding the warnings?

nabobalis:
Fixes #7301

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -56,14 +56,11 @@ release = __version__
 sunpy_version = Version(__version__)
 is_release = not(sunpy_version.is_prerelease or sunpy_version.is_devrelease)
 
-# We want to ignore all warnings in a release version.
-if is_release:
-    warnings.simplefilter("ignore")
+# We want to make sure all the following warnings fail the build
 warnings.filterwarnings("error", category=SunpyDeprecationWarning)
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -56,14 +56,11 @@
 sunpy_version = Version(__version__)
 is_release = not(sunpy_version.is_prerelease or sunpy_version.is_devrelease)
 
-# We want to ignore all warnings in a release version.
-if is_release:
-    warnings.simplefilter("ignore")
+# We want to make sure all the following warnings are not emitted
```
nabobalis, line 59:
```suggestion
# We want to make sure all the following warnings fail the build.
```

dstansby (approved):

wtbarnes (approved):

PR #7378, opened 2024-01-11
add replacement to mapbase for unit: counts / pixel -> ct/pix

alasdairwilson:
EIT maps have the non FITS compliant "counts / pixel" this adds a replacement for them to "ct/pix" in mapbase

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -727,7 +727,8 @@ class GenericMap(NDData):
     def _parse_fits_unit(unit_str):
         replacements = {'gauss': 'G',
                         'dn': 'ct',
-                        'dn/s': 'ct/s'}
+                        'dn/s': 'ct/s',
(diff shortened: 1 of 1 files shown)


nabobalis (approved):

nabobalis:
You want this in a bugfix release?

alasdairwilson:
> You want this in a bugfix release?

Yeah think so, it isn't a breaking bug or anything but it was surely not intended we couldn't parse these units, even though we use some eit data in our examples.

sunpy/map/mapbase.py
```diff
@@ -727,15 +727,16 @@ def max(self, *args, **kwargs):
     def _parse_fits_unit(unit_str):
         replacements = {'gauss': 'G',
```
nabobalis, line 728: I wonder if in future we move this out somewhere and we can extend it without having to hunt for it in this file. But thats not for this PR.

PR #7383, opened 2024-01-11
Increase versions of our deps for 6.0 with a bump to astropy 5.2 + removal of maxwell unit

nabobalis:
I think it's time for 6.0 to increase astropy to 5.2 to align with 12 months.

I also did the other packages and removed the maxwell unit.

PR #7384, opened 2024-01-13
move to pyproject.toml + dedicated tool configs

nabobalis:
Removes setup.{cfg,py} and adds everything into the pyproject.toml file.
Added a twine check to make sure our sdist is correct.
It also tells us what files were packaged in so its, useful.

Doc build is broken but I am waiting for upstream to fix it: https://github.com/OpenAstronomy/sphinx-changelog/issues/22 

Question:
- [x] Changelog?


sunpy/util/tests/test_sysinfo.py
```diff
@@ -8,61 +8,63 @@
...
 
 def test_find_dependencies():
     missing, installed = find_dependencies()
     assert missing == {}
     assert sorted(list(installed.keys())) == sorted(["astropy", "numpy", "packaging", "parfive", "requests"])
 
-    missing, installed = find_dependencies(package="sunpy", extras=["required", "all"])
+    missing, installed = find_dependencies(package="sunpy", extras=["required", *EXTRA_ALL_GROUPS])
     assert missing == {}
-    assert sorted(list(installed.keys())) == sorted(['asdf',
```
nabobalis, line 19: Finally updated these to hopefully be less annoying

README.rst
```diff
@@ -2,107 +2,56 @@
...
-            ><img
-                src="https://img.shields.io/badge/powered%20by-NumFOCUS-orange.svg?style=flat&colorA=E1523D&colorB=007D8A"
-                alt="Powered by NumFOCUS"
-            /></a>
-        </td>
-        </tr>
-    </tbody>
-    </table>
-    <br />
+|SunPy Logo|
```
nabobalis, line 5: Modifying this is going to be a pain that we will never escape.

Cadair: The rst table doesn't look too bad? I appreciate reflowing tables is a pain. We could make it much wider in terms of columns to make it less likely we need to reflow it?

.circleci/config.yml
```diff
@@ -27,6 +27,16 @@ apt-run: &apt-install
     sudo apt install -y libopenjp2-7
 
 jobs:
+  twine-check:
```
Cadair, line 30: Why are we adding this here?

pyproject.toml
```diff
@@ -6,113 +6,298 @@ requires = [
...
+  "h5py>=3.1.0",
+  "matplotlib>=3.5.0",
+  "pandas>=1.2.0",
+]
+visualization = [
+  "matplotlib>=3.5.0",
+  "mpl-animators>=1.0.0",
+]
+all = [
+  "sunpy[asdf,dask,image,jpeg2000,map,net,spice,timeseries,visualization]",
```
Cadair, line 98: This syntax didn't used to be supported right? What version of setuptools etc do we need to support this?

Cadair (commented):
Looks generally good. My only concern is making sure we aren't requiring a newer than specified pip/setuptools to allow the dependency grouping.

.circleci/config.yml
nabobalis: So we know we do not break the sdist? Without it, I would not have known the html table would not render.

pyproject.toml
nabobalis: I will have to check and increase the min version.

.circleci/config.yml
Cadair: but why on circle and not GHA?

pyproject.toml
Cadair: I worry that if it's a pip version we wont be able to use it for a while. We should work on the principle that the min pip version is the one that ships with the min version of Python we have.

Cadair: That being said I suspect it's setuptools not pip.

nabobalis: I know I had someone not able to instal a project due to an old pip that only had a pyproject. I forget if their update of pip also brought in setuptools (i would assume it had it)

.circleci/config.yml
nabobalis: It was easier? I could create a tox env for this and add it to the cron actions?

nabobalis: Does our actions config allow arbitrary commands like this?

Cadair: Not all our actions things have to run in tox. You can add a not-openastronomy workflows jobs in the same way: https://github.com/astropy/astropy/blob/main/.github/workflows/ci_workflows.yml#L20

pyproject.toml
Cadair: I pulled the 3.9.0 docker container.

```
Python 3.9.0 (default, Nov 18 2020, 13:28:38) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pip
>>> pip.__version__
'20.3.1'
>>> import setuptools
>>> setuptools.__version__
'50.3.2'
```

As long as the build dependencies are installed correctly, I don't *think* we need to worry about the setuptools version.

Doing this `pip install git+https://github.com/nabobalis/sunpy@pyproj#egg=sunpy[all]` inside that container, did not work. It started looking for sunpy releases on pypi with all the extras we needed.

Doing it on python 3.10.0 worked though, which would be our minimum Python for the next release? So I think this PR is ok as long as we bump to min 3.10 at the same time.

.circleci/config.yml
Cadair: I wonder if we should add this to one of the OA publish workflows somewhere :thinking:

nabobalis: Should be added to the source dist one no?

pyproject.toml
nabobalis: I will bump

README.rst
nabobalis: Reflowed to much wider

dstansby:
There's lots of changes in here that seem orthogonal to migrating to `pyproject.toml`... is there any chance you can split those into a separate PR to make this easier to review and less likely for a mistake to slip through?

nabobalis:
I think the readme changes and the new sdist check the only ones? 

We had to bump to 3.10 to ensure the pyproject only is supported and that cames with some style changes. I could move that into the other PR I have open about min deps.

sunpy/coordinates/frames.py
```diff
@@ -161,7 +161,7 @@ def represent_as(self, base, s='base', in_frame_units=False):
 
         # If a frame wrap angle is set, use that wrap angle for any spherical representations.
         if self._wrap_angle is not None and \
-           isinstance(data, (UnitSphericalRepresentation, SphericalRepresentation)):
+           isinstance(data, UnitSphericalRepresentation | SphericalRepresentation):
```
nabobalis, line 164: These changes are from the pre-commit now we have a min of 3.10.

pyproject.toml
```diff
@@ -2,117 +2,301 @@
...
```
nabobalis, line 97:
```suggestion
  "sunpy[asdf,dask,image,jpeg2000,map,net,timeseries,visualization]",
```

.github/workflows/ci.yml
```diff
@@ -100,7 +99,7 @@ jobs:
             apt:
               - libopenjp2-7
               - graphviz
-        - linux: py39-online
+        - linux: py312-online
```
CyclingNinja, line 102: Double checking, everything else has been moved to 3.10, is this being 3.12 deliberate

CyclingNinja:
Tests are currently failing with the pandas/pyarrow error, needs merge from main (https://github.com/sunpy/sunpy/commit/b52096077f4857bf9d6695603433dc1951e42b57)

.github/workflows/ci.yml
nabobalis: This was a random choice, I could pick 310,311,312. I just picked 312 as I was lazy.

nabobalis:
I would like:

- https://github.com/sunpy/sunpy/pull/7383
- https://github.com/sunpy/sunpy/pull/7402
- https://github.com/sunpy/sunpy/pull/7400

merged in before this one is please.

CyclingNinja (approved):

Cadair (approved):

alasdairwilson (approved):
nice 👍

setup.py
Cadair, line 1: I wonder if we should keep this.

Without it certain old versions of pip wont work iirc.

Cadair (changes requested):
We need to decide if it's too early to burn setup.py.

setup.py
alasdairwilson: pip >21.3 right which is coincidentally from the same week as python 3.10 was released.

Cadair: Ah, if it's "pip which shipped with python 3.10" then we should be fine as 3.10 is the min version on main now.

wtbarnes:
> We need to decide if it's too early to burn setup.py.

Why? What's the objection?

Cadair:
Editable installs don't work with the version of pip shipped in our minimal python requirements (https://github.com/sunpy/sunpy/pull/7384#discussion_r1479674823)

setup.py
alasdairwilson: I am not sure this is an "honest" test, if you just use the python:3.10 docker image you get pip 23.0.1

`ENV PYTHON_PIP_VERSION=23.0.1`

It is only by forcing to 3.10.0 (or .1 or .2, the pip version updates to 22+ in 3.10.3) that you get an incompatible pip version.

People would have to have not updated the micro version of their python in several years, which isn't impossible, and its not like pip >21.3 is incompatible with 3.10.0 (or indeed with 3.9 etc.).

alasdairwilson: I mean I don't particularly object strongly to keeping setup.py, just providing extra context there.

Cadair: But I feel like we should support Python 3.10.0 as that's the minimum version we claim to support, that's why I tested against it.

nabobalis: Lets bump it.

nabobalis: It is now backk.

Cadair (approved):

PR #7396, opened 2024-01-24
add wheel upload to anaconda sunpy channel

nabobalis:
I added a token to the action secrets page.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -174,6 +174,10 @@ jobs:
     needs: [test]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@main
     with:
+      upload_to_anaconda: ${{ (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') }}
+      anaconda_user: sunpy
+      anaconda_package: sunpy
+      anaconda_keep_n_latest: -1
       sdist: false
       test_extras: 'tests'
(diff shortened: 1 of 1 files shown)


tox.ini
Cadair, line 1: Why do you think we should remove this? Surely all the reasons we added them still apply? online tests are flakey and sometimes re-running helps and sometimes they stall out forever?

Also this change isn't actually what the PR says it is about.

.github/workflows/ci.yml
```diff
@@ -184,11 +188,16 @@ jobs:
...
+      anaconda_token: ${{ secrets.anaconda_token }}
 
   publish_pure:
     needs: [publish]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@main
     with:
+      upload_to_anaconda: ${{ (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') }}
+      anaconda_user: sunpy
+      anaconda_package: sunpy
+      anaconda_keep_n_latest: -1
```
Cadair, line 200:
```suggestion
      anaconda_keep_n_latest: 10
```

tox.ini
nabobalis: I will remove the change and we can debate this in another PR

Cadair (approved):

nabobalis:
I think this counts as a minor change.

PR #7397, opened 2024-01-24
Add pyerfa as core dep and fix imports (fixes devdeps)

nabobalis:
Import has now changed.

### DIFF
--- a/changelog/7397.trivial.rst
+++ b/changelog/7397.trivial.rst
@@ -1,0 +1,2 @@
+``pyerfa`` is now a new direct dependency.
+It has been an indirect dependency from sunpy 3.1, over two years ago.
--- a/setup.cfg
+++ b/setup.cfg
@@ -46,6 +46,7 @@ install_requires =
   numpy>=1.21.0
   packaging>=19.0
   parfive[ftp]>=2.0.0
+  pyerfa
   requests
 
 [options.extras_require]
--- a/sunpy-dev-env.yml
+++ b/sunpy-dev-env.yml
@@ -16,6 +16,7 @@ dependencies:
   - packaging
   - pandas
   - parfive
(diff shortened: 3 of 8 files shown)


ayshih:
("long overdue" because we could have "fixed" those imports once Astropy 4.2 became the minimum dependency)

nabobalis:
Done.

dstansby (changes requested):
The new dependency should get a changelog entry. Also I'm not sure about backporting a new dependency... can we just ignore the warning on 5.x?

nabobalis:
> The new dependency should get a changelog entry. Also I'm not sure about backporting a new dependency... can we just ignore the warning on 5.x?

Hmm, I would say that since pyefra has been a core dependency of astropy for years now, the fact we are adding to ours I don't think changes anything. 

I can add the ignore warning to those releases. 

I will leave this choice up to you and Albert.

ayshih:
While `pyerfa` would become a new *direct* dependency, it has been an indirect dependency since our minimum `astropy` dependency became 4.2 (i.e., `sunpy` 3.1, over two years ago).  That is, making an indirect dependency a direct dependency has zero effect on anyone with a valid installation of `sunpy` 3.1+, since their environment is "guaranteed" to have `pyerfa` present.

My inclination is to go ahead with the backport, because it cleans up the code, but I do think there should be a changelog (`trivial`), primarily to alert anyone who might have a bizarrely inconsistent environment.

nabobalis:
Rebaesd and added changelog.

ayshih (approved):
Can you do me a favor and add it to `sunpy-dev-env.yml` as well?

nabobalis:
Ah I forgot!

dstansby (approved):

Issue #7161, opened 2023-08-21
Divide up and remove `sunpy.physics.differential_rotation`


ayshih:
### Provide a general description of the issue or problem.

`sunpy.physics.differential_rotation` is the only module in that sub-package, and contains three items, each of which has a more appropriate place to live:

- `diff_rot()`, which can be moved to `sunpy.sun.models`
- `solar_rotate_coordinate()`, which can be moved to `sunpy.coordinates.utils`
- `differential_rotate()`, which can be moved to `sunpy.map.maputils`

Then we can get rid of `sunpy.physics.differential_rotation` (after deprecation), and thus `sunpy.physics` as well.

NucleonGodX:
I'd try to work on this one!

tanishy7777:
Is this fixed @ayshih ?

PR #5983, opened 2022-03-21
Deprecate sunpy.io.fits

dstansby:
Fixes https://github.com/sunpy/sunpy/issues/5882. I decided the best way to do this was:

- Rename the original file to `sunpy.io._fits.py`
- Change internal code to import from this
- Put a deprecation warning that emits every time `sunpy.io.fits` is imported

I think I've found everything that needs changing in docs etc. with a grep for `sunpy.io.fits.`

Issue #5292, opened 2021-05-03
Rewrite `differential_rotate()` to use `RotatedSunFrame`+`reproject`


ayshih:
`differential_rotate()` and `RotatedSunFrame`+`reproject` are now giving the same results (see #3931).  That means we can now rewrite `differential_rotate()` to use `RotatedSunFrame`+`reproject`.

wafels:
Just to confirm, both full disk maps and submaps give the same answers (up to the differences due to interpolation methods as noted in #3931) when using the two methods?

ayshih:
I mean, I wouldn't stake my first-born on it, but I'm pretty sure, yes.

wtbarnes:
Another motivator for this proposed change: this function uses `skimage.transform.warp` and thus rewriting it is a necessary step in removing our dependency on `scikit-image`.

PR #7399, opened 2024-01-24
Moved ```sunpy.physics.differential_rotation.diff_rot``` to ```sunpy.sun.models.diff_rot``` and deprecated the function in ```sunpy.physics.differential_rotation```

Ghaithq:
Divided up sunpy.physics according to issue #7161. The tests of each function was also divided according to the position of the relocated functions.

- ```diff_rot()```, was moved to ```sunpy.sun.models```


Ghaithq:
pre-commit.ci autofix

nabobalis:
Thanks for the PR @Ghaithq, however you can not delete the old code as it has to go thourgh a deprecation period first. See https://docs.sunpy.org/en/latest/dev_guide/contents/public_api.html for more infomation. 

The moved functions and tests are fine but you have to keep the old API alive for now.

Ghaithq:
> Thanks for the PR @Ghaithq, however you can not delete the old code as it has to go thourgh a deprecation period first. See https://docs.sunpy.org/en/latest/dev_guide/contents/public_api.html for more infomation.
> 
> The moved functions and tests are fine but you have to keep the old API alive for now.

should i return the old code?

nabobalis:
Yes you need to return the original code, add deprecation warnings and point to the new locations.

Ghaithq:
> Yes you need to return the original code, add deprecation warnings and point to the new locations.

ok I will try to do it right now. Is everything else fine?

sunpy/sun/models.py
```diff
@@ -84,3 +89,119 @@
 evolution = QTable(_t)
 evolution.source = 'Unknown'
 evolution.add_index('time')
+
+
+
+@u.quantity_input
+def diff_rot(duration: u.s, latitude: u.deg, rot_type='howard', frame_time='sidereal'):
```
nabobalis, line 96: Is it time to rename this?

nabobalis:
> > Yes you need to return the original code, add deprecation warnings and point to the new locations.
> 
> ok I will try to do it right now. Is everything else fine?

On a quick glance, it looks fine to me. Locations make sense to me. When the old code is added and deprecated, I will tag in others to review.

sunpy/sun/models.py
Ghaithq: what exactly, the function name? I don't recall renaming anything except ```_get_new_observation``` which I changed to ```_get_new_observation``` since it's no longer a private method.

nabobalis: I did not say you renamed anything. This is a question for the other maintainers if we want to rename this while we are going through this process.

Ghaithq: Oh okay sorry I misunderstood.

Ghaithq:
> > > Yes you need to return the original code, add deprecation warnings and point to the new locations.
> > 
> > 
> > ok I will try to do it right now. Is everything else fine?
> 
> On a quick glance, it looks fine to me. Locations make sense to me. When the old code is added and deprecated, I will tag in others to review.
> 
> You do not need to add the code back but the functions, you could import them inside the functions to avoid having duplicate code.

I already returned it. Should I've not?

nabobalis:
> > > > Yes you need to return the original code, add deprecation warnings and point to the new locations.
> > > 
> > > 
> > > ok I will try to do it right now. Is everything else fine?
> > 
> > 
> > On a quick glance, it looks fine to me. Locations make sense to me. When the old code is added and deprecated, I will tag in others to review.
> > You do not need to add the code back but the functions, you could import them inside the functions to avoid having duplicate code.
> 
> I already returned it. Should I've not?

Well there are several ways to handle it, for example https://github.com/sunpy/sunpy/pull/5983/files moves the code but imports it back and adds a warning into the namespace.

sunpy/physics/differential_rotation.py
```diff
@@ -144,7 +146,7 @@ def diff_rot(duration: u.s, latitude: u.deg, rot_type='howard', frame_time='side
 
     return Longitude(rotation.to(u.deg))
 
-
+@deprecated(since="5.1.1", message="We will be moving this", alternative="sunpy.coordinates.utils._validate_observer_args")
```
nabobalis, line 149: 1) You only need to do this for public API, the leading underscore marks this as private API (its a style choice in Python).

2) This will come out in sunpy 6.0 and not 5.1.1.

Ghaithq: Okay I will fix this. I thought since reflects the version in which the deprecation happened.

Ghaithq:
pre-commit.ci autofix

nabobalis:
Thanks again for doing all of this, I spoke to our coordinates maintainer and he told me about open issue: https://github.com/sunpy/sunpy/issues/5292

He said that moving the functions without any API changes would be a lost opportunity as we would have to break the function twice instead of once. 

So I would suggest that if you could scale back this PR to only moving `diff_rot()` to `sunpy.sun.models` and whatever else codewise has to move to support that and leave the other functions alone.

Would that be ok?

Ghaithq:
> Thanks again for doing all of this, I spoke to our coordinates maintainer and he told me about open issue: #5292
> 
> He said that moving the functions without any API changes would be a lost opportunity as we would have to break the function twice instead of once.
> 
> So I would suggest that if you could scale back this PR to only moving `diff_rot()` to `sunpy.sun.models` and whatever else codewise has to move to support that and leave the other functions alone.
> 
> Would that be ok?

Yeah sure, no problem. Actually can I work on #5292 too? and if I rewrote ```differential_rotate()``` to use ```RotatedSunFrame()```+```reproject()```, functions other than ```diff_rot()``` should remain in ```sunpy.physics```?

nabobalis:
> Yeah sure, no problem. Actually can I work on #5292 too? and if I rewrote `differential_rotate()` to use `RotatedSunFrame()`+`reproject()`, functions other than `diff_rot()` should remain in `sunpy.physics`?

That would be a complex issue to tackle, I would not recommend it but you are free to have a go. 
That should happen in a separate PR. 

For this PR, just focus on diff_rot and we can worry about moving the other functions later on.

Ghaithq:
pre-commit.ci autofix

Ghaithq:
> > Yeah sure, no problem. Actually can I work on #5292 too? and if I rewrote `differential_rotate()` to use `RotatedSunFrame()`+`reproject()`, functions other than `diff_rot()` should remain in `sunpy.physics`?
> 
> That would be a complex issue to tackle, I would not recommend it but you are free to have a go. That should happen in a separate PR.
> 
> For this PR, just focus on diff_rot and we can worry about moving the other functions later on.

Made the changes. Hope this works. If there's any modification please tell me and I'll fix them.

sunpy/physics/differential_rotation.py
```diff
@@ -25,12 +25,14 @@
...
 from sunpy.time import parse_time
 from sunpy.util import expand_list
+from sunpy.util.decorators import deprecated
 from sunpy.util.exceptions import warn_user
 
 __all__ = ['diff_rot', 'solar_rotate_coordinate', 'differential_rotate']
 
 
 @u.quantity_input
+@deprecated( since="6.0",message="We will be moving this", alternative="sunpy.sun.models.diff_rot")
```
nabobalis, line 35:
```suggestion
@deprecated(since="6.0",message="We will be moving this", alternative="sunpy.sun.models.diff_rot")
```

What does the deprecation warning look like?

nabobalis:
Thanks that looks better, can you rebase this PR and fix the test failures please?

sunpy/physics/differential_rotation.py
Ghaithq: Not sure. I copied the line from the documentation and changed the values of parameters.

Ghaithq:
> Thanks that looks better, can you rebase this PR and fix the test failures please?
> 
> Oh can you also change the title to be more detailed please.

Yeah sure. will work on it now.

Ghaithq:
pre-commit.ci autofix

Ghaithq:
> Thanks that looks better, can you rebase this PR and fix the test failures please?
> 
> Oh can you also change the title to be more detailed please.

Am I supposed to fix all checks?

sunpy/physics/differential_rotation.py
nabobalis: Can you share/check what the warning looks like?

nabobalis:
> > Thanks that looks better, can you rebase this PR and fix the test failures please?
> > Oh can you also change the title to be more detailed please.
> 
> Am I supposed to fix all checks?

Just "CI / core / py312 (ubuntu-latest) (pull_request)"

The other one that is failing is a known problem.

sunpy/physics/differential_rotation.py
Ghaithq: when I ran the tests this was the output
```FAILED sunpy/physics/tests/test_differential_rotation.py::test_rigid - sunpy.util.exceptions.SunpyDeprecationWarning: We will be moving this```

nabobalis:
> Should I remove the diff_rot() tests in physics.tests?

We still need the tests, they should be moved and updated to call the new function.

sunpy/physics/differential_rotation.py
nabobalis: In that case, the message doesn't really offer much.

Ghaithq:
> > Should I remove the diff_rot() tests in physics.tests?
> 
> We still need the tests, they should be moved and updated to call the new function.

I moved the tests to sunpy.sun.tests but should I leave the tests in sunpy.physics. I think if I removed them the code coverage will decrease.

sunpy/physics/differential_rotation.py
Ghaithq: Should I make it more detailed?

nabobalis:
> > > Should I remove the diff_rot() tests in physics.tests?
> > 
> > 
> > We still need the tests, they should be moved and updated to call the new function.
> 
> I moved the tests to sunpy.sun.tests but should I leave the tests in sunpy.physics. I think if I removed them the code coverage will decrease.

One set of tests should be enough.

sunpy/physics/differential_rotation.py
nabobalis: That message should be removed.

Ghaithq:
> > > > Should I remove the diff_rot() tests in physics.tests?
> > > 
> > > 
> > > We still need the tests, they should be moved and updated to call the new function.
> > 
> > 
> > I moved the tests to sunpy.sun.tests but should I leave the tests in sunpy.physics. I think if I removed them the code coverage will decrease.
> 
> One set of tests will be enough.
> 
> What you want to add is a unit test that imports the old function and make sures the deprecation warning is raised.

can I use ```test_deprecated_warning_message(since, pending, warning, message, warning_message)```

PR #7402, opened 2024-01-25
Min python version is now py310

nabobalis:
The whatsnew is updated by https://github.com/sunpy/sunpy/pull/7383/ and not here.

This is in preparation for the pyproject.toml where we need to bump to a min of 310 as well as our support cycle.

### DIFF
--- a/sunpy/coordinates/utils.py
+++ b/sunpy/coordinates/utils.py
@@ -420,7 +420,7 @@ def solar_angle_equivalency(observer):
     <Quantity 1.00603718 arcsec>
     """
 
-    if not isinstance(observer, (SkyCoord, BaseCoordinateFrame)):
+    if not isinstance(observer, SkyCoord | BaseCoordinateFrame):
         raise TypeError(
             "Invalid input, observer must be of type SkyCoord or BaseCoordinateFrame.")
     if observer.obstime is None:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,8 +49,7 @@ jobs:
       envs: |
         - windows: py311
         - macos: py310
-        - linux: py39
-        - linux: py39-oldestdeps
+        - linux: py310-oldestdeps
     secrets:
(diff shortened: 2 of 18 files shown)


sunpy/coordinates/utils.py
```diff
@@ -420,7 +420,7 @@ def solar_angle_equivalency(observer):
     <Quantity 1.00603718 arcsec>
     """
 
-    if not isinstance(observer, (SkyCoord, BaseCoordinateFrame)):
+    if not isinstance(observer, SkyCoord | BaseCoordinateFrame):
```
dstansby, line 423: 🥳

dstansby:
The figure devdeps fail is:

> E   astropy.utils.exceptions.AstropyDeprecationWarning: Importing ErfaWarning from astropy.utils.exceptions was deprecated in version 6.1 and will stop working in a future version. Instead, please use from erfa import ErfaWarning

Worth fixing here, or for another PR?

nabobalis:
That is fixed in another PR, waiting for Albert to review

CyclingNinja (approved):

sunpy/coordinates/utils.py
Cadair: Is this actually a good change? It seems amazingly pointless to me.

Cadair (approved):

sunpy/coordinates/utils.py
CyclingNinja: I think its more readable, its much clearer that a logical operator is happening at that point imo

nabobalis: THE RUFF SAYS WE DO IT SO WE DO IT

nabobalis:
Oh that reminds me:

```
E       asdf.exceptions.AsdfWarning: asdf.extensions plugin from package asdf-astropy==0.1.1 failed to load:
E       
E       AsdfWarning: asdf.resource_mappings plugin from package asdf==2.8.3 failed to load:
E       
E       RuntimeError: Unable to locate core schemas
```
can we delete asdf?

PR #7405, opened 2024-01-25
[5.0] Add pyerfa as core dep and fix imports (fixes devdeps) #7397

nabobalis:


setup.cfg
```diff
@@ -46,6 +46,8 @@ install_requires =
   numpy>=1.21.0
   packaging>=19.0
   parfive[ftp]>=2.0.0
+  pyerfa
+  requests
```
nabobalis, line 50:
```suggestion
```

PR #7407, opened 2024-01-25
Information on syncing tags in the newcomers guide 

nabobalis:


wtbarnes (approved):

docs/dev_guide/contents/newcomers.rst
```diff
@@ -164,6 +164,13 @@ This will make submitting changes easier in the long term for you:
 
     $ git clone https://github.com/<username>/sunpy.git sunpy-git
     $ cd sunpy-git
+    # This adds the main sunpy repository as a remote called "upstream".
+    # This will help you keep your fork up to date with the main sunpy repository.
+    $ git remote add upstream https://github.com/sunpy/sunpy.git
+    # These sync up tags between your fork and the main sunpy repository.
+    # This is important to ensure your local version is correct.
+    $ git fetch --tags upstream
+    $ git push -f --tags origin main
```
ayshih, line 173:
```suggestion
    # This retrieves the tags from the main sunpy repository, which are used for determining the version of your fork.
    $ git fetch --tags upstream
```
I see no reason to tell the users to push upstream's tags to their own fork.  That's not necessary for the version determination to work, and could conceivably cause confusion in their future.

ayshih (changes requested):

docs/dev_guide/contents/newcomers.rst
nabobalis: Hmm, I understand but If I view this as making sure their fork is set up if they ever need to clone again, I would lean towards adding the push to their fork.

wtbarnes: Could clarify by just adding a comment that explains why the push command is there.

ayshih: I'm not convinced.

- It is not necessary for the user to get a proper version number, so it is distracting and essentially "dead weight" for this guide.
- It can be useful for a user who wants to install their fork but does not want to connect to upstream as a remote, but that's hardly the kind of development environment we are interested in supporting.
- If the user's `git fetch` does not have the default behavior of getting tags, then the tags on their fork can become stale and inconsistent from the tags on upstream.
- It is glaringly obvious when the user has no tags in their local repo because the version number will be `0.1...`, but it will be more difficult to tell if a user has a stale set of tags.

nabobalis: Fair enough, I stand down.

ayshih (approved):
LGTM

Issue #7166, opened 2023-08-25
Make GenericMap.plot() arguments keyword only


dstansby:
### Describe the feature

See https://github.com/sunpy/sunpy/pull/7163 for context - it would be good to make the arguments to `GenericMap.plot()` keyword only to avoid confusion.

Cadair:
@ayshih mentioned in chat the other day that we should also validate that `annotate=` has type bool as then that would catch some errors where people pass the wrong positional argument until the deprecation period expires.

Cadair:
Haha great. I am not paying much attention to work ATM 😛

ayshih:
Closing because #7421 put it into deprecation

PR #7418, opened 2024-01-28
Made GenericMap.plot() arguments keyword only #7166

yeppAniket:
## PR Description
Made the GenericMap.plot() accept keywords only as arguments, As discussed in #7166 and #7163.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -2421,7 +2421,7 @@ class GenericMap(NDData):
         return figure
 
     @u.quantity_input
-    def plot(self, annotate=True, axes=None, title=True, autoalign=False,
+    def plot(self,*, annotate=True, axes=None, title=True, autoalign=False,
              clip_interval: u.percent = None, **imshow_kwargs):
         """
         Plots the map object using matplotlib, in a method equivalent


nabobalis:
Thanks for the PR @yeppAniket.

For this to work, we have to add warnings first, this change can not happen without forewarning.
If you have a look at https://github.com/sunpy/sunpy/pull/6406/files for example, that will show you what you should add.

yeppAniket:
> For this to work, we have to add warnings first, this change can not happen without forewarning. 
So I need to add a depreciation warning and update the changelog, right?

nabobalis:
> > For this to work, we have to add warnings first, this change can not happen without forewarning.
> 
> So I need to add a depreciation warning and update the changelog, right?

Yes.

yeppAniket:
> Yes.

from sunpy.util.exceptions import SunpyDeprecationWarning

def test_deprecated_positional_args(map_object, axes):
    with pytest.warns(SunpyDeprecationWarning, match=r"Pass annotate=True as keyword args"):
        map_object.plot(axes=axes, annotate=True)

    with pytest.warns(SunpyDeprecationWarning, match=r"Pass title='interpolation' as keyword args"):
        map_object.plot(axes=axes, title='interpolation')

    with pytest.warns(SunpyDeprecationWarning, match=r"Pass annotate=True, title='interpolation' as keyword args"):
        map_object.plot(axes=axes, annotate=True, title='interpolation')

    with pytest.warns(SunpyDeprecationWarning, match=r"Pass clip_interval=(lower, upper) as keyword args"):
        map_object.plot(axes=axes, clip_interval=(10, 90))

    with pytest.warns(SunpyDeprecationWarning, match=r"Pass autoalign='pcolormesh' as keyword args"):
        map_object.plot(axes=axes, autoalign='pcolormesh')
        
Something like this for the warnings?

nabobalis:
Yeah, you do not need to test it for every keyword, one will be enough.

nabobalis:
Superceded by https://github.com/sunpy/sunpy/pull/7419

Issue #7259, opened 2023-10-25
Maps with 8-bit unsigned integer data should probably not set `norm` to `None`


ayshih:
There is very old code (from #1432) that sets the `norm` to `None` for maps with 8-bit unsigned integer data (uint8, i.e., 0 to 255):

https://github.com/sunpy/sunpy/blob/5bffcc28214dcfe1b4d40e811cacc6807243175a/sunpy/map/mapbase.py#L236-L237

Presumably this is done because 8-bit unsigned integer data maps directly to the 8-bit color scale for `matplotlib`.  However, this causes later code (written much later) to fail because it assumes that the `norm` keyword exists only if it is actually a normalizer, e.g., in `plot()`:

https://github.com/sunpy/sunpy/blob/5bffcc28214dcfe1b4d40e811cacc6807243175a/sunpy/map/mapbase.py#L2529-L2532

So, possible solutions include:

- Don't special-case the setting of `norm` for uint8 data (i.e., treat it the same as any other data type)
- Don't set `norm` at all for uint8 data (i.e., `norm` will not be set in `plot_settings`)
- Add special code in `plot()` specific to uint8 data

Choosing between the options depends on exactly how we would like uint8 data to be visualized.

ayshih:
I guess this can be closed since #7422 was merged, although I don't recall us actually having decided that was the preferred way to go. =)  I guess we'll see if anyone complains...

PR #7422, opened 2024-01-30
If data is of uint8, then set plot_settings and norm accordingly

enthusiastic2003:
#7259 If the datatype of the map is uint8, then don't bother creating norm, and create plot_settings accordingly, without the 'norm' key. If data is of any other datatype, then create norm and create plot_settings with the 'norm' key.


.gitignore
```diff
@@ -5,7 +5,6 @@ __pycache__/
 *.py[cod]
 *$py.class
 tmp/
-
```
nabobalis, line 8: This hchange should be reverted.

sunpy/map/mapbase.py
```diff
@@ -235,19 +235,22 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
...
+            self.plot_settings = {'cmap': 'gray',
+                                'norm': norm,
+                                'interpolation': 'nearest',
+                                'origin': 'lower'
+                                }
+        else:
+            self.plot_settings = {'cmap': 'gray',
+                                'interpolation': 'nearest',
+                                'origin': 'lower'
+                                }
```
nabobalis, line 253: Is this not repeating itself?

nabobalis: Just seems very repetitive, can we not improve on that?

sunpy/map/mapbase.py
```diff
@@ -235,19 +235,22 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
         # TODO: This should be a function of the header, not of the map
         self._validate_meta()
 
-        if self.dtype == np.uint8:
-            norm = None
-        else:
+        if self.dtype != np.uint8:
```
ayshih, line 238:
```suggestion
        self.plot_settings = {'cmap': 'gray',
                            'interpolation': 'nearest',
                            'origin': 'lower'
                            }
        if self.dtype != np.uint8:
```

sunpy/map/mapbase.py
```diff
@@ -235,19 +235,22 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
...
             # Put import here to reduce sunpy.map import time
             from matplotlib import colors
             norm = colors.Normalize()
 
         # Visualization attributes
```
ayshih, line 243:
```suggestion
            self.plot_settings['norm'] = colors.Normalize()
```

sunpy/map/mapbase.py
ayshih, line 253:
```suggestion
```

ayshih (commented):
This is one way to detangle the logic and remove the code duplication

ayshih:
As with #7371, I'd like to see developers first sharing opinions on #7259 about whether we should be continuing to treat uint8 data differently before merging this PR

enthusiastic2003:
@nabobalis , Should i add a unit test case?

nabobalis:
Yes please.

nabobalis (approved):

PR #7202, opened 2023-09-16
Remove old limb drawing function

dstansby:
Deprecated in 4.1, so time to 🗑️

PR #7427, opened 2024-01-31
Refine GenericMap.draw_contours() for User-Provided Transformations 

### LINKED ISSUE #7191: `GenericMap.draw_contours()`: don't run internal transform code if `transform` keyword is provided

Mubin17:
Fixes #7191: Refines GenericMap.draw_contours() to internally transform the map to the axes WCS only when the transform keyword argument is not provided.

Note: Following PR builds up the existing PR #7207 which is lacking a figure unit test. I have made an attempt to add a figure unit test and will eventually update figure_hashes.

Also as per SunPy testing guidelines, I've added a fake hash to the JSON files. The real hash will be provided by CircleCi during the PR run, and it will be used to replace the placeholder hash.


Mubin17:
> Thanks for the PR @Mubin17.
> 
> If you used the code from #7202, I would prefer if you use that commit as the basis for this and then update items on top of that. This way we keep track of contributions.

Thanks for the suggestion @nabobalis, will keep that in mind for future PR's.

sunpy/map/mapbase.py
```diff
@@ -2355,7 +2355,11 @@ def draw_contours(self, levels, axes=None, **contour_args):
         # We do this instead of using the `transform` keyword argument so that Matplotlib does not
         # get confused about the bounds of the contours
         if self.wcs is not axes.wcs:
-            transform = axes.get_transform(self.wcs) - axes.transData  # pixel->pixel transform
+            if "transfrom" in contour_args:
```
nabobalis, line 2358:
```suggestion
            if "transform" in contour_args:
```

sunpy/map/tests/test_mapbase.py
```diff
@@ -1546,6 +1546,14 @@ def test_rotation_rect_pixelated_data(aia171_test_map):
     rect_rot_map = rect_map.rotate(30 * u.deg)
     rect_rot_map.peek()
 
+@figure_test
+def test_draw_contours_with_transform(aia171_test_map):
+    aia_map = sunpy.map.Map(aia171_test_map)
+    fig = plt.figure(figsize=(8, 4))
+    ax = fig.add_subplot(projection=aia_map)
+    aia_map.plot(axes=ax, clip_interval=(1, 99.99)*u.percent)
+    aia_map.draw_contours([-2,2]*u.percent, transform=ax.get_transform('pixel'))
```
nabobalis, line 1555: Can you try to replace the transfrom with something like https://matplotlib.org/stable/api/transformations.html#matplotlib.transforms.Affine2D

We want to make sure the transform works and picking one that will produce bad looking contours should make the figure test more useful.

sunpy/map/tests/test_mapbase.py
```diff
@@ -1546,6 +1546,14 @@ def test_rotation_rect_pixelated_data(aia171_test_map):
     rect_rot_map = rect_map.rotate(30 * u.deg)
     rect_rot_map.peek()
 
+@figure_test
+def test_draw_contours_with_transform(aia171_test_map):
```
nabobalis, line 1550: Can you change this test to do the following:

A two-panel figure test, where the first panel is what is already written, and the second one composites in a scaling transform (e.g., by a factor of 2) like adding in https://matplotlib.org/stable/api/transformations.html#matplotlib.transforms.Affine2D

The contours will look "wrong" on the second panel, but the purpose is to confirm that they are "wrong" in the correct fashion.

Mubin17: How does this look:

```python
import astropy.units as u
import sunpy.map
from matplotlib.transforms import Affine2D

fig, axes = plt.subplots(1, 2, figsize=(16, 4))

for ax in axes:
    ax.set_axis_off()

# Panel 1: Original contours
ax1 = axes[0]
ax1 = fig.add_subplot(1, 2, 1, projection=aia_map)
aia_map.plot(axes=ax1, clip_interval=(1, 99.99)*u.percent)
aia_map.draw_contours([-1, 1]*u.percent, transform=ax1.get_transform('pixel'))

# Panel 2: Contours with scaling transform
ax2 = axes[1]
ax2 = fig.add_subplot(1, 2, 2, projection=aia_map)
scaling_transform = Affine2D().scale(2, 2)
composite_transform = ax2.get_transform('pixel') + scaling_transform
aia_map.plot(axes=ax2, clip_interval=(1, 99.99)*u.percent)
aia_map.draw_contours([-1, 1]*u.percent, transform=composite_transform)

# # Change x and y limits by a factor of two
# ax2.set_xlim(ax2.get_xlim()[0]*3, ax2.get_xlim()[1]*3)
# ax2.set_ylim(ax2.get_ylim()[0]*3, ax2.get_ylim()[1]*3)

ax2.set_title('Contours with Scaling Transform')

plt.show()
```
Produces:
Screenshot 2024-02-02 at 2 13 17 PM

If this looks good I will update the tests accordingly.

ayshih: Note that you can't actually see the contours at all on the right-hand panel.  That's because the order of the summation for the composite transform is wrong.  Here's a fixed version of your script (with the plotting code cleaned up as well):
```python
from matplotlib.transforms import Affine2D

import astropy.units as u

import sunpy.map
from sunpy.data.sample import AIA_171_IMAGE

fig = plt.figure(figsize=(11, 4))

# Panel 1: Original contours
ax1 = fig.add_subplot(1, 2, 1, projection=aia_map)
aia_map.plot(axes=ax1, clip_interval=(1, 99.99)*u.percent)
aia_map.draw_contours([-1, 1]*u.percent, transform=ax1.get_transform('pixel'))

# Panel 2: Contours with scaling transform
ax2 = fig.add_subplot(1, 2, 2, projection=aia_map)
scaling_transform = Affine2D().scale(2, 2)
composite_transform = scaling_transform + ax2.get_transform('pixel')
aia_map.plot(axes=ax2, clip_interval=(1, 99.99)*u.percent)
aia_map.draw_contours([-1, 1]*u.percent, transform=composite_transform)

# # Change x and y limits by a factor of two
# ax2.set_xlim(ax2.get_xlim()[0]*3, ax2.get_xlim()[1]*3)
# ax2.set_ylim(ax2.get_ylim()[0]*3, ax2.get_ylim()[1]*3)

ax2.set_title('Contours with Scaling Transform')

plt.show()
```
Figure_1

ayshih: However, the bigger issue is that this test approach isn't good enough, because the figure actually plots the same irrespective of the code change in this PR.  Let me get back to you with a meaningful test.

ayshih: Okay, here's a useful test:
```python
from matplotlib.transforms import Affine2D

import astropy.units as u

import sunpy.map
from sunpy.data.sample import AIA_171_IMAGE, HMI_LOS_IMAGE

hmi_map = sunpy.map.Map(HMI_LOS_IMAGE)
fig = plt.figure(figsize=(16, 4))

# Panel 1: Implicit transfrom
ax1 = fig.add_subplot(1, 3, 1, projection=aia_map)
aia_map.plot(axes=ax1, clip_interval=(1, 99.99)*u.percent)
hmi_map.draw_contours([-10, 10]*u.percent)
ax1.set_title('Default, correct behavior')

# Panel 2: Explicit transform
ax2 = fig.add_subplot(1, 3, 2, projection=aia_map)
aia_map.plot(axes=ax2, clip_interval=(1, 99.99)*u.percent)
hmi_map.draw_contours([-10, 10]*u.percent, transform=ax2.get_transform(hmi_map.wcs))
ax2.set_title('Explicitly specifying the correct transform')

# Panel 3: Explicit transform with wacky rotation
ax3 = fig.add_subplot(1, 3, 3, projection=aia_map)
rotate_transform = Affine2D().rotate_deg_around(512, 512, 90)
composite_transform = rotate_transform + ax3.get_transform(hmi_map.wcs)
aia_map.plot(axes=ax3, clip_interval=(1, 99.99)*u.percent)
hmi_map.draw_contours([-10, 10]*u.percent, transform=composite_transform)
ax3.set_title('Contours rotated by 90 deg CCW')

plt.show()
```
## Before this PR
before
## After this PR
after

wtbarnes: @Mubin17 are you able to modify the figure test using the suggested three panel plot above?

sunpy/map/tests/test_mapbase.py
```diff
@@ -1555,6 +1556,36 @@ def test_rotation_rect_pixelated_data(aia171_test_map):
     rect_rot_map = rect_map.rotate(30 * u.deg)
     rect_rot_map.peek()
 
+@figure_test
+def test_draw_contours_with_transform():
+    # TODO: Use test data
+    from sunpy.data.sample import AIA_171_IMAGE, HMI_LOS_IMAGE
```
nabobalis, line 1562: I think we would need to solve this?

nabobalis (approved):

sunpy/map/tests/test_mapbase.py
wtbarnes: Is there a reason to not use a fixture here?

wtbarnes (changes requested):
the figure test should use fixtures for the maps. Otherwise this looks good to me.

sunpy/map/tests/test_mapbase.py
nabobalis: I can turn it into a fixture, but should we be using the sample data for this? 

I can try with the test aia image and see how well it works first?

nabobalis: Oh we lack a hmi test file. 

I will just turn them into a fixture

nabobalis: I setup two fixtures instead.

nabobalis:
Thanks for the PR @Mubin17

PR #7432, opened 2024-02-02
Update schemas for upcoming ASDF standard 1.6.0

braingram:
## PR Description

asdf is working on making the current development version of the ASDF standard (1.6.0) stable (see https://github.com/asdf-format/asdf/pull/1744 for details). As the 1.6.0 version contains new versions of a few core schemas (including [ndarray](https://github.com/asdf-format/asdf-standard/blob/main/resources/schemas/stsci.edu/asdf/core/ndarray-1.1.0.yaml)) many dependent schemas will require updates to be compatible with the new version (if they contain `$ref` or `tag` links to core schemas).

This PR updates the `sunpy` schemas to be compatible with ASDF standard 1.6.0. One example of a required update (for ASDF 1.6.0 compatibility) is updating the `ndarray` `tag`s found in several `sunpy` schemas for [ndarray-1.1.0](https://github.com/asdf-format/asdf-standard/blob/main/resources/schemas/stsci.edu/asdf/core/ndarray-1.1.0.yaml). For example, the `ndarray-1.0.0` tag in  `generic_map`:
https://github.com/sunpy/sunpy/blob/fb9c9599ae5efa6500e8a9bdba1ed229ad88ce12/sunpy/io/special/asdf/resources/schemas/generic_map-1.1.0.yaml#L17
is updated in this PR to a wildcard expression:
```yaml
 tag: "tag:stsci.edu:asdf/core/ndarray-1.*
```
to accept either `ndarray-1.0.0` `ndarray-1.1.0` (or any future `ndarray` tag with major version `1`).

A new manifest and extension (`sunpy-1.2.0`) were added to use the new schemas.

Additionally a few schemas were updated so their `id` matches the `uri` used to register the schema with asdf.

See https://github.com/asdf-format/asdf/actions/runs/8345268069/job/22843625585?pr=1744 for a test run where ASDF standard 1.6.0 was enabled as the default and the sunpy tests were run with this PR.


nabobalis:
Would you recommend backporting this to the release versions?

braingram:
> Would you recommend backporting this to the release versions?

Good question. Naively I think this should be safe to backport (as it's only adding new schemas and the new manifest). That should allow the released versions to write out ASDF 1.6.0 compliant files (when asdf switches to that as the default, this switch is tentatively planned for asdf 4.0 which is likely a few months out).

Is there anything I can do to make this easier to backport?

Also, the docs failure looks to be unrelated but I'm not that familiar with it to be confident:
https://github.com/sunpy/sunpy/actions/runs/8364240849/job/22899279125?pr=7432#step:10:313
It looks like it's failing from warnings:
```
WARNING: the pydata_sphinx_theme extension is not safe for parallel writing
WARNING: doing serial write
```

braingram:
Thanks! I will open an issue over in asdf milestoned with 4.0 to verify that sunpy 6.0 is out by then.

sunpy/io/special/asdf/resources/schemas/geocentricearthequatorial-1.0.0.yaml
```diff
@@ -1,14 +1,15 @@
...
-id: "asdf://sunpy.org/schemas/geocentricearthequatorial-1.0.0"
+id: "asdf://sunpy.org/sunpy/schemas/geocentricearthequatorial-1.0.0"
 
 title: |
   Represents a Geocentric Earth Equatorial (GEI) coordinate object from SunPy
 
 examples:
   -
     - A GeocentricEarthEquatorial frame without data
+    - asdf-standard-1.5.0
```
Cadair, line 12: Is this something which is affecting how the tests verify this?

sunpy/io/special/asdf/resources/schemas/generic_map-1.0.0.yaml
```diff
@@ -1,7 +1,7 @@
 %YAML 1.1
 ---
 $schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
-id: "asdf://sunpy.org/schemas/generic_map-1.0.0"
+id: "asdf://sunpy.org/sunpy/schemas/generic_map-1.0.0"
```
Cadair, line 4: I assume changing this on the old schemas can't cause a problem? There's no way for external things to be referencing it?

Cadair (approved):

sunpy/io/special/asdf/resources/schemas/geocentricearthequatorial-1.0.0.yaml
braingram: I went looking for documentation on this and found none :facepalm:

This tells the `pytest-asdf` plugin to run this example using ASDF standard 1.5.0. Here's a link to the code that handles this (although it's not that helpful):
https://github.com/asdf-format/asdf/blob/48c5c57bb8a471a3bad7f2d5791a9576422abeb5/pytest_asdf/plugin.py#L184

This (and the other similar additions) is needed as the example uses a tag that is not in ASDF standad 1.6.0. In this example, the following time tag is used:
```yaml
equinox: !time/time-1.1.0
```
which exists in 1.5.0 but not in 1.6.0 (which uses `time-1.2.0` instead). Without this change, when 1.6.0 becomes the default, this example will start failing because `time-1.1.0` is not a valid 1.6.0 tag.

Let me know if more details are helpful or if you have any questions.

sunpy/io/special/asdf/resources/schemas/generic_map-1.0.0.yaml
braingram: I don't foresee this causing any issues.

The `id` here is effectively unused by asdf at the moment. asdf will instead lookup schemas based on the uri used to register them (based on the path and the prefix provided to the `DirectoryResourceMapping`). However, if this schema contained a local `$ref` it's possible that `jsonschema` would have issues as it likely uses the `id` in the schema to resolve the ref.

nabobalis (approved):

sunpy/io/special/asdf/resources/schemas/geocentricearthequatorial-1.0.0.yaml
Cadair: That is what I assumed, thanks for confirming.

nabobalis:
I have manually backported these

braingram:
Thanks again!

braingram:
@nabobalis I release asdf 3.3.0 which introduces a warning for manifests with ids that don't match the registered uri (one of the bugs fixed in this PR). @Cadair noted this new warning in the dkist tests:
https://github.com/DKISTDC/dkist/actions/runs/9936292499/job/27444219001?pr=419#step:10:581
Is it possible to make a sunpy release with the backported changes in this PR? Thanks!

nabobalis:
> @nabobalis I release asdf 3.3.0 which introduces a warning for manifests with ids that don't match the registered uri (one of the bugs fixed in this PR). @Cadair noted this new warning in the dkist tests: [DKISTDC/dkist/actions/runs/9936292499/job/27444219001?pr=419#step:10:581](https://github.com/DKISTDC/dkist/actions/runs/9936292499/job/27444219001?pr=419#step:10:581) Is it possible to make a sunpy release with the backported changes in this PR? Thanks!

Sure.

nabobalis:
The release of 5.1.5 with this patch is out on pypi.

braingram:
Thank you!

PR #7438, opened 2024-02-07
Remove two gallery examples that use mpl-animators

nabobalis:
I think these belong in the mpl-animators gallery.

TODO
- [x] Create gallery: See https://github.com/sunpy/mpl-animators/pull/37


wtbarnes:
I agree in theory, but I object until there is an mpl-animators gallery

nabobalis:
I will do that only after people are ok with this in principle. 
IF so, I will work on the gallery.

nabobalis:
Gallery is DONE

wtbarnes (approved):
Cool, looks good to me. I'm onboard for anything that makes the sunpy gallery more manageable.

PR #7441, opened 2024-02-09
Reduce the tox env_list to the common combinations

Cadair:
Other combinations are possible, but not run by default with a plain `tox run`.


alasdairwilson (approved):
👍

PR #7451, opened 2024-02-16
Add GONG H-Alpha map source

### LINKED ISSUE #6656: No map source to handle GONG H-alpha data
### Describe the bug
Originally raised by nawinnga in the sunpy chat room.

There is no map source to do the necessary translation from the GONG H-alpha headers to standard meta data Map can use related to #6653 and #6655.

It doesn't look like its trivial to find/extract/calculate the necessary WCS information (e.g. PCi_j + CDELT or CDi_j ...) without some reference to what the fits headers mean some info on [this NSO page](https://nso.edu/data/nisp-data/h-alpha/)

### To Reproduce

```
from astropy.io import fits
import sunpy

# to get around  issue #6655
hdul = fits.open('ftp://gong2.nso.edu/HA/haf/202203/20220318/20220318000050Bh.fits.fz')
sunpy.map.Map((hdul[1].data, hdul[1].header))

Map((hdul[1].data, hdul[1].header))
  File "<stdin>", line 1, in <module>
  File "/Users/shane/.virtualenvs/temp/lib/python3.9/site-packages/sunpy/map/map_factory.py", line 331, in __call__
    new_map = self._check_registered_widgets(data, meta, **kwargs)
  File "/Users/shane/.virtualenvs/temp/lib/python3.9/site-packages/sunpy/map/map_factory.py", line 381, in _check_registered_widgets
    return WidgetType(data, meta, **kwargs)
  File "/Users/shane/.virtualenvs/temp/lib/python3.9/site-packages/sunpy/map/mapbase.py", line 232, in __init__
    self._validate_meta()
  File "/Users/shane/.virtualenvs/temp/lib/python3.9/site-packages/sunpy/map/mapbase.py", line 1409, in _validate_meta
    raise MapMetaValidationError('\n'.join(err_message))
sunpy.map.mapbase.MapMetaValidationError: Image coordinate units for axis 1 not present in metadata.
Image coordinate units for axis 2 not present in metadata.
See https://docs.sunpy.org/en/stable/code_ref/map.html#fixing-map-metadata for instructions on how to add missing metadata.
```
Setting the cdelt and cunit at least get lets the map be loaded but the WCS is still non-standard
```
hdul[1].header['cdelt1'] = 1
hdul[1].header['cdelt2'] = 1
hdul[1].header['cunit1'] = 'arcsec'
hdul[1].header['cunit2'] = 'arcsec'

mm = Map((hdul[1].data, hdul[1].header))
WARNING: SunpyMetadataWarning: Missing metadata for observer: assuming Earth-based observer.
For frame 'heliographic_stonyhurst' the following metadata is missing: hgln_obs,hglt_obs,dsun_obs
For frame 'heliographic_carrington' the following metadata is missing: crlt_obs,dsun_obs,crln_obs
 [sunpy.map.mapbase]
WARNING: SunpyUserWarning: Could not determine coordinate frame from map metadata.
Could not determine celestial frame corresponding to the specified WCS object [sunpy.map.mapbase]
<sunpy.map.mapbase.GenericMap object at 0x1297eb160>
SunPy Map
---------
Observatory:		 NSO-GONG
Instrument:
Detector:
Measurement:		 6562.808
Wavelength:		 6562.808
Observation Date:	 2022-03-18 00:00:50
Exposure Time:		 Unknown
Dimension:		 [2048. 2048.] pix
Coordinate System:	 Unknown
Scale:			 [1. 1.] arcsec / pix
Reference Pixel:	 [1023. 1023.] pix
Reference Coord:	 [0. 0.] arcsec
array([[0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       ...,
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0]], dtype=int16)
```

### Screenshots

_No response_

### System Details

OS: Mac OS 10.16 Arch: 64bit, (i386)
python3.9
sunpy: 4.1.0
astropy: 5.1.1
numpy: 1.23.5

### Installation method

pip

samaloney:
## PR Description

Closes #6656 by adding a GONG H-alpha map source.

TODO:
- [ ] Add changelog
- [ ] Update docs and add links to source material
- [ ] Add tests
- [ ] User Earth lat, lon header to set observatory location

#### From sunpy with this PR
gong_sunpy


sunpy/map/sources/gong.py
```diff
@@ -40,7 +51,7 @@ class GONGSynopticMap(GenericMap):
     @classmethod
     def is_datasource_for(cls, data, header, **kwargs):
         return (str(header.get('TELESCOP', '')).endswith('GONG') and
-                str(header.get('CTYPE1', '').startswith('CRLN')))
+                str(header.get('CTYPE1', '')).startswith('CRLN'))
```
samaloney, line 54: Previously this was returning a string `'True'` and happened to work as there were not other GONG clients.

samaloney:
Because of the asdf issue with the `.fz` extension you have to open these maps with the well hidden filetype keyword `Map('/path/to/file/20240216000002Bh.fits.fz', filetype='fits')`

sunpy/map/sources/gong.py
```diff
@@ -4,15 +4,25 @@
...
+__all__ = ['GONGSynopticMap', 'GONGHAlphaMap']
 
 from sunpy.map.mapbase import SpatialPair
 
+_SITE_NAMES = {
+    'LE': 'Learmonth',
+    'UD': 'Udaipur',
+    'TD': 'El Teide',
+    'CT': 'Cerro Tololo',
+    # '': 'Engineering site', TODO check fits header
```
nabobalis, line 22: Do we remove this?

sunpy/map/sources/gong.py
```diff
@@ -61,3 +71,56 @@ def spatial_units(self):
     @property
     def observer_coordinate(self):
         return get_earth(self.date)
+
+
+class GONGHAlphaMap(GenericMap):
```
nabobalis, line 76: Not sure how I feel about the HAlpha, could it be Halpha? .

samaloney: I don't have strong opinion either way, so happy to change.

sunpy/map/sources/gong.py
```diff
@@ -61,3 +71,56 @@ def spatial_units(self):
...
+        return SpatialPair(u.deg, u.deg)
+
+    @property
+    def earth_location(self):
+        """Location of the observatory on Earth"""
+        return EarthLocation.from_geodetic(lat=self.meta['site-lat'] * u.deg, lon=self.meta['site-lon'] * u.deg)
+
+    @property
+    def observer_coordinate(self):
+        return SkyCoord(self.earth_location.itrs).heliographic_stonyhurst
```
nabobalis, line 126: Do we have a test for this?

samaloney: No but I can add one but there is a test for `self.earth_location` and if this were to break it would be in `SkyCoord` or `EarthLocation`

sunpy/map/sources/gong.py
```diff
@@ -61,3 +71,56 @@ def spatial_units(self):
...
+        super().__init__(data, header, **kwargs)
+
+    @classmethod
+    def is_datasource_for(cls, data, header, **kwargs):
+        return (str(header.get('TELESCOP', '')).endswith('GONG') and
+                str(header.get('IMTYPE', '')).startswith('H-ALPHA'))
+
+    @property
+    def date(self):
+        return Time(self.meta.get('date-obs'))
```
wtbarnes, line 103: This is the default behavior in `GenericMap` so this can be removed.

sunpy/map/sources/gong.py
```diff
@@ -61,3 +71,56 @@ def spatial_units(self):
...
+    def nickname(self):
+        site = _SITE_NAMES.get(self.meta.get("sitename", ""), "UNKNOWN")
+        return f'{self.observatory}, {site}'
+
+    @property
+    def spatial_units(self):
+        return SpatialPair(u.deg, u.deg)
+
+    @property
+    def earth_location(self):
```
wtbarnes, line 120:
```suggestion
    def _earth_location(self):
```

Better to make this private since it is not a property that is shared by all maps.

sunpy/map/sources/gong.py
```diff
@@ -61,3 +71,56 @@ def spatial_units(self):
...
+                str(header.get('IMTYPE', '')).startswith('H-ALPHA'))
+
+    @property
+    def date(self):
+        return Time(self.meta.get('date-obs'))
+
+    @property
+    def scale(self):
+        return SpatialPair((self.meta['SOLAR-R'] * u.arcsec) / (self.meta['FNDLMBMI'] * u.pixel),
+                           (self.meta['SOLAR-R'] * u.arcsec) / (self.meta['FNDLMBMA'] * u.pixel))
```
wtbarnes, line 108:
```suggestion
        solar_r = self.meta['SOLAR-R'] * u.arcsec
        return SpatialPair(solar_r / (self.meta['FNDLMBMI'] * u.pixel),
                           solar_r / (self.meta['FNDLMBMA'] * u.pixel))
```

Should we have some way to bail out if these keys don't exist? I don't know if that's a reasonable case to consider.

wtbarnes (changes requested):

sunpy/map/sources/gong.py
```diff
@@ -61,3 +71,56 @@ def spatial_units(self):
     @property
     def observer_coordinate(self):
         return get_earth(self.date)
+
+
+class GONGHalphaMap(GenericMap):
```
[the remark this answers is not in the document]
samaloney: So as far I as I know GONG produces high cadence H-alpha, lower cadence ~1 hours magnetograms, and the carrington rotation magnetic maps.

samaloney:
I think the test fail is related to this issue astropy/astropy/issues/16128

sunpy/map/sources/gong.py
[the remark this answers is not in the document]
wtbarnes: I guess using SDO as the example, we have separate map source classes for the different instruments on the same observatory so it would make sense to follow that same logic here, eg have an H-alpha source and a magnetogram source.

sunpy/map/sources/gong.py
wtbarnes: Is there a reason this can't be removed?

wtbarnes:
> Yea so if I extract the test command and run in a terminal it works I guess it could be something to do with `remote_data` but I tried adding that marker and it didn't work either?
> 
> ```python
> import astropy.units as u
> from numpy.testing import assert_equal
> from sunpy.data.test import get_dummy_map_from_header, get_test_filepath
> 
> gm = get_dummy_map_from_header(get_test_filepath('gong_halpha.header'))
> assert_equal(gm.observer_coordinate.data.xyz, 
>              [146712246479.363, -5563586.169750214, -17605285536.73928] * u.m)
> ```

I'm able to reproduce this test fail locally. The reason is because the last date in the IERS table that astropy is grabbing is 1/12/2024 and the date of this observation is 2/16/2024. Thus, its throwing a warning that you're beyond the last date in the IERS table and your accuracy may be degraded.

sunpy/map/sources/gong.py
samaloney: No I thought it was some how causing the error so added it back.

samaloney:
Yea but is that specific to test runs e.g. strict remote data runner as if I run it from local shell it works?

```
    ~/Projects/sunpy-dev    feat-gong-halpha *3 ?1                                                                                                                                                                                                                                                                                                 ✔  sunpy-dev 
    ~/Projects/sunpy-dev    feat-gong-halpha *3 ?1  ipython                                                                                                                                                                                                                                                                                        ✔  sunpy-dev 
Python 3.11.7 (main, Dec  4 2023, 18:10:11) [Clang 15.0.0 (clang-1500.1.0.2.5)]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.21.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import astropy.units as u
   ...: from numpy.testing import assert_equal
   ...: from sunpy.data.test import get_dummy_map_from_header, get_test_filepath
   ...:
   ...: gm = get_dummy_map_from_header(get_test_filepath('gong_halpha.header'))
   ...: assert_equal(gm.observer_coordinate.data.xyz,
   ...:              [146712246479.363, -5563586.169750214, -17605285536.73928] * u.m)

In [2]:
```

wtbarnes:
Ok I think that should fix the test. Let me know what you think about that. I changed the `assert_equal` to an `isclose` with a tolerance of 1e-9. If we really need the `assert_equal` then we can interrogate what the actual differences are between the two.

wtbarnes:
> Yea but is that specific to test runs e.g. strict remote data runner as if I run it from local shell it works?
> 
> ```
>     ~/Projects/sunpy-dev    feat-gong-halpha *3 ?1                                                                                                                                                                                                                                                                                                 ✔  sunpy-dev 
>     ~/Projects/sunpy-dev    feat-gong-halpha *3 ?1  ipython                                                                                                                                                                                                                                                                                        ✔  sunpy-dev 
> Python 3.11.7 (main, Dec  4 2023, 18:10:11) [Clang 15.0.0 (clang-1500.1.0.2.5)]
> Type 'copyright', 'credits' or 'license' for more information
> IPython 8.21.0 -- An enhanced Interactive Python. Type '?' for help.
> 
> In [1]: import astropy.units as u
>    ...: from numpy.testing import assert_equal
>    ...: from sunpy.data.test import get_dummy_map_from_header, get_test_filepath
>    ...:
>    ...: gm = get_dummy_map_from_header(get_test_filepath('gong_halpha.header'))
>    ...: assert_equal(gm.observer_coordinate.data.xyz,
>    ...:              [146712246479.363, -5563586.169750214, -17605285536.73928] * u.m)
> 
> In [2]:
> ```

Either you're silencing the warning or using a different IERS table (that has a more recent date). Can you check which table you're using?

samaloney:
Hum I cleared the downloaded cache in the process of trying to reproduce the errors so maybe that's related.

```
from astropy.utils import iers

iers.IERS_A_FILE
'finals2000A.all'

iers.IERS_B_FILE
'/Users/sm/.virtualenvs/sunpy-dev/lib/python3.11/site-packages/astropy/utils/iers/data/eopc04_IAU2000.62-now'
```

wtbarnes:
Maybe? Whatever I did also I did not fix the issue on the CI though did fix the fail for me locally. 

Yeah, maybe just using an older file is the easiest solution. People will of course run into this if they use the newer files, but it's probably not a big deal in terms of accuracy or they'll have a more up-to-date IERS table.

sunpy/map/sources/tests/test_gong_halpha_source.py
```diff
@@ -0,0 +1,56 @@
...
+
+def test_date(gong_halpha):
+    """Tests the date property of the GONGHalphaMap map."""
+    assert_equal(Time('2024-02-16T00:00:02'), gong_halpha.date)
+
+
+def test_scale(gong_halpha):
+    """Tests the scale property of the GONGHalphaMap map."""
+    assert_equal(gong_halpha.scale.axis1, 1.0794939681708389 * (u.arcsec / u.pix))
+    assert_equal(gong_halpha.scale.axis1, 1.0794939681708389 * (u.arcsec / u.pix))
```
wtbarnes, line 40: Should this be `axis2` instead?

wtbarnes:
Just to cross-reference, the actual underlying in astropy is now being discussed here: https://github.com/astropy/astropy/issues/13227#issuecomment-1986724516

The tl;dr is that astropy will ignore a just-downloaded IERS table if you have `iers.conf.auto_download = False` even though that table exists on your machine already.

sunpy/map/sources/tests/test_gong_halpha_source.py
```diff
@@ -0,0 +1,57 @@
...
+
+def test_earth_location(gong_halpha):
+    assert_equal(gong_halpha._earth_location.lat, 34.26032998167749*u.deg)
+    assert_equal(gong_halpha._earth_location.lon, -116.92141999386104*u.deg)
+
+
+def test_observer_coordinate(gong_halpha):
+    xyz_expected = [146712246479.363, -5563586.169750214, -17605285536.73928] * u.m
+    with pytest.warns(AstropyWarning, match='Tried to get polar motions for times after IERS data is valid.'):
+        with pytest.warns(AstropyWarning, match='times are outside of range covered by IERS table'):
```
nabobalis, line 56: Had to add an extra warning?

nabobalis (approved):

sunpy/map/sources/tests/test_gong_halpha_source.py
nabobalis: I broke oldest build do this.

```
>       with pytest.warns(AstropyWarning, match='Tried to get polar motions for times after IERS data is valid.'):
E       Failed: DID NOT WARN. No warnings of type (<class 'astropy.utils.exceptions.AstropyWarning'>,) were emitted. The list of emitted warnings is: [].
```

wtbarnes: Again I think we should just choose an older test header to avoid having to deal with this IERS behavior since an upstream fix is needed.

ayshih: Even though the upstream fix (astropy/astropy#16187) will likely be merged imminently, it probably won't be backported, so it will be a long while before it is in all versions of `astropy` we support.  The alternative to picking an older test header is to simply filter away warnings within the unit test.

ayshih: That is, the unit test don't actually care if there's a warning or not, so catching/matching with `pytest.warns()` is unnecessary.

ayshih: (And yes, there are two warnings, from two different part of the `astropy` codebase.)

nabobalis: Done.

alasdairwilson (approved):

wtbarnes (approved):

Issue #7450, opened 2024-02-16
Files downloaded from VSO have their filenames altered if they include more that one dot (a.b.fits -> a_b.fits)


samaloney:
### Describe the bug

Files downloaded from the VSO with file names containing more than one dot `a.b.fits` are save to files with altered filenames (`a_b.fits`) which can cause some confusion.

While investigating #6656 I noticed the file names are altered e.g. a url with file name of `20240209120002Uh.fits.fz` will become `20240209120002uh_fits.fz` when downloaded (I know the filename is not taken from the url).

I've tracked it down to the `slugify` function `slugify('20240209120002Uh.fits.fz')` -> `'20240209120002uh_fits.fz'`. I'm not sure if this is really bug with `slugify` or its use in sanitising possible filenames?

L326

https://github.com/sunpy/sunpy/blob/8349d7f8698da737b884faa81b1773ddd1adaf1f/sunpy/net/vso/vso.py#L290-L335

### To Reproduce

VSO seems to be down right now so can't get a full example but this shows the gist

```python
from sunpy.util.net import slugify

slugify('a.b.c.fits')

#'a_b_c.fits'
```

ViciousEagle03:
I think if we specifically pass the parameter of `delim = '.'` , when we call the function `slugify` as `slugify('a.b.c.fits' , delim='.')` , we will get the desired result for the cases where there are multiple `period '.'`.
So do you want the function `slugify` to default to `delim='.'` instead of delim='_'  or only for cases where there are multiple periods in the `FITS` filename?

nabobalis:
We would need to make sure we preserve the filenames with two dots when it comes to file extensions.

ViciousEagle03:
Could you please review my pull request? Thank you.

ayshih:
Why does `slugify()` even replace periods?  Periods are not typically a character that needs to be sanitized.  Can we simply take periods out of the replace list?

I'm not a fan of a fix that attempts to sometimes replace periods, because that invariably means false positives/negatives and maintenance burden.

samaloney:
I think a `slugify` and a `sanitise_filename` function would have different behaviours. It not totally clear what the original intent of `slugify` was and if that's really what is needed here. The example I found was for periods but there maybe other valid filename characters that are also being replace. I Filename sanitisation is as straightforward as it may seem.

ayshih:
[...]

PR #7453, opened 2024-02-18
Fix filename sanitization for downloaded files (do not replace periods, do not change case, and do not leave Unicode characters decomposed)

ViciousEagle03:
This PR address the issue #7450

### DIFF
--- a/sunpy/util/tests/test_net.py
+++ b/sunpy/util/tests/test_net.py
@@ -18,7 +18,7 @@ def test_content_disposition_unicode():
 
 
 def test_slugify():
-    assert sunpy.util.net.slugify("™bc") == "TMbc"  # Unicode normalization
+    assert sunpy.util.net.slugify("ä™") == "äTM"  # Unicode NFKC normalization
     assert sunpy.util.net.slugify("filegreg") == "filegreg"  # no file extension
     assert sunpy.util.net.slugify("filegreg.fits") == "filegreg.fits"  # one file extension
     assert sunpy.util.net.slugify("file.greg.fits") == "file.greg.fits"  # more than one apparent file extension
--- a/sunpy/util/tests/test_net.py
+++ b/sunpy/util/tests/test_net.py
@@ -19,12 +19,12 @@ def test_content_disposition_unicode():
 
 def test_slugify():
     assert sunpy.util.net.slugify("äb c", "b_c")
-    assert sunpy.util.net.slugify("file.greg.fits") == "file_greg.fits"
-    assert sunpy.util.net.slugify("file.greg.fits", "x") == "filexgreg.fits"
-    assert sunpy.util.net.slugify("file.name.fits.hdu") == "file_name.fits.hdu"
+    assert sunpy.util.net.slugify("file.greg.fits") == "file.greg.fits"
+    assert sunpy.util.net.slugify("file.greg.fits", "x") == "file.greg.fits"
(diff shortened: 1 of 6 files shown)


sunpy/util/tests/test_net.py
```diff
@@ -21,6 +21,7 @@ def test_slugify():
     assert sunpy.util.net.slugify("äb c", "b_c")
     assert sunpy.util.net.slugify("file.greg.fits") == "file_greg.fits"
     assert sunpy.util.net.slugify("file.greg.fits", "x") == "filexgreg.fits"
+    assert sunpy.util.net.slugify("file.name.fits.hdu") == "file_name.fits.hdu"
```
nabobalis, line 24: Can you add the examples from the issue so we can double check?

nabobalis:
Can you also add a changelog please.

sunpy/util/tests/test_net.py
ViciousEagle03: Sure i will add the example,

nabobalis:
Actually in line with https://github.com/sunpy/sunpy/issues/7450#issuecomment-1951436626, I think we should leave periods alone.

ViciousEagle03:
> Actually in line with [#7450 (comment)](https://github.com/sunpy/sunpy/issues/7450#issuecomment-1951436626), his point makes a lot of sense. I think we should leave the periods alone.
So should I remove my fix in `net.py` and remove the periods from the list that `slugify()` function replaces?

ViciousEagle03:
So should I remove my fix in `net.py` and remove the periods from the list that `slugify()` function replaces?

nabobalis:
> So should I remove my fix in `net.py` and remove the periods from the list that `slugify()` function replaces?

Yes please.

samaloney:
Thanks for the PR this addresses the specific issue and look good to me. 

It would be good to consider if could address the the root cause see the discussion on #7450

ViciousEagle03:
Sure, I will look into the root cause

ayshih:
I've directly committed to this PR to re-implement `slugify()` in a more straightforward way

ayshih:
For the Unicode fans out there, `slugify()` has been normalizing to NFKD form (decomposed by compatibility), but I feel it should normalize to NFKC form (decomposed by compatibility, then recomposed by canonical equivalence).  It's weird to me that it characters decomposed.  For example, "ä" becomes two Unicode characters ("a" + U+0308) instead of being left intact.  Any thoughts?

I'll note that there is currently a test line that looks intended to test the decomposition of "ä", but that line is bugged, so the output isn't tested.

nabobalis:
I would be in favour of NFKC

nabobalis:
Online Test failures look to be real.

One is a filename difference, and the other seems to be more results returned from the VSOClient?

ViciousEagle03:
Do you have any tasks for at the moment
Apologies for not being active last week I was occupied with my mid-terms

nabobalis:
> Do you have any tasks for me at the moment . Apologies for not being active last week I was occupied with my mid-terms

I think the online tests need updating but that is it.

nabobalis (approved):

nabobalis:
Thanks for the PR @ViciousEagle03 and @ayshih

nabobalis:
I decided not to backport this and add a whatsnew as I don't want to break previous behaviour for filenames for the released versions.

PR #7457, opened 2024-02-27
Upload to scientific-python-nightly-wheels for binary files

nabobalis:
Missed a job.


.github/workflows/ci.yml
```diff
@@ -191,9 +191,9 @@ jobs:
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@main
     with:
       upload_to_anaconda: ${{ (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') }}
-      anaconda_user: sunpy
+      anaconda_user: scientific-python-nightly-wheels
       anaconda_package: sunpy
-      anaconda_keep_n_latest: -1
+      anaconda_keep_n_latest: 1
```
nabobalis, line 196: I hope this will clean up?

matthewfeickert:
@nabobalis This is currently uploading both the sdist and wheels. Can you please upload wheels only?

nabobalis:
Ah! I have removed it now in the latest commit. Binary wheels only now!

matthewfeickert:
Great and thanks! :D

.github/workflows/ci.yml
```diff
@@ -204,16 +204,12 @@ jobs:
...
     secrets:
       pypi_token: ${{ secrets.pypi_token }}
-      anaconda_token: ${{ secrets.anaconda_token }}
+      anaconda_token: ${{ secrets.anaconda_org_upload_token }}
 
   publish_pure:
     needs: [publish]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@main
     with:
-      upload_to_anaconda: ${{ (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') }}
```
nabobalis, line 213: Anaconda uploads are for wheels only.

Cadair: This should be the universal wheel? maybe it's also the sdist.

Cadair: I think the issue is that if we don't also upload the universal wheel if you try to use the nightly pip index on windows it will explode (as it won't be able to build from source).

nabobalis: It uploaded both the pure wheel and the dist. Maybe if we can get the OA template to only upload the wheels?

PR #7461, opened 2024-02-28
Reimplement Slugify 

ahmedhosssam:
This PR addresses the issue https://github.com/sunpy/sunpy/issues/7450

Note: I know there is another open PR that solves the same issue, but I'm trying to make any code changes to gain more understanding of the code. (I'm preparing myself for gsoc, so I'm not waiting for a merge more than feedback)

## PR Description:
I think that the bug was having more than one dot in some files and changing it results some confusion, so I think this implementation solves the problem.
But there is one failed test that checks for `file.greg.fits` to be `file_greg.fits`
Should we change the test? Or should we stick with the original function?


nabobalis:
Hello @ahmedhosssam, as you have already said, there is already a PR opened to address this here: https://github.com/sunpy/sunpy/pull/7453 and it works by redoing how we normalize the filenames. 

I do not want the limited time our reviewers have to review PRs that will not be merged in.
I hope you understand.

PR #7440, opened 2024-02-08
Fix jsoc keyword not escaping a string for the query

nabobalis:

PR #7467, opened 2024-03-01
Fix jsoc keyword not escaping a string for the query

nabobalis:
Supersedes: https://github.com/sunpy/sunpy/pull/7440/files

Bugfix for 5.0
Bugfix for 5.1
I would consider this a minor change.

### DIFF
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -336,8 +336,8 @@ class JSOCClient(BaseClient):
 
         requests = []
         self.query_args = jsoc_response.query_args
-        supported_protocols = sorted({'fits', 'as-is'})
-        supported_methods = sorted({'url-tar', 'url', 'url-quick'})
+        supported_protocols = sorted(['fits', 'as-is'])
+        supported_methods = sorted(['url-tar', 'url', 'url-quick'])
         for block in jsoc_response.query_args:
 
             ds = self._make_recordset(**block)
@@ -680,16 +680,15 @@ class JSOCClient(BaseClient):
             #
             # Note that whilst it is technically possible to just search by series,
             # this is not allowed here, because some of these would be very large
-            # searches that would make JSOC sad
+            # searches that would make JSOC unhappy with sunpy
             raise ValueError("Time, Wavelength or an explicit PrimeKey must be specified.")
         keys = []
         keyword_info = c.keys(series)
         for key, value in keyword.items():
             if key not in keyword_info:
                 raise ValueError(f"Keyword: '{key}' is not supported by series: {series}")
-            # We have to ensure that any values that are strings are quoted
-            other_value = value['value'] if str(value['value']).isnumeric() else f"'{value['value']}'"
-            keys.append(f"{key}{value['operator']}{other_value}")
+            # We have to ensure that any input values that are strings are quoted
+            keys.append(f"{key}{value['operator']}'{value['value']}'")
         keyword_string = f"[? {' AND '.join(keys)} ?]" if keys else ""
         return f"{series}{primekey_string}{keyword_string}{segment}"
 
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -650,7 +650,7 @@ class JSOCClient(BaseClient):
             # This is executed when wavelength has been passed through PrimeKey().
             wavelength = f"{primekey.pop('WAVELNTH', '')}"
         elif not primekey.get('WAVELNTH'):
-            wavelength = f'{int(np.ceil(wavelength.to(u.AA).value))}'
+            wavelength = f'{int(np.ceil(wavelength.to_value(u.AA)))}'
         else:
             # This is executed when wavelength has been passed both through PrimeKey()
             # and Wavelength().
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -300,11 +300,12 @@ class JSOCClient(BaseClient):
         query = and_(*query)
         blocks = []
         for block in walker.create(query):
-            iargs = kwargs.copy()
-            iargs.update(block)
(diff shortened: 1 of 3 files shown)


sunpy/net/jsoc/jsoc.py
```diff
@@ -650,27 +645,19 @@ def _make_recordset(self, series, start_time='', end_time='', wavelength='',
...
-                                 " or separate them by OR operator.")
-                raise ValueError(error_message)
-        else:
+            timestr = f"{primekey.pop(list(match)[0], '')}" if match else ''
+        if wavelength == '':
             # This is executed when wavelength has been passed through PrimeKey().
-            wavelength = '{}'.format(primekey.pop('WAVELNTH', ''))
+            wavelength = f"{primekey.pop('WAVELNTH', '')}"
+        elif not primekey.get('WAVELNTH'):
+            wavelength = f'{int(np.ceil(wavelength.to(u.AA).value))}'
```
wtbarnes, line 653:
```suggestion
            wavelength = f'{int(np.ceil(wavelength.to_value(u.AA)))}'
```

sunpy/net/jsoc/jsoc.py
```diff
@@ -702,18 +685,19 @@ def _make_recordset(self, series, start_time='', end_time='', wavelength='',
...
         keyword_info = c.keys(series)
         for key, value in keyword.items():
-            if key in keyword_info:
-                keys.append(f"{key}{value['operator']}{value['value']}")
-            else:
+            if key not in keyword_info:
                 raise ValueError(f"Keyword: '{key}' is not supported by series: {series}")
+            # We have to ensure that any values that are strings are quoted
+            other_value = value['value'] if str(value['value']).isnumeric() else f"'{value['value']}'"
+            keys.append(f"{key}{value['operator']}{other_value}")
```
wtbarnes, line 692: This is the bugfix that is the actual focus of this PR right?

wtbarnes: I'm also confused by this line. If as the comment says you are checking if this is a string or not, why not just do `if isinstance(value['value], str)`?

sunpy/net/jsoc/tests/test_jsoc.py
```diff
@@ -371,3 +366,14 @@ def test_empty_response_fetch(client):
...
+
+
+@pytest.mark.remote_data
+def test_string_keyword_is_converted_correctly(client):
+    # Bug reported here:
+    # https://community.openastronomy.org/t/how-to-select-only-fuv-iris-level-1-data-using-jsoc/900
+    responses = client.search(
+        a.Time("2014-01-01T00:00", "2014-01-01T00:05"),
+        a.jsoc.Series("iris.lev1"), a.jsoc.Keyword("INSTRUME") == "FUV"
+        )
```
wtbarnes, line 378:
```suggestion
    )
```

wtbarnes: Am I going crazy? Shouldn't that that last `)` be indented at the same level as `responses`?

wtbarnes (approved):
This should really get a second review from someone who understands the JSOC client much better than I do.

sunpy/net/jsoc/tests/test_jsoc.py
nabobalis: No you are correct.

sunpy/net/jsoc/jsoc.py
nabobalis: Err, I forget now

sunpy/net/jsoc/jsoc.py
```diff
@@ -335,8 +336,8 @@ def request_data(self, jsoc_response, method='url', **kwargs):
 
         requests = []
         self.query_args = jsoc_response.query_args
-        supported_protocols = {'fits', 'as-is'}
-        supported_methods = {'url-tar', 'url', 'url-quick'}
+        supported_protocols = sorted({'fits', 'as-is'})
+        supported_methods = sorted({'url-tar', 'url', 'url-quick'})
```
Cadair, line 340: If we care about these being sorted, why not just use a list? Taking a set and then converting it to a generator seems insane.

sunpy/net/jsoc/jsoc.py
```diff
@@ -650,27 +645,19 @@ def _make_recordset(self, series, start_time='', end_time='', wavelength='',
                                  "prime-keys is not yet supported. Alternative is to "
                                  "use only one of the primekey to query for series data.")
                 raise ValueError(error_message)
-            if match:
```
Cadair, line 653: As valid as these logic refactors are doing them in the middle of this PR which is fixing a legitimate bug, in code that most of us are not familiar with makes reviewing this PR *really* hard.

nabobalis: Don't worry about it.

Cadair: Well if you care about faster reviews then it might be something to think about :wink:

nabobalis: *I won't reply to that with my actual thoughts.*

But yes, that is very true. I will endeavour to make sure my PR's are smaller in the future.

sunpy/net/jsoc/jsoc.py
nabobalis: Fixed.

sunpy/net/jsoc/jsoc.py
nabobalis: None of the unit tests failed for me locally when I did that, so I have no idea why I did that in the first place or the worrying one is that the unit tests don't cover this and I have broken something.

Only time and God can tell us.

nabobalis:
Online fail is:

FAILED ../../.tox/py312-online/lib/python3.12/site-packages/sunpy/map/mapsequence.py::sunpy.map.mapsequence.MapSequence.save

Not due to this PR hopefully.

PR #7469, opened 2024-03-02
Pass in segments keyword to the DRMS to return the SUMS dirs

nabobalis:
Supersedes https://github.com/sunpy/sunpy/pull/7236 

So the data on the JSOC is stored in SUM directories and they can return these locations to a user. To return this information, we just have to add a value for the `seg` key for the DRMS client.

The reason we will want to do this is that it allows a user (and in future the JSOCClient) to use these file locations to download the JSOC data without having to export. The JSOC has asked sunpy to stop doing exports and this is step 1 of that process.

### DIFF
--- a/changelog/7469.feature.rst
+++ b/changelog/7469.feature.rst
@@ -1,1 +1,1 @@
-`sunpy.net.jsoc.JSOCCLient` queries now return the SUMS directory paths as the segment key value in the results table.
+`sunpy.net.jsoc.JSOCClient` queries now return the SUMS directory paths as the segment key value in the results table.
--- a/changelog/7469.feature.rst
+++ b/changelog/7469.feature.rst
@@ -1,0 +1,1 @@
+`sunpy.net.jsoc.JSOCCLient` queries now return the SUMS directory paths as the segment key value in the results table.
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -7,6 +7,7 @@ from pathlib import Path
 
 import drms
 import numpy as np
+import pandas as pd
 import parfive
 from packaging.version import Version
 
@@ -745,11 +746,16 @@ class JSOCClient(BaseClient):
             key = keywords
         log.debug(f"Running following query: {ds}")
         log.debug(f"Requesting following keywords: {key}")
-        result = client.query(ds, key=key, rec_index=is_meta)
+        result = client.query(ds, key=key, seg=segments_passed, rec_index=is_meta)
         if result is None or result.empty:
             return astropy.table.Table()
+        if segments_passed :
+            # DRMS returns a tuple of length 2 if there is segment information
+            query_result, segment_result = result[0], result[1]
+            query_result = pd.concat([query_result, segment_result],axis=1)
         else:
-            return astropy.table.Table.from_pandas(result)
+            query_result = result
+        return astropy.table.Table.from_pandas(query_result)
 
     @classmethod
     def _can_handle_query(cls, *query):
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -753,7 +753,7 @@ class JSOCClient(BaseClient):
             query_result = pd.concat([query_result, segment_result],axis=1)
         else:
             query_result = result
-        if result is None or result.empty:
+        if result is None or query_result.empty:
             return astropy.table.Table()
         return astropy.table.Table.from_pandas(query_result)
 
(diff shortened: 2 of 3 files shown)


sunpy/net/jsoc/tests/test_jsoc.py
```diff
@@ -53,6 +53,8 @@ def test_return_query_args(client):
                         a.jsoc.Segment('Bp') & a.jsoc.Segment('magnetogram'))
     # Because res.query_args is list that contains dict
     assert 'primekey' in res.query_args[0]
+    assert "magnetogram" in res.keys()
+    assert res["magnetogram"][0].startswith("/SUM")
```
wtbarnes, line 57: Shouldn't we check for the existence of 'BP' as well since it is also a segment? I also wonder if a more comprehensive test wouldn't be running a query with and without passing `a.jsoc.Segment` and checking that the length is different / the segment names are now columns with the expected values.

nabobalis: Sounds good, I will do so.

changelog/7469.feature.rst
```diff
@@ -0,0 +1 @@
+`sunpy.net.jsoc.JSOCCLient` queries now return the SUMS directory paths as the segment key value in the results table.
```
nabobalis, line 1:
```suggestion
`sunpy.net.jsoc.JSOCClient` queries now return the SUMS directory paths as the segment key value in the results table.
```

sunpy/net/jsoc/tests/test_jsoc.py
nabobalis: I created a new test with these changes and undid the changes for this test.

sunpy/net/jsoc/tests/test_jsoc.py
```diff
@@ -371,3 +371,26 @@ def test_empty_response_fetch(client):
...
+    assert "magnetogram" not in res.keys()
+    assert "Bp" not in res.keys()
+    assert "magnetogram" in seg_res.keys()
+    assert "Bp" in seg_res.keys()
+    # SUMS directories are like /SUM93/D1044682184/S00000/Bp.fits
+    # So checking the start and end is the easiest way to check the segments
+    assert seg_res["magnetogram"][0].startswith("/SUM")
+    assert seg_res["Bp"][0].startswith("/SUM")
+    assert seg_res["magnetogram"][0].endswith("magnetogram.fits")
+    assert seg_res["Bp"][0].endswith("Bp.fits")
```
nabobalis, line 396: Kind of brute force and simple.

wtbarnes (approved):
Looks good to me, as long as the relevant online tests pass

wtbarnes:
> The reason we will want to do this is that it allows a user (and in future the JSOCClient) to use these file locations to download the JSOC data without having to export. The JSOC has asked sunpy to stop doing exports and this is step 1 of that process.

Given this, should we have an example that shows how to do that, maybe including how to build a header from the metadata and do the stitching together of the file yourself?

nabobalis:
I don't know if it's possible yet. You want it in this PR?

sunpy/net/jsoc/jsoc.py
```diff
@@ -764,11 +765,17 @@ def _lookup_records(self, iargs):
             key = keywords
         log.debug(f"Running following query: {ds}")
         log.debug(f"Requesting following keywords: {key}")
-        result = client.query(ds, key=key, rec_index=isMeta)
-        if result is None or result.empty:
+        result = client.query(ds, key=key, seg=segments_passed, rec_index=isMeta)
+        if segments_passed :
```
nabobalis, line 769:
```suggestion
        if segments_passed:
```

sunpy/net/jsoc/jsoc.py
```diff
@@ -745,11 +746,16 @@ def _lookup_records(self, iargs):
...
         log.debug(f"Requesting following keywords: {key}")
-        result = client.query(ds, key=key, rec_index=is_meta)
+        result = client.query(ds, key=key, seg=segments_passed, rec_index=is_meta)
+        if segments_passed:
+            # DRMS returns a tuple of length 2 if there is segment information
+            query_result, segment_result = result[0], result[1]
+            query_result = pd.concat([query_result, segment_result],axis=1)
+        else:
+            query_result = result
         if result is None or result.empty:
```
nabobalis, line 756:
```suggestion
        if result is None or query_result.empty:
```

nabobalis:
Online fail is: FAILED ../../.tox/py312-online/lib/python3.12/site-packages/sunpy/io/tests/test_filetools.py::test_write_file_fits[aia_171_image.fits] - FileNotFoundError: [Errno 2] No such file or directory: 'aia_171_image.fits'

PR #7470, opened 2024-03-02
Remove vendored `obsgeo_to_frame` in favor of using Astropy

ahmedhosssam:
Hi,
I found that some tests in `sunpy/coordinates/tests/test_wcs_utils.py` should be removed after `Astropy 5.0`, but Astropy is now `6.0` so I think it should be removed, right?
If it's right, I think it will make the whole testing process faster, now pytest runs in less than 120 seconds after it used to take more than 180 seconds on my computer.

Please let me know if I'm missing something.

### DIFF
--- a/sunpy/coordinates/tests/test_wcs_utils.py
+++ b/sunpy/coordinates/tests/test_wcs_utils.py
@@ -343,13 +343,6 @@ def test_frame_mapping_obsgeo_spherical(dkist_location, caplog):
     assert not caplog.records
 
 
-@pytest.mark.parametrize("obsgeo", ([np.nan] * 6, None, [0] * 6, [54] * 5))
-def test_obsgeo_invalid(obsgeo):
-
-    with pytest.raises(ValueError):
-        obsgeo_to_frame(obsgeo, None)
-
-
 def test_observer_hgln_crln_priority():
     """
     When extracting the observer information from a FITS header, ensure
--- a/sunpy/coordinates/tests/test_wcs_utils.py
+++ b/sunpy/coordinates/tests/test_wcs_utils.py
@@ -3,7 +3,7 @@ import numpy as np
 import pytest
 
 import astropy.units as u
-from astropy.coordinates import ITRS, BaseCoordinateFrame, SkyCoord
+from astropy.coordinates import BaseCoordinateFrame, SkyCoord
 from astropy.coordinates.earth import EarthLocation
 from astropy.tests.helper import assert_quantity_allclose
 from astropy.time import Time
@@ -18,7 +18,6 @@ from sunpy.coordinates.frames import (
 )
 from sunpy.coordinates.wcs_utils import (
     _set_wcs_aux_obs_coord,
-    obsgeo_to_frame,
     solar_frame_to_wcs_mapping,
     solar_wcs_frame_mapping,
 )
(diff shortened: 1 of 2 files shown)


ayshih:
The vendored version of `sunpy.coordinates
wcs_utils.obsgeo_to_frame()` should be removed at the same time (this all came in with #5315).

It'd be mind-boggling if these three tests take over a minute to run in your environment.  I suspect that something went awry in your comparison.

ahmedhosssam:
> The vendored version of `sunpy.coordinates wcs_utils.obsgeo_to_frame()` should be removed at the same time (this all came in with #5315).

Done.
Should I remove these tests also?

```
@pytest.mark.parametrize("obsgeo", ([np.nan] * 6, None, [0] * 6, [54] * 5))
def test_obsgeo_invalid(obsgeo):

    with pytest.raises(ValueError):
        obsgeo_to_frame(obsgeo, None)
def test_observer_hgln_crln_priority():
    """
    When extracting the observer information from a FITS header, ensure
    Stonyhurst (HG) coordinates are preferred over Carrington (CR) if present.
    Note that `Map` creates a custom FITS header with a sanitized observer
    location, so it is separately tested in the map module.
    """
    header = {'CRVAL1': 0,
              'CRVAL2': 0,
              'CRPIX1': 5,
              'CRPIX2': 5,
              'CDELT1': 10,
              'CDELT2': 10,
              'CUNIT1': 'arcsec',
              'CUNIT2': 'arcsec',
              'PC1_1': 0,
              'PC1_2': -1,
              'PC2_1': 1,
              'PC2_2': 0,
              'NAXIS1': 6,
              'NAXIS2': 6,
              'CTYPE1': 'HPLN-TAN',
              'CTYPE2': 'HPLT-TAN',
              'date-obs': '1970-01-01T00:00:00',
              'mjd-obs': 40587,
              'hglt_obs': 0,
              'hgln_obs': 0,
              'crlt_obs': 2,
              'crln_obs': 2,
              'dsun_obs': 10,
              'rsun_ref': 690000000}
    wcs = WCS(header)
    c = wcs.pixel_to_world(0, 0)
    assert c.observer.lon == 0 * u.deg
    # Note: don't test whether crlt or hglt is used---according to
    # _set_wcs_aux_obs_coord, those are expected to always be the same and so
    # the same one is always used
```

nabobalis:
Do you mean `test_obsgeo_invalid`? I think that should go but I don't see why the test after that would go?

ahmedhosssam:
All of the tests that I removed already exist in `astropy.wcs.tests.test_utils.py` now, except `test_observer_hgln_crln_priority`, it tests `astropy.wcs.WCS`. As far as I understand, the test should belong to `astropy` rather than `sunpy`. Or the data in `header` are related to `sunpy`?

nabobalis:
I would need Albert to double check but that test should stay in sunpy.

wtbarnes (approved):

ayshih:
> All of the tests that I removed already exist in `astropy.wcs.tests.test_utils.py` now, except `test_observer_hgln_crln_priority`, it tests `astropy.wcs.WCS`. As far as I understand, the test should belong to `astropy` rather than `sunpy`. Or the data in `header` are related to `sunpy`?

It's not readily apparent from the test code, but it internally calls a `sunpy` function to convert the WCS header to a SunPy coordinate frame.  The test should stay.

ahmedhosssam:
> Please make the changes flagged by pre-commit, otherwise looks good to me

Done.

ayshih (approved):

ayshih:
Thanks for the PR!

PR #7477, opened 2024-03-06
Fix VSOClient example doctest

nabobalis:
Been failing.


wtbarnes:
Some context from the VSO side:

> It looks like the results now include an L1 FITS file in addition to lz files.  The EIT code itself hasn't changed.

nabobalis:
Whatever works.

Cadair:
> I wonder if instead we should limit this query to the lz results rather than expanding the list of possible inputs?

I would say the other way around, let's return the L1 files?

I missed this PR exists and have done a L1 only fix along with the LASP issue in #7483

Cadair:
Oh so fun fact, it seems that the `a.Level` attr doesn't filter the results down. I will pull this commit into #7483

nabobalis:
RIP

PR #7483, opened 2024-03-08
Update base URL for EVEClient to new LASP URL

Cadair:
This fixes the online tests by changing the EVE L0CS client to use the new URLs for the data from LASP and account for the fact that the VSO new returns L1 SOHO EIT data.


wtbarnes:
The oldestdeps fail is real. The test you unskipped is not raising the expected warning.

alasdairwilson (approved):

PR #7486, opened 2024-03-08
Flip jp2 data vertically before saving

### LINKED ISSUE #7485: Saving with the ".jp2" extension results in a vertically flipped image
### Describe the bug

Images get flipped when saving as jp2 files.
This was my fault when I added this feature 🎉

### To Reproduce

```
from sunpy.map import Map
m = Map("https://helioviewer.org/jp2/AIA/2024/03/08/304/2024_03_08__00_00_05_123__SDO_AIA_AIA_304.jp2")
m.peek()
m.save("test.jp2")
flipped = Map("test.jp2")
flipped.peek()
```

See screenshots below

### Screenshots

`m.peek()`
image

`flipped.peek()`
image

### System Details

>>> import sunpy
>>> sunpy.util.system_info()
==============================
sunpy Installation Information
==============================

General
#######
OS: Mac OS 14.3.1
Arch: 64bit, (arm)
sunpy: 4.0.dev2573+gf79e5d92d
Installation path: sunpy.egg-info

Required Dependencies
#####################
astropy: 6.0.0
numpy: 1.26.4
packaging: 23.2
parfive: 2.0.2
pyerfa: 2.0.1.1
requests: 2.31.0

Optional Dependencies
#####################
sunpy: 4.0.dev2573+gf79e5d92d

### Installation method

git checkout

dgarciabriseno:
## PR Description

Fixes #7485 

Flipping the data array when saving as jp2 fixes this problem.

### DIFF
--- a/sunpy/io/_jp2.py
+++ b/sunpy/io/_jp2.py
@@ -176,7 +176,7 @@ def write(fname, data, header, **kwargs):
 
     tmpname = fname + "tmp.jp2"
     jp2_data = np.uint8(data)
-    jp2 = Jp2k(tmpname, jp2_data, **kwargs)
+    jp2 = Jp2k(tmpname, np.flip(jp2_data, 0), **kwargs)
 
     # Append the XML data to the header information stored in jp2.box
     meta_boxes = jp2.box
--- a/sunpy/io/_jp2.py
+++ b/sunpy/io/_jp2.py
@@ -176,7 +176,11 @@ def write(fname, data, header, **kwargs):
 
     tmpname = fname + "tmp.jp2"
     jp2_data = np.uint8(data)
-    jp2 = Jp2k(tmpname, np.flip(jp2_data, 0), **kwargs)
+
+    # The jp2 data is flipped when read in, so we have to flip it back before
+    # saving.
+    flipped = np.flip(jp2_data, 0)
+    jp2 = Jp2k(tmpname, flipped, **kwargs)
 
     # Append the XML data to the header information stored in jp2.box
     meta_boxes = jp2.box
--- a/sunpy/io/_jp2.py
+++ b/sunpy/io/_jp2.py
@@ -178,7 +178,7 @@ def write(fname, data, header, **kwargs):
     jp2_data = np.uint8(data)
 
     # The jp2 data is flipped when read in, so we have to flip it back before
-    # saving.
+    # saving. See https://github.com/sunpy/sunpy/pull/768 for context.
     flipped = np.flip(jp2_data, 0)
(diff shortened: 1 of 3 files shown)


dgarciabriseno:
test

sunpy/io/_jp2.py
```diff
@@ -176,7 +176,7 @@ def write(fname, data, header, **kwargs):
 
     tmpname = fname + "tmp.jp2"
     jp2_data = np.uint8(data)
-    jp2 = Jp2k(tmpname, jp2_data, **kwargs)
+    jp2 = Jp2k(tmpname, np.flip(jp2_data, 0), **kwargs)
```
ayshih, line 179: Add a comment that explains that we have to flip the array back here because we flip it upon reading (due to the difference in the origin of the data)

ayshih (changes requested):
You need to fix the relevant unit test (https://github.com/sunpy/sunpy/blob/f79e5d92d82fdca84cceadb98f79a00b7e82acc0/sunpy/io/tests/test_jp2.py#L44-L46) so that it actually verifies the output

dgarciabriseno:
> You need to fix the relevant unit test (
> 
> https://github.com/sunpy/sunpy/blob/f79e5d92d82fdca84cceadb98f79a00b7e82acc0/sunpy/io/tests/test_jp2.py#L44-L46
> ) so that it actually verifies the output

How would you recommend doing that?
I'm not sure the result will match the original fits data array since it casts the data to `uint8` to save it as a jp2. (There's other issues with doing that too, a simple cast isn't sufficient to convert from fits to jp2).

sunpy/io/_jp2.py
dgarciabriseno: Where does it get flipped when its read in?

dgarciabriseno: I added a comment here, and updated the test to verify the data array after reading it back.

sunpy/io/_jp2.py
```diff
@@ -176,7 +176,11 @@ def write(fname, data, header, **kwargs):
 
     tmpname = fname + "tmp.jp2"
     jp2_data = np.uint8(data)
-    jp2 = Jp2k(tmpname, jp2_data, **kwargs)
+
+    # The jp2 data is flipped when read in, so we have to flip it back before
+    # saving.
```
wtbarnes, line 181:
```suggestion
    # saving. See https://github.com/sunpy/sunpy/pull/768 for context.
```

wtbarnes (changes requested):
Just needs a changelog, but other than that looks good to me.

wtbarnes (approved):

nabobalis (approved):

PR #7488, opened 2024-03-09
Pin out pytest 8.1.1

Cadair:
We also have to pin out the yanked 8.1.0 as pypicky doesn't filter yanked versions.


nabobalis (approved):
YOLO

Issue #7478, opened 2024-03-07
Fix inaccuracies in `sunpy.map.maputils.py`


ayshih:
The following calculations in `sunpy.map.maputils.py` ought to be fixed to not use the Cartesian approximation:

https://github.com/sunpy/sunpy/blob/3939b7e56428fd6df65d8f85499ad6dbbae8ccea/sunpy/map/maputils.py#L209
https://github.com/sunpy/sunpy/blob/3939b7e56428fd6df65d8f85499ad6dbbae8ccea/sunpy/map/maputils.py#L260
https://github.com/sunpy/sunpy/blob/3939b7e56428fd6df65d8f85499ad6dbbae8ccea/sunpy/map/maputils.py#L290

simoneferretti:
Hi, I've worked on the file and tried solving the inaccuracies using the separation method from Skycoord.
[maputils.json](https://github.com/sunpy/sunpy/files/14540510/maputils.json)

simoneferretti:
Hi, I'd like to work on this issue

nabobalis:
Go ahead @simoneferretti

simoneferretti:
I've sent a pull request

PaoloBiolghini:
I think I have found a possible solution, I'm going to open a PR asap

nabobalis:
There is already a PR open here: https://github.com/sunpy/sunpy/pull/7489 if you could help that would be ideal than another PR.

PaoloBiolghini:
sorry i didn't see the comment; should i remove my PR and link it to the one you pinged me?

nabobalis:
Can you add code suggestions to the original PR that I linked to?

ayshih:
Since multiple folks have been having difficulties making the appropriate fix, clearly it was not a "good first issue" like I had thought.  I have fixed it myself in #7512.

PR #7489, opened 2024-03-10
Issue #7478: fixes inaccuracies in sunny.map.maputils.py using SkyCoord() function instead of Cartesian approximation

simoneferretti:
## PR Description
pre-commit.ci autofix


simoneferretti:
pre-commit.ci autofix

nabobalis:
Thank you for the PR @simoneferretti.

Unfortunately, the tests are failing on the CI, can you debug and fix them please?

ayshih:
Please don't use `SkyCoord.separation()` because it is too complex and too slow for this (special) case.  The correct equation is straightforward for those who remember trigonometry.

Also, be careful when working with quantities.  You've multiplied a quantity which already has units of arcseconds by `u.deg`.

ayshih:
Thank you for attempting to fix the issue!  However, since the straightforward solution is much simpler than the approach you took, I have submitted #7512 and am closing this PR.

PR #7494, opened 2024-03-12
Updates Sunpy core with package template

CyclingNinja:
* Adds `.cruft.json` 
* General editing and compliance with new package template


.ruff.toml
```diff
@@ -25,12 +25,14 @@ extend-ignore = [
...
-[per-file-ignores]
+[lint.per-file-ignores]
 # Part of configuration, not a package.
 "setup.py" = ["INP001"]
 "conftest.py" = ["INP001"]
-# Implicit-namespace-package. The examples are not a package.
-"docs/*.py" = ["INP001"]
+"docs/*.py" = [
+  "INP001",  # Implicit-namespace-package. The examples are not a package.
+  "E402",  # Module imports not at top of file
```
Cadair, line 34: This rule should only apply to conf.py

Cadair (approved):

CyclingNinja:
pre-commit.ci autofix

CyclingNinja:
pre-commit.ci autofox

wtbarnes (approved):
I am not the most qualified to review this, but this all looks good to me.

Cadair:
I think the remaining todo here is that the oldestdeps build is broken.

nabobalis (approved):

nabobalis (changes requested):
blocking until oldest deps is fixed

Cadair:
The oldestdeps build is also failing on main, so I think that this is good to go.

sunpy/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""
```
nabobalis, line 1: Does this need to be here?

nabobalis (approved):

PR #7502, opened 2024-03-20
Pass from Cartesian approximation to Trigonometric method

PaoloBiolghini:
Trying to sole error #7478 #7489 
I modified the code deleting the carteshian method to find the "coordinate_angles" and creating a trigonometric method to find this variable; 
The following formula represent the new method
image

I have created a intermeiate variable that store the result of the arctan in order to create a code more readable.
From the fact that there are two division going on there is always the possibility that a division by zero error occurs; in order to avoid this scenario I use np.div instead of / operator;
To find the final value I use the cos because in this way i only have to make a check and not two.
If Tx is equal to zero it means that the related coordinate_angle is exaclty Ty

### DIFF
--- a/.test_package_pins.txt
+++ b/.test_package_pins.txt
@@ -1,5 +1,2 @@
 # This file is used to pin test packages across sunpy coordinated packages
 # Do not remove, even if it's empty!
-
-# https://github.com/scientific-python/pytest-doctestplus/issues/243
-pytest<8.1
--- a/sunpy/map/maputils.py
+++ b/sunpy/map/maputils.py
@@ -206,7 +206,7 @@ def contains_full_disk(smap):
     _verify_coordinate_helioprojective(smap.coordinate_frame)
     edge_of_world = _edge_coordinates(smap)
     # Calculate the distance of the edge of the world in solar radii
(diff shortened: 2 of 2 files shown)


nabobalis:
Following https://github.com/sunpy/sunpy/pull/7489#issuecomment-2011076419, I will close this as well.

Thanks for the PR.

ayshih:
Thank you for attempting to fix the issue!  However, your approach is mathematically equivalent to the Cartesian approximation, and is actually numerically worse in some ways.  You can see the equivalence by noting:
```
angle = arctan (T_y / T_x) = arccos (T_x / sqrt(T_x^2 + T_y^2)
```
which means that:
```
T_x / cos(angle) = T_x / (T_x / sqrt(T_x^2 + T_y^2)) = sqrt(T_x^2 + T_y^2)
```
That is, your use of trigonometric functions doesn't actually fix the inaccuracy, but just obscures it.  Please see #7512 for the correct equation.

Incidentally, if you ever have to take arctangents in the future, I'll point out NumPy provides an `arctan2()` that takes two arguments, which means that a division isn't necessary, and also avoids quadrant degeneracy.  That is, you should use `np.arctan2(T_y, T_x)`  instead of `np.arctan(T_y / T_x)`.

PR #7512, opened 2024-03-21
Removed three uses of the small-angle approximation in `sunpy.map.maputils`

### LINKED ISSUE #7478: Fix inaccuracies in `sunpy.map.maputils.py`

ayshih:
There are three functions in `sunpy.map.maputils` that implicitly rely on the small-angle approximation to convert helioprojective theta_x and theta_y to the angular offset from disk center.  (This is equivalent as treating helioprojective Cartesian as if it were a true Cartesian coordinate system.)  This PR has the functions do the proper, accurate computation.

Fixes #7478

### DIFF
--- a/changelog/7512.bugfix.rst
+++ b/changelog/7512.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a very minor inaccuracy in three `sunpy.map` utility functions (:func:`~sunpy.map.contains_full_disk`, :func:`~sunpy.map.coordinate_is_on_solar_disk`, and :func:`~sunpy.map.is_all_off_disk`) resulting from the accidental use of the small-angle approximation.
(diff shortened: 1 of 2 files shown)


ayshih:
When checking over this PR, I realized that two of these functions should simply be calling the third function to avoid code duplication, so now that is done.

nabobalis (approved):

alasdairwilson (approved):

PR #7519, opened 2024-03-21
reduce config?

nabobalis:
Thoughts?


nabobalis:
Oh it broke the figure tests, I will debug

nabobalis:
Closing due to:

I would like a -parallel tox factor to install it and run with -n auto

But I can't figure out how to do it without duplicating the config many times with -online etc

Color is just for ci because GH actions doesn't get detected as automatically supporting it.

PR #7520, opened 2024-03-21
See if we use tmpdir instead of removing files

nabobalis:
This seems to be failing to remove the file, so I just swapped it to use a tmpdir instead.

### DIFF
--- a/sunpy/io/tests/test_filetools.py
+++ b/sunpy/io/tests/test_filetools.py
@@ -1,5 +1,6 @@
 import os
 import pathlib
+from unittest.mock import patch
 
 import numpy as np
 import pytest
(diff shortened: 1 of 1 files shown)


nabobalis:
ResourceWarning?! Can we skip those please? (until we fix them)

wtbarnes:
What's the fix here? Do we have an issue to track this? If this is an upstream problem we should just be ignoring.

nabobalis:
> What's the fix here? Do we have an issue to track this? If this is an upstream problem we should just be ignoring.

This test was failing on the last line which was to remove the file. This shouldn't happen but we also should not be removing files like this in a unit test, we should be using pytest tmpdir to save any files and not worry about cleaning them up manually.

wtbarnes:
I understand that and totally agree tmpdir is the right approach. I meant the bit about the ResourceWarning.

nabobalis:
> I understand that and totally agree tmpdir is the right approach. I meant the bit about the ResourceWarning.

Oh sorry, the resource warning is that a network connection was not closed by the time some other block of python code executed. We've seen this before but I thought we ignored them as we can't fix them without a deep dive to work out where/why these are occuring.

alasdairwilson (approved):

PR #7369, opened 2024-01-06
Warn user when importing submodule without the submodules dependencies

alasdairwilson:
Adds a warning when user imports a submodule without having all the deps defined in that submodules extras.

Closes #7333 

```
In [1]: from sunpy.map import Map
H:\sm2\sunpy\sunpy\util\sysinfo.py:111: ImportWarning: Importing sunpy.map without its extra dependencies may result in errors.
The following packages are not installed:
['matplotlib>=3.5.0']
To install sunpy with these dependencies use `install sunpy[map]`or `install sunpy[all]` for all extras.
```

PR #7527, opened 2024-03-26
Fix issues in #7369

Cadair:
#7369 [broke the DKIST CI](https://github.com/DKISTDC/dkist/actions/runs/8435154989/job/23099968315#step:10:148) because test helpers was importing map and dkist does not depend on the map extra. Also in the process of this I noticed that the strip was stripping more things than it should be.


Cadair:
no backport assuming that #7529  and #7528  are merged.

alasdairwilson (approved):

PR #7531, opened 2024-03-26
Do not optionally depend on truly optional deps

Cadair:
This has come out of #7369

I think one of the foundational issues exposed by #7369 is that our extras do not cleanly separate dependencies that are required to import a subpackage vs dependencies which are only triggered if you call a specific function (or sometimes even a specific function with a specific argument).

The (controversial) approach I have taken in this PR is that we shouldn't even list in our deps packages which fall into the latter category (we for instance do not do so for `jplephem` at the moment).

---

Note that I have exempted "spice" from this culling as it's really "sunpy.coordinates.spice" which is a required dep for a subsubmodule.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,6 @@ image = [
   # !=1.10.0 due to https://github.com/scipy/scipy/issues/17718
   "scipy>=1.8.0,!=1.10.0",
 ]
-jpeg2000 = ["glymur>=0.9.1,!=0.9.5", "lxml>=4.9.0,!=5.0.0"]
 map = [
   "matplotlib>=3.5.0",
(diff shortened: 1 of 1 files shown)


alasdairwilson:
I agree with the philosophy here (and not just because it fixes my mistakes) I think potentially the removed ones here should be moved to their own extras so they can still be pulled in.

although there is a philosophical question about [all] should this pull in every package that sunpy might call or should it pull in every package that we require for standard use: If I am a user and I want to use sunpy, does install sunpy[all] mean I should get dask/scikit-image

nabobalis:
https://github.com/sunpy/sunpy/pull/7536 supecedes this.

PR #7537, opened 2024-03-27
Tidy up of the io doc strings and API page. Deprecate toplevel io namespace. 

nabobalis:
Changelog check will fail as I had to change other changelogs but I did add one.


docs/reference/io.rst
```diff
@@ -1,48 +1,56 @@
...
-metadata, data and mask. In comparison, the mask is not currently saved
-to FITS. The following code shows to to save and load a sunpy Map to an asdf
-file
+sunpy currently implements support for saving `Map <sunpy.map.GenericMap>` and `coordinate frame <sunpy.coordinates.frames>` objects into asdf files.
+As asdf tightly integrates into Python, saving a map to an asdf file will save the metadata, data and mask.
+In comparison, the mask is not currently saved to FITS.
+The following code shows to to save and load a sunpy Map to an asdf file:
 
-.. doctest-requires:: asdf
+.. code-block:: python
```
wtbarnes, line 47: This reads more like a how-to guide. I don't think it belongs in our reference material.

docs/reference/io.rst
```diff
@@ -79,34 +87,11 @@ file
...
 
-When saving a Map to ASDF all maps are saved as a `.GenericMap` and not a specific source class.
+When saving a Map to ASDF, all maps are saved as a `.GenericMap` and not a specific source class.
 This comes with some trade-offs.
 If you are using custom map sources defined outside of the `sunpy` core package, and these sources are imported after asdf has been invoked for the first time (used, not just imported), then they will not be registered with the asdf converter.
 Also if the custom map subclass is not registered with `sunpy.map.Map` upon loading of the map, it will be returned as a `.GenericMap`.
 This approach has been chosen despite these limitations so that once a map is saved to an ASDF file it can always be loaded back into a map rather than the asdf library returning it as a Python dictionary.
-It also follows the philosophy of the way maps are saved and loaded in the FITS format, where the components of the Map are serialised and the way meta data is handled depends solely on the contents of the ``.meta`` attribute.
-
-Internal API
```
wtbarnes, line 89: Is there any part of this section we should preserve in the dev guide?

pytest.ini
```diff
@@ -65,3 +65,4 @@ filterwarnings =
     ignore:unclosed \<socket:ResourceWarning
     # Pending removal from sunpy 7.0
     ignore:The QueryResponse class is deprecated
+    ignore:The toplevel space of the io subpackage was never
```
wtbarnes, line 68: Are we sure that we're importing from file_tools everywhere and not the top level namespace? This would hide it if we are.

docs/reference/io.rst
```diff
@@ -1,48 +1,56 @@
...
+  - GENX
+  - ANA
+  - NOAA SWPC Solar Region Summary (SRS)
+  - ASDF
 
-.. automodapi:: sunpy.io.ana
+- Internal/Private API
+  - FITS
+  - JPEG2000
+  - CDF
```
wtbarnes, line 18: I understand the point here but if we want this to be  private API, why mention it at all?

nabobalis: Removed.

docs/reference/io.rst
nabobalis: I will make it one.

docs/reference/io.rst
nabobalis: We could start adding private api into our dev guide?

pytest.ini
nabobalis: Err, good question. I will do a check with this removed and find out.

sunpy/io/tests/test_filetools.py
```diff
@@ -21,101 +21,101 @@ def test_read_file_network_fits():
...
     with patch("astropy.io.fits.file.download_file", return_value=TEST_AIA_IMAGE) as mock:
-        hdulist = sunpy.io.read_file(url)
+        hdulist = read_file(url)
         assert mock.call_args[0][0] == url
     assert isinstance(hdulist, list)
     assert len(hdulist) == 1
     assert len(hdulist[0]) == 2
     assert isinstance(hdulist[0][0], np.ndarray)
-    assert isinstance(hdulist[0][1], sunpy.io.header.FileHeader)
+    assert isinstance(hdulist[0][1], sunpy.io._header.FileHeader)
```
wtbarnes, line 30: Won't this fail since you've removed `import sunpy.io`?

wtbarnes (approved):
I don't understand how some of these tests aren't throwing an `ImportError` but sure.

wtbarnes (changes requested):
Because this formally deprecates io, this needs a changelog.

wtbarnes (approved):
I'm an idiot. Nevermind. Looks good.

sunpy/io/tests/test_filetools.py
nabobalis: I still import it no?!

nabobalis: Why does that still work?!

It surely won't get it from `import sunpy` s?!

sunpy/io/__init__.py
```diff
@@ -1 +1,12 @@
...
+"""
+This subpackage contains a series of file readers and only for internal use.
+
+None of this code is intended to be used directly by a user and is not part of the sunpy public API
+and is not added to the API reference pages.
+"""
+from sunpy.util.exceptions import warn_deprecated
+
+warn_deprecated('The toplevel space of the io subpackage was never intended for public use and will be removed in the future.')
+from sunpy.io._file_tools import read_file, write_file
```
nabobalis, line 10: I think this is wrong. 

I get the message if I do:
```python
In [3]: import sunpy.io._header
WARNING: SunpyDeprecationWarning: The toplevel space of the io subpackage was never intended for public use and will be removed in the future. [sunpy.io]
```
I should not be raising this warning if someone does import that.

nabobalis: Or even:

```
import sunpy.io.special.asdf
```

As it stands, we can't merge this till I fix this.

sunpy/io/tests/test_filetools.py
nabobalis: I imported it for my sanity.

nabobalis:
TODO: Fix CI

Issue #7535, opened 2024-03-27
Document and possibly raise warning about apparent hard server side limit of 20,000 for helio client


samaloney:
### Describe the bug

Seems to be a (silent?) sever side limit on the max number of record the helio service can return. Only the soft `MaxRecords` limit is documented should probably check if a query hits the limit and raise warning?

### To Reproduce

```
query = Fido.search(a.Time('2022-01-01', '2022-02-01'),
        a.helio.TableName("gevloc_sxr_flare"), 
        a.helio.MaxRecords(99999))
len(query['hec'])
20000

query = Fido.search(a.Time('2022-01-01', '2022-01-15'), 
        a.helio.TableName("gevloc_sxr_flare"),
        a.helio.MaxRecords(99999))
len(query['hec'])
20000

query = Fido.search(a.Time('2022-01-01', '2022-01-10'), 
        a.helio.TableName("gevloc_sxr_flare"), 
        a.helio.MaxRecords(99999))
len(query['hec'])
19653
```

### System Details

```
#######
OS: Mac OS 14.3.1
Arch: 64bit, (arm)
sunpy: 5.0.0
Installation path: /Users/sm/.virtualenvs/ARCCnet/lib/python3.11/site-packages/sunpy-5.0.0.dist-info

Required Dependencies
#####################
astropy: 5.3.2
numpy: 1.25.2
packaging: 23.1
parfive: 2.0.2

Optional Dependencies
#####################
asdf: 2.15.1
asdf-astropy: 0.4.0
beautifulsoup4: 4.12.2
cdflib: 1.1.0
dask: 2023.9.0
drms: 0.6.4
glymur: 0.12.8
h5netcdf: 1.2.0
h5py: 3.9.0
lxml: 4.9.3
matplotlib: 3.7.2
mpl-animators: 1.1.0
pandas: 2.1.0
python-dateutil: 2.8.2
reproject: 0.11.0
scikit-image: 0.21.0
scipy: 1.11.2
sqlalchemy: 2.0.20
tqdm: 4.66.1
zeep: 4.2.1
```

### Installation method

pip

nabobalis:
We can have the MaxRecords attr issue a warning if the value is above 20000?

samaloney:
[...]

PR #7540, opened 2024-03-29
Add exception and warning for a.helio.MaxRecords

ahmedhosssam:
Fixes [#7535](https://github.com/sunpy/sunpy/issues/7535)

### DIFF
--- a/changelog/7540.bugfix.rst
+++ b/changelog/7540.bugfix.rst
@@ -1,1 +1,1 @@
-The maximum records in `~sunpy.net.helio.search` now are 20000.
+The maximum records in `~sunpy.net.helio.HECClient` now are 20000.
--- a/changelog/7540.bugfix.rst
+++ b/changelog/7540.bugfix.rst
@@ -1,0 +1,1 @@
+The maximum records in `~sunpy.net.helio.search` now are 20000.
--- a/sunpy/net/helio/hec.py
+++ b/sunpy/net/helio/hec.py
@@ -176,7 +176,7 @@ class HECClient(BaseClient):
         results = votable_handler(etree.tostring(results))
         table = HECResponse(results.to_table(), client=self)
         if len(table) == max_records:
-            warn_user("Number of results is the same as the " + ("default " if max_records == 500 else "") + "`a.helio.MaxRecords` limit of " + str(max_records) +
+            warn_user("Number of results is the same as current limit. "
                       ". It is possible your query has been truncated. "
                       "If you want to change this, set `a.helio.MaxRecords` to a higher value.")
         return table
--- a/sunpy/net/helio/hec.py
+++ b/sunpy/net/helio/hec.py
@@ -176,15 +176,9 @@ class HECClient(BaseClient):
         results = votable_handler(etree.tostring(results))
         table = HECResponse(results.to_table(), client=self)
         if len(table) == max_records:
-            if max_records == 500:
-                # 500 is the default value for max_records
-                warn_user("Number of results is the same as the default `max_records` of 500. "
-                          "It is possible your query has been truncated. "
-                          "If you want to change this, set `a.helio.MaxRecords` to a higher value.")
-            else:
-                warn_user("Number of results is the same as the `a.helio.MaxRecords` limit of " + str(max_records) +
-                          ". It is possible your query has been truncated. "
-                          "If you want to change this, set `a.helio.MaxRecords` to a higher value.")
+            warn_user("Number of results is the same as the " + ("default " if max_records == 500 else "") + "`a.helio.MaxRecords` limit of " + str(max_records) +
+                      ". It is possible your query has been truncated. "
+                      "If you want to change this, set `a.helio.MaxRecords` to a higher value.")
         return table
 
     def get_table_names(self):
--- a/sunpy/net/helio/hec.py
+++ b/sunpy/net/helio/hec.py
@@ -177,7 +177,7 @@ class HECClient(BaseClient):
         table = HECResponse(results.to_table(), client=self)
         if len(table) == max_records:
             warn_user("Number of results is the same as current limit. "
-                      ". It is possible your query has been truncated. "
+                      "It is possible your query has been truncated. "
                       "If you want to change this, set `a.helio.MaxRecords` to a higher value.")
         return table
 
--- a/sunpy/net/helio/hec.py
+++ b/sunpy/net/helio/hec.py
@@ -152,6 +152,8 @@ class HECClient(BaseClient):
             if isinstance(elem, a.Time):
                 qrdict['Time'] = elem
             elif isinstance(elem, ha.MaxRecords):
+                if elem.value > 20000:
+                    raise ValueError("The maximum allowed value for `a.helio.MaxRecords` is 20000.")
                 qrdict['max_records'] = elem.value
             elif isinstance(elem, ha.TableName):
                 qrdict['table_name'] = elem.value
@@ -173,10 +175,16 @@ class HECClient(BaseClient):
                                                     MAXRECORDS=max_records)
         results = votable_handler(etree.tostring(results))
         table = HECResponse(results.to_table(), client=self)
-        if len(table) == max_records == 500:
-            warn_user("Number of results is the same as the default `max_records` of 500. "
-                      "It is possible your query has been truncated. "
-                      "If you want to change this, set `a.helio.MaxRecords` to a higher value.")
+        if len(table) == max_records:
+            if max_records == 500:
+                # 500 is the default value for max_records
+                warn_user("Number of results is the same as the default `max_records` of 500. "
+                          "It is possible your query has been truncated. "
+                          "If you want to change this, set `a.helio.MaxRecords` to a higher value.")
+            else:
+                warn_user("Number of results is the same as the `a.helio.MaxRecords` limit of " + str(max_records) +
+                          ". It is possible your query has been truncated. "
+                          "If you want to change this, set `a.helio.MaxRecords` to a higher value.")
         return table
 
     def get_table_names(self):
--- a/sunpy/net/helio/hec.py
+++ b/sunpy/net/helio/hec.py
@@ -152,8 +152,6 @@ class HECClient(BaseClient):
             if isinstance(elem, a.Time):
                 qrdict['Time'] = elem
             elif isinstance(elem, ha.MaxRecords):
-                if elem.value > 20000:
-                    raise ValueError("Helio will only return a max of 20000 results.")
                 qrdict['max_records'] = elem.value
             elif isinstance(elem, ha.TableName):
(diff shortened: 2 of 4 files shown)


nabobalis:
Can you add a changelog and a unit test for this please?

sunpy/net/helio/hec.py
```diff
@@ -152,6 +152,8 @@ def search(self, *args, **kwargs):
             if isinstance(elem, a.Time):
                 qrdict['Time'] = elem
             elif isinstance(elem, ha.MaxRecords):
+                if elem.value > 20000:
+                    raise ValueError("The maximum allowed value for `a.helio.MaxRecords` is 20000.")
```
nabobalis, line 156:
```suggestion
                    raise ValueError("Helio will only return a max of 20000 results.")
```

sunpy/net/helio/tests/test_helio.py
```diff
@@ -293,6 +293,9 @@ def test_client_search(client):
     res = client.search(a.Time(start, end), a.helio.TableName(table_name), a.helio.MaxRecords(10))
     assert len(res) == 10
 
+    with pytest.raises(ValueError):
```
nabobalis, line 296: Can you match the message?

sunpy/net/helio/hec.py
```diff
@@ -173,9 +175,9 @@ def search(self, *args, **kwargs):
                                                     MAXRECORDS=max_records)
         results = votable_handler(etree.tostring(results))
         table = HECResponse(results.to_table(), client=self)
-        if len(table) == max_records == 500:
-            warn_user("Number of results is the same as the default `max_records` of 500. "
-                      "It is possible your query has been truncated. "
+        if len(table) == max_records:
+            warn_user("Number of results is the same as the " + ("default " if max_records == 500 else "") + "`a.helio.MaxRecords` limit of " + str(max_records) +
+                      ". It is possible your query has been truncated. "
```
nabobalis, line 180:
```suggestion
            warn_user("Number of results is the same as current limit. "
                      "It is possible your query has been truncated. "
```

How about just?

ahmedhosssam: Yes, but I was thinking about showing more information about the source of the warning.

sunpy/net/helio/tests/test_helio.py
ahmedhosssam: Done

sunpy/net/helio/hec.py
nabobalis: I think it's confusing currently?

nabobalis: This line will also need a unit test

changelog/7540.bugfix.rst
```diff
@@ -0,0 +1 @@
+The maximum records in `~sunpy.net.helio.search` now are 20000.
```
nabobalis, line 1:
```
<unknown>:88: WARNING: py:obj reference target not found: sunpy.net.helio.search
```
This will need to point to the client.

sunpy/net/helio/tests/test_helio.py
```diff
@@ -293,6 +294,14 @@ def test_client_search(client):
     res = client.search(a.Time(start, end), a.helio.TableName(table_name), a.helio.MaxRecords(10))
     assert len(res) == 10
 
+    with pytest.raises(ValueError, match="Helio will only return a max of 20000 results."):
```
nabobalis, line 297: Can this be made into a seperate offline test? 

It does not need to be added to this unit test.

sunpy/net/helio/tests/test_helio.py
```diff
@@ -293,6 +294,20 @@ def test_client_search(client):
     res = client.search(a.Time(start, end), a.helio.TableName(table_name), a.helio.MaxRecords(10))
     assert len(res) == 10
 
+    with pytest.warns(SunpyUserWarning, match="Number of results is the same as current limit. "
+                      "It is possible your query has been truncated. "
+                      "If you want to change this, set `a.helio.MaxRecords` to a higher value."):
+        res = client.search(a.Time(start, end), a.helio.TableName(table_name), a.helio.MaxRecords(10))
```
nabobalis, line 300: Why does this only get raised for this query?

ahmedhosssam: This query could return more than 10 records (I think it could return more than 500) so I think it's logical that it raises a warning when max record is equal to the length of the table.

nabobalis: But isn't the previous line the same?

ahmedhosssam: I think it would be better to put it in a separate test called `test_client_search_warning(client)`, right?
And `test_client_search(client)` would be as before.

nabobalis: Sure but I mean, isn't this test line the same as the one above?

Wouldn't both raise the warning, and if so, we don't need the second test?

ahmedhosssam: Yes it's the same, but it's for a different test purpose, that's why I see it's better to separate them.

nabobalis: So why would we need to repeat it?

ahmedhosssam: Yeah it's not good.

sunpy/net/helio/tests/test_helio.py
```diff
@@ -294,6 +295,26 @@ def test_client_search(client):
     assert len(res) == 10
 
 
+@pytest.mark.remote_data
+def test_client_search_warning(client):
```
nabobalis, line 299: What I was trying to say is that we don't need the same test twice.

nabobalis: Yes, is that not the case with this test?

ahmedhosssam: Yes, so it must be removed and just stick with `test_client_search`, right?

nabobalis: I would say so, is there a need to have the test repeated?

ahmedhosssam: No, I don't think it's possible for `warn_user` not to be raised if we reach `len(table) == max_records`.

sunpy/net/helio/hec.py
```diff
@@ -173,8 +175,8 @@ def search(self, *args, **kwargs):
                                                     MAXRECORDS=max_records)
         results = votable_handler(etree.tostring(results))
         table = HECResponse(results.to_table(), client=self)
-        if len(table) == max_records == 500:
-            warn_user("Number of results is the same as the default `max_records` of 500. "
+        if len(table) == max_records:
```
nabobalis, line 178: Do we not need to modify the `test_client_search` to trigger this warning?

sunpy/net/helio/hec.py
```diff
@@ -152,6 +152,8 @@ def search(self, *args, **kwargs):
             if isinstance(elem, a.Time):
                 qrdict['Time'] = elem
             elif isinstance(elem, ha.MaxRecords):
+                if elem.value > 20000:
+                    raise ValueError("Helio will only return a max of 20000 results.")
```
nabobalis, line 156: @samaloney do want a warning or an error?

samaloney: Yea I think should raise and error as setting it any higher doesn't currently return anymore results but I don't have really strong view either way.

nabobalis: Error it is.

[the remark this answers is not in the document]
nabobalis: I think this would be better as well.

sunpy/net/helio/tests/test_helio.py
```diff
@@ -294,6 +294,14 @@ def test_client_search(client):
     assert len(res) == 10
 
 
+def test_client_search_limit(client):
```
nabobalis, line 297: This test can be reduced to just testing the attr directly instead of doing a search.

sunpy/net/helio/hec.py
```diff
@@ -173,8 +173,8 @@ def search(self, *args, **kwargs):
                                                     MAXRECORDS=max_records)
         results = votable_handler(etree.tostring(results))
         table = HECResponse(results.to_table(), client=self)
-        if len(table) == max_records == 500:
-            warn_user("Number of results is the same as the default `max_records` of 500. "
+        if len(table) == max_records:
+            warn_user("Number of results is the same as current limit. "
```
nabobalis, line 177: So now we just need a unit test that hits this message and I think this PR will be good to go.

ahmedhosssam: Is it possible for `assert len(res) == 10` not to trigger the warning?

nabobalis: It doesn't seem to?

ahmedhosssam: I think we can hits the warning like this:
```python
@pytest.mark.remote_data
def test_client_search(client):
    start = '2005/01/03'
    end = '2005/12/03'
    table_name = 'rhessi_hxr_flare'
    with pytest.warns(SunpyUserWarning, match="Number of results is the same as current limit. "
                                         "It is possible your query has been truncated. "
                       "If you want to change this, set `a.helio.MaxRecords` to a higher value."):
        res = client.search(a.Time(start, end), a.helio.TableName(table_name), a.helio.MaxRecords(10))
    assert len(res) == 10
``` 

But, should we keep `assert len(res) == 10`?

sunpy/net/helio/tests/test_helio.py
```diff
@@ -294,6 +295,11 @@ def test_client_search(client):
     assert len(res) == 10
 
 
+def test_client_search_limit():
```
nabobalis, line 298:
```suggestion
def test_max_records_limit():
```

sunpy/net/helio/tests/test_helio.py
```diff
@@ -5,6 +5,7 @@
 from requests.exceptions import SSLError
 
 from sunpy.net import attrs as a
+from sunpy.net.helio import attrs as ha
```
nabobalis, line 8: Do we need this?

nabobalis (approved):

nabobalis:
Thanks for the PR @ahmedhosssam

nabobalis:
Looks like there is a fail on the CI:

```
FAILED ../../.tox/py312-online/lib/python3.12/site-packages/sunpy/net/helio/tests/test_helio.py::test_HECResponse_iter - sunpy.util.exceptions.SunpyUserWarning: Number of results is the same as cu...
```

nabobalis:
The failing tests are not due to this PR.

nabobalis (approved):

PR #7543, opened 2024-04-01
Adds ValueError to mapbase.submap() when Helioprojection coordinate contains NaN values.

### LINKED ISSUE #5598: Add a helpful error message to `GenericMap.submap()` for helioprojective-related NaNs
It's a recurring source of confusion for users when they try to submap a helioprojective map from one observatory using 2D helioprojective coordinates from a different observatory (or the same observatory at a different time) and one of the corners is off-disk.  Even if the observer-location difference or the observation-time difference is small, formally there is no unambiguous conversion of an off-disk 2D coordinate to a different helioprojective frame because we don't know what assumption to make the coordinate 3D.  (For an on-disk 2D coordinate, we make the broadly safe assumption that the coordinate is on the solar surface in 3D space.)  The coordinate transformation of the corner point returns NaNs, and `submap()` errors ungracefully.

We should add an error message to `submap()` so that it checks if it is about to work with NaNs and tells the user that they likely want to use `Helioprojective.assume_spherical_screen()` so that the off-disk 2D coordinate can be converted to a 3D coordinate.

grahamasam:
This pull request produces a value error when Helioprojection coordinates that have NaN distance values are input as bottom_left or top_right to the submap() function.

Fixes #5598 

Additional Info:
I'm unsure if this solution fully addresses the common user issue presented in #5598. A testcase that mimics one of these specific user issues could be useful to determine what other situations should result in the ValueError. Any suggestions would be greatly appreciated!


sunpy/map/mapbase.py
```diff
@@ -1995,6 +1995,21 @@ def _parse_submap_quantity_input(self, bottom_left, top_right, width, height):
     @_parse_submap_input.register(SkyCoord)
     @_parse_submap_input.register(BaseCoordinateFrame)
     def _parse_submap_coord_input(self, bottom_left, top_right, width, height):
+        # If bottom_left or top_right is a helioprojective coordinate with NaN values,
+        # cannot create meaningful submap
```
nabobalis, line 1999:
```suggestion
```

sunpy/map/mapbase.py
```diff
@@ -1995,6 +1995,21 @@ def _parse_submap_quantity_input(self, bottom_left, top_right, width, height):
...
+        if (isinstance(bottom_left, Helioprojective)):
+            values = bottom_left.distance.to_value()
```
nabobalis, line 2001: Are the values needed for anything else? If not, we could use the new walrus operator?

sunpy/map/mapbase.py
```diff
@@ -1995,6 +1995,21 @@ def _parse_submap_quantity_input(self, bottom_left, top_right, width, height):
...
+            if np.isnan(values).any():
+              raise ValueError("The Helioprojective coordinates input to `submap()` contain NaN values."
```
nabobalis, line 2003: I think this message is the same as below? If so, I think defining it once above would be better.

sunpy/map/mapbase.py
```diff
@@ -1994,6 +1994,18 @@ def _parse_submap_quantity_input(self, bottom_left, top_right, width, height):
     @_parse_submap_input.register(SkyCoord)
     @_parse_submap_input.register(BaseCoordinateFrame)
     def _parse_submap_coord_input(self, bottom_left, top_right, width, height):
+        msg = (
+            "The {quadrant} input contains NaN values. "
+            "It is possible the coordinates are off-disk. "
+            "Consider using Helioprojective.assume_spherical_screen"
```
wtbarnes, line 2000: If we merge #7115, this should change to `SphericalScreen`

changelog/7543.feature.rst
```diff
@@ -0,0 +1 @@
+Updated :meth:`sunpy.map.Map.submap` to check if it is about to work on locations with NaNs now errors and informs the user that they likely want to use :meth:`~sunpy.coordinates.Helioprojective.assume_spherical_screen` so that the off-disk 2D coordinate can be converted to a 3D coordinate.
```
wtbarnes, line 1: Same as below. Should be updated to use `SphericalScreen` once that PR is merged.

sunpy/map/mapbase.py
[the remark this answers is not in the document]
nabobalis: Why am I not allowed?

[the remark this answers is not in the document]
nabobalis: I added more text to the message, see if you like it.

nabobalis:
TODO:
- [ ] Rebase this PR when https://github.com/sunpy/sunpy/pull/7115 is merged in

ayshih (changes requested):
Oops, I guess I haven't looked at this PR before.  As guessed by the author, this PR doesn't actually address the common use case of #5598 because the implementation is at the "wrong" place in the code.  (The included unit tests do not reflect common usage.)  I'll try to provide better guidance soon if someone else doesn't do so first.

ayshih:
Here's an example of the common use case:
```python
>>> import astropy.units as u

>>> import sunpy.map
>>> from sunpy.data.sample import AIA_171_IMAGE

>>> aiamap = sunpy.map.Map(AIA_171_IMAGE)
```
The following submap that includes off-disk pixels works fine because it is in the native coordinate frame, so 2D coordinates do not need to be promoted:
```python
>>> coord_native = SkyCoord(0*u.arcsec, 0*u.arcsec, frame=aiamap.coordinate_frame)
>>> aiamap.submap(coord_native, width=1000*u.arcsec, height=1000*u.arcsec)
```
The following, very similar submap fails because the observer is slightly different, so 2D coordinates need to be transformed to the native coordinate frame, and thus the off-disk 2D corners get a NaN distance when internally promoted.
```python
>>> coord_other = SkyCoord(0*u.arcsec, 0*u.arcsec, frame='helioprojective', observer='earth', obstime=aiamap.date)
>>> aiamap.submap(coord_other, width=1000*u.arcsec, height=1000*u.arcsec)
...
ValueError: cannot convert float NaN to integer
```
The check needs to be on the output of `self._parse_submap_input()` right here:
https://github.com/sunpy/sunpy/blob/e7536b1ddd34eaa9a5f6990627a82f89ab64c6be/sunpy/map/mapbase.py#L1912-L1913

sunpy/map/mapbase.py
```diff
@@ -1912,6 +1912,18 @@ def submap(self, bottom_left, *, top_right=None, width: (u.deg, u.pix) = None, h
...
             bottom_left, top_right, width, height)).T
 
+        msg = (
+            "The provided input coordinates to ``submap`` when transformed to the target "
+            "coordinate frame contain NaN values and are possibly off-disk. "
+            "This means that submap can not crop the map to the desired region, "
+            "as there are no valid pixel coordinates to crop to."
+            "The context manager: Helioprojective.assume_spherical_screen can be used to "
+            "transform the coordinates to a spherical screen where the NaN values are "
+            "transformed as there is a spherical screen where they are valid coordinates."
```
ayshih, line 1922:
```suggestion
            "The provided input coordinates to ``submap`` when transformed to the target "
            "coordinate frame contain NaN values and cannot be used to crop the map. "
            "The most common reason for NaN values is transforming off-disk 2D "
            "coordinates without specifying an assumption (e.g., via the"
            "`Helioprojective.assume_spherical_screen()` context manager) that allows "
            "such coordinates to be interpreted as 3D coordinates."
```

ayshih (approved):
Revised the warning message

nabobalis (approved):

nabobalis:
Thanks for the PR @grahamasam and @ayshih for the review.

changelog/7543.feature.rst
nabobalis, line 1:
```suggestion
Updated :meth:`sunpy.map.GenericMap.submap` to check if it is about to work on locations with NaNs now errors and informs the user that they likely want to use :meth:`~sunpy.coordinates.Helioprojective.assume_spherical_screen` so that the off-disk 2D coordinate can be converted to a 3D coordinate.
```

PR #7545, opened 2024-04-03
Fixed the GEI->HME transformation to preserve the length unit of the coordinate

ayshih:
Thanks to Bob Weigel for reporting this bug!  It turns out that one of our transformation steps – `GeocentricEarthEquatorial` (GEI) to `HeliocentricMeanEcliptic` (HME or HAE) – inadvertently did not preserve the length unit of the coordinate, but instead always returned the coordinate in AU.  (I never noticed because I probably tested that transformation only with coordinates where AU was already the length unit.)  While changing the unit does not make the answer wrong, it can be inconvenient for the user.

This PR makes the amusingly silly-looking fix, and importantly adds a comprehensive unit test that covers each of our transformation steps.

### DIFF
--- a/sunpy/coordinates/_transformations.py
+++ b/sunpy/coordinates/_transformations.py
@@ -1071,7 +1071,7 @@ def gei_to_hme(geicoord, hmeframe):
     earth_object_int = geicoord.cartesian.transform(rot_matrix)
 
     # Find the Sun-object vector in the intermediate frame
-    sun_object_int = sun_earth_int + earth_object_int
+    sun_object_int = earth_object_int + sun_earth_int  # add in this order to preserve the original units
     int_coord = int_frame.realize_frame(sun_object_int)
 
     # Convert to the final frame through HCRS
--- a/changelog/7545.bugfix.rst
+++ b/changelog/7545.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug with any coordinate transformation starting in `~sunpy.coordinates.GeocentricEarthEquatorial` (GEI) returning output with AU as the length unit, rather than preserving the length unit of the initial coordinate.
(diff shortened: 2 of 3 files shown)


ayshih:
I should add that the GEI->HME step is required to convert from GEI to any other frame, so this bug affects all transformations which start from GEI.

sunpy/coordinates/_transformations.py
```diff
@@ -1071,7 +1071,7 @@ def gei_to_hme(geicoord, hmeframe):
     earth_object_int = geicoord.cartesian.transform(rot_matrix)
 
     # Find the Sun-object vector in the intermediate frame
-    sun_object_int = sun_earth_int + earth_object_int
+    sun_object_int = earth_object_int + sun_earth_int
```
nabobalis, line 1074: Why does this fix the bug?!

nabobalis (approved):

sunpy/coordinates/_transformations.py
ayshih: Yup, abstraction makes this look silly.  What happens when you add two unit-bearing vectors is that the second addend is converted to the unit of the first addend, then added, so the output will always have the unit of the first addend.  In this case, `earth_object_int` has the initial units, and `sun_earth_int` has AU units, so we want `earth_object_int` first.  We of course could instead explicitly transform to the unit afterwards, but it's unnecessary as long as we're intentional about addend order.

nabobalis: That makes sense, thanks for the explanation.
I wasn't sure what the type of either earth_object_int and sun_earth_int.

ayshih: They are both `CartesianRepresentation`, with the relevant behavior actually coming from `Quantity`

Cadair: I think what @samaloney said in chat is valid, we should add a comment here.
```suggestion
    # Add sun_earth_int to earth_object_int to maintain the units of earth_object_int in the result.
    sun_object_int = earth_object_int + sun_earth_int
```

ayshih: Added a comment

alasdairwilson (approved):

PR #7547, opened 2024-04-03
[map branch] Migrate `GenericMap` to inherit from `NDCube`

Cadair:
This PR is made from a feature branch[1] which is going to form the core of the NDCube sprint work at the coordination meeting in a few weeks.

The first commit moves all the metadata model properties out on to a mixin class, as the plan is to not touch them during the refactor.

---

[1] oh wow it's been a while since we did this. EDIT FROM NABIL: NOT TRUE


sunpy/util/decorators.py
```diff
@@ -356,6 +356,7 @@ def outer(prop):
         """
         prop: the property method being decorated
         """
+        return prop
```
Cadair, line 359: We need to revert this.

Issue #7550, opened 2024-04-04
sunpy 5.1.1 + pytest 8 on armel & mips64el error


mr-c:
### Describe the bug

Building sunpy 5.1.1 with Pytest 8.1.1 + https://github.com/sunpy/sunpy/pull/7429 we still get two errors:

On armel (arm32v5): `test_map_arithmetic_operations_raise_exceptions[map-warn_context0]` fails, `DID NOT WARN. No warnings of type (<class 'RuntimeWarning'>,) were emitted.`

https://buildd.debian.org/status/fetch.php?pkg=sunpy&arch=armel&ver=5.1.1-4&stamp=1712238082&raw=0

On mips64el: `[doctest] sunpy.coordinates.frames.Helioprojective.assume_spherical_screen`: 
``` 
Differences (unified diff with -expected +actual):
    @@ -2,4 +2,4 @@
         [(   0., 0., 0.99660825), ( 319., 0., 0.99687244),
          ( 638., 0., 0.99778472), ( 957., 0., 1.00103285),
    -     (1276., 0., 1.00125872), (1595., 0., 1.00125872),
    -     (1914., 0., 1.00125872)]>
    +     (1276., 0.,        nan), (1595., 0.,        nan),
    +     (1914., 0.,        nan)]>
```
https://buildd.debian.org/status/fetch.php?pkg=sunpy&arch=mips64el&ver=5.1.1-4&stamp=1712238826&raw=0

Any guidance on how we shall fix this?

### To Reproduce

Here's a dockerfile, build it with `--platform=linux/arm/v5` or `--platform=linux/mips64le` 

``` Dockerfile
FROM docker.io/debian:sid-slim

WORKDIR /build

RUN apt-get update && apt-get install -y --no-install-recommends devscripts curl debian-keyring
RUN dget https://deb.debian.org/debian/pool/main/s/sunpy/sunpy_5.1.1-4.dsc
WORKDIR /build/sunpy-5.1.1
RUN apt-get -y build-dep .
RUN dpkg-buildpackage -uc -us -b
```

For reference, all the extra patches we apply to Sunpy 5.1.1 are at https://salsa.debian.org/debian-astro-team/sunpy/-/tree/master/debian/patches?ref_type=heads

### System Details

```
==============================
sunpy Installation Information
==============================

General
#######
OS: Debian GNU/Linux (trixie/sid, Linux 6.1.0-18-amd64)
Arch: 64bit, ()
sunpy: 5.1.1
[...]

PR #7551, opened 2024-04-04
Fixed an unnecessary division computation for an unsupported division by a Map

ayshih:
Dividing a map by something is implemented as multiplying the map by the inverse of that something.  When that something is also a map, the operation is not supported.  We weren't checking whether the division is supported, and instead relying on checking whether the latter multiplication is supported, but that means that the inverse is computed even when the overall operation is not supported.  This PR simply makes us check the division upfront.

This PR also happens to fix part of #7550

### DIFF
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -3,7 +3,6 @@ Test Generic Map
 """
 import re
 import tempfile
-import contextlib
 
 import matplotlib
 import matplotlib.pyplot as plt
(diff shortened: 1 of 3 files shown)


sunpy/map/tests/test_mapbase.py
```diff
@@ -1732,23 +1732,15 @@ def test_map_arithmetic_neg(aia171_test_map):
...
-    ('foobar', contextlib.nullcontext()),
-    (None, contextlib.nullcontext()),
-    (['foo', 'bar'], contextlib.nullcontext()),
-])
-def test_map_arithmetic_operations_raise_exceptions(aia171_test_map, value, warn_context):
+@pytest.mark.parametrize("value", ['map', 'foobar', None, ['foo', 'bar']])
+def test_map_arithmetic_operations_raise_exceptions(aia171_test_map, value):
     value = aia171_test_map if value == 'map' else value
     with pytest.raises(TypeError):
         _ = aia171_test_map + value
```
nabobalis, line 1738: Do we need these returns?

ayshih:
Of course, as I look at this PR, I wonder why our code goes through with the division when it's going to raise a TypeError anyway...

sunpy/map/tests/test_mapbase.py
ayshih: Well, no, hence the assignment to `_`, but I guess you're asking whether we can just delete the `_ =`?  Does a linter complain if we remove those?

ayshih:
Ah, even better, it's actually a bug in our code.  Stay tuned.

sunpy/map/tests/test_mapbase.py
nabobalis: Yeah, can we just delete those returns? 

If the linter complains, it should be thrown out.

nabobalis:
If that is the case, do we want to backport this bugfix PR?

ayshih:
New version

nabobalis (approved):

sunpy/map/tests/test_mapbase.py
```diff
@@ -1732,23 +1731,15 @@ def test_map_arithmetic_neg(aia171_test_map):
...
 
 
-@pytest.mark.parametrize(("value", "warn_context"), [
-    ('map', pytest.warns(RuntimeWarning)),
-    ('foobar', contextlib.nullcontext()),
-    (None, contextlib.nullcontext()),
-    (['foo', 'bar'], contextlib.nullcontext()),
-])
-def test_map_arithmetic_operations_raise_exceptions(aia171_test_map, value, warn_context):
+@pytest.mark.parametrize("value", ['map', 'foobar', None, ['foo', 'bar']])
```
nabobalis, line 1734: I assume adding a np.nan here would be pointless?

wtbarnes (approved):

Issue #7555, opened 2024-04-05
Build and test against numpy 2.0rc1


Cadair:
### Describe the feature

We need to bump our build deps and test against 2.0.0rc1 at runtime. https://github.com/numpy/numpy/issues/24300#issuecomment-2030603395

We should do this before the next bugfix release.

pllim:
FWIW, also failed here: https://github.com/astropy/astropy-integration-testing/actions/runs/8568043943/job/23485761643?pr=16

xref https://github.com/astropy/astropy-integration-testing/pull/16

nabobalis:
PANDAS!!!!!

PR #7556, opened 2024-04-05
fix sunpy and numpy 2.0

alasdairwilson:
Fixes https://github.com/sunpy/sunpy/issues/7555

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
     secrets:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
-  test:
+  tests:
     needs: [core]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@main
     with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
     secrets:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
-  tests:
+  test:
     needs: [core]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@main
     with:
@@ -71,7 +71,8 @@ jobs:
     needs: [docs]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@main
     with:
-      default_python: '3.12'
+      # Due to mplcario, we need to use Python 3.10 for the wheels
+      default_python: '3.10'
       submodules: false
       coverage: codecov
       toxdeps: tox-pypi-filter
@@ -87,7 +88,6 @@ jobs:
             apt:
               - libopenjp2-7
               - graphviz
-              - libcairo2-dev
         - linux: py312-online
           posargs: -n auto --dist loadgroup --color=yes
     secrets:
@@ -150,7 +150,7 @@ jobs:
         shell: bash -el {0}
         run: |
           pip install --no-deps --no-build-isolation .
-      - name: Run tests
+      - name: Run test
         shell: bash -el {0}
         run: |
           conda list
@@ -174,7 +174,7 @@ jobs:
         github.event_name == 'pull_request' &&
         contains(github.event.pull_request.labels.*.name, 'Run publish')
       )
-    needs: [tests, docs]
+    needs: [test, docs]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@main
     with:
       upload_to_anaconda: ${{ (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
     needs: [core]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@main
     with:
-      default_python: '3.10'
+      default_python: '3.12'
       submodules: false
       pytest: false
       toxdeps: tox-pypi-filter
@@ -71,7 +71,7 @@ jobs:
     needs: [docs]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@main
     with:
-      default_python: '3.10'
+      default_python: '3.12'
       submodules: false
       coverage: codecov
       toxdeps: tox-pypi-filter
@@ -100,14 +100,14 @@ jobs:
       )
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@main
     with:
-      default_python: '3.10'
+      default_python: '3.12'
       submodules: false
       coverage: codecov
       toxdeps: tox-pypi-filter
       posargs: -n auto --color=yes
       envs: |
         - linux: base_deps
-        - linux: py311-devdeps
+        - linux: py312-devdeps
     secrets:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
@@ -122,7 +122,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: '3.10'
+          python-version: '3.12'
       - run: python -m pip install -U --user build
       - run: python -m build . --sdist
       - run: python -m pip install -U --user twine
@@ -144,7 +144,7 @@ jobs:
         with:
           activate-environment: sunpy-test
           environment-file: sunpy-dev-env.yml
-          python-version: "3.11"
+          python-version: "3.12"
       - name: Install sunpy
         shell: bash -el {0}
         run: |
@@ -196,7 +196,7 @@ jobs:
     needs: [publish]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@main
     with:
-      python-version: "3.10"
+      python-version: "3.12"
       test_extras: 'tests'
       test_command: 'pytest -p no:warnings --doctest-rst -m "not mpl_image_compare" --pyargs sunpy'
       submodules: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,6 +87,7 @@ jobs:
             apt:
               - libopenjp2-7
               - graphviz
+              - cairo
         - linux: py312-online
           posargs: -n auto --dist loadgroup --color=yes
     secrets:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,7 @@ jobs:
             apt:
               - libopenjp2-7
               - graphviz
-              - cairo
+              - libcairo2-dev
         - linux: py312-online
           posargs: -n auto --dist loadgroup --color=yes
     secrets:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -174,7 +174,7 @@ jobs:
         github.event_name == 'pull_request' &&
         contains(github.event.pull_request.labels.*.name, 'Run publish')
       )
-    needs: [test, docs]
+    needs: [tests, docs]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@main
     with:
       upload_to_anaconda: ${{ (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') }}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ requires = [
   "setuptools_scm[toml]>=6.2",
   "wheel",
   "extension-helpers",
-  "numpy>=1.25"
+  "numpy>=2.0.0rc1,<2.3",
 ]
 build-backend = "setuptools.build_meta"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ requires = [
   "setuptools_scm[toml]>=6.2",
   "wheel",
   "extension-helpers",
-  "numpy>=2.0.0rc1,<2.3",
+  "numpy>=2.0.0rc1",
 ]
 build-backend = "setuptools.build_meta"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,17 @@ requires = [
   "setuptools_scm[toml]>=6.2",
   "wheel",
   "extension-helpers",
-  "numpy>=2.0.0rc1",
+  # Comments on numpy build requirement range:
+  #
+  #   1. >=2.0.x is the numpy requirement for wheel builds for distribution
+  #      on PyPI - building against 2.x yields wheels that are also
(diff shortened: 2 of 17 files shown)


nabobalis:
Is this why devdeps was failing?

alasdairwilson:
> Is this why devdeps was failing?

I think so, we will see if it passes now

nabobalis (approved):

sunpy/map/mapbase.py
```diff
@@ -529,7 +529,7 @@ def _get_lon_lat(self, frame):
     @property
     def quantity(self):
         """Unitful representation of the map data."""
-        return u.Quantity(self.data, self.unit, copy=False)
+        return u.Quantity(self.data, self.unit, copy=True)
```
nabobalis, line 532: Probs dont want this but lmao

pytest.ini
```diff
@@ -65,3 +65,6 @@ filterwarnings =
     ignore:unclosed \<socket:ResourceWarning
     # Pending removal from sunpy 7.0
     ignore:The QueryResponse class is deprecated
+    # These should have been fixed by numpy 2.0 but possible its due to the
+    # other packages not building agasint it yet?
+    ignore:.*may indicate binary incompatibility.*
```
nabobalis, line 70: This is wrong, as now they are valueerrors instead of warnings.

This will DO NOTHING

pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [build-system]
 requires = [
   "extension-helpers",
-  "numpy>=1.25",
+  "numpy>=2.0.0rc1,<2.3",
```
nabobalis, line 4:
```suggestion
  "numpy>=2.0.0rc1",
```

pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [build-system]
 requires = [
   "extension-helpers",
-  "numpy>=2.0.0rc1,<2.3",
+  "numpy>=2.0.0rc1",
```
alasdairwilson, line 4: Why was there a <2.3 in the first place?

alasdairwilson:
Naive me of 8 hours ago thinking just change pyerfa and all will be well.

pyproject.toml
nabobalis: numpy suggested to use that on release branches and I was going to backport this but then i thought, who cares what numpy thinks.

sunpy/image/tests/test_transform.py
```diff
@@ -1,3 +1,4 @@
+
```
nabobalis, line 1:
```suggestion
```

sunpy/map/mapbase.py
alasdairwilson: Now loads of this has been fixed can this be reverted?

nabobalis: yeah I should do that.

sunpy/time/tests/test_time.py
```diff
@@ -150,14 +150,6 @@ def test_parse_time_individual_numpy_datetime():
     assert dt == Time('2005-02-01', format='isot')
 
 
-def test_parse_time_numpy_datetime_timezone():
```
nabobalis, line 153: I removed this test as numpy plan to remove parsing timezone aware datetimes, so.

sunpy/map/mapbase.py
nabobalis: Doesn't seem to work:

```
>       value = np.array(
            value, dtype=dtype, copy=copy, order=order, subok=True, ndmin=ndmin
        )
E       ValueError: Unable to avoid copy while creating an array as requested.
E       If using `np.array(obj, copy=False)` replace it with `np.asarray(obj)` to allow a copy when needed (no behavior change in NumPy 1.x).
E       For more details, see https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword.

../../.tox/py312-devdeps/lib/python3.12/site-packages/astropy/units/quantity.py:544: ValueError
```

Not sure if this needs to be patched more in astropy?

braingram: I believe the `copy` change was to follow `numpy`s lead. See:
https://github.com/astropy/astropy/issues/16167
Several `astropy` downstream packages have received updates to accommodate this change like the following:
https://github.com/astropy/specutils/commit/5ec4931ff66454b545ccf45013b03b1265b97709

nabobalis: Thanks for the context, I can follow specutils.

nabobalis:
I would expect this to not backport nicely.

CyclingNinja (approved):

sunpy/coordinates/tests/test_frames.py
```diff
@@ -224,12 +224,13 @@ def test_hpc_default_observer():
 
 
 def test_hpc_low_precision_float_warning():
-    hpc = Helioprojective(u.Quantity(0, u.deg, dtype=np.float32),
-                          u.Quantity(0, u.arcsec, dtype=np.float16),
-                          observer=HeliographicStonyhurst(0*u.deg, 0*u.deg, 1*u.AU))
+    with np.errstate(over='ignore'):
```
Cadair, line 227: What's this for ooi?

sunpy/time/timeformats.py
```diff
@@ -80,10 +80,10 @@ class TimeTaiSeconds(TimeFromEpoch):
     --------
     >>> from astropy.time import Time
     >>> t = Time('1958-01-01T00:00:00', format='isot', scale='tai')
-    >>> t.tai_seconds
+    >>> t.tai_seconds  # doctest: +SKIP
```
Cadair, line 83: We should open a tracking issue for when numpy 2.0 is released to un-skip these (and anything else which will need fixing).

Cadair (approved):
Technically looks fine. Is it too early to add user docs about lack of support for opencv / cdflib? Is there a tracking issue on cdflib?

sunpy/time/timeformats.py
nabobalis: https://github.com/sunpy/sunpy/issues/7566

sunpy/coordinates/tests/test_frames.py
nabobalis: For this:

```
=================================== FAILURES ===================================
_____________________ test_hpc_low_precision_float_warning _____________________
[gw0] linux -- Python 3.12.2 /home/runner/work/sunpy/sunpy/.tox/py312-devdeps/bin/python

    def test_hpc_low_precision_float_warning():
>       hpc = Helioprojective(u.Quantity(0, u.deg, dtype=np.float32),
                              u.Quantity(0, u.arcsec, dtype=np.float16),
                              observer=HeliographicStonyhurst(0*u.deg, 0*u.deg, 1*u.AU))

../../.tox/py312-devdeps/lib/python3.12/site-packages/sunpy/coordinates/tests/test_frames.py:227: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../.tox/py312-devdeps/lib/python3.12/site-packages/sunpy/coordinates/frames.py:151: in __init__
    super().__init__(*args, **kwargs)
../../.tox/py312-devdeps/lib/python3.12/site-packages/astropy/coordinates/baseframe.py:317: in __init__
    data = self._infer_data(args, copy, kwargs)  # possibly None.
../../.tox/py312-devdeps/lib/python3.12/site-packages/astropy/coordinates/baseframe.py:470: in _infer_data
    representation_data = representation_cls(copy=copy, **repr_kwargs)
../../.tox/py312-devdeps/lib/python3.12/site-packages/astropy/coordinates/representation/spherical.py:55: in __init__
    super().__init__(lon, lat, differentials=differentials, copy=copy)
../../.tox/py312-devdeps/lib/python3.12/site-packages/astropy/coordinates/representation/base.py:643: in __init__
    super().__init__(*args, **kwargs)
../../.tox/py312-devdeps/lib/python3.12/site-packages/astropy/coordinates/representation/base.py:205: in __init__
    self.attr_classes[component](attr, copy=copy, subok=True)
../../.tox/py312-devdeps/lib/python3.12/site-packages/astropy/coordinates/angles/core.py:604: in __new__
    self._validate_angles()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Latitude 0. arcsec>, angles = <Latitude 0. arcsec>

    def _validate_angles(self, angles=None):
        """Check that angles are between -90 and 90 degrees.
        If not given, the check is done on the object itself.
        """
        # Convert the lower and upper bounds to the "native" unit of
        # this angle.  This limits multiplication to two values,
        # rather than the N values in `self.value`.  Also, the
        # comparison is performed on raw arrays, rather than Quantity
        # objects, for speed.
        if angles is None:
            angles = self
    
        # For speed, compare using "is", which is not strictly guaranteed to hold,
        # but if it doesn't we'll just convert correctly in the 'else' clause.
        if angles.unit is u.deg:
            limit = 90
        elif angles.unit is u.rad:
            limit = 0.5 * np.pi
        else:
            limit = u.degree.to(angles.unit, 90.0)
    
        angles_view = angles.view(np.ndarray)
        if NUMPY_LT_2_0:
            # Ensure ndim>=1 so that comparison is done using the angle dtype.
            # Otherwise, e.g., np.array(np.pi/2, 'f4') > np.pi/2 will yield True.
            angles_view = angles_view[np.newaxis]
    
>       if np.any(np.abs(angles_view) > limit):
E       RuntimeWarning: overflow encountered in cast
```

nabobalis:
Looks like an arm tolerance issue:

```
===================================================================================== FAILURES =====================================================================================
  _______________________________________________________________________________ test_earth_location ________________________________________________________________________________
  
  gong_halpha = <sunpy.map.sources.gong.GONGHalphaMap object at 0x149a85bd0>
  SunPy Map
  ---------
  Observatory:		 NSO-GONG
  Instrument:		...     ...,
         [0, 0, 0, ..., 0, 0, 0],
         [0, 0, 0, ..., 0, 0, 0],
         [0, 0, 0, ..., 0, 0, 0]], dtype=int16)
  
      def test_earth_location(gong_halpha):
  >       assert_equal(gong_halpha._earth_location.lat, 34.2603299[816](https://github.com/sunpy/sunpy/actions/runs/8626444999/job/23644912294?pr=7556#step:7:827)7749*u.deg)
  
  ../venv-test-arm64/lib/python3.10/site-packages/sunpy/map/sources/tests/test_gong_halpha_source.py:48: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  
  args = (<built-in function eq>, <Latitude 34.26032998 deg>, <Quantity 34.26032998 deg>), kwds = {'err_msg': '', 'header': 'Arrays are not equal', 'strict': False, 'verbose': True}
  
      @wraps(func)
      def inner(*args, **kwds):
          with self._recreate_cm():
  >           return func(*args, **kwds)
  E           AssertionError: 
  E           Arrays are not equal
  E           
  E           Mismatched elements: 1 / 1 (100%)
  E           Max absolute difference: 7.10542736e-15
  E           Max relative difference: 2.07395182e-16
  E            x: Latitude(34.26033, unit='deg')
  E            y: Quantity(34.26033, unit='deg')
  
  /Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/contextlib.py:79: AssertionError
  ============================================================================= short test summary info ==============================================================================
  FAILED map/sources/tests/test_gong_halpha_source.py::test_earth_location - AssertionError: 
  Arrays are not equal
  
  Mismatched elements: 1 / 1 (100%)
  Max absolute difference: 7.10542736e-15
  Max relative difference: 2.07395182e-16
   x: Latitude(34.26033, unit='deg')
   y: Quantity(34.26033, unit='deg')
```

.github/workflows/ci.yml
```diff
@@ -87,6 +87,7 @@ jobs:
             apt:
               - libopenjp2-7
               - graphviz
+              - cairo
```
nabobalis, line 90:
```suggestion
              - libcairo2-dev
```

pyproject.toml
```diff
@@ -4,7 +4,7 @@ requires = [
   "setuptools_scm[toml]>=6.2",
   "wheel",
   "extension-helpers",
-  "numpy>=1.25"
+  "numpy>=2.0.0rc1",
```
nabobalis, line 7:
```suggestion
  # Comments on numpy build requirement range:
  #
  #   1. >=2.0.x is the numpy requirement for wheel builds for distribution
  #      on PyPI - building against 2.x yields wheels that are also
  #      ABI-compatible with numpy 1.x at runtime.
  #   2. Note that building against numpy 1.x works fine too - users and
  #      redistributors can do this by installing the numpy version they like
  #      and disabling build isolation.
  #   3. The <2.3 upper bound is for matching the numpy deprecation policy,
  #      it should not be loosened.
  "numpy>=2.0.0rc1,<2.3",
```

nabobalis:
I also broke the gallery build, which needs fixing before merge.

.github/workflows/ci.yml
Cadair, line 1: Some of the changes to this file are going to conflict with the package template. @CyclingNinja We could bump the default python in the template at this point, 3.12 is stable enough.

nabobalis: Is it just the default python version? What else might conflict?

Cadair: the default version yes, and the test/tests thing?

pytest.ini
```diff
@@ -4,17 +4,17 @@ testpaths =
...
     docs/_build
     docs/generated
-    *.egg-info
     examples
     sunpy/_dev
-    .history
-    tools
     sunpy/extern
-    benchmarks
+    tools
```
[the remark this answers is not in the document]
nabobalis: Say we add `sunkit_image/extern` on the sunkit_image, that has to go into the template as well?

.github/workflows/ci.yml
nabobalis: Well I saw docs and thought its time for test to be tests. I can undo it here.

pytest.ini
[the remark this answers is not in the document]
Cadair: extern is in the list in the template for good measure even if it doesn't exist. https://github.com/sunpy/package-template/blob/main/%7B%7B%20cookiecutter.package_name%20%7D%7D/pytest.ini#L15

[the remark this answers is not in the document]
nabobalis: Ok, I will undo the sorting here.

.github/workflows/ci.yml
nabobalis: Undone.

Issue #7560, opened 2024-04-08
sunpy import error in Jupyter Notebook


shah-astro:
### Describe the bug

Running the **"import sunpy"** in a cell of Jupyter Notebook results the following error:
```

---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
Cell In[1], line 1
----> 1 import sunpy

File ~\AppData\Local\anaconda3\envs\nn-gpu\lib\site-packages\sunpy\__init__.py:11
      1 """
      2 ``sunpy``
      3 =========
   (...)
      8 * Documentation: https://docs.sunpy.org/en/stable/
      9 """
     10 from sunpy.tests.self_test import self_test
---> 11 from sunpy.util import system_info
     12 from sunpy.util.config import load_config, print_config
     13 from sunpy.util.logger import _init_log

File ~\AppData\Local\anaconda3\envs\nn-gpu\lib\site-packages\sunpy\util\__init__.py:1
----> 1 from sunpy.util.decorators import *
      2 from sunpy.util.exceptions import *
      3 from sunpy.util.metadata import *

File ~\AppData\Local\anaconda3\envs\nn-gpu\lib\site-packages\sunpy\util\decorators.py:13
     10 from functools import wraps
     12 import astropy.units as u
---> 13 from astropy.nddata import NDData
     15 from sunpy.util.exceptions import SunpyDeprecationWarning, SunpyPendingDeprecationWarning, warn_deprecated
     17 __all__ = ['deprecated']

File ~\AppData\Local\anaconda3\envs\nn-gpu\lib\site-packages\astropy\nddata\__init__.py:14
     11 from astropy import config as _config
     13 from .bitmask import *
---> 14 from .blocks import *
     15 from .ccddata import *
     16 from .compat import *

File ~\AppData\Local\anaconda3\envs\nn-gpu\lib\site-packages\astropy\nddata\blocks.py:8
      2 """
      3 This module includes helper functions for array operations.
      4 """
      6 import numpy as np
----> 8 from .decorators import support_nddata
     10 __all__ = ["reshape_as_blocks", "block_reduce", "block_replicate"]
     13 def _process_block_inputs(data, block_size):

File ~\AppData\Local\anaconda3\envs\nn-gpu\lib\site-packages\astropy\nddata\decorators.py:12
      8 from itertools import islice
     10 from astropy.utils.exceptions import AstropyUserWarning
---> 12 from .nddata import NDData
     14 __all__ = ["support_nddata"]
[...]

PR #7561, opened 2024-04-08
Clean up util imports

Cadair:
This will address #7560 but isn't really a "fix".

We need to be very careful with what we import at the top level inside `util` as we import it from the top `sunpy` namespace, and large import paths will slow that down and also occasionally cause weird issues (apparently).

This is a quick fix, but I am unsure what we should do about this to make sure we keep the import hygiene in util?

### DIFF
--- a/sunpy/util/decorators.py
+++ b/sunpy/util/decorators.py
@@ -9,9 +9,6 @@ import functools
 from inspect import Parameter, signature
 from functools import wraps
 
-import astropy.units as u
-from astropy.nddata import NDData
-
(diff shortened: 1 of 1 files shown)


alasdairwilson (commented):
I am fine with this but this decorator is literally only used inside mapbase (indeed it is spec cifically designed for map arithmetic), why can't it just be put there? Is there something I am missing.

Cadair:
I mean, that seems sensible yeah.

nabobalis:
We do need to deal with util.

nabobalis (approved):

nabobalis (approved):

PR #7574, opened 2024-04-16
Update dev figure hashes

nabobalis:
Need to see if this breaks something.


sunpy/map/tests/test_reproject_to.py
```diff
@@ -51,6 +51,7 @@ def test_reproject_to_hgs(aia171_test_map, hgs_header):
     aia171_test_map.reproject_to(hgs_header).plot()
 
 
+@figure_test
```
nabobalis, line 54: I get that these are not standard figure tests but I still think they should be marked as such.

sunpy/image/tests/test_transform.py
```diff
@@ -296,8 +296,8 @@ def test_clipping(rot30):
     return fig
 
 
-@pytest.mark.filterwarnings("ignore:.*bug in the implementation of scikit-image")
 @figure_test
+@pytest.mark.filterwarnings("ignore:.*bug in the implementation of scikit-image")
```
ayshih, line 300: Revert this because it makes no difference

sunpy/map/tests/test_mapbase.py
```diff
@@ -1654,8 +1654,8 @@ def test_rotation_rect_pixelated_data(aia171_test_map):
     rect_rot_map.peek()
 
 
-@pytest.mark.parametrize('method', _rotation_registry.keys())
 @figure_test
+@pytest.mark.parametrize('method', _rotation_registry.keys())
```
ayshih, line 1658: Revert this because it makes no difference

ayshih (changes requested):
I'm certain that swapping the order of the decorators makes no difference for the issue that was encountered in CI, so I'd like those swaps to be reverted to keep the Git history more clean.

sunpy/map/tests/test_mapbase.py
nabobalis, line 1658:
```suggestion
@pytest.mark.parametrize('method', _rotation_registry.keys())
@figure_test
```

sunpy/image/tests/test_transform.py
nabobalis, line 300:
```suggestion
@pytest.mark.filterwarnings("ignore:.*bug in the implementation of scikit-image")
@figure_test
```

ayshih (approved):

PR #7583, opened 2024-04-23
[map branch] `NDCube.__init__` accesses `.wcs` and chaos unfolds

Cadair:


Cadair (commented):
:shipit: apart from that mapbase test.

PR #7585, opened 2024-04-23
Allow DN as a unit on `GenericMap`

### LINKED ISSUE #6566: Use of `u.DN` unit in place of `u.count`

DN is a commonly used unit in solar physics and often shows up in FITS headers. Currently, we replace DN with "count" in `GenericMap`,

https://github.com/sunpy/sunpy/blob/dbcf1b59d9a2b35d1fcd8ca2219a25dc64dafc96/sunpy/map/mapbase.py#L722-L726

This is primarily because DN is not a unit that is recognized by the FITS standard but also because until recently, the closest representation of DN in `astropy.units` is `u.count`. However, as of astropy 4.3, there is now a `u.DN`: https://docs.astropy.org/en/v4.3.1/units/index.html#module-astropy.units.astrophys. This raises the question of whether we should switch to using `u.DN` instead of `u.count` once we pin to >= 4.3.

The main disadvantage to this would be that any map with units of `u.DN` would not be FITS serializable. 

See also the discussion in the chat on this topic: https://matrix.to/#/!JYqfIVJjWANcHnfktY:cadair.com/$166743304126958mwNUw:matrix.org?via=cadair.com&via=matrix.org&via=openastronomy.org

wtbarnes:
Fixes #6566


sunpy/map/tests/test_mapbase.py
```diff
@@ -1701,10 +1699,10 @@ def test_map_arithmetic_addition_subtraction(aia171_test_map, value):
...
 @pytest.mark.parametrize('value', [
-    10 * u.ct,
-    u.Quantity([10], u.ct),
-    u.Quantity(np.random.rand(128), u.ct),
-    u.Quantity(np.random.rand(128, 128), u.ct),
+    10 * u.s,
+    u.Quantity([10], u.s),
+    u.Quantity(np.random.rand(128), u.s),
+    u.Quantity(np.random.rand(128, 128), u.s),
     10.0,
```
DanRyanIrish, line 1706: Why have `ct` changed to `s`?  Rather than `DN`?

wtbarnes: Fair question. Because it seemed silly to be multiplying (or dividing) `DN` by something in `ct`

nabobalis:
The only tests I could not fix were the asdf ones.

sunpy/map/sources/tests/test_iris_source.py
```diff
@@ -47,7 +47,8 @@ def test_level_number(irismap):
 
 def test_units(irismap):
     """Tests the unit property of the SJIMap"""
-    assert irismap.unit == u.ct
+    with pytest.warns(SunpyMetadataWarning, match="Could not parse unit string"):
+        assert irismap.unit == u.Unit("DN")
```
wtbarnes, line 51: Wait. Why is warning being thrown? What are the units here?

sunpy/map/mapbase.py
```diff
@@ -726,19 +726,22 @@ def max(self, *args, **kwargs):
...
         if isinstance(unit, u.UnrecognizedUnit):
-            warn_metadata(f'Could not parse unit string "{unit_str}" as a valid FITS unit.\n'
-                          f'See {_META_FIX_URL} for how to fix metadata before loading it '
-                          'with sunpy.map.Map.\n'
-                          'See https://fits.gsfc.nasa.gov/fits_standard.html for '
-                          'the FITS unit standards.')
-            unit = None
+            # NOTE: Special case DN here as it is not part of the FITS standard, but
+            # is widely used and is also a recognized astropy unit
+            if 'DN' in str(unit):
```
wtbarnes, line 736: This still seems like a *very* unreliable/unsatisfying way to check for this, but i cannot come up with any other way. Once you parse the unit using `format='fits'` as we do above, the whole thing just becomes a big blob of `UnrecognizedUnit` so it becomes impossible to decompose each piece into `u.Unit` objects.

wtbarnes:
The figure test fails are real and are due to the unit label changing to "DN"

sunpy/map/mapbase.py
```diff
@@ -726,19 +726,24 @@ def max(self, *args, **kwargs):
...
         if isinstance(unit, u.UnrecognizedUnit):
-            warn_metadata(f'Could not parse unit string "{unit_str}" as a valid FITS unit.\n'
-                          f'See {_META_FIX_URL} for how to fix metadata before loading it '
-                          'with sunpy.map.Map.\n'
-                          'See https://fits.gsfc.nasa.gov/fits_standard.html for '
-                          'the FITS unit standards.')
-            unit = None
+            # NOTE: Special case DN here as it is not part of the FITS standard, but
+            # is widely used and is also a recognized astropy unit
+            unit_str_components = unit.to_string().replace('/', ' ').replace('(', '').replace(')', '')
```
nabobalis, line 736: What on earth caused you to need this line? This is great and I am here for it.

wtbarnes: I guess I removed the comment explaining this. Because anything with DN is not FITS-compliant, the whole unit becomes an `UnrecognizedUnit` and you cannot decompose it into individual `Unit` objects. Thus, you have to parse it as a string. The `replace` methods remove non unit characters that could be present in the string. 

The goal here is to breakup the unit into components to check if one of them is DN. Just checking for "DN" in the unit string is ambiguous because it's possible some unit could include a contiguous "dn" in it somewhere. 

If there is a better way to do this, I'm all for it. I don't like this approach.

nabobalis: > I guess I removed the comment explaining this. Because anything with DN is not FITS-compliant, the whole unit becomes an `UnrecognizedUnit` and you cannot decompose it into individual `Unit` objects. Thus, you have to parse it as a string. The `replace` methods remove non unit characters that could be present in the string.
> 
> The goal here is to breakup the unit into components to check if one of them is DN. Just checking for "DN" in the unit string is ambiguous because it's possible some unit could include a contiguous "dn" in it somewhere.
> 

Ah! 

> I don't like this approach.

I do.

What kind of strings do we have to handle?
If I can see some examples, we could hopefully come up with another method.

wtbarnes: See the updated `_parse_fits_unit` tests for some examples.

Cadair: Rather than this abomination why not just do this:
```python
unit = u.Unit(unit_str)
if u.DN not in base_unit.bases:
    warn_metadata(...)
    unit = None
```

wtbarnes: Does that work when you have an `UnrecognizedUnit`? I did not try that but yes that would certainly be preferable (though less fun).

nabobalis: Seems to work, at least the tests pass.

nabobalis (approved):

wtbarnes:
The test fail is real though I don't understand why that isn't warning.

nabobalis:
> The test fail is real though I don't understand why that isn't warning.

I fixed that one due to my mistyping, the online doc test failure is real.

It wants to convert DN and cts. I can tweak the example.

Cadair:
So our solution to not knowing how to fix the asdf tests was to just... not?

nabobalis:
Yes.

PR #7586, opened 2024-04-23
[map branch] Remove `reproject_to` in favor of using the method already defined on `NDCube`

wtbarnes:
This removes the `reproject_to` method on `GenericMap` in favor of using the one already defined on `NDCube`.

TODOs

- [ ]  changelog
- [ ] add a test that metadata is now preserved

### DIFF
--- a/changelog/7586.removal.rst
+++ b/changelog/7586.removal.rst
@@ -1,0 +1,1 @@
+Remove `sunpy.map.GenericMap.reproject_to` in favor of `ndcube.NDCube.reproject_to`.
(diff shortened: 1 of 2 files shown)


ayshih:
Bye bye #7370

This also means that metadata will be handled differently, right?  (xref #5627)

wtbarnes:
> Bye bye #7370

Oops, that's my fault. looks like we'll at least need to override this method then with that bit of logic followed by a `super().reproject_to`.

> This also means that metadata will be handled differently, right? (xref #5627)

Yep, that's correct. I think because NDCube separates WCS and non-WCS metadata, this should solve this problem as we can just propagate all non-WCS metadata to the reprojected map.

Cadair:
> Bye bye, #7370, we hardly knew ye

Why did we do that again :stuck_out_tongue:

PR #7588, opened 2024-04-24
[map branch] `.wcs` setter

Cadair:


sunpy/map/mapbase.py
```diff
@@ -727,6 +728,35 @@ def wcs(self):
...
+        """
+        # Unwrap any wrapper classes to FITS WCS
+        unwrapped, _ = unwrap_wcs_to_fitswcs(wcs)
+        # Convert to a header
+        new_header = unwrapped.to_header()
+        old_wcs_header = self.wcs.to_header()
+        # Reduce the new header to just the keys which differ from the current WCS
+        # We do this to figure out what's been changed post wcslib doing any
+        # conversion (such as arcsec -> deg)
+        changed_header = dict(set(new_header.items()).difference(old_wcs_header.items()))
```
DanRyanIrish, line 745: Does this return a dict of keys with no values?

sunpy/map/mapbase.py
```diff
@@ -727,6 +728,35 @@ def wcs(self):
...
+        # If any of the keys in spatial units are modified wcslib will have
+        # almost certainly changed their units to deg if they were arcsec, so we
+        # have to explicitly check this and convert them back to the original
+        # header units to not confuse people.
+        naughty_key_prefixes = {"CDELT", "CRVAL", "CD"}
+        for prefix in naughty_key_prefixes:
+            if any(k.startswith(prefix) for k in changed_header.keys()):
+                if new_header["CUNIT1"] != self.meta["CUNIT1"]:
```
DanRyanIrish, line 753: Why is `"CUNIT1"` in particular being checked?

sunpy/map/mapbase.py
```diff
@@ -727,6 +728,35 @@ def wcs(self):
...
+                    raise NotImplementedError(
```
DanRyanIrish, line 754: Love this!  But what would need to be done if we hit this error?

sunpy/map/mapbase.py
Cadair: only crazy people have maps with a different unit along both celestial dimensions (also this is just temporary).

sunpy/map/mapbase.py
Cadair: We will probably run into this when we add support for ResampledWCS, we will need to convert the units of the keys in the new header to the units of the original map.

DanRyanIrish (approved):

PR #7594, opened 2024-04-24
Enable use of  SunPyBaseCoordinateFrame outside sunpy

### LINKED ISSUE #7593: Support the use of SunPyBaseCoordinateFrame outside of core library
### Describe the feature

* It would be useful to be able to use `SunPyBaseCoordinateFrame` outside of sunpy but currently doing so is alkward as the `frame_to_wcs` and `wcs_to_frame` mappings are hardcode to the current frames / wcs' and raise an error you you try to do so
* The name clearly implies this is SunPy specific but there a lot of boiler plate code that would have to be copied to provide the same thing.

At the moment to make this work you have to make sure what ever frames are added are added before the sunpy function e.g.

`astropy.wcs.utils.FRAME_WCS_MAPPINGS.insert(1, [my_wcs_to_frame])`

### Proposed solution

Don't raise an error the if the frame is a subclass of `SunPyBaseCoordinateFrame` but not one of the frame implemented in sunpy return `None` 

Specifically L198 raise an error as xcoord and xcoord are undefined
https://github.com/sunpy/sunpy/blob/2281a2198997e8671efd48dcb531b07a98b86ddf/sunpy/coordinates/wcs_utils.py#L173-L198

A simple solution would be to an a final else clause to this if Elias
```python
        elif isinstance(frame, HeliographicStonyhurst):
            xcoord = 'HGLN' + '-' + projection
            ycoord = 'HGLT' + '-' + projection
            wcs.wcs.cunit = ['deg', 'deg']
        else
            return None

```

samaloney:
Closes #7593 by enabling use of `SunPyBaseCoordinateFrame` outside of sunpy by returning None if not a sunpy Frame instead of raising an error.

### DIFF
--- a/sunpy/coordinates/wcs_utils.py
+++ b/sunpy/coordinates/wcs_utils.py
@@ -191,6 +191,9 @@ def solar_frame_to_wcs_mapping(frame, projection='TAN'):
             xcoord = 'HGLN' + '-' + projection
             ycoord = 'HGLT' + '-' + projection
             wcs.wcs.cunit = ['deg', 'deg']
+        else:
+            # A subclass not supported by the core library
+            return None
 
     else:
         return None
--- a/changelog/7594.bugfix.rst
+++ b/changelog/7594.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug that interfered with :func:`astropy.wcs.utils.celestial_frame_to_wcs` when working with a custom subclass of :class:`~sunpy.coordinates.frames.SunPyBaseCoordinateFrame`.
--- a/sunpy/coordinates/tests/test_wcs_utils.py
+++ b/sunpy/coordinates/tests/test_wcs_utils.py
@@ -15,6 +15,7 @@ from sunpy.coordinates.frames import (
     HeliographicCarrington,
     HeliographicStonyhurst,
     Helioprojective,
+    SunPyBaseCoordinateFrame,
 )
 from sunpy.coordinates.wcs_utils import (
     _set_wcs_aux_obs_coord,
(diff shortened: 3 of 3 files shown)


sunpy/coordinates/wcs_utils.py
```diff
@@ -191,6 +191,9 @@ def solar_frame_to_wcs_mapping(frame, projection='TAN'):
             xcoord = 'HGLN' + '-' + projection
             ycoord = 'HGLT' + '-' + projection
             wcs.wcs.cunit = ['deg', 'deg']
+        else:
+            # A subclass but not a sunpy frame
```
ayshih, line 195:
```suggestion
            # A subclass not supported by the core library
```

ayshih (approved):

changelog/7594.feature.rst
```diff
@@ -0,0 +1 @@
+Enable support of subclassing :class:`~sunpy.coordinates.frames.SunPyBaseCoordinateFrame` outside the core library.
```
Cadair, line 1: I think this should mention the fact it's `wcs_to_coordinate_frame` or whatever the astropy api is.

samaloney: Yea it's the `celestial_frame_to_wcs` and then `wcs_to_celestial_frame` the other way. So what something like

```
Enable support of subclassing :class:`~sunpy.coordinates.frames.SunPyBaseCoordinateFrame` outside the core library utilising :func:`astropy.wcs.utils.wcs_to_celestial_frame`.

Comment (Cadair) [reply changelog/7594.feature.rst:1]: I would reorder that sentence a little and say "ignore unknown subclasses of SunPyBaseCoordinateFrame by default in wcs_to_celestial_frame"?

**Comment by ayshih** [comment]
Hmm, I'm inclined to categorize this PR as a bug fix (as opposed to a feature) and backport it

**Review by ayshih — CHANGES_REQUESTED** [review]
Okay, change the changelog entry to be a a bugfix, a la
```
Fixed a bug that interfered with :func:`astropy.wcs.utils.celestial_frame_to_wcs` when working with a custom subclass of :class:`~sunpy.coordinates.frames.SunPyBaseCoordinateFrame`.
```

Cadair (approved):

ayshih (approved):

PR #7601, opened 2024-04-25
fix failing windows tests

exitflynn:
The Scraper branch's windows tests had been failing since we were using `pathlib.Path` in the tests to send the directory path, which comes out with backslashes in windows fashion and would let us down when comparing with a forward-slash `pattern`. The way it's handled in main is we simply just don't use `Path()` to get OS-specific filepaths, and just linux style forward-slash strings. I've reverted it to match with main, though a discussion could be had if we should include `pathlib.Path` generated paths there 🤔 

Another error was because in one of the functions, scraper internally replaces \\ with / in the pattern when in windows. I had previously replaced one of the two pattern variables and had missed the other one


nabobalis (approved):

nabobalis:
Thanks @exitflynn

PR #7622, opened 2024-05-12
Fix devdeps fails

nabobalis:
This just skips the failing ones on the figure tests.

This will need to be un-skipped when opencv does a release.

### DIFF
--- a/sunpy/image/tests/test_transform.py
+++ b/sunpy/image/tests/test_transform.py
@@ -265,6 +265,7 @@ def test_reproducible_matrix_multiplication():
 
 
 @figure_test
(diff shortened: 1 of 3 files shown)


wtbarnes:
What is causing this failure? When can we unskip this?

nabobalis:
> What is causing this failure? When can we unskip this?

Two things:

1. The dask repr causes a runtime warning about the divide that np.errorstate didn't seem to work for me at all.
2. The second one are the figure tests which need to wait for a patched numpy 2.0 compatible release of opencv. They will always fail until that happens.

wtbarnes (approved):

PR #7626, opened 2024-05-14
Add try except around client queries

### LINKED ISSUE #7625: Make Fido.search robust against errors from a single client/s
### Describe the feature

* Currently if one client has and error the entire `Fido.search` crashes which is very annoying especially when it's not the intended client that has the problem.

For example at the time of making this issue the VSO seems to be down but this causes searches for NOAA SRS data to crash while a direct search of the SRS client works.

```

q = Fido.search(a.Time('2022-12-01', '2022-12-31'), a.Instrument.soon)
ConnectionRefusedError                    Traceback (most recent call last)
```

```
from sunpy.net.dataretriever.sources.noaa import SRSClient
srs = SRSClient()
srs.search(a.Time('2022-12-01', '2022-12-31'))
srs
<sunpy.net.dataretriever.client.QueryResponse object at 0x159be8310>
       Start Time               End Time        Instrument Physobs Source Provider
----------------------- ----------------------- ---------- ------- ------ --------
2022-12-01 00:00:00.000 2022-12-01 23:59:59.999       SOON     SRS   SWPC     NOAA
2022-12-02 00:00:00.000 2022-12-02 23:59:59.999       SOON     SRS   SWPC     NOAA
2022-12-03 00:00:00.000 2022-12-03 23:59:59.999       SOON     SRS   SWPC     NOAA
2022-12-04 00:00:00.000 2022-12-04 23:59:59.999 
```

### Proposed solution

If a query can be handled by more then one client and at least one client doesn't error return a result object.

samaloney:
## PR Description
Make `Fido.search` more robust by wrapping the `client.search` calls in a try except. Current code works but introduces silent fails I will add some logging. Need a way to surface the errors to the user in the search results maybe add a `ClientError` that could be include in the result or add and `.error` similarly to the `Fido.fetch`

- [ ] Add `.errors` to the `UnifiedResponse`
- [ ] Reduce log clutter

Closes #7625

### DIFF
--- a/sunpy/net/fido_factory.py
+++ b/sunpy/net/fido_factory.py
@@ -477,9 +477,11 @@ class UnifiedDownloaderFactory(BasicRegistrationFactory):
         candidate_widget_types = self._check_registered_widgets(*query)
         results = []
         for client in candidate_widget_types:
-            tmpclient = client()
-            results.append(tmpclient.search(*query))
-
+            try:
+                tmpclient = client()
+                results.append(tmpclient.search(*query))
+            except Exception:
+                pass
         # This method is called by `search` and the results are fed into a
         # UnifiedResponse object.
         return results
--- a/sunpy/net/fido_factory.py
+++ b/sunpy/net/fido_factory.py
@@ -18,7 +18,7 @@ from packaging.version import Version
 
 from astropy.table import Table
 
-from sunpy import config
+from sunpy import config, log
 from sunpy.net import attr, vso
 from sunpy.net.base_client import BaseClient, QueryResponseColumn, QueryResponseRow, QueryResponseTable
 from sunpy.util.datatype_factory_base import BasicRegistrationFactory, NoMatchError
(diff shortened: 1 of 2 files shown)


Cadair:
I am + on adding a `.error[s]` to `UnifiedResponse` as well. I worry the full `log.exception` might be too much if we do that though, people are likely to think it's all broken even if it isn't.

samaloney:
Ok I'll have a look at adding a `.errors` to the `UnifiedResponse`. Do the logs even show up with the defaults? I could replace the current log with a small message at info/warning and keep the track trace at debug?

Cadair:
I think everything info and above shows up by default.

samaloney:
Ok I've added a comment to the issue to discuss how the API could should look before I go to far with an implementation - https://github.com/sunpy/sunpy/issues/7625#issuecomment-2114439381

nabobalis:
I will close this until the comments on the issue are resolved.

PR #7463, opened 2024-02-29
ADAPT Client

GillySpace27:
## PR Description

Adds functionality to download ADAPT maps using FIDO.
This addresses the following issue: https://github.com/sunpy/sunpy/issues/7152

I believe that the core of this code works but I cant get a dev environment running to test things. I believe I'll have to do more work putting the right parts in the right places for sunpy, but I wanted to get this started by uploading what I have.

PR #7640, opened 2024-05-21
Add Adapt Map

nabobalis:
This comes from PFSSPY

This pairs nicely with https://github.com/sunpy/sunpy/pull/7463 

TODO: 
- [x] Merge in above PR, rebase this and update the map source link.


sunpy/map/sources/adapt.py
```diff
@@ -0,0 +1,37 @@
...
+        ctype1, ctype2 = self.meta['ctype1'], self.meta['ctype2']
+        if ctype1 == 'Long':
+            ctype1 = 'CRLN-CAR'
+        if ctype2 == 'Lat':
+            ctype2 = 'CRLT-CAR'
+        return SpatialPair(ctype1, ctype2)
+
+    @property
+    def date(self):
+        return Time(self.meta.get('date-obs',self.meta.get('maptime')))
```
wtbarnes, line 31: This should probably fall back to `super().date` if neither of those keys can be found.

nabobalis, line 31:
```suggestion
        return Time(self.meta.get('date-obs', self.meta.get('maptime', super().date)))
```

nabobalis: I can't python, please hold

nabobalis: Python better?

sunpy/data/test/adapt.header
```diff
@@ -0,0 +1,75 @@
+SIMPLE  =                    T / file does conform to FITS standard
+BITPIX  =                  -32 / number of bits per data pixel
+NAXIS   =                    3 / number of data axes
+NAXIS1  =                  360 / length of data axis 1
+NAXIS2  =                  180 / length of data axis 2
+NAXIS3  =                   12 / length of data axis 3
```
wtbarnes, line 6: So in our test data, this will result in a 2D map. In the actual ADAPT files, is there a 3rd dimension? If so, I think the Map will complain about this when it drops the third dimension. There is also the question of how we would support taking out dimensions that are not just the 0th dimension.

wtbarnes: This is kind of odd because this is a dimension in the WCS that has no meaning when it comes to the WCS, i.e. you cannot define a `CTYPE` or a `CDELT` for the "ensemble axis"

sunpy/map/sources/adapt.py
```diff
@@ -0,0 +1,37 @@
...
+
+from astropy.time import Time
+
+from sunpy.map.mapbase import GenericMap, SpatialPair
+
+__all__ = ['ADAPTMap']
+
+class ADAPTMap(GenericMap):
+    """
+    ADAPT (Air Force Data Assimilative Photospheric Flux Transport) flux transport model evolves an ensemble of realizations.
```
wtbarnes, line 13:
```suggestion
    ADAPT magnetic flux map
     
    The ADAPT (Air Force Data Assimilative Photospheric Flux Transport) flux transport model evolves an ensemble of realizations, using flux transport processes during periods for which there are no observations, and updates the ensemble using the ensemble least-squares (EnLS) data assimilation method to account for model and observational uncertainties. These data products are Carrington maps with an additional third dimension corresponding to the ensemble of models used to produce the ADAPT map.
```

sunpy/map/sources/tests/test_adapt.py
wtbarnes, line 1: This needs a test of the `date` and `coordinate_system` properties as well since we are special casing those for this map.

sunpy/map/sources/adapt.py
wtbarnes: I lifted most of this text from the NSO ADAPT webpage. It would be useful if someone who knows more about these data products could look over these as well.

wtbarnes (changes requested):

sunpy/data/test/adapt.header
nabobalis: Yes I agree, I need to get a real file and see what it looks like.

But it lacks the complete set of FITS keywords for the third axis.

sunpy/map/sources/tests/test_adapt.py
nabobalis: Added

nabobalis:
This is very much broken with a real FITS file.

sunpy/data/test/adapt.header
wtbarnes: Yeah it's really bizarre...

docs/whatsnew/6.0.rst
```diff
@@ -92,5 +92,12 @@ Pass them as keyword arguments (e.g., ``..., title=True, ...``) instead.
 Support for ADvanced Adaptive Prediction Technique (ADAPT)
 ==========================================================
 
-A new map source has been added (``sunpy.map.sources.ADAPTMap``) to support the ADAPT data files.
+A new map source has been added (`sunpy.map.sources.ADAPTMap`) to support the ADAPT data files.
 In addition, a new client (`sunpy.net.dataretriever.ADAPTClient`) has been added to search and download ADAPT files.
+
+Please note that there are several data arrays within the FITS file and map will try to read them all and fail, so you have to specify the Header Data Pair you want to read.
+In this case, it will always be the first one. You can do this by passing the ``hdus`` argument to the `sunpy.map.Map` call.
```
wtbarnes, line 99:
```suggestion
Please note that there are several data arrays within the ADAPT FITS files and `~sunpy.map.Map` will, by default, try to read them all and fail.
In these cases, you must specify the header-data Pair you want to read.
For these data, it will always be the first one.
You can specify this by passing the ``hdus`` keyword argument to `~sunpy.map.Map`,
```

sunpy/map/sources/adapt.py
```diff
@@ -0,0 +1,51 @@
...
+__all__ = ['ADAPTMap']
+
+class ADAPTMap(GenericMap):
+    """
+    ADAPT magnetic flux map
+
+    The ADAPT (Air Force Data Assimilative Photospheric Flux Transport) flux transport model evolves an ensemble of realizations, using flux transport processes during periods for which there are no observations, and updates the ensemble using the ensemble least-squares (EnLS) data assimilation method to account for model and observational uncertainties.
+    These data products are Carrington maps with an additional third dimension corresponding to the ensemble of models used to produce the ADAPT map.
+
+    The third dimension is ignored and only the first element of the third dimension is used to create the  map.
```
wtbarnes, line 18:
```suggestion
    Here, the third dimension is ignored and only the first element of the third dimension is used to create the  resulting map.
```

sunpy/map/sources/adapt.py
```diff
@@ -0,0 +1,51 @@
...
+
+    .. warning::
+
+        Please note that there are several data arrays within the FITS file and map will try to read them all and fail, so you have to specify the Header Data Pair you want to read.
+        In this case, it will always be the first one.
+        You can do this by passing the ``hdus`` argument to the `sunpy.map.Map` call.
```
wtbarnes, line 24:
```suggestion
        Please note that there are several data arrays within the ADAPT FITS files and `~sunpy.map.Map` will, by default, try to read them all and fail.
        In these cases, you must specify the header-data Pair you want to read.
        For these data, it will always be the first one.
        You can specify this by passing the ``hdus`` keyword argument to `~sunpy.map.Map`
```

wtbarnes (approved):
Minor rewording suggestions, but otherwise looks good to me.

PR #7660, opened 2024-06-04
Handle FITS io differently for time series vs Map

nabobalis:
So for a long time, we have been just loading all the hdu's of a FITS file even if we don't support them or can even load them into a map or time series.

This PR tackles that by explicitly opting for Images only for Map and adding a hack workaround for time series. 

The goal is which we can load ADAPT FITS files which are strange.
Our other choice is to just say that to load a adapt fits, you have to manually pass in the primary header and data pair yourself.

### DIFF
--- a/sunpy/io/_fits.py
+++ b/sunpy/io/_fits.py
@@ -38,8 +38,9 @@ from sunpy.util.io import HDPair
 
 __all__ = ['header_to_fits', 'read', 'get_header', 'write', 'extract_waveunit', 'format_comments_and_history']
 
+SUPPORTED_FITS_IMAGES_TYPES = (fits.CompImageHDU, fits.PrimaryHDU, fits.ImageHDU)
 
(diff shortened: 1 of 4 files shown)


samaloney:
I think as we discussed on the community call you can pass `hdus=1` or even `hdus=[1,2]` but I don't know what that would really do.

https://github.com/sunpy/sunpy/blob/767cd79bc376e419bc5dae257a5d95a126977f5c/sunpy/io/_fits.py#L42

nabobalis:
> I think as we discussed on the community call you can pass `hdus=1` or even `hdus=[1,2]` but I don't know what that would really do.
> 
> https://github.com/sunpy/sunpy/blob/767cd79bc376e419bc5dae257a5d95a126977f5c/sunpy/io/_fits.py#L42

Thanks, I had no idea.

PR #7227, opened 2023-10-09
merge baseurl and pattern for scraper clients (#7077)

nabobalis:
This is the main PR to update the scraper, we want this in for 6.1

PR #7077, opened 2023-06-24
merge baseurl and pattern for scraper clients

exitflynn:
as things are right now, while defining scraper-based dataretriever clients, the `baseurl` which we send to scraper, is a pattern in the strftime format and we also require a second `pattern` string which is the same thing but formatted in the `parse` format.
What this PR does is change this to requiring only one full parse-formatted `pattern` in the scraper and generating the strftime formatted `timepattern` from it.

Having the new parse-format `pattern` around will also help considerably help simplifying many existing functions like `_extractDateURL()` which are using regex for parsing and help with #4888.

closes #7073 

This PR is part of my Google Summer of Code 2023 project.

PR #7664, opened 2024-06-06
Add backwards compatibility for the scraper

exitflynn:
Aims to add backwards compatibility for the base scraper post-rewrite from #7077 so that we can go through a deprecation period for the breaking changes.

Addressing @dstansby's [suggestions](https://github.com/sunpy/sunpy/pull/7227#pullrequestreview-1682939171) on how we can approach this, so far this PR pretty much addresses all the points in terms of the Scraper design, namely the PR:

  - Makes `pattern` optional (ie. , pattern=None)
  - Adds a new keyword-only argument, `format`, to take the new format
  - Raises an error if both pattern and format are provided
  - Uses old behaviour, and raises a deprecation warning saying to use new format=... instead if only `pattern` passed
  - Uses new behaviour if only `format` passed.

Outside of changing `sunpy.net.scraper` I've done the following:
- Added back the tests for the original scraper with the ones I wrote (distinguished by a `new_format` suffix) and they all should be passing now.
- During the rewrite, I had previously moved out `check_timerange` from `scraper` to `scraper_utils` but I moved it back in so long as the old scraper is still around since I felt it'd get a bit too complex otherwise (for example, it'd require internal scraper functions that were removed in the rewrite, etc).
- Updated hopefully all the relevant docstring examples and the base dataretriever clients' code. I am _not_ sure of all the tests passing, i'm curious of the CI's verdict on that rn :P
- I've also updated the `extending_fido.rst` doc about `pattern` vs `format` and have tried to improve some of the documentation I wrote previously explaining the Scraper algorithm.
- *also renamed `timepattern` to ~~`dt_pattern`~~ `datetime_pattern` since it's also appearing in docs rn, felt it more fitting.

The really long file for the scraper tests right now make me doubt a little if maybe I should've skimmed over some tests when adding backwards compatibility, I'd love to know if I've been on the right path with this so far :cowboy_hat_face:

TODO:
- [x] We need to add a test that creates an old style scraper client (basically use one of the old clients) and make sure that it raises the warning correctly and still works. (Added back nearly all the old teststhat expect and catch the warning  and renamed new ones to have a `new_format` suffix)
- [x] Write a migration guide in the developer docs and link to it in the message.

### DIFF
--- a/docs/topic_guide/extending_fido.rst
+++ b/docs/topic_guide/extending_fido.rst
@@ -17,6 +17,21 @@ The main place this is done is when constructing a `~.UnifiedResponse` object, w
 
 .. _sunpy-topic-guide-new-source-for-fido-add-new-scraper-client:
 
+A migration guide for working with / adding new scraper clients:
+===============================================================
+The main change in 5.0 is the removal of the old regex-based pattern system in favor of a new parse-based pattern system.
+This change is to make the pattern system more flexible and easier to use. The new pattern system is based on the `parse <https://github.com/r1chardj0n3s/parse/>`__ format.
+Previously, when defining a new FIDO client or working with the Scraper, a regex-based ``baseurl`` and a parse-styled ``pattern`` attributes were required.
+Now only a single parse-based ``pattern`` is required when writing Fido clients.
+During the deprecation period for the old ``pattern``, the new one is to be passed as a ``format`` attribute to the Scraper.
+Instructions for writing the new pattern are:
+1. Much like the baseurl, the single pattern would cover the entire URL from the protocol, domain to the filename.
+2. Instead of conveying time or numerical information in the URL in a datetime or regex format, it is now done in the `parse <https://github.com/r1chardj0n3s/parse/>`__ format for eg. a ``"%Y"`` is now to be replaced with ``{{year:4d}}``. For a full list of the new keywords corresponding to datetime format, the supported time keys (to be used within ``{{}}``!) are: 'year:4d', 'year:2d', 'month:2d'. 'month_name:l', 'month_name_abbr:l', 'day:2d', 'day_of_year:3d', 'hour:2d', 'minute:2d', 'second:2d', 'microsecond:6d', 'millisecond:3d' and 'week_number:2d'.
+3. The metadata attributes for extraction are written within double curly-braces ``{{}}`` for eg. ``{{ADAPTRealizations:3d}}``.
+4. Single curly-braces ``{}`` are used in case of regular placeholders for Python format strings, each respective value being passed as ``kwargs`` to the Scraper.
+A more detailed explanation of the new pattern system is referred to in the topic-guide below.
+
+
 An explanation of how "scraper" clients work
 ============================================
 
@@ -29,7 +44,7 @@ A new "scraper" client inherits from `~sunpy.net.dataretriever.client.GenericCli
   Each `tuple` contains the "attr" value and its description.
 * A class attribute ``pattern``; this is a string used to match all URLs supported by the client and extract necessary metadata from the matched URLs.
   The time and other metadata attributes for extraction are written within double curly-braces ``{{}}`` in a `parse <https://github.com/r1chardj0n3s/parse/>`__ format.
-  Regular placeholders for Python format strings can still be included via single curly braces ``{}`` with their respective parameters passed as the ``kwargs``.
+  Single curly braces ``{}`` are only used for regular placeholders for Python format strings, each respective value passed as ``kwargs`` to the Scraper.
   An example of how such a pattern looks like is given in the algorithm explanation below.
 
 .. warning::
@@ -125,7 +140,7 @@ This client scrapes all the URLs available under the base url ``http://lasp.colo
 `~sunpy.net.scraper.Scraper` is primarily focused on URL parsing based on time ranges, so the rest of the ``pattern`` specifies where in the URL the time information is located, using `parse <https://github.com/r1chardj0n3s/parse/>`__ notation.
 The ``pattern`` attribute is first filled in with the calculated time-based values, and then used to populate the results table from the URLs matched with the ``pattern``.
 It includes some of the time definitions, as well as names of attrs (in this case "Level").
-The supported time keys are: '{year:4d}', '{year:2d}', '{month:2d}'. '{month_name:l}', '{month_name_abbr:l}', '{day:2d}', '{day_of_year:3d}', '{hour:2d}', '{minute:2d}', '{second:2d}', '{microsecond:6d}', '{millisecond:3d}' and '{week_number:2d}'.
+The supported time keys (to be used within ``{{}}``!) are: 'year:4d', 'year:2d', 'month:2d'. 'month_name:l', 'month_name_abbr:l', 'day:2d', 'day_of_year:3d', 'hour:2d', 'minute:2d', 'second:2d', 'microsecond:6d', 'millisecond:3d' and 'week_number:2d'.
 
 The attrs returned in the ``register_values()`` method are used to match your client to a search, as well as adding their values to the attr.
 This means that after this client has been imported, running ``print(a.Provider)`` will show that the ``EVEClient`` has registered a provider value of ``LASP``.
--- a/docs/topic_guide/extending_fido.rst
+++ b/docs/topic_guide/extending_fido.rst
@@ -30,6 +30,8 @@ A new "scraper" client inherits from `~sunpy.net.dataretriever.client.GenericCli
 * A class attribute ``pattern``; this is a string used to match all URLs supported by the client and extract necessary metadata from the matched URLs.
   The time and other metadata attributes for extraction are written within double curly-braces ``{{}}`` in a `~sunpy.extern.parse.parse` format. Regular placeholders for Python format strings can still be included via single curly braces ``{}`` with their respective parameters passed as the ``kwargs``. An example of how such a pattern looks like is given in the algorithm explanation below.
 
+Note: The scraper has supported regex-based patterns for a long time, which is what it will still expect by default. However that is in the process of being replaced with parse-style patterns in the future versions of the package. To use the newer parse-style patterns, it's currently required to pass it as the ``format`` argument to the scraper.
+
 Each such client relies on the `~sunpy.net.scraper.Scraper` to be able to query for files using the :meth:`~sunpy.net.scraper.Scraper.filelist` method. The general algorithm to explain how the `~sunpy.net.scraper.Scraper` is able to do this is:
 
 A brief explanation of how the Scraper works is as follows:
@@ -48,7 +50,7 @@ For a more in-depth explanation on how this is accomplished internally, see the 
     >>> from sunpy.net import Scraper
     >>> pattern = ('http://proba2.oma.be/{instrument}/data/bsd/{{year:4d}}/{{month:2d}}/{{day:2d}}/'
     ...            '{instrument}_lv1_{{year:4d}}{{month:2d}}{{day:2d}}_{{hour:2d}}{{minute:2d}}{{second:2d}}.fits')
-    >>> s = Scraper(pattern, instrument='swap')
+    >>> s = Scraper(format=pattern, instrument='swap')
     >>> s.dt_pattern
     'http://proba2.oma.be/swap/data/bsd/%Y/%m/%d/swap_lv1_%Y%m%d_%H%M%S.fits'
 
--- a/docs/topic_guide/extending_fido.rst
+++ b/docs/topic_guide/extending_fido.rst
@@ -95,7 +95,7 @@ The files that satisfy these conditions are then added to the output.
     ['http://proba2.oma.be/swap/data/bsd/2015/01/01/swap_lv1_20150101_000857.fits',
     'http://proba2.oma.be/swap/data/bsd/2015/01/01/swap_lv1_20150101_001027.fits',
     '...',
-    'http://proba2.oma.be/swap/data/bsd/2015/01/01/swap_lv1_20150101_235947.fits']
+    'http://proba2.oma.be/swap/data/bsd/2015/01/02/swap_lv1_20150102_233313.fits']
 
 Writing a new "scraper" client
 ==============================
--- a/docs/topic_guide/extending_fido.rst
+++ b/docs/topic_guide/extending_fido.rst
@@ -32,9 +32,16 @@ A new "scraper" client inherits from `~sunpy.net.dataretriever.client.GenericCli
   Regular placeholders for Python format strings can still be included via single curly braces ``{}`` with their respective parameters passed as the ``kwargs``.
   An example of how such a pattern looks like is given in the algorithm explanation below.
 
-Note: The scraper has supported regex-based patterns for a long time, which is what it will still expect by default. However that is in the process of being replaced with parse-style patterns in the future versions of the package. To use the newer parse-style patterns, it's currently required to pass it as the ``format`` argument to the scraper.
+.. warning::
 
-Each such client relies on the `~sunpy.net.scraper.Scraper` to be able to query for files using the :meth:`~sunpy.net.scraper.Scraper.filelist` method. The general algorithm to explain how the `~sunpy.net.scraper.Scraper` is able to do this is:
+    The scraper class has supported regex-based patterns for a long time
+    However, that is in the process of being replaced with parse-style patterns.
+    This is supported by the ``format`` keyword argument to the scraper.
+    The regex-style patterns have be deprecated and will be removed in the future.
+
+Each such client relies on the `~sunpy.net.scraper.Scraper` to be able to query for files using the :meth:`~sunpy.net.scraper.Scraper.filelist` method.
+
+The general algorithm to explain how the `~sunpy.net.scraper.Scraper` is able to do this is:
 A brief explanation of how the Scraper works is as follows:
 
 1. Drop the filename from the pattern and generate a list of directories to search for files.
@@ -44,8 +51,8 @@ A brief explanation of how the Scraper works is as follows:
 
 For a more verbose in-depth explanation on how this is accomplished internally:
 
-The Scraper takes as input the generalised ``pattern`` of how a desired file path looks like, in the ``parse`` format.
-Upon initialisation, a version of the pattern following the datetime format is also internally generated, called the ``datetime_pattern``.
+The scraper takes as input the generalized ``pattern`` of how a desired file path looks like, in the ``parse`` format.
+Upon initialization, a version of the pattern following the datetime format is also internally generated, called the ``datetime_pattern``.
 
 .. code-block:: python
 
@@ -79,7 +86,8 @@ The location given by the filled pattern is visited and a list of files at the l
 This is handled differently depending on whether the pattern is a web URL or a ``file://`` or an ``ftp://`` path in the :meth:`~sunpy.net.scraper.Scraper.filelist` method.
 
 Each filename is then parsed against the remaining portion of the pattern to determine if it matches.
-Each such file is then checked for lying in the intended timerange using the :meth:`~sunpy.net.scraper._check_timerange` method which in turn uses :meth:`sunpy.net.scraper_utils.get_timerange_from_exdict` to get the covered timerange for each file. The files that satisfy these conditions are then added to the output.
+Each such file is then checked for lying in the intended timerange using the :meth:`~sunpy.net.scraper._check_timerange` method which in turn uses :meth:`sunpy.net.scraper_utils.get_timerange_from_exdict` to get the covered timerange for each file.
+The files that satisfy these conditions are then added to the output.
 
 .. code-block:: python
 
@@ -91,7 +99,9 @@ Each such file is then checked for lying in the intended timerange using the :me
 
 Writing a new "scraper" client
 ==============================
-The `~sunpy.net.scraper` thus allows us to write Fido clients for a variety of sources. For a simple example of a scraper client, we can look at the implementation of `sunpy.net.dataretriever.sources.eve.EVEClient` in sunpy.
+
+The `~sunpy.net.scraper` thus allows us to write Fido clients for a variety of sources.
+For a simple example of a scraper client, we can look at the implementation of `sunpy.net.dataretriever.sources.eve.EVEClient` in sunpy.
 
 A version without documentation strings is reproduced below:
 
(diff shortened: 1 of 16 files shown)


nabobalis:
Thanks, I will try to get some time to review this next week.

sunpy/net/scraper_utils.py
```diff
@@ -77,33 +86,66 @@ def date_floor(date, timestep):
 
     return datetime(*time_tup)
 
-def check_timerange(pattern, url, timerange):
+def extractDateURL(self, url):
```
samaloney, line 89:
```suggestion
def extract_date(self, url):
```

sunpy/net/scraper_utils.py
```diff
@@ -77,33 +86,66 @@ def date_floor(date, timestep):
...
-        exdict['year'] = 2000 + exdict['year']
-    if 'month' not in exdict:
-                if 'month_name' in exdict:
-                    exdict['month'] = datetime.strptime(exdict['month_name'], '%B').month
-                elif 'month_name_abbr' in exdict:
-                    exdict['month'] = datetime.strptime(exdict['month_name_abbr'], '%b').month
-    tr = get_timerange_from_exdict(exdict)
-    return tr.intersects(timerange)
+    # remove the user and passwd from files if there:
+    url = url.replace("anonymous:data@sunpy.org@", "")
```
samaloney, line 94: Can a user supply different credentials and if so would this break?

docs/topic_guide/extending_fido.rst
```diff
@@ -27,29 +27,42 @@ A new "scraper" client inherits from `~sunpy.net.dataretriever.client.GenericCli
...
   It returns a dictionary where keys are the supported attrs and values are lists of tuples.
   Each `tuple` contains the "attr" value and its description.
-* A class attribute ``pattern``; this is a string used to match the URLs supported by the client and extract metadata from the matched URLs.
-  The time and other metadata attributes for extraction are written using the `~sunpy.extern.parse.parse` format, using double curly-brackets so to differentiate them from ``kwargs`` parameters which are written in single curly-brackets.
+* A class attribute ``pattern``; this is a string used to match all URLs supported by the client and extract necessary metadata from the matched URLs.
+  The time and other metadata attributes for extraction are written within double curly-braces ``{{}}`` in a `parse` format. Regular placeholders for Python format strings can still be included via single curly braces ``{}`` with their respective parameters passed as the ``kwargs``. An example of how such a pattern looks like is given in the algorithm explanation below.
```
nabobalis, line 31:
```suggestion
  The time and other metadata attributes for extraction are written within double curly-braces ``{{}}`` in a `parse` format.
  Regular placeholders for Python format strings can still be included via single curly braces ``{}`` with their respective parameters passed as the ``kwargs``.
  An example of how such a pattern looks like is given in the algorithm explanation below.
```

sunpy/net/scraper_utils.py
exitflynn: i cant think of an example that'd hit this but i don't think it should break, this was most likely originally added since the current implementation [overrides ftp filelist-ing with these credentials]( https://github.com/sunpy/sunpy/blob/scraper_rewrite/sunpy/net/scraper.py#L207) and this function internally gets called later on.

Generalising it would be a simple regex change though, let me know if it seems like a good change to have.

However I'm a bit skeptic about this check ever passing since after looking up FTP URL structure the second `@` shouldn't be there?

exitflynn:
Thanks for your reviews! 

> This looks great, nice work. I know it came up before but do we have a migration guide or section in the docs explaining how to eventually move to the new format syntax.

A guide for writing clients with the newer format is added in the all-things-scraper-clients / extending-Fido [topic-guide](https://github.com/exitflynn/sunpy/blob/scraper_rewrite/docs/topic_guide/extending_fido.rst) but it's not explicitly about migrating yet. Does what's there work or should I try and address migration more?

docs/topic_guide/extending_fido.rst
```diff
@@ -27,29 +27,43 @@ A new "scraper" client inherits from `~sunpy.net.dataretriever.client.GenericCli
...
-  The time and other metadata attributes for extraction are written using the `~sunpy.extern.parse.parse` format, using double curly-brackets so to differentiate them from ``kwargs`` parameters which are written in single curly-brackets.
+* A class attribute ``pattern``; this is a string used to match all URLs supported by the client and extract necessary metadata from the matched URLs.
+  The time and other metadata attributes for extraction are written within double curly-braces ``{{}}`` in a `parse <https://github.com/r1chardj0n3s/parse/>`__ format.
+  Regular placeholders for Python format strings can still be included via single curly braces ``{}`` with their respective parameters passed as the ``kwargs``.
+  An example of how such a pattern looks like is given in the algorithm explanation below.
+
+Note: The scraper has supported regex-based patterns for a long time, which is what it will still expect by default. However that is in the process of being replaced with parse-style patterns in the future versions of the package. To use the newer parse-style patterns, it's currently required to pass it as the ``format`` argument to the scraper.
```
nabobalis, line 35:
```suggestion
Note: The scraper has supported regex-based patterns for a long time, which is what it will still expect by default.
However that is in the process of being replaced with parse-style patterns in the future versions of the package.
To use the newer parse-style patterns, it's currently required to pass it as the ``format`` argument to the scraper.
```

nabobalis: We want to enable the new format by default and raise a warning if its the old way.

docs/topic_guide/extending_fido.rst
```diff
@@ -27,29 +27,43 @@ A new "scraper" client inherits from `~sunpy.net.dataretriever.client.GenericCli
...
+
+1. Drop the filename from the pattern and generate a list of directories to search for files.
+2. For each directory, get a list of files present in it.
+3. For each file, check if it matches the pattern.
+4. For each file that matches the pattern, check if it is in the timerange. If it is, add it to the output.
+
+For a more verbose in-depth explanation on how this is accomplished internally:
 
-1. It takes as input a generalised ``pattern`` of how a desired filepath looks like, following the ``parse`` format. A version of the pattern following the datetime format is also generated, called the ``timepattern``.
+1. A Scraper object takes as input the generalised ``pattern`` of how a desired filepath looks like, in the ``parse`` format. Upon initialisation, a version of the pattern following the datetime format is also internally generated, called the ``dt_pattern``.
```
nabobalis, line 47:
```suggestion
1. The Scraper takes as input the generalised ``pattern`` of how a desired file path looks like, in the ``parse`` format.
   Upon initialisation, a version of the pattern following the datetime format is also internally generated, called the ``dt_pattern``.
```

docs/topic_guide/extending_fido.rst
```diff
@@ -27,29 +27,43 @@ A new "scraper" client inherits from `~sunpy.net.dataretriever.client.GenericCli
...
 
 .. code-block:: python
 
     >>> from sunpy.net import Scraper
     >>> pattern = ('http://proba2.oma.be/{instrument}/data/bsd/{{year:4d}}/{{month:2d}}/{{day:2d}}/'
     ...            '{instrument}_lv1_{{year:4d}}{{month:2d}}{{day:2d}}_{{hour:2d}}{{minute:2d}}{{second:2d}}.fits')
-    >>> s = Scraper(pattern, instrument='swap')
+    >>> s = Scraper(format=pattern, instrument='swap')
+    >>> s.dt_pattern
```
nabobalis, line 55: Thoughts on renaming this to "datetiime_pattern"?

docs/topic_guide/extending_fido.rst
```diff
@@ -27,29 +27,43 @@ A new "scraper" client inherits from `~sunpy.net.dataretriever.client.GenericCli
...
+    'http://proba2.oma.be/swap/data/bsd/%Y/%m/%d/swap_lv1_%Y%m%d_%H%M%S.fits'
 
-2. The smallest unit of time / time-step for that directory pattern (the full timepattern minus the filename at the end) is then detected by using :meth:`~sunpy.net.scraper_utils.extract_timestep`.
+The smallest unit of time / time-step for that directory-pattern (the full ``dt_pattern`` except the filename at the end) is then internally detected from ``dt_pattern`` by using the :meth:`~sunpy.net.scraper_utils.extract_timestep` util.
```
nabobalis, line 58:
```suggestion
The smallest unit of time / time-step for that directory-pattern (the full ``dt_pattern`` except the filename at the end) is then internally detected from ``dt_pattern`` by using :meth:`~sunpy.net.scraper_utils.extract_timestep`.
```

docs/topic_guide/extending_fido.rst
```diff
@@ -60,9 +74,9 @@ Each such client relies on the `~sunpy.net.scraper.Scraper` to be able to query
...
-5. The name of each file present is then examined to determine if it matches the remaining portion of the pattern using :meth:`~sunpy.extern.parse.parse`.
-6. Each such file is then checked for lying in the intended timerange using the :meth:`~sunpy.net.scraper_utils.check_timerange` method which in turn uses :meth:`sunpy.net.scraper_utils.get_timerange_from_exdict` to get the covered timerange for each file. The files that satisfy these conditions are then added to the output.
+2. The location given by the filled pattern is visited and a list of files at the location is obtained. This is handled differently depending on whether the pattern is a web URL or a ``file://`` or an ``ftp://`` path in the :meth:`~sunpy.net.scraper.Scraper.filelist` method.
+3. Each filename is then parsed against the remaining portion of the pattern to determine if it matches.
+4. Each such file is then checked for lying in the intended timerange using the :meth:`~sunpy.net.scraper._check_timerange` method which in turn uses :meth:`sunpy.net.scraper_utils.get_timerange_from_exdict` to get the covered timerange for each file. The files that satisfy these conditions are then added to the output.
```
nabobalis, line 79:
```suggestion
The location given by the filled pattern is visited and a list of files at the location is obtained.
This is handled differently depending on whether the pattern is a web URL or a ``file://`` or an ``ftp://`` path by :meth:`~sunpy.net.scraper.Scraper.filelist`.

Each filename is then parsed against the remaining portion of the pattern to determine if it matches.
Each such file is then checked for lying in the intended timerange using the :meth:`~sunpy.net.scraper._check_timerange` method which in turn uses :meth:`sunpy.net.scraper_utils.get_timerange_from_exdict` to get the covered timerange for each file.
The files that satisfy these conditions are then added to the output.
```

docs/topic_guide/extending_fido.rst
```diff
@@ -145,18 +159,11 @@ Examples
 
 Suppose any file of a data archive can be described by this URL ``https://some-domain.com/%Y/%m/%d/satname_{SatelliteNumber}_{Level}_%y%m%d%H%M%S_{any-2-digit-number}.fits``:
 
-``baseurl`` becomes ``r'https://some-domain.com/%Y/%m/%d/satname_(\d){2}_(\d){1}_(\d){12}_(\d){2}\.fits'``.
+The new format ``pattern`` becomes ``r'https://some-domain.com{{year:4d}}/{{month:2d}}{{day:2d}}/satname_{SatelliteNumber:2d}_{Level:1d}_{{year:2d}}{{month:2d}}{{day:2d}}{{hour:2d}}{{minute:2d}}{{second:2d}}_{{:2d}}.fits'``.
+The date-time values and any other metadata attributes that we wish to extract are written within double curly-braces ``{{}}``. These metadata attributes are the desired keys for the returned dictionary and they should match with the ``attr.__name__``.
+Note that parts of such attributes can accordingly be omitted to match parts of the filename which are dynamic but not needed to be extracted. For example, ``{{:2d}}`` is used in the above example to match any 2-digit number in the filename. Similarly ``{{}}`` can be used to match a string of any length starting from its position in the filename.
```
nabobalis, line 164:
```suggestion
The new format ``pattern`` becomes ``r'https://some-domain.com{{year:4d}}/{{month:2d}}{{day:2d}}/satname_{SatelliteNumber:2d}_{Level:1d}_{{year:2d}}{{month:2d}}{{day:2d}}{{hour:2d}}{{minute:2d}}{{second:2d}}_{{:2d}}.fits'``.
The datetime values and any other metadata attributes that we wish to extract are written within double curly-braces ``{{}}``.
These metadata attributes are the desired keys for the returned dictionary and they should match with the ``attr.__name__``.
Note that parts of such attributes can accordingly be omitted to match parts of the filename which are dynamic but not needed to be extracted.
For example, ``{{:2d}}`` is used in the above example to match any 2-digit number in the filename.
Similarly ``{{}}`` can be used to match a string of any length starting from its position in the filename.
```

docs/topic_guide/extending_fido.rst
```diff
@@ -175,6 +182,7 @@ So the desired key names for returned dictionary should be written in the ``patt
 
         return adict
 
+
```
nabobalis, line 185:
```suggestion
```

sunpy/net/scraper.py
```diff
@@ -40,18 +54,32 @@ class Scraper:
     Parameters
     ----------
     pattern : `str`
-        A string containing the url with the date and other information to be
+        A string containing the url with the date encoded as datetime formats,
+        and any other parameter as ``kwargs`` as a string format.
+        This can also be a uri to a local file patterns. Deprecated in favor of `format`. Default is `None`.
+
```
nabobalis, line 60: You should remove the line breaks between each parameter.

sunpy/net/scraper.py
```diff
@@ -72,24 +102,55 @@ class Scraper:
...
-        if "year:4d" in pattern and "year:2d" in pattern:
-            pattern = pattern.replace("year:2d", ":2d")
-        self.pattern = pattern
-        self.domain = f"{urlsplit(self.pattern).scheme}://{urlsplit(self.pattern).netloc}/"
-        milliseconds = re.search(r'\%e', self.timepattern)
-        if not milliseconds:
-            self.now = datetime.now().strftime(self.timepattern)
+    def __init__(self, pattern=None, regex=False, format=None, **kwargs):
+        if pattern is not None and format is None:
+            warnings.warn(
```
nabobalis, line 107: I thought we had a helper to raise deprecation warnings for keywords. I can't seem to find it, only for positional arguments.

sunpy/net/scraper.py
```diff
@@ -72,24 +102,55 @@ class Scraper:
...
+                "Using 'pattern' for the URL format is deprecated. "
+                "Please use 'format' instead.",
+                DeprecationWarning,
```
nabobalis, line 110: This should be a sunpydepcreationwaring instead.

sunpy/net/scraper.py
```diff
@@ -72,24 +102,55 @@ class Scraper:
...
+            )
+            self.use_old_format = True
+            self.pattern = pattern
+        elif pattern is None and format is not None:
+            self.use_old_format = False
+            self.pattern = format
```
nabobalis, line 116: So if both are set, it will use format and ignore the pattern?

I would rather this errored and told users to only format.

sunpy/net/scraper.py
```diff
@@ -192,7 +260,7 @@ def filelist(self, timerange):
             return self._ftpfilelist(timerange)
         elif urlsplit(directories[0]).scheme == "file":
             return self._localfilelist(timerange)
-        elif urlsplit(directories[0]).scheme == "http":
+        elif urlsplit(directories[0]).scheme == "http" or urlsplit(directories[0]).scheme == "https":
```
nabobalis, line 263:
```suggestion
        elif urlsplit(directories[0]).scheme in ["http", "https"]:
```

sunpy/net/scraper.py
```diff
@@ -308,14 +397,131 @@ def _httpfilelist(self, timerange):
...
+                exdict['year'] = 2000 + exdict['year']
+            if 'month' not in exdict:
+                        if 'month_name' in exdict:
+                            exdict['month'] = datetime.strptime(exdict['month_name'], '%B').month
+                        elif 'month_name_abbr' in exdict:
+                            exdict['month'] = datetime.strptime(exdict['month_name_abbr'], '%b').month
+            tr = get_timerange_from_exdict(exdict)
+            return tr.intersects(timerange)
+
+    def _URL_followsPattern(self, url):
```
nabobalis, line 439:
```suggestion
    def _url_follows_pattern(self, url):
```
Elsewhere as well.

sunpy/net/scraper.py
```diff
@@ -308,14 +397,131 @@ def _httpfilelist(self, timerange):
...
+            return False
+        else:
+            return parse(self.pattern, url)
+
+
+    def _extract_date(self, url):
+        """
+        Extracts the date from a particular url following the pattern.
+        """
+        # remove the user and passwd from files if there:
```
nabobalis, line 459:
```suggestion
        # Remove the user and password if present
```

sunpy/net/scraper.py
```diff
@@ -308,14 +397,131 @@ def _httpfilelist(self, timerange):
...
+        url = url.replace("anonymous:data@sunpy.org@", "")
+
+        def url_to_list(txt):
+            # Substitutes '.' and '_' for '/'.
+            return re.sub(r'\.|_', '/', txt).split('/')
+
+        # create a list of all the blocks in times - assuming they are all
+        # separated with either '.', '_' or '/'.
```
nabobalis, line 467:
```suggestion
        # Create a list of all the blocks in times
        # assuming they are all separated with either '.', '_' or '/'
```

nabobalis (dismissed):
Thanks for doing all of this! I left some minor comments. I also rebased the base branch, you might need to rebase.

nabobalis:
> but it's not explicitly about migrating yet. Does what's there work or should I try and address migration more?

There should be a short migration guide either in the change log or in the main fido client doc so we can point to it in the deprecation warning.

nabobalis:
We also should document what needs to be removed when we want to drop the old code paths

sunpy/net/tests/test_scraper.py
```diff
@@ -1,94 +1,173 @@
...
-    ('%b%y', relativedelta(months=1)),
-    ('%m%y', relativedelta(months=1)),
-    ('%H%d', relativedelta(hours=1)),
-    ('%y%b', relativedelta(months=1)),
-]
-
 
 def testDirectoryDatePattern():
-    s = Scraper('%Y/%m/%d/%Y%m%d_%H%M%S_59.fit.gz')
+    with pytest.deprecated_call():
```
nabobalis, line 16: When you change to sunpydepcreationwarning, I am unsure if this will work. 

I would also like to check the message is the one we expect to be raised.

sunpy/net/tests/test_scraper.py
```diff
@@ -1,94 +1,173 @@
...
 
-PATTERN_EXAMPLES = [
-    ('%b%y', relativedelta(months=1)),
-    ('%m%y', relativedelta(months=1)),
-    ('%H%d', relativedelta(hours=1)),
-    ('%y%b', relativedelta(months=1)),
-]
-
 
 def testDirectoryDatePattern():
```
nabobalis, line 15: Can we rename these tests to follow the others in sunpy?

`test_directory_date_pattern` and etc.

sunpy/net/scraper.py
```diff
@@ -80,24 +99,77 @@ class Scraper:
     pattern looks with the actual time.
     """
 
-    def __init__(self, pattern, regex=False, **kwargs):
-        if regex:
+    def __init__(self, format=None, pattern=None, regex=False, **kwargs):
```
nabobalis, line 102: Sorry, I was wrong in my previous comment about this. I will blame covid and not my own incompetence.

We want to keep the API the same way around as it will break all the older clients. 

I haven't thought through what I am about to say but I think we would need to do this:

"pattern", "regex "are the first two arguments to the init, then it will be "format" for the new behavior. 

We check if pattern/regex are passed in, we issue a deprecation warning saying to use format keyword and provide a message with a link to a migration guide.

I would also make the format keyword to be required to be a keyword as a temporary workaround. TO avoid any API ambiguity.

exitflynn: > I would also make the format keyword to be required to be a keyword as a temporary workaround.
ooh do you mean having the format keyword as a required argument? because wouldn't that also break all the older clients?

also wish you a fast and comfy recovery from covid!! and thanks for all the reviews, health issues have also got to me these days, i should be able to address them over the weekend

nabobalis: > > I would also make the format keyword to be required to be a keyword as a temporary workaround.
> 
> do you mean having the format keyword as a required argument? because wouldn't that also break all the older clients?

No. I mean the use of `*` in the signature. 

`def __init__(self, pattern=None, regex=False, *, format=None, **kwargs):`

> also wish you a fast and comfy recovery from covid!! and thanks for all the reviews, health issues have also got to me these days, i should be able to address them all over the weekend.

Make sure to look after yourself!

sunpy/net/scraper.py
exitflynn: hey i can't seem to figure out which of the existing decorators would apply in this case, [here they are for reference](https://github.com/sunpy/sunpy/blob/main/sunpy/util/decorators.py). does one of them fit here or does this call for a new one for deprecating specific arguments?

sunpy/net/scraper.py
exitflynn: i think the else condition afterwards would catch that instead and raise the error?

docs/topic_guide/extending_fido.rst
exitflynn: i think that works better, yeah

exitflynn:
Thanks for the reviews Nabil, I've added back the `deprecated_renamed_argument` decorator from [astropy](https://github.com/astropy/astropy/blob/main/astropy/utils/decorators.py#L321) to handle the deprecation warnings and addressed pretty much all of the reviews so far. What remains for now is a proper migration guide at the changelog or the docs (or I'm missing something lol) which I should be able to get to very soon!

nabobalis:
> Thanks for the reviews Nabil, I've added back the `deprecated_renamed_argument` decorator from [astropy](https://github.com/astropy/astropy/blob/main/astropy/utils/decorators.py?rgh-link-date=2024-06-26T22%3A15%3A52Z#L321) to handle the deprecation warnings and addressed pretty much all of the reviews so far. What remains for now is a proper migration guide at the changelog or the docs (or I'm missing something lol) which I should be able to get to very soon!

Sounds good! I will review this again.

nabobalis:
In the meantime I will rebase this PR to fix those conflicts. You will need to reset your local branch when I do this to avoid having to force push again.

sunpy/net/scraper.py
nabobalis: We should check with a unit test.

nabobalis:
TODO:
We need to add a test that creates an old style scraper client (basically use one of the old clients) and make sure that it raises the warning correctly and still works.

nabobalis:
I screwed up the rebase @exitflynn, you can force push your local branch to fix my mistake please?

nabobalis:
I rebased this, hopefully I got everything.

nabobalis (approved):

nabobalis:
That online fail is real.

nabobalis:
Oh its from helio. Nevermind.

nabobalis:
Thanks again @exitflynn

PR #7665, opened 2024-06-07
resample hmi jp2 file

nabobalis:
The original file was by 410 by 410 so I made it 512 by 512, a nicer number we can all agree on.

Oh and made the WCS not be this:

old_hmi

It now looks like:

new_hmi


nabobalis:
This brings up a test case, do we want to validate our test data via figure tests?

nabobalis:
WE HAD THIS AS A FIGURE TEST AND WE NEVER NOTICED?

wtbarnes:
Out of curiosity what did you do to fix the wcs? What was the problem?

nabobalis:
I downloaded the original file from Helioviewer, opened it, resampled it and saved it out.

wtbarnes:
Ah ok so our original file was just messed up somehow.

wtbarnes (approved):

PR #7678, opened 2024-06-19
Scale back ANA deprecation

Cadair:
I don't think #7642 got enough review, I do not for a moment believe that that PR meets the criteria for a single review merge.

I am not going to deprecate ANA without *actively* reaching out to the community giving them an opportunity to say it's being used. I also am unconvinced that the proposed alternative is viable and maintained.

I have opened a discourse post which I have put links to in the messages, but before this gets upgraded to a full deprecation warning we should send out a message (maybe with the release notes for 6.0) to solarnews etc.

Also to be abundantly clear: This is not me endorsing a eventual removal, I am still against getting rid of this at all, but we might as well ask people.


nabobalis:
> I don't think #7642 got enough review, I do not for a moment believe that that PR meets the criteria for a single review merge.

It did.

> I am not going to deprecate ANA without _actively_ reaching out to the community giving them an opportunity to say it's being used. I also am unconvinced that the proposed alternative is viable and maintained.

Our solution is not viable and or maintained.
Can only here prove that this file format is still active use?

> I have opened a discourse post which I have put links to in the messages, but before this gets upgraded to a full deprecation warning we should send out a message (maybe with the release notes for 6.0) to solarnews etc.

I have no issue with reaching out to the community. 

> Also to be abundantly clear: This is not me endorsing a eventual removal, I am still against getting rid of this at all, but we might as well ask people.

This should be removed, this code offers nothing to sunpy.

I am against making this a pending deprecation. 
I suspect I will be out voted.

Cadair:
As mentioned at the community call we should check that `SunpyPendingDeprecationWarning` is show to users by default.

nabobalis:
> As mentioned at the community call we should check that `SunpyPendingDeprecationWarning` is show to users by default.

It did not for me.

PR #7679, opened 2024-06-19
Build aarch64 wheels

Cadair:
This enables aarch64 wheels.

Due to the fact we are currently emulating aarch64 to build and test these (hopefully soon we will have native aarch64 runners) I have setup the tests to only run the `sunpy/io/tests/test_ana.py` file as a way of checking the compiled code works and saving a signifiant amount of build time (this turned out to be harder than it should have been).

When we have native runners for aarch64 we should be able to row this back to be trivial.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -183,7 +183,7 @@ jobs:
       anaconda_package: sunpy
       anaconda_keep_n_latest: 1
       sdist: false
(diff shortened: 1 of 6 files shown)


nabobalis:
Imagine not having to build wheels anymore.

nabobalis:
Should get a feature(?) changelog.

nabobalis (approved):

PR #7681, opened 2024-06-19
TST: fix incorrect upper bound on numpy for oldestdeps tox env

neutrinoceros:
## PR Description
I noticed that oldestdeps jobs were accidentally running numpy 2.0.0 (see [example logs](https://github.com/sunpy/sunpy/actions/runs/9577652835/job/26406459276))
This resolves it.


Cadair:
Thanks!

Cadair (approved):

Cadair:
I really must fix minimum-dependancies so that we can just use that and get rid of this *&$*£(! madness

PR #7682, opened 2024-06-19
Be more rigorous with how we handle dates in map

### LINKED ISSUE #7673: Overhaul how we handle dates/times for maps

This is spun out from #7606.  I propose an overhaul of how we handle dates/times for maps.  To briefly summarize the FITS standard:

- DATE-OBS is commonly used in older files, and usually (but not always!) refers to the start of the observation
- DATE-BEG refers to the start of the observation
- DATE-END refers to the end of the observation
- DATE-AVG refers to the "average" time of the observation, and the standard is suggestive that (but not explicit) that the observer metadata is coupled to this time.  Also, the standard does not prescribe a specific way to calculate this average.

To briefly summarize our current code, GenericMap has the following properties:

- `.date_start`, which is just DATE-BEG
- `.date_end`, which is just DATE-END
- `.date_average`, which is DATE-AVG (if present), or falls back to the arithmetic mean of DATE-BEG and DATE-END (if present)
- `.date`, which is in order of decreasing priority: DATE-OBS > `.date_average` > `.date_start` > `.date_end` > the current time

Issues:
- The `.wcs` property is instantiated with only one of the possible times, with `.wcs.wcs.dateobs` set to `.date`, and the other times are not preserved.  Also, since the coordinate frame constructed from this WCS uses `.date`, it may not be accurate if DATE-OBS is present but is not the same time as the observer metadata (e.g., DATE-AVG).
- The `.observer_coordinate` property is instantiated with its time as `.date`, and as above, that may not be accurate if DATE-OBS is present but is not the same time as the observer metadata (e.g., DATE-AVG).

Proposed overhaul plan:

1. Re-define `.date_average` to be the reference time for observer metadata.  The new logic would be in order of decreasing priority: DATE-AVG > `.date`.  We should remove the assumption that the arithmetic mean of DATE-BEG and DATE-END is the correct way to construct the average time.
2. Re-define `.date` to be the "canonical" time to refer to an observation, which is commonly the start of the observation, and would be the time that, e.g., is used in plot titles.  The new priority would be DATE-OBS > `.date_start` > DATE-AVG > `.date_end` > session time (see https://github.com/sunpy/sunpy/issues/5543#issuecomment-932221515).
3. Re-define `.observer_coordinate` so that it uses `.date_average` instead of `.date`
4. Re-define `.wcs` so that it populates the following:
  - `.wcs.wcs.dateobs` with `.date` (note that this will never be `None`)
  - `.wcs.wcs.dateavg` with `.date_average` (note that this will never be `None`)
  - `.wcs.wcs.datebeg` with `.date_start` if not `None`
  - `.wcs.wcs.dateend` with `.date_end` if not `None`
5. Define our WCS->frame mapping so that it prioritizes `.wcs.dateavg` over `.wcs.dateobs`.  This is actually already the case, but I list it for completeness: https://github.com/sunpy/sunpy/blob/558e55e5f2c1f573d10db464d29061b582b8ece2/sunpy/coordinates/wcs_utils.py#L37

Considerations:
1. If we do the above, we might want to change the name of `.date_average`, and potentially of `.date`, to make it more clear and to enable some deprecation.
2. There would be the potential confusion that `.observer_coordinate.date` would be equal to `.date_average`, and potentially different from `.date.`.

Cadair:
fixes #7673

### DIFF
--- a/docs/whatsnew/6.0.rst
+++ b/docs/whatsnew/6.0.rst
@@ -57,6 +57,15 @@ The following extra group have been added:
 Additionally, the "all" extra group is a meta group that includes all optional dependencies.
 The "dask" extra group has also been removed.
 
+Decoupling dates in `~sunpy.map.Map`
+====================================
+Previously, `~sunpy.map.Map` used the single `~sunpy.map.GenericMap.date` property to serve two, potentially conflicting purposes: the "canonical" observation time and the reference time for the coordinate system.
+In order to properly support map sources where those two times are not the same, there is a new `~sunpy.map.GenericMap.reference_date` property to serve as the reference time for the coordinate system.
+`~sunpy.map.GenericMap.date` continues to serve as the canonical observation time, which is commonly the start of the observation period.
+See the documentation for the respective properties for the logic of parsing the metadata.
+
+An example of a user-facing impact is that, for certain map sources, the time used by the `~sunpy.map.GenericMap.coordinate_frame` and `~sunpy.map.GenericMap.observer_coordinate` properties can be different from the time printed in the default title of plots.
+
 Removal of ``sunpy.database``
 =============================
 
--- a/sunpy/coordinates/wcs_utils.py
+++ b/sunpy/coordinates/wcs_utils.py
@@ -34,7 +34,7 @@ def solar_wcs_frame_mapping(wcs):
     if hasattr(wcs, "coordinate_frame"):
         return wcs.coordinate_frame
 
-    dateobs = wcs.wcs.dateavg or wcs.wcs.dateobs or None
+    dateobs = wcs.wcs.dateavg or wcs.wcs.dateobs or wcs.wcs.datebeg or wcs.wcs.dateend or None
 
     # Get observer coordinate from the WCS auxiliary information
     # Note: the order of the entries is important, as it determines which set
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -864,7 +864,7 @@ class GenericMap(NDData):
     @property
     def date(self):
         """
-        Image observation time.
+        'Canonical' observation time.
 
         For different combinations of map metadata this can return either
         the start time, end time, or a time between these. It is recommended
@@ -874,16 +874,18 @@ class GenericMap(NDData):
 
         Taken from, in order of preference:
 
-        1. The DATE-OBS FITS keyword
-        2. `~sunpy.map.GenericMap.date_average`
-        3. `~sunpy.map.GenericMap.date_start`
+        1. `~sunpy.map.GenericMap.date_start`
+        2. The ``DATE-OBS`` or ``DATE_OBS`` FITS keywords
+        3. `~sunpy.map.GenericMap.date_average`
         4. `~sunpy.map.GenericMap.date_end`
         5. The current time
         """
-        time = self._date_obs
-        time = time or self.date_average
-        time = time or self.date_start
-        time = time or self.date_end
+        time = (
+            self.date_start or
+            self._date_obs or
+            self.date_average or
+            self.date_end
+        )
 
         if time is None:
             if self._default_time is None:
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -611,8 +611,15 @@ class GenericMap(NDData):
         # FITS standard doesn't allow both PC_ij *and* CROTA keywords
         w2.wcs.crota = (0, 0)
         w2.wcs.cunit = self.spatial_units
-        w2.wcs.dateobs = self.date.isot
         w2.wcs.aux.rsun_ref = self.rsun_meters.to_value(u.m)
+
+        w2.wcs.dateavg = self.reference_date.isot
+        if self._date_obs is not None:
+            w2.wcs.dateobs = self._date_obs.isot
+        if self.date_start is not None:
+            w2.wcs.datebeg = self.date_start.isot
+        if self.date_end is not None:
+            w2.wcs.dateend = self.date_end.isot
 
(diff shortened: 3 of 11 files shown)


sunpy/coordinates/wcs_utils.py
```diff
@@ -34,7 +34,7 @@ def solar_wcs_frame_mapping(wcs):
     if hasattr(wcs, "coordinate_frame"):
         return wcs.coordinate_frame
 
-    dateobs = wcs.wcs.dateavg or wcs.wcs.dateobs or None
+    dateobs = wcs.wcs.dateavg or wcs.wcs.dateobs or wcs.wcs.datebeg or wcs.wcs.dateend or None
```
Cadair, line 37: Given we don't know where the WCS has come from, it makes sense to me to use the same hierarchy here as in `reference_date` in map.

sunpy/map/mapbase.py
```diff
@@ -611,9 +611,17 @@ def wcs(self):
         # FITS standard doesn't allow both PC_ij *and* CROTA keywords
         w2.wcs.crota = (0, 0)
         w2.wcs.cunit = self.spatial_units
-        w2.wcs.dateobs = self.date.isot
         w2.wcs.aux.rsun_ref = self.rsun_meters.to_value(u.m)
 
+
```
Cadair, line 616:
```suggestion
```

sunpy/map/sources/psp.py
```diff
@@ -36,6 +36,10 @@ def __init__(self, data, header, **kwargs):
         self.plot_settings['norm'] = ImageNormalize(
             stretch=AsinhStretch(a=0.001), vmin=0)
 
+    @property
+    def reference_date(self):
```
Cadair, line 40: There is a unit test which was asserting this, but I don't know if that's because the test was written to test the default map behaviour or if it was encoding actual knowledge of WISPR data.

I wonder if @svank happens to know which date header in a WISPR file is the one that corresponds to the spacecraft location?

sunpy/map/sources/solo.py
```diff
@@ -46,11 +46,6 @@ def processing_level(self):
             # The level number is prepended by the letter L
             return int(self.meta.get('level')[1:])
 
-    @property
-    def date(self):
```
Cadair, line 50: I assume this isn't wildly controversial, but probably should get its own changelog.

sunpy/map/sources/psp.py
svank: WISPR documents the DATE-* cards on page 8 [here](https://wispr.nrl.navy.mil/sites/wispr.nrl.navy.mil/files/wispr_data_user_guide_v3.pdf). `DATE-OBS` is `DATE-BEG`, and `DATE-AVG` is my choice for a representative spacecraft location. (Each WISPR image is an on-spacecraft sum of 5-ish exposures, running from `BEG` to `END`)

Cadair: I shall swap this to use DATE-AVG on your recommendation, but it would be nice to know from the WISPR team what date stamp they use for the corresponding location they put in the headers.

svank: For one sample file, I can reproduce the HCI coords in the header using `DATE-AVG`, so that must be it.

sunpy/map/mapbase.py
```diff
@@ -613,13 +613,12 @@ def wcs(self):
         w2.wcs.cunit = self.spatial_units
         w2.wcs.aux.rsun_ref = self.rsun_meters.to_value(u.m)
 
-        w2.wcs.dateavg = self.reference_date.isot
-        if self._date_obs is not None:
-            w2.wcs.dateobs = self._date_obs.isot
+        w2.wcs.dateobs = self.date.utc.isot
```
Cadair, line 616: What's the reasoning for including date-obs on the wcs if it hasn't come from the date-obs header?

ayshih: Our metadata approach/model, where a source subclass is supposed to fix a wrong DATE-OBS or redirect to a non-standard keyword by overriding `.date`.  For constructing the WCS object, we should look only at (potentially overridden) properties, not at the keywords themselves.  (Technically `_date_obs` is a property that could be overridden, but it is private and undocumented.)  So, the use cases basically fall into three categories (far from an equipartition):

1. The user wants the canonical observation time on the WCS object and DATE-OBS is correct.
2. The user wants the canonical observation time on the WCS object and the `.date` property has been overridden to not pull it straight from DATE-OBS.
3. The user doesn't care whether the WCS object has the canonical observation time.

Sticking `.date` in `.wcs.wcs.dateobs` in all cases means that (1) and (2) are satisfied and (3) has not been negatively impacted.

Cadair:
@ayshih's commit looks good to me, I think this is ready if Albert does.

ayshih (approved):
I'll approve this PR to count for the my approval of the stuff I didn't write and @Cadair's approval of the stuff he didn't write. =)  We still need another approver.

nabobalis:
This should get a whatsnew entry.

docs/whatsnew/6.0.rst
```diff
@@ -115,16 +124,3 @@ You can specify this by passing the ``hdus`` keyword argument to `~sunpy.map.Map
 .. code-block:: python
 
     >>> sunpy.map.Map("adapt40311_03k012_202401020800_i00005600n1.fts.gz", hdus=0)  # doctest: +SKIP
-
-New `~sunpy.coordinates.PlanarScreen` context manager
```
nabobalis, line 119: This was a duplicate?

ayshih: No, must be a weird merge...

nabobalis (approved):

wtbarnes (approved):

nabobalis:
Thanks everyone. ONWARD TO 6.0

PR #7696, opened 2024-07-01
Update base Norh URL

nabobalis:
See if this works.

### DIFF
--- a/changelog/7696.trivial.rst
+++ b/changelog/7696.trivial.rst
@@ -1,0 +1,1 @@
+Updated the internal URL for the `~sunpy.net.dataretriever.sources.norh.NoRHClient` to point to a HTTPS archive of the NoRH data.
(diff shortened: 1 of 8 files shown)


exitflynn:
welp im beginning to think they might not have just announced a new URL yet, nothing of that sort comes up for the past year when you search `Nobeyama Radioheliograph  "ftp"`

exitflynn:
i've figured out why this is still failing and it exposes an edge-case that is not accounted for in the Scraper, which would be of clients having `http`-baseurls and serving data in extension-less files. `_httpfilelist()` looks for URLs at a directory's webpage [ending with the same extension as the pattern](https://github.com/sunpy/sunpy/blob/main/sunpy/net/scraper.py#L290) (like `.fits`, etc) before parsing it further. This wasn't a problem before since the source was ftp then and it's straightforward to get a list of files in a directory. i could work on fixing this if it'd be more convenient that way, let me know!

ayshih (approved):
I think it'd be good to have a (trivial) changelog entry

nabobalis:
Added the changelog.

nabobalis:
Thanks for the review @ayshih

Issue #7268, opened 2023-10-27
Documentation builds are very slow (and caching is not effective)


ehsteve:
### Describe the performance issue

When i build the docs whether using `make html` or `tox -e build_docs` the build deletes previously generated files such as the api files and then have to regenerate them every time. This makes the build very slow. On my machine, it takes 5 mins for make html and half that time with tox.

Do we need to regenerate those files on every build?

### To Reproduce

make html

or 

tox -e biuld_docs

### Proposed fix

Do not regenerate those files in each build.

Cadair:
The issue seems to be that a large number of files are getting invalidated in sphinx's cache (maybe not all of them?) when it's not clear why!

ayshih:
Is this a recent development?  In the past, my `api` files would definitely be cached when using `make html`, which I know because I'd have to manually clear it out to see API changes that I made reflected in the docs.

`tox` does `tox` things, so I can't speak for `tox`

ayshih:
Hmm, okay, maybe `sunpy.coordinates` API files aren't getting invalidated =)

ayshih:
Heh, maybe it's a Mac thing!

ayshih:
Oh, never mind then

ayshih:
What Sphinx version?

nabobalis:
Yeah I am interested, did they do something in a new release.

nabobalis:
7.2.6 was the version

ayshih:
I can definitely reproduce this now, but I'm befuddled by what has changed.  I'm pretty sure it hadn't been like this a few months ago when I was still working on fixing the inheritance-diagram links.  I tried downgrading to Sphinx 5.3.0, and it's still rebuilding everything.  Maybe something having to do with our updated theme?

ayshih:
Perhaps there is poor caching, but I think there is something really borked with our (new) theme.  I just timed a couple of docs re-builds.  When I re-build the docs under our theme, the writing of all of the (non-cached) HTML files takes ~12 minutes(!).  If I disable our theme and use the default Sphinx theme instead, the same writing takes ~50 *seconds*.  That is a monumental difference!  What is our theme doing?!?

nabobalis:
I wonder if it's worth asking upstream?

wtbarnes:
What happens if you just use the plain pydata theme? That would help nail down whether it's the base theme or something specific to the sunpy one.

ayshih:
[...]

PR #7703, opened 2024-07-03
Maybe fix docs build cache?

alasdairwilson:
Obviously run.id will never work but what will...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,8 +47,20 @@ jobs:
     secrets:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
+  hash-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+      - name: Set cache key
+        id: set-cache-key
+        run: |
+          echo "cache-key=docs-${{ hashFiles('examples/**/*.py', '**/*.rst') }}" >> $GITHUB_ENV
+          echo "::set-output name=cache-key::docs-${{ hashFiles('examples/**/*.py', '**/*.rst') }}"
+        shell: bash
+
   docs:
-    needs: [core]
+    needs: [core, hash-docs]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@main
     with:
       default_python: '3.12'
@@ -59,7 +71,7 @@ jobs:
         docs/_build/
         docs/generated/
         .tox/sample_data/
-      cache-key: docs-${{ hashFiles('examples/**/*.py', '**/*.rst') }}
+      cache-key: ${{ needs.compute-cache-key.outputs.cache-key }}
       libraries: |
         apt:
           - graphviz
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,9 +55,19 @@ jobs:
       - name: Set cache key
         id: set-cache-key
         run: |
+          echo "${{ github.run_id }}"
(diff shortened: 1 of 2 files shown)


nabobalis:
Why would run.id never work?

.github/workflows/ci.yml
```diff
@@ -59,7 +71,7 @@ jobs:
         docs/_build/
         docs/generated/
         .tox/sample_data/
-      cache-key: docs-${{ github.run_id }}
+      cache-key: ${{ needs.compute-cache-key.outputs.cache-key }}
```
nabobalis, line 74: There is another:
```
          cache-key: docs-${{ github.run_id }}
```
under the gallery build, is that fine to stay or does that need the same hash as this?

.gitignore
```diff
@@ -5,6 +5,7 @@ __pycache__/
 *.py[cod]
 *$py.class
 tmp/
+/fakedirectory/
```
nabobalis, line 8:
```suggestion
/properlyfakedirectory/
```

.github/workflows/ci.yml
```diff
@@ -83,7 +95,7 @@ jobs:
             docs/_build/
             docs/generated/
             .tox/sample_data/
-          cache-key: docs-${{ github.run_id }}
+          cache-key: ${{ needs.compute-cache-key.outputs.cache-key }}
```
nabobalis, line 98: No this didnt work.

nabobalis:
I am going to make random changes for fun.

alasdairwilson:
So yeah my attempt obv didnt work but run.id is unique per run so it would never find a cache?

I don't know why what I did didn't work.

Might be worth testing with just a static cache key first tbf.

.github/workflows/ci.yml
```diff
@@ -59,7 +71,7 @@ jobs:
         docs/_build/
         docs/generated/
         .tox/sample_data/
-      cache-key: docs-${{ github.run_id }}
+      cache-key: ${{ github.sha }}
```
nabobalis, line 74: Testing if just the sha makes this work. 

Then we need to work out a more complex hashing scheme.

Cadair:
The cache restore seems to have worked (I think it was before?) but the issue is that sphinx thinks all the source files have changed: https://github.com/sunpy/sunpy/actions/runs/9786453564/job/27021782660?pr=7703#step:10:696

Cadair:
To be clear, I tried this locally and got the same result. In a clean git directory I ran `tox -e build_docs` twice and in the second run I get this:
```
building [html]: targets for 0 source files that are out of date
updating environment: 0 added, 414 changed, 0 removed
```

I am getting the feeling we have looked into this before. @ayshih do you remember anything about this?

Cadair:
Ah maybe what we did before was this: https://github.com/sunpy/sunpy/issues/7268 but that doesn't seem to have resolved the fact the cache is bork.

Cadair:
Actually I think run_id should work for us. The idea of this cache is to make the gallery build not have to rebuild the whole doc tree. According to [the docs](https://docs.github.com/en/actions/learn-github-actions/contexts) run ID is:

> A unique number for each workflow run within a repository. This number does not change if you re-run the workflow run.

That sounds right?

alasdairwilson:
> Actually I think run_id should work for us. The idea of this cache is to make the gallery build not have to rebuild the whole doc tree. According to [the docs](https://docs.github.com/en/actions/learn-github-actions/contexts) run ID is:
> 
> > A unique number for each workflow run within a repository. This number does not change if you re-run the workflow run.
> 
> That sounds right?

You are, of course, correct.

I had interpreted it as a per job id rather than an id for the entire workflow:

I guess that means the original cache key worked but yet the cache...does not.

Closing this I guess.

PR #7704, opened 2024-07-04
Fix crash by checking FILLVAL attribute for floating-point data types

### LINKED ISSUE #7565: read_cdf/TimeSeries crashes if CDF variable doesn't have a 'FILLVAL' attribute
### Describe the bug

Similar to https://github.com/sunpy/sunpy/issues/5907, I stumbled upon CDF files that aren't fully [complying to the CDF ISTP Metadata Guidelines](https://spdf.gsfc.nasa.gov/sp_use_of_cdf.html) by having a variable that doesn't provide a corresponding FILLVAL attribute. `read_cdf()` is expecting this attribute to exist, and otherwise crashes (and with it also `TimeSeries`):
https://github.com/sunpy/sunpy/blob/9f0a37dc0d0a543663e871a3256b147fc004db2e/sunpy/io/_cdf.py#L83-L87

But as has been mentioned in https://github.com/sunpy/sunpy/issues/5907#issuecomment-1054428488, it's not uncommon that a CDF file is not fully following the ISTP standard (these are by the way recent PSP data files, so I'll anyhow contact the instrument team). If the `FILLVAL` information isn't provided by the instrument team, one could just skip the replacement call. So I propose to change `if np.issubdtype(data.dtype, np.floating):` to something like `if np.issubdtype(data.dtype, np.floating) and 'FILLVAL' in attrs:`

### To Reproduce

Because the error message occurs in `read_cdf` when using `TimeSeries`, this is what I'm calling here directly:

``` python
import datetime as dt
from sunpy.io._cdf import read_cdf
from sunpy.net import Fido
from sunpy.timeseries import TimeSeries

dataset = 'PSP_SWP_SPI_SF00_L3_MOM'
trange = a.Time(dt.date(2023,3,14), dt.date(2023,3,15))
result = Fido.search(trange, a.cdaweb.Dataset(dataset))
downloaded_files = Fido.fetch(result)
# ts = TimeSeries(downloaded_files, concatenate=True)
data = read_cdf(downloaded_files.data[0])
```

### Screenshots

_No response_

### System Details

- sunpy version: 5.1.1 and 5.1.2

### Installation method

pip

RehanChalana:
## PR Description

Fix crash by checking FILLVAL attribute for floating-point data types

Fixes #7565

### DIFF
--- a/sunpy/io/_cdf.py
+++ b/sunpy/io/_cdf.py
@@ -83,7 +83,7 @@ def read_cdf(fname):
             # Set fillval values to NaN
             # It would be nice to properley mask these values to work with
             # non-floating point (ie. int) dtypes, but this is not possible with pandas
-            if np.issubdtype(data.dtype, np.floating):
+            if np.issubdtype(data.dtype, np.floating) and 'FILLVAL' in attrs:
                 data[data == attrs['FILLVAL']] = np.nan
 
             # Get units


nabobalis:
Thanks for the PR, I think this needs a changelog entry as a bugfix.

nabobalis:
Hi @RehanChalana, since there has been no activity on this PR. I will be closing it. If you want to continue working on this, please re-open it!

PR #7707, opened 2024-07-05
Fix the docstrings for Map cached properties

ayshih:
We cache a few of the properties of Map (since #4467), but apparently no one in the subsequent four years has noticed or pointed out that the docstrings were not showing up.  This was due to a bug in the call to `functools.wraps()`.

### DIFF
--- a/sunpy/util/decorators.py
+++ b/sunpy/util/decorators.py
@@ -159,7 +159,7 @@ def cached_property_based_on(attr_name):
         """
         prop: the property method being decorated
         """
-        @wraps(outer)
+        @wraps(prop)
         def inner(instance):
             """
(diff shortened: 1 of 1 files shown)


ayshih:
Examples:
https://docs.sunpy.org/en/v5.1.4/generated/api/sunpy.map.GenericMap.html#sunpy.map.GenericMap.observer_coordinate
https://docs.sunpy.org/en/v5.1.4/generated/api/sunpy.map.GenericMap.html#sunpy.map.GenericMap.wcs

nabobalis (approved):

nabobalis:
Bot is asleep, I will do the manual backport.

PR #7710, opened 2024-07-07
Fix cron jobs

nabobalis:


sunpy/timeseries/tests/test_timeseries_factory.py
```diff
@@ -98,6 +99,7 @@ def test_from_url():
     assert isinstance(ts[1], sunpy.timeseries.GenericTimeSeries)
 
 
+@skip_numpy2
 def test_read_cdf():
```
nabobalis, line 103: On conda, cdflib has not been updated yet.

.github/workflows/ci.yml
```diff
@@ -207,7 +206,7 @@ jobs:
         github.event_name == 'pull_request' &&
         contains(github.event.pull_request.labels.*.name, 'Run publish')
       )
-    needs: [test, docs]
+    needs: [test, docs, publish]
```
nabobalis, line 209: Since these are failing to upload sometime, I want to make sure the normal wheels get uploaded before the aarch64 wheels are attempted.

tox.ini
```diff
@@ -134,16 +133,6 @@ commands =
     {posargs}
     python -c 'import pathlib; print("Documentation available under file://\{0\}".format(pathlib.Path(r"{toxinidir}") / "docs" / "_build" / "index.html"))'
 
-[testenv:base_deps]
```
nabobalis, line 137: This was replaced with the core_deps build.

.github/workflows/ci.yml
nabobalis: Also, not sure if I need to report this upstream but both upload jobs are uploading all of the artifacts, regardless of which job created them.

I assume this is not intended behaviour.

Cadair (approved):

wtbarnes (approved):

Issue #7715, opened 2024-07-10
Remove the skip_numpy2 decorator and remove the last test using it.


nabobalis:
On conda, cdflib has not been updated yet.

_Originally posted by @nabobalis in https://github.com/sunpy/sunpy/pull/7710#discussion_r1667726270_
            

In our conda cron job, this test was failing due to numpy 2 and CDF reader on conda forge not having the recipe updated.

nabobalis:
This can not be done until the library on conda-forge has been done.

PR #7716, opened 2024-07-10
Removes skip_numpy2 decorator

Prtm2110:
This PR Removes skip_numpy2 decorator and also takes care of the tests related to it
Solves Issue #7715


nabobalis:
Thanks for the PR @Prtm2110 but this can not be merged as we need this until the code is fixed upstream.

I would suggest another issue.

PR #7717, opened 2024-07-10
Update simple_map and add figure test for it

nabobalis:
Goal is to make the simple map, not 3 by 3 but 9 by 9.

### DIFF
--- a/sunpy/map/tests/conftest.py
+++ b/sunpy/map/tests/conftest.py
@@ -107,12 +107,15 @@ def make_simple_map():
     and scaled differently in each direction.
     """
     data = np.arange(81).reshape((9, 9))
-    ref_coord = SkyCoord(0.0, 0.0, frame='helioprojective', obstime='now', unit='deg',
+    ref_coord = SkyCoord(0.0, 0.0, frame='helioprojective', obstime='2020-01-01 00:00:00', unit='deg',
                          observer=SkyCoord(0 * u.deg, 0 * u.deg, 1 * u.AU, frame='heliographic_stonyhurst'))
     ref_pix = [4, 4] * u.pix
     scale = [2, 1] * u.arcsec / u.pix
     header = sunpy.map.make_fitswcs_header(data, ref_coord, reference_pixel=ref_pix, scale=scale)
     return sunpy.map.Map(data, header)
+
+
+simple_map = pytest.fixture(make_simple_map)
 
 
 @pytest.fixture
@@ -152,6 +155,3 @@ def sample_hmi():
     from sunpy.data.sample import HMI_LOS_IMAGE
 
     return sunpy.map.Map(HMI_LOS_IMAGE)
-
-
-simple_map = pytest.fixture(make_simple_map)
--- a/sunpy/map/tests/conftest.py
+++ b/sunpy/map/tests/conftest.py
@@ -101,9 +101,10 @@ def generic_map():
     return sunpy.map.Map((data, header))
 
 
-def make_simple_map():
+@pytest.fixture
+def simple_map():
     """
-    A simple 9x9 map, with it's center at (0, 0),
+    A simple 9x9 map, with its center at (0, 0),
     and scaled differently in each direction.
     """
     data = np.arange(81).reshape((9, 9))
@@ -113,9 +114,6 @@ def make_simple_map():
     scale = [2, 1] * u.arcsec / u.pix
     header = sunpy.map.make_fitswcs_header(data, ref_coord, reference_pixel=ref_pix, scale=scale)
     return sunpy.map.Map(data, header)
-
-
-simple_map = pytest.fixture(make_simple_map)
 
 
 @pytest.fixture
--- a/sunpy/map/tests/conftest.py
+++ b/sunpy/map/tests/conftest.py
@@ -102,19 +102,17 @@ def generic_map():
 
 
 def make_simple_map():
-    # A 3x3 map, with it's center at (0, 0), and scaled differently in
-    # each direction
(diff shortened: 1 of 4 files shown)


sunpy/map/tests/conftest.py
```diff
@@ -141,14 +139,19 @@ def eit_test_map():
...
 
+
 @pytest.fixture
 def sample_hmi():
     from sunpy.data.sample import HMI_LOS_IMAGE
 
     return sunpy.map.Map(HMI_LOS_IMAGE)
+
+
+simple_map = pytest.fixture(make_simple_map)
```
Cadair, line 157: I'd rather have this under the function, it's pretty weird that it's separate anyway. Keeping it there will make it more obvious.

nabobalis: Fixed.

sunpy/map/tests/conftest.py
```diff
@@ -102,13 +102,14 @@ def generic_map():
...
 def make_simple_map():
-    # A 3x3 map, with it's center at (0, 0), and scaled differently in
-    # each direction
-    data = np.arange(9).reshape((3, 3))
-    ref_coord = SkyCoord(0.0, 0.0, frame='helioprojective', obstime='now', unit='deg',
-                         observer=SkyCoord(0 * u.deg, 0 * u.deg, 1 * u.AU,
-                                           frame='heliographic_stonyhurst'))
-    ref_pix = [1, 1] * u.pix
+    """
+    A simple 9x9 map, with it's center at (0, 0),
```
ayshih, line 106:
```suggestion
    A simple 9x9 map, with its center at (0, 0),
```

sunpy/map/tests/test_mapbase.py
```diff
@@ -849,7 +844,7 @@ def test_resample(simple_map, shape):
 
     resampled_upper_left = resampled.pixel_to_world((shape[0] - 0.5) * u.pix,
```
ayshih, line 845:
```suggestion
    resampled_upper_right = resampled.pixel_to_world((shape[0] - 0.5) * u.pix,
```

sunpy/map/tests/test_mapbase.py
```diff
@@ -849,7 +844,7 @@ def test_resample(simple_map, shape):
...
                                                     (shape[1] - 0.5) * u.pix)
-    original_upper_left = simple_map.pixel_to_world(2.5 * u.pix, 2.5 * u.pix)
+    original_upper_left = simple_map.pixel_to_world(8.5 * u.pix, 8.5 * u.pix)
     assert u.allclose(resampled_upper_left.Tx, original_upper_left.Tx)
     assert u.allclose(resampled_upper_left.Ty, original_upper_left.Ty)
```
ayshih, line 849:
```suggestion
    original_upper_right = simple_map.pixel_to_world(8.5 * u.pix, 8.5 * u.pix)
    assert u.allclose(resampled_upper_right.Tx, original_upper_right.Tx)
    assert u.allclose(resampled_upper_right.Ty, original_upper_right.Ty)
```

sunpy/map/tests/conftest.py
```diff
@@ -102,13 +102,14 @@ def generic_map():
 
 
 def make_simple_map():
```
ayshih, line 104:
```suggestion
@pytest.fixture
def simple_map():
```
(and remove the `simple_map = ...` below)

I don't understand why this isn't made a fixture via decorator like every other fixture we define.  You'll also need to modify the following lines, but those tests should really be using the fixture anyway:
https://github.com/sunpy/sunpy/blob/2b416fdacb755aaf9cf94b960a9995b2997758a4/sunpy/map/tests/test_mapbase.py#L39
https://github.com/sunpy/sunpy/blob/2b416fdacb755aaf9cf94b960a9995b2997758a4/sunpy/map/tests/test_mapbase.py#L1028
https://github.com/sunpy/sunpy/blob/2b416fdacb755aaf9cf94b960a9995b2997758a4/sunpy/map/tests/test_mapbase.py#L1049

sunpy/map/tests/test_mapbase.py
```diff
@@ -1026,7 +1021,6 @@ def test_superpixel_fractional_inputs(generic_map):
 @given(pc=matrix_meta('pc'))
 def test_resample_rotated_map_pc(pc, method):
     smap = make_simple_map()
```
ayshih, line 1023: Change this to use the fixture

sunpy/map/tests/test_mapbase.py
```diff
@@ -1047,7 +1041,6 @@ def test_resample_rotated_map_pc(pc, method):
 @given(cd=matrix_meta('cd'))
 def test_resample_rotated_map_cd(cd, method):
     smap = make_simple_map()
```
ayshih, line 1043: Change this to use the fixture

ayshih (commented):
Some picky comments that are admittedly tangential to the point of this PR

sunpy/map/tests/test_mapbase.py
nabobalis: I think the reason is:

```
Function-scoped fixtures are not reset between examples generated by
`@given(...)`, which is often surprising and can cause subtle test bugs.
```
But I can work around this.

nabobalis:
> Some picky comments that are admittedly tangential to the point of this PR

Happy to fix anything you spot!

nabobalis:
I added figure tests for both map fixtures, someone should look at them and decide if we want to expand or change them.

wtbarnes (approved):
There's currently a conflict with the dev figure hashes, but otherwise it looks good to me

PR #6743, opened 2023-01-28
Deprecate `meta_keywords()`

exitflynn:
This was originally #6722 but I accidentally closed it since I didn't realise that force pushing would do that. 
Would close #6562.

PR #7730, opened 2024-07-16
Parse DN correctly in header helper

wtbarnes:
After #7585, "DN" is now correctly handled by `GenericMap`. However, the header helper was still applying strict checking against the FITS standard when parsing the unit string, meaning one could not pass `unit='DN'` into the header helper. This fixes this bug by pulling the unit parsing logic out of `GenericMap` so it can be used in the header helper as well.

This also removes the `_map_meta_keywords` dictionary which as far as I can tell is unused anywhere and was deprecated in #6743


nabobalis (approved):

wtbarnes:
Wait...this does not do what I want it to do...

nabobalis:
> Wait...this does not do what I want it to do...
> 
> `_parse_fits_unit` still returns a `Unit`. It needs to be a string. Maybe this is not the right approach.

Time for a second method

nabobalis (changes requested):
CI is very broken

wtbarnes:
Oops I forgot it was used other places.

sunpy/map/tests/test_header_helper.py
```diff
@@ -210,12 +210,18 @@ def test_quantity_input(map_data, hpc_coord):
     header = make_fitswcs_header(map_quantity, hpc_coord, unit=override_unit)
     assert header['bunit'] == override_unit.to_string('fits')
 
-
```
nabobalis, line 213: This line should be added back.

sunpy/map/tests/test_header_helper.py
```diff
@@ -210,12 +210,18 @@ def test_quantity_input(map_data, hpc_coord):
...
     # Test that unit can be passed in as a string
     map_unit = u.Unit('ct / (pix s)')
     header = make_fitswcs_header(map_data, hpc_coord, unit=map_unit.to_string())
-    assert header['bunit'] == map_unit.to_string('fits')
+    assert header['bunit'] == map_unit.to_string(format='fits')
+
+
+@pytest.mark.parametrize(('input_unit','output_string'),
+                         [('DN', 'DN'), (u.DN, 'DN'), ('DN / s', 'DN / s')])
+def test_unit_handles_dn(input_unit, output_string, map_data, hpc_coord):
```
nabobalis, line 222:
```suggestion
def test_make_fitswcs_header_handles_dn(input_unit, output_string, map_data, hpc_coord):
```

nabobalis (approved):

nabobalis:
setuptools 71.0.1 came out 3 hours ago and broke things, we can pin out that verison.

wtbarnes:
Oh lovely. I was wondering what in the world was happening.

Cadair:
you broke it: https://github.com/sunpy/sunpy/actions/runs/9992411608/job/27617589005?pr=7730#step:10:757

wtbarnes:
> you broke it: https://github.com/sunpy/sunpy/actions/runs/9992411608/job/27617589005?pr=7730#step:10:757

How is that possible

nabobalis:
It needs to point to the api link and not the import. Is my guess.

changelog/7730.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed a bug where "DN" was not able to be parsed by `~sunpy.map.make_fitswcs_header` due to strict checking
```
nabobalis, line 1:
```suggestion
Fixed a bug where "DN" was not able to be parsed by `~sunpy.map.header_helper.make_fitswcs_header` due to strict checking
```

changelog/7730.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed a bug where "DN" was not able to be parsed by `~sunpy.map.header_helper.make_fitswcs_header` due to strict checking
```
nabobalis, line 1: Hopefully this will work

wtbarnes: Ah got it. Thanks!

PR #7737, opened 2024-07-18
replace usages of ``copy_arrays`` with ``memmap``

zacharyburnett:
## PR Description

follow-on to https://github.com/asdf-format/asdf/pull/1797

### DIFF
--- a/sunpy/io/special/asdf/tests/test_genericmap.py
+++ b/sunpy/io/special/asdf/tests/test_genericmap.py
@@ -48,7 +48,7 @@ def test_genericmap_mask(aia171_test_map, tmpdir):
 @asdf_entry_points
 def test_load_100_file_with_shift():
     fname = get_test_filepath("aiamap_shift_genericmap_1.0.0.asdf")
-    with asdf.open(fname, copy_arrays=True) as af:
+    with asdf.open(fname, memmap=False) as af:
         aiamap = af['object']
         assert isinstance(aiamap, sunpy.map.sources.AIAMap)
         assert "crval1" in aiamap.meta.modified_items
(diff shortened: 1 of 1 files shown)


changelog/7737.trivial.rst
```diff
@@ -0,0 +1 @@
+replace usages of ``copy_arrays`` with ``memmap``
```
nabobalis, line 1: Since this in the test file, this does not need a changelog

nabobalis (approved):

nabobalis:
Thanks for the PR @zacharyburnett

zacharyburnett:
@nabobalis I apologize, I forgot to include that `memmap` only exists in `asdf>=3.1.0`; there should also probably be a pin in `pyproject.toml` to that effect, or alternatively I can add code to check the `asdf` version and use `copy_arrays` or `memmap` accordingly

nabobalis:
As long as our test suite passed, which I think it did even on our oldest deps, I don't see the issue?

nabobalis:
The oldest deps despite the pin, used asdf==3.3.0, I assume maybe a 310 cant install older versions?

zacharyburnett:
it looks like `oldestdeps` in `tox.ini` pins `asdf<2.12.0`
https://github.com/sunpy/sunpy/blob/e5878584d57ed21916b390987cfe7d641387540c/tox.ini#L61
however the project requires `asdf>=2.13.0`
https://github.com/sunpy/sunpy/blob/e5878584d57ed21916b390987cfe7d641387540c/pyproject.toml#L58
and since the `pip install .` step occurs after the dependencies from `toxenv:deps` are installed, this means the oldestdeps in `tox.ini` are overwritten:
https://github.com/sunpy/sunpy/actions/runs/9993966424/job/27622904664#step:10:21

nabobalis:
Wait, our CI wasn't updated to use the new min versions?!

I will have to change that.  

Either way, this is just test changes so I have no issue as long as our tests pass.

zacharyburnett:
We could use something like https://github.com/OpenAstronomy/minimum_dependencies to gather the oldest deps and install those (see https://github.com/spacetelescope/jwst/blob/de39de07b6f0380860caa86e8485666b4cef35c2/tox.ini#L51-L64)

nabobalis:
We already do elsewhere, it just wasn't added to sunpy for some reason I don't know.

PR #7744, opened 2024-07-19
Make XRTMap have default units of DN

wtbarnes:
Fixes a bug where the XRT map source was still using default units of `u.ct` instead of `u.DN`

### DIFF
--- a/changelog/7744.bugfix.rst
+++ b/changelog/7744.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed a bug where `~sunpy.map.sources.XRTMap` was still defaulting to `~astropy.units.ct` rather than
+`~astropy.units.DN`.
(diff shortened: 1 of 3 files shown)


nabobalis (approved):

changelog/7744.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed a bug where `~sunpy.map.sources.XRTMap` was still defaulting to `~astropy.units.ct` rather than
+`~astropy.units.DN`.
```
nabobalis, line 2:
```suggestion
Fixed a bug where `~sunpy.map.sources.XRTMap` was still defaulting to counts rather than DN.
```
<unknown>:9: WARNING: py:obj reference target not found: astropy.units.ct
<unknown>:9: WARNING: py:obj reference target not found: astropy.units.DN

wtbarnes: I knew that was going to happen.

PR #7753, opened 2024-07-24
added unit nitpick

nabobalis:
Also fix for the drms 0.8.0 release which changes the API docs

### DIFF
--- a/sunpy/image/tests/test_transform.py
+++ b/sunpy/image/tests/test_transform.py
@@ -7,7 +7,7 @@ from skimage import transform as tf
 from astropy.coordinates.matrix_utilities import rotation_matrix
 
 from sunpy.image.transform import _rotation_registry, affine_transform
-from sunpy.tests.helpers import figure_test
+from sunpy.tests.helpers import figure_test, skip_windows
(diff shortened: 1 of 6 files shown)


sunpy/image/tests/test_transform.py
```diff
@@ -245,6 +245,7 @@ def test_float32(identity):
     assert np.issubdtype(out_arr.dtype, np.float32)
 
 
+@skip_windows
```
nabobalis, line 248: This is now broken on windows and I don't have the skills to debug why.

nabobalis: I should open an issue to track this.

wtbarnes: This makes me nervous but ok

wtbarnes (approved):

alasdairwilson (approved):

sunpy/image/tests/test_transform.py
nabobalis: We will have to track this down but its a numpy issue on windows. Nothing we can do to fix that outside of reporting it upstream.

Issue #7757, opened 2024-07-25
Have AIA and HMI maps use T_OBS (or T_REC?) for `reference_date`


ayshih:
Apparently T_OBS is the reference observation time for AIA and HMI metadata, at least for Level-1 data products.  (DATE-OBS is derived from T_OBS by subtracting half the exposure time, so would not be a reference time.)  That means that we should use T_OBS for constructing the coordinate frame and observer location.  With #7682, we can now simply override the `reference_date` property so that T_OBS is used.

For derived data products (e.g., HARP images), T_REC might be the correct keyword to use rather than T_OBS.

nabobalis:
> For derived data products (e.g., HARP images), T_REC might be the correct keyword to use rather than T_OBS.

I think we would need to create a Map source for this?

wtbarnes:
> > For derived data products (e.g., HARP images), T_REC might be the correct keyword to use rather than T_OBS.
> 
> I think we would need to create a Map source for this?

Well HARPs are always HMI maps right so we could just have a check in the metadata that says `if HARP then date_key="T_REC"` or something like that. Currently, one can load SHARPs data into a `Map` using just the HMI map source.

nabobalis:
I dont think we have a test header, we should add one and check.

wtbarnes:
Sure, I guess we should test against SHARP and HARP data in addition to the usual HMI data.

nabobalis:
I will add them to the open PR.

PR #7758, opened 2024-07-25
Update reference_date for AIA Map and Gong

nabobalis:
<whimper> The reason that #7758 breaks differential_rotate() is because it chooses to return the output map as the same subclass as input map, but it of course has no idea how to correctly modify the metadata so that the source subclass understands the new reference time. All the function modifies is DATE-OBS. For AIAMap, differential_rotate() would need to modify T_OBS. (Even for subclasses that do not override .reference_date, if the input metadata has DATE-AVG, differential_rotate() would need to modify DATE-AVG.) Thus, the output map has the wrong

Okay, I'm thinking we add ._set_date() and ._set_reference_date() methods to GenericMap, and any subclass that overrides .date/.reference_date needs to override the corresponding setter as well.reference time, which triggers a transformation of the bounding box that isn't supposed to happen, which triggers the submap warning for off-disk corners.

Fixes https://github.com/sunpy/sunpy/issues/7757

TODO:

- [x] Update other SDO Map sources
- [x] Changelog
- [x] Unit tests
- [x] Fix online doctests


sunpy/map/tests/test_mapbase.py
```diff
@@ -380,13 +380,13 @@ def test_coordinate_frame(aia171_test_map):
...
-    assert frame.obstime == aia171_test_map.date
+    assert frame.obstime == aia171_test_map.reference_date
 
 
 def test_heliographic_longitude_crln(hmi_test_map):
     assert_quantity_allclose(hmi_test_map.heliographic_longitude,
-                             hmi_test_map.carrington_longitude - sun.L0(hmi_test_map.date),
-                             rtol=1e-3)  # A tolerance is needed because L0 is for Earth, not SDO
+                             hmi_test_map.carrington_longitude - sun.L0(hmi_test_map.reference_date),
+                             rtol=0.1)  # A tolerance is needed because L0 is for Earth, not SDO
```
nabobalis, line 389: Not sure if this change is correct.

sunpy/map/tests/test_mapbase.py
```diff
@@ -380,13 +380,13 @@ def test_coordinate_frame(aia171_test_map):
     assert frame.observer.lat == aia171_test_map.observer_coordinate.frame.lat
     assert frame.observer.lon == aia171_test_map.observer_coordinate.frame.lon
     assert frame.observer.radius == aia171_test_map.observer_coordinate.frame.radius
-    assert frame.obstime == aia171_test_map.date
+    assert frame.obstime == aia171_test_map.reference_date
```
nabobalis, line 383: Should be this ok?

sunpy/map/sources/sdo.py
```diff
@@ -188,14 +207,15 @@ def scale(self):
     def date(self):
         """
         Image observation time.
+        """
+        return self._get_date('IMG_FRST')
```
nabobalis, line 211: I was not sure what to use for this.

sunpy/physics/tests/test_differential_rotation.py
```diff
@@ -224,14 +226,18 @@ def test_differential_rotate_time_all_on_disk(all_on_disk_map):
...
     # This rotated map should have a larger extent in the x direction
     new_time = all_on_disk_map.date + 48*u.hr
-    with pytest.warns(UserWarning, match="Using 'time' assumes an Earth-based observer"):
-        dmap = differential_rotate(all_on_disk_map, time=new_time)
-    assert dmap.data.shape[1] > all_on_disk_map.data.shape[1]
+    new_observer = get_earth(new_time)
+    with SphericalScreen(new_observer):
+        with pytest.warns(UserWarning, match="Using 'time' assumes an Earth-based observer"):
+            dmap = differential_rotate(all_on_disk_map, time=new_time)
+    assert dmap.data.shape[1] > all_on_disk_map.data.shape[1] -2 # TODO HACK LMAO
```
nabobalis, line 240: This is wrong.

ayshih: Now moot

sunpy/map/tests/test_mapbase.py
ayshih: Should be

sunpy/map/tests/test_mapbase.py
ayshih: I still need to look into why the tolerance needed to be changed

sunpy/map/sources/sdo.py
ayshih: Assuming we don't get actual input from an HMI person, we should stare at bunch of the HMI files to see whether this is an appropriate choice

nabobalis: I should just send an email.

sunpy/map/tests/test_mapbase.py
ayshih: A `.date` needed to be updated to `.reference_date` within `.carrington_longitude`

sunpy/map/sources/sdo.py
```diff
@@ -188,14 +214,21 @@ def scale(self):
...
```
nabobalis, line 218:
```suggestion
        return self._get_date('T_OBS')
```

sunpy/map/sources/sdo.py
```diff
@@ -188,14 +214,21 @@ def scale(self):
     def date(self):
         """
         Image observation time.
+        """
+        return self._get_date('T_OBS')
 
-        This is taken from the 'DATE-OBS' or 'T_OBS' keywords.
+    def _set_date(self, date):
+        self.meta['IMG_FRST'] = parse_time(date).utc.isot
```
ayshih, line 221:
```suggestion
        self.meta['T_OBS'] = parse_time(date).utc.isot
```

_typos.toml
```diff
@@ -1,5 +1,6 @@
 default.extend-ignore-identifiers-re = [
-    "ANDed", # ANDed
+    "ANDed",
+    "FRST", # IMG_FRST keyword
```
ayshih, line 3: This can be removed now?

nabobalis: Oh yes. I forgot.

sunpy/map/sources/sdo.py
nabobalis: Woops. Thanks.

_typos.toml
nabobalis, line 3:
```suggestion
```

nabobalis:
I should fix this properly and fix the last few commits.

Cadair (commented):
Can someone explain the need for the set methods? That's very antithetical to the fundamental model

ayshih:
The `differential_rotate()` function returns a differentially rotated map as an instance of the input's source subclass (e.g., `AIAMap`), and it needs to be able to set the new reference date to be the rotated time.  However, the function has no knowledge of the source-specific keyword that needs to be modified to set the new reference date.  (Previously, the function merely set `DATE-OBS`, which only works if the source doesn't provide `DATE-AVG` or doesn't override `.reference_date` to use a different keyword.)  That is, we need to be able to write the metadata such that the source-specific metadata parsing can understand it.  The expedient solution was to have the source subclasses also define private setters.

An alternative is to follow the approach I have insisted on for the `.reproject_to()` method: return a `GenericMap` rather than pretend the output is of the same source subclass.  However, that would be a breaking change to the API of `differential_rotate()`.

ayshih:
The problem is that we have metadata interpreters, but no metadata writers.  Another alternative would be to store a time shift in the Map instance (defaulting to zero shift).  We wouldn't touch the metadata at all then, but all metadata interpreters – both GenericMap and source-specific overriders – would need to add the time shift in all of the date properties.  That's conceptually cleaner, but then I don't know how we save such a Map to FITS.

nabobalis:
TODO: We need a changelog for diff rot as well

sunpy/map/mapbase.py
```diff
@@ -929,6 +925,19 @@ def reference_date(self):
...
         )
 
+    def _set_reference_date(self, date):
+        """
+        Set the reference date using the same priority as `.GenericMap.reference_date`.
+
+        If a source subclass overrides `.GenericMap.reference_date`, it should override
+        this private method as well.
+        """
+        for keyword in ['date-avg', 'date-obs', 'date-beg', 'date-end']:
```
Cadair, line 935: Rather than reimplementing all this boiler plate code for the getter and setter why don't we take a leaf out of the coordinate frame book and define a list of keys for the various properties on each source?

(I would be ok with that being a followup PR if we want to get this one out)

sunpy/map/mapbase.py
```diff
@@ -2711,10 +2736,16 @@ def contour(self, level, **kwargs):
         >>> aia = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)  # doctest: +REMOTE_DATA
         >>> contours = aia.contour(50000 * u.DN)  # doctest: +REMOTE_DATA
         >>> print(contours[0])  # doctest: +REMOTE_DATA
-            <SkyCoord (Helioprojective: obstime=2011-06-07T06:33:02.770, rsun=696000.0 km, observer=<HeliographicStonyhurst Coordinate (obstime=2011-06-07T06:33:02.770, rsun=696000.0 km): (lon, lat, radius) in (deg, deg, m)
-        (-0.00406308, 0.04787238, 1.51846026e+11)>): (Tx, Ty) in arcsec
+        <SkyCoord (Helioprojective: obstime=2011-06-07T06:33:02.880, rsun=696000.0 km, observer=<HeliographicStonyhurst Coordinate (obstime=2011-06-07T06:33:02.880, rsun=696000.0 km): (lon, lat, radius) in (deg, deg, m)
+        (-0.00406429, 0.04787238, 1.51846026e+11)>): (Tx, Ty) in arcsec
         [(719.59798458, -352.60839064), (717.19243987, -353.75348121),
-        ...
```
Cadair, line 2717: Do we not want to keep the truncation?

sunpy/map/sources/tests/test_eit_source.py
```diff
@@ -5,25 +5,37 @@
...
 __author__ = "Pritish C. (VaticanCameos)"
 
 
-@pytest.fixture(scope="module", params=header_list)
-def eit_map(request):
+@pytest.fixture()
+def eit_map():
     """Creates an EITMap from a FITS file."""
-    return get_dummy_map_from_header(request.param)
+    return sunpy.map.Map(get_test_filepath("EIT/efz20040301.000010_s.fits"))
```
Cadair, line 19: What's going on here?

sunpy/map/sources/tests/__init__.py
```diff
@@ -0,0 +1,6 @@
+def _test_private_date_setters(map_object):
```
Cadair, line 1: This seems like a place people will not look for this, can you put it in a helpers.py?

Cadair (changes requested):
I think this is basically good. I am fine with the use of the private setters to modify the metadata in things like diff rot, I thought we were calling them inside the map sources at first glance.

sunpy/map/mapbase.py
```diff
@@ -199,11 +199,7 @@ def __init_subclass__(cls, **kwargs):
         cls.__doc__ = fix_duplicate_notes(_notes_doc, cls.__doc__)
 
         if hasattr(cls, 'is_datasource_for'):
-            # NOTE: This conditional is due to overlapping map sources in sunpy and pfsspy that
```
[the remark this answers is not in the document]
nabobalis: Restored.

sunpy/map/mapbase.py
nabobalis: I leave this up to @ayshih to decide.

sunpy/map/mapbase.py
nabobalis: I dont see the harm in having the whole thing, its not very long.

sunpy/map/sources/tests/test_eit_source.py
nabobalis: I need to check that date and reference_date are correct but we have like 6 file headers we run these tests though (for no reason I can see). So I just use the first one.

sunpy/map/sources/tests/__init__.py
nabobalis: Do we want people to look for this?

Cadair: people writing or editing these tests may need to look for it.

sunpy/map/sources/tests/test_eit_source.py
Cadair: but you have now done this for *all* the tests? This seems like a backwards step? Why not just make the date tests use the single map?

nabobalis: Yeah and it doesnt matter. My plan is to remove all of these fits files in a future PR.

sunpy/map/sources/tests/__init__.py
nabobalis: Moved

sunpy/map/sources/gong.py
```diff
@@ -34,10 +35,6 @@ class GONGSynopticMap(GenericMap):
     array of 242×256 pixels covering the solar disk. These magnetograms are used to derive
     synoptic maps which show a full-surface picture of the solar magnetic field.
 
-    Notes
-    -----
-    If you have ``pfsspy`` installed this map source will be used instead of the one built into ``pfsspy``.
-
```
Cadair, line 40: Restore this? It's still true right?

sunpy/map/sources/gong.py
```diff
@@ -54,8 +51,20 @@ def is_datasource_for(cls, data, header, **kwargs):
 
     @property
     def date(self):
+        # The FITS file has a date that is made from the date-obs and time-obs keywords
+        # Which is not what date-obs is supposed to be
         return Time(f"{self.meta.get('date-obs')} {self.meta.get('time-obs')}")
 
+    def _set_date(self, date):
+        self.meta['date-obs'], self.meta['time-obs'] = parse_time(date).utc.iso.split(' ')
```
Cadair, line 59: This is insanely picky, but using T feels like a better separator to me.
```suggestion
        self.meta['date-obs'], self.meta['time-obs'] = parse_time(date).utc.isot.split('T')
```

sunpy/map/sources/sdo.py
```diff
@@ -130,6 +144,18 @@ def observatory(self):
         """
         return self.meta.get('telescop', '').split('/')[0]
 
+    @property
+    def reference_date(self):
```
Cadair, line 148: A idle thought, how much other stuff is duplicated between the AIA and HMI sources? Would it be worth having a `SDOMap` class with the shared stuff?

sunpy/map/sources/soho.py
```diff
@@ -143,6 +143,13 @@ def date(self):
         time = self.meta.get('time-obs', self.meta.get('time_obs'))
         return parse_time(f"{date}T{time}")
 
+    def _set_date(self, date):
+        if 'time-obs' in self.meta:
```
Cadair, line 147: It's inconsistent that we "fix" the header here by putting the whole date&time into `date-obs` but we don't do that for GONG.

I am not sure what one we should prefer. Maybe @ayshih has thoughts.

sunpy/map/sources/tests/test_swap_source.py
```diff
@@ -7,20 +7,26 @@
 
 from sunpy.data.test import get_dummy_map_from_header, get_test_filepath
 from sunpy.map.sources.proba2 import SWAPMap
-
-header_files = [
```
Cadair, line 11: What are the differences between these files?

Cadair (commented):
Sorry, more comments.

sunpy/map/sources/gong.py
ayshih: Heh, picky indeed!  If a wayward space were to creep into the `TimeISO` format, I think the chaos would extend far beyond this one line.

sunpy/map/sources/soho.py
ayshih: I wrote this, and the reason that I did it this way is because the overridden `date` property for `LASCOMap` (unlike `GONGSynopticMap`) has a logic branch anticipates that some headers will have a whole DATE-OBS.

I suppose a better approach would be to follow the original header: if it had a whole DATE-OBS, write a whole DATE-OBS; if it had a split DATE-OBS/TIME-OBS, write a split DATE-OBS/TIME-OBS.

sunpy/map/mapbase.py
ayshih: It gets trickier if the reference date is split across multiple keys.  Future PR!

sunpy/map/sources/sdo.py
ayshih: Future PR! =)

Cadair: yes totally.

sunpy/map/sources/tests/test_swap_source.py
nabobalis: Time. It's just a sequence of files.

sunpy/map/sources/gong.py
```diff
@@ -34,10 +35,6 @@ class GONGSynopticMap(GenericMap):
     array of 242×256 pixels covering the solar disk. These magnetograms are used to derive
     synoptic maps which show a full-surface picture of the solar magnetic field.
```
nabobalis, line 37:
```suggestion

    Notes
    -----
    If you have ``pfsspy`` installed this map source will be used instead of the one built into ``pfsspy``.
```

sunpy/map/sources/soho.py
nabobalis: I pushed a rough first go at this for Gong. See what you both think.

Cadair (approved):

changelog/7758.bugfix.2.rst
```diff
@@ -0,0 +1 @@
+Fix `sunpy.physics.differential_rotation.differential_rotate` to update the ``reference_date`` attribute of the input map instead of the ``date``.
```
ayshih, line 1:
```suggestion
Fix :func:`sunpy.physics.differential_rotation.differential_rotate` to update the ``reference_date`` attribute of the input map instead of the ``date``.
```

changelog/7758.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed incorrect ``reference_date`` for SDO/AIA and SDO/HMI data to use the ``T_OBS`` keyword instead of the ``DATE-OBS`` keyword.
```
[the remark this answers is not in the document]
ayshih: Because of the heavy use of sunpy for AIA imagery, it might be helpful to have a second sentence that says, "This correction is typically negligible: for AIA images, the location of SDO has been shifted up to a second in time, which corresponds to a shift in heliographic longitude by only 4 milliarcseconds."

sunpy/map/sources/soho.py
ayshih: I think what you did for `GONGSynopticMap` looks good.  Can you add that here as well? (a few lines down)

sunpy/map/sources/tests/test_mdi_source.py
```diff
@@ -40,6 +41,18 @@ def test_observatory(mdi):
...
+def test_reference_date(mdi):
+    assert mdi.reference_date.isot == "2010-10-15T23:00:11.000"
+
+
+def test_date(mdi):
+    assert mdi.date.isot == "2010-10-15T23:00:11.000"
+
+
+def test_private_date_setters(mdi):
+    _test_private_date_setters(mdi)
```
ayshih, line 53: Can you add another version of this test for `mdi_synoptic`?

ayshih (approved):
Some minor stuff, otherwise let's get this in!

sunpy/map/sources/tests/test_mdi_source.py
nabobalis: Will do.

nabobalis:
Ok I pushed one more commit, hopefully it's all good Albert!

nabobalis:
Online tests are numpy being downgraded to <2 for some reason and that is breaking the map rst file.

PR #7760, opened 2024-07-29
Adds `get_contours` and supports "method" key word a keyword argument to select between `contourpy` and `scikit-image`

Prtm2110:
## PR Description
Solves: #5874

`ContourPy` is giving same result as ski-images i.e. its working as intended


Prtm2110:
Sorry for all these commits, I certainly made some mistake here

pyproject.toml
```diff
@@ -69,6 +69,7 @@ jpeg2000 = [
   "lxml>=4.9.0,!=5.0.0",
 ]
 map = [
+  "contourpy>=1.2.0"
```
nabobalis, line 72: Does this need to be added?

sunpy/map/mapbase.py
```diff
@@ -2730,8 +2730,10 @@ def contour(self, level, **kwargs):
             # find_contours expects a scalar below
             level = level[0]
 
-        contours = measure.find_contours(self.data, level=level, **kwargs)
+        gen = contour_generator(z=self.data, **kwargs)
+        contours = gen.lines(level)
         contours = [self.wcs.array_index_to_world(c[:, 0], c[:, 1]) for c in contours]
+
```
nabobalis, line 2736:
```suggestion
```

pyproject.toml
Prtm2110: I guess not as I tried test ContourPy without ` "contourpy>=1.2.0"` and it worked fine

nabobalis:
Do we need to increase our MPL minimum version or are we ok?

sunpy/map/mapbase.py
```diff
@@ -2720,7 +2720,7 @@ def contour(self, level, **kwargs):
         --------
         skimage.measure.find_contours
```
nabobalis, line 2721: This reference needs to be changed and the one above as well.

Prtm2110:
> Do we need to increase our MPL minimum version or are we ok?

Nope

pyproject.toml
```diff
@@ -69,6 +69,7 @@ jpeg2000 = [
   "lxml>=4.9.0,!=5.0.0",
 ]
 map = [
+  "contourpy>=1.2.0",
```
nabobalis, line 72:
```suggestion
```
I thought this was a dependency of MPL?

sunpy/map/mapbase.py
```diff
@@ -2687,7 +2687,7 @@ def contour(self, level, **kwargs):
         """
         Returns coordinates of the contours for a given level value.
 
-        For details of the contouring algorithm see `skimage.measure.find_contours`.
+        For details of the contouring algorithm, see `contourpy.contour_generator`.
```
nabobalis, line 2690: The library will need to be added to the intersphinx dict in conf.py for this to work.

pyproject.toml
Prtm2110: Yes, understood now as mpl already has `ContourPy` as its dependency then we don't need to add it as well

Prtm2110:
> Do we need to increase our MPL minimum version or are we ok?

Yes [Here](https://github.com/matplotlib/matplotlib/releases/tag/v3.6.0) v3.6.0 is the minimum requirement

sunpy/map/mapbase.py
Prtm2110: I was unaware of that, thanks

pyproject.toml
ayshih: The code (now) does a direct import from `contourpy`, so it needs to be added as an explicit dependency

ayshih (commented):
- I commented above, but put back the explicit dependency on `contourpy`
- We probably don't need to change the minimum version of matplotlib
- I'd argue that this should be classified as a "breaking" change because we pass through additional keywords to the underlying contour-finding function, and I presume that the respective functions in `skimage` and `contourpy` accept different keywords

sunpy/map/tests/test_mapbase.py
```diff
@@ -1505,8 +1505,6 @@ def test_contour_inputs(simple_map):
     with pytest.raises(ValueError, match='Contour levels must be increasing'):
         simple_map.draw_contours([10, -10] * u.dimensionless_unscaled)
 
-    with pytest.raises(ValueError, match=r'The provided level \(1000.0\) is not smaller than the maximum data value \(8\)'):
-        simple_map.draw_contours(1000 * u.dimensionless_unscaled, fill=True)
```
nabobalis, line 1509: Why does this not error anymore?

nabobalis:
Can you check the code in draw_contours?

sunpy/map/tests/test_mapbase.py
Prtm2110: As the above code is part of `@pytest.mark.skipif(Version(matplotlib.__version__) < Version("3.6.0")` it doesn't run and because now min mpl is `v3.5.0` so no error.
I wonder why it is failing when I used v`3.6.0`

Prtm2110:
> draw_contours

Can you please elaborate

sunpy/map/tests/test_mapbase.py
nabobalis: Well the skip if needs to go and we need to increase our minimum version of MPL to 3.6.0 as it supports this feature now.

nabobalis:
> > check the code in draw_contours?
> 
> Can you please elaborate

honestly, I don't have more to say. We have a draw_contours method that needs to be looked over with the changes in MPL in 3.6.0 and countourpy in mind.

pyproject.toml
nabobalis, line 72:
```suggestion
  "contourpy>=1.0.0",
```
Do we need to pin a specific version for MPL?

changelog/7760.breaking.rst
```diff
@@ -0,0 +1 @@
+The `skimage` library has been replaced with `ContourPy` and creating contours
```
nabobalis, line 1:
```suggestion
Replaced the use of scikit-image with ContourPy within `sunpy.map.Map.contour`.
There is no change to returned contours but offers a faster way to contour. 
```

pyproject.toml
Prtm2110: As mentioned [here](https://github.com/sunpy/sunpy/pull/5923#discussion_r819417509) we can go with `>=v1.0.1`

nabobalis: Sounds good

sunpy/map/tests/test_mapbase.py
Prtm2110: Okay.

docs/conf.py
```diff
@@ -204,6 +204,7 @@
     "sunkit_image": ("https://docs.sunpy.org/projects/sunkit-image/en/stable/", None),
     "sunkit_instruments": ("https://docs.sunpy.org/projects/sunkit-instruments/en/stable/", None),
     "zeep": ("https://docs.python-zeep.org/en/stable/", None),
+    "contourpy": ("https://contourpy.readthedocs.io/en/stable/", None),
```
Prtm2110, line 207: Is this correct? the ref is still not working?

sunpy/map/tests/test_mapbase.py
```diff
@@ -1500,7 +1499,6 @@ def test_contour_units(simple_map):
         assert np.all(c1 == c2)
 
 
-@pytest.mark.skipif(Version(matplotlib.__version__) < Version("3.6.0"), reason="Fails on old MPL versions, the first with block raises a different error")
```
Prtm2110, line 1503: Why is this test which was being skipped before is failing now? Is it a problem in the test itself or the code is incorrect somewhere?

changelog/7760.breaking.rst
```diff
@@ -0,0 +1,2 @@
+Replaced the use of scikit-image with ContourPy within `sunpy.map.Map.contour`.
+There is no change to returned contours but offers a faster way to contour.
```
ayshih, line 2:
```suggestion
`sunpy.map.GenericMap.contour` now internally uses ContourPy instead of scikit-image for improved performance.
The API for the method has not changed, but as before, additional keyword arguments are passed through to the underlying function, so some such keyword arguments may no longer be supported.
```

ayshih: This also fixes the broken ref

sunpy/map/tests/test_mapbase.py
```diff
@@ -1500,7 +1499,6 @@ def test_contour_units(simple_map):
...
 def test_contour_inputs(simple_map):
     with pytest.raises(ValueError, match='Contour levels must be increasing'):
         simple_map.draw_contours([10, -10] * u.dimensionless_unscaled)
```
ayshih, line 1504: Having non-increasing contour levels apparently no longer emits an error, so this block can be deleted.

ayshih (changes requested):
Has someone confirmed that the returned contours are in fact identical with this change?  Or are the contours similar but not necessarily identical?

sunpy/map/tests/test_mapbase.py
nabobalis: We should add a test to see what non-increasing contour levels does to the output.

nabobalis:
> Has someone confirmed that the returned contours are in fact identical with this change? Or are the contours similar but not necessarily identical?

None of the figure tests changed? That is the best I got.

ayshih:
> None of the figure tests changed? That is the best I got.

The output of `contour()` is not used by `draw_contours()`, so figure tests are irrelevant

nabobalis:
Wait, why?!

ayshih:
`draw_contours()` calls `matplotlib.axes.Axes.contour()`, but I think it was challenging to pull the coordinates of a contour back out of matplotlib (hardly surprising).  I think that's why `contour()` used to call `skimage.measure.find_contours()` to provide contour coordinates to the user.

Now that `contour()` can directly call the same function that matplotlib is (presumably) using for `matplotlib.axes.Axes.contour()`, `contour()` and `draw_contour()` should be identical now.  My question is equivalent to asking whether `contour()` and `draw_contour()` were identical previously.

nabobalis:
So we should add a figure test doing the comparison of the two methods?

ayshih:
Well, there's not much value to a figure test comparing `skimage` and `contourpy`, and there should be no difference between `contour()` and `draw_contours()` after this PR.  I was just questioning the statement in the changelog, "There is no change to returned contours...", because no one has apparently checked (aside from the very simple unit test, which in fact does have a change in point order).

nabobalis:
Ah ok, we should change the changelog then!

sunpy/map/mapbase.py
```diff
@@ -2730,7 +2730,8 @@ def contour(self, level, **kwargs):
             # find_contours expects a scalar below
             level = level[0]
 
-        contours = measure.find_contours(self.data, level=level, **kwargs)
+        gen = contour_generator(z=self.data, **kwargs)
```
Cadair, line 2733: How much do we care about this method using the same algorithm as `draw_contour`? It seems that there are faster options in contourpy (like the threaded one) we could default to?

nabobalis: In theory now, they should be the same as MPL uses it.

Prtm2110:
> Has someone confirmed that the returned contours are in fact identical with this change? Or are the contours similar but not necessarily identical?

Sorry, They are not even similar because `contourpy.contour_generator` returns `M,N` format unlike `skimage.measure.find_contours` but after the commit I just made they the results are similar but not identical

Prtm2110:
Screenshot from 2024-08-02 12-01-53

nabobalis:
The other choice is to add a method keyword and allow users to select either package? 
If we want to avoid a breaking change.

docs/conf.py
nabobalis: What is not working?

sunpy/map/tests/test_mapbase.py
nabobalis: It was only skipped on oldest deps, it always ran on the CI. 

The test output will have changed and we need to work out why and adjust the test.

Prtm2110:
> The other choice is to add a method keyword and allow users to select either package? 
> If we want to avoid a breaking change.

Should I make these changes then?

nabobalis:
> > The other choice is to add a method keyword and allow users to select either package?
> > If we want to avoid a breaking change.
> 
> Should I make these changes then?

No, wait for the other reviewers to provide their view on this.

sunpy/map/mapbase.py
Cadair: yeah, but my point is that there is apparently a newer algorithm in contourpy not used by mpl (yet) which is faster and can be multi-threaded. https://contourpy.readthedocs.io/en/v1.2.1/user_guide/calculate/threads.html#threads

nabobalis: Lets use that.

sunpy/map/mapbase.py
```diff
@@ -2730,8 +2730,9 @@ def contour(self, level, **kwargs):
             # find_contours expects a scalar below
             level = level[0]
 
-        contours = measure.find_contours(self.data, level=level, **kwargs)
-        contours = [self.wcs.array_index_to_world(c[:, 0], c[:, 1]) for c in contours]
+        gen = contour_generator(z=self.data,name="threaded", **kwargs)
```
nabobalis, line 2733:
```suggestion
        gen = contour_generator(z=self.data, name="threaded", **kwargs)
```
I think we should instead check for name in kwargs. If a user passess in their own name, we want to check that it correctly is set.

Cadair, line 2733: We need to be intentional about enabling threaded, if we do this we are no longer using the *exact* same code path as mpl.

nabobalis: Would you prefer we add an explicit method keyword to choose?

Cadair: no, it's more that we need to decide how important to us algorithm compatibly is to us.

nabobalis: Well I mean, why not let users pick, between MPL, scikit-image and contourpy methods?

Cadair: I am less convinced by keeping skimage as an option, but for contourpy it's more that we need to pick the default.

nabobalis: Fair, personally I lean towards playing it safe, setting name to be the simple non-threaded approach and then if MPL change the default to the threaded one, we can do the same.

Prtm2110:
How can I go about updating figure tests and adding tests in general?

nabobalis:
> How can I go about updating figure tests and adding tests in general?

Do you mean updating the hashes of the figure tests or creating new ones?

For adding tests, the simplest thing to copy an existing one and then update to test a new item or a new objective.

There is some documentation here: https://docs.sunpy.org/en/latest/dev_guide/contents/tests.html

Prtm2110:
> Do you mean updating the hashes of the figure tests or creating new ones?
> 
Both, actually—updating the necessary hashes for this PR and creating new ones. I'm not very familiar with testing figures.
> For adding tests, the simplest thing to copy an existing one and then update to test a new item or a new objective.

The test I added passes locally but fails in the Git checks. Why might this be happening?

nabobalis:
> > Do you mean updating the hashes of the figure tests or creating new ones?
> 
> Both, actually—updating the necessary hashes for this PR and creating new ones. I'm not very familiar with testing figures.

The doc link should help.
Updating the figure hashes is simplest to copy the file from the circleci artifacts page. 

> > For adding tests, the simplest thing to copy an existing one and then update to test a new item or a new objective.
> 
> The test I added passes locally but fails in the Git checks. Why might this be happening?

Your local environment is not up to date is the most common issue.

changelog/7760.breaking.rst
```diff
@@ -0,0 +1,2 @@
+`sunpy.map.GenericMap.contour` now internally uses ContourPy instead of scikit-image for improved performance.
```
nabobalis, line 1: Can the text in this changelog be adjusted for Cadair's comment above?

sunpy/map/mapbase.py
```diff
@@ -2713,14 +2713,23 @@ def contour(self, level, **kwargs):
...
+        (717.17229147, -362.06880784), (717.27714042, -361.9631112 ),
+        (718.43620686, -359.56313541), (718.8672722 , -357.1614    ),
+        (719.58811599, -356.68119768), (721.29217122, -354.76448374),
+        (722.00110323, -352.46446792), (722.08933899, -352.36363319),
+        (722.00223989, -351.99536019), (721.67724425, -352.36263712),
+        (719.59798458, -352.60839064), (717.19243987, -353.75348121),
+        (715.8820808 , -354.75140718), (714.78652558, -355.05102034),
+        (712.68209174, -357.14645009), (712.68639008, -359.54923801),
+        (713.14112796, -361.95311455)]>
         ...
```
nabobalis, line 2726:
```suggestion
```

changelog/7760.breaking.rst
Prtm2110: Is it fine now?

changelog/7760.breaking.rst
```diff
@@ -0,0 +1,2 @@
+`sunpy.map.GenericMap.contour` now uses ContourPy instead of scikit-image for faster performance and consistency with `sunpy.map.GenericMap.draw_contour`, reducing reliance on scikit-image.
```
nabobalis, line 1:
```suggestion
`sunpy.map.GenericMap.contour` now uses ContourPy instead of scikit-image for faster performance and consistency with `sunpy.map.GenericMap.draw_contour`.
```
Is it actually faster? We can't say this if we don't verify it.

nabobalis, line 1: <unknown>:7: WARNING: py:obj reference target not found: sunpy.map.GenericMap.draw_contour [ref.obj]

Prtm2110: Alright, how about 'improved' instead of 'faster' as suggested earlier

nabobalis: Improved how?

Prtm2110: Its fixed now

Prtm2110: I don't know, I tried looking for evidence but didn't find any, so what should I update changelog to?

nabobalis: Either you drop the mention of performance or you benchmark it. That is our two choices.

Prtm2110: **Performance Test**
```
import cProfile

test_code = '''

import astropy.units as u
for x in range(25):
    import sunpy.map
    aiamap = sunpy.map.Map(AIA_193_IMAGE)
    contours = aiamap.contour(50000 * u.DN)
'''

if __name__ == "__main__":
    cProfile.run(test_code)
```
**Using Scikit-Image**
`8894305 function calls (18373215 primitive calls) in 62.470 seconds`
**Using ContourPy**
`17448686 function calls (16975009 primitive calls) in 58.905 seconds`

nabobalis: That is better. Improved it is.

nabobalis (approved):

changelog/7760.breaking.rst
```diff
@@ -0,0 +1,2 @@
+:meth:`sunpy.map.GenericMap.contour` now uses ContourPy instead of scikit-image for improved performance and consistency with :meth:`sunpy.map.GenericMap.draw_contours`.
```
ayshih, line 1: The changelog entry needs to warn that the returned contours will likely be different (in order, if not in values)

sunpy/map/mapbase.py
ayshih, line 2690:
```suggestion
        For details of the contouring algorithm, see :func:`contourpy.contour_generator`.
```

sunpy/map/mapbase.py
```diff
@@ -2696,7 +2696,7 @@ def contour(self, level, **kwargs):
             is not `None`, this must be a `~astropy.units.Quantity` with units
             equivalent to the map data units.
         kwargs :
-            Additional keyword arguments are passed to `skimage.measure.find_contours`.
+            Additional keyword arguments are passed to `contourpy.contour_generator`.
```
ayshih, line 2699:
```suggestion
            Additional keyword arguments are passed to :func:`contourpy.contour_generator`.
```

changelog/7760.breaking.rst
```diff
@@ -0,0 +1,3 @@
...
+The API for the method has not changed, but as before, additional keyword arguments are passed through to the underlying function, so some such keyword arguments may no longer be supported.
+The returned contours may also differ in order and values compared to the previous implementation.
```
nabobalis, line 3:
```suggestion
:meth:`sunpy.map.GenericMap.contour` now uses ContourPy instead of scikit-image for improved performance and consistency with :meth:`sunpy.map.GenericMap.draw_contours`.

This does mean that both that the returned contours may differ in order and values compared to the scikit-image version and as that additional keyword arguments are passed through to the underlying function, specific keyword arguments may no longer be supported with ContourPy.
```

Prtm2110:
Any Other Suggestions?

nabobalis:
> Any Other Suggestions?

Not right now, this needs a second reviewer to approve.

changelog/7760.breaking.rst
```diff
@@ -0,0 +1,3 @@
+:meth:`sunpy.map.GenericMap.contour` now uses ContourPy instead of scikit-image for improved performance and consistency with :meth:`sunpy.map.GenericMap.draw_contours`.
+
+This does mean that both that the returned contours may differ in order and values compared to the scikit-image version and as that additional keyword arguments are passed through to the underlying function, specific keyword arguments may no longer be supported with ContourPy.
```
wtbarnes, line 3: What does this mean? Is this saying that the result of the contour method won't be the same after switching to contourpy? If so, isn't this a breaking change?

wtbarnes (commented):
In terms of code changes, I'm fine with this, but I'm a bit concerned by the differences in results here. How big are these differences? Is it because the algorithm for finding the contours is different? After upgrading to 6.1 and plotting contours, is a user going to find significantly different contours when running the same code as they did with 6.0?

The changelog should make all of that clear.

ayshih:
Based on the little output that I've seen, it looks like the contours may be fundamentally unchanged, but simply can start/end at a different point along a closed contour.  This change will break any downstream code that depended on the exact sequencing of contour points.

Regardless, yes, this is a breaking change with no deprecation period.

wtbarnes:
In that case, this needs a keyword argument to select between contourpy and scikit-image, defaulting to the latter, and a deprecation warning.

wtbarnes (changes requested):
In that case, this needs a keyword argument to select between contourpy and scikit-image, defaulting to the latter, and a deprecation warning.

sunpy/map/mapbase.py
```diff
@@ -2708,20 +2708,23 @@ def plot(self, *, annotate=True, axes=None, title=True, autoalign=False,
 
         return ret
 
-    def contour(self, level, **kwargs):
+    def contour(self, level, method='skimage', **kwargs):
```
nabobalis, line 2711: We now need to raise a deprecation warning, saying that skiimage will be dropped in future and we will default to contourpy

sunpy/map/mapbase.py
```diff
@@ -2708,20 +2708,23 @@ def plot(self, *, annotate=True, axes=None, title=True, autoalign=False,
...
+        For details of the contouring algorithm, see :func:`contourpy.contour_generator` or :func:`contourpy.contour_generator`.
 
         Parameters
         ----------
         level : float, astropy.units.Quantity
             Value along which to find contours in the array. If the map unit attribute
             is not `None`, this must be a `~astropy.units.Quantity` with units
             equivalent to the map data units.
+        method : The `method` keyword determines which contouring method is used and should
+            be specified as either 'contourpy' or 'skimage'.
```
nabobalis, line 2724:
```suggestion
        method : {'contourpy', 'skimage'}
            Determines which contouring method is used and should
            be specified as either 'contourpy' or 'skimage'.
            Defaults to 'skimage'.
```

ayshih:
> In that case, this needs a keyword argument to select between contourpy and scikit-image, defaulting to the latter, and a deprecation warning. 

I disagree with the second and third parts.  I think we should set the default to contourpy (as a breaking change with no deprecation period) and we should keep scikit-image as an option indefinitely (as an import within the method so that it is not a hard dependency).  Deprecation warnings when changing default behavior are very annoying, and there's no reason or benefit to completely removing the usage of scikit-image.

This is analogous to what we did when changing the default for image rotation.

ayshih:
If deprecation is critical, I suggest defining a new method, e.g., `get_contour()`, and then deprecating `contour()`.

wtbarnes:
The spirit of the original issue that this PR addresses was to remove our dependency on scikit-image, hence my suggestion. What's the benefit of having two possible code paths for contouring? Does anyone really need that?

nabobalis:
> If deprecation is critical, I suggest defining a new method, e.g., `get_contour()`, and then deprecating `contour()`.

Oh I like this. Anyone else up for it?

ayshih:
> The spirit of the original issue that this PR addresses was to remove our dependency on scikit-image, hence my suggestion. What's the benefit of having two possible code paths for contouring? Does anyone really need that?

Once the default behavior of the contour method doesn't depend on scikit-image, that issue is resolved as far as I'm concerned.  Due to image rotation, scikit-image will "forever" be an optional dependency of sunpy.image, and hence of sunpy.map, so there's no harder dependency by keeping it as an option here.

Even if the default behaviors of the two contour implementations were identical in output, they have different optional keyword arguments, so it's conceivable to me that someone would still want to call the scikit-image version.

Prtm2110:
So how should I proceed?

nabobalis:
We will need to decide that at the next weekly meeting.

nabobalis:
From the community meeting today:

1) Revert the old method
2) Deprecate the old method in favour of the new method below.
3) Create a new method called "get_contours" that accepts a method keyword that allows users to switch methods (between scikit image and contourpy) but it defaults to "contourpy".
4) Update all the docs and examples where we call contours to use the new method.

Hopefully that is clear, if not do say and we can work on it.

Prtm2110:
I have moved all the unit tests from the `contour` method to the `get_contours` method, which means contour no longer has any unit tests. Is this acceptable, or should I maintain tests for contour as they were before?

If I understand correctly, this change is considered a feature update rather than a breaking change, since user code should not break after this release. Do I need to add two changelog entries: one for the feature update and another for the deprecation?

nabobalis:
> I have moved all the unit tests from the `contour` method to the `get_contours` method, which means contour no longer has any unit tests. Is this acceptable, or should I maintain tests for contour as they were before?

I think that is fine, we just need a unit test checking the deprecation warning for the old contour call. 

> If I understand correctly, this change is considered a feature update rather than a breaking change, since user code should not break after this release. Do I need to add two changelog entries: one for the feature update and another for the deprecation?

This is still a breaking change as we plan to remove the contour method. So as you say, we need two changelogs.

changelog/7760.deprecation.rst
```diff
@@ -0,0 +1 @@
+The :meth:`sunpy.map.GenericMap.contour` method, which still uses skimage, will now raise a deprecation warning. Please use :meth:`sunpy.map.GenericMap.get_contours`, which defaults to using ContourPy, instead.
```
nabobalis, line 1:
```suggestion
:meth:`sunpy.map.GenericMap.contour` method has now been deprecated in favor of :meth:`sunpy.map.GenericMap.get_contours`.
```

changelog/7760.feature.rst
```diff
@@ -0,0 +1,4 @@
+A new method, :meth:`sunpy.map.GenericMap.get_contours`, has been added for drawing contours.
+This method includes a "method" keyword, which defaults to ContourPy for improved performance and consistency with :meth:`sunpy.map.GenericMap.draw_contours`. The skimage option is also supported.
+
+Note that ContourPy may produce different contour results and may not support all scikit-image keyword arguments.
```
nabobalis, line 4:
```suggestion
A new method :meth:`sunpy.map.GenericMap.get_contours`, has been added for getting contours.
By default it now uses ContourPy for improved performance and consistency with :meth:`sunpy.map.GenericMap.draw_contours`. By using the "method" keyword, one can change to scikit-image

Note that ContourPy may produce different contour results and may not support all scikit-image keyword arguments.
```

sunpy/map/mapbase.py
```diff
@@ -2708,6 +2709,7 @@ def plot(self, *, annotate=True, axes=None, title=True, autoalign=False,
 
         return ret
 
+    @deprecated(since="6.0", message="The 'contour' method is deprecated and will be removed in a future release. Please use 'get_contours' instead.", alternative="sunpy.map.GenericMap.get_contours")
```
nabobalis, line 2712:
```suggestion
    @deprecated(since="6.1", message="The 'contour' method is deprecated and will be removed in a future release. Please use 'get_contours'.", alternative="sunpy.map.GenericMap.get_contours")
```

sunpy/map/tests/test_mapbase.py
```diff
@@ -1452,13 +1451,33 @@ def test_submap_inputs(generic_map2, coords):
         smap = generic_map2.submap(*args, **kwargs)
         assert u.allclose(smap.dimensions, (3, 3) * u.pix)
 
+def test_contour_deprecation_warning(simple_map):
+    with pytest.warns(SunpyDeprecationWarning, match="The 'contour' method is deprecated and will be removed in a future release. Please use 'get_contours' instead."):
+        simple_map.get_contours([1.5, 2.5])
```
nabobalis, line 1457:
```suggestion


```

sunpy/map/tests/test_mapbase.py
```diff
@@ -1452,13 +1451,33 @@ def test_submap_inputs(generic_map2, coords):
...
```
nabobalis, line 1456:
```suggestion
        simple_map.contour([1.5, 2.5])
```

Prtm2110: I get I have made a mistake here, but I usually do check the tests. I'm not sure why they weren't failing locally- my repository is up-to-date.

sunpy/map/mapbase.py
```diff
@@ -2765,6 +2767,78 @@ def contour(self, level, **kwargs):
...
+        Parameters
+        ----------
+        level : float, astropy.units.Quantity
+            Value along which to find contours in the array. If the map unit attribute
+            is not `None`, this must be a `~astropy.units.Quantity` with units
+            equivalent to the map data units.
+        method : {'contourpy', 'skimage'}
+            Determines which contouring method is used and should
+            be specified as either 'contourpy' or 'skimage'.
+            Defaults to 'skimage'.
```
nabobalis, line 2785:
```suggestion
            Defaults to 'contourpy'.
```

sunpy/map/mapbase.py
```diff
@@ -2765,6 +2767,78 @@ def contour(self, level, **kwargs):
...
+        kwargs :
+            Additional keyword arguments passed to either :func:`contourpy.contour_generator`
+            or :func:`skimage.measure.find_contours`, depending on the value of the `method` argument.
```
nabobalis, line 2788:
```suggestion
            or :func:`skimage.measure.find_contours`, depending on the value of the ``method`` argument.
```

sunpy/map/mapbase.py
```diff
@@ -2765,6 +2767,78 @@ def contour(self, level, **kwargs):
...
+        --------
+        :func:`contourpy.contour_generator` and :func:`skimage.measure.find_contours`
+        """
+        level = self._process_levels_arg(level)
+        if level.size != 1:
+            raise ValueError("level must be a single scalar value")
+        else:
+            # _process_levels_arg converts level to a 1D array, but
+            # find_contours expects a scalar below
+            level = level[0]
```
nabobalis, line 2826: Is this a problem for contourpy or does it need the same thing to occur?

sunpy/map/tests/test_mapbase.py
nabobalis: It isn't about making a mistake, that's normal. There are plenty of PRs of mine here that were and are riddled with mistakes. 

It's just good practice to check if changes do not break the relevant tests.

sunpy/map/mapbase.py
Prtm2110: No, as per the contourpy docs level (lines) can be float only

sunpy/map/mapbase.py
```diff
@@ -2765,6 +2767,78 @@ def contour(self, level, **kwargs):
...
+            Coordinates of each contour.
+
+        Examples
+        --------
+        >>> import astropy.units as u
+        >>> import sunpy.map
+        >>> import sunpy.data.sample  # doctest: +REMOTE_DATA
+        >>> aia = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)  # doctest: +REMOTE_DATA
+        >>> contours = aia.get_contours(50000 * u.DN, method='contourpy')  # doctest: +REMOTE_DATA
+        >>> print(contours[0])  # doctest: +REMOTE_DATA
```
nabobalis, line 2802: Do we need the print here?

sunpy/map/mapbase.py
nabobalis: So can we drop this else statement or not?

sunpy/map/mapbase.py
```diff
@@ -2708,6 +2709,7 @@ def plot(self, *, annotate=True, axes=None, title=True, autoalign=False,
 
         return ret
 
+    @deprecated(since="6.1", message="The 'contour' method is deprecated and will be removed in a future release. Please use 'get_contours'.", alternative="sunpy.map.GenericMap.get_contours")
     def contour(self, level, **kwargs):
         """
         Returns coordinates of the contours for a given level value.
```
nabobalis, line 2715: I can't leave a comment at the right location now but since we added the deprecated, the example now fails the CI tests. 

I think we can add a IGNORE_WANRINGS flag to doctest to ignore the warning. Hopefully that will work.

sunpy/map/mapbase.py
Prtm2110: No we cannot remove it, several unit tests are failing after removing it.

sunpy/map/mapbase.py
Prtm2110: I guess not, should I remove it to check?

nabobalis: Can you remove it?

sunpy/map/mapbase.py
```diff
@@ -2765,6 +2767,78 @@ def contour(self, level, **kwargs):
...
+            Coordinates of each contour.
+
+        Examples
+        --------
+        >>> import astropy.units as u
+        >>> import sunpy.map
+        >>> import sunpy.data.sample  # doctest: +REMOTE_DATA
+        >>> aia = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)  # doctest: +REMOTE_DATA
+        >>> contours = aia.get_contours(50000 * u.DN, method='contourpy')  # doctest: +REMOTE_DATA
+        >>> print(contours[0])  # doctest: +IGNORE_WARNINGS
```
nabobalis, line 2802: This is the wrong place to make this change and it still needs the REMOTE_DATA

Prtm2110: Sorry, where can I add +IGNORE_WARNINGS then?

nabobalis: It gets added to the example for the old method which is raising the deprecation warning.

Prtm2110: Sorry, yeah it should be for contour

sunpy/map/mapbase.py
```diff
@@ -2734,8 +2736,8 @@ def contour(self, level, **kwargs):
         >>> import sunpy.map
         >>> import sunpy.data.sample  # doctest: +REMOTE_DATA
         >>> aia = sunpy.map.Map(sunpy.data.sample.AIA_171_IMAGE)  # doctest: +REMOTE_DATA
-        >>> contours = aia.contour(50000 * u.DN)  # doctest: +REMOTE_DATA
-        >>> print(contours[0])  # doctest: +REMOTE_DATA
+        >>> contours = aia.contour(50000 * u.DN)  # doctest: +REMOTE_DATA +IGNORE_WARNINGS
+        >>> contours[0]  # doctest: +REMOTE_DATA +IGNORE_WARNINGS
```
nabobalis, line 2740: This shouldn't need the warnings flag but its worth checking.

sunpy/map/mapbase.py
```diff
@@ -2765,6 +2767,78 @@ def contour(self, level, **kwargs):
...
+        if level.size != 1:
+            raise ValueError("level must be a single scalar value")
+        else:
+            # _process_levels_arg converts level to a 1D array, but
+            # find_contours expects a scalar below
+            level = level[0]
+
+        if method == 'contourpy':
+            from contourpy import contour_generator
+            gen = contour_generator(z=self.data, **kwargs)
```
nabobalis, line 2830:
```suggestion
            from contourpy import contour_generator

            gen = contour_generator(z=self.data, **kwargs)
```

sunpy/map/mapbase.py
```diff
@@ -2765,6 +2767,78 @@ def contour(self, level, **kwargs):
...
+            contours = gen.lines(level)
+            contours = [self.wcs.array_index_to_world(c[:, 1], c[:, 0]) for c in contours]
+        elif method == 'skimage':
+            from skimage import measure
+            contours = measure.find_contours(self.data, level=level, **kwargs)
```
nabobalis, line 2835:
```suggestion
            from skimage import measure

            contours = measure.find_contours(self.data, level=level, **kwargs)
```

Prtm2110:
The failed test is unrelated.

changelog/7760.feature.rst
```diff
@@ -0,0 +1,4 @@
+A new method :meth:`sunpy.map.GenericMap.get_contours`, has been added for getting contours.
+By default it now uses ContourPy for improved performance and consistency with :meth:`sunpy.map.GenericMap.draw_contours`. By using the "method" keyword, one can change to scikit-image
```
ayshih, line 2:
```suggestion
Added :meth:`~sunpy.map.GenericMap.get_contours` for getting contours from a `~sunpy.map.Map`.
By default the method uses ContourPy for performance reasons and consistency with :meth:`sunpy.map.GenericMap.draw_contours`.
One can alternatively specify that the method use scikit-image.
```

sunpy/map/mapbase.py
```diff
@@ -2765,6 +2767,80 @@ def contour(self, level, **kwargs):
...
+             (722.00110323, -352.46446792), (722.08933899, -352.36363319),
+             (722.00223989, -351.99536019), (721.67724425, -352.36263712),
+             (719.59798458, -352.60839064), (717.19243987, -353.75348121),
+             (715.8820808 , -354.75140718), (714.78652558, -355.05102034),
+             (712.68209174, -357.14645009), (712.68639008, -359.54923801),
+             (713.14112796, -361.95311455)]>
+
+        See Also
+        --------
+        :func:`contourpy.contour_generator` and :func:`skimage.measure.find_contours`
```
ayshih, line 2818:
```suggestion
        :func:`contourpy.contour_generator`
        :func:`skimage.measure.find_contours`
```

sunpy/map/mapbase.py
```diff
@@ -2765,6 +2767,80 @@ def contour(self, level, **kwargs):
...
 
+    def get_contours(self, level, method='contourpy', **kwargs):
+        """
+        Returns coordinates of the contours for a given level value.
+
+        For details of the contouring algorithm, see :func:`contourpy.contour_generator` or :func:`contourpy.contour_generator`.
+
+        Parameters
+        ----------
+        level : float, astropy.units.Quantity
```
ayshih, line 2778:
```suggestion
        level : float, `~astropy.units.Quantity`
```

ayshih (changes requested):
Just a few more changes!

sunpy/map/mapbase.py
```diff
@@ -2708,6 +2709,7 @@ def plot(self, *, annotate=True, axes=None, title=True, autoalign=False,
 
         return ret
 
+    @deprecated(since="6.1", message="The 'contour' method is deprecated and will be removed in a future release. Please use 'get_contours'.", alternative="sunpy.map.GenericMap.get_contours")
```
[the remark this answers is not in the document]
Prtm2110: By 'automatic desperation message,' do you mean a warning? It has been raised.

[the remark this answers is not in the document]
ayshih: The `deprecated` decorator automatically creates a deprecation message if `message` is not specified, and that automatic message should include the `alternative` if specified.  You have manually specified a message string, which may not be necessary and may not be uniform with other deprecation warnings.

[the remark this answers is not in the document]
Prtm2110: Alright understood, now I have removed the manual message and automatic deprecation message is working correctly.

ayshih (approved):
Looks good to me!  Thanks for your patience!

Note to merger: definitely squash

nabobalis (approved):

nabobalis:
Thanks again @Prtm2110 for the PR

Thank you @ayshih and @wtbarnes for your reviews.

PR #7768, opened 2024-08-03
Ignore dep warning from asdf?

nabobalis:


pytest.ini
```diff
@@ -87,3 +87,5 @@ filterwarnings =
     ignore:This download has been started in a thread which is not the main thread:UserWarning
     # Raised by oldesdeps job for scikit-image
     ignore:Starting with ImageIO v3 the behavior of this function will switch.*:DeprecationWarning
+    # Raised by CI now I think for asdf / jsonschema?
```
braingram, line 90:
```suggestion
    # Warning by attrs used indirectly through asdf. Can be removed when min asdf is 3.4.0
```

braingram (approved):
asdf 3.4.0 is out with a fix to prevent this new deprecation warning (which is coming from attrs: https://github.com/python-attrs/attrs/pull/1323).

However I think it will still appear in the oldest deps so filtering the warning marks sense to me.

nabobalis:
> asdf 3.4.0 is out with a fix to prevent this new deprecation warning (which is coming from attrs: [python-attrs/attrs#1323](https://github.com/python-attrs/attrs/pull/1323)).
> 
> However I think it will still appear in the oldest deps so filtering the warning marks sense to me.

Thanks!

wtbarnes:
That windows fail looks real but unrelated to this PR.

wtbarnes (approved):

Cadair (approved):

Issue #7780, opened 2024-08-17
Wheel installs toplevel "docs", "examples", "licenses" into lib*/python3.x/site-packages/


bnavigator:
### Describe the bug

The current wheels or when building from source, there are toplevel directories  "docs", "examples", "licenses" in the environment's site-packages. This potentially conflicts with other packages exhibiting the same unclean behavior.

### To Reproduce

```python
>>> from importlib.metadata import files
>>> set(f.parts[0] for f in files('sunpy'))
{'docs', 'sunpy', 'examples', 'sunpy-6.0.1.dist-info', 'licenses'}
```

### System Details

sunpy Installation Information
==============================

General
#######
OS:  (, Linux 6.10.3-1-default)
Arch: 64bit, (x86_64)
sunpy: 6.0.1
Installation path: /home/abuild/rpmbuild/BUILDROOT/python-sunpy-6.0.1-0.x86_64/usr/lib64/python3.11/site-packages/sunpy-6.0.1.dist-info

Required Dependencies
#####################
astropy: 6.1.2
numpy: 2.0.0
packaging: 24.1
parfive: 2.1.0
pyerfa: 2.0.1.4
requests: 2.32.3

Optional Dependencies
#####################
sunpy: 6.0.1

Install packages
############
pip: 24.0
setuptools: 70.1.1

### Installation method

pip, from source (.tar.gz)

bnavigator:
Here is what I use for the rpmbuild. You might want to adjust for general application:

```diff
diff -ur sunpy-6.0.1.orig/pyproject.toml sunpy-6.0.1/pyproject.toml
--- sunpy-6.0.1.orig/pyproject.toml	2024-08-16 02:26:49.000000000 +0200
+++ sunpy-6.0.1/pyproject.toml	2024-08-17 21:52:29.312141037 +0200
@@ -149,8 +149,12 @@
 license-files = ["LICENSE.rst"]
 
 [tool.setuptools.packages.find]
+include = ["sunpy*"]
 exclude = ["sunpy._dev"]
 
+[tool.setuptools.exclude-package-data]
+"*" = ["*.c", "*.h"]
+
 [tool.setuptools_scm]
 write_to = "sunpy/_version.py"
 
```

nabobalis:
Thanks for the report @bnavigator, would adding those folders to the mainifest file also work?

bnavigator:
[...]

PR #7781, opened 2024-08-18
provide the code only for a install

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/7780 

TODO:

- [ ] Fix the fail on oldestdeps


Cadair:
Why not exclude the other things in the manifest? Also this should be done in the template.

nabobalis:
We can dron the entire manifest this way. Seems like a neater solution to me? Especially for the template.

nabobalis:
Moved this upstream to https://github.com/sunpy/package-template/pull/111

PR #7791, opened 2024-08-31
Added area-based contouring to `GenericMap.contour()`

### LINKED ISSUE #7420: Area-based contours
### Describe the feature

For X-rays (and radio I think?) we like to put contours on images based on the "area contained." The current [`GenericMap.contour`](https://docs.sunpy.org/en/latest/generated/api/sunpy.map.GenericMap.html#sunpy.map.GenericMap.contour) method, when given percentiles, puts those contours relative to [the max value in the map](https://github.com/sunpy/sunpy/blob/090641f28b96507b78d7e64978f7c41d1c4c717b/sunpy/map/mapbase.py#L2307-L2308).

If we want to do it by area we have to do a little bit of shenanigans with sorted arrays, like [in this gist](https://gist.github.com/settwi/5d3f34b79843df00c2058ec1d49da2ea). We could extend the example to N-dimensions, too.

This would be a pretty small addition to the contour code but probably make people happy!

### Proposed solution

See the request! There is an example. Up for debate if we want N-dimension compatibility and how to add units in logically.

Mubin17:
## PR Description

Fixes #7420
Adds a new area parameter to the `get_contour()` method, allowing users to find contours based on the fraction of the total area enclosed. This is particularly useful for X-ray and radio maps where area containment is a relevant metric.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -2733,6 +2733,12 @@ class GenericMap(NDData):
             Value along which to find contours in the array. If the map unit attribute
             is not `None`, this must be a `~astropy.units.Quantity` with units
             equivalent to the map data units.
+        area : bool, optional
+            If `True`, the contour level is interpreted as the fraction of
+            the total area of the map data. The contour is drawn at the value
+            that corresponds to the specified fraction of area. If `False`,
+            the contour is drawn at the specified level value.
+            Defaults to `False`.
         kwargs :
             Additional keyword arguments are passed to `skimage.measure.find_contours`.
 
@@ -2760,12 +2766,20 @@ class GenericMap(NDData):
          (719.58811599, -356.68119768), (721.29217122, -354.76448374),
          (719.59798458, -352.60839064)]>
 
+        >>> contours = aia.contour(0.99*u.DN, area=True)  # doctest: +REMOTE_DATA
+        >>> print(contours[0])  # doctest: +REMOTE_DATA
+        <SkyCoord (Helioprojective: obstime=2011-06-07T06:33:02.880, rsun=696000.0 km, observer=<HeliographicStonyhurst Coordinate (obstime=2011-06-07T06:33:02.880, rsun=696000.0 km): (lon, lat, radius) in (deg, deg, m)
+        (-0.00406429, 0.04787238, 1.51846026e+11)>): (Tx, Ty) in arcsec
+        [(100.16572879, -1143.25308903), ( 99.68139649, -1143.76979777),
+         (100.159315  , -1145.90002049), (100.68236775, -1143.77222046),
+         (100.16572879, -1143.25308903)]>
+    
         See Also
         --------
         skimage.measure.find_contours
         """
         from skimage import measure
-
+        
         level = self._process_levels_arg(level)
         if level.size != 1:
             raise ValueError("level must be a single scalar value")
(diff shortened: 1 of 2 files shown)


nabobalis:
Thanks for the PR @Mubin17, however a change like this needs to be considered as we plan to replace this method in https://github.com/sunpy/sunpy/pull/7760.

I think there is a discussion to be had on how to handle this.

sunpy/map/mapbase.py
```diff
@@ -2747,19 +2753,42 @@ def contour(self, level, **kwargs):
...
             raise ValueError("level must be a single scalar value")
         else:
             # _process_levels_arg converts level to a 1D array, but
             # find_contours expects a scalar below
             level = level[0]
+        if area:
+            if not 0 <= level <= 1:
+                # print(level)
+                raise ValueError('level must be between 0 and 1 when area is set to True.')
+            # Sort pixels by value, then find the value that corresponds to the
```
nabobalis, line 2781: Author attribution should be given here since I think this is basically the gist code from the original issue?

nabobalis:
So now https://github.com/sunpy/sunpy/pull/7760 is merged in, I think we add this to the new get_contour method.

THe question I have is how should a user switch to the area-based method? Are we going to add a keyword or use the input value to the method?

Mubin17:
@nabobalis I've updated the `get_contours()` method to include an `area` flag, so users can switch to the area-based method by setting area=True.

sunpy/map/mapbase.py
```diff
@@ -2778,7 +2778,7 @@ def contour(self, level, **kwargs):
         contours = [self.wcs.array_index_to_world(c[:, 0], c[:, 1]) for c in contours]
         return contours
 
-    def get_contours(self, level, method='contourpy', **kwargs):
+    def get_contours(self, level, method='contourpy', area=False, **kwargs):
```
nabobalis, line 2781: I am not really pro adding another keyword but I don't have a better idea.

Hopefully the other maintainers will better ideas.

sunpy/map/mapbase.py
```diff
@@ -2812,17 +2818,32 @@ def get_contours(self, level, method='contourpy', **kwargs):
...
+        (719.58811599, -356.68119768), (721.29217122, -354.76448374),
+        (722.00110323, -352.46446792), (722.08933899, -352.36363319),
+        (722.00223989, -351.99536019), (721.67724425, -352.36263712),
+        (719.59798458, -352.60839064), (717.19243987, -353.75348121),
+        (715.8820808 , -354.75140718), (714.78652558, -355.05102034),
+        (712.68209174, -357.14645009), (712.68639008, -359.54923801),
+        (713.14112796, -361.95311455)]>
+        ...
+
+        >>> contours = aia.contour(0.9*u.DN, method='contourpy' area=True)  # doctest: +REMOTE_DATA
```
nabobalis, line 2834:
```suggestion
        >>> contours = aia.contour(0.9*u.DN, method='contourpy', area=True)  # doctest: +REMOTE_DATA
```

sunpy/map/mapbase.py
```diff
@@ -2837,6 +2858,21 @@ def get_contours(self, level, method='contourpy', **kwargs):
             # find_contours expects a scalar below
             level = level[0]
 
+        if area:
+            if not 0 <= level <= 1:
```
nabobalis, line 2862: Why does this need to be here for area?

ayshih:
I think a Boolean keyword makes sense, although I think calling it "area" is potentially confusing.


I'll note that this implementation assumes that all pixels have the same area.  It'd be okay to defer fixing that to a future PR, but then this assumption should be stated in the docstring.

nabobalis:
Thanks for the PR @Mubin17, I will close this for now, if you want to work on it again, please reopen it!

PR #7792, opened 2024-09-01
devdeps and linkcheck tweaks

nabobalis:
Should fix the failing cronjobs.

Need to check if normal tests pass with these changes.


Cadair:
oldestdep fails are real because of numpy repr changes :cry:

nabobalis:
> oldestdep fails are real because of numpy repr changes 😢

Hopefully it is fixed now.

Cadair (approved):

PR #6835, opened 2023-03-11
Adds the gallery example for `mplcairo` plotting 

Telomelonia:
## PR Description
This example will go through how to blend two plots from SunPy using `mplcairo`.

Fixes #6517 

## TODO
- [x] Create an example python file in plotting section for this tutorial
- [x] Check if this is coming in sphinx-gallery docs or not. 
- [x] Changelog.
## Issue
- [x] If the mplcairo is not in local backend then its not working (try something by which this is not required)

PR #7800, opened 2024-09-11
Reworked the blending gallery example so that mplcairo is not used

ayshih:
We have `mplcairo` as a docs dependency because of a single gallery example (#6835).  Unfortunately, it's been over two years since the last release of `mplcairo`, and there are no wheels for Python 3.11+.

This PR reworks the blending gallery example so that `mplcairo` is no longer used.  The blending operation has to be done "manually", but it's not too bad.  Accordingly, this PR also removes `mplcairo` as a dependency.


examples/plotting/screen_blend_mode.py
```diff
@@ -0,0 +1,84 @@
...
+======================
+Blending maps together
+======================
+
+This example shows how to blend two maps.
+
+``matplotlib`` by itself provides only alpha-based transparency for
+superimposing one image onto another, which can be restrictive when trying to
+create visually appealing composite images.  For better results, one can
+use math to perform standard image-processing
```
Cadair, line 11: by the power of "math" :superhero:

Cadair, line 11:
```suggestion
use standard image-processing
```

?

Cadair (commented):
lgtm

examples/plotting/screen_blend_mode.py
ayshih: Now a bit reworded

Cadair (approved):

alasdairwilson (approved):

PR #1736, opened 2016-04-19
Add the Helioprojective Radial Frame to coordinates

Cadair:
This is an addition to the coordinates package for the Helioprojective-Radial frame.

TODO:

- [x] Documentation. Specifcally the fact that the center of the disk is (0, -90) in this frame.

ping @LukeBarnard

PR #7803, opened 2024-09-13
Support for Helioprojective Radial

### LINKED ISSUE #5561: Add Helioprojective Radial frame to coordinates
Hello, 

Please consider adding Helioprojective Radial to the available solar coordinates. 

Best regards

ayshih:
This PR at last implements support for Helioprojective Radial (closes #5561).  (See #1736 for a past attempt.)  The tricky aspect is negotiating between the natural component of impact angle, which is 0 degrees at disk center, and the FITS-WCS component of declination, which is –90 degrees at disk center.  The PR adds:

- A `HelioprojectiveRadial` coordinate frame
- A `make_hpr_header()` header helper
- A `show_hpr_impact_angle()` ~context manager~ function to modify a WCSAxes axis to show impact angle

To do:

- [x] ASDF stuff
- [x] Transformation tests
- [x] Other tests
- [x] Docstring examples
- [x] What's new
- [x] Gallery example


ayshih:
The second image in the gallery example is supposed to be this:
Figure_2
and it renders that way when run from the terminal, but not when rendered by sphinx-gallery.  I might need to bug-fix sphinx-gallery...

ayshih:
> I might need to bug-fix sphinx-gallery...

Okay, I can't really fix this.  The figure rendering in sphinx-gallery happens after the code block is complete – the `plt.show()` command is essentially just decorative – and there's no way to keep the context alive beyond the end of the code block.  I was confused why we didn't encounter this issue previously with the `propagate_with_solar_surface()` context manager, but it seems that our gallery examples use it only for a reprojection call, and not for stuff like distorting grid lines during figure rendering.

Anyway, in this case, the context-manager approach was unnecessarily cute anyway, so I switched it to a function.

sunpy/visualization/visualization.py
```diff
@@ -45,15 +48,51 @@ def axis_labels_from_ctype(ctype, unit):
...
               'SOLY': 'Heliocentric Y'}
 
     label = labels.get(ctype_short, f"{ctype}")
     if unit is not None:
         label += f' [{unit}]'
 
     return label
+
+
+def show_hpr_impact_angle(declination_axis):
```
Cadair, line 67: I have got to have a think about how to patch astropy so we don't have to use the private API here.

ayshih: That's why I was interested in passing a matplotlib formatter into `set_major_formatter()`, but that turned into a whole mess because of how WCSAxes does things.  Even if some change is made to astropy, in the near term we still ought to provide users with an approach that works on older versions of astropy.

ayshih: Yup, definitely

sunpy/visualization/visualization.py
```diff
@@ -45,15 +48,51 @@ def axis_labels_from_ctype(ctype, unit):
...
+    .. minigallery:: sunpy.visualization.show_hpr_impact_angle
+    """
+    if not isinstance(declination_axis, CoordinateHelper):
+        raise TypeError("The input should be one of the WCSAxes coordinate axes.")
+
+    old_format_method = declination_axis._formatter_locator.formatter
+
+    def new_format_method(values, spacing, format="auto"):
+        return old_format_method(values + 90*u.deg, spacing, format=format)
+    declination_axis._formatter_locator.formatter = new_format_method
```
Cadair, line 97: ok, so using https://github.com/astropy/astropy/pull/17020

I could rewrite this as:

```python
    old_format_method = declination_axis.formatter

    def new_format_method(values, spacing, format="auto"):
        return old_format_method(values + 90*u.deg, spacing, format=format)

    declination_axis.set_major_formatter(new_format_method)
```

which does the job.

sunpy/visualization/visualization.py
```diff
@@ -45,15 +48,51 @@ def axis_labels_from_ctype(ctype, unit):
...
+    sunpy.coordinates.HelioprojectiveRadial
+
+    Examples
+    --------
+    .. minigallery:: sunpy.visualization.show_hpr_impact_angle
+    """
+    if not isinstance(declination_axis, CoordinateHelper):
+        raise TypeError("The input should be one of the WCSAxes coordinate axes.")
+
+    old_format_method = declination_axis._formatter_locator.formatter
```
Cadair, line 93: Even with current astropy we can remove one private API access here.
```suggestion
    old_format_method = declination_axis.formatter
```

sunpy/visualization/visualization.py
ayshih: We would need to check the `astropy` version and still use the private API for older versions?

Cadair: yeah or just migrate when we hit >7.0?

nabobalis (approved):
Great job Albert. This is why you Lord of Cord

Cadair (changes requested):
Looks great, just a few little things and asdf being asdf.

sunpy/io/special/asdf/resources/schemas/helioprojectiveradial-1.0.0.yaml
```diff
@@ -0,0 +1,60 @@
...
+allOf:
+  - $ref: "http://astropy.org/schemas/astropy/coordinates/frames/baseframe-1.0.0"
+  - properties:
+      frame_attributes:
+        type: object
+        properties:
+          observer:
+            anyOf:
+              - tag: "tag:sunpy.org:sunpy/coordinates/frames/heliographic_stonyhurst-1.0.0"
+              - tag: "tag:sunpy.org:sunpy/coordinates/frames/heliographic_stonyhurst-1.1.0"
```
braingram, line 50:
```suggestion
              - tag: "tag:sunpy.org:sunpy/coordinates/frames/heliographic_stonyhurst-1.*"
```
Since this is a new schema there's no harm in using wildcards here.

sunpy/io/special/asdf/resources/schemas/helioprojectiveradial-1.0.0.yaml
```diff
@@ -0,0 +1,60 @@
...
+              - type:
+                  string
+          rsun:
+            tag: "tag:stsci.edu:asdf/unit/quantity-1.1.0"
```
braingram, line 54:
```suggestion
            tag: "tag:stsci.edu:asdf/unit/quantity-1.*"
```

sunpy/io/special/asdf/resources/schemas/helioprojectiveradial-1.0.0.yaml
```diff
@@ -0,0 +1,60 @@
...
+          obstime:
+            tag: "tag:stsci.edu:asdf/time/time-1.1.0"
```
braingram, line 56:
```suggestion
            tag: "tag:stsci.edu:asdf/time/time-1.*"
```

braingram (commented):
I only reviewed the asdf part but those updates look good to me. Thanks! Let me know if looking at the other parts would be helpful.

sunpy/map/header_helper.py
```diff
@@ -391,3 +391,97 @@ def make_heliographic_header(date, observer_coordinate, shape, *, frame, project
...
+        frame="helioprojectiveradial",
+        obstime=observer_coordinate.obstime,
+        observer=observer_coordinate,
+        rsun=getattr(observer_coordinate, "rsun", None),
+    )
+
+    # Calculate the correct location for this reference coord, which will likely be far off the data array
+    reference_pixel = [
+        shape[1] / 2. - 0.5,
+        ((90*u.deg - theta_min) / theta_binsize).decompose().value - 0.5
```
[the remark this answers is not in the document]
ayshih: Ha ha, yeah, silly of me

sunpy/coordinates/wcs_utils.py
```diff
@@ -179,6 +182,10 @@ def solar_frame_to_wcs_mapping(frame, projection='TAN'):
             xcoord = 'HPLN' + '-' + projection
             ycoord = 'HPLT' + '-' + projection
             wcs.wcs.cunit = ['arcsec', 'arcsec']
+        elif isinstance(frame, HelioprojectiveRadial):
+            xcoord = 'HRLN' + '-' + projection
+            ycoord = 'HRLT' + '-' + projection
```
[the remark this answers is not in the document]
ayshih: I was simply mimicking the others, of course, so I've now changed them all to f-strings.

sunpy/coordinates/tests/test_frames.py
```diff
@@ -269,6 +270,60 @@ def test_hpc_is_visible_tolerance():
...
+    assert isinstance(hpr1._data, SphericalRepresentation)
+
+    assert hpr1.psi.unit is u.deg
+    assert hpr1.delta.unit is u.deg
+    assert hpr1.distance.unit is u.Mm
+    assert_quantity_allclose(hpr1.psi, 0*u.deg)
+    assert_quantity_allclose(hpr1.delta, 0*u.deg)
+    assert_quantity_allclose(hpr1.distance, 1*u.Mm)
+
+    assert_quantity_allclose(hpr1.theta, 90*u.deg)
```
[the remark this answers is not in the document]
ayshih: Added

Cadair (approved):

svank:
I'm very excited to see this! Just took it for a spin with PUNCH test data and it's working great. Thanks, @ayshih and all!

PR #7812, opened 2024-09-26
Add Angstrom as default unit for HMI wavelength

wtbarnes:
There does not seem to be any metadata in the HMI headers to indicate the unit of the wavelength. If it can't be found, it now will default to Angstrom.

Annoyingly, one of our test HMI files has a blank WAVEUNIT key, but I suspect that's just because it was processed with sunpy and then saved again. I've used that one as a test that it does in fact use WAVEUNIT if it is present.


nabobalis:
If you suspect that about the header, should we not replace it?

sunpy/map/sources/sdo.py
```diff
@@ -130,6 +130,16 @@ def __init__(self, data, header, **kwargs):
                 self._set_symmetric_vmin_vmax()
         self._nickname = self.detector
 
+    @property
+    def waveunit(self):
+        """
+        The `~astropy.units.Unit` of the wavelength of this observation.
+
+        Most HMI files seem to not have a parseable WAVEUNIT key so if it cannot be found
```
nabobalis, line 138: Do they lack the keyword or is it just gibberish to astropy.units?

wtbarnes: They lack the keyword (or any indication as to what the unit might be, we try to parse WAVEUNIT *many* ways in `io.fits`) so `super().waveunit` returns `None`.

wtbarnes:
> If you suspect that about the header, should we not replace it?

Rather than replace that file, I would recommend, in this instance, using a header-only test file and a dummy array. I think it's also fine as is. It does, albeit inadvertently, test the fact that it will prioritize the key if it exists.

Cadair (approved):

PR #7814, opened 2024-09-28
Fixed bug with tilted and offset planar screen

ayshih:
As (self-)caught by @wtbarnes, there's a calculation bug for a planar screen (added in #7115) that is both tilted (the plane is not perpendicular to the observer-Sun direction) and offset (the plane does not go through Sun center).  The incorrect code is:
```python
d_from_plane = (frame.observer.radius - self._distance_from_center) * direction.x
```
when it should be
```python
d_from_plane = frame.observer.radius * direction.x - self._distance_from_center
```
I also changed some variable names on earlier lines to facilitate future maintenance.


nabobalis (approved):

wtbarnes (approved):
Oops

PR #6885, opened 2023-03-29
Modify sample_at_coords method to accept other interpolation methods

OussCHE:
I wrote a draft that might solve the issue #6827 . I saw that @ayshih  just added a new PR #6882 that modifies sample_at_coords(which I can add to the PR).

If you accept the draft, I will start writing the unit tests (and finally modify the examples)

PR #7824, opened 2024-10-05
Adds interpolation options for sample_at_coords using scipy

### LINKED ISSUE #6827: Add interpolation options to `sample_at_coords()`

`sample_at_coords()` currently samples the value of the pixel that is nearest to the provided coordinate, which is essentially nearest-neighbor interpolation.  We should create an interpolation keyword argument that defaults to the nearest-neighbor, but also supports higher orders of interpolation.

tien-vo:
Closes #6827 

## PR Description

This is a vectorized rewrite of the attempt in #6885.

For interpolation methods other than `astropy` and `nearest`, the end-points of the arc along which to sample may not be included in the interpolation. So, it is often better to add the 4 nearest neighbors to the simplices found by `smap.wcs.world_to_array_index` (which uses `astropy.wcs.WCS.all_world2pix`). See below (hollow markers are grid points; red grid points are used for interpolation; blue markers are the arc along which to interpolate):

plot_neighbor_finder

Also, here is a comparison of all supported interpolation methods.

compare_methods

`astropy` and `nearest` agrees exactly, as expected. But there can be quite large differences in the other interpolation methods. I'm wondering if anyone has a good test for higher-order interpolation, perhaps one with a denser map and less noise?

### DIFF
--- a/changelog/7824.feature.rst
+++ b/changelog/7824.feature.rst
@@ -1,2 +1,2 @@
 The `~sunpy.map.maputils.sample_at_coords` now allows interpolation
-with unstructured data using `~scipy.interpolate.griddata`.
+with `~scipy.interpolate.griddata`.
--- a/changelog/7824.feature.rst
+++ b/changelog/7824.feature.rst
@@ -1,0 +1,2 @@
+The `~sunpy.map.maputils.sample_at_coords` now allows interpolation
+with unstructured data using `~scipy.interpolate.griddata`.
--- a/sunpy/map/maputils.py
+++ b/sunpy/map/maputils.py
@@ -166,7 +166,7 @@ def solar_angular_radius(coordinates):
     return sun._angular_radius(coordinates.rsun, coordinates.observer.radius)
 
 
-def get_neighboring_indices(indices, position, dimensions):
+def _get_neighboring_indices(indices, position, dimensions):
     """
     Get the neighboring indices relative to given indices.
 
@@ -193,22 +193,22 @@ def get_neighboring_indices(indices, position, dimensions):
            [0, 1, 2, 3, 4]])
 
     >>> # Right neighbors
-    >>> get_neighboring_indices(indices, [1, 0], [5, 5])
+    >>> _get_neighboring_indices(indices, [1, 0], [5, 5])
     array([[1, 2, 3, 4],
            [0, 1, 2, 3]])
 
     >>> # Left neighbors
-    >>> get_neighboring_indices(indices, [-1, 0], [5, 5])
+    >>> _get_neighboring_indices(indices, [-1, 0], [5, 5])
     array([[0, 1, 2, 3],
            [1, 2, 3, 4]])
 
     >>> # Top neighbors
-    >>> get_neighboring_indices(indices, [0, 1], [5, 5])
+    >>> _get_neighboring_indices(indices, [0, 1], [5, 5])
     array([[0, 1, 2, 3],
            [1, 2, 3, 4]])
 
     >>> # Bottom neighbors
-    >>> get_neighboring_indices(indices, [0, -1], [5, 5])
+    >>> _get_neighboring_indices(indices, [0, -1], [5, 5])
     array([[1, 2, 3, 4],
            [0, 1, 2, 3]])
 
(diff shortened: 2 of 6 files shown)


tien-vo:
pre-commit.ci autofix

nabobalis:
Thanks for the PR @tien-vo. We will need to give @ayshih some time to review this PR.

sunpy/map/maputils.py
```diff
@@ -164,11 +166,73 @@ def solar_angular_radius(coordinates):
     return sun._angular_radius(coordinates.rsun, coordinates.observer.radius)
 
 
-def sample_at_coords(smap, coordinates):
+def get_neighboring_indices(indices, position, dimensions):
```
ayshih, line 169: This function probably should be private (i.e., add an underscore to the beginning of the name)

sunpy/map/maputils.py
```diff
@@ -164,11 +166,73 @@ def solar_angular_radius(coordinates):
...
-    Uses nearest-neighbor interpolation of coordinates in map, as
-    it effectively uses array indexing.
+    Get the neighboring indices relative to given indices.
+
+    Parameters
+    ----------
+    indices : 2-D ndarray of integers with shape (2, n)
+        Indices to find neighbors for.
+    position : 1-D ndarray of shape (2,) or 2-element list
+        x and y position of neighbor relative to given indices.
```
ayshih, line 178: This API seems like awkward and prone to mistakes.  Unless you envision asymmetric use cases, I think the API should instead by specifying the radius in pixels of the neighborhood to return.

sunpy/map/maputils.py
```diff
@@ -164,11 +166,73 @@ def solar_angular_radius(coordinates):
...
+        & (y <= dimensions[1] - 1)
+    )
+    return np.vstack((x[in_bound], y[in_bound]))
+
+
+def sample_at_coords(
+    smap,
+    coordinates,
+    method="astropy",
+    neighbor_positions=[[1, 0], [-1, 0], [0, 1], [0, -1]],
```
ayshih, line 230: See my other comment about this API being awkward.  Also, why are you not including diagonals?

sunpy/map/maputils.py
```diff
@@ -188,10 +274,34 @@ def sample_at_coords(smap, coordinates):
...
+        return u.Quantity(smap.data[*indices], smap.unit)
+
+    extended_indices = indices.copy()
+    dimensions = [smap.dimensions.x.value, smap.dimensions.y.value]
+    for position in neighbor_positions:
+        extended_indices = np.append(
+            extended_indices,
+            get_neighboring_indices(indices, position, dimensions),
+            axis=1,
+        )
```
ayshih, line 294: You don't do anything to filter out duplicates...

changelog/7824.feature.rst
```diff
@@ -0,0 +1,2 @@
+The `~sunpy.map.maputils.sample_at_coords` now allows interpolation
+with unstructured data using `~scipy.interpolate.griddata`.
```
ayshih, line 2: Stating "unstructured data" here is unnecessary and can only lead to confusion

sunpy/map/maputils.py
tien-vo: Yes, I agree. Specifying the radius is probably the better option. I'll change it

sunpy/map/maputils.py
tien-vo: It's just to make the resulting area (red dots in original comment) surrounding the curve (blue dots) a little smoother. Do we want to include the diagonal neighbors (North-East: `[1, 1]`, North-West: `[-1, 1]`, South-East: `[1, -1]`, South-West: `[-1, -1]`) by default? Or maybe we can expose two arguments: `radius: int` and `include_diagonal_neighbors: bool`? I'm not married to the name of the second argument, [scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.generate_binary_structure.html#scipy.ndimage.generate_binary_structure) calls it `connectivity`, so we can call it that, too. But since we're dealing with a 2D image, `connectivity` can only be '1' (N, S, E, W) or '2' (also including NE, NW, SE, SW).

sunpy/map/maputils.py
tien-vo: Good point. I wonder if it affects the higher-order methods. Will look further into this.

tien-vo: The answer is no. But I'll filter duplicates since it will improve performance for `griddata`.

sunpy/map/maputils.py
tien-vo: Helpers are now private.

sunpy/map/maputils.py
tien-vo: Latest commit exposes the pixel radius.

sunpy/map/maputils.py
tien-vo: The `connectivity` thing is too complicated. I decided to just pick points within a pixel `radius`, which is the shortest path from one pixel to another. This will also include the diagonals.

PR #7827, opened 2024-10-09
Cleanup `MapSequence`

wtbarnes:
This PR does a bunch of miscellaneous cleanup on `MapSequence`. Some of it is more controversial than others so I'm happy to remove some parts or move things to other parts as needed. Overall, my aim has been to make the API closer to that of `GenericMap` but with an added dimension. 

Summary of changes below:

- Deprecate the `derotate` keyword. As far as I can tell, this has never actually done anything. 
- Refactor logic of how sort-by methods are selected. This is all internal
- Deprecate the `resample` kwarg in `plot`. This is a separation-of-concerns issue. Why should `plot` be doing any resampling? If you want to plot a resampled map, resample it yourself first.
- Deprecate `resample` in `peek`. See above.
- Do not allow for non-WCS axes to be passed to `plot`. We've not allowed this on `Map` for a long time. I'm not even sure it would work here. I've not deprecated this because I consider it undocumented behavior to begin with.
- Deprecate the method `all_maps_same_shape`. I don't see any need for this to be part of the public API. 
- Deprecate the method `at_least_one_map_has_mask`. Same as above.
- Deprecate the method `as_array` in favor of the new properties `data` and `mask`. Previously, `as_array` returned a masked array if at least one map held a mask and a bare array if not. This refactor now separates these properties similar to how `GenericMap` treats them
- Deprecate the method `all_meta` in favor of the new property `meta`

My aim was to deprecate things rather than remove them outright as that would constitute a breaking change. However, if there's no opposition, I'm also happy to outright break things.

### DIFF
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -16,6 +16,7 @@ from astropy.visualization import ImageNormalize
 from sunpy.map import GenericMap
 from sunpy.map.maputils import _clip_interval, _handle_norm
 from sunpy.util import expand_list
+from sunpy.util.decorators import deprecated, deprecated_renamed_argument
 from sunpy.util.exceptions import warn_user
 from sunpy.visualization import axis_labels_from_ctype, wcsaxes_compat
 
@@ -49,15 +50,18 @@ class MapSequence:
     To coalign a mapsequence so that solar features remain on the same pixels,
     please see the "Coalignment of MapSequences" note below.
 
+    .. note:: The individual components of a `~sunpy.map.MapSequence` can be coaligned
+              using the functions in `sunkit_image.coalignment` or using the
+              `~sunpy.map.GenericMap.reproject` method.
+
     Examples
     --------
     >>> import sunpy.map
     >>> mapsequence = sunpy.map.Map('images/*.fits', sequence=True)   # doctest: +SKIP
-
-    MapSequences can be co-aligned using the routines in `sunkit_image.coalignment`.
     """
 
-    def __init__(self, *args, sortby='date', derotate=False, **kwargs):
+    @deprecated_renamed_argument('derotate', new_name=None, since='6.1',)
+    def __init__(self, *args, sortby='date', derotate=False):
         """Creates a new Map instance"""
 
         self.maps = expand_list(args)
@@ -68,13 +72,18 @@ class MapSequence:
 
         # Optionally sort data
         if sortby is not None:
-            if sortby == 'date':
-                self.maps.sort(key=self._sort_by_date())
-            else:
-                raise ValueError("Only sort by date is supported")
+            if sortby not in self._sort_methods:
+                raise ValueError(f"sortby must be one of the following: {self._sort_methods.keys()}")
+            self.maps.sort(key=self._sort_methods[sortby])
 
         if derotate:
-            self._derotate()
+            raise NotImplementedError("This functionality has not yet been implemented.")
+
+    @property
+    def _sort_methods(self):
+        return {
+            "date": lambda m: m.date,
+        }
 
     def __getitem__(self, key):
         """Overriding indexing operation.  If the key results in a single map,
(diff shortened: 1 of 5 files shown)


sunpy/map/mapsequence.py
```diff
@@ -475,19 +468,38 @@ def peek(self, resample=None, **kwargs):
...
+    def data(self):
+        if not np.all([m.data.shape == self[0].data.shape for m in self]):
+            raise ValueError('Not all maps have the same shape.')
+        data = np.asarray([m.data for m in self.maps])
+        return np.swapaxes(np.swapaxes(data, 0, 1), 1, 2)
+
+    @property
+    def mask(self):
+        if not np.any([m.mask is not None for m in self.maps]):
+            return
```
nabobalis, line 495:
```suggestion
            return None
```

wtbarnes:
I've added a single changelog listing all of the deprecations, but I'm happy to break this up into multiple files if that is preferred.

sunpy/map/mapsequence.py
```diff
@@ -68,13 +76,18 @@ def __init__(self, *args, sortby='date', derotate=False, **kwargs):
 
         # Optionally sort data
         if sortby is not None:
-            if sortby == 'date':
-                self.maps.sort(key=self._sort_by_date())
-            else:
-                raise ValueError("Only sort by date is supported")
+            if sortby not in self._sort_methods:
+                raise ValueError(f"sortby must be one of the following: {self._sort_methods.keys()}")
```
nabobalis, line 80:
```suggestion
                raise ValueError(f"sortby must be one of the following: {list(self._sort_methods.keys())}")
```
Otherwise I think the output looks like:

>>> f'{abc.keys()}'
"dict_keys(['a', 'b'])"

nabobalis (approved):

sunpy/map/mapsequence.py
```diff
@@ -473,19 +471,38 @@ def peek(self, resample=None, **kwargs):
...
     def at_least_one_map_has_mask(self):
         """
-        Tests if at least one map has a mask.
+        True if at least one map has a mask
         """
-        return np.any([m.mask is not None for m in self.maps])
+        return self.mask is not None
 
+    @property
+    def data(self):
```
Cadair, line 489: Needs a docstring.

sunpy/map/mapsequence.py
```diff
@@ -473,19 +471,38 @@ def peek(self, resample=None, **kwargs):
...
+        if not np.all([m.data.shape == self[0].data.shape for m in self]):
+            raise ValueError('Not all maps have the same shape.')
+        data = np.asarray([m.data for m in self.maps])
+        return np.swapaxes(np.swapaxes(data, 0, 1), 1, 2)
+
+    @property
+    def mask(self):
```
Cadair, line 496: needs a docstring.

sunpy/map/mapsequence.py
```diff
@@ -498,43 +515,39 @@ def as_array(self):
...
 
+    @deprecated(since='6.1', message=v61_DEPRECATION_MESSAGE, alternative='meta property')
     def all_meta(self):
         """
         Return all the meta objects as a list.
         """
+        return self.meta
+
+    @property
+    def meta(self):
```
Cadair, line 532: needs a docstring.

sunpy/map/mapsequence.py
```diff
@@ -473,19 +471,38 @@ def peek(self, resample=None, **kwargs):
 
         return MapSequenceAnimator(plot_sequence, **kwargs)
 
+    @deprecated(since='6.1', message=v61_DEPRECATION_MESSAGE, alternative='data property')
```
Cadair, line 474:
```suggestion
```

I don't think we should remove this, as it feels like a useful way to know if `.data` is going to fail and deprecating it without a replacement feels problematic.

sunpy/map/mapsequence.py
```diff
@@ -49,15 +54,18 @@ class MapSequence:
     To coalign a mapsequence so that solar features remain on the same pixels,
     please see the "Coalignment of MapSequences" note below.
 
+    .. note:: The individual components of a `~sunpy.map.MapSequence` can be coaligned
+              using the functions in `sunkit_image.coalignment` or using the
+              `~sunpy.map.GenericMap.reproject` method.
```
Cadair, line 59:
```suggestion
              `~sunpy.map.GenericMap.reproject_to` method.
```

sunpy/map/mapsequence.py
```diff
@@ -498,43 +535,42 @@ def as_array(self):
...
-            if self.at_least_one_map_has_mask():
-                mask_sequence = np.zeros_like(data, dtype=bool)
-                for im, m in enumerate(self.maps):
-                    if m.mask is not None:
-                        mask_sequence[:, :, im] = m.mask
-                return ma.masked_array(data, mask=mask_sequence)
-            else:
-                return data
+        data = self.data
+        if (mask:=self.mask) is not None:
```
Cadair, line 539:
```suggestion
        if (mask := self.mask) is not None:
```

Cadair (approved):

nabobalis:
ONE MORE DOWN. LETS GO

Issue #6522, opened 2022-10-30
Support for `center=...` in `get_rectangle_coordinates`


wtbarnes:
### Describe the feature

I often find myself specifying a rectangular region in terms of the center of that rectangle and the width and height. However, `get_rectangle_coordinates` only accepts a bottom left and top right coordinate or a bottom left and width and height, meaning I often have to translate a center coordinate to a bottom left coordinate in order to specify a rectangular region. It would be convenient if `get_rectangle_coordinates` allowed for passing in a center coordinate as well.

ayshih:
I'll just point out that for a subset of situations (e.g., `Helioprojective` on the solar disk), you can create essentially the rectangle that you want by calling:
```python
rectangle = center.spherical_offsets_by(width/[-2, 2], height/[-2, 2])
```
and skip using `get_rectangle_coordinates()`.  It's not exactly the same rectangle, but depending on where `width` and `height` are coming from, it may actually be formally more correct.

Cadair:
oooh, that should be documented somewhere.

wtbarnes:
(*I originally posted this on #7836, but am moving it here as it seems more sensible to have this discussion on the original issue*)

Before proceeding any further with #7836, I think we need to have a bit of a discussion as to whether we want such an object to live in `sunpy` at all. Some questions we should consider:

1. Is it worth having an additional object just to be able to specify a rectangle using the center coordinate?
2. Is there intended scope beyond just specifying something that is passed to `submap` or `draw_quadrangle`?
3. What about non-rectangular regions?
4. How does this overlap with something like the [`regions`](https://astropy-regions.readthedocs.io/en/stable/) package?

nabobalis:
My very quick thoughts are that unless a new class offers useful functionality, I think expanding the API to submap/draw_quadrangle to allow center=X,Y and width/height makes the most sense to me.

Prtm2110:
Alright, that makes sense! Should I proceed with expanding the API in `get_rectangle_coordinate`, `submap`, and `draw_quadrangle` to include center=X, Y and width/height for rectangle specification? I can modify the existing PR to reflect this approach. Let me know if there are any additional considerations or specific guidelines for this update.

nabobalis:
We have not had time to discuss this at the weekly meeting, so I would hold off until that happens.

PR #7836, opened 2024-10-16
Adds `Rectangle` class with class methods that allow for easier creation of Rectangle objects 

Prtm2110:
## PR Description

This PR introduces the Rectangle class with methods to create a rectangular regions using different input options:

`from_corners`: Allows creation of a rectangle by specifying the bottom-left and top-right coordinates.
`from_bottom_left`: Creates a rectangle using the bottom-left coordinate along with width and height.
`from_center`: Enables rectangle creation from the center coordinate, width, and height, automatically calculating the bottom-left and top-right corners.

Each method can optionally return a tuple of `(bottom_left, top_right)` if `as_tuple=True`, making it flexible to work with both rectangle objects and raw coordinates.

Solves #6522

### DIFF
--- a/sunpy/coordinates/utils.py
+++ b/sunpy/coordinates/utils.py
@@ -391,6 +391,25 @@ class Rectangle:
     def __init__(self, bottom_left: SkyCoord, top_right: SkyCoord):
         """
         Initializes a Rectangle object with bottom-left and top-right coordinates.
+
+        Parameters
+        ----------
+        bottom_left : `~astropy.coordinates.SkyCoord`
+            The bottom-left corner of the rectangle.
+        top_right : `~astropy.coordinates.SkyCoord`
+            The top-right corner of the rectangle.
+
+        Examples
+        --------
+        >>> from astropy.coordinates import SkyCoord
+        >>> import astropy.units as u
+        >>> bl = SkyCoord(10*u.deg, -10*u.deg, frame='heliographic_stonyhurst')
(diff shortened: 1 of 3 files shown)


sunpy/coordinates/utils.py
```diff
@@ -387,6 +387,103 @@ def get_rectangle_coordinates(bottom_left, *, top_right=None,
...
+            raise ValueError("The specified width cannot be negative.")
+        if width > 360 * u.deg:
+            raise ValueError("The specified width cannot be greater than 360 degrees.")
+        if height < 0 * u.deg:
+            raise ValueError("The specified height cannot be negative.")
+        if center.spherical.lat + height / 2 > 90 * u.deg or center.spherical.lat - height / 2 < -90 * u.deg:
+            raise ValueError("The specified height exceeds the maximum latitude.")
+
+        bottom_left = SkyCoord(
+            center.spherical.lon - width / 2,
```
Prtm2110, line 468: Should we call these width and height, considering that we are defining the rectangle from the center? It's implied that the topmost point would be at half the distance from the center.

Prtm2110:
Please let me know what is to be done here, I will try to get it done before the release.

prathamhole14:
Hi @wtbarnes, thank you for the review. If we decide to modify the API of `get_rectangle_coordinates`, I will reset this PR and make the changes here itself. Please reopen this if that's the case.

PR #7837, opened 2024-10-16
Use `sphinxcontrib-bibtex` to reference papers

### LINKED ISSUE #7078: Use `sphinxcontrib-bibtex` to manage references
Throughout the package, we have references to various papers, books, etc. These are largely formatted as links. Instead, we should manage these references with a BibTeX file and the [`sphinxcontrib-bibtex`](https://sphinxcontrib-bibtex.readthedocs.io/en/latest/quickstart.html) extension.

This would provide an easier way to manage references to external resources and provide more informative links to reference materials. As an example of what this looks like in practice, [see the `fiasco` docs](https://fiasco.readthedocs.io/en/stable/api/fiasco.Ion.html#fiasco.Ion.direct_ionization_cross_section).

wtbarnes:
Fixes #7078 

This is a start at using `sphinxcontrib-bibtex` to reference all papers or anything for which we can write a Bibtex entry. This replaces links with references to a centralized bibliography.

I'm sure I'm still missing some paper links that need replacing. 

- [x] Find more references that need replacing??
- [x] Add a section to the dev guide saying how to reference things


nabobalis:
You think its worth a backport or should we just roll it out for 6.1?

wtbarnes:
Nah, not worth a backport.

Issue #7839, opened 2024-10-18
Fix this specific improper comparison between quantities and strings 


GillySpace27:
### Describe the bug

When I was trying to construct a new Fido Client, I ran into an issue where there was a comparison happening that threw exceptions. In the else clause of the following method, you can see the original and my hotfix. 

In sunpy.net.attr.Range:

>  Around line 470
    def __contains__(self, other):
        if isinstance(other, Range):
            return self.min <= other.min and self.max >= other.max
        else:
            return self.min <= other <= self.max
            # return self.min.value <= int(other) <= self.max.value #TODO Fix Hotfix

I believe that it was happening when I was trying to get the client to understand my a.Wavelength. Removing that attr from the query fixes the exception and also returns an result with all wavelengths included, which makes sense. But leaving it in causes the error.

### To Reproduce

    from sunpy.net import Fido
    from sunpy.net import attrs as a
    import astropy.units as u

    # Example usage
    time_range = a.Time("2023-10-11 00:00:00", "2023-10-11 01:00:00")
    instrument = a.Instrument("AIASynoptic")
    wavelength = a.Wavelength(193 * u.angstrom)
    sample = a.Sample(10 * u.minute)
    results = Fido.search(time_range, instrument, wavelength)  # , sample)  # , wavelength)

    print(results)

    Fido.fetch(results, path="DLData/{file}")

Here's the error:

>File "/Users/cgilbert/vscode/sunpy/sunpy/.venv/lib/python3.12/site-packages/astropy/units/quantity_helper/converters.py", line 196, in converters_and_unit
    if can_have_arbitrary_unit(args[i]):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cgilbert/vscode/sunpy/sunpy/.venv/lib/python3.12/site-packages/astropy/units/quantity_helper/converters.py", line 138, in can_have_arbitrary_unit
    return np.all(np.logical_or(np.equal(value, 0.0), ~np.isfinite(value)))
                                ^^^^^^^^^^^^^^^^^^^^
numpy._core._exceptions._UFuncNoLoopError: ufunc 'equal' did not contain a loop with signature matching types (<class 'numpy.dtypes.StrDType'>, <class 'numpy.dtypes._PyFloatDType'>) -> None

> During handling of the above exception, another exception occurred:

>Traceback (most recent call last):
[...]

PR #7840, opened 2024-10-18
Fix the invalid comparison

GillySpace27:
## PR Description

I was making a new client and I found a bug in the attrs module so I'm trying to update that so no-one else spends a week bug-hunting.  This addresses issue #7839 .

### DIFF
--- a/sunpy/net/attr.py
+++ b/sunpy/net/attr.py
@@ -448,7 +448,7 @@ class Range(DataAttr):
         if isinstance(other, Range):
             return self.min <= other.min and self.max >= other.max
         else:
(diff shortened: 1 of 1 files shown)


nabobalis:
Can you also submit the client, on the surface all the other clients seem to be ok with the current attr code.

So I think it is worth working out if this is a change that makes sense or if there is a "gotcha" in the way clients need to be setup to work with the attr system that we never found before.

GillySpace27:
I just opened PR #7841 to address the client's inclusion. :)

nabobalis:
I will close this in favour of the PR that adds the client.

Issue #7828, opened 2024-10-13
Add a new FIDO client that downloads the AIA synoptic Dataset


GillySpace27:
### Describe the feature

Downloading and working with large sets of AIA 4k x 4k images can be time-consuming and expensive. I learned of the existence of another set of AIA images that are hosted on the jsoc website but they aren't able to be seen by FIDO, which feels like a choice someone made. Nevertheless, they are here: [https://jsoc1.stanford.edu/data/aia/synoptic/](https://jsoc1.stanford.edu/data/aia/synoptic/ ) 

Their description: 
>“This directory contains FITS files at a 2-minute cadence which have been produced by processing 4Kx4K AIA level 1 images to level 1.5, then reducing them to 1024x1024 2.4 arc-sec pixels by summing each non NaN valued pixel in the 4x4 pixel cell and dividing by the number of non NaN valued pixels. The pixel value units are DN per original 0.6 arc-sec pixel. The files are stored in a year, month, day, hour, and filename directory hierarchy. These images are produced approximately 7 days after T_OBS since they are derived from definitive level 1 data. The image_times and image_times.json provide the time for the most recent (current final) synoptic file.”

The ability to query and download from this dataset is self-evident, as it does basically all of the prep work to look at the images you need, along with being both spatially binned and time integrated. Just a fantastic dataset (unless you're looking at the most fine-scale or dynamic features that are detectable by the instrument. 

### Proposed solution

Anyways, I've already written a couple of other clients for FIDO now, so I went ahead and implemented a version of a Client that seems to work. I'm sure we'll have to go back and forth but I'm going to set up a pull request draft to get the process going.  I also wouldn't be surprised if there are other approaches than my implementation, which might be more consistent with the other Clients.

wtbarnes:
If this data is available via the JSOC, it should be accessible through drms and then by extension the JSOCClient that already is supported in Fido. Is there some reason that the existing client is not adequate for querying this dataset?

GillySpace27:
Well, I guess I see the product on the status page: [http://jsoc.stanford.edu/data/jsoc_proc_status.htm](http://jsoc.stanford.edu/data/jsoc_proc_status.html)
but I can't use that as the series in an attrs.jsoc.Series call, so I'm not sure what else to do. I can't find it mentioned anywhere else either. Please let me know what you think I should do.

wtbarnes:
[...]

PR #7841, opened 2024-10-18
Add a client to Fido which can discover the synoptic AIA series

GillySpace27:
## PR Description

This PR addresses Issue #7828 by adding a new client into the dataretriever module

Some discussion should be had regarding the best way to call for this Series since it isn't managed by JSOC in the usual way. For the moment if a.Instrument is either "AIA" or "AIASynoptic" the client will be called. a.Resolution == "Synoptic" will also trigger alongside a.Instrument == "AIA". That said, I'm not sure the average user would like to have to start specifying a resolution like that in order to not be served with this series, so I'm thinking the unique instrument might be the way to go.


nabobalis:
It might be tricky but can you undo all of the code formatting changes? There have been discussions around auto formatting the code but they have not been accepted yet

GillySpace27:
Sorry about that!  Of course I set up the formatter to run on save, so I'm not always seeing what changes it makes. How do you think I should go about it? Are there standard approaches?

nabobalis:
Err, honestly I don't know what would be best. It might be simplest to just start again from a clean branch and add just the new client and attr code.

nabobalis:
Replaced by https://github.com/sunpy/sunpy/pull/7842

PR #7844, opened 2024-10-21
`GenericMap.draw_contour()` makes use of the `plot_settings` dictionary

Prtm2110:

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -2474,6 +2474,19 @@ class GenericMap(NDData):
         Extra keyword arguments to this function are passed through to the
         corresponding matplotlib method.
         """
+        contour_params = [
+            'corner_mask', 'colors', 'alpha', 'cmap', 'norm', 'vmin', 'vmax',
+            'origin', 'extent', 'locator', 'extend', 'xunits', 'yunits',
+            'antialiased', 'nchunk', 'linewidths', 'linestyles', 'negative_linestyles',
+            'hatches', 'algorithm', 'clip_path'
+        ]
+
+        plot_settings = self.plot_settings.copy()
+        # Transfer relevant settings to contour_args
+        for key in contour_params:
+            if key in plot_settings:
+                contour_args[key] = plot_settings.pop(key)
+
         axes = self._check_axes(axes)
         levels = self._process_levels_arg(levels)
 
--- a/sunpy/map/tests/test_mapbase.py
+++ b/sunpy/map/tests/test_mapbase.py
@@ -2,6 +2,7 @@
 Test Generic Map
 """
 import re
+import copy
 import tempfile
 from copy import deepcopy
 
@@ -1726,6 +1727,25 @@ def test_draw_carrington_map(carrington_map):
     ax = fig.add_subplot(1, 1, 1, projection=carrington_map)
     carrington_map.plot(axes=ax)
     return fig
+
+
+def test_plot_settings_and_contour_params(simple_map):
+    new_plot_settings = copy.deepcopy(simple_map.plot_settings)
+    new_plot_settings['alpha'] = 0.5
+    new_plot_settings['linewidths'] = 3.0
+
+    # Ensure the new settings reflect the changes
+    assert new_plot_settings['alpha'] == 0.5
+    assert new_plot_settings['linewidths'] == 3.0
+
+    # Set the new plot settings back to the map
+    simple_map.plot_settings = new_plot_settings
+
+    # Ensure draw_contours uses these settings correctly
+    levels = [0.5, 1.0, 1.5]
+    contour_plot = simple_map.draw_contours(levels)
+    assert contour_plot.get_alpha() == 0.5
+    assert contour_plot.get_linewidths()[0] == 3.0
 
 
 @pytest.mark.parametrize('method', _rotation_registry.keys())


sunpy/map/tests/test_mapbase.py
```diff
@@ -1728,6 +1729,25 @@ def test_draw_carrington_map(carrington_map):
...
 
 
+def test_plot_settings_and_contour_params(simple_map):
+    new_plot_settings = copy.deepcopy(simple_map.plot_settings)
+    new_plot_settings['alpha'] = 0.5
+    new_plot_settings['linewidths'] = 3.0
+
+    # Ensure the new settings reflect the changes
+    assert new_plot_settings['alpha'] == 0.5
+    assert new_plot_settings['linewidths'] == 3.0
```
nabobalis, line 1739: What is this testing?

sunpy/map/mapbase.py
```diff
@@ -2474,6 +2474,19 @@ def draw_contours(self, levels, axes=None, *, fill=False, **contour_args):
         Extra keyword arguments to this function are passed through to the
         corresponding matplotlib method.
         """
+        contour_params = [
+            'corner_mask', 'colors', 'alpha', 'cmap', 'norm', 'vmin', 'vmax',
+            'origin', 'extent', 'locator', 'extend', 'xunits', 'yunits',
+            'antialiased', 'nchunk', 'linewidths', 'linestyles', 'negative_linestyles',
+            'hatches', 'algorithm', 'clip_path'
+        ]
```
nabobalis, line 2482: How does this work if MPL update their keywords support?

Prtm2110:
If `contour_args` is directly updated with `plot_settings`, it may pass down irrelevant or unsupported information from `plot_settings`, leading to incorrect contours (as described in the PR). To address this, I created a function that handles this logic. It selectively updates `contour_args` with relevant entries from `plot_settings` while removing unsupported or conflicting keys. This allows the user to override arguments like `origin` without having them unintentionally set.

Here’s the detailed reasoning:

- **`cmap`:** Setting `cmap = None` avoids the error:  
  > *ValueError: Either colors or cmap must be None*  
  Since contour plots handle colors separately via the `colors` argument, passing a colormap (`cmap`) is unnecessary.

- **`interpolation`:** Matplotlib's `contour` does not use this argument. Passing it results in the warning:  
  > *The following kwargs were not used by contour: 'interpolation'*  
  Therefore, this argument is removed.

- **`origin`:** If `'origin': 'lower'` is specified, it is updated to `'origin': None` because `None` is the default for contour plots in Matplotlib. This ensures that the behavior aligns with Matplotlib's expected defaults.


- **`norm`:** The condition:
  A better approach is to let the user control the norm setting in plot_settings['norm'] for this the below logic is to be implemented but its not quite working. 
  ```python
    if contour_args['norm'] == self.norm_to_check_contour:
        contour_args['norm'] = None

sunpy/map/mapbase.py
```diff
@@ -2445,6 +2447,64 @@ def _process_levels_arg(self, levels):
...
+
+        # if contour_args['norm'] == self.norm_to_check_contour:
+        #     contour_args['norm'] = None
+
+        # Remove unsupported parameters for contour plots
+        contour_args['cmap'] = None  # Contour plots can not use 'cmap' if 'colors' is not None
+        contour_args.pop('interpolation')
+        contour_args['norm'] = None
+
+        return contour_args
```
nabobalis, line 2505: What is the plan for this section of code?

nabobalis: Why the need for a seperate function?

Prtm2110: I created it to keep things clean and manageable, it's not essential, should I move it inside `draw contour`?

nabobalis: No, I think it's fine,  I was just curious.

sunpy/map/mapbase.py
```diff
@@ -247,6 +247,8 @@ def __init__(self, data, header, plot_settings=None, **kwargs):
             # Put import here to reduce sunpy.map import time
             from matplotlib import colors
             self.plot_settings['norm'] = colors.Normalize()
+            self.norm_to_check_contour = self.plot_settings['norm']
```
nabobalis, line 250: Is this used somewhere?

sunpy/map/mapbase.py
```diff
@@ -2445,6 +2447,64 @@ def _process_levels_arg(self, levels):
             raise u.UnitsError("This map has no unit, so levels can only be specified in percent "
                                "or in u.dimensionless_unscaled units.")
 
+
+    def _update_contour_args(self, contour_args):
+        """
+        Updates `contour_args` with values from `plot_settings`, ensuring
+        compatibility with Matplotlib's contour plotting. Removes unsupported
```
nabobalis, line 2454: How does it ensure this?

sunpy/map/tests/test_mapbase.py
```diff
@@ -1728,6 +1729,22 @@ def test_draw_carrington_map(carrington_map):
     return fig
 
 
+def test_plot_settings_and_contour_params(simple_map):
```
nabobalis, line 1732: Is there is a small set of keywords this supports, I would like all of them tested if possible.

sunpy/map/mapbase.py
Prtm2110:
```
 default_contour_param = {
            'origin': None,
        }

  # Replace conflicting settings with contour defaults
  for key in original_plot_defaults:
      if key in contour_args and contour_args[key] == original_plot_defaults[key]:
          contour_args[key] = default_contour_param[key]
```
If the dict pair from (just newly concatenated contour_args) is also present in `original_plot_defaults`(this stores the plot_settings that have been set in the init method of `GenericMap` for map plotting) then this value of the pair is to be replaced with the corresponding key's value from `default_contour_param`(this stores the overlapping matplotlib contour parameter and there default vales)

nabobalis: Maybe it's too late in the day for me but I don't see how that ensures compatibility with Matplotlib contour plotting. What if they remove a keyword argument?

Prtm2110: Ah, you mean what happens if an unsupported argument is passed through `plot_settings`, right? 
In that case, it will simply ignore the argument. Ideally, we should warn the user or raise an error, but this isn't feasible because, as far as I know, there’s no straightforward way to retrieve the supported Matplotlib contour keywords. While we could manually create a list of supported keywords, as you mentioned, this might fail if they update their keywords support.

sunpy/map/mapbase.py
Prtm2110:
```
  if contour_args['norm'] == self.norm_to_check_contour:
      contour_args['norm'] = None
```
I expected this if condition to evaluate to True, but it is not (as the two objects are different). Do you have any idea why this might be happening?

sunpy/map/mapbase.py
nabobalis: So in that case, most of this documentation string is incorrect?

changelog/7844.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed a bug where custom values in the ``plot_settings`` dictionary were not being correctly applied in the :meth:`sunpy.map.GenericMap.draw_contours` method, ensuring proper propagation of settings to contour plots.
```
nabobalis, line 1:
```suggestion
Fixed custom values in the ``plot_settings`` dictionary not being correctly applied by :meth:`sunpy.map.GenericMap.draw_contours`.
```

sunpy/map/mapbase.py
```diff
@@ -2445,6 +2447,64 @@ def _process_levels_arg(self, levels):
...
+            and default values applied where necessary.
+
+        Notes
+        -----
+        - 'cmap': Set to `None` to avoid the error "ValueError: Either colors or cmap must be None".
+        - 'interpolation': Removed because Matplotlib's contour function raises the warning
+        "The following kwargs were not used by contour: 'interpolation'".
+        - 'origin': If `'origin': 'lower'` is present, it is replaced with `'origin': None`,
+        as `None` is the default value for Matplotlib's contour plots.
+        """
```
nabobalis, line 2480:
```suggestion
        """
        Updates ``contour_args`` with values from ``plot_settings``.
        
        Parameters
        ----------
        contour_args : dict
            A dictionary of arguments to be used for contour plotting.

        Returns
        -------
        dict
            The updated `contour_args` dictionary.

        Notes
        -----
        - 'cmap': Set to `None` to avoid the error "ValueError: Either colors or cmap must be None".
        - 'interpolation': Removed because Matplotlib's contour function raises the warning
        "The following kwargs were not used by contour: 'interpolation'".
        - 'origin': If `'origin': 'lower'` is present, it is replaced with `'origin': None`,
        as `None` is the default value for Matplotlib's contour plots.
        """
```

sunpy/map/mapbase.py
```diff
@@ -2445,6 +2447,64 @@ def _process_levels_arg(self, levels):
...
+            'origin': None,
+        }
+
+        # Replace conflicting settings with contour defaults
+        for key in original_plot_defaults:
+            if key in contour_args and contour_args[key] == original_plot_defaults[key]:
+                contour_args[key] = default_contour_param[key]
+
+        # if contour_args['norm'] == self.norm_to_check_contour:
+        #     contour_args['norm'] = None
```
nabobalis, line 2500:
```suggestion
```

sunpy/map/mapbase.py
```diff
@@ -2445,6 +2447,64 @@ def _process_levels_arg(self, levels):
...
+
+        # Remove unsupported parameters for contour plots
+        contour_args['cmap'] = None  # Contour plots can not use 'cmap' if 'colors' is not None
+        contour_args.pop('interpolation')
+        contour_args['norm'] = None
```
```suggestion
		# Contour plots can not use 'cmap' if 'colors' is not None
        contour_args['cmap'] = None
        contour_args.pop('interpolation')
        contour_args['norm'] = None
```
How does this account for that comment?

sunpy/map/mapbase.py
```diff
@@ -2474,6 +2534,9 @@ def draw_contours(self, levels, axes=None, *, fill=False, **contour_args):
         Extra keyword arguments to this function are passed through to the
         corresponding matplotlib method.
         """
+
+        contour_args = self._update_contour_args(contour_args)
+
```
nabobalis, line 2543:
```suggestion
        contour_args = self._update_contour_args(contour_args)
```

changelog/7844.bugfix.rst
nabobalis, line 1:
```suggestion
Fixed a bug where custom values in the ``plot_settings`` dictionary were not being correctly applied in the :meth:`sunpy.map.GenericMap.draw_contours` method.
```

sunpy/map/mapbase.py
nabobalis: How are they difference exactly? Difference values?

sunpy/map/mapbase.py
```diff
@@ -2445,6 +2447,64 @@ def _process_levels_arg(self, levels):
...
+            if key in contour_args and contour_args[key] == original_plot_defaults[key]:
+                contour_args[key] = default_contour_param[key]
+
+        contour_args['cmap'] = None  # 'cmap' cannot be used for contour plots when levels are not None,in-case of composite maps.
+        contour_args.update(contour_args_copy)  # If 'draw_contour' is used, this ensures the method arguments are applied.
+
+        contour_args.pop('interpolation')
+        contour_args['norm'] = None
+
+        return contour_args
```
nabobalis, line 2505:
```suggestion
        contour_args.update(plot_settings)
        # Define default settings for normal plots and contour-specific updates
        original_plot_defaults = {
            'origin': 'lower',
        }
        default_contour_param = {
            'origin': None,
        }
        # Replace conflicting settings with contour defaults
        for key in original_plot_defaults:
            if key in contour_args and contour_args[key] == original_plot_defaults[key]:
                contour_args[key] = default_contour_param[key]
        # 'cmap' cannot be used for contour plots when levels are not None,
        # which is the case in composite maps.
        contour_args['cmap'] = None
        # If 'draw_contour' is used, this ensures the method arguments are applied.
        contour_args.update(contour_args_copy)
        contour_args.pop('interpolation')
        contour_args['norm'] = None
        return contour_args
```

Where is interpolation passed and I assume its not allowed to be passed into MPL contours?

sunpy/map/mapbase.py
Prtm2110: For eg like this `<matplotlib.colors.Normalize object at 0x7edef5363520>`
`<matplotlib.colors.Normalize object at 0x7edef53c3550>`

sunpy/map/mapbase.py
Prtm2110: Yes, it's not allowed to be passed

sunpy/map/mapbase.py
nabobalis: Right, so the objects are different, so it won't work. 

What is the purpose of the check?

sunpy/map/mapbase.py
nabobalis: Where does it come from?

nabobalis: Ah its set in the original dict.

sunpy/map/mapbase.py
nabobalis: So you want to set it to None if its been set at all?

Prtm2110: Set it to none if its been originally set in GenericMap

nabobalis: Then can we not check if it exists?

Prtm2110: We can check `isinstance(norm , colors.Normalize())` but the problem with is this that mlp contour accepts string as well as `colors.Normalize()` for norm
https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.contour.html

nabobalis: What if we check if just exists?

Prtm2110: Nope, if we just check if that exits then user passing `norm` in setting or directly in draw_contour is useless

nabobalis: Well if its set by a user directly into draw_contour, cant we just use that and ignore if it exists in plot settings?

Prtm2110: > Well if its set by a user directly into draw_contour, cant we just use that and ignore if it exists in plot settings?

Ah, Thanks for pointing this out, fixed this in new commit

nabobalis: So is this line still needed?

sunpy/map/mapbase.py
```diff
@@ -2445,6 +2447,53 @@ def _process_levels_arg(self, levels):
...
+            'origin': None,
+        }
+        # Replace conflicting settings with contour defaults
+        for key in original_plot_defaults:
+            if key in contour_args and contour_args[key] == original_plot_defaults[key]:
+                contour_args[key] = default_contour_param[key]
+        # 'cmap' cannot be used for contour plots when levels are not None,
+        # which is the case in composite maps.
+        contour_args['norm'] = None
+        contour_args['cmap'] = None
```
nabobalis, line 2490:
```suggestion
        # 'cmap' cannot be used for contour plots when levels are not None,
        # which is the case in composite maps.
        contour_args['cmap'] = None
        contour_args['norm'] = None
```
Maybe add a comment about the norm here?

sunpy/map/mapbase.py
Prtm2110: Still it's a issue if passed through plotsettings (commit I made fixes it for draw_contour only)

Prtm2110: Yes we can, adding a comment about it.

sunpy/map/mapbase.py
```diff
@@ -2445,6 +2445,54 @@ def _process_levels_arg(self, levels):
...
+
+        Parameters
+        ----------
+        contour_args : dict
+            A dictionary of arguments to be used for contour plotting.
+
+        Returns
+        -------
+        dict
+            The updated `contour_args` dictionary.
```
nabobalis, line 2461:
```suggestion
            The updated ``contour_args`` dictionary.
```

sunpy/map/tests/test_compositemap.py
```diff
@@ -65,6 +65,27 @@ def test_plot_composite_map_colors(composite_test_map):
 
 
 def test_plot_composite_map_mplkwargs(composite_test_map):
```
nabobalis, line 67: What is this testing?

Prtm2110: It is testing if the supported mpl contour kwargs which are passed through plot_settings are properly reflected or not

nabobalis: This is just checking if dictionary assignment works tho?

Prtm2110: Yes can simply say that

Prtm2110: Shall I change the test name?

nabobalis: So unless it's testing something related to the code changes in this PR, I don't see the benefit of this test.

nabobalis: So it's essentially not testing anything to do with the code changes in this PR?

nabobalis: Yes, the test needs to be changed to be meaningful. What it should do, I don't know

Prtm2110: I have modified the tests, please review them.

sunpy/map/tests/test_compositemap.py
```diff
@@ -64,7 +64,22 @@ def test_plot_composite_map_colors(composite_test_map):
     composite_test_map.plot(colors='red')
 
 
-def test_plot_composite_map_mplkwargs(composite_test_map):
+def test_plot_composite_map_updated_args(composite_test_map, simple_map):
+    levels = [-1000, -500, -250, 250, 500, 1000] * u.G
+    composite_test_map.set_levels(index=1, levels=levels)
```
nabobalis, line 69: What is this testing?

sunpy/map/tests/test_compositemap.py
nabobalis: I would also check that the work arounds for cmap etc are correct for normal maps instead of just composite maps.

We have to make sure these do not break any existing map calls for draw_contours

Prtm2110: I included `contour_args = {'norm': 'log'}` to confirm whether the original `contour_args` from `draw_contours` take precedence over `plot_settings`. Should I do the same for `cmap`, or would it be better to create a new test specifically for this?

nabobalis: Do the same for cmap in the same test.

sunpy/map/tests/test_compositemap.py
Prtm2110: > I included `contour_args = {'norm': 'log'}` to confirm whether the original `contour_args` from `draw_contours` take precedence over `plot_settings`. 

Testing if the right contours args will be finally passed down to mlp contour

sunpy/map/tests/test_compositemap.py
Prtm2110: Added.

nabobalis (approved):

hayesla (approved):

PR #7851, opened 2024-10-26
Add function and map method for drawing extent of another wcs

### LINKED ISSUE #7615: Add a Map method to draw the extent of the map
It's a common need to plot the extent of one map (e.g., the FOV of an instrument) onto a different WCSAxes (e.g., an image from a different instrument).  One can just use the four corners of the map to draw a box if the pixel->world transformation is the same (i.e., everything is aligned), but this will be inaccurate if the reprojection is nontrivial.

Here's an example of overlaying an HMI SHARP map on top of an AIA map.  Using just the four corners of the HMI SHARP map to draw the extent (red solid line) does not faithfully follow the map bounds due to the WCS projection used for the HMI SHARP map.  Let's add a convenience method that properly plots the map extent (green dashed line).

download (12)

This method would also naturally help those working with two maps from significantly different vantage points.

wtbarnes:
Fixes #7615 

TODOs:

- [x] Figure tests
- [x] Docs for map method
- [x] Better docs for visualization function
- [x] Changelog

### DIFF
--- a/changelog/7851.feature.rst
+++ b/changelog/7851.feature.rst
@@ -1,2 +1,2 @@
-Add `sunpy.visualization.drawing.extent` to visualize the extent of another WCS on an axis.
+Add :func:`sunpy.visualization.drawing.extent` to visualize the extent of another WCS on an axis.
 Add `sunpy.map.GenericMap.draw_extent` to visualize the extent of a map on a given axis.
--- a/changelog/7851.feature.rst
+++ b/changelog/7851.feature.rst
@@ -1,2 +1,2 @@
 Add :func:`sunpy.visualization.drawing.extent` to visualize the extent of another WCS on an axis.
-Add `sunpy.map.GenericMap.draw_extent` to visualize the extent of a map on a given axis.
+Add :meth:`sunpy.map.GenericMap.draw_extent` to visualize the extent of a map on a given axis.
--- a/changelog/7851.feature.rst
+++ b/changelog/7851.feature.rst
@@ -1,2 +1,2 @@
-Add `sunpy.visualization.drawing.extent` and `sunpy.map.GenericMap.draw_extent` to visualize the extent
-of another WCS on a map.
+Add `sunpy.visualization.drawing.extent` to visualize the extent of another WCS on an axis.
+Add `sunpy.map.GenericMap.draw_extent` to visualize the extent of a map on a given axis.
--- a/changelog/7851.feature.rst
+++ b/changelog/7851.feature.rst
@@ -1,0 +1,2 @@
+Add `sunpy.visualization.drawing.extent` and `sunpy.map.GenericMap.draw_extent` to visualize the extent
+of another WCS on a map.
--- a/docs/whatsnew/6.1.rst
+++ b/docs/whatsnew/6.1.rst
@@ -55,3 +55,13 @@ Support has been added to ``sunpy.map.Map`` and ``sunpy.timeseries.TimeSeries`` 
 Support for S3 filesystems is tested, but other fsspec plugins should work for most filetypes, please open issues if errors are encountered with other filesystems.
 
 You will need to install packages for your desired filesystem so for S3 the ``s3fs`` package needs to be installed (use ``pip install sunpy[all,s3]`` or ``conda install sunpy s3fs boto3`` boto3 is currently needed for reading CDF files).
+
+Drawing the extent of a WCS
+===========================
+
+The extent of any WCS with two celestial axes can now be drawn on a `~astropy.visualization.WCSAxes` using
+:func:`sunpy.visualization.drawing.extent`.
+Additionally, :meth:`sunpy.map.GenericMap.draw_extent` has been added as a convenience method for drawing the
+extent of any map.
+
+.. minigallery:: sunpy.visualization.drawing.extent
--- a/docs/whatsnew/6.1.rst
+++ b/docs/whatsnew/6.1.rst
@@ -59,7 +59,7 @@ You will need to install packages for your desired filesystem so for S3 the ``s3
 Drawing the extent of a WCS
 ===========================
 
-The extent of any WCS with two celestial axes can now be drawn on a `~astropy.visualization.WCSAxes` using
+The extent of any WCS with two celestial axes can now be drawn on a `~astropy.visualization.wcsaxes.WCSAxes` using
 :func:`sunpy.visualization.drawing.extent`.
 Additionally, :meth:`sunpy.map.GenericMap.draw_extent` has been added as a convenience method for drawing the
 extent of any map.
--- a/sunpy/visualization/drawing.py
+++ b/sunpy/visualization/drawing.py
@@ -201,7 +201,7 @@ def extent(axes, wcs, *, rsun: u.m = R_sun, **kwargs):
 
     Parameters
     ----------
-    axes : `matplotlib.axes.Axes`
+    axes : `astropy.visualization.wcsaxes.WCSAxes`
         The axes to plot the prime meridian on, or "None" to use current axes.
     wcs : `~astropy.wcs.WCS`
         The WCS that defines the extent to be drawn.
--- a/sunpy/visualization/drawing.py
+++ b/sunpy/visualization/drawing.py
@@ -197,34 +197,45 @@ def prime_meridian(axes, *, rsun: u.m = R_sun, resolution=500, **kwargs):
 @u.quantity_input
 def extent(axes, wcs, *, rsun: u.m = R_sun, resolution=1000, **kwargs):
     """
-    Draws the extent of a given `~astropy.wcs.WCS`.
+    Draws the extent as defined by a given `~astropy.wcs.WCS` instance.
 
     Parameters
     ----------
+    axes : `matplotlib.axes.Axes`
+        The axes to plot the prime meridian on, or "None" to use current axes.
+    wcs : `~astropy.wcs.WCS`
+        The WCS that defines the extent to be drawn.
+    rsun : `~astropy.units.Quantity`
+        Solar radius (in physical length units) at which to draw the solar
+        prime meridian. Defaults to the standard photospheric radius.
+    resolution : `int`
+        The number of points used to represent the prime meridian.
+
+    Returns
+    -------
+    visible : `~matplotlib.patches.Polygon`
+        The patch added to the axes for the visible part of the WCS extent.
+    hidden : `~matplotlib.patches.Polygon`
+        The patch added to the axes for the hidden part of the WCS extent.
     """
     if not wcsaxes_compat.is_wcsaxes(axes):
         raise ValueError('axes must be a WCSAxes')
     axes_frame = wcsapi_to_celestial_frame(axes.wcs)
     shape = wcs.pixel_shape
-    # Find the corners in pixel space of the WCS
+    # Define the corners of the WCS in pixel space starting at the
+    # lower left corner and moving clockwise
     corners = np.array([(0, 0),
                         (0, shape[0]-1),
                         (shape[1]-1, shape[0]-1),
                         (shape[1]-1, 0)])
-    # Find the edges in pixel space of the WCS
-    edges = np.array([(corners[0], corners[1]),
-                      (corners[1], corners[2]),
-                      (corners[2], corners[3]),
-                      (corners[3], corners[0])])
-    edges = np.array([(np.linspace(*e[:,0], resolution), np.linspace(*e[:,1],resolution))
-                      for e in edges])
-    visible = []
-    hidden = []
-    for edge in edges:
-        edge_coord = wcs.pixel_to_world(*edge)
-        _visible, _hidden = _plot_vertices(edge_coord, axes, axes_frame, rsun, close_path=False, **kwargs)
-        visible.append(_visible)
-        hidden.append(_hidden)
+    xy_edges = []
+    for i in range(corners.shape[0]):
+        i_next = 0 if i == (len(corners)-1) else i+1
+        # NOTE: exclude last entry to avoid duplicate entries at corners
+        xy_edges += [[np.linspace(*corners[[i,i_next],0], resolution)[:-1],
+                      np.linspace(*corners[[i,i_next],1], resolution)[:-1]]]
+    edge_coords = wcs.pixel_to_world(*np.hstack(xy_edges))
+    visible, hidden = _plot_vertices(edge_coords, axes, axes_frame, rsun, close_path=False, **kwargs)
     return visible, hidden
 
 
--- a/sunpy/visualization/drawing.py
+++ b/sunpy/visualization/drawing.py
@@ -220,23 +220,11 @@ def extent(axes, wcs, *, rsun: u.m = R_sun, **kwargs):
         raise ValueError('axes must be a WCSAxes')
     axes_frame = wcsapi_to_celestial_frame(axes.wcs)
     shape = wcs.pixel_shape
-    # Define the corners of the WCS in pixel space.
-    # These are, in order:
-    # 1. Lower left corner of the lower left pixel
-    # 2. Upper left corner of the upper left pixel
-    # 3. Upper right corner of the upper right pixel
-    # 4. Lower right corner of the lower right pixel
-    corners = np.array([(-0.5, -0.5),
-                        (-0.5, shape[0]-0.5),
-                        (shape[1]-0.5, shape[0]-0.5),
-                        (shape[1]-0.5, -0.5)])
-    xy_edges = []
-    for i in range(corners.shape[0]):
-        i_next = (i + 1) % corners.shape[0]
-        n_pixels = np.fabs(np.diff(corners[[i,i_next],:],axis=0)).max().astype(int)
-        # Exclude last point to avoid double counting corners
-        xy_edges += [[np.linspace(*corners[[i,i_next],0], n_pixels-1, endpoint=False),
-                      np.linspace(*corners[[i,i_next],1], n_pixels-1, endpoint=False)]]
+    # Traverse the edges of the WCS in pixel space
+    xy_edges = [[np.full(shape[1], -0.5), np.arange(shape[1]) - 0.5],  # left edge
+                [np.arange(shape[0]) - 0.5, np.full(shape[0], shape[1] - 0.5)],  # top edge
+                [np.full(shape[1], shape[0] - 0.5), np.arange(shape[1], 0, -1) - 0.5],  # right edge
+                [np.arange(shape[0], 0, -1) - 0.5, np.full(shape[0], -0.5)]]  # bottom edge
     edge_coords = wcs.pixel_to_world(*np.hstack(xy_edges))
     visible, hidden = _plot_vertices(edge_coords, axes, axes_frame, rsun, close_path=True, **kwargs)
     return visible, hidden
--- a/sunpy/visualization/drawing.py
+++ b/sunpy/visualization/drawing.py
@@ -247,14 +247,13 @@ def _plot_vertices(coord, axes, frame, rsun, close_path=True, **kwargs):
 
     # Get the 2D vertices of the coordinates
     coord = coord.transform_to(frame)
-    # If we have a HPC frame ensure that it's 3D
-    if isinstance(coord.frame, Helioprojective):
-        coord = coord.frame.make_3d()
     Tx = coord.spherical.lon.to_value(u.deg)
     Ty = coord.spherical.lat.to_value(u.deg)
     vertices = np.array([Tx, Ty]).T
-    # Determine which points are visible
-    if hasattr(frame, 'observer'):
+
+    # Determine which points are visible (2D points are always visible)
+    is_2d = (norm := coord.spherical.norm()).unit is u.one and u.allclose(norm, 1*u.one)
+    if not is_2d and hasattr(frame, 'observer'):
         # The reference distance is the distance to the limb for the axes
         # observer
         rsun = getattr(frame, 'rsun', rsun)
--- a/sunpy/visualization/drawing.py
+++ b/sunpy/visualization/drawing.py
@@ -9,6 +9,7 @@ import astropy.units as u
 from astropy.constants import R_sun
 from astropy.coordinates import SkyCoord
 from astropy.visualization.wcsaxes.wcsapi import wcsapi_to_celestial_frame
+from astropy.wcs.wcsapi import BaseHighLevelWCS
 
 from sunpy.coordinates import HeliographicCarrington, HeliographicStonyhurst
 from sunpy.coordinates.frames import HeliocentricInertial, Helioprojective
@@ -215,14 +216,19 @@ def extent(axes, wcs, **kwargs):
     """
     if not wcsaxes_compat.is_wcsaxes(axes):
         raise ValueError('axes must be a WCSAxes')
+    if not isinstance(wcs, BaseHighLevelWCS):
+        raise TypeError("wcs should be a High Level WCS object")
     axes_frame = wcsapi_to_celestial_frame(axes.wcs)
-    shape = wcs.pixel_shape
+    shape = wcs.low_level_wcs.pixel_shape
     # Traverse the edges of the WCS in pixel space
     xy_edges = [[np.full(shape[1], -0.5), np.arange(shape[1]) - 0.5],  # left edge
                 [np.arange(shape[0]) - 0.5, np.full(shape[0], shape[1] - 0.5)],  # top edge
                 [np.full(shape[1], shape[0] - 0.5), np.arange(shape[1], 0, -1) - 0.5],  # right edge
                 [np.arange(shape[0], 0, -1) - 0.5, np.full(shape[0], -0.5)]]  # bottom edge
     edge_coords = wcs.pixel_to_world(*np.hstack(xy_edges))
+    # Filter out any non-SkyCoord coordinates returned for these pixel axes
+    if not isinstance(edge_coords, SkyCoord):
+        edge_coords = [c for c in edge_coords if isinstance(c, SkyCoord)][0]
     visible, hidden = _plot_vertices(edge_coords, axes, axes_frame, R_sun, close_path=True, **kwargs)
     return visible, hidden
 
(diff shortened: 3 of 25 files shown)


wtbarnes:
Docs build and figure test devdeps fails are unrelated.

sunpy/visualization/tests/test_drawing.py
```diff
@@ -18,6 +18,21 @@ def aia171_test_map():
     return Map(get_test_filepath('aia_171_level1.fits'))
 
 
+@pytest.fixture
+def cropped_aia193_sample_map():
+    import sunpy.data.sample
+    m = sunpy.map.Map(sunpy.data.sample.AIA_193_JUN2012)
```
nabobalis, line 24: What is this sample data file?!

wtbarnes: https://docs.sunpy.org/en/stable/generated/gallery/map_transformations/reprojection_aia_euvi_mosaic.html#sphx-glr-generated-gallery-map-transformations-reprojection-aia-euvi-mosaic-py

ayshih (commented):
- I think the map method should be locked to plotting that map's own extent on the specified axes.  I would not accept a potentially unrelated WCS as an argument.
- Instead of using the `resolution` approach, I think it would be better to literally use the pixel corners.  That way, the segmentation of the extent box will not be mismatched from pixelization.  (There can be a different parameter for increasing the sampling points along a pixel edge, but the default would be just the corners.)

wtbarnes:
> * I think the map method should be locked to plotting that map's own extent on the specified axes.  I would not accept a potentially unrelated WCS as an argument.

Agreed. This is also more consistent with the way the other draw methods work, i.e. they "draw themselves". I don't know why I did it backwards. Are you opposed to keeping the function in `sunpy.visualizations.drawing` which takes in any WCS and tries to its extent on an axis?

>     * Instead of using the `resolution` approach, I think it would be better to literally use the pixel corners.  That way, the segmentation of the extent box will not be mismatched from pixelization.  (There can be a different parameter for adding sampling points along a pixel edge, but the default would be just the corners.)

I see what you mean re: pixelization. However, aren't the four corners alone insufficient to adequately define the map extent? For example, in your original issue, the red and green boxes you show have the same corner coordinates even though the FOVs are different. Am I missing something?

ayshih:
> > * I think the map method should be locked to plotting that map's own extent on the specified axes.  I would not accept a potentially unrelated WCS as an argument.
> 
> Agreed. This is also more consistent with the way the other draw methods work, i.e. they "draw themselves". I don't know why I did it backwards. Are you opposed to keeping the function in `sunpy.visualizations.drawing` which takes in any WCS and tries to its extent on an axis?

Nope, not opposed as it is consistent with the limb stuff.

> > * Instead of using the `resolution` approach, I think it would be better to literally use the pixel corners.  That way, the segmentation of the extent box will not be mismatched from pixelization.  (There can be a different parameter for adding sampling points along a pixel edge, but the default would be just the corners.)
> 
> I see what you mean re: pixelization. However, aren't the four corners alone insufficient to adequately define the map extent? For example, in your original issue, the red and green boxes you show have the same corner coordinates even though the FOVs are different. Am I missing something?

To clarify, I meant the outer corners of _every single_ edge pixel, so 2N+2M(+1) points for an NxM array.  For that example image, I think I may have called `sunpy.map.maputils.map_edges()`.

sunpy/visualization/drawing.py
```diff
@@ -194,6 +194,51 @@ def prime_meridian(axes, *, rsun: u.m = R_sun, resolution=500, **kwargs):
...
+    if not wcsaxes_compat.is_wcsaxes(axes):
+        raise ValueError('axes must be a WCSAxes')
+    axes_frame = wcsapi_to_celestial_frame(axes.wcs)
+    shape = wcs.pixel_shape
+    # Define the corners of the WCS in pixel space starting at the
+    # lower left corner and moving clockwise
+    corners = np.array([(0, 0),
+                        (0, shape[0]-1),
+                        (shape[1]-1, shape[0]-1),
+                        (shape[1]-1, 0)])
```
ayshih, line 230: I'd argue these should range from `-0.5` to `shape[index] - 0.5`

wtbarnes:
I've removed the resolution argument and constructed the edges such that the spacing along any given edge has a resolution of 1 pixel, with the left endpoint included and the right endpoint excluded as the latter is the starting point of the subsequent edge. 

The reason I didn't use `sunpy.map.map_edges` was to avoid tying `sunpy.visualizations.drawing.extent` to `GenericMap` and instead just accept a WCS. I suppose we could refactor `map_edges` to use some more general function for finding the extent of the WCS and then use that in both functions.

sunpy/visualization/drawing.py
```diff
@@ -194,6 +194,54 @@ def prime_meridian(axes, *, rsun: u.m = R_sun, resolution=500, **kwargs):
     return visible, hidden
 
 
+@u.quantity_input
+def extent(axes, wcs, *, rsun: u.m = R_sun, **kwargs):
```
Cadair, line 198: Without having looked at this in detail, this feels like something which could be useful to upstream to astropy?

sunpy/visualization/drawing.py
```diff
@@ -194,6 +194,54 @@ def prime_meridian(axes, *, rsun: u.m = R_sun, resolution=500, **kwargs):
...
+    """
+    Draws the extent as defined by a given `~astropy.wcs.WCS` instance.
+
+    Parameters
+    ----------
+    axes : `matplotlib.axes.Axes`
```
Cadair, line 204:
```suggestion
    axes : `astropy.visualization.wcsaxes.WCSAxes`
```

sunpy/visualization/drawing.py
```diff
@@ -194,6 +194,54 @@ def prime_meridian(axes, *, rsun: u.m = R_sun, resolution=500, **kwargs):
...
+    # 2. Upper left corner of the upper left pixel
+    # 3. Upper right corner of the upper right pixel
+    # 4. Lower right corner of the lower right pixel
+    corners = np.array([(-0.5, -0.5),
+                        (-0.5, shape[0]-0.5),
+                        (shape[1]-0.5, shape[0]-0.5),
+                        (shape[1]-0.5, -0.5)])
+    xy_edges = []
+    for i in range(corners.shape[0]):
+        i_next = (i+1)%corners.shape[0]
```
Cadair, line 235:
```suggestion
        i_next = (i + 1) % corners.shape[0]
```

sunpy/visualization/tests/test_drawing.py
```diff
@@ -54,6 +69,14 @@ def test_heliographic_equator_prime_meridian(heliographic_test_map):
     drawing.prime_meridian(axes, color="red")
 
 
+@figure_test
+def test_draw_extent(cropped_aia193_sample_map, euvi195_sample_map):
+    fig = plt.figure()
+    ax = fig.add_subplot(projection=euvi195_sample_map)
+    euvi195_sample_map.plot(axes=ax)
+    drawing.extent(ax, cropped_aia193_sample_map.wcs)
```
Cadair, line 77:
```suggestion
    drawing.extent(ax, cropped_aia193_sample_map.wcs)
    return fig
```

sunpy/visualization/drawing.py
Cadair, line 198: Why do we need rsun here? I get that draw_vertices uses it, but why do we need it here? What does it *dooo*?

sunpy/visualization/drawing.py
```diff
@@ -194,6 +194,54 @@ def prime_meridian(axes, *, rsun: u.m = R_sun, resolution=500, **kwargs):
...
+@u.quantity_input
+def extent(axes, wcs, *, rsun: u.m = R_sun, **kwargs):
+    """
+    Draws the extent as defined by a given `~astropy.wcs.WCS` instance.
+
+    Parameters
+    ----------
+    axes : `matplotlib.axes.Axes`
+        The axes to plot the prime meridian on, or "None" to use current axes.
+    wcs : `~astropy.wcs.WCS`
```
Cadair, line 206:
```suggestion
    wcs : `~astropy.wcs.wcsapi.BaseLowLevelWCS`
```
?

Cadair (commented):
I like this very much.

Cadair (changes requested):
Minor changes then good

sunpy/visualization/tests/test_drawing.py
wtbarnes: Why? None of the other tests return `fig`

sunpy/visualization/drawing.py
wtbarnes: It's used to determine visibility in `_plot_vertices`. @ayshih is going to love this...

https://github.com/sunpy/sunpy/blob/4842174abcdbbb7fd2e634475329ce048cec37d3/sunpy/visualization/drawing.py#L218-L219

wtbarnes: Presumably this should be refactored to just use https://docs.sunpy.org/en/stable/generated/api/sunpy.coordinates.Helioprojective.html#sunpy.coordinates.Helioprojective.is_visible?

hayesla:
this looks great! I had a play around with it and found a weird bug

I was trying it to draw FOV of solar orbiter instruments on FSI - and it seems to have an issue with the SPICE map:

example_extent


to reproduce:
```python
fig = plt.figure(figsize=(12, 6))

ax1 = fig.add_subplot(1, 2, 1, projection=m_fsi_zoom)
ax2 = fig.add_subplot(1, 2, 2, projection=m_fsi_zoom)

m_fsi_zoom.plot(axes=ax1, title='With draw_quadrangle')
m_fsi_zoom.plot(axes=ax2, title='With drawings.extent')


maps = [(hri_map, 'HRI', 'C0'),
        (phi_map, 'PHI', 'C1'),
        (spice_map, 'SPICE', 'C2')]

# old way I was using with draw_quadrangle, and using pixels so that rotation angle taken into account
for m, label, color in maps:
    m_fsi_zoom.draw_quadrangle(
        [0, 0] * u.pix,
        top_right=u.Quantity(m.dimensions),
        label=label,
        edgecolor=color,
        lw=2,
        transform=ax1.get_transform(m.wcs), axes=ax1
    )

# trying with new PR
for m, label, color in maps:
    drawing.extent(ax2, m.wcs, color=color, label=label, lw=2)

ax1.legend()
ax2.legend()
```

it could be the way the SPICE map is or something, but funny that it works for draw_quadrangle way

sunpy/visualization/drawing.py
```diff
@@ -194,6 +194,54 @@ def prime_meridian(axes, *, rsun: u.m = R_sun, resolution=500, **kwargs):
...
+                        (-0.5, shape[0]-0.5),
+                        (shape[1]-0.5, shape[0]-0.5),
+                        (shape[1]-0.5, -0.5)])
+    xy_edges = []
+    for i in range(corners.shape[0]):
+        i_next = (i + 1) % corners.shape[0]
+        n_pixels = np.fabs(np.diff(corners[[i,i_next],:],axis=0)).max().astype(int)
+        # Exclude last point to avoid double counting corners
+        xy_edges += [[np.linspace(*corners[[i,i_next],0], n_pixels-1, endpoint=False),
+                      np.linspace(*corners[[i,i_next],1], n_pixels-1, endpoint=False)]]
```
ayshih, line 239: * This is nigh unreadable =P
* You have an off-by-one error for splitting the edge back into pixels
* You used `wcs.pixel_shape` (as opposed to `wcs.array_shape`), so the shape is in (x, y) order, but then your code treats it as if it's in (y, x) order, hence the bad plotting for rectangular extents

Anyway, here's a rewrite:
```suggestion
    # Traverse the edges of the WCS in pixel space
    xy_edges = [[np.full(shape[1], -0.5), np.arange(shape[1]) - 0.5],  # left edge
                [np.arange(shape[0]) - 0.5, np.full(shape[0], shape[1] - 0.5)],  # top edge
                [np.full(shape[1], shape[0] - 0.5), np.arange(shape[1], 0, -1) - 0.5],  # right edge
                [np.arange(shape[0], 0, -1) - 0.5, np.full(shape[0], -0.5)]]  # bottom edge
```

sunpy/visualization/drawing.py
```diff
@@ -194,6 +194,54 @@ def prime_meridian(axes, *, rsun: u.m = R_sun, resolution=500, **kwargs):
...
 
+@u.quantity_input
+def extent(axes, wcs, *, rsun: u.m = R_sun, **kwargs):
+    """
+    Draws the extent as defined by a given `~astropy.wcs.WCS` instance.
+
+    Parameters
+    ----------
+    axes : `astropy.visualization.wcsaxes.WCSAxes`
+        The axes to plot the prime meridian on, or "None" to use current axes.
```
ayshih, line 205: You didn't edit this after copying it!

sunpy/visualization/drawing.py
```diff
@@ -194,6 +194,54 @@ def prime_meridian(axes, *, rsun: u.m = R_sun, resolution=500, **kwargs):
...
+    wcs : `~astropy.wcs.wcsapi.BaseLowLevelWCS`
+        The WCS that defines the extent to be drawn.
+    rsun : `~astropy.units.Quantity`
+        Solar radius (in physical length units) at which to draw the solar
+        prime meridian. Defaults to the standard photospheric radius.
```
ayshih, line 210: You didn't edit this after copying it, but see also my comment about removing `rsun` as an argument

sunpy/visualization/drawing.py
ayshih: I'm pretty sure that the `rsun` functionality in this entire module is not working as intended...  Fixing it is beyond the scope of this PR, but I'd omit `rsun` here as an argument for `extent()` until the underlying machinery is fixed.

changelog/7851.feature.rst
```diff
@@ -0,0 +1,2 @@
+Add `sunpy.visualization.drawing.extent` to visualize the extent of another WCS on an axis.
```
ayshih, line 1:
```suggestion
Add :func:`sunpy.visualization.drawing.extent` to visualize the extent of another WCS on an axis.
```

changelog/7851.feature.rst
```diff
@@ -0,0 +1,2 @@
...
+Add `sunpy.map.GenericMap.draw_extent` to visualize the extent of a map on a given axis.
```
ayshih, line 2:
```suggestion
Add :meth:`sunpy.map.GenericMap.draw_extent` to visualize the extent of a map on a given axis.
```

sunpy/visualization/tests/test_drawing.py
ayshih: My preference is to return `fig`, and to *also* instantiate `fig` without using pyplot: `fig = Figure()`

ayshih (changes requested):

sunpy/visualization/tests/test_drawing.py
```diff
@@ -18,6 +17,21 @@ def aia171_test_map():
     return Map(get_test_filepath('aia_171_level1.fits'))
 
 
+@pytest.fixture
+def cropped_aia193_sample_map():
+    import sunpy.data.sample
+    m = sunpy.map.Map(sunpy.data.sample.AIA_193_JUN2012)
+    return m.submap(SkyCoord(Tx=300*u.arcsec, Ty=-450*u.arcsec,frame=m.coordinate_frame),
+                    width=500*u.arcsec,
+                    height=500*u.arcsec)
```
ayshih, line 26: How about changing one of these significantly so that the extent is not square?

wtbarnes: Good idea 😛

Cadair:
So this seems to error if you try and plot the extent of a submap over it's origin map. As far as I can tell this is because the hpc coords are never made 3D.

```
In [1]: import matplotlib.pyplot as plt

In [2]: fig = plt.figure()

In [3]: import sunpy.map

In [4]: from sunpy.data.test import get_test_filepath

In [5]: fname = get_test_filepath("aia_171_level1.fits")

In [6]: sm = sunpy.map.Map(fname)

In [7]: crop = sm.submap([50, 50]*u.pixel, top_right=[80, 80]*u.pixel)
NameError                                 Traceback (most recent call last)
Cell In[7], line 1
----> 1 crop = sm.submap([50, 50]*u.pixel, top_right=[80, 80]*u.pixel)

NameError: name 'u' is not defined

In [8]: import astropy.units as u

In [9]: crop = sm.submap([50, 50]*u.pixel, top_right=[80, 80]*u.pixel)

In [10]: sm.plot()
Out[10]: <matplotlib.image.AxesImage at 0x7f64cf59f380>

In [11]: crop.draw_extent()
UnitConversionError                       Traceback (most recent call last)
File ~/.virtualenvs/sunpy-dev/lib/python3.13/site-packages/astropy/units/quantity_helper/helpers.py:89, in get_converters_and_unit(f, unit1, unit2)
     88 try:
---> 89     converters[changeable] = get_converter(unit2, unit1)
     90 except UnitsError:

File ~/.virtualenvs/sunpy-dev/lib/python3.13/site-packages/astropy/units/quantity_helper/helpers.py:38, in get_converter(from_unit, to_unit)
     37 try:
---> 38     converter = from_unit.get_converter(to_unit)
     39 except AttributeError as exc:
     40     # Check for lack of unit only now, to avoid delay for cases where a unit
     41     # was present. Note that cases where dimensionless is expected are
     42     # already short-circuited; here, we cover just the case where, e.g., the
     43     # user has done u.add_enabled_equivalencies(u.dimensionless_angles()).

File ~/.virtualenvs/sunpy-dev/lib/python3.13/site-packages/astropy/units/core.py:1133, in UnitBase.get_converter(self, other, equivalencies)
   1131                 return lambda v: b(converter(v))
-> 1133 raise exc

File ~/.virtualenvs/sunpy-dev/lib/python3.13/site-packages/astropy/units/core.py:1116, in UnitBase.get_converter(self, other, equivalencies)
   1115 try:
-> 1116     return self._apply_equivalencies(
   1117         self, other, self._normalize_equivalencies(equivalencies)
   1118     )
   1119 except UnitsError as exc:
   1120     # Last hope: maybe other knows how to do it?
   1121     # We assume the equivalencies have the unit itself as first item.
   1122     # TODO: maybe better for other to have a `_back_converter` method?

File ~/.virtualenvs/sunpy-dev/lib/python3.13/site-packages/astropy/units/core.py:1067, in UnitBase._apply_equivalencies(self, unit, other, equivalencies)
   1065 other_str = get_err_str(other)
-> 1067 raise UnitConversionError(f"{unit_str} and {other_str} are not convertible")

UnitConversionError: 'm' (length) and '' (dimensionless) are not convertible

During handling of the above exception, another exception occurred:

UnitConversionError                       Traceback (most recent call last)
Cell In[11], line 1
----> 1 crop.draw_extent()

File ~/Git/sunpy/sunpy/map/mapbase.py:2414, in GenericMap.draw_extent(self, axes, **kwargs)
   2411 import sunpy.visualization.drawing
   2413 axes = self._check_axes(axes)
-> 2414 return sunpy.visualization.drawing.extent(
   2415     axes,
   2416     self.wcs,
   2417     **kwargs
   2418 )

File ~/.virtualenvs/sunpy-dev/lib/python3.13/site-packages/astropy/units/decorators.py:316, in QuantityInput.__call__.<locals>.wrapper(*func_args, **func_kwargs)
    314 # Call the original function with any equivalencies in force.
    315 with equiv_context:
--> 316     return_ = wrapped_function(*func_args, **func_kwargs)
    318 # Return
    319 ra = wrapped_signature.return_annotation

File ~/Git/sunpy/sunpy/visualization/drawing.py:226, in extent(axes, wcs, **kwargs)
    221 xy_edges = [[np.full(shape[1], -0.5), np.arange(shape[1]) - 0.5],  # left edge
    222             [np.arange(shape[0]) - 0.5, np.full(shape[0], shape[1] - 0.5)],  # top edge
    223             [np.full(shape[1], shape[0] - 0.5), np.arange(shape[1], 0, -1) - 0.5],  # right edge
    224             [np.arange(shape[0], 0, -1) - 0.5, np.full(shape[0], -0.5)]]  # bottom edge
    225 edge_coords = wcs.pixel_to_world(*np.hstack(xy_edges))
--> 226 visible, hidden = _plot_vertices(edge_coords, axes, axes_frame, R_sun, close_path=True, **kwargs)
    227 return visible, hidden

File ~/Git/sunpy/sunpy/visualization/drawing.py:253, in _plot_vertices(coord, axes, frame, rsun, close_path, **kwargs)
    251     rsun = getattr(frame, 'rsun', rsun)
    252     reference_distance = np.sqrt(frame.observer.radius**2 - rsun**2)
--> 253     is_visible = coord.spherical.distance <= reference_distance
    254 else:
    255     # If the axes has no observer, the entire limb is considered visible
    256     is_visible = np.ones_like(coord.spherical.distance, bool, subok=False)

File ~/.virtualenvs/sunpy-dev/lib/python3.13/site-packages/astropy/units/quantity.py:683, in Quantity.__array_ufunc__(self, function, method, *inputs, **kwargs)
    681     return NotImplemented
    682 else:
--> 683     raise e

File ~/.virtualenvs/sunpy-dev/lib/python3.13/site-packages/astropy/units/quantity.py:628, in Quantity.__array_ufunc__(self, function, method, *inputs, **kwargs)
    623 # Determine required conversion functions -- to bring the unit of the
    624 # input to that expected (e.g., radian for np.sin), or to get
    625 # consistent units between two inputs (e.g., in np.add) --
    626 # and the unit of the result (or tuple of units for nout > 1).
    627 try:
--> 628     converters, unit = converters_and_unit(function, method, *inputs)
    630     out = kwargs.get("out")
    631     # Avoid loop back by turning any Quantity output into array views.

File ~/.virtualenvs/sunpy-dev/lib/python3.13/site-packages/astropy/units/quantity_helper/converters.py:177, in converters_and_unit(function, method, *args)
    174 units = [getattr(arg, "unit", None) for arg in args]
    176 # Determine possible conversion functions, and the result unit.
--> 177 converters, result_unit = ufunc_helper(function, *units)
    179 if any(converter is False for converter in converters):
    180     # for multi-argument ufuncs with a quantity and a non-quantity,
    181     # the quantity normally needs to be dimensionless, *except*
   (...)
    184     # can just have the unit of the quantity
    185     # (this allows, e.g., `q > 0.` independent of unit)
    186     try:
    187         # Don't fold this loop in the test above: this rare case
    188         # should not make the common case slower.

File ~/.virtualenvs/sunpy-dev/lib/python3.13/site-packages/astropy/units/quantity_helper/helpers.py:299, in helper_twoarg_comparison(f, unit1, unit2)
    298 def helper_twoarg_comparison(f, unit1, unit2):
--> 299     converters, _ = get_converters_and_unit(f, unit1, unit2)
    300     return converters, None

File ~/.virtualenvs/sunpy-dev/lib/python3.13/site-packages/astropy/units/quantity_helper/helpers.py:91, in get_converters_and_unit(f, unit1, unit2)
     89     converters[changeable] = get_converter(unit2, unit1)
     90 except UnitsError:
---> 91     raise UnitConversionError(
     92         f"Can only apply '{f.__name__}' function to quantities "
     93         "with compatible dimensions"
     94     )
     96 return converters, unit1

UnitConversionError: Can only apply 'less_equal' function to quantities with compatible dimensions```

**Comment by Cadair** [comment]
I tired to use this with DKIST gwcses and I need this patch for that: https://github.com/wtbarnes/sunpy/pull/29/files

I am not sure how I would write a comprehensive test for this, so have made it a PR for now.

File: sunpy/visualization/drawing.py [hunk 14]
```diff
@@ -208,6 +241,9 @@ def _plot_vertices(coord, axes, frame, rsun, close_path=True, **kwargs):
 
     # Get the 2D vertices of the coordinates
     coord = coord.transform_to(frame)
+    # If we have a HPC frame ensure that it's 3D
+    if isinstance(coord.frame, Helioprojective):
+        coord = coord.frame.make_3d()
```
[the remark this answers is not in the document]
ayshih: I just pushed a better fix

sunpy/visualization/tests/test_drawing.py
```diff
@@ -18,6 +22,21 @@ def aia171_test_map():
...
+                    width=500*u.arcsec,
+                    height=250*u.arcsec)
+
+
+@pytest.fixture
+def euvi195_sample_map():
+    import sunpy.data.sample
+    return Map(sunpy.data.sample.STEREO_A_195_JUN2012)
+
+
```
ayshih, line 39: The figure test shouldn't use sample data (and potentially triggering downloading).  It'd be just as good to use `aia171_test_map` and a comparable map with the observer information modified.

sunpy/visualization/tests/test_drawing.py
```diff
@@ -31,31 +50,43 @@ def heliographic_test_map():
...
     drawing.prime_meridian(axes, color="red")
+    return fig
+
+
+@figure_test
+def test_draw_extent(cropped_aia193_sample_map, euvi195_sample_map):
+    fig = Figure()
+    ax = fig.add_subplot(projection=euvi195_sample_map)
+    euvi195_sample_map.plot(axes=ax)
+    drawing.extent(ax, cropped_aia193_sample_map.wcs)
```
ayshih, line 84: Actually, this unit tests shouldn't even use a Map-based WCS since that could conceivably obscure a bug

sunpy/visualization/tests/test_drawing.py
wtbarnes: I'm now using a non-map WCS so this isn't an issue.

Cadair (approved):

ayshih (approved):
This new functionality warrants a gallery example and a "what's new" entry, but those can be in follow-on PRs.  Let's get this in!

nabobalis:
> This new functionality warrants a gallery example and a "what's new" entry, but those can be in follow-on PRs. Let's get this in!

We can just do it in this PR.

docs/whatsnew/6.1.rst
```diff
@@ -55,3 +55,13 @@ Support has been added to ``sunpy.map.Map`` and ``sunpy.timeseries.TimeSeries``
 Support for S3 filesystems is tested, but other fsspec plugins should work for most filetypes, please open issues if errors are encountered with other filesystems.
 
 You will need to install packages for your desired filesystem so for S3 the ``s3fs`` package needs to be installed (use ``pip install sunpy[all,s3]`` or ``conda install sunpy s3fs boto3`` boto3 is currently needed for reading CDF files).
+
+Drawing the extent of a WCS
+===========================
+
+The extent of any WCS with two celestial axes can now be drawn on a `~astropy.visualization.WCSAxes` using
```
Cadair, line 62:
```suggestion
The extent of any WCS with two celestial axes can now be drawn on a `~astropy.visualization.wcsaxes.WCSAxes` using
```

Cadair:
The devdeps fails are due to timeouts on anaconda.org.

PR #7854, opened 2024-10-28
Use updated astropy.Table api

SolarDrew:
## PR Description

`Table.pformat_all()` is deprecated and it's breaking things. This updates attr to use the new `pformat`.


wtbarnes:
It looks like several of the doctests are now failing because the output of `pformat` is slightly different.

nabobalis (approved):

SolarDrew:
> It looks like several of the doctests are now failing because the output of `pformat` is slightly different.

Serves me right for not checking. Thanks for the fix @Cadair :+1:

PR #7863, opened 2024-10-31
Redesign our scheduled builds

Cadair:
I am trying to address https://github.com/sunpy/package-template/issues/98 and just generally come up with a sane way to roll out cron jobs / extra CI to all the sponsored packages.

To that end, this is my current suggestion, if it works here then I will add it to the template.

This PR does the following things:

* Moves the cron jobs out into it's own workflow file, the main advantage of doing this is that we can trigger it when a label is added to the pull request rather than only triggering on another event and then running if the label is present.
* Moves the notify job out into it's own workflow, which is run after either of the CI and Cron workflows finish. This means that it wont conflict with the template but also means that we will now get two messages for each branch in the matrix rooms (one for each workflow).
* Changes the way we trigger scheduled builds on main to use the `workflow_run` event rather than sharing the way we run scheduled builds on release branches which uses a PAT from a bot user. The main advantage of this is that we can roll out template cron jobs to all the repos which don't use release branches without having to worry about templates etc.

---

We could also choose to spin out the publish job into it's own workflow file which means we could trigger "Run Publish" on label add as well, but then we would get *three* different messages in the matrix room.


nabobalis:
> We could also choose to spin out the publish job into it's own workflow file which means we could trigger "Run Publish" on label add as well, but then we would get _three_ different messages in the matrix room.

It feels like we should at this point?

Cadair:
> It feels like we should at this point?

The other issue now I think about it is that we are only going to be able to conditionally run publish if the *whole* CI workflow passes, which isn't what we do at the moment.

nabobalis:
> > It feels like we should at this point?
> 
> The other issue now I think about it is that we are only going to be able to conditionally run publish if the _whole_ CI workflow passes, which isn't what we do at the moment.

You mean even the online build?

Cadair:
> On the topic of the bot messages, did you not create a github events room? Just move the messages there.

I think that's likely to get them ignored forever.

nabobalis:
> > On the topic of the bot messages, did you not create a github events room? Just move the messages there.
> 
> I think that's likely to get them ignored forever.

Doesn't seem like much of a difference to now? But fair point. 

There is no way to batch the messages in another workflow?

nabobalis:
> > You mean even the online build?
> 
> yeah

That might be a problem.

Cadair:
> There is no way to batch the messages in another workflow?

I mean not easily. It might be possible to store some state in the matrix event and add more reactions to the same message, but that sounds like a good amount of dev work. Other than that it's run a webservice to do the notifications.

nabobalis:
> > There is no way to batch the messages in another workflow?
> 
> I mean not easily. It might be possible to store some state in the matrix event and add more reactions to the same message, but that sounds like a good amount of dev work. Other than that it's run a webservice to do the notifications.

That just seems like work for work's sake. People will have to live with multiple messages. That's my vote.

nabobalis (approved):

.github/workflows/ci.yml
```diff
@@ -15,6 +15,9 @@ on:
   pull_request:
   # Allow manual runs through the web UI
   workflow_dispatch:
+  # Trigger on completion of the scheduled_builds.yml file (only on main)
+  workflow_run:
+    workflows: [Scheduled build triggerer]
```
Cadair, line 20: Ok, here's a question I can't quite decide what option is the least-worst.

As currently implemented here, this triggers when the `scheduled_builds` workflow finishes. This works well for this repo, as we now only have one `scheduled` trigger in that workflow, which triggers everything. However, I want to implement this in the template and I am wondering if having separate `schedule` triggers in the `ci.yml` and `cron.yml` would be more accessible for projects which do not use release branches (almost all of them) and we should just suck it up in sunpy/sunpy and sunpy/ndcube and have three separate, duplicated `schedule` triggers.

nabobalis: Will those projects eventually move to release branches? 

If no, ndcube/sunpy can have repeated triggers.

Cadair: > Will those projects eventually move to release branches?

I mean maybe eventually. Only if they support more than one release at a time.

The alternative is that we mirror what we have here into the package template, but the `scheduled_builds` workflow is a no-op which is just used to trigger the other workflows?

nabobalis: No op sounds ok to me

PR #7867, opened 2024-11-08
Use a fixed time in the past for the gallery example with STEREO locations

ayshih:
The gallery example that reproduces the "Where is STEREO?" plot can no longer plot the location of STEREO-B because ephemeris information is no longer generated (communication with STEREO-B was lost over a decade ago).  This PR changes the example so that instead of generating the plot for "now", it generates it for the time of the sunpy 6.0.3 docs.


nabobalis (approved):

Cadair:
Is it worthy adding a note that it won't work for more recent times?

nabobalis:
Added

PR #7878, opened 2024-11-12
Use main instead of master for figure tests comparison 

nabobalis:
The path for the figures should be main?!


pytest.ini
```diff
@@ -23,7 +23,7 @@ doctest_optionflags =
 addopts =
     --doctest-rst
     -p no:unraisableexception
-    -p no:theadexception
+    -p no:threadexception
```
nabobalis, line 26: TYPO

Cadair (approved):

PR #7889, opened 2024-11-28
Fixes CI - Not Online

nabobalis:
Small normal tweaks

Unsure what to do about the online tests and the JSOC.

PR #7892, opened 2024-12-02
Fix cache configuration for tests against astropy 7

ayshih:
Astropy 7 switched to using pathlib.Path in internal configuration, but we provide strings in our configuration for testing


ayshih:
Rebasing this on #7889 so that CI will even run the online build

nabobalis (approved):

PR #7899, opened 2024-12-05
Adds `show_in_notebook` function for Interactive tables using `Ipydatagrid`

### LINKED ISSUE #7639: Make the attrs table searchable
### Describe the feature

Astropy Table lets you search through the results in a notebook. 

By default, our tables do not have this.
The call is a different one but there must be a way to hook into it and display it like that in a notebook. 

See https://docs.astropy.org/en/stable/api/astropy.table.Table.html#astropy.table.Table.show_in_notebook

### Proposed solution

Make it work.

Prtm2110:
## PR Description

Fixes #7639

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,6 +58,10 @@ asdf = [
   "asdf-astropy>=0.4.0",
   "asdf>=2.13.0",
 ]
+tables = [
+  "ipydatagrid>=1.3.2",
+  "ipython>=8.0.0"
+]
 dask = ["dask[array]>=2022.5.2"]
 image = [
   # !=1.10.0 due to https://github.com/scipy/scipy/issues/17718
@@ -101,7 +105,7 @@ visualization = [
   "mpl-animators>=1.0.0",
 ]
 core = ["sunpy[image,map,net,timeseries,visualization]"]
-all = ["sunpy[core,asdf,jpeg2000,opencv,spice,scikit-image]"]
+all = ["sunpy[core,asdf,jpeg2000,opencv,spice,scikit-image,tables]"]
 tests-only = [
   "hvpy>=1.1.0",
   "jplephem>=2.14",
--- a/sunpy/net/fido_factory.py
+++ b/sunpy/net/fido_factory.py
@@ -214,6 +214,50 @@ class UnifiedResponse(Sequence):
         """
         return type(self)(*[i.show(*cols) for i in self._list])
 
+
+    def show_in_notebook(self, **kwargs):
+        """
+        Display the tables in the current object as interactive grids in a Jupyter Notebook.
+
+        This function utilizes the ``ipydatagrid`` library to render tables as interactive grids.
+        Each table in the current object is converted to a ``pandas.DataFrame``, passed to an
+        ``ipydatagrid.DataGrid`` instance, and displayed directly in the notebook.
+
+        .. note::
+            This function requires the optional dependencies ``pandas``, ``IPython`` and ``ipydatagrid``.
+            Ensure these are installed before calling this method.
+
+        Parameters
+        ----------
+        **kwargs : dict, optional
+            Additional keyword arguments to customize the ``ipydatagrid.DataGrid`` instance.
+            For available options, use ``help(ipydatagrid.DataGrid)`` in a Jupyter Notebook.
+
+        Raises
+        ------
+        ValueError
+            If no tables are available to display.
+
+        Returns
+        -------
+        None
+            The method displays the grids directly in the notebook and does not return anything.
+        """
+        from ipydatagrid import DataGrid
+        from IPython.display import display
+
+        if not self._list:
+            raise ValueError("No tables to display.")
+
+        for table in (self._list):
+            # Identify and exclude multidimensional columns
+            valid_columns = [name for name in table.colnames if len(table[name].shape) <= 1]
+            filtered_table = table[valid_columns]
+            df = filtered_table.to_pandas()
+            grid = DataGrid(df, **kwargs)
+            display(grid)
+
+
     @property
     def all_colnames(self):
         """
--- a/sunpy/net/tests/test_fido.py
+++ b/sunpy/net/tests/test_fido.py
@@ -184,6 +184,14 @@ def test_unifiedresponse_slicing():
 
 
 @pytest.mark.remote_data
+def test_ipydatagrid_table():
+    results = Fido.search(
+        a.Time("2012/1/1", "2012/1/2"), a.Instrument.lyra)
+    assert isinstance(results, UnifiedResponse)
+    assert isinstance(results.show_in_notebook(), None)
+
+
+@pytest.mark.remote_data
 def test_unifiedresponse_slicing_reverse():
     results = Fido.search(
         a.Time("2012/1/1", "2012/1/2"), a.Instrument.lyra)


sunpy/net/fido_factory.py
```diff
@@ -214,6 +214,50 @@ def show(self, *cols):
...
+        Returns
+        -------
+        None
+            The method displays the grids directly in the notebook and does not return anything.
+        """
+        from ipydatagrid import DataGrid
+        from IPython.display import display
+
+        if not self._list:
+            raise ValueError("No tables to display.")
```
nabobalis, line 250: Why should this error?

sunpy/net/tests/test_fido.py
```diff
@@ -173,6 +174,14 @@ def test_unifiedresponse_slicing():
     assert isinstance(results[0], QueryResponseTable)
 
 
+@pytest.mark.remote_data
+def test_ipydatagrid_table():
+    results = Fido.search(
+        a.Time("2012/1/1", "2012/1/2"), a.Instrument.lyra)
+    assert isinstance(results, UnifiedResponse)
+    assert isinstance(results.show_in_notebook(), NoneType)
```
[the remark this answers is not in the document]
Prtm2110: As the function isn't returning anything this can be used to test it, I will look into improving the tests

pyproject.toml
```diff
@@ -58,6 +58,10 @@ asdf = [
   "asdf-astropy>=0.4.0",
   "asdf>=2.13.0",
 ]
+tables = [
+  "ipydatagrid>=1.3.2",
+  "ipython>=8.0.0"
+]
```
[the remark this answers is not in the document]
nabobalis: These then should be moved to test dependencies and the imports handled inside that function as a very optional dependency.

sunpy/net/fido_factory.py
Prtm2110: If `_list` is empty, this raises an error to inform the user of this exact issue. Should I remove it, now I think its unnecessary?

nabobalis: If its empty, why should the code raise an error?

Prtm2110:
[itables](https://github.com/mwouts/itables) can be a great alternative here, as it eliminates the need for additional dependencies. Here's how you can use it:

```python
from itables import show
show(df)
```

If you prefer to avoid the IPython dependency, you can consider the following approach:  
Example Screenshot

Additionally, do you have any suggestions for writing tests for this integration?

nabobalis:
> [itables](https://github.com/mwouts/itables) can be a great alternative here, as it eliminates the need for additional dependencies. Here's how you can use it:

I assume astropy picked the other library for a reason?

What is the difference between them?

Also is this an AI answer?

> Additionally, do you have any suggestions for writing tests for this integration?

Honestly, I do not. How did astropy do it?

sunpy/net/fido_factory.py
```diff
@@ -214,6 +214,53 @@ def show(self, *cols):
         """
         return type(self)(*[i.show(*cols) for i in self._list])
 
+
+    def show_in_notebook(self, **kwargs):
```
nabobalis, line 218: Why can we not use the astropy version of this method?

Prtm2110:
> I assume astropy picked the other library for a reason?
> 
I believe maybe they didn't pick itables because "itables appears to change how dataframes are rendered Notebook-wide which I felt isn't something we're going for." said by exitflynn, although this is not a problem now because itable have `show(df)` method which only converts particular dataframes rather than all.

> What is the difference between them?

If we use itables we can directly use their show method which can display them in jupyter notbook, but in astropy they have used ipydatagrid to single tables and returned them directly, which we cannot do in sunpy as there can be multiple tables in a fido result, also this is the reason why we cannot use astropy's show_in_notebook method.

> Also is this an AI answer?
 
Neither the code nor the comments are ai (although I use grammarly), I guess it is a problem? I will stop that now.
> > Additionally, do you have any suggestions for writing tests for this integration?
> 
> Honestly, I do not. How did astropy do it?
> 
> Since astropy are dealing with these types of dependencies, I think investigating how they do that and copying their solution is the simplest path forward.

Yes, they just checked `isInstance(t.show_in_table(), Datagrid)` but again this wouldn't work here, because we are not returning Datagrid as they are.

nabobalis:
> > I assume astropy picked the other library for a reason?
> 
> I believe maybe they didn't pick itables because "itables appears to change how dataframes are rendered Notebook-wide which I felt isn't something we're going for." said by exitflynn, although this is not a problem now because itable have `show(df)` method which only converts particular dataframes rather than all.

How does it change them compared to the other library?

> > What is the difference between them?
> 
> If we use itables we can directly use their show method which can display them in jupyter notbook, but in astropy they have used ipydatagrid to single tables and returned them directly, which we cannot do in sunpy as there can be multiple tables in a fido result, also this is the reason why we cannot use astropy's show_in_notebook method.

So how does both libraries handle having multiple tables and rendering them in a notebook?

> > Also is this an AI answer?
> 
> Neither the code nor the comments are ai (although I use grammarly), I guess it is a problem? I will stop that now.

No that's fine. It's a common tool and its different from Gen AI.

> > > Additionally, do you have any suggestions for writing tests for this integration?
> > 
> > 
> > Honestly, I do not. How did astropy do it?
> > Since astropy are dealing with these types of dependencies, I think investigating how they do that and copying their solution is the simplest path forward.
> 
> Yes, they just checked `isInstance(t.show_in_table(), Datagrid)` but again this wouldn't work here, because we are not returning Datagrid as they are.

Hmm, I expect that properly testing this kind of output is painful and why it was not done. Can you check how the `quicklook` method is tested inside sunpy?

Prtm2110:
> > > I assume astropy picked the other library for a reason?
> > 
> > 
> > I believe maybe they didn't pick itables because "itables appears to change how dataframes are rendered Notebook-wide which I felt isn't something we're going for." said by exitflynn, although this is not a problem now because itable have `show(df)` method which only converts particular dataframes rather than all.
> 
> How does it change them compared to the other library?
> 
`itable` has this feature as well
```
from itables import init_notebook_mode

init_notebook_mode(all_interactive=True)
```
This converts all dataframes into those interactive tables, this is what exitflynn meant although i guess he missed or they added `show(df)` new method later, which makes changes to particular dataframes only 
> > > What is the difference between them?
> > 
> > 
> > If we use itables we can directly use their show method which can display them in jupyter notbook, but in astropy they have used ipydatagrid to single tables and returned them directly, which we cannot do in sunpy as there can be multiple tables in a fido result, also this is the reason why we cannot use astropy's show_in_notebook method.
> 
> So how does both libraries handle having multiple tables and rendering them in a notebook?
> 

Naturally they cannot render multiple tables, but `itable` renders the tables in distinct boxes with search button (image in the comment above) which looks good.

> > > > Additionally, do you have any suggestions for writing tests for this integration?
> > > 
> > > 
> > > Honestly, I do not. How did astropy do it?
> > > Since astropy are dealing with these types of dependencies, I think investigating how they do that and copying their solution is the simplest path forward.
> > 
> > 
> > Yes, they just checked `isInstance(t.show_in_table(), Datagrid)` but again this wouldn't work here, because we are not returning Datagrid as they are.
> 
> Hmm, I expect that properly testing this kind of output is painful and why it was not done. Can you check how the `quicklook` method is tested inside sunpy?

Yes I had a look, I will try using mocking this will take some time though.

nabobalis:
> Naturally they cannot render multiple tables, but `itable` renders the tables in distinct boxes with search button (image in the comment above) which looks good.

So in your opinion, which library should we use?
 
> Yes I had a look, I will try using mocking.

Thanks

Prtm2110:
> > Naturally they cannot render multiple tables, but `itable` renders the tables in distinct boxes with search button (image in the comment above) which looks good.
> 
> So in your opinion, which library should we use?
> 

itables

Prtm2110:
Also can you please clarify what you mean by `attrs table`, and how they are generated, a code example maybe?

nabobalis:
> Also can you please clarify what you mean by `attrs table`, and how they are generated, a code example maybe?

https://docs.sunpy.org/en/stable/tutorial/acquiring_data/index.html#searching-for-data

Prtm2110:
> > Also can you please clarify what you mean by `attrs table`, and how they are generated, a code example maybe?
> 
> https://docs.sunpy.org/en/stable/tutorial/acquiring_data/index.html#searching-for-data

image
Is the position and the name of search function acceptable? I have created a `_create_table` function to reduce redundancy of code

nabobalis:
Looks fine with me.

changelog/7899.feature.rst
```diff
@@ -0,0 +1 @@
+Added :meth:`sunpy.net.fido_factory.UnifiedResponse.show_in_notebook` to render fido search results and :meth:`sunpy.net.attr.AttrMeta.search` to render attrs tables in form of interactive table using ``itables`` in Jupyter notebooks.
```
nabobalis, line 1: Why the difference in names?

Prtm2110: `a.Instrument.search()` just makes more sense as the only reason we are rendering it using `itables` is because we are trying to search but `show_in_notebook` can be mostly used to make the render pretty and I also believe these names makes them easy to remember.

nabobalis: Does search work in a python terminal? It sounds like it should.

Prtm2110: Oh, okay so shall I rename it to 'show_in_notebook' as well?

nabobalis: Yes please.

Prtm2110: Yes added, can you help me with the failing ref :meth:`sunpy.net.attr.AttrMeta.show_in_notebook`, is it wrong?

nabobalis: The Metaclass is not documented in the API: https://docs.sunpy.org/en/stable/reference/net.html#id38

Prtm2110: Thank you, added.

sunpy/net/attr.py
```diff
@@ -142,6 +160,37 @@ def _repr_html_(self):
...
 
 
+    def show_in_notebook(self, **kwargs):
+        """
+        Display the attrs tables as interactive grids in a Jupyter Notebook.
+
+        This function utilizes the ``itables`` library to render tables as interactive grids.
+
+        .. note::
+            This function requires the optional dependency ``itables``.
```
nabobalis, line 170: Can we not sphinx interlink this package in the docs/conf.py?

sunpy/net/attr.py
```diff
@@ -142,6 +160,37 @@ def _repr_html_(self):
...
+
+        Parameters
+        ----------
+        **kwargs : dict, optional
+            Additional keyword arguments to customize the ``itables.show`` function.
+
+        Returns
+        -------
+        None
+            The method displays the tables directly in the notebook and does not return anything.
```
nabobalis, line 181:
```suggestion
```

sunpy/net/fido_factory.py
```diff
@@ -214,6 +214,43 @@ def show(self, *cols):
...
+
+        Parameters
+        ----------
+        **kwargs : dict, optional
+            Additional keyword arguments to customize the ``itables.show`` function.
+
+        Returns
+        -------
+        None
+            The method displays the tables directly in the notebook and does not return anything.
```
nabobalis, line 237:
```suggestion
```

sunpy/net/attr.py
Prtm2110: As expected its not working (commit I just made), my guess is because the documentation of itables is in different from the other ones like the home/landing page is a quick starter guide i.e. they don't have "stable" documentation as it is a github.io documentation, I could be wrong here.

docs/conf.py
```diff
@@ -228,6 +228,7 @@
     "sunkit_instruments": ("https://docs.sunpy.org/projects/sunkit-instruments/en/stable/", None),
     "zeep": ("https://docs.python-zeep.org/en/stable/", None),
     "contourpy": ("https://contourpy.readthedocs.io/en/stable/", None),
+    "itables": ("https://mwouts.github.io/itables/quick_start.html", None),
```
nabobalis, line 231:
```suggestion
    "itables": ("https://mwouts.github.io/itables/objects.inv", None),
```
This might work, not sure.

Prtm2110: Still not working, should I revert it back?

nabobalis: Yeah

sunpy/net/attr.py
Prtm2110, line 1: This remains unchanged and is still resulting in a document failure. I don’t quite understand why

nabobalis: I would just add one more ` to each of them so they are now code blocks.

Prtm2110: Added, anything else needs attention?

nabobalis: Not for now.

pyproject.toml
```diff
@@ -119,6 +119,7 @@ tests-only = [
   "pytest-mpl>=0.16",
   "pytest-xdist>=3.0.2",
   "pytest>=7.1.0",
+  "itables>=2.2.4",
```
nabobalis, line 122: We will need to decide if this gets its own category and if we install it via "all" as well.

sunpy/net/fido_factory.py
```diff
@@ -214,6 +214,38 @@ def show(self, *cols):
...
+        **kwargs : dict, optional
+            Additional keyword arguments to customize the ``itables.show`` function.
+
+        """
+        try:
+            from itables import show
+        except ImportError:
+            raise ImportError(
+                "`itables` is required to display tables. "
+                "Install itables using `pip install itables`."
```
nabobalis, line 239: I forgot about conda.

sunpy/net/attr.py
```diff
@@ -142,6 +160,33 @@ def _repr_html_(self):
...
+        **kwargs : dict, optional
+            Additional keyword arguments to customize the ``itables.show`` function.
+
+        """
+        try:
+            from itables import show
+        except ImportError:
+            raise ImportError(
+                "`itables` is required to display tables. "
+                "Install itables using `pip install itables`."
```
nabobalis, line 184: I forgot about conda

Prtm2110: Added.

pyproject.toml
Prtm2110: Are there any decisions that have been made?

nabobalis: No there has not been one.

sunpy/net/fido_factory.py
```diff
@@ -214,6 +214,38 @@ def show(self, *cols):
...
+            raise ImportError(
+                "`itables` is required to display tables. "
+                "Install itables using `pip install itables` or `conda install conda-forge::itables`."
+            )
+        for table in self._list:
+            # Identify and exclude multidimensional columns
+            valid_columns = [name for name in table.colnames if len(table[name].shape) <= 1]
+            filtered_table = table[valid_columns]
+            df = filtered_table.to_pandas()
+            show(df, **kwargs)
```
Cadair, line 246: I think it would be a nice addition to add a [caption](https://mwouts.github.io/itables/advanced_parameters.html#caption) to each of these tables with some information like what's in the `__str__` method. I.e. what client are they from etc.

pyproject.toml
```diff
@@ -116,6 +116,7 @@ tests-only = [
...
```
Cadair, line 119: I wonder if it would be worth combining this with `ipywidgets` (an optional dependency for parfive) in a `jupyter` extra we don't include in `all`?

Cadair (commented):
Thanks a lot for working on this @Prtm2110 sorry it took me so long to review it.

I have a couple of relatively minor comments, but overall I think this is a really useful addition. Please let us know if you still have time to work on it, thanks!

sunpy/net/fido_factory.py
```diff
@@ -214,6 +214,38 @@ def show(self, *cols):
         """
         return type(self)(*[i.show(*cols) for i in self._list])
 
+
+    def show_in_notebook(self, **kwargs):
```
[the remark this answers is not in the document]
Prtm2110: Sure that would be great, done!

sunpy/net/tests/test_fido.py
```diff
@@ -173,6 +173,15 @@ def test_unifiedresponse_slicing():
     assert isinstance(results[0], QueryResponseTable)
 
 
+@pytest.mark.remote_data
+def test_show_in_notebook(mocker):
+    results = Fido.search(a.Time('2012/1/1', '2012/1/2'), a.Instrument.aia | a.Instrument.hmi)
+    assert isinstance(results, UnifiedResponse)
+    mock_datagrid =  mocker.patch("itables.show")
+    results.show_in_notebook()
+    assert mock_datagrid.call_count == 2
```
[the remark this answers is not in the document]
Prtm2110: I shall investigate this.

wtbarnes:
This also looks good to me. Thanks for working on this! Two things I'm still not clear on though:

1. Why are we using `itables` over `ipydatagrid`? Both seem relatively popular but the former is maintained by a single person and the latter is maintained by the Jupyter project.
2. How are we handling the `itables`/`ipydatagrid` dependency? Is an additional optional dependency category for notebook-related things being added?

Prtm2110:
> 1. Why are we using `itables` over `ipydatagrid`? Both seem relatively popular but the former is maintained by a single person and the latter is maintained by the Jupyter project.

Hi, ipydatagrid felt much slower and more laggy when I used it, which is why we decided to go with tables. [Ref comment](https://github.com/sunpy/sunpy/pull/7899#issuecomment-2526351035)

> 2. How are we handling the `itables`/`ipydatagrid` dependency? Is an additional optional dependency category for notebook-related things being added?

I'm not quite sure could you please let me know where the dependency should be placed

Cadair:
> 2\. How are we handling the `itables`/`ipydatagrid` dependency? Is an additional optional dependency category for notebook-related things being added?

I've now added an optional `jupyter` extra which also has `ipywidgets` in it for the parfive progress bars.

changelog/7899.feature.rst
```diff
@@ -0,0 +1 @@
+Added :meth:`sunpy.net.fido_factory.UnifiedResponse.show_in_notebook` to render fido search results and :meth:`~sunpy.net.attr.AttrMeta.show_in_notebook` on all attrs to render attrs tables in form of interactive table using ``itables`` in Jupyter notebooks.
```
wtbarnes:
```suggestion
Added :meth:`sunpy.net.fido_factory.UnifiedResponse.show_in_notebook` to render `~sunpy.net.Fido` search results and :meth:`~sunpy.net.attr.AttrMeta.show_in_notebook` on all attrs to render attrs tables in form of interactive table using ``itables`` in Jupyter notebooks.
```

sunpy/net/tests/test_fido.py
```diff
@@ -174,6 +174,24 @@ def test_unifiedresponse_slicing():
     assert isinstance(results[0], QueryResponseTable)
 
 
+@pytest.mark.remote_data
```
alasdairwilson: I know we have random requirements on remote data but do we really need to use a real fido return for this

Cadair: Yeah I agree, there's no need for this to be an online PR.

docs/tutorial/installation.rst
```diff
@@ -132,6 +132,7 @@ Now that we have a fresh virtual environment, we can proceed with installing ``s
     $ conda install sunpy
 
 This will install ``sunpy`` and all of its dependencies.
+If you are planning on using sunpy in jupyter notebooks we also recommend you install the `ipywidgets` and `itables` packages.
```
Cadair:
```suggestion
If you are planning on using sunpy in jupyter notebooks we also recommend you install the ``ipywidgets`` and ``itables`` packages.
```

nabobalis:
Changed test to be offline.

Cadair (approved):

nabobalis:
Thank you again @Prtm2110

PR #7903, opened 2024-12-06
Fixes Regex to Retain Leading Zeros in Time Format

### LINKED ISSUE #7897: Issue with parse time regex dropping 0's on minutes/seconds
### Describe the bug

```
parse_time("202201021230")
 <Time object: scale='utc' format='isot' value=2022-01-02T12:03:00.000>

```

however, 
```
parse_time("20220102123000")
 <Time object: scale='utc' format='isot' value=2022-01-02T12:30:00.000>
```

### To Reproduce

import sunpy

sunpy.map.Map(...)
etc

### Screenshots

_No response_

### System Details

_No response_

### Installation method

_No response_

PredictiveManish:
Fixes #7897 
This PR addresses an issue with the parse_time function where it incorrectly parses time strings with a 4-digit minute and 2-digit second format, leading to the dropping of seconds when the input has 6 digits.

Changes:

- Modified the time parsing regex to preserve leading zeros in minutes and seconds.
- Modified the time parsing regex to handle both 4-digit minutes and 6-digit time formats correctly, ensuring that seconds are retained even when the time string includes the minute and second part in a combined format (e.g., "202201021230" vs "20220102123000").
- Ensured consistent behavior when handling time strings with single-digit minutes/seconds (e.g., "09:05" instead of "9:5").

This fix ensures that both formats are parsed correctly, preserving accurate minute and second information.


nabobalis:
Tests are needed for the new behavior. Using the examples from the original issue should be sufficient.

sunpy/extern/parse.py
```diff
@@ -169,7 +169,7 @@ def __eq__(self, other):
 DAYS_PAT = r"(Mon|Tue|Wed|Thu|Fri|Sat|Sun)"
```
nabobalis, line 169: This file should not be touched, all the changes here need to be reverted.

PredictiveManish: OK, will add tests and change as per the instructions.

PR #7911, opened 2024-12-07
Adds  support for the timestamp %Y%m%d%H%M

ankitkhushwaha:
## PR Description

this fixes [#7897](https://github.com/sunpy/sunpy/issues/7897)
https://github.com/sunpy/sunpy/blob/8c35f7e0fdf46d77faaaa34a1a47e440332326dc/sunpy/time/time.py#L37

user is expected to give proper input as shown in the TIME_FORMAT_LIST 

some of inputs are
 `20140101000001`
`20070504_210812`
`20070504T210812`
`20070504T210812.999999`

in this pr i implemented input validation for parse_time for  above examples

@nabobalis waiting for the suggestions.


PredictiveManish (commented):
Also can make the changes into the format structure for allowing 12 digits format which helps to catch that format easily. %Y%m%d%H%M

iakkhushwaha:
@nabobalis  does `%Y%m%d%H%M` format is allowed for the TIME_FORMAT_LIST ?

If it does, we would need to place it above this line
https://github.com/sunpy/sunpy/blob/8c35f7e0fdf46d77faaaa34a1a47e440332326dc/sunpy/time/time.py#L63

to ensure the `%Y%m%d%H%M` format is matched first. The same would apply for the `%Y%m%dT%H%M` format.

I don't think this is reliable approach rather than tricky one.
In the future this things might breaks if someone else change the order  of items of  TIME_FORMAT_LIST

nabobalis:
Is that the only way? It would be nice to support the timestamps in the original issue.

ankitkhushwaha:
> Is that the only way? It would be nice to support the timestamps in the original issue.

yes we can add support to the timestamps 

but i have one doubt if 

```
parse_time('20280401121')
<Time object: scale='utc' format='isot' value=2028-04-01T12:01:00.000>

```
But  
      
```
parse_time('20280401121000')
<Time object: scale='utc' format='isot' value=2028-04-01T12:10:00.000>

```
but if user want time in format of '%Y%m%d%H%M` we can add support to it anyways

However we dont know user expection what desired time format he wants 
so best way to address the issue to put input validation for that format.  I think so .

if There is better approach in your mind i will be happy to implement it

nabobalis:
> ```
> parse_time('20280401121')
> <Time object: scale='utc' format='isot' value=2028-04-01T12:01:00.000>
> ```

These seems like a format we shouldn't support, its missing one digit.

> ```
> parse_time('20280401121000')
> <Time object: scale='utc' format='isot' value=2028-04-01T12:10:00.000>
> ```
> 
> but if user want time in format of `%Y%m%d%H%M` we can add support to it anyways

I think we should, that seems to be clear enough.

> However we dont know user expection what desired time format he wants so best way to address the issue to put input validation for that format. I think so .

I don't follow.

ankitkhushwaha:
I think it is fine now

sunpy/time/tests/test_time.py
```diff
@@ -211,6 +211,13 @@ def test_parse_time_ISO():
     dt4 = Time('2007-05-04T21:08:00')
     dt5 = Time('2007-05-04')
 
+    assert (
+        parse_time('20070504210812') == dt3 
+    )
+    assert (
+        parse_time('200705042108') == dt4 
+    )
```
nabobalis, line 219:
```suggestion
    assert parse_time('20070504210812') == dt3
    assert parse_time('200705042108') == dt4
```

sunpy/time/tests/test_time.py
```diff
@@ -238,7 +247,6 @@ def test_parse_time_ISO():
         assert is_time_equal(dt, v)
         assert dt.format == 'isot'
```
nabobalis, line 249:
```suggestion


```

sunpy/time/time.py
```diff
@@ -229,6 +232,12 @@ def convert_time_list(time_list, format=None, **kwargs):
 def convert_time_str(time_string, **kwargs):
     if 'TAI' in time_string:
         kwargs['scale'] = 'tai'
+    
+    # if time_string does not have proper format r'20021204([T_])?242012(\.\d{6})?' format 
+    if re.match(r'\d{8}([T_])?' , time_string):                         
+        if not re.match(r'\d{8}([T_])?\d{6}(\.\d{6})?' , time_string):
+            message = 'Input values did not match the format of astro.time.time'
+            raise ValueError(message)
```
nabobalis: Is this being used or tested by anything?

nabobalis:
I think this needs a changelog, a bugfix one.

sunpy/time/time.py
ankitkhushwaha: I have removed it few minutes ago

sunpy/time/time.py
```diff
@@ -41,6 +41,7 @@
     "%Y-%m-%dT%H:%M:%S",  # Example 2007-05-04T21:08:12
```
nabobalis, line 41: I can't add the line suggestion but can you add a comment to this list that it must never be sorted.

changelog/7991.bugfix.rst
```diff
@@ -0,0 +1 @@
+Added support within `~.parse_time` for the timestamp ``%Y%m%d%H%M``.
```
nabobalis, line 1:
```suggestion
Added support within `~.parse_time` for the timestamp ``%Y%m%d%H%M`` and ``%Y%m%d_%H%M``.
```
I forgot about the other one.

sunpy/time/time.py
```diff
@@ -34,13 +34,15 @@
     '%f': r'(?P<microsecond>\d+)',
     '%b': r'(?P<month_str>[a-zA-Z]+)',
 }
+# Do NOT sort this list. The String Parsing is dependent on the specific order within this list.
```
nabobalis, line 37:
```suggestion
# DO NOT SORT THIS LIST.
# The string parsing is dependent on the specific order within this list.
```

sunpy/time/time.py
ankitkhushwaha: I have made the changes.

I have one question ,
can we specify in a comment which elements have an order dependency?

nabobalis (approved):

changelog/7991.bugfix.rst
ankitkhushwaha: I have also added the ``%Y%m%dT%H%M`` in changelog file

sunpy/time/time.py
nabobalis: If you would like.

sunpy/time/time.py
```diff
@@ -34,13 +34,27 @@
     '%f': r'(?P<microsecond>\d+)',
     '%b': r'(?P<month_str>[a-zA-Z]+)',
 }
+# DO NOT SORT THIS LIST.
+# The string parsing is dependent on the specific order within this list.
 TIME_FORMAT_LIST = [
+    """
```
nabobalis, line 40: Why was this added to the list?

nabobalis: Also could this not be a comment with the format it affects?

ankitkhushwaha: I thought this would be best way to explain,
now i have wrote a more concise explanation

nabobalis:
pre-commit.ci autofix

nabobalis:
Thanks for the PR @ankitkhushwaha

PR #7912, opened 2024-12-08
Improve Eclipse Calculations Using Lunar DEM Data

### LINKED ISSUE #7809: Consider improving the accuracy of our eclipse calculations

Currently our eclipse calculations assume that the Moon is a smooth sphere, with two options for the radius to account for the valleys in the lunar limb.  We could consider using lunar altimeter data (e.g., from LRO) to be more accurate, as is done by modern eclipse software.  See also this just-published paper for further inspiration: [Wright & Young (2024), "A Raster-oriented Method for Creating Eclipse Maps"](https://doi.org/10.3847/1538-3881/ad6b23).

naxatra2:
Fixes #7809 
--- 
introduced mathematical improvements to the `eclipse_amount` function incorporating dynamic lunar limb profiles derived from high-resolution Digital Elevation Model (DEM) data (DEM Data should be given by the user)

I have almost only used the given [research paper](https://iopscience.iop.org/article/10.3847/1538-3881/ad6b23#ajad6b23s3) ( given in the issue) as reference 

---

### **Key Changes**
   - Added the `limb_profile_from_dem` function to compute a lunar limb profile using DEM data. This function calculates angular radii at equally spaced azimuthal intervals around the lunar limb.
   - Updated the `eclipse_amount` , which utilizes the computed lunar limb profile to calculate eclipse amounts dynamically.

---

#### **`limb_profile_from_dem` Function**:
This function calculates a lunar limb profile as a one-dimensional array of angular radii. It processes DEM data (latitude, longitude, elevation) to compute:

`limb_profile_from_dem(dem_data, observer_distance, resolution=18000)`
   1. **dem_data**:
      - Lunar DEM data, formatted as [(latitude, longitude, elevation)]
   2. **observer_distance**:
      - Distance of the observer from the Moon in km.
   3. **resolution**:
      - Number of angular intervals for the limb profile (default is 18000). Needs to be calculated be the user seperately and then give as an input.

#### **`eclipse_amount` Function**:
This function calculates the percentage of the Sun obscured by the Moon during an eclipse, using the dynamic lunar limb model:
   1. Now it also needs an additional parameter `dem_data` 

---

### **Data Source**
The SLDEM2015 dataset (More specifically I used the `SLDEM2015_DATA_QUALITY.jp2` file) was used for processing DEM data into the required format. The SLDEM2015 product combines data from the Lunar Reconnaissance Orbiter (LRO) laser altimeter (LOLA) and the Kaguya mission's terrain camera (TC). The data can be accessed at:
[SLDEM2015 Product Page](https://pgda.gsfc.nasa.gov/products/54)

I Only used the data from this source to process the `.jp2` files into (latitude, longitude, elevation) tuples for further computations. To locally test it I had to create a seperate function to process the `.jp2` file in a desirable format. I used the rasterio library for it.


naxatra2:
Even though I have opened a PR, I Still think that it might miss something mathematically. I am new to this domain. Can anybody help ??

nabobalis:
Hello @naxatra2,

Have you been able to test this?

nabobalis:
> Even though I have opened a PR, I Still think that it might miss something mathematically. I am new to this domain. Can anybody help ??

What would you like help with?

nabobalis:
Can you also layout if you used any LLM outputs and if so what was written by an LLM?

nabobalis:
> I have tested the `limb_profile_from_dem `function and it works and creates a limb profile from any input DEM data, It just needs to have a preprocessed DEM file

When you say works, you mean just runs?

What code did you use to process the DEM files?

> But didn't understand like what tests should I do to check the `eclipse_amount ` function.
> 
> image

Well we have two choices:

1. See what the improvements with this method are compared to the original example: https://docs.sunpy.org/en/stable/generated/gallery/showcase/eclipse_amount.html
2. Recreate Figure 10 from the paper
3. Recreate Figure 11 from the paper

To be honest, I am not familiar with the mathematics or even the original eclipse code so these are my 3 quick suggestions.

> I have used this part of that paper for making the limb profiles and then I have written an implementation of the functions. And in the `elipse_amount `function it had considered static radius I changed it to dynamic radius and based on this. Also while creating the limb profiles I got confused a lof so I improved it using LLM.

The issue is that the PR looks to be opened with a LLM comment, the first commit looks like it might also be LLM output. What was the code you wrote and what was the code the LLM wrote?
 
> I have never done such type of testing before. So, wanted help in the testing part of the function. Can you give a basic overview of how to do the tests

There are some basic testing guidelines here: https://docs.sunpy.org/en/latest/dev_guide/contents/tests.html
It isn't the most detailed unfortunately.

naxatra2:
> Well we have two choices:
> See what the improvements with this method are compared to the original example: https://docs.sunpy.org/en/stable/generated/gallery/showcase/eclipse_amount.html
> Recreate Figure 10 from the paper
> Recreate Figure 11 from the paper

> There are some basic testing guidelines here: 
> https://docs.sunpy.org/en/latest/dev_guide/contents/tests.html
> It isn't the most detailed unfortunately.

Okay, I will try to do this and make another commit

nabobalis:
> After running the file through this function. I used the `limb_profile_from_dem` to convert this into a limb profile. Up till this part I checked that my code is working as I intended it to be. 

But what does "working as I intended it" mean in this context? The output was correct? It was just running without any errors?
 
> I thought that I needed to make a very formal PR for doing an actual contribution. So, after making the function `limb_profile_from_dem `(without any comments and bad variable naming), I gave the whole code to chatgpt for improving and adding comments to it. Maybe that's why it looks too much AI generated. I will try not to do it next time.

Thank you.

naxatra2:
I am facing too many issues while testing and creating the limb profiles. I think my approach was wrong considering none of the tests are running. I am closing this PR because while testing I have noticed that my approach has diverged so much from the start that it doesn’t seem worth it anymore, and perhaps I don’t have enough knowledge on this topic right now. Might create a new PR later with a better approach.

nabobalis:
It is a complex issue that is for sure. Thanks for trying, maybe see about an issue which is smaller in scope?

PR #7919, opened 2024-12-10
Handle Missing FILLVAL in read_cdf for Non-Standard CDF Files

Rachit23110261:
## PR Description

Description:
This PR addresses an issue where read_cdf crashes when a CDF file lacks the FILLVAL attribute. A conditional check was added to verify the presence of FILLVAL in variable attributes before attempting to replace its values with NaN. Logs a warning when FILLVAL is missing to notify the user while maintaining functionality.

Changes:
Updated sunpy/io/_cdf.py to include a safe fallback for missing FILLVAL.
Added a warning message using Python's warnings module for better user awareness.

Testing:
All existing tests pass.

### DIFF
--- a/sunpy/io/_cdf.py
+++ b/sunpy/io/_cdf.py
@@ -89,9 +89,10 @@ def read_cdf(fname, **kwargs):
             # Set fillval values to NaN
             # It would be nice to properley mask these values to work with
             # non-floating point (ie. int) dtypes, but this is not possible with pandas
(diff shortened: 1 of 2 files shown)


nabobalis:
Thanks for the PR @Rachit23110261 but I think the one here: https://github.com/sunpy/sunpy/pull/7917is at a more advanced stage. 

Please feel free to tackle another issue.

PR #7922, opened 2024-12-10
 Adds Support for TimeSeries.truncate for out of bounds/partial TimeRanges

ankitkhushwaha:
## PR Description

This fixes [#7914](https://github.com/sunpy/sunpy/issues/7914) 

currently sunpy.timeseries.TimeSeries.truncate(TimeRange) works only for TimeRange within sunpy.timeseries.TimeSeries
But fails for any TimeRange ouside of it.

So i have added support for sunpy.timeseries.TimeSeries.truncate to work with out of bounds and partial TimeRanges 

Now for out of bounds Timeranges it will truncate to nearest time of timeseries.

PR #7936, opened 2024-12-16
Fixes `AttributeError`, Adds support to multi column feature in `GenericTimeSeries.extract` function 

### LINKED ISSUE #7932: `datetime.datetime` object has no attribute `astype` in `GenericTimeSeries._text_summary`
### Describe the bug

using the example mentioned in 

https://github.com/sunpy/sunpy/blob/3e1713f7a9f87627d80d6a9925d9227aea466293/sunpy/timeseries/timeseriesbase.py#L66

causes a error

### To Reproduce

from sunpy.timeseries import TimeSeries
from astropy.time import TimeDelta
import astropy.units as u
import numpy as np
import pandas as pd

times = parse_time("now") - TimeDelta(np.arange(24 * 60) * u.minute)
intensity = np.sin(np.arange(0, 12 * np.pi, step=(12 * np.pi) / (24 * 60)))

df = pd.DataFrame(intensity, index=times, columns=['intensity'])

header = {}
units = {'intensity': u.W / u.m**2}

ts = TimeSeries(df, header, units)
print(ts)

### Screenshots

image

### System Details

_No response_

### Installation method

conda

ankitkhushwaha:
## PR Description

This pr is extended work of #7922 

- [x] Fixes #7932
- [x] Adds  multi column support in `GenericTimeSeries.extract` function
- [ ] Adds Unit tests

### DIFF
--- a/sunpy/timeseries/timeseriesbase.py
+++ b/sunpy/timeseries/timeseriesbase.py
@@ -530,12 +530,15 @@ class GenericTimeSeries:
         `~sunpy.timeseries.TimeSeries`
             A new `~sunpy.timeseries.TimeSeries` with only the selected times.
         """
+        is_int = False
         # Evaluate inputs
         # If given strings, then use to create a sunpy.time.timerange.TimeRange
         # for the SunPy text date parser.
         if isinstance(a, str) and isinstance(b, str):
-            a = TimeRange(a, b)
-        if isinstance(a, TimeRange):
+            time_range = TimeRange(a, b)
+            start = time_range.start.datetime
+            end = time_range.end.datetime
+        elif isinstance(a, TimeRange):
             # If we have a TimeRange, extract the values
             start = a.start.datetime
             end = a.end.datetime
@@ -543,6 +546,16 @@ class GenericTimeSeries:
             # Otherwise we already have the values
             start = a
             end = b
+            is_int = True
+
+        min_time, max_time = self._data.index.min(), self._data.index.max()
+        # Check if the timerange overlaps with the data timerange
+        if not is_int:
+            if min_time >= start and max_time <= end:
+                pass
+            elif (start <= end <= min_time or max_time <= start <= end):
+                message = "Provided timerange is not within the bounds of the timeseries"
+                raise ValueError(message)
 
         # If an interval integer was given then use in truncation.
         truncated_data = self._data.sort_index()[start:end:int]
--- a/sunpy/timeseries/timeseriesbase.py
+++ b/sunpy/timeseries/timeseriesbase.py
@@ -83,6 +83,7 @@ class GenericTimeSeries:
     """
     # Class attribute used to specify the source class of the TimeSeries.
     _source = None
+    _observatory = None
     _registry = dict()
 
     # Title to show when .peek()ing
@@ -121,6 +122,8 @@ class GenericTimeSeries:
             # Should have a list of 3-tuples giving a complex metadata list.
             self.meta = meta
 
+        if isinstance(data.index[0], Time):
+            self._data.index = pd.to_datetime(self._data.index.map(lambda x: x.to_datetime()))
         if units is None:
             self.units = {}
         else:
@@ -158,7 +161,7 @@ class GenericTimeSeries:
         """
         A string/object used to specify the observatory for the TimeSeries.
         """
-        return
+        return self._observatory
 
     @property
     def columns(self):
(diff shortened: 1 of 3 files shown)


sunpy/timeseries/timeseriesbase.py
```diff
@@ -121,6 +122,8 @@ def __init__(self, data, meta=None, units=None, **kwargs):
             # Should have a list of 3-tuples giving a complex metadata list.
             self.meta = meta
 
+        if isinstance(data.index[0], Time):
+            self._data.index = pd.to_datetime(self._data.index.map(lambda x: x.to_datetime()))
```
ankitkhushwaha, line 126: elements of `self._data.index` is changed to `Pandas.Timestamp` because in

https://github.com/sunpy/sunpy/blob/782b242b06724ae315e9aad776c7673b895e3585/sunpy/timeseries/timeseriesbase.py#L249-L250

`Astropy.Time` does not support `round` function

nabobalis: Why is the index an astropy time?

ankitkhushwaha:
The example mentioned in 

https://github.com/sunpy/sunpy/blob/782b242b06724ae315e9aad776c7673b895e3585/sunpy%2Ftimeseries%2Ftimeseriesbase.py#L66

Has astropy time

nabobalis:
Hmm, I think the example needs to be changed instead so it works but at least we should change how the repr works out the time range

ankitkhushwaha:
is user expected to give `Pandas.Timestamp` or `Astropy.Time` ?
It is somewhat tricky, as user working with sunpy library will tends to use `Astropy.Time` more often.
But in some of example even in tests we are using  `Pandas.Timestamp` dataframe.

That's why i changed `self._data.index` is changed to `Pandas.Timestamp` .

nabobalis:
I think we would want both to work.

ankitkhushwaha:
I have tested the code with `pandas.TimeStamp` and `Astropy.Time`  
everything is working fine 
just i have to add some unit tests

Then should i change the example discussed in `GenericTimeSeries` and in repr ? 
any suggestion for the improvement in pr? 
 
i have also added the multi column support in `GenericTimeSeries.extract`

nabobalis:
> I have tested the code with `pandas.TimeStamp` and `Astropy.Time` everything is working fine just i have to add some unit tests
> 
> Then should i change the example discussed in `GenericTimeSeries` and in repr ? any suggestion for the improvement in pr?

Sure

> i have also added the multi column support in `GenericTimeSeries.extract`

Sounds good.

sunpy/timeseries/timeseriesbase.py
```diff
@@ -530,19 +533,32 @@ def truncate(self, a, b=None, int=None):
...
+            end = time_range.end.datetime
+        elif isinstance(a, TimeRange):
             # If we have a TimeRange, extract the values
             start = a.start.datetime
             end = a.end.datetime
         else:
             # Otherwise we already have the values
             start = a
             end = b
+            is_int = True
```
ankitkhushwaha, line 552: should we add input validation to user give only allowed inputs . `Astropy.Time` , `int`

If we use this 
```
a = pd.Timestamp('2011-06-07 06:55:59')
b = pd.Timestamp('2011-06-07 08:55:59')
```
if a , b is within time range code works fine but when not (out of bound timeranges) it cauese error

ankitkhushwaha: i think if we are using `pandas.Timestamp` in `Timeseries.data.index` then we should add also this format in input parameters.

nabobalis: We use it since astropy.time can not be a true pandas index.

sunpy/timeseries/timeseriesbase.py
```diff
@@ -544,6 +550,15 @@ def truncate(self, a, b=None, int=None):
             start = a
             end = b
 
+        min_time, max_time = self._data.index.min(), self._data.index.max()
+        # Check if the timerange overlaps with the data timerange
+        if not (isinstance(start, builtins.int) and isinstance(end, builtins.int)):
```
ankitkhushwaha, line 555: now it will also with
```
a = pd.Timestamp('2011-06-07 06:55:59')
b = pd.Timestamp('2011-06-07 08:55:59')
```

for out of bounds/partial timeranges

ankitkhushwaha:
This PR has some commit from #7922 so i thought to fix this issue after #7922 is merged

PR #7952, opened 2024-12-28
Fix: Added fallback for missing rsun_ref in GONG H-alpha data

### LINKED ISSUE #7951: The value of r_sun in the GONG Ha data does not return to the default value when the differential rotation function is used.
### Describe the bug

When I use GONG Ha data recently, I found that the ['rsun_ref'] is missing in the meta data, which causes for example the projection function can not be realized.

### To Reproduce

File C:\ProgramData\anaconda3\Lib\site-packages\sunpy\physics\differential_[rotation.py:597](http://rotation.py:597/), in differential_rotate(smap, observer, time, **diff_rot_kwargs)
594 out_meta.pop(key)
596 # Add a new HGS observer
--> 597 out_meta.update(get_observer_meta(new_observer, out_meta['rsun_ref']*u.m))
599 if is_sub_full_disk:
600 # Define a new reference pixel and the value at the reference pixel.
601 # Note that according to the FITS convention the first pixel in the
602 # image is at (1.0, 1.0).
603 center_rotated = solar_rotate_coordinate(
604 [smap.center](http://smap.center/), observer=new_observer, **diff_rot_kwargs)

File C:\ProgramData\anaconda3\Lib\site-packages\sunpy\util\[metadata.py:167](http://metadata.py:167/), in MetaDict.getitem(self, key)
163 def getitem(self, key):
164 """
165 Override [] indexing.
166 """
--> 167 return OrderedDict.getitem(self, key.lower())

KeyError: 'rsun_ref'

### Screenshots

_No response_

### System Details

_No response_

### Installation method

pip

PredictiveManish:
Fixes #7951 
This PR Resolves the issue of missing rsun_ref in GONG H-alpha data.

### DIFF
--- a/sunpy/physics/differential_rotation.py
+++ b/sunpy/physics/differential_rotation.py
@@ -510,7 +510,13 @@ def differential_rotate(smap, observer=None, time=None, **diff_rot_kwargs):
 
     # Only this function needs scikit image
     from skimage import transform
-
+    # Ensure 'rsun_ref' consistency
+    rsun_ref = smap.meta.get('rsun_ref', None)
+    if rsun_ref is None:
+        raise KeyError("The input map's metadata is missing 'rsun_ref'.")
+    
+    # Use 'rsun_ref' throughout the function
+    rsun_ref = rsun_ref * u.m
     # Check whether the input contains the full disk of the Sun
     is_sub_full_disk = not contains_full_disk(smap)
     if is_sub_full_disk:
--- a/sunpy/physics/tests/test_differential_rotation.py
+++ b/sunpy/physics/tests/test_differential_rotation.py
@@ -391,3 +391,23 @@ def test_differential_rotation(aia171_test_map):
     with pytest.warns(UserWarning, match="Using 'time' assumes an Earth-based observer"):
         rot_map = differential_rotate(aia171_test_map, time=2*u.day)
     return rot_map.data
+
+def test_rsun_ref_fallback():
+    import astropy.constants as const
+    from sunpy.map import Map
+    
+    data = np.random.rand(2,2) # Example data
+    meta = {'cdelt1': 1, 'cdelt2': 1, 'cunit1': 'arcsec', 'cunit2': 'arcsec', 'crpix1': 1, 'crpix2': 1,
+            'crval1': 0, 'crval2': 0, 'ctype1': 'HPLN-TAN', 'ctype2': 'HPLT-TAN',
+            'date-obs': '2022-01-01T00:00:00'}  # Missing 'rsun_ref'
+    smap = Map(data, meta)
+    
+    # Ensure rsun_ref is not initially in the map metadata
+    assert 'rsun_ref' not in smap.meta
+    
+    # Apply differential rotation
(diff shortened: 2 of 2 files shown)


sunpy/physics/differential_rotation.py
```diff
@@ -510,7 +510,13 @@ def differential_rotate(smap, observer=None, time=None, **diff_rot_kwargs):
...
     # Only this function needs scikit image
     from skimage import transform
-
+    # Ensure 'rsun_ref' consistency
+    rsun_ref = smap.meta.get('rsun_ref', None)
+    if rsun_ref is None:
+        raise KeyError("The input map's metadata is missing 'rsun_ref'.")
+    
+    # Use 'rsun_ref' throughout the function
+    rsun_ref = rsun_ref * u.m
```
nabobalis, line 519: Do we need all of this?

sunpy/physics/tests/test_differential_rotation.py
```diff
@@ -391,3 +391,23 @@ def test_differential_rotation(aia171_test_map):
...
+
+def test_rsun_ref_fallback():
+    import astropy.constants as const
+    from sunpy.map import Map
+    
+    data = np.random.rand(2,2) # Example data
+    meta = {'cdelt1': 1, 'cdelt2': 1, 'cunit1': 'arcsec', 'cunit2': 'arcsec', 'crpix1': 1, 'crpix2': 1,
+            'crval1': 0, 'crval2': 0, 'ctype1': 'HPLN-TAN', 'ctype2': 'HPLT-TAN',
+            'date-obs': '2022-01-01T00:00:00'}  # Missing 'rsun_ref'
+    smap = Map(data, meta)
```
nabobalis, line 403: Can we not just use an existing map fixture?

sunpy/physics/differential_rotation.py
PredictiveManish: Just for error handling..??

sunpy/physics/tests/test_differential_rotation.py
PredictiveManish: Yes will work on this.

sunpy/physics/differential_rotation.py
nabobalis: I guess my actual question is, how does this help solve the original issue? 

It does make the error clearer to users than the current behavior.
I would want some other input but having the code not error but raise a warning saying its missing and using the default value used elsewhere in sunpy seems like the simplest answer.

PredictiveManish: Ok got it, will change and update the error to warning and update for default nature.

ayshih (changes requested):
You need to use the `GenericMap.rsun_meters` property instead of accessing `meta` directly.

PredictiveManish:
> You need to use the `GenericMap.rsun_meters` property instead of accessing `meta` directly.

ok

nabobalis:
Closed in favour of https://github.com/sunpy/sunpy/pull/7953

nabobalis:
Thanks @PredictiveManish

PR #7960, opened 2025-01-02
Handle NaN/masked values in Vizier results

ayshih:
This PR fixes the example "Identifying stars in a STEREO/SECCHI COR2 coronagraph image" by properly cleaning the Vizier results of masked values.  These masked values would otherwise be passed on as NaNs, which results in those stars "disappearing" when `apply_space_motion()` is called.


nabobalis (approved):

ayshih:
## Before this PR
a
## After this PR
b
Note that the "after" image has 7 more marked stars

nabobalis:
You want to add a changelog or should I label as No?

ayshih:
Going to do something different

PR #7961, opened 2025-01-03
Fix Map's handling of PVi_m values

ayshih:
WCS projections can specify `PVi_m` parameters, but Map's handling of those values is bugged (introduced in https://github.com/sunpy/sunpy/pull/5501/commits/fb0554f543bf2e19fed2f70de47c6b4cdc309c89).  Due to an errant 'a' in a regular expression:

https://github.com/sunpy/sunpy/blob/5668bba8ae896c8147c8db4fc7be88b2b04b1fa9/sunpy/map/mapbase.py#L1494

the code looks for `PVi_mA` values (i.e., for the "A" alternative WCS description) instead of `PVi_m` values (for the primary WCS description).  This flub may have resulted from the fact that the FITS standard writes `PVi_ma` where `a` stands in for a blank character or for 'A' through 'Z'.

This bug is only apparent when the WCS projection is one that requires non-default `PVi_m` parameters and there does not exist an "A" alternative WCS description that happens to be a duplicate of the primary WCS description.

This PR also fixes an additional bug the `PVi_m` value would be misassigned when `m` is a two-digit number (i.e., 10 through 99).  For example, `PV2_10` would be assigned to `PV2_1` (and would overwrite the correct value of `PV2_1`).

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1491,12 +1491,12 @@ class GenericMap(NDData):
         """
         Return any PV values in the metadata.
         """
-        pattern = re.compile('pv[0-9]_[0-9]', re.IGNORECASE)
+        pattern = re.compile('pv[0-9]_[0-9]{1,2}$', re.IGNORECASE)
         pv_keys = [k for k in self.meta.keys() if pattern.match(k)]
 
         pv_values = []
         for k in pv_keys:
-            i, m = int(k[2]), int(k[4])
+            i, m = int(k[2]), int(k[4:])
             pv_values.append((i, m, self.meta[k]))
         return pv_values
 
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -1491,7 +1491,7 @@ class GenericMap(NDData):
         """
         Return any PV values in the metadata.
         """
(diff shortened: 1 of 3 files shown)


ayshih:
@Cadair needs to check me on this!

sunpy/map/mapbase.py
```diff
@@ -1491,12 +1491,12 @@ def _pv_values(self):
         """
         Return any PV values in the metadata.
         """
-        pattern = re.compile('pv[0-9]_[0-9]a', re.IGNORECASE)
+        pattern = re.compile('pv[0-9]_[0-9]{1,2}$', re.IGNORECASE)
```
Cadair, line 1494: Maybe I am reading the spec wrong, but there's no technical reason why you couldn't have i>9? right? if you longitude axis was the 10th in your array?

Cadair: I guess that's insanely unlikely to be the case if you are opening it with Map :laughing:

Cadair (approved):

sunpy/map/mapbase.py
ayshih: You are correct: `i` can range from 1 to 99

ayshih: Now fixed by #7985

PR #7968, opened 2025-01-03
Use 3.12 for doc-gallery builds

nabobalis:
No idea if this helps.

Check that astropy is new version on the doc builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,8 +84,7 @@ jobs:
     needs: [docs]
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@main
     with:
-      # Due to mplcario, we need to use Python 3.10 for the wheels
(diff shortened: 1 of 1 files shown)


nabobalis:
astropy==7.0.0

.github/workflows/ci.yml
```diff
@@ -101,6 +100,7 @@ jobs:
             apt:
               - libopenjp2-7
               - graphviz
+              - libcairo2-dev
```
ayshih, line 103: Why is this needed?

ayshih (approved):

.github/workflows/ci.yml
nabobalis: Good question, I needed it on 6.0 branch, as at the time cario was being compiled so I needed it.

nabobalis, line 103:
```suggestion
```

PR #7972, opened 2025-01-05
possibility to load >=3 dimensional data from CDF files

### LINKED ISSUE #6194: Add possibility to load >=3 dimensional data from CDF files
### Describe the feature

At the moment, `read_cdf()` skips data columns that are >=3 dimensional data (cf. https://github.com/sunpy/sunpy/issues/5869 & https://github.com/sunpy/sunpy/pull/5975). I was wondering if there are some possibilities to load this data without too much effort because otherwise we're missing out some nice new spacecraft data.

### Proposed solution

The crucial part is at https://github.com/sunpy/sunpy/blob/main/sunpy/io/cdf.py#L92= where a column is skipped if its dimension is > 2. Would it be a possibility to rewrite it to reflect the dependencies in the column name, like it's done 3 lines below for dim = 2 data:

``` python
elif data.ndim == 2:
    # Multiple columns, give each column a unique label
    for i, col in enumerate(data.T):
        df[var_key + f'_{i}'] = col
        units[var_key + f'_{i}'] = unit
```

So for example, for the `var_key` ELECTRON_FLUX that has (next to EPOCH) an ENERGY and VIEWING dependency, something like "Electron_Flux_Energy1_Viewing1", "Electron_Flux_Energy1_Viewing2", "Electron_Flux_Energy2_Viewing1", and so on. Then we still have a standard DataFrame that is passed to GenericTimeSeries in [L105](https://github.com/sunpy/sunpy/blob/main/sunpy/io/cdf.py#L105=)

sh-abinash-sh:
## PR Description

1. Added handling for variables with dimensions ≥3 by constructing unique column names based on their dependency variables ```(e.g., var_DEP1_val1_DEP2_val2)```.
2. Dependency variable values (```DEPEND_1```, ```DEPEND_2```, etc.) are included in column names to reflect their structure and relationships.
3. Retained existing behavior for 1D and 2D data while enhancing support for higher-dimensional data.

before this PR the code sample below :

```
>>> from sunpy.net import Fido
>>> from sunpy.net import attrs as a
>>> from sunpy.timeseries import TimeSeries

>>> # defining function to quickly try different datasets:
>>> def load_psp(dataset_name):
>>>   dataset = a.cdaweb.Dataset(dataset_name)
>>>   trange = a.Time('2021/07/01', '2021/07/02')
>>>   result = Fido.search(trange, dataset)
>>>   downloaded_files = Fido.fetch(result)
>>>   data = TimeSeries(downloaded_files, concatenate=True)
>>>   df = data.to_dataframe()
>>>   return df, data

```
Throws this warnings :

Screenshot 2025-01-05 170746

While after this PR:

the warning and issue get solved.

Screenshot 2025-01-05 171038

This fixes #6194

### DIFF
--- a/sunpy/io/_cdf.py
+++ b/sunpy/io/_cdf.py
@@ -114,8 +114,12 @@ def read_cdf(fname, **kwargs):
                 unit = u.dimensionless_unscaled
 
             if data.ndim > 2:
-                # Skip data with dimensions >= 3 and give user warning
-                warn_user(f'The variable "{var_key}" has been skipped because it has more than 2 dimensions, which is unsupported.')
+                dependencies = [attrs.get(f"DEPEND_{i}", None) for i in range(1, data.ndim)]
+                dependency_values = [cdf.varget(dep) if dep else range(data.shape[i]) for i, dep in enumerate(dependencies, 1)]
+                for comb in np.ndindex(*(len(vals) for vals in dependency_values)):
+                    column_name = f"{var_key}_" + "_".join([f"{dep}{val}" for dep, val in zip(dependencies, comb)])
+                    df_dict[column_name] = data[(slice(None),) + comb]
+                    units[column_name] = unit
             elif data.ndim == 2:
                 # Multiple columns, give each column a unique label
                 for i, col in enumerate(data.T):


nabobalis:
How is the 3D data stored in the timeseries?

sh-abinash-sh:
> How is the 3D data stored in the timeseries?
> 
> How does a user manipulate or plot the 3D data?
## This is how data 3D data stored in Times series.
1. Dependency Variables (DEPEND_1, DEPEND_2, etc.):
The DEPEND_n attributes of a variable define the axes of the multidimensional data. These dependencies (e.g., energy levels, viewing angles) are used to index the data.

2. Column Naming:
Each slice of the 3D data is assigned a unique column name constructed using:

      The variable name (e.g., var_key).
      The values of the dependency variables (e.g., Energy and Viewing).
      Example column names:
      ELECTRON_FLUX_Energy1_Viewing1
      ELECTRON_FLUX_Energy1_Viewing2
      ELECTRON_FLUX_Energy2_Viewing1
3. Flattened Representation:
The 3D data is flattened into multiple 1D columns. Each column corresponds to one specific combination of dependency values, making it easy to work with in a pandas.DataFrame.

4. Units and Metadata:
Units for each column are stored in the GenericTimeSeries units dictionary, and global attributes (metadata) are preserved.

## How does a user manipulate or plot the 3D data?

1. Users can directly manipulate 3D data by accessing specific columns.
2. Plotting is straightforward, either as individual slices, combined trends, or advanced visualizations like heatmaps and 3D plots.

nabobalis:
Can you show me all of this with examples?

nabobalis:
Thanks for the PR @sh-abinash-sh, I am going to close this for now as I think the original issue needs more thinking before we can tackle it.

PR #7975, opened 2025-01-06
Only run cron 3 times a week

Cadair:
The primary motivation for rekindling this discussion is that we are now paying for 30mins of aarch64 builder per cron run.

Discuss.

### DIFF
--- a/.github/workflows/scheduled_builds.yml
+++ b/.github/workflows/scheduled_builds.yml
@@ -9,7 +9,7 @@ on:
     #        │ │ ┌───────── day of the month (1 - 31)
     #        │ │ │ ┌───────── month (1 - 12 or JAN-DEC)
     #        │ │ │ │ ┌───────── day of the week (0 - 6 or SUN-SAT)
-    - cron: '0 7 * * *'  # Every day at 07:00 UTC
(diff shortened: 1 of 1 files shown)


ayshih:
Can we split it up?  Have the publish builds run every few days, but still have the non-publish builds run daily?

Cadair:
That would be not-trivial given the current setup. I did start poking at refactoring the cron more thoroughly in #7863 which I think would be a pre-requite for that.

nabobalis:
Might as well make it once a week?

Cadair:
Why? This would have it run at the start of the work week, on a Wed before the call and Friday before taking two days off.

nabobalis:
Because no one responds to the runs anyway.

Cadair:
That is another aspect of this, lowering the notifications would maybe make it more likely for people to pay attention. Probably not tho 🙈

ayshih:
Ha, wishful thinking. =P

My general unease is that less frequent runs means that it can be less immediately apparent what the cause is for a newly emerged failure.  This is probably most relevant for the devdeps build against astropy, since there can be a flurry of astropy PRs to sift through.

nabobalis:
We run just the cron cron jobs every day?

nabobalis (approved):

PR #7981, opened 2025-01-07
Add flare event times to GOES NRT example

ehsteve:
## PR Description

This expands on the GOES Near Real-time data example by showing how to use the GOES flare event information which is also provided in near real-time.


ehsteve:
This generates the following plot 
Figure_1

nabobalis (approved):

ehsteve:
Are you going to merge with the test failures? This is all happening so fast!

nabobalis:
> Are you going to merge with the test failures? This is all happening so fast!

Yes.

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -64,9 +67,30 @@
 goes_ts = ts.TimeSeries(goes_data, meta, units, source="xrs")
 
 ###############################################################################
-# Finally, we can plot the timeseries.
+# Now we will download the latest flare event information
+
+flare_events = pd.read_json('https://services.swpc.noaa.gov/json/goes/primary/xray-flares-7-day.json')
```
nabobalis, line 72: So NOAA just keep a list of flares with some basic info for 7 days?

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -64,9 +67,30 @@
...
+
+###############################################################################
+# Next we load it into an astropy Table
```
nabobalis, line 75:
```suggestion
# Next we turn dataframe into an `~astropy.table.Table` in order to plot this data.
```

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -64,9 +67,30 @@
...
+
+goes_class = [str(this_class) for this_class in flare_events['max_class'].values]
+start_time = parse_time(flare_events['begin_time'].values)
+end_time = parse_time(flare_events['end_time'].values)
+peak_time = parse_time(flare_events['max_time'].values)
+flare_list = Table(data={"class": goes_class, "start_time": start_time, "peak_time": peak_time, "end_time": end_time})
```
nabobalis, line 81: Table has a `.frompandas` or something like that, did you try it?

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -64,9 +67,30 @@
 goes_ts = ts.TimeSeries(goes_data, meta, units, source="xrs")
 
 ###############################################################################
-# Finally, we can plot the timeseries.
+# Now we will download the latest flare event information
```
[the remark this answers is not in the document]
ehsteve: Sure. you forgot the "s" at the end of provides.

examples/time_series/goes_xrs_nrt_data.py
ehsteve: If there maybe a better way? Maybe astropy.table can be converted directly?

nabobalis: Table has a `.frompandas` or something like that, worth a shot.

examples/time_series/goes_xrs_nrt_data.py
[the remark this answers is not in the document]
nabobalis: One extra S coming right up.

examples/time_series/goes_xrs_nrt_data.py
ehsteve: That does work though I still have to do the work to convert the times so it really just saves one line of code.

examples/time_series/goes_xrs_nrt_data.py
ehsteve: yes and there is also a latest flare file with just the last one.

examples/time_series/goes_xrs_nrt_data.py
nabobalis: That is unfortunate.

examples/time_series/goes_xrs_nrt_data.py
ehsteve: I guess technically i could just use the pandas dataframe and not use the astropy table at all. I like the idea that event information should be in a Table though. What do you think?

ehsteve: Done.

ehsteve:
This example would be better if astropy had the equivalent of read_json. Then there wouldn't need to be any pandas here.

nabobalis:
They have a json IO plugin that should work https://docs.astropy.org/en/latest/io/unified.html#built-in-table-readers-writers

ehsteve:
That looks like it is just a wrapper around pandas read_json.

nabobalis:
pre-commit.ci autofix

nabobalis (dismissed):

nabobalis:
Sure, I'm about to have a long lunch anyway

ehsteve:
Is timeseries still based on pandas dataframes? I thought the roadmap was to move to astropy TimeSeries objects? I must say that it is now confusing to me that there are two objects in this ecosystem that have the same name and one is only a wrapper around someone else's object.

nabobalis:
> Is timeseries still based on pandas dataframes? I thought the roadmap was to move to astropy TimeSeries objects? I must say that it is now confusing to me that there are two objects in this ecosystem that have the same name and one is only a wrapper around someone else's object.

Don't get this argument started.

ehsteve:
Anyway! have a look, it's cleaner except for the annoying part of having to convert to a pandas dataframe.

ayshih (changes requested):
I have thoughts =)
(will write proper suggestions tonight)

nabobalis:
> I have thoughts =) (will write proper suggestions tonight)

ALBERT NO.

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -1,34 +1,38 @@
...
+# There is also a file with 7 days worth of data.
 
-goes_json_data = pd.read_json("https://services.swpc.noaa.gov/json/goes/primary/xrays-7-day.json")
+goes_data = Table.read('https://services.swpc.noaa.gov/json/goes/primary/xrays-3-day.json', format='pandas.json')
 
 ###############################################################################
 # XRS collects data in two energy channels, "0.05-0.4nm" and "0.1-0.8nm".
 # We separate these "short" and "long" wavelength readings into two arrays.
 
 # This will get us the short wavelength data.
```
ayshih, line 32:
```suggestion
# The recorded flux values alternate between the two XRS energy channels:
# "0.05-0.4nm" and "0.1-0.8nm". We make a `pivot table<https://en.wikipedia.org/wiki/Pivot_table>`__
# that naturally rearranges the data into two flux columns. We then rename the
# columns.

goes_data = goes_data.pivot(
    index='time_tag',
    columns='energy',
    values='observed_flux'
)
goes_data.rename(columns={'0.05-0.4nm': 'xrsa', '0.1-0.8nm': 'xrsb'}, inplace=True)
```
I recognize that pivot tables may not be intuitive to some, but I think makes more sense than making two separate arrays and reforming the `DataFrame`.

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -1,34 +1,38 @@
...
-goes_short = goes_json_data[goes_json_data["energy"] == "0.05-0.4nm"]
+goes_short = goes_data[goes_data["energy"] == "0.05-0.4nm"]
 # This will get us the long wavelength data.
```
ayshih, line 34:
```suggestion
```
Go pivot table!

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -1,34 +1,38 @@
...
-goes_long = goes_json_data[goes_json_data["energy"] == "0.1-0.8nm"]
+goes_long = goes_data[goes_data["energy"] == "0.1-0.8nm"]
```
ayshih, line 35:
```suggestion
```
Go pivot table!

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -42,20 +46,25 @@
...
 # However, here we need to define our own metadata and we will keep it fairly simple.
 
-meta = dict({"instrument": "GOES X-ray sensor", "measurements": "primary", "type": "quicklook"})
+meta = dict(
+    {"instrument": "GOES X-ray sensor", "measurements": "primary", "type": "quicklook"}
+)
 
 ###############################################################################
 #  The final pre-step is create a new `pandas.DataFrame` which we can pass to
 # `sunpy.timeseries.TimeSeries` as the data input.
```
ayshih, line 63:
```suggestion
```
Go pivot table!

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -42,20 +46,25 @@
...
 
-goes_data = pd.DataFrame({"xrsa": goes_short["flux"].values, "xrsb": goes_long["flux"].values}, index=time_array.datetime)
+goes_data = pd.DataFrame(
+    {"xrsa": goes_short['flux'], "xrsb": goes_long['flux']},
+    index=time_array.datetime,
+)
```
ayshih, line 68:
```suggestion
```
Go pivot table!

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -1,34 +1,38 @@
...
 
 # This will get us the short wavelength data.
-goes_short = goes_json_data[goes_json_data["energy"] == "0.05-0.4nm"]
+goes_short = goes_data[goes_data["energy"] == "0.05-0.4nm"]
 # This will get us the long wavelength data.
-goes_long = goes_json_data[goes_json_data["energy"] == "0.1-0.8nm"]
+goes_long = goes_data[goes_data["energy"] == "0.1-0.8nm"]
 
 ###############################################################################
 # `sunpy.timeseries.TimeSeries` requires a datetime index which we can get
```
ayshih, line 38:
```suggestion
# `sunpy.timeseries.TimeSeries` requires a datetime index, which we can get by
# parsing the time strings.

goes_data.index = Time(list(goes_data.index)).datetime
```
(and delete the next three lines)

Unfortunately, because of a bug that needs to be fixed in `parse_time()`, we can't use it here (yet).

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -1,34 +1,38 @@
...
+This example will demonstrate how to make use of GOES XRS Near Real Time data.
+This includes the goes XRS timeseries data as well as the flare times.
+The real-time datasets can all be found at https://services.swpc.noaa.gov/json/goes/primary/
 """
 
 import matplotlib.pyplot as plt
 import pandas as pd
 
 from astropy import units as u
+from astropy.time import TimeDelta
```
ayshih, line 15:
```suggestion
from astropy.time import Time
```
* I need to instantiate `Time` directly because `parse_time()` has a bug
* The use of `TimeDelta` is unnecessary

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -64,9 +73,30 @@
 goes_ts = ts.TimeSeries(goes_data, meta, units, source="xrs")
 
 ###############################################################################
-# Finally, we can plot the timeseries.
+# NOAA also provides the past 7 days of flare information which we can also parse.
+
+flare_events = Table.read("https://services.swpc.noaa.gov/json/goes/primary/xray-flares-7-day.json", format='pandas.json')
```
ayshih, line 78:
```suggestion
flare_events = pd.read_json("https://services.swpc.noaa.gov/json/goes/primary/xray-flares-7-day.json")
```
It's a toss-up between `DataFrame` and `Table` here, but the tiebreaker is that it makes more sense to use `DataFrame` for the other data.

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -64,9 +73,30 @@
...
+
+###############################################################################
+# Finally, we can plot the timeseries and overlay all the flares that occurred.
+# We will do this only for the last day otherwise the plot may be too busy.
 
 fig, ax = plt.subplots()
 goes_ts.plot(axes=ax)
```
ayshih, line 85:
```suggestion
goes_ts.plot(axes=ax)
plot_start = goes_ts.time[-1] - 1 * u.day
ax.set_xlim(
    plot_start.datetime, (plot_start + 1 * u.day).datetime
)
```
More readable to define and use a variable `plot_start`

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -64,9 +73,30 @@
...
+        ax.axvline(peak_time.datetime)
+        ax.axvspan(
+            start_time.datetime,
+            end_time.datetime,
+            alpha=0.2,
+            label=f'{peak_time} {this_flare["max_class"]}',
+        )
+ax.set_xlim(
+    goes_ts.time[-1].datetime, (goes_ts.time[-1] - TimeDelta(1 * u.day)).datetime
+)
```
ayshih, line 100:
```suggestion
```
I moved this before the plot

examples/time_series/goes_xrs_nrt_data.py
ayshih: I do have a question for you: why did you make this plot go backwards in time?  Is that a common way to visualize this data?

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -64,9 +73,30 @@
...
+ax.legend(loc=2)
```
ayshih, line 101:
```suggestion
ax.legend(loc="upper left")
```
Oof, I would never suggest to anyone to use the arcane location codes

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -64,9 +73,30 @@
...
+# Finally, we can plot the timeseries and overlay all the flares that occurred.
+# We will do this only for the last day otherwise the plot may be too busy.
 
 fig, ax = plt.subplots()
 goes_ts.plot(axes=ax)
-
+for this_flare in flare_events:
+    start_time, peak_time, end_time = parse_time(
+        [this_flare["begin_time"], this_flare["max_time"], this_flare["end_time"]]
+    )
```
ayshih, line 89:
```suggestion
for start_time, peak_time, end_time, max_class in zip(parse_time(flare_events["begin_time"]),
                                                      parse_time(flare_events["max_time"]),
                                                      parse_time(flare_events["end_time"]),
                                                      flare_events["max_class"]):
```
This reduces the calls to `parse_time()` and is more readable to me

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -64,9 +73,30 @@
...
+    if start_time > (goes_ts.time[-1] - TimeDelta(1 * u.day)):
+        ax.axvline(peak_time.datetime)
+        ax.axvspan(
+            start_time.datetime,
+            end_time.datetime,
+            alpha=0.2,
+            label=f'{peak_time} {this_flare["max_class"]}',
```
ayshih, line 96:
```suggestion
            label=f'{peak_time} {max_class}',
```
Because I changed the unpacking approach

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -64,9 +73,30 @@
...
+# We will do this only for the last day otherwise the plot may be too busy.
 
 fig, ax = plt.subplots()
 goes_ts.plot(axes=ax)
-
+for this_flare in flare_events:
+    start_time, peak_time, end_time = parse_time(
+        [this_flare["begin_time"], this_flare["max_time"], this_flare["end_time"]]
+    )
+    if start_time > (goes_ts.time[-1] - TimeDelta(1 * u.day)):
```
ayshih, line 90:
```suggestion
    if peak_time > plot_start:
```
Because you still want to plot the flare peak even if the start time happens to be before the plot start

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -64,9 +73,30 @@
...
+        ax.axvline(peak_time.datetime)
+        ax.axvspan(
```
ayshih, line 92:
```suggestion
    if end_time > plot_start:
        ax.axvspan(
```
Because you still want to plot the shaded region as long as it ends within the plot

ayshih (changes requested):
* I think you should use `DataFrame` instead of `Table` throughout
* This definitely warrants a changelog entry
* I do not think this should be backported, but I don't feel strongly about it

examples/time_series/goes_xrs_nrt_data.py
ayshih: `parse_time()` bug to be fixed by #7983

nabobalis:
All of Alberts suggestions are in, I added a changelog. I still want this backported. 

I think there is one open question left to answer.

nabobalis:
Docs are failing due to 
```
<unknown>:1423: WARNING: py:obj reference target not found: aiapy.calibrate.register [ref.obj]
<unknown>:1774: WARNING: py:obj reference target not found: aiapy.calibrate.register [ref.obj]
```
which is because I deleted the aiapy documentation.

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -1,61 +1,65 @@
...
+goes_data.rename(columns={'0.05-0.4nm': 'xrsa', '0.1-0.8nm': 'xrsb'}, inplace=True)
 
 ###############################################################################
-# `sunpy.timeseries.TimeSeries` requires a datetime index which we can get
-# directly and transform into `astropy.time.Time`.
+# `sunpy.timeseries.TimeSeries` requires a datetime index, which we can get by
+# parsing the time strings.
 
-time_array = parse_time(goes_short["time_tag"])
+goes_data.index = Time(list(goes_data.index)).datetime
```
ayshih, line 45:
```suggestion
goes_data.index = parse_time(list(goes_data.index)).datetime
```
After #7983 is merged

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -1,61 +1,65 @@
...
+This example will demonstrate how to make use of GOES XRS Near Real Time data.
+This includes the goes XRS timeseries data as well as the flare times.
+The real-time datasets can all be found at https://services.swpc.noaa.gov/json/goes/primary/
 """
 
 import matplotlib.pyplot as plt
 import pandas as pd
 
 from astropy import units as u
+from astropy.time import Time
```
ayshih, line 15:
```suggestion
```
After #7983 is merged

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -1,61 +1,65 @@
...
 ###############################################################################
-# XRS collects data in two energy channels, "0.05-0.4nm" and "0.1-0.8nm".
-# We separate these "short" and "long" wavelength readings into two arrays.
-
-# This will get us the short wavelength data.
-goes_short = goes_json_data[goes_json_data["energy"] == "0.05-0.4nm"]
-# This will get us the long wavelength data.
-goes_long = goes_json_data[goes_json_data["energy"] == "0.1-0.8nm"]
+# The recorded flux values alternate between the two XRS energy channels:
+# "0.05-0.4nm" and "0.1-0.8nm". We make a `pivot table<https://en.wikipedia.org/wiki/Pivot_table>`__
```
ayshih, line 30:
```suggestion
# "0.05-0.4nm" and "0.1-0.8nm". We make a `pivot table <https://en.wikipedia.org/wiki/Pivot_table>`__
```
I think this will fix the link

ayshih (commented):
Hmm, this example is potentially problematic.  As of the instant of writing, there's only one flare marked on the plot (because apparently there is a lag in identifying flares):
sphx_glr_goes_xrs_nrt_data_001
Thus, this might appear to the user that the code is not working properly.  Furthermore, as solar activity declines, there will increasingly be stretches of time when the flare code doesn't actually have any flares to plot...

nabobalis:
> Hmm, this example is potentially problematic. As of the instant of writing, there's only one flare marked on the plot (because apparently there is a lag in identifying flares):  Thus, this might appear to the user that the code is not working properly. Furthermore, as solar activity declines, there will increasingly be stretches of time when the flare code doesn't actually have any flares to plot...

Is it worth just adding a note stating that? If no flares happen, no flares can be plotted.

hayesla:
> > Hmm, this example is potentially problematic. As of the instant of writing, there's only one flare marked on the plot (because apparently there is a lag in identifying flares):  Thus, this might appear to the user that the code is not working properly. Furthermore, as solar activity declines, there will increasingly be stretches of time when the flare code doesn't actually have any flares to plot...
> 
> Is it worth just adding a note stating that? If no flares happen, no flares can be plotted.

I think we'd need a note for both - the case of the lag is more of an issue I would imagine, especially like in the case Albert showed there are clearly flares but not yet in the flarelist

examples/time_series/goes_xrs_nrt_data.py
ehsteve: I removed this.

ehsteve:
Good point on catching when no flares are identified. To fix this I

- went back to using all 7 days of data to maximize num flares
- now only plot the 5 largest flares and only peak times (otherwise the plot is too busy for 7 days during active times). The flare time spans are no longer visible on 7 day plot.
- plot title now lists the number of flares found

ehsteve:
@ayshih your suggestion `goes_data.index = parse_time(list(goes_data.index)).datetime` gave me an error because parse_time expects more zeros with a time string with Z at the end.

ehsteve:
Here is what the new plot looks like
Figure_1

ehsteve:
I hope you all take it easier on other contributors! This should be an `easy` contribution of only a few lines of extra code in an example and we are already up to 63 interactions here!

ehsteve:
Not that I'm complaining. I can take it. Bring it on!

ayshih:
That's because #7983 needs to be merged first

nabobalis:
it is now merged.

nabobalis:
> I hope you all take it easier on other contributors! This should be an `easy` contribution of only a few lines of extra code in an example and we are already up to 63 interactions here!

Its an example, they always generate the most comments.

ehsteve:
@nabobalis reverted to @ayshih version with parse_time fix. Anything else?

nabobalis:
Whatever else that Albert wants.

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -64,9 +69,30 @@
...
+###############################################################################
+# Finally, we can plot the timeseries and overlay all the flares that occurred.
+# We will do this only for the last day otherwise the plot may be too busy.
 
 fig, ax = plt.subplots()
 goes_ts.plot(axes=ax)
-
+for index, this_flare in largest_flares.iterrows():
+    peak_time = parse_time(this_flare["max_time"])
+    ax.axvline(peak_time.datetime, label=f'{peak_time.datetime} {this_flare["max_class"]}')
```
ayshih, line 95:
```suggestion
if len(largest_flares) > 0:
    for _, this_flare in largest_flares.iterrows():
        peak_time = parse_time(this_flare["max_time"])
        ax.axvline(peak_time.datetime, label=f'{peak_time.datetime} {this_flare["max_class"]}')
```
We can't call `.iterrows()` on an empty list.  There's probably a nicer way to write this, but this fix is acceptable enough to merge.

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -64,9 +69,31 @@
...
+    "https://services.swpc.noaa.gov/json/goes/primary/xray-flares-7-day.json"
+)
+
+###############################################################################
+# Next we sort the list and retain the five largest flares. 
+
+if len(flare_events) > 0:
+    largest_flares = flare_events.sort_values('max_xrlong', ascending=False)[:5]
+else:
+    largest_flares = []
```
nabobalis, line 86: Would checking `.empty` work instead of the len?

examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -64,9 +69,31 @@
...
+
+###############################################################################
+# Finally, we plot the timeseries and overlay the largest flares that occurred.
 
 fig, ax = plt.subplots()
 goes_ts.plot(axes=ax)
-
+if len(largest_flares) > 0:
```
nabobalis, line 93:
```suggestion
if largest_flares:
```
would this work?

examples/time_series/goes_xrs_nrt_data.py
ayshih: Eh, not clearly better as far as I'm concerned

examples/time_series/goes_xrs_nrt_data.py
ayshih: It would work, but I prefer the explicit call in a gallery example because I think falsy values are confusing to the average user

nabobalis:
pre-commit.ci autofix

nabobalis:
Thanks everyone

PR #7982, opened 2025-01-07
Warn when coordinate obstime != to observer time

### LINKED ISSUE #7193: Emit an info message when instantiating an observer-based coordinate where the `obstime` values are not the same

For a coordinate with an observer, there is no requirement that the `obstime` for the observer location be the same as the `obstime` for coordinate.  However, due to the quirkiness that can result from a mismatch in `obstime` – particularly if the `transform_with_sun_center()` context manager is not being used – it'd be prudent for us to emit an informational message that draws attention to the mismatch.  We could report the time difference in the message, and even report how much the Sun has moved during that time difference.

ankitkhushwaha:
## PR Description

fixes #7193

- [x] this pr raises a warning when coordinate obstime is not equal to observer obstime
- [x] Add tests


ankitkhushwaha:
2 tests from reproject.py are failing, will fix it.

ayshih (commented):
Hmm...  My notion had been to the use of `log.info()` instead of `warn_user()`, but it's useful that you chose the latter because that revealed that the message gets triggered quite a few times in our test suite.  Apologies, but I suggest that you hold off on any further work on this PR because the original feature request (#7193) needs more thought.

ankitkhushwaha:
closing it
maybe #7193  will took some more time to resolved

PR #7983, opened 2025-01-08
Fixed a number of bugs associated with parsing time

ayshih:
This PR fixes a number of bugs that affect the parsing of time strings, primarily when they are in a list.

First, the parsing of a list of time strings (as opposed to a single time string) does not automatically set the TAI time scale as it should when "TAI" appears in the time strings:
```python
>>> parse_time("2016.05.04_21:08:12_TAI")
<Time object: scale='tai' format='isot' value=2016-05-04T21:08:12.000>
>>> parse_time(["2016.05.04_21:08:12_TAI"])
<Time object: scale='utc' format='isot' value=['2016-05-04T21:08:12.000']>
```

Second, the regex matching to find candidate time formats is sloppy, so there can be spurious matches that will not parse successfully:

- Periods are not escaped in the regex string, so they become wildcards
- Trailing characters are allowed

These spurious matches do not cause issues for parsing a single time string – aside from a small performance impact – because the code logic will keep trying candidate time formats until one works.  However...

Third, the code logic for parsing a list of time strings is different from parsing a single time string: it tries only the first candidate time format, so a spurious match that comes before the true match results in errors:
```python
>>> parse_time("2007-05-04T21:08:12.999Z")
<Time object: scale='utc' format='isot' value=2007-05-04T21:08:12.999>
>>> parse_time(["2007-05-04T21:08:12.999Z"])
...
ValueError: unconverted data remains: Z
```
and
```python
>>> "parse_time("20070504T210800")
<Time object: scale='utc' format='isot' value=2007-05-04T21:08:00.000>
>>> parse_time(["20070504T210800"])
...
ValueError: time data '20070504T210800' does not match format '%Y%m%dT%H%M%S.%f'
```

A corollary to the above is that the code also doesn't fall back to astropy's parsing if there is no matching format in sunpy:
```python
>>> parse_time('J2000')
<Time object: scale='tt' format='jyear_str' value=J2000.000>
>>> parse_time(['J2000'])
...
TypeError: strptime() argument 1 must be str, not <class 'NoneType'>
```

### DIFF
--- a/sunpy/time/time.py
+++ b/sunpy/time/time.py
@@ -218,54 +218,38 @@ def convert_time_astropy(time_string, **kwargs):
     return time_string
 
 
-@convert_time.register(list)
-def convert_time_list(time_list, format=None, **kwargs):
-    item = time_list[0]
-    # If we have a list of strings, need to get the correct format from our
-    # list of custom formats.
-    if isinstance(item, str) and format is None:
-        if 'TAI' in item:
-            kwargs['scale'] = 'tai'
-        string_format = _get_time_fmt(item)
-        return Time.strptime(time_list, string_format, **kwargs)
-
-    # Otherwise return the default method
-    return convert_time.dispatch(object)(time_list, format, **kwargs)
-
-
 @convert_time.register(str)
+@convert_time.register(list)
 def convert_time_str(time_string, **kwargs):
-    if 'TAI' in time_string:
-        kwargs['scale'] = 'tai'
+    is_single_string = isinstance(time_string, str)
+    is_string_list = isinstance(time_string, list) and isinstance(time_string[0], str)
 
-    for time_format in TIME_FORMAT_LIST:
-        try:
+    if is_single_string or is_string_list:
+        first_item = time_string[0] if is_string_list else time_string
+        if 'TAI' in first_item:
+            kwargs['scale'] = 'tai'
+
+        for time_format in TIME_FORMAT_LIST:
             try:
-                ts, add_one_day = _regex_parse_time(time_string, time_format)
-            except TypeError:
-                break
-            if ts is None:
-                continue
-            t = Time.strptime(ts, time_format, **kwargs)
-            if add_one_day:
-                t += _ONE_DAY_TIMEDELTA
-            return t
-        except ValueError:
-            pass
+                try:
+                    ts, add_one_day = _regex_parse_time(first_item, time_format)
+                except TypeError:
+                    break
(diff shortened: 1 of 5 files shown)


nabobalis (approved):

sunpy/time/time.py
```diff
@@ -217,54 +218,40 @@ def convert_time_astropy(time_string, **kwargs):
...
-        kwargs['scale'] = 'tai'
+    is_single_string = isinstance(time_string, str)
+    is_string_list = isinstance(time_string, list) and isinstance(time_string[0], str)
 
-    for time_format in TIME_FORMAT_LIST:
-        try:
+    if is_single_string or is_string_list:
+        first_item = time_string[0] if is_string_list else time_string
+        if 'TAI' in first_item:
+            kwargs['scale'] = 'tai'
```
Cadair, line 230: Do we support a mix of TAI and non-TAI in a single list?

sunpy/time/time.py
```diff
@@ -217,54 +218,40 @@ def convert_time_astropy(time_string, **kwargs):
...
+                except TypeError:
+                    break
+                if ts is None:
+                    continue
+                if is_single_string:
+                    t = Time.strptime(ts, time_format, **kwargs)
+                    if add_one_day:
+                        t += _ONE_DAY_TIMEDELTA
+                else:
+                    # For a list of strings, we do not try to correct 24:00:00
```
Cadair, line 245: Why?

Cadair (approved):
lgtm

sunpy/time/time.py
ayshih: Nope, because that would be a list with mixed string formats.  Because we pass the entire list to a single `strptime()` call, the list has to have a single string format.  If you have a mixed-format list, you have to call `parse_time()` on the individual elements and then reassemble into a single array `Time`.

sunpy/time/time.py
ayshih: When parsing a single string, we check for "24:00:00" at the same time we check if a string format matches because we need to modify the string before passing it to `strptime()`.  However, for a list of strings, which are typically all of the same string format, we currently take the performance shortcut of looking at only the first element to determine which string formats to try.  That means we never actually look at other elements.

Expanding support to a list is certainly doable, but the performance impact of looking at every element in a long list for "24:00:00" should not be accepted lightly, given how uncommon it should be.  Maybe a keyword option?  (The implementation would be quite similar to also supporting mixed-format lists.)

PR #7986, opened 2025-01-08
Fixed the formatting of GOES XRS channel labels

ayshih:
This PR fixes the formatting of GOES XRS channel labels in TimeSeries plots:

- Fixes the angstrom symbol so that it is not italicized
- Replaces the double hyphen with a longer dash (I chose to use the less correct Unicode minus instead of the correct en dash because I feel, that in Matplotlib's rendering, the en dash is not sufficiently visually distinct from a hyphen)

## Before this PR
baseline (2)

## After this PR
result (4)


nabobalis:
Backport not?

ayshih:
Judgement call, but I landed on "no backport" because a user could conceivably be checking against the strings, and it's not worth it to break any code for this fix.

Cadair (approved):

hayesla (approved):

PR #7988, opened 2025-01-09
update annotate issue#7166 

sriram-dev-9:
Type checking annotate

# Changelog

### Changed
- **`annotate` Argument Validation**: The validation for the `annotate` parameter in the `plot` method has been updated to raise a `TypeError` with a clearer error message if the value is not a boolean. The new message suggests that users explicitly pass `axes=...` if they meant to provide the axes argument instead of `annotate`.

I am a new contributor, and this is my first contribution please excuse me for any mistakes.

### DIFF
--- a/changelog/7988.bugfix.rst
+++ b/changelog/7988.bugfix.rst
@@ -1,0 +1,5 @@
+# Changelog
+
+## [Unreleased] - 2025-01-09
+### Changed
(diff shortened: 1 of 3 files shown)


nabobalis:
Hello @sriram-dev-9, thank you for the pull request.

Can you explain what issue this is fixing?

ayshih:
There is existing code right below the code you added that already checks for `annotate` not being a boolean value (PR was #7163).  If the intent of your PR is to edit the error message, you should modify the existing code rather than adding a new check that supersedes the old check.

nabobalis:
In that case, I will close this PR. @sriram-dev-9, if you want to make a change to the message, please open a new PR.

Issue #7991, opened 2025-01-09
 use declination_axis.set_major_formatter(new_format_method) 


tanishy7777:
### Describe the feature

 Astropy >= 7.0 uses declination_axis.set_major_formatter(new_format_method)

### Proposed solution

replace 

```python
declination_axis._formatter_locator.formatter = new_format_method
```
with 
```python
declination_axis.set_major_formatter(new_format_method)
```

in `sunpy.visualization.visualization.py`

advait-zx:
sunpy.visualization.visualization.py
cant find file location

nabobalis:
I'll have to correct that but until sunpy depends on 7.0, this can't be fixed.

PR #7992, opened 2025-01-09
Update visualization.py

tanishy7777:
Astropy >= 7.0 uses declination_axis.set_major_formatter(new_format_method)
so replaced 

Fixes: #7991 

```python
declination_axis._formatter_locator.formatter = new_format_method
```
with 
```python
declination_axis.set_major_formatter(new_format_method)
```


ayshih:
We cannot make this change until astropy 7.0 becomes our minimum dependency (in about a year).

PR #7834, opened 2024-10-16
Fix CI

nabobalis:

PR #8009, opened 2025-01-10
Adds FakeClient to reproduce work and error property in UnifiedResponse

tanishy7777:
Fixes: #7625 

Here is my plan of action:
Step1: I have reproduced the error that was mentioned in the issue by creating an ExampleClient (code taken from [here](https://docs.sunpy.org/en/stable/topic_guide/extending_fido.html) for the client)
image

Step 2: Add an error property in the `UnifiedResponse` class
Step 3:

### DIFF
--- a/fake_client.py
+++ b/fake_client.py
@@ -1,0 +1,111 @@
+from functools import partial
+from sunpy.net.base_client import BaseClient, convert_row_to_table
+import sunpy.util.net
+
+import sunpy.net.attrs as a
+import sunpy.net
+from sunpy.net.attr import AttrWalker, AttrAnd, AttrOr, DataAttr
+from sunpy.net.base_client import QueryResponseTable
+
+walker = AttrWalker()
+
+@walker.add_creator(AttrOr)
+def create_or(wlk, tree):
+    results = []
+    for sub in tree.attrs:
+        results.append(wlk.create(sub))
+
+    return results
+
+
+@walker.add_creator(AttrAnd, DataAttr)
+def create_and(wlk, tree):
+    result = dict()
+    wlk.apply(tree, result)
+    return [result]
+
+
+@walker.add_applier(a.Time)
+def _(wlk, attr, params):
+    return params.update({'startTime': attr.start.isot,
+                            'endTime': attr.end.isot})
+
+
+@walker.add_applier(a.Level)
+def _(wlk, attr, params):
+    return params.update({'level': attr.value})
+
+
+class ExampleClient(BaseClient):
+    size_column = 'Filesize'
+
+    def search(self, query):
+        queries = walker.create(query)
+        print("Seaching in fake client")
+        results = []
+        for query_parameters in queries:
+            results.append(self._make_search(query_parameters))
+            # try:
+            #     resp = self._make_search(query_parameters)
+            #     results.append(resp)
+            # except ConnectionRefusedError as e:
+            #     # e.client_name = client.__class__.__name__.lower()
+            #     results.append(e)
+
+        return QueryResponseTable(results, client=self)
+    
+    @classmethod
+    def register_values(cls):
+
+        from sunpy.net import attrs
+        adict = {
+        attrs.Instrument: [("LASCO", "Large Angle and Spectrometric Coronagraph")],
+        attrs.Source: [('SOHO', 'Solar and Heliospheric Observatory')],
+        attrs.Provider: [('SDAC', 'Solar Data Analysis Center')],
+        attrs.Detector: [('C1', 'Coronograph 1'),
+                        ('C2', 'Coronograph 2'),
+                        ('C3', 'Coronograph 3')]
+        }
+
+        return adict
+    
+    def _make_search(self, query):
+        raise ConnectionRefusedError                                        
+
+    def _make_filename(path, row, resp, url):
+        # Define a fallback filename based on the information in the search results
+        name = f"row['ID'].fits"
+
+        if resp:
+            cdheader = resp.headers.get("Content-Disposition", None)
+            if cdheader:
+                _, params = sunpy.util.net.parse_header(cdheader)
+                name = params.get('filename', "")
+
+        return path.format(file=name, **row.response_block_map)
+
+    @convert_row_to_table
+    def fetch(self, query_results, *, path, downloader, **kwargs):
+        for row in query_results:
+            filepath = partial(self._make_filename, path, row)
+
+            url = f"https://sfsi.sunpy.org/download/{row['ID']}"
+            downloader.enqueue_file(url, filename=filepath)
+
+    @classmethod
+    def _can_handle_query(cls, *query):
+        query_attrs = set(type(x) for x in query)
+        supported_attrs = {a.Time, a.Level}
+        return supported_attrs.issuperset(query_attrs)
+    
+client = ExampleClient()
+# client.search(a.Time('2014-01-01T00:00:00', '2014-01-01T00:10:00')) 
+print(client.search(a.Time('2014-01-01T00:00:00', '2014-01-01T00:10:00')))
+
+
+
+
+
+
+
+
--- a/sunpy/net/fido_factory.py
+++ b/sunpy/net/fido_factory.py
@@ -54,6 +54,11 @@ class UnifiedResponse(Sequence):
         self._list = []
         self._numfile = 0
         for result in results:
+            if isinstance(result, ConnectionRefusedError):
+                self._list.append(None)
+                client_key = getattr(result, "client_name", len(self._list)-1)
+                self.errors[client_key] = result
+
             if isinstance(result, QueryResponseRow):
                 result = result.as_table()
 
@@ -161,6 +166,24 @@ class UnifiedResponse(Sequence):
         The number of records returned in all responses.
         """
         return self._numfile
+    
+    @property
+    def errors(self):
+        """
+        Returns a list of errors from the query.
+
+        If no errors are present, an empty list is returned.
+        """
+        # assuming that self._list has all the responses including errors
+        err = []
+        for res in self._list:
+            if isinstance(res, QueryResponseTable):
+                err.append(None)
+            else:
+                # connection refused error etc
+                # Not sure about res.errors, need to check
+                err.append(res.errors)
+        return err
 
     def _repr_html_(self):
         nprov = len(self)
@@ -478,8 +501,13 @@ class UnifiedDownloaderFactory(BasicRegistrationFactory):
         candidate_widget_types = self._check_registered_widgets(*query)
         results = []
         for client in candidate_widget_types:
-            tmpclient = client()
-            results.append(tmpclient.search(*query))
+            try:
+                tmpclient = client()
+                results.append(tmpclient.search(*query))
+            except Exception as e:
+                print(f"Error: {e} here")
+                e.client_name = client.__class__.__name__.lower()
+                results.append(e)
 
         # This method is called by `search` and the results are fed into a
         # UnifiedResponse object.
--- a/fido_test.py
+++ b/fido_test.py
@@ -1,0 +1,11 @@
+from sunpy.net import Fido, attrs as a
+import astropy.units as u
+
+
+result = Fido.search(a.Time('2012/3/4', '2012/3/6'), a.Source.soho) 
+print(result)
+
+# Reproduces the bug where Fido.search crashes entirely even 
+# if one of the client has an error using a FakeClient
+
+
--- a/sunpy/net/__init__.py
+++ b/sunpy/net/__init__.py
@@ -13,5 +13,6 @@ from sunpy.net import jsoc as _
 from sunpy.net import vso as _
 from sunpy.net.fido_factory import Fido
 from sunpy.net.scraper import Scraper
+from fake_client import ExampleClient as _    
 
 __all__ = ["Fido", "Scraper"]


tanishy7777:
The way `search` is implemented in a lot of clients is without the try except block.

VSOClient uses try, except and adds exceptions to the response if there are
```python
@deprecated_renamed_argument("response_format", None,"6.0", warning_type=SunpyDeprecationWarning)
def search(self, *query, response_format=None):
if response_format is None:
            response_format = "table"
        query = and_(*query)
        QueryRequest = self.api.get_type('VSO:QueryRequest')
        VSOQueryResponse = self.api.get_type('VSO:QueryResponse')
        responses = []
        exceptions = []
        for block in walker.create(query, self.api):
            try:
                query_response = self.api.service.Query(
                    QueryRequest(block=block)
                )
                for resp in query_response:
                    if resp["error"]:
                        warn_user(resp["error"])
                responses.append(
                    VSOQueryResponse(query_response)
                )
            except Exception as ex:
                exceptions.append(ex)

        responses = self.merge(responses)
        if response_format == "legacy":
            response = QueryResponse.create(responses)
        else:
            response = VSOQueryResponseTable.from_zeep_response(responses, client=self)

        for ex in exceptions:
            response.add_error(ex)

        return response
```

but JSOClient dosent do try except, in case there is a `ConnectionError` here I am not sure it can be handled in `fidofactory.py` consistantly
```python
def search(self, *query, **kwargs):
return_results = JSOCResponse(client=self)
        query = and_(*query)
        blocks = []
        for block in walker.create(query):
            extra_kwargs = kwargs.copy()
            extra_kwargs |= block
            # Update blocks with deep copy of extra_kwargs because in _make_recordset we use .pop() on
            # element from extra_kwargs
            blocks.append(copy.deepcopy(extra_kwargs))
            return_results = astropy.table.vstack([return_results, self._lookup_records(extra_kwargs)])
        return_results.query_args = blocks
        return_results._original_num_rows = len(return_results)
        return return_results
```

Even the `SRSClient` which inherits from `GenericClient` dosent use the try except block. I think we should first make sure all clients return Empty Tables when error is generated as stated  here https://github.com/sunpy/sunpy/issues/7625#issuecomment-2114602574

tanishy7777:
@Cadair @nabobalis

tanishy7777:
I think this adds the error dict as a property to `UnifiedResonse` as shown below.
image

Please let me know if any more changes are needed for this part.
For testing, I think the `fake_client.py` file can be converted to be used with pytest .

@nabobalis @Cadair @samaloney

fake_client.py
samaloney, line 1: Have a look at this PR for how you can raise the error without making fake classes using mocking #7626

sunpy/net/fido_factory.py
```diff
@@ -53,16 +53,27 @@ def __init__(self, *results):
...
         self._numfile = 0
+        self._errors = {}
         for result in results:
-            if isinstance(result, QueryResponseRow):
-                result = result.as_table()
+            if isinstance(result, Exception):
+                arr = np.array([])
+                t = Table(arr, names=())
+                self._errors[result.client.__class__.__name__] = result
+                result = QueryResponseTable(t, client=result.client)
```
samaloney, line 62: I think this can just be `QueryResponseTable([], client=client=result.client)`

sunpy/net/fido_factory.py
```diff
@@ -162,6 +173,16 @@ def file_num(self):
         """
         return self._numfile
 
+    @property
```
samaloney, line 176: So a UnifiedResponse can be sliced by name or number so I think the errors would need to work in a similar way and also update the errors accordingly? So given a response from two clients 

```
res.errors # would return all errors/None
res[0].errors # would only return error from first client
res['a'].errors # would only return error from 'a' client
``` 

or 

```
res.errors[0] # would only return error from first client
res.errors['a'] # would only return error from 'a' client
```

I think the first option would only require the `__getitem__` to be updated and is maybe better anyway @Cadair what do you think?

samaloney (changes requested):
Thanks for the PR! This is a good start as @nabobalis indicated this is a bit complicated and may take a bit of back and forth to get right.

fake_client.py
tanishy7777: got it

sunpy/net/fido_factory.py
tanishy7777: > I think this can just be `QueryResponseTable([], client=result.client)`

got it. I have made the necessary changes

sunpy/net/fido_factory.py
```diff
@@ -53,16 +53,25 @@ def __init__(self, *results):
         """
         self._list = []
         self._numfile = 0
+        self._errors = {}
         for result in results:
-            if isinstance(result, QueryResponseRow):
-                result = result.as_table()
+            if isinstance(result, Exception):
```
nabobalis, line 58: So a client is now returning an Exception if it errors?

sunpy/net/fido_factory.py
```diff
@@ -53,16 +53,25 @@ def __init__(self, *results):
...
+                if isinstance(result, QueryResponseRow):
+                    result = result.as_table()
 
-            if not isinstance(result, QueryResponseTable):
-                raise TypeError(
-                    f"{type(result)} is not derived from sunpy.net.base_client.QueryResponseTable")
+                if isinstance(result, QueryResponseColumn):
+                    result = result.as_table()
+
+                self._errors[result.client.__class__.__name__] = None
```
nabobalis, line 69: What is the reason for this?

tanishy7777: > What is the reason for this?


```python
if isinstance(result, QueryResponseColumn):
          result = result.as_table()
```                    
This was there in the original codebase so I didn't want to remove it.
Also if I dont convert `result` to a table, I can't seem to access the client property and the next line runs into an error.
```python
self._errors[result.client.__class__.__name__] = None
```

nabobalis: I mean the error line, why do we need to set it to None?

sunpy/net/fido_factory.py
tanishy7777: Yes, this is the default behaviour when a request fails for a client we get a `ConnectionRefusedError`
so basically I capture that here and modify the `errors` property accordingly. But if it it is not an Exception. the line I removed and  some other lines are run as shown below:
```python
# If exception raised from client
if isinstance(result, Exception):
                self._errors[result.client.__class__.__name__] = result
                result = QueryResponseTable([], client=result.client)
# if no exception raised run that same code as in the original codebase
# but we add the self._errors[.... line
else:

  if isinstance(result, QueryResponseRow):
      result = result.as_table()

  if isinstance(result, QueryResponseColumn):
      result = result.as_table()

  # this line is added
  self._errors[result.client.__class__.__name__] = None


  if not isinstance(result, QueryResponseTable):
      raise TypeError(
          f"{type(result)} is not derived from sunpy.net.base_client.QueryResponseTable")

```

changelog/8009.feature.rst
```diff
@@ -0,0 +1,2 @@
+Enhanced `sunpy.net.Fido.search()`` to be robust against individual client failures. If a single client fails to respond,
+the search will continue with other clients, ensuring that the functionality is not disrupted by errors from a single external client.
```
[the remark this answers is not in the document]
tanishy7777: > This there a test to confirm this behavior?

Yes, so the code is being tested for `ConnectionError` from the client by adding
`side_effect=ConnectionError('VSO is down')`, which basically makes sure that when we call the function
`sunpy.net.vso.vso.VSOClient.search` we raise a `ConnectionError`

```python
@pytest.mark.remote_data
@mock.patch("sunpy.net.vso.vso.VSOClient.search", side_effect=ConnectionError('VSO is down'))
def test_fido_client_error(vso_search):
    with pytest.raises(ConnectionError):
        q = Fido.search(a.Time('2012/01/01', '2012/01/01'), a.Instrument.soon) # noqa: F841
```

this is added in the `test_fido.py` file

tanishy7777:
@Cadair can you please suggest which of the 2 options would be better https://github.com/sunpy/sunpy/pull/8009?notification_referrer_id=NT_kwDOCIsbr7UxNDIyMDk1MDQ2NjoxNDMzMzQzMTk#discussion_r1913263736

sunpy/net/fido_factory.py
tanishy7777: > I mean the error line, why do we need to set it to None?

oh, actually this was suggested in https://github.com/sunpy/sunpy/issues/7625#issuecomment-2114602574 in the original issue. To have `None` in the errors property to denote no errors from the client that was passed.

sunpy/net/tests/test_fido.py
```diff
@@ -76,6 +76,12 @@ def test_online_fido(query):
     unifiedresp = Fido.search(query)
     check_response(query, unifiedresp)
 
+@pytest.mark.remote_data
+@mock.patch("sunpy.net.vso.vso.VSOClient.search", side_effect=ConnectionError('VSO is down'))
+def test_fido_client_error(vso_search):
+    with pytest.raises(ConnectionError):
+        q = Fido.search(a.Time('2012/01/01', '2012/01/01'), a.Instrument.soon) # noqa: F841
```
nabobalis, line 83:
```suggestion
        Fido.search(a.Time('2012/01/01', '2012/01/01'), a.Instrument.soon)
```

nabobalis: I thought the change was that it won't error but return a results with an error property?

sunpy/net/fido_factory.py
tanishy7777: > Why would we set it to None if there is no error? What is the benefit of this?

Maybe having an errors dictionary to check if there are any errors from the client after you have done a search. 
Like just as a sanity check to verify that all clients have responded maybe. I am not sure though

sunpy/net/tests/test_fido.py
tanishy7777: > I thought the change was that it won't error but return a results with an error property?

I am not sure I understand this. Like are you talking with reference to the above change that you made?

nabobalis: No, to the code changes that you have made.

tanishy7777: > No, to the code changes that you have made.
Yes thats what it does right?
`results` will behave the same way but now I can also access the errors dict 
using results.errors


> No, to the code changes that you have made.

Makes sense. I went through the errors in pytest
image
so the `ConnectionError` is raised even before the search method is run for `vso`

in the `fido_factory.py` method
image
the `client()` method is called which tries to get the url online but here it is not able to find it since we are mocking it. so it throws an error.

I moved the `tmpclient = client()` inside the try block. and made a few changes, now it works!

image

tanishy7777: TLDR: result dosent throw error if a client is failing(even in its build method) and this is the output after running `test_fido.py`
image

nabobalis: That's good.

examples/plotting/add_countour_with_Spherical_screen.py
```diff
@@ -0,0 +1,57 @@
+"""
```
nabobalis, line 1: I assume there was some cross talk between your branches?

tanishy7777: Yea I think so. Will fix it

tanishy7777: yep fixed

sunpy/net/tests/test_fido.py
```diff
@@ -76,6 +76,10 @@ def test_online_fido(query):
     unifiedresp = Fido.search(query)
     check_response(query, unifiedresp)
 
+@mock.patch("sunpy.net.vso.vso.VSOClient.search", side_effect=ConnectionError('VSO is down'))
+def test_fido_client_error(vso_search):
+    results = Fido.search(a.Time("2016/10/01", "2016/10/02"), a.Instrument.aia)
+    assert len(results.errors) > 0
```
nabobalis, line 82: Can we not check the exact error?

sunpy/net/fido_factory.py
```diff
@@ -478,11 +501,15 @@ def _make_query_to_client(self, *query):
...
-            tmpclient = client()
-            results.append(tmpclient.search(*query))
+            try:
+                tmpclient  = client()
+                result = tmpclient.search(*query)
+                result.client = client
+                results.append(result)
+            except Exception as e:
+                e.client = client
+                results.append(e)
```
nabobalis, line 511: For this to work, the client on search has to error if there is a problem?

sunpy/net/fido_factory.py
```diff
@@ -189,6 +211,8 @@ def __str__(self):
             if block.client.info_url is not None:
                 ret += f'Source: {block.client.info_url}\n'
             size = block.total_size()
+            if self.errors[block.client.__name__] is not None:
+                ret += f'Error: {repr(self._errors[block.client.__name__])}\n'
```
nabobalis, line 215: How does this look on output? This also will need a test.

sunpy/net/fido_factory.py
nabobalis: So if there is an error in the client search, the search returns an Exception?

sunpy/net/fido_factory.py
tanishy7777: > For this to work, the client on search has to error if there is a problem?
It will cause an error even if the `client()` method fails like during its initialization, and yes it also fails if `search` method fails.

sunpy/net/tests/test_fido.py
tanishy7777: > Does this also trigger multiple clients?

Yes it does run for all clients, the only change is that the VSO client's search function now returns `ConnectionError`

tanishy7777: > Does this also trigger multiple clients?
image

sunpy/net/fido_factory.py
tanishy7777: image
for example this ConnectionError is raised because in the build method specifically in `vso.py`
image

sunpy/net/fido_factory.py
tanishy7777: image

tanishy7777: I am not sure how to test this ?

nabobalis: Try to capture the output with pytest and see if that works?

tanishy7777: > Try to capture the output with pytest and see if that works?

Like compare the output to the entire string?

nabobalis: Sure, lets try that.

sunpy/net/fido_factory.py
tanishy7777: > So if there is an error in the client search, the search returns an Exception?

yes

nabobalis: How does that work?

sunpy/net/fido_factory.py
tanishy7777: > Sure, lets try that.

Got it will add that

sunpy/net/fido_factory.py
tanishy7777: > How does that work?

I don't understand what do you mean?

nabobalis: I am not familiar with the internals of fido but why are the clients returning errors and not raising them?

tanishy7777: > I am not familiar with the internals of fido but why are the clients returning errors and not raising them?

The clients are raising the Exceptions right? Thats why we can capture them using try-except blocks?

tanishy7777: The flow is something like this 

on calling `Fido.search()` this is method is called from `UnifiedDownloadFactory`(Fido is an instance of this class) which first parses the queries using the attribute walker which then calls the `_make_query_to_client` function after processing the queries into the right format
```python
def search(self, *query):
    query = attr.and_(*query)
    results = query_walker.create(query, self)

    results = [r for r in results if not isinstance(r, vso.VSOQueryResponseTable) or len(r) > 0]
    return UnifiedResponse(*results)
 ```
 
 `make_query_to_client`
This gets all the clients from the registry, and runs **their search method** on the queries. 
If any error is **raised** by the client due to `ConnectionError` or something then try except block captures it.
```python
def _make_query_to_client(self, *query):
    candidate_widget_types = self._check_registered_widgets(*query)
    results = []
    for client in candidate_widget_types:
        try:
            tmpclient  = client()
            result = tmpclient.search(*query)
            result.client = client
            results.append(result)
        except Exception as e:
            e.client = client
            results.append(e)

    return results
```

sunpy/net/fido_factory.py
tanishy7777: > Sure, lets try that.

What if in the future more clients fail due to external reasons. But if we compare the strings directly our tests will fail?

nabobalis: Why would that affect the test output?

tanishy7777: > Why would that affect the test output?

Like there will be more errors in the string?

nabobalis: But why would the test output change?

tanishy7777: Like as far as I understand we are comparing the 
results.__str__ to the entire output string we are expecting using assert?

Or do we check if res.__str__ has "Error" in that, maybe that would work?

nabobalis: But we construct the test, so why would the output change if we add more clients?

tanishy7777: image
Are you suggesting something like this?

tanishy7777: The memory location at which the object is created changes everytime that is also a problem

tanishy7777: image
I was suggesting something like this

tanishy7777: PS: please ignore the file write and print statements I was just testing things out

tanishy7777: > Can assert that the string is in str(results)?
added

changelog/8009.feature.rst
```diff
@@ -0,0 +1,2 @@
+Enhanced :meth:`~sunpy.net.Fido.search` to be robust against individual client failures. If a single client fails to respond,
+the search will continue with other clients, ensuring that the functionality is not disrupted by errors from a single external client.
```
nabobalis, line 2:
```suggestion
Enhanced :meth:`~sunpy.net.Fido.search` to be robust against individual client failures.
If a single client fails to respond, the search will continue with the other clients.
```

sunpy/net/tests/test_fido.py
```diff
@@ -65,8 +65,8 @@ def test_offline_fido(query):
 
 
 @pytest.mark.remote_data
-# Until we get more mocked, we can't really do this to online clients.
-# TODO: Hypothesis this again
+# # Until we get more mocked, we can't really do this to online clients.
+# # TODO: Hypothesis this again
```
nabobalis, line 69: Can this be fixed?

sunpy/net/tests/test_fido.py
```diff
@@ -76,6 +76,14 @@ def test_online_fido(query):
     unifiedresp = Fido.search(query)
     check_response(query, unifiedresp)
```
nabobalis, line 78:
```suggestion


```

sunpy/net/fido_factory.py
```diff
@@ -53,16 +53,28 @@ def __init__(self, *results):
         """
         self._list = []
         self._numfile = 0
+        self._errors = {}
         for result in results:
-            if isinstance(result, QueryResponseRow):
-                result = result.as_table()
+            if isinstance(result, Exception):
+                if hasattr(result.client.__class__, '__name__'):
```
nabobalis, line 59: Why would this not trigger?

sunpy/net/fido_factory.py
nabobalis: As Shane says, we should update `__getitem__`

sunpy/net/fido_factory.py
```diff
@@ -53,16 +53,28 @@ def __init__(self, *results):
...
+                    self._errors[result.client.__class__.__name__] = result
 
-            if isinstance(result, QueryResponseColumn):
-                result = result.as_table()
+                result = QueryResponseTable([], client=result.client)
```
nabobalis, line 62: So we have to return an empty result for the failed client?

sunpy/net/fido_factory.py
```diff
@@ -53,16 +53,28 @@ def __init__(self, *results):
...
+            else:
```
nabobalis, line 63: Is there a need to have the if/else? instead of just if statements?

sunpy/net/fido_factory.py
```diff
@@ -162,6 +174,16 @@ def file_num(self):
         """
         return self._numfile
 
+    @property
+    def errors(self):
+        """
+        Returns a dict of errors for all responses.
+
+        If no errors are present for a client, its value is None.
```
nabobalis, line 182: Is this still true?

sunpy/net/fido_factory.py
```diff
@@ -478,11 +501,15 @@ def _make_query_to_client(self, *query):
         candidate_widget_types = self._check_registered_widgets(*query)
         results = []
         for client in candidate_widget_types:
-            tmpclient = client()
-            results.append(tmpclient.search(*query))
+            try:
+                tmpclient  = client()
+                result = tmpclient.search(*query)
+                # result.client = client
```
nabobalis, line 507:
```suggestion
```

sunpy/net/fido_factory.py
```diff
@@ -478,11 +501,15 @@ def _make_query_to_client(self, *query):
...
+                results.append(result)
+            except Exception as e:
+                e.client = client
```
nabobalis, line 510: This adds a client property to the exception?

tanishy7777: > This adds a client property to the exception?

yes

sunpy/net/fido_factory.py
tanishy7777: > Why would this not trigger?

Trigger when?

sunpy/net/fido_factory.py
nabobalis: The reason being?

sunpy/net/fido_factory.py
nabobalis: The if statement, why would a result, not have a client attached?

sunpy/net/fido_factory.py
tanishy7777: > The reason being?

To add error for the client in the error dictionary and also to print this in __repr__

this is in the __init__ method of UnifiedResponse
```python
if hasattr(result.client, '__name__'):
                    print(f"Error: {result} for {result.client.__name__}")
                    self._errors[result.client.__name__] = result
```

sunpy/net/tests/test_fido.py
tanishy7777: > Can this be fixed?

Yep, as far as I understood, currently the `query` passes a fixed set of parameters. But using `Hypothesis` we want to generate more tests based on some specifcation/strategy right?

So I would possibly create a custom strategy for this?

I didn't understand what the first line is saying though -> "Until we get more mocked, we can't really do this to online clients."

sunpy/net/fido_factory.py
tanishy7777: Yep,
On running,
```python
results = Fido.search((a.Time('2020-01-01', '2020-01-01 00:00:10') | a.Time('2020-01-03', '2020-01-03 00:00:10')) &
                a.Instrument('AIA'))
print("hi", results.errors)
```
image

sunpy/net/fido_factory.py
tanishy7777: Nope its not necessary, I just added it for being explicit about the separation that if an Exception/error has occured it will never go into the other if statements that would have executed if "normal" result was generated. But yep its probably not the best way to handle it should I remove it?

sunpy/net/fido_factory.py
nabobalis: Why do we need an empty table?

sunpy/net/fido_factory.py
```diff
@@ -53,16 +53,30 @@ def __init__(self, *results):
         """
         self._list = []
         self._numfile = 0
+        self._errors = {}
         for result in results:
-            if isinstance(result, QueryResponseRow):
-                result = result.as_table()
+            if isinstance(result, Exception):
+                if hasattr(result.client, '__name__'):
+                    print(f"Error: {result} for {result.client.__name__}")
```
nabobalis, line 60: Remove print statements, if they are debug messages, they should be passed to the logger, if they are important, they should be either raised or passed to the logger as well.

sunpy/net/fido_factory.py
nabobalis: Yes, can you remove it and simplify the nesting please.

sunpy/net/fido_factory.py
nabobalis: We should not do this. There is no need to set the client to be there if it didnt fail.

sunpy/net/fido_factory.py
```diff
@@ -189,6 +213,11 @@ def __str__(self):
             if block.client.info_url is not None:
                 ret += f'Source: {block.client.info_url}\n'
             size = block.total_size()
+
+            if hasattr(block.client, '__name__'):
+                if self.errors[block.client.__name__] is not None:
```
nabobalis, line 218: Does a client never have a name?

sunpy/net/fido_factory.py
```diff
@@ -478,11 +506,16 @@ def _make_query_to_client(self, *query):
...
+                result = tmpclient.search(*query)
+                results.append(result)
+            except Exception as e:
+                print(f"{client} will be excluded from the search results.")
+                print(client.__name__, client.info_url)
+                e.client = client   # client class not instance
+                results.append(e)
 
-        # This method is called by `search` and the results are fed into a
-        # UnifiedResponse object.
```
nabobalis, line 485: Why was this comment removed? 

Feels like it should be at the top of the method tho.

sunpy/net/fido_factory.py
```diff
@@ -478,11 +506,16 @@ def _make_query_to_client(self, *query):
...
-            results.append(tmpclient.search(*query))
+            try:
+                tmpclient  = client()   # client instance
+                result = tmpclient.search(*query)
+                results.append(result)
+            except Exception as e:
+                print(f"{client} will be excluded from the search results.")
+                print(client.__name__, client.info_url)
+                e.client = client   # client class not instance
+                results.append(e)
```
nabobalis, line 517: I would remove the comments, as before, the prints should either be useful and be provided to the logger or they should be removed.

sunpy/net/tests/test_fido.py
nabobalis: Fair enough, we should probably do this in a new PR. 

In the mean time, the line needs to be reverted back to normal.

sunpy/net/tests/test_fido.py
```diff
@@ -77,6 +77,16 @@ def test_online_fido(query):
...
 
 
+@pytest.mark.remote_data
+@mock.patch("sunpy.net.vso.vso.VSOClient.search", side_effect=ConnectionError('VSO is down'))
+def test_fido_client_error(vso_search):
+    results = Fido.search(a.Time("2016/10/01", "2016/10/02"), a.Instrument.aia)
+    assert len(results.errors) > 0
+    print(results.errors)
```
nabobalis, line 85:
```suggestion
```

sunpy/net/tests/test_fido.py
tanishy7777: > Fair enough, we should probably do this in a new PR.

Cool should I open a new PR? or an issue first?

nabobalis: Open a PR

sunpy/net/fido_factory.py
tanishy7777: > Nope its not necessary, I just added it for being explicit about the separation that if an Exception/error has occured it will never go into the other if statements that would have executed if "normal" result was generated. But yep its probably not the best way to handle it should I remove it?

done

sunpy/net/fido_factory.py
tanishy7777: > We should not do this. There is no need to set the client to be there if it didnt fail.

In https://github.com/sunpy/sunpy/issues/7625#issuecomment-2114602574 it was mentioned that `None` as a placeholder would be nice to have, so I added it.

sunpy/net/fido_factory.py
tanishy7777: > Remove print statements, if they are debug messages, they should be passed to the logger, if they are important, they should be either raised or passed to the logger as well.

I checked the `sunpyrc` file by default the log file is commented. So where should I log the message to?


```python
# The file to log messages to
# log_file_path = sunpy.log
```

sunpy/net/fido_factory.py
tanishy7777: > Why do we need an empty table?

because we need to pass the client attribute for the `_repr_` , `__str__` function to work properly.
It uses 'block.client' so for the clients that error attaching an empty table with a client attribute is needed.
The clients that dont error have a client attribute for the table by deafult.


```python
   def __str__(self):
        nprov = len(self)
        if nprov == 1:
            ret = f'Results from {len(self)} Provider:\n\n'
        else:
            ret = f'Results from {len(self)} Providers:\n\n'
        for block in self:
            ret += f"{len(block)} Results from the {block.client.__class__.__name__}:\n"
            if block.client.info_url is not None:
                ret += f'Source: {block.client.info_url}\n'
            size = block.total_size()

            if hasattr(block.client, '__name__'):
                if self.errors[block.client.__name__] is not None:
                    ret += f'Error: {repr(self._errors[block.client.__name__])}\n'

            if np.isfinite(size):
                ret += f'Total estimated size: {size}\n'
            ret += '\n'
            lines = repr(block).split('\n')
            ret += '\n'.join(lines[1:])
            ret += '\n\n'

        return ret


```

sunpy/net/fido_factory.py
tanishy7777: I have removed it as per your other suggestions for cleaner code.

sunpy/net/fido_factory.py
tanishy7777: > Does a client never have a name?

Yes,
the reason I have added this is because of the way `_make_query_to_client` handles errors

```python
     candidate_widget_types = self._check_registered_widgets(*query)
        results = []
        for client in candidate_widget_types:
            try:
                tmpclient  = client()  # Instance of the client
                results.append(tmpclient.search(*query))
            except Exception as e:
                e.client = client       # client class
                results.append(e)

        return results
```

If there is an error then the `class` is added to the `client attribute` to which we can directly call `client.__name__`
but if the result is "normal" then the `instance` of the client is passed, which would give an `AttributeError` if we directly call `client.__name__`

So what this does is that `ret` only needs to have the error statement if `class` is passed from a client. If an `instance` is passed then it will not give an error so we dont check that.
```python
if hasattr(block.client, '__name__'):
    if self.errors[block.client.__name__] is not None:
        ret += f'Error: {repr(self._errors[block.client.__name__])}\n'
```

sunpy/net/fido_factory.py
tanishy7777: > Why was this comment removed?
> 
> Feels like it should be at the top of the method tho.

Added

sunpy/net/fido_factory.py
tanishy7777: > I would remove the comments, as before, the prints should either be useful and be provided to the logger or they should be removed.

Makes sense will remove the comments and removed the print statements

sunpy/net/tests/test_fido.py
```diff
@@ -76,6 +76,14 @@ def test_online_fido(query):
     unifiedresp = Fido.search(query)
     check_response(query, unifiedresp)
 
+@pytest.mark.remote_data
+@mock.patch("sunpy.net.vso.vso.VSOClient.search", side_effect=ConnectionError('VSO is down'))
+def test_fido_client_error(vso_search):
+    results = Fido.search(a.Time("2016/10/01", "2016/10/02"), a.Instrument.aia)
```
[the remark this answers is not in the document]
tanishy7777: > Can we make a query that naturally will return results from two clients and make one of those clients fail?
I am not sure how to do this, should I ask on chat?

sunpy/net/fido_factory.py
nabobalis: There is a logger: https://docs.sunpy.org/en/latest/dev_guide/contents/logger.html

sunpy/net/fido_factory.py
nabobalis: I dont agree.

sunpy/net/tests/test_fido.py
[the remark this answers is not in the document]
nabobalis: Sure

sunpy/net/fido_factory.py
nabobalis: Can we use the class name all the time?

sunpy/net/fido_factory.py
```diff
@@ -53,13 +53,24 @@ def __init__(self, *results):
         """
         self._list = []
         self._numfile = 0
+        self._errors = {}
         for result in results:
+            if isinstance(result, Exception):
+                print(f"Error: {result} for {result.client.__name__}") # TODO: pass this to logger
```
nabobalis, line 59: This needs to be logged.

Cadair:
hey @tanishy7777 are you able to finish this up or should someone else take it over?

samaloney:
@tanishy7777 thanks for all the work after looking at this PR and the code I've come up with an alternative approach https://github.com/samaloney/sunpy/pull/new/feat-fido-client-error

I was trying to update this PR to keep your contributions but I don't seem to have the permissions.

samaloney:
So I've updated this PR to:

* Merge curent main
* Pull the `.errors` property from a specific VSO response table `VSOQueryResponseTable` up to the generic response table `QueryResponseTable`
* Updated `_make_query_to_client` to wrap client calls in try catch and add any error to a newly created `QueryResponseTable` via the error property.
* Restore original `UnifiedResponse`  `__init__`
* Add `.errors` property to `UnifiedResponse` class which will list all errors from all clients

Now give `res = Fido.search(...)`
`res.errors` will give a list of lists of all errors from all clients
`res[0].errors` or `res['vso'].errors` will give errors for that client  

Not sure if we want a keyword to control if errors are raised or not but would be pretty handy to add this later if a need arose.

sunpy/net/dataretriever/sources/lyra.py
```diff
@@ -44,7 +44,6 @@ class LYRAClient(GenericClient):
     2016-01-01 09:41:00.000 2016-01-01 10:40:00.000 PROBA2       LYRA 1200.0 .. 1230.0      ESA irradiance      euv         N/A  2328.75
     2016-01-01 09:41:00.000 2016-01-01 10:40:00.000 PROBA2       LYRA 1200.0 .. 1230.0      ESA irradiance      euv         N/A 419.0625
     <BLANKLINE>
-    <BLANKLINE>
```
nabobalis, line 35: Should this be gone?

changelog/8009.feature.rst
```diff
@@ -0,0 +1 @@
+Enhanced :meth:`~sunpy.net.Fido.search` to be robust against client errors. Any client error will be caught and attached to ``.errors`` property on both the `~sunpy.net.fido_factory.UnifiedResponse` and individual `~sunpy.net.base_client.QueryResponseTable`, the search will continue.
```
nabobalis:
```suggestion
Enhanced :meth:`~sunpy.net.Fido.search` to be robust against client errors.
Any client error will be caught and attached to ``.errors`` property on both the `~sunpy.net.fido_factory.UnifiedResponse` and individual `~sunpy.net.base_client.QueryResponseTable`, the search will continue.
```

sunpy/net/tests/test_fido.py
```diff
@@ -144,9 +152,10 @@ def test_unified_response():
 
 @pytest.mark.remote_data
 def test_no_match():
-    with pytest.raises(DrmsQueryError):
-        Fido.search(a.Time("2016/10/01", "2016/10/02"), a.jsoc.Series("bob"),
-                    a.Sample(10*u.s))
+    res = Fido.search(a.Time("2016/10/01", "2016/10/02"), a.jsoc.Series("bob"),
+                a.Sample(10*u.s))
+    assert res.errors
+    assert isinstance(res[0].errors, DrmsQueryError)
```
nabobalis, line 160: Can we also check the message in the error?

This was an oversight in the original test.

sunpy/net/fido_factory.py
```diff
@@ -465,6 +478,9 @@ def _make_query_to_client(self, *query):
         """
         Given a query, look up the client and perform the query.
 
+        This method is called by `search` and the results are fed into a
```
nabobalis:
```suggestion
        This method is called by ``search`` and the results are fed into a
```

sunpy/net/dataretriever/sources/lyra.py
samaloney: I'm not sure the local and CI tests were failing and traced it to this but doesn't really make sense

sunpy/net/tests/test_fido.py
samaloney: Done

Cadair:
I will review this as soon as I get the chance.

sunpy/net/fido_factory.py
```diff
@@ -162,6 +162,13 @@ def file_num(self):
...
         return self._numfile
 
+    @property
+    def errors(self):
+        """
+        Returns a list of errors for each client.
+        """
+        return [res.errors for res in self._list if res.errors]
+
     def _repr_html_(self):
```
Cadair, line 172: You should add errors to this `_repr_html_` as well.

sunpy/net/fido_factory.py
```diff
@@ -465,6 +478,9 @@ def _make_query_to_client(self, *query):
         """
         Given a query, look up the client and perform the query.
 
+        This method is called by ``search`` and the results are fed into a
+        UnifiedResponse object.
```
Cadair:
```suggestion
        `~sunpy.net.UnifiedResponse` object.
```

Cadair (changes requested):
This looks good, but it needs some narrative documentation to teach people about how to look at and deal with the errors.

tanishy7777:
> @tanishy7777 thanks for all the work after looking at this PR and the code I've come up with an alternative approach https://github.com/samaloney/sunpy/pull/new/feat-fido-client-error
> 
> I was trying to update this PR to keep your contributions but I don't seem to have the permissions.

Thank you for that! I was a bit busy with college and wasnt able to look at this.
I am assuming that now you have got the permissions to commit?

samaloney:
So the test fails seem unrelated and I don't believe the coverage report 👀

Cadair:
Let's merge #7834 and then see what happens here.

sunpy/net/fido_factory.py
```diff
@@ -535,10 +553,13 @@ def _make_query_to_client(self, *query):
         results = []
         for client in candidate_widget_types:
             tmpclient = client()
-            results.append(tmpclient.search(*query))
+            try:
+                res = tmpclient.search(*query)
+            except Exception as err:
+                res = QueryResponseTable([], client=tmpclient, errors=err)
+
```
nabobalis:
```suggestion
```

nabobalis:
**NOTE TO WHOEVER MERGES THIS, SQUASH MERGE**

docs/tutorial/acquiring_data/index.rst
```diff
@@ -245,6 +245,7 @@ In this example we'll search for LYRA or RHESSI data in a given time range:
     <BLANKLINE>
     <BLANKLINE>
 
+Failed client errors be reported in the returned `~sunpy.net.fido_factory.UnifiedResponse` and also on the individual `~sunpy.net.base_client.QueryResponseTable` responses in the ``.errors`` attribute.
```
Cadair:
```suggestion
Failed client errors are reported in the returned `~sunpy.net.fido_factory.UnifiedResponse` and also on the individual `~sunpy.net.base_client.QueryResponseTable` responses in the ``.errors`` attributes.
```

samaloney:
So that specific error seems to only happen sometimes I don't see it in the latest run (https://github.com/sunpy/sunpy/actions/runs/19505650737/job/55842462717?pr=8009) the other test fails I can't reproduce locally but think they are due to rate limits / timeouts.

Issue #6823, opened 2023-03-06
`map` can't handle certain units (e.g., `u.electron`) in memory due to limitation of FITS


starfleetjames:
### Describe the bug

I'm trying to do some manipulation of a `map`, all in memory, to move from an MHD simulation, through an instrument simulator, and ultimately to saving a FITS file on disk in units of `u.dN`. In that process, there is a point where I am multiplying a `map` with `Unit("ph / (Angstrom cm2 s sr)")` by a quantum yield of the detector with `Unit("electron / ph")`. But that results in this error: 
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/masonjp2/opt/anaconda3/envs/suncet/lib/python3.10/site-packages/sunpy/util/decorators.py", line 403, in inner
    return func(instance, value)
  File "/Users/masonjp2/opt/anaconda3/envs/suncet/lib/python3.10/site-packages/sunpy/map/mapbase.py", line 566, in __mul__
    return self._new_instance_from_op(new_data)
  File "/Users/masonjp2/opt/anaconda3/envs/suncet/lib/python3.10/site-packages/sunpy/map/mapbase.py", line 538, in _new_instance_from_op
    new_meta['bunit'] = new_data.unit.to_string('fits')
  File "/Users/masonjp2/opt/anaconda3/envs/suncet/lib/python3.10/site-packages/astropy/units/core.py", line 739, in to_string
    return f.to_string(self)
  File "/Users/masonjp2/opt/anaconda3/envs/suncet/lib/python3.10/site-packages/astropy/units/format/fits.py", line 113, in to_string
    unit = utils.decompose_to_known_units(unit, cls._get_unit_name)
  File "/Users/masonjp2/opt/anaconda3/envs/suncet/lib/python3.10/site-packages/astropy/units/format/utils.py", line 103, in decompose_to_known_units
    new_unit = new_unit * decompose_to_known_units(base, func) ** power
  File "/Users/masonjp2/opt/anaconda3/envs/suncet/lib/python3.10/site-packages/astropy/units/format/utils.py", line 107, in decompose_to_known_units
    func(unit)
  File "/Users/masonjp2/opt/anaconda3/envs/suncet/lib/python3.10/site-packages/astropy/units/format/fits.py", line 107, in _get_unit_name
    cls._validate_unit(name)
[...]

PR #8037, opened 2025-01-24
Fixed Map logic for permitting DN as a unit

ayshih:
This PR fixes a bug in the code that allows DN as a Map unit (#7585).  The existing code has a logical flaw: it should be allowing DN as the only non-FITS unit, but instead it allows any combination of non-FITS units as long as one of the units is DN.  For example, while a map with unit "electron" is not allowed, a map with unit "DN / electron" is mistakenly allowed because it includes DN as one of its non-FITS units.

This bugfix has the unfortunate consequence of making `Map` even less permissive by blocking all non-FITS unit combinations (aside from DN), but I feel we ought to fix the inconsistent behavior in 6.0 (see https://github.com/sunpy/sunpy/issues/6823#issuecomment-2610618498).

### DIFF
--- a/changelog/8037.bugfix.rst
+++ b/changelog/8037.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed the unintended `~sunpy.map.Map` behavior where any combination of non-FITS units were allowed as long as one of the non-FITS units was DN.
+DN is currently the only non-FITS unit permitted in `~sunpy.map.Map`.
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -794,18 +794,26 @@ class GenericMap(NDData):
                         'counts / pixel': 'ct/pix',}
         if unit_str.lower() in replacements:
             unit_str = replacements[unit_str.lower()]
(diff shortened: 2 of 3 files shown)


ayshih:
I welcome discussion about whether this should be merged!  This PR fixes the unintended behavior so it is formally the correct thing to do, but it could potentially be disruptive to any user who happened to be taking advantage of our surprise permissiveness of non-FITS units starting with 6.0.

wtbarnes:
I don't agree with only allowing DN explicitly as the only non-FITS standard unit.

#7585 certainly had unintended consequences as illustrated in the linked comment, but in that PR I did intentionally allow for units *with* DN, not just only DN. The main one I had in mind of course is DN/s, a common unit for any L1 data product normalized to an exposure time. To disallow in 6.1 would be an extremely annoying breaking change.

At the very least, I would suggest also special-casing "DN/s" here. I was trying to avoid an allowed-list of DN-related units in the original PR, but maybe that is the only way.

Cadair (approved):

wtbarnes (approved):

PR #8040, opened 2025-01-26
Option to automatically determine extent when reprojecting a map

### LINKED ISSUE #5858: Add the option to automatically determine the extent of a map when using `reproject_to()`
Based on a discussion on Element and the proof of concept [here](https://colab.research.google.com/drive/1NU3ooz8KNt8zURcrQDKr-yyxTzY_j8CH?usp=sharing), it'd be neat to have the option for `GenericMap.reproject_to()` to automatically determine the extent of the reprojected map.  That is, the CRPIX and NAXIS values would be ignored in the target WCS, and new CRPIX and NAXIS values are calculated to give the tightest extent that contains all of the reprojected data.  There a number of considerations that the proof of concept disregards, so a general implementation needs more thought.

There's also the question of whether we ought to add an option to `reproject_to()`, since the eventual plan is for it to be silently replaced by `NDCube`'s `reproject_to()`.  I'd imagine that implementing such an option in `NDCube`'s `reproject_to()` might be even more fiddly.

ayshih:
This PR adds the option for `GenericMap.reproject_to()` to automatically determine the extent of the output map (closes #5858).  One of the sticking points had been how to code it generally, because sometimes it's sufficient to check only the corners of the input map, but other times you have to check all of the pixels.  I've had the epiphany that we can simply let the user choose the approach, because they know their data.  Here are the options:

- "corners" (fastest) works for a map that is all on disk and the reprojection does not distort the edges
- "edges" (okay) works for a map that is all on disk and the reprojected map is all visible
- "all" (slowest) is the safest because it transforms every pixel (specifically, the corners of every pixel)

Even "all" may not be that noticeable given the performance of the subsequent reprojection.


ayshih:
Here are some image examples for an extreme case of reprojecting an all-on-disk map to a different observer such that part of the map is behind the limb.

`auto_extent=None` (note the wasted space)
download (81)

`auto_extent="corners"` (very bad choice here)
download (82)

`auto_extent="edges"` (some data is still lost near the east limb)
download (83)

`auto_extent="all"` (all data preserved)
download (84)

Cadair:
@ayshih we are ~1 month out from feature freeze, do you want this on the 7.0 milestone?

Cadair:
@astrofrog linked me this as prior art: https://github.com/astropy/reproject/blob/7ece46ccd111a0a324dc44ff1bcf7579428addda/reproject/array_utils.py#L209

ayshih:
> @ayshih we are ~1 month out from feature freeze, do you want this on the 7.0 milestone?

Sure

> @astrofrog linked me this as prior art: https://github.com/astropy/reproject/blob/7ece46ccd111a0a324dc44ff1bcf7579428addda/reproject/array_utils.py#L209

Heh, I didn't know that `sample_array_edges()` existed (it is not documented).  The function can handle a general N-D WCS, in a mostly efficient way, but as reinforced by the function name, it samples the edges rather than using the actual corner pixels.  I would insist that we do the latter.  Of course, `sample_array_edges()` could easily be extended, but I'm simply saying that I would not use the function as is in this PR.

Incidentally, `sample_array_edges()` is an internal function for `reproject_and_coadd()`, which performs smart reprojection of each map based on its sampled edges.  I'll note that by design that the smart reprojection is limited to determining which subregion of the overall extent specified in the target WCS will actually have data from each map.  There is no smart-ness in the sense of shrinking or expanding the overall extent that had been specified in the target WCS.

astrofrog:
Just a quick note that sample_array_edges can do corners by setting n_samples=2

ayshih:
> Just a quick note that sample_array_edges can do corners by setting n_samples=2

Oops, when I wrote "it samples the edges rather than using the actual corner pixels", I should have written "it samples the edges rather than using _the corners of the actual edge pixels_".  That is, I want to use the exact pixel definitions for the "edges" setting in this PR.  For a square array, one can call `sample_array_edges()` with `n_samples=pixels_on_a_side+1`.  However, there's no way to achieve what I want for a rectangular array.

Again, `sample_array_edges()` could easily be enhanced.  It simply doesn't currently do what I need.

sunpy/map/mapbase.py
```diff
@@ -3067,6 +3080,32 @@ def reproject_to(self, target_wcs, *, algorithm='interpolation', return_footprin
...
+        if auto_extent is not None:
+            ny, nx = self.data.shape
+            if auto_extent == 'all':
+                pixels = np.indices((nx + 1, ny + 1)) - 0.5
+            elif auto_extent == 'edges':
+                pixels = [[np.full(ny, -0.5), np.arange(ny) - 0.5],  # left edge
+                          [np.arange(nx) - 0.5, np.full(nx, ny - 0.5)],  # top edge
+                          [np.full(ny, nx - 0.5), np.arange(ny, 0, -1) - 0.5],  # right edge
+                          [np.arange(nx, 0, -1) - 0.5, np.full(nx, -0.5)]]  # bottom edge
+                pixels = np.hstack(pixels)
```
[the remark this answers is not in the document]
ayshih: I've now made a shared function in `sunpy.util`.

Tangential to this PR, but I want to nuke `sunpy.map.map_edges`.

sunpy/map/mapbase.py
```diff
@@ -3020,6 +3022,12 @@ def reproject_to(self, target_wcs, *, algorithm='interpolation', return_footprin
         return_footprint : `bool`
             If ``True``, the footprint is returned in addition to the new map.
             Defaults to ``False``.
+        auto_extent : ``"all"``, ``"edges"``, ``"corners"``, or ``None``
+            If ``None``, the extent of the reprojected map comes from the target WCS.
+            If not ``None``, the extent of the reprojected map is automatically
+            determined by transforming all of the pixels, just the edges, or just the
+            corners of this map.
+            Defaults to ``None``.
```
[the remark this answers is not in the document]
ayshih: Added words

sunpy/map/mapbase.py
```diff
@@ -3000,6 +3001,7 @@ def _check_axes(self, axes, warn_different_wcs=False):
         return axes
 
     def reproject_to(self, target_wcs, *, algorithm='interpolation', return_footprint=False,
+                     auto_extent: Literal[None, 'corners', 'edges', 'all'] = None,
```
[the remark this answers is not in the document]
Cadair: I don't mind useful type hints like this in general, I do however kind of object to duplicating the information in the docstring. If we are going to do this we should have the docstring parse the type hint if there isn't type info in the docstring. i.e. using this: https://github.com/tox-dev/sphinx-autodoc-typehints

sunpy/map/mapbase.py
```diff
@@ -3096,6 +3111,39 @@ def reproject_to(self, target_wcs, *, algorithm='interpolation', return_footprin
...
+                pixels = np.indices((nx + 1, ny + 1)) - 0.5
+            elif auto_extent == 'edges':
+                pixels = grid_perimeter(nx, ny).T - 0.5
+            elif auto_extent == 'corners':
+                pixels = [[-0.5, -0.5, nx - 0.5, nx - 0.5], [-0.5, ny - 0.5, ny - 0.5, -0.5]]
+            else:
+                raise ValueError("The allowed options for auto_extent are 'all', 'edges', 'corners', or None.")
+
+            coords = self.wcs.pixel_to_world(*pixels)
+            out_xpixels, out_ypixels = target_wcs.world_to_pixel(coords)
```
Cadair, line 3126: you could pixel_to_pixel here as well.

sunpy/util/util.py
```diff
@@ -319,3 +322,30 @@ def fix_duplicate_notes(subclass_doc, cls_doc):
         cls_doc += textwrap.indent(subclass_doc, "    ")
 
     return cls_doc
+
+
+def grid_perimeter(nx, ny):
```
Cadair, line 327: I've needed this function a bunch of times. It should be in numpy.

sunpy/map/tests/test_reproject_to.py
```diff
@@ -139,3 +140,37 @@ def test_reproject_to_warn_using_contexts(aia171_test_map, hpc_header):
             # Check if a warning is raised if both context managers are used at the same time.
             with pytest.warns(UserWarning, match="Using propagate_with_solar_surface and SphericalScreen together result in loss of off-disk data."):
                 aia171_test_map.reproject_to(hpc_header)
+
+
+@figure_test
+@pytest.mark.parametrize('auto_extent', [None, 'corners', 'edges', 'all'])
+def test_reproject_to_auto_extent(aia171_test_map, auto_extent):
```
Cadair, line 147: Could we not write a numerical test with a known WCS rather than just a figure test?

sunpy/map/mapbase.py
```diff
@@ -3096,6 +3111,39 @@ def reproject_to(self, target_wcs, *, algorithm='interpolation', return_footprin
             raise ValueError(f"The specified algorithm must be one of: {list(functions.keys())}")
         func = functions[algorithm]
 
+        if auto_extent is not None:
```
Cadair, line 3114: Could we move this to a helper function? I think that it could be useful for people to use with ndcube objects (i.e. use it to calculate adjustments to more complex WCSes).

and you know end up in astropy or reproject :stuck_out_tongue:

sunpy/map/tests/test_reproject_to.py
ayshih: Done

Cadair (changes requested):
Looks good, but I would like to see [this](https://github.com/sunpy/sunpy/pull/8040#discussion_r2077803638) comment about a helper function addressed.

sunpy/map/mapbase.py
ayshih: Moved out to be `sunpy.util.extent_in_other_wcs()`

Cadair (approved):

sunpy/map/mapbase.py
astrofrog: This would definitely be useful in reproject!

nabobalis: LETS GO UPSTREAM

sunpy/util/util.py
```diff
@@ -347,3 +349,58 @@ def grid_perimeter(nx, ny):
              [np.full(ny, nx), np.arange(ny, 0, -1)],  # right edge
              [np.arange(nx, 0, -1), np.zeros(nx)]]  # bottom edge
     return np.hstack(edges).T
+
+
+def extent_in_other_wcs(original_wcs, target_wcs, *, method, original_shape=None, integers=False):
```
nabobalis, line 354: Do we care about some tests for this?

nabobalis (approved):

PR #8043, opened 2025-01-28
Add a hacky script for calculating version bumps

Cadair:
I am not super proud of this, but I got fed up with manually calculating SPEC 0 dependency bumps.

We could (and should?) adjust this to optionally follow the sunpy policy (i.e. astropy versions etc) but this should also remain compatible with other packages.


nabobalis:
Lets do it.

nabobalis (approved):

PR #8045, opened 2025-01-28
Datamanager.cache for Timeseries url handling 

### LINKED ISSUE #2088: Map and Timeseries factories should have better support for urls
we should probably use the astropy download_file machinery for the factory classes, and specifically support passing through the cache keyword argument so the data is not re-downloaded everytime the object is created.

Sauravroy34:
Adds cache for timeseries
fixes #2088


nabobalis:
So how does this differ to Map?

Sauravroy34:
well map factory datamanager cache is used 
https://github.com/Sauravroy34/sunpy/blob/dc6732598a7d13593a7d570978d998fb09fa265e/sunpy/map/map_factory.py#L243-L248


and difference here is that 
-> there is not database (.db) file 
-> path of the cache is different (/.sunpy/cache/download/url) which cannot be changed 
-> The downloaded URL is saved in a hasheddir (which has two components contents and URL)
-> if cache is true then the file is searched in cache dir if it's not present then it is downloaded 
-> if cache is false then contents are stored in tempdir

In this code if cache is true
file is downloaded in cache dir and content is copied into the required file

if it is false
then the file is downloaded in tempdir(eventually it is not saved so no memory redundancy) and content is copied as mentioned  above

nabobalis:
So what does map do to cache files?

Sauravroy34:
> So what does map do to cache files?

it does the something as datamanager.cache does 

```       
  path = str(cache.download(url).absolute())
  pairs = self._read_file(path, **kwargs)
  return pairs
```

here cache dir at sunpy/data/__init__.py (https://github.com/Sauravroy34/sunpy/blob/astropy_download_file_support/sunpy/data/__init__.py)

```
cache = Cache(
    ParfiveDownloader(),
    SqliteStorage(CACHE_DIR + '/cache.db'),
    CACHE_DIR,
    expiry=int(config.get('downloads', 'cache_expiry')) * u.day
)
```
so files are stored in the cache dir  (I guess it is the default Sunpy download directory)
if they are not present they are download (hash of file are stored in database for content comparison)
if they are present they will be cached

nabobalis:
I guess I am just confused without looking at the logic of both timeseries and Map, how they differ, how they handle URLs and cache the data they download and how these code changes help.

Sauravroy34:
well as of now of now timeseries URL functionality does not provide cache mechanism (and we have no option for progress bar)  but map has 
this pr uses astropy download_file for downloading and having a optional cache

nabobalis:
> well as of now of now timeseries URL functionality does not provide cache mechanism (and we have no option for progress bar) but map has this pr uses astropy download_file for downloading and having a optional cache

What is map using that timeseries is not?

Sauravroy34:
Map is using datamanger s cache functionality (https://github.com/sunpy/sunpy/blob/main/sunpy/data/data_manager/cache.py)

while timeseries is using util/net  (https://github.com/sunpy/sunpy/blob/main/sunpy/util/net.py)

nabobalis:
Where is map using the cache?

Sauravroy34:
~/.cache/sunpy this is the cache dir

Sauravroy34:
at lines 16  module is imported and its implemented at lines 243-248

https://github.com/Sauravroy34/sunpy/blob/dc6732598a7d13593a7d570978d998fb09fa265e/sunpy/map/map_factory.py#L16
```
from sunpy.data import cache
```

https://github.com/Sauravroy34/sunpy/blob/dc6732598a7d13593a7d570978d998fb09fa265e/sunpy/map/map_factory.py#L243-L248
 ``` @_parse_arg.register(Request)
    def _parse_url(self, arg, **kwargs):
        url = arg.full_url
        path = str(cache.download(url).absolute())
        pairs = self._read_file(path, **kwargs)
        return pairs
```
Sure we can replicate the same for timeseries as well

**Comment by Sauravroy34** [comment]
> Map is using datamanger s cache functionality (https://github.com/sunpy/sunpy/blob/main/sunpy/data/data_manager/cache.py)
> 
> while timeseries is using util/net (https://github.com/sunpy/sunpy/blob/main/sunpy/util/net.py)
Map factory used to have same download functionality as timeseries have now 
but it was changed after this issue https://github.com/sunpy/sunpy/issues/3144
with this pr 
https://github.com/sunpy/sunpy/pull/3124/files#diff-51c9932a08fa0ea1533e04449c7fb5a0135433e91d62037619d3985177169203

**Comment by nabobalis** [comment]
Right, might as well use the same mechanism as time series.

File: sunpy/util/net.py [hunk 1]
```diff
@@ -1,3 +1,4 @@
+
```
nabobalis, line 1:
```suggestion
```

sunpy/timeseries/timeseries_factory.py
```diff
@@ -407,7 +406,7 @@ def _parse_ts(self, ts, **kwargs):
 
     @_parse_arg.register(Request)
     def _parse_url(self, request, **kwargs):
-        path = download_file(request.full_url, get_and_create_download_dir())
+        path = cache.download(request.full_url).absolute()
```
nabobalis, line 409: Is there a test for this?

nabobalis:
I think the title of the PR and description needs updating. 

Also a changelog entry.

sunpy/timeseries/timeseries_factory.py
Sauravroy34: https://github.com/Sauravroy34/sunpy/blob/main/sunpy/timeseries/tests/test_timeseries_factory.py#L92-L99
in line 92-99 
```
 @pytest.mark.remote_data
def test_from_url():
    # This is the same PSP file we have in our test data, but accessed from a URL
    url = ('https://spdf.gsfc.nasa.gov/pub/data/psp/fields/l2/mag_rtn_1min/2020/'
           'psp_fld_l2_mag_rtn_1min_20200104_v02.cdf')
    ts = sunpy.timeseries.TimeSeries(url)
    assert isinstance(ts[0], sunpy.timeseries.GenericTimeSeries)
    assert isinstance(ts[1], sunpy.timeseries.GenericTimeSeries)
```

nabobalis:
So just to confirm with this change both of these points are fixed https://github.com/sunpy/sunpy/issues/2088#issuecomment-1207434122 ?

Sauravroy34:
> So just to confirm with this change both of these points are fixed [#2088 (comment)](https://github.com/sunpy/sunpy/issues/2088#issuecomment-1207434122) ?

Indeed

nabobalis (approved):

nabobalis:
Looks like a real test failure but it doesn't have anything to do with this PR. We should fix it anyway.

Sauravroy34:
Now 
CI / online / py312-online (ubuntu-latest) (pull_request)  are failing (which has nothing to do with this pr )
```
FAILED ../../.tox/py312-online/lib/python3.12/site-packages/sunpy/net/tests/test_scraper.py::test_ftp - IndexError: list index out of range
FAILED ../../.tox/py312-online/lib/python3.12/site-packages/sunpy/net/tests/test_scraper.py::test_ftp_new_format - IndexError: list index out of range
```

nabobalis:
Thanks again @Sauravroy34

Issue #7266, opened 2023-10-26
`sunpy.coordinates.transformations._times_are_equal()` should have greater tolerance for similar times


SolarDrew:
### Describe the bug

`sunpy.coordinates.transformations._times_are_equal()` appears to check if the times are exactly equal. Obviously this is going to be fine in many situations but I have a case where they are the same to within 1e-9. This is being caught as not equal and breaking a coordinate transformation. A tolerance should be introduced here to address this.

### To Reproduce

from astropy.coordinates import SkyCoord
import sunpy.coordinates
import astropy.units as u

from_coord = SkyCoord(-1000*u.arcsec, 300*u.arcsec,
                      frame="helioprojective", obstime="2023-01-01T00:00", observer="earth")

to_frame = SkyCoord(-1000*u.arcsec, 300*u.arcsec,
                    frame="helioprojective", obstime="2023-01-01T00:00:00.0000001").frame

print(from_coord.obstime)
print(to_frame.obstime)
print(from_coord.obstime - to_frame.obstime)

print(sunpy.coordinates.transformations._times_are_equal(from_coord.obstime, to_frame.obstime))

### System Details

- sunpy 5.0.1
- astropy 5.3.4
- Python 3.11.6
- Arch Linux, Linux 6.5.8-arch1-1

### Installation method

pip

ayshih:
I appreciate that the rigor is frustrating here, but I'm not a fan of making this change.  All of the loopback code – why `_times_are_equal()` even exists – was going to be refactored away; implementing a tolerance means that we can't do that refactoring.  It'd be challenging to implement a tolerance in the upstream loopback code because it's certainly possible that even a time difference as short as a nanosecond results in appreciably non-equivalent frames.

I think we'd have to approach this in the same vein as the previously discussed context manager for saying that frames are equivalent enough.

dstansby:
This is private API anyway - is there an example using only public API that motivates this request for a change?

Cadair:
yes, it's in the OP. It's used inside the coordinate transformations to determine if the observer shift needs to be accounted for.

Cadair:
Doing

```python
from_coord.transform_to(to_frame)
```
triggers an observer shift, is the effect on public API.

dstansby:
And why from a user facing perspective is that an issue? Performance?

ayshih:
It could also be annoying for off-disk coordinates without a screen assumption

Cadair:
That was the one that got us.

[...]

PR #8051, opened 2025-02-02
sunpy.coordinates.transformations._times_are_equal() should have greater tolerance for similar times #7266

parharti:
This PR addresses the issue in the time comparison logic for two Time objects in the TAI scale. Initially, comparing time_1.tai and time_2.tai resulted in incorrect results because .tai returns Time objects, not simple numeric values. To fix this, I switched to using time_1.tai.unix and time_2.tai.unix, which return Unix time as numeric values in seconds, ensuring more accurate and reliable comparisons.

I opted for the absolute difference approach (abs(tai_1 - tai_2) <= atol) instead of using np.allclose() because it is simpler, more transparent, and better suited for directly comparing time differences with a specific tolerance. Although np.allclose() was tested, it gave incorrect results due to the complexity introduced by Quantity objects in astropy.time. The absolute difference method guarantees precision while maintaining clarity.

Thanks to all the contributors for the insightful conversations that helped in this solution!

### DIFF
--- a/sunpy/coordinates/_transformations.py
+++ b/sunpy/coordinates/_transformations.py
@@ -315,14 +315,19 @@ def _check_observer_defined(frame):
                                "but this is valid for only HeliographicCarrington frames.")
 
 
-def _times_are_equal(time_1, time_2):
-    # Checks whether times are equal
-    if isinstance(time_1, Time) and isinstance(time_2, Time):
-        # We explicitly perform the check in TAI to avoid possible numerical precision differences
-        # between a time in UTC and the same time after a UTC->TAI->UTC conversion
-        return np.all(time_1.tai == time_2.tai)
+def time_comparator(time_1, time_2, atol=1e-9):
+    # Convert the TAI times to Unix time (which is based on UTC)
+    tai_1 = time_1.tai.unix
+    tai_2 = time_2.tai.unix
+    # Check if the difference between the two times is within the tolerance
+    return abs(tai_1 - tai_2) <= atol
 
-    # We also deem the times equal if they are both None
+
+def _times_are_equal(time_1, time_2):
+    # If both inputs are valid Time objects, compare them
+    if isinstance(time_1, Time) and isinstance(time_2, Time):
+        return time_comparator(time_1, time_2)
+    # If both times are None, consider them equal
     return time_1 is None and time_2 is None
 
 


nabobalis:
Hello @parharti, thank you for the pull request but I have a few questions.

What bug or problem are you trying to fix? Do you have a code example, before and after showing the problem and that it is fixed?

In addition, the text in your message looks like the output of a LLM model, are you using one to generate text and did you also use it to generate the code changes? I ask since we have a policy on the use of GenAI: https://github.com/sunpy/sunpy#usage-of-generative-ai and we for GSoC are not going to accept students who are unable to be authentic with us.

parharti:
Hey @nabobalis, I understand your concern, but I can assure you that this code was written with full awareness and is not just copied and pasted from an LLM model.

The problem I attempted to solve is an open issue mentioned in this GitHub issue (https://github.com/sunpy/sunpy/issues/7266#issue-1963562481).

I created a separate function, as referenced in the conversation, and implemented a callback function that introduces a tolerance before comparing two times. The previous function treated times as different even when they only varied by a few nanoseconds

nabobalis:
> Hey @nabobalis, I understand your concern, but I can assure you that this code was written with full awareness and is not just copied and pasted from an LLM model.

Thanks but I strongly recommend to limit the use of an LLM, at least when talking to us and writing messages on GitHub.

> The problem I attempted to solve is an open issue mentioned in this GitHub issue ([#7266 (comment)](https://github.com/sunpy/sunpy/issues/7266#issue-1963562481)).
> 
> I created a separate function, as referenced in the conversation, and implemented a callback function that introduces a tolerance before comparing two times. The previous function treated times as different even when they only varied by a few nanoseconds

Can you show that it fixes the problem in the issue? My interpretation of https://github.com/sunpy/sunpy/issues/7266#issuecomment-1782879219 is that this is not as simple as adding a tolerance keyword to "_times_are_equal".

parharti:
time_1 = Time("2025-02-02T00:00:00")
time_2 = Time("2025-02-02T00:00:00.0000000005")
before this two times considered as different but now due to tolerance of nano second they will be considered as same time

parharti:
I would be very grateful if you could review it tell me the edits I should do to make PR successful

nabobalis:
> time_1 = Time("2025-02-02T00:00:00") time_2 = Time("2025-02-02T00:00:00.0000000005") before this two times considered as different but now due to tolerance of nano second they will be considered as same time

I understand that is the core of the original issue but the problem is that my reading of the original issue is that this is not a valid fix. We are introducing a tolerance which can not be changed by a user and needs to be fixed in astropy.

> I would be very grateful if you could review it tell me the edits I should do to make PR successful

Unfortunately, I am not qualified to review this PR, I will alter the person who is.

ayshih:
Apologies, but I am closing this PR.  The approach initially suggested in #7266 is untenable because of the incompatibility with intended refactoring of the coordinate-transformation code.  As @nabobalis indicated, the way forward seemed to be that we would need changes upstream (in `astropy`), and the changes would be complex ones, not simple ones.

Issue #7802, opened 2024-09-12
Add example of time-distance slice plot of sequence of images


hayesla:
### Describe the feature

We should add an example of using [sunpy.map.pixelate_coord_path()](https://docs.sunpy.org/en/stable/generated/api/sunpy.map.pixelate_coord_path.html#sunpy.map.pixelate_coord_path) and [sunpy.map.sample_at_coords()](https://docs.sunpy.org/en/stable/generated/api/sunpy.map.sample_at_coords.html#sunpy.map.sample_at_coords) on a sequence of images to extract a time-distance slice to identify features in solar atmosphere

i.e. 
![oscillation_example](https://github.com/user-attachments/assets/6a97ec77-7730-4735-929d-8a11baaaca0c)

### Proposed solution

The only thing is choosing sample data to do this and getting it on the gallery (i.e. having a longer sequence of images) - anyone have ideas?

Prtm2110:
Hello @hayesla, I tried using `propagate_with_solar_surface` to align a portion of the image and generate aligned slits along the x-axis of a graph, as you have done in your examples. Could you let me know if my approach is acceptable or if you have a better method for this?  

Additionally, I'm facing an issue when using `plot_coord(intensity_coords)` on the aligned images. Here’s a link to my notebook for reference:  [Split Gallery Example](https://github.com/Prtm2110/Split-Gallery-Example/blob/master/split_gallery_ex.ipynb)

nabobalis:
What is the issue you are having with `plot_coord`?

Prtm2110:
I think I fixed it, Could you please check the [Split Gallery Example](https://github.com/Prtm2110/Split-Gallery-Example/blob/master/split_gallery_ex.ipynb)? I want to confirm if the use case and flow of the gallery example are correct before I start working on it.

nabobalis:
Right now its code? I mean it looks ok but the output is different to what is in the original post.

samaloney:
So I was curious about his myself so made this

<img width="912" height="676" alt="Image" src="https://github.com/user-attachments/assets/193e7a4f-6157-4c90-b9f8-d6f3e6768f14" />

Issue #7173, opened 2023-08-25
Fix `assume_spherical_screen()`+`propagate_with_solar_surface()`


ayshih:
The `propagate_with_solar_surface()` context manager can be used to account for solar differential rotation when reprojecting a map to a different observation time, but off-disk data are discarded.  The `assume_spherical_screen()` context manager is typically used to preserve off-disk data in reprojections (e.g., for a change in observer location).  However, using these these two context managers together does not work as one would hope, and most of the off-disk data are still discarded.

This is a very hard problem to solve.  If I could label it `Effort Extreme`, I would.

PR #8057, opened 2025-02-04
Adds example of time-distance slice plot of sequence of images

Prtm2110:
## PR Description
Fixes: #7802

There is a lot to be done here, but this is the initial work. An initial review would be great, especially about the flow of the gallery example.

TODO:
- [x] Changelog

Preview: https://sunpy--8057.org.readthedocs.build/en/8057/generated/gallery/plotting/time_distance.html


nabobalis:
I guess my first comment is that the final figure is not a time-distance diagram.

nabobalis:
Just to clarify further, I think the outcome of the last figure should look like the top figure of Figure 4 in this paper: https://iopscience.iop.org/article/10.1088/0004-637X/697/2/1384/pdf

examples/plotting/time_distance_slice_plot.py
nabobalis, line 1: The name should drop slice I think.

examples/plotting/time_distance_slice_plot.py
```diff
@@ -0,0 +1,104 @@
+"""
+============================================
+Time-Distance Slice Plot of Sequence of Maps
+============================================
```
nabobalis, line 4:
```suggestion
====================================
Time distance for a sequence of maps
====================================
```

examples/plotting/time_distance_slice_plot.py
```diff
@@ -0,0 +1,104 @@
...
+
+This example showcases how you can use :func:`sunpy.map.pixelate_coord_path` and
+:func:`sunpy.map.sample_at_coords()` on a sequence of images to extract a time-distance
+slice and identify features in the solar atmosphere.
```
nabobalis, line 8:
```suggestion
This example showcases how you can use :func:`sunpy.map.pixelate_coord_path` and
:func:`sunpy.map.sample_at_coords()` on a sequence of images to create a time-distance diagram.
```

examples/plotting/time_distance_slice_plot.py
```diff
@@ -0,0 +1,104 @@
...
+from sunpy.net import attrs as a
+
+###############################################################################
+# First, we will import a sequence of images from AIA 171 angstrom separated by 1 hour
+# and create a sequence of maps.
+
+query = Fido.search(a.Time('2018-05-30 02:00:00', '2018-05-30 10:00:00'),
+                    a.Instrument.aia,
+                    a.Wavelength(171*u.angstrom),
+                    a.Sample(1*u.hr))
```
nabobalis, line 28: How many files is this?

examples/plotting/time_distance_slice_plot.py
```diff
@@ -0,0 +1,104 @@
...
+from astropy.visualization import ImageNormalize, SqrtStretch
+
+import sunpy.map
+from sunpy.coordinates import propagate_with_solar_surface
+from sunpy.net import Fido
+from sunpy.net import attrs as a
+
+###############################################################################
+# First, we will import a sequence of images from AIA 171 angstrom separated by 1 hour
+# and create a sequence of maps.
```
nabobalis, line 23:
```suggestion
# First, we will need to acquire a sequence of images from SDO/AIA.
```

examples/plotting/time_distance_slice_plot.py
```diff
@@ -0,0 +1,104 @@
...
+
+query = Fido.search(a.Time('2018-05-30 02:00:00', '2018-05-30 10:00:00'),
+                    a.Instrument.aia,
+                    a.Wavelength(171*u.angstrom),
+                    a.Sample(1*u.hr))
+print(query)
```
nabobalis, line 29:
```suggestion
```

examples/plotting/time_distance_slice_plot.py
```diff
@@ -0,0 +1,104 @@
...
+files = Fido.fetch(query)
+aia_sequence = sunpy.map.Map(files, sequence=True)
```
nabobalis, line 31: You should plot the first one in sequence and plot the object you want to slice on top of it.

examples/plotting/time_distance_slice_plot.py
```diff
@@ -0,0 +1,104 @@
...
+
+fig = plt.figure(figsize=(10, 4))
+ax1 = fig.add_subplot(121, projection=aia_sequence[0])
+aia_sequence[0].plot(axes=ax1)
+ax1.plot_coord(intensity_coords)
+
+ax2 = fig.add_subplot(122)
+ax2.plot(angular_separation, intensity)
+ax2.set_xlabel("Angular distance along slit [arcsec]")
+ax2.set_ylabel(f"Intensity [{aia_sequence[0].unit}]")
```
nabobalis, line 55: This section should talk about where you want to place the slit and why, create it and overplot it on the first image of the sequence and not extract any intensity.

examples/plotting/time_distance_slice_plot.py
```diff
@@ -0,0 +1,104 @@
...
+
+###############################################################################
+# Let's now create a rectangular slice (i.e., a map cutout).
+
+corner = SkyCoord(Tx=-250*u.arcsec, Ty=0*u.arcsec, frame=aia_sequence[0].coordinate_frame)
+width = 250*u.arcsec
+cutout_map = aia_sequence[0].submap(corner, width=width, height=500*u.arcsec)
```
nabobalis, line 62: Why are you cutting out the map in the first sequence?

examples/plotting/time_distance_slice_plot.py
```diff
@@ -0,0 +1,104 @@
...
+
+##################################################################################################
+# Now that we have our cutout, we can reproject each map in our sequence to
+# the WCS of that cutout using the `~sunpy.coordinates.propagate_with_solar_surface`
+# context manager to adjust the field of view of the cutout with the rotation of the solar surface.
+
+with propagate_with_solar_surface():
+    aia_sequence_aligned = sunpy.map.Map([m.reproject_to(cutout_map.wcs) for m in aia_sequence], sequence=True)
```
nabobalis, line 70: What is the purpose of this?

examples/plotting/time_distance_slice_plot.py
```diff
@@ -0,0 +1,104 @@
...
+
+##########################################################################################################
+# Now we can plot the time-distance slice of the sequence of maps using :func:`~matplotlib.pyplot.imshow()`.
```
nabobalis, line 73: So I don't see any place where you create the time-distance image, you need to get the data from the sequence first.

Prtm2110:
> I guess my first comment is that the final figure is not a time-distance diagram.

I interpreted it wrongly, end up aligning maps (which is exactly what we don't want here), Should it look something like this (somewhat at the centre of the sun), I tried to setup the slit outside the limb (at suns corona) for more details in the plot but it gave 
```
ValueError: array is too big; `arr.size * arr.dtype.itemsize` is larger than the maximum possible size.
```
image

nabobalis:
Yeah I dont know what that error is without some more context and code. 

We can have the slit be on disk and follow the active region.

examples/plotting/time_distance_image.py
```diff
@@ -0,0 +1,92 @@
...
+from sunpy.coordinates import propagate_with_solar_surface
+from sunpy.coordinates.screens import SphericalScreen
+from sunpy.net import Fido
+from sunpy.net import attrs as a
+
+###############################################################################
+# First, we will need to acquire a sequence of images from SDO/AIA.
+
+query = Fido.search(
+    a.Time('2025-05-01 00:00:00', '2025-05-01 00:10:00'),
```
nabobalis, line 25: This pulls down 50 images. Maybe too much.

examples/plotting/time_distance_image.py
```diff
@@ -0,0 +1,92 @@
...
+with propagate_with_solar_surface():
+    with SphericalScreen(cut_aia_sequence[0].observer_coordinate):
+        aia_sequence_aligned = sunpy.map.Map([m.reproject_to(cut_aia_sequence[0].wcs) for m in cut_aia_sequence], sequence=True)
+
+###############################################################################
+# Next, we will define a path in space and then use that to fetch the pixel
+# coordinates for every pixel that intersects with the physical path.
+
+# Note this is [X0, X1] and [Y0, Y1] in the coordinate frame of the map.
+line_coords = SkyCoord([900, 1000],[456, 250] , unit=u.arcsec, frame=aia_sequence_aligned[0].coordinate_frame)
```
nabobalis, line 57: This does not return anything interesting.

examples/plotting/time_distance_image.py
```diff
@@ -0,0 +1,92 @@
...
+        cut_sequence.append(aia_map.submap(corner, width=400*u.arcsec, height=400*u.arcsec))
+cut_aia_sequence = sunpy.map.Map(cut_sequence, sequence=True)
+
+###############################################################################
+# Now that we have our cutouts, we can reproject each map in our sequence to
+# the WCS of that cutout using the `~sunpy.coordinates.propagate_with_solar_surface`
+# context manager to adjust the field of view of the cutout with the rotation of the solar surface.
+
+with propagate_with_solar_surface():
+    with SphericalScreen(cut_aia_sequence[0].observer_coordinate):
```
nabobalis, line 49: Is this correct?

ayshih: Nope, you generally can't combine the two context managers (xref #7173)

Cadair: This feature got added, so I think we're ok.

examples/plotting/time_distance_image.py
```diff
@@ -0,0 +1,92 @@
...
+        aia_sequence_aligned = sunpy.map.Map([m.reproject_to(cut_aia_sequence[0].wcs) for m in cut_aia_sequence], sequence=True)
```
Cadair: Do we actually need to reproject the arrays for this example? Can we not compute the SkyCoords for the rotated & screened coordinates for each map and then extract the pixels?

examples/plotting/time_distance_image.py
Cadair:
```suggestion
line_coords = SkyCoord([900, 1000], [456, 250], unit=u.arcsec, frame=aia_sequence_aligned[0].coordinate_frame)
```

examples/plotting/time_distance_image.py
```diff
@@ -0,0 +1,92 @@
+"""
+=====================================================
+Creating a time-distance plot from a sequence of maps
+=====================================================
+
+This example showcases how you can use :func:`sunpy.map.pixelate_coord_path` and
+:func:`sunpy.map.sample_at_coords()` on a sequence of images to create a time-distance diagram.
```
Cadair: Shouldn't this be:

```suggestion
:func:`sunpy.map.sample_at_coords` on a sequence of images to create a time-distance diagram.
```

examples/plotting/time_distance_image.py
```diff
@@ -0,0 +1,92 @@
...
+from sunpy.net import Fido
+from sunpy.net import attrs as a
+
+###############################################################################
+# First, we will need to acquire a sequence of images from SDO/AIA.
+
+query = Fido.search(
+    a.Time('2025-05-01 00:00:00', '2025-05-01 00:10:00'),
+    a.Instrument.aia,
+    a.Wavelength(171*u.angstrom)
```
Cadair:
```suggestion
    a.Wavelength(171*u.angstrom),
    a.Sample(1*u.min),
```

examples/plotting/time_distance_image.py
nabobalis: How does someone do that?

nabobalis:
I rebased, updated the example per Stuart's comments and renamed the file.

examples/plotting/time_distance.py
```diff
@@ -0,0 +1,95 @@
...
+from sunpy.net import attrs as a
+
+###############################################################################
+# First, we will need to acquire a sequence of images from SDO/AIA.
+
+query = Fido.search(
+    a.Time('2025-05-01 00:00:00', '2025-05-01 00:10:00'),
+    a.Instrument.aia,
+    a.Wavelength(171*u.angstrom),
+    a.Sample(1*u.min),
```
samaloney: I'd suggest using the same date as in the issue but possibly for shorter period of time at high cadence to keep the number of files down

examples/plotting/time_distance.py
```diff
@@ -0,0 +1,95 @@
...
+###############################################################################
+# Now that we have our cutouts, we can reproject each map in our sequence to
+# the WCS of that cutout using the `~sunpy.coordinates.propagate_with_solar_surface`
+# context manager to adjust the field of view of the cutout with the rotation of the solar surface.
+
+with propagate_with_solar_surface():
+    with SphericalScreen(cut_aia_sequence[0].observer_coordinate):
+        aia_sequence_aligned = sunpy.map.Map([m.reproject_to(cut_aia_sequence[0].wcs) for m in cut_aia_sequence], sequence=True)
+
+aia_sequence_aligned.plot()
```
samaloney: Few things I'd update to use similar coordinates as in the issue so will actually see something.

I think these two block can be combined into one e.g.

```
reprojected_sub_maps = []
for cur_map in aia_seq:
    cur_map = cur_map/cur_map.exposure_time
    with propagate_with_solar_surface(), SphericalScreen(cur_map.observer_coordinate, only_off_disk=True):
        reprojected_sub_maps.append(cur_map.reproject_to(ref_sub_map.wcs))
sub_map_seq = Map(reprojected_sub_maps, sequence=True)
```

Might be worth noting the the order of the context mangers is important AFAIK

examples/plotting/time_distance.py
```diff
@@ -0,0 +1,95 @@
...
+
+###############################################################################
+# Next, we will define a path in space and then use that to fetch the pixel
+# coordinates for every pixel that intersects with the physical path.
+
+# Note this is [X0, X1] and [Y0, Y1] in the coordinate frame of the map.
+line_coords = SkyCoord([850, 950], [456, 250], unit=u.arcsec, frame=aia_sequence_aligned[0].coordinate_frame)
```
samaloney: Update to roughly match issue

examples/plotting/time_distance.py
```diff
@@ -0,0 +1,95 @@
...
+aia_sequence_aligned[0].plot(axes=ax1)
+ax1.plot_coord(line_coords)
+
+###############################################################################
+# Now let us construct the time-distance diagram. First we need to get the data
+# over the path we defined above for the entire sequence.
+
+intensities = []
+for aia_map in aia_sequence_aligned:
+    intensity_coords = sunpy.map.pixelate_coord_path(aia_map, line_coords)
```
samaloney: Since the maps have been reprojected only need to do this once outside the for loop.

examples/plotting/time_distance.py
```diff
@@ -0,0 +1,95 @@
...
+# This defines the distance along the path in pixels.
+angular_separation = intensity_coords.separation(intensity_coords[0]).to(u.arcsec)
+
+###############################################################################
+# We can now plot the time-distance diagram.
+
+fig = plt.figure(figsize=(10, 4))
+ax1 = fig.add_subplot(111)
+ax1.imshow(intensities, aspect='auto', origin='lower', extent=[0, len(aia_sequence_aligned) * 12, 0, angular_separation[-1].value])
+ax1.set_xlabel('Time (s)')
```
samaloney:
```suggestion
ax1.imshow(intensities.T, aspect='auto', origin='lower', 
           extent=[aia_seq[0].date.datetime,  aia_seq[-1].date.datetime, 
           0, angular_separation[-1].value])
ax1.set_xlabel('Time (UTC)')
```

samaloney (commented):
Nice work @Prtm2110 I think this is nearly there. 

So there at least two-ways to achieve this:
1. by reprojection (as it currently is)
2. by transforming the extracted line coordinates from the first map to the subsequent maps
I guess both should give similar answers but I assume reprojection will be slower, should both be shown / compared here or another example.

Also since this needs to download data might as well use the dates, locations from the shows a nice oscillation.

examples/plotting/time_distance_image.py
samaloney: So you get slightly different results, a few percent, depending on the method. This is for the dates, location from the original issue.

image

samaloney:
@Prtm2110 thanks for the PR do you want to / will you have the time to address the review comments no pressure either way? If not would you mind if I updated the PR to address these?

prathamhole14:
Hi @samaloney sorry for the delay in responding and thank you for the suggestions! I tried producing the same output but unfortunately I am not able to so please if you would like you can totally take over this PR, thank you : )

samaloney:
This is the current final figure

image

But it's using ~120 files and so is quite slow to run and downloads a decent amount of data. ~~I could truncate~~ I have to do something as the doc build is failing due to excessive memory consumption. Could reduce the time range to only show one "period" of the oscillation say 18:14 - 18:19 and, or reduce the cadence but it doesn't look as nice even at 30s cadence.

examples/plotting/time_distance.py
```diff
@@ -0,0 +1,173 @@
...
+    a.Instrument.aia,
+    a.Wavelength(171*u.angstrom),
+)
+files = Fido.fetch(query, site="NSO")
+files = sorted(files)
+
+
+###############################################################################
+# Our target will be a set of loops in the corona above an active region.
+# First load the fits files we downloaded above in to map sequence and create a
```
nabobalis:
```suggestion
# First load the FITS files we downloaded above in to map sequence and create a
```

nabobalis (approved):

nabobalis:
Thank you @Prtm2110 and @samaloney

Cadair:
To know how close the two approaches are, that last figure really needed some colourbars

PR #6392, opened 2022-08-30
Cleaned up existing ASV tests and added new ones

ayshih:
I haven't managed to get ASV working locally.  Can someone give this a try?

PR #8060, opened 2025-02-06
Added an ASV benchmarks CI build for PRs

ayshih:
We have some benchmarks using `asv` (cf. #6392) that we haven't really been taking advantage of.  This PR enables the checking of any performance differences for a given PR through a dedicated `benchmarks` CI build.  This CI build is turned on by adding the label "Run benchmarks" to that given PR.~, and will run after the `core` CI build passes.~

This PR also adds a bunch more ASV benchmarks:
- Coordinate transformations between frames that had been added in the past two years
- Reprojection
- Reprojection plus differential rotation
- Plotting maps with the various settings of `autoalign`

---
Only relevant because of how `asv` builds its working environment, this PR also bumps the minimum required version of `setuptools_scm` due to a bad bug with 8.0.0 (that release is even yanked).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,6 +126,40 @@ jobs:
         - linux: linkcheck
           pytest: false
 
+  benchmarks:
+    needs: [core]
+    if: |
+      github.event_name == 'pull_request' &&
+      contains(github.event.pull_request.labels.*.name, 'Run benchmarks')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sunpy repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - name: Install dependencies
+        run: |
+          sudo apt-get install -y libopencv-dev
+          python -m pip install --upgrade pip
+          pip install asv virtualenv
+      - name: Add machine info for ASV
+        run: asv machine --yes
+      - name: Run benchmarks
+        env:
+          OPENBLAS_NUM_THREADS: 1
+          MKL_NUM_THREADS: 1
+          OMP_NUM_THREADS: 1
+          ASV_FACTOR: 1.1
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          COLUMNS: 150  # set a large terminal width so that large tables are rendered in 2D
+        run: |
+          ASV_OPTIONS="--split --show-stderr --factor $ASV_FACTOR --cpu-affinity 0"
+          asv continuous $ASV_OPTIONS ${BASE_SHA} ${GITHUB_SHA}
+
   sdist_verify:
     if: |
       github.event_name == 'workflow_dispatch' || (
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,40 +126,6 @@ jobs:
         - linux: linkcheck
           pytest: false
 
-  benchmarks:
-    needs: [core]
-    if: |
-      github.event_name == 'pull_request' &&
-      contains(github.event.pull_request.labels.*.name, 'Run benchmarks')
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout sunpy repo
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-      - name: Install Python
-        uses: actions/setup-python@v5
-        with:
-          python-version: "3.11"
-      - name: Install dependencies
-        run: |
-          sudo apt-get install -y libopencv-dev
-          python -m pip install --upgrade pip
-          pip install asv virtualenv
-      - name: Add machine info for ASV
-        run: asv machine --yes
-      - name: Run benchmarks
-        env:
-          OPENBLAS_NUM_THREADS: 1
-          MKL_NUM_THREADS: 1
-          OMP_NUM_THREADS: 1
-          ASV_FACTOR: 1.1
-          BASE_SHA: ${{ github.event.pull_request.base.sha }}
-          COLUMNS: 150  # set a large terminal width so that large tables are rendered in 2D
-        run: |
-          ASV_OPTIONS="--split --show-stderr --factor $ASV_FACTOR --cpu-affinity 0"
-          asv continuous $ASV_OPTIONS ${BASE_SHA} ${GITHUB_SHA}
-
   sdist_verify:
     if: |
       github.event_name == 'workflow_dispatch' || (
--- a/.github/workflows/ci_benchmarks.yml
+++ b/.github/workflows/ci_benchmarks.yml
@@ -1,0 +1,43 @@
+name: CI benchmarks
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled]
+
+concurrency:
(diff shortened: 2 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -126,6 +126,39 @@ jobs:
         - linux: linkcheck
           pytest: false
 
+  benchmarks:
```
Cadair, line 129: I think it would be nicer if this was it's own workflow file?

ayshih: It doesn't have to be this way, but the desire is to have the benchmarks run only after the `core` build actually succeeds

ayshih:
The benchmarks CI build takes about 10 minutes.  Here are example timing tests for the newly added benchmarks:
```
[52.63%] ··· ====== ============= ============= ============= ============= ============= ============= =============
             --                                                    dest                                              
             ------ -------------------------------------------------------------------------------------------------
              src        HCRS          HGS           HGC           HCC           HPC           HPR           HCI     
             ====== ============= ============= ============= ============= ============= ============= =============
              HCRS       n/a       1.35±0.02ms   9.91±0.07ms    2.28±0.1ms   4.10±0.02ms   5.48±0.08ms   4.72±0.05ms 
              HGS    1.49±0.01ms       n/a       8.43±0.05ms     803±4μs     2.49±0.01ms   3.94±0.05ms   3.33±0.02ms 
              HGC    9.76±0.06ms   8.19±0.05ms       n/a       9.06±0.01ms   10.9±0.06ms   12.3±0.08ms   11.5±0.04ms 
              HCC    2.32±0.03ms     721±5μs     9.17±0.02ms       n/a       1.78±0.01ms   3.21±0.03ms   4.09±0.02ms 
              HPC    3.81±0.01ms   2.15±0.01ms   10.7±0.03ms   1.29±0.01ms       n/a       1.40±0.01ms   5.70±0.04ms 
              HPR    5.04±0.05ms   3.34±0.03ms   11.9±0.02ms   2.49±0.02ms   1.12±0.02ms       n/a        6.80±0.1ms 
              HCI    4.85±0.02ms   3.31±0.03ms   11.9±0.05ms   4.21±0.05ms   6.05±0.03ms    7.43±0.1ms       n/a     
             ====== ============= ============= ============= ============= ============= ============= =============
```

```
[53.95%] ··· ===== ============= ============= ============= ============
             --                             dest                         
             ----- ------------------------------------------------------
              src       GEO           MAG            SM          GSM     
             ===== ============= ============= ============= ============
              GEO       n/a         685±3μs     6.26±0.07ms   16.8±0.2ms 
              MAG     839±1μs         n/a       5.45±0.06ms   15.9±0.2ms 
               SM   6.49±0.07ms   5.67±0.05ms       n/a       10.6±0.1ms 
              GSM    17.3±0.2ms    16.3±0.2ms   10.9±0.08ms      n/a     
             ===== ============= ============= ============= ============
```

```
[56.58%] ··· ============ =========
              autoalign            
             ------------ ---------
                False      291±1ms 
                 True      1.50±0s 
              pcolormesh   1.50±0s 
             ============ =========
```

```
[68.42%] ··· map.Reproject.time_reproject_to                                                                                                       ok
[68.42%] ··· =============== ============
                algorithm                
             --------------- ------------
              interpolation    1.18±0s   
                 adaptive     2.20±0.01s 
             =============== ============

[69.74%] ··· map.Reproject.time_reproject_to_plus_diffrot                                                                                          ok
[69.74%] ··· =============== ============
                algorithm                
             --------------- ------------
              interpolation   1.43±0.01s 
                 adaptive     2.44±0.01s 
             =============== ============
```

.github/workflows/ci.yml
Cadair: So one other major advantage to moving this to it's own workflow is that we can have it react to the addition of the label. Currently the CI workflow isn't triggered by label add (because it would restart all the CI), but if we make a new workflow we can add that trigger. In my opinion this outweighs the benefit of early-exit if core fails.

Cadair (approved):
This is obviously a straight up improvement, so I am not going to block it's merge, we can improve it in follow up PRs. Things I would like to see done:

* Move it to it's own workflow (probably)
* Document it here: https://github.com/sunpy/sunpy/blob/main/docs/dev_guide/contents/ci_jobs.rst
* Have the job add a report as a comment to the PR, especially if it fails? Is it possible to get ASV to write a summary out to a text file?

nabobalis:
Do we not want to bring back the benchmark results repo?

Cadair:
Yes, we can do that, but it's actually orthogonal to this which just runs a diff on each PR. I.e. they can both exist independent of each other.

nabobalis:
> Yes, we can do that, but it's actually orthogonal to this which just runs a diff on each PR. I.e. they can both exist independent of each other.

Does a PR benchmark have a baseline so we know if its worse after the patch?

Cadair:
I believe this works by running both before and after every build and comparing.

nabobalis:
Does it create a html page we can display with giles?

Cadair:
No, just build output. Would be nice to have it post as a comment or at least build output.

ayshih:
> * Move it to it's own workflow (probably)

I'll respond in the other thread

> * Document it here: https://github.com/sunpy/sunpy/blob/main/docs/dev_guide/contents/ci_jobs.rst

Ha, I didn't even know that we documented our builds. =)

> * Have the job add a report as a comment to the PR, especially if it fails? Is it possible to get ASV to write a summary out to a text file?

I imagine it could be done; certainly I've seen people do that for pytest-benchmark output.  There probably exists a well-developed action out there.

> Do we not want to bring back the benchmark results repo?

As @Cadair benefits, there are orthogonal benefits for that repo.  I'll caution that the many benchmarks added in this PR means that the total computation time for a single commit has increased by at least a half an order of magnitude.  I think the other repo took many hours to run through the defined range of commits; we will probably be measuring in days if we don't shrink or sparsify the commits.

> Does a PR benchmark have a baseline so we know if its worse after the patch?

Yes, this passes ASV the current head commit and a temporary commit with the PR merged in, and flags any benchmark that has changed by more than 10% (configurable).

.github/workflows/ci.yml
ayshih: Perhaps it can be a reusable workflow that `cron` calls when it completes successfully?  I'm still a novice with GitHub actions/workflows

nabobalis:
> As @Cadair stated, there are orthogonal benefits for that repo. I'll caution that the many benchmarks added in this PR means that the total computation time for a single commit has increased by at least a half an order of magnitude. I think the other repo took many hours to run through the defined range of commits, and that was two years ago. We will probably be measuring the run time in days if we don't shrink or sparsify the commits.

Yeah that would easily timeout on actions if I recall if we were not careful. Hopefully if I ever get the numfocus to get some machines for benchmarking, we can run it on a dedicated machine in the far future.

ayshih:
> * Move it to it's own workflow (probably)

Done, and now triggers when the label is added

> * Document it here: https://github.com/sunpy/sunpy/blob/main/docs/dev_guide/contents/ci_jobs.rst

Done

> * Have the job add a report as a comment to the PR, especially if it fails? Is it possible to get ASV to write a summary out to a text file?

I tried to implement this, but I'm clearly not understanding something.  Let's defer it to a future PR.

Cadair (approved):

PR #8061, opened 2025-02-07
Add Github-Action for new release when tags are pushed to the repository

### LINKED ISSUE #6134: Improve generation of changelog for GitHub release

ankitkhushwaha:
## PR Description

fixes #6134

this pr adds a Github-Action that publish a new release when new tag with `v` tag is pushed to repo.
body of release is updated to text returned by `tools/generate_github_changelog.py`

example of [Published release](https://github.com/ankitkhushwaha/sunpy/releases/tag/v6.0.1.3)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -218,6 +218,37 @@ jobs:
     secrets:
       pypi_token: ${{ secrets.pypi_token }}
 
+  release:
+    if: startsWith(github.ref, 'refs/tags/v')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      actions: none
+      checks: none
+      deployments: none
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Install Pandoc
+        run: sudo apt-get update && sudo apt-get install -y pandoc
+      - name: Generate GitHub Release Notes
+        run: |
+          pip install bleach
+          pandoc -- wrap=none -t markdown_strict CHANGELOG.rst > notes.md
+          python3 -c "import bleach, sys; print(bleach.clean(sys.stdin.read()))" < notes.md > release_notes.md
+          { echo "RELEASE_BODY<<EOF"; cat release_notes.md; echo "EOF"; } >> $GITHUB_ENV
+
+      - name: Create GitHub Release
+        uses: "actions/create-release@v1.1.4"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: ${{ github.ref_name }}
+          body: ${{ env.RELEASE_BODY }}
+          draft: false
+          prerelease: false
+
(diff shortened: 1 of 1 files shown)


nabobalis:
Thanks for the PR @ankitkhushwaha.

As it stands, this does not really integrate with the release workflow and I think that needs thinking about before we continue with this PR.

ankitkhushwaha:
> As it stands, this does not really integrate with the release workflow and I think that needs thinking about before we continue with this PR.

Are there specific steps or dependencies to consider? I'm happy to adjust the implementation accordingly.

nabobalis:
> > As it stands, this does not really integrate with the release workflow and I think that needs thinking about before we continue with this PR.
> 
> Are there specific steps or dependencies to consider? I'm happy to adjust the implementation accordingly.

How does this integrate with the openastro templates which are used for the release?
How do we handle the security of having an action that modifies the repo and prevent bad actors?
Should this into the package template repo instead since this should be rolled out to every sunpy package?

ankitkhushwaha:
i have extended the ci.yml instead of creating a new action.

.github/workflows/ci.yml
```diff
@@ -218,6 +218,37 @@ jobs:
...
       pypi_token: ${{ secrets.pypi_token }}
 
+  release:
+    if: startsWith(github.ref, 'refs/tags/v')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      actions: none
+      checks: none
+      deployments: none
```
ankitkhushwaha, line 228: now action have only permission to write the release.

.github/workflows/ci.yml
```diff
@@ -218,6 +218,37 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Install Pandoc
+        run: sudo apt-get update && sudo apt-get install -y pandoc
+      - name: Generate GitHub Release Notes
+        run: |
+          pip install bleach
+          python tools/generate_github_changelog.py > notes.md
+          python3 -c "import bleach, sys; print(bleach.clean(sys.stdin.read()))" < notes.md > release_notes.md
+          { echo "RELEASE_BODY<<EOF"; cat release_notes.md; echo "EOF"; } >> $GITHUB_ENV
```
ankitkhushwaha, line 239: this way it will prevent the bad actors

ankitkhushwaha:
> Should this into the package template repo instead since this should be rolled out to every sunpy package?

see [#181](https://github.com/sunpy/package-template/pull/181)

ankitkhushwaha:
closing it 
https://github.com/sunpy/package-template/pull/181

PR #8062, opened 2025-02-07
Add line breaks to long signatures in API docs

dstansby:
I remember thinking for *ages* that signatures like [make_fitswcs_header](https://docs.sunpy.org/en/stable/generated/api/sunpy.map.header_helper.make_fitswcs_header.html#sunpy.map.header_helper.make_fitswcs_header) were really hard to read. Now, years later, I return with a solution! 🚀

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -105,6 +105,8 @@ linkcheck_anchors = False
 
 # -- General configuration ---------------------------------------------------
 
+# Wrap large function/method signatures
+maximum_signature_line_length = 80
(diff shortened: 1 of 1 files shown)


Cadair:
Omg yes, this is amazing. Don't suppose you also.want to pr this to the package template?

nabobalis (approved):

nabobalis:
Will this work https://github.com/sunpy/package-template/pull/180 ?

PR #8072, opened 2025-02-10
Add a post to the PR summarizing its CI benchmark results

ayshih:
This is a follow-on to #8060, which added the capability to compare before/after benchmarks for a PR.  This PR excerpts the summary from the benchmarks output and adds a post to the PR.  (On subsequent CI runs for the same PR, it will update that same post instead of adding a new one.)  Because of the permissions necessary to add a post to a PR, this PR needs be merged into `main` before it can be confirmed to be working properly.

### DIFF
--- a/.github/workflows/ci_benchmarks.yml
+++ b/.github/workflows/ci_benchmarks.yml
@@ -1,17 +1,20 @@
 name: CI benchmarks
 
 on:
-  pull_request:
-    types: [opened, synchronize, reopened, labeled]
+  pull_request_target:
+    types: [synchronize, reopened, labeled]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


nabobalis (approved):

ayshih:
For posterity...

This PR was replaced by an alternate approach (#8077).  The approach here required switching the triggering event from `pull_request` to `pull_request_target`, which introduced a number of issues that are non-trivial to impossible to resolve:

- `pull_request_target` is a potential security risk since we run someone else's code (which we have to do in order to benchmark it) with permissive permissions
- `pull_request_target` does not play nice with concurrency groups; a PR triggering the benchmarks CI build – even without the "Run benchmarks" label – would cancel any benchmarks CI build currently running, even on a different PR
- `pull_request` provides a convenient commit with the PR merged, but `pull_request_target` does not

Issue #8022, opened 2025-01-16
SRSClient lacks recent results


nabobalis:
### Describe the bug

The SRSClient uses "ftp://ftp.ngdc.noaa.gov/STP/swpc_products/daily_reports/solar_region_summaries/" which has had an entire arhcive from 1996 to "today" but this dataset has not been updated since 16-09-2024 and thus the online CI is failing on one test which checks for the lastest data.

We either need a better source URL or get someone from NOAA to fix this.

### To Reproduce

See the online CI:

```
FAILED ../../.tox/py312-online/lib/python3.12/site-packages/sunpy/net/dataretriever/sources/tests/test\_noaa.py::test\_srs\_start\_or\_end\_out\_of\_range - assert 0 > 0
where 0 = len(\<sunpy.net.dataretriever.client.QueryResponse object at 0x7f26636b7140>\\n\<No columns>)
```

pythonicforge:
Hey!

To fix this bug, we can have three possible solutions.

1. We can get a better source URL to particularly handle this case where the tests fail for an outdated data from NOAA. 

2. If it is possible, we can also get anyone from NOAA to get this thing fixed.

3. Or we can implement a fallback mechanism, that can modify the SRSClient to account for scenarios where the primary data source is unavailable. Like we can cache some historical data or rely on a mirrored dataset

pythonicforge:
I would like to work on this issue, but I need help on determining which path should I go to fix this.

nabobalis:
Ideally we can get number 2. I am not sure there are any other data sources we can use, that would have to be researched.

pythonicforge:
Exactly, I too think option 2 seems more reliable, but we still need to have some data sources to be researched before implementing it.

nabobalis:
What would we need to research for number 2?

pythonicforge:
I was talking in this perspective, if we consider the work around of implementing some other source to fetch data particularly for this scenario

pythonicforge:
Hey @nabobalis 

I did a mail to NOAA regarding this issue. I haven't heard from them yet, but I expect to get a feedback on this issue soon!

In the meanwhile, shall I start to search for some alternate sources for getting the SRS?

nabobalis:
Unless NOAA have another server we can use, there won't be another source I suspect.

pythonicforge:
[...]

PR #8081, opened 2025-02-10
xfail NOAA SRS FTP tests

wtbarnes:
The Scraper tests that test the FTP functionality use FTP endpoints for some recent NOAA SRS files. These are now failing because those files seem to no longer exist, presumably due to the migration of this dataset from FTP to HTTPS (see discussion in #8022). 

This PR just xfails those tests. We could also

- Skip them
- Delete them
- Find a new FTP endpoint to use in these tests

It doesn't seem that we have any remaining scraper clients that even use FTP so one could also make the case that we deprecate this functionality altogether...


nabobalis (approved):

Cadair:
I think if we don't have any clients left a slow deprecation would be a good idea.

Issue #8092, opened 2025-02-18
Automatic sorting by filename of wildcard input (e.g., to the Map factory) is inconsistent with native OS sorting behavior on Windows


tanishy7777:
### Describe the bug

The `test_patterns` function in `test_map_factory.py` fails on Windows, likely due to differences in sorting behavior between Linux and Windows.
This can be seen in the screenshot attached below

Running `pytest` on Windows 11, python version 3.13.1
![Image](https://github.com/user-attachments/assets/4093197f-16f7-4fe2-b3b1-b837136372b7)

### To Reproduce

run `pytest` in the terminal after setting up the dev environment

### System Details

python version: 3.13.1
sunpy version: 6.1.dev359+gd97980ad8
astropy version: 7.0.0
System info: Windows 11 x64

### Installation method

conda

ayshih:
The issue here is that the implementation has not matched the test since inception (see #3408), but this was only now discovered by the recent change to filenames in the tested directory, namely containing both files that start with lowercase "e" and files that start with uppercase "S".  Notably, on Windows (and perhaps other OSs), `S*` files are sorted after `e*` files, which is not the same as Linux (and other OSs).

### The implementation
Filenames are currently sorted by first running the wildcard input through `glob.glob()`, and then sorting the returned list of strings.  The consequence of this implementation is that the sorting depends on string sorting, which is OS-independent.
https://github.com/sunpy/sunpy/blob/67e305d522d4f8b73bc2355755ea613510860fc6/sunpy/util/io.py#L56-L59

### The test
The sorting is tested against sorting a list returned by `pathlib.Path.glob()`, which returns `Path` (e.g., `WindowsPath` objects, and thus the sorting follows the OS-dependent way of sorting filenames.
https://github.com/sunpy/sunpy/blob/67e305d522d4f8b73bc2355755ea613510860fc6/sunpy/map/tests/test_map_factory.py#L136

### The crux
Do we want the filename sorting to be OS-independent (which means we should fix the test to sort on strings rather than `Path` objects), or do we want the filename sorting to be OS-dependent (which means we should fix the implementation to sort on `Path` objects rather than strings)?

### Details
```python
>>> sorted(glob.glob("*"))
['SOHO_EIT_171_20070601T120013_L1.header',
 'SOHO_EIT_195_20070601T121346_L1.header',
 'SOHO_EIT_284_20070601T120607_L1.header',
 'SOHO_EIT_304_20070601T121937_L1.header',
[...]

PR #8093, opened 2025-02-18
Fixes a unit test since `sunpy.io` expands wildcard input with OS-independent sorting

tanishy7777:
Fixes: #8092

~This is a temporary fix for windows to pass the tests. I have marked the test `test_patterns` in `test_map_factory.py` as xfail.~

Edit by @ayshih: Fixes the test (see #8092 for discussion)

### DIFF
--- a/changelog/8093.bugfix.rst
+++ b/changelog/8093.bugfix.rst
@@ -1,1 +1,0 @@
-Marks `~sunpy.map.tests.test_patterns` as xfail for the tests to pass on windows.
--- a/changelog/8093.bugfix.rst
+++ b/changelog/8093.bugfix.rst
@@ -1,0 +1,1 @@
+Marks `~sunpy.map.tests.test_patterns` as xfail for the tests to pass on windows.
--- a/sunpy/map/tests/test_map_factory.py
+++ b/sunpy/map/tests/test_map_factory.py
@@ -133,6 +133,9 @@ def test_patterns(eit_fits_directory):
     assert ([isinstance(amap, sunpy.map.GenericMap) for amap in maps])
 
     # Test that returned maps are sorted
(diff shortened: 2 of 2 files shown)


sunpy/map/tests/test_map_factory.py
```diff
@@ -93,7 +93,7 @@ def test_composite():
 # Want to check that patterns work, so ignore this warning that comes from
 # the AIA test data
 
-
+@pytest.mark.xfail()
```
nabobalis, line 96: This should only xfail on windows.

changelog/8093.bugfix.rst
```diff
@@ -0,0 +1 @@
+Marks `~sunpy.map.tests.test_patterns` as xfail for the tests to pass on windows.
```
nabobalis, line 1: This does not need a changelog.

sunpy/map/tests/test_map_factory.py
nabobalis: Whatever works.

ayshih (changes requested):
Whoa, stop, why are we xfailing this?  At first glance, this is a bug we need to fix

tanishy7777:
> Whoa, stop, why are we xfailing this? At first glance, this is a bug we need to fix

I have added the changes as suggested in https://github.com/sunpy/sunpy/issues/8092#issuecomment-2667440345 and https://github.com/sunpy/sunpy/issues/8092#issuecomment-2667925412

ayshih (changes requested):
Please add a comment to explain this line for the benefit of future maintainers

tanishy7777:
> Please add a comment to explain this line for the benefit of future maintainers

I have added them

Issue #7655, opened 2024-05-31
Fido.fetch path with JSOC keywords doesn't work 


samaloney:
### Describe the bug

Following the docs it would seem like this should work `f = Fido.fetch(query, path='~/{date}/{file}')` for JSOC queries but it doesn't for valid reasons. 

I think this should at least be documented but ideally `query.path_format_keys()` would only return keys that work?

### To Reproduce

```
query = Fido.search(a.Time(f"2020-01-01", f"2020-01-01T00:01:00"), a.jsoc.Series.aia_lev1_uv_24s,
                    a.Sample(24*u.h), a.jsoc.Notify(<>), a.jsoc.Segment.image)
'date' in query[0].path_format_keys()
# True

f = Fido.fetch(query[0], path='~/{date}/{file}')

KeyError                                  Traceback (most recent call last)
Cell In[182], line 1
----> 1 f = Fido.fetch(query[0], path='~/{date}/{file}')

File /net/maedoc.ap.dias.ie/maedoc/home_cr/smaloney/.virtualenvs/arccnet/lib/python3.10/site-packages/sunpy/net/fido_factory.py:428, in UnifiedDownloaderFactory.fetch(self, path, max_conn, progress, overwrite, downloader, *query_results, **kwargs)
    425     raise ValueError(f"Query result has an unrecognized type: {type(query_result)} "
    426                      "Allowed types are QueryResponseRow, QueryResponseTable or UnifiedResponse.")
    427 for block in responses:
--> 428     result = block.client.fetch(block, path=path,
    429                                 downloader=downloader,
    430                                 wait=False, **kwargs)
    431     if result not in (NotImplemented, None):
    432         reslist.append(result)

File /net/maedoc.ap.dias.ie/maedoc/home_cr/smaloney/.virtualenvs/arccnet/lib/python3.10/site-packages/sunpy/net/base_client.py:324, in convert_row_to_table.<locals>.wrapper(self, query_results, **kwargs)
    322 if isinstance(query_results, QueryResponseRow):
    323     query_results = query_results.as_table()
--> 324 return func(self, query_results, **kwargs)

File /net/maedoc.ap.dias.ie/maedoc/home_cr/smaloney/.virtualenvs/arccnet/lib/python3.10/site-packages/sunpy/net/jsoc/jsoc.py:440, in JSOCClient.fetch(self, jsoc_response, path, progress, overwrite, downloader, wait, sleep, max_conn, **kwargs)
    438     else:
[...]

PR #8111, opened 2025-03-01
Fix: Handle invalid cases properly by catching errors in Fido fetch [#7655]

seika-afk:
Issue : Error on "date" or any other invalid case .

Image displaying the main issue

The code being for this issue :

Code for the Issue

Despite being True , on 'date' being on query , it throws error when passing 'date'
- i couldnt find the exact solution i.e being able to accept  'date'

But alternatively made it so that , it does not cause any error , but ignores that query-component , in case it is valid. And just show a Warning .

changes being :

Changes
Where the code passes through try and catch block , and in case it throws error in try block , it will go through catch block , and replace that component with empty string and , put a Warning .

if There are need for any changes or improvements, kindly let me know :smiley_cat:

### DIFF
--- a/sunpy/net/jsoc/jsoc.py
+++ b/sunpy/net/jsoc/jsoc.py
@@ -526,8 +526,17 @@ class JSOCClient(BaseClient):
                         fname = path.strip(ext)
                     else:
                         fname = path
-                    fname = fname.format(file=filename)
-                    fname = os.path.expanduser(fname)
+                   
+
+                    date_value=request.data.get('date') or request.data.get("DATE") or ""
+
+                    try :
+                        fname = fname.format(file=filename,date=date_value)
+                    except KeyError as e :
+                        log.warning(f"Skipping invalid key in path : {e}")
+                        fname=fname.replace(f"{{{e.args[0]}}}","")
+
+                    fname = os.path.expanduser(fname)   
                     paths.append(fname)
 
         dl_set = True


nabobalis:
Does {date} work with the other Fido clients?

seika-afk:
> Does `{date}` work with the other Fido clients?

Yes Sir, {date} works with other clients like VSO , HEK

> Actually does the JSOCClient format any of the other keywords?

Yes Sir, it does formats them , but just not {date} on the basis of how it is used her .

when i was testing the date with other clients , i tried using date differently and somehow it ended up working :

newer code :


 
and its output :

S


Here Somehow it works :

Key difference :
-Time Formating : Newer code Uses" test_date = "2023-01-01T00:00:00" "and searches for a single timestamp.
-File Retrieval approach : fetches all results in a loop , rather than just query[0]


as for what i should do about this pr. Should i leave it as it is , such that it shows warning , on any invalid prop, not just on date.
or Should I also Edit the documentation part of it  to explain this part , to avoid such issues for users?

nabobalis:
Can you check if we have unit tests for these cases and if not add them?

sunpy/net/jsoc/jsoc.py
```diff
@@ -526,8 +526,17 @@ def get_request(self, requests, path=None, overwrite=False, progress=True,
                         fname = path.strip(ext)
                     else:
                         fname = path
-                    fname = fname.format(file=filename)
-                    fname = os.path.expanduser(fname)
+                   
+
+                    date_value=request.data.get('date') or request.data.get("DATE") or ""
```
nabobalis, line 531: This date comes from the path or the client results?

seika-afk:
> Can you check if we have unit tests for these cases and if not add them?

Yes Sir, I would try my best. I will try to finish it before tomorrow.

unrelated stuff : (Also Sir I would like to apologize for inconsistent PR's ,due to college examinations, it got hindered, but i will try to be as consistent as possible)

nabobalis:
There is no rush, and college examinations are more important.

sunpy/net/jsoc/jsoc.py
seika-afk: > This date comes from the path or the client results?

As here, date is being extracted from request , It comes from client.

seika-afk:
> There is no rush, and college examinations are more important.

Thank you for your understanding sir.Now that examinations are over , i would try to do my best :smile_cat: .

seika-afk:
> Can you check if we have unit tests for these cases and if not add them?

Sir , I checked unit test cases  for "JSOClient" , "VSOClient" ,"HEKClient" , but did not find any test case , under : "sunpy/net/tests/test_fido.py."

So now adding test Cases for them .Before i proceed , Do you have any specific scenarios in mind that you’d like the tests to cover?
For example, should we focus on verifying {date} parsing, error handling, or consistency across clients?.

nabobalis:
> > Can you check if we have unit tests for these cases and if not add them?
> 
> Sir , I checked unit test cases for "JSOClient" , "VSOClient" ,"HEKClient" , but did not find any test case , under : "sunpy/net/tests/test_fido.py."
> 
> So now adding test Cases for them .Before i proceed , Do you have any specific scenarios in mind that you’d like the tests to cover? For example, should we focus on verifying {date} parsing, error handling, or consistency across clients?.

All of those?

seika-afk:
> All of those?

Yes Sir , I tried finding keywords , similar to them . 
-i did locate jsoc in terms : (a.jsoc.Series) . But unrelated to current topic.
-no direct for VSO and HEK
- hence no direct test case for {date} scenerio

nabobalis:
The HEK does no data downloading, so it won't matter for that one.

seika-afk:
> The HEK does no data downloading, so it won't matter for that one.

Got it Sir.Would you prefer I add a new test to specifically check this behavior , or do you have a particular approach in mind?

nabobalis:
> Got it Sir.Would you prefer I add a new test to specifically check this behavior for JSOC and VOS , or do you have a particular approach in mind?

Can you also check one of the dataretriever clients as well? 

Can you also please stop calling me sir.

nabobalis:
> > Can you also check one of the dataretriever clients as well?
> 
> Got it. Before i proceed , just for clarity ,I would like to confirm , whether do you want me to check if the dataretriever client handles {date} correctly, or are you asking me to check if there are existing test cases for it?

Check if there is an existing test, if so, we dont need to add one.

seika-afk:
> Check if there is an existing test, if so, we dont need to add one.

Yes , the "test_client.py" on "sunpy/net/dataretriever/tests/"
certainly includes test cases to date.
-It verifies that QueryResponse correctly handles satrt and end times using parse time . However it doesn not explicitly test {date} handling for dataretriever clients.

on lines (47-50):
```
qr = QueryResponse.create(map_, client=client)
assert qr.time_range.start == parse_time("2013/2/7T4:00:00")
assert qr.time_range.end == parse_time("2013/2/7T5:00:00")

```

PR #8126, opened 2025-03-13
Allow non-FITS-standard units on HMI map

wtbarnes:
Override `unit` on `HMIMap` to allow for units outside of the FITS standard. The main motivation here is to allow for "Mx/cm2" as a unit which is commonly used in the HMI SHARP data products. Without this, warnings are repeatedly issued and `None` is returned as the unit.

We were already allowing for this on the synoptic maps so we may as well do it for all HMI maps.


sunpy/map/sources/sdo.py
```diff
@@ -165,6 +165,15 @@ def reference_date(self):
...
         self.meta['T_OBS'] = parse_time(date).utc.isot
 
+    @property
+    def unit(self):
+        unit_str = self.meta.get('bunit', None)
+        if unit_str is None:
+            return
+        # Maxwells aren't in the IAU unit style manual and therefore not a valid FITS unit
+        # The mapbase unit property forces this validation, so we must override it to prevent it.
+        return u.Unit(unit_str)
```
ayshih, line 175: If we do this, I feel it should be done only for the specific unit expected ("Mx / cm2") rather than allowing any unit

wtbarnes: Agreed. I just copied this straight from `HMISynoptic` source where we were already doing this.

nabobalis:
??

wtbarnes:
oops

wtbarnes:
Online test fails are unrelated

nabobalis (approved):

nabobalis (approved):

Cadair (approved):

PR #8133, opened 2025-03-21
PUNCHMap

### LINKED ISSUE #8132: Add support for PUNCH to Map
### Describe the feature

After some discussion with SunPy folks (thanks!) we'd like to contribute adding support for PUNCH 2D data to the SunPy Map class. We can also add the PUNCH colormap to SunPy. We'll leave the more complex layered data for NDCube usage, and try to nudge users in that direction with example notebooks, etc.

### Proposed solution

We've created a PUNCH map class along with tests, and will create a pull request shortly for feedback to help integrate it into SunPy.

lowderchris:
## PR Description

After some discussion with SunPy folks, we've worked on creating a PUNCHMap class to handle two-dimensional PUNCH data. Closes #8132 . This pull request:

- Adds the core PUNCHMap class file
- Tests this class with a sample PUNCH header file
- Integrates the PUNCH colormap

Definitely let us know of any changes we can make - thanks!


sunpy/map/sources/punch.py
```diff
@@ -0,0 +1,64 @@
...
+        """Instrument identifier."""
+        return self.meta.get("INSTRUME")
+
+
+    # Used by the Map factory to determine if this subclass should be used
+    @classmethod
+    def is_datasource_for(cls, data, header, **kwargs):
+        """Determines if data, header corresponds to a PUNCH image."""
+        # Returns True only if this is data and header from PUNCH
+        return header.get('instrume', '').startswith('NFI') or header.get('instrume', '').startswith('WFI') or header.get('obsrvtry', '').startswith('PUNCH')
```
wtbarnes, line 64:
```suggestion
        instrument = str(header.get('instrume', ''))
        return instrument.startswith('NFI') or instrument.startswith('WFI') or header.get('obsrvtry', '').startswith('PUNCH')
```
The suggested changed should fix the currently failing test. 

Does this source intentionally not require the 'OBSRVTRY' key to start with 'PUNCH'? As written, *any* FITS file that has an instrument key that start with 'NFI' or 'WFI' would be read as a PUNCH map.

sunpy/map/sources/punch.py
```diff
@@ -0,0 +1,64 @@
...
+    * `PUNCH Mission Page <https://punch.space.swri.edu>`__
+    * `PUNCH SOC Development <https://github.com/punch-mission/>`__
+    """
+
+    def __init__(self, data, header, **kwargs):
+        super().__init__(data, header, **kwargs)
+        self._nickname = self.observatory + ' - ' + self.instrument
+        self.plot_settings["cmap"] = "punch"
+        self.plot_settings["title"] = f"{self._nickname}"
+        self.plot_settings["norm"] = LogNorm(vmin=1.77e-15, vmax=3.7e-11)
```
wtbarnes, line 32: Where do these numbers come from? Are they generic enough to be suitable default values for all PUNCH images? If not, we typically leave the minimum and maximum values unset.

wtbarnes (changes requested):
Sorry this has taken so long to review! I've attached a few suggested changes. Looking at the sample header provided, I don't think many of these properties are needed with the exception of `unit` which I think needs some additional logic.

wtbarnes:
pre-commit ci autofix

wtbarnes:
pre-commit.ci autofix

sunpy/map/sources/punch.py
lowderchris: Ah, yes, these were selected to bring out details in particular ranges of the data, and should probably not be hardcoded like that. I'll remove this and make sure that the default still captures the data.

sunpy/map/sources/punch.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+    @property
+    def reference_date(self):
+        """Reference date for the coordinate system."""
+        return self._get_date('DATE-OBS') or super().reference_date
```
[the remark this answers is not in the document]
lowderchris: Aha, I see, we had a bit of back and forth on standards for the various date keywords, but we can certainly jump back to the default for this, which should work just fine.

sunpy/map/sources/punch.py
```diff
@@ -0,0 +1,51 @@
...
+    * `PUNCH Mission Page <https://punch.space.swri.edu>`__
+    * `PUNCH SOC Development <https://github.com/punch-mission/>`__
+    """
+
+    def __init__(self, data, header, **kwargs):
+        super().__init__(data, header, **kwargs)
+        self.nickname = f"{self.observatory} - {self.instrument}"
+        self.plot_settings["cmap"] = "punch"
+
+        self.plot_settings["norm"] = LogNorm()
```
wtbarnes, line 34:
```suggestion
        self.plot_settings["norm"] = ImageNormalize(
            stretch=source_stretch(self.meta, LogStretch()), clip=False)
```
Sorry for the additional suggestion. Better to use normalizations provided by astropy. You'll also need:

```python
from astropy.visualization import ImageNormalize, LogStretch
```

lowderchris: Oh, no worries at all - we really welcome the best practice suggestions here!

sunpy/data/test/punch.header
```diff
@@ -0,0 +1,124 @@
+COMMENT ----- FITS Required ----------------------------------------------------
```
Cadair, line 1: This isn't strictly a SunPy related comment:

Is it valid to have a COMMENT keyword before the `SIMPLE = 'T'` keyword? Section 6.1.1 of the FITS 4 spec states:

> The SIMPLE keyword is required to be the first keyword in the primary header of all FITS files

sunpy/map/sources/punch.py
```diff
@@ -0,0 +1,52 @@
...
+            return
+        return u.def_unit(unit_str)
+
+    # Used by the Map factory to determine if this subclass should be used
+    @classmethod
+    def is_datasource_for(cls, data, header, **kwargs):
+        """Determines if data, header corresponds to a PUNCH image."""
+        # Returns True only if this is data and header from PUNCH
+        instrument = str(header.get('instrume', ''))
+        return instrument.startswith('NFI') or instrument.startswith('WFI') or header.get('obsrvtry', '').startswith('PUNCH')
```
Cadair, line 52: Can we not just rely on `OBSRVTRY = "PUNCH"`?

sunpy/visualization/colormaps/color_tables.py
```diff
@@ -376,3 +377,7 @@ def euvi_color_table(wavelength: u.angstrom):
             "Invalid EUVI wavelength. Valid values are "
             "171, 195, 284, 304."
         )
+
+
+def punch_color_table():
+    return cmap_from_rgb_file("PUNCH", "punch.csv")
```
Cadair, line 383: I think you also need to modify `cm.py` to add it to the list that get's loaded by matplotlib.

Cadair (commented):
Thanks again @lowderchris sorry for the slow review.

sunpy/map/sources/punch.py
lowderchris: That should work, and is much simpler. Originally there was some thought about how to classify data from individual satellites before being merged into a mosaic. Is an individual spacecraft frame PUNCH, or only the mosaic? In the end with the metadata we settled on, OBSRVTRY should always be PUNCH, with INSTRUME and TELESCOP containing the individual spacecraft IDs.

sunpy/data/test/punch.header
lowderchris: Aha, good catch. We validated all of our generated headers with the NASA FITS verifier tool, and it didn't complain, but maybe it should have? Or perhaps a COMMENT card is okay (is it a keyword technically?). We could change this, but it would require some rewriting of our metadata generator in our data pipeline. What do you all think?

Cadair: COMMENT is frequently referred to as a "keyword" in the spec, so my interpretation would be that it is. I think to be safe I would make sure SIMPLE was first. If you look at the first paragraph of 4.4.1.1 it says that the first place you could put a COMMENT key would be after the last NAXISn keyword.

lowderchris: That makes sense. We can take a look at modifying our header generator functions to account for this, and will upload a new sample header from that. Thanks for catching that!

sunpy/map/sources/punch.py
```diff
@@ -0,0 +1,64 @@
...
+        """Reference date for the coordinate system."""
+        return self._get_date('DATE-OBS') or super().reference_date
+
+    @property
+    def unit(self):
+        """Units of the observation."""
+        unit_str = self.meta.get('bunit', self.meta.get("BUNIT"))
+        if unit_str is None:
+            return
+        return u.def_unit(unit_str)
```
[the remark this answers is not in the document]
lowderchris: Absolutely - I made some changes that will hopefully address this, and we're also updating our main header template to reflect this over in punchbowl. Will coordinate changes between the two. Let me know if this looks alright.

sunpy/map/sources/punch.py
```diff
@@ -0,0 +1,58 @@
...
+    def unit(self):
+        """Units of the observation."""
+        unit_str = self.meta.get('bunit', self.meta.get("BUNIT"))
+        if unit_str is None:
+            return
+        parsed_unit = u.Unit(unit_str)
+        if parsed_unit == u.Unit("2.009e+07 W / (sr m2)") or parsed_unit == u.Unit("sqrt(DN)"):
+             return parsed_unit
+        else:
+             return super().unit
```
wtbarnes, line 51: What's the reason for overriding `unit`? If those values are in the "BUNIT" key, then I think the default Map behavior should just work.

sunpy/map/sources/punch.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Used by the Map factory to determine if this subclass should be used
```
nabobalis, line 53:
```suggestion
```

sunpy/map/sources/punch.py
```diff
@@ -0,0 +1,58 @@
...
+    @classmethod
+    def is_datasource_for(cls, data, header, **kwargs):
+        """Determines if data, header corresponds to a PUNCH image."""
+        # Returns True only if this is data and header from PUNCH
```
nabobalis, line 57:
```suggestion
```

sunpy/map/sources/punch.py
```diff
@@ -0,0 +1,58 @@
...
+    consisting of four satellites, together imaging the solar corona in polarized white light.
+    Coordinated observations from three Wide Field Imagers (WFIs) and one Near Field Imager (NFI)
+    are processed and meshed into one virtual observatory with a field of view spanning
+    45-degrees in radius away from the Sun.
+
+    PUNCH launched and began operations on 11 March 2025.
+
+    Notes
+    -----
+
```
nabobalis, line 25:
```suggestion
```

changelog/8133.feature.rst
```diff
@@ -0,0 +1 @@
+Adds support for PUNCH data within the SunPy Map class.
```
nabobalis, line 1:
```suggestion
Added support for PUNCH data within `sunpy.map.Map`.
```

lowderchris:
I've made some changes locally that I'll push up by the end of the day - but it should hopefully resolves the units issue, and the updated header. Thank you all for your patience with this!

sunpy/map/sources/punch.py
```diff
@@ -0,0 +1,52 @@
...
+    # def unit(self):
+    #     """Units of the observation."""
+    #     unit_str = self.meta.get('bunit', self.meta.get("BUNIT"))
+    #     if unit_str is None:
+    #         return
+    #     parsed_unit = u.Unit(unit_str)
+    #     if parsed_unit == u.Unit("2.009e+07 W / (sr m2)") or parsed_unit == u.Unit("sqrt(DN)"):
+    #          return parsed_unit
+    #     else:
+    #          return super().unit
```
nabobalis, line 47:
```suggestion
```

sunpy/map/sources/punch.py
```diff
@@ -0,0 +1,52 @@
...
+    ----------
+    * `PUNCH Mission Page <https://punch.space.swri.edu>`__
+    * `PUNCH SOC Development <https://github.com/punch-mission/>`__
+    """
+
+    def __init__(self, data, header, **kwargs):
+        super().__init__(data, header, **kwargs)
+        self.nickname = f"{self.observatory} - {self.instrument}"
+        self.plot_settings["cmap"] = "punch"
+
```
nabobalis, line 32:
```suggestion
```

sunpy/map/sources/punch.py
lowderchris: This is what I get for commenting out to test and forgetting to remove the old code chunk. Sorry about that!

lowderchris:
@Cadair absolutely - again sorry for the delay on this! Did those changes address everything? I'm happy to do any last fixes today as well if that would make things easier.

nabobalis (approved):

Cadair:
@lowderchris Thanks! We normally write up a little what's new entry for things like this as we are preparing a release. Do you have an example file / plot that would be good to include to show this off?

wtbarnes (approved):

lowderchris:
> @lowderchris Thanks! We normally write up a little what's new entry for things like this as we are preparing a release. Do you have an example file to use / plot you've made that would be good to include to show this off?

Absolutely - would something like the attached plot work? I used this new code to load some PUNCH / WFI data to plot as a map.
punchmap

nabobalis:
Looks great to me! Lets use it.

Cadair:
Thanks very much for your contribution @lowderchris :partying_face:

PR #8134, opened 2025-03-25
Add `zizmor` as a `pre-commit` hook to find potential security vulnerabilities in GitHub Actions

namurphy:
This PR enables a `pre-commit` hook for [`zizmor`](https://woodruffw.github.io/zizmor/), which is a static analysis tool to detect common CI/CD security issues in GitHub workflows that I originally heard about through pyOpenSci.  Scientists tend not to be experts in security over the World Wide Web™, but tools like `zizmor` can help us reduce the number of security vulnerabilities in the heliopythoniverse.  

I attempted this for PlasmaPy first in https://github.com/PlasmaPy/PlasmaPy/pull/2975.  `zizmor` brought up twice as many security vulnerabilities for PlasmaPy than for SunPy, so I guess we're winning over there? 😅

I don't have a good understanding yet regarding most of the issues that `zizmor` is finding, so a possible path forward would be: 

 - Create issues to address the potential security vulnerabilities
 - _Temporarily_ exclude the files that `zizmor` is finding problems in from this hook
 - Merge this PR so that `zizmor` checks new workflow files and some of the existing files
 - Address the potential security vulnerabilities over time


namurphy:
This PR is presently ignoring `.github/workflows/scheduled_builds.yml` since this file is missing the `runs` field.  This was the full error from `zizmor`:

```
failed to register input: .github/workflows/scheduled_builds.yml

Caused by:
    0: invalid GitHub Actions workflow: file://.github/workflows/scheduled_builds.yml
       
       Caused by:
           jobs: data did not match any variant of untagged enum Job at line 21 column 3
    1: invalid GitHub Actions definition: file://.github/workflows/scheduled_builds.yml
       
       Caused by:
           missing field `runs` at line 7 column 1
    2: failed to register input as workflow or action
```

Cadair:
thanks @namurphy I think this will want to get put into the package template, rather than here ultimately, but we can work through issues with this before that.

I do wonder if the pre-commit hooks are the best place for this as opposed to a dedicated github actions job which can be a bit more selective about when it runs.

namurphy:
> I think this will want to get put into the package template, rather than here ultimately, but we can work through issues with this before that.

Agreed! I had the same thought, only a few hours after I submitted this. 😅 

> if the pre-commit hooks are the best place for this as opposed to a dedicated github actions job

That seems like a very reasonable place to put it, perhaps with a corresponding tox environment so that it can be run locally the same way. 🤔

Cadair:
> perhaps with a corresponding tox environment so that it can be run locally the same way. 🤔

:+1:

nabobalis:
There was a discussion around this tool at the Scientific Python summit and its great but due to some false positives, people suggested it would be better to run it over a repo and fix any issues it finds instead of adding it permanently.

Cadair:
I was thinking cron CI or similar. We have many things to fix tho

namurphy:
🪧 Closing this in favor of adding zizmor to the package template.

PR #8155, opened 2025-04-30
Test parfive rc

Cadair:


Cadair:
I think this is now failing in the same way as main?

nabobalis:
VSO issues look upstream, we should change the test to not use that datetime.

PR #8161, opened 2025-05-03
Speed up mesh-based autoalign plotting of Maps

ayshih:
For autoalign plotting of Maps (using `pcolormesh()` internally), it's nice that we can pass matplotlib the pixel->pixel transformation via the `transform` keyword, but that means that matplotlib performs the transformation with every rendering update (e.g., panning).  Instead, we can refactor the approach so that the pixel->pixel transformation is performed only once, and then panning performance greatly improves.  `pcolormesh()` still needs time to render all the data pixels, but `autoalign=True` is far less miserable now for interactive plots.

As a bonus, because we compute the transformation ourselves instead of having matplotlib do it, we can now properly update the plot limits.  Previously, matplotlib's automatic detection would get thrown off by any NaNs in the coordinates.

### DIFF
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -17,7 +17,6 @@ import matplotlib.pyplot as plt
 import numpy as np
 from matplotlib.backend_bases import FigureCanvasBase
 from matplotlib.figure import Figure
-from matplotlib.transforms import Transform
 
 try:
     from dask.array import Array as DaskArray
@@ -61,6 +60,7 @@ from sunpy.util.functools import seconddispatch
 from sunpy.util.util import _figure_to_base64, fix_duplicate_notes
 from sunpy.visualization import axis_labels_from_ctype, peek_show, wcsaxes_compat
 from sunpy.visualization.colormaps import cm as sunpy_cm
+from sunpy.visualization.visualization import _PrecomputedPixelCornersTransform
 
 TIME_FORMAT = config.get("general", "time_format")
 PixelPair = namedtuple('PixelPair', 'x y')
@@ -2846,27 +2846,18 @@ class GenericMap(NDData):
                 imshow_args.setdefault('antialiased', True)
                 imshow_args.setdefault('linewidth', 0)
 
-            # Transform the data corners to the pixel space of the axes
-            self_y, self_x = np.indices((data.shape[0] + 1, data.shape[1] + 1)) - 0.5
-            axes_x, axes_y = astropy.wcs.utils.pixel_to_pixel(self.wcs, axes.wcs, self_x, self_y)
-
-            # Create a lookup table for matplotlib to use
-            class PrecomputedTransform(Transform):
-                input_dims = 2
-                output_dims = 2
-                def transform_non_affine(self, values):
-                    ix, iy = (values + 0.5).astype(int).T
-                    return np.stack([axes_x[iy, ix], axes_y[iy, ix]], axis=1)
+            # Create a lookup table for the transformed data corners for matplotlib to use
+            transform = _PrecomputedPixelCornersTransform(axes, self.wcs)
 
             # Define the mesh in data coordinates in case the transformation results in NaNs
-            ret = axes.pcolormesh(self_x, self_y, data,
+            ret = axes.pcolormesh(transform.data_x, transform.data_y, data,
                                   shading='flat',
-                                  transform=PrecomputedTransform() + axes.transData,
+                                  transform=transform + axes.transData,
                                   **imshow_args)
 
             # Calculate the bounds of the mesh in the pixel space of the axes
-            good = np.logical_and(np.isfinite(axes_x), np.isfinite(axes_y))
-            good_x, good_y = axes_x[good], axes_y[good]
+            good = np.logical_and(np.isfinite(transform.axes_x), np.isfinite(transform.axes_y))
+            good_x, good_y = transform.axes_x[good], transform.axes_y[good]
             min_x, max_x = np.min(good_x), np.max(good_x)
             min_y, max_y = np.min(good_y), np.max(good_y)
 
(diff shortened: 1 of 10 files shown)


ayshih:
So...  This PRs approach does work, but only when the pixel->pixel transformation does not return any NaNs.  `pcolormesh()` does not accept NaNs as input, even though it can clearly handle NaNs that are generated internally by the `transform` function.  The similar, but slower, plotting method `pcolor()` does accept NaNs, but it qualitatively feels like there is no net gain in performance by precomputing the pixel->pixel transformation and then using a slower plotting method.

I need to mull this one over.

Cadair:
Time to open an issue upstream?

ayshih:
@Cadair: As I am wont to, my latest commit monkey patches around the input check in matplotlib. =)

ayshih:
After lots of testing, I have decided that they are making the correct decision to not allow NaN input to `pcolormesh()` (at least by default).  I've adjusted this PR so that the precomputation is stored as a lookup table in a matplotlib transform, and so `pcolormesh()` can be happy and rendering does not require recomputation.

sunpy/map/mapbase.py
```diff
@@ -2841,17 +2840,41 @@ def plot(self, *, annotate=True, axes=None, title=True, autoalign=False,
...
             if imshow_args.get('alpha', 1) != 1:
                 imshow_args.setdefault('antialiased', True)
                 imshow_args.setdefault('linewidth', 0)
 
-            ret = axes.pcolormesh(np.arange(data.shape[1] + 1) - 0.5,
-                                  np.arange(data.shape[0] + 1) - 0.5,
-                                  data, **imshow_args)
+            # Transform the data corners to the pixel space of the axes
+            self_y, self_x = np.indices((data.shape[0] + 1, data.shape[1] + 1)) - 0.5
+            axes_x, axes_y = axes.wcs.world_to_pixel(self.wcs.pixel_to_world(self_x, self_y))
```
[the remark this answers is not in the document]
ayshih: > Isn't this just "pixel_to_pixel" in astropy?

Heh, indeed, I didn't know that existed.

> Alternatively, what about `axes.get_transform(self.wcs)(self_x, self_y)`?

The matplotlib transforms all want 1D, whereas I want to keep it 2D for the lookup table.

sunpy/map/mapbase.py
```diff
@@ -2841,17 +2840,41 @@ def plot(self, *, annotate=True, axes=None, title=True, autoalign=False,
...
+
+            # Create a lookup table for matplotlib to use
+            class PrecomputedTransform(Transform):
+                input_dims = 2
+                output_dims = 2
+                has_inverse = False
+                def transform_non_affine(self, values):
+                    ix, iy = (values + 0.5).astype(int).T
+                    return np.stack([axes_x[iy, ix], axes_y[iy, ix]], axis=1)
```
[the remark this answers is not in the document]
ayshih: I'm hesitant to expose this precomputed transform.  It's safe to use internally because we know exactly what inputs `pcolormesh()` will be sending it: the very pixel corners that had already been computed.  An exposed implementation would have to make decisions or give options for how arbitrary inputs would be handled (e.g., truncating/rounding/interpolating), which could be (slightly) detrimental to performance.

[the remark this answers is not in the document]
Cadair: I'm ok with not making it public, but I do feel like it would be neater if it were self-contained and in it's own place rather than using variables from the containing scope etc.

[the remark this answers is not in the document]
ayshih: Okay, I've moved the class to be a private (for now) class in `sunpy.visualization`

sunpy/map/mapbase.py
```diff
@@ -2762,17 +2763,15 @@ def plot(self, *, annotate=True, axes=None, title=True, autoalign=False,
...
-        beyond the solar disk may not appear, which may also inhibit Matplotlib's
-        autoscaling of the plot limits. The plot limits can be set manually.
-        To preserve the off-disk parts of the map, using the
-        `~sunpy.coordinates.SphericalScreen` context
-        manager may be appropriate.
+        beyond the solar disk may not appear.  To preserve the off-disk parts of the
+        map, using the `~sunpy.coordinates.SphericalScreen` context manager may be
+        appropriate.
         """
         # Todo: remove this when deprecate_positional_args_since is removed
```
Cadair, line 2776: We can pull this for 7.0, don't have to do it now, just saw it.

sunpy/visualization/visualization.py
```diff
@@ -97,3 +100,26 @@ def new_format_method(values, spacing, format="auto"):
...
     declination_axis.set_axislabel('Helioprojective Impact Angle')
+
+
+# TODO: Consider making this transform class public
+class _PrecomputedPixelCornersTransform(Transform):
+    input_dims = 2
+    output_dims = 2
+
+    def __init__(self, axes, data_wcs):
+        """
```
Cadair, line 111: move this docstring up to the class level?

Cadair (approved):

sunpy/map/mapbase.py
ayshih: Removed in #8165

nabobalis: This PR was not rebased with that in?

sunpy/visualization/visualization.py
```diff
@@ -97,3 +100,26 @@ def new_format_method(values, spacing, format="auto"):
...
     declination_axis._formatter_locator.formatter = new_format_method
     declination_axis.set_axislabel('Helioprojective Impact Angle')
+
+
+# TODO: Consider making this transform class public
+class _PrecomputedPixelCornersTransform(Transform):
+    """
+    A matplotlib transform that precomputes the pixel->pixel transformation from a
+    given data WCS to the WCS of the given WCSAxes, and stores it as a lookup table
+    to avoid redundant coordinate transformations.  This transformation is computed
```
nabobalis, line 110:
```suggestion
    to avoid redundant coordinate transformations. This transformation is computed
```

Issue #8172, opened 2025-05-07
RTD pinned to 3.12 due to 3.13 stripping out whitespace


nabobalis:
The compiler now strips common leading whitespace from every line in a docstring. This reduces the size of the bytecode cache (such as .pyc files), with reductions in file size of around 5%, for example in sqlalchemy.orm.session from SQLAlchemy 2.0.

See https://github.com/python/cpython/issues/81283

This means that a lot of the doc magic we do for Map is broken on 3.13. 

Potential fixes:

https://github.com/jupyter/jupyter_client/pull/1052/files
https://github.com/scipy/scipy/pull/21104/files

PR #8173, opened 2025-05-08
Made dynamically constructed docstrings compatible with all Python versions

ayshih:
Python 3.13+ automatically cleans up common indentations in docstrings, which breaks the assumptions behind our `add_common_docstring()` decorator and `fix_duplicate_notes()` function.  This PR fixes them by always removing any common indentations prior to manipulating the docstring.  Along the way, I fixed the wildly wrong argument names and documentation for `fix_duplicate_notes()` and refactored its internals.

For Python 3.12 and earlier, this will technically change a couple of our docstrings when accessed directly via `.__doc__`, but it won't change docstrings as retrieved by `inspect` or as rendered by Sphinx (which uses `inspect`).  Thus, I do not consider this to be a breaking change.

Fixes https://github.com/sunpy/sunpy/issues/8172


nabobalis (approved):

ayshih:
Docs build confirmed to succeed without warnings/errors on both Python 3.12 and 3.13

I believe the just-merged #8165 means that the backports here will have to be manually done

sunpy/util/tests/test_util.py
```diff
@@ -121,3 +121,88 @@ def set_test6_n(self, *args, **kwargs):
             pass
 
     assert util.get_set_methods(A()) == {'test2', 'test6_n'}
+
+
+def test_notes_combined():
+    original_documentation = """
```
Cadair, line 127: eugh I hate it when things are dedented like this, would using `textwrap.dedent` defeat the object of the exercise?

Cadair (approved):

sunpy/util/tests/test_util.py
ayshih: That's fair.  It's ugly as hell.  I'll fix it.

ayshih: Done (using `inspect.cleandoc()`)

PR #8178, opened 2025-05-09
Added the convenience function `sunpy.util.grid_perimeter()`

ayshih:
Both #8040 and #8161 use this new `grid_perimeter()` function, but I decided to break the relevant commit out to a separate PR since I don't know which of those two PRs will go in first, and I don't want this commit to get squashed with other commits.


nabobalis (approved):

Cadair (commented):
Some kind of test?

ayshih:
Ha, already implicitly tested, but good point

nabobalis:
implicitly is good enough for me

PR #8179, opened 2025-05-09
Changed source-code links in documentation to go to GitHub

ayshih:
This is copied largely from [matplotlib](https://github.com/matplotlib/matplotlib/blob/6dcfa9d94bff3d672b59a57e98b1d94ed624b02a/doc/conf.py#L792-L860).  Let's see if it works.

### DIFF
--- a/changelog/8179.doc.rst
+++ b/changelog/8179.doc.rst
@@ -1,0 +1,1 @@
+The source-code links now go to the GitHub repository instead of rendered versions of the source code.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -7,6 +7,8 @@ import os
 import sys
 import datetime
 import warnings
+import tokenize
+from pathlib import Path
 
 from packaging.version import Version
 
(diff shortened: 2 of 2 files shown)


Cadair (commented):
If this works it should go in the package template and get rolled out to all the repos.

ayshih:
This works, but:

- These new source-code links do not work for certain classes because we overwrite their `__module__` attribute and that confuses `inspect`.  These are the classes in `sunpy.net.attrs` and the class `NorthOffsetFrame`.  That said, these classes were also missing source-code links prior to this PR, so this PR did not make anything worse.
- Someone other than me can determine if/how this can go into the package template.  There are currently a couple of fiddly bits that are specific to being part of the core's `conf.py`.

nabobalis (approved):

nabobalis:
Works on RTD.

Cadair:
Ok, I was going to try and move this into the package template before we merged it, but I guess afterwards is fine too.

nabobalis:
> Ok, I was going to try and move this into the package template before we merged it, but I guess afterwards is fine too.

SHOULD BE FASTER THEN

PR #8184, opened 2025-05-10
Add support of row & col slicing in timeseries.TimeSeries

### LINKED ISSUE #6424: Improving functionality of timeseries.TimeSeries (pull out column into timeseries, indexing, slicing)
### Describe the feature

It would be nice to be able to slice out a column from a `TimeSeries` and it returns TimeSeries (holding units, meta). At the moment the only way to do this is to strip it to a pd.DataFrame and pull out the DataFrame column, which isn't ideal if you want to keep associated units etc. 

e.g. 
```python
goes_xrs = sunpy.timeseries.TimeSeries(sunpy.data.sample.GOES_XRS_TIMESERIES)
goes_xrs["xrsa"]
``` 
and this return a sunpy.timeseries.TimeSeries of just the "xrsa" column.

This would also be great for plotting purposes (I know you can pass a column to .plot) but to have individual column timeseries also has its use cases. At the moment, I basically use TimeSeries as a reader and then convert to a pd.DataFrame but this shouldn't be the case. 

It would also be nice to index a TimeSeries at a given index or time, for example

```python
goes_xrs[0]
```

or pass a list or array of indices
```python
goes_xrs[indices]
```
and it would return a TimeSeries at these. I mean maybe you want a method on it instead rather than indexing, such like "truncate_by_indices" or something. but the functionality would be handy for sure.

At the moment to do this you have to pull it out and create a new TimeSeries

```python 
indices = np.arange(100)
goes_df = goes_xrs._data.iloc[[indices]
goes_units = goes_xrs.units
new_xrs = ts.TimeSeries(goes_df, goes_units)

```

### Proposed solution

scope out is Astropyy timeseries can do this, or implement it within the current stack as it would be nice to have soon

ankitkhushwaha:
fixes #6424

### DIFF
--- a/sunpy/timeseries/timeseriesbase.py
+++ b/sunpy/timeseries/timeseriesbase.py
@@ -260,6 +260,36 @@ class GenericTimeSeries:
     def __repr__(self):
         return f"{object.__repr__(self)}\n{self}"
 
+    def __getitem__(self, row_key = None, col_key = None):
+        ts = copy.deepcopy(self)
+
+        if row_key is None:
+            row_key = slice(None)
+        if col_key is None:
+            col_key = slice(None)
+
+        # Row slicing
+        if isinstance(row_key, slice):
+            ts._data = ts._data.iloc[row_key, :]
+        elif isinstance(row_key, TimeRange):
+            ts._data = ts._data.iloc[row_key.start.value:row_key.end.value]
+        elif isinstance(row_key, pd.DatetimeIndex):
+            ts._data[row_key[0]:row_key[-1]]
+        else:
+            if isinstance(row_key, int):
+                col_key = [col_key]
+            ts._data = ts._data.iloc[col_key]
+
+        # Columns slicing
+        if isinstance(col_key, slice):
+            ts._data = ts._data.iloc[:, col_key]
+        else:
+            if isinstance(col_key, str):
+                col_key = [col_key]
+            ts._data = ts._data[row_key]
+
+        return ts
+
     def _repr_html_(self):
         """
         Produces an HTML summary of the timeseries data with plots for use in


ankitkhushwaha:
all the required things are implemented. @nabobalis

sunpy/timeseries/tests/test_timeseriesbase.py
```diff
@@ -520,6 +520,56 @@ def test_repr_html_(generic_ts):
...
+        ),
+    ])
+def test_row_col_slicing(generic_ts, row, col):
+    ts = generic_ts
+    sliced_ts = generic_ts[row, col]
+
+    if row is None:
+        row = slice(None)
+    if col is None:
+        col = slice(None)
```
nabobalis, line 545: Is this handled internally by getitem?

sunpy/timeseries/tests/test_timeseriesbase.py
```diff
@@ -520,6 +520,56 @@ def test_repr_html_(generic_ts):
...
+
+    # Columns slicing
+    if isinstance(col, slice):
+        ts._data = ts._data.iloc[:, col]
+    else:
+        if isinstance(col, str):
+            col = [col]
+        ts._data = ts._data[col]
+
+    assert sliced_ts == generic_ts
```
nabobalis, line 570: I am confused as to what this test is testing? Shouldn't the sliced TS not be the same?

ankitkhushwaha:
Oops! My bad,
 will fix this things

sunpy/timeseries/tests/test_timeseriesbase.py
ankitkhushwaha: yes

sunpy/timeseries/tests/test_timeseriesbase.py
ankitkhushwaha: i have chnaged the ts

wtbarnes:
I'm going to close this PR as it has not seen activity for some time. If there is interest in reviving it, feel free to reopen.

PR #8187, opened 2025-05-12
Automatically detect which autoalignment approach to use for Map plotting

ayshih:
The `autoalign` keyword for Map plotting provides a mesh-based approach (significantly improved in #8161) and an image-based approach (#8036).  This PR makes `autoalign=True` automatically determine whether the image-based approach will suffice or whether the mesh-based approach needs to be used.

This PR also fixes a couple of minor issues with #8036:

* Corrected the indices being used to extract the corners from the perimeter
* Replaced a call to a private matplotlib method with a call to a public one


Cadair (approved):

ayshih:
`autoalign=True` is now the default setting.

I also fixed a bug with mesh-based autoalign where the mesh would be zordered on top of all images.

nabobalis (approved):

PR #8199, opened 2025-05-14
Removed URLS for scripts for miniforge and point to the download page instead

nabobalis:
We get this on stable:
Screenshot 2025-05-14 at 3 28 56 PM

For some reason it only affects stable, not latest, none of the tagged releases. 

The hope is, by removing the urls to exe and sh files, we can restore this page.


nabobalis:
@mihaimaruseac

Cadair:
> The hope is, by removing the urls to exe and sh files, we can restore this page.

Wouldn't the way to be sure be to change the url? i.e. change the filename of the page.

Cadair (approved):

nabobalis:
Seems to be fixed now!

PR #7437, opened 2024-02-06
Adds a warn when using both assume_spherical_screen() and propagate_with_solar_surface() context managers with reproject_to()

Deus1704:
## PR Description
As mentioned in the #7172, this PR warns the user when both the context managers are active at the same time.

PR #8211, opened 2025-05-21
Fixed a bug with tracking our active contexts

ayshih:
#7437 added a warning when both the `assume_spherical_screen()` (now, `SphericalScreen()`) and `propagate_with_solar_surface()` context managers were used together, but there is a bug in the implementation of tracking active contexts that gets confused if a given context manager is used multiple times in a nested structure, e.g.:
```python
with SphericalScreen(...), propagate_with_solar_surface():
    with propagate_with_solar_surface():
        pass

    # Context tracker thinks that the first propagate_with_solar_surface() is no longer active
    # because the second propagate_with_solar_surface() has exited
    ...
```

This PR overhauls the internal tracking of context managers to use the data structure of a stack instead of an unordered set (which was peculiarly implemented as a dictionary of True/False values).  Now the tracking respects the nesting of contexts and can handle repeated contexts.  Plus, the contexts are stored as fully qualified names, which should avoid potential collisions.


ayshih:
While conceptually this can/should be backported to 6.0/6.1, it does change quite a bit with private behavior, so I'm fine with keeping it to 7.0.  The likelihood of this bug affecting users in the wild is low, and the consequence is simply that a warning is not being emitted when it should be.

nabobalis (approved):

sunpy/coordinates/screens.py
```diff
@@ -22,20 +22,21 @@ def __init__(self, only_off_disk=False):
...
     def __enter__(self):
-        ACTIVE_CONTEXTS[self._context_name] = True
+        ACTIVE_CONTEXTS.append(self._context_name)
         self._old_assumed_screen = Helioprojective._assumed_screen  # nominally None
         Helioprojective._assumed_screen = self
 
     def __exit__(self, exc_type, exc_value, exc_tb):
-        ACTIVE_CONTEXTS[self._context_name] = False
+        if (removed := ACTIVE_CONTEXTS.pop()) != self._context_name:
+            raise RuntimeError(f"Cannot remove {self._context_name} from tracking stack because {removed} is last active.")
```
nabobalis, line 39: Codecov says no test? I am not sure this is worth testing but I leave that up to you.

nabobalis (approved):

PR #8212, opened 2025-05-23
Enabled differential rotation of a screen

### LINKED ISSUE #7173: Fix `assume_spherical_screen()`+`propagate_with_solar_surface()`

ayshih:
Fixes #7173

This PR fixes a long-standing incompatibility between the context manager for applying rotation (`propagate_with_solar_surface()`) and the context managers for applying screen assumptions (`PlanarScreen()` and `SphericalScreen()`), which for example resulted in the discarding of most off-disk data in reprojections.  The incompatibility was because the distance to a differentially rotated screen has no straightforward solution and needs to be numerically solved.  This PR implements a relatively lightweight approach – typically takes at most ~8 linear-interpolation steps – to converge to the solution.

This approach works reasonably fast for reprojections (and also for autoalign plotting!).  However, when plotting gridlines on the differentially rotated screen, so many transformations are performed internally that this approach can take a surprisingly long time.  Of course, the upside is that the gridlines are plotted at all, which was not previously possible.

### DIFF
--- a/sunpy/coordinates/screens.py
+++ b/sunpy/coordinates/screens.py
@@ -2,14 +2,17 @@
 Screen class definitions for making assumptions about off-disk emission
 """
 import abc
+import logging
 
 import numpy as np
 
 import astropy.units as u
 from astropy.coordinates.representation import CartesianRepresentation, UnitSphericalRepresentation
 
-from sunpy.coordinates import HeliographicStonyhurst, Helioprojective
+from sunpy import log
+from sunpy.coordinates import HeliographicStonyhurst, Helioprojective, _transformations
 from sunpy.util.decorators import ACTIVE_CONTEXTS
+from sunpy.util.exceptions import warn_user
 
 __all__ = ['BaseScreen', 'SphericalScreen', 'PlanarScreen']
 
@@ -39,6 +42,75 @@ class BaseScreen(abc.ABC):
             raise RuntimeError(f"Cannot remove {self._context_name} from tracking stack because {removed} is last active.")
         Helioprojective._assumed_screen = self._old_assumed_screen
 
+    def _iterate_calculate_distance(self, coord, distance, screen_frame):
+        """
+        Numerically calculates the distance component to a screen to promote a
+        coordinate from 2D to 3D.
+
+        This method should be used only if the screen has been differentially rotated,
+        since otherwise promoting a coordinate from 2D to 3D is straightforward.
+
+        The approach used for numerical solving is to iterate to the solution starting
+        with an initial guess. For each guess, we construct a 3D coordinate and perform
+        a 3D->2D->3D transformation in the native frame of the screen. The amount that
+        the coordinate shifts in the 3D->2D->3D transformation is related to how far off
+        the guess is: when the shift is zero, the guess is correct. To form the next
+        guess, we linearly interpolate/extrapolate between the latest guess and the
+        closest guess over past iterations to estimate where the shift will be zero.
+
+        The maximum number of iterations is hard-coded to 20, but convergence will
+        typically take no more than 6 iterations. When working with an array coordinate,
+        elements that have converged are skipped in subsequent iterations.
+        """
+        # Check if the logging level is at least DEBUG (for performance reasons)
+        debug_output = log.getEffectiveLevel() <= logging.DEBUG
+
+        log.debug("Differentially rotating the screen")
+        use = np.ones(distance.shape, dtype=bool)
+        delta = np.empty_like(distance)
+        next_distance = np.empty_like(distance)
+        for niter in range(20):
+            # If this is not the first iteration, update best distance and calculate new distance
+            if niter > 0:
+                # Calculate for only those pixels that do not meet the tolerance
+                with np.errstate(invalid='ignore'):
+                    # A tolerance of 1e-11 is larger than numerical-precision errors (~1e-12), and only 1.5 meters at 1 AU
+                    use = np.abs(delta / distance) >= 1e-11*u.one
+                if np.sum(use) == 0:
+                    log.debug(f"Solved for the differentially rotated screen after {niter} iterations")
+                    break
+                # Squash small deltas to avoid warnings from later expressions
+                delta = np.where(use, delta, 0)
+
+                if niter == 1:
+                    # For the second iteration, use a simple guess
+                    best_distance, distance = distance, distance + delta
+                    best_delta = delta.copy()
+                else:
+                    # Starting with the third iteration, linearly interpolate between the latest distance and the best distance
+                    with np.errstate(invalid='ignore'):
+                        next_distance[use] = distance[use] - delta[use] * np.nan_to_num((best_distance[use] - distance[use]) / (best_delta[use] - delta[use]))
+
+                    # Update the best distance (not including the most recent calculation)
+                    best_distance = np.where(np.abs(best_delta) < np.abs(delta), best_distance, distance)
+                    best_delta = np.where(np.abs(best_delta) < np.abs(delta), best_delta, delta)
+                    distance = next_distance.copy()
+
+            # Calculate the corresponding delta of a 3D->2D->3D transformation in the native frame of the screen
+            # If delta is zero in that frame, then the 3D point is exactly on the screen, so the guessed distance is correct
+            other_3d = coord[use].realize_frame(coord[use].represent_as(UnitSphericalRepresentation) * distance[use])
+            native_3d = other_3d.transform_to(screen_frame)
+            native_2d = native_3d.realize_frame(native_3d.represent_as(UnitSphericalRepresentation))
+            delta[use] = self.calculate_distance(native_2d) - native_3d.spherical.distance
+
+            if debug_output:
+                log.debug(f"npoints={np.sum(use)}; max distance={distance.max()}, max delta={delta.max()}"
+                          + (f", best distance={best_distance.max()}, best delta={best_delta.max()}" if niter > 0 else ""))
+        else:
+            warn_user(f"Failed to solve for the differentially rotated screen after {niter+1} iterations")
+
+        return distance
+
 
 class SphericalScreen(BaseScreen):
     """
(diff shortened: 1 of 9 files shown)


nabobalis:
Figure test or example gallery worthy?

ayshih:
You can add gridlines to the differentially rotated screen, but it takes a while to compute them:
```python
...
fig = plt.figure(figsize=(10, 5))

ax1 = fig.add_subplot(121, projection=aia_map)
aia_map.plot(axes=ax1)
ax1.coords[0].set_ticks(spacing=3*u.arcmin)
ax1.coords[1].set_ticks(spacing=3*u.arcmin)
ax1.grid(ls='-', lw=1, color='red')

ax2 = fig.add_subplot(122, projection=out_warp)
out_warp.plot(axes=ax2, title=f"Differentially rotated by {(out_warp.date - aia_map.date).to(u.day)}")
overlay = ax2.get_coords_overlay(aia_map.coordinate_frame)
overlay[0].set_ticks(spacing=3*u.arcmin)
overlay[1].set_ticks(spacing=3*u.arcmin)
overlay.grid(ls='-', color='red')

with propagate_with_solar_surface(), SphericalScreen(aia_map.observer_coordinate, only_off_disk=True):
    plt.show()
```
Figure_1

Cadair:
I'm 👎 on backporting this fwiw.

nabobalis:
To 7.0?

Cadair:
Yes. Tho I am willing to be outvoted.

ayshih:
Re backporting to 7.0, I was hesitant at first thought, but I've convinced myself that backporting makes sense.  My thinking:

- There's no API change for inputs or outputs.  The only user-facing change is there will no longer be a warning when calling `reproject_to()` when the combination of context managers was being used (because the bad behavior has been fixed!). 
- There is no (real) disruptive impact on code in the wild.  Anyone already using the combination of context managers in the wild will simply get more accurate results than they did before.  If someone was counting on off-disk data to be discarded, they shouldn't have been using a screen anyway.  Also, the warning that is removed is triggered purely by user choices rather than on inputs, so it's not the type of warning that would be used for logic branching.
- It obviates the problem that the aforementioned warning covered only a fraction of the possible situations where the user could get bad results.  Rather than expanding the warnings at this stage, we should simply fix the bad behavior at the root.

I know it's a bit annoying to squeeze this in during the RC phase.  My feeling is that this PR should go into 7.0.x regardless, even if it has to be 7.0.1.  However, the benefit of getting it into 7.0 is that there can be a what's new entry for it.

nabobalis (approved):
Code looks great, don't understand the math.

ayshih:
Per the discussion at today's dev meeting, I have added a description of the solving approach to the internal docstring

sunpy/coordinates/screens.py
```diff
@@ -120,6 +196,12 @@ def calculate_distance(self, frame):
         # Ignore sqrt of NaNs
         with np.errstate(invalid='ignore'):
             distance = ((-1*b) + np.sqrt(b**2 - 4*c)) / 2  # use the "far" solution
+
+        # Iterate the calculation if differential rotation is being applied
+        if _transformations._autoapply_diffrot is not None and frame.obstime != self._center.obstime:
+            screen_frame = Helioprojective(observer=self._center, obstime=self._center.obstime)
+            distance = self._iterate_calculate_distance(frame, distance, screen_frame)
+
```
wtbarnes, line 204: it would be nice if this didn't have to be repeated for every screen subclass, but for two subclasses it's not that big of a deal.

sunpy/coordinates/screens.py
```diff
@@ -39,6 +42,73 @@ def __exit__(self, exc_type, exc_value, exc_tb):
...
+        log.debug("Differentially rotating the screen")
+        use = np.ones(distance.shape, dtype=bool)
+        delta = np.empty_like(distance)
+        next_distance = np.empty_like(distance)
+        for niter in range(20):
+            # If this is not the first iteration, update best distance and calculate new distance
+            if niter > 0:
+                # Calculate for only those pixels that do not meet the tolerance
+                with np.errstate(invalid='ignore'):
+                    use = np.abs(delta / distance) >= 1e-11*u.one
```
wtbarnes, line 77: Is there any particular rationale for using 1e-11?

sunpy/map/tests/test_reproject_to.py
```diff
@@ -236,3 +222,41 @@ def test_reproject_to_auto_extent_wcs(aia171_test_map, auto_extent, crpix, pixel
...
+        'cunit2': 'arcsec',
+        'ctype1': 'HPLN-TAN',
+        'ctype2': 'HPLT-TAN',
+        'hgln_obs': 0,
+        'hglt_obs': 30,
+        'rsun_ref': aia171_test_map.rsun_meters.value,
+        'dsun_obs': aia171_test_map.dsun.value,
+        'date-obs': new_time.isot,
+        'mjd-obs': new_time.mjd,
+    }
```
wtbarnes, line 250: Why not just use the header helper here?

sunpy/coordinates/screens.py
```diff
@@ -39,6 +42,73 @@ def __exit__(self, exc_type, exc_value, exc_tb):
...
+            other_3d = coord[use].realize_frame(coord[use].represent_as(UnitSphericalRepresentation) * distance[use])
+            native_3d = other_3d.transform_to(screen_frame)
+            native_2d = native_3d.realize_frame(native_3d.represent_as(UnitSphericalRepresentation))
+            delta[use] = self.calculate_distance(native_2d) - native_3d.spherical.distance
+
+            if debug_output:
+                log.debug(f"npoints={np.sum(use)}; max distance={distance.max()}, max delta={delta.max()}"
+                          + (f", best distance={best_distance.max()}, best delta={best_delta.max()}" if niter > 0 else ""))
+        else:
+            warn_user(f"Failed to solve for the differentially rotated screen after {niter+1} iterations")
```
wtbarnes, line 108: Would it be worth reporting the error when the max number of iterations was reached? If things fail, this would at least give the user an idea as to how bad things have gone.

sunpy/coordinates/screens.py
```diff
@@ -39,6 +42,73 @@ def __exit__(self, exc_type, exc_value, exc_tb):
...
+                    # Update the best distance (not including the most recent calculation)
+                    best_distance = np.where(np.abs(best_delta) < np.abs(delta), best_distance, distance)
+                    best_delta = np.where(np.abs(best_delta) < np.abs(delta), best_delta, delta)
+                    distance = next_distance.copy()
+
+            # Calculate corresponding new delta
+            other_3d = coord[use].realize_frame(coord[use].represent_as(UnitSphericalRepresentation) * distance[use])
+            native_3d = other_3d.transform_to(screen_frame)
+            native_2d = native_3d.realize_frame(native_3d.represent_as(UnitSphericalRepresentation))
+            delta[use] = self.calculate_distance(native_2d) - native_3d.spherical.distance
```
wtbarnes, line 102: During our weekly meeting you mentioned that this difference is correlated with the correct solution such that shrinking this delta to zero is equivalent to converging to the correct guess. Could you add an explanatory comment here that summarizes that?

wtbarnes (approved):
I think I've stared at this for long enough that I understand what is happening. The docstring on `_iterate_calculate_distance` is also really helpful. I've added a few (not very substantive) comments, but I'm happy with this as is.

sunpy/coordinates/screens.py
ayshih: Heh, yes, two reasons:

- It's roughly 1 meter at 1 AU (actually, 1.5 meters)
- `delta` is the subtraction of two `distance`-sized values, so I wouldn't want to go below 1e-12 due to numerical precision

sunpy/map/tests/test_reproject_to.py
ayshih: Certainly should for any public example, but for unit tests, I tend to avoid using the header helper when a simple header can be constructed by hand in order to minimize the "blast radius" of test failures if a bug crops up in the header helper

sunpy/coordinates/screens.py
ayshih: It's a good suggestion, but I'm hesitant to report a value here because I can't report it in the "right" frame...  A power user can turn on debug logging and sift through the crazy amount of messages.

sunpy/coordinates/screens.py
ayshih: Added

sunpy/coordinates/screens.py
```diff
@@ -39,6 +42,75 @@ def __exit__(self, exc_type, exc_value, exc_tb):
...
+
+            if debug_output:
+                log.debug(f"npoints={np.sum(use)}; max distance={distance.max()}, max delta={delta.max()}"
+                          + (f", best distance={best_distance.max()}, best delta={best_delta.max()}" if niter > 0 else ""))
+        else:
+            warn_user(f"Failed to solve for the differentially rotated screen after {niter+1} iterations")
```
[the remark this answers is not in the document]
ayshih: Yes, and added a sentence.  In practice, I haven't seen this warning emitted (without dropping the max iterations to a small number).

sunpy/coordinates/screens.py
```diff
@@ -39,6 +42,75 @@ def __exit__(self, exc_type, exc_value, exc_tb):
...
+            # Calculate the corresponding delta of a 3D->2D->3D transformation in the native frame of the screen
+            # If delta is zero in that frame, then the 3D point is exactly on the screen, so the guessed distance is correct
+            other_3d = coord[use].realize_frame(coord[use].represent_as(UnitSphericalRepresentation) * distance[use])
+            native_3d = other_3d.transform_to(screen_frame)
+            native_2d = native_3d.realize_frame(native_3d.represent_as(UnitSphericalRepresentation))
+            delta[use] = self.calculate_distance(native_2d) - native_3d.spherical.distance
+
+            if debug_output:
+                log.debug(f"npoints={np.sum(use)}; max distance={distance.max()}, max delta={delta.max()}"
+                          + (f", best distance={best_distance.max()}, best delta={best_delta.max()}" if niter > 0 else ""))
```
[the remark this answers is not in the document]
ayshih: Now done

PR #8213, opened 2025-05-27
Use new extension helpers support for limited API

Cadair:
Needs https://github.com/astropy/extension-helpers/pull/26

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ requires = [
   "setuptools>=62.1",
   "setuptools_scm[toml]>=8.0.1",
   "wheel",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -3,7 +3,7 @@ requires = [
   "setuptools>=62.1",
   "setuptools_scm[toml]>=8.0.1",
   "wheel",
-  "extension-helpers",
+  "extension-helpers>=1.3.0",
```
astrofrog, line 6: I'd recommend adding <2 as well as in general we recommend pinning to a major version

PR #8225, opened 2025-05-31
Fixed handling of array obstime for a differentially rotated screen

### LINKED ISSUE #8224: Combining screen with `propagate_with_solar_surface` does not work for coordinates with more than one obstime
#8212 added the ability to differentially rotate a screen. Combining `SphericalScreen` with `propagate_with_solar_surface` does not work when transforming coordinates with more than one obstime, e.g. if you have a multidimensional coordinate that has a varying observer coordinate.

```python
from sunpy.coordinates import Helioprojective, get_earth, SphericalScreen, propagate_with_solar_surface

frame = Helioprojective(observer=get_earth(time=['2020-01-01','2020-01-02']))
coord = SkyCoord(0,0,unit='arcsec',frame=frame)
with propagate_with_solar_surface(), SphericalScreen(coord[0].observer):
    coord.make_3d()
```

gives

```python-traceback
Cell In[29], line 6
      4 coord = SkyCoord(0,0,unit='arcsec',frame=frame)
      5 with propagate_with_solar_surface(), SphericalScreen(coord[0].observer):
----> 6     coord.make_3d()

File [~/Documents/codes/sunpy/sunpy/coordinates/frames.py:572](http://localhost:8888/lab/tree/Documents/projects/active-region-evolution-study/notebooks/~/Documents/codes/sunpy/sunpy/coordinates/frames.py#line=571), in Helioprojective.make_3d(self)
    569     d = ((-1*b) - np.sqrt(b**2 - 4*c)) [/](http://localhost:8888/) 2  # use the "near" solution
    571 if self._assumed_screen:
--> 572     d_screen = self._assumed_screen.calculate_distance(self)
    573     d = np.fmin(d, d_screen) if self._assumed_screen.only_off_disk else d_screen
    575 # This warning can be triggered in specific draw calls when plt.show() is called
    576 # we can not easily prevent this, so we check the specific function is being called
    577 # within the stack trace.

File [~/Documents/codes/sunpy/sunpy/coordinates/screens.py:206](http://localhost:8888/lab/tree/Documents/projects/active-region-evolution-study/notebooks/~/Documents/codes/sunpy/sunpy/coordinates/screens.py#line=205), in SphericalScreen.calculate_distance(self, frame)
    203     distance = ((-1*b) + np.sqrt(b**2 - 4*c)) [/](http://localhost:8888/) 2  # use the "far" solution
    205 # Iterate the calculation if differential rotation is being applied
--> 206 if _transformations._autoapply_diffrot is not None and frame.obstime != self._center.obstime:
    207     screen_frame = Helioprojective(observer=self._center, obstime=self._center.obstime)
    208     distance = self._iterate_calculate_distance(frame, distance, screen_frame)

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

ayshih:
Fixes #8224 

I forgot to use an `np.any()` to handle array obstime.

The reason for the other code changes is because support for array obstime means that it's possible to sneak in a point with zero differential rotation, and that point converges on the very first iteration.  That conflicted with my skipping of an array initialization under an implicit assumption that no point would converge immediately.  Anyway, I rearranged the logic so that it is now safe for immediate convergence.

I don't think this needs a separate changelog entry from #8212.

### DIFF
--- a/sunpy/coordinates/screens.py
+++ b/sunpy/coordinates/screens.py
@@ -68,7 +68,6 @@ class BaseScreen(abc.ABC):
         log.debug("Differentially rotating the screen")
         use = np.ones(distance.shape, dtype=bool)
         delta = np.empty_like(distance)
-        next_distance = np.empty_like(distance)
         for niter in range(20):
             # If this is not the first iteration, update best distance and calculate new distance
             if niter > 0:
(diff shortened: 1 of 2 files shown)


wtbarnes:
Is there a test that covers the case where there's more than one obstime?

PR #8234, opened 2025-06-09
Add tests for ASDF serialization of custom times

braingram:

## 🚨 IMPORTANT 🚨

Thank you.

- Be wary of posting output from Large Language Models or similar generative AI as comments on GitHub or any other platform, as such comments tend to be formulaic and low quality content.
  If you use generative AI tools as an aid in developing code or documentation changes, ensure that you fully understand the proposed changes and can explain why they are the correct approach and an improvement to the current state.


braingram:
I don't expect this PR needs a changelog entry. Let me know if I should add one. Otherwise I don't have permissions to apply the corresponding label.

docs segfaulted but are also segfaulting on main: https://app.readthedocs.org/projects/sunpy/builds/28479071/

braingram:
@wtbarnes would you give this a whirl? With the latest asdf-astropy (0.8.0) the custom times should save again. This PR is sort of "belt-and-suspenders" to make sure any future asdf changes (which include sunpy tests as part of downstream) will test against these time formats. If testing in an existing environment you'll also need asdf-standard 1.3.0 for the updated time schema.

sunpy/io/special/asdf/tests/test_time.py
```diff
@@ -0,0 +1,24 @@
...
+
+from asdf.testing.helpers import roundtrip_object
+from astropy.time import Time
+from astropy.utils import minversion
+
+# This registers TimeUTime in astropy.time.Time
+from sunpy.time import TimeUTime  # NOQA
+
+if not minversion("asdf_astropy", "0.8.0"):
+    pytest.skip("these tests require asdf_astropy >= 0.8.0", allow_module_level=True)
```
Cadair: I think this should work and is a bit easier?

```suggestion
# This registers TimeUTime in astropy.time.Time
from sunpy.time import TimeUTime  # NOQA

pytest.importorskip("asdf_astropy", "0.8.0")
```

Cadair (approved):

Cadair:
Thanks a lot @braingram :rocket:

sunpy/io/special/asdf/tests/test_time.py
braingram: Thanks! I didn't know about that option :)

Cadair: It returns the module, so `asdf_astropy = pytest.importorskip(...)` if you actually want to use it later in the code.

PR #8257, opened 2025-06-12
Protected against a bad interaction between astropy's C fast parser of time strings and numpy>=2.3

ayshih:
Even if #8251 didn't look awful, I think we need to add internal code to protect against the segfaults of astropy/astropy#18254.

Even after astropy/astropy#18265 is merged and gets into releases, I think the user experience of segfaults is so bad that this protection should not be removed for a long while.  We'll just refine it with a version check of astropy.

### DIFF
--- a/changelog/8257.bugfix.rst
+++ b/changelog/8257.bugfix.rst
@@ -1,0 +1,1 @@
+Prevented a potential segmentation fault when calling :func:`~sunpy.time.parse_time` on a list of more than 500 strings in a format not specifically recognized by `sunpy.time` itself and instead has to fall back on recognition by `astropy.time`.
--- a/sunpy/time/tests/test_time.py
+++ b/sunpy/time/tests/test_time.py
@@ -356,6 +356,13 @@ def test_parse_time_list_3():
     assert np.all(parse_time(tstrings) == Time.strptime(tstrings, '%Y-%b-%d'))
 
 
+def test_parse_time_long_list():
+    # Check whether lists of >500 strings cause a segmentation fault
+    # https://github.com/astropy/astropy/issues/18254
+    assert isinstance(parse_time(['2001-Jan-01']*501), Time)  # sunpy format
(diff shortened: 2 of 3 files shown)


examples/time_series/goes_xrs_nrt_data.py
```diff
@@ -44,7 +43,7 @@
 # `sunpy.timeseries.TimeSeries` requires a datetime index, which we can get by
 # parsing the time strings.
 
-goes_data.index = [Time(t).datetime for t in goes_data.index]
+goes_data.index = parse_time(list(goes_data.index)).datetime
```
Cadair, line 46: Do we actually need the `list(` here? can't parse_time accept a pandas column?

ayshih: It's because our dispatch has `list` registered, but doesn't know what to do with a pandas column

Cadair: huh, I really thought we had support for that in parse_time :thinking:

nabobalis: Feature request issue time

alasdairwilson (commented):
Should add test(s) for parse time with these lists of strings so that a gallery example isn't the only place we can catch this.

ayshih:
Test added and docs build successfully.  I'm undecided whether to add a changelog entry for this.

Cadair:
> Test added and docs build successfully. I'm undecided whether to add a changelog entry for this.

I would just on the off chance it catches someone by surprise.

ayshih:
Changelog entry added

examples/time_series/goes_xrs_nrt_data.py
ayshih: Now fixed by #8356.  `pandas.Index` was not supported, even though `pandas.Series` and `pandas.DatetimeIndex` were supported.

PR #8258, opened 2025-06-12
Removed the vestigial notion that `reproject` is an optional dependency

ayshih:
`reproject` has been a dependency of `sunpy.map` for nearly 4 years now (see #5470).  It's way overdue to remove the docstring note and the `ImportError` check in `reproject_to()`.


nabobalis (approved):

Cadair:
Haha, the irony that the NDCube migration will add that import check back.

PR #8265, opened 2025-06-13
Added the `%Y-%m-%dT%H:%M:%SZ` time string format

ayshih:
We currently directly support `%Y-%m-%dT%H:%M:%S.%fZ`, but we don't directly support `%Y-%m-%dT%H:%M:%SZ` (without the fractional second), which means that parsing that format falls through to `astropy.time` parsing.  This PR adds that format, plus it also adds a log debug message so that it can be easily seen when we aren't directly supporting a format.


nabobalis (approved):

nabobalis:
Changelog?

nabobalis (approved):

PR #8271, opened 2025-06-13
remove xdist as hard test dep

nabobalis:


Cadair (approved):
Assuming the tests pass

PR #8308, opened 2025-07-25
Use string for JSOC email in cutouts example instead of os var

TimMonko:
## PR Description

Running locally on sunpy==7.0.0 on Windows I had to change to use a string for my e-mail, which is also how the comment above says it needs to be done.  This updates the example to be easier to work with. 
Note, this is my first time ever using sunpy so maybe I did something wrong 🙃 🎉

### DIFF
--- a/examples/acquiring_data/downloading_cutouts.py
+++ b/examples/acquiring_data/downloading_cutouts.py
@@ -6,8 +6,6 @@ Requesting cutouts of AIA images from the JSOC
 This example shows how to request a cutout of a series of
 AIA images from the JSOC.
 """
-import os
-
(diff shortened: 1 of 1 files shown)


nabobalis:
Hello @TimMonko, the os var is set up for the examples to run on our continuous integration. 

For a user, they need to register their own email address with the JSOC and then replace that line with their own email instead. The JSOC requires an email address to acquire data.

TimMonko:
Sounds great! Thanks for your kindness! 👍 
I did figure out the string part, but did not realize the os.environ was for CI
I think the comment about it in the example is enough

PR #8327, opened 2025-08-18
use new pytest asdf plugin

braingram:
## PR Description

Update test dependencies to use new https://pypi.org/project/pytest-asdf-plugin/ which is a feature-compatible replacement for the bundled asdf pytest plugin (which will soon be deprecated).

For context the pytest-asdf-plugin is responsible for finding and converting schema files into "tests" (to check the schema against the metaschema and validated any contained "examples"). Since both the bundled and now external plugins are feature-compatible there should be no change in test number (or new failures, etc).

On main 2493 tests passed:
https://github.com/sunpy/sunpy/actions/runs/17033516101/job/48280847581#step:10:5692

With this PR 2493 tests passed:
https://github.com/sunpy/sunpy/actions/runs/17048240280/job/48329382329?pr=8327#step:10:5693

However I cannot find a CI run that runs the schema checks (if all runs use `--pyargs` they are possibly all ignoring the schemas: https://github.com/asdf-format/pytest-asdf-plugin/issues/8).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,6 +113,7 @@ tests-only = [
   "hvpy>=1.1.0",
   "jplephem>=2.19",
   "pytest-astropy>=0.11.0",
+  "pytest-asdf-plugin>=0.1.1",
   "pytest-mpl>=0.17.0",
   "pytest>=7.4.0",
 ]


braingram:
I'll see if I can relax the pytest pin to fix the oldestdeps. What's the benefit in testing with an old pytest?

braingram:
Closing and re-opening to retrigger CI.

nabobalis:
> I'll see if I can relax the pytest pin to fix the oldestdeps. What's the benefit in testing with an old pytest?

We try to track the min version of pytest used by debian, sunpy is packaged for debian and we don't want to break that testing. However, I have not checked what version of pytest they use these days, if it's possible it would be nice to bump it to 8.0 min.

braingram:
> > I'll see if I can relax the pytest pin to fix the oldestdeps. What's the benefit in testing with an old pytest?
> 
> We try to track the min version of pytest used by debian, sunpy is packaged for debian and we don't want to break that testing. However, I have not checked what version of pytest they use these days, if it's possible it would be nice to bump it to 8.0 min.

Thanks! That is great to know. I relaxed the pin in `pytest-asdf-plugin` to allow 7.x versions (in the 0.1.1 version of `pytest-asdf-plugin`) so hopefully the oldestdeps will pass (once the CI picks up the new version).

braingram:
Closing and reopening again because uv didn't see the new version.

nabobalis:
I just saw your comment, all the runs use pyargs to run the unit tests. Will that be a problem?

braingram:
> I just saw your comment, all the runs use pyargs to run the unit tests. Will that be a problem?

Thanks. I think it's always been a problem :) Within the scope of this PR there's no new problems with `--pyargs` usage and the new `pytest-asdf-plugin` but it also doesn't fix the issue where no schema tests are being run. For fixing that we could either:
- change the sunpy CI to run tests with a predictable path (which `--pyargs` generally don't provide)
- change `pytest-asdf-plugin` to allow for unpredictable paths. The exact fix for that isn't yet decided on but see https://github.com/asdf-format/pytest-asdf-plugin/issues/8 for one idea.

braingram:
docs gallery and py312-online look to be failing the same way on main:
https://github.com/sunpy/sunpy/actions/runs/17033516101

nabobalis:
Thanks again @braingram

PR #8332, opened 2025-08-27
Apply standard plot formatting on draw_contours

### LINKED ISSUE #8331: `Map.draw_contours()` does not apply our standard plot formatting
### Describe the bug

`GenericMap.draw_contours()` does not apply our standard plot formatting (compared to `GenericMap.plot()`), which is important when the plot is only contours rather than contours overlaid on another map.  The plot has no title, and the axes do not have the nice names.  Furthermore, the aspect ratio does not nicely default to equal aspect.

Encountered by @jessiemcd

### To Reproduce

The left panel shows our standard plot formatting.  The right panel shows plotting only contours.
```python

import astropy.units as u

import sunpy.map
from sunpy.data.sample import HMI_LOS_IMAGE

hmi = sunpy.map.Map(HMI_LOS_IMAGE)

fig = plt.figure(figsize=(12, 4))

ax1 = fig.add_subplot(121, projection=hmi)
hmi.plot(axes=ax1)

ax2 = fig.add_subplot(122, projection=hmi)
hmi.draw_contours([-100, 100] * u.gauss, axes=ax2)

plt.show()
```

### Screenshots

Image

### System Details

_No response_

### Installation method

_No response_

bhavyaKhatri2703:
## PR Description

This PR fixes the lack of standard plot formatting in GenericMap.draw_contours().

Adds an annotate keyword (default True) and Title field to draw_contours() so that axis labels, titles, and aspect ratio are applied consistently, similar to plot().

Figure_1

Fixes #8331


sunpy/map/mapbase.py
```diff
@@ -2664,6 +2668,17 @@ def draw_contours(self, levels, axes=None, *, fill=False, **contour_args):
...
+
+        if annotate:
+            axes.set_xlabel(
+                axis_labels_from_ctype(self.coordinate_system[0], self.spatial_units[0])
+            )
+            axes.set_ylabel(
+                axis_labels_from_ctype(self.coordinate_system[1], self.spatial_units[1])
+            )
+            axes.set_title(title)
+            axes.set_aspect("equal", adjustable="datalim")
```
nabobalis: How does this look for non-equal data?

sunpy/map/mapbase.py
```diff
@@ -2612,7 +2612,10 @@ def draw_contours(self, levels, axes=None, *, fill=False, **contour_args):
             Determines the style of the contours:
             - If `False` (default), contours are drawn as lines using :meth:`~matplotlib.axes.Axes.contour`.
             - If `True`, contours are drawn as filled regions using :meth:`~matplotlib.axes.Axes.contourf`.
-
+        annotate : `bool` , optional
+            -If `True` , apply standard map formatting
+        title : `str`
+            Title of the contour map.
```
nabobalis: Do we need to add these? 

Could we not check if there is an existing axis and if not, label it automatically? Also with the title, that can be set after the plot has been created, I don't know if I see the need for that as a new keyword input.

nabobalis:
Thanks for the PR @bhavyaKhatri2703, I have asked Albert to take a deeper look into the PR.

For now I just have a few comments and the need for us to add a few test cases to ensure this is working as intended. See https://docs.sunpy.org/en/latest/dev_guide/contents/tests.html#figure-unit-tests for some guidance on how to add them. 

If you get stuck, please do ask.

sunpy/map/mapbase.py
bhavyaKhatri2703: yes we can do that. Remove annotate and title parameters , and apply labels if no existing axes are provided.
pls let me know if i should make those changes

ayshih: > Could we not check if there is an existing axis and if not, label it automatically?

An existing axis can need to be formatted as well.  For example, creating multiple subplots in a figure means that each axis already exists, but is unformatted.

The alternative to a control keyword would be to do some form of automatic detection to distinguish between a base plotting and an overlay plotting.  That is, check the existing title and axis labels to see if they appear to be unformatted...

ayshih: We do not need a `title` keyword argument at all.  `annotate=True` should set the standard title derived from the map.

ayshih: > The alternative to a control keyword would be to do some form of automatic detection to distinguish between a base plotting and an overlay plotting.

Hmm, we actually ought to do something like this for `plot()` as well...

ayshih:
Thanks for creating this PR!  Some of your code has issues, but it's moot because I think we should implement the better handling of formatting I suggested above.  Here's a branch showing my draft implementation: https://github.com/sunpy/sunpy/compare/main...ayshih:sunpy:plot_formatting

Let me know if you want to incorporate my code changes into your PR – and then you can do the important stuff of adding corresponding figure tests – or if you instead want me to take over your PR.

bhavyaKhatri2703:
> Thanks for creating this PR! Some of your code has issues, but it's moot because I think we should implement the better handling of formatting I suggested above. Here's a branch showing my draft implementation: [main...ayshih:sunpy:plot_formatting](https://github.com/sunpy/sunpy/compare/main...ayshih:sunpy:plot_formatting)
> 
> Let me know if you want to incorporate my code changes into your PR – and then you can do the important stuff of adding corresponding figure tests – or if you instead want me to take over your PR.

Thanks for reviewing my PR! Can i merge your changes into my branch? and I'll try to work on adding the figure tests.

ayshih:
> Can i merge your changes into my branch? and I'll try to work on adding the figure tests.

Yes, please go ahead!

I did *not* do anything about the aspect ratio.  Upon further thought, it may not make sense to impose a default aspect ratio for `draw_*` methods, so I think that should be deferred to a potential future PR.

bhavyaKhatri2703:
> > Can i merge your changes into my branch? and I'll try to work on adding the figure tests.
> 
> Yes, please go ahead!
> 
> I did *not* do anything about the aspect ratio.  Upon further thought, it may not make sense to impose a default aspect ratio for `draw_*` methods, so I think that should be deferred to a potential future PR.

just a question do I need to add test functions for every draw_* function for checking auto formatting?

ayshih:
Rather than writing a separate figure test for each method, you can probably set it up as a parametrized figure test, where the values for the parameter are the methods to be tested.

bhavyaKhatri2703:
@ayshih can u please review this pr now?

nabobalis:
> @ayshih can u please review this pr now?

Please do not @ people for reviews. Albert is aware of this PR and will review it when he gets a chance.

ayshih (changes requested):
You're on the right track, but you need to change the parametrized figure test because you aren't actually testing the various methods as intended.

You also need to add a changelog entry.

sunpy/map/tests/test_plotting.py
```diff
@@ -76,6 +76,7 @@ def test_peek_grid_aia171(aia171_test_map):
     aia171_test_map.peek(draw_grid=True)
 
 
+
```
ayshih, line 79:
```suggestion
```
Revert this spurious newline

sunpy/map/tests/test_plotting.py
```diff
@@ -474,3 +475,23 @@ def test_plot_unit8(aia171_test_map):
...
+        "bottom_left": (50, 50) * u.pix,
+        "width": 30 * u.pix,
+        "height": 50 * u.pix
+    }),
+    ("draw_contours", {"levels": u.Quantity(np.arange(1, 100, 10), 'percent')}),
+])
+def test_auto_format_plot_methods(aia171_test_map, method, kwargs):
+    fig = Figure()
+    ax = fig.add_subplot(projection=aia171_test_map)
+    aia171_test_map.plot(axes=ax)
```
ayshih, line 495:
```suggestion
```
You need to take this line out, because as written you are testing the autoformatting of only plot() because that is the first method called

sunpy/tests/figure_hashes_mpl_382_ft_261_astropy_702_animators_121.json
```diff
@@ -90,4 +95,4 @@
   "sunpy.visualization.tests.test_drawing.test_draw_extent": "d9e5182e20feff6257c3209c2202727a8e628842ca3c0c80fd1cafbc64376388",
   "sunpy.visualization.tests.test_drawing.test_draw_extent_3d": "1d329ecd0e31f4d4ecfb3b75e255c117dc631bd39be9a8a862fb7974004ff07c",
   "sunpy.visualization.tests.test_visualization.test_show_hpr_impact_angle": "43f6c3fea4238064ce503d699cba6457b3eb07ee58ea2a52f43556f42ee7efc8"
-}
+}
```
ayshih: Revert this change

ayshih (changes requested):

bhavyaKhatri2703:
so i am struggling with passing the test for the draw_grid method it keeps giving me the warning :

"The conversion of these 2D helioprojective coordinates to 3D is all NaNs because off-disk coordinates need an additional assumption to be mapped to calculate distance from the observer. Consider using the context manager `SphericalScreen()`"

nabobalis:
I am not sure if https://docs.sunpy.org/en/stable/generated/gallery/plotting/offdisk_contours.html will be helpful reading.

Issue #8335, opened 2025-09-02
[7.0.1] Compilation error in sunpy/io/src/ana/_pyana.c


olebole:
### Describe the bug

I am working to package sunpy 7.0.1 for Debian. When building from source, I get the following error, which I do not fully understand:
```
running build_ext
building 'sunpy.io._pyana' extension
creating build/temp.linux-x86_64-cpython-313/sunpy/io/src/ana
x86_64-linux-gnu-gcc -g -O2 -Werror=implicit-function-declaration -ffile-prefix-map=/build/sunpy-7.0.1=. -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -I/usr/lib/x86_64-linux-gnu/python3-numpy/numpy/_core/include -I/usr/include/python3.13 -c sunpy/io/src/ana/_pyana.c -o build/temp.linux-x86_64-cpython-313/sunpy/io/src/ana/_pyana.o -std=c99 -O3 -w
In file included from /usr/include/python3.13/object.h:1232,
                 from /usr/include/python3.13/Python.h:73,
                 from sunpy/io/src/ana/_pyana.c:14:
/usr/include/python3.13/cpython/object.h:147:8: error: redefinition of 'struct _typeobject'
  147 | struct _typeobject {
      |        ^~~~~~~~~~~
sunpy/io/src/ana/_pyana.c:10:8: note: originally defined here
   10 | struct _typeobject {
      |        ^~~~~~~~~~~
/usr/include/python3.13/object.h: In function 'PyType_HasFeature':
/usr/include/python3.13/object.h:1248:21: error: 'PyTypeObject' {aka 'struct _typeobject'} has no member named 'tp_flags'
 1248 |         flags = type->tp_flags;
      |                     ^~
In file included from /usr/include/python3.13/Python.h:64:
/usr/include/python3.13/cpython/objimpl.h: In function '_PyObject_SIZE':
/usr/include/python3.13/cpython/objimpl.h:6:40: error: 'PyTypeObject' {aka 'struct _typeobject'} has no member named 'tp_basicsize'
    6 |     return _Py_STATIC_CAST(size_t, type->tp_basicsize);
      |                                        ^~
/usr/include/python3.13/pyport.h:16:47: note: in definition of macro '_Py_STATIC_CAST'
   16 | #  define _Py_STATIC_CAST(type, expr) ((type)(expr))
[...]

PR #8336, opened 2025-09-02
Remove old ANA compat block

nabobalis:
Fixes https://github.com/sunpy/sunpy/issues/8335

### DIFF
--- a/sunpy/io/src/ana/_pyana.c
+++ b/sunpy/io/src/ana/_pyana.c
@@ -6,11 +6,6 @@ Based on Michiel van Noort's IDL DLM library 'f0' which contains
 a cleaned up version of the original anarw routines.
 */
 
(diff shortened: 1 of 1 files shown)


wtbarnes:
Why do we no longer need this block of code?

nabobalis:
Causing a crash in Debian and we can remove it without harm.

Cadair:
I wonder what version of gcc the old manylinux build images are using. We definitely need to test this on all the wheels ~and maybe even conda-forge as well~ conda-forge seems to be using gcc 15. The first release of gcc 10.1 was 2020, so old but not silly.

Cadair:
If the wheels build, then I don't see a problem removing this.

Cadair (approved):

PR #8339, opened 2025-09-03
Updated time-distance example

nabobalis:
See https://github.com/sunpy/sunpy/pull/8057#issuecomment-3250413156

### DIFF
--- a/examples/plotting/time_distance.py
+++ b/examples/plotting/time_distance.py
@@ -140,28 +140,31 @@ extent = [aia_seq[0].date.datetime, aia_seq[-1].date.datetime,
 # Plot the reference submap, line and extracted data from both methods and
 # the difference between them.
 
-fig = plt.figure(figsize=(10, 10), layout="constrained")
+fig = plt.figure(figsize=(10, 5), layout="constrained")
 left, right = fig.subfigures(nrows=1, ncols=2, width_ratios=[0.6, 0.75])
 left_ax = left.add_subplot(111, projection=reprojected_sub_maps[0])
 right_ax = right.subplot_mosaic([['repro'], ['trans'], ['diff']],
                                 sharex=True, sharey=True)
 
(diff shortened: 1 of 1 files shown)


examples/plotting/time_distance.py
```diff
@@ -139,7 +140,7 @@
 # Plot the reference submap, line and extracted data from both methods and
 # the difference between them.
 
-fig = plt.figure(figsize=(8, 5), layout="constrained")
+fig = plt.figure(figsize=(10, 10), layout="constrained")
```
nabobalis: Not sure how I feel about this change.

examples/plotting/time_distance.py
```diff
@@ -148,12 +149,19 @@
...
+)
+right_ax['trans'].imshow(
+    intensities_transform.T, aspect='auto', interpolation='none',
+    extent=extent, cmap=imag_ax.get_cmap()
+)
+right_ax['diff'].imshow(
+    (intensities_reproject-intensities_transform).T, interpolation='none',
+    aspect='auto', extent=extent, cmap='bwr'
+)
+plt.colorbar(right_ax['diff'].images[0], ax=right_ax['diff'], label='Difference')
```
Cadair: I think maybe the main image also needs a colour bar, otherwise you don't have a reference for the scale on the difference image. Like is the min/max of the main image 0<>10 if so the +-5 on the difference is a lot etc.

examples/plotting/time_distance.py
Cadair: yeah it's not as neat as it was before, maybe just make it slightly wider than it was before to accommodate the colour bars?

nabobalis: Changed.

examples/plotting/time_distance.py
nabobalis: Added

nabobalis:
SOMEONE APPROVE IT

PR #8322, opened 2025-08-11
Add earth to scale example

nabobalis:
~This does add a pooch dep which I think I am fine with.~

TODO:
- [x] Changelog
- [x] Pooch? POOOCH? POOOOOOOCH?

PR #8340, opened 2025-09-04
Updates to the Earth-scale example

ayshih:
This PR modifies the Earth-scale example (#8322):

* A low-resolution Earth image has been added to the repo (`sunpy.data.EARTH_IMAGE`) so that it does not need to be downloaded for this example.
* The Earth is now plotted using mesh-based plotting instead of image-based plotting.  Depending on the WCS of the underlying map, image-based plotting can sometimes exhibit artifacts that necessitate the use of a clip path, and for some WCSs, even that is not enough.  Mesh-based plotting is the most robust approach.
* The Earth was upside-down (for those of us on the northern hemisphere).


examples/plotting/adding_earth.py
```diff
@@ -52,21 +52,23 @@
 earth_center = SkyCoord(earth_x, earth_y, frame=cutout_map.coordinate_frame)
 earth_wcs = sunpy.map.make_fitswcs_header(
     (earth.height, earth.width), earth_center,
-    scale=earth_diameter / ([earth.width, earth.height] * u.pix)
+    scale=earth_diameter / ([earth.width, -earth.height] * u.pix)
```
Cadair: Is it worth commenting as to why this is negative?

ayshih: Probably =)

ayshih: It felt too awkward to explain the negative scale, so instead I chose to the Earth image upside-down up front and explain it there.

Issue #8344, opened 2025-09-12
JSOC `Cutout` attribute does not properly handle coordinate input


ayshih:
Our JSOC `Cutout` attribute doesn't transform the input coordinate all, but instead implicitly assumes that the coordinate is already in a `Helioprojective` frame with SDO as the observer:

https://github.com/sunpy/sunpy/blob/21cd5b062dd65fa5ad0b86bcc3ce1718bea887ba/sunpy/net/jsoc/attrs.py#L160-L173

That is, supplying a non-`Helioprojective` coordinate will emit a non-helpful error, and supplying a `Helioprojective` coordinate will simply ignore whatever observer is set for that coordinate.

`bl` and `tr` should be transformed to a `Helioprojective` frame with Earth as the observer, and the approximation of Earth for SDO's location should be stated in the docstring.

Cadair:
Bonus points for using spice to get sdo? 🤪

ayshih:
Or JPL Horizons, and avoid downloading a SPICE kernel.

Of course, an unfortunate consequence of having an internal coordinate transformation is that a screen will be needed if the cutout corners go beyond the limb.

wtbarnes:
Whoops 🙈

wtbarnes:
I cannot believe it has taken this long for anyone to discover this. This feature is 4 years old.

Cadair:
> an unfortunate consequence of having an internal coordinate transformation is that a screen will be needed if the cutout corners go beyond the limb.

This is particularly upsetting if we do use Earth and someone uses an aiamap frame for the input.

wtbarnes:
This is only tangentially related to this problem, but I want to document it here, at least for now.

@ayshih was describing to me some strange behavior of the cutout service when tracking was turned on and when the coordinates of the cutout were off limb and in particular were outside of the FoV of AIA. It would be good to have an example of this particular mode of failure.

If we can figure out particular sets of coordinates for which the JSOC cutout API behaves unexpectedly, we could protect against this on our side, either in `drms` or in the `Cutout` attr.

Relevant links:
- Source code for `im_patch` processing option in DRMS, http://jsoc.stanford.edu/doxygen_html/im__patch_8c-source.html
- API documentation for `im_patch` processing option in DRMS, http://jsoc.stanford.edu/doxygen_html/group__im__patch.html

ayshih:
[...]

PR #4595, opened 2020-10-21
Add attr for requesting cutouts from jsoc

wtbarnes:
Fixes #1259 

Currently, this PR also relies on sunpy/drms#43 and should not be merged until a new version of drms is released that includes that PR.

- [x] Add example usage in `JSOCClient` docs or example gallery
- [x] Attr docstring
- [x] Tests???
- [x] Changelog

Documentation for the `im_patch`: http://jsoc.stanford.edu/doxygen_html/group__im__patch.html

Example usage:

```python
import astropy.time
import astropy.units as u
from astropy.coordinates import SkyCoord

from sunpy.net import jsoc
from sunpy.net import attrs
from sunpy.coordinates import Helioprojective

from aiapy.util import sdo_location

tstart = astropy.time.Time('2015-10-17T04:33:30.000')
duration = 1 * u.minute

obs = sdo_location(tstart)
frame = Helioprojective(observer=obs, obstime=obs.obstime)

width = 345.6 * u.arcsec
height = 345.6 * u.arcsec
Txc = -517.2 * u.arcsec
Tyc = -246 * u.arcsec

bottom_left = SkyCoord(Tx=Txc-width/2, Ty=Tyc-height/2, frame=frame)

c = jsoc.JSOCClient()

q = c.search(
    attrs.Time(tstart, tstart + duration),
    attrs.Wavelength(171*u.angstrom),
    attrs.Sample(12*u.s),
    jsoc.attrs.Series('aia.lev1_euv_12s'),
    jsoc.attrs.Notify('first.last@email.com'),
    jsoc.attrs.Segment('image'),
    jsoc.attrs.Cutout(bottom_left, width, height, tracking=True),
)

req = c.request_data(q)
resp = c.get_request(req)
```

PR #8346, opened 2025-09-16
Added input validation for the JSOC `Cutout` attribute

ayshih:
This PR fixes two issues with the JSOC `Cutout` attribute (added in #4595):

- Verifies that the input is a helioprojective coordinate and explicitly documents that the observer frame attribute of the coordinate is ignored.  (This PR does *not* perform any coordinate transformation, so this PR does not close issue #8344.)
- Verifies that the center of the cutout is on the solar disk if tracking is enabled (default setting).  See https://github.com/sunpy/sunpy/issues/8344#issuecomment-3292990861


wtbarnes (approved):
We should have a tracking issue to actual deal with the possibility of non-HPC coordinates as inputs.

nabobalis (approved):

PR #8348, opened 2025-09-17
Fixed doctest that depended on a name

ayshih:
We have a failing doctest because the PI for STEREO/IMPACT changed.  It seems prudent to simply replace the name with an ellipsis since we don't particularly care about whose name is returned.


ayshih:
This PR should be backported, so I added a `trivial` changelog entry

PR #8356, opened 2025-09-18
Fixed parsing of SunPy-supported time formats in common not-quite-lists

ayshih:
This PR fixes the parsing of SunPy-supported time formats in two common not-quite-lists: `pandas.Series` and `astropy.table.Column`.  Previously, they are passed directly to `Time()`, so our time formats were not supported.


ayshih:
Ah, one more to add, hang on

ayshih:
Now good to go again

I'll also point out that this extends to the protection of #8257 to these not-quite-lists, which is important while we are still waiting for astropy 7.1.1 to come out

PR #8358, opened 2025-09-22
Implements Sphinx tagging for the gallery

### LINKED ISSUE #2164: Add some metadata to each example.
[Matplotlib's gallery](http://matplotlib.org/gallery.html) has tagged each example with something like this:

```
.. only:: html

 .. rst-class:: sphx-glr-signature

    Keywords: matplotlib code example, codex, python plot, pyplot
    `Gallery generated by Sphinx-Gallery
    <https://sphinx-gallery.readthedocs.io>`_
```
Which allows users to search the example gallery. Which they suggest how to do here:
https://matplotlib.org/faq/howto_faq.html#how-to-search-examples

We should follow this example and do it for our own examples.

CyclingNinja:
Fixes #2164 

This is the initial PR for adding tags to the gallery.

This uses https://github.com/sphinx-gallery/sphinx-gallery/pull/1555

The most important thing to discuss here is the set of tags and also use it to improve the sphinx-gallery PR.


nabobalis:
I wouldn't call it elegant but I assume that's just due to sphinx.

Cadair:
We're looking at improvements we can make to sphinx-tags but getting the tags in and getting it working is the goal of this PR.

docs/conf.py
```diff
@@ -310,6 +311,12 @@
     'only_warn_on_example_error': True,
 }
 
+# sphinx tags
```
nabobalis:
```suggestion
# -- Sphinx Tags ----------------------------------------------------------------
```

docs/conf.py
```diff
@@ -310,6 +311,12 @@
...
+tags_create_tags = True
+tags_overview_title = 'SunPy Examples Gallery Tags'
```
nabobalis:
```suggestion
tags_overview_title = 'sunpy Examples Gallery Tags'
```
?

nabobalis (approved):

Cadair:
I want to add tags to all the examples and discuss the initial set of tags we want to use in this PR so I'm marking as draft.

docs/conf.py
Cadair: This is the title of the page shouldn't it be title case?

nabobalis: I guess but it looks ugly.

nabobalis:
> I want to add tags to all the examples and discuss the initial set of tags we want to use in this PR so I'm marking as draft.

add all of them

docs/conf.py
CyclingNinja: With @Cadair on this, not starting a title with a capital makes me itch. Also as they're community examples really, should it not be title case?

Cadair:
On tags and types of tags:

- I think an obvious set of tags is the instrument of the data the example uses.
- I think adding tags which map 1:1 to the sub-galleries is pointless, but is it worth tagging the high level data objects used in an example? i.e. should all examples which make a map have a `"Map"` tag? What about anything that uses a coordinate transform having a coordinates tag?
- What about data sources / data providers? Should all the examples which query the VSO have a VSO tag?
- What about formatting? I'm leaning towards tags in title case with spaces and symbols as needed? i.e. `Computer Vision` and `Coordinates & Times` rather than `computer_vision` and `coordinates_and_times` as examples.

alasdairwilson:
> On tags and types of tags:
> 
> * I think an obvious set of tags is the instrument of the data the example uses.
> * I think adding tags which map 1:1 to the sub-galleries is pointless, but is it worth tagging the high level data objects used in an example? i.e. should all examples which make a map have a `"Map"` tag? What about anything that uses a coordinate transform having a coordinates tag?
> * What about data sources / data providers? Should all the examples which query the VSO have a VSO tag?
> * What about formatting? I'm leaning towards tags in title case with spaces and symbols as needed? i.e. `Computer Vision` and `Coordinates & Times` rather than `computer_vision` and `coordinates_and_times` as examples.

Yeah any example with data from an instrument gets that instruments tag(s)

I don’t think 1:1 or near-1:1 tags are pointless as providing alternative ways of browsing is not necessarily a bad thing, like hitting "Map" and getting all of the map subgalleries examples plus also like 4 scattered ones that have been considered as being heavily Map focused in other galleries or the showcase.

Data sources...errr not convinced by this, on the one hand I can see someone clicking "VSO" in an attempt to find out how to download data from the VSO but in reality most of the examples would just be like this random submap example that happens to use AIA data or whatever. i.e. There’s a difference between examples about VSO and examples that happen to call VSO and we don't have many of the former, if there are enough of the former then great those could be tagged.

Yes formatting human readable please, if there is no technical reason why we cant store "Coordinates & Times" in the list of tags then there is no reason we should avoid it.

Its gonna be really hard to pin down "categories" of tag but I assume there should be rough rules, uses an instrument is a good rule for an instrument tag, demonstrates a method on a class (Map, Timeseries) can be tagged as such and I mean demonstrates rather than just uses.

Then I guess broad intent, is it data retreival, is it visualisation, time series analysis, image processing, coordinate transformation

Another category could be science-driven?

Flares, Active regions, coronal holes, sunspots, solar wind, i.e. the thing that i am interested in studying, here are some examples that study those things. though I understand we are light on some of these.

That makes me think of wavelength, we could have Radio, X-ray, EUV, spectral,

IO tags or just IO as a tag? cause we could tag stuff as like NC, ASDF, FITS but we are generally data container agnostic so maybe just io could be a thing, though im struggling to think if we ever do io without data retrieval.

examples/acquiring_data/search_cdaweb.py
```diff
@@ -8,6 +8,7 @@
 CDAWeb stores data from from current and past space physics missions, and is
 full of heliospheric insitu datasets.
 """
+# sphinx_gallery_tags = ["Acquiring Data", "CDAWeb" "Solar Orbiter"]
```
samaloney:
```suggestion
# sphinx_gallery_tags = ["Acquiring Data", "CDAWeb", "Solar Orbiter"]
```

samaloney:
Maybe merge `processing` and `analysis`?

examples/plotting/xy_lims.py
```diff
@@ -7,6 +7,8 @@
 range using Matplotlib's ``set_xlim`` and ``set_ylim`` when plotting a
 Map with WCSAxes.
 """
+# sphinx_gallery_tags = ["Visualization", "AIA", "RHESSI", "Coordinates", "Map"]
```
wtbarnes:
```suggestion
# sphinx_gallery_tags = ["Visualization", "AIA", "Coordinates", "Map"]
```

Cadair:
I've added a section to the developer docs where I discuss the philosophy of tagging, comments on that should be added and then I can add / remove more tags.

https://github.com/sunpy/sunpy/pull/8358/changes#diff-544eb2b6c970cb0f4cfe8ee97a83ba904e5c6e89aeddc0bfbc2aab73fc0ed2b1R51

examples/units_and_coordinates/spice.py
```diff
@@ -29,6 +29,8 @@
     installed.
 
 """
+# sphinx_gallery_tags = ["Coordinates"]
```
Cadair:
```suggestion
# sphinx_gallery_tags = ["Coordinates", "Parker Solar Probe", "Solar Orbiter"]
```

pyproject.toml
```diff
@@ -135,7 +135,7 @@ docs = [
   "sphinx-changelog>=1.5.0",
   "sphinx-copybutton>=0.5.0",
   "sphinx-design>=0.5.0",
-  "sphinx-gallery>=0.14.0",
+  "sphinx-gallery@git+https://github.com/sphinx-gallery/sphinx-gallery@master",
```
Cadair:
```suggestion
  "sphinx-gallery>=0.21.0",
```

.pre-commit-config.yaml
```diff
@@ -55,9 +55,17 @@ repos:
       - id: generate-sunpy-net-hek-attrs
         name: generate sunpy.net.hek.attrs
         entry: python tools/hek_mkcls.py
-        language: system
+        language: python
         pass_filenames: false
         files: (^tools/(hek_mkcls|hektemplate)\.py$)|(^sunpy/net/hek/attrs\.py$)
+      - id: example-gallery-must-have-tags
```
Cadair, line 61: Turns out adding a pre-commit hook was the easiest way to make sure I hadn't forgotten any examples lol

ayshih (approved):

wtbarnes (approved):

PR #8367, opened 2025-10-07
Updates from package template

Cadair:
This PR has been generated by a script, it should update the repo with the latest changes from the package template.


changelog/8367.breaking.rst
```diff
@@ -0,0 +1,5 @@
+Increased minimum version of Python to 3.12.
+Increased minimum version of NumPy to 1.26.0.
+Increased minimum version of Matplotlib to 3.8.0.
+Increased minimum version of SciPy to 1.12.0.
+Increased minimum version of astropy to 6.1.0.
```
nabobalis: Woops.

Update the full list when its been worked out.

pyproject.toml
```diff
@@ -21,16 +21,15 @@ build-backend = "setuptools.build_meta"
...
+requires-python = ">=3.12"
 readme = { file = "README.rst", content-type = "text/x-rst" }
 license = { file = "LICENSE.rst" }
 authors = [
   { name = "The SunPy Community", email = "sunpy@googlegroups.com" },
 ]
 dependencies = [
   "astropy>=6.1.0",
-  "numpy>=1.25.0",
-  "packaging>=23.2",
```
Cadair, line 33: Do we not use `packaging`?

nabobalis: I forget, I feel like we added it due to older versions of python.

Remove it, if it breaks, we add it back :P

nabobalis:
We need to update:

[tool.distutils.bdist_wheel]
py-limited-api = "cp311"

and the hex code in the ANA file.

pyproject.toml
Cadair: Well it's definitely an implicit dependency but it'd be nice to be explicit about it.

nabobalis: It is called directly in verison.py

Cadair: yes, we use it explicitly in multiple places so we should add it back.

nabobalis: So I will add it back.

pyproject.toml
```diff
@@ -49,35 +49,37 @@ classifiers = [
...
   "mpl-animators>=1.2.0",
-  "reproject>=0.12.0",
-  "scipy>=1.11.0",
+  "reproject>=0.13.0",
+  "scipy>=1.12.0",
 ]
 opencv = ["opencv-python>=4.8.0.74"]
 net = [
-  "beautifulsoup4>=4.13.0",
+  "beautifulsoup4>=4.13.1",
```
nabobalis, line 82: This was a yanked release, so I just bumped the minor.

pyproject.toml
```diff
@@ -21,20 +21,20 @@ build-backend = "setuptools.build_meta"
...
   "astropy>=6.1.0",
-  "numpy>=1.25.0",
+  "fsspec>=2023.6.0",
+  "numpy>=1.26.0",
   "packaging>=23.2",
   "parfive[ftp]>=2.1.0",
   "pyerfa>=2.0.1.1",
-  "requests>=2.32.0",
-  "fsspec>=2023.6.0",
+  "requests>=2.32.1",
```
nabobalis, line 37: This was a yanked release, so I just bumped the minor.

nabobalis:
I pulled the 3.12 container and it had pip 25.0.1 installed, so I removed setup.py

Cadair:
Looks like the build issues have been fixed, so now it's real 3.14 compat test fails.

nabobalis:
Might be fixed, lets find out.

Cadair:
Why did you merge  this 314 still failed.

Cadair:
The numpy matrix multi issue on py314 is due to tox collecting the wrong version of numpy (2.2 vs 2.3). The bug is fixed on 2.3

PR #8369, opened 2025-10-08
Make the test suite run with only "core" extras

Cadair:
This was initiated by the fact the conda recipe can't install spiceypy on ppcle64, which broke the tests, but it spiralled a little.

This PR makes the changes needed to run the test suite with only the "core" extra which is currently the ``image,map,net,timeseries,visualization`` extras. So only the main subpackage deps rather than more "optional optionals" like spiceypy or opencv.

I've made this the "core" CI job as well as it's less likely to break on major Python versions, and also should be slightly faster to install (40s apparently).

### DIFF
--- a/sunpy/image/transform.py
+++ b/sunpy/image/transform.py
@@ -322,8 +322,8 @@ def _rotation_skimage(image, matrix, shift, order, missing, clip):
     """
     try:
         import skimage.transform
-    except ImportError:
-        raise ImportError("The scikit-image package is required to use this rotation method.")
+    except ImportError as e:
+        raise ImportError("The scikit-image package is required to use this rotation method.") from e
 
     # Make the rotation matrix 3x3 to include translation of the image
     skmatrix = np.zeros((3, 3))
@@ -384,8 +384,8 @@ def _rotation_cv2(image, matrix, shift, order, missing, clip):
     """
     try:
         import cv2
-    except ImportError:
-        raise ImportError("The opencv-python package is required to use this rotation method.")
+    except ImportError as e:
+        raise ImportError("The opencv-python package is required to use this rotation method.") from e
 
     _CV_ORDER_FLAGS = {
         0: cv2.INTER_NEAREST,


sunpy/image/tests/test_transform.py
```diff
@@ -1,5 +1,7 @@
 import numpy as np
 import pytest
+
+pytest.importorskip("skimage")
```
Cadair: This one needs fixing, we shouldn't skip this whole module, but we use a sample image from skimage.

Cadair:
The followup question to this PR is should the conda package only depend on the "core" deps? Should "core" be our default install for pip users? etc.

pyproject.toml
```diff
@@ -139,6 +143,7 @@ docs-gallery = [
   "sunpy[docs]",
   "astroquery>=0.4.6",
   "jplephem>=2.19",
+  "pillow",
```
nabobalis, line 146: Did I drop this by mistake?

Cadair: It's used in the Earth to scale example, but I think we only had it installed as a transitive dependency.

nabobalis:
> The followup question to this PR is should the conda package only depend on the "core" deps? Should "core" be our default install for pip users? etc.

SUNPY-BASE WILL LIVE AFTER ALL.

I WAS RIGHT WHEN I WANTED THAT

Cadair:
We can split the conda feedstock, but even our "core" deps are a pretty long list, so I'm still kinda dubious on the value in doing it.

alasdairwilson:
> The followup question to this PR is should the conda package only depend on the "core" deps? Should "core" be our default install for pip users? etc.

instinct is, I dont think conda should only depend on the core deps, I think conda shockingly doesn;t support extras in any kind of equivalent way so it would be a pain for a user to switch to pip to grab them.

As for core being default install, yes 100%, I have always hated that you install sunpy and map and timeseries don't work.

PR #8370, opened 2025-10-08
Actually set title in MapSequence.plot

### LINKED ISSUE #8368: `MapSequence.plot` ignores `title` if `annotate=True`
### Describe the bug

When calling `plot` on `MapSequence`, the `title=` keyword is ignored when `annotate=True`. This is in contrast to `GenericMap.plot` which respects `title=` even when `annotate=True`. `MapSequence.plot` respects `title=` when `annotate=False`, but this produces less-than-desirable axes labels.

Like `GenericMap.plot`, it should be possible to set the title on a `MapSequence` animation while still also preserving the nice axes labels.

The problem is this line https://github.com/sunpy/sunpy/blob/03470e8f936ddf0f0ef37ec216b2a5f70dc51bbd/sunpy/map/mapsequence.py#L347 which does not account for the existence of the `title=` keyword.

### To Reproduce

```Python
import sunpy.data.sample
import sunpy.map
m = sunpy.map.Map(sunpy.data.sample.AIA_094_IMAGE)
mseq = sunpy.map.Map([sunpy.data.sample.AIA_094_IMAGE, sunpy.data.sample.AIA_131_IMAGE],
                     sequence=True)
# Sets title to "My Title"
m.plot(annotate=True, title='My Title')
# Ignores title
mseq.plot(annotate=True, title='My Title')
# Sets title but does not set axes labels
```

### Screenshots

_No response_

### System Details

_No response_

### Installation method

_No response_

wtbarnes:
Fixes #8368

### DIFF
--- a/sunpy/map/mapsequence.py
+++ b/sunpy/map/mapsequence.py
@@ -343,8 +343,8 @@ class MapSequence:
         removes = []
 
         # Normal plot
-        def annotate_frame(i):
-            axes.set_title(f"{self[i].name}")
+        def annotate_frame(i, title=None):
+            axes.set_title(f"{self[i].name}" if title is None else title)
             axes.set_xlabel(axis_labels_from_ctype(self[i].coordinate_system[0],
                                                    self[i].spatial_units[0]))
             axes.set_ylabel(axis_labels_from_ctype(self[i].coordinate_system[1],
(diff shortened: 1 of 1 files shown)


nabobalis:
We want to add a test?

wtbarnes:
> We want to add a test?

Yeah, should do.

PR #8382, opened 2025-10-24
Add key to `reproject_to` to preserve original date

### LINKED ISSUE #8323: Flag added to sunpy.coordinates.propagate_with_solar_surface to keep original obs time
### Describe the feature

It's a pretty common task to do something like align a stack of AIA cutouts. When you use `sunpy.coordinates.propagate_with_solar_surface`, though, it forcibly replaces the data's obs_time with that of the reference map to which it was aligned. The header info can't be updated or overwritten, so you're stuck with a stack of files with the same time stamp.

### Proposed solution

It would be great to have a flag like `obs_time_override` or something that lets you keep your image's actual time stamp.

wtbarnes:
Fixes #8323 

I'm not completely sure if this is the best way to modify the resulting date. I tried to modify the date on the WCS object itself but for some reason I don't seem to be able to modify any of the date properties in-place. Modifying the `GenericMap.meta` object seemed the most straightforward.

### DIFF
--- a/docs/whatsnew/7.1.rst
+++ b/docs/whatsnew/7.1.rst
@@ -74,7 +74,7 @@ Now all queries will have an ``.errors`` property that contains any client error
 Preserving the observation time of a map when reprojecting
 ==========================================================
 
-When calling `~sunpy.map.GenericMap.reproject_to`, it is not possible to preserve the original observation time
+When calling `~sunpy.map.GenericMap.reproject_to`, it is now possible to preserve the original observation time
 of the map when reprojecting to a new WCS through the ``preserve_date_obs`` keyword argument. This does not affect
 the reprojection itself, which instead depends on the reference times of the two coordinate frames. This is useful,
 for example, when reprojecting a series of images to a common coordinate frame and you want to preserve the


sunpy/map/mapbase.py
```diff
@@ -3195,6 +3201,10 @@ def reproject_to(self, target_wcs, *, algorithm='interpolation', return_footprin
         # Create and return a new GenericMap
         outmap = GenericMap(output_array, target_wcs.to_header(),
                             plot_settings=self.plot_settings)
+        if preserve_date:
+            # date-avg is the time used as the coordinate frame obstime and
+            # will already be set by the target_wcs
+            outmap.meta['date-obs'] = self.date.isot
```
Cadair: I think the correct way to do this would be with the `_set_date` method: https://github.com/sunpy/sunpy/blob/228be4348fedfa5e5ecb8c898a18dbf06226e547/sunpy/map/mapbase.py#L1031

Some classes override this to fix the metadata in their way.

sunpy/map/mapbase.py
```diff
@@ -3132,6 +3133,11 @@ def reproject_to(self, target_wcs, *, algorithm='interpolation', return_footprin
             target WCS, the extent will be shifted/expanded/cropped by an integer number
             of pixels.
             Defaults to ``None``.
+        preserve_date : `bool`
+            If ``True``, the ``'DATE-OBS'`` of the original map will be preserved in the
+            header of the reprojected map. Note that this will not affect the obstime of
+            the reprojected map as this coordinate frame is defined by ``'DATE-AVG`` which
+            will be the obstime of the target WCS.
```
Cadair: I think if you change to using `_set_date` it might be better to reword this to use property names rather than FITS headers.

wtbarnes:
I think there are a few things I need to think more carefully about here, namely:

- `target_wcs` might not come from a `.wcs` Map property. What happens if `"DATE-OBS"` is the only date stored in `target_wcs`?
- Probably should avoid modifying  `.meta` after creation of the new map. One option is to modify the output of `target_wcs.to_header()`

sunpy/map/mapbase.py
wtbarnes: Ok I went back and forth on using this and `_set_reference_date` and ultimately decided against it. My reasoning for just setting the keys explicitly in the target header before passing that to reproject is because using the `_set_*` methods could inadvertently lead to one action undoing the other, depending on what keys are present on the target header. For example, if `target_wcs` has no `DATE-AVG` key, neither will `outmap` and thus using `_set_reference_date` results in overriding the `DATE-OBS` key if we've already called `_set_date`, thereby undoing the effect of `preserve_date`. If the ordering of operations is reversed, the reference date gets reset to the  date of the original WCS. Overall, I think it is more clear (and less error prone) to just set the keys explicitly. Also, `reproject_to` always produces `GenericMap` so we don't have to worry about the behavior of any subclasses.

sunpy/map/mapbase.py
wtbarnes: As noted above, I didn't end up using those methods so the docstring is unchanged.

samaloney:
Hum I think this could be a slippery slope would creating a new map class for reprojected data with a new attribute be crazy?

Because a reprojected map is different to a standard map and should have a slightly different API?

```
class ReprojectedMap(GenericMap)
    """
    A map which has been retrojected ....
    """
    def __init__(args, **kwargs)
        orig_date_obs = kwargs.get('original_date_obs', None)
        super().__init__(args, **kwargs)
        self.original_date_obs == orig_date_obs
```

sunpy/map/mapbase.py
```diff
@@ -3192,8 +3198,16 @@ def reproject_to(self, target_wcs, *, algorithm='interpolation', return_footprin
         if return_footprint:
             output_array, footprint = output_array
 
+        target_header = target_wcs.to_header()
+        if preserve_date:
+            # Explicitly not using _set_date or _set_reference_date. We do not know whether
+            # target_header has a DATE-AVG key and if it does not, we would inadvertently
+            # overwrite DATE-OBS (that we are supposed to be preserving).
+            target_header['DATE-OBS'] = self.date.utc.isot
+            target_header['DATE-AVG'] = target_header.get('DATE-AVG', 'DATE-OBS')
```
Cadair: Is this what you mean to be doing? this is `DATE-AVG` from the input WCS or the date of the map? (you've just changed what `DATE-OBS` is in the line above)?

Cadair:
> Hum I think this could be a slippery slope would creating a new map class for reprojected data with a new attribute be crazy?

I'm still a hard -1 on map sources with different API. What do you think the slippery slope is?

sunpy/map/mapbase.py
wtbarnes: 🤦‍♂️ The order should be reversed.

wtbarnes:
> Hum I think this could be a slippery slope would creating a new map class for reprojected data with a new attribute be crazy?

Also not a fan of subclasses for reprojected data, but I'm not sure I understand what the advantage would even be of this. The different properties `date` and `reference_date` already allow for what you're suggesting here.

sunpy/map/mapbase.py
```diff
@@ -3132,6 +3133,11 @@ def reproject_to(self, target_wcs, *, algorithm='interpolation', return_footprin
             target WCS, the extent will be shifted/expanded/cropped by an integer number
             of pixels.
             Defaults to ``None``.
+        preserve_date : `bool`
+            If ``True``, ``'DATE-OBS'`` of the original map will be preserved in the
+            header of the reprojected map. Note that this will not affect the obstime of
+            the reprojected map as this coordinate frame is defined by ``'DATE-AVG'`` which
+            will be the obstime of the target WCS.
```
ayshih:
```suggestion
            If ``True``, the observation time of the reprojected map is set to the
            observation time of the original map instead of the observation time defined in the
            target WCS.  This does not affect the reprojection itself, which instead depends on
            the reference times of the two coordinate frames.
            Defaults to ``False``.
```
Your description is too jargon-y, incorrect in some cases, and very easy to misinterpret even in the cases that it is correct

wtbarnes: I accidentally committed before seeing you had made a suggested fix but I copied and pasted that in.

wtbarnes: FWIW, based on previous reviews, I think you and @Cadair have different opinions about the level of "jargon" to be included in this docstring. Fight amongst each other if you wish.

ayshih: > you and @Cadair have different opinions about the level of "jargon" to be included in this docstring

Heh.  I'm not opposed to jargon, but there needs to be a non-jargon-y version that is accessible to folks who aren't steeped in FITS WCS, especially in the parameter description.  A jargon-y version can go in the notes section.

sunpy/map/tests/test_reproject_to.py
```diff
@@ -259,3 +260,35 @@ def test_reproject_to_screen_plus_diffrot(aia171_test_map, screen):
...
+    return get_dummy_map_from_header('EIT_header/efz20040301.000010_s.header')
+
+
+@pytest.fixture
+def m_eit_2():
+    return get_dummy_map_from_header('EIT_header/efz20040301.120010_s.header')
+
+
+def test_reproject_to_preserve_date_obs(m_eit_1, m_eit_2):
+    "Test that preserve_date_obs preserves the date of the original date in DATE-OBS"
```
Cadair:
```suggestion
    """Test that preserve_date_obs preserves the date of the original date in DATE-OBS"""
```

sunpy/map/tests/test_reproject_to.py
```diff
@@ -259,3 +260,35 @@ def test_reproject_to_screen_plus_diffrot(aia171_test_map, screen):
...
+    with pytest.warns(SunpyUserWarning, match="rsun mismatch detected: "):
+        m_2_r = m_eit_2.reproject_to(m_eit_1.wcs, preserve_date_obs=True)
+    assert m_2_r.date == m_eit_2.date
+    assert m_2_r.date_average == m_eit_1.date
+    assert m_2_r.coordinate_frame.obstime == m_eit_1.coordinate_frame.obstime
+
+
+def test_reproject_to_preserve_date_obs_target_has_no_date_avg(m_eit_1, m_eit_2):
+    "Test that date is preserved even when target WCS does not have a date-avg"
```
Cadair:
```suggestion
    """Test that date is preserved even when target WCS does not have a date-avg"""
```

sunpy/map/mapbase.py
```diff
@@ -3134,8 +3141,16 @@ def reproject_to(self, target_wcs, *, algorithm='interpolation', return_footprin
         if return_footprint:
             output_array, footprint = output_array
 
+        target_header = target_wcs.to_header()
+        if preserve_date_obs:
+            # Explicitly not using _set_date or _set_reference_date. We do not know whether
+            # target_header has a DATE-AVG key and if it does not, we would inadvertently
+            # overwrite DATE-OBS (that we are supposed to be preserving).
+            target_header['DATE-AVG'] = target_header.get('DATE-AVG', 'DATE-OBS')
```
Cadair: Should this follow the same chain of keys as the date property, i.e. fallback to DATE-BEG?

changelog/8382.feature.rst
```diff
@@ -0,0 +1,2 @@
+Add a keyword argument to :meth:`~sunpy.map.GenericMap.reproject_to` to preserve the ``DATE-OBS``
+key in the reprojected map.
```
ayshih:
```suggestion
Add a keyword argument to :meth:`~sunpy.map.GenericMap.reproject_to` to preserve the original observation time (typically the ``DATE-OBS`` header key) instead of the observation time in the target WCS.
```

sunpy/map/mapbase.py
wtbarnes: Maybe? My reasoning here was based on the expected availability of keys within the target WCS rather than the way we construct any of the date properties on the resulting map.

My understanding was that a FITS-WCS (and thus the resulting header) will always have a DATE-OBS key and thus is safe to fall back to. Depending on how it was constructed, it may also have a DATE-AVG key.

wtbarnes:
> We discussed this at the community meeting and decided it needs some examples updating / adding.

I've modified the time-distance plot and active region tracking examples to use this keyword. In the latter, I've added a few sentences of explanation as to why. I did not feel like this needed its own example as the AR tracking is one is the best illustration of why you'd want to use it.

sunpy/map/mapbase.py
Cadair: I don't *think* it's true a WCS object will always have a `DATE-OBS` key, and even if we enforce it must this line is still wrong.

Assuming we want to error if `DATE-AVG` or `DATE-OBS` is not in the target header then this line should be:
```suggestion
            target_header['DATE-AVG'] = target_header.get('DATE-AVG', target_header['DATE-OBS'])
```

Cadair:
I think this is good, but @wtbarnes I have one comment that needs resolving, and can you add something to the what's new?

sunpy/map/mapbase.py
ayshih: I agree that the `DATE-OBS` key may not be present; I recall that used to be the case for some SUVI files.  I think it's fine for this code to require `DATE-OBS`, but let's explicitly check for this situation – `preserve_date_obs=True` but no `DATE-AVG` or `DATE-OBS` – and raise a helpful error for the user.

wtbarnes: Wow what is wrong with me. Thanks for the fix. This is what happens when I try to do too many things at once.

Cadair: I agree with Albert that it would be best to raise a custom error here. Can you add that @wtbarnes?

wtbarnes: In progress. Please hold.

wtbarnes: Done. Feel free to reword the exception if you think it does not properly capture the ambiguity of not having either keyword.

docs/whatsnew/7.1.rst
```diff
@@ -70,3 +70,14 @@ Fido.search will now continue on error
 We have enhanced `sunpy.net.Fido`'s ``search`` method to be continue if a client errors on search.
 
 Now all queries will have an ``.errors`` property that contains any client errors that occurred during the search as the search completes.
+
+Preserving the observation time of a map when reprojecting
+==========================================================
+
+When calling `~sunpy.map.GenericMap.reproject_to`, it is not possible to preserve the original observation time
```
Cadair:
```suggestion
When calling `~sunpy.map.GenericMap.reproject_to`, it is now possible to preserve the original observation time
```

PR #8386, opened 2025-10-31
Remove deprecated `assume_spherical_screen`

wtbarnes:
`assume_spherical_screen` was deprecated in 6.0 https://docs.sunpy.org/en/stable/whatsnew/6.0.html#deprecate-assume-spherical-screen. Time for it to go

Still needs a changelog and for all the references to this method in the docs to be fixed since it no longer exists

### DIFF
--- a/docs/whatsnew/6.0.rst
+++ b/docs/whatsnew/6.0.rst
@@ -172,7 +172,7 @@ Another example is that the filename "aia.lev1.171A_2020_06_07T06_33_09.35Z.imag
     This will most likely force a redownload of files from several data providers.
 
 Deprecate ``assume_spherical_screen``
-======================================================================
+=====================================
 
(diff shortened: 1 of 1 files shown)


docs/whatsnew/6.0.rst
```diff
@@ -171,10 +171,10 @@ Another example is that the filename "aia.lev1.171A_2020_06_07T06_33_09.35Z.imag
 
     This will most likely force a redownload of files from several data providers.
 
-Deprecate `~sunpy.coordinates.Helioprojective.assume_spherical_screen`
+Deprecate ``assume_spherical_screen``
 ======================================================================
```
nabobalis:
```suggestion
Deprecate ``assume_spherical_screen``
=====================================
```

PR #8388, opened 2025-11-01
Adapt client filename convention

### LINKED ISSUE #8385: ADAPTClient fails after October 2024 due to filename pattern change
### Describe the bug

Thank you for developing SunPy! This is my first time reporting an issue, so please forgive me if I make any mistakes.

### Summary

When searching for ADAPT maps using SunPy's Fido client, no results are returned for certain dates, even though the data exists on the NSO GONG server. This seems to be due to a change in the filename pattern.

### Possible reason

The filename pattern for ADAPT maps appears to have changed. The change is not reflected in the official document (<https://gong.nso.edu/adapt/maps/adapt_filename_notes.txt>) nor in the SunPy ADAPT client code. This results in the client failing to locate and return the available data for specific dates.

If we look at the data available on the NSO server <https://gong.nso.edu/adapt/maps/gong/2024/>, we can see that the files for October 2024 and later exist. However the filename pattern seems to have changed as follows:

- old: adapt40311_03k012_202409300000_i00020600n1.fts.gz
- new: adapt40311_041012_202409300000_i00020600n1.fts.gz
- new: adapt40311_041012_202410010000_i00005600n1.fts.gz

- old: adapt41311_03k012_202409300000_i00020600n1.fts.gz
- new: adapt41311_041012_202409300000_i00020600n1.fts.gz
- new: adapt41311_041012_202410010000_i00005600n1.fts.gz

In old style, `k` in `03k012` represents month for code version, while in new style, `1` (after `4`) in `041012` seems to represent month (or minor version number?) of the code. This change is likely the cause of the issue.

### Tried to fix it and failed

Initially, I thought the following change in the SunPy ADAPT client code might fix the issue:

```log
diff --git a/sunpy/net/dataretriever/sources/adapt.py b/sunpy/net/dataretriever/sources/adapt.py
index 86bb12a04..f104e8a1c 100644
--- a/sunpy/net/dataretriever/sources/adapt.py
+++ b/sunpy/net/dataretriever/sources/adapt.py
@@ -84,7 +84,7 @@ class ADAPTClient(GenericClient):
             ADAPTResolution: [('1', '1.0 deg'), ('2', '0.2 deg')],
             ADAPTVersionYear: [(str(i), f"Code version year -> {2000 + i}") for i in range(1, 20)],
             ADAPTRealizations: [(str(i), f"Number of Realizations -> {i}") for i in range(1, 20)],
-            ADAPTVersionMonth: [(chr(i+96), f"Code version month -> {i}") for i in range(1, 13)],
+            ADAPTVersionMonth: [(chr(i+96), f"Code version month -> {i}") for i in range(1, 13)] + [(str(i), f"Code version number -> {i}") for i in range(0, 10)],
             ADAPTEvolutionMode: [('a', 'Data assimilation step'), ('i', 'Intermediate step'), ('f', 'Forecast step')],
             ADAPTHelioData: [('n', 'Not added or no data'), ('f', 'Far-side'), ('e', 'Emergence'), ('b', 'Both emergence & far-side')],
             ADAPTMagData: [('0', 'Not added or no data'), ('1', 'Mag-los'), ('2', 'Mag-vector'), ('3', 'Mag- both los & vector'),
```

I found that this change alone did not resolve this issue. In the following `pattern`, `{{ADAPTVersionMonth:1l}}` need to be changed to `{{ADAPTVersionMonth:1d}}` to match `adapt40311_041012_...`. However, this change does not seem to be compatible with the previous data files like `adapt40311_03k012_...`.

```python
pattern = r'https://gong.nso.edu/adapt/maps/gong/{{year:4d}}/adapt{{ADAPTFileType:1d}}{{ADAPTLonType:1d}}{{ADAPTInputSource:1d}}{{ADAPTDataAssimilation:1d}}{{ADAPTResolution:1d}}' + \
    '_{{ADAPTVersionYear:2d}}{{ADAPTVersionMonth:1l}}{{ADAPTRealizations:3d}}_{{year:4d}}{{month:2d}}{{day:2d}}{{hour:2d}}{{minute:2d}}' + \
    '_{{ADAPTEvolutionMode:1l}}{{days_since_last_obs:2d}}{{hours_since_last_obs:2d}}{{minutes_since_last_obs:2d}}{{seconds_since_last_obs:2d}}{{ADAPTHelioData:1l}}{{ADAPTMagData:1d}}.fts.gz'
```

Further modifications seem necessary, especially regarding how the `pattern` is parsed.

### To Reproduce

```python
import astropy.units as u
from astropy.time import Time
from sunpy.net import Fido


def search_adapt_maps(time: Time):
    return Fido.search(
        a.Time(time - 1.0 * u.hour, time + 1.0 * u.hour),
        a.Instrument("adapt"),
    )


# Succeeds
time = Time("2024-09-30")
result = search_adapt_maps(time)
print(result)

print("=" * 80)

# Fails
time = Time("2024-10-01")
result = search_adapt_maps(time)
print(result)

### Screenshots

Results from 1 Provider:

2 Results from the ADAPTClient:

       Start Time               End Time        Instrument Provider Source ADAPTFileType ADAPTLonType ADAPTInputSource ADAPTDataAssimilation ADAPTResolution ADAPTVersionYear ADAPTRealizations ADAPTVersionMonth ADAPTEvolutionMode ADAPTHelioData ADAPTMagData days_since_last_obs hours_since_last_obs minutes_since_last_obs seconds_since_last_obs
----------------------- ----------------------- ---------- -------- ------ ------------- ------------ ---------------- --------------------- --------------- ---------------- ----------------- ----------------- ------------------ -------------- ------------ ------------------- -------------------- ---------------------- ----------------------
2024-09-30 00:00:00.000 2024-09-30 00:00:59.999      ADAPT      NSO   GONG             4            0                3                     1               1                3                12                 k                  i              n            1                   0                    2                      6                      0
2024-09-30 00:00:00.000 2024-09-30 00:00:59.999      ADAPT      NSO   GONG             4            1                3                     1               1                3                12                 k                  i              n            1                   0                    2                      6                      0


================================================================================
Results from 1 Provider:

0 Results from the ADAPTClient:

<No columns>




### System Details

- SunPy: latest on 2025-10-31 (commit 8ea0c40128fd1606c5d97918e64688d3d44ff275 in main branch)
    - Installation: `-e sunpy[all]`
- AstroPy: 7.1.1
- Python: 3.13.7
- OS: macOS

### Installation method

git checkout

### PR DESCRIPTION
**Pull request by iijimahr**
Fixes #8385

This PR fixes ADAPTClient returning no results for ADAPT maps after October 2024,  
caused by a change in the filename convention on the NSO GONG server.  
Two URL patterns (old/new) are now supported and automatically switched based on the query time.

### DIFF
--- a/sunpy/net/dataretriever/sources/tests/test_adapt.py
+++ b/sunpy/net/dataretriever/sources/tests/test_adapt.py
@@ -85,10 +85,11 @@ def test_fetch_working(adapt_client):
     start = '2024/09/29 02:00:00'
     end = '2024/09/29 02:00:59.999'
     tr = a.Time(start, end)
-    qr = adapt_client.search(tr, a.Instrument.adapt)[0]
-    mock_qr = mock_query_object_old_pattern(adapt_client)[0]
+    queries = adapt_client.search(tr, a.Instrument.adapt)
 
     # Testing old pattern before September 2024
+    qr = queries[0]
+    mock_qr = mock_query_object_old_pattern(adapt_client)[0]
     assert mock_qr['Source'] == qr['Source']
     assert mock_qr['Provider'] == qr['Provider']
     assert mock_qr['Instrument'] == qr['Instrument']
@@ -101,7 +102,7 @@ def test_fetch_working(adapt_client):
     assert len(download_list) == 1
 
     # Testing old pattern after October 2024 (without downloading)
-    qr_new = adapt_client.search(tr, a.Instrument.adapt)[2]
+    qr_new = queries[2]
     mock_qr_new = mock_query_object_new_pattern(adapt_client)[0]
     assert mock_qr_new['Source'] == qr_new['Source']
     assert mock_qr_new['Provider'] == qr_new['Provider']

### CODE REVIEW

File: sunpy/net/dataretriever/sources/tests/test_adapt.py
Comment (nabobalis) [inline sunpy/net/dataretriever/sources/tests/test_adapt.py:1]: Can you add a remote_data test that checks that before and after still works?
Also that a mixed query which hits both file formats works? 

Maybe that last one is all we need.

**Comment by nabobalis** [comment]
This will need a bugfix changelog, see https://github.com/sunpy/sunpy/blob/main/changelog/README.rst on how to add one, thanks.

File: sunpy/net/dataretriever/sources/tests/test_adapt.py
Comment (iijimahr) [reply sunpy/net/dataretriever/sources/tests/test_adapt.py:1]: Thank you @nabobalis for review!

The current implementation checks a mixed query that hits both file formats. Since the remote data test takes about 42 seconds, I’m not sure if it’s worth adding separate queries for the old and new filename patterns. Please let me know if that’s necessary.

I’ve also refactored the code for clarity.

File: changelog/8388.bugfix.rst [hunk 1]
```diff
@@ -0,0 +1 @@
+Fix `~sunpy.net.dataretriever.ADAPTClient` returning no results for ADAPT maps after October 2024.  The filename convention was changed on the NSO GONG server.
```
nabobalis:
```suggestion
Fix `~sunpy.net.dataretriever.ADAPTClient` returning no results for ADAPT maps after October 2024.
The filename convention was changed on the NSO server.
```

sunpy/net/dataretriever/sources/tests/test_adapt.py
nabobalis: That is a long test. 

Can we change the test to do one online/remote-server query which will fetch one old file and new one file and then we can check both are returned?

iijimahr: Currently, only a single query is executed, and only one file is fetched. In the query, both the new and old filename patterns are checked.

This may be partly because I’m accessing the NSO server from Japan, and partly because the number of registered parameters has increased to match multiple formats:

```python
 # at sunpy/net/dataretriever/sources/adapt.py:ADAPTClient.register_values
ADAPTVersionMonth: [(chr(i+96), f"Code version month -> {i}") for i in range(1, 13)] + [(str(i), f"Code version number -> {i}") for i in range(0, 10)],
```

In the version on the SunPy default branch, the remote data test took about 20 seconds to complete.

nabobalis: Ah yeah, I didn't consider that. 20 seconds is fine.

But I am wondering if we should make sure the query works if you hit both the old and new file format at the same time with one query.

iijimahr: Yes, the current test already performs a single query that covers the transition period (from 2024-09-28T01:00 to 2024-09-30T23:00) in which both filename patterns exist, so both the old and new filename formats are hit within the same query.

The tested urls are (3k012 and 041012):
```python
# old
'url': ("https://gong.nso.edu/adapt/maps/gong/2024/adapt40311_03k012_202409290200_i00013600n1.fts.gz")

# new
'url': ("https://gong.nso.edu/adapt/maps/gong/2024/adapt40311_041012_202409290200_i00013600n1.fts.gz")
```

nabobalis: But it does two queries no?

iijimahr: You are right! I noticed that I had added an unnecessary second search query.
I removed that line, and the time required for the remote data test has decreased to 28 seconds.

Thank you very much for checking this.

nabobalis (approved):

nabobalis:
Thank you very much @iijimahr, when we get the situation from the NSO, we will merge this.

nabobalis:
I think based on what we know right now, I want to merge this, get a fix out and if we need to revert it in the future, we can do so very easily

PR #8391, opened 2025-11-05
Fix Python 3.14

Cadair:
It seems that one of our non-minimal deps is causing numpy 2.2 to be installed (not 2.3) which is why py314-minimal works but py314 doesn't.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,6 @@ jobs:
       posargs: -n auto --color=yes
       envs: |
         - linux: py314-minimal
-          toxargs: -vv
     secrets:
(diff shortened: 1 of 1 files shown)


nabobalis (approved):

.github/workflows/ci.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       toxdeps: tox-pypi-filter
       posargs: -n auto --color=yes
       envs: |
-        - linux: py313-minimal
+        - linux: py314-minimal
```
Cadair, line 35: we should have reverted this. py314 is still slow because there's no skimage wheels.

PR #8393, opened 2025-11-06
Remove inline-authorship information

Cadair:
fixes sunpy/sunpy#3650


Cadair:
This PR should not be merged until there is some consensus (which I think we are close to) in #3650. Specifically input from Albert as the main dissenting voice.

nabobalis:
I added the text from the issue about authorship going forward.

Ideally this should be applied for every package under the project.

alasdairwilson (approved):

ayshih:
> Specifically input from Albert as the main dissenting voice.

I have no objection given that the stance paragraph has been added

PR #8398, opened 2025-11-13
Add a new map source for HXI on the ASO-S mission

daozjing:
Hello,

In this PR, I am adding a new map source for the advanced space-based solar observatory (ASO-S) misson. 

Especally,I am working on the HXIMap currently for the Hard X-ray Imager (HXI) which is one of the three payloads of ASO-S.

- Features:
    - enable read the cube-imaging .fits file which output from the HXI analysis software, as the sunpy map
    - enable read more than one images/maps from a single .fits file
    - ...
- What I am doing:
    - add file asos.py to ./sunpy/map/sources
    - add file test_asos_hxi_source.py to ./sunpy/map/sources/tests
    - add a .fits file to ./sunpy/data/test for testing
    - write from .asos import * to ./sunpy/map/sources/\_\_init\_\_.py

### DIFF
--- a/sunpy/map/sources/tests/test_asos_hxi_source.py
+++ b/sunpy/map/sources/tests/test_asos_hxi_source.py
@@ -45,7 +45,7 @@ def test_private_date_setters(hxi_map):
 
 def test_measurement(hxi_map):
     """Tests the measurement property of the HXIMap object."""
-    assert hxi_map.measurement in [[20, 30], [20, 30]] * u.keV
+    assert u.allclose(hxi_map.measurement, [20, 30] * u.keV)
 
 
 def test_norm_clip(hxi_map):


sunpy/map/sources/tests/test_asos_hxi_source.py
```diff
@@ -0,0 +1,94 @@
+"""Test cases for ASO-S HXIMap subclass"""
```
nabobalis:
```suggestion
"""
Test cases for ASO-S HXIMap subclass.
"""
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+        """
+        If the WAVEUNIT FITS keyword is not present, defaults to keV.
+        """
+        unit = self.meta.get("waveunit", 'keV')
+        return u.Unit(unit)
+
+#     @property
+#     def wavelength(self):
+#         return u.Quantity([self.meta['energy_l'], self.meta['energy_h']],
+#                           unit=self.waveunit)
```
nabobalis:
```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+    def wavelength(self):
+        return [self.meta['energy_l'], self.meta['energy_h']]*self.waveunit
+
+    @property
+    def observatory(self):
+        return self.meta.get('ORIGIN').split('/')[0]
+
+    @property
+    def instrument(self):
+        #return self.meta['INSTRUME']
```
nabobalis:
```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+    
+    @property
+    def detector(self):
+        return self.instrument
+
+    @classmethod
+    def is_datasource_for(cls, data, header, **kwargs):
+        """Determines if header corresponds to an HXI image"""
+        #print(header.get('instrume') == 'HXI',header.get('ORIGIN'))
+        #return header.get('instrume') == 'HXI'
```
nabobalis:
```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+        if not (set(['HGLN_OBS','HGLT_OBS','DSUN_OBS']) < set(self.fits_header)):
+            
+            from sunpy.coordinates import get_earth
+            observer = get_earth(self.date_start)
+            
+            if observer.frame.name=='heliographic_stonyhurst':
+                self.meta['HGLN_OBS'] = observer.lon.to_value(u.deg)
+                self.meta['HGLT_OBS'] = observer.lat.to_value(u.deg)
+                self.meta['DSUN_OBS'] = observer.radius.to_value(u.m)
+                #log.info("Assuming Earth-based observer to set dsun_obs,hgln_obs,hglt_obs.")
```
nabobalis:
```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
+'''ASO-S Map subclass definitions, currently for the HXI payload is available.'''
+
+import astropy.units as u
+#from sunpy import log
+from sunpy.time import parse_time
+from sunpy.map.mapbase import GenericMap, SpatialPair
+
+__author__="FuYu@PMO, from ASO-S/HXI Team"
```
nabobalis:
```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
+'''ASO-S Map subclass definitions, currently for the HXI payload is available.'''
+
+import astropy.units as u
+#from sunpy import log
```
nabobalis:
```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
+'''ASO-S Map subclass definitions, currently for the HXI payload is available.'''
```
nabobalis:
```suggestion
"""
ASO-S Map subclass definitions.
"""
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+from sunpy.map.mapbase import GenericMap, SpatialPair
+
+__author__="FuYu@PMO, from ASO-S/HXI Team"
+
+__all__ = ['HXIMap']
+
+
+class HXIMap(GenericMap):
+    """
+    HXI Image Maps (single map or multi-maps from one .fits file).
```
nabobalis:
```suggestion
    HXI Image Map Source.
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+    ----------
+    * `ASO-S Mission Page <http://aso-s.pmo.ac.cn>`_
+    * `ASO-S Data Center <http://aso-s.pmo.ac.cn/sodc/dataArchive.jsp>`_
+    * `ASO-S/HXI Flare List <http://aso-s.pmo.ac.cn/hxi_flare/hxi_flare_list.html>`_
+    * `The Tests and Calibrations of the Hard X-ray Imager Aboard ASO-S <https://link.springer.com/article/10.1007/s11207-024-02392-x>`_
+    
+    History:
+    --------
+        2025 08 27, FuYu@PMO, ASO/HXI Team. rsun_meters use the default value from sunpy (not fix).
+             11 11, add 'fix_hxi_observer' keyword.
```
nabobalis:
```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+    The Hard X-ray Imager (HXI, Zhang et al. 2019b; Su et al. 2019) is one of the three payloads 
+    of the advanced space-based solar observatory (ASO-S, Gan et al. 2019, 2023).
+    
+    The energy ranges are ~15-300 keV for imaging and ~10-300 keV for spectra (Su et al. 2024).
+
+    ASO-S was launched on 9 October 2022 (UTC+8).
+
+    References
+    ----------
+    * `ASO-S Mission Page <http://aso-s.pmo.ac.cn>`_
```
nabobalis:
```suggestion
    * `ASO-S Mission Page <http://aso-s.pmo.ac.cn>`__
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+    * `ASO-S Data Center <http://aso-s.pmo.ac.cn/sodc/dataArchive.jsp>`_
```
nabobalis:
```suggestion
    * `ASO-S Data Center <http://aso-s.pmo.ac.cn/sodc/dataArchive.jsp>`__
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+    * `ASO-S/HXI Flare List <http://aso-s.pmo.ac.cn/hxi_flare/hxi_flare_list.html>`_
```
nabobalis:
```suggestion
    * `ASO-S/HXI Flare List <http://aso-s.pmo.ac.cn/hxi_flare/hxi_flare_list.html>`__
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+    * `The Tests and Calibrations of the Hard X-ray Imager Aboard ASO-S <https://link.springer.com/article/10.1007/s11207-024-02392-x>`_
```
nabobalis:
```suggestion
    * `The Tests and Calibrations of the Hard X-ray Imager Aboard ASO-S <https://link.springer.com/article/10.1007/s11207-024-02392-x>`__
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+class HXIMap(GenericMap):
+    """
+    HXI Image Maps (single map or multi-maps from one .fits file).
+    
+    The Hard X-ray Imager (HXI, Zhang et al. 2019b; Su et al. 2019) is one of the three payloads 
+    of the advanced space-based solar observatory (ASO-S, Gan et al. 2019, 2023).
+    
+    The energy ranges are ~15-300 keV for imaging and ~10-300 keV for spectra (Su et al. 2024).
```
nabobalis: Can you link to all of these citations?

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+
+    ASO-S was launched on 9 October 2022 (UTC+8).
```
nabobalis:
```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+
+__all__ = ['HXIMap']
+
+
+class HXIMap(GenericMap):
+    """
+    HXI Image Maps (single map or multi-maps from one .fits file).
+    
+    The Hard X-ray Imager (HXI, Zhang et al. 2019b; Su et al. 2019) is one of the three payloads 
+    of the advanced space-based solar observatory (ASO-S, Gan et al. 2019, 2023).
```
nabobalis:
```suggestion
    of the Advanced Space-based Solar Observatory (ASO-S, Gan et al. 2019, 2023).
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+    
+    History:
+    --------
+        2025 08 27, FuYu@PMO, ASO/HXI Team. rsun_meters use the default value from sunpy (not fix).
+             11 11, add 'fix_hxi_observer' keyword.
+    """
+
+    def __init__(self, data, header, **kwargs):
+        super().__init__(data, header, **kwargs)
+        asos_hxi, hxi = self.meta.get('ORIGIN').split(' ')[0], self.meta.get('INSTRUME','') #deal with different headers
```
nabobalis:
```suggestion
        asos_hxi, hxi = self.meta.get('ORIGIN').split(' ')[0], self.meta.get('INSTRUME','')
```
What are the different headers?

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+        self._nickname =  asos_hxi if 'HXI' in asos_hxi else asos_hxi+'/'+hxi              #self.detector
```
nabobalis:
```suggestion
        self._nickname =  asos_hxi if 'HXI' in asos_hxi else asos_hxi+'/'+hxi
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+        self.plot_settings['cmap'] = 'rhessi' #Using default cmap the same as RHESSI
```
nabobalis:
```suggestion
        self.plot_settings['cmap'] = 'rhessi'
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+        
+        if kwargs.get('fix_hxi_observer',True):
+            self._fix_hxi_observer()
```
nabobalis: I am not sure this is a supported keyword for Map initialization. 
This should be called every time if there is no way to determine if it needs to be called.

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+            
+    def _fix_hxi_observer(self):
+        """
+        Add the missing info. of 'dsun_obs','hgln_obs' and 'hglt_obs' for the fits_header, manually.
```
nabobalis:
```suggestion
        Add the missing information for 'dsun_obs', 'hgln_obs' and 'hglt_obs'.
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+        ASO-S is located near Earth, so the observer is assumed as Earth to get the info. 
+        according to the 'date_start' with sunpy.coordinates.get_earth
+        
+        Usage: set keyword 'fix_hxi_observer' when using sunpy.map.Map(), default:True.
+               if not fix, the sunpy will fix it automatically in some cases.
```
nabobalis:
```suggestion
        ASO-S is located near Earth, so the observer is assumed as Earth.
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+        """
+        if not (set(['HGLN_OBS','HGLT_OBS','DSUN_OBS']) < set(self.fits_header)):
+            
+            from sunpy.coordinates import get_earth
```
nabobalis: I think this can be at the top of the file.

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+    def __init__(self, data, header, **kwargs):
+        super().__init__(data, header, **kwargs)
+        asos_hxi, hxi = self.meta.get('ORIGIN').split(' ')[0], self.meta.get('INSTRUME','') #deal with different headers
+        self._nickname =  asos_hxi if 'HXI' in asos_hxi else asos_hxi+'/'+hxi              #self.detector
+        self.plot_settings['cmap'] = 'rhessi' #Using default cmap the same as RHESSI
+        
+        if kwargs.get('fix_hxi_observer',True):
+            self._fix_hxi_observer()
+            
+    def _fix_hxi_observer(self):
```
nabobalis: I am not sure this is the way this is meant to be fixed. Someone else with more experience with Map and WCS will need to chime in.

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+                self.meta['HGLN_OBS'] = observer.lon.to_value(u.deg)
+                self.meta['HGLT_OBS'] = observer.lat.to_value(u.deg)
+                self.meta['DSUN_OBS'] = observer.radius.to_value(u.m)
+                #log.info("Assuming Earth-based observer to set dsun_obs,hgln_obs,hglt_obs.")
+            else:
+                raise ValueError("observer.frame.name not eq 'heliographic_stonyhurst', not support")
+
+    def _get_cmap_name(self):
+        """
+        Using default cmap the same as RHESSI
```
nabobalis:
```suggestion
        Using the same cmap as RHESSI.
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+        """
+        return "rhessi" 
+
+    def _rotation_matrix_from_crota(self):
+        """
+        Rotation in CROTA. HXI may not need
```
nabobalis:
```suggestion
        Rotation from CROTA.
```

sunpy/data/test/hxi_imgcube_01e02t_20230501_130758_HXI_CLEAN.fits
nabobalis, line 1: Can this be turned into a FITS header such as https://github.com/sunpy/sunpy/blob/main/sunpy/data/test/20181209_180305_kcor_l2.header?

There are functions which will create a fake FITS file from this and allow us not to check in the binary file if we can avoid it.

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+        """
+        return super()._rotation_matrix_from_crota(crota_key='CROTA')
+
+    @property
+    def spatial_units(self):
+        """
+        If cunit{1 or 2} are equal to 'arcsecs', assumes that cunit{1 or 2}
+        respectively are intended to be 'arcsec'.
+        """
```
nabobalis:
```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+            units[1] = 'arcsec'
+        units = [None if unit is None else u.Unit(unit.lower()) for unit in units]
+        return SpatialPair(units[0], units[1])
+
+    @property
+    def coordinate_system(self):
+        """
+        If CTYPE{1 or 2} are equal to 'solar_x' or 'solar_y', assumes that CTYPE{1 or 2}
+        respectively are intended to be 'HPLN-TAN' or 'HPLT-TAN'.
+        """
```
nabobalis:
```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+            ctypes[0] = 'HPLN-TAN'
+        if ctypes[1] == 'solar_y':
+            ctypes[1] = 'HPLT-TAN'
+        return SpatialPair(*ctypes)
+    
+    @property
+    def date_start(self):
+        """DATE-OBS is the start date"""
+        date_obs = self.meta.get('DATE-OBS')
+        if len(date_obs.split(" ")[0])==9:
```
nabobalis: What format is DATE-OBS in that it needs to be split?

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+            date_obs = date_obs[:7]+'20'+date_obs[7:]
+        return parse_time(date_obs)
+
+    @property
+    def date_end(self):
+        return self.date_start + self.meta.get('exptime')*u.second
+    
+    @property
+    def _date_obs(self):
+        """using date_start"""
```
nabobalis:
```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+        return self.date_start
+        
+    @property
+    def reference_date(self):
+        """using date_start, compatible with the old version of sunpy """
```
nabobalis: What old version of sunpy?

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+    def img_bkg(self):
+        return self.meta.get('BKG')
+
+    @property
+    def img_counts(self):
+        return self.meta.get('COUNTS')
+
+    @property
+    def img_algorith(self):
+        return self.meta.get('ALGORITH')
```
nabobalis: I am not sure what the policy is on new properties onto map sources. Someone else will need to chime in.

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+    
+    @property
+    def waveunit(self):
+        """
+        If the WAVEUNIT FITS keyword is not present, defaults to keV.
+        """
```
nabobalis:
```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,178 @@
...
+        unit = self.meta.get("waveunit", 'keV')
+        return u.Unit(unit)
```
nabobalis:
```suggestion
        return u.Unit(self.meta.get("waveunit", 'keV'))
```

nabobalis (commented):
Thank you for the pull request. I would love to see this land in the next release of sunpy.

I left a quick review with no sleep, so it's a bit rougher than I would like. Hopefully it will fix most of the style issues.

sunpy/map/sources/asos.py
daozjing: Since these additional properties are not under the specification of the Sunpy map, they can be removed. 
If someone needs them, they can get from the fits_header of the map.

sunpy/map/sources/asos.py
daozjing: The reference_date is defined in the mapbase.py.

In the old version (e.g., 6.1.1) of sunpy, the value of reference_date is given by the following order:
  self._get_date('date-avg') or
  self._get_date('date-obs') or
  ...

In the recent version (e.g., 7.0.3), reference_date is given by 
  self._get_date('date-avg') or
  self._date_obs or
  ...
  
However, the 'date-obs' which directly got from the meta may not fit the standard time format.
In the recent version, self._date_obs avoids the bug from 'date-obs'.

sunpy/data/test/hxi_imgcube_01e02t_20230501_130758_HXI_CLEAN.fits
daozjing: This FITS file contains more than one FITS header. However, I will try to work on that.

sunpy/map/sources/asos.py
daozjing: When using sunpy.map.Map(), any additional keyword will enter kwargs

Cadair: To make sure all the maps have the same interface we don't allow map sources to have custom keyword arguments.

However, this `_fix_hxi_observer` method should be changed anyway, so this is irrelevant.

sunpy/map/sources/asos.py
Cadair: This is not the way this should be addressed. Here's an example of how it's done in AIA:

https://github.com/sunpy/sunpy/blob/65d7eb1d0ad690ba0d5083db74cdfff6856a8dc6/sunpy/map/sources/sdo.py#L67-L75

The `_supported_observer_coordinates` method should return a list of tuples where each tuple contains two elements, the first a list of keys needed to be present in the header to construct the given frame and the second a dictionary of keyword args to `SkyCoord`.

A more general point is that no map source should modify it's own `.meta` dictionary. All the "fixing" is done in properties.

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,160 @@
...
+
+__all__ = ['HXIMap']
+
+
+class HXIMap(GenericMap):
+    """
+    HXI Image Map Source.
+    
+    The Hard X-ray Imager (HXI, Zhang et al. 2019b; Su et al. 2019) is one of the three payloads 
+    of the Advanced Space-based Solar Observatory (ASO-S, Gan et al. 2019, 2023).
```
Cadair: We added a thing recently for citations with bibtex right @wtbarnes ?

sunpy/map/sources/asos.py
Cadair: yeah we don't allow map sources to define new properties, only override ones in `GenericMap`.

sunpy/map/sources/asos.py
daozjing: e.g., '01-May-23 13:08:16.962', which should be fixed like this '01-May-2023 13:08:16.962'.

sunpy/map/sources/asos.py
daozjing: Under my test, '_supported_observer_coordinates' just provids additional representations of observer coordinates,
e.g., in the 'heliocentricmeanecliptic' frame for AIA in the sdo.py.

The key to the problem is that regardless of the reference frame, relevant coordinates need to be obtained from the metadata (e.g., self.meta.get('haex_obs') for AIA).


However, the metadata here does not provide relevant information e.g.,dsun_obs,hgln_obs,hglt_obs for the 'heliographic_stonyhurst' frame.

In this case, for example, using test_map.plot() will received warning messages as follow:
    - WARNING: SunpyMetadataWarning: Missing metadata for observer: assuming Earth-based observer.
    - For frame 'heliographic_stonyhurst' the following metadata is missing: dsun_obs,hgln_obs,hglt_obs

For an Earth-based observer, we can consider fixing it manually or ignoring it (Sunpy will assume an Earth-based observer).

In fact, '_set_reference_date' (although it is a built-in method) is used to set self.meta['t_obs'] in sdo.py.

sunpy/map/sources/asos.py
wtbarnes: Yes. See this section of our developer guide https://docs.sunpy.org/en/latest/dev_guide/contents/documentation.html#citing-publications

sunpy/map/sources/asos.py
daozjing: Sure, I will do it.

sunpy/map/sources/asos.py
Cadair: Ah, I missed the fact that the headers do not have the observer location in them. Thanks to @ayshih for reminding me of this, but you should set this property:

https://github.com/sunpy/sunpy/blob/65d7eb1d0ad690ba0d5083db74cdfff6856a8dc6/sunpy/map/mapbase.py#L1315-L1320

to return a SkyCoord object.

daozjing: It works  well like this way.

sunpy/map/sources/asos.py
daozjing: No problem, I will deal with.

sunpy/map/sources/asos.py
daozjing: They have been removed

sunpy/data/test/hxi_imgcube_01e02t_20230501_130758_HXI_CLEAN.fits
nabobalis: What else is stored in the file?

sunpy/map/sources/asos.py
nabobalis: I see, I would only consider the oldest version being 7.0, no need to check what happens on 6.X.

sunpy/data/test/hxi_imgcube_01e02t_20230501_130758_HXI_CLEAN.fits
daozjing: In general, the file contains more than one hdu units for images from different energy and time.

daozjing: I have replaced the original FITS file with two .header files.

sunpy/map/sources/asos.py
```diff
@@ -78,35 +77,44 @@ def coordinate_system(self):
...
+        """
+        DATE-OBS is the start_date.
+        Here do some thing to deal with date string like '01-May-23 13:08:16.962' or
+        ' 1-May-2023 13:08:16.962' (the latter one is output from SSWIDL's map2fits)
+        """
+        date_obs = self.meta.get('DATE-OBS').strip()
+        tmp = date_obs.split(" ")[0]
+        if len(tmp)==9 and len(tmp.split("-")[-1])==2:
             date_obs = date_obs[:7]+'20'+date_obs[7:]
         return parse_time(date_obs)
```
nabobalis: Does parse_time not already handle the full date-obs?

changelog/8398.feature.rst
```diff
@@ -0,0 +1 @@
+Added a new map source for ASO-S misson, currently added HXIMap for the Hard X-ray Imager (HXI) payload.
```
nabobalis:
```suggestion
Added a new map source for ASO-S mission, currently added HXIMap for the Hard X-ray Imager (HXI) payload.
```
Also HXIMAP should be linked using sphinx syntax .

sunpy/data/test/hxi_imgcube_01e02t_20230501_130758_HXI_CLEAN_0.header
nabobalis, line 1: So just to clarify, the original FITS file has two HDUs?

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,135 @@
...
+__all__ = ['HXIMap']
+
+
+class HXIMap(GenericMap):
+    """
+    HXI Image Map Source.
+
+    The Hard X-ray Imager (HXI, :cite:t:`Zhang_asos_mission_hxi_2019, Su_asos_hxi_simu_soft_2019`)
+    is one of the three payloads of the Advanced Space-based Solar Observatory
+    (ASO-S, :cite:t:`Gan_asos_mission_overview_2019, Gan_asos_issue_overview_2023`), which designed
```
nabobalis:
```suggestion
    (ASO-S, :cite:t:`Gan_asos_mission_overview_2019, Gan_asos_issue_overview_2023`), which is designed
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,135 @@
...
+    to observe hard X-ray (HXR) spectra and images of solar flares. Having 91 subcollimators to
```
nabobalis:
```suggestion
    to observe Hard X-Ray (HXR) spectra and images of solar flares. Having 91 subcollimators to
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,135 @@
...
+    modulate incident X-rays, HXI can obtain 91 modulation data and 45 visibilities to reconstruct
```
nabobalis:
```suggestion
    modulate incident x-rays, HXI can obtain 91 modulation data and 45 visibilities to reconstruct
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,135 @@
...
+
+    In addition, the energy ranges are ~15-300 keV for imaging and ~10-300 keV for spectra, more information
+    please refer to :cite:t:`Su_asos_hxi_tests_calibrations_2024`.
+
+    References
+    ----------
+    * `ASO-S Mission Page <http://aso-s.pmo.ac.cn>`__
+    * `ASO-S Data Center <http://aso-s.pmo.ac.cn/sodc/dataArchive.jsp>`__
+    * `ASO-S/HXI Flare List <http://aso-s.pmo.ac.cn/hxi_flare/hxi_flare_list.html>`__
+    * `The tests and calibrations paper <https://link.springer.com/article/10.1007/s11207-024-02392-x>`__ - :cite:t:`Su_asos_hxi_tests_calibrations_2024`
```
nabobalis:
```suggestion
    * The tests and calibrations paper: :cite:t:`Su_asos_hxi_tests_calibrations_2024`
```
Not sure if this will work.

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,135 @@
...
+        super().__init__(data, header, **kwargs)
+        self._nickname =  self.meta.get('nickname', 'ASO-S/HXI')
+        self.plot_settings['cmap'] = 'rhessi'
+
+    @property
+    def _default_observer_coordinate(self):
+        """
+        The default observer coordinate to use when there is insufficient information
+        in the metadata. This should be overridden by map sources as appropriate.
+        """
```
nabobalis:
```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,135 @@
...
+        if not (set(['HGLN_OBS','HGLT_OBS','DSUN_OBS']) < set(self.fits_header)):
+            return  get_earth(self.reference_date)
+
+    def _get_cmap_name(self):
+        """
+        Using the same cmap as RHESSI.
+        """
```
nabobalis:
```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,135 @@
...
+        return super()._rotation_matrix_from_crota(crota_key='CROTA')
+
+    @property
+    def spatial_units(self):
+        units = [self.meta.get('cunit1', None), self.meta.get('cunit2', None)]
+        if self.meta['cunit1'] == 'arcsecs':
+            units[0] = 'arcsec'
+        if self.meta['cunit2'] == 'arcsecs':
+            units[1] = 'arcsec'
+        units = [None if unit is None else u.Unit(unit.lower()) for unit in units]
```
nabobalis, line 71: Are the units in the FITS header upper case?

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,135 @@
...
+    def date_end(self):
+        return self.date_start + self.meta.get('exptime')*u.second
+
+    @property
+    def _date_obs(self):
+        return self.date_start
+
+    @property
+    def reference_date(self):
+        """using date_start, compatible with the old version of sunpy """
```
nabobalis:
```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        DATE-OBS is the start_date.
+        Here do some thing to deal with date string like '01-May-23 13:08:16.962' or
+        ' 1-May-2023 13:08:16.962' (the latter one is output from SSWIDL's map2fits)
+        """
+        date_obs = self.meta.get('DATE-OBS').strip()
+        tmp = date_obs.split(" ")[0]
+        if len(tmp)==9 and len(tmp.split("-")[-1])==2:
+            date_obs = date_obs[:7]+'20'+date_obs[7:]
+        return parse_time(date_obs)
```
nabobalis: Does parse_time not handle the original DATE-OBS value?

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,135 @@
...
+    def waveunit(self):
+        return u.Unit(self.meta.get("waveunit", 'keV'))
+
+    @property
+    def wavelength(self):
+        waves = [self.meta.get('energy_l', None), self.meta.get('energy_h', None)]
+        if None in waves:
+            return None
+        else:
+            return waves*self.waveunit
```
nabobalis, line 121: Is it common to not have energy_l and energy_h in the FITS headers?

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,135 @@
...
+
+    @property
+    def observatory(self):
+        return self.meta.get('observatory', 'ASO-S')
+
+    @property
+    def instrument(self):
+        return self.meta.get('INSTRUME','HXI')
```
nabobalis:
```suggestion
        return self.meta.get('INSTRUME', 'HXI')
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,135 @@
...
+
+    @property
+    def detector(self):
+        return self.meta.get('detector','HXI')
```
nabobalis:
```suggestion
        return self.meta.get('detector', 'HXI')
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,135 @@
...
+
+    @classmethod
+    def is_datasource_for(cls, data, header, **kwargs):
+        """Determines if header corresponds to an HXI image"""
```
nabobalis:
```suggestion
        """
        Determines if header corresponds to a HXI image.
        """
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,135 @@
...
+        tag_ori = 'HXI' in header.get('ORIGIN') if header.get('ORIGIN') is not None else False
+        return tag_ori or header.get('INSTRUME') == 'HXI'
```
nabobalis:
```suggestion
        return 'HXI' in header.get('ORIGIN', '') or header.get('INSTRUME') == 'HXI'
```
My brain is too tired to work out if this is even valid.

sunpy/map/sources/tests/test_asos_hxi_source.py
```diff
@@ -0,0 +1,96 @@
...
+from sunpy.map import Map
+from sunpy.map.sources.asos import HXIMap
+from sunpy.util.exceptions import SunpyMetadataWarning
+from .helpers import _test_private_date_setters
+
+
+@pytest.fixture(scope="module")
+def hxi_map():
+    """Note: HXIMap may have more than one images"""
+    #ms = Map(get_test_filepath("hxi_imgcube_01e02t_20230501_130758_HXI_CLEAN.fits"))
```
nabobalis:
```suggestion
```

sunpy/map/sources/tests/test_asos_hxi_source.py
```diff
@@ -0,0 +1,96 @@
...
+from sunpy.data.test import get_dummy_map_from_header, get_test_filepath
+from sunpy.map import Map
+from sunpy.map.sources.asos import HXIMap
+from sunpy.util.exceptions import SunpyMetadataWarning
+from .helpers import _test_private_date_setters
+
+
+@pytest.fixture(scope="module")
+def hxi_map():
+    """Note: HXIMap may have more than one images"""
```
nabobalis:
```suggestion
```
This should be in the Map Source but also from what I recall, Map can not handle FITS with multiple images in them, it just loads the first one unless you do some magic.

sunpy/map/sources/tests/test_asos_hxi_source.py
```diff
@@ -0,0 +1,96 @@
...
+        new_tmp_map = tmp_map._new_instance(tmp_map.data, tmp_map.meta)
+        assert new_tmp_map.plot_settings['norm'].vmin is None
+        assert new_tmp_map.plot_settings['norm'].vmax is None
+        assert new_tmp_map.plot_settings['cmap'] == new_tmp_map._get_cmap_name()
+
+
+def test_wcs(hxi_map):
+    # Smoke test that WCS is valid and can transform from pixels to world coordinates
+    # with pytest.warns(SunpyMetadataWarning,
+    #                   match='Missing metadata for observer: assuming Earth-based observer.*'):
```
nabobalis:
```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,135 @@
...
+    The Hard X-ray Imager (HXI, :cite:t:`Zhang_asos_mission_hxi_2019, Su_asos_hxi_simu_soft_2019`)
+    is one of the three payloads of the Advanced Space-based Solar Observatory
+    (ASO-S, :cite:t:`Gan_asos_mission_overview_2019, Gan_asos_issue_overview_2023`), which designed
+    to observe hard X-ray (HXR) spectra and images of solar flares. Having 91 subcollimators to
+    modulate incident X-rays, HXI can obtain 91 modulation data and 45 visibilities to reconstruct
+    images with a spatial resolution as high as ~3.1 arcsec.
+
+    In addition, the energy ranges are ~15-300 keV for imaging and ~10-300 keV for spectra, more information
+    please refer to :cite:t:`Su_asos_hxi_tests_calibrations_2024`.
+
```
nabobalis, line 27: There should be a note about the files having more than one image in them and maybe an example on how to load all them via Map?

sunpy/map/sources/tests/test_asos_hxi_source.py
daozjing: In fact (under my test), if there are multiple HDU units in a FITS file, as long as each HDU contains data and header,
Map will parse them into multiple maps one by one and store them in a list.

For example, for a FITS file with two HDUs, when  

&emsp;  hdus = fits.open(fits_file)

if (hdus[0].data, hdus[0].header) and (hdus[1].data, hdus[1].header) are effective, then the FITS file can be read with Map, e.g.,

&emsp;  _maps_ = Map(fits_file)

here _maps_ is a list, including two maps from hdus[0] and hdus[1], respectively.

sunpy/map/sources/asos.py
daozjing: There is no difference for loading them, except that the return is a Map object for one image or a list of multiple Map objects for more than one images.

sunpy/map/sources/asos.py
daozjing: It works well and is more concise.

sunpy/map/sources/asos.py
daozjing: I'm not sure, in fact, line 69 can be removed, at least for now.

sunpy/map/sources/asos.py
daozjing: Yes. For date strings, such as '01-May-23 13:08:16.962' or ' 1-May-2023 13:08:16.962' (the latter one exported from SSWIDL's map2fits), parse_time will raise ValueError.

sunpy/map/sources/asos.py
daozjing: If someone uses SSWIDL's map2fits to export a FITS file, the energy_l and energy_h will be lost.

sunpy/data/test/hxi_imgcube_01e02t_20230501_130758_HXI_CLEAN_0.header
daozjing: It may contain one HDU or more than one HDUs, depending on the user exporting FITS from the HXI analysis software.

sunpy/map/sources/asos.py
daozjing: Yes. For date strings, such as '01-May-23 13:08:16.962' or ' 1-May-2023 13:08:16.962' (the latter one exported from SSWIDL's map2fits), parse_time will raise ValueError.

sunpy/map/sources/asos.py
nabobalis: I wonder how hard it would be to add support for that time format in parse_time.

sunpy/map/sources/asos.py
nabobalis: I was always under the impression that Map only loads the first HDU and stops.

daozjing: This is attributed to the _map_factory.py_. The _\_parse_arg()_ function can parse FITS file to data_header_pairs=[(data1,header1),(data2,header2),...], then each (data,header) pair in the data_header_pairs list will be processed one by one in _\_\_call\_\_()_.

nabobalis:
Can you also please add a entry to the 7.1.rst under docs/whatsnew? You can check I think 7.0 for an example of adding an entry for a new map source.

nabobalis:
You will also need to add `ASO` to the list of allowed words under https://github.com/daozjing/sunpy/blob/asos_source/_typos.toml to fix the spell check on the pre-commit.

sunpy/map/sources/asos.py
nabobalis: What would be the common use case to use SSWIDL's map2fits?

daozjing: In my experience, fits2map and map2fits are very general and convenient tools for converting between FITS and Map in the SSWIDL ecosystem.

Some users may be accustomed to read a Map with fits2map and do some processing, and then save the Map to FITS with map2fits.

daozjing:
> Can you also please add a entry to the 7.1.rst under docs/whatsnew? You can check I think 7.0 for an example of adding an entry for a new map source.

Done.

nabobalis:
Thank you @daozjing. Sorry for the rush and lots of review comments. There is a release of 7.1 coming soon and Map sources (outside of example pull requests) always generate a ton of comments as there are fixed ways of creating them that isn't really documented anywhere.

sunpy/map/sources/asos.py
nabobalis: @sunpy/sunpy-maintainers thoughts?

sunpy/map/sources/asos.py
nabobalis: While I don't disagree, I am not sure if Map was designed to open FITS files if a user has mangled up the header compared to how a mission provides it directly. I guess there is no harm in it.

daozjing:
> Thank you @daozjing. Sorry for the rush and lots of review comments. There is a release of 7.1 coming soon and Map sources (outside of example pull requests) always generate a ton of comments as there are fixed ways of creating them that isn't really documented anywhere.

Thank you @nabobalis for your patience and meticulousness in helping optimize this PR. This is my first time making a PR and I have learned a lot.

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,128 @@
...
+            return  get_earth(self.reference_date)
+
+    def _get_cmap_name(self):
+        return "rhessi"
+
+    def _rotation_matrix_from_crota(self):
+        """
+        Rotation from CROTA.
+        """
+        return super()._rotation_matrix_from_crota(crota_key='CROTA')
```
Cadair: This doesn't match the test headers, they seem to have `CROTA1` and `CROTA2` which means the default of using `CROTA2` should work? i.e. you don't need this method if the test headers are correct.

```suggestion
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,128 @@
...
+        return SpatialPair(*ctypes)
+
+    @property
+    def date_start(self):
+        """
+        DATE-OBS is the start_date.
+        Here do some thing to deal with date string like '01-May-23 13:08:16.962' or
+        ' 1-May-2023 13:08:16.962' (the latter one is output from SSWIDL's map2fits)
+        """
+        date_obs = self.meta.get('DATE-OBS').strip()
```
Cadair, line 90: I suggest the first thing this method should do is see if the format parses with `parse_time`, so that if a file is "fixed" then it still works.

sunpy/map/sources/asos.py
Cadair: It appears unambiguous, so it should be possible to add a parser for it. If ssw writes things like this then that's a good enough reason to me to add it.

sunpy/map/sources/asos.py
Cadair: Yeah, that's not really a supported use case, but it's good to be defensive in how we write these things anyway.

sunpy/data/test/hxi_imgcube_01e02t_20230501_130758_HXI_CLEAN_0.header
Cadair, line 1: So that I understand, how are these headers made from the mission supplied data? i.e. if I wanted to recreate these from scratch what would I have to do?

sunpy/map/sources/tests/test_asos_hxi_source.py
```diff
@@ -0,0 +1,90 @@
...
+
+from sunpy.data.test import get_dummy_map_from_header, get_test_filepath
+from sunpy.map.sources.asos import HXIMap
+from .helpers import _test_private_date_setters
+
+
+@pytest.fixture(scope="module")
+def hxi_map():
+    ms = [get_dummy_map_from_header(get_test_filepath(f"hxi_imgcube_01e02t_20230501_130758_HXI_CLEAN_{i}.header")) for i in range(2)]
+    return ms
```
Cadair: A better way to do this would be this:

```suggestion
@pytest.fixture(scope="module", params=[0,1])
def hxi_map(request):
    return get_dummy_map_from_header(get_test_filepath(f"hxi_imgcube_01e02t_20230501_130758_HXI_CLEAN_{request.param}.header"))
```

This will make a [parameterized fixture](https://docs.pytest.org/en/stable/how-to/fixtures.html#parametrizing-fixtures) which means that each test which uses it will get executed multiple times.

This means that all the tests can have the for loops removed and act like they are working on a single header (but all will get run for each header).

sunpy/data/test/hxi_imgcube_01e02t_20230501_130758_HXI_CLEAN_0.header
daozjing: I think one method is, firstly to export an image FITS file using HXI data analysis software, then read it as hdu using astropy.io.fits. Finally, the .header file can be exported by hdu.header.totextfile().

sunpy/map/sources/asos.py
daozjing: I added a try... except... to do this.

sunpy/data/test/hxi_imgcube_01e02t_20230501_130758_HXI_CLEAN_0.header
Cadair: Please can you add some links to the documentation of this map source about the analysis software that produces the data? I'd like it to be clear to people where the data this map source is for comes from.

docs/whatsnew/7.1.rst
```diff
@@ -64,6 +64,11 @@ Added support for querying data from `solarnet API <https://solarnet.oma.be/>`__
     <BLANKLINE>
     <BLANKLINE>
 
+Added support for new observatories
+===================================
+
+Added a new map source for the ASO-S mission, currently added `~sunpy.map.sources.HXIMap` for the Hard X-ray Imager (HXI) payload, which enables the loading of the hard X-ray imaging FITS file that was exported from the HXI analysis software.
```
Cadair, line 70: Could you make an example plot to include here? It would be nice to have a visual demonstration.

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,125 @@
...
+        try:
+            return parse_time(date_obs)
+        except ValueError:
+            tmp = date_obs.split(" ")[0]
+            if len(tmp)==9 and len(tmp.split("-")[-1])==2:
+                date_obs = date_obs[:7]+'20'+date_obs[7:]
```
Cadair:
```suggestion
            if len(tmp) == 9 and len(tmp.split("-")[-1]) == 2:
                date_obs = date_obs[:7] + '20'+date_obs[7:]
```

docs/whatsnew/7.1.rst
daozjing: Yes, I would be happy to add an example. To do this, it seems that I should firstly add a FITS file to the SunPy Data Repository (https://github.com/sunpy/data) ?

sunpy/data/test/hxi_imgcube_01e02t_20230501_130758_HXI_CLEAN_0.header
daozjing: Yes, I can add some links and explanations under the HXIMap class.

docs/whatsnew/7.1.rst
Cadair: For the what's new we normally just include a png in the notes.

Do you want to add an example to the example gallery as well? We generally think carefully about adding more sample data, but we can do if we want to.

daozjing: I have made a PR to add a FITS file (about 60KB) using git lfs, but I am not sure if it will be approved.
However, I can first complete the current task by making a png for the what's new.

nabobalis: The data repo for sunpy is mostly focused on examples which are added to the gallery. We use it to reduce network calls and prevent server outages from breaking the documentation build where possible.

Is the actual data only 60KB? That seems very small.

daozjing: Yes, for the current example.

sunpy/map/sources/tests/test_asos_hxi_source.py
```diff
@@ -0,0 +1,79 @@
...
+    # in the constructor
+    new_hxi_map = hxi_map._new_instance(hxi_map.data, hxi_map.meta)
+    assert new_hxi_map.plot_settings['norm'].vmin is None
+    assert new_hxi_map.plot_settings['norm'].vmax is None
+    assert new_hxi_map.plot_settings['cmap'] == new_hxi_map._get_cmap_name()
+
+
+def test_wcs(hxi_map):
+    # Smoke test that WCS is valid and can transform from pixels to world coordinates
+        hxi_map.pixel_to_world(0*u.pix, 0*u.pix)
```
Cadair:
```suggestion
    hxi_map.pixel_to_world(0*u.pix, 0*u.pix)
```

sunpy/map/sources/tests/test_asos_hxi_source.py
```diff
@@ -0,0 +1,79 @@
...
+def test_new_instance_preserves_plot_settings(hxi_map):
+    # Tests that the _new_instance method preserves the plot_settings
+    # of the old instance. This is done on the HXI source as the HXIMap
+    # constructor explicitly sets the cmap and norm and we want to test
+    # that _new_instance persists the old custom plot_settings
+    hxi_map.plot_settings['norm'] = ImageNormalize(vmin=0.1, vmax=42)
+    hxi_map.plot_settings['cmap'] = 'inferno'
+    new_hxi_map = hxi_map._new_instance(hxi_map.data,
+                                              hxi_map.meta,
+                                              plot_settings=hxi_map.plot_settings)
```
Cadair:
```suggestion
    new_hxi_map = hxi_map._new_instance(hxi_map.data,
                                        hxi_map.meta,
                                        plot_settings=hxi_map.plot_settings)
```

sunpy/map/sources/tests/test_asos_hxi_source.py
```diff
@@ -0,0 +1,79 @@
...
+    assert hxi_map.date.isot in ["2023-05-01T13:07:58.713", "2023-05-01T13:08:16.962"]
+
+
+def test_private_date_setters(hxi_map):
+    _test_private_date_setters(hxi_map)
+
+
+def test_measurement(hxi_map):
+    """Tests the measurement property of the HXIMap object."""
+    assert hxi_map.measurement in [[20, 30], [20, 30]] * u.keV
```
Cadair: This comparison is weird, it seems to be testing if it's one of `[20, 30]*u.keV` or `[20, 30]*u.keV` which are the same thing?

I think this would be better?
```suggestion
    assert u.allclose(hxi_map.measurement, [20, 30] * u.keV)
```

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,140 @@
...
+        else:
+            return waves*self.waveunit
+
+    @property
+    def observatory(self):
+        return self.meta.get('observatory', 'ASO-S')
+
+    @property
+    def instrument(self):
+        return self.meta.get('INSTRUME', 'HXI')
```
Cadair: Just for consistency:

```suggestion
        return self.meta.get('instrume', 'HXI')
```

Cadair (approved):
A few minor tweaks but I think this is ready.

Thank you very much for your contribution @daozjing :tada:

docs/whatsnew/7.1.rst
Cadair: I think we should postpone a new gallery example & associated sample data to a followup PR. @daozjing if you want to add something to the example gallery using this data then we will probably add something to the data repo to go with that.

sunpy/map/sources/tests/test_asos_hxi_source.py
daozjing: For the two current test files, the hxi_map.measurement values are the same.

docs/whatsnew/7.1.rst
daozjing: Thanks. I also think this can be another follow-up task.

sunpy/map/sources/asos.py
```diff
@@ -0,0 +1,140 @@
...
+from sunpy.coordinates import get_earth
+from sunpy.map.mapbase import GenericMap, SpatialPair
+from sunpy.time import parse_time
+
+__all__ = ['HXIMap']
+
+
+class HXIMap(GenericMap):
+    """
+    HXI Image Map Source.
```
hayesla, line 16: Are these FITS files supported here from the output of the HXI image construction software (i.e. a user would first have to construct an image) or does the HXI team provide these FITS? its fine either way, but maybe it should be explicitly mentioned here

Cadair: It's the output of some post-processing. After a conversation we put some documentation here: https://github.com/sunpy/sunpy/pull/8398/files#diff-7561363d2e75732cbd704c7ba3c8d42fdb1f788502e52ed97745fc153f77cc9aR28-R34

samaloney (approved):
This looks good.

I wonder if we should also override the `.unit` attribute so it return an quantity from the unit string from the header `UNIT    = 'photons cm-2 s-1 asec-2'` this won't parse as is so would need to fix it `'photon cm-2 s-1 arcsec-2'`

Cadair:
Merged as https://github.com/sunpy/sunpy/commit/bea1cb85da5db38b511b4eb1e01c3b82e717a2aa

Cadair:
Thanks a lot for this PR @daozjing sorry to have merged this in a different PR, but I'm trying to make the 7.1 release today, so I wanted to get this in.

Could you please comment over on #8410 about your affiliation for the author list? Thanks!

PR #8402, opened 2025-11-15
Enable `pytest-xdist` for figure tests

ConorMacBride:
Hi, `pytest-mpl` `0.18.0` now supports generating HTML summaries for the figure tests when using `pytest-xdist`


ConorMacBride:
`pytest-mpl` appears to be working as expected here. I set it to 4 workers as `auto` created 40+.

Cadair:
Thanks!

PR #8404, opened 2025-11-17
Support `EarthLocation` as observer in `make_fitswcs_header`

### LINKED ISSUE #8326: Add support for generating OBSGEO keys for Ground Based Observers in fits header helper
### Describe the feature

If you have a ground based observer the "most correct" observer location keys are the OBSGEO-[XYZ] triplet. We should support a `EarthLocation` instance as an input to `make_fitswcs_header` and set these keys in the header rather than HGLN/T ones.

### Proposed solution

_No response_

aadya940:
Closes #8326 .

### DIFF
--- a/sunpy/map/header_helper.py
+++ b/sunpy/map/header_helper.py
@@ -2,7 +2,7 @@ import numpy as np
 
 import astropy.units as u
 import astropy.wcs
-from astropy.coordinates import SkyCoord, EarthLocation
+from astropy.coordinates import EarthLocation, SkyCoord
 
 from sunpy import log
 from sunpy.coordinates import frames, sun
@@ -128,7 +128,7 @@ def make_fitswcs_header(data,
         if isinstance(observer, EarthLocation):
             observer_meta = get_earth_observer_meta(observer)
             meta_wcs.update(observer_meta)
-            
+
             observer_hgs = observer.get_itrs(obstime=coordinate.obstime).transform_to(
                 frames.HeliographicStonyhurst(obstime=coordinate.obstime)
             )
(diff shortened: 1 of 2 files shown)


sunpy/map/header_helper.py
```diff
@@ -75,6 +77,9 @@ def make_fitswcs_header(data,
         Units of the array data of the Map. This will populate the the ``'bunit'`` meta keyword.
         If ``data`` is a `~astropy.units.Quantity`, the unit specified here will take precedence
         over the unit information attached to ``data``.
+    observer : `~astropy.coordinates.EarthLocation`, optional
+        Ground-based observer location. If provided, OBSGEO-X/Y/Z keywords will be set
+        in the header instead of heliographic observer keywords.
```
Cadair: I don't think this statement is true (and even if it is you should have a test for it). I think if you pass in a coordinate with an observer and the `observer=` keyword then it will put both in.

I'm not sure what the best way to handle this conflict is. I don't think that disallowing `coordinate` to have an observer is a great idea, as it's kinda hard to remove one if you have got a coordinate from somewhere else. Perhaps the best solution would be to detect the case and issue a `SunPyUserWarning` saying that `coordinate.observer` is being ignored in favour of `observer=`.

aadya940: Ah, missed that case!

sunpy/map/tests/test_header_helper.py
```diff
@@ -321,3 +322,40 @@ def test_make_hpr_header(hgs_coord):
...
+    data = np.zeros((512, 512))
+
+    # Conflict case.
+    with pytest.warns(SunpyUserWarning, match="coordinate.observer.*ignoring"):
+        header = make_fitswcs_header(data, coord_with_observer, observer=greenwich)
+
+    assert 'obsgeo-x' in header
+    assert 'obsgeo-y' in header
+    assert 'obsgeo-z' in header
+    assert 'rsun_obs' in header
```
Cadair, line 363: You should add asserts here that `HGLN-OBS` and `HGLT-OBS` are not in the header.

sunpy/map/header_helper.py
```diff
@@ -119,8 +127,27 @@ def make_fitswcs_header(data,
...
+            meta_wcs.update(observer_meta)
+
+            observer_hgs = observer.get_itrs(obstime=coordinate.obstime).transform_to(
+                frames.HeliographicStonyhurst(obstime=coordinate.obstime)
+            )
+            dsun_obs = observer_hgs.radius
+            meta_wcs['rsun_obs'] = sun._angular_radius(coordinate.rsun, dsun_obs).to_value(u.arcsec)
+        else:
+            raise ValueError("`observer` parameter must be an `EarthLocation` instance.")
+    elif getattr(coordinate, 'observer', None) is not None:
```
Cadair, line 151: I *think* (I haven't actually checked) that it's the `_get_wcs_meta` function on line 126 which adds the observer coordinate information.

aadya940: Yes thats correct, its mentioned in the docstrings I fixed that via removing `observer=` from the `coordinate` kwarg.

PR #8407, opened 2025-11-18
Pixel support for shift_reference_coord

nabobalis:
Goal here is that by adding this to sunpy, we can avoid having to do the hard stuff in sunkit-image regarding pixel shifts.

No idea if this is even correct. 

If this is good, having it in 7.1 would be helpful. 

- [x] Fix code
- [x] Add changelog

### DIFF
--- a/changelog/8407.feature.rst
+++ b/changelog/8407.feature.rst
@@ -1,0 +1,1 @@
+Added support for pixel inputs to `~sunpy.map.GenericMap.shift_reference_coord`.


sunpy/map/mapbase.py
```diff
@@ -1169,37 +1169,41 @@ def center(self):
         return self.wcs.pixel_to_world(*center)
 
     @u.quantity_input
-    def shift_reference_coord(self, axis1: u.deg, axis2: u.deg):
+    def shift_reference_coord(self, axis1, axis2):
```
Cadair, line 1171: I think you can specify multiple units with `quantity_input` somehow?

nabobalis: Added.

sunpy/map/mapbase.py
```diff
@@ -1167,38 +1167,42 @@ def center(self):
...
             A new shifted Map.
         """
         new_meta = self.meta.copy()
-
-        # Update crvals
+        # If we are given pixel we turn them into physical units to shift CRVAL1/2
+        if axis1.unit.is_equivalent(u.pix):
+            axis1 = axis1*self.scale[0]
+        if axis2.unit.is_equivalent(u.pix):
+            axis2 = axis2*self.scale[1]
```
Cadair, line 1202: This isn't going to always be correct, for example non-square pixels or a skew.

More over, I'm not actually sure we want to enable this, surely if you want to shift in pixels you want to shift the reference pixel and not the reference coordinate? @wtbarnes said this was motivated by the alignment work in sunkit-image? What functionality do you actually need?

nabobalis: If I want to shift the reference coordinate but I know the offset in pixels, what should one do?

Cadair: Why wouldn't you change the reference pixel?

I don't know if there's an easy generic answer to that question.

nabobalis: Is that the same operation?

nabobalis: I guess I just wonder, in the coalignment API we work out the pixel shift between two maps. What is the best way to update the pointing for that? 

Do we update the reference pixel, the reference coordinate or something else?

Cadair: They are very subtly different but for coalignment you should be updating the reference pixel.

Cadair:
@nabobalis / @wtbarnes Do you still need this or something like it for sunkit-image? Or should we close this or what do you want to do?

wtbarnes:
My opinion is close it. This wont make or break a sunkit-image release and I'm fairly sure this isn't actually what we need anyway.

PR #8409, opened 2025-11-20
Propagate `JSOCClient.fetch` kwargs down to `drms`

wtbarnes:
Previously, we were not passing `sleep` from `JSOCClient.fetch` down to `drms.ExportRequest.wait` so it had no effect there. This PR fixes that and also propagates the `timeout` and `retries_notfound` kwargs.

I would consider this a bugfix as the docstring for `fetch`

### DIFF
--- a/changelog/8409.feature.rst
+++ b/changelog/8409.feature.rst
@@ -1,3 +1,3 @@
 Pass ``sleep``, ``timeout``, and ``retries_notfound`` from :meth:`~sunpy.net.jsoc.JSOCClient.fetch`
-to :meth:`drms.ExportRequest`. Previously, ``sleep`` was a keyword to :meth:`~sunpy.net.jsoc.JSOCClient.fetch`
+to :meth:`drms.ExportRequest.wait`. Previously, ``sleep`` was a keyword to :meth:`~sunpy.net.jsoc.JSOCClient.fetch`
 but was not being propagated correctly.


sunpy/net/jsoc/jsoc.py
```diff
@@ -369,7 +369,8 @@ def request_data(self, jsoc_response, method='url', **kwargs):
 
     @convert_row_to_table
     def fetch(self, jsoc_response, path=None, progress=True, overwrite=False,
-              downloader=None, wait=True, sleep=10, max_conn=default_max_conn, **kwargs):
+              downloader=None, wait=True, sleep=10, timeout=None, retries=5,
+              max_conn=default_max_conn, **kwargs):
```
Cadair: You shouldn't change the ordering of kwargs which can also be positional arguments. You should add the new kwargs right before **kwargs

Cadair:
This seems fine, but it's a bit of a stretch to be a bugfix what with adding new kwargs.

wtbarnes:
Fair, though it seems silly to have a separate PR just for two extra keywords. I'm fine not labeling it a bugfix. It doesn't really need a backport.

sunpy/net/jsoc/jsoc.py
wtbarnes: Whoops. Done.

changelog/8409.feature.rst
```diff
@@ -0,0 +1,3 @@
+Pass ``sleep``, ``timeout``, and ``retries_notfound`` from :meth:`~sunpy.net.jsoc.JSOCClient.fetch`
+to :meth:`drms.ExportRequest`. Previously, ``sleep`` was a keyword to :meth:`~sunpy.net.jsoc.JSOCClient.fetch`
```
Cadair:
```suggestion
to :meth:`drms.ExportRequest.wait`. Previously, ``sleep`` was a keyword to :meth:`~sunpy.net.jsoc.JSOCClient.fetch`
```

samaloney (approved):

PR #8420, opened 2025-11-26
Fix tests for astropy 7.2

Cadair:
For the record I bisected this back to this commit in astropy: https://github.com/astropy/astropy/commit/1ba18e8ace31b168d284a4104a4ed04b94ef9a44

### DIFF
--- a/sunpy/map/tests/test_map_factory.py
+++ b/sunpy/map/tests/test_map_factory.py
@@ -13,6 +13,7 @@ from astropy.wcs import WCS
 import sunpy
 import sunpy.map
 from sunpy.data.test import get_dummy_map_from_header, get_test_data_filenames, get_test_filepath, rootdir
+from sunpy.io._fits import read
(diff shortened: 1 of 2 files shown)


sunpy/map/tests/test_map_factory.py
```diff
@@ -362,16 +362,6 @@ def test_map_jp2_HMI():
     plt.colorbar()
 
 
-def test_map_fits():
```
Cadair, line 365: I removed this as it's very much a duplicate of the other test I fixed. If people think it's worthwhile I could modify this test to verify that the memmap kwarg makes it down to the io layer?

hayesla (approved):

pllim:
Thanks! p.s. Why didn't sunpy catch this earlier? Does sunpy not test against astropy nightly wheel?

Cadair:
I think we just weren't on top of it, we have #8429 now though :)

PR #8429, opened 2025-11-27
Attempt to open issues based on devdeps

Cadair:
This is an experiment so we notice devdep issues more.

It seems to work based on this commit: https://github.com/sunpy/sunpy/actions/runs/19741600045/job/56567049378 but it can't open an issue from a PR.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -55,7 +55,7 @@ jobs:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
   open-issue:
-    if: failure()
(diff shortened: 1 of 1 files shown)


pllim:
My colleagues and I tried something similar in the past (not using GHA but same concept). What we ended up with is the bot spamming us with issues about the same failure when we can't get to it fast enough.

I didn't read the code carefully on a phone. Does this prevent duplicate issues?

Cadair:
yes, it edits any existing issue with the same title.

pllim:
Exciting. Hope it works!

Cadair:
@meeseeksdev backport to 7.1

PR #8441, opened 2025-12-02
Add a cooldown for dependabot updates

namurphy:
A [dependency cooldown](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns) is "a window of time between when a dependency is published and when it’s considered suitable for use." Because most vulnerabilities are discovered with a few days to a week after the release, dependency cooldowns are an easy way to mitigate the majority of supply chain attacks. The pull request updates `.github/dependabot.yml` adds a [cooldown period for dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/optimizing-pr-creation-version-updates#setting-up-a-cooldown-period-for-dependency-updates) updates of 7 days.  ❄️

Related to #8134.


namurphy:
Superseded by #8545.

PR #8442, opened 2025-12-03
Bugfix net dependency on map

samaloney:
Move the map/coordinates related imports inline into the `Cutout` attr as only required here.

### DIFF
--- a/sunpy/net/jsoc/attrs.py
+++ b/sunpy/net/jsoc/attrs.py
@@ -180,7 +180,6 @@ class Cutout(DataAttr):
             # import here so net won't depend on map
             from sunpy.map.maputils import coordinate_is_on_solar_disk
             if not coordinate_is_on_solar_disk(center):
-
                 raise ValueError("Tracking is enabled, but the center of the cutout "
                                  f"(Tx={center_x}, Ty={center_y}) is not on the solar disk.")
 


sunpy/net/jsoc/attrs.py
```diff
@@ -168,6 +165,12 @@ class Cutout(DataAttr):
     def __init__(self, bottom_left, top_right=None, width: u.arcsec = None,
                  height: u.arcsec = None, tracking=False, register=False,
                  nan_off_limb=False):
+
+        # import here so net won't depend on map
+        from sunpy.coordinates.frames import Helioprojective
+        from sunpy.coordinates.utils import get_rectangle_coordinates
```
Cadair: Do we need to move coordinates? the coordinate submodule is "core" right it always works?

sunpy/net/jsoc/attrs.py
```diff
@@ -168,6 +165,12 @@ class Cutout(DataAttr):
...
+        from sunpy.map.maputils import coordinate_is_on_solar_disk
```
Cadair: If we wanted to we could move this import inside a `if tracking` block to reduce runtime if you aren't using tracking?

sunpy/net/jsoc/attrs.py
samaloney: I guess I was being cautious will only move the map related import.

Cadair:
This should have a changelog because of backports.

sunpy/net/jsoc/attrs.py
```diff
@@ -180,9 +176,13 @@ def __init__(self, bottom_left, top_right=None, width: u.arcsec = None,
...
         center = SkyCoord(center_x, center_y, frame=bottom_left.frame)
-        if tracking and not coordinate_is_on_solar_disk(center):
-            raise ValueError("Tracking is enabled, but the center of the cutout "
-                             f"(Tx={center_x}, Ty={center_y}) is not on the solar disk.")
+        if tracking:
+            # import here so net won't depend on map
+            from sunpy.map.maputils import coordinate_is_on_solar_disk
+            if not coordinate_is_on_solar_disk(center):
+
+                raise ValueError("Tracking is enabled, but the center of the cutout "
```
Cadair, line 183:
```suggestion
            if not coordinate_is_on_solar_disk(center):
                raise ValueError("Tracking is enabled, but the center of the cutout "
```

samaloney:
The test fails seem unrelated to me

Cadair:
@meeseeksdev backport to 7.0

ayshih:
Incidentally, that function should be under `sunpy.coordinates` rather than `sunpy.map`.  I've created #8445.

PR #5825, opened 2022-01-19
Set symmetric colorbar limits on HMI maps

dstansby:
We already do this on MDI maps, so it seems sensible to do it for HMI maps too.

xref https://github.com/sunpy/sunpy/issues/5637

PR #8447, opened 2025-12-03
Use `CenteredNorm` for magnetic-field maps

ayshih:
To have a symmetric plotting range for magnetic-field maps of HMI and MDI, we currently manually set the `vmin` and `vmax` on the norm (e.g., #5825), but that requires us to read in the data array during instantiation.  I just found out that `matplotlib.colors.CenteredNorm` has existed for many years, so is the far better approach, because then the plotting range is determined only when the data is actually being plotted.

### DIFF
--- a/changelog/8447.bugfix.rst
+++ b/changelog/8447.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug with HMI and MDI magnetic-field maps where the data array would always be read during map instantiation.
--- a/sunpy/map/mapbase.py
+++ b/sunpy/map/mapbase.py
@@ -605,14 +605,6 @@ class GenericMap(NDData):
     def _meta_hash(self):
         return self.meta.item_hash()
 
-    def _set_symmetric_vmin_vmax(self):
-        """
-        Set symmetric vmin and vmax about zero
-        """
-        threshold = np.nanmax(abs(self.data))
-        self.plot_settings['norm'].vmin = -threshold
-        self.plot_settings['norm'].vmax = threshold
-
     @property
     @cached_property_based_on('_meta_hash')
     def wcs(self):
--- a/sunpy/map/sources/sdo.py
+++ b/sunpy/map/sources/sdo.py
@@ -1,6 +1,7 @@
 """SDO Map subclass definitions"""
 
 import numpy as np
+from matplotlib.colors import CenteredNorm
 
 import astropy.units as u
 from astropy.coordinates import CartesianRepresentation, HeliocentricMeanEcliptic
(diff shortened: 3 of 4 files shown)


ayshih:
I have confirmed that this PR fixes when RAM is used by Maps that are able to be memory mapped (e.g., HMI).  The Virtual Memory Size (VMS) should always go up during instantiation because the data array can now be addressed, so the question is whether the Resident Set Size (RSS) increases during instantiation as well.

### Before this PR
```python
>>> import psutil

>>> import sunpy.map
>>> from sunpy.data.sample import AIA_171_IMAGE, HMI_LOS_IMAGE

>>> # Initial memory usage
>>> process = psutil.Process()
>>> print(f"RSS: {process.memory_info().rss / 2**20} MB, VMS: {process.memory_info().vms / 2**20} MB")
RSS: 239.68359375 MB, VMS: 197.84765625 MB

>>> # AIA maps are unable to be memory mapped, so both RSS and VMS go up
>>> aia = sunpy.map.Map(AIA_171_IMAGE)
WARNING: VerifyWarning: Invalid 'BLANK' keyword in header.  The 'BLANK' keyword is only applicable to integer data, and will be ignored in this HDU. [astropy.io.fits.hdu.image]
>>> print(f"RSS: {process.memory_info().rss / 2**20} MB, VMS: {process.memory_info().vms / 2**20} MB")
RSS: 244.15625 MB, VMS: 202.86328125 MB

>>> # HMI maps are able to be memory mapped, so RSS should not go up as much as VMS goes up
>>> hmi = sunpy.map.Map(HMI_LOS_IMAGE)
>>> print(f"RSS: {process.memory_info().rss / 2**20} MB, VMS: {process.memory_info().vms / 2**20} MB")
RSS: 252.5 MB, VMS: 211.41796875 MB
```

### After this PR
```python
>>> import psutil

>>> import sunpy.map
>>> from sunpy.data.sample import AIA_171_IMAGE, HMI_LOS_IMAGE

>>> # Initial memory usage
>>> process = psutil.Process()
>>> print(f"RSS: {process.memory_info().rss / 2**20} MB, VMS: {process.memory_info().vms / 2**20} MB")
RSS: 240.86328125 MB, VMS: 199.66796875 MB

>>> # AIA maps are unable to be memory mapped, so both RSS and VMS should go up
>>> aia = sunpy.map.Map(AIA_171_IMAGE)
WARNING: VerifyWarning: Invalid 'BLANK' keyword in header.  The 'BLANK' keyword is only applicable to integer data, and will be ignored in this HDU. [astropy.io.fits.hdu.image]
>>> print(f"RSS: {process.memory_info().rss / 2**20} MB, VMS: {process.memory_info().vms / 2**20} MB")
RSS: 245.15625 MB, VMS: 203.68359375 MB

>>> # HMI maps are able to be memory mapped, so RSS should not go up as much as VMS goes up
>>> hmi = sunpy.map.Map(HMI_LOS_IMAGE)
>>> print(f"RSS: {process.memory_info().rss / 2**20} MB, VMS: {process.memory_info().vms / 2**20} MB")
RSS: 245.359375 MB, VMS: 211.9765625 MB
```

PR #8452, opened 2025-12-03
Improvements to Error and test skipping when ANA extension not built

Cadair:


sunpy/io/ana.py
```diff
@@ -35,9 +36,13 @@ def check_ana_installed(func):
         "C extension for ANA is missing. For more details see: "
         "https://docs.sunpy.org/en/stable/installation.html#installing-without-conda"
     )
-    if _pyana is None and os.environ.get("SUNPY_NO_BUILD_ANA_EXTENSION") is None:
```
Cadair, line 38: Nabil added this to fix the source dist build, but I don't *think* we need it now I've moved the import error to function call time rather than import time.

PR #8459, opened 2025-12-04
Added a test for successful map memory mapping

ayshih:
After #8447, we can now test that our HMI sample image is memory mapped as intended.  I originally tried to write this test using our HMI test image, but its data array is much smaller (80 kB instead of 8 MB), so the test would give confusing/unreliable results as Python managed memory in parallel.

This PR adds `psutil` as a test dependency

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,102 @@
+7.1.0 (2025-12-08)
+==================
+
+Breaking Changes
+----------------
+
+- Increased minimum versions for these dependencies:
+
+  - Python >= 3.12
+  - astropy >= 6.1.0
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -110,6 +110,7 @@ jupyter = [
 tests-only = [
   "hvpy>=1.1.0",
   "jplephem>=2.19",
+  "psutil",
```
Cadair: you need to add a minimum version for oldestdeps to work

ayshih: Okay.  I picked 6.0.0 (released in Jun 2024) because it looks like 5.9.0 released in Dec 2021

PR #8468, opened 2025-12-17
Moved `towncrier` from conda to pip in dev env YAML

ayshih:
For our development environment YAML, `towncrier` is installed by conda, while `sphinx-changelog` has to be installed by pip.  Since `sphinx-changelog` pins the version of `towncrier`, it means that the conda-installed `towncrier` gets uninstalled and then the pinned version of `towncrier` gets installed by pip.  This PR moves `towncrier` from the conda section to the pip section to avoid this reinstallation, and this has no impact to anything else because `towncrier` is not a dependency of any of our conda-installed dependencies.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -111,6 +111,7 @@ jobs:
           activate-environment: sunpy-test
           environment-file: sunpy-dev-env.yml
           python-version: "3.13"
+          conda-remove-defaults: "true"
       - name: Install sunpy
         shell: bash -el {0}
         run: |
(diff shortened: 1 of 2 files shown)


Cadair (approved):

samaloney (commented):
If `sphinx-changelog` installs `towncrier` do we need a explicit install of `towncrier`? I guess it might be clearer if issues crop up.

ayshih:
I just noticed we haven't been disabling the `defaults` conda channel for the CI run, so I've now done that too

ayshih:
> If `sphinx-changelog` installs `towncrier` do we need a explicit install of `towncrier`?

It doesn't have to be explicitly listed, and in fact it's not in our `pyproject.toml`.  However, I'd like to keep it listed in the YAML to be explicit that it will be installed via `pip` even though it is available via `conda`.

samaloney (approved):

PR #8417, opened 2025-11-24
Revised the time-distance plot example

ayshih:
This PR moves the time-distance plot example (#8057) to the "showcase" category, and fixes a bunch of issues with the example.  Since this example has not been a release yet, I don't think this PR needs its own changelog entry.

PR #8469, opened 2025-12-19
Fixed changelog link to moved gallery example

ayshih:
I forgot to update the link in the changelog entry for the time-distance gallery example when I moved it in #8417


Cadair:
I already did this in #8483

PR #8477, opened 2026-01-07
Ensured that the SDO Map reference date is in the same scale as the observation time

ayshih:
For both `AIAMap` and `HMIMap`, the `.date` property typically comes from `DATE-OBS`, while the `.reference_date` property typically comes from `T_OBS` (unless `T_OBS` has been removed during a processing step).  Potentially confusingly, `DATE-OBS` is in UTC, while `T_OBS` is in TAI.  As `Time` objects, this is not technically a problem, because the time scale is part of the object.  However, because times are almost always in UTC, it is quite common for a user to not be super-diligent about time scales  (e.g., before converting to strings or `datetime` objects), so user code can get tripped up with `.reference_date` being in a different time scale than `.date`.

This PR changes the `.reference_date` property for both `AIAMap` and `HMIMap` so that the returned time is in the same time scale as the `.date` property.  If `.date` happens to be in TAI, then `.reference_date` will also be TAI.  But, typically, this means that `.reference_date` now returns UTC time.  The conversion from TAI to UTC is lossless, and a user can explicitly convert it (back) to TAI time if that's what they want.

xref #7682 and #7758

### DIFF
--- a/changelog/8477.breaking.rst
+++ b/changelog/8477.breaking.rst
@@ -1,0 +1,1 @@
+Adjusted the ``reference_date`` property of `~sunpy.map.sources.AIAMap` and `~sunpy.map.sources.HMIMap` so that its return value is in the same time scale as the ``date`` property, in order to avoid subtle user mistakes.
--- a/sunpy/map/sources/sdo.py
+++ b/sunpy/map/sources/sdo.py
@@ -5,6 +5,7 @@ from matplotlib.colors import CenteredNorm
 
 import astropy.units as u
 from astropy.coordinates import CartesianRepresentation, HeliocentricMeanEcliptic
+from astropy.time import Time
 from astropy.visualization import AsinhStretch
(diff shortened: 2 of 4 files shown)


samaloney (commented):
I guess there maybe some overhead in creating a new Time object would `return getattr(date, self.date.scale, date)` work?

ayshih:
> I guess there maybe some overhead in creating a new Time object would `return getattr(date, self.date.scale, date)` work?

I added a check to avoid unnecessary instantiation

Cadair (approved):

changelog/8477.breaking.rst
Cadair, line 1: we should make this a breaking change so that users are aware.

ayshih: Done

wtbarnes (approved):

PR #8480, opened 2026-01-17
Add CDF time support to parse_time

Omiii-215:
This PR adds native support for Common Data Format (CDF) time types in `sunpy.time.parse_time`.

Specifically, it enables parsing of `CDF_EPOCH`, `CDF_EPOCH16`, and `CDF_TIME_TT2000` values directly into `astropy.time.Time` objects by leveraging `cdflib` when it is available.
Changes

 **`sunpy/time/time.py`**

  Extends `convert_time` to recognize `format` values starting with `cdf_` (`cdf_epoch`, `cdf_epoch16`, `cdf_tt2000`).
  Uses a lazy import of `cdflib` to perform the conversion, with a clear error if it is not installed.

**`sunpy/time/tests/test_cdf_time.py`**

  Adds unit tests covering:

   Scalar and array conversions for `CDF_TIME_TT2000`
   `CDF_EPOCH` conversions
   Proper error handling when `cdflib` is not installed

**`docs/how_to/parse_time.rst`**

  Updates the documentation with a new section explaining how to work with CDF epoch formats.


CHANGELOG.rst
Cadair, line 1: please revert the changes here, I'll fix CI issues in #8483

examples/saving_and_loading_data/cdf_time_conversion.py
```diff
@@ -0,0 +1,45 @@
...
+###############################################################################
+# The Common Data Format (CDF) has a number of standard time formats that are
+# not supported by `astropy.time` by default. These are ``CDF_EPOCH``,
+# ``CDF_EPOCH16`` and ``CDF_TT2000``.
+#
+# SunPy provides a way to convert these times to `~astropy.time.Time` objects
+# using the `cdflib` library.
+
+###############################################################################
+# First we need to make sure that `cdflib` is installed.
```
Cadair: We don't guard imports in the gallery, we generally just put a line like this in telling the reader they need a package: https://github.com/sunpy/sunpy/blob/main/examples/map_transformations/reprojection_align_aia_hmi.py#L14

examples/saving_and_loading_data/cdf_time_conversion.py
Cadair, line 1: I'm not sure we need an example for this. Unless someone comes up with something which is closer to a real workflow (i.e. involving a real CDF file, where the time hasn't already been converted), I'd probably suggest removing this.

sunpy/time/tests/test_cdf_time.py
```diff
@@ -0,0 +1,47 @@
...
+
+from sunpy.time import parse_time
+
+# Critical: Skip entire file if cdflib is missing
+cdflib = pytest.importorskip("cdflib")
+# Also skip if cdflib is present but too old to have epochs_astropy
+try:
+    import cdflib.epochs_astropy  # noqa: F401
+except ImportError:
+    pytest.skip("cdflib installed but too old (missing epochs_astropy)", allow_module_level=True)
```
Cadair: I'd drop this and make sure the min version in our pyproject.toml file has support.

sunpy/time/time.py
```diff
@@ -167,6 +167,13 @@ def find_time(string, format):
 
 @singledispatch
 def convert_time(time_string, format=None, **kwargs):
+    if format in ['cdf_epoch', 'cdf_epoch16', 'cdf_tt2000']:
+        try:
+            from cdflib.epochs_astropy import CDFAstropy
+        except ImportError:
+            raise ImportError("cdflib must be installed to support CDF time formats.")
```
Cadair:
```suggestion
        except ImportError as err:
            raise ImportError("cdflib must be installed to support CDF time formats.") from err
```

Cadair (changes requested):
Hi @Omiii-215 thanks for your interest in contributing to sunpy.

I'd like you to bear in mind our policy on output from LLMs in future comments: https://github.com/sunpy/sunpy?tab=readme-ov-file#usage-of-generative-ai

I think that this contribution is probably valid, but I don't know much about CDF times, so I'd like to see a more "real world" example of why someone might want to parse a time like this to `parse_time` perhaps @hayesla or @jgieseler has some ideas on that?

examples/saving_and_loading_data/cdf_time_conversion.py
Omiii-215: I agree with this the example isn’t a true end-to-end workflow. I added it mainly to try and give users a illustration  the new parse_time support based on common cdflib usage but it won't reflect accuracy, I'll Remove it.

ayshih (changes requested):
You are critically missing any unit tests that confirm that the returned `Time` objects convert to the correct values in other formats (e.g., ISO).  As it stands, your unit tests would pass even if definitions are internally screwed up.

Omiii-215:
@ayshih ohh okay, Ig the current tests only validate round tripping cause I cross verified ? within the same format and don’t verify correctness across the definitions. I can add tests that convert the parsed CDF times to ISO (and possibly JD) and will try compare against known values if you have a preferred approach or reference values pls recommend also @hayesla and @Cadair if you could help in ? maybe adding the ISO `t.isot == 2000-01-01T12:00:00.000` format as mentioned?

jgieseler:
> I think that this contribution is probably valid, but I don't know much about CDF times, so I'd like to see a more "real world" example of why someone might want to parse a time like this to `parse_time` perhaps @hayesla or @jgieseler has some ideas on that?

Hi all! I was also wondering if there was a specific use-case motivating this PR. Usually when loading CDF files, I directly use `cdflib.epochs.CDFepoch.to_datetime()` to obtain the EPOCH values as datetime. 

A "real world" example for using `sunpy.time.parse_time` to work with CDF times that I could think of would e.g. be that someone who is not familiar with CDF files gets some CDF data sent as an ASCII file, and now they want to read this in. Without knowing anything about cdflib and CDF formats, they could just dumb the time column to `parse_time` and start using the data.

PR #8482, opened 2026-01-18
 Make map factory tests robust against expected warnings

vishwas-droid:
While running the map factory tests, I noticed that a couple of expected
warnings were being emitted on newer Python versions.

Instead of suppressing them globally, this change handles the warnings
explicitly within the test, which makes the intent clearer and avoids
future CI issues as Python and Astropy evolve.

No functional behavior is changed.


vishwas-droid:
I’m going to close this PR for now and plan to reopen it later once I’ve refined the approach a bit more.

Thanks!

PR #8488, opened 2026-01-26
Test out new external-publish for upload to pypi

### LINKED ISSUE #8394: How we build a universal wheel leads to a potential upload error

Since we build both binary wheels (with ANA) and a universal wheel (without ANA) we have this config in our CI:

```
  publish:
    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@v2
    with:
      sdist: false
      ...

  publish_pure:
    needs: [publish]
    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@v2
    with:
      env: |
        SUNPY_NO_BUILD_ANA_EXTENSION: 1
    secrets:
```

This leads to this graph:

Image

You will note that the binary wheels get uploaded in the step before the source dist (and universal wheel) is built.

This leads to a condition where we could end up with a release on PyPI without a sdist (and universal wheel) which isn't ideal.

I'm not sure what the easiest fix for this would be when using the reusable workflows.

Cadair:
This should fix #8394

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -164,7 +164,7 @@ jobs:
         SUNPY_NO_BUILD_ANA_EXTENSION: 1
 
   upload:
-    # if: startsWith(github.ref, 'refs/tags/v')
+    if: startsWith(github.ref, 'refs/tags/v')
     name: Upload to PyPI
     runs-on: ubuntu-latest
     needs: [publish, publish_pure]


.github/workflows/ci.yml
```diff
@@ -146,21 +148,42 @@ jobs:
...
       test_extras: 'tests'
       test_command: 'pytest -p no:warnings --doctest-rst -m "not mpl_image_compare" --pyargs sunpy'
       submodules: false
       env: |
         SUNPY_NO_BUILD_ANA_EXTENSION: 1
-    secrets:
-      pypi_token: ${{ secrets.pypi_token }}
+
+  upload:
+    # if: startsWith(github.ref, 'refs/tags/v')
```
Cadair:
```suggestion
    if: startsWith(github.ref, 'refs/tags/v')
```

PR #8493, opened 2026-01-29
Update scraper to properly handle local searches on windows

samaloney:
## PR Description

URIs are of the form `file:///C:/path/to/dir` where `file://` is the scheme and `/C:/path/to/dir` is the path or `/path/to/dir` on linux or Mac. The old code relied on the URIs with the leading `/` removed from the path part so `os.listdir(C:/path/to/dir)`
would work as `os.listdir(/C:/path/to/dir)` doesn't. These are not correct URIs and is inconsistent with the internal prefix used when on windows in the code.

This PR modifies the code to use proper URIs through out and adds additional assert in an existing test to test this behaviour.

### DIFF
--- a/changelog/8493.bugfix.rst
+++ b/changelog/8493.bugfix.rst
@@ -1,0 +1,1 @@
+Fix ~`sunpy.net.scraper.Scraper` to correctly handle proper local file URI on Windows and make internal code more consistent.


samaloney:
I don't think the test fails are related to changes in this PR

samaloney:
Would be good to get this back ported too

Cadair (approved):

Cadair:
This is challenging to backport to 7.0, so I'm not going to unless someone specifically requests it.

Issue #8494, opened 2026-01-29
Stacking coordinates, some with distance and some without leads to weird errors


Cadair:
### Describe the bug

This is probably not actually a sunpy bug, but I'm opening this issue to track it and also because of `HelioProjective.make_3d()` this is an issue which will probably disproportionately affect us.

### To Reproduce

Take the following two frames, one has distance and one doesn't:

```python
start_coord = SkyCoord(10*u.arcsec, 10*u.arcsec, obstime="2026-01-01", observer="earth", frame="helioprojective").frame
next_coord = SkyCoord(10*u.arcsec, 10*u.arcsec, obstime="2026-01-01", observer="earth", frame="helioprojective").frame.make_3d()
```

Doing this:
```python
SkyCoord([start_coord, next_coord])
```
Works:
```
<SkyCoord (Helioprojective: obstime=2026-01-01T00:00:00.000, rsun=695700.0 km, observer=<HeliographicStonyhurst Coordinate for 'earth'>): (Tx, Ty) in arcsec
    [(10., 10.), (10., 10.)]>
```
(All be it drops the distance).

Doing this however:

```
SkyCoord([next_coord, start_coord])
```
Does not work
```
---------------------------------------------------------------------------
UnitConversionError                       Traceback (most recent call last)
File ~/.local/share/mamba/envs/sunpy-dev/lib/python3.14/site-packages/astropy/units/quantity.py:964, in Quantity.to_value(self, unit, equivalencies)
    963 try:
--> 964     scale = self.unit._to(unit)
    965 except Exception:
    966     # Short-cut failed; try default (maybe equivalencies help).

File ~/.local/share/mamba/envs/sunpy-dev/lib/python3.14/site-packages/astropy/units/core.py:624, in UnitBase._to(self, other)
    622         return self_decomposed.scale / other_decomposed.scale
--> 624 raise UnitConversionError(f"'{self!r}' is not a scaled version of '{other!r}'")

UnitConversionError: 'Unit(dimensionless)' is not a scaled version of 'Unit("AU")'

During handling of the above exception, another exception occurred:

UnitConversionError                       Traceback (most recent call last)
Cell In[42], line 1
----> 1 SkyCoord([next_coord, start_coord])

File ~/.local/share/mamba/envs/sunpy-dev/lib/python3.14/site-packages/astropy/coordinates/sky_coordinate.py:229, in SkyCoord.__init__(self, copy, *args, **kwargs)
[...]

PR #8503, opened 2026-02-09
Add to_2d() method to SunPyBaseCoordinateFrame (#8494)

Maya-Mohamed:
This PR adds a to_2d() method to the SunPyBaseCoordinateFrame class. This method allows users to easily convert a 3D coordinate frame to a 2D representation by dropping the distance component.
This is particularly useful for resolving UnitConversionError when attempting to stack SkyCoord objects where some frames have distance information (e.g., after calling make_3d()) and others do not. This issue was highlighted in #8494 and affects downstream tools like reproject.


Cadair:
This PR isn't the solution we are looking for so I'm closing it.

PR #8509, opened 2026-02-12
correction onf log(0) situation in drawing

yaochengchen:
Fixes HTML rendering for GenericTimeSeries when the data contains zero values.

If a log y-scale is applied while the series includes 0 values, taking log(0) is undefined (→ -inf) and matplotlib may warn/fail because the data cannot be log-scaled. This PR avoids forcing log scaling in _repr_html_() for data that contains 0 values, preventing invalid log transforms and the resulting warnings.

Adds a regression test to ensure _repr_html_() does not emit log-scaling warnings when the timeseries values has zero.

### DIFF
--- a/sunpy/timeseries/tests/test_timeseriesbase.py
+++ b/sunpy/timeseries/tests/test_timeseriesbase.py
@@ -1,6 +1,7 @@
 import re
 import copy
 import datetime
+import warnings
 from collections import OrderedDict
 
 import numpy as np
@@ -19,10 +20,10 @@ import sunpy.timeseries
 from sunpy.tests.helpers import figure_test
 from sunpy.time import TimeRange, parse_time
 from sunpy.timeseries import TimeSeriesMetaData
+from sunpy.timeseries.timeseriesbase import GenericTimeSeries
 from sunpy.util import SunpyUserWarning
 from sunpy.util.metadata import MetaDict
-from sunpy.timeseries.timeseriesbase import GenericTimeSeries
-import warnings
+
 # Test fixtures are in ../conftest.py
 
 


yaochengchen:
Thoses failing tests are no from my modifications.

sunpy/timeseries/tests/test_timeseriesbase.py
```diff
@@ -556,6 +558,27 @@ def test_timeseries_array():
...
+
+    bad = [
+        ww for ww in w
+        if "log-scal" in str(ww.message).lower()
+        and (
+            "no positive values" in str(ww.message).lower()
+            or "cannot be log-scaled" in str(ww.message).lower()
+        )
+    ]
+    assert not bad
```
Cadair, line 580: Our default pytest config (https://github.com/sunpy/sunpy/blob/main/pytest.ini#L52) already converts warnings to errors, so this isn't needed. The test which would have raised the warning should fail before your fix and pass after without any of this.

yaochengchen: Thanks for the review.

After re-examining this more carefully, I realised that the behaviour I was trying to fix is not actually reproducible in a meaningful way.

If the data contain both 0 and positive values (e.g. `[0, 1, 2]`), `set_yscale("log")` does not raise the “cannot be log-scaled / no positive values” warning — matplotlib handles this case.
If the data are all zeros, the failure comes instead from `astropy.visualization.hist()` (Scott bin width on constant data), which is unrelated to the log-scaling logic and already causes the test to fail under our warnings-as-errors configuration.

So this PR does not address a real, isolated issue in practice. I’m happy to close it.

PR #8515, opened 2026-02-18
Fix operator precedence bug in EITMap.is_datasource_for

yaochengchen:
## PR Description

Fixes an operator precedence issue in EITMap.is_datasource_for.

Previously, due to Python’s operator precedence rules (and > or), the LEVEL check only applied to the TELESCOP condition. This caused headers with INSTRUME == "EIT" to be identified as EITMap even when LEVEL == "L1", preventing EITL1Map from correctly claiming Level 1 data.

This PR adds explicit parentheses so that the LEVEL is None condition applies to both INSTRUME and TELESCOP checks.

A regression test is added to ensure that Level 1 headers are not incorrectly identified as EITMap.

### DIFF
--- a/changelog/8512.doc.rst
+++ b/changelog/8512.doc.rst
@@ -1,0 +1,1 @@
+Fix typos and grammar in installation.rst
--- a/docs/tutorial/installation.rst
+++ b/docs/tutorial/installation.rst
@@ -16,7 +16,7 @@ As well as containing a Python installation, this virtual environment provides a
 If you already have Python and ``conda`` working you can skip the next section.
 
 `If you are using Anaconda, we recommend that you uninstall it as the default package channel(s) have a restrictive license which means you might not be able to use it for free <https://sunpy.org/posts/2024/2024-08-09-anaconda/>`__.
-Instead, we recommend that you use miniforge which is a minimal installer that setups ``conda`` with the ``conda-forge`` channel, which is free to use for everyone.
+Instead, we recommend that you use miniforge which is a minimal installer that sets up ``conda`` with the ``conda-forge`` channel, which is free to use for everyone.
 If you are using miniforge, you can skip the next section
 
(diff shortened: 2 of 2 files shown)


changelog/8515.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix incorrect Level 1 detection in `EITMap`.
```
Cadair:
```suggestion
Fix incorrect Level 1 detection in `sunpy.map.sources.EITMap`.
```

sunpy/map/sources/soho.py
```diff
@@ -111,7 +111,7 @@ def _supported_observer_coordinates(self):
     @classmethod
     def is_datasource_for(cls, data, header, **kwargs):
         """Determines if header corresponds to an EIT image"""
-        return (header.get('instrume') == 'EIT' or header.get('telescop') == 'Extreme-ultraviolet Imaging Telescope (EIT)'  and header.get('level') is None)
+        return ((header.get('instrume') == 'EIT' or header.get('telescop') == 'Extreme-ultraviolet Imaging Telescope (EIT)') and header.get('level') is None)
```
samaloney, line 114: Should this be anything but '1'?

yaochengchen: The level is None check is pre-existing logic; this PR only fixes operator precedence so that it behaves as originally intended.

Notably, EITL1Map.is_datasource_for already uses (A or B) and header.get('level') == "L1", not A or (B and ...), which indicates the intended pattern is (A or B) and C.

This change simply makes EITMap consistent with that structure.

Cadair: > Should this be anything but '1'?

I think really that `EITMap` is for level 0 data, if some level 1.5 or 2 or some other magical data appeared it'd probably be best to use the level 1 map as the base rather than this, or you'd need a new class anyway.

PR #8516, opened 2026-02-18
fix SymLogNorm calling bug in gong.py for matplotlib > 3.2

yaochengchen:
## PR Description

Fix a Matplotlib compatibility warning in GONGMagnetogramMap by using explicit linthresh for SymLogNorm.

### DIFF
--- a/changelog/8516.bugfix.rst
+++ b/changelog/8516.bugfix.rst
@@ -1,0 +1,1 @@
+Fix a Matplotlib compatibility warning in GONGMagnetogramMap by using explicit linthresh for SymLogNorm.


Cadair:
Can you provide some reference to why passing this as a positional argument is deprecated? From a look at the mpl source this seems fine?

yaochengchen:
> Can you provide some reference to why passing this as a positional argument is deprecated? From a look at the mpl source this seems fine?

Thanks for checking. I re-examined my code and realized the issue was actually due to a different mistake on my side. I initially relied on AI analysis, which led me to the wrong conclusion about `SymLogNorm`.

Sorry for the confusion and for wasting your time. I will close this PR.

PR #8517, opened 2026-02-19
Switch pre-commit to use file and type or type_or rather than exclude

samaloney:
Switch to specifying the type/s and file/s we want the checks to run on rather than excluding those that we don't. I think this is more explicit and marginally more maintainable. If we add `.md` files just need to add `markdown` to the `types` or `types_or`

Related to https://github.com/sunpy/package-template/pull/248 and https://github.com/sunpy/package-template/issues/247

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,6 +28,8 @@ repos:
         exclude: *exclude_sunpy_dirs
       - id: check-yaml
         types: [yaml]
+      - id: check-toml
+        types: [toml]
       - id: debug-statements
         types: [python]
         exclude: *exclude_sunpy_dirs


samaloney:
Not sure if this really needs a change log?

.pre-commit-config.yaml
```diff
@@ -1,32 +1,47 @@
...
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v6.0.0
     hooks:
       - id: check-ast
+        types: [python]
       - id: check-case-conflict
+        types: [python]
       - id: trailing-whitespace
+        files: ^sunpy/(?!(data|extern|io/src/ana)/)
+        types_or: [python, rst]
```
samaloney, line 27: Any other file types we want to do this on?

.pre-commit-config.yaml
```diff
@@ -1,32 +1,47 @@
...
       - id: check-yaml
+        types: [yaml]
       - id: debug-statements
+        types: [python]
       - id: check-added-large-files
         args: ["--enforce-all", "--maxkb=1054"]
-        exclude: ""
       - id: end-of-file-fixer
+        types_or: [python, rst]
```
samaloney, line 39: Any other file types we want to do this on?

.pre-commit-config.yaml
```diff
@@ -1,32 +1,47 @@
...
+        files: ^sunpy/(?!(data|extern|io/src/ana)/)
       - id: mixed-line-ending
-  - repo: https://github.com/adhtruong/mirrors-typos
-    rev: v1.38.1
+        types: [text]
+        files: ^sunpy/(?!(data|extern|io/src/ana)/)
+  - repo: https://github.com/crate-ci/typos
```
samaloney, line 44: I don't think this mirror repo is required anymore https://github.com/crate-ci/typos/blob/master/docs/pre-commit.md

.pre-commit-config.yaml
```diff
@@ -1,32 +1,47 @@
...
+    rev: v1.34.0
     hooks:
       - id: typos
+        types_or: [python, rst]
```
samaloney, line 48: If I leave this as `text` we pick up some mistake or at least what it think are mistake in some json files

samaloney:
Move `^sunpy/(?!(data|extern|io/src/ana)/)` to top level and see if can reduce to only `extern`

samaloney:
Ok so it seems that any top level `files` or `exclude` can only be added to not overwritten. So taken a different approach with a yaml anchor so the regex is shared. Trying to commit a large file will now cause a fail. I've also enabled the toml check.

If I remove the `tools` folder from the exclude we get these fails not sure if we care to fix these or not.

```
pre-commit run -a                                                                                                                                                                    ✔  sunpy-dev 
ruff (legacy alias)......................................................Failed
- hook id: ruff
- exit code: 1
- files were modified by this hook

E722 Do not use bare `except`
  --> tools/version_bumps.py:25:9
   |
23 |         try:
24 |             version = Version(ver)
25 |         except:
   |         ^^^^^^
26 |             continue
27 |         release_date = None
   |

Found 2 errors (1 fixed, 1 remaining).

isort....................................................................Failed
- hook id: isort
- files were modified by this hook

Fixing /Users/sm/Projects/sunpy-dev/tools/update_attrs_json.py
Fixing /Users/sm/Projects/sunpy-dev/tools/version_bumps.py
Skipped 3 files

check python ast.........................................................Passed
check for case conflicts.................................................Passed
trim trailing whitespace.................................................Passed
check yaml...............................................................Passed
debug statements (python)................................................Passed
check for added large files..............................................Passed
fix end of files.........................................................Passed
mixed line ending........................................................Passed
typos....................................................................Passed
generate sunpy.net.hek.attrs.............................................Passed
git diff.................................................................Failed
- hook id: git-diff
- exit code: 1

diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index fe1990d23..88c4f94af 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,7 +7,7 @@ repos:
         args: ["--fix"]
         types: [python]
         # Define here once and then reference YAML achore
-        exclude: &exclude_sunpy_dirs ^(sunpy/(data|extern)/|tools/)
+        exclude: &exclude_sunpy_dirs ^sunpy/(data|extern)
   - repo: https://github.com/PyCQA/isort
     rev: 7.0.0
     hooks:
diff --git a/tools/update_attrs_json.py b/tools/update_attrs_json.py
index c678c6665..3f9e941ca 100755
--- a/tools/update_attrs_json.py
+++ b/tools/update_attrs_json.py
@@ -3,8 +3,8 @@
 import sunpy
 from sunpy.net.cdaweb.helpers import _update_cdaweb_dataset_data
 from sunpy.net.jsoc import JSOCClient
-from sunpy.net.vso import VSOClient
 from sunpy.net.solarnet import SOLARNETClient
+from sunpy.net.vso import VSOClient

 print(f"Updating the attrs json files using sunpy {sunpy.__version__}...")

diff --git a/tools/version_bumps.py b/tools/version_bumps.py
index 9ea01c3fa..42f13e8db 100644
--- a/tools/version_bumps.py
+++ b/tools/version_bumps.py
@@ -1,12 +1,12 @@
+import argparse
 import collections
-from packaging.version import Version
-import requests
 from datetime import datetime, timedelta
-from dateutil.parser import parse as parse_date
+from functools import cache
 from importlib import metadata
+
+import requests
 from packaging.requirements import Requirement
-from functools import cache
-import argparse
+from packaging.version import Version


 @cache
 ```

